PR #7321, opened 2018-04-21
configure: make -annotate fatal, and color error and warnings

SkySkimmer:
Warnings are just too hard to see. Also there is no point keeping a
noop option except to point people at the replacement, which does not require configure to succeed.

### DIFF
--- a/configure.ml
+++ b/configure.ml
@@ -21,11 +21,18 @@ let distributed_exec = ["coqtop";"coqc";"coqchk";"coqdoc";"coqworkmgr";
 
 let verbose = ref false (* for debugging this script *)
 
+let red, yellow, reset =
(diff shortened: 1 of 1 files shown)


SkySkimmer:
(prompted by a recent coqclub thread where someone was confused due to not noticing the -annotate warning)

Zimmi48:
@ejgallego Can you take over as the secondary maintainer for `./configure`?

ejgallego:
> @ejgallego Can you take over as the secondary maintainer for ./configure?

No problem.

ejgallego (approved):
I think this is ready to merge, @SkySkimmer ?

SkySkimmer:
Should be.

PR #7317, opened 2018-04-20
Fix #6798: coqchk ignores ugraph when comparing constant instances

SkySkimmer:
Fixes #6798.

PR #7328, opened 2018-04-22
Fix #7327: coqchk subtyping of polymorphic constants

### LINKED ISSUE #7327: coqchk subtyping of polymorphic constants broken
This code path is a bit annoying to get to. Original trigger in HoTT after #7325 #7317.
```coq
Set Universe Polymorphism.

Module Type T.
  Axiom foo : Type.
  Parameter bar@{u v|u = v} : foo@{u}.
End T.

Module M.
  Axiom foo : Type.
  Axiom bar@{u v|u = v} : foo@{v}. (* not syntactically equal to foo@{u}, forcing a lookup in the ugraph *)
End M.

Module F (A:T). End F.

Module X := F M.
```
causes 
```
Fatal Error: Anomaly "Universe Var(1) undefined."
```

SkySkimmer:
Depends on #7317 (to be able to make the test work).

### DIFF
--- a/checker/closure.ml
+++ b/checker/closure.ml
@@ -135,22 +135,16 @@ let betaiotazeta = mkflags [fBETA;fIOTA;fZETA]
  * instantiations (cbv or lazy) are.
  *)
 
-type 'a tableKey =
-  | ConstKey of 'a
-  | VarKey of Id.t
-  | RelKey of int
-
 type table_key = Constant.t puniverses tableKey
+
+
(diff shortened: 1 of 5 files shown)


ppedrot (approved):
Module subtyping is a real can of worms. There is a plentiful amount of suspicious code in there. At least the current fix seems correct to me.

ppedrot:
After being done with #7317 the first thing I'll do is merge this one, so shout out if you're not happy with the fix.

PR #6556, opened 2018-01-06
Remove dir-locals and ship suggested helper hooks instead.

SkySkimmer:
Alternative to #1136.

People interested in those functions can copy `coqdev.el` somewhere and load it, or copy parts of it to their init. Those not interested can just ignore it.

PR #6372, opened 2017-12-10
[make] Move makefiles to their own subdirectory.

ejgallego:
This helps with #6137. I didn't complicate the system a lot, just
added a new makefile prefix to the proper places.

PR #7337, opened 2018-04-24
dir-locals: add bug-reference-mode variables

SkySkimmer:
This makes it so that after `M-x bug-reference-mode` when you see `#4567` in the code it gets turned into a link (open with `C-c RET`) leading to #4567.

### DIFF
--- a/dev/tools/coqdev.el
+++ b/dev/tools/coqdev.el
@@ -103,5 +103,17 @@ Note that this function is executed before _Coqproject is read if it exists."
      2 (3 . 4) (5 . 6)))
   (add-to-list 'compilation-error-regexp-alist 'coq-backtrace))
 
+(defvar bug-reference-bug-regexp)
+(defvar bug-reference-url-format)
+(defun coqdev-setup-bug-reference-mode ()
+  "Setup `bug-reference-bug-regexp' and `bug-reference-url-format' for Coq.
+
+This does not enable `bug-reference-mode'."
+  (let ((dir (coqdev-default-directory)))
+    (when dir
+      (setq-local bug-reference-bug-regexp "#\\(?2:[0-9]+\\)")
+      (setq-local bug-reference-url-format "https://github.com/coq/coq/issues/%s"))))
+(add-hook 'hack-local-variables-hook #'coqdev-setup-bug-reference-mode)
+
 (provide 'coqdev)
 ;;; coqdev ends here


Zimmi48 (dismissed):
You just got rid of `.dir-locals.el` in #6556. Why recreate it now?

SkySkimmer:
Those were setting non-safe variables and/or using dynamic values (coq absolute path) which doesn't work well in dir-locals. This OTOH is well suited to dir-locals.

Zimmi48:
Still, I don't see the point of cluttering the root folder more when all the people interested in this function are probably already requiring the `coqdev.el` file anyways.

ejgallego:
This seems blocked, as a heavy emacs user I will take it unless someone volunteers soon.

ejgallego:
By the way I am in favor of having `.dir-locals.el` back; this is a dot file and clutter, which I dislike, is hard a reason to complain when other easier stuff such as #6372 is not included.

SkySkimmer:
>I still fail to see the point of reintroducing this file. It is inconsistent with the many functions that are safe as well and are part of coqdev.el.

Which functions are these? Recall that using the Coq directory requires arbitrary code execution ie is unsafe.

Zimmi48:
This one I suppose:

```
;; This Elisp snippet adds a regexp parser for the format of Anomaly
;; backtraces (coqc -bt ...), to the error parser of the Compilation
;; mode (C-c C-c: "Compile command: ..."). File locations in traces
;; are recognized and can be jumped from easily in the *compilation*
;; buffer.
(defvar compilation-error-regexp-alist-alist)
(defvar compilation-error-regexp-alist)
(with-eval-after-load 'compile
  (add-to-list
   'compilation-error-regexp-alist-alist
   '(coq-backtrace
     "^ *\\(?:raise\\|frame\\) @ file \\(\"?\\)\\([^,\" \n\t<>]+\\)\\1,\
      lines? \\([0-9]+\\)-?\\([0-9]+\\)?\\(?:$\\|,\
      \\(?: characters? \\([0-9]+\\)-?\\([0-9]+\\)?:?\\)?\\)"
     2 (3 . 4) (5 . 6)))
  (add-to-list 'compilation-error-regexp-alist 'coq-backtrace))
```

And also the ones which are added in https://github.com/coq/coq/pull/7391/files#diff-262bc68bf79f9b22e2e939b2016fe836

SkySkimmer:
>This one I suppose:

Needs add-to-list, and those variables aren't marked safe.

>And also the ones which are added in https://github.com/coq/coq/pull/7391/files#diff-262bc68bf79f9b22e2e939b2016fe836

That's a function to be called interactively.

ejgallego:
The argument of "too many files" is IMHO just a soft one; one file is not gonna make a lot of difference, so IMHO it shouldn't block this; as for whether `.dir-locals` is the best technical solution, I'm still thinking.

ejgallego:
Well, given that it is only to variables, it seems indeed a bit overkill to readd a file. @SkySkimmer , what is the problem with adding this to `coqdev.el` ?

SkySkimmer:
coqdev.el doesn't apply to everyone, whereas this change should (bug-reference-mode is not enabled by default so people who don't want it won't see anything).
coqdev.el is heavier, having to mess with hooks and setq-local.

ejgallego:
Yeah that I figure it out; I dunno how to proceed here; I kind of understand both your point of view and @Zimmi48 's point of view. I'd say if @Zimmi48 is hard set on the "no more files" let's add this to the heavier version, otherwise we merge as-is. @Zimmi48 ?

ejgallego:
Indeed a file for just two variables seems like too much noise for little gain, thus I suggest @SkySkimmer to implement Théo's solution if he has a free minute and doesn't mind.

SkySkimmer:
>Indeed a file for just two variables seems like too much noise for little gain, thus I suggest @SkySkimmer to implement Théo's solution if he has a free minute and doesn't mind, anyways you would need to rebase to add the code owners and I think I do see the advantage of having a centralized emacs setup.

Changed.

ejgallego:
Will test ASAP.

ejgallego (approved):
Works for me (TM)

PR #7338, opened 2018-04-24
[api] Move `hint_info_expr` to `Typeclasses`.

ejgallego:
`hint_info_expr`, `hint_info_gen` do conceptually belong to the
typeclasses modules and should be able to be used without a dependency
on the concrete vernacular syntax.

### DIFF
--- a/parsing/g_vernac.ml4
+++ b/parsing/g_vernac.ml4
@@ -645,7 +645,7 @@ GEXTEND Gram
 
       | IDENT "Existing"; IDENT "Instances"; ids = LIST1 global;
         pri = OPT [ "|"; i = natural -> i ] ->
-         let info = { hint_priority = pri; hint_pattern = None } in
+         let info = { Typeclasses.hint_priority = pri; hint_pattern = None } in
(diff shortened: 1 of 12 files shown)


herbelin (approved):
A new step in a more rational and modular organization. 👍 

@mattam82 seems to be the canonical assignee, right?

ppedrot:
@mattam82 doesn't look very responsive so I'm assigning myself.

ppedrot (approved):
LGTM. Actually, I'd even go to the point where we should split this type into one dedicated to hints and the other to the vernac AST, but thay might be a slightly extremist point of view.

ejgallego:
> that might be a slightly extremist point of view.

This is not so extremist, there seem to be a couple of things we could refine in this sense, and interpretation is called sometimes a bit too late in the process.

Also, typeclasses is in the pretyper but using the Term AST, so yeah, even if code is more verbose we should aim for a complete phase-separation IMVHO.

ppedrot:
This is a common sense PR, I don't think we formally need @mattam82's approval. I'll let you until tomorrow to complain just in case, otherwise I'll merge this PR.

PR #7340, opened 2018-04-25
Remove DirClosedSection.

jashug:
This has been around for at least 16 years, with the comment
"this won't last long I hope".
https://github.com/coq/coq/commit/12965209478bd99dfbe57f07d5b525e51b903f22#diff-1a3a6f7bd5b2cf1bc6dd43ee04bbc3eaR112

**Kind:** simplification

### DIFF
--- a/library/lib.ml
+++ b/library/lib.ml
@@ -580,13 +580,11 @@ let close_section () =
   in
   let (secdecls,mark,before) = split_lib_at_opening oname in
   lib_state := { !lib_state with lib_stk = before };
-  let full_olddir = !lib_state.path_prefix.obj_dir in
   pop_path_prefix ();
   add_entry oname (ClosedSection (List.rev (mark::secdecls)));
   let newdecls = List.map discharge_item secdecls in
   Summary.unfreeze_summaries fs;
-  List.iter (Option.iter (fun (id,o) -> add_discharged_leaf id o)) newdecls;
-  Nametab.push_dir (Nametab.Until 1) full_olddir (DirClosedSection full_olddir)
+  List.iter (Option.iter (fun (id,o) -> add_discharged_leaf id o)) newdecls
 
 (* State and initialization. *)
 
--- a/library/libnames.ml
+++ b/library/libnames.ml
@@ -174,8 +174,6 @@ type global_dir_reference =
   | DirOpenModtype of object_prefix
   | DirOpenSection of object_prefix
   | DirModule of object_prefix
-  | DirClosedSection of DirPath.t
-      (* this won't last long I hope! *)
 
 let eq_op op1 op2 =
   DirPath.equal op1.obj_dir op2.obj_dir &&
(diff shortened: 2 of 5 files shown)


herbelin:
If I understand correctly, the effect of this PR is only that the error message when doing `Print Section S` on a closed section `S` will display `Unknown section` rather than `Cannot print the contents of a closed section`.

To be honest, I forgot that this command existed. It was useful in a terminal-style interaction but it is not useful in a document-based interaction.

Maybe `Print Section` could be removed fully, for simplification. This is the kind of question that may be worth to be asked to a panel of users.

In any case, I'm with this PR. The message `Unknown section` could then be changed into `No such open section`.

jashug:
One other effect this has is it allows you to define multiple sections with the same name.

herbelin:
> One other effect this has is it allows you to define multiple sections with the same name.

Ah, and is this perceived as rather good or rather bad?

ejgallego:
Note that sections do qualify names for a while, there was some discussion recently; what to think about is a different story.

jashug:
Sections only qualify names while they are open. So
```
Section A.
Definition foo := 1.
Print A.foo.
End A.
Print foo.
Fail Print A.foo.
```

One of the ways to make sections more lightweight would be to not have sections qualify names ever, and I'm looking into that option, but shouldn't impact this PR specifically.

@ejgallego Do you happen to have a link to that discussion?

jashug:
I understand @silene is on vacation, as secondary maintainer, @gares, would you take a look?

gares (approved):

herbelin:
> @herbelin are you OK with this PR being merged? If so I'll take care of this.

Allowing defining multiple sections with the same name would be a change of habits and I have no idea though whether this is more desirable than the converse.

I find unnamed sections interesting (as a pair of matching parentheses) but I feel that allowing several sections with same name is a bit counterintuitive. Probably the question is also about what the section name is supposed to convey. Is it only for working in a local context? Is it also for sectioning the file into parts? But then, different level of comments (e.g. `(** *`, `(** **`, etc.) are maybe just enough. 

Having to close a section for "technical reasons" and reopening it with the same name looks to me like a weakness of the system. Couldn't the "technical reason" be addressed without closing/reopening the section?

To make quick progresses on the PR, implementing a warning telling a section is reusing a name looks to me like a good compromise.

herbelin (commented):
Summary as a review: technically I'm ok with the PR. In the intend, I'm a bit bothered by the change of semantics and I'd be interested in seeing this change integrated into a more general reflection on section. (Or, in the meantime, to have a warning on a section with repeated name.)

gares:
Hum, but that warning would be triggered exactly by the use case I had in mind (eg use the same name to signal that, one day, the two sections should be merged).

In other words I think I prefer `Section Foo. ... End. Section Foo. ... End` to `Section Foo. End. Section FooBis. ... End.` if the two sections are, morally, about the same concept.

If you don't oppose I'd merge it without the warning (i.e. as it is now). An one can add an issue/pr to gather more opinions on the subject. Are you ok with this plan?

herbelin:
> If you don't oppose I'd merge it without the warning (i.e. as it is now). And one can open an issue/pr to gather more opinions on the subject. Are you OK with this plan?

OK, let's go this way (but that would be interesting to know what are the "technical reasons" you are referring to; is it about generalizing a definition from within a section, about exporting a notation across a section, ...).

gares:
Yep, the fact that definitions are only abstracted at the end of the section, that means that, polymorphic lists
are not polymorphic until you close their section (just to give a silly example). In large proofs the context can be large, and reopening the section has a visible cost (other than a purely aesthetic one).

OK, I'll merge this one then

PR #7341, opened 2018-04-25
Don't recurse into closed modules/sections in split_lib.

jashug:
Putting `assert false` in the successful recursive case never triggered.
Apparently all users use `split_lib_at_opening` to find open at current nesting level?
`split_lib` appears to be dead code currently, might also be candidate for removal.
Doing so would allow to simplify `split_lib_gen`, since we only expect one matching element.

**Kind:** simplification.

### DIFF
--- a/library/lib.ml
+++ b/library/lib.ml
@@ -183,22 +183,11 @@ let split_lib_gen test =
     | before -> after,equal,before
   in
   let rec findeq after = function
-    | hd :: before ->
-      	if test hd
-	then Some (collect after [hd] before)
-	else (match hd with
-		| (sp,ClosedModule  seg)
-		| (sp,ClosedSection seg) ->
-		    (match findeq after seg with
-		       | None -> findeq (hd::after) before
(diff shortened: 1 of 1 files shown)


herbelin:
It seems that you are right, and that `split_lib` (`contents_after`) is not used anymore. The best person to know is probably @gares.

(For instance, I thought `split_lib` was used for `Reset` but not anymore, so a post-stm cleanup of `lib.ml` since indeed worth.)

ejgallego:
> (For instance, I thought split_lib was used for Reset but not anymore, so a post-stm cleanup of lib.ml since indeed worth.)

Indeed, there are other parts that need a bit of work to be better integrated with the top-level document manager such as the save/vo path.

jashug:
@gares ping: any thoughts on this?

gares (approved):
I checked at the history and this code was added to let one `BackTo` into a section. It can be safely removed, since all that is now taken care by the STM.

PR #7301, opened 2018-04-18
[sphinx] Backport forgotten updates during the migration & other improvements

Zimmi48:
Backport forgotten updates during the migration + some other Sphinx fixes.

**Kind:** documentation

Fixes / closes #7123.

It was less bad than I expected.

PR #7343, opened 2018-04-25
Add missing index entries.

Zimmi48:
This commit was originally part of #7301 but it reveals a bug in the Sphinx architecture.

cc @cpitclaudel

### DIFF
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -216,6 +216,11 @@ numbers (see :ref:`datatypes`).
    Negative integers are not at the same level as :token:`num`, for this
    would make precedence unnatural.
 
+.. index::
+   single: Set (sort)
+   single: Prop
+   single: Type
+
 Sorts
 -----
 
@@ -262,6 +267,8 @@ fun and forall gets identical. Moreover, parentheses can be omitted in
 the case of a single sequence of bindings sharing the same type (e.g.:
 :g:`fun (x y z : A) => t` can be shortened in :g:`fun x y z : A => t`).
 
+.. index:: fun ... => ...
+
 Abstractions
 ------------
 
@@ -281,6 +288,8 @@ fun :token:`ident`\ :math:`_{n}` : :token:`type` => :token:`term`”. If
 a let-binder occurs in
 the list of binders, it is expanded to a let-in definition (see
 Section :ref:`let-in`).
+
+.. index:: forall
 
 Products
 --------
@@ -320,6 +329,11 @@ The notation :n:`(@ident := @term)` for arguments is used for making
 explicit the value of implicit arguments (see
 Section :ref:`explicit-applications`).
 
+.. index::
+   single: ... : ... (type cast)
+   single: ... <: ...
+   single: ... <<: ...
+
 Type cast
 ---------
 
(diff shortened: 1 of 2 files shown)


doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -217,6 +217,11 @@ numbers (see :ref:`datatypes`).
 Sorts
 -----
 
+.. index::
+   single: Set
```
Zimmi48: Adding this index entry reveals a bug:
The next index entry `Set (cmd)` now appears as `(cmd)`.
On the other hand, if I put `Set (sort)` here, `Set (cmd)` is displayed normally but `Set (sort)` is displayed as `(sort)`.
I have observed the same bug with the two indices for `term` (which I added in this PR).

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -330,6 +343,8 @@ the type of :token:`term` to be :token:`type`.
 Inferable subterms
 ------------------
 
+.. index:: _
```
Zimmi48: This does not appear in the symbol section of the index but in a new `_` section. Not great but not a critical problem either.

doc/sphinx/language/gallina-specification-language.rst
cpitclaudel: I investigated this, and I'm not sure it's a bug :)

It's definitely a rendering issue, though; it currently looks confusing.  It comes from here: https://github.com/sphinx-doc/sphinx/blob/93837bb01ecb185d8ee64b1f794f010c510b32e1/sphinx/environment/adapters/indexentries.py#L114

We could either monkey-patch or open an issue to make this customizable.

doc/sphinx/language/gallina-specification-language.rst
cpitclaudel: Why not write this as `single: _` instead?

cpitclaudel: Oh, wait, I see what you mean.  Please ignore the noise

Zimmi48:
More generally, I don't really like the current situation of the index pages (cf. the general index with symbols at the beginning and at the end, the tactic index, and the error index).

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -321,6 +330,10 @@ Section :ref:`explicit-applications`).
 Type cast
 ---------
 
+.. index::
+   single: ... : ... (type cast)
+   single: ... <: ... (type cast)
```
maximedenes: Can you also add `... <<: ...`? Shouldn't we explain what those casts mean, though?

maximedenes (changes requested):
Looks good modulo one change.

doc/sphinx/language/gallina-specification-language.rst
Zimmi48: Of course we should. See #7846.

maximedenes:
@Zimmi48 does this still need an infrastructure fix?

Zimmi48:
Let me rebase this and let's evaluate what we think of the rendered result.

cpitclaudel (approved):

Zimmi48:
Rendered: https://coq.gitlab.io/-/coq/-/jobs/98524916/artifacts/_install_ci/share/doc/coq/sphinx/html/genindex.html

I'm still not convinced by the rendered output of `Set` and `let ... := ...` though.

Zimmi48:
I modified the PR so that indexes point to the section titles and not the text below.
I've also added a small doc of `<<:`. @maximedenes can you review again?
Rendered: https://coq.gitlab.io/-/coq/-/jobs/98558916/artifacts/_install_ci/share/doc/coq/sphinx/html/genindex.html.
I'm still not convinced by the way `Set` and `let ... := ...` are rendered but I think this PR can be merged and I'll open an issue on all the problem of the indexes I can see.

maximedenes (approved):

Issue #7171, opened 2018-04-04
Windows CI: add check that installer works


MSoegtropIMC:
CI currently builds a Windows installer, but doesn't check it in any way.
The Windows CI should run the installer and compare the install destination folder with the "template" folder.

MSoegtropIMC:
The Coq Platform CI does a smoke test for all delivered modules using the installed Coq.

MSoegtropIMC:
The PR which removed the old build scripts and switched to Coq Platform was (#13598).

PR #6962, opened 2018-03-10
[ci] [gitlab] Add some tags for fast runner selection.

ejgallego:
This allows jobs to be picked by workers tagged `fast-runner`.

PR #6057, opened 2017-11-02
Start a release process documentation.

Zimmi48:
These notes were largely inspired by [the release process documentation of NixOS](https://nixos.org/nixos/manual/index.html#release-process).

It is only a beginning, a lot of things are missing. When this is finished this should allow to check one point of #5976.
I am creating this now, so that we don't forget again to put the alpha tag when branching.

PR #7344, opened 2018-04-25
Windows packaging build with Gitlab CI

maximedenes:
This PR makes it possible to use a specific runner on Inria CloudStack to automatically generate signed Windows packages (for protected branches, PRs will generate unsigned packages once we set up the Gitlab fork for PR mirroring).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -134,6 +134,24 @@ before_script:
     OPAM_SWITCH: "edge"
     OPAM_VARIANT: "+flambda"
 
+.windows-template: &windows-template
+  stage: test
+  artifacts:
+    name: "%CI_JOB_NAME%"
+    paths:
+      - dev\nsis\*.exe
+      - coq-opensource-archive-windows-*.zip
+    expire_in: 1 week
+  dependencies: []
+  tags:
+    - windows
+  before_script: []
+  script:
+    - call dev/ci/gitlab.bat
+  only:
+    variables:
+      - $WINDOWS == "enabled"
+
 build:base:
   <<: *build-template
   variables:
@@ -159,6 +177,16 @@ build:edge+flambda:
     OPAM_VARIANT: "+flambda"
     COQ_EXTRA_CONF: "-native-compiler no -coqide opt -flambda-opts "
     COQ_EXTRA_CONF_QUOTE: "-O3 -unbox-closures"
+
+windows64:
+  <<: *windows-template
+  variables:
+    ARCH: "64"
+
+windows32:
+  <<: *windows-template
+  variables:
+    ARCH: "32"
 
 warnings:base:
   <<: *warnings-template
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -970,6 +970,10 @@ function make_lablgtk {
     # These changes are included in dev/build/windows/patches_coq/lablgtk-2.18.3.patch
 
     log2 make world
+
+    # lablgtk does not escape FINDLIBDIR path, which can contain backslashes
+    sed -i "s|^FINDLIBDIR=.*|FINDLIBDIR=$PREFIXOCAML/libocaml/site-lib|" config.make
+
     log2 make install
     log2 make clean
     build_post
--- a/dev/ci/gitlab.bat
+++ b/dev/ci/gitlab.bat
@@ -1,0 +1,50 @@
+@ECHO OFF
+
+REM This script builds and signs the Windows packages on Gitlab
+
+if %ARCH% == 32 (
+  SET ARCHLONG=i686
+  SET CYGROOT=C:\cygwin
+  SET SETUP=setup-x86.exe
+)
+
+if %ARCH% == 64 (
+  SET ARCHLONG=x86_64
+  SET CYGROOT=C:\cygwin64
+  SET SETUP=setup-x86_64.exe
+)
+
+powershell -Command "(New-Object Net.WebClient).DownloadFile('http://www.cygwin.com/%SETUP%', '%SETUP%')"
+SET CYGCACHE=%CYGROOT%\var\cache\setup
+SET CI_PROJECT_DIR_MFMT=%CI_PROJECT_DIR:\=/%
+SET CI_PROJECT_DIR_CFMT=%CI_PROJECT_DIR_MFMT:C:/=/cygdrive/c/%
+SET DESTCOQ=C:\coq%ARCH%_inst
+SET COQREGTESTING=Y
+SET PATH=%PATH%;C:\Program Files\7-Zip\;C:\Program Files\Microsoft SDKs\Windows\v7.1\Bin
+
+if exist %CYGROOT%\build\ rd /s /q %CYGROOT%\build
+if exist %DESTCOQ%\ rd /s /q %DESTCOQ%
+
+call %CI_PROJECT_DIR%\dev\build\windows\MakeCoq_MinGW.bat -threads=1 ^
+  -arch=%ARCH% -installer=Y -coqver=%CI_PROJECT_DIR_CFMT% ^
+  -destcyg=%CYGROOT% -destcoq=%DESTCOQ% -cygcache=%CYGCACHE% ^
+  -addon=bignums -make=N ^
+  -setup %CI_PROJECT_DIR%\%SETUP% || GOTO ErrorExit
+
+copy "%CYGROOT%\build\coq-local\dev\nsis\*.exe" dev\nsis || GOTO ErrorExit
+7z a coq-opensource-archive-windows-%ARCHLONG%.zip %CYGROOT%\build\tarballs\* || GOTO ErrorExit
+
+REM DO NOT echo the signing command below, as this would leak secrets in the logs
+IF DEFINED WIN_CERTIFICATE_PATH (
+  IF DEFINED WIN_CERTIFICATE_PASSWORD (
+    ECHO Signing package
+    @signtool sign /f %WIN_CERTIFICATE_PATH% /p %WIN_CERTIFICATE_PASSWORD% dev\nsis\*.exe
+    signtool verify /pa dev\nsis\*.exe
+  )
+)
+
+GOTO :EOF
+
+:ErrorExit
+  ECHO ERROR %0 failed
+  EXIT /b 1
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,10 +10,6 @@ image:
 environment:
   CYGMIRROR: http://ftp.inf.tu-dresden.de/software/windows/cygwin32
   matrix:
-  - USEOPAM: false
-    ARCH: 32
-  - USEOPAM: false
-    ARCH: 64
   - USEOPAM: true
     ARCH: 64
 
@@ -21,11 +17,3 @@ build_script:
 - cmd: 'call %APPVEYOR_BUILD_FOLDER%\dev\ci\appveyor.bat'
 
 test: off
-
-artifacts:
-  - path: 'dev\nsis\*.exe'
-    name: installer
-
-  - path: 'coq-opensource-archive-*.zip'
-    name: opensource-archive
-


.gitlab-ci.yml
```diff
@@ -229,6 +229,23 @@ build:bleeding-edge:
     EXTRA_CONF: "-native-compiler yes -coqide opt"
     EXTRA_PACKAGES: "$COQIDE_PACKAGES"
 
+build:windows:
+  stage: build
```
SkySkimmer: Might be better in stage: test to avoid holding those jobs up.

ejgallego (approved):
Looks good to me.

ejgallego:
Should this be allowed to fail as to be more resilient wrt INRIA Cloudstack going down?

ejgallego:
Let me know when you want this in.

MSoegtropIMC:
@maximedenes : If I understand this right, the secrets are only available on special INRIA owned machines and running CI on these machines requires special privileges. Can you describe what these privileges are and how you ensure that only INRIA folks allowed to sign binaries in the name of INRIA obtain such privileges?

In general I have my doubts that automating signing is worth the effort. Thinking about how to make this secure is a lot of effort - it is just too easy to miss something. As Edgar Allen Poe wrote 175 years back:

> Few persons can be made to believe that it is not quite an easy thing to invent a method of secret writing which shall baffle investigation. Yet it may be roundly asserted that human ingenuity cannot concoct a cipher which human ingenuity cannot resolve..."

The same applies to all security. I think it makes sense to have a human in the loop for signing.

Technically the  changes are fine, but I have the security concerns stated above and these code changes are by far not sufficient to judge the security of this.

Best regards,

Michael

SkySkimmer (dismissed):
Like with #6962 I am concerned that the build will fail for anyone who doesn't have a tagged runner registered. eg https://gitlab.com/SkySkimmer/coq/pipelines/21065247 currently stalled for lack of runner, I'll kill it at some point if it doesn't die by itself.

I have considered the following solutions:
- we just make it easy to request a runner (simple but makes it harder to use gitlab CI for newcomers, I don't like this solution)
- we use `if: $WINDOWS == "enabled"` and people with access to a windows runner put it in their project variables. This works for this PR but not for #6962 as that one tags jobs that the rest depend upon (so not running them breaks everything). cf https://docs.gitlab.com/ee/ci/variables/README.html#variables-expressions https://gitlab.com/help/ci/variables/README#secret-variables
- (can be seen as an extension of the above) we have a `.gitlab-tagged.yml` using `include: .gitlab-ci.yml` and containing tagged copies of the normal jobs and the windows job, and people with tagged runners can set their project to use gitlab-tagged. cf https://gitlab.com/help/user/project/pipelines/settings#custom-ci-config-path

I think 3 is overkill for this PR so I request you implement 2.

SkySkimmer:
Also I set the WIN_CERTIFICATE_PASSWORD variable to protected on gitlab/coq in case a wild `printenv` appears.

SkySkimmer:
Some interesting stuff in this log https://gitlab.com/coq/coq/-/jobs/64917989
- Coq's make install complains about not having native compute stuff to install
- mingw stuff is downloaded again after bignums is finished (why?)

.gitlab-ci.yml
SkySkimmer: The artifact upload seems to take a while so definitely put the job in the most parallel stage.

MSoegtropIMC:
@SkySkimmer : the warnings about native compute a normal - there is no native compute on Windows as yet (I am working on it, it currently would need a fully working bash and this mean a fully working Cygwin)

The download of mingw stuff are the sources. These are part of the open source package created together with the installer. The point of this is that GPL stuff requires that you make the sources included in your binaries accessible - redirections are not allowed. So there is a mechanism which takes care to get each and every source file included in the binary build artefacts, and this means sources of all MinGW libraries.

What concerns me more is this line:

Saving to: ‘V8.8+beta1.zip’

I think I requested this to be fixed ... (it is not a reasonable name for a module source package - it should be renamed). I will create an issue to fix it.

SkySkimmer:
Weird error in travis OSX job.

SkySkimmer:
>WARNING: Uploading artifacts to coordinator... failed  id=64917989 responseStatus=504 Gateway Time-out status=504 Gateway Time-out token=TxH_LzFc
> RETRY x2 failed

The windows artifacts are <500MB on appveyor, ~120MB for the regular gitlab artifacts. So it's not orders of magnitude more but it's still a possible failure.
Does the windows runner have good upload speed?
Maybe splitting the artifacts into multiple archives would workaround the timeout (ie just keep the original archives as-downloaded).

It's concerning that artifact upload failure doesn't fail the job. Maybe we should have a second job depending on this one building some arbitrary development (plus it would solve #7171).

>The download of mingw stuff are the sources. These are part of the open source package created together with the installer. The point of this is that GPL stuff requires that you make the sources included in your binaries accessible - redirections are not allowed. So there is a mechanism which takes care to get each and every source file included in the binary build artefacts, and this means sources of all MinGW libraries.

Could we keep the sources on some inria server and avoid repacking them at every job? Especially since it's a likely cause of the failure above.

SkySkimmer:
> https://gitlab.com/SkySkimmer/coq/pipelines/21065247 currently stalled for lack of runner, I'll kill it at some point if it doesn't die by itself.

It died on its own.

MSoegtropIMC:
> Could we keep the sources on some inria server and avoid repacking them at every job? Especially since it's a likely cause of the failure above.

Well it is part of the process of building an installer as much as building the installer itself. The files change frequently (about once a week), so if one separates this from building the installer, it might be hard to track the relation. In the end there is likely more reason to regression test this than to regression test the build of the installer itself.

But what would be great is to have a way to keep copies of all downloaded artifacts on some INRIA server and use this as backup source. I planned to have this on a private Amazon S3 bucket.

maximedenes:
> In general I have my doubts that automating signing is worth the effort.

Well, I've been signing packages for a few years now, and honestly I would prefer to spend this time on development.

> If I understand this right, the secrets are only available on special INRIA owned machines and running CI on these machines requires special privileges. Can you describe what these privileges are and how you ensure that only INRIA folks allowed to sign binaries in the name of INRIA obtain such privileges?

The runners that have the secret are used only on the protected branches of the main gitlab Coq repo. So folks who get the privilege to sign are people who have push access to that gitlab repo, or the GitHub repo (because of the mirroring).

.gitlab-ci.yml
maximedenes: I also thought I could do a later `package` stage. Especially because it is going to be required on macOS anyway, I think. Does it sound good?

maximedenes:
> Also I set the WIN_CERTIFICATE_PASSWORD variable to protected on gitlab/coq in case a wild printenv appears.

Sure, I had done it but disabled just while this PR was being tested.

maximedenes:
> Coq's make install complains about not having native compute stuff to install
> mingw stuff is downloaded again after bignums is finished (why?)

This is interesting but should be fixed separately.

.gitlab-ci.yml
SkySkimmer: >I also thought I could do a later package stage. Especially because it is going to be required on macOS anyway, I think. Does it sound good?

It's less parallel, what's the benefit? Also if any of the tests break it will block the package stage.
Why would it be required on apple?

Zimmi48:
> The runners that have the secret are used only on the protected branches of the main gitlab Coq repo.

How do you ensure this? It is because we intend to push branches corresponding to PRs on another GitLab repo, right? We should document this (maybe in this code).

maximedenes:
> How do you ensure this? It is because we intend to push branches corresponding to PRs on another GitLab repo, right? We should document this (maybe in this code).

Even if we push those branches, they wouldn't be protected, would they?

MSoegtropIMC:
> The runners that have the secret are used only on the protected branches of the main gitlab Coq repo. So folks who get the privilege to sign are people who have push access to that gitlab repo, or the GitHub repo (because of the mirroring).

I see. Well you have to decide if this is in accordance with your policies of handling signature keys, including the policies of all institutions involved in issuing the key. In my experience it is already a lot of work to find out what the policies are you need to comply to. Typically if you have a key which is signed by an authority trusted by default by Windows, you have signed something that you take reasonable care that nobody gets access to the key, who is not part of the legal entity to which the key has been issued. If you are lucky, not complying to this policy just leads to revocation of the certificate by the issuer (when they find out). If you are unlucky, there are substantial fines for negligent key leakage. If you considered this already, fine. If not, please check the trust chain of your key and the policies of the parties involved.

sorry for making your life harder!

SkySkimmer:
Sorry it's `only:` https://docs.gitlab.com/ee/ci/yaml/#only-and-except-complex it seems I keep confusing gitlab and travis today.

maximedenes:
> I see. Well you have to decide if this is in accordance with your policies of handling signature keys, including the policies of all institutions involved in issuing the key. In my experience it is already a lot of work to find out what the policies are you need to comply to. Typically if you have a key which is signed by an authority trusted by default by Windows, you have signed something that you take reasonable care that nobody gets access to the key, who is not part of the legal entity to which the key has been issued. If you are lucky, not complying to this policy just leads to revocation of the certificate by the issuer (when they find out). If you are unlucky, there are substantial fines for negligent key leakage. If you considered this already, fine. If not, please check the trust chain of your key and the policies of the parties involved.
> 
> sorry for making your life harder!

I understand your point of view, and I think you are right™. However, I'm afraid we have to be pragmatic: Inria does not have a team responsible for prompt signing, and won't let Coq developers access the key. So we got a key from the Inria foundation. As of today, I'm the only Coq developer part of that entity. My boss (who signed the forms for the key) is ok with they certificate being used as part of an automated process (it is already in use for other software). I don't want to have to go through a manual process relying on two VMs each time we do a release. So either we do things in a way that is not fully right, or we simply don't sign our binaries (I'm also fine with that, but IMHO it would be a regression for our users).

maximedenes:
> negligent key leakage

I hope we can make the argument that mandating core developers to produce signed binaries is not the same as leaking the key in an uncontrolled way.

Zimmi48:
> My boss (who signed the forms for the key) is ok with they certificate being used as part of an automated process (it is already in use for other software).

That's a key point.

MSoegtropIMC:
> I hope we can make the argument that mandating core developers to produce signed binaries is not the same as leaking the key in an uncontrolled way.

I would say this depends on to whom and by whom the key is issued. If the key is issued to say INRIA itself, giving anybody outside of INRIA the possibility to sign using this key could be called negligent key leakage. If the key is issued to say "The Coq Foundation" and not signed with a key owned by INRIA but by an independent authority, it should be fine that everyone who has write rights to the Coq repo can use this key for signing. In case the key is issued to "The Coq Foundation" but signed by a key issued to INRIA, I would say it is a corner case. The key INRIA has might not allow INRIA to issue keys for use outside of their organization - one would have to read the fine print.

But of cause I cannot give you legal counsel. All I say is that matters might be more complicated than you think. At least you should review to whom *exactly* and by whom the key is issued and read the fine print of the contract you signed to get the key. If you have no doubts after this, fine. Otherwise I would discuss this once more with your boss / your legal department / ... or possibly get a key which matches the requirements.

The whole trust system of Windows relies on the idea that if someone messes with your system, at least you can find out whom to sue. If you install software signed by people where it is pretty unknown who they are, it is your own problem. But in my opinion people have a right to reasonably trust software signed by large organizations, and the policies of trusted authorities typically reflect this. 

Once this mechanism is installed and working it is not very likely to be reviewed again. So better get it right.

maximedenes:
In fact, there is an easy solution:
- separate building and signing
- move the runner with secret to my own fork
- make the signing stage require a manual action (Gitlab supports it)

That way, I'll be the one manually producing the signed packages (and having access to the key), but it will require only clicking a button. It should address @MSoegtropIMC 's legal concerns.

MSoegtropIMC:
Indeed this is likely much less work than going through the fine print of the contract.

A good solution!

maximedenes:
I addressed @SkySkimmer 's remarks. The current version no longer relies on secrets for the main repo.

.gitlab-ci.yml
```diff
@@ -229,6 +229,26 @@ build:bleeding-edge:
...
+      - dev\nsis\*.exe
+      - coq-opensource-archive-windows-*.zip
+    expire_in: 1 week
+  dependencies: []
+  tags:
+    - windows
+  before_script:
+    - SET ARCH=64
+  script:
+    - powershell -Command "(New-Object Net.WebClient).DownloadFile('http://www.cygwin.com/setup-x86_64.exe', 'setup-x86_64.exe')"
```
Zimmi48: IIRC the download and installation of dependencies take a long time in the whole packaging process. In the future, we could end up saving a lot of time by using a personalized Docker image, especially if it is cached on the runners. (Not a request for this PR.)

.gitlab-ci.yml
```diff
@@ -229,6 +229,26 @@ build:bleeding-edge:
...
+    name: "$CI_JOB_NAME"
+    paths:
+      - dev\nsis\*.exe
+      - coq-opensource-archive-windows-*.zip
+    expire_in: 1 week
+  dependencies: []
+  tags:
+    - windows
+  before_script:
+    - SET ARCH=64
```
Zimmi48: It is for testing only that you only put "64" and you plan to also add "32" later?

dev/ci/gitlab.bat
```diff
@@ -0,0 +1,49 @@
...
+call %CI_PROJECT_DIR%\dev\build\windows\MakeCoq_MinGW.bat -threads=1 ^
+  -arch=%ARCH% -installer=Y -coqver=%CI_PROJECT_DIR_CFMT% ^
+  -destcyg=%CYGROOT% -destcoq=%DESTCOQ% -cygcache=%CYGCACHE% ^
+  -addon=bignums -make=N ^
+  -setup %CI_PROJECT_DIR%\%SETUP% || GOTO ErrorExit
+
+copy "%CYGROOT%\build\coq-local\dev\nsis\*.exe" dev\nsis || GOTO ErrorExit
+7z a coq-opensource-archive-windows-%ARCHLONG%.zip %CYGROOT%\build\tarballs\* || GOTO ErrorExit
+
+REM DO NOT echo the signing command below, as this would leak secrets in the logs
```
Zimmi48: You could also make the logs of your GitLab fork private, or have the secrets on a third, private repo, say "gitlab.com/inria-foundation/coq" or "gitlab.com/coq-consortium/coq".

Zimmi48 (commented):
I suppose you plan to strip off `dev/ci/appveyor.bat` and `appveyor.yml` later on.

.gitlab-ci.yml
```diff
@@ -229,6 +229,26 @@ build:bleeding-edge:
...
+  script:
+    - powershell -Command "(New-Object Net.WebClient).DownloadFile('http://www.cygwin.com/setup-x86_64.exe', 'setup-x86_64.exe')"
+    - call dev/ci/gitlab.bat
+  only:
+    variables:
+      - $WINDOWS == "enabled"
```
SkySkimmer: It seems the job wasn't created, did you set the variable after pushing or is there a problem?
(also the build job failed and you need to rebase)

maximedenes: Yeah I set it after. I should probably relaunch it.

.gitlab-ci.yml
maximedenes: How does Docker work on Windows?

.gitlab-ci.yml
maximedenes: Yeah, I forgot to create a second job. Will fix.

SkySkimmer:
You need to rebase so that the linux jobs pass (they block the test stage)

.gitlab-ci.yml
```diff
@@ -229,6 +247,18 @@ build:bleeding-edge:
     EXTRA_CONF: "-native-compiler yes -coqide opt"
     EXTRA_PACKAGES: "$COQIDE_PACKAGES"
 
+build:windows64:
```
SkySkimmer: I would just call the jobs `windowsxx`

MSoegtropIMC:
One more note: the creation of the open source package is only required for official downloadable installers. So it would make sense to do this only when signing. The problem is that Cygwin changes frequently and it could happen that the packages changed between build and signing - it can even happen that the source packages used for building don't exist any more 2 days later. Cygwin keeps the latest 2 versions and if there are a few updates in 2 days, there will be issues.

This would be solved by using a fixed Cygwin from a repo we control for each version, as I suggested before.

maximedenes:
> One more note: the creation of the open source package is only required for official downloadable installers. So it would make sense to do this only when signing.

Good point, I will make it configurable.

> The problem is that Cygwin changes frequently and it could happen that the packages changed between build and signing

For now, I will do the signing step rebuild the binary. Not a big deal, as long as everything is automated. That will enforce that the opensource archive is in sync with the released binary.

.gitlab-ci.yml
maximedenes: Done, and rebased.

.gitlab-ci.yml
```diff
@@ -229,6 +247,18 @@ build:bleeding-edge:
     EXTRA_CONF: "-native-compiler yes -coqide opt"
     EXTRA_PACKAGES: "$COQIDE_PACKAGES"
 
+windows64:
+  <<: *windows-template
+  variables:
+    ARCH: "64"
+    SETUP: "setup_x86-64.exe"
```
SkySkimmer: SETUP is reset in the gitlab.bat so not needed here.

maximedenes: It's now used in `.gitlab-ci.yml`. I should probably remove it from `gitlab.bat`.

SkySkimmer: Why not include the download in the .bat?

maximedenes: How?

.gitlab-ci.yml
Zimmi48: No experience with it but it seems to be supported.

.gitlab-ci.yml
maximedenes: No, the part that performs the download is PowerShell.

SkySkimmer:
Is it OK to keep the open source files only as long as we keep the binaries (1 week currently)?

.gitlab-ci.yml
SkySkimmer: Can powershell not be called from bat?

maximedenes: I have no idea, I know nothing about Windows dev.

maximedenes: But anyway, this download is setting up dependencies, so I'm not shocked by seeing it in `.gitlab-ci.yml`. Isn't it what we do for other builds?

Zimmi48:
Yes, because there are various ways to satisfy the requirements of the GPL but one of them is to propose to download everything at once from the same place.

Zimmi48:
Concretely this means that if a user downloads the binary but not the sources and after one week wants to distribute the binary further, they could get in trouble (not us). If they downloaded both, then no problem. This is all very theoretical. No one is ever going to sue for this.

MSoegtropIMC:
@Zimmi48 : The open source archive is there to fulfill legal requirements - it has not real practical use. I would think that one fulfills the word and spirit of the GPL, if one makes the open source archive only available for official releases and not for intermediate test builds, at least if these official releases are regular. If this is questionable, I could go and ask the GNU lawyers.

Again all this would be solved by making sure that the contents of the open source archive wouldn't change over a release. I suggested to download all this from a Coq specific Repo, which is updated shortly after each release to the latest version. The open source archive can then be created when this repo changes. It doesn't make that much sense to recreate it a few 1000 times on each CI run.

SkySkimmer:
>Exception calling "DownloadFile" with "2" argument(s): "An exception occurred d
uring a WebClient request."
At line:1 char:40
>+ (New-Object Net.WebClient).DownloadFile <<<< ('http://www.cygwin.com/$SETUP',
 $SETUP)
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : DotNetMethodException

MSoegtropIMC:
P.S.: the main benefit of not changing Cygwin any time in the release cycle is that a change from cygwin can't break our CI. Solving the open source archive issue is just a side benefit.

Zimmi48:
> official releases and not for intermediate test builds

As soon as these test builds are available for general download, I'm afraid the requirements kick in.

> Again all this would be solved by making sure that the contents of the open source archive wouldn't change over a release. I suggested to download all this from a Coq specific Repo, which is updated shortly after each release to the latest version. The open source archive can then be created when this repo changes. It doesn't make that much sense to recreate it a few 1000 times on each CI run.

I fully agree with this view and solution, this is also what I had in mind above when I suggested to use a homemade docker image. Unfortunately, I have very little experience with Docker and zero experience with Docker for Windows, so I don't know how doable this proposal here. But even if not done with Docker but rather in a ad-hoc way, I would still support this solution.

SkySkimmer:
@maximedenes while trying to get it to work you may want to run windows in the first stage, so that you avoid wasting time waiting for linux jobs
Maybe even make a new stage before all the others, or delete the other jobs.

maximedenes:
> @maximedenes while trying to get it to work you may want to run windows in the first stage, so that you avoid wasting time waiting for linux jobs
> Maybe even make a new stage before all the others, or delete the other jobs.

That's ok, I'm multitasking.

MSoegtropIMC:
@Zimm48: Using docker with locaI repo is an interesting approach. I have some experience with docker, so please feel free to create a ticket for this and assign it to me.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -959,6 +959,10 @@ function make_lablgtk {
     "$TARGET_ARCH-strip.exe" --strip-unneeded src/dlllablgtk2.dll
 
     log2 make world
+
+    # lablgtk does not escape FINDLIBDIR path, which can contain backslashes
+    sed -i "s|^FINDLIBDIR=.*|FINDLIBDIR=$PREFIXOCAML/libocaml/site-lib|" config.make
+
```
MSoegtropIMC: Can you explain under which circumstances FINDLIBDIR contains backslashes? 
FINDLIBDIR should be in what I call MFMT.
See the docs in MakeCoq_MinGW.bat:

REM WFMT = windows format (C:\..)          Used in this batch file.
REM CFMT = cygwin format (\cygdrive\c\..)  Used for Cygwin PATH varible, which is : separated, so C: doesn't work.
REM MFMT = MinGW format (C:/...)           Used for the build, because \\ requires escaping. Mingw can handle \ and /.

SkySkimmer:
still broken

.gitlab-ci.yml
```diff
@@ -173,6 +174,24 @@ before_script:
     TEST_TARGET: "$CI_JOB_NAME"
     EXTRA_PACKAGES: "$TIMING_PACKAGES"
 
+.windows-template: &windows-template
+  stage: windows
+  artifacts:
+    name: "$CI_JOB_NAME"
```
SkySkimmer: Needs to be $env:foo or %foo% (not sure which).

SkySkimmer:
Fixed the setup download.
It seems we have only the one windows runner, with 2 jobs and 1h15 per job that's going to be quite slow.
The artifact timeout from before scares me too.

dev/build/windows/makecoq_mingw.sh
maximedenes: This path is built by lablgtk calling `ocamlfind`, so we don't have control over the format.

SkySkimmer:
Artifact name looks fixed.
Please move the jobs back to the `test` stage. Also squash.
Do we want both jobs considering how long they take?

maximedenes:
I'll move the job, squash and register 2 runners.

Zimmi48:
Please remove the packaging jobs from AppVeyor at the same time. Given how slow AppVeyor is to build pull requests compared to the rest of our CI, we could use the speedup.

maximedenes:
Squashed, rebased, now with 2 workers, moved to test stage, removed packaging jobs from AppVeyor. Please merge if CI is green.

ejgallego:
Can you update naming to `windows:32bit` and `windows:64bit`?

[we cannot use `ci:foo` for Make reasons tho]

maximedenes:
`32bit` or `32bits`?

ejgallego:
We are not yet fully settled on a consiste naming scheme, opam uses the `32bit` prefix so that's what we have used for now.

maximedenes:
I think this is going to make the build failed because it will produce an invalid file name under windows (because of colon). So I'd rather keep things as they are.

Zimmi48:
In the commit message, you say "a one click operation" on your repo, but it is not obvious to me what the setup for this one click operation is. However, it is rather clear why it will sign only on your repo only: the secrets will be there and not elsewhere.

The removal of the packaging jobs from AppVeyor could have implied a clean-up of the AppVeyor batch script too. But this clean-up can also be done later.

When this is merged, it should probably get a bit of documentation, probably as part of #6057.

Zimmi48:
Did you make sure that the build was also working on your repo when the signing secrets are there?

SkySkimmer:
The runner failed:
>'"git"' is not recognized as an internal or external command,
operable program or batch file.

maximedenes:
Working on it.

maximedenes:
> Did you make sure that the build was also working on your repo when the signing secrets are there?

This PR was tested with and without the secrets, though not on my repo. I plan to setup the "one click" operation as a deployment environment on my repo, making another PR here if needed. This PR, in its current state, does not introduce any regression on that matter, since signing was already done outside of CI.

ejgallego:
> I think this is going to make the build failed because it will produce an invalid file name under windows (because of colon). So I'd rather keep things as they are.

I would recommend not depending on the `CI JOB NAME` as to name the artifacts, but if you want to do that for now that's OK.

maximedenes:
@SkySkimmer I believe this is finally ready.

ejgallego:
It seems that windows runners can only run one concurrent job, pity as this makes scaling a bit more painful.

Anyways, we should really increase the number of windows runners as the current capacity is clearly not enough.

PR #7347, opened 2018-04-25
Fix for #7081 (Windows lablgtk) and #7083 (Windows logging)

### LINKED ISSUE #7081: Windows build: strip in lablgtk build can fail randomly
#### Version

All version since 8.5

#### Operating system

Windows

In the windows build symbols are stripped from the lablgtk dll, because otherwise flexdll goes wild. 

$TARGET_ARCH-strip.exe --strip-unneeded src/dlllablgtk2.dll

The dll is locked in Windows for a short time after creation and if a strip is done during this time, the strip fails. With some logging changes this leads to reproducible errors, otherwise this exits only rarely. Fix is to strip during link phase right away with

ocamlmklib -ldopt "-link -Wl,-s"

-ldopt is the okamlmklib linker prefix option
-link is the flexlink linker prefix option
-Wl, is the gcc (linker driver) linker prefix option
-s is the gnu linker option for stripping symbols

Will do a pull request as soon as the script clean up is merged.

### LINKED ISSUE #7083: Windows build: Unify build logging to console (for appveyor) and files
Logging is handled a bit randomly currently. In general logging is done to files, but this is not so useful in appveyor. The idea is to log to both a file and the console, also prepending the build step when logging to the console (which is the file name when logging to files).
This is currently blocked by #7081 - if the logging is changed, the lablgtk build fails reproducibly (found by @JasonGross a while back).

MSoegtropIMC:
Fixes #7081 
Fixes #7083

### DIFF
--- a/dev/build/windows/patches_coq/lablgtk-2.18.3.patch
+++ b/dev/build/windows/patches_coq/lablgtk-2.18.3.patch
@@ -1,6 +1,12 @@
-diff -u -r lablgtk-2.18.3/configure lablgtk-2.18.3.patched/configure
---- lablgtk-2.18.3/configure	2014-10-29 08:51:05.000000000 +0100
-+++ lablgtk-2.18.3.patched/configure	2015-10-29 08:58:08.543985500 +0100
+diff/patch file created on Wed, Apr 25, 2018 11:08:05 AM with:
+difftar-folder.sh ../coq-msoegtrop/dev/build/windows/source_cache/lablgtk-2.18.3.tar.gz lablgtk-2.18.3 1
+TARFILE= ../coq-msoegtrop/dev/build/windows/source_cache/lablgtk-2.18.3.tar.gz
+FOLDER= lablgtk-2.18.3
+TARSTRIP= 1
+TARPREFIX= lablgtk-2.18.3/
+ORIGFOLDER= lablgtk-2.18.3.orig
+--- lablgtk-2.18.3.orig/configure	2014-10-29 08:51:05.000000000 +0100
++++ lablgtk-2.18.3/configure	2018-04-25 10:58:54.454501600 +0200
 @@ -2667,7 +2667,7 @@
  fi
  
@@ -10,10 +16,8 @@ diff -u -r lablgtk-2.18.3/configure lablgtk-2.18.3.patched/configure
    { $as_echo "$as_me:${as_lineno-$LINENO}: WARNING: Ignoring ocamlfind" >&5
  $as_echo "$as_me: WARNING: Ignoring ocamlfind" >&2;}
    OCAMLFIND=no
-
-diff -u -r lablgtk-2.18.3/src/glib.mli lablgtk-2.18.3.patched/src/glib.mli
---- lablgtk-2.18.3/src/glib.mli	2014-10-29 08:51:06.000000000 +0100
-+++ lablgtk-2.18.3.patched/src/glib.mli	2016-01-25 09:50:59.884715200 +0100
+--- lablgtk-2.18.3.orig/src/glib.mli	2014-10-29 08:51:06.000000000 +0100
++++ lablgtk-2.18.3/src/glib.mli	2018-04-25 10:58:54.493555500 +0200
 @@ -75,6 +75,7 @@
    type condition = [ `ERR | `HUP | `IN | `NVAL | `OUT | `PRI]
    type id
@@ -22,10 +26,8 @@ diff -u -r lablgtk-2.18.3/src/glib.mli lablgtk-2.18.3.patched/src/glib.mli
    val add_watch :
      cond:condition list -> callback:(condition list -> bool) -> ?prio:int -> channel -> id
    val remove : id -> unit
-
-diff -u -r lablgtk-2.18.3/src/glib.ml lablgtk-2.18.3.patched/src/glib.ml
---- lablgtk-2.18.3/src/glib.ml	2014-10-29 08:51:06.000000000 +0100
-+++ lablgtk-2.18.3.patched/src/glib.ml	2016-01-25 09:50:59.891715900 +0100
+--- lablgtk-2.18.3.orig/src/glib.ml	2014-10-29 08:51:06.000000000 +0100
++++ lablgtk-2.18.3/src/glib.ml	2018-04-25 10:58:54.479543500 +0200
 @@ -72,6 +72,8 @@
    type id
    external channel_of_descr : Unix.file_descr -> channel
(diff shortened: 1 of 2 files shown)


dev/build/windows/patches_coq/lablgtk-2.18.3.patch
```diff
@@ -10,10 +16,8 @@ diff -u -r lablgtk-2.18.3/configure lablgtk-2.18.3.patched/configure
    { $as_echo "$as_me:${as_lineno-$LINENO}: WARNING: Ignoring ocamlfind" >&5
  $as_echo "$as_me: WARNING: Ignoring ocamlfind" >&2;}
    OCAMLFIND=no
-
-diff -u -r lablgtk-2.18.3/src/glib.mli lablgtk-2.18.3.patched/src/glib.mli
```
SkySkimmer: Is removing the diff -u lines intentional?

MSoegtropIMC: More or less yes. I use a script to compare tar files with folders now to create patches to tar files (difftar-folder.sh in /dev/build/windows folder). This script doesn't produce these lines cause of the way diff is called, but of cause I might be able to change the script. I created this script, because it was a bit unreliable/inconvenient to create patches to tar files in other ways.

The information is anyway redundant. The two change lines below give the complete information on the patch (including if it is unified (-u) format or not).

I added an additional header (which is ignored by patch) which details how the script was called.

MSoegtropIMC:
I fixed the Travis error (two lines with trailing white space - sorry I had only appveyor enanbled in my fork, will change this). Appveyor went through on my fork, see:

https://ci.appveyor.com/project/MSoegtropIMC/coq/build/issue-7083-logging~47

so you might want to integrate it right away.

Btw.: I can also merge this, if someone tells me to do so. The rules are that one must not merge his own PRs, but I guess for such trivial changes, it would be ok to do so if someone else with write right agrees.

MSoegtropIMC:
@Zimmi48 : Sorry, I don't understand this. The errors shown by the latest travis run are clearly fixed by the commit c2b2105 I did this morning. Do you have an explanation on what is going on?

Zimmi48:
Ah I see! Actually you need to fix the problematic commit (using `git rebase -i` to squash 263ee1ca39ae0a6ea823d8984f2f135f942afbc3 and c2b2105). You can't just push an additional commit fixing a previous one in the same PR.

SkySkimmer:
The incremental whitespace lint is per-commit otherwise diff heuristics can make errors appear from nowhere.

MSoegtropIMC:
@Zimmi48 : I did the rebase as you suggested, but I don't think this is really what I want. Now I have all commits done inbetween in this PR, need all the involved's peoples review and everything.

In general I avoid rebases, since they change the history. And I have the feeling that this PR is now pretty messed up.

ejgallego:
@MSoegtropIMC the rebase didn't certainly go fine, be sure you used the proper branch.

MSoegtropIMC:
@Zimmi48 : well a rebase of my own branch would be a nop.
What exactly should I have rebased?

ejgallego:
At worst you can `reset --hard` to master and use `cherry-pick`

MSoegtropIMC:
Yes. The question was, what rebase did you suggest, so that Travis would pass?

MSoegtropIMC:
(Without messing up the PR ...)

SkySkimmer:
The merge commit messed you up I think.
Resetting to 45bf9c60fe should work.

Zimmi48:
When rebasing to squash commits together, you can just do something like `git rebase -i HEAD~3` and then change a `pick` into `f` and moving lines around.

ejgallego:
I think `git rebase origin/master` is usually what you want, where `origin/master` points to a recent Coq `master`.

As for the whitespace rebase, you want to be very careful, I always use `git rebase --whitespace=fix HEAD~N` where `N` is the amount of local commits.

MSoegtropIMC:
I see, so in the end you suggested to squash my 3 commits.

Honestly I don't know if I like this. The two first commits are separate for a reason (they are for different issues). Also I think squashing commits after creating a PR makes the life of reviewers much harder.

Is there no better way to make travis happy?

SkySkimmer:
I don't know why @Zimmi48 is talking about squashing, the issue was whitespace.
This should work:
```bash
git reset --hard 45bf9c60fe
git rebase whitespace=fix HEAD~3
```

Zimmi48:
No I didn't suggested to squash the three commits together, only the one with the whitespace fix with the commit which had the whitespace problem. Anyways, do as @SkySkimmer said, it should work: and you should get only two commits in the end. Although this:

```
git reset --hard c07df4b531cfc01b94d657aa486eebda5315a561
git rebase whitespace=fix HEAD~2
```
should work as well and be even simpler.

MSoegtropIMC:
@Zimmi48 , @SkySkimmer : thanks, the branch is fixed now. I used

git reset --hard c07df4b531cfc01b94d657aa486eebda5315a561
git rebase --whitespace=fix HEAD~2

The pull request also seems to look ok (except maybe the reviwer list is longer, but still only one reviewer is required.

SkySkimmer:
I invoked mystical powers and cleaned up the reviewer list.

MSoegtropIMC:
Does someone know what the Sphinx / Opam failure on gitlab is (I couldn't find an issue report about this)

/usr/lib/python3/dist-packages/sphinx/environment/__init__.py:704: RemovedInSphinx17Warning: html_use_smartypants option is deprecated. Smart quotes are on by default; if you want to disable them, use the smartquotes option.
  RemovedInSphinx17Warning)

Exception occurred:
  File "/usr/lib/python3.6/subprocess.py", line 410, in run
    stderr=stderr)
subprocess.TimeoutExpired: Command '['/builds/coq/coq/bin/coqdoc', '--body-only', '--no-glob', '--no-index', '--no-externals', '-s', '--html', '--stdout', '--utf8', '/tmp/coqdoc-04grhe6_.v']' timed out after 2 seconds

Zimmi48:
@MSoegtropIMC I think it is related to 1b462cfd2b03fa90e30b6d53de3adb33bc756eaf (which was attempting to fix it) and that it is an error that still occurs spuriously from time to time. I'm not sure there is an opened issue report about it. In any case, this GitLab CI status is still experimental right now so it is safe to ignore. Given the scope of your PR, the only real important CI result will be AppVeyor (unfortunately it is also the one that takes the longest).

maximedenes:
@Zimmi48 I'm not sure it is exactly the same thing, because this timeout seems to be when calling `coqdoc`, not `coqtop`.

Zimmi48:
OK, unrelated to this PR in any case. AppVeyor passed BTW.

MSoegtropIMC:
@maximedenes : so just your review is missing. You can have a look at the new log format here:

https://ci.appveyor.com/api/buildjobs/9ab35mv5t0g50mh4/log

Enabling something like this failed before due to the lablgtk build issue, so these two fixes are combined in this PR.

A few features of the new logging:
- everything which goes into a log file is also output to stdout/stderr
- the logfile name prepended to each line
- Everything which went to stderr can be searched for by searching for .err: since all error log files have the extension .err

In case you want me to merge this PR (despite the rule that people shouldn't merge their own PRs) just let me know.

maximedenes (approved):
I'm not qualified to say if the flexdll fix is the right one, but this is obviously a desirable change, and CI is green, so let's merge :)

Sorry for the delay.

MSoegtropIMC:
Interestingly, just yesterday a new fix for the flexdll issue was proposed:

https://github.com/alainfrisch/flexdll/issues/6#issuecomment-386876135

Another fix would be not build a dll but a static library and link flexdll statically - not sure if this works in all circumstances.

But I would say the current solution is quite ok, definitely an order of magnitude more elegant and hopefully many orders of magnitude more stable. FYI: the old version first let make fail, then strip flexdll and then rerun make.

Thanks for merging!

PR #7349, opened 2018-04-25
Fix an anomaly when calling Obligation 0 or Obligation -1.

Zimmi48:
Fix an anomaly when calling `Obligation 0` or `Obligation -1`.

I didn't change the parser which expects an integer instead of a positive number
because changing it would also mean having worse error messages because of our
current LL parser. The error message would have been:

```
Syntax error: 'Tactic' ':=' expected after 'Obligation' (in
    [vernac:command]).
```

### DIFF
--- a/vernac/obligations.ml
+++ b/vernac/obligations.ml
@@ -958,7 +958,7 @@ and obligation (user_num, name, typ) tac =
   let num = pred user_num in
   let prg = get_prog_err name in
   let obls, rem = prg.prg_obligations in
-    if num < Array.length obls then
+    if num >= 0 && num < Array.length obls then
       let obl = obls.(num) in
 	match obl.obl_body with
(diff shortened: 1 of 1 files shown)


Zimmi48:
@JasonGross https://circleci.com/gh/coq/coq/34218 shows an interesting (but also pretty rare, I think) error of the timing log.

JasonGross:
@ejgallego @Zimmi48 this one is interesting ( and different from the usual one, because it's an exception in the python script, not an error in the test-suite).  My first guess would be that the timing log is interleaved incorrectly; are we using `--output-sync` or whatever?  I'll dig into the log momentarily...

Zimmi48:
@mattam82 Ping

mattam82 (approved):
Good catch

PR #7351, opened 2018-04-25
Always print explanation for univ inconsistency, rm Flags.univ_print

SkySkimmer:
This removes the Flags.univ_print in the kernel, making it possible to
put the univ printing flag ownership back in Detyping.

The lazyness is because getting an explanation may be costly and we
may discard it without printing.

See benches
- with lazy
  https://ci.inria.fr/coq/view/benchmarking/job/benchmark-part-of-the-branch/406/console
- without lazy
  https://ci.inria.fr/coq/view/benchmarking/job/benchmark-part-of-the-branch/405/console

Notably without lazy mathcomp odd_order is +1.26% with some lines
showing significant changes, eg PFsection11 line 874 goes from 11.76s
to 13.23s (+12%).
(with lazy the same development has -1% overall and the same line goes
from 11.76s to 11.23s (-4%) which may be within noise range)

### DIFF
--- a/dev/top_printers.ml
+++ b/dev/top_printers.ml
@@ -162,8 +162,8 @@ let pp_state_t n = pp (Reductionops.pr_state n)
 (* proof printers *)
 let pr_evar ev = Pp.int (Evar.repr ev)
 let ppmetas metas = pp(Termops.pr_metaset metas)
-let ppevm evd = pp(Termops.pr_evar_map ~with_univs:!Flags.univ_print (Some 2) evd)
-let ppevmall evd = pp(Termops.pr_evar_map ~with_univs:!Flags.univ_print None evd)
+let ppevm evd = pp(Termops.pr_evar_map ~with_univs:!Detyping.print_universes (Some 2) evd)
+let ppevmall evd = pp(Termops.pr_evar_map ~with_univs:!Detyping.print_universes None evd)
 let pr_existentialset evars =
   prlist_with_sep spc pr_evar (Evar.Set.elements evars)
 let ppexistentialset evars =
--- a/kernel/uGraph.ml
+++ b/kernel/uGraph.ml
@@ -21,7 +21,7 @@ open Univ
 (* Revisions by Bruno Barras, Hugo Herbelin, Pierre Letouzey, Matthieu
    Sozeau, Pierre-Marie Pédrot, Jacques-Henri Jourdan *)
 
-let error_inconsistency o u v (p:explanation option) =
+let error_inconsistency o u v p =
   raise (UniverseInconsistency (o,Universe.make u,Universe.make v,p))
 
 (* Universes are stratified by a partial ordering $\le$.
@@ -557,8 +557,7 @@ let get_explanation strict u v g =
   else match traverse strict u with Some exp -> exp | None -> assert false
 
 let get_explanation strict u v g =
-  if !Flags.univ_print then Some (get_explanation strict u v g)
-  else None
+  Some (lazy (get_explanation strict u v g))
 
 (* To compare two nodes, we simply do a forward search.
    We implement two improvements:
--- a/kernel/univ.ml
+++ b/kernel/univ.ml
@@ -541,11 +541,11 @@ let constraint_type_ord c1 c2 = match c1, c2 with
 (* Universe inconsistency: error raised when trying to enforce a relation
    that would create a cycle in the graph of universes. *)
 
-type univ_inconsistency = constraint_type * universe * universe * explanation option
+type univ_inconsistency = constraint_type * universe * universe * explanation Lazy.t option
 
 exception UniverseInconsistency of univ_inconsistency
 
-let error_inconsistency o u v (p:explanation option) =
+let error_inconsistency o u v p =
   raise (UniverseInconsistency (o,make u,make v,p))
 
 (* Constraints and sets of constraints. *)    
@@ -1235,13 +1235,16 @@ let explain_universe_inconsistency prl (o,u,v,p) =
     | Eq -> str"=" | Lt -> str"<" | Le -> str"<=" 
   in
   let reason = match p with
-    | None | Some [] -> mt()
+    | None -> mt()
     | Some p ->
-      str " because" ++ spc() ++ pr_uni v ++
+      let p = Lazy.force p in
+      if p = [] then mt ()
+      else
+        str " because" ++ spc() ++ pr_uni v ++
 	prlist (fun (r,v) -> spc() ++ pr_rel r ++ str" " ++ pr_uni v)
-	p ++
+          p ++
 	(if Universe.equal (snd (List.last p)) u then mt() else
-	    (spc() ++ str "= " ++ pr_uni u)) 
+           (spc() ++ str "= " ++ pr_uni u))
   in
     str "Cannot enforce" ++ spc() ++ pr_uni u ++ spc() ++
       pr_rel o ++ spc() ++ pr_uni v ++ reason
--- a/kernel/univ.mli
+++ b/kernel/univ.mli
@@ -205,7 +205,7 @@ val enforce_leq_level : Level.t constraint_function
   Constraint.t...
 *)
 type explanation = (constraint_type * Universe.t) list
-type univ_inconsistency = constraint_type * Universe.t * Universe.t * explanation option
+type univ_inconsistency = constraint_type * Universe.t * Universe.t * explanation Lazy.t option
 
 exception UniverseInconsistency of univ_inconsistency
 
--- a/lib/flags.ml
+++ b/lib/flags.ml
@@ -60,7 +60,6 @@ let profile = false
 let ide_slave = ref false
 
 let raw_print = ref false
-let univ_print = ref false
 
 let we_are_parsing = ref false
 
--- a/lib/flags.mli
+++ b/lib/flags.mli
@@ -42,9 +42,6 @@ val we_are_parsing : bool ref
 (* Set Printing All flag. For some reason it is a global flag *)
 val raw_print : bool ref
 
-(* Univ print flag, never set anywere. Maybe should belong to Univ? *)
-val univ_print : bool ref
-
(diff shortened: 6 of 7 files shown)


ejgallego (approved):
Very good, much needed cleanup.

Indeed laziness is a must here as in many other parts, sadly we lack a unified pattern along the codebase.

Something weird is the conversion of UniverseInconsistency in univdecls to a regular user error, as opposed to the current himsg / explainErr organization; but well, it also seems like the exception printers should indeed be registered at each component point instead of there.

ejgallego:
> The lazyness is because getting an explanation may be costly and we
> may discard it without printing.

who's getting an explanation and not printing it?

ejgallego:
> who's getting an explanation and not printing it?

It is the tactic layer? If so, we should absolutely fix that.

SkySkimmer:
>who's getting an explanation and not printing it?

Any tactic like `foo || bar` where foo fails because of a universe inconsistency.
`Reductionops.infer_conv` when syntactic comparison fails because of an inconsistency.
etc

ejgallego:
Ah, somehow I would have expected that `univ_inconsistency` would be:
```ocaml
type univ_inconsistency = constraint_type * Universe.t * Universe.t * UGraph.t
```
and that `explain_universe_inconsistency` would actually call the "explainer"

ejgallego:
`UGraph.t` is a functional structure, right? If so, then I'd prefer to embed pointers in exceptions instead of elaborated terms [either under a closure or not]

SkySkimmer:
`type univ_inconsistency` is defined in the wrong module for that.

SkySkimmer:
>Indeed the errors look conceptually different to me

What do you mean?

>Also, what do you think about fixing the weird call in univdecls?

Some other PR can do that.

ejgallego:
I mean that the errors happen in different contexts/algorithms, so it could make sense to separate them.

So do you want this to go as in?

SkySkimmer:
What errors are you talking about?

ejgallego:
> What errors are you talking about?

I mean the error in `univ` vs the error `uGraph`.

SkySkimmer:
There's a short path for an error on `Prop = Set` and an error on `u+k+1 <= u` (k>=1).
The both seem like they could result in uncaught exceptions since the user may not expect `Constraint.t -> Constraint.t` functions to do constraint checking.
I don't know how the second can happen. The coverage is interesting btw https://x80.org/coq2017/report/file0247.html#L658 the anomaly happens a few times (!!) but not the special case.

ejgallego:
Sounds good, given that this seems to be impacting performance it is IMHO worth to try to allocate only the exception instead of the whole lazy closure.

SkySkimmer:
>Sounds good, given that this seems to be impacting performance it is IMHO worth to try to allocate only the exception instead of the whole lazy closure.

Bench says it does nothing: https://ci.inria.fr/coq/view/benchmarking/job/benchmark-part-of-the-branch/409/console (this is moving the exception to ugraph vs current state of this PR)
```

┌────────────────────────┬─────────────────────────┬─────────────────────────────────────┬─────────────────────────────────────┬─────────────────────────┬─────────────────┐
│                        │      user time [s]      │             CPU cycles              │          CPU instructions           │  max resident mem [KB]  │   mem faults    │
│                        │                         │                                     │                                     │                         │                 │
│           package_name │     NEW     OLD PDIFF   │           NEW           OLD PDIFF   │           NEW           OLD PDIFF   │     NEW     OLD PDIFF   │ NEW OLD PDIFF   │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│               coq-hott │  328.15  329.43 -0.39 % │  915995669296  917803819053 -0.20 % │ 1426060712100 1426415583343 -0.02 % │  588908  588812 +0.02 % │   1   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│ coq-mathcomp-character │  274.27  274.96 -0.25 % │  762796228831  764705721178 -0.25 % │ 1085215313521 1085390254533 -0.02 % │ 1126292 1125208 +0.10 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│       coq-fiat-parsers │  697.10  698.28 -0.17 % │ 1930882690460 1934350807917 -0.18 % │ 2958234783232 2957764791023 +0.02 % │ 3376660 3376096 +0.02 % │   0  67 -100.00 % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│            coq-unimath │ 1271.82 1272.44 -0.05 % │ 3521484943670 3527116566828 -0.16 % │ 5990951571402 5990834491445 +0.00 % │ 1092408 1092524 -0.01 % │   1   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│       coq-math-classes │  227.97  227.92 +0.02 % │  627478005655  626651329384 +0.13 % │  783903650577  784085652822 -0.02 % │  532428  532388 +0.01 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│           coq-compcert │  871.45  871.19 +0.03 % │ 2414665584528 2413909959715 +0.03 % │ 3473361931491 3473851436528 -0.01 % │ 1327636 1327612 +0.00 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│  coq-mathcomp-solvable │  201.01  200.90 +0.05 % │  557334410372  557341817854 -0.00 % │  768118844321  768365871947 -0.03 % │  852016  867180 -1.75 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│  coq-mathcomp-fingroup │   64.64   64.60 +0.06 % │  178176084896  178352037878 -0.10 % │  238034815850  238215265714 -0.08 % │  590732  589636 +0.19 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│         coq-coquelicot │   81.60   81.54 +0.07 % │  225073539866  225023633392 +0.02 % │  265491865307  265708561880 -0.08 % │  716380  716288 +0.01 % │  13   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│   coq-mathcomp-algebra │  185.12  184.98 +0.08 % │  513670198306  512721448553 +0.19 % │  691915760632  691930210493 -0.00 % │  643516  643184 +0.05 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│     coq-mathcomp-field │  459.70  459.25 +0.10 % │ 1278972946701 1277500196024 +0.12 % │ 2060554625051 2060904866519 -0.02 % │  814376  814492 -0.01 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│              coq-color │  724.29  723.10 +0.16 % │ 2001088007770 1998361909702 +0.14 % │ 2371419164885 2371857820162 -0.02 % │ 1446072 1448668 -0.18 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│          coq-fiat-core │  111.13  110.93 +0.18 % │  307925926399  308449489907 -0.17 % │  374661140557  375321651040 -0.18 % │  499352  499256 +0.02 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│              coq-flocq │  263.64  262.99 +0.25 % │  731816231657  731265126379 +0.08 % │  906599859736  907963166294 -0.15 % │ 1245740 1245672 +0.01 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│            coq-bignums │   80.26   79.86 +0.50 % │  220943357728  220360085592 +0.26 % │  285557447803  285879602387 -0.11 % │  528512  528564 -0.01 % │   0   0  +nan % │
├────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│ coq-mathcomp-ssreflect │   45.47   45.06 +0.91 % │  123187031845  123358879775 -0.14 % │  144776643336  144839487022 -0.04 % │  536744  536780 -0.01 % │   0   0  +nan % │
└────────────────────────┴─────────────────────────┴─────────────────────────────────────┴─────────────────────────────────────┴─────────────────────────┴─────────────────┘
```
and it breaks the test suite so let's not bother. Maybe we will do it some other PR.

ejgallego:
Ok.

ejgallego:
I am always sad to let opportunities for further cleanup slip

SkySkimmer:
I consider it more an experiment than a further cleanup.

ejgallego:
I was thinking of creating two exceptions, as they do seem conceptually different to me, removing the exception in the lower layer is indeed an experiment.

SkySkimmer:
How is `Prop = Set` different from any other inconsistent constraint?

ejgallego:
One is coming from checking the other from inference?

SkySkimmer:
No.

PR #7358, opened 2018-04-26
Fix #7356: missing lift when interpreting default instances of evars

### LINKED ISSUE #7356: Failure to interpret the instance of an existential variable
#### Version

Since 8.5.

```coq
Check fun (P : nat -> Prop) (x:nat) (h:P x) => exist _ ?[z] (h : P ?z).
```
This should work but fails with `Cannot interpret ?z in current context: no binding for h.`.

Fix to come (de Bruijn bug).

herbelin:
**Kind:** bug fix

Fixes / closes #7356.


gares (approved):

gares:
I renamed the PR, since I don't think it is nice to consider this a "De Bruijn bug" (even if he may have some responsibility in this class of bugs ;-)

PR #7361, opened 2018-04-26
Towards selecting "best" unification failure among several

herbelin:
**Kind:** enhancement

This is both a shy step towards better understanding the kind of unification failures we may encounter and a concrete improvement of some particular kind of unification error, namely when a first-order unification fails because of a typing error.

The concrete example solved is:
```coq
Check fun P (x:nat) (h:P x) => exist _ x (h : P x).
```
Formerly: 
```
Unable to unify "P x" with "?P x" (cannot satisfy constraint "?P x" == "P x").
```
With the PR:
```coq
Unable to unify "P x" with "?P x"
(unable to find a well-typed instantiation for "?P": cannot ensure that
"nat -> Type" is a subtype of "nat -> Prop").
```

In the longer term, whatever direction is taken for unification (rewritten on a new backtracking engine, integrated to the backtracking monad, or any other approach), I feel that we could be able to classify the different kind of branching in unification and to fine-tune what kind of failures to report (maybe several sometimes).

Also, about the specific example above, even if the PR improves the explanation of the failure, I think that it should succeed, in the sense that nothing says that `P` should necessarily be in a sort which excludes `Prop`. Anyone with ongoing projects towards making the example succeed?

### DIFF
--- a/pretyping/evarconv.ml
+++ b/pretyping/evarconv.ml
@@ -227,7 +227,10 @@ let check_conv_record env sigma (t1,sk1) (t2,sk2) =
 (* Precondition: one of the terms of the pb is an uninstantiated evar,
  * possibly applied to arguments. *)
 
-let join_failures evd1 evd2 e1 e2 = (evd2,e2)
+let join_failures evd1 evd2 e1 e2 =
+  match e1, e2 with
+  | _, CannotSolveConstraint (_,ProblemBeyondCapabilities) -> (evd1,e1)
+  | _ -> (evd2,e2)
 
 let rec ise_try evd = function
     [] -> assert false
--- a/pretyping/evarconv.ml
+++ b/pretyping/evarconv.ml
@@ -227,13 +227,20 @@ let check_conv_record env sigma (t1,sk1) (t2,sk2) =
 (* Precondition: one of the terms of the pb is an uninstantiated evar,
  * possibly applied to arguments. *)
 
+let join_failures evd1 evd2 e1 e2 = (evd2,e2)
+
 let rec ise_try evd = function
     [] -> assert false
   | [f] -> f evd
   | f1::l ->
       match f1 evd with
       | Success _ as x -> x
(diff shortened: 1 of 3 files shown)


gares:
I'm a bit lost. I see the error message is nicer but I don't get why the pretyper decided to put `?P` in `Type` in the first place. Can you explain that?

SkySkimmer:
>I'm a bit lost. I see the error message is nicer but I don't get why the pretyper decided to put ?P in Type in the first place. Can you explain that?

When typing `h:P x` the type of `P x` needs to be a sort, so we get a universe variable `u` and `P : nat -> Type@{u}`. Universe variables are all `>= Set` so it then cannot be made into a Prop.

herbelin:
> When typing `h:P x` the type of `P x` needs to be a sort, so we get a universe variable `u` and `P : nat -> Type@{u}`. Universe variables are all >= `Set` so it then cannot be made into a `Prop`.

From a user point of view, this is a pretty disconcerting approximation and loss of solution with respect to the initial problem, isn't it?

herbelin:
@mattam82: Did you recover some available time and be willing to be assignee for this PR?

ppedrot (commented):
I understand the goals but I don't understand the heuristic to backtrack on the first error if a constraint cannot be solved. Can you explain the intent @herbelin?

herbelin:
It backtracks on the 1st failure only if the second is `ProblemBeyondCapabilities`. I.e. it is a heuristic which starts from the assumption that we can always have a more informative error than `ProblemBeyondCapabilities`. Ideally, one could try to compare all possible pairs of errors and select for each pair which one is most relevant. I worked empirically, starting with one example, but maybe we should try as soon as now to have a more global view??

ppedrot:
This PR doesn't look hurtful, so I guess we can merge it even if we don't really know where we're heading. I'll merge it soon assuming nobody reacts.

ppedrot (approved):

PR #7363, opened 2018-04-26
[ci] Fix another issue with the timing tests

JasonGross:
There was recently a spurious failure on AppVeyor (I've forgotten which
PR).  This commit fixes that particular failure.


SkySkimmer:
I'll take this PR.

SkySkimmer (approved):
TBH I don't understand what this file tries to test with all these patches but I'll trust @JasonGross that it's useful.
If nobody pipes up I'll merge in a few days.

JasonGross:
> TBH I don't understand what this file tries to test with all these patches

The important thing for this file to test is that passing the makefile the relevant timing arguments results in a log which is appropriately-parseable by the python script, and that the python script picks up the correct lines.  If, e.g., `coqc -time` stopped working, or the makefile invocation for the timing stuff broke, this file would catch that.  Admittedly, it was originally supposed to test more than that, but timing is so unreliable in Coq that we can't really do that.

PR #7365, opened 2018-04-27
Mini fixes in the tactics chapter

herbelin:
**Kind:** bug fix 

In passing, fixing two broken links and adding a few `:tacn:` hyperlinks.

### DIFF
--- a/doc/sphinx/proof-engine/tactics.rst
+++ b/doc/sphinx/proof-engine/tactics.rst
@@ -96,10 +96,10 @@ bindings_list`` where ``bindings_list`` may be of two different forms:
 
 + A bindings list can also be a simple list of terms :n:`{* term}`.
   In that case the references to which these terms correspond are
-  determined by the tactic. In case of ``induction``, ``destruct``, ``elim``
-  and ``case`` (see :ref:`ltac`) the terms have to
+  determined by the tactic. In case of :tacn:`induction`, :tacn:`destruct`, :tacn:`elim`
+  and :tacn:`case`, the terms have to
   provide instances for all the dependent products in the type of term while in
-  the case of ``apply``, or of ``constructor`` and its variants, only instances
+  the case of :tacn:`apply`, or of :tacn:`constructor` and its variants, only instances
   for the dependent products that are not bound in the conclusion of the type
   are required.
 
--- a/doc/sphinx/proof-engine/tactics.rst
+++ b/doc/sphinx/proof-engine/tactics.rst
@@ -503,7 +503,7 @@ Applying theorems
 
 .. tacv:: eapply {+, @term with @bindings_list} in @ident as @intro_pattern.
 
-   This works as :tacn:`apply ... in as`  but using ``eapply``.
+   This works as :tacn:`apply ... in ... as`  but using ``eapply``.
 
 .. tacv:: simple apply @term in @ident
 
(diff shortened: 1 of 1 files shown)


Zimmi48 (approved):
Thanks but may I request that #7301 is merged before? I can't be expected to rebase it once more. (Note that I ignore if these two PRs will conflict.)

herbelin:
> Thanks but may I request that #7301 is merged before?

Sure. You may even cherry-pick this PR to yours if you want. I just checked and there was just one redundancy. I just removed it from my PR so that they can be merged.

Zimmi48:
> You may even cherry-pick this PR to yours if you want.

I don't want to change it because then the tests would have to run again and this would delay merging once more ;-)

Zimmi48:
There are fatal warnings when using a reference to an non-existing label. But apparently, there are no such warnings when the same happens with `:tacn:` or `:cmd:` or `:opt:`. Maybe because the first one is natively supported by Sphinx while the second one was added by @cpitclaudel?

herbelin:
> Users can use it to better understand error messages they get. Currently the index is incomplete.
> What I propose is just to make it more complete. 

That's a good point. So, in the short term, maybe the index should be renamed into "Errors and Warnings (selection)"?

A quick grep gives 72 `CWarnings.create`, about 900 `user_err`, 190 errors centralized in `himsg.ml`, and about 200 errors or warnings in the index. So a ratio of 1/6, though probably among the most standard and thus useful ones.

So, we can maybe try to make progress. I don't know how we could organize but the experience showed that we can achieve big tasks collectively. So, assuming we add an argument to each `user_err` to tell to which command it pertains, that we add systematically in the source of the manual a link to a section listing the errors attached to each command, that we have a tool to automatically translate such `user_err` into a list of errors for the given command, and, finally, that we share the work for indexing all such `user_err`. Do you think it is feasible?

Additionally, that would help us to have more uniform error messages.

cpitclaudel:
> Maybe because the first one is natively supported by Sphinx while the second one was added by @cpitclaudel?

That's right.  In coqdomain.py, can you try changing the roles section from this:

```
    roles = {
        # Each of these roles lives in a different semantic “subdomain”
        'cmd': XRefRole(),
        'tac': XRefRole(),
        'tacn': XRefRole(),
        'opt': XRefRole(),
        'thm': XRefRole(),
        'prodn' : XRefRole(),
        'exn': XRefRole(),
        'warn': XRefRole(),
```

to this?

```
    roles = {
        # Each of these roles lives in a different semantic “subdomain”
        'cmd': XRefRole(warn_dangling=True),
        'tac': XRefRole(warn_dangling=True),
        'tacn': XRefRole(warn_dangling=True),
        'opt': XRefRole(warn_dangling=True),
        'thm': XRefRole(warn_dangling=True),
        'prodn' : XRefRole(warn_dangling=True),
        'exn': XRefRole(warn_dangling=True),
        'warn': XRefRole(warn_dangling=True),
```

Zimmi48:
This works! (some fatal warnings are emitted) Great!

maximedenes:
@herbelin Would you mind rebasing this? It is next in my Sphinx-PRs list. Thanks!

herbelin:
Rebase done.

maximedenes (approved):

PR #7367, opened 2018-04-27
circle CI: do not use cache from old config.yml versions

SkySkimmer:
It begs things to break when the cache is out of sync with the system packages.


ejgallego:
Indeed it makes sense, I wonder why the second line was there in the first place?

SkySkimmer:
I didn't expect it to introduce unstability, so I thought it would just speed things up when adding a new job.
In hindsight it's not worth it.

ejgallego (approved):
I do agree, adding a new job is rare enough and rebuilding the cache is a good sanity check.

ejgallego:
Waited fro appveyor to finish just to have it randomly fail :S

PR #7369, opened 2018-04-27
An attempt to distinguish "sort" variables from "Type" variables in pretyping (fixes #4328, #3529)

herbelin:
**Kind:** bug fix / enhancement

Fixes / closes #4328. Added: also fixes #3529.

When an evar has to be instantiated by an unkown sort, a level >= `Set` is generated, possibly prevening further unification of the sort with `Prop` to happen (see discussion at #7361 and #4328).

This relaxes the constraint that such unknown sort is necessarily >= `Set`.

At the current time, this is done in an ad hoc way, while waiting for the feedback from universe experts.

The only kernel modification is to allow to declare variables >= `Prop` instead of necessarily >= `Set`. I thus make here the assumption that the kernel graph algorithm, which includes `Prop` as a node, is indeed designed to deal correctly with `Prop` as a level.

The main modification is otherwise in `uState.ml`, where I added an extra case to collapse `t` into `Prop` whenever we need `t <= Prop`. If what I'm doing makes sense, this shall have to be more smoothly integrated to the rest of the `process_universe_constraints` algorithm.

### DIFF
--- a/engine/uState.ml
+++ b/engine/uState.ml
@@ -481,7 +481,7 @@ let emit_side_effects eff u =
   let uctxs = Safe_typing.universes_of_private eff in
   List.fold_left (merge true univ_rigid) u uctxs
 
-let new_univ_variable ?loc rigid name
+let new_univ_variable ?loc ?(above_set=true) rigid name
   ({ uctx_local = ctx; uctx_univ_variables = uvars; uctx_univ_algebraic = avars} as uctx) =
   let u = UnivGen.new_univ_level () in
   let ctx' = Univ.ContextSet.add_universe u ctx in
@@ -499,13 +499,13 @@ let new_univ_variable ?loc rigid name
     | Some n -> add_uctx_names ?loc n u uctx.uctx_names
     | None -> add_uctx_loc u loc uctx.uctx_names
   in
-  let initial =
-    UGraph.add_universe u false uctx.uctx_initial_universes
-  in                                                 
+  let add u =
+    if above_set then UGraph.add_universe u false
+    else UGraph.add_universe_unconstrained u in
   let uctx' =
     {uctx' with uctx_names = names; uctx_local = ctx';
-                uctx_universes = UGraph.add_universe u false uctx.uctx_universes;
-                uctx_initial_universes = initial}
+                uctx_universes = add u uctx.uctx_universes;
+                uctx_initial_universes = add u uctx.uctx_initial_universes}
   in uctx', u
 
 let make_with_initial_binders e us =
--- a/pretyping/evardefine.ml
+++ b/pretyping/evardefine.ml
@@ -163,7 +163,7 @@ let rec evar_absorb_arguments env evd (evk,args as ev) = function
 (* Refining an evar to a sort *)
 
 let define_evar_as_sort env evd (ev,args) =
-  let evd, u = new_univ_variable univ_rigid evd in
+  let evd, u = new_univ_variable univ_flexible evd in
   let evi = Evd.find_undefined evd ev in 
   let s = Type u in
   let concl = Reductionops.whd_all (evar_env evi) evd evi.evar_concl in
--- a/engine/evd.ml
+++ b/engine/evd.ml
@@ -784,7 +784,7 @@ let new_univ_level_variable ?loc ?name rigid evd =
     ({evd with universes = uctx'}, u)
 
 let new_univ_variable ?loc ?name rigid evd =
-  let uctx', u = UState.new_univ_variable ?loc rigid name evd.universes in
+  let uctx', u = UState.new_univ_variable ?loc ~above_set:false rigid name evd.universes in
     ({evd with universes = uctx'}, Univ.Universe.make u)
 
 let new_sort_variable ?loc ?name rigid d =
--- a/engine/uState.mli
+++ b/engine/uState.mli
@@ -107,7 +107,7 @@ val merge : ?loc:Loc.t -> bool -> rigid -> t -> Univ.ContextSet.t -> t
 val merge_subst : t -> UnivSubst.universe_opt_subst -> t
 val emit_side_effects : Safe_typing.private_constants -> t -> t
 
-val new_univ_variable : ?loc:Loc.t -> rigid -> Id.t option -> t -> t * Univ.Level.t
+val new_univ_variable : ?loc:Loc.t -> ?above_set:bool -> rigid -> Id.t option -> t -> t * Univ.Level.t
 val add_global_univ : t -> Univ.Level.t -> t
 
 (** [make_flexible_variable g algebraic l]
--- a/test-suite/bugs/closed/3529.v
+++ b/test-suite/bugs/closed/3529.v
@@ -1,0 +1,16 @@
+(* Check that Prop can be inferred to fill some sorts *)
+
+Module Example1.
+Class R T := { f: T -> Prop }.
+Instance r : R Prop := {| f P := P -> P |}.
+Definition foo : R Prop := Build_R _ (fun P => P -> P).
+End Example1.
+
+Require Coq.Setoids.Setoid.
+Import Coq.Setoids.Setoid.
+
+Module Example2.
+Class ILogicOps Frm := { lentails: relation Frm }.
+Instance ILogicOps_Prop : ILogicOps Prop | 2 := {| lentails P Q := P -> Q |}.
+Definition ILogicOps_Prop' : ILogicOps Prop := {| lentails P Q := P -> Q |}.
+End Example2.
--- a/test-suite/bugs/closed/4328.v
+++ b/test-suite/bugs/closed/4328.v
@@ -1,6 +1,6 @@
 Inductive M (A:Type) : Type := M'.
 Axiom pi : forall (P : Prop) (p : P), Prop.
-Definition test1 A (x : _) := pi A x.           (* success     *)
-Fail Definition test2 A (x : A) := pi A x.           (* failure ??? *)
-Fail Definition test3 A (x : A) (_ : M A) := pi A x. (* failure     *)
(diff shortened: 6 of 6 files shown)


engine/uState.ml
```diff
@@ -226,7 +226,11 @@ let process_universe_constraints ctx cstrs =
                     let levels = Univ.Universe.levels l in
                     let fold l' local =
                       let l = Univ.Universe.make l' in
-                      if Univ.Level.is_small l' || is_local l' then
+                      if Univ.Level.is_prop r' then begin
+                        vars := Univ.LMap.add l' None !vars;
```
SkySkimmer: This seems wrong. instantiate_variable should be sufficient.

kernel/uGraph.ml
```diff
@@ -521,7 +522,13 @@ let add_universe vlev strict g =
     in
     let entries = UMap.add vlev (Canonical v) g.entries in
     let g = { entries; index = g.index - 1; n_nodes = g.n_nodes + 1; n_edges = g.n_edges } in
-    insert_edge strict (get_set_arc g) v g
+    insert_edge strict (get_arc g) v g
+
+let add_universe vlev strict g =
+  add_universe_above vlev strict get_set_arc g
```
SkySkimmer: Don't pass a function, pass the actual universe. ie `add_universe_above vlev strict Level.set g`.

SkySkimmer:
>The only kernel modification is to allow to declare variables >= Prop instead of necessarily >= Set. I thus make here the assumption that the kernel graph algorithm, which includes Prop as a node, is indeed designed to deal correctly with Prop as a level.

The graph should deal with it OK, but
- it will allow universes `Prop < i < Set` with unclear consequences 
- various places may assume that variables are >= Set. eg Universes.normalize_context_set which includes minimisation to Set and drops constraints `Set <= foo` if that's disabled. I don't know if anywhere in the kernel makes that assumption.

engine/uState.ml
herbelin: I did not change the `rigid` status given to sort variables in `Evarutil.new_type_evar` (and ...), so, the level `l'` is not added to `vars`/`uctx_univ_variables` (in `new_univ_variable`) and `instantiate_variable` would fail without the explicit extra `Univ.LMap.add`.

More fundamentally, I did not understand whether the status `flexible` was reserved to universe polymorphism or if it could have been used in the current (non-polymoprhic) situation.

kernel/uGraph.ml
herbelin: Done. By the way, is there a way to declare a universe without having an edge on it, since, here, after all, the condition `>= Prop` always holds, so I would assume that we can just declare a node without an edge?

engine/uState.ml
SkySkimmer: Making arbitrary levels flexible is definitely wrong.
rigid is for `Type` (without annotation) and named levels, I think everything else should be flexible.
Evarutil.new_type_evar is passed the desired rigidity so I don't understand what you mean there.

SkySkimmer: define_evar_as_sort uses univ_rigid, this seems questionable.

kernel/uGraph.ml
SkySkimmer: I don't understand what you mean. Creating the edge is what ensures the condition holds. It will make it so if we have `foo <= Prop` it will identify with Prop, and it will prevent `Type@{bar} : Prop` which has I have no idea what consequences.

herbelin:
> it will allow universes `Prop < i < Set` with unclear consequences

With the PR, I still get a universe inconsistency on:
```
Universe i.
Constraint Prop < i.
Constraint i < Set.
```
and I'd feel safer to understand what still prevents this (even though that, at first view, this variant of ¬CH is consistent).

About the second point, is there some particular parts of the code that I should check.

kernel/uGraph.ml
herbelin: Ah, yes, you're right. I was reasoning with a (natively) well-founded graph in mind, but in practice, the algorithm is for any graph, so the constraint is needed.

SkySkimmer:
>and I'd feel safer to understand what still prevents this (even though that, at first view, this variant of ¬CH is consistent).

`Universe i.` doesn't go through the new code path. In fact it doesn't go through UState at all, the universe is created in Declare.do_universe and added to the environment in Declare.add_universe.

herbelin:
OK, with the implicit semantics that `Prop` is excluded` from the range of variable universes.

How would you generate a problem `Prop < i < Set` then, for an `i` which is not an explicit universe?

engine/uState.ml
herbelin: I updated the PR, fixing `define_evar_as_sort` and now interpreting flexible as existential (the usual way then).

SkySkimmer:
I don't know if it's actually possible, because strict constraints come from `i+1 <= j` constraints and `Prop+1` is reduced to `Set+1` in Universe.super so we can't get a `Prop+1 <= j` constraint.

herbelin:
Up to the name of new function (`add_unconstrained_universe`) and new option added (`above_set`) which I don't find very nice but which have the advantage of being strict extensions of the API, the PR looks ok to me. So, (further) comments welcome.

Also, once names of function/option stabilized, I can eventually add some quick documentation of the API additions.

SkySkimmer:
Multiple CI and test suite failures.

pretyping/evardefine.ml
```diff
@@ -163,7 +163,7 @@ let rec evar_absorb_arguments env evd (evk,args as ev) = function
 (* Refining an evar to a sort *)
 
 let define_evar_as_sort env evd (ev,args) =
-  let evd, u = new_univ_variable univ_rigid evd in
+  let evd, u = new_univ_variable univ_flexible_alg evd in
```
SkySkimmer: I don't think allowing algebraic is correct here.
IIUC for instance `let s := _ in let t := _ : s in fun x : t => ...` will define_evar_as_sort on the hole for `s` so we would get an algebraic in a let body.

herbelin: Seems indeed to make sense.

herbelin:
@SkySkimmer: let's see if the `alg` fix improves things, but there is at least one issue remaining:
```
Universe i.
Axiom M : Type@{i} -> Type.
Axiom pi : forall (P : Prop) (p : P), Prop.
Definition test1 (A : ?[X])  (x : A) (_ : M A) := pi A x. (* success *)
Definition test2 (A : ?[X]) (_ : M A) (x : A) := pi A x. (* universe inconsistency "Type <= Prop" *)
```
This shows that the inference is still not robust up to the order of arguments. This seems to be caused by the `refresh_universes` in `Evarsolve.evar_define`. In unifying `?X :<= Type@{i}`, it refreshes `i` with a rigid universe. I'd be tempted to refresh it instead with a flexible status (i.e. calling `refresh_universes` with `~status:univ_flexible` as it is done in some other places of `Evarsolve`. Would this sound correct to you?.

Incidentally, I saw that we spend time in checking the type of evars used as type of variables. Wouldn't it be worth to have a dedicated kind of evars for types of variables since the type of these types shall not be part of a unification problem and shall not have to be inferred explicitly, nor to be unified explicitly with some other type. What do you think?

SkySkimmer:
I don't understand refresh much, I would say try it and if it doesn't blow up the CI we can ask @mattam82.

herbelin:
@skyskimmer: Yet another source of asymmetry:
```coq
Check fun (M : Prop) (f : ?[A] -> Prop) => (f Prop, f M). (* OK *)
Check fun (M : Prop) (f : ?[A] -> Prop) => (f M, f Prop). (* universe inconsistency *)
```
It seems that `refresh_universe` should refresh `Prop` into some flexible `Type` when the constraints is `Prop <= ?A`. Shall I do it?

herbelin:
I'm (temporarily) renouncing to improve this further. It solves a few unexpected failures when unification needs `Prop` subtyping but not yet all. Ready as it is on my side.

SkySkimmer:
@herbelin the failures in hott, fiat-crypto and equations are real incompatibilities, you need to investigate and decide whether to restore compatibility or provide an overlay.

herbelin:
The failure in Equations comes from the fact that the PR changes the order in which polymorphic arguments are generated:
```coq
simplification_sigma1@{i j k} : forall (A : Type@{i}) (P : A -> Type@{j}) (B : Type@{k}), ...
```
is now
```coq
simplification_sigma1@{j k i} : forall (A : Type@{i}) (P : A -> Type@{j}) (B : Type@{k}), ...
```
I don't know what I should do. Is this a bug that I should solve, or is the order in which variables are generalized unspecified and the correct fix is to explicitly give the order of universe variable at the definition of `simplification_sigma1`?

herbelin:
> I think the order when not annotated is unspecified.

So, presumably an evar-evar order issue? I don't remember well where we were on this question

ppedrot (commented):
I don't have any opinion on this PR. We need to think about the incompatibilities though.

mattam82:
@herbelin Indeed it seems natural to refresh with a flexible. However I guess the issue here is that it uses a rigid universe to avoid ever capturing a template polymorphic universe, which is basically what the refinements of refresh I introduced are about.

mattam82:
For simplification_sigma1 I think the right fix is to be explicit in the signature of the function.

mattam82:
About refreshing Prop to a flexible type when encountering Prop <= ?A that sounds ok too.

herbelin:
Thanks for the comments. I have to concentrate back to this PR, which I'll do at some time.

SkySkimmer:
I think this may be superseded by https://github.com/coq/coq/pull/9444

ppedrot:
Given that it was clearly labelled as an attempt, and that it failed, I think we should close this PR.

PR #7374, opened 2018-04-27
[sphinx] More fatal warnings.

Zimmi48:
This first commit was proposed by @cpitclaudel in #7365.
In order to merge this, we need to fix these new warnings.
But before getting started on this task, let's wait on #7301 being merged.

### DIFF
--- a/doc/sphinx/addendum/generalized-rewriting.rst
+++ b/doc/sphinx/addendum/generalized-rewriting.rst
@@ -599,6 +599,7 @@ Notice that the syntax is not completely backward compatible since the
 identifier was not required.
 
 .. cmd:: Add Morphism f : @ident
+   :name: Add Morphism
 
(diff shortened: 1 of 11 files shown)


Zimmi48:
This is now on top of #7301 and includes #7365 and fixes all new warnings.

Zimmi48:
Merging this and other Sphinx PRs at the same time without testing their interaction is dangerous, since this adds new fatal warnings. The best course of action would be to do a batch merge and running CI before pushing upstream. Another solution is to merge this and to request other Sphinx PRs to rebase before merging them.

maximedenes:
Will merge right after #7365.

Zimmi48:
Unfortunately, this was merged without being tested on a recent version of the manual. And since it introduced new fatal warnings, it broke the compilation, as could be expected. Fortunately, #7487 already fixes this.

Issue #7377, opened 2018-04-28
Latex packages for building coq doc


SkySkimmer:
What are the latex packages necessary to compile Coq documentation?

In dev/doc/setup.txt and INSTALL.doc we say that for debian-like distributions package texlive-math-extra is used to compile the doc, but this package is not provided in ubuntu 18.04 (just released) or current debian.

Also with sphinx the latex we use has been reduced so the packages used may have reduced too.

ejgallego:
I'd bet that `texlive-math-extra` can be safely removed. Let me do a doc build and I see.

ejgallego:
Docs build fine in 17.10 without that package; as for Sphinx, I got success but had still to install `sphinxcontrib-bibtex` by hand, 18.04 will include this package so we can test again when it is released.

ejgallego:
So I suggest we try a 18.04 build without `pip3` but adding the https://packages.ubuntu.com/bionic/python3-sphinxcontrib.bibtex package and basic texlive + hevea which is still needed.

PR #7379, opened 2018-04-28
[doc] Update Sphinx build instructions for Debian derivatives.

ejgallego:
This can still be improved as I am sure we have too many packages listed.

cc: #7377 

Depends on #7381.

### DIFF
--- a/INSTALL.doc
+++ b/INSTALL.doc
@@ -31,16 +31,26 @@ To produce all the documents, the following tools are needed:
   - Antlr4 runtime for Python 3
 
 
-Under Debian based operating systems (Debian, Ubuntu, ...) a
-working set of packages for compiling the documentation for Coq is:
+Under recent Debian based operating systems (Debian 10 "Buster",
+Ubuntu 18.04, ...) a working set of packages for compiling the
+documentation for Coq is:
 
-  texlive texlive-latex-extra texlive-math-extra texlive-fonts-extra
-  texlive-humanities texlive-pictures latex-xcolor hevea python3
-  python-pip3
+  texlive-latex-extra texlive-fonts-recommended hevea python3-sphinx
+  python3-pexpect python3-sphinx-rtd-theme python3-bs4
+  python3-sphinxcontrib.bibtex python3-pip
 
-To install the Python packages required to build the user manual, run:
+Then, install the Python3 Antlr4 package:
+
+  pip3 install antlr4-python3-runtime
+
+Nix users should get the correct development environment to build the
+Sphinx documentation from Coq's `default.nix`. [Note Nix setup doesn't
+include the LaTeX packages needed to build the full documentation.]
+
+If you are in an older/different distribution you can install the
+Python packages required to build the user manual using python3-pip:
+
   pip3 install sphinx sphinx_rtd_theme beautifulsoup4 antlr4-python3-runtime pexpect sphinxcontrib-bibtex
-
 
 Compilation
 -----------
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,7 @@ plugins/micromega/csdpcert
 plugins/micromega/.micromega.ml.generated
 kernel/byterun/dllcoqrun.so
 coqdoc.sty
+coqdoc.css
 time-of-build.log
 time-of-build-pretty.log
 time-of-build-before.log


INSTALL.doc
```diff
@@ -34,17 +34,18 @@ To produce all the documents, the following tools are needed:
...
-  imagemagick
-  python3 python-pip3
+  imagemagick python3-sphinx python3-sphinxcontrib-bibtex
 
-To install the Python packages required to build the user manual, run:
-  pip3 install sphinx sphinx_rtd_theme beautifulsoup4 antlr4-python3-runtime pexpect sphinxcontrib-bibtex
+If you need to install the Python packages required to build the user
+manual in older distributions, install the python3-pip package and run:
 
+  pip3 install sphinxcontrib-bibtex
```
Zimmi48: I would suggest that you keep the full `pip3` command so that people willing to use `pip` know what they have to install (consider other distributions which may not be providing the Sphinx packages -- I don't know if there are any though).

Comments from @maximedenes or @cpitclaudel welcome.

ejgallego: That's a good idea but I'll put it in a separate section, note that I think the previous `pip3` command was not really exhaustive, it also seemed to include some strange stuff.

Zimmi48: OK that works for me.

Zimmi48 (approved):
Looks good!
Indeed the list of LaTeX packages might still be too long but this PR doesn't make things worse. Anyways, we should look into reducing these dependencies, for instance by removing the tutorials from the repo.

INSTALL.doc
cpitclaudel: I think everything in that pip command was needed; what was wrong in it?

I'd recommend against using Debian packages for sphinx.  It's much better if we're all running an up-to-date version.

ejgallego: Note that Nix is also not using `pip`.

ejgallego: > I think everything in that pip command was needed; what was wrong in it?

Doesn't pip have dependency tracking?

Zimmi48: Hum, but Nix pretty much always has up-to-date versions of packages.

ejgallego: > Hum, but Nix pretty much always has up-to-date versions of packages.

That was not true last time I looked [like OCaml 4.06.0 took a long time to arrive, for example]

ejgallego: Anyways, Debian support is out of the question, I think, so for practical purposes documentation will have to work with Sphinx 1.7.4

maximedenes: > That was not true last time I looked [like OCaml 4.06.0 took a long time to arrive, for example]

I'm pretty sure you made this false claim before, and were told it was BS. OCaml 4.06.0 was announced on Nov 3, 2017 on the caml-list, and committed to Nixpkgs by @vbgl on... Nov 3, 2017. How is that a long time?

maximedenes: See https://github.com/NixOS/nixpkgs/commit/4f7c0aed38d0dc3f7f8650d89949b1914b54aef1 if you don't trust me.

maximedenes: And I can confirm Nixpkgs as all our documentation packages up to date, I pushed/updated the missing ones myself.

ejgallego: Maybe it was 4.05.0... I don't remember now, the chat was on gitter anyways; but indeed Nix doesn't get automagically up to date so it is subject to all the typical problems every package system has. It is very tiring it seems everyone here has discovered the holy grail, either in Cargo, Nix, or whatever the new fad is, however when it comes to code, merges, and PRs, the powers of the holy grail are not so effective.

ejgallego: Indeed at least for `nums` the situation took a while to get sorted out in Nix 4.06.0...

ejgallego: Also, for 4.06.0 we were testing the beta version way before the release date, which was not possible to do with Nix as nobody had volunteered to do the packaging. So yeah, Nix is not so much up to date than Debian `sid` or Ubuntu....

Zimmi48: The big difference is that it's much more easy to get a package in nixpkgs than in Debian. If we were relying on Nix for our CI testing and a package was missing, we would push it ourselves. And even before it is accepted, you can already use it ;-)

ejgallego: Really folks, I am not sure if you are trolling anymore or you are serious:

> And even before it is accepted, you can already use it ;-)

Obviously Debian can do this, and in fact, this is how we bootstrapped the whole CI infrastructure as a recent OPAM is not available in the default Travis image [which is Ubuntu 12.04!!]

Going back ontopic, Ubuntu stable is shipping 1.6.7 and that for me seals the deal for the short term.

It is not clear if Debian Buster will use sphinx 1.8 or 1.6; both PIP and Debian sid have today 1.7.4, whereas Nix provides 1.7.2. Oh!! How could that happen? It turns out that Debian is more up to date than Nix...

ejgallego:
Ok, did another pass and synched with the other PRs as they were conflicting otherwise.

INSTALL.doc
maximedenes: > It is not clear if Debian Buster will use sphinx 1.8 or 1.6; both PIP and Debian sid have today 1.7.4, whereas Nix provides 1.7.2. Oh!! How could that happen? It turns out that Debian is more up to date than Nix...

I'm not sure what you are trying to say, but to add a bit of context, Sphinx 1.7.3 was released 6 days ago, and 1.7.4 3 days ago. So it sounds ok to me that the packages haven't been updated yet. But again, I'm probably missing the point of these comparisons anyway.

INSTALL.doc
```diff
@@ -34,17 +31,25 @@ To produce all the documents, the following tools are needed:
...
+  python3-pexpect python3-sphinx-rtd-theme python3-bs4
+  python3-sphinxcontrib.bibtex python3-pip
 
-To install the Python packages required to build the user manual, run:
-  pip3 install sphinx sphinx_rtd_theme beautifulsoup4 antlr4-python3-runtime pexpect sphinxcontrib-bibtex
+Then, install the Python3 Antlr4 package:
+
+  pip3 install antlr4-python3-runtime
+
+Nix users should get the correct development enviroment from the main
```
maximedenes: typo

INSTALL.doc
ejgallego: I do remember the original discussion now, and your use of the word "BS" is just "BS" [and not very polite]. The day 4.06.0 was released, Coq [except for lablgtk] was ready for it. However, would I have been using Nix exclusively, that wouldn't have had happened, as when I asked how could I try 4.06.0+beta in Nix the answer was "you need to wait" [or to package it myself I guess].

> But again, I'm probably missing the point of these comparisons anyway.

People is claiming "Debian cannot do this" or "Debian ships old Python packages" or "Debian is hard to get packages into", which are all, TTBOMK not faithful to reality, on the other hand people is proposing Nix as an alternative for the base image, and while it'd be cool to have an image with Nix, I am failing to see what we are exactly gaining and I do see that we are losing first-class support for the most used Linux platform.

INSTALL.doc
ejgallego: Point taken, I will open a different issue for removing LaTeX from Coq.

ejgallego: See #7390

PR #7381, opened 2018-04-28
[gitlab] Update base image to Ubuntu bionic + some improvements.

ejgallego:
We move gitlab runners to Ubuntu 18.04 "Bionic"; this allows us to
install most base dependencies using APT, and opens up the door to
saving quite a bit of time by creating a custom docker image
[c.f. #7383]

This change comes with an update of dependencies; we tweak them. Also:

- we add a more precise cache `key` constraint; this is still done
  manually, we should develop an automated way in another PR. The
  format is `$image-v$date-$hour-$minute`

- we export `DEBIAN_FRONTEND=noninteractive` as to avoid problems with
  package installs that ask for interactive input.

- we install Sphinx Python packages using `apt` save for
  `python3-antlr4`, which is still unpackaged [see
  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=897129 ]

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,12 +1,15 @@
-image: ubuntu:xenial
+image: ubuntu:bionic
 
 stages:
   - opam-boot
   - build
   - test
 
+# some default values
 variables:
-  # some default values
+  # Format: $IMAGE-V$DATE-$HOUR-$MINUTE
+  CACHEKEY: bionic-V2018-04-29-00-50
+  DEBIAN_FRONTEND: "noninteractive"
   NJOBS: "2"
   COMPILER: "4.02.3"
   CAMLP5_VER: "6.14"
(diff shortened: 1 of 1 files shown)


SkySkimmer:
>[ERROR] lablgtk.2.18.6 is not available because your system doesn't comply with
        ocaml-version >= "4.06".


What even

.gitlab-ci.yml
```diff
@@ -16,27 +16,23 @@ variables:
...
+  TIMING_PACKAGES: "time python3"
 
   COQIDE_PACKAGES: "libgtk2.0-dev libgtksourceview2.0-dev"
   #COQIDE_PACKAGES_32BIT: "libgtk2.0-dev:i386 libgtksourceview2.0-dev:i386"
-  COQIDE_OPAM: "lablgtk-extras"
-  COQIDE_OPAM_BE: "lablgtk.2.18.6 lablgtk-extras.1.6"
-  COQDOC_PACKAGES: "texlive-latex-base texlive-latex-recommended texlive-latex-extra texlive-math-extra texlive-fonts-recommended texlive-fonts-extra latex-xcolor ghostscript transfig imagemagick tipa python3-pip"
-  COQDOC_OPAM: "hevea"
-  SPHINX_PACKAGES: "bs4 sphinx sphinx_rtd_theme pexpect antlr4-python3-runtime sphinxcontrib-bibtex"
+  COQIDE_OPAM: "lablgtk.2.18.6 conf-gtksourceview.2"
```
SkySkimmer: Why are you setting the version here?

ejgallego:
@SkySkimmer note the workaround for the cache issue, it is not the holy grail but better than nothing

.gitlab-ci.yml
```diff
@@ -178,16 +174,16 @@ opam-boot:
   cache:
     paths: &cache-paths
       - .opamcache
-    key: main
+    key: main-bionic-V2018-04-28-A
```
SkySkimmer: I don't like this system. Updating the keys everytime we need to is going to be a pain.
How about instead of using `key` we do
- at the beginning of opam-boot script, `if [ "$(cat .opamcache/key)" != "$(md5sum .gitlab-ci.yml)" ]; then rm -rf .opamcache; fi`
- at the end `md5sum .gitlab-ci.yml > .opamcache/key`

ejgallego: Not so much of pain IMO but indeed prone to manual errors. I dunno, we can do the hack too if you prefer until `key` gets stronger in GitLab. Anyways #7383 should help with some problems IMHO.

SkySkimmer: If you go with the yml key you should use a variable `CACHEKEY: "V..."` instead of repeating it 3 times.
Also I don't see the point of putting "bionic" in there.

ejgallego: Good idea, key variable refactored.

> Also I don't see the point of putting "bionic" in there.

I am trying to put as much info as I can; bionic is useful as for example having a different key let's say, for Nix.

SkySkimmer:
The job in a previous pipeline https://gitlab.com/coq/coq/-/jobs/65514721 got stuck when apt-get wanted user interaction for tzdata.
I remember encountering this in the past but I don't remember how I got rid of it.

ejgallego:
> I remember encountering this in the past but I don't remember how I got rid of it.

Should be solved by the current version of the PR [see `DEBIAN_FRONTEND`]

.gitlab-ci.yml
```diff
@@ -7,6 +7,8 @@ stages:
 
 variables:
   # some default values
+  CACHEKEY: bionic-V2018-04-28-B
```
SkySkimmer: I don't like having some crazy ad-hoc thing for the key.
Just use a timestamp or an incrementing number.

ejgallego: That's a timestamp.

ejgallego: format = `$IMAGE-v$DATE-$VERSION`

ejgallego: Where `$REVISION` is needed as to avoid shooting myself in the foot today.

SkySkimmer: Don't put image in it, and if you put the date in it just make it precise enough to serve as the whole thing.

ejgallego: > Don't put image in it,

Why? It seems useful to me.

.gitlab-ci.yml
```diff
@@ -16,26 +18,26 @@ variables:
...
-  SPHINX_PACKAGES: "bs4 sphinx sphinx_rtd_theme pexpect antlr4-python3-runtime sphinxcontrib-bibtex"
+  SPHINX_PACKAGES: "antlr4-python3-runtime"
+  COQIDE_OPAM: "lablgtk.2.18.5 conf-gtksourceview.2"
+  COQIDE_OPAM_BE: "lablgtk.2.18.6 conf-gtksourceview.2"
+  COQDOC_PACKAGES: "texlive-latex-recommended texlive-fonts-recommended transfig imagemagick hevea python3-sphinx python3-pexpect python3-sphinx-rtd-theme python3-bs4 python3-sphinxcontrib.bibtex python3-pip"
   ELPI_OPAM: "elpi"
 
-
 before_script:
+  - cat /proc/{cpu,mem}info || true
```
SkySkimmer: ??

ejgallego: > ??

I guess this is clear; I want to get more information about runners; there are quite wild differences; suggestions welcome, including using `docker` client commands.

ejgallego:
Ok, so I think this is coming very close to be ready on my end.

ejgallego:
Rebased on top of #7386 as to avoid a conflict.

ejgallego:
Will also make the commit message more precise.

Zimmi48 (commented):
Nothing to say. My request for review is just because of this PR dependency.

SkySkimmer:
Removing @letouzey review request as it's only for the dependency PR.

ejgallego:
I'm happy with this, IMHO we can merge as gitlab is still having some issues with the caches [so it seems] and this should definitively help.

ejgallego:
Sure.

.gitlab-ci.yml
```diff
@@ -7,6 +7,8 @@ stages:
 
 variables:
   # some default values
+  CACHEKEY: bionic-V2018-04-29-00-50
```
SkySkimmer: Also put the format in a comment.

SkySkimmer (approved):
Waiting for CI.
Is there a preferred timezone for the cache key?

PR #7387, opened 2018-04-29
[gitlab] [circleci] Use a Docker base image for CI

### LINKED ISSUE #7383: [ci] [gitlab] Create custom bionic docker build image.
In #7381, we set the base image to `ubuntu:bionic`. This is good but not good enough, as the base image is missing the set of apt packages that we need in latter phases, thus for example when the build phase is trigerred `apt` has to be run again to install the dependencies [which the cached .opam depends on]

We should indeed create an `ubuntu:bionic-coqbuild` image that includes the apt packages so indeed all the remaining setup work is just to do the `.opam` cache and we are ready to rumble.

ejgallego:
We provide a custom `Dockerfile` for Coq's CI system, based on
`ubuntu:bionic`. The image includes the required set of packages and
OPAM switches.

This greatly simplifies the Gitlab and Circle scripts, at the cost of
having to push a Docker build for them to depend on.

Travis is not included in this PR as it requires significant more
refactoring due to lack of native Docker support.

This is work in progress but ready, a build hook is used so the image
is properly tagged in the Docker autobuilder.

We need to improve the autobuilder setup but this last point requires
some design on how to trigger it.

Fixes #7383

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,83 +8,29 @@ defaults:
     # reference syntax)
     working_directory: ~/coq
     docker:
-      - image: ocaml/opam:ubuntu
+      - image: coqci/base:V2018-05-05
 
   environment: &envvars
-    # required by some of the targets, e.g. compcert, passed for
-    # instance to opam to configure the number of parallel jobs
-    # allowed
-    NJOBS: 2
-    COMPILER: "system"
-    CAMLP5_VER: "6.14"
     NATIVE_COMP: "yes"
-
-    # some useful values
-    TIMING_PACKAGES: &timing-packages "time python"
 
 version: 2
 
 before_script: &before_script
-  name: Install system packages
+  name: Setup OPAM Switch
   command: |
     echo export TERM=xterm >> ~/.profile
     source ~/.profile
-    printenv
-    if [ -n "${EXTRA_PACKAGES}" ]; then sudo apt-get update -yq && sudo apt-get install -yq --no-install-recommends ${EXTRA_PACKAGES}; fi
     echo . ~/.profile >> $BASH_ENV
-
-opam-switch: &opam-switch
-  name: Select opam switch
-  command: |
-    opam switch ${COMPILER}
+    printenv | sort
+    opam switch "$COMPILER"
     opam config list
     opam list
-
-.opam-boot-template: &opam-boot-template
-  <<: *params
-  steps:
-    - checkout
-    - run: *before_script
-    - run:
-        name: Cache selection
-        command: |
-          # We can't use environment variables in cache names
-          # So put it in a file and use the checksum
-          echo "$COMPILER" > COMPILER
-    - restore_cache:
-        keys:
-          - coq-opam-cache-v1-{{ arch }}-{{ checksum "COMPILER" }}-{{ checksum ".circleci/config.yml" }}-
-    - run:
-        name: Update opam lists
-        command: |
-          opam repository set-url default https://opam.ocaml.org
-          opam update
-    - run:
-        name: Install opam packages
-        command: |
-          opam switch -j ${NJOBS} ${COMPILER}
-          opam install -j ${NJOBS} -y camlp5.${CAMLP5_VER} ocamlfind ${EXTRA_OPAM}
-    - run:
-        name: Clean cache
-        command: |
-          rm -rf ~/.opam/log/
-    - save_cache:
-        key: coq-opam-cache-v1-{{ arch }}-{{ checksum "COMPILER" }}-{{ checksum ".circleci/config.yml" }}-
-        paths:
-          - ~/.opam
-    - persist_to_workspace:
-        root: &workspace ~/
-        paths:
-          - .opam/
 
 .build-template: &build-template
   <<: *params
   steps:
     - checkout
     - run: *before_script
-    - attach_workspace: &attach_workspace
-        at: *workspace
-    - run: *opam-switch
     - run: &build-configure
         name: Configure
         command: |
@@ -96,7 +42,7 @@ opam-switch: &opam-switch
           make -j ${NJOBS}
           make test-suite/misc/universes/all_stdlib.v
     - persist_to_workspace:
-        root: *workspace
+        root: &workspace ~/
         paths:
           - coq/
 
@@ -106,7 +52,9 @@ opam-switch: &opam-switch
   <<: *params
   steps:
     - run: *before_script
-    - attach_workspace: *attach_workspace
+    - attach_workspace: &attach_workspace
+        at: *workspace
+
     - run:
         name: Test
         command: |
@@ -115,18 +63,9 @@ opam-switch: &opam-switch
         root: *workspace
         paths:
           - coq/
-  environment: &ci-template-vars
-    <<: *envvars
-    EXTRA_PACKAGES: *timing-packages
 
 # Defines individual jobs, see the workflows section below for job orchestration
 jobs:
-
-  opam-boot:
-    <<: *opam-boot-template
-    environment:
-      <<: *envvars
-      EXTRA_OPAM: "ocamlgraph elpi"
 
   # Build and prepare test environment
   build: *build-template
@@ -136,24 +75,15 @@ jobs:
 
   color:
     <<: *ci-template
-    environment:
-      <<: *ci-template-vars
-      EXTRA_PACKAGES: *timing-packages
 
   compcert:
     <<: *ci-template
 
   coq-dpdgraph:
     <<: *ci-template
-    environment:
-      <<: *ci-template-vars
-      EXTRA_PACKAGES: "time python autoconf automake"
 
   coquelicot:
     <<: *ci-template
-    environment:
-      <<: *ci-template-vars
-      EXTRA_PACKAGES: "time python autoconf automake"
 
   elpi:
     <<: *ci-template
@@ -172,15 +102,9 @@ jobs:
 
   fiat-parsers:
     <<: *ci-template
-    environment:
-      <<: *ci-template-vars
-      EXTRA_PACKAGES: *timing-packages
 
   flocq:
     <<: *ci-template
-    environment:
-      <<: *ci-template-vars
-      EXTRA_PACKAGES: "time python autoconf automake"
 
   math-classes:
     <<: *ci-template
@@ -193,9 +117,6 @@ jobs:
 
   hott:
     <<: *ci-template
-    environment:
-      <<: *ci-template-vars
-      EXTRA_PACKAGES: "time python autoconf automake"
 
   iris-lambda-rust:
     <<: *ci-template
@@ -214,9 +135,6 @@ jobs:
 
   sf:
     <<: *ci-template
-    environment:
-      <<: *ci-template-vars
-      EXTRA_PACKAGES: "time python wget"
 
   unimath:
     <<: *ci-template
@@ -226,14 +144,11 @@ jobs:
 
 workflows:
   version: 2
+
   # Run on each push
   main:
     jobs:
-      - opam-boot
-
-      - build:
-          requires:
-            - opam-boot
+      - build
 
       - bignums: &req-main
           requires:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,76 +1,35 @@
-image: ubuntu:bionic
+image: coqci/base:V2018-05-05
 
 stages:
-  - opam-boot
   - build
   - test
 
 # some default values
 variables:
-  # Format: $IMAGE-V$DATE-$HOUR-$MINUTE
-  CACHEKEY: bionic-V2018-04-29-00-50
-  DEBIAN_FRONTEND: "noninteractive"
-  NJOBS: "2"
-  COMPILER: "4.02.3"
-  CAMLP5_VER: "6.14"
-  OPAMROOT: "$CI_PROJECT_DIR/.opamcache"
-  OPAMROOTISOK: "true"
-
-  # some useful values
-  COMPILER_32BIT: "4.02.3+32bit"
-
-  COMPILER_BLEEDING_EDGE: "4.06.1"
-  CAMLP5_VER_BLEEDING_EDGE: "7.05"
-
-  TIMING_PACKAGES: "time python3"
-
-  COQIDE_PACKAGES: "libgtk2.0-dev libgtksourceview2.0-dev"
-  #COQIDE_PACKAGES_32BIT: "libgtk2.0-dev:i386 libgtksourceview2.0-dev:i386"
-  COQIDE_OPAM: "lablgtk.2.18.5 conf-gtksourceview.2"
-  COQIDE_OPAM_BE: "lablgtk.2.18.6 conf-gtksourceview.2"
-  COQDOC_PACKAGES: "texlive-latex-extra texlive-fonts-recommended hevea python3-sphinx python3-pexpect python3-sphinx-rtd-theme python3-bs4 python3-sphinxcontrib.bibtex python3-pip"
-  SPHINX_PACKAGES: "antlr4-python3-runtime"
-  ELPI_OPAM: "elpi"
+  # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
+  # for reference]
+  CACHEKEY: bionic_coq-V2018-05-03
+  # By default, jobs run in the base switch; override to select another switch
+  OPAM_SWITCH: "base"
+  # Used to select special compiler switches such as flambda, 32bits, etc...
+  OPAM_VARIANT: ""
 
 before_script:
   - cat /proc/{cpu,mem}info || true
   - ls -a # figure out if artifacts are around
-  - printenv
-#  - if [ "$COMPILER" = "$COMPILER_32BIT" ]; then sudo dpkg --add-architecture i386; fi
-  - apt-get update -qq && apt-get install -y -qq m4 opam ${EXTRA_PACKAGES}
-  # This should be replaced by standard debian packages once python3-antlr4 makes to the archive.
-  - if [ -n "${PIP_PACKAGES}" ]; then pip3 install ${PIP_PACKAGES}; fi
-  # if no cache running opam config fails!
-  - if [ -d .opamcache ]; then eval $(opam config env); fi
+  - printenv | sort
+  - declare -A switch_table
+  - switch_table=( ["base"]="$COMPILER" ["be"]="$COMPILER_BE" )
+  - opam switch -y "${switch_table[$OPAM_SWITCH]}$OPAM_VARIANT"
+  - eval $(opam config env)
+  - opam list
+  - opam config list
 
-################ OPAM SYSTEM ######################
-# - use cache between pipelines
-# - use artifacts between jobs
-#   (in https://gitlab.com/SkySkimmer/coq/-/jobs/63255417
-#    the cache wasn't available at the build step)
-# every non opam-boot job must set dependencies (for ci it's in the template)
-# otherwise all opam-boot artifacts are used together and we get some random switch
+################ GITLAB CACHING ######################
+# - use artifacts between jobs                       #
+######################################################
 
-# set cache key when using
-.opam-boot-template: &opam-boot-template
-  stage: opam-boot
-  artifacts:
-    name: "opam-$COMPILER"
-    paths:
-      - .opamcache
-    expire_in: 1 week
-  script:
-    - opam init -a -y -j $NJOBS --compiler=${COMPILER} default https://opam.ocaml.org
-    - eval $(opam config env)
-    - opam update
-    - opam config list
-    - opam list
-    - opam install -j ${NJOBS} -y camlp5.${CAMLP5_VER} ocamlfind num ${EXTRA_OPAM}
-    - rm -rf ~/.opam/log/
-    - opam list
-
-# TODO figure out how to build doc for installed coq
-# set dependencies when using
+# TODO figure out how to build doc for installed Coq
 .build-template: &build-template
   stage: build
   artifacts:
@@ -80,21 +39,16 @@ before_script:
       - config/Makefile
       - test-suite/misc/universes/all_stdlib.v
     expire_in: 1 week
-  dependencies:
-    - not-a-real-job
   script:
     - set -e
-    - printenv
-    - opam config list
-    - opam list
 
     - echo 'start:coq.config'
     - ./configure -prefix "$(pwd)/_install_ci" ${EXTRA_CONF}
     - echo 'end:coq.config'
 
     - echo 'start:coq.build'
-    - make -j ${NJOBS} byte
-    - make -j ${NJOBS}
+    - make -j "$NJOBS" byte
+    - make -j "$NJOBS"
     - make test-suite/misc/universes/all_stdlib.v
     - echo 'end:coq:build'
 
@@ -106,12 +60,9 @@ before_script:
 
     - set +e
 
-# set dependencies when using
 .warnings-template: &warnings-template
   # keep warnings in test stage so we can test things even when warnings occur
   stage: test
-  dependencies:
-    - not-a-real-job
   script:
     - set -e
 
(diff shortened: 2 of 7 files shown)


ejgallego:
Seems to work fine and the time savings are considerable; I think we should enable the automatic building and deploy both on travis and gitlab.

ejgallego:
Time savings example:
- https://gitlab.com/coq/coq/-/jobs/65537826
- https://gitlab.com/ejgallego/coq/-/jobs/65538732

.travis.yml
```diff
@@ -43,7 +43,7 @@ env:
   - CAMLP5_VER=".6.14"
   - CAMLP5_VER_BE=".7.03"
   - FINDLIB_VER=".1.4.1"
-  - FINDLIB_VER_BE=".1.7.3"
+  - FINDLIB_VER_BE=".1.8.0"
```
SkySkimmer: Why are you changing this in this PR?
It breaks travis btw.

dev/ci/ci-compcert.sh
```diff
@@ -5,7 +5,6 @@ ci_dir="$(dirname "$0")"
 
 CompCert_CI_DIR="${CI_BUILD_DIR}/CompCert"
 
-opam install -j "$NJOBS" -y menhir
```
SkySkimmer: Travis needs to install this.

.circleci/config.yml
```diff
@@ -8,83 +8,29 @@ defaults:
...
-    CAMLP5_VER: "6.14"
     NATIVE_COMP: "yes"
 
-    # some useful values
-    TIMING_PACKAGES: &timing-packages "time python"
-
 version: 2
 
 before_script: &before_script
   name: Install system packages
```
SkySkimmer: This name should be changed to "opam switch" or some such thing.

SkySkimmer (dismissed):
Cool but has some bugs.

Zimmi48 (commented):
Nice.

.travis.yml
ejgallego: 1.8.0 is installed by the docker image, so this is in preparation of switching Travis to it.

dev/ci/ci-compcert.sh
ejgallego: Ditto, will switch to the image.

.circleci/config.yml
ejgallego: As you wish.

.circleci/config.yml
```diff
@@ -8,83 +8,29 @@ defaults:
...
     source ~/.profile
-    printenv
-    if [ -n "${EXTRA_PACKAGES}" ]; then sudo apt-get update -yq && sudo apt-get install -yq --no-install-recommends ${EXTRA_PACKAGES}; fi
     echo . ~/.profile >> $BASH_ENV
-
-opam-switch: &opam-switch
-  name: Select opam switch
-  command: |
+    printenv
     opam switch ${COMPILER}
```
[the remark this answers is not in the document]
ejgallego: COMPILER is set by the Docker image, however my 1996-like upload speed means that this ran before the new image was in, re-running.

ejgallego:
Running with new image.

ejgallego:
Ok, so modulo a pass adding image versioning and documentation I consider this PR close to be ready. Travis work is deferred to a further PR, as well as autobuild; IMHO, the speed gains are worth getting this without that those features.

ejgallego:
Gains in circle are modest as the apt setup there was fairly minimal and the cache did the rest; however in Gitlab we had to install quite heave TeX packages and this helps.

Zimmi48:
So would this setup be much less useful if we were to tackle the LaTeX dependencies instead?

ejgallego:
> So would this setup be much less useful if we were to tackle the LaTeX dependencies instead?

I don't think so.

ejgallego:
Automated run being done at https://hub.docker.com/r/ejgallego/coq/builds/

I tried to create Coq docker organization but unfortunately docker requires 4 letters for an organization name, and no dashes, non-alpha characters. I took `coqci` for now, but yeah, we need to think for a good name [or contact docker to ask them to lift the restriction for us].

At least at the beginning, it looks to me that it would be more convenient to store the docker file in a separate repository for easier autobuild integration [tags in the repo do provide image version numbers], but we could also add a `.docker` dir to refine the automated build process. I dunno.

ejgallego:
Ok, I consider this ready; we need to setup a better automatic build system but the current manual approach should work for now while we decide how to version images.

ejgallego:
Note the latency for the image build.

SkySkimmer:
So what's the process for changes which need to change the image? eg adding some new development which needs some opam package.

ejgallego:
> So what's the process for changes which need to change the image? eg adding some new development which needs some opam package.

You tweak the docker file and upload it somewhere, quoting the readme:
```
- Build the image `docker build -t coqci:$VERSION .`

To upload to your hub:

- Create a hub.docker.com account.
- Login into your space `docker login --username=$USER`
- Push the image:
  + `docker tag coq_dev $USER/coqci`
  + `docker push $USER/coqci`
```
Then you need to modify the CI files.

It is even better to support autobuilding; in fact I am autobuilding, see: https://hub.docker.com/r/ejgallego/coq/builds/

The autobuild will add a version; version can be derived from many places, but indeed we need to discuss that and how to trigger the autobuild. IMHO for now we can live with manually setting the autobuild.

ejgallego:
I didn't figure out yet a good solution, a possibility is to have a separate repository for the Docker file and trigger the build from there, but I prefer to think a bit more to see what ideas we can come up with.

ejgallego:
All of this is very new for me and at least my brain needs time to process...

ejgallego:
Also gitlab has its per-repo registry.

ejgallego:
I think I'm gonna make the CI files to point to the `coqci` repository instead of my personal `ejgallego` one, and add to the hook a push to that organization repository so any admin on the docker org can update the image.

But that's a very minor tweak, so the rest of the patch can be reviewed.

.circleci/config.yml
SkySkimmer: ping

SkySkimmer: I mean ping

SkySkimmer (changes requested):

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN apt-get update -qq && apt-get install -y -qq m4 wget time gcc-multilib opam \
+        libgtk2.0-dev libgtksourceview2.0-dev \
+        texlive-latex-extra texlive-fonts-recommended hevea \
+        python3-sphinx python3-pexpect python3-sphinx-rtd-theme python3-bs4 python3-sphinxcontrib.bibtex python3-pip
+
+RUN pip3 install antlr4-python3-runtime
+
+# Basic OPAM setup
+ENV NJOBS="2" \
+    OPAMROOT=/root/.opamcache \
```
SkySkimmer: There's no reason to use a nonstandard opam root with docker (the reason was that gitlab doesn't let us cache stuff outside the repository directory).

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+    OPAMROOTISOK="true"
+
+ENV COMPILER="4.02.3" \
+    BASE_OPAM="num ocamlfind jbuilder"
+
+RUN opam init -a -y -j $NJOBS --compiler=$COMPILER default https://opam.ocaml.org && eval $(opam config env) && opam update
```
SkySkimmer: Is opam update useful immediately after init?

dev/ci/docker/bionic_coq/hooks/post_push
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
SkySkimmer: Is this a git hook or some other meaning of hook?

dev/ci/docker/bionic_coq/Dockerfile
ejgallego: IMHO I like we make the directory explicit.

dev/ci/docker/bionic_coq/Dockerfile
ejgallego: I guess it is as long as they are in the same run.

dev/ci/docker/bionic_coq/hooks/post_push
ejgallego: This is a `hub.docker.com` hook, see the README for the link to the docs. I'll do a pass tho.

ejgallego:
I have updated the docs, regarding the two comments I'd prefer to keep things as they are now.

SkySkimmer:
>I have updated the docs, regarding the two comments I'd prefer to keep things as they are now.

What about the circle "update system packages" job name?

ejgallego:
> What about the circle "update system packages" job name?

Indeed, fixed.

SkySkimmer (approved):
Just waiting for CI.

SkySkimmer:
In 6ac3b1bf22257ad1aceee9698b8333ded50a17fb based on this PR I experiment with the gitlab registry.
The interesting changes are at the top until the new docker-boot job, the rest is just disabling stuff to speed up testing.
A `docker-boot` job is added which builds the docker image and pushes it to the registry, the other jobs running on the image from the registry.
The boot job reuses the already pushed image if it exists and matches, this is important to cut down its runtime.

Making a new image takes 30min for all the switches https://gitlab.com/SkySkimmer/coq/-/jobs/66650546
For testing I removed all jobs except boot and main build, and after that run I built only the standard switch.
Building 1 switch takes around 20min https://gitlab.com/SkySkimmer/coq/-/jobs/66656942
When the registry is up to date it takes 2min https://gitlab.com/SkySkimmer/coq/-/jobs/66658790 (I don't know about all switches but probably no more than 5min)

Each pipeline pushes the image to `$CI_REGISTRY_IMAGE:$CI_PIPELINE_ID` which is eg `registry.gitlab.com/skyskimmer/coq:21557176` which is what the other jobs use to run themselves.
For caching it pulls from and pushes to a constant name, in this test `$CI_REGISTRY_IMAGE:bionic-test`. I don't know if this should be versioned. We might also want to pull from the main coq repo to avoid redoing work.

The question of having 1 image or 1 image/switch remains open.

Using the gitlab registry doesn't really work for circle CI, so the dockerhub account would have to remain (or circle could return to an opam-boot job, or be removed from our CI). There are similar concerns with travis if we want to move it to docker.

.circleci/config.yml
```diff
@@ -8,83 +8,29 @@ defaults:
...
     docker:
-      - image: ocaml/opam:ubuntu
+      - image: coqci/base:V2018-05-03
 
   environment: &envvars
-    # required by some of the targets, e.g. compcert, passed for
-    # instance to opam to configure the number of parallel jobs
-    # allowed
-    NJOBS: 2
-    COMPILER: "system"
```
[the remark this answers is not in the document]
ejgallego: Yup, it does, why it should not? Compiler is set by the image and then circle makes sure we are always on the proper switch.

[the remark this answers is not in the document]
SkySkimmer: >Yup, it does, why it should not?

Didn't it break on gitlab? The yml still sets compiler/etc there.

ejgallego:
> In 6ac3b1b based on this PR I experiment with the gitlab registry.

That's cool!

> The question of having 1 image or 1 image/switch remains open.

IMHO I would use one image for now, it is faster and simpler. Note also that gitlab/circle have limited amount for caching images, IIUC; so having one image makes it more likely that it is on the cache as the pressure is lower.

The issue of sync seems to remain tho?

Why doesn't the gitlab registry work for other projects? Cannot it be made public?

SkySkimmer:
WDYM by sync exactly?

SkySkimmer:
That's only for caching though. When we change the image there may be some time during which different commits alternatively push their version to that tag slowing things down but semantically it's transparent.

.gitlab-ci.yml
```diff
@@ -1,76 +1,33 @@
...
 variables:
-  # Format: $IMAGE-V$DATE-$HOUR-$MINUTE
-  CACHEKEY: bionic-V2018-04-29-00-50
-  DEBIAN_FRONTEND: "noninteractive"
-  NJOBS: "2"
+  # Format: $IMAGE-V$DATE
+  CACHEKEY: bionic_coq-V2018-05-03
+  # This shouldn't be needed here, but for some reason they are not
+  # properly picked from the docker image.
   COMPILER: "4.02.3"
```
SkySkimmer: Why do we still set this here but not in circle?

ejgallego: I thought about this, but is it possible to set that in circle altogether? The `save-cache` entry is gone, so I am not sure we can do this.

I prefer to have this info here for now as to document wha'ts going on in case the cache resurrects in the future.

On the other hand, we need to discuss the future or CircleCI.

ejgallego:
> That's only for caching though. When we change the image there may be some time during which different commits alternatively push their version to that tag slowing things down but semantically it's transparent.

I need to understand it better, I guess you'll open a PR once you are happy with it, right?

SkySkimmer:
I mean COMPILER not CACHEKEY

Gaëtan Gilbert

SkySkimmer:
>I need to understand it better, I guess you'll open a PR once you are happy with it, right?

Once this one is in essentially.

ejgallego:
> I mean COMPILER not CACHEKEY

Ah oh, we discussed that above [or in gitter, I don't recall]

It is such that the `variables` entry in Gitlab's templates [`.ci-template` for example] is properly set; when the `.yml` is processed, the variables coming from the image are still not in scope, so that would result in `OPAM_SWÍTCH` being set to empty.

In Circle, `OPAM_SWITCH` doesn't depend on the job so we don't need this hack.

Another possibility would be to set `OPAM_SWITCH` to a symbolic value `{base,base32,be,be+flambda}` and have a switch statement in `before_script`.

In fact I like this.

ejgallego:
Ok, I've done some updates, use `git diff -u a841b2250..7bbc1dca0` to review.

ejgallego:
Need an image build tho, so we need to wait around one hour for the autobuild to complete.

ejgallego:
Yet another update:

`git diff -u a841b2250..16eb6738c`

ejgallego:
Please review.

SkySkimmer:
I don't have your random commits, is this just the OPAM_SWITCH stuff?

Gaëtan Gilbert

Zimmi48:
You can't get them locally of course, but you can use GitHub web diff interface to see them: https://github.com/coq/coq/compare/a841b2250...16eb6738c

ejgallego:
> I don't have your random commits, is this just the OPAM_SWITCH stuff?

Yup; I thought that would have worked if you do first a `git fetch --all` [and you have the ejgallego remote]

I was trying to improve the review process for force push, that as you noted is not easy.

ejgallego:
> You can't get them locally of course, but you can use GitHub web diff interface to see them: a841b22...16eb673

Oh cool Théo! I'll update my workload to use this instead.

IMHO indeed it is useful to help reviewers to follow force pushes. I may even include them in the first issue in the future.

ejgallego:
> You can't get them locally of course, but you can use GitHub web diff interface to see them: a841b22...16eb673

Oh, the diff shown there is not correct I think. I see this on my local terminal:
```diff
egallego@rochefort:~/external/coq-master$ git diff -u a841b2250..16eb6738c | cat
diff --git a/.circleci/config.yml b/.circleci/config.yml
index f6a0d44f3..07b728edf 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ defaults:
     # reference syntax)
     working_directory: ~/coq
     docker:
-      - image: coqci/base:V2018-05-03
+      - image: coqci/base:V2018-05-05
 
   environment: &envvars
     NATIVE_COMP: "yes"
@@ -16,12 +16,12 @@ defaults:
 version: 2
 
 before_script: &before_script
-  name: Install system packages
+  name: Setup OPAM Switch
   command: |
     echo export TERM=xterm >> ~/.profile
     source ~/.profile
     echo . ~/.profile >> $BASH_ENV
-    printenv
+    printenv | sort
     opam switch "$COMPILER"
     opam config list
     opam list
diff --git a/.gitlab-ci.yml b/.gitlab-ci.yml
index 74a8b2550..f73d6caae 100644
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,4 +1,4 @@
-image: coqci/base:V2018-05-03
+image: coqci/base:V2018-05-05
 
 stages:
   - build
@@ -6,19 +6,22 @@ stages:
 
 # some default values
 variables:
-  # Format: $IMAGE-V$DATE
+  # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
+  # for reference]
   CACHEKEY: bionic_coq-V2018-05-03
-  # This shouldn't be needed here, but for some reason they are not
-  # properly picked from the docker image.
-  COMPILER: "4.02.3"
-  COMPILER_32BIT: "4.02.3+32bit"
-  COMPILER_BE: "4.06.1"
+  # By default, jobs run in the base switch; override to select another switch
+  OPAM_SWITCH: "base"
+  # Used to select special compiler switches such as flambda, 32bits, etc...
+  OPAM_VARIANT: ""
 
 before_script:
   - cat /proc/{cpu,mem}info || true
   - ls -a # figure out if artifacts are around
   - printenv | sort
-  - opam switch -y "$OPAM_SWITCH"
+  - declare -A switch_table
+  - switch_table=( ["base"]="$COMPILER" \
+                   ["be"]="$COMPILER_BE"
+  - opam switch -y "${switch_table[$OPAM_SWITCH]}$OPAM_VARIANT"
   - eval $(opam config env)
   - opam list
   - opam config list
@@ -121,32 +124,28 @@ before_script:
   dependencies:
     - build
   variables: &ci-template-vars
-    OPAM_SWITCH: "$COMPILER"
     TEST_TARGET: "$CI_JOB_NAME"
 
 build:
   <<: *build-template
   variables:
-    OPAM_SWITCH: "$COMPILER"
     EXTRA_CONF: "-native-compiler yes -coqide opt -with-doc yes"
 
 # no coqide for 32bit: libgtk installation problems
 build:32bit:
   <<: *build-template
   variables:
-    OPAM_SWITCH: "$COMPILER_32BIT"
+    OPAM_VARIANT: "+32bit"
     EXTRA_CONF: "-native-compiler yes"
 
 build:bleeding-edge:
   <<: *build-template
   variables:
-    OPAM_SWITCH: "$COMPILER_BE"
+    OPAM_SWITCH: be
     EXTRA_CONF: "-native-compiler yes -coqide opt"
 
 warnings:
   <<: *warnings-template
-  variables:
-    OPAM_SWITCH: "$COMPILER"
 
 # warnings:32bit:
 #   <<: *warnings-template
@@ -156,42 +155,38 @@ warnings:
 warnings:bleeding-edge:
   <<: *warnings-template
   variables:
-    OPAM_SWITCH: "$COMPILER_BE"
+    OPAM_SWITCH: be
 
 test-suite:
   <<: *test-suite-template
   dependencies:
     - build
-  variables:
-    OPAM_SWITCH: "$COMPILER"
 
 test-suite:32bit:
   <<: *test-suite-template
   dependencies:
     - build:32bit
   variables:
-    OPAM_SWITCH: "$COMPILER_32BIT"
+    OPAM_VARIANT: "+32bit"
 
 test-suite:bleeding-edge:
   <<: *test-suite-template
   dependencies:
     - build:bleeding-edge
   variables:
-    OPAM_SWITCH: "$COMPILER_BE"
+    OPAM_SWITCH: be
 
 validate:
   <<: *validate-template
   dependencies:
     - build
-  variables:
-    OPAM_SWITCH: "$COMPILER"
 
 validate:32bit:
   <<: *validate-template
   dependencies:
     - build:32bit
   variables:
-    OPAM_SWITCH: "$COMPILER_32BIT"
+    OPAM_VARIANT: "+32bit"
 
 ci-bignums:
   <<: *ci-template
diff --git a/dev/ci/docker/bionic_coq/Dockerfile b/dev/ci/docker/bionic_coq/Dockerfile
index c0974ef28..5698175c5 100644
--- a/dev/ci/docker/bionic_coq/Dockerfile
+++ b/dev/ci/docker/bionic_coq/Dockerfile
@@ -18,20 +18,18 @@ ENV NJOBS="2" \
 ENV COMPILER="4.02.3" \
     BASE_OPAM="num ocamlfind jbuilder"
 
-RUN opam init -a -y -j $NJOBS --compiler=$COMPILER default https://opam.ocaml.org && eval $(opam config env) && opam update
+RUN opam init -a -y -j $NJOBS --compiler="$COMPILER" default https://opam.ocaml.org && eval $(opam config env) && opam update
 
-# Setup of the regular switch
+# Setup of the base switch
 ENV CAMLP5_VER="6.14" \
     COQIDE_OPAM="lablgtk.2.18.5 conf-gtksourceview.2" \
     CI_OPAM="menhir elpi ocamlgraph"
 
-RUN opam switch -y -j $NJOBS $COMPILER && eval $(opam config env) && \
+RUN opam switch -y -j $NJOBS "$COMPILER" && eval $(opam config env) && \
     opam install -j $NJOBS $BASE_OPAM camlp5.$CAMLP5_VER $COQIDE_OPAM $CI_OPAM
 
-# 32bit switch
-ENV COMPILER_32BIT="4.02.3+32bit"
-
-RUN opam switch -y -j $NJOBS $COMPILER_32BIT && eval $(opam config env) && \
+# base+32bit switch
+RUN opam switch -y -j $NJOBS "${COMPILER}+32bit" && eval $(opam config env) && \
     opam install -j $NJOBS $BASE_OPAM camlp5.$CAMLP5_VER
 
 # BE switch
@@ -42,6 +40,6 @@ ENV COMPILER_BE="4.06.1" \
 RUN opam switch -y -j $NJOBS $COMPILER_BE && eval $(opam config env) && \
     opam install -j $NJOBS $BASE_OPAM camlp5.$CAMLP5_VER_BE $COQIDE_OPAM_BE
 
-# Flambda switch
+# BE+flambda switch
 RUN opam switch -y -j $NJOBS "${COMPILER_BE}+flambda" && eval $(opam config env) && \
     opam install -j $NJOBS $BASE_OPAM camlp5.$CAMLP5_VER_BE $COQIDE_OPAM_BE
diff --git a/dev/ci/docker/bionic_coq/hooks/post_push b/dev/ci/docker/bionic_coq/hooks/post_push
index 6015c9b43..008ef1af3 100755
--- a/dev/ci/docker/bionic_coq/hooks/post_push
+++ b/dev/ci/docker/bionic_coq/hooks/post_push
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
-docker tag $IMAGE_NAME $DOCKER_REPO:V2018-05-03
-docker push $DOCKER_REPO:V2018-05-03
+COQCI_VERSION=V2018-05-05
+docker tag $IMAGE_NAME $DOCKER_REPO:$COQCI_VERSION
+docker push $DOCKER_REPO:$COQCI_VERSION
 
-docker tag $IMAGE_NAME coqci/base:V2018-05-03
-docker push coqci/base:V2018-05-03
+docker tag $IMAGE_NAME coqci/base:$COQCI_VERSION
+docker push coqci/base:$COQCI_VERSION
```

ejgallego:
> I don't have your random commits, is this just the OPAM_SWITCH stuff?

Ah, it is also easy to do `git fetch ejgallego $COMMIT`, so I think the diff should work.

At least myself, I do have all the main developers remotes added.

ejgallego:
Ok, new image pushed; also note that this should fix the recent compcert failure on Gitlab [cc: @Zimmi48 ] , IMHO it is more efficient to get this in that to do a temporal fixup.

Zimmi48:
I thought that you couldn't fetch single dangling commits using git.

ejgallego:
> I thought that you couldn't fetch single dangling commits using git.

Well, I tried here and it worked (TM) , but I have no idea whether this will work in general:
```
egallego@rochefort:~/external/coq-master$ git fetch ejgallego af19d08003173718c0f7c070d0021ad958fbe58d
De github.com:ejgallego/coq
 * branch                af19d08003173718c0f7c070d0021ad958fbe58d -> FETCH_HEAD
```

Zimmi48:
IIUC compcert was failing on GitLab CI and this PR fixes it. Either we choose to backport it to `v8.8` or we prepare a special fix for this in `v8.8`. We can also wait for a few days before doing anything to see how thing go with GitLab CI.

ejgallego:
Oh, good point; I think this is OK to backport to 8.8 as the version requirements are so far the same, and it doesn't change any code of Coq itself.

PR #7397, opened 2018-04-30
[ci] Fix #7396: VST is broken

### LINKED ISSUE #7396: VST script broken
We make .loadpath but this doesn't exist anymore.
https://github.com/PrincetonUniversity/VST/commit/90e164d01f74b2ad1aaa348379e8dbed9989154b

ejgallego:
This is due to our CI script relying on their makefile internals,
unfortunately we still have to do this to avoid timeouts.

### DIFF
--- a/dev/ci/ci-vst.sh
+++ b/dev/ci/ci-vst.sh
@@ -8,6 +8,10 @@ VST_CI_DIR="${CI_BUILD_DIR}/VST"
 # opam install -j ${NJOBS} -y menhir
 git_checkout "${VST_CI_BRANCH}" "${VST_CI_GITURL}" "${VST_CI_DIR}"
 
-# Targets are: msl veric floyd progs , we remove progs to save time
-# Patch to avoid the upper version limit
-( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true .loadpath version.vo msl veric floyd )
+# HACK: from the upstream makefile:
+#
+# default_target: _CoqProject version.vo msl veric floyd progs
+#
+# We have to omit progs as otherwise we timeout on Travis; once we
+# move to Gitlab we will able to just use `make`
+( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true _CoqProject version.vo msl veric floyd )


dev/ci/ci-vst.sh
```diff
@@ -8,6 +8,10 @@ VST_CI_DIR="${CI_BUILD_DIR}/VST"
...
 
-# Targets are: msl veric floyd progs , we remove progs to save time
-# Patch to avoid the upper version limit
-( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true .loadpath version.vo msl veric floyd )
+# HACK: from the upstream makefile:
+#
+# default_target: _CoqProject version.vo msl veric floyd progs
+#
+# We have to omit progs as otherwise we timeout on Travis; once we
+# move to Gitlab we will able to just use `make`
```
Zimmi48: Unfortunately, that would mean also removing Travis support, which is easier to run for our contributors.

Zimmi48 (approved):
Yes as a quick solution. But what about asking them to provide a real target for this? Say `lite`.

dev/ci/ci-vst.sh
SkySkimmer: >Unfortunately, that would mean also removing Travis support, which is easier to run for our contributors.

I think we will end up with a ci-vst-lite.sh and a ci-vst.sh for travis/gitlab.

dev/ci/ci-vst.sh
```diff
@@ -8,6 +8,10 @@ VST_CI_DIR="${CI_BUILD_DIR}/VST"
...
+( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true _CoqProject version.vo msl veric floyd )
```
SkySkimmer: Maybe we should use `make -o progs` (assume-old) since that's closer to what we mean.
I tested with a dummy makefile and it worked as I expected.

SkySkimmer (approved):

dev/ci/ci-vst.sh
ejgallego: > Unfortunately, that would mean also removing Travis support, which is easier to run for our contributors.

Realistically Travis support will be phased out at some point, Gitlab is pretty to run by the way; they can just clone the main Coq repos there and hit "mirror".

PR #7400, opened 2018-05-01
ci-vst.sh: use -o progs

SkySkimmer:
This is closer to what we mean than reproducing the default target
without progs.

### DIFF
--- a/dev/ci/ci-vst.sh
+++ b/dev/ci/ci-vst.sh
@@ -8,10 +8,6 @@ VST_CI_DIR="${CI_BUILD_DIR}/VST"
 # opam install -j ${NJOBS} -y menhir
 git_checkout "${VST_CI_BRANCH}" "${VST_CI_GITURL}" "${VST_CI_DIR}"
 
-# HACK: from the upstream makefile:
-#
-# default_target: _CoqProject version.vo msl veric floyd progs
-#
-# We have to omit progs as otherwise we timeout on Travis; once we
-# move to Gitlab we will able to just use `make`
-( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true _CoqProject version.vo msl veric floyd )
+# We have to omit progs as otherwise we timeout on Travis; on Gitlab
+# we will be able to just use `make`
+( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true -o progs )


ejgallego:
I am OK with that but it still relies on an "internal" name, so breakage is possible.

dev/ci/ci-vst.sh
```diff
@@ -8,10 +8,6 @@ VST_CI_DIR="${CI_BUILD_DIR}/VST"
...
-# HACK: from the upstream makefile:
-#
-# default_target: _CoqProject version.vo msl veric floyd progs
-#
-# We have to omit progs as otherwise we timeout on Travis; once we
-# move to Gitlab we will able to just use `make`
-( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true _CoqProject version.vo msl veric floyd )
+# We have to omit progs as otherwise we timeout on Travis; on Gitlab
+# we will able to just use `make`
+( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true -o progs )
```
ejgallego: Please improve/keep the old documentation, for the casual reader it may be clear what is going on.

SkySkimmer: It says "we omit progs", that's as clear as it gets.

ejgallego: Something such as:
```
The VST default is as of today:
default_target: ....
Due to build time constraints on Travis we omit the `progs` target, etc...
```

SkySkimmer:
I think that's not a useful addition.


Gaëtan Gilbert

dev/ci/ci-vst.sh
Zimmi48: I think the comment is clear enough like this. However, it contains a typo: "we will *be* able to"...

ejgallego:
I'd prefer the documentation to explain the issue in full detail.

SkySkimmer:
It is in full detail.

Gaëtan Gilbert

ejgallego:
No, it is not satisfactory to me. Even after reading that I'd have to look at the VST makefile to understand what is going on, so I'd like to see more verbose documentation [that never hurt anyone BTW].

ejgallego:
So you don't plan to amend this?

ejgallego:
> So you don't plan to amend this?

:/

SkySkimmer:
I was going to but procrastination is too strong.

Gaëtan Gilbert

PR #7402, opened 2018-05-02
[ci]: add pidetop (fix #7336)

### LINKED ISSUE #7336: add pidetop to CI
that should also help #6859

gares:
We test pidetop compiles, installs and can be run by coq

### DIFF
--- a/dev/ci/ci-pidetop.sh
+++ b/dev/ci/ci-pidetop.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+# $0 is not the safest way, but...
+ci_dir="$(dirname "$0")"
+. "${ci_dir}/ci-common.sh"
+
+pidetop_CI_DIR="${CI_BUILD_DIR}/pidetop"
+
+git_checkout "${pidetop_CI_BRANCH}" "${pidetop_CI_GITURL}" "${pidetop_CI_DIR}"
+
+( cd "${pidetop_CI_DIR}" && coq_makefile -f Make -o Makefile.top && make -f Makefile.top "-j${NJOBS}" && make install-toploop -f Makefile.top )
+
+echo -en '4\nexit' | coqtop -main-channel stdfds -toploop pidetop
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,6 +209,9 @@ jobs:
   mtac2:
     <<: *ci-template
 
+  pidetop:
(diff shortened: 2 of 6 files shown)


dev/ci/ci-pidetop.sh
```diff
@@ -0,0 +1,16 @@
...
+pidetop_CI_DIR="${CI_BUILD_DIR}/pidetop"
+
+git_checkout "${pidetop_CI_BRANCH}" "${pidetop_CI_GITURL}" "${pidetop_CI_DIR}"
+
+( cd "${pidetop_CI_DIR}" && coq_makefile -f Make -o Makefile.top && make -f Makefile.top "-j${NJOBS}" && make install-toploop -f Makefile.top )
+
+# Test the toploop can be loaded by Coq (grep init detects if it was loaded,
+# the || true is there because 'xxx' is not a valid message and we run with -o
+# pipefail)
+echo xxx | (coqtop -main-channel stdfds -toploop pidetop || true) | grep init
```
ejgallego: I am not sure if I like this test, WDYT @SkySkimmer ?

gares: It may help to know what is wrong with it, in your opinion.

SkySkimmer: Is there some reason we shouldn't use a valid message?

SkySkimmer (approved):
I would prefer to use a valid message for the test if possible.

dev/ci/ci-pidetop.sh
gares: I can try to see if there is an equivalent of `Quit.` (now that I think at it, there should be a `Terminate` one).

ejgallego:
So what @SkySkimmer said.

gares:
The original protocol has no "exit", exiting is signaled by closing the channel (and not exit code is pecified AFAICT).
I've added a message "exit" that calls `exit 0`, so that we can tell if coq exited abruptly or because out message got the expected handler being called.

ejgallego (approved):
Looks good, will merge as soon as CI is ready.

ejgallego:
Note the backporting does need to update the branch.

Zimmi48:
There is no `v8.8` branch and this wouldn't work with the new `exit` message anyways, so I suppose we'll just test this in 8.9 and forwards.

ejgallego:
:+1:

Issue #6354, opened 2017-12-08
Install target generates many warnings when `-native-compiler=no`.


ejgallego:
#### Version
Coq 8.8+alpha [7b40908bfbc255d51384e88a73fa5d98380b237f]

#### Operating system
Linux / Ubuntu + OPAM / OCaml 4.06.0

#### Description of the problem

When doing `./configure -prefix=u -native-compiler=no && make && make install` , `master` will generate lots of warnings of the form:
```
install: cannot stat 'plugins/fourier/.coq-native/NCoq_fourier_Fourier_util.o': No such file or directory
```
I couldn't observe this in `v8.7`.

advorkin:
Hello,

We are having the same problem.    

Getting 910 cannot stat messages. Ex.
install: cannot stat 'theories/Reals/.coq-native/NCoq_Reals_DiscrR.o': No such file or directory

Using options -bytecode-compiler no -native-compiler yes

Thank you!

advorkin:
I'm having the same problem with 8.7.1 and -native-compiler yes.  8.7.0 worked...  I would appreciate your advice.  Thank you.

SkySkimmer:
Should be fixed by https://github.com/coq/coq/pull/7593

PR #7406, opened 2018-05-02
Fix #7214: install knows which ml files do not get compiled to cmx.

### LINKED ISSUE #7214: install complains because of a few not compiled ml files
eg in https://gitlab.com/SkySkimmer/coq/-/jobs/62215853
```
install: cannot stat 'clib/deque.cmi': No such file or directory
...
install: cannot stat 'clib/deque.cmx': No such file or directory
install: cannot stat 'plugins/extraction/big.cmx': No such file or directory
install: cannot stat 'toplevel/coqtop_byte_bin.cmx': No such file or directory
```
deque has never been used, not sure about the others.

SkySkimmer:
(we just delete the deque files since they're useless)


advorkin:
Hello,

I removed deque.* and modified Makefile.install per your instructions.  Unfortunately, same problem.

Getting 910 cannot stat messages.  Ex. 
install: cannot stat 'theories/Reals/.coq-native/NCoq_Reals_DiscrR.o': No such file or directory

Using options -bytecode-compiler no -native-compiler yes

Thank you!

SkySkimmer:
Then you want bug  #6354 ^^

advorkin:
Thank you for responding.  But that particular bug uses  -native-compiler=no option, thus I wasn't sure if it's the same problem.

ppedrot (approved):
I don't remember why I introduced Deque in the first place, it might have been an incomplete patch that got in somehow. Anyways, it's fine to remove it as it's not used.

SkySkimmer:
Can someone take charge of this PR? @ppedrot ?

ppedrot:
I do not really understand the Makefile part, but I can take it if needed.

ppedrot:
Maybe @gares or @ejgallego should have a look at this PR? They're more fluent in the build system than I am.

gares:
I don't get why the Makefile should be touched by this PR.

ppedrot:
@SkySkimmer Can you explain the Makefile part of the PR?

Makefile.install
```diff
@@ -103,7 +103,9 @@ INSTALLCMI = $(sort \
 	$(foreach lib,$(CORECMA), $(addsuffix .cmi,$($(lib:.cma=_MLLIB_DEPENDENCIES))))) \
 	$(PLUGINS:.cmo=.cmi)
 
-INSTALLCMX = $(sort $(filter-out checker/% ide/% tools/% dev/% configure.cmx, $(MLFILES:.ml=.cmx)))
+INSTALLCMX = $(sort $(filter-out checker/% ide/% tools/% dev/% \
+	configure.cmx toplevel/coqtop_byte_bin.cmx plugins/extraction/big.cmx, \
+	$(MLFILES:.ml=.cmx)))
```
SkySkimmer: MLFILES is gotten by `find . -name '*.ml'` and these specific files don't get compiled so we shouldn't try to install their cmx.

Zimmi48:
@gares Does this explanation satisfy you?

Makefile.install
gares: My question is: why removing the `.ml` files this PR is removing is not enough?

SkySkimmer:
Because coqtop_byte_bin is not removed

Gaëtan Gilbert

gares:
ok

PR #7135, opened 2018-03-31
Introducing an explicit `Declare Scope` command

herbelin:
**Kind:** feature / enhancement / partial redesign / wish

In some sense: "closes #7120".

This PR adds a command `Declare Scope id` to register a scope and displays a warning the first time a non-explicitly registered scope is referred to.

This answers a [suggestion](https://github.com/coq/coq/pull/6247#issuecomment-377202990) initially made in the discusion for #6247 and turned into a proper issue #7120 by @jashug.

As far as I could judge (hoping I did not miss a case), behavior of existing scripts should be unchanged up to the new output of a warning. As for the new command, it makes a scope visible at require time as it was before for scopes implicitly declared by either `Delimit`, `Undelimit`, `Bind` or `Notation`.

Part of the remaining issues are discussed at #7120. Wrt to @jashug's [comment](), the current PR deactivates warning at require time, except for primitive notation (more difficult).

To do:
- [X] Corresponding documentation was added / updated.
- [x] Entry added in CHANGES.

Possibly to do:
- [ ] unify the different definitions of a scope `pair_scope` (e.g. by moving the notations for `fst` and `snd` in `Datatypes.v`, also giving notations for `projT1` and `projT2` at the same time)?
- [ ] define a common scope for `CMorphisms` and `Morphisms`, as well as `CEquivalences` and `Equivalences`?
- [ ] unify names of scopes (with e.g. `_scope` convention) in standard library

PR #496, opened 2017-03-22
Numeral notation (revisited)

letouzey:
### What ? ###

```
Numeral Notation mytype myparse myprint : scope.
Numeral Notation mytype myparse myprint : scope (warning after 5000).
Numeral Notation mytype myparse myprint : scope (abstract after 5000).
```
where:
  - `mytype` should be an inductive type (think of `nat` for instance)
  - `myparse` should be a Coq function of type `... -> mytype`. It could also be partial : `... -> option mytype`. And the starting could be either `Decimal.int`, `Decimal.uint` (new signed or unsigned base-10 coding of integers) or `Z`.
  - `myprint` should be a Coq function of type `mytype -> ...`. it could be partial : `mytype -> option ...`. And the result could be either `Decimal.int`, `Decimal.uint` or `Z`.
  - No correlation between the types of the parser and printer.
  - *New*: the `(asbtract after ...)` option allows to leave large numbers in the form `myparser 123456` where 123456 is typically a `Decimal.uint`, ie. far less subject to stack overflow.

### How ? ###

  - Direct reuse of half of @roglo 's code for [PR#156](https://github.com/coq/coq/pull/156) (the one handling inductive type via conversion with respect to `Z`, not the Ltac part meant for Reals). See also [CEP-002](https://github.com/coq/ceps/blob/master/text/002-numeral-notation.md). Concerning the reals, I propose to wait first for [PR#415](https://github.com/coq/coq/pull/415) that will probably avoid the need for an Ltac machinery.
  - Added similar code for conversion with respect to `Decimal.uint` and `Decimal.int` (see proposed `Coq.Init.Decimal`).
  - Changed `Constrexpr.Numeral` : now a raw string of digits + a boolean sign instead of a bigint.
  - Numeral Notation via Decimal is fully bigint-free.
  - For now, light integration in Coq:
     - A new plugin `numeral_notation_plugin.cmxs` (autoloaded in Prelude).
     - I propose to use it for nat,Z,N,pos but it's easy to decide otherwise (see the last commits in this PR).
     - So the 3 solutions could be used almost simultaneously (historical code, numeral notation via Z, or via Decimal).

  - On the Coq side:
     - A new `Coq.Init.Decimal` file (not exported).
     - A few more functions in `Nat` or `BinPosDef` or `BinIntDef` (for instance `Nat.to_uint`, `Nat.to_int`, ...)
     - Some bijection (or quasi-bijection) proofs in  `Coq.Numbers.DecimalNat` and alii.

[...]

PR #7410, opened 2018-05-02
Splitting primitive numeral parser/printer for positive, N, Z into three files

herbelin:
**Kind:** redesign/cleanup

This is part of #7135 and intended to increase modularity in dealing with the primitive parsers for N, Z and positive. In particular, in a view where scopes are declared before declaring primitive parsers, we get a better modularity by successively, for each type, declaring the scope, loading the primitive parser/printer, declaring the delimiter, binding the scope to a type.

Risk of conflict with #496 though.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -143,8 +143,8 @@ BTAUTOCMO:=plugins/btauto/btauto_plugin.cmo
 RTAUTOCMO:=plugins/rtauto/rtauto_plugin.cmo
 NATSYNTAXCMO:=plugins/syntax/nat_syntax_plugin.cmo
 OTHERSYNTAXCMO:=$(addprefix plugins/syntax/, \
-        z_syntax_plugin.cmo \
-        r_syntax_plugin.cmo \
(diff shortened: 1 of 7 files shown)


ejgallego (commented):
I already noticed this in #7135 and indeed it does make sense; a question tho is if code can be factored among the files; in particular it is strange to need `make_dir` etc.. shouldn't be these provided by the kernel?

ppedrot (commented):
No strong opinion.

ejgallego (approved):
That's a welcome change and opens up the door to further improvements and refactorings.

Please, add an entry to `dev/doc/changes.md` as developers may need to be aware of this change.

herbelin:
OK, added a note to `dev/doc/changes.md`.

ejgallego:
I am gonna merge this pretty quick, this has been opened for 2 months so I think developers had reasonble time to react.

ejgallego:
The CI seems in sorry state but not the fault of this PR.

Issue #7413, opened 2018-05-03
coqdep: spurious warning with -f


gares:
```shell
gares@gargamel:~/COQ/coq/foo$ cat > _CoqProject
-I src 
-R theories Foo

src/a.ml
gares@gargamel:~/COQ/coq/foo$ touch src/a.ml
touch: cannot touch 'src/a.ml': No such file or directory
gares@gargamel:~/COQ/coq/foo$ mkdir src
gares@gargamel:~/COQ/coq/foo$ mkdir theories
gares@gargamel:~/COQ/coq/foo$ touch src/a.ml
gares@gargamel:~/COQ/coq/foo$ ../bin/coqdep -f _CoqProject 
*** Warning: a.ml already found in src (discarding src/a.ml)
gares@gargamel:~/COQ/coq/foo$ 
```

I suspect this is related to #6851

gares:
@SkySkimmer please assign it to me if it does not ring a bell.
IMO the path specified with `-I` is considered twice.

SkySkimmer:
It's the same as in 8.7 doing `coqdep src/a.ml -I src/` (which prints the same warning).
Possible fixes:
- have coqdep only read .v files from coqproject (this causes weird semantics for people using coqdep with -f and -c)
- don't print a warning when we see the same file twice from different sources. (so eg `coqdep src/a.ml src/a.ml` would not warn. We can bikeshed about the equivalence we use, eg should `coqdep src/a.ml "$PWD/src/a.ml"` warn, etc)

SkySkimmer:
>have coqdep only read .v files from coqproject

(to be specific I mean ignoring non-v files, but still reading -I etc)

SkySkimmer:
Anyways please pick one or 2 changes to do.

gares:
I tried to understand the code emitting the warning, that was added in 21a1c8abdd5807427849c34875c294cb6ad7d899 and refiend in 229dd629d2425750ec4e4ea5598804745979b6d2 .

To my understanding:
I list `a.ml` in my `_CoqProject` to tell `coq_makefile` to build it.
I add `-I whatever` to tell `coq_makefile` to pass it to `ocamlc` when compiling and `coqc` when compiling `.v` files that may `Declare ML Module`.
I expect all my `.ml` files be in paths listed with `-I`.
I may list more paths with `-I`, eg to use an external library, and such library may have `.cm*` files that are named like mines. So maybe this the is reason of existence of the warning?

[...]

PR #6859, opened 2018-02-28
[stm] Make toplevels standalone executables.

ejgallego:
We turn coqtop "plugins" into standalone executables, which will be
installed in `COQBIN` and located using the standard `PATH`
mechanism. Using dynamic linking for `coqtop` customization didn't
make a lot of sense, given that only one of such "plugins" could be
loaded at a time.  This cleans up some code and solves two problems:

- `coqtop` needing to locate plugins,
- dependency issues as plugins in `stm` depended on files in `toplevel`.

In order to implement this, we do some minor cleanup of the toplevel
API, making it functional, and implement uniform build rules. In
particular:

- `stm` and `toplevel` have become library-only directories,
- a new directory, `topbin`, contains the new executables,
- 4 new binaries have been introduced, for `coqidetop` and for the stm workers `coq{query,tactic,proof}worker`.

PR #7414, opened 2018-05-03
Add .byte targets for every bestocaml target

SkySkimmer:
This makes it easier to compile our executables for debug purposes.

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -240,6 +240,10 @@ $(OCAMLOPT) $(MLINCLUDES) $(OPTFLAGS) $(LINKMETADATA) -o $@ -linkpkg $(1) $^ && 
 $(OCAMLC) $(MLINCLUDES) $(BYTEFLAGS) $(CUSTOM) -o $@ -linkpkg $(1) $^)
 endef
 
+define ocamlbyte
+$(OCAMLC) $(MLINCLUDES) $(BYTEFLAGS) $(CUSTOM) -o $@ -linkpkg $(1) $^
+endef
+
 # Camlp5 settings
 
 CAMLP5DEPS:=grammar/grammar.cma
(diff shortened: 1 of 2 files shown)


Zimmi48:
I lack Makefile-reader power and I have always been a bit lost among all the developer-friendly targets we provide. Maybe this could be a good opportunity to start a small documentation about them? (In `dev/doc`.) If you don't have the will to do this, can you at least explain quickly how to use what this PR provides?

Zimmi48:
BTW, is it really useful to backport this PR given that most of the development today happens on master?

SkySkimmer:
This started when investigating #7413, I added the coqdep target, did `make bin/coqdep.byte` then ran ocamldebug on it.

Zimmi48:
I guess I see then. This might semantically "conflict" with #6859 BTW.

gares:
> This might semantically "conflict" with #6859 BTW.

I see conflicts, but why "semantic"?

gares (approved):

gares:
> Maybe this could be a good opportunity to start a small documentation about them

To my eyes this PR just adds `make bin/whatever.byte` targets, so nothing very unexpected.

ejgallego (commented):
Note that with Dune support coming soon, developers will almost always better served by using the builtin support for this kind of stuff it does provide.

So indeed we could consider whether we want to spend a lot of time tweaking the complex make setup.

Zimmi48:
> To my eyes this PR just adds make bin/whatever.byte targets, so nothing very unexpected.

Indeed.

> I see conflicts, but why "semantic"?

Well, it seems that all the new `.byte` targets are added manually. When #6859 is merged there will be more executable files, so more potential `.byte` which won't have been added. I'm saying all this with a very remote perspective though (as I said before, I don't understand and don't want to understand Makefile language. I'm very eager for Dune to be our main build system because it also means I won't have to understand that part of the code).

gares:
> I'm very eager for Dune to be our main build system because it also means I won't have to understand that part of the code

I guess you will have to understand dune.

I'm fine postponing the merge of this PR until #6859 is merged not to cause conflicts if @SkySkimmer is fine with it.

SkySkimmer:
>I'm fine postponing the merge of this PR until #6859 is merged not to cause conflicts if @SkySkimmer is fine with it.

I don't understand that PR but it seems to have had some issues getting merged.

Issue #7032, opened 2018-03-21
Official agreement from code owners required


Zimmi48:
It was decided yesterday to distribute the merging process (it used to be done by a single person).
The new process is described here: https://github.com/coq/coq/blob/master/dev/doc/MERGING.md
The maintainers are listed here: https://github.com/coq/coq/blob/master/.github/CODEOWNERS

Given that maintainers of different components are now responsible for merging pull requests (after following the process), it is necessary to give push access to more people.

@amahboubi @aspiwack @barras @bgregoir @ejgallego @fajb @forestjulien @gares @herbelin @JasonGross @letouzey @Matafou @mattam82 @maximedenes @ppedrot @silene @SkySkimmer @thery @vbgl @Zimmi48 and **Pierre Corbineau**: you are the listed code owners (did I forget anyone?). Have you read the merging documentation? Do you agree with the process? If yes, please post a comment here saying:

> I agree to be a code owner. I have read the merging documentation. I will follow the process and use my merging right wisely.

If not, please say so.

I will need to add all the people who agreed to be code owners as members of the new @coq/pushers team which will give them (if they hadn't already) the right to merge approved pull requests into master. For those who are not yet part of the @coq organization, this will also require accepting the invitation to join the organization on GitHub.

Zimmi48:
I agree to be a code owner. I have read the merging documentation. I will follow the process and use my merging right wisely.

Théo Zimmermann

thery:
I agree to be a code owner. I have read the merging documentation. I
will follow the process and use my merging right wisely.

ejgallego:
```coq
I agree to be a code owner. I have read the merging documentation. I will follow the process and use my merging right wisely.
```

ppedrot:
I agree to be a code owner. I have read the merging documentation. I will follow the process and use my merging right wisely. And my credit card number is not 5100 4057 7539 9735.

maximedenes:
> And my credit card number is not 5100 4057 7539 9735.

Just curious, what is not its expiration date?

ppedrot:
@maximedenes Definitely not 42/18.

Matafou:
I agree to be a code owner. I have read the merging documentation. I will follow the process and use my merging right wisely.

Could you make the documentation of the merging script a bit more precise please: under which branch should it be launched? A local branch with upstream set to official coq master (or any branch we want to merge in)?

bgregoir:
[...]

PR #7417, opened 2018-05-03
Micromega clean-up

maximedenes:
We add .mli files, removed dead code and use standard combinators
instead of redefined ad-hoc ones in a few places.

A lot of cleaning still has to be done on this code: documenting the
interfaces, resolving the many abstraction leaks. I suspect there is
still a lot of code duplication.

@Zimmi48 I'm not sure which "part" label I should use.


maximedenes:
@Zimmi48 I see that @fajb was not automatically asked a review as a code owner. Does it mean we should add him to the Coq organization?

Zimmi48:
@fajb never replied to the question whether he agreed to be a code owner in #7032. Unless you have more direct ways to reach him, we may need to find another principal maintainer for micromega.

maximedenes:
Did you drop him an e-mail?

Zimmi48:
No, I didn't. I only did it for the only person who didn't have a GitHub account already.

On your previous question (which "part" label), the answer is: it's always OK to create new labels.

fajb:
Hello, 

I have received the previous notifications...
For the "code owner" thing, I received the mail during vacations and, since then,  did not find time to read the instructions...
Sure, I'd rather like to keep an eye on the evolution of the code and be notified if it something happens.
If there is a bug report, I can also spend some time to fix it.
Yet, from my side, I do not intend to improve it. This is unfortunate as, I agree, there would be much cleaning to do.
(But also much to do to get eventually a decent efficient and complete decision procedure.)

Anyway, I had a quick look at your modifications. They are fine by me except, maybe, that printing code seems to have disappeared. Would it be useful for debugging purposes? Or is-it simply redundant with code existing elsewhere ?
Anyway, thanks for the maintenance.

Best,
PS: If you look at my commit history, it seems I could live without a commit right that I may only use once every few years.

maximedenes:
> They are fine by me except, maybe, that printing code seems to have disappeared. Would it be useful for debugging purposes?

Thanks for the feedback, @fajb ! A lot of the debugging printing code is already plugged in, under an `if debug` conditional. I didn't remove that part, only what was really dead code. Maybe you could tell me what printing functions look useful to you, we should bring them back and probably register them as printers for `ocamldebug` (I don't think they were).

fajb:
Keep it as it is.
If they were useful, time will tell!

Zimmi48:
Hello @fajb! Nice to know that you are correctly receiving the notifications. https://github.com/coq/coq/blob/master/dev/doc/MERGING.md shouldn't be too long to read but one important point is that this code owner / maintainer status does not require you to actively commit code, only to review and merge other people's changes. If you think that's doable, that would be great! Otherwise, your reviews are still very much appreciated.

maximedenes:
@SkySkimmer I'm pretty sure I ran `git rebase --whitespace=fix`. Isn't it supposed to fix all whitespace errors the linter can find?

Zimmi48:
Not the blank line problems at EOF which is what is reported here:

```
plugins/micromega/sos.ml:1112: new blank line at EOF.
```

maximedenes:
Ok, this is a dealbreaker for me. I don't want to do manual whitespace fixes, especially that I don't see any issue with one blank line. Will report an issue separately.

maximedenes:
Done in #7463.

maximedenes:
@Zimmi48 This PR is a bit special, as it was "morally" approved by @fajb, but we don't explicitly have a principal maintainer. Can you merge it? If it passes CI, of course.

Zimmi48:
I can. Although if @fajb doesn't want to assume the role of principal maintainer, we have to find someone else.

Zimmi48:
Can you rebase to have a recent CI build?

maximedenes:
> Can you rebase to have a recent CI build?

I rebased today. Isn't it recent enough?

> I can. Although if @fajb doesn't want to assume the role of principal maintainer, we have to find someone else.

Indeed, we should clarify that. But meanwhile, I'd prefer to see this non-invasive PR merged. Otherwise I'll close it, because I need to reduce my todo-list.

Zimmi48:
That's weird. It looks like the linter fix in #7464 did not actually fix the problem it was supposed to fix.
cc @SkySkimmer

SkySkimmer:
>That's weird. It looks like the linter fix in #7464 did not actually fix the problem it was supposed to fix.

Seems like somehow `blank-at-eol` is what complains about "new blank line at EOF".

Zimmi48:
Then it's going to complain also on Travis branch build after we merge, and in every subsequent linter test, right? In this case, I cannot merge as-is. `git` is not very helpful here: have a newline at EOF is actually helpful so we do not want to disable this test completely, but it is testing more and something that it can't fix automatically :disappointed:

SkySkimmer:
No this is pre-commit.

 >we do not want to disable this test completely

We already did in #7464 though.

Gaëtan Gilbert

Zimmi48:
Sorry, I didn't understand your reply.

SkySkimmer:
The part of the linter which errors is run on the diff for each commit in a PR, so future PRs won't be impacted.

>having a newline at EOF is actually helpful so we do not want to disable this test completely

>We already did in #7464 though.

My mistake, that test is not disabled and not related to the error, it's implemented by our custom script and not through git.

Zimmi48:
OK thanks for the info. I'll merge the PR now then.

Regarding why the linter fix didn't work, this is very strange. Git's documentation says "trailing-space being a short-hand to cover both blank-at-eol and blank-at-eof" which seems to confirm that your fix at #7464 was the right one. If we have this problem again in the future, we should investigate if it's not a bug in git.

plugins/micromega/csdpcert.mli
```diff
@@ -0,0 +1,9 @@
+(************************************************************************)
+(*         *   The Coq Proof Assistant / The Coq Development Team       *)
+(*  v      *   INRIA, CNRS and contributors - Copyright 1999-2018       *)
+(* <O___,, *       (see CREDITS file for the list of authors)           *)
+(*   \VV/  **************************************************************)
+(*    //   *    This file is distributed under the terms of the         *)
+(*         *     GNU Lesser General Public License Version 2.1          *)
+(*         *     (see LICENSE file for the text of the license)         *)
+(************************************************************************)
```
Zimmi48: What does it mean when a .ml exports nothing at all? Doesn't it mean that the .ml file could be removed completely?

SkySkimmer: It's only there for the side effects.

SkySkimmer: Or rather this file is compiled into a standalone executable.

Zimmi48: OK thanks.

PR #7419, opened 2018-05-03
Remove 100 occurrences of Evd.empty

maximedenes:
We address the easy ones, but they should probably be all removed.

Before:

```
git grep Evd.empty | wc -l
166
```

After:

```
git grep Evd.empty | wc -l
66
```

### DIFF
--- a/plugins/funind/functional_principles_proofs.ml
+++ b/plugins/funind/functional_principles_proofs.ml
@@ -230,7 +230,7 @@ let isAppConstruct ?(env=Global.env ()) sigma t =
   with Not_found -> false
 
 let nf_betaiotazeta = (* Reductionops.local_strong Reductionops.whd_betaiotazeta  *)
-  Reductionops.clos_norm_flags CClosure.betaiotazeta  Environ.empty_env Evd.empty
+  Reductionops.clos_norm_flags CClosure.betaiotazeta  Environ.empty_env @@ Evd.from_env Environ.empty_env
 
 
 exception NoChange
@@ -1099,10 +1099,12 @@ let prove_princ_for_struct (evd:Evd.evar_map ref) interactive_proof fun_num fnam
     let get_body const =
       match Global.body_of_constant const with
 	| Some (body, _) ->
+          let env = Global.env () in
+          let sigma = Evd.from_env env in
 	     Tacred.cbv_norm_flags
 	       (CClosure.RedFlags.mkflags [CClosure.RedFlags.fZETA])
-	       (Global.env ())
-	       (Evd.empty)
+               env
+               sigma
 	       (EConstr.of_constr body)
 	| None -> user_err Pp.(str "Cannot define a principle over an axiom ")
     in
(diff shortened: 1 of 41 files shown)


plugins/funind/invfun.ml
```diff
@@ -67,7 +67,7 @@ let observe_tac s tac g =
 let nf_zeta =
   Reductionops.clos_norm_flags  (CClosure.RedFlags.mkflags [CClosure.RedFlags.fZETA])
     Environ.empty_env
-    Evd.empty
+    (Evd.from_env Environ.empty_env)
```
SkySkimmer: I don't think this change does anything.
Well, I guess it makes it easier to find other Evd.empty.

gares (approved):
ssr + pretyping looks good to me

gares:
You were so close https://youtu.be/-hHsHGmoG0U?t=89

plugins/funind/invfun.ml
maximedenes: I know it doesn't change much, but there are two cleanups needed. One is to remove `Evd.empty` so that we get universe constraints right in general. The other is to stop passing around empty maps, to get the evar substitution right. This PR is a step only for the first cleanup.

herbelin (approved):
Purpose of the PR ok to me.

plugins/funind/invfun.ml
forestjulien: @maximedenes : maybe I'm wrong but I don't think the evd change anything here since simple zeta normalize the term and, as far as I understand things, the evars and universes constraints are already solved in the term before its reduction. 
If this vision is false then the modif in get_body (file plugins/funind/functional_principles_proofs.ml) can change the behavior of function, an I will need more tests before approving the patch here.

plugins/funind/functional_principles_proofs.ml
```diff
@@ -1098,10 +1098,12 @@ let prove_princ_for_struct (evd:Evd.evar_map ref) interactive_proof fun_num fnam
...
       match Global.body_of_constant const with
 	| Some (body, _) ->
+          let env = Global.env () in
+          let sigma = Evd.from_env env in
 	     Tacred.cbv_norm_flags
 	       (CClosure.RedFlags.mkflags [CClosure.RedFlags.fZETA])
-	       (Global.env ())
-	       (Evd.empty)
+               env
+               sigma
```
maximedenes: @forestjulien I believe you were commenting this line.

> I will need more tests

You mean manual tests?

forestjulien: yes

maximedenes: IMHO it would be better to extend the funind test suite, rather than rely on manual testing.

ppedrot (approved):
Globally looks OK to me, even though some places actually look like they should be using kernel terms instead but don't because of API deficiency.

plugins/funind/functional_principles_proofs.ml
forestjulien: Ok I'm convinced. The word "test" was probably erroneous. I wanted to convince myself we cannot have any problem in the future nor with an unknown contrib.

forestjulien (approved):

PierreCorbineau (approved):

ppedrot:
Essentially waiting for @mattam82's input.

ppedrot:
@mattam82 This PR is trivial enough so as to consider that if you don't complain very soon, it will be merged regardless.

ppedrot:
@maximedenes Can you rebase please?

maximedenes:
Rebased.

ppedrot:
So it seems we have two CI failures.
- fiat-crypto dies from memory exhaustion, which happens from time to time. I don't know whether this PR worsens the situation.
- dpdgraph is broken by something that looks like an unrelated change.

@maximedenes any comments?

maximedenes:
@ppedrot the failures look unrelated to me, don't they?

ppedrot:
I guess so, let's merge then.

Issue #7422, opened 2018-05-04
[api] Removal of the `*ops` modules


ejgallego:
In the current codebase we find this pervasive pattern:
```ocaml
open Foo
open Foops
```
with in some cases, `Foo` being an "interface only file" while in others, `Foo` contains also functions.

IMO, we should remove this pattern an unify and reduce the number of modules. Less modules is IMHO a key point in order to make the API accessible.

Another problem this pattern has is the development of API documentation (#7155), see discussions in #6512, https://github.com/coq/ceps/pull/19 and likely in many other places [please link to discussions below].

This change has no downside I can think of. In the rare cases where this cleanup may make the size of the kernel larger, we could perfectly mitigate this using other techniques [if we really worry about that we should have a `safe_kernel` directory and a `kernel` one that export the first and adds utilities.

Zimmi48:
I would rather call them `kernel` and `kernelops` ^^

maximedenes:
I support this, the separation looks very artificial in many cases, makes it harder to find API info, and does not seem to bring any immediate benefits.

Issue #7155, opened 2018-04-02
ML API documentation meta issue


ejgallego:
C̶o̶q̶ ̶c̶u̶r̶r̶e̶n̶t̶l̶y̶ ̶p̶r̶o̶v̶i̶d̶e̶s̶ ̶a̶ ̶w̶a̶y̶ ̶t̶o̶ ̶g̶e̶n̶e̶r̶a̶t̶e̶ ̶a̶ ̶d̶o̶c̶u̶m̶e̶n̶t̶a̶t̶i̶o̶n̶ ̶o̶f̶ ̶t̶h̶e̶ ̶m̶l̶-̶l̶e̶v̶e̶l̶ ̶i̶m̶p̶l̶e̶m̶e̶n̶t̶a̶t̶i̶o̶n̶ ̶u̶s̶i̶n̶g̶ ̶t̶h̶e̶ ̶`̶s̶o̶u̶r̶c̶e̶-̶d̶o̶c̶`̶,̶ ̶`̶.̶m̶l̶i̶-̶d̶o̶c̶`̶,̶ ̶`̶m̶l̶-̶d̶o̶c̶`̶ ̶m̶a̶k̶e̶ ̶t̶a̶r̶g̶e̶t̶s̶.̶

[...]

PR #6512, opened 2017-12-27
[api] Relocate `intf` modules according to dependency-order.

ejgallego:
In a component-based source code organization of Coq `intf` doesn't
fit very well, as it sits in bit of "limbo" between different
components, and indeed, encourages developers to place types in
sometimes random places wrt the hierarchy. For example, lower parts of
the system reference `Vernacexpr`, which morally lives in a pretty
higher part of the system.

We move all the files in `intf` to the lowest place their dependencies
can accommodate:

- `Misctypes`: is used by Declaremod, thus it has to go in `library` or below. Ideally, this file would disappear.
- `Evar_kinds`: it is used by files in `engine`, so that seems its proper placement.
- `Decl_kinds`: used in `library`, seems like the right place. [could also be merged]
- `Glob_term`: belongs to pretyping, where it is placed.
- `Locus`: ditto.
- `Pattern`: ditto.
- `Genredexpr`: depended by a few modules in `pretyping`, seems like the righ place.
- `Constrexpr`: used in `pretyping`, the use is a bit unfortunate and could be fixed, as this module should be declared in `interp` which is the one eliminating it.
- `Vernacexpr`: same problem than `Constrexpr`; this one can be fixed as it contains stuff it shouldn't. The right place should be `parsing`.
- `Extend`: Is placed in `pretyping` due to being used by `Vernacexpr`.
- `Notation_term`: First place used is `interp`, seems like the right place.

Additionally, for some files it could be worth to merge files of the form `Foo` with `Foo_ops` in the medium term, as to create proper ADT modules as done in the kernel with `Name`, etc...

Depends on #6859

PR #7423, opened 2018-05-04
[api] Merge `Declareops` into `Declarations`

ejgallego:
This is an experiment in trying to get a more uniform API by merging
together definitions with their operating interfaces. The main
motivation to help the API documentation work (#7155), but this also
has a net effect on the number of `open` needed in the codebase and no
downside I can see.

Also see discussions in #6512, https://github.com/coq/ceps/pull/19 and likely elsewhere.

IMO reducing the number of modules is a key point in order to make the
API accessible, if this PR finds some success the pattern:

```ocaml
open Foo
open Fooops
```

should hopefully go away.

cc: #7422


kernel/declarations.ml
```diff
@@ -300,3 +300,382 @@ and _ module_retroknowledge =
...
+
+(************************************************************************)
+(*         *   The Coq Proof Assistant / The Coq Development Team       *)
+(*  v      *   INRIA, CNRS and contributors - Copyright 1999-2018       *)
+(* <O___,, *       (see CREDITS file for the list of authors)           *)
+(*   \VV/  **************************************************************)
+(*    //   *    This file is distributed under the terms of the         *)
+(*         *     GNU Lesser General Public License Version 2.1          *)
+(*         *     (see LICENSE file for the text of the license)         *)
+(************************************************************************)
```
Zimmi48: This copy-paste is very radical, you could have removed the headers :laughing:

forestjulien (approved):

ppedrot (dismissed):
In this specific case, I am against the merge of the structures together with the operations. Indeed, the fact we have a separate, declarative-only interface for the kernel makes it easy to review the core structures that ought to be manipulated in there. For instance, it helps quickly updating the structures used in the checker with the ones used in the kernel. If we start to stuff it with proof-relevant information (i.e. programs), then it becomes harder to understand this with a quick glance.

More generally, I think we should strive for a kernel that is as understandable and simple as possible, even if this locally goes against recommended engineering practices.

ejgallego:
Umm, I am afraid Pierre-Marie that I have quite a problem following your argument here, I think that would work if the "operations" as you call them would be outside the trusted base, but they are actually inside; so in fact I think review is easier here, not to say usage.

So how does it having things scattered around help to have _"a kernel that is as understandable and simple as possible"_?

ejgallego:
Not to say about having proper API docs...

maximedenes:
I'm with @ejgallego on this one. Opposing "a kernel that is as understandable and simple as possible" and recommended engineering practices is very strange to me. I suggest we instead implement the standard recommended practices, an tweak a bit the codebase to make it easy to understand what the kernel is and review it.

We could even seize the opportunity of making clearer what the TCB is, which I suspect many people who look at the code think is the same thing as the kernel. Today, we have functions that are in the TCB outside of the kernel, and functions in the kernel only used by untrusted clients (so it is not in the TCB).

> it helps quickly updating the structures used in the checker with the ones used in the kernel

The fact that it helps to develop the checker by copy/pasting changes from the kernel doesn't sound very convincing to me. We should rather stop doing that.

maximedenes:
And, to add something, the core structures manipulated in the kernel should indeed be easy to understand (thanks to a documentation), but it should be the case of all components of the system. So it seems to me that @ppedrot is a bit biased by the fact that today, the documentation of the kernel APIs is superior. But that doesn't mean we can't add anything there, if we document it properly.

ppedrot:
Okay, okay, I'm not complaining that loudly either. You can dismiss my review if you want...

ejgallego:
> Okay, okay, I'm not complaining that loudly either. You can dismiss my review if you want...

Pierre-Marie, I take very seriously your comments and I am trying hard to understand what the advantages of having the interface split are [even so then I'd prefer to have `Declarations_intf` then included by `Declarations`]

I am failing to do so; only advantage I see is to avoid duplication but here IMHO [apart from the checker setup but I'm with Max here] I don't see how we can profit from it. Anyways, I consider this as an RFC and not urgent, it'd be great to discuss about this a bit in person, but yeah, I dunno when.

I am confident saying tho that experimental evidence seems to favor having compact interfaces; apart from the obvious problems with proliferation of modules and types [where Coq is now well beyond what a single human can grasp], IMHO modules such as `Constr` should expose the complete term-manipulation API. Not doing so has led to plugin writes to duplicate large parts of this infrastructure, [see Talia Ringer Pumpkin's for example], and worse, it makes hard for people to contribute/express their needs in terms of API.

The TCB issue IMHO should be tackled separatedly from the API issue IMO.

Zimmi48:
You mean Talia Ringer's Pumpkin.

herbelin:
As far as I'm concerned, I also found pleasant that `Declarations` has only types. This is the informative part of the API, and still now, I regularly refer to this well-isolated part of the API to remember what the exact types are.

On the other side, I can also adapt to whatever largely-supported alternative model.

In practice, I'm a bit afraid that if we start moving the `Foo_ops` file to the `Foo` files (there are syntactically-speaking 18 of them), this will induce a burden on plugin maintainers which is maybe not so necessary. If the point is to better allow users not to duplicate code by finding it easily in the API, wouldn't a header to `Foo` telling to look at `Foo_ops` for related functions be sufficient?

For instance, I would feel better if the change is accompanied by objective feedbacks from plugin developers (Talia Ringer, Beta Ziliani, ...) on the pros and cons associated to one or the other structuration of the API. 

About "standard recommended practices", my experiences is that they are not free from "fashion of the day", so, I would not lift "standard recommended practices", whatever they are, as necessarily to be followed.

On the other side, "fashion of the day" is good to let people feel in known territory, and if our way of doing gives the feeling, outside, that Coq is a non-standard harsh-to-understand territory, it might indeed be useful to follow current recommended practices to make it feel less frightening.

By the way, what pointers to the current state of "standard recommended practices" could we use as a reference?

> IMHO modules such as Constr should expose the complete term-manipulation API. Not doing so has led to plugin writes to duplicate large parts of this infrastructure, [see Talia Ringer's Pumpkin for example], and worse, it makes hard for people to contribute/express their needs in terms of API.

I'm sensitive to this issue. How much are we good or bad here? I'm maybe in an optimistic day, but the situation does not look so bad to me. Maybe are we missing links from `Constr` to `EConstr`, to `Termops`? Maybe should we clarify the `Term`/`Constr` terminology?

Hopefully the CIW will be an opportunity to let people contribute/express their needs in terms of API.

ejgallego:
@herbelin thanks for you remarks, a few comments/questions:

> this will induce a burden on plugin maintainers which is maybe not so necessary

What would be such a burden? Note that this change doesn't require any code change from plugin developers [for the moment]

> wouldn't a header to Foo telling to look at Foo_ops for related functions be sufficient?

well, the point is that this seems overkill, why the indirection? I am failing to see the advantage and having 18 more modules it is indeed a big burden in terms of name memory for developers. Specially as, contrary to `Declare` most of the other files are not type-only files but they do include functions.

> I also found pleasant that Declarations has only types. This is the informative part of the API, and still now, I regularly refer to this well-isolated part of the API to remember what the exact types are.

I am not sure how adding some functions make the above impossible but indeed I could understand if you folks want to have a `Declare_intf` module only for the interface, still the expose module should include functions. The typical use case for the above is to avoid duplication which is not the case here.

> standard recommended practices

I am not sure about STP, but indeed this `ops` structuration is not seen a lot in the wild and i guess it is due to not having any advantage and in particular forcing devs to search two modules instead of one. The goal here is make the API smaller, thus purely quantitative.

> How much are we good or bad here?

Based on the duplication of term-manipulation functions on the codebase I'd dare to say we are doing less than optimal here. Other may confirm, but the typical case is that devs look at `Constr`, see all the lacking functions, then they write their own. An the fact that functions are distributed over two layers seems to be very confusing, quoting an [active plugin author](https://gitter.im/coq/coq?at=5af1bc7340f24c430460b54f):

> So I think what I would expect from the a codebase like Coq's is that there are layers (or phases?). The kernel is clearly the lowest layer and it's mostly separate from other stuff as far as I can tell. But I don't actually know if any of the other code corresponds to some layer model. To me, the ..ops suffix usually communicates that they call functions from the module without the suffix and probably a lot of other stuff. so they are in some sense on a higher layer, and I don't dislike that particular distinction.
I say layers because in the end Coq is an interpreter/compiler and I would expect the outermost layer to be user-facing stuff and then there would be a lot of layers inbetween that grow more and more "removed" from anything the user sees
It might be the case that these layers exists (and maybe are not spelled out anywhere) and my difficulties come from not knowing them. But I suppose the recent discussion about how to declare a definition shows that there is some kind of problem with the structure of the code, or at least with the documentation?

So indeed a big part of this proposal is making writing API doc easier, and for that you usually want a complete "coherent" unit for each concept in the codebase.

ejgallego:
Thanks for the comments @herbelin , a few more remarks:

> a mental model that they started to get familiar with.

well, this may be a bit questionable given the evidence in terms of code duplication, and also because the `ops` model is a minority in terms of the total codebase. And I do believe it is so because in general most modules tend to be self-contained.

> It is more about the mental model of separating types and methods.

Indeed I am having trouble understanding this model as for me types are characterized by their operation set [algebras if you want], but I admit that is just my inability to understand the other model so I do appreciate comments. 

What are the cases where plugin writers would be expected to operate on types / constructors directly? In most cases I'd say that if plugin writers have to do that it is because the right API is missing for them.

> It is only smaller in the number of files...

But this is very important IMHO! Think of code completion! Now I have to try `Constr.`, then maybe `Term.`, then maybe `Termops.` , etc... it is hard to understand what's available to me.

> Then, there was the objective of Bruno to move out of the kernel what was not strictly needed there, which was the origin of Termops, Inductiveops, Reductionops.

Thanks for the historical overview, indeed I do agree that TCB issues are a separate issue and that should be also addressed, in this PR and associated issue I am thinking about the externally observable API, it could be well implemented as a layer on top of a TCB `kernel/tcb`.

> Then, there was the objective of Pierre to make the compilation more modular, by drawing less dependencies for the sole reason that a file needed for its types was also dependent on the files needed only for the implementation of its methods.

Yeah it'd be great to hear from Pierre about the concrete issues, at least in the current codebase such occurrences are rare, if a file is depending on a type it is needing some operations about it so in fact the interfaces were not really improving modularity a lot AFAICS.

We had some circularity issues in the past but IMHO that was due to our build and debug system not being user-friendly enough as for developers to implement the proper workaround thus people tended to add types at the lowest level they could.

herbelin:
> Indeed I am having trouble understanding this model as for me types are characterized by their operation set [algebras if you want].

A type can also be seen as determining its algebra just by its constructors. Maybe in play is the difference of view between a type as carrying a semantics (no need to explicit accessors or constructors), or as hiding an implementation. In the second case, there is indeed no other way that characterizing the type by its methods.

Well, there are then the derived combinators and I guess that this is mainly what we are talking about here.

There are also types serving the purpose of communication, and not intended to be equipped with methods.

> Yeah it'd be great to hear from Pierre about the concrete issues, at least in the current codebase such occurrences are rare, if a file is depending on a type it is needing some operations about it so in fact the interfaces were not really improving modularity a lot AFAICS.

> We had some circularity issues in the past but IMHO that was due to our build and debug system not being user-friendly enough as for developers to implement the proper workaround thus people tended to add types at the lowest level they could.

Maybe. I must confessed that I was gladly surprised that you managed to dispatch `intf` w/o apparently increasing the complexity of the dependency graph.

> Thanks for the historical overview, indeed I do agree that TCB issues are a separate issue and that should be also addressed, in this PR and associated issue I am thinking about the externally observable API, it could be well implemented as a layer on top of a TCB kernel/tcb.

On the specific question of `constr` combinators `fold`, `map`, ... I found that the attempt to restrict the TCB was actually a bit hopeless. First we don't count `lib`, `clib`, which may in some cases be critical for the TCB (e.g. the error-prone implementation of smart or tail-recursive `map`). Then, moving around a function from `Termops` to `Term` back and forth depending on whether we use it in the kernel is a bit painful. For simplicity, we should rather accept that standard combinators are in one single place, accessible from the kernel even if not strictly used by the kernel. Or, maybe, the kernel should be split into a general-purpose `kernellib` part declaring the basic types (`Names`, `Constr`, `Context`, `Declarations`, ...), seen as a library in the same spirit as `lib`, and a part implementing the type-checker properly speaking (`Safe_typing`, `Typeops`, ...), with a clear API made of the inference rules of the pCuCiC.

> But this is very important IMHO! Think of code completion! Now I have to try Constr., then maybe Term., then maybe Termops. , etc... it is hard to understand what's available to me.

Yes, it is probably indeed about understanding the motivations of the structure of an API, whatever this structure is. So, let's go towards an API which is well-explained.

Zimmi48 (approved):
Currently `declaration.ml` doesn't have any corresponding `declaration.mli` (famous unresolved `.mli` / `.ml` debate for files with only types). This PR fixes this because this file is not just types anymore. Consequently, I'm very much in favor of this PR and against any file that has only types.

~~(Note that I didn't read the discussion above fully yet.)~~ (now done)

mattam82:
I'd be much happier if Foo.bar 's operations were mainly found in Foo. for autocompletion / discovery reasons. As a plugin maintainer I find renaming the easiest part of maintainance.

mattam82 (approved):

herbelin (approved):
I'm basically ok with whatever local solution is eventually adopted about `Declareops`.

ejgallego:
Not worth the rebase, the original issue is still open tho.

PR #7435, opened 2018-05-05
[gitlab] Add bleeding-edge flambda build.

ejgallego:
I also introduced a bit more systematic job naming.

This is as we are getting close to being able to remove quite a few
builds from Travis.

Note the extra quote rule for the flambda parameters, but unless we
can assign arrays to Gitlab variables there is not a good way to do
this I'm afraid.

Depends on #7387


ejgallego:
another option for naming is `build:base` and `build:edge`

SkySkimmer:
Don't add some unrelated commit.

SkySkimmer:
This cache key is identical to the one in #7442 
Maybe we should use another system, date isn't great for parallel development.

ejgallego:
> Maybe we should use another system, date isn't great for parallel development.

I dunno, it is very readabable and gives good info, we could try adding a version number or a hash?

SkySkimmer:
Incrementing version doesn't work for parallel either.
Maybe use the hash of Dockerfile.

Gaëtan Gilbert

ejgallego:
Yup we can add a hash. Anyways I expect the date to work reasonably well as the update frequency should be really low-level.

ejgallego:
Rebased and build pushed; hopefully this is the last docker update for a while.

.gitlab-ci.yml
```diff
@@ -211,10 +242,8 @@ ci-equations:
 ci-fcsl-pcm:
   <<: *ci-template
 
-# ci-fiat-crypto:
-#   <<: *ci-template
-#   # out of memory error
-#   allow_failure: true
+ci-fiat-crypto:
+  <<: *ci-template
```
SkySkimmer: Why are you resurrecting fiat-crypto?

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -43,4 +43,4 @@ RUN opam switch -y -j $NJOBS $COMPILER_BE && eval $(opam config env) && \
 
 # BE+flambda switch
 RUN opam switch -y -j $NJOBS "${COMPILER_BE}+flambda" && eval $(opam config env) && \
-    opam install -j $NJOBS $BASE_OPAM camlp5.$CAMLP5_VER_BE $COQIDE_OPAM_BE
+    opam install -j $NJOBS $BASE_OPAM camlp5.$CAMLP5_VER_BE $COQIDE_OPAM_BE $CI_OPAM
```
SkySkimmer: Why?

ejgallego: If possible I'd like this switch to be ready to build any CI too; if needed and without a new Docker rebuild.

I can document this in the commit log if you want.

.gitlab-ci.yml
ejgallego: Why not? As far as I can tell it should work fine , workers have more than enough ram.

The idea of this PR was actually to allow to remove all Travis jobs but for the linter and the Mac one, so this is the reason I add it back.

ejgallego: I can document better if you want.

ejgallego: In fact I should add the lint job too so we are complete?

ejgallego: Maybe the workers got upgraded?

ejgallego: The specs are pretty beefy these days, so if you think that's OK I'll add the linter job to this PR so we have a complete Travis replacement just in case [except for OSX, but that's a different story].

SkySkimmer: The main reason the linter isn't on gitlab is that it uses the TRAVIS_COMMIT_RANGE variable. There's no equivalent on gitlab, especially considering we run as a normal branch job.
Do you have a plan to deal with that?

ejgallego: Nope, in fact I ported everything but the linter due to it looking complicated. So let's the linter remain there for now, if Travis becomes a linter-only job for PR that's fine [as we can interpret the red light as a certain linter failure then]

ejgallego: In `oexp` I used, `git rev-list --ancestry-path %s..%s`, to get a commit range, I dunno if that could work here.

ejgallego:
> Maybe just make the configure option additive, then you would pass -flambda-opt -O3 -flambda-opt -unbox-closures

I thought of that, but common practice in flag setup is to use one quoted parameter. I guess it works like this so I wouldn't complicate the thing too much. In fact until recently the plan was to use Dune for CI and that allows for a more structured setup.

SkySkimmer (approved):

Issue #6919, opened 2018-03-05
[CI] Meta-Issue for Gitlab migration


ejgallego:
It is not a secret that Coq CI demands are pretty high. As of today we need more than 10 hours build time of a medium-size CPU instance to test one PR. Would we do more serious testing this constraint would easily multiply by 10, so free plans are not an option. Custom commercial plans could be considered, however, Gitlab offers the possibility of running our own runners in our own servers. This is very attractive for several reasons:

- No time limit per build: we can test developments that take a few hours to compile.
- No limit on the number of instances: Gitlab can spin spot AWS instances so latency is minimal.
- Open source: Gitlab is the only mature integral hosting/CI infrastructure that is open source.
- Cleanup of the CI system: right now we support 3 different systems and we cannot standardize on a single platform, with the overhead this implies.
- All kind of custom artifact building and deployment.

Most of the work to migrate to Gitlab is already in place thanks to @SkySkimmer . I have setup a couple of reasonably powerful runners on MINES test servers so we should be covered in this sense too. The runners have been tested and they do seem to work and don't involve any undesired hack. A possible plan of action is:

- s̶e̶t̶u̶p̶ ̶a̶ ̶p̶r̶o̶p̶e̶r̶ ̶C̶o̶q̶ ̶m̶i̶r̶r̶o̶r̶ ̶a̶t̶ ̶h̶t̶t̶p̶s̶:̶/̶/̶g̶i̶t̶l̶a̶b̶.̶c̶o̶m̶/̶c̶o̶q̶/̶c̶o̶q̶ ̶,̶ ̶r̶e̶g̶i̶s̶t̶e̶r̶ ̶t̶h̶e̶ ̶`̶e̶t̶i̶l̶i̶o̶`̶ ̶r̶u̶n̶n̶e̶r̶ ̶[̶3̶2̶ ̶c̶o̶r̶e̶s̶/̶3̶2̶G̶i̶B̶ ̶R̶A̶M̶]̶.̶
[...]

PR #7436, opened 2018-05-05
[travis] Remove some more jobs from PR testing now that they are on Gitlab.

ejgallego:
This is a "test" PR, but could be merged if we like it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,16 +49,22 @@ env:
   - NATIVE_COMP="yes"
   - COQ_DEST="-local"
   - MAIN_TARGET="world"
-  # Main test suites
-  matrix:
-  - TEST_TARGET="test-suite" COMPILER="4.02.3+32bit"
-  - TEST_TARGET="validate"                           TW="travis_wait"
-  - TEST_TARGET="validate"   COMPILER="4.02.3+32bit" TW="travis_wait"
-  - TEST_TARGET="validate"   COMPILER="${COMPILER_BE}+flambda" CAMLP5_VER="${CAMLP5_VER_BE}" NATIVE_COMP="no" EXTRA_CONF="-flambda-opts -O3" FINDLIB_VER="${FINDLIB_VER_BE}"
 
 matrix:
 
   include:
+    - if: NOT (type = pull_request)
+      env:
+      - TEST_TARGET="test-suite" COMPILER="4.02.3+32bit"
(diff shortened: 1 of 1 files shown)


ejgallego:
Rebased and updated to latest setup, only lint and osx jobs are remaining.

Build on branch [thus full] is at:
https://travis-ci.org/ejgallego/coq/builds/376133831

I would suggest we also remove some jobs from the full build, in particular we could only do one test suite and validate, but that can happen in a separate PR I guess.

ejgallego:
Théo I agree with the plan and indeed it's time we close the documentation issue. However this PR is IMHO a step towards that, I think it is good to remove the testing in gradual steps.

Fiat crypto is readded back in the depending PR if I am not mistaken. It works fine AFAICT

Zimmi48:
> Fiat crypto is readded back in the depending PR if I am not mistaken. It works fine AFAICT

OK, I thought @SkySkimmer was doubting this. Does it work fine even with shared runners?

ejgallego:
> OK, I thought @SkySkimmer was doubting this. Does it work fine even with shared runners?

So far I haven't found a problem with fiat-crypto and given the specs of the shared runners I would be surprised to see one. The problem in the past was likely that `fiat-crypto` itself was way more demanding or that the runners had lower specs.

Zimmi48:
OK great to hear!

Zimmi48:
So given the pretty terrible situation of Travis tests currently (lots of problems installing dependencies) and given how easy it is to setup GitLab CI support for one's fork, I would say it is time to remove these jobs from Travis for real (both on PRs and on branches). I'm going to do the update to the CI doc. The present PR should also include a removal of the Travis badge...

Zimmi48:
Cf. #7482.

ejgallego:
IMHO it is still too early to move from Travis, reliability problems are not exclusive of it [gitlab has it's own bunch of problems]. In particular I'd like to see #6919 closed first.

Zimmi48:
At least, move these jobs that you keep but are redundant to the allow failure category. I don't want to have Travis badge in the README red because yet another time out or failed to fetch issue. The color of the Travis badge should only reflect the status of the macOS job.

SkySkimmer (approved):

SkySkimmer:
OK I'm not going to wait for gitlab's windows.

PR #7437, opened 2018-05-05
[coqdep] Minor cleanups.

ejgallego:
- Remove inclusion of the `tactics` directory, this is coming from a
  time loadable modules were found there, now all are under `plugins`.
- Remove 4 dependencies on modules so we can bootstrap coqdep earlier.
- Use `Format` instead of `Printf` for printing.

### DIFF
--- a/tools/coqdep_common.ml
+++ b/tools/coqdep_common.ml
@@ -8,9 +8,9 @@
 (*         *     (see LICENSE file for the text of the license)         *)
 (************************************************************************)
 
-open Printf
-open Coqdep_lexer
+open Format
 open Unix
+open Coqdep_lexer
 open Minisys
 
 (** [coqdep_boot] is a stripped-down version of [coqdep], whose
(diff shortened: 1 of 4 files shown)


ejgallego:
The idea here is to remove `coqdep_boot`, I realized about this as in Dune we can build the full coqdep directly, so its existence is just an artifact of the makefile setup.

In principle, we would have a single coqdep that depends on a `config` lib that contains the config info and can be built without any special setup, in this config lib we could centralize the miscellaneous bits found now in `Coq_config`, `Flags`, and `Envars`. In fact, there is not a lot to store there other than the coqlib location function and maybe a hardcoded path for it in ML if the user wishes so.

The dependency on `CoqProject_file` is legitimate, and it should become a dependency on `coqfind` at some point.

tools/coqdep_common.ml
```diff
@@ -20,14 +20,15 @@ open Minisys
     options (see for instance [option_dynlink] below).
 *)
 
+let coqdep_warning args =
+  eprintf "*** Warning: @[";
+  kfprintf (fun fmt -> fprintf fmt "@]\n%!") err_formatter args
+
```
ppedrot: I don't remember whether the proper hooks were plugged before this patch, but it would be nice to have terminal colouring in that function. I remember that @SkySkimmer added lightweight support for colour in `configure`, the same thing should be done here.

ejgallego: That never had color except for a couple of cases where it was calling Coq's feedback, but IMHO it should use that. Note that this component is critical dependency-wise as it has to be bootstrapped as soon as possible. Ideally that just depend on `coqfind`.

ppedrot (approved):
Looks good to me.

ejgallego:
Appveyor failure spurious, it is really unreliable.

PR #7438, opened 2018-05-05
[gitlab] Do expensive builds with the flambda compiled Coq.

ejgallego:
Depends on #7435


ejgallego:
Some interesting timing data from Gitlab [I checked the builds were done in similar spec'd runners]

| Contrib | Regular Build | Flambda Build |
| --------+---------------+---------------|
| color     |       | 19:32 |
| formal-topology   | 47:39 | 21:06 |
| geocoq    | 47:52 | 22:04 |
| iris      |       |  |
| math-comp | 28:45 | |
| vst       |       | |

Empty data is waiting for a rebuild as the jobs didn't build in the same worker.

If this data is true, it means we are close to stupid.

Note that artifact download is a bit larger in the regular build, but
I don't think this can make up for such impact.

ejgallego:
cc: @ppedrot

.gitlab-ci.yml
```diff
@@ -125,6 +125,14 @@ before_script:
   variables: &ci-template-vars
     TEST_TARGET: "$CI_JOB_NAME"
 
+.ci-template-flambda: &ci-template-flambda
+  <<: *ci-template
+  dependencies:
+    - build:be+flambda
+  variables:
+    OPAM_SWITCH: "be"
```
SkySkimmer: You need to `<<: *ci-template-vars`.
All the data is invalid except for color: TEST_TARGET doesn't get set so we run `make -f Makefile.ci` and the default target is color.

SkySkimmer (dismissed):
x_x

.gitlab-ci.yml
SkySkimmer: (comment is shown outdated because it's made on a single commit and the diff with all commits looks different)

SkySkimmer:
Looks like runner 40786 finished "math-comp" color twice as fast as the other runners though.

ejgallego:
> Looks like runner 40786 finished "math-comp" color twice as fast as the other runners though.

Likely it has two threads, that's the variation.

.gitlab-ci.yml
ejgallego: Ahh, working so late can never be good. Let's try again-

ejgallego: > You need to <<: *ci-template-vars.

My lack of knowledge of yaml bite here; I assumed that `<<: *ci-template` would make the latter variable declaration "additive". Let's try again.

SkySkimmer:
Disable shared runners on your repo and rerun both master and this PR, then you can compare times on the same runner.

ejgallego:
> Disable shared runners on your repo and rerun both master and this PR, then you can compare times on the same runner.

Yup that's a good idea, will have to wait a few days tho, anyways this is not urgent at all.

ejgallego:
Rebased, IMHO at least the `ci-template-flambda` target is interesting. On the other hand it is not clear what jobs we should add to this target. I would at least have one, if only for fun.

ejgallego:
@RalfJung reports on gitter:

> okay I finally finished the flambda test; with the flag you suggested total user time for lambdaRust is 27min. that's a 15% speedup compared to vanilla ocaml.

SkySkimmer (approved):

ejgallego:
So I dunno about the list of "expensive" build I chose, let's keep an eye to see how they are doing, maybe we want to revert some to use the regular Coq.

PR #7254, opened 2018-04-15
Inductive params

jashug:
**Kind:** cleanup
Closes coq/coq#7253
- [x] Entry added in CHANGES.
- [x] Add test

I wonder how often this is used in practice. I hope CI will show it is not used very often.

Looks like this feature was last touched by @herbelin in https://github.com/coq/coq/commit/1f798216ede7e3813d75732fbebc1f8fbf6622c5

PR #7439, opened 2018-05-06
[WIP] Add support for explicitly uniform inductive parameters

jashug:
**Kind:** feature

TL;DR: Support `Uniform Inductive` which fixes the parameters in the body of the inductive.

#7254 is a proposal to remove ad-hoc filling of inductive parameters. That makes the pattern below not work, where we have parameters we don't want to mention, and are uniform but not inferable:
```
Inductive expr {vars : Type} : Type :=
  add : expr -> expr -> expr.
```
This is commonly used by fiat and @JasonGross.

This feature is a proposal to support
```
Uniform Inductive expr (vars : Type) : Type :=
  add : expr -> expr -> expr.
```
where `Uniform` indicates all parameters are uniform. Inside the constructors, `expr` has type `Type`, so there is no need to infer a value for vars. After pretyping, constructors are generalized over the uniform parameters.

As a user, I think this is a useful semantic indicator, and can also save on writing out the parameters in the constructors when you want them to be uniform.

My current implementation is rather hacky, but it is remarkably short (17 lines for the meat of the code; hopefully it is not too terse). Suggestions welcome.

- [ ] Support `Set Uniform Inductives` to make this default (`NonUniform` prefix to escape)
- [ ] Add tests
- [ ] Corresponding documentation was added.
- [ ] Entry added in CHANGES.

---

In #7254, @herbelin suggested filling the remaining holes with the default heuristics (i.e. the canonical instantiation `(var:=var)`) using an evar flag or evar kind. I think the approach I take here is simpler and easier to understand / control (no need to track whether each hole is a parameter of the current inductive type).

### DIFF
--- a/library/decl_kinds.ml
+++ b/library/decl_kinds.ml
@@ -22,6 +22,8 @@ type private_flag = bool
 
 type cumulative_inductive_flag = bool
 
+type uniform_inductive_flag = bool
+
 type theorem_kind =
   | Theorem
   | Lemma
--- a/parsing/g_vernac.ml4
+++ b/parsing/g_vernac.ml4
@@ -162,11 +162,12 @@ GEXTEND Gram
       | IDENT "Let"; id = identref; b = def_body ->
           VernacDefinition ((DoDischarge, Let), (lname_of_lident id, None), b)
       (* Gallina inductive declarations *)
-      | cum = OPT cumulativity_token; priv = private_token; f = finite_token;
+      | cum = OPT cumulativity_token; priv = private_token;
+        uni = OPT uniformity_token; f = finite_token;
         indl = LIST1 inductive_definition SEP "with" ->
 	  let (k,f) = f in
           let indl=List.map (fun ((a,b,c,d),e) -> ((a,b,c,k,d),e)) indl in
-          VernacInductive (cum, priv,f,indl)
+          VernacInductive (cum, priv, uni, f, indl)
       | "Fixpoint"; recs = LIST1 rec_definition SEP "with" ->
           VernacFixpoint (NoDischarge, recs)
       | IDENT "Let"; "Fixpoint"; recs = LIST1 rec_definition SEP "with" ->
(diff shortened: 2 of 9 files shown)


ppedrot:
I think this is typical use-case for the yet-to-be merged attribute system. For the record, I am personally against the modifier inflation for toplevel declarations.

jashug:
@ppedrot I'm sympathetic to that position. I really want to have a way to say something like
`Inductive I (X : Set) with non_uniform (x : X) : nat -> Type := C : forall y, I y 3 -> I x 5.`,
some way of specifying both uniform and non-uniform parameters at once. However, I couldn't come up with a nice syntax, or a way to be backwards compatible.

Once attributes is merged, this should be extremely straightforward to convert, since I basically copied the `Cumulative` flag (which I assume would also become an attribute).

Zimmi48:
That attributes are not there yet is clearly not a good reason to block this PR in the meantime. Hopefully, we will get attributes before the 8.9 release anyways and this new syntax won't even make it into a released version...

gares:
> However, I couldn't come up with a nice syntax, or a way to be backwards compatible.

I think that the syntax for inductive type declaration did grow organically for very long.
IMO it is better to try to design a new one in order to toss away al backward compatibility issues (e.g. the fact that non implicit but uniform parameters have to be repeated). It can then be compiled in terms of the current syntax/functionalities.

The following snippet is just brainstorming, if there is interest we can try to write an RFC (Cep).
```
(Data|Relation) name [[uniform] (p1 : P1) ... (pn : Pn)] [, [nonuniform] (a1 : A1) ... (an : An)] [, [index] (i1 : I1) ... (in : IN)].
```
Up there there is no `Prop` or `Type` but rather a keyword to tell the intended use of the inductive type.
All kind of parameters are in the style that is supported for constructors (eg `K (a1 : T1) ... (an : Tn)`) that I find quite clear to understand.

A new syntax is clearly out of the scope of this PR, but it may be the good way to get what this PR tries to achieve.

jashug:
Closed in favor of #7703, which uses an option instead of this psuedo-attribute.

PR #6808, opened 2018-02-21
Add unit tests to test-suite

psteckler:
This PR creates a way to run unit tests on arbitrary ML code using the OUnit framework, from within the test-suite. Any ML code that's visible from Coq should be able to run in tests.

The ML code gets compiled and run, which generates a log file that can be used by the existing test-suite Makefile. Each ML file generates a single log file, which may contain the results of multiple tests. Each test result gets an entry in the log file. Using some helper functions to create tests gives a name for tests, which name is visible in the test results.

See `test-suite/unit-tests/clib-tests/inteq.ml` for an example test file.

`test-suite/Makefile` has a new target, `unit-tests`. That target is built when `make` is run without arguments. The `clean` target removes some additional files (`.cmo`, `.cache`, etc.) that get created in the `unit-tests` subdirectory.

This is a proof-of-concept, and surely can be improved.

**Kind:** feature / infrastructure.

Fixes #6755.

- [x] Corresponding documentation was added / updated.
- [x] Entry added in CHANGES.

PR #7442, opened 2018-05-06
Gitlab: build docker image in pipeline and use through registry.

SkySkimmer:
A docker-boot job is added which builds the docker image and pushes it
to the registry, the other jobs running on the image from the
registry.

The boot job reuses the already pushed image if it exists and matches,
this is important to cut down its runtime.

Making a new image takes 30min for all the switches
https://gitlab.com/SkySkimmer/coq/-/jobs/66650546
For testing I removed all jobs except boot and main build, and after
that run I built only the standard switch.

Building 1 switch takes around 20min
https://gitlab.com/SkySkimmer/coq/-/jobs/66656942
When the registry is up to date it takes 2min
https://gitlab.com/SkySkimmer/coq/-/jobs/66658790
(I don't know about all switches but probably no more than 5min)

Each pipeline pushes the image to $CI_REGISTRY_IMAGE:$CI_PIPELINE_ID
which is eg registry.gitlab.com/skyskimmer/coq:21557176 and is what
the other jobs use to run themselves.

For caching it pulls from and pushes to a constant name, in this test
$CI_REGISTRY_IMAGE:$CACHEKEY. We might also want to pull from the main
coq repo to avoid redoing work.

The question of having 1 image or 1 image/switch remains open.

Using the gitlab registry doesn't really work for circle CI, so the
dockerhub account would have to remain (or circle could return to an
opam-boot job, or be removed from our CI). There are similar concerns
with travis if we want to move it to docker.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -28,6 +28,9 @@ docker-boot:
     - if docker pull "$IMAGE"; then echo "Image prebuilt!"; exit 0; fi
     - docker build -t "$IMAGE" .
     - docker push "$IMAGE"
+  except:
+    variables:
+      - $SKIP_DOCKER == "true"
 
 before_script:
   - cat /proc/{cpu,mem}info || true
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,6 +1,7 @@
-image: coqci/base:V2018-05-07-V2
+image: "$IMAGE"
 
 stages:
+  - docker
   - build
   - test
 
@@ -9,10 +10,24 @@ variables:
   # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
   # for reference]
   CACHEKEY: bionic_coq-V2018-05-07-V2
+  IMAGE: "$CI_REGISTRY_IMAGE:$CACHEKEY"
   # By default, jobs run in the base switch; override to select another switch
   OPAM_SWITCH: "base"
   # Used to select special compiler switches such as flambda, 32bits, etc...
   OPAM_VARIANT: ""
+
+docker-boot:
+  stage: docker
+  image: docker:stable
+  services:
+    - docker:dind
+  before_script: []
+  script:
+    - docker login -u gitlab-ci-token -p "$CI_JOB_TOKEN" "$CI_REGISTRY"
+    - cd dev/ci/docker/bionic_coq/
+    - if docker pull "$IMAGE"; then echo "Image prebuilt!"; exit 0; fi
+    - docker build -t "$IMAGE" .
+    - docker push "$IMAGE"
 
 before_script:
   - cat /proc/{cpu,mem}info || true
--- a/dev/ci/README.md
+++ b/dev/ci/README.md
@@ -96,7 +96,8 @@ PR by running GitLab CI on your private branches. To do so follow these steps:
 2. Click on "New Project".
 3. Choose "CI / CD for external repository" then click on "GitHub".
 4. Find your fork of the Coq repository and click on "Connect".
-5. You are encouraged to go to the CI / CD general settings and increase the
+5. If GitLab did not do so automatically, [enable the Container Registry](https://docs.gitlab.com/ee/user/project/container_registry.html#enable-the-container-registry-for-your-project).
+6. You are encouraged to go to the CI / CD general settings and increase the
    timeout from 1h to 2h for better reliability.
 
 Now everytime you push (including force-push unless you changed the default
@@ -137,3 +138,27 @@ Currently, available artifacts are:
 
 As an exception to the above, jobs testing that compilation triggers
 no OCaml warnings build Coq in parallel with other tests.
+
+### GitLab and Windows
+
+
+If your repository has access to runners tagged `windows`, setting the
+secret variable `WINDOWS` to `enabled` will add jobs building Windows
+versions of Coq (32bit and 64bit).
+
+The Windows jobs are enabled on Coq's repository, where pipelines for
+pull requests run.
+
+### GitLab and Docker
+
+System and opam packages are installed in a Docker image. The image is
+automatically built and uploaded to your GitLab registry, and is
+loaded by subsequent jobs.
+
+The Docker building job reuses the uploaded image if it is available,
+but if you wish to save more time you can skip the job by setting
+`SKIP_DOCKER` to `true`.
+
+This means you will need to change its value when the Docker image
+needs to be updated. You can do so for a single pipeline by starting
+it through the web interface.


ejgallego:
I understand there is not a safe way to push to `hub.docker` even for a few contributors?

ejgallego:
> What's your threat model?

A non-authorized person overwriting a tag there.

Ideally, I'd be cool if I could set up a secret on my own Gitlab repo and that secret would be used in a PR to push to `coqci` [as well as to the Gitlab registry], but I don't know how Gitlab secrets work [yet].

Other than that, I like this patch, will review later today or tomorrow I hope.

SkySkimmer:
Better question, what's your use case?
Putting a secret in a protected variable on gitlab shares it only with protected branches (ie master) so that would work, but I don't know if it's useful for what you want.

ejgallego:
> Better question, what's your use case?

Use case is as follows:
1. I update dockerfile.
2. I submit a pull request -> that will trigger a pull request.
3. The CI system detects that I have push rights to `hub.docker.com/coqci`, then it pushes it there.

ejgallego:
I think this looks good.

ejgallego:
On the other hand I am not sure this is worth it. How often are we supposed to update the image?

I would say we would update the Docker image very very infrequently, basically only on OCaml releases or maybe new dependencies coming in. Let's say once in a month.

On the other hand, this will add a 3 / 4 minutes job for **all** pipelines. Let's say we have 15 pipelines per day, that's 15*3 = 45 minutes * 30 days, around 1 full day (24h) of time lost per month.

I am not sure it is worth it IMO.

ejgallego:
Ideally we'd have this job only run when the branch has a special name [we'd need to improve the bot tho]

ejgallego:
Yeah I have been thinking a bit more about this and IMHO generating a tag for each build seems a bit too much, I am not sure Docker is really designed for that, maybe yes.

Also, I don't see the problem with Circle using gitlab's registry, `docker pull registry.gitlab.com/coq/coq:bionic-V2018-05-06` should work well there, right?

ejgallego:
So this is what I suggest for now, we add to the job:
```yaml
only:
  - docker-update
```
and we create such branch in the gitlab repository; so any time we want to update the image we have to push to that branch, then do a pull request. WDYT?

Zimmi48:
I am very much in favor of this PR because we should avoid to depend on too many external services (like DockerHub). Furthermore, the new Docker based solution is making life harder for new contributors (cf. #6808) so this PR is a big improvement.

However, I agree with @ejgallego that we should try to avoid the wasted build time (if we can). Would it be quicker to have a solution based on whether git detects some change to the Docker image code? Or based on a label put on the PR? Or a comment in a git commit? In any case, I absolutely want to avoid the branch-name-based solution proposed by @ejgallego because branch names can't be changed for already opened PRs and I would prefer to reduce any requirement of a manual action.

ejgallego:
> a label put on the PR?

That's almost the same than having a regexp for branch name. I don't think that's so bad.

In the end I think that caching will always make the life of new contributors difficult as it involves some trust [or less of caching], so what the right tradeoff is, I don't know.

ejgallego:
> Furthermore, the new Docker based solution is making life harder for new contributors (cf. #6808) so this PR is a big improvement.

IMHO this event "a new contributor needing to a dependency to the base build system" should be quite rare, of course if it turns out that this is not the case we should reevaluate.

Note that the contributor can also build and push a Docker image without too much problems; only pain is the lack of a single place to put the tag.

ejgallego:
> No because you can set a label on an open PR, but you can't change the branch name.

So I dunno how this would work? Maybe the bot could map the label to a branch name?

The issue is that Gitlab's conditions don't know about labels.

SkySkimmer:
We could do
```yml
only:
  - trigger
  - web
```
Then a user can rebuild by starting the pipeline through the web UI, and we can also have a bot watching for github labels or whatever we want.

ejgallego:
> Then a user can rebuild by starting the pipeline through the web UI, and we can also have a bot watching for github labels or whatever we want.

That'd be also fine to me.

Note that this PR in its current state would be also fine to me, but I'd like more if we could indeed save that boot step.

SkySkimmer:
>Yeah I have been thinking a bit more about this and IMHO generating a tag for each build seems a bit too much, I am not sure Docker is really designed for that, maybe yes.

I think docker is fine with it but it's not super useful for us so I changed it to using only CACHEKEY.

Zimmi48:
I think we are thinking too hard about the build time problem because if we want to save precious minutes of build times, we should start elsewhere (e.g. not recompile Bignums + Math-Classes + Corn several times).

ejgallego:
> I think docker is fine with it but it's not super useful for us so I changed it to using only CACHEKEY.

My worries were indeed about garbage collection, it wouldn't take long for the registry to have thousand of tags, and I dunno how you can clean that up. Would the regsitry support something such as "expiry" I'd be more comforable with it.

ejgallego:
> I think we are thinking too hard about the build time problem because if we want to save precious minutes of build times, we should start elsewhere (e.g. not recompile Bignums + Math-Classes + Corn several times).

We try to do as best as we can latency-wise, and given the limitations of Gitlab. So far what we have here is close to optimal latency-wise, how would you layout things?

SkySkimmer:
>My worries were indeed about garbage collection, it wouldn't take long for the registry to have thousand of tags, and I dunno how you can clean that up. Would the regsitry support something such as "expiry" I'd be more comfortable with it.

Actually I hit delete on one of the tags for the coq/coq registry and it removed all the aliases :v

SkySkimmer:
>We try to do as best as we can latency-wise, and given the limitations of Gitlab. So far what we have here is close to optimal latency-wise, how would you layout things?

We could remove the mathclasses job since it gets built for formal-topology

ejgallego:
> Actually I hit delete on one of the tags for the coq/coq registry and it removed all the aliases :v

I don't know if that's good or terrible.

ejgallego:
Adding a tag as to know when to look at this again. My opinion: we want this in 100%, but it would be great to implement an `only` condition so we save the boot step in the large majority of jobs.

SkySkimmer:
What kind of `only` do you want?

ejgallego:
> What kind of only do you want?

Your only proposal looked fine to me, any other method that won't trigger the rebuild job always will be OK to me. If you don't like `only` we could even merge this as is, I dunno, WDYT?

I do prefer to save that 5 minutes, IMHO that's a lot of latency overall for what it will be 2/3 Docker updates per year.

.gitlab-ci.yml
```diff
@@ -1,22 +1,40 @@
...
+  before_script: []
+  script:
+    - docker login -u gitlab-ci-token -p "$CI_JOB_TOKEN" "$CI_REGISTRY"
+    - cd dev/ci/docker/bionic_coq/
+    - if docker pull "$IMAGE"; then echo "Image prebuilt!"; exit 0; fi
+    - docker build -t "$IMAGE" .
+    - docker push "$IMAGE"
+  except:
+    variables:
+      - $SKIP_DOCKER == "true"
```
SkySkimmer: After thinking about it I want the default to just work and that means always running.
So we use a variable to disable the job (we can set the variable in the web UI when manually running to differ from repo policy).

SkySkimmer:
I removed all the extra `bionic_coq` from paths and variables as they're useless and would be a pain to sync when we upgrade.

Zimmi48:
Is the postpush hook still used with this configuration?

SkySkimmer:
>Is the postpush hook still used with this configuration?

Circle still uses dockerhub.

ejgallego:
> We can add linux when we have more than 1 dockerfile but bionic is no good.

Why is not good?

I can live with the current `SKIP_DOCKER` policy, however I'm not sure I can take a generic naming, specially when we are open to adding nix images, etc...

SkySkimmer:
It should be auto enabled.

Gaëtan Gilbert

SkySkimmer:
OK good point, I'll revert

Gaëtan Gilbert

ejgallego (changes requested):
I think documentation `dev/ci/README.md` should be updated to reflect the variable policy.

I'd suggest adding a section "gitlab configuration variables". Windows variables should be there too.

SkySkimmer:
In what way?

Gaëtan Gilbert

SkySkimmer:
>I think documentation dev/ci/README.md should be updated to reflect the variable policy.

Changed.

dev/ci/README.md
```diff
@@ -137,3 +137,26 @@ Currently, available artifacts are:
 
 As an exception to the above, jobs testing that compilation triggers
 no OCaml warnings build Coq in parallel with other tests.
+
```
ejgallego: You need to update the above section to instruct users to enable their private registry as otherwise Gitlab will fail now.

dev/ci/README.md
```diff
@@ -137,3 +137,26 @@ Currently, available artifacts are:
...
+System and opam packages are installed in a Docker image. The image is
+automatically built and uploaded to your GitLab registry, and is
+loaded by subsequent jobs.
+
+The Docker building job reuses the uploaded image if it is available,
+but if you wish to save more time you can skip it by setting
+`SKIP_DOCKER` to `true`.
+
+This means you will need to change its value when the Docker image
```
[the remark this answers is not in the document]
SkySkimmer: >"change its value back to false"

false isn't a default, any non `true` value works.

SkySkimmer:
Changed further.

ejgallego (approved):
It looks OK to me so I'll merge once CI gives a positive answer.

ejgallego:
I am triggering the build with the proper thing set, let's see...

ejgallego:
https://gitlab.com/coq/coq/pipelines/22002967

ejgallego:
Umm, why is circle not switched?

AFAICS there is no problem in circle pulling from Docker's registry; so at least let's make its config point to the new official home for the Docker image.

Zimmi48:
It would have been less complications if CircleCI had been removed completely.

ejgallego:
> It would have been less complications if CircleCI had been removed completely.

There is not rush; it is easy to modifiy a line.

ejgallego:
@SkySkimmer going to merge this soon, but what should we go with Circle?

Do you want to address it in a separate PR? If you want to push the change, rebasing should help with build time thanks to #7503 having been merged.

ejgallego:
Ok let's get this going and address Circle on a different PR.

Zimmi48:
This PR also doesn't update the README about docker. This should probably be done when this is stabilized.

PR #7446, opened 2018-05-07
[lib] Re-add `set_timeout` to help users workaround #7408

ejgallego:
It seems like #7408 will need some potentially intrusive work, so
let's add the low-level hook back so third party developments can work
well with `8.8.1/master` for the moment.


ppedrot (approved):
LGTM.

Issue #7265, opened 2018-04-16
Definition with wildcards generates anomaly "grounding a non evar-free term"


anton-trunov:
#### Version

Coq master (3e7863e9369d38537685576a8642dbe0c062d0c5)

#### Operating system

macOS 10.13.3 High Sierra

#### Description of the problem

Please consider the following minimized example:

```coq
Definition foo : _ := _.
```

It generates the following anomaly:

>Error: Anomaly "grounding a non evar-free term"
>Please report at http://coq.inria.fr/bugs/.

It's a regression since Coq 8.7.2, which correctly reports:

>Error: Cannot infer this placeholder of type "Type".

ejgallego:
This is due to #6454 , and due to code using `EConstr.to_constr` incorrectly without properly flagging it. Will fix. [Not an 8.8 regression tho]

maximedenes:
@ejgallego any update on this?

ejgallego:
Didn't have time to look into this, anyways running `-debug` on should point the location and the problem should be easy to fix.

anton-trunov:
current Coq master
```
$ bin/coqtop -debug
Welcome to Coq 192.168.1.136:/Users/anton/prj/coq/coq,master (fd146ca38202c9843b4240cbdac0ae75f57e4d67)

Coq < Definition foo : _ := _.
Toplevel input, characters 0-24:
> Definition foo : _ := _.
> ^^^^^^^^^^^^^^^^^^^^^^^^
Error: Anomaly "in econstr: grounding a non evar-free term"
Please report at http://coq.inria.fr/bugs/.
frame @ file "toplevel/vernac.ml", line 172, characters 2-60
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "stm/stm.ml", line 600, characters 33-68
raise @ file "pervasives.ml", line 32, characters 17-33
frame @ file "stm/stm.ml", line 2629, characters 4-60
frame @ file "stm/stm.ml", line 2519, characters 4-105
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "stm/stm.ml", line 935, characters 6-10
frame @ file "stm/stm.ml", line 2392, characters 16-43
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "stm/stm.ml", line 1088, characters 12-91
raise @ file "clib/exninfo.ml", line 65, characters 2-15
[...]

PR #7449, opened 2018-05-07
[vernac] taint two out-of-api `to_constr` use in `comDefinition`.

### LINKED ISSUE #7448: New anomaly with Program Definition
#### Version

8.9+alpha (6c8b00e47334f60f200256d45a5542fa80ce4b12)

```
Require Program.Tactics.
Program Definition foo := _.
```

gives:

```
Anomaly "grounding a non evar-free term"
```

In `v8.8` (0c88f3b50d6e213fc7bfb02525c4e349a630275c and 8.8.0), it gives:

```
Cannot infer an internal placeholder of type "Type".
```

Putting the 8.9+beta1 milestone because someone should probably self-assign this.

ejgallego:
This should fix #7448 and #7265.

### DIFF
--- a/vernac/comDefinition.ml
+++ b/vernac/comDefinition.ml
@@ -90,7 +90,7 @@ let interp_definition pl bl poly red_option c ctypopt =
      Note: in program mode some evars may remain. *)
   let ctx = List.map Termops.(map_rel_decl (to_constr ~abort_on_undefined_evars:false evd)) ctx in
   let c = Term.it_mkLambda_or_LetIn (EConstr.to_constr ~abort_on_undefined_evars:false evd c) ctx in
-  let tyopt = Option.map (fun ty -> Term.it_mkProd_or_LetIn (EConstr.to_constr evd ty) ctx) tyopt in
+  let tyopt = Option.map (fun ty -> Term.it_mkProd_or_LetIn (EConstr.to_constr ~abort_on_undefined_evars:false evd ty) ctx) tyopt in
   (* Keep only useful universes. *)
   let uvars_fold uvars c =
     Univ.LSet.union uvars (universes_of_constr env evd (of_constr c))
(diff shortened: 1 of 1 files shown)


ejgallego:
Why is this not asking the codeowner to do the review @maximedenes @Zimmi48

maximedenes:
The codeowners file is now fixed. This PR is pending on a review by @mattam82.

ejgallego:
Note that in this case CODEOWNERS doesn't do the best job of identifying the best reviewer in the sense that here the "right" person to review this kind of change is the maintainer of `EConstr`; we have this problem in other areas of the codebase where a patch modifying how an API is used should be reviewed by the API implementer not the user.

maximedenes:
I agree it is a problem that exists, but I'm not so sure it is the case here. This PR is about a newly introduce invariant not respected by a piece of code. It looks ok to me to have the maintainer of that piece of code review it, not necessarily the maintainer of the code that enforces the invariant.

ejgallego:
Well indeed both should have a look to this, as it is requested. Maybe in fact we should even open a bug about the implementation of `comDefinition` et al.

ppedrot (commented):
This is merely side-stepping the problem, which is that Program is misusing evar-bearing terms. The current solution is thus a quick workaround, but a proper fix should be designed at some point.

maximedenes:
@ppedrot Do we have an issue for the program-with-evars pb? If yes, can you give a pointer? If not, can you open it (you probably know better what has to be done)? Thanks!

ejgallego:
Indeed we should open the issue. Given that @mattam82 has hardware problems and this is a trivial fix to a bug impacting a couple of users I'd suggest that either @ppedrot or @maximedenes take care of it.

ppedrot (approved):
In case this was not clear, I'm fine with merging this PR as-is.

ppedrot:
@mattam82 Are you fine with this PR?

PR #7450, opened 2018-05-07
Remove an internal function from an mli.

Zimmi48:
This trivial commit removes a function from an mli that wasn't used anywhere. Are these kinds of clean-ups wanted or not?

### DIFF
--- a/tactics/tactics.ml
+++ b/tactics/tactics.ml
@@ -4927,6 +4927,7 @@ let shrink_entry sign const =
   } in
   (const, args)
 
+(** Internal function only used by tclABSTRACT *)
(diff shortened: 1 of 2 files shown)


ejgallego:
I think they are very valuable as IMHO API size is a concern.

Zimmi48:
Again, the CODEOWNER feature does not seem to work.

Zimmi48:
So it seems that this function is actually used by fiat-parsers.

ppedrot:
This function was introduced by @JasonGross to be used in a plugin. I'd think he's still using it...

Zimmi48:
OK I should have checked the blame :(

JasonGross:
I believe this is still used in fiat.

Issue #7433, opened 2018-05-05
[sphinx] .. coqtop:: should be used with reset setting more often


Zimmi48:
#### Version

8.8.0

#### Description of the problem

The small discussion here https://github.com/coq/coq/pull/7391#discussion_r185034208 made me realize that a lot of `Reset Initial` were wrongly removed during the Sphinx migration (not converted to `.. coqtop:: reset`. It's not obvious because of the (stupid) nested lemma feature of Coq. But it still appears from time to time, for instance here https://coq.inria.fr/refman/proof-engine/tactics.html#coq:tacn.decompose (after `Qed` we can see that there is a subgoal left to prove).

I discourage working on this before #7301 and #7374 are merged because of the huge risk of conflicts.

Zimmi48:
It would certainly help to make progress on this if we had a solution to #7755.

Zimmi48:
Not an actual blocker for a beta release.

Zimmi48:
Fixed

PR #7170, opened 2018-04-04
Script to identify the code owner for given files

SkySkimmer:
Can be used with git diff --name-only to identify owners for changes
in given commits.

Example output (commit range f29f8f80c8ad94576c7a36f3f638866c208338a0 to this commit):
```
toplevel/coqloop.ml: @ejgallego
toplevel/g_toplevel.ml4: @ejgallego
stm/stm.ml: @gares
stm/vernac_classifier.ml: @gares
parsing/g_vernac.ml4: @herbelin
printing/ppvernac.ml: @herbelin
intf/vernacexpr.ml: @letouzey
Makefile.doc: @letouzey
vernac/vernacentries.ml: @mattam82
vernac/vernacprop.ml: @mattam82
doc/refman/Coercion.tex: @maximedenes
doc/refman/Extraction.tex: @maximedenes
doc/refman/Nsatz.tex: @maximedenes
doc/refman/Polynom.tex: @maximedenes
doc/refman/Program.tex: @maximedenes
doc/refman/Reference-Manual.tex: @maximedenes
doc/refman/Setoid.tex: @maximedenes
doc/sphinx/addendum/extraction.rst: @maximedenes
doc/sphinx/addendum/generalized-rewriting.rst: @maximedenes
doc/sphinx/addendum/implicit-coercions.rst: @maximedenes
doc/sphinx/addendum/nsatz.rst: @maximedenes
doc/sphinx/addendum/program.rst: @maximedenes
doc/sphinx/addendum/ring.rst: @maximedenes
doc/sphinx/index.rst: @maximedenes
doc/sphinx/language/gallina-extensions.rst: @maximedenes
dev/tools/check-owners.sh: Nobody
test-suite/coq-makefile/timing/run.sh: Nobody
```

PR #7451, opened 2018-05-07
Introduce an option to allow nested lemma, and turn it off by default.

Zimmi48:
**Kind:** enhancement.

- [x] Corresponding documentation was added / updated.
- [x] Entry added in CHANGES.

The nested lemma feature has been bugging me for a while. It is a source of confusion for beginners who forget a `Qed`:

```
Lemma foo : True.
easy.
(* No more subgoals. *)

Lemma bar : True.
apply foo.
(* The reference foo was not found in the current environment. *)
```

It is also the reason why #7433 was not detected sooner.

Independently of these reasons, it has been deprecated for a while *but* we cannot actually remove the feature because lots of experimented users are actually relying on it.

So instead, what I propose in this PR is to create a new option to control whether the nested lemma feature is allowed or not. And turn this option off by default (this is the whole point). We could also (not necessarily, and not done currently in this PR) remove the deprecation warning... When we really intend to remove the feature, we can mark the option itself as deprecated (again, not done currently).

So what do people think of this proposal?

Implementation side, I was surprised that this was more difficult than I expected. I am still not sure I made the right choices so please review. I am not sure of the location of the option in particular (at the beginning it was next to the function raising the error message in `proof_global`).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,9 @@ Vernacular Commands
 
 - Removed deprecated commands Arguments Scope and Implicit Arguments
   (not the option). Use the Arguments command instead.
+- Nested proofs may be enabled through the option `Nested Proofs Allowed`.
+  By default, they are disabled and produce an error. The deprecation
+  warning which used to occur when using nested proofs has been removed.
 
 Tactics
 
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -2768,7 +2768,7 @@ let allow_nested_proofs = ref false
 let _ = Goptions.declare_bool_option
     { Goptions.optdepr  = false;
       Goptions.optname  = "Nested Proofs Allowed";
-      Goptions.optkey   = ["Nested";"Proofs";"Allowed"];
+      Goptions.optkey   = Vernac_classifier.stm_allow_nested_proofs_option_name;
       Goptions.optread  = (fun () -> !allow_nested_proofs);
       Goptions.optwrite = (fun b -> allow_nested_proofs := b) }
 
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -2113,12 +2113,6 @@ let delegate name =
   || VCS.is_vio_doc ()
   || !cur_opt.async_proofs_full
 
-let warn_deprecated_nested_proofs =
-  CWarnings.create ~name:"deprecated-nested-proofs" ~category:"deprecated"
-         (fun () ->
-          strbrk ("Nested proofs are deprecated and will "^
-                    "stop working in a future Coq version"))
-
 let collect_proof keep cur hd brkind id =
  stm_prerr_endline (fun () -> "Collecting proof ending at "^Stateid.to_string id);
  let no_name = "" in
@@ -2200,8 +2194,7 @@ let collect_proof keep cur hd brkind id =
         assert (VCS.Branch.equal hd hd' || VCS.Branch.equal hd VCS.edit_branch);
         let name = name ids in
         `MaybeASync (parent last, accn, name, delegate name)
-    | `Sideff _ ->
-       warn_deprecated_nested_proofs ();
+    | `Sideff (CherryPickEnv,_) ->
         `Sync (no_name,`NestedProof)
     | _ -> `Sync (no_name,`Unknown) in
  let make_sync why = function
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -2771,6 +2771,14 @@ let process_back_meta_command ~newtip ~head oid aast w =
     VCS.commit id (Alias (oid,aast));
     Backtrack.record (); if w == VtNow then ignore(finish ~doc:dummy_doc); `Ok
 
+let allow_nested_proofs = ref false
+let _ = Goptions.declare_bool_option
+    { Goptions.optdepr  = false;
+      Goptions.optname  = "Nested Proofs Allowed";
+      Goptions.optkey   = ["Nested";"Proofs";"Allowed"];
+      Goptions.optread  = (fun () -> !allow_nested_proofs);
+      Goptions.optwrite = (fun b -> allow_nested_proofs := b) }
+
 let process_transaction ~doc ?(newtip=Stateid.fresh ())
   ({ verbose; loc; expr } as x) c =
   stm_pperr_endline (fun () -> str "{{{ processing: " ++ pr_ast x);
@@ -2802,6 +2810,15 @@ let process_transaction ~doc ?(newtip=Stateid.fresh ())
 
       (* Proof *)
       | VtStartProof (mode, guarantee, names), w ->
+
+         if not !allow_nested_proofs && VCS.proof_nesting () > 0 then
+           "Nested proofs are not allowed unless you turn option Nested Proofs Allowed on."
+           |> Pp.str
+           |> (fun s -> (UserError (None, s), Exninfo.null))
+           |> State.exn_on ~valid:Stateid.dummy Stateid.dummy
+           |> Exninfo.iraise
+         else
+
           let id = VCS.new_node ~id:newtip () in
           let bname = VCS.mk_branch_name x in
           VCS.checkout VCS.Branch.master;
--- a/stm/vernac_classifier.ml
+++ b/stm/vernac_classifier.ml
@@ -54,13 +54,20 @@ let idents_of_name : Names.Name.t -> Names.Id.t list =
   | Names.Anonymous -> []
   | Names.Name n -> [n]
 
+let stm_allow_nested_proofs_option_name = ["Nested";"Proofs";"Allowed"]
+
+let options_affecting_stm_scheduling =
+  [ Vernacentries.universe_polymorphism_option_name;
+    stm_allow_nested_proofs_option_name ]
+
 let classify_vernac e =
   let static_classifier ~poly e = match e with
     (* Univ poly compatibility: we run it now, so that we can just
      * look at Flags in stm.ml.  Would be nicer to have the stm
      * look at the entire dag to detect this option. *)
     | ( VernacSetOption (_, l,_) | VernacUnsetOption (_, l))
-      when CList.equal String.equal l Vernacentries.universe_polymorphism_option_name ->
+      when CList.exists (CList.equal String.equal l)
+        options_affecting_stm_scheduling ->
        VtSideff [], VtNow
     (* Qed *)
     | VernacAbort _ -> VtQed VtDrop, VtLater
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -1339,7 +1339,8 @@ using the keyword :cmd:`Qed`.
 
 .. note::
 
-   #. Several statements can be simultaneously asserted.
+   #. Several statements can be simultaneously asserted provided option
+      :opt:`Nested Proofs Allowed` was turned on.
 
    #. Not only other assertions but any vernacular command can be given
       while in the process of proving a given assertion. In this case, the
--- a/doc/sphinx/proof-engine/proof-handling.rst
+++ b/doc/sphinx/proof-engine/proof-handling.rst
@@ -113,6 +113,8 @@ list of assertion commands is given in :ref:`Assertions`. The command
       Aborts the editing of the proof named :token:`ident` (in case you have
       nested proofs).
 
+      .. seealso:: :opt:`Nested Proofs Allowed`
+
    .. cmdv:: Abort All
 
       Aborts all current goals.
@@ -559,6 +561,17 @@ Controlling the effect of proof editing commands
    When the option is off, binders are discharged on the statement to be
    proved and a tactic such as :tacn:`intro` (see Section :ref:`managingthelocalcontext`)
    has to be used to move the assumptions to the local context.
+
+
+.. opt:: Nested Proofs Allowed
+
+   When turned on (it is off by default), this option enables support for nested
+   proofs: a new assertion command can be inserted before the current proof is
+   finished, in which case Coq will temporarily switch to the proof of this
+   *nested lemma*. When the proof of the nested lemma is finished (with :cmd:`Qed`
+   or :cmd:`Defined`), its statement will be made available (as if it had been
+   proved before starting the previous proof) and Coq will switch back to the
+   proof of the previous assertion.
 
 
 Controlling memory usage
--- a/stm/vernac_classifier.mli
+++ b/stm/vernac_classifier.mli
@@ -25,3 +25,4 @@ val classify_as_query : vernac_classification
 val classify_as_sideeff : vernac_classification
 val classify_as_proofstep : vernac_classification
 
+val stm_allow_nested_proofs_option_name : string list
--- a/test-suite/bugs/closed/2969.v
+++ b/test-suite/bugs/closed/2969.v
@@ -12,6 +12,7 @@ eexists.
 reflexivity.
 Grab Existential Variables.
 admit.
+Admitted.
 
 (* Alternative variant which failed but without raising anomaly *)
 
@@ -24,3 +25,4 @@ clearbody n n0.
 exact I.
 Grab Existential Variables.
 admit.
+Admitted.
--- a/test-suite/bugs/closed/3377.v
+++ b/test-suite/bugs/closed/3377.v
@@ -5,6 +5,7 @@ Record prod A B := pair { fst : A; snd : B}.
 Goal fst (@pair Type Type Type Type).
 Set Printing All.
 match goal with |- ?f ?x => set (foo := f x) end.
+Abort.
 
 Goal forall x : prod Set Set, x = @pair _ _ (fst x) (snd x).
 Proof.
@@ -12,6 +13,6 @@ Proof.
   lazymatch goal with
     | [ |- ?x = @pair _ _ (?f ?x) (?g ?x) ] => pose f
   end.
-
 (* Toplevel input, characters 7-44:
 Error: No matching clauses for match. *)
+Abort.
--- a/test-suite/bugs/closed/4069.v
+++ b/test-suite/bugs/closed/4069.v
@@ -41,6 +41,8 @@ Proof. f_equal.
   8.5: 2 goals, skipn n l = l -> k ++ skipn n l = skipn n l
     and skipn n l = l
 *)
+Abort.
+
 Require Import List.
 Fixpoint replicate {A} (n : nat) (x : A) : list A :=
   match n with 0 => nil | S n => x :: replicate n x end.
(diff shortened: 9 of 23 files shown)


Zimmi48:
The code owner feature is definitely broken. Fortunately, we have @SkySkimmer tool in #7170 which gives:

```
vernac/lemmas.mli: @mattam82
vernac/lemmas.ml: @mattam82
proofs/pfedit.mli: @ppedrot
proofs/pfedit.ml: @ppedrot
proofs/proof_global.mli: @ppedrot
proofs/proof_global.ml: @ppedrot
```

SkySkimmer:
Why make the argument optional?

Zimmi48:
To do a lighter refactoring. But if you think it should be mandatory, I can adapt.

SkySkimmer:
I certainly want to know which places if any need `allow_nested:true`.

Zimmi48:
OK, I'll change this (not right now though).

gares:
> So what do people think of this proposal?

thanks!

ejgallego (commented):
I am not sure what to think of this:
- the good side is that it would remove lots of code in the current implementation,
- on the other hand, I think that supporting nested lemmas doesn't pose any particular technical problem, and the feature does indeed seem to make sense to quite a few experienced users. [The first particular technical problem is the duplication of functionalities among the STM and `Proof_global` which instead of working together fight against each other].

This last point brings us to a question on where to hook the warning, @Zimmi48 has chosen [_wisely IMHO_] to hook it in the execution layer, however this implies some particular tradeoffs, it could be indeed interesting to put the warning in the STM.

Zimmi48:
@ejgallego It seems that the first part of your review comment is discussing whether to remove the nested lemma feature eventually. This PR does not take any side on this question.

maximedenes:
> It seems that the first part of your review comment is discussing whether to remove the nested lemma feature eventually. This PR does not take any side on this question.

I think that's precisely the problem. I'm not in favor of introducing a flag if the plan is to make it stay. For example, this is going to make testing nested lemma more difficult.

That said, I'm in favor of removing the nested lemmas feature, after addressing the various user needs that they seem to meet today. The first step would probably be to provide a decent asynchronous IDE. On the other hand, I wouldn't cry if they stayed, if properly supported.

So in short, if we agree to remove nested lemmas in the long term, I'm in favor of this PR, otherwise I'm against.

Zimmi48:
> Because you need to test interactions between features. Adding flags is an obvious source of exponential growth in the number of test cases needed.

Oh but here the flag does not create two widely different behaviors that would need to be tested against each orthogonal feature, it's more like a warning. So I don't think that this (otherwise valid) argument applies. In particular, having the feature on by default doesn't mean that we properly test it.

maximedenes:
> the flag does not create two widely different behaviors that would need to be tested against each orthogonal feature

Ok, maybe you can argue that this is a special case. The `Universe Polymorphic` flag you took as an example definitely does not fall in this category, though.

Zimmi48:
Indeed, that was a pretty bad comparison.

ejgallego:
Maybe I am confused as to why do you want to turn this option on by default. This IMHO looks as a "Coq 9" breakage-level, quite a few users do have nested stuff.

ejgallego:
In that sense my point of view is that I tend to consider the current warning the one that should go, but YMMV.

ejgallego:
I see @Zimmi48 , umm, IMVVHO I think that the plan for the future of proof scripts themselves is not very explicitly stated yet, so I am not sure that doing this change without discussing a bit more about this is the best action path.

IMHO I would just wait for the two upcomign WG/CIM and see where are we going on the proof script sense.

Zimmi48:
> so I am not sure that doing this change without discussing a bit more about this is the best action path.

Again, the goal of this PR was not to take a side on the debate of whether nested lemmas should go or not. What is the problem of considering this an expert-only feature and hiding it behind an option? It's not much different from having to do `Require Extraction` to use extraction. The error message even provides the fix.

ejgallego:
IMHO setting the option OFF by default looks a bit like gratuitous breakage, specially if after discussion we decide to keep nested lemmas in. To me, the default only makes sense if we are set on having nested lemmas go, and IMHO we should discuss that a bit more if possible.

Zimmi48:
It's not gratuitous. The goal is to solve a real issue I have been confronted to several times which is that when people forget a `Qed` this feature prevents them from having any warning or error message. That being said, the balance is tipping against this change so I'm not going to work more on this PR unless I get real support (for now, the only support I got was from @gares).

Zimmi48:
Having this option on by default would (AFAICS) have zero utility.

SkySkimmer:
I very approve of disabling nesting by default.

ejgallego:
> I'm also supporting the fact that nested lemmas should be disabled by default.

What are the problems @ppedrot you see with nested lemmas [honest question, I'm curious]

I'd dare to say that in maths nested / auxiliary lemmas are quite common, thus hence my reluctance.

maximedenes:
> I'd dare to say that in maths nested / auxiliary lemmas are quite common, thus hence my reluctance.

Auxiliary, obviously. But nested, do you have an example?

Zimmi48:
> I am not "against" this change per se, but I'd like to avoid the situation where we push the change, then we declare that nested lemmas are OK.

Once again, it looks like I failed to get that clear: this PR is not taking any side on whether nested lemmas are a good or a bad feature. It is just classifying it as an "advanced" feature that is fine to disable by default (again, like Extraction for instance).

Zimmi48:
But I'm OK with alternative syntax proposal if you prefer this. This would indeed solve my problem.

ejgallego:
> Auxiliary, obviously. But nested, do you have an example?

I fail to locate from memory any paper that is not written by me and states a lemma inside a proof, but I'd dare to say there are quite a few.

ppedrot:
> What are the problems @ppedrot you see with nested lemmas [honest question, I'm curious]

I've also been bitten by the same kind of problems that is mentioned in the first post, i.e. an unclosed proof. Although a trivial problem, I find it rather frustrating that Coq cannot tell me I forgot something.

> I'd dare to say that in maths nested / auxiliary lemmas are quite common, thus hence my reluctance.

Nested lemmas are actually not nested, because they're accessible from the outside. In Coq they're just a syntactic artifact, they don't bring anything at deployment time. While writing a proof it's another story, but then I believe it's mostly an UI issue.

ejgallego:
> It is just classifying it as an "advanced" feature that is fine to disable by default

Yeah in a sense I am questioning if making the vernacular conditional on an "advanced" flag is the way to go, here indeed we'd face indeed option creep. I'd say that I'd prefer a vernacular with few options like those. So we either fully embrace such style of proof, or we get rid of it.

> (again, like Extraction for instance).

IMHO the comparison is not fully fair, in the sense that extraction was removed from the init file as to make startup lighter, but this change doesn't do anything like that.

ejgallego:
So indeed the main problem to solve here seems to be bad parenthezization.

> they don't bring anything at deployment time.

The only benefit is for authors to organize their development like that.

> While writing a proof it's another story, but then I believe it's mostly an UI issue.

Well, in a sense the "I forgot to close a proof" issue seems indeed very close to be an UI issue too.

By the way, I think that true "nested" lemmas are pervasive in developments such as mathcomp, it is just that they use `have` of `suff` instead of the Coq machinery.

ejgallego:
> By the way, I think that true "nested" lemmas are pervasive in developments such as mathcomp, it is just that they use have of suff instead of the Coq machinery.

So let me try to rephrase my worry a bit more once that I have mentioned this, and independently of this proposal that maybe is the right thing to do.

"How would we like the document vernacular (*) to reflect the internal proof structure?"

Indeed, it would make sense to me, that the vernacular would provide structure for subproofs `have`, `suff`, and `case` analysis, side obligations at least. But indeed the current sub lemma support is not bringing anything very useful here, IMHO.

So for me parenthesization is a valid reason to disable "nesting" of lemmas, however the argument that such pattern cannot be properly supported (used in the past) is not a valid one AFAIAC.

(*) Document vernacular here == structure the UI understands.

Zimmi48:
> By the way, I think that true "nested" lemmas are pervasive in developments such as mathcomp, it is just that they use have of suff instead of the Coq machinery.

That's not the same at all as these depend on the local context. And vanilla Coq supports this too with `assert` and `enough`.

> So for me parenthesization is a valid reason to disable "nesting" of lemmas, however the argument that such pattern cannot be properly supported (used in the past) is not a valid one AFAIAC.

Thanks. Indeed, I wasn't using this other argument.

ejgallego:
> Thanks. Indeed, I wasn't using this other argument.

Yeah so indeed IIUC the deprecation was done for that other reasons. I guess the current proposal looks OK to me then.

Zimmi48:
I made the `~allow_nested` parameter mandatory as recommended by @SkySkimmer and I added a commit that removes the deprecation warning. Documentation will wait because I got distracted by the state of the manual.

proofs/pfedit.ml
```diff
@@ -149,7 +149,7 @@ let next = let n = ref 0 in fun () -> incr n; !n
 let build_constant_by_tactic id ctx sign ?(goal_kind = Global, false, Proof Theorem) typ tac =
   let evd = Evd.from_ctx ctx in
   let terminator = Proof_global.make_terminator (fun _ -> ()) in
-  start_proof id goal_kind evd sign typ terminator;
+  start_proof id ~allow_nested:true goal_kind evd sign typ terminator;
```
Zimmi48: `build_constant_by_tactic` is used by `abstract`.

vernac/comFixpoint.ml
```diff
@@ -255,7 +255,7 @@ let declare_fixpoint local poly ((fixnames,fixdefs,fixtypes),pl,ctx,fiximps) ind
       Some (List.map (Option.cata (EConstr.of_constr %> Tactics.exact_no_check) Tacticals.New.tclIDTAC)
         fixdefs) in
     let evd = Evd.from_ctx ctx in
-    Lemmas.start_proof_with_initialization (Global,poly,DefinitionBody Fixpoint)
+    Lemmas.start_proof_with_initialization ~allow_nested:true (Global,poly,DefinitionBody Fixpoint)
```
Zimmi48: I haven't fully understood what `declare_fixpoint` does but it may well be that this should be set to `false` instead.

vernac/lemmas.ml
```diff
@@ -465,7 +473,8 @@ let start_proof_com ?inference_hook kind thms hook =
     else (* We fix the variables to ensure they won't be lowered to Set *)
       Evd.fix_undefined_variables evd
   in
-    start_proof_with_initialization kind evd decl recguard thms snl hook
+    start_proof_with_initialization
+      kind evd decl ~allow_nested:!allow_nested_proofs recguard thms snl hook
```
Zimmi48: `start_proof_com` is the one used by `Lemma` and the like.

vernac/obligations.ml
```diff
@@ -950,7 +950,7 @@ let rec solve_obligation prg num tac =
     Proof_global.make_terminator
       (obligation_terminator prg.prg_name num guard hook auto) in
   let hook ctx = Lemmas.mk_hook (obligation_hook prg obl num auto ctx) in
-  let () = Lemmas.start_proof_univs ~sign:prg.prg_sign obl.obl_name kind evd (EConstr.of_constr obl.obl_type) ~terminator hook in
+  let () = Lemmas.start_proof_univs ~allow_nested:true ~sign:prg.prg_sign obl.obl_name kind evd (EConstr.of_constr obl.obl_type) ~terminator hook in
```
Zimmi48: I don't know really what this code does.

vernac/lemmas.ml
```diff
@@ -359,9 +359,9 @@ let start_proof id ?pl kind sigma ?terminator ?sign c ?init_tac ?(compute_guard=
     | None -> initialize_named_context_for_proof ()
   in
   !start_hook c;
-  Pfedit.start_proof id ?pl kind sigma sign c ?init_tac terminator
+  Pfedit.start_proof id ~allow_nested:true ?pl kind sigma sign c ?init_tac terminator
```
Zimmi48: I could also add the `~allow_nested` parameter to `Lemmas.start_proof` to show further where it's being called (it is called in Funind in particular).

vernac/comFixpoint.ml
SkySkimmer: I think this is what happens for `Fixpoint foo (n:nat) : True.` so it should allow nested iff the option is on.

vernac/lemmas.ml
SkySkimmer: For instance the call in classes.ml should respect the option.

Zimmi48: OK

ejgallego: Yup there are quite a few callers, so maybe you want instead of plug the option at the STM level? In fact `pstates` may be close to be gone, you can check nesting quite easily in `process_transaction` I think.

Zimmi48 (commented):
Thanks @ejgallego. I followed your suggestion of putting the check for nested lemma in the STM. However, I need help on how to raise an error here.

The previous version has been safely saved at https://github.com/Zimmi48/coq/tree/nested-lemma/first-trial

stm/stm.ml
```diff
@@ -2802,6 +2803,13 @@ let process_transaction ~doc ?(newtip=Stateid.fresh ())
 
       (* Proof *)
       | VtStartProof (mode, guarantee, names), w ->
+
+         if not !allow_nested_proofs && VCS.proof_nesting () > 0 then
+           "Nested proofs are not allowed unless you turn option Nested Proofs Allowed on"
+           |> Pp.str
+           |> fun s -> CErrors.user_err ?loc s
```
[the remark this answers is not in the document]
ejgallego: I'm much afraid that there is no "support" for your use case, that anomaly seems in general bogus as many things can go wrong without a `safe_id`. What the code does now is to use `Exninfo.iraise (State.exn_on ~valid Stateid.dummy exn)`, that may be the most sensible thing to do given the current code.

stm/stm.ml
```diff
@@ -2201,7 +2195,6 @@ let collect_proof keep cur hd brkind id =
         let name = name ids in
         `MaybeASync (parent last, accn, name, delegate name)
     | `Sideff _ ->
```
ejgallego: While you are at it make this explicit `    | `Sideff (CherryPickEnv,_) -> `

Zimmi48:
@ejgallego I tried to implement your suggestion but something is off because the fake-ide tests which involve nested lemmas (and to which I added the option) are failing (they were not in the initial version).

ejgallego:
> @ejgallego I tried to implement your suggestion but something is off because the fake-ide tests which involve nested lemmas (and to which I added the option) are failing (they were not in the initial version).

I'd be glad to have a look if you think that could be of help [next week tho]

gares:
I think `sed -i 'somethink'` works on both osx (non-gnu) sed and linux sed, so the changes are good to me.

gares:
I'm looking at the failures right now

gares:
The reason is subtle. Since the value of the option affects the *scheduling* phase of the STM (rather than just the *execution* of commands), the value of the bool ref has to be update before scheduling takes place.
The same happens with `Set Universe Polymorphism`.

I'm pushing to your branch the same fix I used for the other flag: the classifier tells the STM that such command has to be executed immediately (eg as "Notation").

The nicer fix would be to make the STM accumulate these options while doing the scheduling, or having the attribute system take care of that (IIRC I discussed it with @maximedenes and @vbgl ). The fact (attribute) that a Lemma admits nested (sub) lemmas would need the same treatment of `Set Universe Polymorphism` and the "being polymorphic" attribute. Hopefully there two quirks will go away together.

gares (approved):

Zimmi48:
Thanks a lot @gares! I'm waiting on the merge of #7468 and #7461 before adding the doc to this PR.

Zimmi48:
The overlays are done and are incredibly simple: only one file of Color and one file of Corn required the option. This confirms that people don't use this feature much in finished code even if they rely on it during development.

The documentation remains to be done: I got side tracked one again (cf. #7487).

stm/vernac_classifier.ml
```diff
@@ -54,13 +54,20 @@ let idents_of_name : Names.Name.t -> Names.Id.t list =
   | Names.Anonymous -> []
   | Names.Name n -> [n]
 
+let stm_allow_nested_proofs_option_name = ["Nested";"Proofs";"Allowed"]
+
+let options_affecting_stm_scheuling =
```
ejgallego: Spelling of `scheuling`

ejgallego:
Indeed you can remove the overlays now I think. I support this patch, will review once the doc is in place.

It also needs an assignee.

Zimmi48:
This PR is now ready. The overlays have been dropped and the documentation has been added. Did you know that there wasn't a proper documentation of the nested lemma feature so far?
Regarding the assignee, given that the only affected component is the STM, it is either @gares or @ejgallego.

CHANGES
```diff
@@ -18,6 +18,8 @@ Vernacular Commands
 
 - Removed deprecated commands Arguments Scope and Implicit Arguments
   (not the option). Use the Arguments command instead.
+- Nested proofs are not marked as deprecated anymore but using them
+  now requires option `Nested Proofs Allowed` to be turned on.
```
ejgallego: Looks fine but maybe I would tweak the phrasing a bit more. "Using "nested proofs", that is to say, opening a `Lemma` inside the proof of another `Lemma` will produce now an error by default. You can keep using them by activating the `Nested Proof Allowed` option."

ejgallego (approved):
LGTM.

CHANGES
Zimmi48: Thanks for your comment. I'm unconvinced by my current formulation, but I'm not sure I really prefer the one you are proposing. Any other suggestion anyone?

SkySkimmer: Nested proofs may be enabled through the option `...`. By default they are disabled and produce an error. The deprecation warning which used to occur when using nested proofs has been removed.

Zimmi48: Thanks @SkySkimmer!

Zimmi48:
This now depends on #7517.

CHANGES
ejgallego: I think I prefer a phrasing that says first "now nested proofs will produce an error"

This is the most relevant bit of information for the user "action X, that used to work, now produces an error." Then, we should add some info how to fix it.

ejgallego:
> Regarding the assignee, given that the only affected component is the STM, it is either @gares or @ejgallego.

I am not sure how things are supposed to work here, as in principle I'd have to be delegated by the main maintainer to take over this PR [which I'd be happy to do].

On the other hand Enrico has a small patch here, so I dunno. Feel free to assign me if you want me to drive the merging of this PR.

Zimmi48:
Yeah, I feel like @gares kind of co-authored this PR so it would make sense that you drive it.

ejgallego (changes requested):
Our friend the linter complains.

ejgallego (approved):
Latest nits are addressed, this is going to be merged very soon, last call to complain!

ejgallego:
Oh I just see this needs a rebase, sorry @Zimmi48 !

Zimmi48:
No worries, that was a trivial one.

PR #7453, opened 2018-05-07
Refuse to parse empty [Context] command.

### LINKED ISSUE #7452: Anomaly when attempting to use an empty Context
#### Version
Coq 8.8, see file for more information.

#### Operating system
NixOS

While working on UniMath, I noticed that the following file produces an anomaly: https://github.com/siddharthist/UniMath/blob/b1f3fbbc5f0811af714f4c71645701caac3a98cb/UniMath/CategoryTheory/Subcategory/Limits.v

Using Jason's (amazing!) bug isolator, I was able to minimize it to the following:
```coq
(* -*- mode: coq; coq-prog-args: ("-indices-matter" "-type-in-type" "-noinit" "-w" "-notation-overridden,-local-declaration,+uniform-inheritance,-deprecated-option" "-R" "." "UniMath" "-top" "Limits") -*- *)
(* File reduced by coq-bug-finder from original input, then from 190 lines to 4 lines, then from 20 lines to 4 lines *)
(* coqc version 8.8.0 (May 2018) compiled on May 7 2018 12:51:22 with OCaml 4.05.0
   coqtop version 8.8.0 (May 2018) *)

Context .
```
I can confirm that when replacing the original file with that one line, the same error occurs.

Naturally, that line doesn't make any sense! I'll have to fix it in my pull request. But it might be nicer to have Coq give a better error message, with a line number etc.

SkySkimmer:
The error is now:
    Syntax error: [constr:binder] expected after 'Context' (in [vernac:gallina_ext]).

Close #7452.

### DIFF
--- a/vernac/g_vernac.ml4
+++ b/vernac/g_vernac.ml4
@@ -631,8 +631,8 @@ GEXTEND Gram
          t = class_rawexpr ->
           VernacCoercion (CAst.make ~loc:!@loc @@ ByNotation ntn, s, t)
 
-      | IDENT "Context"; c = binders ->
-	  VernacContext c
+      | IDENT "Context"; c = LIST1 binder ->
+          VernacContext (List.flatten c)
 
       | IDENT "Instance"; namesup = instance_name; ":";
 	 expl = [ "!" -> Decl_kinds.Implicit | -> Decl_kinds.Explicit ] ; t = operconstr LEVEL "200";


herbelin:
Adding a test?

SkySkimmer:
How do I test that something doesn't parse? It can't Fail after all.
Output test maybe?

herbelin:
> How do I test that something doesn't parse? It can't Fail after all.

Indeed. I realize that I actually encountered the same problem in the past and renounced to have a test.

mattam82 (approved):

mattam82:
The documentation reflects this right?

ejgallego:
`FailParse` could be added if we think it is worth it. [Not sure about this point]

ejgallego:
Note that a sensible alternative to have a better error message is to actually parse the entry and call `user_err`. Anyways, this PR needs an assignee.

SkySkimmer:
The file moved and so did the code owner (@herbelin -> @mattam82) ^^

Zimmi48:
> Note that a sensible alternative to have a better error message is to actually parse the entry and call user_err. Anyways, this PR needs an assignee.

Yes, that's what I did in #7349 for instance.

mattam82:
The doc does reflect this.

mattam82:
Not sure about the milestone. It could be backported

Zimmi48:
@mattam82 Please always choose a milestone even if you're not sure. It can always be changed later. Generally speaking, you should think twice before skipping a check from the merge script.

mattam82:
The point was to let SkySkimmer choose it here

Zimmi48:
Then make a choice, and then ask for confirmation ;-)
In any case, transforming an anomaly into a normal error sounds like something that should be backported indeed.

PR #7456, opened 2018-05-07
[toplevel] Don't ignore output filename provided by user in -o

### LINKED ISSUE #7447: -o option has stopped working for directories
8.8.0

Linux

Since I updated to coq 8.8.0, the -o option is no longer working like it used to. I am using this option to have a separate build directory, so my Makefile produces a command like coqc -R . Lib -o build/directory/file.vo directory/file.v. This used to work but this is no longer the case. Instead it creates file.vo in the same directory as file.v.

### LINKED ISSUE #7457: #7456 ([toplevel] Don't ignore output filename provided by user in -o) should have a test case in the test-suite
Adding an issue for this as requested by https://github.com/coq/coq/pull/7456#issuecomment-387233510

Probably the thing to do is to find the right category/folder for it, and add a custom `run.sh` script (as in the `coq-makefile` folder and the timing tests therein) to run `coqc` with `-o` on a dummy file and check that it creates a `.vo` file in the right place.

ejgallego:
This was a silly bug introduced in
675a1dc401eb9a5540ba5bc9a522c1f84d4c3d54 that forgot to properly
forward the command line option.

Thanks to @SkySkimmer for finding out the problem.

closes #7447
closes #7457

### DIFF
--- a/test-suite/misc/coqc_dash_o.sh
+++ b/test-suite/misc/coqc_dash_o.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+DOUT=misc/tmp_coqc_dash_o/
+OUT=${DOUT}coqc_dash_o.vo
+
+
+mkdir -p "${DOUT}"
+rm -f "${OUT}"
+$coqc misc/coqc_dash_o.v -o "${OUT}"
+if [ ! -f "${OUT}" ]; then
+  printf "coqc -o not working"
+  exit 1
+fi
+rm -fr "${DOUT}"
+exit 0
--- a/test-suite/misc/coqc_dash_o.v
+++ b/test-suite/misc/coqc_dash_o.v
@@ -1,0 +1,1 @@
+Definition x := nat.
--- a/toplevel/coqtop.ml
+++ b/toplevel/coqtop.ml
@@ -315,11 +315,12 @@ let compile cur_feeder opts ~echo ~f_in ~f_out =
   CoqworkmgrApi.giveback 1
 
 let compile_file cur_feeder opts (f_in, echo) =
+  let f_out = opts.compilation_output_name in
   if !Flags.beautify then
     Flags.with_option Flags.beautify_file
-      (fun f_in -> compile cur_feeder opts ~echo ~f_in ~f_out:None) f_in
+      (fun f_in -> compile cur_feeder opts ~echo ~f_in ~f_out) f_in
   else
-    compile cur_feeder opts ~echo ~f_in ~f_out:None
+    compile cur_feeder opts ~echo ~f_in ~f_out
 
 let compile_files cur_feeder opts =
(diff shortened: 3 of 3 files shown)


JasonGross:
It might be useful to add a test file to the test-suite

ejgallego:
> It might be useful to add a test file to the test-suite

Good idea, I might be short on cycles tho (*) and this fix is important, so please open an issue so we don't forget, or feel free to push to the branch.

(*) I ignore how the test suite works so that'd take me some time.

gares:
I think https://github.com/coq/coq/blob/master/test-suite/misc/exitstatus.sh is a good example to test this functionality.

gares:
I'm preparing the test and pushing it to the branch of @ejgallego

gares:
I should have added this test when `-o` was added. Here it is, not very deep, but have detected the regression.

gares:
I think this fix is to be merged in 8.8, so in any case it has to be checked by @Zimmi48 or @maximedenes ... I guess you guys could also do the review

test-suite/misc/coqc_dash_o.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+$coqc misc/coqc_dash_o.v -o /tmp/coqc_dash_o.vo
```
ejgallego: This is insecure and bad practice, at the very least we should use `mktemp` or a local path. I see we have more stuff like this in the test suite.

ejgallego (commented):
Fix insecure file creation.

test-suite/misc/coqc_dash_o.sh
Zimmi48: Interesting! I didn't know about `mktemp`.

ejgallego: I tried to request changes to my own PR and I discovered I couldn't do it XD

gares:
Hum, while playing with coqc, I tried to pass to it -no-glob, but it still generates a .glob file.

```
gares@gargamel:~/COQ/wt/7456$ touch a.v
gares@gargamel:~/COQ/wt/7456$ bin/coqc -no-glob a
gares@gargamel:~/COQ/wt/7456$ ls
a.glob a.v a.vo
```

@ejgallego does this ring a bell? Otherwise I'll open a separate issue.

maximedenes:
@gares please report the `-no-glob` issue. I did some benchmarking with it not so long ago, I remember being surprised I didn't measure any difference...

maximedenes (approved):

ejgallego:
From a quick look the logic for `-no-glob` is the same than before, but we could well have a bug as the logic itself is strange.

Zimmi48 (approved):

gares:
I can't reproduce the no-glob bug anymore, so I guess I was dreaming. Sorry.

PR #7459, opened 2018-05-08
Try to fix CODEOWNERS

maximedenes:
It seems the code owners feature is no longer working for us. It has happened in the past because of syntax issues in `.github/CODEOWNERS`. I don't know if the `{..,..}` syntax is supported in file paths.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -71,9 +71,9 @@
 /engine/          @ppedrot
 # Secondary maintainer @aspiwack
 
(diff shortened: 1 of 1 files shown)


SkySkimmer:
I'm pretty sure it's supposed to be supported, but who knows how github implemented their thing.

Zimmi48 (approved):
Sorry, I didn't realize this could have been the cause of the failure. Let's merge this to see if it solves the problem.

Zimmi48:
It worked.

PR #7464, opened 2018-05-09
Make whitespace linter not check for trailing space.

### LINKED ISSUE #7463: Linter reports whitespace issues not fixed by `git rebase --whitespace=fix`
Previous discussions on the linter has concluded that whitespace linting was mainly about aesthetics, IIUC.

I think it is fine to care about aesthetics, as long as the cost is low or non-existent. Running `git rebase --whitespace=fix` as part of my PR workflow is fine IMHO.

However, the linter currently reports "whitespace errors" which are not fixed by that `git` command, like a blank line at the end of a file. So please let me know if a command fixing all issues reported by the linter exists, and if not, what checks we should disable to make it true.

SkySkimmer:
git does not know how to fix this automatically when they appear by
removing a chunk of text at the end of the file.

eg
```
foo

bar
```
to
```
foo

```

Trailing space checking was enabled because it was default in git, which is not much of a reason.

This leaves checking for tab-in-indent (ocp-indent dislikes it), presence of at least 1 newline at the end of files (careless scripts can get bitten by its lack), and that we don't forget the .sh at the end of new overlays.

Closes #7463.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,48 +5,51 @@
 # Because our commit hook automatically does [apply whitespace=fix] we
 # disable whitespace checking for all files except those where we want
 # it. Otherwise rogue global configuration and forgotten local
-# configuration can break commits.
+# configuration can break commits. Note that git cannot fix but can
+# detect a blank-at-eof when it comes from removing a chunk of text at
+# the end of the file, leaving an extra newline from before that
+# chunk, so we disable blank-at-eof.
 * -whitespace
 
 # tabs are allowed in Makefiles.
-Makefile* whitespace=trailing-space
-tools/CoqMakefile.in whitespace=trailing-space
+Makefile* whitespace=blank-at-eol
+tools/CoqMakefile.in whitespace=blank-at-eol
 
 # in general we don't want tabs.
-*.asciidoc whitespace=trailing-space,tab-in-indent
-*.bib whitespace=trailing-space,tab-in-indent
-*.c whitespace=trailing-space,tab-in-indent
-*.css whitespace=trailing-space,tab-in-indent
-*.dtd whitespace=trailing-space,tab-in-indent
-*.el whitespace=trailing-space,tab-in-indent
-*.g whitespace=trailing-space,tab-in-indent
-*.h whitespace=trailing-space,tab-in-indent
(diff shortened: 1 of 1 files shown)


Zimmi48:
git doesn't know how to fix trailing spaces or just trailing newlines?

SkySkimmer:
newlines

Gaëtan Gilbert

Zimmi48:
But we can't remove the check for trailing newlines without removing the check for trailing spaces? (At least that's how I understand your PR.)

SkySkimmer:
I removed both because people keep running into them and they're 
useless. We could keep the trailing space check if we wanted to.

Gaëtan Gilbert

Zimmi48:
I see. The only thing that I dislike about trailing whitespaces is seeing this red whitespaces in git diffs (aesthetics). So I'm fine with removing this check.

Zimmi48:
We are BTW lacking owners for this file and a few other root dotfiles.

ejgallego:
> I see. The only thing that I dislike about trailing whitespaces is seeing this red whitespaces in git diffs (aesthetics). So I'm fine with removing this check.

They are indeed very annoying for colors sometimes, so if that particular check is easy to keep I would be happy, I though rebase could take care of those.

SkySkimmer:
So do I put back the end of line check?

gares:
If the problem is just about git diff showing red marks, there are options to not show whitespace stuff, or even change the colour...

maximedenes:
> So do I put back the end of line check?

I would put back end of line, but not end of file, if possible. That is, check what `git` can automatically fix.

maximedenes:
> That is, check what git can automatically fix.

To be more precise, IIUC `git` can not always fix end of line spaces in some corner cases, but that's not so critical IMHO.

SkySkimmer:
>|git| can not always fix end of line spaces in some corner cases

Where do you get that from?

Gaëtan Gilbert

maximedenes:
Forget what I wrote, I'm too tired to read correctly. It doesn't change my suggestion anyway.

SkySkimmer:
>I would put back end of line, but not end of file, if possible. That is, check what git can automatically fix.

OK I put it back.

maximedenes (approved):

PR #7467, opened 2018-05-09
Remove unused references from biblio.

Zimmi48:
A few reference have moved around when they were not in alphabetical order already due to the method I used to remove unused ones.

### DIFF
--- a/doc/sphinx/biblio.bib
+++ b/doc/sphinx/biblio.bib
@@ -121,9 +121,7 @@ s},
   pages       = {85--95},
   month       = {November},
   year        = {2000},
-  url         =
-    "{\sf ftp://ftp.inria.fr/INRIA/Projects/coq/David.Delahaye/papers/}"#
-    "{\sf LPAR2000-ltac.ps.gz}"
+  url         = {http://www.lirmm.fr/\%7Edelahaye/papers/ltac\%20(LPAR\%2700).pdf}
 }
 
 @Article{Dyc92,
--- a/doc/sphinx/biblio.bib
+++ b/doc/sphinx/biblio.bib
@@ -3,101 +3,11 @@
 @String{lnai  = "Lecture Notes in Artificial Intelligence"}
 @String{SV    = "{Sprin-ger-Verlag}"}
 
-@InProceedings{Aud91,
-  author      = {Ph. Audebaud},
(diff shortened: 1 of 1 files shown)


maximedenes (approved):
Welcome cleanup, thanks! How did you detect the unused ones?

Zimmi48:
Rebased.

ejgallego:
I indeed had the same thought, IMHO having a large related biblio to come with Coq is a nice thing.

ejgallego:
We could make this formal as in "papers using and citing Coq", some people tried to do that in a wiki page but IMHO a bib entry + generated page works well too. Example:

https://github.com/math-comp/math-comp/wiki/Publications

Zimmi48:
I don't understand this request why we would want to keep a copy in the archive: I don't think that anyone ever relied on this list (at least outside of the dev team); it hasn't ever been advertised. And the repository doesn't sound like the right place to store this anyways. If you want to start a wiki page, this will be a very welcome addition that may reveal very useful (in particular for students). Including the bibtex entries in the wiki page can be done with code blocks...

ejgallego:
> why we would want to keep a copy in the archive

As @herbelin mentioned this is important historical data, and Coq is above all a research project so shipping a historical bibliography is IMVHO proper.

herbelin:
> who really looked at it in the past?

It has been useful for me at some time. I agree that it is almost invisible from people other than developers. Then, a quick solution would be to copy it (as a bib file) on the wiki.

Zimmi48:
OK indeed that's the quickest solution so I'll implement it.

Zimmi48:
Done here: https://github.com/coq/coq/wiki/Published-Works
with a link on the main page, here: https://github.com/coq/coq/wiki#coq-community
Thus, I think nothing is delaying this PR anymore.

maximedenes:
@ejgallego @herbelin Can I merge now?

ejgallego:
Sure, I wasn't meaning to block this PR but more to state that I like software that comes with a biblio, as I think Hugo does.

But indeed taking care of the biblio is a lot of work so unless someone volunteers to make it "official" and supported I am fine with its move to the other repository.

herbelin:
Same as @ejgallego. Some work should be done to sort in the biblio what "resisted to time" but I have no time to do it currently and the move to the wiki is good enough. And after all, if we later need a reference which has been removed by this PR, it shall also be easy to add it again.

PR #7470, opened 2018-05-09
use at least 6 Xs in mktemp filename templates

### LINKED ISSUE #7469: mktemp calls fail on OpenBSD due to insufficient X's
#### Version
8.8.0

#### Operating system
OpenBSD 6.3

The OpenBSD `mktemp` utility fails with an error if less than 6 `X`s appear in the given filename pattern. This causes `test-suite/save-logs.sh` to fail when running `make test-suite`. Also affected is `dev/tools/pre-commit`.

pesco:
OpenBSD mktemp fails with an error otherwise.

**Kind:**  bug fix

Fixes #7469

### DIFF
--- a/dev/tools/pre-commit
+++ b/dev/tools/pre-commit
@@ -14,9 +14,9 @@ then
 
     # We fix whitespace in the index and in the working tree
     # separately to preserve non-added changes.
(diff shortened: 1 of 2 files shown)


SkySkimmer (approved):
XXX_XXX

PR #7471, opened 2018-05-09
[travis] Fix version of camlp5 for OCaml 4.06.1.

Zimmi48:


ejgallego (approved):
Fix is correct, Travis script is indeed buggy.

SkySkimmer:
I guess opam init is noop even if the compiler is not the current one?

ejgallego:
> I guess opam init is noop even if the compiler is not the current one?

Yeah that's a problem we've been carrying with us since at least the original script I think.

ejgallego:
> Yeah that's a problem we've been carrying with us since at least the original script I think.

So in fact we need to do a switch; Théo, could you add the following line to this PR?
```
- opam init -j ${NJOBS} --compiler=${COMPILER} -n -y
- opam switch "$COMPILER"
- eval $(opam config env)
```

ejgallego:
Thanks, let's see how this goes in the CI.

Zimmi48:
It passed here, and on my branch as well: https://travis-ci.org/Zimmi48/coq/builds/376959995

ejgallego:
Will not wait for Appveyor as it seems it would introduce a > 24 hour lag.

PR #7477, opened 2018-05-10
Support for notations with autonomous only-parsing and only-printing declarations.

### LINKED ISSUE #7462: (only printing) notations should not remove (only parsing) notations
#### Version

8.8.0

This code should work
```coq
Notation "$ x" := (@id nat x) (only parsing, at level 0).
Notation "$ x" := (@id bool x) (only printing, at level 0).
Check $1.
(* Error: Unknown interpretation for notation "$ _". *)
```
Similarly, this code should work:
```coq
Notation "$ x" := (@id bool x) (only printing, at level 0).
Notation "$ x" := (@id nat x) (only parsing, at level 0).
Check $1. (* Error: Syntax Error: Lexer: Undefined token *)
```

See also #5567
cc @herbelin

herbelin:
**Kind:** bug fix / wish

Fixes / closes #7462.

- [X] Entry added in CHANGES (put in section 8.9 / to be put in section 8.8.1 if seen as a fix).

Added note: #5567 not fixed.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -33,6 +33,14 @@ Tactic language
   called by OCaml-defined tactics.
 - Option "Ltac Debug" now applies also to terms built using Ltac functions.
 
+Changes from 8.8.0 to 8.8.1
+===========================
+
+Notations
+
+- Fixed unexpected collision between only-parsing and only-printing
+  notations (issue #7462).
+
 Changes from 8.8+beta1 to 8.8.0
 ===============================
 
--- a/interp/notation.ml
+++ b/interp/notation.ml
@@ -429,13 +429,15 @@ let rec find_without_delimiters find (ntn_scope,ntn) = function
 
 (* Uninterpreted notation levels *)
 
-let declare_notation_level ntn level =
+let declare_notation_level ?(onlyprint=false) ntn level =
   if String.Map.mem ntn !notation_level_map then
     anomaly (str "Notation " ++ str ntn ++ str " is already assigned a level.");
-  notation_level_map := String.Map.add ntn level !notation_level_map
+  notation_level_map := String.Map.add ntn (level,onlyprint) !notation_level_map
 
-let level_of_notation ntn =
-  String.Map.find ntn !notation_level_map
+let level_of_notation ?(onlyprint=false) ntn =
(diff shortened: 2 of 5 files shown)


ejgallego:
Taking this as interp's 2nd.

ejgallego:
I am neutral as whether this should go into 8.8.1 , I guess that if @JasonGross would appreciate it, the change is small enough as not pose significant risk I believe.

On an related note, we should move metasyntax from `vernac`, the Dune build indicates that with current code it can sit in `printing`, or `parsing` if a bit of refactoring is done.

Zimmi48:
It would make sense to consider this as a bug fix: only printing and only parsing notations should not interact. I don't see any threat to compatibility either. @maximedenes if you disagree, please say so. However, I would then rephrase the entry in CHANGES: "Fixed unwanted collision between only-parsing and only-printing notations (cf. #7462)."

ejgallego (approved):
Looks good to me, I wonder whether we should make the `onlyprint` parameter mandatory but it is likely not a good idea.

@herbelin , would you mind implementing @Zimmi48 's suggestion in the wording of CHANGES?

herbelin:
Added a section on 8.8.1 changes and reworded.

ejgallego:
Yup it seems like the phrasing could use some improvement, Théo's working looked OK to me.

Zimmi48:
@ejgallego You probably missed the recent update by @herbelin. The phrasing is now closer to what I had suggested and LGTM.

ejgallego:
Indeed I checked but for some reason I saw the old page. Looks fine to me, but maybe we would like to add the issue / PR number to changes? [I tend to find this useful]:

>  Fixed unexpected collision between only-parsing and only-printing notations. (#7477)

herbelin:
OK, added issue number.

We shall have to remember though that the list is not exhaustive.

I used "issue" while, formerly, "BZ" was used. "BZ" seems now inappropriate but I did not take the responsibility of changing the existing BZ into "issue" or whatever short terminology is used to refer to "issues".

Zimmi48:
Issue is fine. Just `#` is also fine since there are no more collisions between PR numbers and bug report / issue numbers.

PR #7479, opened 2018-05-10
Move 4722 (dangling symlink) to misc tests, remove dangling symlink from repo

### LINKED ISSUE #7065: Get rid of dangiling symlink in repository
#### Version

8.7

#### Operating system
Linux

The Coq git currently contains a dangling symlink: `test-suite/bugs/closed/4722/tata`. This triggers warnings in some tools, e.g. opam when pinning the Coq dir (it uses rsync which complains about the bad symlink), or gitlab CI caches.

The symlink is a test for #4722. This test could be moved to a misc test (suggested by @SkySkimmer) so that instead of storing a dangling symlink in the repo, we create it at test time.

RalfJung:
**Kind:** bug fix

Fixes #7065


SkySkimmer:
Should we remove the link after running coqtop? It seems like if some tool has issues with it, the issues will remain if the tool runs the test suite.

test-suite/misc/4722.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+set -e
+
+if ! test -d misc/4722; then
```
SkySkimmer: I don't like this test for some reason. Would you mind doing 
```bash
mkdir -p misc/4722
ln -sfT toto misc/4722/tata
touch misc/4722.v
```
instead?

SkySkimmer (approved):
>Like so?

Yes but please squash.

test-suite/misc/4722.sh
RalfJung: *shrug* I find that entirely unreadable and much less clear than my code, but I don't care strongly.

RalfJung:
> Yes but please squash.

Done.

test-suite/misc/4722.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+set -e
+
+# create test files
+mkdir -p misc/4722
+ln -sfT toto misc/4722/tata
```
SkySkimmer: Sorry, it seems `ln -T` doesn't exist on OSX.
It shouldn't be necessary in this case (tata should never be a directory) so just remove it.

SkySkimmer (changes requested):

test-suite/misc/4722.sh
RalfJung: Fixed.

SkySkimmer (approved):

PR #7025, opened 2018-03-20
Coq makefile: provide variables to extend the flags passed to coq, coqchk, coqdoc

RalfJung:
This provides a way to pass extra flags to coq/coqchk/coqdoc without overwriting the default flags.

Fix #6659

PR #7481, opened 2018-05-11
document 7025 (coq_makefile flag variables)

gares:
Now that the chapter about coq_makefile is ported to sphinx, #7025 can be properly documented.

@RalfJung please check this


RalfJung:
Would it make sense to mention the default values? I think it would be helpful.

gares:
done

RalfJung (approved):
Looks good!

I would have mentioned explicitly that the EXTRA variables are by default empty, for extra clarity. That's not needed though.

gares:
ltac2 failure is clearly unrelated

maximedenes (approved):
LGTM

PR #7484, opened 2018-05-11
Fix non-portable shebang in test-suite.

Zimmi48:
**Kind:** bug fix

So that I can run the test-suite locally.

### DIFF
--- a/test-suite/misc/coqc_dash_o.sh
+++ b/test-suite/misc/coqc_dash_o.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 DOUT=misc/tmp_coqc_dash_o/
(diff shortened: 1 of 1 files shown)


SkySkimmer:
How hard would it be to add a nix job to gitlab?

Zimmi48:
Not hard at all. There is already a Docker image for NixOS BTW (for CircleCI but it should work perfectly for GitLabCI as well): https://hub.docker.com/r/nixorg/nix/tags/

gares (approved):

gares:
FTR, do you confirm that `/bin/sh` exists on nix? My bad I used `bash` but at least `sh` must be there https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s04.html

PR #7486, opened 2018-05-11
Update old parts of CHANGES to use current bug numbers.

Zimmi48:
This is a little clean-up in `CHANGES` which was prompted by @herbelin's comment here https://github.com/coq/coq/pull/7477#issuecomment-388371837. The idea is to remove the references to Bugzilla bug numbers (starting with `BZ#`) but this implies using the current numbers for those which have been renumbered. Given that Bugzilla is not accessible anymore, I guess it makes sense to allow for easier access to the content of these bug reports. However, it feels also a bit weird to "rewrite history" like this...

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -96,7 +96,7 @@ Tactics
   profiling, and "Set NativeCompute Profile Filename" customizes
   the profile filename.
 - The tactic "omega" is now aware of the bodies of context variables
-  such as "x := 5 : Z" (see BZ#148). This could be disabled via
+  such as "x := 5 : Z" (see #1362). This could be disabled via
   Unset Omega UseLocalDefs.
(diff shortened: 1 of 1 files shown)


SkySkimmer:
>However, it feels also a bit weird to "rewrite history" like this...

History is written by the winners. And Coq is a winner. Let's do it.

herbelin:
> History is written by the winners. And Coq is a winner. Let's do it.

????

Somehow I disagreee on the two first points... or, rather, things are not so simple.

In any case, `CHANGES` was already rewritten to use `BZ`, so it can be rewritten to not use it.

Zimmi48:
@herbelin can you be the assignee here?

herbelin:
Sure (bypassing the failures which have no reasons to be related). Shall do soon.

PR #7189, opened 2018-04-06
Fix #5539: algebraic universe produced by cases.

SkySkimmer:
The stdlib and test suite go through, let's see the CI.

Issue #5539, opened 2017-05-17
Universe polymorphism breaks type checking of term that Coq itself printed


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#5539
From: @andersk
Reported version: 8.6
CC:   coq-bugs-redist@lists.gforge.inria.fr

SkySkimmer:
Case compilation is generating an algebraic universe:
```coq
Print follow.

follow@{Top.7 Top.10} = 
fix follow (n : nat) : D n -> Prop :=
  match n as n0 return (D n0 -> Prop) with
  | 0 =>
      fun d : D 0 =>
      match d in (D n0) return match n0 with
                               | 0 => Prop
                               | S _ => IDProp
                               end with
      | DO => True
      | DS _ _ => idProp
      end
  | S n' =>
      fun d : D (S n') =>
      match
        d as d0 in (D n0)
        return
(***************************************************)
          (match n0 as x return (D x -> Type@{max(Set+1,Top.10)}) (* !!! HERE *) with
(***************************************************)
           | 0 => fun _ : D 0 => IDProp
           | S n1 => fun _ : D (S n1) => (D n1 -> Prop) -> Prop
           end d0)
      with
      | DO => idProp
      | DS n0 d' => fun f : D n0 -> Prop => f d'
      end (follow n')
  end
     : forall n : nat, D n -> Prop
(* Top.7 Top.10 |=  *)
```

simpl does refolding which hides the algebraic universe.

Fix incoming.

PR #7490, opened 2018-05-11
Forbid pretyping algebraic universes in most positions.

SkySkimmer:
They are allowed when we pretype with [IsType] constraint and from
that in the codomain of products and in the body of letins.

This should cut down on algebraic-universe-in-constraint anomalies.

I'm hoping nobody has these annotations anywhere else, let's see what CI says.

@mattam82 does that seem about right to you?

- [x] Entry added in CHANGES.

### DIFF
--- a/pretyping/pretyping.ml
+++ b/pretyping/pretyping.ml
@@ -150,16 +150,22 @@ let interp_universe_level_name ~anon_rigidity evd qid =
         with UGraph.AlreadyDeclared -> evd
       in evd, level
 
-let interp_universe ?loc evd = function
+let interp_universe ?loc ?(allow_alg=false) evd = function
   | [] -> let evd, l = new_univ_level_variable ?loc univ_rigid evd in
 	    evd, Univ.Universe.make l
   | l ->
+    (if not allow_alg
+     then match l with
+       | [Some (_,0)] | [None] -> ()
+       | _ ->
+         user_err ?loc ~hdr:"interp_universe"
+           Pp.(str "Algebraic universe not allowed here."));
     List.fold_left (fun (evd, u) l ->
       let evd', u' =
         match l with
         | Some (l,n) ->
            (* [univ_flexible_alg] can produce algebraic universes in terms *)
-           let anon_rigidity = univ_flexible in
+          let anon_rigidity = univ_flexible in (* TODO if allow_alg then flex_alg *)
            let evd', l = interp_universe_level_name ~anon_rigidity evd l in
            let u' = Univ.Universe.make l in
            (match n with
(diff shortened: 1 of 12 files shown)


SkySkimmer:
This isn't the right check, eg it allows `Variable foo : Type@{i+1}.`

OTOH there is only so much we can enforce, `Definition foo@{i} : Type@{i+1} := Type@{i}.` is fine but `Definition foo@{i j} : Type@{i+1} := Type@{j}.` isn't.

mattam82 (approved):
I think this goes in the right direction without polluting the code too much (thanks to the economic argument passing, my first experiment was much worse 4 years back). Indeed we cannot do a purely syntactic check to rule out all the wrong cases but that's already a big improvement over anomalies.

SkySkimmer:
Here's an additional commit adding a kernel time check.
It breaks the build due to #5539 (fix waiting in #7189).

maximedenes:
@mattam82 can you self-assign?

pretyping/pretyping.ml
```diff
@@ -205,16 +205,22 @@ let interp_universe_level_name ~anon_rigidity evd r =
...
+       | _ ->
+         user_err ?loc ~hdr:"interp_universe"
+           Pp.(str "Algebraic universe not allowed here."));
     List.fold_left (fun (evd, u) l ->
       let evd', u' =
         match l with
         | Some (l,n) ->
            (* [univ_flexible_alg] can produce algebraic universes in terms *)
-           let anon_rigidity = univ_flexible in
+          let anon_rigidity = univ_flexible in (* TODO if allow_alg then flex_alg *)
```
mattam82: Shall we leave this as a TODO really?

mattam82: Nevermind

mattam82:
LGTM. With the CHANGES I will merge once CI is done

SkySkimmer:
> It breaks the build due to #5539 (fix waiting in #7189).

Still broken, seems like it was a different bug.

SkySkimmer:
Now depends on #7696.

ppedrot:
@SkySkimmer You should rebase.

SkySkimmer:
Rebased.

SkySkimmer:
I'm going to give up on this PR. Maybe someday we will better understand algebraics and come back to it.

jashug:
@SkySkimmer I'd like to try modifying this PR, but can't get a copy since the branch has been deleted. If you still have the tree somewhere, would you mind sending me a reference?

SkySkimmer:
>@SkySkimmer I'd like to try modifying this PR, but can't get a copy since the branch has been deleted. If you still have the tree somewhere, would you mind sending me a reference?

I clicked the restore branch button

jashug:
Thanks, that worked. If my modifications are successful, would you prefer I reopen this PR and push to your branch, or that I open a new PR?

SkySkimmer:
Make a new PR

PR #7491, opened 2018-05-11
Fix #7421: constr_eq ignores universe constraints.

### LINKED ISSUE #7421: Ltac constr_eq does not check that universes match
#### Version

~~~~~
The Coq Proof Assistant, version 8.8.0 (May 2018)
compiled on May 3 2018 7:11:49 with OCaml 4.06.0
~~~~~

#### Operating system

Arch Linux, up to date as of today.

Ltac match and constr_eq do not check that universes match. This is counter to the expectation that terms considered equal by Ltac match or constr_eq are convertible.

~~~~~
Universe i.
Goal False.
  unify (Type@{i}) (Type@{i}). (* succeeds as expected *)
  unify (Type@{i+1}) (Type@{i+1}). (* succeeds as expected *)
  constr_eq (Type@{i}) (Type@{i}). (* succeeds as expected *)
  constr_eq (Type@{i+1}) (Type@{i+1}). (* succeeds as expected *)
  match constr:(Type@{i}) with Type@{i} => idtac end. (* succeeds as expected *)
  match constr:(Type@{i+1}) with Type@{i+1} => idtac end. (* succeeds as expected *)
  Fail unify (Type@{i}) (Type@{i+1}). (* fails as expected *)

  Fail match constr:(Type@{i}) with Type@{i+1} => idtac end. (* succeeds unexpectedly *)
  Fail constr_eq (Type@{i}) (Type@{i+1}). (* succeeds unexpectedly *)
  Fail constr_eq Prop Set. (* succeeds unexpectedly *)
  Fail constr_eq Prop Type.
Abort.
~~~~~

(included examples from: #7310)

SkySkimmer:
The test isn't quite the one in #7421 because that use of algebraic
universes is wrong.

Ltac `match` is a different code path, I'll split it to its own issue.

NB: the implementation of the tactic was moved out of the .ml4 as camlp5 barfs on `match with exception`.

### DIFF
--- a/tactics/tactics.ml
+++ b/tactics/tactics.ml
@@ -5034,6 +5034,26 @@ let tclABSTRACT ?(opaque=true) name_op tac =
     else name_op_to_name name_op (DefinitionBody Definition) "_subterm" in
   abstract_subproof ~opaque s gk tac
 
+let constr_eq ~strict x y =
+  let fail = Tacticals.New.tclFAIL 0 (str "Not equal") in
+  let fail_universes = Tacticals.New.tclFAIL 0 (str "Not equal (due to universes)") in
+  Proofview.Goal.enter begin fun gl ->
+    let env = Tacmach.New.pf_env gl in
+    let evd = Tacmach.New.project gl in
+      match EConstr.eq_constr_universes env evd x y with
+      | Some csts ->
+        let csts = UnivProblem.to_constraints ~force_weak:false (Evd.universes evd) csts in
+        if strict then
+          if Evd.check_constraints evd csts then Proofview.tclUNIT ()
+          else fail_universes
+        else
+          (match Evd.add_constraints evd csts with
+           | evd -> Proofview.Unsafe.tclEVARS evd
+           | exception Univ.UniverseInconsistency _ ->
+             fail_universes)
+      | None -> fail
+  end
+
 let unify ?(state=full_transparent_state) x y =
   Proofview.Goal.enter begin fun gl ->
   let sigma = Proofview.Goal.sigma gl in
--- a/test-suite/bugs/closed/7421.v
+++ b/test-suite/bugs/closed/7421.v
@@ -1,0 +1,39 @@
+
+
+Universe i j.
+
+Goal False.
+Proof.
+  Check Type@{i} : Type@{j}.
+  Fail constr_eq_strict Type@{i} Type@{j}.
+  assert_succeeds constr_eq Type@{i} Type@{j}. (* <- i=j is forgotten after assert_succeeds *)
+  Fail constr_eq_strict Type@{i} Type@{j}.
+
+  constr_eq Type@{i} Type@{j}. (* <- i=j is retained *)
+  constr_eq_strict Type@{i} Type@{j}.
+  Fail Check Type@{i} : Type@{j}.
+
+  Fail constr_eq Prop Set.
+  Fail constr_eq Prop Type.
+
+  Fail constr_eq_strict Type Type.
+  constr_eq Type Type.
+
+  constr_eq_strict Set Set.
+  constr_eq Set Set.
+  constr_eq Prop Prop.
+
+  let x := constr:(Type) in constr_eq_strict x x.
+  let x := constr:(Type) in constr_eq x x.
+
+  Fail lazymatch type of prod with
+       | ?A -> ?B -> _ => constr_eq_strict A B
+       end.
+  lazymatch type of prod with
+  | ?A -> ?B -> _ => constr_eq A B
+  end.
+  lazymatch type of prod with
+  | ?A -> ?B -> ?C => constr_eq A C
+  end.
+
+Abort.
--- a/CHANGES
+++ b/CHANGES
@@ -28,6 +28,12 @@ Tactics
 - The `simple apply` tactic now respects the `Opaque` flag when called from
   Ltac (`auto` still does not respect it).
 
+- Tactic `constr_eq` now adds universe constraints needed for the
+  identity to the context (it used to ignore them). New tactic
+  `constr_eq_strict` checks that the required constraints already hold
+  without adding new ones. Preexisting tactic `constr_eq_nounivs` can
+  still be used if you really want to ignore universe constraints.
+
 Tools
 
 - Coq_makefile lets one override or extend the following variables from
--- a/doc/sphinx/proof-engine/tactics.rst
+++ b/doc/sphinx/proof-engine/tactics.rst
@@ -3949,9 +3949,20 @@ succeeds, and results in an error otherwise.
    :name: constr_eq
 
    This tactic checks whether its arguments are equal modulo alpha
-   conversion and casts.
+   conversion, casts and universe constraints. It may unify universes.
 
 .. exn:: Not equal.
+.. exn:: Not equal (due to universes).
+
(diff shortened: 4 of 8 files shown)


JasonGross (changes requested):
I think that `constr_eq` should not have side effects.  (And `match` to `idtac` should definitely not have side-effects.)  It should check that the universes are compatible, but it should *not* add any induced constraints to the universe graph.

More broadly, there are two sorts of things I might want to do: I might want to check if two things are the same (for various notions of the same) (and fail otherwise), and I might want to enforce that two are the same (for various notions of the same), updating the environment as necessary.  The first is generally what I mean when I write `match`, `constr_eq`, etc; the second is what I mean when I write, e.g., `unify`.

SkySkimmer:
>I think that constr_eq should not have side effects. (And match to idtac should definitely not have side-effects.) It should check that the universes are compatible, but it should not add any induced constraints to the universe graph.

Makes sense, changed.

andres-erbsen:
Hmm. `constr_eq` does not consider two evars equal. Both evars and universes are implicitly existentially quantified in some sense, so perhaps constr_eq shouldn't consider two different but compatible universe variables equal either? I am guessing that this is not the functionality what comes in common usage, but given that `constr_eq` is a low-level primitive, it is probably worth taking the time to be principled and not unnecessarily opinionated about it.

test-suite/bugs/closed/7421.v
```diff
@@ -0,0 +1,16 @@
...
+Proof.
+  Fail constr_eq Type@{i} Type@{j}.
+  Fail constr_eq Prop Set.
+  Fail constr_eq Prop Type.
+Abort.
+
+Constraint i = j.
+Goal False.
+Proof.
+  constr_eq (Type@{i}) (Type@{j}).
```
JasonGross: I think the following should also work, and should get tests:
```coq
constr_eq Type Type.
constr_eq Set Set.
constr_eq Prop Prop.
let x := constr:(Type) in constr_eq x x.
lazymatch type of prod with
| ?A -> ?B -> _ => constr_eq A B
end.
lazymatch type of prod with
| ?A -> ?B -> ?C => constr_eq A C
end.
```

tactics/tactics.ml
```diff
@@ -5038,6 +5038,18 @@ let tclABSTRACT ?(opaque=true) name_op tac =
...
 
+let constr_eq x y =
+  Proofview.Goal.enter begin fun gl ->
+    let env = Tacmach.New.pf_env gl in
+    let evd = Tacmach.New.project gl in
+      match EConstr.eq_constr_universes env evd x y with
+      | Some csts ->
+        let csts = Universes.to_constraints ~force_weak:false (Evd.universes evd) csts in
+        if Evd.check_constraints evd csts then Proofview.tclUNIT ()
+        else Tacticals.New.tclFAIL 0 (str "Not equal")
```
JasonGross: Perhaps this should get a different error message?  (e.g., "Not equal (universe mismatch)")

JasonGross (dismissed):

JasonGross:
> so perhaps `constr_eq` shouldn't consider two different but compatible universe variables equal either

This doesn't work very well.  For example, this would make `constr_eq Type Type` fail, and if `prod` were universe-polymorphic, it would make `constr_eq prod prod` fail.  The difference between evars and universes is that you have to wrap something in `open_constr` to get fresh implicit evars, but you don't have to do anything special to get fresh implicit universes.  I can see there being a case for having a `constr_eq_strict_universes` tactic, but, so long as the default in Coq is to not write universes, the default should also be to not care too much about universe identity.

SkySkimmer:
Either `constr_eq Type Type` fails or it has side effects. In the later case if you want the side effect free version you can wrap it in `assert_succeeds`.
So do I make it have side effects or not?

JasonGross:
I want there to be a tactic called something like `constr_eq_add_constraints` which adds constraints, one called something like `constr_eq_strict_univs` which fails on `Type Type`, and I want the tactic called `constr_eq` to be `assert_succeeds constr_eq_add_constraints` (with correct error messages).

SkySkimmer:
I definitely don't want the default looking one to add constraints and then forget them. If I say `constr_eq foo bar; baz` I should be able to assume in baz that they're actually equal.

SkySkimmer:
Meanwhile it seems cross-crypto expects it to add constraints (it probably doesn't care if we forget them though).

andres-erbsen:
cross-crypto uses constr_eq in tests of Ltac code which is itself experimental. it'll be easy to port those uses to any other mechanism of term equally testing. I think it is much more important to decide on a reasonable semantics for universe variables in Ltac; I will just fix up cross-crypto for whatever we decide here.

I am still of the opinion that universe variables should be treated like other evars. I understand that they're currently not treated that way, but I don't think there is a good reason for why that makes sense. at the very least, keeping evar semantics straight is already hard enough, and I'm worried that treating universe variables differently would result in a whole new can of worms.

ppedrot:
FTR, I'm deeply convinced that equality on terms is a delicate matter, as witnessed by the vast array of subtly distinct functions in the OCaml codebase. There is no reason such a complexity should not be reflected in the tactic language...

SkySkimmer:
Updated.

SkySkimmer:
In 4728.v this bug causes a new error `Error: The reference numgoals was not found in the current environment.`
I don't understand why.

SkySkimmer:
Any Ltac/parsing experts around?

ppedrot:
I don't understand how that was parsing before. The `constr_eq` term is expecting two constrs as arguments. I suspect this is due to CAMLP5 being clever and factorizing rules in the previous state of the code, which it cannot do anymore now due to the fact there are two entries in the tactic extension.

SkySkimmer:
>I don't understand how that was parsing before. The constr_eq term is expecting two constrs as arguments. I suspect this is due to CAMLP5 being clever and factorizing rules in the previous state of the code, which it cannot do anymore now due to the fact there are two entries in the tactic extension.

OK I changed the bug file.

SkySkimmer:
Added CHANGES.

ppedrot (changes requested):
Can you split the TACTIC EXTEND entry into two separate ones? This will provide a fully backwards compatible syntax. Then you can revert the change on the test file.

SkySkimmer:
Done

ppedrot:
@JasonGross are you at least not objecting to the current state of the PR?

ppedrot (approved):

JasonGross:
The variant(s) of `constr_eq` should be documented in the refman and added to the tactic index, but otherwise I am not objecting to the current form of this PR.

SkySkimmer:
>The variant(s) of constr_eq should be documented in the refman and added to the tactic index, but otherwise I am not objecting to the current form of this PR.

Doc added.

ppedrot:
There is only one place using `globals`, and it is using in a very hackish way. This code should rather be browsing the module structure rather than the structureless global blob.

Also, making a function available doesn't mean we endorse its use. Often people go to the shortest solution, so that we should strive for easiness of correct code and clumsiness of the undesirable one.

SkySkimmer:
>There is only one place using globals, and it is using in a very hackish way.

Comment for #7745?

ppedrot:
Yep sorry!

PR #7493, opened 2018-05-12
Minor update of the documentation about the rcfile

### LINKED ISSUE #6662: What does -q do in coqc?
I find it not clear to me what the `-q` command does in `coqc`. In the manual of `coqtop`, it says `-q` means "skip loading of rcfile." However, what does rcfile do? And in what cases shall we not load them? I believe this should be stated in the manual, since beginners like me don't understand.

Sorry that I don't know if this question is suitable for the Issues page. If not, where shall I raise this kind of questions? Thanks!

herbelin:
**Kind:** documentation

To some extend, fixes #6662.

This makes more precise the doc about the `coqrc` file in the reference manual. This incidentally fixes a missing parenthesis. This gives an alternative less-idiomatic name to the pre-unix expression "rcfile".

We also use a `~` to refer to the home directory. Maybe this is too unix specific, but I don't know how this could/should be said on, e.g. windows.

Added: sphinx search seems to have a little problem. If I e.g. search for [coqrc](https://coq.inria.fr/refman/search.html?q=coqrc&check_keywords=yes&area=default) in the online manual, then the search tool does not see that a sequence of `~~~~~~~~~~~` is actually the underlining of the previous section header. Don't know if this is easily fixable from our side.

### DIFF
--- a/doc/sphinx/practical-tools/coq-commands.rst
+++ b/doc/sphinx/practical-tools/coq-commands.rst
@@ -55,15 +55,20 @@ Customization at launch time
 By resource file
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-When |Coq| is launched, with either ``coqtop`` or ``coqc``, the resource file
-``$XDG_CONFIG_HOME/coq/coqrc.xxx`` is loaded, where ``$XDG_CONFIG_HOME``
+When |Coq| is launched, with either ``coqtop`` or ``coqc``, the
+resource file ``$XDG_CONFIG_HOME/coq/coqrc.xxx``, if it exists, will
+be implicitly prepended to any document read by Coq, whether it is an
+interactive session or a file to compile. Here, ``$XDG_CONFIG_HOME``
 is the configuration directory of the user (by default its home
-directory ``/.config`` and ``xxx`` is the version number (e.g. 8.8). If
+directory ``~/.config``) and ``xxx`` is the version number (e.g. 8.8). If
 this file is not found, then the file ``$XDG_CONFIG_HOME/coqrc`` is
-searched. You can also specify an arbitrary name for the resource file
+searched. If not found, it is the file ``~/.coqrc.xxx`` which is searched,
+and, if still not found, the file ``~/.coqrc``. If the latter is also
+absent, no resource file is loaded.
+You can also specify an arbitrary name for the resource file
 (see option ``-init-file`` below).
 
-This file may contain, for instance, ``Add LoadPath`` commands to add
+The resource file may contain, for instance, ``Add LoadPath`` commands to add
 directories to the load path of |Coq|. It is possible to skip the
 loading of the resource file with the option ``-q``.
 
--- a/man/coqtop.1
+++ b/man/coqtop.1
@@ -110,7 +110,7 @@ print Coq version and exit
 
 .TP
 .B \-q
-skip loading of rcfile
+skip loading of rcfile (resource file) if any
 
 .TP
 .BI \-init\-file \ filename


doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -56,14 +56,17 @@ By resource file
...
 this file is not found, then the file ``$XDG_CONFIG_HOME/coqrc`` is
-searched. You can also specify an arbitrary name for the resource file
+searched. If not found, it is the file ``~/.coqrc.xxx`` which is searched,
+and, if still not found, the file ``~/.coqrc``. If the latter is also
+absent, no resource file is loaded.
+You can also specify an arbitrary name for the resource file
 (see option ``-init-file`` below).
 
-This file may contain, for instance, ``Add LoadPath`` commands to add
+The resource file may contain, for instance, ``Add LoadPath`` commands to add
```
maximedenes: Isn't this considered bad practice?

man/coqtop.1
```diff
@@ -110,7 +110,7 @@ print Coq version and exit
 
 .TP
 .B \-q
-skip loading of rcfile
+skip loading of rcfile (resource file if any)
```
maximedenes: I would write skip loading of rcfile (resource file) if any

doc/sphinx/practical-tools/coq-commands.rst
ejgallego: It is a bit archane but not a bad practice per se, it is equivalente to setting some global -R / -Q options.

Zimmi48: Isn't it basically the only point of the existence of this file anyways?

man/coqtop.1
herbelin: Right, changed.

No opinion otherwise on the rc file and on whether `Add LoadPath` is "good" or "bad".

ejgallego (approved):
The patch looks good to me. Taking over this as the `toplevel` maint.

doc/sphinx/practical-tools/coq-commands.rst
ejgallego: @herbelin you may want to be more explicit and reflect the current semantics: "the content of the resource file will be appended to the beginning of each document read by Coq".

ejgallego:
Waiting on @herbelin to answer my question above, just in case he wants to do that tweak. [If you do so rebase as the CI will run much faster]

doc/sphinx/practical-tools/coq-commands.rst
herbelin: But it is not included in vo files, isn't it?

ejgallego: Yup, I think it is included.

herbelin: Ah, interesting!

But after all, as soon as we accept any command there, we have to store the potential dependencies somewhere, and there is no other place for that than the .vo file.

I changed into: "When Coq is launched, with either coqtop or coqc, the
resource file $XDG_CONFIG_HOME/coq/coqrc.xxx, if it exists, will
be implicitly prepended to any document read by Coq, whether it is an
interactive session or a file to compile.". Is that ok?

ejgallego: Yup, that's seems a bit clearer about the purpose, thanks. It is hard to come up with good wording for this kind of stuff.

As far as I can trace back in history the commands in the file were always processed "normally", as @Zimmi48 points out indeed the utility of such rc file in the current form is a bit questionable, but it doesn't create too much of a problem I'd say. For libraries we want to provide higher-level primitives anyways.

PR #7495, opened 2018-05-12
Fix restrict_universe_context

### LINKED ISSUE #7472: universe restriction drops equations
Checkout: https://github.com/LPCIC/coq-elpi/tree/bug-mimin (commit f214d4141c).
```
$ make theories/derive/tests/test_param1.vo
...
UNIVERSES:
 {elpi.derive.tests.test_param1.15 elpi.derive.tests.test_param1.14 elpi.derive.tests.test_param1.13 elpi.derive.tests.test_param1.12} |= Set < elpi.derive.tests.test_param1.12
                                                 Set < elpi.derive.tests.test_param1.14
                                                 Set < elpi.derive.tests.test_param1.15
                                                 elpi.derive.tests.test_param1.15 < elpi.derive.tests.test_param1.14
                                                 Set <= elpi.derive.tests.test_param1.14
                                                 Set <= elpi.derive.tests.test_param1.15
                                                 elpi.derive.tests.test_param1.6 <= elpi.derive.tests.test_param1.14
                                                 elpi.derive.tests.test_param1.6 <= elpi.derive.tests.test_param1.15
                                                 elpi.derive.tests.test_param1.13 = elpi.derive.tests.test_param1.15
                                                 
ALGEBRAIC UNIVERSES:{}
UNDEFINED UNIVERSES:
WEAK CONSTRAINTS:

{elpi.derive.tests.test_param1.13 elpi.derive.tests.test_param1.6}
UNIVERSES:{elpi.derive.tests.test_param1.13} |= 
ALGEBRAIC UNIVERSES:{}
UNDEFINED UNIVERSES:
WEAK CONSTRAINTS:
```
The used ones are printed between the second set of constraints, namely `{elpi.derive.tests.test_param1.13 elpi.derive.tests.test_param1.6}`.

SkySkimmer:
This should fix #7472 although I didn't test it.

Instead of messing with a buggy custom graph algorithm we go through the ugraph algorithm in a slightly brute-force manner.
It seems this impacts some tricky heuristic causing HoTT and formal-topology (and a HoTT-based test suite file) to fail. Still investigating.

Bench: https://ci.inria.fr/coq/view/benchmarking/job/benchmark-part-of-the-branch/426/console
Despite the unsophisticated algorithm there is little impact, I think we rarely have many new universes / constraints per object.
```
┌──────────────────────────┬─────────────────────────┬───────────────────────────────────────┬───────────────────────────────────────┬─────────────────────────┬─────────────────┐
│                          │      user time [s]      │              CPU cycles               │           CPU instructions            │  max resident mem [KB]  │   mem faults    │
│                          │                         │                                       │                                       │                         │                 │
│             package_name │     NEW     OLD PDIFF   │            NEW            OLD PDIFF   │            NEW            OLD PDIFF   │     NEW     OLD PDIFF   │ NEW OLD PDIFF   │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│               coq-sf-vfa │   26.19   26.48 -1.10 % │    72078267639    72152115741 -0.10 % │    93867490685    94035628199 -0.18 % │  541376  541380 -0.00 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                coq-sf-lf │   20.78   20.88 -0.48 % │    56161474465    56271106701 -0.19 % │    73618512048    73872348912 -0.34 % │  429260  429140 +0.03 % │   7   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│              coq-bignums │   79.78   80.08 -0.37 % │   220549849581   220674642906 -0.06 % │   285941360780   286264820515 -0.11 % │  528684  528732 -0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│          coq-fiat-crypto │ 4108.89 4113.51 -0.11 % │ 11397157329157 11410919665169 -0.12 % │ 18427840323976 18442342141631 -0.08 % │ 3185336 3147660 +1.20 % │   8  33 -75.76 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│         coq-math-classes │  228.35  228.58 -0.10 % │   628530108909   629035769844 -0.08 % │   788150193061   786417458071 +0.22 % │  532584  533152 -0.11 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                 coq-corn │ 1563.74 1564.48 -0.05 % │  4341727574523  4325870881332 +0.37 % │  6452506241544  6450126963962 +0.04 % │  875020  874056 +0.11 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│    coq-mathcomp-fingroup │   64.94   64.97 -0.05 % │   178952540988   179084420300 -0.07 % │   239204768022   239794506122 -0.25 % │  588384  589000 -0.10 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│           coq-coquelicot │   83.52   83.54 -0.02 % │   230493921022   230181842101 +0.14 % │   275444082307   275437213366 +0.00 % │  721580  720136 +0.20 % │  29   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                coq-flocq │  263.23  263.24 -0.00 % │   730312386310   730385581913 -0.01 % │   903594177302   903317960182 +0.03 % │ 1245704 1245604 +0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│       coq-mathcomp-field │  460.97  460.98 -0.00 % │  1282709326728  1282401193211 +0.02 % │  2066762859212  2064538964300 +0.11 % │  814584  814392 +0.02 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│         coq-fiat-parsers │  703.45  703.24 +0.03 % │  1946437216614  1946269578221 +0.01 % │  2975529193808  2979563842492 -0.14 % │ 3376576 3376520 +0.00 % │   0   9 -100.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│    coq-mathcomp-solvable │  201.21  201.12 +0.04 % │   558641643648   558190473032 +0.08 % │   770409292656   770917220716 -0.07 % │  839588  866264 -3.08 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│ coq-mathcomp-real-closed │  171.78  171.54 +0.14 % │   477975725875   477399075006 +0.12 % │   699118837987   698698054275 +0.06 % │  839508  837272 +0.27 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                coq-color │  723.37  722.23 +0.16 % │  1999700567406  1995187718879 +0.23 % │  2385012886632  2384593340636 +0.02 % │ 1450604 1446404 +0.29 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│             coq-compcert │  875.59  873.48 +0.24 % │  2425409110563  2419988860901 +0.22 % │  3483911297171  3481702485365 +0.06 % │ 1335604 1333456 +0.16 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                  coq-vst │ 3849.68 3840.16 +0.25 % │ 10701114539363 10672001365828 +0.27 % │ 14014496246065 14003087057603 +0.08 % │ 2232116 2234312 -0.10 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│              coq-unimath │ 1291.28 1287.94 +0.26 % │  3576386632018  3566802329864 +0.27 % │  6047356902985  6047511516079 -0.00 % │ 1071100 1071192 -0.01 % │   1   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│   coq-mathcomp-ssreflect │   45.31   45.19 +0.27 % │   123795015548   123548759317 +0.20 % │   145512283840   145469700027 +0.03 % │  537116  537276 -0.03 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│     coq-mathcomp-algebra │  185.64  185.10 +0.29 % │   514971975928   513827131380 +0.22 % │   693801057199   693315221524 +0.07 % │  650772  645120 +0.88 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│               coq-geocoq │ 3104.64 3092.45 +0.39 % │  8610245606594  8578404414075 +0.37 % │ 13830358266032 13818865418258 +0.08 % │ 1239700 1210724 +2.39 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│               coq-sf-plf │   53.94   53.63 +0.58 % │   148135110655   147999661663 +0.09 % │   187565992538   187696689307 -0.07 % │  514416  514684 -0.05 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│            coq-fiat-core │  111.95  111.25 +0.63 % │   310466951147   309300228426 +0.38 % │   380270573253   379700140505 +0.15 % │  502092  500004 +0.42 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│   coq-mathcomp-odd-order │ 1420.54 1411.10 +0.67 % │  3957117294869  3931405783138 +0.65 % │  6664863723229  6666033452918 -0.02 % │ 1416680 1393308 +1.68 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│   coq-mathcomp-character │  276.06  274.16 +0.69 % │   767362696214   762136638822 +0.69 % │  1091194393647  1087543890712 +0.34 % │ 1130488 1120628 +0.88 % │   0   0  +nan % │
└──────────────────────────┴─────────────────────────┴───────────────────────────────────────┴───────────────────────────────────────┴─────────────────────────┴─────────────────┘
```

### DIFF
--- a/engine/univops.ml
+++ b/engine/univops.ml
@@ -35,79 +35,14 @@ let universes_of_constr env c =
     | _ -> Constr.fold aux s c
   in aux LSet.empty c
 
-type graphnode = {
-  mutable up : constraint_type LMap.t;
-  mutable visited : bool
-}
-
-let merge_types d d0 =
-  match d, d0 with
-  | _, Lt | Lt, _ -> Lt
-  | Le, _ | _, Le -> Le
-  | Eq, Eq -> Eq
-
-let merge_up d b up =
-  let find = try Some (LMap.find b up) with Not_found -> None in
-  match find with
-  | Some d0 ->
-    let d = merge_types d d0 in
-    if d == d0 then up else LMap.add b d up
-  | None -> LMap.add b d up
-
-let add_up a d b graph =
-  let node, graph =
-    try LMap.find a graph, graph
-    with Not_found ->
-      let node = { up = LMap.empty; visited = false } in
-      node, LMap.add a node graph
-  in
-  node.up <- merge_up d b node.up;
-  graph
-
-(* for each node transitive close until you find a non removable, discard the rest *)
-let transitive_close removable graph =
-  let rec do_node a node =
-    if not node.visited
-    then
-      let keepup =
-        LMap.fold (fun b d keepup ->
-            if not (LSet.mem b removable)
-            then merge_up d b keepup
-            else
-              begin
-                match LMap.find b graph with
-                | bnode ->
-                  do_node b bnode;
-                  LMap.fold (fun k d' keepup ->
-                    merge_up (merge_types d d') k keepup)
-                    bnode.up keepup
-                | exception Not_found -> keepup
-              end
-          )
-          node.up LMap.empty
-      in
-      node.up <- keepup;
-      node.visited <- true
-  in
-  LMap.iter do_node graph
-
-let restrict_universe_context (univs,csts) keep =
-  let removable = LSet.diff univs keep in
-  let (csts, rem) =
-    Constraint.fold (fun (a,d,b as cst) (csts, rem) ->
-        if LSet.mem a removable || LSet.mem b removable
-        then (csts, add_up a d b rem)
-        else (Constraint.add cst csts, rem))
-      csts (Constraint.empty, LMap.empty)
-  in
-  transitive_close removable rem;
-  let csts =
-    LMap.fold (fun a node csts ->
-        if LSet.mem a removable
-        then csts
-        else
-          LMap.fold (fun b d csts -> Constraint.add (a,d,b) csts)
-            node.up csts)
-      rem csts
-  in
+let restrict_universe_context (univs, csts) keep =
+  let removed = LSet.diff univs keep in
+  if LSet.is_empty removed then univs, csts
+  else
+  let allunivs = Constraint.fold (fun (u,_,v) all -> LSet.add u (LSet.add v all)) csts univs in
+  let g = UGraph.empty_universes in
+  let g = LSet.fold UGraph.add_universe_unconstrained allunivs g in
+  let g = UGraph.merge_constraints csts g in
+  let allkept = LSet.diff allunivs removed in
+  let csts = UGraph.constraints_for ~kept:allkept g in
   (LSet.inter univs keep, csts)
--- a/kernel/uGraph.ml
+++ b/kernel/uGraph.ml
@@ -503,7 +503,7 @@ let insert_edge strict ucan vcan g =
     let () = cleanup_universes g in
     raise e
 
-let add_universe vlev strict g =
+let add_universe_gen vlev g =
   try
     let _arcv = UMap.find vlev g.entries in
       raise AlreadyDeclared
@@ -520,8 +520,14 @@ let add_universe vlev strict g =
     }
     in
     let entries = UMap.add vlev (Canonical v) g.entries in
-    let g = { entries; index = g.index - 1; n_nodes = g.n_nodes + 1; n_edges = g.n_edges } in
-    insert_edge strict (get_set_arc g) v g
+    { entries; index = g.index - 1; n_nodes = g.n_nodes + 1; n_edges = g.n_edges }, v
+
+let add_universe vlev strict g =
+  let g, v = add_universe_gen vlev g in
+  insert_edge strict (get_set_arc g) v g
+
+let add_universe_unconstrained vlev g =
+  fst (add_universe_gen vlev g)
 
 exception Found_explanation of explanation
 
@@ -696,6 +702,9 @@ let enforce_univ_lt u v g =
     error_inconsistency Lt u v (get_explanation false v u g)
 
 let empty_universes =
+  { entries = UMap.empty; index = 0; n_nodes = 0; n_edges = 0 }
+
+let initial_universes =
   let set_arc = Canonical {
     univ = Level.set;
     ltle = LMap.empty;
@@ -717,9 +726,6 @@ let empty_universes =
   let entries = UMap.add Level.set set_arc (UMap.singleton Level.prop prop_arc) in
   let empty = { entries; index = (-2); n_nodes = 2; n_edges = 0 } in
   enforce_univ_lt Level.prop Level.set empty
-
-(* Prop = Set is forbidden here. *)
-let initial_universes = empty_universes
 
 let is_initial_universes g = UMap.equal (==) g.entries initial_universes.entries
 
@@ -779,6 +785,42 @@ let constraints_of_universes g =
   in
   let csts = UMap.fold constraints_of g.entries Constraint.empty in
   csts, UF.partition uf
+
+(* domain g.entries = kept + removed *)
+let constraints_for ~kept g =
+  (* rmap: partial map from canonical universes to kept universes *)
+  let rmap, csts = LSet.fold (fun u (rmap,csts) ->
+      let arcu = repr g u in
+      if LSet.mem arcu.univ kept then
+        LMap.add arcu.univ arcu.univ rmap, enforce_eq_level u arcu.univ csts
+      else
+        match LMap.find arcu.univ rmap with
+        | v -> rmap, enforce_eq_level u v csts
+        | exception Not_found -> LMap.add arcu.univ u rmap, csts)
+      kept (LMap.empty,Constraint.empty)
+  in
+  let rec add_from u csts todo = match todo with
+    | [] -> csts
+    | (v,strict)::todo ->
+      let v = repr g v in
+      (match LMap.find v.univ rmap with
+       | v ->
+         let d = if strict then Lt else Le in
+         let csts = Constraint.add (u,d,v) csts in
+         add_from u csts todo
+       | exception Not_found ->
+         (* v is not equal to any kept universe *)
+         let todo = LMap.fold (fun v' strict' todo ->
+             (v',strict || strict') :: todo)
+             v.ltle todo
+         in
+         add_from u csts todo)
+  in
+  LSet.fold (fun u csts ->
+      let arc = repr g u in
+      LMap.fold (fun v strict csts -> add_from u csts [v,strict])
+        arc.ltle csts)
+    kept csts
 
 (** [sort_universes g] builds a totally ordered universe graph.  The
     output graph should imply the input graph (and the implication
--- a/dev/ci/user-overlays/07495-gares-elpi-test-bug.sh
+++ b/dev/ci/user-overlays/07495-gares-elpi-test-bug.sh
@@ -1,0 +1,8 @@
+if [ "$CI_PULL_REQUEST" = "7495" ] || [ "$CI_BRANCH" = "fix-restrict" ]; then
+
+    # this branch contains a commit not present on coq-master that triggers
+    # the universe restriction bug #7472
+    Elpi_CI_BRANCH=overlay-7495
+    Elpi_CI_GITURL=https://github.com/LPCIC/coq-elpi.git
+
+fi
--- a/engine/univops.mli
+++ b/engine/univops.mli
@@ -14,5 +14,8 @@ open Univ
 (** The universes of monomorphic constants appear. *)
 val universes_of_constr : Environ.env -> constr -> LSet.t
 
-(** Shrink a universe context to a restricted set of variables *)
+(** [restrict_universe_context (univs,csts) keep] restricts [univs] to
+   the universes in [keep]. The constraints [csts] are adjusted so
+   that transitive constraints between remaining universes (those in
+   [keep] and those not in [univs]) are preserved. *)
 val restrict_universe_context : ContextSet.t -> LSet.t -> ContextSet.t
--- a/kernel/uGraph.mli
+++ b/kernel/uGraph.mli
@@ -49,13 +49,15 @@ exception AlreadyDeclared
 
 val add_universe : Level.t -> bool -> t -> t
 
+(** Add a universe without (Prop,Set) <= u *)
+val add_universe_unconstrained : Level.t -> t -> t
+
 (** {6 Pretty-printing of universes. } *)
 
 val pr_universes : (Level.t -> Pp.t) -> t -> Pp.t
 
 (** The empty graph of universes *)
 val empty_universes : t
-[@@ocaml.deprecated "Use UGraph.initial_universes"]
 
 val sort_universes : t -> t
 
@@ -63,6 +65,12 @@ val sort_universes : t -> t
    [csts] are the non-Eq constraints and [partition] is the partition
    of the universes into equivalence classes. *)
 val constraints_of_universes : t -> Constraint.t * LSet.t list
+
+(** [constraints_for ~kept g] returns the constraints about the
+   universes [kept] in [g] up to transitivity.
+
+    eg if [g] is [a <= b <= c] then [constraints_for ~kept:{a, c} g] is [a <= c]. *)
+val constraints_for : kept:LSet.t -> t -> Constraint.t
 
 val check_subtype : AUContext.t check_function
 (** [check_subtype univ ctx1 ctx2] checks whether [ctx2] is an instance of


SkySkimmer:
@gares could you check that it works for your bug?

Since it's hard to trigger we will have to wait for unit tests to put a regression test in.

gares:
Once merged I can enable restriction in elpi, and this triggers the bug... I'll prepare an overlay for you with minimisation on

SkySkimmer:
Feel free to push to my branch.

Gaëtan Gilbert

SkySkimmer:
Incidentally fixes the example in  #5200.

gares:
elpi passes! I see 3 tests failing:
- test suite fails `bugs/closed/4544`
- Formal topolgy fails with `Error: Universe instance should have length 4`, maybe now restriction leaves  more/less universes around and hence one has to pass more/less universe instances.
- Hott fails with `Error: Universe {HoTT.Modalities.Modality.895} is unbound.`, I've no clue

gares:
I think @mattam82 is the most reasonable assignee here, given that @SkySkimmer authors the PR. Maybe also @ppedrot understand this can shepherd. In case he can just override my assignment of this PR to Matthieu.

gares:
the rebase dropped my commit, I hope it was not intended since I put it back

SkySkimmer:
There was no rebase, I did `push -f --all` to pick up any branches lying on my disk.

gares:
Well same result.

For the formal-topology overlay, I'm not sure I did the right thing.
I made a PR with the right set of changes, see https://github.com/bmsherman/topology/pull/17 but
I made the PR on master, while the current branch being tested by ci includes extra commits from other overlays I guess

gares:
> I did push -f --all to pick up any branches lying on my disk.

well, given you did not pull this branch, you dropped my commit. No problem, but next time beware -f

SkySkimmer:
They will also need to make a branch for 8.8 testing unless this gets 
backported.

Gaëtan Gilbert

gares:
I can't fix the test suite alone. The first error is about a missing instance, I passed `l` there, but then you get the error we also have on hott (indeed the test is taken from hott) that is "unknown...". I'll stop here.

engine/univops.ml
```diff
@@ -35,79 +35,29 @@ let universes_of_constr env c =
...
+  let g = UGraph.empty_universes in
+  let g = LSet.fold UGraph.add_universe_unconstrained allunivs g in
+  let g' = UGraph.merge_constraints csts g in
+  let removed = LSet.diff univs keep in
+  let allkept = LSet.diff allunivs removed in
+  let csts, _g = LSet.fold (fun u cstsg ->
+      LSet.fold (fun v (csts,g as cstsg) ->
+          (** Fold through all pairs of kept universes. We avoid
+             recomputing for swapped order pairs. *)
+          if Level.compare u v >= 0 then cstsg
```
mattam82: Why > here?

mattam82 (commented):
I'm trying to understand the difference it introduces. It can certainly select a different subset of the constraints. There are cases where it now selects a different subset of the kept universes, eliminating more spurious ones as well? I think overall the code is better that way, reusing the existing uGraph implementation.

engine/univops.ml
```diff
@@ -35,79 +35,29 @@ let universes_of_constr env c =
...
+          else match is_constrained u v g' with
+            | None -> cstsg
+            | Some cst ->
+              if UGraph.check_constraint g cst then cstsg
+              else Constraint.add cst csts, UGraph.enforce_constraint cst g)
+        allkept cstsg)
+      allkept (Constraint.empty,g)
   in
   (LSet.inter univs keep, csts)
```
SkySkimmer: The universes we keep are exactly the same, as indicated by this line not changing. The constraints are not the same.

engine/univops.ml
SkySkimmer: >Why > here?

What do you mean?

mattam82: Why not ```Level.compare u v == 0```

SkySkimmer:
See the comment immediately above

Gaëtan Gilbert

engine/univops.ml
mattam82: Right, I don't understand the comment either... That's meant to avoid treating u = l, v = l twice?

SkySkimmer:
If the universes are u and v we will see (u,u) (v,v) (u,v) and (v,u). We 
skip (u,u) (v,v) because in a valid graph there are no interesting 
constraints between them. We skip (u,v) or (v,u) because we don't care 
about the order.

Gaëtan Gilbert

mattam82:
Good, I see now.

mattam82 (approved):

gares:
I think this bug/PR should have high priority for two reasons:
- it breaks scripts (more/less univ variables) hence it should be merged soon in order give users the time to handle this. Note that there is no way to have the same .v file work before/after this PR AFAICT (if the file is affected by the bug, of course).
- the bug popped up in my code when I ported it to the recommendations done https://github.com/ybertot/plugin_tutorials that we are about to give to others in a few weeks.

gares:
You can't use fist in a lemma statement, at least not in a decent way.

SkySkimmer:
So for instance in 4544.v `O_ind_internal` has the following universes before this patch:
```
u a i j k |= Set < a
                Set < i
                Set < j
                a < u
                a <= i
                a <= j
                i <= k
                j <= k
```
and after
```
u a i j k |= Set < u
                Set < a
                Set < i
                Set < j
                Set < k
                a < u
                a <= i
                a <= j
                a <= k
                i <= k
                j <= k
```
This is equivalent but then if you Show Universes at the end of `inO_equiv_inO` you can see that after the patch the extra universe n°954 remains, and Show Proof says it should be the `k` in some `O_ind_internal` instances. So it looks like one of those extra redundant constraints get in the way.

If we add a shortcut to the universe restriction such that whenever we remove no universes we return the input as-is it successfully works around the issue in this case. This is a kinda hacky way to do so but let's see how far it gets us.

gares:
I'm not following, the universe variables are the same (while the constraints are different but equivalent).
IIRC the file breaks because there are less variables before the `|=` symbol.

SkySkimmer:
Minimisation on equivalent constraints may produce non equivalent results.
At least I think so.

mattam82:
Indeed, minimization is currently sensitive to the constraint set and not the induced relation. I.e. there is no “canonical”minimal set of constraints for equivalent constraint sets, it’s rather the transitive closure which is canonical, on which minimization performs a priori less simplifications. Maybe we should try that though to be more stable.

mattam82:
Another way would be to minimize wrt the uGraph but that involves adding more code in the kernel for that purpose and it might not be easy as the graph structure is backwards for this computation (you want the predecessors rather than the successors in that case).

SkySkimmer:
For restrict a way to remove a universe from the ugraph while preserving relations between others that go through the removed universe would be useful.

SkySkimmer:
>the graph structure is backwards for this computation

Is there a reason for that or could it be switched? It doesn't seem like the ugraph would care if it's talking about `<` or `>`.

gares:
Wait a sec, I'm afraid we are not understanding each other.
I'm not talking about the edges of the graph. I do agree that transitively closing two graphs that are semantically equivalent makes them also syntactically equivalent. So this is probably the way to go if one wants to have a stable notion of "normal" graph.

*BUT* I'm not talking about the edges, I'm talking about the *nodes* (the things sitting on the left of `|=`, the ones that become parameters of the definitions).
It can't make sense that a node is considered garbage (ie. not reachable by a sub graph) if you look at the graph as it is or to its transitive closure. It is a semantic property.
I understand that the set of edges/nodes that are necessary in order to typecheck a term is not easy to deremine without putting the term and its inner types in normal form (and you don't want that). This was my question at the working group. But once you have a graph of constraints coming from tactics/pretypes, the notion of garbage is, IMO, well defined and stable.

What am I missing.

SkySkimmer:
Minimisation has a semantic effect (it doesn't preserve the semantic equivalence) and it isn't `Proper` as setoids say.

You define `foo{a b c} (* a <= b, b <= c *)`. When you use it somewhere universes `a0, b0, c0` are produced. Then minimization adds `a0 = b0 = c0`.
With the new restrict you get `foo@{a b c} (* a <= b, b <= c, a <= c *)` (because (a,c) gets processed first for whatever reason) and this breaks minimization (see NB) so when you use it you get 3 universes when you only wanted one.
(NB: this is a fake example, getting a real example of minimisation being sensitive is beyond the time I want to spend on it)

mattam82:
In other words, restrict is not the culprit, it’s the minimization done at
uses of the result of restrictions that’s unstable. Is that clearer @gares ?
— Matthieu
Le ven. 18 mai 2018 à 11:09, Gaëtan Gilbert <notifications@github.com> a
écrit :

gares:
No, and I'm afraid it is not clean because I don't have a spec for minimization.
I do have one for restriction (it is exactly a GC). Can you give me one for minimization, so that I can understand why there is not such a thing as absolute (unique) minimum?

mattam82:
Minimization tries to lower the flexible universes to their minimal level.
If the constraints contain redundant constraints then a universe might get
two or more predecessors instead of one and not be lowered (unless it is
also algebraic). If it has one predecessor, we transform the <= into a =
and proceed recursively.
— Matthieu

gares:
I try to craft an example (`v` variables, `x` globals):
```
{ x1, x2, v |= x1 <= x2, x2 <= v }          -> { x1, x2    |= x1 <= x2, v = x2 }
{ x1, x2, v |= x1 <= x2, x2 <= v, x1 <= v } -> { x1, x2, v |= ... }
```

JasonGross:
> If the constraints contain redundant constraints then a universe might get two or more predecessors instead of one and not be lowered (unless it is also algebraic).  If it has one predecessor, we transform the <= into a = and proceed 

If, before checking for uniqueness of predecessors, you remove all predecessors which are less than or equal to all other predecessors (and then put one back if you've removed them all), does that fix the sensitivity to transitive closure?

SkySkimmer:
New commit is a better version 
```
Smarter fix for restrict_universe_context using the UGraph

This time the shortcut (if is_empty removed ...) is unecessary. See
pipeline https://gitlab.com/SkySkimmer/coq/pipelines/22278594 which
has the shortcut disabled (and a slight bug because instead of using
enforce_eq_level I added the constraint sometimes getting u = u.
```
I'll leave it with separate commits for a bit so people can have a look.

SkySkimmer:
For the new version (vs master):
https://ci.inria.fr/coq/view/benchmarking/job/benchmark-part-of-the-branch/434/console
```

┌──────────────────────────┬─────────────────────────┬───────────────────────────────────────┬───────────────────────────────────────┬─────────────────────────┬─────────────────┐
│                          │      user time [s]      │              CPU cycles               │           CPU instructions            │  max resident mem [KB]  │   mem faults    │
│                          │                         │                                       │                                       │                         │                 │
│             package_name │     NEW     OLD PDIFF   │            NEW            OLD PDIFF   │            NEW            OLD PDIFF   │     NEW     OLD PDIFF   │ NEW OLD PDIFF   │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                coq-sf-lf │   20.72   20.94 -1.05 % │    56586894766    56654246973 -0.12 % │    73891870225    74069232425 -0.24 % │  429656  429608 +0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│   coq-mathcomp-ssreflect │   45.30   45.37 -0.15 % │   123806720642   123799292129 +0.01 % │   145548136255   145771497523 -0.15 % │  536796  536716 +0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                coq-flocq │  263.12  263.14 -0.01 % │   731219442568   731902717211 -0.09 % │   904156796927   905285612183 -0.12 % │ 1245728 1245560 +0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│               coq-sf-plf │   53.98   53.98 +0.00 % │   148727380769   148430869002 +0.20 % │   187815126446   187950661875 -0.07 % │  514592  514412 +0.03 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│              coq-bignums │   80.01   79.92 +0.11 % │   220489851144   220318437126 +0.08 % │   285975572343   286193859558 -0.08 % │  519628  519620 +0.00 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│     coq-mathcomp-algebra │  185.96  185.71 +0.13 % │   513703263527   512680842050 +0.20 % │   692541240217   691814201723 +0.11 % │  652800  644336 +1.31 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│               coq-geocoq │ 3102.08 3096.87 +0.17 % │  8607052732530  8591803586226 +0.18 % │ 13801076120917 13796676884727 +0.03 % │ 1239940 1240020 -0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                coq-color │  722.42  720.76 +0.23 % │  1997153880301  1993083136609 +0.20 % │  2380268365514  2379847535519 +0.02 % │ 1445680 1445768 -0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│    coq-mathcomp-solvable │  203.03  202.55 +0.24 % │   561273093929   559138979266 +0.38 % │   771762732374   771080124299 +0.09 % │  839696  842096 -0.29 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│             coq-compcert │  870.01  867.28 +0.31 % │  2409939537904  2403686186618 +0.26 % │  3481007314018  3481556723029 -0.02 % │ 1387860 1387936 -0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│ coq-mathcomp-real-closed │  171.90  171.33 +0.33 % │   476570138411   475428734025 +0.24 % │   695992758165   696231970223 -0.03 % │  838800  838260 +0.06 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│       coq-mathcomp-field │  460.27  458.73 +0.34 % │  1277819418902  1272754858352 +0.40 % │  2055922942355  2056146495448 -0.01 % │  809880  808924 +0.12 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                 coq-hott │  336.38  335.20 +0.35 % │   935005249418   927958727299 +0.76 % │  1432342299484  1432631241535 -0.02 % │  588980  588912 +0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│              coq-unimath │ 1296.82 1292.27 +0.35 % │  3592638477235  3578213482818 +0.40 % │  6048026234513  6047947608826 +0.00 % │ 1079620 1079608 +0.00 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│          coq-fiat-crypto │ 4116.47 4099.58 +0.41 % │ 11423720793012 11376309732568 +0.42 % │ 18432090209390 18433184974321 -0.01 % │ 3188188 3195900 -0.24 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│            coq-fiat-core │  112.06  111.59 +0.42 % │   310915867493   310488028499 +0.14 % │   379974873653   380017034773 -0.01 % │  498240  497644 +0.12 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                 coq-corn │ 1569.84 1562.97 +0.44 % │  4341219649599  4322259729286 +0.44 % │  6439793133435  6439697642863 +0.00 % │  933372  933504 -0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│           coq-coquelicot │   83.74   83.36 +0.46 % │   231267932037   230913631171 +0.15 % │   275685961424   275903096266 -0.08 % │  721424  720880 +0.08 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│   coq-mathcomp-character │  276.42  275.13 +0.47 % │   766367056434   762832465191 +0.46 % │  1086733439366  1086875019240 -0.01 % │ 1143688 1142244 +0.13 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│         coq-fiat-parsers │  704.06  700.61 +0.49 % │  1950895020109  1938700556753 +0.63 % │  2974055646686  2973910848249 +0.00 % │ 3378628 3376620 +0.06 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│         coq-math-classes │  228.62  227.47 +0.51 % │   626481108311   624990542994 +0.24 % │   780713787090   780832986925 -0.02 % │  533480  532120 +0.26 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│                  coq-vst │ 3853.29 3833.01 +0.53 % │ 10717454939540 10657425284106 +0.56 % │ 14007728670257 14007469696424 +0.00 % │ 2228092 2227588 +0.02 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│               coq-sf-vfa │   26.56   26.42 +0.53 % │    72573328792    72412050020 +0.22 % │    94212802699    94368738572 -0.17 % │  537092  537168 -0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│    coq-mathcomp-fingroup │   65.62   65.08 +0.83 % │   179771023970   178827292739 +0.53 % │   239795638550   239263004142 +0.22 % │  588928  588412 +0.09 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│   coq-mathcomp-odd-order │ 1443.08 1429.46 +0.95 % │  4010343917347  3971055820611 +0.99 % │  6663998632137  6661443108835 +0.04 % │ 1374596 1368920 +0.41 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼─────────────────┤
│      coq-formal-topology │   37.98   37.57 +1.09 % │   101550100197   101540779000 +0.01 % │   126597350115   126737890258 -0.11 % │  482496  482284 +0.04 % │   0   0  +nan % │
└──────────────────────────┴─────────────────────────┴───────────────────────────────────────┴───────────────────────────────────────┴─────────────────────────┴─────────────────┘


```

kernel/uGraph.ml
```diff
@@ -786,6 +786,42 @@ let constraints_of_universes g =
   let csts = UMap.fold constraints_of g.entries Constraint.empty in
   csts, UF.partition uf
 
+(* domain g.entries = kept + removed *)
+let constraints_for ~kept g =
+  (* rmap: partial map from canonical universes to kept universes *)
+  let rmap, csts = LSet.fold (fun u (rmap,csts) ->
+      let arcu = repr g u in
+      if LSet.mem arcu.univ kept then
+        LMap.add arcu.univ arcu.univ rmap, enforce_eq_level u arcu.univ csts
```
mattam82: Ok, we add u = u so that the LMap.find on kept canonical universes works.

kernel/uGraph.ml
```diff
@@ -786,6 +786,42 @@ let constraints_of_universes g =
...
+      else
+        match LMap.find arcu.univ rmap with
+        | v -> rmap, enforce_eq_level u v csts
+        | exception Not_found -> LMap.add arcu.univ u rmap, csts)
```
mattam82: Here I'm a bit worried: rmap is in the process of being built, so we depend on the fold order somehow?

Otherwise, it looks to me that indeed, constraints_for will for sure not return constraints mentioning anything else than the kept universes.

mattam82 (approved):

SkySkimmer:
Squashed.

kernel/uGraph.ml
SkySkimmer: >Here I'm a bit worried: rmap is in the process of being built, so we depend on the fold order somehow?

The fold order impacts which kept universe is canonical when the ugraph considers a removed universe to be canonical.

mattam82: @SkySkimmer could you confirm or infirm my worry?

SkySkimmer: What is the worry exactly? ie what can go wrong?

mattam82: Good, thanks for the explanation.

gares:
Maybe this is not the right place but I'd like to know the answer to the point raised by @JasonGross, that is also where I wanted to go with my example: understand if this minimization can be specified at the semantic level of the graph, and not at its syntactic level.

That is "having 1 syntactic predecessor" makes no sense, since the two graphs in my example are semantically equivalent but only in one case the node `v` is purged.
Intuitively `v` has to be purged (made `=` to `x`) if:
- `v` has no successors (strict or weak, eg `v` is a node on the border of the graph)
- `x` is the max of all the predecessors of `v`.
If there is a node `x` with such property then `v=x` (`v` is collapsed to `x`).

Does this make sense?

mattam82:
I don’t see how to do this properly without basically doing a transitive
closure computation. Ie when there is more than one predecessor, check if
one of them is above all the others in the transitive closure.

herbelin:
@mattam82, @SkySkimmer:

I have a somewhat related minimization question. Minimization is currently done by joining universes different from `Prop` and `Set`. Would it be easy to instrument it so that it knows how to minimize using `Prop` for the universes which are not constrained below and using `Set` for the universes which are constrained only by `Set` below. This is in relation with #7369: in practice, there are two kinds of existential variables for universes, either `Type` where I did not mention the level, or `?[X]` when it eventually happens that it has to be instantiated by some sort. In the latter case, `Prop` and `Set` should remain potential solutions and that would be useful for #7369 that minimization handles this case.

I tried to hack `Universe.compute_lbound` in this direction but that's working "too often" and it seems that what I fail to see is the meaning of `uctx_univ_variables` and `uctx_univ_algebraic`.

When a variable occurs in `uctx_univ_variables` or `uctx_univ_algebraic`, is there an implicit assumption that it should be higher than `Set`? According to `UState.new_univ_variable` it is only greater or equal to `Set`, so `Set` should in theory be a possible instance, but then, in minimization, `Set` seems to be forbidden as a possible solution for such a variable.

Conversely, if I'd like in the cases above that some given universe variable is allowed to be minimized to `Set` (or even to `Prop` using the variant of `UState.new_univ_variable` in #7369), should a specific information be added somewhere in the `UState.t` or is this information retrievable from the data already present?

Typical examples where this would be useful are the following:
```coq
Check fun b : bool => if b then bool else True.  (* works *)
Check fun b : bool => if b then True else bool.  (* fails *)
```
or
```coq
Check bool = True. (* works *)
Check True = bool. (* fails *)
```
Both failing examples would work and return the same sort as in the succeeding examples if we were able in #7369 to minimize the graph `Set <= u` for `u` allowed to be `Set`.

gares:
> I don’t see how to do this properly without basically doing a transitive closure computation.

AFAICT (from the comments) this PR does the TC of the graph. Am I wrong? (otherwise my example is bogus).

SkySkimmer:
>AFAICT (from the comments) this PR does the TC of the graph. 

It stops at the first kept universe in any transitive chain.

gares:
OK. Anyway, is it true that if we perform/reason-on the TC closure of the graph then restriction/minimization is a semantic property and not a syntactic one?

I agree it is still an heuristic in the sense that one could be smarter and collapse more/less, but if we fix the criteria as in https://github.com/coq/coq/pull/7495#issuecomment-391328473 then: is the algorithm working at the semantic level?

mattam82:
Yes, it can be made a semantic property of the TC of the graph. Actually, it might be feasible, I did not experiment.

mattam82:
@gares what do you think, shall we merge this cleanup? @SkySkimmer it seems the HoTT failure happens no more, as there's no overlay and it works on gitlab, right?
@herbelin could you open a separate issue about the treatment of Prop/Set please?

SkySkimmer:
>@SkySkimmer it seems the HoTT failure happens no more, as there's no overlay and it works on gitlab, right?

Yes.

herbelin:
>  @herbelin could you open a separate issue about the treatment of Prop/Set please?

I shall maybe do at some time (I'm in workshop this week and my time is pretty limited). #7369 covers part of the state of my reflection otherwise.

gares:
The cleanup should be merged IMO.
I'm not competent enough to try out the TC thingy, so I hope someone else does.

mattam82:
Thanks for the answers everyone, I'll merge.

gares:
It's fine, I'll eventually uncomment restriction in elpi, I should be able to update the branch we track in CI next week.

PR #7482, opened 2018-05-11
Update CI documentation following recent evolutions.

Zimmi48:
**Kind:** documentation

This allows checking the two first items at #6807.

Further improvements possible.

PR #7503, opened 2018-05-14
[ci] [circleci] Remove jobs done in Gitlab efficiently.

ejgallego:
Following the migration to Gitlab (#6919) we reduce Circle load, see
also discussion in #7436 and #7482 .

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -160,22 +160,22 @@ workflows:
           requires:
             - build
             - bignums
-      - compcert: *req-main
-      - coq-dpdgraph: *req-main
-      - coquelicot: *req-main
-      - cross-crypto: *req-main
-      - elpi: *req-main
-      - equations: *req-main
-      - geocoq: *req-main
-      - fcsl-pcm: *req-main
-      - fiat-crypto: *req-main
-      - fiat-parsers: *req-main
-      - flocq: *req-main
+      # - compcert: *req-main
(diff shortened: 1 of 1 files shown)


SkySkimmer (commented):
Why not go all the way and remove circle?

ejgallego:
> Why not go all the way and remove circle?

Eventually we may do that I guess, but I'd like to go step by step. Gitlab still needs more testing IMO, so it is good the leave the fallbacks in place for now.

ejgallego:
This just reduces the load on circle a lot so overall latency should improve.

SkySkimmer:
It reduces the load a lot because there isn't much left, so in what way is it a fallback?

ejgallego:
> It reduces the load a lot because there isn't much left, so in what way is it a fallback?

We still build Coq and a few contribs there; so that's an extra sanity check.

ejgallego:
We will eventually drop it, but for this kind of stuff I like to go slooow.

ejgallego:
slooow = let's have Gitlab do a few more thousand builds and then we'll see.

ejgallego:
Another heavy dependency that's not properly captured in Circle is math-comp. But again I dunno what to do in Gitlab, using stages would be bad for latency, but indeed maybe that's the best after all.

SkySkimmer (approved):

PR #7505, opened 2018-05-14
Pick up user overlays when running GitLab CI on PRs.

Zimmi48:
**Kind:** infrastructure.

I recently noticed that user-overlays were not picked up when running GitLab CI on PRs (because the branch name is `pr-NNNN` and the `CI_PULL_REQUEST` variable is not defined). This PR fixes this by finding the pull request number from the branch name and defining `CI_PULL_REQUEST`.

### DIFF
--- a/dev/ci/ci-common.sh
+++ b/dev/ci/ci-common.sh
@@ -10,6 +10,10 @@ if [ -n "${GITLAB_CI}" ];
 then
     export COQBIN="$PWD/_install_ci/bin"
     export CI_BRANCH="$CI_COMMIT_REF_NAME"
+    if [[ ${CI_BRANCH#pr-} =~ ^[0-9]*$ ]]
+    then
+        export CI_PULL_REQUEST="${CI_BRANCH#pr-}"
+    fi
 else
     if [ -n "${TRAVIS}" ];
     then


SkySkimmer:
The alternative would be to have overlays test for the branch name directly.
Considering we don't run anything using overlays on travis PRs any more that may make more sense than emulating a travis variable.

Zimmi48:
That's not really a Travis variable, see `export CI_PULL_REQUEST="$TRAVIS_PULL_REQUEST"`.
If we choose another way, this code should be cleaned up anyways.

SkySkimmer:
CI_PULL_REQUEST is emulating a travis variable.

Gaëtan Gilbert

ejgallego:
> The alternative would be to have overlays test for the branch name directly.

You mean adding `a CI_BRANCH_NAME = 'pr-XXXX'` test?

SkySkimmer:
Yes

Gaëtan Gilbert

ejgallego:
> Yes

That's what I've been doing, I dunno, I think this change makes sense. Only point is that we have to keep it sync with the bot, but that doesn't seem like an issue to me.

Zimmi48:
Until CircleCI is removed, we still need to support it as well, so that means three possibilities in the test. But anyways, the workaround you propose means that there is no urgency in merging this so we can discuss a bit more.

ejgallego (approved):
Ideally overlays would not even need the test as the system would detect their presence and apply them automatically, so this change is welcome.

ejgallego:
Umm, for example in https://gitlab.com/coq/coq/-/jobs/68094691 I don't see the variable set.

Zimmi48:
@ejgallego Not sure what you meant in your comment above but the link you gave isn't for a build for this PR. Here's one for this PR https://gitlab.com/coq/coq/-/jobs/68070514 and we can read:

```
+ export CI_PULL_REQUEST=7505
```

ejgallego:
Pss, I explicitly checked that I was in the good PR, but even so I got it wrong. Sorry for the noise.

PR #7507, opened 2018-05-14
gitlab CI: fix [warnings] template

SkySkimmer:
We never actually used the -warn-error flag...

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -76,7 +76,7 @@ before_script:
 
     - set +e
   variables: &warnings-variables
-    COQ_EXTRA_CONF: "-native-compiler yes -coqide byte -byte-only"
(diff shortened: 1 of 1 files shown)


Zimmi48:
We might have used it correctly at the beginning. The `-warn-error` flag changed recently. However, it is bad that it doesn't fail when used like that.

SkySkimmer:
This doesn't have anything to do with the flag's change.
You can look at the initial (merged) gitlab commit, it was never used https://github.com/coq/coq/commit/149b0c422027a31972b62457af1bf97bd016e26c#diff-96edf7a6f008de9e928d04e1ae5e12a5R96

ejgallego (approved):

ejgallego:
Certainly the way redefinitions work in yaml + the need to remember to include each subsection is a big source of bugs.

PR #7514, opened 2018-05-14
[ci] Don't build lite versions of CI developments.

ejgallego:
In the original Travis CI setup, the per-job time limit was an
issue. However, Gitlab has much improved this problem due to

 a) Coq not being built for each contrib,
 b) user-configurable time limit.

We thus disable the expensive builds from Travis and instruct Gitlab
to build [fiat-crypto, geocoq, math-comp, unimath, vst] in full.

We also update the math-comp script as the odd-order theorem lives in
a separate repository and it is a key CI case.

### DIFF
--- a/dev/ci/ci-fiat-crypto.sh
+++ b/dev/ci/ci-fiat-crypto.sh
@@ -6,6 +6,9 @@ ci_dir="$(dirname "$0")"
 fiat_crypto_CI_DIR="${CI_BUILD_DIR}/fiat-crypto"
 
 git_checkout "${fiat_crypto_CI_BRANCH}" "${fiat_crypto_CI_GITURL}" "${fiat_crypto_CI_DIR}"
+
 ( cd "${fiat_crypto_CI_DIR}" && git submodule update --init --recursive )
 
-( cd "${fiat_crypto_CI_DIR}" && make lite lite-display )
+fiat_crypto_CI_TARGETS="lite lite-display"
+
+( cd "${fiat_crypto_CI_DIR}" && make ${fiat_crypto_CI_TARGETS} )
--- a/dev/ci/ci-math-comp.sh
+++ b/dev/ci/ci-math-comp.sh
@@ -5,11 +5,10 @@ ci_dir="$(dirname "$0")"
 . "${ci_dir}/ci-common.sh"
 
 mathcomp_CI_DIR="${CI_BUILD_DIR}/math-comp"
+oddorder_CI_DIR="${CI_BUILD_DIR}/odd-order"
 
-checkout_mathcomp "${mathcomp_CI_DIR}"
+git_checkout "${mathcomp_CI_BRANCH}" "${mathcomp_CI_GITURL}" "${mathcomp_CI_DIR}"
+git_checkout "${oddorder_CI_BRANCH}" "${oddorder_CI_GITURL}" "${oddorder_CI_DIR}"
 
-# odd_order takes too much time for travis.
-( cd "${mathcomp_CI_DIR}/mathcomp"                  && \
-  sed -i.bak '/PFsection/d'                  Make && \
-  sed -i.bak '/stripped_odd_order_theorem/d' Make && \
-  make Makefile.coq && make -f Makefile.coq all )
+( cd "${mathcomp_CI_DIR}/mathcomp" && make && make install )
+( cd "${oddorder_CI_DIR}/" && make )
--- a/dev/ci/ci-vst.sh
+++ b/dev/ci/ci-vst.sh
@@ -5,9 +5,6 @@ ci_dir="$(dirname "$0")"
 
 VST_CI_DIR="${CI_BUILD_DIR}/VST"
 
-# opam install -j ${NJOBS} -y menhir
 git_checkout "${VST_CI_BRANCH}" "${VST_CI_GITURL}" "${VST_CI_DIR}"
 
-# We have to omit progs as otherwise we timeout on Travis; on Gitlab
-# we will be able to just use `make`
-( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true -o progs )
+( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true )
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,13 +87,7 @@ matrix:
       - TEST_TARGET="ci-equations"
     - if: NOT (type = pull_request)
       env:
-      - TEST_TARGET="ci-geocoq"
-    - if: NOT (type = pull_request)
-      env:
       - TEST_TARGET="ci-fcsl-pcm"
-    - if: NOT (type = pull_request)
-      env:
-      - TEST_TARGET="ci-fiat-crypto"
     - if: NOT (type = pull_request)
       env:
       - TEST_TARGET="ci-fiat-parsers"
@@ -102,22 +96,13 @@ matrix:
       - TEST_TARGET="ci-flocq"
     - if: NOT (type = pull_request)
       env:
-      - TEST_TARGET="ci-formal-topology"
-    - if: NOT (type = pull_request)
-      env:
       - TEST_TARGET="ci-hott"
-    - if: NOT (type = pull_request)
-      env:
-      - TEST_TARGET="ci-iris-lambda-rust"
     - if: NOT (type = pull_request)
       env:
       - TEST_TARGET="ci-ltac2"
     - if: NOT (type = pull_request)
       env:
       - TEST_TARGET="ci-math-classes"
-    - if: NOT (type = pull_request)
-      env:
-      - TEST_TARGET="ci-math-comp"
     - if: NOT (type = pull_request)
       env:
       - TEST_TARGET="ci-mtac2"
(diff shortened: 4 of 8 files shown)


dev/ci/ci-math-comp.sh
```diff
@@ -6,10 +6,6 @@ ci_dir="$(dirname "$0")"
 
 mathcomp_CI_DIR="${CI_BUILD_DIR}/math-comp"
 
-checkout_mathcomp "${mathcomp_CI_DIR}"
```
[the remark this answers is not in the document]
ejgallego: Killed.

dev/ci/ci-vst.sh
```diff
@@ -10,4 +10,4 @@ git_checkout "${VST_CI_BRANCH}" "${VST_CI_GITURL}" "${VST_CI_DIR}"
 
 # We have to omit progs as otherwise we timeout on Travis; on Gitlab
 # we will be able to just use `make`
-( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true -o progs )
+( cd "${VST_CI_DIR}" && make IGNORECOQVERSION=true )
```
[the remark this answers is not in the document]
ejgallego: Done.

ejgallego:
> Did you intend to base this on the PR removing jobs from circle CI?

Oh indeed that's a prerequisite, sorry for that. In fact this may need further adjustment, `formal-topology` and `iris` are both usually very close to the Travis limit too so maybe they should go.

ejgallego:
#7503 needs to be merged first in order to avoid Circle timeouts.

dev/ci/ci-fiat-crypto.sh
```diff
@@ -8,4 +8,4 @@ fiat_crypto_CI_DIR="${CI_BUILD_DIR}/fiat-crypto"
 git_checkout "${fiat_crypto_CI_BRANCH}" "${fiat_crypto_CI_GITURL}" "${fiat_crypto_CI_DIR}"
 ( cd "${fiat_crypto_CI_DIR}" && git submodule update --init --recursive )
 
-( cd "${fiat_crypto_CI_DIR}" && make lite lite-display )
+( cd "${fiat_crypto_CI_DIR}" && make )
```
JasonGross: This should be `make coq selected-specific-display`, or, to build even a bit more, `make coq nonautogenerated-specific nonautogenerated-specific-display`.  (The `*-display` targets are not built by default.)

JasonGross (changes requested):

dev/ci/ci-fiat-crypto.sh
ejgallego: Thanks @JasonGross , PR updated. What to build is your call folks.

JasonGross (approved):

dev/ci/ci-fiat-crypto.sh
ejgallego: Umm, with this configuration fiat-crypto fails:
```
File "./src/Curves/Weierstrass/Projective.v", line 130, characters 3-7:
Error: Attempt to save an incomplete proof (in proof exceptional_2P_eq_2Q)
```
any idea @JasonGross ?

ejgallego: That's with `fiat_crypto_CI_TARGETS='coq nonautogenerated-specific nonautogenerated-specific-display'`

ejgallego: https://gitlab.com/coq/coq/-/jobs/68227040

JasonGross: Very strange.  It builds fine on master on our travis: https://travis-ci.org/mit-plv/fiat-crypto/jobs/378977060#L895 (This is with 8.master~git~201805141404+19072-0~daily308~ubuntu14.04.1 from https://launchpad.net/~jgross-h/+archive/ubuntu/coq-master-daily)

JasonGross: I think this is most likely a bug in Coq on that platform, given that [the line it's pointing at is `Qed`](https://github.com/mit-plv/fiat-crypto/blob/251ea49a661aef7c075ace80867006183ab0cdea/src/Curves/Weierstrass/Projective.v#L130) and the line before that is `par: abstract solve [ fsatz | cbv [id] in * ; fsatz ].` and there should be no way to get that error message after `par: abstract` (I can get `Attempt to save a proof with shelved goals`, but this is a different message.)

ejgallego: Interesting, let's rebuild!

Maybe `par` is not fully deterministic, but CI is using flambda on 4.06.0 so maybe that could be a factor.

ejgallego: https://gitlab.com/coq/coq/-/jobs/68243830

JasonGross: Same error.  I wonder if this is related to the bug that SIGINT in the middle of par crashes/kills coqtop...

ejgallego: Yeah something bizarre is going on; I guess that this is due to `par` going OOM in these workers. So what would be a more conservative setting? I'll try with `selected-specific-display`

JasonGross: Peak memory usage of 1353868 ko should not make Coq OOM.  (Peak memory usage of the failed file on gitlab is 884004 ko, and peak memory usage of the only file running at the same time is 468468 ko.)  We don't have a target other than `lite` that excludes `Projective.v` (it's included in `coq`).  I could make one (though I'm a bit hesitant to add yet another target to our current dozen or so), but I think it's reasonable to expect that this should work.

JasonGross: Maybe you should try running fiat-crypto with `-j1` instead of with `-j2` or whatever?

Zimmi48 (approved):
Approval on the principle. I let the details to @SkySkimmer.

SkySkimmer:
I'm pretty sure fiat crypto is failing because of a memory issue. If I limit the memory usage of all coq processes to 3GB I can reproduce every time, at 4GB never.

SkySkimmer:
>Peak memory usage of 1353868 ko should not make Coq OOM. (Peak memory usage of the failed file on gitlab is 884004 ko, and peak memory usage of the only file running at the same time is 468468 ko.) We don't have a target other than lite that excludes Projective.v (it's included in coq). I could make one (though I'm a bit hesitant to add yet another target to our current dozen or so), but I think it's reasonable to expect that this should work.

`time` doesn't see the memory usage of the `par` workers.
Consider `foo.v`:
```coq
Ltac loop i acc := match i with 0 => idtac | S ?i => loop i constr:(cons i acc) end.

Goal True /\ False.
  split.
  all: (idtac; match goal with |- True => exact I | |- False => loop 1000 constr:(@nil nat) end).
Abort.
```
```script
$ time -f "(real: %e, user: %U, sys: %S, mem: %M ko)" coqc foo.v
(real: 9.00, user: 8.90, sys: 0.06, mem: 309264 ko)

$ sed -i.bak 's/all:/par:/' foo.v
$ time -f "(real: %e, user: %U, sys: %S, mem: %M ko)" coqc foo.v
(real: 9.55, user: 0.04, sys: 0.02, mem: 58012 ko)
```
So only the `real` time is useful when using `par`.

ejgallego:
Yup my impression was that `par` is extremely demanding memory-wise, thanks for confirming @sampablokuper . So `fiat-crypto` will have to be kept on the `lite` target for now it seems.

Zimmi48:
Indeed formal-topology and iris-lambda-rust do time out often so please remove them from Travis as well:
- https://travis-ci.org/coq/coq/builds/378619461
- https://travis-ci.org/coq/coq/builds/378685865

JasonGross:
> So fiat-crypto will have to be kept on the lite target for now it seems.

I've added a target `nobigmem`, so you can now try `make print-nobigmem nobigmem nonautogenerated-specific nonautogenerated-specific-display`.  Does that work?

SkySkimmer:
This is on an unrelated commit: https://gitlab.com/coq/coq/-/jobs/68457413
 >Makefile:119: *** invalid syntax in conditional.  Stop.

Gaëtan Gilbert

JasonGross:
> Makefile:119: *** invalid syntax in conditional.  Stop.

Oops, that's what I get for not testing locally.  Sorry.  Should be fixed now.

ejgallego:
> I've added a target nobigmem, so you can now try make print-nobigmem nobigmem nonautogenerated-specific nonautogenerated-specific-display. Does that work?

Thanks @JasonGross , build with the new targets pushed.

ejgallego:
Fiat crypto changes reverted.

ejgallego:
> Indeed formal-topology and iris-lambda-rust do time out often so please remove them from Travis as well:

Granted.

JasonGross:
@ejgallego can you try `make print-nobigmem nobigmem nonautogenerated-specific nonautogenerated-specific-display` for fiat-crypto once more?  I've removed what I think should be all of the major offenders (rather than just the one issue from before).

ejgallego:
@JasonGross see #7525

Issue #7161, opened 2018-04-03
Nightly builds roadmap


Zimmi48:
Providing nightly Coq builds would simply mean having a webpage where users can download a binary package corresponding to the latest version of Coq master branch.
This should include at least the packages we provide now for released versions (MacOS, Windows 32 bits and Windows 64 bits) but it would be cool to provide more (e.g. a Debian / Ubuntu `.deb` binary package, a Nix package with a binary cache, etc).

Nightly builds are useful to ensure that all our users have the possibility of testing development (alpha) versions, including those who are on platforms where compiling Coq from sources is harder. In particular, developers of Coq packages we test in CI must be able to run their packages with the latest version of Coq master.
Additionally, providing a pre-compiled Linux package (`.deb` or Nix) would help gain build time for people testing their package on CI with the development version of Coq.

We are already quite close to providing nightly builds. Here are the steps I see:

- [x] Build MacOS and Windows packages on CI.
- [ ] *(optional)* Build Linux binary packages on CI.
- [ ] Give a precise version number to these packages. See #483.
      What is the current status of this PR? @ejgallego do you intend to take care of finishing it and do you think that Dune is going to help here?
- [ ] Choose where to host these packages. Because Travis doesn't provide native build artifact support, the MacOS package currently uses bintray. With a proper version number, it would make sense to use this platform for nightly builds and we could upload Windows package over there as well.

ejgallego:
> What is the current status of this PR? @ejgallego do you intend to take care of finishing it and do you think that Dune is going to help here?

Yes and yes.

JasonGross:
Note that I have ~nightly builds at https://launchpad.net/~jgross-h/+archive/ubuntu/coq-master-daily, https://launchpad.net/~jgross-h/+archive/ubuntu/coq-8.8-daily, etc, and packages for the various released versions at https://launchpad.net/~jgross-h/+archive/ubuntu/many-coq-versions .  I do not endorse using the scripts that I'm using to do this; they are cargo-culted from the released debian packages and have many artifacts because of that.  I made them basically just so that I could test whatever versions of Coq I wanted on travis without needing to wait for them to build.

Zimmi48:
@maximedenes @ejgallego @SkySkimmer once the MacOS and Windows packages are built on GitLab CI, will there be a natural place to store them and distribute them to users?

Zimmi48:
[...]

Issue #6505, opened 2017-12-26
Provide a way for make doc to use installed Coq for its dependencies.


RalfJung:
Coq should provide a make target that builds only documentation, so that we can offer a `coq-doc` opam package. Also see https://github.com/coq/coq/issues/4855.

herbelin:
Maybe I did not understand the question, but isn't `make doc` what you'd like to do? This builds all of:
- the reference manual
- the "tutorial"
- the tutorial on inductive types
- the coqdoc-ization of the standard library

(Notice in particular that this needs to recompile all the standard library.)

RalfJung:
> this needs to recompile all the standard library

That's pretty annoying for the opam use case... we want to have the documentation in a separate package so that the main coq package doesn't have to depend on sphinx or whatever. If all this needs is the .vo files, can't it use the ones that are installed?

Zimmi48:
Indeed, it should be theoretically doable to make the doc OPAM package depend on the Coq OPAM package and reuse the installed `.vo`. It will probably require tweaking the make targets, but I think that is a good strategy.

SkySkimmer:
The sphinx part is quite easy:
- point COQBIN to the installed coq at https://github.com/coq/coq/blob/7dd881fc72d62eb0c1f1e5063eb3a8ed268fb5d5/Makefile.doc#L54
- remove -boot at https://github.com/coq/coq/blob/7dd881fc72d62eb0c1f1e5063eb3a8ed268fb5d5/doc/tools/coqrst/repl/coqtop.py#L45

ejgallego:
See also #16198

Zimmi48:
The related issue precisely indicates that the present issue is solved since the Dune consolidation PR. It is the other case (building the doc with a non-installed Coq) that has become more complicated. Thus, a coq-doc package could now be introduced in the core-dev repo.

ejgallego:
Yes, let's close this as "completed" as it works now (and in fact this is how the CI builds the manual)

Note that with recent dune directory targets work pretty reasonably, so maybe for 8.18 we should dump the `coq-doc` dune dep and finally fix https://github.com/coq/coq/issues/4855

PR #7515, opened 2018-05-14
gitlab: build sphinx doc in separate job

SkySkimmer:
The system to avoid passing BOOT to coqtop while sphinxing is a bit inelegant, suggestions welcome.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -55,6 +55,7 @@ before_script:
     paths:
       - _install_ci
       - config/Makefile
+      - config/coq_config.py
       - test-suite/misc/universes/all_stdlib.v
     expire_in: 1 week
   script:
@@ -70,7 +71,7 @@ before_script:
 
     - echo 'start:coq.build'
     - make -j "$NJOBS" byte
-    - make -j "$NJOBS"
+    - make -j "$NJOBS" world $EXTRA_TARGET
     - make test-suite/misc/universes/all_stdlib.v
     - echo 'end:coq:build'
 
@@ -108,6 +109,19 @@ before_script:
 # artifacts are used together and we may get some random Coq. To that
 # purpose, we add a spurious dependency `not-a-real-job` that must be
 # overridden otherwise the CI will fail.
+
+.doc-templare: &doc-template
+  stage: test
+  dependencies:
+    - not-a-real-job
+  script:
+    - SPHINXENV='COQBIN="'"$PWD"'/_install_ci/bin/" COQBOOT=no'
+    - make -j "$NJOBS" SPHINXENV="$SPHINXENV" SPHINX_DEPS= sphinx
+    - make install-doc-sphinx
+  artifacts:
+    name: "$CI_JOB_NAME"
+    paths:
+      - _install_ci/share/doc/coq/
 
 # set dependencies when using
 .test-suite-template: &test-suite-template
@@ -181,7 +195,9 @@ before_script:
 build:base:
   <<: *build-template
   variables:
-    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt -with-doc yes"
+    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
+    # coqdoc for stdlib, until we know how to build it from installed Coq
+    EXTRA_TARGET: "stdlib"
 
 # no coqide for 32bit: libgtk installation problems
 build:base+32bit:
@@ -227,6 +243,11 @@ warnings:edge:
   variables:
     <<: *warnings-variables
     OPAM_SWITCH: edge
+
+documentation:
+  <<: *doc-template
+  dependencies:
+    - build:base
 
 test-suite:base:
   <<: *test-suite-template
--- a/Makefile.doc
+++ b/Makefile.doc
@@ -31,6 +31,7 @@ DVIPS:=dvips
 HTMLSTYLE:=coqremote
 
 # Sphinx-related variables
+SPHINXENV:=COQBIN="$(CURDIR)/bin/"
 SPHINXOPTS= -j4
 SPHINXBUILD= sphinx-build
 SPHINXBUILDDIR= doc/sphinx/_build
@@ -55,7 +56,7 @@ endif
 
 sphinx: $(SPHINX_DEPS)
 	$(SHOW)'SPHINXBUILD doc/sphinx'
-	$(HIDE)COQBIN="$(abspath bin)" $(SPHINXBUILD) -W -b html $(ALLSPHINXOPTS) doc/sphinx $(SPHINXBUILDDIR)/html
+	$(HIDE)$(SPHINXENV) $(SPHINXBUILD) -W -b html $(ALLSPHINXOPTS) doc/sphinx $(SPHINXBUILDDIR)/html
 	@echo
 	@echo "Build finished. The HTML pages are in $(SPHINXBUILDDIR)/html."
 
--- a/doc/tools/coqrst/repl/coqtop.py
+++ b/doc/tools/coqrst/repl/coqtop.py
@@ -41,10 +41,13 @@ class CoqTop:
                            the ansicolors module)
         :param args:       Additional arugments to coqtop.
         """
+        BOOT = True
+        if os.getenv('COQBOOT') == "no":
+            BOOT = False
         self.coqtop_bin = coqtop_bin or os.path.join(os.getenv('COQBIN', ""), "coqtop")
         if not pexpect.utils.which(self.coqtop_bin):
             raise ValueError("coqtop binary not found: '{}'".format(self.coqtop_bin))
-        self.args = (args or []) + ["-boot", "-color", "on"] * color
+        self.args = (args or []) + ["-boot"] * BOOT + ["-color", "on"] * color
         self.coqtop = None
 
     def __enter__(self):


ejgallego:
Umm, I was under the impression that we didn't want to do this as to get a full package in phase 1, right?

I am OK with any choice.

.gitlab-ci.yml
```diff
@@ -48,7 +49,7 @@ before_script:
 
     - echo 'start:coq.build'
     - make -j "$NJOBS" byte
-    - make -j "$NJOBS"
+    - make -j "$NJOBS" world $EXTRA_TARGET
```
ejgallego: Psss, I am not happy to see such hacks, in fact, `coq_config.py` should have never been introduced in the first place like it is now [a configure target] but instead it should have been generated using a make rule [as the only thing it contains is the version]. But when I saw it I didn't want to complain as they doc people had enough of a party back then.

.gitlab-ci.yml
```diff
@@ -37,6 +37,7 @@ before_script:
     paths:
       - _install_ci
       - config/Makefile
+      - config/coq_config.py
```
ejgallego: Psss, I am not happy to see such hacks, in fact, coq_config.py should have never been introduced in the first place like it is now [a configure target] but instead it should have been generated using a make rule [as the only thing it contains is the version]. But when I saw it I didn't want to complain as they doc people had enough of a party back then.

Makefile.common
```diff
@@ -14,6 +14,9 @@
 # Executables
 ###########################################################################
 
+COQBIN:=$(PWD)/bin/
+export COQBIN
+
```
ejgallego: This looks very suspicious, in particular for non-local builds.

SkySkimmer:
I thought it was more about not having to figure out the separate 
building while migrating.

Gaëtan Gilbert

.gitlab-ci.yml
```diff
@@ -155,7 +168,9 @@ before_script:
 build:base:
   <<: *build-template
   variables:
-    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt -with-doc yes"
+    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
+    # coqdoc for stdlib, until we know how to build it from installed Coq
+    EXTRA_TARGET: "stdlib"
```
ejgallego: Why is this new target needed?

SkySkimmer:
To build it without building sphinx

Gaëtan Gilbert

ejgallego (commented):
I think this PR will have a better time if you actually split it in 2 [one for the changes to Make / Sphinx] and other for the changes to CI.

ejgallego:
> I thought it was more about not having to figure out the separate building while migrating.

I have no idea, all that I have is vague recollection. So CI-wise, my only question is what about the artifacts and here @Zimmi48 is the main person linking at them.

Am I correct in the current status of this PR won't persist the documentation artifact?

SkySkimmer:
If people are using the artifacts we could make the doc job artifact the 
whole install directory. It's a bit of redundant upload and if the main 
use of artifacts is looking at the doc it makes the download slower but 
not in a huge way.
I don't think anyone is actually using the artifacts though, considering 
they have some silly path baked in.

My main goal with this PR is to avoid having sphinx issues like the 
timeout one stop the entire pipeline.

Gaëtan Gilbert

ejgallego:
> My main goal with this PR is to avoid having sphinx issues like the
> timeout one stop the entire pipeline.

Indeed seems like a sensible goal, and I had also thought of that, not because of the errors but also due to the extra latency Sphinx introduced.

Thus, in general terms I'm OK with this PR provided the artifact issue is clarified [cc:  #7161 ]

ejgallego:
As far as my competences go, I can take care of the patch except for the `COQBIN` change, for that, you either revert it or get a `Makefile` maintainer to approve it.

Zimmi48:
I would also like to see the patch split in two: one part fixing #6505 and one part using this for GitLab CI.
I want to keep access to the doc artifacts (especially the manual). I think they are very useful to be able to review a documentation PR.
Overall, I'm in favor of this PR.

.gitlab-ci.yml
ejgallego: Ping.

Makefile.common
ejgallego: Is this still needed?

.gitlab-ci.yml
SkySkimmer: >To build it without building sphinx

ejgallego:
I think @Zimmi48 's concerns about failing are a bit less; this looks OK to me, however I would like to understand a bit more what's going on, I am not sure I follow the targets/path setting here.

Could you answer my questions @SkySkimmer ? Merci.

Makefile.common
SkySkimmer: It's so we can override it in gitlab: `make COQBIN=...`

ejgallego: But my worry here is that the path may not be correct, right? In non-local builds we should use `DESTDIR`?

ejgallego:
I am unable to understand this, I guess I am too tired to look at Makefiles. I'll try again soon, but if someone can review the make part I'll merge happily.

Makefile.common
gares: I'm against this change. If you need `$COQBIN` to hold a good value, export it in the CI script.

Makefile.doc
```diff
@@ -55,7 +55,7 @@ endif
 
 sphinx: $(SPHINX_DEPS)
 	$(SHOW)'SPHINXBUILD doc/sphinx'
-	$(HIDE)COQBIN="$(PWD)/bin" $(SPHINXBUILD) -W -b html $(ALLSPHINXOPTS) doc/sphinx $(SPHINXBUILDDIR)/html
```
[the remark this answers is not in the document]
gares: Oh, I did not know COQBIN was used internally, now I better understand what you are trying to do.
I'll be more explicit in the review

gares (changes requested):
Put `COQBIN?=$(shell pwd)/bin` somewhere, do not export it and invoke sphinx as `COQBIN=$(COQBIN) $(SPHIX...)`

AFAIK COQBIN is not used internally in the makefile, it is just supported by coq_makefile. I guess this is why Clement knew about this variable and used it here. I dont think he should have.

The better solution would be to drop COQBIN, use `PATH` and have the makefile add to PATH `$(PWD)/bin` before running sphinx. Also, is seems the python class accepts an explicit path to coqtop, but I don'k know if there is a command line flag to sphinx that set it (in that case you don't even need to set PATH).

SkySkimmer:
New version.

Makefile.common
```diff
@@ -14,6 +14,8 @@
 # Executables
 ###########################################################################
 
+COQBIN:=$(CURDIR)/bin/
```
gares: I prefer `?=`. For example:

```
gares@gargamel:~$ cd /tmp/
gares@gargamel:/tmp$ cat > Makefile
X:=3
all:
	echo $(X)
gares@gargamel:/tmp$ make
echo 3
3
gares@gargamel:/tmp$ make X=4
echo 4
4
gares@gargamel:/tmp$ X=4 make
echo 3
3
gares@gargamel:/tmp$ cat > Makefile
X?=3
all:
	echo $(X)
gares@gargamel:/tmp$ make
echo 3
3
gares@gargamel:/tmp$ make X=4
echo 4
4
gares@gargamel:/tmp$ X=4 make
echo 4
4
gares@gargamel:/tmp$ 
```

SkySkimmer: I prefer `:=` so that I don't have to worry about unclean environment.

Makefile.common
ejgallego: I agree with Enrico, this looks fishy somehow.

SkySkimmer:
It was changed.

Gaëtan Gilbert

ejgallego (changes requested):
I understand this better now, indeed `COQBIN` should not be used in core Makefiles, as this will break for example building the documentation with Dune, which takes care to properly bring the right coqtop into scope.

I think sphinx should rely on PATH, or understand the value provided in `COQTOPBEST` which is the proper variable AFAICS.

SkySkimmer:
I don't understand what you want done. It sounds like a refactoring that's out of scope for this PR.
I just want the sphinx errors to stop blocking my builds.

ejgallego:
> I don't understand what you want done. It sounds like a refactoring that's out of scope for this PR.
> I just want the sphinx errors to stop blocking my builds.

Sorry if our two request for changes are not clear. I think that the modification we are asking for is:
- don't use `COQBIN`, it was not meant to be used by the build system and its introduction is an error;
- so drop the COQBIN part, and instead either use the already existing `COQTOPBEST` or set the path properly and let Sphinx use the `coqtop` in-scope.

SkySkimmer:
Updated

Makefile.common
```diff
@@ -10,6 +10,9 @@
 
 -include config/Makefile
 
+PATH:=$(CURDIR)/bin:$(PATH)
+export PATH
+
```
ejgallego: Sorry if I am slow here, why do we need to update the `PATH` here?

[Note that I am not saying this is a bad thing to do].

ejgallego (dismissed):
Looks better now, I still have one question tho.

ejgallego:
I didn't mean to have a "request changes" here, only a commnet, but Github did like that...

Makefile.common
SkySkimmer: We need to update it somewhere, here seems like the right place.

ejgallego: What component does need the new path? AFAICS we also set `PATH` in the Sphinx call, so this should not be needed, right?

I am fine with setting PATH globally and in fact this is how the Dune system works, but then we'd need to tweak quite a few items in the makefile. IMHO not worth it.

gares (approved):

ejgallego:
I repeat here as I don't see my question:

Why do we need to export `PATH`, what component does need it? AFAICS we also set `PATH+  in the Sphinx call, so the export should not be needed, right?

I am fine with setting `PATH` globally and in fact this is how the Dune system works, but then we'd need to tweak quite a few items in the makefile. IMHO not worth it.

ejgallego:
Oh so certainly I looked at wrong diff or my memory failed today.

But this doesn't seem broken, right? In fact, I may be happy to merge it, but I'd like to see what @gares and @letouzey think of settting up PATH in makefile. Indeed, that interacts with many components and has the potential to simplify quite a few things.

ejgallego:
> has the potential to simplify quite a few things.

On the other hand it is hard to predict what the interactions will be.

SkySkimmer:
In the log for the new job it says things like "/usr/bin/x86_64-linux-gnu-ld: cannot open output file bin/ocamllibdep: No such file or directory"
A previous CI run failed but maybe it was spurious and led to me noticing the above.

ejgallego:
> In the log for the new job it says things like "/usr/bin/x86_64-linux-gnu-ld: cannot open output file bin/ocamllibdep: No such file or directory"
> A previous CI run failed but maybe it was spurious and led to me noticing the above.

Ok let's have a look, indeed setting `PATH` just for Sphinx may be the way to go for the moment.

ejgallego:
We need to improve more but this looks good to me. Note that this PR conflicts with #7684 , I prefer the naming there as IMHO we should prefix variables.

I propose we merge this and I rebase on top and tweak the variable.

ejgallego (approved):

PR #7521, opened 2018-05-15
Fix soundness bug with VM/native and cofixpoints

### LINKED ISSUE #7333: vm_compute segfaults / Anomaly with cofix
#### Version
The Coq Proof Assistant, version 8.8.0 (April 2018)
compiled on Apr 18 2018 15:14:15 with OCaml 4.05.0

and

The Coq Proof Assistant, version 8.7.2 (March 2018)
compiled on Mar 4 2018 20:17:32 with OCaml 4.04.2

#### Operating system

Linux 4.14.29 #1-NixOS SMP x86_64 GNU/Linux

The following (minimized) file causes a segfault when executing the last line:

```coq
CoInductive wrap : Type :=
 | item : unit -> wrap.

Definition extract (t : wrap) : unit :=
match t with
| item x => x
end.

CoFixpoint close u : unit -> wrap :=
match u with
| tt => item
end.

Definition table : wrap := close tt tt.

Eval vm_compute in (extract table).
Eval vm_compute in (extract table). (* Segfault on the second vm_compute *)
```

Alternately the follow file causes an `Anomaly "Uncaught exception Not_found."`:

```coq

Set Primitive Projections.
CoInductive wrap : Type :=
 item { extract : unit }.

CoFixpoint close u : unit -> wrap := 
match u with 
| tt => item
end.

Definition table : wrap := close tt tt.

Eval vm_compute in (extract table).
Eval vm_compute in (extract table). (* Anomaly on the second vm_compute *)
```

maximedenes:
This fix includes a nice cleanup in the kernel, removing a hook and unifying `pre_env` and `env`. The cleanup makes the fix convenient because the bytecode and native compilers can then access standard reduction functions. In particular, we eta-expand cofixpoints when required, to fix the soundness bug.

I'm not sure if this qualifies for backporting, though. Let's see what CI says.

This is still work in progress, it is missing a `CHANGES` entry and some test cases.

Fixes / closes #7333

- [ ] Corresponding documentation was added / updated.
- [x] Entry added in CHANGES.


ejgallego:
CI too old.

kernel/safe_typing.ml
```diff
@@ -953,3 +953,125 @@ Would this be correct with respect to undo's and stuff ?
 let set_strategy e k l = { e with env =
    (Environ.set_oracle e.env
       (Conv_oracle.set_strategy (Environ.oracle e.env) k l)) }
+
+(** Register retroknowledge hooks *)
+
+open Retroknowledge
+
+(* the Environ.register function syncrhonizes the proactive and reactive
```
SkySkimmer: typo

kernel/environ.mli
```diff
@@ -28,16 +28,60 @@ open Declarations
...
+  env_named_ctx : Context.Named.t;
+  env_named_map : (Context.Named.Declaration.t * lazy_val) Id.Map.t;
+}
+
+type rel_context_val = private {
+  env_rel_ctx : Context.Rel.t;
+  env_rel_map : (Context.Rel.Declaration.t * lazy_val) Range.t;
+}
+
+type env = {
```
ppedrot: I'm not sure that making the `env` type fully transparent is a good idea. We should instead provide a view function from and into this record representation.

maximedenes: Ok, but this is orthogonal, since `pre_env` is already transparent today.

ppedrot: Yep, but nobody uses `pre_env` whereas `env` is a pervasive, opaque, datastructure.

maximedenes: Ok, but isn't making it private enough? What do you want to ensure with the view?

ppedrot: Marking it private is a good start, but in general there is no reason to export the implementation of a type if nobody relies on it. That makes changing the representation at a later time harder, and the fact we already changed a few implementation details of environments makes me wary that it ends up having unexpected costs.

Also, it might be enough to check for the direct uses of the fields and provide the corresponding functions in the API.

maximedenes: Ok, I'll make it private and re-export the projections, although I'm not convinced it will make the code more robust.

maximedenes:
One thing that was a bit painful is that there were two functions named the same way in `Pre_env` and `Environ`, namely `lookup_named_val`, but not doing the same thing (nor having the same type). One was taking the environment as input, the other a context. One was returning a declaration, the other a VM value.

kernel/reduction.ml
```diff
@@ -789,24 +789,6 @@ let infer_conv_leq ?(l2r=false) ?(evars=fun _ -> None) ?(ts=full_transparent_sta
     env univs t1 t2 = 
   infer_conv_universes CUMUL l2r evars ts env univs t1 t2
 
-(* This reference avoids always having to link C code with the kernel *)
```
ppedrot: Is there a justification for this comment? What do we gain by not always linking C code? Can that affect e.g. JsCoq (cc @ejgallego).

ppedrot (commented):
I'm just thinking about that and it might be a stupid idea, but it may be interesting to write a post-mortem for every soundness bug of the kernel, like people do in security. This would be nice both for documentation for the future, and also because it'd be more transparent to external users.

kernel/reduction.mli
```diff
@@ -118,9 +118,12 @@ val betazeta_appvect : int -> constr -> constr array -> constr
 
 val dest_prod       : env -> types -> Context.Rel.t * types
 val dest_prod_assum : env -> types -> Context.Rel.t * types
+val dest_lam        : env -> types -> Context.Rel.t * types
```
ppedrot: Just realizing that, but the implicit documentation that the argument and result are types is wrong. This function operates on terms, not types, and hence should be using `constr`. Same for the function below that was there before the patch.

kernel/reduction.ml
ejgallego: It is hard to infer the historical context as this has been there for a long time and the toolchain has changed quite a bit. In jscoq I do provide JS stubs for the C functions so I think that shouldn't matter there.

[In fact if someone would like to implement the stubs as either an Emscripten version of the C VM or as a native WebASM target vm_compute would happily work, see https://github.com/ejgallego/jscoq/issues/10]

herbelin:
> I'm just thinking about that and it might be a stupid idea, but it may be interesting to write a post-mortem for every soundness bug of the kernel, like people do in security. 

Yes, I think it is important that we do that!

maximedenes:
> I'm just thinking about that and it might be a stupid idea, but it may be interesting to write a post-mortem for every soundness bug of the kernel, like people do in security. This would be nice both for documentation for the future, and also because it'd be more transparent to external users.

Indeed, I've thought about this before, it would be interesting to explore the possibilities. Can you open an issue so that we keep it in mind?

maximedenes:
I will add the `CHANGES` entry in an 8.8-specific PR fixing bug without the refactoring part, using a hook instead. @ppedrot I consider this ready (modulo CI testing).

ppedrot:
As far as I'm concerned, I don't even understand in detail the invariant that was broken by not η-expanding cofixpoints in the VM. I trust your patch, but I'd be happy to know more about it and see this documented somewhere...

Zimmi48:
Sorry @maximedenes, let me correct myself. You need to add the CHANGES entry in both PRs since each one is going to be merged in a different branch. But in the two PRs, it must be the same message in the same section (Changes between 8.8.0 and 8.8.1)

maximedenes:
@ppedrot I added a bit of description of the invariant. Basically the problem is that the memoization code is inserted after the last lambda, and doesn't expect to be applied to more arguments.

ppedrot:
Let's wait for the CI to terminate and merge.

ppedrot:
It looks like Ltac2 needs an overlay.

Zimmi48:
How come this PR did not add any tests? Sorry for not noticing earlier.

ppedrot:
Indeed, it would be better to add a test. @maximedenes Can you open a new PR?

PR #6454, opened 2017-12-17
[econstr] Flag to make `to_constr` fail if its output contains evars

ejgallego:
We forbid calling `EConstr.to_constr` on terms that are not evar-free,
as to progress towards enforcing the invariant that `Constr.t` is
evar-free. [c.f. #6308]

Due to compatibility constraints we provide an optional parameter to
`to_constr`, `abort` which can be used to overcome this restriction
until we fix all parts of the code.

Now, grepping for `~abort:false` should return the questionable
parts of the system.

Not a lot of places had to be fixed, some comments:

- problems with the interface due to `Evd/Constr` [`Evd.define` being
  the prime example] do seem real!
- inductives also look bad with regards to `Constr/EConstr`.
- code in plugins needs work.

A notable user of this "feature" is `Obligations/Program` that seem to
like to generate kernel-level entries with free evars, then to scan
them and workaround this problem by generating constants.

PR #7524, opened 2018-05-16
[ssr] fix after to_constr ~abort_on_undefined_evars was added

gares:
Now that I understand the semantics of the flag, which I did not at the time #6454 was merged, I can fix ssr.

### DIFF
--- a/plugins/ssr/ssrequality.ml
+++ b/plugins/ssr/ssrequality.ml
@@ -287,7 +287,10 @@ let foldtac occ rdx ft gl =
     (fun env c _ h -> try find_T env c h ~k:(fun env t _ _ -> t) with NoMatch ->c),
     (fun () -> try end_T () with NoMatch -> fake_pmatcher_end ())
   | _ -> 
-    (fun env c _ h -> try let sigma = unify_HO env sigma (EConstr.of_constr c) (EConstr.of_constr t) in EConstr.to_constr sigma (EConstr.of_constr t)
+    (fun env c _ h ->
+       try
+         let sigma = unify_HO env sigma (EConstr.of_constr c) (EConstr.of_constr t) in
+         EConstr.to_constr ~abort_on_undefined_evars:false sigma (EConstr.of_constr t)
     with _ -> errorstrm Pp.(str "fold pattern " ++ pr_constr_pat t ++ spc ()
       ++ str "does not match redex " ++ pr_constr_pat c)), 
     fake_pmatcher_end in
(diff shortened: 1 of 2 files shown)


gares:
For reviewing:
- ssrequality: after unifying the pattern with a sub term of the goal we can't be sure the sub term being selected is not evar free (if the goal contains an evar, for example)
- ssrview: the local function is called `evars_of_econstr` and indeed expects the input to be an open term. I guess I did not know the function in evarutil I now call.

plugins/ssr/ssrequality.ml
```diff
@@ -287,7 +287,10 @@ let foldtac occ rdx ft gl =
     (fun env c _ h -> try find_T env c h ~k:(fun env t _ _ -> t) with NoMatch ->c),
     (fun () -> try end_T () with NoMatch -> fake_pmatcher_end ())
   | _ -> 
-    (fun env c _ h -> try let sigma = unify_HO env sigma (EConstr.of_constr c) (EConstr.of_constr t) in EConstr.to_constr sigma (EConstr.of_constr t)
+    (fun env c _ h ->
+       try
+         let sigma = unify_HO env sigma (EConstr.of_constr c) (EConstr.of_constr t) in
+         EConstr.to_constr ~abort_on_undefined_evars:false sigma (EConstr.of_constr t)
```
ppedrot: This is highly suspicious code. What are you intending here?

gares: I do agree, so I'll try to explain.

Here `(sigma,t)` is an open constr and `(sigma0,c)` is a subterm of the goal, that an open constr. `sigma` extends `sigma0`.
We replace `c` with roughly `nf_evars sigma1 t` where `sigma1` contains the unifier of `t` with `c`.
- If the unifier assigns evars in `c`, then the code is incorrect (since we later work in `sigma0`).
- If the unifier assigns evars in `t` then the code is now correct (up to universe constraints, should be later inferred by `convert_concl` I think).
- If the unifier assigns evars in `t` to terms containing evars (i.e. c is not evar free) then the `~abortion` is triggered for no good reason (since these evars are in `sigma0`).

So, this code from another geologic era (pre-universes) works fine (even today I believe) if the goal contains no evar.
If the goal contains evars and you are lucky it used to work before `~abort` and will work after this PR (it does not work
now, and emilio did not spot it because 99% of the tests on ssreflect work on evar free goals).

I hope I made the misery clear, and that this PR is not making things worse. Does it make sense to you?

SkySkimmer: Why not use Evarutil.nf_evars_universes?

gares: Thanks for the hint, I'll look at that.

gares: I meant, that I may use this API in another PR to improve this code.
The current PR is just a fix.

maximedenes (approved):
This is an OK temporary fix, but the code should be cleaned-up at some point.

PR #7525, opened 2018-05-16
[ci] Try to build more of fiat-crypto.

ejgallego:
See #7514


JasonGross (approved):
Seems to have succeeded this time

SkySkimmer:
I got an OOM error at https://gitlab.com/SkySkimmer/coq/-/jobs/68999982

JasonGross:
I think either the ci instances should be given more RAM, or you should build fiat-crypto with -j1.

Zimmi48:
Many similar failures:
- https://gitlab.com/coq/coq/-/jobs/69329709
- https://gitlab.com/coq/coq/-/jobs/69301077
- https://gitlab.com/coq/coq/-/jobs/68945389
- https://gitlab.com/coq/coq/-/jobs/68938186
- https://gitlab.com/coq/coq/-/jobs/68918461

We should either revert this PR or find a way to appropriately tag the fiat-crypto job so that only the runners which are powerful enough to handle it are used...

JasonGross:
Can't we just force fiat-crypto to build with `-j1`?

Zimmi48:
I may be wrong but it seems to me that this is going to often make the fiat-crypto build longer than 2h. What about first building the targets that can be handled with `-j2` then the targets than require more memory with `-j1`? (If it is well understood which are which...)

ejgallego:
We could just try to detect how many cores the worker has, on the other that's not so easy to do. I suggest we just tune the build down a bit.

PR #7526, opened 2018-05-16
[circle] Use Docker image from Gitlab registry.

ejgallego:


Zimmi48:
Maybe you can update the README and remove the push hook at the same time?

ejgallego:
> Maybe you can update the README and remove the push hook at the same time?

Indeed it does make sense.

.circleci/config.yml
```diff
@@ -46,7 +47,9 @@ before_script: &before_script
         paths:
           - coq/
 
-  environment: *envvars
+  environment:
+    <<: *envvars
+    NATIVE_COMP: "yes"
```
SkySkimmer: Why this change?

ejgallego: Otherwise the image vars are not picked up in subjobs.

dev/ci/docker/README.md
```diff
@@ -1,37 +1,13 @@
 ## Overall Docker Setup for Coq's CI.
```
SkySkimmer: We should say somewhere "if you change the dockerfile you need to change the cachekey in gitlab and circle yml files"

Zimmi48 (commented):
You could also remove the three lines about `coqci` since it seems we're not going to use this organization after all.

dev/ci/docker/README.md
ejgallego: I think now things are simple enough that even a commit check could work to ensure this.

ejgallego:
Sure, will do another pass.

dev/ci/docker/README.md
Zimmi48: That would be great!

ejgallego:
Ok, changes done. I don't know what the best way to ensure automatically these 3:
```
$ git grep 'CACHEKEY:'
.circleci/config.yml:    CACHEKEY: "bionic_coq-V2018-05-07-V2"
.gitlab-ci.yml:  CACHEKEY: "bionic_coq-V2018-05-07-V2"
dev/ci/docker/bionic_coq/Dockerfile:# CACHEKEY: "bionic_coq-V2018-05-07-V2"
```
are kept in sync. I won't have the time to work on it soon so feel free to tackle on the pre-commit check.

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -1,3 +1,6 @@
+# CACHEKEY: "bionic_coq-V2018-05-07-V2"
```
Zimmi48: What is the point of repeating the CACHEKEY here? Unless this isn't actually a real comment...

ejgallego: The point is to allow a script to check that this version here is in sync with what's inside `.gitlab-ci.yml`

SkySkimmer: For automatic checking we should be able to check that any commit modifying the dockerfile also modifies the cachekey in gitlab.yml.

ejgallego: That's also a valid way to do it I guess. Also cool would be to autogenerate `.gitlab-ci.yml` and `Dockerfile` using a promote rule.

SkySkimmer:
I have no clue what that is.

Gaëtan Gilbert

ejgallego:
> I have no clue what that is.

I have no clude what "what" is but I infer you talk about promote rules.

A promote make rule is a build rule that allows to check consistent version of auto-generated files in the repository, kind of releasing a file. Thus, for example here, `.gitlab-ci.yml` could depend on the hash of the Dockerfile, so when you want to update it you just do "promote".

The build rule can also warn you when the file is out of date [as it will generate a temp version and then diff].

SkySkimmer (approved):

PR #7522, opened 2018-05-16
[ocaml] Update required OCaml version to 4.05.0

ejgallego:
Closes #7380. Ubuntu 18.04 and Debian Buster will ship this OCaml
version so it makes sense we bump our dependency to 4.05.0 as we can
use some newer compiler features.

PR #7527, opened 2018-05-16
[windows] Don't build menhir and int anymore in the packaging scripts.

ejgallego:
As pointed out by @MSoegtropIMC
[here](https://github.com/coq/coq/pull/7522#issuecomment-389478963)
there are not needed to build the packages, so not building them will
save a couple of minutes.

### DIFF
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -847,7 +847,7 @@ function make_ocaml {
 
 function make_ocaml_tools {
   make_findlib
-  make_menhir
+  # make_menhir
   make_camlp5
 }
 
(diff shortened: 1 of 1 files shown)


MSoegtropIMC (approved):
Changes are fine.

ejgallego:
@MSoegtropIMC as I understand, you are the one supposed to merge this change.

MSoegtropIMC:
@ejgallego : btw.: another not so necessary build time consumer is sed. I build it from sources since cygwin made a fix to not strip \r characters. Since the OCaml tools and Coq are MinGW and not Cygwin, they produce \r on output and this leads to errors. But there shouldn't be that many places where this hurts and adding an extra \r strip using tr shouldn't hurt, but it might be that various libs need minor patched.

ejgallego:
Thanks @MSoegtropIMC , I was indeed confused as you hadn't assigned the PR to yourself.

Thanks for the `sed` advice, indeed I had read some problems you folks were discussing about it, in my Dune tree the only place `sed` is used tho is `make_opcodes` in the kernel, I am not sure if we could make it work in a compatible way, but for sure it is worth trying.

Issue #6938, opened 2018-03-07
A blueprint for the "Coq 9" project.


ejgallego:
@mattam82 and others have hinted a few times the possibility of a "Coq 9" event happening some time soon. Indeed, as the version bump in "Coq 9" indicates, it would be a version introducing non-trivial incompatibilities in the system.

With 8.8 almost frozen, it could be a good moment to start to think what the plan could be. IMHO there are two main topics to address, **what** and **how**:

### Candidates to be included in Coq 9:

- Removal of legacy unification (#930, #991)
- Improvements to the notation and printing system (#6247 , #6455)
- Changes in the stdlib (#306 #140)
- New build system (#6857)
- STM 2.0, removal of the XML protocol.
- Vernac / state refactoring.
- Uniform scoping of objects.
- Module system?

### How would a plan look like?

The main problem with such "disruptive" developments is that they are in conflict with a regular release schedule. How can we reconcile the two?

Indeed, it seems that the only way to achieve such a large-scale project is to enable both Coq 8.x and the future Coq 9 to coexist. There are several approaches to achieve this, but the most promising one is to have a in-tree fork of the codebase, and allow the users to enable Coq 9 features. This supposes a bit of maintenance cost at first, but with a bit of care it can be done all rigth, and at Coq 9 time compatibility structures can be removed.

Another possibility is to generate two different binaries, `coqtop` and `coqtop9`, so the difference between the systems are more marked, and we don't have to mess so much with aliases and modules.

ejgallego:
So folks, what compat-breaking changes you would like to see in Coq 9?

maximedenes:
Nice idea! What about a grammar for Coq documents? Along the lines of https://github.com/gares/ceps/blob/3ccc8e86f823dbff1f011ae7443acb11c968e85d/text/v-file-format.md

herbelin:
> what compat-breaking changes you would like to see in Coq 9?

In complemente to what has already been said:

- It would be a significant step to have an equality defined by abstraction on a direction, cubical style, together with quotients and HITs, the former being required, as far as I can see, to have the second computing well. I'm sure @mattam82 has that in mind too.
[...]

Issue #6768, opened 2018-02-15
Towards a fix to the loss of subject reduction with corecursion?


herbelin:
I wonder whether we could solve the failure of subject reduction with corecursion by using an asymmetric conversion algorithm. Let me remind the failure of subject reduction with (so-called) co-fixpoints:
```coq
CoInductive I := C : I -> I.
CoFixpoint infty := C infty.
Definition unfold : infty = C infty :=
  match infty as x return match x with C n => x = C n end with
  | C n => eq_refl (C n)
  end.
Fail Definition nf_unfold : infty = C infty := Eval lazy in unfold.
```
It fails because we have to compare `C infty = C infty ≤ infty = C infty`, i.e. `C infty ≤ infty` and reduction does not allow to unfold `infty` since noone is asking for destructing it.

But, to some extend, someone is asking for destructing it: it is the other side of the conversion problem. So, the same way as we implemented η where the presence of a constructor on one side makes legitimate to η-expand on the other side, can't we accept that a constructor on one side makes legitimate to cofix-expand on the other side.

The problem is, that, on the contrary of η-expansion for a connective, there is no finite type structure to ensure termination of the process, in the sense that η-expanding on one side might lead to require to η-expand on the other side, and this infinitely many often.

So, the question is: could we solve the problem by allowing cofix-expansion only on one side? After all, subject reduction generates oriented conversion problems (i.e. "type of subject" ≤ "object"), so we may be able to preserve this information of orientation in the conversion algorithm.

What do you think?

ppedrot:
My take on that is that coinductive types should not enjoy dependent elimination, because in most models you can actually show that this does not hold. There is a very good reason for that, which is that cofixpoints have no canonical forms, even if you try performing η-expansions. For instance, how do you expand a cofixpoint that's doing two steps at once.

So, I think we should simply forbid coinductive types defined through constructors, and force people to use primitive records instead, as these elegantly solve the problem.

I was advocating for a warning on positively defined co-inductive types in 8.8, but I don't know if the discussion converged.

maximedenes:
> I was advocating for a warning on positively defined co-inductive types in 8.8, but I don't know if the discussion converged.

Maybe a good topic for a WG discussion?

herbelin:
@ppedrot: I know this view of you, but the current situation is that Coq's coinductive are actually greatest fixpoints of positive types (i.e. if I understand category theory language well enough, infinite limits of finite collmits) and not greatest fixpoints of negative types. 

[...]

Issue #5303, opened 2017-01-10
Printing of and error messages involving Records, Structures and Variants + doc of Set Printing All


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#5303
From: @amahboubi
Reported version: 8.5
CC:   @ejgallego, @herbelin

amahboubi:
Sorry for the late feedback @herbelin : for me this is issue can be closed, thanks for the fix!

PR #7536, opened 2018-05-16
Deprecate positive coinductive types

ppedrot:
Adding a warning when declaring positive coinductive types.

Such types a fundamentally wrong and are known to break subject reduction of CIC. We advertise for the use of negative coinductive types instead, i.e. defined by means of primitive records.

Due to the fact that the kernel is currently unable to handle mutually corecursive negative records, we currently only produce the warning on coinductive types which only made of one block, which is the vast majority of the uses. When this is solved, the warning needs to be adapted accordingly.

We adapted the standard library and tests to use primitive corecursive records.

Most of the changes are actually due to the Coq "design pattern" that CoInductive types do not generate induction schemes.

The only part of the stdlib that was really using them is the Stream library, which is probably not much used in the wild.

Depending on the case, tests were either adapted, or removed if they were made moot by the transition, e.g. if they relied on properties not holding with negative coinductive types.

This was documented both in the Coq reference manual and in the CHANGES.

### DIFF
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -1173,7 +1173,8 @@ equality:
    Axiom Stream_ext : forall (s1 s2: Stream), EqSt s1 s2 -> s1 = s2.
 
 As of Coq 8.9, it is now advised to use negative co-inductive types rather than
-their positive counterparts.
+their positive counterparts. As such, the latter are deprecated and their
+support is expected to be discontinued in a future version of Coq.
 
 .. seealso::
    :ref:`primitive_projections` for more information about negative
--- a/test-suite/failure/cofixpoint.v
+++ b/test-suite/failure/cofixpoint.v
@@ -1,9 +1,10 @@
 (* A bug in the guard checking of nested cofixpoints. *)
 (* Posted by Maxime Dénès on coqdev (Apr 9, 2014).    *)
 
-CoInductive CoFalse := .
+Set Primitive Projections.
+CoInductive CoFalse := { coFalse : False }.
 
-CoInductive CoTrue := I.
+CoInductive CoTrue := { coTrue : True }.
 
 Fail CoFixpoint loop : CoFalse :=
   (cofix f := loop with g := loop for f).
--- a/theories/Lists/Streams.v
+++ b/theories/Lists/Streams.v
@@ -9,17 +9,26 @@
 (************************************************************************)
 
 Set Implicit Arguments.
-Set Primitive Projections.
+
+Set Warnings "-deprecated-positive-coinductive".
 
 (** Streams *)
 
 CoInductive Stream (A : Type) :=
-  Cons { hd : A; tl : Stream A }.
+  Cons : A -> Stream A -> Stream A.
 
 Section Streams.
   Variable A : Type.
 
   Notation Stream := (Stream A).
+
+Definition hd (x:Stream) := match x with
+                            | Cons a _ => a
+                            end.
+
+Definition tl (x:Stream) := match x with
+                            | Cons _ s => s
+                            end.
 
 Fixpoint Str_nth_tl (n:nat) (s:Stream) : Stream :=
   match n with
@@ -28,6 +37,17 @@ Fixpoint Str_nth_tl (n:nat) (s:Stream) : Stream :=
   end.
 
 Definition Str_nth (n:nat) (s:Stream) : A := hd (Str_nth_tl n s).
+
+
+Lemma unfold_Stream :
+ forall x:Stream, x = match x with
+                      | Cons a s => Cons a s
+                      end.
+Proof.
+  intro x.
+  case x.
+  trivial.
+Qed.
 
 Lemma tl_nth_tl :
  forall (n:nat) (s:Stream), tl (Str_nth_tl n s) = Str_nth_tl n (tl s).
@@ -39,7 +59,6 @@ Hint Resolve tl_nth_tl: datatypes.
 Lemma Str_nth_tl_plus :
  forall (n m:nat) (s:Stream),
    Str_nth_tl n (Str_nth_tl m s) = Str_nth_tl (n + m) s.
-Proof.
 simple induction n; simpl; intros; auto with datatypes.
 rewrite <- H.
 rewrite tl_nth_tl; trivial with datatypes.
@@ -47,17 +66,15 @@ Qed.
 
 Lemma Str_nth_plus :
  forall (n m:nat) (s:Stream), Str_nth n (Str_nth_tl m s) = Str_nth (n + m) s.
-Proof.
 intros; unfold Str_nth; rewrite Str_nth_tl_plus;
  trivial with datatypes.
 Qed.
 
 (** Extensional Equality between two streams  *)
 
-CoInductive EqSt (s1 s2: Stream) : Prop := eqst {
-  eqst_hd : hd s1 = hd s2;
-  eqst_tl : EqSt (tl s1) (tl s2);
-}.
+CoInductive EqSt (s1 s2: Stream) : Prop :=
+    eqst :
+        hd s1 = hd s2 -> EqSt (tl s1) (tl s2) -> EqSt s1 s2.
 
 (** A coinduction principle *)
 
@@ -69,13 +86,11 @@ Ltac coinduction proof :=
 (** Extensional equality is an equivalence relation *)
 
 Theorem EqSt_reflex : forall s:Stream, EqSt s s.
-Proof.
 coinduction EqSt_reflex.
 reflexivity.
 Qed.
 
 Theorem sym_EqSt : forall s1 s2:Stream, EqSt s1 s2 -> EqSt s2 s1.
-Proof.
 coinduction Eq_sym.
 + case H; intros; symmetry ; assumption.
 + case H; intros; assumption.
@@ -84,7 +99,6 @@ Qed.
 
 Theorem trans_EqSt :
  forall s1 s2 s3:Stream, EqSt s1 s2 -> EqSt s2 s3 -> EqSt s1 s3.
-Proof.
 coinduction Eq_trans.
 - transitivity (hd s2).
   + case H; intros; assumption.
@@ -99,7 +113,6 @@ Qed.
 
 Theorem eqst_ntheq :
  forall (n:nat) (s1 s2:Stream), EqSt s1 s2 -> Str_nth n s1 = Str_nth n s2.
-Proof.
 unfold Str_nth; simple induction n.
 - intros s1 s2 H; case H; trivial with datatypes.
 - intros m hypind.
@@ -112,7 +125,6 @@ Qed.
 Theorem ntheq_eqst :
  forall s1 s2:Stream,
    (forall n:nat, Str_nth n s1 = Str_nth n s2) -> EqSt s1 s2.
-Proof.
 coinduction Equiv2.
 - apply (H 0).
 - intros n; apply (H (S n)).
@@ -133,7 +145,7 @@ Inductive Exists ( x: Stream ) : Prop :=
   | Further : Exists (tl x) -> Exists x.
 
 CoInductive ForAll (x: Stream) : Prop :=
-    HereAndFurther { Forall_here : P x; Forall_further : ForAll (tl x) }.
+    HereAndFurther : P x -> ForAll (tl x) -> ForAll x.
 
 Lemma ForAll_Str_nth_tl : forall m x, ForAll x -> ForAll (Str_nth_tl m x).
 Proof.
@@ -151,7 +163,6 @@ Hypothesis InvThenP : forall x:Stream, Inv x -> P x.
 Hypothesis InvIsStable : forall x:Stream, Inv x -> Inv (tl x).
 
 Theorem ForAll_coind : forall x:Stream, Inv x -> ForAll x.
-Proof.
 coinduction ForAll_coind; auto.
 Qed.
 End Co_Induction_ForAll.
@@ -171,7 +182,6 @@ induction n.
 - reflexivity.
 - simpl.
   intros s.
-  cbn.
   apply IHn.
 Qed.
 
@@ -219,8 +229,7 @@ Lemma Str_nth_tl_zipWith : forall n (a:Stream A) (b:Stream B),
 Proof.
 induction n.
 - reflexivity.
-- cbn.
-  intros ? ?.
+- intros [x xs] [y ys].
   unfold Str_nth in *.
   simpl in *.
   apply IHn.
--- a/theories/Lists/Streams.v
+++ b/theories/Lists/Streams.v
@@ -9,26 +9,17 @@
 (************************************************************************)
 
 Set Implicit Arguments.
+Set Primitive Projections.
 
 (** Streams *)
 
 CoInductive Stream (A : Type) :=
-  Cons : A -> Stream A -> Stream A.
+  Cons { hd : A; tl : Stream A }.
 
 Section Streams.
   Variable A : Type.
 
   Notation Stream := (Stream A).
-
-
-Definition hd (x:Stream) := match x with
-                            | Cons a _ => a
-                            end.
-
-Definition tl (x:Stream) := match x with
-                            | Cons _ s => s
-                            end.
-
 
 Fixpoint Str_nth_tl (n:nat) (s:Stream) : Stream :=
   match n with
@@ -37,17 +28,6 @@ Fixpoint Str_nth_tl (n:nat) (s:Stream) : Stream :=
   end.
 
 Definition Str_nth (n:nat) (s:Stream) : A := hd (Str_nth_tl n s).
-
-
-Lemma unfold_Stream :
- forall x:Stream, x = match x with
-                      | Cons a s => Cons a s
-                      end.
-Proof.
-  intro x.
-  case x.
-  trivial.
-Qed.
 
 Lemma tl_nth_tl :
  forall (n:nat) (s:Stream), tl (Str_nth_tl n s) = Str_nth_tl n (tl s).
@@ -59,6 +39,7 @@ Hint Resolve tl_nth_tl: datatypes.
 Lemma Str_nth_tl_plus :
  forall (n m:nat) (s:Stream),
    Str_nth_tl n (Str_nth_tl m s) = Str_nth_tl (n + m) s.
+Proof.
 simple induction n; simpl; intros; auto with datatypes.
 rewrite <- H.
 rewrite tl_nth_tl; trivial with datatypes.
@@ -66,15 +47,17 @@ Qed.
 
 Lemma Str_nth_plus :
  forall (n m:nat) (s:Stream), Str_nth n (Str_nth_tl m s) = Str_nth (n + m) s.
+Proof.
 intros; unfold Str_nth; rewrite Str_nth_tl_plus;
  trivial with datatypes.
 Qed.
 
 (** Extensional Equality between two streams  *)
 
-CoInductive EqSt (s1 s2: Stream) : Prop :=
-    eqst :
-        hd s1 = hd s2 -> EqSt (tl s1) (tl s2) -> EqSt s1 s2.
+CoInductive EqSt (s1 s2: Stream) : Prop := eqst {
+  eqst_hd : hd s1 = hd s2;
+  eqst_tl : EqSt (tl s1) (tl s2);
+}.
 
 (** A coinduction principle *)
 
@@ -86,11 +69,13 @@ Ltac coinduction proof :=
 (** Extensional equality is an equivalence relation *)
 
 Theorem EqSt_reflex : forall s:Stream, EqSt s s.
+Proof.
 coinduction EqSt_reflex.
 reflexivity.
 Qed.
 
 Theorem sym_EqSt : forall s1 s2:Stream, EqSt s1 s2 -> EqSt s2 s1.
+Proof.
 coinduction Eq_sym.
 + case H; intros; symmetry ; assumption.
 + case H; intros; assumption.
@@ -99,6 +84,7 @@ Qed.
 
 Theorem trans_EqSt :
  forall s1 s2 s3:Stream, EqSt s1 s2 -> EqSt s2 s3 -> EqSt s1 s3.
+Proof.
 coinduction Eq_trans.
 - transitivity (hd s2).
   + case H; intros; assumption.
@@ -113,6 +99,7 @@ Qed.
 
 Theorem eqst_ntheq :
  forall (n:nat) (s1 s2:Stream), EqSt s1 s2 -> Str_nth n s1 = Str_nth n s2.
+Proof.
 unfold Str_nth; simple induction n.
 - intros s1 s2 H; case H; trivial with datatypes.
 - intros m hypind.
@@ -125,6 +112,7 @@ Qed.
 Theorem ntheq_eqst :
  forall s1 s2:Stream,
    (forall n:nat, Str_nth n s1 = Str_nth n s2) -> EqSt s1 s2.
+Proof.
 coinduction Equiv2.
 - apply (H 0).
 - intros n; apply (H (S n)).
@@ -145,7 +133,7 @@ Inductive Exists ( x: Stream ) : Prop :=
   | Further : Exists (tl x) -> Exists x.
 
 CoInductive ForAll (x: Stream) : Prop :=
-    HereAndFurther : P x -> ForAll (tl x) -> ForAll x.
+    HereAndFurther { Forall_here : P x; Forall_further : ForAll (tl x) }.
 
 Lemma ForAll_Str_nth_tl : forall m x, ForAll x -> ForAll (Str_nth_tl m x).
 Proof.
@@ -163,6 +151,7 @@ Hypothesis InvThenP : forall x:Stream, Inv x -> P x.
 Hypothesis InvIsStable : forall x:Stream, Inv x -> Inv (tl x).
 
 Theorem ForAll_coind : forall x:Stream, Inv x -> ForAll x.
+Proof.
 coinduction ForAll_coind; auto.
 Qed.
 End Co_Induction_ForAll.
@@ -182,6 +171,7 @@ induction n.
 - reflexivity.
 - simpl.
   intros s.
+  cbn.
   apply IHn.
 Qed.
 
@@ -229,7 +219,8 @@ Lemma Str_nth_tl_zipWith : forall n (a:Stream A) (b:Stream B),
 Proof.
 induction n.
 - reflexivity.
-- intros [x xs] [y ys].
+- cbn.
+  intros ? ?.
   unfold Str_nth in *.
   simpl in *.
   apply IHn.
--- a/doc/sphinx/changes.rst
+++ b/doc/sphinx/changes.rst
@@ -806,6 +806,11 @@ Details of changes in 8.9+beta1
 Kernel
 
 - Mutually defined records are now supported.
+- Positive coinductive types, i.e. the ones defined by a list of constructors,
+  are now deprecated. They were a source of loss of subject reduction in CIC.
+  It is now advised to use their negative counterparts instead, that is, by
+  setting the "Primitive Projections" flag and defining them as records with
+  projections.
 
 Notations
 
--- a/kernel/safe_typing.ml
+++ b/kernel/safe_typing.ml
@@ -859,10 +859,23 @@ let check_mind mie lab =
     (* The label and the first inductive type name should match *)
     assert (Id.equal (Label.to_id lab) oie.mind_entry_typename)
 
+let warn_positive_coinductive =
+  let open Pp in
+  CWarnings.create ~name:"deprecated-positive-coinductive" ~category:"deprecated"
+    (fun lbl -> str "Coinductive type " ++ Label.print lbl ++ str " is defined \
+      as an algebraic type, which breaks subject reduction of CIC. Please use \
+      a primitive coinductive record instead. Algebraic coinductive types are \
+      deprecated and are to be removed from Coq.")
+
+let check_positive_coinductive lbl mib = match mib.mind_record, mib.mind_finite with
+| (NotRecord | FakeRecord), CoFinite -> warn_positive_coinductive lbl
+| _ -> ()
+
 let add_mind l mie senv =
   let () = check_mind mie l in
   let kn = MutInd.make2 senv.modpath l in
   let mib = Indtypes.check_inductive senv.env kn mie in
+  let () = check_positive_coinductive l mib in
   let mib =
     match mib.mind_hyps with [] -> Declareops.hcons_mind mib | _ -> mib
   in
--- a/test-suite/bugs/closed/bug_2319.v
+++ b/test-suite/bugs/closed/bug_2319.v
@@ -1,6 +1,7 @@
 Section S.
 
-  CoInductive A (X: Type) := mkA: A X -> A X.
+  Set Primitive Projections.
+  CoInductive A (X: Type) := mkA { a : A X }.
   Variable T : Type.
 
   (* This used to loop (bug #2319) *)
--- a/test-suite/bugs/closed/bug_4375.v
+++ b/test-suite/bugs/closed/bug_4375.v
@@ -89,8 +89,8 @@ Print a.
 Print b.
 *)
 
-Polymorphic CoInductive foo@{i} (T : Type@{i}) : Type@{i} :=
-| A : foo T -> foo T.
+Set Primitive Projections.
+Polymorphic CoInductive foo@{i} (T : Type@{i}) : Type@{i} := A { a : foo T }.
 
 Polymorphic CoFixpoint cg@{i} (t : Type@{i}) : foo@{i} t :=
   @A@{i} t (cg t).
--- a/test-suite/bugs/closed/bug_5215.v
+++ b/test-suite/bugs/closed/bug_5215.v
@@ -256,10 +256,7 @@ Definition S_nat_func : Functor Type_Cat Type_Cat :=
 
 Definition S_nat_alg_cat := Algebra_Cat S_nat_func.
 
-CoInductive CoNat : Set :=
-  | CoO : CoNat
-  | CoS : CoNat -> CoNat
-.
+CoInductive CoNat : Set := { coNat : option CoNat }.
 
 Definition S_nat_coalg_cat := @CoAlgebra_Cat Type_Cat S_nat_func.
 
--- a/test-suite/failure/guard_cofix.v
+++ b/test-suite/failure/guard_cofix.v
@@ -5,9 +5,10 @@ subterm3.v). Posted on Coq-club by Maxime Dénès (02/26/2014). *)
 
 (* First example *)
 
-CoInductive CoFalse : Prop := CF : CoFalse -> False -> CoFalse.
+Set Primitive Projections.
+CoInductive CoFalse : Prop := CF { cL : CoFalse; cR : False }.
 
-CoInductive Pandora : Prop := C : CoFalse -> Pandora.
+CoInductive Pandora : Prop := C { c : CoFalse }.
 
 Axiom prop_ext : forall P Q : Prop, (P<->Q) -> P = Q.
 
@@ -31,7 +32,7 @@ Lemma H : omega = CoFalse.
 Proof.
 apply prop_ext; constructor.
   induction 1; assumption.
-destruct 1; destruct H0.
+destruct 1; destruct cR0.
 Qed.
 
 Fail CoFixpoint loop' : CoFalse :=
--- a/test-suite/modules/errors.v
+++ b/test-suite/modules/errors.v
@@ -34,10 +34,10 @@ Module MA6 : SA6. Inductive TA6 (A B:Type):= CA6 : A -> TA6 A B. Fail End MA6.
 Reset Initial.
 
 Module Type SA7. Inductive TA7 (A:Type) := CA7 : A -> TA7 A. End SA7.
-Module MA7 : SA7. CoInductive TA7 (A:Type):= CA7 : A -> TA7 A. Fail End MA7.
+Module MA7 : SA7. Set Primitive Projections. CoInductive TA7 (A:Type):= CA7 { ca7 : A -> TA7 A }. Unset Primitive Projections. Fail End MA7.
 Reset Initial.
 
-Module Type SA8. CoInductive TA8 (A:Type) := CA8 : A -> TA8 A. End SA8.
+Module Type SA8. Set Primitive Projections. CoInductive TA8 (A:Type) := CA8 { ca8 : A -> TA8 A }. Unset Primitive Projections. End SA8.
 Module MA8 : SA8. Inductive TA8 (A:Type):= CA8 : A -> TA8 A. Fail End MA8.
 Reset Initial.
 
--- a/test-suite/output/Fixpoint.v
+++ b/test-suite/output/Fixpoint.v
@@ -44,7 +44,9 @@ fix even_pos_odd_pos 2 with (odd_pos_even_pos n (H:odd n) {struct H} : n >= 1).
   omega.
 Qed.
 
+Set Primitive Projections.
 CoInductive Inf := S { projS : Inf }.
+Unset Primitive Projections.
 Definition expand_Inf (x : Inf) := S (projS x).
 CoFixpoint inf := S inf.
 Eval compute in inf.
--- a/test-suite/prerequisite/ssr_mini_mathcomp.v
+++ b/test-suite/prerequisite/ssr_mini_mathcomp.v
@@ -196,7 +196,7 @@ Definition clone_subType U v :=
 
 Variable sT : subType.
 
-CoInductive Sub_spec : sT -> Type := SubSpec x Px : Sub_spec (Sub x Px).
+Variant Sub_spec : sT -> Type := SubSpec x Px : Sub_spec (Sub x Px).
 
 Lemma SubP u : Sub_spec u.
 Proof. by case: sT Sub_spec SubSpec u => T' _ C rec /= _. Qed.
@@ -209,7 +209,7 @@ Definition insub x :=
 
 Definition insubd u0 x := odflt u0 (insub x).
 
-CoInductive insub_spec x : option sT -> Type :=
+Variant insub_spec x : option sT -> Type :=
   | InsubSome u of P x & val u = x : insub_spec x (Some u)
   | InsubNone   of ~~ P x          : insub_spec x None.
 
@@ -568,7 +568,7 @@ Fixpoint nth s n {struct n} :=
 
 Fixpoint rcons s z := if s is x :: s' then x :: rcons s' z else [:: z].
 
-CoInductive last_spec : seq T -> Type :=
+Variant last_spec : seq T -> Type :=
   | LastNil        : last_spec [::]
   | LastRcons s x  : last_spec (rcons s x).
 
@@ -1292,7 +1292,7 @@ Open Scope big_scope.
 (* packages both in in a term in which i occurs; it also depends on the       *)
 (* iterated <op>, as this can give more information on the expected type of   *)
 (* the <general_term>, thus allowing for the insertion of coercions.          *)
-CoInductive bigbody R I := BigBody of I & (R -> R -> R) & bool & R.
+Variant bigbody R I := BigBody of I & (R -> R -> R) & bool & R.
 
 Definition applybig {R I} (body : bigbody R I) x :=
   let: BigBody _ op b v := body in if b then op v x else x.
--- a/test-suite/success/Cases.v
+++ b/test-suite/success/Cases.v
@@ -1211,11 +1211,14 @@ Type
   end.
 
 
-
-CoInductive SStream (A : Set) : (nat -> A -> Prop) -> Type :=
-    scons :
-      forall (P : nat -> A -> Prop) (a : A),
-      P 0 a -> SStream A (fun n : nat => P (S n)) -> SStream A P.
+Set Primitive Projections.
+CoInductive SStream (A : Set) (P : nat -> A -> Prop) : Type :=
+    scons {
+      SStream_a : A;
+      SStream_hd : P 0 SStream_a;
+      SStream_tl : SStream A (fun n : nat => P (S n));
+    }.
+Unset Primitive Projections.
 Parameter B : Set.
 
 Type
--- a/test-suite/success/Fixpoint.v
+++ b/test-suite/success/Fixpoint.v
@@ -11,23 +11,28 @@ Fixpoint f (n:nat) (m:=pred n) (l:listn m) (p:=S n) {struct l} : nat :=
 
 Eval compute in (f 2 (consn 0 0 niln)).
 
-CoInductive Stream : nat -> Set :=
-  Consn : forall n, nat -> Stream n -> Stream (S n).
+Set Primitive Projections.
+CoInductive Stream (n : nat) : Set :=
+  Consn {
+    Stream_hd : nat;
+    Stream_tl : Stream (pred n);
+    Stream_eq : 0 <> n;
+  }.
 
 CoFixpoint g (n:nat) (m:=pred n) (l:Stream m) (p:=S n) : Stream p :=
     match n return (let m:=pred n in forall l:Stream m, let p:=S n in Stream p)
     with
-    | O => fun l:Stream 0 => Consn O 0 l
+    | O => fun l:Stream 0 => Consn (S O) 0 l (O_S 0)
     | S n' =>
       fun l:Stream n' =>
       let l' :=
-        match l in Stream q return Stream (pred q) with Consn _ _ l => l end
+        match l in Stream _ return Stream _ with Consn _ _ l _ => l end
       in
-      let a := match l with Consn _ a l => a end in
-      Consn (S n') (S a) (g n' l')
+      let a := match l with Consn _ a l _ => a end in
+      Consn (S (S n')) (S a) (g n' l') (O_S (S n'))
    end l.
 
-Eval compute in (fun l => match g 2 (Consn 0 6 l) with Consn _ a _ => a end).
+Eval compute in (fun l => match g 2 (Consn (S 0) 6 l (O_S 0)) with Consn _ a _ _ => a end).
 
 (* Check inference of simple types in presence of non ambiguous
    dependencies (needs revision 10125) *)
--- a/test-suite/success/Inductive.v
+++ b/test-suite/success/Inductive.v
@@ -67,9 +67,11 @@ Check (fun x:I1 =>
 Set Implicit Arguments.
 Unset Strict Implicit.
 
-CoInductive LList (A : Set) : Set :=
-  | LNil : LList A
-  | LCons : A -> LList A -> LList A.
+Set Primitive Projections.
+
+CoInductive LList (A : Set) : Set := { llist : option (prod A (LList A)) }.
+Definition LNil A := @Build_LList A None.
+Definition LCons A x l := @Build_LList A (Some (pair x l)).
 
 Arguments LNil [A].
 
--- a/test-suite/success/RecTutorial.v
+++ b/test-suite/success/RecTutorial.v
@@ -1083,22 +1083,14 @@ Proof.
 Qed.
 
  Set Implicit Arguments.
+ Set Primitive Projections.
 
- CoInductive Stream (A:Set) : Set   :=
- |  Cons : A -> Stream A -> Stream A.
+ CoInductive Stream (A:Set) : Set   := Cons { head : A; tail : Stream A }.
 
- CoInductive LList (A: Set) : Set :=
- |  LNil : LList A
- |  LCons : A -> LList A -> LList A.
+ CoInductive LList (A: Set) : Set := { llist : option (prod A (LList A)) }.
 
-
-
-
-
- Definition head (A:Set)(s : Stream A) := match s with Cons a s' => a end.
-
- Definition tail (A : Set)(s : Stream A) :=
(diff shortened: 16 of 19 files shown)


plugins/ssr/ssrbool.v
```diff
@@ -585,9 +587,11 @@ Lemma rwP2 : reflect Q b -> (P <-> Q).
 Proof. by move=> Qb; split=> ?; [apply: appP | apply: elimT; case: Qb]. Qed.
 
 (*  Predicate family to reflect excluded middle in bool.                      *)
-CoInductive alt_spec : bool -> Type :=
+Unset Elimination Schemes.
+Inductive alt_spec : bool -> Type :=
   | AltTrue of     P : alt_spec true
   | AltFalse of ~~ b : alt_spec false.
+Unset Elimination Schemes.
```
gares: You mean `Set` here?

gares (dismissed):
I think that we can live with the eliminators being generated in ssr.
(unless there is a name conflict, and I don't think there are).

ejgallego:
@ppedrot IMHO you should push a CI build making this warning fatal to measure the true impact of this change.

gares:
cc @amahboubi do you recall why `CoInductive` is used in MC? Was it just not to generate eliminators?

SkySkimmer:
It may be useful to provide more guidance for translating coinductives with multiple constructors and with indices.

herbelin:
I'm fine with this PR but what would then be your recommended alternatives to e.g.:
```coq
CoInductive conat : Type := O : conat | S (n:conat) : conat.
```
and
```coq
Require Import Streams.
Inductive tree : Type := Node : Stream tree -> tree.
```

ppedrot:
@herbelin, @SkySkimmer 

```
CoInductive conat : Type := O : conat | S (n:conat) : conat.
```
becomes
```
CoInductive conat := { Conat : option conat }.
```
and @herbelin's second example is still valid. More generally, replacing the constructors is done by first writing a inductive with the same shape but which has one parameter `R : Type` instead of the recursive calls, and you instantiate that as a nested coinductive type.

As for indices, you have to add them by hand as equalities in the record constructor of the coinductive type (or using the same inductive shape trick).

Allowing mixed inductive-coinductive declarations would make that more natural, but unluckily this is not implemented as of today.

plugins/ssr/ssrbool.v
```diff
@@ -1376,7 +1386,9 @@ Notation "[ 'qualify' 'an' x : T | P ]" := (Qualifier 2 (fun x : T => P%B))
 Section KeyPred.
 
 Variable T : Type.
-CoInductive pred_key (p : predPredType T) := DefaultPredKey.
+Set Elimination Schemes.
+Inductive pred_key (p : predPredType T) := DefaultPredKey.
+Unset Elimination Schemes.
```
amahboubi: Set and Unset have been swapped here, haven't they?

plugins/ssr/ssreflect.v
```diff
@@ -275,10 +277,14 @@ Notation "{ 'type' 'of' c 'for' s }" := (dependentReturnType c s)
...
-CoInductive phantom T (p : T) := Phantom.
+Unset Elimination Schemes.
+Inductive phantom T (p : T) := Phantom.
+Set Elimination Schemes.
 Arguments phantom : clear implicits.
 Arguments Phantom : clear implicits.
-CoInductive phant (p : Type) := Phant.
+Unset Elimination Schemes.
+Inductive phant (p : Type) := Phant.
+Unset Elimination Schemes.
```
amahboubi: Do you mean Set here?

amahboubi:
Indeed, as far as I know CoInductive is used to avoid generating eliminators here. Would it be possible to implement a variant of `Inductive` which packages:
```
Unset Elimination Scheme.
Inductive ...
Set Elimination Scheme.
```
? I am afraid I do not have a good syntax/name to propose though, which might be a sign that it is not a good idea...

ppedrot:
@amahboubi This is actually a pretty clear use case of attributes. Last time I heard about it, we were waiting for @vbgl to finish the implementation.

plugins/ssr/ssrbool.v
ppedrot: Yes.

plugins/ssr/ssreflect.v
ppedrot: Indeed.

ppedrot:
So it seems compcert was using Coq streams. I'll see if I can come up with a backward-compatible fix.

SkySkimmer:
`Variant` doesn't scheme, although it only works for non recursive types.

amahboubi:
Thanks @ppedrot ! @vbgl, is there a place where I can see what attributes are/will be? I could only find 
https://github.com/vbgl/coq/commits/attributes.

gares:
> Variant doesn't scheme, although it only works for non recursive types.

I was trying to look at the doc for `Variant` but it seems empty: https://coq.inria.fr/distrib/current/refman/language/gallina-extensions.html#coq:cmd.variant . I arrived here from the index.

gares:
I found the doc, but it contains no index entry:
  https://coq.inria.fr/distrib/current/refman/language/gallina-specification-language.html#variants
:-(

According to the doc, it is what we want.

Zimmi48:
@amahboubi There is this CEP https://github.com/coq/ceps/pull/31 (but it's probably not up-to-date).

ejgallego:
cc: #6938

ppedrot:
I've added overlays to CompCert and MathClasses that simply add a copy of the legacy Streams module inside their code. The checker seems to choke on negative coinductive streams though.

amahboubi:
Hi @Zimmi48, thanks for the pointer to the relevant CEP.
@gares : a while ago, I tried to understand if `Variant` was what we needed in mathcomp. At the time I had some trouble understanding `Variant` vs `Inductive` vs `Record`, and I submitted issue https://github.com/coq/coq/issues/5303. It seems that @herbelin's fix has solved all the blocking issues though so we might try to use it now.

herbelin:
@amahboubi: isn't otherwise `Unset Elimination Schemes` convenient for your need?

herbelin:
@ppedrot:
```coq
CoInductive conat := { Conat : option conat }.
```

OK, I see. So, you don't want to remove coinductive types with positive connectives inside, you only want to guard the recursive call in a coinductive type with at least one occurrence of a primitive projection at the head.

And if you want to do this, this is as a way to forbid dependent elimination for coinductive types and thus as a way to address the subject reduction failure, right?

More generally, I understand that your point is to forbid any ways which provide iterated η-expansions without keeping a marker of how many unfolding have been done. With this respect, `eqst` is a "good" way to prove η-expansion (up to the identification of `EqSt` and `=`) as it uses an unreducible expression `{ steq_tl := ... }`, but `match`, which may disappear in an interaction with its argument, is not.

Also, I wonder whether it would make sense to symmetrically guard any inductive type by a "positive" "constructor", i.e. to forbid `Inductive I := { it : I }` when the record is negative. For instance, in
```coq
Set Primitive Projections.
Inductive I := { it : I }.
Fixpoint f x := f x.(it).
```
is `f` morally acceptable even though there is no positive constructor to guard it? Or do we accept it because it is always possible to consider that there is an implicit positive `Build_I` behind the scene which can be used as unlocking the unfolding of the recursive function?

But then, couldn't we do the same for coinductive types and assume that in `Inductive conat := O | S (n:conat)`, there is also an implicit native projection encapsulating the positive part? Otherwise said, couldn't we just keep the syntax as it is but instead restrict `match` to behave as if there was an implicit `Conat` projection, i.e. by ensuring that the dependency in the coinductive argument in the return clause of a `match` is itself protected by a `match`? At least, this would rule out this [example](https://github.com/coq/coq/issues/6768#issue-297400317) of failure of subject reduction.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1031,6 +1031,51 @@ we have to construct an infinite proof of equality, that is, an infinite object
...
+*positive co-inductive types*, is known to break subject reduction. It is a
+fundamental issue related to the duality of computational behavior between
+inductive and co-inductive types.
+
+Since the introduction of primitive records in Coq 8.5, an alternative
+presentation is available, called *negative co-inductive types*. This consists
+in defining a co-inductive type as a primitive record type through its
+projections. Such a technique is akin to the *co-pattern* style that can be
+found in e.g. Agda, and preserves subject reduction.
+
```
herbelin: Can you clarify that the problem is not with negative

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1031,6 +1031,51 @@ we have to construct an infinite proof of equality, that is, an infinite object
 of type :g:`(EqSt s1 s2)`. We will see how to introduce infinite objects in
 Section :ref:`cofixpoint`.
 
+Caveat
+++++++
+
+The ability to define co-inductive types by constructors, hereafter called
+*positive co-inductive types*, is known to break subject reduction. It is a
+fundamental issue related to the duality of computational behavior between
+inductive and co-inductive types.
```
herbelin: Can you explain more? I thought that the problem was about the undecidability of equality over infinite objects. Are you saying that a similar dual subject reduction problem can be obtained with inductive types?

ppedrot:
> OK, I see. So, you don't want to remove coinductive types with positive connectives inside, you only want to guard the recursive call in a coinductive type with at least one occurrence of a primitive projection at the head.
>
> And if you want to do this, this is as a way to forbid dependent elimination for coinductive types and thus as a way to address the subject reduction failure, right?

Correct.

> Also, I wonder whether it would make sense to symmetrically guard any inductive type by a "positive" "constructor", i.e. to forbid Inductive I := { it : I } when the record is negative.

Actually I'm surprised that this is accepted by the kernel. Isn't that another source of SR failure?

> But then, couldn't we do the same for coinductive types and assume that in Inductive conat := O | S (n:conat), there is also an implicit native projection encapsulating the positive part?

We could do that indeed, but I'm afraid that it would lead to a rather intricate implementation of the guard condition. The nice part about describing cofixpoint through co-patterns is that you clearly see that something is decreasing syntactically, namely the context.

> Otherwise said, couldn't we just keep the syntax as it is but instead restrict match to behave as if there was an implicit Conat projection, i.e. by ensuring that the dependency in the coinductive argument in the return clause of a match is itself protected by a match?

I don't have a proof of that but I think that would work.

amahboubi:
@herbelin Yes, in principle,and this is what @ppedrot did here, but then you have to re-`Set Elimination Schemes' back after the definition, which again is ok but a bit painful. See for instance the typos @gares and myself found while reviewing the present PR. `Variant` seems to be exactly what we need in fact. @herbelin, do you know where this name comes from?

herbelin:
> Variant seems to be exactly what we need in fact. @herbelin, do you know where this name comes from?

The terminology was introduced by @aspiwack. As far as I understood, it is reminiscent of the use of the term [variant type](https://en.wikipedia.org/wiki/Variant_type) in programming.

aspiwack:
The motivation for introducing `Variant` was, in part, to avoid srr's need for non-recursive coinductive types, just to avoid generating a scheme.

The reason for the name, was that there already was a non-recursive `Record` keyword. But I was told that reusing `Record` for types defined as variants would be confusing (which, to be fair, is more than believable). So, I bit the bullet and used `Variant` by symmetry. I could have used `BiInductive`, which kind of makes sense, but I deemed it ugly, and it left `Record` hanging.

By the way, before I implemented `Variant`, the development version of Coq had a flag (turned on by default) which prevented `Record`-s from generating schemes. So I just had to generalise (and rename) this flag before it made it into a release :slightly_smiling_face:

amahboubi:
Thanks for the reference, it was very useful! The name seems to refer to 2 unrelated things then. I still do not know how I would teach it though, specially to an audience less cultured in CS (including myself). @aspiwack, what other usage did you have in mind, apart from the specification predicates used in mathcomp. Is there any library already using ``Variant``?

mattam82:
One can think of it as a generalization of C’s enum as well, if that can
help.
Le mer. 23 mai 2018 à 14:48, Assia Mahboubi <notifications@github.com> a
écrit :

JasonGross:
> Is there any library already using Variant?

I think we use it in a couple of places in fiat-crypto (or did at one point) (maybe also cross-crypto), to declare non recursive inductive enumeration type.

aspiwack:
> The name seems to refer to 2 unrelated things then. I still do not know how I would teach it though, specially to an audience less cultured in CS (including myself).

Well, not entirely. You use `Variant` in Coq to introduce a non-recursive variant type (and `Record` for a non-recursive record type). When you need recursive thing, you have to specify `Inductive` or `CoInductive`, depending on what kind of fixed point you need. I'd be entirely happy with someone finding a good name for the non-recursive type and deprecate both `Record` and `Variant`, though.

At the end of the day, these are just names. As long as you know they exist, it doesn't matter too much what they are. But of course, good names facilitate learning.

> what other usage did you have in mind, apart from the specification predicates used in mathcomp.

To be honest, I don't remember. I mean, clearly, the usage I had in mind was telling Coq I was defining a non-recursive data type. One of the benefit I was hoping to gain from this was to prevent automatic generation of induction schemes in that case, motivated by ssr's use of `CoInductive` for this reason.

Using `CoInductive` bothered me for two reasons. First, it is incompatible with changes such as this PR: dependent pattern matching on coinductive types is problematic (one way to see it is that on coinductive records, it is roughly the same as eta-conversion, which is undecidable), but it is, of course, used for the non-recursive types in ssreflect. The mere mention that dependent pattern matching on coinductive was a historical mistake (though made knowingly) used to make Georges quite uncomfortable. So `Variant` was an answer to this.

The other reason is that `CoInductive` is fundamentally wrong for the job: even non-recursive coinductive can be built by co-fixed points, which means that they are suspended. Extensionally, it makes no difference, since everything terminates, but intentionally, non-recursive co-inductive type have one extra point (which is the reason why I implemented `Variant` as inductive types rather than co-inductive).

The rest of the considerations, I'm not entirely sure, they were probably of more theoretical a nature. Maybe I was hoping that non-recursive types could eventually be treated specially by Coq, in order to have a stronger conversion (_e.g._ doing some case-splitting during conversion). I had already started looking into sequent calculus at the time, and maybe this factored in my feeling that this was a missing feature of Coq.

It was also the time where I introduced recursive (both `Inductive` and `CoInductive`) records. Partly to introduce the style of programming with coinductives that this PR makes official. Mostly actually. But this created a weird gap in that we had inductive and codinductive variants and records, as well as non-recursive records. But non-recursive variants were absent. So maybe it was just me being weirded out by this, and I saw that as cleaning.

herbelin:
> I'd be entirely happy with someone finding a good name for the non-recursive type and deprecate both `Record` and `Variant`, though.

Note that OCaml supports for instance
```ocaml
type sum_of_records = C of { prc1 : int ; prc2 : int } | D of { prd : int }
```
Why couldn't we support it too? But to do so, we would also need a terminology which subsumes both `Variant` and `Record`.

IIRC, @gares had proposed once `Type` which by default would mean non-recursive. Then `Inductive` and `CoInductive` would be attributes, say `Type [Inductive]` or whatever syntax, and `Inductive` be a compatibility notation for `Type [Inductive]`. But maybe too big changes at the moment. Maybe to be considered in relation with expressing indices differently, as in, say:
```coq
Type vect A n := nil when n = 0 | cons p (a:A) & vect A p when n = S p.
```
which could make sense in the context of a relevant equality, as Nicolas T. mentioned it once.

At some time, the question of accepting a type like `sum_of_records` above seemed to me related to the question of whether we can and want to make a distinction between a `match`-based conjunction/∑-type and a projection-based conjunction/∑-type, but the more I'm thinking to this, the more I'm getting confused as the two are isomorphic in the context of intuitionistic logic, in the presence of their η-rules. So, besides the impact of one representation over the other to control the undecidability of η-rules in the coinductive case, as this PR does, I'm a bit lost. Don't know whether @ppedrot. @mattam82 or @aspiwack (or maybe GMM, or maybe GS) have ideas clearer on the question or if we should brainstorm further.

aspiwack:
I don't have clear ideas, just vague intuitions. I suggest, however, that this particular discussion should be held somewhere else.

ppedrot:
> I suggest, however, that this particular discussion should be held somewhere else.

I can only support this statement...

xavierleroy:
The discussion at https://github.com/AbsInt/CompCert/pull/231 shows that the proposed changes to the Streams standard library break not just CompCert but more generally parsers generated by Menhir to Coq.  This said I know of only two such parsers in use (CompCert and Velus) and @jhjourdan has plans for alternatives to Streams, so we will manage.

This said again, this PR fails my expectations regarding Coq's development process, namely that for every important change between versions N and N+1 a way should be provided for users to make their development compatible with both versions N and N+1.  Here I was informed that the changes to the Streams library are breaking and nothing will be done about it.

ejgallego:
I didn't follow this issue closely but I don't see why this change should be breaking the property Xavier mentions, Coq 8.9 could ship both stream libraries, tho the `Makefile` for users would need some tweaks to select the legacy or negative version or even better they could use `coqfind`.

Zimmi48:
> This said again, this PR fails my expectations regarding Coq's development process, namely that for every important change between versions N and N+1 a way should be provided for users to make their development compatible with both versions N and N+1. Here I was informed that the changes to the Streams library are breaking and nothing will be done about it.

I can only agree that this is a problem here. Since plans are supposed to be communicated very soon about a Standard Library 2.0, with many breaking changes, wouldn't it make sense to do this change to the Streams library as part of this more general project? In the meantime, this PR could still get in but the Streams library would need to continue to use deprecated features (using a warning directive to turn the warning off locally).

ppedrot:
@Zimmi48 Is there a way to mark a Coq definition as deprecated? This would be useful in this case. If so I'm fine with leaving the legacy implementation of `Streams` in the next version. At some point we need to perform the switch though, and I don't know of any easy way to do it in Coq directly. The only sane way I can think of is to externalize streams as an independent library so as to version it in a Coq-independent way.

ejgallego:
> The only sane way I can think of is to externalize streams as an independent library so as to version it in a Coq-independent way.

@ppedrot I'd suggest using the upcoming `coqfind` + package format for this. I think that should be ready before September.

This way, users could declare dependencies such as:
```
(libraries stdlib.streams)
```
vs
```
(libraries stdlib.legacy-streams)
```
and the `Stream` prefix should be bound to the proper path.

Zimmi48:
Only notations can be used to raise a deprecation warning in a Coq file. But the deprecation message isn't customizable.

@maximedenes Do you want to comment on the possibility to perform this change of implementation as part of the Stdlib 2 project?

maximedenes:
> @maximedenes Do you want to comment on the possibility to perform this change of implementation as part of the Stdlib 2 project?

Sure. A library of streams seems of general interest, and so should be part of Stdlib 2. Going for negative coinductive types instead of what we currently have makes a lot of sense to me. A tool to generate co-induction principles would also be a good candidate for integration IMHO.

herbelin:
I can only share this enthousiasm about extending the deprecation mechanism, about generating automatically coinduction schemes, about developing a more consistent library of `Streams` (consistent in the sense of thickness), but to come back to the initial question, is it clear that we have investigated all issues on how to address the undecidability of η on coinductive types?

After I eventually realized that negative and positive tuples are basically isomorphic in the presence of their η-rules (see #7613), I start to be skeptical that either one or the other of (head-)negative coinductive types and (head-)positive coinductive types is fondamentally better than the other one. My understanding of coinductive types in general is limited but could it be that if we cannot identify both kinds of representation because of rejecting η (due to its undecidability), they however both remain of equal technical and conceptual interest (with appropriate restrictions on each of the presentation, as in particular suggested above, to preserve subject reduction)?

mattam82:
I think as we don't have a well-behaved theory and implementation of positive coinductives which maintains both SR and decidable type-checking currently, it is most reasonable to inform users that they should rather use negative ones instead. The Agda community seems to have made the switch without too many hiccups, as far as I know. 

Technically, the restriction you're talking about of verifying that there are matches in the return clause of a dependent match on a coinductive looks quite complicated syntactically to me, and I'm not sure what it will really means semantically.

ppedrot:
> Technically, the restriction you're talking about of verifying that there are matches in the return clause of a dependent match on a coinductive looks quite complicated syntactically to me, and I'm not sure what it will really means semantically.

For the record, that's **exactly** [Baclofen Type Theory](https://hal.inria.fr/hal-01441829/document), a.k.a. predicates being restricted to algebra morphisms blah blah blah. I find that quite funny, but that might well be a mere coincidence. (Something tells me it isn't, but I have shares in this.)

But yeah, I side with @mattam82 on this one, we have a design that is known to work (negative coinductive types), while the way to make positive coinductive types work still belongs to the realm of fundamental research. So I'd favour deprecating the latter as a first step. Anyways, even assuming we knew how to properly implement them, positive coinductive types would still incur a heavy backwards incompatibility, so...

aspiwack:
Let me insist again that we take research discussions to another venue (probably an issue, please cc me!) and focus, in this PR on the questions at hand: does this PR offer a good enough migration strategy, if not, how could it be devised?  And more generally is it a good idea?

Let's avoid holding this PR because there are problems which are not solved. The only question is whether we want to stay in the partly broken situation or fix it and break some things, with the limited understanding that we have today.

mattam82:
@ppedrot ok I see it now. It's kind of the effect of cofix unfolding that's getting purified here then, ensuring the discriminee's reduction and its copies in the type are kept in sync.

@herbelin assuming we find solutions to the remaining compatibility/transition issues, do you agree with this path of deprecation of positives in favor of negatives for the time being?

Zimmi48:
@aspiwack The other issue you're talking about already exists and is #7613. I think @herbelin's comment here comes from an uneasiness to deprecate positive coinductive types when he thinks that maybe the problems that they have could be fixed instead.

@gares The `Variant` documentation has been fixed in master, cf. https://coq.gitlab.io/-/coq/-/jobs/72012591/artifacts/_install_ci/share/doc/coq/sphinx/html/language/gallina-specification-language.html#variants

Zimmi48:
My personal (unexperimented) feeling is that the negative presentation of co-inductive types is more natural, and more symetric. However, I feel like the documentation should be refactored further. Indeed, only the positive presentation is currently described in the first chapter of the reference manual. Records and primitive projections are only mentioned in the first section of the second chapter (Extensions of Gallina). The presentation of co-recursive functions also uses the positive presentation of co-inductive types...

That you must wait the second chapter to hear about primitive projections gives them a less important status that what they actually represent: they have their own constructor in the type of terms, and yet, they are not mentioned in the grammar of terms at the beginning of the Gallina chapter. IMHO sections, modules, advanced match constructs and existential variables are truly extensions and are thus well located in the second chapter but it is not the case for primitive projections which are more fundamental and are not activated by default for compatibility reasons only.

JanBessai:
Maybe it helps to have an example that is larger/different from streams.
Here is an implementation of an efficient dependently typed coinductive memoization structure:
- In the [old style](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664/086cbc7edc6748a22ef61eadc9deb78416552084) using constructors
- In the [new style](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664/35675340d2e6785d47aea7cc0d675541b1779612) using records
- As [a diff](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664/revisions?diff=split) between the two

My résumé   of the changes:
- The [data-type definition](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664#file-memotree-v-L11-L19) of `Delay` got slightly [more complicated](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664/086cbc7edc6748a22ef61eadc9deb78416552084#file-memotree-v-L10-L12) involving an extra type to encode the sum. It still works when changing `Variant` to `Inductive` in `AllowRight`. I'm not sure where Variant would be necessary, but it surely introduces the risk of having to carefully update the standard library and use the least possible thing in every place. For example `sumbool`, `pair`, `or`, ... should all be `Variant` to avoid having to copy them, when a `Variant` version is needed.
- I find the trees easier to understand when they have proper record labels. It feels like traditional OO-Programing. Though I'm not sure how to describe them in written informal math text. For inductive types I'd write "the least set closed under the rules...". What is the correct analogon for the coinductive definition with records?
- [Function `get_memo`](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664#file-memotree-v-L57) got more compact using let instead of [let + match](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664/086cbc7edc6748a22ef61eadc9deb78416552084#file-memotree-v-L51-L55).
- [Predicate `IsMemoTree`](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664#file-memotree-v-L80-L83) can avoid [forall-quantifying](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664/086cbc7edc6748a22ef61eadc9deb78416552084#file-memotree-v-L68) over the tree, which is nice. It allows to avoid [the `IsFMemoTree_parts` tactical](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664/086cbc7edc6748a22ef61eadc9deb78416552084#file-memotree-v-L92-L96) to get rid of existential equalities introduced by inversion. However, I'm not sure why Coq (8.7.2) doesn't infer the implicit parameter `n` in `leftBranchOk` and `rightBranchOk` when using `IsFMemoTree (leftBranch t)` instead of `@IsFMemoTree _ (leftBranch t)`.
- The [ugly `unroll_MemoTree`-hack](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664/086cbc7edc6748a22ef61eadc9deb78416552084#file-memotree-v-L70-L73) is gone. This is a huge plus :+1: .
- In `IsFMemoTree_memo` I'm not sure why `constructor; auto using IH`, which was fine in [the previous version](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664/086cbc7edc6748a22ef61eadc9deb78416552084#file-memotree-v-L89) no longer works in [the record version](https://gist.github.com/JanBessai/5d68965507368ce597538b99a4f73664#file-memotree-v-L97-L100).


Aside from the example I wonder how [Paco](https://plv.mpi-sws.org/paco/) is related to the considerations.

SkySkimmer:
> I'm not sure where Variant would be necessary

Variant is just Inductive + check that it's not recursive + disable automatic generation of schemes (foo_ind etc), it's never necessary.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1045,6 +1045,51 @@ co-inductive definitions are also allowed.
...
+
+   Set Primitive Projections.
+   CoInductive Stream : Set := Seq { hd : nat; tl : Stream }.
+   CoInductive EqSt (s1 s2: Stream) : Prop := eqst {
+     eqst_hd : hd s1 = hd s2;
+     eqst_tl : EqSt (tl s1) (tl s2);
+   }.
+
+Some properties that hold over positive streams are lost when going to the
+negative presentation, typically when they imply equality over streams.
```
maximedenes: imply or involve?

maximedenes (dismissed):
I support the change, but I would like to first have support for mutual negative coinductives. Is it part of your other PR on mutual records?

doc/sphinx/language/gallina-specification-language.rst
ppedrot: Both are true!

plugins/ssr/ssrbool.v
```diff
@@ -456,9 +456,11 @@ Section BoolIf.
 
 Variables (A B : Type) (x : A) (f : A -> B) (b : bool) (vT vF : A).
 
-CoInductive if_spec (not_b : Prop) : bool -> A -> Set :=
+Unset Elimination Schemes.
+Inductive if_spec (not_b : Prop) : bool -> A -> Set :=
```
JasonGross: Why use `Unset Elimination Schemes. Inductive` rather than `Variant`?  It has the same effect, right?

theories/Lists/Streams.v
```diff
@@ -144,7 +132,7 @@ Inductive Exists ( x: Stream ) : Prop :=
   | Further : Exists (tl x) -> Exists x.
 
 CoInductive ForAll (x: Stream) : Prop :=
-    HereAndFurther : P x -> ForAll (tl x) -> ForAll x.
+    HereAndFurther { Forall_here : P x; Forall_further : ForAll (tl x) }.
```
JasonGross: Reuse the capitalization of above: `ForAll_*` rather than `Forall_*`?

theories/Lists/Streams.v
```diff
@@ -175,11 +164,11 @@ Variables A B : Type.
 Variable f : A -> B.
 CoFixpoint map (s:Stream A) : Stream B := Cons (f (hd s)) (map (tl s)).
 
-Lemma Str_nth_tl_map : forall n s, Str_nth_tl n (map s)= map (Str_nth_tl n s).
+Lemma Str_nth_tl_map : forall n s, EqSt (Str_nth_tl n (map s)) (map (Str_nth_tl n s)).
```
JasonGross: Is this no longer true with the `=`?

doc/sphinx/language/gallina-specification-language.rst
herbelin: @ppedrot: In the absence of answers to previous questions, allow me to ask again: wouldn't it be more correct to say that it is not the move to positive to negative which makes that some properties implying/involving equality are lost but only the amount of η-equality that we allow in each case.

Otherwise said, we could restrict η-equality for the positive formulation or we could add more η-equality for the negative formulation, and this is what matters for the "more equalities over streams".

My understanding of the reason why people tends to consider that negative streams solve 
the extensionality of equality issue leading to a loss of subject reduction is that η is clearly disjoint from the logical rules in the negative case, while it comes implicitly in the dependent form of `match` in the positive case. Couldn't a view be adopted which separates concerns and which questions the relevance of necessarily hard-wiring some form of extensionality in dependent `match`?

mattam82: From what I understand, for a decidable, SR preserving positive treatment of coinductives we'd need to switch to Baclofen type theory or something similar which restricts eta-equalities on the positive coinductives/avoids the implicit eta-equality which is part of dependent elimination. I think we all agree on that no? I suspect having such a variant is part of @ppedrot 's research project as well :) Maybe we could add something to that effect to the reference manual explanation?

ppedrot: > SR preserving positive treatment of coinductives we'd need to switch to Baclofen type theory or something similar which restricts eta-equalities on the positive coinductives/avoids the implicit eta-equality which is part of dependent elimination.

That's my understanding, but this is still a rough one. So yes, it boils down to the problem of eta-rules, but all of this is mixed together in the current presentation of type theory.

> Couldn't a view be adopted which separates concerns and which questions the relevance of necessarily hard-wiring some form of extensionality in dependent match?

As explained by @mattam82, that's just BTT. But practical usability of that theory is not clear yet.

theories/Lists/Streams.v
ppedrot: Nope.

plugins/ssr/ssrbool.v
ppedrot: I am personally not super convinced by the magical `Variant` keyword that makes coffee. We should strive for atomic behaviours, because it makes decoupling easy. At least the intent is clear here, and we could switch to attributes when they're available.

herbelin:
@mattam82, @ppedrot: I think we agree on the analysis, yes. The reason for such big moves that @ppedrot are mentioning remains however unclear to me. Wouldn't a BTT-like restriction of `match` for coinductive types adding a test that any occurrence of `x` in the return predicate is exactly itself protected by a `match` already fix the subject reduction issue? As far as I can see, this would only affect the attempts to prove eta-like statements on coinductive types and would otherwise presumably preserve most of existing developments.

In particular, with such a minimal change, why would have the guard condition or tactics to be changed? I don't understand what you have in mind when you say, @ppedrot, that that would still require an important reworking of both the kernel and the tactic layers.

ppedrot:
> As far as I can see, this would only affect the attempts to prove eta-like statements on coinductive types and would otherwise presumably preserve most of existing developments.

I don't think so. Virtually all uses of coinductive types in proof use `destruct x` over a goal that is **not** of the form `match x with ... end`. So in practice that means that it'll fail immediately.

herbelin:
I just tried to pass `Streams.v` (mentally) and it seems that only the eta-expansion `unfold_Stream` fails (of course). The 8 calls to `case` are non dependent and the unique call to `intros (a,b) (c,d)` mentions only the `hd` or `tl` of a stream. Not a big library though...

ppedrot:
@herbelin After a quick look, `Str_nth_tl_map` and `Str_nth_tl_zipWith` are also wrong, which in turns requires changing the lemmas that depend on them.

herbelin:
Are you sure? I see only an `induction n` in `Str_nth_tl_map`.

ppedrot:
@herbelin I had to change them for this PR, so I am pretty sure. The problem comes from the calls to `intros` with a splitting pattern, which is essentially performing dependent induction.

theories/Lists/Streams.v
ppedrot: Let me correct myself, this looks indeed still true with equality. Should I fix this?

herbelin:
There is an `intros` with splitting pattern in `Str_nth_tl_zipWith` but not in `Str_nth_tl_map`.

And the `intros` with splitting pattern in `Str_nth_tl_zipWith` only involves `hd` or `tl` of the streams `a` and `b` (after reduction, if I'm not mistaken).

ppedrot:
Indeed, I was wrong, we do not use dependent induction on this. And actually the result still hold with primitive streams, so that I can adapt the PR to keep equality.

ppedrot:
Rebased and adapted now that we have mutual records.

test-suite/failure/cofixpoint.v
```diff
@@ -1,9 +1,10 @@
 (* A bug in the guard checking of nested cofixpoints. *)
 (* Posted by Maxime Dénès on coqdev (Apr 9, 2014).    *)
 
-CoInductive CoFalse := .
+Set Primitive Projections.
+CoInductive CoFalse := { coFalse : False }.
 
-CoInductive CoTrue := I.
+CoInductive CoTrue := { coTrue : True }.
```
JasonGross: Wouldn't the appropriate translation of this be
```coq
CoInductive CoTrue := I {}.
```
?

SkySkimmer: That's no a primitive record.

herbelin: > That's no a primitive record.

I'm lost. In which sense do you say that?

SkySkimmer: A primitive record must have at least one field.

herbelin: Ah, ok, but then this is in Coq custom-made terminology.

SkySkimmer: >Ah, ok, but then this is in Coq custom-made terminology.

I don't know what that means.

herbelin: I mean that whatever way we morally define record and primitive, i.e. by record = negative, or record = with (explicit) primitive projections, or record = with compactly represented projections, or record = with {| ... |} syntax, and whatever way we morally define primitive, i.e. by primitive = negative, or primitive = with explicit primitive projections, or primitive = with compactly represented projections, there is no reason to make a difference between a tuple type with at least 1 projection and a tuple type with no projections.

Then, the only possible explanation is that the terminology that you are using comes from the fact that Coq does not implement η for tuple types with no projections. This is in this sense that I say that this is a Coq custom-made terminology.

gares:
@ppedrot could you split the commit fixing the stdlib in two, one just for ssr. Then put it aside and try to simply `CoInductive -> Inductive`. If it works, as I expect, then I prefer this to the unset/set litany. Or Variant, assuming we understood what it does (I'm not sure I did).

The point being that this was a *minuscule* optimization (no need to generate the eliminators). If it costs more than 2 characters (eg `Co`) then I'm not so sure it is worth it.

ppedrot:
For the record, I have a branch that implements BTT for positive coinductive types. Writing it down made me realize the main limitation of this approach compared to the negative one: the linearity check is highly non-modular, in a way that is worse than conversion.

Essentially to check that `x` appears linearly in a term `t`, one may need to perform **full reduction** of `t`, including notably arbitrary δ reduction if `x` appears as an argument of a constant. This could be very slow in general, while having a dedicated term node would ensure statically that the return clause starts with a linear pattern-matching.

The current check made in my branch is just checking a fast-path on the provided return clause and then goes directly on normalizing it. I find this quite inelegant.

ppedrot:
I have implemented the changes suggested by @gares, and I have also reintroduced the `Streams` library as it was before, with the warning turned down.

I am quite suspicious of the latter change, actually. CompCert will definitely have to stick to the new library at some point and I don't agree with @xavierleroy's analysis of the situation: it is perfectly possible to make CompCert compatible with both 8.8 and 8.9 at the same time, this is what the PR overlay is doing. The implicit backward compatibility contract of Coq doesn't state that we shouldn't break developments, otherwise it would be impossible to go forward. What it says is that it should be *possible* to make a development compatible with Coq versions *n* and *n+1*, which is very different. Otherwise I don't understand how it will be possible to get rid of the legacy stream library.

maximedenes:
> I don't agree with @xavierleroy's analysis

Can you give a pointer?

ppedrot:
@maximedenes This is just above in this conversation, at https://github.com/coq/coq/pull/7536#issuecomment-392344093.

Zimmi48:
@ppedrot Your analysis of the definition of "backward compatibility" is right. The patch is here https://github.com/ppedrot/CompCert/commit/09ed44d2e75b1accfd0c1f56b6cfdb3539d70ba9 and I must agree: just copying the library from Coq to CompCert is a pretty good way of making the patch backward compatible... So to me it is indeed a satisfying solution...

xavierleroy:
**@Zimmi48's theorem.**  _All changes to the Coq standard library are backward compatible._

Proof.  Every Coq development can take a private copy of the Coq standard library, making these developments immune to changes in Coq's standard library.  Qed.

maximedenes:
I see @xavierleroy 's point. What we could do is:
- In 8.9: deprecate positive coinductive types, provide a library of negative Streams, keep the one of positive Streams (locally disabling the warning)
- In 8.10: remove the library of positive Streams (and maybe positive coinductive types, or we could do that one cycle later)

ppedrot:
@maximedenes Whatever the decision is, the current PR is, as far as I am concerned, ready. It doesn't remove the legacy stream library, and whichever transition scheme we stick to, this can be done at a later time.

maximedenes:
> It doesn't remove the legacy stream library,

Sorry, I realized this after posting. But then, I'm a bit lost. Why is any overlay needed at all, then?

ppedrot:
The overlay is not needed, I can get rid of it.

maximedenes:
> CompCert will definitely have to stick to the new library at some point

Let me rephrase: where is this library? Sorry, I struggle finding information in this giant PR discussion.

xavierleroy:
Dear colleagues: my increasingly tongue-in-cheek comments are intended to make you think harder about best practices for evolving standard libraries.  I have the feeling some of you don't understand the word "compatibility" at all.

In the particular case of the Stream library, whatever you do, we'll survive.  It's a trivial library, probably doesn't have many users, and may not be needed in the future, in the case of CompCert at least.

ppedrot:
@maximedenes The updated stream library is in the history of this PR, at 0bf3f2761b3be7bc35e20799868aaa4e619f7ebf more precisely.

maximedenes:
> @maximedenes The updated stream library is in the history of this PR, at 0bf3f27 more precisely.

Thanks!

> I have perfectly no idea on how to do that.

So what is hard about making a version where the two libraries coexist?

ppedrot:
> So what is hard about making a version where the two libraries coexist?

In terms of pure code, nothing. But how should we name the deprecated and the replacement libraries so that the transition is easy?

mattam82:
@ppedrot You suggest to add a term node / variant of match predicate
ensuring linearity, how would that look like precisely?
Le mar. 24 juil. 2018 à 15:35, Maxime Dénès <notifications@github.com> a
écrit :

maximedenes:
> But how should we name the deprecated

That one should stay the same for the next version.

> and the replacement libraries

Pick a name that doesn't contain `New` ;) Maybe something involving codatatypes, or negative something...

If needed, in 8.10, we drop the old `Streams` library and rename the new one into `Streams`, keeping a deprecated alias for one more version.

Isn't that it?

ppedrot:
@mattam82 Just as `Case` but the return clause is not a function `λ(x : indices)(p : I p x). R`, rather an array of functions corresponding to each η-expanded type.

ppedrot:
> That one should stay the same for the next version.

Hum, we did that for the engine API in Coq and it turned out to be a real mess. Your proposal forces the downstream users to change their code *twice*, which is a burden.

I'd rather move `Stream` to another name, and make the clients of the library change their `Require Import` once, and implement the new negative stream directly as `Coq.List.Stream`.

maximedenes:
> make the clients of the library change their Require Import once

But this won't be backward-compatible for them.

ppedrot:
But then we can do that in two steps:
- in 8.9 introduce a new deprecated stream library `Legacy.Streams` that is just a copy of the current one, without touching the latter, with a deprecation warning (if ever we get that in 8.9, sigh) for users of `List.Streams`.
- in 8.10 reimplement negative streams in `List.Streams`
- in 8.11 remove the legacy implementation

mattam82:
@ppedrot What's the type of the whole case then?

ppedrot:
Another linear match whose return branches are a sort, I guess. We probably need some equation for linear cases which do not use their arguments though.

herbelin:
Hi, I'm bit lost on the status of the problem. Is the following formulation faithful to the general perception of the situation:

---
The CIC fails to satisfy subjection reduction in the presence of dependent pattern-matching on coinductive types, which are "positive" types, i.e. types defined by their constructors. Even if this failure of subjection reduction is logically and computationally "safe", it is a bit unsatisfactory from the conceptual point of view.

A natural solution to the problem is to use "negative" types, i.e. "codata", i.e. types defined by their destructors, as was e.g. done in Agda. The main drawback of this solution in the context of Coq is that users shall have to change all code relying on positive coinductive types. Also, standard types such as `colist` would have to be defined in a somehow more convoluted way.

Another solution is Baclofen type theory which is however a relatively young technique compared to negative types, more difficult to implement, and with potential complexity issues at type-checking time. With Baclofen type theory, existing positive definitions of coinductive types could be preserved
and only the proofs which rely on the provability of eta-expansion would be invalidated.

To be noted also is that in the current positive-style library of `Streams`, the only lemma exploiting 
eta-expansion is `unfold_Stream` and that this lemma shall be invalidated anyway.

---

Assuming that the above summary is "consensual" enough, maybe a good approach would be to evaluate what parts of developments among those we track would be invalidated by already rejecting eta-expansion for positive co-inductive types. If all developments using positive coinductive types have anyway to be redone, then it is maybe not asking too much that they be redone using negative coinductive types?

@ppedrot:
> Just as Case but the return clause is not a function λ(x : indices)(p : I p x). R, rather an array of functions corresponding to each η-expanded type.

Couldn't we just have a check of well-formedness at type-checking time of the `match` and keep the same syntax as currently?

ppedrot:
So it is likely that this PR won't make it into 8.9, for a lack of convergence. Nonetheless, I think that the documentation part of it should definitely be part of that release, so that at least people are aware of the problem and the potential solution. Negative coinductive types should be advertised more.

Zimmi48:
@ppedrot If you intend to push the documentation part, let me repeat my comment:

>I feel like the documentation should be refactored further. Indeed, only the positive presentation is currently described in the first chapter of the reference manual. Records and primitive projections are only mentioned in the first section of the second chapter (Extensions of Gallina). The presentation of co-recursive functions also uses the positive presentation of co-inductive types...
>
>That you must wait the second chapter to hear about primitive projections gives them a less important status that what they actually represent: they have their own constructor in the type of terms, and yet, they are not mentioned in the grammar of terms at the beginning of the Gallina chapter. IMHO sections, modules, advanced match constructs and existential variables are truly extensions and are thus well located in the second chapter but it is not the case for primitive projections which are more fundamental and are not activated by default for compatibility reasons only.

ppedrot:
@Zimmi48 But documenting that properly is a lot of work that will not happen for 8.9. I'd rather just have the warning about the loss of SR that is absent from the doc AFAICT, and a short message about how to side-step this. So, essentially a slightly tweaked version of the documentation commit of this PR that doesn't mention the deprecation.

Zimmi48 (commented):
Now that the documentation already mentions the problem with positive co-inductive types, I have even more trouble accepting to go as far as deprecating the thing without refactoring the doc first...

gilhur:
I have one suggestion.

Since the eta expansion is accepted as intensional equality, how about accepting the same expansion as intensional equality?

For eample, consider the following coinductive type.
```
Set Primitive Projections.
CoInductive conat := { Conat : option conat }.
```
As far as I can see, it is not possible to prove the following eta expansion (please correct me if I am wrong).
```
forall n: conat, n = {| Conat := n.(Conat) |}
```

I tried to update existing serious coinductive proofs with primitive projections and the lack of this expansion became the (only) source of major troubles.

ppedrot:
> As far as I can see, it is not possible to prove the following eta expansion (please correct me if I am wrong).

Indeed, this is a archetypal example of a property that is lost.

> I tried to update existing serious coinductive proofs in the proposed approach and the lack of this expansion became the (only) source of major troubles.

This η-rule is akin to function extensionality: it is not provable but it is consistent, and adding it to CIC without breaking good metatheoretical properties (like subject reduction or decidability of type-checking) is hard and an active research area. You can still add it as an axiom though, just as you'd do with funext.

gilhur:
> This η-rule is akin to function extensionality: it is not provable but it is consistent, and adding it to CIC without breaking good metatheoretical properties (like subject reduction or decidability of type-checking) is hard and an active research area. You can still add it as an axiom though, just as you'd do with funext.

This is different from what I understand.

The function extensionality corresponds to the following bisimilarity axiom, which is hard to make provable without breaking subject reduction etc.
```
CoInductive bisim_conat : conat -> conat -> Prop := ...
Axiom bisim_conat_eq: forall n m: conat, bisim_conat n m -> n = m
```

However, the eta equality for funtions, `f = (fun x => f x)`, is currently supported as intensional equality (ie, `reflexivity` proves it).
Similarly, I think the same eta equality for coinductive types, for example `n = {| Conat := n.(Conat) |} `, can be simply treated as intensional equality without breaking any nice property.

Please fix me if my understanding is wrong.
Thanks!

gilhur:
Thanks to [Problem with eta](https://github.com/DeepSpec/InteractionTrees/pull/41#discussion_r), I now understand that eta-equality for recursively-defined coinductive data can cause a subject-reduction problem.

So, I will use the eta-equality axiom for now.

ppedrot:
FTR, I stumbled over [this neat article](http://www.cs.swan.ac.uk/~csetzer/articles/CMCS2018/bergerSetzerProceedingsCMCS18.pdf) by Berger and Setzer about the undecidability of an definitional equality satisfying η-expansion for streams.

maximedenes:
@ppedrot what should be done with this PR? It was marked as `needs: discussion` but I don't remember it being discussed at the last WG, was it?

ppedrot:
It was not discussed at the last WG, but I think I remember we did discuss it at another WG without reaching any consensus. Status is unclear.

- @Zimmi48 wants the documentation about primitive projections and co-inductive types to be wholly rewritten, but that's a huge task I am not really willing to engage into when I am only concerned about metatheoretical properties of Coq.
- @herbelin would like to have a BTT-like system implemented in Coq. While not very hard (I have a branch that does this) I think this is not a very elegant solution. This is indeed solving the issue, but it doesn't make the co-inductive model much clearer either (reasoning about positive coinductive types is weird, with the match-based ι-rule and whatnot)
- I claim this PR is trivial and nonetheless a good step towards a solution whatsoever, as this is just a warning to indicate to our end users that they should switch to negative co-inductive types. We don't have that many people using coinductive types around, I think that by now most users are aware of the concept, through Agda evangelism for instance.

Zimmi48:
That's a good summary and I'll weaken my position a bit: deprecating means intending to eventually remove, so I would just like to know where will the (negative) coinductive types be documented after the removal. This could be in the current section of the first chapter, with an ad-hoc forward link to the section on primitive records, or this could be in the current section on primitive records of the second chapter. [I agree that my initial request (moving the documentation of records and primitive records to the first chapter) is unreasonable for this PR.] If this is the first solution, I would just advise to invert the presentation order in the current section on coinductive type: first the recommended (negative) approach, then the deprecated (positive) approach with the caveat. I claim that it is wrong to put forward the solution that we want to get rid of. I think that such an inversion is reasonable enough to be done as part of this PR.

Zimmi48:
Of course, there is no point in doing so if @herbelin cannot be convinced first.

herbelin:
As far as elegance is concerned, I personally find protecting à la BTT the main argument of the return clause of a coinductive starting with a positive type constructor more "elegant" in the sense that I find it more to the point, i.e. capturing exactly what the problem is, i.e. that in `νX.F(X)` only the `ν` is negative and that there is no reason to force `F` to be negative itself, i.e. no reason to force an indirection  to have `F` starting with a positive type constructor. In particular, dually, we do not force an indirection to use a negative type at the top of an inductive type.

I.e. why to force writing:
```
Variant F A := None | Some : A -> F A.
CoInductive C := { it : F C }.
```
to simulate
```
CoInductive C := None | Some : C -> C.
```
and not to force writing
```
Record F A := { fst : nat ; snd : A }.
Inductive I := C : F I -> I.
```
for the dual situation of a declaration
```
Inductive I := { fst : nat ; snd : I }.
```

Additionally, more developments would be preserved without a change. On the other side, maybe a discussion with Agda people would be worth to know if they would have used BTT-style if it had been known that it could be used.

aspiwack:
I agree with @herbelin that, in the context of this change, the variant syntax for coinductive types should remain with the semantics of a one-field encoding.

I'll admit that my understanding of BTT is really very limited (very!). And I don't understand what a BTT-style limitation would entail for this particular case. I'd be interested if someone could summarise this particular solution space.

maximedenes:
> Additionally, more developments would be preserved without a change. On the other side, maybe a discussion with Agda people would be worth to know if they would have used BTT-style if it had been known that it could be used.

That sounds like a reasonable approach indeed. How about trying to start a discussion on the Agda-list, @ppedrot? After all, such a discussion sounds scientifically interesting.

maximedenes:
I'm afraid we are going to have to bump the milestone, though.

maximedenes:
@ppedrot why put this in the list of topics to be discussed at the next WG?

ppedrot:
@maximedenes should your comment read "why **not** put this"? Otherwise I don't understand it.

In any case, I am still pushing forward this PR. We have a CIC metatheory misdemeanour at hand, something should be done about it.

maximedenes:
> should your comment read "why **not** put this"? Otherwise I don't understand it.

Oh yes, lol, sorry. I obviously meant we could discuss it :)

xavierleroy:
Concerning CompCert specifically: @jhjourdan removed all dependencies on the Stream standard library module, and used a coinductive record type to define streams.  So you can break the Stream module as much as you want.

xavierleroy:
Concerning the usability of coinductive definitions in general: I am still concerned that several uses of coinduction, such as Capretta's delay monad or my and Grall's coinductive big-step semantics, are sums of products, and not products of sums, and therefore will not fit the "only records are coinductive" presentation well.  It would be instructive to try and code Capretta's delay monad using coinductive records.

ppedrot:
@xavierleroy it's not hard to have coinductive sums, you just have to put the sum as a nested inductive inside the record. AFAICT that's what they do in Agda, and in Coq it'd look like this:
```
CoInductive Delay A := { node : A + Delay A }.

Definition now {A} (x : A) := {| node := inl x |}.
Definition later {A} (x : Delay A) := {| node := inr x |}.

CoFixpoint bind {A B} (f : A -> Delay B) (x : Delay A) : Delay B :=
match x.(node _) with
| inl x => f x
| inr y => {| node := inr (bind f y) |}
end.
```

xavierleroy:
> it's not hard to have coinductive sums, you just have to put the sum as a nested inductive inside the record.

Sure, it works for simple cases like infinite lists and the delay monad.  Now if you throw some dependencies in, as in the case of a set of inference rules to be interpreted coinductively, it's less obvious:
```
CoInductive evalinf: term -> Prop :=
  | evalinf_app_l: forall a b,
      evalinf a -> evalinf (App a b)
  | evalinf_app_r: forall a v b,
      eval a v -> evalinf b -> evalinf (App a b)
  | evalinf_app_beta: forall a x c b v,
      eval a (Lam x c) -> eval b v -> evalinf (subst c x v) -> evalinf (App a b).
```
No matter how nicely dual codata is to data, definitions by axioms and inference rules are sums of products, not products of sums.

ppedrot:
@xavierleroy just do what Paco does, i.e. open co-recursion on the inductive predicate. I agree it's not pretty and there should be a bit of language support, but who prevents you from writing the following?
```
Inductive evalinf₀ (self : term -> Prop) : term -> Prop :=
  | evalinf_app_l: forall a b,
      self a -> evalinf₀ self (App a b)
  | evalinf_app_r: forall a v b,
      eval a v -> self b -> evalinf₀ self (App a b)
  | evalinf_app_beta: forall a x c b v,
      eval a (Lam x c) -> eval b v -> self (subst c x v) -> evalinf₀ self (App a b).

CoInductive evalinf : term -> Prop := Evalinf : forall t, evalinf₀ evalinf t -> evalinf t.
```

ppedrot:
I am closing this PR in favour of #10764, which is a tad less backwards incompatible. If ever we judge that positive coinductive types are indeed hard to use compared to their negative variants, we'll always be able to resurrect it.

PR #7537, opened 2018-05-17
Improve the Gallina chapter of the reference manual.

Zimmi48:
Fixes around the documentation of terms.
Including adding missing irrefutable-patterns to the grammar of binders.

Things that are still missing after this commit in the grammar of terms are:
- explicit `Type` annotations (cc @SkySkimmer)
- primitive projections

Additionally, what is `term :>`? It is in the grammar but not described in the rest of the section below.

EDIT: depends on #7606

### DIFF
--- a/doc/sphinx/README.rst
+++ b/doc/sphinx/README.rst
@@ -32,7 +32,16 @@ Names (link targets) are auto-generated for most simple objects, though they can
 - Vernacs (commands) have their name set to the first word of their signature.  For example, the auto-generated name of ``Axiom @ident : @term`` is ``Axiom``, and a link to it would take the form ``:cmd:`Axiom```.
 - Vernac variants, tactic notations, and tactic variants do not have a default name.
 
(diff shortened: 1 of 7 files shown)


Zimmi48:
Thanks! Is that documented anywhere? I guess documenting it is just a matter of adding one sentence in the "Cast" subsection...

SkySkimmer:
>Thanks! Is that documented anywhere? I guess documenting it is just a matter of adding one sentence in the "Cast" subsection...

It's mentioned in https://github.com/coq/coq/blob/master/doc/sphinx/addendum/program.rst#syntactic-control-over-equalities
>Also, :g:`term :>` explicitly asks the system to coerce term to its support type. It can be useful in notations, for example [...]

Zimmi48:
I think we just need to create a specific issue regarding the grammars in the reference manual (see also #7583).

Zimmi48:
I've finished this PR. Note that it is based upon #7606 because in some cases we have a common documentation for several objects. It leaves no object without body in this chapter (sometimes it uses the `:undocumented:` flag, in particular for many error messages). But we cannot automatically check and enforce this (until #7602 is solved).

maximedenes (approved):
I approve this PR blindly :)

PR #7540, opened 2018-05-17
gitlab windows: do not make artifacts

SkySkimmer:
- the runners are insecure
- (the real reason) I want to see if it's significantly faster


SkySkimmer:
40 min https://gitlab.com/coq/coq/-/jobs/68924719 (this)
vs
75 min https://gitlab.com/coq/coq/-/jobs/68870312 (master)

We seem to have enough windows runners so I'll close but let's remember the results if that changes.

PR #7550, opened 2018-05-18
[CI] Fix the script used by math-classes.

ppedrot:
We call configure to properly regenerate the Makefile and its dependencies.

The current build script of math-classes is somewhat strange as it has a hardcoded makefile that is never regenerated. This is troublesome if one wishes to add files to the library, or switch Coq versions.

This is required for the CI of #7536 to succeed.


ejgallego (approved):
Thanks.

PR #7557, opened 2018-05-19
Add test cases from #7554

### LINKED ISSUE #7554: Anomalies with Program
This is with Coq 8.7.2.
```
Require Import Coq.Program.Tactics.
Program Lemma foo: forall P, forall H, forall (n:nat), P n.
```
yields
```
Ltac call to "program_simpl" failed.
Ltac call to "program_simpl" failed.
Anomaly "Uncaught exception Not_found." Please report at http://coq.inria.fr/bugs/.
```

and
```
Require Import Coq.Program.Tactics.
Program Lemma foo: forall a (P : list a -> Prop), forall H, forall (n:list a), P n.
```
yields
```
Anomaly "Evar ?X1 was not declared." Please report at http://coq.inria.fr/bugs/.
```
but note that 
```
Require Import Coq.Program.Tactics.
Program Lemma foo: forall (a : Type) (P : list a -> Prop), forall H, forall (n:list a), P n.
```
yields
```
The statement obligations could not be resolved automatically, write a statement definition first.
```

(I am not blocked on this, I just found it by accident and report it as told :-))

tchajed:
Failed in v8.7.2 but were fixed by v8.8.0.

Closes #7554.

### DIFF
--- a/test-suite/bugs/closed/7554.v
+++ b/test-suite/bugs/closed/7554.v
@@ -1,0 +1,12 @@
+Require Import Coq.Program.Tactics.
+
+(* these should not result in anomalies *)
+
+Fail Program Lemma foo:
+  forall P, forall H, forall (n:nat), P n.
+
+Fail Program Lemma foo:
+  forall a (P : list a -> Prop), forall H, forall (n:list a), P n.
+
+Fail Program Lemma foo:
+  forall (a : Type) (P : list a -> Prop), forall H, forall (n:list a), P n.


test-suite/bugs/closed/gh7554.v
```diff
@@ -0,0 +1,5 @@
+Require Import Coq.Program.Tactics.
+
+Fail Program Lemma foo: forall (a : Type) (P : list a -> Prop), forall H, forall (n:list a), P n.
```
Zimmi48: This one never generated anomalies. Did you mean to put:

```
Program Lemma foo: forall P, forall H, forall (n:nat), P n.
```

instead?

Zimmi48 (changes requested):
Thank you for adding these tests.
Can you remove the `gh` prefix from the filename? This was something that I had proposed when I didn't think we would be able to migrate existing bugs from Bugzilla but given that we finally managed to, we didn't resort to using this prefix.

test-suite/bugs/closed/gh7554.v
tchajed: Yes, good catch. I fixed this by putting all three examples.

Zimmi48 (approved):

PR #6171, opened 2017-11-16
[proof] Removal of global proof state.

ejgallego:
Currently, proofs are stored in a global variable in `Proof_global`,
which is then kept in sync with the `Stm` notion of proof for each
particular state.

In this commit, we make some partial progress towards removal of such
global state in favor of a functional handling of the proof state. For
now, we deprecate functions accessing the global proof state to get an
idea of the remaining work.

Completing this work seems doable for 8.8, however the question on
compatibility should be addressed. There are 3 possible choices:

- Deprecate all the interfaces, but remove the global proof state in
  8.9 . This implies having Coq use some of the bad interfaces itself.

- Hard removal of the impure interfaces.

- Removal of the global proof state, keep the deprecated interfaces
  and install a private compatibility hook for the STM to update the
  proof state while the transition happens.

I think the best option is 2, even if a bit painful. 1 has a big risk
of delaying work and so the work may never happen. 3 is a bit of a
hack and IMO the advantages are not so clear.

EDIT: this depends on #7196.

PR #7558, opened 2018-05-20
[api] Make `vernac/` self-contained.

ejgallego:
We make the vernacular implementation self-contained in the `vernac/`
directory. To this extent we relocate the parser, printer, and AST to
the `vernac/` directory, and move a couple of hint-related types to
`Hints`, where they do indeed belong.

IMO this makes the code easier to understand, and provides a better
modularity of the codebase as now all things under `tactics` have 0
knowledge about vernaculars.

The vernacular extension machinery has also been moved to `vernac/`,
this will help when #6171 [proof state cleanup] is completed along
with a stronger typing for vernacular interpretation that can
distinguish different types of effects vernacular commands can perform.

This PR introduces some very minor source-level incompatibilities due
to a different module layering [thus deprecating is not
possible]. Impact should be relatively minor.

Overlays:
- https://github.com/Mtac2/Mtac2/pull/90
- https://github.com/ppedrot/ltac2/pull/60
- https://github.com/mattam82/Coq-Equations/pull/94


ejgallego:
Adjusted reviewers based on code owners, `dev/base_include/top_printers` seem to be orphan, should we open a bug?

ejgallego:
For those wondering about the motivation of the vernac refactorings, it is mainly due to wanting to improve a bit the interfaces exported by `Vernacinterp` and friends so document managers have an easier time to implement incremental compilation and parallel processing.

ejgallego:
Most of the patch is just `PCoq -> PVernac` renaming. Pity we don't have packing yet, but I understand that Dune will provide facilitates in the future that would allow one to extend modules, so for example `PCoq` could have been extended by `vernac` so the users above wouldn't need to change it.

ppedrot (approved):
LGTM.

maximedenes:
I can take this PR as the vernac secondary maintainer.

maximedenes (approved):
Looks good to me as well. This PR contains one unrelated, but very minor change.

maximedenes:
@herbelin anything to add?

maximedenes:
If no objection, I will merge tomorrow.

PR #7559, opened 2018-05-20
Existing Class noop when already a class + warning.

### LINKED ISSUE #5012: [Existing Class] breaks [Instance := { }] (incorrect error message "Unbound method name foo of class Foo")

Original bug ID: BZ#5012
From: @JasonGross
Reported version: 8.6

SkySkimmer:
Fix #5012.

### DIFF
--- a/vernac/record.ml
+++ b/vernac/record.ml
@@ -562,12 +562,18 @@ let add_inductive_class ind =
 	cl_unique = !typeclasses_unique }
   in add_class k
 
+let warn_already_existing_class =
+  CWarnings.create ~name:"already-existing-class" ~category:"automation" Pp.(fun g ->
+      Printer.pr_global g ++ str " is already declared as a typeclass.")
+
 let declare_existing_class g =
-  match g with
-  | ConstRef x -> add_constant_class x
-  | IndRef x -> add_inductive_class x
-  | _ -> user_err ~hdr:"declare_existing_class"
-		      (Pp.str"Unsupported class type, only constants and inductives are allowed")
+  if Typeclasses.is_class g then warn_already_existing_class g
+  else
+    match g with
+    | ConstRef x -> add_constant_class x
+    | IndRef x -> add_inductive_class x
+    | _ -> user_err ~hdr:"declare_existing_class"
+             (Pp.str"Unsupported class type, only constants and inductives are allowed")
 
 open Vernacexpr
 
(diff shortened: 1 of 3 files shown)


vernac/record.ml
```diff
@@ -562,12 +562,18 @@ let add_inductive_class ind =
 	cl_unique = !typeclasses_unique }
   in add_class k
 
+let warn_already_existing_class =
+  CWarnings.create ~name:"already-existing-class" ~category:"automation" Pp.(fun g ->
+      Printer.pr_global g ++ str " already declared as a typeclass.")
```
JasonGross: Should there be an "is" to make this a grammatical sentence?

JasonGross (approved):
Thanks! I think this error should be documented in the refman. (But I'll approve this as-is because it's already better than the current situation.)

SkySkimmer:
Added verb to the message.
>Thanks! I think this error should be documented in the refman. (But I'll approve this as-is because it's already better than the current situation.)

It's a warning, not an error. Do we document those?

JasonGross:
24ththink we should document warnings.  For example, the warning about the uniform inheritance condition should have documentation explaining that the coercion isn't actually added, and there should be documentation explaining the warning message about ltac identifier name clashes.  This warning message doesn't seem particularly important to explain further in the docs (and I mentioned it because I thought we did, in fact, list warnings along with errors, though I realize now we might not), but I think it's worth working towards a documentation state where all errors and warnings are at least listed in the doc.

Zimmi48:
> It's a warning, not an error. Do we document those?

We are supposed to. We have the `.. warn::` directive for that purpose. Warnings are listed with errors in the "Errors and Warnings Index". I realize though that I just introduced a new error without thinking of documenting it.

JasonGross:
> I realize though that I just introduced a new error without thinking of documenting it.

Perhaps the PR template "documentation" bullet should include "(including documenting any errors or warnings modified)"?

ejgallego:
> Perhaps the PR template "documentation" bullet should include "(including documenting any errors or warnings modified)"?

We should do this automatically.

Zimmi48:
> We should generate the documentation automatically from the error registration table instead of relying on a manual process.

Ideally yes, but until this is the case, adapting to template to avoid people to forget is not a bad idea. @JasonGross would you mind opening a PR to propose this change?

SkySkimmer:
Can someone take charge of this PR?

maximedenes (approved):
Doc looks ok to me. I think @mattam82 should be the assignee.

SkySkimmer:
@mattam82 ping

mattam82 (approved):
Good catch, the warning addition looks fine to me.

PR #7561, opened 2018-05-20
[api] [parsing] Move Egram* to `vernac/`

ejgallego:
The extension mechanism is specific to metasyntax and vernacinterp,
thus it makes sense to place them next to each other.


META.coq
```diff
@@ -6,6 +6,18 @@ version = "8.8"
 directory = ""
 requires = "camlp5"
 
+package "grammar" (
```
SkySkimmer: Why's this moving?

ejgallego: It is not just a move, see the diff, the old META entry for grammar was just wrong, `grammar` is just a plugin for `camlp5` [that we build outside the main coq tree] and thus doesn't belong in the main META dependency tree. In fact, in the Dune build we don't provide a grammar library and dynlink it, but we do distribute the whole driver [which is way faster for a start]

ejgallego: In fact in my main working tree the parsing directory is gone altogether, hard to show the full thing tho, merging of this stuff is like, super slow, and if I stack people do complain.

Zimmi48: You could give a link to your main branch though so that people can have a look at the whole thing without feeling a pressing need to review.

herbelin:
> The extension mechanism is specific to metasyntax and vernacinterp,
thus it makes sense to place them next to each other.

Is this for technical reasons (e.g. simplifying dependencies) or for conceptual reasons?

I'm asking because I got used to think at `parsing` as low-level (connecting Camlp5 to Coq) and `vernac` as higher level (connecting the user to  Coq), thus with low-level `Egramcoq` in `parsing` but you are apparently defending a view in which low vs high-level is maybe less important than a feature-based separation.

But then, why not e.g. to move `metasyntax.ml` in `parsing` instead?

ejgallego:
> Is this for technical reasons (e.g. simplifying dependencies) or for conceptual reasons?
> 
> I'm asking because I got used to think at parsing as low-level (connecting Camlp5 to Coq) and vernac as higher level (connecting the user to Coq), thus with low-level Egramcoq in parsing but you are apparently defending a view in which low vs high-level is maybe less important than a feature-based separation.
> 
> But then, why not e.g. to move metasyntax.ml in parsing instead?

I guess it is a quite subjective choice, I am trying to pack together things that are coupled. In #7558 I took the choice of making the system more modular thus I put the parsing part for the vernacular in that module.

But indeed there are other views, I guess we can discuss in person soon!

ejgallego:
@herbelin thinking twice about it, this PR by itself doesn't make a lot of sense, closing in favor of #7558 , which includes this commit but in a better context.

PR #6857, opened 2018-02-28
[build] Preliminary support for building with `dune`.

ejgallego:
[Dune](https://github.com/ocaml/dune) is a compositional declarative
build system for OCaml. It provides automatic generation of
`version.ml`, `.merlin`, `META`, `opam`, API documentation; install
management; easy integration with external libraries, test runners,
and modular builds.

In particular, Dune uniformly handles components regardless whether
they live in, or out-of-tree. This greatly simplifies cases where a
plugin [or CoqIde] is checked out in the current working copy but then
distributed separately [and vice-versa]. Dune can thus be used as a
more flexible `coq_makefile` replacement.

For now we provide experimental support for a Dune build. In order to
build Coq + the standard library with Dune type:

```
$ make -f Makefile.dune world
```

This PR includes a preliminary, developer-only preview of Dune for
Coq. There is still ongoing work, see
https://github.com/coq/coq/issues/8052 for tracking status towards
full support.

## Technical description.

Dune works out of the box with Coq, once we have fixed some modularity
issues. The main remaining challenge was to support `.vo` files.

As Dune doesn't support custom build rules yet, to properly build
`.vo` files we provide a small helper script `tools/coq_dune.ml`. The
script will scan the Coq library directories and generate the
corresponding rules for `.v -> .vo` and `.ml4 -> .ml` builds. The
script uses `coqdep` as to correctly output the dependencies of
`.v` files. `coq_dune` is akin to `coq_makefile` and should be able to
be used to build Coq projects in the future.

Due to this pitfall, the build process has to proceed in three stages: 1) build `coqdep` and `coq_dune`; 2) generate `dune` files for
`theories` and `plugins`; 3) perform a regular build with all
targets are in scope.

### Why Dune?

Coq has a moderately complex build system and it is not a secret that
many developer-hours have been spent fighting with `make`.

In particular, the current `make`-based system does offer poor support
to verify that the current build rules and variables are coherent, and
requires significant manual, error-prone. Many variables must be
passed by hand, duplicated, etc... Additionally, our make system
offers poor integration with now standard OCaml ecosystem tools such
as `opam`, `ocamlfind` or `odoc`. Another critical point is build
compositionality. Coq is rich in 3rd party contributions, and a big
shortcoming of the current make system is that it cannot be used to
build these projects; requiring us to maintain a custom tool,
`coq_makefile`, with the corresponding cost.

In the past, there has been some efforts to migrate Coq to more
specialized build systems, however these stalled due to a variety of
[...]

PR #7563, opened 2018-05-21
[ide] Use Dune to support separate CoqIDE build.

ejgallego:
Packagers have long wanted to build CoqIDE against the installed Coq
libraries instead of having to rebuild parts of Coq itself.

In fact, this is indeed possible since since some time, as for example
other topleves in OPAM such as `coq-serapi` do.

We thus provide the needed `dune` files so CoqIDE can be built
separately by issuing `jbuilder build -p coqide`.

This can be backported to 8.8.1, but IMO protocol files should be
moved to its own directory as done in `master`; I don't see a lot of
risk on it tho.

Note that as of today, the resulting files have to be installed in the
same place where the `coqworker*` binaries are installed.  This could
be improved by tweaking the search heuristic of coqide / coqlib.

The Dune-generated `coqide.install` file does the right thing in that
regard.

### DIFF
--- a/ide/coqide.opam
+++ b/ide/coqide.opam
@@ -1,0 +1,20 @@
+opam-version: "1.2"
+maintainer:  "The Coq Development Team <coqdev@sympa.inria.fr>"
+authors:     "The Coq Development Team"
+homepage:    "https://coq.inria.fr/"
+bug-reports: "https://github.com/coq/coq/issues"
+dev-repo:    "https://github.com/coq/coq.git"
+license:     "LGPL-2.1"
+
+name: "coqide"
+available: [ocaml-version >= "4.02.3"]
+
+depends: [
+  "coq"       { >= "8.9" & < "8.10" }
+  "camlp5"
+  "jbuild"    { build }
+  "ocamlfind" { build }
+  "num"
+]
+
+build: [["jbuilder" "build" "-p" name "-j" jobs]]
--- a/ide/jbuild
+++ b/ide/jbuild
@@ -1,0 +1,27 @@
+(jbuild_version 1)
+
+(executable
+ ((name idetop)
+  (public_name coqidetop.opt)
+  (package coqide)
+  (modules (idetop))
+  (libraries (coq.toplevel coqide.protocol))
+  (flags -rectypes)
+  (link_flags -linkall)))
+
+(rule
+ ((targets (coqide_main.ml))
+  (deps (coqide_main.ml4))
+  ; XXX: when Coq is built with dune the following line will do the job:
+  ; (action (run coqmlp5 -loc loc -impl ${<} -o ${@}))))
+  (action (system "camlp5o pa_extend.cmo q_MLast.cmo pa_macro.cmo pr_o.cmo -loc loc -impl ${<} -o ${@}"))))
+
+(executable
+ ((name coqide_main)
+  (public_name coqide)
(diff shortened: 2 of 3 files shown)


ide/_tags
```diff
@@ -0,0 +1,16 @@
...
+
+<protocol>: include
+
+# Language server
+<idetop.*>: package(coq.toplevel)
+<idetop.native>: linkall
+
+# CoqIDE itself
+not <idetop.ml>: package(lablgtk2.sourceview2)
+<coqide_main.ml>:  syntax(camlp4o), package(camlp4.macro)
```
gares: Why camlp4 and not camlp5?

gares:
Given the current trend, I'd prefer to see a dune file to build coqide rather than an ocamlbuild one. Why did you chose ocamlbuild here?

ide/_tags
ejgallego: That's a good question, it likely comes from the time the legacy repo was created.

ejgallego:
> Given the current trend, I'd prefer to see a dune file to build coqide rather than an ocamlbuild one. Why did you chose ocamlbuild here?

Oh, that may work, let me try; there is a couple of catches:
- we may need to provide an OPAM file too,
- when building Coq with Dune `kernel.cmxa` will store some necessary linking flags, so the linking is a bit different if Coq was built with Dune or if it was built with without. We can live with that but it's a bit annoying. We could also teach `make` to properly set the flags on kernel but I am not sure if that's gonna be easy.

ejgallego:
Ok, Dune files added from #6857 , it works here:
- built Coq with `-coqide no -prefix foo`
- built coqide with `jbuilder build -p coqide`
- `opam-install` does the rest.

gares:
Do you want to change
  https://github.com/coq/coq/blob/master/Makefile.build#L571
to pass `-cclib -lcoqrun` ?
And add another specific rule for `kernel.cmxa`?

Is this what you talk about in your comments?

ejgallego:
> Do you want to change
> https://github.com/coq/coq/blob/master/Makefile.build#L571
> to pass -cclib -lcoqrun ?
> And add another specific rule for kernel.cmxa?
> 
> Is this what you talk about in your comments?

Something like that, I was no aware we were already doing that for the `cma`. The thing is that I am afraid to touch these stuff as things tend to break in windows / OSX, etc...

I don't fully understand these details, but Dune indeed adds some magic like that when the stanza contains `cfiles` https://github.com/coq/coq/pull/6857/files#diff-9b787ba8a386f7571be136e18f455203R7

So I think you are right and the fact the we don't do this for `kernel.cmxa`  should be considered "a bug".

ejgallego:
Indeed #7575  works, so I assume you want to keep this PR depending on it then.

gares (approved):

ejgallego:
Ready to go.

ppedrot (commented):
If people knowledgeable of build systems assess that this is OK, I'll blindly trust them.

ejgallego:
Apart from the tweak due to a merge problem, the rest is just adding an extra file that has no effect on the main part, it is interesting for packagers tho.

ejgallego:
CI failures are spurious.

ejgallego:
This didn't gather enough interest I guess and anyways #6857 should be merged before 8.9 so this won't be necessary.

PR #7564, opened 2018-05-21
Move interning the [hint_pattern] outside the Typeclasses hooks.

### LINKED ISSUE #7562: Typeclasses pretyping phase depends on uninterpreted term AST + Hooks.
In `pretyping/typeclasses` we can find:

```
type hint_info_expr = Constrexpr.constr_expr hint_info_gen
val add_instance_hint_hook :
  (global_reference_or_constr -> GlobRef.t list ->
   bool (* local? *) -> hint_info_expr -> Decl_kinds.polymorphic -> unit) Hook.t
```

where the hook is set in `vernac/classes.ml` and does the interpretation of the "pattern". I think that something fishy is going on here; `pretyping` is considered below `interp` and indeed it shouldn't depend on uninterpreted ASTs.

I tried to give it a go, but the code flow is a bit more complex of what I can take in these moments.

IMO, this should be fixed and the hooks done away with.

Coincidentally, `Typeclasses` is basically the only important user of the dreaded `Hook.t` in the codebase.

SkySkimmer:
Close #7562.
The hooks are more about interaction with tactics/hints.ml so I don't see them going away.

### DIFF
--- a/pretyping/typeclasses.ml
+++ b/pretyping/typeclasses.ml
@@ -30,7 +30,7 @@ type 'a hint_info_gen =
     { hint_priority : int option;
       hint_pattern : 'a option }
 
-type hint_info_expr = Constrexpr.constr_pattern_expr hint_info_gen
+type hint_info = (Misctypes.patvar list * Pattern.constr_pattern) hint_info_gen
 
 let typeclasses_unique_solutions = ref false
 let set_typeclasses_unique_solutions d = (:=) typeclasses_unique_solutions d
@@ -80,7 +80,7 @@ type typeclass = {
   cl_props : Context.Rel.t;
 
   (* The method implementaions as projections. *)
-  cl_projs : (Name.t * (direction * hint_info_expr) option
+  cl_projs : (Name.t * (direction * hint_info) option
 	      * Constant.t option) list;
   
   cl_strict : bool;
@@ -92,7 +92,7 @@ type typeclasses = typeclass Refmap.t
 
 type instance = {
   is_class: GlobRef.t;
-  is_info: hint_info_expr;
+  is_info: hint_info;
   (* Sections where the instance should be redeclared,
      None for discard, Some 0 for none. *)
   is_global: int option;
--- a/pretyping/typeclasses.mli
+++ b/pretyping/typeclasses.mli
@@ -21,7 +21,7 @@ type 'a hint_info_gen =
     { hint_priority : int option;
       hint_pattern : 'a option }
 
-type hint_info_expr = Constrexpr.constr_pattern_expr hint_info_gen
+type hint_info = (Misctypes.patvar list * Pattern.constr_pattern) hint_info_gen
 
 (** This module defines type-classes *)
 type typeclass = {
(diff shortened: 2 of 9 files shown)


ppedrot (commented):
I'm slightly afraid that `constr_pattern` might contain data whose serialization is somewhat broken. We store `hint_info` in the summary, so better double check.

SkySkimmer:
I think we already store constr_pattern for hints: https://github.com/coq/coq/blob/d6eb4a26648817f6b034e95c02622cadf0fa65ca/tactics/hints.ml#L147

ppedrot (approved):
Cannot be wronger than what we currently do, so good to me.

SkySkimmer:
Should there be a subst/discharge for it though?

ejgallego:
> The hooks are more about interaction with tactics/hints.ml so I don't see them going away.

Interesting, however they make using the `pretyper` alone fail in a few cases; maybe a "Hook" is not the proper registration mechanism?

Sorry if I say stupid things as it is taking me some effort to get familiar with the code.

ejgallego:
> they make using the pretyper alone fail in a few cases

[Yes, I know a user that is trying to use only from pretyping to below, he has to link a few extra things to make it work]

ejgallego:
Thanks for having a look.

ejgallego:
I am not sure tho we should fully close that bug, it would be good to have a reminder to look a the hooks at some point.

I yet have not found a Hook that was not masking suspicious activity.

pretyping/typeclasses.ml
```diff
@@ -31,6 +31,7 @@ type 'a hint_info_gen =
       hint_pattern : 'a option }
 
 type hint_info_expr = Constrexpr.constr_pattern_expr hint_info_gen
```
ejgallego: I think `hint_info_expr` should be moved to `tactics/hints` then.

ejgallego (dismissed):

pretyping/typeclasses.ml
SkySkimmer: >I think hint_info_expr should be moved to tactics/hints then.

You did that in some commit in #7570 but it seems like it depends on some previous stuff, what do?

ejgallego: I think I had checked that it could be done in the context of this PR, but I was likely wrong.

If my request is not possible please dismiss my review and merge.

SkySkimmer:
Who is in charge of this PR?
```
pretyping/typeclasses.mli: @mattam82
pretyping/typeclasses.ml: @mattam82
vernac/classes.ml: @mattam82
tactics/class_tactics.ml: @ppedrot
tactics/hints.mli: @ppedrot
tactics/hints.ml: @ppedrot
```

ppedrot:
That's a tie, so I can assign it to myself as @mattam82 is less responsive than I am.

SkySkimmer:
@ejgallego done

ejgallego:
Merci !

ejgallego (approved):
Looks good, thanks @SkySkimmer and @ppedrot

ejgallego:
@ppedrot may be busy so taking over this on the advice of @maximedenes

PR #7570, opened 2018-05-21
[api] Move `Constrexpr` to the `interp` module.

ejgallego:
Continuing the interface cleanup we place `Constrexpr` in the
internalization module, which is the one that eliminates it.

This slims down `pretyping` considerably, including removing the
`Univdecls` module which existed only due to bad dependency ordering
in the first place.

D̶e̶p̶e̶n̶d̶s̶ ̶o̶n̶ ̶#̶7̶5̶5̶8̶ ̶a̶n̶d̶ ̶#̶7̶5̶6̶4̶ ̶


interp/constrexpr.ml
```diff
@@ -17,6 +17,7 @@ open Decl_kinds
 
 (** [constr_expr] is the abstract syntax tree produced by the parser *)
 
+type universe_decl = (lident list, Univ.Constraint.t) gen_universe_decl
```
SkySkimmer: The earliest definition of this type is in ustate, for some reason it was simultaneously (f72a67569ec8cb9160d161699302b67919da5686) created in evd (later moved to ustate) and univdecls.

ejgallego: I think the reason this kind of stuff happened is that we had some subtle cyclic dependency mess in the system, but these should be gone by now and enforced once the Dune build is in the CI.

I think the situation is a bit better now, will try to make this type a deprecated alias.

ejgallego: Fixed, thanks for the tip.

ejgallego: There was indeed quite a few duplication.

interp/constrexpr.ml
```diff
@@ -17,7 +17,10 @@ open Decl_kinds
 
 (** [constr_expr] is the abstract syntax tree produced by the parser *)
 
-type universe_decl_expr = (lident list, Glob_term.glob_constraint list) gen_universe_decl
+type universe_decl = UState.universe_decl
+[@@ocaml.deprecated "Please use [UState.universe_decl]"]
```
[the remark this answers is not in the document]
ejgallego: Good catch, I tend to do that to help plugins but in this case there's no point.

interp/constrexpr_ops.mli
```diff
@@ -124,3 +124,10 @@ val error_invalid_pattern_notation : ?loc:Loc.t -> unit -> 'a
 
 (** Placeholder for global option, should be moved to a parameter *)
 val asymmetric_patterns : bool ref
+
+(** Local universe and constraint declarations. *)
+val interp_univ_decl : Environ.env -> universe_decl_expr ->
+                       Evd.evar_map * universe_decl
+
+val interp_univ_decl_opt : Environ.env -> universe_decl_expr option ->
```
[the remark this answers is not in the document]
ejgallego: The thing is that the type is only used in the AST tho, so I dunno, we could indeed put `universe_decl_expr` in the core pretyping AST `Glob_term` but it would be then unused there.

herbelin:
Question in passing: Can you explain why you need the lexer in `interp`?

ejgallego:
> Question in passing: Can you explain why you need the lexer in interp?

Hi Hugo, `Notation_terms` depends on `Extend`, which depends on `CLexer`.

I didn't look at optimizing that, it is likely possible to do better.

ejgallego:
> I didn't look at optimizing that, it is likely possible to do better.

Indeed, that can be improved by splitting `notation_term` in the parsing part and intepretation part.

ejgallego:
@herbelin , please see the latest commit for a WIP split of notation interpretation and parsing. It needs some tweaks on the organization of the stuff in `parsing/` , but you can see the idea.

Actually resolving this was a very good idea IMO.

ejgallego:
By the way @herbelin , would you like to shepherd this PR?

herbelin:
> By the way @herbelin , would you like to shepherd this PR?

I can formally be the assignee, but, to be honest, I lost track of all the changes going on, so, if someone else (in addition to you) has a better overview than me on the ongoing moves, I'll be happy to let another person "shepherd".

Among the points I'm lost about is the following question: what is the impact of the restructurations made in this PR from the point of view of a plugin developer using the API?

herbelin:
@Zimmi48: ok thanks.

ejgallego:
> Among the points I'm lost about is the following question: what is the impact of the restructurations made in this PR from the point of view of a plugin developer using the API?

This is a very good question. The main idea these these commit and the depended #7558 implement is the ability to see "coq module" as a more self-contained programming unit. So once this is in, we'd like to start #7155 , and the idea would be to add a page for each directory or "Coq unit".

So in particular, the overview for `coq.interp` would be: "The interp layer has as a mission to elaborate Coq's term abstract syntax into core AST that will be fed to type inference. This layer thus defines the input AST type `constr_expr`, as well as the interpretation functions [foo....] and the AST for notation patterns]"

For vernac, we would write something similar, as well as for pretyping.

So the main task of this PR is to move the parsing part out of `interp` and interpretation is done without knowledge of the parsing. Thus, a contributor wanting to work / understand `interp` doesn't have to even know of the existence of `camlp5`.

Zimmi48:
> So once this is in, we'd like to start #7155 , and the idea would be to add a page for each directory or "Coq unit".

How do you intend to organize this documentation? I would personally appreciate having almost one README per sub-directory (as I've started doing or encouraging in some places). The advantage would be that a new contributor can easily browse the GitHub repo and understand what each component is about. But other solutions are good to me to.

In any case, when you are happy with the state of a sub-component (for instance with this PR), what about adding an embryo of such documentation right away? This would also help the reviewers understand the purpose of the changes (I think) and help everyone see that this is not just pointless code moving with no clear target, like refactoring without a plan can sometimes be.

ejgallego:
> How do you intend to organize this documentation?

We will use `odoc` so indeed we will have an introduction to each directory, we could make the README link to it. And of course we will have an intro README.

I guess you can see `odoc` examples to get an idea. Also, `odoc` is under development these days so what the new version will do is not set in stone.

maximedenes:
@ppedrot Can you have a look at this?

SkySkimmer (approved):

mattam82 (approved):
Fine with me, I like the refactoring

maximedenes:
Since people seem ok with this PR, if no objection, I'll merge it later today.

ppedrot (approved):
I guess that's OK.

ejgallego:
Merci !

PR #6423, opened 2017-12-14
[rfc] [ocaml] Enable warning "27" unused variable.

ejgallego:
IMHO this warning is important to be on, specially in the upper layers
that are supposed to handle environments in "let style".

Lots of funny stuff in the diff.

Unfortunately CAMPLX does not understand `[@@@ocaml.warning "-27"]` so
this PR is not ready for merge as `ml4` files will trigger the warning
in abundance due to unused `loc` generated by parsing rules.

Suggestions on how to disable the warning for `ml4`-generated files
are welcome.

PR #7571, opened 2018-05-21
[kernel] Compile with almost all warnings enabled.

ejgallego:
This is a partial resurrection of #6423 but only for the kernel.

IMHO, we pay a bit of price for this but it is a good safety
measure.

Only warning "4: fragile pattern matching" and "44: open hides a type"
are disabled.

We would like to enable 44 for sure once we do some alias cleanup.

### DIFF
--- a/kernel/dune
+++ b/kernel/dune
@@ -13,3 +13,8 @@
 
 (documentation
  (package coq))
+
+; In dev profile, we check the kernel against a more strict set of
+; warnings.
+(env
+ (dev (flags :standard -w +a-4-44-50)))
--- a/kernel/nativelib.ml
+++ b/kernel/nativelib.ml
@@ -40,7 +40,7 @@ let include_dirs () =
   [Filename.get_temp_dir_name (); coqlib () / "kernel"; coqlib () / "library"]
 
 (* Pointer to the function linking an ML object into coq's toplevel *)
-let load_obj = ref (fun x -> () : string -> unit)
+let load_obj = ref (fun _x -> () : string -> unit)
 
 let rt1 = ref (dummy_value ())
 let rt2 = ref (dummy_value ())
@@ -113,7 +113,7 @@ let call_compiler ?profile:(profile=false) ml_filename =
     let res = CUnix.sys_command (ocamlfind ()) args in
     let res = match res with
       | Unix.WEXITED 0 -> true
-      | Unix.WEXITED n | Unix.WSIGNALED n | Unix.WSTOPPED n ->
+      | Unix.WEXITED _n | Unix.WSIGNALED _n | Unix.WSTOPPED _n ->
         warn_native_compiler_failed (Inl res); false
     in
     res, link_filename
@@ -158,7 +158,7 @@ let call_linker ?(fatal=true) prefix f upds =
   (try
     if Dynlink.is_native then Dynlink.loadfile f else !load_obj f;
     register_native_file prefix
-   with Dynlink.Error e as exn ->
+   with Dynlink.Error _ as exn ->
      let exn = CErrors.push exn in
      if fatal then iraise exn
      else if !Flags.debug then Feedback.msg_debug CErrors.(iprint exn));
--- a/kernel/subtyping.ml
+++ b/kernel/subtyping.ml
@@ -138,7 +138,7 @@ let check_inductive cst env mp1 l info1 mp2 mib2 spec2 subst1 subst2 reso1 reso2
   let mib2 =  Declareops.subst_mind_body subst2 mib2 in
   let check_inductive_type cst name t1 t2 =
     check_conv (NotConvertibleInductiveField name)
-      cst (inductive_is_polymorphic mib1) infer_conv_leq env t1 t2
+      cst (inductive_is_polymorphic mib1) (infer_conv_leq ?l2r:None ?evars:None ?ts:None) env t1 t2
   in
 
   let check_packet cst p1 p2 =
@@ -162,10 +162,10 @@ let check_inductive cst env mp1 l info1 mp2 mib2 spec2 subst1 subst2 reso1 reso2
 	cst
   in
   let mind = MutInd.make1 kn1 in
-  let check_cons_types i cst p1 p2 =
+  let check_cons_types _i cst p1 p2 =
     Array.fold_left3
       (fun cst id t1 t2 -> check_conv (NotConvertibleConstructorField id) cst
-	(inductive_is_polymorphic mib1) infer_conv env t1 t2)
+        (inductive_is_polymorphic mib1) (infer_conv ?l2r:None ?evars:None ?ts:None) env t1 t2)
       cst
       p2.mind_consnames
       (arities_of_specif (mind, inst) (mib1, p1))
@@ -229,7 +229,7 @@ let check_constant cst env l info1 cb2 spec2 subst1 subst2 =
   let check_conv cst poly f = check_conv_error error cst poly f in
   let check_type poly cst env t1 t2 =
     let err = NotConvertibleTypeField (env, t1, t2) in
-    check_conv err cst poly infer_conv_leq env t1 t2
+    check_conv err cst poly (infer_conv_leq ?l2r:None ?evars:None ?ts:None) env t1 t2
   in
   match info1 with
     | Constant cb1 ->
@@ -268,14 +268,14 @@ let check_constant cst env l info1 cb2 spec2 subst1 subst2 =
 		 Anyway [check_conv] will handle that afterwards. *)
 	      let c1 = Mod_subst.force_constr lc1 in
 	      let c2 = Mod_subst.force_constr lc2 in
-	      check_conv NotConvertibleBodyField cst poly infer_conv env c1 c2))
-   | IndType ((kn,i),mind1) ->
+              check_conv NotConvertibleBodyField cst poly (infer_conv ?l2r:None ?evars:None ?ts:None) env c1 c2))
+   | IndType ((_kn,_i),_mind1) ->
        CErrors.user_err Pp.(str @@
        "The kernel does not recognize yet that a parameter can be " ^
        "instantiated by an inductive type. Hint: you can rename the " ^
        "inductive type and give a definition to map the old name to the new " ^
        "name.")
-   | IndConstr (((kn,i),j),mind1) ->
+   | IndConstr (((_kn,_i),_j),_mind1) ->
       CErrors.user_err Pp.(str @@
        "The kernel does not recognize yet that a parameter can be " ^
        "instantiated by a constructor. Hint: you can rename the " ^
--- a/kernel/term.ml
+++ b/kernel/term.ml
@@ -54,13 +54,13 @@ let mkProd_wo_LetIn decl c =
   let open Context.Rel.Declaration in
   match decl with
   | LocalAssum (na,t) -> mkProd (na, t, c)
-  | LocalDef (na,b,t) -> subst1 b c
+  | LocalDef (_na,b,_t) -> subst1 b c
 
 let mkNamedProd_wo_LetIn decl c =
   let open Context.Named.Declaration in
   match decl with
     | LocalAssum (id,t) -> mkNamedProd id t c
-    | LocalDef (id,b,t) -> subst1 b (subst_var id c)
+    | LocalDef (id,b,_t) -> subst1 b (subst_var id c)
 
 (* non-dependent product t1 -> t2 *)
 let mkArrow t1 t2 = mkProd (Anonymous, t1, t2)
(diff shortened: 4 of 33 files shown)


Zimmi48:
This last commit only touches the kernel so the number of required reviews didn't make any sense. This was only due to the dependency on the Dune PR. I don't see any conflict happening with any of the previous PRs you based this on so I don't understand why you chose to stack in that case. This is only going to delay the merge of this very PR.

ejgallego:
> This last commit only touches the kernel so the number of required reviews didn't make any sense. This was only due to the dependency on the Dune PR. I don't see any conflict happening with any of the previous PRs you based this on so I don't understand why you chose to stack in that case. This is only going to delay the merge of this very PR.

This is just stacked on the Dune PR, that's the one that depends on all the rest of PRs, the main reason being that the Dune commit itself is not ready for merge yet [but essential to develop some other work] but the dependencies are, so they stack like this.

But yeah, I have refrained to submit my `misctypes+bye` branch as the stacking is extreme. Pity we cannot use of any development momentum.

Zimmi48:
But why does this require the Dune commit?! What is the connection between the two?

ejgallego:
> But why does this require the Dune commit?! What is the connection between the two?

I use Dune to implement the per-module warning.

Zimmi48:
OK, but once you had implemented it, couldn't you make it independent? I don't mean to annoy you but only to help here with merging this quicker ;-)

ejgallego:
> OK, but once you had implemented it, couldn't you make it independent? I don't mean to annoy you but only to help here with merging this quicker ;-)

This depends on #6857 I think, so will have to wait.

ejgallego:
> This depends on #6857 I think, so will have to wait.

But I think we can get all the pieces ready pretty quickly; even if the commit chain looks large all the changes are trivial, the hard one was #6859 

On the other hand we could do as in #7563 and spin off the kernel part to do a test here, indeed maybe that's a good idea.

ejgallego:
I guess that's a matter of style; I always start from al setup where all warnings are on. Unused variables seem pretty dangerous for kernel code tho.

maximedenes (dismissed):
Please ping once dependencies have been merged.

ejgallego (commented):
CI is OK.

SkySkimmer (commented):
Test to see what happens to the review request.

ejgallego:
@maximedenes this is ready to review.

kernel/dune
```diff
@@ -4,6 +4,7 @@
  (public_name coq.kernel)
  (wrapped false)
  (modules_without_implementation cinstr nativeinstr)
+ (flags :standard -w +a-4-44-50)
```
ejgallego: this should be likely done in an `(env ....)` declaration.

maximedenes:
> [SkySkimmer](/SkySkimmer) self-requested a review

@SkySkimmer you recently self-requested a review, do you want to have a look at it?

SkySkimmer:
I guess my comment from back when still stands:
>I'm not convinced that enabling warning `27: innocuous unused variable` is a good idea.
`44: "open" shadowing already defined identifier` seems like it could get seriously annoying when moving some function and leaving a deprecated version at the original site.

@ejgallego said
>I always start from al setup where all warnings are on. 

Coq is not a new project so I don't see how this is relevant.
>Unused variables seem pretty dangerous for kernel code tho.

The ones that get touched in this patch seem all fine though.
I don't think the cost of dealing with the pointless warnings is worth it.

maximedenes (dismissed):
I don't like that we enable some warnings that we need to disable locally (like 41 and 45). Better not activate them, or fix the code to not trigger them.

Also, are the changes of the form `{x ; y}` -> `{x; y; _}` really an improvement? Less sure about this one.

SkySkimmer (commented):
In the end I'm willing to try it out in practice. I agree with @maximedenes that the locally disabled warnings would be better fixed.
At least a comment of why it's disabled for those of us who don't know the warning codes by heart ;)

gares:
> Also, are the changes of the form `{x ; y}` -> `{x; y; _}` really an improvement? Less sure about this one.

According to my experience, records can be used in two ways:
1. to name structured data
2. to enable easy projection

For 1, the warning is really important. Eg I use it in AST in a compilation phase, when I add a field I must do something at compile time hence having the compiler signal it is relevant. In this case it is really like a tuple, but with names.

For 2, the warning is silly. The whole point is to let one write `{ foo }`. Eg when implementing a set of flags, by definition you don't read all of them in one go. Using `x.foo` is less explicit: you don't immediately see that a piece of code is not using the other fields, you have to read it all.

I believe the warning should be attached to a data type, not be on/off globally. It should be the designer of the data type to signal its meaning. But this message is out of place here, unless @gasche reads it ;-)

I'm not in favor of enabling 9 globally.


But more importantly I'm not in favor of retrofitting warning and fixing the code like that.
I see code like `| err -> ...` turned into `| _err -> ...`. Either there is a bug, and `err` shall be used, or there is no bug and the code shall be changed to make the pattern exhaustive. Or `  let (_, _, labs, pbs, letsubst) =` becoming `  let (_, _, labs, pbs, _letsubst) =`: again if the last component should not be used one would put a `_` there, not `_letsubst` given that the other two discarded components are not named.

ejgallego:
> I'm not in favor of enabling 9 globally.

Note this is just for the kernel. @gares, of your 2 use cases, which one do you think it is the prevalent one in `kernel  ?

gares:
`cClosure` is using `info` as a set of flags, as far as I can tell.
`Conv_oracle` is using `oracle` as a tuple, even if it looks like it should be 2 data structures, one for the strategy and one for opacity (if it was so you would not have needed to add ;_ in set/get_strategy since the data being dropped is the one about the opacity that should have not been passed AFAICT).
`Opauqeproofs` should define `opaquetab` as an opaque data and really hide the `int` your drop with `; _` since there is an invariant linking two fields of the record (eg all `set_whatever` must use and update the `int`, looks like they do).
I don't know much about `UGraph` or `modops`.
For `Term_typing` I don't know why seff is like that today, while constant_whatever seems to be a data type that collects many infos one never really needs at once, hence the inprecise pattern matching on the record seems legit.

gares:
> Note this is just for the kernel.

This is why I'm being so picky. If these warnings reveal a bug we should fix it, while the current patch seems to fix the warning under the assumption the code is correct.

ejgallego:
> if the last component should not be used one would put a `_` there, not `_letsubst` given that the other two discarded components are not named.

This is a matter of style I guess; I like naming arguments as it makes the code more clear, and here we are talking about supposed safety-critical code. So I'd say that rather the parameters where `_` is used should be fixed.

kernel/nativelib.ml
```diff
@@ -158,7 +158,7 @@ let call_linker ?(fatal=true) prefix f upds =
   (try
     if Dynlink.is_native then Dynlink.loadfile f else !load_obj f;
     register_native_file prefix
-   with Dynlink.Error e as exn ->
+   with Dynlink.Error _e as exn ->
```
gasche: I would just use `_` here as the name is not informative.

kernel/subtyping.ml
```diff
@@ -138,7 +138,7 @@ let check_inductive cst env mp1 l info1 mp2 mib2 spec2 subst1 subst2 reso1 reso2
   let mib2 =  Declareops.subst_mind_body subst2 mib2 in
   let check_inductive_type cst name t1 t2 =
     check_conv (NotConvertibleInductiveField name)
-      cst (inductive_is_polymorphic mib1) infer_conv_leq env t1 t2
+      cst (inductive_is_polymorphic mib1) (infer_conv_leq ?l2r:None ?evars:None ?ts:None) env t1 t2
```
gasche: Maybe give a local name to this specialization of `infer_conv_leq` in this module to avoid repeating it several times?

kernel/term.ml
```diff
@@ -276,7 +276,7 @@ let decompose_prod_n_assum n =
       | Prod (x,t,c)    -> prodec_rec (Context.Rel.add (LocalAssum (x,t)) l) (n-1) c
       | LetIn (x,b,t,c) -> prodec_rec (Context.Rel.add (LocalDef (x,b,t)) l) (n-1) c
       | Cast (c,_,_)      -> prodec_rec l n c
-      | c -> user_err (str  "decompose_prod_n_assum: not enough assumptions")
+      | _c -> user_err (str  "decompose_prod_n_assum: not enough assumptions")
```
gasche: I would just use `_` here and in others case where the whole pattern is concerned and the name was not informative in the first place.

kernel/vmvalues.ml
```diff
@@ -333,6 +336,7 @@ struct
   | RelKey i -> combinesmall 3 (Int.hash i)
   | EvarKey evk -> combinesmall 4 (Evar.hash evk)
 end
+[@@@ocaml.warning "+41"]
```
gasche: Just using `[@@ocaml.warning "-41"]` here to disable the warning locally (for this structure item `module IdKeyHash = ...`) should work, I believe, or you could disable it within the submodule with `struct [@@@ ...]` (no need for a close then).

kernel/vmvalues.ml
```diff
@@ -132,12 +132,14 @@ type id_key =
...
 | EvarKey of Evar.t
 
+[@@@ocaml.warning "-41"]
 let eq_id_key k1 k2 = match k1, k2 with
 | ConstKey c1, ConstKey c2 -> Constant.equal c1 c2
 | VarKey id1, VarKey id2 -> Id.equal id1 id2
 | RelKey n1, RelKey n2 -> Int.equal n1 n2
 | EvarKey evk1, EvarKey evk2 -> Evar.equal evk1 evk2
 | _ -> false
+[@@@ocaml.warning "+41"]
```
gasche: Why not just disambiguate the labels to silence the warning?

gasche (commented):
I looked at the PR. Overall the changes seem sensible to me and I think they improve the codebase; if it was my project, I would merge it.

There are two specific patterns of change that I don't like:

- replacing `| foo ->` into `| _foo ->` rather than `| _ ->` does not improve readability unless `foo` is a particularly informative name
- I'm not fond of your bracketed imperative use of `[@@@warning "+n"]...[@@@warning "-n"]`. You should be able to scope the warning better to use only one occurrence. Also, it seems that several occurrences of this pattern occur at places where the code could easily be fixed: for +41, you could just use an explicit module path to disambiguate (disambiguating just one of the constructor, or adding a type annotation, should be enough), and for 45 you can use `open!` instead of `open` at the place that shadows an identifier (or use an explicit path to disambiguate the identifier). In particular, I don't personally think that it is ever a good idea to silence 41, you should always fix the code.

kernel/vmvalues.ml
gasche: (I think it's always a good idea to respect warning 41 and fix the code.)

ejgallego:
Thanks a lot for the review and feedback @gasche . A few comments:

> * You should be able to scope the warning better to use only one occurrence.

IIRC I had to use the `[@@@ ...]` form due to problems with OCaml 4.02.3. So this has to stay until we raise the minimal required version.

> (I think it's always a good idea to respect warning 41 and fix the code.)

Indeed, but I wanted this patch to be "no code changes at all". But indeed I can add an extra commit that actually modifies the code.

Regarding your comments, I do agree with all of them. However I won't implement them until I get confirmation this PR is going to be merged and in which form so I can avoid useless work.

Zimmi48:
>So this has to stay until we raise the minimal required version.

Then it could make sense to merge #7522 first.

gares:
> However I won't implement them until I get confirmation this PR is going to be merged and in which form

My position is that the PR is fine as long as the fixes to the code are actually fixes and not just silencing the warning. An extra commit would help reviewing.

Enabling all warnings, or just the ones for which a fix is doable in a reasonable amount of time is up to you I guess.

ejgallego:
This is already taking 6 months so if we don't make progress soon I will orphan this PR.

kernel/subtyping.ml
ejgallego: It only occurs a couple of times so IMHO we can live with that.

kernel/nativelib.ml
ejgallego: Done.

kernel/vmvalues.ml
ejgallego: I had to use a type annotation due to the type being defined in the same module.

maximedenes (approved):

ejgallego:
Merci !

maximedenes:
@ejgallego I missed that this PR does not update merlin. Can you do so ASAP? Kernel development is really painful otherwise.

ejgallego:
Ups indeed sorry, as I am usually using the auto-generated one I didn't have this problem.

PR #6801, opened 2018-02-21
Highlight differences between successive proof steps (color, underline, etc.)

jfehrle:
This is a low-risk change because there is no change in behavior unless the -diffs command-line parameter is passed to coqtop.  For coqide, the user can pass `-diffs off` or use `Set Diffs off.` at the Coq command line to disable diffs (they default to enabled).

Please see proof_diffs.ml and pp_diff.ml comments for a description.

**Kind:** feature.

- [ ] Corresponding documentation was added / updated.
- [x] Entry added in CHANGES.

Note: the HTML output code was removed.  It will appear in a later PR.
Examples (from the HTML output):

The darker green text was added by the most recent proof step.  (Deletions are not shown. There is a switch in the code to show both the new line as well as the old line with the deletions highlighted in red.)

diff add

This shows both additions and deletions:

diff add and remove

There is special handling for hypothesis lines in the form "n, m : nat".  In the following example, "n : nat" became "n, m : nat".

n comma m

Example from coqtop:  Note that Wikipedia advises that most ANSI terminal interpreters don't support strikeout text, so it's not used for coqtop.

coqtop diff add and remove

EDIT: Fixes #1763.
EDIT: Updated screen shots and text on 6/30.

PR #7572, opened 2018-05-21
Remove unused Printer.printer_pr override mechanism.

jfehrle:
**Kind:**  feature.

This functionality made it possible to override the pr_subgoals, pr_subgoal and pr_goal methods in Printer.ml.  @ejgallego suggested this could be removed.  It simplifies the changes in #6801.

Should this be mentioned in CHANGES?

### DIFF
--- a/printing/printer.ml
+++ b/printing/printer.ml
@@ -494,7 +494,7 @@ let pr_transparent_state (ids, csts) =
 	str"CONSTANTS: " ++ pr_cpred csts ++ fnl ())
 
 (* display complete goal *)
-let default_pr_goal gs =
+let pr_goal gs =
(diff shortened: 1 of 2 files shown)


ejgallego (approved):
Thanks @jfehrle , I am fine with this removal and indeed it has been on my radar for quite a bit of time as it seems unused.

It think it is safe to say that as of today at least toplevels follow a different goal printing path.

What to do for plugins, that's an interesting question, and I think the API would have to be redesigned anyways.

Zimmi48:
Watch out @jfehrle that you have opened this branch on the main Coq repo. Please reopen a new PR with a copy from your fork.

Issue #6724, opened 2018-02-09
What commit from 3rd party developments should CI test?


SkySkimmer:
Currently we test some branch, usually master, occasionally some 8.8 specific branch. This regularly causes CI failures when Coq master and the 3rd party branch have moved on but CI is on some branch with an older base (mostly these failures are in Equations).
It also means that if I look at some old commit I can't reproducibly run CI on it because the 3rd party branch has moved on from what it was at the time of the commit.
What can we do about this?

Zimmi48:
I think that the conclusion that was reached on Gitter was:
- we should always test specific hashes of the external developments instead of testing a branch
- we should periodically update these hashes (probably a bot can be programmed to do that by opening a PR updating the hashes every week or every two weeks and using CI to confirm that the PR can be merged).

This approach has many advantages:
- as you say it means that testing again an older commit won't fail because the external development has received backward-incompatible fixes
- if CircleCI is testing a feature branch that is not up-to-date with master, it won't yield unwanted failures
- our CI won't break when external developers mess up
- it will provide a pretty good way of dealing with dependencies between external developments (in particular it should help with the stdpp -> iris -> lambda-rust dependency chain and with #6639 - VST not rebuilding CompCert)

I was considering implementing this using GitHub feature of being able to get an archive for a given commit hash. Example: https://github.com/mattam82/Coq-Equations/archive/7e507d5.tar.gz. It would be much more efficient than non-shallow cloning, which would be the other option.
Notice that in this specific example I used the a commit hash that is in a PR that is not merged yet (at the time of writing). This GitHub (mis-)feature could come in handy in our case to manage "overlays".
The same feature exists with GitLab. Example (using once again a commit from an unmerged PR): https://gitlab.mpi-sws.org/FP/iris-coq/repository/d0253c0/archive.tar.gz
This feature even exists on gforge. Example: https://scm.gforge.inria.fr/anonscm/gitweb?p=coquelicot/coquelicot.git;a=snapshot;h=0e01746;sf=tgz. But of course, in this case there are no PRs.

If no one is opposed to this approach, I'm ready to volunteer to implement it (unless someone else wants to do it).

ejgallego:
Hi Théo, I'm not sure that the motivation to modify this aspect of the
CI is strong enough.

IMHO, the current "continuous" approach has worked very well with
_near to zero_ problems, and the benefits IMHO again have been great
at a very reasonable cost. Ensuring that everything is "in sync" is a
[...]

PR #7574, opened 2018-05-22
Improve merging and overlay documentations.

Zimmi48:
Clarification prompted by @jfehrle.

**Kind:** documentation

### DIFF
--- a/dev/ci/README.md
+++ b/dev/ci/README.md
@@ -107,19 +107,32 @@ there are some.
 
 You can also run one CI target locally (using `make ci-somedev`).
 
-Whenever your PR breaks tested developments, you should either adapt it
-so that it doesn't, or provide a branch fixing these developments (or at
-least work with the author of the development / other Coq developers to
-prepare these fixes). Then, add an overlay in
-[`dev/ci/user-overlays`](/dev/ci/user-overlays) (see the README there)
-as part of your PR.
-
-The process to merge your PR is then to submit PRs to the external
-development repositories, merge the latter first (if the fixes are
-backward-compatible), and merge the PR on Coq then.
-
 See also [`test-suite/README.md`](/test-suite/README.md) for information about adding new tests to the test-suite.
 
+### Breaking changes
+
+When your PR breaks an external project we test in our CI, you must prepare a
+patch (or ask someone to prepare a patch) to fix the project:
+
+1. Fork the external project, create a new branch, push a commit adapting
+   the project to your changes.
+2. Test your pull request with your adapted version of the external project by
+   adding an overlay file to your pull request (cf.
+   [`dev/ci/user-overlays/README.md`](/dev/ci/user-overlays/README.md)).
+3. Fixes to external libraries (pure Coq projects) *must* be backward
+   compatible (i.e. they should also work with the development version of Coq,
+   and the latest stable version). This will allow you to open a PR on the
+   external project repository to have your changes merged *before* your PR on
+   Coq can be integrated.
+
+   On the other hand, patches to plugins (projects linking to the Coq ML API)
+   can very rarely be made backward compatible and plugins we test will
+   generally have a dedicated branch per Coq version.
+   You can still open a pull request but the merging will be requested by the
+   developer who merges the PR on Coq. There are plans to improve this, cf.
+   [#6724](https://github.com/coq/coq/issues/6724).
+
+Moreover your PR must absolutely update the [`CHANGES`](/CHANGES) file.
 
 Advanced GitLab CI information
 ------------------------------
--- a/dev/ci/user-overlays/README.md
+++ b/dev/ci/user-overlays/README.md
@@ -1,8 +1,21 @@
 # Add overlays for your pull requests in this directory
 
-An overlay is a file containing very simple logic to test whether we are currently building a specific pull request or git branch (useful so that overlays work on your own fork) and which changes some of the variables whose default can be found in [`ci-basic-overlay.sh`](/dev/ci/ci-basic-overlay.sh).
+When your pull request breaks an external development we test in our CI, you
+must prepare a patch (or ask someone to prepare a patch) to fix this development.
+Backward compatible patches are to be preferred, especially on libraries (it is
+harder to make backward compatible patches for plugins).
 
-The name of your overlay file should be of the form `five_digit_PR_number-GitHub_handle-branch_name.sh`.
+Once you have a patched version, you can add an overlay to your pull request:
+this is a file which defines where to look for the patched version so that
+testing is possible. It changes the value of some variables from
+[`ci-basic-overlay.sh`](/dev/ci/ci-basic-overlay.sh) (generally both the
+`_CI_BRANCH` and the `_CI_GITURL` variables of a given development at once).
+
+The file contains very simple logic to test the pull request number or branch
+name and apply it only in this case.
+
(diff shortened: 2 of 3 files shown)


dev/ci/user-overlays/README.md
```diff
@@ -1,8 +1,21 @@
...
+this is a file which defines where to look for the patched version so that
+testing is possible. It changes the value of some variables from
+[`ci-basic-overlay.sh`](/dev/ci/ci-basic-overlay.sh) (generally both the
+`_CI_BRANCH` and the `_CI_GITURL` variables of a given development at once).
+
+The file contains very simple logic to test the pull request number or branch
+name and apply it only in this case.
+
+The name of your overlay file should be of the form
+`five_digit_PR_number-GitHub_handle-branch_name.sh`.
```
SkySkimmer: I think we should relax this to 5 digits then anything `.sh`.
`[0-9]{5}-[a-zA-Z0-9-_]+.sh`

SkySkimmer (approved):

dev/ci/user-overlays/README.md
Zimmi48: OK

dev/ci/user-overlays/README.md
```diff
@@ -1,8 +1,22 @@
 # Add overlays for your pull requests in this directory
 
-An overlay is a file containing very simple logic to test whether we are currently building a specific pull request or git branch (useful so that overlays work on your own fork) and which changes some of the variables whose default can be found in [`ci-basic-overlay.sh`](/dev/ci/ci-basic-overlay.sh).
+When your pull request breaks an external development we test in our CI, you
+must prepare a patch (or ask someone to prepare a patch) to fix this development.
```
jfehrle: "development" is not quite right.  Maybe "external code" or "external project", then at the end of the sentence "the" (or "that") instead of "this".  (Because "external" seems for something distant while "this" seems for something close IMHO.)

dev/doc/MERGING.md
```diff
@@ -92,7 +92,9 @@ messy history when there are conflicts.
 ### What to do if the PR has overlays
 
 If the PR breaks compatibility of some developments in CI, then the author must
-have prepared overlays for these developments (see [`dev/ci/README.md`](/dev/ci/README.md))
+have prepared overlays for these developments (see
```
jfehrle: Same comment on "developments" here.

dev/ci/user-overlays/README.md
```diff
@@ -1,8 +1,22 @@
...
+Backward compatible patches are to be preferred, especially on libraries (it is
+harder to make backward compatible patches for plugins).
```
jfehrle: Can you perhaps add a pointer to instructions for preparing a patch file (or files)?  Do we use sed or something else?  And where do patch files go in the source tree?

dev/ci/user-overlays/README.md
```diff
@@ -1,8 +1,22 @@
 # Add overlays for your pull requests in this directory
 
-An overlay is a file containing very simple logic to test whether we are currently building a specific pull request or git branch (useful so that overlays work on your own fork) and which changes some of the variables whose default can be found in [`ci-basic-overlay.sh`](/dev/ci/ci-basic-overlay.sh).
+When your pull request breaks an external development we test in our CI, you
```
jfehrle: How about adding something like "Overlays provide a way to use a modified version of external code when building a specific PR.  The overlay specifies which fork and branch to use for the external code.  See dev/ci/user-overlays/."?

Also, aren't there some additional steps needed to finish up?  I'm guessing the workflow is like this:
- create a fork+branch for the external code (in the external repository) with the desired changes
- create an overlay file for your PR
- once the PR compiles, get the external change merged
- get the PR merged

It seems like there is a failure window between the last 2 steps in which all Coq builds other than the specified PR could fail if the external change doesn't build with the old version of Coq source.  Maybe those types of changes should be discouraged/banned.  But if they are allowed, it's operationally awkward.  If there was a way to check whether the external PR has been merged into the external repository, then Coq could apply the patch in the other builds appropriately.  But maybe the cost of doing so is greater than the benefit.

In #6801, it looks like I need to add '()' to calls to Printer.pr_open_subgoals because I added optional parameters.  This would fall into the not pretty case I mentioned, unless I keep the old routine unchanged and define a parallel Printer.pr_open_subgoals2.

Zimmi48: Indeed, when a patch is not backward-compatible there is currently a problem which you identify. I plan to fix this problem by changing the way the overlay system works (cf. #6724) but here, I'm trying to document the present (far from ideal) situation.

ejgallego:
@Zimmi48 signal when you think this is ready, I'm assuming you will still do an extra pass.

Zimmi48:
Thank you for your comment @jfehrle. I've completely reworded the explanations. Let me know what you think now.

dev/ci/README.md
```diff
@@ -107,19 +107,31 @@ there are some.
...
+
+When your PR breaks an external project we test in our CI, you must prepare a
+patch (or ask someone to prepare a patch) to fix the project:
+
+1. Fork the external project, create a new branch, push a commit adapting
+   the project to your changes.
+2. Test your pull request with your adapted version of the external project by
+   adding an overlay file to your pull request (cf.
+   [`dev/ci/user-overlays/README.md`](/dev/ci/user-overlays/README.md)).
+3. Fixes to external libraries should be backward compatible (i.e. they should
```
jfehrle: Maybe clearer as "Ideally, fixes to external libraries ..." since there is the more complex case below.

Zimmi48: I am doing a distinction between pure Coq projects (libraries) vs projects that bind to the Coq ML API (plugins). In the case of libraries, it is really important that the fix is backward compatible because we test the same version of the library on `master` and on the stable branch (currently `v8.8`). This will also become less important when #6724 is implemented.

dev/doc/MERGING.md
```diff
@@ -43,10 +43,30 @@ A maintainer is expected to be reasonably reactive, but no specific timeframe is
...
+If overlays are missing, ask the author to prepare them and label the PR with
+the [needs: overlay](https://github.com/coq/coq/labels/needs%3A%20overlay) label.
+
+When fixes are ready, there are two cases to consider:
+
+- For patches that are backward compatible (best scenario), then you should get
+  external project maintainers to integrate them before merging the PR.
+- For patches that are not backward compatible (which is often the case when
+  patching plugins after an update to the Coq API), then you can proceed to
+  merge the PR and then notify project maintainers they can merge the patch.
```
jfehrle: A bit clearer to say "the external project maintainers" or "the external maintainers".

When expressing conditionality, "then" is used only after an "if", so "then" should be removed in 2 places here.

jfehrle (approved):
This is much clearer.  Just minor wording tweaks.

dev/ci/README.md
```diff
@@ -107,19 +107,31 @@ there are some.
...
+   also work with the development version of Coq, and the latest stable
+   version). This will allow you to open a PR on the external project
+   repository to have your changes merged *before* your PR on Coq can be
+   integrated.
+
+   On the contrary, patches to plugins can very rarely be made backward
```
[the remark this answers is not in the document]
Zimmi48: I'll put "On the other hand" instead.

Issue #6475, opened 2017-12-19
Improve / fix linking of the VM in bytecode builds.


ejgallego:
#6038 introduced a bug when linking the VM in bytecode mode as it doesn't set the path properly (which likely causes #5992):

```makefile
  -I kernel/byterun -dllpath $(abspath kernel/byterun) -cclib -lcoqrun -dllib -lcoqrun \
```
which is obviously incorrect as the path should be `$(COQLIBINSTALL)/kernel/byterun`.

However, a couple of questions appear wrt VM linking in the byte version:
- do we still want to link dynamically the VM, or should using `-I kernel/byterun -custom kernel/byterun/libcoqrun.a` be a better strategy?
- is the use of `VMBYTEFLAGS` when linking `kernel.cma` still necessary?

ejgallego:
After some brief chat with @xavierleroy (thanks!), it seems that the only drawback of statically linking Coq's VM would be that the resulting file format is a bit of a strange mix, so packages should be careful, for example when calling `strip`. @gares , do you think that this would be a problem in Debian?

Also, quoting https://caml.inria.fr/pub/docs/manual-ocaml/intfc.html#sec418

> For new or rarely-used libraries, static linking is much simpler to set up in a portable way.

So indeed, given the totally "private" nature of Coq's VM I am inclined here to propose that we build standalone `coqtop.byte` executables. Derived toplevels such as SerAPI already link the VM statically [otherwise managing the .so becomes a bit annoying].

This would certainly require a changes entry.

xavierleroy:
@ejgallego, you're misunderstanding how much packagers for Linux distributions hate the bizarre ELF-executable-with-OCaml-bytecode-appended file format produced by `ocamlc -custom`.  First, the `strip` command destroys the bytecode, so packagers must make special exceptions not to strip those files after installing.  Second, Debian, always of the opinionated kind, told us in no uncertain terms that they refuse to have executable files that are neither ELF nor scripts.  

That discussion took place >10 years ago, and as a result both the OCaml distro and the Coq distro stopped using `ocamlc -custom`, switching to DLLs instead.  Why do you want to regress on this issue?  The work to turn the VM interpreter in a DLL has already be done!  

> For new or rarely-used libraries, static linking is much simpler to set up in a portable way.

This feels like it was written 15 years ago, when DLL support in OCaml was the exception rather than the rule.  Today it *might* remain true for executables that are used only locally and never installed, but not for stuff that users and esp. packagers will see.

[...]

PR #7575, opened 2018-05-22
[build] Add -cclib -lcoqrun options to build of kernel.cmxa.

ejgallego:
It seems that it is standard practice in the OCaml world to set the
`-cclib` flags at library creation time, at least in native libraries.

Indeed, this seems to make linking easier as seen for example in #7563.

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -393,7 +393,6 @@ $(COQTOPEXE): $(TOPBIN:.opt=.$(BEST))
 bin/%.opt$(EXE): topbin/%_bin.ml $(LINKCMX) $(LIBCOQRUN)
 	$(SHOW)'COQMKTOP -o $@'
 	$(HIDE)$(OCAMLOPT) -linkall -linkpkg $(MLINCLUDES) \
-	                   -I kernel/byterun/ -cclib -lcoqrun \
 			   $(SYSMOD) -package camlp5.gramlib \
 			   $(LINKCMX) $(OPTFLAGS) $(LINKMETADATA) $< -o $@
 	$(STRIP) $@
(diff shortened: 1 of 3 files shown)


ejgallego:
So far this works as intended in #7563

Makefile.build
```diff
@@ -577,6 +576,11 @@ kernel/kernel.cma: kernel/kernel.mllib
 	$(SHOW)'OCAMLC -a -o $@'
 	$(HIDE)$(OCAMLC) $(MLINCLUDES) $(BYTEFLAGS) $(VMBYTEFLAGS) -a -o $@ $(filter-out %.mllib, $^)
 
+# Specific rule for kernel.cmxa as to adjoin -cclib -lcoqrun
+kernel/kernel.cmxa: kernel/kernel.mllib
+	$(SHOW)'OCAMLOPT -a -o $@'
+	$(HIDE)$(OCAMLOPT) $(MLINCLUDES) $(OPTFLAGS) -I kernel/byterun/ -cclib -lcoqrun -a -o $@ $(filter-out %.mllib, $^)
```
SkySkimmer: It's not new in this PR, but why do we use `filter-out` instead of `$<`?

ejgallego: No idea.

gares: I'm a bit rusty, why only the .cmxa archive needs the extra flag? and not the .cma ?

gares (approved):

Makefile.build
ejgallego: The `cma` file gets the `VMBYTEFLAGS`; the way C code is linked in bytecode builds is a bit different c.f. #6475 , so I dunno what the right magic here should be.

ejgallego: So likely the `-cclib` flag only works in bytecode if using `-custom` as in native.

gares:
I've now merged the overlay and re-run the job.
Still there is something fishy: what did jbuilder do? There was no configuration file, hence it did not build it. Still exit code 0?

gares:
Really?
```
gares@gargamel:~$ eval `opam config --root=~/opam-roots/4.06/ env`
gares@gargamel:~$ mkdir foo
gares@gargamel:~$ cd foo
gares@gargamel:~/foo$ jbuilder  # opens the manpage
gares@gargamel:~/foo$ jbuilder build @install # terminates immediately
gares@gargamel:~/foo$ echo $?
0
```

ejgallego:
I see what you mean, and indeed that's a bit confusing but it makes sense.

- `jbuilder build @install` is the "developer" mode build, it will collect all the packages that are in the tree marked by the `dune-workspace` and prepare a composed build. If there are 0 packages, you get the identity.

If you want to build a package in "release" mode, you want to use `jbuilder -p $PACKAGE build`. That should complaint about the missing package.

ejgallego:
I think Pierre may be busy so requesting the secondary maintainer to take over the PR @gares

PR #7578, opened 2018-05-22
Allow make clean to work on a fresh clone

### LINKED ISSUE #7542: `make clean` should not fail on a fresh clone
#### Version

8.9+alpha (5281317cb558f2b9aa6f854b9c7aeb617beba8e6)

#### Operating system

Linux

`make clean` should succeed on a fresh clone.  Since a2f4a43833a29a5ede5905225b814c33e3a46132, it says
```
make -C test-suite clean
make[1]: Entering directory '/home/jgross/Documents/repos/coq-temp/test-suite'
Makefile:28: ../config/Makefile: No such file or directory
make[1]: *** No rule to make target '../config/Makefile'.  Stop.
make[1]: Leaving directory '/home/jgross/Documents/repos/coq-temp/test-suite'
Makefile:216: recipe for target 'indepclean' failed
make: *** [indepclean] Error 2
```

`git bisect` says:

> a2f4a43833a29a5ede5905225b814c33e3a46132 is the first bad commit
> commit a2f4a43833a29a5ede5905225b814c33e3a46132
> Author: Paul Steckler <steck@stecksoft.com>
> Date:   Fri Feb 23 14:23:19 2018 -0500
> 
>    add unit tests to test suite
> 
> ```
> :100644 100644 e2a97b3a121dde0aac31a64802a53dfba31b17ac f1960ba684548bda3fae4c1d80f435be49406e96 M      .gitignore
> :100644 100644 dbe152b5b5d37493bbd2ecefea95f5fd05a4b67e 44d24eaa28b75bf37c395099707771306ebb6eff M      .travis.yml
> :100644 100644 793afb661ec7b08b6129ff02279424fdb59cdfbf f914de5a61dc9d89f37be313850c6a0a4b305f72 M      Makefile
> :100644 100644 eed41fbe71cdad0cfb8fe1355a7b82beed9efbf4 f3b39ff2608e38a64ce62a000757beaa475a7eff M      Makefile.common
> :040000 040000 c57b6a5071bcfdefdbe6252d5fdd7ea40ce8f63c 62bf595f35c595a26d61ff1eb29ea1e23ad472c7 M      dev
> :040000 040000 a2a4fe38e37973706681359d520fa8bc433463dd 4ca11cc640035fed631b961e50eefda9bcf98d33 M      test-suite
> ```
> bisect run success

JasonGross:
The file `config/Makefile` doesn't exist unless we run `./configure`.
We shouldn't have to run `./configure` to run `make clean`.  We still
error if we try to make any non-clean target in the test-suite Makefile
without first running configure; I did it this way because it is the
minimal change in behavior.  The simpler change would just be to never
error if `config/Makefile` doesn't exist (just change `include
../config/Makefile` to `-include ../config/Makefile`).

We also now test this property when building on CI.

This fixes #7542

Fixes / closes #7542


gares (approved):
the test-suite part is OK, I let the others looks at the CI part

SkySkimmer (approved):

ejgallego (commented):
This looks pretty convoluted to me, not ticking the "red box", but not happy with this hack.

`make clean` should just use `rm -rf` and have the proper targets declare _unconditionally_, not this strange hack adding more complexity to an already complex setup.

ejgallego (commented):
In fact I think it is the change causing this bug what is incorrect and should be reverted. make is not an easy business but indeed using include like that is IMO a bug.

JasonGross:
> make clean should just use rm -rf 

This has nothing to do with the content of the clean target.  The issue is that the Makefile unconditionally includes a configure-generated file, even for targets ( like clean) which don't need it, and errors if the file doesn't exist.  We can either remove the use of configure-generated files ( seems hard), only error on non-existence when doing something other than clean ( what this pr does), or never error on non-existence.

JasonGross:
Btw, the simpler version of this PR would have the diff
```diff
 # Includes
 ###########################################################################
 
-include ../config/Makefile
+-include ../config/Makefile
 include ../Makefile.common
 
 #######################################################################
```
I can do that instead if it's preferred.

JasonGross:
Wait, I'm confused.  `../Makefile.common` already issues `-include config/Makefile`, so why are we including it again here?  I guess the simpler version of this PR just has the diff
```diff
 # Includes
 ###########################################################################
 
-include ../config/Makefile
 include ../Makefile.common
 
 #######################################################################
```
Shall I do this instead?

SkySkimmer:
The recursive include is relative to the original directory, so it 
doesn't work.

Gaëtan Gilbert

ejgallego:
The `-` include should work as the build system doesn't know about `config/Makefile` in the dependency chain; otherwise I am fine with the PR [which can be simplified maybe a bit not to introduce the extra variable].

I was in a sense whinning about the second inclusion.

SkySkimmer:
No need for 2 commits, just squash.

JasonGross:
I don't know how to amend a commit from the github web interface (and I don't have time at this second to edit on the command line).

JasonGross:
Squashed

JasonGross:
GitLab fails on ci-fiat-crypto, ci-ltac2, ci-equations, ci-coq-dpdgraph, and succeeds on the others.  I think this is just waiting for a review from @ejgallego ?

ejgallego:
Hi Jason I am OK with the current form, note that I didn't mean to block the first version in the frist place.

SkySkimmer:
@gares the test suite part is the main part so I think you're in charge of this PR.

PR #7582, opened 2018-05-22
[ci] Build fiat-crypto targets in sequence

JasonGross:
This should hopefully alleviate memory problems on gitlab, by first
building the `lite` targets, and then building the remaining
not-that-big targets.

**Kind:** bug fix / performance / infrastructure.


ejgallego:
Thanks, let's see how it goes.

JasonGross:
GitLab is happy with fiat-crypto on this one (https://gitlab.com/coq/coq/-/jobs/69835422)

ejgallego:
WDYT @SkySkimmer ?

SkySkimmer:
I don't see why this would help.
Maybe we should rerun the fiat-crypto job a few times to build confidence.

ejgallego:
> I don't see why this would help.

I think the problem is too much parallelism, so this patch reduces it by sequentializing so no targets from the split can run together.

> Maybe we should rerun the fiat-crypto job a few times to build confidence.

The first build actually happened in the worker with 1 thread so that's usually the problematic case, but it took quite a bit of time to finish.

ejgallego:
Build time for fiat-crypto [worker]
- master: 67m 15s [gitlab 2.30ghz 1 thread]
- this patch:

ejgallego (approved):
This patch cannot make things worse and Jason knows what he's doing here.

SkySkimmer (approved):

SkySkimmer:
@ejgallego can merge this

Gaëtan Gilbert

Zimmi48:
Here are two more "out of memory" failures after this being merged:

- https://gitlab.com/coq/coq/-/jobs/70229540
- https://gitlab.com/coq/coq/-/jobs/70364291

In both cases, the runner had only one cpu core.
Since we are able to print this information, can't we also extract it to set `NJOBS` accordingly to the number of cores? (It could even help make all the builds run faster.)

ejgallego:
> Since we are able to print this information, can't we also extract it to set NJOBS accordingly to the number of cores?

Unfortunately that information is not reliable, in some setups `/proc/cpuinfo` may return 32 CPUs if the host has them but only one would be available to the build, it depends on your container method.

In order to do this stuff reliably we need `gitlab-runner` to export to us the resource limit, so maybe we should open a bug request.

Zimmi48:
OK, another proposal then: now that the targets are divided in two parts, can we put NJOBS back to 1 for the second make?

ejgallego:
Sure that sounds fine, maybe we could just remove one expensive target? IMHO anyways the free Gitlab runners are awesome as sometimes we get 200 runners so the backlog clears every days, but long running CI tends to get preempted and I think it is natural that a free service does. So maybe we should just set up our own `fast-runner`s and be done.

PR #7593, opened 2018-05-24
Don't try to install native compiled files if native-compile is not set

jfehrle:
**Kind:** infrastructure.

`make install` tries to install native-compiled files even when the build is not using native-compile (the default option on win32 and cygwin).  This generates a vast number of error messages similar to those shown below.  It is also surprisingly slow, at least on my system.  This change makes the installation conditional.

```
install: cannot stat 'plugins/setoid_ring/.coq-native/NCoq_setoid_ring_ArithRing.cm*': No such file or directory
install: cannot stat 'plugins/setoid_ring/.coq-native/NCoq_setoid_ring_ZArithRing.o': No such file or directory
```

### DIFF
--- a/Makefile.install
+++ b/Makefile.install
@@ -82,7 +82,7 @@ endif
 
 install-tools:
 	$(MKDIR) $(FULLBINDIR)
-	# recopie des fichiers de style pour coqide
+	# copy style files for coqide
 	$(MKDIR) $(FULLCOQLIB)/tools/coqdoc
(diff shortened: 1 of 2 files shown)


jfehrle:
@MSoegtropIMC You may remember me mentioning this issue several months ago.

MSoegtropIMC:
I support this change.

Mid term I try to solve this by actually supporting native compute on Windows.

Zimmi48:
This is probably also useful for people on Linux who wish to compile with `-native-compiler no`.

maximedenes (approved):
LGTM, thanks!

jfehrle:
Thank you, sir.

PR #7594, opened 2018-05-24
Fix #5983 (many frequent AppVeyor failures) by increasing spawn timeout.

### LINKED ISSUE #5983: test-suite/success/paralleltac.v is non-deterministic
[`test-suite/success/paralleltac.v`](https://github.com/coq/coq/blob/master/test-suite/success/paralleltac.v) succeeded on [a first Travis build](https://travis-ci.org/coq/coq/builds/226960068) but failed on [a second build after an empty commit](https://travis-ci.org/coq/coq/jobs/227525341).

#### Failure log (edit to complete)

- On Travis on 2017-05-01.
- On AppVeyor on 2017-12-11.
- On AppVeyor on 2017-12-14.
- On AppVeyor on 2017-12-22.

Zimmi48:
**Kind:** bug fix

Fixes #5983

The same workaround was used for the Sphinx timeouts.

### DIFF
--- a/lib/spawn.ml
+++ b/lib/spawn.ml
@@ -10,7 +10,7 @@
 
 let proto_version = 0
 let prefer_sock = Sys.os_type = "Win32"
-let accept_timeout = 2.0
+let accept_timeout = 10.0
 
 let pr_err s = Printf.eprintf "(Spawn  ,%d) %s\n%!" (Unix.getpid ()) s
 let prerr_endline s = if !Flags.debug then begin pr_err s end else ()


gares (approved):

ejgallego:
@Zimmi48 maybe you would like to print a bit more information on what's gone wrong with the timeout, such as the original process name, etc...? The current message is close to useless.

So if you feel like doing that revision I'll wait, otherwise merging pretty soon.

ppedrot:
It seems slightly strange to hardcode that value btw. Couldn't we pass it through the environment or arguments instead?

ejgallego:
> It seems slightly strange to hardcode that value btw. Couldn't we pass it through the environment or arguments instead?

I'm not sure we have an use case for this, after all it is not clear that the windows problems are due to this, but likely some other threading in windows (TM) issue. So I prefer that for this experiment we do like this.

By the way, something that would really make sense on windows would be up the used ocaml version, maybe we want to merge the appveyor bits of #7522 ? IMHO that's way more likely to help here than this timeout stuff.

ejgallego:
So by the way do we have enough evidence that this will fix Appveyor's build? Just wondering.

Zimmi48:
> The current message is close to useless.

You mean the commit message?

> So by the way do we have enough evidence that this will fix Appveyor's build?

It passed twice already (here and on my fork). If I reword the commit, we'll have two more runs. Currently, almost half of the runs are failing because of this. And if it ends up not being enough, we can still reopen the issue.

> By the way, something that would really make sense on windows would be up the used ocaml version, maybe we want to merge the appveyor bits of #7522 ? IMHO that's way more likely to help here than this timeout stuff.

If you want to prepare a PR upping OCaml version on AppVeyor *now*, I'm OK for this PR to wait. But if you don't have time for this now, then this PR must get in because the errors are too frequent to be bearable (I don't want the mergers to "learn" to skip failed tests and too many failing tests is exactly doing this).

ejgallego:
> You mean the commit message?

I mean the timeout message.

> And if it ends up not being enough, we can still reopen the issue.

That's fine to me, I was just wondering if the closing in the commit message was just too optimistic.

> If you want to prepare a PR upping OCaml version on AppVeyor now, I'm OK for this PR to wait. But if you don't have time for this now, then this PR must get in because the errors are too frequent to be bearable (I don't want the mergers to "learn" to skip failed tests and too many failing tests is exactly doing this).

The PR is already prepared, but no harm in merging this first, it will either work or not so we can see.

So if you feel like improving the error message so we know a bit more what's not connecting back I'll wait, otherwise merging soon.

Zimmi48:
I don't feel like improving the error message because I personally don't really know what this code is doing. Maybe a request for @gares instead.

ejgallego:
> I don't feel like improving the error message because I personally don't really know what this code is doing. Maybe a request for @gares instead.

Ok so I'll merge as-is.

ejgallego:
> Maybe a request for @gares instead.

Let's open an issue to improve this tho.

gares:
The message means: I (Coq) spawned a worker telling it to connect back. It did not connect back in $n *seconds*, something is terribly wrong. Feel free to improve the wording.

Given that a second on a current machine is like a geologic era I thought this would never happen unless the machine is trashing.... In such a case an error is better than waiting forever IMHO.

ejgallego:
I was of your opinion but you know windows is special, and indeed doesn't like processes too much xd

PR #7596, opened 2018-05-24
[ci] [windows] Use newer OCaml version.

ejgallego:


Zimmi48 (approved):
Let's see how the AppVeyor build goes.

Zimmi48:
It seems that there is a problem with the num library.

ejgallego:
That's my fault on careless cherry picking...

ejgallego:
Well indeed it seems that #7594 really improved the problem, so in fact windows is so slow as the timeout to matter, it seems.

maximedenes (commented):
Don't we want to test the version we use to build the Windows package? We could bump that one at the same time, WDYT? Or go for a middle ground, like 4.05 for both.

Zimmi48:
> Don't we want to test the version we use to build the Windows package? We could bump that one at the same time, WDYT? Or go for a middle ground, like 4.05 for both.

:+1:

ejgallego:
> Don't we want to test the version we use to build the Windows package? We could bump that one at the same time, WDYT? Or go for a middle ground, like 4.05 for both.

My personal recommendation for Windows is to use the latest OCaml version, it makes no sense to use an old OCaml version for Windows given how many fixes for the platform each release includes.

Most of the work needed to do it has happened at #7522 , however not surprisingly lablgtk fails to build; we shouldn't depend on so many internals of the packages; I dunno how can fix that because if my understanding of Michael's setup is correct he needs now 10 hours to build Coq, so not easy to debug a problem then.

ejgallego:
So yeah I would vote for fixing #7522, then merging the Windows part on `master`.

Zimmi48:
At some point, we should just stop supporting CoqIDE on Windows, and recommend using vscode with vscoq instead. Unfortunately, as I understand it vscoq is not really maintained right now and more work is still needed to support it better.

ejgallego:
Well CoqIDE on Windows works fine if you build `lablgtk` for example with OPAM, so IMHO that discussion is offtopic for this PR.

What I am going to do is to split the Windows changes from #7522 to this PR and make it to depend on this one.

ejgallego:
Regarding your vscoq recommendation, have you looked at it? What is your assement on the implementation / madurity?

Zimmi48:
I haven't yet unfortunately (I meant to for a long time). My assessment is based on the apparent fact that it is recommended by teachers to their students at US universities (as an alternative to Proof-General). For sure, a deeper assessment would be required.

ejgallego:
Ok, build changes done, now we need to fix the build.

MSoegtropIMC (approved):
Thanks for the update (would have been my job!)

Changes are all fine.

MSoegtropIMC:
P.S.: Of cause I assume the lablgtk build errors on gitlab will be fixed. In case you need help with this, please let me know.

ejgallego:
This has been ready for a while.

MSoegtropIMC:
Sorry, no time before / during CIW. Will merge this weekend or latest on Monday.

maximedenes (approved):

PR #7556, opened 2018-05-18
Add a setting to warn about empty object in the refman

cpitclaudel:
**Kind:** infrastructure

Part of https://github.com/coq/coq/issues/7407

- [x] Corresponding documentation was added / updated.

This adds a setting to warn about empty Coq objects:

```
reading sources... [  5%] addendum/extended-pattern-matching
/build/coq/docguide/doc/sphinx/addendum/extended-pattern-matching.rst:265: (INFO/1) No contents in directive coq:opt
reading sources... [  7%] addendum/extraction
/build/coq/docguide/doc/sphinx/addendum/extraction.rst:107: (INFO/1) No contents in directive coq:cmd
/build/coq/docguide/doc/sphinx/addendum/extraction.rst:108: (INFO/1) No contents in directive coq:cmd
/build/coq/docguide/doc/sphinx/addendum/extraction.rst:109: (INFO/1) No contents in directive coq:cmd
/build/coq/docguide/doc/sphinx/addendum/extraction.rst:288: (INFO/1) No contents in directive coq:cmdv
reading sources... [ 10%] addendum/generalized-rewriting
/build/coq/docguide/doc/sphinx/addendum/generalized-rewriting.rst:182: (INFO/1) No contents in directive coq:cmd
/build/coq/docguide/doc/sphinx/addendum/generalized-rewriting.rst:199: (INFO/1) No contents in directive coq:cmd
/build/coq/docguide/doc/sphinx/addendum/generalized-rewriting.rst:229: (INFO/1) No contents in directive coq:cmd
/build/coq/docguide/doc/sphinx/addendum/generalized-rewriting.rst:539: (INFO/1) No contents in directive coq:tacv
/build/coq/docguide/doc/sphinx/addendum/generalized-rewriting.rst:542: (INFO/1) No contents in directive coq:tacv
/build/coq/docguide/doc/sphinx/addendum/generalized-rewriting.rst:545: (INFO/1) No contents in directive coq:tacv
/build/coq/docguide/doc/sphinx/addendum/generalized-rewriting.rst:548: (INFO/1) No contents in directive coq:tacv
/build/coq/docguide/doc/sphinx/addendum/generalized-rewriting.rst:551: (INFO/1) No contents in directive coq:tacv
/build/coq/docguide/doc/sphinx/addendum/generalized-rewriting.rst:576: (INFO/1) No contents in directive coq:cmd
[...]

PR #7606, opened 2018-05-25
Allow more than one signature and name per Sphinx object

cpitclaudel:
As discussed in GH-7556.  This makes it possible to write this:

```rest
  .. cmdv:: Lemma @ident {? @binders} : @type
            Remark @ident {? @binders} : @type
            Fact @ident {? @binders} : @type
            Corollary @ident {? @binders} : @type
            Proposition @ident {? @binders} : @type
     :name: Lemma; Remark; Fact; Corollary; Proposition

     These commands are all synonyms of :n:`Theorem @ident {? @binders } : type`.
```

### DIFF
--- a/doc/sphinx/proof-engine/ltac.rst
+++ b/doc/sphinx/proof-engine/ltac.rst
@@ -178,7 +178,7 @@ Sequence
 A sequence is an expression of the following form:
 
 .. tacn:: @expr ; @expr
-   :name: ;
+   :name: ltac-seq
 
    The expression :n:`@expr__1` is evaluated to :n:`v__1`, which must be
    a tactic value. The tactic :n:`v__1` is applied to the current goal,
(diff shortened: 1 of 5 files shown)


doc/sphinx/proof-engine/ltac.rst
```diff
@@ -178,7 +178,7 @@ Sequence
 A sequence is an expression of the following form:
 
 .. tacn:: @expr ; @expr
-   :name: ;
+   :name: ltac-seq
```
Zimmi48: Can't you escape the `;` character instead? With this renaming, this index becomes completely useless.

cpitclaudel: Ah, good point.  The problem is that this is already super brittle (the `name` tags are used as HTML anchors, and `;` can't be in one, so the cleanup / normalization process turns this into an empty id.

We could use a separate `:index:` option.

cpitclaudel: OK, I'm happy to work on this. Will keep you posted.

Zimmi48: In the meantime, I would be OK to merge this and just removing this index.

Zimmi48 (approved):
In fact, let's merge as-is for now and fix the `;` index later.

maximedenes (approved):
I haven't checked the rendered output, but I like the idea.

PR #7607, opened 2018-05-26
Simplify reification of predicate in bytecode and native compilers

maximedenes:


maximedenes:
Any takers? @ppedrot ? @silene ?

ppedrot:
I stumbled upon this code already and had a hard time understanding what the `dep` flag was standing for. I still don't quite understand, so I'm not sure I should take this PR.

ppedrot:
Also, why do we rely on the shape of the type of the predicate? We know statically the number of arguments, don't we?

maximedenes:
> I stumbled upon this code already and had a hard time understanding what the dep flag was standing for. I still don't quite understand, so I'm not sure I should take this PR.

I can explain : in a previous life, the return predicate could have two different shapes, depending on whether the match was dependent or not. When it was, the predicate was like today. When not, the quantification over the scrutinee was dropped.

maximedenes:
> Also, why do we rely on the shape of the type of the predicate? We know statically the number of arguments, don't we?

The number yes, but we need to build the type of the domains, not so sure if it would be easy to do otherwise.

ppedrot:
> I can explain : in a previous life, the return predicate could have two different shapes, depending on whether the match was dependent or not. When it was, the predicate was like today. When not, the quantification over the scrutinee was dropped (i.e. was not there).

Can you point to the commit that removed this non-uniformity?

herbelin:
> Can you point to the commit that removed this non-uniformity?

This quite old but maybe was it done only by steps (in the absence of a way to enforce the uniformity in the ML syntax). For instance, this was done in `indrec.ml` in 2004 (d63e8fab799).

ppedrot:
@maximedenes Can you rebase so that I can merge?

ppedrot:
@maximedenes ping!

maximedenes:
Rebased.

pretyping/vnorm.ml
```diff
@@ -313,8 +310,8 @@ and nf_predicate env sigma ind mip params v pT =
       let params = if Int.equal n 0 then params else Array.map (lift n) params in
       let dom = mkApp(mkIndU ind,Array.append params rargs) in
       let body = nf_vtype (push_rel (LocalAssum (name,dom)) env) sigma vb in
-      true, mkLambda(name,dom,body)
-    | _ -> false, nf_val env sigma v crazy_type
+      mkLambda(name,dom,body)
+    | _ -> nf_val env sigma v crazy_type
```
ppedrot: Thinking about it, shouldn't this pattern be an `assert false`? The return predicate is supposed to always be dependent now so there shouldn't be a shortage of lambdas.

ppedrot:
Now there is a linting error.

mattam82 (approved):
I had no idea there was a time where the predicate depended on the dependency. Ok with the changes.

pretyping/vnorm.ml
ppedrot: Come on, you fixed the linting but didn't set the assertion failure?

maximedenes: I can look at the code but not now. Fixing the whitespace didn't require opening the file.

ppedrot:
I am waiting for the addition of the assertion failure mentioned above. If you really don't have time @maximedenes, I can merge it as-is nonetheless, I am just a bit concerned that we will forget about this if so. This is just a cleanup so there is no actual hurry.

PR #7614, opened 2018-05-27
Simplify the code that handles trust of side-effects in kernel typing.

ppedrot:
For some reason the code was returning int option when only the value of the integer was relevant.

Trivial change that does not modify the extensional behaviour.

### DIFF
--- a/kernel/term_typing.ml
+++ b/kernel/term_typing.ml
@@ -179,36 +179,36 @@ let rec is_nth_suffix n l suf =
   | _ :: l -> is_nth_suffix (pred n) l suf
 
 (* Given the list of signatures of side effects, checks if they match.
- * I.e. if they are ordered descendants of the current revstruct *)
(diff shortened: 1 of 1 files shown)


ppedrot:
I think it was just laziness, as the code relies on the option value inside the loop.

The side-effect code is half-broken anyways, I am pretty sure I can find a proof of False, if not several, by just using safe ML code. It helps preventing involuntarily errors though. I have further cleanups pending, and at some point I really want to make it work properly.

gares:
> The side-effect code is half-broken anyways

This is the kind of bullshit comment that makes me postpone the task of reviewing your (very welcome) cleanup.

ppedrot:
@gares What if I expose a soundness bug only using the safe interface?

gares:
If the bug also comes with a PR that fixes it, then I'll pay you a beer.

maximedenes (approved):
LGTM

PR #7616, opened 2018-05-28
Fix #7615: Functor inlining drops universe substitution.

### LINKED ISSUE #7615: Functor inlining drops universe substitution
The `Mod_subst.subst_con0` function completely disregards the instance it is applied to whenever it results in an inlining.

This allows to derive this nifty proof of False.
```
Set Universe Polymorphism.

Module Type S.
Parameter Inline T@{i} : Type@{i+1}.
End S.

Module F (X : S).
Definition X@{j i} : Type@{j} := X.T@{i}.
End F.

Module M.
Definition T@{i} := Type@{i}.
End M.

Module N := F(M).

Require Import Hurkens.

Definition eqU@{i j} : @eq Type@{j} N.X@{i Set} Type@{i} := eq_refl.
Definition bad@{i} : False := TypeNeqSmallType.paradox N.X@{i Set} eqU.

Print Assumptions bad.
```

ppedrot:
We store the universe context in the inlined terms and apply it to the instance provided to the substitution function. Technically the context is not needed, but we use it to assert that the length of the instance corresponds, just in case.

Note: to backport without affecting the API, it is actually enough to add the `Vars.subst_instance_constr` line.

### DIFF
--- a/kernel/mod_subst.ml
+++ b/kernel/mod_subst.ml
@@ -24,7 +24,7 @@ open Constr
    is the term into which we should inline. *)
 
 type delta_hint =
-  | Inline of int * constr option
+  | Inline of int * (Univ.AUContext.t * constr) option
   | Equiv of KerName.t
 
 (* NB: earlier constructor Prefix_equiv of ModPath.t
@@ -158,7 +158,7 @@ let find_prefix resolve mp =
 
 (** Applying a resolver to a kernel name *)
 
-exception Change_equiv_to_inline of (int * constr)
+exception Change_equiv_to_inline of (int * (Univ.AUContext.t * constr))
 
 let solve_delta_kn resolve kn =
   try
@@ -300,9 +300,10 @@ let subst_con0 sub (cst,u) =
   let knu = KerName.make mpu dir l in
   let knc = if mpu == mpc then knu else KerName.make mpc dir l in
   match search_delta_inline resolve knu knc with
-    | Some t ->
+    | Some (ctx, t) ->
       (* In case of inlining, discard the canonical part (cf #2608) *)
-      Constant.make1 knu, t
+      let () = assert (Int.equal (Univ.AUContext.size ctx) (Univ.Instance.length u)) in
+      Constant.make1 knu, Vars.subst_instance_constr u t
     | None ->
       let knc' =
         progress (kn_of_delta resolve) (if user then knu else knc) ~orelse:knc
@@ -482,7 +483,7 @@ let gen_subst_delta_resolver dom subst resolver =
       | Equiv kequ ->
 	  (try Equiv (subst_kn_delta subst kequ)
 	   with Change_equiv_to_inline (lev,c) -> Inline (lev,Some c))
-      | Inline (lev,Some t) -> Inline (lev,Some (subst_mps subst t))
+      | Inline (lev,Some (ctx, t)) -> Inline (lev,Some (ctx, subst_mps subst t))
       | Inline (_,None) -> hint
     in
     Deltamap.add_kn kkey' hint' rslv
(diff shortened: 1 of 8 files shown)


JasonGross:
This should probably get a test-suite file.

kernel/modops.ml
```diff
@@ -403,7 +403,11 @@ let inline_delta_resolver env inl mp mbid mtb delta =
 	    | Undef _ | OpaqueDef _ -> l
 	    | Def body ->
 	      let constr = Mod_subst.force_constr body in
-	      add_inline_delta_resolver kn (lev, Some constr) l
+              let ctx = match constant.const_universes with
```
mattam82: Just nitpicking, isn’t there a helper function doing that already?

mattam82 (approved):
Good catch, the fix looks fine.

kernel/modops.ml
ppedrot: I don't remember one.

SkySkimmer: Declareops.constant_polymorphic_context

Zimmi48:
This fixes a critical bug so it should get a CHANGES entry.

kernel/mod_subst.ml
```diff
@@ -24,7 +24,7 @@ open Constr
    is the term into which we should inline. *)
 
 type delta_hint =
-  | Inline of int * constr option
+  | Inline of int * (Univ.AUContext.t * constr) option
```
maximedenes: Why not use `in_univ_context`?

ppedrot: That's not the same type, here we want an abstract context rather than a named one.

ppedrot:
@Zimmi48 done.

SkySkimmer:
>Technically the context is not needed, but we use it to assert that the length of the instance corresponds, just in case.

I don't see the point in doing that.

maximedenes:
I agree with @ppedrot, although I like code without assertions :)

SkySkimmer:
This isn't 0-cost, it literally broke the checker.

ppedrot:
Oh, yes, indeed, I forgot to port the checker... Let me fix that.

ppedrot:
I have fixed the checker, but the resulting code is a gaping backdoor. It marginally better than before (at least we check the arity of the universe instance) but we still blindly trust the vo file to contain well-typed inlined terms. As I wrote in the comment, we should not even be using substitutions in the checker but for some reason we do.

maximedenes:
I agree that it is very suspicious that the checker uses delta resolvers. @ppedrot do you want to discuss that a bit?

ppedrot:
@maximedenes We should discuss it indeed, but the current PR is quite orthogonal to this issue.

Zimmi48:
@ppedrot There are so many CI failures, it is hard to see whether they are all spurious. Would you mind rebasing to check?

ppedrot:
@Zimmi48 Done. That said, I'd pretty much bet that there is no contribution tracked in the CI (or even any contribution *at all*) excercising this code path.

Zimmi48:
> That said, I'd pretty much bet that there is no contribution tracked in the CI (or even any contribution at all) excercising this code path.

Would you be able to come up with a test suite file going through this code path with success then?

ppedrot:
Yep, the file added to the test-suite is checking that. But I was talking about real developments.

Zimmi48:
Sure, I was talking about something that would not end up with `Fail` though ;-)

ppedrot:
OK, we should indeed write some successful test, but I'm not sure this endeavour really pertains to this PR, which is only fixing a critical bug. I'd recommend merging it and opening some separate bug report.

Zimmi48:
@maximedenes I think this PR has been ready to merge for a while now :)

herbelin:
By the way, about collecting critical bugs found in the history of Coq, @ppedrot or @maximedenes, do you have some concrete plan? I can otherwise start to collect what I remember and at some time we cross our results, and one of us, or all of us put it in readable form somewhere? Or whatever other suggestion.

maximedenes:
I don't have any concrete plans, so if you can collect some information to have something to start from, it would be very helpful indeed.

maximedenes (approved):
I don't fully understand the code of inlining with delta resolvers, but this looks good to me.

PR #7620, opened 2018-05-28
[ssr] rewrite: turn anomaly into regular error

### LINKED ISSUE #7255: ssr: anomaly occurs when trying to rewrite with a logical equivalence
#### Version

Coq master (4c466c429902d0ef0bd0da0f42e1a3b6d9f7d32a)

#### Operating system

macOS 10.13.3 High Sierra

Please consider the following minimized example:

```coq
From Coq Require Import ssreflect Setoid.

Structure SEProp := {prop_of : Prop; _ : prop_of <-> True}.

Fact anomaly: forall P : SEProp, prop_of P.
Proof.
move=> [P E].
rewrite E.
```

`rewrite E` produces an anomaly:

>Error: Anomaly "cvtac's exception: Pretype_errors.PretypeError(_, _, _)"
>Please report at http://coq.inria.fr/bugs/.

Obverse that if I simplify the goal, then the anomaly goes away:
```coq
Fact no_anomaly: forall P : SEProp, prop_of P.
Proof. by move=> [P /= ->]. Qed.
```

Also, it seems that the usage of SSReflect is essential here, because 
```coq
From Coq Require Import Setoid.
Structure SEProp := {prop_of : Prop; _ : prop_of <-> True}.

Fact no_anomaly2: forall P : SEProp, prop_of P.
Proof. intros [P E]. Fail rewrite E. Abort.
```

results in the following error message:

> build_signature: no constraint can apply on a dependent argument

gares:
I think the bug was introduces when apply_type was made safe.
In the test joint to #7255 rewrite (setoid case) generates an
ill-typed goal and apply_type raises a Pretype_error.

It is unclear to me why the tactic monad does not turn the
pretype_error into a UserError

Fix #7255

### DIFF
--- a/plugins/ssr/ssrequality.ml
+++ b/plugins/ssr/ssrequality.ml
@@ -417,8 +417,6 @@ let rwcltac cl rdx dir sr gl =
       then errorstrm Pp.(str "Rewriting impacts evars")
       else errorstrm Pp.(str "Dependent type error in rewrite of "
         ++ pr_constr_env (pf_env gl) (project gl) (Term.mkNamedLambda pattern_id (EConstr.Unsafe.to_constr rdxt) (EConstr.Unsafe.to_constr cl)))
-    | CErrors.UserError _ as e -> raise e
-    | e -> anomaly ("cvtac's exception: " ^ Printexc.to_string e);
   in
   tclTHEN cvtac' rwtac gl
 
--- a/test-suite/ssr/ssr_rew_illtyped.v
+++ b/test-suite/ssr/ssr_rew_illtyped.v
@@ -1,0 +1,9 @@
+From Coq Require Import ssreflect Setoid.
+
+Structure SEProp := {prop_of : Prop; _ : prop_of <-> True}.
+
+Fact anomaly: forall P : SEProp, prop_of P.
+Proof.
+move=> [P E].
+Fail rewrite E.
+Abort.


SkySkimmer:
>It is unclear to me why the tactic monad does not turn the
pretype_error into a UserError

Why would it do that?

gares:
Maybe you are right, and the exception should be fatal.
But it is not clear to me, not documented AFAICT.

The label is `~typecheck`, and I've added it, but I would have chosen `~assert_well_typed` if it was intended to be a fatal error to pass an ill typed term. So maybe we should rename the flag?

gares:
@maximedenes

maximedenes:
@gares I think you forgot to add me as a reviewer.

plugins/ssr/ssrequality.ml
```diff
@@ -417,7 +417,7 @@ let rwcltac cl rdx dir sr gl =
       then errorstrm Pp.(str "Rewriting impacts evars")
       else errorstrm Pp.(str "Dependent type error in rewrite of "
         ++ pr_constr_env (pf_env gl) (project gl) (Term.mkNamedLambda pattern_id (EConstr.Unsafe.to_constr rdxt) (EConstr.Unsafe.to_constr cl)))
-    | CErrors.UserError _ as e -> raise e
+    | (CErrors.UserError _ | Pretype_errors.PretypeError _) as e -> raise e
     | e -> anomaly ("cvtac's exception: " ^ Printexc.to_string e);
```
maximedenes: What is this trying to catch? Why not just let the exception bubble up?

gares: I don't know what I was trying to catch here. IMO it is my (bad?) habit of having assertions that backfires me (as the "no safe_id attached" one). `cvtac` is supposed to be a tactic, hence, IMO, it should only fail with `CErrors.UserError`. On purpose I type-check the inferred predicate and raise a different exception (few lines above) to re-raise a better error message. It is unclear to me if `Pretype_errors.PretypeError _` is a legit exception that can be raised by tactics.

maximedenes:
@gares this is missing a test case

gares:
there you go

Zimmi48:
Same remark as in another PR of yours: this will need a rebase to make the CI happy.

gares:
ping @maximedenes

gares:
ci failures on gitlab seem spurious, I've restarted the windows one (segmentation fault)

gares:
Ready to be merged

Zimmi48:
@gares Any reason not to put this in the 8.8.1 milestone? Given the size of the patch and what it's changing, I would say it is backportable.

maximedenes (approved):

Zimmi48:
@gares ping

PR #7621, opened 2018-05-28
Tactics.introduction: remove dangerous option ~check

gares:
In locally nameless mode (proof mode) names in the context *must*
be distinct otherwise the term representation makes no sense.

### DIFF
--- a/plugins/funind/functional_principles_proofs.ml
+++ b/plugins/funind/functional_principles_proofs.ml
@@ -598,7 +598,7 @@ let treat_new_case ptes_infos nb_prod continue_tac term dyn_infos =
 	Proofview.V82.of_tactic (intro_using heq_id);
 	onLastHypId (fun heq_id -> tclTHENLIST [
 	(* Then the new hypothesis *)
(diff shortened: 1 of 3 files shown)


gares:
Checks have passed, this is an easy one! Reviewers, crash it!

ppedrot (approved):
Why did we have that in the first place?

herbelin:
> Why did we have that in the first place?

I guess it is to save the time of some check when it is clear what the result of the check will be???

Well, here, the added time is in the size of the context, I don't know if it is noticeable. Also, `intro_then_gen` which is the main client of `introduction` could in theory take benefit of `check:false` since it already ensures the freshness itself.

JasonGross:
Is there a user-visible change here?  e.g., could building a functional scheme fail in the past if you were unlucky with section variable names?

gares:
I frankly don't know, but if it used to work by chance then I prefer it to fail predictably now (and eventually fix the code).

herbelin:
I apparently arrive after the battle, but I missed the point of removing the test. I would have said that a best fix would have been to exploit the possibility of a check (e.g. in `intro_then_gen`), even if maybe not significantly noticeable in terms of performance??

Zimmi48:
There wasn't a performance assessment BTW.

PR #7625, opened 2018-05-28
Fix #7017: Problem with extraction of Decimal.v

### LINKED ISSUE #7017: Problem with extraction of Decimal.v (the type int is redefined and masks OCaml's one)
#### Version

version 8.8+alpha (February 2018), c921825

When I run the following:
```
Require Import Extraction BinPos.
Require Import ExtrOcamlNatInt.

Recursive Extraction Library BinPos.
```
And then I try:
```
ocamlbuild BinPos.cmo
```

It fails with `This expression has type int but an expression was expected of type Decimal.int`.

The problem is that `Decimal.v` redefines a type `int` and then the `open Decimal` masks the OCaml type `int`.

Should `int` be renamed as something like `coq_int`?

maximedenes:
Since OCaml builtins cannot be used under qualified names, they should
not be masked by user defined types, so we treat them like keywords.

Fix #7017


letouzey (changes requested):
On the long term, I think that's the proper way to fix this issue. But the compatibility consequences will be huge for many users, especially if you want to push that in 8.8.1 : I'm afraid a *lot* of manual code around extracted code will have to be adapted  (e.g. `bool` or `list` becoming `bool0`, `list0` or `coq_bool`, `coq_list` depending of your extraction mode). Hopefully, many users already map these types to the real OCaml ones through `Extract Inductive`, and hence won't be affected, or in a limited way. But that's certainly not all of our users. So I rather suggest we keep this proposed fix for 8.9, and we do something less drastic for 8.8.1, like temporary renaming `Decimal.int` or temporary moving the `int<->nat,Z,...` conversion functions away from `PeanoNat`, `BinInt` and co (hence cutting the dependency over `Decimal`). Or even removing this `Decimal` stuff from the 8.8 branch for now. Neither solution is particularly appealing to me, but any of them is more reasonable than an extraction change of this scale in a minor version.

maximedenes:
@letouzey Thanks for the feedback, I'll change the milestone. Are you willing to be the assignee and merge it?

maximedenes:
In fact, is seems to break stuff, so I need to have a look first.

maximedenes:
extraction has no secondary maintainer, who's willing to be the assignee? @ppedrot ?

ejgallego:
I am wary of this fix, I think that extraction should not silently rename such names on clash but fail, instead forcing the user to specify a mapping (let it be to the built in or not)

maximedenes:
Ok, but this is already not the behavior of extraction on keywords. So I think you're request is orthogonal to this fix. If you think otherwise, I'll drop the fix since it is going to be much more work.

maximedenes:
> instead forcing the user to specify a mapping

Maybe the misunderstanding about the amount of work comes from here: do you think there's already the machinery in place for specifying such mappings?

ejgallego:
> Why ?

The thing that doesn't feel right is that your reaction is not constructive, to a comment we maybe cannot address your answer is "I drop the fix" which is obviously not the best thing to do in this case.

If you cannot implement this then you could say "I'm afraid we'll have to get this like that and open an issue for later work". Unless of course yourself don't believe this should get in.

maximedenes:
> The thing that doesn't feel right is that your reaction is not constructive, to a comment we maybe cannot address your answer is "I drop the fix" which is obviously not the best thing to do in this case.
> 
> If you cannot implement this then you could say "I'm afraid we'll have to get this like that and open an issue for later work". Unless of course yourself don't believe this should get in.

Ah ok, so maybe I simply overestimated the importance of your remark. But it was just a sign I was taking it seriously. This whole thing in extraction about uncontrolled blacklisting, silent renamings, etc feels very shaky, so I'm also not comfortable with making it even more complex.

Zimmi48:
Indeed @maximedenes, the reply "I'll drop the fix" felt like a threat (of the "or else..." kind) even though you didn't mean it that way. If you had said "This would represent more work that I'm willing to put into this so if you feel strongly about this, we'll have to label this as help wanted.", this would indeed have avoided any misunderstanding.

maximedenes:
> the reply "I'll drop the fix" felt like a threat

It's a bit sad that everything tends to get interpreted like this in this team, but ok, note taken.

JasonGross:
What about adding a command that lets users extend the blacklist?  (Or using making the blacklist a setting so that users can `Set Extraction Blacklist Append ...` or whatever)  Then the specific `ExtrOcaml*` files can add the types and functions they use in extraction to the blacklist, and since CompCert does not use ExtrOcamlNatInt, it will not be impacted by this change.

SkySkimmer:
Extraction Blacklist is about filenames according to the doc, although it doesn't seem to be used by the code, so I don't know what you mean.

JasonGross:
This PR adds a bunch of things to a list. (or a set?  A list that is used to build a set?)  I am proposing that we allow users to extend that set, and that we move the listing to the `ExtrOcaml*.v` files.  If blacklist is already taken, then maybe `Set Extraction Keyword List Append`?  Or whatever name describes it well

SkySkimmer:
IIUC it's a set of things that must be renamed for extraction to be correct. I don't think it makes sense to make the user extend that.
Maybe we should go with the current PR and add a bunch of Extract Inductive by default for the ones which are reasonable (typically option but not int).

mattam82:
`Extraction Blacklist` should be extended to handle identifiers also, it serves precisely the same purpose of avoiding clashes.

mattam82:
And `Extraction Blacklist` works by the way, we use it daily in MetaCoq.

liyishuai:
What's the status of this PR? Can the fix be backported to 8.8?

Zimmi48:
> Can the fix be backported to 8.8?

You should not expect any more releases of Coq 8.8 or 8.9 if that's what you're asking. If this is just about backporting to v8.8 and v8.9 without a release, this can be discussed. But it's not like this PR has made much progress since it was opened very long ago...

liyishuai:
@bcpierce00 @Zdancewic may I propose a higher priority in merging this pull request? This would benefit our developments under @DeepSpec.

maximedenes:
Note that starting with OCaml 4.08, a cleaner fix is possible, since the built-in `int` type can be referred to in an absolute way: `Stdlib.Int.t`. It would then be enough to blacklist the `Stdlib` module name.

Not sure how far we are from requiring OCaml 4.08.

bcpierce00:
We'd be happy with any solution that reduces the current number of hacks required.  :-)

I suppose if we're not ready to require 4.08, there could be a flag that says "I'm using 4.08; please do this the right way"...

Zimmi48:
4.08.1 is in Debian testing (Bullseye), which is not expected before "well into 2021", so if we keep our current policy regarding this, we should update our minimal supported version around Coq 8.14.

ejgallego:
Also note that it is now possible to have the OCaml target backend target a `coq-core.plugins.extraction-runtime` library in a kinda clean way [not sure about the other backends tho]

PR #7628, opened 2018-05-29
[default.nix] Update dependency list

vbgl:
Use OCaml 4.06 as the default.
Drop ocp-index (I cannot manage to build it with OCaml 4.06).
Adds hostname and ounit, which were missing.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -67,6 +67,7 @@ stdenv.mkDerivation rec {
     python
     rsync
     which
+    ocamlPackages.ounit
 
   ] else []) ++ (if lib.inNixShell then [
     ocamlPackages.merlin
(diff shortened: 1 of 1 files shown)


Zimmi48 (approved):

vbgl:
Yes. There is a bug in `opam-installer` (fixed there: https://github.com/ocaml/opam/commit/718d6198e338069852dcdde067f56df236235370). It lead to introduce a similar bug in `nixpkgs` (at https://github.com/NixOS/nixpkgs/blob/8f813241af2e005afd0acb182a7fd04f8a9d61b2/pkgs/development/tools/ocaml/findlib/default.nix#L53-L54).

I’ve a patch that fixes it as part of a larger plan that get rid of some bloated software (aka `opam`).

ejgallego:
Note that we will produce `coq.install` files tho as a mean to indicate packages which files to install.

maximedenes:
@vbgl any update on this?

vbgl:
The fix to nixpkgs (https://github.com/NixOS/nixpkgs/pull/41591) has just got merged.

PR #7629, opened 2018-05-29
Fix anomaly in autoapply when an unbound hint name is provided

### LINKED ISSUE #7627: "autoapply I using I" raises an anomaly.
#### Version
8.8.0

```
Goal True.
autoapply I using I.
```

raises:
```
Ltac call to "autoapply (constr) using (preident)" failed.
Ltac call to "autoapply (constr) using (preident)" failed.
Anomaly "Uncaught exception Not_found."
```

cc @mattam82

Armael:
Fixes #7627 .

### DIFF
--- a/tactics/class_tactics.ml
+++ b/tactics/class_tactics.ml
@@ -1229,8 +1229,11 @@ let is_ground c =
 let autoapply c i =
   let open Proofview.Notations in
   Proofview.Goal.enter begin fun gl ->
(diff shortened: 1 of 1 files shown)


mattam82 (approved):
Good catch. I don't think a test-suite is necessary here and am confident the rest of CI tests would pass.

maximedenes:
> Good catch. I don't think a test-suite is necessary here and am confident the rest of CI tests would pass.

@Why not add a regression test?

PR #7635, opened 2018-05-30
Define rec_declaration in terms of prec_declaration.

SimonBoulier:
And similarly for fixpoint and cofixpoint.
No impact on API.

**Kind:** cleanup

Change made with @herbelin.

### DIFF
--- a/kernel/constr.ml
+++ b/kernel/constr.ml
@@ -107,20 +107,12 @@ type t = (t, t, Sorts.t, Instance.t) kind_of_term
 type constr = t
 
 type existential = existential_key * constr array
-type rec_declaration = Name.t array * constr array * constr array
-type fixpoint = (int array * int) * rec_declaration
-  (* The array of [int]'s tells for each component of the array of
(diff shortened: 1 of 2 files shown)


SkySkimmer (approved):

kernel/constr.ml
```diff
@@ -72,10 +72,22 @@ type case_info =
 type 'constr pexistential = existential_key * 'constr array
 type ('constr, 'types) prec_declaration =
     Name.t array * 'types array * 'constr array
+
 type ('constr, 'types) pfixpoint =
     (int array * int) * ('constr, 'types) prec_declaration
+  (* The array of [int]'s tells for each component of the array of
```
maximedenes: Can you move this comment to the `.mli` file? It really describes the API rather than internal implementation.

SimonBoulier:
Done.

kernel/constr.ml
SimonBoulier: Done

maximedenes (approved):
Looks very good, thanks for contributing!

PR #7642, opened 2018-05-30
Gitlab: retry failed jobs once

SkySkimmer:
Not sure if we want this but let's talk about it.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -50,6 +50,7 @@ before_script:
 # TODO figure out how to build doc for installed Coq
 .build-template: &build-template
   stage: build
+  retry: 1
   artifacts:
     name: "$CI_JOB_NAME"
     paths:


Zimmi48:
See also coq/bot#3.

ejgallego:
Maybe worth doing only on failed base builds; and maybe a couple of the heavier ones, as these are more likely to be interrupted.

Tho I agree that the bot can do much better as gitlab should report if the cause of the failure was due to the script or the worker.

Zimmi48:
> Maybe worth doing only on failed base builds

Why not for the base builds, they don't cost so much to retry.

> GitLab should report if the cause of the failure was due to the script or the worker

It does but it still misses a lot of runner failures (that are displayed as script failures). Should we report this whenever we see it?

SkySkimmer:
It properly retried jobs which failed due to being unrebased ^^

ejgallego:
So do we want this retry configuration? Seems a bit expensive to me?

Zimmi48:
Let's put it only for the base builds for now.

ejgallego (changes requested):
What @Zimmi48 said, let's try with base builds only for now.

SkySkimmer:
Changed

ejgallego (approved):

PR #7644, opened 2018-05-31
[vernac] make “Coercion” an attribute

vbgl:


stm/vernac_classifier.ml
```diff
@@ -199,11 +199,8 @@ let classify_vernac e =
   in
   let rec static_control_classifier ~poly = function
     | VernacExpr (f, e) ->
-      let poly = List.fold_left (fun poly f ->
-          match f with
-          | VernacPolymorphic b -> b
-          | (VernacProgram | VernacLocal _) -> poly
-        ) poly f in
+      let _, atts = Vernacentries.attributes_of_flags f Vernacinterp.{ loc = None ; locality = None ; polymorphic = poly ; program = false ; coercion = false; } in
```
gares: I don't know if it exists (if not it should be added), but I'd like to see a `empty_flags` or `default_flags` here rather than the this record, otherwise every time you add a field or change one you touch the stm for no reason.

vernac/lemmas.ml
```diff
@@ -42,6 +42,14 @@ let call_hook fix_exn hook l c =
     let e = CErrors.push e in
     iraise (fix_exn e)
 
+let bind_declaration_hook (>>=)
```
gares: Boh?

gares (changes requested):

ejgallego:
Would like to have a look a this too.

vernac/vernacentries.ml
```diff
@@ -2250,32 +2258,36 @@ let with_fail st b f =
...
+       | VernacPolymorphic _ ->
+         user_err Pp.(str "Polymorphism specified twice")
+       | VernacLocal b when Option.is_empty atts.locality ->
+         (polymorphism, { atts with locality = Some b })
+       | VernacLocal _ ->
+         user_err Pp.(str "Locality specified twice")
+       | VernacCoercion when not atts.coercion ->
+         (polymorphism, { atts with coercion = true })
+       | VernacCoercion ->
+         user_err Pp.(str "Coercion mode specified twice")
```
Zimmi48: I wonder how to raise these error messages. I haven't been able to trigger them. Could you add output test cases showing them?

Zimmi48 (commented):
Coercion is a weird choice for the next attribute given that it currently has a special syntax which is not just a leading flag. Of course, in the future, we might change this syntax to become instead:

```
"coercion"
Definition foo ...
```

and only then it will make sense...

vbgl:
Thanks for the comments, Théo. Indeed, I might be doing things in a weird order. Concrete syntax for attributes is what’s coming next. Spoiler: https://github.com/vbgl/coq/commit/cef7f46b5694ebb61c7aa9850e06697bde0d416b

I think I’ve addressed Enrico’s comments.

maximedenes:
> Would like to have a look a this too.

@ejgallego Can you have a look? Would be good to get this merged. Thanks!

ejgallego:
After a quick look I am failing to see why this is an improvement, on the contrary, it seems to make the code more convoluted.

Would you mind to explain what the gain is?

gares (approved):

ejgallego (changes requested):
Flagging so I ensure my question is answered.

vbgl:
The gain is that to determine what vernacular command defines a coercion, we no longer rely on the name of the command (here: “Coercion”) but on the attributes of this command (derived from flags decorating the command and global settings, if relevant). This further enables to contemplate attaching the “Coercion” property to other commands (“Instance”, “Axiom”, “Let”, …) and also getting rid of the specific “Coercion” syntax.

ppedrot:
This is a voicing of my support for this PR.

Zimmi48:
> This further enables to contemplate attaching the “Coercion” property to other commands (“Instance”, “Axiom”, “Let”, …) and also getting rid of the specific “Coercion” syntax.

Although "Instance" should probably be seen as an attribute as well :wink:

JasonGross:
> and also getting rid of the specific “Coercion” syntax.

I am opposed to removing Coercion as its own vernacular.  There must remain some way of defining existing constants (e.g., Zpos) to be coercions.

ppedrot:
What does the fact that we use the Coercion syntax have to do with this PR? We can provide whatever user-facing syntax for standard cases of attributes as we already do today.

vbgl:
El 22/6/18 a las 16:19, Jason Gross escribió:
>     and also getting rid of the specific “Coercion” syntax.
> 
> I am opposed to removing Coercion as its own vernacular. There must
> remain some way of defining existing constants (e.g., Zpos) to be coercions.
> 

This is not what I meant. I was talking of this grammar rule, for
defining a new constant and making it, at once, a coercion:

https://github.com/coq/coq/blob/1750982027f5465c8f7c10af44e4f3674edc1ecd/vernac/g_vernac.ml4#L622-L624

ejgallego:
I see and like the idea, however I am not sure that abusing the `attr` field [badly named by me] should be the way to go. Instead we should provide a proper registration mechanism.

Indeed, in a sense my complain is that the attribute hook is still called from `vernac_definition`, this is the part where I don't see the gain.

In fact maybe we should cleanup the whole file at once and define all the vernaculars using strongly typed extension points; this has a considerable cost tho in terms of maintenability, so I dunno.

herbelin:
I have nothing special to say except that I see this PR only now and want to apologize for it. The case of `Coercion` is an interesting one.

ejgallego:
This PR has gone stale enough that we may even close it; however it is interesting nonetheless.

gares:
> This PR has gone stale enough that we may even close it; however it is interesting nonetheless.

@vbgl are the reamrks by @ejgallego clear to you? Do you intend to follow one of the two proposed paths?

My review still stands, I don't mind if it is possible at the ML level to attach a coercion attribute to a non coercible object. Actually it is the typical invariant for which types are overkilling IMO. In a system with no subtyping such as the one of OCaml it is going to backfire. The added value of separating metadata (such as being a coercion) from the syntax tree is much more important to me than being able to express at the ML level that a `Begin Section` is a coercion. We already have dynamic checks about unsupported attributes BTW.

vbgl:
Let’s close this. The implementation is wrong. I may come back to it when time allows.

PR #7648, opened 2018-05-31
Indicate in the doc that clearbody can take several idents

TheoWinterhalter:
**Kind:** documentation.

Change documentation of `clearbody` to account for the fact that it can take several identifiers.

- [ ] Entry added in CHANGES.

### DIFF
--- a/doc/sphinx/proof-engine/tactics.rst
+++ b/doc/sphinx/proof-engine/tactics.rst
@@ -893,17 +893,14 @@ quantification or an implication.
    This clears the hypothesis :n:`@ident` and all the hypotheses that depend on
    it.
 
-.. tacv:: clearbody @ident
+.. tacv:: clearbody {+ @ident}
    :name: clearbody
 
-   This tactic expects :n:`@ident` to be a local definition then clears its
-   body. Otherwise said, this tactic turns a definition into an assumption.
+   This tactic expects :n:`{+ @ident}` to be local definitions and clears their
+   respective bodies.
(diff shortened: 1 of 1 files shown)


doc/sphinx/proof-engine/tactics.rst
```diff
@@ -901,6 +893,18 @@ quantification or an implication.
    This clears the hypothesis :n:`@ident` and all the hypotheses that depend on
    it.
 
+.. tacv:: clearbody @ident
+   :name: clearbody
+
+   This tactic expects :n:`@ident` to be a local definition then clears its
```
Zimmi48:
```
:token:`ident`
```
should generally be preferred over
```
:n:`@ident`
```
(cf. https://github.com/coq/coq/tree/master/doc/sphinx#coq-roles)
and I'm not sure that the word "then" is appropriate in this case (there is no "if"). I know that you didn't write this but since you are moving these lines, why not improve them at the same time?

Zimmi48 (commented):
Thank you for this PR. I understand that you added a variant for `clearbody` because this was done similarly for `clear` but I would rather have generalized the grammar of the existing entry into `clearbody {+ @ident }` without creating a variant. I would have done similarly for `clear`, maybe even going to `clear {* @ident }` to include three variants at once. This is of course just a matter of personal taste.

Zimmi48:
Looks good, thanks!

maximedenes (approved):

maximedenes:
Thanks for contributing!

PR #79, opened 2015-06-25
Let the kernel assume that a (co-)inductive type is positive

aspiwack:
For the sake of experimentation, or just because it is an assumption no different in essence from an `Axiom`, it is useful to use a type that Coq does not recognise as positive.

The syntax I chose for this was `Assume [Positive] Inductive`. Another possibility may be `Assuming [Positive] Inductive`. The square bracket delimit a list, because of potential future things that could be assumed (also in Definitions) like `Assume [Guarded]` or `Assume [Guarded Wellsized]` (to ignore guard conditions and universe inconsistencies).

Here is a view of coqide with this feature:

assumepositive

Does anyone want to review the (minimal) changes to the kernel? (I'm fairly confident about them, I believe the test-suite thingies that I added to have a good coverage). Any advice or comment on the syntax?

PR #7651, opened 2018-05-31
[Work and discussion in progress] Add `Unguarded` syntax for (co)fixpoints and `Assumed Positive` for (co)inductive types.

SimonBoulier:
Disable guard check for (co)fixpoints when defined with `Ungarded` keyword.
We can now write:
```
Unguarded Fixpoint f (n : nat) : False
  := f n.

Unguarded Fixpoint ackermann (m n : nat) {struct m} : nat.
  destruct m. exact (S n).
  destruct n. exact (ackermann m 1).
  exact (ackermann m (ackermann (S m) n)).
Defined.

Compute (ackermann 3 4).

Print Assumptions ackermann.
(* ackermann is assumed to be guarded *)

Unguarded CoFixpoint fer (s : Stream nat) : Stream nat.
Proof.
  exact (fer s).
Defined.
```

**Kind:** feature.

- [x] Entry added in CHANGES.

I know that this PR is far from being perfect, any comment welcome.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -37,6 +37,9 @@ Vernacular Commands
 - Nested proofs may be enabled through the option `Nested Proofs Allowed`.
   By default, they are disabled and produce an error. The deprecation
   warning which used to occur when using nested proofs has been removed.
+- Guard check can be temporary disabled to define non terminating fixpoints
+  or non productive cofixpoints with the commands `Unguarded Fixpoints`
+  and `Unguarded CoFixpoint`.
 
 Coq binaries and process model
 
--- a/CHANGES
+++ b/CHANGES
@@ -38,8 +38,11 @@ Vernacular Commands
   By default, they are disabled and produce an error. The deprecation
   warning which used to occur when using nested proofs has been removed.
 - Guard check can be temporary disabled to define non terminating fixpoints
-  or non productive cofixpoints with the commands `Unguarded Fixpoints`
+  or non productive cofixpoints with the commands `Unguarded Fixpoint`
   and `Unguarded CoFixpoint`.
+- Positivity check can be temporary disabled to define non positive inductive
+  or coinductive types with the commands `Assumed Positive Inductive`
+  and `Assumed Positive CoInductive`.
 
 Coq binaries and process model
 
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -1141,10 +1141,16 @@ constructions.
 
    .. example::
 
+<<<<<<< 8a15241643eaf41b8b367f05b465fd90153437a2
       The recursive call may not only be on direct subterms of the recursive
       variable :g:`n` but also on a deeper subterm and we can directly write
       the function :g:`mod2` which gives the remainder modulo 2 of a natural
       number.
+=======
+The check of decreasing argument can be temporary disabled using ``Unguarded``.
+Warning: this can break the consistency of the system, use at your own risk.
+Unchecked fixpoint are printed by ``Print Assumptions``.
+>>>>>>> Fix documentation.
 
          Fixpoint mod2 (n:nat) : nat :=
          match n with
@@ -1260,6 +1266,7 @@ Definitions of recursive objects in co-inductive types
       mutually dependent methods.
 
    .. cmdv:: Unguarded CoFixpoint @ident {? @binders } {? : @type } := @term
+
       Define a cofixpoint without checking productivity.
       Warning: this can break the consistency of the system, use at your own risk.
       Unchecked cofixpoint are printed by ``Print Assumptions``.
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -508,12 +508,12 @@ The Vernacular
                       : | ( `ident` … `ident` : `term` ) … ( `ident` … `ident` : `term` )
    definition         : [Local] Definition `ident` [`binders`] [: `term`] := `term` .
                       : | Let `ident` [`binders`] [: `term`] := `term` .
-   inductive          : Inductive `ind_body` with … with `ind_body` .
-                      : | CoInductive `ind_body` with … with `ind_body` .
+   inductive          : [Assumed Positive] Inductive `ind_body` with … with `ind_body` .
+                      : | [Assumed Positive] CoInductive `ind_body` with … with `ind_body` .
    ind_body           : `ident` [`binders`] : `term` :=
                       : [[|] `ident` [`binders`] [:`term`] | … | `ident` [`binders`] [:`term`]]
    fixpoint           : [Unguarded] Fixpoint `fix_body` with … with `fix_body` .
-                      : | CoFixpoint `cofix_body` with … with `cofix_body` .
+                      : | [Unguarded] CoFixpoint `cofix_body` with … with `cofix_body` .
    assertion          : `assertion_keyword` `ident` [`binders`] : `term` .
    assertion_keyword  : Theorem | Lemma
                       : | Remark | Fact
@@ -740,6 +740,13 @@ Simple inductive types
       The types of the constructors have to satisfy a *positivity condition*
       (see Section :ref:`positivity`). This condition ensures the soundness of
       the inductive definition.
+
+   .. cmdv:: Assumed Positive Inductive @ident : {? @sort } := {? | } @ident : @type {* | @ident : @type }
+      :name: Assumed Positive Inductive
+
+      The positivity condition checking can be disabled using :cmd:`Assumed Positive
+      Inductive`. Warning: this can break the consistency of the system, use at your
+      own risk. Unchecked inductive types are printed by :cmd:`Print Assumptions`.
 
    .. exn:: The conclusion of @type is not valid; it must be built from @ident.
 
@@ -1045,6 +1052,13 @@ co-inductive definitions are also allowed.
    object of type :g:`(EqSt s1 s2)`. We will see how to introduce infinite
    objects in Section :ref:`cofixpoint`.
 
+.. cmdv:: Assumed Positive CoInductive @ident : {? @sort } := {? | } @ident : @type {* | @ident : @type }
+   :name: Assumed Positive CoInductive
+
+   The positivity condition checking can be disabled using :cmd:`Assumed Positive
+   CoInductive`. Warning: this can break the consistency of the system, use at your
+   own risk. Unchecked coinductive types are printed by :cmd:`Print Assumptions`.
+
 Definition of recursive functions
 ---------------------------------
 
@@ -1141,16 +1155,12 @@ constructions.
 
    .. example::
 
-<<<<<<< 8a15241643eaf41b8b367f05b465fd90153437a2
       The recursive call may not only be on direct subterms of the recursive
       variable :g:`n` but also on a deeper subterm and we can directly write
       the function :g:`mod2` which gives the remainder modulo 2 of a natural
       number.
-=======
-The check of decreasing argument can be temporary disabled using ``Unguarded``.
-Warning: this can break the consistency of the system, use at your own risk.
-Unchecked fixpoint are printed by ``Print Assumptions``.
->>>>>>> Fix documentation.
+
+      .. coqtop:: all
 
          Fixpoint mod2 (n:nat) : nat :=
          match n with
@@ -1184,10 +1194,12 @@ Unchecked fixpoint are printed by ``Print Assumptions``.
             end.
 
    .. cmdv:: Unguarded Fixpoint @ident @binders {? {struct @ident} } {? : @type } := @term
+      :name: Unguarded Fixpoint
 
-      The check of decreasing argument can be temporary disabled using ``Unguarded``.
-      Warning: this can break the consistency of the system, use at your own risk.
-      Unchecked fixpoint are printed by ``Print Assumptions``.
+      The check of decreasing argument can be temporary disabled using :cmd:`Unguarded Fixpoint`.
+      Warning: this can break the consistency of the system, use at your own risk. Decreasing
+      argument can still be specified but the decrease is not checked anymore. Unchecked fixpoint
+      are printed by :cmd:`Print Assumptions`.
 
       .. example::
          False can be proved in the following way.
@@ -1199,20 +1211,22 @@ Unchecked fixpoint are printed by ``Print Assumptions``.
 
       .. example::
 
-         Decreasing argument can still be specified but the decrease is not checked anymore.
-         Here is for instance the definition of the Ackermann function.
-
          .. coqtop:: all
 
-            Unguarded Fixpoint ackermann (m n : nat) {struct m} : nat.
-              destruct m. exact (S n).
-              destruct n. exact (ackermann m 1).
-              exact (ackermann m (ackermann (S m) n)).
-            Defined.
+            Unguarded Fixpoint ackermann (m n : nat) {struct m} : nat :=
+              match m with
+              | 0 => S n
+              | S m =>
+                  match n with
+                  | 0 => ackermann m 1
+                  | S n => ackermann m (ackermann (S m) n)
+                  end
+              end.
 
             Print Assumptions ackermann.
-            (* ackermann is assumed to be guarded *)
 
+         Note that the proper way to define the Ackermann function is to use
+         well-founded recursion (see. :cmd:`Program Fixpoint`).
 
 .. _cofixpoint:
 
@@ -1266,10 +1280,11 @@ Definitions of recursive objects in co-inductive types
       mutually dependent methods.
 
    .. cmdv:: Unguarded CoFixpoint @ident {? @binders } {? : @type } := @term
+      :name: Unguarded CoFixpoint
 
-      Define a cofixpoint without checking productivity.
-      Warning: this can break the consistency of the system, use at your own risk.
-      Unchecked cofixpoint are printed by ``Print Assumptions``.
+      Define a cofixpoint without checking productivity. Warning: this can
+      break the consistency of the system, use at your own risk. Unchecked
+      cofixpoint are printed by :cmd:`Print Assumptions`.
 
 .. _Assertions:
 
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -512,7 +512,7 @@ The Vernacular
                       : | CoInductive `ind_body` with … with `ind_body` .
    ind_body           : `ident` [`binders`] : `term` :=
                       : [[|] `ident` [`binders`] [:`term`] | … | `ident` [`binders`] [:`term`]]
-   fixpoint           : Fixpoint `fix_body` with … with `fix_body` .
+   fixpoint           : [Unguarded] Fixpoint `fix_body` with … with `fix_body` .
                       : | CoFixpoint `cofix_body` with … with `cofix_body` .
    assertion          : `assertion_keyword` `ident` [`binders`] : `term` .
    assertion_keyword  : Theorem | Lemma
@@ -1146,8 +1146,6 @@ constructions.
       the function :g:`mod2` which gives the remainder modulo 2 of a natural
       number.
 
-      .. coqtop:: all
-
          Fixpoint mod2 (n:nat) : nat :=
          match n with
          | O => O
@@ -1178,6 +1176,37 @@ constructions.
             | leaf b => 1
             | cons t f' => (tree_size t + forest_size f')
             end.
+
+   .. cmdv:: Unguarded Fixpoint @ident @binders {? {struct @ident} } {? : @type } := @term
+
+      The check of decreasing argument can be temporary disabled using ``Unguarded``.
+      Warning: this can break the consistency of the system, use at your own risk.
+      Unchecked fixpoint are printed by ``Print Assumptions``.
+
+      .. example::
+         False can be proved in the following way.
+
+         .. coqtop:: all
+
+            Unguarded Fixpoint f (n : nat) : False
+              := f n.
+
+      .. example::
+
+         Decreasing argument can still be specified but the decrease is not checked anymore.
+         Here is for instance the definition of the Ackermann function.
+
+         .. coqtop:: all
+
+            Unguarded Fixpoint ackermann (m n : nat) {struct m} : nat.
+              destruct m. exact (S n).
+              destruct n. exact (ackermann m 1).
+              exact (ackermann m (ackermann (S m) n)).
+            Defined.
+
+            Print Assumptions ackermann.
+            (* ackermann is assumed to be guarded *)
+
 
 .. _cofixpoint:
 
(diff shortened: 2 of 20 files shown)


CHANGES
```diff
@@ -37,6 +37,9 @@ Vernacular Commands
 - Nested proofs may be enabled through the option `Nested Proofs Allowed`.
   By default, they are disabled and produce an error. The deprecation
   warning which used to occur when using nested proofs has been removed.
+- Guard check can be temporary disabled to define non terminating fixpoints
+  or non productive cofixpoints with the commands `Unguarded Fixpoints`
```
Zimmi48: typo: spurious s

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -515,7 +515,7 @@ The Vernacular
                       : | CoInductive `ind_body` with … with `ind_body` .
    ind_body           : `ident` [`binders`] : `term` :=
                       : [[|] `ident` [`binders`] [:`term`] | … | `ident` [`binders`] [:`term`]]
-   fixpoint           : Fixpoint `fix_body` with … with `fix_body` .
+   fixpoint           : [Unguarded] Fixpoint `fix_body` with … with `fix_body` .
```
Zimmi48: You should also add this on the CoFixpoint line below.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1069,7 +1069,7 @@ one can define the addition function as :
    | S p => S (add p m)
    end.
 
-The ``{struct`` :token:`ident```}`` annotation may be left implicit, in this case the
+The ``{struct`` :token:`ident` ``}`` annotation may be left implicit, in this case the
```
Zimmi48: Note that this is already fixed in #7537 and the best fix goes further than this.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1153,6 +1153,37 @@ reduction will only be performed when the argument in position of the
 decreasing argument (which type should be in an inductive definition)
 starts with a constructor.
 
+The check of decreasing argument can be temporary disabled using ``Unguarded``.
+Warning: this can break the consistency of the system, use at your own risk.
```
Zimmi48: This documentation is not done in the recommended style, the documentation must come after the variant grammar and must be indented. Since this chapter is heavily reindented by #7537 anyways, I suggest you rebase on top or wait for it to get merged. In any case, you will have so many conflicts in this file that the best way to solve them will be to copy / paste this part.

Zimmi48 (commented):
Since we can also do fixed points using `Theorem`, `Lemma`, etc, would it make sense to also add this new syntax for these commands?

doc/sphinx/language/gallina-specification-language.rst
SimonBoulier: Ok thanks, I will do this after merge then.

vernac/comFixpoint.ml
```diff
@@ -243,7 +243,11 @@ let interp_fixpoint ~cofix l ntns =
   check_recursive true env evd fix;
   (fix,pl,Evd.evar_universe_context evd,info)
 
-let declare_fixpoint local poly ((fixnames,fixdefs,fixtypes),pl,ctx,fiximps) indexes ntns =
+(* guarded: Some b -> set check_guarded to b, None -> does not change it *)
+let declare_fixpoint guarded local poly ((fixnames,fixdefs,fixtypes),pl,ctx,fiximps) indexes ntns =
```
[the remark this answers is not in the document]
SimonBoulier: I named guarded, but do you suggest I also name the other arguments?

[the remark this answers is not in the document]
ejgallego: Maybe that would be a good idea, what do other devs think?

Zimmi48:
> Since we can also do fixed points using Theorem, Lemma, etc, would it make sense to also add this new syntax for these commands?

I withdraw this because I'm actually doubtful that doing fixed points using `Theorem` is a good idea anyways.

ppedrot:
***Please. Stop. Inflating. The. Definition. Modifier. Syntax. Again.***

There have been a shitload of arguments against definition modifiers, and it all boils down that it doesn't scale. I am too tired to repeat them here and I think there are pointers in #79.  I am fundamentally against this PR.

ppedrot (changes requested):
To expand on my above comment, I am even surprised that people that were (I think) at the heated Coq workgroup where we discussed the issue comment on that PR as if it was OK in principle, and that they did not try to deter @SimonBoulier from implementing it at the CIW... Is that the famous "History repeats twice, first as a tragedy, second as a farce" effect?

Zimmi48:
@SimonBoulier The documentation PR has been merged so you might want to rebase now.

@ppedrot I completely disagree with your argument to reject this change. This argument has been stupidly used to delay #79 for three years now. The problem you are pointing at is a well known syntax problem and must be addressed independently without preventing people to work on new features in the meantime. Attributes are now on their way thanks to @vbgl and will probably be finished for 8.9 so it means that the current syntax will probably never make it into a released version as-is. But even if it wasn't the case, that wouldn't be a good reason to delay the availability of this feature.

ejgallego:
I like this feature and its current syntax. We are taking attributes as a kind of "silver bullet" , but the more I think about them the less convinced I am they are going to bring significant benefits.

Attributes are _hard_ to handle, and worse, they tend to override all kind of static checks due to their "generic" nature. So IMVHO we shouldn't delay this PR just for that.

maximedenes:
> So IMVHO we shouldn't delay this PR just for that.

I tend to agree with that part.

ppedrot:
> The problem you are pointing at is a well known syntax problem and must be addressed independently without preventing people to work on new features in the meantime. 

Wrong. Syntax is non-local by nature, so this can affect people that are unrelated to this change. Plus in the current implementation it is a hell to get it right. There is no way to compute the syntax conflict because CAMLP5, but we would be surprised if this were possible. This is the kind of area where conservativity should be of utmost importance, because it will be hard to backtrack on that.

But like, I don't understand why it is taking THREE YEARS. Like I can implement a syntax that scales properly in literally half a day. Should I do that or what?

ppedrot:
And also,

> But even if it wasn't the case, that wouldn't be a good reason to delay the availability of this feature.

We've already discussed that, and the answer is: write a goddamn plugin.

Zimmi48:
> But like, I don't understand why it is taking THREE YEARS. Like I can implement a syntax that scales properly in literally half a day. Should I do that or what?

Seriously? And you didn't do it?
Anyways, now your help on this is not needed anymore since it is well underway thanks to @vbgl.

ppedrot:
> Seriously? And you didn't do it?

This ended up in the sinkhole of workgroups and PR bikeshedding, as all syntactic considerations do. But yes, luckily @vbgl has started doing it with a wide approval, although I am still curious to see the moment where we start discussing on the actual syntax.

ejgallego:
IMO "Write a plugin" is hardly an answer to most problems, but rather a good way to add some maintenance burden and testing overhead. This seems to me like a well-delimited functionality that applies to two core vernaculars.

ejgallego (changes requested):
I am convinced now that this property is local to the vernaculars and thus doesn't belong in an attribute.

vernac/vernacinterp.mli
```diff
@@ -15,6 +15,7 @@ type deprecation = bool
 type atts = {
   loc : Loc.t option;
   locality : bool option;
+  check_guard : bool option; (* None -> use global setting (true by default); Some -> override *)
   polymorphic : bool;
```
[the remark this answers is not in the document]
ejgallego: I mean this should not be an attribute, but an entry in the corresponding `vernac_expr` constructor.

Zimmi48:
It seems Emilio that you have a diverging opinion of what should be an attribute than what was in the CEP and what is being implemented. IMO the great benefit of attributes is that they make the parsing phase more tractable by moving some of the complexity / error handling from the parsing file to the interpretation phase.

gares:
> this property is local to the vernaculars

Can you explain what you mean?

gares:
I like the feature, I'm fine with the current "prefix" syntax and the plan to turn it into one of the many tags/attributes. I dislike the word `Unguarded`, it is too obscure. All users understand "termination checking" or "total functions" but "guardedness condition" is, frankly, even hard to pronounce.

One thing is quite fishy to my eyes, please @SimonBoulier answer the following concern.
Fixpoints are not named, eg Fixpoint is just `Definition` + `fix`. I wonder how print assumption can detect the unguarded fixpoint if it is copied in my proof by unifying a goal with the type of a lemma (containing the unguarded fixpoint) and unification has unfolded the fixpoint (i.e. lose his name). Is the `fix` term constructor containing the `tainted` bit (which I doubt given how little files did change) or is the `Definition` containing it?
I'm not 100% sure I can prove False undetected, but maybe @maximedenes can ;-)

ejgallego:
Well indeed maybe "convinced" is far too strong. I tend to have a more "strongly typed" vision of commands, where the type for each vernacular reflects what the command depends on.

I never understood the attributes proposal in a very wide scope. To my point of view attributes are something that are, in principle, data pplicable to most commands, such as location, locality, or preprocessing information. However here we have only two commands that can understand the unguarded bit so I am not sure that should be an attribute. Certainly when introducing the attr record I was not thinking of that case.

It would seem to me that such information would be an attribute if attached to terms, in such a way that would mean "don't check termination" in the checking of that term.

But I dunno, I guess I'm certainly fine if people thinks that we should use a vernac attribute for that, but I fear that we end up with a record which is a bit messy in the sense of having weak "uniformity" properties. Maybe a good solution is indeed to further refine "attributes" as to have more representative classes. For example we could have "definition_attr" for attributes specific to definition-like commands, "scoped_obj" attributes for commands that support scope, etc...

Anyways take my opinion more as an evolving reflection than as a strong statement on the solution to choose.

ejgallego:
I would certainly enjoy github refining the "red" review button as to have a "strong" and "weak" request for changes.

gares:
OK, but I still don't get exactly what your review talks about. Here I see
- a feature (letting one switch off termination checking)
- a syntax in the current style and a proposal (switch to a uniform attribute syntax)
- an implementation

I read @ppedrot and @ejgallego reviews as "No to the syntax". I fully understand this POV, even if I've a different opinion. *but* this PR is more than just some syntax, namely it adds a feature and an provides an implementation. Do you guys agree that the feature is OK? Do you have remarks about the implementation?

ejgallego:
Ive only looked at the connection between the syntax and the implementation so far, but not at the implementation itself.

Zimmi48:
> One thing is quite fishy to my eyes, please @SimonBoulier answer the following concern.
Fixpoints are not named, eg Fixpoint is just Definition + fix. I wonder how Print Assumptions can detect the unguarded fixpoint if it is copied in my proof by unifying a goal with the type of a lemma (containing the unguarded fixpoint) and unification has unfolded the fixpoint (i.e. lose his name). Is the fix term constructor containing the tainted bit (which I doubt given how little files did change) or is the Definition env entry that contains it?

Simon explained that this new feature doesn't go as far as he would have liked because the check disabling only occurs at definition time. So if the fixpoint ended up in a proof it would have to be typechecked again and this would fail.

Zimmi48:
> Certainly when introducing the attr record I was not thinking of that case.

Indeed, it is funny that you were the one that introduced this attr record but had a completely different understanding of what an attribute was compared to most other developers. Remember how long it took to merge this PR? That was in part due to the very high confusion surrounding its purpose. Contrary to the PR that introduced the attr records, the attribute feature has had a CEP which is clear and detailed so I would defer any further design discussion there.

> But I dunno, I guess I'm certainly fine if people thinks that we should use a vernac attribute for that, but I fear that we end up with a record which is a bit messy in the sense of having weak "uniformity" properties. Maybe a good solution is indeed to further refine "attributes" as to have more representative classes. For example we could have "definition_attr" for attributes specific to definition-like commands, "scoped_obj" attributes for commands that support scope, etc...

I would personally push for an attribute record that is extensible, even if it requires some form of dynamic type checking, and I've said so to @vbgl in private.

> It would seem to me that such information would be an attribute if attached to terms, in such a way that would mean "don't check termination" in the checking of that term.

This would be better than the current proposal but would also require a much deeper refactoring that I'm not sure that Simon is ready to undertake.

> I would certainly enjoy github refining the "red" review button as to have a "strong" and "weak" request for changes.

I think you should just dismiss your "request for changes" to transform it into a simple "comment review".

> a syntax in the current style and a proposal (switch to a uniform attribute syntax)

The current proposal uses the attribute system internally but not on the user-side because there is no syntax for attributes yet.

library/decl_kinds.ml
```diff
@@ -12,6 +12,8 @@
 
 type discharge = DoDischarge | NoDischarge
 
+type check_guard = bool option
```
Zimmi48: I don't like `bool option`. Why don't you create a three-valued type like is done around? Why do you need three values BTW?

printing/printer.ml
```diff
@@ -915,9 +915,9 @@ let pr_assumptionset env sigma s =
       | Constant kn ->
           safe_pr_constant env kn ++ safe_pr_ltype typ
       | Positive m ->
-          hov 2 (MutInd.print m ++ spc () ++ strbrk"is positive.")
+          hov 2 (MutInd.print m ++ spc () ++ strbrk"is assumed to be positive")
       | Guarded kn ->
-          hov 2 (safe_pr_constant env kn ++ spc () ++ strbrk"is positive.")
+          hov 2 (safe_pr_constant env kn ++ spc () ++ strbrk"is assumed to be guarded")
```
Zimmi48: If this is not affecting any output test, could you add one? I'd like to understand in which situation these messages are seen.

stm/vernac_classifier.ml
```diff
@@ -202,7 +202,7 @@ let classify_vernac e =
       let poly = List.fold_left (fun poly f ->
           match f with
           | VernacPolymorphic b -> b
-          | (VernacProgram | VernacLocal _) -> poly
+          | (VernacProgram | VernacLocal _ | VernacGuarded _) -> poly
```
Zimmi48: This is not going to scale if we get many attributes. You should just use `_ ->` here.

vernac/comFixpoint.mli
```diff
@@ -80,12 +78,13 @@ val interp_fixpoint :
 (** Registering fixpoints and cofixpoints in the environment *)
 (** [Not used so far] *)
 val declare_fixpoint :
-  locality -> polymorphic ->
+  check_guard -> locality -> polymorphic ->
```
Zimmi48: The type signature doesn't correspond to the implementation (compilation is broken).
Can you make these three first parameters named ones since you are changing the API anyways?
BTW, what is this "Not used so far" comment above?

vernac/comFixpoint.mli
```diff
@@ -19,12 +19,10 @@ open Vernacexpr
 (** Entry points for the vernacular commands Fixpoint and CoFixpoint *)
 
 val do_fixpoint :
-  (* When [false], assume guarded. *)
-  locality -> polymorphic -> (fixpoint_expr * decl_notation list) list -> unit
+  check_guard -> locality -> polymorphic -> (fixpoint_expr * decl_notation list) list -> unit
```
Zimmi48: The removal of the comment is only making things less clear because the semantics of `check_guard` is non-obvious.

vernac/g_vernac.ml4
```diff
@@ -78,9 +78,15 @@ GEXTEND Gram
     ]
   ;
   vernac:
+    [ [ IDENT "Unguarded";   (f, v) = vernac_loc -> (VernacGuarded false  :: f, v)
+      (* Disabled to not confuse the user: for the moment there is not way to disable guard check globally. *)
+      (* | IDENT "Guarded"; (f, v) = vernac_loc -> (VernacGuarded true :: f, v) *)
```
Zimmi48: Why don't you define this option? It should be really easy to add one. Indeed, this had me confused about the reasons for a three-valued flag.

vernac/vernacentries.ml
```diff
@@ -2185,6 +2186,13 @@ let check_vernac_supports_polymorphism c p =
     | VernacExtend _ | VernacUniverse _ | VernacConstraint _) -> ()
   | Some _, _ -> user_err Pp.(str "This command does not support Polymorphism")
 
+(* Vernaculars that take a guarded flag *)
+let check_vernac_supports_guarded c p =
+  match p, c with
+  | None, _ -> ()
+  | Some _, (VernacFixpoint _ | VernacCoFixpoint _) -> ()
+  | Some _, _ -> user_err Pp.(str "This command does not support Guarded")
```
Zimmi48: @ejgallego Cf. here and the above function about the Polymorphic flag to better understand the attribute system.

ejgallego:
> the attribute feature has had a CEP which is clear and detailed so I would defer any further design discussion there.

I am not sure which CEP you are referring to. If that's https://github.com/aspiwack/ceps/blob/attributes/text/attributes.md , it is very far from detailed. At most a sketch of wishes as many important details seem missing, including some key crucial technical ones. In particular, it doesn't address vernac execution at all, but only syntax. Also, the PR seems only to discuss attributes for declarations and inductives, I am very OK with them as long as they are done with the proper scoping.

In fact, my attribute record seems fairly orthogonal to the CEP.

> I would personally push for an attribute record that is extensible, even if it requires some form of dynamic type checking, and I've said so to @vbgl in private.

Well that system is just a map from `string` to `string` and while I am not opposed to that I do fail to see how having this can make the vernacular layer better, at least when we are talking about core functionality. 

I wish indeed we had a system to better classify commands, as to be able to say "this command supports locality, termination checking, and program mode". We could either use variants, GADT, or objects to achieve this.

Note that the CEP talks more about plugins, but here it seems clear to me that this guarded flag is better stored in the `VernacDefinition` constructor.

> I think you should just dismiss your "request for changes" to transform it into a simple "comment review".

In this case I am still feeling like strongly recommending to remove "VernacGuarded" case in favor of a command-specific entry, so I think I'll keep it for a while. I just want to highlight that I am not talking categorically here nor I am making any general statement about attributes, but I'd like people to reflect about it.

> The current proposal uses the attribute system internally but not on the user-side because there is no syntax for attributes yet.

So indeed it seems to me that the dual would be what was proposed and then would maybe help alleviate the parsing "problems". Parsing can be done in a more convenient way, however internally the new `Guarded` constructor doesn't make a lot of sense IMVHO.

vernac/vernacentries.ml
ejgallego: I am not sure what do you mean, I do understand this mechanism as I basically started it and IIRC I suggested / discussed the current design with Maxime some time ago. Code like this seems quite fragile, even if I am the one to blame and it is not even usable by vernacs declared using the interp table. It seems clear to me that we should not foster this wrapping style but indeed declaration-only attributes should be properly scoped.

gares:
> Simon explained that this new feature doesn't go as far as he would have liked because the check disabling only occurs at definition time. So if the fixpoint ended up in a proof it would have to be typechecked again and this would fail.

Right! Much better this way ;-)

ejgallego:
@SimonBoulier personally I'd say that if you remove the `VernacGuarded` "attribute" and instead put this info in the `Definition` constructor itself that should be merged without too much problems.

Indeed it seems to me that quite a few design work on what "attributes" are and how are they gonna be implemented is needed, but IMHO that should be orthogonal to this PR.

Zimmi48:
Except that this is precisely what @ppedrot opposed to, isn't it?

ejgallego:
The way I understand it is that @ppedrot complains about the syntax, I am referring to the AST itself.

At least to me it is clear that the AST needs a bit more of design work so I think that what I suggest is a reasonable compromise in order not to get this PR stuck.

About the syntax, I have no particular strong opinion.

ppedrot:
To be clear, I'm just opposed to the syntax extension. How hackish the implementation is can be frowned upon from an engineering point of view, but this is one order of magnitude less problematic as it does not appear directly to the user.

Quizz: without looking at the code, which permutation makes the syntax parsable in the current PR?
```
{Unguarded, Local, Polymorphic, Program} Fixpoint f (n : nat) := tt.
```

mattam82:
I think Unguarded/General fxpoint could have even a larger scope than Program, and apply to Lemma, Inductive as well, ie any command where an unguarded fixpoint might be typechecked. In the current codebase there’s no parsing support for attributes so I didn’t see a problem with the implementation. When they’re available we’ll use them. I suppose there will be a mechanism to check the support of an attribute by a given command, @vbgl @maximedenes ?

maximedenes:
> I suppose there will be a mechanism to check the support of an attribute by a given command, @vbgl @maximedenes ?

Yes, I believe that is the plan.

Zimmi48:
> I think Unguarded/General fxpoint could have even a larger scope than Program, and apply to Lemma, Inductive as well, ie any command where an unguarded fixpoint might be typechecked.

For `Inductive`, do you mean you mean that `Unguarded` should give access to the feature that was implemented in #79? In that case, I wonder if `Unguarded` is the right generic terminology.

For `Lemma`, `Theorem`, etc, yes, this is what I suggested in https://github.com/coq/coq/pull/7651#pullrequestreview-124815588. However, I withdrew this in https://github.com/coq/coq/pull/7651#issuecomment-393805776 because I'm doubtful about the interest of using `Theorem` to define a (co-)fixpoint. I wonder if we shouldn't deprecate this feature instead. WDYT?

SkySkimmer:
>For Inductive, do you mean you mean that Unguarded should give access to the feature that was implemented in #79? In that case, I wonder if Unguarded is the right generic terminology.

No, it's that you could write `Unguarded Inductive foo := bar : (fix f (x:unit) : empty := f x) -> foo.`

nomeata:
This might be very useful to me! But I often work with `Definition` and local `fix`, and from the description it is unclear to me whether there is a way to mark a local `fix` as `Unguarded`. Will that be supported?

Zimmi48:
It was discussed and this PR doesn't go as far because it doesn't change the AST of terms (we can't attach `Unguarded` to a term). However, I guess the options proposed in https://github.com/SimonBoulier/TypingFlags should do what you want. BTW @SimonBoulier what about first proposing to integrate these options to Coq. This was the approach recently followed by @jashug in #7703 to avoid the debates around attributes.

SimonBoulier:
> what about first proposing to integrate these options to Coq

That's my plan! I do this as soon as my thesis is written  :-)

herbelin (commented):
OK to me as far as I'm concerned.

gares (approved):
wrt the classifier

maximedenes (changes requested):
This can now use attributes.

maximedenes:
It seems to me that this PR could be rebased on top of attributes, now that we have them.

Zimmi48:
I think this is simply waiting on @SimonBoulier finishing writing their thesis first.

SimonBoulier:
Sorry I am not very available those days but I will do it.
What is the deadline for the 8.9? October, the 1st?

Zimmi48:
October 1st is the date of the beta release so it will be too late for sure, but it is already quite late. If it is merged now, it goes in 8.10 by default unless developers convince the release manager how important it is to include this in 8.9.

SimonBoulier:
Ok, it will be for 8.10 then.

SimonBoulier:
Superseded by  #9004.

PR #7652, opened 2018-05-31
Explicitly require python2 in python scripts in tools/

Armael:
On my system, `python` is pointing to `python3`. This explicitly requires `python2` in the python scripts in tools/. Otherwise, these scripts do not work when called with python 3.

### DIFF
--- a/tools/make-both-single-timing-files.py
+++ b/tools/make-both-single-timing-files.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 import sys
 from TimeFileMaker import *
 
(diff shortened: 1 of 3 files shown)


ejgallego:
Umm, I'd rather have things working with python3. Would it be very hard to do so?

Zimmi48:
I suppose we could merge this for 8.8.1 and port the scripts to python3 in 8.9+beta1. (BTW, supporting python2 is useless, that should only be done when there are good reasons for it. It's not the case here.)

JasonGross (approved):
I don't presently have the time to port these to python3, but my guess is that if you update the file/Unicode handing, that'll be the only change necessary.  I coded them in python2 mostly from habit.

PR #7655, opened 2018-05-31
Refactor parsing rules for Hint Resolve -> and Hint Resolve <-

Armael:
Joint work with @Zimmi48 .

This PR refactors the parsing rules for `Hint Resolve -> ...` and `Hint Resolve <- ...`. Previously there were 4 dedicated rules in extratactics.ml4 and 1 ad-hoc rule in g_proofs.ml4; after the refactoring only 2 rules remain, written in the same place as other hints. 

We could go further with the refactoring but we believe this is a useful first step.


ppedrot (approved):
LGTM. Is there a reason why this was defined as a separate entry before?

ejgallego (approved):

Zimmi48:
We asked @Herbelin and he didn't think their was a particular reason. We haven't looked at the git history?

ppedrot:
@mattam82 any comments on this PR before I merge it?

Armael:
FWIW I asked @mattam82 during the CIW if he was OK with this change, and the answer was "yes".

Armael:
Thanks!

PR #7656, opened 2018-05-31
CI for QuickChick and ext-lib

lemonidas:
**Kind:** infrastructure.

### DIFF
--- a/dev/ci/ci-ext-lib.sh
+++ b/dev/ci/ci-ext-lib.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+ci_dir="$(dirname "$0")"
+
+# This script could be included inside other ones
+# Let's avoid to source ci-common twice in this case
+if [ -z "${CI_BUILD_DIR}" ];
+then
+    . "${ci_dir}/ci-common.sh"
+fi
+
+ext_lib_CI_DIR="${CI_BUILD_DIR}/ExtLib"
+
+git_checkout "${ext_lib_CI_BRANCH}" "${ext_lib_CI_GITURL}" "${ext_lib_CI_DIR}"
+
+( cd "${ext_lib_CI_DIR}" && make && make install)
--- a/dev/ci/ci-quickchick.sh
+++ b/dev/ci/ci-quickchick.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+ci_dir="$(dirname "$0")"
+
+# This script could be included inside other ones
+# Let's avoid to source ci-common twice in this case
+if [ -z "${CI_BUILD_DIR}" ];
+then
+    . "${ci_dir}/ci-common.sh"
+fi
+
+quickchick_CI_DIR="${CI_BUILD_DIR}/Quickchick"
+
+install_ssreflect
+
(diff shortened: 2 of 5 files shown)


ejgallego (approved):
LGTM

ejgallego:
Umm, @lemonidas , the QuickChick makefile seems to be returning `1`, can you check what's going on? As we discussed, `make ci-quickchick` on an updated coq build should do the work.

Zimmi48:
If I understand correctly, QuickChick's master branch tracks Coq master and QuickChick's 8.8 branch tracks Coq v8.8. Note that the README is saying that the master branch depends on Coq 8.7 so this should be fixed.

dev/ci/ci-quickchick.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+ci_dir="$(dirname "$0")"
+
+# This script could be included inside other ones
+# Let's avoid to source ci-common twice in this case
+if [ -z "${CI_BUILD_DIR}" ];
```
SkySkimmer: Don't bother with this conditional.

dev/ci/ci-ext-lib.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+ci_dir="$(dirname "$0")"
+
+# This script could be included inside other ones
+# Let's avoid to source ci-common twice in this case
+if [ -z "${CI_BUILD_DIR}" ];
```
SkySkimmer: Don't bother with this conditional.

dev/ci/ci-quickchick.sh
ejgallego: Note that this is done already in some files, hence the origin from the cut-and-paste.

ejgallego:
Needs to be rebased on top of #7680

ejgallego:
This is still broken, strange, as the log doesn't seem very informative. Are you sure your makefile is safe in the use of `make -j`. It is easy to write `Makefiles` not safe when enabling a paralell build.

SkySkimmer:
QuickChick is using the record-in-constructor ocaml feature which isn't available in 4.02.3:
https://github.com/QuickChick/QuickChick/blob/c513d3178f0fad8c8351438fce97083b283a3ab8/quickChickTool/quickChickToolTypes.ml#L15-L27

dev/ci/ci-quickchick.sh
SkySkimmer: >Note that this is done already in some files, hence the origin from the cut-and-paste.

That's no reason to keep the copy when flagged in review though.

ejgallego:
Oh, thanks @SkySkimmer , this explains why I couldn't reproduce. [looking forward to having build-envs for the CI]

@lemonidas , can you then use the `ci-template-flambda` template in the gitlab file?

[We could also add a `ci-template-edge` I guess]

dev/ci/ci-quickchick.sh
lemonidas: Do you want me to remove the entire conditional altogether or just remove the condition and leave the "then" branch body?

PR #7660, opened 2018-05-31
Add codeowner for timing python scripts

JasonGross:
**Kind:** infrastructure.

As per https://github.com/coq/coq/pull/7652#issuecomment-393668390


maximedenes (approved):
We would need a secondary maintainer too, but ok for now.

PR #7649, opened 2018-05-31
Remove some dead code in class_tactics.ml

Armael:

PR #7661, opened 2018-06-01
Non-backtracking hints

Armael:
This is joint work with @Zimmi48 and @MSoegtropIMC .
~~This is based on top of #7655 and #7649 which should be merged first.~~

**Kind:** feature.

This pull-request proposes to add a new `NonBacktracking` flag for auto hints (more specifically, for `Hint Resolve`, `Hint Immediate` and `Hint Extern`). For example, to declare a Hint Resolve as nonbacktracking, one can write:

```NonBacktracking Hint Resolve foo : mybase```

Support for the flag has been implemented in `typeclasses eauto`.

The semantics of this flag is as follows: during proof search, if a nonbacktracking hint is applied with success, then alternative possible hints are thrown away and never considered -- the proof search never "backtracks" the application of the nonbacktracking hint. 

The intent of adding this flag is that it seems particularly useful to perform forward reasoning, or more generally for invertible steps (e.g. saturating the context). Backtracking the application of such steps is useless and easily leads to exponential blowup of the search.

Work left to be done:
- [ ] Update the documentation
- [ ] Clarify whether the flag makes sense for Hint Unfold
- [ ] Update the debugging messages of typeclasses eauto
- [ ] Add an entry in CHANGES.

### DIFF
--- a/test-suite/success/CommitHint.v
+++ b/test-suite/success/CommitHint.v
@@ -1,0 +1,24 @@
+
+Class E (a b : nat) := e{}.
+
+Class C (a b : nat) := c{}.
+
+Class D (a:nat) := d{}.
+
+Instance d2 : D 2 := {}.
+
+Instance c2 : C 0 2 | 10 := {}.
+
+Instance exy (a b:nat) `{C a b} `{D b} : E a b := {}.
+
+Module Back.
+  Instance c0 : C 0 0 | 1 := {}.
+  Type _ : E _ _.
+End Back.
+
+Module NoBack.
+  Type _ : E _ _.
+  Instance c0 : C 0 0 ! 1 := {}.
+  Fail Type _ : E _ _.
+  Type _ : C 0 0.
+End NoBack.
--- a/pretyping/typeclasses.ml
+++ b/pretyping/typeclasses.ml
@@ -8,7 +8,6 @@
 (*         *     (see LICENSE file for the text of the license)         *)
 (************************************************************************)
 
-(*i*)
 open Names
 open Globnames
 open Term
@@ -19,12 +18,16 @@ open Util
 open Typeclasses_errors
 open Context.Rel.Declaration
 
-(*i*)
+type hint_commit =
+  | NoCommit (** the default *)
+  | Commit (** do not try to backtrack the application of this hint *)
 
 (* Core typeclasses hints *)
-type 'a hint_info_gen =
-    { hint_priority : int option;
-      hint_pattern : 'a option }
+type 'a hint_info_gen = {
+  hint_priority : int option;
+  hint_pattern : 'a option;
+  hint_commit : hint_commit;
+}
 
 type hint_info = (Pattern.patvar list * Pattern.constr_pattern) hint_info_gen
 
--- a/pretyping/typeclasses.mli
+++ b/pretyping/typeclasses.mli
@@ -15,10 +15,16 @@ open Environ
 
 type direction = Forward | Backward
 
+type hint_commit =
+  | NoCommit (** the default *)
+  | Commit (** do not try to backtrack the application of this hint *)
+
 (* Core typeclasses hints *)
-type 'a hint_info_gen =
-    { hint_priority : int option;
-      hint_pattern : 'a option }
+type 'a hint_info_gen = {
+  hint_priority : int option;
+  hint_pattern : 'a option;
+  hint_commit : hint_commit;
+}
 
 type hint_info = (Pattern.patvar list * Pattern.constr_pattern) hint_info_gen
 
--- a/tactics/class_tactics.ml
+++ b/tactics/class_tactics.ml
@@ -516,11 +516,11 @@ let make_resolve_hyp env sigma st only_classes pri decl =
             (List.map_append
              (fun (path,info,c) ->
               let h = IsConstr (EConstr.of_constr c,Univ.ContextSet.empty) [@ocaml.warning "-3"] in
-              make_resolves env sigma ~name:(PathHints path) info NoCommit ~check:true ~poly:false h)
+              make_resolves env sigma ~name:(PathHints path) info ~check:true ~poly:false h)
                hints)
         else []
       in
-        (hints @ make_resolves env sigma pri NoCommit ~name ~check:false ~poly:false (IsGlobRef id))
+        (hints @ make_resolves env sigma pri ~name ~check:false ~poly:false (IsGlobRef id))
     else []
 
 let make_hints g (modes,st) only_classes sign =
--- a/tactics/class_tactics.ml
+++ b/tactics/class_tactics.ml
@@ -311,7 +311,7 @@ let rec e_trivial_fail_db only_classes db_list local_db secvars =
     begin fun gl ->
     let tacs = e_trivial_resolve db_list local_db secvars only_classes
                                  (pf_env gl) (project gl) (pf_concl gl) in
-      tclFIRST (List.map (fun (x,_,_,_,_) -> x) tacs)
+      tclFIRST (List.map (fun (x,_,_,_,_,_) -> x) tacs)
     end
   in
   let tacl =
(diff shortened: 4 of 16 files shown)

### CODE REVIEW

**Comment by ejgallego** [comment]
A minor note about terminology, in logic programming I've seldom seen `NonBacktracking`, but rather `Deterministic` or `Commited`, and indeed it sounds a bit weird to me.

Also, is there an overview of the roadmap and design of instance search somewhere? I seem to recall that there was some discussion about cuts already.

**Comment by Zimmi48** [comment]
> A minor note about terminology, in logic programming I've seldom seen NonBacktracking, but rather Deterministic or Commited, and indeed it sounds a bit weird to me.

We were not sure about the `NonBacktracking` terminology. Maybe `Committed` sounds better indeed.

> Also, is there an overview of the roadmap and design of instance search somewhere? I seem to recall that there was some discussion about cuts already.

Not AFAIK.
The discussion about cuts is at coq/ceps#30 and #6285. But here we are not doing a cut since all the previous backtracking point are kept. We are just making sure that we do not create a new backtracking point.
My personal roadmap is not about instance search but about proof search in general. I want to finish #721, rename a few commands, options and file names to make it clear that the proof search is not just for typeclasses, and refactor the documentation.

**Comment by ejgallego** [comment]
Thanks for the summary Théo.

> we do not create a new backtracking point.

Well, that's usually done in LP using a cut, isn't it? Is there not overlap in functionality there with #6285 ? Maybe @gares can also comment.

**Comment by MSoegtropIMC** [comment]
@ejgallego : my Prolog is a bit rusted, but afair in Prolog a cut means to remove all existing backtracking points - in a sense confirming what has been achieved so far and discarding all other alternatives seen in the past.

The semantic of Théo's and Armael's work is to say "in case this hint applies, don't consider the option of not applying it - just always apply it". Backtracking points from the past are not changed in any way.

**Comment by MSoegtropIMC** [comment]
P.S.: in a sense "NoBacktracking" is really confusing, because NoBacktracking hints can still be backtracked. If a previous backtracking point is considered, the NoBacktracking hints applied in between will be backtracked. They will just not be backtracked on their own right.

**Comment by MSoegtropIMC** [comment]
P.P.S.: I guess we really need a good example - I am working on it.

**Comment by ejgallego** [comment]
> in case this hint applies, don't consider the option of not applying it - just always apply it

what if two different hints apply?

In Prolog, it is very common to write:
``` 
a(Pat) :- !, b.
```
for example.

**Comment by MSoegtropIMC** [comment]
> what if two different hints apply?

Then both are applied in an arbitrary order - and this order is then fixed by what this PR offers. The use of "apply" and the comparison with Prolog is misleading here. Both work with backward inference, where you usually can only apply one or another theorem (or rule in Prolog) and this application will change the goal such that the other theorem / rule cannot be applied any more.

Forward reasoning does not change the goal, so it is usually the case that when two hints apply, both of them can be applied in parallel. The sole goal of this PR is to avoid that these parallel "applications" are tried in every possible order during backtracking of backward application.

**Comment by Zimmi48** [comment]
> Then both are applied in an arbitrary order

This is assuming a case where applying one doesn't prevent to apply the other (like in forward reasoning).

**Comment by Armael** [comment]
I just rebased on top of master, now that the dependencies have been merged.

Wrt the terminology, I agree that "NonBacktracking" is not so good, and "Committed" sounds better indeed.

**Comment by andres-erbsen** [comment]
How about `Forward` instead of `Committed`/`NonBacktracking`?

**Comment by MSoegtropIMC** [comment]
@andres-erbsen : we plan to have (some time) a Hint Forward, which matches on hypothesis rather than the goal. I guess it would get confusing to have a "Forward Hint" and a "Hint Forward" (and even a "Forward Hint Forward").

**Comment by MSoegtropIMC** [comment]
Btw.: as test case I decided to implement the Rush Hour puzzle (https://www.thinkfun.com/play-online/rush-hour/) in Coq. It offers quite a range of complexity and, depending on the search strategy, interesting interaction between forward and backward search, although the first step in backward search is not so nice (since one doesn't know the final setup, just the position of the goal car).

ETA: next week.

**Review by mattam82 — APPROVED** [review]
Hi guys, sorry for the large delay, I'm coming back to this now! Modulo the renaming to Commited in the syntax and code for consistency, I like the change to the code. If you can rebase and add a test-suite file demonstrating the feature I'll chaperon it.

**Comment by MSoegtropIMC** [comment]
@mattam82 : I still have a good example on my todo list. I discarded the rush hour puzzle example cause of licensing issues and also because it doesn't factorize in forward and backward moves elegantly. Let me resume think about it ... maybe I have something from automation I can share.

Regarding the name: how about "Commutative Hint" or "Permutable Hint". This expresses the idea that the hint can be exchanged in order with other commutative hints without affecting the result. The other names discussed either describe technical details on how this is implemented or specific use cases like forward search, but goal changing hints might also be commutative (e.g. certain normalizing sets of rewrites).

**Comment by Armael** [comment]
Thanks for coming back to this PR!
I just rebased it on top of master, and renamed the option. I went with `Commited` which I think gives a good intuition for the behavior of the flag.

There are still a couple TODOs to adress, I'll try to address them.

**Comment by mattam82** [comment]
It seems it will require a few overlays as well.

**Comment by Zimmi48** [comment]
There are too many failures for all of them to come from this PR. @Armael can you rebase once more?

**Comment by Armael** [comment]
I rebased: I'm not sure what's happening for flocq, hott and windows64. The other look like they possibly need an overlay indeed.

**Comment by Zimmi48** [comment]
This error in HoTT is worrying:
```
File "./theories/Basics/Equivalences.v", line 157, characters 0-206:
Error:
Unable to satisfy the following constraints:
In environment:
A : Type
B : Type
C : Type
f : A -> B
g : B -> C
H : IsEquiv f
H0 : IsEquiv (g o f)

?IsEquiv : "IsEquiv f"

?H : "IsEquiv (g o f o _^-1)"
```

**Comment by Zimmi48** [comment]
Unless this is finished quickly, this will have to be delayed to 8.11. I personally won't have time to work on this for the 8.10 feature freeze next week.

**Comment by MSoegtropIMC** [comment]
My plan was to finalize and test its usability during the implementer workshop this year. I think even if it can be finished before, it doesn't make that much sense to add this without also providing evidence for its usefulness and some tutorial.

**Comment by maximedenes** [comment]
I'll rebase this.

**Comment by maximedenes** [comment]
FTR, the rebase was made much more difficult than it should have been due to a lot of unrelated whitespace changes.

**Comment by Zimmi48** [comment]
I've added this PR to my TODO. Hopefully, we can have it for 8.13.

File: test-suite/success/CommitHint.v [hunk 1]
```diff
@@ -0,0 +1,24 @@
...
+
+Module Back.
+  Instance c0 : C 0 0 | 1 := {}.
+  Type _ : E _ _.
+End Back.
+
+Module NoBack.
+  Type _ : E _ _.
+  Instance c0 : C 0 0 ! 1 := {}.
+  Fail Type _ : E _ _.
```
SkySkimmer, line 22: There may be some inefficiency or bug in the implementation, here with `Typeclasses eauto := debug.` I get
~~~
Debug: 1: looking for (E ?a ?b) with backtracking
Debug: 1.1: simple apply exy on (E ?a ?b), 2 subgoal(s)
Debug: 1.1-1 : (C ?a ?b)
Debug: 1.1-1: looking for (C ?a ?b) with backtracking
Debug: 1.1-1.1: exact c0 on (C ?a ?b), 0 subgoal(s)
Debug: 1.1-2 : (D 0)
Debug: 1.1-2: looking for (D 0) without backtracking
Debug: 1.1-2: no match for (D 0), 0 possibilities
Debug: 1: looking for (E ?a ?b) with backtracking
Debug: 1.1: simple apply exy on (E ?a ?b), 2 subgoal(s)
Debug: 1.1-1 : (C ?a ?b)
Debug: 1.1-1: looking for (C ?a ?b) with backtracking
Debug: 1.1-1.1: exact c0 on (C ?a ?b), 0 subgoal(s)
Debug: 1.1-2 : (D 0)
Debug: 1.1-2: looking for (D 0) without backtracking
Debug: 1.1-2: no match for (D 0), 0 possibilities
~~~
note the doubled `exact c0` attempt.

mattam82: These are two runs of typeclasses eauto, I guess because we run one which is allowed to fail and then one which is not when performing `Type _ : E _ _`.

PR #7663, opened 2018-06-01
test suite: make target to regenerate failing output tests

SkySkimmer:


Zimmi48:
Please document in `test-suite/README.md`.

SkySkimmer:
done

Gaëtan Gilbert

gares (changes requested):
I think the code is fine but the name of the option is perfectible, in particular it says what the implementation does while it does not tell you why you should use it.

Imho `APPROVE_CURRENT_OUTPUT` is more clear (also `ACCEPT` could work).

The best documentation would be, IMO, to mention this variable in the log file "to accept the new output re-run make with APPROVE_CURRENT_OUTPUT=1".

ejgallego:
We could as well name it "promote", as it currently called in Dune which support this natively.

[I ignore if Gaëtan took inspiration from there]

ejgallego:
On the other hand I like more the split of the action in "diff" and "promote". I don't see why running the test suite should be necessary to update the files. IMHO, the new output should be saved and then promoted with a different rule.

SkySkimmer:
>On the other hand I like more the split of the action in "diff" and "promote". I don't see why running the test suite should be necessary to update the files. IMHO, the new output should be saved and then promoted with a different rule.

OK, we now have target `approve-output` to do this.
Also updated the test suite readme.

gares (approved):

gares:
ci failures seem spurious

PR #7670, opened 2018-06-01
Add is_literal tactic

maximedenes:
This PR is trying to address a request by @MSoegtropIMC.

One question I have for @MSoegtropIMC is: why isn't it enough to know if a term is closed (does not contain free variables)?

### DIFF
--- a/plugins/ltac/extratactics.ml4
+++ b/plugins/ltac/extratactics.ml4
@@ -950,6 +950,22 @@ TACTIC EXTEND is_const
     | _ -> Tacticals.New.tclFAIL 0 (Pp.str "not a constant") ]
 END;;
 
+let rec is_literal sigma t =
+  match EConstr.kind sigma t with
+  | Ind _ -> true
+  | Construct _ -> true
+  | App(f, args) -> is_literal sigma f && Array.for_all (is_literal sigma) args
+  | _ -> false
+
+TACTIC EXTEND is_literal
+| [ "is_literal" constr(x) ] -> [
+  Proofview.tclEVARMAP >>= fun sigma ->
+  if is_literal sigma x then
+    Proofview.tclUNIT ()
+  else
+    Tacticals.New.tclFAIL 0 (Pp.str "not a literal") ]
+END;;
+
 (* Command to grab the evars left unresolved at the end of a proof. *)
 (* spiwack: I put it in extratactics because it is somewhat tied with
    the semantics of the LCF-style tactics, hence with the classic tactic


plugins/ltac/extratactics.ml4
```diff
@@ -950,6 +950,22 @@ TACTIC EXTEND is_const
     | _ -> Tacticals.New.tclFAIL 0 (Pp.str "not a constant") ]
 END;;
 
+let rec is_literal sigma t =
+  match EConstr.kind sigma t with
+  | Ind _ -> true
+  | Construct _ -> true
+  | App(f, _) -> is_literal sigma f
```
SkySkimmer: Why is `S x` "literal"?

maximedenes: It's not, of course! :)

MSoegtropIMC:
@maximedenes : a closed term can contain function calls. I wouldn't name such a term a literal. A literal is in my view a number, a string, a list or possibly other structure of literals, but something like sin(3) I wouldn't call a literal. Of cause it would also make sense to have a function is_closedterm.

If it is of advantage to use is_closedterm instead of is_literal in my application I would have check. One can argue that forward rewriting with f(3) = x is also not a good idea, so I agree that it might be better to concentrate on closed terms.

In general I think that for automation it helps a lot to have a variety of term classification functions.

plugins/ltac/extratactics.ml4
```diff
@@ -950,6 +950,22 @@ TACTIC EXTEND is_const
     | _ -> Tacticals.New.tclFAIL 0 (Pp.str "not a constant") ]
 END;;
 
+let rec is_literal sigma t =
+  match EConstr.kind sigma t with
+  | Ind _ -> true
+  | Construct _ -> true
+  | App(f, args) -> is_literal sigma f && Array.for_all (is_literal sigma) args
+  | _ -> false
```
gares: It is a bit weird that `0 :: nil` is a literal and that `nat` is a literal but `nat :: nil` is not (because `Type` is not and `::` has an implicit)

Zimmi48: So sorts should be considered as litterals, right?

JasonGross:
I think it might be better to provide `has_lambda` and `is_closedterm` (though it's still not entirely clear to me what we should say about `is_closedterm mycons` given `Definition mycons := cons.`), and combine them at the user-level rather than the OCaml level.

plugins/ltac/extratactics.ml4
MSoegtropIMC: I don't have a strong opinion on if nat should be a literal or not. My main use of is_literal is to avoid rewriting a literal with something else. E.g. if I have an equation which after some destructs ends up to be 1=x to avoid that 1 is replaced with x in some automated rewrite with hypothesis. Obviously it usually wouldn't make a lot of sense to rewrite nat = T either.

Still it also would make sense to keep a connection to what people usually call a literal in other languages. Many languages have a notion of list or other structures literals, but I I don't think that if you ask some people (without this context) if they think a type name is a literal, they would probably say no.

So in summary: I am open to treat nat is literals in the tactic we discuss here, but I woudln't call it is_literal then, but maybe is_groundterm.

Maybe it makes more sense to have is_groundterm and if someone want's is_literal in the strict sense, he can still use is_groundterm and check if the head symbol is a constructor.

In this case I would probably use is_gorundterm directly in my tests.

MSoegtropIMC:
@JasonGross : is_closedterm /\ ! has_lambda is still something else than is_groundterm or is_literal.

Something which is "is_closedterm /\ ! has_lambda" could still have function application, while is_groundterm and is_literal should only contain constructors and names of inductive types.

But as I said before, I think it is very helpful for automation to have a rich (of cause reasonably orthogonal) set of term classification functions.

Something which would also be very useful would be functions which *count* the number of various base constructs in a term (constructors, function applications, free variables, inductive type names, evars, ...)

MSoegtropIMC:
@JasonGross : The nomenclature in discussions at the CIW converged to:

**ground term**: a **fully reduced** term without free variables, which (I think) is a term which contains only  constructors, applications of constructors and names of inductive types.

**literal term**: a ground term where the head symbol is a constructor. This is what in most programming languages would be called a literal.

**closed term**: a possibly **not fully reduced** term which does not contain free variables (not sure about evars). Since with sections (as far as I understand) reduction can introduce free section variables, one would have to define if for unreduced terms "no free variables" applies to the term as is or the reduced term. The same applies to criteria like "has_lambda". Since classification functions for ltac should be efficient, I think they should work on unreduced terms, which makes properties like "closed_term" or "has_lambda" conceptually difficult.

JasonGross:
Ah, that makes sense.

> a **fully reduced** term without free variables which (I think) is a term which contains only constructors, applications of constructors and names of inductive types.

I think it should also include sorts and names of coinductive types.  And `fun x : nat => x` is fully reduced and without free variables.  Is it ground?  What about `fix id (n : nat) : nat := match n with O => O | S n' => id n' end`?  It would seem a tiny bit strange, to me, to call `@pair nat nat O O` ground/literal, but to not call `@existT nat (fun _ => nat) O O` ground/literal...

MSoegtropIMC:
@JasonGross : getting closer :-)

I would guess from the discussion at the CIW that the proper definition of "ground term" is "fully reduced term without free variables", including all the other things you listed.

Let's then define a "literal term" to be a term containing only constructors and names of inductive types. One could name a literal term where the head symbol is a constructor a "value literal term" and otherwise a "type literal term" and the union of both "literal term".

So `@existT nat (fun _ => nat) O O` would be ground but not literal, since it contains a lambda. A literal should be really something very simple. As I mentioned, I use in my automation the idea that literals are the simplest thing in the world and that a term would always become more complicated by replacing a literal with a variable. For literal as defined above this works well in my automation. But I think one can't say this of terms which include lambdas.

I hope these definitions are acceptable and clear for all involved in the discussion here.

JasonGross:
Ok, I think I'm sold on this categorization.  I think I agree with @gares, though, that sorts (and names of coinductives) should also be literals, so that `0 :: nil` and `nat :: nil` are both literals.

ppedrot (commented):
This looks like a typical no-implementation-fits-all problem to me, as witnessed by the comments in this thread. I can only lament that we have to resort to writing such a function in ML due to the lack of expressivity of Ltac, where any sane language would allow to write this out using properly expressive primitive functions.

ppedrot:
@JasonGross Indeed. I'm not against the inclusion of this into Ltac1, but I cannot help feeling this is a Danaides' barrel.

Zimmi48:
@ppedrot In order to allow Michael to test Ltac2, would you mind releasing a beta version now? If done quickly, it could allow including it in the Windows package for 8.8.1.

MSoegtropIMC:
@ppedrot : well with the is_ctor function Jason integrated a while ago (it was just not documented and in the one year Coq break I had in the last 12 months I forgot about it) one can implement is_groundterm reasonably well in Ltac. One can even implement is_ctor in Ltac, but it requires a quod libet sandbox to provide the parameters and then to see if it can be destructed.

The problem is that term classification is a very common task in gating in automation. Gating means it is executed just to check if some automation applies and thus also executed if a certain hint extern does not apply. I asked for this function because it doesn't make that much sense to spend cycles on such trivial things with Ltac functions.

In general I think that automation does profit from a reasonably complete and orthogonal set of term classification functions, not only elementary things like is_ctor but also complete term classification like is_groundterm.

Of cause I agree that it is most important to have a set of elementary classification functions like is_ctor.

JasonGross:
@MSoegtropIMC The issue, as I see it, is that in Ltac1, walking an entire term is relatively slow, so it's inefficient to implement things like `is_ctor` and `is_literal` by hand.  By contrast, in Ltac2, you can basically just transcribe the OCaml implementation and have it run almost as fast, because Ltac2 gives you things like `kind_of_term` and matching on that datastructure.  So, in Ltac1, it bascially needs to be implemented in OCaml for speed and to avoid kludges, but in Ltac2, it's quite easily implemented as part of a sort of standard library of tactics, and doesn't need to be in the core.

MSoegtropIMC:
@JasonGross : indeed a good point. In case a speed difference between Ltac2 and OCaml is not so significant, the need for the higher level functions is not really there.

But I guess we can agree on that is_ctor should be there? Or has ltac2 ways to look into the data structures and decide this without a quod libet sandbox?

How about having is_trivial now as intermediate solution until Ltac 2 is there and put a comment in the manual that it will most likely be deprecated when Ltac 2 arrives and should only be used by people who are willing to change?

JasonGross:
@MSoegtropIMC Ltac2 has [`Constr.kind : constr -> kind`](https://github.com/ppedrot/ltac2/blob/master/theories/Constr.v#L23) which gives back an element of the datastructure
```coq
Ltac2 Type kind := [
| Rel (int)
| Var (ident)
| Meta (meta)
| Evar (evar, constr array)
| Sort (sort)
| Cast (constr, cast, constr)
| Prod (ident option, constr, constr)
| Lambda (ident option, constr, constr)
| LetIn (ident option, constr, constr, constr)
| App (constr, constr array)
| Constant (constant, instance)
| Ind (inductive, instance)
| Constructor (constructor, instance)
| Case (case, constr, constr, constr array)
| Fix (int array, int, ident option array, constr array, constr array)
| CoFix (int, ident option array, constr array, constr array)
| Proj (projection, constr)
].
```
So you can peel off the `App` nodes and then just check if the kind of your term is `Constr.kind.Constructor`.


To be clear, I'm not opposed to this PR, but just noting that things are nicer in Ltac2.

ppedrot (dismissed):
Can you split the entries and revert the change in the test? When this is done if everybody agrees I'll merge the PR.

ppedrot:
What do we do with this PR? It seems people do not really agree on what a literal is, and I think we should rather provide some kind of more atomic primitive to allow them to implement it precisely...

MSoegtropIMC:
@ppedrot : can we delay this decision two months? I was out of serious Coq work for a while and didn't really have time to test the various options. The elementary tests exist meanwhile (just not yet documented) and it might be that this is good enough from a performance point of view. Other options are LTAC-2 and MTAC which both now come with the Windows installer (with 8.8.2). I will return to serious Coq work soon and test all this in practice. Then I can give better comments. Since this is quite controversial, some performance measurements would help in the argument (one way or the other).

ppedrot:
@MSoegtropIMC ping.

ppedrot:
Now that Ltac2 is part of Coq, I think we can close this PR.

MSoegtropIMC:
I plan to spend the developer workshop with automation tests (also performance) and examples. After that I wanted to clean up all related PRs. I hope it is ok to wait one month.

ppedrot:
@MSoegtropIMC can we close it now?

ppedrot:
I'm taking the liberty to close it, reopen if not happy.

MSoegtropIMC:
@ppedrot: sorry for being so unresponsive on this - since I moved to mostly doing numerical algorithm verification, the structure of my automation problems changed a lot so right now I am not able to judge on real world scenarios if this is needed or not. Some first tests of Ltac-2 make me think that the facilities for term classification in Ltac-2 are sufficiently powerful and performant for my past and future needs, but I cannot really say something substantial on the performance.

So yes, at least for the time being i agree to closing this.

PR #7677, opened 2018-06-02
[api] Remove Misctypes

ejgallego:
The `Misctypes` file contains a popurri of type definitions, however, this diverse nature makes it impossible to map it to a concrete component, and kind of forces an obscure nature on it.

We relocate the definitions in Misctypes to other modules; for the most part, I think the relocation makes sense, removing this catch-all file from the tree.


gares (approved):
Misc = Mess

gares:
I think your approach with a compat module full of deprecations is the way to go

ejgallego:
I think that most of the types in `Misctypes` pertain to @ppedrot 's domain [tho a few additional reviews would be great], so I am proposing him to be the assignee of this PR.

maximedenes (approved):
I reviewed only the kernel changes, they look ok to me.

kernel/names.mli
```diff
@@ -605,3 +605,9 @@ type evaluable_global_reference =
   | EvalConstRef of Constant.t
 
 val eq_egr : evaluable_global_reference ->  evaluable_global_reference -> bool
+
+(** Located identifiers and objects with syntax. *)
+
+type lident = Id.t CAst.t
```
ppedrot: I don't think that these types should be put in the kernel. They refer to AST types that live well upwards.

engine/evarutil.mli
```diff
@@ -22,12 +22,22 @@ open EConstr
 val new_meta : unit -> metavariable
 val mk_new_meta : unit -> constr
 
+(** General evar naming using intro patterns  *)
+type intro_pattern_naming_expr =
+  | IntroIdentifier of Id.t
+  | IntroFresh of Id.t
+  | IntroAnonymous
+
```
ppedrot: This should probably be put into Namegen, but I don't have a strong opinion on that.

ppedrot (changes requested):
Everything is mostly OK, except for the fact that the located versions of AST datatypes were put into the kernel. I really don't think it is a good idea, and this seems reflected by the fact they're not used that down.

kernel/names.mli
ejgallego: Where do you want to put them? Names is not such a bad place after all I think for them.

engine/evarutil.mli
ejgallego: Sounds reasonable, let me see how things would look

kernel/names.mli
ejgallego: We are talking like 4 lines; in fact I'd even merge `Names` and `Nameops` and get rid of one module. We have too many modules.

ejgallego: Imagine having all identifier-related stuff in a single module, can't be true!

ppedrot: Names in already too big for its own good, IMHO... I'd rather split it.

ejgallego: I hope we are not leading to a world where 20% of the TCB are `open Foo` statements XD

ejgallego: I'd rather have these 4 silly lines not block this PR for very long. I still think that `Names` is not such a bad place for the located objects as IMHO it seems natural that the module defining `Id.t` also defines it's located version.

Noted that this is already used in the next layer `library/` so this is far from "well upwards" but instead it is "immediatedly upwards".

ppedrot: OK, OK, fine with me.

ppedrot (approved):

kernel/names.mli
ejgallego: I understand your concern, but couldn't do better :( but yeah, if we find another place I'll be very happy to move it. Note indeed that Names and Subst have an special place within the kernel as they are used by both the checker and the kernel itself. So in fact they should be placed in a different place, even in clib. We have loads of stuff there that the kernel uses so indeed that's a tricky question.

ppedrot:
I'll leave a few days for other reviewers to comment on this PR. If they stay silent though, this will be considered as tacit acceptance.

engine/evarutil.mli
ejgallego: I looks better indeed, pushing the change to `Namegen`.

ejgallego:
So I another small pass; IMHO it is not perfect but I think it is a good step. I do agree with the `lid` etc... comment, however we are talking about 3 lines and indeed the kernel could have some use for these structures so let's keep it there for now.

PR #7679, opened 2018-06-02
Clean native compilation of primitive projections

ppedrot:
More straightforward native compilation of primitive projections.

Instead of having a constant-based compilation of projections, we generate them at the compilation time of the inductive block to which they pertain.

This is a first step towards being able to replace the kername-based representation of projections with a inductive + inded based one.

### DIFF
--- a/kernel/declarations.ml
+++ b/kernel/declarations.ml
@@ -52,7 +52,7 @@ type inline = int option
 type projection_body = {
   proj_ind : MutInd.t;
   proj_npars : int;
-  proj_arg : int;
+  proj_arg : int; (** Projection index, starting from 0 *)
   proj_type : types; (* Type under params *)
   proj_eta : constr * types; (* Eta-expanded term and type *)
   proj_body : constr; (* For compatibility with VMs only, the match version *)
--- a/kernel/nativecode.ml
+++ b/kernel/nativecode.ml
@@ -53,7 +53,7 @@ type gname =
   | Gind of string * inductive (* prefix, inductive name *)
   | Gconstruct of string * constructor (* prefix, constructor name *)
   | Gconstant of string * Constant.t (* prefix, constant name *)
-  | Gproj of string * Constant.t (* prefix, constant name *)
+  | Gproj of string * inductive * int (* prefix, inductive, index (start from 0) *)
   | Gcase of Label.t option * int
   | Gpred of Label.t option * int
   | Gfixtype of Label.t option * int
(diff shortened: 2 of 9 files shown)


kernel/declarations.ml
```diff
@@ -52,7 +52,7 @@ type inline = int option
 type projection_body = {
   proj_ind : MutInd.t;
   proj_npars : int;
-  proj_arg : int;
+  proj_arg : int; (** Projection index, starting from 0 *)
```
SkySkimmer: I believe this includes parameters, ie `proj_arg` for `fst` is `2`. It may be worth checking and including this information in the comment.

ppedrot: Nope, I checked, this is really the projection index.

kernel/nativecode.ml
```diff
@@ -1979,12 +1958,41 @@ let compile_mind prefix ~interactive mb mind stack =
...
+      let i = push_symbol (SymbProj (ind, j)) in
+      let accu = MLapp (MLprimitive Cast_accu, [|MLlocal cf_uid|]) in
+      let accu_br = MLapp (MLprimitive Mk_proj, [|get_proj_code i;accu|]) in
+      let code = MLmatch(asw,MLlocal cf_uid,accu_br,[|[((ind,1),cargs)],MLlocal ci_uid|]) in
+      let code = MLlet(cf_uid, MLapp (MLprimitive Force_cofix, [|MLlocal c_uid|]), code) in
+      let gn = Gproj ("", (pb.proj_ind, j), pb.proj_arg) in
+      Glet (gn, mkMLlam [|c_uid|] code) :: acc
+    in
+    let projs = match mb.mind_record with
+    | None | Some None -> []
```
maximedenes: What on earth is that?

maximedenes (approved):
The change looks ok, I tested it a bit.

kernel/nativecode.ml
ppedrot: (Discussed on gitter but for the record:) Crappy native primproj API fixed in a later PR.

ppedrot:
Just in case that takes ages, this PR is purely about native compilation, so it's not completely necessary for @mattam82 to give a review (that's always better though).

mattam82 (approved):
LGTM, as far as I understand the native comp code

maximedenes:
@ppedrot the same vio failure shows up here. Can you report the spurious failure while I merge?

PR #7680, opened 2018-06-02
[ci] Expose updated `OCAMLPATH` for CI users.

ejgallego:
This is needed for CI packages that use `META.coq` such as in
https://github.com/coq/coq/pull/7656 .

### DIFF
--- a/dev/ci/ci-common.sh
+++ b/dev/ci/ci-common.sh
@@ -8,6 +8,7 @@ export NJOBS
 
 if [ -n "${GITLAB_CI}" ];
 then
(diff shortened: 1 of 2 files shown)


SkySkimmer:
Is this something that should be concatenated to the old value like PATH?

ejgallego:
> Is this something that should be concatenated to the old value like PATH?

In our setup OCAMLPATH will be always empty, but yeah, we could.

SkySkimmer (approved):
Is this why the quickchick PR is failing?

PR #7682, opened 2018-06-02
Fixes #7641: more detailed message about disjunctive patterns with different variables

### LINKED ISSUE #7641: "The components of this disjunctive pattern must bind the same variables." should name which variables overlap
#### Version

8.8.0

Consider, e.g.,
```coq
Inductive foo := bar | baz.

Definition beq (x y : foo) : bool
  := match x, y with
     | bar, barr
     | baz, baz
       => true
     | _, _ => false
     end.
(* Error: The components of this disjunctive pattern must bind the same variables. *)
```
I want to be told what my spelling mistake here is, and where.  That is, I'd like to see something like
```
Error:
The components of this disjunctive pattern must bind the same variables.
The component
"bar, barr"
binds variable "barr" while the component
"baz, baz"
binds no variables.
```

cc @herbelin

herbelin:
**Kind:** error message improvement

This is slight quick improvement on the current situation, addressing wish #7641.

Could still be made more detailed with more time.

Note that this is a new example where one may want to hightlight several locations at the same time.

### DIFF
--- a/interp/constrintern.ml
+++ b/interp/constrintern.ml
@@ -1141,9 +1141,18 @@ let check_number_of_pattern loc n l =
   if not (Int.equal n p) then raise (InternalizationError (loc,BadPatternsNumber (n,p)))
 
 let check_or_pat_variables loc ids idsl =
-  if List.exists (fun ids' -> not (List.eq_set (fun {loc;v=id} {v=id'} -> Id.equal id id') ids ids')) idsl then
-    user_err ?loc  (str
-    "The components of this disjunctive pattern must bind the same variables.")
+  let eq_id {v=id} {v=id'} = Id.equal id id' in
+  (* Collect remaining patterns which do not have the same variables as the first pattern *)
+  let idsl = List.filter (fun ids' -> not (List.eq_set eq_id ids ids')) idsl in
+  match idsl with
+  | ids'::_ ->
+    (* Look for an [id] which is either in [ids] and not in [ids'] or in [ids'] and not in [ids] *)
+    let ids'' = List.subtract eq_id ids ids' in
+    let ids'' = if ids'' = [] then List.subtract eq_id ids' ids else ids'' in
+    user_err ?loc
+      (strbrk "The components of this disjunctive pattern must bind the same variables (" ++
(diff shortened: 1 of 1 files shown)


interp/constrintern.ml
```diff
@@ -1141,9 +1141,16 @@ let check_number_of_pattern loc n l =
...
+  let eq_id {v=id} {v=id'} = Id.equal id id' in
+  let idsl = List.filter (fun ids' -> not (List.eq_set eq_id ids ids')) idsl in
+  match idsl with
+  | ids'::_ ->
+    let ids'' = List.subtract eq_id ids ids' in
+    let ids'' = if ids'' = [] then List.subtract eq_id ids' ids else ids'' in
+    user_err ?loc
+      (strbrk "The components of this disjunctive pattern must bind the same variables (" ++
+       Id.print (List.hd ids'').v ++ strbrk " is not bound in all patterns).")
+  | [] -> ()
```
ejgallego: It feels like the list manipulation functions here are high-level operations that should be properly abstracted and named, right?

ejgallego: Indeed I mean to name both:
```
let idsl = List.filter (fun ids' -> not (List.eq_set eq_id ids ids')) idsl in
```
and
```
let ids'' = List.subtract eq_id ids ids' in
let ids'' = if ids'' = [] then List.subtract eq_id ids' ids else ids'' in
```
to my taste we have too many occurences of this style in the code; I'd be great to start commonalization/deduplication. Also, that would allow people to improve the implementation in the future.

I think it would be good to tweak it, but if you don't feel like doing let me know and I'll merge as-is.

herbelin: >  if you don't feel like doing let me know and I'll merge as-is.

Indeed, making such a cleaning properly would require time which is not prioritary in my short term.

herbelin: OK, done. Tell me if it is clear enough.

ejgallego: Thanks a lot Hugo, merging as soon as CI passes.

ejgallego (approved):

ejgallego (approved):

PR #7177, opened 2018-04-05
Unifying names of "smart" combinators + adding combinators in CArray

herbelin:
**Kind:** architecture

This is part of "improving" the API where "improving" means here easier to learn and remember, more regular, even if it implies extra "burden" for who already uses the API.

Concretely:
- We model the name of `smartmap` and `smartfoldmap` on the name of the underlying combinators (and deprecate other names).
- We add two new smart combinators that will be useful for coq/ceps#34.

More generally, here are questions to efficiency experts:
- Couldn't we have a more precise name than `smart` to describe these combinators? For instance, I'm thinking at `homogeneous` or, shortlier, `homo`.
- What about having a specific name space, say `List.Homo`, `Array.Homo` for such combinators, as we have e.g. `Array.Fun1`, reusing then the "standard" naming scheme within the given name spaces?
- What about using more pervasively these combinators in the implementation whenever the `map` is homogeneous?

I can eventually do all these changes.

PS: This is part of ongoing implementation on coq/ceps#34, isolating these commits for easier "domain-specific" reviewing and discussion.

PR #7683, opened 2018-06-02
[lib] Fix wrong deprecation annotations.

ejgallego:
Introduced in #7177


herbelin:
Well, indeed. Shall merge in the afternoon if owner @ppedrot don't do it in the meantime.

ppedrot (approved):
Trivially OK.

PR #7687, opened 2018-06-03
[ci] [docker] Pin specific versions of OPAM CI dependencies.

ejgallego:
Things like Menhir or Elpi seem to fragile as to allow a
non-deterministic install in the Dockefile, we thus specify the
version for all `CI_OPAM` packages.

cc: https://github.com/AbsInt/CompCert/issues/234

### DIFF
--- a/dev/ci/docker/bionic_coq/Dockerfile
+++ b/dev/ci/docker/bionic_coq/Dockerfile
@@ -1,4 +1,4 @@
-# CACHEKEY: "bionic_coq-V2018-05-07-V2"
+# CACHEKEY: "bionic_coq-V2018-06-04-V2"
 # ^^ Update when modifying this file.
 
 FROM ubuntu:bionic
@@ -8,7 +8,7 @@ ENV DEBIAN_FRONTEND="noninteractive"
 
 RUN apt-get update -qq && apt-get install -y -qq m4 wget time gcc-multilib opam \
         libgtk2.0-dev libgtksourceview2.0-dev \
-        texlive-latex-extra texlive-fonts-recommended hevea \
+        texlive-latex-extra texlive-fonts-recommended texlive-science \
         python3-sphinx python3-pexpect python3-sphinx-rtd-theme python3-bs4 python3-sphinxcontrib.bibtex python3-pip
 
 RUN pip3 install antlr4-python3-runtime
@@ -19,15 +19,19 @@ ENV NJOBS="2" \
     OPAMROOTISOK="true"
 
 # Base opam is the set of base packages required by Coq
-ENV COMPILER="4.02.3" \
-    BASE_OPAM="num ocamlfind jbuilder ounit"
+ENV COMPILER="4.02.3"
 
 RUN opam init -a -y -j $NJOBS --compiler="$COMPILER" default https://opam.ocaml.org && eval $(opam config env) && opam update
 
-# Setup of the base switch; CI_OPAM contains Coq's CI dependencies.
+# Common OPAM packages.
+# `num` does not have a version number as the right version to install varies
+# with the compiler version.
+ENV BASE_OPAM="num ocamlfind.1.8.0 jbuilder.1.0+beta20 ounit.2.0.8" \
+    CI_OPAM="menhir.20180530 elpi.1.0.3 ocamlgraph.1.8.8"
+
+# BASE switch; CI_OPAM contains Coq's CI dependencies.
 ENV CAMLP5_VER="6.14" \
-    COQIDE_OPAM="lablgtk.2.18.5 conf-gtksourceview.2" \
-    CI_OPAM="menhir elpi ocamlgraph"
+    COQIDE_OPAM="lablgtk.2.18.5 conf-gtksourceview.2"
 
 RUN opam switch -y -j $NJOBS "$COMPILER" && eval $(opam config env) && \
     opam install -j $NJOBS $BASE_OPAM camlp5.$CAMLP5_VER $COQIDE_OPAM $CI_OPAM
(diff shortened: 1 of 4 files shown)


dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -19,15 +19,16 @@ ENV NJOBS="2" \
     OPAMROOTISOK="true"
 
 # Base opam is the set of base packages required by Coq
+# `num` does not have a dependency as it varies on the compiler version.
 ENV COMPILER="4.02.3" \
-    BASE_OPAM="num ocamlfind jbuilder ounit"
+    BASE_OPAM="num ocamlfind.1.8.0 jbuilder.1.0+beta20 ounit.2.0.8"
```
SkySkimmer: While you're at it move BASE_OPAM to the next ENV directive, immediately before it's used.

SkySkimmer:
How did this break CI, did we regenerate the docker image since the menhir update?

ejgallego:
> How did this break CI, did we regenerate the docker image since the menhir update?

Yes. I cleaned the wrong image so I had to regenerate and we hit the compert bug that by the way this PR workarounds.

dev/ci/docker/bionic_coq/Dockerfile
ejgallego: Moved and tweaked a bit.

SkySkimmer (approved):

ejgallego:
By the way Gaëtan (@SkySkimmer ) should be tweak the `docker-boot` job so that it uses cached layers?

It seems that we we will often update the upper layer, so that could help a bit in the bootstrap time.

SkySkimmer:
>By the way Gaëtan (@SkySkimmer ) should be tweak the docker-boot job so that it uses cached layers?

We would have to figure out which image to pull as a cache base.
Also it would mean that updating the cache key doesn't run a full rebuild so we would want some way to disable the cache.

ejgallego:
> Also it would mean that updating the cache key doesn't run a full rebuild so we would want some way to disable the cache.

I mean that if the `apt` hashed layer was already stored in the registry, the build could just pull it, right?

The hashes ensure that we cannot mess things up. But I didn't have time to fully look at it, there may be more tricky parts; it was just a thinking out loud moment.

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -19,15 +19,18 @@ ENV NJOBS="2" \
...
 # Base opam is the set of base packages required by Coq
-ENV COMPILER="4.02.3" \
-    BASE_OPAM="num ocamlfind jbuilder ounit"
+ENV COMPILER="4.02.3"
 
 RUN opam init -a -y -j $NJOBS --compiler="$COMPILER" default https://opam.ocaml.org && eval $(opam config env) && opam update
 
-# Setup of the base switch; CI_OPAM contains Coq's CI dependencies.
+# Common OPAM packages
+# `num` does not have a dependency as it varies on the compiler version.
```
Zimmi48: Do you mean a version number?

Zimmi48 (commented):
Being more deterministic sounds like a good idea.

dev/ci/docker/bionic_coq/Dockerfile
ejgallego: Indeed.

ejgallego:
Let's wait to merge this one to see what upstream does; in https://github.com/AbsInt/CompCert/pull/236 they are proposing to require an even newer menhir version so until that is resolved I'm afraid we'll have to live with the current status, or merge this PR and an additional.

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -19,15 +19,19 @@ ENV NJOBS="2" \
...
+ENV COMPILER="4.02.3"
 
 RUN opam init -a -y -j $NJOBS --compiler="$COMPILER" default https://opam.ocaml.org && eval $(opam config env) && opam update
 
-# Setup of the base switch; CI_OPAM contains Coq's CI dependencies.
+# Common OPAM packages.
+# `num` does not have a version number as the right version to install varies
+# with the compiler version.
+ENV BASE_OPAM="num ocamlfind.1.8.0 jbuilder.1.0+beta20 ounit.2.0.8" \
+    CI_OPAM="menhir.20180528 elpi.1.0.3 ocamlgraph.1.8.8"
```
gares: I don't think this is a good idea.

IMO we should ask the upstreams of our CI module X to provide an opam file (when they have external dependencies) and then run `opam install --deps-only X` instead of hard coding the dependencies here.

Otherwise I (with the upstream hat) I've an hard time to update `coq-elpi`. It is no more enough to move the `coq-master` branch Coq-CI is tracking, but I also have to make a PR for this file, and I don't even know it exists.

gares (changes requested):

dev/ci/docker/bionic_coq/Dockerfile
Zimmi48: The docker based system saves us a lot of build time but doesn't allow the solution that you propose. Anyways, you shouldn't have to worry whenever you update your dependencies. I plan to implement commit pinning soon as when we have that, this problem will go away.

Zimmi48: Not that your request for changes is not relevant as this PR doesn't change the model. In the previous situation, we also used a fixed set of dependencies but we didn't know which ones exactly.

gares: > The docker based system saves us a lot of build time but doesn't allow the solution that you propose

I don't get it. Exactly as you run `opam install elpi.XXX` in the script above, you could run `opam install coq-elpi --deps-only`. Then as soon as the docker image is generated  the dependencies of coq-elpi are going to get right.

> Not that your request for changes is not relevant as this PR doesn't change the model.

Maybe what I'm missing is *when* the image is generated. If one need to change this file in oder to have it re-generated then you can discard my review.

ejgallego: Regeneration is on every build unless you set a variable to cache the image, which we do.

We could do what you propose but I am not sure I like the non-determinism there; there is always the tradeoff with caching; what you are telling us is that we shouldn't cache elpi, indeed, we could have a more complex logic to be smarter in caching, but I think for now the tradeoff is acceptable.

ejgallego: Also note that docker tends not to like non-deterministic commands, IIUC.

ejgallego: So, in other words, today our Docker cache invalidation is manual, and for the moment it will remain so, thus depending packages must manually signal a cache invalidation attempt or override the status in their `ci-` script. The same was true before this PR, but it now makes explicit the values used in the cached image, which is a good thing "(TM)"

Zimmi48:
> Regeneration is on every build unless you set a variable to cache the image, which we do.

This is incorrect. Regeneration actually requires changing the docker file. And @gares said:

> Maybe what I'm missing is when the image is generated. If one need to change this file in oder to have it re-generated then you can discard my review.

> So, in other words, today our Docker cache invalidation is manual, and for the moment it will remain so, thus depending packages must manually signal a cache invalidation attempt or override the status in their ci- script. The same was true before this PR, but it now makes explicit the values used in the cached image, which is a "good thing"™

:+1:

SkySkimmer:
>This is incorrect. Regeneration actually requires changing the docker file

No, we pull based on the CACHEKEY so when it changes there is no cache.

Zimmi48:
Yes, is it different from what I said?

SkySkimmer:
The CACHEKEY in the dockerfile is only a comment, it doesn't cause or block rebuilding the image.

Zimmi48:
Hum you're right. I've forgotten we haven't implemented a system that takes the hash of the docker file yet. So in fact, regenerating an image only requires changing the CACHEKEY in the GitLab file (and turning the appropriate environment variable on).

ejgallego:
We run the regen at every build, but indeed we have yet another caching layer with cachekey.

ejgallego:
Ok, this should be ready to get in once the CI passes.

ejgallego:
My proposal to dismiss @gares review is "docker file only provides a baseline for the CI system, while this is not optimal wrt dependencies, caching considerations still make the tradeoff worth it. CI developments with more specific needs should setup their own enviroment in the CI script".

gares (approved):

PR #7684, opened 2018-06-02
[doc] Build ML API documentation artifact.

ejgallego:
This is not optimal for we have to rebuild the `.cmi` as
`ocamldoc` cannot yet use the `_install_ci/` directory.

Overall the `mli` documentation is in a sorry state, however, I think
this is a first step in order to improve it.

Note that the `ml-doc` target seems broken in OCaml 4.07.0, needs
investigation.

PR #7688, opened 2018-06-03
[doc] [README] [ci] Add links to documentation builds.

ejgallego:
cc: #7161
cc: https://github.com/coq/www/pull/69

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,3 +54,15 @@ used, and include a complete source example leading to the bug.
 ## Contributing
 
 Guidelines for contributing to Coq in various ways are listed in the [contributor's guide](CONTRIBUTING.md).
+
+## Continuous builds
+
+Coq's Continuous Integration system provides documentation and binary
+builds for all the development branches. In particular, you can access:
+
+- Coq's Documentation [master branch]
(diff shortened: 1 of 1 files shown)


ejgallego:
Rendered: https://github.com/ejgallego/coq/blob/25798f84bad109a8e3b69ec369c2971632abcaaa/README.md

Zimmi48 (commented):
Putting these links in the README may make more sense that what I proposed in coq/www#69.
However, I wouldn't create a new section for this, I would put the links in the already existing "documentation" section.
Additionally, the current state of the standard library documentation isn't acceptable and we should not link to it until we fix it.

ejgallego:
> Additionally, the current state of the standard library documentation isn't acceptable and we should not link to it until we fix it.

We have a bug for it?

Zimmi48:
Not yet. Feel free to open one, or I will eventually.

ejgallego:
> Not yet. Feel free to open one, or I will eventually.

Thanks for opening #7720

Zimmi48:
This PR was semantically depending on #7684. If you want to have it merged soon, you should remove the mli docs. The stdlib also doesn't make much sense to keep until #7720 is solved. Moreover the link to the refman needs to be updated after #7515. Finally, as I said in my review comment, I would prefer these links (this link actually, after the removal of the other two) to be in the preexisting documentation section of the README. If you don't have time for this, I can also take over.

PR #7690, opened 2018-06-03
Fixing typos in file Berardi.v

herbelin:
**Kind:** two mini typos

One of the typos requires a compatibility notation. Maybe not worth a word in CHANGES though as it is somehow more a cultural file than a library file, but could do also if someone thinks it should.


Zimmi48:
I don't believe the CHANGES entry is required.

gares (approved):

ppedrot (approved):

PR #7691, opened 2018-06-03
Fixing spelling of statment/statement in two API types

herbelin:
**Kind:** API mispelling

I realized that	we mispelled "statment" in the past. This PR proposes to fix this. I don't know however if the cost for incompatibility is worth correcting the spelling.

It is about two	grammar types, so maybe	not used so much, but still an API break.

Wondering how would developers feel about this PR, whether they are native English or not English speakers, and Coq or plugin developers.

- [ ] Entry added in `changes.md` (if considered worth)
- [ ] deprecation phase (if considered worth)

### DIFF
--- a/parsing/extend.ml
+++ b/parsing/extend.ml
@@ -116,7 +116,7 @@ and 'a rules =
 type 'a production_rule =
 | Rule : ('a, 'act, Loc.t -> 'a) rule * 'act -> 'a production_rule
 
-type 'a single_extend_statment =
+type 'a single_extend_statement =
   string option *
   (** Level *)
   gram_assoc option *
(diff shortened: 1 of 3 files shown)


SkySkimmer (commented):
Yeah fix it.
This isn't something where backward compat is critical like HTTP Referer.
In fact it looks like an internal type that plugins aren't expected to use.

ppedrot (approved):
I'm all for the patch, and favouring quick merge without neither changes nor deprecation. This is a really internal API.

ppedrot:
(If people do not agree with my position, please comment further. Otherwise I'll merge soon.)

Issue #5351, opened 2017-02-10
Intuition does not simplify P -> Q, P in hypotheses with universe polymorphism


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#5351
From: @tchajed
Reported version: trunk
CC:   coq-bugs-redist@lists.gforge.inria.fr

See also: [BZ#5350](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%235350%22)

SkySkimmer:
In this nonlinear ltac pattern https://github.com/coq/coq/blob/bb63a9f2209108424bd3dadc1c479f9ddc0e1803/theories/Init/Tauto.v#L28 different instances of a constant will not match.

herbelin:
Hum, what to conclude?

ppedrot:
We can use constr_eq and remove the non-linear match. Not sure about efficiency though.

PR #7696, opened 2018-06-04
Remove some univ_flexible_alg from cases

SkySkimmer:
There are a couple left which seem OK.

### DIFF
--- a/pretyping/cases.ml
+++ b/pretyping/cases.ml
@@ -995,7 +995,7 @@ let expand_arg tms (p,ccl) ((_,t),_,na) =
 
 let use_unit_judge env evd =
   let j, ctx = coq_unit_judge !!env in
-  let evd' = Evd.merge_context_set Evd.univ_flexible_alg evd ctx in
+  let evd' = Evd.merge_context_set Evd.univ_flexible evd ctx in
     evd', j
 
 let add_assert_false_case pb tomatch =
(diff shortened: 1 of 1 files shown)


SkySkimmer:
See #7490 for context.

mattam82:
@SkySkimmer is there a way to restart build:base on gitlab's CI?

SkySkimmer:
>@SkySkimmer is there a way to restart build:base on gitlab's CI?

I added you to the gitlab Coq group so you should have a button now.
I also clicked the button for you in this case.

SkySkimmer:
It seems like this breaks quickchick somehow.

SkySkimmer:
>It seems like this breaks quickchick somehow.

They have a match somewhere with `return Type@{max(Set,Ui,Uv)}` which now causes the `hlist_tl` constant to have an extra universe. Then intuition fails to unify different instances.

SkySkimmer:
Basically hitting #5351.

ppedrot:
Hopefully #8905 will be the correct fix allowing this PR to go through (if still of relevance).

mattam82 (approved):
Let's hope this just goes through!

SkySkimmer:
@mattam82 should be good to merge

ppedrot:
Let me take over.

mattam82:
Ok

mattam82:
A PR needs to be merged upstream that’s all.

ejgallego:
> A PR needs to be merged upstream that’s all.

Ah, I missed the overlay sorry.

PR #7701, opened 2018-06-05
Deprecate unused internal ComInductive API

jashug:
**Kind:** cleanup
- [x] Entry added in dev/doc/changes.md.

No one seems to be using these outside of `comInductive.ml` anymore.
They were marked as "Internal" and "Exported for Funind", and Funind doesn't use them anymore.
In the spirit of pruning the API, I suggest deprecation followed by removal.

I also have a branch removing this API here: https://github.com/jashug/coq/tree/remove-internal-comInductive-API
with Travis run at https://travis-ci.org/jashug/coq/builds/388047239

A note: deprecating the types `structured_one_inductive_expr` and `structured_inductive_expr` raises warnings when they are referenced in the following deprecated functions.

### DIFF
--- a/dev/doc/changes.md
+++ b/dev/doc/changes.md
@@ -60,6 +60,9 @@ Vernacular commands
   `tactics/`. In all cases adapting is a matter of changing the module
   name.
 
+- Part of the Internal API in `vernac/comInductive` has been deprecated
+  because it appeared to have no more users.
+
 ### Unit testing
 
   The test suite now allows writing unit tests against OCaml code in the Coq
--- a/vernac/comInductive.mli
+++ b/vernac/comInductive.mli
@@ -51,13 +51,16 @@ type structured_one_inductive_expr = {
   ind_arity : constr_expr;
   ind_lc : (Id.t * constr_expr) list
 }
+[@@ocaml.deprecated "This API is internal"]
 
 type structured_inductive_expr =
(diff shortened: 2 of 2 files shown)


ejgallego (commented):
This crossed my head when doing the command split; however, I was conservative due to a function not having any users doesn't mean it _should not_ have them.

In particular, I wouldn't be surprised if some plugin is reimplementing the functionality there and just the connection was lost.

Other than that, interpreting inductives is a useful operation and in fact it looks to me that it should be moved to `interp`, tho that would require a bit of refactoring in order to properly handle the notation state.

But it could also be argued that inductive interpretation is internal to `vernac` and that should not be exported, I dunno.

SkySkimmer:
>A note: deprecating the types structured_one_inductive_expr and structured_inductive_expr raises warnings when they are referenced in the following deprecated functions.

You can put `[@@@ocaml.warning "-3"]` before the deprecated stuff to disable the warning. See eg engine/universes.mli

herbelin (commented):
This can be seen as internal, but at the same time not necessarily unexpected functions. Maybe they can be kept, together  with a comment, as they are somehow the side-effect free components of `do_mutual_inductive`, and valuable as such.

My 2c.

jashug:
I have no specific need for this PR. I noticed it while working on #7703, but if the general feeling is that these are a worthy part of the API, lets close.

PR #7703, opened 2018-06-05
Add an option to force parameters to be uniform

jashug:
**Kind:** feature

Supersedes #7439.

We support an option `Set Uniform Inductive Parameters` that abstracts inductives over their parameters before type-checking the constructors. This allows to write ex.
```
Inductive list (A : Type) := nil : list | cons : A -> list -> list.
```
This patch does not include earlier syntax which allowed for `Uniform Inductive` and `NonUniform Inductive`, as that can wait for attributes to be implemented.

- [x] Corresponding documentation was added / updated.
- [x] Entry added in CHANGES.

### DIFF
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -905,6 +905,32 @@ Parametrized inductive types
      sort for the inductive definition and will produce a less convenient
      rule for case elimination.
 
+.. opt:: Uniform Inductive Parameters
+
+     When this option is set (it is off by default),
+     inductive definitions are abstracted over their parameters
+     before typechecking constructors, allowing to write:
+
+     .. coqtop:: all undo
+
+        Set Uniform Inductive Parameters.
+        Inductive list3 (A:Set) : Set :=
+        | nil3 : list3
+        | cons3 : A -> list3 -> list3.
+
+     This behavior is essentially equivalent to starting a new section
+     and using :cmd:`Context` to give the uniform parameters, like so
+     (cf. :ref:`section-mechanism`):
+
+     .. coqtop:: all undo
+
+        Section list3.
+        Context (A:Set).
+        Inductive list3 : Set :=
+        | nil3 : list3
+        | cons3 : A -> list3 -> list3.
+        End list3.
+
 .. seealso::
    Section :ref:`inductive-definitions` and the :tacn:`induction` tactic.
 
--- a/vernac/comInductive.ml
+++ b/vernac/comInductive.ml
@@ -146,7 +146,6 @@ let interp_cstrs env sigma impls mldata arity ind =
   let sigma, (ctyps'', cimpls) =
     on_snd List.split @@
     List.fold_left_map (fun sigma l ->
-        on_snd (on_fst EConstr.Unsafe.to_constr) @@
         interp_type_evars_impls env sigma ~impls l) sigma ctyps' in
   sigma, (cnames, ctyps'', cimpls)
 
@@ -327,14 +326,17 @@ let check_param = function
 | CLocalPattern {CAst.loc} ->
     Loc.raise ?loc (Stream.Error "pattern with quote not allowed here")
 
-let interp_mutual_inductive (paramsl,indl) notations cum poly prv finite =
+let interp_mutual_inductive_gen env0 (uparamsl,paramsl,indl) notations cum poly prv finite =
   check_all_names_different indl;
   List.iter check_param paramsl;
-  let env0 = Global.env() in
+  if not (List.is_empty uparamsl) && not (List.is_empty notations)
+  then user_err (str "Inductives with uniform parameters may not have attached notations.");
   let pl = (List.hd indl).ind_univs in
   let sigma, decl = interp_univ_decl_opt env0 pl in
+  let sigma, (uimpls, ((env_uparams, ctx_uparams), useruimpls)) =
+    interp_context_evars env0 sigma uparamsl in
   let sigma, (impls, ((env_params, ctx_params), userimpls)) =
-    interp_context_evars env0 sigma paramsl
+    interp_context_evars ~impl_env:uimpls env_uparams sigma paramsl
   in
   let indnames = List.map (fun ind -> ind.ind_name) indl in
 
@@ -346,15 +348,15 @@ let interp_mutual_inductive (paramsl,indl) notations cum poly prv finite =
   let sigma, arities = List.fold_left_map (fun sigma -> interp_ind_arity env_params sigma) sigma indl in
 
   let fullarities = List.map (fun (c, _, _) -> EConstr.it_mkProd_or_LetIn c ctx_params) arities in
-  let env_ar = push_types env0 indnames fullarities in
+  let env_ar = push_types env_uparams indnames fullarities in
   let env_ar_params = EConstr.push_rel_context ctx_params env_ar in
 
   (* Compute interpretation metadatas *)
   let indimpls = List.map (fun (_, _, impls) -> userimpls @
     lift_implicits (Context.Rel.nhyps ctx_params) impls) arities in
   let arities = List.map pi1 arities and aritypoly = List.map pi2 arities in
-  let impls = compute_internalization_env env0 sigma ~impls (Inductive (params,true)) indnames fullarities indimpls in
-  let ntn_impls = compute_internalization_env env0 sigma (Inductive (params,true)) indnames fullarities indimpls in
+  let impls = compute_internalization_env env_uparams sigma ~impls (Inductive (params,true)) indnames fullarities indimpls in
+  let ntn_impls = compute_internalization_env env_uparams sigma (Inductive (params,true)) indnames fullarities indimpls in
   let mldatas = List.map2 (mk_mltype_data sigma env_params params) arities indnames in
 
   let sigma, constructors =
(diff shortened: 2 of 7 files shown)


doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -900,6 +900,23 @@ Parametrized inductive types
      because the conclusion of the type of constructors should be :g:`listw A`
      in both cases.
 
+     This behavior can be turned off with the option :opt:`Uniform Inductive Parameters`.
```
Zimmi48: Since you are defining the option at this point, you should rather write:

```
.. opt:: Uniform Inductive Parameters

   When this option is set (it is off by default)...

```

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -900,6 +900,23 @@ Parametrized inductive types
...
+
+     .. coqtop:: all
+
+        Set Uniform Inductive Parameters.
+        Inductive list3 (A:Set) : Set :=
+        | nil3 : list3
+        | cons3 : A -> list3 -> list3.
+        Unset Uniform Inductive Parameters.
+
+     A parametrized inductive definition can be defined using annotations
```
Zimmi48: This paragraph is a copy from the one below. Probably a rebasing issue.

lib/flags.mli
```diff
@@ -92,6 +92,10 @@ val is_universe_polymorphism : unit -> bool
 val make_polymorphic_inductive_cumulativity : bool -> unit
 val is_polymorphic_inductive_cumulativity : unit -> bool
 
+(** Global uniform inductive parameters flag *)
+val make_uniform_inductive_parameters : bool -> unit
+val is_uniform_inductive_parameters : unit -> bool
```
Zimmi48: @ejgallego Do you want to comment on the appropriate location for defining a new flag / option? I seem to recall that we prefer to define options at the point where they are used nowadays.

gares:
Curiosity: the API provided by `elpi` to define inductive types uses this *very same convention*: parameters are not to be repeated (indeed they cannot vary). Conversely non-uniform parameters are given the syntax of indexes (they can vary and they must be provided), that is non-uniform parameters are uniform indexes.

vernac/comInductive.ml
```diff
@@ -328,14 +327,17 @@ let check_param = function
 | CLocalPattern {CAst.loc} ->
     Loc.raise ?loc (Stream.Error "pattern with quote not allowed here")
 
-let interp_mutual_inductive (paramsl,indl) notations cum poly prv finite =
+let interp_mutual_inductive_gen env0 (uparamsl,paramsl,indl) notations cum poly prv finite =
   check_all_names_different indl;
   List.iter check_param paramsl;
-  let env0 = Global.env() in
+  if not (List.is_empty uparamsl) && not (List.is_empty notations)
+  then user_err (str "Inductives with uniform parameters may not have attached notations.");
```
gares: May I ask if there is any deep problem here, or it is just lack of time? @herbelin do you think it is hard to support this?

mattam82:
Definitely like the feature and it sounds simpler than the first one involving unification. I’d like to have notations working with those as well. I suppose it’s a matter of putting the right number of underscores in the « exported » notations at each inductive occurrence. One extension option I’m immediately thinking of is to make the parameters automatically maximally implicit in the constructors, but that’s just a suggestion.

SkySkimmer:
Would it be useful to have a syntax for non-uniform parameters when the option is set? eg
```coq
Inductive Acc {A} (R : A -> A -> Prop) | x := Acc_intro : (forall y, R y x -> Acc R y) -> Acc R x.
```
with the pipe indicating the end of uniform parameters.

gares:
> Would it be useful to have a syntax

I'm sorry but this is off topic. The option is to no pass `R` (and not just `A`), and is orthogonal to `|` to separate parameters from non-unif parameters/indexes

vernac/comInductive.ml
herbelin: Hard to support notations? If the notation for the type involves at least one of the arguments that would not be mentioned as in e.g. the case of `\/`, or `/\`, the notation would stop to make sense with this at least one argument not being mentioned.

andres-erbsen:
How does this feature compare to using a new section and Context variables for uniform parameters?

Zimmi48:
Good question. I suppose the two are equivalent and the documentation should probably make this explicit.

jashug:
Thank you all for the feedback.

@gares @herbelin For notations, yes, the issue is that they have to be abstracted over the parameters. I don't know how to do substitutions in constr_expr, but I suppose you could do so. I think you'll still end up with some weird edge cases like
```
Ltac get_head x := match x with ?f _ => get_head f | _ => x end.
Ltac return_head x := let x := get_head x in exact x.
Inductive Foo (n : nat) := C : # where "#" := (ltac:(return_head Foo)).
```
If my understanding is correct, this will work to define `C` with `#` meaning `Foo : Type`, but outside `#` will mean `Foo : nat -> Type`.
The issue is almost exactly the same as having notations survive sections. I suppose you could have the notation only used for defining the constructors, and not be declared globally.

@mattam82 If you use `Set Implicit Arguments. Set Contextual Implicit. Set Maximal Implicit Insertion.` then the parameters should be set maximal implicit for constructors. I can add that to the documentation if you think it would be useful.
I had intended to keep pushing for the unification approach as well. Unification can unify the parameters of the conclusion with the input parameters whether uniform or not, and this is a more principled approach than the `complete_conclusion` removed in #7254. #7254 loses the heuristic that unspecified implicit parameters (curly braces) to recursive arguments are the same as the input parameters. For non-uniform inductive parameters, that heuristic is sometimes wrong, and overrides the correct solution found by unification. This PR attempts to let users be explicit about when they are dealing with uniform parameters and when not.

@SkySkimmer The first commit introduces code to support such a split. @gares proposed a different syntax variant in #7439. If you can come up with a syntax that makes @ppedrot happy, I'm all for it (personally I'd rather have such a way of specifying both uniform and non-uniform parameters in one line (and it might be adaptable to the `Fixpoint` command as well?))

@andres-erbsen The only difference is that type inference gets the constructors and indices to work with this way, so you can say e.x. `Inductive Foo A := C : A -> Foo.` and infer that `A` is a `Type`, while `Context A.` has nothing to go on. I'll add that to the documentation.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -905,6 +905,32 @@ Parametrized inductive types
...
+     .. coqtop:: all
+
+        Set Uniform Inductive Parameters.
+        Inductive list3 (A:Set) : Set :=
+        | nil3 : list3
+        | cons3 : A -> list3 -> list3.
+        Unset Uniform Inductive Parameters.
+
+     This behavior is essentially equivalent to starting a new section
+     and using `Context` to give the uniform parameters, like so:
```
Zimmi48: The proper way to quote a command is:

```
:cmd:`Context`
```

lib/flags.mli
ejgallego: Well indeed this is documented at the top of `Flags.mli`.

Please move the flags to the proper module. Also, don't pollute the namespace with `make/is` fucntions.

doc/sphinx/language/gallina-specification-language.rst
jashug: Thanks. I hadn't seen `doc/sphnix/README`, that gave me a better sense of how the new documentation works. I fixed this, and also used `undo` for the examples.

lib/flags.mli
jashug: Sorry about that, I was blindly mimicking the cumulativity flag above, and didn't read `flags.mli`. I moved the option to `vernacentries.ml`, which is where it is used, but if it would be better in `ComInductive` let me know.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -905,6 +905,32 @@ Parametrized inductive types
...
+     inductive definitions are abstracted over their parameters
+     before typechecking constructors, allowing to write:
+
+     .. coqtop:: all undo
+
+        Set Uniform Inductive Parameters.
+        Inductive list3 (A:Set) : Set :=
+        | nil3 : list3
+        | cons3 : A -> list3 -> list3.
+        Unset Uniform Inductive Parameters.
```
Zimmi48: Since you use `undo`, the `Unset` command here is inessential.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -905,6 +905,32 @@ Parametrized inductive types
...
+
+     This behavior is essentially equivalent to starting a new section
```
Zimmi48: You could perhaps add ```(cf. :ref:`section-mechanism`)``` here.

Zimmi48 (approved):
I approve the documentation part.

Note that there is another example of defining inductive types within sections in the "mutual inductive types" section below. Maybe this part of the doc could be made to use uniform inductive types instead.

vernac/vernacentries.ml
```diff
@@ -542,6 +542,13 @@ let should_treat_as_cumulative cum poly =
     else user_err Pp.(str "The NonCumulative prefix can only be used in a polymorphic context.")
   | None -> poly && Flags.is_polymorphic_inductive_cumulativity ()
 
+let uniform_inductive_parameters = ref false
+
+let should_treat_as_uniform _ =
```
ppedrot: Use a `()` pattern here.

vernac/comInductive.mli
```diff
@@ -19,9 +19,14 @@ open Decl_kinds
...
 
+type uniform_inductive_flag =
+  | UniformParameters
+  | NonUniformParameters
+
 val do_mutual_inductive :
   (one_inductive_expr * decl_notation list) list -> cumulative_inductive_flag ->
-  polymorphic -> private_flag -> Declarations.recursivity_kind -> unit
+  polymorphic -> private_flag -> uniform_inductive_flag ->
+  Declarations.recursivity_kind -> unit
```
ppedrot: Might be worth giving a label to this argument. The other ones are lacking it, but it doesn't justify not doing it systematically when introducing new arguments.

ppedrot (dismissed):

jashug:
Addressed changes requested by @ppederot

jashug:
#7254 is looking good, with buy-in from @JasonGross, so I believe this PR is just waiting on approval from @mattam82. The gitlab failure of Equations is almost certainly spurious (it was green before), but I can't restart it myself.

Zimmi48:
@jashug Yes, exit code 137 always mean spurious failure (runner system failure in GitLab terminology). I've restarted the job. You can request to join the Coq organization on GitLab https://gitlab.com/coq to be able to restart failed job.

mattam82:
I think you’ll have to rebase first to make equations go through.

mattam82 (approved):

jashug:
Rebased, all green.

JasonGross:
Should I merge the overlay into fiat-crypto now?

jashug:
Well, I believe that there is general agreement to move forward with this and #7254, so it seems unlikely that the overlay will change and it is backwards compatible. Do note though that the overlay is not necessary for this PR, but only for #7254.

PR #7715, opened 2018-06-05
Simplify the cooking of primitive projections.

ppedrot:
There is no need to expand a primitive projection with the section parameters and universes, for one good reason: they are never applied neither to parameters nor universe instances.

### DIFF
--- a/kernel/cooking.ml
+++ b/kernel/cooking.ml
@@ -126,16 +126,15 @@ let expmod_constr cache modlist c =
 	    | Not_found -> Constr.map substrec c)
 
       | Proj (p, c') ->
-          (try 
-	     let p' = share_univs (ConstRef (Projection.constant p)) Univ.Instance.empty modlist in
-	     let make c = Projection.make c (Projection.unfolded p) in
-	     match kind p' with
-	     | Const (p',_) -> mkProj (make p', substrec c')
-	     | App (f, args) -> 
-	       (match kind f with 
-	       | Const (p', _) -> mkProj (make p', substrec c')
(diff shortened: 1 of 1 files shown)


kernel/cooking.ml
```diff
@@ -126,16 +126,15 @@ let expmod_constr cache modlist c =
...
-	     | Const (p',_) -> mkProj (make p', substrec c')
-	     | App (f, args) -> 
-	       (match kind f with 
-	       | Const (p', _) -> mkProj (make p', substrec c')
-	       | _ -> assert false)
-	     | _ -> assert false
+          (try
+            (** No need to expand parameters or universes for projections *)
+            let map cst =
+              let _ = Cmap.find cst (fst modlist) in
```
SkySkimmer: What's this `find` for?

ppedrot: Only constants that are in the substitution need to be expanded. If this check is not done, `pop_con` might trigger an anomaly.

SkySkimmer (approved):
I'd prefer something like
```ocaml
if Cmap.mem (Projection.constant p) (fst modlist) then
  let p = pop_proj p in 
  let c' = substrec c' in 
  mkProj (p, c')
else Constr.map substrec c
```
(with a new `pop_proj` combinator) over exception flow control but it's not a blocker.

ppedrot:
@SkySkimmer the goal is to reduce the dependency in the kernel on functions that rely on the constant-using encoding of projections, so your patch goes against this.

maximedenes (approved):

PR #7736, opened 2018-06-07
[ci] Fix GeoCoq after https://github.com/GeoCoq/GeoCoq/issues/12

ejgallego:
We build a bit of math-comp so it works, also patch in the old-style
until the upstream issue is fixed.

Let's see how this works.

### DIFF
--- a/dev/ci/ci-common.sh
+++ b/dev/ci/ci-common.sh
@@ -93,17 +93,26 @@ install_ssreflect()
   git_checkout "${mathcomp_CI_BRANCH}" "${mathcomp_CI_GITURL}" "${mathcomp_CI_DIR}"
 
   ( cd "${mathcomp_CI_DIR}/mathcomp" && \
-    sed -i.bak '/ssrtest/d'     Make && \
-    sed -i.bak '/odd_order/d'   Make && \
-    sed -i.bak '/all\/all.v/d'  Make && \
-    sed -i.bak '/character/d'   Make && \
-    sed -i.bak '/real_closed/d' Make && \
-    sed -i.bak '/solvable/d'    Make && \
-    sed -i.bak '/field/d'       Make && \
-    sed -i.bak '/fingroup/d'    Make && \
-    sed -i.bak '/algebra/d'     Make && \
-    make Makefile.coq && make -f Makefile.coq all && make install )
+    make Makefile.coq && \
+    make -f Makefile.coq ssreflect/all_ssreflect.vo && \
+    make -f Makefile.coq install )
 
   echo -en 'travis_fold:end:ssr.install\\r'
+
+}
+
+# this installs just the ssreflect + algebra library of math-comp
+install_ssralg()
+{
+  echo 'Installing ssralg' && echo -en 'travis_fold:start:ssralg.install\\r'
+
+  git_checkout "${mathcomp_CI_BRANCH}" "${mathcomp_CI_GITURL}" "${mathcomp_CI_DIR}"
+
+  ( cd "${mathcomp_CI_DIR}/mathcomp" && \
+    make Makefile.coq && \
+    make -f Makefile.coq algebra/all_algebra.vo && \
+    make -f Makefile.coq install )
+
+  echo -en 'travis_fold:end:ssralg.install\\r'
 
 }
--- a/dev/ci/ci-geocoq.sh
+++ b/dev/ci/ci-geocoq.sh
@@ -7,4 +7,6 @@ GeoCoq_CI_DIR="${CI_BUILD_DIR}/GeoCoq"
 
 git_checkout "${GeoCoq_CI_BRANCH}" "${GeoCoq_CI_GITURL}" "${GeoCoq_CI_DIR}"
 
+install_ssralg
+
 ( cd "${GeoCoq_CI_DIR}" && ./configure.sh && make )


SkySkimmer:
What's with this crazy sed hack, can't we just `make algebra` or something?

Zimmi48:
I would indeed suppose that math-comp authors have support for this kind of things.

ejgallego:
The hack is not so crazy IMO.

But indeed, we could use the opam infrastructure I guess; I guess againt that I originally did like that because I wanted full control over the general `Make` file.

ejgallego:
Anyways the hack has been there for ages so I'd rather not address it in this PR which is meant to be a "quick" fix.

ejgallego:
Also note that the fix should work even after the authors fix the upstream version.

ejgallego:
Something strange is that their Travis suceeded for `master`, maybe Jason packages are outdated?

Zimmi48:
Or maybe, they don't test everything.

ejgallego:
Yeah their running time is suspiciously short.

Zimmi48:
Looks like the sed hack is not working:

```
make[1]: Entering directory '/builds/coq/coq/_build_ci/math-comp/mathcomp'
COQC ssreflect/ssrnotations.v
COQC ssreflect/ssreflect.v
ssreflect/ssrnotations (real: 0.14, user: 0.08, sys: 0.05, mem: 104152 ko)
make[2]: *** No rule to make target 'fingroup/fingroup.vo', needed by 'algebra/ssrnum.vo'.  Stop.
make[2]: *** Waiting for unfinished jobs....
COQC ssreflect/ssrbool.v
ssreflect/ssreflect (real: 0.18, user: 0.14, sys: 0.04, mem: 119436 ko)
ssreflect/ssrbool (real: 0.19, user: 0.12, sys: 0.06, mem: 127068 ko)
Makefile.coq:328: recipe for target 'all' failed
make[1]: *** [all] Error 2
```

SkySkimmer:
Still not working. The other sed is failing now.

gares:
> What's with this crazy sed hack, can't we just make algebra or something?

yYou can `make -C algebra` but you have to `make -C ssreflect && make -C fingroup` first.

dev/ci/ci-common.sh
```diff
@@ -107,3 +107,24 @@ install_ssreflect()
...
 }
+
+# this installs just the ssreflect + algebra library of math-comp
+install_ssralg()
+{
+  echo 'Installing ssreflect' && echo -en 'travis_fold:start:ssr.install\\r'
+
+  git_checkout "${mathcomp_CI_BRANCH}" "${mathcomp_CI_GITURL}" "${mathcomp_CI_DIR}"
+
+  ( cd "${mathcomp_CI_DIR}/mathcomp" && \
```
gares: no please

gares (dismissed):

dev/ci/ci-common.sh
ejgallego: This hack was already in place and thus orthogonal to the purpose of this PR. Please let's move the discussion to its own issue  / PR.

ejgallego: This hack was already in place and thus orthogonal to the purpose of this PR. Please let's move the discussion to its own issue  / PR.

gares: This is new code. Fix *this* pr please. The fix is trivial. I m sorry I did not suggest this before.

dev/ci/ci-geocoq.sh
```diff
@@ -7,4 +7,7 @@ GeoCoq_CI_DIR="${CI_BUILD_DIR}/GeoCoq"
 
 git_checkout "${GeoCoq_CI_BRANCH}" "${GeoCoq_CI_GITURL}" "${GeoCoq_CI_DIR}"
 
+install_ssralg
+
+sed -i.bak "1iSet Nested Proofs Allowed\n" "${GeoCoq_CI_DIR}/Meta_theory/Models/hilbert_to_tarski.v"
```
Zimmi48: This is missing a dot and is thus making the build fail:

```
Warning: There is no option Nested Proofs Allowed Require Import Setoid.
[unknown-option,option]
File "./Meta_theory/Models/hilbert_to_tarski.v", line 231, characters 0-58:
Error:
Nested proofs are not allowed unless you turn option Nested Proofs Allowed on.
```

Zimmi48 (commented):
Given that this PR needs to be fixed once more, and given that an author of math-comp told you how to replace the horrible sed hack with a cleaner way, I suggest you do follow the recommendation while fixing the PR.

Zimmi48:
The nested lemma has been removed so this part of the fix is not needed anymore. However, I don't understand why you have still not taken into account Enrico's proposal to replace the sed hack.

ejgallego:
> I don't understand why you have still not taken into account Enrico's proposal to replace the sed hack.

I already explained it; the `sed` hack has been in place for a long time and it is orthogonal to this PR and not trivial to eliminate; please open a bug if you want it replaced.

Zimmi48:
I don't understand. We are talking about the sed hack that you are introducing, not the one that was already present. And Enrico provided you a replacement. If this replacement doesn't actually work, I'm fine with the current state of the PR. But if it does work, then why not use it? I suppose you either think that we are asking you to also remove the preexisting sed hack or you see why Enrico's drop in replacement won't be enough but you didn't explain why.

ejgallego:
Just a matter of lack of time @Zimmi48

Zimmi48:
Passed here: https://gitlab.com/Zimmi48/coq/-/jobs/73927070

Sorry @ejgallego I had to reintroduce the `Set Nested Proofs Allowed` option I had asked you to remove. Turns out that, contrarily to what was claimed in GeoCoq/GeoCoq#12, the nested lemma has not been removed (cc. GeoCoq/GeoCoq@e140b35f2a18446e8164b7b9a126aba147c38bb9).

Zimmi48:
Wait a minute.

Zimmi48:
So 83bfdcf can be dropped again after all: the nested lemma had not been removed in GeoCoq/GeoCoq@a3ebcab73d65a0baa6f889baa8ec0ec26dc795ca but it has in GeoCoq/GeoCoq@e140b35f2a18446e8164b7b9a126aba147c38bb9. However, this commit didn't remove the corresponding proof so GeoCoq must be broken again right now.

ejgallego:
Yeah we are having trouble with this PR XD

Zimmi48:
Contrary to what I was thinking, GeoCoq is fixed and this PR is finally ready to merge @SkySkimmer (commits squashed).

SkySkimmer (approved):

PR #7745, opened 2018-06-08
Make type Environ.globals abstract + simplify Environ.retroknowledge

SkySkimmer:
It feels a bit cleaner this way.

### DIFF
--- a/dev/top_printers.ml
+++ b/dev/top_printers.ml
@@ -232,7 +232,7 @@ let ppenv e = pp
 let ppenvwithcst e = pp
   (str "[" ++ pr_named_context_of e Evd.empty ++ str "]" ++ spc() ++
    str "[" ++ pr_rel_context e Evd.empty (rel_context e) ++ str "]" ++ spc() ++
-   str "{" ++ Cmap_env.fold (fun a _ s -> Constant.print a ++ spc () ++ s) (Obj.magic e).env_globals.env_constants (mt ()) ++ str "}")
+   str "{" ++ Environ.fold_constants (fun a _ s -> Constant.print a ++ spc () ++ s) e (mt ()) ++ str "}")
 
 let pptac = (fun x -> pp(Ltac_plugin.Pptactic.pr_glob_tactic (Global.env()) x))
 
--- a/kernel/environ.ml
+++ b/kernel/environ.ml
@@ -55,7 +55,8 @@ type globals = {
   env_projections : projection_body Cmap_env.t;
   env_inductives : mind_key Mindmap_env.t;
   env_modules : module_body MPmap.t;
-  env_modtypes : module_type_body MPmap.t}
+  env_modtypes : module_type_body MPmap.t;
+}
 
 type stratification = {
   env_universes : UGraph.t;
@@ -86,7 +87,7 @@ type rel_context_val = {
 }
 
 type env = {
-  env_globals       : globals;           (* globals = constants + inductive types + modules + module-types *)
+  env_globals       : globals;
   env_named_context : named_context_val; (* section variables *)
   env_rel_context   : rel_context_val;
   env_nb_rel        : int;
@@ -207,6 +208,9 @@ let lookup_named_val id env =
 
 let lookup_named_ctxt id ctxt =
   fst (Id.Map.find id ctxt.env_named_map)
+
+let fold_constants f env acc =
+  Cmap_env.fold (fun c (body,_) acc -> f c body acc) env.env_globals.env_constants acc
 
 (* Global constants *)
 
--- a/kernel/environ.mli
+++ b/kernel/environ.mli
@@ -318,6 +318,7 @@ open Retroknowledge
 (** functions manipulating the retroknowledge 
     @author spiwack *)
 val retroknowledge : (retroknowledge->'a) -> env -> 'a
+[@@ocaml.deprecated "Use the record projection."]
 
 val registered : env -> field -> bool
 
--- a/kernel/environ.mli
+++ b/kernel/environ.mli
@@ -46,13 +46,8 @@ type constant_key = constant_body * (link_info ref * key)
 
 type mind_key = mutual_inductive_body * link_info ref
 
-type globals = {
-  env_constants : constant_key Cmap_env.t;
-  env_projections : projection_body Cmap_env.t;
-  env_inductives : mind_key Mindmap_env.t;
-  env_modules : module_body MPmap.t;
-  env_modtypes : module_type_body MPmap.t
-}
+type globals
+(** globals = constants + projections + inductive types + modules + module-types *)
 
 type stratification = {
   env_universes : UGraph.t;
@@ -70,7 +65,7 @@ type rel_context_val = private {
 }
 
 type env = private {
-  env_globals       : globals;           (* globals = constants + inductive types + modules + module-types *)
+  env_globals       : globals;
   env_named_context : named_context_val; (* section variables *)
   env_rel_context   : rel_context_val;
   env_nb_rel        : int;
@@ -174,6 +169,9 @@ val reset_with_named_context : named_context_val -> env -> env
 
 (** This removes the [n] last declarations from the rel context *)
 val pop_rel_context : int -> env -> env
+
+(** Useful for printing *)
+val fold_constants : (Constant.t -> constant_body -> 'a -> 'a) -> env -> 'a -> 'a
 
 (** {5 Global constants }
   {6 Add entries to global environment } *)
--- a/kernel/nativecode.ml
+++ b/kernel/nativecode.ml
@@ -1845,7 +1845,7 @@ and apply_fv env sigma univ (fv_named,fv_rel) auxdefs ml =
   in
   let auxdefs = List.fold_right get_rel_val fv_rel auxdefs in
   let auxdefs = List.fold_right get_named_val fv_named auxdefs in
-  let lvl = Context.Rel.length env.env_rel_context.env_rel_ctx in
+  let lvl = Context.Rel.length (rel_context env) in
   let fv_rel = List.map (fun (n,_) -> MLglobal (Grel (lvl-n))) fv_rel in
   let fv_named = List.map (fun (id,_) -> MLglobal (Gnamed id)) fv_named in
   let aux_name = fresh_lname Anonymous in
@@ -1854,7 +1854,7 @@ and apply_fv env sigma univ (fv_named,fv_rel) auxdefs ml =
 and compile_rel env sigma univ auxdefs n =
   let open Context.Rel.Declaration in
   let decl = lookup_rel n env in
-  let n = List.length env.env_rel_context.env_rel_ctx - n in
+  let n = List.length (rel_context env) - n in
   match decl with
   | LocalDef (_,t,_) ->
       let code = lambda_of_constr env sigma t in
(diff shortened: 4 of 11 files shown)


ejgallego (commented):
I still think that we gain is relative [I am skeptical about us being able to change the internals without altering the ML signature] and API wise makes completion / discovery a bit worse.

Anyways I feel that the private record vs abstract type has the potential to become the Quevedo vs Góngora battle of our times, so my next comment on this issue will be of the form of a letter to some obscure journal.

ppedrot (approved):
Small, clean patch. I'm all for it.

dev/top_printers.ml
```diff
@@ -232,7 +232,7 @@ let ppenv e = pp
 let ppenvwithcst e = pp
   (str "[" ++ pr_named_context_of e Evd.empty ++ str "]" ++ spc() ++
    str "[" ++ pr_rel_context e Evd.empty (rel_context e) ++ str "]" ++ spc() ++
-   str "{" ++ Cmap_env.fold (fun a _ s -> Constant.print a ++ spc () ++ s) (Obj.magic e).env_globals.env_constants (mt ()) ++ str "}")
```
SkySkimmer: (5e60af46bdcb5aa487737961859f80181486516b introduced this magic)

ppedrot: Pretty ugly indeed...

kernel/environ.mli
```diff
@@ -319,7 +313,7 @@ val remove_hyps : Id.Set.t -> (Context.Named.Declaration.t -> Context.Named.Decl
 open Retroknowledge
 (** functions manipulating the retroknowledge 
     @author spiwack *)
-val retroknowledge : (retroknowledge->'a) -> env -> 'a
+val retroknowledge : env -> retroknowledge
```
SkySkimmer: @aspiwack what was the reason for writing this function in this way?

aspiwack: Oh! It was a very nice reason! An important reason! A reason that will really impress you! A re…

I don't actually remember. Maybe I thought it was a very clever type for some reason? Maybe I had a particular pattern in mind. Maybe I just didn't realise that I was overcomplicating things.

ppedrot: @aspiwack You seem to like this kind of pattern, I remember that you introduced a similar one for global proof data access. Let me dare a bit of software psychanalysis: I suppose that you're thinking in a richer language where the `'a` return type is constrained to have some property (e.g. `hProp`) so that the accessor looks like a higher induction principle. Somewhere in between your brain and the keyboard this gets compiled on the fly to the unexpressive OCaml language...

aspiwack: Sure. Let's say that. At this point, I'll take any excuse.

SkySkimmer: >@aspiwack You seem to like this kind of pattern, I remember that you introduced a similar one for global proof data access. Let me dare a bit of software psychanalysis: I suppose that you're thinking in a richer language where the 'a return type is constrained to have some property (e.g. hProp) so that the accessor looks like a higher induction principle. Somewhere in between your brain and the keyboard this gets compiled on the fly to the unexpressive OCaml language...

Wouldn't it make more sense to constrain in such a way that the access is global, ie the retroknowledge doesn't leak beyond the function call? A bit like the old Sigma stuff, or the haskell local state monad.

ppedrot: @SkySkimmer OCaml has side-effects! Boo, boo, that breaks almost of the parametricity reasoning...

SkySkimmer:
There's also Obj.magic, and yet we use abstract types.

Gaëtan Gilbert

ppedrot:
@SkySkimmer OK, nobody will believe me on this one but:

> Repeat after me: Obj.magic is not part of the OCaml language.

[Xavier Leroy](https://en.wikiquote.org/wiki/Xavier_Leroy) who happens to have his own Wikiquote page.

maximedenes:
> It feels a bit cleaner this way.

@SkySkimmer can you give a more objective motivation for this change?

SkySkimmer:
No.

maximedenes:
> No.

Ok, then I oppose to this change.

maximedenes:
> Ok, then I oppose to this change.

Apart from the retroknowledge refactoring, btw, which I do find simpler than it was previously.

ppedrot:
@maximedenes I can give one: good practices.

There is no reason to let a type transparent (even only privately so) if it's not going to be used as such. Environments are a pervasive data structure, and it would be a pity for external users to take advantage of implementation details.

Judging from the fact most of the plugins writers outside (and actually even inside sometimes) have no frigging idea of what they're writing and are mainly guided by the types from the API, it looks like a necessary conservation measure to me.

maximedenes:
I don't understand the argument. Don't get me wrong, I do like abstraction, and we all know it is a fundamental principle that makes software development scale.

But in this precise case do we gain abstraction over anything? It's a bit hard to read from the patch: is there any projection that is not exported in the new interface?

maximedenes:
> Judging from the fact most of the plugins writers outside the dev team (and actually even inside sometimes) have no frigging idea of what they're writing

FTR, you can't expect people to have an idea of what they are doing when API documentation is non-existing... IMHO that's what we should be addressing, instead of exposing records through an isomorphism instead of directly and claiming we gain anything by doing it.

ejgallego:
Yeah I feel that usually, this kind of opaqueness is indeed rather a barrier than a help for other developers. Understanding what's on a type is an important practical property.

For example, in the recent printing diff patch, a Pp to HTML function is proposed, however, instead of doing the usual fold over the document type, the developer didn't discover the structure and wrote a very convoluted indirection thru Format.

aspiwack:
> But in this precise case do we gain abstraction over anything? It's a bit hard to read from the patch: is there any projection that is not exported in the new interface?

If there are only projections, then it's motivation enough to have the type be private. A more useful question, in this context, would be, are there field update functions which are not exported?

If they are all exported, then we can ask the question: does it mean that the semantics of `env` is that it is a tuple of all these things.

(for the records, considering how small this patch is I doubt we will find it to be the case. But who knows)

ppedrot:
OK, let me be more precise. I am not really concerned about the fields we really provide through projection wrappers. My main concern is that people start using the `env_global` field accessing directly the maps in there, when there are a family of `lookup_*` functions for that.

In particular and to give an example, I am working on a patch that adds a notion of alias to inductive types, in order to properly implement `Include` and get rid of the canonical part of inductive names. This would change the type of `env_global`, but the lookup functions would be untouched, as they would resolve the alias on the fly. If people start accessing this field directly, it means a less robust code w.r.t. to changes like that.

Also, I should add that find your argument of *this is a change to the kernel that doesn't bring much* is a bid of a bad faith one. May I remind you that you were the one in the first place to recently expose the internals of environments, and as such, if we did value kernel conservativity, the burden of the argumentation would be upon you?

aspiwack:
> the type is already private. This change is making it opaque. Are you saying it is enough?

No, I had just missed that it was private. Sorry for the confusion. To be honest, I don't think it matters either way, it's neither much work to redo or revert.

maximedenes:
Since there was a bit of meta-communication around this PR on gitter, I would like to make something clear: this patch has a few cons (mentioned by @ejgallego ), and before the last message of @ppedrot, no concrete reason was given in favor of it. So I thought the balance is in favor of integrating the `env` part of the patch (there is still the retroknowledge simplification). I did not mean to be aggressive at all, and I'll revise my judgement if concrete motivations can be given, of course. In particular, I didn't mean to resort to a "conservativity" argument.

Sorry for those misunderstandings, I've been lacking sleep for a while now. Nothing I can do about it, and it's putting my in the uncomfortable situation of having to review PRs while being exhausted.

@ppedrot about your example, I'm missing something (probably again because of lack of sleep): this patch exposes `globals : env -> globals`. How is that going to prevent the scenario that you are describing? I thought it was the `globals` type that should be made opaque, instead.

herbelin:
😩=😓+😒
long😴=😄

SkySkimmer:
> I thought it was the globals type that should be made opaque, instead.

Updated.

maximedenes:
> about your example, I'm missing something (probably again because of lack of sleep): this patch exposes globals : env -> globals. How is that going to prevent the scenario that you are describing? I thought it was the globals type that should be made opaque, instead.

@ppedrot ping

ppedrot:
There is only one place using globals, and it is using in a very hackish way. This code should rather be browsing the module structure rather than the structureless global blob.

Also, making a function available doesn't mean we endorse its use. Often people go to the shortest solution, so that we should strive for easiness of correct code and clumsiness of the undesirable one.

kernel/nativecode.ml
```diff
@@ -1841,7 +1841,7 @@ and apply_fv env sigma univ (fv_named,fv_rel) auxdefs ml =
   in
   let auxdefs = List.fold_right get_rel_val fv_rel auxdefs in
   let auxdefs = List.fold_right get_named_val fv_named auxdefs in
-  let lvl = Context.Rel.length env.env_rel_context.env_rel_ctx in
+  let lvl = Context.Rel.length (rel_context env) in
```
maximedenes: I think I prefer the original code in this file. In fact, I believe we could remove the aliases of the projections, but it's ok if we don't do it now.

kernel/environ.ml
```diff
@@ -675,11 +679,11 @@ let remove_hyps ids check_context check_value ctxt =
 open Retroknowledge
 
 (* lifting of the "get" functions works also for "mem"*)
-let retroknowledge f env =
-  f env.retroknowledge
+let retroknowledge env =
```
maximedenes: This function doesn't seem useful anymore, I would deprecate it in favor of the projection.

kernel/nativelambda.ml
```diff
@@ -441,7 +441,7 @@ let is_lazy prefix t =
 	let entry = mkInd (fst c) in
 	(try
 	    let _ =
-	      Retroknowledge.get_native_before_match_info (!global_env).retroknowledge
+              Retroknowledge.get_native_before_match_info (retroknowledge !global_env)
```
maximedenes: I prefer the original code in this file too.

maximedenes (dismissed):
Welcome cleanup, thanks! Only a few nits that I'd like to see addressed.

kernel/nativecode.ml
SkySkimmer: >I think I prefer the original code in this file. In fact, I believe we could remove the aliases of the projections, but it's ok if we don't do it now.

Why?

maximedenes: This argument was made before, but we tend to have a lot of useless aliases that make the code base larger and harder to browse. If they have no point, why keep them?

SkySkimmer: Here going through a function is simpler than going through 2 projection layers.

SkySkimmer:
Changed the retroknowledge thing.

kernel/nativecode.ml
maximedenes: Ok, keep this one if you prefer it.

SkySkimmer:
@mattam82 ping

maximedenes (approved):

maximedenes:
@SkySkimmer I can merge it, this is a kernel change, the rest is just propagation.

maximedenes:
Can you rebase to trigger a recent CI build? I can merge today as soon as it has run.

ejgallego:
Now it is impossible to inspect the globals present in the environment, either for debug or serialization use.... this breaks SerAPI at least; really I don't get the point of these changes.

Additionally, this seems like an API regression to me.

Note that the faux opaqueness also makes very hard to program in a robust way; with the previous verson of the code, I could match the record and use a warning to be sure that my pattern match is complete; now, how can the compiler check that I am inspecting the type in an injective way :S

PR #7748, opened 2018-06-08
Add a bit of doc to EConstr.decompose_lam*

SkySkimmer:


ppedrot (commented):
The documentation is not quite true, these functions are also skipping casts.

SkySkimmer:
changed

Gaëtan Gilbert

herbelin:
An alternative is to separate concerns and to stop skipping casts. (Maybe too adventurous though.)

ppedrot (approved):

PR #7749, opened 2018-06-08
[doc] Disable smartquotes conversion

### LINKED ISSUE #7742: Sphinx docs: Quotes are turned into unicode quotes, leading to syntax error when copy-pasting
#### Version

8.8.0

#### Operating system
Not applicable

The example `Notation` commands at https://coq.inria.fr/refman/user-extensions/syntax-extensions.html (and probably most of the quotes in the entire documentation) are writen in a way that one cannot copy-paste them. For example:
```
Notation “A /\ B” := (and A B).
```
The correct way would be
```
Notation "A /\ B" := (and A B).
```
Notice the different quotes.

This is a regression compared to the old documentation at https://coq.inria.fr/distrib/8.7.2/refman/syntax-extensions.html where commands could be copy-pasted from the documentation into Coq.

cpitclaudel:
**Kind:** documentation / bug fix / infrastructure.

Fixes / closes GH-7742

### DIFF
--- a/doc/sphinx/conf.py
+++ b/doc/sphinx/conf.py
@@ -201,9 +201,9 @@ html_static_path = ['_static']
 # The empty string is equivalent to '%b %d, %Y'.
 #html_last_updated_fmt = None
 
-# If true, SmartyPants will be used to convert quotes and dashes to
-# typographically correct entities.
-html_use_smartypants = False # FIXME wrap code in <code> tags, otherwise quotesget converted in there
+# FIXME: this could be re-enabled after ensuring that smart quotes are locally
+# disabled for all relevant directives
+smartquotes = False
 
 # Custom sidebar templates, maps document names to template names.
 #html_sidebars = {}


doc/sphinx/conf.py
```diff
@@ -201,9 +201,9 @@
 # The empty string is equivalent to '%b %d, %Y'.
 #html_last_updated_fmt = None
 
-# If true, SmartyPants will be used to convert quotes and dashes to
-# typographically correct entities.
-html_use_smartypants = False # FIXME wrap code in <code> tags, otherwise quotesget converted in there
+# FIXME: this could be re-enabled after ensuring that smart quotes are locally
+# disabled for all relevant directives
```
Zimmi48: Indeed, it's a bit sad to lose smart quotes in many places. But as a quick fix it makes sense.

maximedenes (approved):
LGTM as a quick fix indeed

PR #7750, opened 2018-06-08
Handle mutual records in the kernel

ppedrot:
This PR adapts the data structures used in the kernel in order to handle mutual primitive records. There is no user-facing support yet, this will come with a subsequent PR.

Essentially three changes:
- the `proj_ind` field now contains the full inductive instead of just the `MutInd.t`
- the `mind_record` field contains a more informative type that is generalized to handle mutual records
- the record-data generating function from `Indtypes` has been adapted to the mutual case.

### DIFF
--- a/kernel/entries.ml
+++ b/kernel/entries.ml
@@ -49,9 +49,9 @@ type one_inductive_entry = {
   mind_entry_lc : constr list }
 
 type mutual_inductive_entry = {
-  mind_entry_record : (Id.t option) option; 
-  (** Some (Some id): primitive record with id the binder name of the record
-      in projections.
+  mind_entry_record : (Id.t array option) option;
+  (** Some (Some ids): primitive records with ids the binder name of each
+      record in their respective projections. Not used by the kernel.
       Some None: non-primitive record *)
   mind_entry_finite : Declarations.recursivity_kind;
   mind_entry_params : (Id.t * local_entry) list;
--- a/kernel/nativecode.ml
+++ b/kernel/nativecode.ml
@@ -1990,8 +1990,10 @@ let compile_mind prefix ~interactive mb mind stack =
       Glet (gn, mkMLlam [|c_uid|] code) :: acc
     in
     let projs = match mb.mind_record with
-    | None | Some None -> []
-    | Some (Some (id, kns, pbs)) -> Array.fold_left_i add_proj [] pbs
+    | NotRecord | FakeRecord -> []
+    | PrimRecord info ->
+      let _, _, pbs = info.(i) in
+      Array.fold_left_i add_proj [] pbs
     in
     projs @ constructors @ gtype :: accu :: stack
   in
--- a/kernel/nativecode.ml
+++ b/kernel/nativecode.ml
@@ -1965,6 +1965,7 @@ let compile_mind prefix ~interactive mb mind stack =
     in
     let constructors = Array.fold_left_i add_construct [] ob.mind_reloc_tbl in
     let add_proj j acc pb =
+      let () = assert (eq_ind ind pb.proj_ind) in
       let tbl = ob.mind_reloc_tbl in
       (* Building info *)
       let ci = { ci_ind = ind; ci_npar = nparams;
@@ -1985,7 +1986,7 @@ let compile_mind prefix ~interactive mb mind stack =
       let accu_br = MLapp (MLprimitive Mk_proj, [|get_proj_code i;accu|]) in
       let code = MLmatch(asw,MLlocal cf_uid,accu_br,[|[((ind,1),cargs)],MLlocal ci_uid|]) in
       let code = MLlet(cf_uid, MLapp (MLprimitive Force_cofix, [|MLlocal c_uid|]), code) in
-      let gn = Gproj ("", (pb.proj_ind, j), pb.proj_arg) in
+      let gn = Gproj ("", ind, pb.proj_arg) in
       Glet (gn, mkMLlam [|c_uid|] code) :: acc
     in
     let projs = match mb.mind_record with
@@ -2052,7 +2053,7 @@ let compile_deps env sigma prefix ~interactive init t =
   | Construct (((mind,_),_),u) -> compile_mind_deps env prefix ~interactive init mind
   | Proj (p,c) ->
     let pb = lookup_projection p env in
-    let init = compile_mind_deps env prefix ~interactive init pb.proj_ind in
+    let init = compile_mind_deps env prefix ~interactive init (fst pb.proj_ind) in
     aux env lvl init c
   | Case (ci, p, c, ac) ->
       let mind = fst ci.ci_ind in
--- a/checker/cic.mli
+++ b/checker/cic.mli
@@ -207,7 +207,7 @@ type inline = int option
     always transparent. *)
 
 type projection_body = {
-  proj_ind : MutInd.t;
+  proj_ind : inductive;
   proj_npars : int;
   proj_arg : int;
   proj_type : constr; (* Type under params *)
--- a/checker/cic.mli
+++ b/checker/cic.mli
@@ -252,9 +252,10 @@ type recarg =
 
 type wf_paths = recarg Rtree.t
 
-type record_body = (Id.t * Constant.t array * projection_body array) option
-    (* The body is empty for non-primitive records, otherwise we get its
-       binder name in projections and list of projections if it is primitive. *)
+type record_info =
+| NotRecord
+| FakeRecord
+| PrimRecord of (Id.t * Constant.t array * projection_body array) array
 
 type regular_inductive_arity = {
   mind_user_arity : constr;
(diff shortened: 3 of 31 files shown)


gares:
Is there a usecase?

ppedrot:
@gares To fully implement deprecation of positive coinductive types, we also need to provide mutual negative coinductive types. I don't know if this is widely used in the wild, but there are a few tests in the test-suite. Furthermore there is no reason not to provide this facility.

maximedenes:
> Furthermore there is no reason not to provide this facility.

This is almost always a bad argument. When introducing a new feature, I recommend comparing the benefits (new use cases covered, etc.) to the cost (implementation, maintenance of the code, documentation, maintenance of the documentation, etc).

In this particular case my feeling is that the costs should not be very high. It seems to me the kernel almost implements it already. But I haven't looked the patch, so we'll need to confirm that. As for the documentation, I hope this PR does a good job at explaining not only that we support mutual records, but also shows some examples of the translation from positive coinductive types.

maximedenes:
> There is no user-facing support yet, this will come with a subsequent PR.

Oh, and I see this only now. I'm not sure I understand why the two are split. Is this PR introducing dead code?

ppedrot:
> Oh, and I see this only now. I'm not sure I understand why the two are split. Is this PR introducing dead code?

No, it's just making the data structure general enough. As such it reduces the probability of bugs because it makes the invariants more explicit.

Also, the other PR is ready but I'm waiting for this one to be merged before publishing it.

maximedenes:
> Also, the other PR is ready but I'm waiting for this one to be merged before publishing it.

I don't see why. IMHO that PR should be part of https://github.com/coq/coq/projects/25

ppedrot:
@maximedenes I know you like to have a global view, and this is why the goal of the project is textually described in there, split in various posts. But really, I forced myself to publish dependent PRs and the least I can say is that I don't like it, both for technical and psychological reasons. On the technical level, @herbelin started commenting on #7751 while all the material discussed actually pertained to the current PR, all because the commits are duplicated here and there. On the psychological level, dependent PRs add a number of PRs whose merge is not going to be immediate, as we need to review them separately and then discuss them alike. This results in a lot of open PRs for no good reasons. I'd really prefer pushing the branches on my repo, describing them in the project and then opening PRs one-by-one.

herbelin:
>  @herbelin started commenting on #7751

This was probably a mistake actually. I suspect I saw first #7750 and then for some reason I had a window open on #7751, was surprised that it was about the vm, still saw the changes in `declarations.mli`, thought I made a mistake when thinking I had seen a PR not about the vm, but posted the comment there.

I'd be happy to discuss anywhere centralized. It is indeed a pity to have so many discussions dispatched. Maybe we would also need some wiki-style or text-file place to maintain a summary of factual knowledge cleaned from the obsolete intermediate steps of the discussions.

ppedrot:
As mentioned on gitter, this PR inadvertently fixes a bug that was inadvertently introduced in #7679. Luckily it's not a soundness bug but it prevents native compilation of most coinductive data types.

The index used for the scheme of the name of projections is wrong. On a current master, it generates `proj_n_n_foobar` where `n` is the projection index, where it should instead read `proj_i_n_foobar` where `i` is the block index and `n` the projection index. This is not serious because the kernel does not accept mutual records currently, so in practice it just fails, but luckily the current PR that allows this also fixes the bug.

maximedenes:
@ppedrot Could you add a test case for the bug that was "inadvertently" fixed? Thanks!

ppedrot:
@maximedenes How does one write a test for the native compiler?

maximedenes:
You can use a cast with `<<:`. If the native compiler is disabled, it will fall back to the VM, which shouldn't be a problem.

ppedrot:
@maximedenes I mean, I want a test file to be compiled with the native compiler on, not merely testing convertibility. I am not sure it is the case by default in the test-suite, otherwise I would probably have remarked it there.

maximedenes:
If you use `<<:`, all dependencies needed to evaluate the conversion test will be compiled on the fly. They are not precompiled on the test-suite, and we could try to do that too, indeed (like for stdlib). But for the scope of this PR, a cast that will trigger compilation of relevant primitive projections is enough.

ppedrot:
@maximedenes Test added.

ppedrot:
Test still broken. §*#@ the day when it was decided that constructor indices start at 1.

ppedrot:
@maximedenes Correcting myself: this failure is not due to my patches. The source of the bug is already there in 8.8 and is essentially due to the fact that the `Nativecode.eq_gname` function always returns false on `Gproj` nodes. Probably a source of soundness failure. I'm currently trying to weaponize it.

ppedrot:
This really looked like it was either an eisenbug or something wrong on my side, as I did not manage to reproduce it after some cleanup. There is still a issue w.r.t. semantics of `eq_gname` though, even though I don't know if it's doable to exploit it.

maximedenes:
IIRC, those equality functions are used to define alpha-equivalence, used to factorize some code. With a function that returns `false` more often than it should, it's probably easy to exhibit an exponential behavior in the compiler. However, I'm not convinced it would be enough to break soundness. Of course, if the function returns `true` when it shouldn't, that's a different story...

maximedenes:
In any case, it is definitely worth fixing, so thanks for reporting.

ppedrot:
> IIRC, those equality functions are used to define alpha-equivalence

It's also used for the `HashtblGlobal` functor, which assumes the relation to be an equivalence one. I don't know what happens if you break this invariant.

Furthermore, the CI is still failing on that file while I cannot reproduce on my side, which is definitely strange.

ppedrot:
> Furthermore, the CI is still failing on that file while I cannot reproduce on my side, which is definitely strange.

It is a segfault, and it is only happening when using flambda. Do we actually test native compilation with this flag in general? I don't know where to check that in the CI configuration files.

Otherwise I'd blame some aggressive optimization that is performed by flambda that is invalidated by native compilation translation scheme, e.g. due to the fact that `Nativevalues.t` is exported as a function type. If this is the case it would require serious care to be sound w.r.t. OCaml semantics, or switching to a lower-level backend like Malfunction.

kernel/nativecode.ml
```diff
@@ -1985,12 +1986,14 @@ let compile_mind prefix ~interactive mb mind stack =
       let accu_br = MLapp (MLprimitive Mk_proj, [|get_proj_code i;accu|]) in
       let code = MLmatch(asw,MLlocal cf_uid,accu_br,[|[((ind,1),cargs)],MLlocal ci_uid|]) in
       let code = MLlet(cf_uid, MLapp (MLprimitive Force_cofix, [|MLlocal c_uid|]), code) in
-      let gn = Gproj ("", (pb.proj_ind, j), pb.proj_arg) in
+      let gn = Gproj ("", ind, pb.proj_arg) in
```
SkySkimmer: `j` is the projection number, so why is it getting changed to the inductive number?

ppedrot: That was the bug I mentioned above.

ppedrot:
@maximedenes OK, so the CI failure has nothing to do with my patches. Coq 8.8 + flambda already segfaults on my test. What do?

maximedenes:
Let's see if there's a quick fix to the flambda problem. Otherwise, we can merge this PR without the test and open a separate one with the test case and a "need: independent fix" label.

ppedrot:
I reverted the test for the time being, so that it doesn't delay merging of this PR.

kernel/entries.ml
```diff
@@ -49,7 +49,7 @@ type one_inductive_entry = {
   mind_entry_lc : constr list }
 
 type mutual_inductive_entry = {
-  mind_entry_record : (Id.t option) option; 
+  mind_entry_record : (Id.t array option) option;
   (** Some (Some id): primitive record with id the binder name of the record
```
maximedenes: The comment is outdated. And... `(Id.t array option) option`, seriously? The time has come for a dedicated type, I'm afraid.

maximedenes (dismissed):
Very welcome cleanup, just one nit.

kernel/entries.ml
ppedrot: I can change the type into another PR. It is more related to the upper layers, so it would make sense to do this in the vernac-facing PR.

SkySkimmer:
Shouldn't mind_record (and the mind_record_entry) move to the inductive packets instead of making it an array?
Must all types of a mutual block be of the same primitiveness?

ppedrot:
@SkySkimmer Currently, primitiveness is global to the inductive definition. It might be possible to also allow per-block distinction but:
- it is not clear whether it is OK to do so (e.g. regarding SR et al.)
- that would require changing a lot of code, including wreaking havoc in extraction

maximedenes:
@SkySkimmer are you satisfied with the answer? Can I merge this PR?

maximedenes (approved):

SkySkimmer:
>@SkySkimmer are you satisfied with the answer? Can I merge this PR?

Yes

Issue #7613, opened 2018-05-27
On the compatibility between match-based tuples and projection-based tuples


herbelin:
One of the concrete issues I encountered when trying to move tuples (`prod`, `conj`, `sig`, `sigT`, ...) to native projections is a difference of behavior of `intros` on a goal of the form `match t with (x,y) => P end`. With `match`-based tuples, introduction are blocked. With "projection"-based tuples, this is syntactic sugar for `let x := fst t in let y := snd t in P` and `intros` happily introduces `x` and `y` in the context.

What to think about that?

I thought for some times that `match`-based tuples or even ∑-types and `projection`-based tuples or ∑-types where fundamentally different. Indeed, they are in the context of linear logic, and this difference is essential to understand the "big picture" (thanks in passing to @ppedrot for having clarify this to me a few months ago).

I now realized that even if fundamentally different (different β-normal forms in particular), they are actually strictly isomorphic in intuitionistic logic as soon as they are considered with their definitional η-rules, i.e., respectively, with:

```
Γ ⊢ t : a:A ⊗ B    Γ, x : a:A ⊗ B ⊢ E[x] : P(x)
——————————————————————————————————————————————————
Γ ⊢ E[t] ≡ match t with (x,y) => E(x,y) end : P(t)

Γ ⊢ t : a:A & B
————————————————————————————————
Γ ⊢ t ≡ (fst t, snd t) : a:A & B
```
(I'm using Agda-style notations of a ∑-type `∑a:A.B` so as to be able to distinguish between the match-based version `a:A ⊗ B` and the projection-based `a:A & B` incarnations.)

So, again, what to think about that?

- Should we go towards removing any observational difference between the two isomorphic `match`-based and projection-based tuples and ∑-types?
  - In particular, should we consider that `intros` should not fail on `match t with (x,y) => P(x,y) end` (even in the `match`-based case) and instead apply η on the fly to rephrase it into `let x := match t with (x,y) => x end in let y := match t with (x,y) => y end in P(x,y)` (`match`-based do not currently have η implemented in the conversion but we could think at providing it at least for this purpose)?
  - One may otherwise prefer instead that a goal `match z with (x,y) => P(x,y) end` reduces to `forall x y, P(x,y)`, but it is not clear how to recognize that we want this behavior when seeing `let x := fst z in let y := snd z in P(x,y)` in the projection-based case.
[...]

PR #6754, opened 2018-02-13
Better elaboration of pattern-matchings on primitive projections

mattam82:
Better representation of cases on primitive projections. Work in progress.
Now using let-ins consistently for the thing to destruct and each projection
to avoid blowups. Currently based on 8.7 for testing on current developments.

**Kind:** feature / performance.

Fixes / closes #5701

- [x] Corresponding documentation was added / updated.
- [x] Entry added in CHANGES.

PR #7751, opened 2018-06-08
Adapt the VM to use index-based projections.

ppedrot:
This depend on #7750 [merged].

Part of the project to switch to an index-based representation of primitive projections.

### DIFF
--- a/dev/vm_printers.ml
+++ b/dev/vm_printers.ml
@@ -16,8 +16,8 @@ let ppripos (ri,pos) =
       print_string "structured constant\n"
   | Reloc_getglobal kn ->
     print_string ("getglob "^(Constant.to_string kn)^"\n")
-  | Reloc_proj_name p ->
-    print_string ("proj "^(Constant.to_string p)^"\n")
+  | Reloc_proj_name ((mind, n), i) ->
+    print_string ("proj "^(MutInd.to_string mind)^","^(string_of_int n)^","^(string_of_int i)^"\n")
   );
    print_flush ()
 
--- a/kernel/cbytecodes.ml
+++ b/kernel/cbytecodes.ml
@@ -128,7 +128,7 @@ type instruction =
   | Ksetfield of int
   | Kstop
   | Ksequence of bytecodes * bytecodes
-  | Kproj of int * Constant.t  (* index of the projected argument,
+  | Kproj of (inductive * int)  (* index of the projected argument,
                                            name of projection *)
   | Kensurestackcapacity of int
 (* spiwack: instructions concerning integers *)
@@ -239,6 +239,9 @@ let pp_sort s =
   | Prop Pos -> str "Set"
   | Type u -> str "Type@{" ++ Univ.pr_uni u ++ str "}"
 
+let pp_proj ((mind, n), i) =
+  MutInd.print mind ++ str"#" ++ int n ++ str "." ++ int i
+
 let rec pp_struct_const = function
   | Const_sort s -> pp_sort s
   | Const_ind (mind, i) -> MutInd.print mind ++ str"#" ++ int i
@@ -310,7 +313,7 @@ let rec pp_instr i =
 
   | Kbranch lbl -> str "branch " ++ pp_lbl lbl
 
-  | Kproj(n,p) -> str "proj " ++ int n ++ str " " ++ Constant.print p
+  | Kproj p -> str "proj " ++ pp_proj p
 
   | Kensurestackcapacity size -> str "growstack " ++ int size
 
--- a/kernel/cbytecodes.mli
+++ b/kernel/cbytecodes.mli
@@ -88,8 +88,7 @@ type instruction =
   | Ksetfield of int                    (** accu[n] = sp[0] ; sp = pop sp *)
   | Kstop
   | Ksequence of bytecodes * bytecodes
-  | Kproj of int * Constant.t  (** index of the projected argument,
-                                            name of projection *)
+  | Kproj of (inductive * int)  (** inductive, index of the projected argument *)
   | Kensurestackcapacity of int
 
 (** spiwack: instructions concerning integers *)
--- a/kernel/cbytegen.ml
+++ b/kernel/cbytegen.ml
@@ -479,8 +479,8 @@ let rec compile_lam env cenv lam sz cont =
 
   | Lval v -> compile_structured_constant cenv v sz cont
 
-  | Lproj (n,kn,arg) ->
-     compile_lam env cenv arg sz (Kproj (n,kn) :: cont)
+  | Lproj (p, arg) ->
+     compile_lam env cenv arg sz (Kproj p :: cont)
 
   | Lvar id -> pos_named id cenv :: cont
 
--- a/kernel/cemitcodes.ml
+++ b/kernel/cemitcodes.ml
@@ -27,7 +27,7 @@ type reloc_info =
   | Reloc_annot of annot_switch
   | Reloc_const of structured_constant
   | Reloc_getglobal of Names.Constant.t
-  | Reloc_proj_name of Constant.t
+  | Reloc_proj_name of (inductive * int)
 
 let eq_reloc_info r1 r2 = match r1, r2 with
 | Reloc_annot sw1, Reloc_annot sw2 -> eq_annot_switch sw1 sw2
@@ -36,7 +36,7 @@ let eq_reloc_info r1 r2 = match r1, r2 with
 | Reloc_const _, _ -> false
 | Reloc_getglobal c1, Reloc_getglobal c2 -> Constant.equal c1 c2
 | Reloc_getglobal _, _ -> false
-| Reloc_proj_name p1, Reloc_proj_name p2 -> Constant.equal p1 p2
+| Reloc_proj_name (i1, n1), Reloc_proj_name (i2, n2) -> eq_ind i1 i2 && Int.equal n1 n2
 | Reloc_proj_name _, _ -> false
 
 let hash_reloc_info r =
@@ -45,7 +45,7 @@ let hash_reloc_info r =
   | Reloc_annot sw -> combinesmall 1 (hash_annot_switch sw)
   | Reloc_const c -> combinesmall 2 (hash_structured_constant c)
   | Reloc_getglobal c -> combinesmall 3 (Constant.hash c)
-  | Reloc_proj_name p -> combinesmall 4 (Constant.hash p)
+  | Reloc_proj_name (ind, n) -> combinesmall 4 (combine (ind_hash ind) n)
 
 module RelocTable = Hashtbl.Make(struct
   type t = reloc_info
@@ -284,7 +284,7 @@ let emit_instr env = function
       if n <= 1 then out env (opSETFIELD0+n)
       else (out env opSETFIELD;out_int env n)
   | Ksequence _ -> invalid_arg "Cemitcodes.emit_instr"
-  | Kproj (n,p) -> out env opPROJ; out_int env n; slot_for_proj_name env p
+  | Kproj (_, n as p) -> out env opPROJ; out_int env n; slot_for_proj_name env p
   | Kensurestackcapacity size -> out env opENSURESTACKCAPACITY; out_int env size
   (* spiwack *)
   | Kbranch lbl -> out env opBRANCH; out_label env lbl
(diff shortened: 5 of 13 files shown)


herbelin:
Hi @ppedrot, thanks for this PR!

I see that you added some documentation in `declarations.mli` and it seems that part of the terminology is clear. For instance, it seems undisputable that the terminology "record" at least means that the type is a (possibly dependent) conjunctive type with named projections defined at declaration time.

However, I feel that the terminology is otherwise often confused. Typically, should the terminology "record" be associated to other meanings as well? Let me give a list of questions:

1. does the terminology "record" mean that the type is usable with `{| |}` syntax?
   1. if yes, does it mean usable exclusively with only the `{| |}` syntax or also with a `Build_record`-style syntax?
1. does "record" tell something about the compactness of the representation of projections?
   1. in particular, does the terminology "primitive record" include projections being represented in a compact way?
   1. does the terminology "fake record" include projections not being represented in a compact way?
1. does "record" tell something about whether projections are normal expressions or expanded into `match` e.g. by tactics `cbn`, `cbv`, ...?
    1. does the terminology "primitive record" include or not projections being normal expressions?
    1. and does the terminology "fake record" include or not projections being expanded into `match`?
1. does "record" tell something about whether the components of a `record` are evaluated strictly or not (i.e. whether `(t,u).(fst)` forces the evaluation of `t` and `u` or not) by `cbn`, `cbv`, ...?
1. does "record" tell something about whether a `match` syntax is provided for the type and, if yes, whether a `match` is a normal expression or is expanded into "projections"?
   1. does the terminology "primitive record" include or not `match` being provided and if yes being normal or not?
   1. I imagine that the terminology "fake record" refers to pre-8.5 "records", and consequently that it includes `match` syntax which is not normalized to projections; is that correct?
1. does "record" exclude or not being recursive (a μ-fixpoint on types) or corecursive (a ν-fixpoint)?

In the same vein, should the terminology "primitive projections" be used in the sense of compact representation, what may suggest the word "primitive" taken from the point of view of a dedicated efficient implementation?

Or should it be used in the sense of projections being normal expressions, what may suggest the word "primitive" taken from the logical point of view (i.e. in the sense of defined from projections)?

That's a lot of questions and I hope you're ok that I'm asking them. I'm sure it shall be useful to everyone to see answers given in `declarations.mli`.

ppedrot:
@herbelin I think you are mixing several loosely related concepts in this comment. Furthermore, as a generic remark, this PR is only concerned about the VM, you're probably referring to the PRs that it depends on.

> does the terminology "record" mean that the type is usable with {| |} syntax?

This is a syntactic problem unrelated to the kernel. There is already several similar notions that only exist at the syntactic level and are desugared or interpreted before reaching the kernel, e.g. `let`-style matching and the like.

> does "record" tell something about the compactness of the representation of projections?

I guess that by that you mean that parameters are not present in the projection node. For now there is no non-compact primitive projection system in the kernel, and I doubt of the usefulness of the existence of such a concept. Dually, you could also ask for pattern-matching with repeated parameters in the matching branch, and thanks to our users' sanity, this has never been asked.

Points 3., 4. and 5. are unrelated to the kernel. I'll just comment briefly on 4. by stating that I believe that the reduction semantics of primitive projections in the kernel is probably somewhat broken. In denotational models you need negative pairs to be lazy, while we evaluate them eagerly in CIC. It works because we have no effects around but I'd bet it would turn problematic if this were the case.

> does "record" exclude or not being recursive (a μ-fixpoint on types) or corecursive (a ν-fixpoint)?

That one is tricky. Currently the kernel partially supports recursive records (with a few bugs I think) and co-recursive records. I don't think we have a clear story for the full picture yet, but we know that some choices work and some other do not.

herbelin:
@ppedrot: Thanks for the quick answer.

> this PR is only concerned about the VM, you're probably referring to the PRs that it depends on.

Yes, it is maybe not the best place for addressing all these questions, but since you introduced terminology like `PrimitiveRecord` and `FakeRecord`, I thought it was important to understand what they mean as soon as now.

> I think you are mixing several loosely related concepts in this comment.

Asking for a clarification of what is related to records and what is not is precisely the point of my list of questions. For instance, we have #6651 which is precisely waiting for an answer to the question of what are the cases for which it is meaningful to use the dot notation.

> I guess that by that you mean that parameters are not present in the projection node.

Yes.

> For now there is no non-compact primitive projection system in the kernel, and I doubt of the usefulness of the existence of such a concept.

Sure. But the question is then whether primitive projection means the same as primitive record or not. For instance, you introduced the terminology `FakeRecord` which corresponds, if I'm not mistaken, to current `Record` keyword and to current `Inductive I := { ... }`, when `Primitive` is not set. The question is: what is the definition of `FakeRecord`? For instance, does it necessarily imply no compact representation of projections and if yes why? In other words, does the terminology "primitive" only refers to the existence of a proper ML constructor for projections in the syntax of terms independently of whether a `match` syntax in the kernel is also available or not for this type or does it strictly refer to a "negative" semantics of the conjunction for which `match` is not available in the kernel.

Based on the first experiments I made, the question seems important because if we want to take benefit from a compact internal representation of projections for all conjunctive types defined in the library, incompatibilities are waiting for us if at the same time we stop making `match` natively available.

The question seems also important from a computational point of view. There might be situations where having a single `match t with` in which `t` is evaluated once and for all is preferable to having a duplication of `fst t` and `snd t` expressions which would reevaluate `t` several times. Told otherwise, it is unclear to me that there are no situations where both primitive `match` and primitive projections are worth to be available.

> Dually, you could also ask for pattern-matching with repeated parameters in the matching branch, and thanks to our users' sanity, this has never been asked.

Sorry, I don't see what you mean. What would not be sane for users?

> I believe that the reduction semantics of primitive projections in the kernel is probably somewhat broken. In denotational models you need negative pairs to be lazy, while we evaluate them eagerly in CIC. It works because we have no effects around but I'd bet it would turn problematic if this were the case.

I was actually wondering about effects. For instance, you probably already investigated what would break in the isomorphism between positive tuples and negative tuples in the presence of effects?

Thanks again for your answers.

ppedrot:
> But the question is then whether primitive projection means the same as primitive record or not.

I don't think that there is a computational way to implement primitive projections (i.e. as indices) on non-primitive records. You'll get into the horrible hell of η-expansions of sum types which is IIRC undecidable in general and ill-typed in dependent types in particular. It is a pity that the primitive record feature was named primitive *projections*, because it is not quite about projections.

> The question is: what is the definition of FakeRecord?

This flag is not used by the kernel. It is only a marker for extraction so as to generate ML records. As far as the kernel is concerned, `FakeRecord` and `NotRecord` represent the same reality.

> incompatibilities are waiting for us if at the same time we stop making match natively available

Incompatibilities are like everywhere with primitive projections set by default, not only regarding matches. There are issues with name generation, computational behaviour, term matching, etc. In retrospective I am deeply convinced that, albeit attractive, it was a *very bad idea* to try to provide the user with a partial compatibility layer. In practice it doesn't work, as our dear primitive projections user can testify (@RalfJung, @robbertkrebbers...) and furthermore it really makes the kernel implementation more complex and less trustable.

> The question seems also important from a computational point of view. There might be situations where having a single match t with in which t is evaluated once and for all is preferable to having a duplication of fst t and snd t expressions which would reevaluate t several times. Told otherwise, it is unclear to me that there are no situations where both primitive match and primitive projections are worth to be available.

That's why we have #6754 and primitive let-bindings in CIC.

> Sorry, I don't see what you mean. What would not be sane for users?

The fact that primitive projections can be formulated without parameters is a lucky optimization, but is not  the core of the concept. Up to duality, this is the same as forcing the branches of standard pattern-matching over inductive types in CIC to contain the parameters of the inductive types.

Both are equivalent from the equational point of view, but it is not from a complexity one as this forces the kernel to duplicate information that can be re-inferred. So my point was that it's stupid to include data that is not necessary, it can only make things slower.

> I was actually wondering about effects. For instance, you probably already investigated what would break in the isomorphism between positive tuples and negative tuples in the presence of effects?

That's pretty well-known, this is essentially the difference between `A ⊗ B` and `A & B` in linear logic, which are *not* isomorphic. Indeed, `[A ⊗ B] := T ([A] × [B])` and `[A & B] := [A] × [B]`.

herbelin:
> I don't think that there is a computational way to implement primitive projections (i.e. as indices) on non-primitive records.

I don't understand what you call indices here? Is it In the sense of "realargs"? If yes, I don't see what you mean.

Anyway, @mattam82 made a compatibility level which can represent `match t with (x,y) => x end` with a primitive projection. Don't you like it?

> You'll get into the horrible hell of η-expansions of sum types ...

But positive conjunctions are simpler than sum types. As discussed in #7613, η-expansions for 
positive conjunctions seem direct (already because we can mimic the way it is done for the negative representation).

> which is IIRC undecidable in general

The undecidability is about recursive types. η-expansion on `nat` or on negative streams is undecidable but it is not related to being positive or negative, or to being a conjunction or a disjunction. It is related to being infinite, isn't it?

> and ill-typed in dependent types in particular.

I have some ideas on this question... From a "moral" point of view, there is no reason that commutative cuts and η-expansions are ill-typed. The solution I think is to embed `generalization` in the syntax (as Pierre B. was doing for small inversion). Embedding generalization in the syntax is what e.g. Lasson is doing to have a "pure" version of `J`/`eq_rect` in his paper on omega-groupoids laws, where "pure", here, means that his language, which is talking only about equality, does not mention other logical concepts than equality. 

Otherwise said, the current usage is to simulate generalizations with β-expansions across a `match` but I'm sure that the concept makes sense in itself as part of the notion of "dependent pattern-matching" without having to require explicitly dependent products. By embedding generalization in the syntax, it is not anymore an artificial β-"computation". It becomes what we need in the exact form we need.

> It is a pity that the primitive record feature was named primitive *projections*, because it is not quite about projections.

On the other side, if they were called primitive projections, it is probably also because under some other points of view, it is really the projection which is primitive and not necessarily the type.

> Incompatibilities are like everywhere with primitive projections set by default ... it was a *very bad idea* to try to provide the user with a partial compatibility layer. In practice it doesn't work, as our dear primitive projections user can testify (@RalfJung, @robbertkrebbers...)

This is exactly my point. The compatibility level was on top of a primitive-record-based implementation of primitive projections. I think it is worth to look at a positive-record-based implementation of primitive projections in parallel, or maybe even in superimposition, to a negative-based implementation of primitive projections.

Why to suffer incompatibilities if we can have the advantages of positive-based and negative-based records at the same time?

> The fact that primitive projections can be formulated without parameters is a lucky optimization, but is not  the core of the concept. Up to duality, this is the same as forcing the branches of standard pattern-matching over inductive types in CIC to contain the parameters of the inductive types.

I'm lost, but we would better make progress with a live discussion (why a "projection"/`match` duality rather than a "projection"/"constructors", or a "tuple"/`match` duality?)

> That's pretty well-known, this is essentially the difference between `A ⊗ B` and `A & B` in linear logic, which are *not* isomorphic. Indeed, `[A ⊗ B] := T ([A] × [B])` and `[A & B] := [A] × [B]`.

There is still something that I don't understand. Are you attaching a `T` to any connective or you are only inserting `T` at the border between the positive and negative phases? In the second case, why would individual connectives lost their η-rules in the process? Isn't it only `T` which breaks η and not the underlying ⊗ or & structure? And, very informally, can't we keep the intuitionistic structure at the same time as the `T` structure, so that `!A ⊗ !B` and `!(A & B)` remains isomorphic, even in the presence of an additional `T` structure.

I don't know when we shall have the opportunity to meet. I have a lot to discuss with you.

maximedenes:
This discussion is very long. @ppedrot is any of it relevant for this PR? Can you summarize what its status is?

ppedrot:
@maximedenes The discussion is unrelated to this PR, actually. It should have taken place in #7750.

herbelin (commented):
Relying on the experts. OK to me.

maximedenes (changes requested):
Looks good, modulo a comment. Do you know if we test VM + primitive projections in the test suite? I would like to see at least a test case for a record with two projections.

maximedenes:
@ppedrot ping

ppedrot:
@SkySkimmer This PR is morally superseded by #7908, but the latter uses `Projection.Repr.t` instead of just a pair of an inductive and an integer. While it is better to use dedicated types to carry semantics, `Projection.Repr.t` also contains irrelevant information like the number of parameters and the label of the projection. So I don't know whether it is better to merge this PR and rebase #7908 atop of it, or simply drop this one. What do you think, @SkySkimmer?

SkySkimmer:
I think carrying irrelevant information is fine. It's not like we reallocate the record in the VM. So drop this one.

ppedrot:
All right, lets close.

PR #7759, opened 2018-06-10
Workaround to fix #7731 (printing not splitting line at break hint).

### LINKED ISSUE #7731: Printing bug: Newline added in incorrect place
#### Version

8.8.0

Consider the following file:
```coq
From Coq Require Import Utf8.
Section test.
Context (atomic_update : forall {TX TY : Type} (Eo Em : nat) (α : TX -> Prop) (β Φ : TX -> TY -> Prop), Prop).
Context (val loc : Type) (mapsto : loc -> val -> Prop) (XX: nat).

Notation "'AU' '{{' ∀ x : Tx , α } } @ Eo , Em '{{' ∃ y : Ty , β , 'RET' Φ } }" :=
  (atomic_update Tx Ty Eo Em (fun x => α) (fun x y => β) (fun x y => Φ))
  (at level 20, Eo, Em, α, β, Φ at level 200, x, y at level 99, format
"'[   ' 'AU'  '{{'  ∀  x  :  Tx ,  α  } }  '/' @  Eo ,  Em  '/' '[   ' '{{'  ∃  y  :  Ty ,  β ,  '/' RET  Φ  } } ']' ']'").

Notation "l  ↦  v" := (mapsto l v) (at level 20).

Definition PRE (P : Prop) := P.
Definition ENT (P : list Prop) (Q : Prop) := Q.
Definition AND (P Q : Prop) := P /\ Q.

Notation "P ∗ Q" := (AND P Q) (at level 80, right associativity).
Notation "⊤" := XX.

Notation "Γ  '_' : P" := (cons (PRE P) Γ)
  (at level 1, P at level 200,
   left associativity, format "Γ '_'  :  '[' P ']' '//'", only printing).
Notation "" := nil (only printing).

Notation "Δ '--------------------------------------' ∗ Q" :=
  (ENT Δ Q)
  (at level 1, Q at level 200, left associativity,
  format "Δ '--------------------------------------' ∗ '//' Q '//'", only printing).


Lemma test (Φ : nat -> Prop) l x :
  ENT (cons (PRE( AU {{ ∀ xx : val, l ↦ xx ∗ l ↦ xx ∗ l ↦ xx }} @ ⊤, ⊤
            {{ ∃ yyyy : val, l ↦ yyyy ∗ l ↦ x ∗ l ↦ x ∗ l ↦ x ∗ l ↦ x ∗ l ↦ x,
               RET Φ 0 }})) nil)
  True.
Proof. 
Show.
Abort.
End test.
```
and run it using `coqc filename.v`. That shows the following output:
```
1 subgoal
  
  atomic_update : ∀ TX TY : Type, nat
                                  → nat
                                    → (TX → Prop)
                                      → (TX → TY → Prop)
                                        → (TX → TY → Prop) → Prop
  val : Type
  loc : Type
  mapsto : loc → val → Prop
  XX : nat
  Φ : nat → Prop
  l : loc
  x : val
  ============================
  _ : AU {{ ∀ xx : val, l ↦ xx ∗ l ↦ xx ∗ l ↦ xx }} @ 
         ⊤, ⊤
         {{ ∃ yyyy : val, l ↦ yyyy ∗ l ↦ x ∗ l ↦ x ∗ l ↦ x ∗ l ↦ x ∗ l ↦ x,
            RET Φ 0 }}
  --------------------------------------∗
  True
```
Notice how, in the line starting `_ : AU`, there is a linebreak after the `@` even though the format for that notation says to break *before* the `@`! So, I think there's a bug here, and I'd expect the linebreak to occur either before the `@` or after the second `⊤`.

Note that this example is extremely fragile. Changing `Γ  '_' : P` to remove the `Γ` (because we only have a singleton context here) makes the bug disappear. Using `0` instead of `⊤` makes the bug disappear. Also I was not able to reproduce this bug in a way that it shows in emacs, but only when using `Show.`. I suppose getting it in emacs is possible though, I just have not found the right term.

herbelin:
**Kind:** bug fix / workaround

Closes #7731.

In some cases, Format's inner boxes inside an outer box act as break hints, even though there are already "better" break hints in the outer box.

We work around this "feature" by not inserting a box around the default printing rule for a notation if there is no effective break points in the box. Hoping that it has no unseen unexpected side effects besides fixing the issue.

### DIFF
--- a/test-suite/output/Notations3.out
+++ b/test-suite/output/Notations3.out
@@ -241,3 +241,8 @@ Notation
 Notation
 "( x , y , .. , z )" := pair .. (pair x y) .. z : core_scope
 (default interpretation)
+1 subgoal
+  
+  ============================
+  ##@%
+   ^^^
--- a/test-suite/output/Notations3.v
+++ b/test-suite/output/Notations3.v
@@ -385,3 +385,17 @@ Module LocateNotations.
 Locate "exists".
 Locate "( _ , _ , .. , _ )".
 End LocateNotations.
+
+Module Issue7731.
+
+Axiom (P : nat -> Prop).
+Parameter (X : nat).
+Notation "## @ E ^^^" := (P E) (at level 20, E at level 1, format "'[ ' ## '/' @ E '/' ^^^ ']'").
+Notation "%" := X.
+
+Set Printing Width 7.
+Goal ## @ % ^^^.
+Show.
+Abort.
+
+End Issue7731.
--- a/vernac/metasyntax.ml
+++ b/vernac/metasyntax.ml
@@ -1312,8 +1312,15 @@ let make_pa_rule level (typs,symbols) ntn need_squash =
 
 let make_pp_rule level (typs,symbols) fmt =
   match fmt with
-  | None -> [UnpBox (PpHOVB 0, make_hunks typs symbols level)]
-  | Some fmt -> hunks_of_format (level, List.split typs) (symbols, parse_format fmt)
+  | None ->
+     let hunks = make_hunks typs symbols level in
+     if List.exists (function _,(UnpCut (PpBrk _) | UnpListMetaVar _) -> true | _ -> false) hunks then
+       [UnpBox (PpHOVB 0,hunks)]
+     else
+       (* Optimization to work around what seems an ocaml Format bug (see Mantis #7804/#7807) *)
+       List.map snd hunks (* drop locations which are dummy *)
+  | Some fmt ->
+     hunks_of_format (level, List.split typs) (symbols, parse_format fmt)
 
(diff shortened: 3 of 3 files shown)


ejgallego:
I guess no reason to delay this, @herbelin , @mattam82 , are you ok with me helping here? If so, please rebase and assign to me.

herbelin:
Rebased and assigned to @ejgallego.

ejgallego (approved):
Wow, what a bug! Congrats @herbelin and this was a tough one.

I guess the workaround is ugly but we have no other choice. Maybe you could add a comment to the code in metasyntax pointing to the OCaml bug report?

ejgallego:
LGTM, merging as soon as CI passes.

ejgallego:
The CI seems in sorry state but not the fault of this PR.

PR #7761, opened 2018-06-10
Fixing #7700: section variables bound to abbreviations were not found

### LINKED ISSUE #7700: Notation aliases for Lets are broken
#### Version

8.8.0

This should work
```coq
Section foo.
  Let x := Set.
  Notation y := x.
  Check y. (* Error: The reference y was not found in the current environment. *)
```

herbelin:
**Kind:** bug fix

Fixes #7700

Surprisingly enough this is a serious missing case of the globalization of variables which has remained non reported for more than a decade.

Redundancy between finding section variables in both `interp_var` and `interp_qualid` could probably be cleaned, if some courageous soul is wanting to investigate that.

### DIFF
--- a/interp/constrintern.ml
+++ b/interp/constrintern.ml
@@ -994,9 +994,12 @@ let glob_sort_of_level (level: glob_level) : glob_sort =
   | GType info -> GType [sort_info_of_level_info info]
 
 (* Is it a global reference or a syntactic definition? *)
-let intern_qualid qid intern env ntnvars us args =
+let intern_qualid ?(no_secvar=false) qid intern env ntnvars us args =
   let loc = qid.loc in
   match intern_extended_global_of_qualid qid with
+  | TrueGlobal (VarRef _) when no_secvar ->
+      (* Rule out section vars since these should have been found by intern_var *)
+      raise Not_found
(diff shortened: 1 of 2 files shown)


interp/constrintern.ml
```diff
@@ -994,9 +994,12 @@ let glob_sort_of_level (level: glob_level) : glob_sort =
   | GType info -> GType [sort_info_of_level_info info]
 
 (* Is it a global reference or a syntactic definition? *)
-let intern_qualid qid intern env ntnvars us args =
+let intern_qualid ?(no_secvar=false) qid intern env ntnvars us args =
```
ejgallego: Maybe the parameter should be named `avoid_secvar`?

herbelin: If you want.

ejgallego: Not a strong feeling towards it, I was just thinking out loud and suggesting an alternative; so if you like the new one better update the branch, otherwise I'll merge tomorrow.

[Yes, I do agree that part could use a cleanup, also abbreviations could use some love]

ejgallego (approved):
This part of the code is indeed sensible but as far as I can tell the patch looks OK. Thanks!

PR #7762, opened 2018-06-10
Closes #7699: confusing undocumented multi-class syntax of Bind Scope

### LINKED ISSUE #7699: Bind Scope does not correctly handle types which are not identifiers
#### Version

8.8.0

This code does not behave as I expect it to:

```coq
Axiom foo : option unit -> Type.
Axiom bar : option nat -> Type.
Delimit Scope a_scope with a.
Delimit Scope b_scope with b.
Bind Scope a_scope with option unit.
Bind Scope b_scope with option nat.
Notation "()" := (Some tt) : a_scope.
Notation "()" := (Some O) : b_scope.
Set Printing All.
Check bar (). (* bar (@Some nat O) : Type *)
Check foo ().
(*Error:
The term "@Some nat O" has type "option nat"
while it is expected to have type "option unit".
*)
```

cc @davidben, from whom I discovered that `Bind Scope a_scope with option unit` is not an error.

cc @herbelin

herbelin:
**Kind:** syntax clarification 

Fixes #7699.

Only `Bind Scope a_scope with class` is documented. Do not provide `Bind Scope a_scope with class1 .. classn` as it is a source of confusion. Think e.g. as `Bind Scope b_scope with option nat`.

Note that the standard library was using the feature though in the form of `Bind Scope nat_scope with Nat.t nat`. Note also that some users do it (e.g. @jasongross in issue #3732). So, even if not documented, it might be worth a message in `CHANGES` if the decision is to restrict the syntax.

If ever we really want the syntax, maybe better to provide `Bind Scope a_scope with class1, ..., class2`? That would be less confusing, and that would let open a syntax `Bind Scope a_scope with class_pattern` with `class_pattern` being e.g. `option _`.

### DIFF
--- a/vernac/g_vernac.mlg
+++ b/vernac/g_vernac.mlg
@@ -490,6 +490,10 @@ let warn_deprecated_include_type =
   CWarnings.create ~name:"deprecated-include-type" ~category:"deprecated"
          (fun () -> strbrk "Include Type is deprecated; use Include instead")
 
+let warn_deprecated_bind_several_scopes =
+  CWarnings.create ~name:"deprecated-bind-multiple-scopes" ~category:"deprecated"
+         (fun () -> strbrk "Binding a scope to several types at the same time is deprecated.")
+
 }
 
 (* Modules and Sections *)
@@ -1119,7 +1123,9 @@ GRAMMAR EXTEND Gram
 	 { VernacDelimiters (sc, None) }
 
      | IDENT "Bind"; IDENT "Scope"; sc = IDENT; "with";
-       refl = LIST1 class_rawexpr -> { VernacBindScope (sc,refl) }
+       refl = LIST1 class_rawexpr ->
+         { if List.length refl > 1 then warn_deprecated_bind_several_scopes ();
+           VernacBindScope (sc,refl) }
 
      | IDENT "Infix"; op = ne_lstring; ":="; p = constr;
          modl = [ "("; l = LIST1 syntax_modifier SEP ","; ")" -> { l } | -> { [] } ];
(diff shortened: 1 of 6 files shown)


SkySkimmer:
Should we have a deprecation step?

herbelin:
> Should we have a deprecation step?

No idea of whether it is worth since it is undocumented (I don't remember the jurisprudence). I added one though.

Pending on a [PR](https://github.com/coq/bignums/pull/12) to bignums.

herbelin:
@zimmi48: since this PR is about synchronizing the code and documentation (deciding somehow a bit arbitrarily to change the code to adjust to the documentation), would you be ok to give your opinion and eventually be the assignee?

Zimmi48:
Yes.

vernac/g_vernac.mlg
```diff
@@ -490,6 +490,10 @@ let warn_deprecated_include_type =
   CWarnings.create ~name:"deprecated-include-type" ~category:"deprecated"
          (fun () -> strbrk "Include Type is deprecated; use Include instead")
 
+let warn_deprecated_bind_several_scopes =
+  CWarnings.create ~name:"deprecated-bind-multiple-scopes" ~category:"deprecated"
+         (fun () -> strbrk "Binding of several scopes at the same time is deprecated. ")
```
Zimmi48: This message is a bit confusing. I thought the scope was e.g. `Z_scope`. If so, then the command always take one single scope but binds several items to this scope at once.

Zimmi48 (approved):
Looks good to go in 8.9. Please open another PR targeting 8.10 removing the syntax and including an entry in CHANGES.

vernac/g_vernac.mlg
herbelin: I tried another formulation: "Binding a scope to several types at the same time is deprecated".

herbelin:
> Please open another PR targeting 8.10 removing the syntax and including an entry in CHANGES.

Done, #8558.

vernac/g_vernac.mlg
Zimmi48: Sounds better.

silene:
Given the discussion on #8558, it seems like a documentation fix would be better.

PR #7768, opened 2018-06-10
Fix #7723 (vm_compute segfault and proof of false)

### LINKED ISSUE #7723: vm_compute segfaults with universe polymorphism
#### Version

8.8.0

```coq
Set Universe Polymorphism.

Inductive decision_tree : Type := .

Fixpoint first_satisfying_helper {A B} (f : A -> option B) (ls : list A) : option B
  := match ls with
     | nil => None
     | cons x xs
       => match f x with
          | Some v => Some v
          | None => first_satisfying_helper f xs
          end
     end.

Axiom admit : forall {T}, T.
Definition dtree4 : option decision_tree :=
  match first_satisfying_helper (fun pat : nat => Some pat) (cons 0 nil)
  with
  | Some _ => admit
  | None => admit
  end
.
Definition dtree'' := Eval vm_compute in dtree4. (* segfault *)
```

More involved example, with other relevant error messages, coming up soon.

cc @maximedenes @SkySkimmer @mattam82

maximedenes:
I don't claim full understanding of this code, but the fix looks ok.

- [X] Entry added in CHANGES.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -88,6 +88,8 @@ Kernel
 
 - Fix a critical bug with cofixpoints and vm_compute/native_compute (#7333).
 - Fix a critical bug with inlining of polymorphic constants (#7615).
+- Fix a critical bug with universe polymorphism and vm_compute (#7723). Was
+  present since 8.5.
 
 Notations
 
--- a/kernel/cbytecodes.ml
+++ b/kernel/cbytecodes.ml
@@ -217,6 +217,7 @@ type vm_env = {
 
 
 type comp_env = {
+    arity : int;                 (* arity of the current function, 0 if none *)
     nb_uni_stack : int ;         (* number of universes on the stack,      *)
                                  (* universes are always at the bottom.    *)
     nb_stack : int;              (* number of variables on the stack       *)
--- a/kernel/cbytecodes.mli
+++ b/kernel/cbytecodes.mli
@@ -159,6 +159,7 @@ type vm_env = {
 
 
 type comp_env = {
+    arity : int;                 (* arity of the current function, 0 if none *)
     nb_uni_stack : int ;         (** number of universes on the stack       *)
     nb_stack : int;              (** number of variables on the stack       *)
     in_stack : int list;         (** position in the stack                  *)
--- a/kernel/cbytegen.ml
+++ b/kernel/cbytegen.ml
@@ -258,14 +258,12 @@ let pos_universe_var i r sz =
   (* Compilation of a universe variable can happen either at toplevel (the
   current closure correspond to a constant and has local universes) or in a
   local closure (which has no local universes). *)
-  if r.nb_uni_stack != 0 then begin
+  if r.nb_uni_stack != 0 then
     (* Universe variables are represented by De Bruijn levels (not indices),
     starting at 0. The shape of the stack will be [v1|..|vn|u1..up|arg1..argq]
     with size = n + p + q, and q = r.arity. So Kacc (sz - r.arity - 1) will access
     the last universe. *)
-    Printf.eprintf "pos_univ i=%i sz=%i arity=%i nb_univ=%i\n" i sz r.arity r.nb_uni_stack;
     Kacc (sz - r.arity - (r.nb_uni_stack - i))
-  end
   else
     let env = !(r.in_env) in
     let db = FVuniv_var i in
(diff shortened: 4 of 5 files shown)


Zimmi48:
Unrelated to the main topic here but may I ask why you pushed this branch to https://gitlab.com/coq/coq?

Zimmi48:
Also, not completely specific to this PR, but given the very high number of critical bug fixes that will be part of 8.8.1, maybe a general comment could be added to the CHANGES file (most of them seem to imply the universes and either a reduction machine or modules).

maximedenes:
> Unrelated to the main topic here but may I ask why you pushed this branch to https://gitlab.com/coq/coq?

Sorry, I meant to push to my gitlab fork instead.

maximedenes:
> maybe a general comment could be added to the CHANGES file

I think each critical bug should get a separate `CHANGES` entry. Do you mean another more general one on top of that? What would it say?

Zimmi48:
> Sorry, I meant to push to my gitlab fork instead.

You should set-up your GitLab fork as a CI/CD mirror, don't you think. (Cf. explanations in https://github.com/coq/coq/tree/master/dev/ci#information-for-developers).

> I think each critical bug should get a separate CHANGES entry. Do you mean another more general one on top of that? What would it say?

Yes, of course. I meant an introduction comment to reassure our users. For instance, by saying that all these bugs are using complex combinations of advanced features and none of them pass the checker (I don't know if that's true). Anyways, that's just an idea, not a requirement.

BTW, I think we should also signal for each critical bug since when it was present (I guess the answer is 8.5 for most).

JasonGross:
@Zimmi48 in "most of them seem to imply the universes and either a reduction machine or modules" by "imply" do you mean "implicate" (or "involve")?

ppedrot:
I don't fully understand the VM invariants, but couldn't we be in a situation where we sandwich universe and term arguments, so that just counting their respective number is incorrect?

mattam82:
I think the VM relies on the fact that universes are quantified in a prenex way throughout, at least when we adapted the code in 8.5

ppedrot (approved):
I convinced myself that this was indeed correct.

ppedrot:
@maximedenes Can you rebase quickly so that this is backported to 8.8.1?

Zimmi48:
Don't worry though as this PR and #7798 are the two blockers for 8.8.1.
BTW, it is usually OK to fix the conflicts manually during the merge when only CHANGES is affected but in this case I would prefer if it could be fixed by rebasing instead because otherwise it means I have to fix them once again during backporting ;-)

Zimmi48:
@maximedenes Can you rebase so that @ppedrot can merge?

maximedenes:
I'll rebase once @bgregoir has reviewed it, otherwise it may be for nothing.

Zimmi48:
You ask for his review 10 days ago and he didn't react yet. Can you ensure that he does today?

maximedenes:
> You asked for his review 10 days ago and he didn't react yet. Can you ensure that he does today?

We have looked at it, and the fix looks dead wrong so it is going to require more work.

Zimmi48:
OK thanks for letting me know. The 8.8.1 release is going to have to wait then :'(

maximedenes:
Yeah sorry about that, but I don't want to merge an incorrect fix in a hurry. Unfortunately, I'm lacking sleep again but it should be better next week.

maximedenes:
Here's an example showing that the patch is incorrect:

```coq
Set Universe Polymorphism.
Definition bar@{i} := Type@{i}.
Definition foo@{i j} (x y z : nat) :=
  @id Type@{j} bar@{i}.
Eval vm_compute in foo.
```

I'l put it in the test suite.

maximedenes:
@ppedrot This should be ready, modulo CI. I added some test case and comments.

ppedrot:
Spurious CI failures.

Zimmi48:
Is it OK to backport this even with the change in the `comp_env` type? I guess it is an that's not really an interface that Coq plugin developers are supposed to be interfacing with...

Zimmi48:
I did backport it like this. Let me know if that was wrong.

ppedrot:
I don't see any problem for the backport personally, this is really an internal API.

herbelin:
@Zimmi48: Do you also intend to backport this to versions prior to 8.8?

And do I understand correctly that the bug starts with universe polymorphism in 8.5?

maximedenes:
> And do I understand correctly that the bug starts with universe polymorphism in 8.5?

I haven't checked if it could already be exploited, but yes, the incorrect code was there since 8.5.

PR #7770, opened 2018-06-10
Move indices on top on the TOC. Closes #7764.

### LINKED ISSUE #6573: The credits chapter is so long, it shouldn't come before the table of contents
#### Version

8.7.1

The PDF output of the Reference Manual shows the introduction, then the credits, then (at page 23) the table of contents. Putting the table of contents right after the introduction would mean at page 5, this would be much more reasonable.

### LINKED ISSUE #7764: Make the manual indexes quickly accessible
#### Version

8.8.0

AFAIU, the reference manual is not meant to be read sequentially. And in my everyday work, I often need to read the description of some tactic, option, etc. So, every time I use the web-based Sphinx docs, I need to scroll down, because the table of contents does not fit the screen of my laptop.

Moving the indexes to the beginning of the ToC or creating a page with all the indexes (I'm willing to make it my default page) would certainly help. Incidentally, in the PDF version the ToC fits the screen, so there is no problem there.

Zimmi48:
**Kind:** documentation

Grants / closes #7764

This moves the various indexes on top of the Table of Contents, as suggested in #7764. This may look a bit weird so this change needs to be discussed.

Here's how it looks with this change:
- first commit only https://coq.gitlab.io/-/coq/-/jobs/73926160/artifacts/_install_ci/share/doc/coq/sphinx/html/index.html
- both commits https://coq.gitlab.io/-/coq/-/jobs/73989049/artifacts/_install_ci/share/doc/coq/sphinx/html/index.html

The second commit closes #6573.


Zimmi48:
Sounds good but I'm not fully sure how you view this: could you describe your idea in greater details or open an alternative PR to this one so that we get a better idea and can compare the various solutions?
In particular, I'm not sure whether you intend this quick links section to be in the TOC or at the top of the index page.

Zimmi48:
I've taken this PR one step further because it felt weird to have the indexes on top of the TOC and then the part "The language". So I've added a new part "Preamble" with "Introduction" and "Credits" in between. Now this looks really good to me. In particular, when we are on the main page, the "Introduction" page appears highlighted. This allows to see very clearly where the manual really begins.

Additionally, this second commit makes Credits an independent chapter of Introduction. I think this is really important if we want people to find and read Credits. The current situation makes them almost disappear.

anton-trunov:
@Zimmi48 This looks awesome! And it's much easier to use for a quick lookup! Thank you!

maximedenes (approved):
Looks very good to me. I never know, is it "indexes" or "indices"?

anton-trunov:
>I never know, is it "indexes" or "indices"?

@maximedenes If it is "lists of names or topics that are referred to in a book" then the it's called "indexes". If it's plural of the common mathematical term, then it's "indices".

maximedenes:
@anton-trunov Thanks a lot! Will merge, then.

maximedenes:
The documentation build on gitlab seems to have failed, retrying.

maximedenes:
Failed again. @Zimmi48, any idea where that could come from?

Zimmi48:
I recently discovered this to two plural thing. Would you like me to amend the commit?

Zimmi48:
As for the CI failures, they are always the same. It's very annoying see #7829.

Zimmi48:
This time the documentation build on CI passed: https://coq.gitlab.io/-/coq/-/jobs/75409965/artifacts/_install_ci/share/doc/coq/sphinx/html/index.html

PR #7772, opened 2018-06-11
[native_compute] Delay computations with toplevel match

maximedenes:
This is an easy improvement on examples like:

Fixpoint zero (n : nat) :=
  match n with
  | O => O
  | S p => zero p + zero p
  end.

Definition foo := if true then zero 100 else 0.

Eval native_compute in if true then 0 else foo.

I didn't add a test case, because our current testing infrastructure is too
fragile for that.

### DIFF
--- a/kernel/nativelambda.ml
+++ b/kernel/nativelambda.ml
@@ -432,7 +432,6 @@ module Renv =
 	r
   end
 
-(* What about pattern matching ?*)
 let is_lazy prefix t =
   match kind t with
   | App (f,args) ->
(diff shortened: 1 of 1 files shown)


ppedrot:
Why precisely this heuristic? We don't we do that for every non-value?

ppedrot:
For the lack of contenders, I'm assigning myself.

maximedenes:
> Why precisely this heuristic? We don't we do that for every non-value?

I believe this is what it was trying to implement, just missing the match case.

The real problem is that this test should be done after compilation, through a straightforward analysis of the generated code (whether it is a value or not). Here we are trying to guess the shape of the generated code from the input, which is ugly.

Nonetheless, this PR is an improvement.

ppedrot (approved):

ppedrot:
CI failures look spurious.

PR #6121, opened 2017-11-08
Fix the check in configure.ml following instructions in INSTALL.

Zimmi48:

PR #7774, opened 2018-06-11
[build] Fix checks and notes noting 4.02.1 instead of 4.02.3

ejgallego:
Bumping to 4.02.3 was decided some time ago in the WG, however a
couple of places escaped updating.

### DIFF
--- a/INSTALL.ide
+++ b/INSTALL.ide
@@ -22,7 +22,7 @@ Else, read the rest of this document to compile your own CoqIde.
 
 COMPILATION REQUIREMENTS
 
-- OCaml >= 4.02.1 with native threads support.
+- OCaml >= 4.02.3 with native threads support.
 - make world must succeed.
 - The graphical toolkit GTK+ 2.x. See http://www.gtk.org.
   The official supported version is at least 2.24.x.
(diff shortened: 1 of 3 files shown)


Zimmi48:
This was already proposed in #6121. I'd rather not have this discussion again. This "inconsistency" will disappear by itself when we upgrade to 4.05 anyways.

maximedenes:
One thing that I really don't like with the status quo (which is due to a misunderstanding during the WG discussion btw), is that I don't think anything would detect if we release a 8.8.1 that does not compile with 4.02.1.

ejgallego:
I didn't want to get into another version fight, but we discussed this in person this year and indeed using 4.02.1 is a bug; to all extents 4.02.1 is unsupported today. 

Also, now it seems like we have a real reason to bump the dependency to 4.02.3 so let's do it.

ejgallego:
Note than in my view the upgrade to 4.05.0 will likely happen for 8.10, so we shouldn't rely too much on it for the short term.

SkySkimmer (approved):

ejgallego:
@letouzey may be busy and I am configure's 2nd, so I'd suggest @maximedenes takes care of this.

maximedenes:
> This was already proposed in #6121. I'd rather not have this discussion again. This "inconsistency" will disappear by itself when we upgrade to 4.05 anyways.

That discussion concluded it should not revert a commit of mine that was meant to be temporary. I never understood why we didn't fix the problem. So let's do it now!

PR #7775, opened 2018-06-11
Fix handling of universe context for expanded program obligations.

### LINKED ISSUE #75: Equations lemma takes a long time to prove
The `Admit` in the end takes 30 seconds to go through. Apparently it is due to lemma  `foldLeft_unfold_eq' taking a while to prove. (thanks to the people on the coq IRC channel)

Removing  `(l: List T)` from either `f_type` or  `foldLeft_type` makes the `Admit` go through quite fast. 
Replacing Cons_type with List as well.

```coq
Require Equations.Equations.

Set Program Mode.

Definition ifthenelse b A (e1: true = b -> A) (e2: false = b -> A): A. Admitted.
  
Definition List (T: Type): Type. Admitted.
Definition isCons (T: Type) (src: List T): bool. Admitted.
Definition Cons_type (T: Type): Type := {src: List T | (isCons T src = true)}.
Definition head (T: Type) (src: Cons_type T): T. Admitted.
Definition tail (T: Type) (src: Cons_type T): List T. Admitted.
Definition f_type (T R: Type) (l: List T): Type := R -> (T -> R).
Definition foldLeft_type (T R: Type) (l: List T): Type := R.

Equations foldLeft (T R: Type) (l: List T) (z: R) (f: f_type T R l): foldLeft_type T R l := 
foldLeft T R l z f by rec 0 lt :=
foldLeft T R l z f :=
  ifthenelse (isCons _ l) R
             (fun _ => foldLeft T R (tail T l) (f z (head T l)) f)
             (fun _ => z).

Admit Obligations.
```

mattam82:
The universe context was dropped even though it isn't added to the
global universes yet. Keep it so that it is properly defined with
the constant the expanded obligation appears in.

**Kind:**  bug fix

Fixes / closes mattam82/Coq-Equations#75

### DIFF
--- a/vernac/obligations.ml
+++ b/vernac/obligations.ml
@@ -865,7 +865,7 @@ let obligation_terminator name num guard hook auto pf =
       else UState.union prg.prg_ctx ctx
     in
     let uctx = UState.const_univ_entry ~poly:(pi2 prg.prg_kind) ctx in
-    let (_, obl) = declare_obligation prg obl body ty uctx in
+    let (defined, obl) = declare_obligation prg obl body ty uctx in
     let obls = Array.copy obls in
     let _ = obls.(num) <- obl in
     let prg_ctx =
@@ -874,10 +874,12 @@ let obligation_terminator name num guard hook auto pf =
             polymorphic obligation with the existing ones *)
         UState.union prg.prg_ctx ctx
       else
-        (** The first obligation declares the univs of the constant,
+        (** The first obligation, if defined,
+            declares the univs of the constant,
             each subsequent obligation declares its own additional
             universes and constraints if any *)
-        UState.make (Global.universes ())
+        if defined then UState.make (Global.universes ())
+        else ctx
     in
     let prg = { prg with prg_ctx } in
     try


Zimmi48:
@mattam82 ping (if you want this in 8.8.1)

mattam82:
Not sure it’s easy, I don’t think it’s possible to produce such a case from
a Coq .v file readily. We’d need a call to add_definition with an empty
obligations array. I’ll check.

Le dim. 17 juin 2018 à 10:26, Maxime Dénès <notifications@github.com> a
écrit :

Zimmi48:
You could also try using the unit test framework for this...

Zimmi48:
Interesting test case @mattam82, although this one fails in 8.8.0 (and current v8.8) but already succeeds in master. It would be good to know what fixed it in master.

Zimmi48:
You will need to rebase on current master to have CI run properly BTW.

mattam82:
I was trying to find a test-case which would trigger the problem but this one actually doesn't and was certainly fixed by one of the commits related to program's handling of universes in master.

mattam82:
I couldn't use the unit testing either, what would be needed is a mock plugin and a .v file using it. I've no time to do that now.

Zimmi48:
Could you push the test-suite file that was fixed in master but not in v8.8 in another PR so that we can merge it and find which commit fixed it?

maximedenes:
> I couldn't use the unit testing either, what would be needed is a mock plugin and a .v file using it. I've no time to do that now.

Sounds a bit heavy indeed. Thanks for the answer.

mattam82:
It's fixing a universe undefined bug, because we're dropping the constraints of the "expanded" term. When it is defined as a constant the universe constraints are registered with it. Otherwise they have to be kept and declared with the enclosing term in which the obligation body is expanded.

maximedenes (approved):

PR #7782, opened 2018-06-12
[VM] Rename reloc -> cenv

maximedenes:
The renaming is internal only. I believe the name reloc is legacy and
a bit confusing now that the structure contains a full compilation
environment.


maximedenes:
@ppedrot can you take this one? It is a trivial alpha renaming.

ppedrot (approved):
Does look like α-renaming to me.

PR #7721, opened 2018-06-06
Add a note about [ci skip] in CI README.

Zimmi48:
**Kind:** documentation

Follows a discussion at #7601.

PR #7783, opened 2018-06-12
Move INSTALL.doc to doc/README.md and improve a few things.

Zimmi48:
**Kind:** documentation

Move `INSTALL.doc` to `doc/README.md` and improve a few things. You should look commit by commit to see diffs showing the improvements.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,13 +1,20 @@
-                           The Coq documentation
-                           =====================
+The Coq documentation
+=====================
 
 The Coq documentation includes
 
 - A Reference Manual
 - A document presenting the Coq standard library
 
+The documentation of the latest released version is available on the Coq
+web site at [coq.inria.fr/documentation](http://coq.inria.fr/documentation).
+
+Additionnally, you can view the documentation for the current master version at
+<https://gitlab.com/coq/coq/-/jobs/artifacts/master/file/_install_ci/share/doc/coq/sphinx/html/index.html?job=documentation>.
+
 The reference manual is written is reStructuredText and compiled
-using Sphinx (see `doc/sphinx/README.rst`) to learn more.
+using Sphinx (see [`sphinx/README.rst`](/doc/sphinx/README.rst))
+to learn more about the format that is used.
 
 The documentation for the standard library is generated from
 the `.v` source files using coqdoc.
@@ -15,12 +22,8 @@ the `.v` source files using coqdoc.
 Prerequisite
 ------------
 
-To produce all the documents, the following tools are needed:
+To produce the complete documentation in HTML, the following tools are required:
 
-  - latex (latex2e)
-  - pdflatex
-  - dvips
-  - makeindex
   - Python 3
   - Sphinx 1.6.5 (http://www.sphinx-doc.org/en/stable/)
   - sphinx_rtd_theme
@@ -28,56 +31,64 @@ To produce all the documents, the following tools are needed:
   - beautifulsoup4
   - Antlr4 runtime for Python 3
 
+To produce the documentation in PDF and PostScript formats the following
+additional tools are required:
+
+  - latex (latex2e)
+  - pdflatex
+  - dvips
+  - makeindex
 
 Under recent Debian based operating systems (Debian 10 "Buster",
 Ubuntu 18.04, ...) a working set of packages for compiling the
 documentation for Coq is:
 
-  texlive-latex-extra texlive-fonts-recommended python3-sphinx
-  python3-pexpect python3-sphinx-rtd-theme python3-bs4
-  python3-sphinxcontrib.bibtex python3-pip
+    texlive-latex-extra texlive-fonts-recommended
+    python3-sphinx python3-pexpect python3-sphinx-rtd-theme
+    python3-bs4 python3-sphinxcontrib.bibtex python3-pip
 
 Then, install the Python3 Antlr4 package:
 
-  pip3 install antlr4-python3-runtime
+    pip3 install antlr4-python3-runtime
 
 Nix users should get the correct development environment to build the
-HTML documentation from Coq's `default.nix`. [Note Nix setup doesn't
-include the LaTeX packages needed to build the full documentation.]
+HTML documentation from Coq's [`default.nix`](/default.nix) (note this
+doesn't include the LaTeX packages needed to build the full documentation).
 
-If you are in an older/different distribution you can install the
-Python packages required to build the user manual using python3-pip:
+If you are in an older / different distribution / operating system, you can
+install the Python packages required to build the reference manual using
+python3-pip:
 
-  pip3 install sphinx sphinx_rtd_theme beautifulsoup4 antlr4-python3-runtime pexpect sphinxcontrib-bibtex
+    pip3 install sphinx sphinx_rtd_theme beautifulsoup4 antlr4-python3-runtime pexpect sphinxcontrib-bibtex
 
 Compilation
 -----------
 
 To produce all documentation about Coq, just run:
 
-   ./configure            (if you hadn't already)
-   make doc
+    ./configure           # (if you hadn't already)
+    make doc
 
 
 Alternatively, you can use some specific targets:
 
-   make doc-ps
-        to produce all PostScript documents
+- `make doc-ps`
+  to produce all PostScript documents
 
-   make doc-pdf
-        to produce all PDF documents
+- `make doc-pdf`
+  to produce all PDF documents
 
-   make doc-html
-        to produce all html documents
+- `make doc-html`
+  to produce all HTML documents
 
-   make sphinx
-        to produce the HTML version of the reference manual
+- `make sphinx`
+   to produce the HTML version of the reference manual
 
-   make stdlib
-        to produce all formats of the Coq standard library
+- `make stdlib`
+  to produce all formats of the Coq standard library
 
 
-Also note the "-with-doc yes" option of ./configure to enable the
+Also note the `-with-doc yes` option of `./configure` to enable the
 build of the documentation as part of the default make target.
 
 
@@ -86,6 +97,7 @@ Installation
 
 To install all produced documents, do:
 
-  make DOCDIR=/some/directory/for/documentation install-doc
+    make DOCDIR=/some/directory/for/documentation install-doc
 
-DOCDIR defaults to /usr/share/doc/coq
+`DOCDIR` defaults to `/usr/share/doc/coq` or whatever value was specificied
+through the `-docdir` option of `./configure`.
--- a/doc/README.md
+++ b/doc/README.md
@@ -62,7 +62,7 @@ To produce all documentation about Coq, just run:
 Alternatively, you can use some specific targets:
 
    make doc-ps
-   	to produce all PostScript documents
+        to produce all PostScript documents
 
    make doc-pdf
         to produce all PDF documents
--- a/doc/README.md
+++ b/doc/README.md
@@ -13,25 +13,41 @@ Additionnally, you can view the documentation for the current master version at
 <https://gitlab.com/coq/coq/-/jobs/artifacts/master/file/_install_ci/share/doc/coq/sphinx/html/index.html?job=documentation>.
 
 The reference manual is written is reStructuredText and compiled
-using Sphinx (see [`sphinx/README.rst`](/doc/sphinx/README.rst))
+using Sphinx. See [`sphinx/README.rst`](sphinx/README.rst)
 to learn more about the format that is used.
 
 The documentation for the standard library is generated from
 the `.v` source files using coqdoc.
 
-Prerequisite
+Dependencies
 ------------
 
-To produce the complete documentation in HTML, the following tools are required:
+### HTML documentation
 
-  - Python 3
-  - Sphinx 1.6.5 (http://www.sphinx-doc.org/en/stable/)
-  - sphinx_rtd_theme
-  - pexpect
-  - beautifulsoup4
-  - Antlr4 runtime for Python 3
+To produce the complete documentation in HTML, you will need Coq dependencies
+listed in [`INSTALL`](../INSTALL). Additionally, the Sphinx-based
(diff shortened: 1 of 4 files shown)


doc/README.md
```diff
@@ -0,0 +1,103 @@
...
+  - dvips
+  - makeindex
+
+Under recent Debian based operating systems (Debian 10 "Buster",
+Ubuntu 18.04, ...) a working set of packages for compiling the
+documentation for Coq is:
+
+    texlive-latex-extra texlive-fonts-recommended
+    python3-sphinx python3-pexpect python3-sphinx-rtd-theme
+    python3-bs4 python3-sphinxcontrib.bibtex python3-pip
```
jfehrle: Would it be correct/useful to refer to these as "python packages"?  Also installed with pip?  Or are they some kind of Debian-specific packages?

doc/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+  - Python 3
+  - Sphinx 1.6.5 (http://www.sphinx-doc.org/en/stable/)
+  - sphinx_rtd_theme
+  - pexpect
+  - beautifulsoup4
+  - Antlr4 runtime for Python 3
+
+To produce the documentation in PDF and PostScript formats the following
+additional tools are required:
```
jfehrle: which are installed as binaries (not Python packages).  (The reader should search for and install each tool)

doc/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Installation
+------------
+
+To install all produced documents, do:
+
+    make DOCDIR=/some/directory/for/documentation install-doc
+
+`DOCDIR` defaults to `/usr/share/doc/coq` or whatever value was specificied
+through the `-docdir` option of `./configure`.
```
jfehrle: Seems to me the usual practice should be to use `-docdir`, with `DOCDIR` as a less-preferred alternative.  Where does the doc end up if you only specify `-local`?  Seems to me you shouldn't need an install step for the -local case (just slows down the edit-build-review cycle).

doc/README.md
```diff
@@ -0,0 +1,103 @@
...
+python3-pip:
+
+    pip3 install sphinx sphinx_rtd_theme beautifulsoup4 antlr4-python3-runtime pexpect sphinxcontrib-bibtex
+
+Compilation
+-----------
+
+To produce all documentation about Coq, just run:
+
+    ./configure           # (if you hadn't already)
```
jfehrle: Maybe `./configure <PARAMETERS>`?

jfehrle:
Thanks for renaming INSTALL.doc and the wording improvements.  Is it possible to also rename Makefile.doc, which my IDE treats as an MS Word file?  (I asked the vendor whether there's a way to disable that behavior in my IDE but haven't heard anything back yet.)

doc/README.md
Zimmi48: Well it is supposed to work even without parameters.

doc/README.md
Zimmi48: If you use `-local` you are not supposed to use install indeed.

doc/README.md
jfehrle: It would be nice to get these to work on Windows but probably not necessary for what I want to do.

Cygwin has a long list of packages that can be installed: (https://cygwin.com/packages/package_list.html) but it appears not to include the 4 listed here.  (cygwin packages are new to me)

doc/README.md
Zimmi48: All the ones starting with `python3` are Python packages and can be installed using pip.

jfehrle:
Off topic: Would it make sense to make coq available as a cygwin package?

Zimmi48:
Thanks for your comment @jfehrle. I've taken them into account and reworded / reorganized a lot of stuff in this document.

About the `Makefile.doc` question, I suppose you could open a PR proposing to change this to `Makefile.docs`. Mind you that you would also have to update it in the three following places:

```
$ git grep "Makefile\.doc"
.github/CODEOWNERS:/Makefile.doc        @maximedenes
Makefile.build:include Makefile.doc      ## provides the 'documentation' rule
dev/doc/build-system.dev.txt:- Makefile.doc : specific rules for compiling the documentation.
```

doc/README.md
```diff
@@ -19,52 +19,50 @@ to learn more about the format that is used.
...
+### HTML documentation
 
-  - Python 3
-  - Sphinx 1.6.5 (http://www.sphinx-doc.org/en/stable/)
-  - sphinx_rtd_theme
-  - pexpect
-  - beautifulsoup4
-  - Antlr4 runtime for Python 3
+To produce the complete documentation in HTML, you will need the dependencies
+of Coq listed in [`INSTALL`](/INSTALL). Additionally, the Sphinx-based
```
jfehrle: "Coq dependencies" is more idiomatic but "dependencies of Coq" is quite understandable.

doc/README.md
```diff
@@ -19,52 +19,50 @@ to learn more about the format that is used.
 The documentation for the standard library is generated from
```
jfehrle: The closing parenthesis for the comment "see INSTALL to learn more..." should be after "... the format that is used."  Or it could a sentence on its own without parentheses.

doc/README.md
```diff
@@ -0,0 +1,102 @@
...
+Compilation
+-----------
+
+To produce all documentation about Coq in all formats, just run:
+
+    ./configure           # (if you hadn't already)
+    make doc
+
+
+Alternatively, you can use some specific targets:
```
jfehrle: Maybe "you can make specific targets".

doc/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+  - latex (latex2e)
+  - pdflatex
+  - dvips
+  - makeindex
+
+Install them using your package manager. E.g. on Debian / Ubuntu:
+
+    apt install texlive-latex-extra texlive-fonts-recommended
+
```
jfehrle: I had hoped to give you instructions on how to install using the cygwin package manager, but that isn't working for me yet.

doc/README.md
Zimmi48: > Or does "make sphinx" do the compile, too?

It does. And having an installed Coq is not sufficient.

doc/README.md
```diff
@@ -0,0 +1,102 @@
...
+- `make doc-html`
+  to produce all HTML documents
+
+- `make sphinx`
+   to produce the HTML version of the reference manual
+
+- `make stdlib`
+  to produce all formats of the Coq standard library
+
```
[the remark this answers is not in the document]
Zimmi48: I prefer the current formulation for now since there is no way to produce the reference manual in PDF or PS format at the present time. (Will probably be reintroduced again in the future.)

doc/README.md
jfehrle: Isn't it true that if you're able to build Coq from the source tree, then you've satisfied the dependencies?  The way it's worded suggests that the reader should go read INSTALL, which may be unnecessary.  Maybe just add something like "(If you're able to build Coq, the dependencies are already satisfied.)"

doc/README.md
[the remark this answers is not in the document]
jfehrle: Ah, then the doc should say that there's currently no way to produce the ref manual in PDF/PS.  As currently worded, I would expect "all PDF documents" and "all HTML documents" to produce the same documents (in different formats).

doc/README.md
maximedenes: @Zimmi48 what about this suggestion?

Zimmi48: I don't like it too much but if you'd prefer like this I can do the change.

maximedenes:
Ok, I think this is good to merge.

maximedenes (approved):
Was already carefully re-read by @jfehrle , so I didn't do it again. I support the move.

maximedenes (changes requested):
Wait! Why is the gilab run shown as success, but in fact empty? I'm pretty sure this actually breaks the build because you forgot to fix a path in the Wnidows script. If our CI infrastructure can show success without running any test, that should be fixed.

SkySkimmer:
>Wait! Why is the gilab run shown as success, but in fact empty? I'm pretty sure this actually breaks the build because you forgot to fix a path in the Wnidows script. If our CI infrastructure can show success without running any test, that should be fixed.

Looks like [ci skip] on the last commit causes that.

maximedenes:
> Looks like [ci skip] on the last commit causes that.

Oh, CI shows a green mark and "pipeline skipped". But I think it is generally a bad idea to skip CI.

Zimmi48:
Hum, what's the problem here, I only touch non-compiled files. Did you miss #7721 @maximedenes?

maximedenes:
> Hum, what's the problem here? I only touch non-compiled files.

Yes, but that's unrelated to breaking the build or not. Can you launch a build?

This should fail:

https://github.com/coq/coq/blob/df35025b2be4a0dc9aadecc0e3110a21012683cf/dev/build/windows/makecoq_mingw.sh#L1137

except if we ignore the result, of course, but then it's another problem.

maximedenes:
> Did you miss #7721 @maximedenes?

I don't know if you were expecting me to see it. There is close to 0 probability that I see a PR submitted and merged the same day, when I'm not involved as the author, the assignee or mentioned.

Zimmi48:
Hum, very sorry about that. I removed the `[ci skip]` to confirm the failure and I shall fix it tomorrow. Maybe we should revert #7721 as well then.

Zimmi48:
I've fixed the Windows install step. I don't like these specific quirks at all and will open an issue about this.

Zimmi48:
See #7896

MSoegtropIMC (commented):
Shouldn't we add instructions on how to build it on Windows to the INSTALL file?

Zimmi48:
Yes, as soon as someone manages to actually build it on Windows (I assume you are talking about the doc). @jfehrle tried and nearly succeeded but I don't know what his situation is now. In any case, this can be added in a later PR.

MSoegtropIMC:
@Zimmi48 : I was more talking about building on Windows in general, but I guess it is best if we do this together with doc building + building for plugin developers.

@ejgallego : did you find some time to check Merlin + VisualStudio Code? The rest in terms of plugin development on Windows is working. If you don't have time I can also look into it.

jfehrle:
I have "make sphinx" working on Windows.  The problem I described in coqdev (New failure for "make sphinx") appears to be my mistake.  My setup for OCaml and other tools is a mess--I have multiple cygwin installations when I should have only one.  I use one to compile Coq and use another to make sphinx--neither can do both :-(.  Reinstalling my entire toolchain should fix that.

The non-obvious parts of the setup were related to support for `/cygdrive/c/...` pathnames:
- Installing cygwin make (non-cygwin make won't work)
- PR #7815

Zimmi48:
I think I'll let you @MSoegtropIMC and @jfehrle document the setup in `INSTALL` and `doc/README.md` at once: if I'm not mistaken, the recommendation for developing Coq on Windows should be:

Install a full cygwin (avoid finding out later than we are missing a few packages), then:

```
OPAM_SWITCH=4.06.1+mingw64c

wget https://github.com/fdopen/opam-repository-mingw/releases/download/0.0.0.1/opam64.tar.xz
tar -xf opam64.tar.xz
bash opam64/install.sh

opam init -a mingw https://github.com/fdopen/opam-repository-mingw.git --comp $OPAM_SWITCH --switch $OPAM_SWITCH
eval "$(opam config env)"
opam install -y num ocamlfind camlp5 ounit
```

(these lines are taken from `dev/ci/appveyor.sh`, maybe they should be factorized out so that users can execute them via a shell wrapper)

Then: install pip packages as described in `doc/README.md`.

MSoegtropIMC:
@Zimmi48 : well the way the official installers are built is entirely different. You call a batch file which first installs a fresh cygwin with the required packages and then compiles OCaml and Coq from sources without opam.

But I wouldn't recommend this method for development of Coq, und I guess unless you want to do Coq development there is no good reason to compile it from sources. For Coq development I would suggest to use opam, because e.g. Merlin is pretty much impossible to compile without opam.

But I would still recommend to have a batch file which sets up everything automatically. The simple reason being that (in my experience, sorry no offense to anybody on this list) people are to lazy to read instructions which are longer than 5 lines and simply do it wrong and then complain that it doesn't work. In my experience simply providing a batch file is less maintenance effort.

And I think it makes sense to document both methods - the way the official installers are built e.g. in case someone wants to create a custom installer with specific plugins and the opam method intended for Coq development.

Btw.: For Coq development there is also the option to create (with opam) cygwin executables of coq. I didn't as yet test which method works better in practice. In the past OCaml debugging did work better with cygwin than with MinGW, but that might have changed.

I already have such a batch file (worked on it in Nice). The only thing missing is that Merlin does not communicate properly with Visual Studio Code and @ejgallego wanted to look into this. So we have these options now:

* drop Visual Studio Code support for now and recommend (and install and setup) emacs instead
* wait for Emilio to look into Merlin and Visual Studio Code
* wait until I had time to look into this issue (might take 3 or 4 weeks from now, since I would have to work into the structure of Visual Studio Code plugins - maybe not a bad things anyway because the Coq plugin is also way suboptimal.

What are your suggestions?

Zimmi48:
> What are your suggestions?

Perfect is the enemy of the good. I would suggest sharing what you've already managed to reach (possibly without merlin support yet) and enhance it later on.

BTW Emilio is already busy on so many topics, and like many other devs he tends to take on new responsibilities too easily. For VSCode + merlin on Windows, I suggest you talk instead to other OCaml + Windows users outside of the Coq developers.

More precisely, if what is available here https://github.com/reasonml-editor/vscode-reasonml doesn't work on Windows, open an issue on their issue tracker.

MSoegtropIMC:
@Zimmi48 : let me check if I can get Merlin to work easily with Emacs - I guess this should be much easier. I wanted to publish a somewhat rounded up package, which doesn't open more questions than it answers. I am just terribly busy the next 2 weeks or so.

maximedenes (approved):

PR #7786, opened 2018-06-12
In "redundant clause" pattern-matching error, show also the pattern (closes #7777)

### LINKED ISSUE #7777: "Error: This clause is redundant." does not give adequate location information when the clause contains multiple alternatives
#### Version

8.8.0

```coq
Inductive foo := aa | bb | cc.
Definition bar (f : foo) : Set
  := match f with
     | aa
     | bb
     | bb
       => nat
     | cc => nat
     end.
(* Toplevel input, characters 56-92:
> Definition bar (f : foo) : Set   := match f with      | aa      | bb      | bb        => nat      | cc => nat      end.
>                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This clause is redundant.
*)
```
I would like to know which of the alternatives is redundant.

cc @herbelin

herbelin:
Closes #7777 according to the discussion there.

The refined error message is particularly useful when the pattern is part of a disjunction.

Maybe this could be improved though, not mentioning the pattern when there is no disjunction, but that would be more work.

- [X] Added output test in test-suite

### DIFF
--- a/test-suite/output/Cases.out
+++ b/test-suite/output/Cases.out
@@ -169,3 +169,5 @@ fun x : K => match x with
              | _ => 2
              end
      : K -> nat
+The command has indeed failed with message:
+Pattern "S _, _" is redundant in this clause.
--- a/test-suite/output/Cases.v
+++ b/test-suite/output/Cases.v
@@ -217,3 +217,6 @@ Check fun x => match x with a3 | a4 => 3 | _ => 2 end.
 Check fun x => match x with a3 => 3 | a2 | a1 => 4 | _ => 2 end.
 Check fun x => match x with a4 => 3 | a2 | a1 => 4 | _ => 2 end.
 Check fun x => match x with a3 | a4 | a1 => 3 | _ => 2 end.
+
+(* Test redundant clause within a disjunctive pattern *)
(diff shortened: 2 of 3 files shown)


ppedrot (approved):

ppedrot:
As discussed on gitter the phrasing sounds strange, could you simplify / correct it, e.g. removing the check for 'multiple'?

mattam82 (approved):

herbelin:
I have no particular share on `multiple`. This is just the terminology used in the reference manual ([here](https://coq.inria.fr/refman/language/gallina-extensions.html#multiple-and-nested-pattern-matching)). What would you propose? To simply drop `multiple`? If noone objects, that's ok to me.

ppedrot:
@herbelin I think the simpler is to drop 'multiple' indeed. If you can rebase and fix this, I'll merge the PR.

herbelin:
> I think the simpler is to drop 'multiple' indeed.

OK, done, and rebased. Thanks.

Zimmi48:
> This branch is 1 commit ahead, 493 commits behind coq:master. 

For some reason, this branch was not properly rebased but @coqbot didn't catch it. I'm looking into this.

PR #7787, opened 2018-06-12
Fixes #7780: missing lift in expanding alias under a binder in unification

### LINKED ISSUE #7780: Stack overflow in refine / elaboration
#### Version

8.8.0

#### Operating system

Linux

```coq
Goal forall T (t : T) (P P0 : T -> Set), option (option (list (P0 t)) -> option (list (P t))).
  intros ????.
  refine (Some
            (fun rls
             => let mkcons := _ in
                let default _ := _ in
                match rls with
                | Some _ => _
                | None => _
                end)).
  (* Error: Stack overflow. *)
```

I also get a stack overflow on
```coq
Definition foo T (t : T) (P P0 : T -> Set) : option (option (list (P0 t)) -> option (list (P t)))
:= Some
     (fun rls
      => let mkcons := _ in
         let default _ := _ in
         match rls with
         | Some _ => _
         | None => _
         end).
```

herbelin:
**Kind:** bug fix

Fixes #7780 (see commit for the exact short fix)

- [X] Added test in test-suite

### DIFF
--- a/pretyping/evarsolve.ml
+++ b/pretyping/evarsolve.ml
@@ -929,7 +929,7 @@ let invert_arg_from_subst evd aliases k0 subst_in_env_extended_with_k_binders c_
     with Not_found ->
       match expand_alias_once evd aliases t with
       | None -> raise Not_found
-      | Some c -> aux k c in
+      | Some c -> aux k (lift k c) in
   try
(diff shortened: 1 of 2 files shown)


mattam82 (approved):
Alright with me, aliases are already k0 lifted but have to be lifted again when expanded under the additional k binders.

PR #7791, opened 2018-06-12
Deprecating the declaration of arbitrary terms as hints.

ppedrot:
We restrict hints to be global references, so as to further simplify the implementation. Allowing arbitrary terms makes it difficult or expensive to handle properly some actions like universe contexts or hint equality. Ultimately, the `IsConstr` constructor for hints should also be removed.

This feature was seemingly not used in the stdlib. I don't think it is use much anyways.

- [x] Entry added in CHANGES.


Zimmi48:
It is used in the wild actually: cf. https://github.com/Zimmi48/coq/tree/clean-up/remove-constr-in-hints and the failures at https://gitlab.com/Zimmi48/coq/pipelines/22914759

Zimmi48:
Many of these failures are spurious or unrelated but at least Color is really using the feature.

ppedrot:
@Zimmi48 FTR, color is a bit... *peculiar* in its use of Coq features. I had to port it by hand after the deprecation of identifier-less fix tactic invocations, and the least I can say is that it seems to be either from another Coq era, or maybe even another development plane!

Zimmi48:
I did this in the CIW indeed.

ppedrot:
Anyways, if Color is the only user, it's probably doable to deprecate it anyways. They'll find a way around for the next version, e.g. defining compatibility global values.

Zimmi48:
I'll try to confirm this.

JasonGross:
Also used in fiat-parsers, if I recall correctly.  I'm not sure if it's used in fiat-crypto.

gares:
> like universe contexts

The beauty of let-polymorphism strikes back ;-)

mattam82 (approved):
Yes please! For discharge as well...

Zimmi48:
Indeed, it was to help with discharge that I had started this branch.

Zimmi48:
So it's in fact used in color, cross-crypto, Equations, fiat-crypto, fiat-parsers, math-classes, HoTT, stdpp, math-comp, UniMath, VST, and the test-suite. So I wouldn't say it not much used.

Zimmi48:
https://github.com/fblanqui/color/blob/d2ff19f24dd56be7bae8a835e58bb62a32fb81d6/Util/Relation/RelExtras.v#L67-L69

https://github.com/adampetcher/fcf/blob/cbd288a6c125a1b47899499fad1006206d8e19df/src/FCF/Broken/Procedure.v#L178-L180

Sorry that was FCF, not fiat-crypto.

https://github.com/mattam82/Coq-Equations/blob/1c8b501a5503eb3006fd1c100a6041ef07fca78b/test-suite/rec.v#L135

https://github.com/thery/coqprime/blob/59e3bf69a84c593ad733b83dbcfa90036f5d052a/src/Coqprime/PrimalityTest/Cyclic.v#L57

Sorry that was coq-prime, not fiat-crypto.

https://github.com/mit-plv/fiat/blob/9b8229bc438becf6694e4ea02cbd486ac1dfd34e/src/Common/MSetExtensions.v#L110

https://github.com/math-classes/math-classes/blob/d1903b554b7912d272f80ce016def6d6495c64e6/theory/cut_minus.v#L15

https://github.com/HoTT/HoTT/blob/092ebba509710c6e3eadbe03bc68049dd0b4ea0c/coq/theories/Init/Peano.v#L35-L36

https://gitlab.mpi-sws.org/robbertkrebbers/coq-stdpp/blob/aa942ca85603ae4e88a963abc7691e77ed3c46a7/theories/namespaces.v#L97

https://github.com/math-comp/math-comp/blob/bcba70f86ac39c3d234da0d696c4b32a741e71cf/mathcomp/algebra/ssrnum.v#L1259

https://github.com/UniMath/UniMath/blob/3f8d8cc076c3adc5956653f1d04bd56256b6ec23/UniMath/Foundations/PartA.v#L348

https://github.com/PrincetonUniversity/VST/blob/e679c09e351369dbfefda59c05e3b1aac9ecbae8/msl/sepalg.v#L331

Note that for Equations, math-comp, UniMath and VST, the use case is `Hint Resolve @foo.`

ppedrot:
> Note that for Equations, math-comp, UniMath and VST, the use case is Hint Resolve @foo.

Is there a difference of semantics with `Hint Resolve foo`, at least in the monomorphic case?

Zimmi48:
I wouldn't have expected one. I was very surprised to see this. But the fact that it was used let me think that there is actually one, whether it's on purpose or not.

ppedrot:
@mattam82 et al. What do we do with this PR then? Is it reasonable to merge it or should I close it for a better future?

ppedrot:
Also note that deprecated doesn't necessarily mean we're going to remove the feature soon. It can just be a signal for the user not to use this construct.

Zimmi48 (approved):
I think we should proceed with the merge and we can take our time indeed to really remove the feature.

Zimmi48:
If you want an all green CI, a rebase could help. Otherwise, things are already good like this.

ppedrot:
Rebased. Next in line should be the `prolog` tactic.

mattam82:
I'm using @foo because I really want the reference to be used as a hint, not it's partial application, I expect others do it for the same reason.

JasonGross:
@mattam82 
> I'm using `@foo` because I really want the reference to be used as a hint, not it's partial application, I expect others do it for the same reason.

Except that, as far as I can tell, `Hint Resolve @foo` and `Hint Resolve foo` both use the reference, and `Hint Resolve (foo)` uses the partial application.  Perhaps people use `Hint Resolve @foo` because they assume that `Hint Resolve foo` will use the partial application (because that would make sense), even though it doesn't?

ppedrot:
@mattam82 Really? Doesn't the parser always consider that `foo` stands for the global?

mattam82:
At first it does yes, but if your code path just type checks it it can turn into an application to implicits... I think I added the global_or_constr specifically to handle this case cleanly.

ppedrot:
Is there an easy test to check that?

ejgallego:
I'd like to support this as hints are the single place in `tactics` that do require calling `interp/` and IMHO it'd be great to be able to enforce this invariant.

vbgl:
Can we have a rebase to fix the conflicts and run the CI again? Otherwise please update the milestone. Thanks.

maximedenes:
@ppedrot what's the status of this?

ppedrot:
@maximedenes not sure. AFAICT we still did not understand the functor use case. It doesn't prevent us from deprecating the code and letting the users react to that though.

maximedenes:
> AFAICT we still did not understand the functor use case

What do you mean by "the functor use case"?

ppedrot:
@maximedenes writing globref hints inside a functor can generate non-globref hints at instantiation time, and it seems that there are people relying on that.

ppedrot:
After a bit of code-diving, AFAICT modulo not-so-relevant subtleties like monomorphic constraints, `Hint Resolve @foo` and `Hint Resolve foo` behave the same. My guess is that the reason why people wrote the former is probably due to the old behaviour of the command where the latter would indeed add implicit arguments around. I don't know when this was changed though.

Thus this looks safe to merge for me, maybe after trying to remove all uses to the `@foo` variant from our CI.

maximedenes:
What will be the behavior w.r.t functor inlining?

ppedrot:
@maximedenes unchanged, this is unrelated to the current PR which only affects the surface syntax.

maximedenes:
Ok, but it means we won't be able to remove the corresponding code, right? Is there a plan for that?

ppedrot:
I don't know what to do with this PR. I think it should eventually be merged, but there is quite a bit of code to clean first.

maximedenes:
> I think it should eventually be merged, but there is quite a bit of code to clean first.

What code do you mean? You wrote in the PR description that this PR is not used in stdlib. I thought the only concern with this PR is that it is partial, in the sense that we can't remove the dubious code because of functor inlining.

maximedenes:
But it probably is a good step nonetheless, isn't it?

ppedrot:
I think it is a good step. My concern was about external developments, not the stdlib.

maximedenes:
> My concern was about external developments

Ok, but the sooner we merge this, the more help we can get from downstream authors, I guess. Should we merge?

maximedenes:
I'm a bit surprised this does not require a doc update, though. Did you check if hints on terms were mentioned in the refman?

ppedrot:
Indeed, the refman needs to be updated.

ppedrot:
Should be ready now. Just in case users find this really, really annoying, we *might* still undeprecate it.

ppedrot:
To whomever is reading this, this PR should either be merged or closed. In the first case it needs an assignee. Commentators seemed to be vaguely in favour of that option.

ejgallego:
@ppedrot if nobody else volunteers I could help with that in ~2 weeks time.

maximedenes:
@ppedrot the warning is emitted by the parser?!

ppedrot:
@maximedenes that's the standard practice for deprecation it seems, there are other instances spread in the parsing files.

maximedenes:
> that's the standard practice for deprecation it seems

Seems like a very strange practice to me, what's the rationale behind it? Tactic deprecation was certainly not done at parsing time.

maximedenes:
There does not seem to be so many instances in parsers, hopefully. One case when it makes sense is when a syntax is deprecated (replaced by another one which gives the same AST). When the feature itself is deprecated, it shouldn't be the parser which triggers the warning. For example, another component could be declaring a constr as a hint and bypass the warning.

ppedrot:
Do you want me to move the code to the interpretation phase?

maximedenes:
> Do you want me to move the code to the interpretation phase?

This is a hint deprecation, so it should go to the hints system. If you need another API which does not trigger the warning because internal components use it, you can expose another one, marked as not to be used anymore. This way we will communicate the deprecation not only to users, but also to plugin developers.

Does it sound ok?

ppedrot:
@maximedenes should be ready now. I do think we're overengineering this deprecation, which should have occurred a long time ago.

maximedenes:
> should be ready now

Great, thanks!

> I do think we're overengineering this deprecation

Well, deprecations tend to stay for years (cf the deprecation of `Focus`), so better do it properly.

> which should have occurred a long time ago.

I agree with that part.

maximedenes (approved):
Looks good, thanks!

charguer:
I had a hard time following all the discussion. In summary, could you please confirm that I can replace my occurences of `Hint Resolve @foo` with `Hint Resolve foo` even in cases where `foo` begins with maximally-inserted arguments (typically, typeclasses)?  

If I understand well, this would break backward compatibility, because I'm pretty sure that `Hint Resolve foo` used to add as hint a specific, arbitrary instantiation of `foo`. (I'm not sure until which version of Coq, though). 

Furthermore, in terms of documentation, it would be useful for the CHANGELOG entry to mention the name of the depreciation identifier (that is, "deprecated-hint-constr"), it is useful for searching the changelog when encountering the message. Also, this changelog entry could recall the interpretation of `Hint Resolve foo` with respect to maximally inserted arguments.

ppedrot:
> I had a hard time following all the discussion. In summary, could you please confirm that I can replace my occurences of Hint Resolve @foo with Hint Resolve foo even in cases where foo begins with maximally-inserted arguments (typically, typeclasses)?

Yes. The two commands should be equivalent for monomorphic constants.

> If I understand well, this would break backward compatibility,

No changes were introduced by this PR in the semantics of hints, the change of behaviour was introduced ages ago.

ppedrot:
FTR, hints were made to accept global references in d97cd41db7786ee5172bb00fa2efd1c25ce44a4e more than 7 years ago...

PR #7793, opened 2018-06-12
[ci] update docker image to include elpi 1.0.4

gares:


SkySkimmer:
You need to update CACHEKEY in dockerfile, gitlab yml and circle yml.
Then manually run a pipeline with variable SKIP_DOCKER=false.

ejgallego:
As @SkySkimmer says. Also note @gares that if you think it would make sense to update elpi often, then we can move it from the docker file to an `opam install elpi` in the `ci-elpi.sh` file that should be less of a burden to maintain.

Your call tho, it doesn't take a lot of effort to rebuild the Docker image.

ejgallego (changes requested):
Either:
- move the elpi install to `ci-elpi`
- update the Docker image version `CACHEKEY`

gares:
Thanks for the explanations.

I moved the installation of elpi to `ci-elpi`. I kept the version, even if I believe it is not a good idea, it should just pick the latest IMHO.

Would it be fine if I remove the version from `ci-elpi`? If so please give me the time to force push again.

Frankly, given how hard it is to update the docker, having versioned dependencies in `CI_OPAM` is a very bad idea. I think you should move them to the corresponding `ci-foobar` unless they take a very long time to build (and I don't think it is the case for ocamlgraph nor menhir) and they are used by multiple ci jobs (that is not the case AFAICT).

gares:
right, I'll try

gares:
nope :-/

If I'm not mistaken, in order to ask opam to resolve the dependencies one has to have the package in the data base. You can `pin` a file (part of the git checkout) but this also triggers its installation (that you don't want, only only want to install the package dependencies). One would need to fake a repo, that is make a directory called `packages/`, and check out there... also, coq-elpi depends on coq, and one would need to tell opam that coq.dev is installed even if it is not (not via opam). You can `--fake` it, but it is hackish.

I prefer the status quo wrt this PR.

gares:
To recap what one needs in order to use `--deps-ponly` on package `ci-foo` in a clean way:
- `opam repo add coq-core-dev https://coq.inria.fr/opam/core-dev/`
- `opam repo add coq-extra-dev https://coq.inria.fr/opam/extra-dev/` (have in there `foo.ci.dev`)
- `opam install --deps-only foo.ci.dev`
where `foo.ci.dev` points to the same branch tracked by `ci-foo`

Zimmi48:
OK, too bad. Then things look good like this, but you still need to update the CACHEKEY in both `.gitlab-ci.yml` and `.circleci/config.yml`. Then someone has to trigger a manual build to generate the new image.

ejgallego:
> given how hard it is to update the docker

Updating the Docker is not hard, and it reasonably documented.

> having versioned dependencies in CI_OPAM is a very bad idea.

No, it is not. It saves build time, and creates few problems as most packages can work OK with a reasonbly wide range of versions.

> I moved the installation of elpi to ci-elpi. I kept the version, even if I believe it is not a good idea, it should just pick the latest IMHO.

As you are the upstream here you should do whatever your package needs to work, it is your call.

ejgallego (changes requested):
Remove the changes to the docker file.

Zimmi48:
> Remove the changes to the docker file.

Is the install step in `ci-elpi` going to work with the image that already has elpi in it?

Zimmi48:
> Updating the Docker is not hard, and it reasonably documented.

It is reasonably documented but it's still hard.

SkySkimmer:
It's manual, not hard.

Gaëtan Gilbert

ejgallego:
> It's manual, not hard.

^^

ejgallego:
> Is the install step in ci-elpi going to work with the image that already has elpi in it?

It shouldn't matter too much, but of course now `ci-elpi` is a non deterministic script so the possibility of failure due to opam is always there.

gares:
> Is the install step in ci-elpi going to work with the image that already has elpi in it?

If 1.0.3 is already installed then the command would just do nothing. 
It would work by writing `opam install elpi.1.0.4` but I prefer to let the script pick the last version hence my will to change the docker image.

> It saves build time, and creates few problems as most packages can work OK with a reasonbly wide range of versions.

@ejgallego I know I've hard times communicating with you online.
I'm open to your suggestions to improve the situation.
My suggestion (or better polite request) is that you *do read* my messages.

In a discussion you can't just claim the opposite without providing an argument.
I've said "unless they take a very long time to build (I don't think it is the case for ocamlgraph nor menhir)".  https://github.com/coq/coq/pull/7793#issuecomment-396855179

Building elpi takes 5 seconds on my PC, and there is only one user on CI (coq-elpi). So you save nothing.
It has no place in the docker file and it causes the problem raised by @Zimmi48 (first paragraph of this message).

gares:
@Zimmi48 I think I did apply the changes you request (propagate the cache key to circle and gitlab)

SkySkimmer:
>Anyways, I think we are going to deprecate non-versioned depbuild installs so this may be a good moment to start enforcing that , @SkySkimmer , what do you think?

I don't really know where we should go with CI versioning.
We should keep dependencies in the docker though.

ejgallego:
> I don't really know where we should go with CI versioning.

It looks to me that at least for the "base system" we should aim for a reproducible build, for the contribs themselves (c.f. #6724 ) I do think we have a harder time due to frequency update.

> We should keep dependencies in the docker though.

I do agree; however it seems that @gares is requesting an exception for elpi. Personally I am OK with an exception, if you are OK with it.

But I think we both agree in that `CI_OPAM` stays in Docker.

ejgallego (approved):
Official policy is still to have build-deps versioned in Docker, however we grant an exception to Elpi due to upstream request.

Zimmi48 (dismissed):
@gares I preferred the version where you were giving the version of elpi explicitly.

ejgallego:
Docker pipeline build at https://gitlab.com/coq/coq/pipelines/23775040

SkySkimmer:
>I do agree; however it seems that @gares is requesting an exception for elpi. Personally I am OK with an exception, if you are OK with it.

I don't see why there should be an exception.

ejgallego:
> @gares I preferred the version where you were giving the version of elpi explicitly.

Hihi, we don't converge on this XDD

So if we are going to give an explicit version then I guess it makes sense to do the update in Docker then.

ejgallego:
> I don't see why there should be an exception.

Well if elpi is going to release a new version like every 2 days then our Docker setup is not ready for that. But I see your point.

What do you propose? I am fine either way.

SkySkimmer:
Do the update in docker like it was originally.

Gaëtan Gilbert

ejgallego (changes requested):
Ok, so sorry for the roller coaster discussion, both Théo and Gaëtan prefer a deterministic dependency which I also do, so please remove the modification in `ci-elpi` and bump the Docker dependency [and CACHEKEY again].

gares:
I do as you say guys.
If you are OK to have 2 different versions of elpi hardcoded in the CI files... I'm OK too.

Zimmi48:
What you implemented is what @ejgallego was asking before @SkySkimmer's opposition. In that case, the logical follow-up would have been to remove elpi from the image in the next PR updating it.

I'm personally fine with the current state of the PR. However, I'm afraid that what @SkySkimmer has requested is to go back to updating this in the Docker image only...

gares:
Sorry I got lost. It should be ok now.

ejgallego (approved):

Zimmi48:
@ejgallego Are you launching the manual CI build?

ejgallego:
I will launch the manual build and also remove the old image as Enrico didn't bump the Docker image version so the cached version will be incorrectly used.

ejgallego:
Pipeline build https://gitlab.com/coq/coq/pipelines/23792092

SkySkimmer (approved):

gares:
thanks!

PR #7797, opened 2018-06-13
Remove reference name type.

maximedenes:
`reference` was defined as `Ident` or `Qualid`, but the `qualid` type already
permits empty paths. So we had effectively two representations for
unqualified names, that were not seen as equal by `eq_reference`.

We remove the reference type and replace its uses by `qualid`.

This is still work in progress. We need to:
- [X] clean up a few changes we made before introducing `qualid_is_ident`
- [X] port plugins tracked in CI
- [X] add documentation in dev/doc/changes.md

### DIFF
--- a/library/libnames.mli
+++ b/library/libnames.mli
@@ -65,23 +65,28 @@ val restrict_path : int -> full_path -> full_path
     qualifications of absolute names, including single identifiers.
     The [qualid] are used to access the name table. *)
 
-type qualid
+type qualid_r
+type qualid = qualid_r CAst.t
 
-val make_qualid : DirPath.t -> Id.t -> qualid
+val make_qualid : ?loc:Loc.t -> DirPath.t -> Id.t -> qualid
 val repr_qualid : qualid -> DirPath.t * Id.t
 
 val qualid_eq : qualid -> qualid -> bool
 
 val pr_qualid : qualid -> Pp.t
 val string_of_qualid : qualid -> string
-val qualid_of_string : string -> qualid
+val qualid_of_string : ?loc:Loc.t -> string -> qualid
 
 (** Turns an absolute name, a dirpath, or an Id.t into a
    qualified name denoting the same name *)
 
-val qualid_of_path : full_path -> qualid
-val qualid_of_dirpath : DirPath.t -> qualid
-val qualid_of_ident : Id.t -> qualid
+val qualid_of_path : ?loc:Loc.t -> full_path -> qualid
+val qualid_of_dirpath : ?loc:Loc.t -> DirPath.t -> qualid
+val qualid_of_ident : ?loc:Loc.t -> Id.t -> qualid
+
+val qualid_is_ident : qualid -> bool
+val qualid_path : qualid -> DirPath.t
+val qualid_basename : qualid -> Id.t
 
 (** Both names are passed to objects: a "semantic" [kernel_name], which
    can be substituted and a "syntactic" [full_path] which can be printed
@@ -124,20 +129,6 @@ val eq_global_dir_reference :
   global_dir_reference -> global_dir_reference -> bool
 
 (** {6 ... } *)
-(** A [reference] is the user-level notion of name. It denotes either a
-    global name (referred either by a qualified name or by a single
-    name) or a variable *)
-
-type reference_r =
-  | Qualid of qualid
-  | Ident of Id.t
-type reference = reference_r CAst.t
-
-val eq_reference : reference -> reference -> bool
-val qualid_of_reference : reference -> qualid CAst.t
-val string_of_reference : reference -> string
-val pr_reference : reference -> Pp.t
-val join_reference : reference -> reference -> reference
 
 (** some preset paths *)
 val default_library : DirPath.t
--- a/dev/base_include
+++ b/dev/base_include
@@ -229,7 +229,7 @@ let pf_e gl s =
 let _ = Flags.in_debugger := false
 let _ = Flags.in_toplevel := true
 let _ = Constrextern.set_extern_reference
-  (fun ?loc _ r -> CAst.make ?loc @@ Libnames.Qualid (Nametab.shortest_qualid_of_global Id.Set.empty r));;
+  (fun ?loc _ r -> Nametab.shortest_qualid_of_global ?loc Id.Set.empty r);;
 
 let go () = Coqloop.(loop ~opts:Option.(get !drop_args) ~state:Option.(get !drop_last_doc))
 
--- a/dev/ci/user-overlays/07797-rm-reference.sh
+++ b/dev/ci/user-overlays/07797-rm-reference.sh
@@ -1,0 +1,20 @@
+_OVERLAY_BRANCH=rm-reference
+
+if [ "$CI_PULL_REQUEST" = "7797" ] || [ "$CI_BRANCH" = "_OVERLAY_BRANCH" ]; then
+
+    Equations_CI_BRANCH="$_OVERLAY_BRANCH"
+    Equations_CI_GITURL=https://github.com/maximedenes/Coq-Equations.git
+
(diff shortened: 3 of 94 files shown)


library/libnames.mli
```diff
@@ -65,23 +65,28 @@ val restrict_path : int -> full_path -> full_path
     qualifications of absolute names, including single identifiers.
     The [qualid] are used to access the name table. *)
 
-type qualid
+type qualid_r
+type qualid = qualid_r CAst.t
```
SkySkimmer: Why do we now expose this?

maximedenes: We don't expose more, `qualid_r` is the old `qualid` and the new `qualid` was used in ad-hoc ways in many places.

herbelin (commented):
At some time, the parser was building `Qualid` only with non-empty path (and probably it is still the case, at least in `g_prim.ml4`), so in practice, even if not enforced by the OCaml syntax, I would suspect that the invariant was satisfied and that no collision even happened.

On the other side, if I remember correctly, the only real interest in differentiating between `Ident` in `Qualid` was for `Constrintern` (`intern_applied_reference`, `find_pattern_variable`, ...). The test `qualid_is_ident` is as good with this respect.

If I understand correctly, you are keeping the difference between the concept `reference` (in the sense of ast-level version of `GlobRef.t`/variable) and the technical incarnation of this concept (an uninterpreted list of identifiers). Seems good.

So, altogether, I find that this PR makes sense (with the usual reserve of the opinion of plugin developers, but that's routine!).

gares (approved):
Correct wrt ssr & ide.

If possible I'd made `qualid` a module and have `Qualid.is_ident` and `Qualid.basename` for example. But I was born with C, so I'm also fine with `qualid_` ;-)

maximedenes:
@silene you're the owner of `Libnames`. Do you approve this PR?

maximedenes:
In fact, @gares is the secondary owner and already approved it. So the only remaining step is to provide overlays for all plugins.

gares:
I've only checked ssr and ide as my review says. Which other files do you want me to check?

maximedenes:
I think the idea if that the owner of `Libnames` should approve the principle of the change. All the rest is just propagation.

maximedenes:
@ppedrot In Ltac2, I read:

```coq
(** Beware, at the raw level, [Qualid [id]] has not the same meaning as
    [Ident id]. The first is an unqualified global reference, the second is
    the dynamic reference to id. *)
```

which seems completely contradictory with the whole point of this PR. Can you explain a bit more?

ppedrot:
@maximedenes That's just Ltac2 wrongly exploiting the difference between both representations. I need to fix that separately.

maximedenes:
> That's just Ltac2 wrongly exploiting the difference between both representations. I need to fix that separately.

Ok, so I will submit a first blind patch that makes it compile, and will let you fix the semantics.

ejgallego:
Can't you provide a compatibility layer for plugins?

maximedenes:
> Can't you provide a compatibility layer for plugins?

For now, I prefer to see errors on plugins that I plan to port myself, because there is already an ocean of deprecation warnings produced by plugins we track on CI.

But once that is done, I would be happy to implement a layer. Do you have a suggestion on how to do that?

ejgallego:
> But once that is done, I would be happy to implement a layer. Do you have a suggestion on how to do that?

Well it depends how many function signatures you have changed; the layer is just to keep the old type along with the old functions around, but in deprecated form. Maybe is too much of a nuisance, I dunno; but I do find it valuable; removing plugin warnings is usually very easy.

By the way, if plugins are using deprecated stuff don't hesitate to submit a bug report so they fix it.

maximedenes:
> Well it depends how many function signatures you have changed; the layer is just to keep the old type along with the old functions around, but in deprecated form. Maybe is too much of a nuisance, I dunno; but I do find it valuable; removing plugin warnings is usually very easy.

I agree, but I'm not sure how to do that on this concrete example, because the constructors of the type that disappears were used directly. I can't replace them by functions.

maximedenes:
> The type has to stay of course, in deprecated form.

Ok, but almost all plugins were expecting to find this type in `contrexpr` nodes. So how to preserve compatibility?

ejgallego:
Yeah that seems like too much work.

ejgallego:
You can maybe make `reference = qualid` and then deprecate it. That won't be 100% but at least should give some warning? Dunno if that's worth it.

ejgallego:
I am also not sure I am a fan of the new naming, as `qualid` now also stands for unqualified `id`s. This is confusing IMHO.

maximedenes:
> I am also not sure I am a fan of the new naming, as qualid now also stands for unqualified ids. This is confusing IMHO.

I think it is pretty standard that qualified identifiers allow empty prefixes, isn't it?

ejgallego:
> I think it is pretty standard that qualified identifiers allow empty prefixes, isn't it?

Not in Coq I think.

ppedrot:
@maximedenes Ltac2 made not to observe this difference.

maximedenes:
Ok, I ported all plugins. A side remark: having more uniform build systems for all those plugins wouldn't hurt.

I hope the naming issue is not blocking, as far as I'm concerned, this PR is ready modulo CI.

Zimmi48:
Indeed, in the refman the notion of qualid is a superset for the notion of ident: https://coq.inria.fr/refman/language/gallina-specification-language.html#grammar-token-qualid

ejgallego:
> I don't think it is the case.

Oh indeed you are right, `shortest_qualid_of_foo` will return an empty prefix. Sorry for the noise.

gares:
@maximedenes can you ping the upstreams you made an overlay, I've to leave the lab now

maximedenes:
> can you ping the upstreams you made an overlay, I've to leave the lab now

Sure, thanks for merging! Maybe I don't need to ping you, though ;)

herbelin:
Hi, I noted a regression introduced by this PR. At the time of printing, the information of whether we have a variable or a global reference seems to be lost. This is visible for instance by the loss of color when printing a constant or inductive or constructor.

Example:
```
Check nat.
(* formerly: nat is displayed in light green in coqop and navy in coqide *)
(* now: no color *)
```
I don't know what would be the easier to fix it. At parsing time, we are not able to make the difference, but starting with the translation `constr_expr -> glob_constr` we should be able to tell the difference. So, there is here an asymmetry between parsing and printing which is delicate to deal with.

maximedenes:
Hi, I'd like to understand better. How was the distinction between variables and global references done previously?

herbelin:
I think it was `Ident id` vs `Qualid id` in `Ppconstr.pr_reference`.

herbelin:
@maximedenes: Maybe a cool way to fix the problem while at the same time making `constr_expr` more expressive is to enrich the `cAst.t` record. I don't know what e.g. @ejgallego has in mind, but maybe trying:
```
type 'a CAst.t = {
  v   : 'a;
  loc : Loc.t option;
  alt : 'a option
}
```
and in the case of reference, there could be a `alt` node carrying the full qualified path. (Then, we would have a choice to make for section variables, whether to qualify them or not.)

An alternative fix would be to add a field `CVar` to `constr_expr` which would make `constr_expr` a bit closer to `glob_constr`.

What do you think?

ejgallego:
> and in the case of reference, there could be a `alt` node carrying the full qualified path. (Then, we would have a choice to make for section variables, whether to qualify them or not.)

And what would that field mean in all other constructors of the AST?

herbelin:
> And what would that field mean in all other constructors of the AST?

It is an optional field, to give a possible alternative rendering (to be then propogated to `Pp.t` or to whatever outputting format). (I'm not saying that this is _the_ solution, just trying to find one.)

PR #7801, opened 2018-06-13
[vernac] Add option to force building really mutual induction schemes

mattam82:
Currently, if one of the inductives is non recursive, build_mutual_induction_scheme defaults to a
case analysis schems taking fewer predicates and methods just for that
inductive. This irregularity prevents doing a combined scheme afterwards
to gather all eliminators into one, as combined scheme expects all the
eliminators to have the same predicates and methods. I have a use case
in building function graphs in Equations where some of the inductives
might not be recursive but I expect many other use cases could exist. It's borderline a bug, at least I didn't expect the command to work like this.

Shall we provide user syntax for this?

**Kind:** bug fix / feature

### DIFF
--- a/pretyping/indrec.ml
+++ b/pretyping/indrec.ml
@@ -304,7 +304,7 @@ let make_rec_branch_arg env sigma (nparrec,fvect,decF) f cstr recargs =
   process_constr env 0 f (List.rev cstr.cs_args, recargs)
 
 (* Main function *)
-let mis_make_indrec env sigma listdepkind mib u =
+let mis_make_indrec env sigma ?(force_mutual=false) listdepkind mib u =
   let nparams = mib.mind_nparams in
   let nparrec = mib.mind_nparams_rec in
   let evdref = ref sigma in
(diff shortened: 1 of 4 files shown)


mattam82:
Ideally I'd like this in 8.8.1 or .2 if there is one but it can also wait.

Zimmi48:
If you add a user syntax, this clearly becomes 8.9 material. Whereas if you don't this is indeed borderline and could be acceptable as part of a patch-level release... Maybe you can propose a syntax in another PR on top of this one?

gares (approved):
Looks fine to me. 

IMHO this can go in 8.8.1 since the API is not broken.

mattam82:
Yep, I'd prefer to have the backward compatible API in 8.8.1 and think of user syntax later. In that case no CHANGES/doc needed. ci-vst's error seems spurious so this should actually be ready to merge.

Zimmi48:
@gares You're the main (secondary) code owner here, can you self-assign?

gares:
ci failure is spurious

PR #7802, opened 2018-06-13
make clean: remove g_*.ml files

SkySkimmer:
People keep running into issues with these since the one PR which moved them. This isn't a very general solution but it's better than nothing.

Since our CI cleans before building it will detect if someone adds a non-generated file named g_foo.ml.

NB: g_*.ml is already in our gitignore.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,6 +183,7 @@ indepclean:
 	rm -f $(GENFILES)
 	rm -f $(COQTOPBYTE) $(CHICKENBYTE) $(TOPBYTE)
 	find . \( -name '*~' -o -name '*.cm[ioat]' -o -name '*.cmti' \) -delete
+	find . -name 'g_*.ml' -delete
 	rm -f */*.pp[iox] plugins/*/*.pp[iox]
 	rm -rf $(SOURCEDOCDIR)
 	rm -f toplevel/mltop.byteml toplevel/mltop.optml


SkySkimmer:
@gares can you take this since @letouzey is inactive?

gares:
Given the extension (also belonging to real files) I prefer the pattern to be based on first finding .ml4 files and then removing the corresponding ml one. You could try something like this (but I think we already have  a list of ML4 files around):
```make
rm -f (addsuffix .ml,$(basename $(wildcard */g_*.ml4 plugins/*/g_*.ml4)))
```

Zimmi48:
@gares This is supposed to work around a problem that many users have encountered after they did `git pull` and a `.ml4` file had moved. So your solution won't work.

SkySkimmer:
I think @gares means we can delete any ml file which has the name of a ml4 file.

gares:
Make clean should remove what was generated by the Makefile, if you `make` then `git pull` then `git clean` I don't expect it to clean renamed files.

What I don't like about this patch is that `make clean` may remove `.ml` files that are not generated (by the old or current Makefile). Again, it is not a big deal for `.cmx`, say, but `.ml` if what I produce as a human.

I know `git clean -dfx` is dangerous, so I don't run it in the middle of my dev (before a commit).
I don't think `make clean` should be dangerous.

I've just checked the makefile, just use `rm -f $(ML4FILES:.ml4=.ml)`.

SkySkimmer:
>Make clean should remove what was generated by the Makefile, if you make then git pull then git clean I don't expect it to clean renamed files.

The goal is to clean files that were generated by files which are now renamed.

gares:
Then put `git clean -dfx` in your CI.

gares (changes requested):
use git clean

SkySkimmer:
It's not for CI, it's for users who pull and get a broken working directory.

gares:
I don't like this change.
The generated files are confused with the non-generated ones, this is the problem,
otherwise alienclean could deal with this.

I think we should rename the `.ml` (generated) in `.mlwhatever`.
Then I'd be fine with a `find ... -delete` since `.mlwhatever` would be *always* safe to delete.

gares (changes requested):
To safely detect which files come from renamed `.ml4` files, make the extension of pre-processed `.ml4` files *different* from `.ml`, eg `.mlexpanded`

gares:
One not knowing this convention may call a file `g_something.ml`.
He may not call it `something.mlexpanded`.
IMHO, they are fundamentally different.

ejgallego:
> make the extension of pre-processed .ml4 files different from .ml, eg .mlexpanded

Please _don't_, if your build system is a mess don't make the life of saner people painful, using non-standard suffixes and breaking editor support among other things.

gares:
Generated files don't even need to exist in the first place. This is how CoqMakefile deals with .ml4.
In the Makefile of Coq the choice to save on disk the intermediate data (mashalled AST) with extension .ml.

These .ml file are not .ml files, they are generated and your editor is not letting you edit them.
If you use READABLE_ML4 then these .ml become readable (a debugging feature).

In Dune these problems don't exist, and `git clean dfx` can be replaced by `rm -rf _build`.
I don't see how cleaning up the mess above would make the life of others harder.

maximedenes:
The real fix for this pb is a hygienic build.

PR #7803, opened 2018-06-13
[TYPO FIX] elimitate -> eliminate

bollu:
**Kind:** typo fix.


herbelin (approved):
Of course! Thanks. Shall merge tomorrow.

herbelin:
@Zimmi48: I don't know what label to put here (fix, cleanup, documentation ?), leaving empty. As for the milestone, there is no user impact so I put 8.9 but change it to 8.8.1 if you think it is worth.

Zimmi48:
I often put fix and documentation together and I agree it's not worth to backport.

PR #7805, opened 2018-06-13
Towards listing the critical bugs of the history of Coq.

### LINKED ISSUE #7604: Post-mortem description of kernel soundness bugs
Following a proposition made in PR #7521, I think that every kernel soundness bug should give rise to a post-mortem describing the root of the problem, the consequences and the fix that has been implemented. This is on par with what is performed the domain of security breaches.

herbelin:
(Edited to keep the summary up-to-date)

This is a very first try at listing the critical bugs found in Coq since version 7.0 (2001).

Caution: _It has not been double-checked. It may contain errors or be missing bugs._

Without counting the 3 pending critical bugs, Enrico and I collected 18 bugs (4 about `match`, `fix`, `cofix`, 4 about universes out of which 2 with modules, 2 about prim. proj., 5 with `lazy`, `vm_compute` or `native_compute`, 3 conflicts with the "intended models"). So, altogether, the current count should be 21.

In particular, I let open questions regarding fixes made to the kernel but which I could not decide easily from the description or patch if they were critical/exploitable (out of which the 2 extra critical bugs later noticed by Enrico).

I skipped what seemed to have been bugs in beta-releases.

Need double-checks, decision about the format, etc. In particular, I put quite a lot of details, which I agree are not necessary for a quick reading. On the other side, if someone has more information on the "essence" of the bugs, that would certainly be useful. 

**Kind:** documentation

Closes #7604

There a lot of `?` in the text. Don't hesitate to update and push to this PR.

Extra note: the last two reported items were mistakenly missing in this PR. See #7917 for the missing part.

### DIFF
--- a/dev/doc/critical-bugs
+++ b/dev/doc/critical-bugs
@@ -1,0 +1,226 @@
+Preliminary compilation of critical bugs in stable releases of Coq
+==================================================================
+  WORK IN PROGRESS WITH SEVERAL OPEN QUESTIONS
+
+
+To add: #7723 (vm_compute universe polymorphism), #7695 (modules and algebraic universes), #7615 (lost functor substitutions)
+
+Typing constructions
+
+  component: "match"
+  summary: substitution missing in the body of a let
+  introduced: ?
+  impacted released versions: V8.3-V8.3pl2, V8.4-V8.4pl4
+  impacted development branches: none
+  impacted coqchk versions: ?
+  fixed in: master/trunk/v8.5 (e583a79b5, 22 Nov 2015, Herbelin), v8.4 (525056f1, 22 Nov 2015, Herbelin), v8.3 (4bed0289, 22 Nov 2015, Herbelin)
+  found by: Herbelin
+  exploit: test-suite/success/Case22.v
+  GH issue number: ?
+  risk: ?
+
+  component: fixpoint, guard
+  summary: missing lift in checking guard
+  introduced: probably from V5.10
+  impacted released versions: probably V5-V7, V8.0-V8.0pl4, V8.1-V8.1pl4
+  impacted development branches: v8.0 ?
+  impacted coqchk versions: ?
+  fixed in: master/trunk/v8.2 (ff45afa8, r11646, 2 Dec 2008, Barras), v8.1 (f8e7f273, r11648, 2 Dec 2008, Barras)
+  found by: Barras
+  exploit: test-suite/failure/guard.v
+  GH issue number: none
+  risk: unprobable by chance
+
+  component: cofixpoint, guard
+  summary: de Bruijn indice bug in checking guard of nested cofixpoints
+  introduced: after V6.3.1, before V7.0
+  impacted released versions: V8.0-V8.0pl4, V8.1-V8.1pl4, V8.2-V8.2pl2, V8.3-V8.3pl2, V8.4-V8.4pl4
+  impacted development branches: none
+  impacted coqchk versions: ?
+  fixed in: master (9f81e2c36, 10 Apr 2014, Dénès), v8.4 (f50ec9e7d, 11 Apr 2014, Dénès), v8.3 (40c0fe7f4, 11 Apr 2014, Dénès), v8.2 (06d66df8c, 11 Apr 2014, Dénès), v8.1 (977afae90, 11 Apr 2014, Dénès), v8.0 (f1d632992, 29 Nov 2015, Herbelin, backport)
+  found by: Dénès
+  exploit: ?
+  GH issue number: none ?
+  risk: ?
+
+  component: inductive types, elimination principle
+  summary: de Bruijn indice bug in computing allowed elimination principle
+  introduced: 23 May 2006, 9c2d70b, r8845, Herbelin (part of universe polymorphism)
+  impacted released versions: V8.1-V8.1pl4, V8.2-V8.2pl2, V8.3-V8.3pl2, V8.4-V8.4pl4
+  impacted development branches: none
+  impacted coqchk versions: ?
+  fixed in: master (8a01c3685, 24 Jan 2014, Dénès), v8.4 (8a01c3685, 25 Feb 2014, Dénès), v8.3 (2b3cc4f85, 25 Feb 2014, Dénès), v8.2 (459888488, 25 Feb 2014, Dénès), v8.1 (79aa20872, 25 Feb 2014, Dénès)
+  found by: Dénès
+  exploit: see GH#3211
+  GH issue number: #3211
+  risk: ?
+
+  component: universe subtyping
+  summary: bug in Prop<=Set conversion which made Set identifiable with Prop, preventing a proof-irrelevant interpretation of Prop
+  introduced: V8.2 (bba897d5f, 12 May 2008, Herbelin)
+  impacted released versions: V8.2-V8.2pl2
+  impacted development branches: none
+  impacted coqchk versions: ?
+  fixed in: master/trunk (679801, r13450, 23 Sep 2010, Glondu), v8.3 (309a53f2, r13449, 22 Sep 2010, Glondu), v8.2 (41ea5f08, r14263, 6 Jul 2011, Herbelin, backport)
+  found by: Georgi Guninski
+  exploit: test-suite/bugs/closed/4294.v
+  GH issue number: #4294
+  risk: ?
+
+Module system
+
+  component: modules, universes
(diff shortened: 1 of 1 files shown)


dev/doc/critical-bugs
```diff
@@ -0,0 +1,221 @@
...
+  GH issue number: #4157
+  risk: 
+
+  component: "virtual machine" (compilation to bytecode ran by a C-interpreter)
+  summary: wrong universe constraints
+  introduced: possibly exploitable from V8.1; exploitable at least from V8.5
+  impacted released versions: V8.1-V8.4pl5 unknown, V8.5-V8.5pl3, V8.6-V8.6.1, V8.7.0-V8.7.1
+  impacted development branches: unknown for v8.1-v8.4, none from v8.5
+  impacted coqchk versions: none (no virtual machine in coqchk)
+  fixed in: master (c9f3a6cbe5c41, 12 Feb 2018, PR#6713, Dénès), v8.7 (c058a4182, 15 Feb 2018, Zimmermann), v8.6 (a2cc54c6, 21 Feb 2018, Herbelin), v8.5 (d4d550d0f1a, 21 Feb 2018, Herbelin)
```
Zimmi48: I'm just the committer for this commit. The listed author is Maxime Dénès.

herbelin: I clarified your role by adding the mention "backport" (here and everywhere else relevant). That'd be ok?

Zimmi48: Yes

JasonGross:
Could we make this a `.md` file, and have the "exploit" be a link to the actual file in the test-suite?
(Something like (untested) `sed s',exploit: \([^\.]*\.v\)$,exploit: [\1](../../\1),g' -i`)

dev/doc/critical-bugs
```diff
@@ -0,0 +1,221 @@
...
+
+  component: "native" conversion machine (translation to OCaml which compiles to native code)
+  summary: translation of identifier from Coq to OCaml was not bijective, leading to identify True and False
+  introduced: V8.5
+  impacted released versions: V8.5-V8.5pl1
+  impacted development branches: none
+  impacted coqchk versions: none (no native computation in coqchk)
+  fixed in: master/trunk/v8.6 (a91d15dcb4, 19 May 2016, letouzey), v8.5 (088b3161c93, 19 May 2016, letouzey), 
+  found by: Letouzey, Dénès
+  exploit: lost?
```
JasonGross: I think this one might actually be 244d7a9aafe7ad613dd2095ca3126560cb3ea1d0, which includes the exploit in the commit message?

> `Unicode.ascii_of_ident` is now truly injective
>
> A non-ASCII char is now converted to `_UUxxxx_` with xxxx being its unicode index
 in hexa. And any preexisting `_UU` substring in the ident is converted to `_UUU`.
 The switch from `__Uxxxx_` to `_UUxxxx_` is cosmetic, it just helps the extraction
 (less __ in names). But the other part of the patch (detection of preexisting
 `_UU` substrings) is critical to make `ascii_of_ident` truly injective and avoid
 the following kind of proof of False via native_compute :
>
>      Definition α := 1.
>      Definition __U03b1_ := 2.
>      Lemma oups : False.
>      Proof.
>       assert (α = __U03b1_). { native_compute. reflexivity. }
>       discriminate.
>      Qed.

I think the commit you're pointing at (a91d15dcb4) is just cosmetic?

JasonGross (changes requested):

dev/doc/critical-bugs
```diff
@@ -0,0 +1,221 @@
...
+  exploit: need to find the example again
+  GH issue number: no
+  risk: unlikely to be exploited by chance
+
+  component: library of extensional sets, guard condition
+  summary: guard condition was unknown to be inconsistent with propositional extensionality in library Sets
+  introduced: not a bug per se but an incompatibility discovered late
+  impacted released versions: technically speaking from V6.1 with the introduction of the Sets library which was then inconsistent from the very beginning without we knew it
+  impacted coqchk versions: ?
+  fixed by constraining the guard condition: bac7079739, 28 Oct 2004, Barras, Dénès
```
JasonGross: bac7079739 is 
> Set devient predicatif par defaut
>
> git-svn-id: svn+ssh://scm.gforge.inria.fr/svn/coq/trunk@4726 85f007b7-540e-0410-9357-904b9bb8a0f7

by @herbelin committed on Oct 28, 2003...

Also, is this the one whose proof is
```coq
Axiom ext : (False -> False) = True.
Fixpoint oops (x : True) : False
  := match x with
     | I => oops (match ext in (_ = T) return T with
                  | eq_refl => fun absurd => match absurd with end
                  end)
     end.
```
which is still present as late as 8.4pl6?

JasonGross (dismissed):

dev/doc/critical-bugs
herbelin: Bad copy-paste. I changed for a more relevant commit, but I'm not sure it was fixed by only  one commit and if one commit exactly this one. I'm unaware of the exact history and if @maximedenes could give the needed clarifications and fill the missing bits, that would be wonderful.

dev/doc/critical-bugs
herbelin: Right. Fixed. Thanks.

herbelin:
> Could we make this a .md file, and have the "exploit" be a link to the actual file in the test-suite?
(Something like (untested) sed s',exploit: \([^\.]*\.v\)$,exploit: [\1](../../\1),g' -i)

We can certainly eventually do that. As a first step, my idea was to help with the historical digging, to get confirmation by the authors that these was critical bugs. I dont' know what was the idea of @ppedrot at the beginning when he opened #7604.

herbelin:
> @herbelin If you think that you want to stop there, we can also consider merging as it is right now and let other people like @JasonGross open new PRs to propose a different format.

I shall not really have time to work more on it now. So, your proposal is ok to me if it is ok for others.

maximedenes:
I like the idea of tracking critical bugs, and use it as a way of showing we are taking them seriously. It would be good to add a number to identify each of the listed soundness bug, in the spirit of CVE. Maybe like CSB-2018-XXX (CSB = Coq Soundness Bug).

We should also document what needs to be done when a new soundness bug is discovered.

herbelin:
Ideally, someone with "physical access" to @barras should ask him if among the various bugs he fixed in the kernel along the years, there are still some not in my list which were critical.

Zimmi48:
OK then I shall merge this in the beginning of next week. Anyone should feel free to push commits on top of this branch in the meantime. Also, maybe it would make sense to add an entry in CHANGES to mention the existence of this new document?

gares:
Dear Hugo, I think this is also a critical bug (that I don't find in the list). This is a commit fixing it in 8.2 c1d988904483eb1f3a8917ea08fced1240e3844b IIRC it was not there in 8.1. I think I did fix it also in 8.3, 8.4 and 8.5

gares:
component: modules
summary: universe constraints for module subtyping not stored in vo files
introduced: 8.2 (b3d3b566c5b)
impacted released versions: 8.2, 8.3, 8.4
impacted development branches: v8.5
impacted coqchk versions: none
fixed in: v8.2 (c1d9889), v8.3 (8056d02), v8.4 (a07deb4ea), trunk (0cd0a3ecdc7) Mar 5, 2014, Tassi
found by: Tassi by running coqchk on the mathematical components library
exploit: requires multiple files, no test provided
GH issue number: #3243
risk: could be exploited by mistake

gares:
I'm not 100% sure about the commit that introduced it (I'm not an expert of this code).

gares:
I also don't see the bug fixed with this commit a8b034513e0c03ceb7e154

herbelin:
@gares: I added #3243. Thanks so much.

About Georges' bug in the lazy machine, there was indeed a 6-line exploit. I added it an entry for it. Thanks also very much.

Zimmi48 (approved):
I'll merge this now and I'll let whoever wants to change the format to a Markdown file do it in a subsequent PR.

gares:
> I added it an entry for it.

Sorry but I don't see that

gares:
@Zimmi48

gares:
I think @herbelin can just open another PR with the extra commits by rebasing his `master+history-critical-bugs` on today's master.

herbelin:
Explanation: I had renamed the branch and pushed to the wrong name. Sorry about that. See #7917 for the two extra items.

PR #7815, opened 2018-06-15
On cygwin, pass the filename in a format that coqdoc understands.

jfehrle:

"make sphinx" failed on Windows/cygwin because Python passed a filename to coqdoc in the form "/cygdrive/c/...", which coqdoc doesn't understand.  Coqdoc does understand "C:/dir/...".

### DIFF
--- a/doc/tools/coqrst/coqdoc/main.py
+++ b/doc/tools/coqrst/coqdoc/main.py
@@ -20,6 +20,7 @@ lexer.
 """
 
 import os
+import platform
 from tempfile import mkstemp
 from subprocess import check_output
 
@@ -36,6 +37,9 @@ def coqdoc(coq_code, coqdoc_bin=None):
     """Get the output of coqdoc on coq_code."""
     coqdoc_bin = coqdoc_bin or os.path.join(os.getenv("COQBIN"), "coqdoc")
     fd, filename = mkstemp(prefix="coqdoc-", suffix=".v")
+    if platform.system().startswith("CYGWIN"):
+        # coqdoc currently doesn't accept cygwin style paths in the form "/cygdrive/c/..."
+        filename = check_output(["cygpath", "-w", filename]).decode("utf-8").strip()
     try:
         os.write(fd, COQDOC_HEADER.encode("utf-8"))
         os.write(fd, coq_code.encode("utf-8"))


doc/tools/coqrst/coqdoc/main.py
```diff
@@ -32,10 +33,17 @@
 COQDOC_SYMBOLS = ["->", "<-", "<->", "=>", "<=", ">=", "<>", "~", "/\\", "\\/", "|-", "*", "forall", "exists"]
 COQDOC_HEADER = "".join("(** remove printing {} *)".format(s) for s in COQDOC_SYMBOLS)
 
+def tempfile():
+    fd, filename = mkstemp(prefix="coqdoc-", suffix=".v")
+    if platform.system().startswith("CYGWIN"):
+        # coqdoc doesn't accept cygwin style paths in the form "/cygdrive/c/..."
+        filename = check_output(["cygpath", "-w", filename]).strip().decode("utf-8")
```
cpitclaudel: I'd flip the strip and the decode.  Also, you want to encode the filename before passing it, I think.

doc/tools/coqrst/coqdoc/main.py
```diff
@@ -32,10 +33,17 @@
...
+    return fd, filename
+
 def coqdoc(coq_code, coqdoc_bin=None):
     """Get the output of coqdoc on coq_code."""
     coqdoc_bin = coqdoc_bin or os.path.join(os.getenv("COQBIN"), "coqdoc")
-    fd, filename = mkstemp(prefix="coqdoc-", suffix=".v")
+    fd, filename = tempfile()
```
cpitclaudel: I think tempfile should take the same arguments as mkstemp.

cpitclaudel (commented):
I think the code is fine (I added suggestions), but I wonder why it's needed.
Is it because you're compiling coqdoc with a non-cygwin OCaml? IOW, shouldn't coqdoc be the one who understands these paths?
It seems strange to build cygwin-specific knowledge into this python script.
You might also be able to work around the issue by using relative paths, btw.  Does it work to take the path to the tempfile relative to the current folder and pass that to coqdoc? This could be a good workaround.

doc/tools/coqrst/coqdoc/main.py
jfehrle: Whitespace characters won't be changed by the decode, and they can't be part of a miultityte character.  The encode would be needed if the path to the coq source tree has utf-8 characters; so fat I think we haven't used them in any file/directory names in the source tree.

doc/tools/coqrst/coqdoc/main.py
jfehrle: Or just inline the code instead of having a separate function--my original tempfile() was more complex.

jfehrle:
> I wonder why it's needed.  Is it because you're compiling coqdoc with a non-cygwin OCaml? IOW, shouldn't coqdoc be the one who understands these paths?

I wondered this myself.  Not so easy to investigate--probably not well-documented in cygwin/OCaml and probably requires much experimentation re-installing of tools and full rebuilds of coq. 

My guess at the moment is that `4.02.3+mingw64c` doesn't include the cygdrive support.  I'll try that tomorrow.

```
$ opam list
# Installed packages for 4.02.3+mingw64c:
base-bigarray      base  Bigarray library distributed with the OCaml compiler
base-bytes         base  Bytes library distributed with the OCaml compiler
base-ocamlbuild    base  OCamlbuild binary and libraries distributed with the OC
base-threads       base  Threads library distributed with the OCaml compiler
base-unix          base  Unix library distributed with the OCaml compiler
camlp4           4.02+7  Camlp4 is a system for writing extensible parsers for p
camlp5             7.05  Preprocessor-pretty-printer of OCaml
conf-m4               1  Virtual package relying on m4
conf-which            1  Virtual package relying on which
ocamlbuild            0  Build system distributed with the OCaml compiler since
ocamlfind         1.8.0  A library manager for OCaml
ounit             2.0.8  Unit testing framework loosely based on HUnit. It is si
```

jfehrle:
Looks like the official release version of Coq for Windows doesn't support cygdrive path names, either.  But I expect changing the Windows build process to fix this is not at all simple--the current process applies several patches and runs from a .bat script.  This PR seems to me a reasonable quick fix.

```
$ echo $X
/cygdrive/c/Users/Jim.DESKTOP-MMPNVGH/Documents/coqSource/coq/test3.v

$ echo $Y
C:/Users/Jim.DESKTOP-MMPNVGH/Documents/coqSource/coq/test3.v

$ ls $X
/cygdrive/c/Users/Jim.DESKTOP-MMPNVGH/Documents/coqSource/coq/test3.v

$ ls $Y
C:/Users/Jim.DESKTOP-MMPNVGH/Documents/coqSource/coq/test3.v

$ coqtop -v
The Coq Proof Assistant, version 8.6.1 (July 2017)
compiled on Jul 6 2017 10:24:59 with OCaml 4.02.3

$ coqtop -compile $X
Error:
Can't find file
/cygdrive/c/Users/Jim.DESKTOP-MMPNVGH/Documents/coqSource/coq/test3.v

$ coqtop -compile $Y

$   (i.e., success)

```

doc/tools/coqrst/coqdoc/main.py
cpitclaudel: > Whitespace characters won't be changed by the decode, and they can't be part of a miultityte character.

OK, that's fair :) I like to not reason about these things, but the argument is convincing.

> so fat I think we haven't used them in any file/directory names in the source tree.

But it's not a source tree file, right? It's wherever mkstemp creates it.  It's OK, though — python3 takes a string for that argument, so I think we're good.

cpitclaudel:
OK, I'm fine with this workaround then.  Thanks for looking into this.

jfehrle:
I will push a modified version addressing your concerns within a day or so.

jfehrle:
I made the tweaks we discussed.  @cpitclaudel, would you mark this as reviewed?  Could you also do the merge?  Thanks a bunch for your help!  Looking forward to revising some documentation...

Zimmi48:
Right now @cpitclaudel don't have the push rights to do the merge. So it's either @maximedenes who merges this, or @cpitclaudel after I give him push rights, if he wants to be an owner.

cpitclaudel (approved):
LGTM. Do we need the final encode() call, though?  I think you previously pointed out that check_output takes strings.

cpitclaudel:
I'm a bit weary of taking on additional maintenance responsibilities, and I'm not sure how much responsiveness I can commit to.  As a compromise, what do you think of having me as secondary owner for that code?

jfehrle:
> Do we need the final encode() call, though? I think you previously pointed out that check_output takes strings.

I removed it.  "make sphinx" seemed to work correctly with the encode in place, though.

jfehrle:
The gitlab failures are unrelated to this change, which only modifyies a isingle python file.  So I think this is ready to merge...

maximedenes (approved):
Unrelated failures indeed, thanks!

jfehrle:
Thanks @maximedenes !

PR #7817, opened 2018-06-15
Clarifying that the "print" functions in Names are mainly for debugging

herbelin:
**Kind:** cleanup

The `print` and `to_string` functions in `Names` were printing full names (not taking into account visibility). They were also printing `#` for sections, which is not part of the user-level syntax of names.

We requalify them as "debug" printers. In passing, this allows to catch quite a number of misuses of these functions in code supposed to be read by the user (Added: submitted separately in #8534 on which this PR now depends).

Use of these printers in reporting anomalies are expected and were kept.

~~Still not perfect (see a few FIXME).~~

This is a step in helping external contributors and plugin developer not to use the wrong printing functions. We shall still have to tell where are the good functions (namely in `Printer`, or for code below `Printer`, for lack of anything better, using `Termops.print_constr`).

The design choices can be discussed. This is just a proposal.

- [ ] Entry to add in `dev/doc/changes.md`

### DIFF
--- a/checker/environ.ml
+++ b/checker/environ.ml
@@ -118,7 +118,7 @@ let anomaly s = anomaly (Pp.str s)
 let add_constant kn cs env =
   if Cmap_env.mem kn env.env_globals.env_constants then
     Printf.ksprintf anomaly ("Constant %s is already defined.")
-      (Constant.to_string kn);
+      (Constant.debug_to_string kn);
   let new_constants =
     Cmap_env.add kn cs env.env_globals.env_constants in
   let new_globals =
@@ -184,7 +184,7 @@ let lookup_mind kn env =
 let add_mind kn mib env =
   if Mindmap_env.mem kn env.env_globals.env_inductives then
     Printf.ksprintf anomaly ("Mutual inductive block %s is already defined.")
-      (MutInd.to_string kn);
+      (MutInd.debug_to_string kn);
   let new_inds = Mindmap_env.add kn mib env.env_globals.env_inductives in
   let kn1,kn2 =  MutInd.user kn, MutInd.canonical kn in
   let new_inds_eq = if KerName.equal kn1 kn2 then
--- a/kernel/mod_subst.ml
+++ b/kernel/mod_subst.ml
@@ -85,11 +85,11 @@ let is_empty_subst = Umap.is_empty
 let string_of_hint = function
   | Inline (_,Some _) -> "inline(Some _)"
   | Inline _ -> "inline()"
-  | Equiv kn -> KerName.to_string kn
+  | Equiv kn -> KerName.debug_to_string kn
 
 let debug_string_of_delta resolve =
   let kn_to_string kn hint l =
-    (KerName.to_string kn ^ "=>" ^ string_of_hint hint) :: l
+    (KerName.debug_to_string kn ^ "=>" ^ string_of_hint hint) :: l
   in
   let mp_to_string mp mp' l =
     (ModPath.to_string mp ^ "=>" ^ ModPath.to_string mp') :: l
--- a/checker/subtyping.ml
+++ b/checker/subtyping.ml
@@ -114,7 +114,7 @@ let check_inductive  env mp1 l info1 mib2 spec2 subst1 subst2=
         let v = Univ.ACumulativityInfo.variance cumi in
         let v' = Univ.ACumulativityInfo.variance cumi' in
         if not (Array.for_all2 Univ.Variance.check_subtype v' v) then
-          CErrors.anomaly Pp.(str "Variance mismatch for " ++ MutInd.print kn)
+          CErrors.anomaly Pp.(str "Variance mismatch for " ++ MutInd.debug_print kn)
       in
       let auctx = Univ.ACumulativityInfo.univ_context cumi in
       let auctx' = Univ.ACumulativityInfo.univ_context cumi' in
--- a/checker/typeops.ml
+++ b/checker/typeops.ml
@@ -75,7 +75,7 @@ let judge_of_constant env (kn,u as cst) =
   let _cb =
     try lookup_constant kn env
     with Not_found ->
-      failwith ("Cannot find constant: "^Constant.to_string kn)
+      failwith ("Cannot find constant: "^Constant.debug_to_string kn)
   in
   let ty, cu = constant_type env cst in
   let () = check_constraints cu env in
@@ -158,7 +158,7 @@ let judge_of_inductive_knowing_parameters env (ind,u) (paramstyp:constr array) =
   let specif =
     try lookup_mind_specif env ind
     with Not_found ->
-      failwith ("Cannot find mutual inductive block: "^MutInd.to_string (fst ind))
+      failwith ("Cannot find mutual inductive block: "^MutInd.debug_to_string (fst ind))
   in
   type_of_inductive_knowing_parameters env (specif,u) paramstyp
 
@@ -172,7 +172,7 @@ let judge_of_constructor env (c,u) =
   let specif =
     try lookup_mind_specif env ind
     with Not_found ->
-      failwith ("Cannot find mutual inductive block: "^MutInd.to_string (fst ind))
+      failwith ("Cannot find mutual inductive block: "^MutInd.debug_to_string (fst ind))
   in
   type_of_constructor (c,u) specif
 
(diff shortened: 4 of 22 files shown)


ppedrot (dismissed):
You should mark the old functions as deprecated, it makes the transition easier for plugin writers. We will remove them in the next version.

herbelin:
> You should mark the old functions as deprecated, it makes the transition easier for plugin writers. We will remove them in the next version.

Done.

checker/environ.ml
```diff
@@ -120,7 +120,7 @@ let anomaly s = anomaly (Pp.str s)
 let add_constant kn cs env =
   if Cmap_env.mem kn env.env_globals.env_constants then
     Printf.ksprintf anomaly ("Constant %s is already defined.")
-      (Constant.to_string kn);
+      (Constant.debug_to_string kn);
```
maximedenes: This is not really debugging code, it is just error printing. What should it use instead of a debug printer?

kernel/declareops.ml
```diff
@@ -289,9 +289,9 @@ let hcons_mind mib =
 (** {6 Stm machinery } *)
 
 let string_of_side_effect { Entries.eff } = match eff with
```
maximedenes: Should this function also get `debug` prefix?

kernel/mod_subst.ml
```diff
@@ -78,11 +78,11 @@ let is_empty_subst = Umap.is_empty
 let string_of_hint = function
```
maximedenes: Same here, the `debug` prefix should be propagated.

maximedenes (dismissed):
I like the idea of guiding API users better. I'm not completely sure it should be done via function names rather than good documentation, but I'm not opposed either. What I don't like, though, is that kernel error reporting now relies on `debug_XXX` functions, when it is not really debugging code. We can think about it. Meanwhile, can you collect all the fixes done in this PR (as in, the parts that stopped relying on debug printers) in another one, that can be merged immediately? Thanks!

checker/environ.ml
herbelin: It is anomaly printing. I used `debug` as a continuation of existing terminology but we could use `internal`, or `unsafe`, or whatever other terminology.

Not that I'm a big fan of cryptic anomaly message. Users see some of them sometimes and they deserve an effort for understandable anomalies.

If we use the `Constant` printer, it is because we have nothing else at hand.

herbelin:
> I'm not completely sure it should be done via function names rather than good documentation, but I'm not opposed either.

If you'd like to write good documentation instead, which is indeed another possibility, I gladly let you take this PR over.

About splitting the PR, I'm a bit worry on one side by my time which is very limited these weeks and on the other side by that we have some difficulty to _significantly_ go down again the number of about 150 PRs open. So, I prefer not to promise anything related to this suggestion.

kernel/declareops.ml
herbelin: If the `debug` prefix is definitively adopted then ok to generalize it. I did not do it because I was unsure `debug` was really the best terminology.

An other alternative to the `debug` prefix could be `basic`, or something on this base. What do you think?

ejgallego (commented):
I am also not a fan of the `debug` prefix, I dunno.

I'd rather use a module wrapping the "internal" and external parts of the API. Anyways, would we follow something such as CEP#12 we would have a proper interface for object such that the interface would expose the constraints on `print` etc...

Anyways at least for errors we will require error registration soon so that should improve the printing situation a bit.

herbelin:
@ejgallego, @maximedenes: then maybe can we indeed share the work. I make a PR with only the fixes and one of you address the issue of documenting the API?

ejgallego:
> I make a PR with only the fixes

That would be awesome @herbelin ; please do so. I am still wondering what the best choice here would be.

ejgallego:
@herbelin , first sorry for the delay in commenting here; second, I have been thinking a bit more about this and indeed I tend to think that as @maximedenes has proposed, function names may not be the best alternative here. For example, it makes sense that modules defining core ADT provide a `print` function, and whether it is a debug utility or not is not so clear (for example the printer may want to use not for debug purposes).

May I propose we instead take advantage of the API-documentation system, which is now built for every PR even if it needs some love, to better specify what these print functions do "low-lever printer?".

IMHO we should encourage developers to look at the API docs, in fact we should even add some links to current docs to README, etc...

herbelin:
I then created #8534 for the "fix" part. This PR is then only about a renaming.

ejgallego:
> This PR is then only about a renaming.

Great, thanks @herbelin . For the renaming then I feel like I'd prefer the current `print` name to remain and improve the API docs instead. However this is just a soft preference so I wonder what others would think, for example @ppedrot as he introduced quite a few of this modules if I am not mistaken.

herbelin:
> For the renaming then I feel like I'd prefer the current print name to remain and improve the API docs instead.

Note that there are already functions `MBId.debug_to_string`, `ModPath.debug_to_string`, `KerName.debug_to_string`, `Constant.debug_to_string`, `Constant.debug_print`, `MutInd.debug_to_string`, `MutInd.debug_print`, which add one or two bits of information on top of `MBId.to_string`, `ModPath.to_string`, `KerName.to_string`, `Constant.to_string`, `Constant.print`, `MutInd.to_string`, `MutInd.print`, which are also, even without the `debug_` prefix, internal printers (i.e. printing a `#` which has no meaning from the user syntax point of view).

The two bits of extra information are a key for distinguishing module variables with same name, and the canonical kernel name when it differs from the user name.

ppedrot:
@herbelin You should probably rebase now that the sub-PR has been merged in master.

ejgallego:
Thanks @herbelin for the rebase. After reading the patch again my opinion is that I'd rather keep the original names and improve the API documentation instead.

The fact that these printers are "debug-only" only matters if you look at the system from a very high-layers, however, users of `coq.kernel` may not have the same view. Anyways maybe the main technical argument to keep the current naming is that we would expect for such core types in the future to conform to some signature:
```ocaml
module type Base : sig
   type t

   val print : t -> Pp.t
   val equal : t -> -> t -> bool
```
etc... I guess that was the main reason to modularize them in 4490dfcb94057dd6518963a904565e3a4a354bac so we may as well take advantage of that.

herbelin:
> After reading the patch again my opinion is that I'd rather keep the original names and improve the API documentation instead.

As discussed above, I'm ok so that one of you instead improves documentation.

Note that after #8555, the situation is that we don't print any more `#` in the non-debug printers, which is nice, but the printed names are still not "user-level" names, since for them to be user-level names (as of now), the section path would have to be printed.

ejgallego:
@ppedrot what are your thoughts here?

ejgallego:
I think we can safely more this to the 8.10 milestone.

ppedrot:
So what? Following @ejgallego's comments, should we close this PR and open a new documentation one or what?

ppedrot:
@herbelin ping.

herbelin:
> @herbelin ping.

I thought the agreement was that @maximedenes or @ejgallego clarify the `Names` API so that it is clear that the `KerName` printers are not to be exposed to users but only for internal messages such as anomaly.

I can alternatively encapsulate the printers in an `Internal` module, if the `debug` word is too connotated.

ppedrot:
@maximedenes @ejgallego ping.

ejgallego:
> @maximedenes @ejgallego ping.

I don't think I have more to add than what I said in https://github.com/coq/coq/pull/7817#issuecomment-427859928

IMHO the names should remain the same as to advance towards uniform signatures for modules; but of course OMMV. I would improve the API documentation of the Names module in order to help API users, we have now "readable" API docs so this can be done I think.

herbelin:
@ejgallego: I can actually do a change of documentation myself. If I say, next to `Constant.print` and `MutInd.print`: "This printer displays internal kernel names. Messages addressed to users should use instead `Printer.pr_constant`, are you ok?

ejgallego:
> "This printer displays internal kernel names. Messages addressed to users should use instead `Printer.pr_constant`, are you ok?

I am more than OK @herbelin ; note that my comments are just a very soft opinion of course I am OK with any other approach people would like to follow.

herbelin:
But then, there is a little discrepancy with `Names.Id.print` which is ok for users and which has no equivalent in `Printer`...

herbelin:
OK, I opened #9413.

PR #7819, opened 2018-06-15
Ho matching occ sel

mattam82:
This is the big higher-order matching update commit of unifall, which extends the capabilities of evar_conv_x and second_order_matching to be configurable by flags and occurrence selection strategy. It's not called much at this point, only by abstract_list_all_with_dependencies and when calling `solve_unif_constraints_with_heuristics` with an explicit `with_ho:true` argument.

**Kind:**  feature

Mainly the internally used interfaces of evarconv changed and code in pretyping was adapted. The the_conv_x and the_conv_x_leq functions have been deprecated in favor of unify_delay/unify_leq_delay and the complete unify function. The passing of flags changed as well to support parameterizing evar_conv with transparent states, frozen_evars and flags for betaiota reduction and canonical structure resolution.

### DIFF
--- a/engine/evd.mli
+++ b/engine/evd.mli
@@ -77,6 +77,14 @@ sig
 
 end
 
+module Abstraction : sig
+  type t = bool list
+
+  val identity : t
+
+  val abstract_last : t -> t
+end
+
 (** {6 Evar infos} *)
 
 type evar_body =
@@ -94,6 +102,10 @@ type evar_info = {
   (** Boolean mask over {!evar_hyps}. Should have the same length.
       When filtered out, the corresponding variable is not allowed to occur
       in the solution *)
+  evar_abstract_arguments : Abstraction.t;
+  (** Boolean information over {!evar_hyps}, telling if an hypothesis instance
+      can be immitated or should stay abstract in HO unification problems
+      and inversion (see [second_order_matching_with_args] for its use). *)
   evar_source : Evar_kinds.t located;
   (** Information about the evar. *)
   evar_candidates : econstr list option;
--- a/engine/evd.mli
+++ b/engine/evd.mli
@@ -104,7 +104,7 @@ type evar_info = {
       in the solution *)
   evar_abstract_arguments : Abstraction.t;
   (** Boolean information over {!evar_hyps}, telling if an hypothesis instance
-      can be immitated or should stay abstract in HO unification problems
+      can be imitated or should stay abstract in HO unification problems
       and inversion (see [second_order_matching_with_args] for its use). *)
   evar_source : Evar_kinds.t located;
   (** Information about the evar. *)
--- a/engine/evd.mli
+++ b/engine/evd.mli
@@ -78,7 +78,11 @@ sig
 end
 
 module Abstraction : sig
-  type t = bool list
+  type abstraction =
+    | Abstract
+    | Imitate
+
+  type t = abstraction list
 
   val identity : t
 
--- a/pretyping/evarconv.ml
+++ b/pretyping/evarconv.ml
@@ -788,7 +788,15 @@ and evar_eqappr_x ?(rhs_is_already_stuck = false) flags env evd pbty
           else UnifFailure (i,NotSameHead)
         and f3 i = miller true (sp1,al1) appr1 appr2 i
         and f4 i = miller false (sp2,al2) appr2 appr1 i
-        and f5 i = consume true appr1 appr2 i in
+        and f5 i =
+          (** We ensure failure of consuming the stacks does not
+             propagate an error about unification of the stacks while
+             the heads themselves cannot be unified, so we return
+             NotSameHead. *)
+          match consume true appr1 appr2 i with
+          | Success _ as x -> x
+          | UnifFailure _ -> quick_fail i
+        in
         ise_try evd [f1; f2; f3; f4; f5]
 
     | Flexible ev1, MaybeFlexible v2 ->
--- a/pretyping/evarconv.ml
+++ b/pretyping/evarconv.ml
@@ -420,7 +420,10 @@ let compare_cumulative_instances evd variances u u' =
 
 let conv_fun f flags on_types =
   let typefn env evd pbty term1 term2 =
-    f { (default_flags env) with with_cs = flags.with_cs } env evd pbty term1 term2
+    let flags = { (default_flags env) with
+      with_cs = flags.with_cs;
+      frozen_evars = flags.frozen_evars }
+    in f flags env evd pbty term1 term2
   in
   let termfn env evd pbty term1 term2 =
     f flags env evd pbty term1 term2
@@ -1397,50 +1400,45 @@ let second_order_matching flags env_rhs evd (evk,args) (test,argoccs) rhs =
        Feedback.msg_debug Pp.(str"abstracting: " ++
                                 prc env_rhs (mkVar id) ++ spc () ++
                                 prc env_rhs c);
-      let rec force_instantiation evd = function
-      | (evk,evty,inst,abstract)::evs ->
-         let evk = Option.default evk (Evarutil.advance evd evk) in
-         let evd =
-            if is_undefined evd evk then
-              (* We try abstraction or concretisation for *)
-              (* this unconstrained occurrence *)
-              (* and we use typing to propagate this instantiation *)
-              (* We avoid making an arbitrary choice by leaving candidates *)
-              (* if both can work *)
-              let evi = Evd.find_undefined evd evk in
-              let vid = mkVar id in
-              let candidates = [inst; vid] in
-              try
-                let evd, ev = Evarutil.restrict_evar evd evk (Evd.evar_filter evi) (Some candidates) in
-                let evi = Evd.find evd ev in
-                (match evar_candidates evi with
-                 | Some [t] ->
-                    if not (noccur_evar env_rhs evd ev (EConstr.of_constr t)) then
-                      raise (TypingFailed evd);
-                    let evd = Evd.define ev (EConstr.of_constr t) evd in
-                    check_evar_instance evar_unify flags evd ev (EConstr.of_constr t)
-                 | Some l when abstract = Abstraction.Abstract &&
-                     List.exists (fun c -> isVarId evd id (EConstr.of_constr c)) l ->
-                    let evd = Evd.define ev vid evd in
-                    check_evar_instance evar_unify flags evd ev vid
-                 | _ -> evd)
-              with e -> user_err (Pp.str "Cannot find an instance")
-            else
-              ((if !debug_ho_unification then
-                  let evi = Evd.find evd evk in
-                  let env = Evd.evar_env evi in
-                  Feedback.msg_debug Pp.(str"evar is defined: " ++
-                     int (Evar.repr evk) ++ spc () ++
-                     prc env (match evar_body evi with Evar_defined c -> c
-                                                     | Evar_empty -> assert false)));
-               evd)
-          in
-          force_instantiation evd evs
-      | [] ->
-          abstract_free_holes evd l
-      in
-      force_instantiation evd !evsref
-  | [] ->
+     let rec force_instantiation evd = function
+     | (evk,evty,inst,abstract)::evs ->
+       let evk = Option.default evk (Evarutil.advance evd evk) in
+       let evd =
+         if is_undefined evd evk then
+         (* We try abstraction or concretisation for *)
+         (* this unconstrained occurrence *)
+         (* and we use typing to propagate this instantiation *)
+         (* We avoid making an arbitrary choice by leaving candidates *)
+         (* if both can work *)
+         let evi = Evd.find_undefined evd evk in
+         let vid = mkVar id in
+         let candidates = [inst; vid] in
+           try
+             let evd, ev = Evarutil.restrict_evar evd evk (Evd.evar_filter evi) (Some candidates) in
+             let evi = Evd.find evd ev in
+               (match evar_candidates evi with
+               | Some [t] ->
+                 if not (noccur_evar env_rhs evd ev (EConstr.of_constr t)) then
+                   raise (TypingFailed evd);
+                 instantiate_evar evar_unify flags evd ev (EConstr.of_constr t)
+               | Some l when abstract = Abstraction.Abstract &&
+                          List.exists (fun c -> isVarId evd id (EConstr.of_constr c)) l ->
+                 instantiate_evar evar_unify flags evd ev vid
+               | _ -> evd)
+           with e -> user_err (Pp.str "Cannot find an instance")
+         else
+           ((if !debug_ho_unification then
+               let evi = Evd.find evd evk in
+               let env = Evd.evar_env evi in
+               Feedback.msg_debug Pp.(str"evar is defined: " ++
+                 int (Evar.repr evk) ++ spc () ++
+                 prc env (match evar_body evi with Evar_defined c -> c
+                   | Evar_empty -> assert false)));
+            evd)
+       in force_instantiation evd evs
+     | [] -> abstract_free_holes evd l
+     in force_instantiation evd !evsref
+   | [] ->
      if Evd.is_defined evd evk then
        (** Can happen due to dependencies: instantiating evars in the arguments of evk might
            instantiate evk itself. *)
@@ -1453,25 +1451,23 @@ let second_order_matching flags env_rhs evd (evk,args) (test,argoccs) rhs =
           end;
         evd)
      else
-       let evd =
-         try
-           let evi = Evd.find_undefined evd evk in
-           let evenv = evar_env evi in
-           let evdref = ref evd in
-           let rhs' = nf_evar !evdref rhs' in
+       try
+         let evi = Evd.find_undefined evd evk in
+         let evenv = evar_env evi in
+         let evdref = ref evd in
+         let rhs' = nf_evar !evdref rhs' in
            if !debug_ho_unification then
              Feedback.msg_debug Pp.(str"abstracted type before second solve_evars: " ++
                                       prc evenv rhs');
-           (** solve_evars is not commuting with nf_evar, because restricting
-               an evar might provide a more specific type. *)
-           let evd, _ = !solve_evars evenv evd rhs' in
-           evdref := evd;
-           if !debug_ho_unification then
-             Feedback.msg_debug Pp.(str"abstracted type: " ++ prc evenv (nf_evar !evdref rhs'));
-           let flags = default_flags_of TransparentState.full in
-           Evarsolve.check_evar_instance evar_unify flags !evdref evk rhs'
+         (** solve_evars is not commuting with nf_evar, because restricting
+             an evar might provide a more specific type. *)
+          let evd, _ = !solve_evars evenv evd rhs' in
+          evdref := evd;
+          if !debug_ho_unification then
+            Feedback.msg_debug Pp.(str"abstracted type: " ++ prc evenv (nf_evar !evdref rhs'));
+          let flags = default_flags_of TransparentState.full in
+            Evarsolve.instantiate_evar evar_unify flags !evdref evk rhs'
          with IllTypedInstance _ -> raise (TypingFailed evd)
-       in Evd.define evk rhs' evd
   in
   let evd = abstract_free_holes evd subst in
   evd, true
@@ -1484,6 +1480,8 @@ let default_evar_selection flags evd (ev,args) =
     | _ :: args, a :: abs ->
       let spec =
         if not flags.allow_K_at_toplevel then
+          (* [evar_absorb_arguments] puts an Abstract flag for the
+             toplevel binders that were absorbed. *)
           let occs =
             if a == Abstraction.Abstract then Locus.AtLeastOneOccurrence
             else Locus.AllOccurrences
@@ -1624,8 +1622,7 @@ let rec solve_unconstrained_evars_with_candidates flags evd =
       | a::l ->
           (* In case of variables, most recent ones come first *)
           try
-            let evd = check_evar_instance evar_unify flags evd evk a in
-            let evd = Evd.define evk a evd in
+            let evd = instantiate_evar evar_unify flags evd evk a in
             match reconsider_unif_constraints evar_unify flags evd with
             | Success evd -> solve_unconstrained_evars_with_candidates flags evd
             | UnifFailure _ -> aux l
@@ -1645,8 +1642,7 @@ let solve_unconstrained_impossible_cases env evd =
       let evd' = Evd.merge_context_set Evd.univ_flexible_alg ?loc evd' ctx in
       let ty = j_type j in
       let flags = default_flags env in
-      let evd' = check_evar_instance evar_unify flags evd' evk ty in
-        Evd.define evk ty evd'
+      instantiate_evar evar_unify flags evd' evk ty
     | _ -> evd') evd evd
 
 let solve_unif_constraints_with_heuristics env
--- a/pretyping/evarconv.ml
+++ b/pretyping/evarconv.ml
@@ -1127,16 +1127,14 @@ type occurrence_match_test =
   env -> evar_map -> constr ->
   env -> evar_map -> int -> constr -> constr -> bool * evar_map
 
-type prefer_abstraction = bool
-
 type occurrence_selection =
   | AtOccurrences of Locus.occurrences
-  | Unspecified of prefer_abstraction
+  | Unspecified of Abstraction.abstraction
 
 type occurrences_selection =
   occurrence_match_test * occurrence_selection list
 
-let default_occurrence_selection = Unspecified false
+let default_occurrence_selection = Unspecified Abstraction.Imitate
 
 let default_occurrence_test ~frozen_evars ts _ origsigma _ env sigma _ c pat =
   let flags = { (default_flags_of ~subterm_ts:ts ts) with frozen_evars } in
@@ -1406,7 +1404,8 @@ let second_order_matching flags env_rhs evd (evk,args) (test,argoccs) rhs =
                       raise (TypingFailed evd);
                     let evd = Evd.define ev (EConstr.of_constr t) evd in
                     check_evar_instance evd ev (EConstr.of_constr t) (conv_fun evar_conv_x flags)
-                 | Some l when abstract && List.exists (fun c -> isVarId evd id (EConstr.of_constr c)) l ->
+                 | Some l when abstract = Abstraction.Abstract &&
+                     List.exists (fun c -> isVarId evd id (EConstr.of_constr c)) l ->
                     let evd = Evd.define ev vid evd in
                     check_evar_instance evd ev vid (conv_fun evar_conv_x flags)
                  | _ -> evd)
@@ -1468,10 +1467,14 @@ let default_evar_selection flags evd (ev,args) =
   let rec aux args abs =
     match args, abs with
     | _ :: args, a :: abs ->
-       let spec = if not flags.allow_K_at_toplevel then
-                    AtOccurrences (if a then Locus.AtLeastOneOccurrence else Locus.AllOccurrences)
-                  else Unspecified a in
-       spec :: aux args abs
+      let spec =
+        if not flags.allow_K_at_toplevel then
+          let occs =
+            if a == Abstraction.Abstract then Locus.AtLeastOneOccurrence
+            else Locus.AllOccurrences
+          in AtOccurrences occs
+        else Unspecified a
+      in spec :: aux args abs
     | l, [] -> List.map (fun _ -> default_occurrence_selection) l
     | [], _ :: _ -> assert false
   in aux (Array.to_list args) evi.evar_abstract_arguments
--- a/pretyping/evarconv.ml
+++ b/pretyping/evarconv.ml
@@ -154,31 +154,41 @@ let position_problem l2r = function
   | CUMUL -> Some l2r
 
 (* [occur_rigidly ev evd t] tests if the evar ev occurs in a rigid
-   context in t
+   context in t. Precondition: t has a rigid head.
 
-  That function should be an over approximation of occur-check, it can
-  return true even if the occur-check would fail on the normal form, as
-  otherwise we will postpone unsolvable constraints while maybe a
-  reduction would have allowed unification (see bug 3539 for example).
+   That function is an under approximation of occur-check, it can return
+   false even if the occur-check would succeed on the normal form.  This
+   means we might postpone unsolvable constraints which will ultimately
+   result in an occur-check after reductions. If it returns true, we
+   know that the occur-check would also return true on the normal form.
 
-  The boolean indicates if the term is a rigid head. For applications,
-  this means than an occurrence of the evar in arguments should be looked
-  at to find an occur-check. *)
+   The boolean indicates if the term is a rigid head. For applications,
+   this means than an occurrence of the evar in arguments should be looked
+   at to find an occur-check. *)
 
-let occur_rigidly (evk,_ as ev) evd t =
+let occur_rigidly flags env evd (evk,_) t =
   let rec aux t =
     match EConstr.kind evd t with
-    | App (f, c) -> if aux f then Array.exists aux c else false
-    | Construct _ | Ind _ | Sort _ | Meta _ | Fix _ | CoFix _ | Int _ -> true
-    | Proj (p, c) -> not (aux c)
-    | Evar (evk',_) -> if Evar.equal evk evk' then raise Occur else false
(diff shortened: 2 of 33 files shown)


engine/evd.mli
```diff
@@ -98,6 +106,10 @@ type evar_info = {
   (** Boolean mask over {!evar_hyps}. Should have the same length.
       When filtered out, the corresponding variable is not allowed to occur
       in the solution *)
+  evar_abstract_arguments : Abstraction.t;
+  (** Boolean information over {!evar_hyps}, telling if an hypothesis instance
+      can be immitated or should stay abstract in HO unification problems
+      and inversion (see [second_order_matching_with_args] for its use). *)
```
gares: Am I understanding well that this "abstraction" thing lets you chose a solution for
```
H |- ?e[H] === f(H)
```
among:
- `H |- ?e := fun _ => f(H)`
- ` H |- ?e := fun H1 => f(H1)`
In which case the boolean corresponding to H is supposed to be true/false?

pretyping/evarconv.ml
```diff
@@ -103,28 +127,48 @@ type flex_kind_of_term =
...
+  That function should be an over approximation of occur-check, it can
+  return true even if the occur-check would fail on the normal form, as
+  otherwise we will postpone unsolvable constraints while maybe a
+  reduction would have allowed unification (see bug 3539 for example).
+
+  The boolean indicates if the term is a rigid head. For applications,
+  this means than an occurrence of the evar in arguments should be looked
+  at to find an occur-check.
+
+  TODO: replace with a test on the normal form and evaluate performance.
```
gares: This should be a true issue, not a comment.
Or just discarded given that, in general, it is arbitrarily expensive.

ppedrot (commented):
Not my area of competence.

engine/evd.mli
```diff
@@ -98,6 +106,10 @@ type evar_info = {
   (** Boolean mask over {!evar_hyps}. Should have the same length.
       When filtered out, the corresponding variable is not allowed to occur
       in the solution *)
+  evar_abstract_arguments : Abstraction.t;
+  (** Boolean information over {!evar_hyps}, telling if an hypothesis instance
+      can be immitated or should stay abstract in HO unification problems
```
herbelin: Typo: imitated

pretyping/evarconv.ml
```diff
@@ -43,6 +57,16 @@ let _ = Goptions.declare_bool_option {
   Goptions.optwrite = (fun a -> debug_unification:=a);
 }
 
+let debug_ho_unification = ref (false)
+let _ = Goptions.declare_bool_option {
+  Goptions.optdepr = false;
+  Goptions.optname =
+    "Print debug information for the higer-order unification algorithm";
```
herbelin: Typo: higher.

Also, `optname` is used for printing a message on `Test`. It should take the place of `s` in `The s mode is off/on`. So, it should be something like `the higher-order unification algorithm
debug printing`.

herbelin (commented):
Looks very good. 

Looked only at a little part yet. Shall continue later.

engine/evd.mli
mattam82: Exactly. If the argument is marked abstract then the abstraction solution `(fun H1 => f (H1))` will be chosen, if possible.

pretyping/evarconv.ml
mattam82: Indeed, will remove, I guess we all know it would be prohibitive.

mattam82:
Rebased and first comments addressed

engine/evd.mli
gares: I think you should clarify the comment, maybe copy pasting the example up there.

In particular it is still not clear to me what `true` means.
In the example above you could just say "when the bool for H is true".

An alternative is to use a variant instead of the boolean, so that one would have `Imitate` or `Abstract` and then the comment would not be needed.

engine/evd.mli
```diff
@@ -77,6 +77,14 @@ sig
 
 end
 
+module Abstraction : sig
+  type t = bool list
```
gares: Can we use a variant instead of `bool` so that the meaning of `true` would be clearer?

pretyping/evarconv.ml
```diff
@@ -133,7 +175,8 @@ let occur_rigidly (evk,_ as ev) evd t =
     | Proj (p, c) -> not (aux c)
     | Evar (evk',_) -> if Evar.equal evk evk' then raise Occur else false
     | Cast (p, _, _) -> aux p
-    | Lambda _ | LetIn _ -> false
+    | Lambda (na, t, b) -> aux b
+    | LetIn (na, _, _, b) -> aux b
```
gares: Can you explain this change please?
Why are you skipping the body of the let-in and the type of the lambda?

gares:
The changes to evarconv are quite substantial, I really need more time to read the diff.
I think we should also benchmark this change (it should not change the performance as far as I understand it, but still...)

proofs/goal.ml
```diff
@@ -65,6 +65,7 @@ module V82 = struct
     let evi = { Evd.evar_hyps = hyps;
 		Evd.evar_concl = concl;
 		Evd.evar_filter = Evd.Filter.identity;
+                Evd.evar_abstract_arguments = Evd.Abstraction.identity;
```
gares: there is something wrong with the indentation

engine/evd.mli
mattam82: How do "Abstract" and "Transparent" sound?

pretyping/evarsolve.mli
```diff
@@ -31,14 +40,16 @@ val expand_vars_in_term : env -> evar_map -> constr -> constr
    some problems that cannot be solved in a unique way (except if choose is
    true); fails if the instance is not valid for the given [ev] *)
 
-type conv_fun =
-  env ->  evar_map -> conv_pb -> constr -> constr -> unification_result
+type types_or_terms = bool
```
gares: Can you please describe what the boolean stands for and why it is relevant to the conversion (unification?) function to know that?

pretyping/evarconv.ml
mattam82: It is an approximation: for lambdas and let-ins, short of doing reduction we don't know if an evar occurrence in the binding type or body of the let-in is indeed rigid if we reduced, so we only check the body. If it appears rigidly in the body then we know it would appear rigidly in the normal form as well.

pretyping/evarsolve.ml
```diff
@@ -1601,46 +1622,33 @@ let rec invert_definition conv_algo choose env evd pbty (evk,argsv as ev) rhs =
...
-           pr_evar_map evd' ++ fnl() ++
-           str "----> " ++ int ev ++ str " := " ++
-           print_constr body);
-        raise e in*)
-    let evd' = check_evar_instance evd' evk body conv_algo in
-    Evd.define evk body evd'
+    let evd' = Evd.define evk body evd' in
+    (** Check after definition, as checking could involve the same
+        evar definition problem again otherwise *)
+    check_evar_instance evd' evk body conv_algo
```
gares: This is quite counterintuitive.

I understand that conv_algo is eally a unif_algo, so we may loop.

But if evk does not occur in body, how can it occur (checked above) in the type of body?
Can you explain better why this change is needed?

pretyping/evarsolve.mli
gares: I think you should also rename conv_fun into unif_fun since it is really unification.

engine/evd.mli
gares: In the comment below you talk about abstraction and imitation, IIRC.

pretyping/evarconv.ml
```diff
@@ -537,27 +592,91 @@ and evar_eqappr_x ?(rhs_is_already_stuck = false) ts env evd pbty
...
 		(* Happens thanks to w_unify building ill-typed terms *) 
 		  default
 	      in f evd
 	  end
       | _ -> default evd
   in
-  let flex_rigid on_left ev (termF, skF as apprF) (termR, skR as apprR) =
+  let rec first_order flexrigid env i t1 t2 sk1 sk2 =
+    (* Try first-order unification *)
+    match ise_stack2 false env i (evar_conv_x flags) sk1 sk2 with
```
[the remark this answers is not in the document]
mattam82: The thing is that zipping is expensive I guess, so we don't want to factor here. Note that this is just a refactorization of the original code (previously from L596 on). The changes to evar_conv itself is this refactoring and the passing of a record of flags instead of a tuple. Indeed I had generalized this first_order by a flexrigid flag at some point but it is not used in the end, I will clean that up.

pretyping/evarconv.ml
```diff
@@ -569,6 +688,7 @@ and evar_eqappr_x ?(rhs_is_already_stuck = false) ts env evd pbty
 	    (fun () ->
 	      ise_try evd 
 	        [eta;(* Postpone the use of an heuristic *)
+                 (* (fun i -> first_order true env i termF termR skF skR); *)
```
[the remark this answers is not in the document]
mattam82: Yep, remnant of an experiment.

pretyping/evarconv.ml
gares: I'm sorry but I'm totally lost.

Why recurring under lambdas, but not in the branches of a match for example.
Also, if you don't go in the type of a lambda, how can this code "over approximate" occur check (that I guess looks everywhere).

Also if t is `(fix ....) t1... ev.. tn` it says true, while if t it is `def t1..ev..tn` is says false (`def` being a constant).
Both context are not necessarily rigid, so I really don't get what this code is supposed to do.

Finally, `Proj` flips the boolean: turning a regular projection into a primitive one changes what this code does?

It is good that you've add a comment, but I'm still puzzled and I don't understand how to explain what
this code does.

I looked at the bug you mention, it is "?X ?Y == X? (mk_pair (fst ?Y) (snd ?Y))". Can you explain the role played by this code? I expect it is `ev=?Y` and `t=(mk_pair (fst ?Y) (snd ?Y))`. When `fst` is primitive the `raise Occur` code path is reached, but if `fst` is a regular constant it is not.

pretyping/evarsolve.mli
gares: So the bool is `true` when `conv_fun` is  called on a term?
I'm really asking to document the API so that it is easier to understand.

May I suggest renaming `conv_fun` -> `unifier` and `conv_fun_bool` -> `conversion_check`?
And `types_or_term` -> `disable_delta` ? (or something more precise than `types_or_terms`).

pretyping/evarsolve.ml
mattam82: IIRC the problem is due to constraints and candidates and not the precise typechecking we are trying to perform. If conv_algo indeed tries to solve some constraint involving evk it might try to instantiate evk again. Sadly it's been a year since I looked into this so I don't have a precise example of a problematic case to give. Another way would be to force evk to be considered frozen in the check_evar_instance's conv_algo calls to prevent such instantiations.

gares: This is what I though too: one could write the code in the expected order but freeze the evar under processing.

pretyping/evarconv.ml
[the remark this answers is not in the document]
gares: I see, so I'm fine with not factoring this code.

pretyping/evarconv.ml
```diff
@@ -1110,117 +1211,289 @@ let set_solve_evars f = solve_evars := f
...
-        let evty = set_holes evdref cty subst in
-        let instance = Filter.filter_list filter instance in
-        let evd = !evdref in
-        let (evd, ev) = new_evar_instance sign evd evty ~filter instance in
-        evdref := evd;
-        evsref := (fst (destEvar !evdref ev),evty)::!evsref;
-        ev in
-      set_holes evdref (apply_on_subterm env_rhs evdref set_var c rhs) subst
+  let fixed = ref Evar.Set.empty in
+  let rec set_holes env_rhs evdref rhs = function
```
gares: I think it would make the code simpler to understand if this function would take in input an evd, and use an evdref only locally to call other functions that use the imperative API

pretyping/evarconv.ml
```diff
@@ -862,72 +923,84 @@ and evar_eqappr_x ?(rhs_is_already_stuck = false) ts env evd pbty
...
 	| Construct _, Construct _ ->
-	  rigids env evd sk1 term1 sk2 term2
+           rigids env evd sk1 term1 sk2 term2
+
+        | Evar (sp1,al1), Evar (sp2,al2) -> (** Frozen evars *)
+          if Evar.equal sp1 sp2 then
+            match ise_stack2 false env evd (evar_conv_x flags) sk1 sk2 with
+            |None, Success i' ->
+              (** FIXME: solve_refl can restrict the evar, do we want to allow that? *)
+              Success (solve_refl flags (fun p env i pbty a1 a2 ->
```
gares: In general, when unifying X with X, pruning X is OK to me if the two explicit substitution do not unify.
But here it seems that X is frozen (rigid case) hence pruning should be forbidden.

I think the FIXME should be addressed.

pretyping/evarsolve.ml
```diff
@@ -1344,7 +1362,7 @@ let solve_refl ?(can_drop=false) conv_algo env evd pbty evk argsv1 argsv2 =
   let args = Array.map2 (fun a1 a2 -> (a1, a2)) argsv1 argsv2 in
   let untypedfilter =
     restrict_upon_filter evd evk
-      (fun (a1,a2) -> conv_algo env evd Reduction.CONV a1 a2) args in
+      (fun (a1,a2) -> conv_algo false env evd Reduction.CONV a1 a2) args in
   let candidates = filter_candidates evd evk untypedfilter NoUpdate in
   let filter = closure_of_filter evd evk untypedfilter in
```
gares: To fix the FIXME about froze, it seem you just need to add a boolean "forbid pruning" and use it here

gares (changes requested):
All the code that I tried to understand makes more sense than before. The changes are too extensive to be sure the code is correct.

I think the only points I'd like to see addressed are the ones that make the code more clear (Eg there are many evdref here and there that seems to be linearly used).

I'd also like a high level explanation (somewhere in a comment) on what "frozen evars" are used for.

gares:
I'm sorry it took me so long. The new code seems to be a good improvement.

herbelin:
> @herbelin said: Looked only at a little part yet. Shall continue later.

@gares having made a very good job at reviewing this PR, I'm tempted to rely on his review and to simply recommend the merge as soon as he's satisfied.

engine/evd.mli
mattam82: Right. Abstract and Imitate are the better names

pretyping/evarconv.ml
mattam82: Sorry for the confusion, that function is tricky indeed. The comment was wrong. And the definition as well. What we want is an under-approximation of occur-check: it should return true only if an occur-check would return true on the normal form, but it might return false while the occur-check on the normal form would return true. So, the fix/cofix/meta cases were wrong: they are not definitely rigid heads. Proj is not either, unless it is not unfoldable, and same goes for constants. I adapted the code and the comment to reflect this.

In the example bug, occur_rigidly will now return false in both cases.

proofs/goal.ml
mattam82: The surrounding lines used tabs, I changed them to spaces.

mattam82:
I implemented a few of the changes but will need more time to finish it. I'll come back to it in september. As @herbelin guessed, this is actually only one part of unifall, the more interesting change is the move to evar-based clauses coming on top of this, which won't make it to 8.9 anyway. I'll give you guys a ping when all the comments are addressed.

pretyping/evarconv.mli
```diff
@@ -17,14 +17,32 @@ open Locus
...
+
+val conv_fun : unify_fun -> unify_flags -> Evarsolve.conv_fun
+
 exception UnableToUnify of evar_map * Pretype_errors.unification_error
 
 (** {6 Main unification algorithm for type inference. } *)
 
-(** returns exception NotUnifiable with best known evar_map if not unifiable *)
+(** returns exception UnableToUnify with best known evar_map if not unifiable *)
 val the_conv_x     : env -> ?ts:transparent_state -> constr -> constr -> evar_map -> evar_map
```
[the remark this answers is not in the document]
mattam82: Yes indeed, done.

pretyping/evarconv.ml
```diff
@@ -1398,17 +1678,39 @@ let make_opt = function
...
+  match evar_conv_x flags env evd CONV t1 t2 with
+  | Success evd' -> evd'
+  | UnifFailure (evd',e) -> raise (UnableToUnify (evd',e))
+
+let unify_leq flags env evd t1 t2 =
+  match evar_conv_x flags env evd CUMUL t1 t2 with
+  | Success evd' -> evd'
+  | UnifFailure (evd',e) -> raise (UnableToUnify (evd',e))
+
+let unify_with_heuristics flags ~with_ho env evd cv_pb ty1 ty2 =
```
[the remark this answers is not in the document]
mattam82: That's right, it's not apparent enough. I went the relatively easy way, renaming unify_with_heurisitcs to unify and unify/unify_leq to unify_delay and unify_leq_delay which makes it clearer that they delay constraints. Changing their types significantly to project out the constraints would be overkill I think, as this interface fits the uses in typing which expects to gather constraints and call solve_unif_constraints_with_heuristics at the end.

pretyping/evarconv.ml
mattam82: Indeed. I could avoid it entirely in second_order_matching, there remains cases in apply_on_subterm and thin_evars which make use of map_constr_with_binders that doesn't naturally allow side effects on the evar_map.

pretyping/evarconv.ml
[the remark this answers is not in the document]
mattam82: I removed the flexrigid as it is unused

pretyping/evarconv.ml
mattam82: I should be more obvious what the function does now.

pretyping/evarconv.ml
```diff
@@ -1110,117 +1211,289 @@ let set_solve_evars f = solve_evars := f
...
+    UnifFailure (evd, CannotSolveConstraint (pb,ProblemBeyondCapabilities))
 
-let apply_conversion_problem_heuristic ts env evd pbty t1 t2 =
-  let t1 = apprec_nohdbeta ts env evd (whd_head_evar evd t1) in
-  let t2 = apprec_nohdbeta ts env evd (whd_head_evar evd t2) in
+let is_beyond_capabilities = function
+  | CannotSolveConstraint (pb,ProblemBeyondCapabilities) -> true
+  | _ -> false
+
+(* TODO frozen *)
```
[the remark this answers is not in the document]
mattam82: Right. What remained to do was to handle frozen evars correctly here, I just implemented it and removed the todo. Frozen evars cannot be instantiated so this was just a matter of restricting the rules so that no frozen evars is instantiated.

pretyping/evarsolve.ml
mattam82: Yes indeed, solve_refl succeeds only when there is no restriction needed.

mattam82:
> All the code that I tried to understand makes more sense than before. The changes are too extensive to be sure the code is correct.

The whole thing needed quite a bit of overhaul indeed. Hopefully the CI can confirm I did not introduce bugs in the many refactorings I did.

> I think the only points I'd like to see addressed are the ones that make the code more clear (Eg there are many evdref here and there that seems to be linearly used).

I tried to improve that in the latest version. If you still see some I'd be happy to fix them.

> I'd also like a high level explanation (somewhere in a comment) on what "frozen evars" are used for.

In my last commit I documented the unify_flags better, are you satisfied with it? The frozen_evars set allows to do higher-order matching easily instead of unification, ensuring we do not touch (at least a subset of) evars that were already present in a goal.

mattam82:
@gares @herbelin this is ready for another round. It might have CI issues due to things I did this afternoon but should be mostly ready now.

engine/evd.mli
mattam82: I went for the Imitate and Abstract variants here.

mattam82:
Bench here: https://ci.inria.fr/coq/view/benchmarking/job/benchmark-part-of-the-branch/552

pretyping/evarconv.ml
mattam82: This is now adressed in solve_refl, indeed.

pretyping/evarconv.ml
```diff
@@ -1110,117 +1211,289 @@ let set_solve_evars f = solve_evars := f
...
+            let body = match evar_body evi with Evar_empty -> assert false | Evar_defined c -> c in
+            Feedback.msg_debug Pp.(str"evar was defined already as: " ++ prc evenv body)
+          end;
+        evd)
+     else
+       let evd =
+         try
+           let evi = Evd.find_undefined evd evk in
+           let evenv = evar_env evi in
+           let evdref = ref evd in
```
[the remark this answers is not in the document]
mattam82: This is gone

mattam82:
I rebased to latest master and updated the overlays for equations and mtac2. I expect fiat-crypto will fail for an unrelated reason again. @gares can you take it? I addressed all your previous comments I believe.

ejgallego:
It seems it's important to get this early on to have 8.10 testing; maybe we could flag PRs such as this one with `priority: high` ? WDYT folks [cc @Zimmi48 @vbgl ]

Zimmi48:
Unless someone else wants to take over, this just depends on @mattam82's schedule.

mattam82:
This depends on reviewers accepting it mainly

gares:
I'm sorry I did not have time, I'll try to review it today.

gares (approved):

gares:
The code is fine to me. Just one thing: there was a bench, but changes happened before. Should we run another one, or the changes were just minor (as the ones I've asked IIRC).

mattam82:
I'll run another one just to be on the safe side

gares:
Sorry but I don't see the job on pendulum

mattam82:
That's because I didn't launch it yet :)

mattam82:
https://ci.inria.fr/coq/view/benchmarking/job/benchmark-part-of-the-branch/649/console should have it in a while

mattam82:
```
┌────────────────────────┬──────────────────────────┬──────────────────────────────────────────────┬──────────────────────────────────────────────┬───────────────────────────────┬──────────────────────────┐
│                        │      user time [s]       │                  CPU cycles                  │               CPU instructions               │     max resident mem [KB]     │        mem faults        │
│                        │                          │                                              │                                              │                               │                          │
│           package_name │     NEW     OLD  PDIFF   │               NEW               OLD  PDIFF   │               NEW               OLD  PDIFF   │        NEW        OLD PDIFF   │    NEW     OLD   PDIFF   │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│       coq-fiat-parsers │    3.80    6.79 -44.04 % │        9352008007       16669498033 -43.90 % │       17404823174       34295538284 -49.25 % │     178928     178888 +0.02 % │      0      19 -100.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│    coq-formal-topology │   29.92   30.12  -0.66 % │       82968339384       82945233729  +0.03 % │      108362229572      108237689264  +0.12 % │     468052     467796 +0.05 % │     21       1 +2000.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│             coq-sf-plf │   43.08   43.28  -0.46 % │      120654331889      120741873356  -0.07 % │      152792037284      153285650180  -0.32 % │     492004     491604 +0.08 % │     48      11 +336.36 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│  coq-mathcomp-fingroup │   54.39   54.55  -0.29 % │      152025685795      151900555890  +0.08 % │      198534520156      198381121022  +0.08 % │     574712     574264 +0.08 % │     77       2 +3750.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│     coq-mathcomp-field │  220.38  220.82  -0.20 % │      614514664266      615345124630  -0.13 % │      893093377341      892466957007  +0.07 % │     745128     749384 -0.57 % │      0       0    +nan % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│            coq-unimath │ 2052.43 2055.10  -0.13 % │     5714491695316     5722017905319  -0.13 % │    10001611864877    10003650991588  -0.02 % │     918620     956536 -3.96 % │    516     256 +101.56 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│ coq-mathcomp-ssreflect │   38.99   39.02  -0.08 % │      107616601662      107431379926  +0.17 % │      128621724423      128505874043  +0.09 % │     523840     523336 +0.10 % │     63      11 +472.73 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│ coq-mathcomp-character │  251.34  250.92  +0.17 % │      700449113940      698455433493  +0.29 % │      965792482354      963758807955  +0.21 % │    1066528    1067956 -0.13 % │     10     112  -91.07 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│  coq-mathcomp-solvable │  190.90  190.57  +0.17 % │      532760481833      531116173499  +0.31 % │      713147437010      712385984562  +0.11 % │     851248     851600 -0.04 % │     13       2 +550.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│           coq-compcert │ 1026.92 1024.38  +0.25 % │     2862780560381     2853836340453  +0.31 % │     4368391283067     4365948284142  +0.06 % │    1305272    1313264 -0.61 % │    222     153  +45.10 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│               coq-hott │  342.54  341.44  +0.32 % │      929957320168      927686396783  +0.24 % │     1409791362707     1407037744352  +0.20 % │     479572     479392 +0.04 % │    451     220 +105.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│        coq-lambda-rust │ 2084.09 2075.95  +0.39 % │     5809276883336     5787594188268  +0.37 % │     7971340446904     7958329162423  +0.16 % │    1382104    1371696 +0.76 % │     36       0    +nan % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│   coq-mathcomp-algebra │  152.50  151.87  +0.41 % │      425126870228      423599555780  +0.36 % │      535773774750      534944764124  +0.15 % │     632716     637296 -0.72 % │     10      41  -75.61 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│               coq-corn │ 1501.60 1494.94  +0.45 % │     4192562269589     4171278804873  +0.51 % │     6174162600454     6168785701392  +0.09 % │     838644     855300 -1.95 % │      1       1   +0.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│              coq-flocq │  567.03  564.00  +0.54 % │     1577889444308     1566619943533  +0.72 % │     2283176485889     2281376272751  +0.08 % │    1718312    1720396 -0.12 % │    308     334   -7.78 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│ coq-mathcomp-odd-order │ 1387.70 1379.93  +0.56 % │     3868582104088     3846126901324  +0.58 % │     6371600506770     6369090604462  +0.04 % │    1357624    1350144 +0.55 % │    102      53  +92.45 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│              coq-verdi │  123.88  123.12  +0.62 % │      344302382253      342433255538  +0.55 % │      443626757939      441852133380  +0.40 % │     611508     593860 +2.97 % │      9      10  -10.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│              coq-color │  685.86  681.41  +0.65 % │     1918838131653     1905035902795  +0.72 % │     2290518264169     2282691127510  +0.34 % │    1380196    1386408 -0.45 % │    107      60  +78.33 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│             coq-geocoq │ 2003.04 1989.57  +0.68 % │     5589655700519     5553279642936  +0.66 % │     8059662835769     8056798689862  +0.04 % │    1200832    1188004 +1.08 % │    185     234  -20.94 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│         coq-verdi-raft │ 1376.15 1365.94  +0.75 % │     3841325002710     3812219329878  +0.76 % │     5235535090210     5217747518544  +0.34 % │    1808824    1800432 +0.47 % │      0       0    +nan % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│       coq-math-classes │  218.26  216.56  +0.79 % │      610596627198      605467647643  +0.85 % │      769732060610      767794937634  +0.25 % │     517308     514536 +0.54 % │     22      17  +29.41 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│        coq-fiat-crypto │ 3982.86 3951.73  +0.79 % │    11067126908774    10978858990892  +0.80 % │    17704598405129    17601190421045  +0.59 % │    2468212    2466912 +0.05 % │    727    1016  -28.44 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│            coq-bignums │   70.32   69.49  +1.19 % │      195455222789      193746656851  +0.88 % │      257376034424      257310508892  +0.03 % │     527812     523072 +0.91 % │    169     160   +5.62 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼──────────────────────────┤
│         coq-coquelicot │   74.52   73.48  +1.42 % │      204126564104      201738931769  +1.18 % │      244862453736      243379038792  +0.61 % │     691280     687128 +0.60 % │    205      35 +485.71 % │
└────────────────────────┴──────────────────────────┴──────────────────────────────────────────────┴──────────────────────────────────────────────┴───────────────────────────────┴──────────────────────────┘
```

I think it is in the margin of noise (note I didn't rebase since feb 8th)

SkySkimmer:
Note that fiat-parsers is currently silently broken, similar to what happened in https://github.com/mit-plv/fiat/pull/21

gares:
Ok, I'll merge asap

pretyping/unification.ml
```diff
@@ -141,16 +141,25 @@ let abstract_list_all env evd typ c l =
   evd,(p,typp)
 
 let set_occurrences_of_last_arg args =
-  Some AllOccurrences :: List.tl (Array.map_to_list (fun _ -> None) args)
+  Evarconv.AtOccurrences AllOccurrences ::
+    List.tl (Array.map_to_list (fun _ -> Evarconv.Unspecified Abstraction.Abstract) args)
```
mattam82: Dangerous, we don’t check that this list is non-empty

PR #7821, opened 2018-06-15
[refine] obey the use_unification_heuristics flag

mattam82:
Refine now respects this existing flag so users can control when and when not to resolve unification constraints heuristically. Observationally this does not change anything yet as the heuristics used during typechecking (i.e. evaraconv) are the same as the ones used by solve_remaining, and we silently ignore during typechecking of the term by refine if the constraints can be solved or not, but still try it even if the flag is off.

**Kind:** enhancement

### DIFF
--- a/plugins/ltac/extratactics.ml4
+++ b/plugins/ltac/extratactics.ml4
@@ -293,7 +293,7 @@ open Vars
 
 let constr_flags () = {
   Pretyping.use_typeclasses = true;
-  Pretyping.solve_unification_constraints = true;
+  Pretyping.solve_unification_constraints = Pfedit.use_unification_heuristics ();
   Pretyping.use_hook = Pfedit.solve_by_implicit_tactic ();
   Pretyping.fail_evar = false;
   Pretyping.expand_evars = true }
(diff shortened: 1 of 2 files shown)


ppedrot (approved):
I guess that's OK...

maximedenes:
Can we get a test case?

ppedrot:
That's in the little check boxes in @mattam82's original post :)

maximedenes:
> That's in the little check boxes in @mattam82's original post :)

Very good! Let's put a label then :)

mattam82:
I can't cook up an example right now where this shows up, as all the heuristics in the current evar_conv are also done during unification. E.g. first-order solutions to ?X 0 ~= 0 < 0 are not postponed and higher-order unification is disabled, and we catch errors from solve_unif_constraints in understand_ltac used by refine, so if none apply we keep the constraints around (except if fail_evar is true). The difference of behavior is that with this patch when doing Unset Solve Unification Constraints [refine foo] really does not call solve_unif_constraints_with_heuristics anymore (as can be seen by tracing), neither during typechecking of [foo] nor after the dot.

ppedrot:
@mattam82 IIUC you consider this PR ready? If so can you rebase it on a recent master for the CI?

mattam82:
Yes, will do.

ppedrot:
@mattam82 ping.

mattam82:
Done

PR #7822, opened 2018-06-15
cArray: proper invalid_arg exceptions

mattam82:
Just fix the strings which are otherwise misleading.


ppedrot (approved):
Not that important, as the OCaml compiler itself complains if your code depends on the string value of `Invalid_Arg`. Still a sensible patch.

PR #7823, opened 2018-06-15
[tactics] function to gather undef evars of the goal

mattam82:
Add a function to compute the undefined evars of a goal. Used by the new apply/rewrite but of more general interest.

### DIFF
--- a/proofs/tacmach.mli
+++ b/proofs/tacmach.mli
@@ -95,7 +95,7 @@ val refine           : constr -> tactic
 val pr_gls    : goal sigma -> Pp.t
 val pr_glls   : goal list sigma -> Pp.t
 
-(* Variants of [Tacmach] functions built with the new proof engine *)
+(** Variants of [Tacmach] functions built with the new proof engine *)
 module New : sig
   val pf_apply : (env -> evar_map -> 'a) -> Proofview.Goal.t -> 'a
   val pf_global : Id.t -> Proofview.Goal.t -> GlobRef.t
@@ -139,4 +139,6 @@ module New : sig
 
   val pf_nf_evar : Proofview.Goal.t -> constr -> constr
 
+  (** Gathers the undefined evars of the given goal. *)
+  val pf_undefined_evars : Proofview.Goal.t -> Evar.Set.t
 end
--- a/proofs/tacmach.ml
+++ b/proofs/tacmach.ml
@@ -227,4 +227,9 @@ module New = struct
 
   let pf_nf_evar gl t = nf_evar (project gl) t
 
+  let pf_undefined_evars gl =
+    let sigma = Proofview.Goal.sigma gl in
+    let ev = Proofview.Goal.goal gl in
+    let evi = Evd.find sigma ev in
+      Evarutil.filtered_undefined_evars_of_evar_info sigma evi
 end


ppedrot (commented):
Shouldn't you use `Evarutil.filtered_undefined_evars_of_evar_info` directly instead, and potentially use the caching mechanism?

engine/evarutil.ml
```diff
@@ -740,6 +740,13 @@ let undefined_evars_of_named_context evd nc =
     nc
     ~init:Evar.Set.empty
 
+let undefined_evars_of_econstr_named_context evd nc =
+  Context.Named.fold_outside
```
gares: If only we were using standard names, eg `fold_right` here one could ask why not using `fold_left` given that `union` is commutative and `fold_left` is more stack efficient.

mattam82:
@ppedrot Absolutely right, I changed the code to be just a wrapper around filtered_...

Zimmi48 (commented):
Try not to add new functions to an interface file without a documentation comment. In this precise case, I would be interested to read documentation rather than trying to understand what the function does from looking at the code. In particular, I'm wondering: does it get undefined evars that are currently mentioned in the current goal, or does it get all evars, including the ones that are coming from other goals?

gares:
@Zimmi48 I agree in general, but here the type is quite clear: `Goal.t -> Evd.Set.t`.

Another comment: IIRC to get a `Goal.t` one needs to `Goal.enter` and in such a context my feeling is that one should just use the `'a tactic` monad type, in this case `Evd.Set.t tactic`. Can you comment? Can you point me to the user of this API?

Zimmi48:
> I agree in general, but here the type is quite clear: Goal.t -> Evd.Set.t.

I don't know. Does the type answer my question about which undefined evars are returned?

mattam82:
A typical call is the following, under a `Goal.enter` for gl indeed, it gathers the undefined evars of the goal that unification should freeze (consider as rigid). The pf_ set of functions is here precisely to inspect the Goal.t structure without going under the monad, so it looks legitimate to me.
This is used in the "new" rewrite coming later on.
```
  let frozen_evars = if frzevars then Tacmach.New.pf_undefined_evars gl else Evar.Set.empty in
```

I added a comment.

ppedrot:
@mattam82 Note that computing frozen_evars upfront is a source of slowness in VST. Wrapping it in a lazy gave me a 2% speedup, which consisted in practice in dividing the time a handful of hugely slow computations by 4 or 5.

mattam82:
Yep, I know!

ppedrot:
I still have the patch, it's quite trivial. Want me to publish the PR or do you have it on your side?

mattam82:
The code snippet I gave is in the unifall branch, I guess I could maybe use
your PR if you publish it somewhere.

mattam82:
Rebased, if CI passe I consider this ready.

Zimmi48:
I'm not seeing what "kind" label to apply there. Should we create a new `kind: api` label? Or should we extend the meaning of `kind: infrastructure`?

mattam82:
Maybe an: Internal (API, ML documentation) would be generally useful

proofs/tacmach.mli
```diff
@@ -139,4 +139,6 @@ module New : sig
 
   val pf_nf_evar : Proofview.Goal.t -> constr -> constr
 
+  (** Gathers the undefined evars of the current goal. *)
```
gares: Sorry to nitpick on this detail, but given the questions of @Zimmi48 ...
It is not the current goal, but rather the goal given as input.

mattam82:
I fixed the comment and rebased.

Zimmi48:
Only CI failures are unrelated: Windows builds are broken everywhere.

Zimmi48:
Actually, they have stopped failing, so I restarted them.

mattam82:
@gares @ppedrot this is ready, so whoever wants to merge can

ppedrot (approved):
Let's merge.

PR #370, opened 2016-11-21
A tentative fix for #4413

herbelin:
This is a regression wrt v8.4 related to the change of order of resolution of evar-evar unification problems.

One possible "fix" is to give priority to the instantiation of the non pre-existing evar to solve a ?x=?y problem. Note that it happens to incidentally fix also #4095!

Started a coq-contribs test but I feel this kind of "fix" should also be tested on contribs using type inference intensively [Edited afterwards: no failures on contribs [here](https://ci.inria.fr/coq/view/coq-contribs/job/coq-contribs/603/console), nor for a simpler version based on  trunk+4413-evar-evar-fix2 [here](https://ci.inria.fr/coq/view/coq-contribs/job/coq-contribs/605/console)].

PR #991, opened 2017-08-22
Unifall: episode II - apply

mattam82:
This set of commits (on top of unifall-infra) ports apply to the evar_conv unification and evar-based clauses, leaving almost everything else unchanged (it still impacts reflexivity and eauto's exact which use apply or clenv's unify primitive). Right now I'm just curious about potential compatibility issues on Travis. The ones I found in the stdlib are either:
- apply/eapply discrepancies, apply was allowing itself to produce dependent subgoals sometimes
- different occurrence selection in higher-order unification, due to different heuristics being used. This is rare as often the user provides enough info (e.g. a predicate) to disambiguate.

Let's see what Travis has in store.

Depends on #7825.

PR #7825, opened 2018-06-15
[tactics] Refine test for unresolved evars: not reachable from initial evars

### LINKED ISSUE #4095: Regression: constr forces typeclass resolution that it did not previously force

Original bug ID: BZ#4095
From: @JasonGross
Reported version: 8.5
CC:   @mattam82

See also: [BZ#4413](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%234413%22)

### LINKED ISSUE #4413: Type inference engine incorrectly fails in the presence of evars (Regression since 8.4)

Original bug ID: BZ#4413
From: @JasonGross
Reported version: 8.5
CC:   @herbelin, @jonleivent

See also: [BZ#3872](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%233872%22)
See also: [BZ#3655](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%233655%22)
See also: [BZ#4095](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%234095%22)

### LINKED ISSUE #4631: stack overflow in [refine] (confusion about evar dependencies?) (regression since 8.4)

Original bug ID: BZ#4631
From: @JasonGross
Reported version: 8.5
CC:   @gares, @SkySkimmer

mattam82:
The test is refined to handle aliases: i.e. undefined evars coming from
restrictions and evar-evar unifications with an initial evar are not
considered fresh unresolved evars. To check this, we generalize
the restricted_evars map of the evar_map into an aliased_evars map,
which is extended in Evd.restrict and Evd.define_with_evar cases.
This is now on top of PR #8741. I basically implements the proposal
in PR #370 to have a test of unresolved evars independent of the evar-evar
orientation order.

This allows [apply] to refine an evar with a new one if it results from a
[clear] request or an evar-evar solution only, otherwise the new evar was
considered fresh and an error was incorrectly raised.

**Kind:** bug fix
- Fixes #4413
- Fixes #4095
- Fixes #4631
- [ ] Added test-suite files

### DIFF
--- a/doc/changelog/02-specification-language/07825-rechable-from-evars.rst
+++ b/doc/changelog/02-specification-language/07825-rechable-from-evars.rst
@@ -1,0 +1,9 @@
+- **Changed:**
+  In :tacn:`refine`, new existential variables unified with existing ones are no
+  longer considered as fresh. The behavior of :tacn:`simple refine` no longer depends on
+  the orientation of evar-evar unification problems, and new existential variables
+  are always turned into (unshelved) goals. This can break compatibility in
+  some cases (`#7825 <https://github.com/coq/coq/pull/7825>`_, by Matthieu
+  Sozeau, with help from Maxime Dénès, review by Pierre-Marie Pédrot and
+  Enrico Tassi, fixes `#4095 <https://github.com/coq/coq/issues/4095>`_ and
+  `#4413 <https://github.com/coq/coq/issues/4413>`_).
--- a/engine/evarutil.mli
+++ b/engine/evarutil.mli
@@ -112,6 +112,10 @@ val gather_dependent_evars : evar_map -> Evar.t list -> (Evar.Set.t option) Evar
     solved. *)
 val advance : evar_map -> Evar.t -> Evar.t option
 
+(** [reachable_from_evars sigma seeds] computes the descendents of
+    evars in [seeds] by restriction or evar-evar unifications in [sigma]. *)
+val reachable_from_evars : evar_map -> Evar.Set.t -> Evar.Set.t
+
 (** The following functions return the set of undefined evars
     contained in the object, the defined evars being traversed.
     This is roughly a combination of the previous functions and
@@ -234,8 +238,8 @@ exception ClearDependencyError of Id.t * clear_dependency_error * GlobRef.t opti
 
 (** Restrict an undefined evar according to a (sub)filter and candidates.
     The evar will be defined if there is only one candidate left,
-@raise ClearDependencyError NoCandidatesLeft if the filter turns the candidates
-  into an empty list. *)
+    @raise ClearDependencyError NoCandidatesLeft if the filter turns the candidates
+    into an empty list. *)
 
 val restrict_evar : evar_map -> Evar.t -> Filter.t ->
   ?src:Evar_kinds.t Loc.located -> constr list option -> evar_map * Evar.t
--- a/engine/evd.ml
+++ b/engine/evd.ml
@@ -453,7 +453,7 @@ end
 
 type evar_flags =
   { obligation_evars : Evar.Set.t;
-    restricted_evars : Evar.t Evar.Map.t;
+    aliased_evars : Evar.t Evar.Map.t;
     typeclass_evars : Evar.Set.t }
 
 type side_effect_role =
@@ -631,7 +631,7 @@ let add_with_name ?name ?(typeclass_candidate = true) d e i = match i.evar_body 
     associated to an evar, so we prevent registering its typeclass status. *)
 let add d e i = add_with_name ~typeclass_candidate:false d e i
 
-(*** Evar flags: typeclasses, restricted or obligation flag *)
+(*** Evar flags: typeclasses, aliased or obligation flag *)
 
 let get_typeclass_evars evd = evd.evar_flags.typeclass_evars
 
@@ -659,29 +659,28 @@ let is_obligation_evar evd evk =
 let inherit_evar_flags evar_flags evk evk' =
   let evk_typeclass = Evar.Set.mem evk evar_flags.typeclass_evars in
   let evk_obligation = Evar.Set.mem evk evar_flags.obligation_evars in
-  if not (evk_obligation || evk_typeclass) then evar_flags
-  else
-    let typeclass_evars =
-      if evk_typeclass then
-        let typeclass_evars = Evar.Set.remove evk evar_flags.typeclass_evars in
-        Evar.Set.add evk' typeclass_evars
-      else evar_flags.typeclass_evars
-    in
-    let obligation_evars =
-      if evk_obligation then
-        let obligation_evars = Evar.Set.remove evk evar_flags.obligation_evars in
-        Evar.Set.add evk' obligation_evars
-      else evar_flags.obligation_evars
-    in
-    { evar_flags with obligation_evars; typeclass_evars }
+  let aliased_evars = Evar.Map.add evk evk' evar_flags.aliased_evars in
+  let typeclass_evars =
+    if evk_typeclass then
+      let typeclass_evars = Evar.Set.remove evk evar_flags.typeclass_evars in
+      Evar.Set.add evk' typeclass_evars
+    else evar_flags.typeclass_evars
+  in
+  let obligation_evars =
+    if evk_obligation then
+      let obligation_evars = Evar.Set.remove evk evar_flags.obligation_evars in
+      Evar.Set.add evk' obligation_evars
+    else evar_flags.obligation_evars
+  in
+  { obligation_evars; aliased_evars; typeclass_evars }
 
 (** Removal: in all other cases of definition *)
 
 let remove_evar_flags evk evar_flags =
   { typeclass_evars = Evar.Set.remove evk evar_flags.typeclass_evars;
     obligation_evars = Evar.Set.remove evk evar_flags.obligation_evars;
-    (* Restriction information is kept. *)
-    restricted_evars = evar_flags.restricted_evars }
+    (* Aliasing information is kept. *)
+    aliased_evars = evar_flags.aliased_evars }
 
 (** New evars *)
 
@@ -809,7 +808,7 @@ let create_evar_defs sigma = { sigma with
 
 let empty_evar_flags =
   { obligation_evars = Evar.Set.empty;
-    restricted_evars = Evar.Map.empty;
+    aliased_evars = Evar.Map.empty;
     typeclass_evars = Evar.Set.empty }
 
 let empty_side_effects = {
@@ -872,70 +871,11 @@ let evar_source evk d = (find d evk).evar_source
 let evar_ident evk evd = EvNames.ident evk evd.evar_names
 let evar_key id evd = EvNames.key id evd.evar_names
 
-let define_aux def undef evk body =
-  let oldinfo =
-    try EvMap.find evk undef
-    with Not_found ->
-      if EvMap.mem evk def then
-        anomaly ~label:"Evd.define" (Pp.str "cannot define an evar twice.")
-      else
-        anomaly ~label:"Evd.define" (Pp.str "cannot define undeclared evar.")
-  in
-  let () = assert (oldinfo.evar_body == Evar_empty) in
-  let newinfo = { oldinfo with evar_body = Evar_defined body } in
-  EvMap.add evk newinfo def, EvMap.remove evk undef
+let get_aliased_evars evd = evd.evar_flags.aliased_evars
 
-(* define the existential of section path sp as the constr body *)
-let define_gen evk body evd evar_flags =
-  let (defn_evars, undf_evars) = define_aux evd.defn_evars evd.undf_evars evk body in
-  let last_mods = match evd.conv_pbs with
-  | [] ->  evd.last_mods
-  | _ -> Evar.Set.add evk evd.last_mods
-  in
-  let evar_names = EvNames.remove_name_defined evk evd.evar_names in
-  { evd with defn_evars; undf_evars; last_mods; evar_names; evar_flags }
-
-(** By default, the obligation and evar tag of the evar is removed *)
-let define evk body evd =
-  let evar_flags = remove_evar_flags evk evd.evar_flags in
-  define_gen evk body evd evar_flags
-
-(** In case of an evar-evar solution, the flags are inherited *)
-let define_with_evar evk body evd =
-  let evk' = fst (destEvar body) in
-  let evar_flags = inherit_evar_flags evd.evar_flags evk evk' in
-  define_gen evk body evd evar_flags
-
-let is_restricted_evar evd evk =
-  try Some (Evar.Map.find evk evd.evar_flags.restricted_evars)
+let is_aliased_evar evd evk =
+  try Some (Evar.Map.find evk evd.evar_flags.aliased_evars)
   with Not_found -> None
-
-let declare_restricted_evar evar_flags evk evk' =
-  { evar_flags with restricted_evars = Evar.Map.add evk evk' evar_flags.restricted_evars }
-
-(* In case of restriction, we declare the restriction and inherit the obligation
-   and typeclass flags. *)
-
-let restrict evk filter ?candidates ?src evd =
-  let evk' = new_untyped_evar () in
-  let evar_info = EvMap.find evk evd.undf_evars in
-  let id_inst = Filter.filter_list filter evar_info.evar_hyps.env_named_var in
-  let evar_info' =
-    { evar_info with evar_filter = filter;
-      evar_candidates = candidates;
-      evar_source = (match src with None -> evar_info.evar_source | Some src -> src);
-      evar_identity = Identity.make id_inst;
-    } in
-  let last_mods = match evd.conv_pbs with
-  | [] ->  evd.last_mods
-  | _ -> Evar.Set.add evk evd.last_mods in
-  let evar_names = EvNames.reassign_name_defined evk evk' evd.evar_names in
-  let body = mkEvar(evk',id_inst) in
-  let (defn_evars, undf_evars) = define_aux evd.defn_evars evd.undf_evars evk body in
-  let evar_flags = declare_restricted_evar evd.evar_flags evk evk' in
-  let evar_flags = inherit_evar_flags evar_flags evk evk' in
-  { evd with undf_evars = EvMap.add evk' evar_info' undf_evars;
-    defn_evars; last_mods; evar_names; evar_flags }, evk'
 
 let downcast evk ccl evd =
   let evar_info = EvMap.find evk evd.undf_evars in
@@ -1223,6 +1163,73 @@ let pr_shelf evd =
   let open Pp in
   if List.is_empty evd.shelf then str"(empty stack)"
   else prlist_with_sep (fun () -> str"||") (prlist_with_sep spc Evar.print) evd.shelf
+
+let define_aux def undef evk body =
+  let oldinfo =
+    try EvMap.find evk undef
+    with Not_found ->
+      if EvMap.mem evk def then
+        anomaly ~label:"Evd.define" (Pp.str "cannot define an evar twice.")
+      else
+        anomaly ~label:"Evd.define" (Pp.str "cannot define undeclared evar.")
+  in
+  let () = assert (oldinfo.evar_body == Evar_empty) in
+  let newinfo = { oldinfo with evar_body = Evar_defined body } in
+  EvMap.add evk newinfo def, EvMap.remove evk undef
+
+(* define the existential of section path sp as the constr body *)
+let define_gen evk body evd evar_flags =
+  let (defn_evars, undf_evars) = define_aux evd.defn_evars evd.undf_evars evk body in
+  let last_mods = match evd.conv_pbs with
+  | [] ->  evd.last_mods
+  | _ -> Evar.Set.add evk evd.last_mods
+  in
+  let evar_names = EvNames.remove_name_defined evk evd.evar_names in
+  { evd with defn_evars; undf_evars; last_mods; evar_names; evar_flags }
+
+(** By default, the obligation and evar tag of the evar is removed *)
+let define evk body evd =
+  let evar_flags = remove_evar_flags evk evd.evar_flags in
+  define_gen evk body evd evar_flags
+
+(** In case of an evar-evar solution, the flags are inherited *)
+let define_with_evar evk body evd =
+  let evk' = fst (destEvar body) in
+  let evar_flags = inherit_evar_flags evd.evar_flags evk evk' in
+  let evd = unshelve evd [evk] in
+  let future_goals = FutureGoals.remove evk evd.future_goals in
+  let evd = { evd with future_goals } in
+  define_gen evk body evd evar_flags
+
+(* In case of restriction, we declare the aliasing and inherit the obligation
+   and typeclass flags. *)
+
+let restrict evk filter ?candidates ?src evd =
+  let evk' = new_untyped_evar () in
+  let evar_info = EvMap.find evk evd.undf_evars in
+  let id_inst = Filter.filter_list filter evar_info.evar_hyps.env_named_var in
+  let evar_info' =
+    { evar_info with evar_filter = filter;
+      evar_candidates = candidates;
+      evar_source = (match src with None -> evar_info.evar_source | Some src -> src);
+      evar_identity = Identity.make id_inst;
+    } in
+  let last_mods = match evd.conv_pbs with
+  | [] ->  evd.last_mods
+  | _ -> Evar.Set.add evk evd.last_mods in
+  let evar_names = EvNames.reassign_name_defined evk evk' evd.evar_names in
+  let body = mkEvar(evk',id_inst) in
+  let (defn_evars, undf_evars) = define_aux evd.defn_evars evd.undf_evars evk body in
+  let evar_flags = inherit_evar_flags evd.evar_flags evk evk' in
+  let evd = { evd with undf_evars = EvMap.add evk' evar_info' undf_evars;
+    defn_evars; last_mods; evar_names; evar_flags }
+  in
+  (* Mark new evar as future goal, removing previous one,
+    circumventing Proofview.advance but making Proof.run_tactic catch these. *)
+  let evd = unshelve evd [evk] in
+  let evd = remove_future_goal evd evk in
+  let evd = declare_future_goal evk' evd in
+  (evd, evk')
 
 (**********************************************************)
 (* Accessing metas *)
--- a/proofs/refine.ml
+++ b/proofs/refine.ml
@@ -75,6 +75,8 @@ let generic_refine ~typecheck f gl =
   let future_goals, sigma = Evd.pop_future_goals sigma in
   (* Select the goals *)
   let future_goals = Evd.FutureGoals.map_filter (Proofview.Unsafe.advance sigma) future_goals in
+  let shelf = Evd.shelf sigma in
+  let future_goals = Evd.FutureGoals.filter (fun ev -> not @@ List.mem ev shelf) future_goals in
   (* Proceed to the refinement *)
   let sigma = match Proofview.Unsafe.advance sigma self with
   | None ->
--- a/tactics/tacticals.ml
+++ b/tactics/tacticals.ml
@@ -695,6 +695,8 @@ module New = struct
   (* Check that holes in arguments have been resolved *)
 
   let check_evars env sigma extsigma origsigma =
+    let reachable = lazy (Evarutil.reachable_from_evars sigma
+                            (Evar.Map.domain (Evd.undefined_map origsigma))) in
     let rec is_undefined_up_to_restriction sigma evk =
       if Evd.mem origsigma evk then None else
       let evi = Evd.find sigma evk in
@@ -710,7 +712,12 @@ module New = struct
     let rest =
       Evd.fold_undefined (fun evk evi acc ->
         match is_undefined_up_to_restriction sigma evk with
-        | Some (evk',evi) -> (evk',evi)::acc
+        | Some (evk',evi) ->
+           (* If [evk'] descends from [evk] which descends itself from
+              an originally undefined evar in [origsigma], it is a not
+              a fresh undefined hole from [sigma]. *)
+           if Evar.Set.mem evk (Lazy.force reachable) then acc
+           else (evk',evi)::acc
         | _ -> acc)
         extsigma []
     in
--- a/tactics/tactics.ml
+++ b/tactics/tactics.ml
@@ -5184,14 +5184,14 @@ end
 
 (** Tacticals defined directly in term of Proofview *)
 module New = struct
-  open Genredexpr
-  open Locus
-
   let reduce_after_refine =
-    reduce
-      (Lazy {rBeta=true;rMatch=true;rFix=true;rCofix=true;
-             rZeta=false;rDelta=false;rConst=[]})
-      {onhyps = Some []; concl_occs = AllOccurrences }
+    (* For backward compatibility reasons, we do not contract let-ins, but we unfold them. *)
+    let redfun env t =
+      let open CClosure in
+      let flags = RedFlags.red_add_transparent allnolet TransparentState.empty in
+      clos_norm_flags flags env t
+    in
+    reduct_in_concl ~check:false (redfun,DEFAULTcast)
 
   let refine ~typecheck c =
     Refine.refine ~typecheck c <*>
--- a/engine/evarutil.ml
+++ b/engine/evarutil.ml
@@ -516,12 +516,7 @@ let restrict_evar evd evk filter ?src candidates =
   let candidates = Option.map (filter_effective_candidates evd evar_info filter) candidates in
   match candidates with
   | Some [] -> raise (ClearDependencyError (*FIXME*)(Id.of_string "blah", (NoCandidatesLeft evk), None))
-  | _ ->
-     let evd, evk' = Evd.restrict evk filter ?candidates ?src evd in
-     (* Mark new evar as future goal, removing previous one,
-        circumventing Proofview.advance but making Proof.run_tactic catch these. *)
-     let evd = Evd.remove_future_goal evd evk in
-     (Evd.declare_future_goal evk' evd, evk')
+  | _ -> Evd.restrict evk filter ?candidates ?src evd
 
 let rec check_and_clear_in_constr env evdref err ids global c =
   (* returns a new constr where all the evars have been 'cleaned'
@@ -703,9 +698,21 @@ let rec advance sigma evk =
   match evi.evar_body with
   | Evar_empty -> Some evk
   | Evar_defined v ->
-      match is_restricted_evar sigma evk with
+      match is_aliased_evar sigma evk with
       | Some evk -> advance sigma evk
       | None -> None
+
+let reachable_from_evars sigma evars =
+  let aliased = Evd.get_aliased_evars sigma in
+  let rec search evk visited =
+    if Evar.Set.mem evk visited then visited
+    else
+      let visited = Evar.Set.add evk visited in
+      match Evar.Map.find evk aliased with
+      | evk' -> search evk' visited
+      | exception Not_found -> visited
+  in
+  Evar.Set.fold (fun evk visited -> search evk visited) evars Evar.Set.empty
 
 (** The following functions return the set of undefined evars
     contained in the object, the defined evars being traversed.
--- a/engine/evd.mli
+++ b/engine/evd.mli
@@ -284,8 +284,11 @@ val restrict : Evar.t-> Filter.t -> ?candidates:econstr list ->
     possibly limiting the instances to a set of candidates (candidates
     are filtered according to the filter) *)
 
-val is_restricted_evar : evar_map -> Evar.t -> Evar.t option
-(** Tell if an evar comes from restriction of another evar, and if yes, which *)
+val get_aliased_evars : evar_map -> Evar.t Evar.Map.t
+(** The map of aliased evars *)
+
+val is_aliased_evar : evar_map -> Evar.t -> Evar.t option
+(** Tell if an evar has been aliased to another evar, and if yes, which *)
 
 val set_typeclass_evars : evar_map -> Evar.Set.t -> evar_map
 (** Mark the given set of evars as available for resolution.
@@ -387,7 +390,6 @@ val push_future_goals : evar_map -> evar_map
 val pop_future_goals : evar_map -> FutureGoals.t * evar_map
 
 val fold_future_goals : (evar_map -> Evar.t -> evar_map) -> evar_map -> evar_map
-
 
 val remove_future_goal : evar_map -> Evar.t -> evar_map
 
--- a/engine/proofview_monad.ml
+++ b/engine/proofview_monad.ml
@@ -202,6 +202,11 @@ module type State = sig
   val modify : (t->t) -> unit Logical.t
 end
 
+module type Reader = sig
+  type t
+  val get : t Logical.t
+end
+
 module type Writer = sig
   type t
   val put : t -> unit Logical.t
--- a/engine/proofview_monad.mli
+++ b/engine/proofview_monad.mli
@@ -79,7 +79,7 @@ val with_empty_state : goal -> goal_with_state
 val map_goal_with_state : (goal -> goal) -> goal_with_state -> goal_with_state
 
 (** Type of proof views: current [evar_map] together with the list of
-    focused goals. *)
+    focused goals, locally shelved goals and globally shelved goals. *)
 type proofview = {
   solution : Evd.evar_map;
   comb : goal_with_state list;
@@ -114,6 +114,10 @@ module type State = sig
   val get : t Logical.t
   val set : t -> unit Logical.t
   val modify : (t->t) -> unit Logical.t
+end
+module type Reader = sig
+  type t
+  val get : t Logical.t
 end
 
 module type Writer = sig
--- a/engine/termops.ml
+++ b/engine/termops.ml
@@ -310,15 +310,15 @@ let pr_evar_map_gen with_univs pr_evars env sigma =
 
 let pr_evar_list env sigma l =
   let open Evd in
-  let pr_restrict ev =
-    match is_restricted_evar sigma ev with
+  let pr_alias ev =
+    match is_aliased_evar sigma ev with
     | None -> mt ()
-    | Some ev' -> str " (restricted to " ++ Evar.print ev' ++ str ")"
+    | Some ev' -> str " (aliased to " ++ Evar.print ev' ++ str ")"
   in
   let pr (ev, evi) =
     h 0 (Evar.print ev ++
       str "==" ++ pr_evar_info env sigma evi ++
-      pr_restrict ev ++
+      pr_alias ev ++
       (if evi.evar_body == Evar_empty
        then str " {" ++ pr_existential_key sigma ev ++ str "}"
        else mt ()))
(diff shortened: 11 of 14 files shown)


engine/evarutil.mli
```diff
@@ -226,6 +230,10 @@ type clear_dependency_error =
 
 exception ClearDependencyError of Id.t * clear_dependency_error * GlobRef.t option
 
+(** Marks an evar that has been defined by another evar by projection.
+    Used to handle checking of created goals correctly. *)
+val evar_evar_solution : bool Store.field
```
ppedrot: Like the resolvable field, you should use `unit Store.field` to save space on the most common case, which is probably going to be `false`.

mattam82: Done

mattam82:
Rebased and fixed according to @ppedrot's comments

tactics/tacticals.ml
```diff
@@ -516,7 +517,10 @@ module New = struct
     let rest =
       Evd.fold_undefined (fun evk evi acc ->
         match is_undefined_up_to_restriction sigma evk with
-        | Some (evk',evi) -> (evk',evi)::acc
+        | Some (evk',evi) when not (Evd.mem origsigma evk)
```
ppedrot: You shouldn't use `Evd.mem` if what you want to check is that the evar is undefined in the original evar map. Also, I am slightly concerned by the conceptual complexity of this function, which starts applying many heuristics and relies on those fragile flags. Maybe we should just sit down and seriously ponder the whole evar aliasing mess instead.

mattam82:
I rebased this on top of PR #8741 to profit from the existing restricted_evars set defined there. This makes all the evar_info flag manipulations explicit and contained in evd.

mattam82:
Bench: https://ci.inria.fr/coq/view/benchmarking/job/benchmark-part-of-the-branch/554/ (pending)

ppedrot (approved):
LGTM. You should probably rebase and remove the unrelated overlay, I don't think it's a good practice to side-step slow-responding upstream inside individual PRs.

mattam82:
Yes. There is still an issue in iris-lambda to be resolved before this is ready.

engine/evarutil.mli
```diff
@@ -126,6 +126,10 @@ val gather_dependent_evars : evar_map -> Evar.t list -> (Evar.Set.t option) Evar
     solved. *)
 val advance : evar_map -> Evar.t -> Evar.t option
 
+(** [reachable_from_evars sigma seeds evs] computes if one of [evs] is a descendent
+    of an evar of [seeds] by restriction or evar-evar unifications in [sigma]. *)
+val reachable_from_evars : evar_map -> Evar.Set.t -> Evar.Set.t -> bool
```
gares: Maybe it is just a dejavu but I seem to recall we discussed about this API and my impression was that what one wants is
```
val reachable_from_evars : evar_map -> Evar.Set.t -> Evar.Set.t 
```
And the compose the result with `Evar.Set.mem` in order to get the boolean.
Would it be possible to get this API? Or it would be inefficient.

mattam82: I don't remember that solution being discussed but it does make sense. The current implementation can stop early using an exception but basically builds a visited set that could just be completed and returned instead. I can try that as an alternative solution.

gares: My use case is to shelve goals (evars) that are reachable from already existing goals. Hence having the set, rather than a membership test, is probably simpler.

mattam82: Yes, that makes sense. Note that reachable_from_evars just computes a transitive closure by aliasing so it containts the defined evars as well, which is necessary for `check_evars`. It is easy to filter them out if necessary.

mattam82: I just pushed a new commit implementing that

gares (approved):

RalfJung:
@mattam82 Which are these two occurrences?  Also, @robbertkrebbers is our expert for the IPM/MoSeL tactics :)

mattam82:
@RalfJung I’ll have to find them again, it’s where the scripts break with
this branch :)

robbertkrebbers:
I'm trying to figure out what's going on, is there a chance this PR affects the behavior of `subst`?

mattam82:
That’s unlikely, it is more about which calls to non-e variants are allowed
(e.g apply) and which goals they produce.
Le mer. 31 oct. 2018 à 09:49, Robbert Krebbers <notifications@github.com> a
écrit :

robbertkrebbers:
I double checked that I did not do anything wrong, but some of the failures on Iris have to do with subst. It goes wrong here:

https://gitlab.mpi-sws.org/FP/iris-coq/blob/19aae59ad3806d286e6d3a07038da77ba99c29fd/theories/program_logic/adequacy.v#L90

The `simplify_eq` performs `subst` (among other things); however `subst` is failing. The goal after the `simplify_eq` is:

```
Irrelevant hypotheses
H2 : t1' = e1' :: t1''
"H" : ▷ (|={∅,⊤}=> state_interp σ1' (κs0 ++ κs') (Init.Nat.pred (length t1'))
                   ∗ WP e1' @ s; ⊤ {{ v, Φ v }} ∗ wptp s t1'')
--------------------------------------∗
|={∅,∅,⊤}▷=> |={⊤,∅}▷=>^n ∃ (e2 : expr Λ) (t2' : list (expr Λ)),
                            ⌜t2 = e2 :: t2'⌝
                            ∗ state_interp σ2 κs' (Init.Nat.pred (length t2))
                              ∗ WP e2 @ s; ⊤ {{ v, Φ v }} ∗ wptp s t2'
```

Now calling `subst` doesn't do anything, and calling `subst t1'` results in:

```
Ltac call to "subst (ne_var_list)" failed.
Cannot find any non-recursive equality over t1'.
```

With the commit on which this MR is based (0ac673e562c34245e4e48efc428d808e917be79b) the `subst` worked like it used to be.

I tried to create a minimal example, but failed.

robbertkrebbers:
For the regression in Iris's `atomic.v` I found the following minimal example:

```coq
Definition Bar (P P' : Prop) := P -> P'.
Lemma bar_id P : Bar P P.
Proof. unfold Bar. auto. Qed.

Goal exists P, Bar P P.
eexists. notypeclasses refine (bar_id _).
(* I get an explicit goal `Prop` in this PR *)
(* While the `Prop` is shelved in master *)
```

Since the `Prop` in question is already on the shelve, I see no reason why it should be unshelved by the above `refine`. As such, I would say this PR introduces an unwanted regression.

mattam82:
Looking at your example again, that's debatable: one can also understand that when doing `refine (bar_id _)`, the `_` is turned into a new goal, which solves the shelved variable. I'm wondering what others think about this, @aspiwack and @herbelin in particular?

mattam82:
@herbelin @aspiwack any idea here?

aspiwack:
Sorry I let that slide earlier. I'm having trouble understanding what's
happening in this discussion thread. Could you summarize the question?

mattam82:
The question might reduce to: if ?X1 is a shelved evar in your goal, and you apply a lemma which unifies it with a new evar ?X2 (through an input "_" for example), shall it turn into a new goal or stay shelved?

robbertkrebbers:
My opinion: I'm very much in favor of it being shelved. 

-The original evar was shelved for a reason in the first place, so it's strange that by applying a lemma that evar (well, an evar that's unified to it) becomes unshelved out of the sudden.
- In the case of Iris, this issue (evar being unshelved) happens during type class search. Type class should never create unshelved goals.

mattam82:
Ok. Just note that it's non-uniform behavior, if you refine with `id _` instead of `_` then you'll get a new subgoal and make the shelved evar disappear.

robbertkrebbers:
> In the case of Iris, this issue (evar being unshelved) happens during type class search. Type class should never create unshelved goals.

I took a look at the Iris issue again; the creation of the evar did not happen due to type class search, but due to applying a type class instance by hand using `notypeclasses refine`. Indeed, in the following the goal **is** shelved:

```
Class Bar (P P' : Prop) := bar : P -> P'.
Instance bar_id P : Bar P P.
Proof. unfold Bar. auto. Qed.

Goal exists P, Bar P P.
eexists. apply _.
```

So, since the issue is not part of type class search, it turns that Iris can easily be adapted to this MR (modulo the `subst` bug that I mentioned in https://github.com/coq/coq/pull/7825#issuecomment-435183484).

Hence, as long as we agree that type class search should never create shelved goals (which appeared this MR had no effect on), then we're good no matter what choice IMHO.

mattam82:
Ok. That's fine, dependent subgoals will always be shelved in typeclass resolution, unless you explicitely do a `notypeclasses refine (bar_id _)` in a `Hint Extern` to explicitely produce a goal for `_`, consistently with the change in this PR. And this is similarly taken care of in the following unifall-apply PR: when applying a typeclass instance `foo : forall (x : A). P x`, the goal corresponding to `x`, if any, is shelved explicitly.

I still have to figure out the subst problem to make progress here.

robbertkrebbers:
> Ok. That's fine, dependent subgoals will always be shelved in typeclass resolution

Good, then I I'm fine with the proposed change. I have pushed a fix to Iris: https://gitlab.mpi-sws.org/FP/iris-coq/commit/394075ccaa09678d9707afd1e533d136f9747375

> I still have to figure out the subst problem to make progress here.

Alright, let me know. once that's done. Then I can check if there are other regressions in the Iris-related projects.

ppedrot:
@mattam82 What's the status of this?

maximedenes:
@mattam82 Should https://github.com/coq/coq/pull/991 be rebased on this?

maximedenes:
I'll rebase this.

maximedenes:
> I still have to figure out the subst problem to make progress here.

OMG, took me ages to realize what is happening: the hypothesis that `subst` is supposed to substitute now has an invisible cast to `Prop`. Of course, `subst` is stupid enough that it does not recognize the equality below the cast.

mattam82:
@maximedenes I think I've searched for a long time too and couldn't figure out why the hell this cast was introduced in the first place, did you find out?

maximedenes:
> I think I've searched for a long time too and couldn't figure out why the hell this cast was introduced in the first place, did you find out?

All I know is that it is a `REVERTcast` introduced by `convert_concl`. It could be from a `Hint Unfold`, for example. How do I get a resolution trace showing the hints that are applied?

maximedenes:
Note: I found that the cast is inserted by `reduce_after_refine`. After discussing with @mattam82 and @herbelin, we decided to call reduction functions directly, without inserting a cast, in this case.

maximedenes (commented):
IIUC, this PR is changing the behavior of the low-level `refine`. More precisely, when an existential variable introduced in the arguments of `refine` is unified with a pre-existing evar, it inherits some of its associated data. It makes aliased evars (solutions of evar/evar problems) be treated like "restricted" evars. It is not clear to me why we couldn't do that by removing the special handling of restricted variables, rather than adding aliased evars there.

More concrelely, why do we need to keep an `aliased_evars` map, rather than using the evar map to define the reachability relation?

maximedenes:
The fiat-crypto-legacy failure is very hard to debug, even with full traces to compare. @JasonGross do you think you could help us on understanding what changed behavior?

The only thing I see is a new goal with the following shape: `H := ..., ... |- H`. Not sure what generated it.

JasonGross:
@maximedenes Sorry for the delay; I missed this.  (If I'm non-responsive for too long in the future, feel free to ping me again, or email me.)

I don't have time to dig into it much now (I'll try to make time sometime this week), but my current guess is that there's a change somewhere in the behavior of `shelve_unifiable` or `simple refine` with `open_constr`?
Where does the new goal show up?  In particular, does the new goal show up (a) before the call to `Compilers.Z.Bounds.Pipeline.ReflectiveTactics.do_reflective_pipeline`, (b) before the subcall to `Compilers.Z.Bounds.Pipeline.ReflectiveTactics.solve_side_conditions`, (c) as a result of `do_reify` in `solve_side_conditions`, or (d) somehow in one of the branches of `solve_post_reified_side_conditions`?  If it's (a), it's going to be a bit of a pain to track down, as it's probably somewhere in `refine_to_reflective_glue` in `src/Compilers/Z/Bounds/Pipeline/Glue.v`...

maximedenes:
After a painful exploration of millions of Ltac debug trace, @mattam82 and I concluded that the change of semantics in fiat-crypto-legacy was expected. I added an overlay.

maximedenes:
@mattam82 the compcert error is due to the a change in the order of subgoals produced by refine. Is this a known effect of the patch?

maximedenes:
> the compcert error is due to the a change in the order of subgoals produced by refine. Is this a known effect of the patch?

In fact, the new order corresponds to the order in which the corresponding holes appear in the term, so it looks like a bug fix.

doc/changelog/02-specification-language/07825-rechable-from-evars.rst
```diff
@@ -0,0 +1,8 @@
+- In `refine`, new existential variables unified with existing ones are no
+  longer considered as fresh. The behavior of `simple refine` no longer depend on
+  the orientation of evar-evar unification problems, new existential variables
```
jfehrle:
```suggestion
  the orientation of evar-evar unification problems, and new existential variables
```

Also, is "orientation" the right word here?  I wasn't sure what was meant, but that may just be my lack of knowledge.

jfehrle (commented):
Wording suggestions

doc/changelog/02-specification-language/07825-rechable-from-evars.rst
JasonGross: I think "orientation" is correct; to me it means "left-to-right" vs "right-to-left", i.e., the difference between `unify x y` and `unify y x` when you need to bias one side over the other (e.g., if you're unifying two evars and need to decide which one gets filled with the other)

doc/changelog/02-specification-language/07825-rechable-from-evars.rst
```diff
@@ -0,0 +1,8 @@
+- In `refine`, new existential variables unified with existing ones are no
```
Zimmi48:
```suggestion
- **Changed:**
  In :tacn:`refine`, new existential variables unified with existing ones are no
```

doc/changelog/02-specification-language/07825-rechable-from-evars.rst
```diff
@@ -0,0 +1,8 @@
...
+  longer considered as fresh. The behavior of `simple refine` no longer depend on
```
Zimmi48:
```suggestion
  longer considered as fresh. The behavior of :tacn:`simple refine` no longer depends on
```

maximedenes:
@Zimmi48 it seems I can't apply your suggestion, GH says I don't have push rights, which can't be true, I pushed several times to this branch :(

I guess @mattam82 should be able to accept the changes.

maximedenes:
@ppedrot All green! Finally :)

doc/changelog/02-specification-language/07825-rechable-from-evars.rst
jfehrle: "left-to-right" and "right-to-left" suggest "direction" as a useful wording.  Your choice.

JasonGross:
I am very confused by the behavior of Coq after this PR.  Consider:
```coq
Record T (x : nat) := { t : x = x }.

Goal exists x, T x.
  eexists.
  simple refine {| t := _ |}.
  (* 2 focused subgoals
(shelved: 1) (ID 5)

  ============================
  nat

subgoal 2 (ID 10) is:
 ?x = ?x
*)
  Unshelve.
  (* 2 subgoals (ID 10)

  ============================
  ?x = ?x

subgoal 2 (ID 5) is:
 nat
*)
```
Why am I told that there is a shelved goal after `simple refine`, and why does `Unshelve` swap the order of my goals?

JasonGross:
My sense is that this change is a step backwards in the usability of `simple refine`.  Why is it better this way?  For example, previously `simple refine (existT _ _ _)` would always create exactly two goals.  Now it will sometimes create two goals, sometimes create three goals, and sometimes create four goals, depending on whether or not the arguments to `sigT` are evars.

maximedenes:
> Why am I told that there is a shelved goal after `simple refine`

This looks strange indeed. @mattam82

> Now it will sometimes create two goals, sometimes create three goals, and sometimes create four goals, depending on whether or not the arguments to `sigT` are evars.

Really? Do you have an example?

@mattam82 I did notice a regression:

```
Record T (x : nat) := { t : x = x }.

Goal exists x, T x.
  eexists.
  refine {| t := _ |}.
```

now produces `?Goal = ?Goal`, instead of `?x = ?x` previously.

mattam82:
There's a bug lurking somewhere:
```
  (* 2 focused subgoals
(shelved: 1) (ID 5)

  ============================
  nat

subgoal 2 (ID 10) is:
 ?x = ?x
*)
```

The (shelved: 1) indicates that the nat goal is both shelved and unshelved, that should be fixed. This explains the subsequent behavior of Unshelve. Regarding the behavior of the `simple refine {| t := _ |}` it makes sense to me: you are refining with `Build_T ?y ?z` on a goal `T ?x`. The number of subgoals is going to always be the number of evars from the refining term `Build_T ?y ?z : T ?y` that are not unified with non-evar terms in the unification problem `T ?y = T ?x`. Here we have an evar aliasing `?x := ?y` or `?y := ?x` and the (new) spec is that the result should be independent of evar unification direction and simple refine should not shelve goals. You hence get goals for `?y` and `?z`.

JasonGross:
> Really? Do you have an example?

@maximedenes Consider
```coq
Goal { A : _ & { P : _ & @sigT A P } }.
  epose _ as A.
  epose _ as P.
  exists A, P.
```
In this state, `simple refine (existT _ _ _)` should give two goals.  If I first `subst P`, I get three goals.  If I instead `subst A P`, then `simple refine (existT _ _ _)` gives four subgoals, I believe.

> The number of subgoals is going to always be the number of evars from the refining term `Build_T ?y ?z : T ?y` that are not unified with non-evar terms in the unification problem `T ?y = T ?x`.

This behavior makes `simple refine` effectively unusable in automation, because it means that my automation needs to basically do unification by hand just to figure out how many subgoals remain, and there is literally no support in Ltac (other than `..`) for making tactics that can do different things depending on how many goals there are.  If the evar terms are not all in a row (and thus amenable to being shelved with `..`).  I think literally the only thing automation can do to be robust here is to `set` all of the evars before `simple refine` and `subst` them after.  This can incur serious performance overhead in large goals with large contexts, and is anyway a kludge that I consider to be an indication of poorly-designed tactics.  (It is something I've used to prevent `rewrite` and `cbv` or `cbn` from messing with evars when they shouldn't.)

I believe the only spec that makes `simple refine` usable in automation is if the number of subgoals is the number of holes which are not unified with *any* term in the unification problem.  Furthermore, I think unification should prefer instantiating new holes with old ones, rather than the other way around.

mattam82:
I don’t think the result should be order dependent, however you have a point that if the original evar is shelved that could be inherited as well. But then the spec still depends on the status of the evars in the conclusion: if you unify _ with an existing evar ?goaln in the conclusion that is a dependent subgoal (not shelved), then do you think no new subgoal should appear or that one should appear, solving ?goaln? On the implementation side, to enforce that no subgoal appears in theses cases we would need Proofview.advance to reason about shelved subgoals: if an evar advances to a shelved subgoal is stays shelved and advance should return None. I think failure to unshelve the goals we advance to in the current version is the reason for the bug you found.

JasonGross:
> however you have a point that if the original evar is shelved that could be inherited as well.

If this fixes the issue I brought up with `sigT` / `existT`, then I'm very strongly in favor.  Note that this should also fix the issue with fiat-crypto legacy without any need for an overlay.

> But then the spec still depends on the status of the evars in the conclusion: if you unify _ with an existing evar ?goaln in the conclusion that is a dependent subgoal (not shelved), then do you think no new subgoal should appear or that one should appear, solving ?goaln?

If I'm understanding correctly, I think the difference between these shouldn't be visible to the user (except in the naming of goal evars perhaps)?  Do you have an example goal + tactic invocation demonstrating the two different behaviors?  (It's easier for me to think about user-visible behavior than about underlying implementation.)

mattam82:
Suppose you have three goals, where the first depends on the third and the second is independent. I do `refine` with a term that instantiates the occurrence of the third goal with an evar (say X) and completes the subgoal. Should there be a new subgoal for X, solving the third subgoal, and hence in order: X and the original second subgoal. Or should the evar X be ‘resolved’ by the third subgoal dependency and then we’ll have, in order the original second subgoal and then the original third goal.

JasonGross:
I think in this case it does not matter too much.  The thing I'd be inclined to do, if it makes sense, is to add an option parameterizing unification and make the default setting whichever one is backwards compatible.  Failing that, I'd default to choosing whichever option most closely matches the previous behavior.  If neither option matches the previous behavior, I think it'd be more intuitive to have the latter option, that "the evar X be ‘resolved’ by the third subgoal dependency and then we’ll have, in order the original second subgoal and then the original third goal.".  I see the following argument as weakly in favor of this choice: existing evars may be encoding useful information, for example by their goal ordering (and fresh holes in an argument to `simple refine` can't encode such information), and reshuffling goals as a side effect of refining is surprising (even if the actual thing going on is old goals being solved by new goals, it appears to the user to be a shuffling).

maximedenes:
@mattam82 so, what should we do? IIUC, we could inherit the shelved status by enriching the `future_goals` data (or moving evarconv to the monad, but that's probably not going to happen soon).

mattam82:
I enriched the future_goal status. Funny thing is, this revealed that #10298 was not properly fixed in 8.11.0: the Enc typeclass was still being resolved previously, because shelved goals being unified to other evars were not tracked correctly by refine. With this patch, the bug is finally fixed.

JasonGross:
Failure in fiat-crypto seems to reveal a genuine bug:
> Error: Anomaly "grounding a non evar-free term"
 Please report at http://coq.inria.fr/bugs/

Perhaps this should be minimized and added to the test-suite?

The f-c-l failure is confusing, idk why it's not finding a target that exists....

JasonGross:
Oh, I bet the f-c-l failure is because this PR has a flat-crypto-legacy overlay that points to a stale commit.  If the behavior mentioned above is fixed, the overlay should no longer be necessary.

SkySkimmer:
Commit message is missing its newlines

mattam82:
Hmm, the fiat-crypto failure is a bit worrrying. I'll investigate... fiat-crypto-legacy fails early due to build issues? @maximedenes any idea on that one?

mattam82:
@JasonGross  Oh I didn't see your comment at first, let me repush without the overlay, we'll see. If you can launch a minimizer for fiat-crypto that would be great

@SkySkimmer indeed

JasonGross:
```coq
(* -*- mode: coq; coq-prog-args: ("-emacs" "-w" "+implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern" "-top" "Func" "-R" "." "Top") -*- *)
(* File reduced by coq-bug-finder from original input, then from 466 lines to 305 lines, then from 312 lines to 47 lines, then from 62 lines to 47 lines, then from 383
lines to 47 lines, then from 911 lines to 116 lines, then from 690 lines to 220 lines, then from 717 lines to 374 lines, then from 815 lines to 517 lines, then from 798 lines to 538 lines, then from 820 lines to 539 lines, then from 781 lines to 539 lines, then from 804 lines to 539 lines, then from 776 lines to 537 lines, then from 766 lines to 537 lines, then from 776 lines to 537 lines, then from 768 lines to 537 lines, then from 791 lines to 537 lines, then from 2308 lines to 538 lines, then from 760 lines to 537 lines, then from 760 lines to 537 lines, then from 1798 lines to 538 lines, then from 1294 lines to 537 lines, then from 1998 lines to 538 lines, then from 929 lines to 792 lines, then from 434 lines to 132 lines, then from 412 lines to 364 lines, then from 208 lines to 96 lines, then from 2149 lines to 143 lines, then from 201 lines to 145 lines, then from 161 lines to 145 lines *)
(* coqc version 8.12+alpha (February 2020) compiled on Feb 10 2020 18:42:28 with OCaml 4.06.1
   coqtop version jgross-Leopard-WS:/home/jgross/Documents/repos/coq2,pr/7825 (98d00cacfaf63e66394398a787a044c42a8ab1a3) *)

Global Set Asymmetric Patterns.

Module Export Language.
  Module Export Compilers.
    Module Export type.
      Inductive type {base_type : Type} := base (t : base_type) | arrow (s d : type).
      Global Arguments type : clear implicits.

      Fixpoint final_codomain {base_type} (t : type base_type) : base_type
        := match t with
           | base t
             => t
           | arrow s d => @final_codomain base_type d
           end.

      Fixpoint for_each_lhs_of_arrow {base_type} (f : type base_type -> Type) (t : type base_type) : Type
        := match t with
           | base t => unit
           | arrow s d => f s * @for_each_lhs_of_arrow _ f d
           end.

    End type.
    Notation type := type.type.
    Bind Scope etype_scope with type.type.
    Infix "->" := type.arrow : etype_scope.
    Module base.
      Module Export type.
        Inductive type {base_type : Type} := type_base (t : base_type) | prod (A B : type) | list (A : type) | option (A : type) | unit.
        Global Arguments type : clear implicits.
      End type.
      Notation type := type.type.

    End base.

    Module Export expr.
      Section with_var.
        Context {base_type : Type}.
        Local Notation type := (type base_type).
        Context {ident : type -> Type}
                {var : type -> Type}.

        Inductive expr : type -> Type :=
        | Ident {t} (idc : ident t) : expr t
        | Var {t} (v : var t) : expr t
        | Abs {s d} (f : var s -> expr d) : expr (s -> d)
        | App {s d} (f : expr (s -> d)) (x : expr s) : expr d
        | LetIn {A B} (x : expr A) (f : var A -> expr B) : expr B
        .
      End with_var.

      Module Export APIExports.
        Module base.
          Module Export type.
            Inductive base := Z.
          End type.
          Notation type := (Language.Compilers.base.type APIExports.base.type.base).
        End base.
        Module Import API.
          Notation type := (Language.Compilers.type APIExports.base.type).
          Axiom ident : type -> Type.
          Notation expr := (@expr APIExports.base.type ident).

          Class parameters :=
            {
            }.
          Axiom base_Z : APIExports.base.type.
          Axiom base_listZ : APIExports.base.type.
          Module Export rep.
            Section rep.
              Context {p : parameters}.

              Class rep (t : base.type) :=
                { ltype : Type;
                  rep_p := p }.
              Instance listZ_local {zrep : rep base_Z} : rep base_listZ.
              Admitted.

              Instance Z : rep base_Z.
              Admitted.
            End rep.
          End rep.

          Section defs.
            Context {p : parameters}

                    {listZ : rep.rep base_listZ}.

            Fixpoint base_ltype (t : base.type) : Type :=
              match t with
              | base.type.prod a b => base_ltype a * base_ltype b
              | _ => rep.ltype (rep:=listZ)
              end.
            Fixpoint ltype (t : type.type base.type) : Type :=
              match t with
              | type.base t => base_ltype t
              | type.arrow s d => unit
              end.

          End defs.
          Context {p : parameters}.

          Axiom dummy_base_ltype
            : forall {p : parameters} {listZ : rep base_listZ}
                     (t : base.type), base_ltype t.
          Axiom base_list_lengths : base.type -> Type.
          Axiom cmd : forall {p : parameters}, Type.
          Axiom translate_cmd
            : forall {p : parameters} {t}
                     (e : API.expr (var:=(@ltype p (@listZ_local p (@Z p)))) (type.base t)) (nextn : nat),
              nat * @base_ltype p (@listZ_local p (@Z p)) t * cmd.

          Axiom cmd_skip : forall {p}, @cmd p.

          Fixpoint list_lengths t : Type :=
            match t with
            | type.base t => base_list_lengths t
            | _ => unit
            end.

          End defs.
          Context {p : parameters}.

          Axiom dummy_base_ltype
            : forall {p : parameters} {listZ : rep base_listZ}
                     (t : base.type), base_ltype t.
          Axiom base_list_lengths : base.type -> Type.
          Axiom cmd : forall {p : parameters}, Type.
          Axiom translate_cmd
            : forall {p : parameters} {t}
                     (e : API.expr (var:=(@ltype p (@listZ_local p (@Z p)))) (type.base t)) (nextn : nat),
              nat * @base_ltype p (@listZ_local p (@Z p)) t * cmd.

          Axiom cmd_skip : forall {p}, @cmd p.

          Fixpoint list_lengths t : Type :=
            match t with
            | type.base t => base_list_lengths t
            | _ => unit
            end.

          Fixpoint translate_func' {t} (e : @API.expr ltype t) (nextn : nat)
            : type.for_each_lhs_of_arrow ltype t ->
              type.for_each_lhs_of_arrow list_lengths t ->
              nat * base_ltype (type.final_codomain t) * cmd :=
            match e with
            | expr.Abs (type.base s) d f =>

              fun (args : base_ltype s * type.for_each_lhs_of_arrow _ d)
                  (lengths : base_list_lengths s * _) =>
                translate_func' (f (fst args)) nextn (snd args) (snd lengths)

            | expr.Ident (type.base b) idc =>
              fun (_ _:unit) => translate_cmd (expr.Ident idc) nextn
            | expr.Var (type.base b) v =>
              fun (_ _:unit) => translate_cmd (expr.Var v) nextn
            | expr.App _ (type.base b) f x =>
              fun (_ _:unit) => translate_cmd (expr.App f x) nextn
            | expr.LetIn _ (type.base b) x f =>
              fun (_ _:unit) => translate_cmd (expr.LetIn x f) nextn

            | _ => fun _ _ => (0%nat, dummy_base_ltype _, cmd_skip)
            end.
(* Error: Anomaly "grounding a non evar-free term"
Please report at http://coq.inria.fr/bugs/. *)
```

JasonGross (dismissed):
Please add this to the test-suite (in addition to the example from fiat-crypto that produces the anomaly) : 
```coq
Goal { A : _ & { P : _ & @sigT A P } }.
  epose _ as A.
  epose _ as P.
  exists A, P.
  (* Regardless of which evars are in the goals vs the hypotheses,
     [simple refine (existT _ _ _)] should leave over two goals *)
  simple refine (existT _ _ _); let n := numgoals in guard n = 2. Undo.
  subst P.
  simple refine (existT _ _ _); let n := numgoals in guard n = 2. Undo.
  subst A.
  simple refine (existT _ _ _); let n := numgoals in guard n = 2. Undo.
Abort.
```

JasonGross:
Fiat-Crypto-Legacy is revealing a genuine bug, I believe.  (Perhaps we were too hasty to remove it.)  I'm working on minimizing a case where `a; b.` leaves over 14 subgoals but `a. all: b.` leaves over 13 sugoals (and the latter is the correct behavior).

JasonGross:
Here is a reduced version of the bug from fiat-crypto-legacy, I believe:
```coq
Goal { A : _ & { P : _ & @sigT A P } }.
  epose _ as A;
  epose _ as P;
  exists A, P;
  (* Regardless of which evars are in the goals vs the hypotheses,
     [simple refine (existT _ _ _)] should leave over two goals.  This
     should be true even when chained with epose. *)
  assert_succeeds (simple refine (existT _ _ _); let n := numgoals in guard n = 2);
  subst P;
  assert_succeeds (simple refine (existT _ _ _); let n := numgoals in guard n = 2);
  subst A;
  assert_succeeds (simple refine (existT _ _ _); let n := numgoals in guard n = 2).
  (* fails *)
Abort.
```
Somehow the number of goals remaining is different depending on whether the `epose`(s) are separated from the subsequent tactics via `;` or via `.`....

JasonGross:
Also, since this touches unification, I think it'd be nice to test fiat-crypto-legacy again after the two issues described above are resolved.

mattam82:
Thanks @JasonGross indeed this catches a discrepancy in the handling of the local shelf

mattam82:
This seems to reveal a deeper problem with the implementation of `epose` and other such tactics that take open terms: they all rely on Proofview.run (which interprets "tactic.") to gather the generated evars at the end of execution and shelve them. It means we don't know the `_` ought to be shelved after `epose _ as A`.

herbelin:
>  they all rely on Proofview.run (which interprets "tactic.") to gather the generated evars at the end of execution and shelve them. It means we don't know the _ ought to be shelved after epose _ as A.

Trying to connect to the context of the issue: do you mean that the `tclWITHHOLES` surrounding the call to `epose` (= `letin_tac`) should somehow be moved in the code of `letin_tac`?

Zimmi48:

maximedenes:
@mattam82 do you have a clear view of what needs to be done for this PR? It would be nice to unblock it, as many other potential fixes are waiting for it. I'd be happy to help, but I'm not sure how we should address the issue revealed by fiat-crypto-legacy.

ejgallego:
A (very) random thought on this PR, if evars are meant to be pointers, when a variable is unified with an older one shouldn't this just set the pointer properly? (The young -> old invariant is important for other reasons in the WAM)

Maybe that's easier than keeping the map, tho `evar_body` should maybe not be exposed in the current form.

maximedenes:
> A (very) random thought on this PR, if evars are meant to be pointers, when a variable is unified with an older one shouldn't this just set the pointer properly? (The young -> old invariant is important for other reasons in the WAM)
> 
> Maybe that's easier than keeping the map, tho `evar_body` should maybe not be exposed in the current form.

At first sight, this seems like a reasonable idea for a new Coq. However, in practice it is going to change the ordering of unification problems, the solutions, etc. So a transition seems really hard.

Note that it is still something that will be easier to explore once unifall is complete and at least we have to worry only about one unification engine, not two and their interactions.

maximedenes:
@Zimmi48 any idea why no Gitlab CI pipieline is shown in the checks here?

maximedenes:
> any idea why no Gitlab CI pipieline is shown in the checks here?

In fact the `pr-7825` branch on gitlab does not seem to contain my last commit.

Zimmi48:
I'll look into it.

SkySkimmer:
>https://github.com/coq/coq/pull/7825#issuecomment-584784896

This may be linked to https://github.com/coq/coq/issues/12910

mattam82:
> A (very) random thought on this PR, if evars are meant to be pointers, when a variable is unified with an older one shouldn't this just set the pointer properly? (The young -> old invariant is important for other reasons in the WAM)
> 
> Maybe that's easier than keeping the map, tho `evar_body` should maybe not be exposed in the current form.

Aliasing would not be in general preserved by nf_evar'ing the bodies of evar_info's, so it's nice to keep it separate.
While there is a points-to relation here, the contexts/types of evars might change in ways we want to remember. Also, we have no use yet of the young -> old relationship, and I don't know what the WAM is :)

ejgallego:
> Aliasing would not be in general preserved by nf_evar'ing the bodies of evar_info's, so it's nice to keep it separate.
> While there is a points-to relation here, the contexts/types of evars might change in ways we want to remember. Also, we have no use yet of the young -> old relationship, and I don't know what the WAM is :)

WAM = Warren Abstract Machine; indeed that'd rely on having evars be truly pointers, copying a term as in `nf_evar` is indeed a showstopper, tho I never understood why that copy is needed in the first place.

maximedenes:
> WAM = Warren Abstract Machine; indeed that'd rely on having evars be truly pointers, copying a term as in `nf_evar` is indeed a showstopper, tho I never understood why that copy is needed in the first place.

But what if young := old is ill-typed?

ejgallego:
> But what if young := old is ill-typed?

Not familiar enough with the internals, but coming from an evar~evar problem that should be the same whether you update the evar_map or use the trail ? [Depending on the representation having the map could be fine I guess]

maximedenes:
@mattam82 the ci-rewriter failure shows that run_on_evars complains about some evar leaks. What's a typical way for resolution to leak evars?

mattam82:
> @mattam82 the ci-rewriter failure shows that run_on_evars complains about some evar leaks. What's a typical way for resolution to leak evars?

It seems to mean in this case that typeclass resolution produced an evar map with undefined evars that are not shelved, and not goals either, otherwise resolution would have failed. This means that there are dangling evars somewhere, probably introduced by some tactic that does not register its future goals somehow, or due to the change of handling of future goals, this check is now incorrect (e.g. it might be evars that are older in the stack of future goals, and which should be rather considered ok).

maximedenes:
> (e.g. it might be evars that are older in the stack of future goals, and which should be rather considered ok)

Ah ah, seems like an interesting possibility. Will investigate more, thanks!

JasonGross:
If you need help tracing the tactic execution in the rewriter, let me know.

maximedenes:
@JasonGross Actually, it fails on a definition (`VerifiedRewriter_of_Rewriter` in `AllTactics`). But the typeclass resolution trace is a bit sparse, so any information is welcome.

maximedenes:
Making the type of one argument is enough to make it succeed: `{pkg_proofs : @package_proofs _ _ pkg}`.

JasonGross:
@maximedenes That is strange, because `package` (the type of `pkg`) is a class, so it should be picked up automatically by typeclass resolution as it's in context.  In 8.11, the following `Check` has all evars resolved:
```coq
    Set Printing All.
    Check forall {exprInfo : ExprInfoT}
                 {exprExtraInfo : ExprExtraInfoT}
                 {pkg : package}
                 {pkg_proofs : package_proofs}
                 (R : RewriterT),
        True.
    (* forall (exprInfo : ExprInfoT) (exprExtraInfo : @ExprExtraInfoT exprInfo)
  (pkg : @package (@base exprInfo) (@ident exprInfo))
  (_ : @package_proofs (@base exprInfo) (@ident exprInfo) pkg)
  (_ : @RewriterT exprInfo exprExtraInfo pkg), True
     : Prop
*)
```
I assume that you get unresolved evars on this `Check` with this branch?
(Note, btw, that the arguments to the type of `pkg` are forced by the inference of `pkg` as an argument to `RewriterT`; if we drop the `RewriterT`, then we get:
```coq
    Set Printing All.
    Check forall {exprInfo : ExprInfoT}
                 {exprExtraInfo : ExprExtraInfoT}
                 {pkg : package}
                 {pkg_proofs : package_proofs},
        True.
    (* forall (exprInfo : ExprInfoT) (exprExtraInfo : @ExprExtraInfoT exprInfo)
  (pkg : @package ?base ?ident) (_ : @package_proofs ?base ?ident pkg), True
     : Prop
where
?base : [exprInfo : ExprInfoT
         exprExtraInfo : @ExprExtraInfoT exprInfo
         pkg : @package ?base ?ident |- Type] (pkg cannot be used)
?ident : [exprInfo : ExprInfoT
          exprExtraInfo : @ExprExtraInfoT exprInfo
          pkg : @package ?base ?ident
         |- forall _ : type.type (base.type.type ?base), Type] (pkg cannot be used)
*)
```
)

maximedenes:
I think we are in tricky case: `exact pkg` (which is executed by the TC resolution) is allowed to leave some unresolved shelved goals (probably because they unify with pre-existing evars), and somehow we don't track this properly and end up loosing these evars.

I guess I have all the rewriter-specific info I needed, thanks!

JasonGross:
Btw, I expect this standalone example will show the same behavior (though I haven't tested it on this branch):
```coq
Class A := { a : nat }.
Axiom B : forall {n : nat}, Type.
Existing Class B.
Axiom C : forall {n : nat} {b : @B n}, Type.
Existing Class C.
Axiom D : forall {x : A} {y : @B (@a x)}, Type.
Existing Class D.
Set Printing All.
Definition foo
  : forall {x : A}
           {y : B}
           {z : C}
           {w : D},
    True.
Proof. constructor. Qed.
Check @foo.
(* @foo
     : forall (x : A) (y : @B (@a x)) (_ : @C (@a x) y) (_ : @D x y), True
*)
Check forall {x : A}
             {y : B}
             {z : C},
    True.
(* forall (x : A) (y : @B ?n) (_ : @C ?n y), True
     : Prop
where
?n : [x : A  y : @B ?n |- nat] (y cannot be used)
*)
```

maximedenes:
> I expect this standalone example will show the same behavior

Indeed, it does :) thanks again, it will be easier to work on that minimized example.

ppedrot:
Do I read correctly that this is now ready?

maximedenes:
@ppedrot I believe it is. Note the second commit which circumvents a bug which was hidden by a bogus non-propagation of evar TC-related flags (fixed by this PR).

After a discussion with @mattam82 and @SkySkimmer, we agreed that https://github.com/coq/coq/pull/12920 would take more time/thought, so it was safer to go with the solution in the second commit of this PR for now.

Note that we managed to not need overlays, in the end, which seems like good news.

maximedenes:
FTR, the changes asked by @JasonGross have been addressed and the examples added to the test suite.

engine/evd.ml
```diff
@@ -542,6 +546,17 @@ end = struct
...
+  let replace evk evk' stack =
+    let replace ev = if Evar.equal ev evk then evk' else ev in
+    let replace fgl =
+      { principal = Option.map replace fgl.principal;
+        comb = List.map replace fgl.comb;
+      }
+    in
+    List.map replace stack
+    *)
+
```
ejgallego, line 559: This is dead code right?

dev/ci/user-overlays/07825-reachable-from-evars.sh
```diff
@@ -0,0 +1,9 @@
+#if [ "$CI_PULL_REQUEST" = "7825" ] || [ "$CI_BRANCH" = "reachable-from-evars" ]; then
```
ppedrot, line 1: The overlay can be safely removed.

proofs/refine.ml
```diff
@@ -47,6 +47,7 @@ let typecheck_evar ev env sigma =
 let generic_refine ~typecheck f gl =
   let sigma = Proofview.Goal.sigma gl in
   let env = Proofview.Goal.env gl in
+  if !Flags.debug then Feedback.msg_debug Pp.(str"sigma before generic_refine: " ++ Termops.pr_evar_map None env sigma);
```
ejgallego, line 50: It seems to me that this debug could would make the `-debug` option unusable, right?

engine/evd.ml
```diff
@@ -494,6 +494,10 @@ module FutureGoals : sig
   val add : principal:bool -> Evar.t -> stack -> stack
   val remove : Evar.t -> stack -> stack
 
+(*
```
ppedrot, line 497: More dead code.

ejgallego (changes requested):
The use of `!Flags.debug` seems to need tweaking.

tactics/tactics.ml
```diff
@@ -5184,14 +5184,14 @@ end
...
-      (Lazy {rBeta=true;rMatch=true;rFix=true;rCofix=true;
-             rZeta=false;rDelta=false;rConst=[]})
-      {onhyps = Some []; concl_occs = AllOccurrences }
+    (* For backward compatibility reasons, we do not contract let-ins, but we unfold them. *)
+    let redfun env t =
+      let open CClosure in
+      let flags = RedFlags.red_add_transparent allnolet TransparentState.empty in
+      clos_norm_flags flags env t
+    in
+    reduct_in_concl ~check:false (redfun,DEFAULTcast)
```
ppedrot, line 5194: This might actually be beneficial for performances, maybe a bench could be informative.

ppedrot (approved):

maximedenes:
All comments addressed, should be ready for merge @ppedrot

ppedrot:
Last round for comment, will merge soon.

mattam82:
At laaaaast.... PR 7925 has come around :) Props to Maxime and Gaëtan!

ejgallego (approved):

JasonGross:
Sorry for my belated response to the review request here, but as long as this PR doesn't require an overlay to work with f-c-l (nor any of the other developments) and as long as it adds test-suite files for all of the incompatibilities found along the way, then I'm happy to have had this merged.  It looks though like it's missing a test case for the typeclasses issue that came up in https://github.com/coq/coq/pull/7825#issuecomment-681156343 , perhaps a new test-suite file should be added for that in a new PR?  (The procedure I'm following is that any time something in the CI catches something not caught by the test-suite, it seems like a good idea to add a minimal example to the test-suite.)

PR #6428, opened 2017-12-14
Version 8.7 of #6328 (fix #6313: lost goals in nested ltac in refine)

herbelin:
A version of #6328 for v8.7 which does not change the API.

It still does not take into account the shelved/given-up status of remaining goals in `ltac:()` (as I plan to do in #6328) but it fixes the critical lost of the status of goals which resulted in wrongly shelving all the goals generated up to the point of a `ltac:()` inside a `refine`.

PR #6328, opened 2017-12-06
Fix #6313: lost goals in nested ltac in refine

herbelin:
This fixes what is actually a critical bug in nesting calls to the proof engine as `ltac:(...)`-in-terms does.

The fix is simply to save the list of pending goals (the "future" goals) and to restore them (as was done already for "side effects"), in `refine_by_tactic`.

Some issue remains with `ltac:(tac)`: what is it supposed to do with the shelf. If the shelf produced by `tac` is non empty, should the call fail or return the shelf? The current situation is not good as the shelf is simply lost, i.e. tacitly moved in the ether.

This should probably be versed to 8.7.1, but I modified the API, so a version of the PR which does not change the API should be done. But, first, let's answer the shelve in `ltac:()` question.

(Edited for clarification.)

PR #7826, opened 2018-06-15
evd: restrict_evar with candidates, can raise NoCandidatesLeft

mattam82:
When restricting an evar with candidates, raise an exception if this
restriction would leave the evar without candidates, i.e. unsolvable,
preventing the [clear] request.

We make the new evar a future goal and remove the old one.
If we did nothing, [unshelve tac] would work correctly as it
uses [Proofview.advance] to find the shelved goals, going through
the cleared evar. But [Unshelve] would fail as it expects only
undefined evars on the shelf and throws away the defined ones.

***This was before Hugo's changes in the future_goal interface, maybe removing is not necessary anymore***

This is used in ho-matching: we attach candidates to evars and this enforces that restriction during unifications do not result in evars having no candidates left. This looks generally meaningful to me.

**Kind:** bug fix / feature

No test-suite here, related to ho-matching...
But good to integrate it independently to see if it raises unexpected incompatibilities.

### DIFF
--- a/engine/evarutil.ml
+++ b/engine/evarutil.ml
@@ -426,10 +426,6 @@ let push_rel_context_to_named_context ?hypnaming env sigma typ =
 
 let default_source = Loc.tag @@ Evar_kinds.InternalHole
 
-let restrict_evar evd evk filter ?src candidates =
-  let evd, evk' = Evd.restrict evk filter ?candidates ?src evd in
-  Evd.declare_future_goal evk' evd, evk'
-
 let new_pure_evar_full evd evi =
   let (evd, evk) = Evd.new_evar evd evi in
   let evd = Evd.declare_future_goal evk evd in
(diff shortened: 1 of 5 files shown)


mattam82:
@herbelin I believe it's fine to manipulate the future goals this way, it maintains the invariant that only undefined evars appear in the shelves (global and local).

herbelin (commented):
Well-spotted, this indeed reveals a hole in tracking available candidates.

About the `future_goals` question, I unfortunately don't have a lot of time these days to carefully think at it. My quick intuition is that my changes to `future_goal` would not help in your case. (On the other side, the configuration that you mention is maybe one of the remaining bits I did not understood in a strange behavior I had with #6428, see this [comment](https://github.com/coq/coq/pull/6328#issuecomment-368159039)).

Having `clear` failing when the `clear` can decidably be told to lead to an unsolvable goal seems to be the reasonable option.

In any case, I believe this can be merged quickly. Unless @ppedrot wants to do it, I can be the assignee and merge it on Monday evening if noone has anymore to say by then.

ppedrot (commented):
Nothing to comment on.

herbelin:
@Zimmi48: is it normal that VST fails after taking more than 2 hours and that the documentation target fails with "Error while sending the following sentence to coqtop: From Coq Require Import ssreflect ssrfun ssrbool List."?

Zimmi48:
For the first, yes it can happen sometimes with slow runners. We should add tags to make sure that only some limited kinds of runners are used to build some long developments like VST.
For the second, this is a spurious failure that used to happen a lot but it has been fixed in recent master.
Restarting both jobs.

Issue #7713, opened 2018-06-05
Spurious failure of the vio2vo test


ppedrot:
In two unrelated PRs (#7679 and #7004), the vio2vo test from the Coq test-suite failed spuriously.

Links to the failures:
- https://gitlab.com/coq/coq/-/jobs/72408667
- https://gitlab.com/coq/coq/-/jobs/72483736

Zimmi48:
New occurrence:

- https://gitlab.com/coq/coq/-/jobs/72821129

Looks like it has suddenly become very frequent.

cc @gares

maximedenes:
Why don't we see any output in the tests?

gares:
I'm sorry but so far many seemingly STM related tests  were observed to fail randomly.
As far as I know it has never been the fault of Coq, but rather the VMs and the optimistic value of the argument of -j that made it fail.

I've no clue what you are doing on CI. Does it fail on your PC?

Zimmi48:
> Does it fail on your PC?

Good question. If I see one of these tests fail on my PC I'll report it immediately with the log.

Zimmi48:
That being said, I run the test-suite locally much less frequently than I check CI results.

SkySkimmer:
>Why don't we see any output in the tests?

What do you mean?

Zimmi48:
Indeed, we're just stupid and we forgot that GitLab upload artifacts for failed tests: https://gitlab.com/coq/coq/-/jobs/72821129/artifacts/raw/test-suite/logs/coq-makefile/vio2vo.log

Zimmi48:
The log file I linked to is the following BTW:

```
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
make[1]: Entering directory '/builds/coq/coq/test-suite/coq-makefile/vio2vo/_test'
COQDEP VFILES
CAMLDEP src/test_aux.mli
CAMLDEP src/test.mli
COQDEP src/test_plugin.mlpack
CAMLDEP src/test_aux.ml
CAMLDEP -pp src/test.ml4
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
CAMLC -c src/test_aux.mli
CAMLOPT -c -for-pack Test_plugin src/test_aux.ml
CAMLC -c src/test.mli
CAMLOPT -pp -c -for-pack Test_plugin src/test.ml4
CAMLOPT -pack -o src/test_plugin.cmx
[...]

PR #7828, opened 2018-06-15
[Spawn] no control sock on unix

gares:
Optimistically we also put the timeout back to a reasonable (for a human) value.
If it fails on windows, we can revert last commit.

### DIFF
--- a/lib/spawn.ml
+++ b/lib/spawn.ml
@@ -15,14 +15,12 @@ let accept_timeout = 10.0
 let pr_err s = Printf.eprintf "(Spawn  ,%d) %s\n%!" (Unix.getpid ()) s
 let prerr_endline s = if !Flags.debug then begin pr_err s end else ()
 
-type req = ReqDie | ReqStats | Hello of int * int
-type resp = RespStats of Gc.stat
+type req = ReqDie | Hello of int * int
 
 module type Control = sig
   type handle
 
   val kill : handle -> unit
-  val stats : handle -> Gc.stat
   val wait : handle -> Unix.process_status
   val unixpid : handle -> int
   val uid : handle -> string
@@ -43,7 +41,6 @@ module type MainLoopModel = sig
 end
 
 (* Common code *)
-let assert_ b s = if not b then CErrors.anomaly (Pp.str s)
 
 (* According to http://caml.inria.fr/mantis/view.php?id=5325
  * you can't use the same socket for both writing and reading (may change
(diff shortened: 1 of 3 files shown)


ejgallego:
I won't be able to review this for a while, feel free to take over.

Zimmi48:
The removal of `stats` (probably not fully related to this PR) makes this non-backportable (not that this was very important to backport...).

ejgallego:
After a very brief look, I have a few comments:
- the timeout drop looks unrelated, why do we want to lower it again? So far, it has proven to be effective and indeed I can believe 2sec is too short for win.
- it seems to me that declaring this "fixes" #7713 is a bit too strong, in the sense that we have no evidence yet [Théo was on 4.06.1] and I am not sure I understand the reason why this is supposed to fix the problem.
- relatedly, I am not sure I understand the current solution either, AFAICT the control channel is nothing more than a "gimmick", as OCaml doesn't do preemptive multithreading; which means that we have implemented a solution that won't work reliably. I dislike such solution, I do either prefer something that is broken 100% of the time or 0% of the time.

ejgallego:
So from reading #7713 it is not very clear to me what the "race" is. IMO I'd like to see at least a try to describe what the problem could be, blindly applying changes to a concurrent subsystem will lead us nowhere.

gares:
The PR removed the channel *and* the thread that fails to connect back in 2 seconds.

It is hard to believe it does not fix the problem (on unix, where this PR applies), since with this PR nobody will connect back and nobody will expect someone else to connect back.
So yes, Fix-on-Unix would be more appropriate than Fix, but still...

The timeout to 2 seconds is unrelated, and as I say in my message " If it fails on windows, we can revert last commit.". I'm using this PR to make an attempt. If my theory about this issue is correct, then the timeout is unrelated: the controller thread is stuck because of a bug, not because the system is slow. We will see, I'll try to re-run the job on the windows workers a few times.

ejgallego:
I feel like I am still missing something, but let's see. Théo, do you think this PR is testable?

gares:
> The removal of stats (probably not fully related to this PR

Without the control channel I don't know how to implement stats. I could keep the call and make the value produced be a dummy...

I'm not so sure it is worth backporting this.

Zimmi48:
> Without the control channel I don't know how to implement stats. I could keep the call and make the value produced be a dummy...

Keep it like this then.

> Théo, do you think this PR is testable?

I could try to reproduce the issue by running the test case many times to see it fail, then test that this isn't the case anymore with this PR. However, given Enrico's explanation, I'm under the impression that this is unnecessary as the error could not possibly happen anymore in Unix systems with this change.

Regarding the decreasing of the timeout value, this decrease was done to solve a frequent AppVeyor failure. If I understand this PR correctly, it doesn't change anything on Windows. Then, there's no reason why the failure wouldn't come back. So I'm indeed more reluctant about this second commit.

Zimmi48:
I managed to have 3 failures out of 30 runs when building Coq with `-j4` at the same time in another directory.
I tried to increase the failure rate using `-j8` in this other build but I only got 4 failures out of 30 runs.
Anyways, that's enough to be testable so let me test this PR.

gares:
I'm sorry I think the changes I wanted to make on the timeout was to set it to `(-1.0)` that means no timeout, as discussed in another thread. I don't know why I ended up putting 2 seconds.

If the problem is that the machines are slow, this is the final solution.

Zimmi48:
0 failure out of 60 trials. I confirm that this PR (in its previous form where the timeout wasn't disabled) does fix the problem on Linux. But this was expected indeed. As for changing the value of the timeout, once again, I don't think it is required now. But I don't really care either.

ppedrot:
@gares While you're at it, you should also probably remove the `ReqStats` constructor.

gares:
done and removed the the timeout thing so that the PR can be merged

ejgallego (commented):
I don't have the time to review this in depth now, but given the feedback from other developers I'll merge in a couple of days if there is no further oposition.

PR #7813, opened 2018-06-14
Workaround for #2800: handling non-value arguments in tactics.

cmangin:
Although the fix is not a proper one, it seems to solve every instance of #2800 that I could test, I would be welcoming any counterexample.
Also, I have no idea what would be the proper fix. See the issue discussion for more details.

**Kind:** bug fix.
Fixes #2800 
- [x] Added / updated test-suite

PR #7831, opened 2018-06-15
Fix for issue #7707: include Ltac2 and Equations in Windows build

### LINKED ISSUE #7707: Include Ltac2 and Equations in the Windows package
I think it would be good to encourage our users to start testing what we see as a future replacement of things like Ltac, Funind, Program... So I suggest we distribute Ltac2 0.1 and Equations 1.0 in the Windows packages for 8.8.1 like we did with Bigumns in 8.7.2.

cc @gares @MSoegtropIMC @ppedrot @mattam82 @cmangin

MSoegtropIMC:
On the way I also fixed some minor issues with calling MakeCoq_MinGW from cygwin.

Fixes / closes #7707


dev/build/plugin-versions/bignums.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+case "$1" in
+    git-master|/cygdrive*)
+        # First get the SHA of master/HEAD and then download a zip of exactly this state
+        # Note: the access token is a public access token of MSoegtropIMC
+        BIGNUMS_SHA=$(curl \
+            -H 'Authorization: token 79b1bf691f933bb0185f0fc72816e5ff387bc6ea' \
```
Zimmi48: You shouldn't have pushed this to GitHub, it looks like a personal access token. You should revoke it ASAP. Furthermore, you don't need this to get what you are looking for. You should use https://github.com/coq/bignums/archive/master.zip instead.

dev/build/plugin-versions/bignums.sh
```diff
@@ -0,0 +1,16 @@
...
+            -H 'Accept: application/vnd.github.VERSION.sha' \
+            'https://api.github.com/repos/coq/bignums/commits/HEAD?sha=master')
+
+        echo https://github.com/coq/bignums/archive "$BIGNUMS_SHA" zip 1 "bignums-$BIGNUMS_SHA"
+        ;;
+    *)
+        echo https://github.com/coq/bignums/archive V8.8+beta1 zip 1 bignums-8.8+beta1
+        ;;
+esac
```
Zimmi48: Add newlines at end of files.

dev/build/plugin-versions/bignums.sh
MSoegtropIMC: @Zimmi48 : yes it is a personal access token, but it has exactly the same access rights as public users have anyway (that means it has zero extra rights). As far as I can tell you need an access token to access the API, otherwise the number of accesses you can make from one machine is very restricted (to 50 per hour or so). With an access token you can do 5000 per hour. The worst thing which can happen is that someone uses it for something else and steals from the 5000 requests per hour, but there is no good reason for doing this.

The reason I am not using master.zip is that the downloaded files are cached. This mechanism includes the SHA of the version in the cached archive name. If a new version is available, it will be downloaded and compiled automatically. Without such a mechanism one cannot reasonably cache downloads of plugins.

Zimmi48: Thanks for the explanations. It looks reasonable then. I'll have a more thorough look at the PR later.

dev/build/plugin-versions/bignums.sh
MSoegtropIMC: Done!

dev/build/plugin-versions/bignums.sh
MSoegtropIMC: Just a note: Since I am also not sure what is the right procedure, I tried it without a token like this:

for i in {1..100}; do curl -H 'Accept: application/vnd.github.VERSION.sha'   'https://api.github.com/repos/MSoegtropIMC/coq/commits/HEAD?sha=master'; done

The first 50 or so go fine, after this I get this error:

{
  "message": "API rate limit exceeded for <myip>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}

Maybe someone can review the link given above and see if there is a better way.
According to the github doc, this restriction is per hour.

Of cause I am open to better suggestions.

SkySkimmer: If you have git installed, `git ls-remote https://github.com/coq/bignums refs/heads/master` will print (right now) `1cafe43151450d7a3f7a0b2c8036851900e3eab9        refs/heads/master`, and `git ls-remote https://github.com/coq/bignums refs/heads/master | cut -f 1` will give just the sha.

SkySkimmer: (I ran 100 in a row without issue)

MSoegtropIMC: @SkySkimmer : thanks, I will switch to the method you suggest. It also has the advantage that it is not limited to repositories in github. I didn't know that git can get such information without cloning the repo - interesting.

MSoegtropIMC:
After rebasing to the merge of PR #7813 from today, both equations and ltac-2 work fine with master.
I sqaushed everything, hope CI runs through now.

dev/build/plugin-versions/bignums.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+case "$1" in
+    git-master|/cygdrive*)
+        # First get the SHA of master/HEAD and then download a zip of exactly this state
+        # Note: the access token is a public access token of MSoegtropIMC
```
Zimmi48: Your are not using this method anymore thus, this comment should be removed or adapted. BTW, I have just checked the various permission levels for personal access token and I don't see any that only gives read access. So I strongly suggest that you still revoke the access token that was included in an earlier version of this PR.

MSoegtropIMC: Yes you are right, I should remove the comment on the access token and also fix the Readme file.

Regarding access rights of access tokens: when you create an access token the systems asks which rights to give and you simply don't tick any right at all. Github then displays the access rights of this token as "public access" meaning (as far as I can tell) it as the same rights an unauthenticated user has. But I already deleted the token yesterday after switching to SkySkimmer's method, since I create tokens for specific applications and delete them when they are not needed any more.

Zimmi48: Oh you are right! I had forgotten you could not check any box at all :)

Zimmi48 (approved):
Looks good. Thank you! Maybe you should also update the `.github/CODEOWNER` file to define owners for `dev/build/plugins-versions`

MSoegtropIMC:
@Zimmi48 : would you recommend to set the code owner of each addon version file to the respective addon owner, or should I assign the ownership of this folder to me, since it is (currently) only used for Windows?

Zimmi48:
I think you should assign the whole folder to you.

MSoegtropIMC:
I added the folder to the code owners file.

Btw.: I plan to also do changes to the NSIS installer, but after cleaning this up in general (#7837).

dev/build/plugin-versions/bignums.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+case "$1" in
+    git-master|/cygdrive*)
+        # First get the SHA of master/HEAD and then download a zip of exactly this state
+        BIGNUMS_SHA=$(git ls-remote 'https://github.com/coq/bignums' refs/heads/master | cut -f 1)
+
+        echo https://github.com/coq/bignums/archive "$BIGNUMS_SHA" zip 1 "bignums-$BIGNUMS_SHA"
+        ;;
+    *)
+        echo https://github.com/coq/bignums/archive V8.8+beta1 zip 1 bignums-V8.8+beta1
```
Zimmi48: There is a V8.8.0 tag, and even if it points to the same commit, it's probably more appropriate.

MSoegtropIMC:
I thought about allowing only /cygdrive, git-master, git-vX.Y.Z and vX.Y.Z patterns in the case and throw and error in *). The git-v.X.Y.Z and vX.Y.Z case could be converted with sed to a matching bignum version number (assuming one is willing to use the same version names). How about this? The disadvantage would be that a bignum label would have to be created, even if adjustments to a new Coq version are not required. Maybe only the future can tell what is more desirable.

Btw.: the difference between git-vX.Y.Z and vX.Y.Z is where it gets the sources from. The first is from git (as a zip), the second is from some INRIA repository - possibly anyway deprecated.

Zimmi48:
Don't rely on Inria's servers: you shouldn't make any assumption on what they will distribute. Certainly not the plugins anyways. As for the version name, I don't understand why not all kinds of naming schemes would be supported.

MSoegtropIMC:
@Zimmi: I think the mapping of version numbers is not trivial. Essentially there are two options:

a) always use the same version for Coq and bignums - easy except that if you supply a local folder (the /cygwin/... thing) one can't really tell what version it is - I assume master. As I said this has the disadvantage that one would have to create labels in bignums, even if nothing does change. This might be desirable in the sense of "we explicitly state that this version of bignums has been tested with this version of Coq, so we create a label). But one then really has to create the labels always, for all plugins.

b) State in this file an explicit mapping, say Coq 8.8.0, 8.8.1 and 8.8.2 use bignums 8.8.0, but Coq 8.8.3 and Coq 8.8.4 use bignums 8.8.3

c) In either case one can keep the mapping history in this file or one can only handle master and the latest version here and keep the information about old version mappings in the git history, or one can keep the full version mapping here. I would rather vote for the latter, because this serves as good documentation of compatibiity, but I am open to the other version. Since currently there is only master and latest, currently both ways are identical.

The mechanism I provided can all cases ad variants. It can also handle it differently for every library and I would also say that the case switching is reasonably obvious to serve as documentation on what the mapping is. But I cannot decide which method is more desirable for the maintainers of Coq and bignums.

If someone tells me what method is desirable, I am happy to adjust the files accordingly. Or I can just fix the version as you mentioned, and we delay this decision until we have a more informed view on how this will behave in the future. The path would then be keep explicit mapping tables until one fells that it is time for clean-up and generalization

Zimmi48:
I would be in favor of each branch containing one single version to be used: then there is no question of determining what the current version of Coq is. Moreover, this mapping is likely to be pretty trivial: the API is not going to change in v8.8 so current Bignums 8.8.0 is going to be compatible with all of Coq versions 8.8.x. In the case of a plugin that is actively developed like Equations, we could change the version number for 8.8.2, not because the current version has become incompatible but because there were improvements to Equations in the meantime.

Anyways, my personal medium-term plan about all this is to change the way we handle version of external projects we test in CI (cf. #6724). When I finally find time to implement what I propose here, then I think I'll just plug your code to use the exact same version as is tested in CI.

maximedenes:
So, what should we do? Update the bignums tag for now and find a more general strategy later?

Zimmi48:
Yes, although even like this it is okay since both tags point to the same commit.

MSoegtropIMC:
Ok, I will change it a bit later so that CI can run overnight.

Besides the concrete version I think it is better to see what the future brings and do a 1:1 mapping for the next 3 releases or so and then discuss what might be a reasonable strategy.

MSoegtropIMC:
Sorry had network issues yesterday evening, will do it some time before noon.

MSoegtropIMC:
I did a local build and tested the installer - all fine.
I see failures on circleCI regarding ltac2. This did work before - do I need to rebase again?

Zimmi48:
You mean on GitLab CI? Yes, if you could rebase again it would help make CI green because non backward-compatible patches have been merged into some plugins in the meantime.

MSoegtropIMC:
sorry, yes, I meant gitlab CI. I just pushed the rebase.

maximedenes:
I realized recently that this PR has a drawback: since the windows build script does not take overlays into account, any PR needing an overlay for bignums, ltac2 or equations is going to make the windows build fail. Do we want to address that before extending the set of packages included?

Zimmi48:
That's a very good point, and this can be solved by using the same system as in the rest of the CI, i.e.:
- sourcing `ci-common.sh`
- using the value of the environment variables (for instance `$ltac2_CI_GITURL` and `$ltac2_CI_BRANCH`).

MSoegtropIMC:
I must admit, I don't understand the overlay system and related issues and can't comment on it or fix it. Is there some documentation I can read?

Zimmi48:
`dev/ci/ci-common.sh` defines a bunch of bash variables and functions, in particular it does:

https://github.com/coq/coq/blob/915452f9a73d25e45131edb08531c29a79ab7020/dev/ci/ci-common.sh#L45-L49

`ci-basic-overlay.sh` contain default locations and branches, for instance:

https://github.com/coq/coq/blob/915452f9a73d25e45131edb08531c29a79ab7020/dev/ci/ci-basic-overlay.sh#L59-L63

These variables can be overridden in pull request testing using overlays, for instance:

https://github.com/coq/coq/blob/915452f9a73d25e45131edb08531c29a79ab7020/dev/ci/user-overlays/07213-ppedrot-fast-constr-match-no-context.sh#L3-L4

So simply doing as I said in my previous message (sourcing `ci-common.sh` and using the values of `..._CI_GITURL` and `..._CI_BRANCH`) will be enough to:

1. Ensure that the versions of the plugins that are packaged on Windows are the same that were tested in our Linux CI.
2. Allow overlays for pull requests to apply there as well.

More reading: https://github.com/coq/coq/tree/master/dev/ci/user-overlays#add-overlays-for-your-pull-requests-in-this-directory

MSoegtropIMC:
@Zimmi48 : I see, so if I understand this right that this solves the same problem as I tried to solve with the /dev/build/plugin-versions mechanism I introduced. Shouldn't I just remove what I introduced then and use this mechanism instead? Or woudln't this be appropriate, because configuring the installer is not just CI testing. Still I guess that the installer best uses the same versions as CI tests and it doesn't make that much sense to have two mechanisms for the same thing. It must just be clear than to everybody that what the installer contains is controlled by files in /dev/ci.

On the other hand I am not sure if it is desirable to always pull in master versions rather than specific labels as the CI plugin version control seems to do.

Or should we do it the other way around? Use my mechanism for stuff which goes into the installer and reuse this in CI for those plugins it does handle?

Zimmi48:
> I see, so if I understand this right that this solves the same problem as I tried to solve with the /dev/build/plugin-versions mechanism I introduced. Shouldn't I just remove what I introduced then and use this mechanism instead?

I think this would indeed be the right course of action.

> On the other hand I am not sure if it is desirable to always pull in master versions rather than specific labels as the CI plugin version control seems to do.

I agree, although we can decide to setup a specific policy for stable branches (use a tag instead of a branch, this is supported as well) given that the installers we'll distribute in the end will be the ones we built on stable branches.

You could even consider going one step further and trying to share the build instructions from:

https://github.com/coq/coq/blob/915452f9a73d25e45131edb08531c29a79ab7020/dev/ci/ci-ltac2.sh#L1-L10

Not sure it's worth it though...

MSoegtropIMC:
@Zimmi48 : ok, let me review the mechanism used in CI and conclude if I think it is appropriate for release building or not. In case I think it is appropriate, I will switch to using it, otherwise I will discuss here what issues I see. Just as I said, I am super busy the next 2 weeks. Can this wait so long?

Zimmi48:
> Can this wait so long?

No, it really can't but I am OK to take over and adapt this PR for you if you want. I wouldn't go as far as my last suggestion in this first step though.

MSoegtropIMC:
@Zimmi48 : that's fine (remove the mechanism I introduced and switch to the CI mechanism). If it can't wait, please take this over.

Zimmi48:
This should be ready now. I activated the build of Ltac2 and Equations. I don't understand why it hadn't been the case until now.
BTW I think the current build logs are *way* too verbose. I had to scroll for very long minutes to get to the point where the plugins were built.

SkySkimmer:
> BTW I think the current build logs are way too verbose. I had to scroll for very long minutes to get to the point where the plugins were built.

Now that we use gitlab we could pipe the logs into files, split in whatever way is convenient, and upload them as artefacts. (we can do this for some logs leaving others to stdout as well)

MSoegtropIMC (commented):
Thanks, looks indeed fine now.

maximedenes (approved):

maximedenes:
@Zimmi48 did you remove the overlay files on purpose?

Zimmi48:
Yes, a bit of clean-up doesn't do harm from time to time and I wanted to remove any trace of a `.git` extension (to be able to append `/archive` instead).

SkySkimmer:
>Yes, a bit of clean-up doesn't do harm from time to time and I wanted to remove any trace of a .git extension (to be able to append /archive instead).

You should add something to the overlay readme saying people should use the url without .git then.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -1211,6 +1214,10 @@ function make_coq {
     # 2.) clean of test suites fails with "cannot run complexity tests (no bogomips found)"
     # make clean
 
+    # Copy these files somewhere the plugin builds can find them
+    cp dev/ci/ci-basic-overlay.sh /build/
+    cp -r dev/ci/user-overlays /build/
```
Zimmi48: In #7928, @MSoegtropIMC asked where:

```
. /build/ci-basic-overlay.sh
for overlay in /build/user-overlays/*.sh; do
```

come from. This is where. Note that I just imitated what you had done previously with the `dev/ci/plugins_version/` directory.

> Or is the /build folder something else which exists on the CI machines?

No, it's really the `/build` folder as you created it, nothing else.

Issue #7839, opened 2018-06-16
Pervasive use of non-type-safe primitives in coqchk


whitequark:
I had an idea that would involve cross-compilation of `coqchk` and feeding it untrusted input, so I decided to see if that can be done. I was quite surprised to see 137 instances of `Obj.*` in TCB, 40 instances of the "%identity" primitive under a different name, 51 instance of `Array.unsafe_*`, 9 instances of `String.unsafe_*`, and especially the `values.ml` and `validate.ml` files from coqchk. 

Needless to say, I have no confidence at all that coqchk will do its job on potentially malicious input. This is not at all what I expect from a proof checker.

To add to this, the use of Obj ties Coq to a specific version of a specific implementation of the OCaml runtime, which I can't imagine is desirable.

To add more to this, the use of OCaml marshaling for `vo` files makes it significantly harder than necessary to implement a hypothetical `safecoqchk` that would not suffer from any of these issues.

I understand that my first major interaction with the Coq community being a complaint does not look good, but I hope it's clear that I only wish Coq be the best it can be. In any case, I'll be glad to fix some of these once there's consensus on what should be done.

I would really like to see Coq:
  - [ ] use a proper serialization format for `vo` files;
  - [x] remove `Hashcons.Hobj` entirely, it is not used anywhere in Coq (#7840);
  - [x] remove `CEphemeron`, which is only necessary because of the use of marshalling, or in any case, use GADTs and standard library ephemerons instead (#7850);
  - [ ] use GADTs for implementing `Store`, since there's `option` overhead anyway;
  - [ ] ~use [inline records](https://blog.janestreet.com/ocaml-4-03-everything-else/) for implementing `CList` since that allows doing the obvious thing in a type-safe way~ wouldn't work because CList.list has to be compatible with Prelude.list;
  - [ ] ~rely on flambda to optimize the type-safe but more naive implementations of select primitives in `CMap`, `CSet` and `Int` that currently use `Obj.magic` to expose the standard library internals~ flambda cannot do this;
  - [x] use either a proper custom block in `Canary`, or at least transmute a closure without making any particular assumptions about its layout, or remove it (#7841);
  - [ ] use the new Spacetime profiler instead of `CObj` or at least not make `CObj` a hard requirement for `coqchk`;
  - [ ] make the use of unsafe Array/String primitives optional with a switch;
[...]

PR #7841, opened 2018-06-16
Remove Canary

whitequark:
See #7839 for motivation.


ppedrot (approved):
Fine with me, this was mostly used for debugging anyways.

ppedrot (changes requested):
You also need to adapt the checker/values.ml file accordingly, as you changed the memory representation of kernel names.

whitequark:
Done.

whitequark:
I think the `coq-dpdgraph` test fails because of [this line](https://github.com/Karmaki/coq-dpdgraph/blob/master/graphdepend.ml4#L54-L55). `GlobRef.ConstRef` has a `KerPair`, which is a pair of `KerNames`, which used to have a `canary`. So the old canary implementation didn't even catch bugs...

ppedrot:
Why do you think so? It is only comparing the integer part AFAICS...

whitequark:
I'm not entirely sure, but `Canary.obj` has quite surprising behavior:
```ocaml
# let obj = Obj.new_block Obj.closure_tag 0;;
val obj : Obj.t = <abstr>
# obj = obj;;
Exception: Invalid_argument "compare: functional value".
Raised by primitive operation at unknown location
Called from file "toplevel/toploop.ml", line 180, characters 17-56
# obj < obj;;
Exception: Invalid_argument "compare: functional value".
Raised by primitive operation at unknown location
Called from file "toplevel/toploop.ml", line 180, characters 17-56
# compare obj obj;;
- : int = 0
```
From the docs of `Canary`:
```
val obj : t
(** Canary. In the current implementation, this object is marshallable,
    forbids generic comparison but still allows generic hashes. *)
```
In other words, the purpose of Canary was to disallow polymorphic comparison, but it doesn't really do that, and in a way relevant to the behavior of those lines. But of course I may be wrong.

ppedrot:
@whitequark I don't think it has anything to do with that. If I understand the diff correctly, it is simply that the output is sensitive to the order in which the hashtable storing the graph is traversed. As this patch changes the hash function, you get a different permutation. That said, the code is still wrong as it relies on generic hashtables for global references. So, the plugin is incorrect no matter what. I will write a patch to solve this issue.

ppedrot:
Karmaki/coq-dpdgraph#49 makes this PR go through.

ppedrot (approved):

maximedenes (approved):

maximedenes:
> Are you OK with the PR?

Yes, feel free to merge!

PR #7844, opened 2018-06-16
Remove Elpi from Travis.

Zimmi48:
Elpi is currently failing on Travis but not on GitLab CI, likely because of the recent dependency update that was only done in the Docker image.

Rather than trying to keep the version of dependencies in sync with GitLab CI, we choose to remove Elpi from Travis. Travis tests are already far from complete compared to GitLab CI so this doesn't do harm.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,11 +82,6 @@ matrix:
       - TEST_TARGET="ci-coquelicot"
     - if: NOT (type = pull_request)
       env:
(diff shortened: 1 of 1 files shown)


Zimmi48:
CI testing is rather useless for this PR, but the build of my branch shows the build going through with Elpi not in there: https://travis-ci.org/Zimmi48/coq/builds/393138691

ejgallego (approved):

PR #7848, opened 2018-06-17
Fix a typo in documentation

whitequark:


Zimmi48 (approved):
Thanks!

Zimmi48:
It looks like you based this branch on an outdated version of master. This is going to make most of CI fail. However, since the fix is trivial, I'll merge as-is.

PR #7849, opened 2018-06-17
Fix #7615: Functor inlining drops universe substitution.

### LINKED ISSUE #7615: Functor inlining drops universe substitution

ppedrot:
Backport of #7616.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Changes from 8.8.0 to 8.8.1
 Kernel
 
 - Fix a soundness bug with CoFixpoints and VM/native_compute (see #7333).
+- Fix a critical bug with inlining of polymorphic constants (#7615).
 
 Notations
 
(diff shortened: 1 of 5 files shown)


Zimmi48:
Thanks!

maximedenes (approved):
Looks good to me, but I have only a partial understanding of this code.

PR #7850, opened 2018-06-17
Rewrite CEphemeron to use a type-safe implementation based on GADTs

whitequark:
See #7839 for motivation.

**Kind:** cleanup.

This eliminates 2 uses of Obj from TCB.

Moreover, with the new implementation, a key that is marshalled and then unmarshalled does not compare equal with itself even using the polymorphic equality function.

See the comment in CEphemeron.ml for details on implementation.

In order to convince myself that the new implementation does not leak memory, I've used the following adapted example:
<details><summary>test_ephemeron.ml</summary>

```ocaml
type 'a typ = ..

type boxkey = Box : 'a typ -> boxkey [@@unboxed]

type 'a key = 'a typ * boxkey

module EHashtbl = Ephemeron.K1.Make(struct
  type t = boxkey
  let equal = (==)
  let hash  = Hashtbl.hash
end)

type value = { get : 'k. 'k typ -> 'k } [@@unboxed]

let values : value EHashtbl.t =
  EHashtbl.create 1001

module type Unit = sig end
let index = ref 0

let create : type v. v -> v key =
  fun value ->
  let module M = struct
    type _ typ += Typ : v typ

    let get : type k. k typ -> k =
      fun typ ->
      match typ with
      | Typ -> value
      | _ -> assert false

    let boxkey = Box Typ
    let key    = Typ, boxkey
    let value  = { get }
  end in
  incr index; let index = !index in
  Gc.finalise_last (fun _ -> Printf.printf "collected boxkey %d\n%!" index) M.boxkey;
  Gc.finalise_last (fun _ -> Printf.printf "collected key %d\n%!" index) M.key;
  EHashtbl.add values M.boxkey M.value;
  M.key

let get : type v. v key -> v =
  fun (typ, boxkey) ->
  (EHashtbl.find values boxkey).get typ

let contains : type v. v key -> bool =
  fun (typ, boxkey) ->
  EHashtbl.mem values boxkey

type bar = { mutable fld : int }

let make_bar i =
  { fld = i }

let show_alloc n =
  Gc.compact ();
  EHashtbl.clean values;
  print_string "gc! ";
  print_int n;
  print_string ": live: ";
  print_int (Gc.stat ()).live_words;
  print_newline ()

let () =
  show_alloc 1;
  let v1 = "foo" in
  show_alloc 2;
  (fun () ->
    show_alloc 3;
    let k1 = create v1 in
    show_alloc 4;
    (fun () ->
      show_alloc 5;
      let v2 = (fun () ->
        show_alloc 6;
        let v2 = make_bar 1 in
        Printf.printf "v2: %d\n%!" v2.fld;
        Gc.finalise_last (fun _ -> Printf.printf "collected value 2\n%!") v2;
        show_alloc 7;
        let k2 = create v2 in
        show_alloc 8;
        Printf.printf "k1: %s, k2: %d\n%!" (get k1) (get k2).fld;
        Printf.printf "n = %d\n%!" (EHashtbl.length values);
        v2
      ) () in
      show_alloc 9;
      Printf.printf "n = %d\n%!" (EHashtbl.length values);
      Printf.printf "v2: %d\n%!" v2.fld
    ) ();
    show_alloc 10;
    Printf.printf "k1: %s\n%!" (get k1);
    show_alloc 11;
    Printf.printf "n = %d\n%!" (EHashtbl.length values);
  ) ();
  show_alloc 12;
  Printf.printf "n = %d\n" (EHashtbl.length values)
```
</details>

In order to convince myself that the new implementation does not add space overhead (both the keys and the values are unboxed), I've checked that unboxing does what it should by diffing the output of `-dlambda`:
<details><summary>lambda.diff</summary>

```diff
--- unboxed.sexp  2018-06-17 10:21:21.906857693 +0000
+++ boxed.sexp    2018-06-17 10:21:30.482920519 +0000
@@ -1,53 +1,47 @@
 (setglobal CEphemeron!
   (let
     (EHashtbl/1006 =
        (module-defn(EHashtbl/1006) cEphemeron.ml(x):x-x
          (apply (field 13 (field 0 (global Ephemeron!)))
            (let
              (equal/1008 =
                 (function prim/1316 prim/1315 stub (== prim/1316 prim/1315))
               hash/1009 = (field 20 (global Hashtbl!)))
              (makeblock 0 equal/1008 hash/1009))))
      values/1115 = (apply (field 0 EHashtbl/1006) 1001)
      create/1116 =
        (function value/1119
          (let
            (M/1128 =
               (module-defn(M/1128) cEphemeron.ml(x):x-x
                 (let
                   (Typ/1120 = (makeblock 248 "Typ" (caml_fresh_oo_id 0))
                    get/1121 =
                      (function typ/1124
                        (if (== typ/1124 Typ/1120) value/1119
                          (raise
                            (makeblock 0 (global Assert_failure/28g)
                              [0: "cEphemeron.ml" 27 13]))))
-                   boxkey/1125 = (makeblock 0 Typ/1120)
-                   key/1126 = (makeblock 0 Typ/1120 boxkey/1125)
-                   value/1127 = (makeblock 0 get/1121))
-                  (makeblock 0 Typ/1120 get/1121 boxkey/1125 key/1126
-                    value/1127))))
+                   key/1126 = (makeblock 0 Typ/1120 Typ/1120))
+                  (makeblock 0 Typ/1120 get/1121 Typ/1120 key/1126 get/1121))))
            (seq
              (apply (field 4 EHashtbl/1006) values/1115 (field 2 M/1128)
                (field 4 M/1128))
              (field 3 M/1128))))
      InvalidKey/1129 =
        (makeblock 248 "CEphemeron.InvalidKey" (caml_fresh_oo_id 0))
      get/1130 =
        (function param/1135
          (try
-           (apply
-             (field 0
-               (apply (field 6 EHashtbl/1006) values/1115
-                 (field 1 param/1135)))
+           (apply (field 6 EHashtbl/1006) values/1115 (field 1 param/1135)
              (field 0 param/1135))
           with exn/1202
            (if (== exn/1202 (global Not_found/22g)) (raise InvalidKey/1129)
              (reraise exn/1202))))
      *match*/1201 =
        (let (foo/1136 = "foo" k2/1137 = (apply create/1116 foo/1136))
          (apply (field 1 (global Printf!))
            [0: [11: "k2: " [2: 0a [12: '\n' 0a]]] "k2: %s\n"]
            (apply get/1130 k2/1137))))
     (makeblock 0 EHashtbl/1006 values/1115 create/1116 InvalidKey/1129
       get/1130)))
```
</details>

Indeed, the listing above also shows that the amount of indirection remained the same, and the space overhead is lower because the collection queue and per-key finalizers are no longer necessary.

EDIT: depends on ~~#7522~~.


whitequark:
The CircleCI build failed because it runs 4.02.3 and Ephemeron is introduced in 4.03.0.

ppedrot:
So this depends on #7522.

Zimmi48:
You should be aware that Emilio was proposing the update of the minimal OCaml version to happen in 8.10, so this means that the PR that this depends upon won't be merged before September.

ejgallego:
Indeed a lot of cleanups are pending in uping the required OCaml version. In my view 8.10 indeed should seriously consider using 4.05.0.

As for 8.9, I am personally neutral, I guess the RM should propose what to do here.

whitequark:
@ppedrot Benchmarks, done on 4.06.1 x86_64 Linux:
```
Throughputs for "CEphemeronOld.create", "CEphemeron.create" each running 3 times for at least 3 CPU seconds:
CEphemeronOld.create:  3.11 WALL ( 3.09 usr +  0.02 sys =  3.11 CPU) @ 4157614.66/s (n=12914682)
                       3.12 WALL ( 3.11 usr +  0.01 sys =  3.12 CPU) @ 4497974.28/s (n=14045343)
                       3.13 WALL ( 3.13 usr +  0.00 sys =  3.13 CPU) @ 4535824.04/s (n=14199819)
   CEphemeron.create:  3.30 WALL ( 3.30 usr +  0.00 sys =  3.30 CPU) @ 3953796.33/s (n=13053253)
                       3.26 WALL ( 3.26 usr +  0.00 sys =  3.26 CPU) @ 4002351.45/s (n=13053253)
                       3.28 WALL ( 3.28 usr +  0.00 sys =  3.28 CPU) @ 3984510.68/s (n=13053253)
Result:
                          Rate            CEphemeron.create CEphemeronOld.create
   CEphemeron.create 3980219+- 19051/s                   --                  -9%
CEphemeronOld.create 4397138+-161568/s                  10%                   --
Throughputs for "CEphemeronOld.get", "CEphemeron.get" each running 3 times for at least 3 CPU seconds:
CEphemeronOld.get:  3.16 WALL ( 3.16 usr +  0.00 sys =  3.16 CPU) @ 36176994.38/s (n=114380369)
                    3.17 WALL ( 3.16 usr +  0.00 sys =  3.17 CPU) @ 36098647.33/s (n=114380369)
                    3.18 WALL ( 3.17 usr +  0.00 sys =  3.18 CPU) @ 36013437.14/s (n=114380369)
   CEphemeron.get:  3.18 WALL ( 3.18 usr +  0.00 sys =  3.18 CPU) @ 15617025.69/s (n=49617477)
                    3.18 WALL ( 3.17 usr +  0.00 sys =  3.17 CPU) @ 15627847.14/s (n=49617477)
                    3.17 WALL ( 3.17 usr +  0.00 sys =  3.17 CPU) @ 15647136.60/s (n=49617477)
Result:
                        Rate           CEphemeron.get CEphemeronOld.get
   CEphemeron.get 15630670+-11831/s                --              -57%
CEphemeronOld.get 36096360+-63452/s              131%                --
```
<details><summary>bench_cEphemeron.ml</summary>

```ocaml
open Benchmark

type r = { f: int }

let blackhole v = ignore (Sys.opaque_identity v)

let count = ref 0
let maybe_clear f =
  incr count;
  if !count = 10000 then begin
    count := 0;
    f ();
  end

let () =
  let value = { f = 1 } in
  let res = throughputN ~repeat:3 3
    [("CEphemeronOld.create", 
      (fun () -> 
        CEphemeronOld.create value |> blackhole;
        maybe_clear CEphemeronOld.clear), ());
     ("CEphemeron.create",    
      (fun () -> 
        CEphemeron.create value |> blackhole;
        maybe_clear CEphemeron.clear), ())] 
  in
  print_endline "Result:";
  tabulate res;
  let keyold = CEphemeronOld.create value in
  let key = CEphemeron.create value in
  let res = throughputN ~repeat:3 3
    [("CEphemeronOld.get",    (fun () -> CEphemeronOld.get keyold |> blackhole), ());
     ("CEphemeron.get",       (fun () -> CEphemeron.get key |> blackhole), ())] 
  in
  print_endline "Result:";
  tabulate res
```
</details>
I looked at a few potential improvements but none made a difference.

ppedrot:
Large scale bench running here: https://ci.inria.fr/coq/job/benchmark-part-of-the-branch/452.

whitequark:
By the way, I just realized that
```ocaml
type 'a key = 'a typ * boxkey
```
can be replaced with just
```ocaml
type 'a key = 'a typ
```
if the compiler *guarantees* unboxing, such that `Box x == Box x` always holds. But it doesn't make this implementation any faster, just marginally more space-efficient.

gares:
Dear @whitequark, I'm the "author" of the original code (largely inspired from alt-ergo IIRC).

I think you are mixing 3 changes here, two of which I'm 100% in favor, and another one that leaves me less enthusiastic:
- using the ephemerons provided by recent OCaml rather than the ones built using custom finalizers
- using GADT to drop calls to `Obj.`
- using extensible types to achieve uniqueness

If I'm not mistaken only the latter requires a recent OCaml version and, to my eyes, bring little benefit.

If I'm right that the three changes are orthogonal, could you make a simpler PR only including the first two changes?

ppedrot:
@gares AFAIU you need extensible types to be type-safe. Otherwise I don't know how to do this.

gares:
With GADT you get the existential type you need in order to put any value in the map (without needing to go trough Obj.t), and with a boxed key you get uniqueness (by the allocator) rather than by the semantics of `+=`.

ppedrot:
But the type constraint is provided by the parameter of the extensible type. Otherwise you don't know anything with `Box`, it's just an existential wrapper. I might be wrong and if so I'd be curious to see your proposal.

whitequark:
@gares However, there is no way to match on this existential to extract the value contained in the map. It's true that physical equality suffices to achieve type safety, but the OCaml type system is not powerful enough to derive the equality witness from physical equality alone.

SkySkimmer:
>If I'm not mistaken only the latter requires a recent OCaml version and, to my eyes, bring little benefit.

Ephemeron is @since 4.03.0

gares:
Wrt performances, I estimate that we use ~ 2K ephemerons per file (max) and the the performance of get is largely irrelevant. Wrt memory consumption we can't really measure it using coqc, since no ephemeron is expected to be garbage. Only interactive use should eventually trigger a collection.

gares:
> Ephemeron is @SInCE 4.03.0

Thanks, hence my request of a split should be ignored.

whitequark:
Moreover, I've actually considered whether there could (or should) be any way, perhaps in the OCaml stdlib, to go from `'a 'b. fun (a:'a) (b:'b) -> a == b` to `('a, 'b) eq` (with `eq` defined as `type (_, _) eq = Eq : ('e, 'e) Eq`) and this can actually *break* type safety.

Consider a library that has:
```ocaml
type 'a foo = private unit
type phantoma
type phantonb
val f : unit -> phantom a
val g : unit -> phantom b
val h : phantom a -> unit
```

Consider that `g` could do nothing and `f` perform some operation `h` depends on. If it was possible to cast from `phantom b` to `phantom a`, then it would be possible to call `h` without ever calling `f`, which is clearly undesirable.

gares:
> but the OCaml type system is not powerful enough to derive the equality witness from physical equality alone.

I guess you are right, but I'm not sure I would buy the use of extensible type to get rid of an "axiom" as (pseudo code) the following one:
```
let pointer_eq (x : 'a) (y : 'b) : (eq 'a 'b) option = if x == y then Some(Obj.magic refl) else None
```

ppedrot:
This is unsound in general, if you use e.g.
```
type _ foo = Foo : int foo
type _ bar = Bar : string bar
```
Obviously `Foo == Bar` but the two types better not be convertible. That's like the Agda bug on injective parameters or something.

gares:
@ppedrot for "boxed" values, I meant.  The argument of @whitequark still holds, even if it is exactly type safety that breaks, rather the use we all make to impose an order in a set of APIs...

whitequark:
And the last few comments here are an excellent illustration for why I believe the TCB should contain *zero* instances of `Obj`...

ppedrot:
@whitequark Well, if we had some semantic model of OCaml we could prove that by hand. Unluckily, apart from CakeML I don't know of such project, let alone for OCaml.

ppedrot:
@gares Also, even for boxed values, the compiler is free to share the representation as long as the type is pure, as this is compatible with OCaml semantics. This would still break soundness in this case.

gares:
> And the last few comments here are an excellent illustration for why I believe the TCB should contain zero instances of Obj...

FTR I like your code because it is simpler (especially wrt the code performing the free), not because it is more type safe. The use of `Obj` done before was correct as far as I can tell.

whitequark:
> FTR I like your code because it is simpler (especially wrt the code performing the free), not because it is more type safe. The use of Obj done before was correct as far as I can tell.

I also think that the use of `Obj` was correct here. The point of removing it was not restoring soundness, but making it easier to convince (oneself, or anyone else) that the code is indeed sound.

ppedrot:
@whitequark If you want a real source of potential memory unsoundness, you should look at `store.ml`. I have *fix this* somewhere on my TODO list for ages. Luckily it's not used in an unsound way currently (I think), and it's not used by the checker either.

whitequark:
> @whitequark If you want a real source of potential memory unsoundness, you should look at store.ml. I have fix this somewhere on my TODO list for ages. Luckily it's not used in an unsound way currently (I think), and it's not used by the checker either.

Luckily, I have it on my radar, and already consider some potential fixes.

ppedrot:
The proper fix is to reuse `dyn.ml`, I think. For that one, I don't know how to do it without using unsafe code though.

whitequark:
screenshot_20180619_141958

Oh, dear.

whitequark:
> The proper fix is to reuse dyn.ml, I think. For that one, I don't know how to do it without using unsafe code though.

Regarding the proper fix, are you sure reusing `dyn.ml` is the right path? The code in `store.ml` would seem to suggest that it is performance-sensitive, all the `unsafe_get` and such. (Amusingly, it does not use `unsafe_blit` for some reason.) If it can tolerate a comparison or two, I can just use the same GADT trick there as here.

ppedrot:
Dyn is actually correct, as far as my knowledge of OCaml semantics goes. Which is *not* the case for Store.

ppedrot:
@whitequark The problem is that the key index has to be preserved by marshalling, which doesn't work with the solution discussed in this PR (which turns out to be the one feature we rely on).

clib/cEphemeron.ml
```diff
@@ -8,84 +8,104 @@
...
-    with InvalidKey -> None
-  with
-  | None -> ()
-  | Some v -> f v
+let iter_opt (typ, boxkey) f =
+  try f ((EHashtbl.find values boxkey).get typ)
+  with Not_found -> ()
 
-let clear () = run_collection ()
+(* NB: clear only removes dead bindings, so it uses Ephemeron.S.clean *)
```
gares: Can you explain this comment? I've hard time understand what it means.
The semantics of the previous clear was to force a collection, that looks like "only removes dead bindings".

whitequark: A function `clear` calling a function `EHashtbl.clean` when `EHashtbl.clear` also exists looks like a typo. Here it is not a typo, the semantics of `clear` correspond to `EHashtbl.clean`. Maybe the `CEphemeron.clear` should just be renamed.

whitequark: Renamed.

whitequark:
> @whitequark If you want a real source of potential memory unsoundness, you should look at store.ml. I have fix this somewhere on my TODO list for ages. Luckily it's not used in an unsound way currently (I think), and it's not used by the checker either.

> The proper fix is to reuse dyn.ml, I think. For that one, I don't know how to do it without using unsafe code though.

@ppedrot 
Actually, Coq's Dyn is not only *unsafe*, it's just unsound, period. Consider that `Dyn.create` has the signature `string -> 'a tag` and there is nothing keeping the caller from doing something like:
```ocaml
let t1: int tag = Dyn.create "int" in
let t2: string tag = Dyn.create "int" in
...
```

ppedrot (approved):
I forgot to comment on this PR but I am OK on principle with it.

ejgallego:
> EDIT: I forgot about the need to update the OCaml version. Let's wait for #7522 then...

This can move forward now as the minimal OCaml is 4.05.0

ppedrot:
@whitequark I took the freedom to rebase myself. If the CI agrees I'll merge it quickly.

ejgallego:
> @whitequark I took the freedom to rebase myself. If the CI agrees I'll merge it quickly.

Still outdated it seems?

ppedrot:
This outdated check is mysterious to me...

ppedrot:
No observable change on the bench, I'll merge the PR.
```
┌──────────────────────────┬─────────────────────────┬───────────────────────────────────────┬───────────────────────────────────────┬─────────────────────────┬──────────────────────┐
│                          │      user time [s]      │              CPU cycles               │           CPU instructions            │  max resident mem [KB]  │      mem faults      │
│                          │                         │                                       │                                       │                         │                      │
│             package_name │     NEW     OLD PDIFF   │            NEW            OLD PDIFF   │            NEW            OLD PDIFF   │     NEW     OLD PDIFF   │  NEW  OLD    PDIFF   │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│             coq-compcert │  886.78  891.90 -0.57 % │  2458162393677  2472100737228 -0.56 % │  3560895788313  3560153882970 +0.02 % │ 1342820 1343168 -0.03 % │  309  118  +161.86 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│                coq-sf-lf │   44.18   44.41 -0.52 % │   121245935871   121958085398 -0.58 % │   196811224285   196930412510 -0.06 % │  428188  428092 +0.02 % │  116    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│                coq-flocq │  589.09  592.06 -0.50 % │  1640019517737  1647281891097 -0.44 % │  2375551009943  2378184252907 -0.11 % │ 1761620 1761908 -0.02 % │  206   46  +347.83 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│              coq-bignums │   98.08   98.50 -0.43 % │   271942709717   272485920994 -0.20 % │   387290283101   387544213972 -0.07 % │  537944  538588 -0.12 % │   11    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│         coq-math-classes │  247.22  247.83 -0.25 % │   681961440164   682087873171 -0.02 % │   891779755682   891844691927 -0.01 % │  524844  524900 -0.01 % │   84   84    +0.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│         coq-fiat-parsers │  718.08  719.72 -0.23 % │  1993721340239  1998739477625 -0.25 % │  3064092616047  3064468963130 -0.01 % │ 2931020 2897592 +1.15 % │    3  342   -99.12 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│    coq-mathcomp-solvable │  220.07  220.53 -0.21 % │   610965989938   611474470352 -0.08 % │   872222055102   873539878661 -0.15 % │  833704  814916 +2.31 % │    8  118   -93.22 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│               coq-sf-plf │   74.01   74.16 -0.20 % │   204899507169   205269102411 -0.18 % │   298032948833   298169714603 -0.05 % │  518944  518760 +0.04 % │   15    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│           coq-coquelicot │  103.28  103.46 -0.17 % │   285422826354   285602648060 -0.06 % │   383925328164   383979038301 -0.01 % │  716640  717228 -0.08 % │  281   13 +2061.54 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│                 coq-hott │  341.57  341.78 -0.06 % │   944823062158   945377263127 -0.06 % │  1438818037095  1439441697372 -0.04 % │  610220  610368 -0.02 % │   71    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│ coq-mathcomp-real-closed │  187.65  187.70 -0.03 % │   522140829836   521994177015 +0.03 % │   788589986009   788637558658 -0.01 % │  838292  837908 +0.05 % │    0    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│   coq-mathcomp-character │  292.95  293.02 -0.02 % │   814503387075   814451193594 +0.01 % │  1184497222576  1184602149506 -0.01 % │ 1118068 1123044 -0.44 % │    0    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│    coq-mathcomp-fingroup │   84.64   84.64 +0.00 % │   233938033785   234161463954 -0.10 % │   346798776188   346938901173 -0.04 % │  588540  588984 -0.08 % │   97    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│   coq-mathcomp-ssreflect │   65.25   65.21 +0.06 % │   179498153435   179515208239 -0.01 % │   254762408712   254770293700 -0.00 % │  534648  535408 -0.14 % │    2    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│     coq-mathcomp-algebra │  205.16  204.96 +0.10 % │   569282394087   568605149993 +0.12 % │   797280168774   797096956198 +0.02 % │  644956  645308 -0.05 % │  174    9 +1833.33 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│              coq-unimath │ 1323.07 1321.26 +0.14 % │  3671658876393  3666449982586 +0.14 % │  6208484208210  6208993901476 -0.01 % │ 1046940 1053728 -0.64 % │  355   19 +1768.42 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│      coq-formal-topology │   57.78   57.70 +0.14 % │   157754697217   157933955877 -0.11 % │   237695747717   237920328154 -0.09 % │  480052  480644 -0.12 % │    0    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│          coq-fiat-crypto │ 6324.36 6315.34 +0.14 % │ 17587048524053 17561954675408 +0.14 % │ 28982435078914 28989956328725 -0.03 % │ 2631364 2637844 -0.25 % │ 1030 1345   -23.42 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│                 coq-corn │ 1549.70 1547.21 +0.16 % │  4301196116891  4296569393774 +0.11 % │  6358640558308  6358879481521 -0.00 % │  862620  862612 +0.00 % │   44   44    +0.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│            coq-fiat-core │  135.05  134.82 +0.17 % │   375289895617   375768962859 -0.13 % │   501446666925   501273006209 +0.03 % │  498700  499780 -0.22 % │   27   62   -56.45 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│               coq-sf-vfa │   47.09   46.97 +0.26 % │   130099933254   129962571050 +0.11 % │   207371706792   207517654974 -0.07 % │  541424  541240 +0.03 % │   28    5  +460.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│       coq-mathcomp-field │  466.40  465.15 +0.27 % │  1297903378242  1294405862333 +0.27 % │  2116264939838  2116237781984 +0.00 % │  806224  806144 +0.01 % │   26  173   -84.97 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│               coq-geocoq │ 2831.54 2823.15 +0.30 % │  7871800923734  7847042831591 +0.32 % │ 12494447130981 12501538293674 -0.06 % │ 1287496 1287508 -0.00 % │   78  115   -32.17 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│                coq-color │  738.16  735.61 +0.35 % │  2044884478215  2038391235172 +0.32 % │  2457486823896  2457271169088 +0.01 % │ 1427120 1403428 +1.69 % │   39   30   +30.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│   coq-mathcomp-odd-order │ 1411.16 1406.24 +0.35 % │  3930719117268  3917830159293 +0.33 % │  6609170059357  6613347505806 -0.06 % │ 1352732 1349488 +0.24 % │    0    1  -100.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│                coq-gpfsl │ 1212.59 1207.96 +0.38 % │  3372855945403  3361912785978 +0.33 % │  4420510194225  4433426571100 -0.29 % │ 1623596 1624780 -0.07 % │   64    5 +1180.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼─────────────────────────┼──────────────────────┤
│                  coq-vst │ 1717.50 1710.32 +0.42 % │  4775242376923  4754806392535 +0.43 % │  6509603172926  6501311722459 +0.13 % │ 2223496 2222000 +0.07 % │  198  363   -45.45 % │
└──────────────────────────┴─────────────────────────┴───────────────────────────────────────┴───────────────────────────────────────┴─────────────────────────┴──────────────────────┘
```

PR #7855, opened 2018-06-18
Update section on adding your project to CI and link to example PR.

Zimmi48:
**Kind:** documentation

### DIFF
--- a/dev/ci/README.md
+++ b/dev/ci/README.md
@@ -47,16 +47,13 @@ CI.
 
 ### Add your development by submitting a pull request
 
-Add a new `ci-mydev.sh` script to [`dev/ci`](.) (have a look at
-[`ci-coq-dpdgraph.sh`](ci-coq-dpdgraph.sh) or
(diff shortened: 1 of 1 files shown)


SkySkimmer:
Are we dropping travis?

ejgallego (approved):
I am OK with this; we should discuss what to do with Travis, in my view I'd like to have it as backup at least until INRIA can provide custom workers, or `-local` goes away.

But I guess we can omit the Travis information for now, as it is likely that Travis will be only used for very specific targets in the future.

Zimmi48:
> Are we dropping travis?

Eventually, I think so. The recent ExtLib + QuickChick were approved without adding anything to Travis. I have been removing failing jobs from Travis so currently Travis is very incomplete compared to GitLab CI. I prefer not to recommend adding to Travis to avoid some later annoyance (like timeouts, not-up-to-date dependencies, etc). That doesn't mean that it's forbidden to add to Travis of course.

ejgallego:
@SkySkimmer are you ok with this?

SkySkimmer (approved):

PR #7856, opened 2018-06-18
Fix #7829: Spurious documentation failures.

### LINKED ISSUE #7829: Spurious documentation failures
We observe an alarming rate of failures of the documentation build on GitLab these days:

- https://gitlab.com/coq/coq/-/jobs/75178114
- https://gitlab.com/coq/coq/-/jobs/75139529
- https://gitlab.com/coq/coq/-/jobs/75139882
- https://gitlab.com/coq/coq/-/jobs/75143042
- https://gitlab.com/coq/coq/-/jobs/75164104

In all these cases, the error is the same:

```
Exception occurred:
  File "/usr/lib/python3/dist-packages/pexpect/expect.py", line 70, in timeout
    raise TIMEOUT(msg)
pexpect.exceptions.TIMEOUT: Timeout exceeded.
...
Error while sending the following sentence to coqtop: From Coq Require Import ssreflect ssrfun ssrbool List.
```

What is especially weird is that I'm under the impression that these failures started appearing after the merge of #7284 but what is weird is that this PR didn't touch the faulty line. It only added https://github.com/coq/coq/pull/7284/files#diff-d80ca9da497679f1161bfbf7e7ab7ce1R509 in some other place.
So I'm tempted by two solutions. Either separate these imports in several lines to make it less likely to reach timeout (which is already at 10s), or revert the change I just linked to, to see if it changes anything (in which case, it looks like a bug).

Zimmi48:
We split a Require Import in two to avoid reaching the timeout.

**Kind:** infrastructure.

Hopefully enough to fix #7829.

### DIFF
--- a/doc/sphinx/proof-engine/ssreflect-proof-language.rst
+++ b/doc/sphinx/proof-engine/ssreflect-proof-language.rst
@@ -3657,7 +3657,8 @@ selective rewriting, blocking on the fly the reduction in the term ``t``.
 
   .. coqtop:: reset
 
-     From Coq Require Import ssreflect ssrfun ssrbool List.
+     From Coq Require Import ssreflect ssrfun ssrbool.
+     From Coq Require Import List.
      Set Implicit Arguments.
      Unset Strict Implicit.
      Unset Printing Implicit Defensive.


Zimmi48:
@maximedenes This passed on GitLab CI. Would you like me to re-run the documentation build a couple times to check whether this really fixes the problem?

Zimmi48:
Passed twice, running a third time just to be sure.

Zimmi48:
Passed three times. Time to merge I think.

Zimmi48:
I never backported this because I never observed this problem in `v8.8`. This might have something to do with #7515 or with a change in Coq making `coqtop` process this `Require Import` more slowly.

PR #7866, opened 2018-06-19
Implementation of mutual records in the higher strata

### LINKED ISSUE #7183: Mutually coinductive records are not handled
Mutually coinductive records are currently not handled by the kernel. The reason is quite stupid if I understand correctly, as it is due to the fact extraction cannot handle them properly. This is a limitation that prevents the transition to coinductive negative types, so it should be swiftly lifted.

ppedrot:
Depends on #7750.

Fixes #7183.

This is the companion PR of #7750 that adds the handling of mutual record types to the higher layers, namely the horrendous files in `vernac` that are doing literally *n'importe quoi* due to the accumulation of minute changes across the span of several years. In retrospect adapting the kernel was easier. This allowed (or actually required) to clean up up this mess thoroughly. So in addition to a nice user-facing feature, we also have a probably slightly less buggy code now.

- [x] Entry added in CHANGES.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,10 @@
 Changes from 8.8.2 to 8.9+beta1
 ===============================
 
+Kernel
+
+- Mutually defined records are now supported.
+
 Tactics
 
 - Added toplevel goal selector ! which expects a single focused goal.
--- a/doc/sphinx/language/gallina-extensions.rst
+++ b/doc/sphinx/language/gallina-extensions.rst
@@ -23,8 +23,9 @@ expressions. In this sense, the :cmd:`Record` construction allows defining
 .. _record_grammar:
 
   .. productionlist:: `sentence`
-     record         : `record_keyword` `ident` [ `binders` ] [: `sort` ] := [ `ident` ] { [ `field` ; … ; `field` ] }.
+     record         : `record_keyword` `record_body` with … with `record_body`
      record_keyword : Record | Inductive | CoInductive
+     record_body    : `ident` [ `binders` ] [: `sort` ] := [ `ident` ] { [ `field` ; … ; `field` ] }.
      field          : `ident` [ `binders` ] : `type` [ where `notation` ]
                     : | `ident` [ `binders` ] [: `type` ] := `term`
 
@@ -167,12 +168,13 @@ and the syntax `term.(@qualid` |term_1| |term_n| `)` to `@qualid` |term_1| `…`
 In each case, `term` is the object projected and the
 other arguments are the parameters of the inductive type.
 
+
 .. note:: Records defined with the ``Record`` keyword are not allowed to be
    recursive (references to the record's name in the type of its field
    raises an  error). To define recursive records, one can use the ``Inductive``
    and ``CoInductive`` keywords, resulting in an inductive or co-inductive record.
-   A *caveat*, however, is that records cannot appear in mutually inductive
-   (or co-inductive) definitions.
+   Definition of mutal inductive or co-inductive records are also allowed, as long
+   as all of the types in the block are records.
 
 .. note:: Induction schemes are automatically generated for inductive records.
    Automatic generation of induction schemes for non-recursive records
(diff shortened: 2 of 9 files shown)


Zimmi48:
You feel like the refman itself doesn't need to be updated? Or is it because you feel the refman section on Records is so lame that you're worried that starting to extend it is going to get you to do much more work than you intended? (See also https://github.com/coq/coq/pull/7536#issuecomment-394630730).

ppedrot:
@Zimmi48 Yes, we need to seriously rework the record section. It's out of the scope of this PR I'm afraid.

Zimmi48:
I'm fine with leaving this out of this PR but we should have an open issue about it with a 8.9+beta1 milestone.

maximedenes:
Maybe I'm missing something, but I don't how such a feature could be added without at least being mentioned in the manual.

ppedrot:
Would a short sentence saying "Mutual records are also allowed" would please you? That's about what is done for coinductive types in the refman.

Zimmi48:
> Maybe I'm missing something, but I don't how such a feature could be added without at least being mentioned in the manual.

It should be added to the manual but given that the section on records needs much more work, this could be done at the same time.

Zimmi48:
> Would a short sentence saying "Mutual records are also allowed" would please you? That's about what is done for coinductive types in the refman.

No, I wouldn't be satisfied by this. A grammar of records is given in the manual and this grammar doesn't show the syntax for mutual records. (And BTW I think this grammar should be merged back into the grammar for Inductive in the first chapter).

maximedenes:
@Zimmi48 I think we are doing this too often: we postpone something that should be done in a PR, we open an issue, never schedule the work or even say who's in charge, and of course the subsequent change is never done, leave the system in an undesirable state.

Even if we can't make the documentation very good now, we should update all mentions of records that seem to indicate they cannot be mutual (like in the grammars), and add a short mention like @ppedrot suggested.

Zimmi48:
> never schedule the work or even say who's in charge

I don't think that's true. I think that when we have insisted that an issue be open about missing documentation with a proper milestone, this has worked OK.

But I'm not against adapting the doc now, however, I simply won't be satisfied with a small mention like the one @ppedrot proposes.

That being said, I really think the state of the record documentation really needs work and I would have loved this to be an opportunity to force one of you to work on it (I feel like I'm the only one who's trying to make the doc better these days :wink: ).

maximedenes:
> That being said, I really think the state of the record documentation is bad and this part of the doc really needs work and I would have loved this to be an opportunity to force one of you to work on it (I feel like I'm the only one who's trying to make the doc better these days  ).

I agree that the documentation is a priority. The problem is that we are overloaded, and it is not surprising, since we are a few developers trying to maintain 500000 lines of code... These days I'm working on what I think is another priority, namely the standard library. So I'm glad that you are improving the documentation, but I don't think it makes sense to try to "force people to work on it" when they would already be happy to do so.

ppedrot:
So what should I do regarding this PR then?

Zimmi48:
@maximedenes seems to think that you should update the doc as part of this PR, and this implies extending the grammar of records in the extensions to gallina chapter.

herbelin (commented):
Very nice initiative. This was indeed missing for elegance and completeness of the design.

Otherwise, maybe I missed it in which case I apologize but could you tell more about your vision on what would be the best to do in terms of elegance, compatibility, expressivity and efficiency of inductive and coinductive types in the system? (See also the discussion [here](https://github.com/coq/coq/pull/7536#issuecomment-394773825)).

ppedrot:
I have minimally updated the doc to mention mutually (co)recursive records.

Zimmi48:
Thanks!

maximedenes (dismissed):
Looks good. The documentation update is sufficient IMHO. However, in the test case, can you seize the opportunity of adding some VM computations on such records? I guess it wasn't done as part of the kernel counterpart of this PR for lack of user syntax, so now is the time.

maximedenes (approved):

PR #7868, opened 2018-06-19
[coqtop] Give priority to stdlib load path over current directory

maximedenes:
When initilazing the load path, coqtop adds implicit bindings for stdlib
and for the current directory (-R stdlib Coq -R . ""). In case of
a clash, the binding of the current directory had priority, which makes
it impossible to edit stdlib files (when they Require files from the
same directory) using PG, or from CoqIDE when launched from the directory
containing the file.

Example to reproduce the problem:

```
cd plugins/omega
coqide Omega.v
```

some of the Requires will fail.

### DIFF
--- a/toplevel/coqinit.ml
+++ b/toplevel/coqinit.ml
@@ -92,6 +92,14 @@ let libs_init_load_path ~load_init =
   let coqpath = Envars.coqpath in
   let coq_path = Names.DirPath.make [Libnames.coq_root] in
 
+  (* current directory (not recursively!) *)
+  [ { recursive = false;
+      path_spec = VoPath { unix_path = ".";
+                           coq_path = Libnames.default_root_prefix;
+                           implicit = false;
+                           has_ml = AddTopML }
+    } ] @
+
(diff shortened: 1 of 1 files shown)


ejgallego:
If this breaks no CI I am OK with the patch, anyways the mess we have there is epic.

maximedenes:
@ejgallego the three failures look unrelated: two of them are no space left on kaguya, the third is an ML compilation error in Ltac2.

ejgallego:
I'll push this change but the whole thing is very suspicious; hopefully I can remove the all hardcoded path loading from Coq soon.

ejgallego (approved):
No harm in doing the swapping so it seems, however the whole semantics of path loading needs to be tweaked.

maximedenes:
> I'll push this change but the whole thing is very suspicious; hopefully I can remove the all hardcoded path loading from Coq soon.

I agree, this is just a minimal change that should improve the situation locally. But a larger clean-up would be well deserved.

PR #7798, opened 2018-06-13
Remove hack skipping comparison of algebraic universes in subtyping.

SkySkimmer:
When inferring [u <= v+k] I replaced the exception and instead add
[u <= v]. This is trivially sound and it doesn't seem possible to have
the one without the other (except specially for [Set <= v+k] which was
already handled).

I don't know an example where this used to fail and now succeeds (the
point was to remove an anomaly, but the example
~~~coq
Module Type SG. Definition DG := Type. End SG.
Module MG : SG. Definition DG := Type : Type. Fail End MG.
~~~
now fails with universe inconsistency.

Fix #7695 (soundness bug!).

PR #7869, opened 2018-06-19
Fix Univ.enforce_leq dropped constraints when algebraic on the right

SkySkimmer:
There's probably a proof of false using subtyping if someone wants to
look.

NB: the checker doesn't handle algebraics on the right.

### DIFF
--- a/kernel/univ.ml
+++ b/kernel/univ.ml
@@ -674,12 +674,7 @@ let check_univ_leq u v =
   Universe.for_all (fun u -> check_univ_leq_one u v) u
 
 let enforce_leq u v c =
-  let rec aux acc v =
-  match v with
-  | v :: l ->
-    aux (List.fold_right (fun u -> constraint_add_leq u v) u c) l
-  | [] -> acc
-  in aux c v
+  List.fold_left (fun c v -> (List.fold_left (fun c u -> constraint_add_leq u v c) c u)) c v
 
 let enforce_leq u v c =
   if check_univ_leq u v then c


SkySkimmer:
I don't know why fold_right was used but fold_left is standard when the order doesn't matter.

Zimmi48:
Hum, the code is broken.

Zimmi48:
@SkySkimmer VST passed so I think it would make sense that you rebased #7798 on top of this one to check whether removing this instability is enough to make it go through

mattam82 (approved):
That’s pretty bad indeed, and certainly triggerable through module subtyping.

SkySkimmer:
>There's probably a proof of false using subtyping if someone wants to
look.

On second thought I think it's sound, when doing `max(a_i){i<=k} <= max(b_j){j<=l}` it becomes `max(a_i){i<=k} <= b_l` so it picks a random level to be smaller than based on the hash order. The main issue is the unstability of the algorithm.

mattam82:
Right, slightly restrictive and really unstable but it was sound indeed.
Le mar. 19 juin 2018 à 22:38, Gaëtan Gilbert <notifications@github.com> a
écrit :

Zimmi48:
@maximedenes CI failures are all unrelated: can you merge?

SkySkimmer:
>@maximedenes CI failures are all unrelated: can you merge?

@mattam82 is maintainer for the univ.ml (with me but I'm PR author ^^) though?

Zimmi48:
Sorry I meant to ping @mattam82 but I was probably bitten by GitHub auto-complete.

Issue #7872, opened 2018-06-20
omega, romega, lia: pick one


maximedenes:
The redundancy between these three tactics is putting some burden on maintainers of the system. For example, #186 is much more costly to implement than it should be. I believe the situation is also confusing newcomers. I would like to understand what is the plan to resolve this redundancy, and help implement it.

Here's my understanding of the situation:
- The implementation of `romega` was started 17 years ago (according to its header). It is meant as a reflection-based replacement of `omega` but hasn't reached full compatibility. I don't know if it is the sign of serious difficulty in resolving the incompatibilities.
- `romega` is undocumented
- when in the compatible fragment, `romega` should always be preferred to `omega` (reflection should bring some performance improvements).
- `lia` has incomparable power (sometimes it can do more than `(r)omega`, sometimes less (that's what I thought, although the manual seems to say that it is strictly more powerful).
- the three tactics are basically not maintained externally

I believe our manpower is way too limited (compared to the size of the code base) to afford maintaining the three. Note however that `lia` is part of the larger `micromega`, for which I don't think there exists a general replacement. Not maintaing two of them doesn't necessarily mean making them disappear, but rather moving them out of the Coq repo and calling for external maintainers.

So I believe the first question that needs to be evaluated is: can `lia` be used as a replacement for `(r)omega` in practice today? I believe @RalfJung experimented it recently, maybe he can give some feedback.

If the answer is no, then maybe we should favor `romega`, starting with writing down all known incompatibilties that have been resisting for all those years.

@ppedrot do you know if it would be hard to rebind `(r)omega` to `lia`, so that we could see what it gives on our CI?

Zimmi48:
I think @letouzey probably has some experience to share on this question.

RalfJung:
> I believe @RalfJung experimented it recently, maybe he can give some feedback.

You must be confusing me with someone... I don't remember ever doing experiments like that.^^

However, what I can say is that in Iris, was prefer to use `lia` because it is much faster than `omega`. I did not know about `romega`.

maximedenes:
@Zimmi48 I don't know how I forgot to cite Pierre's nickname, he's clearly the one who can answer some of the questions above. Thanks!

@RalfJung Really? I remember a discussion on gitter where somebody was impressed by the performance of `lia` and started using it in a development as a replacement for `omega`. I thought it was you in Iris. If not, maybe @Zimmi48 remembers, I believe he was part of the discussion too. But maybe that also never happened :)

[...]

PR #7878, opened 2018-06-20
Remove omega and romega.

maximedenes:
We move `zify` to `micromega` and rebind `omega` and `romega` to `lia`.

Lia cannot be boostrapped in exactly the same way as omega, so the
stdlib will have to be adapted so as not to rely on lia inside ZArith.
Laurent Théry offered to remove the admits there.

This is work in progress, but it is looking good: CompCert has around 3000 calls to `omega` and requires only a 4-line patch.


ppedrot:
I'd be curious to see a bench also!

andres-erbsen:
If other benchmarks look promising, please benchmark this change on https://github.com/mit-plv/bedrock2 and dependencies even though these are not normally a part of CI/benchmarks. Or just tell me when and how I should do it. Thanks!

JasonGross:
Maybe we should add bedrock2 to opam and the CI?

JasonGross (changes requested):
As codeowner of the Compat files (is that all I'm requested for here?  Is there a way to get GitHub to tell me?), I approve.

Separately, this needs updates to the documentation.  Also maybe `omega` and `romega` should emit deprecation warnings?  Also is it worth keeping `romega` alive as a separate plugin on opam?  (Do we know where `romega` and `lia` don't overlap?)

JasonGross:
@andres-erbsen  Note that bedrock2 cannot currently be benched because it will fail to build (by construction) for any `coqc` not reporting as 8.7.2.

maximedenes:
> Separately, this needs updates to the documentation. Also maybe omega and romega should emit deprecation warnings? Also is it worth keeping romega alive as a separate plugin on opam? (Do we know where romega and lia don't overlap?)

Sure, this is in progress as indicated by the corresponding label.

plugins/btauto/Algebra.v
```diff
@@ -84,9 +84,9 @@ Ltac case_decide := match goal with
   let H := fresh "H" in
   define (@decide P D) b H; destruct b; try_decide
 | [ |- context [Pos.compare ?x ?y] ] =>
-  destruct (Pos.compare_spec x y); try (exfalso; zify; romega)
+  destruct (Pos.compare_spec x y); try (exfalso; lia)
```
fajb: In exfalso; lia, exfalso is redundant for lia. This would only be an optimisation if the conclusion is a complex arithmetic formula.

plugins/btauto/Algebra.v
```diff
@@ -417,14 +417,14 @@ Qed.
 Hint Extern 5 =>
```
fajb: This whole tactic can be optimised away and directly replaced by lia.
My guess is that zify improved and is now dealing with Pos.max

fajb (approved):
This is a bold move but that could be the time to make it. 
As I told @maximedenes  in a private message both omega and lia have known shortcoming -- meaning they are incomplete for linear arithmetics.

Having a drop-in replacement will require some work and extensive benchmarking.

thery:
Ok I think I've removed all the "admit"

maximedenes:
@thery Impressive! Thanks a lot for your help :)

herbelin (commented):
Same remarks as for `fourier`. There has been work, sweat, passion put in developing the different plugins. Whatever decision is taken so as to combine the respective strengthes of every work in a common framework, so as to simplify the pedagogy, so as to simplify the maintainance, so as to improve our understanding of the underlying techniques, I wish that it is done in a way respectful of the work of everyone done in the past.

Zimmi48:
> I wish that it is done in a way respectful of the work of everyone done in the past.

More than the initial author, the one who could be bothered by the removal of `romega` at this point is @letouzey who has taken time to clean it up last summer. But he's already said:

> So for now I'd vote for romega, even if the actual winner will probably be the one able to replace omega without too much drama.

(cf. https://github.com/coq/coq/issues/7872#issuecomment-398814289)

Zimmi48:
Here as well, if deemed valuable, `romega` could continue being maintained as part of https://github.com/coq-community...

bollu:
For some more context, I actually benchmarked on the files linked by @maximedenes , and here were the results on my machine:

```
(** With  lia **)
╭─bollu@cantordust ~/work/dependence-analysis-coq/dependence-loop  ‹master*›
╰─$ time coqc proof.v
coqc proof.v  6.41s user 0.17s system 99% cpu 6.583 total

(** Stash changes **)
╭─bollu@cantordust ~/work/dependence-analysis-coq/dependence-loop  ‹master*›
╰─$ git stash
Saved working directory and index state WIP on master: 814382e add example of creating a _coinductionn principle_ from gitter
HEAD is now at 814382e add example of creating a _coinductionn principle_ from gitter

(** run with omega **)
╭─bollu@cantordust ~/work/dependence-analysis-coq/dependence-loop  ‹master*›
╰─$ time coqc proof.v
coqc proof.v  2.80s user 0.17s system 99% cpu 2.971 total
```

This is embedded in the commit message https://github.com/bollu/dependence-analysis-coq/commit/bfa4d23f93817cb71088599699e28cff0189489f

ppedrot (approved):

gares (approved):
makefile changes are ok

Zimmi48:
Rebase when you push. Your branch is not up-to-date with master so in order to avoid various problem, GitLab CI won't start. See also #8079.

samuelgruetter:
@maximedenes why do you have to make PRs to many repos on the CI? Do you want this to be a breaking change with the next Coq version? That is, will all projects updating to the next Coq version have to do diffs of the form

```
-Require Import Coq.omega.Omega.
+Require Import Coq.micromega.Lia.
```

in all their files? Why do you not keep a file that I can import with `Require Import Coq.omega.Omega.` which defines `omega` as an alias for `lia`, and emits a deprecation warning?

maximedenes:
@smauelgruetter Good question :) That's not the most common diff, in fact I usually replace : `Require Import Coq.omega.Omega.` with `Require Import Coq.ZArith.ZArith.`. I think it is a sound change anyway.

We have no machinery to emit a deprecation warning at `Require`, so unfortunately I couldn't proceed as you suggest.

But anyway, many patches I'm submitting are fixing scripts that relied on incompleteness bugs of `omega` (basically, I have to remove a few lines of script here and there).

maximedenes:
I guess we could, but it will leave some noise in our `Makefile`, and I'm not sure what it would bring since:
- we won't have a way to generate a deprecation warning, so the problem of removing this file will just be delayed, not solved
- the change from `Require Coq.omega.Omega` to `Require Coq.ZArith.ZArith` is backward compatible, so I don't see why it couldn't be done now downstream.

samuelgruetter:
Here's a way to emit a deprecation warning: https://github.com/maximedenes/coq/pull/62
What do people think about it?

herbelin:
In passing, after so many years, it is great that an observational superset of `omega`, `romega` of `lia` is finally found, which integrates the best of what we know how to provide. Hats off.

Zimmi48:
>I don't think this is doable because external developments precisely rely on what the standard library does, through leakage of hints...

I only meant removing the dependency on `omega` in opaque proofs, not in hint definitions or such.

herbelin:
What is the status of this PR? Is this 8.9 milestone? Also, omega and romega do not occur anymore in the CODEOWNERS file, am I supposed to be the assignee? Can someone make a summary of what it changes user side in practice?

Zimmi48:
Given how close the branching is now I would recommend only deprecating omega / romega in 8.9 and doing the actual removal (without compatibility layer) after the branching.

silene:
While experimenting with the replacement of `omega` by `lia`, I have encountered yet another issue with `zify`, namely #8399 . While I am in favor of removing `omega` and `romega`, we cannot replace them with `lia` until the issues with `zify` are fixed. Perhaps an intermediate solution would be to replace them by

```
Ltac omega := unfold Z.succ in * ; unfold Z.pred in * ; xlia zchecker.
```

which is `lia` without `zify`.

pi8027:
Now, `lia` seems to be able to solve all the problems that `omega` can solve. https://github.com/coq/coq/labels/part%3A%20micromega

Zimmi48:
Maybe it's time to officially deprecate omega then (in 8.11 I suppose, because Coq 8.10 did not receive all the recent lia improvements).

pi8027:
> Given how close the branching is now I would recommend only deprecating omega / romega in 8.9 and doing the actual removal (without compatibility layer) after the branching.

> Maybe it's time to officially deprecate omega then (in 8.11 I suppose, because Coq 8.10 did not receive all the recent lia improvements).

@Zimmi48 I agree with you. If we deprecate them without actual removal, we could receive new issues with detailed comparisons of `lia`, `omega`, and `romega`, like #9921. If we remove `omega` and `romega` silently, we will lose a chance to identify hidden issues of Micromega.

Zimmi48:
`romega` has already been removed silently though

ppedrot:
Isn't this basically superseded by #11286? The rebase of the current PR looks like it's going to be a mess...

Zimmi48:
Superseded by #13741.

PR #7888, opened 2018-06-21
Clarify the message "this hint will only be used by eauto"

Armael:
**Kind:** documentation (user message improvement)

This clarifies slightly the message "this hint will only be used be eauto", that is printed when doing `Hint Resolve` on a hint that requires instantiating evars to be applied.

The new message is: `the hint: ... will only be used by eauto, because it requires instantiating evars`.
Bikeshedding on the exact text of the clarified message is of course welcome.

### DIFF
--- a/tactics/hints.ml
+++ b/tactics/hints.ml
@@ -828,38 +828,48 @@ let make_exact_entry env sigma info poly ?(name=PathAny) (c, cty, ctx) =
 
 let make_apply_entry env sigma (eapply,hnf,verbose) info poly ?(name=PathAny) (c, cty, ctx) =
   let cty = if hnf then hnf_constr env sigma cty else cty in
-    match EConstr.kind sigma cty with
-    | Prod _ ->
-        let sigma' = Evd.merge_context_set univ_flexible sigma ctx in
-        let ce = mk_clenv_from_env env sigma' None (c,cty) in
-	let c' = clenv_type (* ~reduce:false *) ce in
-        let pat = Patternops.pattern_of_constr env ce.evd (EConstr.to_constr ~abort_on_undefined_evars:false sigma c') in
-        let hd =
-	  try head_pattern_bound pat
-          with BoundPattern -> failwith "make_apply_entry" in
-        let nmiss = List.length (clenv_missing ce) in
-        let secvars = secvars_of_constr env sigma c in
-	let pri = match info.hint_priority with None -> nb_hyp sigma' cty + nmiss | Some p -> p in
-	let pat = match info.hint_pattern with
-	| Some p -> snd p | None -> pat
-	in
-	if Int.equal nmiss 0 then
-	  (Some hd,
-          { pri; poly; pat = Some pat; name;
-            db = None;
-            secvars;
-            code = with_uid (Res_pf(c,cty,ctx)); })
-	else begin
-	  if not eapply then failwith "make_apply_entry";
-          if verbose then
-	    Feedback.msg_info (str "the hint: eapply " ++ pr_leconstr_env env sigma' c ++
-	    str " will only be used by eauto");
-          (Some hd,
-           { pri; poly; pat = Some pat; name;
-	     db = None; secvars;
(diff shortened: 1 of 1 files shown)


ppedrot:
I don't think that *instantiating* is the proper term, but I cannot see how to phrase it in a better way. Any ideas, people?

SkySkimmer:
What's an example of when this happens?

Armael:
Artificial example:
```coq
Hypothesis L : forall (x y z: nat),
  x = y ->
  y = z.

Hint Resolve L.
```

aspiwack:
In a discussion on Gitter, we became convinced that speaking of instantiating was misleading here because the problem was rather evars which were _not_ instantiated. There was a bit of a back and forth to find a phrasing which was both technically correct and intuitive to grasp. We came up with the following:

> The hint Foo will only be used by eauto, because applying Foo would leave variables a, x, y as unresolved existential variables.

For instance, in your lemma `L` above:

> The hint L will only be used by eauto, because applying L would leave variable x as an unresolved existential variable.

(keeping Coq's traduction to have messages specialise to the singular form)

SkySkimmer:
So it's because there is a variable which appears in an hypothesis yet not in the goal?

Armael:
@SkySkimmer yes, AFAICT

I like both @gmalecha and @aspiwack suggestions. However, by looking at the code I have no clue how to retrieve the names of the "unresolved variables". Could anyone give me a hint?

Armael:
Ok, I just pushed a new version of the patch, using @aspiwack wording. I also took the liberty of re-indenting the whole function, in order to get rid of tabulations.

ppedrot (approved):

Zimmi48:
Ideally, it would be great to have an output test for this, but I don't want to make your life harder as a contributor, so not a strong request.

Cf. https://github.com/coq/coq/tree/master/test-suite#adding-a-test

Armael:
That's fine, I'll add some output tests later today.

tactics/hints.ml
```diff
@@ -828,38 +828,50 @@ let make_exact_entry env sigma info poly ?(name=PathAny) (c, cty, ctx) =
...
+      (Some hd,
+       { pri; poly; pat = Some pat; name;
+         db = None;
+         secvars;
+         code = with_uid (Res_pf(c,cty,ctx)); })
+    else begin
+      if not eapply then failwith "make_apply_entry";
+      if verbose then begin
+        let variables =
+          if nmiss > 1 then str "variables" else str "variable"
```
Zimmi48: I think there is a standard plural function defined somewhere for this.

ppedrot:
@Armael You should rebase also, to please the CI.

gares:
I like the current formulation

Armael:
Rebased on top of master + used `CString.plural`

ppedrot:
The reason the message does not appear is that you use `msg_info` rather than `msg_notice`. This looks semantically correct to me, the former being for additional information and the latter for stuff that the user requested to be printed. I don't know whether we should adapt the test-suite makefile to be more verbose or just do not add an output test for this feature.

SkySkimmer:
Maybe we could pass -verbose or whatever the option was for that one test file (the test suite already has support for per file arguments by parsing the emacs file local stuff).

Zimmi48:
Good idea @SkySkimmer, although I realize that this is something else that is not documented as part of https://github.com/coq/coq/blob/master/test-suite/README.md.

Zimmi48:
The first line of the test file would need to be:

```
(* -*- mode: coq; coq-prog-args: ("-verbose") *)
```

Zimmi48:
As reported by @Armael on Gitter the `-verbose` doesn't exist.
The one we should use here is `-lv` or `-load-verbose`, however this is supposed to take the filename as argument and replace the `-l` option that is normally used to load it.
Consequently, I don't think that this solution works either. Sorry @Armael for having proposed to add a test case and delayed this PR in vain.

ppedrot:
OK let's merge then.

PR #7889, opened 2018-06-21
Cleanup reduction effects: they only work on constants.

SkySkimmer:
We only ever call `reduction_effect_hook` on constants, so there's no
point allowing it to be declared with globrefs. There is also no point
using a constr map instead of constant map.

(Technically there was a call of the effect hook on projections, but
that can never match a globref so it was useless)

### DIFF
--- a/pretyping/tacred.ml
+++ b/pretyping/tacred.ml
@@ -539,7 +539,7 @@ let reduce_mind_case_use_function func env sigma mia =
 let match_eval_ref env sigma constr stack =
   match EConstr.kind sigma constr with
   | Const (sp, u) ->
-     reduction_effect_hook env sigma (EConstr.to_constr sigma constr)
+     reduction_effect_hook env sigma sp
         (lazy (EConstr.to_constr sigma (applist (constr,stack))));
      if is_evaluable env (EvalConstRef sp) then Some (EvalConst sp, u) else None
   | Var id when is_evaluable env (EvalVarRef id) -> Some (EvalVar id, EInstance.empty)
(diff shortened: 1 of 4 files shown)


ppedrot (approved):
Very nice cleanup. Term comparison should be haram, and this patch removes a dependency on the constant -> projection encoding, so it's all good for me.

SkySkimmer:
It also removes a `constr_of_global`, so you can declare effects on polymorphic constants ^^

herbelin (commented):
OK to me.

pretyping/tacred.ml
```diff
@@ -550,16 +550,14 @@ let match_eval_ref env sigma constr stack =
...
-     reduction_effect_hook env sigma (EConstr.to_constr sigma constr)
+     reduction_effect_hook env sigma sp
         (lazy (EConstr.to_constr sigma (applist (constr,stack))));
     if is_evaluable env (EvalConstRef sp) then
       let u = EInstance.kind sigma u in
       Some (EConstr.of_constr (constant_value_in env (sp, u)))
     else
       None
   | Proj (p, c) when not (Projection.unfolded p) ->
-     reduction_effect_hook env sigma (EConstr.to_constr ~abort_on_undefined_evars:false sigma constr)
```
gares: I've no clue what "reduction_effect_hook" is, but this line seems a change in semantics

SkySkimmer: See commit message.

gares (approved):

SkySkimmer:
@mattam82 ping

ppedrot:
I am taking over, @mattam82 is overloaded and I don't think this simple patch deserves waiting any further.

PR #7890, opened 2018-06-21
Inline a function from Quote used in setoid_ring.

ppedrot:
The code was wrong as it relies once again on term equality and fails on polymorphic constants. Quote is bound to disappear, so we write a correct version of this 10-line function in setoid_ring.


maximedenes (approved):
Nice clean-up!

ppedrot:
@maximedenes You should probably assign this PR to yourself, the change is quite trivial.

maximedenes:
Ok, I hope @amahboubi doesn't mind.

PR #7897, opened 2018-06-22
Remove fourier plugin

maximedenes:
As stated in the manual, the fourier tactic is subsumed by lra.

Porting the standard library did not reveal any incompatibility, except in a couple of places relying on the weakness of `fourier`, where `lra` is a bit more powerful.

I will provide a compatibility layer, but I'd like to see a first CI run.


Zimmi48 (dismissed):
I don't like this compatibility layer (`Ltac fourier := lra.`) at all (same is true for `omega` / `romega` BTW).
I'd rather deprecate `fourier` in 8.9, advising to use `lra` instead, then merge this PR in 8.10 (it's not so far away, little more than 2 summer months away from the branching) without any compatibility layer.

Zimmi48:
Dropping right now would be also OK given that this is a plugin that could be maintained somewhere else. That's really the redefinition of `fourier` that I don't find right.

maximedenes:
> Dropping right now would be also OK given that this is a plugin that could be maintained somewhere else. That's really the redefinition of fourier that I don't find right.

I plan to implement a deprecation machinery to print a warning signalling users that this rebinding is done and they should update their code.

We are tracking 4 million of lines in our CI (including Coq), I don't think it is reasonable to ask to maintain patches for 2 months... Here it's still not so bad, but for `omega`...

Zimmi48:
> We are tracking 4 millions of lines in our CI (including Coq), I don't think it is reasonable to ask to maintain patches for 2 months...

I guess we can see your patches as an assessment of the feasibility to remove such and such tactic. However, it's not clear why we (you in this case) should be the ones to patch all this code in the end. If we end up deprecating the tactics first, and your patches have got conflicts in the meantime, then you could submit them as-is and ask people to adapt them on their own. In fact, since they are backward compatible patches, you could even submit them now. It will reduce the number of deprecation warnings to fix for external project maintainers when they start using a version of Coq in which the given tactic is marked as deprecated.

In any case, I don't think that this is a good argument to rush into doing things :wink:

maximedenes:
We discussed a bit more offline. One piece of information that @Zimmi48 didn't have is that other patches will be made much simpler after this work, like #186 (and more generally, refactorings that impact many plugins).

herbelin (commented):
I'd like to see evidences that removing plugins like `fourier` will significantly help in #186. Could you provide some? On the psychological note, I'm also wondering whether you did talk with Loïc Pottier. Does he support the view that you are defending?

Otherwise, I have no strong opinion. On one side, the plugin is in practice easy to maintain, so, no need to remove it and/or break the compatibility per se. On the other side, simplifying the pedagogy and the code base by providing only one instance of Fourier's algorithm which subsumes what all of `fourier` and `lra` are able to do, is certainly valuable. So, my synthetic opinion is that as soon as everyone is satisfied and respected in his work, I rally to the majority.

In any case, comparing the exact strength of `fourier` and `lra` is something to be done, so this is also a reason to support this PR. Could you by the way comment on the way reification is done in each of the two plugins?

Ah, another issue: I find the name `fourier` easier to remember than `lra`.

ppedrot:
@herbelin On a somewhat unrelated note, I agree with the fact that I have a hard time remembering what are all those three-letter tactics. I don't know if I'm the only one out there, yet the shortage of characters from the Great Perl War is far behind us.

fajb:
I will only comment on the names
- Fourier is the name of a mathematician and an algorithm. 
- lra comes the SMT (also three letters) community.
http://smtlib.cs.uiowa.edu/logics.shtml
This is the name of a logic.

Unfortunately, currently both plugins are using a variant of fourier elimination instead of  a more scalable simplex implementation.

herbelin:
> Unfortunately, currently both plugins are using a variant of fourier elimination instead of a more scalable simplex implementation.

By curiosity, what do you mean by "more scalable"?

JasonGross:
I assume "more scalable" means "better asymptotic complexity" (but please correct me if I'm wrong).

fajb:
fourier elimination has exponential complexity in the number of  constraints/variables. 
This is a worst case that shows in practice.
Despise a worst-case exponential complexity, the practical complexity of the simplex is very good.
All state-of-the-art SMT solvers implement a simplex...

herbelin:
I think we already had this discussion actually. Using the word `fourier` would become misleading if the algorithm for "linear real arithmetic" is eventually reimplemented using a simplex. Fair enough.

Side (naive) question: would it make sense to have `lra` and `lia` merged? I.e. that a unique tactic indifferently solves integer or real arithmetic (and maybe even combinations of them)?

fajb:
That would make sense and would reduce further the number of tactics!

What would be possible without much effort would be to rify the goal (like zify but over R) and then run the existing lra. However, that would not be complete as lra does not exploit that certain variables are integral.

Getting a full combination would require some work. I would need to give it some thoughts  but that seems doable. 
This would also have the additional advantage of simplifying the code: currently there is code for R, Q and Z. With the drawback that R has axioms whereas Q and Z are axiom-free.

Zimmi48:
Evaluating whether we can remove some tactics and what would be the impact on performance is good. Removing tactics / plugins from the main official Coq repo is good to reduce the maintenance cost (and honestly we should seize every chance we have of reducing the maintenance cost when it has low impact on our users), and this does not represent a dismissal of the value of such past contributions (the same is true when we rewrite a pre-existing feature in a new way). And if the plugin is still considered valuable to keep maintaining, this can still be done externally (for instance as part of https://github.com/coq-community).

maximedenes:
> On the psychological note, I'm also wondering whether you did talk with Loïc Pottier. Does he support the view that you are defending?

As a funny coincidence, Loïc came to visit us today, so I took a few minutes to tell him. He said "feel free to do what you like with the code".

herbelin:
I'm glad to learn that Loïc feels detached about `fourier`. I actually did not mean that he would take it bad. My recommendation was more general. I believe that it is always better to justify one's position rather than behaving looking arbitrary (by the way, weren't you one of those to blame inria selection committees for their opacity a couple of years ago?).

Back to Loïc, how is he doing? I always appreciated the contributions he made to Coq.

Zimmi48:
> by the way, weren't you one of those to blame inria selection committees for their opacity a couple of years ago?

I think this has no connection whatsoever with the current discussion. We are not having a chat between developers. This is a public place.

> Back to Loïc, how is he doing?

Better to ask this on Gitter.

Zimmi48:
Back to the actual topic of the PR:

> As stated in the manual, the fourier tactic is subsumed by lra.

is a justification for the PR. There is no "behaving looking arbitrary".

herbelin:
> Getting a full combination would require some work. I would need to give it some thoughts but that seems doable.
> This would also have the additional advantage of simplifying the code: currently there is code for R, Q and Z. With the drawback that R has axioms whereas Q and Z are axiom-free.

Actually, and more generally, how difficult would it be to add a generic level for binding any implementation of N, Z, Q, R, ... to the internal language of micromega (like it is for `ring`). As far as I understand, the reification file `coq_micromega.ml` is quite generic already with these `zop_table` and co. How realistic would it be to move the building of these tables from the ML level to the Coq level (e.g. via type classes)?

Of course, I'm not asking in the context of this PR. It is more to have an idea of where we could eventually go.

herbelin:
@zimmi48: I think my questions and analogies are relevant to explain the message I'm trying to transmit. What alternative ways would you propose?

Zimmi48:
> I think my questions and analogies are relevant to explain the message I'm trying to transmit.

I think they are not and I suggest you always consider in front of what public you are speaking. In particular, what each of us may think of Inria's selection committee is clearly a private question. Furthermore, I think you are being unfair to Maxime in saying that his proposal looks arbitrary. Be careful Hugo: you want us to show consideration and respect for the work of past contributors, but at times your remarks can sound dismissive of the work of present contributors. That's not a good way of attracting new motivated contributors (and keeping the ones we already have)...

herbelin:
@Zimmi48: so, we continue the discussion online or offline?

In any case, there is nothing "secret" in saying that inria committees apply rules which they don't necessarily explain. I was just trying to explain using a mirror argument why I was considering that it is important to develop convincing justifications.

And I don't see what is wrong in asking a developer (or an external contributor) to develop arguments. I guess this is the rule of the game, and it is worth even in non-scientific contexts

Zimmi48:
> we continue the discussion online or offline?

Either way, but not in this PR thread. I'd prefer offline though.

> And I don't see what is wrong in asking a developer (or an external contributor) to develop arguments. I guess this is the rule of the game, and it is worth even in non-scientific contexts

Sure, there's nothing wrong in general, but be careful how you express such a request. The way you are talking to Maxime is... strange.

fajb:
Having a ring-like interface would be great: it does not seems completely straightforward.
A difficulty is to recognise constants for instance in Q and  build from it an Ocaml rational. 

Also, the reification of R is quite special as the a term like  1 / 3 is recognised as a rational constant even if it is written Rmult R1 (Rinv (Rplus R1 (Rplus R1 R1))) but maybe this would be the case for other R-like structures.

For those who agree to build on top of the standard library, an extensible tactic à la zify may be another way be extensible without writing ML code.

ppedrot (approved):

maximedenes:
> On a somewhat unrelated note, I agree with the fact that I have a hard time remembering what are all those three-letter tactics. I don't know if I'm the only one out there, yet the shortage of characters from the Great Perl War is far behind us.

@ppedrot We could think of a renaming at the same time, to minimize incompatible changes. What name would you like?

ppedrot:
I don't know, something like `rarith` or similar.

maximedenes:
The bench didn't show any change: https://ci.inria.fr/coq/job/benchmark-part-of-the-branch/468/console

Some developments are using `fourier` there: flocq, coquelicot (I think).

Compatibility is excellent (only one small patch, for CoRN).

I can prepare patches for developments, porting them to `lra` so that we don't rely on the compatibility layer. Then we can merge this PR either with or without the compat layer. @Zimmi48 what has your preference?

silene:
> I don't know, something like `rarith` or similar.

As @fajb explained, the names `lia` and `lra` come from an official nomenclature (more precisely, the official names are `qf_lia` and `qf_lra`). I am not too fond of Coq rolling its own nomenclature, especially one that is less precise than the widely accepted one. In particular, the `rarith` name no longer carries the `linear` part, which makes the tactic conflict with the `nra` tactic (which, as you can guess, is just short for the official `qf_nra` theory).

maximedenes:
Ok, this should be ready. It doesn't require any overlays anymore. @gares can you be the assignee?

maximedenes:
The failures seem spurious.

gares (approved):

maximedenes:
Thanks!

Zimmi48:
@maximedenes When porting qarith-stern-brocot following the removal of this plugin, there was only one incompatibility to solve Zimmi48/qarith-stern-brocot@226327e84b312ca5c5feaf070869798f2798c29e, replacing all `fourier` by `lra` is trivial Zimmi48/qarith-stern-brocot@653b029f2f3a0a221263fb6f7071894af8894e68. However, in this commit I can't remove `Require Import Fourier` without introducing an incompatibility (likely because of scopes being opened or things like that...).

maximedenes:
Indeed, you may need to add `Open Scope real_scope` somewhere. But IMHO it is probably a bug of the development you are looking at to rely on the scopes internally used by `Fourier`. The scoping of scopes (pun intended) makes it hard to detect such situations, I hope to improve on that independently.

Zimmi48:
A bug, but not one easily noticed as you remarked.

PR #891, opened 2017-07-17
Check universes are declared

mattam82:
This is on top of PR #890  [merged]
This is a tentative to ensure during typechecking in the kernel that universes in the terms are properly declared, even though they don't necessarily appear in constraints. Currently the kernel accepts term with undeclared universes and fails at their first use. I fear this is going to be costly (esp. on universe polymorphic code) and we should maybe do this more globally as a well-formedness test before typing, collecting all the universes of the term and checking their well-formedness at that time (maybe also checking that there is no algebraic universe in the wrong place at that time).

PR #7901, opened 2018-06-22
Fix #7900 (dropped universes in obligation side effects)

### LINKED ISSUE #7900: Undefined universe with Program, polymorphism and side effects
~~~coq
Require Import Coq.Program.Program.
Set Universe Polymorphism.
Set Printing Universes.

Axiom ALL : forall {T:Prop}, T.

Inductive Expr : Set := E (a : Expr).

Parameter Value : Set.

Fixpoint eval (e: Expr): Value :=
  match e with
  | E a => eval a
  end.

Class Quote (n: Value) : Set :=
  { quote: Expr
    ; eval_quote: eval quote = n }.

Program Definition quote_mult n
        `{!Quote n} : Quote n :=
  {| quote := E (quote (n:=n)) |}.


Next Obligation.
Proof.
  destruct Quote0 as [q eq].
  subst.
  apply ALL.
Qed.
Print quote_mult_obligation_1.
~~~
See #891 for context.

SkySkimmer:
That bug is hard to make a test case for without #891.

I don't understand side effects and program enough to be sure that this is the right fix, but it seems reasonable at least.

### DIFF
--- a/engine/uState.ml
+++ b/engine/uState.ml
@@ -447,10 +447,7 @@ let merge ?loc sideff rigid uctx ctx' =
           uctx_univ_algebraic = LSet.union uctx.uctx_univ_algebraic levels }
         else { uctx with uctx_univ_variables = uvars' }
   in
-  let uctx_local =
-    if sideff then uctx.uctx_local
-    else ContextSet.append ctx' uctx.uctx_local
-  in
+  let uctx_local = ContextSet.append ctx' uctx.uctx_local in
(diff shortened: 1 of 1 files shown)


SkySkimmer:
Looks like some stuff breaks.

ppedrot:
You don't know where you have stepped. This code is nightmare.

SkySkimmer:
＊runs away＊

PR #7902, opened 2018-06-22
Use a homebrew parser to replace the GEXTEND extension points of Camlp5

ppedrot:
In an effort to reduce dependency on Camlp5, the use of Camlp5 GEXTEND macro has been replaced by a homebrew parser relying on ocamlyacc.

The parser is stupid and the syntax is almost the same as the previous one. The only difference is that one needs to wrap OCaml code between { braces } so that quoting works out of the box. Files requiring such a syntax are handled specifically by the type system and need to have a .mlg extension instead of a .ml4 one. Currently mlg files can only contain GRAMMAR EXTEND statements. They do not
support TACTIC, VERNAC nor ARGUMENT EXTEND.

All ml4 files containing only GEXTEND statements were ported in a semi-automatic way, i.e. using crappy regular expressions that turned out to work surprisingly well and resorting to manual modification when needed. Files containing a mix of both that were not ported yet are:
- plugins/funind/g_indfun.ml4
- plugins/ltac/g_ltac.ml4
- plugins/ltac/g_obligations.ml4
- plugins/ltac/g_rewrite.ml4
- plugins/ssr/ssrparser.ml4
- plugins/ssr/ssrvernac.ml4
- plugins/ssrmatching/ssrmatching.ml4

They should be split into a GEXTEND part and another extension part. Handling Coq EXTEND macros is delayed to another PR, where most of them could be replaced by pure OCaml code.

The feature has been documented in `dev/doc/changes.md`.

I would be happy if this PR could be quickly discussed and (hopefully) merged, as git is sometimes stupid when it comes to rebasing, as it doesn't realize that two files are the same. This leads to having to redo a lot of manual work.

- [x] Entry added in CHANGES.


maximedenes:
> In an effort to reduce dependency on Camlp5

Is this still part of a plan to embed the parsing engine of Camlp5? Or is the plan that the engine is released separately so we can depend on only that?

JasonGross:
> git is sometimes stupid when it comes to rebasing, as it doesn't realize that two files are the same

I believe you can kludge around this by having one commit that renames the files without editing, and another commit that edits the newly named files.  (This may, of course, break the invariant that all commits should compile, but you might be able to kludge that with symlinks...)

ppedrot:
This is orthogonal to the decision of embedding or not. This PR partially removes the dependency of Coq on the "preprocessing" part of Camlp5, and leaves the grammar engine dependency untouched. Whether we embed the engine or we release the grammar engine separately, we will just have to adapt the wrapper in `pcoq.ml`. The homebrew preprocessor does not rely on Camlp5 API, it uses the safe API defined in Pcoq.

Zimmi48:
One especially good argument I've heard against the use of a custom parser is the lack of tooling support. `.ml4` files were already pretty badly supported by OCaml tools but it's going to be worse with `.mlg`. It might be time to consider ppx (which is well supported by Merlin) instead (even if it does way too much).

ppedrot:
@Zimmi48 I tried that, and the moral is that it is completely unreadable. Note that ocamlyacc files suffer from the same shortcoming.

ppedrot:
Ltac2 needs an overlay, I'll take care of it.

gares:
You probably need to fix Coq Makefile too.

May I ask why the extra `{}` ? They are everywhere and the expressions are already delimited by `->` and `]`so I quite don't see why they would be needed.

ppedrot:
> You probably need to fix Coq Makefile too.

I might need help on that one.

> May I ask why the extra {} ? They are everywhere and the expressions are already delimited by -> and ]so I quite don't see why they would be needed.

Otherwise you need a full parser of OCaml expressions, as these tokens are not necessarily well-bracketed in OCaml. Using delimiters that are always well-bracketed allows for a stupid parser implementation were we just dump anything that is in between them (modulo a bit of semantic like merely counting the number of other delimiters).

gares:
Hum, for coq_makefile you have to:
- add rules to `CoqMakefile.in` for `.mlg` as you did in the other makefiles
- drop the ones for .ml4 (or maybe keep them, did't check if you dropped them)
- patch the CoqProject_file parser to accept mlg in place of ml4

Fine for `{}`, but if that is the only reason then moving from `[ ... -> ... ]` to `[ ... ]  -> [ ... ]` seems more consistent with what we do for `TACTIC EXTEND` and company.

herbelin (commented):
No particular comments. There are various extensions in sight about the extensible grammar engine and having the expertise on it seems a good way to go (and maybe, one day, making our extensions available for the OCaml community back).

ppedrot:
@gares I also needed a delimiter that was not part of the tokens used for the grammar parsing, otherwise it would have required a more complex lexing/parsing algorithm.

ejgallego:
I am not sure this is much of an improvement, we are just replacing a piece of non-standard code with another piece of non-standard code. Maintenance burden seems the same to me.

ppedrot:
@ejgallego You really shouldn't compare the maintenance cost of camlp5 vs. the one of a few hundred lines of ocamlyacc. The former needs a release for every OCaml release, breaks each time I try to use a slightly fancy OCaml compiler flag, and is accumulating tens of thousands of semi-automatically generated code in its repository. The latter is essentially stable *for ever*.

ejgallego:
You have a point but IMO that doesn't mean you couldn't fix that problem in Camlp5. Note that I am not against this PR, just commenting; I feel like as @Zimmi48 has pointed out, I would welcome a more standard solution such as PPX but on the other side indeed, we are not ready for it.

I am a bit short on time these days so in order not to block this I have only one key question for my review, what are the dependencies of the preprocessor tool?

ppedrot:
> I am a bit short on time these days so in order not to block this I have only one key question for my review, what are the dependencies of the preprocessor tool?

None, it is a self-contained ocamlyacc parser with a backend written in pure OCaml. In particular it can be compiled at the very beginning of the compilation phase of Coq, just as we do today for the camlp5 drivers in grammar.

ejgallego (approved):
LGTM.

ppedrot:
Are there many plugins using the `GEXTEND` construction? I'm a bit struggling with `coq_makefile` now, I do not know whether it is that urgent to port it yet.

Makefile.build
```diff
@@ -359,11 +364,15 @@ GRAMC := $(OCAMLFIND) ocamlc $(CAMLFLAGS) $(CAMLDEBUG) $(USERFLAGS) \
...
 	@touch grammar/test.mlp
-	$(HIDE)$(GRAMC) -pp '$(CAMLP5O) -I $(MYCAMLP5LIB) $^ -impl' -impl grammar/test.mlp -o grammar/test
+	$(HIDE)$(GRAMC) -pp '$(CAMLP5O) $^ -impl' -impl grammar/test.mlp -o grammar/test
 	@rm -f grammar/test.* grammar/test
 	$(SHOW)'OCAMLC -a $@'
 	$(HIDE)$(GRAMC) $^ -linkall -a -o $@
 
+grammar/gramext: $(GRAMEXTCMO) grammar/gramext_main.ml
+	$(SHOW)'OCAMLC -a $@'
+	$(HIDE)$(GRAMC) $^ -linkall -o $@
```
ejgallego: This binary will have to be installed, so I propose you use here the final name, which should have coq on it. In the Dune build we already build an optimized preprocessor, named `coqp5`, maybe you could reuse that name?

grammar/gramext_main.ml
```diff
@@ -0,0 +1,290 @@
...
+(*    //   *      This file is distributed under the terms of the       *)
+(*         *       GNU Lesser General Public License Version 2.1        *)
+(************************************************************************)
+
+open Lexing
+open Gramext_ast
+open Format
+
+let fatal msg =
+  let () = Printf.eprintf "Error: %s\n%!" msg in
```
ejgallego: Umm, don't open `Format` then use `Printf`. I suggest you use `Format` everywhere.

grammar/gramext_main.ml
```diff
@@ -0,0 +1,290 @@
...
+  Printf.sprintf "File \"%s\", line %d, characters %d-%d:" file line bpos epos
+
+let parse_file f =
+  let chan = open_in f in
+  let lexbuf = Lexing.from_channel chan in
+  let () = lexbuf.lex_curr_p <- { lexbuf.lex_curr_p with pos_fname = f } in
+  let ans =
+    try Gramext_parse.file Gramext_lex.token lexbuf
+    with
+    | Gramext_lex.Lex_error (loc, msg) ->
```
ejgallego: Please do proper cleanup in exceptions, such as closing file descriptors.

grammar/gramext_main.ml
```diff
@@ -0,0 +1,290 @@
...
+      let () = Printf.eprintf "%s\n%!" (pr_loc loc) in
+      fatal "syntax error"
+  in
+  let () = close_in chan in
+  ans
+
+module StringSet = Set.Make(String)
+
+let string_split s =
+  let len = String.length s in
```
ejgallego: Yet another string split function. We may have like 20 in Coq.

grammar/gramext_main.ml
```diff
@@ -0,0 +1,290 @@
...
+  let () = fprintf chan "let _ = @[" in
+  let () = fprintf chan "@[<v>%a@]" print_local ext in
+  let () = List.iter (fun e -> print_entry chan ext.gramext_name e) ext.gramext_entries in
+  let () = fprintf chan "()@]\n" in
+  ()
+
+end
+
+let pr_ast chan = function
+| Code s -> fprintf chan "%s\n%!" s.code
```
ejgallego: In almost all cases you want to use "@\n", not "\n".

ejgallego: Also, adding the flush here is fragile, ensure flush and mandatory boxing in the top level call instead.

parsing/extend.ml
```diff
@@ -101,7 +101,7 @@ type ('self, 'a) symbol =
 | Aself : ('self, 'self) symbol
 | Anext : ('self, 'self) symbol
 | Aentry : 'a entry -> ('self, 'a) symbol
-| Aentryl : 'a entry * int -> ('self, 'a) symbol
+| Aentryl : 'a entry * string -> ('self, 'a) symbol
```
ejgallego: What's the reason for this?

grammar/gramext_main.ml
ppedrot: Blame the OCaml stdlib that only introduced that function in 4.04.

Makefile.build
ppedrot: Technically no plugin rely on it yet, but it is probably going to be the case indeed. I don't have anything against calling it `camlp5`. Do other people agree?

maximedenes: I prefer `coqp4`. This is not even a troll :)

maximedenes: What does `coqp5` mean?

SkySkimmer: coqppx?
camlpcoq?

ppedrot: > What does coqp5 mean?

An ancient african word that means *I do not want Coq to depend on camlp5*?

ppedrot: If anything, `coqpp` would be closer to reality.

maximedenes: I like `coqpp`.

ppedrot:
Let's go for `coqpp` then.

ppedrot:
Nitpicks addressed and binary renamed to `coqpp`.

ppedrot:
FTR, I am leaving the porting of `coq_makefile` and the system-wide installation of `coqpp` for another PR, as they should be part of the same bundle. For now external plugins cannot use the homebrew preprocessor.

grammar/gramext_main.ml
ejgallego: Add a note/fixme so we can clean when we up the ocaml version

ppedrot:
To whomever wants to be the assignee, I consider this PR to be ready.

ejgallego:
> To whomever wants to be the assignee, I consider this PR to be ready.

Who should be indeed?

parsing/extend.ml
ejgallego: OK, it just seems weird to see all the new `int_of_string`, also these are no guarded so we can get an anomaly. So I guess I am missing something.

ejgallego:
I am OK with this, but will wait 48h before merge [at least] as this is a large change. I will also do a more in-depth review but I don't expect to find any significant problem.

People in reviews, @herbelin , @gares, please do share your thoughts.

ejgallego:
@ppedrot this is failing on Windows :S

ppedrot:
@ejgallego I don't think it is related, it used to have a green CI, so it's probably transient. I am relaunching it to check.

dev/ci/user-overlays/07902-ppedrot-camlp5-parser.sh
```diff
@@ -0,0 +1,8 @@
+_OVERLAY_BRANCH=camlp5-parser
+
+if [ "$CI_PULL_REQUEST" = "7902" ] || [ "$CI_BRANCH" = "$_OVERLAY_BRANCH" ]; then
+
+    ltac2_CI_BRANCH="$_OVERLAY_BRANCH"
+    ltac2_CI_GITURL=https://github.com/ppedrot/ltac2.git
```
Zimmi48: Remove the `.git` to have the Windows build succeed. See #7964

Zimmi48:
It wasn't spurious and it is fixed now (cf. my last, outdated, inline comment).

ejgallego:
> It wasn't spurious and it is fixed now (cf. my last, outdated, inline comment).

Thanks @Zimmi48 , indeed that was my impression but could remember.

ejgallego:
We could do a small command line tool to search for all builds of a given PR in the CI system, which is cumbersome to do now.

Zimmi48:
Actually GitLab CI shows the list of previous failed jobs: see the right panel here https://gitlab.com/coq/coq/-/jobs/78462738

ejgallego:
Merging tomorrow afternoon if no one comments more, I think 10 days is a resonable timeframe.

maximedenes:
Where are `coqpp` tests?!

ppedrot:
What do you want to test? We already have a bunch of `mlg` files in the repo now. The only similar test for camlp5 is `test-suite/coq-makefile/template/src/test.ml4` which does not test camlp5 per se but rather coq_makefile. As we do not support coqpp in coq_makefile yet this makes no sense to have the equivalent test for coqpp.

PR #7912, opened 2018-06-24
Simplify effects API

ppedrot:
This PR simplifies the kernel API for definition side-effects.

It does so by making the kernel mostly parametric in the effect roles, which are only threaded through it and never manipulated for their contents. This allows to get rid of kernel-facing code that was duplicating its own variant of side-effect role for nothing, and results in a code that just manipulate lists abstractly.

Furthermore, redundant API has been compacted so that that there are less exported kernel functions.

### DIFF
--- a/kernel/safe_typing.mli
+++ b/kernel/safe_typing.mli
@@ -44,7 +44,7 @@ type 'a safe_transformer = safe_environment -> 'a * safe_environment
 type private_constants
 
 val side_effects_of_private_constants :
-  private_constants -> Entries.side_effect list
+  private_constants -> Entries.side_eff list
 (** Return the list of individual side-effects in the order of their
     creation. *)
 
--- a/kernel/safe_typing.mli
+++ b/kernel/safe_typing.mli
@@ -41,12 +41,7 @@ type 'a safe_transformer = safe_environment -> 'a * safe_environment
 
 (** {6 Stm machinery } *)
 
-type private_constant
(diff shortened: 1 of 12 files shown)


kernel/safe_typing.mli
```diff
@@ -41,29 +41,20 @@ type 'a safe_transformer = safe_environment -> 'a * safe_environment
...
-type private_constant
 type private_constants
 
-type private_constant_role =
-  | Subproof
-  | Schema of inductive * string
-
 val side_effects_of_private_constants :
-  private_constants -> Entries.side_effect list
+  private_constants -> Entries.side_eff list
```
SkySkimmer: Maybe this name change could be made unnecessary.

ppedrot: The `side_eff` type corresponds to the previous one with the same name, while `side_effect` also contains the trusted environment. So, it's more that the `side_effects_of_private_constants` has changed type rather than the type has changed name.

maximedenes (approved):
Looks superfically good to me, but I'm not competent, so I'd like to see a review by @gares. Do we test these features in the test-suite btw?

ppedrot:
We do test them, and probably more than in the wild. There are a few tricky tests mixing universes and side-effects for instance.

gares (approved):

gares:
@maximedenes LGTM

PR #7917, opened 2018-06-25
Critical bugs: added #3243 and Gonthier's bug in lazy machine.

herbelin:
Both reminded by Enrico.

**Kind:** documentation

This is the continuation of #7805 (there was a synchronisation problem with my branches).

### DIFF
--- a/dev/doc/critical-bugs
+++ b/dev/doc/critical-bugs
@@ -81,6 +81,20 @@ Module system
   GH issue number: #4294
   risk: ?
 
+Module system
+
+  component: modules, universes
+  summary: universe constraints for module subtyping not stored in vo files
+  introduced: presumably 8.2 (b3d3b56)
+  impacted released versions: 8.2, 8.3, 8.4
+  impacted development branches: v8.5
+  impacted coqchk versions: none
+  fixed in: v8.2 (c1d9889), v8.3 (8056d02), v8.4 (a07deb4), trunk (0cd0a3e) Mar 5, 2014, Tassi
+  found by: Tassi by running coqchk on the mathematical components library
+  exploit: requires multiple files, no test provided
+  GH issue number: #3243
+  risk: could be exploited by mistake
+
 Universes
 
   component: template polymorphism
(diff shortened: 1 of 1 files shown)


gares:
thanks

SkySkimmer:
Was the lazy machine bug really a soundness bug? Since the fix is to reduce more it seems like an incompleteness bug.
Also I still haven't seen an example where the code before the fix went wrong.

maximedenes:
> Was the lazy machine bug really a soundness bug? Since the fix is to reduce more it seems like an incompleteness bug.

It was a soundness bug, because the machine was relying else on invariants that were invalidated. I guess @barras must have Georges' example in his archives. Or maybe Georges still has it.

Zimmi48 (approved):
Should I wait to see whether the exploit can be retrieved before merging this?

And BTW, should we give code ownership of this file to someone else?

gares:
There was a match on the (supposedly) whnf of the terms, and a catch all `_, _ -> false`.
The side effect was breaking the invariant that a term, once put in whnf, stays in whnf.
The cache for constants was shared among LHS and RHS, and while `id x` is in whnf, `(fun y => y) x` is not (you need a more complex term, IIRC, but the idea was this one).

gares:
I think having it in the test-suite would be of little value (today) since the cache was recently made into two (one for the LHS and one for the RHS) IIRC.

gares:
> There was a match on the (supposedly) whnf of the terms, and a catch all _, _ -> false.

Sorry, the bug was that, in the lambda case, stacks were ignored (since they had to be empty). So it was definitely a soundness bug.

herbelin:
For the record, I have a copy of the corresponding proof of `False` in my mailbox but it would indeed be of little interest as a regression test for the reason mentioned by @gares. In my opinion, the interest is more in the achievement itself which is quite sophisticated.

maximedenes:
I would be curious to see the term :)

gares:
yes, please post it ;-)

herbelin:
I sent it to @maximedenes. You can ask him.

I suspect that both @ggonthier and @barras would be ok to make it public (even if it is a pretty abstruse proof). Originally, it was only sent privately to some developers, so I'm following the initial intention of the author.

maximedenes:
I got explicit approval from Georges to publish the proof (Georges simply wanted to do a "responsible disclosure" of the bug at the time), so here it is:

```coq
CoInductive bar : Type := Bar (_ : unit -> bool).
Definition foo T b : Type := let (b) := b : bar in if b tt then T else unit.
Definition magic_obj T : T :=
  let t _ := true in let bart := Bar (let b := t in b) in
  let barf := let (b) := bart in Bar (if b tt then fun _ => false else t) in
  let fubar : foo T barf := tt in fubar : foo T (if t tt then bart else barf).
```

PR #7918, opened 2018-06-25
Mini-update of version history with recent changes.

herbelin:
**Kind:** documentation

I find convenient to have this kind of information at hand (e.g. I needed it to get the global picture for critical bugs).

### DIFF
--- a/dev/doc/versions-history.tex
+++ b/dev/doc/versions-history.tex
@@ -395,7 +395,17 @@ Coq V8.7 beta 1 & released 6 September 2017 & \feature{bundled with Ssreflect pl
 
 Coq V8.7 beta 2 & released 6 October 2017 & \\
 
-Coq V8.7 & released 18 October 2016 & \\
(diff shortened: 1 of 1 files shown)


herbelin:
I don't think that the credits chapter has all the information that the version-history file has. And if it has, it certainly does not have it in the same synthetic way.

I'm anyway ok to continue to maintain the file as long as it is useful for me.

Zimmi48 (approved):
OK then.

Do you want to keep it a `.tex` file though?

herbelin:
Is there a way to render the tables and tabulations nicely in .md? If yes, we can move it to markdown.

Zimmi48:
Tables, yes: https://help.github.com/articles/organizing-information-with-tables/

Zimmi48:
@herbelin Shoud I merge this now and leave the conversion for later?

herbelin:
It'd be preferable, yes, if ok for you.

PR #7919, opened 2018-06-25
Fix equality check on global names from native compute.

ppedrot:
Not sure it could have led to a soundness bug, but this is definitely serious enough to deserve a backport. Also made the code robust by listing all the constructors.

I am assigning @maximedenes by default, this is simple enough to be reviewed and merged switfly.


maximedenes (approved):
Thanks!

PR #6247, opened 2017-11-25
Support for custom entries in notations

herbelin:
This is the part of #982 dedicated to supporting custom entries.

There is still some work to do, starting with documentation, continuing with a few open questions. I'm opening a new PR so that #982 which seems to be now relatively stable can be considered independently.

Typical (new) open questions about custom entries:

Do we automatically add a special notation `Notation "( x )" := x.` in each new custom entry, or do we ask the user to do it? This would be convenient in a custom entry with levels and precedences since parentheses are the natural ingredient to go back from the lower to the higher precedence.

Do we automatically add a rule to parse identifiers in each new custom entry, or do we let the user  do it on demand? Again, this would be convenient in a custom entry used to represent expressions of a given language.

In any case, if we let the user decide, we would need to provide a specific user API for that since this goes beyond what the `Notation` mechanism is able to do at the moment.

PS: for convenience, this PR currently includes #978 and #982.

PR #7923, opened 2018-06-26
Documentation for custom entries

herbelin:
**Kind:** documentation / fine-tuning

~~This collects the part of #6247 which needs some more work (documentation + decide how involved the example of the documentation can be).~~ ~~This extends #6247 with documentation, fixes and improved support for "grammar coercions".~~ This extends #8115 with documentation (the last two commits).

- [X] Documentation in reference manual
- [X] Entry to be added in CHANGES
- [X] Fine-tuning to make the example of the documentation working (parsing of parentheses, of identifiers, of rules without a symbol)
- [ ] Feedback on documentation needed

Note: This is now on top of ~~#6247~~#8115 (which provides the feature part of custom entries and which is the new version of #6247).


stm/vernac_classifier.ml
```diff
@@ -176,7 +176,7 @@ let classify_vernac e =
     (* These commands alter the parser *)
     | VernacOpenCloseScope _ | VernacDelimiters _ | VernacBindScope _
     | VernacInfix _ | VernacNotation _ | VernacNotationAddFormat _
-    | VernacSyntaxExtension _
+    | VernacSyntaxExtension _ | VernacDeclareCustomEntry _
```
gares: Does declaring an entry change the parser in a visible way?
Adding notation to that entry do, of course. But just declaring the entry seems not to require `VtNow`.

herbelin: OK, I moved to `VtLater`. Note that this is relevant for #6247, so I'm changing it there.

herbelin: I made the change but now I have a doubt. Where should I indicate that a `Notation` referring to a custom entry `foo` should be evaluated only after the corresponding `Declare Custom Entry foo` has been evaluated.

gares: I guess the most important thing is that the error message is clear, eg "No custom entry named 'foo'".

ejgallego: If the command doesn't modify the actual grammar parsed `VtLater` tends to be safe I think.

gares: Given that `Notation` is `VtNow` the STM (of today) will evaluate all the commands coming before `Notation` in order to execute `Notation`, hence `Notation` is executed in a state where the entry is declared (assuming there is a command `Declare Custom..` before the `Notation` populating the new entry).

In the model we whish to obtain one day, both `Notation` and `Declare Custom Entry` will be commands that have their execution broken into two pieces: changing the parser + the rest.
In the case of `Declare Custom Entry` the "rest" is likely to be empty, while for `Notation` the rest is going to be about checking the interpretation and add the notation to other data structures like the ones used for printing. Both change some data structures relative to parsing, even if only `Notation` in a visible way.

So in a sense `Declare Custom Entry` is closer to `Notation`, but as of today putting `VtNow` (whilst being safe) is not needed. You can leave it if you prefer, maybe with a comment "not strictly needed since just adding a non terminal does not change the language accepted by the parser".

herbelin: Very clear explanation, thanks. I did not realize that the current implementation forces evaluation of all commands preceding a `Notation`.

herbelin:
Even if there are still possible improvements to do in style (*), the documentation seems now ok in terms of contents. Feedback is welcome.

The design of the custom entry feature seems to stabilize with a realistic example shown in the manual. The "ident" and "parentheses" features discussed in #6247 are implemented by means of "grammar coercions". The "field" feature (@andres-erbsen [here](https://github.com/coq/coq/pull/6247#issuecomment-395241913) and [here](https://github.com/coq/coq/pull/6247#issuecomment-401097675)), maybe generally the availability of any `g_prim` entry, is not done.

(*) e.g. non systematic use of `cmd::`, alternance of a "pedagogical" style with a "factual" style, terminology sometimes informal (can we more clearly define "term", "expression", "entry", "grammar", "rule", ...), not very smooth English, ...

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -687,6 +687,31 @@ side. E.g.:
    Notation "'apply_id' f a1 .. an" := (.. (f a1) .. an)
      (at level 10, f ident, a1, an at level 9).
 
+Custom entries
+~~~~~~~~~~~~~~
+
+New grammar entries can be defined using the ``Declare Custom Entry``
```
Zimmi48: Indeed, I haven't followed these PRs really closely but if the `Declare Custom Entry` command is a new one, it is really important that you present it in the following way:

```
.. cmd:: Declare Custom Entry @ident

   This command allows to define new grammar entries that can later be referred to 
   using the entry name :n:`custom @ident`.

   .. example::

      For instance, we may want to define an *ad hoc* parser for arithmetical operations
      and proceed as follows:

      .. coqtop::
```

In particular, the use of `cmd` allows to find this command in the command index.

If you haven't yet, I suggest you read https://github.com/coq/coq/tree/master/doc/sphinx#documenting-coq-with-sphinx
I find this documentation very helpful to better understand how the new format of the manual works.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -705,6 +730,8 @@ following figure. The optional :token:`scope` is described in the Section 12.2.
                  : | CoFixpoint `cofix_body` [`decl_notation`] with … with `cofix_body` [`decl_notation`].
    decl_notation : [where `string` := `term` [: `scope`] and … and `string` := `term` [: `scope`]].
    modifiers     : at level `natural`
+                 : in custom `ident`
+                 : in custom `ident` at level `natural`
```
Zimmi48: I have tried to remove most uses of `natural` and replace them with `num` but apparently failed to do so in this grammar.
Note that using `num` would put a link to https://coq.inria.fr/refman/language/gallina-specification-language.html#grammar-token-num while using `natural` would put a link to https://coq.inria.fr/refman/proof-engine/ssreflect-proof-language.html#grammar-token-natural (`natural` is defined as `num` or `ident` in the SSR chapter).

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: Hum, I'm very sorry, I had again pushed to a wrong branch name and you did not see the correct version.

Anyway, the correct version had a `.. cmd:: Local Declare Custom Entry @ident` but no `.. cmd:: Declare Custom Entry @ident`, so I added one.

I'm unsure to  have fully understand the difference between single and double backticks. Is it that we use double backticks when in regular text and single ones when argument of a sphinx role/macro? Also, I guess that I should write ``:n:`custom @ident` `` rather than ``` ``custom`` `ident` ``` to mean that I'm thinking at generic code where `ident` is a metavariable, and that I should write ``:g:`custom foo` `` to mean that I'm thinking at a specific example with `foo` taking the place of the metavariable?

My use of ``:g:`...` `` vs ``...`` seems still a bit random.

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: OK, I made the change.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -687,13 +687,126 @@ side. E.g.:
    Notation "'apply_id' f a1 .. an" := (.. (f a1) .. an)
      (at level 10, f ident, a1, an at level 9).
 
+Custom entries
+~~~~~~~~~~~~~~
+
+.. cmd:: Declare Custom Entry @ident
+
+New term and pattern grammar entries, called *custom entries*, can be
```
Zimmi48: The content of the block should be indented at the level of `cmd` to be understood as the body of the command definition. See how I did it in my comment on your previous version.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -687,13 +687,126 @@ side. E.g.:
...
+declared using the :n:`Declare Custom Entry @ident` command and later
```
Zimmi48: I suggest you reformulate as in my previous proposition to avoid repeating this command uselessly.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -687,13 +687,126 @@ side. E.g.:
...
+   | [x y + z] => [x + y z]
+   | y => [y + e]
+   end.
+
+Custom entries have levels, like the main grammar of terms and grammar
+of patterns have. The lower level is 0 and this is the level used by
+default to put rules delimited with tokens on both ends. The level is
+left to be inferred by Coq when using :n:`in custom @ident``. The
+level is otherwise given explicitly by using the syntax :n:`in custom
+@ident at level @num`, where :n:`@num` refers to the level.
```
Zimmi48: You can use ```:token:`num` ``` instead of ```:n:`@num` ```. See https://github.com/coq/coq/tree/master/doc/sphinx#coq-roles

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -687,13 +687,126 @@ side. E.g.:
...
+
+Levels are cumulative: a notation at level ``n`` of which the left end
```
Zimmi48: I would use ```:token:`num` ``` instead of ``n`` in this paragraph. But that's a matter of taste.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -687,13 +687,126 @@ side. E.g.:
...
+x y) z` using the syntax ``(x + y) z``. Similarly, :g:`Notation "{ x }"
+:= x (in custom expr at level 0, x constr)` gives a way to let any
+arbitrary expression which is not in handled by the custom entry
+``expr`` be parsed or printed by the main grammar of term up to the
+insertion of a pair of curly brackets.
+
+.. cmd:: Print Grammar @ident.
+
+   This displays the state of the grammar for terms and grammar for
+   patterns associated to the custom entry ``ident``.
```
Zimmi48: ```:token:`ident` ```

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -687,13 +687,126 @@ side. E.g.:
...
 Summary
 ~~~~~~~
 
 **Syntax of notations**
 
-The different syntactic variants of the command Notation are given on the
-following figure. The optional :token:`scope` is described in the Section 12.2.
+The different syntactic forms taken by the commands declaring
+notations are given below. The optional :token:`scope` is described in
+Section 12.2.
```
Zimmi48: Watch out that your change to this two lines is going to conflict with the improvements at #7969.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -745,13 +869,18 @@ following figure. The optional :token:`scope` is described in the Section 12.2.
...
 
 .. cmd:: Local Notation @notation
 
    Notations survive modules unless the command ``Local Notation`` is used instead
    of ``Notation``.
 
+.. cmd:: Local Declare Custom Entry @ident
+
+   Custom entries survive modules unless the command ``Local Declare
+   Custom Entry`` is used instead of ``Declare Custom Entry``.
```
Zimmi48: Use ```:cmd:`Declare Custom Entry` ``` here to get a hyperlink.

Zimmi48 (commented):
> I'm unsure to have fully understand the difference between single and double backticks. Is it that we use double backticks when in regular text and single ones when argument of a sphinx role/macro? Also, I guess that I should write ```:n:`custom @ident` ``` rather than ``` ``custom`` `ident` ``` to mean that I'm thinking at generic code where ident is a metavariable, and that I should write ```:g:`custom foo` ``` to mean that I'm thinking at a specific example with foo taking the place of the metavariable?

Yes, single backticks always follow a `:foo:` and are never supposed to be alone (except in production lists).
`@ident` inside a `:n:` block or ```:token:`ident` ``` (or ``` `ident` ``` inside the grammar) means that this is something generic and will put a link to the grammar entry definition. On the contrary `as ident` would mean that the user needs to type `ident` literally instead of replacing it with a value. This distinction seems of particular importance in this chapter.

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: I knew it but forgot. Fixed.

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: Done.

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: I tried both version but couldn't understand the difference. They both seems to link to where the entry is defined in the manual, and the only difference seems to be a color blue vs a color green.

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: If I do that, I should do it for all remaining occurrences in the paragraph, and it would look a bit strange to explain a story using a "placeholder" name rather than a concrete name, wouldn't it?

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: Done.

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: The two changes seems compatible though.

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: Done.

doc/sphinx/user-extensions/syntax-extensions.rst
Zimmi48: Yes.

doc/sphinx/user-extensions/syntax-extensions.rst
Zimmi48: Yes, I'm not exactly sure either.

doc/sphinx/user-extensions/syntax-extensions.rst
Zimmi48: Yes, you would need to change it everywhere. I don't think it would look strange but as I said, it is a matter of taste.

maximedenes (dismissed):
Willl review the doc once the feature is merged.

Zimmi48:
The title of this PR is highly misleading as it is not at all about just documenting the feature, but also squashing together some commits from the previous PR. I think the plan is to close the depending PR and request this one to be reviewed. But anyways, there is a rebase to do before it can be properly reviewed. The title of the PR should also be changed.

Zimmi48:
Seems like I can't edit review dismissal comments, but you meant #8115 instead.

Zimmi48 (approved):
I understand that the documentation for custom entries that was supposed to be handled in this PR has been accidentally merged in the feature PR #8115.

Well, no problem at all because this documentation is great as-is. I just had very few trivial fixes to make, so I opened #8189.

Zimmi48:
So I suggest to just close the present PR and move on.

ejgallego:
Indeed I did read the documentation in the merged PR and it looked good to me so I guessed Hugo meant to tackle the docs in the original PR, which IMO makes more sense (merging a feature without documentation is so so)

So indeed as this PR is emtpy as-is, and given Théo assessment, let's close this PR and open a doc review issue if Hugo feels like it.

herbelin:
OK, closing. And anyway, there is #8189 available for a second pass.

PR #7924, opened 2018-06-26
Ad hoc fix for #5696, #7903 (ltac subterms and open subterms in notations).

### LINKED ISSUE #5696: Tactics in terms break unrelated .. binders in notations

Original bug ID: BZ#5696
From: @JasonGross
Reported version: 8.8+alpha
CC:   @herbelin, @jadephilipoom

### LINKED ISSUE #7903: Tactics in terms in notations break binders / open term
#### Version

8.8.0

```coq
Notation foo x f := (fun x : nat => f).
Notation bar x f := (let z := ltac:(exact 1) in (fun x : nat => f)).
Check foo x (x + x). (* success *)
Check bar x (x + x). (* Error: The reference x was not found in the current environment. *)
```

See also #5696, #3304, #3468

cc @herbelin , @ppedrot , @andres-erbsen

herbelin:
We fix the issue by removing terms of the substitutions which have free variables and are thus not interpretable in the context of the `ltac:` subterm.

This remains rather ad hoc, since, in an expression `Notation f x := ltac:(foo)`, we interpret `x` at calling time of `foo` rather than at use time of `x` in `foo`, even if `x` is not necessary.

Independently, we could also imagine that binders in the ltac expressions are then interpreted but that would start to be (very) complicated.

Note that this will presumably "fix" ltac2 quotations at the same time.

**Kind:** bug fix

Closes #5696. Closes #7903.

### DIFF
--- a/interp/constrintern.ml
+++ b/interp/constrintern.ml
@@ -710,10 +710,12 @@ let instantiate_notation_constr loc intern intern_pat ntnvars subst infos c =
       let arg = match arg with
       | None -> None
       | Some arg ->
-        let mk_env (c, (tmp_scope, subscopes)) =
+        let mk_env id (c, (tmp_scope, subscopes)) map =
           let nenv = {env with tmp_scope; scopes = subscopes @ env.scopes} in
-          let gc = intern nenv c in
-          (gc, Some c)
+          try
+            let gc = intern nenv c in
+            Id.Map.add id (gc, Some c) map
+          with GlobalizationError _ -> map
         in
         let mk_env' (c, (onlyident,(tmp_scope,subscopes))) =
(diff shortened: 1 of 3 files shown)


ejgallego:
@herbelin be sure to add a review request for PRs where I am the secondary maintainer, otherwise I may miss them.

ejgallego:
Silly linter problem making the CI fail.

ppedrot (approved):

ejgallego (approved):

herbelin:
Linter fixed.

ejgallego:
Equations failure is spurious, it didn't took long for the code to go stale. I think asking for a rebase is overkill.

PR #7925, opened 2018-06-26
Clean transparent state

ppedrot:
We move `Names.transparent_state` to its own module `TranspState`. We furthermore turn it into a named record and provide accessors functions. This was documented in `dev/changes.md`. Due to the architecture of the `Names` module I could not add deprecation around the old API (this would have created a dependency cycle), but they are probably rarely used by plugins.


gares:
Hum, two things:
- between `TranspState` and `TransparentState` I prefer the latter, given how little the former saves you and how obvious the latter makes the transition (in habit)
- if we have to rename this, I think we could also take the occasion to pick a name that does not mean "the state is transparent" but rather "constants to be (not) unfolded"

Bottom line: you don't need to clean the state, it is already transparent...

mattam82:
Agreed with @gares here, it should already have been transparency_state or unfolding_state, and equations will likely need an overlay.

ppedrot:
So should I call this module `TransparencyState`?

gares:
I would propose `UnfoldBlacklist` but looking at its content I'm not sure I totally understand which piece of info it holds. I suspect is maps constants names to `YesUnfoldMe`, `NoDon'tUnfoldMe`, `DoAsYouWhish`...
`UnfoldingPreference`?

ppedrot:
It just holds a boolean for every constant / variable, contrarily to `conv_oracle` which contains a full strategy.

ppedrot:
So, any contender for the name?

maximedenes (approved):
No strong opinion. Change to record looks good.

maximedenes:
This needs a rebase, it seems.

herbelin (commented):
OK to me.

Zimmi48 (approved):
This makes things much clearer, thanks!
You could have made the type opaque and added setters to change the opacity of one variable / constant but I suppose you didn't because there was no point in doing this...
Why are some o(caml)doc comments above what they document and some below?

ppedrot:
This still needs an overlay though, I'll try to find time to write this down.

SkySkimmer:
>This still needs an overlay though, I'll try to find time to write this down.

How's it going?

ppedrot:
Almost done, it's just a mess to compile dependencies that are implicit in the CI. I'll update this shortly.

ejgallego:
What do you mean @ppedrot?

Zimmi48:
Submodules, I suppose.

ppedrot:
For some reason the overlays are not working properly, but the respective downstream changes do compile manually. Anybody understanding this better than me?

Zimmi48:
The error in Windows is because the overlay for equation was not correctly picked up. Looking at the log, `CI_PULL_REQUEST` and `CI_BRANCH` were given empty values:

```
+13:37:08 '[' -n '' ']'
+13:37:08 export CI_BRANCH=
+13:37:08 CI_BRANCH=
+13:37:08 export CI_PULL_REQUEST=
+13:37:08 CI_PULL_REQUEST=
+13:37:08 export TRAVIS_BRANCH=
+13:37:08 TRAVIS_BRANCH=
+13:37:08 export TRAVIS_PULL_REQUEST=
+13:37:08 TRAVIS_PULL_REQUEST=
```

@SkySkimmer @ejgallego @MSoegtropIMC I wondered what changed and could have broken this (it used to work, right?)

ppedrot:
As far as I am concerned, the PR is ready. CI failure on windows is spurious and reflects an unrelated bug of the latter.

MSoegtropIMC:
@ppedrot : ok @Zimmi48 found the issue and fixed it in #8918. So I agree that the Windows64 failure was due to a bug in my CI code.

SkySkimmer:
>@ppedrot : ok @Zimmi48 found the issue and fixed it in #8918. So I agree that the Windows64 failure was due to a bug in my CI code.

That is not a @Zimmi48 PR ;)

maximedenes (changes requested):
Sorry, but @gares' comment was ignored: can we get `TransparentState` instead of `TranspState`? I can do the change if you like.

maximedenes:
Also, is the Dune failure real?

ppedrot:
@maximedenes I quite don't care about the name. If you have time to perform the diff and adapt the overlays I'd be glad but otherwise I'll do it.

ppedrot:
This is ready. The windows64 failure is spurious.

maximedenes (approved):

PR #7928, opened 2018-06-26
Fix 'unbound variable' issue on Windows packaging jobs.

Zimmi48:
Since #7831 was merged, the Windows packaging jobs are failing with:
```
/build/user-overlays/00669-maximedenes-ssr-merge.sh: line 3: CI_PULL_REQUEST: unbound variable
ERROR MakeCoq_MinGW.bat failed
ERROR dev/ci/gitlab.bat failed
```
This was hard to anticipate given that it works well on pull requests.
The fix I'm proposing is to remove `set -o nounset` which is causing the failure. I'm not expecting this proposed fix to get approved given that it was put there for a reason, but it's hard to reason about our shell scripts if we don't use consistent programming styles in the whole code base (and indeed, we frequently assume that variables can be unset, that's why we put them in quotes to compare them to their expected value...).

### DIFF
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -1437,12 +1437,16 @@ function make_addon_equations {
 }
 
 function make_addons {
-  if [ -n "${GITLAB_CI}" ]; then
+  if [ -n "$GITLAB_CI" ]; then
     export CI_BRANCH="$CI_COMMIT_REF_NAME"
-    if [[ ${CI_BRANCH#pr-} =~ ^[0-9]*$ ]]
-    then
+    if [[ ${CI_BRANCH#pr-} =~ ^[0-9]*$ ]]; then
       export CI_PULL_REQUEST="${CI_BRANCH#pr-}"
+    else
+      export CI_PULL_REQUEST=""
     fi
+  else
+    export CI_BRANCH=""
+    export CI_PULL_REQUEST=""
   fi
   . /build/ci-basic-overlay.sh
   for overlay in /build/user-overlays/*.sh; do


MSoegtropIMC:
I don't agree to the proposed fix. Allowing unset variables will lead to maintainability issues before long. It is not unreasonable to expect that shell variables are defined before used. An alternative would be something around

    if [ -z ${CI_PULL_REQUEST+defined} ]
    then
        CI_PULL_REQUEST=
    fi

For an explanation search for ${parameter+word} in
http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html

(Please include the above link as a comment in case you use this method).

Note: "" are not required because the result is either null or defined, but our shell lint might require quotes.

Alternatively you can use ${CI_PULL_REQUEST+} wherever you use CI_PULL_REQUEST.

Zimmi48:
Thanks for the hints.

SkySkimmer:
Or we can just set the variable to empty string in the non-PR branch.

Zimmi48:
I went for the easiest solution.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -1439,10 +1439,14 @@ function make_addon_equations {
...
-    if [[ ${CI_BRANCH#pr-} =~ ^[0-9]*$ ]]
-    then
+    if [[ ${CI_BRANCH#pr-} =~ ^[0-9]*$ ]]; then
       export CI_PULL_REQUEST="${CI_BRANCH#pr-}"
+    else
+      export CI_PULL_REQUEST=""
     fi
+  else # assume local
+    export CI_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
+    export CI_PULL_REQUEST=""
```
MSoegtropIMC: You are aware that this script usually doesn't run inside a git repository? I don't understand what git is supposed to do here.

There are two typical ways how this script gets Coq:
1.) As a zip from github
2.) From a local folder (might be git or not)
In either case the coq code ends up in /build/coq-xyz, which never has any git stuff in it.
The makecoq_mingw.sh script usually has /build as working directory and enters the component subdirectories with build_prep.

Maybe you can explain what exactly you want to achieve in which build environment and then we discuss how to achieve this.

Please don't forget the initial purpose of these scripts: setup a clean environment and do everything in a clean room way.

Zimmi48:
> You are aware that this script usually doesn't run inside a git repository? I don't understand what git is supposed to do here.

Sorry, I had forgotten about this. I had just copied this code from https://github.com/coq/coq/blob/04e0f9fde8789a28b66f24000ac8c831ff0815af/dev/ci/ci-common.sh#L27-L30 and I had supposed it might be useful here as well (when testing locally within a git repo) but indeed, it doesn't really make any sense...

Zimmi48:
@MSoegtropIMC Can you review and merge?

MSoegtropIMC:
@Zimmi48 : To review this I need to understand a bit better. Can you tell me where the files

   . /build/ci-basic-overlay.sh
   for overlay in /build/user-overlays/*.sh; do

come from? Are they copied there by the outer CI test scripts? What happens if a user downloadeds the scripts to build Coq (which is the documented method). I guess these files won't exist then?

The /build folder is a special folder in the cygwin structure of the windows build. Since when are copied files into this folder and what are they used for otherwise? Or is the /build folder something else which exists on the CI machines? If so again it likely won't work because this script runs under cygwin and /build is something like C:\cygwin\build.

In short: these scripts are not primarily CI scripts and they should not be changed such that they work only in CI.

Zimmi48:
All these questions are interesting but are not relevant to the present PR. They should have been asked on #7831. This PR doesn't change anything to the system that was put in place in #7831 it only fixes a problem when these variables are undefined.
I'll answer these questions by commenting on the code of #7831. But this shouldn't delay the present PR.

MSoegtropIMC:
Thanks for the explanation. I will do the merge after lunch.

MSoegtropIMC (approved):
With the additional explanation the changes are understandable and fine.

Zimmi48:
Note that the Elpi failure is expected and unrelated.

PR #7932, opened 2018-06-26
CoqIDE scrolls the proof buffer down to the first goal.

### LINKED ISSUE #7931: request: focus on *goal* in coqide
8.8.

mac os

Every step in coqide moves the focus back to the beginning of the hypotheses list.
This can be very frustrating when working with long list of hypotheses, as it requires scrolling back down after every step. Would be much better if it'd focus on the first goal.

cmangin:
**Kind:** feature.
Closes #7931.

The proof buffer in CoqIDE now focuses on the first goal as requested (I actually found myself to be annoyed by this a few times).
The change is trivial (and I think the insertion of the tag was actually forgotten, as the focusing part was already there), but I have no opinion on the precise design decision.

By that I mean that this PR makes it so that the last line displayed in the proof buffer is the last line of the first goal. I could make it so that it is the next line to give a bit of breathing room for instance. I don't know what is best.

### DIFF
--- a/ide/wg_ProofView.ml
+++ b/ide/wg_ProofView.ml
@@ -103,7 +103,7 @@ let mode_tactic sel_cb (proof : #GText.view_skel) goals ~unfoc_goals hints = mat
           else []
         in
         proof#buffer#insert (goal_str ~shownum:true 1 goals_cnt);
-        insert_xml proof#buffer (Richpp.richpp_of_pp width cur_goal);
+        insert_xml ~tags:[Tags.Proof.goal] proof#buffer (Richpp.richpp_of_pp width cur_goal);
         proof#buffer#insert "\n"
       in
       (* Insert remaining goals (no hypotheses) *)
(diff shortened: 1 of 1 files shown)


gares:
Can you make an example or a screenshot? From the description it is hard to understand what the change does.

cmangin:
Sure. I am executing the following code that just generates some big goal with a lot of hypotheses:
```coq
Fixpoint alot (T : Prop) (n : nat) :=
match n with
| O => T
| S n => alot T n /\ T
end.

Fixpoint sobig (P : nat -> Type) (T : Prop) (n : nat) : Prop :=
match n with
| O => T
| S n => P n -> sobig P T n
end.

Goal alot (sobig (fun n => n = n) True 20) 5.
compute.
repeat (apply conj); intros;
replace True with (alot True 20) by admit; compute.
```

Without this PR, the proof window displays the following:
Without

With this PR, it displays the following:
With

I guess this is probably a matter of taste, if there are people strongly inclined towards both choices, I could try to add a setting.

gares (approved):

ppedrot:
It seems to have been the purpose of 08b7a19, but I'm not really sure.

cmangin:
You are right. Then 85fca507c6c4810d0858d6fbd8f5a1ece52e755c removed the use of tags, and f461e76 further removed the now dead code.
I've slightly changed my code to be closer to the original one, it should be better.

Zimmi48:
So indeed, this is a regression from 8.6. Maybe this deserves to be backported after all.

ppedrot (approved):
Given that this kind of code is not really reviewable, I trust that it's doing the right thing (although it probably doesn't anyways, no doubt we'll find some corner-case).

Zimmi48:
@ppedrot Can you merge?

PR #7473, opened 2018-05-09
[ci] Add mit-plv/cross-crypto

JasonGross:
I followed the code for fiat-crypto / fiat-parsers.  I hope I didn't
miss anything.

**Kind:** infrastructure.

PR #7934, opened 2018-06-26
Add mit-plv/bedrock2-ci to CI

andres-erbsen:
Previously discussed at https://github.com/coq/coq/pull/7878, https://github.com/mit-plv/bedrock2/issues/12, and https://github.com/mit-plv/bedrock2-ci/pull/1, development CI configuration at https://github.com/mit-plv/bedrock2-ci/blob/master/.travis.yml. I followed what Jason did to add cross-crypto at https://github.com/coq/coq/pull/7473.


.circleci/config.yml
```diff
@@ -78,6 +78,9 @@ jobs:
   # Build and prepare test environment
   build: *build-template
 
+  bedrock2:
+    <<: *ci-template
```
SkySkimmer: We're not adding to circle any more.

.gitlab-ci.yml
```diff
@@ -309,6 +309,9 @@ validate:edge+flambda:
 ci-bignums:
   <<: *ci-template
 
+ci-bignums:
+  <<: *ci-bedrock2
```
SkySkimmer: This is wrong.
- not bignums
- the yaml variable is ci-template

dev/ci/ci-basic-overlay.sh
```diff
@@ -147,6 +147,12 @@
 : "${bignums_CI_BRANCH:=master}"
 : "${bignums_CI_GITURL:=https://github.com/coq/bignums}"
 
+########################################################################
+# bedrock2
+########################################################################
+: "${bignums_CI_BRANCH:=master}"
+: "${bignums_CI_GITURL:=https://github.com/mit-plv/bedrock2-ci}"
```
SkySkimmer: `bignums` is wrong (bad copy paste?).

dev/ci/c-bedrock2.sh
```diff
@@ -0,0 +1,12 @@
...
+
+ci_dir="$(dirname "$0")"
+. "${ci_dir}/ci-common.sh"
+
+cross_crypto_CI_DIR="${CI_BUILD_DIR}/bedrock2-ci"
+
+git_checkout "${cross_crypto_CI_BRANCH}" "${cross_crypto_CI_GITURL}" "${cross_crypto_CI_DIR}"
+( cd "${cross_crypto_CI_DIR}" && git submodule update --init --recursive )
+
+( cd "${cross_crypto_CI_DIR}" && env EXPECTED_COQC_VERSION="$(coqc --version | sed -n 's/The Coq Proof Assistant, version \([^ ]*\) .*/\1/p')" make )
```
SkySkimmer: Why is this script talking about cross-crypto?

SkySkimmer (dismissed):
Please fix the copy pasting errors.

SkySkimmer:
Do we want to backport this? Is it compatible with Coq 8.8?

andres-erbsen:
Fixed the noted copy-paste errors.

bedrock2 is built with 8.7 and 8.8 (and now master) so I think backporting would make sense.

dev/ci/c-bedrock2.sh
```diff
@@ -0,0 +1,12 @@
...
+
+ci_dir="$(dirname "$0")"
+. "${ci_dir}/ci-common.sh"
+
+bedrock2_CI_DIR="${CI_BUILD_DIR}/bedrock2-ci"
+
+git_checkout "${bedrock2_CI_BRANCH}" "${bedrock2_CI_GITURL}" "${bedrock2_CI_DIR}"
+( cd "${bedrock2_CI_DIR}" && git submodule update --init --recursive )
+
+( cd "${bedrock2_CI_DIR}" && env EXPECTED_COQC_VERSION="$(coqc --version | sed -n 's/The Coq Proof Assistant, version \([^ ]*\) .*/\1/p')" make )
```
maximedenes: This is a hack, can't we get upstream support for it instead?

dev/ci/c-bedrock2.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+ci_dir="$(dirname "$0")"
+. "${ci_dir}/ci-common.sh"
+
+bedrock2_CI_DIR="${CI_BUILD_DIR}/bedrock2-ci"
```
maximedenes: Why not point to the bedrock2 repo instead? This will make it more difficult for us to submit patches upstream.

maximedenes (dismissed):
If I understand correctly, this setup will require two patches each time we break compatibility (one for bedrock2, and one to bump the submodule in bedrock2-ci). Can we use the bedrock2 repo directly, instead?

dev/ci/c-bedrock2.sh
samuelgruetter: I also think that this is a hack and I'll remove the `EXPECTED_COQC_VERSION` check from bedrock2.

dev/ci/c-bedrock2.sh
samuelgruetter: I see, so you're saying that if a Coq change breaks bedrock2 and you create a patch, you'd like to make only one PR (against bedrock2) and avoid an extra PR against bedrock2-ci, right?

bedrock2 depends on two other repos (bbv and riscv-coq), and they have to be separate repos. We did spend some thought on how to put everything together for CI, see eg [here](https://github.com/mit-plv/bedrock2/issues/12#issuecomment-399248610). The main concerns are diamond dependencies and having a fast workflow.

So, an alternative CI setup which would be more convenient for the Coq developers would be to have bbv and riscv-coq as git submodules of bedrock2. Like this, if a patch for bbv or riscv-coq is needed, it would still require two PRs, but patches for bedrock2 (where most of the code lives) would only require one PR. This seems possible, but once we get bitten by diamond dependencies on bbv, we'll have to complicate the build setup of bedrock2 a bit by allowing it to depend on a bbv at a different path, ignoring its bbv submodule. I'll keep thinking about how to do that nicely.

@maximedenes  does that make sense, or did I misunderstand something?

Zimmi48:
@andres-erbsen I don't see a problem with your own setup to run CI but I agree with @maximedenes that we shouldn't have to care about it. It seems that the problem here is similar to the one we have with stdpp-iris-lambda-rust and which was handled in a (different) ad hoc way. I think the problem is on our side: we should support such standard use case and have a generic way to answer the questions you were debating in mit-plv/bedrock2#12. I'll try to open an issue about this soon to have a generic discussion.

SkySkimmer (approved):
You need to squash commits.

.gitlab-ci.yml
```diff
@@ -309,6 +309,9 @@ validate:edge+flambda:
 ci-bignums:
   <<: *ci-template
 
+ci-bedrock2:
```
ejgallego: Alpha order is incorrect here.

.travis.yml
```diff
@@ -68,6 +68,9 @@ matrix:
     - if: NOT (type = pull_request)
       env:
       - TEST_TARGET="ci-bignums"
+    - if: NOT (type = pull_request)
+      env:
+      - TEST_TARGET="ci-bedrock2"
```
ejgallego: Don't update Travis anymore.

Makefile.ci
```diff
@@ -9,6 +9,7 @@
 ##########################################################################
 
 CI_TARGETS=ci-bignums \
+    ci-bedrock2 \
```
ejgallego: Problem with alpha sorting.

ejgallego (changes requested):

ejgallego (approved):
I am personally OK adding this as-is and improving the setup later, however I shall respect @maximedenes choice in that regard, so won't merge until I get green light from him.

maximedenes:
> I am personally OK adding this as-is and improving the setup later, however I shall respect @maximedenes choice in that regard, so won't merge until I get green light from him.

I was about to ask your opinion and ready to follow it :)

ejgallego:
> I was about to ask your opinion and ready to follow it :)

My opinion is that the testing benefit may outweight the temporary fixing annoyances; in particular as I hope we don't break bedrock so often; but I am clearly biased towards more extensive testing.

maximedenes:
We can merge as is and when we submit a bedrock2 patch, ask upstream to update the submodule.

samuelgruetter:
Not sure if everyone noticed but about 9 hours ago, I changed the CI setup of bedrock2, see [this comment](https://github.com/mit-plv/bedrock2/issues/12#issuecomment-400571524).
The dependencies (bbv and riscv-coq) are now submodules of bedrock2, and the "umbrella repo" bedrock2-ci does not appear in this PR any more.
Does that address your patch workflow concerns?

ejgallego:
> Does that address your patch workflow concerns?

Hi @samuelgruetter I am fine with the setup for now, mergin as soon as CI passes.

ejgallego:
@andres-erbsen note the silly linter error in Travis, can you fix and push again? Thanks.

ejgallego:
While you are at it with the linter please rebase to master.

PR #7937, opened 2018-06-27
Mention Consortium in README

maximedenes:
We are now actively looking for sponsors, let's make our communication
more visible.


Zimmi48 (commented):
I suggest you move this new section just above the Bug reports, because I don't like separating Bug reports and Contributing... (Other solution, would be to put it at the end.)

maximedenes:
Ok, I put it at the end.

Zimmi48 (approved):

PR #7939, opened 2018-06-27
Turn the CoqProject_file module into a pure ML file

ppedrot:
We also seize the opportunity to make the parsing algorithm slightly less crazy.

### DIFF
--- a/lib/coqProject_file.ml
+++ b/lib/coqProject_file.ml
@@ -90,37 +90,50 @@ let rec post_canonize f =
 
 exception Parsing_error of string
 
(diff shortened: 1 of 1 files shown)


ejgallego:
I think that I am delegating this to @gares , `coqfind` will introduce a different project format and uses its own camlp5-free parser.

maximedenes:
Maybe we can update the owner, then.

gares (approved):
I don't have time to check the code, but I'm ok with the change if it passes the tests

ppedrot:
Elpi failure is spurious, it requires merging two PRs to compile on master.

PR #7941, opened 2018-06-27
Extend QuestionMark to produce a better error message in case of missing record field

### LINKED ISSUE #6394: Record error messages should be improved.
```
Record foo (T : Type) :=
  { ops1 : T -> T -> T;
    ops2 : T -> T -> T }.

Definition x :=
  {| ops1 := Nat.add |}.
```

Coq replies:
```
Cannot infer this placeholder of type "nat -> nat -> nat".
```

Of course, I want the name of the missing fields.

bollu:
**Kind:** feature 

Fixes / closes #6394 
Workaround to PR: #7804 

Extend `QuestionMark` as suggested in #7804 to cleanly produce error messages for the case of a missing record field.

### DIFF
--- a/engine/evar_kinds.ml
+++ b/engine/evar_kinds.ml
@@ -21,12 +21,21 @@ type matching_var_kind = FirstOrderPatVar of Id.t | SecondOrderPatVar of Id.t
 
 type subevar_kind = Domain | Codomain | Body
 
+(* maybe this should be a Projection.t *)
+type record_field = { fieldname : Constant.t; recordname : Names.inductive }
+
+type question_mark = {
+     qm_obligation: obligation_definition_status;
+     qm_name: Name.t;
+     qm_record_field: record_field option;
+}
+
 type t =
   | ImplicitArg of GlobRef.t * (int * Id.t option)
      * bool (** Force inference *)
   | BinderType of Name.t
   | NamedHole of Id.t (* coming from some ?[id] syntax *)
-  | QuestionMark of obligation_definition_status * Name.t
+  | QuestionMark of question_mark
   | CasesType of bool (* true = a subterm of the type *)
   | InternalHole
   | TomatchTypeParameter of inductive * int
--- a/engine/termops.ml
+++ b/engine/termops.ml
@@ -114,7 +114,7 @@ let pr_evar_suggested_name evk sigma =
   | None -> match evi.evar_source with
   | _,Evar_kinds.ImplicitArg (c,(n,Some id),b) -> id
   | _,Evar_kinds.VarInstance id -> id
-  | _,Evar_kinds.QuestionMark (_,Name id) -> id
+  | _,Evar_kinds.QuestionMark {Evar_kinds.qm_name = Name id} -> id
   | _,Evar_kinds.GoalEvar -> Id.of_string "Goal"
   | _ ->
       let env = reset_with_named_context evi.evar_hyps (Global.env()) in
--- a/dev/ci/user-overlays/07941-bollu-questionmark-into-record-for-missing-record-field-error.sh
+++ b/dev/ci/user-overlays/07941-bollu-questionmark-into-record-for-missing-record-field-error.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+if [ "$CI_PULL_REQUEST" = "7941" ] || [ "$CI_BRANCH" = "jun-27-missing-record-field-error-message-quickfix" ]; then
+    Equations_CI_BRANCH=overlay-question-mark-extended-for-missing-record-field
+    Equations_CI_GITURL=https://github.com/bollu/Coq-Equations
+fi
--- a/engine/evar_kinds.mli
+++ b/engine/evar_kinds.mli
@@ -22,19 +22,18 @@ type matching_var_kind = FirstOrderPatVar of Id.t | SecondOrderPatVar of Id.t
 type subevar_kind = Domain | Codomain | Body
 
 (* maybe this should be a Projection.t *)
+(* Represents missing record field *)
 type record_field = { fieldname : Constant.t; recordname : Names.inductive }
 
 type question_mark = {
      qm_obligation: obligation_definition_status;
      qm_name: Name.t;
+     (* Tracks if the evar represents a missing record field *)
      qm_record_field: record_field option;
 }
 
-let default_question_mark = {
-  qm_obligation=Define true;
-  qm_name=Anonymous;
-  qm_record_field=None;
-}
+(* Default value of question_mark which is used most often *)
+val default_question_mark : question_mark
 
 type t =
   | ImplicitArg of GlobRef.t * (int * Id.t option)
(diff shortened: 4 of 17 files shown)


engine/evar_kinds.ml
```diff
@@ -21,12 +21,16 @@ type matching_var_kind = FirstOrderPatVar of Id.t | SecondOrderPatVar of Id.t
...
+type is_missing_record_field = { fieldname : Constant.t; recordname : Names.inductive }
+
 type t =
   | ImplicitArg of GlobRef.t * (int * Id.t option)
      * bool (** Force inference *)
   | BinderType of Name.t
   | NamedHole of Id.t (* coming from some ?[id] syntax *)
-  | QuestionMark of obligation_definition_status * Name.t
+  | QuestionMark of obligation_definition_status * Name.t *
+   is_missing_record_field option
```
ppedrot: While you're at it you should probably use a record here. This would prevent a lot of noise in the changes below, and you could also define a default record with everything set to `None` or something and use the design pattern of records using `with` clauses.

ppedrot: I'd have written it
```
type question_mark = {
     qm_obligation: obligation_definition_status;
     qm_name: Name.t;
     qm_record_field: is_missing_record_field option;
}                                                                                                                             
```
and probably `type record_field` instead of `type is_missing_record_field`.

ejgallego: I am not very fond of the the `qm_` pattern, it tends to get kind of messy; a non-perfect alternative @bollu is is put the record definition inside a module.

ppedrot: Using a module here is really heave artillery. This is not a very important type, I don't think it is worth it.

ppedrot:
@bollu modulo a few nitpicks, this PR is essentially lacking overlays. [Here](https://github.com/coq/coq/blob/master/dev/ci/user-overlays/README.md) is a description of the steps to perform to do so. You'll also need to rebase your PR and squash the commits.

bollu:
@ppedrot: Tried adding an overlay, let's say how it goes :D

bollu:
@ppedrot when I rebase, should the commit adding the overlay be separate?
Also, for whatever reason, the appveyour build fails with:

```
Makefile:77: cannot run complexity tests (no bogomips found)
3821BUILDING SUMMARY FILE
3822Makefile:77: cannot run complexity tests (no bogomips found)
```

ppedrot:
Usually we keep the feature and overlay commits separated indeed.

Zimmi48:
You can ignore this weird failure. AppVeyor is failing all over the place since yesterday: https://ci.appveyor.com/project/coq/coq/history

bollu:
@ppedrot - Rebased, cleaned up the history. Everything should compile now.

Any other nits to pick?

engine/evar_kinds.ml
```diff
@@ -21,12 +21,21 @@ type matching_var_kind = FirstOrderPatVar of Id.t | SecondOrderPatVar of Id.t
...
 
+(* maybe this should be a Projection.t *)
+type record_field = { fieldname : Constant.t; recordname : Names.inductive }
+
+type question_mark = {
+     qm_obligation: obligation_definition_status;
+     qm_name: Name.t;
+     qm_record_field: record_field option;
+}
+
```
ppedrot: I think you should define somewhere a default record instance
```
let default_question_mark = {
  Evar_kinds.qm_obligation=Evar_kinds.Define true;
  Evar_kinds.qm_name=Anonymous;
  Evar_kinds.qm_record_field=None;
}
```
and replaces all the explicit record creations in the other files by `{ default_question mark with ... }`. The one thing I don't know is where to put this default instance though. Probably not in `Evar_kinds` because this is a type-only file. Probably in `pretyping/glob_ops.ml`.

ppedrot (changes requested):

engine/evar_kinds.ml
Zimmi48: Death to type only files! (I'm serious.) Duplicate said file in a `.ml` and a `.mli` and put the default instance there.

ppedrot: I personally disagree, this leads to a shitload of redundancy when describing AST types. Also, as I mentioned already, OCaml relies on the same technique, so it is not as if we were something frowned upon by the language designers...

Zimmi48: Does OCaml have `.ml` only files? I don't think so. Anyways, I'm not sure that the OCaml compiler code is the right place to look at if you want advice on how to write OCaml code. As if you recommended reading the code of Coq's standard library to know how to write proper Coq code...

ppedrot: Fair enough. I prefer having duplicated code than no mli, so @bollu please copy this file into a mli file and add the default record in there.

bollu: @ppedrot is this notion of a "type-only file" something enforced by the coq codebase style guide? Why would there be such a concept as a "type-only file"? (OCaml noob, trying to understand the motivation for this)

Zimmi48: @bollu No it's something that has been changing a lot recently. And I personally want to get rid of them, and now I've @ppedrot on board.

ppedrot (approved):

bollu:
@ppedrot It looks like [vst failed](https://gitlab.com/coq/coq/-/jobs/82160635), with
```
COQC msl/Extensionality.v
While loading initial state:
Warning: Cannot open Free [cannot-open-path,filesystem]
While loading initial state:
Warning: Cannot open Custom [cannot-open-path,filesystem]
While loading initial state:
Warning: Cannot open Free [cannot-open-path,filesystem]
While loading initial state:
Warning: Cannot open Custom [cannot-open-path,filesystem]
File "./msl/Extensionality.v", line 1, characters 15-29:
Error: Unable to locate library VST.msl.Axioms.
```
I think this is something that has broken upstream, correct?

Also, the windows builds have been broken.

ppedrot:
@mattam82, @herbelin are you fine with the solution and the patch? I'll leave you a bit of time to answer, if I don't hear from you I'll consider that the simplicity of this PR doesn't require your explicit consent and merge it.

Zimmi48:
@bollu It's probably time to remove "[WIP]" from the title.

mattam82:
@ppedrot indeed it's fine

engine/termops.ml
```diff
@@ -189,7 +189,7 @@ let pr_decl (decl,ok) =
 
 let pr_evar_source = function
   | Evar_kinds.NamedHole id -> Id.print id
-  | Evar_kinds.QuestionMark _ -> str "underscore"
+  | Evar_kinds.QuestionMark _ -> str "underscore termops"
```
ppedrot: Just saw that: what's the reason for that diff? Is it a spurious change? In any case you should probably revert this line.

bollu: Hm, yes, this seems spurious, let me change this.

bollu:
@ppedrot: Updated!

PR #7948, opened 2018-06-28
Syntax for naming an existential variable

amblafont:
I knew this feature existed but I did not remember the syntax and I could not find the right syntax in the manual (hopefully it was mentionned recently in the mailing list)

**Kind:** documentation

### DIFF
--- a/doc/sphinx/language/gallina-extensions.rst
+++ b/doc/sphinx/language/gallina-extensions.rst
@@ -2228,7 +2228,7 @@ existential variable used in the same context as its context of definition is wr
    Check (fun x y => _) 0 1.
 
 Existential variables can be named by the user upon creation using
-the syntax ``?``\ `ident`. This is useful when the existential
+the syntax ``?``\ [`ident`]. This is useful when the existential
 variable needs to be explicitly handled later in the script (e.g.
 with a named-goal selector, see :ref:`goal-selectors`).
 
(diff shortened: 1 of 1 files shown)


doc/sphinx/language/gallina-extensions.rst
```diff
@@ -2228,7 +2228,7 @@ existential variable used in the same context as its context of definition is wr
    Check (fun x y => _) 0 1.
 
 Existential variables can be named by the user upon creation using
-the syntax ``?``\ `ident`. This is useful when the existential
+the syntax ``?``\ [`ident`]. This is useful when the existential
```
Zimmi48: Use ```:n:`?[@ident]` ``` instead.

Zimmi48 (commented):
Thanks!
I was the one who documented this and this was apparently badly migrated to Sphinx (cf. https://coq.inria.fr/distrib/V8.7.2/refman/gallina-ext.html#ExistentialVariables).

doc/sphinx/language/gallina-extensions.rst
amblafont: I don't know what that means but ok, I'll just copy and paste :)

Zimmi48 (approved):
Rendered here: https://coq.gitlab.io/-/coq/-/jobs/78085713/artifacts/_install_ci/share/doc/coq/sphinx/html/language/gallina-extensions.html#existential-variables

Note: when you can, we usually prefer two commits changing the same line in a single PR to be squashed together. Nonetheless, I'm going to merge this PR as-is because I want to put the 8.8.1 tag ASAP.

herbelin:
I don't know if this is worth publicity, but for the record, the following syntax is also allowed: `?[?foo]` which names the existential variable using the `foo` prefix and adding a numeric suffix or increasing an existing numeric suffix if `foo` already exists.

Issue #7555, opened 2018-05-18
Anomaly when running guardedness checker 


tchajed:
#### Version

```text
The Coq Proof Assistant, version 8.8+alpha (May 2018)
compiled on May 18 2018 14:55:55 with OCaml 4.06.0
```

master, a0da3a68d121

#### Operating system

macOS 10.13

#### Description of the problem

The following results in `Anomaly "grounding a non evar-free term"`.

```coq
Definition trivial : unit.
  Guarded.
```

This does still work, but just about any other use of `Guarded.` doesn't:

```coq
Definition trivial : unit.
  exact tt.
  Guarded.
```

Not fixed by #7449 or #7530.

ppedrot:
Duplicate of #7286.

PR #7952, opened 2018-06-28
Fix an algorithmic issue in the vernac guardedness checker.

### LINKED ISSUE #7286: Guarded relies on non-ground EConstr
Since the merge of #6454, the `Guarded` command is essentially useless, because it relies on term normalization that contain evars, and raises an anomaly as soon as there is an open goal.

```
Goal True.
Guarded.
(** Anomaly "grounding a non evar-free term" *)
```

ppedrot:
Calling the O(n) EConstr.to_constr function at every node is a very bad idea (tm).

EDIT: Fixes #7286.

### DIFF
--- a/pretyping/inductiveops.ml
+++ b/pretyping/inductiveops.ml
@@ -750,8 +750,11 @@ let type_of_projection_knowing_arg env sigma p c ty =
    syntactic conditions *)
 
 let control_only_guard env sigma c =
+  let c = Evarutil.nf_evar sigma c in
   let check_fix_cofix e c =
-    match kind (EConstr.to_constr sigma c) with
+    (** [c] has already been normalized upfront *)
+    let c = EConstr.Unsafe.to_constr c in
+    match kind c with
     | CoFix (_,(_,_,_) as cofix) ->
       Inductive.check_cofix e cofix
     | Fix (_,(_,_,_) as fix) ->


pretyping/inductiveops.ml
```diff
@@ -758,8 +758,11 @@ let type_of_projection_knowing_arg env sigma p c ty =
    syntactic conditions *)
 
 let control_only_guard env sigma c =
+  let c = Evarutil.nf_evar sigma c in
   let check_fix_cofix e c =
-    match kind (EConstr.to_constr sigma c) with
+    (** [c] has already been normalized upfront *)
+    let c = EConstr.Unsafe.to_constr c in
```
ejgallego: Don't we have a safe method to do this?

ppedrot: As witnessed by #7286, this is hopeless, the argument might contain unresolved evars.

mattam82 (approved):
Looks good. I believe using the to_constr there is better than rewriting everything with Constr.

pretyping/inductiveops.ml
```diff
@@ -758,8 +758,11 @@ let type_of_projection_knowing_arg env sigma p c ty =
    syntactic conditions *)
 
 let control_only_guard env sigma c =
+  let c = Evarutil.nf_evar sigma c in
```
SkySkimmer: Why don't we just to_constr here?

ppedrot: Because the types do not agree. There is no `iter_full_binder` for kernel terms...

SkySkimmer: So add it.

SkySkimmer: In fact it used to be on kernel terms until the commit which introduced this issue.

ejgallego: Indeed this is what I was going to say, the origin of this problem was a try not to duplicate that function; it is clear now that the motivation was ill-founded so we are bound to duplication.

ejgallego:
Add a test case for #7555 #7286

pretyping/inductiveops.ml
ppedrot: There are no `full_binders` functions in the kernel, I am not really willing to introduce just that one function there. Either we are uniform and provide this variant for all functions or we keep it for EConstr. Such functions are not required by the kernel so I'd advocate against kernel code inflation.

maximedenes: I don't understand this. Why would such combinators be needed on `econstr` but not `constr`? Kernel code inflation is irrelevant here IMHO, the code is generic, doesn't increase the complexity for bugs, and is low risk if tested properly.

ppedrot: I am fine with adding it, but then we need all such functions, not just `iter`.

maximedenes: I agree, but that's certainly not an argument for not adding one ;)

gares:
What is the vernacular guardedness checker?

pretyping/inductiveops.ml
ppedrot: OK, so it turns out it is harder than I thought. In order to implement the `full_binder` variants you need term lifting. But the latter is only defined in `Vars` so that the variants cannot be defined in the `Constr` module. What do?

SkySkimmer: Define it in vars, or move lifting to constr
Or have a module outside the kernel for stuff on constr that the kernel doesn't use.

ppedrot: Guys, this is really starting to be very complex for a simple fix that is 3-line patch...

ejgallego: Well, indeed; but on the other hand this is highligthing serious issues with the API; given that the fix seems far from "critical" we can take some time to discuss what'd be the proper way to address this.

SkySkimmer:
>What is the vernacular guardedness checker?

`Guarded.`

pretyping/inductiveops.ml
SkySkimmer: >OK, so it turns out it is harder than I thought. In order to implement the full_binder variants you need term lifting. But the latter is only defined in Vars so that the variants cannot be defined in the Constr module. What do?

Actually I don't see where fold_with_full_binders needs term lifting. It needs Context but that was moved before Constr so it's not a problem.

ppedrot: It needs term lifting to push fixpoint and cofixpoint binders. Look at `Environ.push_rec_types`.

SkySkimmer: The current code does 
~~~ocaml
  | Fix (_,(lna,tl,bl)) ->
      let n' = CArray.fold_left2 (fun c n t -> g (LocalAssum (n, t)) c) n lna tl in
      let fd = Array.map2 (fun t b -> (t,b)) tl bl in
      Array.fold_left (fun acc (t,b) -> f n' (f n acc t) b) acc fd
~~~
so is it broken?

ppedrot: Looks like it...

ppedrot: Interstingly it seems it has been broken since the very beginning, see d9dc86a6f7194c2e7ea704c95495955ca4f4b08d from November 2000 (!).

SkySkimmer: The annoying part is that IIUC once we have rec-rec fixpoints the lift will go away x_x

ppedrot: Well, I think we still have time to see that coming.

SkySkimmer:
I'll merge monday afternoon if nobody complains.

ejgallego:
I am not happy to see this in, but it is better than what we have now in master so green light.

PR #7080, opened 2018-03-27
Swapping Context and Constr and defining declarations on constr in Constr

herbelin:
**Kind:** architecture

The motivation for this PR is to eventually be able to use contexts of declarations in the definition of the `Case` constructor of `constr` (as discussed in the last steps of CEP [#34](https://github.com/coq/ceps/pull/34)).

It can also be seen as moving `Context.Rel.Declaration.t` and the 5 other `Context.*.t` of `context.ml` (i.e. the instantiation of the generic `pt` to `constr`/`types`) to a place where they are less isolated that they became after the generalization of `Context` to support `EConstr`. With this move, `Context` becomes exclusively polymorphic.

From the API point of view, this implies that `Constr` now includes `Context` and that the `t` types of `Context` are now only available in `Constr.Context.Rel.Declaration` and co and not any
more in `Context.Rel.Declaration`. So, more `open Constr` are explicitly needed (I hope this does not go against other ongoing design choices).

The inclusion of `Context` in `Constr` is done to the price of a pretty heavy boilerplate. Typically, I would just have wanted to be able to say:
```
include Context
  extending Rel.Declaration with t = (constr, types) Rel.Declaration.pt
  extending Rel with t = Declaration.t list
  extending Named.Declaration with t = (constr, types) Named.Declaration.pt
  extending Named with t = Named.Declaration.t list
  extending Compacted.Declaration with t = (constr, types) Compacted.Declaration.pt
  extending Compacted with t = Compacted.Declaration.t list
```
but I did not find how to do it concisely.

Note: I did not put a kind label. Maybe `redesign` is the closest one that I could found but it is not about a feature, it is purely architectural.

PR #7953, opened 2018-06-28
More efficient abstraction over variables in Cooking.

ppedrot:
Instead of repeatedly replacing the variables with a De Bruijn index and closing it, we do this in one pass. We furthermore share the abstraction over the context.

This source of slowdown was observed in lambda-rust.

### DIFF
--- a/kernel/cooking.ml
+++ b/kernel/cooking.ml
@@ -24,6 +24,7 @@ open Declarations
 open Univ
 
 module NamedDecl = Context.Named.Declaration
+module RelDecl = Context.Rel.Declaration
 
 (*s Cooking the constants. *)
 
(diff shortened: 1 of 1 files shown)


gares:
I've the impression that the new function should be somewhere in `Context` and be named `Named.to_rel` or `Rel.of_named`.

I'm also inferring from the code that `subst_var` replaces a variable by its position in the list. It is documented (I've just checked), but I think that, given the name is not particularly clear, you should sneak a comment in your code, eg "subst: maps named variables to De Bruijn indexes, the value of the index is the position of the variable in the list".

ppedrot:
> I've the impression that the new function should be somewhere in Context and be named Named.to_rel or Rel.of_named.

This requires explicitly depending on terms, so it goes against #7080. Better not mess too much with side PRs.

> given the name is not particularly clear, you should sneak a comment in your code

Will do.

ppedrot:
Added a comment.

gares:
> This requires explicitly depending on terms, so it goes against #7080

Can you be more explicit?

I see the modules I named shall be moved elsewhere, but this is not the point.
My point is that a function converting a context from named to rel should be in `Context`.
You can named it `rel_context_of_named_context` if this is how the types are called after #7080 was merged.

ppedrot:
I don't think either it makes much sense to put it in Cooking. Do you really think it's worthwhile to export this code?

maximedenes (approved):
The idea looks good, I tend to agree with @gares that `abstract_context` could find a better place, but I've no suggestion.

maximedenes:
So @gares, what should we do?

PR #7956, opened 2018-06-28
Rebuild coqtop$(EXE) in "make coqbinaries" in addition to coqtop.opt$(EXE).

### LINKED ISSUE #7758: Make target coqbinaries only rebuilds coqtop.opt, not coqtop
#### Version

Coq 8.9+alpha

#### Operating system

Windows and Linux

Since #6859, running `make` will generate two identical binaries `bin/coqtop` and `bin/coqtop.opt`. However, changing something then running `make coqbinaries` will only regenerate `bin/coqtop.opt`, leaving `bin/coqtop` unchanged.

cc @ejgallego

jfehrle:
Rebuild *.exe in "make coqbinaries" in addition to *.opt.exe for top binaries

Fixes #7758.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -211,7 +211,7 @@ archclean: clean-ide optclean voclean
 	rm -f $(ALLSTDLIB).*
 
 optclean:
-	rm -f $(COQTOPEXE) $(CHICKEN) $(TOPBIN)
+	rm -f $(COQTOPEXE) $(CHICKEN) $(TOPBINOPT)
 	rm -f $(TOOLS) $(PRIVATEBINARIES) $(CSDPCERT)
 	find . -name '*.cmx' -o -name '*.cmx[as]' -o -name '*.[soa]' -o -name '*.so' | xargs rm -f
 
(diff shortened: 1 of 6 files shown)


SkySkimmer:
As mentioned in the issue, this should be only for coqtop, not coqproofworker etc.

jfehrle:
Ah, I wasn't sure what @ejgallego meant by "The other 3 binaries are internal so indeed I guess the explicit naming is preferrable."  Emilio, do you agree I should not copy coqproofworker.opt.exe to coqproofworker.exe (and the other 2)?  Only coqtop.opt.exe?

ejgallego:
> do you agree I should not copy coqproofworker.opt.exe to coqproofworker.exe (and the other 2)? Only coqtop.opt.exe?

I do agree, only coqtop I think.

ejgallego:
The current maintainer for the make system is @gares

gares:
The makefile contains code like `...  && $(STRIP) $@ && $(CODESIGN) ...` hence you cannot put `$(HIDE)` as part of the value of `CODESIGN`.

jfehrle:
I tried a lot of alternatives for this define but couldn't get it to to work.  I suppose the alternatives are a) insert conditionals wherever there is CODESIGN command or b) let it be since it's a minor issue.

```
define codesign
if [ "$(CODESIGN)" != true ] then $(CODESIGN) $@; fi
endef
```

jfehrle:
This is ready for review.

FWIW, I noticed there were some places in the makefiles where only CODESIGN or STRIP are called, rather than both.  Hope that's correct.

ejgallego (approved):
The parts I understand look good to me.

gares (approved):
This PR is 1 line fix (`Makefile.build`, line 398) plus two unrelated changes, namely the renaming of `TOPBIN` to `TOPBINOPT` and (in a separate commit) the hiding of a useless (on windows and linux) message.

I won't nitpick this time, but next time please separate unrelated changes in separate PRs/commits.

ejgallego:
@gares are you taking this?

Issue #7915, opened 2018-06-25
deprecation warning in misctypes points to the wrong file


gares:
```
Warning 3: deprecated: IntroAnonymous
Use version in [Evarutil]
```
I found it in `Namegen`

PR #7961, opened 2018-06-29
[api] Fix wrong deprecation warning (#7915)

ejgallego:
Fixes: #7915.

Due to a change in the original misctypes removal PR, the deprecation
notice went out of sync.


gares (approved):

gares:
I think I can take this one even if I'm not the owner, it is so trivial..

PR #7962, opened 2018-06-29
[ssr] let the user override intro patterns

gares:
This PR revives the code doing dynamic dispatching of ssr2 (once demoed at a working group).
It is *very* preliminary, mainly to benchmark the cost of dynamic dispatching.

CC: @CohenCyril @amahboubi and @bgregoir  showed interest in this recently (maybe give a look at the test file). If the cost of dispatching is not too bad I can try to let one override also the other intro patterns.

As usual comments on the code by the Ltac internal's guru are welcome, CC: @ppedrot 

bench says no overhead:
```
┌──────────────────────────┬─────────────────────────┬─────────────────────────────────────┬─────────────────────────────────────┬─────────────────────────┬─────────────────┐
│                          │      user time [s]      │             CPU cycles              │          CPU instructions           │  max resident mem [KB]  │   mem faults    │
│                          │                         │                                     │                                     │                         │                 │
│             package_name │     NEW     OLD PDIFF   │           NEW           OLD PDIFF   │           NEW           OLD PDIFF   │     NEW     OLD PDIFF   │ NEW OLD PDIFF   │
├──────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│     coq-mathcomp-algebra │  198.96  199.93 -0.49 % │  552412295463  555015863617 -0.47 % │  777929340685  778356542439 -0.05 % │  646336  644156 +0.34 % │   0   1 -100.00 % │
├──────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│       coq-mathcomp-field │  462.97  464.90 -0.42 % │ 1288279041998 1293496276708 -0.40 % │ 2116792705975 2117369395833 -0.03 % │  803568  803504 +0.01 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│   coq-mathcomp-odd-order │ 1425.66 1430.76 -0.36 % │ 3971889780869 3986579773744 -0.37 % │ 6707676093761 6703742475407 +0.06 % │ 1366216 1350324 +1.18 % │ 224   0  +nan % │
├──────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│   coq-mathcomp-character │  288.55  289.47 -0.32 % │  802184064308  804989530668 -0.35 % │ 1170307404943 1170233786423 +0.01 % │ 1131468 1114704 +1.50 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│    coq-mathcomp-fingroup │   81.24   81.44 -0.25 % │  224222977563  224849540621 -0.28 % │  332908333324  332835955836 +0.02 % │  588648  586928 +0.29 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│    coq-mathcomp-solvable │  216.45  216.79 -0.16 % │  600219161579  601616391073 -0.23 % │  858087279404  857342084062 +0.09 % │  832552  860220 -3.22 % │   5   0  +nan % │
├──────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│ coq-mathcomp-real-closed │  184.10  184.23 -0.07 % │  511931441712  513005335333 -0.21 % │  777385527178  777131383258 +0.03 % │  832640  831680 +0.12 % │   2   1 +100.00 % │
├──────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│   coq-mathcomp-ssreflect │   62.26   62.21 +0.08 % │  170882509840  170337389872 +0.32 % │  241273247139  241037074639 +0.10 % │  533280  531708 +0.30 % │   0   0  +nan % │
├──────────────────────────┼─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────┼─────────────────┤
│              coq-bignums │   96.16   95.88 +0.29 % │  266073464377  265101193852 +0.37 % │  377734059966  377957425485 -0.06 % │  516732  516492 +0.05 % │  11   0  +nan % │
└──────────────────────────┴─────────────────────────┴─────────────────────────────────────┴─────────────────────────────────────┴─────────────────────────┴─────────────────┘
```

### DIFF
--- a/plugins/ssr/ssrcommon.ml
+++ b/plugins/ssr/ssrcommon.ml
@@ -1237,9 +1237,7 @@ let gentac gen gl =
 let genstac (gens, clr) =
   tclTHENLIST (old_cleartac clr :: List.rev_map gentac gens)
 
-let gen_tmp_ids
-  ?(ist=Geninterp.({ lfun = Id.Map.empty; extra = Tacinterp.TacStore.empty })) gl
-=
+let gen_tmp_ids ?(ist=Geninterp.empty_interp_sign) gl =
   let gl, ctx = pull_ctx gl in
   push_ctxs ctx
     (tclTHENLIST
--- a/plugins/ssr/ssrcommon.ml
+++ b/plugins/ssr/ssrcommon.ml
@@ -27,6 +27,7 @@ open Ssrmatching_plugin
 open Ssrmatching
 open Ssrast
 open Ssrprinters
+open Ssrdispatch
 
 module RelDecl = Context.Rel.Declaration
 module NamedDecl = Context.Named.Declaration
@@ -63,8 +64,11 @@ let rec check_hyps_uniq ids = function
   | [] -> ()
 
 let check_hyp_exists hyps (SsrHyp(_, id)) =
-  try ignore(Context.Named.lookup id hyps)
-  with Not_found -> errorstrm Pp.(str"No assumption is named " ++ Id.print id)
+  try ignore(Context.Named.lookup id hyps); Proofview.tclUNIT ()
+  with Not_found ->
+   Proofview.(tclLIFT (NonLogical.raise
+     CErrors.(UserError (Some "ssreflect",
+       Pp.(str"No assumption is named " ++ Id.print id)))))
 
 let test_hypname_exists hyps id =
   try ignore(Context.Named.lookup id hyps); true
@@ -151,7 +155,6 @@ open Genarg
 open Stdarg
 open Pp
 
-let errorstrm x = CErrors.user_err ~hdr:"ssreflect" x
 let anomaly s = CErrors.anomaly (str s)
 
 (* Tentative patch from util.ml *)
@@ -1438,8 +1441,11 @@ end
 
 let return ~orig_name:_ ~new_name:_ = tclUNIT ()
 
-let tclINTRO_ID id = tclINTRO ~id:(Some id) ~conclusion:return
-let tclINTRO_ANON = tclINTRO ~id:None ~conclusion:return
+let tclINTRO_ID = register_overloaded "intro_id" (Arg(Stdarg.wit_ident, Nil))
+  (fun id -> tclINTRO ~id:(Some id) ~conclusion:return)
+
+let tclINTRO_ANON = register_overloaded "intro_anon" Nil
+   (tclINTRO ~id:None ~conclusion:return)
 
 let tclRENAME_HD_PROD name = Goal.enter begin fun gl ->
   let convert_concl_no_check t =
--- a/test-suite/ssr/ssr_overload_ipats.v
+++ b/test-suite/ssr/ssr_overload_ipats.v
@@ -1,0 +1,24 @@
+Require Import ssreflect.
+
+Axiom frl : (nat -> Prop) -> Prop.
+Axiom frl_intro : forall P : nat -> Prop, (forall x, P x) -> (frl P).
+Notation "`all x , P" := (frl (fun x : _ => P)) (x ident, at level 20).
+
+Lemma test : frl (fun x => x = 0).
+Proof.
+Fail move=> x.
+Fail move=> ?.
+Abort.
+
+Hint Extern id 0 (frl _) => apply: frl_intro => id : ssr_intro_id.
+Hint Extern 0 (frl _) => apply: frl_intro => ?  : ssr_intro_anon.
+Hint Extern 0 (frl _) => apply: frl_intro => _  : ssr_intro_drop.
+Hint Extern id 0 => idtac "check" id : ssr_check_hyp_exists.
+Hint Extern id 0 => clear id || idtac : ssr_intro_delayed_clear.
+Hint Extern o d x 0 => idtac "occ" o "dir" d "subject" x : ssr_intro_subst.
+
+Lemma test : forall w, `all y, `all x, (x = w).
+Proof.
+move=> ? _ [|x]. Undo.
+move=> w {-3}<- x {y q}.
+Abort.
--- a/plugins/ltac/g_ltac.ml4
+++ b/plugins/ltac/g_ltac.ml4
@@ -342,9 +342,11 @@ GEXTEND Gram
           Vernacexpr.VernacProof (ta,Some l) ] ]
   ;
   hint:
-    [ [ IDENT "Extern"; n = natural; c = OPT Constr.constr_pattern ; "=>";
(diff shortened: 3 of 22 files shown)


plugins/ssr/ssrcommon.ml
```diff
@@ -1440,7 +1440,20 @@ end
 
 let return ~orig_name:_ ~new_name:_ = tclUNIT ()
 
-let tclINTRO_ID id = tclINTRO ~id:(Some id) ~conclusion:return
+
+
+let tclINTRO_ID_ltac ist = function
```
gares: This ugly code can be removed by using GADTs, as I did here: https://github.com/LPCIC/elpi/blob/master/src/elpi_API.mli#L507 (actually it would be simpler here). The FFI then looks like:  https://github.com/LPCIC/elpi/blob/master/src/elpi_builtin.ml#L476 where the ML function has a type described by the GADT. Do we have anything like that in Coq already? @ppedrot ?

test-suite/ssr/ssr_overload_ipats.v
```diff
@@ -0,0 +1,17 @@
...
+Axiom frl : (nat -> Prop) -> Prop.
+Axiom frl_intro : forall P : nat -> Prop, (forall x, P x) -> (frl P).
+Notation "`all x , P" := (frl (fun x : _ => P)) (x ident, at level 20).
+
+Lemma test : frl (fun x => x = 0).
+Proof.
+Fail move=> x.
+Abort.
+
+Ltac ssr_intro_id id ::= apply: frl_intro; ssr_intro_id_orig id.
```
gares: @CohenCyril this is how it works

CohenCyril: Would it be possible to index the tactic to use with the head symbol of the goal?
Like
```
Add SsrIntro head_symbol := my_tactic
```

gares: Open question: how do you pass to the hints in a db some arguments (the name of the hyp in this case). @Zimmi48 do you happen to know if that is feasible?

Zimmi48: Sorry @gares, I missed the mention somehow. I am not exactly sure what you have in mind, but it seems to me that you would have to rely on passing arguments to the tactic that you put in your hint extern. Something along the lines of:

```
Ltac foo x := idtac x.
Hint Extern 1 => foo : my_ssr_hints.
Goal True.
  (* move will internally call foo with argument string:("bla") *)
  foo string:("bla")
```

Not sure how realistic this is in practice. But it's definitely OK to extend the "Hint" primitives for such use-case if required.

gares: Your example is mostly OK, the minor difference is that the main call is not `foo` but an ML piece of code, eg `move=> bla`.

If you point me to the code of hints I'll give it a try, maybe it is not that hard. I think the syntax would be more like
```
HintExtern 1 => fun id => foo id : hintdb.
```

Zimmi48: The main code of hints is in `tactics/hints.ml{i,}`.

gares: Thanks, loks like there is nothing to change, since the tactic is an expression (hence can be a fun...). I'll try to hook this up and see how it goes.

gares:
I'm finally happy with the Ltac/ML bridge (its API).

@maximedenes you could give a look at `ssreflect.v` and `ssrcommon.ml` in order to get an idea of what this overloading mechanism would look like on the ssr side of things. Maybe also try to propose a rephrasing of the comment in `ssrdispatch.mli`, I really cant make it decent.

@CohenCyril the file in the test-suite has one example of what the user is expected to do in order to overload `=> x` (no more Ltac symbol overriding, but rather a new Hint Extern as you suggested).

@ppedrot at some point in time I thought that the `ffi` datatype that I expose in `ssrdispatch` was somewhere in ltac, but I could not find it (I found one, but seems tied to the parsing code). It is used in `ssrcommon` to give you an idea of what I need. When you have time please review `ssrdispatch.ml` since I'm not very confident I'm calling Ltac in the most orthodox way (you gave me that code once, but the context was different).

ppedrot:
This PR has stalled. I am closing it, @gares you should reopen it when deemed ready.

maximedenes:
@ppedrot AFAICT, this PR was expecting actions from three developers, including you and me. Why not answer the message above rather than closing the PR?

ppedrot:
I think we discussed at least some of the above points at a CUDW, didn't we?

maximedenes:
Honestly, I don't remember. We should at least log the conclusions here.

CohenCyril:
@gares @ppedrot @maximedenes Sorry for not answering eariler, even though this PR is of strong interest to me.
I like it as it is, I would love if we could move forward with it.

CohenCyril:
also, I do not remember what was discussed and/or decided at the CUDW...

gares:
I don't recall either.

I believe that there are two possible existing clients, and we should come up with some real test case:
- MC analysis (epsilon/delta)
- Iris

Also, the current code only considers ltac1 override while one should also consider ltac2 these days I guess.

One option is to revive this at the next CUDW, if there is interest

mattam82:
Two more clients for this:
- After discussion with Cyril, we think `Hint Extern If` should rather be replaced by a `Hint Extern Rec self pat => tac` for typeclasses resolution, where `self` references to the ongoing typeclass resolution tactic. In addition to using the exceptions of Ltac2, we should be able to express all the desired cuts we want flexibly using that.
- Definitely I would like to make depelim / dependent destruction of equations part of the intro patterns syntax, and this could be a way to help (although I guess for now you considered only string arguments, we could have more general Ltac2 datastructures there).

I'd like to discuss this (together with the ongoing typeclass changes) at the Coq Call tomorrow.
@gares @ppedrot just a heads up :)

PR #7964, opened 2018-06-29
Document that GITURL variables shouldn't have a trailing .git anymore.

Zimmi48:
This allows to append `/archive` at the end (cf. #7831).


ejgallego (approved):
LGTM, thanks. Let's keep an eye tho on not introducing github-specific stuff here c.f. #7707

Zimmi48:
Good point. I'll keep that in mind. See also https://github.com/coq/coq/issues/6724#issuecomment-401592887.

ejgallego:
Yeah we may need to add another variable for ARCHIVE; but for now things are looking good, great job!

PR #7969, opened 2018-06-30
doc: typesetting and hyperlinks in Syntax Extensions

Lysxia:
**Kind:** documentation

### DIFF
--- a/doc/sphinx/user-extensions/syntax-extensions.rst
+++ b/doc/sphinx/user-extensions/syntax-extensions.rst
@@ -10,15 +10,14 @@ parses and prints objects, i.e. the translations between the concrete
 and internal representations of terms and commands.
 
 The main commands to provide custom symbolic notations for terms are
-``Notation`` and ``Infix``. They are described in section :ref:`Notations`. There is also a
-variant of ``Notation`` which does not modify the parser. This provides with a
-form of abbreviation and it is described in Section :ref:`Abbreviations`. It is
+:cmd:`Notation` and :cmd:`Infix`; they will be described in the
+:ref:`next section <Notations>`. There is also a
+variant of :cmd:`Notation` which does not modify the parser; this provides a
+form of :ref:`abbreviation <Abbreviations>`. It is
 sometimes expected that the same symbolic notation has different meanings in
-different contexts. To achieve this form of overloading, |Coq| offers a notion
-of interpretation scope. This is described in Section :ref:`Scopes`.
-
-The main command to provide custom notations for tactics is ``Tactic Notation``.
-It is described in Section :ref:`TacticNotation`.
+different contexts; to achieve this form of overloading, |Coq| offers a notion
+of :ref:`interpretation scopes <Scopes>`.
+The main command to provide custom notations for tactics is :cmd:`Tactic Notation`.
 
 .. coqtop:: none
 
@@ -44,7 +43,7 @@ logical conjunction (and). Such a notation is declared by
 
    Notation "A /\ B" := (and A B).
 
-The expression :g:`(and A B)` is the abbreviated term and the string ``"A /\ B"``
+The expression :g:`(and A B)` is the abbreviated term and the string :g:`"A /\ B"`
 (called a *notation*) tells how it is symbolically written.
 
 A notation is always surrounded by double quotes (except when the
@@ -66,15 +65,15 @@ example.
 
 A notation binds a syntactic expression to a term. Unless the parser
 and pretty-printer of Coq already know how to deal with the syntactic
-expression (see 12.1.7), explicit precedences and associativity rules
-have to be given.
+expression (see :ref:`ReservingNotations`), explicit precedences and
+associativity rules have to be given.
 
 .. note::
 
    The right-hand side of a notation is interpreted at the time the notation is
-   given. In particular, disambiguation of constants, implicit arguments (see
-   Section :ref:`ImplicitArguments`), coercions (see Section :ref:`Coercions`),
-   etc. are resolved at the time of the declaration of the notation.
+   given. In particular, disambiguation of constants, :ref:`implicit arguments
+   <ImplicitArguments>`, :ref:`coercions <Coercions>`, etc. are resolved at the
+   time of the declaration of the notation.
 
 Precedences and associativity
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1017,8 +1027,8 @@ Interpretation scopes used in the standard library of Coq
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 We give an overview of the scopes used in the standard library of Coq.
-For a complete list of notations in each scope, use the commands Print
-Scopes or Print Scope scope.
+For a complete list of notations in each scope, use the commands ``Print
+Scopes`` or ``Print Scope scope``.
```
Zimmi48: You could make this ```:cmd:`Print Scopes` or :cmd:`Print Scope` ``` by first adding indexes to these command "variants" (variants do not get an index by default). Their definitions are just a bit below and to add an index, use option `:name:`. Note that I don't really understand why both are marked as variants, probably one of them shouldn't.

If you do this change, you can also reindent the variants inside the block defined by the main command...

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -984,7 +994,7 @@ Binding types of arguments to an interpretation scope
 
    :cmd:`About`
      The command to show the scopes bound to the arguments of a
-     function is described in Section 2.
+     function is described in _extensionsofgallina.
```
Zimmi48: According to the last version of the manual before the migration https://coq.inria.fr/distrib/V8.7.2/refman/syntax-extensions.html#scopes, the link was just to the definition of the command `About` (https://coq.inria.fr/distrib/V8.7.2/refman/vernacular.html#About). So you can probably remove this second link and rephrase the whole sentence.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -690,10 +695,14 @@ side. E.g.:
...
-**Syntax of notations**
+.. _NotationSyntax:
+
+Syntax of notations
++++++++++++++++++++
 
 The different syntactic variants of the command Notation are given on the
-following figure. The optional :token:`scope` is described in the Section 12.2.
+following figure. The optional :token:`scope` is described in
+:ref:`Scopes`.
```
Zimmi48: As you can see the token `scope` is not clickable because it wasn't defined anywhere. I don't really know whether it should be added to the grammar below, or be defined in the text by replacing the use of ```:token:`scope` ``` at one point by ```:production:`scope` ``` as described here https://github.com/coq/coq/tree/master/doc/sphinx#coq-roles. By looking at the Coq parser, we can see that scopes are just idents. So I guess it makes sense to not extend the grammar (like types which are just special kinds of terms, scopes are just idents with a special role) but to use an inline `production` instead.

Zimmi48 (commented):
Thanks a lot for this! I've added a few comments hinting how to improve this chapter even more. These are are mere suggestions, not requests.

doc/sphinx/user-extensions/syntax-extensions.rst
Lysxia: This sentence actually seems out of place, and there's another `seealso` block pointing to that command in the subsection above. So I'll remove it.

doc/sphinx/user-extensions/syntax-extensions.rst
Lysxia: Okay, let's use `production`.

Lysxia:
Thanks for the nice suggestions!

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1116,18 +1119,19 @@ Displaying informations about scopes
...
-   scope is bound, if any.
-
-.. cmdv:: Print Scopes
+.. cmd:: Print Scopes
 
    This displays all the notations, delimiting keys and corresponding
    class of all the existing interpretation scopes. It also displays the
    lonely notations.
 
+.. cmdv:: Print Scope scope
```
Zimmi48: If you don't mind, add three more spaces in front of this whole block (then the variant will appear more clearly as a variant).

Zimmi48: And BTW you can use `@scope` in the `cmdv` to make it clickable and displayed differently.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1116,18 +1119,19 @@ Displaying informations about scopes
...
+   :name: Print Scope
+
+   This displays all the notations defined in interpretation scope ``scope``.
```
Zimmi48: You can use ```:token:`scope` ``` here.

Zimmi48 (commented):
Some more comments on your most recent modifications.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -892,27 +901,27 @@ Binding arguments of a constant to an interpretation scope
    turn to be themselves arguments of a reference are interpreted
    accordingly to the arguments scopes bound to this reference.
 
-.. cmdv:: Arguments @qualid : clear scopes
+   .. cmdv:: Arguments @qualid : clear scopes
 
-   Arguments scopes can be cleared with :n:`Arguments @qualid : clear scopes`.
+      Arguments scopes can be cleared with :n:`Arguments @qualid : clear scopes`.
```
Zimmi48: Maybe this sentence would be worth to reformulate, because like this it's pretty useless.

Zimmi48 (approved):
Once again, thanks a lot for this clean-up / fix-up.
While going through the rendered version, I noticed a few other problems. Feel free to fix them or not, I'll merge this PR either way:

- The first command (`Notation`) has its description that is not indented.
- The first command is just a name, it only presents its grammar through examples. You can reuse what is given for `Infix` and abbreviations to improve it.
- `modifier` is not a grammar entry: `modifiers` is but the `s` should probably be removed since it only represents one modifier at a time.
- In the grammar `num` should be used instead of `natural`.

Lysxia:
Thanks, I'll leave it at that for the moment.

Zimmi48:
OK, merging then. Same for your other PR?

Lysxia:
Sure!

Issue #5975, opened 2017-10-18
Coq 8.7.0 test-suite fails (when Coq not configured -local and not installed)


JasonGross:
I'm not sure what's up with `Error: cannot guess a path for Coq libraries; please use -coqlib option`.  See [this launchpad log](https://launchpadlibrarian.net/341301069/buildlog_ubuntu-precise-amd64.coq-8.7.0_8.7.0-1~precise~ppa79_BUILDING.txt.gz) for more details.
```
make[5]: Entering directory `/<<PKGBUILDDIR>>/test-suite'
Makefile:177: warning: undefined variable `TRAVIS'
logs/misc/printers.log
==========> TESTING misc/printers.sh <==========
Error during initialization: 
Error: cannot guess a path for Coq libraries; please use -coqlib option
0m0.000000s 0m0.000000s
0m0.050000s 0m0.000000s
==========> FAILURE <==========
    misc/printers.sh...Error!

logs/coq-makefile/uninstall1.log
../template/init.sh: line 8: git: command not found
Error: cannot guess a path for Coq libraries; please use -coqlib option
==========> FAILURE <==========
    coq-makefile/uninstall1/run.sh...Error!

logs/coq-makefile/findlib-package.log
../template/init.sh: line 8: git: command not found
make[5]: Entering directory `/<<PKGBUILDDIR>>/test-suite/coq-makefile/findlib-package/findlib/foo'
make[5]: Leaving directory `/<<PKGBUILDDIR>>/test-suite/coq-makefile/findlib-package/findlib/foo'
Error: cannot guess a path for Coq libraries; please use -coqlib option
==========> FAILURE <==========
    coq-makefile/findlib-package/run.sh...Error!

logs/coq-makefile/arg.log
../template/init.sh: line 8: git: command not found
Error: cannot guess a path for Coq libraries; please use -coqlib option
==========> FAILURE <==========
    coq-makefile/arg/run.sh...Error!

logs/coq-makefile/extend-subdirs.log
../template/init.sh: line 8: git: command not found
Error: cannot guess a path for Coq libraries; please use -coqlib option
==========> FAILURE <==========
    coq-makefile/extend-subdirs/run.sh...Error!

logs/coq-makefile/only.log
../template/init.sh: line 8: git: command not found
Error: cannot guess a path for Coq libraries; please use -coqlib option
==========> FAILURE <==========
    coq-makefile/only/run.sh...Error!

logs/coq-makefile/coqdoc1.log
[...]

PR #7973, opened 2018-07-01
Add a test build on NixOS to GitLab CI.

Zimmi48:
Unfortunately, we can't run the test-suite until #5975 is fixed.
We pin `default.nix` again to make the CI build predictable.
As in Windows builds, we need to override the default `before_script`. Shouldn't we get rid of this default and put it in the template instead?

**Kind:** infrastructure.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -233,6 +233,22 @@ windows32:
   variables:
     ARCH: "32"
 
+pkg:nix:
+  image: nixorg/nix:latest # Minimal NixOS image which doesn't even contain git
+  stage: test
+  variables:
+    GIT_STRATEGY: none
+  dependencies: [] # We don't need to download build artifacts
+  before_script: [] # We don't want to use the shared 'before_script'
+  script:
+    - export TMPDIR=$PWD
+    - nix-build "$CI_PROJECT_URL/-/archive/$CI_COMMIT_SHA.tar.gz" -K
+  artifacts:
+    name: "$CI_JOB_NAME.logs"
+    when: on_failure
+    paths:
+      - nix-build-coq.drv-0/*/test-suite/logs
+
 warnings:base:
   <<: *warnings-template
 
--- a/default.nix
+++ b/default.nix
@@ -21,7 +21,11 @@
 # Once the build is finished, you will find, in the current directory,
 # a symlink to where Coq was installed.
 
-{ pkgs ? (import <nixpkgs> {})
+{ pkgs ?
+    (import (fetchTarball {
+      url = "https://github.com/NixOS/nixpkgs/archive/060a98e9f4ad879492e48d63e887b0b6db26299e.tar.gz";
+      sha256 = "1lzvp3md0hf6kp2bvc6dbzh40navlyd51qlns9wbkz6lqk3lgf6j";
+    }) {})
 , ocamlPackages ? pkgs.ocaml-ng.ocamlPackages_4_06
 , buildIde ? true
 , buildDoc ? true
--- a/default.nix
+++ b/default.nix
@@ -9,14 +9,14 @@
 
 # nix-shell supports the --arg option (see Nix doc) that allows you for
 # instance to do this:
-# $ nix-shell --arg ocamlPackages "(import <nixpkgs> {}).ocamlPackages_latest" --arg buildIde false
+# $ nix-shell --arg ocamlPackages "(import <nixpkgs> {}).ocaml-ng.ocamlPackages_4_05" --arg buildIde false
 
 # You can also compile Coq and "install" it by running:
 # $ make clean # (only needed if you have left-over compilation files)
 # $ nix-build
 # at the root of the Coq repository.
 # nix-build also supports the --arg option, so you will be able to do:
-# $ nix-build --arg doCheck false
+# $ nix-build --arg doInstallCheck false
 # if you want to speed up things by not running the test-suite.
 # Once the build is finished, you will find, in the current directory,
 # a symlink to where Coq was installed.
@@ -25,7 +25,7 @@
 , ocamlPackages ? pkgs.ocaml-ng.ocamlPackages_4_06
 , buildIde ? true
 , buildDoc ? true
-, doCheck ? true
+, doInstallCheck ? true
 }:
 
 with pkgs;
@@ -58,13 +58,13 @@ stdenv.mkDerivation rec {
              ps.antlr4-python3-runtime ps.sphinxcontrib-bibtex ]))
      antlr4
 
-  ] else []) ++ (if doCheck then
+  ] else []) ++ (if doInstallCheck then
 
     # Test-suite dependencies
     # ncurses is required to build an OCaml REPL
     optional (!versionAtLeast ocaml.version "4.07") ncurses
     ++ [
-    python
+    python2
     rsync
     which
     ocamlPackages.ounit
@@ -90,10 +90,18 @@ stdenv.mkDerivation rec {
 
   prefixKey = "-prefix ";
 
-  buildFlags = [ "world" ] ++ optional buildDoc "doc-html";
+  buildFlags = [ "world" "byte" ] ++ optional buildDoc "doc-html";
 
-  installTargets = [ "install" ] ++ optional buildDoc "install-doc-html";
+  installTargets =
+    [ "install" "install-byte" ] ++ optional buildDoc "install-doc-html";
 
-  inherit doCheck;
+  inherit doInstallCheck;
+
+  preInstallCheck = ''
+    patchShebangs tools/
+    patchShebangs test-suite/
+  '';
+
+  installCheckTarget = [ "check" ];
 
 }
--- a/default.nix
+++ b/default.nix
@@ -40,9 +40,9 @@ stdenv.mkDerivation rec {
   name = "coq";
 
   buildInputs = [
-
-    # Coq dependencies
     hostname
+    python2 time # coq-makefile timing tools
+
   ] ++ (with ocamlPackages; [
     ocaml
     findlib
@@ -67,13 +67,9 @@ stdenv.mkDerivation rec {
     # Test-suite dependencies
     # ncurses is required to build an OCaml REPL
     optional (!versionAtLeast ocaml.version "4.07") ncurses
-    ++ [
-    python2
-    rsync
-    which
-    ocamlPackages.ounit
+    ++ [ ocamlPackages.ounit rsync which ]
 
-  ] else []) ++ (if lib.inNixShell then [
+  else []) ++ (if lib.inNixShell then [
     ocamlPackages.merlin
     ocamlPackages.ocp-indent
     ocamlPackages.ocp-index
--- a/default.nix
+++ b/default.nix
@@ -42,51 +42,33 @@ stdenv.mkDerivation rec {
   buildInputs = [
     hostname
     python2 time # coq-makefile timing tools
-
-  ] ++ (with ocamlPackages; [
-    ocaml
-    findlib
-    camlp5_strict
-    num
-
-  ]) ++ (if buildIde then [
-
-    # CoqIDE dependencies
-    ocamlPackages.lablgtk
-
-  ] else []) ++ (if buildDoc then [
-
+  ]
+  ++ (with ocamlPackages; [ ocaml findlib camlp5_strict num ])
+  ++ optional buildIde ocamlPackages.lablgtk
+  ++ optionals buildDoc [
     # Sphinx doc dependencies
     pkgconfig (python3.withPackages
       (ps: [ ps.sphinx ps.sphinx_rtd_theme ps.pexpect ps.beautifulsoup4
              ps.antlr4-python3-runtime ps.sphinxcontrib-bibtex ]))
-     antlr4
-
-  ] else []) ++ (if doInstallCheck then
-
+    antlr4
+  ]
+  ++ optionals doInstallCheck (
     # Test-suite dependencies
     # ncurses is required to build an OCaml REPL
     optional (!versionAtLeast ocaml.version "4.07") ncurses
     ++ [ ocamlPackages.ounit rsync which ]
-
-  else []) ++ (if lib.inNixShell then [
-    ocamlPackages.merlin
-    ocamlPackages.ocp-indent
-    ocamlPackages.ocp-index
-
-    # Dependencies of the merging script
-    jq
-    curl
-    git
-    gnupg
-
-  ] else []);
+  )
+  ++ optionals lib.inNixShell (
+    [ jq curl git gnupg ] # Dependencies of the merging script
+    ++ (with ocamlPackages; [ merlin ocp-indent ocp-index ]) # Dev tools
+  );
 
   src =
     if lib.inNixShell then null
     else
       with builtins; filterSource
-        (path: _: !elem (baseNameOf path) [".git" "result" "bin"]) ./.;
+        (path: _:
+           !elem (baseNameOf path) [".git" "result" "bin" "_build_ci"]) ./.;
 
   prefixKey = "-prefix ";
 
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,4 @@
+# Some developers don't want a pinned nix-shell by default.
+# If you want to use the pin nix-shell or a more sophisticated set of arguments:
+# $ nix-shell default.nix
+import ./default.nix { pkgs = import <nixpkgs> {}; }
--- a/tools/CoqMakefile.in
+++ b/tools/CoqMakefile.in
@@ -65,20 +65,20 @@ VERBOSE ?=
 # Time the Coq process (set to non empty), and how (see default value)
 TIMED?=
 TIMECMD?=
-# Use /usr/bin/env time on linux, gtime on Mac OS
+# Use command time on linux, gtime on Mac OS
 TIMEFMT?="$* (real: %e, user: %U, sys: %S, mem: %M ko)"
 ifneq (,$(TIMED))
-ifeq (0,$(shell /usr/bin/env time -f $(TIMEFMT) true >/dev/null 2>/dev/null; echo $$?))
-STDTIME?=/usr/bin/env time -f $(TIMEFMT)
+ifeq (0,$(shell command time -f $(TIMEFMT) true >/dev/null 2>/dev/null; echo $$?))
+STDTIME?=command time -f $(TIMEFMT)
 else
 ifeq (0,$(shell gtime -f $(TIMEFMT) true >/dev/null 2>/dev/null; echo $$?))
 STDTIME?=gtime -f $(TIMEFMT)
 else
-STDTIME?=time
+STDTIME?=command time
 endif
 endif
 else
-STDTIME?=/usr/bin/env time -f $(TIMEFMT)
+STDTIME?=command time -f $(TIMEFMT)
 endif
 
 # Coq binaries


ejgallego (commented):
I fail to see why this is useful. The purpose of Coq's CI is not to test every know distribution, much less a minoritary one.

I guess this should rather be on Nix's CI system.

Zimmi48:
This was requested to me after I fixed yet another incompatibility that had been introduced in Coq.

Zimmi48:
This test can also be very useful to prevent any "accidental" introduction of a new dependency.

Of course, those two advantages would be much stronger if we could run the test-suite too.

ejgallego:
What part of Coq is this testing that is not tested elsewhere?



> This was requested to me after I fixed yet another incompatibility that had been introduced in Coq.

Can you say more?

> This test can also be very useful to prevent any "accidental" introduction of a new dependency.

Umm, why our current system does not prevent this?

.gitlab-ci.yml
```diff
@@ -233,6 +233,17 @@ windows32:
   variables:
     ARCH: "32"
 
+nixos:
+  image: nixorg/nix:latest # Minimal NixOS image which doesn't even contain git
```
ejgallego: This seems like a non-reproducible recipe.

default.nix
```diff
@@ -21,7 +21,11 @@
 # Once the build is finished, you will find, in the current directory,
 # a symlink to where Coq was installed.
 
-{ pkgs ? (import <nixpkgs> {})
+{ pkgs ?
```
ejgallego: Why this change? Is that fetching a concrete Nix image?

.gitlab-ci.yml
```diff
@@ -233,6 +233,17 @@ windows32:
   variables:
     ARCH: "32"
 
+nixos:
```
ejgallego: name this `distro:nixos` please.

Zimmi48:
Sorry, I wasn't expecting this change to be controversial, so I hadn't really prepared arguments to defend it.

> The purpose of Coq's CI is not to test every know distribution, much less a minoritary one.

This is not any distribution, this is a test that everything works with the Nix package manager, which is of much more general interest. We already support this package manager by providing a `default.nix` so it's probably a good idea to test it as well. Using a NixOS image instead of an Ubuntu image with Nix on it allows for a much lighter image (download time is ridiculous) with zero unneeded package. That's why I'm saying that it can help to track Coq's dependencies better. For instance, when `git`, `rsync` and `which` were added in the test-suite. Not something that can be detected with a full-fledged Ubuntu image...

Zimmi48:
It's funny because I was expecting instead the reaction to be: oh wow, how short and elegant it is to add a NixOS CI test.

Zimmi48:
Here was the request to add this job by @SkySkimmer: https://github.com/coq/coq/pull/7484#issuecomment-388373819

.gitlab-ci.yml
```diff
@@ -233,6 +233,17 @@ windows32:
...
+  image: nixorg/nix:latest # Minimal NixOS image which doesn't even contain git
+  stage: test
+  variables:
+    GIT_STRATEGY: none
+  dependencies: [] # We don't need to download build artifacts
+  before_script: [] # We don't want to use the shared 'before_script'
+  script:
+    - nix build -f "$CI_PROJECT_URL/-/archive/$CI_COMMIT_SHA.tar.gz" --arg doCheck false
+    # We can only test the build and not the test-suite until #5975 is fixed.
```
SkySkimmer: We can install then run the test suite like gitlab does.

Zimmi48: That's not really a supported workflow in Nix. That would be a bit hackish. I prefer to wait and see if Dune is really going to solve this problem. If not, I'll propose to remove the problematic tests.

Zimmi48: Actually I was wrong: this is supported by Nix (https://nixos.org/nixpkgs/manual/#ssec-installCheck-phase). So I'll try that indeed.

SkySkimmer: > The installCheck phase checks whether the package was installed correctly by running its test suite against the installed directories. The default installCheck calls make installcheck. 

We could probably add such a target to the Makefile. It would simplify the gitlab yml too.

Zimmi48: Indeed. And this reinforces my conviction that it is wrong that currently the `check` target fails if configured with `-prefix`. Anyways, for now I'm remapping `installCheck` to use the `check` target and I'm getting pretty close to being able to run the test-suite successfully.

Zimmi48:
In the nixpkgs manual, the `check` target is meant to be run after build but before install, and the `installCheck` target is meant to run after install. This is probably coming from a standard convention. We shouldn't provide a `check` target that fails if it is run before install. We could consider splitting our test-suite in two parts, one that is run by `check` and one that is run by `installCheck` (other solutions are possible).

Zimmi48:
This is ready to merge if CI passes.

default.nix
```diff
@@ -93,8 +93,19 @@ stdenv.mkDerivation rec {
 
   buildFlags = [ "world" ] ++ optional buildDoc "doc-html";
 
-  installTargets = [ "install" ] ++ optional buildDoc "install-doc-html";
+  postBuild = "make byte"; # Needed by the test-suite
+  # But it is not recommended to do it in parallel with make world
```
SkySkimmer: No -j?

SkySkimmer:
Doesn't work: 
~~~
$ nix-build -f "$CI_PROJECT_URL/-/archive/$CI_COMMIT_SHA.tar.gz" -j2 -K
error: unrecognised flag '-f'
~~~

.gitlab-ci.yml
ejgallego: I know who has built this image but I don't see why is this relevant here. My question is still the same, this seems like a non-reproducible setup, doesn't it?

default.nix
```diff
@@ -87,12 +91,25 @@ stdenv.mkDerivation rec {
...
+
+  inherit doInstallCheck;
+
+  preInstallCheck = ''
+    patchShebangs tools/
+    patchShebangs test-suite/
+  '';
 
-  inherit doCheck;
+  installCheckTarget = [ "check" ];
```
SkySkimmer: This includes `validate`, is that intended?
I don't think you can avoid setting the variables that the gitlab test suite job does to make it work with installed Coq, that's why I thought about having a separate installcheck target.

ejgallego:
> Using a NixOS image instead of an Ubuntu image with Nix on it allows for a much lighter image (download time is ridiculous) with zero unneeded package. That's why I'm saying that it can help to track Coq's dependencies better. For instance, when git, rsync and which were added in the test-suite. Not something that can be detected with a full-fledged Ubuntu image...

Ubuntu base image is 80 MiB and contains just very basic packages, all in all I don't see an improvement in download time / image size. In fact, runners will cache the docker image but not the NixOS one so this part is not an improvement.

I made 3 questions above but I still got no answer, so tagging until they get answered.

ejgallego (changes requested):
Answer my questions.

ejgallego:
> For instance, when git, rsync and which were added in the test-suite. Not something that can be detected with a full-fledged Ubuntu image...

`git` and `rsync` are not part of the base Ubuntu image, is that the case for the NixOS image?

ejgallego:
> It's funny because I was expecting instead the reaction to be: oh wow, how short and elegant it is to add a NixOS CI test.

Well my role as a reviewer is to ask very different questions, like "why do we need this?", "is this redudant with the test we already have?", "is the addition of a new job justified?", etc...

ejgallego:
Indeed NixOS is special enough that we may want to run a build job.

aspiwack:
@ejgallego I must admit that I have some doubts that such a trivial PR is worse that much emotion.

Nevertheless, I can provide some answers, I think

- It would be absolutely possible to use a Ubuntu image, install Nix on it, then call the same `default.nix`. It would have the same effect, but it does feel more complex than using this docker image.
- The `latest` tag is, indeed, strictly speaking non-reproducible. But because a specific version of nixpkgs is pinned for the build, the only issue would be if a newer version of nix could no longer read the version of nixpkgs under consideration. I don't know whether this ever happens, but we're talking guaranteed reproducibility for years. I doubt that we need more than that (and as far as reproducibility is concerned, since mirrors can disappear from the internet, it's probably very much theoretical to ask for more than a few years). The rest of the build will be identical, because it's the version of nixpkgs, not of nix, which drives the build. (this is the same reason why running in an Ubuntu image would make no difference)

.gitlab-ci.yml
Zimmi48: I'm unsure what exactly you are asking? Is this because of `latest`? What can I do to make it reproducible? Use a hash instead?

default.nix
Zimmi48: That's a solution indeed, if you are not worried about the build time.

default.nix
Zimmi48: Yes, it does include `validate`. I could restrict to `test-suite` if deemed sufficient.

.gitlab-ci.yml
ejgallego: From my part I can live with it if it is hard to improve, even if there are some components that worry such as openssl or git. WDYT @SkySkimmer ?

.gitlab-ci.yml
ejgallego: So maybe `pkg:nixos` ?

Based on what you say we would for example also add a test to build an opam package, so we would have `pkg:opam` and `pkg:nixos`. How does that sound?

Zimmi48:
> git and rsync are not part of the base Ubuntu image, is that the case for the NixOS image?

I would expect they are either part of the base image, or that they are installed at some point (I haven't checked). Otherwise, how does GitLab pulls the repository?

I thought I had answered your questions but I'm going to try again:

> What part of Coq is this testing that is not tested elsewhere?

Similarly to when we test the compilation and test-suite with several versions of OCaml, this is not testing a different part but rather a different setup. BTW, it would be really easy to test different versions of OCaml with the Nix setup.

> Can you say more?

There is a number of incompatibilities that were fixed over time by either myself or @vbgl. Do you want the list?

> Umm, why our current system does not prevent this?

The `git` case is certainly one that was missed by our current system.

ejgallego:
Thanks for the answers:

> > git and rsync are not part of the base Ubuntu image, is that the case for the NixOS image?
> 
> I would expect they are either part of the base image, or that they are installed at some point (I haven't checked). Otherwise, how does GitLab pulls the repository?

They are not part of the base image [check with `docker run -it ubuntu:bionic /bin/bash`, I think that gitlab pulls the repos from the runner, but I am not sure.

> > What part of Coq is this testing that is not tested elsewhere?
> 
> Similarly to when we test the compilation and test-suite with several versions of OCaml, this is not testing a different part but rather a different setup.

I see, more below.

> BTW, it would be really easy to test different versions of OCaml with the Nix setup.

Well in the current setup it is fairly easy too and in fact it has outperformed Nix in the past as OPAM has official status wrt OCaml. IMHO these kind of statements are a bit "misleading", as they imply NixOS brings and improvement here, but in fact it does not.

> > Can you say more?
> 
> There is a number of incompatibilities that were fixed over time by either myself or @vbgl. Do you want the list?

Indeed sorry as this is really my only question [remarks aside], so far I can see the `/bin/bash` case, what more differences in setup does NixOS help with? I don't need an exhaustive list, just an informal list of things that differ to get an idea.

> > Umm, why our current system does not prevent this?
> 
> The git case is certainly one that was missed by our current system.

`git` may be indeed special, but so far here is another part that where I fail to see an improvement wrt current setup. I could be wrong of course.

Summarizing, and until further correction, so far it seems to me that the only advantage NixOS does bring to our CI setup is to test against hardcoded paths.

That doesn't seem a lot, however, on the other hand, if we see NixOS as a packaging job indeed it does make sense to test the build script `default.nix` as it would make sense to test the opam files once they are auto generated, so it seems OK to have this job in the CI.

SkySkimmer:
Probably git is pulled in by opam.

SkySkimmer:
https://gitlab.com/SkySkimmer/coq/-/jobs/66654642
>Selecting previously unselected package git.
Preparing to unpack .../226-git_1%3a2.17.0-1ubuntu1_amd64.deb ...
Unpacking git (1:2.17.0-1ubuntu1) ...

ejgallego:
> Probably git is pulled in by opam.

It should not, but indeed we have a bug in our Dockerfile, we should run apt with with `--no-install-recommends` as to really test strict dependencies.

gares:
> And this reinforces my conviction that it is wrong that currently the |check| target fails if configured with |-prefix|.

@SkySkimmer please have a look at #7977

ejgallego (approved):
I would like to learn more about what is NixOS testing that the standard CI run does not do; apart from path hardcoding which is IMO useful.

Anyways, if we see this as a packaging job then indeed it does make sense to test our `default.nix` job, so green light here.

Note that we should qualify the job as to be `pkg:nixos` or something like that for when we introduce `pkg:opam`.

Zimmi48:
FWIW pulling the Docker image takes less than 20 seconds (I've observed the build in real time - and GitLab helps by showing the time elapsed since the beginning).

> It should not, but indeed we have a bug in our Dockerfile, we should run apt with with --no-install-recommends as to really test strict dependencies.

Well, if you don't install `git` you have to change the "git strategy" as I did in the nixos job.

> Well in the current setup it is fairly easy too and in fact it has outperformed Nix in the past as OPAM has official status wrt OCaml. IMHO these kind of statements are a bit "misleading", as they imply NixOS brings and improvement here, but in fact it does not.

Don't worry I'm planning to show that Nix can outperform OPAM soon XD

> Indeed sorry as this is really my only question [remarks aside], so far I can see the /bin/bash case, what more differences in setup does NixOS help with? I don't need an exhaustive list, just an informal list of things that differ to get an idea.

So far, mostly this and the dependency on Python 2 that had gone unnoticed.

> Anyways, if we see this as a packaging job then indeed it does make sense to test our default.nix job, so green light here.

This is indeed how I see it.

> Note that we should qualify the job as to be pkg:nixos or something like that for when we introduce pkg:opam.

Agreed. I'll put `pkg:nix`. I prefer this to `distro:nixos`.

ejgallego:
> Agreed. I'll put pkg:nix. I prefer this to distro:nixos.

Sounds good, we'll add opam and likely `deb` at some point.

ejgallego:
Note that still there are some naming discrepancies [in the commit msg] wrt to Nix vs NixOS

Zimmi48:
This is blocked by a consistent failure of the timing test. Here's the complete log: [timing.log](https://github.com/coq/coq/files/2156404/timing.log). The most interesting part being:

```
--- time-of-build-before.log.desired.processed	2018-07-02 18:29:54.820287819 +0000
+++ time-of-build-before.log.processed	2018-07-02 18:29:54.818287815 +0000
@@ -1,5 +1,11 @@
 COQDEP VFILES
 COQC Slowv
-Slow (real: , user: , sys: , mem: ko)
+------
+real	ms
+user	ms
+sys	ms
 COQC Fastv
-Fast (real: , user: , sys: , mem: ko)
+------
+real	ms
+user	ms
+sys	ms
```

It is working inside `nix-shell` though. I've tried adding `time` as an explicit dependency but this didn't change anything. @vbgl Could it be that the time command doesn't work in Nix builds?

In the meantime, I'm trying to understand why I don't manage to upload artifacts for failed test logs: https://gitlab.com/coq/coq/-/jobs/78880289
@SkySkimmer Any advice?

SkySkimmer:
https://docs.gitlab.com/ee/ci/yaml/#artifacts-paths
>You can only use paths that are within the project workspace.

So /tmp/ stuff won't work. Maybe if you do `ln -s tmp /tmp/` you can use a relative `tmp/...` path? Try and see.

Zimmi48:
Thanks! The error message wasn't very explicit ;-)
Anyways, instead of the hack you suggested, I've just told Nix to use the current directory as the `tmp` directory.

Zimmi48:
The artifact uploading works now. So the only two blocking things are to figure out what is going on with the timing test, and to decide whether we keep things as they are now or go back to doing the check before install thanks to #7977.

Zimmi48:
So in fact, the explanation is not that `time` doesn't work on Nix builds but that it outputs a different format when called as `time` or or as `/path/to/time`. Why? Because when calling `time`, it is a shell function that we are calling (https://askubuntu.com/a/434294/769627).

gares:
Yep, time is a bash shell builtin, if we rely on the `gnu time` utility we should call it via `/usr/bin/time`.

Zimmi48:
Alright this PR should be really ready now. I have fixed the timing tools on NixOS.
I have tried taking advantage of #7977 but I haven't managed to, so I'm going to keep things as they are now.

FTR the errors I got were:
```
Error: Cannot find a physical path bound to logical path matching suffix <> and prefix Coq.
```

@vbgl Could you review the changes to `default.nix`?

gares:
> I have tried taking advantage of #7977 but I haven't managed to, so I'm going to keep things as they are now.

Can you tell more? What gave this error?

SkySkimmer:
>FTR the errors I got were:

Without knowing what you did to adapt it's not really possible to investigate.

SkySkimmer (changes requested):

Zimmi48:
> Can you tell more? What gave this error?

I added this line:

```
checkFlags = [ "CAML_LD_LIBRARY_PATH=$(PWD)/kernel/byterun" "COQLIB=$(PWD)" ];
```

But looking at some other parts of the log, I can see:

```
Fatal error: exception Sys_error("/build/coq/test-suite/coq-makefile/arg/_test/plugins: No such file or directory")
```
I might have made wrong assumptions on the value of `$PWD` at check time. I don't know what other variable I could have used. Maybe @vbgl knows.

SkySkimmer:
If I understand correctly that makes it run `make var=foo check` but then `var` is not exported, it's only set in the makefile. @gares was using `var=foo make check`.

Zimmi48:
Ah, I didn't know the difference. I'm going to try again.

Zimmi48:
In this branch: https://github.com/Zimmi48/coq/tree/save/ci/gitlab/nixos-2
I still get this error:

```
Makefile:423: warning: undefined variable 'TGTS'
Makefile:655: warning: undefined variable 'src/test_plugin_MLPACK_DEPENDENCIES'
Fatal error: cannot load shared library dllnums
Reason: dllnums.so: cannot open shared object file: No such file or directory
make[5]: *** [Makefile:668: theories/test.vo] Error 2
make[4]: *** [Makefile:329: all] Error 2
make[4]: Leaving directory '/build/coq/test-suite/coq-makefile/findlib-package/_test'
make[3]: *** [Makefile:752: byte] Error 2
make[3]: Leaving directory '/build/coq/test-suite/coq-makefile/findlib-package/_test'
    coq-makefile/findlib-package/run.sh...Error!
```

gares:
This is what `CAML_LD_LIBRARY_PATH=$(PWD)/kernel/byterun` is supposed to fix

gares:
Oh no sorry, it is not the vm, it is the .so part of bignum

Zimmi48:
Are you sure it's not the num library from OCaml? In fact, I'm guessing that Nix had already set `CAML_LD_LIBRARY_PATH` and I overrode it. I should have appended to it.

gares:
Yes, you should append the path leading to byterun

gares:
Anyway, I think the nix way is a bit stretching it in this case.

The manpage of `ocamlrun` says that `CAML_LD_LIBRARY_PATH` is for *additional* paths to be scanned for .so objects...  and this is in line with that [LD_LIBRARY_PATH is for](http://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html): hacks and hijacking, not regular function.

I understand why they did it this way, but they should have asked the upstreams for for another variable name to avoid backstabbing their users: setting `(CAML_)LD_LIBRARY_PATH` should not hide stuff that is installed system wide, IMHO. [troll] What is next? Setting LD_PRELOAD system wide?

gares (changes requested):
The `which time` part should be refined as follows:
```
gares@gargamel:~$ export PATH=/bin
gares@gargamel:~$ which time
gares@gargamel:~$ echo $?
1
gares@gargamel:~$ which time || echo time
time
gares@gargamel:~$ export PATH=/bin:/usr/bin/
gares@gargamel:~$ which time || echo time
/usr/bin//time
```

default.nix
vbgl: I guess we discussed that already; and we were in favor of not pinning within `default.nix` (which is used daily by developers). If you really need pinning, please do so in a separate file that has to be explicitly called; in the CI script, you may call `nix-build -f reproducible.nix …`, but when working as usual, I’ll only run plain `nix-shell`.

Zimmi48:
I'm almost back to where I was with `installCheck`, only the timing test is still failing, this time with:

```
Makefile:71: warning: undefined variable '*'

--- time-of-build-before.log.desired.processed	2018-07-03 14:07:28.134304031 +0000
+++ time-of-build-before.log.processed	2018-07-03 14:07:28.133304032 +0000
@@ -1,5 +1,2 @@
-COQDEP VFILES
-COQC Slowv
 Slow (real: , user: , sys: , mem: ko)
-COQC Fastv
 Fast (real: , user: , sys: , mem: ko)
```

Zimmi48:
@gares I don't think you have reviewed the last version of the PR. There is no `which time` anymore. I'm using `command time` as was requested by @SkySkimmer.

aspiwack:
Is testing the test-suite that useful in this particular job?

SkySkimmer:
Is that undefined variable expected?

default.nix
```diff
@@ -36,64 +40,50 @@ stdenv.mkDerivation rec {
...
-    num
-
-  ]) ++ (if buildIde then [
-
-    # CoqIDE dependencies
-    ocamlPackages.lablgtk
-
-  ] else []) ++ (if buildDoc then [
-
+    python2 time # coq-makefile timing tools
```
vbgl: Why do we need `python2`? Can we contemplate porting that to python 3?

(I’m looking at the `python3` a few lines below…)

Zimmi48:
> Is that undefined variable expected?

I would guess not.

gares (approved):

vbgl (changes requested):

Zimmi48:
> Is testing the test-suite that useful in this particular job?

Yes, I wouldn't have discovered that the timing tools were broken on NixOS if I hadn't try to have the test-suite pass. Anyways, we are nearly there now and we have two options: find out what's going on with the above mentioned error of the timing test and this undefined variable, or merge the current version which is using the `installCheck` phase and which passed CI.

aspiwack:
> we have two options: find out what's going on with the above mentioned error of the timing test and this undefined variable, or merge the current version which is using the installCheck phase and which passed CI.

Merge first, improve later. I see no reason of delaying this.

Zimmi48:
> I guess we discussed that already; and we were in favor of not pinning within default.nix (which is used daily by developers). If you really need pinning, please do so in a separate file that has to be explicitly called; in the CI script, you may call nix-build -f reproducible.nix …, but when working as usual, I’ll only run plain nix-shell.

That's not exactly how I reminded the conclusion. In my mind, the conclusion was: we are not going to keep pinning now for something that will happen in an hypothetical future. And the hypothetical future is now.
I don't understand why we would not like to pin by default given that `default.nix` provides the option to override this default. Here are my reasons:

1. We are testing the package that we are providing (and that people can install by doing `nix-env -if https://github.com/coq/coq/archive/master.zip`).
2. If we are not pinning the version that we are using, we'll forget to update it.
3. What's the problem with using a pinned version for `nix-shell` anyways?
4. And if you really don't want to use the pin version and that typing `nix-shell --arg pkgs "import <nixpkgs> {}"` is too hard, you can set up a bash alias `coq-dev: nix-shell --arg pkgs "import <nixpkgs> {}`. FTR my personal default nixpkgs channel is not nixpkgs-unstable and is too old for Coq development so I have to do something like this already and it's not a big deal.

> Merge first, improve later. I see no reason of delaying this.

I usually agree but this PR is changing `doCheck` into `doInstallCheck`. If the plan is to go back to `doCheck` immediately after, I prefer to do it at once. That being said, I have no idea how to fix the last issue I reported, so I'm proposing to merge as-is unless someone has an idea.

maximedenes:
There are many things I don't follow here, and I don't have time answer immediately, please don't pin again by default meanwhile.

vbgl:
Can we pin in default.nix and un-pin in shell.nix? So that there are good defaults both for developers and for end-users?

maximedenes:
> Can we pin in default.nix and un-pin in shell.nix? So that there are good defaults both for developers and for end-users?

That would work for me.

Zimmi48:
> Can we pin in default.nix and un-pin in shell.nix? So that there are good defaults both for developers and for end-users?

Yes and that would be fine to me as a way to make you and @maximedenes happy. So I'm going to push a commit doing this even if I still fail to see why unpinning does to help developers.

vbgl:
I think that good defaults for developers **is** important.

That being said, I’m not completely convinced that what I just suggested completely solves the issue of pinning. Can you please clarify how/when/why/… will the reference to `nixpkgs` be updated?

Zimmi48:
> I think that good defaults for developers is important.

I agree. I just disagree on what these good defaults are.

> That  being said, I’m not completely convinced that what I just suggested  completely solves the issue of pinning. Can you please clarify  how/when/why/… will the reference to nixpkgs be updated?

Don't worry, I'll personally use the pinned version from `default.nix` instead of the unpinned version from `shell.nix` so I'll update the pin as often as I update my nixpkgs-unstable channel (every other month or so).

vbgl:
Of course, if the pinned version always matches your default version, you won’t understand the issue…

Zimmi48:
So @SkySkimmer @vbgl are we gonna merge this?

SkySkimmer:
Why is it building camlp5, is there not a prebuilt version?

default.nix
```diff
@@ -36,64 +40,50 @@ stdenv.mkDerivation rec {
...
     else
       with builtins; filterSource
-        (path: _: !elem (baseNameOf path) [".git" "result" "bin"]) ./.;
+        (path: _:
+           !elem (baseNameOf path) [".git" "result" "bin" "_build_ci"]) ./.;
 
   prefixKey = "-prefix ";
 
-  buildFlags = [ "world" ] ++ optional buildDoc "doc-html";
+  buildFlags = [ "world" "byte" ] ++ optional buildDoc "doc-html";
```
SkySkimmer: Is building byte simultaneously OK?

SkySkimmer (approved):
The log is very bloated by nix complaining about "wrong elf type".

Zimmi48:
> Why is it building camlp5, is there not a prebuilt version?

I guess because we are not using the default OCaml version from nixpkgs. Anyways my next PR will add caching.

Zimmi48:
Only CI failure is mtac2, which is broken with current master.

Zimmi48:
@vbgl I've addressed your request. Can you approve the PR?

vbgl (approved):

vbgl:
Théo, you haven’t answered my questions: why do we pin the reference to nixpkgs? to which version? when do we update and how?

I nonetheless recommend merging.

Zimmi48:
> Théo, you haven’t answered my questions: why do we pin the reference to nixpkgs? to which version? when do we update and how?

Oh! That's the second time in this PR that I think I have answered the questions and the reviewer says I haven't :scream: 
Why pin? For reproducibility. In fact, the NixOS image I use doesn't even contains a nixpkgs channel and encourages its users to pin (https://hub.docker.com/r/nixorg/nix/)

Which version? The latest version of nixpkgs when the pin is updated.

When update? As often as we want to. In the beginning, I'll update by hand. When I'll proceed to implement the proposal #6724, I'll include it in the regular (weekly) automatic update.

Zimmi48:
@SkySkimmer Can you merge?

Zimmi48:
FTR here is the diff to go back to check before install which almost works except for the problem with the timing test mentioned above:

```diff
diff --git a/default.nix b/default.nix
index 1be274081..c76526266 100644
--- a/default.nix
+++ b/default.nix
@@ -16,7 +16,7 @@
 # $ nix-build
 # at the root of the Coq repository.
 # nix-build also supports the --arg option, so you will be able to do:
-# $ nix-build --arg doInstallCheck false
+# $ nix-build --arg doCheck false
 # if you want to speed up things by not running the test-suite.
 # Once the build is finished, you will find, in the current directory,
 # a symlink to where Coq was installed.
@@ -27,9 +27,10 @@
       sha256 = "1lzvp3md0hf6kp2bvc6dbzh40navlyd51qlns9wbkz6lqk3lgf6j";
     }) {})
 , ocamlPackages ? pkgs.ocaml-ng.ocamlPackages_4_06
+, buildByte ? true
 , buildIde ? true
 , buildDoc ? true
-, doInstallCheck ? true
+, doCheck ? true
 }:
 
 with pkgs;
@@ -46,14 +47,15 @@ stdenv.mkDerivation rec {
   ++ (with ocamlPackages; [ ocaml findlib camlp5_strict num ])
   ++ optional buildIde ocamlPackages.lablgtk
   ++ optionals buildDoc [
+
     # Sphinx doc dependencies
     pkgconfig (python3.withPackages
       (ps: [ ps.sphinx ps.sphinx_rtd_theme ps.pexpect ps.beautifulsoup4
              ps.antlr4-python3-runtime ps.sphinxcontrib-bibtex ]))
     antlr4
   ]
-  ++ optionals doInstallCheck (
-    # Test-suite dependencies
+  ++ optionals doCheck ( # Test-suite dependencies
+
     # ncurses is required to build an OCaml REPL
     optional (!versionAtLeast ocaml.version "4.07") ncurses
     ++ [ ocamlPackages.ounit rsync which ]
@@ -72,18 +74,19 @@ stdenv.mkDerivation rec {
 
   prefixKey = "-prefix ";
 
-  buildFlags = [ "world" "byte" ] ++ optional buildDoc "doc-html";
+  buildFlags =
+   [ "world" ] ++ optional buildByte "byte" ++ optional buildDoc "doc-html";
 
   installTargets =
-    [ "install" "install-byte" ] ++ optional buildDoc "install-doc-html";
+    [ "install"] ++ optional buildByte "install-byte"
+    ++ optional buildDoc "install-doc-html";
 
-  inherit doInstallCheck;
+  inherit doCheck;
 
-  preInstallCheck = ''
+  preCheck = ''
     patchShebangs tools/
     patchShebangs test-suite/
+    export CAML_LD_LIBRARY_PATH=$PWD/kernel/byterun:$CAML_LD_LIBRARY_PATH
+    export COQLIB=$PWD
   '';
-
-  installCheckTarget = [ "check" ];
-
 }
```

PR #7977, opened 2018-07-02
allow `make check` to succeed when -prefix is given to ./configure and make install not run yet

gares:
In others PR I read that @SkySkimmer wants to be able to run make check without having installed Coq but having configured Coq for system-wide installation (eg `./configure -prefix /usr`). Well, here it is.

On my PC this patch is sufficient to configure Coq in system wide mode and run
```
CAML_LD_LIBRARY_PATH=$PWD/kernel/byterun COQLIB=$PWD make -j2 -C test-suite
```
before installing Coq.

### DIFF
--- a/lib/envars.ml
+++ b/lib/envars.ml
@@ -110,6 +110,7 @@ let check_file_else ~dir ~file oth =
   if Sys.file_exists (path / file) then path else oth ()
 
 let guess_coqlib fail =
+  getenv_else "COQLIB" (fun () ->
   let prelude = "theories/Init/Prelude.vo" in
   check_file_else ~dir:Coq_config.coqlibsuffix ~file:prelude
     (fun () ->
@@ -117,6 +118,7 @@ let guess_coqlib fail =
       then Coq_config.coqlib
       else
         fail "cannot guess a path for Coq libraries; please use -coqlib option")
+  )
 
 (** coqlib is now computed once during coqtop initialization *)
 


gares:
In others PR I read that @SkySkimmer wants to be able to run `make check` without having installed Coq but having configure Coq for system-wide installation. Well, here it is.

FTR: `CAML_LD_LIBRARY_PATH` is only needed for byte code (eg coqtop.byte) and there is a coq_makefile tests running `make byte`.

ejgallego (approved):
I am OK with this, note that I have been trying to rewrite this function for a while [to use coqfind].

gares:
@ejgallego I don't know how to test it in our CI, help is welcome. Or maybe we can ask @SkySkimmer to confirm this is what he wants and let him add `make check` here and there on CI

SkySkimmer:
That's easy, just combine it with the nix CI PR.

Gaëtan Gilbert

SkySkimmer:
>On my PC this patch is sufficient to configure Coq in system wide mode and run

Maybe those variables should be set automatically in the makefiles?

Zimmi48:
Please let me know if you want me to rebase the Nix CI PR on top of this.

gares:
> Maybe those variables should be set automatically in the makefiles?

You may provide a `make check LOCAL=1` in order to force the two variables. I'm not doing it in this PR.



> Please let me know if you want me to rebase the Nix CI PR on top of this.

Do as you like, but I'm not recommending making PR depending on PR ;-P

gares:
CI failures seem spurious (ltac2, that also fails windows), while FCF fails with code 137 and no error message

PR #7979, opened 2018-07-02
TACTIC EXTEND in coqpp

ppedrot:
This is a followup of #7902. It allows `coqpp` to also handle `TACTIC EXTEND` macros.

(Minor) syntax changes have been documented in `dev/` and as many files as possible have been ported.

### DIFF
--- a/dev/doc/changes.md
+++ b/dev/doc/changes.md
@@ -95,19 +95,73 @@ Primitive number parsers
 
 ### Transitioning away from Camlp5
 
-In an effort to reduce dependency on Camlp5, the use of Camlp5 GEXTEND macro
-is discouraged. Most plugin writers do not need this low-level interface, but
-for those who do, the transition path is somewhat straightforward. To convert
-a ml4 file containing only standard OCaml with GEXTEND statements, the following
-should be enough:
-- replace its "ml4" extension with "mlg"
-- replace GEXTEND with GRAMMAR EXTEND
-- wrap every occurrence of OCaml code into braces { }
+In an effort to reduce dependency on camlp5, the use of several grammar macros
+is discouraged. Coq is now shipped with its own preprocessor, called coqpp,
+which serves the same purpose as camlp5.
+
+To perform the transition to coqpp macros, one first needs to change the
+extension of a macro file from `.ml4` to `.mlg`. Not all camlp5 macros are
+handled yet.
+
+Due to parsing constraints, the syntax of the macros is slightly different, but
+updating the source code is mostly a matter of straightforward
+search-and-replace. The main differences are summarized below.
+
+#### OCaml code
+
+Every piece of toplevel OCaml code needs to be wrapped into braces.
 
 For instance, code of the form
 ```
 let myval = 0
+```
+should be turned into
+```
+{
+let myval = 0
+}
+```
 
+#### TACTIC EXTEND
+
+Steps to perform:
+- replace the brackets enclosing OCaml code in actions with braces
+- if not there yet, add a leading `|̀  to the first rule
+
+For instance, code of the form:
+```
+TACTIC EXTEND my_tac
+  [ "tac1" int_or_var(i) tactic(t) ] -> [ mytac1 ist i t ]
+| [ "tac2" tactic(t) ] -> [ mytac2 t ]
+END
+```
+should be turned into
+```
+TACTIC EXTEND my_tac
+| [ "tac1" int_or_var(i) tactic(t) ] -> { mytac1 ist i t }
+| [ "tac2" tactic(t) ] -> { mytac2 t }
+END
+```
+
+#### VERNAC EXTEND
+
+Not handled yet.
+
+#### ARGUMENT EXTEND
+
+Not handled yet.
+
+#### GEXTEND
+
+Most plugin writers do not need this low-level interface, but for the sake of
+completeness we document it.
+
+Steps to perform are:
+- replace GEXTEND with GRAMMAR EXTEND
+- wrap every occurrence of OCaml code in actions into braces { }
+
+For instance, code of the form
+```
 GEXTEND Gram
    GLOBAL: my_entry;
 
@@ -119,10 +173,6 @@ END
 ```
 should be turned into
 ```
-{
-let myval = 0
-}
-
 GRAMMAR EXTEND Gram
    GLOBAL: my_entry;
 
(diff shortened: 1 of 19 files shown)


grammar/coqpp_lex.mll
```diff
@@ -105,6 +108,7 @@ rule extend = parse
 (** Standard *)
 | ident { IDENT (Lexing.lexeme lexbuf) }
 | qualid { QUALID (Lexing.lexeme lexbuf) }
+| number { INT (int_of_string (Lexing.lexeme lexbuf)) }
```
ejgallego: that looks very very fishy.

grammar/coqpp_main.ml
```diff
@@ -260,22 +262,77 @@ let print_ast chan ext =
...
 
+module TacticExt :
+sig
+
+val print_ast : Format.formatter -> tactic_ext -> unit
+
+end =
+struct
+
+let rec print_symbol chan = function
```
ejgallego: `chan` here is misleading, it should be `fmt`

grammar/coqpp_lex.mll
ppedrot: ?

grammar/coqpp_main.ml
```diff
@@ -260,22 +262,77 @@ let print_ast chan ext =
...
+let print_rules chan rules =
+  fprintf chan "Tacentries.([@[<v>%a@]])" print_rules rules
+
+let print_ast chan ext =
+  let pr chan () =
+    let level = match ext.tacext_level with None -> 0 | Some i -> i in
+    fprintf chan "Tacentries.tactic_extend %s \"%s\" ~level:%i %a"
+      plugin_name ext.tacext_name level print_rules ext.tacext_rules
+  in
+  let () = fprintf chan "let () = @[%a@]\n" pr () in
```
ejgallego: `\n` -> `@\n`

grammar/coqpp_main.ml
ppedrot: I used chan all over the place already, this is just me being consistent with the code.

grammar/coqpp_main.ml
ejgallego: This makes discovering badly balanced boxed easier.

grammar/coqpp_parse.mly
```diff
@@ -10,13 +10,59 @@
...
+    check_separator sep;
+    Uopt entry
+  else if l = 7 && coincide s "tactic" 0 && '5' >= s.[6] && s.[6] >= '0' then
+    let n = Char.code s.[6] - 48 in
+    check_separator sep;
+    Uentryl ("tactic", n)
+  else
+    let s = match s with "hyp" -> "var" | _ -> s in
+    check_separator sep;
+    Uentry s
```
ejgallego: Bufffff, this looks like quite an error-prone mess; a table + generic routine should be preferred.

grammar/coqpp_main.ml
ppedrot: The generated code is horrible anyways, I don't think it is really important. Will fix it though.

ejgallego (commented):
Minor comments, in general it looks OK to me.

grammar/coqpp_parse.mly
ppedrot: That's historical stuff that ought to be removed, and this is just copy-pasting from the corresponding camlp5 file. I know this should be shared, but this would introduce useless dependencies between the camlp5 macros and coqpp. So I went for the easy path, thinking we can get rid of camlp5 soon enough.

grammar/coqpp_main.ml
ejgallego: Yeah so maybe a quick search/replace would help; this will greatly mislead people IMVHO.

grammar/coqpp_parse.mly
ppedrot: I can probably seize the opportunity to get rid of this special case though.

grammar/coqpp_lex.mll
ejgallego: Oh never mind...

grammar/coqpp_main.ml
ppedrot: Greatly? Don't you think you're exaggerating a bit? But yeah, let's search-and-replace.

grammar/coqpp_parse.mly
ejgallego: I also dislike all the hardcoded numbers + indexes; I don't think cleaning this up is a hard requirement but we are introducing very sketchy code in that routine so I'd be good to lessen the technical debt if we can.

ppedrot: Ultimately and hopefully all this mess will be replace by Ltac2 that has a sane compositional system for notation scopes.

grammar/coqpp_main.ml
ppedrot: That's really changing like every line in `coqpp_main`. I don't think it's very good from the point of view of readability of the diff. Should I do that in another commit at least?

dev/doc/changes.md
```diff
@@ -133,9 +183,12 @@ my_entry:
 END
 ```
 
-Currently mlg files can only contain GRAMMAR EXTEND statements. They do not
-support TACTIC, VERNAC nor ARGUMENT EXTEND. In case you have a file containing
-both kinds at the same time, it is recommended splitting it in two.
+Caveats:
+- No ̀`GLOBAL` entries mean that they are all local, while camlp5 special-cases
```
Zimmi48: You have a weird backtick here.

dev/doc/changes.md
```diff
@@ -133,9 +183,12 @@ my_entry:
...
+  this as a shorthand for all global entries. Solution: always define a GLOBAL
```
Zimmi48: Maybe you want to put this GLOBAL in verbatim as well.

ppedrot:
Comments above addressed.

grammar/coqpp_main.ml
ejgallego: Oh sure do as you wish, if you think this is bad we can may don't do it; but I see the confusion between the two `fprintf` often enough as for this to worry me.

maximedenes:
> While you are at it please add grammar/coqpp to .gitignore; note that I am not fond of randomly placing binaries around the tree...

Why do we put the binary there? It should be placed alongside other binaries IMHO. Our build is already non-hygienic enough...

ppedrot:
@maximedenes I am planning to write a PR to bundle coqpp with the other Coq binaries, so that external plugins can use it, but I need a little bit of help with `coq_makefile`.

ppedrot:
(In particular this means that this PR is ready for me, up to fixes asked by other reviewers.)

ppedrot:
See #7984 for a proper handling of the `coqpp` binary.

ppedrot:
@ejgallego do you want to assign this PR to yourself?

ejgallego:
Letting a couple of days for comments.

ejgallego (approved):

PR #7109, opened 2018-03-29
Term combinators respecting the canonical structure of branches and return predicate

herbelin:
**Kind:** architecture (on top of #7080)

This is a new step in the direction of CEP https://github.com/coq/ceps/pull/34.

This commit proposes some combinators which seem appropriate in the direction of seeing the spine of lambda and let-ins in the branches and return predicate of a `Case` node (i.e. Coq's `match`) as an integral part of the structure of a `Case`.

The current proposal is certainly not the final version for such combinators yet, but feedback is needed so I'm submitting them as they are now. In particular, I have the following questions:

- Do we continue to replicate `Constr` to `Term`, or do we progressively deprecate `Term`?
- Do we try to have a minimal number of functions in `Term` or do we prefer
  to share code between kernel (`Term`/`Constr`) and higher levels (`Termops`/`EConstr`)? For instance, do we share the code of `EConstr.map` and `Constr.map` (which is  not the case currently)?
- Also, do we put (effective code of) combinators in `EConstr` or in `Termops`, or do we put them in `EConstr` and replicate them in `Termops`?

PS: I introduced a `userview` which changes the semantics of what is a subterm of a `match`. For instance, the body of the return predicate, the body of a branch, the types and let's in the context of a return predicate or branch are now subterms of a `match`. Also, as a consequence, there is no more need for β for contracting the branches of a `Case` after a ɩ: this shall now be part of ɩ. On the other side a branch of the form `Lambda (name, type_of_pattern_variable, body)` is not anymore a subterm (on which `pattern` would apply). The use of a `userview` is transitional. The idea is that eventually only the `userview` is used.

PR #7983, opened 2018-07-03
Turn a dead branch into an assertion failure in VM reification.

ppedrot:
In #7607, dead code that used to handle non-dependent return predicates was removed. This made the reification branch expecting non-functions in predicates dead code. We fix this by using an assert instead.

### DIFF
--- a/pretyping/vnorm.ml
+++ b/pretyping/vnorm.ml
@@ -311,7 +311,7 @@ and nf_predicate env sigma ind mip params v pT =
       let dom = mkApp(mkIndU ind,Array.append params rargs) in
       let body = nf_vtype (push_rel (LocalAssum (name,dom)) env) sigma vb in
       mkLambda(name,dom,body)
-    | _ -> nf_val env sigma v crazy_type
+    | _ -> assert false
 
 and nf_args env sigma vargs ?from:(f=0) t =
   let t = ref t in


JasonGross:
Is this actually dead code?  E.g., if I use `Derive` / `eexists `, and then use `change` or `set ` to turn a match return predicate into something that is not judgmentally a function, and then fill the evar with that, will vm_computing the resulting term hit this anomaly?

ppedrot:
@JasonGross It better be, otherwise #7607 is incorrect. I think that the code base (including the pretyper) enforces that case statements always statically have the right shape, even if the return type is an evar. This is what #7109 is enforcing by typing.

JasonGross:
@ppedrot This code currently gives "Error: Anomaly "Not enough lambda/let's." Please report at http://coq.inria.fr/bugs/." at `Defined` in 8.8; does it work the same way in this branch?
```coq
Definition P (x : nat) := match x with O => unit | S _ => nat end.
Definition foo (x : nat) : { y : P x | y = match x return P x with O => tt | S n' => n' end }.
Proof.
  change (fun x => P x) with P.
  vm_compute.
  eexists.
  reflexivity.
Defined.
```

ppedrot:
@JasonGross Same anomaly.

maximedenes (approved):
LGTM, thanks!

PR #7985, opened 2018-07-03
Remove letouzey from CODEOWNERS since he left the Coq organization.

SkySkimmer:
This PR assumes that secondary maintainers are willing to become primary, concerned are @maximedenes @herbelin @ppedrot @mattam82 @gares @ejgallego 
Let's give them time to react if that's not accurate.

It also leaves omega/romega without any maintainers.

We should find secondary maintainers for those components but IMO there's no point delaying this PR until it can be done.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -137,8 +137,7 @@
 /plugins/derive/        @aspiwack
 # Secondary maintainer @ppedrot
 
-/plugins/extraction/    @letouzey
-# Secondary maintainer @maximedenes
+/plugins/extraction/    @maximedenes
 
 /plugins/firstorder/   @PierreCorbineau
 # Secondary maintainer @herbelin
@@ -158,10 +157,6 @@
 
 /plugins/nsatz/         @thery
 # Secondary maintainer @ppedrot
-
-/plugins/omega/         @letouzey
-
-/plugins/romega/        @letouzey
 
 /plugins/setoid_ring/   @amahboubi
 # Secondary maintainer @bgregoir
@@ -219,44 +214,34 @@
 
 ########## Standard library ##########
 
-/theories/Arith/          @letouzey
-# Secondary maintainer @herbelin
+/theories/Arith/          @herbelin
 
-/theories/Bool/           @letouzey
-# Secondary maintainer @herbelin
+/theories/Bool/           @herbelin
 
 /theories/Classes/        @mattam82
 # Secondary maintainer @herbelin
 
-/theories/FSets/          @letouzey
-# Secondary maintainer @herbelin
+/theories/FSets/          @herbelin
 
-/theories/Init/           @letouzey
-# Secondary maintainer @ppedrot
+/theories/Init/           @ppedrot
 
-/theories/Lists/          @letouzey
-# Secondary maintainer @ppedrot
+/theories/Lists/          @ppedrot
 
 /theories/Logic/          @herbelin
 # Secondary maintainer @ppedrot
 
-/theories/MSets/          @letouzey
-# Secondary maintainer @herbelin
+/theories/MSets/          @herbelin
 
-/theories/NArith/         @letouzey
-# Secondary maintainer @herbelin
+/theories/NArith/         @herbelin
 
-/theories/Numbers/        @letouzey
-# Secondary maintainer @herbelin
+/theories/Numbers/        @herbelin
 
-/theories/PArith/         @letouzey
-# Secondary maintainer @herbelin
+/theories/PArith/         @herbelin
 
 /theories/Program/        @mattam82
 # Secondary maintainer @herbelin
 
-/theories/QArith/         @letouzey
-# Secondary maintainer @herbelin
+/theories/QArith/         @herbelin
 
 /theories/Reals/          @silene
 # Secondary maintainer @ppedrot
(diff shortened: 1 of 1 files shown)


ejgallego (approved):
I am OK.

ppedrot (approved):

maximedenes (changes requested):
I can't be the primary maintainer for extraction. There are many bugs to address and I don't have enough time at the moment. There should either be a real primary maintainer, or we should declare it unmaintained.

Zimmi48:
Primary maintainers are about merging PRs, not fixing bugs though.

maximedenes:
> Primary maintainers are about merging PRs, not fixing bugs though.

It may be the current status quo (although I don't think that's really the case), but I think we should aim for a workflow where maintainers maintain (and also merge PRs).

.github/CODEOWNERS
```diff
@@ -159,10 +158,6 @@
 /plugins/nsatz/         @thery
 # Secondary maintainer @ppedrot
 
-/plugins/omega/         @letouzey
-
-/plugins/romega/        @letouzey
```
ejgallego: Can't you keep this and use `@nobody`?

Zimmi48: `@nobody` is a real person though so you probably meant `@ghost`.

Zimmi48:
> It may be the current status quo (although I don't think that's really the case), but I think we should aim for a workflow where maintainers maintain (and also merge PRs).

That's not the deal that was made when people agreed to be code owner. So we can change our terminology to distinguish code owner and maintainer if you want, but I don't think we should aim for such a workflow, at least not in the medium-term.

Zimmi48:
So you may refuse to become the primary code owner, but you can't use this argument as it sends a bad signal to other people already or about to become code owners.

.github/CODEOWNERS
gares: The other day I discovered you can cite "teams", eg `@coq/pushers`, so maybe this is what we should do...

gares:
> we can change our terminology to distinguish code owner and maintainer if you want

The confusion pissed me off already quite a few times, and Maxime is aware of that.

For example I'm "maintaining" CoqIDE that I think should have never existed in the first place (why on earth would one write a UI in OCaml?). I can surely review and merge patches on CoqIDE, since I know its code, but expecting me to fix bugs that are not super critical is stretching my involvement too much. 

I think we should at least clarify it at the next WG. Having two roles may confuse users, that refer to maintainers as "people that may know", according to my experience, not people that are expected to do.

SkySkimmer:
>I can't be the primary maintainer for extraction. There are many bugs to address and I don't have enough time at the moment. There should either be a real primary maintainer, or we should declare it unmaintained.

Removed.

I also added a default entry `/* @coq/pushers`, no clue if it will work.

Zimmi48:
> I also added a default entry /* @coq/pushers, no clue if it will work.

I would do some experimentation elsewhere before doing so. I'm afraid that it means it is going to ask a review from everyone in the team everytime a file without code owner is modified (for instance every time `CHANGES` is modified).

SkySkimmer:
Here's a manual review request for pushers, any effect on your side?

Zimmi48:
Yes, similar to the one I get when I am personally requested a review.

gares (approved):

maximedenes:
> The confusion pissed me off already quite a few times, and Maxime is aware of that.
> 
> For example I'm "maintaining" CoqIDE that I think should have never existed in the first place (why on earth would one write a UI in OCaml?). I can surely review and merge patches on CoqIDE, since I know its code, but expecting me to fix bugs that are not super critical is stretching my involvement too much.
> 
> I think we should at least clarify it at the next WG. Having two roles may confuse users, that refer to maintainers as "people that may know", according to my experience, not people that are expected to do.

Fair enough. So maybe we should also list maintainers. For example, assignments for issues should target maintainers, and PR assignments should be target code owners. Let's discuss it separately.

maximedenes (approved):

maximedenes:
> maximedenes approved these changes on behalf of coq/pushers

Wait! Why does GitHub say that? I'm supposed to be an owner of this file.

.github/CODEOWNERS
```diff
@@ -1,6 +1,9 @@
 # This file describes the maintainers for the main components. See
 # `dev/doc/MERGING.md`.
 
+# default
+/*      @coq/pushers
```
SkySkimmer: Do keep this change?

Zimmi48:
It's because @SkySkimmer also requested a review from the team. Therefore, it's not requesting a review from me (as a member of the team) anymore.

.github/CODEOWNERS
Zimmi48: No!

maximedenes: Maybe not, it seems to have strange interactions with individual owners.

maximedenes: In fact, you can put me back for extraction, then we clarify the distinction between maintainers and owners. Sorry for the noise.

Zimmi48: My main concern is not to spam code owners who are not watching the repository and thus should only receive e-mails / notifications when their part of the code is affected.

maximedenes (approved):

PR #7989, opened 2018-07-04
[ci] Avoid annoying detached head warning.

ejgallego:


dev/ci/ci-common.sh
```diff
@@ -69,7 +69,7 @@ git_checkout()
     if [ ! -d .git ] ; then git clone "${_DEPTH[@]}" "${_URL}" . ; fi && \
     echo "Checking out ${_DEST}"                                      && \
     git fetch "${_URL}" "${_BRANCH}"                                  && \
-    git checkout "${_COMMIT}"                                         && \
+    git -c advice.detachedHead=false checkout "${_COMMIT}"            && \
```
Zimmi48: There is a simpler way to avoid this warning: use `git checkout --detach`

SkySkimmer: @ejgallego do you want to use --detach?

ejgallego: I don't mind either way, I did indeed know about `--detach` but it changes the semantics, so I dunno; as you folks prefer.

SkySkimmer (approved):

PR #7990, opened 2018-07-04
Convert timing tool to python3

### LINKED ISSUE #7485: Timing tool should be upgraded to python3.
#### Version

Current master 83b2e3430da9ba6d234a78c5b2910070c13d4152 but it has always been the case IIRC.

#### Operating system

NixOS 18.03.132021.c0c5571ec1a (Impala) but it was also the case with older versions of this system.

The log is:

```
+ set -e
++ dirname ./run.sh
+ cd .
++ cd ../../../..
++ pwd
+ COQLIB=/home/theo/coq
+ export COQLIB
+ ./001-correct-diff-sorting-order/run.sh
+ set -e
++ dirname ./001-correct-diff-sorting-order/run.sh
+ cd ./001-correct-diff-sorting-order
+ /home/theo/coq/tools/make-both-time-files.py time-of-build-after.log.in time-of-build-before.log.in time-of-build-both.log
  File "/home/theo/coq/tools/make-both-time-files.py", line 13, in <module>
    left_dict = get_times(args[1])
  File "/home/theo/coq/tools/TimeFileMaker.py", line 44, in get_times
    lines = f.read()
  File "/nix/store/128m147iv3n0dnlq1pwd2pjfqxlsaisk-python3-3.6.4-env/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1396: ordinal not in range(128)
```

I'm using the dependencies provided by `default.nix`. Is this reproducible on other system when inside `nix-shell` (possibly with option `--pure`)? @maximedenes Since you are also on NixOS, have you had the same failures?

CC @JasonGross

jashug:
**Kind:** bug fix / feature.
Fixes / closes #7485

This makes the test-suite pass on my machine, with Python 3.5.
I also changed the `INFINITY` constant to use a unicode escape. I'm not sure if that is tested or not.

The slight changes in behavior are probably because Python 3.5 randomizes dictionary iteration order (it was never guaranteed to be consistent), and floating point addition is not associative. That was a fun bug to track down.


jashug:
Checked the dependence on iteration order by reverting to Python 2 but sorting before summing, and I get the same results as sorting before summing in Python 3.

Zimmi48:
Thanks!

CI shows some remaining problems:

Common error on all external project tests (which are tested with timing tools):

```
Aggregating timing log... 
Makefile.ci:62: recipe for target 'ci-bedrock2' failed
  File "./tools/make-one-time-file.py", line 19, in <module>
    times_dict = get_times(sys.argv[1])
  File "/builds/coq/coq/tools/TimeFileMaker.py", line 41, in get_times
    with open(file_name, 'r', encoding="utf-8") as f:
TypeError: 'encoding' is an invalid keyword argument for this function
```

There are also dependencies to fix in `.travis.yml`.

gares (approved):
I'm fine with the change, and I blindly trust you on the python code.

gares:
@Zimmi48 this is because CI is still using python2 to execute the script, it seems python3 open() has such a kwd: https://docs.python.org/3/library/functions.html

Zimmi48:
> Maybe @jashug could be more precise in the shebang line and select python3 rather than the "ambiguous" python

That's what he did. And I don't think this is the reason. On Travis, the test is failing because python3 is not available. But on GitLab CI, both python2 and python3 are available.

Zimmi48:
Or there is a missed call to an ambiguous `python` somewhere else.

gares:
You are right, the shebang line is ok. I misread a diff...

jashug:
So I revised the code, and it works with both python2 and python3 for me now.
Compatibility wasn't too hard, just do a lot of `import __future__` and get the Python 3 `open` method from `io`.
The `PYTHONIOENCODING` is because when python tries to output unicode to stdout (which is used in the test-suite), it encodes it with it's best guess of the expected encoding, which defaulted to `ascii`, which didn't work. For comparison, the previous code was using explicit utf8 byte sequences.

Zimmi48:
To make Travis pass you should still either make the shebang generic, or update the python version that gets installed, or both.

jashug:
I couldn't find a place in `.travis.yml` to change `python2` to `python3`, so I make the shebang generic.
If someone more familiar can make travis use `python3`, that would be wonderful too.

Zimmi48:
@jashug It seems that you forgot to push.

Sorry for my remark about `.travis.yml` being quite imprecise. This `brew` system on macOS makes it very difficult to understand which version of `python` is installed, especially given that some packages are depending transitively on `python2` or `python3`. So for instance, these two lines:

https://github.com/coq/coq/blob/073c612261cd474a84bddd94b80697cbc8c28488/.travis.yml#L227-L228

will install `python2` as a dependency. Installing `python3` would require adding two lines like this:

https://github.com/coq/coq/blob/073c612261cd474a84bddd94b80697cbc8c28488/.travis.yml#L245-L246

But better keep things as they are with a generic shebang (it's less complicated and it will allow testing compatibility with python2 and python3).

jashug:
Right, sorry about that. Pushed, and CI is all green.

JasonGross (approved):

JasonGross:
> The PYTHONIOENCODING is because when python tries to output unicode to stdout (which is used in the test-suite), it encodes it with it's best guess of the expected encoding, which defaulted to ascii, which didn't work. 

Is the invocation of the scripts in the Makefile the best place to put this?  Rather than, say, the test-suite invocations, or as an `export` variable in the `Makefile` (which would be more easily overridable)?  I ask because I'm not familiar with the terminal encoding setup, and I'm not familiar with why you'd ever want it to be anything other than utf-8.

jashug:
I don't know where the best place to put the variable is, but I don't think it's too bad in the Makefile. The old version always outputted utf-8, and the current version (run through the Makefile) should too. 
`PYTHONIOENCODING` also only applies to output to standard out, and output to files is also hardcoded as utf-8.
For people who want a different output encoding, they should be able to convert after the fact (as a workaround).
Another place you could put it which I just thought of is in the shebang line.

Zimmi48:
I've seen this error on a recent master build (https://gitlab.com/coq/coq/-/jobs/79981084):

```
Aggregating timing log...
  File "./tools/make-one-time-file.py", line 21, in <module>
    print_or_write_table(table, sys.argv[2:])
  File "/builds/coq/coq/tools/TimeFileMaker.py", line 209, in print_or_write_table
    print(table)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2500' in position 2280: ordinal not in range(128)
```

It might be related to this change, so I'm posting it just in case. I've just relaunched the build to confirm it was a spurious failure.

jashug:
I found where `make-one-time-file.py` was called without `PYTHONIOENCODING`, and made a PR to set it.
The error is strange though, since that isn't the infinity sign, but a box-drawing character, though not used by `TimeFileMaker.py`. So for some reason the character `─` appeared in the output in that run, and python choked. (If it had tried to output infinity instead, it still would have choked though).

Zimmi48:
Not to worry. Thanks for the quick diagnosis.

PR #7991, opened 2018-07-04
Make Travis faster by removing more builds.

Zimmi48:
We remove all the testing of external projects from Travis. The current state was inconsistent anyways.
We keep the matrix of test-suite / validate / warnings / doc targets with different OCaml versions.
If people want to do more thorough testing on their personal branches, they will have to setup GitLab CI as explained in dev/ci/README.md. It's not difficult to setup.

The motivation is that yesterday it look longer for just the linter than for the full GitLab CI test suite to run on some PRs, due to a high number of recently merged PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,9 +67,6 @@ matrix:
       - TEST_TARGET="validate"   COMPILER="${COMPILER_BE}+flambda" CAMLP5_VER="${CAMLP5_VER_BE}" NATIVE_COMP="no" EXTRA_CONF="-flambda-opts -O3" FINDLIB_VER="${FINDLIB_VER_BE}"
     - if: NOT (type = pull_request)
       env:
-      - TEST_TARGET="ci-bignums"
-    - if: NOT (type = pull_request)
-      env:
       - TEST_TARGET="ci-color"
     - if: NOT (type = pull_request)
       env:
@@ -85,9 +82,6 @@ matrix:
       - TEST_TARGET="ci-equations"
     - if: NOT (type = pull_request)
       env:
-      - TEST_TARGET="ci-fcsl-pcm"
-    - if: NOT (type = pull_request)
-      env:
       - TEST_TARGET="ci-fiat-parsers"
     - if: NOT (type = pull_request)
       env:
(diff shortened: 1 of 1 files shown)


ejgallego:
> The current state was inconsistent anyways.

What do you mean by inconsistent?

> The motivation is that yesterday it look longer for just the linter than for the full GitLab CI test suite to run on some PRs, due to a high number of recently merged PRs.

Well, that's not strong enough IMHO; it is normal that some times CI systems are overloaded from time to time.

For now, I'd like to keep some actual CI testing in Travis.

ejgallego (changes requested):
I'd like the testing to remain for a while more.

So either we can close this PR, or if you want I'd be OK with removing `fcsl-pcm` and `math-classes`.

Zimmi48:
I'll do what you propose as it's better than nothing.

Zimmi48:
Compared to your proposal, I've also removed bignums since it's already tested in color anyways.

ejgallego (approved):
Ok we can take this. As for general CI testing in Travis, I'd like to keep it functioning for now; Travis has scheduled some improvements and we don't know for sure we'll stop using it.

Zimmi48:
> Travis has scheduled some improvements and we don't know for sure we'll stop using it.

Indeed, I'm planning to continue using Travis for CI testing in coq-community so I may have some feedback to give about new goodies (like the new "Checks" feature).

Zimmi48:
I don't think we can expect any serious use of Travis in the future until they resolve their fetching bug IMHO. I'm restarting such failed jobs all the time, here is one instance I did not restart: https://travis-ci.org/coq/coq/builds/401436863

ejgallego:
If what worries you is the red light let's just remove it from the readme. IMHO keeping the build jobs for now costs very little and allows for some experiments if we want.

Zimmi48:
No, I want the green light because there is stuff that is tested only on Travis.

ejgallego:
> No, I want the green light because there is stuff that is tested only on Travis.

Only the linter, right?

Zimmi48:
Also the macOS tests (basically, what is described in the CI documentation).

ejgallego:
Ah, indeed the OSX migration to Gitlab is still not done.

Zimmi48:
No, but @maximedenes is going to work on this really soon (but we'll probably keep the test-suite on Travis, like we did with AppVeyor).

JasonGross:
You can pass the option that allows installation of unauthenticated packages, and that seems to get rid of the fetching issues for me.

PR #7984, opened 2018-07-03
Compile `coqpp` inside the `bin/` folder and make it available after installation

ppedrot:
Compile `coqpp` inside the `bin/` folder and make it available after installation. No need to put it inside `.gitignore`.

PR #7994, opened 2018-07-04
Make bin/ in makefile, not configure.

SkySkimmer:
See https://github.com/coq/coq/pull/7984#issuecomment-402315772 for context.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,10 @@ endif
 
 MAKE_OPTS := --warn-undefined-variable --no-builtin-rules
 
-submake: alienclean
+bin:
+	mkdir bin
+
+submake: alienclean | bin
 	$(MAKE) $(MAKE_OPTS) -f Makefile.build $(MAKECMDGOALS)
 
 noconfig:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -156,7 +156,10 @@ endif
 
 MAKE_OPTS := --warn-undefined-variable --no-builtin-rules
 
-submake: alienclean
+bin:
+	mkdir bin
+
+submake: alienclean | bin
```
ejgallego: Umm, I guess the rules generating the binaries should depend on the `bin` target? Anyways I am not an expert in `Makefile`.

Also, shouldn't this use `mkdir -p bin`?

gares:
```
$ mkdir a
$ mkdir a # error
```
I guess `-p` is needed but may not be spot by CI

SkySkimmer: I tried making rules generating things in bin depend on it but could not figure out how to make it work in a generic way (`bin/%: bin` didn't have an effect) and doing it per-rule seems like a pain.

About -p, the rule shouldn't get run if the directory already exists.

gares: ah right. then I'm OK with it.

gares (approved):

PR #7998, opened 2018-07-04
[coqpp] Move to its own directory.

ejgallego:
Coqpp has nothing to do with `grammar`, we thus place it in its own
directory, which will prove convenient in more modular build systems.

Note that we add `coqpp` to the list of global includes, we could have
indeed added some extra rules, but IMHO not worth it as hopefully
proper containment will be soon checked by Dune.


ppedrot:
> Coqpp has nothing to do with grammar

I disagree. The grammar `folder` is about the implementation of Coq-specific macros in camlp5. coqpp is willing to replace the latter, so for the time being they are cohabitating there, but the camlp5 code is going to disappear soon enough.

ejgallego:
> I disagree. The grammar folder is about the implementation of Coq-specific macros in camlp5. coqpp is willing to replace the latter, so for the time being they are cohabitating there, but the camlp5 code is going to disappear soon enough.

They are totally disjoint tools as in both a chicken and a lamb are edible.

The camlp5 code is not going away so soon I think due to plugins [at least 8.10] so I think I'd like to have each animal in its own directory.

ppedrot:
> The camlp5 code is not going away so soon I think due to plugins [at least 8.10] so I think I'd like to have each animal in its own directory.

I am still not convinced. You chose to call the new directory `coqpp` but it doesn't sound right as this would be the only binary with a folder to its name. What is your motivation for the split? You mention modularity, but there is currently little point in compiling separately `coqpp` and the `camlp5` macro files. Can you elaborate on that?

ejgallego:
> I am still not convinced.

I think you are complicating your point of view too much, this is an obvious change.

> You chose to call the new directory coqpp but it doesn't sound right as this would be the only binary with a folder to its name.

Both `coqtop` and `coqdoc` for example live on their own folders.

> What is your motivation for the split? You mention modularity, but there is currently little point in compiling separately coqpp and the camlp5 macro files. Can you elaborate on that?

Well they have nothing to do with each other, that's the motivation for the split. They already compile separatedly and I think having them together sends the wrong signal. In fact, we should move both tools to the proper `pp` folder as they are build-only components; and not exported as components in the META file.

ejgallego:
Here you can see what's wrong with the current layout: https://github.com/ejgallego/coq/blob/7cd9bc68d9a4625adc0834c40659850fc4347606/grammar/dune

ppedrot:
Note that I am not opposed to this change, I am just mostly unconvinced. Unluckily it seems I am the sole owner here, so I will summon people like @gares and @maximedenes that understand more about the build system, and I'll follow their opinion.

ejgallego:
The rule for directories such as `grammar` is that one (1) directory == one (1) `cma` file. That's another reason not to have the random files there.

ppedrot (approved):
Fine, I am now not-not-convinced.

gares (approved):

Issue #7753, opened 2018-06-09
Provide instructions and tools so that external projects can easily setup CI


Zimmi48:
We encourage external project authors to add their project to our CI but this generally means setting up CI on their side as well to make sure that they don't break compatibility with tested versions of Coq.

However, we have failed to provide instructions on how to do so, so far.

@thery has written some notes on how to set up Travis using OPAM to install Coq: https://github.com/thery/note/blob/master/travis.md

It is also possible to use Docker with Travis and some projects are doing that: https://github.com/math-comp/multinomials/blob/master/.travis.yml

IIUC @ejgallego's plan is to automatically build a Docker image containing Coq and its dependencies on our CI, store it in the GitLab Docker registry, and make it available for project maintainers to use on their side.
The main advantage of Docker over OPAM for external projects would be a reduced build time.
A caveat is that we might ship more dependencies than what is actually needed in the Docker image given that our own base image includes dependencies of external projects as well. Unless there exists a technical solution to avoid this. To discuss.

Travis is probably OK for most external projects but for some projects with long build time like VST, math-comp or GeoCoq, we can recommend using GitLab CI instead to avoid timeouts and to allow testing the project in full. Setting up GitLab CI for testing branches of a GitHub project is really easy these days. The only unsupported part is testing pull requests coming from forks. But I've programmed @coqbot (cf. https://github.com/coq/bot) to synchronize PRs on Coq's GitHub repository to GitLab. I can quite easily adapt it be able to synchronize any PR from a fork on any GitHub repository to a GitLab repository with same name and same owner, as long as it has push access to the GitLab repository (project maintainers will only have to set up a webhook, setup their GitLab CI-only mirror, and grant push access to @coqbot to this mirror).

So the plan looks like this:

- [x] Automatically generate Docker images for Coq branch builds.
- [x] Document the setup using Docker on Travis.
- [x] Document the setup using Docker on GitLab.
- [ ] Generalize @coqbot to be able to sync PRs on various repositories.

Zimmi48:
Note that within coq-community, I'm now pursuing another approach based on Nix (+ Travis). If it turns out to be satisfying (I think it will), it will get sufficiently documented to serve the community of package authors at large.

See for instance the lemma-overloading package where this approach is already implemented: https://github.com/coq-community/lemma-overloading

Zimmi48:
[...]

PR #8001, opened 2018-07-05
Cache the build of the Nix package using Cachix.

Zimmi48:
**Kind:** infrastructure.

This is one more step towards #7161 and #7753.

This PR adds support for the binary cache https://coq.cachix.org for the Coq Nix package.
More precisely, if the `CACHIX_SIGNING_KEY` variable is defined, it will sign the produced packages and push them to the above mentioned binary cache. I'll put this variable on GitLab CI as a protected variable, which means that it will only apply to builds on protected branches, such as master or v8.8.

Since the variable is also defined on my personal GitLab fork, and the build already passed there https://gitlab.com/Zimmi48/coq/-/jobs/79593568, the build of the Nix package for this PR should be quite fast (just fetching binaries).

This was initially inspired by the way Cachix does for its own package (https://github.com/cachix/cachix/, see in particular https://github.com/cachix/cachix/blob/7e31c778db910c829db4235a065f8c5ba46da20f/.travis.yml#L15) but I've changed the way to fetch the package into `nix-build -E "import (fetchTarball $CI_PROJECT_URL/-/archive/$CI_COMMIT_SHA.tar.gz) {}"` because this will avoid depending on the exact URL from where the tarball is downloaded. Instead, the hash of the tarball is computed after download which means that I can compile `$CI_PROJECT_URL/-/archive/$CI_COMMIT_SHA.tar.gz` and then install the package from https://github.com/coq/coq/tarball/master.

It also means that I can depend on this package in other packages. This allows me to setup CI on some external package in the simplest way: https://github.com/Zimmi48/aac-tactics/commit/b788c75f93c2edb6836b73d295e31224830f4a64.  See the corresponding build here https://travis-ci.com/Zimmi48/aac-tactics/builds/78159513. It ends up failing because aac_tactics is not yet compatible with Coq master. But notice how Coq is fetched and not rebuilt.

cc @domenkozar: I'm sure you'll be interested to hear about this application of Cachix.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -245,7 +245,8 @@ pkg:nix:
   before_script: [] # We don't want to use the shared 'before_script'
   script:
     - export TMPDIR=$PWD
-    - nix-build "$CI_PROJECT_URL/-/archive/$CI_COMMIT_SHA.tar.gz" -K
+    - nix-env -if https://github.com/cachix/cachix/tarball/master --substituters https://cachix.cachix.org --trusted-public-keys cachix.cachix.org-1:eWNHQldwUO7G2VkjpnjDbWwy4KQ/HNxht7H4SSoMckM=
+    - nix-build "$CI_PROJECT_URL/-/archive/$CI_COMMIT_SHA.tar.gz" -K --extra-substituters https://coq.cachix.org --trusted-public-keys "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= coq.cachix.org-1:Jgt0DwGAUo+wpxCM52k2V+E0hLoOzFPzvg94F65agtI=" | if [ ! -z "$CACHIX_SIGNING_KEY" ]; then cachix push coq; fi
   artifacts:
     name: "$CI_JOB_NAME.logs"
     when: on_failure
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -240,13 +240,25 @@ pkg:nix:
   image: nixorg/nix:latest # Minimal NixOS image which doesn't even contain git
   stage: test
   variables:
+    # By default we use coq.cachix.org as an extra substituter but this can be overridden
+    EXTRA_SUBSTITUTERS: https://coq.cachix.org
+    EXTRA_PUBLIC_KEYS: coq.cachix.org-1:Jgt0DwGAUo+wpxCM52k2V+E0hLoOzFPzvg94F65agtI=
+    # The following variables should not be overridden
     GIT_STRATEGY: none
+    CACHIX_PUBLIC_KEY: cachix.cachix.org-1:eWNHQldwUO7G2VkjpnjDbWwy4KQ/HNxht7H4SSoMckM=
+    NIXOS_PUBLIC_KEY: cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
+
   dependencies: [] # We don't need to download build artifacts
   before_script: [] # We don't want to use the shared 'before_script'
   script:
+    # Use current worktree as tmpdir to allow exporting artifacts in case of failure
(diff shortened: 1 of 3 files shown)


vbgl (approved):
Can you please document where these random-looking strings come from (or at least what they are)? In case they have to be updated some day. Thanks.

.gitlab-ci.yml
```diff
@@ -245,7 +245,8 @@ pkg:nix:
   before_script: [] # We don't want to use the shared 'before_script'
   script:
     - export TMPDIR=$PWD
-    - nix-build "$CI_PROJECT_URL/-/archive/$CI_COMMIT_SHA.tar.gz" -K
+    - nix-env -if https://github.com/cachix/cachix/tarball/master --substituters https://cachix.cachix.org --trusted-public-keys cachix.cachix.org-1:eWNHQldwUO7G2VkjpnjDbWwy4KQ/HNxht7H4SSoMckM=
```
ejgallego: Indeed sorry, put the updatable parts in variables `COQ_NIX_FOO`

Zimmi48:
The new commit adds documentation comments and the possibility to override the additional cache.

Zimmi48:
@SkySkimmer: can you review and merge?

SkySkimmer (changes requested):
The exports and keys should be in a `variables` section.

Zimmi48:
Done

.gitlab-ci.yml
```diff
@@ -241,11 +241,21 @@ pkg:nix:
...
+    EXTRA_PUBLIC_KEYS: "${EXTRA_PUBLIC_KEYS-coq.cachix.org-1:Jgt0DwGAUo+wpxCM52k2V+E0hLoOzFPzvg94F65agtI=}"
+
   dependencies: [] # We don't need to download build artifacts
   before_script: [] # We don't want to use the shared 'before_script'
   script:
+    # Use current worktree as tmpdir to allow exporting artifacts in case of failure
     - export TMPDIR=$PWD
-    - nix-build "$CI_PROJECT_URL/-/archive/$CI_COMMIT_SHA.tar.gz" -K
+    # Install Cachix as documented at https://github.com/cachix/cachix
+    - nix-env -if https://github.com/cachix/cachix/tarball/master --substituters https://cachix.cachix.org --trusted-public-keys cachix.cachix.org-1:eWNHQldwUO7G2VkjpnjDbWwy4KQ/HNxht7H4SSoMckM=
```
SkySkimmer: This key and the other one should also be in variables.

.gitlab-ci.yml
```diff
@@ -241,11 +241,21 @@ pkg:nix:
   stage: test
   variables:
     GIT_STRATEGY: none
+    # By default we use coq.cachix.org as an extra substituter but this can be overridden
+    EXTRA_SUBSTITUTERS: "${EXTRA_SUBSTITUTERS-https://coq.cachix.org}"
+    EXTRA_PUBLIC_KEYS: "${EXTRA_PUBLIC_KEYS-coq.cachix.org-1:Jgt0DwGAUo+wpxCM52k2V+E0hLoOzFPzvg94F65agtI=}"
```
SkySkimmer: Just put in the value, project variables have precedence over job variables.

Zimmi48: Ah, great to know!

.gitlab-ci.yml
Zimmi48: I don't see why you would want to replace the official cache for this program without replacing the source for the program as well. This might be over-thinking it.

SkySkimmer: It has nothing to do about replacing, it's about moving the magic value out of the script.

SkySkimmer (approved):

Zimmi48:
CI passed.

domenkozar:
FYI: once stackage LTS-12 hits nixpkgs (hopefully anytime now), it's going to be easier to bootstrap cachix.

Issue #8004, opened 2018-07-05
Universe polymorphism issue new to 8.8


jwiegley:
#### Version

The Coq Proof Assistant, version 8.8.1 (June 2018)
compiled on Jun 30 2018 8:37:45 with OCaml 4.05.0

#### Operating system

macOS 10.13.5

#### Description of the problem

The attached code works fine with 8.5-8.7, but fails in 8.8. It works if I disable `Universe Polymorphism` for this example, but this is something I cannot do in the full context where this is used. I'd like to know if this is a bug, or if not, an understanding of why it has become an issue in 8.8.

[Bug.zip](https://github.com/coq/coq/files/2168095/Bug.zip)

SkySkimmer:
Seems to work again in master.
Will try to bisect.

SkySkimmer:
Appears e6c87412d70b71daaf417bd4b8e4ae6f1f28515b 
Fixed f54cbf01ad2aab14eed671e02b98f6962b05a38d
That commit could probably be backported if we want to.

jwiegley:
I can easily wait for the next point release of 8.8 too, if I'm the only one this affects.

Zimmi48:
Unfortunately, you'll have to since the 8.8.1 release has already been tagged.

JasonGross:
If this appeared and was fixed without anyone noticing, it should get a test-case in the test-suite.

PR #8008, opened 2018-07-06
Add test for #8004.

SkySkimmer:


Zimmi48:
Merging this one as well since this is the same test case as I just merged in v8.8.

PR #8012, opened 2018-07-06
Set PYTHONIOENCODING in ci-wrapper

jashug:
**Kind:** bug fix
Fix unicode encoding bug noticed after #7990 was merged.

This should work, but I didn't test it. Take it with a grain of salt.
I can write something which always outputs UTF-8, but I'll have to case on `sys.version_num`.

@JasonGross @Zimmi48

### DIFF
--- a/dev/ci/ci-wrapper.sh
+++ b/dev/ci/ci-wrapper.sh
@@ -21,6 +21,7 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd "${DIR}/../.."
 
 export TIMED=1
+export PYTHONIOENCODING=UTF-8
 "${DIR}/${CI_SCRIPT}" 2>&1 | tee time-of-build.log
 travis_fold 'start' 'coq.test.timing' && echo 'Aggregating timing log...'
 python ./tools/make-one-time-file.py time-of-build.log


SkySkimmer:
This seems fragile. Is there some reason we can't set whatever this is in the python program itself?

jashug:
Unicode is text, bytes are how computers store text. `TimeFileMaker.py` assembles a block of text, and wants to output it. For files that we open, we can tell python to store the text by encoding it using `UTF-8`, and write the text. For standard output, Python opens it using it's best guess for the encoding. You can tell python explicitly what encoding to use with the `PYTHONIOENCODING` environment variable.

So far is true for both Python 2 and Python 3. In both versions, you can write an explicit byte stream to standard out, but the way to do so is different.
Python 2: `print(s.encode('utf-8'), file=sys.stdout)`.
Python 3: `print(s.encode('utf-8'), file=sys.stdout.buffer)`
`sys.stdout.buffer` does not exist in Python 2, so we could write
```
try:
    binary_stdout = sys.stdout.buffer
except AttributeError:
    binary_stdout = sys.stdout
print(s.encode('utf-8'), file=binary_stdout)
```

Would this be a preferable approach?

jashug:
@Zimmi48 rebased

JasonGross:
> Would this be a preferable approach?

Yes, I'd prefer having this set by the script in this way than by needing to make sure we set the environment correctly on all calls.

ejgallego:
I am a bit of the opinion of @SkySkimmer here, setting this in the script is a sure recipe for divergence as soon as a developer uses this script by other means.

The first question is: why are we using unicode? If that's only for a couple of symbols, we could just don't use unicode and be done with it.

Or, if unicode needs to be used, then the python program should ensure the invariant _locally_.

JasonGross:
@ejgallego do you object to the proposed solution of:
```python
try:
    binary_stdout = sys.stdout.buffer
except AttributeError:
    binary_stdout = sys.stdout
print(s.encode('utf-8'), file=binary_stdout)
```
?

As for why Unicode: so that we support Unicode in .v files, in .v filenames, etc, and also because being able to use the infinity symbol is nice.

ejgallego:
> @ejgallego do you object to the proposed solution of:

I am not Python expert so I can't judge; it seems strange tho that you have to setup the encoding each time you output; maybe you should setup a common global object with the right setup.

JasonGross:
Eh, I think it's standard for python.  An alternative would be:
```python
try:
    binary_stdout = sys.stdout.buffer
except AttributeError:
    binary_stdout = sys.stdout
def print_utf8(s): print(s.encode('utf-8'), file=binary_stdout)
```
and then just call `print_utf8`.

jashug:
@ejgallego The "Pythonic way", as I understand it, is to just set up the encoding once when you open a file. `sys.stdout` makes that complicated, because it is opened by the interpreter before your program is loaded.
The thread https://mail.python.org/pipermail/python-ideas/2012-June/015395.html includes some takes on the issue by core Python developers. Their take seems to be that your programs don't have any business telling the surrounding environment what encoding to use. If you need to output binary data to stdout, you can, but it's rare (and Python 2 and Python 3 do it differently). If you are talking to standard output, the assumption is that you are talking text, and so `stdout` is treated as a text file, with an encoding dictated by the surrounding environment.
As for why Unicode, I would answer that the output of these programs is text, meant to be read by humans, and thus manipulating binary representations of that text is not the appropriate level of abstraction. You can assemble text by manipulating byte sequences, but that's a bit like writing higher-order functions as transformations on syntax.

JasonGross:
This should be closed now that #8015 has been merged, right?

jashug:
Yep

PR #8023, opened 2018-07-09
Introduce a team of code owners for the CI system.

Zimmi48:
This means that all the members of the team will receive a review request for PRs on the CI, but only one of them will need to approve the PR, and this will remove the review request for the others.

Currently the team contains Emilio and Gaetan, the two current code owners of these files. It makes sense to start experimenting on this component since they had already decided to make their role symmetric.

Updating the list of maintainers can be done by updating the list members, and without changing the CODEOWNER file.

If this is seen as a good idea, we can slowly convert the rest of the CODEOWNER file using this style (add a documentation-maintainers team, build-system-maintainers team, etc.).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,21 +8,14 @@
 
 ########## CI infrastructure ##########
 
-/dev/ci/               @ejgallego
-# Secondary maintainer @SkySkimmer
+/dev/ci/          @coq/ci-maintainers
+/.circleci/       @coq/ci-maintainers
+/.travis.yml      @coq/ci-maintainers
+/.gitlab-ci.yml   @coq/ci-maintainers
 
 /dev/ci/user-overlays/*.sh @ghost
 # Trick to avoid getting review requests
 # each time someone adds an overlay
-
-/.circleci/       @SkySkimmer
-# Secondary maintainer @ejgallego
-
-/.travis.yml      @ejgallego
-# Secondary maintainer @SkySkimmer
-
-/.gitlab-ci.yml   @SkySkimmer
-# Secondary maintainer @ejgallego
 
 /appveyor.yml          @maximedenes
 /dev/ci/appveyor.*     @maximedenes


SkySkimmer:
>This means that all the members of the team will receive a review request for PRs on the CI, but only one of them will need to approve the PR, and this will remove the review request for the others.

Do you know if it works when the PR is from a member of the team?

ejgallego (commented):
works for me

Zimmi48:
> Do you know if it works when the PR is from a member of the team?

Good question. Do you want to get the answer before approving this? (I can try doing some tests in another repository.) Otherwise I guess we'll see, it shouldn't be much different than the situation today when the principal maintainer submits a PR...

SkySkimmer (approved):

ejgallego (approved):

maximedenes:
> Good question. Do you want to get the answer before approving this? (I can try doing some tests in another repository.) Otherwise I guess we'll see, it shouldn't be much different than the situation today when the principal maintainer submits a PR...

I think it would be good to test that.

Zimmi48:
> I think it would be good to test that.

This is requiring me more work, so unless you see a real problem, I'd prefer to merge as-is.

maximedenes (approved):
Ok, we'll see what it gives in practice.

ejgallego:
So now it turns out that I am getting a review request on my own PRs; I don't see a way to remove that other than removing the team.

This does indeed a bit of noise unfortunately to my review requested list :(

Zimmi48:
So it looks then like the answer to this question:

> Do you know if it works when the PR is from a member of the team?

was yes, it even works too often.

Anyways, IMO this is a GitHub-bug, so I'm going to report it. Hopefully they fix it and we can expand the use of this feature.

maximedenes:
> So now it turns out that I am getting a review request on my own PRs

I would be curious to see how you review your PRs, and how you reply to your reviews :)

Zimmi48:
You can leave review comments on your own PRs. But if you do this, it will remove the request for review from the rest of the team.

Zimmi48:
@ejgallego Could you test and confirm what I've just said? If it's really the case, I'll follow up on my GitHub bug report because this is another issue.

ejgallego:
> @ejgallego Could you test and confirm what I've just said? If it's really the case, I'll follow up on my GitHub bug report because this is another issue.

I can comment, but not approve or denegate; after commenting the review request is still shown.

SkySkimmer:
If I'm the one commenting it goes away as usual.

Zimmi48:
> I can comment, but not approve or denegate; after commenting the review request is still shown.

OK, so at least that is good.

PR #8029, opened 2018-07-10
Fix #7845: put -R . Top at beginning of args instead of at end

### LINKED ISSUE #7845: coq_makefile should not add -R . Top to the end when -Q is passed, even when there is no common root
#### Version

8.8.0

Consider this:
```bash
$ mkdir -p theories; touch a.v theories/b.v
$ coq_makefile -Q theories Foo a.v theories/b.v -o Makefile
Warning: theories (used in -R or -Q) is not a subdirectory of the current directory

Warning: no common logical root
Warning: in such case INSTALLDEFAULTROOT must be defined
Warning: the install-doc target is going to install files
Warning: in orphan_Foo_Top
$ make VERBOSE=1
"coqdep" -dyndep var  -I . -Q theories Foo -R . Top a.v theories/b.v > ".coqdeps.d" || ( RV=$?; rm -f ".coqdeps.d"; exit $RV )
make --no-print-directory -f "Makefile" pre-all
if [ "8.8.0" != "8.8.0" ]; then\
  echo "W: This Makefile was generated by Coq 8.8.0";\
  echo "W: while the current Coq version is 8.8.0";\
fi
make --no-print-directory -f "Makefile" real-all
"coqc"   -q  -I . -Q theories Foo -R . Top  a.v
Warning: /home/jgross/Documents/repos/fiat-crypto/food/theories was
previously bound to Foo; it is remapped to Top.theories
[overriding-logical-loadpath,loadpath]
"coqc"   -q  -I . -Q theories Foo -R . Top  theories/b.v
Warning: /home/jgross/Documents/repos/fiat-crypto/food/theories was
previously bound to Foo; it is remapped to Top.theories
[overriding-logical-loadpath,loadpath]
make --no-print-directory -f "Makefile" post-all
```
This interferes with my coq-bug-minimizer, because it uses `make` to generate dependencies and .glob files.

I would prefer for `coq_makefile` to only add `-Q . Top` when there is no common root, but I'd be okay with it adding `-R . Top` to the *beginning* of the list rather than the end.

cc @gares

SkySkimmer:
Anyone feel free to push a test suite addition.

### DIFF
--- a/tools/coq_makefile.ml
+++ b/tools/coq_makefile.ml
@@ -309,8 +309,8 @@ let ensure_root_dir
     let source x = {thing=x; source=CmdLine} in
     let here_path = { path = "."; canonical_path = here } in
     { project with
-        ml_includes = source here_path :: ml_includes;
-        r_includes = source (here_path, "Top") :: r_includes }
+        ml_includes = ml_includes @ [source here_path];
+        r_includes = r_includes @ [source (here_path, "Top")] }
 ;;
 
 let warn_install_at_root_directory 


JasonGross (approved):
Thanks!  Here is a test (I'm still working on getting google approval to push code / make commits, but I think this is short enough to not matter; could you push something like it to the test-suite?)
```bash
mkdir -p theories; touch a.v theories/b.v; (echo 'Require Import Foo.b.' > a.v)
coq_makefile -Q theories Foo a.v theories/b.v -o Makefile
make clean && make
```

This fails on current 8.8, and should succeed on this branch.

SkySkimmer:
Looks like SF is failing.

JasonGross:
Oh, I see.  If we do `-R . Top`, then we can `Require Import` without prefixing with `Top.`.  So we should either do `-Q . ""`, or keep `-R . Top` (but still move it to the beginning).  The latter is more backwards-compatible, but the former seems possibly nicer?

JasonGross:
I'd suggest leaving `-R . Top`, and moving it to the beginning, and then we can even probably backport this to 8.8.2.  (Unless you have strong feelings about `Top` vs `""`.)

SkySkimmer:
I don't like `""` so that's fine.

JasonGross (changes requested):
This does not actually fix the problem (and, in fact, makes it worse); I was not careful enough in my testing to diagnose the issue, so I'm going to go update the issue with a more precise description.

JasonGross:
I think the actual issue that needs to be fixed here is #8038

SkySkimmer:
>This does not actually fix the problem (and, in fact, makes it worse)

OK then.

PR #8031, opened 2018-07-10
Remove auto-retry in GitLab CI now that @coqbot is handling it.

Zimmi48:
**Kind:** infrastructure.

@coqbot is now automatically retrying builds which failed because of a runner failure (those without the magic sentence "The build completed normally"). You may have noticed already, but it's also adding new status checks for real failures too.


SkySkimmer (approved):
Praise coqbot

ejgallego (approved):
Hail @coqbot

PR #8034, opened 2018-07-10
[travis] Try to workaround the repeated APT failures by using Jason Gross's suggestion.

Zimmi48:
**Kind:** infrastructure.

cf. https://github.com/mit-plv/fiat-crypto/blob/390dee3042911a36f51910ff9afd657f2106c922/.travis.yml#L21-L22 and https://github.com/coq/coq/pull/7991#issuecomment-403548687

We don't benefit from this in every build because some of them are overriding the install steps. It should be possible to make this more generic. I'd be happy to have someone push a commit doing this on top of this branch. It can also wait for later.


ejgallego (approved):
I am not sure this should make an effect but why not?

PR #8035, opened 2018-07-10
Fix #8033: Tactic assert-suceeds has a typo in its name in the manual

### LINKED ISSUE #8033: Doc of "assert_succeeds" and "assert_fails" should ideally be explained at a high level.
The index incorrectly references the tactic as "assert_suceed" with a single "c".
https://coq.inria.fr/refman/coq-tacindex.html

By the way, the documentation of "assert_succeeds" and "assert_fails" should ideally be explained at a high level, independently of their implementation.
"Coq provides a derived tactic to check that a tactic fails" does not quite describe the arguments, the return value and the (absence of) sided-effects of the tactic.

https://coq.inria.fr/refman/proof-engine/ltac.html#coq:tacn.assert-suceeds

Thanks

charguer:
Fixed typo in documentation.

### DIFF
--- a/doc/sphinx/proof-engine/ltac.rst
+++ b/doc/sphinx/proof-engine/ltac.rst
@@ -117,7 +117,7 @@ mode but it can also be used in toplevel definitions as shown below.
                      : | numgoals
                      : | guard `test`
                      : | assert_fails `tacexpr3`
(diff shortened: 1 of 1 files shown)


Zimmi48 (approved):
Thanks!

PR #8037, opened 2018-07-10
Export a function to apply toplevel tactic values in Tacinterp.

ppedrot:
This is a function that keeps being asked or reimplemented. It doesn't hurt adding it to the Ltac API.

### DIFF
--- a/plugins/ltac/tacinterp.ml
+++ b/plugins/ltac/tacinterp.ml
@@ -141,16 +141,6 @@ let extract_trace ist = match TacStore.get ist.extra f_trace with
 | None -> []
 | Some l -> l
 
-module Value = struct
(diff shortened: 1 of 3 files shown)


ppedrot:
@gares Most of the uses I encountered were using `Val.t` directly because they were plugging into the Ltac runtime. This version is more atomic, and it would require boilerplate to undo what the injection would do. Also, you don't always have access to the generic argument type, at least directly.

gares (approved):

ppedrot:
@gares You should probably assign this PR to yourself. You're the one that I heard complaining most about the lack of this function after all...

ppedrot:
@gares ping!

PR #8000, opened 2018-07-05
Fix #7854: Native compilation + flambda trigger SEGFAULT.

ppedrot:
We use a more abstract representation for accumulators in the native compilation scheme, that requires less fiddling with low-level implementation details. It might be slower though.

It should probably be backported, as it doesn't break the API.

Fixes #7854.

PR #7871, opened 2018-06-20
[ci] Upgrade edge OCaml to 4.07.0 and Dune to 1.0.0

ejgallego:
- We update the OCaml version used in the base CI image.
- Windows / OSX image building is also updated to use newer OCaml.
- We also update Dune to 1.0.0.

PR #8040, opened 2018-07-11
[ci] Enable native compiler in `egde:flambda` build.

ejgallego:
OCaml 4.07.0 should have fixed the [memory eating bug](https://caml.inria.fr/mantis/view.php?id=7630)

Depends on #8000

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -200,7 +200,7 @@ build:edge+flambda:
   variables:
     OPAM_SWITCH: edge
     OPAM_VARIANT: "+flambda"
-    COQ_EXTRA_CONF: "-native-compiler no -coqide opt -flambda-opts "
+    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt -flambda-opts "
     COQ_EXTRA_CONF_QUOTE: "-O3 -unbox-closures"
 
 windows64:
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ matrix:
       - TEST_TARGET="validate"   COMPILER="4.02.3+32bit" TW="travis_wait"
     - if: NOT (type = pull_request)
       env:
-      - TEST_TARGET="validate"   COMPILER="${COMPILER_BE}+flambda" CAMLP5_VER="${CAMLP5_VER_BE}" NATIVE_COMP="no" EXTRA_CONF="-flambda-opts -O3" FINDLIB_VER="${FINDLIB_VER_BE}"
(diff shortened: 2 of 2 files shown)


SkySkimmer:
I assume the dependency is #7871?

ejgallego:
> I assume the dependency is #7871?

Yup sorry.

ejgallego:
Full travis build: https://travis-ci.org/ejgallego/coq/builds/402707917

ppedrot:
This should probably also depend on #8000.

ejgallego:
> This should probably also depend on #8000.

No need to; the stdlib compiles fine so far and the don't conflict.

ejgallego:
> No need to; the stdlib compiles fine so far and the don't conflict.

What I ignore is if we ever test the result of the native compilation.

ppedrot:
> What I ignore is if we ever test the result of the native compilation.

I doubt it, as the accumulator function which is used virtually everytime we have an open function does not work at all with flambda.

ejgallego:
> I doubt it, as the accumulator function which is used virtually everytime we have an open function does not work at all with flambda.

So that seems like a critical bug on the test-suite, doesn't it?

ppedrot:
@ejgallego Probably.

ejgallego:
> So that seems like a critical bug on the test-suite, doesn't it?

I am wrong, indeed the test suite seems to fail.

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN opam init -a -y -j $NJOBS --compiler="$COMPILER" default https://opam.ocaml.
 # Common OPAM packages.
 # `num` does not have a version number as the right version to install varies
 # with the compiler version.
-ENV BASE_OPAM="num ocamlfind.1.8.0 jbuilder.1.0+beta20 ounit.2.0.8" \
+ENV BASE_OPAM="num ocamlfind.1.8.0 dune.1.0.0 ounit.2.0.8" \
     CI_OPAM="menhir.20180530 elpi.1.0.4 ocamlgraph.1.8.8"
```
MSoegtropIMC: @ejgallego : can you explain the consequences of the jbuilder vs dune change? Is this related to #6857 ? Isn't in this PR intentionally?

ejgallego: They are the same software, `jbuilder` got renamed to `dune` due to trademark issues.

SkySkimmer (approved):

ejgallego:
Should be ready [provided CI works]

SkySkimmer (approved):

PR #6930, opened 2018-03-06
Make -warn-error fail on warnings emitted by coqc on stdlib.

maximedenes:
We turn all Coq warnings that are on by default into errors.

Fails for now, as some warnings on deprecated notations are emitted.

PR #8041, opened 2018-07-11
[ci] Remove warning jobs in favor of default `-warn-error yes`

ejgallego:
As discussed in #6930, we remove the warnings jobs and instead do
require the developers to submit a clean build.

### DIFF
--- a/configure.ml
+++ b/configure.ml
@@ -647,9 +647,10 @@ let camltag = match caml_version_list with
     48: implicit elimination of optional arguments: too common
     50: unexpected documentation comment: too common and annoying to avoid
     56: unreachable match case: the [_ -> .] syntax doesn't exist in 4.02.3
+    58: "no cmx file was found in path": See https://github.com/ocaml/num/issues/9
     59: "potential assignment to a non-mutable value": See Coq's issue #8043
 *)
-let coq_warnings = "-w +a-4-9-27-41-42-44-45-48-50-59"
+let coq_warnings = "-w +a-4-9-27-41-42-44-45-48-50-58-59"
 let coq_warn_error =
     if !prefs.warn_error
     then "-warn-error +a"
--- a/configure.ml
+++ b/configure.ml
@@ -647,8 +647,9 @@ let camltag = match caml_version_list with
     48: implicit elimination of optional arguments: too common
     50: unexpected documentation comment: too common and annoying to avoid
     56: unreachable match case: the [_ -> .] syntax doesn't exist in 4.02.3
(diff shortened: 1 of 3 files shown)


ejgallego:
I'm making this depend on #7871 as I am seeing some strange things that i cannot reproduce on 4.07.0

ejgallego:
It seems the build system is broken when not using -local, needs more research.

SkySkimmer:
>It seems the build system is broken when not using -local

Is it really about -local?

ejgallego:
> Is it really about -local?

It seemed to me, but now I have no clue. Very weird.

ejgallego:
The thing is that I cannot reproduce reliably, WTH

ejgallego:
Maybe it is due to -O options

ejgallego:
indeed.

configure.ml
```diff
@@ -647,8 +647,10 @@ let camltag = match caml_version_list with
     48: implicit elimination of optional arguments: too common
     50: unexpected documentation comment: too common and annoying to avoid
     56: unreachable match case: the [_ -> .] syntax doesn't exist in 4.02.3
+    59: "no cmx file was found in path" See https://github.com/ocaml/num/issues/9
+    59: "A potential assignment to a non-mutable value was detected in this source file." See issue #8043
```
SkySkimmer: These comments are a bit messed up.

SkySkimmer (changes requested):

SkySkimmer:
Dependency merged.

configure.ml
ejgallego: Let's see now.

SkySkimmer: Still messed up: the commit for -59 is adding a duplicated comment.

ejgallego: should be fixed now.

configure.ml
```diff
@@ -647,8 +647,11 @@ let camltag = match caml_version_list with
     48: implicit elimination of optional arguments: too common
     50: unexpected documentation comment: too common and annoying to avoid
     56: unreachable match case: the [_ -> .] syntax doesn't exist in 4.02.3
+    58: "no cmx file was found in path": See https://github.com/ocaml/num/issues/9
+    59: "A potential assignment to a non-mutable value was detected in this source file.":
+        See Coq's issue #8043
```
SkySkimmer: It would be nicer it this fit in 1 line, maybe
`59: potential assignment to a non-mutable value, see #8043`

ejgallego: Fixed.

SkySkimmer (approved):

PR #8054, opened 2018-07-12
[dev] Autogenerate OCaml dev files.

ejgallego:
For now we only copy the templates, but we could do more fancy stuff.

This helps to be compatible with build systems that take care of these
files automatically, see:
https://github.com/coq/coq/pull/6857#discussion_r202096579


ejgallego:
Where to insert the dependency is indeed a good question, suggestions welcome.

.gitignore
```diff
@@ -179,3 +179,8 @@ test-suite/.nra.cache
 
 plugins/ssr/ssrparser.ml
 plugins/ssr/ssrvernac.ml
+
+# ocaml dev files
+.merlin
+ide/.merlin
```
SkySkimmer: redundant

Makefile.dev
```diff
@@ -85,13 +85,28 @@ endif
 # But these partial targets could be quite handy for quick builds
 # of specific components of Coq.
 
+###########################################################################
+# OCaml dev files
+###########################################################################
+camldevfiles: $(MERLINFILES) META.coq
+	echo "foo $(MERLINFILES)"
```
SkySkimmer: ?

dev/tools/coqdev.el
```diff
@@ -33,7 +33,7 @@
 
 (defun coqdev-default-directory ()
   "Return the Coq repository containing `default-directory'."
-  (let ((dir (locate-dominating-file default-directory "META.coq")))
+  (let ((dir (locate-dominating-file default-directory "META.coq.in")))
```
SkySkimmer: If that file is meant to leave it may be worth finding something more robust.

Makefile.dev
ejgallego: Debug code, fixed sorry.

dev/tools/coqdev.el
ejgallego: We could use `dune-project` once it is in the tree.

ejgallego: In fact Tuareg should learn to do that automatically.

SkySkimmer: This is coq specific though.

ejgallego: Once we use a standard build system I see little coq-specific stuff there.

SkySkimmer:
little <> nothing

Gaëtan Gilbert

ejgallego:
> little <> nothing

`coqdev.el` can just depend on Tuareg and use its project location.

SkySkimmer:
It's not a submode of tuareg

Gaëtan Gilbert

gares (approved):

ppedrot (commented):
I don't quite know why I am supposed to be an owner here, probably the move of the `ide` file. Absolutely no comment on this PR.

Zimmi48 (commented):
I don't quite get how this PR represents an improvement by itself. Is it just a preparatory step in order to have Dune properly generate these files but still allow one to use make and then only copy them? In that case, it means that longer term, these `.in` files would get completely removed, right?

Weak request: could you use this opportunity to move these top-level files (`.merlin.in` and `META.coq.in`) to a sub-folder (for instance `dev/`)?

ejgallego:
> I don't quite get how this PR represents an improvement by itself. Is it just a preparatory step in order to have Dune properly generate these files but still allow one to use make and then only copy them?

Yes. Also this is useful for make maintainers if they want to set version number, flags, etc... automatically.

> Weak request: could you use this opportunity to move these top-level files (.merlin.in and META.coq.in) to a sub-folder (for instance dev/)?

In fact I tried this approach first, but it is worse I think. How would you properly resolve naming and location?

Zimmi48:
> In fact I tried this approach first, but it is worse I think. How would you properly resolve naming and location?

I don't know. Thanks for trying.

ejgallego:
> I don't know. Thanks for trying.

Maybe we can move `META` but the way is now is more standard. Tho, I can confirm that I will remove `META` support from `make` as soon as I can. I do not want to maintain this file. That means in practice when opam packages are built with Dune.

Regarding `.merlin`, the problem is that `dev` folder could also contain a `.merlin` file, so it seems tricky to distinguish. Note that as is stands today, Dune doesn't generate this top-level file.

Zimmi48:
> Note that as is stands today, Dune doesn't generate this top-level file.

But the plan is that it does? Because otherwise, I wouldn't see the point of having this copying step.

ejgallego:
> But the plan is that it does?

It is not needed, instead, it generates a `.merlin` file for each library under the proper directory.

Zimmi48:
Oh I see, so no top-level `.merlin` at all then.

ejgallego:
> Oh I see, so no top-level .merlin at all then.

Yeah, the make system could do the same if they wanted of course.

maximedenes (approved):
codeowners change looks good

ejgallego:
@gares ping, it seems this is not on your merge queue, however it is approaved.

PR #7530, opened 2018-05-16
Evarutil.finalize: combine minimize, to_constr and restrict.

SkySkimmer:
I finally figured out how to get the univ variables for variable numbers of constrs (for universe restriction) in a sensible way, leading to this combination. See diff for what it looks like in practice.

cc @ybertot 

If you look at the context of the changes you may notice some
Pretyping.check_evars calls. This is in some way redundant with
to_constr's abort_on_undefined_evars, and is sometimes called after
a non-evar-permissive to_constr (indicating a probable bug introduced
by the abort flag).

I didn't fix any of those, leaving the order as-is. It should be done
in a separate PR, but we may want to consider how to deal with the
redundancy in a more principled manner. Maybe to_constr should be
raising a sensible error, not an anomaly? There is also
flush_and_check_evars which is used only once.

PR #8056, opened 2018-07-13
[performance] Evarutil.finalize redux

jashug:
@SkySkimmer 
This is a fork of coq/coq#7530, `SkySkimmer:evar-finalize`, to investigate
the performance loss there. The uses of finalize in `vernac/comAssumption.ml` and
`vernac/comDefinition.ml` at most change whether the universes are
calculated after normalization by either `Univops.universes_of_constr c`
or `EConstr.universes_of_constr sigma (EConstr.of_constr c)` to
calculating before normalizing by `EConstr.universes_of_constr sigma c`.

The remaining changes in #7530, `vernac/classes.ml`,
`vernac/comFixpoint.ml`, and `vernac/record.ml`
look like they change functionality more dramatically,
so I'll put them in a separate commit.

I'd like this to be benchmarked, so we can see if the time increase in hott and odd-order persists with this reduced change set.

### DIFF
--- a/engine/evarutil.ml
+++ b/engine/evarutil.ml
@@ -54,6 +54,20 @@ let new_global evd x =
 (* Expanding/testing/exposing existential variables *)
 (****************************************************)
 
+let finalize ?abort_on_undefined_evars ?(skip_restrict=false) env sigma f =
+  let sigma = minimize_universes sigma in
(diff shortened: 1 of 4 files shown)


jashug:
Benchmarks can be run on branches without an associated PR. Sorry for the noise.

Issue #7631, opened 2018-05-29
native_compute fails to compile an example in Coq 8.8


erikmd:
#### Version

The Coq Proof Assistant, version 8.8.0 (May 2018)
compiled on May 28 2018 14:37:34 with OCaml 4.02.3

#### Operating system

GNU/Linux for amd64

#### Description of the problem

Trying to run the example below (taken from https://github.com/coq/coq/issues/4357#issuecomment-337536667) with native_compute, I get an anomaly:

```coq
Require Import Omega.
Require Import List.

(** A toy example of reflexive tactic around the Fibonacci sequence *)
Fixpoint fib (n : nat) : N :=
  match n with
  | 0 | 1 => 1%N
  | S (S n as n') => (fib n + fib n')%N
  end.

Definition goal := forall n : nat, 0 <= n <= 12 -> (2 | fib (2 + 3 * n))%N.

Lemma sound : forall n : N, (n mod 2 = 0)%N -> (2 | n)%N.
Proof.
now intros n H; apply N.mod_divide.
Qed.

Ltac tac1 :=
match goal with
| [ n : nat, Hn : ?a <= ?n <= ?b |- (2 | fib (2 + 3 * ?n))%N ] =>
  apply sound;
  apply N.eqb_eq;
  let f := fresh "f" in
  pose (f n := fib (2 + 3 * n));
  let H := fresh "H" in
  assert (In (f n) (map f (seq a (S b)))) as H;
  [ apply in_map; apply in_seq; omega
  | fold (f n); generalize (f n) H;
    apply forallb_forall with (f := (fun n => n mod 2 =? 0)%N);
    native_compute; reflexivity]
end.

Theorem thm1 : goal.
Proof.
red; intros n H.
tac1.
Abort.
```

#### Actual result

```sh
$ coqc native.v
File "/tmp/Coq_nativec6ec5f.native", line 10, characters 2-19:
Error: Unbound value const_Nnative_fib
File "./native.v", line 39, characters 0-5:
Warning: Native compiler exited with status 2
[native-compiler-failed,native-compiler]
File "./native.v", line 39, characters 0-5:
Error: Anomaly "Compilation failure."
Please report at http://coq.inria.fr/bugs/.
```

erikmd:
FYI I've tested this example with the other Coq versions I have by hand on my machine, and it triggers the same error with Coq 8.5.3, 8.6.1 & 8.7.0

PR #8058, opened 2018-07-13
Simplify code for [Definition := Eval ...]

### LINKED ISSUE #3690: [Eval compute in] fails to minimize universes

Original bug ID: BZ#3690
From: @JasonGross
Reported version: 8.5

SkySkimmer:
This fixes #3690 for real (it was added to the test suite as closed even though it was wontfix for some reason) by restricting universes after running the Eval.

Note the commit touching nativenorm, I don't know if that's the right fix but without it the test for #7631 breaks with assert false. I don't know if there's a way to trigger it with the current system but it's not good for a function taking econstr to fail when the constr is not evar normal.
I have further simplification in mind which would may make it suboptimal to econstr.to_constr before the eval so I avoided it but if someone strongly thinks it's better I can deal with it.

### DIFF
--- a/pretyping/nativenorm.ml
+++ b/pretyping/nativenorm.ml
@@ -132,15 +132,15 @@ let construct_of_constr_notnative const env tag (mind, _ as ind) u allargs =
   (mkApp(mkConstructU((ind,i),u), params), ctyp)
  
 
-let construct_of_constr const env tag typ =
+let construct_of_constr const env sigma tag typ =
   let t, l = app_type env typ in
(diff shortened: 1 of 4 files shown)


SkySkimmer:
(because github is stupid it's showing the commits in the wrong order, the nativenorm fix should be first)

Zimmi48:
You can avoid this problem by using `--ignore-date` option of `git rebase`.

SkySkimmer:
I'm going to reopen this with another branch name.

PR #8059, opened 2018-07-13
Simplify code for [Definition := Eval ...]

### LINKED ISSUE #3690: [Eval compute in] fails to minimize universes

SkySkimmer:
This fixes #3690 for real (it was added to the test suite as closed even though it was wontfix for some reason) by restricting universes after running the Eval.

Note the commit touching nativenorm, I don't know if that's the right fix but without it the test for #7631 breaks with assert false. I don't know if there's a way to trigger it with the current system but it's not good for a function taking econstr to fail when the constr is not evar normal.
I have further simplification in mind which would may make it suboptimal to econstr.to_constr before the eval so I avoided it but if someone strongly thinks it's better I can deal with it.

### DIFF
--- a/pretyping/nativenorm.ml
+++ b/pretyping/nativenorm.ml
@@ -132,15 +132,15 @@ let construct_of_constr_notnative const env tag (mind, _ as ind) u allargs =
   (mkApp(mkConstructU((ind,i),u), params), ctyp)
  
 
-let construct_of_constr const env tag typ =
+let construct_of_constr const env sigma tag typ =
   let t, l = app_type env typ in
-  match kind t with
+  match EConstr.kind_upto sigma t with
   | Ind (ind,u) -> 
       construct_of_constr_notnative const env tag ind u l
   | _ -> assert false
 
-let construct_of_constr_const env tag typ = 
-  fst (construct_of_constr true env tag typ)
+let construct_of_constr_const env sigma tag typ =
+  fst (construct_of_constr true env sigma tag typ)
 
 let construct_of_constr_block = construct_of_constr false
 
@@ -207,9 +207,9 @@ let rec nf_val env sigma v typ =
       let env = push_rel (LocalAssum (name,dom)) env in
       let body = nf_val env sigma (f (mk_rel_accu lvl)) codom in
       mkLambda(name,dom,body)
-  | Vconst n -> construct_of_constr_const env n typ
+  | Vconst n -> construct_of_constr_const env sigma n typ
   | Vblock b ->
-      let capp,ctyp = construct_of_constr_block env (block_tag b) typ in
+      let capp,ctyp = construct_of_constr_block env sigma (block_tag b) typ in
       let args = nf_bargs env sigma b ctyp in
       mkApp(capp,args)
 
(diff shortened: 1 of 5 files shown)


pretyping/nativenorm.ml
```diff
@@ -132,15 +132,15 @@ let construct_of_constr_notnative const env tag (mind, _ as ind) u allargs =
   (mkApp(mkConstructU((ind,i),u), params), ctyp)
  
 
-let construct_of_constr const env tag typ =
+let construct_of_constr const env sigma tag typ =
```
maximedenes: @SkySkimmer did you check if `vnorm` needs a similar fix?

maximedenes (approved):
The nativenorm change looks ok to me.

pretyping/nativenorm.ml
SkySkimmer: vnorm does
~~~ocaml
  (** This evar-normalizes terms beforehand *)
  let c = EConstr.to_constr ~abort_on_undefined_evars:false sigma c in
  let t = EConstr.to_constr ~abort_on_undefined_evars:false sigma t in
~~~
so no.
I'm going to add vm_compute to that one test file for future safety.

test-suite/bugs/closed/3956.v
```diff
@@ -129,13 +129,13 @@ Module Comodality_Theory (F : Comodality).
       := IdmapM FPM.
     Module cip_FPM := FPM.coindpathsM FPM cmpinv_o_cmp_M idmap_FPM.
     Module cip_FPHM <: HomotopyM FPM cmpM.PM cip_FPM.fhM cip_FPM.fkM.
-      Definition m : forall x, cip_FPM.fhM.m@{i j} x = cip_FPM.fkM.m@{i j} x.
+      Definition m : forall x, cip_FPM.fhM.m x = cip_FPM.fkM.m x.
```
herbelin: I did not understand in which sense this is related to this PR.

herbelin (approved):
Looks much nicer than before!

test-suite/bugs/closed/3956.v
SkySkimmer: This is about this part:

>This fixes #3690 for real (it was added to the test suite as closed even though it was wontfix for some reason) by restricting universes after running the Eval.

herbelin: Ah, ok, you meant #3956, not #3960. Thanks.

SkySkimmer: No, I did mean #3960. The change in the test for #3956 is a side effect of the fix.

herbelin: Sorry, I don't understand, there is no `Eval` in  #3960. And no ` 3960.v` in your commits. I'm confused.

SkySkimmer: Sorry it was #3690 (eval compute in fails to minimize universes).

SkySkimmer:
ping @mattam82

SkySkimmer:
ping @mattam82 ping

mattam82 (approved):
Much nicer indeed, and bravo on the fix thanks to the improved minimization!

mattam82:
Are the ci issues related?

SkySkimmer:
Don't think so, but have a rebase regardless.

SkySkimmer:
Ready to merge

ejgallego (approved):

PR #8062, opened 2018-07-13
Add Z.div_mod_to_quot_rem tactic, put it in zify

JasonGross:
The various (micr)omega tactics now support `Z.div` and `Z.modulo`.

I briefly looked into supporting `Nat.div` and `Nat.modulo`, but the
conversions between `Z.div` and `Nat.div` are defined in `ZArith.Zdiv`,
which depends on `Omega`, which depends on `PreOmega`, which is where
`zify` is defined.

**Kind:** feature

- [x] Entry added in CHANGES.

### DIFF
--- a/doc/sphinx/addendum/micromega.rst
+++ b/doc/sphinx/addendum/micromega.rst
@@ -145,7 +145,7 @@ weakness, the :tacn:`lia` tactic is using recursively a combination of:
 + linear *positivstellensatz* refutations;
 + cutting plane proofs;
 + case split.
-  
+
 Cutting plane proofs
 ~~~~~~~~~~~~~~~~~~~~~~
 
@@ -250,6 +250,8 @@ obtain :math:`-1`. By Theorem :ref:`Psatz <psatz_thm>`, the goal is valid.
 
 .. [#] Support for :g:`nat` and :g:`N` is obtained by pre-processing the goal with
   the ``zify`` tactic.
+.. [#] Support for :g:`Z.div` and :g:`Z.modulo` is obtained by pre-processing the goal with
+  the `Z.div_mod_to_quot_rem` tactic, which is called by `zify`.
 .. [#] Sources and binaries can be found at https://projects.coin-or.org/Csdp
 .. [#] Variants deal with equalities and strict inequalities.
 .. [#] In practice, the oracle might fail to produce such a refutation.
--- a/doc/sphinx/addendum/micromega.rst
+++ b/doc/sphinx/addendum/micromega.rst
@@ -250,8 +250,16 @@ obtain :math:`-1`. By Theorem :ref:`Psatz <psatz_thm>`, the goal is valid.
 
 .. [#] Support for :g:`nat` and :g:`N` is obtained by pre-processing the goal with
   the ``zify`` tactic.
-.. [#] Support for :g:`Z.div` and :g:`Z.modulo` may be obtained by pre-processing the goal with
-  the ``Z.div_mod_to_quot_rem`` tactic after manually running ``zify``.
+.. [#] Support for :g:`Z.div` and :g:`Z.modulo` may be obtained by
+  pre-processing the goal with the ``Z.div_mod_to_equations`` tactic (you may
+  need to manually run ``zify`` first).
+.. [#] Support for :g:`Z.quot` and :g:`Z.rem` may be obtained by pre-processing
+  the goal with the ``Z.quot_rem_to_equations`` tactic (you may need to manually
+  run ``zify`` first).
+.. [#] Note that support for :g:`Z.div`, :g:`Z.modulo`, :g:`Z.quot`, and
+  :g:`Z.rem` may be simultaneously obtained by pre-processing the goal with the
+  ``Z.to_euclidean_division_equations`` tactic (you may need to manually run
+  ``zify`` first).
 .. [#] Sources and binaries can be found at https://projects.coin-or.org/Csdp
 .. [#] Variants deal with equalities and strict inequalities.
 .. [#] In practice, the oracle might fail to produce such a refutation.
--- a/doc/sphinx/addendum/micromega.rst
+++ b/doc/sphinx/addendum/micromega.rst
@@ -251,7 +251,8 @@ obtain :math:`-1`. By Theorem :ref:`Psatz <psatz_thm>`, the goal is valid.
 .. [#] Support for :g:`nat` and :g:`N` is obtained by pre-processing the goal with
   the ``zify`` tactic.
 .. [#] Support for :g:`Z.div` and :g:`Z.modulo` may be obtained by pre-processing the goal with
-  the ``Z.div_mod_to_quot_rem`` tactic after manually running ``zify``.
+  the ``Z.div_mod_to_quot_rem`` tactic after manually running ``zify``.  Note that additionally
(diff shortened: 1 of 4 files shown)


JasonGross:
Queued up a bench at https://ci.inria.fr/coq/view/opam/job/benchmark-part-of-the-branch/493/ (if I did it right)

Zimmi48:
This PR is having a huge impact on the build time of fiat-crypto. It reached the 2h time limit 19 times (with @coqbot restarting the job each time). I should probably disable @coqbot retry for timeouts!

JasonGross:
Hm, interesting.  I will dig into it.  I can also split off the commit at adds the `div_mod_to_quot_rem` tactic to be separate from `zify`, if that'd be useful.

maximedenes:
@JasonGross doesn't that extend `lia` beyond the scope of linear arithmetic?

JasonGross:
@maximedenes Does it?  It will let you prove things like `0 < b -> 0 <= a mod b < b`, and it'll let you prove things like `0 < b -> 0 <= (a mod b) + (c mod b) < b + b`, but there's a sense in which I'd still consider these in the realm of "linear arithmetic".  I think it also lets `lia` prove `(2 * x) / 2 = x`, because we effectively get the equation `2*x = 2*q + r /\ 0 <= r < 2`, and I'd claim that this is also still linear arithmetic.
I think this is essentially all of the power that it adds to `lia`: rather than treating `a mod b` and `a / b` as just opaque `Z` atoms, we also get the equation `(b = 0 /\ a mod b = 0 /\ a / b = 0) \/ (0 <= a mod b < b \/ b < a mod b <= 0)`, plus the equation `a = b * (a / b) + (a mod b)` in the right-hand case, which is useful for `lia` iff `b` is a literal.

(Oh, but, hm, I guess I need to update the tactic to not generalize things like `2/2`, if `lia`/`romega` currently recognize that as `1`; do they?)

maximedenes:
@fajb what's your take on that? Does it qualify as linear arithmetic?

Zimmi48:
https://ci.inria.fr/coq/view/opam/job/benchmark-part-of-the-branch/493/console confirmed the huge impact on fiat-crypto as it was stuck for two days. The rest of the bench shows no significant impact. Could it be an infinite loop?

JasonGross:
Infinite loop: No.  Not unless `lia` or `nia` has an infinite loop inside of them.  It's not terribly surprising that this happens, in some sense, because this tactic adds a bunch of disjunctions to the context, as well as a bunch of new products, and the tactics are non-linear in the number of disjunctions.  I'm going to try a slightly different strategy here, of posing two implications rather than a disjunction.

Is there a way to access the build log of the bench, to figure out which file it got stuck on?

JasonGross:
Is there any description of the performance / time-complexity of `nia`?

fajb:
The support for div and mod proposed by @JasonGross can only improve the expressiveness.
Being picky, this only falls in linear arithmetics if the divisor is a constant.

Regarding complexity, there are at least 2 exponentials.
- 1 is due to the propositional structure of the formula
  lia/nia constructs a naive dnf
- 1 is due to the arithmetic part (fourier elimination)

Dealing with div and mod makes both worse...

If you have some engineering time to spare, we can discuss ways to tame both...

PS: Would it be possible to improve the tactics and substitute the equations on the fly? I think that could help a bit.

fajb:
Stairecase wit: there are 2 (undocumented?) options to control the tactics a bit.
When set, completness is lost but that may be useful to get more predictable and faster failures.

Set Lia Enum false prevents a second enumeration phase that is (rarely) necessary for completeness. 
Set Lia Depth x also limits the proof search.

JasonGross:
I have a bit of engineering time to spare, and would be interested in figuring out how to tame these exponentials.

I guess the fallback plan is to make the tactic available but not part of zify, and document it as a manual preprocessor to run to get completeness over div and mod

fajb:
I do not have easy solutions but ideally here is a workplan:

- Decent Propositional Reasoning
  Ideally, lia/nia should only care about conjunctions of arithmetic propositions
  and leave propositional reasoning to a dedicated tactic.
  `intuition lia` is a candidate but: 
   1- For completeness, it should be taught that arithmetic propositions are decidable.
       (I guess, this is the easy part)
   2- For efficiency, it should be taught  backjumping.
       Backjumping is a standard optimisation of sat-solver but also of tableau-based prover.
       If you case-split; explore case1 and notice that the proof does not need it; do not explore case2

- Arithmetic reasoning
  Ideally, lia/nia should be efficient for deciding conjunctions.
  Replace Fourier-Motzkin  with the Simplex + (branch and bound or cutting planes)
  
Of course, this is easier said than done.

andres-erbsen:
Could the preprocessing tactic be added without making it a part of zify by default so it does not slow down the standard library but can be used in developments that need it? It is a common copy-paste for me.

JasonGross:
I have updated this with a commit that leaves the tactic in, but makes it not part of `nia`.  The documentation alludes to how it can be run.

andres-erbsen:
The `CHANGES` entry probably needs to be updated.

Also, given that all of `div`, `mod`, `quot`, `rem` are in the standard library but this tactic only uses 2 of them, pehaps it should instead be called something like `div_mod_to_equations`?

fajb (approved):
I find it very nice to have support for 'div' and 'mod'.

PS: Does anyone have an idea how zify could be extended in a modular way?  
(Adding a tactic per derived arithmetic function does not seem very user-friendly)\
Could zify - [op1...opn] be a longer-term solution?

doc/sphinx/addendum/micromega.rst
```diff
@@ -250,6 +250,8 @@ obtain :math:`-1`. By Theorem :ref:`Psatz <psatz_thm>`, the goal is valid.
 
 .. [#] Support for :g:`nat` and :g:`N` is obtained by pre-processing the goal with
   the ``zify`` tactic.
+.. [#] Support for :g:`Z.div` and :g:`Z.modulo` may be obtained by pre-processing the goal with
+  the ``Z.div_mod_to_quot_rem`` tactic after manually running ``zify``.
```
jfehrle: This would be clearer without the passive voice.  Something like:

```.. [#] To use :g:`Z.div` and :g:`Z.modulo`, preprocess the goal with  the ``Z.div_mod_to_quot_rem`` tactic after using ``zify``.```

I'm guessing "manually" is not really needed.

jfehrle (approved):
The doc looks OK for form and clarity.  I can't comment on accuracy.

doc/sphinx/addendum/micromega.rst
JasonGross: I used passive voice to mimic the bullet point above.  Saying `manually` is needed, because, in contrast with automatic support for `N` and `nat`, the user must run `zify` explicitly if they want to support div and mod mixed with `N` and `nat`.  However, this is not required in all cases (in particular, if the equations all only mention `Z`).  I don't think your phrasing is fully accurate; you can *use* `Z.div` and `Z.modulo` without the tactics; the only issue is that `nia` and `lia` won't be able to do anything with them.

I have adjusted the end of the sentence to say "(you may need to manually run `zify` first)".

jfehrle: OK

PR #8065, opened 2018-07-13
Remove useless libobject in Implicit Types

maximedenes:
I'm still learning libobject 101, so I'm not sure what to think of this one. This command declares an object that is 'Keep', so will be stored in `.vo` files and "module keep" objects (IIUC). However, it does not have an effect at `Require` or `Import`. So not sure if it does anything else than using memory.

### DIFF
--- a/interp/reserve.ml
+++ b/interp/reserve.ml
@@ -15,8 +15,6 @@ open Util
 open Pp
 open Names
 open Nameops
-open Libobject
-open Lib
 open Notation_term
 open Notation_ops
 open Globnames
@@ -77,14 +75,10 @@ let notation_constr_key = function (* Rem: NApp(NRef ref,[]) stands for @ref *)
   | NRef ref -> RefKey(canonical_gr ref), None
   | _ -> Oth, None
 
-let cache_reserved_type (_,(id,t)) =
+let set_reserved_type (id,t) =
   let key = fst (notation_constr_key t) in
   reserve_table := Id.Map.add id t !reserve_table;
   reserve_revtable := keymap_add key (id, t) !reserve_revtable
-
-let in_reserved : Id.t * notation_constr -> obj =
-  declare_object {(default_object "RESERVED-TYPE") with
-    cache_function = cache_reserved_type }
 
 let declare_reserved_type_binding {CAst.loc;v=id} t =
   if not (Id.equal id (root_of_id id)) then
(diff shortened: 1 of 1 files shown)


maximedenes:
@ppedrot can you have a look at this one? Thanks!

ppedrot (approved):
Seems OK.

ppedrot:
@herbelin Any comments before I merge this PR?

SkySkimmer (approved):
You have my keyboard

herbelin:
> Any comments before I merge this PR?

There is/was an invariant that the contents of tables is strictly under control of registered vernac commands, via the `cache`, `load` and `import` commands.

For the record, this invariant was necessary to replay the effects of commands when backtracking points were not put at each sentence but only at key places such as beginning of section. I'm not fully aware of the current status but it is probably the case that there are currently backtracking points associated to each commands and that replaying objects using the `cache` function is not needed any more (to the exceptions of exported objects in sections and modules).

Even if the invariant is maybe not technically needed now for objects which don't survive sections nor modules, I think we should rather go the other way round, seek for uniformity with other commands and add support for the `Global` attribute.

maximedenes:
> I think we should rather go the other way round, seek for uniformity with other commands and add support for the Global attribute.

I guess that would make sense too.

> There is/was an invariant that the contents of tables is strictly under control of registered vernac commands, via the cache, load and import commands.

I'm not sure what this means exactly. What are "tables"?

> I'm not fully aware of the current status but it is probably the case that there are currently backtracking points associated to each commands and that replaying objects using the cache function is not needed any more (to the exceptions of exported objects in sections and modules).

Yes, that's my understanding of the current situation.

> For the record, this invariant was necessary to replay the effects of commands when backtracking points were not put at each sentence but only at key places such as beginning of section.

So you mean for example it was needed when backtracking to the middle of a closed section? AFAIU this is indeed now handled by the STM.

ppedrot:
So what do we do? If the STM is handling this fine, there is no reason to store the object in the stack, except if we want its effects to survive sections and modules. Does the latter make sense?

herbelin:
> So what do we do? If the STM is handling this fine, there is no reason to store the object in the stack, except if we want its effects to survive sections and modules. Does the latter make sense?

If the backtracking is handling this fine, we can indeed update our invariants and not require that every change to the global state goes through an object. This would then apply to the case when a command does not survive neither to sections nor to modules.

However, I feel that we have more to gain at supporting surviving sections and modules (with an improved support of the Local/Global flag directly visible for the user) than at reducing the size of the code base for cases which are anyway non-standard cases.

ejgallego:
> STM is handling this fine

`s/STM/Summary/`

PR #8068, opened 2018-07-14
[build] Build Coq and plugins with `-strict-sequence`

### LINKED ISSUE #8067: Coq doesn't compile with `-strict-sequence`
Not clear who's at fault here; adding `-strict-sequence` to flags makes compilation fail in different ways.

I am unsure about the usefulness of the flag, but stricter builds envs tend to include it by default so hence reporting.

ejgallego:
Fixes #8067. This is becoming the default in many developments, so it
makes sense to require it too, both for Coq and for Plugins.

### DIFF
--- a/configure.ml
+++ b/configure.ml
@@ -475,6 +475,7 @@ let coq_bin_annot_flag = if !prefs.bin_annot then "-bin-annot" else ""
 (* This variable can be overriden only for debug purposes, use with
    care. *)
 let coq_safe_string = "-safe-string"
+let coq_strict_sequence = "-strict-sequence"
 
 let cflags = "-Wall -Wno-unused -g -O2"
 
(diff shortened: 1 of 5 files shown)


ejgallego:
PR on github is outdated due to github problems, proceed with care.

ejgallego:
Once https://github.com/QuickChick/QuickChick/pull/94 is merged this will be ready to go.

SkySkimmer:
(and this is basically a stronger warning 10)

ejgallego:
> Do we really want to enforce that? We don't for warn-error.

This is becoming a kind of default flag in the OCaml world [at least in default dev profiles] so I indeed propose we anticipate it becoming the default in the future.

IMHO it is indeed good style.

ejgallego:
> IMHO it is indeed good style.

And we don't have any incompatibility.

ejgallego:
For warnings it is a bit more complicated tho.

ppedrot (approved):

SkySkimmer (approved):

ejgallego:
It seems to me that this is really in the domain of Makefile cc: @gares

gares (approved):

gares:
All changes are trivial but for the ones that were reviewed (ie interp and funind related changes)

PR #8069, opened 2018-07-14
Only check overlay extensions on git-tracked files

JasonGross:
This way, when editors leave over temporary files from editing
user-overlays, we don't prevent commits unless they are added to git.

**Kind:** feature / infrastructure.

### DIFF
--- a/dev/tools/check-overlays.sh
+++ b/dev/tools/check-overlays.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
-for f in dev/ci/user-overlays/*
+for f in $(git ls-files "dev/ci/user-overlays/")
 do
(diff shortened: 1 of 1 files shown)


Zimmi48:
@SkySkimmer It looks like you are not listed as official owner for this file.

SkySkimmer (changes requested):
This actually breaks the pre commit hook when committing through magit, because magit calls git with --literal-pathspecs (disabling interpretation of *) which also applies in the hook.
I think the easiest solution is to just `git ls-files "dev/ci/user-overlays/"`.

JasonGross:
> `git ls-files "dev/ci/user-overlays/"`

Changed

SkySkimmer (approved):

Issue #7602, opened 2018-05-25
Plan to get rid of objects without body


Zimmi48:
#7556 added a new warning for objects without a body (generally this is due to the body being incorrectly indented and not recognized as such). Cf. https://github.com/coq/coq/tree/master/doc/sphinx#improper-nesting.
It also introduced the `:undocumented:` flag to put in place of a body when needed (cf. https://github.com/coq/coq/tree/master/doc/sphinx#coq-objects).

However, this warning is currently turned off by default because there are too many occurrences and we cannot address them in one day. What we could do would be to address this warning chapter by chapter and to turn the warning on only for chapters that are free of it (to make sure that it doesn't come back). This would require enhancing the configuration option added by #7556 to allow a finer-grained configuration (chapter by chapter). Any help on this topic from any volunteer would be welcome.

Zimmi48:
@cpitclaudel Shouldn't `prodn` be allowed without a body? And in fact, bodies should probably be made optional for `exn` as well given how frequent it is that they are not documented.

PR #8072, opened 2018-07-15
Fixes for chapters 'Vernacular commands', 'Proof handling' and 'Tactics' of the Reference Manual.

wkolowski:
**Kind:** documentation

This time I did some more serious changes.

The remark that "parentheses are necessary" for functional induction was incorrect - I checked in Coq 8.6, 8.8.1 and 8.9 alpha and they are not.

In the entry for the tactic red, there was a name conflict: t was both the conclusion of the goal and the body of c. I changed the first use of t to T. I also added the error message "No head constant to reduce".

The example given for hnf was incorrect, because forall n : nat, (plus (S n) (S n)) is not even a valid term.

There also seems to be some inconsistencies in the syntax, because for the unfold tactic you can use a modifier "in (type of H)", but for instantiate it is

instantiate ( @num := @term ) in ( Value of @ident )
instantiate ( @num := @term ) in ( Type of @ident )

I fixed the documentation entry for unfold which incorrectly stated that the syntax for unfold is "unfold ident in (Type of ident)".

I also renamed the section "Automatizing" to "Automating".

Some general remarks:
1. I think it would be a good idea to merge the sections "Automation" and "Automating", since they're dealing with similar matters.
2. It would be a good idea to put documentation for all the tactics for inversion together (now functional inversion is apart from ordinary inversion).
3. The section "Classical tactics" seems sad, as it has just one entry for the classical_left tactic. I think it would be better to put classical_left near ordinary left.

Also, some sentences could be reformulated:
"If the conclusion of the type of term does not match the goal and the conclusion is an inductive type isomorphic to a tuple type, then the tuple is (recursively) decomposed and the first component of the tuple of which a non-dependent premise matches the conclusion of the type of ident."

This one is quite impenetrable and I don't even know how to improve it.


doc/sphinx/proof-engine/tactics.rst
```diff
@@ -2804,14 +2804,15 @@ the conversion in hypotheses :n:`{+ @ident}`.
...
+     forall (x:T1) ... (xk:Tk), T
 
-   with :g:`t` :math:`\beta`:math:`\iota`:math:`\zeta`-reducing to :g:`c t`:sub:`1` :g:`... t`:sub:`n` and :g:`c` a
+   with :g:`T` :math:`\beta`:math:`\iota`:math:`\zeta`-reducing to :g:`c t`:sub:`1` :g:`... t`:sub:`n` and :g:`c` a
    constant. If :g:`c` is transparent then it replaces :g:`c` with its
    definition (say :g:`t`) and then reduces
    :g:`(t t`:sub:`1` :g:`... t`:sub:`n` :g:`)` according to :math:`\beta`:math:`\iota`:math:`\zeta`-reduction rules.
 
 .. exn:: Not reducible.
+.. exn:: No head constant to reduce.
```
Zimmi48: Please had `:undocumented:` to these two error messages: see https://github.com/coq/coq/tree/master/doc/sphinx#coq-objects and #7602.

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -2862,7 +2862,7 @@ the conversion in hypotheses :n:`{+ @ident}`.
 
    + A constant can be marked to be unfolded only if applied to enough
      arguments. The number of arguments required can be specified using the
-     ``/`` symbol in the arguments list of the ``Arguments`` vernacular command.
+     ``/`` symbol in the argument list of the ``Arguments`` vernacular command.
```
Zimmi48: A possible improvement would be to replace ``` ``Arguments`` ``` with ``` :cmd:`Arguments` ```

Zimmi48 (commented):
I only went over the changes quickly for now. I'll try to do a more complete review tomorrow and share some thoughts about the Tactics chapter.

maximedenes (approved):
Looks very good, thanks!

Zeimer:
Thanks for the suggestions, I applied them.

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -3072,10 +3075,10 @@ Conversion tactics applied to hypotheses
    listed in this section.
 
    If :n:`@ident` is a local definition, then :n:`@ident` can be replaced by
-   (Type of :n:`@ident`) to address not the body but the type of the local
+   (type of :n:`@ident`) to address not the body but the type of the local
```
Zimmi48: I had missed this: it would be even better to put the verbatim quotes around the whole expression:
```
:n:`(type of @ident)`
```

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -3535,15 +3538,14 @@ Hint databases defined in the Coq standard library
...
-actual content of a database is the collection of the hints declared
+actual content of a database is the collection of hints declared
 to belong to this database in each of the various modules currently
-loaded. Especially, requiring new modules potentially extend a
-database. At Coq startup, only the core database is non empty and can
-be used.
+loaded. Especially, requiring new modules may extend the database.
+At Coq startup, only the core database is nonempty and can be used.
 
 :core: This special database is automatically used by ``auto``, except when
```
Zimmi48: Could use ```:tacn:`auto` ``` to have a link to the tactic definition.

Zimmi48 (approved):
Here are a few more comments of stuff I missed when doing a first quick review. It's very minor so tell me if you don't have time to address them.

Zimmi48:
I agree with your general remarks. In its current state, this chapter looks too much like a very long list of tactics with not so much global consistency. I've considered starting to split things. In particular, I would like to create a chapter about "Automation" gathering information from both this chapter, and the typeclass chapter. But as a first step, just reducing the number of sections by gathering things together like you propose would be a good step forward:

- moving "Controlling automation" inside "Automation"
- "Checking properties of terms" would make more sense as part of the Ltac chapter ("The tactic language")
- moving "Equality and inductive sets" inside "Equality"
- moving "Inversion" inside "Case analysis and induction" like you proposed
- moving `classical_left` and `classical_right` next to `left` and `right` as you proposed
- moving the tactics from "Automatizing" into "Decision procedures"
- removing the last section "Simple tactic macros"

If you want to tackle this big refactoring, let's first merge the present PR and do this in a subsequent one.

Zeimer:
Ok, let's merge it and leave the new suggestions for later (I feel bad doing 2-line commits).

When I finish fixing the chapters by themselves, I will probably do a global consistency check (there are many spelling inconsistencies, like "a hypothesis" vs "an hypotehsis") and I will include them at this point.

herbelin:
> There also seems to be some inconsistencies in the syntax, because for the `unfold` tactic you can use a modifier "in (type of H)", but for `instantiate` it is
> 
> `instantiate ( @num := @term ) in ( Value of @Ident )`
> `instantiate ( @num := @term ) in ( Type of @Ident )`

Indeed! And it happens to be an unnoticed left-over of the syntax of Coq before the revision that happened in Coq 8.0 in 2004. I made #8110 to fix this inconsistency (it includes fixing the reference manual).

PR #8077, opened 2018-07-16
Fix #7291: unify tactic should have more descriptive error messages.

### LINKED ISSUE #7291: unify tactic should have more descriptive error messages
#### Version

8.8+beta1

The `unify` tactic should describe the terms which are not unifiable, possibly elaborating them if they seem to unify with the current printing mode.  I expect the error message of, e.g., `unify 1 2` to be similar to the error message of `let __ := constr:(ltac:(reflexivity) : 1 = 2) in idtac`.

ppedrot:
We simply reuse the same exception as the pretyper instead of raising the generic failure exception in Tactics.unify.

Fixes / closes #7291.

### DIFF
--- a/doc/sphinx/proof-engine/tactics.rst
+++ b/doc/sphinx/proof-engine/tactics.rst
@@ -3976,7 +3976,7 @@ succeeds, and results in an error otherwise.
    This tactic checks whether its arguments are unifiable, potentially
    instantiating existential variables.
 
-.. exn:: Not unifiable.
+.. exn:: Unable to unify @term with @term.
 
 .. tacv:: unify @term @term with @ident
 
--- a/tactics/tactics.ml
+++ b/tactics/tactics.ml
@@ -5062,6 +5062,7 @@ let constr_eq ~strict x y =
 
 let unify ?(state=full_transparent_state) x y =
   Proofview.Goal.enter begin fun gl ->
+  let env = Proofview.Goal.env gl in
   let sigma = Proofview.Goal.sigma gl in
   try
     let core_flags =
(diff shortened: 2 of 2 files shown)


herbelin:
Short and effective. Shall merge this WE if no objections by then.

JasonGross (dismissed):
I'm hesitant to block merging this PR, but note that this makes the documentation of `unify` incorrect, which still [claims that `Not unifiable` is an error message](https://coq.inria.fr/refman/proof-engine/tactics.html#coq:exn.not-unifiable).

herbelin:
OK. @JasonGross, do you intend the [following](https://github.com/herbelin/github-coq/tree/ppedrot-fix-7291%2Bdoc) patch in the documentation of `unify`?

@ppedrot: do you want to do it, or do you want me to add a doc commit, or to squash an update of the doc to your commit, or whatever solution that suits you?

(Incidentally, I made a confusion thinking I had announced this PR for late Monday while I had written WE. Anyway, the countdown is restarted by @JasonGross' remark.)

herbelin:
`Unable to unify @term with @term` is ok to me. I just imitated what was done for the other instance of the same message in the chapter. If no objection, we can then change both occurrences.

My use of `intend` was a mistake, sorry. I rather meant "is this what you have in mind".

ppedrot:
I have added the documentation change suggested by @JasonGross.

ppedrot:
@herbelin should be ready to go.

JasonGross (approved):

JasonGross:
@herbelin Huh, that's subtle.  If you had said "is this patch what you intended", I think I would not have stumbled over the wording at all.  I think that the difference between this phrasing and what you said is that my brain automatically elaborates this phrasing into "is the change in this patch a change that satisfies what you intended to request", while with yours, my brain struggled to find a sensible elaboration: "do you intend this patch" got read as "do you intend to write this patch" (which makes no sense because you already wrote it), and then as "do you intend to apply this patch" (which makes no sense because I'm not the assignee nor the author), and then as "do you intend to suggest this patch" (which makes no sense because you already suggested it).

herbelin:
@jasongross: :smile: (and sorry for the confusion)

@ppedrot: would you like to change the second occurrence of the error message in the refman so that they uniformly follow the same format? (Or maybe would you prefer this is done in a separate PR?)

ppedrot:
@herbelin Let's open another PR, this one is only about the `unify` tactic.

herbelin:
> Let's open another PR, this one is only about the unify tactic.

OK, see #8148.

PR #8079, opened 2018-07-16
Document the automatic use of the rebase label.

Zimmi48:
**Kind:** documentation


SkySkimmer:
I didn't know you could do link abbreviations like that in mardown.

Zimmi48:
> I didn't know you could do link abbreviations like that in mardown.

I learned this only recently. Otherwise, I would have used it more.

maximedenes (approved):

PR #8086, opened 2018-07-17
Improved chapter 'The tactic language' of the Reference Manual.

wkolowski:
**Kind:** documentation

Most of the changes are typos and grammar errors, but I made one significant change in the entry for the tactic fail: I removed a sentence which seemed false for me, reworded the description of the tactic gfail and added a lengthy example to make things clearer.

But there are still some problems with the description of fail:

"These variants fail with an error message or an error level even if there are no goals left. Be careful however if Coq terms have to be printed as part of the failure: term construction always forces the tactic into the goals, meaning that if there are no goals when it is evaluated, a tactic call like let x := H in fail 0 x will succeed."

This remark is bad. Not only the phrase "term construction forces that tactic into the goals" is unclear, but also for me the tactic "let x := 42 in fail 0 x" fails with message "Error: Tactic failure: 42." when there are unsolved goals and fails with message "Error: No such goal." when all goals have been solved. I couldn't make this tactic succeed.

I think that someone should rewrite the documentation for fail and gfail from scratch.

Other remarks:

1. "We can restrict the application of a tactic to a subset of the currently focused goals"

This sentence suggests that by default all goals are focused and we can only unfocus some. I think that the "focusing" hasn't been explained well enough (but I read this part of the manual a year ago, maybe it got improved in the meantime).

2. The entry "Filling a term context" is bad.

Not only there's a blatant grammar error "This expression evaluates replaces the hole", but the whole construct is badly documented. It would be nice to have an example to see what it's really doing.

Moreover, there was an undocumented error message: "Unbound context identifier ident". What's even more, the error message is slightly bugged, because for the tactic 

let x := context ident [x] in x.

it displays "Unbound context identifierident." (without the obvious space).

This left me wondering: is this "context" construct different from the one that can be used inside "match context" branches?

### DIFF
--- a/doc/sphinx/proof-engine/ltac.rst
+++ b/doc/sphinx/proof-engine/ltac.rst
@@ -572,7 +572,9 @@ Failing
    .. tacv:: gfail
       :name: gfail
 
-      This variant fails when used after :n:`;` even if there are no goals left (see the example).
+      This variant fails even when used after :n:`;` and there are no goals left.
+      Similarly, ``gfail`` fails even when used after ``all:`` and there are no
+      goals left. See the example for clarification.
 
    .. tacv:: gfail {* message_token}
 
@@ -603,6 +605,9 @@ Failing
          Proof. trivial. fail. Abort.
 
          Goal True.
+         Proof. trivial. all: fail. Qed.
+
+         Goal True.
          Proof. gfail. Abort.
 
          Goal True.
@@ -610,6 +615,9 @@ Failing
 
          Goal True.
          Proof. trivial. gfail. Abort.
+
+         Goal True.
+         Proof. trivial. all: gfail. Abort.
 
 Timeout
 ~~~~~~~
@@ -921,7 +929,10 @@ produce subgoals but generates a term to be used in tactic expressions:
    value of :n:`@ident` by the value of :n:`@expr`.
 
    .. exn:: Not a context variable.
+      :undocumented:
+
    .. exn:: Unbound context identifier @ident.
+      :undocumented:
 
 Generating fresh hypothesis names
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -1212,6 +1223,7 @@ following:
 +-----------------+-----------------------------------------------+
 
 .. exn:: Debug mode not available in the IDE
+   :undocumented:
 
 A non-interactive mode for the debugger is available via the option:
 
--- a/doc/sphinx/proof-engine/ltac.rst
+++ b/doc/sphinx/proof-engine/ltac.rst
@@ -10,8 +10,8 @@ This chapter gives a compact documentation of |Ltac|, the tactic language
 available in |Coq|. We start by giving the syntax, and next, we present the
 informal semantics. If you want to know more regarding this language and
 especially about its foundations, you can refer to :cite:`Del00`. Chapter
-:ref:`detailedexamplesoftactics` is devoted to giving examples of use of this
-language on small but also with non-trivial problems.
+:ref:`detailedexamplesoftactics` is devoted to giving small but nontrivial
+use examples of this language.
 
 .. _ltac-syntax:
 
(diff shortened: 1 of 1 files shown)


Zimmi48:
Here is an example showing the issue that the manual was talking about:

```
Goal True.
easy.
all: let x := constr:(S 0) in gfail 0 x.
```

(this succeeds)

One key ingredient is that you need the `all:` selector. Otherwise the default selector `1:` is used, and indeed, there is no such goal.

It can sound strange that using the `all:` selector will make most tactics succeed when there are no goals left (except `gfail` basically) but it kind of make sense when used like this for instance:

```
Goal True.
easy.
all:let n := numgoals in guard n=0.
```

So this explains what the doc meant, but this doesn't make the doc good. We could at the very least try to make this sentence into a separate note, including this example.

> "We can restrict the application of a tactic to a subset of the currently focused goals"

I think the focusing is indeed badly explained in the manual. Let me explain it quickly to you and maybe you'll find the time to make this into an improvement to the doc.
If a tactic creates several sub-goals, then you have several focused goals. You can use bullets or focusing brackets to focus on just one of them. When you have several focused goals, you can choose to apply a tactic to only one of them (the default goal selector being an implicit `1:`) or to several of them (`1-3:`) or to all of them (`all:`). It is true that this sentence sounds like the default is `all:` and not `1:`.

For the `context` question, I need some Ltac-expert to answer (@ppedrot?)

doc/sphinx/proof-engine/ltac.rst
```diff
@@ -546,15 +548,13 @@ Failing
 
    This is the always-failing tactic: it does not solve any
    goal. It is useful for defining other tacticals since it can be caught by
-   :tacn:`try`, :tacn:`repeat`, :tacn:`match goal`, or the branching tacticals. The
-   :tacn:`fail` tactic will, however, succeed if all the goals have already been
-   solved.
```
Zimmi48: Rather than removing this unclear sentence, another possibility would be to explain what is meant more precisely: `all: fail` always fails unless there are not goals left. Similarly, and probably more useful, `easy; fail` will succeed if `easy` succeeds (while `easy; gfail` will always fail). Actually, you have already added this info in the `gfail` description, so maybe it's enough.

doc/sphinx/proof-engine/ltac.rst
```diff
@@ -572,7 +572,7 @@ Failing
    .. tacv:: gfail
       :name: gfail
 
-      This variant fails even if there are no goals left.
+      This variant fails when used after :n:`;` even if there are no goals left (see the example).
```
Zimmi48: I would rather say "This variant fails even when used after `n` and there are no goals left"

doc/sphinx/proof-engine/ltac.rst
```diff
@@ -582,10 +582,35 @@ Failing
...
+      .. coqtop:: all
+
+         Goal True.
+         Proof. fail. Abort.
+
+         Goal True.
+         Proof. trivial; fail. Qed.
+
+         Goal True.
+         Proof. trivial. fail. Abort.
```
Zimmi48: This example doesn't seem very useful given that there is nothing specific to `fail` here. Unless you add `all:` in front of `fail`.

doc/sphinx/proof-engine/ltac.rst
```diff
@@ -896,6 +921,7 @@ produce subgoals but generates a term to be used in tactic expressions:
    value of :n:`@ident` by the value of :n:`@expr`.
 
    .. exn:: Not a context variable.
+   .. exn:: Unbound context identifier @ident.
```
Zimmi48: Would be great if you could add `:undocumented:` options to these two `exn`.

doc/sphinx/proof-engine/ltac.rst
```diff
@@ -1185,6 +1211,8 @@ following:
 | r string:       | advance up to the next call to “idtac string” |
 +-----------------+-----------------------------------------------+
 
+.. exn:: Debug mode not available in the IDE
```
Zimmi48: Same remark.

doc/sphinx/proof-engine/ltac.rst
cmangin: It looks like the `Not a context variable` one is not even implemented actually.
```coq
Goal True.
let ident := idtac in
let x := context ident [O] in x.
(* Anomaly "Uncaught exception Ltac_plugin.Taccoerce.CannotCoerceTo("a term context")." *)
```

cmangin:
About contexts, there is actually an example, but in the "Pattern matching on terms" section. In both cases `context` refers to the same construct, but in one case (`match`) it introduces a name binding a new context, and in the other case (`let ... in`) it fills a preexisting context and returns a term.

Zeimer:
I applied the suggestions, expanded the description of gfail and added two cases to the example using all: so that it's even more verbose now :)

Zimmi48 (approved):

PR #8089, opened 2018-07-18
Remove declare_object for SsrHave NoTCResolution.

maximedenes:
IIUC, this was a hack to make `Set SsrHave NoTCResolution` behave like
`Global Set SsrHave NoTCResolution`. I don't think it is needed (just
let the user write the desired locality), but if it is, the right way of
doing it is to let clients of Goptions specify a default locality.

- [x] Entry added in CHANGES.


gares (approved):

gares:
looks good, but as you say you have to add a line to CHANGES

maximedenes:
`CHANGES` entry added, should be ready modulo CI.

PR #8090, opened 2018-07-18
Fix check that sections are all-mono or all-polymorphic

SkySkimmer:
Without this commit the following fails at `End Bar.`:
~~~coq
Unset Universe Polymorphism.
Section Foo.
  Variable A : Type.

  Section Bar.
    Set Universe Polymorphism.
    Universe i.
    Definition foo := A -> Type@{i}.
  End Bar.
End Foo.
Print foo.
~~~

The interaction of universes and sections could use some improvement but let that not stand in the way of this patch.

### DIFF
--- a/library/lib.ml
+++ b/library/lib.ml
@@ -428,24 +428,27 @@ let add_section () =
   sectab := ([],(Names.Cmap.empty,Names.Mindmap.empty),
                 (Names.Cmap.empty,Names.Mindmap.empty)) :: !sectab
 
-let check_same_poly p vars =
-  let pred = function Context _ -> p = false | Variable (_, _, poly, _) -> p != poly in
-  if List.exists pred vars then
-    user_err Pp.(str  "Cannot mix universe polymorphic and monomorphic declarations in sections.")
+let check_section_poly p =
+  let check_one (vars,_,_) =
+    let pred = function Context _ -> p = false | Variable (_, _, poly, _) -> p != poly in
+    if List.exists pred vars then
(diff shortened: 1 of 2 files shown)


Zimmi48:
And with this commit, where does it fail?

SkySkimmer:
See test suite change.

mattam82:
It could be checked in the Set function for that option indeed.

JasonGross:
Shouldn't it be possible to mix monomorphic and polymorphic things in a section as long as either there are no section-local variables/universes, or the definitions are completely disjoint?

E.g.,
```coq
Section foo.
  Monomorphic Definition a := nat.
  Polymorphic Definition b := nat.
  Set Universe Polymorphism.
  Variable x : nat.
  Definition c := nat.
  Unset Universe Polymorphism.
  Definition d := nat.
End foo.
```

Also, I think that whatever happens, (un)setting universe polymorphism as the first line in a top-level section should continue to work fine, and we should possibly test this in the test-suite.

JasonGross:
Is this code that used to work that would now fail?
```coq
Unset Universe Polymorphism.
Section Foo.
  Variable A : Type.

  Section Bar.
    Local Set Universe Polymorphism.
    Variable B : Type.
    Definition foo := B.
  End Bar.
  Definition bar := A.
End Foo.
Print foo.
Print bar.
```

Zimmi48:
Seems like the discussion on this PR hasn't converged yet. Given that 8.8.2 will happen soon, it won't probably be part of it.

PR #8094, opened 2018-07-18
Remove `Automatic Coercions Import` option.

maximedenes:
This option made the Coercions command follow non-standard scoping rules
(effect on `Require` rather than `Import`). It was already marked for deletion
in 8.8.

- [X] Corresponding documentation was added / updated (including any warning and error messages added / removed / modified).
- [X] Entry added in CHANGES.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -149,6 +149,8 @@ Vernacular commands
 
 - `Hypotheses` and `Variables` can now take implicit binders inside sections.
 
+- Removed deprecated option `Automatic Coercions Import`.
+
 Tools
 
 - The `-native-compiler` flag of `coqc` and `coqtop` now takes an argument which can have three values:
--- a/doc/sphinx/addendum/implicit-coercions.rst
+++ b/doc/sphinx/addendum/implicit-coercions.rst
@@ -322,21 +322,8 @@ are also forgotten.
 Coercions and Modules
 ---------------------
 
-.. flag:: Automatic Coercions Import
-
-   Since |Coq| version 8.3, the coercions present in a module are activated
-   only when the module is explicitly imported. Formerly, the coercions
-   were activated as soon as the module was required, whether it was
-   imported or not.
-
-   This option makes it possible to recover the behavior of the versions of
-   |Coq| prior to 8.3.
-
-.. warn:: Coercion used but not in scope: @qualid. If you want to use this coercion, please Import the module that contains it.
-
-   This warning is emitted when typechecking relies on a coercion
-   contained in a module that has not been explicitely imported. It helps
-   migrating code and stop relying on the option above.
+The coercions present in a module are activated only when the module is
+explicitly imported.
 
 Examples
 --------
--- a/pretyping/classops.ml
+++ b/pretyping/classops.ml
@@ -120,9 +120,6 @@ let class_tab =
 let coercion_tab =
   Summary.ref ~name:"coercion_tab" (CoeTypMap.empty : coe_info_typ CoeTypMap.t)
 
-let coercions_in_scope =
-  Summary.ref ~name:"coercions_in_scope" GlobRef.Set_env.empty
-
 module ClPairOrd =
 struct
   type t = cl_index * cl_index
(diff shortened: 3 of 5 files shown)


ppedrot (approved):

maximedenes:
I think this should be integrated only once a version has been released with https://github.com/coq/coq/pull/8104

maximedenes:
> This explains the needs:discussion label I guess

Well, it is rather than the deprecation warning was not implemented properly...

ppedrot:
@maximedenes Now that #8104 is part of 8.9, you should probably rebase this PR so that we can fix CI devs and merge it.

Zimmi48:
>@maximedenes Now that #8104 is part of 8.9, you should probably rebase this PR so that we can fix CI devs and merge it.

Careful when you say that a PR is part of 8.9. It hasn't been backported yet and @silene just postponed it to 8.10. I have alerted him by private e-mail that this particular PR really needed to be backported to 8.9 though.

maximedenes:
> @silene just postponed it to 8.10

How can we (PR authors) be notified when that happens?

Zimmi48:
@maximedenes I think the best solution is simply to have @coqbot do the notification. I will add this feature today.

ppedrot:
@maximedenes The depending PR is definitely both in master and 8.9 now, so you should rebase this one and urge downstream devs to fix their code.

ppedrot:
@maximedenes pingy-ping. This is, I believe, an important and urgent change.

maximedenes:
@ppedrot This should now be ready. Since you think it is an important change, would you mind pinging the formal-topology authors so that they fix the warning(s)? Thanks a lot!

Zimmi48:
formal-topology doesn't look like an active project anymore. Maybe it's time to reconsider it being in Coq's CI. Maybe we could suggest to the author to submit their work to coq-community/corn so that it is maintained there instead...

ppedrot:
@maximedenes can you rebase this PR? Should be easier to merge now that formal topology is out of the way.

vbgl:
This breaks corn quite fundamentally. They have a pattern for notations that should now be also used for coercions: some modules are not meant to be imported and define a “notations” submodule; this submodule only is imported by clients.

With this PR, coercions defined in the non-imported module are no longer visible.

Zimmi48:
>They have a pattern for notations that should now be also used for coercions

That sounds like a reasonable change to do but might not be trivial.

vbgl:
I’ve started to address the issue of math-classes and Corn. I cannot understand what’s going on in the `ode/SimpleIntegration.v` file: the coercions that are inserted with and without this PR are very different. Suggestions welcome.

vbgl:
I think this is ready. Ping assignee @ppedrot

ppedrot:
What about the CI failure? Is this spurious?

ejgallego:
> What about the CI failure? Is this spurious?

Yes @ppedrot

ppedrot:
Let's merge then.

herbelin:
For the record, adapting the Coq contribution `algebra`, as well as `lin-alg` and `or-stab` which depend on it, is not easy (note: I consider these contributions as samples of what users could be confrounted to in general).

I have a first version which works by making explicit a sequence of individual `Export` so as to control the order of declaration of coercions, but this is quite heavy (all files have virtually to do the same long sequence of explicit `Export`).

Somehow, the problem is that the order of coercions is immutable, so that, when an ambiguous path conflict arises, we are constrained to keep the path defined first.

I'm not fully aware of what has already been proposed yet but the following approach may help to improve the situation.

If we had a `Remove Coercion` command, we could reset the state of coercion paths, add a new coercion which does not trigger a clash, and re-add the first coercion in a second time, thus forcing the conflict to be resolved differently.

Similarly, if we had a `Remove Path` command, we could prevent a conflict.

If redeclaring coercions were allowed and if conflicts were resolved by taking the new coercion path, we could also update the state by re-issuing commands such as `Coercion` or re-issuing some `Export` having an effect on coercions. (*)

[Incidentally, we could also improve the genesis of coercion paths (they sometimes have loops...). And improve the printing as well (it may be the case that a sequence of coercions is not printed while the corresponding path is in fact an ambiguous path which would not be used back at parsing time, see for instance #5462).]

Anyway, for sure, the "logic" of `Require`, `Export` and `Import` needs we work at it. So, thanks, @maximedenes, to take care of it.

Issue #10825 discussed optimization of imports, which is wrong for coercions, as noticed in #10825.

I don't know how users addressed the coercion path changes in practice when migrating to 8.10, but my feeling is that we should provide a way to have a better control on coercions, at least for 8.11. This could apply to canonical projections as well.

I'd be tempted to explore (*), which would presumably also solve #9584, but I was wondering about the general opinion.

vsiles:
@herbelin Hi ! By any chance can you share your diff that fixes `algebra` for 8.10 or 8.11, even if it's a temporary one ?

herbelin:
Hi too @vsiles! The version [here](https://github.com/herbelin/algebra/releases/tag/v8.10.0) compiles with 8.10 but 8.11 apparently fails. Hope it helps though...

PR #8096, opened 2018-07-19
Higher-level libobject API for objects with fixed scopes

maximedenes:
This is the very first step of a libobject cleanup. It should not change semantics.

This is still work in progress.

The next steps will be:
- A similar API for objects with locality
- Studying the few non-standard objects remaining
- Writing or fixing discharge functions for many objects
- Starting to unify the behaviors (either by making the meaning of `Local` and `Global` uniform, or by providing new attributes for scoping
- Removing "superglobals" (effects on `Require`)

This is essentially a libobject change, so the natural assignee is @gares (since @silene seems offline).

### DIFF
--- a/interp/declare.ml
+++ b/interp/declare.ml
@@ -445,11 +445,9 @@ let assumption_message id =
 (** Monomorphic universes need to survive sections. *)
 
 let input_universe_context : Univ.ContextSet.t -> Libobject.obj =
-  declare_object
-    { (default_object "Monomorphic section universes") with
-      cache_function = (fun (na, uctx) -> Global.push_context_set false uctx);
-      discharge_function = (fun (_, x) -> Some x);
-      classify_function = (fun a -> Dispose) }
+  declare_object @@ local_object "Monomorphic section universes"
+    ~cache:(fun (na, uctx) -> Global.push_context_set false uctx)
+    ~discharge:(fun (_, x) -> Some x)
 
 let declare_universe_context poly ctx =
   if poly then
--- a/library/libobject.mli
+++ b/library/libobject.mli
@@ -119,6 +119,51 @@ val classify_object : obj -> obj substitutivity
 val discharge_object : object_name * obj -> obj option
 val rebuild_object : obj -> obj
 
+(** Higher-level API for objects with fixed scope.
+
+- Local means that the object cannot be imported from outside.
+- Global means that it can be imported (by importing the module that contains the
+object).
+- Superglobal means that the object survives to closing a module, and is imported
+when the file which contains it is Required (even without Import).
+- With the nodischarge variants, the object does not survive section closing.
+  With the normal variants, it does.
+
+We recommend to avoid declaring superglobal objects and using the nodischarge
+variants.
+*)
+
+val local_object : string ->
+  cache:(object_name * 'a -> unit) ->
+  discharge:(object_name * 'a -> 'a option) ->
+  'a object_declaration
+
+val local_object_nodischarge : string ->
+  cache:(object_name * 'a -> unit) ->
+  'a object_declaration
+
+val global_object : string ->
+  cache:(object_name * 'a -> unit) ->
+  subst:(Mod_subst.substitution * 'a -> 'a) option ->
+  discharge:(object_name * 'a -> 'a option) ->
+  'a object_declaration
+
+val global_object_nodischarge : string ->
+  cache:(object_name * 'a -> unit) ->
+  subst:(Mod_subst.substitution * 'a -> 'a) option ->
+  'a object_declaration
+
+val superglobal_object : string ->
+  cache:(object_name * 'a -> unit) ->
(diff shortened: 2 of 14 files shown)


library/libobject.mli
```diff
@@ -113,6 +113,51 @@ val classify_object : obj -> obj substitutivity
...
+We recommend to avoid declaring superglobal objects and using the nodischarge
+variants.
+*)
+
+val local_object : string ->
+  cache:(Libnames.object_name * 'a -> unit) ->
+  discharge:(Libnames.object_name * 'a -> 'a option) ->
+  'a object_declaration
+
+val local_object_nodischarge : string ->
```
SkySkimmer: Why not make the discharge argument optional?

maximedenes:
What this PR does it basically the first step towards making obvious how to answer https://github.com/coq/coq/wiki/Default-Locality-Behavior-of-Commands (in the code)

ejgallego:
Let me very quickly suggest that you implement a `Print Objects` command so we can see the status of libobject at any time; this should be a debug command and could go away at some point.

maximedenes:
> Let me very quickly suggest that you implement a Print Objects command so we can see the status of libobject at any time; this should be a debug command and could go away at some point.

It would dump all the registered libobjects and their kind (local, global, superglobal or ad-hoc), right?

ejgallego:
> It would dump all the registered libobjects and their kind (local, global, superglobal or ad-hoc), right?

In fact the more it can dump the more we know the inversion of control is going away!

I'd like in fact to have this to understand a bit better how objects are sent to workers, etc...

library/libobject.mli
```diff
@@ -113,6 +113,51 @@ val classify_object : obj -> obj substitutivity
...
+  subst:(Mod_subst.substitution * 'a -> 'a) option ->
+  'a object_declaration
+
+val superglobal_object : string ->
+  cache:(Libnames.object_name * 'a -> unit) ->
+  subst:(Mod_subst.substitution * 'a -> 'a) option ->
+  discharge:(Libnames.object_name * 'a -> 'a option) ->
+  'a object_declaration
+
+val superglobal_object_nodischarge : string ->
```
gares: How can something be superglobal but not survive sections?

gares (approved):

gares:
I like the approach, the code change in most cases is trivial hence I think the list of owners is a bit too long.
Only the owners of libobject should be concerned.

library/libobject.mli
maximedenes: Do you mean "how" or "why"?

ppedrot (approved):
Dunno whether I'm supposed to be an owner, but I do like the approach.

maximedenes:
> This is essentially a libobject change, so the natural assignee is @gares (since @silene seems offline).

@gares ping

gares:
I'll be back on wednesday

maximedenes:
> I'll be back on wednesday

@gares ping

gares:
There is still a needs progress label, and I also think you should rebase.

gares:
ping @maximedenes

maximedenes:
@gares rebased

ejgallego (commented):
This seems like a core API that has changed, however no entry in `dev/doc` ?

maximedenes:
> This seems like a core API that has changed, however no entry in dev/doc ?

Right, I forgot to do it. It'll have to wait, though.

interp/declare.ml
```diff
@@ -474,11 +474,10 @@ let assumption_message id =
...
 
 let input_universe_context : Univ.ContextSet.t -> Libobject.obj =
-  declare_object
-    { (default_object "Monomorphic section universes") with
-      cache_function = (fun (na, uctx) -> Global.push_context_set false uctx);
-      discharge_function = (fun (_, x) -> Some x);
-      classify_function = (fun a -> Dispose) }
+  declare_object @@ local_object "Monomorphic section universes"
+    ~cache:(fun (na, uctx) -> Global.push_context_set false uctx)
+    ~subst:None
```
SkySkimmer: No `subst` argument to local_object ^^

maximedenes: Ok, I'm even too tired to get a commit right, so I should clearly not be doing any technical work.

amahboubi:
HI @maximedenes, my review is requested here but I do not feel competent enough on this code. I like the announced objective, but cannot make any useful comment at this stage.

gares:
ping @maximedenes the doc that is missing does not seem to be a daunting task, your should be able to reuse the .mli comment.

gares:
ping @maximedenes

maximedenes:
> ping @maximedenes

Again, this is not on the top of my list at the moment, it will have to wait as written above.

maximedenes:
> Well, the last message from you was on the 29th of August ;-)

Sure, no problem in pinging. I simply meant that it was still on my radar but we are very active on topics with higher priority.

maximedenes:
Rebased and added documentation. Should be ready modulo CI.

PR #8055, opened 2018-07-13
Fast accumulator check in native compilation

ppedrot:
This PR replaces the call to `Nativevalues.is_accu` function in native compilation of fixpoints by a trivial ML pattern-matching. The latter is way faster than the former, as it is just a jump, while the `is_accu` function needs to call low-level C functions that perform a lot of useless checks.

The same trick could be used for cofixpoints, but I'll leave that to another PR.

This mitigates #8019.

@maximedenes is without a doubt the natural assignee.

PR #8097, opened 2018-07-19
Use more efficient accu check for cofix unfolding in native compilation.

ppedrot:
Same patch as #8055 but now for cofixpoints instead of fixpoints.


maximedenes (approved):
I forgot, is this tested in the test suite?

ppedrot:
@maximedenes Since the merge of #8000, it should be indeed.

maximedenes:
@ppedrot it seems this needs a rebase

PR #8101, opened 2018-07-19
Remove ClosedModule and ClosedSection from libstack

maximedenes:
Seems unused and probably holds a lot of pointers.

### DIFF
--- a/library/lib.ml
+++ b/library/lib.ml
@@ -27,11 +27,10 @@ type node =
   | CompilingLibrary of object_prefix
   | OpenedModule of is_type * export * object_prefix * Summary.frozen
   | OpenedSection of object_prefix * Summary.frozen
-  | ClosedSection of library_segment
 
-and library_entry = object_name * node
+type library_entry = object_name * node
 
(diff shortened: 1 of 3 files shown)


maximedenes:
> I don't see anything wrong with this commit, but I am wondering whether these entries might have been used by some external tool. In that case, dropping the library_segment arguments from the entries but keeping the constructors might make sense.

Unless we have a concrete evidence of that (the CI doesn't show anything like that), I would prefer to clean the code.

maximedenes:
@silene can you be the assignee? Failures are spurious, AFAICT.

Zimmi48:
@maximedenes It would be better if you rebased so that your branch is actually tested on Equations.

maximedenes:
Rebase done.

maximedenes:
> @silene can you be the assignee? Failures are spurious, AFAICT.

@silene ping

maximedenes:
@gares can you take this one too?

gares (approved):

PR #8102, opened 2018-07-20
Fix #8043: Unsafe assignment in checker.

ppedrot:
We use a dedicated mutable constructor to perform a Landin knot.

### DIFF
--- a/checker/validate.ml
+++ b/checker/validate.ml
@@ -85,6 +85,7 @@ let rec val_gen v ctx o = match v with
   | Fail s -> fail ctx o ("unexpected object " ^ s)
   | Annot (s,v) -> val_gen v (ctx/CtxAnnot s) o
   | Dyn -> val_dyn ctx o
+  | Proxy { contents = v } -> val_gen v ctx o
 
 (* Check that an object is a tuple (or a record). vs is an array of
    value representation for each field. Its size corresponds to the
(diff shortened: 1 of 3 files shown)


gares:
Can't you just revert 578d73a8b8eb6623bfa ? I really don't see any use for that (in votour I mean)

gares:
Said otherwise: the code in that file should be generated by a ppx, spending time on it seems really futile, especially for data that is not even used by the checker.

ppedrot:
@gares I guess I could, I think I introduced that for debugging. That said, the current PR is helpful even in general, because it allows to abstract over the OCaml limitation that recursive data structures need to be syntactically constructors.

ppedrot:
@gares I remember why I introduced that: I was debugging memory allocation of vio files, and for that you need to observe the STM data structures. Does it really hurt to have it here?

gares:
Well, I'd say no it does not hurt, but there was a bugreport and your fix... hence it happens to have some cost to keep this code around.

herbelin:
Any conclusion to this discussion? In any case, the only official reviewer mentioned should assign it to him.

ppedrot:
@maximedenes could you take care of this one?

maximedenes:
> @maximedenes could you take care of this one?

Sure, will do.

maximedenes (approved):
The `vi`-related memory issues are far from solved, so I think it makes sense to keep the debugging code meanwhile.

Zimmi48:
@maximedenes Here we go with the notification you requested in https://github.com/coq/coq/pull/8094#issuecomment-426231518 (this was only a test and will happen every time in the future).

PR #8103, opened 2018-07-20
Coercions cleanup: use GlobRef.t instead of constr

maximedenes:


plugins/firstorder/ground.ml
```diff
@@ -13,17 +13,16 @@ open Formula
...
-    try
-      let kn= fst (destConst (Classops.get_coercion_value coe)) in
-	predref:=Names.Cpred.add kn !predref
-    with DestKO -> ()
+  let f coe =
+    match coe.Classops.coe_value with
+    | ConstRef c -> predref := Names.Cpred.add c !predref
+    | _ -> ()
   in
     List.iter f (Classops.coercions ());
```
SkySkimmer: Why don't we fold through that list instead of itering with a ref?

plugins/ssr/ssrvernac.ml4
```diff
@@ -359,13 +358,12 @@ let coerce_search_pattern_to_sort hpat =
...
   let coerce hp coe_index =
-    let coe = Classops.get_coercion_value coe_index in
+    let coe_ref = coe_index.Classops.coe_value in
     try
-      let coe_ref = global_of_constr coe in
       let n_imps = Option.get (Classops.hide_coercion coe_ref) in
       mkPApp (Pattern.PRef coe_ref) n_imps [|hp|]
-    with _ ->
-    errorstrm (str "need explicit coercion " ++ pr_constr_env env sigma coe ++ spc ()
+    with Not_found | Option.IsNone ->
```
SkySkimmer: What raises Not_found here?

pretyping/classops.ml
```diff
@@ -298,12 +293,9 @@ let lookup_path_to_sort_from env sigma s =
...
-  let c, _ =
-    Reductionops.whd_all_stack env Evd.empty (EConstr.of_constr coe.coe_value)
-  in
-  match EConstr.kind Evd.empty (** FIXME *) c with
-  | Construct (cstr,u) ->
-      (cstr, Inductiveops.constructor_nrealargs cstr -1)
+  match coe.coe_value with
+  | ConstructRef c ->
+      (c, Inductiveops.constructor_nrealargs c -1)
   | _ ->
```
SkySkimmer: Here we used to reduce, so this function would still succeed on `Coercion s := S.` no? I don't know how it's used though.

SkySkimmer (commented):
Please check that it's OK to drop the reduction in get_coercion_constructor.

pretyping/classops.ml
maximedenes: Right, this should be fixed.

plugins/ssr/ssrvernac.ml4
maximedenes: hide_coercion, potentially

plugins/firstorder/ground.ml
maximedenes: What does it change?

SkySkimmer:
It's cleaner to avoid using a ref all else being equal

Gaëtan Gilbert

maximedenes:
> It's cleaner to avoid using a ref all else being equal

I would use `fold` myself there, but it seems to me like a matter of taste.

pretyping/classops.ml
maximedenes: @SkySkimmer any suggestion on what is the right API to reduce in non-constructor cases?

ejgallego:
> I would use fold myself there, but it seems to me like a matter of taste.

A `ref` tends to have more cost / be harder to optimize by the compiler [usually if optimization kicks in both tend to produce similar performing code AFAIU].

gares (approved):

gares:
I think I cover most of the patch (ssr, pretyping) and printing is trivial.
For firstorder I guess we need @PierreCorbineau

pretyping/classops.ml
```diff
@@ -297,31 +292,25 @@ let lookup_path_to_fun_from env sigma s =
 let lookup_path_to_sort_from env sigma s =
   apply_on_class_of env sigma s lookup_path_to_sort_from_class
 
+let mkNamed = function
```
gares: This one should really be somewhere

gares:
I think I repaired the code, but I also think that the feature should be removed.

What is does:
```
Coercion somenat := Some (A:=nat).
Check (fun x => match x with
                | None => 0
                | O => 0
                | S n => n
                end).
```
In the elaboration of the match (on yet to be typed terms) it tries to see if a coercion can be
injected around the patterns (if the body of the coercion is a constructor...).
Of course the thing is not only surprising, but also quite broken (eg if you put the `O` branch
first it does not find the coercion).

I think we should put a deprecation warning in the branch of the code I've just repaired and then
purge it. AFAIK it is not even documented, so we could also try to purge it directly.

pretyping/classops.ml
ppedrot: No, because it is disregarding the universe variables. We shouldn't make users believe that this is fine.

maximedenes: @ppedrot I don't want to do things that are not considered "fine". So what should we do here, instead?

pretyping/classops.ml
```diff
@@ -292,12 +292,19 @@ let lookup_path_to_fun_from env sigma s =
...
+  | IndRef i -> EConstr.mkInd i
+
 let get_coercion_constructor env coe =
-  match coe.coe_value with
-  | ConstructRef c ->
-      (c, Inductiveops.constructor_nrealargs c -1)
-  | _ ->
-      raise Not_found
+  let evd = Evd.from_env env in
+  let red x = fst (Reductionops.whd_all_stack env evd x) in
```
ppedrot: This is using the reductionops reduction that is very inefficient. Because you're just looking at the head constructor of an expression, you should use `Reductionops.clos_whd_flags` instead.

gares: Unless I'm mistaken, I used the function that was used before.

plugins/firstorder/ground.ml
ppedrot: I have this change somewhere in one of my PRs but I don't remember where. @maximedenes this style is pretty ugly considering the intended semantics.

pretyping/classops.ml
ppedrot: If you really desire to call a combinator (even though there is no reason for that here) I'd use `UnivGen.fresh_global_instance` or `Evd.fresh_global`.

pretyping/classops.ml
ppedrot: That doesn't make it right, unfortunately. Calls to the Reductionops machine should be really considered twice, they're rarely doing what you're actually wanting.

maximedenes: > That doesn't make it right, unfortunately. Calls to the Reductionops machine should be really considered twice, they're rarely doing what you're actually wanting.

Is the message "you should should call the `Reductionops` machine, but instead use `Reductionops.clos_whd_flags`"? That sounds like a joke.

ppedrot: `Reductionops.clos_whd_flags` is not the Reductionops machine, it's the kernel machine. Maybe the phrasing *Reductionops machine* is misleading, but this module actually implements a full-blown (again, technically two slightly distinct with little code sharing) reduction machine that does wonders like refolding, russian motorbikes and *retour de l'être aimé*.

maximedenes: This is a change that is completely orthogonal to this PR. I'll be willing to do it, even on a large scale (because many pieces are calling the wrong one, I think), but only once we have an understandable API.

plugins/firstorder/ground.ml
maximedenes: Will fix, even if orthogonal, because this is small.

pretyping/classops.ml
ppedrot: Addendum: `clos` probably stands for `(C)Closure`.

maximedenes:
Should be ready now, modulo CI. Who can take it?

pretyping/classops.ml
maximedenes: Ok, I don't think it would bring anything.

plugins/firstorder/ground.ml
maximedenes: Fixed.

maximedenes:
> Should be ready now, modulo CI. Who can take it?

@gares you're the secondary maintainer of `pretyping`, can you be the assignee?

PR #8104, opened 2018-07-20
 Warnings on coercions used without being Imported

maximedenes:
This warning makes it much easier to stop relying on `Set Automatic
Coercions Import`. Tested with success on math-classes.

Depends on #8103 [merged].

### DIFF
--- a/doc/sphinx/addendum/implicit-coercions.rst
+++ b/doc/sphinx/addendum/implicit-coercions.rst
@@ -325,6 +325,12 @@ Coercions and Modules
    This option makes it possible to recover the behavior of the versions of
    |Coq| prior to 8.3.
 
+.. warn:: Coercion used but not in scope: @qualid. If you want to use this coercion, please Import the module that contains it.
+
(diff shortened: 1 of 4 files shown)


maximedenes:
@gares can you be the assignee for this one too? Thanks!

Zimmi48 (changes requested):
I believe this is a case that would really benefit from an addition to the reference manual, since it supports documenting warnings with the `.. warn::` object. I suggest you document this warning inside the description of the corresponding (deprecated) option.

Not all my comments are hard requests but I would like to see at least some documentation or a clearer warning message, ideally both.

gares (approved):
My only concern was already reported by @Zimmi48 (making the text of the warning more explicit on the remedy).

herbelin (approved):
Not a heavy user of coercion and thus difficult to make up a proper opinion, but that looks a priori like a useful proposal.

maximedenes:
@Zimmi48 comments addressed.

Zimmi48 (approved):

PR #8105, opened 2018-07-20
Remove environment passing to cache_coercion function

SkySkimmer:


SkySkimmer:
Rebased and extended after #8103 which reduces usage of the environment.

herbelin:
It seems to make more sense than the previous code (more consistent use of `cache_object` and cie, more consistent code for `vernac_print`). I guess however that at some time, all these functions shall have to take explicitly a `state` as input (is that correctly the consensus for the longer term?)

herbelin (approved):
Side comment: I would recommend that the author gives a minimal description of his PR :wink:. Otherwise, looks ok, as said previously.

@mattam82: I'm assigning. I shall consider by default that you shall not want to make a review (since the PR is rather cosmetic) and merge withing the next few days.

ejgallego (approved):

SkySkimmer:
Conflict gone.

Issue #5505, opened 2017-04-29
[instantiate] does not properly restrict the context of evars


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#5505
From: @JasonGross
Reported version: 8.6

See also: [BZ#5504](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%235504%22)

herbelin:
Duplicate of #5378.

Issue #5504, opened 2017-04-29
[instantiate] can create ill-typed terms


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#5504
From: @JasonGross
Reported version: 8.6

See also: [BZ#5505](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%235505%22)

herbelin:
Duplicate of #5378.

PR #8110, opened 2018-07-21
Fixing capital letters in the "in" syntax of instantiate.

herbelin:
This trace of V7-style capital letters in an admittedly uncommon syntax remained unnoticed for 14 years.

**Kind:** bug fix

Found by @zeimer (see #8072).

The PR is purely about the syntax side (not about the technical issues with `instantiate`, #5378, #5504, #5505, ...).

### DIFF
--- a/doc/sphinx/proof-engine/tactics.rst
+++ b/doc/sphinx/proof-engine/tactics.rst
@@ -1323,8 +1323,8 @@ name of the variable (here :g:`n`) is chosen based on :g:`T`.
    changes in the goal, its use is strongly discouraged.
 
 .. tacv:: instantiate ( @num := @term ) in @ident
-.. tacv:: instantiate ( @num := @term ) in ( Value of @ident )
-.. tacv:: instantiate ( @num := @term ) in ( Type of @ident )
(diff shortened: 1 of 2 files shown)


ppedrot (commented):
Isn't that changing the keyword status of `type` or `value` in any way? I think it does not but I'm asking just in case.

ppedrot:
After pondering a bit, I think this PR should probably deserve comments by more devs. I understand that it brings uniformity to a rarely used syntax, but let's summon people just in case by marking this PR a needing discussion.

maximedenes (approved):

herbelin:
> by marking this PR a needing discussion.

Anyone willing to discuss this PR after the summer breaks?

> Isn't that changing the keyword status of type or value

It does not.
```coq
Definition Value := 0. (* ok *)
```

The reason (to double-check) is that `VERNAC EXTEND` detects that it comes after another token, `in`, so that they don't need to be declared as tokens themselves.

JasonGross (approved):

ppedrot (approved):
I don't think that this PR will trigger a fierce opposition nor any enthusiasm from other devs, so I'll just leave a few days before merging it quietly.

Issue #8026, opened 2018-07-09
Named goals in {…}


nomeata:
Now that we have the `2: {…}` syntax, I was hoping I can do this:

```
Lemma foo (n: nat) (P : nat -> Prop):
  P n.
Proof.
  intros.
  refine (nat_ind _ ?[Base] ?[Step] _).
  [Base]: { … }.
  [Step]: { … }.
Qed.
```
but it is not yet supported, it seems. Is there a plan to support this?
EDIT: supported since 8.9

And, as a next step, is there a way of indicating the name for a named goals already in the lemma, so that this would work (note that the assumptions of `nat_ind_named` have names):
```
Lemma nat_ind_named:
  forall (P : nat -> Prop) (n: nat)
  (base : P 0)
  (step : forall n, P n -> P (S n)),
  P n.
Proof. intros. apply nat_ind; assumption. Qed.
  
Lemma foo (n: nat) (P : nat -> Prop):
  P n.
Proof.
  intros.
  apply nat_ind_named.
  [base]: { … }.
  [step]: { … }.
Qed.
```

Zimmi48:
@nomeata Thanks for your request. This had been my plan following the introduction of `2: { … }` but I was waiting for user requests to confirm the need. See #8112 for the first part of your request.

For the second part, I would indeed support such a feature (it would be very useful to have this at least for auto-generated induction principles for inductive types), but I'm afraid there are deeper unanswered design questions about the lives of names in general. See also #7997 and #268. I'm sure that @ppedrot can comment on the present proposal as well.

While waiting for a big redesign that would help solving all the names issues, what about a tactic that would allow renaming sub-goals, for instance:

```
Lemma foo (n: nat) (P : nat -> Prop):
  P n.
Proof.
  intros.
  apply nat_ind_named; name_goals [ base | step ].
```

where `name_goals [ base | step ]` would be equivalent to `[ refine ?[base] | refine ?[step] ]`?

I'm not sure it would represent such a big improvement over just using a plain `refine` though...

nomeata:
I guess one would wrap it in a tactic in either case, so the gain is small - but but completely zero, as a `named_goal` tactic in the documentation would highlight this pattern!

Zimmi48:
[...]

PR #8112, opened 2018-07-21
Add support for focusing on named goals using brackets.

Zimmi48:
**Kind:** feature

Grants the first part of #8026

  - [x] Including any warning and error messages added / removed / modified).
- [x] Entry added in CHANGES.

### DIFF
--- a/ide/coq_lex.mll
+++ b/ide/coq_lex.mll
@@ -23,7 +23,11 @@ let number = [ '0'-'9' ]+
 
 let string = "\"" _+ "\""
 
-let undotted_sep = (number space* ':' space*)? '{' | '}' | '-'+ | '+'+ | '*'+
+let alpha = ['a'-'z' 'A'-'Z']
+
+let ident = alpha (alpha | number | '_' | "'")*
+
+let undotted_sep = ((number | '[' ident ']') space* ':' space*)? '{' | '}' | '-'+ | '+'+ | '*'+
 
 let vernac_control = "Fail" | "Time" | "Redirect" space+ string | "Timeout" space+ number
 
--- a/CHANGES
+++ b/CHANGES
@@ -48,6 +48,12 @@ Tactics
   may need to add `Require Import Lra` to your developments. For compatibility,
   we now define `fourier` as a deprecated alias of `lra`.
 
+Focusing
+
+- Focusing bracket `{` now supports named goal selectors,
+  e.g. `[x]: {` will focus on a goal (existential variable) named `x`.
+  As usual, unfocus with `}` once the sub-goal is fully solved.
+
 Standard Library
 
 - Added `Ascii.eqb` and `String.eqb` and the `=?` notation for them,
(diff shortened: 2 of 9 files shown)


ide/coq_lex.mll
```diff
@@ -23,7 +23,11 @@ let number = [ '0'-'9' ]+
 
 let string = "\"" _+ "\""
 
-let undotted_sep = (number space* ':' space*)? '{' | '}' | '-'+ | '+'+ | '*'+
+let alpha = ['a'-'z' 'A'-'Z']
+
+let ident = alpha (alpha | number | '_' | "'")*
```
Zimmi48: For now, this is a clear under-approximation. I'm not against some help in making it less so.

ppedrot (approved):
LGTM. Reviewing this patch made me realize that we're still using this horrible Obj hack in the Proof module. I'll submit a fix for it.

Zimmi48:
@ppedrot What do you think of the parsing under-approximation in CoqIDE? Ideally, we should stop requesting editors to parse Coq files (cf. SerAPI which does the splitting of sentences on its own), but in the meantime what to do? Keep it an under-approximation or improve this? Do you have suggestions?

ppedrot:
The current splitting algorithm is already an approximation IIRC, so I don't think it really hurts to extend it a bit. We could probably defer it to coq itself, but then I am slightly concerned about the potential cost to it, as it would introduce latency if done improperly. Anyways we can always enhance this after this PR.

cpitclaudel (approved):
Docs look good.

cpitclaudel:
Do recent Coq versions have a solution for this in Ltac?  That is, can you run a piece of Ltac in the context of a named evar? The one solution I'm familiar with is `instantiate (1 := ltac:(…))`, which is a bit unwieldy.

cpitclaudel:
> but you cannot use this as part of a larger Ltac macro (goal selectors are not technically part of Ltac).

Right, that's what I meant.  Could your code be extended to allow `instantiate ([x] := ltac:(…))`

Zimmi48:
Sorry but my previous answer was quite incomplete, you can do the following:

```coq
Goal exists n : nat, n = n.
eexists ?[x]; only [x]: exact 0.
```

Basically you add the `only` keyword to be able to use a goal selector inside Ltac (documented here https://coq.inria.fr/refman/proof-engine/ltac.html#coq:tacv.only and available since 8.6). Sorry for forgetting about this earlier on. I would definitely recommend using this over `instantiate` whenever you can.

herbelin:
@cpitclaudel, there is already syntax for that:
```coq
Goal exists x, x = 1.
eexists ?[x].
instantiate (x := ltac:(apply S)).
```
Note ongoing discussions on the status of `instantiate (n:=...)`. In particular, `instantiate (x:=...)` should not be the same as `instantiate (n:=...)`. The first one is in the context of the evar. The second one should morally be in the context of the goal in which the evar appears (since the evar is then in general subject to a substitution).

cpitclaudel:
Thanks @Zimmi48 and @herbelin , this is great to know :)

Zimmi48:
@ppedrot: @mattam82 is only a code owner because of the trivial change in `vernacentries.ml` so I don't think you need to wait for his review before merging.

ppedrot:
All right, let's do it then.

PR #8115, opened 2018-07-22
Support for custom entries in notations (take 2, feature part)

herbelin:
**Kind:** feature

This is the feature part of #7923, which is also an extension/fix of #6247. It contains @ppedrot patch for marschallable custom entry data. I failed to make an online diff with github (see below on how to get one). Besides a couple of alpha-conversion, the main additions are in `constrextern.ml` and in `notation.ml`, to fix printing bugs and add supports for coercions from an entry level to another entry or to another level of the same entry.

Documentation is in #7923.

Please don't hesitate to ask questions.

(*) the lines to copy-paste to get a diff between #6247 and this PR:
```
git remote add herbelin https://github.com/herbelin/github-coq
git fetch herbelin
git diff herbelin/master+notation-custom-entry herbelin/master+notation-custom-entry-feature-new
```


ejgallego:
Taking on this as `interp` 2nd; @ppedrot please can you give an eye?

@herbelin , build seems broken.

ppedrot (approved):
I am definitely not a specialist of Metasyntax, so I trust Hugo on the correctness of the feature. I haven't spotted anything suspicious in the overall patch, so it's good to go for me (provided this works as intended).

herbelin:
> build seems broken.

Indeed, sorry about it. My last changes had a typo and had broken the printer and a test. Now fixed. It remains a problem in relation with @ppedrot's patch that I don't have time to investigate now (something wrong at module closure).

ejgallego:
@herbelin still broken on a few places including the test-suite.

herbelin:
When I looked at it yesterday, the failures `success/Notations.v`, `success/Notations2.v` and `output/Notations2.v` were related to @ppedrot's patch. I suspect a priori that the other failures are also related to the backtracking patch (a problem of notations wrongly disappearing at module closure, and of custom entry names wrongly not disappearing, even without `Import`). My time is short these days, so unless @ppedrot has time, it shall wait for a couple of more days, sorry.

ejgallego:
I see, I'll try to have a look too but I can't promise anything as I am on holidays.

gares (approved):

herbelin:
Good news. I found the bugs in the backtracking patch (a typo and a missing `try`/`with`). There was also another little regression when redefining a notation.

All seems now green (phew.... ).

interp/constrexpr.ml
```diff
@@ -20,7 +20,10 @@ type universe_decl_expr = (lident list, Glob_term.glob_constraint list) UState.g
 type ident_decl = lident * universe_decl_expr option
 type name_decl = lname * universe_decl_expr option
 
-type notation = string
+type notation_entry = InConstrEntry | InCustomEntry of string
```
ejgallego: Why is `constr` special? IMHO, we could benefit from a more generic handling here.

interp/constrextern.ml
```diff
@@ -325,16 +333,16 @@ let is_zero s =
   in aux 0
 
 let make_notation_gen loc ntn mknot mkprim destprim l bl =
-  match ntn,List.map destprim l with
+  match snd ntn,List.map destprim l with
```
ejgallego: Not a fan of all the `snd/fst` added here; would it be too much of a pain to turn this into a record?

interp/notation.ml
```diff
@@ -39,6 +39,30 @@ open Context.Named.Declaration
...
+| (InConstrEntry | InCustomEntry _), _ -> false
+
+let notation_entry_level_eq s1 s2 = match (s1,s2) with
+| InConstrEntrySomeLevel, InConstrEntrySomeLevel -> true
+| InCustomEntryLevel (s1,n1), InCustomEntryLevel (s2,n2) -> String.equal s1 s2 && n1 = n2
+| (InConstrEntrySomeLevel | InCustomEntryLevel _), _ -> false
+
+let notation_eq (from1,ntn1) (from2,ntn2) =
+  notation_entry_level_eq from1 from2 && String.equal ntn1 ntn2
+
```
ejgallego: Shouldn't this be in in constrexpr_ops ?

ejgallego (approved):
This looks pretty good to me, congrats @herbelin (and @ppedrot ) on the nice patch.

I have a few (non-blocking) comments; I will wait for Hugo's answer before merging; based on the discussion on other PRs I think no other issue remains, if you want to hold the merge please speak soonish.

interp/notation.ml
herbelin: This would indeed be consistent with type `notation` being in `constrexpr.ml`. But `Constrexpr_ops` depends on `Smartlocate` which depends on `Notation` what requires `Notation` before `Constrexpr_ops`. On the other side, `eq_notation` is used in `Notation` (e.g. for `scope_eq`). So, we would need to break the circularity. Maybe it is `Constrexpr_ops.is_constructor` (sole dependency of `Constrexpr_ops` into `Smartlocate` thus `Notation`), which is not at its right place, but then what to do with `Constrexpr_ops.cases_pattern_fold_names` which depends on `is_constructor` (making it abstract over the latter and pass it as an argument?).

In short, it is a bit of a headache...

interp/constrextern.ml
herbelin: If you insist very strongly, I can try, but to tell the truth I'm a bit exhausted by this PR and I would prefer that such a change is done at a later step when there are less other things to finish.

interp/constrexpr.ml
herbelin: Currently, `constr` remains special. For instance, insertion of parentheses at printing time could now be done generically as it is done for custom entry (in particular one could eventually get rid of the `if prec_less` in `ppconstr.ml`). The special treatment of `{` ... `}` is also only for `constr`.

Eventually, the code could not have to distinguish between `InConstrEntry` and `InCustomEntry` (or between `InConstrEntrySomeLevel` and `InCustomEntryLevel`), but that's not yet the case.

interp/constrextern.ml
herbelin: Yes, retrieving information in the ocean of PR comments is difficult. Maybe adding the question of dependencies internal to `interp` in the [Maintenance](https://github.com/coq/coq/projects/8) or [API](https://github.com/coq/coq/projects/8) project, and the question of unifying `constr` and `custom` in a `Notations` project?

ejgallego:
Merged, thanks @herbelin ; can you update #7135 and #7923 ?

herbelin:
Hum, I'm very sorry, I made a mistake: my last forced update to this PR included the two "documentation" commits that were supposed to be only in #7923. (Usually, I was pushing only `HEAD~2` to this branch, and `HEAD` to #7923, but I apparently forgot to add the `HEAD~2`.)

Maybe can we (@ejgallego?) revert the two unintended documentation commits so that we can discuss about them calmly in #7923?

ejgallego:
Can you rather amend the doc PR?

herbelin:
> Can you rather amend the doc PR?

That's a possibility. Then I have to think not to update the branch bound to #7923 (so that it does not close automatically). At the current time, the two documentation commits (in `doc/sphinx/user-extensions/syntax-extensions.rst`) are exactly the same as those that were committed, so #7923 is a faithful place to comment about the documentation.

interp/constrextern.ml
Zimmi48: Whereas I encourage you to use projects to organize (part of) your work, projects without a clear intent and without regular updates like the two you mentioned are pretty useless. So opening issues, possibly with "help wanted" labels seems like a better solution.

Issue #8078, opened 2018-07-16
Move Windows test-suite to GitLab CI and get rid of AppVeyor.


Zimmi48:
AppVeyor is now consistently failing on all builds, again! (Mostly 1h timeouts)

@maximedenes How reasonable would it be to move the Windows test-suite to GitLab CI? If I understand correctly, the main limitation is the number of VMs we have access to.

maximedenes:
I don't know, can you measure the current load? We shouldn't count on more VMs from the Inria CloudStack, because we need to create macOS VMs and it will almost exhaust our limited quota.

Zimmi48:
I see. So this would mean adding a new build and putting more load on existing Windows runners.
At this very moment, we have 10 jobs (i.e. 5 pipelines) in the queue on GitLab CI (not started) and they are all Windows. On the other hand, we have 6 jobs in the queue on AppVeyor, and this correspond to the observation that AppVeyor usually finishes after GitLab CI.
How many VMs can we have and how many are we currently using for Windows? I don't think we should have half Windows and half macOS VMs given that we could have 3 Windows jobs for only one macOS job (assuming we keep Travis for running the test-suite on macOS).
Another option would be to stop running the 32 bit packaging job on PRs given that both packaging jobs usually fail at the same time and succeed at the same time.

maximedenes:
How clear is it that appveyor has lower availability than Inria's CloudStack?

SkySkimmer:
On gitlab we could put the test suite and the packaging in the same job without worrying about 1h timeouts right? Is that the only reason they were separate?

Zimmi48:
> How clear is it that appveyor has lower availability than Inria's CloudStack?

It's not fully clear but what I really want to address is the unreliability of AppVeyor and not the availability level. AppVeyor gives us a single job at the time and we have about 5 Windows runners IIUC.

> On gitlab we could put the test suite and the packaging in the same job without worrying about 1h timeouts right? Is that the only reason they were separate?

I believe you are right.

Zimmi48:
Another difference is that the test-suite is currently depending on OPAM but the packaging jobs are not (this difference may be seen as a bad idea though).

ejgallego:
I am not sure I understand the point about OPAM [is done on purpose so we test both setups], but I don't see ourselves abandoning AppVeyor at least until the INRIA CI setup is properly secured.

Zimmi48:
[...]

PR #8136, opened 2018-07-24
Do not run 32-bit Windows builds on pull requests.

Zimmi48:
I have never seen this build fail without the 64-bit Windows build failing as well. So it is just a waste of time to test it on every pull request and we are short in Windows test machines.

**Kind:** infrastructure.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -212,6 +212,9 @@ windows32:
   <<: *windows-template
   variables:
     ARCH: "32"
+  except:
+    - /^pr-.*$/
+
 
 pkg:nix:
   image: nixorg/nix:latest # Minimal NixOS image which doesn't even contain git


ejgallego:
Ummm, we are not short on windows runners AFAICT, it is just that we haven't enabled them.

Zimmi48:
> Ummm, we are not short on windows runners AFAICT, it is just that we haven't enabled them.

Indeed, I've been wondering why we only had three Windows jobs running at the same time, on the following runners:
- windows-untrusted5
- windows-untrusted-i02
- INRIACI-UO50J28

while we have way more VMs that report as up and that are registered as runners. Do you understand what's going on?

Anyways, even if we can fix this (we could do it by deleting existing non-functioning VMs and creating and registering them again if there is no better way), I still believe that the present PR should be merged: as I said I've never seen a Windows 32 bit build fail without the Windows 64 bit build failing as well, so it still looks like a loss of energy to do this test on every PR. Furthermore (and more importantly), I'd like to get rid of AppVeyor (#8078) but the limiting factor is our capacity on GitLab CI, so removing this one job would allow to put a Windows test-suite job instead.

ejgallego:
> while we have way more VMs that report as up and that are registered as runners. Do you understand what's going on?

I don't; in the past we had 6 runners, but it seems that the CI handling is done quite opaquely. [Not even an action log AFAICS, but there could be one in the admin console]

As long as we plan to distribute both 32 and 64 bit versions for Windows I'd like to see both jobs remain active. Removing Appveyor is an orthogonal thing and poses different kind of challenges AFAICT.

ejgallego (changes requested):
My preliminary position here is that as long as we are shipping a 32bit windows build I'd rather keep the job in place.

@SkySkimmer ?

SkySkimmer:
I don't know.
Maybe @maximedenes or @MSoegtropIMC have an opinion.

ejgallego:
Certainly the missing Windows runners should be restored ASAP @maximedenes

MSoegtropIMC:
As far as I know many "power Coq users" prefer the 32 bit version because it is faster (pointers are smaller and need lass cache space and Coq data is likely mostly pointers). So I would say that both, the 32 bit and the 64 bit version are important.

On the other hand I tend to agree that testing both 10x a day is a waste of resources (think of Coq's CO2 footprint ;-) But if we disable one or the other, I would insist on having e.g. a daily build which tests more things, which I think is anyway a good idea. Both are required, and I think the last thing @maximedenes wants when doing a release is a not working windows installer.

One more note: in case we have a daily build, I would add a test which actually installs Coq from the installer and runs a few regression tests to see if it actually works.

Zimmi48:
> But if we disable one or the other, I would insist on having e.g. a daily build which tests more things, which I think is anyway a good idea.

Not that this PR only removes the test on pull requests, but keeps it in the master branch, so each time a pull request is merged, the build is still tested. The only (very low) risk that this PR could pose would be to find out that we have broken the build after a PR was merged. But there wouldn't be any risk of discovering the problem at release time.

MSoegtropIMC:
@Zimmi48 : thanks, I am good to go then.

off topic: Would you recommend to add the "test if the installer is not just a bunch of random bytes" test in the same way, e.g. run only on a PR? It might not take so long - I wouldn't run a largish regression test, just some tests that it makes sense and that the included addons are accessible. Total run time might be 10 minutes or so.

Zimmi48:
Yes, I would recommend adding this test to all packaging jobs, including the ones that run on a PR.

Zimmi48:
@ejgallego I suggest to go ahead and merge this PR. We are back to having 6 Windows runners which is better, but this is still the limiting factor of our CI, and it will become even more so once we add the macOS packaging job.

MSoegtropIMC:
I agree. Without a test that the installer actually does something useful, this CI test is not so useful anyway.

ejgallego (approved):
Merging as @Zimmi48 requested.

ejgallego:
Please rebase so the CI runs again.

Zimmi48:
Done.

PR #8137, opened 2018-07-24
Fix 8132.  Print the content of body, not its type.

### LINKED ISSUE #8132: Proof diff mode incorrectly renders hypotheses
```
coqtop -diffs on
> Goal True. pose 3.
1 subgoal
  n := nat : nat
  ============================
  True
```

jfehrle:
**Kind:** bug fix.

Fixes #8132

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -130,7 +130,7 @@ Display diffs between proof steps
 
 - coqtop and coqide can now highlight the differences between proof steps
   in color. This can be enabled from the command line or the
-  "Set Diffs on|off|removed" command. Please see the documentation for
+  `Set Diffs "on"|"off"|"removed"` command. Please see the documentation for
   details.  Showing diffs in Proof General requires small changes to PG
   (under discussion).
 
--- a/printing/proof_diffs.ml
+++ b/printing/proof_diffs.ml
@@ -251,6 +251,11 @@ let pr_letype_core goal_concl_style env sigma t =
 let pp_of_type env sigma ty =
   pr_letype_core true env sigma EConstr.(of_constr ty)
 
+let pr_leconstr_core goal_concl_style env sigma t =
+  Ppconstr.pr_lconstr_expr (Constrextern.extern_constr goal_concl_style env sigma t)
+
+let pr_lconstr_env env sigma c = pr_leconstr_core false env sigma (EConstr.of_constr c)
+
 (* fetch info from a goal, returning (idents, map, concl_pp) where
 idents is a list with one entry for each hypothesis, each entry is the list of
 idents on the lhs of the hypothesis.  map is a map from ident to hyp_info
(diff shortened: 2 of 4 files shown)


ppedrot:
This does not solve all of the problem, now I get spurious scope openings in hypotheses. For instance,
```
Goal forall n, n + 1 = n.
Proof.
intros.
set (m := n + 1).
```
prints
```
1 subgoal
n : nat
m := (n + 1)%nat : nat
m = n
```
instead of
```
1 subgoal
n : nat
m := n + 1 : nat
m = n
```

jfehrle:
The latest code handles @ppedrot's case using the technique in Printer.pr_compacted_decl.

I see that Printer.pr_rel_decl has code to print anonymous names as "<>".  What commands would generate that kind of output?

jfehrle:
The failure is for ci-equations, not related to my changes. (Github shows the same failure twice.)

CHANGES
```diff
@@ -130,7 +130,7 @@ Display diffs between proof steps
 
 - coqtop and coqide can now highlight the differences between proof steps
   in color. This can be enabled from the command line or the
-  "Set Diffs on|off|removed" command. Please see the documentation for
+  Set Diffs "on"|"off"|"removed" command. Please see the documentation for
```
herbelin: Maybe `Set Diffs "on"|"off"|"removed"` (but see also #8180).

printing/proof_diffs.ml
```diff
@@ -251,6 +251,11 @@ let pr_letype_core goal_concl_style env sigma t =
 let pp_of_type env sigma ty =
   pr_letype_core true env sigma EConstr.(of_constr ty)
 
+let pr_leconstr_core goal_concl_style env sigma t =
+  Ppconstr.pr_lconstr_expr (Constrextern.extern_constr goal_concl_style env sigma t)
+
+let pr_lconstr_env env sigma c = pr_leconstr_core false env sigma (EConstr.of_constr c)
+
```
herbelin: This is already in `printer.ml`, right?

herbelin (approved):
This looks good to me. More precisely, following `Printer.pr_compacted_decl`/`Printer.pr_rel_decl` is the right thing to do, I believe.

> I see that Printer.pr_rel_decl has code to print anonymous names as "<>". What
commands would generate that kind of output?

I believe that this should not show in practice instead maybe in debugging. In general, a name is preventively assigned to variables of a context (using `Namegen.make_all_name_different`).

herbelin:
This seems with high priority. Will merge tomorrow.

herbelin:
> Will merge tomorrow.

I realize that my report is contradictory. As a first step, the two questions deserve an answer.

printing/proof_diffs.ml
jfehrle: Yes, it is.  Proof_diffs can't access functions in Printer because of the linking order--but I can have Printer use the copy in Proof_diffs so there aren't 2 copies.

The relationship between the two files is a bit tricky because (conceptually) the diff code prints the old and new proofs then computes the differences between the Pp.t's.  And we chose to make printing a proof with diffs part of Printer.ml (pr_open_subgoals_diff).  That seemed to make the most sense.

CHANGES
jfehrle: I will make this change and the pr_lconstr_env change below.

printing/proof_diffs.ml
herbelin: OK. Maybe this means that `printer.ml` should be split into a low part and a more higher part for goals. Maybe @ejgallego has some insight of what to do here. But this starts to go beyond the scope of this PR. So, unless this new dependency is a problem for someone, I shall merge (new target) tomorrow.

ejgallego: Yeah, I think that printer indeed seems to have too much stuff. IMHO a splitting into more concrete printers would help, but I decided to defer it for later due to the hassles merging the goal diff pr.

jfehrle:
Thanks for the review @herbelin!

PR #8140, opened 2018-07-25
Fix #7887: Lia struggles with goals in `Type`

### LINKED ISSUE #7887: Lia struggles with goals in `Type`
`lia` fails on the following goal:

```coq
Require Import Lia.

Axiom P : nat -> Set.

Goal forall n : nat, n < 0 -> P n.
intros. lia.
```

while `omega` succeeds. Coq complains about a type error, so I believe the fix should not be too hard, I'll have a look.

maximedenes:
Credits to @ppedrot for the Ltac tricks.

### DIFF
--- a/plugins/micromega/Lia.v
+++ b/plugins/micromega/Lia.v
@@ -23,7 +23,12 @@ Declare ML Module "micromega_plugin".
 
 
 Ltac preprocess :=
-  zify ; unfold Z.succ in * ; unfold Z.pred in *.
+  intros;
+  let T := match goal with |- ?P => P end in let s := type of T in
+  match s with
+  | Prop => zify ; unfold Z.succ in * ; unfold Z.pred in *
+  | _ => exfalso; zify ; unfold Z.succ in * ; unfold Z.pred in *
+  end.
 
 Ltac zchange := 
   intros __wit __varmap __ff ;
--- a/test-suite/bugs/closed/7887.v
+++ b/test-suite/bugs/closed/7887.v
@@ -1,0 +1,7 @@
+Require Import Lia.
+
+Axiom P : nat -> Set.
+
+Goal forall n : nat, n < 0 -> P n.
+intros. lia.
+Qed.


maximedenes:
Any takers? @ppedrot @gares ?

ppedrot:
Isn't it a bit strange that the tactic will do two different things depending whether the goal is a proposition or a type? The current situation seems better than before, but couldn't we try to handle more stuff in zify?

Zimmi48:
@maximedenes Seems to me that you didn't give enough time to the code owner (@fajb) to react before asking for someone else to take this...

maximedenes:
> Seems to me that you didn't give enough time to the code owner (@fajb) to react before asking for someone else to take this...

Ok, what is the recommended time?

ppedrot:
I can step down if @fajb wants to take over, let him change the assignee if so.

maximedenes:
> The current situation seems better than before, but couldn't we try to handle more stuff in zify?

We could, maybe. But this PR is a strict improvement, and `zify` is already a bottleneck on many examples, so I'd prefer to not push too much stuff there. Also, if the goal is already on `Z`, I don't think `zify` should do anything.

Zimmi48:
> Ok, what is the recommended time?

I don't know. In #8098, I put 5 working days. This seemed reasonable...

fajb (approved):
This is good to me: this prevents a low-level error and increases the applicability of the tactic.

maximedenes:
@fajb would you mind merging this? Thanks!

plugins/micromega/Lia.v
```diff
@@ -23,7 +23,12 @@ Declare ML Module "micromega_plugin".
 
 
 Ltac preprocess :=
-  zify ; unfold Z.succ in * ; unfold Z.pred in *.
+  intros;
+  let T := match goal with |- ?P => P end in let s := type of T in
+  match s with
+  | Prop => zify ; unfold Z.succ in * ; unfold Z.pred in *
+  | _ => exfalso; zify ; unfold Z.succ in * ; unfold Z.pred in *
```
JasonGross: What's the reason to duplicate the branches, rather than write
```coq
match s with
| Prop => idtac
| _ => (* we can't solve non-Prop goals except by proving False *) exfalso
end;
zify ; unfold Z.succ in * ; unfold Z.pred in *.
```
?

maximedenes: If that's equivalent, I prefer your version indeed.

JasonGross (changes requested):
I would suggest changing the branches as I suggest (or at least turn `match s` into `lazymatch s`), or else if you end up with a case where `zify` fails on a `Prop` goal, this will mask it, and if `zify` is failing on the context, it will take twice as long to fail.

plugins/micromega/Lia.v
JasonGross: It's almost equivalent.  The current version will backtrack across failures of `zify` in the first branch.  If this is the behavior you want, then I'd suggest leaving it and making a note of such.  In cases where `zify` succeeds, the two versions are equivalent.

maximedenes: No, you're right, it's probably not what we want.

JasonGross: Well, almost equivalent.  If `Z.succ` or `Z.pred` shows up only in the goal, and also the user issues `Opaque Z.succ` or `Opaque Z.pred` before calling this tactic, then the version I suggest will not backtrack across the unfolding failure, while the verseion currently in this PR will.  But, again, I think this backtracking behavior is not the behavior you want `lia` to have.

ppedrot:
@maximedenes Are you willing to implement the changes proposed above?

ppedrot:
@maximedenes ping.

maximedenes:
> Are you willing to implement the changes proposed above?

I won't have time until a while (we made sure this was no longer on the critical path for `Coqlib`), so feel free to take over. Also I'm an Ltac beginner...

ppedrot:
@maximedenes Is this superseded by #9856?

maximedenes:
Yes, it should be.

maximedenes:
But please confirm it before closing this.

fajb:
I confirm #9856 handles goals in Type.

PR #8141, opened 2018-07-25
Diff option in CoqIDE

### LINKED ISSUE #8146: CoqIDE displays diffs even if nobody told it to do so
The CHANGES file says:
```
- coqtop and coqide can now highlight the differences between proof steps
  in color. This can be enabled from the command line or the
  "Set Diffs on|off|removed" command. Please see the documentation for
  details.  Showing diffs in Proof General requires small changes to PG
  (under discussion).
```
but here I see the diffs and I did not pass any options, nor write `Set...` in my document.

ppedrot:
This is a followup of #6801. We add a way to switch the diff option inside CoqIDE be means of a menu, instead of setting it by default at the start.

EDIT: Fixes #8146.

### DIFF
--- a/ide/coq.ml
+++ b/ide/coq.ml
@@ -530,20 +530,31 @@ let break_coqtop coqtop workers =
 
 module PrintOpt =
 struct
-  type t = string list
+  type _ t =
+  | BoolOpt : string list -> bool t
+  | StringOpt : string list -> string t
+
+  let opt_name (type a) : a t -> string list = function
+  | BoolOpt l -> l
+  | StringOpt l -> l
+
+  let opt_data (type a) (key : a t) (v : a) = match key with
+  | BoolOpt l -> Interface.BoolValue v
+  | StringOpt l -> Interface.StringValue v
 
   (* Boolean options *)
 
-  let implicit = ["Printing"; "Implicit"]
-  let coercions = ["Printing"; "Coercions"]
-  let raw_matching = ["Printing"; "Matching"]
-  let notations = ["Printing"; "Notations"]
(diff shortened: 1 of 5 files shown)


jfehrle:
Thanks for adding the GUI controls for the diff setting.  Still so much for me to learn about the CoqIDE code.  Also, the IDE development process on Windows is painfully awkward.

gares (changes requested):

ppedrot:
Actually I did test it, but I didn't quite understand what the removed flag was supposed to do. No wonder why it printed the same as the other diff!

gares:
It print ~~stuff~~ for things that went away.

gares (approved):

Zimmi48:
>If you enter the command Set Diffs "on". in the command line, it is applied only for the next proof step. I think it should persist.

This is pretty bad, but for consistency with other view options, it would make sense to completely ignore `Set Diffs "on"` in CoqIDE and print a warning.

jfehrle:
The other view options have the same problem but show no warning in CoqIDE.  For example, `Unset Printing All.`  If CoqIDE is going to go to the trouble of detecting and ignoring some Set and Unset commands, it's almost as easy to automatically set the option in the menus--and simpler for users.

Zimmi48:
`Unset Printing All.` does show a warning: `Set this option from the IDE menu instead`.
But I am not defending this UX here, I'm rather advocating for consistency. I don't know why this design choice was made. If you think it should be reverted, better open a specific issue to discuss it.

jfehrle:
You're right.  I missed the orange-on-yellow message

Issue #7971, opened 2018-06-30
Intuition's precedence (and others) are broken and need be redesigned/deprecated and replaced


Blaisorblade:
#### Description of the problem

While `assert X by a; b; c` means `(assert X by a); b; c` because `by` takes a `tactic3`, `assert X by intuition a; b; c` means `assert X by intuition (a; b; c)` because `intuition` takes a `tactic`.

That's giving `intuition X` the same precedence treatment as parentheses `(X)` — or `forall y, X`.
If I wrote an arithmetic expression grammar and added a normal operator with that precedence, we'd hopefully all agree it would not be good. It seems clear that `intuition` is not as central an operator to have such precedence rules.

Changing precedence would break existing scripts, so this can only be fixed through deprecation — probably this should be queued up for some planned big rewrite (say, Ltac2).

Sadly, `Print grammar tactic` reveals a number of other occurrences. Worst of all, `by` does not have a coherent precedence — `  | IDENT "stepl"; constr:constr; "by"; tactic:tactic`, since `assert ... by tac` takes `tak` as a `tacexpr3`, and that's far more common (and reasonable).

I expect that most systems with customizable operator precedence (Haskell, Agda, maybe even Coq's precedence levels for notations) do *not* let you declare operators with the precedence of parentheses.

#### Version

8.7.2

Zimmi48:
It's funny because I'm often annoyed to have to put parentheses for the `by` clause of `assert`. On the other hand, the precedence for `intuition` seems wrong indeed. And what do you think of the precedence of `now`? Sometimes, I end up writing `assert ... by now ...` just to avoid putting parentheses.

In any case, this doesn't really need to be part of a larger project, because it is really easy to adapt scripts in a backward compatible way: just add more parentheses.

Blaisorblade:
Thanks for being open to a change.

> It's funny because I'm often annoyed to have to put parentheses for the by clause of assert

True, but for Chlipala-style proofs (which never use `.`, and which are my preference these days) you need some parens for assert either way... so in my personal life I don't mind too much. But I don't want each tactic to have different rules which I must look up in Coq's sources because the reference doesn't have them and `Print Grammar` doesn't print them.

Incidentally, perfect documentation would be better but reduce the problem only a bit — this isn't something a user should consume brainspace for.

> And what do you think of the precedence of now?

[...]

Issue #7676, opened 2018-06-01
Associativity differs between /\ and * and between \/ and +


olaure01:
#### Version
Coq 8.8.0

#### Operating system
Ubuntu 16.04

#### Description of the problem
Binary constructors ```/\``` and ```\/``` (Prop -> Prop -> Prop) associate to the right while * and + (Type -> Type -> Type) associate to the left, thus leading to a strange behaviour when one wants to move from Prop to Type.

```
(* Constuctors \/ and + associates differently *)
(* Constuctors /\ and * associates differently *)

(* \/ associates to the right *)
(* A \/ B \/ C = A \/ (B \/ C) *)
Goal forall A B C : Prop, B -> A \/ B \/ C.
Proof.
intros.
(* Check (A \/ (B \/ C)). *)
right.
left.
assumption.
Qed.

(* + associates to the left *)
(* A + B + C = (A + B) + C *)
Goal forall A B C : Prop, B -> {A} + {B} + {C}.
Proof.
intros.
(* Check (({A} + {B}) + {C}). *)
left.
right.
assumption.
Qed.

(* + associates to the left *)
(* A + B + C = (A + B) + C *)
Goal forall A B C, B -> A + B + C.
Proof.
intros.
(* Check ((A + B) + C)%type. *)
left.
right.
assumption.
Qed.



(* /\ associates to the right *)
(* A /\ B /\ C = A /\ (B /\ C) *)
Goal forall A B C : Prop, A /\ B /\ C -> B.
Proof.
intros A B C H.
(* Check (A /\ (B /\ C)). *)
destruct H as [H1 H2].
destruct H2 as [H2 H3].
assumption.
Qed.

(* * associates to the left *)
(* A * B * C = (A * B) * C *)
Goal forall A B C, A * B * C -> B.
Proof.
intros A B C H.
(* Check ((A * B) * C)%type. *)
destruct H as [H1 H3].
destruct H1 as [H1 H2].
assumption.
Qed.
```

Zimmi48:
cc @maximedenes 
(That's the kind of issues that should be gathered in a GitHub project so that these historical mistakes are not made again in Standard Library 2.0.)

herbelin:
To feed the discussions, note the following constraints we had at the time of moving from 7.0 syntax to 8.0 syntax:
- Since `sumor` takes its `Type` argument on the left, `+` has to be associative on the left.
[...]

PR #8142, opened 2018-07-25
Improved the grammar and spelling of chapter 'Syntax extensions and interpretation scopes' of the Reference Manual.

wkolowski:
**Kind:** documentation

Mostly grammar and spelling issues. Two more serious changes:
1. I changed a broken example that defined and with the notation /\ to one that defines and' using the notation &.
2. A table stated that the default level for tactic is 5, but the description said it's 0. I changed the table entry to 0 for consistency, but didn't check if this is really the case (no idea how to test this).

### DIFF
--- a/doc/sphinx/user-extensions/syntax-extensions.rst
+++ b/doc/sphinx/user-extensions/syntax-extensions.rst
@@ -50,11 +50,11 @@ A notation is always surrounded by double quotes (except when the
 abbreviation has the form of an ordinary applicative expression;
 see :ref:`Abbreviations`). The notation is composed of *tokens* separated by
 spaces. Identifiers in the string (such as ``A`` and ``B``) are the *parameters*
-of the notation. They must occur at least once each in the denoted term. The
+of the notation. Each of them must occur at least once in the denoted term. The
 other elements of the string (such as ``/\``) are the *symbols*.
 
 An identifier can be used as a symbol but it must be surrounded by
-simple quotes to avoid the confusion with a parameter. Similarly,
+single quotes to avoid the confusion with a parameter. Similarly,
 every symbol of at least 3 characters and starting with a simple quote
 must be quoted (then it starts by two single quotes). Here is an
 example.
@@ -119,7 +119,7 @@ command understands. Here is how the previous examples refine.
    Notation "A /\ B" := (and A B) (at level 80, right associativity).
    Notation "A \/ B" := (or A B) (at level 85, right associativity).
 
-By default, a notation is considered non associative, but the
+By default, a notation is considered non-associative, but the
 precedence level is mandatory (except for special cases whose level is
 canonical). The level is either a number or the phrase ``next level``
 whose meaning is obvious.
@@ -139,7 +139,7 @@ instance define prefix notations.
    Notation "~ x" := (not x) (at level 75, right associativity).
 
 One can also define notations for incomplete terms, with the hole
-expected to be inferred at typing time.
+expected to be inferred during typechecking.
 
 .. coqtop:: in
 
@@ -185,14 +185,14 @@ rules. Some simple left factorization work has to be done. Here is an example.
    Notation "x < y" := (lt x y) (at level 70).
    Notation "x < y < z" := (x < y /\ y < z) (at level 70).
 
-In order to factorize the left part of the rules, the sub-expression
-referred by ``y`` has to be at the same level in both rules. However the
+In order to factorize the left part of the rules, the subexpression
+referred to by ``y`` has to be at the same level in both rules. However the
 default behavior puts ``y`` at the next level below 70 in the first rule
-(``no associativity`` is the default), and at the level 200 in the second
+(``no associativity`` is the default), and at level 200 in the second
 rule (``level 200`` is the default for inner expressions). To fix this, we
 need to force the parsing level of ``y``, as follows.
 
-.. coqtop:: all
+.. coqtop:: in
 
    Notation "x < y" := (lt x y) (at level 70).
    Notation "x < y < z" := (x < y /\ y < z) (at level 70, y at next level).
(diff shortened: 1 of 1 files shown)


Zimmi48:
@herbelin or @ppedrot might know about the levels.

ppedrot:
Default tactic level is 5, which is the root of #7971 btw.

Zeimer:
I changed the default level to 5.

herbelin:
All changes look very relevant, thanks a lot.

What was otherwise the problem with `/\`? Somehow I'm attached to this example and if it can be kept, I would prefer.

Zeimer:
Try pasting it to CoqIDE and see what happens (an error). I tried a few ways to fix it, but they didn't work. The fastest fix I could come up with was replacing the example. This probably has something to do with redefining stuff already present in the standard library.

herbelin:
> Try pasting it to CoqIDE and see what happens (an error).

OK, good remark. The failure is due to the type of constructors to be parsed in scope `type_scope`, so, even if one might think that the `where` clause would take precedence over any former declaration of a notation for `/\`, this is not the case unless we put it in `type_scope`.

But scopes are explained only later on and I realize that this is actually not the first example of `/\` in the manual, so, ok for the change.

herbelin:
One more comment actually: I would add a `right associativity` in the `Reserved Notation` for `&` as it is the recommended practice to put one.

[Incidentally, there are inconsistencies in the choice of `right` and `left` associativity in the design of the standard library. I hope I do not overlook anything important, but I feel that if some unity should be adopted, `right` associativity would be the most convenient (especially in conjunctive objects, so that they are nested like `exists` or `sig`). See also #7676.]

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -899,11 +905,11 @@ Binding arguments of a constant to an interpretation scope
    the scope is limited to the argument itself. It does not propagate to
    subterms but the subterms that, after interpretation of the notation,
    turn to be themselves arguments of a reference are interpreted
-   accordingly to the arguments scopes bound to this reference.
+   accordingly to the argument scopes bound to this reference.
 
    .. cmdv:: Arguments @qualid : clear scopes
 
-      Arguments scopes can be cleared with :n:`Arguments @qualid : clear scopes`.
+      Argument scopes can be cleared with :n:`Arguments @qualid : clear scopes`.
```
Zimmi48: It would make sense to reformulate differently: this command can be used to clear argument scopes of `qualid`.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1319,16 +1327,16 @@ tactic language. Tactic notations obey the following syntax:
...
+   a module is closed unless the command ``Local Tactic Notation`` is used instead
+   of :cmd:`Tactic Notation`.
 
 .. rubric:: Footnotes
 
 .. [#and_or_levels] which are the levels effectively chosen in the current
    implementation of Coq
 
-.. [#no_associativity] Coq accepts notations declared as ``no associative`` but the parser on
+.. [#no_associativity] Coq accepts notations declared as ``non-associative`` but the parser on
```
Zimmi48: It was in fact ``no associativity`` and has been fixed in another PR (this explains the conflict).

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1446,18 +1454,16 @@ tactic language. Tactic notations obey the following syntax:
...
-.. [#no_associativity] Coq accepts notations declared as ``no
-   associativity`` but the parser on which Coq is built, namely
-   Camlp5, currently does not implement ``no associativity`` and
-   replaces it with ``left associativity``; hence it is the same for
-   Coq: ``no associativity`` is in fact ``left associativity``, for
-   the purposes of parsing
+.. [#no_associativity] Coq accepts notations declared as non-associative but the parser on
+   which Coq is built, namely Camlp4, currently does not implement ``no associativity`` and
+   replaces it by a ``left associativity``; hence it is the same for Coq: ``no associativity``
+   is in fact ``left associativity``.
```
Zimmi48: There are still problems with the conflict resolution here. Camlp4 had been changed in Camlp5 and this has been reverted, the part "for the purposes of parsing" had been added and this has been removed. A "a" had been removed and has reappeared. Given the recent changes to this paragraph, you can just keep it unchanged in this PR.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1026,11 +1032,11 @@ Binding arguments of a constant to an interpretation scope
    the scope is limited to the argument itself. It does not propagate to
    subterms but the subterms that, after interpretation of the notation,
    turn to be themselves arguments of a reference are interpreted
-   accordingly to the arguments scopes bound to this reference.
+   accordingly to the argument scopes bound to this reference.
 
    .. cmdv:: Arguments @qualid : clear scopes
 
-      Arguments scopes can be cleared with :n:`Arguments @qualid : clear scopes`.
+      This command can be used to clear argument scopes of ``qualid``.
```
Zimmi48: Using ``` :token:`qualid` ``` instead of ``` ``qualid`` ``` would be even better.

Zeimer:
@herbelin 
I won't bother myself with adding "right associativity" for now.

@Zimmi48 
So, will this PR be merged any soon? I would like to do a global consistency check for spelling etc. without running into trouble in the future when rebasing on top of this or something like that.

Zimmi48:
@Zeimer I'm just waiting on the fix to the incorrect conflict resolution mentioned above.

Zeimer:
Oh, now I understand what you mean. I recall doing some conflict resolution during a rebase some time ago, but I thought it was some kind of "regression". I will try to fix this soon.

Zimmi48 (approved):

Zimmi48:
I won't wait for the linter for this PR because of Travis being seriously affected at the moment (https://www.traviscistatus.com/incidents/118bjpkgkg00).

PR #8144, opened 2018-07-25
Improved grammar and spelling for chapters 'Utilities' and 'CoqIDE' of the Reference Manual.

wkolowski:
**Kind:** documentation

Mostly grammar and spelling. There were two more serious issues related to utf8:

1. The command "Require utf8" didn't work, so I replaced it with "Require Import Unicode.Utf8"
2. Documentation for coqdoc states that "The default behavior is to assume ASCII 7 bit input files", but in my opinion this contradicts the description of the option -utf8 (if I recall correctly, coqdoc replaces "->" with "→" by default).

### DIFF
--- a/doc/sphinx/practical-tools/coqide.rst
+++ b/doc/sphinx/practical-tools/coqide.rst
@@ -27,7 +27,7 @@ is shown in the figure :ref:`CoqIDE main screen <coqide_mainscreen>`.
 At the top is a menu bar, and a tool bar
 below it. The large window on the left is displaying the various
 *script buffers*. The upper right window is the *goal window*, where
-goals to prove are displayed. The lower right window is the *message
+goals to be proven are displayed. The lower right window is the *message
 window*, where various messages resulting from commands are displayed.
 At the bottom is the status bar.
 
@@ -62,8 +62,8 @@ In the figure :ref:`CoqIDE main screen <coqide_mainscreen>`,
 the running buffer is `Fermat.v`, all commands until
 the ``Theorem`` have been already executed, and the user tried to go
 forward executing ``Induction n``. That command failed because no such
-tactic exists (tactics are now in lowercase…), and the wrong word is
-underlined.
+tactic exists (names of standard tactics are written in lowercase),
+and the failing command is underlined.
 
 Notice that the processed part of the running buffer is not editable. If
 you ever want to modify something you have to go backward using the up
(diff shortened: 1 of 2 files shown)


doc/sphinx/practical-tools/coqide.rst
```diff
@@ -62,8 +62,8 @@ In the figure :ref:`CoqIDE main screen <coqide_mainscreen>`,
 the running buffer is `Fermat.v`, all commands until
 the ``Theorem`` have been already executed, and the user tried to go
 forward executing ``Induction n``. That command failed because no such
-tactic exists (tactics are now in lowercase…), and the wrong word is
-underlined.
+tactic exists (names of standard tactics are writen in lowercase with
+words separated by underscores), and the failing command is underlined.
```
gares: The underscore part is incorrect, since there are standard tactics (eg `simple apply`) that does not use `_` as a separator but rather ` `.

Zeimer: I removed the mention of underscores.

gares (approved):

Zimmi48:
>Documentation for coqdoc states that "The default behavior is to assume ASCII 7 bit input files", but in my opinion this contradicts the description of the option -utf8 (if I recall correctly, coqdoc replaces "->" with "→" by default).

This text dates back from 0eed97d80d4753095dcb0fe0b10b368f942e422f (14 years ago). Maybe @herbelin can tell more about how coqdoc support for UTF8 evolved since then.

herbelin:
>Documentation for coqdoc states that "The default behavior is to assume ASCII 7 bit input files", but in my opinion this contradicts the description of the option -utf8 (if I recall correctly, coqdoc replaces "->" with "→" by default).

At first view this assumption is correct. The replacement of "->" with "→" is only with option `-utf8`.

More precisely, it seems that `coqdoc` is a bit inconsistent.

With no options, when parsing Coq identifiers, it assumes them to be utf-8 encoded (which is indeed the convention for Coq). But when producing the html file, it adds a `charset=iso-8859-1` header by default. I guess it should instead be `charset=ASCII`.

With options `-utf8` everything seems fine: LaTeX requires `utf8x`, html declares `charset=utf-8`, identifiers are parsed assuming utf-8, symbols like "->" are replaced with "→" .

With options `--latin1`, it is almost fine: LaTeX requires `latin1`, html declares `charset=utf-8`. The only issue is that Coq identifiers are considered in utf-8 (and there is no other choice than that!). So, the option `--latin1` makes sense only for a file which uses no utf-8 in the Coq part and latin-1 in the comments.

doc/sphinx/practical-tools/utilities.rst
```diff
@@ -533,7 +533,7 @@ or
 
 
 It gives the |Latex| and HTML texts to be produced for the given |Coq|
-token. One of the |Latex| or HTML text may be omitted, causing the
+token. Either the |Latex| or the HTML rule be omitted, causing the
```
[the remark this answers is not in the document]
Zeimer: Fixed.

PR #8145, opened 2018-07-25
Improved grammar and spelling in chapter 'Extended pattern matching' of the Reference Manual.

wkolowski:
**Kind:** documentation

Mostly grammar and spelling issues.

General remarks:
1. The examples are really bad. Last time I checked (i.e. a year ago) most of them didn't really require the provided annotations.
2. "You can only catch let bindings in mode where you bind all variables and so you have to use @ syntax." - What kind of mode was meant here?

In general I think that this chapter should be rewritten from scratch with better examples where the annotations can't be omitted.

### DIFF
--- a/doc/sphinx/addendum/extended-pattern-matching.rst
+++ b/doc/sphinx/addendum/extended-pattern-matching.rst
@@ -11,7 +11,7 @@ Extended pattern-matching
 
 This section describes the full form of pattern-matching in |Coq| terms.
 
-.. |rhs| replace:: right hand side
+.. |rhs| replace:: right hand sides
 
 Patterns
 --------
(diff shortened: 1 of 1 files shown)


doc/sphinx/addendum/extended-pattern-matching.rst
```diff
@@ -448,7 +448,7 @@ Using pattern matching to write proofs
 In all the previous examples the elimination predicate does not depend
 on the object(s) matched. But it may depend and the typical case is
 when we write a proof by induction or a function that yields an object
-of dependent type. An example of proof using match in given in Section
+of a dependent type. An example of proof using match in given in Section
 8.2.3.
```
Zimmi48: This should now be: ```in the description of :tacn:`refine` ```

Zimmi48 (approved):
Thanks!
You should open an issue about this chapter needing a rewrite (at least the examples) so that we don't forget (and maybe some knowledgeable contributor can help if the main developers do not find the time).
More generally speaking, I suggest you open issues about all the remaining problems that you noticed and were not fixed by the time of merging your PRs.

doc/sphinx/addendum/extended-pattern-matching.rst
Zeimer: Fixed.

PR #8148, opened 2018-07-25
Doc: preliminary work before #7291 which add an "Unable to unify" message

herbelin:
This is a minor PR in preparation of #8077, as suggested by @ppedrot [there](https://github.com/coq/coq/pull/8077#issuecomment-407671606).

We adopt the convention that error messages with a template in the tactic chapter use the sphinx syntax used in defining syntax rules to describe the template. I did not try to look at how it was done in other chapters though.

**Kind:** documentation

### DIFF
--- a/doc/sphinx/proof-engine/tactics.rst
+++ b/doc/sphinx/proof-engine/tactics.rst
@@ -281,7 +281,7 @@ Applying theorems
    :g:`t`:sub:`n` in the goal. See :tacn:`pattern` to transform the goal so that it
    gets the form :g:`(fun x => Q) u`:sub:`1` :g:`...`  :g:`u`:sub:`n`.
 
-   .. exn:: Unable to unify ... with ... .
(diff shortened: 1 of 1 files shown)


doc/sphinx/proof-engine/tactics.rst
```diff
@@ -281,7 +281,7 @@ Applying theorems
    :g:`t`:sub:`n` in the goal. See :tacn:`pattern` to transform the goal so that it
    gets the form :g:`(fun x => Q) u`:sub:`1` :g:`...`  :g:`u`:sub:`n`.
 
-   .. exn:: Unable to unify ... with ... .
+   .. exn:: Unable to unify @term with @term.
```
Zimmi48: You could even do `@term__1 with @term__2` (but the present form is fine).

Zimmi48 (approved):
:+1:

PR #8151, opened 2018-07-25
Vector: expose ++ to user

liyishuai:
**Kind:** feature
- [x] Entry added in CHANGES.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -139,6 +139,10 @@ Display diffs between proof steps
   details.  Showing diffs in Proof General requires small changes to PG
   (under discussion).
 
+Notations
+
+- Added [++] infix for [VectorDef.append].
+
 Changes from 8.8.0 to 8.8.1
 ===========================
 
--- a/CHANGES
+++ b/CHANGES
@@ -141,7 +141,7 @@ Display diffs between proof steps
 
 Notations
 
-- Added [++] infix for [VectorDef.append].
+- Added `++` infix for `VectorDef.append`.
   Note that this might cause incompatibilities if you have, e.g., list_scope
   and vector_scope both open with vector_scope on top, and expect `++` to
   refer to `app`.
--- a/CHANGES
+++ b/CHANGES
@@ -142,6 +142,10 @@ Display diffs between proof steps
 Notations
 
 - Added [++] infix for [VectorDef.append].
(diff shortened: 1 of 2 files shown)


herbelin:
This was indeed missing. Well-spotted.

In theory, there could be incompatibilities but the PR is going the expected way and the change is documented. So that's all good to me. Thanks

herbelin:
PS: Assigning it to me, since apparently single reviewer. Shall merge between Thursday late and Friday morning if no other comments by then.

Zimmi48:
Actually @liyishuai had already prepared a fix for cross-crypto and this fix has been merged, so we only have to relaunch the test (I'm doing it).

Zimmi48:
Hum, there are two weird things with CI. First, @coqbot should have overridden the failed status of cross-crypto since it is now a success. Second, this build "skipped" a lot of jobs and I have no idea why: https://gitlab.com/coq/coq/pipelines/26497628 (cc @SkySkimmer @ejgallego).

@liyishuai To be sure that everything is good and that there are no more incompatibilities, a rebase of this PR against the latest master would allow a fresh CI run. Should the entry in CHANGES mention that this is a potential source of incompatibility?

SkySkimmer:
Very strange pipeline indeed, no clue what's going on.

herbelin:
Rescheduled for Monday.

CHANGES
```diff
@@ -134,6 +134,14 @@ Display diffs between proof steps
   details.  Showing diffs in Proof General requires small changes to PG
   (under discussion).
 
+Notations
+
+- Added [++] infix for [VectorDef.append].
```
herbelin: Is there any reason to use here `[` ... `]` and not `` ` `` ... `` ` `` like in the rest of the paragraph? 

[Note that there is an ongoing [discussion](https://github.com/coq/coq/issues/8180) on the style to use for  `CHANGES` and, even if w/o final conclusion yet, it seems that we are going towards adopting markdown for `CHANGES`.]

herbelin (commented):
Hi, looking again at your changes to `CHANGES`, I realized that I had one minor comment.

CHANGES
liyishuai: I realized there are two styles after the first commit. Maybe I should unify into ``` ` ``` style.

Issue #8018, opened 2018-07-08
coqpp: locations and tooling


gares:
I'm writing this after a discussion with @maximedenes.
- It seems `coqpp` does not put in the generated text pragmas as `#line 123 "origfile.mlg"` hence locations are lost and error are signaled on the generated file, not on the original file
- It also seems that `coqpp` is not used as in `ocamlc -pp coqpp` hence it leaves around generated files but it does not need to

ejgallego:
> It also seems that coqpp is not used as in ocamlc -pp coqpp hence it leaves around generated files but it does not need to

That's a matter of style, this is what was done with `camlp5` AFAICT; dev-wise, you want to look at the intermediary files often.

maximedenes:
@ppedrot any progress on that? Not having locations is a very serious regression w.r.t. camlp5.

SkySkimmer:
>That's a matter of style, this is what was done with `camlp5` AFAICT; dev-wise, you want to look at the intermediary files often.

I guess instead of READABLE_ML4 we should have KEEP_GENERATED. I don't think you often look at the binary camlp5 output.

ejgallego:
> I guess instead of READABLE_ML4 we should have KEEP_GENERATED. I don't think you often look at the binary camlp5 output.

That mostly matters for the workflow of the dependency rules, the dependency chain is different and so is parallelism. In the Dune version we will surely keep the generated files for now as it simpler [and they don't bother as they are placed in _build].

Zimmi48:
I'm moving this to 8.10+beta1 because the PR fixing this was merged in this milestone but if this regression is concerning for 8.9, maybe it should be backported...

Zimmi48:
I take this as a no.

PR #8161, opened 2018-07-26
Implement VERNAC EXTEND in coqpp

ppedrot:
Quite straightforward. Only one file has been ported, as all other users of this macro also declare `ARGUMENT EXTEND` extensions.

### DIFF
--- a/dev/doc/changes.md
+++ b/dev/doc/changes.md
@@ -159,7 +159,33 @@ END
 
 #### VERNAC EXTEND
 
-Not handled yet.
+Steps to perform:
+- replace the brackets enclosing OCaml code in actions and rule classifiers with
+  braces
+- if not there yet, add a leading `|̀  to the first rule
+
+Handwritten classifiers declared through the `CLASSIFIED BY` statement are
+considered OCaml code, so they also need to be wrapped in braces.
+
+All extension macros are now behaving as if they were declared `FUNCTIONAL`,
+which means that they must return an interpretation state. This should
+currently be done by merely returning the `st` variable that is implicitly
+passed to the rule code.
+
+For instance, code of the form:
+```
+VERNAC COMMAND EXTEND my_command CLASSIFIED BY classifier
+  [ "foo" int(i) ] => [ classif' ] -> [ cmd1 i ]
+| [ "bar" ] -> [ cmd2 ]
+END
+```
+should be turned into
+```
+VERNAC COMMAND EXTEND my_command CLASSIFIED BY { classifier }
+  [ "foo" int(i) ] => { classif' } -> { let () = cmd1 i in st }
+| [ "bar" ] -> { let () = cmd2 in st  }
+END
+```
 
 #### ARGUMENT EXTEND
 
--- a/dev/doc/changes.md
+++ b/dev/doc/changes.md
@@ -167,11 +167,6 @@ Steps to perform:
 Handwritten classifiers declared through the `CLASSIFIED BY` statement are
 considered OCaml code, so they also need to be wrapped in braces.
 
-All extension macros are now behaving as if they were declared `FUNCTIONAL`,
-which means that they must return an interpretation state. This should
-currently be done by merely returning the `st` variable that is implicitly
-passed to the rule code.
-
 For instance, code of the form:
 ```
 VERNAC COMMAND EXTEND my_command CLASSIFIED BY classifier
@@ -182,8 +177,8 @@ END
 should be turned into
 ```
 VERNAC COMMAND EXTEND my_command CLASSIFIED BY { classifier }
-  [ "foo" int(i) ] => { classif' } -> { let () = cmd1 i in st }
-| [ "bar" ] -> { let () = cmd2 in st  }
+| [ "foo" int(i) ] => { classif' } -> { cmd1 i }
+| [ "bar" ] -> { cmd2 }
 END
 ```
 
--- a/plugins/derive/g_derive.mlg
+++ b/plugins/derive/g_derive.mlg
@@ -8,13 +8,21 @@
 (*         *     (see LICENSE file for the text of the license)         *)
 (************************************************************************)
 
+{
+
 open Stdarg
+
+}
 
 DECLARE PLUGIN "derive_plugin"
 
+{
+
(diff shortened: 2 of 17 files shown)


dev/doc/changes.md
```diff
@@ -150,7 +150,33 @@ END
...
+VERNAC COMMAND EXTEND my_command CLASSIFIED BY classifier
+  [ "foo" int(i) ] => [ classif' ] -> [ cmd1 i ]
+| [ "bar" ] -> [ cmd2 ]
+END
+```
+should be turned into
+```
+VERNAC COMMAND EXTEND my_command CLASSIFIED BY { classifier }
+  [ "foo" int(i) ] => { classif' } -> { let () = cmd1 i in st }
+| [ "bar" ] -> { let () = cmd2 in st  }
```
gares: can't you add this "context" in your preprocessor? I don't see why forcing people to return an unbound variable can be helpful (today and even tomorrow when st will be bound for real).

gares: Said otherwise, I'd have `VERNAC....` that is the one that expects the st to be returned and `LEGACY VERNAC ...` that is supposed to die and adds the `let () = ... in st` wrapper. So that when you remove LEGACY you get a sensible type error.

ppedrot: I didn't really understand @ejgallego's plan on this feature. The current situation does not make sense at all, because the returned state is never taken into account by the vernac interpretation. Adding a LEGACY marker would ease the transition, but the default case will require fixing anyways as it is broken...

ejgallego: In order to handle locality we had to change the extension type, so I opted for a quite general one instead of having to define many. That being said feel free to refine with attributes or something.

gares: In other words the current code is `LEGACY` and there is no infrastructure to remove the `LEGACY` (given that the result of functional code is discarded).

In any case, I prefer the code written between `{}` to make sense (in the old imperative model or in the new functional and only half baked mode). This PR makes the code just look silly, so please hide the silly context in the macro (with or without LEGACY, this is up to you).

ejgallego: I wouldn't worry too much about this, but if you want, make the return type an option type, so the top level interpreter can know what to do (`None` would mean legacy)

Anyways, I feel that problems with the ongoing migration to pure state handling shouldn't delay this PR too much.

herbelin:
I'm a bit confused. According to the CODEOWNERS file, @maximedenes should be assignee, but he is not even  in the list of reviewers.

ppedrot:
Finally I made the port behave as the non-FUNCTIONAL extension, because the functional API is never going to make it in 8.9. So, for me this is now ready to go.

gares (approved):

aspiwack (approved):
My review is nominally requested, but the changes to my bits are trivial. So, on my end, nothing to report.

plugins/derive/g_derive.mlg
```diff
@@ -8,13 +8,21 @@
...
+
 let classify_derive_command _ = Vernacexpr.(VtStartProof ("Classic",Doesn'tGuaranteeOpacity,[]),VtLater)
 
-VERNAC COMMAND EXTEND Derive CLASSIFIED BY classify_derive_command
+}
+
+VERNAC COMMAND EXTEND Derive CLASSIFIED BY { classify_derive_command }
 | [ "Derive" ident(f) "SuchThat" constr(suchthat) "As" ident(lemma) ] ->
-     [ Derive.start_deriving f suchthat lemma ]
+  { Derive.start_deriving f suchthat lemma }
```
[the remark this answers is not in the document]
ppedrot: This allows to parse OCaml code without having to write an OCaml parser, using the same kind of trick as in e.g. melt. We have to wrap quotes inside delimiters that are well-parenthesized in the object language, so that it is just a matter of keeping a count of how many delimiters are opened so far (and handling a few constructs like strings and comments).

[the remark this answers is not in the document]
herbelin: I guess the question was about why moving from square brackets to curly brackets to delimit OCaml code. I guess that the reason is by consistency with the newly introduced curly brackets in `mlg` files. OK, you could ask why not to use square brackets in `mlg` files by consistency with `VERNAC EXTEND`. Then, I would guess that this is by consistency with `ocamlyacc`. The hysterical paths followed by syntax are often tortuous...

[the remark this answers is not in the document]
ppedrot: You have to change at least one of the uses of brackets into something else, because you cannot use it both for quotes and as a token. I went for braces for OCaml code because this was minimal in changes regarding camlp5-style grammar extension.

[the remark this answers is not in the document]
aspiwack: It was about both. And the answers are perfectly clear, thanks :slightly_smiling_face:

maximedenes:
@ppedrot please fix https://github.com/coq/coq/issues/8405 and https://github.com/coq/coq/issues/8018 before porting more files to coqpp, otherwise it is going to make everybody's life more painful. Thanks!

ejgallego:
So does this PR preserve a variant of the functional vernac extend?

We are very very close to actually use it for real.

ppedrot:
No, but if this lands in master I can adapt the PR for that.

ejgallego:
So who's gonna be the assignee here? Should I?

maximedenes:
> No, but if this lands in master I can adapt the PR for that.

Is this ok with you, @ejgallego ?

ejgallego:
> Is this ok with you, @ejgallego ?

It is OK to me.

maximedenes:
> What I exactly mean is that the `.ml` files produced by coqp5 can use `coq.gramlib` instead of `camlp5.gramlib`, right?

@ejgallego do you want this change before merging this PR?

ejgallego:
> @ejgallego do you want this change before merging this PR?

I think no change regarding the vendoring of camlp5 is needed here. This is [at the moment] purely a linking issue. Only problem I see with this PR is that a lot of code is not tested/used, but I think we can trust @ppedrot to implement the missing bits.

maximedenes:
> I think no change regarding the vendoring of camlp5 is needed here. This is [at the moment] purely a linking issue. Only problem I see with this PR is that a lot of code is not tested/used, but I think we can trust @ppedrot to implement the missing bits.

Ok, thanks, so merging.

PR #8164, opened 2018-07-26
Add information to option type errors

tchajed:
Print the expected and actual types for the option value (which is one of bool,
int, or string). The boolean type is still a bit confusing for users, since
most options take a boolean [On|Off] but it's usually implicitly "On".

**Kind:** feature. (error messages)


tchajed:
This error message isn't documented so I didn't change any documentation.

gares (approved):

gares:
thanks!

PR #6597, opened 2018-01-16
Binary, Octal, and Hex conversions between [positive], [Z], [N], [nat] and [string]

JasonGross:
@davidben pointed out to me that Coq doesn't have conversations between numbers and strings.  This adds such conversions for `positive`, `N`, `nat`, and `Z` in binary, octal, and hex.

Edit: <strike>This is on top of #6596</strike> (not anymore)

- [x] Entry added in CHANGES.

PR #8169, opened 2018-07-26
NArith: add sized N2Bv

liyishuai:
**Kind:** feature.
- [x] Entry added in CHANGES.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -52,6 +52,8 @@ Standard Library
   Solution: wrap `_ =? _` in `(_ =? _)%Z` (or whichever scope you
   want).
 
+- Added `Ndigits.N2Bv_sized`, and proved some lemmas about it.
+
 Tools
 
 - Coq_makefile lets one override or extend the following variables from
--- a/theories/NArith/Ndigits.v
+++ b/theories/NArith/Ndigits.v
@@ -517,6 +517,23 @@ Definition N2Bv (n:N) : Bvector (N.size_nat n) :=
     | Npos p => P2Bv p
   end.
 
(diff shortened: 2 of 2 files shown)


herbelin:
Looks indeed useful, thanks.

I incidentally have an open question at large. What would be the clearer: to use `xH`, `xI`, `xO` or to use their notations `1`, `~1`, `~0`? This would mean also changing `P2Bv`.

Anyway, the PR looks consistent enough in style with the rest of the file as it is. In the absence of further comments within a couple of days, as canonical assignee, I'll shall merge it (typically on next Wednesday).

liyishuai:
I've realized that there already exists [N2Bv_gen](https://coq.inria.fr/library/Coq.NArith.Ndigits.html#N2Bv_gen).
Shall we revert these changes to avoid duplication?

herbelin:
> I've realized that there already exists N2Bv_gen.

Ow. That would be better to avoid duplication indeed. But if you did not see it that's probably that `N2Bv_gen` is not visible enough. Also, is it worth to have `P2Bv_gen` (or `P2Bv_sized`)? If the author of the file (@letouzey) could comment, that would be helpful.

liyishuai:
It would be worth keeping `P2Bv_gen` if there are real world practices that use `positive` explicitly.

liyishuai:
https://github.com/coq/coq/pull/6597#discussion_r213880595 shows that converting between other types and `positive` is rather tricky.
I think `positive` should be limited as a helper type for constructing `Z` and `N`. There are `NArith` and `ZArith`, but nobody complained about arithmetic support for `positive`.

PR #8171, opened 2018-07-27
Bvector: add BVeq and some notations

liyishuai:
**Kind:** feature.

- [x] Entry added in CHANGES.

### DIFF
--- a/theories/Bool/Bvector.v
+++ b/theories/Bool/Bvector.v
@@ -70,6 +70,8 @@ Definition BVor := @Vector.map2 _ _ _ orb.
 
 Definition BVxor := @Vector.map2 _ _ _ xorb.
 
+Definition BVeq m n := @Vector.eqb bool eqb m n.
+
 Definition BshiftL (n:nat) (bv:Bvector (S n)) (carry:bool) :=
   Bcons carry n (Vector.shiftout bv).
 
@@ -99,3 +101,12 @@ Fixpoint BshiftRa_iter (n:nat) (bv:Bvector (S n)) (p:nat) : Bvector (S n) :=
 
 End BOOLEAN_VECTORS.
 
+Module BvectorNotations.
+Delimit Scope Bvector_scope with Bvector.
+Notation "^~ x" := (Bneg _ x) (at level 35, right associativity) : Bvector_scope.
+Infix    "^&"   := (BVand  _) (at level 40, left  associativity) : Bvector_scope.
+Infix    "^⊕"   := (BVxor  _) (at level 45, left  associativity) : Bvector_scope.
+Infix    "^|"   := (BVor   _) (at level 50, left  associativity) : Bvector_scope.
+Infix    "=?"   := (BVeq _ _) (at level 70, no    associativity) : Bvector_scope.
+Open Scope Bvector_scope.
+End BvectorNotations.
--- a/CHANGES
+++ b/CHANGES
@@ -92,6 +92,9 @@ Standard Library
   impacts users running Coq without the init library (`-nois` or
   `-noinit`) and also issuing `Require Import Coq.Init.Datatypes`.
 
+- Added `Bvector.BVeq` that decides whether two `Bvector`s are equal.
+- Added notations for `BVxor`, `BVand`, `BVor`, `BVeq` and `BVneg`.
+
 Tools
 
 - Coq_makefile lets one override or extend the following variables from


theories/Bool/Bvector.v
```diff
@@ -99,3 +101,12 @@ Fixpoint BshiftRa_iter (n:nat) (bv:Bvector (S n)) (p:nat) : Bvector (S n) :=
 
 End BOOLEAN_VECTORS.
 
+Module BvectorNotations.
+Delimit Scope Bvector_scope with Bvector.
+Infix    "^"   := (BVxor  _)                                    : Bvector_scope.
+Infix    "&"   := (BVand  _) (at level 40, left  associativity) : Bvector_scope.
+Infix    "|"   := (BVor   _) (at level 50, left  associativity) : Bvector_scope.
+Infix    "=?"  := (BVeq _ _) (at level 70, no    associativity) : Bvector_scope.
+Notation "~ x" := (Bneg _ x) (at level 75, right associativity) : Bvector_scope.
```
gares: In Coq `~` is used for propositional negation.
In [MathComp](https://guthub.com/math-comp/math-comp) `~~` (with no space) is used for boolean negation.

I think you should avoid clashing with propositional negation, and eventually pick the same notation as in MathComp

Also, `==` is the symbol used in Coq and MathComp for decideable comparisons (with different implementations, but still). So please use that symbol.

theories/Bool/Bvector.v
```diff
@@ -99,3 +101,12 @@ Fixpoint BshiftRa_iter (n:nat) (bv:Bvector (S n)) (p:nat) : Bvector (S n) :=
 
 End BOOLEAN_VECTORS.
 
+Module BvectorNotations.
+Delimit Scope Bvector_scope with Bvector.
+Infix    "^"   := (BVxor  _)                                    : Bvector_scope.
+Infix    "&"   := (BVand  _) (at level 40, left  associativity) : Bvector_scope.
```
gares: Boolean conjunction/disjunction is denoted `&&` and `||` in Coq.
You should use the same symbols for these I think.

gares (dismissed):
Please uniform the notations.

I also recommend you get in touch with @maximedenes for the https://github.com/coq/stdlib2/ project. One of the things that will be set is a collection of notations.

theories/Bool/Bvector.v
herbelin: I was at the same time wondering the same thing, then I saw https://en.wikipedia.org/wiki/Bitwise_operations_in_C#Bitwise_operators. No conclusion though.

theories/Bool/Bvector.v
herbelin: @gares: I don't understand your remark. The notation `=?` seems to be the preferred way to write a function from `A -> A -> bool` in the standard library, doesn't it? The symbol `==` is rarely used in the standard library (it is used e.g. for quotient equality on `Q`, locally for equality modulo in `Zdiv`, locally for equality of denotations in `Int.v`, ...).

The notation `~` is clearly motivated by the usage in C. Its clash with `not` is a bit worrying. Its precedence level (75) is also worrying: it is higher than the levels of atomic predicates (70).

Personally, I find `~~` pretty disconcerting. I understand the analogy with `||` and `&&` but it very easy to confuse it with a double negation. 

Using `ocaml` syntax, i.e. infix `lor`, `land`, `lxor`, as well as prefix `lnot`, would have the serious drawback that this would make them tokens and thus unusable as identifiers.

herbelin:
Idem, @maximedenes is coordinating (among others) a unification of terminology and notations on the standard library, so his opinion is requested.

My general feeling is that wanting different operators for similar concepts in different types (as it is done in programming languages) would not scale in the case of a proof assistant. So, scopes, or type classes, or any mechanism to reuse similar notations in different contexts should be used.

The reason `||` and `&&` were used on `bool` is by mimetism of programming languages (the C and OCaml culture in particular). Now that they are used, there are probably no reason to change them (except that they look a bit ad hoc, compared to mathematical practice).

In any case, to be concrete, in the short term, I would be ok with `|` and `&` and `^` which are exactly in the programming language spirit of `||` and `&&`. I have however no good solution for `Bneg`.

Also for the internal names, I see that the usage has been taken to prefix the names of `Bvector` with `B` or `BV`, so let's continue like this, even if the usage evolved to rather use qualified names. By consistency with the `beq` schemes used elsewhere for equality to `bool`, I would nevertheless provide an alias (i.e. an additional `Notation beq := BVeq` with comment "alternative name following standard rule for naming equality to bool").

These opinions are only mine, and I'd be happy to hear other visions.

theories/Bool/Bvector.v
gares: You are right about `=?`, I've mistaken it for comparison (the ternary one).

I'm not so sure `==` is not used
```
$ git grep ' == ' theories | wc -l
1231
```
Mostly in `.../Numbers/...`, but it does not stand for decidable equality most of the times, so my remark should be dropped.

gares:
I don't understand the remark. Sure in C they are written like that. In Coq they are already written differently (as in `&&`), so why would be OK to write these operations in two different ways in Coq's library.

gares:
I may have missed the point of this PR.  If it is to have "C like" notations, then please discard my review.

liyishuai:
My purpose is to have *some* notation for these operators. C style is more of a candidate than reference.

herbelin:
Seems like I'm the canonical assignee.

JasonGross (dismissed):
What level does `|` need to be at to not break the syntax for `Inductive`, `match`, Ltac `match`, and `first`?

In any case, please add a test-suite file that `Import`s these notations after `Require Import ZArith` (to check for lack of clashes with Z notations), and then tests that all of the standard building blocks using `|` and `&` still work correctly (namely, ` Inductive`, Gallina `match`, Ltac `match`, the `first` tactic with at least two branches, the `[ > ... ]` Ltac construction with at least two branches, and the notation for `sigT`)

JasonGross:
To throw another proposal in the mix: in fiat-crypto, we use `&'` and `|'` for `Z.land` and `Z.lor` (which, btw, I think are confusingly named, for they are bitwise operations, not logical ones).  I imagine we would use `~'` for `Z.lneg` if it came up.

liyishuai:
Indeed `|` broke the inductive definition:
```
From Coq Require Import Bvector ZArith.
Import BvectorNotations.

Inductive foo : Bvector 0 -> Prop :=
  Foo1 : foo Bnil
| Foo2 : foo Bnil.
```
> Error: The reference Foo2 was not found in the current environment.

herbelin:
Let me try to give extra food for a general discussion on notations for or, and, xor, neg (and my apologies to @liyishuai for trying to open the discussion beyond the scope of only this PR).

I personally feel a bit uncomfortable with using a prime symbol (i.e. `&;` etc) as it makes me feel that the notations are then auxiliary and not worth a proper symbol. I don't know however how shared is this perception.

I would e.g. prefer going in the direction of @gares' proposal, namely using `||` for `BVor` and `&&` for `BVand`, or in the direction of the second proposal of ~jasongross, namely `^|`, `^&`, `^~`. Note that a tentative alternative could be to reuse `+` and `*` but that would not work anymore for `nat` as e.g. in `Nat.lxor` and cie, since there would be no way to distinguish `xor` from `add` in this latter case just from the type.

But then, what to do for `BVxor`? If we don't follow C anyway, I start to be a bit skeptical about `^` which is also for power, and skeptical also about `^^` (also used for power in the standard library, and also a bit cryptic in my test). I would like to add `(+)` as an alternative, or even `⊕` if we accept to commit to symbols which require an input method or copy-paste to be entered. Note: see also discussion on notations for xor [here](https://en.wikipedia.org/wiki/Exclusive_or#Alternative_symbols).

What to do with `BVneg`? The symbol `~` has not the expected precedence (I don't think that we want `~ x |' y` to be interpreted by default as `~ (x |' y)`). Since the symbol is anyway prefix, my proposal would be to simply keep `BVneg` (or, to avoid the `BV` prefix, to provide an alternative name `neg` in name space `Bvector`, so that we would write `neg (x || y)`.

As discussed, the notation `=?` seems the right way to go consistently with the rest of the standard library.

Please anyone having an idea on advantages and drawbacks of this or that notation comment, so that we can collect a maximum of perceptions and make a fair synthesis. This is important I think in the context of a standard library made to scale and last.

herbelin:
@maximedenes: can we count on you to express, ideally tomorrow, your vision on the question of notations?

For instance, would there be oppositions to adopt @JasonGross second proposal, i.e. `^&`, `^|`, `^~`, as well as `^⊕` (despite it is non-pure ascii), and, of course `?=` (see my previous comment for motivations, but using `^~` for negation, which was not discussed above)?

maximedenes:
> can we count on you to express, ideally tomorrow

I'm afraid not. In the context of stdlib2, I plan to allocate some time to browse all the libraries around and make a synthesis, but that will happen later. In the context of the current stdlib, I don't have any opinion, but of course won't oppose anything, then. Note that discussion notation choices is very much dependent on the context, since constraints (conflicts, idioms, etc) are global to a library (at least).

herbelin:
@maximedenes: By library, do you mean a topic (like string manipulation, real analysis, rational numbers, linear algebra, ...), or do you mean a group of maintainers (like the Coq standard library, MathComp, std++, Fiat, MathClasses, CoRN, HoTT, ...). As long as there are different groups of people involved, having a global choice of notations seem difficult to achieve (and, as I said in other occasions, we could in this case fall back on an architecture supporting synonyms or redefinitions of notations, as e.g. done in Agda, where you can change a notation or an identifier at module import).

Thanks in advance for a confirmation of what you had in mind when using the word "library"!

@liyishuai: Assuming that @JasonGross who made the main part of the proposal I gave above agrees, would you be ok to have a `^`-prefixed naming model rather than a `'`-postfix naming model?

liyishuai:
> For instance, would there be oppositions to adopt @JasonGross second proposal, i.e. `^&`, `^|`, `^~`, as well as `^⊕` (despite it is non-pure ascii), and, of course `?=` (see my previous comment for motivations, but using ^~ for negation, which was not discussed above)?

@herbelin Did you mean `=?`, as in the previous comment?
I think the '^'-prefixed naming model is good. Changes underway.

herbelin:
> Did you mean `=?`

Yes, of course, sorry.

herbelin:
@gares: It seems to me that the discussion made in the PR addressed your initial request for changes, right? Are you satisfied as it is?

@jasongross: do you still think that a test file is necessary, now that `|` is not used anymore?

In passing (and beyond this PR), it seems worth that we eventually define the same notations for `land`, `lor`, `lxor` in `BinIntDef`, what do you think?

herbelin:
@gares, @jasongross: Since I suspect that the answers to both questions shall be positive, I shall take a silence for an approval.

herbelin (approved):

herbelin:
I'm taking the responsibility of evaluating that @gares' remarks have been addressed (effort for uniformisation done, @maximedenes contacted) and shall merge tomorrow. There is still time to say something @gares if ever I'm wrong in my evaluation.

PR #8175, opened 2018-07-27
Do not call 'auto with *' in (d) intuition but 'auto' instead.

ppedrot:
This is a kamikaze PR. Fire in the hole!

Fixes #4949.

### DIFF
--- a/theories/Init/Tauto.v
+++ b/theories/Init/Tauto.v
@@ -88,10 +88,10 @@ Local Ltac tauto_gen flags := tauto_intuitionistic flags || tauto_classical flag
 Ltac tauto := with_uniform_flags ltac:(fun flags => tauto_gen flags).
 Ltac dtauto := with_power_flags ltac:(fun flags => tauto_gen flags).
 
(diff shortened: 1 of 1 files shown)


ppedrot:
@SkySkimmer nope, I am not that stupid!

ppedrot:
Let's close this and ponder about a proper fix...

JasonGross:
I think the proper fix involves adding a variant of `info_(e)auto` which names the databases the hints are pulled from, and then doing exactly what this PR does, but also adding a thing in `Coq.Compat.Coq88` which defines `intuition` to be `intuition info_with_db_info_auto with *`, so that it is quite easy to pull the relevant databases needed from the output of the tactic.

JasonGross:
If you set things up right, and manage to get it to spit out location information, it might even be possible to write an automated translation tool which first compiles with `-compat 8.8`, aggregates the databases for each location across all files in the project, and then updates the calls to `intuition` which mention more than `core` to `intuition auto with ...`.

PR #8182, opened 2018-07-28
Handle diffs better for the "Undo" command.

jfehrle:
**Kind:** bug fix / feature.

(I'll file a bug report if that's really needed, but this is on the borderline between a bug and a feature.  Hypothetically if the spec says "Coq crashes if you do X", then it's not a bug--just a (probably) dumb spec.)

Current code shows diffs between the current proof state and the proof state before the last command/tactic was applied.  That's probably not what the user wants to see if they enter an `Undo` command--they want to see the diffs that were previously displayed for that proof state.  Diffs should really be between the current proof state and the proof state  before the last tactic was applied.

Here's what you get now: the first and third diffs correspond to the same proof state but the diff info is not the same--not helpful.

bad undo

And here is the improved behavior; the first and third diffs are the same.  Much better.

image

The "Back" command doesn't keep enough info to allow finding the correct prior proof state.  Not planning to address that unless someone else wants to provide the necessary data.  I'm not seeing how "Back" is a particularly useful command.

This change also handles going backward in CoqIDE using the "Backward One Command" button.

Note there are other cases where the diffs should use different data.  For example, `all: swap 1 2` only changes the proof view, not the underlying proof state.  More on that later...

### DIFF
--- a/stm/stm.mli
+++ b/stm/stm.mli
@@ -110,6 +110,10 @@ val add : doc:doc -> ontop:Stateid.t -> ?newtip:Stateid.t ->
   bool -> Vernacexpr.vernac_control CAst.t ->
   doc * Stateid.t * [ `NewTip | `Unfocus of Stateid.t ]
 
+(* Returns the proof state before the last tactic that was applied at or before
+the specified state.  Used to compute proof diffs. *)
+val get_prev_proof : doc:doc -> Stateid.t -> Proof.t option
+
 (* [query at ?report_with cmd] Executes [cmd] at a given state [at],
    throwing away side effects except messages. Feedback will
    be sent with [report_with], which defaults to the dummy state id *)
--- a/ide/idetop.ml
+++ b/ide/idetop.ml
@@ -211,15 +211,13 @@ let add_diffs oldp newp intf =
     { intf with fg_goals = { first_goal with goal_hyp = hyps_pp_list; goal_ccl = concl_pp } :: tl }
 
 let goals () =
-  let oldp =
-    try Some (Proof_global.give_me_the_proof ())
-    with Proof_global.NoCurrentProof -> None in
   let doc = get_doc () in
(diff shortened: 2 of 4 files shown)


jfehrle:
@gares Perhaps you should review first since the most significant changes are in Stm.ml.

stm/stm.mli
```diff
@@ -110,6 +110,9 @@ val add : doc:doc -> ontop:Stateid.t -> ?newtip:Stateid.t ->
   bool -> Vernacexpr.vernac_control CAst.t ->
   doc * Stateid.t * [ `NewTip | `Unfocus of Stateid.t ]
 
+(* Returns the proof state before the most recent proof step.  Used to compute proof diffs *)
+val get_prev_proof : doc:doc -> Proof.t option
```
gares: I don't like the API to be aware of the current state, or at least, this is where ae are going with the entire API so the new code should not be stateful.
I think you should take in input the `Stateid.t` and use `Stm.get_current_state` at call site (and the comment needs to be fixed accordingly).

gares (changes requested):

jfehrle:
@gares, I made the changes you suggested.  @ejgallego and @ppedrot, would you look at your parts--very small changes.  Thanks,

jfehrle:
The test failure appears random and unrelated to the changes.

Zimmi48:
Yes, apparently someone broke OPAM and this is the reason for the failing Windows test.

gares (approved):

jfehrle:
Thanks for your help!

Zimmi48:
Yes @gares, the failure occurred at artifact uploading. I have some code that is supposed to handle this kind of failures (https://github.com/coq/bot/blob/b2cd2b70920b86b86138c04878433b78b6962dc0/bot.ml#L516-L535) but apparently not all of them (in this case it failed silently). I'll try to improve this.

PR #8183, opened 2018-07-28
Merge script: add a sanity check for reviews.

Zimmi48:
**Kind:** infrastructure.

### DIFF
--- a/dev/tools/merge-pr.sh
+++ b/dev/tools/merge-pr.sh
@@ -198,6 +198,21 @@ if [ -z "$(git config user.signingkey)" ]; then
   warning "gpg will guess a key out of your git config user.* data"
 fi
 
+# Sanity check: at least one "APPROVED" review
+REVIEWS=$(curl -s "$API/pulls/$PR/reviews")
+APPROVED=$(echo "$REVIEWS" | jq -rc 'map(select(.state == "APPROVED"))')
+if [ "$APPROVED" = "[]" ]; then
+  error "no approved review, please approve before merging"
+  ask_confirmation
+fi
+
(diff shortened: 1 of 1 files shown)


SkySkimmer:
The merge script is already kinda slow, do we really need to add another API call? Github already blocks pushing without an approve review.

JasonGross:
> Github already blocks pushing without an approve review.

It does?  I know it blocks merging from the web interface without an approved review.  It also requires that all pushes to the master branch are merge commits of current master with an open PR which has at least one approved review?

SkySkimmer:
Something like that. Observationally I couldn't push without first 
approving and I couldn't push 2 PRs at once.

Gaëtan Gilbert

Zimmi48:
> Github already blocks pushing without an approve review.

The goal was to give a better (preventive) error message than the one you get from GitHub, especially as some developers may not realize that the pushing operation failed. I didn't realize the merge script was slow. I'm fine with closing this PR if you prefer.

The other question is do we enforce the pushing restrictions for admins as well.

>It also requires that all pushes to the master branch are merge commits of current master with an open PR which has at least one approved review?

Yes, and in fact I recently realized that it even prevents merging two PRs from the command-line then pushing. This one is a bug IMHO and I've reported it. It made me backtrack on enforcing the pushing restrictions for admins.

maximedenes (commented):
No opinion. Let's discuss it... or not.

Zimmi48:
For discussion to happen, there should be a WG. I'm sad because I liked having a regular physical meetup with you folks. Anyways, there are better things to discuss than this PR. I think I'm not going to improve on the merge script anymore: instead I'm going to come up with a merge process involving @coqbot.

PR #8184, opened 2018-07-29
Improved grammar and spelling in chapters 'Extraction', 'Program' and 'ring and field' chapters of the Reference Manual.

wkolowski:
**Kind:** documentation

Mostly grammar and spelling. More serious changes:
1. The syntax for "Solve Obligations" was incorrect. The clause "with expr" is indeed optional, but "of ident" can not stand alone by itself — it requires the "with expr" part. I fixed this.
2. I changed the explanation of reflection a bit

There's also an issue I didn't fix: in the chapter about the Program command the word "inequality" means "<>", but in other chapters it means "< <= > >=". I bet this kind of inconsistency may be also present in other chapters.

### DIFF
--- a/doc/sphinx/addendum/extraction.rst
+++ b/doc/sphinx/addendum/extraction.rst
@@ -116,13 +116,13 @@ be optimized in order to be efficient (for instance, when using
 induction principles we do not want to compute all the recursive calls
 but only the needed ones). So the extraction mechanism provides an
 automatic optimization routine that will be called each time the user
-want to generate |OCaml| programs. The optimizations can be split in two
+wants to generate an |OCaml| program. The optimizations can be split in two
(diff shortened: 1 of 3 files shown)


Zimmi48 (approved):

Zimmi48:
>There's also an issue I didn't fix: in the chapter about the Program command the word "inequality" means "<>", but in other chapters it means "< <= > >=". I bet this kind of inconsistency may be also present in other chapters.

If we believe these wikipedia pages, both meaning are corrects but I agree that it can be confusing:

https://en.wikipedia.org/wiki/Inequation
https://en.wikipedia.org/wiki/Inequality_(mathematics)

I'm not sure how to improve the consistency / clarity here.

PR #8185, opened 2018-07-30
Improved grammar and spelling in the remaining chapters of the Reference Manual.

wkolowski:
**Kind:** documentation

Mostly grammar and spelling plus some formatting (I changed the very inconsistent use of ":" in the chapter Generalized Rewriting).

There are a few issues:

1. In the chapter Nsatz, the syntax entry "nsatz with radicalmax:=num%N strategy:=num%Z parameters:= var* variables:= var*" looks very bad, not to mention that it is badly explained (no examples).

2. In the chapter Generalized rewriting, example 9 is not totally clear. For me it does not generate any subgoal. It would be nice it this example was written down explicitly. Try this code:

```coq
Require Import Setoid.

Definition eq0 (x y : nat) : Prop := x = y /\ x <> 0.

(* For [div]. *)
Require Import Nat.

Add Morphism div
with signature (eq ==> eq0 ==> eq) as div_mor.
Proof.
  intros y x. generalize dependent y.
  induction x as [| x']; destruct y0; inversion 1; auto.
Qed.

Goal forall x y n : nat, x = y -> div x n = div y n.
Proof.
  intros. rewrite H. auto.
Qed.
```
Not to mention that the examples are not using the directive ".. example". By the way, is it possible to add a title when using this directive?

3. In the chapter Generalized Rewriting, covariance and contravariance are explained badly. There are some confusing statements, like "the second occurrence of x occurs covariantly in (x / x)" (and the confusion is that the (x / x) being referred to is not a standalone expression, but a subexpression of something else).

### DIFF
--- a/doc/sphinx/addendum/generalized-rewriting.rst
+++ b/doc/sphinx/addendum/generalized-rewriting.rst
@@ -26,12 +26,12 @@ on the type class infrastructure. It also improves on and generalizes
 the previous implementation in several ways:
 
 
-+ User-extensible algorithm. The algorithm is separated in two parts:
-  generations of the rewriting constraints (done in ML) and solving of
++ User-extensible algorithm. The algorithm is separated into two parts:
+  generation of the rewriting constraints (written in ML) and solving
   these constraints using type class resolution. As type class
   resolution is extensible using tactics, this allows users to define
   general ways to solve morphism constraints.
-+ Sub-relations. An example extension to the base algorithm is the
++ Subrelations. An example extension to the base algorithm is the
   ability to define one relation as a subrelation of another so that
   morphism declarations on one relation can be used automatically for
   the other. This is done purely using tactics and type class search.
@@ -58,7 +58,7 @@ Relations and morphisms
 ~~~~~~~~~~~~~~~~~~~~~~~
 
 A parametric *relation* ``R`` is any term of type
-``forall (x1 :T1 ) ... (xn :Tn ), relation A``.
+``forall (x1 : T1) ... (xn : Tn), relation A``.
 The expression ``A``, which depends on ``x1 ... xn`` , is called the *carrier*
 of the relation and ``R`` is said to be a relation over ``A``; the list
 ``x1,...,xn`` is the (possibly empty) list of parameters of the relation.
@@ -66,25 +66,25 @@ of the relation and ``R`` is said to be a relation over ``A``; the list
 **Example 1 (Parametric relation):**
 
 It is possible to implement finite sets of elements of type ``A`` as
-unordered list of elements of type ``A``.
-The function ``set_eq: forall (A: Type), relation (list A)``
+unordered lists of elements of type ``A``.
+The function ``set_eq: forall (A : Type), relation (list A)``
 satisfied by two lists with the same elements is a parametric relation
 over ``(list A)`` with one parameter ``A``. The type of ``set_eq``
-is convertible with ``forall (A: Type), list A -> list A -> Prop.``
+is convertible with ``forall (A : Type), list A -> list A -> Prop.``
 
 An *instance* of a parametric relation ``R`` with n parameters is any term
-``(R t1 ... tn )``.
+``(R t1 ... tn)``.
 
 Let ``R`` be a relation over ``A`` with ``n`` parameters. A term is a parametric
 proof of reflexivity for ``R`` if it has type
-``forall (x1 :T1 ) ... (xn :Tn), reflexive (R x1 ... xn )``.
+``forall (x1 : T1) ... (xn : Tn), reflexive (R x1 ... xn)``.
 Similar definitions are given for parametric proofs of symmetry and transitivity.
 
 **Example 2 (Parametric relation (cont.)):**
 
 The ``set_eq`` relation of the previous example can be proved to be
 reflexive, symmetric and transitive. A parametric unary function ``f`` of type
-``forall (x1 :T1 ) ... (xn :Tn ), A1 -> A2`` covariantly respects two parametric relation instances
+``forall (x1 : T1) ... (xn : Tn), A1 -> A2`` covariantly respects two parametric relation instances
 ``R1`` and ``R2`` if, whenever ``x``, ``y`` satisfy ``R1 x y``, their images (``f x``) and (``f y``)
 satisfy ``R2 (f x) (f y)``. An ``f`` that respects its input and output
 relations will be called a unary covariant *morphism*. We can also say
@@ -92,7 +92,7 @@ that ``f`` is a monotone function with respect to ``R1`` and ``R2`` . The
 sequence ``x1 ... xn`` represents the parameters of the morphism.
 
 Let ``R1`` and ``R2`` be two parametric relations. The *signature* of a
-parametric morphism of type ``forall (x1 :T1 ) ... (xn :Tn ), A1 -> A2``
+parametric morphism of type ``forall (x1 : T1) ... (xn : Tn), A1 -> A2``
 that covariantly respects two instances :math:`I_{R_1}` and :math:`I_{R_2}` of ``R1`` and ``R2``
 is written :math:`I_{R_1} ++> I_{R_2}`. Notice that the special arrow ++>, which
 reminds the reader of covariance, is placed between the two relation
@@ -120,14 +120,14 @@ parameters is any term :math:`f \, t_1 \ldots t_n`.
 
 **Example 3 (Morphisms):**
 
-Continuing the previous example, let ``union: forall (A: Type), list A -> list A -> list A``
+Continuing the previous example, let ``union: forall (A : Type), list A -> list A -> list A``
 perform the union of two sets by appending one list to the other. ``union` is a binary
 morphism parametric over ``A`` that respects the relation instance
 ``(set_eq A)``. The latter condition is proved by showing:
 
 .. coqtop:: in
 
-  forall (A: Type) (S1 S1’ S2 S2’: list A),
+  forall (A : Type) (S1 S1’ S2 S2’ : list A),
     set_eq A S1 S1’ ->
     set_eq A S2 S2’ ->
     set_eq A (union A S1 S2) (union A S1’ S2’).
@@ -137,7 +137,7 @@ for all ``A``.
 
 **Example 4 (Contravariant morphism):**
 
-The division function ``Rdiv: R -> R -> R`` is a morphism of signature
+The division function ``Rdiv : R -> R -> R`` is a morphism of signature
 ``le ++> le --> le`` where ``le`` is the usual order relation over
 real numbers. Notice that division is covariant in its first argument
 and contravariant in its second argument.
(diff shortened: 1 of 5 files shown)


Zimmi48:
>Not to mention that the examples are not using the directive ".. example". By the way, is it possible to add a title when using this directive?

I don't think so but let's ask @cpitclaudel.

Zimmi48:
`nsatz` is missing a tactic index entry.

I suggest applying this diff:

```diff
diff --git a/doc/sphinx/addendum/nsatz.rst b/doc/sphinx/addendum/nsatz.rst
index 387d61495..c7275e025 100644
--- a/doc/sphinx/addendum/nsatz.rst
+++ b/doc/sphinx/addendum/nsatz.rst
@@ -7,6 +7,9 @@ Nsatz: tactics for proving equalities in integral domains
 
 :Author: Loïc Pottier
 
+.. tacn:: nsatz
+   :undocumented:
+
 The tactic `nsatz` proves goals of the form
 
 :math:`\begin{array}{l}
@@ -68,10 +71,9 @@ Buchberger algorithm.
 This computation is done after a step of *reification*, which is
 performed using :ref:`typeclasses`.
 
-The ``Nsatz`` module defines the tactic `nsatz`, which can be used without
-arguments, or with the syntax:
+.. tacv:: nsatz with radicalmax:=@num%N strategy:=@num%Z parameters:=[{*, @ident}] variables:=[{*, @ident}]
 
-| nsatz with radicalmax:=num%N strategy:=num%Z parameters:= :n:`{* var}` variables:= :n:`{* var}`
+   Most complete syntax for :tacn:`nsatz`.
 
 where:
```

And to reindent the following paragraphs (instead of the `:undocumented:` option in particular).

Note that the actual grammar definition of the extended `nsatz` is:

```coq
Tactic Notation "nsatz" "with"
 "radicalmax" ":=" constr(radicalmax) 
 "strategy" ":=" constr(info) 
 "parameters" ":=" constr(lparam)
 "variables" ":=" constr(lvar):=
  intros;
  try apply (@psos_r1b _ _ _ _ _ _ _ _ _ _ _);
  match goal with |- (@equality ?r _ _ _) =>
    repeat equalities_to_goal;
    nsatz_generic radicalmax info lparam lvar
  end.
```

with some Ltac code to decompose the lists of terms `lparam` and `lvar` passed as arguments (pretty horrible IMHO). And thus, my proposed tactic description above is actually an approximation.

@maximedenes Since you are in a clean-up mood, do you think that something should be done here?

cpitclaudel:
> Not to mention that the examples are not using the directive ".. example". By the way, is it possible to add a title when using this directive?

No, unfortunately (the example in the dev guide suggests it, but the "title" thus created doesn't look special :confused:) 

I think we should fix this by giving the first line title status.  I'll open a PR.

Zimmi48:
@Zeimer #8191 has been merged. So if you rebase, you can start using it.

doc/sphinx/addendum/generalized-rewriting.rst
```diff
@@ -175,11 +175,11 @@ has also been declared as a morphism.
 Adding new relations and morphisms
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-A parametric relation :g:`Aeq: forall (y1 : β1 ... ym : βm )`,
+A parametric relation :g:`Aeq: forall (y1 : β1 ... ym : βm)`,
 :g:`relation (A t1 ... tn)` over :g:`(A : αi -> ... αn -> Type)` can be
 declared with the following command:
 
-.. cmd::  Add Parametric Relation (x1 : T1) ... (xn : Tk) : (A t1 ... tn) (Aeq t′1 ... t′m ) {? reflexivity proved by refl} {? symmetry proved by sym} {? transitivity proved by trans} as @ident
+.. cmd::  Add Parametric Relation (x1 : T1) ... (xn : Tk) : (A t1 ... tn) (Aeq t′1 ... t′m) {? reflexivity proved by refl} {? symmetry proved by sym} {? transitivity proved by trans} as @ident
```
Zimmi48: If you have the courage, it would be great to reformulate this command to put the command grammar first, then the (indented) description.

doc/sphinx/addendum/generalized-rewriting.rst
```diff
@@ -226,7 +226,7 @@ replace terms with related ones only in contexts that are syntactic
 compositions of parametric morphism instances declared with the
 following command.
 
-.. cmd:: Add Parametric Morphism (x1 : T1 ) ... (xk : Tk ) : (f t1 ... tn ) with signature sig as @ident
+.. cmd:: Add Parametric Morphism (x1 : T1) ... (xk : Tk) : (f t1 ... tn) with signature sig as @ident
```
Zimmi48: Same here.

Zimmi48 (approved):

doc/sphinx/addendum/generalized-rewriting.rst
```diff
@@ -175,11 +175,11 @@ has also been declared as a morphism.
...
-A parametric relation :g:`Aeq: forall (y1 : β1 ... ym : βm )`,
+A parametric relation :g:`Aeq: forall (y1 : β1 ... ym : βm)`,
 :g:`relation (A t1 ... tn)` over :g:`(A : αi -> ... αn -> Type)` can be
 declared with the following command:
 
-.. cmd::  Add Parametric Relation (x1 : T1) ... (xn : Tk) : (A t1 ... tn) (Aeq t′1 ... t′m ) {? reflexivity proved by refl} {? symmetry proved by sym} {? transitivity proved by trans} as @ident
+.. cmd::  Add Parametric Relation (x1 : T1) ... (xn : Tk) : (A t1 ... tn) (Aeq t′1 ... t′m) {? reflexivity proved by refl} {? symmetry proved by sym} {? transitivity proved by trans} as @ident
 
 after having required the ``Setoid`` module with the ``Require Setoid``
 command.
```
Zimmi48: Just below there is a ```:g:`@ident` ``` which should become a ```:token:`ident` ```

Zimmi48:
Regarding example 9, here is how I can reach the goal that is described:

```
Require Import Setoid.

Definition eq0 (x y : nat) : Prop := x = y /\ x <> 0.

(* For [div]. *)
Require Import Nat.

Add Morphism div
with signature (eq ==> eq0 ==> eq) as div_mor.
Proof.
  intros y x. generalize dependent y.
  induction x as [| x']; destruct y0; inversion 1; auto.
Qed.

Goal forall x y n : nat, x = y -> div x n = div y n.
Proof.
  intros. apply div_mor; trivial.
  unfold eq0; split; trivial.
```

I didn't use setoid rewriting, I applied the morphism directly.
I can imagine how using `setoid_replace` could work to get a similar result, but I didn't manage to guide it to use `div_mor` instead of replacing using Leibniz equality. As for `setoid_rewrite`, it wants to solve all the obligations it generates, so I don't see how the behavior described could be exhibited using it. @mattam82 can you help us understanding this example?

Zeimer:
The diff didn't work, so I applied it manually. Also there was a conflict when I tried to add nsatz to the tactic index, so I had to rename the whole chapter (only internally) from _nsatz to _nsatz_chapter. I indented all the command descriptions. I changed the examples to use the .. example:: directive with titles and it looks way better now.

doc/sphinx/addendum/generalized-rewriting.rst
```diff
@@ -149,74 +148,71 @@ always the intended equality for a given structure.
...
-
-after having required the ``Setoid`` module with the ``Require Setoid``
-command.
+   This command declares a parametric relation :g:`Aeq: forall (y1 : β1 ... ym : βm)`,
+   :g:`relation (A t1 ... tn)` over :g:`(A : αi -> ... αn -> Type)`.
 
-The :g:`@ident` gives a unique name to the morphism and it is used
-by the command to generate fresh names for automatically provided
-lemmas used internally.
+   The :token:`@ident` gives a unique name to the morphism and it is used
```
Zimmi48: When using `:token:` you shouldn't put the `@`. The `@` is used inside `:n:` blocks to denote grammar tokens.

doc/sphinx/addendum/generalized-rewriting.rst
```diff
@@ -149,74 +148,71 @@ always the intended equality for a given structure.
...
 
-  Add Parametric Relation (A : Type) : A (@eq A)
-    [reflexivity proved by @refl_equal A]
-  ...
+     Add Parametric Relation (A : Type) : A (@eq A)
+       [reflexivity proved by @refl_equal A]
+     ...
 
 Some tactics (:tacn:`reflexivity`, :tacn:`symmetry`, :tacn:`transitivity`) work only on
 relations that respect the expected properties. The remaining tactics
```
Zimmi48: Only spotted now: there is a ``` `replace` ``` on the line below that should become a ``` :tacn:`replace` ```

doc/sphinx/addendum/generalized-rewriting.rst
```diff
@@ -226,103 +222,104 @@ replace terms with related ones only in contexts that are syntactic
...
 
-The command declares ``f`` as a parametric morphism of signature ``sig``. The
-identifier ``id`` gives a unique name to the morphism and it is used as
-the base name of the type class instance definition and as the name of
-the lemma that proves the well-definedness of the morphism. The
-parameters of the morphism as well as the signature may refer to the
-context of variables. The command asks the user to prove interactively
-that ``f`` respects the relations identified from the signature.
+   This command declares ``f`` as a parametric morphism of signature ``sig``. The
+   identifier ``id`` gives a unique name to the morphism and it is used as
```
Zimmi48: `id` should be ``` :token:`ident` ``` here

doc/sphinx/addendum/generalized-rewriting.rst
```diff
@@ -551,21 +547,20 @@ pass additional arguments such as ``using relation``.
 .. tacv:: setoid_replace @term with @term [in @ident] [using relation @term] [by @tactic]
```
Zimmi48: The optional parts should normally be enclosed in `{?` and `}` instead of `[` and `]`.

doc/sphinx/addendum/nsatz.rst
```diff
@@ -1,63 +1,55 @@
...
+   P_1(X_1, \ldots, X_n) = Q_1(X_1, \ldots, X_n), \ldots, P_s(X_1, \ldots, X_n) = Q_s(X_1, \ldots, X_n) \\
+   \vdash P(X_1, \ldots, X_n) = Q(X_1, \ldots, X_n) \\
+   \end{array}`
 
-It also proves formulas
+   where :math:`P, Q, P_1, Q_1, \ldots, P_s, Q_s` are polynomials and :math:`A` is an integral
+   domain, i.e. a commutative ring with no zero divisors. For example, :math:`A`
+   can be :math:`\mathbb{R}`, :math:`\mathbb{Z}`, or :math:`\mathbb{Q}`.
+   Note that the equality :math:`=` used in these goals can be
+   any setoid equality (see :ref:`tactics-enabled-on-user-provided-relations`) , not only Leibnitz equality.
```
Zimmi48: There is a typo there, it should be Leibniz.

Zimmi48 (commented):
Some more comments on stuff that I hadn't spotted before.

doc/sphinx/addendum/generalized-rewriting.rst
Zeimer: Got 'em all.

PR #6952, opened 2018-03-09
Issues with printing based on scope (#873): performance and use of abbreviation for printing.

herbelin:
Fixes #6777, #6536 and #6416.

Warning: it builds on top of #6247 with which it interferes.

I made some cleaning at the same time, so, it might not be unwise to not commit it in urgency on a day of freeze.

PR #6949, opened 2018-03-08
Revert PR #873: New strategy based on open scopes for deciding…

maximedenes:
This revert was decided at the last WG, because the functionality didn't seem fully ready, as it introduced some regressions. In particular, nothing prevents us from merging again once the regressions are fixed.

This (hopefully) fixes #6777, #6536 and #6416 .

@herbelin I need some help fixing the test-suite, can you take over and push to my branch? Thanks!

PR #873, opened 2017-07-11
New strategy based on open scopes for deciding which notation to use among several of them

herbelin:
Last August, Beta and Ralf asked that the scopes be taken into account for choosing which notation to use when several of them are available for the same expression. This was a very relevant remark and I finally found some time to satisfy their wish.

The current strategy to print an expression is to use the latest available notation declared for this expression. The proposed strategy is to search in the stack of open scopes for the first available notation usable for this expression. E.g.:

```coq
Delimit Scope line_scope with line.
Notation "{ }" := nil (format "{ }") : line_scope.
Notation "{ x }" := (cons x nil) : line_scope.
Notation "{ x ; y ; .. ; z }" :=  (cons x (cons y .. (cons z nil) ..)) : line_scope.
Notation "[ ]" := nil (format "[ ]") : matx_scope.
Notation "[ l ]" := (cons l%line nil) : matx_scope.
Notation "[ l ; l' ; .. ; l'' ]" :=  (cons l%line (cons l'%line .. (cons l''%line nil) ..))
  (format "[ '[v' l ; '/' l' ; '/' .. ; '/' l'' ']' ]") : matx_scope.

Open Scope matx_scope.
Check [[1;2;3];[4;5;6];[7;8;9]].
(*
[{1; 2; 3};
 {4; 5; 6};
 {7; 8; 9}]
     : list (list nat)
*)
```

At the current time, it is more at the stage of a request for comments. In particular, the question of what priority to give to abbreviations (syntactic definitions) is unclear. For instance, if I have:
```coq
Notation "@@" := Prop : symbol_scope.
Notation PROP := Prop.
Open Scope symbol_scope.
```
do I want `Prop` to be displayed `@@` (because `symbol_scope` is open last) or `PROP` (as currently)?

PR #8187, opened 2018-07-30
Notation printing based on scopes (take 2, including bug fixes)

### LINKED ISSUE #6416: Coq master is about 5%--20% slower than Coq 8.7.0 on various files in fiat-crypto, and over 100x slower at printing
#### Version

master (2c2a080, I think; c.f. #6415 for the uncertainty)

#### Operating system

Linux

On Coq 8.7.0, [this travis job (targets `selected-specific selected-specific-display` on fiat-crypto master)](https://travis-ci.org/mit-plv/fiat-crypto/jobs/315687769) completes in 37 minutes.  On Coq master [the same targets](https://travis-ci.org/mit-plv/fiat-crypto/jobs/315687773) time out after 49.5 minutes.  To pick a single file, on 8.7.0:
```
src/Specific/X25519/C64/ladderstep (real: 213.89, user: 213.15, sys: 0.52, mem: 1620916 ko)
```
on master:
```
src/Specific/X25519/C64/ladderstep (real: 279.82, user: 278.60, sys: 0.53, mem: 1587116 ko)
```
Additionally, the `src/Specific/*/*Display.v` files, which basically just do a single `Check` command take much, much longer on master (from the fact that 5 minutes passed with only three of them completing on `make -j2`, at least one such file must take at least 2.5 minutes.  On 8.7.0, they are much quicker (tens of seconds, I believe, at most).  (I have travis output `Travis keep-alive spew` every five minutes to keep the build alive even with files that take more than 10 minutes.)

### LINKED ISSUE #6536: [regression] It is no longer possible to override a notation with an alias
#### Version

master

Since #873, not only is printing hundreds of times slower on large terms, it is also the case that I can no longer override notations by declaring new aliases (and, additionally, there is no way to override notations in multiple scopes with a new notation in a single scope).  That is, the following does not work, when I expect it to:

```coq
Notation hide := (_ = _).
Goal 1 = 1.
```

I would suggest allowing aliases to be bound to scopes for printing priority purposes (e.g., `Notation hide := (_ = _) : core_scope.`), and additionally adding a flag that allows users to choose between latest-declared versus most-relevant-scope for printing.

### LINKED ISSUE #6777: Regression in notation printing
From mathcomp, `fingroup/quotient.v`:
```
Variables (gT : finGroupType) (Q A : {set gT}).
Notation H := <<A>>.
Check <<A>>.
(*
@generated gT A
     : @set_of (FinGroup.arg_finType (FinGroup.base gT))
         (Phant (Finite.sort (FinGroup.arg_finType (FinGroup.base gT))))
*)
Check H.
(*
@generated gT A
     : @set_of (FinGroup.arg_finType (FinGroup.base gT))
         (Phant (Finite.sort (FinGroup.arg_finType (FinGroup.base gT))))
*)
```
In both cases, the system prints `<<A>>` and not `H`.

I don't know when the regression took place.

herbelin:
**Kind:** feature

This is a second take at #873 which could not go in 8.8 (see revert in #6949). This second take includes #6952 which was providing fixes to blocking problems (#6777, #6536 and #6416) introduced by #873.

- [x] Unclear whether it is worth a line in the reference manual
- [X] Entry added in CHANGES.

Closes #6777
Closes #6536
Closes #6416

### DIFF
--- a/doc/sphinx/user-extensions/syntax-extensions.rst
+++ b/doc/sphinx/user-extensions/syntax-extensions.rst
@@ -1340,7 +1340,16 @@ Coq will use the following rules for printing priorities:
   have a delimiter are considered, giving priority to the most
   recently defined (or imported) ones. The corresponding delimiter is
   inserted, making the corresponding scope the most recent explicitly
-  open scope for all subterms of the current term.
+  open scope for all subterms of the current term. As an exception to
+  the insertion of the corresponding delimiter, when an expression is
+  statically known to be in a position expecting a type and the
+  notation is from scope ``type_scope``, and the latter is closed, the
+  delimiter is not inserted. This is because expressions statically
+  known to be in a position expecting a type are by default
+  interpreted with `type_scope` temporarily activated. Expressions
+  statically known to be in a position expecting a type typically
+  include being on the right-hand side of `:`, `<:`, `<<:` and after
+  the comma in a `forall` expression.
 
 - As a refinement of the previous rule, in the case of applied global
   references, notations in a non-opened scope with delimiter
--- a/doc/sphinx/user-extensions/syntax-extensions.rst
+++ b/doc/sphinx/user-extensions/syntax-extensions.rst
@@ -1040,6 +1040,8 @@ interpreted in the scope stack extended with the scope bound tokey.
    To remove a delimiting key of a scope, use the command
    :n:`Undelimit Scope @scope`
 
+.. _ArgumentScopes:
+
 Binding arguments of a constant to an interpretation scope
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
@@ -1306,6 +1308,44 @@ Displaying information about scopes
       This displays all the notations defined in the interpretation scope :token:`scope`.
       It also displays the delimiting key if any and the class to which the
       scope is bound, if any.
+
+Impact of scopes on printing
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When several notations are available for printing the same expression,
+Coq will use the following rules for printing priorities:
+
+- If two notations are available in different scopes which are open,
+  the notation in the more recently opened scope takes precedence.
+
+- If two notations are available in the same scope, the more recently
+  defined (or imported) notation takes precedence.
+
+- Abbreviations and lonely notations, both of which have no scope,
+  take precedence over a notation in an open scope if and only if the
+  abbreviation or lonely notation was defined (or imported) more
+  recently than when the corresponding scope was open. They take
+  precedence over any notation not in an open scope, whether this scope
+  has a delimiter or not.
+
+- A scope is *active* for printing a term either because it was opened
+  with :cmd:`Open Scope`, or the term is the immediate argument of a
+  constant which temporarily opens a scope for this argument (see
+  :ref:`Arguments <ArgumentScopes>`) in which case this temporary
+  scope is the most recent open one.
+
+- In case no abbreviation, nor lonely notation, nor notation in an
+  explicitly open scope, nor notation in a temporarily open scope of
+  arguments, has been found, notations in those closed scopes which
+  have a delimiter are considered. A notation is chosen and the
+  corresponding delimiter is inserted, making the corresponding scope
+  the most recent explicitly open scope for all subterms of the
+  current term.
+
+- A scope can be closed by using :cmd:`Close Scope` and its delimiter
+  removed by using :cmd:`Undelimit Scope`. To remove automatic
+  temporary opening of scopes for arguments of a constant, use
+  :ref:`Arguments <ArgumentScopes>`.
 
 .. _Abbreviations:
 
--- a/doc/sphinx/user-extensions/syntax-extensions.rst
+++ b/doc/sphinx/user-extensions/syntax-extensions.rst
@@ -1337,10 +1337,22 @@ Coq will use the following rules for printing priorities:
 - In case no abbreviation, nor lonely notation, nor notation in an
   explicitly open scope, nor notation in a temporarily open scope of
   arguments, has been found, notations in those closed scopes which
-  have a delimiter are considered. A notation is chosen and the
-  corresponding delimiter is inserted, making the corresponding scope
-  the most recent explicitly open scope for all subterms of the
-  current term.
+  have a delimiter are considered, giving priority to the most
+  recently defined (or imported) ones. The corresponding delimiter is
+  inserted, making the corresponding scope the most recent explicitly
+  open scope for all subterms of the current term.
+
+- As a refinement of the previous rule, in the case of applied global
+  references, notations in a non-opened scope with delimiter
+  specifically defined for this applied global reference take priority
+  over notations in a non-opened scope with delimiter for generic
+  applications. For instance, in the presence of ``Notation "f ( x
+  )" := (f x) (at level 10, format "f ( x )") : app_scope`` and
+  ``Notation "x '.+1'" := (S x) (at level 10, format "x '.+1'") :
+  mynat_scope.`` and both of ``app_scope`` and ``mynat_scope`` being
+  bound to a delimiter *and* both not opened, the latter, more
+  specific notation will always take precedence over the first, more
+  generic one.
 
 - A scope can be closed by using :cmd:`Close Scope` and its delimiter
   removed by using :cmd:`Undelimit Scope`. To remove automatic
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -148,6 +148,11 @@ Notations
   that will do it automatically, using the output of coqc. The script
   contains documentation on its usage in a comment at the top.
 
+- When several notations are available for the same expression,
+  priority is given to latest notations defined in the scopes being
+  opened, in order, rather than to the latest notations defined
+  independently of whether they are in an opened scope or not.
+
 Tactics
 
 - Added toplevel goal selector `!` which expects a single focused goal.
--- a/dev/doc/changes.md
+++ b/dev/doc/changes.md
@@ -152,6 +152,12 @@ Termops
 - Internal printing functions have been placed under the
   `Termops.Internal` namespace.
 
+Notations:
+
+- Notation.availability_of_notation is not anymore needed: if a
+  delimiter is needed, it is provided by Notation.uninterp_notation
+  which fails in case the notation is not available.
+
 ### Unit testing
 
 The test suite now allows writing unit tests against OCaml code in the Coq
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -263,6 +263,11 @@ let rec insert_pat_coercion ?loc l c = match l with
   | [] -> c
   | ntn::l -> CAst.make ?loc @@ CPatNotation (ntn,([insert_pat_coercion ?loc l c],[]),[])
 
+let add_lonely keyrule seen =
+  match keyrule with
+  | NotationRule (None,ntn) -> ntn::seen
+  | SynDefRule _ | NotationRule (Some _,_) -> seen
+
 (**********************************************************************)
 (* conversion of references                                           *)
 
@@ -387,7 +392,7 @@ let rec extern_cases_pattern_in_scope (custom,scopes as allscopes) vars pat =
   with No_match ->
     try
       if !Flags.in_debugger || !Flags.raw_print || !print_no_symbol then raise No_match;
-      extern_notation_pattern allscopes vars pat
+      extern_notation_pattern allscopes [] vars pat
         (uninterp_cases_pattern_notations scopes pat)
     with No_match ->
     let loc = pat.CAst.loc in
@@ -441,13 +446,14 @@ let rec extern_cases_pattern_in_scope (custom,scopes as allscopes) vars pat =
       insert_pat_coercion coercion pat
 
 and apply_notation_to_pattern ?loc gr ((subst,substlist),(nb_to_drop,more_args))
-    (custom, (tmp_scope, scopes) as allscopes) vars =
+    (custom, (tmp_scope, scopes) as allscopes) lonely_seen vars =
   function
-    | NotationRule (sc,ntn),key ->
+    | NotationRule (sc,ntn),key,need_delim ->
       begin
         match availability_of_entry_coercion custom (fst ntn) with
         | None -> raise No_match
         | Some coercion ->
+            let key = if need_delim || List.mem ntn lonely_seen then key else None in
             let scopt = match key with Some _ -> sc | _ -> None in
 	    let scopes' = Option.List.cons scopt scopes in
 	    let l =
@@ -470,8 +476,8 @@ and apply_notation_to_pattern ?loc gr ((subst,substlist),(nb_to_drop,more_args))
               (insert_pat_delimiters ?loc
                  (make_pat_notation ?loc ntn (l,ll) l2') key)
       end
-    | SynDefRule kn,key ->
-      assert (key = None);
+    | SynDefRule kn,key,need_delim ->
+      assert (key = None && need_delim = false);
       match availability_of_entry_coercion custom InConstrEntrySomeLevel with
       | None -> raise No_match
       | Some coercion ->
@@ -489,9 +495,9 @@ and apply_notation_to_pattern ?loc gr ((subst,substlist),(nb_to_drop,more_args))
       in
       assert (List.is_empty substlist);
       insert_pat_coercion ?loc coercion (mkPat ?loc qid (List.rev_append l1 l2'))
-and extern_notation_pattern allscopes vars t = function
+and extern_notation_pattern allscopes lonely_seen vars t = function
   | [] -> raise No_match
-  | (keyrule,pat,n as _rule,key)::rules ->
+  | (keyrule,pat,n as _rule,key,need_delim)::rules ->
     try
       if is_inactive_rule keyrule then raise No_match;
       let loc = t.loc in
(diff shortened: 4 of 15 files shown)


JasonGross (changes requested):
I'd like to see documentation for this in the reference manual.  In particular, I think the refman should include:
- The search order for printing notations
- A mention of where lonely notations fall in the search order
- A mention of where abbreviations (`Notation` without `"`s) fall in the search order
- A mention of what is required to remove a notation from the search order (as far as I'm aware, in 8.8, you need `Close Scope` (if the scope is open) + `Undelimit Scope` (if the scope is delimited) + `Arguments : clear implicits` or `Unbind Scope` (does this exist?) (if the scope is bound))

Zimmi48:
Is the update to `test-suite/.csdp.cache` really related?

herbelin:
> Is the update to test-suite/.csdp.cache really related?

Indeed, no. Thanks for noticing.

@JasonGross: I understand your concerns... I'll try to do this, but that may require some thinking and thus time that I shall probably not have in the short term.

ejgallego:
@herbelin I assume that the performance problems we witnessed last time should be solved, right? 

@JasonGross @herbelin , what would be a reasonable compromise to try to advance with this PR?

JasonGross:
I'd be fine if the following text appeared somewhere in the reference manual (assuming it's correct; part of why I'm asking for it in the doc is that I'm not entirely sure what the search order is; what's below reflects what I think it ought to be):

```
When several notations are available for printing the same expression, Coq will use the following rules for printing priorities:

- If two notations are available in different scope which are open, the notation in the more recently opened scope takes precedence.
- If two notations are available in the same scope, the more recently defined (or imported) notation takes precedence.
- Abbreviations and lonely notations, both of which have no scope, take precedence over a notation in a scope if and only if the abbreviation or lonely notation was defined more recently than the notation in the scope.
- Notations which live in scopes that are closed are never used for printing.
```

ejgallego:
@JasonGross 's review still seems to apply.

herbelin:
@jasongross: Thanks for your text. I shall add a note on the use of delimiters when a scope is closed, and a change to item 3. But before that, I'm still wondering what to do here:
```coq
Declare Scope mynat_scope.
Delimit Scope mynat_scope with mynat.
Notation "# x" := (S x) (at level 20).
Notation "## x" := (S x) (at level 20) : mynat_scope.
Check fun x => S x.
```
Should we see `fun x => # x` (current behavior with this PR, which gives precedence to lonely notations over notations in scope) or `fun x => (## x)%mynat` (current behavior with master, which gives precedence to order of definition).

herbelin:
> Should we see `fun x => # x` (current behavior with this PR, which gives precedence to lonely notations over notations in scope) or `fun x => (## x)%mynat` (current behavior with master, which gives precedence to order of definition).

I opted for `fun x => # x`.

@JasonGross: I added the documentation on printing priority rules. Tell me if it is ok for you.

Zimmi48 (approved):

ejgallego:
Needs fixing it seems @herbelin

ejgallego:
I will also launch a bench, just in case; after the bench is completed I'll merge.

herbelin:
CI ok, "benchmark-part-of-the-branch" started with number 526.

ejgallego:
> CI ok, "benchmark-part-of-the-branch" started with number 526.

Thanks @herbelin . To what me respects this seems ready to go. I'll let some time so @JasonGross can comment, will do another review myself and then merge hopefully ASAP.

ejgallego:
@JasonGross please re-review, thanks.

JasonGross (changes requested):
Looks good modulo the small changes I'm requesting for increased accuracy

ejgallego:
@herbelin waiting for a new revision.

ejgallego:
Folks maybe we could move this so we can merge quick? The PR has been ready for long only pending some minor doc updates.

herbelin:
I had some comments to take into account. Let me try to do it now.

ejgallego:
Merci @herbelin.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1305,6 +1305,30 @@ Displaying information about scopes
...
+
+- Abbreviations and lonely notations, both of which have no scope,
+  take precedence over a notation in a scope if and only if the
+  abbreviation or lonely notation was defined (or imported) more
+  recently than when the corresponding scope was open, or, if the
+  scope is not open and has a delimiter, more recently than when the
+  notation in the scope was defined.
+
+- Notations which live in scopes that are closed and which have no
+  delimiters are never used for printing. A scope can be closed by
```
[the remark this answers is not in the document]
herbelin: @jasongross: I fixed the inaccuracy and extended the explanations. However, I realized that the new semantics adds a case when notations are chosen "arbitrarily"  by looking in a map ordered alphabetically, which basically says that the notation is chosen based on the alphabetical order of the scope in which it lives. Here is a typical example:
```
Declare Scope a.
Declare Scope c.
Declare Scope b.
Notation "#" := None : a.
Notation "###" := None : c.
Notation "##" := None : b.
Delimit Scope a with a.
Delimit Scope c with c.
Delimit Scope b with b.
Check None.
(* ### *)
```
I.e., scope `c` is chosen because last in alphabetic order, independently of the order of declarations.

This does not look satisfactory. I'll see how I can change this.

herbelin:
Note: In doubt, I kept the item in `Changes.md` in section 8.9.0beta but I suspect that it is too late for 8.9.0. What are the news about it?

ejgallego:
> What are the news about it?

No idea. As @Zimmi48 placed this on the 8.10 deadline I suggest you accordingly move the changes entry. When they backport it should be corrected I think, and we can always amend master in each 8.9.x release.

doc/sphinx/user-extensions/syntax-extensions.rst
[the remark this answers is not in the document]
ejgallego: @herbelin I am adding thus a needs progress label to this PR. Please remove it when yo consider it ready for merge consideration again.

Zimmi48:
Sorry for changing the milestone while I'm not the RM and thus not the one taking the decision here. I thought it was pretty clear to everyone that it was too late for 8.9. Guillaume has been denying backporting to most feature PRs (he only backported when there were strong reasons to do so). But anyways, that's not my call as I said.

doc/sphinx/user-extensions/syntax-extensions.rst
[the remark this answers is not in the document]
herbelin: @jasongross: It remains to decide of a rule for ordering notations which are not in an open scope. Maybe as before, i.e. in the order of registration of notations, unless you have arguments for a different order?

herbelin:
@JasonGross: I opted for giving priority to most recent notations in case of notations in a non-open scope with a delimiter. I extended the documentation with an additional (rare) particular case about generic applicative notations. Please tell if this is of for you.

ejgallego:
@herbelin is there a risk of efficiency / run time problems with this new approach?

ejgallego:
Test suite seems broken.

ejgallego:
> Test suite seems broken.

Also printing problem in lambdaRust

herbelin:
> Also printing problem in lambdaRust

The printing change in Iris is questioning. There is a notation in scope `stdpp_scope` which is used non ambiguously for parsing a statement but this notation starts with a `forall`, so that, at printing time, the unicode notation for `forall` in scope `type_scope` takes precedence. If `type_scope` becomes prioritary, it is because the goal is a statement, what activates the scope `type_scope`, thus hiding the (more refined) notation (namely `"'{{{' P } } } e {{{ x .. y , 'RET' pat ; Q } } }"`) in `stdpp_scope`. With the priority based on scope put in place with this PR, I don't see how the user can get control on `type_scope`. So some thinking is needed.

ejgallego:
Thanks for the analysis, it is very interesting, Iris is indeed a very advanced user of notations, sometimes I wonder if they would not better served by a plugin. Anyways, shouldn't they wrap the `forall` on a constructor so this works a bit better?

cc: @RalfJung 

p.s: You may want to run a rebase to trigger the CI.

RalfJung:
Cc @robbertkrebbers

RalfJung:
@herbelin Which notations are we talking about that are causing problems? Are we talking about [Texan triples](https://gitlab.mpi-sws.org/FP/iris-coq/blob/master/theories/bi/weakestpre.v#L74)?
I don't understand why this should need a constructor wrapping `forall`; it doesn't wrap any of the other connectives (`-∗`, `▷`, `□` and `WP`) either. Not being able to have any notation involving `forall` or `->` sounds like a rather big regression.

If this is about scopes determining precedence, why does `stdpp_scope` not take precedence over `type_scope`?

(I am also surprised to see a behavior-changing new feature being scheduled for inclusion in 8.9 longer after that got branched off.)

robbertkrebbers:
To add to the latter, in case of equal precedence, I would be fine of falling back to declaration/scope opening order, but this should be a last resort.

Zimmi48:
>(I am also surprised to see a behavior-changing new feature being scheduled for inclusion in 8.9 longer after that got branched off.)

I'm ready to bet that it will never get into 8.9: that's why I initially changed the milestone. But indeed, that's not my call.

herbelin:
@RalfJung, @robbertkrebbers: I adopted a different strategy for printing terms known to be types. It does not open the scope `type_scope` any more. Instead, it remembers that if, after exploring notations in the current open scopes, it happens to use a notation from the scope `type_scope` it does not need a delimiter. This should address the Iris issue.

About the idea of using the stack of open scopes to prioritize notations, this arrived as the conclusion of a discussion with @beta-ziliani and @RalfJung (see the examples it allows in test file [Notations3.v](https://github.com/coq/coq/blob/c3007e5063dba40bb0ca0c94dd61b39715846a42/test-suite/output/Notations3.v#L420-L473)). By using the order of opening of scopes rather than ordering of notations, the idea is also that is easy to `Open`/`Close` a scope manually when needed, or to use `%some_scope` in notations to force the scope used for printing a subterm of a notation. However, other ways to give priorities to notations can be imagined and, @robbertkrebbers, if you have some idea of a language for ordering notations, that may be worth.

robbertkrebbers:
> if you have some idea of a language for ordering notations, that may be worth.

The easiest thing to do would be to give both scopes and notations an (optional) level as a natural number (much like the levels for parsing). For the case of scopes, this can be part of the new `Declare Scope` command; a proposed syntax is `Declare Scope foo_scope (at printing level X)`. For notations you could do something like `Notation ... := ... (at printing level X)`.

FWIW: I don't really like explicit levels, and would much rather see something where you can write down ordering constraints. However, for parsing we also have explicit levels, so it would make sense to stick with that for this feature too.

herbelin:
@robbertkrebbers: maybe could we first collect examples which would benefit from an explicit ordering?

I'm now reorganizing the code so that the rules for ordering notations are centralized, so, that should be relatively easy to eventually take into account constraints.

There is another "natural" order on notations which is pattern inclusion. So, when a notation abbreviated a larger subterm than another (e.g. a notation for `@eq nat _ _` would capture a larger subterm than a notation for `@eq _ _ _`), one could decide to use the more specialized one. 
I have no idea whether this is what we would like in general or not. Here also would we need to collect examples.

herbelin:
@jasongross: I made a new update on the documentation of printing, mentioning the case of `type_scope`. Tell me if ok for you and if I can dismiss your request.

There shall still be something to clarify, which is the priority to give to abbreviations. For instance, there is in `ZArith` an abbreviation `Notation Z.pos := Zpos` (in `Coq.ZArith.BinInt`) which creates a conflict with the primitive notation for `Z`: for instance `Zpos xH` is currently printed `1%Z` because natural numbers notation are prioritary, but, if we would use scopes to order natural numbers notation and abbreviations, the abbreviation would become prioritary and one would get `Z.pos 1`. [If this is the only example, we can consider that this a library design issue though.]

Beside the question on abbreviations, there is potentially the question of efficiency. The slowness in #6416 is fixed afaicj but notation printing remains pretty sequential. If there are still slow examples, we can certainly optimize the algorithm (e.g. by building the list of possible notations lazily, or by indexing the notations on a more diverse class of keys).

In any case, I'm afraid that improving notations is endless, so I would be ok to merge this now, even more that it is the beginning of the 8.10 cycle and there will be time to address new possible questions related to this PR if needed.

robbertkrebbers:
@herbelin Consider the following from Iris:

```
wand : iProp → iProp → iProp.
fupd : coPset → coPset → iProp → iProp.
```

We have the following notations:

```
(P -∗ Q) := (wand P Q).
(|={E1,E2}=> P) := (fupd E1 E2 P).
(|={E}=> P) := (fupd E E P).
(P ={E1,E2}=∗ Q) := (P -∗ |={E1,E2}=> Q).
(P ={E}=∗ Q) := (P -∗ |={E}=> Q).
```

The notations that are further below are preferred.

Right now, we get this to work by essentially putting them in the right order. However, that's already very fragile, since the first notation is in another file than the last 4 notations. Worse, there are even more overlapping notations that are in different files and controlling the order is near impossible.

Furthermore, although the order is intuitively something like pattern order, I think that will be very hard to be made precise. Note that for many notations the top-level term is different.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1307,6 +1309,57 @@ Displaying information about scopes
...
+  the notation in the more recently opened scope takes precedence.
+
+- If two notations are available in the same scope, the more recently
+  defined (or imported) notation takes precedence.
+
+- Abbreviations and lonely notations, both of which have no scope,
+  take precedence over a notation in an open scope if and only if the
+  abbreviation or lonely notation was defined (or imported) more
+  recently than when the corresponding scope was open. They take
+  precedences over any notation not in open scope, whether this scope
```
JasonGross: "not in open scope" -> "not in an open scope"

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1307,6 +1309,57 @@ Displaying information about scopes
...
+  has a delimiter or not.
+
+- A scope is active for printing a term either because it was opened
```
JasonGross: Perhaps emphasize "active" (or "active for printing a term") with italics, to indicate that this is the definition of that phrase.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1307,6 +1309,57 @@ Displaying information about scopes
...
+
+- In case no abbreviation, nor lonely notation, nor notation in an
+  explicitly open scope, nor notation in a temporarily open scope of
+  arguments, has been found, notations in those closed scopes which
+  have a delimiter are considered, giving priority to the most
+  recently defined (or imported) ones. The corresponding delimiter is
+  inserted, making the corresponding scope the most recent explicitly
+  open scope for all subterms of the current term. As an exception to
+  the insertion of the corresponding delimiter, when an expression is
+  statically known to be in position of type and the notation is from
```
JasonGross: "statically known to be in position of type":
1. I think "in a position expecting a type" is better grammatically
2. The full enumeration of this is "on the right-hand-side of `:`, `<:`, or `<<:`, and after the comma in `forall ..., ....`", right?  Did I miss any?  Perhaps this should be elaborated here?

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1307,6 +1309,57 @@ Displaying information about scopes
...
+  scope ``type_scope``, and the latter is closed, the delimiter is not
+  inserted. This is because expressions statically known to be in
+  position of type are by default interpreted with `type_scope`
```
JasonGross: "in position of type" -> "in a position expecting a type"

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1307,6 +1309,57 @@ Displaying information about scopes
...
+  temporarily activated.
+
+- As a special rule of the previous item, in case of applied global
```
JasonGross: "special rule" -> "special case", maybe?  (I've never heard "special rule of" used before)
"in case of" -> "in the case of"

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1307,6 +1309,57 @@ Displaying information about scopes
...
+  references, notations in a non-opened scope with delimiter
+  specifically defined for this applied global reference are
+  prioritary over notations in a non-opened scope with delimiter for
```
JasonGross: spelling and grammar: "are prioritary" -> "take priority"

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1307,6 +1309,57 @@ Displaying information about scopes
...
+  generic applications (such as e.g. ``Notation "f ( x )" := (f x) (at
```
JasonGross: This needs a more complete example.  I think I understand it, but I'm not sure.  You're saying that in
```coq
Notation "x '.+1'" := (S x) (at level 10, format "x '.+1'") : mynat_scope.
Notation "f ( x )" := (f x) (at level 10, format "f ( x )") : app_scope.
```
the notation `.+1` takes priority over `f(x)`, even though the latter was declared more recently and neither scope is open?

Also, how is this a special case of the previous item?

JasonGross (approved):
Doc looks good to me, modulo the minor suggestions I've made.

ejgallego:
@herbelin , should we merge or do you consider @robbertkrebbers 's remarks in need to be addressed?

If we shall merge, please rebase and I'll merge soon.

Zimmi48:
I suggest taking into account the suggestions by Jason before merging.

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: I'm afraid that giving an exact list may easily become out-of-sync with the implementation. Another case would be e.g. in the `return` clause of a `match` (a quick answer is obtained by grepping `intern_type` and `interp_type`). I wrote `Expressions statically known to be in a position expecting a type typically include being on the right-hand side ...`. Does it look acceptable to you?

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: I put `a refinement of the previous rule`. Would that be ok?

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: > This needs a more complete example.

I elaborated. Tell me if ok for you.

I otherwise took into account all other local corrections you proposed.

ejgallego:
This looks ready to me, I will wait a bit to see if @JasonGross would like to reply to @herbelin , in the absence of more comments I will merge tomorrow [hoping no conflicts arise]

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1307,6 +1309,65 @@ Displaying information about scopes
...
+  the notation in the more recently opened scope takes precedence.
+
+- If two notations are available in the same scope, the more recently
+  defined (or imported) notation takes precedence.
+
+- Abbreviations and lonely notations, both of which have no scope,
+  take precedence over a notation in an open scope if and only if the
+  abbreviation or lonely notation was defined (or imported) more
+  recently than when the corresponding scope was open. They take
+  precedences over any notation not in an open scope, whether this scope
```
JasonGross:
```suggestion
  precedence over any notation not in an open scope, whether this scope
```

JasonGross (approved):
Doc looks good, modulo the one typo that I've noted ("take precedence" is the plural form of the singular "takes precedence")

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: Fixed. Thanks.

ejgallego:
Failure is spurious,merging.

Issue #8207, opened 2018-08-03
Overwriting previous delimiting key; Ambiguous paths ... are not official Warning messages


andrew-appel:
#### Description of the problem

Run this command (assuming the standard mathcomp installation for Coq 8.8.0):
`Require Import mathcomp.ssreflect.ssrbool.`

and I get these messages:
```
Overwriting previous delimiting key bool in scope bool_scope
Ambiguous paths:
[pred_of_mem_pred; sort_of_simpl_pred] : mem_pred >-> pred_sort
```

The word "Warning" does not appear, and in particular there is no indication of what Set Warning flag controls this message.

That part seems to be a Coq bug.  

For the ssreflect team, I wonder whether it's possible to write a line such as, 
`From mathcomp.ssreflect Require Import ssreflect seq ssrbool ssrnat ssrfun eqtype seq fintype finfun.
`
and not get any warning messages at all.  Am I doing something wrong?

tchajed:
@andrew-appel Thanks for the bug report! I'm always a fan of improving error messages.

These two warnings not using the new configurable warning system is a clear bug, so I submitted PRs to fix those two issues.

However, this doesn't address the fact that ssreflect triggers these warnings. The overwritten scope is bool (from the Coq prelude) being replaced with B in ssrbool and the ambiguous coercions paths are just an artifact of ssreflect's coercions. I'm not sure what ssreflect could do to avoid these, other than locally disabling the warnings since they're probably harmless for these two cases.

Zimmi48:
Yes, we'll probably have to disable these warnings locally since they are a conscious choice. The longer term solution (i.e. in stdlib2) is to be consistent between the choices made in Coq's prelude and in SSR (cc @gares @amahboubi @maximedenes).

pi8027:
I believe that this problem has been fixed in #9743. May I close?

gares:
> For the ssreflect team, I wonder whether it's possible to write a line such as,

As of today there is no way :-/

PR #8210, opened 2018-08-03
Make overwrite notation scope key a warning

tchajed:
Previously was a feedback and was therefore not configurable using the
warning system.

**Kind:** feature (fix to user messages).

Partial fix for #8207

- [ ] Entry added in CHANGES.

No CHANGES entry seemed necessary; this is part of making the new warning system apply to more existing error messages.

### DIFF
--- a/doc/sphinx/user-extensions/syntax-extensions.rst
+++ b/doc/sphinx/user-extensions/syntax-extensions.rst
@@ -987,6 +987,8 @@ interpreted in the scope stack extended with the scope bound tokey.
    To bind a delimiting key to a scope, use the command
    :n:`Delimit Scope @scope with @ident`
 
+   .. warn:: Overwriting previous delimiting key @ident in scope @scope
+
 .. cmd:: Undelimit Scope @scope
 
    To remove a delimiting key of a scope, use the command
--- a/interp/notation.ml
+++ b/interp/notation.ml
@@ -191,6 +191,11 @@ let make_current_scopes (tmp_scope,scopes) =
 (**********************************************************************)
 (* Delimiters *)
 
+let warn_overwrite_scope scope =
+  CWarnings.create ~name:"overwrite-delimit-key" ~category:"parsing"
+    (fun oldkey -> str "Overwriting previous delimiting key " ++ str oldkey ++
+                   str " in scope " ++ str scope)
+
 let declare_delimiters scope key =
   let sc = find_scope scope in
   let newsc = { sc with delimiters = Some key } in
@@ -199,9 +204,8 @@ let declare_delimiters scope key =
     | Some oldkey when String.equal oldkey key -> ()
     | Some oldkey ->
         (** FIXME: implement multikey scopes? *)
-	Flags.if_verbose Feedback.msg_info
-	  (str "Overwriting previous delimiting key " ++ str oldkey ++ str " in scope " ++ str scope);
-	scope_map := String.Map.add scope newsc !scope_map
+      warn_overwrite_scope scope oldkey;
+      scope_map := String.Map.add scope newsc !scope_map
   end;
   try
     let oldscope = String.Map.find key !delimiters_map in
--- a/plugins/ssr/ssrbool.v
+++ b/plugins/ssr/ssrbool.v
@@ -11,6 +11,7 @@
 (* This file is (C) Copyright 2006-2015 Microsoft Corporation and Inria. *)
 
 Require Bool.
+
 Require Import ssreflect ssrfun.
 
 (******************************************************************************)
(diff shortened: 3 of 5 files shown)


interp/notation.ml
```diff
@@ -199,9 +204,9 @@ let declare_delimiters scope key =
     | Some oldkey when String.equal oldkey key -> ()
     | Some oldkey ->
         (** FIXME: implement multikey scopes? *)
-	Flags.if_verbose Feedback.msg_info
-	  (str "Overwriting previous delimiting key " ++ str oldkey ++ str " in scope " ++ str scope);
-	scope_map := String.Map.add scope newsc !scope_map
+      warn_overwrite_scope scope oldkey
+      ;
+  scope_map := String.Map.add scope newsc !scope_map
```
Zimmi48: The indentation is very odd here, and why the semi-colon `;` is on a newline?

Zimmi48 (dismissed):
Apart from the minor indentation remark, this LGTM :sparkles:

Zimmi48:
Because you are creating a new warning and the standard library must compile without warnings, you'll have to fix the standard library. That's the reason for the failing CI.

tchajed:
Fixed the indentation issue and also locally disabled the warning in ssreflect. Here it's pretty clear that ssreflect wants a different delimiting key, which seems fine and innocuous.

maximedenes (changes requested):
A warning should not report false positives. A way to override a delimiter without triggering the warning should be provided, or the warning should be turned off by default.

Zimmi48:
>A warning should not report false positives.

Certainly not a false positive indeed. Are you simply asking that the standard library be fixed in a way other than locally disabling the warning?

Zimmi48:
@tchajed There are still issues to fix in the test-suite. You can probably use the target `accept-output` to do so (see https://github.com/coq/coq/tree/master/test-suite#fixing-output-tests).

tchajed:
@herbelin thanks for the tip, that avoids the warning and clarifies intent

The test suite now passes locally.

herbelin:
> overwrite notation scope key

Actually, I realize that this message/warning is misleading. It affects only printing, for which the new key is now used, but not parsing where both keys are available. So, the message should rather be something like "Key newkey shall be used instead of oldkey for printing in scope myscope whenever a delimiter is needed.", or the semantics be changed to override the key also for parsing, but supporting several keys for a given scope does not seem bad a priori, so, I would say, better to change only the message.

Zimmi48:
I think it is related to @herbelin's last comment but the SSR "fix" is apparently breaking external developments. For instance, GeoCoq now fails with:

```
File "./algebra/ssralg.v", line 4110, characters 16-45:
Error: Unknown scope delimiting key bool.
```

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -987,6 +987,8 @@ interpreted in the scope stack extended with the scope bound tokey.
    To bind a delimiting key to a scope, use the command
    :n:`Delimit Scope @scope with @ident`
 
+   .. warn:: Overwriting previous delimiting key @ident in scope @scope
```
Zimmi48: Add `:undocumented:` on the next line (to replace the missing description). We are trying to avoid all objects without a body but we can't turn the warning on yet because there are many remaining.

gares (approved):

Zimmi48:
It feels like this PR is in a dead-lock. @maximedenes opposed but I'd like a clarification from him, @herbelin proposed a solution to avoid disabling the warning but this doesn't seem to be backward compatible. In any case, this is probably going to miss 8.8.2.

maximedenes:
> @maximedenes opposed but I'd like a clarification from him

My point was that there should be backward-compatible fixes that don't involve disabling the warning, in both stdlib and ssr.

herbelin:
> My point was that there should be backward-compatible fixes that don't involve disabling the warning, in both stdlib and ssr.

What kind of fixes are you thinking about? Fixes in `ssrbool.v`?

Anyway, I don't see the point of providing `B` in addition to `bool` in ssr, or at least not in standard. The same way as we tend to require an explicit `Import FooBarNotations` to get some notations visible, I don't see the need for wanting a `B` unless we start to intensively delimit boolean scopes.

So, my recommendation so that this PR makes progresses is: 1) to change the warning so that it reflects what exactly happens (i.e. overriding the forme key only for printing), and 2) to make the availability of `B` an explicit request (or to move it to MathComp).

herbelin (changes requested):
See my comments in the discussion. In particular, the message should be modified.

herbelin:
@gares: any objection to moving the definition of `B` to math-comp?

ppedrot:
This PR has stalled. Any updates?

ppedrot:
No answer, so closing the PR. Please reopen when updated.

PR #8211, opened 2018-08-03
Make ambiguous coercion paths a warning

tchajed:
Previously was a feedback message and therefore not configurable with
the warning system.

**Kind:** feature (fix for user messages)

Partial fix for #8207

- [ ] Entry added in CHANGES.

No CHANGES entry seemed necessary; this is part of making the new warning system apply to more existing error messages.

### DIFF
--- a/pretyping/classops.ml
+++ b/pretyping/classops.ml
@@ -327,6 +327,9 @@ let message_ambig env sigma l =
   str"Ambiguous paths:" ++ spc () ++
   prlist_with_sep fnl (fun ijp -> print_path env sigma ijp) l
 
+let warn_ambiguous_paths env sigma =
+  CWarnings.create ~name:"ambiguous-paths" ~category:"typechecker" (message_ambig env sigma)
+
 (* add_coercion_in_graph : coe_index * cl_index * cl_index -> unit
                          coercion,source,target *)
 
@@ -380,8 +383,8 @@ let add_coercion_in_graph env sigma (ic,source,target) =
       old_inheritance_graph
   end;
   let is_ambig = match !ambig_paths with [] -> false | _ -> true in
-  if is_ambig && not !Flags.quiet then
-    Feedback.msg_info (message_ambig env sigma !ambig_paths)
+  if is_ambig then
+    warn_ambiguous_paths env sigma !ambig_paths
 
 type coercion = {
   coercion_type   : coe_typ;
--- a/doc/sphinx/addendum/implicit-coercions.rst
+++ b/doc/sphinx/addendum/implicit-coercions.rst
@@ -138,7 +138,7 @@ Declaring Coercions
   .. exn:: @qualid does not respect the uniform inheritance condition.
   .. exn:: Found target class ... instead of ...
 
-  .. warn:: Ambiguous path.
+  .. warn:: Ambiguous paths: ...
 
      When the coercion :token:`qualid` is added to the inheritance graph,
      invalid coercion paths are ignored; they are signaled by a warning
--- a/plugins/ssr/ssrbool.v
+++ b/plugins/ssr/ssrbool.v
@@ -1196,7 +1196,11 @@ Definition in_mem T x mp := nosimpl pred_of_mem T mp x.
 
 Prenex Implicits mem.
 
+(* This coercion, together with [sort_of_simpl_pred], is ambiguous. However,
+[pred_of_mem] will be used as the coercion [mem_pred >-> pred_sort]. *)
+Set Warnings "-ambiguous-paths".
 Coercion pred_of_mem_pred T mp := [pred x : T | in_mem x mp].
+Set Warnings "ambiguous-paths".
 
 Definition eq_mem T p1 p2 := forall x : T, in_mem x p1 = in_mem x p2.
 Definition sub_mem T p1 p2 := forall x : T, in_mem x p1 -> in_mem x p2.
--- a/test-suite/output/AmbiguousPathsWarning.out
+++ b/test-suite/output/AmbiguousPathsWarning.out
@@ -1,0 +1,3 @@
+File "stdin", line 3, characters 0-41:
+Warning: Ambiguous paths: [A_to_unit'] : A >-> unit
+[ambiguous-paths,typechecker]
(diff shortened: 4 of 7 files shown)


Zimmi48 (approved):

Zimmi48:
As for the other PR, you will need to fix the standard library to not emit this warning.

tchajed:
I locally disabled the warning for ssreflect, but I don't completely understand how the ambiguous paths are resolved. Using `Print Coercion Paths mem_pred simpl_pred` I was able to confirm that `pred_of_mem_pred` is actually being used as a coercion, though. It would be nice if someone that understood coercions and/or what ssreflect is doing here could explain so that I can put the explanation in the comment.

maximedenes (dismissed):
Same remark as for the other warning: a warning should not report false positives. If this is a legit use case, the warning should not be triggered.

tchajed:
@maximedenes I don't understand your concern here. Isn't the whole point of having a warning rather than an error is that it isn't always a bug in the code? There are lots of cases where a programming language cannot determine whether a use case is legit or not. In this case I believe Coq always takes the first path and ambiguous paths are ignored. In ssrbool the coercion creates one ambiguous path, but this path was never the desired coercion and so the warning is spurious, but there's no way for Coq to know that this was intentional.

It would be great if ssreflect didn't trigger this warning, but I don't think fixing ssreflect should be required to fix this bug and make it a proper warning. We might need a hacky solution like disabling the warning for ssreflect for now, but I think it's a separate issue to make ssreflect not trigger this warning at all, and one that I don't know how to solve.

As far as making this a warning enabled by default, I would prefer to be warned about ambiguous paths. Experts that know what they're doing can disable the warning.

tchajed:
I did discover one issue with locally disabling this warning: importing this file still triggers the warning, unless the importing file has the warning disabled. This is annoying and almost defeats the purpose of locally disabling; perhaps ssreflect needs a `Global Set Warnings "-ambiguous-paths".`, which would be unfortunate but I think it's no worse than the way things were before.

JasonGross:
>  ssreflect needs a Global Set Warnings "-ambiguous-paths"., which would be unfortunate but I think it's no worse than the way things were before.

This would mean that any file depending on any file which uses ssr would not warn on overlapping coercions, which seems rather unfortunate.  I want at least a warning message if I create a coercion which is not going to be used, even in files where I use ssr tactics.

tchajed:
Fair enough; I think the current behavior should be changed so that locally disabling a warning has the same effect when the module is imported, and until that happens users of ssreflect will have to tolerate this warning upon importing ssrbool (including as a transitive dependency).

I agree with you that the warning from library code is better than mysteriously having warnings changed by imports.

Zimmi48:
Let's tag this as `needs: independent fix` because of the import problem. @maximedenes may have an idea about this.

gares (approved):
I'm OK with the change in ssrbool.v

This warning is clearly a border line one. What makes it so is, in my opinion, the message. The user does not care that the path is ambiguous. The user cares about which coercion path Coq will take. IMO the message should be made more precise telling the user which path will be taken (multiple paths from A to B: list... Path x will be used).

I think that warning options should be supported as an atttribute as in `#[warnings "-amb.."] Coercion...`.

Zimmi48:
>I think that warning options should be supported as an atttribute as in #[warnings "-amb.."] Coercion....

Massive :+1:

Zimmi48:
It looks like this is going to stay blocked for a while because of the import problem.

herbelin:
Is there a good solution with imports? Would it work for instance, if, any time a path is hidden, we register that fact in the lib_stk and the .vo file, so that, at import, we first remove the paths that shall be overwritten before adding the new coercion, thus preventing a predictible warning to happen. Of course, we would still get a warning if the collision comes originates from disjoint modules?

@gares remark seems also fully right. The message should also tell what shall be used.

herbelin:
>  I think this is a more general problem of the warning system: we have the same problem with overridden notations.

Yes, this is more general problem, but we need to look for solutions, no? And after all, what I'm proposing may also work for notations.

Zimmi48:
>Yes, this is more general problem, but we need to look for solutions, no? And after all, what I'm proposing may also work for notations.

Right I simply meant: it would be better to tackle the general problem at once and not piece by piece.

herbelin:
> it would be better to tackle the general problem at once and not piece by piece.

I agree with this objective, but I don't see how yet how we could deal with that other than `declare_object` per `declare_object`.

pretyping/classops.ml
```diff
@@ -327,6 +327,9 @@ let message_ambig env sigma l =
   str"Ambiguous paths:" ++ spc () ++
   prlist_with_sep fnl (fun ijp -> print_path env sigma ijp) l
 
+let warn_ambiguous_paths env sigma =
+  CWarnings.create ~name:"ambiguous-paths" ~category:"typechecker" (message_ambig env sigma)
```
SkySkimmer: CWarnings.create should not be in a lambda, instead do
~~~ocaml
let warn_ambiguous_paths = 
  CWarnings.create ~name:... ~category:... (fun (env,sigma,l) -> message_ambig env sigma l)
~~~
(I would also inline message_ambig)

SkySkimmer (changes requested):
I don't understand @maximedenes's request for changes. Isn't there already a warning, which uses custom code to print itself? At least with this PR people can unset it without having to use -quiet.

Zimmi48:
My dismissal of the stale request for change was probably a bit exaggerate given that in this specific case, there was also the problem about the emitting of the warning at import time. It would be good if people that know about the warning machinery (e.g. @maximedenes could help debug this).

pretyping/classops.ml
Zimmi48: Could this have anything to do with the issue when locally disabling the warning?

SkySkimmer:
>there was also the problem about the emitting of the warning at import time

If I understand the discussion correctly this problem is preexisting, ie it's exactly #8207. This PR is just a cleanup.

vbgl:
Is someone going to work on this during the coming weeks? If not, please remove/postpone the milestone label. Thanks.

tchajed:
Postponed, if I work on this it'll only be later.

pi8027:
Hello. I think that this PR can be closed because of #9743. I have relaxed the ambiguous path condition by checking convertibility of coercion paths and turned ambiguous paths messages to warnings to do some output tests.

Issue #6213, opened 2017-11-21
Proposal: implement typeclass search cahce


vzaliva:
Following discussion at the coq-dev mailing list, I am opening this issue to discuss implementing a mechanism to cache typeclass search results to speed up repeated searches.

vzaliva:
from @mattam82  "AFAIR Jason Gross had a proposal in this direction some time ago. In general it is hard because proof search is context-dependent, a query in one context might have a different answer in another. But for an already run (typeclass) resolution in the same context and with the same hints yes, a cache could be kept. I think it would not be too hard to implement on top of the current implementation."

from @JasonGross  "Here's another proposal: in my experience, most of the time spent in typeclass resolution is spent on subtrees that eventually fail.  The cache can simply record which typeclasses goals eventually fail for successful typeclass resolutions, without recording their descendents.  Typeclass resolution using the cache should trust the cache about failures, but if the entire resolution fails, it should try again without the cache.  Since typeclass resolution is fully backtracking, this cannot change the success/failure behavior of type class resolution, though of course the cacheless search can get much slower under your nose without you noticing, so I'd want a setting that disables using the cache."

vzaliva:
I agree with @JasonGross that most of the time following dead branches takes most of the time. But this is just a subjective observation. We can try to analyze some logs to quantify it and see what would be the most optimal way to purge the search tree: positive (guiding towards some branches) or negative (rejecting some branches early).

The positive strategy has an advantage that it could be expressed in terms of current hints mechanism. The cache will become just additional hint database activated during particular search. Not sure if it is easy to reject branches via Hints...

vzaliva:
I plan to try to work on this feature in my spare time. I am new to Coq internals so, I will appreciate some guidance. Few initial questions I have:

1. How the mechanism will be exposed to the user? Should it be a global setting which user just activates (e.g. `Set Typeclass Caching`), or perhaps it should be a command which wraps particular tactics application? (e.g. `cached setoid_rewrite`)
1. How the implementation should be organized (code-wise)? Should just extend existing typeclass search mechanism? One idea is to have an internal API for plugging external search/cache implementations and this to be the first one to use this. There might be possible to create others.
1. How the cache is be stored? We will have the cache for multiple queries. Perhaps a single cache file per .v?
1. Should we try to use some existing mechanism to implement cache application (e.g. temporary transient hints databases which are activated if cache has chosen to be valid)?
[...]

PR #721, opened 2017-05-31
eauto based on the typeclasses eauto implementation

Zimmi48:
This PR plugs the `eauto` tactic to a new implementation based on the one of `typeclasses eauto` in `class_tactics.ml`.

~~The issue preventing to replace `eauto` is the search-depth semantics. I fail to see how to imitate the current search-depth semantics of `eauto` (threading the decrease of the depth among sub-goals) within the tactic engine (see #5555).~~
EDIT: (almost) addressed

I have also tried to see if I could replace `auto` but I encountered at least two issues: obligation solving becoming more powerful; immediate hints failing when they would succeed with the original `auto`.

Similarly, I tried to see if I could replace `trivial` but it was too powerful, due to some discrepancy with `eauto`. Here is an example showing the difference of power between `trivial` and `eauto` in 8.6:

```coq
Require Import BinInt BinNat.
Lemma foo (n m : Z) (p p0 : positive)
  (H : forall p q : positive, Z.abs_N (Z.pos p ÷ Z.pos q) = (N.pos p / N.pos q)%N) :
  Z.abs_N (Z.pos p ÷ Z.pos p0) = (Z.abs_N (Z.pos p) / Z.abs_N (Z.pos p0))%N.
Proof.
  debug trivial with nocore. (* assumption. (*fail*) intro. (*fail*) simple apply H. (*fail*) *)
  info_eauto 1 with nocore. (* simple apply H. *)
Qed.
```

~~FTR 107f835 is a cherry-pick of e6c41b9 which was merged into v8.6 (this branch should be rebased once v8.6 is merged into trunk).~~
EDIT: Done.

~~9ab3d55 is an independent (small) refactoring which could be moved out of this PR.~~
EDIT: I dropped this one because if conflicts with #638.

~~Also, @mattam82 do you agree with 083ee07? I am betting that there was a missing `if only_classes` test there. If that's correct, it could be made into an independent bug-fix. Although I haven't identified a bug that it could trigger (but I haven't looked, maybe one could cook such a bug).~~
EDIT: See #6841.

PR #8212, opened 2018-08-03
Typeclass cache

vzaliva:
**Kind:** performance

Initial implementation of typeclass caching per #6213.

### DIFF
--- a/dev/doc/typeclass_caching.md
+++ b/dev/doc/typeclass_caching.md
@@ -52,16 +52,6 @@ Coq-8.9 (master):
  | min_goals=4    | 3:34 |  5.4% |           1155 |
 ```
 
-# Documentation #
-
-Added vernacular commands:
-
-* `Set Typeclasses Caching`
-* `Unset Typeclasses Caching`
-* `Test Typeclasses Caching`
-* `Set Typeclasses Caching Mingoals n`
-* `Test Typeclasses Caching Mingoals`
-
 # TODO #
 * Can try to optimize further the performance of `tc_cache_compare` function
 * Try to figure out why switching from 8.7 to 8.9 increased number of cache hits
--- a/dev/doc/typeclass_caching.md
+++ b/dev/doc/typeclass_caching.md
@@ -2,33 +2,27 @@
 
 # Status #
 
-Caching works and on our tests (compiling math-classes) shows 27%
+Caching works and in our tests (compiling math-classes) shows 27%
 speed up.
 
 # Discussion #
 
-We are caching failures of typeclass resolution. The cache is
-basically a set and we check for the membership. It is invalidated by
-changes in hints databases or other variables affecting typeclass
-resolution mechanims.
+We are caching failures of typeclass resolution. The cache is a set
+and we check for the membership. It is invalidated by changes in hints
+databases or other variables affecting typeclass resolution mechanism.
 
-Comparing goals is computationally heavy. Due to the nature of
-comparison we could not only define equality predicate on goals, not
-orderding. Additionally it is not possible to define a hash function
-to use a hash set.
-
-We need to strike a ballance between the cost of cache lookups and
-number of cache hits. Both addition and lookup of the new goals to
-cache are heavy operation (`O(n)`). Our initial naive implementation
-("Strict match") gave us 2.49% hit rate with max cache size ~24000
-entries.
+Comparing goals is computationally heavy, so we need to strike a
+balance between the cost of cache lookups and the number of cache
+hits. Both addition and lookup of the new goals to the cache are heavy
+operations (`O(n)`). Our initial naive implementation ("Strict match")
+gave us 2.49% hit rate with max cache size ~24000 entries.
 
 Implementing more intelligent match up to unresolved evars ("Evar
 match") increased the hit ratio to 14% and max cache size become more
 manageable: 3995.
 
 Our final observation was that due to the cost of cache lookup it does
-not make sense to check it for goals which are the leafs in the proof
+not make sense to check it for goals which are the leaves in the proof
 search tree. We introduced a `min_goals` parameter which controls how
 many dependent goals a goal must have to be included in caching. This
 slightly decreased cache size and marginally improved performance.
@@ -50,7 +44,7 @@ Coq-8.7:
 Switching to Coq-8.8 master branch significantly improved cache
 performance:
 
-Coq-8.8 (master):
+Coq-8.9 (master):
 ```
  | Experiment     | Time | Hits  | Max cache size |
  | -------------- | ---- | -- -- | -------------- |
@@ -68,17 +62,10 @@ Added vernacular commands:
 * `Set Typeclasses Caching Mingoals n`
 * `Test Typeclasses Caching Mingoals`
 
-# Links #
-
-* Experimental implementation for Coq-8.7
-  https://github.com/vzaliva/coq/tree/typeclass-cache
-* Ticket https://github.com/coq/coq/issues/6213
-
-
 # TODO #
-* Can try to optimize further performance of `tc_cache_compare` function
-* Try to figure out why switching from 8.7 to 8.8 increased number of cache hits
+* Can try to optimize further the performance of `tc_cache_compare` function
+* Try to figure out why switching from 8.7 to 8.9 increased number of cache hits
 * Persistently save and load the cache
-* Do not cache failures due to hint cut.
-* Debug command to examine the cache?
+* Do not cache failures due to hint cut
+* Debug command to examine the cache contents?
 * Add test cases for caching to testsuite/output
--- a/dev/doc/typeclass_caching.md
+++ b/dev/doc/typeclass_caching.md
@@ -1,0 +1,84 @@
+# Development notes for typeclass resolution caching #
+
+# Status #
+
+Caching works and on our tests (compiling math-classes) shows 27%
+speed up.
+
+# Discussion #
+
+We are caching failures of typeclass resolution. The cache is
+basically a set and we check for the membership. It is invalidated by
+changes in hints databases or other variables affecting typeclass
+resolution mechanims.
+
+Comparing goals is computationally heavy. Due to the nature of
+comparison we could not only define equality predicate on goals, not
+orderding. Additionally it is not possible to define a hash function
+to use a hash set.
+
+We need to strike a ballance between the cost of cache lookups and
+number of cache hits. Both addition and lookup of the new goals to
+cache are heavy operation (`O(n)`). Our initial naive implementation
+("Strict match") gave us 2.49% hit rate with max cache size ~24000
+entries.
+
+Implementing more intelligent match up to unresolved evars ("Evar
+match") increased the hit ratio to 14% and max cache size become more
+manageable: 3995.
+
+Our final observation was that due to the cost of cache lookup it does
+not make sense to check it for goals which are the leafs in the proof
+search tree. We introduced a `min_goals` parameter which controls how
+many dependent goals a goal must have to be included in caching. This
+slightly decreased cache size and marginally improved performance.
+
+## Benchmarks: ##
+
+Coq-8.7:
+```
+ | Experiment     | Time | Hits  | Max cache size |
+ | -------------- | ---- | -- -- | -------------- |
+ | Baseline       | 3:40 |     0 |              0 |
+ | Strict match   | 7:43 | 2.49% |         ~24000 |
+ | Evar match     | 5:41 |   14% |           3955 |
+ | min_goals=3    | 4:46 |       |              ? |
+ | min_goals=4    | 4:37 | 4.09% |            875 |
+ | min_goals=10   | 4:47 | 1.62% |            211 |
+```
+
+Switching to Coq-8.8 master branch significantly improved cache
+performance:
+
+Coq-8.8 (master):
+```
+ | Experiment     | Time | Hits  | Max cache size |
+ | -------------- | ---- | -- -- | -------------- |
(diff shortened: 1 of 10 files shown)


dev/doc/typeclass_caching.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Discussion #
+
+We are caching failures of typeclass resolution. The cache is
+basically a set and we check for the membership. It is invalidated by
+changes in hints databases or other variables affecting typeclass
+resolution mechanims.
+
+Comparing goals is computationally heavy. Due to the nature of
+comparison we could not only define equality predicate on goals, not
```
Zimmi48: Unclear sentence.

dev/doc/typeclass_caching.md
```diff
@@ -0,0 +1,84 @@
...
+orderding. Additionally it is not possible to define a hash function
+to use a hash set.
+
+We need to strike a ballance between the cost of cache lookups and
```
Zimmi48: The right spelling is balance.

dev/doc/typeclass_caching.md
```diff
@@ -0,0 +1,84 @@
...
+ | Evar match     | 5:41 |   14% |           3955 |
+ | min_goals=3    | 4:46 |       |              ? |
+ | min_goals=4    | 4:37 | 4.09% |            875 |
+ | min_goals=10   | 4:47 | 1.62% |            211 |
+```
+
+Switching to Coq-8.8 master branch significantly improved cache
+performance:
+
+Coq-8.8 (master):
```
Zimmi48: master is actually called Coq 8.9 currently

dev/doc/typeclass_caching.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Documentation #
+
+Added vernacular commands:
+
+* `Set Typeclasses Caching`
+* `Unset Typeclasses Caching`
+* `Test Typeclasses Caching`
+* `Set Typeclasses Caching Mingoals n`
+* `Test Typeclasses Caching Mingoals`
```
Zimmi48: This should move to the user documentation (reference manual), as part of the options in the type class chapter. The directive is `.. opt::` (no need to list `Set` / `Unset` / `Test` commands).

dev/doc/typeclass_caching.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Links #
+
+* Experimental implementation for Coq-8.7
+  https://github.com/vzaliva/coq/tree/typeclass-cache
```
Zimmi48: You can probably remove this

tactics/class_tactics.ml
```diff
@@ -37,7 +37,7 @@ module NamedDecl = Context.Named.Declaration
 
 (** Options handling *)
 
-let typeclasses_debug = ref 0
+let typeclasses_debug = ref 1
```
Zimmi48: What is this change?

tactics/class_tactics.ml
```diff
@@ -78,6 +78,16 @@ let get_typeclasses_verbose () =
 let set_typeclasses_depth d = (:=) typeclasses_depth d
 let get_typeclasses_depth () = !typeclasses_depth
 
+let typeclasses_caching = ref true
+let set_typeclasses_caching d = (:=) typeclasses_caching d
+let get_typeclasses_caching () = !typeclasses_caching
+
+(** [typeclasses_caching_min_subgoals] controls how much subgoals had to be checked before failed goal be cached. It is performance optimzation to avoid caching 'easy' goals, where overhad of the cache overcomes it's potential benefits. *)
```
Zimmi48: Split comments onto separate lines (avoid lines longer than 80 characters).
Typos:
- how much -> how many
- had to -> have to
- be cached -> is cached
- overhad -> overhead
- it's potential benefits -> its potential benefits

tactics/class_tactics.ml
```diff
@@ -588,6 +614,49 @@ module Search = struct
...
+      tc_cache_goal_concl : Constr.constr ;
+
+      tc_cache_info: autoinfo;
+      tc_cache_global_hints: hint_db list }
+
+  let tc_cache_entry_cmp a b =
+    let (>>==) f c = if f != 0 then f else Lazy.force c in
+    (* We define our own version of `CErrror.noncritical` to
+          exclude errors like: Anomaly "Universe Coq.Classes.CMorphisms.397 undefined." *)
+    let noncritical x = if CErrors.is_anomaly x then true else CErrors.noncritical x in
```
Zimmi48: These anomalies always felt really weird but now we are masking them? Why not fix the actual problem which is that these errors should not have been classified as anomalies in the first place?

tactics/class_tactics.ml
```diff
@@ -588,6 +614,49 @@ module Search = struct
...
+      >>== lazy (List.compare Hint_db.compare
+                   a.tc_cache_global_hints b.tc_cache_global_hints)
+
+  module TypeclassCache = Set.Make(struct
+                              type t = tc_cache_entry
+                              let compare = tc_cache_entry_cmp
+                            end)
+
+  let typeclass_cache = Summary.ref ~name:"typeclass_cache"
+                          TypeclassCache.empty
```
Zimmi48: The newline seems unnecessary here.

tactics/class_tactics.ml
```diff
@@ -661,154 +730,214 @@ module Search = struct
...
-    let secvars = compute_secvars gl in
-    let poss =
-      e_possible_resolve hints info.search_hints secvars info.search_only_classes env sigma concl in
-    (* If no goal depends on the solution of this one or the
+    if !typeclasses_caching &&
+         TypeclassCache.exists
+           (fun x -> tc_cache_entry_cmp
+                       {
+                         tc_cache_goal_sigma  = sigma ;
+                         tc_cache_goal_concl   = EConstr.Unsafe.to_constr concl  ;
```
Zimmi48: What is justifying to use `Econstr.Unsafe.to_constr` here? Can't the type of `tc_cache_goal_concl` be changed instead to store `Econstr.t`?

tactics/hints.mli
```diff
@@ -162,6 +162,7 @@ module Hint_db :
     val cut : t -> hints_path
 
     val unfolds : t -> Id.Set.t * Cset.t
+    val compare: t -> t -> int
```
Zimmi48: This should come with a comment. And since this function was not exposed before, wouldn't it be preferable to use a ADT as return type?

Zimmi48 (commented):
Please avoid the whitespace only changes as they are going to make my life miserable when trying to rebase #721. Furthermore, they are making the diff unnecessarily more complicated to review.

Note that this is only a very superficial review.

Zimmi48:
You will need to fix the warnings that the compilation of Coq is emitting on your branch. Use `./configure -profile devel` to avoid any such surprise in the future.

tactics/class_tactics.ml
ejgallego: Indeed, this cannot be merged as-is.

tactics/class_tactics.ml
vzaliva: I just enabled to see the debug messages as it tested in CI. I will remove it once CI passed.

tactics/class_tactics.ml
vzaliva: I am afraid I do not have enough understanding of that is causing these anomalies to fix it myself. Shall I open a ticket?

tactics/class_tactics.ml
Zimmi48: OK

tactics/class_tactics.ml
Zimmi48: Yes, and ping @mattam82 there.

tactics/hints.mli
Zimmi48: IMO the standard library of OCaml is not a good source of inspiration for how things should be done. I was thinking of a return type of the form `Lt | Eq | Gt`. Of course, I'm also OK with keeping things as they are, as long as the comment explains how hints are compared by this function.

doc/sphinx/addendum/type-classes.rst
```diff
@@ -525,6 +525,22 @@ Options
...
    backtrack on it, assuming that it is canonical.
 
+.. opt:: Typeclasses Caching
+
+   This option (on by default) controls whenever the typeclass
+   resolution mechanism should use a cache to speed-up repeated
+   lookups.
+
+
+.. opt:: Set Typeclasses Caching Mingoals @num
```
Zimmi48: Remove "Set"

doc/sphinx/addendum/type-classes.rst
```diff
@@ -525,6 +525,22 @@ Options
...
+
+   This option which controls how many dependent goals a failed goal
```
Zimmi48: Remove "which"

doc/sphinx/addendum/type-classes.rst
```diff
@@ -525,6 +525,22 @@ Options
...
+   must have to be included in the cache. The value ``0`` will cause
+   caching all failed resolutions. Increasing the value of this the
```
Zimmi48: Remove "the"

Zimmi48:
- Equations fails with an anomaly.
- Iris fails with an out-of-memory (cache too big?)
- VST fails with a unification failure (could it be typeclass search that has become weaker?)

vzaliva:
Let me investigate further why the `Equations`, `Iris`, and `VST` compilation fails.

ppedrot (commented):
Unluckily, I think that this cache mechanism is completely unsound. First, there is no guarantee that unification is stable by the provided equivalence of states. Due to bugs, in general it is not, e.g. when the result depends on some fixed name. But even without taking these bugs into consideration, I don't know what is enforced by the unification algorithm. Furthermore, Extern hints can perform arbitrary computations and observe goals in many ways which break the equivalence used by the cache. So I think that rather than a cache, this should be dubbed a variant hint algorithm, which is never guaranteed to produce the same results as the original one.

tactics/class_tactics.ml
```diff
@@ -661,154 +732,214 @@ module Search = struct
...
-        (pr_depth info.search_depth ++ str": looking for " ++
-           Printer.pr_econstr_env (Goal.env gl) sigma concl ++
-           (if backtrack then str" with backtracking"
-            else str" without backtracking"));
-    let secvars = compute_secvars gl in
-    let poss =
-      e_possible_resolve hints info.search_hints secvars info.search_only_classes env sigma concl in
-    (* If no goal depends on the solution of this one or the
+    if !typeclasses_caching &&
+         TypeclassCache.exists
```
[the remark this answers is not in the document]
vzaliva: I do not know why I used `exists` instead of `mem` :) Anyway, switching to `mem` sped up things a bit and now with cache, the compilation of my benchmark it is 3:02 (compared to 4:55 without)!

tactics/class_tactics.ml
```diff
@@ -588,6 +617,48 @@ module Search = struct
...
+
+      tc_cache_info: autoinfo;
+      tc_cache_global_hints: hint_db list }
+
+  let tc_cache_entry_cmp a b =
+    let (>>==) f c = if f != 0 then f else Lazy.force c in
+    (* We define our own version of `CErrror.noncritical` to
+          exclude errors like: Anomaly "Universe Coq.Classes.CMorphisms.397 undefined." *)
+    let noncritical x = if CErrors.is_anomaly x then true else CErrors.noncritical x in
+    let def = Int.compare (Obj.magic (Obj.repr a)) (Obj.magic (Obj.repr b)) in
```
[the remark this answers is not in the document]
vzaliva: I need to establish some sort of ordering, so `==` is not enough. Any suggestions on how I can do this without `Obj.magic` are very welcome.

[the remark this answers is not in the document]
ppedrot: But actually, this is incorrect. The GC can move pointers around, which invalidates your order and breaks the invariants of the map structure, leading to garbage...

vzaliva:
@Zimmi48 Latest CI build succeeded. I've fixed an issue where incorrect hint comparison function was used. It looks like it does not try to compile VST, Iris, and Equations which failed before. How can I run a test build of these packages? (of course, I can try to do it manually, but I was wondering if there is an easy automated way).

gares:
I looked at the code and tried it out a little.

I was suspecting some logical soundness issue, but I was just mislead by some names that I believe are not well chosen (eg cache_global_hints seems to also includes instances that come from implication, eg local instances). Also the code and the debug messages seem to talk about a cache (that is typically positive) while the implemented cache is a negative one.

The problem @ppedrot points to is real, it will hit you. The GC of OCaml is a moving one (imagine a fancy mark and sweep) and hence you cannot assume the value of a pointer to be stable. You should hash the data in order to obtain a stable number out of it, and base your order relation on it. 

The use of the evar map you do is very odd, you store them in the summary for example and you compare terms belonging to two sometimes related sometimes not evar maps. At the current state of the code of Coq, I've the impression your code works, but I would suggest changing the approach, drop the evar map from the cache, and put in there constr obtained by using the (safe) to_constr coercion. It may be slower in some cases, but the code would be less fragile, IMO. Also, the summary survives the entire file. If you open, then close, then reopen sections, the very same term `Var "x"` can me completely different things, and I've the impression that the first order comparison you do does not check the `Environment.env`s your terms lives in are related. 

I think the PR, even at its current state, should provide a few test suite files. IMHO the tests should be in output/ and be accompained with the trace turning on the right debugging flags. It seems the only reasonable way to test that the cache kicks in when it should.

SkySkimmer:
This PR is dead. Reopen it once there is progress.

PR #8216, opened 2018-08-04
Fix docs on arguments to setoid_replace

### LINKED ISSUE #8213: Documentation for setoid_replace has the order of arguments wrong
#### Version
8.8

#### Operating system
n/a

The documentation has the following:
```
setoid_replace term with term [in ident] [using relation term] [by tactic]
```
resulting in keyword order `with, in, using, by`, whereas [the implementation] is as follows:
```coq
Tactic Notation "setoid_replace" constr(x) "with" constr(y)
  "using" "relation" constr(rel)
  "in" hyp(id)
  "by" tactic3(t) :=
  setoidreplacein (rel x y) id ltac:(t).
```
resulting in keyword order `with, using, in, by`. Looks like the fix is to make the switch [here](https://github.com/coq/coq/blob/15331729aaab16678c2f7e29dd391f72df53d76e/doc/refman/Setoid.tex#L541-L545):
```latex
\tacindex{setoid\_replace}
\texttt{setoid\_replace} \textit{term} \texttt{with} \textit{term}
~\zeroone{\texttt{in} \textit{ident}}
~\zeroone{\texttt{using relation} \textit{term}}
~\zeroone{\texttt{by} \textit{tactic}}
```
cc @mattam82

langston-barrett:
Fixes #8213

**Kind:** documentation

### DIFF
--- a/doc/sphinx/addendum/generalized-rewriting.rst
+++ b/doc/sphinx/addendum/generalized-rewriting.rst
@@ -544,7 +544,7 @@ pass additional arguments such as ``using relation``.
 .. tacv:: setoid_rewrite {? @orientation} @term {? at @occs} {? in @ident}
    :name: setoid_rewrite
 
-.. tacv:: setoid_replace @term with @term {? in @ident} {? using relation @term} {? by @tactic}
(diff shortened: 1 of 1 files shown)


Zimmi48 (approved):
Thanks!

PR #8217, opened 2018-08-04
Fixes #8215: "critical" type inference bug in interpreting evars by name

### LINKED ISSUE #8215: A "proof" of False by wrong context of existential variable
I find a "proof" of False that has no goals left, although error is detected by Qed.

#### Version

compiled on Jul 3 2018 10:30:00 with OCaml 4.06.1

#### Operating system
Windows 10

```
Goal False.
Proof.
  pose (n := 1).
  evar (m:nat).
  Show Existentials.
  (* Existential 1 = ?m : [n := 1 : nat |- nat] (shelved) *)
  subst n.
  Fail pose (i := ?m).
  (* The command has indeed failed with message:
     Cannot interpret ?m in current context: no binding for n. *)
  pose (n := 0).
  (* Now n is bound with another value. *)
  pose (i := ?m).
  (* It succeeds... But introducing a new variable that only
     shares the name should be irrelevant. *)
  assert (i = m).
  {
    unfold i. unfold m. reflexivity.
  }
  assert (m = 1).
  {
    subst m.
    instantiate (1 := n).
    reflexivity.
  }
  (* ?m is replaced with different n's. *)
  subst n m i.
  assert (0 <> 1) by discriminate.
  tauto.
  (* No more subgoals.
     But the proof is obviously wrong. *)
  Show Proof.
Qed. (* It fails here. *)
```

herbelin:
When interpreting an existential variable "?n@{inst}" in the current context, check that variables bound to local definitions are replaced by variables with convertible body.

Also give a message explaining the type error or non-convertibility error rather than wrongly saying that there is no binding for the variable.

**Kind:** bug fix

Fixes / closes #8215

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -81,6 +81,11 @@ Specification language
   variables) may occasionally change type inference in incompatible
   ways, especially regarding the inference of the return clause of "match".
 
+- Fixing a missing check in interpreting instances of existential
+  variables which are bound to local definitions might exceptionally
+  induce an overhead if the cost of checking the conversion of the
+  corresponding definitions is additionally high (PR #8215).
+
 Standard Library
 
 - Added `Ascii.eqb` and `String.eqb` and the `=?` notation for them,
--- a/pretyping/pretyping.ml
+++ b/pretyping/pretyping.ml
@@ -960,20 +960,50 @@ let rec pretype k0 resolve_tc (tycon : type_constraint) (env : GlobEnv.t) evdref
 and pretype_instance k0 resolve_tc env evdref loc hyps evk update =
   let f decl (subst,update) =
     let id = NamedDecl.get_id decl in
+    let b = Option.map (replace_vars subst) (NamedDecl.get_value decl) in
     let t = replace_vars subst (NamedDecl.get_type decl) in
+    let check_body id c =
+      match b, c with
+      | Some b, Some c ->
+         if not (is_conv !!env !evdref b c) then
+           user_err ?loc  (str "Cannot interpret " ++
+             pr_existential_key !evdref evk ++
+             strbrk " in current context: binding for " ++ Id.print id ++
+             strbrk " is not convertible to its expected definition (cannot unify " ++
+             quote (print_constr_env !!env !evdref b) ++
+             strbrk " and " ++
+             quote (print_constr_env !!env !evdref c) ++
+             str ").")
+      | Some b, None ->
+           user_err ?loc  (str "Cannot interpret " ++
+             pr_existential_key !evdref evk ++
+             strbrk " in current context: " ++ Id.print id ++
+             strbrk " should be bound to a local definition.")
+      | None, _ -> () in
+    let check_type id t' =
+      if not (is_conv !!env !evdref t t') then
+        user_err ?loc  (str "Cannot interpret " ++
+          pr_existential_key !evdref evk ++
+          strbrk " in current context: binding for " ++ Id.print id ++
+          strbrk " is not well-typed.") in
     let c, update =
       try
         let c = List.assoc id update in
         let c = pretype k0 resolve_tc (mk_tycon t) env evdref c in
+        check_body id (Some c.uj_val);
         c.uj_val, List.remove_assoc id update
       with Not_found ->
       try
-        let (n,_,t') = lookup_rel_id id (rel_context !!env) in
-        if is_conv !!env !evdref t (lift n t') then mkRel n, update else raise Not_found
+        let (n,b',t') = lookup_rel_id id (rel_context !!env) in
+        check_type id (lift n t');
+        check_body id (Option.map (lift n) b');
+        mkRel n, update
       with Not_found ->
       try
-        let t' = !!env |> lookup_named id |> NamedDecl.get_type in
-        if is_conv !!env !evdref t t' then mkVar id, update else raise Not_found
+        let decl = lookup_named id !!env in
+        check_type id (NamedDecl.get_type decl);
+        check_body id (NamedDecl.get_value decl);
+        mkVar id, update
       with Not_found ->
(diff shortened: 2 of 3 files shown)


herbelin:
> Does this need a bench?

I don't think so. This only affects the interpretation of expressions of the form `?n` or `?n@{instance}` which are presumably very rare, and moreover only in the presence of local definitions, and moreover, the added conversion check would be (potentially arbitrarily) costly only if someone explicitly change the value of a local definition, which seems to be an even rarer situation.

JasonGross:
I think the performance cost of this this deserves documentation somewhere.  (At the very least in a code comment, maybe in CHANGES, maybe elsewhere.)

Zimmi48:
Seems to have not progressed by lack of handling from @mattam82.
@herbelin Note that @JasonGross suggested an entry in CHANGES.
In any case, it seems that this is going to be delayed to 8.9.

herbelin:
@Jasongross: my feeling is that the cases where it can be sensitive for performance are extremely rare and of the same nature as any kind of typing situation sensitive to conversion (but I also know that I easily underestimate how far some users exploit Coq in its corners).

Would you like e.g. a comment like:

"Fixing a missing check in interpreting instances of existential variables which are bound to local definitions might exceptionally induce an overhead if the cost of checking the conversion of the corresponding definitions is additionally high (#8215)."

@Zimmi48: what about soliciting @ppedrot as reviewer and assignee?

Zimmi48:
>@Zimmi48: what about soliciting @ppedrot as reviewer and assignee?

Sure.

herbelin:
@ppedrot: could you supervise this? (I still can add a comment to CHANGES if @jasongross agrees on my formulation.)

herbelin:
I finally added anyway a comment to `CHANGES`. @ppedrot, this is a relatively important fix and I'm assigning it to you, hoping that you agree.

ppedrot (approved):
As mentioned, potentially quite expensive but there is no other way around I can think of, and rare enough that it shouldn't be a problem hopefully.

ppedrot:
Will merge promptly if no answer from @mattam82 but I'd still be curious to see a bench on a development using named evars (is there any?).

herbelin:
> I'd still be curious to see a bench on a development using named evars (is there any?).

Would indeed maybe quicker to build one by hand than finding one.

ppedrot:
CI failures are unrelated, so I'll merge anyways.

silene:
There is something I don't understand. Why does Coq allow for local definitions to be cleared while they are still in use by existential variables? I suppose that preventing clearing might be annoying in some use cases. But in that case, would it not be better if Coq was to clear local definitions from existential variables? I cannot imagine a single use case where it makes sense to first clear a local definition, then recreate one with the exact same value (modulo conversion), and finally use it to instantiate an existential variable. Would clearing local definitions become too costly for Coq?

herbelin:
@silene:
> Why does Coq allow for local definitions to be cleared while they are still in use by existential variables?

It could indeed be done differently and have #8215 behave like this:
```coq
Goal False.
Proof.
  pose (n := 1).
  evar (m:nat).
  subst n.
  Set Printing Existential Instances.
  (* "m := ?m : nat" rather than "m := ?m@{n:=1} : nat" as before *)
```
This would not change the need for the PR though. Your proposal is about modifying the behavior of `subst`.

herbelin:
The strategy is different between `clear` and `subst`:
```coq
Goal let x:=1 in exists z, z =0.
intros.
eexists ?[z].
Set Printing Existential Instances.
(* ?z@{x:=x} = 0 *)
subst x.
(* ?z@{x:=1} = 0 *)
Undo.
clear x.
(* ?z = 0 *)
Undo.
unfold x; clear x.
(* ?z@{x:=1} = 0 *)
```
> I suppose that preventing clearing might be annoying in some use cases.

In any cases, `clear` is never forbidden. We instead restrict the evars of the goal so that they do not depend anymore on the cleared variable. E.g., if `?e` is defined in context `a,b,c:nat` and we have an occurrence `?e@{a:=x,b:=y,c:=x}` in a goal with context `x,y:nat`, then a `clear x` will not fail but rather restrict `?e` so that it depends only on `b`.

herbelin:
> Would clearing local definitions [in the definition context of the evar] become too costly for Coq?

I don't think so. We could probably have `subst` behaving like `clear`, if we want.

herbelin:
> > The strategy is different between `clear` and `subst`
>
> How so? As your example shows, we end up with `?z@{x:=1} = 0` in both cases.

No, with `clear x`, it gives `?z = 0`.

The second `clear x` gives `?z@{x:=1} = 0` but it is the `unfold x` which gives `?z@{x:=1} = 0`, and, then `clear` has no `x` variable anymore to clear in `?z` (i.e. the `x` on the l.h.s. of `:=` is the `x` in the context of `?z`, not the `x` of the current goal which is the `x` to which `clear x` applies to.

I showed the `unfold x; clear x` example by comparison because this is what `subst` does under the hood.

So, I'm interpreting your proposal as changing `subst` so that `x:=x` is removed rather than replaced by `x:=1`. (Maybe `unfold` on a local variable could be changed too.)

But again, in any case, this is independent of this PR. This PR fixes a bug, independently of what tactics do.

herbelin:
For the sake of completeness wrt your other questions:

> Is there any realistic situation where it is more useful than ending up with ?z = 0?

Probably no.

> Is it because another subgoal might still have `x` in its context and we don't want to prevent the script to instantiate `?z` using `x` in that other subgoal?

No, instantiating an evar by `x` is independent of the question of whether `x` exists in the goal which refers to the evar. I can solve by reflexivity a goal `|- ?z@{x:=1} = 1` by setting `?z := x` even though `x` does not occur in this goal. What matters is the `x` in the defining context of the evar.

PR #8228, opened 2018-08-09
(Partially) Revert "Make Environ.globals abstract."

ejgallego:
This reverts commit 3984f3c1db51f7b788ad49eafb7647774e8d1f53 by adding back a view to access globals. This is important information needed in a few use cases.

Such changes broke clients wanting to inspect the enviroment, see
https://github.com/coq/coq/pull/7745#issuecomment-411597610

### DIFF
--- a/kernel/environ.mli
+++ b/kernel/environ.mli
@@ -46,8 +46,18 @@ type constant_key = Opaqueproof.opaque constant_body * (link_info ref * key)
 
 type mind_key = mutual_inductive_body * link_info ref
 
-type globals
-(** globals = constants + projections + inductive types + modules + module-types *)
+module Globals : sig
+  type t
+
+  type view =
+    { constants : constant_key Cmap_env.t
+    ; inductives : mind_key Mindmap_env.t
+    ; modules : module_body MPmap.t
+    ; modtypes : module_type_body MPmap.t
+    }
+
+  val view : t -> view
+end
 
 type stratification = {
   env_universes : UGraph.t;
@@ -67,7 +77,7 @@ type rel_context_val = private {
 }
 
 type env = private {
-  env_globals       : globals;
+  env_globals       : Globals.t;
   env_named_context : named_context_val; (* section variables *)
   env_rel_context   : rel_context_val;
   env_nb_rel        : int;
--- a/kernel/environ.ml
+++ b/kernel/environ.ml
@@ -50,12 +50,19 @@ type constant_key = Opaqueproof.opaque constant_body * (link_info ref * key)
 
 type mind_key = mutual_inductive_body * link_info ref
 
-type globals = {
-  env_constants : constant_key Cmap_env.t;
-  env_inductives : mind_key Mindmap_env.t;
-  env_modules : module_body MPmap.t;
-  env_modtypes : module_type_body MPmap.t;
-}
+module Globals = struct
+
+  type view =
+    { constants : constant_key Cmap_env.t
+    ; inductives : mind_key Mindmap_env.t
+    ; modules : module_body MPmap.t
+    ; modtypes : module_type_body MPmap.t
+    }
+
+  type t = view
+
+  let view x = x
+end
 
 type stratification = {
   env_universes : UGraph.t;
@@ -88,7 +95,7 @@ type rel_context_val = {
 }
 
 type env = {
-  env_globals       : globals;
+  env_globals       : Globals.t;
   env_named_context : named_context_val; (* section variables *)
   env_rel_context   : rel_context_val;
   env_nb_rel        : int;
@@ -110,11 +117,12 @@ let empty_rel_context_val = {
 }
 
 let empty_env = {
-  env_globals = {
-    env_constants = Cmap_env.empty;
-    env_inductives = Mindmap_env.empty;
-    env_modules = MPmap.empty;
-    env_modtypes = MPmap.empty};
+  env_globals =
+    { Globals.constants = Cmap_env.empty
+    ; inductives = Mindmap_env.empty
+    ; modules = MPmap.empty
+    ; modtypes = MPmap.empty
+    };
   env_named_context = empty_named_context_val;
   env_rel_context = empty_rel_context_val;
   env_nb_rel = 0;
@@ -215,29 +223,29 @@ let lookup_named_ctxt id ctxt =
   fst (Id.Map.find id ctxt.env_named_map)
 
 let fold_constants f env acc =
-  Cmap_env.fold (fun c (body,_) acc -> f c body acc) env.env_globals.env_constants acc
+  Cmap_env.fold (fun c (body,_) acc -> f c body acc) env.env_globals.Globals.constants acc
 
 let fold_inductives f env acc =
-  Mindmap_env.fold (fun c (body,_) acc -> f c body acc) env.env_globals.env_inductives acc
+  Mindmap_env.fold (fun c (body,_) acc -> f c body acc) env.env_globals.Globals.inductives acc
 
 (* Global constants *)
 
 let lookup_constant_key kn env =
-  Cmap_env.find kn env.env_globals.env_constants
+  Cmap_env.find kn env.env_globals.Globals.constants
 
 let lookup_constant kn env =
-  fst (Cmap_env.find kn env.env_globals.env_constants)
+  fst (Cmap_env.find kn env.env_globals.Globals.constants)
 
 (* Mutual Inductives *)
 let lookup_mind kn env =
-  fst (Mindmap_env.find kn env.env_globals.env_inductives)
+  fst (Mindmap_env.find kn env.env_globals.Globals.inductives)
 
 let mind_context env mind =
   let mib = lookup_mind mind env in
   Declareops.inductive_polymorphic_context mib
 
 let lookup_mind_key kn env =
-  Mindmap_env.find kn env.env_globals.env_inductives
+  Mindmap_env.find kn env.env_globals.Globals.inductives
 
 let oracle env = env.env_typing_flags.conv_oracle
 let set_oracle env o =
@@ -468,10 +476,10 @@ let no_link_info = NotLinked
 
 let add_constant_key kn cb linkinfo env =
   let new_constants =
-    Cmap_env.add kn (cb,(ref linkinfo, ref None)) env.env_globals.env_constants in
+    Cmap_env.add kn (cb,(ref linkinfo, ref None)) env.env_globals.Globals.constants in
   let new_globals =
     { env.env_globals with
-	env_constants = new_constants } in
+        Globals.constants = new_constants } in
   { env with env_globals = new_globals }
 
 let add_constant kn cb env =
@@ -598,10 +606,10 @@ let template_polymorphic_pind (ind,u) env =
   else template_polymorphic_ind ind env
   
 let add_mind_key kn (_mind, _ as mind_key) env =
-  let new_inds = Mindmap_env.add kn mind_key env.env_globals.env_inductives in
+  let new_inds = Mindmap_env.add kn mind_key env.env_globals.Globals.inductives in
   let new_globals =
     { env.env_globals with
-        env_inductives = new_inds; } in
+        Globals.inductives = new_inds; } in
   { env with env_globals = new_globals }
 
 let add_mind kn mib env =
@@ -688,22 +696,22 @@ let keep_hyps env needed =
 
 let add_modtype mtb env =
   let mp = mtb.mod_mp in
-  let new_modtypes = MPmap.add mp mtb env.env_globals.env_modtypes in
-  let new_globals = { env.env_globals with env_modtypes = new_modtypes } in
+  let new_modtypes = MPmap.add mp mtb env.env_globals.Globals.modtypes in
+  let new_globals = { env.env_globals with Globals.modtypes = new_modtypes } in
   { env with env_globals = new_globals }
 
 let shallow_add_module mb env =
   let mp = mb.mod_mp in
-  let new_mods = MPmap.add mp mb env.env_globals.env_modules in
-  let new_globals = { env.env_globals with env_modules = new_mods } in
+  let new_mods = MPmap.add mp mb env.env_globals.Globals.modules in
+  let new_globals = { env.env_globals with Globals.modules = new_mods } in
   { env with env_globals = new_globals }
 
 let lookup_module mp env =
-    MPmap.find mp env.env_globals.env_modules
+    MPmap.find mp env.env_globals.Globals.modules
 
 
-let lookup_modtype mp env = 
-  MPmap.find mp env.env_globals.env_modtypes
+let lookup_modtype mp env =
+  MPmap.find mp env.env_globals.Globals.modtypes
 
 (*s Judgments. *)
 


ppedrot (changes requested):
I am slightly tired to always repeat the arguments against transparency of datatypes, so I'll refer to the arguments given in #7745 for a global recap. In your particular case, the fact that you rely on the effective implementation of the type seems to always arise from the same need: serialization. I remember you had the same problem with serialization of formatting data.

Obviously there is a tension between the need to export data representation for encoding purposes and the advantages provided by opaqueness from the point of view of robustness to change and static correction. But I don't think we should make the type transparent for all clients, in particular in both introduction and elimination directions. In order to please everybody, I think the best here is to provide an alias `globals_view` type to the opaque `globals` type and a function that exposes the implementation, well-documented to NOT BE USED by casual users.

Also, you probably shouldn't pattern-match on this type, it is more of a blob than anything else. Pattern-matching only makes sense for free datatypes, not complex quotients. So the exhaustive warning is quite not meaningful.

ejgallego:
Even if you have repeated the "arguments" often, they don't seem less bogus to me. People need to use the APIs in several ways and the only thing this "transparency" thing achieves is the introduction of noise and hacks and a loss of uniformity.

> In your particular case, the fact that you rely on the effective implementation of the type seems to always arise from the same need: serialization. I remember you had the same problem with serialization of formatting data.

It is not just serialization, debug, analysis, learning... The data in this patch is critical data and any reasonable user of the environment will have to inspect it.

> I'll refer to the arguments given in #7745

The main argument there was "do you have an objective motation" and the answer was ["no"](https://github.com/coq/coq/pull/7745#issuecomment-396586275)

> I remember you had the same problem with serialization of formatting data.

"Formatting data" is the printing format of Coq [something that will become closer to HTML with] and obviously downstream clients have to use it. We implemented the "wrapper" and so far it has only brought noise and 0 advantages, as the next major revision of `Pp.t` won't be compatible for good reason.

> advantages provided by opaqueness from the point of view of robustness to change and static correction.

This is a fallacy, at least in this case. What are the advantages? None. Benefit? My software was broken and now I have to lose my very little time in this useless discussion instead of actually coding. You are slightly tired, I am **extremely tired** of people that don't use the API I use messing with them. Maybe as the user of them I know what should work well.

Note how good this approach is that you have to add a comment [not even using the proper docstring] to explain what `globals` do contain, thing that was not needed before as the things was self-documenting.

> I think the best here is to provide an alias globals_view type to the opaque globals type and a function that exposes the implementation, well-documented to NOT BE USED by casual users.

First, this is pure noise; second, there is no reason to hide what's on the env from any user. The type is private so it is already read-only and can be safely used by plugins.

> So the exhaustive warning is quite not meaningful.

Pattern matching on records is totally fine and OCaml warning 9 has proven essential in the maintenance of large codebases.

Anyways, I'll stop here as the damage/benefit ratio seems overwhelming; I am not even gonna get started on people defending "opaqueness" then being OK with Coq relying on `Marshal` in a fundamental way. I'll tell you a little secret, if we ever want to have something that resembles efficiently parallel document processing we better stop relying on `Marshal`.

This change broke third party software such as SerAPI as thus it should reverted. I didn't add SerAPI to the CI purely as a deference to the developers as it is a fairly complex piece of software and I can take care of it on my own, but if that's the only way to proceed then I will.

SkySkimmer (commented):
TBH I erred in favor of abstractness for this type, maybe it was a mistake. So I'm fine with this PR.

gares:
How large is the code generated by ppx in serapi (in binary size)?

I'm a bit tired of these discussions. If the type is transparent just to enable (better) serialization, I believe we should put the serializer in the module and keep the type opaque. Adding to the mli a function `t -> xml` or similar is not going to be controversial. And yes, depend on ppx? (since we don't want to write the serializer by hand).

ejgallego:
This is not just about serialization, as discussed in my comment; even documentation is affected.

mattam82:
I think @ppedrot gave a convincing argument that opacity would be necessary to implement Include the right way for example, and if memory serves right, this type transiently included a projection map also, so it does change from time to time while providing a stable lookup_* interface. I’m tempted to side with @ppedrot here and provide and explicitly unstable/tied to the implementation view here, or better yet, as Enrico suggests make the serialization part of the module. I think we should maintain the comment though explaining that it contains inductives, constants modules and module types which are accessible with the corresponding lookup functions. It’s a very low-level representation not maintaining the order of declarations in the environment. I wonder how you can make use of it besides serialization

gares:
Emilio could you answer the question about the size? Maybe it is a silly metric but it is the only problem I could see in putting a serialiser there.

gares:
For all other uses I think you can live with an opaque data type and an API on it. Maybe the API lacks functions, if so we can add them.

I really don't see how these discussions end up generating so much heat, frankly.

ejgallego:
> I think @ppedrot gave a convincing argument that opacity would be necessary to implement Include the right way for example, and if memory serves right, this type transiently included a projection map also, so it does change from time to time while providing a stable lookup_* interface.

Umm, it seems to me that API stability and "opacity" are really different notions; you can expose a record for documentation / code cleanliness issues, but promise nothing about its stability.

AFAICT Coq has 0 API stability guarantees anyway so I don't see how the 8.8 status-quo would prevent anything.

> How large is the code generated by ppx in serapi (in binary size)?
> 
> I'm a bit tired of these discussions. If the type is transparent just to enable (better) serialization, I believe we should put the serializer in the module and keep the type opaque. Adding to the mli a function t -> xml or similar is not going to be controversial. And yes, depend on ppx (since we don't want to write the serializer by hand).

I have no idea about the code size, but IMHO we rather keep serialization out of this file for now.

> For all other uses I think you can live with an opaque data type and an API on it. Maybe the API lacks functions, if so we can add them.

I'm afraid that this amounts to duplicating the whole map interface, so it doesn't seem to me like a sensible solution. For example, I may add diff capability to `Map.t`, then, we would have to redo the whole API here? That doesn't seem right.

Indeed, given the coding style in this file IMHO that's a clear call to expose the projections again.

gares:
> For example, I may add diff capability to Map.t, then, we would have to redo the whole API here? That doesn't seem right.

What seems natural to me is to have an API in `Environ.mli` that looks like:
```
type env_diff = whatever
val diff : env -> env -> env_diff
```
You seem to want to code this in another file, or better in a plugin/toplevel. It that correct?

> I have no idea about the code size, but IMHO we rather keep serialization out of this file for now.

Please tell us "why" and why "for now". This sentence contains very little information to me.

ejgallego:
> You seem to want to code this in another file, or better in a plugin/toplevel. It that correct?

That's orthogonal IMHO, I am not sure what I would like to see there yet, it was just an example of "custom API" vs "standard API".

Being able to poke / print the globals and their maps was very useful as Gaëtan has pointed out.

> Please tell us "why" and why "for now". This sentence contains very little information to me.

Well, first why is because this is not about serialization, that will remain to be Marshall for a while. We don't even have a design for a new serialization system so it is hard to know what format, constraints, etc... we should have. Apart from that Coq cannot depend on external libs and without that doing serialization is just silly.

ejgallego:
This may not show on @maximedenes review list due to aggressive filtering, pinging.

gares:
I was looking at the rest of the API. `env` is a private record and there is no other function letting one change the value of the `globals` field. If it is a read-only field then I think this PR is fine (no invariant can be broken). If there are plans to change the data type, a comment should suffice to warn the API user.

maximedenes (approved):
I don't really understand what opacity of the type brings us, so I'm ok with the patch. That being said, I don't care much about it.

ppedrot:
So far the invariant was that we only used an environment for lookup. Now that we expose everything, it'll make client depend on implementation details like order of folding over maps, representation of indirections, etc., etc.

I am quite pissed off that a middle ground cannot be reached. I understand the architectural limitations of having serialization outside of the module that defines the relevant structures, but I'm amazed that people are not forecasting the issues of exposing internal APIs. Like, it is not as if one of the people implied in this thread were not [recently precisely complaining about technical shortcomings of this design choice](https://inbox.ocaml.org/caml-list/7fed9f2d-953b-8153-0acf-97a7c984f440@lexifi.com/T/#m2fdff43b3873398f1a900e4c31faab6f857c578b).

SkySkimmer:
Maybe Environ should be really restricted to single lookups and if you want to fold through globals you have to use the module structures in the safe_env? Which then has a specified order and so on.

gares:
Let's try to cut it:
- @ejgallego if it is blocking for you, eg serapi can't compile, please add to this PR the CI job for it
- anybody in favor of opacity: please add to this PR an extra commit re-opacifying as much as possible the data type and eventually adding the needed API so that CI still goes trough

If we stay on the abstract level, discussion will never converge. Let's try to tackle the issue that made @ejgallego open this PR in the first place.

ejgallego:
@skyskimmer that seems like a pointless reimplementation of a very standard ocaml API.

@gares , serapi can work either way (if opaque then it will just show `opaque`) my point is that given the rest of the file the choice seems clear.

gares:
This long discussion for what, then? Aesthetic? I feel I wasted my time trying to help settle this. I guess you'll have to continue without me. Good luck.

ejgallego:
Oh I would not say this is aesthetic but an API issue. The API proposed of course allows quite a bit more functionality, but he'll won't freeze over if we make this moduke unusable.

ejgallego:
Well, I think we have reached a point where we are bike-shedding over this [IMO trivial and pretty clear given the style in the rest of the file] PR.

So in the interest of all of us let's reach a quick conclusion: the maintainer doesn't want to commit but he approves, the developer who introduced the change does agree, so it boils down I think to see if @ppedrot  reconsiders. So please @ppedrot give your final word here so we can merge or close.

If this PR is not merged the effect is that clients won't be able to know which globals does the environment have. Omitting this functionality from the API seems like an obnoxious omission to me, but effects won't be the end of the world.
  
Tools that can dump the env [such as SerAPI 0.6] will just output an `Opaque` term for that field; this will also affect other features, for example the parallel process queue I have been working on won't be able to perform environment sanity checks, etc... which indeed seems discouraging.

ppedrot:
I stand my ground: you should provide a `globals_view` type in the mli that is just internally `globals` and export an identity function `globals_view -> globals`. I am not asking that much, and this makes everybody happy, so why not do it?

maximedenes:
@ejgallego do you plan to provide the view @ppedrot mentioned in the API?

ejgallego:
> @ejgallego do you plan to provide the view @ppedrot mentioned in the API?

No. I don't think it makes sense, why for this record and not for the rest of the records in such file?

I am tired of wasting my time in this stuff.

maximedenes:
> why for this record and not for the rest of the records in such file

Well, I guess we'd have to start somewhere, right? If you want to do it for all records, I don't think @ppedrot will object.

ejgallego:
> Well, I guess we'd have to start somewhere, right? If you want to do it for all records, I don't think @ppedrot will object.

I don't want to do so; first, because that "view" thing is just terrible coding style without any other benefit than to annoy people, second, because that would make all my ppx machinery useless.

maximedenes:
> because that "view" thing is just terrible coding style

You don't need to convince me, I was the first one to oppose this change. But after all, it does not change much.

> that would make all my ppx machinery useless.

I've never used `ppx`. Is there no way to customize it, to make it go through the view when you need to serialize the original type?

SkySkimmer:
Why a view function instead of something like 
~~~ocaml
type globals_internal = (*the record*)
type globals = private globals_internal
~~~
?

ppedrot:
That works for me as well, as the type-checker will catch casts statically.

ejgallego:
> type globals_internal = (*the record*) type globals = private globals_internal

@SkySkimmer why is that different from:
```
type globals = private { ... (* the record *) }
```
I didn't do that for consistency with the coding style. The main type is private tho so you don't really gain anything other than to annoy clients.

ppedrot:
This has not the same semantics. Private records prevent you from building them and only allows projections, while private types prevent casting from the base types but allows explicit casting to it using `(_ :> _)` syntax.

ppedrot:
> you don't really gain anything other than to annoy clients

Also, that's the point. If you are messing with that type you're definitely doing either something very peculiar or outright wrong.

ejgallego:
> Is there no way to customize it, to make it go through the view when you need to serialize the original type?

Oh sure, obviously there is a way to do anything you want. Maybe we could instead modify the OCaml compiler so it understands the next syntax / pattern.

ejgallego:
> If you are messing with that type you're definitely doing either something very peculiar or outright wrong.

Oh sure, querying which globals are in scope seems like a very exotic need....

Maybe indeed the case is that we don't really want to look what gets added there, specially in the case of workers.

ejgallego:
Anyways, I suggest we just stop beating this dead horse folks.

ppedrot:
You might believe the opposite, but I am really sympathetic of your serialization goals, and it seems to me that the solution proposed by @SkySkimmer should work (as long as ppx serialization understands private types, which it should, right?).

ejgallego:
> it seems to me that the solution proposed by @SkySkimmer should work

Can you explain what _concrete_ bad behavior of a client does the "two types" solution prevent when compared with the single-record one?

ppedrot:
A bad behaviour is observing a projection without realizing that in general, you shouldn't. With a private record, the compiler allows you to project the fields making the abstraction layer moot, while with a private proxy type, you have to cast it explicitly before projecting. If you don't do it you get a type error.

ejgallego:
> A bad behaviour is observing a projection without realizing that in general, you shouldn't. With a private record, the compiler allows you to project the fields making the abstraction layer moot, while with a private proxy type, you have to cast it explicitly before projecting. If you don't do it you get a type error.

I am not asking about "abstract" reasoning, I know enough OCaml I think to understand what the different at the level of static semantics is. Let state my question again: can you explain what is the concrete misuse of the API the two types solution would prevent?

p.s: There is no reason a client should not be able to observe a projection. Let me restate, this PR doesn't imply that the interface is stable [now that API docs are built again we could estate expected interface stability] If you are messing with that kind of low-level types, of course you should not expect it to remain stable between versions. But guess that, in the real world many clients [such as for example rich top-levels] need to hook to compiler internals in order to work; c'est la vie.

maximedenes:
> Oh sure, obviously there is a way to do anything you want. Maybe we could instead modify the OCaml compiler so it understands the new syntax / programming pattern.

I doubt this ironic tone was required. I'll leave the discussion.

ejgallego:
> I doubt this ironic tone was required.

My answer is not ironic at all.

ejgallego:
> can you explain what is the concrete misuse of the API the two types solution would prevent?

@ppedrot ping

ejgallego:
Sooner of later we are going to need this so the mistake must be corrected.

mattam82:
@ejgallego IIUC you don't want to try the proxy/private type solution? Can you point to code in SerAPI dealing with this problem?

ejgallego:
> @ejgallego IIUC you don't want to try the proxy/private type solution? Can you point to code in SerAPI dealing with this problem?

Private type is OK; however I don't see the point as `environ` is already private so why to forbid people using the record? There is no risk of doing silly stuff, and we get nuisance IMVHO, specially when it was already this way in 8.8 and nobody was really bothered by, Gaëtan indeed confirmed that the change was not really on purpose.

But indeed I can make it private, work for env serialization is in the `v8.9+maps`, but it is not working as it was pending on this.

The thing is that if the type is not private you can just do `[@@deriving sexp]` and be done with it, otherwise, you lose that ability, so you need to introduce a wrapper.

Anyways I will workaround the thing my own way using a type copy and `Obj.magic`. There is still a lot of "platform" work to be done IMHO as to be able to delegate proofs more efficiently.

gares:
> Anyways I will workaround the thing my own way using a type copy and `Obj.magic`. There is still a lot of "platform" work to be done IMHO as to be able to delegate proofs more efficiently.

Why is this PR open again then? (at least, in its current shape)

ejgallego:
> Why is this PR open again then? (at least, in its current shape)

I'd rather not use `Obj.magic`, also, #8495 is still open AFAICT.

SkySkimmer:
So are we doing this or what?

ejgallego:
This is still a merge for me .

We got more use cases, for example icoq may want to use this for properly computing a file digest.

ejgallego:
Well this is my oldest PR and I'd like it to get unstuck; I think it is fair to say that objectively the blockage of this PR [as well as for example #9443] has been a loss of time for 0 gain, not to speak that this is an important API regression from 8.8.

I guess we need some kind of "Technical Committee" to decide what to do in this case, it could well be the RM or the steering committee.

gares:
We should have discussed it at the WG, why didn't you put it on the schedule?

ejgallego:
> We should have discussed it at the WG, why didn't you put it on the schedule?

I was afraid of it hahaha , but indeed, I still am.

The point is that technically I still fail to see what problems this does introduce, the API was around since 8.0 and the type is still private; so I could close and instead do in my plugins `Obj.magic` but really when you look at the big picture it seems a bit strange I'd say.

yangky11:
I would prefer to expose `globals` in the interface file. I'm building an application that interacts with Coq. It needs to access the program state including the global environment. Making globals transparent (maybe private) is the easiest way to go, and I don't see any potential misuse.

vbgl:
I’ve postponed this PR to 8.11, as the discussion did not reach any satisfying conclusion.

ejgallego:
Yet another use case that is not possible to do today without this; verify that what the UI believes is true and what the kernel thinks is true are the same.

ejgallego:
I guess there is little point in dragging this PR for more time; I've made the record private; is that OK @ppedrot ?

Not a big effect tho as no API takes globals as input.

ppedrot:
@ejgallego the requested change was adding a private proxy type, not to make the record itself private. As argued above, the latter is virtually equivalent to exporting the whole type.

gares:
> I guess there is little point in dragging this PR for more time

There is still the needs:discussion label, and in person discussion did not happen, so arguing the PR should be merged is against our procedures. Please schedule a slot at the next WG.

ejgallego:
> There is still the needs:discussion label, and in person discussion did not happen, so arguing the PR should be merged is against our procedures.

Where is this procedure specified?

maximedenes:
> Where is this procedure specified?

I don't think it has been written down formally, but the meaning of the label seemed clear so far. What do you think needs clarification, "needs"? "discussion"?

SkySkimmer:
>What do you think needs clarification, "needs"? "discussion"?

Neither of those imply in-person-ness.

maximedenes:
> Neither of those imply in-person-ness.

Ok, so maybe @Zimmi48 could add a description for the label which clarifies the meaning.

gares:
> Where is this procedure specified?

For once, the script to merge PR complains if there is a `needs: ...` label.

This label is added, as far as I can tell, when things are not converging on the PR comments, or when the proposed change has implications that needs agreement among developers.

In this particular case the PR discussion is *not* converging, so discussion in person is needed and I suggested above.

If you like I can edit the label description, or propose an edit to whatever document you think fits.

ejgallego:
Indeed the way we have handled the flag doesn't imply "in-personess" as Gaëtan points out, in many cases we have resolved the discussions online.

Honestly, I don't understand what the purpose of your remark about the label is.

It seems obvious to me than before merging and updated version both the label and the requests for changes should we addresses, as we do for all PRs.

gares:
And how do you resolve `needs: discussion`? You did not solve that, and you did not even bring up the topic when we had the opportunity. So do it next time, and in the meanwhile please stop putting pressure on the PR with sentences like "I guess there is little point in dragging this PR for more time;".

SkySkimmer:
That's an argument about the particular PR, not the label policy

Gaëtan Gilbert

ejgallego:
@ppedrot what do you think now?

@maximedenes do you want this discussed in a WG, or would an agreement here work for you to remove the label?

The API for the kernel environment is indeed a complex subject as discussed here and elsewhere [for example what MetaCoq uses, etc...]

On the other hand this fixes a serious regression that has like 5 projects using Obj.magic just to read what the kernel knows, and indeed it is puzzling we let such a regression slip for 3 releases.

ejgallego:
> And how do you resolve `needs: discussion`?

Like asking the person who put the label if he's OK to remove it?

> You did not solve that, and you did not even bring up the topic when we had the opportunity. So do it next time, and in the meanwhile please stop putting pressure on the PR with sentences like "I guess there is little point in dragging this PR for more time;".

This is offtopic, factually incorrect, and accusatory for a second time; let's move the discussion elsewhere please.

ejgallego:
> I guess there is little point in dragging this PR for more time;

Just for the record, what the sentence means is "I'll implement the solution you requested for now"

SkySkimmer:
>indeed it is puzzling we let such a regression slip for 3 releases.

I think it's pretty clear that we don't ming breaking plugins if they're not in CI

ejgallego:
> I think it's pretty clear that we don't mind breaking plugins if they're not in CI

Yeah, but this seems like very basic functionality to me; but indeed the way our review model is, kinda incentives racing when a conflict may appear.

A good example is #7745 , imagine I put a request for changes there, which I didn't just out of luck; then I would have blocked that PR for quite a long time, and I would have "won" the race, in the sense that I would be on the other side than I am now in this PR. Yup this models seems pretty negative to me as incentives having less changes / harsh reviews.

maximedenes:
> do you want this discussed in a WG, or would an agreement here work for you to remove the label?

Either would work for me, but I would like that we stop putting pressure on @ppedrot. He made his opinion clear already (many times). Either we follow his recommendation and ask if he's ok, then remove the label, or we wait until the WG to discuss more.

> I'll implement the solution you requested for now

I think what was perceived as just putting pressure once more on @ppedrot was the fact that you didn't implement what he requested (a private proxy), but something else (a private record), despite him having clearly made the distinction between the two (see earlier comments), and then asked if the PR could be merged.

I see that you now implemented the private proxy, so I guess it was just a mistake, but it probably led to this confusing discussion.

Just to recall some context, I personally approved this PR in its initial version. I'm also ok with it in its current version. So I'm certainly not hostile to this patch, just trying to make sure we can discuss in the best conditions.

Zimmi48:
> > Neither of those imply in-person-ness.
> 
> Ok, so maybe @Zimmi48 could add a description for the label which clarifies the meaning.

I'm pretty sure this label actually *had* a description of the form "The PR should be discussed at the next WG." Has someone removed it? Anyway, I am definitely not against remote discussions outside the scope of a WG, although *sometimes* when a written discussion cannot converge, it is useful to have a video-chat / call with the persons involved. It doesn't need to be in the context of a WG.

ejgallego:
> I'm pretty sure this label actually _had_ a description of the form "The PR should be discussed at the next WG." Has someone removed it?

I seem to recall indeed it had one, maybe you check the edit logs? But indeed we should document labels on `CONTRIBUTING` and consider a bug any difference between what says in CONTRIBUTING and what is written on github.

Zimmi48:
> > I'm pretty sure this label actually _had_ a description of the form "The PR should be discussed at the next WG." Has someone removed it?
> 
> I seem to recall indeed it had one, maybe you check the edit logs? But indeed we should document labels on `CONTRIBUTING` and consider a bug any difference between what says in CONTRIBUTING and what is written on github.

I don't have an edit log for the label, I can only tell you it was last updated on January 11th.

I agree in principle with the documentation of labels in the contributing guide. For now, only a few labels are documented. I did not document the "needs: discussion" one because there was never a formal discussion / agreement on what was the process for such PRs.

ejgallego:
Ping @ppedrot [I guess you could also be the assignee if you are OK with the patch]

ppedrot (approved):

ppedrot:
What's the status of the "discussion" around this PR? While it's probably overblown for the current issue, I do think we still need some technical plan to solve the diverging need between SerAPI-style interactions and static opacity checks in the sensitive areas of the codebase. Maybe some support from OCaml itself would help.

ejgallego:
> What's the status of the "discussion" around this PR?

IMHO we can merge, @maximedenes is OK and the discussion was mainly between us. I agree with the plan you proposed for now [implemented here]

> While it's probably overblown for the current issue, I do think we still need some technical plan to solve the diverging need between SerAPI-style interactions and static opacity checks in the sensitive areas of the codebase. Maybe some support from OCaml itself would help.

Indeed we got entangled over a silly discussion; for me the discussion has little to do with SerAPI, but as @mattam82 pointed out, it is more about how the kernel exposes its data. Indeed I do agree that the kernel is special enough as to expose the info in the way it wants, but such info has to be exposed. For example I need this API for search, UI, and ML projects [none of them depending on SerAPI directly].

Indeed we have quite a bit of duplication w.r.t. where is this information stored. As of today, each time we add a definition it has to be kept in sync along the libobject, nametab, and the kernel, all of that manually. I'd be great if we could somehow improve that, but IMHO that discussion should happen as we go.

ejgallego:
@vbgl IMHO this PR would be quite useful if backported to 8.10 , leaving that for your consideration tho.

kernel/environ.mli
```diff
@@ -46,8 +46,18 @@ type constant_key = Opaqueproof.opaque constant_body * (link_info ref * key)
...
+  type t
+
+  type view =
+    { constants : constant_key Cmap_env.t
+    ; inductives : mind_key Mindmap_env.t
+    ; modules : module_body MPmap.t
+    ; modtypes : module_type_body MPmap.t
+    }
+
+  val view : t -> view
```
SkySkimmer: Why a function instead of `type t = private view`?

ejgallego: See the discussion above.

SkySkimmer: I do not see.

ejgallego: I was referring to this https://github.com/coq/coq/pull/8228#issuecomment-514552020 , but indeed, I am likely not understanding.

SkySkimmer: Your first version was 
~~~ocaml
type globals = private { bla bla }
~~~
what is wanted is
~~~ocaml
type globals_internal = { bla bla }
type globals = private globals_internal
~~~

ejgallego: What is the difference tho? I mean, that way you also expose the internal implementation.

SkySkimmer:
The version with a view function is fine IMO, it's just not what I expected

Gaëtan Gilbert

ejgallego:
> The version with a view function is fine IMO, it's just not what I expected

Yeah, I'd say let's go this way, anyways in my impression this API is highly unstable and will likely evolve once use cases start to surface more prominently.

ejgallego:
Ping. I'm fine if you would like to discuss this in the WG tho.

gares:
WG discussion: we don't need discussion anymore.

ppedrot:
Still needs a rebase though.

ejgallego:
> Still needs a rebase though.

Does it? I see no conflict.

ppedrot:
Something miraculous happened in the meantime? I'll merge this.

ppedrot:
Should turn my tongue 7 times: the problem is that it doesn't compile after a merge on master, so you still have to rebase. To see this you'd need to rerun the CI.

ejgallego:
Thanks @ppedrot , fixed; indeed I had assumed a broken compilation to be the case for the rebase, but I was curious if you had any other reason.

PR #8230, opened 2018-08-09
fix formulation of the Euclid Theorem in comment

samuelgruetter:


theories/Numbers/Integer/Abstract/ZDivEucl.v
```diff
@@ -13,7 +13,7 @@ Require Import ZAxioms ZMulOrder ZSgnAbs NZDiv.
 (** * Euclidean Division for integers, Euclid convention
 
     We use here the "usual" formulation of the Euclid Theorem
-    [forall a b, b<>0 -> exists b q, a = b*q+r /\ 0 < r < |b| ]
+    [forall a b, b<>0 -> exists r q, a = b*q+r /\ 0 < r < |b| ]
```
jashug: Should the last clause also be changed to `0 <= r < |b|`?

samuelgruetter: fixed!

herbelin:
Indeed, thanks. Shall merge after some delay if ever someone has more comments to do.

herbelin (approved):

Issue #8045, opened 2018-07-11
Windows packaging builds are overly verbose.


ejgallego:
Windows packaging builds are overly verbose and include things such as decompression / untarring logs; NSIS debug information, etc... which makes them very hard to read, and waste quite a lot of space.

Please, provide less verbose logging by default, and if necessary, provide a debug option or upload an artifact.

MSoegtropIMC:
Well I think ignoring information you don't need is less troublesome than conjuring information you don't have out of thin air. I think rerunning jobs to get additional information is not a good solution. I think this way much more time (and resources) are wasted than by having verbose log files. But the system already has log files in an extra folder, so this could be uploaded as artifact instead.

I generally like the set -x of the main script for debugging, but I could put this into a log file as well (it currently isn't).

What would you like to see in the logs? Just module start / end messages and the last few lines of error messages?

Zimmi48:
Indeed, I would support putting a bit less in the logs that are displayed and storing the rest of the logs as build artifacts. I don't have any good suggestion on what to keep / what to remove.

ejgallego:
Example: https://gitlab.com/coq/coq/-/jobs/107883687

It is literally hard to understand what is going on with the current log format.

Zimmi48:
Indeed, it's a bit hard to read because of artifact upload but the log verbosity is otherwise a problem solved. Maybe there should be a bit of documentation (in `dev/ci/README.md`) mentioning that most of the log is to be browsed through artifacts.

ejgallego:
> most of the log is to be browsed through artifacts.

Please don't fix this issue like that. Some of us just want to click and have a single message telling us what went wrong with the build. Logs are OK for those needing the details.

I think the issue is simple, when a build goes wrong, there is only 1 relevant piece of information, the rest is just making the person looking at the logs lose their time.

MSoegtropIMC:
@ejgallego : would you consider this as fixed meanwhile? There have been some PRs integrated to improve the logging, most notably to put all more verbose logging into separate files and upload them as artifacts.

ejgallego:
IMHO we still have plenty of room to improve, so I'd like this to be open. It still takes me a lot of time to navigate the logs, and on top of that I have to open a file; compared to the setup we have on Linux + Dune for example this can get better.

It may make sense to add a flag so that the default CI build is silent, but you can turn it verbose from the web interface.

MSoegtropIMC:
[...]

PR #8243, opened 2018-08-12
Remove xargs from "make clean" so it won't fail on Cygwin

jfehrle:
**Kind:** infrastructure.

`xargs` fails on Cygwin when I run `make clean`.  This uses the `-delete` option on find instead except for one place where only `-exec rm {}` would work.

(Edited: Was originally about doing a `make clean` for the Windows build.  Most of the discussion is out of date.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ cleankeepvo: indepclean clean-ide optclean cruftclean depclean docclean devdoccl
 objclean: archclean indepclean
 
 cruftclean: ml4clean
-	find . -name '*~' -o -name '*.annot' | xargs rm -f
+	find . \( -name '*~' -o -name '*.annot' \) -exec rm -f {} +
 	rm -f gmon.out core
 
 camldevfilesclean:
@@ -208,7 +208,7 @@ camldevfilesclean:
 indepclean:
 	rm -f $(GENFILES)
 	rm -f $(COQTOPBYTE) $(CHICKENBYTE) $(TOPBYTE)
-	find . \( -name '*~' -o -name '*.cm[ioat]' -o -name '*.cmti' \) -delete
+	find . \( -name '*~' -o -name '*.cm[ioat]' -o -name '*.cmti' \) -exec rm -f {} +
 	rm -f */*.pp[iox] plugins/*/*.pp[iox]
 	rm -rf $(SOURCEDOCDIR)
 	rm -f toplevel/mltop.byteml toplevel/mltop.optml
@@ -239,7 +239,7 @@ archclean: clean-ide optclean voclean
 optclean:
 	rm -f $(COQTOPEXE) $(CHICKEN) $(TOPBINOPT)
 	rm -f $(TOOLS) $(PRIVATEBINARIES) $(CSDPCERT)
-	find . -name '*.cmx' -o -name '*.cmx[as]' -o -name '*.[soa]' -o -name '*.so' | xargs rm -f
+	find . \( -name '*.cmx' -o -name '*.cmx[as]' -o -name '*.[soa]' -o -name '*.so' \) -exec rm -f {} +
 
 clean-ide:
 	rm -f $(COQIDECMO) $(COQIDECMX) $(COQIDECMO:.cmo=.cmi) $(COQIDEBYTE) $(COQIDE)
@@ -252,10 +252,10 @@ ml4clean:
 	rm -f $(GENML4FILES) $(GENMLGFILES)
 
 depclean:
-	find . $(FIND_SKIP_DIRS) '(' -name '*.d' ')' -print | xargs rm -f
+	find . $(FIND_SKIP_DIRS) '(' -name '*.d' ')' -exec rm -f {} +
 
 cacheclean:
-	find theories plugins test-suite -name '.*.aux' -delete
+	find theories plugins test-suite -name '.*.aux' -exec rm -f {} +
 
 cleanconfig:
 	rm -f config/Makefile config/coq_config.ml myocamlbuild_config.ml dev/ocamldebug-coq dev/camlp5.dbg config/Info-*.plist
@@ -263,14 +263,18 @@ cleanconfig:
 distclean: clean cleanconfig cacheclean timingclean
 
 voclean:
-	find theories plugins test-suite \( -name '*.vo' -o -name '*.glob' -o -name "*.cmxs" -o -name "*.native" -o -name "*.cmx" -o -name "*.cmi" -o -name "*.o" \) -delete
-	find theories plugins test-suite -name .coq-native -empty -delete
+	find theories plugins test-suite \( -name '*.vo' -o -name '*.glob' -o -name "*.cmxs" \
+	-o -name "*.native" -o -name "*.cmx" -o -name "*.cmi" -o -name "*.o" \) -exec rm -f {} +
+	find theories plugins test-suite -name .coq-native -empty -exec rm -f {} +
 
 timingclean:
-	find theories plugins test-suite \( -name '*.v.timing' -o -name '*.v.before-timing' -o -name "*.v.after-timing" -o -name "*.v.timing.diff" -o -name "time-of-build.log" -o -name "time-of-build-before.log" -o -name "time-of-build-after.log" -o -name "time-of-build-pretty.log" -o -name "time-of-build-both.log" \) -delete
+	find theories plugins test-suite \( -name '*.v.timing' -o -name '*.v.before-timing' \
+	  -o -name "*.v.after-timing" -o -name "*.v.timing.diff" -o -name "time-of-build.log" \
+	  -o -name "time-of-build-before.log" -o -name "time-of-build-after.log" \
+	  -o -name "time-of-build-pretty.log" -o -name "time-of-build-both.log" \) -exec rm -f {} +
 
 devdocclean:
-	find . -name '*.dep.ps' -o -name '*.dot' | xargs rm -f
+	find . \( -name '*.dep.ps' -o -name '*.dot' \) -exec rm -f {} +
 	rm -f $(OCAMLDOCDIR)/*.log $(OCAMLDOCDIR)/*.aux $(OCAMLDOCDIR)/*.toc
 	rm -f $(OCAMLDOCDIR)/ocamldoc.sty $(OCAMLDOCDIR)/coq.tex
 	rm -f $(OCAMLDOCDIR)/html/*.html
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -126,14 +126,14 @@ clean:
 	$(SHOW) 'RM        <**/*.stamp> <**/*.vo> <**/*.vio> <**/*.log> <**/*.glob>'
 	$(HIDE)find . \( \
 	  -name '*.stamp' -o -name '*.vo' -o -name '*.vio' -o -name '*.log' -o -name '*.glob' \
-	  \) -print0 | xargs -0 rm -f
+	  \) -exec rm -f {} +
 	$(SHOW) 'RM        <**/*.cmx> <**/*.cmi> <**/*.o> <**/*.test>'
 	$(HIDE)find unit-tests \( \
 	  -name '*.cmx' -o -name '*.cmi' -o -name '*.o' -o -name '*.test' \
-	  \) -print0 | xargs -0 rm -f
+	  \) -exec rm -f {} +
 distclean: clean
 	$(SHOW) 'RM        <**/*.aux>'
-	$(HIDE)find . -name '*.aux' -print0 | xargs -0 rm -f
+	$(HIDE)find . -name '*.aux' -exec rm -f {} +
 
 #######################################################################
 # Per-subsystem targets


jfehrle:
Oops, I only meant to submit the last commit.  Will fix momentarily.

MSoegtropIMC (changes requested):
@jfehrle : I would do it in a different way: exclude the unneeded files in the tar. It doesn't make much sense to tar a lot of files, untar them and then run make clean (which in an empty folder is a substantial amount of work because first the dependencies must be built) to delete them.

jfehrle:
> I would do it in a different way: exclude the unneeded files in the tar.

That would be quite complex to code since the set of files removed by `make clean` is not simple to describe, as you can see below.  It would also present an ongoing maintenance issue: when the behavior of `make clean` changes from time to time, this .sh file would need to be updated to match it.  While there is some inefficiency in removing the files after the tar rather than filtering them out beforehand, this would not impact the continuous integration since that copies from a clean source tree.

>run make clean (which in an empty folder is a substantial amount of work because first the dependencies must be built)

Huh?  What dependencies need to be built?  I'm not aware that anything needs to be built before running `make clean`.

```
$ make clean
rm -f
rm -f ide/input_method_lexer.ml
rm -f ide/highlight.ml ide/config_lexer.ml ide/config_parser.mli ide/config_parser.ml
rm -f ide/utf8_convert.ml
rm -rf
rm -f bin/coqtop.exe  bin/coqtop.opt.exe bin/coqproofworker.opt.exe bin/coqtacticworker.o
pt.exe bin/coqqueryworker.opt.exe
rm -f bin/coqdep.exe bin/coq_makefile.exe bin/coq-tex.exe bin/coqwc.exe bin/coqdoc.exe bi
n/coqc.exe bin/coqworkmgr.exe bin/coqpp.exe bin/fake_ide.exe bin/ocamllibdep.exe bin/coqd
ep_boot.exe plugins/micromega/csdpcert.exe
find . -name '*.cmx' -o -name '*.cmx[as]' -o -name '*.[soa]' -o -name '*.so' | xargs rm -
f
find theories plugins test-suite \( -name '*.vo' -o -name '*.glob' -o -name "*.cmxs" -o -
name "*.native" -o -name "*.cmx" -o -name "*.cmi" -o -name "*.o" \) -delete
find theories plugins test-suite -name .coq-native -empty -delete
rm -rf _build
rm -f test-suite/misc/universes/all_stdlib.*
rm -f coqpp/coqpp_lex.ml ide/config_lexer.ml ide/coq_lex.ml ide/protocol/xml_lexer.ml ide
/utf8_convert.ml tools/coqdep_lexer.ml tools/coqdoc/cpretty.ml tools/coqwc.ml tools/ocaml
libdep.ml coqpp/coqpp_parse.ml parsing/g_constr.ml parsing/g_prim.ml plugins/btauto/g_bta
uto.ml plugins/cc/g_congruence.ml plugins/ltac/coretactics.ml plugins/ltac/g_eqdecide.ml
plugins/ltac/g_tactic.ml plugins/micromega/g_micromega.ml plugins/nsatz/g_nsatz.ml plugin
s/omega/g_omega.ml plugins/quote/g_quote.ml plugins/romega/g_romega.ml plugins/rtauto/g_r
tauto.ml toplevel/g_toplevel.ml vernac/g_proofs.ml vernac/g_vernac.ml kernel/copcodes.ml
 kernel/byterun/coq_jumptbl.h
rm -f bin/coqtop.byte.exe  bin/coqtop.byte.exe bin/coqproofworker.byte.exe bin/coqtacticw
orker.byte.exe bin/coqqueryworker.byte.exe
find . \( -name '*~' -o -name '*.cm[ioat]' -o -name '*.cmti' \) -delete
rm -f */*.pp[iox] plugins/*/*.pp[iox]
rm -rf
rm -f toplevel/mltop.byteml toplevel/mltop.optml
rm -f glob.dump
rm -f config/revision.ml revision
rm -f plugins/micromega/.micromega.ml.generated
make -C test-suite clean
make[1]: Entering directory '/cygdrive/c/bin/cygwin_coqB/build/coq-local/test-suite'
Makefile:77: cannot run complexity tests (no bogomips found)
rm -f trace .lia.cache output/MExtraction.out
RM        <**/*.stamp> <**/*.vo> <**/*.vio> <**/*.log> <**/*.glob>
RM        <**/*.cmx> <**/*.cmi> <**/*.o> <**/*.test>
make[1]: Leaving directory '/cygdrive/c/bin/cygwin_coqB/build/coq-local/test-suite'
rm -f ide/coqide_main.ml parsing/cLexer.ml plugins/derive/g_derive.ml plugins/extraction/
g_extraction.ml plugins/firstorder/g_ground.ml plugins/funind/g_indfun.ml plugins/ltac/ex
traargs.ml plugins/ltac/extratactics.ml plugins/ltac/g_auto.ml plugins/ltac/g_class.ml pl
ugins/ltac/g_ltac.ml plugins/ltac/g_obligations.ml plugins/ltac/g_rewrite.ml plugins/ltac
/profile_ltac_tactics.ml plugins/setoid_ring/g_newring.ml plugins/ssr/ssrparser.ml plugin
s/ssr/ssrvernac.ml plugins/ssrmatching/g_ssrmatching.ml parsing/g_constr.ml parsing/g_pri
m.ml plugins/btauto/g_btauto.ml plugins/cc/g_congruence.ml plugins/ltac/coretactics.ml pl
ugins/ltac/g_eqdecide.ml plugins/ltac/g_tactic.ml plugins/micromega/g_micromega.ml plugin
s/nsatz/g_nsatz.ml plugins/omega/g_omega.ml plugins/quote/g_quote.ml plugins/romega/g_rom
ega.ml plugins/rtauto/g_rtauto.ml toplevel/g_toplevel.ml vernac/g_proofs.ml vernac/g_vern
ac.ml
find . -name '*~' -o -name '*.annot' | xargs rm -f
rm -f gmon.out core
find . '(' -name '{arch}' -o -name '.svn' -o -name '_darcs' -o -name '.git' -o -name '.bz
r' -o -name 'debian' -o -name "${GIT_DIR}" -o -name '_build' -o -name '_build_ci' -o -nam
e '_install_ci' -o -name 'user-contrib' -o -name 'test-suite' -o -name '.opamcache' -o -n
ame '.coq-native' ')' -prune -o '(' -name '*.d' ')' -print | xargs rm -f
rm -f doc/*/*.dvi doc/*/*.aux doc/*/*.log doc/*/*.bbl doc/*/*.blg doc/*/*.toc \
doc/*/*.idx doc/*/*~ doc/*/*.ilg doc/*/*.ind doc/*/*.dvi.gz doc/*/*.ps.gz doc/*/*.pdf.gz\

doc/*/*.???idx doc/*/*.???ind doc/*/*.v.tex doc/*/*.atoc doc/*/*.lof\
doc/*/*.hatoc doc/*/*.haux doc/*/*.hcomind doc/*/*.herrind doc/*/*.hidx doc/*/*.hind \
doc/*/*.htacind doc/*/*.htoc doc/*/*.v.html
rm -f doc/stdlib/index-list.html doc/stdlib/index-body.html \
  doc/stdlib/*Library.coqdoc.tex doc/stdlib/library.files \
  doc/stdlib/library.files.ls doc/stdlib/FullLibrary.tex
rm -f doc/*/*.ps doc/*/*.pdf doc/*/*.eps doc/*/*.pdf_t doc/*/*.eps_t
rm -rf doc/stdlib/html doc/tutorial/tutorial.v.html
rm -f doc/common/version.tex
rm -f doc/coq.tex
rm -rf doc/sphinx/_build
find . -name '*.dep.ps' -o -name '*.dot' | xargs rm -f
rm -f /*.log /*.aux /*.toc
rm -f /ocamldoc.sty /coq.tex
rm -f /html/*.html
rm -f .merlin ide/.merlin plugins/.merlin META.coq
```

MSoegtropIMC:
Did you try a make clean in a fresh directory? I would expect that it is a substantial amount of work, because I think it first computes all dependencies. In my experience with other projects (especially coqmake generated projects) a make clean in a fresh folder takes substantially longer than a make clean after a build.

Anyway, this problem will go away soon. I am about to release the developer build scripts.

SkySkimmer:
The `clean` targets do not cause dependency building actually.

ppedrot (commented):
No comment.

MSoegtropIMC:
One more note on this: make clean AFTER a build used to crash on cygwin because of some command line length restriction of xargs. I don't think this has ever been fixed, because it is a Windows only problem and a make clean is not really required. The solution would be to replace find | xargs rm combinations with individual calls to rm - which is of cause slower but doesn't crash.

If make clean doesn't compute dependencies and make clean does actually work, the PR is fine.

SkySkimmer:
We use `find -delete` in some of the clean targets, maybe we should just harmonize on that.

jfehrle:
Yeah, the make clean is failing in both the windows32 and windows64 jobs.  On my own system, it succeeds in one version of Cygwin and fails in another.  (It succeeds when I run the .bat file.)  I haven't tried to track down why and what version.  On the other hand, if the command is `make clean || true` then the failure is unsightly but not really an issue.

I won't get back to this PR for some days or weeks; at the moment I'm focusing on finishing some things I want to get into 8.9.  

You mentioned you'll have some new developer scripts soon--when do you think they'll be ready?  I'd be interested in reviewing them and trying them even before they're pushed to master.

jfehrle:
I will probably take a look at find -delete but not right away.

MSoegtropIMC:
@jfehrle : can you please attach a file list (find . -type f) of your clean directory and after a build (I don't do Windows in place builds, so it would be difficult for me to get this). I can see if I can come up with a simple tar exclude file. Please either attach the zippd file to this PR.

jfehrle:
@SkySkimmer I can use `-delete` except in one place because `FIND_SKIP_DIRS` has a `-prune` inside, which is incompatible with `-delete`.  Any ideas on other ways to rewrite this?

Also, I don't see why I get the error when the list of files only has 11 entries--how could that be too big??

Makefile:
```
	find . $(FIND_SKIP_DIRS) '(' -name '*.d' ')' -print | xargs rm -f
```

```
XARGS
find . '(' -name '{arch}' -o -name '.svn' -o -name '_darcs' -o -name '.git' -o -name '.bzr' -o -name 'debian' -o -name "${GIT_DIR}" -o -name '_build' -o -name '_build_ci' -o -name '_install_ci' -o -name 'user-contrib' -o -name 'test-suite' -o -name '.opamcache' -o -name '.coq-native' ')' -prune -o '(' -name '*.d' ')' -print | xargs -0 rm -f
assertion "bc_ctl.arg_max >= LINE_MAX" failed: file "/usr/src/findutils-4.6.0-1.x86_64/src/findutils-4.6.0/xargs/xargs.c", line 500, function: main


find . '(' -name '{arch}' -o -name '.svn' -o -name '_darcs' -o -name '.git' -o -name '.bzr' -o -name 'debian' -o -name "${GIT_DIR}" -o -name '_build' -o -name '_build_ci' -o -name '_install_ci' -o -name 'user-contrib' -o -name 'test-suite' -o -name '.opamcache' -o -name '.coq-native' ')' -prune -o '(' -name '*.d' ')' -delete
find: The -delete action automatically turns on -depth, but -prune does nothing when -depth is in effect.  If you want to carry on anyway, just explicitly use the -depth option.
make: *** [Makefile:241: depclean] Error 1
```
```
LIST FILES
find . '(' -name '{arch}' -o -name '.svn' -o -name '_darcs' -o -name '.git' -o -name '.bzr' -o -name 'debian' -o -name "${GIT_DIR}" -o -name '_build' -o -name '_build_ci' -o -name '_install_ci' -o -name 'user-contrib' -o -name 'test-suite' -o -name '.opamcache' -o -name '.coq-native' ')' -prune -o '(' -name '*.d' ')' -print
./.mlfiles.d
./.mllibfiles.d
./.vfiles.d
./checker/.mlfiles.d
./checker/.mllibfiles.d
./kernel/byterun/coq_fix_code.c.d
./kernel/byterun/coq_interp.c.d
./kernel/byterun/coq_memory.c.d
./kernel/byterun/coq_values.c.d
./plugins/.mlfiles.d
./plugins/.mllibfiles.d
$ xargs --show-limits < /dev/null
Your environment variables take up 5497 bytes
POSIX upper limit on argument length (this system): 24455
POSIX smallest allowable upper limit on argument length (all systems): 4096
Maximum length of command we could actually use: 18958
Size of command buffer we are actually using: 24455                    <= bc_ctl.arg_max
Maximum parallelism (--max-procs must be no greater): 2147483647
```

jfehrle:
@MSoegtropIMC And here is the before file: [before.txt](https://github.com/coq/coq/files/2321139/before.txt)

SkySkimmer:
>@SkySkimmer I can use -delete except in one place because FIND_SKIP_DIRS has a -prune inside, which is incompatible with -delete. Any ideas on other ways to rewrite this?

Use `-exec` then.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -1437,7 +1441,8 @@ function make_addon_equations {
 }
 
 function make_addons {
-  if [ -n "$GITLAB_CI" ]; then
+  echo GITLAB_CI = ${GITLAB_CI:=foobar}
+  if [ "${GITLAB_CI}" != "foobar" ]; then
```
SkySkimmer: Why this change?

dev/build/windows/makecoq_mingw.sh
```diff
@@ -1164,6 +1164,10 @@ function make_coq {
         COQ_BUILD_PATH=/build/coq-local
         tar -zcf $TARBALLS/coq-local.tar.gz --exclude-vcs -C "${COQ_VERSION%/*}" "${COQ_VERSION##*/}"
```
SkySkimmer: Looking at the `tar` manpage there is
~~~
       --exclude-vcs-ignores
              Exclude files that match patterns read  from  VCS-specific  ignore  files.   Supported
              files are: .cvsignore, .gitignore, .bzrignore, and .hgignore.
~~~
Does this do what you want?

MSoegtropIMC:
@jfehrle : thanks for supplying the files!

I analyzed the delta and exclude filtering using these commands:

````
FILTER='\.exe$|\.d$|\.cm[oixas]+$|\.[ao]$|\.vo$|\.glob$|/g_[^/]+\.ml$|.merlin$|^\./doc/sphinx/_build|^\./doc/sphinx/_static.*\.png$'
comm -23 <(sort after.txt) <(sort before.txt) > newfiles.txt
egrep -v -e $FILTER newfiles.txt > newfilesred.txt
egrep -v -e $FILTER before.txt > beforered.txt
comm -23 <(sort before.txt) <(sort beforered.txt) > beforeremoved.txt
````

The comm 23 thing is my favorite way to compute set differences in bash. This shows that the files can be mostly filtered out by extension or path. I attached the output files "newfilesred.txt" and "beforeremoved.txt". I would think the files in beforeremoved are not there in a clean repo. The newfilesred mostly contains generated ML files and a few obscurities. There seems to be a name pattern g_<something>.ml for generated ML files, which I filtered out, but this pattern is not always used. Maybe we can use this name pattern for all generated .ml files and ignore these files for the time beeing or explicitly exclude them (individually be name).

Anyhow this doesn't look like something terrible complicated to do with a tar exclude file.

[beforeremoved.txt](https://github.com/coq/coq/files/2321332/beforeremoved.txt)
[newfilesred.txt](https://github.com/coq/coq/files/2321333/newfilesred.txt)

**Comment by MSoegtropIMC** [comment]
One more note on make clean: in my opinion a make clean should remove exactly those files which can be produced by the make file and not do wildcard removes. This typically does require to compute dependencies because finding out what can be build requires to find all the intermediate files and this requires the dependencies.

**Comment by MSoegtropIMC** [comment]
@SkySkimmer : the limits only apply until (including) Windows XP / Server 2003.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms682653(v=vs.85).aspx

**BUT**: some cygwin programs - as far as I can remember a few years back this included xargs - do their own checks of what they think is reasonable and throw an error in case they think they might exceed the no longer existing Windows limits. To get around this, I tried to keep the environment small, mostly by reducing the PATH and also removing the OLD_PATH environment variable cygwin creates on start. Almost all programs are fine though. I had only issues with very selected programs.

In case we have serious issues with this, one could check the sources and supply a patch to cygwin. XP is no longer supported since April 2014, so XP specific hacks can be removed now.

https://www.microsoft.com/en-us/windowsforbusiness/end-of-xp-support

**Comment by MSoegtropIMC** [comment]
I tested the maximum command line length with CreateProcessA. Results:

When calling cmd.exe, the CreateProcessA call succeeds up to a command line length of 32766. It fails from 32767 on. cmd.exe itself fails with command lines longer than 8191.

I didn't check how this behaves when the Unicode variant is called.

The C code (Compiled / tested with Visual Studio 2015) is attached.
[Caller.cpp.txt](https://github.com/coq/coq/files/2323008/Caller.cpp.txt)

**Comment by MSoegtropIMC** [comment]
@SkySkimmer suggested to use --exclude-vcs-ignores (the comment was on code and deleted with a commit --amend). I think this is the best option. It should do exactly what we want.

Of cause it makes sense to fix make clean anyway, but I wouldn't call it after the tar copy and use --exclude-vcs-ignores when taring instead.

**Comment by jfehrle** [comment]
@MSoegtropIMC I'm abandoning adding a `make clean` as I originally planned--other priorities call plus I figured out a workable local alternative.  The real underlying problem I have compiling CoqIDE is that I can't compile it with Cygwin.  The Cygwin tools I've tried won't link with `FLEXDLL`.  Perhaps I'll switch my Coq development to minGW sometime.

What remains here is converting `make clean` to not rely on xargs, allowing me to do a `make clean` in Cygwin without an error.  I also took out the `v` from the untar--it gives thousands of lines of uninteresting output.  @ejgallego mentioned this in https://github.com/coq/coq/issues/8045.  Are you OK with that?

@SkySkimmer Would you review the xargs change?

I'll update the PR description momentarily.

**Review by SkySkimmer — COMMENTED** [review]
I think I would prefer to use the same solution for everything if possible (so -exec).
Any reason not to?

**Comment by jfehrle** [comment]
> I would use `-exec +` over `-exec ;` for this

Didn't know about that one.  Seems like that might that be vulnerable to the same command line length issue as `xargs`.  But in this case, the number of `*.d` files is certainly small.

Pushed this change.

File: dev/build/windows/makecoq_mingw.sh [hunk 3]
```diff
@@ -267,7 +267,7 @@ function get_expand_source_tar {
       return 1
     fi
   else
-    logn untar tar xvaf "$TARBALLS/$name.$3" --strip $strip
+    logn untar tar xaf "$TARBALLS/$name.$3" --strip $strip
   fi
```
MSoegtropIMC: I put in the v for a reason. I like to see what files are expanded. Sometimes strange things happen in CI even when untaring, and then it is good to know how far the untar get. I generally don't think that long log files are a problem, as long as they are not so long that downloading the raw log file takes considerable time.

MSoegtropIMC (changes requested):

MSoegtropIMC:
@jfehrle : regading make clean: I will put the cvs ignore thing in when taring, also for efficiency reasons.

Sorry for the continued delays with the developer environment - but I first need to get my other PR in, which takes longer than expected. I didn't know that you work with opam cygwin. You read on coq-dev that opam cygwin 64 has serious propblems and opam cygwin 32 works better? Which one are you using? 32 or 64?

jfehrle:
> I put in the v for a reason.
And so it shall remain.  Please cancel your review.

Looks like github just went down.

dev/build/windows/makecoq_mingw.sh
ejgallego: They are definitively a problem as they are impossible to make a sense of, cf #8045

MSoegtropIMC (dismissed):

MSoegtropIMC:
One note: I originally assigned this to me because it originally only contained changes in files owned by me. Since now all changes in files owned by me have been removed and changes in files owned by others have been added, I wonder if someone else want's to take this.

jfehrle:
@gares Would you care to review and merge this since you're the code owner for the Makefiles?

Thanks.

jfehrle:
> I think I would prefer to use the same solution for everything if possible (so -exec).
Any reason not to?

I missed this comment earlier.  They all use -exec ... + now.

gares:
CI fails on windows

gares (approved):
I'm fine with the change, but it seems the CI failure is related

SkySkimmer:
It's failing in appveyor but not gitlab windows.
Also see https://ci.appveyor.com/project/coq/coq/build/master~6574#L3865
>/bin/sh: line 1:  3932 Segmentation fault

I don't think that's supposed to happen.

jfehrle:
> but it seems the CI failure is related

Hard to see how this connected.  I rebased on master, let's see if it still happens.

jfehrle:
@gares Does anyone else need to review this?   If not, would you merge it when convenient?  Thanks.

gares:
I'll merge when I'm back in my office.

jfehrle:
@gares Are you back in the office yet?

gares:
Oh sorry, I'm back but I've seen the PR is assigned to @MSoegtropIMC so I thought he was supposed to merge it. I'll merge it now.

Makefile
```diff
@@ -252,25 +252,29 @@ ml4clean:
...
 	rm -f config/Makefile config/coq_config.ml myocamlbuild_config.ml dev/ocamldebug-coq dev/camlp5.dbg config/Info-*.plist
 
 distclean: clean cleanconfig cacheclean timingclean
 
 voclean:
-	find theories plugins test-suite \( -name '*.vo' -o -name '*.glob' -o -name "*.cmxs" -o -name "*.native" -o -name "*.cmx" -o -name "*.cmi" -o -name "*.o" \) -delete
-	find theories plugins test-suite -name .coq-native -empty -delete
+	find theories plugins test-suite \( -name '*.vo' -o -name '*.glob' -o -name "*.cmxs" \
+	-o -name "*.native" -o -name "*.cmx" -o -name "*.cmi" -o -name "*.o" \) -exec rm -f {} +
+	find theories plugins test-suite -name .coq-native -empty -exec rm -f {} +
```
maximedenes: @jfehrle @gares why change this line? There was no call to `xargs` and the change is incorrect. The one on the line above also looks useless. Same question for `timingclean`.

maximedenes: Ok, I see why @SkySkimmer suggested to use `-exec rm` instead of `-delete`. I don't know why, but here it is not correct as done currently because `.coq-native` is a directory.

gares: oops, let's open an issue

jfehrle: Submitted PR https://github.com/coq/coq/pull/8497 to use `rm -rf`.

Issue #8245, opened 2018-08-12
creating n evars takes O(n^2.5) time; 4GB RAM for 8k


andres-erbsen:
#### Version

~~~~~
The Coq Proof Assistant, version 8.8.1 (July 2018)
compiled on Jul 5 2018 7:25:50 with OCaml 4.06.0
~~~~~

#### Operating system

Arch Linux, up to date as of yesterday. `Linux ashryn 4.17.14-arch1-1-ARCH #1 SMP PREEMPT Thu Aug 9 11:56:50 UTC 2018 x86_64 GNU/Linux`.

#### Description of the problem

Hopefully this is just another manifestation of the issues underlying #8231 and #8237, but I am opening an issue anyway just to be sure.

Creating evars is unexpectedly expensive. Creating 8000 evars takes a minute and a half and 4GB of RAM. The overall performance seems to scale worse than O(n^2) and better than O(n^3), so maybe O(n^2.5). I have not measured memory requirements for individual runs, but I suspect it is O(n^2).

~~~~~~
Goal True. Time do 250 (let H := fresh in evar (H : unit)). Abort. Optimize Heap.
(* Finished transaction in 0.066 secs (0.042u,0.023s) (successful) *)
Goal True. Time do 500 (let H := fresh in evar (H : unit)). Abort. Optimize Heap.
(* Finished transaction in 0.192 secs (0.192u,0.s) (successful) *)
Goal True. Time do 1000 (let H := fresh in evar (H : unit)). Abort. Optimize Heap.
(* Finished transaction in 0.798 secs (0.779u,0.016s) (successful) *)
Goal True. Time do 2000 (let H := fresh in evar (H : unit)). Abort. Optimize Heap.
(* Finished transaction in 3.692 secs (3.599u,0.083s) (successful) *)
Goal True. Time do 4000 (let H := fresh in evar (H : unit)). Abort. Optimize Heap.
(* Finished transaction in 18.662 secs (18.329u,0.282s) (successful) *)
Goal True. Time do 8000 (let H := fresh in evar (H : unit)). Abort.
(* Finished transaction in 97.144 secs (95.799u,1.073s) (successful) *)

(* 119.28user 1.68system 2:01.31elapsed 99%CPU (0avgtext+0avgdata 4050804maxresident)k
0inputs+0outputs (0major+1272385minor)pagefaults 0swaps *)
~~~~~~

JasonGross:
[...]

PR #8246, opened 2018-08-13
Implementing an internal basic version of the "pose" tactic independent of the multi-usage internal "letin_tac"

ppedrot:
The previous implementation was calling a lot of useless unification even though the net effect of the tactic was simply to add a binding to the environment.

Interestingly the base tactic was used in several higher level tactics, including evar and ssreflect pose.

Part of the fix for #8245.

### DIFF
--- a/plugins/ltac/evar_tactics.ml
+++ b/plugins/ltac/evar_tactics.ml
@@ -94,7 +94,7 @@ let let_evar name typ =
     in
     let (sigma, evar) = Evarutil.new_evar env sigma ~src ~naming:(Namegen.IntroFresh id) typ in
     Tacticals.New.tclTHEN (Proofview.Unsafe.tclEVARS sigma)
-    (Tactics.letin_tac None (Name.Name id) evar None Locusops.nowhere)
+    (Tactics.pose_tac (Name.Name id) evar)
   end
   
 let hget_evar n =
--- a/plugins/ltac/tacinterp.ml
+++ b/plugins/ltac/tacinterp.ml
@@ -1664,16 +1664,18 @@ and interp_atomic ist tac : unit Proofview.tactic =
         (* We try to fully-typecheck the term *)
           let flags = open_constr_use_classes_flags () in
           let (sigma,c_interp) = interp_open_constr ~flags ist env sigma c in
-          let let_tac b na c cl eqpat =
-            let id = Option.default (make IntroAnonymous) eqpat in
-            let with_eq = if b then None else Some (true,id) in
-            Tactics.letin_tac with_eq na c None cl
-          in
           let na = interp_name ist env sigma na in
+          let let_tac =
+            if b then Tactics.pose_tac na c_interp
+            else
+              let id = Option.default (make IntroAnonymous) eqpat in
+              let with_eq = Some (true, id) in
+              Tactics.letin_tac with_eq na c_interp None Locusops.nowhere
+          in
           Tacticals.New.tclWITHHOLES ev
           (name_atomic ~env
             (TacLetTac(ev,na,c_interp,clp,b,eqpat))
-            (let_tac b na c_interp clp eqpat)) sigma
+            let_tac) sigma
         else
         (* We try to keep the pattern structure as much as possible *)
           let let_pat_tac b na c cl eqpat =
--- a/plugins/ssr/ssrfwd.ml
+++ b/plugins/ssr/ssrfwd.ml
@@ -25,9 +25,7 @@ module RelDecl = Context.Rel.Declaration
 (** 8. Forward chaining tactics (pose, set, have, suffice, wlog) *)
 (** Defined identifier *)
 
-
-let settac id c = Tactics.letin_tac None (Name id) c None
-let posetac id cl = Proofview.V82.of_tactic (settac id cl Locusops.nowhere)
+let posetac id cl = Proofview.V82.of_tactic (Tactics.pose_tac (Name id) cl)
 
 let ssrposetac (id, (_, t)) gl =
   let ist, t =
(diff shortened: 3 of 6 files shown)


herbelin:
Are you sure you prefer this code duplication rather than optimizing `letin_tac` so that it bypasses unification when called with a `nowhere`? This would also optimize the `remember in |-` case if I'm correct.

ppedrot:
I think I prefer duplication so that it is clear what every tactic does. Otherwise due to combinatorial explosion, we end up having a lot of potential code paths, including a few that are never tested, and most of the arguments of the tactics are useless. Furthermore I don't really know how it would be possible to provide a more efficient implementation without essentially rewriting the few lines that this PR adds.

As for writing `remember with |-`, I think you can blame the author of this kind of code...

(FTR, we should probably also perform the same kind of changes for similarly related tactics like assert and the like.)

herbelin:
> I think I prefer duplication so that it is clear what every tactic does. Otherwise due to combinatorial explosion, we end up having a lot of potential code paths, including a few that are never tested, and most of the arguments of the tactics are useless.

An eternal debate... With more code sharing, fixes and improvements (and bugs) are shared. With more distinct paths, it is more code to maintain (but it might also be more convenient in case of different maintainers). Reading code with well-identified code paths might be easier than reading code with many internal code paths. But maybe a compromise is also possible to make code with several code paths not to convoluted.

> Furthermore I don't really know how it would be possible to provide a more efficient implementation without essentially rewriting the few lines that this PR adds.

Bypassing unification when called with a nowhere clause would not be ok?

ppedrot:
> Bypassing unification when called with a nowhere clause would not be ok?

But then you still end up calling an *usine à gaz* performing a lot of tactic machinery just to pose a let binding in the context, which should be like one of the primitive operations. The current state of this code reflects the limitations of the old engine, I'd bet.

gares:
Changes to ssr are fine.

The new code is in line with the current code of tactics, that is horrible.
The code should just refine `let x := t : _ in _` and check that `x` is free but you can't because the pretyper that accepts _ takes in input glob terms, while `t` is a constr.

Circumvent that is what makes to code horrible:
- the refine primitive does not accept `_` hence you must put an evar in there
- creating the evar in the longer context is what takes most of the code
- and you need a context, that you cannot create unless you typecheck `t` first (you could put an evar there, but I guess you wanted to be faster)
- `t` being a (open) constr is likely to be already typed, but anyway we type check it, and not to type check it 3 times we call `refine ~typecheck:false`
- refreshing universes is also fishy. If the input term was a raw/glob one, this is what the pretyper would do I guess (typical ambiguity) but here it is not clear that one wants a fresh universe. Or maybe it is just because of algebraic universes one cannot really use. A comment would be welcome.
- not to talk about the fact that, given that type checking and refreshing universes are not in the monad, one has to use unsafe primitives in order to do what one has to do.

My impression is that, at the status quo of the API, one should have two primitives along these lines:
```
val pose_open_constr : constr * types -> Id.t -> unit tactic
val pose_glob : glob_term -> Id.t -> unit tactic
val typeof : ?refresh:bool -> constr -> types tactic
```
The latter would just refine the `let-in` and check the ident is fresh.
The former would at least make it clear we are not type checking the term again, since it comes with a type (that one may/not have refreshed).
The extra `typeof` would compose decenlty with the former, in case one builds a term by hand.
How to get an open constr plus its type using ARGUMENT EXTEND is not clear to me, but ideally should compose with the former.

I'm not sure all that can be fixed in this PR, but at least the title should reflect what this PR actually optimizes (that is not clear)

ppedrot:
> The new code is in line with the current code of tactics, that is horrible.

That's what you get for writing in proof assembly, I'm afraid...

> and you need a context, that you cannot create unless you typecheck t first (you could put an evar there, but I guess you wanted to be faster)

That's pretty important, because otherwise the syntactic form of the type that appears in the context would be decided by unification, which is a very bad thing. Retyping is not specified, but it has a more easily definable semantics than pretyping.

> refreshing universes is also fishy. If the input term was a raw/glob one, this is what the pretyper would do I guess (typical ambiguity) but here it is not clear that one wants a fresh universe. Or maybe it is just because of algebraic universes one cannot really use. A comment would be welcome.

That one is from the `letin_tac` code. I agree it's pretty ugly, but unluckily, as you highlighted, there is no flag in retyping that would prevent generating algebraic universes.

> My impression is that, at the status quo of the API, one should have two primitives along these lines:

I agree for typeof, but I really don't agree about the `glob_constr` API. This ties the semantics of this tactic to the one of the pretyper, which is highly unspecified and subject to random subtle changes from time to time.

More generally, there is a notable tension between low-level tactics whose behaviour is fully specified but do little (and even so there is a lot of room for moving parts, e.g. here name generation and syntactic form generated by the retyping algorithm) and user-facing coffee-brewing tactics. The former are the ones you need for OCaml / Ltac(2) programming, while the latter is more purposed towards interactive use. The `pose` tactic is simple enough that the same program can cover both uses, but advanced examples like `destruct` or even `intros` are definitely not.

gares:
So I guess we agree on most of the things (but the title of the PR is still pushy/non-informative).

Wrt the recurrent argument that the pretyper is not^Wcannot specified, I hope to put an end to it with my ongoing research. Still, even today, there are not many ways to type check a let-in. Your code mimics what any elaborator would do, and calls "unspecified" code exactly when it would do.

herbelin:
@ppedrot: I took the initiative to choose a less crazy title for the PR. Change it as you wish to better respect your intent if needed.

I can be assignee and merge it if, @ppedrot, it makes things clearer from your point of view. Unless @gares wants to do it. CODEOWNERS-ly, it should be @mattam82, but I guess, @mattam82, that you would not mind not being released from this kind of duty?

ppedrot:
Failure was spurious, should I rerun a CI? Otherwise can you merge, @gares ?

Issue #7653, opened 2018-05-31
Provide an id to allow matching goals in prior proof steps with the corresponding goal in a later proof step


jfehrle:
#### Description of the problem

Currently there's no easy way to match a goal in a prior proof step with the current proof step--even if the goals are unchanged.  The current proof diff code (#6801) deals with this crudely--it compares the first focused goal in the old and new proof.  This is frequently correct, but there are common use cases where the new first goal is not a transformation of the old first goal, in which case showing diffs is not really useful.  Also, some tactics modify multiple goals--it would be nice to show diffs on all goals that have changed but there's no easy way to know which ones should be compared.

If a proof step converts G -> G' , both goals should have the same id (maybe call is "ancestry id"?).

ejgallego:
Indeed that would be great, however it is a bit tricky to implement due to the primitive proof engine not having a refine primitive, so for those using `tclEVARS` is a bit hard to track what happened.

We need to think about this.

Zimmi48:
Indeed, it's far from trivial. Currently we can only provide a very basic notion of relationship between goals: which one generated which. It has been considered for a long time by @herbelin and me and maybe some others to add a notion of side-goals and main goal, but there is no such thing right now.

> If a proof step converts G -> G' , both goals should have the same id (maybe call is "ancestry id"?).

Well, no, because a goal G can generate several sub-goals and then you would end up with all your goals sharing the same id. What you could want is each goal having an id (they actually already have one) and each goal storing the id of its parent (not currently done).

jfehrle:
> Well, no, because a goal G can generate several sub-goals and then you would end up with all your goals sharing the same id. What you could want is each goal having an id (they actually already have one) and each goal storing the id of its parent (not currently done).

I had supposed that new goals could be distinguished from existing goals and hence could be assigned a new id and that the id would simply be copied when G -> G'.  But that distinction may not possible to make.  Doing it as you suggest is reasonable.

Zimmi48:
When a goal is refined into several subgoals, the corresponding hole (existential variable) is instantiated with a term containing several new holes. No hole has a particular status with respect to the previous one. At least, not in the current model. As I was explaining, @herbelin and I considered tagging these hints with "main goal" or "side goal" annotations, but even in this case, there could be several main goals.

[...]

PR #8247, opened 2018-08-13
Show diffs on multiple changed goals; match old and new goal info

jfehrle:
**Kind:** bug fix / feature.

Currently the diff code only compares the first current goal of the old and new proof states.  While this is correct often enough to be useful, it doesn't handle the cases where:
- Multiple goals are changed or created with one tactic, such as "split."
- The order of the goals is changed, e.g. with "all: swap 1 3."

Note: the first commit changes the behavior of the recently-added Stm.get_prev_proof function.  This is so that tactics and commands that don't change the underlying proof, such as "all : swap 1 3", still show the same diffs from the last proof-changing tactic, though in a different spot in the output.

I hope we can get this into 8.9.

For example:

image

### DIFF
--- a/printing/printer.ml
+++ b/printing/printer.ml
@@ -494,17 +494,17 @@ let pr_transparent_state (ids, csts) =
 	str"CONSTANTS: " ++ pr_cpred csts ++ fnl ())
 
 (* display complete goal
- prev_gs has info on the previous proof step for diffs
- gs has info on the current proof step
+ og_s has goal+sigma on the previous proof step for diffs
+ g_s has goal+sigma on the current proof step
  *)
-let pr_goal ?(diffs=false) ?prev_gs gs =
-  let g = sig_it gs in
-  let sigma = project gs in
+let pr_goal ?(diffs=false) ?og_s g_s =
+  let g = sig_it g_s in
+  let sigma = project g_s in
   let env = Goal.V82.env sigma g in
   let concl = Goal.V82.concl sigma g in
   let goal =
     if diffs then
-      Proof_diffs.diff_goals ?prev_gs (Some gs)
+      Proof_diffs.diff_goal ?og_s g sigma
     else
       pr_context_of env sigma ++ cut () ++
         str "============================" ++ cut ()  ++
@@ -525,13 +525,18 @@ let pr_goal_name sigma g =
 let pr_goal_header nme sigma g =
   let (g,sigma) = Goal.V82.nf_evar sigma g in
   str "subgoal " ++ nme ++ (if should_tag() then pr_goal_tag g else str"")
-  ++ (if should_gname() then str " " ++ Pp.surround (pr_existential_key sigma g) else mt ())  
+  ++ (if should_gname() then str " " ++ Pp.surround (pr_existential_key sigma g) else mt ())
 
 (* display the conclusion of a goal *)
-let pr_concl n sigma g =
+let pr_concl n ?(diffs=false) ?og_s sigma g =
   let (g,sigma) = Goal.V82.nf_evar sigma g in
   let env = Goal.V82.env sigma g in
-  let pc = pr_goal_concl_style_env env sigma (Goal.V82.concl sigma g) in
+  let pc =
+    if diffs then
+      Proof_diffs.diff_concl ?og_s sigma g
+    else
+      pr_goal_concl_style_env env sigma (Goal.V82.concl sigma g)
+  in
   let header = pr_goal_header (int n) sigma g in
   header ++ str " is:" ++ cut () ++ str" "  ++ pc
 
@@ -698,13 +703,25 @@ let print_dependent_evars gl sigma seeds =
   in
   constraints ++ evars ()
 
+module GoalMap = Evar.Map
+
 (* Print open subgoals. Checks for uninstantiated existential variables *)
 (* spiwack: [seeds] is for printing dependent evars in emacs mode. *)
 (* spiwack: [pr_first] is true when the first goal must be singled out
    and printed in its entirety. *)
-(* [prev] is the previous proof step, used for diffs *)
-let pr_subgoals ?(pr_first=true) ?(diffs=false) ?prev
+(* [os_map] is derived from the previous proof step, used for diffs *)
+let pr_subgoals ?(pr_first=true) ?(diffs=false) ?os_map
                         close_cmd sigma ~seeds ~shelf ~stack ~unfocused ~goals =
+  let diff_goal_map =
+    match os_map with
+    | Some (_, diff_goal_map) -> diff_goal_map
+    | None -> GoalMap.empty
+  in
+
+  let map_goal_for_diff ng = (* todo: move to proof_diffs.ml *)
+    try GoalMap.find ng diff_goal_map  with Not_found -> ng
+  in
+
   (** Printing functions for the extra informations. *)
   let rec print_stack a = function
     | [] -> Pp.int a
@@ -738,23 +755,23 @@ let pr_subgoals ?(pr_first=true) ?(diffs=false) ?prev
     else str" " (* non-breakable space *)
   in
 
+  let get_ogs g =
+    match os_map with
+    | Some (osigma, _) -> Some { it = map_goal_for_diff g; sigma = osigma }
+    | None -> None
+  in
   let rec pr_rec n = function
     | [] -> (mt ())
     | g::rest ->
-        let pc = pr_concl n sigma g in
+       let og_s = get_ogs g in
+       let pc = pr_concl n ~diffs ?og_s sigma g in
         let prest = pr_rec (n+1) rest in
         (cut () ++ pc ++ prest)
   in
   let print_multiple_goals g l =
     if pr_first then
-      let prev_gs =
-        match prev with
-        | Some (prev_goals, prev_sigma) ->
-          if prev_goals = [] then None
-          else Some { it = List.hd prev_goals; sigma = prev_sigma}
-        | None -> None
-      in
-      pr_goal ~diffs ?prev_gs { it = g ; sigma = sigma }
+      let og_s = get_ogs g in
+      pr_goal ~diffs ?og_s { it = g ; sigma = sigma }
       ++ (if l=[] then mt () else cut ())
       ++ pr_rec 2 l
     else 
@@ -797,7 +814,7 @@ let pr_subgoals ?(pr_first=true) ?(diffs=false) ?prev
 	++ print_dependent_evars (Some g1) sigma seeds
       )
 
-let pr_open_subgoals_diff ?(quiet=false) ?(diffs=false) ?prev_proof proof =
+let pr_open_subgoals_diff ?(quiet=false) ?(diffs=false) ?oproof proof =
   (* spiwack: it shouldn't be the job of the printer to look up stuff
      in the [evar_map], I did stuff that way because it was more
      straightforward, but seriously, [Proof.proof] should return
@@ -833,15 +850,15 @@ let pr_open_subgoals_diff ?(quiet=false) ?(diffs=false) ?prev_proof proof =
      let { Evd.it = bgoals ; sigma = bsigma } = Proof.V82.background_subgoals p in
      let bgoals_focused, bgoals_unfocused = List.partition (fun x -> List.mem x goals) bgoals in
      let unfocused_if_needed = if should_unfoc() then bgoals_unfocused else [] in
-     let prev = match prev_proof with
-       | Some op ->
-         let (ogoals , _, _, _, _) = Proof.proof op in
-         let { Evd.it = obgoals; sigma = osigma } = Proof.V82.background_subgoals op in
-         let obgoals_focused = List.filter (fun x -> List.mem x ogoals) obgoals in
-         Some (obgoals_focused, osigma)
-       | None -> None
+     let os_map = match oproof with
+       | Some op when diffs ->
+         let (_,_,_,_, osigma) = Proof.proof op in
+         let diff_goal_map = Proof_diffs.make_goal_map oproof proof in
+         Some (osigma, diff_goal_map)
+       | _ -> None
      in
-     pr_subgoals ~pr_first:true ~diffs ?prev None bsigma ~seeds ~shelf ~stack:[] ~unfocused:unfocused_if_needed ~goals:bgoals_focused
+     pr_subgoals ~pr_first:true ~diffs ?os_map None bsigma ~seeds ~shelf ~stack:[]
+        ~unfocused:unfocused_if_needed ~goals:bgoals_focused
   end
 
 let pr_open_subgoals ~proof =
@@ -1023,22 +1040,14 @@ let print_and_diff oldp newp =
   | Some proof ->
     let output =
       if Proof_diffs.show_diffs () then
-        try pr_open_subgoals_diff ~diffs:true ?prev_proof:oldp proof
+        try pr_open_subgoals_diff ~diffs:true ?oproof:oldp proof
         with Pp_diff.Diff_Failure msg -> begin
           (* todo: print the unparsable string (if we know it) *)
-          Feedback.msg_warning Pp.(str ("Diff failure:" ^ msg ^ "; showing results without diff highlighting" ));
+          Feedback.msg_warning Pp.(str ("Diff failure: " ^ msg) ++ cut()
+              ++ str "Showing results without diff highlighting" );
           pr_open_subgoals ~proof
         end
       else
         pr_open_subgoals ~proof
     in
     Feedback.msg_notice output;;
-
-(* Do diffs on the first goal returning a Pp. *)
-let diff_pr_open_subgoals ?(quiet=false) o_proof n_proof =
-  match n_proof with
-  | None -> Pp.mt ()
-  | Some proof ->
-    try pr_open_subgoals_diff ~quiet ~diffs:true ?prev_proof:o_proof proof
-    with Pp_diff.Diff_Failure _ -> pr_open_subgoals ~proof
-      (* todo: print the unparsable string (if we know it) *)
--- a/printing/printer.mli
+++ b/printing/printer.mli
@@ -171,26 +171,46 @@ val pr_transparent_state   : transparent_state -> Pp.t
 
 (** Proofs, these functions obey [Hyps Limit] and [Compact contexts]. *)
 
-val pr_goal                : ?diffs:bool -> ?prev_gs:(goal sigma) -> goal sigma -> Pp.t
+(** [pr_goal ~diffs ~og_s g_s] prints the goal specified by [g_s].  If [diffs] is true,
+    highlight the differences between the old goal, [og_s], and [g_s].  [g_s] and [og_s] are
+    records containing the goal and sigma for, respectively, the new and old proof steps,
+    e.g. [{ it = g ; sigma = sigma }].
+*)
+val pr_goal                : ?diffs:bool -> ?og_s:(goal sigma) -> goal sigma -> Pp.t
 
-(** [pr_subgoals ~pr_first ~prev_proof pp sigma seeds shelf focus_stack unfocused goals]
-   prints the goals of the list [goals] followed by the goals in
-   [unfocused], in a short way (typically only the conclusion) except
-   for the first goal if [pr_first] is true. Also, if [diffs] is true
-   and [pr_first] is true, then highlight diffs relative to [prev] in the
-   output for first goal. This function prints only the
-   focused goals unless the conrresponding option
-   [enable_unfocused_goal_printing] is set. [seeds] is for printing
-   dependent evars (mainly for emacs proof tree mode). *)
-val pr_subgoals            : ?pr_first:bool -> ?diffs:bool -> ?prev:(goal list * evar_map) -> Pp.t option -> evar_map
+(** [pr_subgoals ~pr_first ~diffs ~os_map close_cmd sigma ~seeds ~shelf ~stack ~unfocused ~goals]
+   prints the goals in [goals] followed by the goals in [unfocused] in a compact form
+   (typically only the conclusion).  If [pr_first] is true, print the first goal in full.
+   [close_cmd] is printed afterwards verbatim.
+
+   If [diffs] is true, then highlight diffs relative to [os_map] in the output for first goal.
+   [os_map] contains sigma for the old proof step and the goal map created by
+   [Proof_diffs.make_goal_map].
+
+   This function prints only the focused goals unless the corresponding option [enable_unfocused_goal_printing] is set.
+   [seeds] is for printing dependent evars (mainly for emacs proof tree mode).  [shelf] is from
+   Proof.proof and is used to identify shelved goals in a message if there are no more subgoals but
+   there are non-instantiated existential variables.  [stack] is used to print summary info on unfocused
+   goals.
+*)
+val pr_subgoals            : ?pr_first:bool -> ?diffs:bool -> ?os_map:(evar_map * Evar.t Evar.Map.t) -> Pp.t option -> evar_map
                              -> seeds:goal list -> shelf:goal list -> stack:int list
                              -> unfocused: goal list -> goals:goal list -> Pp.t
 
 val pr_subgoal             : int -> evar_map -> goal list -> Pp.t
-val pr_concl               : int -> evar_map -> goal -> Pp.t
 
-val pr_open_subgoals_diff  : ?quiet:bool -> ?diffs:bool -> ?prev_proof:Proof.t -> Proof.t -> Pp.t
-val diff_pr_open_subgoals  : ?quiet:bool -> Proof.t option -> Proof.t option -> Pp.t
+(** [pr_concl n ~diffs ~og_s sigma g] prints the conclusion of the goal [g] using [sigma].  The output
+    is labelled "subgoal [n]".  If [diffs] is true, highlight the differences between the old conclusion,
+    [og_s], and [g]+[sigma].  [og_s] is a record containing the old goal and sigma, e.g. [{ it = g ; sigma = sigma }].
+*)
+val pr_concl               : int -> ?diffs:bool -> ?og_s:(goal sigma) -> evar_map -> goal -> Pp.t
+
+(** [pr_open_subgoals_diff ~quiet ~diffs ~oproof proof] shows the context for [proof] as used by, for example, coqtop.
+    The first active goal is printed with all its antecedents and the conclusion.  The other active goals only show their
+     conclusions.  If [diffs] is true, highlight the differences between the old proof, [oproof], and [proof].  [quiet]
+     disables printing messages as Feedback.
+*)
+val pr_open_subgoals_diff  : ?quiet:bool -> ?diffs:bool -> ?oproof:Proof.t -> Proof.t -> Pp.t
 val pr_open_subgoals       : proof:Proof.t -> Pp.t
 val pr_nth_open_subgoal    : proof:Proof.t -> int -> Pp.t
 val pr_evar                : evar_map -> (Evar.t * evar_info) -> Pp.t
--- a/printing/proof_diffs.ml
+++ b/printing/proof_diffs.ml
@@ -14,39 +14,26 @@ Proof General requires minor changes to make the diffs visible, but this code
 shouldn't break the existing version of PG.  See pp_diff.ml for details on how
 the diff works.
 
-Diffs are computed for the hypotheses and conclusion of the first goal between
-the old and new proofs.
+Diffs are computed for the hypotheses and conclusion of each goal in the new
+proof with its matching goal in the old proof.
 
-Diffs can be enabled with the Coq commmand "Set Diffs on|off|removed." or
-'-diffs "on"|"off"|"removed"' on the OS command line.  The "on" option shows only the
-new item with added text, while "removed" shows each modified item twice--once
-with the old value showing removed text and once with the new value showing
-added text.
+Diffs can be enabled in coqtop with 'Set Diffs "on"|"off"|"removed"' or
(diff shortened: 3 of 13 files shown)


jfehrle:
@ejgallego @Zimmi48 This solves the underlying problem of #7653.

printing/proof_diffs.ml
```diff
@@ -317,31 +333,244 @@ let hyp_list_to_pp hyps =
...
+  !nid_to_oid
+
+
+(* based on the code in show_proof *)
+let to_constr p =
+  let open CAst in
+  let pprf = Proof.partial_proof p in
+  (* todo: When does pprf have multiple elements?  Can it have zero elements? *)
+  let t = List.hd pprf in
+  let sigma, env = Pfedit.get_current_context ~p () in
```
jfehrle: I'm pretty sure zero elements is not possible here.  If it's possible to have multiple elements, I'd like to know how to generate such a case so I can try it.

jfehrle:
The ci-fiat-crypto failure looks unrelated.

@herbelin, most of the changes are in the printing directory, so perhaps you should be the first reviewer?  Thanks.

printing/proof_diffs.ml
Zimmi48: @ejgallego can probably answer this. I think there is support for multiple proofs that is only used by the Derive plugin AFAIR.

jfehrle: Thanks.  BTW, I tried `Set Nested Proofs Allowed.' and entering a new Goal but I still didn't get multiple entries in pprf.

SkySkimmer: It might be for
~~~coq
Lemma foo : A with bar : B.
Proof. exact A_val. exact B_val. Qed.
~~~
I didn't check so this is just a conjecture.

jfehrle: Ah, can you send me the exact text to try?

image

ppedrot (changes requested):
Apart from the stylistic changes mentioned, I find the solution implemented here horribly contrived and fragile. I guess it's OK because this is just for user-side printing, but still, it bugs me as an ugly hack that is affecting the proof manipulating API when this should ideally be performed at the UI level.

jfehrle:
I would have preferred a different approach myself.  Ideally, associating goals in the new proof state with their antecedents would be handled in the lower level code when a tactic is applied.  I was advised that this was not a simple change.  My understanding is that tactics provide the new proof state as a whole, so the associations are not trivially available even at that level, so a good bit of refactoring would be needed.  I thought it was better to implement this outside the lower level code so it's much less likely to raise any critical issue (e.g. the diff code is in try blocks and the user has to enable the feature), particularly as someone relatively new to Coq and who has not worked at all with the lower level code.

printing/printer.mli
```diff
@@ -171,23 +171,25 @@ val pr_transparent_state   : transparent_state -> Pp.t
 
 (** Proofs, these functions obey [Hyps Limit] and [Compact contexts]. *)
 
-val pr_goal                : ?diffs:bool -> ?prev_gs:(goal sigma) -> goal sigma -> Pp.t
+val pr_goal                : ?diffs:bool -> ?prev_g_s:(goal sigma) -> goal sigma -> Pp.t
 
-(** [pr_subgoals ~pr_first ~prev_proof pp sigma seeds shelf focus_stack unfocused goals]
+module GoalMap = Evar.Map
```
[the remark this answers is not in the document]
jfehrle: Got it.

What I really wanted was to define a Map module in goal.mli.  But that opens Util, which defines its own conflicting Map module--it took me a couple hours to realize that.  I wasn't keen on moving my Map definition before the `open`.

I also experimented for a couple hours without success with variations such as:

```
module Map : sig include Map.S with type key = goal end      (* in .ml file *)
module Map = Map.Make(struct type t = goal let compare = Evar.compare end)      (* in .mli file *)
```

It's not easy to figure out what might work from the OCaml doc or the other sources I checked on the internet.

stm/stm.ml
```diff
@@ -1231,9 +1231,23 @@ end = struct (* {{{ *)
...
-      let did = fold_until back_tactic 1 id in
-      get_proof ~doc did
-    with Not_found -> None
+      let np = get_proof ~doc id in
+      match np with
+      | None -> None
+      | Some cp ->
+        let did = ref id in
+        let rv = ref np in
+        while
```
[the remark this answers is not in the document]
jfehrle: Most imperative languages have a `do ... while (condition)` that would work nicely here that OCaml doesn't support.  What I did is an analog.  I can add a ref bool `done` flag and move the test into the body.

But if you think there shouldn't be ref variables as well, what would you suggest replacing them with?

jfehrle:
Updated.  Let me know if you have more comments.  Thanks.

printing/proof_diffs.ml
Zimmi48:
``` coq
Variables A B : Set.
Inductive tree : Set := node : A -> forest -> tree

with forest : Set :=
| leaf : B -> forest
| cons : tree -> forest -> forest.

Theorem tree_size : tree -> nat with forest_size : forest -> nat.
```

jfehrle: Thanks for the example text.  I still find pprf has only one element.  Perhaps it is a vestige and the code no longer generates pprf with more than one element.

ppedrot (approved):

jfehrle:
Fixed a minor bug in the diff printing adn did a little refactoring for clarity in the code I've added.

printing/printer.ml
```diff
@@ -494,17 +494,17 @@ let pr_transparent_state (ids, csts) =
 	str"CONSTANTS: " ++ pr_cpred csts ++ fnl ())
 
 (* display complete goal
- prev_gs has info on the previous proof step for diffs
- gs has info on the current proof step
+ prev_g_s has goal+sigma on the previous proof step for diffs
```
herbelin: Is this now `og_s`?

printing/proof_diffs.ml
```diff
@@ -14,8 +14,8 @@ Proof General requires minor changes to make the diffs visible, but this code
 shouldn't break the existing version of PG.  See pp_diff.ml for details on how
 the diff works.
 
-Diffs are computed for the hypotheses and conclusion of the first goal between
-the old and new proofs.
+Diffs are computed for the hypotheses and conclusion of the each goal in the new
```
herbelin: the each -> each?

printing/proof_diffs.ml
```diff
@@ -317,31 +327,277 @@ let hyp_list_to_pp hyps =
...
+
+let diff_goal ?og_s ng ns =
+  let (hyps_pp_list, concl_pp) = diff_goal_info (unwrap og_s) (goal_info ng ns) in
   let open Pp in
   v 0 (
     (hyp_list_to_pp hyps_pp_list) ++ cut () ++
     str "============================" ++ cut () ++
     concl_pp);;
+
+(* CODE TO DETERMINE WHICH GOALS TO COMPARE BETWEEN THE OLD AND NEW PROOFs *)
```
herbelin: Why uppercase?

printing/proof_diffs.ml
```diff
@@ -317,31 +327,277 @@ let hyp_list_to_pp hyps =
...
+open CAst
+
+(* code for Array.iter2, which was added in 4.03
+todo: remove when we no longer support compiling with 4.02 *)
+let my_array_iter2 f a b =
+  if Array.length a <> Array.length b then
+    invalid_arg "Array.iter2: arrays must have the same length"
+  else
+    for i = 0 to Array.length a - 1 do f (Array.unsafe_get a i) (Array.unsafe_get b i) done
+
```
herbelin: This is already in clib/cArray.ml (available as `Array.iter2`).

printing/proof_diffs.ml
herbelin: It seems that the ability to have several goals in `partial_proof` is not used in general. It is used in the `derive` plugin but I did not investigate more.

herbelin (commented):
This is ok as far as I'm concerned. See also the various comments.

jfehrle:
I've made changes addressing your comments.  Thanks for the review!

@gares, any comments on the stm change?

stm/stm.ml
```diff
@@ -1231,9 +1231,22 @@ end = struct (* {{{ *)
...
-      let did = fold_until back_tactic 1 id in
-      get_proof ~doc did
-    with Not_found -> None
+      let np = get_proof ~doc id in
+      match np with
+      | None -> None
+      | Some cp ->
+        let did = ref id in
+        let rv = ref np in
+        let _done = ref false in
```
gares: why `_done`? The variable is used, did you get a warning?

gares (approved):

stm/stm.ml
jfehrle: You mean why the name `_done` and not `done`?  `done` is a keyword, it can't be used for a variable name, right?

Zimmi48: Indeed, but then find another name. @gares is right: pre-pending `_` is conventionally reserved to unused variables.

herbelin: Note: for a function called `match`, we used `match_`!

printing/proof_diffs.ml
```diff
@@ -317,31 +324,270 @@ let hyp_list_to_pp hyps =
...
     concl_pp);;
+
+
+(*** Code to determine which calls to compare between the old and new proofs ***)
+
+open Glob_term
+open Names
+open CAst
+
+let term_to_name x =
```
ejgallego: I am not sure `glob_expr` should be used here; this PR certainly deserves a deeper look.

stm/stm.ml
jfehrle: Ah, I was unaware of the convention.

printing/proof_diffs.ml
jfehrle: This code was based on reverse engineering `show_proof`, the code for the `Show Proof` command, which is based entirely on same type that `x` has.  If that's good for `Show Proof`, why wouldn't it be good here?

BTW `x` is of type `'a glob_constr_r`; `glob_expr` is not (no longer?) a defined type in source tree.

ejgallego: > If that's good for Show Proof, why wouldn't it be good here?

I'm much afraid the code for `Show Proof` may not be good.

jfehrle: What data structure would you suggest using instead?

This code is far simpler than `Show Proof`--it's just a somewhat verbose tree traversal constructed mechanically from the type definition.  If `Show Proof` is using this data structure, it seems overzealous to say that diffs shouldn't.  If and when `Show Proof` is restructured, I can make the corresponding changes here.

The consequences of a failure to match or an incorrect match here are limited--the diff code will catch the failure to match, then print an error message and the proof step without diffs.  If there's an incorrect match, the diffs will look a bit odd, but the user can turn off diffs if that's distracting--and I'll make improvements.  `match_goals` is used only when multiple old goals have been rewritten, I suspect most proof steps don't do that.

jfehrle:
I think "heuristics" is the wrong term; if my assumptions are correct, this is a full solution.

I added a description of the matching in comments just before Proof_diffs.make_goal_map_i (and a reference to them in the commit message).  I also adjusted some variable names to be more precise about things that are goal ids versus evar names.

Also, `_done` is now `done_.

printing/proof_diffs.ml
ejgallego: I didn't see where show proof uses a glob_constr , care to point that out?

jfehrle:
@herbelin Would you mark your review as "Approved"?

Still waiting to hear back from @ejgallego.  There's not a lot of time left before 8.9 beta if significant changes are needed.

printing/proof_diffs.ml
ejgallego: I stilll don't see where the `glob_constr` is used.

ejgallego (commented):
I am fine with the PR, it seems to improve code in a key few places.

I am however concerned with the goal matching going on in `Proof_diffs`, the algorithm is not clearly described IMO, and the style is a bit far from functional programming. Also, I am not sure what is going on with the use of `glob_expr`. WDYT @ppedrot ?

proofs/proof.ml
```diff
@@ -452,3 +452,12 @@ module V82 = struct
     { pr with proofview ; shelf }
 
 end
+
+let all_goals p =
+  let add gs set =
+    List.fold_left (fun s g -> Goal.Set.add g s) set gs in
+  let (goals,stack,shelf,given_up,_) = proof p in
+    let set = add goals Goal.Set.empty in
+    let set = List.fold_left (fun s gs -> let (g1, g2) = gs in add g1 (add g2 set)) set stack in
```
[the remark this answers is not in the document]
jfehrle: This is a set, not a map.  You mean I should update lib/cSet.ml?  To simplify line 461 only?  `stack` is a list of (list of goals, list of goals)--that seems very specific to this routine, hard to imagine that would be reused elsewhere if it was in cSet.

jfehrle:
Pushed an updated description.

herbelin:
@ejgallego: you seem to be the one who looked the most closely at this PR, do you want to be the assignee?

herbelin:
Maybe should I be the assignee...

If I understand correctly, all local comments have been addressed but the remark of @ppedrot about the API level at which the feature should be implemented is still open, is that right?

If yes, could @ppedrot explain what he has in mind so as to understand if a modified implementation is worth in the longer term?

As of today, the current implementation is anyway the only implementation available, so why not experimenting with it as it is?

If this position is acceptable for everyone, I shall merge by Monday (@silene: this shall be ok?).

ppedrot:
My remark does not prevent merging as-is. I don't have a better proposal currently, but eventually this code should be moved upwards to the UI where it really belongs.

ejgallego:
> If I understand correctly, all local comments have been addressed but the remark of @ppedrot about the API level at which the feature should be implemented is still open, is that right?

@herbelin So far most of the code changes in this PR do seem like an improvement, however, the use of `glob_expr` in `Proof_diffs` seems strange and I would like to understand it better before we merge. Unfortunately I am a bit short on time to be able to look at that code in detail.

That being said, I don't expect that issue to be a blocker for 8.9, so if you think the code there seems OK feel free to merge.

ejgallego:
> eventually this code should be moved upwards to the UI where it really belongs.

What code? IMHO the diff code does belong in the printing subsystem as editors do have a hard time making proper diffs without having access to the underlying AST.

herbelin:
> That being said, I don't expect that issue to be a blocker for 8.9, so if you think the code there seems OK feel free to merge.

@ejgallego: You probably looked at the code more than I did. So, the question is more to me whether the PR improves the diff feature or not even if the implementation can be made differently. For the first question, is the answer correctly a yes? Then, wrt the implementation, will it be possible that @ppedrot, you and @jfehrle find a protocol to make the code evolve? Eventually, I can spend some time to understand the issues which @ppedrot and you are raising but I'm sorry that my experience in UI is really too weak to have a proper opinion myself yet.

ejgallego:
> You probably looked at the code more than I did. So, the question is more to me whether the PR improves the diff feature or not even if the implementation can be made differently.

I am concerned about some of the code, in the diffs file; the rest seems fine. My concern is big enough as to make this wait unless some other person can have a look at the use of `Glob_term`.

I feel like @ppedrot has some other concerns.

jfehrle:
Are you discussing who feels comfortable doing the merge or whether this can be merged as-is?

If neither of you (@herbelin, @ejgallego) feel comfortable merging this, then perhaps @ppedrot would be willing to do so since he knows the lower level code better.


Regarding whether the code can be merged as-is:

> I feel like @ppedrot has some other concerns.

He approved the PR 3 weeks ago, so his concerns were not that great.  And more recently he said "My remark does not prevent merging as-is."

@ejgallego: Given:
* your earlier statement "I am fine with the PR, it seems to improve code in a key few places"
* that this change is low risk (the feature can be turned off),
* that there are no specific comments I can respond to and 
* that I've offered to rework it in the future if necessary,

why delay this fix/improvement any longer?  Let's not make the best the enemy of the good (as Voltaire put it).

ejgallego:
@jfehrle I am sorry if I haven't been clear, but my request about why the code is using glob_constr has still not been answered and that's a critical concern.

The code you pointed out above has nothing to do with the use of an internal AST which worries me.

jfehrle:
I have no idea why that would be a critical concern.  Would you give at least a little detail so I understand your concern?  What are the alternatives?  Would it help to move some of these routines to a different file?

> The code you pointed out above has nothing to do with the use of an internal AST which worries me.

What does "The code you pointed out above" refer to?

ejgallego:
See this comment:

https://github.com/coq/coq/pull/8247#discussion_r211646080

jfehrle:
> I am not sure `glob_expr` should be used here; this PR certainly deserves a deeper look.

What would address your concerns?  Further review by @ppedrot or someone else on this point?  I don't know what you're thinking here so I don't know what to do about it.

silene:
> If this position is acceptable for everyone, I shall merge by Monday (@silene: this shall be ok?).

The date does not matter much, since v8.9 was branched last week. This does not seem like a fix, so I will backport it only if developers make a good case for it.

Zimmi48:
@silene 
>This does not seem like a fix, so I will backport it only if developers make a good case for it.

This is a refinement of a new (imperfect) feature that has never been part of any release so it seems like a good candidate for backporting.

ejgallego:
> What would address your concerns?

My question is, "why are you using a `glob_expr` instead of a `constr_expr`". Why do you need that printing function there.

You already tried to give an answer by pointing to some code that does nothing like yours, so that's not a very good one I'd dare to say.

ppedrot:
Just to be clear: this code is wrong, but AFAIU there is no way to make it correct, because it relies on false premises. So diffing subgoals is just a horrible heuristic, which is OK because this is just UI. That's the reason why I'd like to see this code in another component I don't want to have to care about.

ejgallego:
> Just to be clear: this code is wrong, but AFAIU there is no way to make it correct, because it relies on false premises. So diffing subgoals is just a horrible heuristic

I think you are a bit radical here. Diffing in general is a heuristic anyways; so IMO the real question here is: "what is the heuristic? Is it implemented in a reasonable way?"

Note that we could move this to some upper layer, but printing seems OK. `toplevel` uses it so it cannot go too high anyways.

jfehrle:
> My question is, "why are you using a `glob_expr` instead of a `constr_expr`".

I'm completely new to this part of the code and unfamiliar with the difference between the two.  I don't know enough about them to respond intelligently.  I was able to reverse-engineer something based on glob_constr (not glob_expr).  I can look at changing it if it's important, but I probably would need some help getting pointed in the right direction.

> Why do you need that printing function there.

Are you referring to show_proof?  That's just what I used to figure out how the data structure works.

printing/proof_diffs.ml
jfehrle: > I stilll don't see where the `glob_constr` is used.

Sorry, I didn't see this question until just now.  In Constrextern.extern, the `c` in `match c with` is of type `glob_constr_r`.

ejgallego:
Do you mean that the code is not actually used?

jfehrle:
> Do you mean that the code is not actually used?

The diffs code does not call show_proof, if that is your question.

ejgallego:
Oh indeed I see. Note that coq is a mature system thus new dead debug code is not allowed in the production release.

Additionally it had the effect of delaying this PR.

ejgallego:
Sorry Jim, I am totally lost. I guess I'll have to wait until I have time for a proper review.

jfehrle:
Any idea when you'll have time for a proper review?
I'd be happy to do a Skype call at that time; that might help get us into sync on the PR.

ejgallego:
> Any idea when you'll have time for a proper review?

Early next week for sure. I think beta is still a couple of weeks away so we should be good.

ejgallego:
Anyways let me state my position here, if I had to chose whether I recommend this going into 8.9 or waiting for 8.10 in this current state I'd recommend getting the PR in, but I think it'd be wise to do one more pass before merging, in particular `Glob_constr` shouldn't be used in `Proof_diffs`.

jfehrle:
> It has never been part of any release, but it does modify some external interfaces. For example, this pull request adds an og_s argument to various exported functions. I have absolutely no idea what og_s means and it is not documented anywhere, as far as I can tell.

Certainly I can add a description of the og_s arguments.  I would observe, though, that
* Only one out of about 50 functions defined in `printer.mli` has any description of the purpose or arguments.  (Though it's not unreasonable to require more documentation for newly-added items.)
* I expect there won't be a lot of people writing code against these changed external interfaces.

Zimmi48:
>Though it's not unreasonable to require more documentation for newly-added items

:+1: Fortunately, the PR review system allows us to enforce better quality on new code.

jfehrle:
The example in issue https://github.com/coq/coq/issues/8481 breaks the tree-matching heuristic.  It doesn't handle the `1,4: pose True.` case because the new proof is not just the old proof with substitutions for one or more goals.  (If the `pose` is done in 2 steps, the match succeeds with a simpler heuristic: if only one goal is added and one goal is removed between the steps, then the new goal is a modification of the old goal.)

```
Unnamed_thm < repeat apply conj.
  :
Unnamed_thm < Show Proof.
(conj ?Goal (conj ?Goal0 (conj ?Goal1 ?Goal2)))

Unnamed_thm < 1,4:pose True.
Warning:
Diff failure: Unable to match goals between old and new proof states (5)
Showing results without diff highlighting
  :
Unnamed_thm < Show Proof.
(conj (let P := True in ?Goal1)
   (conj ?Goal (conj ?Goal0 (let P := True in ?Goal2))))
```

jfehrle:
> The example in issue #8481 breaks the tree-matching heuristic.

I spoke too soon.  It was an easily-fixable bug.

jfehrle:
I pushed the latest, updating comments in printer.mli as requested and fixing the bug I mentioned.

@ejgallego, please look at this.  If you want to discuss on gitter, please suggest a time between 9 AM and 9 PM California time.  There's not much time left before 8.9 if I'm going to make non trivial changes.

ejgallego:
@jfehrle , there is something I am not understanding. Why `idetop` has to know about the diff goals stuff?

That seems strange. I would suppose that the toplevels would just call the proper goal printing routine and it would the routine itself who would make the choice. Semantic information should be properly encapsulated.

Ok, I did a review, and I think all looks good but I am convinced there is a problem in `Proof_diffs`: the type of `match_goals` is incorrect, it should take `constr_expr` not `glob_expr`. This is indeed because the implementation of `to_constr` is incorrect. You should instead just use `Constrextern.extern_constr` and variants. Unfortunately this is a serious problem and it must be corrected before the code can go in. While you are at if please don't use exceptions for control in match_goals, but a boolean or even a richer type.

Please also remove the debug code from Proof_diffs. It should either go to `debug_printers`, or you can use serialization.

jfehrle:
> there is something I am not understanding. Why idetop has to know about the diff goals stuff?

At the least, I think this code should select the old proof step rather than having lower levels pick it.  And, IIRC, I got circular dependencies when I tried to call `Stm.get_prev_proof` from `Printer` or `Proof_diffs`.

I can try to move more of the diff-related code here into Proof_diffs.  But trying to hide all awareness of diffs here means scattering the changes widely and maybe replicating some routines.  I think the loss of locality would make it harder to follow what's happening.  I'd probably need to change `Proof.map_structured_proof`, `Idetop.process_goal`, then the print routines the latter calls: `pr_goal_concl_style_env` and `pr_compacted_decl`, and pass various wrapper variables hither and yon.  Trying to hide details is not always an improvement.

> the type of match_goals is incorrect, it should take constr_expr not glob_expr. This is indeed because the implementation of to_constr is incorrect. You should instead just use Constrextern.extern_constr and variants. Unfortunately this is a serious problem and it must be corrected before the code can go in.

Would you explain the difference between the two types and their uses, and why it is a serious problem?  Otherwise I've learned nothing.

This is a lot of stuff to change.

ejgallego:
Will answer later on idetop as it is not so urgent. Wrt glob_constr, we decided some time ago that it is an internal datatype. Even if it works for your uses you should not rely on it. A very famous plug-in was implement over that type and the pain was famous. I wrote an explanation some time ago but it had been lost in the mail archives. Hope to do a blog post some day 

Thus, you can either perform the matching directly on constr (which seems like the best choice for the purpose I think) or in constr_expr, which is the AST pre-printing. Note that the change should be straightforward, it looks harder than it actually is.

ejgallego:
On idetop: indeed the upper layer should select the previous state but then delegate to the printing routines what to print. Thus, I think there should not be any conditional on the proof diffs state.

ejgallego:
I don't see how this change would replicate code, as far as I can see it should have the contrary effect.

Regarding glob_constr, it is really a kind of low-level AST, needed by some particular implementation reasons , but indeed it supposes some limitations as it is generated without input from the pretyper. This impedes to use type-based resolution for many things. But the short story is, don't use that type. And please file a big from the part of the codebase you got that code as it must be fixed.

jfehrle:
I tried moving `export_pre_goals`, `process_goal_diffs`, etc. from `idetop.ml` into `proof_diffs.ml`.   First I get the error "Unbound module Interface".  Then, if I add, say, `let open Interface in` in the appropriate place, I still get the same error message, but now it's on the `let open`.  It seems like these 2 routines have to remain in idetop.ml because they rely on `Interface`.

Defining identical types `goal`, `pre_goals` and `goals` in Proof_diffs seems to make them incompatible with the same types in Interface.  My impression is that defining types identically in 2 different files makes them incompatible--the OCaml doc doesn't say anything about this.  That makes it mighty hard to use!

I don't fully understand how export_pre_goals works, in particular `Interface.whatever` and especially `pgs.Proof.whatever`.  So it becomes a long session of guesswork and experimentation, and even spending an hour or two researching it on the web is often unhelpful.

I don't understand why `Proof.map_structured_proof` refers to `zipper` rather than `stack`.  It's confusing.

So--can you explain in some detail how I can do what you're suggesting?  Or shall we let it be?

jfehrle:
Furthermore, `process_goal` has to remain in idetop.ml; it can't go into proof_diffs.ml because it calls routines in printer.ml.  For efficiency reasons, it's better to call `Proof_diffs.make_goal_map` once for each call to `goals` rather than creating it repeatedly for each call to `process_goal`.  That would mean keeping most of the diff code in idetop.ml.

Zimmi48:
>My impression is that defining types identically in 2 different files makes them incompatible

That is correct. It is true in Coq too.

>Or shall we let it be?

Indeed, I think that @ejgallego views this as less of a concern than the question of `glob_constr` vs `constr` or `constr_expr`.

ejgallego:
I think you are trying to do a different, way more complicated thing to what I had in mind, moving interface to printer is not going to work obviously.

jfehrle:
I changed this to use `const_expr` instead of `glob_constr` as you requested.  I removed the debug code from `match_goals` (though I kept `db_goal_map`).  I didn't know what to do for idetop.

The exceptions raised in `match_goals` are for violations of the routine's assumptions, which I believe won't happen.  If they are raised, they should be shown to the user and reported as bugs.  I wouldn't change that.  (Also, FWIW, it seems to me that ending the tree traversal when I get one of these errors would mean checking after every call to `constr_expr`, so many lines would change for an improbable error.  I suppose I could set a `string ref` to indicate an error was found, but in that case it seems odd to continue the traversal anyway.)

Please take another look.  If it looks OK, please merge it.  And if not, let me know.

ejgallego:
> I changed this to use `const_expr` instead of `glob_constr` as you requested.

After thinking a bit more, wouldn't it make sense to do the matching over `Constr.t` directly? Think about it, you may want to refine it later, as matching may be more precise over the former.

I still have a gripe about the use of imperative code in `Proof_diffs`. That style is a bit far from the OCaml used in code, but what should we do? If I understand correctly you don't have a lot of OCaml experience, so it will take more iterations to get the code to a more idiomatic style. I was meaning to pass the map itself as a functional value; same for idetop, with the right higher-order pattern you can remove the conditional.

All in all, I don't oppose getting the code as-is, despite some OCaml annoyances in general the patch looks pretty good. This has already the green tick of two devs; but I am not sure they looked at the totality of the patch. I think I will let 2/3 days for @herbelin to look at the current version, and unless he complains I'll merge.

jfehrle:
OK, sounds good.

> After thinking a bit more, wouldn't it make sense to do the matching over Constr.t directly?

I don't know what the difference is between using Constr.t and const_expr, so I have no opinion.  Also, the matching is conceptually quite simple even though it's 150+ lines of code, not sure what might be gained.

> I was meaning to pass the map itself as a functional value; same for idetop, with the right higher-order pattern you can remove the conditional.

So something like add the map and the old proof as the initial arguments to `process_goal`, then curry those values for the diffs case?  I don't think I can add optional arguments to `process_goal` without changing the declaration for `Proof.map_structured_proof`, right?  (which I thought I should not change)

Thanks for the review.

ejgallego:
> I don't know what the difference is between using Constr.t and const_expr, so I have no opinion. Also, the matching is conceptually quite simple even though it's 150+ lines of code, not sure what might be gained.

There is a bit of text here: https://github.com/ejgallego/coq-serapi/blob/v8.8/FAQ.md#how-many-asts-does-coq-have also in the genarg.mli header: https://github.com/coq/coq/blob/master/lib/genarg.mli

_"hellish"_ could be a good way to explain it. Basically in constr you have less elaborated terms, so you gain a "purer" form to work with.

Regarding the changes, indeed, you could try like that; I don't think changing `Proof.map_structured_proof` is frowned upon. To what me regards I think that the API in proof is far from finish so of course it can take improvements from users.

Thanks for the patch and your patience.

ejgallego (approved):

ejgallego:
Last call for @herbelin 's comments [or request for more time], otherwise I shall merge tomorrow.

PR #8249, opened 2018-08-14
Remove unneeded file stm/workerLoop.mli

jfehrle:
Remove unneeded file: workerLoop.ml and workerLoop.mli were moved to toplevel in commit 382ee49
but stm/workerLoop.mli was not removed as it should have been.  Hasta la vista.

**Kind:** cleanup


gares (approved):

gares:
I'm on VAC, but I'll take care of this one as soon as I'm back

PR #8254, opened 2018-08-15
Inline the definition of CClosure.mk_clos_deep.

ppedrot:
An important part of this function was dead code, due to the fact it was only used for whd evaluation of specific shapes of constr.

### DIFF
--- a/kernel/cClosure.ml
+++ b/kernel/cClosure.ml
@@ -551,45 +551,7 @@ let mk_clos_vect env v = match v with
   [|mk_clos env v0; mk_clos env v1; mk_clos env v2; mk_clos env v3|]
 | v -> Array.Fun1.map mk_clos env v
 
-(* Translate the head constructor of t from constr to fconstr. This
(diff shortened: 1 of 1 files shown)


maximedenes:
Can we get a bench, just in case?

ppedrot:
Bench running here: https://ci.inria.fr/coq/job/benchmark-part-of-the-branch/503/.

ppedrot:
The bench had to be aborted due to fiat-crypto not compiling at that point, but it only showed a slight speedup, which was not unexpected. AFAICT this PR is good to go.

herbelin:
Expected assignee is @maximedenes.

maximedenes (approved):

maximedenes:
@ppedrot sorry, this needs a rebase. I can merge as soon as it is done. Thanks!

ppedrot:
@maximedenes ping. The bedrock failure is spurious.

PR #8255, opened 2018-08-15
Fast typing of application nodes

### LINKED ISSUE #8232: Checking non-dependent applications takes quadratic time
#### Version

~~~~~
compiled on Jul 5 2018 7:25:50 with OCaml 4.06.0
~~~~~

#### Operating system

Arch Linux, up to date as of today. `Linux ashryn 4.17.13-arch1-1-ARCH #1 SMP PREEMPT Mon Aug 6 23:06:31 UTC 2018 x86_64 GNU/Linux`

Checking non-dependent function applications takes quadratic time. This benchmark applies an axiom `f` to `n` arguments of type `unit` using `exact` and measures time taken during tactic execution and `Qed`.

~~~~~
echo; python -c 'n=1000; print ("Axiom f : " + "unit->"*n + "unit.\nGoal unit.\nTime exact (f "+" tt"*n + ").\nTime Qed.")' > /tmp/a.v && coqc /tmp/a.v 2>&1
echo; python -c 'n=2000; print ("Axiom f : " + "unit->"*n + "unit.\nGoal unit.\nTime exact (f "+" tt"*n + ").\nTime Qed.")' > /tmp/a.v && coqc /tmp/a.v 2>&1
echo; python -c 'n=4000; print ("Axiom f : " + "unit->"*n + "unit.\nGoal unit.\nTime exact (f "+" tt"*n + ").\nTime Qed.")' > /tmp/a.v && coqc /tmp/a.v 2>&1
echo; python -c 'n=8000; print ("Axiom f : " + "unit->"*n + "unit.\nGoal unit.\nTime exact (f "+" tt"*n + ").\nTime Qed.")' > /tmp/a.v && coqc /tmp/a.v 2>&1
echo; python -c 'n=12000; print ("Axiom f : " + "unit->"*n + "unit.\nGoal unit.\nTime exact (f "+" tt"*n + ").\nTime Qed.")' > /tmp/a.v && coqc /tmp/a.v 2>&1
echo; python -c 'n=16000; print ("Axiom f : " + "unit->"*n + "unit.\nGoal unit.\nTime exact (f "+" tt"*n + ").\nTime Qed.")' > /tmp/a.v && coqc /tmp/a.v 2>&1
echo; python -c 'n=20000; print ("Axiom f : " + "unit->"*n + "unit.\nGoal unit.\nTime exact (f "+" tt"*n + ").\nTime Qed.")' > /tmp/a.v && coqc /tmp/a.v 2>&1
~~~~~

Data:

~~~~~
# n     exact Qed
# 1000  0.015 0.01
# 2000  0.055 0.04
# 4000  0.225 0.095
# 8000  0.764 0.379
# 12000 1.688 0.854
# 16000 3.034 1.54
# 20000 5.131 2.474
~~~~~

Plot of the first two columns and a fitted quadratic:

quadratic

ppedrot:
This PR implements a type-checking algorithm for application nodes that is asymptotically faster than the current implementation. It does it by changing the representation of `FProd` nodes so that they contain term closures instead of `fconstr`. The rationale is that it contains an open term, and we reuse the same trick as already implemented for lambda and case representation. Then we leverage this to substitute lazily values in product types.

This shouldn't change the type generated by the typechecker.

Fixes #8232. The pretyper is still quadratic but I don't think there is hope for that one though.

### DIFF
--- a/kernel/cClosure.ml
+++ b/kernel/cClosure.ml
@@ -300,7 +300,7 @@ and fterm =
   | FCoFix of cofixpoint * fconstr subs
   | FCaseT of case_info * constr * fconstr * constr array * fconstr subs (* predicate and branches are closures *)
   | FLambda of int * (Name.t * constr) list * constr * fconstr subs
-  | FProd of Name.t * constr * constr * fconstr subs
+  | FProd of Name.t * fconstr * constr * fconstr subs
   | FLetIn of Name.t * fconstr * fconstr * constr * fconstr subs
   | FEvar of existential * fconstr subs
   | FLIFT of int * fconstr
@@ -586,10 +586,10 @@ let rec to_constr lfts v =
         Term.compose_lam (List.rev tys) f
     | FProd (n, t, c, e) ->
       if is_subs_id e && is_lift_id lfts then
-        mkProd (n, t, c)
+        mkProd (n, to_constr lfts t, c)
       else
         let subs' = comp_subs lfts e in
-        mkProd (n, subst_constr subs' t, subst_constr (subs_lift subs') c)
+        mkProd (n, to_constr lfts t, subst_constr (subs_lift subs') c)
     | FLetIn (n,b,t,f,e) ->
       let subs = comp_subs (el_lift lfts) (subs_lift e) in
         mkLetIn (n, to_constr lfts b,
@@ -872,7 +872,7 @@ and knht info e t stk =
     | CoFix cfx -> { norm = Cstr; term = FCoFix (cfx,e) }, stk
     | Lambda _ -> { norm = Cstr; term = mk_lambda e t }, stk
     | Prod (n, t, c) ->
-      { norm = Whnf; term = FProd (n, t, c, e) }, stk
+      { norm = Whnf; term = FProd (n, mk_clos e t, c, e) }, stk
     | LetIn (n,b,t,c) ->
       { norm = Red; term = FLetIn (n, mk_clos e b, mk_clos e t, c, e) }, stk
     | Evar ev -> { norm = Red; term = FEvar (ev, e) }, stk
@@ -996,7 +996,7 @@ and norm_head info tab m =
           let c = mk_clos (subs_lift e) f in
           mkLetIn(na, kl info tab a, kl info tab b, kl info tab c)
       | FProd(na,dom,rng,e) ->
-          mkProd(na, kl info tab (mk_clos e dom), kl info tab (mk_clos (subs_lift e) rng))
+          mkProd(na, kl info tab dom, kl info tab (mk_clos (subs_lift e) rng))
       | FCoFix((n,(na,tys,bds)),e) ->
           let ftys = Array.Fun1.map mk_clos e tys in
           let fbds =
--- a/kernel/cClosure.ml
+++ b/kernel/cClosure.ml
@@ -300,7 +300,7 @@ and fterm =
   | FCoFix of cofixpoint * fconstr subs
   | FCaseT of case_info * constr * fconstr * constr array * fconstr subs (* predicate and branches are closures *)
   | FLambda of int * (Name.t * constr) list * constr * fconstr subs
-  | FProd of Name.t * fconstr * fconstr
+  | FProd of Name.t * constr * constr * fconstr subs
   | FLetIn of Name.t * fconstr * fconstr * constr * fconstr subs
   | FEvar of existential * fconstr subs
   | FLIFT of int * fconstr
@@ -584,9 +584,12 @@ let rec to_constr lfts v =
         let tys = List.mapi (fun i (na, c) -> na, subst_constr (subs_liftn i subs) c) tys in
         let f = subst_constr (subs_liftn len subs) f in
         Term.compose_lam (List.rev tys) f
-    | FProd (n,t,c)   ->
-        mkProd (n, to_constr lfts t,
-                   to_constr (el_lift lfts) c)
+    | FProd (n, t, c, e) ->
+      if is_subs_id e && is_lift_id lfts then
+        mkProd (n, t, c)
+      else
+        let subs' = comp_subs lfts e in
+        mkProd (n, subst_constr subs' t, subst_constr (subs_lift subs') c)
     | FLetIn (n,b,t,f,e) ->
       let subs = comp_subs (el_lift lfts) (subs_lift e) in
         mkLetIn (n, to_constr lfts b,
@@ -869,7 +872,7 @@ and knht info e t stk =
     | CoFix cfx -> { norm = Cstr; term = FCoFix (cfx,e) }, stk
     | Lambda _ -> { norm = Cstr; term = mk_lambda e t }, stk
     | Prod (n, t, c) ->
-      { norm = Whnf; term = FProd (n, mk_clos e t, mk_clos (subs_lift e) c) }, stk
+      { norm = Whnf; term = FProd (n, t, c, e) }, stk
     | LetIn (n,b,t,c) ->
       { norm = Red; term = FLetIn (n, mk_clos e b, mk_clos e t, c, e) }, stk
     | Evar ev -> { norm = Red; term = FEvar (ev, e) }, stk
@@ -992,8 +995,8 @@ and norm_head info tab m =
       | FLetIn(na,a,b,f,e) ->
           let c = mk_clos (subs_lift e) f in
           mkLetIn(na, kl info tab a, kl info tab b, kl info tab c)
-      | FProd(na,dom,rng) ->
-          mkProd(na, kl info tab dom, kl info tab rng)
+      | FProd(na,dom,rng,e) ->
+          mkProd(na, kl info tab (mk_clos e dom), kl info tab (mk_clos (subs_lift e) rng))
       | FCoFix((n,(na,tys,bds)),e) ->
           let ftys = Array.Fun1.map mk_clos e tys in
           let fbds =
--- a/kernel/typeops.ml
+++ b/kernel/typeops.ml
@@ -151,6 +151,11 @@ let type_of_abstraction _env name var ty =
 let make_judgev c t = 
   Array.map2 make_judge c t
 
+let rec check_empty_stack = function
+| [] -> true
+| CClosure.Zupdate _ :: s -> check_empty_stack s
+| _ -> false
+
 let type_of_apply env func funt argsv argstv =
   let open CClosure in
   let len = Array.length argsv in
@@ -159,7 +164,9 @@ let type_of_apply env func funt argsv argstv =
   let rec apply_rec i typ =
     if Int.equal i len then term_of_fconstr typ
     else
-      let typ, _ = whd_stack infos tab typ [] in
+      let typ, stk = whd_stack infos tab typ [] in
+      (** The return stack is known to be empty *)
+      let () = assert (check_empty_stack stk) in
       match fterm_of typ with
       | FProd (_, c1, c2, e) ->
         let arg = argsv.(i) in
--- a/kernel/typeops.ml
+++ b/kernel/typeops.ml
@@ -152,27 +152,33 @@ let make_judgev c t =
   Array.map2 make_judge c t
 
 let type_of_apply env func funt argsv argstv =
+  let open CClosure in
   let len = Array.length argsv in
-  let rec apply_rec i typ = 
-    if Int.equal i len then typ
-    else 
-      (match kind (whd_all env typ) with
-      | Prod (_,c1,c2) ->
-	let arg = argsv.(i) and argt = argstv.(i) in
-	  (try
-	     let () = conv_leq false env argt c1 in
-	       apply_rec (i+1) (subst1 arg c2)
-	   with NotConvertible ->
-	     error_cant_apply_bad_type env
-	       (i+1,c1,argt)
-	       (make_judge func funt)
-	       (make_judgev argsv argstv))
-	    
+  let infos = create_clos_infos all env in
+  let tab = create_tab () in
+  let rec apply_rec i typ =
+    if Int.equal i len then term_of_fconstr typ
+    else
+      let typ, _ = whd_stack infos tab typ [] in
+      match fterm_of typ with
+      | FProd (_, c1, c2, e) ->
+        let arg = argsv.(i) in
+        let argt = argstv.(i) in
+        let c1 = term_of_fconstr (mk_clos e c1) in
+        begin match conv_leq false env argt c1 with
+        | () -> apply_rec (i+1) (mk_clos (Esubst.subs_cons ([| inject arg |], e)) c2)
+        | exception NotConvertible ->
+          error_cant_apply_bad_type env
+            (i+1,c1,argt)
+            (make_judge func funt)
+            (make_judgev argsv argstv)
+        end
       | _ ->
-	error_cant_apply_not_functional env 
-	  (make_judge func funt)
-	  (make_judgev argsv argstv))
-  in apply_rec 0 funt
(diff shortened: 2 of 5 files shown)


gares:
I think it would be easier to understand if the "growing substitution was just an argument of apply_rec rather than a node in the fconstr. Instead of iterating subst1 one would keep the arguments processed so far and do a substn. Going trough the cclosure business seems identical (in performance) but requires mastering that code, that is way more general than what we are doing here.

Or at least, this is my understanding of this patch.
And, If I'm not mistaken, the code I suggest should also work for the pretyper (I interpret your comment as "I can't do it in the pretyper because we don't use fconstr there).

ppedrot:
> I think it would be easier to understand if the "growing substitution" was just an argument of apply_rec rather than a node in the fconstr. Instead of iterating subst1 one would keep the arguments processed so far and do a substn.

This is less efficient in general, as you need to perform an inconditional O(n) substitution instead of delaying it to the conversion check through the machine. I had tried to do that at first but I realized this, and it makes the change moot from the point of view of asymptotic complexity.

> And, If I'm not mistaken, the code I suggest should also work for the pretyper (I interpret your comment as "I can't do it in the pretyper because we don't use fconstr there", please confirm).

The issue with the pretyper is indeed that it manipulates terms, not terms withing a substitution, so you have to pay for this cost even if it turns out to be useless. But then your approach suffers from the same issue as above.

kernel/cClosure.ml
```diff
@@ -368,7 +368,7 @@ and fterm =
   | FCoFix of cofixpoint * fconstr subs
   | FCaseT of case_info * constr * fconstr * constr array * fconstr subs (* predicate and branches are closures *)
   | FLambda of int * (Name.t * constr) list * constr * fconstr subs
-  | FProd of Name.t * fconstr * fconstr
+  | FProd of Name.t * constr * constr * fconstr subs
```
gmalecha: I don't completely understand the code here, but why is it `constr * constr * fconstr subs` rather than `fconstr * constr * fconstr subs`? My understanding is that the first argument (after the name is the type), that isn't open, and we lose sharing (the purpose of `fconstr` if I understand correctly) if we convert it back into a `constr` during reduction.

All this said, this seems to be the way that lambda and case work, so...I'm not certain.

gares:
> This is less efficient in general, as you need to perform an inconditional O(n) substitution instead of delaying it to the conversion check through the machine. I had tried to do that at first but I realized this, and it makes the change moot from the point of view of asymptotic complexity.

You are right, with your approach the reduction machine shares more computations.
Said that, I would really be surprised if that changed in practice since the complexity factor does not come for the size/complexity of the arguments of the application (what you optimize better using the machine), but rather their number (the n in your O(n)).
`substn` also does not lose sharing, so the additional cost of traversing the types of the arguments (assuming there is nothing to substitute, the worst of the simple approach) is matched by the cost of the subsequent comparison. That is, identity-map + comparison = comparison (in both time and space).

You said you started writing the simpler code. Do you still have it around (for a bench)?

kernel/typeops.ml
```diff
@@ -151,27 +151,33 @@ let make_judgev c t =
...
+  let rec apply_rec i typ =
+    if Int.equal i len then term_of_fconstr typ
+    else
+      let typ, _ = whd_stack infos tab typ [] in
+      match fterm_of typ with
+      | FProd (na, c1, c2, e) ->
+        let arg = argsv.(i) in
+        let argt = argstv.(i) in
+        let c1 = term_of_fconstr (mk_clos e c1) in
+        begin match conv_leq false env argt c1 with
```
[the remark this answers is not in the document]
gmalecha: I agree that the fast path is important. I'm not well versed enough to know how expensive the conversion check is in the case of syntactic equality, but I'll keep looking at the code. Suppose that I have two larger terms that differ only at some lead node. Syntactic equality will fail, but conversion will succeed. Will the conversion be slow (because it will reduce the full term) or fast because it will only convert the small term?

gmalecha:
@gares Doesn't `substn` always create a copy of the term? So won't your "identity-map" actually be "copy term"? Therefore, you might lose a physical equality, or at least you could do a lot of allocation. I agree that it doesn't change the asymptotic complexity though.

gares:
Last time I checked it was a "smart map" ie using == to allocate only if necessary.

ppedrot:
@gares I don't understand the algorithm you're thinking about, at least to obtain the complexity you're mentioning. We seem to agree that applying an arbitrary substitution to a term of size n is O(n). But then, how do you thread this substitution inside the `apply_rec` function?

Let's paraphrase the typing rule. Say you want to type the application of a term `t : T` to `u_1 : A_1 ... u_n : A_n`. You need first to evaluate `T` in whnf so as to check it is of the form `forall x : X, S`, then check that `X ≡ A_1`, and then recursively perform this check with `S{x := u_1}` and `u_2 ... u_n`. The problem is that the shape of the type might depend on the argument `u_1`, so that you **have** to substitute it upfront for the whnf to make sense.

For instance, if you want to evaluate `forall b : bool, if b then nat -> nat else nat` applied to arguments `true 0`. You cannot delay the substitution of `true` here otherwise you'll fail badly.

My patch allows this by leveraging the delayed substitution mechanism of the kernel, instead of having to do that in some ad-hoc way.

gares:
You are right, sorry for the noise

herbelin:
I couldn't grasp from a quick reading if the discussion between @gmalecha and @ppedrot had converged. In any case, this PR has no assignee. It should be @maximedenes.

gmalecha:
@herbelin I think what is here is an improvement over what exists right now.

I don't understand the code well enough to know for certain, but from my tracing, I think that it might still be possible to improve things. To be certain, I would need to better understand how `fconstr` and the unification algorithm work.

maximedenes:
@ppedrot how about benchmarking this?

ppedrot:
I did this at some point, but no idea which run it was. I'll try to relaunch a bench.

maximedenes:
> I did this at some point, but no idea which run it was. I'll try to relaunch a bench.

@ppedrot ping

ppedrot:
I am pretty sure I redid it but cannot find it *again*. Sigh, let's run yet another bench and raise the entropy of the universe.

ppedrot:
This is https://ci.inria.fr/coq/job/benchmark-part-of-the-branch/522/console. It seems that unimath is slightly faster and VST slightly slower, but I find this strange. Last time there was really no difference above 1%, so this might be noise. I can retry with those two extremes to be sure.

maximedenes:
Doesn't this show that this optimization is useless?

ppedrot:
This is not an optimization, this patch changes the algorithmic complexity of the typing function. In most cases it is useless, but for huge terms it might be important.

ppedrot:
Let me try a variant of this patch where the domain of products is still a closure. This should combine the best of both worlds, i.e. fast typing of applications while keeping a back-and-forth profile similar to the one we have today.

ppedrot:
I pushed an additional patch that lets the domain of `FProd` as an `fconstr`, and it seems to essentially make unimath faster and everything else unchanged, while still solving the algorithmic issue.
```
│              coq-unimath │ 1301.78 1320.63 -1.43 % │  3614480067689  3666574500355 -1.42 % │  6036550521115  6207254365343 -2.75 % │  866704 1053868 -17.76 % │  348   21 +1557.14 % │
│               coq-sf-vfa │   46.58   46.83 -0.53 % │   128767370448   129181795503 -0.32 % │   205046937468   205311122658 -0.13 % │  537120  541316  -0.78 % │   28    6 +366.67 % │
│      coq-formal-topology │   57.25   57.54 -0.50 % │   156743949164   156592440569 +0.10 % │   235620302475   235699180506 -0.03 % │  480708  480620  +0.02 % │    0    0  +nan % │
│               coq-geocoq │ 2809.68 2818.87 -0.33 % │  7812339303607  7838691032417 -0.34 % │ 12499196197126 12496265376303 +0.02 % │ 1287196 1287508  -0.02 % │  392  400 -2.00 % │
│                coq-flocq │  584.74  586.24 -0.26 % │  1627168070995  1630908611583 -0.23 % │  2366532464648  2368776509495 -0.09 % │ 1760556 1576580 +11.67 % │  298   16 +1762.50 % │
│                 coq-hott │  339.48  340.13 -0.19 % │   941192875734   942984364137 -0.19 % │  1428946254745  1437157361458 -0.57 % │  610420  610348  +0.01 % │  219    0  +nan % │
│                coq-gpfsl │ 1198.69 1199.61 -0.08 % │  3339415049489  3340085370948 -0.02 % │  4354370465113  4398379811345 -1.00 % │ 1761604 1584420 +11.18 % │   21   41 -48.78 % │
│   coq-mathcomp-ssreflect │   64.82   64.85 -0.05 % │   178625952913   178172323168 +0.25 % │   253026519511   252668040681 +0.14 % │  535800  535244  +0.10 % │    2    0  +nan % │
│               coq-sf-plf │   73.67   73.70 -0.04 % │   203554161654   203702804145 -0.07 % │   296027547168   295909705395 +0.04 % │  517080  518700  -0.31 % │   15    0  +nan % │
│       coq-mathcomp-field │  465.92  465.68 +0.05 % │  1296522871665  1296071953630 +0.03 % │  2109541880860  2114127249441 -0.22 % │  807748  805908  +0.23 % │    0  156 -100.00 % │
│                coq-color │  735.46  735.02 +0.06 % │  2038216770989  2036021764478 +0.11 % │  2458738844733  2455294816734 +0.14 % │ 1425944 1427260  -0.09 % │   44   45 -2.22 % │
│         coq-math-classes │  247.06  246.81 +0.10 % │   680502702953   680036181407 +0.07 % │   889583119478   889684437784 -0.01 % │  524864  524960  -0.02 % │   84   84 +0.00 % │
│ coq-mathcomp-real-closed │  187.59  187.38 +0.11 % │   521867586062   521715344304 +0.03 % │   785759414481   786488545757 -0.09 % │  830668  837800  -0.85 % │    0    0  +nan % │
│   coq-mathcomp-odd-order │ 1414.36 1410.44 +0.28 % │  3939450337173  3927832137151 +0.30 % │  6600981184617  6610049183240 -0.14 % │ 1361244 1356452  +0.35 % │    1    0  +nan % │
│         coq-fiat-parsers │  716.94  714.84 +0.29 % │  1990538749464  1984692983804 +0.29 % │  3054620016480  3061833593778 -0.24 % │ 2589880 2896624 -10.59 % │   31  444 -93.02 % │
│             coq-compcert │  884.20  881.34 +0.32 % │  2451298239279  2443897940945 +0.30 % │  3562093855713  3558389009623 +0.10 % │ 1349960 1343244  +0.50 % │  104   26 +300.00 % │
│   coq-mathcomp-character │  293.58  292.38 +0.41 % │   815605621451   812968203868 +0.32 % │  1179579793919  1182371909362 -0.24 % │ 1122380 1122964  -0.05 % │    0    0  +nan % │
│     coq-mathcomp-algebra │  205.32  204.48 +0.41 % │   569705546805   567966477855 +0.31 % │   795708884658   795041207026 +0.08 % │  644764  645468  -0.11 % │   23    9 +155.56 % │
│    coq-mathcomp-solvable │  220.66  219.66 +0.46 % │   611601063570   609983204552 +0.27 % │   865337324559   871360238793 -0.69 % │  857712  814920  +5.25 % │   10  169 -94.08 % │
│           coq-coquelicot │  103.19  102.72 +0.46 % │   285608049194   284482910161 +0.40 % │   383199997975   381836761055 +0.36 % │  715648  717132  -0.21 % │  280   13 +2053.85 % │
│                  coq-vst │ 1718.44 1709.78 +0.51 % │  4777210472508  4752823953913 +0.51 % │  6519214939655  6505737643930 +0.21 % │ 2223688 2222504  +0.05 % │  273  270 +1.11 % │
│            coq-fiat-core │  134.53  133.78 +0.56 % │   374100182913   373624701383 +0.13 % │   499101186365   499031759442 +0.01 % │  501336  499648  +0.34 % │  395  180 +119.44 % │
│    coq-mathcomp-fingroup │   84.65   84.17 +0.57 % │   234141984994   232853199808 +0.55 % │   344276617385   344705413025 -0.12 % │  586336  588964  -0.45 % │  212    0  +nan % │
│                 coq-corn │ 1549.90 1540.94 +0.58 % │  4302193921893  4275718589491 +0.62 % │  6335685732869  6356612773925 -0.33 % │  858392  862868  -0.52 % │   44   44 +0.00 % │
│                coq-sf-lf │   43.54   43.27 +0.62 % │   120030497884   119677292263 +0.30 % │   194642822510   194728311210 -0.04 % │  429096  427908  +0.28 % │  116    0  +nan % │
│              coq-bignums │   98.17   97.47 +0.72 % │   271413540950   269833156283 +0.59 % │   385871845120   385405491193 +0.12 % │  541160  538464  +0.50 % │   15    0  +nan % │
│          coq-fiat-crypto │ 6357.78 6309.76 +0.76 % │ 17688578218922 17550716416314 +0.79 % │ 29027552221663 28981537632469 +0.16 % │ 2617332 2642160  -0.94 % │ 1271 1346 -5.57 % │
└──────────────────────────┴─────────────────────────┴───────────────────────────────────────┴───────────────────────────────────────┴──────────────────────────┴───────────────────| 
```

kernel/typeops.ml
```diff
@@ -151,27 +151,33 @@ let make_judgev c t =
...
+  let tab = create_tab () in
+  let rec apply_rec i typ =
+    if Int.equal i len then term_of_fconstr typ
+    else
+      let typ, _ = whd_stack infos tab typ [] in
+      match fterm_of typ with
+      | FProd (_, c1, c2, e) ->
+        let arg = argsv.(i) in
+        let argt = argstv.(i) in
+        let c1 = term_of_fconstr c1 in
```
pythonesque: Two comments:

First, you should include a comment to explain why you know you can drop the stack (it is initially empty).  It turns out to be true that there are no shifts in the stack, I believe; but the reason is not obvious since whd_stack does not call fapp_stack (it is only the case here because [1] e doesn't have any shifts or lifts to start with, and [2] there are never any shifts in the returned stack (or shifts or lifts in substitutions in the returned fconstr) if you only call kni on [1] a stack without shifts and [2] a fconstr whose substitutions have no shifts or lift; which is an invariant that both isn't actually documented anywhere, and wasn't even true before this PR).  Alternatively, you could just uselessly run fapp_stack here.

Second, term_of_fconstr can be rather expensive; it might be nice to delay doing this (especially since conv_leq just turns the constr back into a fconstr) and do a fast path equality on fterms instead.  It's true conversion can be arbitrarily expensive, and the "fast path" equality still has to walk both terms at least once if it succeeds.  But I would think you could still do fast path equality of fterms at lower cost than either full conversion, or converting to to_constr and then doing fast path equality there; and if that failed, the slow path would not be as slow (no need to convert back for conv_leq).  It would be more complicated, though; it would be something similar to the existing conv_leq with all reduction (beta, delta, iota, etc.) disabled (but not identical, because I think that would cause assertions in the current function to throw).

ppedrot:
@pythonesque I have added an assertion for your first point. Regarding your second point, I don't know how to do that properly. Checking for pointer equality is meaningless for fconstrs, and converting back the argument to constr for the fast-path would make the whole trick moot. So AFAIC, this PR is ready for me.

maximedenes:
@pythonesque does it look good to you, now?

pythonesque:
Yes, it looks good now (I think any zupdates are safe to avoid applying here, and if that's not true it will be caught later during reduction).

maximedenes (approved):

ppedrot:
@maximedenes ping.

maximedenes:
> @maximedenes ping.

Ah ah, this PR was already on my screen ;)

PR #8256, opened 2018-08-16
[ci] Fix QuickChick by adding new simple-io dependency.

Zimmi48:
**Kind:** infrastructure.

QuickChick CI is currently broken on v8.8 because of QuickChick/QuickChick#101 which added a dependency on a new Coq package (simple-io).

@Lysxia @liyishuai let us know when such things happen. In particular, let us know when/if you eventually merge this in master because we would need to apply the same changes to Coq master.

@SkySkimmer The `ci-simple-io.sh` script doesn't contain:

```bash
ci_dir="$(dirname "$0")"

# This script could be included inside other ones
# Let's avoid to source ci-common twice in this case
if [ -z "${CI_BUILD_DIR}" ];
then
    . "${ci_dir}/ci-common.sh"
fi
```

because I've seen you say somewhere that it was actually useless. Is that correct? If so, then we might want to clean-up these scripts.


SkySkimmer:
I think you still need to include ci-common, the unneeded part is doing it conditionally.

Zimmi48:
Oh right! Fixed.

SkySkimmer (approved):

PR #8260, opened 2018-08-17
Tweak diff options in CoqIDE

jfehrle:
1) Make the diff setting a persistent setting.
2) Changing the diff setting from the menu will reprint the proof.
3) Generate a warning message if the user enters a "Set Diffs xx" command.
4) Adjust display names for diffs in View menu.

**Kind:** bug fix / feature.

Addresses a few issues identified at the end of #8141 (post-merge).

### DIFF
--- a/ide/coqide.ml
+++ b/ide/coqide.ml
@@ -1046,16 +1046,24 @@ let build_ui () =
       ~accel:"F1"
       ~callback:(cb_on_current_term MiscMenu.show_hide_query_pane);
     GAction.group_radio_actions
-      ~callback:begin function
-        | 0 -> List.iter (fun o -> Opt.set o "off") Opt.diff_item.Opt.opts
-        | 1 -> List.iter (fun o -> Opt.set o "on") Opt.diff_item.Opt.opts
-        | 2 -> List.iter (fun o -> Opt.set o "removed") Opt.diff_item.Opt.opts
-        | _ -> assert false
+      ~init_value:(
+        let v = diffs#get in
+        List.iter (fun o -> Opt.set o v) Opt.diff_item.Opt.opts;
+        if v = "on" then 1
+        else if v = "removed" then 2
+        else 0)
+      ~callback:begin fun n ->
+        (match n with
+        | 0 -> List.iter (fun o -> Opt.set o "off"; diffs#set "off") Opt.diff_item.Opt.opts
+        | 1 -> List.iter (fun o -> Opt.set o "on"; diffs#set "on") Opt.diff_item.Opt.opts
+        | 2 -> List.iter (fun o -> Opt.set o "removed"; diffs#set "removed") Opt.diff_item.Opt.opts
+        | _ -> assert false);
+        send_to_coq (fun sn -> sn.coqops#show_goals)
         end
       [
-        radio "Unset diff" 0 ~label:"Unset _Diff";
-        radio "Set diff" 1 ~label:"Set Di_ff";
-        radio "Set removed diff" 2 ~label:"Set _Removed Diff";
+        radio "Unset diff" 0 ~label:"_Don't show diffs";
+        radio "Set diff" 1 ~label:"Show diffs: only _added";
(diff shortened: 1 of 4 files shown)


ide/coqide.ml
```diff
@@ -1046,16 +1046,24 @@ let build_ui () =
       ~accel:"F1"
       ~callback:(cb_on_current_term MiscMenu.show_hide_query_pane);
     GAction.group_radio_actions
-      ~callback:begin function
-        | 0 -> List.iter (fun o -> Opt.set o "off") Opt.diff_item.Opt.opts
-        | 1 -> List.iter (fun o -> Opt.set o "on") Opt.diff_item.Opt.opts
-        | 2 -> List.iter (fun o -> Opt.set o "removed") Opt.diff_item.Opt.opts
-        | _ -> assert false
+      ~init_value:(
+        let v = diffs#get in
```
ppedrot: I don't think this code is run at the right time. The initial value is taken when the module is linked, and loading the preferences has not occurred yet at that point. Rather, you should tie the value of this menu to the diff preference by using its connect signal so that they are always synchronized.

ide/coqide.ml
```diff
@@ -1046,16 +1046,24 @@ let build_ui () =
...
+        if v = "on" then 1
+        else if v = "removed" then 2
+        else 0)
+      ~callback:begin fun n ->
+        (match n with
+        | 0 -> List.iter (fun o -> Opt.set o "off"; diffs#set "off") Opt.diff_item.Opt.opts
+        | 1 -> List.iter (fun o -> Opt.set o "on"; diffs#set "on") Opt.diff_item.Opt.opts
+        | 2 -> List.iter (fun o -> Opt.set o "removed"; diffs#set "removed") Opt.diff_item.Opt.opts
+        | _ -> assert false);
+        send_to_coq (fun sn -> sn.coqops#show_goals)
```
ppedrot: Same thing: let the diff preference `connect` callback do the work of setting the `Coq.PrintOpt.diff_item` option rather than doing at the same time.

ide/coqide.ml
jfehrle: The current code works correctly.  This code is inside the build_ui function, which is executed **after** the object files are loaded.  coqide_main.ml4 ends with:

```
let () =
  load_prefs ();
  :
```

Which would be executed at object file load time.

I'm not seeing how I could do this with connect.  I expect I need code something like this:

```
rb_unset = radio "Unset diff" 0 ~label:"_Don't show diffs";
rb_unset#connect#activate ~callback:(... diffs#set "off");
rb_unset#connect#SOMETHING ~callback:(... return diffs#get)
```

In which the radio button calls the SOMETHING callback to use the preference value when it is bringing up the menu.  I've spent a good while looking at gAction.ml in the GTK library but I don't see anything candidate for SOMETHING.  (BTW, the experimentation goes very slow because it takes about an hour for me to rebuild CoqIDE on Windows--I'm stuck doing full builds.)

After playing with that a couple hours I also tried this:

```
        let active = diffs#get in
        let to_str v = if v = 1 then "on" else if v = 2 then "removed" else "off" in
        let rb1 = radio "Unset diff" 0 ~active ~label:"_Don't show diffs" in
```

but the compiler objects--there is no `~active` on `radio` (= GAction.add_radio_action).

If you can advise me some on how to get connect to work, I'll try it, but considering that this code works correctly, how about if we let it be?

jfehrle:
@ppedrot This code works just fine and I can't figure out how to use a connect callback after several hours fumbling around.  I don't have GTK 2 documentation and there are no examples for radio buttons in the Coq code.

Would you approve the change or otherwise respond to my comments from last week?  Thanks.

ppedrot (approved):
For the time being it is OK to merge it as-is because it is already working, if I have time I'll come up with a later cleanup.

jfehrle:
Thanks, can you also do the merge?  I understand submitters are not supposed to merge their own changes.

jfehrle:
Awesome, thanks!

PR #8264, opened 2018-08-17
More efficient computation of avoided variables during pretyping.

ppedrot:
Slightly better allocation and semantic-wise.


herbelin:
Looks good. No need to be delayed I guess.

herbelin:
Assigning myself and planning to merge on Monday evening. Still time to say something...

Issue #6567, opened 2018-01-09
Inconsistency between the syntax of Theorem ... with and Fixpoint ... with


Zimmi48:
#### Version

8.7.1

#### Description of the problem

```coq
Inductive tree := Leaf : tree | Node : forest -> tree
with forest := Tree : tree -> forest | Merge : forest -> forest -> forest.

Fixpoint size (t : tree) : nat with size_f (f : forest) : nat.
Proof.
  - destruct t as [ | f ].
    + exact 1.
    + exact (1 + size_f f).
  - destruct f as [ t | f1 f2 ].
    + exact (size t).
    + exact (size_f f1 + size_f f2).
Defined.
```

and

```coq
Theorem size (t : tree) : nat with size_f (f : forest) : nat.
Proof.
  - destruct t as [ | f ].
    + exact 1.
    + exact (1 + size_f f).
  - destruct f as [ t | f1 f2 ].
    + exact (size t).
    + exact (size_f f1 + size_f f2).
Defined.
```

are both accepted. (They are the same modulo the keyword `Fixpoint` or `Theorem`.)

```coq
Theorem size : tree -> nat with size_f : forest -> nat.
Proof.
  - intros [].
    + exact 1.
    + exact (1 + size_f f).
  - intros [ t | f1 f2 ].
    + exact (size t).
    + exact (size_f f1 + size_f f2).
Defined.
```

is also accepted but not:

```coq
Fixpoint size : tree -> nat with size_f : forest -> nat.
```

which fails with "A fixpoint needs at least one parameter."

Shouldn't the two syntax be consistent in what they accept? Or is there a reason for the way things are?

Note that the manual correctly describes `Fixpoint ... with ... .` as a variant of `Theorem ... with ... .` where binders are not optional (they are not enclosed in square brackets).

Zimmi48:
While making the allowed syntax for `Definition` and `Theorem` / `Lemma` converge a related question arises: should the form `Definition ... with ...` be allowed or should the form `Theorem ... with ...` be removed?

SkySkimmer:
Note that 
~~~coq
Class Foo T := foo : T -> nat.

Theorem size : Foo tree with size_f : Foo forest.
~~~
is also rejected, so `Theorem with` only allows skipping the automatic introduction (IIUC), it doesn't allow using  a convertible form of the type.

This is a small enough difference that I would lean to wontfix (and maybe open an issue to allow using Fixpoint with an unreduced type).

[...]

PR #8265, opened 2018-08-17
Deprecate [Theorem ... with].

Zimmi48:
**Kind:** clean-up

This is a proposal to deprecate (and subsequently remove the syntax) `Theorem ... with` which allows to define mutually recursive fixed points in proof-editing mode. The migration path is trivial (replace `Theorem` with `Fixpoint` which also supports not providing a body and using the proof-editing mode instead). Preliminary testing has shown that this is only used (among the projects in our CI) in Equations (cc @mattam82) and CompCert (cc @xavierleroy).

I find the command inconsistent and confusing because `Theorem` will change its meaning if there is a `with` or not (will insert a `fix` / `cofix` or not) whereas `Fixpoint` / `CoFixpoint` won't (they will always insert a `fix` / `cofix`).
There are other additional inconsistencies due to a partly different implementation. See #6567 (should the limitation of the `Fixpoint` form noted in this issue be addressed before deprecating the `Theorem` form?).

This makes it also more complicated to unite `Definition` and `Theorem` as was my intention with the project of introducing `Theorem ... := ...`. In this project, the only practical remaining difference between `Definition` and `Theorem` will be the default resulting opacity (but even this could be made tweakable using an attribute).

### DIFF
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -1313,28 +1313,6 @@ Chapter :ref:`Tactics`. The basic assertion command is:
 
       These commands are all synonyms of :n:`Theorem @ident {? @binders } : type`.
 
-.. cmdv:: Theorem @ident {? @binders } : @type {* with @ident {? @binders } : @type}
-
-   This command is useful for theorems that are proved by simultaneous induction
-   over a mutually inductive assumption, or that assert mutually dependent
-   statements in some mutual co-inductive type. It is equivalent to
-   :cmd:`Fixpoint` or :cmd:`CoFixpoint` but using tactics to build the proof of
-   the statements (or the body of the specification, depending on the point of
-   view). The inductive or co-inductive types on which the induction or
-   coinduction has to be done is assumed to be non ambiguous and is guessed by
-   the system.
-
-   Like in a :cmd:`Fixpoint` or :cmd:`CoFixpoint` definition, the induction hypotheses
-   have to be used on *structurally smaller* arguments (for a :cmd:`Fixpoint`) or
-   be *guarded by a constructor* (for a :cmd:`CoFixpoint`). The verification that
-   recursive proof arguments are correct is done only at the time of registering
-   the lemma in the environment. To know if the use of induction hypotheses is
-   correct at some time of the interactive development of a proof, use the
-   command :cmd:`Guarded`.
-
-   The command can be used also with :cmd:`Lemma`, :cmd:`Remark`, etc. instead of
-   :cmd:`Theorem`.
-
 .. cmdv:: Definition @ident {? @binders } : @type
 
    This allows defining a term of type :token:`type` using the proof editing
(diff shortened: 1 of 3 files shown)


cpitclaudel (approved):
The doc changes look good to me (I don't have an opinion on the code change).

Zimmi48:
I'd like to have this in 8.9* so can developers comment and possibly also users of the feature @xavierleroy (for CompCert) and @mattam82 (for Equations)?

*If this eventually get merged, I'd like to have it merged rather sooner than later so that we can start benefiting from this clean-up in 8.10.

Zimmi48:
ping @mattam82 @herbelin @xavierleroy (if nobody protests, it probably means that it is OK to remove this syntax, doesn't it?)

herbelin:
> ping @mattam82 @herbelin @xavierleroy (if nobody protests, it probably means that it is OK to remove this syntax, doesn't it?)

I have mixed feelings. In some sense, `Theorem with` is often deceiving and thus not to recommend but the reason is that it is difficult to control the guard in interactive mode and one easily arrives to a proof which we believe is guarded but which, e.g. due to some rewriting in the way, fails to be recognized as guarded.

In itself, I don't see it as a problem of `Theorem with`, but as a weakness of the guard checker. In particular, interactive development of `Fixpoint` suffers also of this weakness.

About the unification between `Definition` and `Theorem`, maybe the young generation is so much fed of proofs-as-programs correspondence that they feel that it is an error not to unify the syntax of `Definition` and `Theorem`. For me, they remain different and, in particular, I don't see `Theorem with` as bad, but maybe I'm too old in my way of thinking :smile:

xavierleroy:
First, I think changes to the vernacular language, especially removal of constructs that are actively used in one of your flagship projects, need to be justified better than just "I find it confusing", need to be discussed widely (e.g. on coq-dev), and need consensus beyond @Zimmi48 and @herbelin.

Second, for a very long time Coq sucked at proofs by mutual induction.  The `Scheme...with` dance is awful, `Combined Scheme` (introduced by @mattam82 after I complained) is only slightly better.  `Theorem...with` is perhaps a bit temperamental but opens new, better ways to conduct those proofs by mutual induction.  This should be encouraged and developed further rather than removed.

Third, I did not know that `Fixpoint ... with` can be used in conjunction with interactive proof mode.  That may be obvious because "it's just like `Definition`", but it is not intuitive to me so I would not have guessed.  @mattam82: did you know about this when you introduced the `Theorem...with` syntax?

Fourth, I agree with @herbelin that for presentation purposes it helps to write `Theorem` for things that are theorems and `Definition` or `Fixpoint` for things that are definitions.  I am familiar with Curry-Howard, thank you very much, but the people who read and review the CompCert development (e.g. certification authorities) may not, and I see no reason to confuse them further.

So, in summary: you need to build a stronger case for the removal of this feature and offer better alternatives than `Fixpoint...with` in proof mode.

Zimmi48:
Thank you very much for this reply. No need to get anxious: if I pinged you three times to get your opinion, it's because I thought I needed it before proceeding. I'll close this for now since the goal was to get it in 8.9 but longer term, would an alternative like `Inductive Theorem ... with` (which would also have the advantage of supporting `Inductive Theorem` for non-mutual cases; possibly with the attribute syntax for the `Inductive` part), be a good solution to you?

PR #8274, opened 2018-08-18
Restore previous printing filtering; always reprint the context after "Set Diffs"

jfehrle:
**Kind:** bug fix / feature.

Bug fix: restore previous printing behavior that was unintentionally changed in 7d2a9df
(current code always prints context, should print only if the proof has changed).

Bug fix: fix message that came out as "Error: Error: -diffs requires ..."

Enhancement: always print the context after the "Set Diffs" command.

### DIFF
--- a/toplevel/coqloop.ml
+++ b/toplevel/coqloop.ml
@@ -330,17 +330,33 @@ let cproof p1 p2 =
 
 let drop_last_doc = ref None
 
+(* todo: could add other Set/Unset commands, such as "Printing Universes" *)
+let print_anyway_opts = [
+  [ "Diffs" ];
+  ]
+
+let print_anyway c =
+  let open Vernacexpr in
+  match c with
+  | VernacExpr (_, VernacSetOption (_, opt, _))
+  | VernacExpr (_, VernacUnsetOption (_, opt)) ->
+    List.mem opt print_anyway_opts
+  | _ -> false
+
 (* We try to behave better when goal printing raises an exception
    [usually Ctrl-C]
 
    This is mostly a hack as we should protect printing in a more
    generic way, but that'll do for now *)
-let top_goal_print oldp newp =
+let top_goal_print ~doc c oldp newp =
   try
     let proof_changed = not (Option.equal cproof oldp newp) in
-    let print_goals = not !Flags.quiet &&
-                      proof_changed && Proof_global.there_are_pending_proofs () in
-    if print_goals then Printer.print_and_diff oldp newp;
+    let print_goals = proof_changed && Proof_global.there_are_pending_proofs () ||
+                      print_anyway c in
+    if not !Flags.quiet && print_goals then begin
+      let dproof = Stm.get_prev_proof ~doc (Stm.get_current_state ~doc) in
+      Printer.print_and_diff dproof newp
+    end
   with
   | exn ->
     let (e, info) = CErrors.push exn in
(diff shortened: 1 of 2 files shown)


jfehrle:
Should I also reprint the context after other Set/Unset commands that affect printing, such as "Set Printing Universes"?

jfehrle:
@ejgallego Can you take a look at this sometime soon?  Thanks.

toplevel/coqloop.ml
```diff
@@ -330,17 +330,33 @@ let cproof p1 p2 =
...
-let top_goal_print oldp newp =
+let top_goal_print ~doc c oldp newp =
   try
     let proof_changed = not (Option.equal cproof oldp newp) in
-    let print_goals = not !Flags.quiet &&
-                      proof_changed && Proof_global.there_are_pending_proofs () in
-    if print_goals then Printer.print_and_diff oldp newp;
+    let print_goals = proof_changed && Proof_global.there_are_pending_proofs () ||
+                      print_anyway c in
+    if not !Flags.quiet && print_goals then begin
```
ejgallego: Why has the `quiet` flag been moved out of the main condition?

ejgallego:
I am not sure I understand what the problem is here; if the proof has not changed why is the command analysis necessary?

toplevel/coqloop.ml
jfehrle: It seemed clearer to me this way.  NP to change.

jfehrle:
> if the proof has not changed why is the command analysis necessary?

If you enter a `Set Diffs` command, the proof is not changed, but you probably want to see the context using the new setting.  Current code doesn't reprint the context.  (This may also be worth doing for other commands that change the appearance of the context, such as `Set Printing Universes`.)

toplevel/coqloop.ml
ejgallego: I'm OK with the current version.

ejgallego:
> If you enter a Set Diffs command, the proof is not changed, but you probably want to see the context using the new setting. Current code doesn't reprint the context. (This may also be worth doing for other commands that change the appearance of the context, such as Set Printing Universes.)

I see, there is indeed a paradigm conflict wrt the terminal-based model and the more graphical, modern style of interaction advocated by UIs. In the latter, UIs control when the goals will be shown, and in fact usually get a rich representation from the server [Coq] and can handle the different display options locally.

So for example, the way this would work with diffs is that CoqIDE / Emacs actually issue a call `(ShowGoals ... (diff_with foo_stm))` and indeed, there is no Coq option to be set per-se. @jfehrle you may want to have a look to the SerAPI display protocol, which while incomplete, hints some of the ideas we'd like UI handling to go.

ejgallego (approved):
LGTM, will merge soon.

jfehrle:
> you may want to have a look to the SerAPI display protocol

I'll try to do that at some point.  I think I might start by considering what various use cases in the GUI, then design the protocol--maybe you're doing that or you've thought about/used the tools enough that the GUI is obvious.

Examples (off the top of my head, probably several use cases): Users might like to efficiently replay the steps of their proof without having to (slowly) reprove them.  They might want to see some statistical info on their proof--how many steps, which goals took the most steps to prove, which lemmas were used.  They might want to label certain steps so they can jump to them for personal review or to show others.  They might want to annotate some steps, then be able to see those notes later.  They might want the ability to do branching--alternative versions and the ability to switch between them.

... then I'd figure out the API.

jfehrle:
Thanks for the review and merge!

PR #8278, opened 2018-08-20
Do not inline let-bound functions in clambda optimization.

### LINKED ISSUE #8277: Out of memory on reasonably sized proof.
#### Version

The Coq Proof Assistant, version 8.8.1 (August 2018)
compiled on Aug 4 2018 12:29:40 with OCaml 4.07.0

#### Operating system

Linux

Coq raises an Out of Memory on my machine (about 8G of RAM) when checking a proof term of "reasonnable size", whereas an equivalent proof script type-checks in less than a second. Both file are available on [this gist](https://gist.github.com/Gbury/6d42a203e44d73be8cacc4f15e489593) (as term.v and script.v respectively).

For the context: this proof was automatically generated by a theorem prover, which generated both the script (`script.v` in the gist), and the proof term (`term.v` in the gist), in about 0.1s. The term in `term.v` should be relatively identical to the term built by the script. Note that essentially the same proof term was generated for the [dedukti logical framework](https://github.com/Deducteam/Dedukti) (modulo some encoding), which had no problem checking it instantly.

Maybe this is related to the structure of the generated proof (using a lot of let bindings). If so, I'm open to suggestions on how to produce better proofs.

Finally, this proof was generated from a series of problem of increasing size (from 1 to 100), and the example in the gist was produced using the the smallest size that Coq did not manage to verify on my machine (which is size 6). Thus, if needed, bigger (and smaller) instances of the same kind of proof can easily be generated (up to about size 50).

ppedrot:
This was triggering an exponential blowup in the size of the generated
intermediate VM code.

Fixes #8277.

### DIFF
--- a/kernel/clambda.ml
+++ b/kernel/clambda.ml
@@ -269,7 +269,7 @@ let lam_subst_args subst args =
 let can_subst lam =
   match lam with
   | Lrel _ | Lvar _ | Lconst _
-  | Lval _ | Lsort _ | Lind _ | Llam _ -> true
+  | Lval _ | Lsort _ | Lind _ -> true
   | _ -> false
 
 let rec simplify subst lam =


Zimmi48:
@maximedenes Would you have some time to review and merge this?

maximedenes:
@bgregoir does it look good to you?

@ppedrot does `nativelambda` need the same fix?

maximedenes (approved):
After discussing with @bgregoir, it appears that anyway this could have been useful only for iterators, which are not there yet, and will probably be compiled in a way that does not rely on this heuristics.

So let's merge!

Zimmi48:
@maximedenes ping
VST failure is probably unrelated. I don't know if you want to make sure there's no issue with a last rebase.

PR #8282, opened 2018-08-21
[coq_makefile] print all options (Fix #7529)

### LINKED ISSUE #7529: Help option is not recognized by CoqMakefile
#### Version

8.8

#### Operating system

N/A

The man page of `coq_makefile` says:

```
-h     Will give you a description of the whole list of options of coq_makefile.
```

However, typing `coq_makefile -h` produces:

```
Unknown option -h
Available arguments:
[file.v]: Coq file to be compiled
[file.ml[i4]?]: Objective Caml file to be compiled
[file.ml{lib,pack}]: ocamlbuild file that describes a Objective Caml
  library/module
[any] : subdirectory that should be "made" and has a Makefile itself
  to do so. Very fragile and discouraged.
[-extra result dependencies command]: add target "result" with command
  "command" and dependencies "dependencies". If "result" is not
  generic (do not contains a %), "result" is built by _make all_ and
  deleted by _make clean_.
[-extra-phony result dependencies command]: add a PHONY target "result"
 with command "command" and dependencies "dependencies". Note that
 _-extra-phony foo bar ""_ is a regular way to add the target "bar" as
 as a dependencies of an already defined target "foo".
[-I dir]: look for Objective Caml dependencies in "dir"
[-R physicalpath logicalpath]: look for Coq dependencies resursively
  starting from "physicalpath". The logical path associated to the
  physical path is "logicalpath".
[-Q physicalpath logicalpath]: look for Coq dependencies starting from
  "physicalpath". The logical path associated to the physical path
  is "logicalpath".
[VARIABLE = value]: Add the variable definition "VARIABLE=value"
[-byte]: compile with byte-code version of coq
[-opt]: compile with native-code version of coq
[-arg opt]: send option "opt" to coqc
[-install opt]: where opt is "user" to force install into user directory,
  "none" to build a makefile with no install target or
  "global" to force install in $COQLIB directory
```

gares:


gares:
I asked the review by @Zimmi48 for two reasons. To back up @silene and to notify him that this seems to be a regression (see the bugreport)

Zimmi48 (approved):
Indeed. This is backportable IMO.

Issue #5038, opened 2016-08-24
Computation of the head of a term unfolds let-bindings


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#5038
From: @jadephilipoom
Reported version: 8.5
CC:   @andres-erbsen, @silene, @JasonGross, @ppedrot, @psteckler

ppedrot:
AFAICT, this has been solved by #9915.

PR #8285, opened 2018-08-21
Fixing #8270: cbn was applying zeta even when not asked for.

### LINKED ISSUE #8270: cbn without zeta substitutes let-bindings
Version 8.8.1.

`cbn`, when called without the zeta reduction flag, performs a sort of intermediate reduction on let-bindings where it substitutes the variable in the body of the let, without eliminating the let-binding.  This is a bug because without zeta, `cbn` shouldn't change the lets at all.  This bug makes `cbn` unsuitable for performing reduction when using let-bindings to avoid an exponential blowup in term size, which is inconvenient when one wants to simplify applications of fixpoints to such expressions.

Related to #5327 along the general theme that it's inconvenient to perform reduction in Coq without unfolding let-bindings.

```
Goal let x := 0 in
     let y := x in
     y = 0.
  (* We use "cofix" as an example because there are obviously no
     cofixpoints in sight.  This problem arises with any set of
     reduction flags (not including zeta where the lets are of course reduced away) *)
  cbv cofix.
  (* as expected, the goal is unchanged. *)
  Set Debug RAKAM.
  cbn cofix.
  (* cbn substitutes the bodies of the let-expressions without
     eliminating them.  This is different from zeta-reduction (which
     would eliminate the let-bindings entirely) and is a bug; cbn ought
     to leave the goal unchanged just like cbv. *)
```

[Output from Debug RAKAM](https://gist.github.com/ichung/0e2f84912ca961927a9a8108656ed1cf)

herbelin:
**Kind:** bug fix

Fixes / closes #8270

Using the same fix as in `cbv`: when `zeta` is not set, we discard the body of the let-in when pushing the declaration on the environment.

### DIFF
--- a/pretyping/reductionops.ml
+++ b/pretyping/reductionops.ml
@@ -628,6 +628,18 @@ let safe_meta_value sigma ev =
   try Some (Evd.meta_value sigma ev)
   with Not_found -> None
 
+let strong_with_flags whdfun flags env sigma t =
+  let push_rel_check_zeta d env =
+    let open CClosure.RedFlags in
+    let d = match d with
+      | LocalDef (na,c,t) when not (red_set flags fZETA) -> LocalAssum (na,t)
+      | d -> d in
+    push_rel d env in
+  let rec strongrec env t =
+    map_constr_with_full_binders sigma
+      push_rel_check_zeta strongrec env (whdfun flags env sigma t) in
+  strongrec env t
+
 let strong whdfun env sigma t =
   let rec strongrec env t =
     map_constr_with_full_binders sigma push_rel strongrec env (whdfun env sigma t) in
--- a/pretyping/reductionops.mli
+++ b/pretyping/reductionops.mli
@@ -144,6 +144,9 @@ val pr_state : state -> Pp.t
 
 (** {6 Reduction Function Operators } *)
 
+val strong_with_flags :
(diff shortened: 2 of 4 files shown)


ppedrot (approved):
LGTM. Aren't there other instances of the problem in the code base?

herbelin:
> Aren't there other instances of the problem in the code base?

I indeed remember having found more than one inconsistency between `lazy`, `cbv` and `cbn` with respect to zeta and/or delta of local variables.

Searching in the issues, I found [this comment](https://github.com/coq/coq/issues/5038#issuecomment-337546607) of #5038 where I'm defending that `cbn` is right and that it is `cbv`  and `lazy` which are wrong. So, it is really about what is the intended meaning of strong reduction (should strong reduction consider that it re-apply weak-head reduction in the context with the binders it traversed - as claimed in #5038 -, or should strong reduction consider bindings are being part not of the context but of the term being strongly reduced as implicitly suggested in #8270).

So, I don't know what is the smarter semantics to implement, but, for sure, a decision must be taken...

herbelin:
Rebased.

ppedrot:
@herbelin The linter is complaining about the trailing newline in the test file, can you fix it?

herbelin:
Ah, sorry; linter fixed.

PR #8286, opened 2018-08-21
Fixing #7867: class error message tried to print a "fun" with no binders

### LINKED ISSUE #7867: Anomaly "File "printing/ppconstr.ml", line 398, characters 14-20: Assertion failed."
#### Version

8.8.0 and master

```
Class class := { foo : nat }.
Instance : class := { foo := 0 ; bar := 0 }.
```

gives:

```
Error: Anomaly "File "printing/ppconstr.ml", line 398, characters 14-20: Assertion failed."
```
The line number is for current master, i.e. this is:

https://github.com/coq/coq/blob/6715e6801c1d285a12eeca55dd8b831d7efb8c0d/printing/ppconstr.ml#L398

herbelin:
**Kind:** bug fix

Fixes / closes #7867.

Note: the printer cannot print a `fun` or a `forall` with no binders. It is not exactly clear who should be responsible of ensuring that an internal `CLamdaN` or internal `CProdN` node has at least one binder. We fix the bug by modifying the parser so that it does not generate such degenerate `CLamdaN` or `CProdN` node.

### DIFF
--- a/parsing/g_constr.mlg
+++ b/parsing/g_constr.mlg
@@ -249,7 +249,7 @@ GRAMMAR EXTEND Gram
 
   record_field_declaration:
     [ [ id = global; bl = binders; ":="; c = lconstr ->
-      { (id, mkCLambdaN ~loc bl c) } ] ]
+      { (id, if bl = [] then c else mkCLambdaN ~loc bl c) } ] ]
   ;
   binder_constr:
     [ [ "forall"; bl = open_binders; ","; c = operconstr LEVEL "200" ->
--- a/test-suite/bugs/closed/7867.v
+++ b/test-suite/bugs/closed/7867.v
@@ -1,0 +1,4 @@
+(* Was a printer anomaly due to an internal lambda with no binders *)
+
+Class class := { foo : nat }.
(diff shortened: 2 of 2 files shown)


herbelin:
Hi @ejgallego, would you have time to review this trivial PR?

ejgallego:
> Hi @ejgallego, would you have time to review this trivial PR?

Sure, but please, next time add a request for review earlier, otherwise I may not notice I should take over.

ejgallego (dismissed):
Umm, this seems like a large hammer to pull [modifying the core AST functions] in order to fix #7867 which is indeed due to incomplete code in the instance declaration. On the other hand I can understand that this default behavior can simplify a few programming cases.

I am not very happy to see the fix in this form, especially due to the bad error message which is giving us a hint. IMHO we can:
- keep the code as it, however document the functions and the special pattern/case for the empty list.
- `assert false` if the binder list is empty, instead of silently returning the constructor, so the invariant won't be broken by users.
- keep this code as is before the PR and modify the instance code with uses smarter code for record display.

ejgallego (approved):
LGTM, thanks!

PR #8289, opened 2018-08-21
Cachix can now be installed from nixpkgs.

Zimmi48:
So we use the same version of nixpkgs to install Cachix and the dependencies of Coq.


SkySkimmer (approved):

Zimmi48:
I realized (thanks to CI) that I depended on a file that was not checked out (since the git strategy is set to none is this job). So know I `wget` the `pin.nix` file before running the command that uses it.

Zimmi48:
I'm stupid. `wget` isn't available either. Well, I don't know how to do what I wanted except by duplicating the pin info in `.gitlab-ci.yml`.

SkySkimmer:
What's wrong with git_strategy: fetch?

Zimmi48:
It would require `git` in the base image.

Zimmi48:
I can reopen if ever @vbgl or someone else has an idea.

PR #8290, opened 2018-08-21
Fix #8288: cumulativity inferance ignores args to bound variables

### LINKED ISSUE #8288: Anomaly when using Polymorphic Inductive Cumulativity
#### Version
8.8.0

#### Operating system
Ubuntu 18.04.1

The code in the gist below raises an anomaly `Invalid subtyping relation`.  It is using `Set Polymorphic Inductive Cumulativity`. If I remove that line, it works. BTW, most likely we shouldn't use inductive cumulativity in this code.

https://gist.github.com/beta-ziliani/2779d3a9ede67fc911c78a24890598c2

SkySkimmer:
(NB: only variables whose binder is inside the constructor argument,
ie

        Inductive foo := C : forall A : Type -> Type, A Type -> foo.

does not trigger the bug because A becomes a RelKey.)

Who knows how long this would have been undetected (and it would have been a proof of False) if infercumulativity was in the kernel.

### DIFF
--- a/pretyping/inferCumulativity.ml
+++ b/pretyping/inferCumulativity.ml
@@ -99,7 +99,7 @@ let rec infer_fterm cv_pb infos variances hd stk =
   | FEvar ((_,args),e) ->
     let variances = infer_stack infos variances stk in
     infer_vect infos variances (Array.map (mk_clos e) args)
-  | FRel _ -> variances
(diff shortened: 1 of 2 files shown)


Zimmi48:
@mattam82 ping

mattam82 (approved):
Right, clearly missing case.

PR #8293, opened 2018-08-22
Fix typo of caracterisation -> c*h*aracterisation

bollu:
**Kind:** documentation.

I noticed the typo, so I grepped and fixed the offending typo across all files.


herbelin:
No reason to have compcert path failure in vst related.

Issue #8091, opened 2018-07-18
Eval ... in performs universe minimization even when this is useless


ppedrot:
#### Version

Coq 8.9

#### Description of the problem

The following script displays the issue:
```
Set Universe Polymorphism.
Unset Universe Minimization ToSet.

Inductive prod@{i j} (A : Type@{i}) (B : Type@{j}) := pair : A -> B -> prod A B.

Definition fst@{i j} {A B} (p : prod@{i j} A B) := match p with pair _ _ x _ => x end.

Ltac generate n :=
match n with
| 0 => constr:(Type)
| S ?n =>
  let r := generate n in
  constr:(pair _ _ Type r)
end.

Definition foo := ltac:(let n := generate 1000 in exact n).

Time Eval lazy in fst foo.
```
The last command takes about one second on my machine even though the reduction in itself is instantaneaous. The reason is that Coq is performing universe minimization just for the sake of it.

SkySkimmer:
Isn't it rather that Eval should print the universe context like Check does?

andres-erbsen:
Why does universe minimization with 1000 universes take an entire second to begin with? Perhaps it would be good to make a test case that shows that time even after this issue is fixed?

SkySkimmer:
`generate n` produces something like `4 * n + 2` universes
Whether that makes it more reasonable is up to you.

PR #8295, opened 2018-08-22
Fix #8291: print universe names in universe context for Check.

### LINKED ISSUE #8291: Check doesn't support nonstrict universe declaration (in printing)
~~~coq
Unset Strict Universe Declaration.
Set Printing Universes.
Check Type@{i}.
(* Type@{i}
     : Type@{i+1}
(* {Top.13} |=  *)
*)
~~~

SkySkimmer:
Why were we avoiding the shadowing of the old `sigma`?

### DIFF
--- a/test-suite/output/UnivBinders.out
+++ b/test-suite/output/UnivBinders.out
@@ -48,6 +48,9 @@ Type@{Top.17} -> Type@{v} -> Type@{u}
 (* u Top.17 v |=  *)
 
 foo is universe polymorphic
+Type@{i} -> Type@{j}
+     : Type@{max(i+1,j+1)}
+(* {j i} |=  *)
 Monomorphic mono = Type@{mono.u}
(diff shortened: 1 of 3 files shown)


SkySkimmer:
Added the print for Eval for #8091 since I don't want to deal with a conflict for this trivial stuff.

Zimmi48:
@mattam82 ping

mattam82 (approved):
It's not clear to me why we avoided the shadowing before, maybe @herbelin has an idea. The code looks simpler and cleaner with your changes.

Zimmi48:
@mattam82 Don't forget to self-assign when you are the only code owner.

herbelin:
@mattam82: it is in your code (0132b5b51fc), so I guess this was a typo (actually favored by me not shadowing for no particular good reason). @skyskimmer's fix is then good :+1:.

Issue #8298, opened 2018-08-22
OCaml 4.07 download path is incorrect (non standard github archive path which sometimes doesn't work)


MSoegtropIMC:
The download path for OCaml 4.07 in the windows build scripts is a non standard github archive path. It depends on the day, the weather and the internet client if the download with this path works or not.
Replace this with a standard gihub archive path.

Zimmi48:
FTR, the non-standard path is https://github.com/ocaml/ocaml/archive/4.07.0/ocaml-4.07.0.tar.gz. From what you said in your PR, it works only in Firefox, but not with wget, Chrome or IE. I couldn't reproduce this on Linux (it worked for me in Chromium, Firefox, and with `wget`). So I suppose indeed, it depends on the day as well.

Zimmi48:
PR merged

MSoegtropIMC:
My guess is that it depends on which specific github frontend server the requests ends up at, Likely they have slightly different software versions which lead to different results.

What I learned from this is that testing an URL in a browser is not enough, one also needs to look at it.

PR #8300, opened 2018-08-22
Fix issue #8298 OCaml 4.07 download path is incorrect

MSoegtropIMC:
This fixes issue #8298.

The current download path looks like a kind of typo. Interestingly it does work sometimes (I checked it in firefox when reviewing the commit). For me the current path works only in Firefox, but not with wget, Chrome or IE. For CI @jfehrle reported that it fails sometimes. The new path is the github standard archive path and works in any browser.

Cheers,

Michael


Zimmi48 (approved):
Yes, it was just a misunderstanding from me of the `build_prep` function.

PR #8281, opened 2018-08-21
Trivial Sphinx fix in  doc.

Zimmi48:
**Kind:** documentation fix

Problem noticed by @jfehrle in #8279.

PR #8266, opened 2018-08-17
Minor Sphinx improvements in the bullet documentation.

Zimmi48:
And fixing a problem with nested proofs.
The second commit makes `bullet` an actual grammar token. @cpitclaudel There are many such issues in the manual. Would there be a way to get a warning to detect them?

**Kind:** documentation

PR #8301, opened 2018-08-23
Documentation for proof diffs

jfehrle:
**Kind:** documentation

This is written assuming that #8247 makes it into 8.9, so this should be merged after that.

I hope the reviewer can take a look at the output.

Surprisingly, `make sphinx` is running efficiently now--it runs in less than a minute.  Odd.  I wonder if the file staleness check might be comparing dates without times.


doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+
+.. _diffs_coqide_removed:
+
+  .. image:: ../_static/diffs-coqide-removed.png
+     :alt: |CoqIDE| with Set Diffs removed
+
+**How to enable diffs**
+
+.. opt:: Set Diffs %( "on" %| "off" %| "removed" %)
+.. This ref doesn't work: :opt:`Set Diffs %( "on" %| "off" %| "removed" %)`
```
jfehrle: Issues with references:
- While the parameters (e.g. "on|"off"|"removed" should appear in the `opt ::` line, that seems to become part of the tag name.  I wanted line 581 to be a link to the option definition but couldn't get it to work with or without the parameters.
- There should be square brackets around the options instead of parentheses but I couldn't get that to work, either.  Known issue?

Zimmi48:
Rendered:
- https://coq.gitlab.io/-/coq/-/jobs/91412685/artifacts/_install_ci/share/doc/coq/sphinx/html/practical-tools/coq-commands.html#by-command-line-options
- https://coq.gitlab.io/-/coq/-/jobs/91412685/artifacts/_install_ci/share/doc/coq/sphinx/html/proof-engine/proof-handling.html#showing-differences-between-proof-steps

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
 
+.. _showing_diffs:
+
+Showing differences between proof steps
+---------------------------------------
+
+
+Coq can automatically highlight the differences between successive proof steps.
+For example, the following screen shots of coqtop and CoqIDE show the same
+“intros” tactic.  This tactic creates two new hypotheses, highlighted in green.
```
Zimmi48: `“intros”` -> ```:tacn:`intros` ```

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+The conclusion is entirely in pale green because it’s changed, but no tokens were added
+to it.  The third screen shot uses the "removed" option, so it shows the conclusion a
+second time with the old text, with deletions marked in red.  Also, since the hypotheses are
+new, no line of old text is shown for them.
+
+.. _diffs_coqtop_on:
+
+  .. image:: ../_static/diffs-coqtop-on.png
```
Zimmi48: I would crop the top of this image which is confusing.

doc/sphinx/proof-engine/proof-handling.rst
Zimmi48: Keep it with parentheses (like for https://coq.gitlab.io/-/coq/-/jobs/91412685/artifacts/_install_ci/share/doc/coq/sphinx/html/proof-engine/proof-handling.html#coq:opt.bullet-behavior-none-strict-subproofs).
It's very weird that the option doesn't appear in the index. @cpitclaudel Any insight?

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+
+  .. image:: ../_static/diffs-coqide-removed.png
+     :alt: |CoqIDE| with Set Diffs removed
+
+**How to enable diffs**
+
+.. opt:: Set Diffs %( "on" %| "off" %| "removed" %)
+.. This ref doesn't work: :opt:`Set Diffs %( "on" %| "off" %| "removed" %)`
+
+The “on” option highlights added tokens in green, while the “removed” option
```
Zimmi48: Indent this paragraph.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+
+2. For each goal in the new proof state, determine what old goal to compare
+   it to--the one it is derived from or is the same as.  Match the hypotheses by
+   name (order is ignored), handling compacted items specially.
+
+3. For each hypothesis and conclusion (the “items”) in each goal, pass
+   them as strings to the lexer to break them into tokens.  Then apply the standard
+   Myers diff algorithm to the tokens and add appropriate highlighting.
+
+Notes:
```
Zimmi48: Use `.. note::`

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+
+
+In CoqIDE, diffs should be enabled from the View menu.  Don’t use the ``Set Diffs``
+command in CoqIDE.  You can change the background colors shown for diffs from the
+Edit | Preferences | Tags panel by changing the settings for the ``diff.added``,
+``diff.added.bg``, ``diff.removed`` and ``diff.removed.bg`` tags.  This panel also
+allows let you control other attributes of the highlights, such as the foreground
+color, bold, italic, underline and strikeout.
+
+Note: As of this writing (August 2018), Proof General will need minor changes
```
Zimmi48: Use `.. note::`
BTW, is there an open issue on PG issue tracker? You could point to it.

Zimmi48 (commented):
Partial comments (I didn't read the doc fully yet).

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+The “on” option highlights added tokens in green, while the “removed” option
+additionally reprints items with removed tokens in red.  Unchanged tokens in
+modified items are shown with pale green or red.  (Colors are user-configurable.)
+
+For coqtop, showing diffs can be enabled when starting coqtop with the
+``-diffs on|off|removed`` command-line option or with the ``Set Diffs``
+command within Coq.  You will need to provide the ``-color on|auto`` command-line option when
+you start coqtop in either case.
+
+Colors for coqtop can be configured by setting the ``COQ_COLORS`` environment
```
Zimmi48: This is mostly repetitive with the existing documentation of this environment variable. Can't you rather add a reference to it and improve / expand this documentation (use `.. seealso::`).

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+color (0, 0, 240) and unmodified text in a modified item to a blue background.
+The other tags used by diffs are ``diff.removed`` and ``diff.removed.bg``.  Note
+that if you specify ``COQ_COLORS``, the predefined attributes are ignored.
+
+
+In CoqIDE, diffs should be enabled from the View menu.  Don’t use the ``Set Diffs``
+command in CoqIDE.  You can change the background colors shown for diffs from the
+Edit | Preferences | Tags panel by changing the settings for the ``diff.added``,
+``diff.added.bg``, ``diff.removed`` and ``diff.removed.bg`` tags.  This panel also
+allows let you control other attributes of the highlights, such as the foreground
```
Zimmi48: Typo: remove "allows"

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+**How diffs are calculated**
+
+Diffs are calculated as follows:
+
+1. Select the old proof state to compare to, which is the proof state before
+   the last tactic that changed the proof.  Changes that only affect the view
+   of the proof, such as ``all: swap 1 2``, are ignored.
+
+2. For each goal in the new proof state, determine what old goal to compare
+   it to--the one it is derived from or is the same as.  Match the hypotheses by
```
Zimmi48: did you mean to use — instead?

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+   name (order is ignored), handling compacted items specially.
+
+3. For each hypothesis and conclusion (the “items”) in each goal, pass
+   them as strings to the lexer to break them into tokens.  Then apply the standard
+   Myers diff algorithm to the tokens and add appropriate highlighting.
```
Zimmi48: If it's standard it probably has a Wikipedia page that you could maybe link to.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+
+Notes:
+
+* Aside from the highlights, the "on" option output should be identical
+  to the undiffed output.
+* Deleted goals (i.e. completed in the last proof step) will not be shown
```
Zimmi48: Don't call this "deleted goals", "completed goals" seems clear enough (and honestly, this point seems like stating the obvious).

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+  even with the "removed" setting.
+* If you get a ``Diff_Failure`` message, that may be a bug.  While unlikely,
```
Zimmi48: I'm not sure what you are saying here. If it is an anomaly, then it *is* most definitely a bug and should be reported as such. I'm not sure about the worthiness of mentioning the `Back` command since nobody uses this AFAICT.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+  it’s possible that the string generated for some item can’t be parsed by the
+  lexer.  You may also see this if you use the :cmd:`Back` command, in which case the
+  diff code won’t get the right old version of the proof.  In contrast, :cmd:`Undo` and
+  buttons for going backward in the proof in CoqIDE will work fine.
+
+The following screen shots show diffs working with multiple goals and with compacted
```
Zimmi48: Isn't the spelling screenshot more common?

doc/sphinx/proof-engine/proof-handling.rst
cpitclaudel: No particular idea

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+command in CoqIDE.  You can change the background colors shown for diffs from the
+Edit | Preferences | Tags panel by changing the settings for the ``diff.added``,
+``diff.added.bg``, ``diff.removed`` and ``diff.removed.bg`` tags.  This panel also
+allows let you control other attributes of the highlights, such as the foreground
+color, bold, italic, underline and strikeout.
+
+Note: As of this writing (August 2018), Proof General will need minor changes
+to be able to show diffs correctly.  We hope it will support this feature soon.
+
+**How diffs are calculated**
```
cpitclaudel: This should be a title, rather than bold text, I think

doc/sphinx/proof-engine/proof-handling.rst
cpitclaudel: Right, it would be nice to know which minor adjustments are needed.  Even better, it would be nice to implement the feature in PG before merging this, and then we don't even need to document the issue.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +539,112 @@ Requesting information
...
+The following screen shots show diffs working with multiple goals and with compacted
+hypotheses.  In the first one, notice that the goal ``P 1`` is not highlighted at
+all after the split because it has not changed.
+
+.. _diffs_coqtop_multigoal:
+
+  .. image:: ../_static/diffs-coqtop-multigoal.png
+     :alt: coqtop with Set Diffs on with multiple goals
+
+.. _diffs_coqtip_compacted:
```
cpitclaudel: Probably should be `coqtop` instead of `coqtip`?

cpitclaudel:
I'm not sure it's worth showing examples of the way things look in coqtop — does anyone actually use the plain coqtop?

Conversely, I think the `coqtop` blocks are processed using coqtop, so showing the examples directly in the manual might be better than screenshots.

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: I thought I should show the prior state state (the `forall` line) for context.  I'll include the `forall` as regular text.

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: You're sure?  There are 3 notes together here.  I think they're all minor points that don't deserve an attention-grabbing box around them.

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: I'll open an issue in PG and reference it.  I was told there are 2 versions: a general PG and a Coq-specific PG.  Is https://github.com/ProofGeneral/PG the correct project?

I assumed I'd get better attention from the PG folks after the feature is officially released.  I've not interacted with the PG folks (except a little with Erik Dorel-Martin, who I met in Nice) so I don't know if or when they would make changes.  Suggestions welcome.

> Right, it would be nice to know which minor adjustments are needed.

Presumably in the PR, not here.

> it would be nice to implement the feature in PG before merging this

You make it sound so easy.  I've never even used PG, I'm not an emacs user, haven't done emacs lisp and I don't know how the PG team operates.  Maybe none of these will present any difficulties but I don't know yet.  Still busy making sure the last few changes get into 8.9.

doc/sphinx/proof-engine/proof-handling.rst
Zimmi48: I think what `intros` does is sufficiently clear to everyone that you don't need to give the context.

doc/sphinx/proof-engine/proof-handling.rst
Zimmi48: We're just trying to be consistent around the manual.

doc/sphinx/proof-engine/proof-handling.rst
Zimmi48: There is only one version and it is Coq-specific and it is the right URL you liked to.

>I assumed I'd get better attention from the PG folks after the feature is officially released.

That's a bad idea: we generally try to have PG support land as early as possible. Note that @cpitclaudel knows what he's talking about since he's also a PG maintainer.

>You make it sound so easy. I've never even used PG, I'm not an emacs user, haven't done emacs lisp and I don't know how the PG team operates. Maybe none of these will present any difficulties but I don't know yet. Still busy making sure the last few changes get into 8.9.

But do you know for sure that minor adjustments are needed? I didn't try it yet.

jfehrle: Well, I spoke to Erik about supporting diffs in PG when I was in Nice, then I emailed him once the diff code was in master.  I didn't hear back from him.  I would be delighted if one of the PG maintainers would make the changes.

Adjustments needed (I'll put more detail into the PR):
1. Define default color/highlight for tags diff.removed, diff.added, diff.removed.bg, diff.added.bg
2. Semantic extension: tags in the form "start.*" indicate the highlight continues until a matching "end.*" tag is seen.  But turn off such highlighting for white space at the beginning and end of lines.

jfehrle: PS PG will work just fine if diffs are not enabled.  It's only if diffs are enabled.

jfehrle: Looks pretty good to me with -emacs mode.  @ejgallego advised me to change Idetop.goals, which IIUC is for the XML interface, which I also did.  If PG simply accepts ANSI escape codes embedded in the output, then it's probably really simple.  Can PG/emacs handle underlining (or strikeout?).  I would expect it can handle background colors with 24-bit rgb colors.  If it can handle strikeout, I may want to make that available in coqtop--I had read that most terminal software doesn't support strikeout.

Also, the colors I chose may not be great defaults in emacs.  Is there a good way to provide different defaults from PG?  Or is color selection usually done just by setting COQ_COLORS?

Is there already clean way to pass `-diffs "on"` from PG so the user doesn't have to do that every time they start PG (if they want that setting)?  Or is that a small change?

Perhaps you can try the code in master with PG with -emacs and let me know how it looks?

image

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: OK, but I think it would be a `:see`, right?

doc/sphinx/proof-engine/proof-handling.rst
cpitclaudel: >  If PG simply accepts ANSI escape codes embedded in the output, then it's probably really simple.

We don't currently do, AFAIK, but it could be done.

> Can PG/emacs handle underlining (or strikeout?).

I expect so.

>  I would expect it can handle background colors with 24-bit rgb colors.

Not on typical terminals (these tend to support only 256 colors), though the latest versions of Emacs can be configured to output 24-bit color codes on appropriately configured terminals.  
Of course, the GUI Emacs doesn't have trouble with 24-bit color.

> Also, the colors I chose may not be great defaults in emacs. Is there a good way to provide different defaults from PG? Or is color selection usually done just by setting COQ_COLORS?

We currently ignore the colors produced by coqtop (or, more precisely, colors are currently disabled with -emacs).  You mentioned 24-bit colors; do you actually make use of those, or do you use the more typical 256-color terminal palette?

Can you show what the raw output (non-highlighted) looks like?

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: Yes.  I expected Sphinx would convert it.  Apparently not, which awkward since there's no key on the keyboard for an em dash.

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: I can remove "standard".  There's no wikipedia article or other succinct description.  Myer's original paper is here: http://www.xmailserver.org/diff2.pdf.  I think people know what a diff is, just like they know what a sort is without a reference.  You think it's worth linking to the paper?

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: OK, completed goals.  My thought was: If you see a diff in github, for example, you *always* see what's been deleted.  People might expect it (if they actually use the "removed" option).

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: I didn't have time to modify the lexer so that, for diffs, it accepts all strings.  That's next on my list.  @ejgallego and I thought it was pretty unlikely that something printed in the proof would not be lexable.  I do want to fix that, though, so the underlying diff code can later be used with any input strings, such as for error messages for type errors (diffing the expected and actual) or perhaps Gallina code.  Or other things.

I'll remove this.

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: Apparently, yes.

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: It should be subordinate to the subsection "Showing differences between proof steps".  I tried making this a subsubsection but Sphinx gave an error for the input shown below.  

```
How diffs are calculated
^^^^^^^^^^^^^^^^^^^^^^^^
```

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: Yes.  I followed the pattern and defined a link for the image but there is no reference to it.  Is it necessary to define a link for every image?

jfehrle:
> I'm not sure it's worth showing examples of the way things look in coqtop

I assumed few people would use coqtop to write proofs (or want to).  The coqtop highlights are different from CoqIDE so I thought that was worth showing in at least the first of the 5 screenshots.  I used coqtop for the last two screen shots because it shows the before and after proof states in one image.

> Conversely, I think the coqtop blocks are processed using coqtop, so showing the examples directly in the manual might be better than screenshots.

If the blocks show the highlights that would be fine.  I haven't yet looked at how to show coqtop blocks.  Is there an easy way to turn on diffs on a per-example basis?  That requires a code change?  Either the command line option to `coqtop` or include `Set Diffs "on"` in the script, but I don't want `Set Diffs "on"` to show in the example.

Also, FWIW, it may be helpful to show diffs may be helpful on some other unrelated examples.

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: > Does Windows offer something like compose keys?

I'm not aware of anything like that.  MS Word converts `--` to an em dash on the fly.  I haven't needed to type em dashes in other contexts in Windows.

cpitclaudel:
> That requires a code change? Either the command line option to coqtop or include Set Diffs "on" in the script, but I don't want Set Diffs "on" to show in the example.

It might require tweaking to support highlighting in full, and we may want to add a simple way to enable diffs for a single sample, but for now just adding the following should do:

```
.. coqtop:: none
    Set Diffs On
```

doc/sphinx/proof-engine/proof-handling.rst
cpitclaudel: No, it shouldn't be

doc/sphinx/proof-engine/proof-handling.rst
cpitclaudel: Which error?

doc/sphinx/proof-engine/proof-handling.rst
cpitclaudel: Oh, are diffs based on string comparisons?

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: The coqtop colors are 24-bit.  It was tricky to get shades of green and red that worked reasonably well with a black background and without changing the existing syntactic highlights.

```
  ; "diff.added"       , make ~bg_color:(`RGB(0,141,0)) ~underline:true ()
  ; "diff.removed"     , make ~bg_color:(`RGB(170,0,0)) ~underline:true ()
  ; "diff.added.bg"    , make ~bg_color:(`RGB(0,91,0)) ()
  ; "diff.removed.bg"  , make ~bg_color:(`RGB(91,0,0)) ()
```

The CoqIDE colors are exactly the ones that GitHub uses, which I think are better than the coqtop colors.  Depends on whether PG usually has a light or dark background.  They require 24 bit color:

```
    ("diff.added", make_tag ~bg:"#b6f1c0" ~underline:true ());
    ("diff.removed", make_tag ~bg:"#f6b9c1" ~strikethrough:true ());
    ("diff.added.bg", make_tag ~bg:"#e9feee" ());
    ("diff.removed.bg", make_tag ~bg:"#fce9eb" ());
```

Here is the raw output, omitting the escape character that should appear before each `[`:

```$ rlwrap bin/coqtop.exe -diffs removed -emacs -color on
Welcome to Coq DESKTOP-MMPNVGH:/cygdrive/c/Users/Jim.DESKTOP-MMPNVGH/Documents/coqSource/coq,diffs_doc (c66c27781c555ec7301300cbf0d0342394c03981)

<prompt>Coq < 1 || 0 < </prompt>Parameter P : nat -> Prop.
<infomsg>P is declared</infomsg>

<prompt>Coq < 2 || 0 < </prompt>Goal P 1 /\ P 2.
1 subgoal (ID 1)

  ============================
  [48;2;0;91;0m[48;2;0;141;0;4mP 1[37m /\[39m P 2[48;2;0;91;0;24m

<prompt>Unnamed_thm < 3 |branch| 0 < </prompt>split.
2 subgoals (ID 3)

  ============================
  [48;2;91;0;0mP 1[37m [39m[48;2;170;0;0;4m[37m/\[39m P 2[48;2;91;0;0;24m
  [48;2;0;91;0mP 1

subgoal 2 (ID 4) is:
 P 2
```

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: Yes, diff starts with strings, which it passes through the lexer to get a list of strings/tokens that the diff treats as opaque values--["abc"] compared to ["abcd"] gives one deletion ("abc") and one insertion ("abcd").  Whitespace differences are ignored since they are not tokens.

doc/sphinx/proof-engine/proof-handling.rst
jfehrle:
```
Warning, treated as error:
/cygdrive/c/Users/Jim.DESKTOP-MMPNVGH/Documents/coqSource/coq/doc/sphinx/proof-engine/proof-handling.rst:569:Title level inconsistent:

How to enable diffs
^^^^^^^^^^^^^^^^^^^
make[1]: *** [Makefile.doc:59: sphinx] Error 2
```

jfehrle:
My output.  I think the extra blank lines are a Windows artifact.

image

jfehrle:
I made many of the updates, but not done yet.  Looking for @cpitclaudel's thoughts on getting the highlights to show.

doc/sphinx/README.rst
```diff
@@ -202,6 +202,9 @@ In addition to the objects above, the ``coqrst`` Sphinx plugin defines the follo
           Print nat.
           Definition a := 1.
 
+    The blank line after the directive is required.  If you begin a proof,
+    include an ``Abort`` afterwards to reset coqtop for the next example.
```
Zimmi48: You also need to modify the file this one is generated from (see comments at top).

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -393,6 +393,9 @@ The following example script illustrates all these features:
     { trivial. }
     assumption.
 
+  .. coqtop:: none
+
+    Abort.
```
Zimmi48: Note: this one is already fixed by #8266 and this change will create a conflict when #8266 is merged.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +546,137 @@ Requesting information
...
+
+  .. image:: ../_static/diffs-coqide-removed.png
+     :alt: |CoqIDE| with Set Diffs removed
+
+**How to enable diffs**
+
+.. opt:: Set Diffs %( "on" %| "off" %| "removed" %)
+.. This ref doesn't work: :opt:`Set Diffs %( "on" %| "off" %| "removed" %)`
+
+  The “on” option highlights added tokens in green, while the “removed” option
```
Zimmi48: Note you should also indent the comment above, otherwise Sphinx interprets this paragraph as part of the comment.

doc/sphinx/proof-engine/proof-handling.rst
Zimmi48: I don't know what is a `:see`. But you can you (top-level) `.. seealso::` directive (if you want): it works like `.. note::`.

doc/sphinx/proof-engine/proof-handling.rst
Zimmi48: Use:

    How diffs are calculated
    ````````````````````````

because this is how the previous sub-section work in this chapter.

doc/sphinx/proof-engine/proof-handling.rst
Zimmi48: Yes, I think it would make sense to add the paper to the manual's bibliography.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -539,6 +546,137 @@ Requesting information
...
+* Goals completed in the last proof step will not be shown even with the
+  "removed" setting.
+
+The following screenshots show diffs working with multiple goals and with compacted
+hypotheses.  In the first one, notice that the goal ``P 1`` is not highlighted at
+all after the split because it has not changed.
+
+  .. coqtop:: none
+
+    Set Diffs "on".
```
Zimmi48: I generally recommend using one more space of indentation (again for consistency with the rest of the manual).

Zimmi48:
Testing on my side:
screenshot_2018-08-24 proof handling coq 8 9 alpha documentation
I think the problem is that the colors are only used for the input and not the output. @cpitclaudel Aren't the input colors coming from coqdoc and not coqtop?

jfehrle:
What are the rules to identify various levels of titles/sections (which I can add to the README)?  I got confused from http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections because we seem to have a different convention, i.e.:

- "Proof Handling" (chapter) has "--------" above and below it.
- "Switching on/off the proof editing mode" (section) has "----------" below it
- "Proof using options" (subsection) has "`````````````" below it

A couple of places have "============", such as "How to read this book" in the Introduction, which appears to give a section-sized title without putting it in the table of contents

jfehrle:
> I don't know what is a :see. But you can you (top-level) .. seealso:: directive (if you want): it works like .. note::.

Sorry, I meant something like:
```
see Section :ref:`controlling-display`
```

jfehrle:
> Yes, I think it would make sense to add the paper to the manual's bibliography.

I added this entry to biblio.bib but I don't see it in zebibliography.html.  What do I need to do?  Keep in mind I'm in a Windows environment, there were some parts of the tool chain I couldn't easily install.

```
@Article{Myers,
  author    = {Eugene Myers},
  title     = {An O(ND) Difference Algorithm and its Variations},
  journal   = {Algorithmica},
  volume    = {1},
  number    = {2},
  year      = {1986},
  bibsource = {https://link.springer.com/article/10.1007%2FBF01840446},
  url       = {http://www.xmailserver.org/diff2.pdf}
}
```

jfehrle:
@Zimmi48 Addressed your latest points and rebased.  Not done yet.  See earlier comments.

doc/sphinx/proof-engine/proof-handling.rst
cpitclaudel: Thanks, I looked into this.  The parser in Emacs doesn't support these 24-bit color codes :/
Adding them shouldn't be too hard, but I don't really have the bandwidth for that at the moment.
Is there a way to specify COQ_COLORS that would make Coq use the more common 8-color palette? It includes standard and bright red and green, so we could use this for Emacs.

cpitclaudel:
> I think the problem is that the colors are only used for the input and not the output. @cpitclaudel Aren't the input colors coming from coqdoc and not coqtop?

We use coqdoc for the input and coqtop's ANSI colors for the output (see https://coq.inria.fr/distrib/current/refman/user-extensions/proof-schemes.html for an example)

But the ANSI color parser only handles the standard 8 colors, not 24-bit ones, so we'd need to update the python parser if we want the colors rendered in the manual.  It might be worth doing if we want to show diffs in other examples; otherwise, screenshots might be just fine.

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: Yes, setting this in COQ_COLORS should work just fine.  Tag names are shown above.

Should I still submit an issue for PG?  Seems like you may be able to explain what's needed at this point better than I can.  If you create the issue, please make sure I see it so I can comment on it.

cpitclaudel: Done at https://github.com/ProofGeneral/PG/issues/381 . Thanks for taking the time to explain!

jfehrle: Thanks for your help!

Zimmi48:
>I added this entry to biblio.bib but I don't see it in zebibliography.html. What do I need to do? Keep in mind I'm in a Windows environment, there were some parts of the tool chain I couldn't easily install.

You need to `:cite:` the paper.

>I got confused from http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections because we seem to have a different convention, i.e.:

If there are standard conventions, they weren't respected during the migration and I would support / merge a PR that changes the manual to follow the standard conventions.

>But the ANSI color parser in the manual only handles the standard 8 colors, not 24-bit ones, so we'd need to update the python parser if we want the colors rendered in the manual. It might be worth doing if we want to show diffs in other examples; otherwise, screenshots might be just fine.

It would be great to support this in the manual. Couldn't another solution be to tweak `COQ_COLORS` when compiling the manual?

cpitclaudel:
> It would be great to support this in the manual. Couldn't another solution be to tweak COQ_COLORS when compiling the manual?

Yeah, that should work too :)

jfehrle:
> Couldn't another solution be to tweak COQ_COLORS when compiling the manual?

In general I think it's better to show the actual colors you'd see as a user.  That idea is confounded by the fact that the colors will be different for coqtop, CoqIDE and PG.  Even if we use the coqtop colors, they're tuned for light type on a black background--they may not look that good with a white background.

I suggest we enhance the Python code so it can handle the 24-bit colors, then if we still don't like the look we can set COQ_COLORS.  Can we live with screen shots for now?

Also, FWIW, I'm not so keen on having color highlighting for tactics and commands in the examples because you don't see that in the user interface.  `Show Match nat.` will not have color highlights in the UI:

image

Zimmi48:
>In general I think it's better to show the actual colors you'd see as a user. That idea is confounded by the fact that the colors will be different for coqtop, CoqIDE and PG. Even if we use the coqtop colors, they're tuned for light type on a black background--they may not look that good with a white background.

The background won't be the same in coqtop for different users. Can't we find a set of color that works well with a clear background for the manual? Being able to show proof diffs in the manual seems like something that could be useful.

>Can we live with screen shots for now?

For CoqIDE yes, but for coqtop I am convinced (since Clément has made the remark) that we shouldn't put any screenshot since we don't want to give any incentive to users to use coqtop.

>Also, FWIW, I'm not so keen on having color highlighting for tactics and commands in the examples because you don't see that in the user interface. Show Match nat. will not have color highlights in the UI:

Once again it will depend on the user and their UI and I don't see the problem with more syntactic color highlighting.

jfehrle:
Also, FYI, I notice the capitalization of the paper title in the bibliography gets changed.  The output is:

image

But `O(ND)` in the biblio.bib should not have been changed to `o(nd)`.  Not gonna worry about that.

jfehrle:
> Can't we find a set of color that works well with a clear background for the manual?

Of course, but for the light grey background used in the examples.  "clear" is not a background color, eh?  @cpitclaudel, will we be able to configure COQ_COLORS for Sphinx in time for 8.9?  If it's simple I could add that in this PR (send the changes or give me some directions).

> we don't want to give any incentive to users to use coqtop.

Is there a desire to make coqtop go away at some point?  I agree we don't want to encourage using coqtop for proof development.  But it is useful for developers.

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -95,6 +97,15 @@ list of assignments of the form ``name=``:n:``{*; attr}`` where
 ANSI escape code. The list of highlight tags can be retrieved with the
```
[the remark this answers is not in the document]
Zimmi48: @jfehrle This is fixed in #8281.

Zimmi48:
>Is there a desire to make coqtop go away at some point? I agree we don't want to encourage using coqtop for proof development. But it is useful for developers.

Exactly, we'll likely keep it because it's useful for developers, but the reference manual is more of a user manual.

Zimmi48:
@jfehrle Regarding your question on capitalization, see https://tex.stackexchange.com/questions/10772/bibtex-loses-capitals-when-creating-bbl-file

doc/sphinx/biblio.bib
```diff
@@ -294,6 +294,17 @@ @MastersThesis{Mun94
...
 }
 
+@Article{Myers,
+  author    = {Eugene Myers},
+  title     = {An O(ND) Difference Algorithm and its Variations},
+  journal   = {Algorithmica},
+  volume    = {1},
+  number    = {2},
+  year      = {1986},
+  bibsource = {https://link.springer.com/article/10.1007%2FBF01840446},
```
Zimmi48: Add a backslash before the %, otherwise it is interpreted as a comment.

doc/sphinx/practical-tools/coq-commands.rst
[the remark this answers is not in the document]
jfehrle: I remembered seeing it but I couldn't find it before.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -572,6 +579,166 @@ Requesting information
...
+   them as strings to the lexer to break them into tokens.  Then apply the
+   Myers diff algorithm :cite:`Myers` on the tokens and add appropriate highlighting.
+
+Notes:
+
+* Aside from the highlights, output for the "on" option should be identical
+  to the undiffed output.
+* Goals completed in the last proof step will not be shown even with the
+  "removed" setting.
+
```
jfehrle: These are 2 minor comments that IMHO shouldn't have an attention-grabbing `Note` box around them.  If you like, I could avoid the word `Note` in the text.

jfehrle:
I think this is good, please take a look.  Still waiting for the other PR to merge.

doc/sphinx/README.rst
Zimmi48: Don't forget about this.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -423,6 +423,9 @@ The following example script illustrates all these features:
     assumption.
     Qed.
 
+  .. coqtop:: none
+
+    Abort.
```
Zimmi48: This addition has become useless since the introduction of the Qed above.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -572,6 +579,166 @@ Requesting information
...
+
+..
+
+  .. image:: ../_static/diffs-coqtop-on3.png
+     :alt: coqtop with Set Diffs on
+
+How to enable diffs
+```````````````````
+
+.. opt:: Set Diffs %( "on" %| "off" %| "removed" %)
```
Zimmi48: The option / flag itself shouldn't contain the `Set` part. See other examples in the manual.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -572,6 +579,166 @@ Requesting information
...
+``Edit | Preferences | Tags`` panel by changing the settings for the ``diff.added``,
+``diff.added.bg``, ``diff.removed`` and ``diff.removed.bg`` tags.  This panel also
+lets you control other attributes of the highlights, such as the foreground
+color, bold, italic, underline and strikeout.
+
+Note: As of this writing (August 2018), Proof General will need minor changes
+to be able to show diffs correctly.  We hope it will support this feature soon.
+See https://github.com/ProofGeneral/PG/issues/381 for the current status.  Note
+that there may be a delay between resolving this issue and its appearance in the
+next released version of Proof General.
```
Zimmi48: I think you can drop the last sentence since (AFAIU) both the documented way of installing PG (using `git clone`) and the MELPA package allow getting a version that follows master.

doc/sphinx/proof-engine/proof-handling.rst
Zimmi48: I don't mind keeping things as-is if you like.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -572,6 +579,166 @@ Requesting information
...
+* Aside from the highlights, output for the "on" option should be identical
+  to the undiffed output.
+* Goals completed in the last proof step will not be shown even with the
+  "removed" setting.
+
+.. comment The following screenshots show diffs working with multiple goals and with compacted
+   hypotheses.  In the first one, notice that the goal ``P 1`` is not highlighted at
+   all after the split because it has not changed.
+
+    .. comment todo: Use this script and remove the screenshots when COQ_COLORS
```
Zimmi48: You could have avoided repeating `.. coment` by putting it on the line above at the same level as the previous one. All what follows and is indented is considered to be part of the comment block.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -572,6 +579,166 @@ Requesting information
...
+Notes:
+
+* Aside from the highlights, output for the "on" option should be identical
+  to the undiffed output.
+* Goals completed in the last proof step will not be shown even with the
+  "removed" setting.
+
+.. comment The following screenshots show diffs working with multiple goals and with compacted
+   hypotheses.  In the first one, notice that the goal ``P 1`` is not highlighted at
+   all after the split because it has not changed.
```
Zimmi48: It seems that this commented out text is pretty much the same that the one which follows a bit later.

doc/sphinx/README.rst
Zimmi48: Sorry you had already taken this comment into account.

doc/sphinx/proof-engine/proof-handling.rst
jfehrle: That was intentional.  coqtop will show the before state and the tactic in the box whereas the screen shots don't show either of these.  So they get different text.  So I saved it.

jfehrle:
OK, covered you latest comments.

Zimmi48:
Rendered: https://coq.gitlab.io/-/coq/-/jobs/93182682/artifacts/_install_ci/share/doc/coq/sphinx/html/proof-engine/proof-handling.html#showing-differences-between-proof-steps

Zimmi48 (approved):
Looks good. Waiting for #8247 to get merged before merging this one. Please ping me when this is the case.

And let's not forget to check the status of https://github.com/ProofGeneral/PG/issues/381 to know if the text must be updated.

ejgallego:
@Zimmi48 , ready to go IIUC.

jfehrle:
@Zimmi48 This is ready, please merge into master.

What do I need to do now to get this and #8247 into 8.9?  Email Guillaume and make my case to him?

ejgallego:
> What do I need to do now to get this and #8247 into 8.9? Email Guillaume and make my case to him?

We do not know yet. So far having set the milestone to 8.9 should do the trick for now.

Zimmi48:
@jfehrle I plan to write to Guillaume myself about the beta release so I'll have a word in particular about these two PRs.

Please rebase this PR so that CI runs one last time before merging.

jfehrle:
Rebased.

jfehrle:
@Zimmi48 Would you merge this into master and arrange for backport into 8.9?

Zimmi48:
Oops, sorry I thought I had merged this.

jfehrle:
Thanks.  When can we get this into 8.9?

Zimmi48:
That's @silene's call. But don't worry I'm sure it will get in before the beta.

jfehrle:
> There is only one version and it is Coq-specific and it is the right URL you liked to.

@Zimmi48, what is the Proof General branch name for the Coq-specific version of PG?  Is it `async`?  Or `master`?

Zimmi48:
Both were advertised as Coq-specific not so long ago. It seems that Proof General has gained once more support for more than one proof assistant in its `master` branch. `async` is definitely Coq-specific, but not the version that most people use.

jfehrle:
The master branch passes `-emacs` to Coq.  Supporting diffs in PG on this branch will require very minor changes to Coq (maybe ~10 lines) so highlighting tags are embedded in the output, plus PG changes to convert the tags to highlights.  The async branch appears to receive XML; I haven't experimented with that version yet to see if any Coq change is necessary.

Are there any other branches we'd want to support?  master is most used, async is second?

My timing is not great.  I should have started on this a month or two ago.  I was hoping someone with PG development experience would take this on.  It's questionable whether I can get the Coq changes ready and merged for 8.10.  If I miss 8.10 and the changes are sufficiently small/obvious, perhaps they could go into 8.10.x and/or 8.9.x?

**Off-topic**: I was able to install Linux for Windows+Ubuntu and then Coq, emacs and PG on top of it.  I had to work around an OPAM issue with LfW (https://github.com/ocaml/opam-repository/issues/13858) by building a few things outside of OPAM, such as Coq.

The Ubuntu website says it needs 25GB of disk space.  The actual requirement is much less, I saw my free space drop by only 5GB, and that may overstate it.  I had bought an external SSD with the intention of putting LfW+Ubuntu on it, but I learned that LfW must be installed on the system drive.  It appears that LfW won't read external SSDs.  Perhaps that's because it would be more complex to handle cleanly removing external drive within LfW.  LfW seems slower for building Coq; I've seen comments suggesting that I/O is slower within LfW.

LfW is able to access Windows files, so I didn't need to install my usual editor inside LfW.  Files in the Ubuntu file system appear to be available outside LfW, although the path to the Ubuntu file system is very long and not well documented.  I only discovered it by chance.

Some of the limitations mentioned on the LfW website no longer apply, namely having to reboot after installing LfW and not being able to read LfW files from Windows processes.

Overall, LfW seems workable.

Zimmi48:
Cool experience report on Linux on Windows!

Regarding adapting what `-emacs` sends, I wouldn't do the changes unless it is requested by PG maintainers: indeed, what's the point of doing this if they do not then adapt PG to be able to support the diff feature? If they want to support it and plan to use it, then obviously it makes more sense.

For the async branch, no change on the Coq side should be necessary since it is using the same protocol as CoqIDE. All the changes to support diffs should be done on the PG side in that case, it seems.

jfehrle:
Well, I submitted https://github.com/ProofGeneral/PG/issues/381 last August, which @cpitclaudel and @erikmd commented on.  They suggested supporting diffs in both the master and async branches.  Since no PG developer has volunteered to make the change, I'm looking at doing it myself.  (I guess that would technically make me a PG maintainer.)

> For the async branch, no change on the Coq side should be necessary since it is using the same protocol as CoqIDE.

I added a check in 8.10 requiring `-color` be enabled in order to enable diffs.  I need to check that this doesn't create a problem for PG/async.  And perhaps adjust the check.

Would you respond to the questions in my previous message?

Zimmi48:
Well obviously, if you are ready to do the changes to PG yourself, you should go ahead. In this case, I would recommend (but maybe this is already what you planned for) to do the changes to Coq and PG in parallel so that you can validate them.

>I added a check in 8.10 requiring -color be enabled in order to enable diffs.

Isn't it something coqtop-specific? I don't understand why the XML protocol would change the tags it sent based on this command-line option.

>Would you respond to the questions in my previous message?

>Are there any other branches we'd want to support? master is most used, async is second?

This is a question for PG maintainers to answer. I can only guess.

> If I miss 8.10 and the changes are sufficiently small/obvious, perhaps they could go into 8.10.x and/or 8.9.x?

This is a question for RMs to answer. 8.10.x maybe (seems acceptable at least for 8.10.0 which isn't due for almost two more months). 8.9.x I doubt it because I expect 8.9.1 to be released soon and to be the last 8.9 release.

jfehrle:
> Isn't it something coqtop-specific? I don't understand why the XML protocol would change the tags it sent based on this command-line option.

IIRC, diff tags appear in the XML only if diffs are enabled.  And -color is required to enable diffs.  I'll figure it out and do the right thing.

PR #8302, opened 2018-08-23
Fix #7795: UGraph.AlreadyDeclared with Program

### LINKED ISSUE #7795: Anomaly "Uncaught exception UGraph.AlreadyDeclared."
#### Version

8.8.0
coq-stdpp dev.2018-06-10.1.aa942ca8

#### Operating system

Debian

https://gist.github.com/jad-hamza/4dd6d528a353d8268a39a89da0668f72

This example gives the error message:

```
Error: Anomaly "Uncaught exception UGraph.AlreadyDeclared."
```

Maybe related to https://github.com/coq/coq/issues/7792?

Removing the `stdpp.set` import prevents the problem. I'm not sure whether this is a Coq bug or a stdpp problem, so I also posted there: https://gitlab.mpi-sws.org/robbertkrebbers/coq-stdpp/issues/20

SkySkimmer:
This change is based on noticing that we use a default value for the
`sideff` argument even though we have a similarly named `side_eff`
available. Someone who knows how side effects and universes are
supposed to interact should check this.

Also @jad-hamza please have a look to see if this fixes the real world instance of the bug.

### DIFF
--- a/proofs/pfedit.ml
+++ b/proofs/pfedit.ml
@@ -173,8 +173,8 @@ let build_by_tactic ?(side_eff=true) env sigma ?(poly=false) typ tac =
       const_entry_body = Future.chain ce.const_entry_body
         (fun (pt, _) -> pt, ()) } in
   let (cb, ctx), () = Future.force ce.const_entry_body in
-  let univs' = Evd.merge_context_set Evd.univ_rigid (Evd.from_ctx univs) ctx in
-  cb, status, Evd.evar_universe_context univs'
+  let univs = UState.merge side_eff Evd.univ_rigid univs ctx in
+  cb, status, univs
 
 let refine_by_tactic env sigma ty tac =
   (** Save the initial side-effects to restore them afterwards. We set the
--- a/test-suite/bugs/closed/7795.v
+++ b/test-suite/bugs/closed/7795.v
@@ -1,0 +1,65 @@
+
+
+Definition fwd (b: bool) A (e2: A): A. Admitted.
+
+Ltac destruct_refinement_aux T :=
+  let m := fresh "mres" in
+  let r := fresh "r" in
+  let P := fresh "P" in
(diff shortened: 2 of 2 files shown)


ppedrot (commented):
I summon @mattam82 for the review of this PR, because it is both about universe constraints and Program. Apart for the `side_eff` flag, I don't really see the difference introduced by this patch, as `Evd.merge_context_set` internally uses `UState.merge`.

ppedrot:
This is another argument in favour of trying to get ride of side-effects (by 8.10 hopefully?).

SkySkimmer:
>This is another argument in favour of trying to get ride of side-effects (by 8.10 hopefully?).

Is there a plan for that somewhere?

ppedrot:
Weren't you the one advocating for opening a bug around local definitions and abstract?

SkySkimmer:
Is that the same as removing side effects?

Gaëtan Gilbert

ppedrot:
Indeed. Side-effects are the data structure used to carry local definitions inside proofs.

Zimmi48:
@mattam82 ping

mattam82:
It seems correct to use `side_eff` indeed, when we inline the definitions in the body and force it we get an restriction of the `univs` context in `ctx`, hence the same universe declarations can appear.  This is because we call close_proof with now = true in this code path. This is truly messy code though, I can't quite get what happens when we have side effects in the term produced by build_constant_by_tactic in the following case:
```
Ltac constraint := exact (let f := fun u : Type => I in let g := Type in f g).
Obligation Tactic := abstract (intros; constraint).
Definition T : True -> True := _.
(** abstract side effect not inlined, universes are repeated *)
```

I'd also love to see a path towards removing side-effects or making their handling much cleaner.

mattam82 (approved):

PR #8303, opened 2018-08-23
Better controls for template polymorphism

SkySkimmer:
This adds atributes `template` and `notemplate` to control whether an inductive type should be made template polymorphic.
When neither is specified, templateness is inferred as previously unless new option `Auto Template Polymorphism` is unset.

There are tests for the basic functionality (in addition to the stdlib of course) but could be more extensive, eg records are a different code path, etc
Feel free to post some tests to this PR and I'll add them (or if you have the perms you can push directly).

Eventually we may want to move towards remove the automatic option (and someday maybe even remove template completely) but let's get the controls in first.

### DIFF
--- a/doc/sphinx/language/cic.rst
+++ b/doc/sphinx/language/cic.rst
@@ -1025,8 +1025,26 @@ the Type hierarchy.
 Template polymorphism
 +++++++++++++++++++++
 
-Inductive types declared in :math:`\Type` are polymorphic over their arguments
-in :math:`\Type`. If :math:`A` is an arity of some sort and :math:`s` is a sort, we write :math:`A_{/s}`
+Inductive types can be made polymorphic over their arguments
+in :math:`\Type`.
+
+.. opt:: Auto Template Polymorphism
+
+  This option, enabled by default, makes every inductive type declared
+  at level :math:`Type` (without annotations or hiding it behind a
+  definition) template polymorphic.
+
+  This can be prevented using the ``notemplate`` attribute.
+
+  An inductive type can be forced to be template polymorphic using the
+  ``template`` attribute.
+
+  Template polymorphism and universe polymorphism (see Chapter
+  :ref:`polymorphicuniverses`) are incompatible, so if the later is
+  enabled it will prevail over automatic template polymorphism and
+  cause an error when using the ``template`` attribute.
+
+If :math:`A` is an arity of some sort and :math:`s` is a sort, we write :math:`A_{/s}`
 for the arity obtained from :math:`A` by replacing its sort with :math:`s`.
 Especially, if :math:`A` is well-typed in some global environment and local
 context, then :math:`A_{/s}` is typable by typability of all products in the
--- a/kernel/indtypes.ml
+++ b/kernel/indtypes.ml
@@ -280,7 +280,7 @@ let typecheck_inductive env mie =
     List.fold_left
       (fun (env_ar,l) ind ->
          (* Arities (without params) are typed-checked here *)
-	 let expltype = ind.mind_entry_template in
+         let template = ind.mind_entry_template in
          let arity =
 	   if isArity ind.mind_entry_arity then
 	     let (ctx,s) = dest_arity env_params ind.mind_entry_arity in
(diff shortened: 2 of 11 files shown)


Zimmi48 (approved):
Approving the doc part (although having an example would be even better).
I'm also supportive of the general idea of this PR.

mattam82 (approved):
The code looks good to me (note I'm learning about the attributes system with this patch). I think these controls can be very useful at least to evaluate the transition to a system with no template poly and for users who want control on the sometimes surprising "auto-activation" of template polymorphism depending on the arity's sort annotation.

SkySkimmer:
@maximedenes ping

mattam82:
In principle, with cumulative polymorphism, nothing should prevent moving
to univ polymorphism except for the lowering to Prop behavior. That should
be rare in the current stdlib.
Le jeu. 13 sept. 2018 à 22:22, Hugo Herbelin <notifications@github.com> a
écrit :

SkySkimmer:
All the tactics which are broken with polymorphic stdlib constants

Gaëtan Gilbert

SkySkimmer:
> I'm sorry, I don't understand your sentence. Wasn't some part of it
 > inadvertently cut?

Option is template polymorphic, suppose you make it cumulative 
polymorphic, then any tactic which manipulates options internally and 
uses constr_of_global for it will break.

Gaëtan Gilbert

herbelin:
@skyskimmer: you mean that we need first to move all calls to `constr_of_global` in tactics to, I don't remember what is the canonical replacement, something like `Evd.fresh_global` I guess? If this is what is expected, I can do my share if it can help.

herbelin:
>  I believe we could support it but I’d rather not do that and focus on removing the implicit Prop <= Type instead

We probably already talked about it. If I remember correctly I heard pedagogical motivations (like people wondering why `Inductive unit := I.` is put in `Prop`, and it is hard to explain that `Prop` can be seen as the sort of the subsingleton types), technical motivations (like extraction expected to be simpler), semantical motivations (difficult to handle for the set-theoretic model without relying on a judgemental equality - but don't we need anyway a judgemental equality for the set-theoretic model?). Did I exhaust the list of motivations?

On the other side, aren't we also going towards other troubles in terms of compatibility or complexification of the implementation? Did someone already try to remove `Prop <= Type` in real life?

JasonGross:
I think there might be a place in fiat-parsers where I make a bunch of definitions with `prod` in a module functor and instantiate in one place with prop things and in another place with type things.

Also, I find it quite useful to have `sig` of two `Prop`s land in `Prop`, because this lets me use it as a `Prop` (like `ex`) but destruct it even in any informative context (unlike `ex`).

SkySkimmer:
I wonder if it would be possible to combine removing `Prop <= Set` and adding the possibility to instantiate a universe binder with `Prop`.
If I have `foo@{i j | i <= j}` we can make it so instantiating either with `Prop` forces the other to be `Prop`, but for `foo@{i j | i < j}` we can't tell whether it's only to have `Type@{i} : Type@{j}` (in which case instantiating `i` with `Prop` could be allowed) or if it uses the implied `i <= j`.
Maybe making an error when instantiating wit `Prop` in a strict constraint but supporting nonstrict ones would be enough for the uses template polymorphism has (eg `sigT` has no constraints if using the algebraic type, or nonstrict constraints if using a max).
This method shouldn't require anything from the model (`sigT@{Prop Prop}` is just the copy `Inductive sigTP (A:Prop) (B:A->Prop) : Prop := ...`), I don't know about extraction.

Even if we don't remove `Prop <= Set` I think being able to instantiate polymorphic binders with `Prop` would still be useful.

mattam82:
@letouzey started experimenting this in real life indeed. Of course that means a more complex higher-level for a simpler kernel. I think you explained the motivations indeed. About the set-theoretic model the problem is rather that one has to use a trace encoding.

mattam82:
Remember that there’s an implicit Set <= i constraint on the universe
polymorphic universes when type checking definitions, so we would have to
weaken this to Prop <= i to ensure we are not introducing contradictions.
Then indeed i’s could be instantiated with Prop. The question that remains
would be what to do with remaining Prop </<= i </<= Set constraints. Would
it be fine to allow universes between Prop and Set? I believe
consistency-wise it would be fine, we’d still get an assignment of naturals
to variables making the constraints hold. We’d have to change some
fast-path checks in univ though.
Le sam. 15 sept. 2018 à 19:42, Gaëtan Gilbert <notifications@github.com> a
écrit :

PR #8313, opened 2018-08-24
[ci-fiat-crypto] Test extraction

JasonGross:
**Kind:** infrastructure.

On top of #8312 

Currently makes ocamlopt stack-overflow on the CI (but not on my machine).  Is this flambda?  Something else?

### DIFF
--- a/dev/ci/ci-fiat-crypto.sh
+++ b/dev/ci/ci-fiat-crypto.sh
@@ -9,4 +9,7 @@ git_checkout "${fiat_crypto_CI_BRANCH}" "${fiat_crypto_CI_GITURL}" "${fiat_crypt
 
 ( cd "${fiat_crypto_CI_DIR}" && git submodule update --init --recursive )
 
-( cd "${fiat_crypto_CI_DIR}" && make new-pipeline )
+# We need a larger stack size to not overflow ocamlopt+flambda when
+# building the executables.
+# c.f. https://github.com/coq/coq/pull/8313#issuecomment-416650241
+( cd "${fiat_crypto_CI_DIR}" && ulimit -s 32768 && make new-pipeline c-files )


ejgallego:
@JasonGross it could well be flambda, that's easy to try with OPAM; for sure worth reporting to the OCaml devs if true.

ejgallego:
So ideally we should obtain the `.ml` file that causes the problem in order to submit an OCaml bug report.

I may not be able to do it soon, sorry, but I can help with the OCaml investigation if you provide the file.

JasonGross:
Here is the .ml file: [unsaturated_solinas.tar.gz](https://github.com/coq/coq/files/2328849/unsaturated_solinas.tar.gz)

It does indeed seem to be an flambda issue
```
$ ocamlopt -v
The OCaml native-code compiler, version 4.07.0
Standard library directory: /usr/local/google/home/jagro/.opam/4.07.0/lib/ocaml
$ time ocamlopt -w -20 unsaturated_solinas.ml -o unsaturated_solinas

real	0m5.957s
user	0m5.484s
sys	0m0.452s
```
vs
```
$ ocamlopt -v
The OCaml native-code compiler, version 4.07.0
Standard library directory: /usr/local/google/home/jagro/.opam/4.07.0+flambda/lib/ocaml
$ time ocamlopt -w -20 unsaturated_solinas.ml -o unsaturated_solinas
Fatal error: exception Stack overflow

real	0m3.274s
user	0m3.160s
sys	0m0.108s
```

ejgallego:
Great, thanks @JasonGross ; I will submit a bug report to OCaml once I can investigate a bit, feel of course free to do it yourself and I'm just back from holidays and carry a large backlog.

As for Coq, maybe we could increase the stack size? What does happen if you increase the stack size for `ocamlopt.opt` ?? [use `ulimit`]

ejgallego:
OCaml report here: https://caml.inria.fr/mantis/view.php?id=7842

It seems like the flambda middle-end could use some tail-call optimizations, so reporting anyways.

ejgallego:
> Where should this be done? Inside ci-fiat-crypto.sh?

I think so, just before the call to the proper target [add a small comment please pointing to this issue]

JasonGross:
done

JasonGross:
16384 isn't enough for all the files; I'm going to try 32768

ejgallego (approved):
CI failure is spurious; [the mythical ltac_prof stuff].

Issue #8321, opened 2018-08-25
Add more useful addons to the Windows Installer


MSoegtropIMC:
Add some more useful addons to the Windows installer.

I frequently (or at least occassionally) use:
- Princeton VST + the required CompCert tools
- QuickChick
- Menhir and Mnehir Coq library
- math components + ssreflect
- coquelicot
- aactactics

Zimmi48:
When it was just about including a few small plugins with the Coq Windows installer, it was quite reasonable to add this to the build time but of course with larger developments like VST, it becomes more complicated. What we would need ideally would be a generic framework to build a Windows installer for any package in the Coq index. And such various builds could be executed in parallel. The Windows users would just have to download from the website the installers for whatever package they are interested in.

MSoegtropIMC:
@Zimmi48 : yes, this is my plan. But I can't get there all in one step. My mid term plan is to have an automated developer setup, which allows to compile plugins for a given Coq version and second to have an installer with downloadable plugins.

Maybe mid term I will also switch to opam. For the developer environment I plan to provide both: opam and my scripts.

Zimmi48:
This issue has been completed and should be closed, right?

MSoegtropIMC:
Yes - meanwhile this is all included.

PR #8306, opened 2018-08-23
Add Mtac2 1.0 to Windows packages.

Zimmi48:
Like #8305 but for master.

**Kind:** infrastructure.

- [x] Entry added in CHANGES.

PR #8305, opened 2018-08-23
Add Mtac2 1.0 to Coq 8.8 Windows packages.

Zimmi48:
**Kind:** infrastructure.

- [x] Entry added in CHANGES.

Now that Mtac2 1.0 has been adapted to Coq 8.8, we add it to the Windows installer to give it more chances to "compete against" Ltac and Ltac2.
Note that this is a 8.8-only PR because Mtac2 1.0 is not yet compatible with Coq 8.9 / master (the `master-sync` branch that is tested in CI for Coq master is an outdated version of Mtac2 so it wouldn't make sense to distribute it in the Windows package). We hope that Mtac2 1.0 will be made compatible with Coq 8.9 by the time of the release but in the meantime, it would make sense to remove Mtac2 from CI for Coq master.

PR #7528, opened 2018-05-16
Document how to set up a Windows runner on CloudStack

maximedenes:

PR #8322, opened 2018-08-25
Additional addons for Windows installer

MSoegtropIMC:
This PR implements Issue #8321, adding more addons.

I don't enable these by default for the time beeing, because the build time gets really long.

It is more intended for e.g. teachers who want to ease installation of required tools.

Also I don't think I can enable this on CI - build time is more than 2 hours.

**ATTENTION**: There is one "major" change:

The coq library is now installed under <installroot>/lib/coq instead of <installroot>/lib. The main reason is that otherwise ocamlfind can't find libraries. One could better separate files used at Coq run time and plugin build time, but I think it doesn't hurt to have the library under Windows under the same folder as under Linux.

On the way I also did  few enhancements and reliability fixes to the build script.

I will merge this with the MTAC2 PR after some initial CI runs.

### DIFF
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -1647,7 +1647,7 @@ function make_addon_equations {
   installer_addon_dependency equations
   if build_prep_overlay Equations; then
     installer_addon_section equations "Equations" "Coq plugin for defining functions by equations" ""
-    # Note: PATH is autmatically saved/restored by build_prep / build_post
+    # Note: PATH is automatically saved/restored by build_prep / build_post
     PATH=$COQBIN:$PATH
     logn coq_makefile ${COQBIN}coq_makefile -f _CoqProject -o Makefile
     log1 make
@@ -1695,7 +1695,7 @@ function make_addon_ssreflect {
 function make_addon_ltac2 {
   installer_addon_dependency ltac2
   if build_prep_overlay ltac2; then
-    installer_addon_section ltac2 "Ltac-2" "Coq plugin with the Ltac-2 enhanced tactics language" ""
+    installer_addon_section ltac2 "Ltac-2" "Coq plugin with the Ltac-2 enhanced tactic language" ""
     log1 make all
     log2 make install
     build_post
@@ -1707,6 +1707,7 @@ function make_addon_ltac2 {
 function make_addon_unicoq {
   installer_addon_dependency unicoq
   if build_prep_overlay unicoq; then
+    installer_addon_section unicoq "Unicoq" "Coq plugin for an enhanced unification algorithm" ""
     log1 coq_makefile -f Make -o Makefile
     log1 make
     log2 make install
@@ -1721,6 +1722,7 @@ function make_addon_mtac2 {
   make_addon_unicoq
   installer_addon_dependency_end
   if build_prep_overlay mtac2; then
+    installer_addon_section mtac2 "Mtac-2" "Coq plugin for a typed tactic language for Coq." ""
     log1 coq_makefile -f _CoqProject -o Makefile
     log1 make
     log2 make install
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -2,7 +2,7 @@
 
 ###################### COPYRIGHT/COPYLEFT ######################
 
-# (C) 2016 Intel Deutschland GmbH
+# (C) 2016..2018 Intel Deutschland GmbH
 # Author: Michael Soegtrop
 #
 # Released to the public by Intel under the
@@ -238,9 +238,10 @@ fi
 # $1 file server name including protocol prefix
 # $2 file name (without extension)
 # $3 file extension
-# $4 number of path levels to strip from tar (usually 1)
-# $5 module name (if different from archive)
-# $6 expand folder name (if different from module name)
+# $4 [optional] number of path levels to strip from tar (usually 1)
+# $5 [optional] module name (if different from archive)
+# $6 [optional] expand folder name (if different from module name)
+# $7 [optional] module base name (used as 2nd choice for patches, defaults to $5)
 # ------------------------------------------------------------------------------
 
 function get_expand_source_tar {
@@ -261,6 +262,12 @@ function get_expand_source_tar {
     folder=$6
   else
     folder=$name
+  fi
+
+  if [ "$#" -ge 7 ] ; then
+    basename=$7
+  else
+    basename=$name
   fi
 
   # Set logging target
@@ -314,8 +321,11 @@ function get_expand_source_tar {
   fi
 
   # Patch if patch file exists
+  # First try specific patch file name then generic patch file name
   if [ -f "$PATCHES/$name.patch" ] ; then
     log1 patch -p1 -i "$PATCHES/$name.patch"
+  elif  [ -f "$PATCHES/$basename.patch" ] ; then
+    log1 patch -p1 -i "$PATCHES/$basename.patch"
   fi
 
   # Go back to base folder
@@ -340,6 +350,7 @@ function get_expand_source_tar {
 # $3 file extension
 # $4 [optional] number of path levels to strip from tar (usually 1)
 # $5 [optional] module name (if different from archive)
+# $6 [optional] module base name (used as 2nd choice for patches, defaults to $5)
 # ------------------------------------------------------------------------------
 
 function build_prep {
@@ -356,6 +367,15 @@ function build_prep {
     name=$2
   fi
 
+  if [ "$#" -ge 6 ] ; then
+    basename=$6
+  else
+    basename=$name
+  fi
+
+  # Set installer section to not set by default
+  installersection=
+
   # Check if build is already done
   if [ ! -f "$FLAGFILES/$name.finished" ] ; then
     BUILD_PACKAGE_NAME=$name
@@ -365,7 +385,7 @@ function build_prep {
 
     touch "$FLAGFILES/$name.started"
 
-    get_expand_source_tar "$1" "$2" "$3" "$strip" "$name"
+    get_expand_source_tar "$1" "$2" "$3" "$strip" "$name" "$name" "$basename"
 
     cd "$name"
 
@@ -381,6 +401,61 @@ function build_prep {
 }
 
 # ------------------------------------------------------------------------------
+# Like build_prep, but gets the data from an entry in ci-basic-overlay.sh
+# This assumes the following definitions exist in ci-basic-overlay.sh,
+# or in a file in the user-overlays folder:
+# $1_CI_REF
+# $1_CI_ARCHIVEURL
+# $1_CI_GITURL
+# ATTENTION: variables in ci-basic-overlay.sh are loaded by load_overlay_data.
+# load_overlay_data is is called at the end of make_coq (even if the build is skipped)
+#
+# Parameters
+# $1 base name of module in ci-basic-overlay.sh, e.g. mathcomp, bignums, ...
+# ------------------------------------------------------------------------------
+
+function build_prep_overlay {
+  urlvar=$1_CI_ARCHIVEURL
+  gitvar=$1_CI_GITURL
+  refvar=$1_CI_REF
+  url=${!urlvar}
+  git=${!gitvar}
+  ref=${!refvar}
+  ver=$(git ls-remote "$git" "refs/heads/$ref" | cut -f 1)
+  if [[ "$ver" == "" ]]; then
+      # $1_CI_REF must have been a tag or hash, not a branch
+      ver="$ref"
+  fi
+  build_prep "$url" "$ver" tar.gz 1 "$1-$ver" "$1"
+}
+
+# ------------------------------------------------------------------------------
+# Load overlay version variables from ci-basic-overlay.sh and user-overlays/*.sh
+# ------------------------------------------------------------------------------
+
+function load_overlay_data {
+  if [ -n "${GITLAB_CI+}" ]; then
+    export CI_BRANCH="$CI_COMMIT_REF_NAME"
+    if [[ ${CI_BRANCH#pr-} =~ ^[0-9]*$ ]]; then
+      export CI_PULL_REQUEST="${CI_BRANCH#pr-}"
+    else
+      export CI_PULL_REQUEST=""
+    fi
+  else
+    export CI_BRANCH=""
+    export CI_PULL_REQUEST=""
+    # Used when building 8.8.0 with the latest scripts
+    export TRAVIS_BRANCH=""
+    export TRAVIS_PULL_REQUEST=""
+  fi
+
+  for overlay in /build/user-overlays/*.sh; do
+    . "$overlay"
+  done
+  . /build/ci-basic-overlay.sh
+}
+
+# ------------------------------------------------------------------------------
 # Finalize a module build
 # - create name.finished
 # - go back to base folder
@@ -392,6 +467,7 @@ function build_post {
     touch "$FLAGFILES/$BUILD_PACKAGE_NAME.finished"
     PATH=$BUILD_OLDPATH
     LOGTARGET=other
+    installer_addon_end
   fi
 }
 
@@ -522,6 +598,101 @@ function files_to_nsis {
   tr '/' '\\' < "/build/filelists/$1" | sed -r 's/^\.(.*)\\([^\\]+)$/SetOutPath $INSTDIR\\\1\nFile ${COQ_SRC_PATH}\\\1\\\2/' > "/build/filelists/$1.nsh"
 }
 
+# ------------------------------------------------------------------------------
+# Create an nsis installer addon section
+#
+# parameters
+# $1 identifier of installer section and base name of file list files
+# $2 human readable name of section
+# $3 description of section
+# $4 flags (space separated list of keywords): off = default off
+#
+# $1 must be a valid NSIS identifier!
+# ------------------------------------------------------------------------------
+
+function installer_addon_section {
+  installersection=$1
+  list_files "addon_pre_$installersection"
+
+  echo 'LangString' "DESC_$1" '${LANG_ENGLISH}' "\"$3\"" >> "/build/filelists/addon_strings.nsh"
+
+  echo '!insertmacro MUI_DESCRIPTION_TEXT' '${'"Sec_$1"'}' '$('"DESC_$1"')' >> "/build/filelists/addon_descriptions.nsh"
+
+  local sectionoptions=
+  if [[ "$4" == *off* ]] ; then sectionoptions+=" /o" ; fi
+
+  echo "Section $sectionoptions \"$2\" Sec_$1" >> "/build/filelists/addon_sections.nsh"
+  echo 'SetOutPath "$INSTDIR\"' >> "/build/filelists/addon_sections.nsh"
+  echo '!include "..\..\..\filelists\addon_'"$1"'.nsh"' >> "/build/filelists/addon_sections.nsh"
+  echo 'SectionEnd' >> "/build/filelists/addon_sections.nsh"
+}
+
+# ------------------------------------------------------------------------------
+# Start an installer addon dependency group
+#
+# parameters
+# $1 identifier of the section which depends on other sections
+# The parameters must match the $1 parameter of a installer_addon_section call
+# ------------------------------------------------------------------------------
+
+dependencysections=
+
+function installer_addon_dependency_beg {
+  installer_addon_dependency "$1"
+  dependencysections="$1 $dependencysections"
+}
+
+# ------------------------------------------------------------------------------
+# End an installer addon dependency group
+# ------------------------------------------------------------------------------
+
+function installer_addon_dependency_end {
+  set -- $dependencysections
+  shift
+  dependencysections="$*"
+}
+
+# ------------------------------------------------------------------------------
+# Create an nsis installer addon dependency entry
+# This needs to be bracketed with installer_addon_dependencies_beg/end
+#
+# parameters
+# $1 identifier of the section on which other sections might depend
+# The parameters must match the $1 parameter of a installer_addon_section call
+# ------------------------------------------------------------------------------
+
+function installer_addon_dependency {
+  for section in $dependencysections ; do
+    echo '${CheckSectionDependency} ${Sec_'"$section"'} ${Sec_'"$1"'} '"'$section' '$1'" >> "/build/filelists/addon_dependencies.nsh"
+  done
+}
+
+# ------------------------------------------------------------------------------
+# Finish an installer section after an addon build
+#
+# This creates the file list files
+#
+# parameters: none
+# ------------------------------------------------------------------------------
+
+function installer_addon_end {
+  if [ -n "$installersection" ]; then
+    list_files "addon_post_$installersection"
+    diff_files "addon_$installersection" "addon_post_$installersection" "addon_pre_$installersection"
+    files_to_nsis "addon_$installersection"
+  fi
+}
+
+# ------------------------------------------------------------------------------
+# Set all timeouts in all .v files to 1000
+# Since timeouts can lead to CI failures, this is useful
+#
+# parameters: none
+# ------------------------------------------------------------------------------
+
+function coq_set_timeouts_1000 {
+  find . -type f -name '*.v' -print0 | xargs -0 sed -i 's/timeout\s\+[0-9]\+/timeout 1000/'
+}
 
 ###################### MODULE BUILD FUNCTIONS #####################
 
@@ -1287,6 +1458,8 @@ function make_coq {
 
     build_post
   fi
+
+  load_overlay_data
 }
 
 ##### GNU Make for MinGW #####
@@ -1450,36 +1623,17 @@ function make_coq_installer {
   fi
 }
 
-###################### ADDONS #####################
+###################### ADDON COQ LIBRARIES / PLUGINS / TOOLS #####################
 
 # The bignums library
 # Provides BigN, BigZ, BigQ that used to be part of Coq standard library
 
 function make_addon_bignums {
-  bignums_SHA=$(git ls-remote "$bignums_CI_GITURL" "refs/heads/$bignums_CI_REF" | cut -f 1)
-  if [[ "$bignums_SHA" == "" ]]; then
-      # $bignums_CI_REF must have been a tag / commit and not a branch
-      bignums_SHA="$bignums_CI_REF"
-  fi
-  if build_prep "$bignums_CI_ARCHIVEURL" "$bignums_SHA" zip 1 "bignums-$bignums_SHA"; then
+  installer_addon_dependency bignums
+  if build_prep_overlay bignums; then
+    installer_addon_section bignums "Bignums" "Coq library for fast arbitrary size numbers" ""
     # To make command lines shorter :-(
     echo 'COQ_SRC_SUBDIRS:=$(filter-out plugins/%,$(COQ_SRC_SUBDIRS)) plugins/syntax' >> Makefile.coq.local
-    log1 make all
-    log2 make install
-    build_post
-  fi
-}
-
-# Ltac-2 plugin
-# A new (experimental) tactic language
-
-function make_addon_ltac2 {
-  ltac2_SHA=$(git ls-remote "$ltac2_CI_GITURL" "refs/heads/$ltac2_CI_REF" | cut -f 1)
-  if [[ "$ltac2_SHA" == "" ]]; then
-      # $ltac2_CI_REF must have been a tag / commit and not a branch
-      ltac2_SHA="$ltac2_CI_REF"
-  fi
-  if build_prep "$ltac2_CI_ARCHIVEURL" "$ltac2_SHA" zip 1 "ltac2-$ltac2_SHA"; then
     log1 make all
     log2 make install
     build_post
@@ -1490,12 +1644,9 @@ function make_addon_ltac2 {
 # A function definition plugin
 
 function make_addon_equations {
-  Equations_SHA=$(git ls-remote "$Equations_CI_GITURL" "refs/heads/$Equations_CI_REF" | cut -f 1)
-  if [[ "$Equations_SHA" == "" ]]; then
-      # $Equations_CI_REF must have been a tag / commit and not a branch
-      Equations_SHA="$Equations_CI_REF"
-  fi
-  if build_prep "$Equations_CI_ARCHIVEURL" "$Equations_SHA" zip 1 "Equations-$Equations_SHA"; then
+  installer_addon_dependency equations
+  if build_prep_overlay Equations; then
+    installer_addon_section equations "Equations" "Coq plugin for defining functions by equations" ""
     # Note: PATH is autmatically saved/restored by build_prep / build_post
     PATH=$COQBIN:$PATH
     logn coq_makefile ${COQBIN}coq_makefile -f _CoqProject -o Makefile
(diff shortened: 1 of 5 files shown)


MSoegtropIMC:
P.S.: please let me first fix the (likely apprearing) CI issues and squash everything. I put a not here as soon as it is ready for review.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -63,6 +63,12 @@ RMDIR_BEFORE_BUILD=1
 #   Relocatble builds are first configured with prefix=./ then build and then
 #   reconfigured with prefix=<installroot> before make install.
 
+###################### COMPONENT VERSIOn OVERRIDES #####################
+
+# This is a temporary solution until the overlay mechanism is finalized
+
+# VST master did not work on 2018-08-19 (365d1b2d42eb31fff58bac560c078858602ba436)
+VST_CI_BRANCH=e49605cf1f1e5ae3bbec3d6554122427a94ae986
```
Zimmi48: VST master has been fixed in the meantime.

MSoegtropIMC: Thanks for the hint on VST - I will change it!

I have one strange effect: at home I get bignums-649dad440287151fd1937defb852886f192b8820 which is the head of v8.8 branch, not from master. On CI I get the head of the master branch which seems to fail. Any idea what is going on here?

Zimmi48: That's weird, could it be a caching problem? In any case, the diff between the tip of `v8.8` and `master` is limited to `.gitignore`, `.travis.yml` and `default.nix` (https://github.com/coq/bignums/compare/v8.8...master?expand=1) not something that should affect the build.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -1136,7 +1207,7 @@ function copy_coq_license {
     install -D README.md                      "$PREFIXCOQ/license_readme/coq/ReadMe.md"
     install -D CHANGES                        "$PREFIXCOQ/license_readme/coq/Changes.txt"
     install -D INSTALL                        "$PREFIXCOQ/license_readme/coq/Install.txt"
-    install -D doc/README.md                  "$PREFIXCOQ/license_readme/coq/ReadMeDoc.md"
+    install -D doc/README.md                  "$PREFIXCOQ/license_readme/coq/ReadMeDoc.md" || true
```
Zimmi48: I'd like to insist that this is not appropriate for the Windows build scripts to install more files than the normal install targets do. If you feel like these files should be installed, please open a PR modifying the Makefile instead and remove these lines from here. There is no reason why these files should be installed on Windows but not on macOS nor Linux.

MSoegtropIMC: @Zimmi48 : can we do this cleanup as a separate step? There is more to it and I don't want to make it too complicated. I will do this soon, but I don't want to touch the make files in this step as well. I think it would be cleaner to have a separate issue where changes here and the make file are done in a synchronized and thorough way.

My next steps in the pipeline regarding the installer are:

1.) Windows development environment (separate installation of Cygwin and compilation, support in place compilation, support opam based builds, support cygwin and MinGW builds)

2.) Switch from NSIS to WIX (Windows Installer XML Toolkit). WIX easily supports downloadable components and I just like MSI installers more than exe installers.

3.) Cleanup - I could do the cleanup of readme and license files at any time before - the additional addons also require some thoughts here. Another cleanup step is either completely removing (as absolete) or factoring out the support for old versions.

MSoegtropIMC:
@Zimmi48 : yes the patches to aac-tactcis and QuickChick are discussed with the maintainers. For QuickChick I got feedback that they take my changes (the issue here is that they buid things so that they are only compatible with a byte code version of Coq). For aac-tactics I don't have feedback yet. I am not sure that they take this. The way I patched this is how the tutorial should look in a properly installed version where files are where they belong. The way they do it probably works better for a stand alone compilation without installation.

MSoegtropIMC:
@Zimmi48 : regarding the build time: I think it would be nice to have a small installer and a big one - although mid term my goal is to have addons as downloadable components (after switching from NSIS to WIX). I plan to build the components in parallel then (as much as possible - there are a few dependencies). What to do short term I leave to the CI experts. One could enable a few of the quick to build and generally useful addons by default.

For compcert I am not sure if it makes any sense to install the .VO files, which are not required for using compcert (rather than verifying it). I added compcert because VST needs clightgen unless one does only exercises where the clightgen ouput is provided. Also I want to build VST on top of the original compcert files rather than their own copy, but this logic is not in yet (VST has an option for this which I need to set in case compcert is also there).

dev/build/windows/makecoq_mingw.sh
MSoegtropIMC: The v8.8 vs master issue with bignums is my bad: I edit the build files in a branch which is based on master, but when I build it I didn't use the local option, but build Coq 8.8.1, which has a difference ci_basic_overlays file which uses bignums v8.8. Time to get rid of this "build something else than you have" mess.

Still need to find out why CI fails ...

MSoegtropIMC:
The issue (hopefully) was a line I added to make the live of plugin writers easier:

export COQLIB=$RESULT_INSTALLDIR_MFMT/lib/coq/

I used CFMT as for COQBIN, which doesn't work because all programs are MinGW and cannot handle cygwin paths (/cygdrive/c/...). MFMT is C:/xyz WFMT is C:\xyz MinGW (actually all Windows programs not doing buggy path manipulations as well) can handle / instead of \, so MFMT is a good choice. CFMT is only intended for Linux path variables, where : is used as separator and for script internal usage (not exported variables). But maybe I change all usages of CFMT to MFMT where CFMT is not required.

MSoegtropIMC:
Btw.: Does the shell script linter still run? I expected to need two CI iterations just for that. Maybe I am meanwhile so well-behaved to get it right on first try, but I can't find the script lint anywhere anymore.

SkySkimmer:
We don't lint scripts (beyond basic whitespace linting).
There was a PR fixing some lint errors but it didn't add automatic linting (and I'm pretty sure lint errors remain).

MSoegtropIMC:
I am sure some sort of bash linting did run some time in the past, because I had CI failures cause of that. I found the feedback useful.

Actually I don't know much about bash linting. I could run it locally as well for my own good, but I don't know which tool and config file was used back then. Some pointers would be helpful.

dev/build/windows/makecoq_mingw.sh
MSoegtropIMC: @Zimmi48 : VST master (https://github.com/PrincetonUniversity/VST/commit/4ad211edd35cd21b3ffe1a63871cfe2ed914d183) definitely does not build, at least not in default configuration or if Compcert is not installed. Right now I cannot test it with compcert beeing installed because compcert requires menhirlib and menhirlib seems to have a regression - not sure how compcert is built in CI. I attached the VST error log.
The pinned version (e49605cf1f1e5ae3bbec3d6554122427a94ae986) does compile fine. So I keep the pinned version. I will discuss this with Andrew and Lennart.

MSoegtropIMC: Sorry, forgot the log ...

[VST-4ad211edd35cd21b3ffe1a63871cfe2ed914d183-make.log.txt](https://github.com/coq/coq/files/2321652/VST-4ad211edd35cd21b3ffe1a63871cfe2ed914d183-make.log.txt)
[VST-4ad211edd35cd21b3ffe1a63871cfe2ed914d183-make.err.txt](https://github.com/coq/coq/files/2321653/VST-4ad211edd35cd21b3ffe1a63871cfe2ed914d183-make.err.txt)

MSoegtropIMC:
One note: CI passes, but only because the additional addons are not enabled. With Coq master menhir-lib fails. See Issue #8328.

dev/build/windows/makecoq_mingw.sh
Zimmi48: Yes.

Zimmi48:
>For aac-tactics I don't have feedback yet.

Who did you write to? If it is to Thomas Braibant & Damien Pous, you shouldn't expect any feedback. The current maintainer of aac-tactics is @fakusb. The README of the aac-tactics repository hasn't been updated yet to reflect this. Feel free to open a PR on this repository, but note https://github.com/coq-community/aac-tactics/pull/7#issuecomment-409206838.

>Also I want to build VST on top of the original compcert files rather than their own copy, but this logic is not in yet (VST has an option for this which I need to set in case compcert is also there).

I also suggested to do this in CI in #6639 but as of today it is not the case yet (VST builds its own copy of compcert).

>Actually I don't know much about bash linting. I could run it locally as well for my own good, but I don't know which tool and config file was used back then. Some pointers would be helpful.

The author of the linting PR used "shellcheck".

MSoegtropIMC:
@Zimmi48 : I think I write to @fakusb (at least the abbreviation and the real name would fit).

MSoegtropIMC:
I am a bit confused by the recent failure. Is says that the OCaml num lib is already installed. Can it be that cgywin istallations are resued? It almost looks like it. Below are the final lines from the cygwin install:

```
solving: 0 tasks, update: no, use test packages: no
Augmented Transaction List: is empty
Augmented Transaction List: is empty
Changing gid back to original
```

Reusing cygwin installations does not work reliably, unless some additional (complicated and hard to maintain) steps are taken.

Zimmi48:
@MSoegtropIMC Yes, it could be the case due to the way VMs are currently used as runners: registered as shell executor, the VMs are not cleared between two independent builds; adding a clearing step, or even better switching to a docker-based setup, would solve this problem.

See also: https://github.com/coq/coq/pull/7528#pullrequestreview-120703994, #7428.

MSoegtropIMC:
@Zimmi48 : yes, I have the docker on my todo - sorry was really busy the last few months.

Would it be Ok to simply delete the cygwin folder before installation? I hope it isn't used by several builds in parallel! It would be easy to change the CI script.

One could also reuse cygwin (would safe a few minutes), but it is tricky and maintenance intensive and in my opinion not worth it.

MSoegtropIMC:
@Zimmi48 : It passes now (with just an additional ocamlfind list || true at the beginning). This shows that the CI build as is is unreliable cause of the cygwin reuse. But I would prefer to fix the cygwin CI stuff in a separate PR. This PR is already quite large - not in terms of number of lines, but in terms of what it pulls in.

**Status**: QuickChick does not compile with Coq master - it fails compiling some OCaml code. AACtactics also fails with Coq master in some OCaml code.  Mebhirlib and compcert compile fine with the supplied patch. Both QuickChick and AACtactics work with 8.8.1, so it is still useful to have these in. For Quickchick I supplied patches for several versions.

**Next steps**: My plan is to supply an installer for 8.8.1 with all addons, and add CI for all addons. For 8.8.2 I likely will also make the installer after the facts - as soon as all components are adjusted to 8.8.2. After that I hope with CI support that this builds out of the box. In parallel I wanted to work on the WIX replacement for NSIS.

So unless you want me to change things, I would think this is ready for merging.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -337,16 +348,39 @@ function build_prep {
   fi
 }
 
+# ------------------------------------------------------------------------------
+# Like build_prep, but gets the data from an entry in ci-basic-overlay.sh
+# This assumes the following definitions exist in ci-basic-overlay.sh
+# $1_CI_BRANCH
+# $1_CI_GITURL
```
Zimmi48: This doc is slightly confusing since this function does not call `ci-basic-overlay.sh` but just uses the environment variables that were defined in this file and sourced before-hand.

Zimmi48:
>My plan is to supply an installer for 8.8.1 with all addons, and add CI for all addons. For 8.8.2 I likely will also make the installer after the facts - as soon as all components are adjusted to 8.8.2.

Why do you say this? It shouldn't be expected that there is any incompatibility between 8.8.1 and 8.8.2 so it would be preferable that the installer scripts in 8.8.2 actually work with 8.8.2...

>So unless you want me to change things, I would think this is ready for merging.

I don't request any change. I'm a bit uncomfortable with this PR adding more "garbage" (patches in particular) but this is your realm and I can understand you want this in as a first step. I'm not the secondary code owner for these files in any case, @maximedenes is.

MSoegtropIMC:
> Why do you say this? It shouldn't be expected that there is any incompatibility between 8.8.1 and 8.8.2 so it would be preferable that the installer scripts in 8.8.2 actually work with 8.8.2...

Well assuming 8.8.2 is taken from master there are quite a few incompatibilities. At the user level there is the change in the fix tactic which in 8.8.1 can be called with just one argument but in 8.8.2 the documented two arguments are required. This is what the menhir-lib patch is about. Then there seem to be quite a few changes for plugin developers - at least several versions of QuickChick and AACtactics compile fine with 8.8.1 but not with Coq master. I was myself quite astonished how much fell apart when I switched from 8.8.1 to master. I want to provide the "deepspec" installer for 8.8.2 as soon as possible, but from what I have seen I don't expect this will happen next week (when 8.8.2 is expected if I remember right).

> I'm a bit uncomfortable with this PR adding more "garbage" (patches in particular)

I hope I can remove them soon. The idea is to add the addons to CI, so that in the future patches are not required. For the time being I think it is a good idea to test the same patches as I supply upstream - even if they are not tested in CI right now I test them at least locally. Of cause it would help to have a daily or weekly build with extended installer, just I don't know how to do this with gitlab.

Btw.: the "eternal patches", e.g. for lablgtk are there because they fix cygwin specific issues the upstream maintainers have limited interest in - at least if the patch means additional complexity for Linux. Patches for Coq addons are an entirely different story and should have a different dynamic.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -119,7 +126,11 @@ mkdir -p "$PREFIXCOQ/bin"
 mkdir -p "$PREFIXOCAML/bin"
 
 # This is required for building addons and plugins
+# This must be CFMT (/cygdrive/c/...) otherwise coquelicot 3.0.2 configure fails.
+# coquelicot uses which ${COQBIN}/coqc to check if coqc exists. This does not work with COQBIN in MFMT.
 export COQBIN=$RESULT_INSTALLDIR_CFMT/bin/
+# This must be MFMT (C:/) otherwise bignums 68a7a3d7e0b21985913a6c3ee12067f4c5ac4e20 fails
```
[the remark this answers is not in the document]
MSoegtropIMC: I guess the comment is misleading. MFMT should be the default. We build MinGW binaries and these cannot handle cygwin (CFMT) paths. If you give something like /cygdrive/c/bla to a MinGW application, it has no idea this should be C:\bla. So it is expected that paths should be MFMT, otherwise they cannot be given to any of the executables we build (ocamlc, ocamlfind, coqc, ...) . What is more strange is if something has to be CFMT. This typically only happens if the path is added to $PATH or used in other situations where the : hurts. As it looks "where" in cygwin cannot handle MFMT paths.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -1436,8 +1496,157 @@ function make_addon_equations {
...
+    log1 make $MAKE_OPT
+    log1 install_addon_vst
+    build_post
+  fi
+}
+
+# ssreflect
+
+function make_addon_ssreflect {
+  if build_prep https://github.com/math-comp/math-comp/archive mathcomp-1.7.0 tar.gz ; then
```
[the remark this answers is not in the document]
MSoegtropIMC: A historical artifact - I did this before I wrote build_prep_overlay and head some funny effects (which come from mixing versions with my scripts). I thought about changing it, but then had to many real issues switching from 8.8.1 to master. It is no issue to change it, if you want.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -1436,8 +1496,157 @@ function make_addon_equations {
...
+  fi
+}
+
+# Quickchick Randomized Property-Based Testing Plugin for Coq
+
+function make_addon_quickchick {
+  make_addon_ssreflect
+  make_addon_extlib
+  make_addon_simple_io
+  make_ocamlbuild
```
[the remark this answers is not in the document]
MSoegtropIMC: I am experimenting with switching to opam. The main issue with opam is that it is tricky to get an "installation template" and executables with proper built in search paths. For developers I will provide a opam variant, since the path issue only exists for building installers. This will give me some means to judge long term maintainability of an opam solution.

Without using opam I don't see a way to reuse the information. I could extract it from opam files, but I don't think it is really worth the effort.

dev/build/windows/makecoq_mingw.sh
MSoegtropIMC: True. ci-basic-overlay.sh is called by make_addons, so this function only works for addons. This should be documented.

I guess I will document this and also use build_prep_overlay wherever it applies. It will get tomorrow, though.

Zimmi48:
>Well assuming 8.8.2 is taken from master there are quite a few incompatibilities.

I think you missed the point of what a patch-level release is supposed to be so to clear things up, no 8.8.2 is not going to be taken from master but from v8.8 where I've carefully backported a set of documentation improvement and bug fixes that preserves the compatibility (see https://github.com/coq/coq/projects/19#column-2960008). So don't you worry about all these incompatibilities that you discovered in master.

Zimmi48:
>The idea is to add the addons to CI, so that in the future patches are not required. For the time being I think it is a good idea to test the same patches as I supply upstream - even if they are not tested in CI right now I test them at least locally.

Right, but for instance in the case of QuickChick or VST, the projects are already in the CI, so the necessity of an additional patch is weird.

MSoegtropIMC:
> Right, but for instance in the case of QuickChick or VST, the projects are already in the CI, so the necessity of an additional patch is weird.

It is a different thing if something works in place or after installation. Some packages, e.g. VST and AACtactics seem to be mostly used "in place" and not installed. VST doesn't seem to have a make install at all. AAC uses weird path settings which are not appropriate for installed software. The files compile in place but after installation e.g. for AACtactics the Tutorial.v file doesn't work without changes. For Quickchick the issue is that the make file only works for a byte-code compiled Coq, while in the WIndows installer we only compile and install a native version.

At some point we will need to have a post check for the installer in CI, which e.g. checks if stuff really works after installation. Then such effects will be visible in CI.

MSoegtropIMC:
@Zimmi48 : taking your information on 8.8.2 vs master I reviewed this again concentrating on making it very nice for 8.8.x instead of making it half nice on master.

There is one issue: quite a few of the components in the overlay file in the Coq 8.8 branch refer to master branches, a few to component 8.8 branches.

For building an installer I generally need tags / commits, not branches and definitely not master branches. Otherwise one would get a different thing when later recompiling a specific version - like 8.8.1.

There are two ways to fix this:

1.) replace master with an appropriate tag / commit in the overlays files in the Coq 8.8 branch. I know what to put in there, but I don't know how to make a PR for a specific branch.

2.) use specific version in my build files, where I cannot live with what is in the overlay file.

I would prefer 1.), but I am flexible. I am currently doing 2.) adding comments in my file on why I don't use the overlay file data when I don't, but this got so frequent that I started to think about 1.)

What needs to be discussed is what to do with cases where the overlay file points to a 8.8 branch of a component. For CI this is absolutely appropriate - for building installers not. How to solve this? Can we add another variable for such cases to the overlays file which always points to a very specific version? We could call it e.g. XYZ_CI_TAG which is the latest known good tag or commit.

If we do this, we also need to discuss it it makes sense to require that XYZ_CI_BRANCH is a branch and XYZ_CI_TAG is a tag or commit.

MSoegtropIMC:
@Zimmi48 : Option 2.) doesn't really work in a sensible way. I did now carefully select what I think is most reasonable for 8.8.x, but quite a few components don't build with Coq master then. Most notably even Equations does not compile. I selected tag v1.1-8.8 as most reasonable for Coq 8.8.x. This compiles fine with 8.8.1 but not with master.

Is it OK to add the XYZ_CI_TAG fields to the ci_basic_overlays.sh file?

If so, do I need to make a separate PR for 8.8 and master with the appropriate different values?

Zimmi48:
Hi @MSoegtropIMC, thanks for raising these issues.

I agree with your concern: I'd prefer to have only tags or fixed commits in the overlay file in the v8.8 branch. The principal reason why it's not the case already is because our current scripts want a ref that `git clone` can use: branches and tags work but not commits. I'm going to try to propose a change that would provide the required flexibility.

>If so, do I need to make a separate PR for 8.8 and master with the appropriate different values?

Something like that, yes. This is what I did e.g. with #8305 and #8306.

MSoegtropIMC:
@Zimmi48 : for the time being, would it be OK to add the TAG fields where I need them? I think for CI it does make sense to specify a branch - which specifies the latest code for the given Coq branch, and a tag which specifies the "last known good state". CI should be tested against both from time to time, but usually I would use the branches in CI. Releases should always use the tags.

So how about I add the TAG fields to 8.8 and master and we see later how to make use of this in CI? Otherwise I would have to do really ugly hacks to get my stuff nice for 8.8.x and passing CI on master.

Zimmi48:
Let's see what the discussion at #8348 gives but I don't think it would make any sense to add tags for master because they will become outdated very quickly. Tags are good for stable branches like v8.8. If it helps you do things more cleanly, I would recommend changing the base branch of this PR (from master to v8.8), merging in v8.8 first, and then adapting when forward-porting to master (a bit unusual as a workflow but OK for infrastructure changes I believe).

MSoegtropIMC:
I agree that if one applies a certain rigor to it, it also works if this information is kept in different branches (master and release branches).

One question: I also want to get this PR into 8.8.2. Would it be OK that in master only the addons enabled in master CI would compile (bignums, equations, ltac2 and likely mtac2). I think I could manage this somehow without disturbing what I want to have in 8.8.2. Some other addons would have to be fixed later. In case you get your PR in soon, I could also rebase on it, but this would then have to happen some time soon if 8.8.2 is next week.

Zimmi48:
>One question: I also want to get this PR into 8.8.2. Would it be OK that in master only the addons enabled in master CI would compile (bignums, equations, ltac2 and likely mtac2).

Definitely OK yes.

Zimmi48:
@MSoegtropIMC I didn't understand what you meant by:

>I could attach my tested makecoq_mingw.sh to my PR (I didn't push it as yet because some minor hacks would be required to get it pass the master CI).

MSoegtropIMC:
I have a version of the script which explicitly picks tags or commit hashes for all addons I build. This is tested with both the 8.8.1 tag and the 8.8 branch. Quite a bit of this (I think even all) is still on branches (master or other branches) in the Coq 8.8 branch. So at some point you need to make a commit to the 8.8 branch which picks tags - unless I misunderstood your plans.

Since I have a set of well tested tags for the addons I build, I thought you might want to reuse these. I will attach the file here in a few minutes (I have this on a different PC).

Zimmi48:
Perfect, thank you very much!

MSoegtropIMC:
P.S.: of cause I can also prepare a PR for 8.8 as soon as #8348 is merged.

Zimmi48:
>I can also prepare a PR for 8.8 as soon as #8348 is merged

...and backported. Indeed, this would probably be the simplest solution.

MSoegtropIMC:
> ...and backported. Indeed, this would probably be the simplest solution.

Ok, let's do it this way. Could you already have a look at the file attached above if you have any objections against my picks in the make_addon functions? If you want to change something, I could already test it in parallel.

Zimmi48:
OK, here we go.

- For Ltac2, you point to the commit ppedrot/ltac2@49736b7a1acc23cda6c869645f4a8755ca1b8b65 but this commit has also a tag name: `0.1`. Not a big deal though.
- For CompCert, the only reason version 3.3 doesn't compile with 8.8.1 should be that their Makefile implements a version check. Since you are already passing `-ignore-coq-version` to `./configure`, 3.3 should work fine with 8.8.1. If it's not the case, then it's worrying.
- Regarding VST, I received an e-mail from Andrew Appel saying that they fixed the problem you reported. Isn't it the case? And there still is no install target?
- For aac-tactics and ext-lib you are using `v8.8` but these are branches, not tags or commits.
- For simple-io, you are using the `master` branch instead of a tag or commit.

MSoegtropIMC:
Thanks for the review - looks like I stopped with the cleanup before the last 4 items ... sorry for that.

I will update this, run tests a gain and come back to you.

Zimmi48:
@MSoegtropIMC I just tested the v3.3 tag with 8.8.0, 8.8.1 and the tip of v8.8 and it succeeds in all three cases... FTR I used OCaml 4.06.

ejgallego:
> @Zimmi48 : Regarding CompCert: the 3.3 tag fails with the Coq v8.8 branch with the below (about 35% through).

This is due to OCaml 4.07.0, not Coq.

Zimmi48:
>This is due to OCaml 4.07.0, not Coq.

Good to know, thanks.

MSoegtropIMC:
Regarding CompCert: I guess it is best then to use the commit hash I picked, which works both with 4.06 and 4.07.

ejgallego:
Note that compcert is not free software so we should be careful about distribution

Zimmi48:
Agreed.

BTW, #8348 has been merged and backported so you can start working on the v8.8 version of this PR whenever you want. I suggest you do the pinning in a preliminary, separate commit (but in the same PR).

Zimmi48:
>Note that compcert is not free software so we should be careful about distribution

Good point. Distribution is allowed though, but we should be careful to ship the license with it:

https://github.com/AbsInt/CompCert/blob/b4b1fa3d47c37829be30367ced9aa89abcd7f5c7/LICENSE#L103-L107

There should probably be a mention that external packages that are provided with the Coq installer come under their own licenses.

MSoegtropIMC:
@ejgallego : Regarding compcert license and distribution with "extended Coq": yes, I know. Still it is open source and it is encouraged to be tested even by industrial users. Of cause the license file should be at a prominent place. Maybe we should get @xavierleroy 's opinion on it. Originally I only wanted to include the GPL part (clightgen and the interface files used by VST), but I found that there is no dedicated install for that.

Zimmi48:
>Still it is open source

The code is not open source according to the official [*Open Source Definition*](https://en.wikipedia.org/wiki/The_Open_Source_Definition), although I suppose you meant "the code is open for people to see".

>Certainly there should be a green light from CompCert copyright holders before it is officially shipped.

Anyways, IIRC the goal was to setup the infrastructure to make it possible to build CompCert in the Windows package but not have it shipped by default, right? We'll keep the shipped packages down to a reasonable set of small plugins (bignums, equations, ltac2, mtac2, aac-tactics?, quickchick?) until we have a bigger Windows infrastructure.

MSoegtropIMC:
As I said, I want to ship clightgen, which, as far as I understand (need to double check), is GPL plus other parts of CompCert used by VST which are GPL. Afaik VST only uses the GPL parts of CompCert.

Without clighgen, VST can only be used for the provided exercises. Everybody who wants to try VST on his own code needs CompCert clightgen.

Zimmi48:
>Afaik VST only uses the GPL parts of CompCert.

Not exactly: https://github.com/PrincetonUniversity/VST/blob/4da6fdcc7d67a0c016956c7a8b606a7a774f76ff/LICENSE#L11-L14

xavierleroy:
Concerning CompCert's licensing, you're right that it is neither free software nor open source.  The publically-available sources are usable for educational, research or evaluation purposes only (with the exception of some library modules that are also dual-licensed under the GPL).  This doesn't prevent CompCert from being included in a software distribution, e.g. a package collection, but end users need to be aware of CompCert's licensing conditions.  For instance, it's better if CompCert is not installed by default but needs user action.

Concerning VST and CompCert: the original intent was that the Coq development of VST uses only CompCert Coq modules that are dual-licensed (GPL + noncommercial license).  Maybe this is no longer true but possibly it could be arranged.  However, the clightgen tool is not free software and has the same usage restrictions as the whole CompCert compiler, just because clightgen reuses two compilation passes of CompCert.  A commercial version of clightgen, with technical support and unrestricted uses, is available from AbsInt.

MSoegtropIMC:
@xavierleroy : thanks for the clarification - especially regarding clightgen I got it wrong then. The misunderstanding comes from these lines in the CompCert (3.3) license file:
```
The following files in this distribution are dual-licensed both under
the INRIA Non-Commercial License Agreement and under the Free Software
Foundation GNU General Public License, either version 2 or (at your
option) any later version:
:
:
  cfrontend/C2C.ml
  cfrontend/Clight.v
  cfrontend/ClightBigstep.v
  cfrontend/Cop.v
  cfrontend/CPragmas.ml
  cfrontend/Csem.v
  cfrontend/Cstrategy.v
  cfrontend/Csyntax.v
  cfrontend/Ctypes.v
  cfrontend/Ctyping.v
  cfrontend/PrintClight.ml
  cfrontend/PrintCsyntax.ml
  
  backend/Cminor.v
  backend/PrintCminor.ml

  all files in the cparser/ directory

  all files in the exportclight/ directory
```
Without digging into the details I assumed that clightgen is exactly "cparser + cfrontend + cminor backend + exportclight". I hope you see how this can lead to the wrong conclusion regarding clightgen I came to. I think you should clarify this in the license file.

Would you agree to include CompCert (or possibly just clightgen and the files required by VST) into the Coq extended installer under these conditions:

- It is a separate component which is by default not selected.
- If the CompCert component is selected, the CompCert license file is shown in the installer before installation and the user has to click an "Agree" button to continue with the installation.

MSoegtropIMC:
P.S.: Of cause I would do the same "license accept" to VST since recent version do include non GPL files.

Since this is a rather recent change (the change in the VST license file linked by @Zimmi48 is 4 days old) an alternative would be to install the VST version from before including the new files (about 4 weeks back), which works fine with Coq 8.8. I would have to double check the license.

Btw.: a "make install-gpl" in CompCert would make my live packaging things together a bit easier.

MSoegtropIMC:
@Zimmi48 : I wonder what the next steps are now. I guess the most reasonable procedure would be:

1.) finish this PR so that it uses whatever version it gets from ci_basic_overlay.sh (usually a master branch)
2.) do as little changes as possible to that the addons included in CI do build
3.) merge this PR to master
4.) prepare another PR targeting v8.8 which includes changes to ci_basic_overlay.sh picking the discussed and tested tags and hashes and possibly undoing any hacks done in 2.) above
5.) merge the other PR to v8.8

Is this the correct procedure? I thought about rebasing this PR to v8.8, but came to the conclusion that this would be a bad idea.

Regarding VST: would you prefer to include an older version which only includes GPL files or should I do a "dispaly and accept license" for VST in the installer? I understood Xavier so that the new files VST is using now (I think this has mostly to do with 64 bit porting) might be GPLed as well. So we can wait how this discussion ends or do the license displaying right away.

Zimmi48:
>Is this the correct procedure? I thought about rebasing this PR to v8.8, but came to the conclusion that this would be a bad idea.

The procedure you propose would be OK, but I guess I would prefer making this PR target v8.8 first since this is what we want to see working first.

Regarding VST, I am fine pining an older version for v8.8 and we'll see how the licensing discussion evolves for 8.9.

xavierleroy:
> Without digging into the details I assumed that clightgen is exactly "cparser + cfrontend + cminor backend + exportclight".

That's very close.  The backend/ subdirectory is not used at all because clightgen stops at Clight, not Cminor.  However, not all files in cfrontend/ are dual-licensed, and clightgen uses some of those non-dual-licensed files, in an essential way...

I agree this licensing stuff is complicated and error prone.

> Would you agree to include CompCert (or possibly just clightgen and the files required by VST) into the Coq extended installer under these conditions:
>    It is a separate component which is by default not selected.
>   If the CompCert component is selected, the CompCert license file is shown in the installer before installation and the user has to click an "Agree" button to continue with the installation.

Yes. that's fine.  Thanks!

MSoegtropIMC:
> The procedure you propose would be OK, but I guess I would prefer making this PR target v8.8 first since this is what we want to see working first.

OK, but I will create a new branch based on v8.8 and a new PR then. Rebasing this PR to v8.8 might get tricky cause of the many changes on makecoq_mingw.sh from different sites. So I do as stated above, bit I will exchange the order (first finish v8.8 and then master).

Zimmi48:
>OK, but I will create a new branch based on v8.8 and a new PR then. Rebasing this PR to v8.8 might get tricky cause of the many changes on makecoq_mingw.sh from different sites.

Yes, that's fine, although if you want to keep it in this PR you can also do a hard reset of this branch:

```bash
# working on branch-foo-v8.8
git checkout msoegtrop-dsss2018
git branch msoegtrop-dsss2018-save # save the state of this branch in case it's useful
git reset --hard branch-foo-v8.8
git push -f
```

But do as you prefer.

MSoegtropIMC:
@Zimmi48 : time to continue work here ...

I thought about how to handle build issues with master branches of addons:

1.) add CI for all addons still not having separate CI (if any - didn't check)
2.) If the coq master Linux based CI fails with addon master, adjust the ci-basic-overlays.sh file to a version which does work
3.) If just the windows CI fails with addon master either override the version somewhere or if no version builds with master, provide a patch
4.) Adjust the windows build system such that patches can either include a version or just the base name (I guess the aac patch would work for pretty much any version until this is fixed).

The override could either be done at the beginning of makecoq_mingw.sh or in a separate file, say ci-windows-overlays.sh.

Is this agreeable or would you do it in another way?

Zimmi48:
1 and 2 sound good. Regarding 3 and 4, I would very much prefer if we could do otherwise: if you can work out a patch, then submit a PR upstream. If upstream doesn't want to merge the patch (which should be very rare), then we can discuss what strategy to adopt.

I also have concerns with the Windows job becoming too long on master (which I didn't have on v8.8 because it is run much less frequently). There are other concerns regarding the Windows CI anyways so I think we should plan a conference-call with @ejgallego and @maximedenes...

MSoegtropIMC:
@Zimmi48 : regarding run time: I have a few plans:

1.) Reuse addons which just contain .vo files from the Linux CI and just build plugin addons. The majority of the build time goes into library addons, not into the plugins.
2.) Don't upload the open source archive except for release builds. The upload really takes long and it is fairly useless. This is already implemented in v8.8.
3.) Possibly run a reduced addon version in pre-commit quality gate tests and do e.g. a daily extended build. What is in where should depend on popularity and breaking rate of the addon.
4.) Possibly for PCQG also avoid uploading the installer (it also takes 15 minutes). I think it is sufficient to upload the installer daily.
5.) In a future PR switch to downloadable addons. I am not sure if this is doable with NSIS. A switch to WIX would be a major effort. In my opinion WIX is more reasonable than NSIS, but also substantially more complicated. But meanwhile I get along with NSIS - I still don't like it, but I can live with it and can get things done.

Sure we can have a conference call on the topic. Just drop me a mail when you have time.

dev/ci/ci-basic-overlay.sh
```diff
@@ -226,3 +226,18 @@
...
+########################################################################
+: "${menhirlib_CI_REF:=master}"
+: "${menhirlib_CI_GITURL:=https://gitlab.inria.fr/fpottier/coq-menhirlib}"
+: "${menhirlib_CI_ARCHIVEURL:=${menhirlib_CI_GITURL}/-/archive}"
+
+########################################################################
+# aac-tactics
+########################################################################
+# Note: this is the latest commit of the v8.8 branch as of August 31st 2018
+: "${aactactis_CI_REF:=master}"
```
SkySkimmer: Is the comment a survivor from a 8.8-only PR?

MSoegtropIMC: @SkySkimmer : yes, thanks!

Please note that this PR is not yet ready for review. I still need some experimenting what works in CI and what not (which is unfortunately not the same as what works locally - something I am researching). There are still a few things missing, among others a final cleanup, but also the dependency checking.

Maybe we should have a "work in progress" label ... I really don't know how to handle these CI centric PRs in a good way.

SkySkimmer: >Maybe we should have a "work in progress" label ... I really don't know how to handle these CI centric PRs in a good way.

See `needs: progress` which I just put on this PR.

Zimmi48:
I see that this PR still includes a patch to QuickChick and you even had to workaround the fact that a patch should normally be associated to a specific version but we are tracking a branch in master. How hard would it be to submit a PR to QuickChick developers instead as I've repeatedly asked for? I don't think that I'm ready to accept this PR with a new patch in master. If I did accept patches in 8.8.2, it was because it was tied to a specific version and (more specifically) because there was little time.

Zimmi48:
Oh I had forgotten that you had already replied on this topic:

>For QuickChick I got feedback that they take my changes (the issue here is that they buid things so that they are only compatible with a byte code version of Coq).

So how is this going on? Why isn't the patch in yet? Probably if you had gone through a PR (https://github.com/QuickChick/QuickChick/pulls) it would have been quicker.

MSoegtropIMC:
@Zimmi48 : regarding QuickChick: one of the things I had to patch is already in QuickChick master and I removed it from the patch.
For the native vs byte build: I need to discuss it with the QuickChick team again. They said the change is reasonable, but it might be that it makes it harder for them to debug their code. I guess the makefile needs a bit of a larger rework so that one can choose to make a bytecode or native build from the command line. My patch simply switches from bytecode to native. Since the Windows Coq is a native build, we need native plugins.
I would think this all needs a bit of time. I guess they will see the advantage of having this in the Coq installer and of having "proper" make files and will support us with it. It just might take a short while.

MSoegtropIMC:
From my point of view, this PR is finished now - please do the reviews (I think I fixed all previous review comments).

This is what is in 8.8.2 with these differences:
- VST and AAC disabled cause of build issues (I will investigate later)
- all addons are on master except cases where this doesn't work easily (e.g. coquelicot, see comments in code)
- I add functionality to have patches without version number/hash because specific patches obviously don't work with branches since branches are translated to hashes
- I left the function to sed timeouts to large values in, although this is currently not used since VST is disabled. I might need this later

Zimmi48 (commented):
Looks pretty good, thanks for the work you put into this Michael!

But why don't all the addons have `installer_addon_section`?

Regarding the QuickChick patch, it is small enough so I'm OK to merge this as-is but we should look into fixing this upstream and removing the patch.

Zimmi48:
You told me you were pretty sure that this wouldn't be completely straightforward to backport. If that's the case, please put this in the 8.10+beta1 milestone and then take care of opening the backport PR to v8.9. If that's not the case, then you can put the 8.9+beta1 milestone.

MSoegtropIMC:
@Zimmi48 : thanks for the review!

I guess with the missing installer sections you refer to matc2 and unicoq - this is indeed something I overlooked during merging. I will fix it (plus the 2 typos).

Regarding 8.9 vs 8.10: I actually don't know how easy it will be in 8.9. It might or might not work out of the box. When do you need to know? Thursday evening I should know more.

MSoegtropIMC:
I fixed the mtac2 and unicoq installation and tested the installer. Sorry that I completely lost mtac-2 from my radar when testing the installer before.

I tried successfully the following with CoqIde installed with the installer from the gitlab aritafcts:

```
Require Import Bignums.BigNumPrelude.
Require Import Coquelicot.Coquelicot.
Require Import Equations.Equations.
Require Import ExtLib.ExtLib.
Require Import Ltac2.Ltac2.
Require Import Mtac2.Mtac2.
Require Import mathcomp.ssreflect.all_ssreflect.
Require Import MenhirLib.Main.
Require Import QuickChick.QuickChick.
Require Import SimpleIO.CoqPervasives.
Require Import Unicoq.Unicoq.
```

Some produce warnings mostly on operator parsing inconsistencies and I have to load Mtac2 before mathcomponents for whatever reason. But basically it seems to work and I guess few people use all this together.

I guess I can ignore the bedrock failure, so I would declare the PR as "ready to merge".

Zimmi48 (approved):
Looks great, thanks!

PR #8326, opened 2018-08-26
Mention PRINT_LOGS=1 when failing test suite

SkySkimmer:


test-suite/report.sh
```diff
@@ -0,0 +1,55 @@
...
+        fi
+    done
+fi
+
+if grep -q -F 'Error!' summary.log ; then
+    echo FAILURES;
+    grep -F 'Error!' summary.log;
+    if [ -z "$TRAVIS" ] && [ -z "$PRINT_LOGS" ]; then
+        echo 'To print details of failed tests, rerun with environment variable PRINT_LOGS=1'
+        echo 'eg "make PRINT_LOGS=1" from the test suite directory"'
```
Zimmi48: Maybe advise more strongly to rerun by using `make` from the test-suite directory, because if people rerun with `make test-suite PRINT_LOGS=1` they will have the bad surprise to see the full test-suite being run again.

SkySkimmer: I changed it a bit. There's only so much we can protect people against our own system though.

gares (approved):

Zimmi48:
OK

PR #8333, opened 2018-08-27
Fix URIs in the configuration script

pi8027:
**Kind:** bug fix.

The naming conversion of documentation URIs seems to be changed since V8.8+beta1 (see: https://coq.inria.fr/distrib/). Currently `configure.ml` doesn't follow it and some documentation of external libraries point wrong URIs (e.g., `plus` and `minus` in http://math-comp.github.io/math-comp/htmldoc/mathcomp.ssreflect.ssrnat.html).


ejgallego (approved):
Well-spotted indeed, the current system is a mess.

PR #8340, opened 2018-08-28
Put camldevfiles targets in Makefile

SkySkimmer:
There's no need to build dependencies for it.


ejgallego (approved):
Nice, thanks!

gares (approved):

JasonGross:
This broke the `states` target, which lives in `Makefile.dev`, and depends on `camldevfiles`, because `Makefile.dev` does not include `Makefile`.  This breaks HoTT's CI, but was not caught by Coq's CI, because nothing on Coq's CI tests `make states`.

PR #8341, opened 2018-08-28
Reduce universe redeclarations (catching AlreadyDeclared)

SkySkimmer:
NB: Include is still causing repeat declarations.

### DIFF
--- a/kernel/term_typing.ml
+++ b/kernel/term_typing.ml
@@ -256,6 +256,8 @@ let infer_declaration (type a) ~(trust : a trust) env (dcl : a constant_entry) =
       let tyj = infer_type env typ in
       let proofterm =
         Future.chain body (fun ((body,uctx),side_eff) ->
+          (* don't redeclare universes which are declared for the type *)
+          let uctx = Univ.ContextSet.diff uctx univs in
           let j, uctx = match trust with
           | Pure ->
             let env = push_context_set uctx env in
--- a/CHANGES
+++ b/CHANGES
@@ -205,6 +205,12 @@ Tools
   and `print-pretty-single-time-diff` now correctly label the "before" and
   "after" columns, rather than swapping them.
 
+Kernel
+
+- The kernel does not tolerate capture of global universes by
+  polymorphic universe binders, fixing a soundness break (triggered
+  only through custom plugins)
+
 Changes from 8.8.0 to 8.8.1
 ===========================
 
--- a/dev/doc/critical-bugs
+++ b/dev/doc/critical-bugs
@@ -109,6 +109,16 @@ Universes
   GH issue number: none
   risk: unlikely to be activated by chance
 
+  component: universe polymorphism
+  summary: universe polymorphism can capture global universes
+  impacted released versions: V8.5 to V8.8
+  impacted coqchk versions: V8.5 to current (NOT FIXED)
+  fixed in: 2385b5c1ef
+  found by: Gaëtan Gilbert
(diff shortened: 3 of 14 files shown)


kernel/term_typing.ml
```diff
@@ -267,6 +267,8 @@ let infer_declaration (type a) ~(trust : a trust) env (dcl : a constant_entry) =
       let tyj = infer_type env typ in
       let proofterm =
         Future.chain body (fun ((body,uctx),side_eff) ->
+          (* don't redeclare universes which are declared for the type *)
+          let uctx = Univ.ContextSet.diff uctx univs in
```
SkySkimmer: Not sure if this is the right place to do this, but it seemed cleaner than doing it in the proof_global function and it makes OK sense to allow the redundancy in the entry.

SkySkimmer:
BTW allowing redeclarations for monomorphic definitions is fine but for polymorphic definitions it seems inconsistent, allowing to capture global universes.
ie a plugin could produce something like
~~~coq
Universe i.
Definition Ti := Type@{i}.
Polymorphic Definition T@{i (* this is the same as the global i *)|(*no constraints*)} : Ti -> Type@{i} := fun x => x.
Check T@{j}. (* T@{j} : Ti -> Type@{j} (* j |= *) *)
(* !!! the substitution of j doesn't go through Ti *)
~~~
I haven't tried to make such a plugin so maybe something prevents it, but it's definitely dangerous.

After this PR it would produce an anomaly. However the checker may still allow it (something for a future PR) (to trick the checker you may need to use `Var 0` as your global universe since the checker sees universe-abstracted terms).

SkySkimmer:
Inconsistency from plugin confirmed, see test.

SkySkimmer:
Not sure if we want to actually include the test, testing for inconsistencies through plugins is a bit complex. And testing inconsistencies in the checker is essentially impossible.

ejgallego:
> Not sure if we want to actually include the test, testing for inconsistencies through plugins is a bit complex.

Why?

JasonGross (dismissed):
This merits a line in dev/doc/critical-bugs, right?

SkySkimmer:
Only 12204ee and 2385b5c are bug related, the others are just cleanups.

Zimmi48 (approved):
Approving the doc part.

mattam82 (approved):
The fix looks good to me. I just think we should go further and disallow Level.Var to appear in monomorphic constant or inductive contexts altogether. @SkySkimmer is Include the only remaining case where redeclarations happen now?

SkySkimmer:
>The fix looks good to me. I just think we should go further and disallow Level.Var to appear in monomorphic constant or inductive contexts altogether. 

Next PR I guess.

>@SkySkimmer is Include the only remaining case where redeclarations happen now?

I don't know.

Zimmi48:
@SkySkimmer Shouldn't this come with a CHANGES entry? If yes, add it to the 8.8.2 section in both this PR and #8360.

SkySkimmer:
Added changes

Zimmi48:
Great thanks! @maximedenes you're the only code owner whose review is missing from this (critical) PR.

maximedenes (commented):
I declare myself incompetent.

Zimmi48:
@mattam82 ping

silene:
If I am not mistaken, this is already part of 8.9, so please complain if I am wrongly removing from the project.

PR #8348, opened 2018-08-28
Download tarball instead of cloning external projects.

Zimmi48:
Downloading tarballs instead of cloning using git allows to use commits as references and not just branches or tags. This will be especially useful on stable branches: testing the master branch of an external project is fragile because upstream maintainers can decide suddenly to stop supporting an older version of Coq. In the past, I've tried mitigating this by using tags instead but tags do not always exist. That's why CoLoR testing is currently disabled on `v8.7`:

https://github.com/coq/coq/blob/81ab3a7d50a4fd18476074822a8189ab0fc370b8/.gitlab-ci.yml#L224-L225

We keep using git clone for projects on gforge.inria.fr and for projects pulling dependencies through git submodules.

**Kind:** infrastructure.


SkySkimmer (commented):
Seems useful enough.

Some failures because you partially ported git_checkout calls to use the archive url but not using the git_download function.

Zimmi48:
Thanks for the review @SkySkimmer!

> What is --strip? My tar manpage doesn't have it.

I think `--strip` and `--strip-components` are the same but given that only the latter appears in my manual to, I've replaced it.

MSoegtropIMC (commented):
I thought about if it would make sense to have separate entries for the last know good tag/commit hash and the proper branch for each component. If you have only one, you need to change a lot of things when making a label which I think gets difficult. In the end as long as something is a branch, the REFs should be branches. If something is a tag, teh REFs should be tags as well. I don't see how you want to manage this in GIT. So I think it is easier to have xyz_CI_BRANCH which always points to the appropriate branch (say master in the master branch and v8.8 in the v8.8 branch) and xyz_CI_TAG which always points to the last known good tag.

Then when you build, you have to know if you build a release (which uses xyz_CI_TAG) or do CI for a branch (which uses xyz_CI_BRANCH) and one can even do both in CI from time to time.

In case you think only one is enough, I would like to understand how you envision to handle this in GIT.

ejgallego (commented):
I am not sure I like this; first, it is github specific and may not work for all repositories, which seems like an important downside to me; I'd rather keep the branch scheme and add a `_COMMIT` variable, and have the checkout script use infinite depth if the commit is requested.

Also, it seems hard to know whether this scheme will suffice when the pinning mechanism has not been proposed yet [and it is far from trivial I think]

On a lesser inconvenient is that this breaks a workflow for overlays, having a git repository in `_ci` at hand is good for quick patches; and indeed we should/may take advantage of a compositional build in the future to build the ci stuff.

Zimmi48:
Thanks for the comments @MSoegtropIMC and @ejgallego!

Note that this is just infrastructure that makes it *possible* to pin but no pinning is done in this PR. My experience as a stable branch manager makes me strongly believe that pinning on stable branches is very useful. On master, it is another debate and I don't want to venture into it just now.

Note also that this is not GitHub specific, as it also supports GitLab and BitBucket. I think it is reasonable to expect that projects that are added in CI should use one of these three platforms (self-hosted instances included).

>If you have only one, you need to change a lot of things when making a label which I think gets difficult.

I don't understand what you mean here: for the time being I plan to keep branches (no tags / no fixed commits) on master and fix commits or use tags when available on stable branches (but they don't have to be changed anymore once they are fixed). If we still want to be able to update some of them on stable branches, we could keep the information about the corresponding branch in a comment.

>I would like to understand how you envision to handle this in GIT.

I don't understand the question.

>On a lesser inconvenient is that this breaks a workflow for overlays, having a git repository in _ci at hand is good for quick patches

I don't get how this breaks the workflow for overlays. My claim in this PR is that it doesn't change anything (except having to also override the `ARCHIVEURL` variable *if* the fork is not hosted on the same platform as the upstream project).

Zimmi48:
>I mean it breaks the workflow: a) make a change to coq; b) make ci-all c) go under _build_ci and fix the broken plugins.

I've tried very hard to not break this workflow (and my tests seem to confirm that I've succeeded). If you see anything problematic, I'll fix it.

ejgallego:
> I've tried very hard to not break this workflow (and my tests seem to confirm that I've succeeded). If you see anything problematic, I'll fix it.

I guess I am missing something, do you mean that there still is a `.git` repos under `_build_ci/$FOO/.git` ?

Zimmi48:
Ah right, sorry I missed the info on part of your workflow. Let me think about it and find a way to give some choice between tarballs and git. I'd prefer to use tarballs on the CI to save bandwidth and time but it would make sense to do a (deep) git clone when working locally.

ejgallego:
> Ah right, sorry I missed the info on part of your workflow. Let me think about it and find a way to give some choice between tarballs and git. I'd prefer to use tarballs on the CI to save bandwidth and time but it would make sense to do a (deep) git clone when working locally.

Yeah, note that I am OK with tarballs if you want to go that way, as long as we keep the git repos URL in the overlay file I will survive.

Zimmi48:
Let me spend one hour on this issue. I don't have a lot of time on my hands right now but it would be a shame to not even try. If I don't manage, I would appreciate to have it merged nonetheless.

Zimmi48:
I think I've managed. At least, it works locally. Let see if it also works OK on CI.

dev/ci/ci-common.sh
```diff
@@ -46,6 +46,40 @@ done
...
+    cd "$DEST"
+    wget "$ARCHIVEURL/$REF.tar.gz"
+    tar xvfz "$REF.tar.gz" --strip-components=1
+    rm -f "$REF.tar.gz"
+  else # When on local machine, we clone git repo to help prepare overlays
+    local GITURL_VAR="${PROJECT}_CI_GITURL"
+    local GITURL="${!GITURL_VAR}"
+    local REF_VAR="${PROJECT}_CI_REF"
+    local REF="${!REF_VAR}"
+    git clone "$GITURL" "$DEST"
```
SkySkimmer: Could you replicate the stuff in `git_checkout` where it still works if the directory already exists?
Or perhaps even call the `git_checkout` function here.

ejgallego (approved):
Looks great, and adds a much needed refactoring to `git_download`.

dev/ci/ci-common.sh
Zimmi48: I've considered this but my past experience with the way `git_checkout` behaves is far from satisfying so that's why I've preferred to aim for a simpler approach: if the directory already exists either delete it or manually checkout the right branch yourself.
It will also avoid any problem if the data was previously fetched as a tarball (if you had defined `$CI`).
Does this sound OK?

SkySkimmer: > Does this sound OK?

No, it means I can't do `make ci-foo` after fixing some file in foo to see if it worked.

ejgallego: > my past experience with the way git_checkout behaves is far from satisfying

Can you say more? A key goal I have is actually remove all custom build rules for ci projects as they move to Dune [then for example if mathcomp depends on coq, dune will use the local coq and track all the global deps]

So IMHO it is important to support @SkySkimmer workflow.

Zimmi48: OK I think I had misunderstood @SkySkimmer's initial comment so answering differently:
`git_download` will not fail if the directory already exists, it will just print a warning and skip directly to the build phase. So your workflow (a very important one indeed) is fully supported.

dev/ci/ci-common.sh
```diff
@@ -46,6 +46,40 @@ done
...
+# if it does, it will do nothing except print a warning
+# (this can be useful when testing make ci-* locally).
+# When the variable $CI is undefined, it uses git clone to perform the
+# download.
+git_download()
+{
+  local PROJECT=$1
+  local DEST=$2
+
+  if [ -d "$DEST" ]; then
```
SkySkimmer: Oh I see there's this bit

Zimmi48: This was inspired by your first review comment about `mkdir -p` succeeding too often.

Zimmi48:
So, does everything look good then? If yes, then I suggest to not delay the merge to allow @MSoegtropIMC to rebase #8322 on top.

ejgallego:
> So, does everything look good then? If yes, then I suggest to not delay the merge to allow @MSoegtropIMC to rebase #8322 on top.

Umm, why does github say that only Michael is the owner of the files in this PR?

ejgallego:
I assume @coq/ci-maintainers are the owners of this PR, despite the review panel missing the shield.

ejgallego:
@SkySkimmer will wait for your OK.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -1421,12 +1421,12 @@ function make_addon_ltac2 {
...
-      # $Equations_CI_BRANCH must have been a tag and not a branch
-      Equations_SHA="$Equations_CI_BRANCH"
+      # $Equations_CI_REF must have been a tag / commit and not a branch
+      Equations_SHA="$Equations_CI_REF"
   fi
-  if build_prep "${Equations_CI_GITURL}/archive" "$Equations_SHA" zip 1 "Equations-$Equations_SHA"; then
+  if build_prep "$Equations_CI_ARCHIVEURL" "$Equations_SHA" zip 1 "Equations-$Equations_SHA"; then
     # Note: PATH is autmatically saved/restored by build_prep / build_post
     PATH=$COQBIN:$PATH
     logn coq_makefile ${COQBIN}coq_makefile -f _CoqProject -o Makefile
```
ejgallego: The amount of boilerplate in these windows scripts is a bit suboptimal.

dev/ci/ci-bedrock2.sh
```diff
@@ -5,7 +5,7 @@ ci_dir="$(dirname "$0")"
 
 bedrock2_CI_DIR="${CI_BUILD_DIR}/bedrock2"
 
-git_checkout "${bedrock2_CI_BRANCH}" "${bedrock2_CI_GITURL}" "${bedrock2_CI_DIR}"
+git_checkout "${bedrock2_CI_REF}" "${bedrock2_CI_GITURL}" "${bedrock2_CI_DIR}"
```
ejgallego: Ideally we would just say `git_checkout bedrock` , but I lack the require sh-fu to make that happen.

dev/ci/ci-template.sh
```diff
@@ -1,12 +0,0 @@
...
-ci_dir="$(dirname "$0")"
-. "${ci_dir}/ci-common.sh"
-
-Template_CI_BRANCH=master
-Template_CI_GITURL=https://github.com/Template/Template
-Template_CI_DIR="${CI_BUILD_DIR}/Template"
-
-git_checkout "${Template_CI_BRANCH}" "${Template_CI_GITURL}" "${Template_CI_DIR}"
-
-( cd "${Template_CI_DIR}" && make )
```
ejgallego: Why this removal?

dev/ci/ci-common.sh
```diff
@@ -46,6 +46,40 @@ done
...
+    echo "Warning: download and unpacking of $PROJECT skipped because $DEST already exists."
+  elif [ -n "$CI" ]; then # When on CI, we download tarballs
```
SkySkimmer: Perhaps this should be controlled by a variable like `FORCE_GIT` (default value `${CI:+1}` ie "not $CI") then we could set it locally to eg the bedrock script and get rid of git_checkout.

dev/ci/ci-template.sh
SkySkimmer: I guess telling people to look at some existing script is easier to maintain than keeping around this example one.

dev/ci/ci-common.sh
```diff
@@ -46,6 +46,40 @@ done
...
+# [git_download project dest] will download the given [ref] of [project]
+# and unpack it in [dest] if the folder does not exist already;
+# if it does, it will do nothing except print a warning
+# (this can be useful when testing make ci-* locally).
+# When the variable $CI is undefined, it uses git clone to perform the
+# download.
+git_download()
+{
+  local PROJECT=$1
+  local DEST=$2
```
SkySkimmer: Is there some reason to have `DEST` different from `PROJECT`?

MSoegtropIMC:
> So, does everything look good then? If yes, then I suggest to not delay the merge to allow @MSoegtropIMC to rebase #8322, Additional addons for Windows installer on top.

@Zimmi48 : thanks for caring, but under the agreement that my checking at first only hast to be perfect for 8.8.2 and for master only the things checked in CI need to work, I don't depend on this. Of cause I mid term depend on it to make all addons work both in master and in the release branch.

dev/build/windows/makecoq_mingw.sh
Zimmi48: This boilerplate is going away in @MSoegtropIMC's PR.

dev/ci/ci-template.sh
Zimmi48: Yes

dev/ci/ci-bedrock2.sh
Zimmi48: I can make this happen in a subsequent PR.

dev/ci/ci-common.sh
Zimmi48: Good idea. I'll have a try.

dev/ci/ci-common.sh
Zimmi48: Nope, I considered removing it but this means that:

https://github.com/coq/coq/blob/7ca59e2db97d78c511856375991400e44a6c1934/dev/ci/ci-bignums.sh#L12-L16

would become

```shell
git_download bignums

(cd "${CI_BUILD_DIR}/bignums" && make && make install)
```

Does this look good? If yes, I can do it.

SkySkimmer: The capitalisation of some likely-temporary directory doesn't seem important.

Zimmi48:
>under the agreement that my checking at first only hast to be perfect for 8.8.2 and for master only the things checked in CI need to work, I don't depend on this

This PR is required to be able to pin commits for v8.8 (currently we can only pin tags and most of the time appropriate tags don't exist).

MSoegtropIMC:
> This PR is required to be able to pin commits for v8.8 (currently we can only pin tags and most of the time appropriate tags don't exist).

For the time being I have explicit versions in my script. I will change it as soon as this PR plus the PR which pins the versions for 8.8.2 is in (you can take the versions from my PR then).

If you think it is timewise feasible for 8.8.2, I can also wait for the two PRs, but I don't see why we should go through this - it is easy to change afterwards.

dev/ci/ci-common.sh
Zimmi48: Ah yes, I was not really asking about the capitalization, more about the rest. But it seems you find it OK, so I will go for it.

Zimmi48:
>If you think it is timewise feasible for 8.8.2, I can also wait for the two PRs, but I don't see why we should go through this - it is easy to change afterwards.

I'd rather have a solution that is clean from the beginning if we can. Timewise, it shouldn't a problem. 8.8.2 won't definitely happen before next week (there is no commitment to a specific date as far as 8.8.2 is concerned). I'm also waiting to see if #8341 should be backported.

MSoegtropIMC:
> Timewise, it shouldn't a problem. 8.8.2 won't definitely happen before next week 

Fine, then I wait.

Shall I send you the versions I picked (tested with 8.8.1 tag and 8.8 branch) ? I could attach my tested makecoq_mingw.sh to my PR (I didn't push it as yet because some minor hacks would be required to get it pass the master CI).

Is there already a PR to do the picks for 8.8.2?

dev/ci/ci-common.sh
```diff
@@ -44,29 +44,39 @@ for overlay in "${ci_dir}"/user-overlays/*.sh; do
...
-  mkdir -p "${_DEST}"
-  ( cd "${_DEST}"                                                     && \
-    if [ ! -d .git ] ; then git clone "${_DEPTH[@]}" "${_URL}" . ; fi && \
-    echo "Checking out ${_DEST}"                                      && \
-    git fetch "${_URL}" "${_BRANCH}"                                  && \
-    git -c advice.detachedHead=false checkout "${_COMMIT}"            && \
-    echo "${_DEST}: $(git log -1 --format='%s | %H | %cd | %aN')" )
+  local PROJECT=$1
+  local DEST="$CI_BUILD_DIR/$PROJECT"
+  : "${FORCE_GIT:=${CI:+1}}" # FORCE_GIT default value is 1 if $CI is undefined
```
Zimmi48: This doesn't work as intended. @SkySkimmer do you know how to fix this?

SkySkimmer: Indeed I misread the doc
Just use an `if` then.

Zimmi48: OK ^^

Zimmi48:
>Shall I send you the versions I picked (tested with 8.8.1 tag and 8.8 branch) ? I could attach my tested makecoq_mingw.sh to my PR (I didn't push it as yet because some minor hacks would be required to get it pass the master CI).

Let's switch back this discussion to #8322.

dev/ci/ci-fiat-parsers.sh
```diff
@@ -3,8 +3,6 @@
 ci_dir="$(dirname "$0")"
 . "${ci_dir}/ci-common.sh"
 
-fiat_parsers_CI_DIR="${CI_BUILD_DIR}/fiat"
+git_download fiat_parsers
```
JasonGross: I believe you're missing `FORCE_GIT=1` here; the fiat-parsers Makefile tries to update / init submodules itself.

dev/ci/ci-fiat-crypto-legacy.sh
```diff
@@ -3,12 +3,11 @@
...
-
-( cd "${fiat_crypto_CI_DIR}" && git submodule update --init --recursive )
+FORCE_GIT=1
+git_download fiat_crypto
 
 fiat_crypto_legacy_CI_TARGETS1="print-old-pipeline-lite old-pipeline-lite lite-display"
 fiat_crypto_legacy_CI_TARGETS2="print-old-pipeline-nobigmem old-pipeline-nobigmem nonautogenerated-specific nonautogenerated-specific-display"
-( cd "${fiat_crypto_CI_DIR}" && make ${fiat_crypto_legacy_CI_TARGETS1} && make -j 1 ${fiat_crypto_legacy_CI_TARGETS2} )
+
+( cd "${CI_BUILD_DIR}/fiat_crypto" && git submodule update --init --recursive && \
```
JasonGross: Do you need the trailing slash inside parentheses in bash?  (I know you don't in python)

JasonGross (dismissed):

dev/ci/ci-fiat-crypto-legacy.sh
Zimmi48: I have no idea.

Zimmi48:
CI results: GitLab CI OK, Travis CI OK but did not report it (weird), AppVeyor times out (unrelated).

ejgallego:
Waiting for @SkySkimmer green flag.

dev/ci/ci-common.sh
```diff
@@ -44,29 +44,41 @@ for overlay in "${ci_dir}"/user-overlays/*.sh; do
...
-    if [ ! -d .git ] ; then git clone "${_DEPTH[@]}" "${_URL}" . ; fi && \
-    echo "Checking out ${_DEST}"                                      && \
-    git fetch "${_URL}" "${_BRANCH}"                                  && \
-    git -c advice.detachedHead=false checkout "${_COMMIT}"            && \
-    echo "${_DEST}: $(git log -1 --format='%s | %H | %cd | %aN')" )
+  local PROJECT=$1
+  local DEST="$CI_BUILD_DIR/$PROJECT"
+  if [[ ! -n "$FORCE_GIT" && ! -n "$CI" ]]; then
+      FORCE_GIT=1
+  fi
```
SkySkimmer: Just use `if [ -n "$FORCE_GIT" ] || [ -n '$CI" ]` in the below condition

SkySkimmer (changes requested):

dev/ci/ci-common.sh
Zimmi48: OK

Zimmi48: Wait this isn't the same, did you mean: `if [ -n "$FORCE_GIT" ] || [ ! -n "$CI" ]`?

SkySkimmer: right
but `-z "$CI"` is cleaner than `!`

Zimmi48: And even this is less powerful: when `$CI` is undefined, setting `FORCE_GIT=0` would allow to download the tarball nonetheless. Whereas with this test, you would have to set `$CI` to do the same.

SkySkimmer:
Don't add that ability until we have some use for it.

Gaëtan Gilbert

Zimmi48:
OK

Zimmi48:
@SkySkimmer Updated. Please confirm that this corresponds to your expectations.

dev/ci/ci-common.sh
```diff
@@ -44,29 +44,38 @@ for overlay in "${ci_dir}"/user-overlays/*.sh; do
...
-    echo "Checking out ${_DEST}"                                      && \
-    git fetch "${_URL}" "${_BRANCH}"                                  && \
-    git -c advice.detachedHead=false checkout "${_COMMIT}"            && \
-    echo "${_DEST}: $(git log -1 --format='%s | %H | %cd | %aN')" )
+  local PROJECT=$1
+  local DEST="$CI_BUILD_DIR/$PROJECT"
+
+  if [ -d "$DEST" ]; then
+    echo "Warning: download and unpacking of $PROJECT skipped because $DEST already exists."
+  elif [ -n "$FORCE_GIT" ] || [ -z "$CI" ]; then
```
SkySkimmer: Actually that's still inverted isn't it?
Maybe we should use `[ "" = "$foo" ]` instead of `[ -z "$foo" ]` / `[ "" != "$foo" ]` instead of `[ -n "$foo" ]` to make it easier to read through the `test` flags.

Zimmi48: Yep.

Zimmi48:
@SkySkimmer Done.

Note that this is now on top of #8368.

SkySkimmer (approved):

Zimmi48:
@ejgallego CI has passed (again): can you merge?

JasonGross:
Was this tested with overlays?

dev/ci/ci-fiat-parsers.sh
JasonGross: Feel free to also update the submodule in the build script.  In any case, I believe I've fixed the silent incorrect success with https://github.com/mit-plv/fiat/commit/d595813799d2f141fcba9c8c66c5308783a4ccb2

Zimmi48:
>Was this tested with overlays?

Oh, I wanted to do it but I forgot. Did you experience any trouble?

Zimmi48:
Oh I see you already reported it at #8348

ejgallego:
@Zimmi48 this PR introduced a non trivial regression: the old CI code had code to output the exact git commit being tested in the CI. This allowed to faithfully reproduce a build locally [modulo the Gitlab Env]

However, now when fetching a `master` branch, no information about the tip of the tree is displayed.

SkySkimmer:
We could probably do a `git ls-remote` to get that information.
@Zimmi48 can you make a PR?

ejgallego:
> We could probably do a git ls-remote to get that information.
> @Zimmi48 can you make a PR?

Should we do both after and before? I am afraid of the race condition, by Murphy's law one day we will get crazy debugging a problem because of a race here.

ejgallego:
> Can you open a bug report and assign to me?

Done thanks.

PR #8349, opened 2018-08-28
Mention dev/doc/critical-bugs in CONTRIBUTING

JasonGross:
**Kind:** infrastructure.


Zimmi48 (changes requested):
Sorry but I think this case is not frequent enough to justify making the template more heavy. Furthermore, critical bug fixes generally come from core developers who should be aware of this file. To raise more awareness towards such expectations (mainly for future contributors) I would recommend adding them to the contributing guide.

JasonGross:
*nods*  This was after noticing that #8341 didn't add an entry.  I've moved it to `CONTRIBUTING.md`.

Zimmi48 (approved):

PR #8356, opened 2018-08-30
String: add reverse function

liyishuai:
**Kind:** feature

- [x] Entry added in CHANGES.

### DIFF
--- a/doc/changelog/10-standard-library/08356-rev_string.rst
+++ b/doc/changelog/10-standard-library/08356-rev_string.rst
@@ -1,0 +1,6 @@
+- Added ``String.rev`` and proved some lemmas about it
+  (`#8356 <https://github.com/coq/coq/pull/8356>`_, by Yishuai Li).
+  Note that this might cause incompatibilities if you have, e.g.,
+  ``List`` and ``String`` both imported with ``String`` on top,
+  and expect ``rev`` to refer to ``List.rev``.
+  Solution: qualify ``rev``.
--- a/theories/Strings/String.v
+++ b/theories/Strings/String.v
@@ -84,6 +84,15 @@ Fixpoint append (s1 s2 : string) : string :=
   end
 where "s1 ++ s2" := (append s1 s2) : string_scope.
 
+Theorem append_EmptyString_l : forall s : string, EmptyString ++ s = s.
+Proof. reflexivity. Qed.
+
+Theorem append_EmptyString_r : forall s : string, s ++ EmptyString = s.
+Proof. induction s; simpl; f_equal; auto. Qed.
+
+Theorem append_assoc : forall s1 s2 s3, s1 ++ s2 ++ s3 = (s1 ++ s2) ++ s3.
+Proof. induction s1; simpl; intros; f_equal; auto. Qed.
+
 (******************************)
 (** Length                    *)
 (******************************)
@@ -152,6 +161,74 @@ intros s2 n; rewrite plus_comm; simpl; auto.
 intros a s1' Rec s2 n; case n; simpl; auto.
 generalize (Rec s2 O); simpl; auto. intros.
 rewrite <- Plus.plus_Snm_nSm; auto.
+Qed.
+
+(** Appending strings sum their lengths *)
+
+Theorem append_length : forall s1 s2 : string,
+    length (s1 ++ s2) = length s1 + length s2.
+Proof. induction s1; simpl; auto. Qed.
+
+(** *** Reverting strings  *)
+
+Fixpoint rev (s : string) : string :=
+  match s with
+  | EmptyString => EmptyString
+  | String c s' => rev s' ++ String c EmptyString
+  end.
+
+Lemma rev_app_distr : forall s1 s2 : string,
+    rev (s1 ++ s2) = rev s2 ++ rev s1.
+Proof with simpl; auto.
+  induction s1 as [| c s IHs1]; intros...
+  - rewrite append_EmptyString_r...
+  - rewrite IHs1.
+    rewrite append_assoc...
+Qed.
+
+Remark rev_unit : forall (s : string) (c : ascii),
+    rev (s ++ String c EmptyString) = String c (rev s).
+Proof. intros; rewrite rev_app_distr; auto. Qed.
+
+Lemma rev_involutive : forall s : string, rev (rev s) = s.
+Proof with simpl; auto.
+  induction s as [| c s IHs]...
+  rewrite rev_unit.
+  rewrite IHs...
+Qed.
+
+(** Compatibility with other operations *)
+
+Lemma rev_length : forall s : string, length (rev s) = length s.
+Proof with simpl; auto.
+  induction s...
+  rewrite append_length.
+  rewrite IHs.
+  elim (length s)...
+Qed.
+
+Fixpoint rev_append (s acc : string) : string :=
+  match s with
+  | EmptyString => acc
+  | String c s' => rev_append s' (String c acc)
+  end.
+
+Definition rev' (s : string) : string :=
+  rev_append s EmptyString.
+
+Lemma rev_append_rev : forall s1 s2 : string,
+    rev_append s1 s2 = rev s1 ++ s2.
+Proof with simpl; auto.
+  induction s1...
+  intros; rewrite <- append_assoc...
+Qed.
+
+Lemma rev_alt : forall s : string, rev s = rev' s.
+Proof.
+  intros; unfold rev'.
+  rewrite rev_append_rev.
+  rewrite append_EmptyString_r.
+  reflexivity.
 Qed.
 
 (** *** Substrings *)


liyishuai:
If I want to add another feature (converting `string <-> Vector.t ascii`) that require changes proposed here, should I push to this branch or open a new PR?

CHANGES
```diff
@@ -65,6 +65,12 @@ Standard Library
 
 - Added `Ndigits.N2Bv_sized`, and proved some lemmas about it.
 
+- Added `String.rev` and proved some lemmas about it.
+  Note that this might cause incompatibilities if you have, e.g., List and
+  String both imported with String on top, and expect `rev` to refer to
+  `List.rev`.
+  Solution: quantify `rev`.
```
liyishuai: Is this the correct solution?

jashug: "quantify" -> "qualify"

liyishuai:
I'm requesting 2 approving reviews to merge QuickChick/QuickChick#104.
The fix for other projects are under way.

xavierleroy:
Waitaminute.  I already find it annoying that String defines a `++` notation that can hide that of List.  But at least the functions are named differently (String.append vs List.app).  Now you want to add a(nother) name conflict between List and String?

ejgallego:
> Waitaminute. I already find it annoying that String defines a ++ notation that can hide that of List. But at least the functions are named differently (String.append vs List.app). Now you want to add a(nother) name conflict between List and String?

Well, there seem to be several issues here; first, is that Coq maybe should provide a type class for monoid-like structures and `++` should be a notation for the generic operator, then let the type inference to do the rest.

Regarding naming, it seems to me that uniform naming reign over conflicts, and use name qualification resolve conflicts.

That being said, I have no opinion on this PR per-se as I use a different solution for Vectors/Lists.

liyishuai:
The updated QuickChick requires [coq-simple-io](https://github.com/Lysxia/coq-simple-io). Shall we add it to Coq CI? @Lysxia

xavierleroy:
Strings in the Coq stdlib are a disaster.  Defining `string = list ascii` would give all the operations and theorems and notations of lists for free.  (No type classes; just parametric polymorphism.)  Making `string` an abstract type would support extraction to native OCaml strings.  The current implementation, with an inductive that is isomorphic to `list ascii` but distinct, is the worst of both worlds: no reuse and no efficient extraction.  Copying bits and pieces from module List to module String is not the way forward.

Zimmi48:
Summoning @maximedenes, as the project lead of stdlib2, to this discussion.

Zimmi48:
@liyishuai coq-simple-io was already added to the v8.8 CI (see #8256), it will be easy to replicate this in master.

Zimmi48:
See #8368

herbelin (approved):
The contents looks good to me.

About the overlap of notations, I don't know what to do. Isn't the overlap solvable by playing with `Close Scope` and `Undelimit Scope`?

As a general rule, though independent of this PR, a direction could be to put the string notations in a submodule `StringNotations`, as it is already done for `ListNotations`, `EqNotations`, `ProperNotations`, `VectNotations`, `IfNotations`, ... @maximedenes promised to take the responsibility of a redesign of the standard library and that could be one rule which we could adopt.

About name overlap, as @ejgallego says, the rule which we tend to adopt is indeed rather to reuse similar names on purpose and to recommend qualification for disambiguation. Ideally, one should be able to import objects of a module selectively (names, coercions, hints, notations, ...) but we are far from reaching this stage, despite the efforts of some developers. (In comparison, Agda is e.g. more flexible than Coq on importing selectively.)

About a different implementation of `String`, why not. I personally have no particular experience in using `String`, so I'm ready to rely on @xavierleroy's experience. A request for pulling a simple list-based variants of `String` might actually find support from a couple of users.

maximedenes:
> As a general rule, though independent of this PR, a direction could be to put the string notations in a submodule StringNotations, as it is already done for ListNotations, EqNotations, ProperNotations, VectNotations, IfNotations, ... @maximedenes promised to take the responsibility of a redesign of the standard library and that could be one rule which we could adopt.

I would rather implement partial `Import`s. Easy to do technically, but we need to get the design right, which is tricky.

liyishuai:
I think partial import refers to that in [Haskell](https://wiki.haskell.org/Import)?

herbelin:
@liyishuai: so, the same as what I was mentioning from [Agda](http://wiki.portal.chalmers.se/agda/pmwiki.php?n=ReferenceManual.Modules#mods) in my third paragraph?

liyishuai:
Yes they look quite similar. I also find them more convenient than the all-or-nothing import we currently have in Coq.

herbelin:
> Yes they look quite similar. I also find them more convenient than the all-or-nothing import we currently have in Coq.
> 
Definitely!

herbelin:
Shall we leave the PR as it is and keep the suggestions made in the discussion for future work (uniformisation of names, export of notations on explicit demand, a finer-grained `Import` command)?

In the absence of further comments, I shall merge on Wednesday evening (Paris time).

Zimmi48:
Isn't it a bit early to announce a merge date given that this PR is compatibility breaking and no overlay has been prepared for compcert, VST and fiat-parsers? Note also that the CHANGES entry should probably be moved to a new post 8.9 section.

herbelin:
> Isn't it a bit early to announce a merge date given that this PR is compatibility breaking and no overlay has been prepared for compcert, VST and fiat-parsers?

Oh, right, I missed that.

For CompCert, the [fix](https://github.com/liyishuai/CompCert/commit/c5a6c76c34be48d141ac70680c322f05be5a026c) is short and simple (adding two qualitifications), so the question is whether @xavierleroy is ok with the principle of recommending a qualifier-based model, or if we instead try to favor strict compatibility (to some price to be evaluated), or, maybe, if we even target a longer-term decision on the design of the `String` library (e.g. by defining `string` from `list`), or even of the `List` library (e.g. by trying to align the naming policy there even closer to the one of OCaml `list.ml`, typically moving `app` to `append`). [Note also that I'm not a specialist of the `String` library, so, I'm acting about it only as a coordinator of the PR and I'm ready to hear any good idea.]

For VST, the first shown failure seems similar to the one of CompCert. For fiat-parsers, this seems less obvious, but the collision between `String.rev` and `List.rev` is the most plausible too.

[Hoping that I did not miss something of the discussion.]

Zimmi48:
>For VST, the first shown failure seems similar to the one of CompCert.

Indeed, VST contains a copy of CompCert so likely the fix is the same.

xavierleroy:
I already said that I'm opposed to this PR, which just adds problems on top of a problematic-by-design library.  Merge if you must, but prepare overlays and all that stuff because CompCert will not be updated preventively.

herbelin:
@xavierleroy: your alternative is then that another `String` library based on `list` is developed instead, right? Which would not be difficult anyway to do, especially if it happens to be an opinion shared by other users of the `String` library.

Regarding qualifying `List.rev`, by grepping `List.` in CompCert, I see a lot of occurrences of `List.map`, `List.length`, etc., so it does not seem to me that it would be excessive to change a few `rev` into `List.rev`. Of course, if your worry is strict compatibility, then, this is another issue, which goes far beyond the scope of this PR and which would deserve a proper discussion between top users.

herbelin:
@xavierleroy: I opened issue #8483 to collect ideas about the design of the `String` library. Please check that what I said there reflects what you meant.

I'm proposing to support this PR in that providing more functions for manipulating `String` shall be eventually necessary anyway and that it seems to me more rational to reuse the same naming and notations as `List` than inventing new names. However, building a new `String` library based on @xavierleroy's suggestions seem doable relatively quickly, so if a clear support can be found in this direction, it could make this PR quickly obsolete.

Zimmi48:
I think going against a clear opposition like this from @xavierleroy should not be done without a discussion at the WG.

herbelin:
@xavierleroy: I'm not so familiar with extraction nor efficiency in general. What actually prevents to have an efficient extraction with the current (non abstract) implementation of `string`. Extracting `string` to `string`, `String.eqb` to `(=)`, `String.append` to `^`, etc. would not work, or would not be enough?

I mean: there is a file `ExtrOcamlString` in the Coq archive which extracts `string`  to `char list` for [simplicity](https://github.com/coq/coq/commit/e5034fbbf23edb17dd88175b2f44b2dc9b4110fa#diff-3359939b615d37e8440e72cf3209e146) but which could have extracted directly to `string` as well.

xavierleroy:
> What actually prevents to have an efficient extraction [...]

As far as I can see, you need specific OCaml code in an extraction plugin to generate string literals instead of list literals.  The remainder of the mapping could probably be done with `Extract Constant` directives.  There might be other "gotchas".

herbelin:
> As far as I can see, you need specific OCaml code in an extraction plugin to generate string literals instead of list literals. 

I see, so assuming a binding of Coq `string` to OCaml `string`, say,
```coq
Extract Inductive string => string [
 """""" (* binding for EmptyString *)
 "(fun (a,s) -> String.make 1 a ^ s)" (* binding for String *)
]
"(fun f0 fS s => let l = String.length s in
  if l = 0 then f0 else fS s.[0] (String.sub s 1 (l-1)))" (* binding for "match" on string *).
```
one would still have an artificial overhead in building string literals character by character (since, e.g. `"Hi"`would be extracted to `((fun (a,s) -> String.make 1 a ^ s) ('H', ((fun (a,s) -> String.make 1 a ^ s) ('i',"")))`). So, indeed, one would somehow need in extraction special code.

(Actually, the problem is similar with integer literals, right?)

Back to the Coq `String` library, I'm still concerned by how to make progresses. For extraction, we need a type `string` which is distinct from `list` but the question of efficiency of string extraction would not be different with a different library, i.e. one would have in all cases to use `Extract Constant` directives to take benefit of the actual OCaml primitives, as well as hoping for a variant of primitive numeral/string printers targetted for extraction, right?

In particular, @xavierleroy, if one would e.g. start an alternative library based, say, on:
```coq
Inductive string := String : list ascii-> string.

Definition lift {A} (f : list ascii -> A) '(String s) := f s.
```
would that meet your expectations? I.e. distinct type so that extraction can be customized + easy reuse of the existing combinators on `list` by simple insertion of the `lift` from above.

herbelin:
I'm still wondering what to get out of the discussion on this PR. @xleroy, would you be ready to support the "burden" of a move to another `String` library which would provide almost for free all `lIst` combinators, as described in my last comment

(Note that my proposal above, in addition to preserve native extraction to `string` and inherit `list` operators, would preserve easy definition of primitive parsers/printers as legitimately required in #8483. With a bit of support for the `string` to `list ascii` isomorphism, we could even inherit the properties on `list`.)

SkySkimmer:
This PR is dead. It may be reopened if it ever gets revived.

herbelin:
Basically, the PR was good to me (unless we move in a short term to a new String library, in which case, we can target the new library).

Maybe someone who followed all discussions on a new string library (the coq-club 6 Nov 2018 discussion or stdlib2) can comment?

ppedrot:
I side with @SkySkimmer: this PR is dead, and the experiment just showed that reopening it has not turned it magically alive again. Let's close it with a wooden stake.

PR #8357, opened 2018-08-30
fix #8347 unification of evars reduces in types unecessarily

### LINKED ISSUE #8347: unification of evars reduces in types unecessarily?

Sometimes the unification of evar performs some reduction over there type, which is a bit annoying in some situation. Here the situation where it is annoying: I want to write a tactic that generates a subgoal of which type is determined by unification of an evar.

Definition eq_one (i:nat) := i = 1.
(* eq_one is delta_reducible but I don't want it to be reduced. *)
Lemma foo: (eq_one 1 -> False) -> False.
Proof.
  intros H. 
  (* I want to prove the premiss of H (eq_one 1) BUT without writing its type. *)
  (* So I create an evar which will be instantiated to the type of H by the next tactic: *)
  let ev := open_constr:((_:Prop)) in
    assert (id:ev).                   (* create a subgoal of type ?ev. *)
  2:specialize H with (1:=id). (* instantiate ?ev to the premiss of H. *)

Here I expect that my sugoal type has gone from ?ev to (eq_one ?x) but I get (?x = 1). eq_one has been reduced.

Matafou:
Fixing a side effect of evar unification (hnf applied to the matching type).

**Kind:** bug fix.

Fixes / closes #8347

### DIFF
--- a/dev/ci/user-overlays/08357-Matafou-fix-specialize-#8347.sh
+++ b/dev/ci/user-overlays/08357-Matafou-fix-specialize-#8347.sh
@@ -1,0 +1,9 @@
+if [ "$CI_PULL_REQUEST" = "8357" ] || [ "$CI_BRANCH" = "fix-specialize-#8347" ]; then
+
+    Corn_CI_REF=fix-specialize-#8347
+    Corn_CI_GITURL=https://github.com/Matafou/corn
+
+    unimath_CI_REF=fix-specialize-#8347
+    unimath_CI_GITURL=https://github.com/Matafou/UniMath
+
+fi
--- a/proofs/clenv.ml
+++ b/proofs/clenv.ml
@@ -23,7 +23,6 @@ open EConstr
 open Vars
 open Reduction
 open Reductionops
-open Tacred
 open Pretype_errors
 open Evarutil
 open Unification
@@ -56,8 +55,6 @@ let refresh_undefined_univs clenv =
     let map_freelisted f = { f with rebus = subst_univs_level_constr subst f.rebus } in
       { clenv with evd = evd'; templval = map_freelisted clenv.templval;
 	templtyp = map_freelisted clenv.templtyp }, subst
-
-let clenv_hnf_constr ce t = hnf_constr (cl_env ce) (cl_sigma ce) t
 
 let clenv_get_type_of ce c = Retyping.get_type_of (cl_env ce) (cl_sigma ce) c
 
@@ -496,7 +493,7 @@ let clenv_unify_binding_type clenv c t u =
 	  TypeNotProcessed, clenv, c
 
 let clenv_assign_binding clenv k c =
-  let k_typ = clenv_hnf_constr clenv (clenv_meta_type clenv k) in
+  let k_typ = clenv_meta_type clenv k in
   let c_typ = nf_betaiota clenv.env clenv.evd (clenv_get_type_of clenv c) in
   let status,clenv',c = clenv_unify_binding_type clenv c c_typ k_typ in
   { clenv' with evd = meta_assign k (c,(Conv,status)) clenv'.evd }


Matafou:
This PR is a proposition, and also to see if something breaks.
This change should be discussed by expert of evar unification.

Zimmi48:
@Matafou the build raises warnings (which are fatal) so we can't see if anything breaks afterwards.

Matafou:
hmmm lint-repository.sh rejects this commit but it consistsonlyin removing a few words from the code. I think this is a bug...

SkySkimmer:
It doesn't though? https://travis-ci.org/coq/coq/jobs/422492013
The OSX job is the cause of the travis failure.

ppedrot:
There are two failures, are they related?

Matafou:
Yes at least one is related. An strange `apply with` relying on the unfolding of a constant. Strangely this behavior of apply with cannot be done with `apply` without  `with` (even without my change). I am still invastigating the other failures but I guess they are similar.

Matafou:
More precisely, in `UniMath/UniMath/RealNumbers/NonnegativeReals.v` the goal is of the form
```
   x : Dcuts   r : NonnegativeRationals   
   Hr : r ∈ Dcuts_half x
  ============================
  (r <= r + r)%NRat
```
where (with master branch) the following tactic works:
``` apply is_Dcuts_bot with (1:= Hr).``` (1)
but not this one:
```apply (is_Dcuts_bot _ _ Hr).``` (2)
which should be equivalent imho. Ont he other hand this works (3):
```
red in Hr.
red in Hr.
apply (is_Dcuts_bot _ _ Hr).
```
With my current suggestion only (3) works.

My remarks:
- My modification looks like a regression
- on the other hand it makes (1) and (2) more consistent with each other.
- But I can see why we may want (1) and (2) to (consistently) succeed instead. In this case I wonder how to make my initial ltac behave as expected.

herbelin:
What is the error message in (1) and (2)? Would it work with `refine` instead? Don't know what other thinks, but if it is a failure of unification, and if `refine` succeeds, I would adapt `UniMath` to your patch.

Incidentally, if we remove the `nf_betaiota` as well, what happens? (The `nf_betaiota` is also from before 7.0.)

ppedrot (commented):
I don't have an opinion on the change. It does make the semantics slightly easier but it is (marginally) not backwards compatible.

ppedrot:
Is there any future for this PR? Is the new behaviour the one we want or not? (@herbelin is probably more informed on that question.)

herbelin:
I would say that the new behavior is better (closer to the user expectation) and, if there is only one ci failure, that's not too bad. We could try to move the `hnf_constr` in `w_merge` so that the UniMath compatibility is preserved but `Meta`s are not supposed to be the future, so is it worth to work on `w_merge` while there is an ongoing move to using only `Evar`?

So, as far as I'm concerned I would support the PR.

Matafou:
There is a second failure in  ci-formal-topology. I did not have time to investigate it yet but it seems to be the same kind.

ppedrot:
@Matafou do you plan working on this PR? Otherwise, @herbelin are you willing to take it up?

Matafou:
I think I understand roughly your explanation let me sum up my understanding of things:

**Currently** (without this PR):
1) in some situation like this apply and (apply with) do not behave consistently.
2) applywith behaves better than apply in these situations but for a bad reason: this early unfolding works around a weakness of the unification algorithm.
3) (apply with) has a bad side effect on the type of existential variables (#8347): The unfolding that happens during unification should not change the type of the the evar itself, but only the the unification problem (if it is possible, maybe using a second evar?).

**With this PR**:
1) (apply with) become more like apply, which is a regression of applywith even if it makes things more consistent.
2) The bad side effect is fixed, which is a progress.


Ma analysis is that there are two different problems to solve
1) The backtracking mechanism of unification that does not backtrack unfolding
2) the bad side effect

** 4 solutions in (my) order of preference**
a) solve 1), it would allow to remove the hnf in applywith with no regression and thus make 2) disappear.
b) not solve 1) (not worth the effort?) and fix the bad side effect without making it regress
c) this PR: not solve 1) (not worth the effort?) but remove the hnf in applywith and make it regress.
d) do nothing.

@herbelin do you think the fix you suggest in unification is possible (with possible impact in performances)?
Would it be possible to still make this hnf but without impacting the final type of the evar?

Matafou:
Sorry for the close/open... Do we need to re-add the 8.10beta milestone flag?

ejgallego:
> Sorry for the close/open... Do we need to re-add the 8.10beta milestone flag?

If you would like to see this merged on 8.10 sure.

Matafou:
I think this PR (once rebased) is OK to be merged because it is a small regression vs a cleaner behaviour. Unless @hugo thinks we can solve the problem without regression in 8.10.

ppedrot:
@Matafou in this case, you should probably write overlays for the broken contribs.

Matafou:
Salut Pierre-Marie
Je veux bien mais je n’ai aucune idée de ce que c’est . Et même Google ne m’aide pas beaucoup. Tu as un pointeur sur une doc?
P

Pierre

________________________________
De : Pierre-Marie Pédrot <notifications@github.com>
Envoyé : mercredi, février 6, 2019 12:33 PM
À : coq/coq
Cc : Pierre Courtieu; Mention
Objet : Re: [coq/coq] fix #8347 unification of evars reduces in types unecessarily (#8357)


@Matafou<https://github.com/Matafou> in this case, you should probably write overlays for the broken contribs.

ejgallego:
> Salut Pierre-Marie Je veux bien mais je n’ai aucune idée de ce que c’est . Et même Google ne m’aide pas beaucoup. Tu as un pointeur sur une doc? P

https://github.com/coq/coq/blob/master/dev/ci/README-developers.md#breaking-changes

Matafou:
Thanks I will do that next week.

Pierre

________________________________
De : Emilio Jesús Gallego Arias <notifications@github.com>
Envoyé : jeudi, février 7, 2019 1:44 AM
À : coq/coq
Cc : Pierre Courtieu; Mention
Objet : Re: [coq/coq] fix #8347 unification of evars reduces in types unecessarily (#8357)


Salut Pierre-Marie Je veux bien mais je n’ai aucune idée de ce que c’est . Et même Google ne m’aide pas beaucoup. Tu as un pointeur sur une doc? P

https://github.com/coq/coq/blob/master/dev/ci/README-developers.md#breaking-changes

Matafou:
Overlays added.

On Corn it is a bit more worrying: with this PR some (one) `stepl` tactic makes a different choice for the equality relation. I wonder how this PR can have this side effect... I am not expert enough in type classes to understand what happens.
It is however solved easily by replacing stepl by transitivity.

maximedenes:
@Matafou I'm always worried with this kind of changes, as they break script in subtle ways, and can't really be even documented because the whole thing is already more or less undocumented.

So whether we want to go this way probably depends on how strong the motivation for the change is. It is not obvious from the description, can you say more about it?

maximedenes:
Nevermind, I just found it in the linked issue, sorry.

Matafou:
Well it is more a case were I am bitten by his unexpected unfolding than a real motivation.

The real motivation imho is that (1) eapply (foo _ _ _ H) and (2) eapply foo with (2:=H) do not behave the same. The latter performs an initial the hnf on clenv while the former does not.

I was bitten by this because (2) reduced the type of the evar too much for my needs, but this can be considered harmless.

BUT UniMath showed one case where (2) is strictly better than (1), which I consider a bug.

My patch makes it more consistent by making it behave (2) behave like (1). Fortunately I think incompatibilities will be rare because unification performs unfolding anyway. If I understood correctly Hugo it is only a weakness of the unification algorithm that make sometimes the initial unfolding better.

Matafou:
Overlays seem OK at last.

I leave it to the specialists if this is to be merged or not.

ppedrot:
This PR doesn't seem to be converging. If somebody wants to push it more strongly, please resuscitate it. In the meantime I am closing it.

herbelin:
I would still lean to think that the change is saner but I have no real time to take over the PR.

At least, I added a comment on the [UnificationProblems](https://github.com/coq/coq/wiki/UnificationProblems) wiki page to remember the `evarconv` issue.

herbelin:
PS: is there a way to keep links to the CI results for closed/merged PRs?

SkySkimmer:
>PS: is there a way to keep links to the CI results for closed/merged PRs?

Go to the commit list https://github.com/coq/coq/pull/8357/commits then click the red cross / green check / orange circle (depending on if CI failed / succeeded / didn't finish).

herbelin:
Ah thanks. Then those inconsistencies found by CI are very few.

PR #8360, opened 2018-08-30
Backport fix for universe polymorphism capturing global universes

SkySkimmer:
This is a backport of the part of #8341 which fixes the inconsistency.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,12 @@ Tools
   and `print-pretty-single-time-diff` now correctly label the "before" and
   "after" columns, rather than swapping them.
 
+Kernel
+
+- The kernel does not tolerate capture of global universes by
+  polymorphic universe binders, fixing a soundness break (triggered
+  only through custom plugins)
+
 Changes from 8.8.0 to 8.8.1
 ===========================
 
--- a/kernel/environ.ml
+++ b/kernel/environ.ml
@@ -194,8 +194,7 @@ let push_constraints_to_env (_,univs) env =
 
 let add_universes strict ctx g =
   let g = Array.fold_left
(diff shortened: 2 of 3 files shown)


Zimmi48:
The log file says `coq-makefile` was not found:

```
==========> TESTING misc/poly-capture-global-univs.sh <==========
misc/poly-capture-global-univs.sh: line 10: coq_makefile: command not found
0m0.001s 0m0.000s
0m0.001s 0m0.001s
    misc/poly-capture-global-univs.sh...Error!
```

The difference between master and v8.8 is likely due to some other change that was not backported. Any candidate change in Makefile targets?

Zimmi48:
Note that it didn't just fail on Nix, it also failed on AppVeyor (Windows) and Travis (macOS).

Zimmi48:
And also in the rest of test-suite (Linux) target on your Travis: https://travis-ci.org/SkySkimmer/coq/builds/422589619
So what does GitLab CI do specially to make it work there?

Zimmi48:
Given that this is only a critical bug if one goes through plugins and there are more open bugs like this and that we cannot understand why the test-suite is failing, I guess we will just skip this in 8.8.2.

SkySkimmer:
The test is a `misc/` script so inherently tricky, maybe we can just merge the fix and not the test in 8.8.2?

Zimmi48:
If you are confident that the fix is the same, then indeed, this solution is good to me.

SkySkimmer:
Test dropped.

Zimmi48:
@SkySkimmer Contrarily to what @coqbot seems to believe, this PR is not up-to-date with v8.8 which explains the failure of HoTT (still testing the master branch instead of the V8.8 tag). Can you rebase once more?

Issue #8223, opened 2018-08-07
Grammar for SSReflect style Inductive types


anton-trunov:
#### Version

Coq 8.8.1

#### Operating system

macOS 10.13.6

#### Description of the problem

I failed to find the grammar for SSReflect style `Inductive` definitions in the reference manual. An example of such a definition can be found [here][1]:

```coq
Inductive list (A : Type) : Type := nil | cons of A & list A.
```

[1]: https://coq.inria.fr/refman/proof-engine/ssreflect-proof-language.html#anonymous-arguments

gares:
@amahboubi do you recall if the documentation of that syntax was in the latex user manual and was lost during the migration to sphinx?

gares:
JFTR, such notation is available everywhere, you can even write
```coq
Lemma foo of False & True.
```

anton-trunov:
> FTR, such notation is available everywhere

Oh, I didn't know that, thank you! But you probably meant something like this?
```coq
Lemma foo of False & True : False.
```

Otherwise I get:
> Error: Syntax error: ':' expected after [constr:binders] (in [vernac:gallina]).

herbelin:
BTW, what is the general feeling about this notation? It is a bit a trick to provide `of t` as a strict abbreviation for `(_:t)` but the OCaml-reminiscent syntax `Inductive list (A : Type) : Type := nil | cons of A & list A.` is somehow cool.

Currently, it is a hack, since e.g. you can also write `Inductive list (A : Type) : Type := nil | cons & A of list A.` which is difficult to justify, but I guess we could agree on a clean improvement of it.

@mattam82, @ppedrot, @anton-trunov, @gares, ... what do you think?

anton-trunov:
@herbelin Having a uniform syntax would be great, indeed!

Zimmi48:
Honestly, I can't see the gain of this syntax over
```
Inductive list (A : Type) : Type := nil | cons (_ : A) (_ : list A).
```

Trying to reproduce OCaml syntax seems generally like a bad idea to me because the choices made for the syntax of OCaml are often pretty bad (and in particular in the case of the definitions of constructors of ADTs).

However, I also think that it's bad that the SSR plugin is more than just about tactics, so if people like this syntax and care about it, I wouldn't mind having it moved (with a cleaner implementation) outside of the plugin and made generally available.

amahboubi:
[...]

PR #8372, opened 2018-08-31
Grammar rule in the documentation of ssreflect's anonymous arguments

amahboubi:
**Kind:** documentation

Addresses issue #8223.

### DIFF
--- a/doc/sphinx/proof-engine/ssreflect-proof-language.rst
+++ b/doc/sphinx/proof-engine/ssreflect-proof-language.rst
@@ -444,11 +444,16 @@ not its name, one usually uses “arrow” abstractions for prenex
 arguments, or the ``(_ : term)`` syntax for inner arguments. In |SSR|,
 the latter can be replaced by the open syntax ``of term`` or
 (equivalently) ``& term``, which are both syntactically equivalent to a
-``(_ : term)`` expression.
+``(_ : term)`` expression. This feature almost behaves as the
+following extension of the binder syntax:
 
-For instance, the usual two-constructor polymorphic type list, i.e.
-the one of the standard List library, can be defined by the following
-declaration:
+.. prodn::
+   binder += & @term | of @term
+
+Caveat: ``& T`` and ``of T`` abbreviations have to appear at the end
+of a binder list. For instance, the usual two-constructor polymorphic
+type list, i.e. the one of the standard ``List`` library, can be
+defined by the following declaration:
 
 .. example::
 
@@ -5387,7 +5392,7 @@ Tacticals
 
 discharge :ref:`discharge_ssr`
 
-.. prodn:: tactic += @tacitc => {+ @i_item }
+.. prodn:: tactic += @tactic => {+ @i_item }
 
 introduction see :ref:`introduction_ssr`
 


Zimmi48:
@amahboubi It seems you have messed up with git. We should only see 86cddb0e709c885589b56a19aa5f5b958c33ea97. One solution would be
```
git checkout master
git pull
git checkout ssr-anon-args-doc
git reset --hard master
git cherry-pick 86cddb0e709c885589b56a19aa5f5b958c33ea97
git push -f
```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -444,10 +444,13 @@ not its name, one usually uses “arrow” abstractions for prenex
 arguments, or the ``(_ : term)`` syntax for inner arguments. In |SSR|,
 the latter can be replaced by the open syntax ``of term`` or
 (equivalently) ``& term``, which are both syntactically equivalent to a
-``(_ : term)`` expression.
+``(_ : term)`` expression:
+
+.. prodn::
+   binder += binder | & term | of term
```
Zimmi48: For the grammar tokens at the left of the `+=` in a `prodn` directive, you need to put an `@` in front. This would give `binder += @binder | & @term | of @term`.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -444,10 +444,13 @@ not its name, one usually uses “arrow” abstractions for prenex
...
 
 For instance, the usual two-constructor polymorphic type list, i.e.
-the one of the standard List library, can be defined by the following
+the one of the standard `List` library, can be defined by the following
```
Zimmi48: To have a verbatim rendering you need two backticks before and after: ``` ``List`` ```

amahboubi:
Thanks for the instantaneous answer @Zimmi48. Yes I also realized that too late and things should be in better shape now.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
amahboubi: Thanks. Side remark: references to libraries are very heterogeneous (in particular, with or with .v). Is there a general policy?

Zimmi48:
>Thanks. Side remark: references to libraries are very heterogeneous (in particular, with or with .v). Is there a general policy?

There is none AFAIK. Feel free to open a PR homogenizing this stuff and documenting the convention :wink:

Zimmi48:
Rendered: https://gitlab.com/coq/coq/-/jobs/93741707/artifacts/file/_install_ci/share/doc/coq/sphinx/html/proof-engine/ssreflect-proof-language.html

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -444,10 +444,13 @@ not its name, one usually uses “arrow” abstractions for prenex
 arguments, or the ``(_ : term)`` syntax for inner arguments. In |SSR|,
 the latter can be replaced by the open syntax ``of term`` or
 (equivalently) ``& term``, which are both syntactically equivalent to a
-``(_ : term)`` expression.
+``(_ : term)`` expression:
+
+.. prodn::
+   binder += @binder | & @term | of @term
```
Zimmi48: I only noticed this when reading the rendered doc but I think the `+=` operator means you're not supposed to repeat `binder` on the right. It could just be:
```rst
.. prodn::
   binder += & @term | of @term
```

Zimmi48 (changes requested):
I think this addition to the doc makes it a lot clearer :tada:

Zimmi48 (approved):
Perfect, thanks! Would you mind squashing your commits together?

Zimmi48:
>I'm unsure that this is going in the right direction to write explicitly in the manual that...

>I find that that would not go in the right direction to document...

Sorry but I don't agree with this kind of thinking. What exists *has to* be documented. We need to stop thinking that because something is undocumented it won't be used. This is independent on the question whether we want to deprecate it and remove it (which is totally a discussion that we can have).

gares (approved):

Zimmi48:
>maybe would you like first to hear from @ggonthier what was his intent?

That'd be an interesting question to ask but it doesn't change what I think of documenting the feature while it exists. Note that whatever his initial intent was, the Coq developers can now decide to do whatever we want with this feature: restrict it, deprecate it and remove it entirely, etc.

>now that what we are discussing at documenting it, it may not be too late to really think at what we want

It's never too late to change, and in #8223 we talked about moving this to the main code out of the SSReflect plugin. If this was done, of course this would need to be implemented differently then. But again, this is unrelated to documentation. I don't think there are any excuse to not document something. What we can do though, is to explicitly add to the doc that this is considered a hack and merely intended to be used as part of an inductive definition.

amahboubi:
HI @herbelin, the use case mentioned by Georges is indeed specially useful when programming with canonical structures.
Here is a typical example: library [poly](https://github.com/math-comp/math-comp/blob/1dcea6744ecc79546905d41a23801b1fc7b60c5a/mathcomp/algebra/poly.v#L127) defines a datastructure `polynomial` for polynomials, as lists of coefficients, with a non-zero last element. This datastructure is parametrized with an instance of ring, which provides the "zero" element used in the test. Now one still wants to have access to a notation like `{poly int}` for the type of polynomials with integer coefficients, as opposed to passing the explicit instance of ring structure available on type `int`. But the term `polynomial int`is ill-typed, and the failure does not trigger the unifcation-based canonical instance look-up. The definition of the notation `{poly int}` uses a phantom type to pose a unifcation problem triggering the the look-up, which allows the user to benefit from the declared canonical instance of ring.
These "phantom types" are thus part of the standard canonical structures tool box, and by nature, this toolbox involves definitions with anonymous arguments. 
Our [paper](https://hal.inria.fr/hal-00816703v2) with @gares  provides another (more detailed) example of this idiom, p12, for a technique to improve error messages and ease debugging. The notation `[find v | t1 ∼t2 ] rest` is similar to the previous example for polynomials. It uses the standard `(_ : T)` notation because our purpose was to stay plugin-independent.
The Mathcomp library features a few other use-cases of this syntax, for instance in the statement of proof-dependent definitions. [Here](https://github.com/math-comp/math-comp/blob/1dcea6744ecc79546905d41a23801b1fc7b60c5a/mathcomp/algebra/ssralg.v#L1286) is the example of the Frobenius morphism, `(fun x => x ^ p)` which requires the exponent `p` to be the characteristic of the underlying ring/field.

Zimmi48:
@herbelin Do you mind if I merge this documentation PR and the discussion on this feature and whether it should get restricted can continue nonetheless here on somewhere else?

herbelin:
Thanks for clarifying that the motivation was eventually explicitly also for discharged binders (and apparently for the 3 occurrences of `fun ... of ...` that I found by grepping MathComp), but my question is also about whether it would make sense to think at the general level of Coq, rather than only at the level of the idioms of Ssreflect, and to propose a general syntax for `(_:T)`, at least when it is about what @ggonthier calls discharged binders, and, if yes, to understand what is the general scheme for this syntax. For instance, I realize that the syntax documented by @amahboubi is actually misleading if not wrong, since, due to precedence levels, it can only come at the end of a list of binders , and thus, is not strictly speaking an abbreviation for `(_:T)`. Also, I'm not sure that things like `Definition f & nat of nat :=` is intended. So, wouldn't it be worth to think again at the feature, and to make it more robust, especially in a context where we have access to `g_constr.ml` and don't need a hack, also deciding whether it is really an abbreviation for `(_:T)` or something more restricted.

Zimmi48:
I take this as a "yes, I do mind".

amahboubi:
By the way, the purpose of my previous message was precisely to illustrate the general interest of such a notation.
My link to the Frobenius morphism also illustrates that a simple search based on the `fun ... of ...` pattern would not catch all the instances. For instance in library [morphism](https://github.com/math-comp/math-comp/blob/master/mathcomp/fingroup/morphism.v), you will find 3  instances in a row of similar proof-dependent definitions: `dom` the [domain](https://github.com/math-comp/math-comp/blob/1dcea6744ecc79546905d41a23801b1fc7b60c5a/mathcomp/fingroup/morphism.v#L137) for the domain of a group morphism, and `morphim` and `morphpre` (which come right after `dom`) for the image and preimage of such a morphism. All of them are of the shape `Definition ... of...`.

ggonthier:
That the `of T` / `& T` abbreviations have to appear at the end of a binder list is a conscious design tradeoff. It avoids spurious brackets around `T` in the common use cases, and isn't too much of an issue when one adopts a prenex style of putting all named before unnamed arguments (a practice also useful in dealing with a similar restriction of the `-> T` notation having to appear after `forall` or discharged binders).
   I decided not to impose any order on the use of `of T` vs. `& T` because this is purely a style issue, and I didn't want system code to be prescriptive on such a matter (I'm no big fan of the `Variable` / `Variables` warning either). Indeed I've leaned towards using `fun ... & phant_id u v => ...` without a leading `of` for instance factories - there are over 30 instances in `ssralg.v` alone (as
 @amahboubi pointed out,  a search for `fun ... of` misses most of the use cases).

ejgallego:
Indeed this seems like a useful and reasonably tested addition to binder syntax so IMHO it would make sense to move it to the core grammar; I say that as a user of such form of course so I am very partial to this party.

The main issue would be whether enabling this by default would break any 3rd party development.

ejgallego:
> I'm no big fan of the Variable / Variables warning either

This warning does also exist in other variations such as `Implicit Type(s)`, please open an issue if you would like the warning to be removed or the singular form to be deprecated.

Zimmi48:
The warning is also off by default IIRC.

>The main issue would be whether enabling this by default would break any 3rd party development.

But CI is great for testing this.

herbelin:
> fun ... & phant_id u v => ... without a leading of for instance factories - there are over 30 instances in ssralg.v alone

Ah, I thought that the idiom was to use `&` only after a `of`, so I did not try to grep it. Thanks for telling.

I'm not sure I'm convinced though. I was expecting a more universal approach and, @ggonthier, the more I read your argumentation, the more I feel it is just an idiom rather than a standard solution.

I still like the idea to be able to write `Inductive list A := nil | cons of A & list A` and I'd be in favor of supporting it. I'm however less convinced by the other uses. I'm unsure that `foo of some type` or `foo & some type` are clearer than simply providing a dedicated general syntax for `(_ : T)` as most of us initially thought this PR was about. The symbol `&` suggests a conjunction, so, making `& T` an abbreviation for `_:T` would either need to rethink at it in another role than the role of conjunction, or to forbid it in first position, but that would then be quite artificial to allow a notation for `_:T` only from the second position. As far as `of` is concerned, let's try to make the mental effort that a notation `Definition foo n of typ :=` denotes an object taking an `n` and some (object of type) `typ`. OK, why not. But this would be restricted to `_:T` in last positions because (even by changing the precedences), this would start to be strange to write `Definition foo n of t p of t q := ...` (or at least, we would start to diverge quite a lot from the original motivation of specifically using `of` and `&` over other symbols because of the analogy with OCaml syntax of inductive types). So, if the purpose is before all to abbreviate `_:T`, and not to echo to the syntax of OCaml inductive types, I don't see why `of` and `&` would be better than any other symbols. In particular, a neutral symbol, which does not prejudge of the existence or not of other binders before or after, would look better to me. Actually, I just checked and, unless I mixed up my tests, I didn't see the `` ` `` turning the corresponding argument into an implicit argument, even when a class, so it basically does the same as `of`, up to the question of parentheses (do we agree on this observation?).

About parentheses, saying that parentheses would be spurious if writing `of (P x)` rather than `of P x` seems a bit excessive to me. Parentheses are anywhere needed somewhere to represent tree-like data. So, the spurious is obviously here a spurious according to your own aesthetics, which, as you said, prefers to sacrifice a binder coming after a `of`/`&` rather than seeing parentheses (or some other form of delimiters). Note that I'm perfectly ok with the fact that we all have our own criteria for aesthetism. If I'm making this remark it is only to clarify that the word spurious is somehow an appropriate word in the discussion in that it conveys a moral meaning while it is only about personal tastes.

So, if we summarize, there is already a notation for `_:T` in vanilla coq, and, indeed ``Inductive nat := O | S `(nat).`` works perfectly (even if I did not realize it until today!!). This notation is versatile, it is a bit an arbitrary symbol but it has no connotation such as `&` which comes with the connotation of connecting two things.

Incidentally, there is no reason to expect parentheses in `` `(nat)``, and this constraint could be dropped. Of course, keeping parentheses around an application remains needed if one accepts to prefer versatility.

Now, more generally, what to do with the cool ideas of ssreflect, what to do with the redundancies between cool ideas of ssreflect and features from coq? By initiating this discussion, I had hopes that a discussion with ssreflect developers could converge towards the general interest of the diversity of all users. Maybe I'm wrong, but I feel that your position, @ggonthier, is more to preserve ssreflect as it is, as a plugin with its idioms and redundancies than at working altogether at providing a language which rubs out the existing idiosynchracies and targets all users, all styles, all habits. Am I correct?
But are you sure that this is the best we can do? Don't underestimate the impact of your attitude on other ssreflect developers. If yourself shows turned towards the future, towards wishing a synthesis of the best of what ssreflect provides and of what coq provides, rather than preserving idiosynchracies a priori, I would not be surprised to see other ssreflect developers welcoming the change, with maybe other potential positive impacts in sight...

amahboubi:
> simply providing a dedicated general syntax for (_ : T) as most of us initially thought this PR was about.

This PR is only about documenting existing existing content.

maximedenes (approved):
This PR is obviously right. @Zimmi48, can you merge?

maximedenes:
I suggest to move the discussion somewhere else, it has nothing to do with this PR which is a documentation fix.

ggonthier:
I'm afraid I have to disagree with many of the points made by @herbelin, but also agree with @maximedenes that this discussion of going off topic, so I'll wait for it to be moved before posting a reply.

amahboubi:
@Zimmi48, @maximedenes : I am done.

Zimmi48 (approved):
Approving the new version of this PR which fixes the problem noticed by @herbelin.

This PR being a documentation PR only and the feature being present in 8.8 (and before), I continue to claim (and I insist) that it is necessary to merge this, independent of the discussions on the evolution of this feature (which should indeed happen in their own issue / and separately from a more general discussion about SSR).

I will thus merge this tomorrow, unless I get a formal opposition before then (and I won't accept any ambiguous one). I hope I won't get any formal opposition, but if I get one, this would have to be delayed until the next WG.

herbelin:
I'm ok with the PR as it is.

I otherwise opened #8413 which I called "General discussions on the design of ssreflect" and with a discussion header which you will find [here](https://github.com/coq/coq/issues/8413#issuecomment-418765415). Is it ok for those who have expressed an interest in pursuing the discussion? Or would you prefer a more focused topic specifically on `&`/`of`? Or on the more general topic of abbreviating `_:T`? Or on any other level of discussion you would be interested in?

Zimmi48:
It seems that GitLab CI had had some troubles recently and it has skipped the doc build, so closing / reopening to get a documentation build.

Zimmi48:
I'll rebase myself.

amahboubi:
Hi @Zimmi48.  Sorry I did not realize this was a request. Is it helpful if I do it now? I can.

Zimmi48:
No, all is good now. Thanks!

amahboubi:
Ok, great. Thanks!

Zimmi48:
Rendered: https://coq.gitlab.io/-/coq/-/jobs/96556744/artifacts/_install_ci/share/doc/coq/sphinx/html/proof-engine/ssreflect-proof-language.html#anonymous-arguments

PR #7062, opened 2018-03-24
Slightly refining some error messages about unresolvable evars.

herbelin:
**Kind:** enhancement

This is a slight enhancement in explaining the location of unresolved evars obtained by product or lambda refinement. For instance:
```coq
Goal forall a f, f a = 0.
(* before: Cannot infer an internal placeholder of type "Type" (unlocated) *)
(* after: Cannot infer the domain of the type of f (located at f) *)
```
```coq
Goal forall f P, P (f 0).
(* before: Cannot infer the type of f in environment: x : nat (already located at f) *)
(* after: Cannot infer the codomain of the type of f in environment: x : nat (located at f) *)
```

Note: I don't think this is worth being reported in documentation.

PR #8373, opened 2018-08-31
Fixes #8369: Not_found in printing message about an unresolved subevar.

### LINKED ISSUE #8369: Anomaly when using pattern-matching under lambda
#### Version

master (bf1446294dba45d3ea9b7bb39d2fc96617848c03)

#### Operating system

macOS 10.13.6

```coq
Definition foo := fun '(u, v) p2 => (u, v).
```

master (bf1446294dba45d3ea9b7bb39d2fc96617848c03): 
> Error: Anomaly "Uncaught exception Not_found." Please report at http://coq.inria.fr/bugs/.

Coq 8.8.1:
> Error: Cannot infer an internal placeholder of type "Type" in environment:
> pat, p0 : ?T0 * ?T1

herbelin:
We were trying to print the name of a resolved evar. As a fix, we use the name of the evar if it has a known user name but print an anonymous "subterm" expression if it has no known user name.

Maybe should we not even try to use the original name as this name may have been "donated" to another (unresolved) evar in situations of instantiations which do not "essentially" change the view we have at the evar (i.e. `intro`, `assert`)?

**Kind:** bug fix

Fixes #8369


mattam82:
> Maybe should we not even try to use the original name as this name may have been "donated" to another (unresolved) evar in situations of instantiations which do not "essentially" change the view we have at the evar (i.e. intro, assert)?

I'm not sure I understand. Isn't evar_ident always returning None for an evar that's been defined?

I tried restarting the ci-hott target but the failure looks totally unrelated, maybe a rebase is needed? The last one looks recent though.

herbelin:
@mattam82: You're probably right. I'll need more time to think at it.

herbelin:
I have "existential" doubts. I introduced in #7062 a way to refer to subterms of an evar so that we can produce the following kind of error message:
```coq
Goal forall a f, f a = 0
>               ^
Error: Cannot infer the domain ?T of the type of f.
```
Elaborating on #8369, I wonder if, finally, it could not be as simple and more informative to give:
```coq
Goal forall a f, f a = 0
>               ^
Error: Cannot infer ?T in the partial instance "?T -> nat"
found for the type of f.
```
Similarly, what would be the best here (for #8369):
```coq
Definition foo := fun '(u, v) p2 => (u, v).
>                   ^^^^^^^^^^^^^^^^^^^^^^^^
Error: Cannot infer ?T in the partial instance
"?T@{pat:=pat; p0:=pat} -> ?T0 * ?T1" found for the type
of this pattern-matching problem in environment:
pat, p0 : ?T0 * ?T1
```
or (following the strategy introduced in #7062):
```coq
Definition foo := fun '(u, v) p2 => (u, v).
>                   ^^^^^^^^^^^^^^^^^^^^^^^^
Error: Cannot infer the domain of a subterm of the type
of this pattern-matching problem in environment:
pat, p0 : ?T0 * ?T1
```

Any opinions?

herbelin:
I think I'm finally going to backtrack on #7062 and to rather adopt the style `Cannot infer ?T in the partial instance "?T -> nat" found for the type of f`. This is both more informative and simpler to implement.

herbelin:
@mattam82: This PR is again ready for collecting opinions. I finally simplified things, tell me if you like it better.

herbelin:
@mattam82: if you like the new version of the error messages, this PR shall not take you a lot of time. If you don't like it, I can continue to work on it.

mattam82 (approved):
I like the new error message better indeed, and the code looks fine.

mattam82:
@herbelin This seems to have broken quite a few things in the CI.

herbelin:
> This seems to have broken quite a few things in the CI.

It is because I forgot to add `bug_` prefix to the bug number. Restarting.

ppedrot:
@herbelin Overlays have been merged downstream, you should probably relaunch / rebase this PR.

mattam82:
@herbelin I guess a rebase is all that's needed here.

herbelin:
@mattam82:
> @herbelin I guess a rebase is all that's needed here.

It seems indeed to have been the case. CI now all green.

mattam82:
I think this could be backported if you think it's worth it.

PR #6791, opened 2018-02-20
Removing compatibility support for versions older than 8.5.

Zimmi48:
**Kind:** clean-up

Following the conclusion at https://github.com/coq/coq/pull/887#issuecomment-364409212, which itself was following the conclusion of the WG at https://github.com/coq/coq/pull/884#issuecomment-351010823, I removed the deprecation for most compatibility notations, which allowed me in turn to remove the `VOld` and `V8_5` compatibility flags and to turn the warnings for the remaining deprecated notations on.

- [x] Entry added in CHANGES.

I guess the only documentation that's needed is an entry in CHANGES for the removal of the 8.5 compatibility support, right?

PR #8374, opened 2018-09-01
Remove -compat 8.6 and document the compat updates to do as part of a release.

### LINKED ISSUE #8311: Coq master -compat 8.8 should Import Coq.Compat.Coq88
#### Version

master (8.9+alpha)

This should compile with `coqc -compat 8.8`, but it does not:
```coq
Import Coq.Compat.Coq88.
```

This is absolutely mandatory for projects that use `-compat` options and are tested on Coq's CI.  Also critical for being able to test the compat options in the test-suite.

JasonGross:

Fixes / closes #8311

As requested in https://github.com/coq/coq/issues/8311#issuecomment-415976318, the release process now describes the steps to take.

All automatable steps are taken by the new script `dev/tools/update-compat.py`

I've tried to make the script relatively easy to update if functions get
renamed or moved, but since it's doing unstructured source manipulation,
it is sort-of fragile.

We could plausibly add a file to the test-suite to ensure that we catch
script-breakage early, but this would require dropping compatibility
support much earlier in the development cycle (the compatibility changes
would have to come right when the new version is branched, rather than
shortly before the beta release). 
(Is this a thing we want to do?  The test-suite update would just involve running `dev/tools/update-compat.py --assert-unchanged` on every build.)


Zimmi48:
>This would require dropping compatibility support much earlier in the development cycle (the compatibility changes would have to come right when the new version is branched, rather than
shortly before the beta release). Is this a thing we want to do?

I think so yes. Dropping compatibility helps the development move forward more freely so this seems like something that should happen early in the development cycle.

dev/tools/update-compat.py
```diff
@@ -0,0 +1,254 @@
...
+EXTRA_HEADER = '\n(** Compatibility file for making Coq act similar to Coq v%s *)\n'
+FLAGS_MLI_PATH = os.path.join(ROOT_PATH, 'lib', 'flags.mli')
+FLAGS_ML_PATH = os.path.join(ROOT_PATH, 'lib', 'flags.ml')
+COQARGS_ML_PATH = os.path.join(ROOT_PATH, 'toplevel', 'coqargs.ml')
+G_VERNAC_PATH = os.path.join(ROOT_PATH, 'vernac', 'g_vernac.mlg')
+TEST_SUITE_PATHS = tuple(os.path.join(ROOT_PATH, 'test-suite', 'success', i)
+                         for i in ('CompatOldFlag.v', 'CompatPreviousFlag.v', 'CompatCurrentFlag.v'))
+TEST_SUITE_DESCRIPTIONS = ('current-minus-two', 'current-minus-one', 'current')
+# sanity check that we are where we think we are
+assert(os.path.normpath(os.path.realpath(SCRIPT_PATH)) == os.path.normpath(os.path.realpath(os.path.join(ROOT_PATH, 'dev', 'tools'))))
```
Zimmi48: Does this mean that the script has to be run from inside the `dev/tools` directory? Some of the scripts in there assume on the contrary that they are run as `dev/tools/script` from the root of the repo.

dev/tools/update-compat.py
```diff
@@ -0,0 +1,254 @@
...
+
+def get_header():
+    with open(HEADER_PATH, 'r') as f: return f.read()
+
+HEADER = get_header()
+
+def get_version():
+    with open(CONFIGURE_PATH, 'r') as f:
+        for line in f.readlines():
+            found = re.findall(r'let coq_version = "([0-9]+\.[0-9]+)', line)
```
Zimmi48: Hopefully this will become much simpler in the future with a file containing almost nothing else than the version number.

Zimmi48 (approved):
Looks great! Thank you so much.

Can you also update the CODEOWNER file (adding yourself as the primary owner of the new script, you can add me as second if no other volunteer).

dev/tools/update-compat.py
JasonGross: No.  It assumes that the script lives in `dev/tools`, though; copying it somewhere else will not work.  I define `SCRIPT_PATH` based on `__file__` so that it does not matter what the current directory is.  Perhaps this deserves a comment?

Zimmi48: More comments is always good. But OK, this is great and robust :tada:

ejgallego (commented):
I have no idea how compat is supposed to work, it seems strange that we don't introduce a `V8_8` flag as soon as we branch.

Regarding the script, I am not sure if such a tool will prove to be too fragile.

Zimmi48:
>I have no idea how compat is supposed to work, it seems strange that we don't introduce a V8_8 flag as soon as we branch.

Indeed, @JasonGross I think it would make sense to move these new instructions to the first section "as soon as the previous version branched off master".

JasonGross:
I have updated the location of the instructions, have run the script (and found two minor typos in it when running it a second time), updated CODEOWNERS, and added it to the CI.

Note that this means that we will drop support for `-compat 8.7` as soon as `v8.9` is branched.

JasonGross:
Updated things to also include editing `doc/stdlib/index-list.html.template` (which I had forgotten about)

JasonGross:
Rebased again to take care of compat notations (another thing that I forgot about)

dev/tools/update-compat.py
JasonGross: Updated with comment

JasonGross:
Removing `-compat 8.6` breaks a large number of contributions, because it removes all of the deprecated compatibility notations.  I don't want to be responsible for fixing the test-suite and preparing overlays for everything for removing `-compat 8.6`.  What should I do?  I can split off the part of the PR that updates the compat versions and the test-suite, or I can update the script to keep around four `-compat` options rather than three...

Zimmi48:
Thanks @JasonGross, this is great but this goes now way beyond 8.8.2 material. Please open a new PR with just 50e1afa97b9b0efb8f9b6769ce9a1d8bed529223 that I can backport. As for finishing the present PR, indeed this is not a task that you can do alone.

Zimmi48:
@JasonGross Regarding the use of deprecated compatibility aliases, you can at least fix the ones that are used in fiat-parsers and cross-crypto.
Recall that you were the one to ask me in #6791 to keep the deprecation of a few aliases:

>I think that any notations of the form `Zfoo := Z.foo` or `Nfoo:= N.foo` or probably even `Pfoo:= Pos.foo` should be deprecated; if I recall correctly, the argument for non-deprecation was that sometimes the deprecated names are shorter or more sensible, but I think this argument doesn't really hold for names that differ only by a `.`.

Zimmi48:
Let's open issues asking maintainers to make their project build with `-w "+compatibility-notation"`.

Zimmi48:
I created issues on all the projects that were using these notations for sure. I am not sure for Coquelicot, GeoCoq, Iris and QuickChick which depend on SSReflect, formal-topology and Corn which depend on Math-Classes, and VST which depends on CompCert.

JasonGross:
@Zimmi48 #8387

JasonGross:
I've fixed fiat-crypto and made a PR for HoTT.  I'll leave cross-crypto for @andres-erbsen (I don't actually work on that project), and it looks like it actually stems from fcf not being compatible, though cross-crypto doesn't seem to be tracking fcf master...?

ppedrot (approved):
LGTM.

gares (approved):

JasonGross:
Note that first first three commits must be backported to the v8.9 branch before the beta (else `-compat 8.9` won't work correctly on the beta), and the final commit must not be backported to v8.9.

Zimmi48:
Then it would be simpler to put the last commit in a separate (dependent) PR.

JasonGross:
I can do that, but then this PR won't pass the CI, because #8444 was merged before this PR, and this PR adds a test to the CI to make sure that the compat updater is a no-op on the current code base (to ensure that we catch it early if it's going to stop working).

Zimmi48:
Then keep it like this.

Zimmi48:
And then the usual process is to put this in the 8.10+beta1 milestone, and once it is merged to open a v8.9-only PR with the bits to backport.

herbelin (approved):
Looks ok for what I looked at. Leaving it to more experts than me.

maximedenes (approved):
codeowners changes ok

Zimmi48:
@JasonGross Unfortunately, as I just commented in #8383, the removal of the compatibility notations is *not* going to happen in Coq 8.9 so could you please split up from this PR what should absolutely get in Coq 8.9, without the test-suite update since it would break CI (probably without the script or the release-doc updates either), and with a commit postponing the compat version of compatibility notations instead of removing them? Then we can work at merging this PR right after the 8.9.0 release (but without any kind of pressure).

Zimmi48:
>Are you asking for 2?

I'm asking for 2.

>Which would include the script and the updates to releasing and the CI check.

You can't include the update to the test-suite though because as soon as you do this then you have to remove compat 8.7 as well and then it implies removing the compatibility notations as well. Anyways, I advise you to only put the minimum required for 8.9 in this PR so that we can take time to do the rest properly and devise about the future of compat (in particular, I doubt that removing compat should always happen as soon as we bump the version number).

JasonGross:
> You can't include the update to the test-suite though because as soon as you do this then you have to remove compat 8.7 as well and then it implies removing the compatibility notations as well.

There will be separate updates to the test-suite: one for removing -compat 8.6, and another for removing -compat 8.7

> in particular, I doubt that removing compat should always happen as soon as we bump the version number

I think we should have a consistent policy for this.  The policy might be "master supports 4 compat versions, releases support 3", which would say that we remove the old compat versions right before we branch.  Other policies are also possible, but I think we need to pick one.

JasonGross:
@Zimmi48 I've made the version for 8.9 at #8634

JasonGross:
Closing this as superseded by #8634, #8637, #8638

PR #8376, opened 2018-09-01
Source basic overlay after user overlays to fix #8375 following #8348.

### LINKED ISSUE #8375: Instructions for overlays do not work
#### Version

8.9+alpha

The instructions for overlays say:
> If the fork is not on the same platform (e.g. GitHub instead of GitLab), it is
necessary to redefine the `_CI_ARCHIVEURL` variable as well.

But this is misleading.  I have
```bash
if [ "$CI_PULL_REQUEST" = "8064" ] || [ "$CI_BRANCH" = "numeral-notation-4" ]; then
    HoTT_CI_REF=fix-for-numeral-notations
    HoTT_CI_GITURL=https://github.com/JasonGross/HoTT
fi
```
and in [this build](https://gitlab.com/coq/coq/-/jobs/93836648), it issues
```bash
wget https://github.com/HoTT/HoTT/archive/fix-for-numeral-notations.tar.gz
```
which is wrong.  So it seems the `_CI_ARCHIVEURL` variable always needs to be redefined.  I believe #8348 is the culprit.

cc @Zimmi48

Zimmi48:
**Kind:** infrastructure fix.

Fixes / closes #8375 

~~Not tested yet.~~

### DIFF
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -1448,10 +1448,10 @@ function make_addons {
     export CI_BRANCH=""
     export CI_PULL_REQUEST=""
   fi
-  . /build/ci-basic-overlay.sh
   for overlay in /build/user-overlays/*.sh; do
     . "$overlay"
   done
+  . /build/ci-basic-overlay.sh
 
   for addon in $COQ_ADDONS; do
     "make_addon_$addon"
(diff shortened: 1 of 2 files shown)


SkySkimmer (approved):

Zimmi48:
@SkySkimmer Is it OK if I push an additional commit to this branch fixing this?

SkySkimmer:
Really? When I `git checkout svdfd` it fails.

SkySkimmer:
Also since there's no conflict put it in another PR.

Zimmi48:
>Really? When I git checkout svdfd it fails.

I think it depends on your git version.

>Also since there's no conflict put it in another PR.

Ah sorry, I had pushed it already. I can still undo this if you want.

Zimmi48:
The overlays were also loaded in the Windows build script so the loading order had to be changed there as well.

PR #8394, opened 2018-09-03
Print the entire string to the CoqIDE screen, e.g. for "Print Options"

### LINKED ISSUE #8378: "Print Options" produces no output in CoqIDE
#### Version

~~8.8+~~ 8.9+alpha

Running the command "Print Options." in CoqIDE (from an editor panel) produces no output.  It should.

### LINKED ISSUE #8398: Incorrect printing of record types in coqide
#### Version

8.9

```
Record foo := { bar : nat }.
Print foo.
```

When run in Coqide, this gives

```
Record foo : Set := Build_foo { bar : nat
For Build_foo: Argument scope is [nat_scope]
```

So, there is as missing bracket, as well as a missing line break.

jfehrle:
If the string matches the regex, output the part before the regex match with full highlights.  (Duh)

**Kind:** bug fix.

Fixes / closes #8378

I think this is critical to include in 8.9.

This is a case where unit testing would be helpful, though it would require stubbing out some GTK/CoqIDE dependencies--infrastructure we don't have today.

EDIT: fixes #8398 too

### DIFF
--- a/ide/ideutils.ml
+++ b/ide/ideutils.ml
@@ -71,15 +71,15 @@ let insert_xml ?(mark = `INSERT) ?(tags = []) (buf : #GText.buffer_skel) msg =
   let rmark = `MARK (buf#create_mark buf#start_iter) in
   (* insert the string, but don't apply diff highlights to white space at the begin/end of line *)
   let rec insert_str tags s =
+    let etags = try List.hd !dtags :: tags with hd -> tags in
     try
-      let _ = Str.search_forward nl_white_regex s 0 in
+      let start = Str.search_forward nl_white_regex s 0 in
+      insert_with_tags buf mark rmark etags (String.sub s 0 start);
       insert_with_tags buf mark rmark tags (Str.matched_group 1 s);
       let mend = Str.match_end () in
       insert_str tags (String.sub s mend (String.length s - mend))
-    with Not_found -> begin
-      let etags = try List.hd !dtags :: tags with hd -> tags in
+    with Not_found ->
       insert_with_tags buf mark rmark etags s
-      end
   in
   let rec insert tags = function
   | PCData s -> insert_str tags s


jfehrle:
I think the failure is unrelated--this PR only changes IDE code.

Off topic: there is only one failure here, ci-hott, that gets reported twice.  Would be nice to only report it once.

Zimmi48:
@jfehrle The fact that it is reported twice is a feature, not a bug: it gives you a direct link to the GitLab CI failure. Anyways, that failure is unrelated. I restarted it to get an all green CI, but this shouldn't prevent proceeding with the merge in any case.

ppedrot (approved):
Rereading this code, I am wondering why you did not use the tag mechanism to output the diff information directly on the nodes instead of relying on this kind of hacks. I think I remember you had problems with diffing boxes, but still, the current solution strikes me as a hack more than anything else. The more I look back at the diff mechanism the more I grow queasy, but for the time being let's forget the implementation as long as it works as intended. 

I am also expecting performance issues now but this is the kind of problems that only occur with @JasonGross-style huge outputs, so I'll consider it a minor annoyance.

ejgallego:
> The more I look back at the diff mechanism the more I grow queasy, but for the time being let's forget the implementation as long as it works as intended.

Indeed the system is supposed to emit tags, why these are lost in CoqIDE, I am not sure.

Indeed I think that the diff algorithm could take a different approach, but the code should be modular enough as for this not to be a problem once we can use a better diff method [likely html-based diff]

jfehrle:
I'm open to improving this later on, particularly if we run into performance issues.  I expected fairly small outputs (a screenful) and that performance considerations are unimportant here.  insert_str is O(length * # of newlines) because of the String.sub and String.length.  It could be rewritten to be O(n) though that may be a bit less readable.

The Pp.t structure isn't ideal for representing diffs.  As @ejgallego suggests, HTML is a better though not maybe not a perfect fit, either.  Every significant piece of software has parts that are elegant and parts that are less elegant.  The realistic question is whether we've made a good choice balancing the cost/benefit of a new feature while considering whether to refactor or redesign the code the feature interfaces to.

Thanks for the review.

ejgallego:
> The Pp.t structure isn't ideal for representing diffs. As @ejgallego suggests, HTML is a better though not maybe not a perfect fit, either.

I wonder what do you mean by "representing". Certainly HTML/Pp.t seem to me more than good enough for the purpose of _sending the diff information to the UI_.

By using tags, (`div/span`) the UI should be able to do a pretty good job IMHO.

jfehrle:
> I wonder what do you mean by "representing".

The current semantics of Pp.t are not an ideal fit.  But it's what we have today.  Trying to work around those makes the code more complex.

> By using tags, (div/span) the UI should be able to do a pretty good job IMHO.

Pretty good, yes, but if the browser picks the line breaks, you end up highlighting whitespace at the beginning or end of the line.  Doesn't look so good.

ejgallego:
> Pretty good, yes, but if the browser picks the line breaks, you end up highlighting whitespace at the beginning or end of the line. Doesn't look so good.

Yeah line breaks should go away in the next revision of `Pp.t` hopefully.

PR #8400, opened 2018-09-04
[doc] also scan plugins/ to build the lirbary index

### LINKED ISSUE #8220: doc of ssrbool.v is not listed on the website
it looks like that theories/ is listed, but plugins/ is not

gares:
CC: math-comp/math-comp#215
Fix: #8220 

This is only a partial fix since:
- wrt SSReflect, the doc generated skips the headers, that is the most important thing, but at least we have one less problem to fix (the files are processed by coqdoc and indexed now!)
- wrt the rest of plugins/*/*.v, to make the doc compile I had to declare they should be hidden. I imagine some of them should not be, but in this case one also need to classify the files an update the index template. I'm not competent to do this.


Zimmi48:
Overall the impression that gives me is: this system is way too complicated, coqdoc should come with a standard procedure on how to document libraries and Coq should use it. @ejgallego maybe Dune can bring improvements in this area as well?

Regarding this particular PR, it LGTM but I'm not very confident when looking at Makefile hacks so @maximedenes can I let you review and merge?

gares:
I don't touch any makefile, just a shell script that fills in the index "template".

Yes, horribly complex and especially "manual". You add a file, and you have to fix the template...

Zimmi48:
Right, that's a script not a Makefile (but one called from a Makefile), not that this makes me more confident ^^
Of course, if nobody else takes this, I will still "review" and merge it.

gares:
I've just rebased this an included it in #8408 so that we can see the output.
Tell me if I should just close this one, all in all if #8408 is not OK and merged, I don't want this one to be merged either.

Zimmi48:
If you consider the two as being so tied up, then yes, it makes sense to close this one.

PR #8408, opened 2018-09-05
[doc] [ssr] Fix rendering

gares:
We make the files in plugins/ssr/*.v look nicer when rendered in HTML.
We used the same sed code we use in math-comp in order to mangle the sources.
We also did some adjustments by hand, such has hacking the .css in place.

Let's have CI build the html for us and let's see how it looks like.

CC @amahboubi @CohenCyril 

depends on #8400


Zimmi48:
What prevents from applying the same styling tricks to the full stdlib if that makes the rendering nicer?

Regarding CI, something odd is going on: maybe of the jobs are marked as skipped: https://gitlab.com/coq/coq/pipelines/29448515

SkySkimmer:
Is there some reason we can't directly put in the `<style>` element instead of going through javascript?

gares:
> What prevents from applying the same styling tricks to the full stdlib if that makes the rendering nicer?

In these files there are comments that use indentation in a "semantic" way.
Hence we need a monospace font, otherwise things looks very confusing.

Typically monospace fonts are not used for text (as it is the case in the standard css) and I believe all comments are text, or use others means (like bulleting) to give a structure.

gares:
> Is there some reason we can't directly put in the <style> element instead of going through javascript?

I guess I could do that, but I don't know if `<style>` can be put everywhere (coqdoc lets you put arbitrary html code only inside `<body>...<div class="doc">`)

gares:
> Regarding CI

I've no clue, I did nothing special. I could not change the status of skipped of `ci-refman` either.

SkySkimmer:
I think it's fine https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style
>The <style> element can be included inside the <head> or <body> of the document, and the styles will still be applied

Zimmi48 (approved):

Zimmi48:
As you mentioned, we won't be able to see the output before #8400 is merged.

gares:
Ci does not like me :-/ This time the job failed because of a cryptographic error, whatever that means... will we ever be able to see the html?

gares:
noo, it does not build stdlib, it just build sphinx. I guess I need some help to change ci and artifacts upload. Anyone?

Zimmi48:
Hum, yeah I think that was one of the drawbacks of moving this to a test-phase job. @SkySkimmer?

SkySkimmer:
stdlib doc should be built by the build:base job (due to the `EXTRA_TARGET: "stdlib"` bit) but it doesn't get installed because I didn't notice it was missing.
Possibilities:
- we make all build-template jobs build the stdlib doc and install it
- we add some INSTALL_EXTRA_TARGET variable and use it to install stdlib in the build:base job

We could also try to build it in a separate job but there were some issues whose details I don't remember so maybe not for this PR.

gares:
@SkySkimmer I need help, it seems I'm not able to re-use the already compiled coq

gares:
I've rebased this PR, that I forgot to push forward. It is quite important for mathcomp that the doc of ssr*.v in Coq is made available.

@coq/ci-maintainers please check the ci part that was impacted by a conflict and that I don't fully understand.

gares:
@Zimmi48 I've just rebased since I had a spurious CI failure I believe.

I'm not so sure on how to proceed here. The CI part is to have this part of the html being rendered too. I could remove last commit if it is problematic (I don't know how to fix CI, really).

Can we have this merged by 8.9 is out?

Zimmi48:
I don't have any problem with testing this manually and merging this without any CI update if that's needed. I'm not quite sure I understand what the current situation and blocking points are.

Zimmi48:
>Yes. Is the rest of the stdlib also in the pdf refman?

I'm not talking about the pdf refman but the pdf doc of the stdlib (i.e. the `doc/stdlib/Library.pdf` output of `make doc`). It looks like it is already scarce compared to what is in the HTML version anyways.

gares:
Didn't even know t exists ;-) and I guess it is not made from the template I modified, otherwise they would be in sync... Maybe we should open an issue for that problem if it is already present.

gares:
Thanks for merging. Do you think someone (you or silene or ?) could populate https://coq.inria.fr/distrib/V8.9+beta1/stdlib/ ? I see the refman but not the stdlib. Just to be sure how it will look like on the website before the final release.

Zimmi48:
@gares I'll forward you an e-mail about that.

PR #8409, opened 2018-09-05
[vernac] Restrict the “deprecated” attribute to a few vernacular commands

### LINKED ISSUE #8382: Deprecated attribute should handle Definition as well and emit a warning when it doesn't handle a command.
#### Version

8.9+alpha

The deprecated attribute (documented here: https://gitlab.com/coq/coq/-/jobs/artifacts/master/file/_install_ci/share/doc/coq/sphinx/html/language/gallina-specification-language.html?job=documentation#attributes) supports Ltac definitions and tactic notations. It should be extended to also support Gallina definitions and notations (currently the only way of deprecating a definition is to define a compatibility alias, i.e. abusing the compat system, see also #6840).

All the commands that do not support the deprecated attribute should fail when it's used (actually I would be fine with only emitting a warning). It is completely inconsistent that
```
#[deprecated(since="8.9.0", note="Use True instead.")] Set Universe Polymorphism.
```
succeeds without even emitting a warning, while
```
#[polymorphic] Set Universe Polymorphism.
```
fails with
```
This command does not support Polymorphism
```

vbgl:
Motivated by #8382.

### DIFF
--- a/vernac/vernacentries.ml
+++ b/vernac/vernacentries.ml
@@ -2262,6 +2262,12 @@ let check_vernac_supports_polymorphism c p =
     | VernacExtend _ | VernacUniverse _ | VernacConstraint _) -> ()
   | Some _, _ -> user_err Pp.(str "This command does not support Polymorphism")
 
+(* Vernaculars that may handle a deprecation flag *)
+let check_vernac_supports_deprecation c d =
+  if d <> None then match c with
+  | VernacExtend _ -> ()
+  | _ -> user_err (Pp.str "This command does not support deprecation")
+
 (** A global default timeout, controlled by option "Set Default Timeout n".
     Use "Unset Default Timeout" to deactivate it (or set it to 0). *)
 
@@ -2401,6 +2407,7 @@ let interp ?(verbosely=true) ?proof ~st {CAst.loc;v=c} =
     | c ->
       check_vernac_supports_locality c atts.locality;
       check_vernac_supports_polymorphism c polymorphism;
+      check_vernac_supports_deprecation c atts.deprecated;
       let polymorphic = Option.default (Flags.is_universe_polymorphism ()) polymorphism in
       Flags.make_universe_polymorphism polymorphic;
       Obligations.set_program_mode atts.program;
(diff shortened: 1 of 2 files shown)


vernac/vernacentries.ml
```diff
@@ -2262,6 +2262,12 @@ let check_vernac_supports_polymorphism c p =
     | VernacExtend _ | VernacUniverse _ | VernacConstraint _) -> ()
   | Some _, _ -> user_err Pp.(str "This command does not support Polymorphism")
 
+(* Vernaculars that may handle a deprecation flag *)
+let check_vernac_supports_deprecation c d =
+  if d <> None then match c with
+  | VernacExtend _ -> ()
+  | _ -> user_err (Pp.str "This command does not support deprecation")
+
```
ejgallego: I am sorry but I fail to see what the logic is here.

What you really want IMO anyways is a table that maps commands to supported features, and it is built from the declaration of the vernac.

vbgl: You are perfectly right. We need to design a good way for defining what attributes are supported and how they are handled.

My reasoning was that since the issue #8382 had a high-priority label, it would deserve this kind of quick-and-dirty patch mimicking what is done for other attributes. I agree that it’s quite bad to ignore attributes without notice.

ejgallego: I mean, this check will forbid `#[deprecated] Definition`, right?

vbgl: Yes, it does. Is there already a mechanism for deprecating definitions?

Zimmi48:
>Is there already a mechanism for deprecating definitions?

Not yet, although this was kind of the other half of my request at #8382. I understand if you don't want to be the one implementing it, but then this PR should not be marked as closing #8382 completely. The high priority label was indeed more about the part that you are fixing in this PR.

maximedenes:
can this get an assignee? @mattam82 ?

herbelin (changes requested):
Otherwise, same comments as @ejgallego. The interest in a generlc attribute mechanism is that it is managed by means of extensible tables. Having a `check_vernac_supports` function for each attribute shall not scale. I don't remember, do you already have plans to move to an extensible mechanism?

PR #8250, opened 2018-08-14
Introduce a team of code owners for the documentation.

Zimmi48:
As was previously done for the CI, this means that there are no more principal / secondary code owners, and we can update the team of owner without changing the `CODEOWNERS` file. All the members of the team can choose to review and self-assign any documentation PR that is not their own. All the members of the team receive notifications for all new PRs affecting the `doc/` folder.

For now the team members are @cpitclaudel @maximedenes @Zimmi48. It will be better for me since I will get review requests (I was already acting as principal maintainer) and it will allow @cpitclaudel to review and merge my own documentation PRs.

@silene I didn't put you in the @coq/doc-maintainers team yet. Should I do it?

PR #8411, opened 2018-09-05
Changes to default.nix to be able to use Dune.

Zimmi48:
**Kind:** infrastructure.

Question in passing: @maximedenes @vbgl should we form a nix-team (like there is a ci-team #8023 and a doc-team #8250) for shared code ownership of the Nix files?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -23,8 +23,8 @@
 
 { pkgs ?
     (import (fetchTarball {
-      url = "https://github.com/NixOS/nixpkgs/archive/060a98e9f4ad879492e48d63e887b0b6db26299e.tar.gz";
-      sha256 = "1lzvp3md0hf6kp2bvc6dbzh40navlyd51qlns9wbkz6lqk3lgf6j";
+      url = "https://github.com/NixOS/nixpkgs/archive/4477cf04b6779a537cdb5f0bd3dd30e75aeb4a3b.tar.gz";
+      sha256 = "1i39wsfwkvj9yryj8di3jibpdg3b3j86ych7s9rb6z79k08yaaxc";
     }) {})
 , ocamlPackages ? pkgs.ocaml-ng.ocamlPackages_4_06
 , buildIde ? true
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -45,6 +45,7 @@ stdenv.mkDerivation rec {
   buildInputs = [
     hostname
     python2 time # coq-makefile timing tools
+    jbuilder # why is this not part of ocamlPackages and why isn't it called dune yet?
```
vbgl: What is the purpose of this comment? Here are answers to the questions.

`jbuilder` is no OCaml library; its behavior does not depend on the version of the OCaml compiler used to compile it; therefore, it does not belong to the `ocamlPackages` set.

It is indeed not renamed yet, probably by lack of manpower.

Zimmi48: Thanks. It was mostly an poor-man's inline comment. I'll remove it.

Zimmi48:
Note that #8412 is requiring Dune 1.1 so we can either wait to merge this PR until nixpkgs has the updated version, or wait to merge #8412 instead.

Zimmi48:
#8412 is going to get merged, so let's wait for a nixpkgs update before merging this one.

ejgallego:
> #8412 is going to get merged, so let's wait for a nixpkgs update before merging this one.

The NiX update should be trivial @Zimmi48 , Dune 1.1 is fully compatible with 1.0

vbgl:
I just tried: it breaks the `re` tests (at least with `nixpkgs` on darwin)… This is irrelevant for Coq but it shows that this is neither “trivial” nor “fully compatible”.

What can be done, is to override it in our `default.nix`:

```
let dune =
  overrideDerivation jbuilder (o: {
    name = "dune-1.1.1";
    src = fetchFromGitHub {
      owner = "ocaml";
      repo = "dune";
      rev = "1.1.1";
      sha256 = "0v2pnxpmqsvrvidpwxvbsypzhqfdnjs5crjp9y61qi8nyj8d75zw";
    };
  });
in
```

Zimmi48:
Good idea @vbgl. Let's do this.

Zimmi48:
Maybe I should move `dune` to the shell-only part until we actually start using this in `nix-build` as well...

ejgallego:
> it breaks the re tests (at least with nixpkgs on darwin)

Fairly odd, that looks like a Dune bug. So far the update on OPAM only broke a handful of developments and that was due to a bug in Dune 1.1.0 [fixed in 1.1.1] But yeah, that's a young tool.

vbgl (approved):

PR #8412, opened 2018-09-05
[dune] [ci] Fix and test release profile + use 1.1 dune-workspace

ejgallego:
Dune 1.1 allows us to define the `env` flags in the workspace file,
which is a better place than the current situation.

Along the way, We fix the build flags for release mode [missing
`-rectypes` and add a `release` build profile CI job.


.gitlab-ci.yml
```diff
@@ -217,10 +215,17 @@ build:edge+flambda:
...
   <<: *dune-template
   variables:
     OPAM_SWITCH: edge
+    DUNE_TARGET: world
+
+build:egde:dune:release:
+  <<: *dune-template
+  variables:
+    OPAM_SWITCH: edge
+    DUNE_TARGET: release
```
Zimmi48: The number of different build jobs is becoming a bit too much :(

dune-workspace
```diff
@@ -0,0 +1,6 @@
+(lang dune 1.1)
+
+; Add custom flags here. Default developer profile is `dev`
+(env
+ (dev     (flags :standard -rectypes -w -9-27-50))
+ (release (flags :standard -rectypes)))
```
Zimmi48: I don't get it. We are silencing some warnings in dev mode but not in release mode? Why would we want users to see warnings that we don't see?

Zimmi48:
If #8411 is merged first, then this PR will break it. If you want to proceed and merge this, #8411 would have to wait that Dune is updated to 1.1 in nixpkgs.

gares:
It is not fully in topic but not off topic either: in another PR it was pointed out that dune natively supports a "profile" that uses opaque in order to speed up re-compilation. Could we have that too? Is it already there?

gares (approved):
for what I understand, this it ok

.gitlab-ci.yml
ejgallego: This would be the seed for the OPAM packaging job, it could be moved along the `pkg:nix` job tho.

dune-workspace
ejgallego: The `env` profile uses a much stricter set of warnings than the `release` one.

ejgallego:
> It is not fully in topic but not off topic either: in another PR it was pointed out that dune natively supports a "profile" that uses opaque in order to speed up re-compilation. Could we have that too? Is it already there?

This the built-in `dev` profile which is the default if you don't use `-p` .

.gitlab-ci.yml
ejgallego: In fact the system is ready to generate the OPAM packages now, but I was planning to wait to:
- add documentation build target
- add test-suite build target
- fix the couple of bugs around [such as coqide path detection]
- wait for the better version number system to land in Dune.

Zimmi48: Fine by me to do this in several steps. But indeed, do move this next to the pkg:nix job. If fact, once the documentation / test-suite targets are ready, we could even use this to build the Nix package.

ejgallego: Ok, moving to `pkg:dune-release`

.gitlab-ci.yml
```diff
@@ -234,6 +233,12 @@ windows32:
   except:
     - /^pr-.*$/
 
+pkg:dune-release:
+  <<: *dune-template
+  stage: test
```
SkySkimmer: When putting stuff in the test stage we should use `dependencies: []` to avoid downloading the various built coqs.

SkySkimmer: In fact let's put it in the dune-template as that seems simpler.

ejgallego: Good point, thanks.

Makefile.dune
```diff
@@ -10,10 +10,11 @@ BUILD_CONTEXT=_build/default
...
 	@echo "Welcome to Coq's Dune-based build system. Targets are:"
-	@echo "  - states: build a minimal functional coqtop"
-	@echo "  - world:  build all binaries and libraries"
-	@echo "  - clean:  remove build directory and autogenerated files"
-	@echo "  - help:   show this message"
+	@echo "  - states:  build a minimal functional coqtop"
+	@echo "  - world:   build all binaries and libraries"
+	@echo "  - release: build Coq in release mode"
+	@echo "  - clean:   remove build directory and autogenerated files"
+	@echo "  - help:    show this message"
```
SkySkimmer: Do we really want to align stuff here? It ends up making updates annoying.

ejgallego: Given that this is a document I consider formatting important. The alternative would be to use a table generator.

dune-workspace
SkySkimmer: I don't want to look up the dune defaults to figure out which warnings are on, can we do like in configure and use `+a...`?

.gitlab-ci.yml
```diff
@@ -234,6 +233,12 @@ windows32:
   except:
     - /^pr-.*$/
 
+pkg:dune-release:
```
SkySkimmer: What's this `pkg` prefix?

dune-workspace
ejgallego: We could do our own profile indeed, but here I think I do prefer relying on Dune's. It is one thing less to worry about, and profiles can be always be printed easily.

.gitlab-ci.yml
ejgallego: We took that prefix for package manager dependent builds. We have NiX, then this one in "release" will build an opam package (and maybe upload it).

An important consideration of the release mode tho is that is uses stronger optimizations, so maybe the flambda version should be built like that. I dunno.l, lots of room here.

SkySkimmer (changes requested):
Request fix for the `dependencies` issue.

@Zimmi48 @ejgallego after that do I merge or do I wait for nix stuff to happen?

.gitlab-ci.yml
ejgallego: > In fact let's put it in the dune-template as that seems simpler.

I've leave it like this, but note that Dune has no problem in reusing previous builds in an incremental way.

Zimmi48:
@SkySkimmer You can proceed with merging.

.gitlab-ci.yml
SkySkimmer: >I've left it like this, but note that Dune has no problem in reusing previous builds in an incremental way.

Including across profiles?
And there's certainly no point in downloading the various `make` based builds.

SkySkimmer (approved):

.gitlab-ci.yml
ejgallego: Builds live in "contexts", the only one is default but you for example could declare one for `bisect_pxx` etc... they are independent.

`(env ...)` profiles are supposed to be tracked in the dependency chain, so indeed if you change some stuff or profile it may trigger a rebuild. Note that this is still being developed AFAICT.

ejgallego: > And there's certainly no point in downloading the various make based builds.

Sure, I was just mentioning in case in the future we need it. The dependency tracking is more useful for the developer tho.

PR #8415, opened 2018-09-05
[bin] Fix binary location procedure to work with symlinks.

ejgallego:
In #7860 `Sys.executable_name` was introduced to locate coq private
binaries; however, this breaks use cases with symlinks, such as Dune.

In order to fix this, we adopt a simpler strategy; we will always
adapt the name originally provided by the user, and only add a
directory if it was originally present.

This should work (hopefully) in all cases.

This fixes `coqide` not working on Dune builds.

### DIFF
--- a/ide/ideutils.ml
+++ b/ide/ideutils.ml
@@ -328,15 +328,18 @@ let coqtop_path () =
       | None ->
         try
           let new_prog = System.get_toplevel_path "coqidetop" in
-            if Sys.file_exists new_prog then new_prog
+            (* The file exists or it is to be found by path *)
+            if Sys.file_exists new_prog ||
+               CString.equal Filename.(basename new_prog) new_prog
+            then new_prog
 	    else
 	      let in_macos_bundle =
 		Filename.concat
 		  (Filename.dirname new_prog)
 		  (Filename.concat "../Resources/bin" (Filename.basename new_prog))
 	      in if Sys.file_exists in_macos_bundle then in_macos_bundle
-                 else "coqidetop"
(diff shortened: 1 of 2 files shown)


ejgallego:
@Zimmi48 this should fix the coqide problem.

ppedrot (commented):
This doesn't look very good to me to replace a function that is semantically supposed to be the binary name with an access to the argument array, without a real specification. I understand this as a workaround, but I think it should probably be considered and reported upstream as a bug of the OCaml function.

ejgallego:
> This doesn't look very good to me to replace a function that is semantically supposed to be the binary name with an access to the argument array, without a real specification.

What do you mean? So far both the old behavior and the new one a fully specified AFAICT.

But we change the strategy, indeed, the way we want to locate binaries now is based on the original name the user/script provides; not on the real binary location which can be something strange due to symlinks.

The strategy is:
- if the provided name doesn't contain a path (basename == real_name) then don't add any path and rely on `PATH` for location.
- otherwise, if the user specified a particular binary path, use it.

> I understand this as a workaround, but I think it should probably be considered and reported upstream as a bug of the OCaml function.

This is not a workaround, but a change on strategy. The upstream function is supposed to resolve symlinks so I dunno a bug would be accepted.

ejgallego:
Note that there are quite a few instances in the codebase where location of things is a mess or broken wrt symlinks, hopefully a next PR will unify and fix all of them.

ppedrot (approved):
All right, that's good enough for me. Name resolution is a mess anyways.

ejgallego:
Merci !

Issue #7192, opened 2018-04-07
Print Assumptions confused by nested modules


andrew-appel:
#### Operating system: Windows
In the following program, `Print Assumptions E.D.f` should report "closed" but wrongly believes that `E.D.f` is an axiom.

```
Definition a := True.

Module Type B.
Axiom f : Prop.
End B.

Module Type C.
Declare Module D : B.
End C.

Module E: C.
Module D <: B.
Definition f := a.
End D.
End E.

Print Assumptions E.D.f.
  (* Axioms: 
      E.D.f : Prop
  *)
Print Assumptions a.
  (* Closed under the global context *)
```

herbelin:
Hi @andrew-appel, I think you mean `E <: C`. Otherwise your redefinition of `D` is not taken into account.

andrew-appel:
But I claim that E.D.f is supported by a proof, that is, "a" is a witness for "f", and therefore "f" is not an assumption.  I should be able to have an opaque module that has a proof, a witness, for its implementation.

herbelin:
You're right, this is a bug. `Print Assumptions` looks at the implementation of modules (bypassing the apparent signature) but it does so only at the first level of nesting. This should be extended to submodules.

PR #8417, opened 2018-09-05
Fixing #8416: Print Assumptions missing module information from compiles files

### LINKED ISSUE #8416: Print Assumptions spuriously reports axioms from module types defined in another file

Define the following in a file "test.v":

```
Module Type A.
  Axiom f : True.
End A.

Module M : A.
  Lemma f : True. trivial. Qed. 
End M.
```

Then, if we have in another file

```
Require Import test.

Lemma test : True.
Proof. apply M.f. Qed.

Module N : A.
  Lemma f : True. trivial. Qed. 
End N.

Lemma test2 : True.
Proof. apply N.f. Qed.

Print Assumptions test.
Print Assumptions test2.
```

The first print assumption command claims there's an axiom:
```
Axioms:
M.f : True
```
while the second properly reports that the term is closed under the global context. Under 8.8.0, both are reported as closed. This is causing lots of spurious assumptions to be listed in my developments that use mathcomp libraries.

herbelin:
This fixes the fix 1522b989 to #7192: an incorrect permutation of code in 1522b989 was losing the case `MPfile`. Because of that, the original fix to #7192 was also not effective for instances of #7192 across files.

**Kind:** bug fix

Fixes / closes #8416 

### DIFF
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -106,7 +106,8 @@ SUBSYSTEMS := $(VSUBSYSTEMS) misc bugs ide vio coqchk coqwc coq-makefile unit-te
 
 PREREQUISITELOG = prerequisite/admit.v.log			\
   prerequisite/make_local.v.log prerequisite/make_notation.v.log \
-  prerequisite/bind_univs.v.log
+  prerequisite/bind_univs.v.log prerequisite/module_bug8416.v.log \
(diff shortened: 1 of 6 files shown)


mattam82 (approved):
The fix of the fix looks good to me, and the test-suite files are really helpful!

Issue #8371, opened 2018-08-31
Better documentation of the dependencies needed to compile the doc


amahboubi:
#### Version
 8.9+alpha (August 2018)

#### Operating system

Linux (Ubuntu 16.04)

#### Description of the problem

The description of the [dependencies](https://github.com/coq/coq/tree/master/doc#dependencies) does not mention the versions required for the Python packages. At least for some older version of Ubuntu, it would be useful. See issue #8223 for an example of failures.

Zimmi48:
@cpitclaudel said:
> The import issue you ran into could have been due to an incorrect recent change, too (I think we were originally hoping to not depend on anything too recent in Sphinx; it could be that I or someone else recently used a new feature without meaning to)

We should probably do the same for documentation dependencies as we do for the rest of our dependencies: i.e. pin them to our minimal required versions in our CI.

cc @ejgallego

Zimmi48:
I don't know either, but indeed, pinning to the minimal required versions would probably require going through pip for all of these dependencies, unless we decide that the lower bound is what Ubuntu bionic provides.

cpitclaudel:
> Doc dependencies are installed thru apt thus distro-pinned already.

Cool, so I didn't discretely break everything :)

>  however I am unsure how does pinning work in pip

You can add `==version` after the name of the package: `pip3 install --user sphinx==1.7.8`

ejgallego:
> unless we decide that the lower bound is what Ubuntu bionic provides.

I think that's a good compromise; similarly we will use OCaml 4.05.0 for 8.10 hopefully which is what Ubuntu / Debian stable use. That seems a safe choice for most people, while not being _too_ old.

Zimmi48:
OK, so it only remains to be documented.

Zimmi48:
The lower version bounds of the documentation dependencies are now documented. As for the upper version bounds, we strive to support the latest versions although at the current time, this is not the case (see #9126). We view this as a bug that we should fix. Nonetheless, it could make sense to document explicitly what the latest tested version is...

Issue #7390, opened 2018-04-29
[doc] Moving away from LaTeX


ejgallego:
After the completion of the Sphinx migration [#6818], there are only residual TeX files in the tree; we all tend to agree that they would rather be moved to other formats. In today's `master` we can see the remaining files:

```
egallego@rochefort$ find . -name *.tex
./vernac/doc.tex
./interp/doc.tex
./lib/doc.tex
./pretyping/doc.tex
./proofs/doc.tex
./library/doc.tex
./tactics/doc.tex
./kernel/doc.tex
```
I will take care of removing these once #7155 lands, I will move the corresponding text there to the proper files. We are waiting on a new `odoc` upstream release for that tho + Dune.
```
./dev/v8-syntax/syntax-v8.tex
./dev/v8-syntax/memo-v8.tex
./dev/ocamldoc/header.tex
./dev/doc/naming-conventions.tex
./dev/doc/versions-history.tex
./dev/doc/newsyntax.tex
./dev/doc/minicoq.tex
```
No idea on what to do with these files? Should be converted to MarkDown / RST ?
```
./doc/tools/Translator.tex
./doc/common/title.tex
./doc/common/version.tex
./doc/common/macros.tex
./doc/stdlib/Library.coqdoc.tex
./doc/stdlib/Library.tex
./doc/tutorial/Tutorial.tex
./doc/tutorial/Tutorial.v.tex
./doc/whodidwhat/whodidwhat-8.5update.tex
./doc/whodidwhat/whodidwhat-8.2update.tex
./doc/whodidwhat/whodidwhat-8.4update.tex
./doc/whodidwhat/whodidwhat-8.3update.tex
./doc/RecTutorial/recmacros.tex
./doc/RecTutorial/RecTutorial.tex
./doc/RecTutorial/coqartmacros.tex
```
I guess a few of these files should be ported to Sphinx; dunno for the rest.

SkySkimmer:
I don't know how coqdoc works, does it go through latex when producing other formats?

ejgallego:
> I don't know how coqdoc works, does it go through latex when producing other formats?

I does not.

ejgallego:
Note that the test suite does indeed depend on LaTeX so we are not gonna be moving so fast from it [unless we only test the html targets].

Zimmi48:
I would already be interested in having less dependencies for users who do not run the test-suite.

Zimmi48:
To answer your initial questions:

```
[...]

Issue #7378, opened 2018-04-28
Copyright dates in the manual are "1999-1970" when building with Nix.


ejgallego:
Originally to be seen at https://coq.inria.fr/distrib/current/refman/

Now, you need to build the manual with Nix to observe this behavior.

Zimmi48:
This has been fixed in master and v8.8.

ejgallego:
Well, this was closed but the website is still wrong.

maximedenes:
@Zimmi48 how did you build it? This is a bug I've observed on NixOS, due to timestamp 0 being put on sphinx files. I'm surprised you don't have the same issue since you use NixOS too.

Zimmi48:
So indeed, I don't know how I did it, but I'm back to seeing "1999-1970" again :(

Zimmi48:
Actually I know where I had seen correct dates: in the GitLab artifacts. So just use the artifact instead of what you generate on your machine next time you upload the manual.

maximedenes:
Yes, in fact the deployment should be automated of course. I plan to do that once we move the website to a standard provider.

Zimmi48:
Could you create an issue (on the website repo) so that others are informed of this plan?

Zimmi48:
BTW, if the copyright dates are based on a timestamp, I'm having a hard time understanding what this is useful for:

https://github.com/coq/coq/blob/b2f746e41abf53fc481f90804ba4d70edd73fc86/doc/sphinx/conf.py#L74

Zimmi48:
Fixed as part of https://coq.inria.fr/distrib/V8.8.1/refman/ by using the artifact from GitLab CI.
However, I'd still like to understand whether using timestamps for this is a good idea and what is the purpose of the line quoted above.

Zimmi48:
cc @cpitclaudel FTR I still see "© Copyright 1999-1970, Inria." when building on NixOS loading dependencies using `nix-shell`. I don't know if it's reproducible on non-NixOS when loading dependencies using `nix-shell` or if it goes away on NixOS when installing Python packages using `pip` (have never tried yet).

ejgallego:
Is this fixed @Zimmi48 ? [Seems so]

Zimmi48:
This is a Nix-specific problem that is still there. The workaround is to use the artifact generated by GitLab CI.

ejgallego:
Ok so should we retitle the bug?

ejgallego:
[...]

PR #8418, opened 2018-09-05
Add a PDF version of the manual

### LINKED ISSUE #7500: PDF version of the manual
@cpitclaudel said it would be doable to extend the current Sphinx infrastructure to also be able to generate a single-page PDF document of the reference manual (like it used to be done with the LaTeX manual).

Although not as important as it used to be, it would be neat to have such version, for instance for embedding in the Zenodo record, like it was done for previous versions (cf. https://zenodo.org/record/1174360).

cpitclaudel:
**Kind:** documentation / infrastructure.

I've tried to split this into reasonably-sized commits. There were mistakes in some of the LaTeX formulae that the HTML build didn't catch, and in the bib file; I also used this opportunity to document more DOs and DONTs in the documentation's dev guide.

There are plenty of further tweaks we could make to make the manual look prettier, but this is enough to build it.

Closes #7500


ejgallego:
Note this is in partial conflict with #7390

cpitclaudel:
@ejgallego How so?

ejgallego:
> @ejgallego How so?

We thought we would remove the dependency on LaTeX from the doc system, but indeed we may keep it then and adjust the issue correspondingly.

Zimmi48:
Indeed, this PDF manual has been asked for by several people and if in addition it allows to catch more errors, I can only conclude that #7390 will be a "won't fix". Note that I myself have a use case for the PDF version of the manual which is when updating the Zenodo record.

doc/sphinx/credits.rst
```diff
@@ -1,12 +1,5 @@
-.. include:: preamble.rst
-.. include:: replaces.rst
-
 .. _credits:
 
--------------------------------------------
-Credits
--------------------------------------------
```
Zimmi48: The Credits chapter must be a proper chapter, independent of the introduction (see https://coq.inria.fr/refman/credits.html), so why remove the title from here?

doc/sphinx/index.latex.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Introduction
+------------
+
+.. include:: introduction.rst
+
+Credits
+-------
+
+.. include:: credits.rst
```
Zimmi48: Credits must not be part of the introduction, and certainly not come before the table of contents. See also #6573.

Makefile.doc
```diff
@@ -58,11 +58,19 @@ ifndef QUICK
...
+	$(HIDE)$(SPHINXENV) $(SPHINXBUILD) -b $* \
+		$(ALLSPHINXOPTS) doc/sphinx $(SPHINXBUILDDIR)/$*
+
+sphinx-pdf: sphinx-latex
+	$(SHOW)'LATEXMK $(SPHINXBUILDDIR)/latex'
+	$(HIDE)+$(MAKE) -C $(SPHINXBUILDDIR)/latex
+
+sphinx:
+	+$(MAKE) sphinx-html
+	+$(MAKE) sphinx-pdf
```
Zimmi48: The `sphinx-pdf` target should be included in the `doc-pdf` target. There are also quite a few places which mention the `sphinx` target that should be fixed to use `sphinx-html` instead (the other solution would be to keep `sphinx` with its current meaning and add a new `sphinx-all` target).

doc/sphinx/README.rst
```diff
@@ -364,6 +362,32 @@ DON'T
...
+
+DO
+  .. code::
+
+     :n:`name={*; attr}`
+
+DON'T
+  .. code::
+
+     ``name=``:n:``{*; attr}``
```
Zimmi48: I guess you meant to use single quotes on the right side.

doc/sphinx/README.rst
```diff
@@ -377,6 +401,56 @@ DON'T
...
+-----------------------------------------------------------
+
+DO
+  .. code::
+
+     A tactic of the form:
+
+     .. coqdoc::
+
+        do [ t1 | … | tn ].
```
Zimmi48: Note that currently there is a rendering problem (with margins) when using the `.. coqdoc::` directive which encourages people to use `.. coqtop:: in` instead.

doc/sphinx/README.template.rst
```diff
@@ -153,6 +179,86 @@ DON'T
...
+        If you assume the axiom above, …
+
+DON'T
+  .. code::
+
+     Here is a useful axiom:
+
+     .. example::
+
+        .. code::
```
Zimmi48: You probably meant `coqdoc` here and BTW this part is only in the template but not in the generated README.

doc/sphinx/biblio.bib
```diff
@@ -136,7 +136,7 @@ @InProceedings{Del00
   pages       = {85--95},
   month       = {November},
   year        = {2000},
-  url         = {http://www.lirmm.fr/\%7Edelahaye/papers/ltac\%20(LPAR\%2700).pdf}
+  url         = {http://www.lirmm.fr/%7Edelahaye/papers/ltac%20(LPAR%2700).pdf}
```
Zimmi48: The problem when you omit the escaping is that the end of the line gets interpreted as a comment.

doc/sphinx/conf.py
```diff
@@ -66,11 +73,38 @@
...
+
 # The master toctree document.
-master_doc = 'index'
+# We create this file in `copy_master_doc` above.
+master_doc = "index"
 
 # General information about the project.
-project = 'Coq'
+project = 'The Coq Reference Manual'
 copyright = '1999-2018, Inria'
```
Zimmi48: Can you tell me about what is the use of this copyright line? It seems it isn't used because the build uses timestamps instead. That's not a good idea because when building on Nix, timestamps are reset to a default value (1970). This doesn't really make sense from a copyright law point of view either. So it would be best if this line was actually used.

doc/sphinx/conf.py
```diff
@@ -257,57 +293,57 @@
...
-#latex_logo = None
-
-# For "manual" documents, if this is true, then toplevel headings are parts,
-# not chapters.
-#latex_use_parts = False
+# latex_logo = "../../ide/coq.png"
 
 # If true, show page references after internal links.
-#latex_show_pagerefs = False
+# latex_show_pagerefs = True
```
Zimmi48: Changing a commented value sounds strange ^^

doc/sphinx/index.latex.rst
```diff
@@ -0,0 +1,81 @@
...
+
+License
+-------
+
+.. include:: license.rst
+
+The language
+------------
+
+.. toctree::
```
Zimmi48: No indexes in this table of matters? And any way to share this between the two versions. I'm foreseeing people adding a chapter and forgetting to update the PDF table of matter or moving chapters around but only in one version.

doc/sphinx/language/gallina-extensions.rst
```diff
@@ -85,8 +83,8 @@ To build an object of type :n:`@ident`, one should provide the constructor
 .. _record-named-fields-grammar:
 
   .. productionlist::
-    term : {| [`field_def` ; … ; `field_def`] |}
-    field_def : name [binders] := `term`
+    record_term : {| [`field_def` ; … ; `field_def`] |}
+    field_def : name [binders] := `record_term`
```
Zimmi48: I think `term` was really meant there and the above should have been an *extension* of the term syntax.
If that helps, I think that this should not be presented in the extension chapter but in the specification chapter anyways, as part of the main grammar.

doc/sphinx/language/gallina-extensions.rst
```diff
@@ -157,9 +155,9 @@ It can be activated for printing with
 .. _record_projections_grammar:
 
   .. productionlist:: terms
-     term : term `.` ( qualid )
-          : | term `.` ( qualid arg … arg )
-          : | term `.` ( @`qualid` `term` … `term` )
+     projection : projection `.` ( `qualid` )
+          : | projection `.` ( `qualid` `arg` … `arg` )
+          : | projection `.` ( @`qualid` `term` … `term` )
```
Zimmi48: Same remark. I think `term` was really meant but that should have been an extension. I also think that this should be part of the main grammar in the first chapter.

doc/sphinx/practical-tools/coqide.rst
```diff
@@ -263,8 +261,8 @@ for the ∀ symbol. A list of symbol codes is available at
 An alternative method which does not require to know the hexadecimal
 code of the character is to use an Input Method Editor. On POSIX
 systems (Linux distributions, BSD variants and MacOS X), you can
-use `uim` version 1.6 or later which provides a :math:`\LaTeX`-style input
-method.
+use `uim` version 1.6 or later which provides a
+:math:`\text{\LaTeX}`-style input method.
```
Zimmi48: Can't we just say LaTeX? This sounds overly complicated just to get a fancy font.

doc/tools/coqrst/coqdomain.py
```diff
@@ -1108,4 +1144,7 @@ def setup(app):
     # Tell Sphinx about extra settings
     app.add_config_value("report_undocumented_coq_objects", None, 'env')
 
-    return {'version': '0.1', "parallel_read_safe": True}
+    meta = { "version": "0.1",
+             "env_version": 1, # Increment this when the contents of env are changed
```
Zimmi48: I'm not sure where this `env` is defined or what will happen if someone forgets to increment this.

Zimmi48 (changes requested):
Thanks a lot for this. Overall, what I understands LGTM, apart from a few comments below.

doc/sphinx/biblio.bib
Zimmi48: Or maybe it's just GitHub which is stupid.

Zimmi48:
You have two failures:

- a warning treated as an error
  ```
  /builds/coq/coq/doc/sphinx/coq-cmdindex.rst:2:Field list ends without a blank line; unexpected unindent.
  ```

- the `default.nix` file is set to only build the `doc-html` target, and yet it now tries to build the pdf target as well:
  ```
  LATEXMK doc/sphinx/_build/latex
  make[4]: Entering directory '/builds/coq/coq/nix-build-coq.drv-0/324syvibpz8ghvmby850xcj4lvivp280-source/doc/sphinx/_build/latex'
  latexmk -pdf -dvi- -ps-  'CoqRefMan.tex'
  /nix/store/8zkg9ac4s4alzyf4a8kfrig1j73z66dw-bash-4.4-p23/bin/bash: latexmk: command not found
  make[4]: *** [Makefile:32: CoqRefMan.pdf] Error 127
  ```

anton-trunov:
Hi @cpitclaudel, thank you for this PR.
Is there a rendered PDF available?

Zimmi48:
@anton-trunov 
>Is there a rendered PDF available?

Our CI will generate one as soon as the build is fixed.

doc/sphinx/index.latex.rst
cpitclaudel: > Credits must not be part of the introduction

They are not :)

> and certainly not come before the table of contents. See also #6573.

They don't ^^

The source of the confusion might be that in LaTeX, the table of contents is generated from the `toctree`s, but it always appears at the very beginning of the document.

cpitclaudel:
Thanks for proofreading! I agree with \approx all comments :) I'll make the relevant fixes.

doc/sphinx/index.latex.rst
Zimmi48: OK. Better like this than nothing.

doc/sphinx/README.rst
Zimmi48: No sorry, never filled an issue for this. Here is an example though: https://coq.inria.fr/refman/proof-engine/ssreflect-proof-language.html#pattern-assignment (the first block is a coqdoc).

doc/sphinx/index.latex.rst
Zimmi48: OK perfect. I couldn't check the rendered version yet so I had to guess from the code.

Makefile.doc
cpitclaudel: Thanks for pointing this out.  Question: do you know why `install-doc-printable` and `install-doc-sphinx` have no dependencies?

doc/sphinx/conf.py
cpitclaudel: I think the line is used, actually: at the bottom of the refman that's current online I see "© Copyright 1999-2018, Inria.".  Where is it not using it?

doc/sphinx/index.latex.rst
cpitclaudel: Indices are built automatically, and in LaTeX builds (unlike in HTML) you don't have to add them explicitly to the TOC.

doc/sphinx/language/gallina-extensions.rst
cpitclaudel: I added FIXMEs

doc/tools/coqrst/coqdomain.py
cpitclaudel: I improved the comment

cpitclaudel:
I've processed most comments, and added questions to the ones I wasn't sure how to fix.
The warning is a bit of an issue: it's a sphinx bug, which got fixed in more recent versions.  Which version is running on the CI? https://github.com/sphinx-doc/sphinx/issues/2505

doc/sphinx/credits.rst
cpitclaudel: Argh, that one is tricky.  Top-level titles have to be in the index file for LaTeX.  I'll look for a fix.

cpitclaudel: I wrapped the credits file into a `credits-wrapper` file.  It's ugly, but I don't think it'll hurt maintainability.

doc/sphinx/README.rst
cpitclaudel: OK, I have a patch for this, based on this branch.  I'll post it after we merge this.

Makefile.doc
Zimmi48: I don't. It's probably a mistake.

doc/sphinx/conf.py
Zimmi48: This line at the bottom was generated using the timestamp. See https://coq.inria.fr/distrib/V8.8.0/refman/ which was built on NixOS.

doc/sphinx/introduction.rst
```diff
@@ -41,6 +35,11 @@ are processed from a file.
    The `coqtop` read-eval-print-loop can also be used directly, for debugging
    purposes.
 
+   .. [#PG] Proof-General is available at https://proofgeneral.github.io/.
```
Zimmi48: I want to check where the footnote is rendered in the HTML output.

Zimmi48 (approved):

anton-trunov:
>  Our CI will generate one as soon as the build is fixed.
Thank you, @Zimmi48 

@cpitclaudel I saw the PDF you sent to the coqdev mailing list -- looks very cool, indeed. Thanks again!

Is it possible to include one more level into the PDF's table of contents? The old refman had three levels of nesting, i.e. one could easily navigate to a section `x.y.z`, whereas the current one only supports two levels (`x.y`). For instance, I can only navigate to "The SSReflect proof language" section, but not to its subsections.

Zimmi48:
Regarding the minimal required version, as mentioned in #8371, CI is currently testing with the version provided in Ubuntu bionic. Given that we are already requiring some dependencies that can only be installed using pip, I don't have a problem bumping the minimal required version for Sphinx itself (and fix the doc of the dependencies and how to get them). You'll need to interact with @ejgallego regarding the update of the dependencies in the Docker image.

(Maybe the only down-side is that I would like to include this PR in 8.8.2 --to have a PDF version of the manual for 8.8-- but bumping dependency requirements in a patch-level release is usually considered bad practice; OTOH these minimal dependencies for the Sphinx doc were never documented in 8.8 as recently noted on Coq-Club --doc/README.md is not part of 8.8.1-- and in #8371.)

doc/sphinx/conf.py
cpitclaudel: What sorcery is this? ^^ OK, I'll check.

cpitclaudel: I tried `faketime -f "-10y" make sphinx-html` and changing to 2016 instead of 2018.  In both cases the copyright line didn't change.  Could there be something nix-specific going on?

Zimmi48: Let's continue this discussion in #7378.

Zimmi48:
@cpitclaudel I can take care of updating the minimal Sphinx dependency if you confirm that we need 1.8.0 or above. Ideally, we should now the minimal required version for all the Python packages so that we pin them all.

Zimmi48:
@cpitclaudel I don't have the permission to push to your branch but here b7d1fe281a1c87131cf3ca8405ea082647cc42bb is a commit on top of (a rebased version of this branch) which updates the required version of Sphinx to 1.7.0 in the documentation and in the CI settings.

While doing this, I realized that this PR creates a (very frequent) race condition in the Coq build when building (from a clean state) with `make -j4 doc`. This is definitely related to LaTeX because `make -j4 doc-html` doesn't have this race. Could a Makefile-competent person review the relevant parts of this PR? (@SkySkimmer, @gares or @ejgallego?)

SkySkimmer (commented):
@Zimmi48 can you say more about the race condition?

Makefile.doc
```diff
@@ -58,17 +58,25 @@ ifndef QUICK
...
-	@echo "Build finished. The HTML pages are in $(SPHINXBUILDDIR)/html."
+# sphinx-html and sphinx-latex
+sphinx-%: $(SPHINX_DEPS)
+	$(SHOW)'SPHINXBUILD doc/sphinx ($*)'
+	$(HIDE)$(SPHINXENV) $(SPHINXBUILD) -b $* \
+		$(ALLSPHINXOPTS) doc/sphinx $(SPHINXBUILDDIR)/$*
+
+sphinx-pdf: sphinx-latex
+	$(SHOW)'LATEXMK $(SPHINXBUILDDIR)/latex'
+	$(HIDE)+$(MAKE) -C $(SPHINXBUILDDIR)/latex
```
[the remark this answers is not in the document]
cpitclaudel: Good point, thanks; fixed.

Makefile.doc
```diff
@@ -58,17 +58,25 @@ ifndef QUICK
...
+
+sphinx:
+	+$(MAKE) sphinx-html
+	+$(MAKE) sphinx-pdf
```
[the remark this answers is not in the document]
cpitclaudel: I'm not sure these two targets are safe to run in parallel.  I'm basing that reasoning off of https://github.com/sphinx-doc/sphinx/issues/2946 and the fact that Sphinx caches are shared across build targets.

cpitclaudel:
> @cpitclaudel I don't have the permission to push to your branch but here b7d1fe2 is a commit on top of (a rebased version of this branch) which updates the required version of Sphinx to 1.7.0 in the documentation and in the CI settings.

Thanks, very nice :) I've pushed that patch to the branch.  I took the liberty to change the code to require 1.8.0 (since that's what the commit message said). I've also checked the "allow commit from maintainers box", sorry about that.

ejgallego (changes requested):
Ummm, I'd rather avoid going back to `pip` if possible.

Also, 1.8.0 will make Coq impossible to package even in recent distributions. This is a serious issue. Reading the PR and the linked sphinx issues it is not clear to me what the required minimal version is, so let's first find that and try to work out with the _minimum_ possible.

Another concern with the current patch is the lack of reproducibility / pinning in the sphinx call. Note that this is already a but in the current version of the Dockerfile.

cpitclaudel:
> Also, 1.8.0 will make Coq impossible to package even in recent distributions. This is a serious issue. 

Indeed, good point.  Is the current dependency on `antlr4-python3-runtime` OK because we check in the autogenerated parser files?

> Reading the PR and the linked sphinx issues it is not clear to me what the required minimal version is, so let's first find that and try to work out with the minimum possible.

The only problem, I think, is a warning on some older versions of Sphinx.  I can try to bisect to see which one.

> Another concern with the current patch is the lack of reproducibility / pinning in the sphinx call. Note that this is already a but in the current version of the Dockerfile.

I'm not sure I understand that part :)

cpitclaudel:
> The only problem, I think, is a warning on some older versions of Sphinx. I can try to bisect to see which one.

Done. 1.6.7 is bad; 1.7.0 is good. Not sure how to see which version is packaged by Ubuntu (I don't know how to interpret `Version: 1.3.6-2ubuntu1.2`)?

The issue is just a warning; ignoring it is fine:

```
/build/coq/pdf/doc/sphinx/coq-cmdindex.rst:2: WARNING: Field list ends without a blank line; unexpected unindent.
/build/coq/pdf/doc/sphinx/coq-exnindex.rst:2: WARNING: Field list ends without a blank line; unexpected unindent.
/build/coq/pdf/doc/sphinx/coq-optindex.rst:2: WARNING: Field list ends without a blank line; unexpected unindent.
/build/coq/pdf/doc/sphinx/coq-tacindex.rst:2: WARNING: Field list ends without a blank line; unexpected unindent.
/build/coq/pdf/doc/sphinx/genindex.rst:2: WARNING: Field list ends without a blank line; unexpected unindent.
```

There's another issue related to this bit of code:

```
    def _target_id(self, name):
        # Use `name[1]` instead of ``nodes.make_id(name[1])`` to work around
        # https://github.com/sphinx-doc/sphinx/issues/4983
        return 'grammar-token-{}'.format(name[1])
```

That issue is fixed in 8.0, so the code should be updated to read `nodes.make_id(name[1])` when we switch to 8.0.

@ejgallego , what do you think / what do you recommend?

SkySkimmer:
>Not sure how to see which version is packaged by Ubuntu (I don't know how to interpret Version: 1.3.6-2ubuntu1.2)?

It should mean 1.3.6, plus some patches either backported from later versions or custom to the distribution.

Zimmi48:
>can you say more about the race condition?

- https://travis-ci.com/Zimmi48/coq/jobs/145252746#L5477-L5479
- https://travis-ci.com/Zimmi48/coq/jobs/145252748#L5691-L5693
- https://travis-ci.com/Zimmi48/coq/jobs/145225042#L4198-L4200 (ltac file too short, most common error)
- https://travis-ci.com/Zimmi48/coq/jobs/145194901#L4812-L4814

ejgallego:
> Indeed, good point. Is the current dependency on `antlr4-python3-runtime` OK because we check in the autogenerated parser files?

This unversioned dependency is currently a bug of our Dockerfile. Note however that a person willing to package Coq would have an easier time getting a single dependency in shape than 10.

> I'm not sure I understand that part :)

The version for all dependencies installed by pip should be specified so we get a deterministic build.

Regarding package versions, here you can see the versions shipped by Debian/Ubuntu:
- https://packages.ubuntu.com/search?suite=all&arch=any&searchon=names&keywords=python-sphinx
- https://packages.debian.org/search?keywords=python-sphinx

So indeed the optimal thing would be to keep us in 1.6.7, however if we really need to up the dependency we can to 1.7.9 for the moment, at the cost of leaving Ubuntu stable users alone.

Zimmi48:
>The version for all dependencies installed by pip should be specified so we get a deterministic build.

:+1: I didn't do it because I didn't know what values to put.

ejgallego:
>  I didn't do it because I didn't know what values to put.

Yeah the thing is that using `apt` defaults saves us quite a bit of effort in this front.

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -9,11 +9,10 @@ ENV DEBIAN_FRONTEND="noninteractive"
 RUN apt-get update -qq && apt-get install --no-install-recommends -y -qq \
         m4 automake autoconf time wget rsync git gcc-multilib opam \
         libgtk2.0-dev libgtksourceview2.0-dev \
-        texlive-latex-extra texlive-fonts-recommended texlive-science tipa \
-        python3-sphinx python3-pexpect python3-sphinx-rtd-theme python3-bs4 python3-sphinxcontrib.bibtex \
-        python3-setuptools python3-wheel python3-pip
+        texlive-full python3-pip python3-setuptools
```
ejgallego: `texlive-full` seems also overkill here;  note that the size of the Docker image is an important variable.

Zimmi48: OK I see the problem. Compared to the packages that were previously there, one additional dependency is `xelatex`.

ejgallego: As I was fearing image size has increased by almost 1GiB (https://gitlab.com/coq/coq/container_registry) IMO we cannot take that.

This PR seems a bit too large BTW, wouldn't it make sense to start integrating some of the features?

SkySkimmer: texlive-full depends on various heavy doc packages which we should be able to avoid.
eg on my system there is `texlive-latex-extra-doc` at 534MB uncompressed.
This is debian so maybe ubuntu does it differently but I doubt that.

BTW maybe we should `apt-get clean` to get rid of the downloaded .deb files after we're done installing.

cpitclaudel: > This PR seems a bit too large BTW, wouldn't it make sense to start integrating some of the features?

I'm fine with that :) But what features are you thinking of?

cpitclaudel:
>  Ah sorry for the commit message but was 1.8.0 actually released (and not just beta)? My testing shows that 1.7.0 is good enough to make the build go through without warnings. And setting to 1.8.0 would just fail at pip install.

1.8.0 is available now :)

> at the cost of leaving Ubuntu stable users alone.

I'm not sure what that entails, so I'd like to delegate that decision to people who do.

I think I've highlighted the two issues with using non-current versions: with < 1.7.0 we get a few spurious warnings, and with < 1.8.0 we generate a few non-standard HTML anchors.

My preference to fix the first issue is to update to 1.7.0, and on platforms that can't do that to use 1.6+ with `SPHINXWARNERROR=0`.  To fix the second issue, I think coqdomain.py should branch on Sphinx's version number.  I will submit that fix after we merge this.

cpitclaudel:
> Yeah the thing is that using apt defaults saves us quite a bit of effort in this front.

Does it? How do people on non-`apt` systems know what to install?

ejgallego:
It does saves us from setting versions numbers ourselves as it is a curated set of packages

cpitclaudel:
Yes, of course. But what versions do we tell people on macOS, Fedora, or Windows to install?

Zimmi48:
1.8.0 is available now indeed and guess what? It broke the build of Coq's refman: https://travis-ci.org/coq/coq/builds
Fortunately, Travis caught this early because we didn't pin any package. I think we should keep it like this: pin to the minimum version dependency in the Docker image for GitLab CI and not pin in the Travis configuration to always test the latest version.

Regarding splitting this PR up, yes if you want to push the commits fixing parts of the doc without changing the infrastructure to another PR, I can merge it more swiftly. To merge the present PR with the LaTeX infrastructure, it will be necessary to understand how to avoid the above-mentioned race condition.

Finally, regarding texlive-full, indeed very sorry about this. Please backtrack on this change but note that it is important to add the new dependency on xelatex.

Zimmi48:
I hope we can finish and merge this PR this week-end because it's the last thing blocking 8.8.2 (BTW, it deserves a CHANGES entry in the 8.8.2 section).

Zimmi48:
>A deprecation warning treated as error, right?

Not clear to me:

```
Running Sphinx v1.8.0
/usr/local/lib/python3.4/dist-packages/sphinx/util/requests.py:72: UserWarning: Some links may return broken results due to being unable to check the Server Name Indication (SNI) in the returned SSL cert against the hostname in the url requested. Recommended to install requests-2.4.1+.
  'Some links may return broken results due to being unable to '
Extension error:
Could not import extension coqrst.coqdomain (exception: cannot import name 'MathDirective')
```

>I'm traveling this week-end, so I'm not going to be very responsive, unfortunately. But with a list of issues, I can try to do a bit. Can you clarify what the race condition is? Also, which other things need changing at this point?

I am not able to reproduce this race condition anymore so never mind this. Other things to do:
- backtrack on the use of `texlive-full` in Travis and Docker;
- pin as much as possible in Docker (to the minimum supported version, and update the README accordingly) and unpin in Travis to let it catch errors with latest versions early;
- add an entry in CHANGES in the 8.8.2 section.

I think that's it. As for the minimum Sphinx version, given that we will indeed abandon Ubuntu / Debian stable packagers, feel free to pick 1.8.0 if you like.

Zimmi48:
To start the Docker build, some manual operation is needed. You'll need to ask someone to do it, or request to join the gitlab.com coq organization and start it manually yourself (`SKIP_DOCKER` needs to be `false`).

ejgallego:
> given that we will indeed abandon Ubuntu / Debian stable packages

I am not against this in principle but note that using 1.8.0 has far more serious problems as it means Coq cannot be packaged even for the unstable version.

I'd say that unless some stronger motivation is given I am not willing to sacrifice packaging capabilities for such a fringe feature the PDF manual is (and it seems "bugs" in older versions are nits anyways)

Zimmi48:
@ejgallego Sorry, can you confirm that you are OK with moving to 1.7.0 (possibly 1.7.9) but not 1.8.0? I'm not sure I understood your comment correctly.

Zimmi48:
1. I see the fact that the PDF format was dropped in 8.8 without any discussion a regression, especially given that it was not initially intended that way by @cpitclaudel. Furthermore, I've received enough complaints / requests to consider that this is something important and that I consider a blocker for 8.8.2.
2. After pondering a bit, I don't think changing the minimum Sphinx version in 8.8.2 would be right so I propose to change it to 1.8.0 in master and v8.9 but when backporting to v8.8 I will rather disable the warning-as-failure.
3. I am willing to take care of fixing what needs to be fixed in CI files (@cpitclaudel is it OK if I force-push to your branch?) but I will still need help because...
4. The race condition problem is back: https://travis-ci.com/Zimmi48/coq/builds/85180121 (it probably hadn't disappeared but I didn't know how to reproduce it).

ejgallego:
> 2\. I propose to change it to 1.8.0 in master and v8.9 but when backporting to v8.8 I will rather disable the warning-as-failure.

IMVHO I'd try to keep the current sphinx version for 8.9.

8.10 will be disruptive due to the Ocaml version so that seems like a better moment to try to go to 1.7.9 [or hopefully 1.8.0 if Debian catches up]

Zimmi48:
OK but then we must accept to disable fatal Sphinx warnings in 8.9 (that sounds OK though given that it's just a few).

ejgallego:
> OK but then we must accept to disable fatal Sphinx warnings in 8.9 (that sounds OK though given that it's just a few).

Yeah I guess we can live with that in the 8.8 and 8.9 branch. So for master, I am OK moving to 1.7.9.

ejgallego:
> I see a few `make: Circular ide/.merlin <- submake dependency dropped.` messages. Are these expected, or new?

A rebase should get rid of those.

Zimmi48:
Hum, no @ejgallego this is a build on an already rebased branch.

BTW, CI passed for this branch on my fork (except the problem on Travis), so I'm going to force-push it here soon.

Zimmi48:
Once again, I didn't manage to reproduce the problem locally. That being said, I obtained this result on the first run:

```
[]\TU/lmr/m/n/10 Syntax error: [prim:reference] expected after

Underfull \hbox (badness 10000) in paragraph at lines 913--915
\TU/lmr/m/n/10 [prim:reference] (in [vernac:command])
[513] [514] [515]) (./CoqRefMan.aux) )
(see the transcript file for additional information)
Output written on CoqRefMan.pdf (518 pages).
Transcript written on CoqRefMan.log.
=== TeX engine is 'XeTeX'
Latexmk: Index file 'CoqRefMan.idx' was written
Latexmk: Log file says output to 'CoqRefMan.pdf'
Latexmk: All targets (CoqRefMan.pdf) are up-to-date
make[4]: Leaving directory '/home/theo/git/coq/doc/sphinx/_build/latex'
make[3]: Leaving directory '/home/theo/git/coq'
make[2]: Leaving directory '/home/theo/git/coq'
make[1]: Leaving directory '/home/theo/git/coq'
make: *** [Makefile:166: submake] Error 2
```

on the second one:

```
The LaTeX files are in doc/sphinx/_build/latex.
Run 'make' in that directory to run these through (pdf)latex
(use `make latexpdf' here to do that automatically).
make -C doc/sphinx/_build/latex
make[4]: Entering directory '/home/theo/git/coq/doc/sphinx/_build/latex'
latexmk -pdf -dvi- -ps-  'CoqRefMan.tex'
Latexmk: This is Latexmk, John Collins, 19 Jan. 2017, version: 4.52c.
Latexmk: All targets (CoqRefMan.pdf) are up-to-date
make[4]: Leaving directory '/home/theo/git/coq/doc/sphinx/_build/latex'
make[3]: Leaving directory '/home/theo/git/coq'
make[2]: Leaving directory '/home/theo/git/coq'
make[1]: Leaving directory '/home/theo/git/coq'
```

Another reason to worry is that this re-run printed twice:

```
Running Sphinx v1.7.6
```

Note that I was using `make -j4 doc` so it still seems that there is some kind of race there even if it did not provoke any failure.

cpitclaudel:
> Another reason to worry is that this re-run printed twice:

That is expected: we need two calls to Sphinx to generate both html and LaTeX.

ejgallego:
> Hum, no @ejgallego this is a build on an already rebased branch.

Umm, something is fishy then, this was supposed to be solved.

Zimmi48:
OK so in fact I had this error in my build log:
```
Error:
while loading /home/theo/git/coq/plugins/syntax/numeral_notation_plugin.cmxs:
error loading shared library: /home/theo/git/coq/plugins/syntax/numeral_notation_plugin.cmxs: file too short
```
and yet the build went all the way (up to producing the PDF manual but still showing the error as above). So I had misinterpreted the error above, it had nothing to do with the "Syntax error" I started my snippet at. And this also sort of explains why I couldn't reproduce the issue that was showing with Travis. For some reason, the build continues past the error when running with `make -j4 doc` whereas it does not when building with `make -j4` after adding `-with-doc yes` to configure. I just tried that and got:

```
COQC -noinit theories/Init/Tactics.v
OCAMLOPT -pack -o plugins/funind/recdef_plugin.cmx
rm -f bin/coqtop && cp bin/coqtop.opt bin/coqtop
COQC -noinit theories/Init/Prelude.v
make[3]: execvp: bin/coqtop: Text file busy
make[3]: *** [Makefile.build:829: theories/Init/Prelude.vo] Error 127
make[3]: *** Waiting for unfinished jobs....
OCAMLOPT -shared -o plugins/ltac/ltac_plugin.cmxs
OCAMLOPT  plugins/ltac/tauto.ml
OCAMLOPT  plugins/cc/g_congruence.ml
OCAMLOPT  plugins/firstorder/ground.ml
make[3]: Leaving directory '/home/theo/git/coq'
make[2]: *** [Makefile:166: submake] Error 2
make[2]: Leaving directory '/home/theo/git/coq'
make[1]: *** [Makefile.doc:71: sphinx] Error 2
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/home/theo/git/coq'
make: *** [Makefile:166: submake] Error 2
```

Running again from the failed position I got:
```
OCAMLOPT -shared -o plugins/btauto/btauto_plugin.cmxs
COQC -noinit theories/Init/Logic_Type.v
COQC -noinit theories/Init/Specif.v
gcc: error: plugins/btauto/btauto_plugin.cmxs.startup.o: No such file or directory
File "caml_startup", line 1:
Error: Error during linking
make[3]: *** [Makefile.build:751: plugins/btauto/btauto_plugin.cmxs] Error 2
make[3]: Leaving directory '/home/theo/git/coq'
make[2]: *** [Makefile:166: submake] Error 2
make[2]: Leaving directory '/home/theo/git/coq'
make[1]: *** [Makefile.doc:71: sphinx] Error 2
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/home/theo/git/coq'
make: *** [Makefile:166: submake] Error 2
```

Running again I got:

```
COQC      theories/Logic/JMeq.v
OCAMLOPT -pack -o plugins/funind/recdef_plugin.cmx
OCAMLOPT -pack -o plugins/funind/recdef_plugin.cmx
gcc: error: ./theories/Logic/.coq-native/NCoq_Logic_JMeq.cmxs.startup.o: No such file or directory
File "caml_startup", line 1:
Error: Error during linking
File "./theories/Logic/JMeq.v", line 147, characters 0-49:
Warning: Removed file ./theories/Logic/JMeq.vo
Error: Native compiler exited with status 2

make[3]: *** [Makefile.build:834: theories/Logic/JMeq.vo] Error 1
make[3]: *** Waiting for unfinished jobs....
COQC      theories/Program/Wf.v
COQC      theories/Program/Equality.v
COQC      theories/Sets/Finite_sets.v
File "./theories/Program/Equality.v", line 13, characters 15-19:
Error: Unable to locate library JMeq.

make[1]: *** [Makefile.build:836: theories/Program/Equality.vo] Error 1
make[1]: *** Waiting for unfinished jobs....
make[3]: Leaving directory '/home/theo/git/coq'
make[2]: *** [Makefile:166: submake] Error 2
make[2]: Leaving directory '/home/theo/git/coq'
make[1]: *** [Makefile.doc:71: sphinx] Error 2
make[1]: Leaving directory '/home/theo/git/coq'
make: *** [Makefile:166: submake] Error 2
```

Running again:

```
COQC      theories/Wellfounded/Lexicographic_Product.v
COQC      theories/Wellfounded/Union.v
COQC      theories/Classes/Morphisms_Prop.v
File "./theories/Wellfounded/.coq-native/NCoq_Wellfounded_Lexicographic_Product.native", line 1:
Error: I/O error: ./theories/Wellfounded/.coq-native/NCoq_Wellfounded_Lexicographic_Product.native: No such file or directory
File "./theories/Wellfounded/Lexicographic_Product.v", line 168, characters 0-9:
Warning: Removed file ./theories/Wellfounded/Lexicographic_Product.vo
Error: Native compiler exited with status 2

make[1]: *** [Makefile.build:836: theories/Wellfounded/Lexicographic_Product.vo] Error 1
make[1]: *** Waiting for unfinished jobs....
COQC      theories/Classes/Equivalence.v
:
COQC      theories/Wellfounded/Wellfounded.v
File "./theories/Wellfounded/Wellfounded.v", line 15, characters 15-36:
Error: Unable to locate library Lexicographic_Product.

make[3]: *** [Makefile.build:836: theories/Wellfounded/Wellfounded.vo] Error 1
make[3]: *** Waiting for unfinished jobs....
make[3]: Leaving directory '/home/theo/git/coq'
make[2]: *** [Makefile:166: submake] Error 2
make[2]: Leaving directory '/home/theo/git/coq'
make[1]: *** [Makefile.doc:71: sphinx] Error 2
make[1]: Leaving directory '/home/theo/git/coq'
make: *** [Makefile:166: submake] Error 2
```

And after that the compilation finally succeeded.

So there definitely *is* a race!

---

Regarding the minimal version, I didn't realize that @ejgallego had said 1.7.9 is OK for master for now and 1.8.0 if Debian catches up. But in fact, nixpkgs is even behind (1.7.8) and updating it to 1.7.9 or 1.8.0 is not trivial enough for me to do it, so I'd rather keep a minimal dependency equal or below 1.7.8 for now (for master).

ejgallego:
> I'd rather keep a minimal dependency equal or below 1.7.8 for now (for master).

Yeah, as we are gonna keep a lower one for 8.9, let's land the lower one in master for now too.

Definitively 1.8.0 / 1.7.9 is very fresh stuff to make it the _minimal_ required version.

Zimmi48:
>Yeah, as we are gonna keep a lower one for 8.9, let's land the lower one in master for now too.

We still need to bump it in master to avoid turning fatal warnings off. But 1.7.0 is probably enough. @cpitclaudel?

Zimmi48:
I believe that I've fixed the race condition. Regarding the Sphinx update I've settled on 1.7.0 because this is what's in nixos-18.03. But indeed, the bump will be for master only.
@ejgallego Can you approve the PR?
@cpitclaudel Can you check that you're fine with my changes?
I'd like to merge pretty soon.

doc/README.md
```diff
@@ -28,18 +28,18 @@ To produce the complete documentation in HTML, you will need Coq dependencies
 listed in [`INSTALL`](../INSTALL). Additionally, the Sphinx-based
 reference manual requires Python 3, and the following Python packages:
 
-    sphinx sphinx_rtd_theme beautifulsoup4 antlr4-python3-runtime pexpect sphinxcontrib-bibtex
+  - sphinx >= 1.8.0
```
ejgallego: this seems to contradict what is in the docker file?

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -7,13 +7,19 @@ LABEL maintainer="e@x80.org"
...
-
-RUN pip3 install antlr4-python3-runtime
+        # Dependencies of stdlib and sphinx doc
+        texlive-latex-extra texlive-fonts-recommended texlive-xetex \
+        latexmk xindy python3-pip python3-setuptools \
+        # Dependencies of source-doc and coq-makefile
+        texlive-science tipa
+
+# More dependencies of the sphinx doc
+RUN pip3 install sphinx==1.7.0 sphinx_rtd_theme beautifulsoup4 \
```
ejgallego: Do you need to install using pip the rest of python packages too?

Note that they should also be version-pinned as to obtain a reproducible build.

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -7,13 +7,19 @@ LABEL maintainer="e@x80.org"
 ENV DEBIAN_FRONTEND="noninteractive"
 
 RUN apt-get update -qq && apt-get install --no-install-recommends -y -qq \
+        # Dependencies of the test-suite and of external projects
```
ejgallego: This is inaccurate, for example `opam` and `gcc-multilib` are core dependencies of the image itself.

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -7,13 +7,19 @@ LABEL maintainer="e@x80.org"
...
         m4 automake autoconf time wget rsync git gcc-multilib opam \
+        # Dependencies of CoqIDE
```
ejgallego: Technically of `lablgtk`

ejgallego:
> @ejgallego Can you approve the PR?

I've launched a Docker build in gitlab, let's see how it goes first. https://gitlab.com/coq/coq/pipelines/30604790

Note a few comments meanwhile.

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -7,13 +7,19 @@ LABEL maintainer="e@x80.org"
...
-RUN pip3 install antlr4-python3-runtime
+        # Dependencies of stdlib and sphinx doc
+        texlive-latex-extra texlive-fonts-recommended texlive-xetex \
+        latexmk xindy python3-pip python3-setuptools \
+        # Dependencies of source-doc and coq-makefile
+        texlive-science tipa
+
+# More dependencies of the sphinx doc
+RUN pip3 install sphinx==1.7.0 sphinx_rtd_theme beautifulsoup4 \
+        antlr4-python3-runtime pexpect sphinxcontrib-bibtex
```
ejgallego: For example I don't see a reason you install `pexpect` using pip. Similarly for beautifuldsoup4.

ejgallego:
> I've launched a Docker build in gitlab, let's see how it goes first.

@Zimmi48 the whole OPAM ecosystem seems to have problem so lets see if the image can build.

Zimmi48:
@ejgallego I had already launched a build: https://gitlab.com/coq/coq/pipelines/30603958

doc/README.md
Zimmi48: Oops, thanks for noticing.

dev/ci/docker/bionic_coq/Dockerfile
Zimmi48: I did it like this because I didn't know which ones were depending on `sphinx`. But yes, I agree they should be version pinned, I just have trouble picking versions.

dev/ci/docker/bionic_coq/Dockerfile
Zimmi48: OK I will extend the comment.

dev/ci/docker/bionic_coq/Dockerfile
Zimmi48: OK will update.

Zimmi48:
Looking at the dependencies more closely, it seems that none of the other Python packages depend on Sphinx so I'll backtrack and put them back in the apt install step.

ejgallego:
> Looking at the dependencies more closely, it seems that none of the other Python packages depend on Sphinx so I'll backtrack and put them back in the apt install step.

Indeed; after looking at the image things are looking good I think; will have another look tonight to see how the comments are going and approve.

Remember to update the docker image version if you modify the dockerfile [that's a pain I know]

Zimmi48:
>Remember to update the docker image version if you modify the dockerfile [that's a pain I know]

I did.

ejgallego (approved):
Looks great to me, amazing job folks.

Zimmi48:
I'll merge as soon as the following two points are achieved:
- CI passes
- @cpitclaudel confirms he's happy with the three commits I pushed on top of his branch.

Zimmi48:
```
Exception occurred:
  File "/usr/lib/python3/dist-packages/pybtex/scanner.py", line 118, in required
    raise TokenRequired(description, self)
pybtex.scanner.TokenRequired: syntax error in line 27: '}' expected
The full traceback has been saved in /tmp/sphinx-err-3y2xouz5.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!

Exception occurred:
  File "/usr/lib/python3/dist-packages/pybtex/scanner.py", line 118, in required
    raise TokenRequired(description, self)
pybtex.scanner.TokenRequired: syntax error in line 27: '}' expected
The full traceback has been saved in /tmp/sphinx-err-8fh_2btg.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
Makefile.doc:63: recipe for target 'sphinx-latex' failed
```

What's going on? Does it mean that even 1.7.0 is not good?

ejgallego:
> What's going on? Does it mean that even 1.7.0 is not good?

Umm, maybe you need to install the sphinx-related packages with pip? I guess there are 2 candidates.

Makefile.doc
```diff
@@ -67,9 +67,8 @@ sphinx-%: $(SPHINX_DEPS)
 sphinx-pdf: sphinx-latex
 	+$(MAKE) -C $(SPHINXBUILDDIR)/latex
 
-sphinx:
-	+$(MAKE) sphinx-html
-	+$(MAKE) sphinx-pdf
+sphinx:\
+  sphinx-html sphinx-pdf
```
cpitclaudel: Are you sure about this? I don't think it's safe to run sphinx-html and sphinx-pdf in parallel.

cpitclaudel:
> @cpitclaudel confirms he's happy with the three commits I pushed on top of his branch.

I'm surprised by the commit that fixes the race condition.  Instead of making the `sphinx` target depend on both `sphinx-html` and `sphinx-pdf`, I think it would be better to make `sphinx` depend on the dependencies of `sphinx-html` and `sphinx-pdf`, or to make it depend on one and run the other in its body.

Zimmi48:
OK that sounds good to me as well.

Zimmi48:
In Travis, there are still some issues:

```
Output written on CoqRefMan.pdf (511 pages).
Transcript written on CoqRefMan.log.
Latexmk: Index file 'CoqRefMan.idx' was written
Latexmk: Missing input file: 'CoqRefMan.ind' from line
  'No file CoqRefMan.ind.'
Latexmk: Log file says output to 'CoqRefMan.pdf'
Latexmk: All targets (CoqRefMan.pdf) are up-to-date
Collected error summary (may duplicate other messages):
  makeindex CoqRefMan.idx: Command for 'makeindex CoqRefMan.idx' gave return code 256
Latexmk: Use the -f option to force complete processing,
 unless error was exceeding maximum runs of latex/pdflatex.
make[4]: *** [CoqRefMan.pdf] Error 12
make[4]: Leaving directory `/home/travis/build/Zimmi48/coq/doc/sphinx/_build/latex'
make[3]: *** [sphinx-pdf] Error 2
```

In GitLab CI we get a similar warning but no error:

```
Latexmk: Summary of warnings:
  Latex failed to resolve 826 reference(s)
Rule 'makeindex CoqRefMan.idx': File changes, etc:
   Non-existent destination files:
      'CoqRefMan.ind'
```

Could it be that the older version available on Travis fails when the newer on one on GitLab CI doesn't?
Is is OK if we disable the LaTeX builds on Travis given that they are already tested on GitLab CI?

@ejgallego In the end, I finally picked versions for all the python dependencies (see message of commit 4a48cfb) and I install them all with pip (but it's fully reproducible since they are pinned). Is that OK with you? In particular, this is targeting Coq 8.10 and all of these are available as part of Debian Buster so it seems consistent with the move to OCaml 4.05.

ejgallego:
> @ejgallego In the end, I finally picked versions for all the python dependencies (see message of commit [4a48cfb](https://github.com/coq/coq/commit/4a48cfbe34f3c8087c913d7a1090fa67eea15ae2)) and I install them all with pip (but it's fully reproducible since they are pinned). Is that OK with you?

I am not why you didn't follow my original advice, that was to install with PIP Sphinx-only dependencies. In the previous version you moved to apt also some sphinx packages, and now we have back unrelated python packages into pip. IMHO beautifulsoup and pexpect are core python packages and should be installed with python itself. [antlr4 should be too but it is still not in apt]

ejgallego (changes requested):
`pexpect` and `beautifulsoup` are system packages and should be installed as such.

Zimmi48:
@ejgallego My rationale is not important enough for me to argue so I've implemented the change that you just requested.

ejgallego:
> @ejgallego My rationale is not important enough for me to argue so I've implemented the change that you just requested.

What's your rationale? I am lost I guess... I was under the impression that it was more of a typo [sorry it is hard to track comments in this PR due to the size]

I mean, setuptools for example is in the same category than `pexpect`, so I dunno, I'd try to minimize pip deps as in the end I think it is better to integrate with package managers; but we can keep as previously if you think that's the right thing.

Now that you have done the change let's see if that goes tru, otherwise let's do as in https://github.com/coq/coq/commit/4a48cfbe34f3c8087c913d7a1090fa67eea15ae2

ejgallego:
In fact my general worry was more like "if pexpect doesn't work when in the system image, we must have some serious issue somewhere", but yeah, this is being CS-level paranoid.

Zimmi48:
@ejgallego I have no doubt CI will go through. My rationale was just to have a consistent way of picking minimum versions. But indeed, this doesn't necessarily make a lot of sense. So I'm totally fine with the current approach.

ejgallego (approved):

ejgallego:
CI image looks pretty good, CI for doc has passed, finally ready to go!

herbelin:
:+1:

Zimmi48:
@cpitclaudel I'm trying to backport this without updating the dependencies (on Sphinx in particular), and I get the following error:
```
Exception occurred:
  File "/usr/lib/python3/dist-packages/pybtex/scanner.py", line 118, in required
    raise TokenRequired(description, self)
pybtex.scanner.TokenRequired: syntax error in line 27: '}' expected
```
(cf. https://gitlab.com/coq/coq/-/jobs/99930034)
which I had also got earlier on when testing this PR with Sphinx 1.7.0.
Any idea how to workaround this bug?

Zimmi48:
@cpitclaudel It seems that it is due to your "unescaping" of `%` sign in bibtex: https://bitbucket.org/pybtex-devs/pybtex/issues/103/pybtexscannertokenrequired-syntax-error-in.

Zimmi48:
Since we need to workaround this bug of this older sphinx-bibtex in v8.8 and v8.9, I simply removed this link from the backported version.

cpitclaudel:
:+1: Thanks a lot @Zimmi48 for all your help, and everyone for the reviews

cpitclaudel:
> Any idea how to workaround this bug?

Could this be due to pybtex getting confused? I think your workaround is perfect.

PR #8420, opened 2018-09-06
[pfedit] Fix master build due to merge conflict

ejgallego:


ejgallego:
This needs review cc @SkySkimmer

proofs/pfedit.ml
```diff
@@ -173,7 +173,7 @@ let build_by_tactic ?(side_eff=true) env sigma ?(poly=false) typ tac =
       const_entry_body = Future.chain ce.const_entry_body
         (fun (pt, _) -> pt, ()) } in
   let (cb, ctx), () = Future.force ce.const_entry_body in
-  let univs = UState.merge side_eff Evd.univ_rigid univs ctx in
+  let univs = UState.merge ~sideff:side_eff ~extend:status Evd.univ_rigid univs ctx in
```
SkySkimmer: I don't think that's the right value for `extend`, it should probably be `true`.
Isn't the `status` about using `admit`?

SkySkimmer (changes requested):

proofs/pfedit.ml
ejgallego: I have no idea, setting to true for now.

SkySkimmer (approved):

SkySkimmer:
I'm going to merge since otherwise everything is broken + this is what was intended in #8302.

PR #8422, opened 2018-09-06
Modules and definitions now share the same namespace.

ppedrot:
This prevents naming a definition and a module the same. This will turn out to be useful when we implement side-effects as local definitions, so that the nametab can reuse the same syntax for module and definition qualification.

Furthermore the current situation is a bit clunky when it comes to named universes, as both a definition and a module can generate qualified universe names. This can result in strange situations.

The only places that relied on the old behaviour were tests, either to precisely check this property, or that use standard naming schemes that clashed with one another (foo, bar, etc.).

If this doesn't break anything in the CI, this should ideally go into 8.9 so that we can implement internal definitions for 8.10.


test-suite/output/UnivBinders.v
```diff
@@ -43,18 +43,18 @@ Print mono.
 Check mono.
 Check Type@{mono.u}.
 
-Module mono.
-  Fail Monomorphic Universe u.
+Module Mono.
+(*   Fail Monomorphic Universe u. *)
```
SkySkimmer: Just remove the line no?

SkySkimmer (approved):

test-suite/output/UnivBinders.v
maximedenes: Then put a comment.

ppedrot: Added.

ppedrot:
It seems that HoTT, elpi and lambda-rust rely on this feature. What should we do? Just print a warning for 8.9?

Zimmi48:
Make this a deprecated warning so that you can forbid it in 8.10 (i.e. after branching) and implement internal definitions right away (or am I missing some additional step)?

ppedrot:
The error has been turned into a warning. If this gets into 8.9 we'll be able to remove the feature in master as explained above.

gares:
In some tests in elpi I do name modules as inductive types.
Eg `inductive nat := .... elpi derive ....   Check nat.induction`.
Is this problematic too?

gares:
Actually I had a use case for constants too (pseudo syntax).
`Module foo. Definition body := ... End foo. Lemma foo := foo.body.`
The objective being to systematically provide both the transparent and opaque variant of `foo` as `foo` and `foo.body`.

ppedrot:
@gares Your use case is precisely the feature I am trying to provide through internal definitions. The fact that you use modules for that is the only way to do this currently, but as soon as we have internal definitions you should be able to produce that without such an encoding.

SkySkimmer:
Damn buttons

SkySkimmer:
>@gares Your use case is precisely the feature I am trying to provide through internal definitions. The fact that you use modules for that is the only way to do this currently, but as soon as we have internal definitions you should be able to produce that without such an encoding.

But what to do in the interval?

gares:
> but as soon as we have internal definitions you should be able to produce that without such an encoding.

Could you explain the feature you have in mind? Your plan?

I'm really using modules as name spaces, nothing more. Another way to put it, I'd like constants/inductive names to possibly contain `.` (that happens to be the module separator, but still).
But, I also do believe that it may be useful to have a notion of "all the things with prefix nat.", that would be the module name today.

So please elaborate your plan.

gares:
wrt this PR, You removed possible collisions between definitions and modules.
My concern is that I don't see why you should let inductive and module names to collide.

(hence a report of my current use case, and my full support to provide any other mean for doing so).

ejgallego:
> Could you explain the feature you have in mind? Your plan?

I second that, it seems it would be good if you discussed a bit how are you planning to replace side effects.

In particular, I may be a bit worried in that I'd dare to say that many other cleanups should happen first, like removing `Global.env ()` etc....

ppedrot:
> My concern is that I don't see why you should let inductive and module names to collide.

They don't. When I wrote definitions, I meant any non-module object that had a label in the kernel. This includes definitions.

> Could you explain the feature you have in mind? Your plan?

The ultimate goal is to get rid of side-effects, and I need two things to do that:
1. having a way to qualify kernel objects within another definition
2. designing private definitions

Contrarily to the module system, this system would be outside of the kernel. Namespacing is only visible through the Nametab module, no need to implement a complex trusted system to provide this feature.

Feature 1. would allow the user to give a user-facing name of a definition that is qualified by another definition. Typical use case:

```
Inductive foo := ...
Definition foo_rect := ...
Name foo_rect as foo.rect.
```
which would create an alias eliminated at name resolution time. Obviously you could also do that directly without going through an intermediate definition, by relying on some unique kernel name generation that nobody cares about because it would'nt be visible to the user.

Feature 2 is straightforward: it provides a way to hide the name of a constant outside of some implementation scope, so that in practice you cannot write a program that manipulates this name. Obviously we have dependent types so that such definitions can still be unfolded in tactic mode and whatnot, but this would cover 95% of the use cases of (non-functor) modules without the added complexity that comes with them.

With that in hand, you can easily replace side-effects with internal definitions whose names are private both in defined and qed cases. And you remove the trusting code out of the kernel, as this morally undoes what the side-effect mechanism introduction did.

ppedrot:
> In particular, I may be a bit worried in that I'd dare to say that many other cleanups should happen first, like removing Global.env () etc....

This is orthogonal to this issue.

ejgallego:
So in your design @ppedrot , what does `Qed` return?

mattam82:
Could you elaborate on how ```abstract``` would change in your proposed design?

ejgallego:
> This is orthogonal to this issue.

I am quite skeptical private definitions created by tactics can work reliably [hint today they don't] unless we don't clean up the env mess first.

mattam82:
Maybe not much changes and you just continue to carry an extension of the env in the evar_map (the current ```effects``` field)? At Qed you declare those constants as internal ones.

ppedrot:
I mean, in terms of kernel handling. That is, Qed would still be delegated but no side-effect inlining would take place, instead the future proof would declare the local subproofs as inaccessible names.

ejgallego:
> I mean, in terms of kernel handling. That is, Qed would still be delegated but no side-effect inlining would take place, instead the future proof would declare the local subproofs as inaccessible names.

Ah OK, I wonder what that does mean; in principle you don't need to declare the names innacessible, just not to declare them, right?

gares:
Which data structure do you plan to change and what does "backpatch" exactly mean.

ejgallego:
> making the constant body mutable

Please not; there is no reason the update function cannot be functional.

gares:
Nah, no mutability. After re-reading the thing 10 times I guess @ppedrot wants just to put in the future not a simple term but a more complex data structure, that is a list of constants plus the main term (a local env). It has always been an option, but you need to fix the system in many other places (discharge, print assumptions, module instantiation...).

ejgallego:
Yeah that's what I wondering, what does the local `env` exactly is supposed to contain. That's tricky.

Unfortunately due to the way OCaml works I am convinced that proof checking has to be managed with a queue, in the same way typical single-threaded environments [node, etc...] work today.

So the core of the system shall be functional and the proof queue shall back patch indeed the env as new proofs do arrive.

ppedrot:
Anyways, could we still consider this PR for inclusion. Even without the long-term plan, it makes sense insofar as that considering modules and definitions to be in different namespaces is a lie. In OCaml, this is enforce by syntactical ways (namely upper vs. lowercase), but in Coq, there are places where syntax alone does not disambiguate what we are talking about, e.g. `Print`. The ambiguous design choice has forced us to provide an alternative syntax for projections...

gares:
IMHO Print is the corner case, since it is overloaded. In the language of terms, for example, there is no confusion (modules alone don't exist). Do you have a more convincing example than Print?

ppedrot:
Well, now there is the universe naming scheme. `x.i` refers both to the definition and the module `x`.

gares:
Could you elaborate, I'm not familiar with that.

ppedrot:
Global monomorphic universes now get a user-readable name that is related to the point where it was defined. So for instance, if you write `Definition t@{u} := Type@{u}.` this defines a new global universe, and its user-facing name is `t.u`. But modules also have the same feature, if you write `Universe u`, the universe will be named `M.u` where `M` is the current module.

gares:
I must have missed this change or this feature.
I'm frankly quite puzzled about the syntax, `u` is not a sub component of `t`, eg I can write `Type@{t.u}` and this has nothing to do with `t` (I even imagine one can type check this without looking a `t` (both its body and its type).

SkySkimmer:
Then what name would you use? If you just call it `u` everything breaks as it stops the relatively common idiom
~~~coq
Definition foo@{i j} := ...
Definition bar@{i j} := ...
(* universes on foo and bar should be unrelated *)
~~~
And if I give universes names I want them to be retained so that printing is readable.

gares:
This is because the names `i` and `j` are made accessible. They could be used for printing only, and be displayed as `Type{foo#i}`.

I've just tried and they work in input too, eg `Check Type@{foo.i}`... Do we need this feature?

gares:
I mean, if you need to use the same universe multiple times you can name it with `Universe`.

SkySkimmer:
>This prevents naming a definition and a module the same. This will turn out to be useful when we implement side-effects as local definitions, so that the nametab can reuse the same syntax for module and definition qualification.

We could also keep it the way it is now, implement local definitions and accept that if names are reused it will sometimes error.

ppedrot:
So what? Should I close this PR?

ppedrot:
Let's close this for the time being.

PR #8423, opened 2018-09-06
coqpp: allow DEPRECATED when declaring tactics

vbgl:
#7907 introduced the syntax in the legacy `grammar/tacextend.mlp` file. Now it is also recognized by the new coqpp.

### DIFF
--- a/coqpp/coqpp_parse.mly
+++ b/coqpp/coqpp_parse.mly
@@ -62,7 +62,7 @@ let parse_user_entry s sep =
 %token <string> IDENT QUALID
 %token <string> STRING
 %token <int> INT
-%token VERNAC TACTIC GRAMMAR EXTEND END DECLARE PLUGIN
+%token VERNAC TACTIC GRAMMAR EXTEND END DECLARE PLUGIN DEPRECATED
 %token LBRACKET RBRACKET PIPE ARROW COMMA EQUAL
 %token LPAREN RPAREN COLON SEMICOLON
 %token GLOBAL FIRST LAST BEFORE AFTER LEVEL LEFTA RIGHTA NONA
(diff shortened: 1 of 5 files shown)


maximedenes:
Thanks! While you are at it, maybe adding a one-line comment at the top of each of the two files duplicating the syntax would be good, to prevent other occurrences of the mistake I did.

coqpp/coqpp_parse.mly
```diff
@@ -108,8 +108,13 @@ vernac_extend:
...
 tactic_extend:
-| TACTIC EXTEND IDENT tactic_level tactic_rules END
-  { TacticExt { tacext_name = $3; tacext_level = $4; tacext_rules = $5 } }
+| TACTIC EXTEND IDENT tactic_deprecated tactic_level tactic_rules END
+  { TacticExt { tacext_name = $3; tacext_deprecated = $4; tacext_level = $5; tacext_rules = $6 } }
+;
+
+tactic_deprecated:
+| { None }
+| DEPRECATED IDENT { Some $2 }
```
ppedrot: This should probably be a string rather than an identifier.

maximedenes: Why? You mean, it may need to be qualified?

vbgl: The use-case I have is as follows (excerpts):

```
let romega_depr = Vernacinterp.mk_deprecation ()

(* … *)

TACTIC EXTEND romega
DEPRECATED romega_depr
|  …
```

Using a string would require to put double quotes around the name of the constant `romega_depr`, right?

ppedrot: Ah, I thought this was the message payload. So rather than STRING, you should then use the CODE token so that it can be an arbitrary piece of OCaml (wrapped between braces). The old macro system is quite limited, let's take advantage of the switch to be as expressive as possible.

vbgl: If I go for CODE, how do I print it?

ppedrot: Just print the code field of the record.

vbgl: Indeed… Done. Thanks.

maximedenes:
@ppedrot can you be the assignee? Thanks!

ppedrot (approved):

PR #8419, opened 2018-09-06
Remove romega in favor of lia

vbgl:
This is a subset of #7878. This should lead to fewer compatibility issues than its equivalent with omega because `romega` relies on `zify`, just like `lia`.

One motivation is to reduce the work for coqlib (#186).

- [x] Added / updated test-suite

- [x] Corresponding documentation was added / updated (including any warning and error messages added / removed / modified).
- [x] Entry added in CHANGES.

PR #8425, opened 2018-09-06
Deprecate romega in favor of lia

vbgl:
#8419 removes the romega tactics, which is too big of a change. This PR only deprecates it.

- [x] Entry added in CHANGES.

NB: this branch is based on #8423.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -48,6 +48,8 @@ Tactics
   may need to add `Require Import Lra` to your developments. For compatibility,
   we now define `fourier` as a deprecated alias of `lra`.
 
+- The `romega` tactics have been deprecated; please use `lia` instead.
+
 Focusing
 
 - Focusing bracket `{` now supports named goal selectors,
--- a/doc/sphinx/addendum/omega.rst
+++ b/doc/sphinx/addendum/omega.rst
@@ -26,7 +26,9 @@ Description of ``omega``
 .. tacv:: romega
    :name: romega
 
-   To be documented.
+   .. deprecated:: 8.9
+
+      Use :tacn:`lia` instead.
 
 Arithmetical goals recognized by ``omega``
 ------------------------------------------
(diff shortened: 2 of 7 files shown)


herbelin (approved):
Looks making sense, very good.

Like for [fourier/lra](https://github.com/coq/coq/pull/7897#issuecomment-399696815), I wonder about what is the best name for the semantical superset of lia, omega and romega. Somehow, even if not reflecting what it does, `omega` has become a popular name among users. On the other side, my feeling is that less people are attached to the name `romega`, but one would need to ask to know. The name `lia` is very good to express what it does, but pretty cryptic if one does not know in advance what the acronym denotes. I wonder whether we should think at something more immediately clear, especially for newcomers, like `arith`, or `larith`. At the same time, `omega` has become popular without denoting anything particular at first, so one may get used to `lia`. Moreover, in contrast with `lra` which (at least me) I easily mispells, `lia` is pronouncable and easier to remember for me. Thus, I probably would lean by default towards keeping `lia` as well as `omega` aliasing to it, but I'd be curious to hear user feedback on this question, before really making up a personal opinion.

Side question: why twice `nia` (whose name is cryptic as well for a newcomer) in `Remainder_equiv`?

vbgl:
Thanks for the comments. Indeed, `nia` was not needed there.

Zimmi48 (approved):
Thanks. A good first step indeed for 8.9.

maximedenes:
@vbgl can you rebase now that #8423 is merged?

Zimmi48 (changes requested):
Sorry I forgot about this earlier: this PR should also modify the doc (which mentions `romega` more than it documents it) to add a `.. deprecated::` directive (search for other uses of this directive in the manual for examples). The PR removing romega should also be updated to remove all mentions from the manual.

Zimmi48 (approved):
Perfect, thanks!

ppedrot (approved):

Zimmi48:
Everyone approved. Who merges? @ppedrot?

PR #8434, opened 2018-09-07
Canonical representation of kernel substitutions

ppedrot:
For some reason the code was implementing substitutions as pairs of maps, but the invariant ensured actually no observable difference between fetching a module ident from one or the other. The split seems to be mostly due to historical reasons. We make this invariant static by representing substitutions as a single map.


maximedenes (approved):

silene:
Is there a reason to have this in 8.9?

ppedrot:
Nope, you can remove the milestone.

PR #8436, opened 2018-09-07
Move maps & sets indexed by GlobRef.t into the kernel

vbgl:
This moves the modules `RefOrdered`, `RefOrdered_env`, `Refmap`, `Refmap_env`, `Refset`, and `Refset_env` from `Globnames` to `Names`.

### DIFF
--- a/kernel/names.ml
+++ b/kernel/names.ml
@@ -935,7 +935,7 @@ end
 
 type projection = Projection.t
 
-module GlobRef = struct
+module GlobRefInternal = struct
 
   type t =
     | VarRef of variable           (** A reference to the section-context. *)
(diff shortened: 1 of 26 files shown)


kernel/names.ml
```diff
@@ -951,11 +951,71 @@ module GlobRef = struct
...
+  type t = GlobRef.t
+  let compare gr1 gr2 =
+    GlobRef.global_ord_gen compare ind_user_ord constructor_user_ord gr1 gr2
+  let equal gr1 gr2 =
+    GlobRef.global_eq_gen equal eq_user_ind eq_user_constructor gr1 gr2
+  let hash gr = GlobRef.global_hash_gen hash ind_user_hash constructor_user_hash gr
+end
+
+module Refmap = HMap.Make(RefOrdered)
+module Refset = Refmap.Set
```
ejgallego: I think current trend is to use `GlobRef.Map` etc...

library/globnames.ml
```diff
@@ -87,65 +87,14 @@ let printable_constr_of_global = function
...
+module RefOrdered_env = Names.RefOrdered_env
 
-(* Alternative sets and maps indexed by the user part of the kernel names *)
+module Refmap = Names.Refmap
+module Refset = Names.Refset
 
-module Refmap_env = HMap.Make(RefOrdered_env)
-module Refset_env = Refmap_env.Set
+module Refmap_env = Names.Refmap_env
+module Refset_env = Names.Refset_env
```
ejgallego: This needs a deprecation notice.

ejgallego (approved):
Modulo the module naming, this looks good to me.

kernel/names.ml
ppedrot: Note that for constants and inductives, we still use the old scheme. The reason is probably that there are two different notions of maps, one based on canonical names and the other on user names.

amahboubi (approved):
The change in the code of the ring tactic is innocuous. But it's the only part I've checked.

vbgl:
I’ve just pushed a patch with the renaming suggested by Emilio. I was a bit annoyed by all these `t` types. I don’t know if there is a way to achieve the same effects without introducing yet an other name for the type t:

```
module M = struct
  type t = A
  type alias = t
  module Inner = struct
    type t = alias
  end
end
```

SkySkimmer:
`type nonrec t = t`

ejgallego:
> I’ve just pushed a patch with the renaming suggested by Emilio. I was a bit annoyed by all these t types. I don’t know if there is a way to achieve the same effects without introducing yet an other name for the type t:

Indeed, apart from the solution of @SkySkimmer , `include` can help sometimes, and having a signature for Ordered may allow a slightly cleaner style as done for `Id.Map` in the `mli`.

ppedrot (approved):

maximedenes:
@vbgl do you want to squash the commits? Aren't they fine as they are?

maximedenes (approved):

vbgl:
Yes. I will squash them (some changes introduced in the first one are reverted in the second one).

I thought that this PR was waiting for approval from a few more people: this is why i waited for squashing.

maximedenes:
You can squash before they review, so that CI runs meanwhile. But anyway for this kind of PRs (one real change and some propagation), I think the merging documentation says you can proceed with approval only by the maintainers impacted by the change, not the propagation.

herbelin (approved):
The move seems good. Regarding the recommended naming policy, I'm trusting the advice of @ejgallego who has a good overview of the current state of the API.

Issue #8178, opened 2018-07-27
Documentation doesn't explain clearly enough how to use options and the difference between flags and options.


jfehrle:
#### Description of the problem

The command `Set Printing Compact Contexts on." is documented but not accepted in Coqtop.  Should be made to work or removed from the doc.  If the option is removed, references to it in the comments before Printer.bld_sign_env and pr_context_limit_compact and use of the option value in Printer.should_compact should also be removed.

https://coq.inria.fr/distrib/current/refman/proof-engine/vernacular-commands.html?highlight=compact#coq:opt.printing-compact-contexts

```
Unnamed_thm < Set Printing Compact Contexts on.
Toplevel input, characters 0-33:
> Set Printing Compact Contexts on.
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning: There is no option Printing Compact Contexts on.
[unknown-option,option]
```

Zimmi48:
The correct way to call the option is `Set Printing Compact Contexts.`
I don't see the "on" part in the documentation you linked.

jfehrle:
OK, fair enough, not a bug in the option.

On the other hand, I've read parts of the doc about setting flags and options more than a couple times, and they're still surprisingly confusing for what should be a simple topic.  The doc doesn't usefully define the differences between flags and options.  The syntax for setting them is identical (or at least confusingly similar).  Are flags predefined (not user-defined)?  And if so, shouldn't the doc have a flags index at the beginning of the document?  Or have a combined index of flags and options and identify which are which-if it really matters.

The general documentation for setting an option **only** shows the following form.  It does not show the form Set _option_ at all (which **is** shown for flags).

![image](https://user-images.githubusercontent.com/1253341/43348002-e1e6e15a-91c5-11e8-8200-4d06a7a0b5c2.png)

I think might be friendlier to show _Set_unset_ with a hyperlink to the full syntax (Local, Global, Export) instead of _Option_ in all the option boxes.

![image](https://user-images.githubusercontent.com/1253341/43348285-55939f98-91c7-11e8-8bec-eb90f83d586e.png)

It would also help if the parsing and error messages were a bit friendlier (less misleading), for example it sure would help the user to say that the "on" is extraneous.

```
> Set Printing Compact Contexts on.
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning: There is no option Printing Compact Contexts on.
[unknown-option,option]
```

or that something is missing:

```
[...]

PR #481, opened 2017-03-14
[option] Remove support for non-synchronous options.

ejgallego:
Inspired by #5229 , which
this PR solves, I propose to remove support for non-synchronous
options.

It seems the few uses of `optsync = false` we legacy and shouldn't
have any impact.

Moreover, non synchronous options may create particularly tricky
situations as for instance, they won't be propagated to workers.

PR #8439, opened 2018-09-07
Update documentation of options and flags

jfehrle:
**Kind:** documentation

Fixes some of the issues in #8178

Getting into 8.9 would be nice but not urgent.

1. Defines a new NotationObject type for flags, since the commands for flags and options have different syntax (Set) and semantics (Unset).  Flags are still included in the renamed Options and Flags Index.
2. Replace `:option:` with `:flag:` where needed.
3. Rewrote the general discussion of flags and options and their commands.
4. Added explicit `:name:`s so the index doesn't include flags, option and command parameters (except where needed for disambiguation).


Zimmi48 (changes requested):
Awesome job :tada: See comments below for requests for modifications.

doc/sphinx/addendum/type-classes.rst
```diff
@@ -528,6 +528,7 @@ Options
    backtrack on it, assuming that it is canonical.
 
 .. flag:: Typeclasses Debug {? Verbosity @num}
+   :name:Typeclasses Debug Verbosity
```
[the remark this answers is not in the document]
jfehrle: Yeah, neither fish nor fowl.

doc/sphinx/addendum/universe-polymorphism.rst
```diff
@@ -354,7 +354,7 @@ This minimization process is applied only to fresh universe variables.
 It simply adds an equation between the variable and its lower bound if
 it is an atomic universe (i.e. not an algebraic max() universe).
 
-.. opt:: Universe Minimization ToSet
+.. flag:: Universe Minimization ToSet
 
    Turning this option off (it is on by default) disallows minimization
```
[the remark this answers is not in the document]
jfehrle: Idiosyncratic capitalization for "ToSet".

doc/sphinx/coq-optindex.rst
```diff
@@ -1,5 +1,5 @@
 .. hack to get index in TOC
 
 -----------------
-Option index
+Option and flag index
 -----------------
```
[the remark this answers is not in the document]
jfehrle: Longer, not just equal?

doc/tools/coqrst/coqdomain.py
```diff
@@ -316,11 +316,10 @@ class OptionObject(NotationObject):
 
     Example::
 
-       .. opt:: Nonrecursive Elimination Schemes
+       .. opt:: Hyps Limit num
```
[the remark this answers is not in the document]
cpitclaudel: Yes, a test would likely be good.

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -78,145 +78,109 @@ Displaying
...
 -----------------------------
 
-|Coq| configurability is based on flags (e.g. :flag:`Printing All`), options
-(e.g. :opt:`Printing Width`), or tables (e.g. :cmd:`Add Printing Record`). The
-names of flags, options and tables are made of non-empty sequences of
-identifiers (conventionally with capital initial letter). The general commands
-handling flags, options and tables are given below.
+Coq has many settings to control its behavior.  Setting types include flags, options
+and tables.  Flags are boolean values, such as :flag:`Printing All`.  Options have numeric or
+string values, such as :opt:`Printing Width`.  Tables contain sets of strings.
```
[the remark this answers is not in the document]
jfehrle: Should I also replace @flag with :production:`flag` in the following commands?

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -78,145 +78,109 @@ Displaying
...
-   .. cmdv:: Global Unset @flag
-
-      This command switches :n:`@flag` off. The original
-      state of :n:`@flag` is *not* restored at the end of the module. Additionally,
-      if set in a file, :n:`@flag` is switched off when the file is `Require`-d.
+.. cmd::  {? Local | Global | Export } Set @flag
+   :name: Set
 
-   .. cmdv:: Export Unset @flag
+   Sets :n:`@flag` on. Scoping qualifiers are
```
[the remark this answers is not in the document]
jfehrle: Both forms are OK.  "Turns the flag on" is a little more idiomatic, but since the commands are "Set" and "Unset", I thought it would be better for readers to see the same verbs in the description.

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -78,145 +78,109 @@ Displaying
...
 
 .. cmd:: Test @flag
 
-   This command prints whether :n:`@flag` is on or off.
+   Prints the current value of :n:`@flag`.
 
 
-.. cmd:: Set @option @value
+.. cmd:: {? Local | Global | Export } Set @option @value
+   :name: Set (option)
```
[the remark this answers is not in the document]
jfehrle: I want to extend the notation to support `{| alt1 | alt }`, which is just what's needed here.  That appears to be a trivial change, but I haven't figured out how to install Antlr yet, as I noted elsewhere.  If you know how...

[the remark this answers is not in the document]
jfehrle: Like this?
```
:name: Set `option`
```

and not

```
:name: Set @option
```
?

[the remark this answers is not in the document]
cpitclaudel: What platform are you on?  On Debian derivatives you can just run `apt install antlr`.  Alternatively, there are installation instructions for OSX, Windows, and GNU/Linux systems on http://www.antlr.org/, in the quick start box.  You can cycle through using the small dots at the bottom right of the quick start tab.

jfehrle:
BTW, is there a convention for using this button--is it for the submitter or the reviewer?

image

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -78,145 +78,109 @@ Displaying
...
+.. cmd:: Print Options
 
-   This command prints the current value of :n:`@option`.
+   Prints the current value of all flags and options, and the names of all tables.
 
 
 .. TODO : table is not a syntax entry
 
 .. cmd:: Add @table @value
    :name: Add `table` `value`
```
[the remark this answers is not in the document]
jfehrle: This has some problems.  I have

```
.. cmd:: Add @table @string
   :name: Add @table @string

   Adds :token:`string` to :token:`table`.
```

which renders as 

image

Seems strange to have `table` in the description show up in black.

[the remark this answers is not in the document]
jfehrle: Also,

```
.. cmd:: Remove @table @string
   :name: Remove @table @string
```

gives this index entry:

image

Do you think that's preferable to this style for the index?

image

doc/sphinx/proof-engine/vernacular-commands.rst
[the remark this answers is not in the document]
jfehrle: The `:name:` has to be indented like this:

```
.. cmd:: {? Local | Global | Export } Set @option @num
         {? Local | Global | Export } Set @option @string
   :name: Set @option @num; Set @option @string
```

jfehrle:
I fixed what I could and pushed the changes.  Let me know what you think.

doc/sphinx/proof-engine/vernacular-commands.rst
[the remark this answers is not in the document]
jfehrle: @cpitclaudel I'm on Windows.  I clicked on "Download" on www.antlr.org I didn't read most of the home page.  On the download page, I missed the sentence `All users should download the ANTLR tool itself and then choose a runtime target below, unless you are using Java which is built into the tool jar.', which IMHO should have been the first sentence on the page.

At the moment, when I try the link in that sentence or in the quick start box I get `This site can’t be reached
antlr.org refused to connect.`  Perhaps that's just transitory.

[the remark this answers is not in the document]
jfehrle: Ah, the site is back up.

jfehrle:
I added support for alternatives in notation using {| ... }

For example: 

```
.. cmd:: {? Local | Global | Export } Set @option {| @num | @string }
   :name: Set @option
```

renders as:

image

Please take a look.  I plan to squash the last 4 or 5 commits before you merge.  Too complicated to keep changes separate when they affect the same files.

cpitclaudel:
Hmm, should it actually be `{? Local | Global | Export }`, not `{|? Local | Global | Export }` or `{? {| Local | Global | Export }}`?
This local or global or export thing is an alternative too, albeit optional, isn't it?

Also, did you update the LaTeX rendering code?

jfehrle:
> Hmm, should it actually be ...

I don't get your question.  The expression I have allows for Local, Global, Export or empty.  What would you do differently, and why?  AFAIK formats are only single characters, so I think `{|?` would not fit the current code very well.

> Also, did you update the LaTeX rendering code?

There's more to do?  Where is that code?  I've never used LaTeX.

cpitclaudel:
> I don't get your question. 

I don't understand the semantics of `{|`, that's all :)
If `{|` indicates an alternation, then I would expect to have to write `{? {| Local | Global | Export }}`.

> There's more to do? Where is that code?

I think you updated sphinx.py, but you also want to update `html.py`, `latex.py`, `plain.py`, and `regexp.py`

doc/tools/coqrst/coqdomain.py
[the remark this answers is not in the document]
Zimmi48: I'll open an issue so that we don't forget.

Zimmi48:
Note: I'm answering some of your inline comments here so that you see the answer.

>Idiosyncratic capitalization for "ToSet".

I believe this is due to a limitation of the parser which prevents having `Set` as part of the name of an option (although we managed for `Inductive` so I'm not sure that this is actually impossible).

>Longer, not just equal?

Longer or equal.

>Should I also replace `@flag` with :production:`flag` in the following commands?

No you can use ```:production:`flag` ``` only once (at the definition point).

>BTW, is there a convention for using this button--is it for the submitter or the reviewer?

It's a new thing. I'd say it's mostly useful when some inline comment was taken into account but is still shown, so I'd say it's for the submitter. The reviewer can always "unresolve" if they disagree.

@cpitclaudel Can we have support for `@` in command names / index?

>I notice that if I have both of the following, the index entry points to the first item and there is no error message for the second item Should there be an error message? On the one hand, this frequently may be an error but it may also be that the command appears multiple times to express variations. Wouldn't an error message would force every variant to have its own index entry?

Are you sure that the second one doesn't appear after the first but with just the differing part shown? I agree it's not very visible.

doc/sphinx/addendum/extended-pattern-matching.rst
```diff
@@ -262,9 +262,9 @@ When we use parameters in patterns there is an error message:
         | cons A _ l' => l'
         end).
 
-.. opt:: Asymmetric Patterns
+.. flag:: Asymmetric Patterns
 
-This option (off by default) removes parameters from constructors in patterns:
+This flag (off by default) removes parameters from constructors in patterns:
```
Zimmi48: This body needs to be indented.

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -79,82 +79,80 @@ Flags, Options and Tables
...
+
+.. FIXME Convert `Extraction Language OCaml` to an option.
 
 Flags, options and tables are identified by a series of identifiers, each with an initial
 capital letter.
 
-.. TODO : flag is not a syntax entry
-
 .. cmd::  {? Local | Global | Export } Set @flag
    :name: Set
```
Zimmi48: I'm not sure I understand what you said: did you try putting `:name: Set @flag` here?

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -79,82 +79,80 @@ Flags, Options and Tables
...
-   :name: Remove `table` `value`
+.. cmd:: Remove @table @string
+   :name: Remove @table @string
 
-   Removes :n:`@value` from :n:`@table`.
+   Removes :token:`string` from :token:`table`.
 
-.. cmd:: Test @table @value
-   :name: Test `table` `value`
+.. comment THIS DIDN'T WORK, e.g. for "Test Search Blacklist "foo"."
```
Zimmi48: Add a FIXME for making it easier to find this comment by grepping.

Zimmi48:
>I'm not sure. In the index, yes, but not in HTML anchors

Yeah, I don't really care if this is omitted from the anchors as long as the indexes render them well.

doc/sphinx/proof-engine/proof-handling.rst
```diff
@@ -449,6 +450,7 @@ The following example script illustrates all these features:
 Set Bullet Behavior
 ```````````````````
 .. opt:: Bullet Behavior  %( "None" %| "Strict Subproofs" %)
```
Zimmi48: This option could use your new `{| }` box.

Zimmi48 (approved):
Great job! I'll wait for @cpitclaudel's green light on the changes to the Sphinx infrastructure before merging. I'll give also a few days to give an opportunity to other developers to comment. Note that as long as this stays a doc-only PR, I think this can be included in 8.8.2.

jfehrle:
@cpitclaudel OK, let me know if this clarifies things:
* `{| ... }` puts its contents in a box.  It generates the same html as `{? ... }` but without the superscript.
* The `|` characters that appear inside the `{| ... }` are ordinary characters--no special handling. 
 They are conventionally used for alternatives.  

In my earlier example, look at the `{| @num | @string }` part of the line, which is rendered as:

image

The first part of the example has `{? Local | Global | Export } `, which already generates a box around the `Local | Global | Export`.  I haven't tried the 2 alternate forms you mentioned.  I expect they would add unwanted additional boxes.

The changes I made are tiny: I added one `if` in sphinx.py and a single character to TacticNotations.g.

> you also want to update `html.py`, `latex.py`, `plain.py`, and `regexp.py`

The changes needed there are probably quite small, but I have no experience with Latex and don't know how to test the changes.  And I have some Latex configuration problem.  I'm not eager to dig into this since I don't plan to use Latex, so perhaps I'll withdraw this change.  If you have time to do the Latex part, I'd be delighted.

BTW, I don't see a `latex.py`.

```
$ time env nice make -j 4 doc-pdf
rm -f
make: Circular .merlin.in <- submake dependency dropped.
make: Circular ide/.merlin.in <- submake dependency dropped.
make: Circular plugins/.merlin.in <- submake dependency dropped.
make: Circular META.coq.in <- submake dependency dropped.
make --warn-undefined-variable --no-builtin-rules -f Makefile.build doc-pdf
make[1]: Entering directory '/cygdrive/c/Users/Jim.DESKTOP-MMPNVGH/Documents/coqSource/coq'
(cd doc/stdlib;\
latex -interaction=batchmode Library;\
latex -interaction=batchmode Library > /dev/null;\
../tools/show_latex_messages -no-overfull Library.log)
This is pdfTeX, Version 3.14159265-2.6-1.40.19 (TeX Live 2018/Cygwin) (preloaded format=latex)
 restricted \write18 enabled.
entering extended mode
! LaTeX Error: File `ucs.sty' not found. (file Library.tex)

Type X to quit or <RETURN> to proceed,
or enter new name. (Default extension: sty)

Enter file name:
! Emergency stop. (file Library.tex)
<read *>

l.4 \usepackage
make[1]: *** [Makefile.doc:133: doc/stdlib/Library.dvi] Error 1
make[1]: Leaving directory '/cygdrive/c/Users/Jim.DESKTOP-MMPNVGH/Documents/coqSource/coq'
make: *** [Makefile:166: submake] Error 2

real    0m7.137s
user    0m1.849s
sys     0m4.852s
```

jfehrle:
> for "options and flags" and "errors, warnings". ... I think that we should seek consistency between these two index names: either use a comma or a "and".

"and" is standard English usage.  Commas would be understood but substandard.

Zimmi48:
Let's put "and" in both index titles then.

Zimmi48:
OK I see what you mean: `.. cmd:: Test @flag` has no explicit `:name:` so the computed name is `Test` which is shown in the index. Then `.. cmd:: Test @option` has no explicit `:name:` so the computed name is again `Test` and it isn't shown in the index because it is already there (maybe it should indeed emit a warning in this case). Then `.. cmd:: Test @table for @string` has an explicit name `Test @table for @string` so this appears in the index.

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -78,145 +78,107 @@ Displaying
...
+and tables:
 
-.. TODO : flag is not a syntax entry
+* A :production:`flag` has a boolean value, such as :flag:`Asymmetric Patterns`.
+* An :production:`option` has a numeric or string value, such as :opt:`Firstorder Depth`.
+* A :production:`table` contains a set of strings.
+* In addition, some commands provide settings, such as :cmd:`Extraction Language OCaml`.
 
-.. cmd:: Set @flag
+.. FIXME Convert `Extraction Language OCaml` to an option.
```
herbelin: One possibility is that support for declaring new types of options is added to `library/goptions.ml`.

The FIXME should not stay in the reference manual in any case.

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -78,145 +78,107 @@ Displaying
...
-.. cmd:: Test @option
+.. cmd:: Remove @table @string
+   :name: Remove @table @string
 
-   This command prints the current value of :n:`@option`.
+   Removes :token:`string` from :token:`table`.
 
+.. comment THIS DIDN'T WORK, e.g. for "Test Search Blacklist "foo"."
+   cmd:: Test @table @string
+   :name: Test @table @string
```
herbelin: This syntax does not exist. What exists is `Test @table` which is synonymous for `Print Table @table` (and of course also `Test @table for @string` and `Test @table for @qualid`).

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -78,145 +78,107 @@ Displaying
...
 
-.. TODO : table is not a syntax entry
+   Reports whether :token:`table` contains :token:`string`.
 
-.. cmd:: Add @table @value
-   :name: Add `table` `value`
+.. cmd:: Test @table for @string
+   :name: Test @table for @string
```
herbelin: Two kinds of table are currently supported: for string (e.g. "Search Blacklist") and for global reference (e.g. "Printing Coercion"). So, `Test @table for @qualid` should be added.

herbelin (changes requested):
A very useful PR.

doc/sphinx/README.rst
```diff
@@ -114,15 +117,25 @@ Here is the list of all objects of the Coq domain (The symbol :black_nib: indica
 
              Raised if :n:`@tactic` does not fully solve the goal.
 
-``.. opt::`` :black_nib: A Coq option.
+``.. flag::`` :black_nib: A Coq flag (aka a binary-valued option).
```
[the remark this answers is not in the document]
jfehrle: aka = also known as
I can make this "i.e." if that's clearer.

doc/sphinx/proof-engine/vernacular-commands.rst
jfehrle: `FIXME` doesn't appear in the document; above @Zimmi48 is asking me to add `FIXME` comments.

Are you OK with that?

doc/sphinx/proof-engine/vernacular-commands.rst
jfehrle: Will do.  Presumably `Add table` and `Remove table` also take qualids.

I wish there was a way to extract this information systematically rather than having to read code and/or rely on reviewers.

doc/sphinx/proof-engine/vernacular-commands.rst
jfehrle: I notice that `Add Printing Coercion` is documented as taking a `qualid`, while the other `Add Printing XXX` command say they take `ident`s.  Is that correct or are they all qualids?

jfehrle:
I will back out the addition and use of the `{| ... }` construct because I can't run latex locally to verify that it works correctly for that case.  I don't have enough free disk space to install Linux for Windows, which I apparently would need to be able to generate doc with latex.  I hope someone else can do that (we need such a construct).

@ppedrot Do you approve the changes in the first commit "Define flags ..."?

jfehrle:
Pushed the changes addressing the comments and a few other details.  @herbelin please take another look, thanks.

doc/sphinx/README.rst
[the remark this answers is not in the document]
herbelin: I'm not native English speaker but I thought that aka was a bit informal jargon. I thought that the standard usage was to use "i.e.' or "also known as" in full. But again, I'm not familiar with the effective usages.

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -78,145 +78,107 @@ Displaying
...
-(e.g. :opt:`Printing Width`), or tables (e.g. :cmd:`Add Printing Record`). The
-names of flags, options and tables are made of non-empty sequences of
-identifiers (conventionally with capital initial letter). The general commands
-handling flags, options and tables are given below.
+Coq has many settings to control its behavior.  Setting types include flags, options
+and tables:
 
-.. TODO : flag is not a syntax entry
+* A :production:`flag` has a boolean value, such as :flag:`Asymmetric Patterns`.
+* An :production:`option` has a numeric or string value, such as :opt:`Firstorder Depth`.
```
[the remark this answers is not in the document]
herbelin: OK

doc/sphinx/proof-engine/vernacular-commands.rst
herbelin: Ah sorry, I did not realize that `FIXME` was a sphinx keyword!

doc/sphinx/proof-engine/vernacular-commands.rst
herbelin: Ah, well spotted. They indeed all take a qualid.

doc/sphinx/proof-engine/vernacular-commands.rst
herbelin: > Presumably Add table and Remove table also take qualids.

Yes, `Remove ...` also takes a qualid for those table which takes a qualid (like the `Printing ...` tables).

>  wish there was a way to extract this information systematically rather than having to read code and/or rely on reviewers.

You mean that the information is given at the time of the parsing rule? Maybe something is doable in this direction.

doc/sphinx/proof-engine/vernacular-commands.rst
Zimmi48: Not a keyword but Sphinx comments because of the `..` and the fact it is not a known directive.

doc/sphinx/README.rst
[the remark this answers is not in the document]
jfehrle: Yes.  Webster's dictionary includes "aka" as an abbreviation and doesn't say it's informal.  https://www.merriam-webster.com/dictionary/aka.  On the other hand, it's usually used to give aliases for people, so it's a bit playful to use it here.

doc/sphinx/proof-engine/vernacular-commands.rst
jfehrle: FWIW, I see there is an `Unset Printing Records` command, but none of the other tables support this.

jfehrle:
How does `synchronously` relate to options, flags and tables?  I don't have a good guess what this might be--is it still a real feature?

```
.. cmd:: Print Table Printing If
    :
The list of inductive types for which pattern-matching is written
using an ``if`` expression is managed synchronously. This means that it is
sensitive to the command ``Reset``.
```

```
Coq < Print Options.
Synchronous options:
  Asymmetric Patterns: off
  Automatic Coercions Import: off [DEPRECATED]
  Automatic Introduction: on [DEPRECATED]
```

SkySkimmer:
About synchronous see #481

jfehrle:
Thanks.  I removed "synchronously" text above for 2 `Print Table` commands.

@Zimmi48 Does removing the useless/confusing "Synchronous options:" below have to be a separate PR?

```
Coq < Print Options.
Synchronous options:
  Asymmetric Patterns: off
  Automatic Coercions Import: off [DEPRECATED]
```

jfehrle:
Also, I notice the doc uses `pattern-matching` in many places.  I think `pattern matching` is standard usage.  Shall I go ahead and change those as part of this PR?

jfehrle:
Added "table" as a new type and updated the doc to use it.  Look at the commit "Revise doc for tables".

I need a little help with the wording for `Printing Let`, in which "written" is confusingly overused:

```
Specifies a set of qualids for which pattern matching is written using a let expression. Note that
this only applies to pattern matching written with match. It doesn't affect pattern-matching
explicitly written with a destructuring let.
```

Zimmi48:
>Does removing the useless/confusing "Synchronous options:" below have to be a separate PR?

These options have been removed and were left-over in the doc? Sure, feel free to do the clean-up as part of this PR.

SkySkimmer:
>These options have been removed and were left-over in the doc?

No, it's that they're now the only kind of option so there's no need for the header.

SkySkimmer:
>I need a little help with the wording for Printing Let, in which "written" is confusingly overused:

The first one could be "displayed" and the others `input (as)`.

jfehrle:
Found and fixed more options that are really flags.  I notice that the Extraction and Omega flags are not accepted by the command line.  I suppose that's because they're in plugins and the flags are only available if you load the plugin?  Perhaps I should add some text saying that?

```
Auto Template Polymorphism
Extraction Optimize
Extraction Conservative Types
Extraction KeepSingleton
Extraction AutoInline
Extraction SafeImplicits
Printing Allow Match Default Clause  (and added missing "Match")
Stable Omega
Omega UseLocalDefs
Omega System
Omega Action
SsrHave NoTCResolution

**Comment by jfehrle** [comment]
Pushed updates.

> Also, I notice the doc uses "pattern-matching" in many places.

I'll make that a separate PR since there are messages in the code using the hyphenated form.

> Does removing the useless/confusing "Synchronous options:" below have to be a separate PR?

@Zimmi48 You were talking about putting this in 8.8.2 provided there are no code changes, but removing this header is a (very small) code change.  Let me know.

@cpitclaudel Do you approve the notation additions? (in the "Define flage ..." commit)

**Comment by Zimmi48** [comment]
@jfehrle OK thanks for the clarification. Yes, this kind of changes is acceptable for a patch-level release. We generally backport improvements to user messages when they do not imply large refactorings.

**Comment by Zimmi48** [comment]
I would like @herbelin to confirm that he's happy with the current state before merging.

Regarding options of omega and extractions, yes they require loading plugins but they are both documented in a chapter that already mentions this explicitly (around the top).

**Comment by jfehrle** [comment]
Changed the "Synchronous options:" header in `Print Options` to just say "Options:"

**Comment by Zimmi48** [comment]
@herbelin ping

File: doc/sphinx/proof-engine/vernacular-commands.rst (hunk 14 above)
Comment (herbelin) [reply doc/sphinx/proof-engine/vernacular-commands.rst]: Actually, `Printing Records` is not a table, so there is a bug in the example given in the manual.  It could be instead ``... or tables (e.g. :cmd:`Add Printing Let`)``.

**Review by herbelin — APPROVED** [review]

**Comment by herbelin** [comment]
> Changed the "Synchronous options:" header in Print Options to just say "Options:"

:+1:

> Also, I notice the doc uses "pattern-matching" in many places. I think "pattern matching" is standard usage. Shall I go ahead and change those as part of this PR?

I would say yes. (However, I guess that you mean hyphen in expressions such as "It doesn't affect pattern-matching explicitly written with a destructuring let", not in situations like "pattern-matching problem" about which I was - hopefully correctly - tought that an hypen is needed?)

Otherwise, looks very good to me (thanks a lot).

Note a last bug to fix though: the example `Add Printing Record` to be changed into `Add Printing Let`.

**Comment by jfehrle** [comment]
Hi Hugo,
I'm baffled by this email--I can't find it in the GitHub page for the issue, and lines shown with "+" below were deleted prior to my last push, which was 24 hours or more earlier.  I didn't find any example of Test @table [@string | @qualid] that worked.
Is it possible your comment went onto an obsolete version of the changes?
Conversely, I didn't receive an email with your comment visible in GitHub ending with "Note a last bug to fix though: the example Add Printing Record to be changed into Add Printing Let."
Jim      From: Hugo Herbelin <notifications@github.com>
 To: coq/coq <coq@noreply.github.com> 
Cc: Jim Fehrle <jfehrle@sbcglobal.net>; Mention <mention@noreply.github.com>
 Sent: Wednesday, September 19, 2018 5:38 AM
 Subject: Re: [coq/coq] Update documentation of options and flags (#8439)
   
@herbelin commented on this pull request.In doc/sphinx/proof-engine/vernacular-commands.rst:>  
+.. comment THIS DIDN'T WORK, e.g. for "Test Search Blacklist "foo"."
+   cmd:: Test @table @string
+   :name: Test @table @string
Actually, Printing Records is not a table, so there is a bug in the example given in the manual. It could be instead ... or tables (e.g. :cmd:`Add Printing Let`).—

**Comment by jfehrle** [comment]
> (However, I guess that you mean hyphen in expressions such as "It doesn't affect pattern-matching explicitly written with a destructuring let", not in situations like "pattern-matching problem" about which I was - hopefully correctly - tought that an hypen is needed?)

I found no authoritative answer.  I would remove all the hyphens for consistency.  I think it's hard to articulate (and then follow) a guideline when to use the hyphen and when not to.

If I Google "pattern-matching", all the responses use "pattern matching".  The Wikipedia article on pattern matching doesn't use a hyphen for the phrase.

Googling for ""pattern-matching problem", I do find a couple links/papers that use the hyphen in a few places (e.g. in the bibliography), but it seems to me mostly they do without.  (based on 3 minutes research).

In https://www.grammarbook.com/punctuation/hyphens.asp, rule 1 seems the closest to what you were taught, but "pattern matching" is a noun.  Also https://en.wikipedia.org/wiki/Hyphen#Compound_modifiers has a discussion.

**Comment by jfehrle** [comment]
PS I counted only 16 uses of "pattern-matching" that proceeded a noun (like your example "pattern-matching problem") out of a total of 64 uses of the phrase.

**Comment by Zimmi48** [comment]
>Is it possible your comment went onto an obsolete version of the changes?

@jfehrle It is always possible to reply to comments on outdated / resolved / masked threads, see https://github.com/coq/coq/pull/8439#discussion_r218785412.

File: doc/sphinx/language/gallina-extensions.rst [hunk 18]
```diff
@@ -100,19 +100,22 @@ to be all present if the missing ones can be inferred or prompted for
        Rat_bottom_cond := O_S 1;
        Rat_irred_cond := one_two_irred |}.
 
-This syntax can be disabled globally for printing by
+This syntax can be disabled globally for display by
 
 .. cmd:: Unset Printing Records
```
Zimmi48: Following up on Hugo's remark, it would seem that `Printing Records` (with an "s") is a flag whereas `Printing Record` (without an "s") is a table. @herbelin Is that correct? @jfehrle Can you fix this?

doc/sphinx/language/gallina-extensions.rst
```diff
@@ -100,19 +100,22 @@ to be all present if the missing ones can be inferred or prompted for
...
-to get record syntax or
+to see record syntax or
 
-.. cmd:: Add Printing Constructor @ident
+.. table:: Printing Constructor @qualid
+   :name: Printing Constructor
 
-to get constructor syntax.
+to see constructor syntax.  See the :cmd:`Add @table` and :cmd:`Remove @table` commands
+to change the set of overridden types.
```
Zimmi48: While you are at it, can you reformulate this paragraph so that each object (flag, table) gets a body (indented description).

Zimmi48 (commented):
I hope that my comment below properly clarifies the issue with `Printing Records` and `Printing Record`. While you are at it, could you scroll through the diff and indent the descriptions of options / flags / tables that are not yet indented? This would be a huge gain towards #7602.

jfehrle:
> It is always possible to reply to comments on outdated / resolved / masked threads

Sure.  However, clicking on the link in the email wasn't bringing up the message when I tried it 3 or 4 times earlier today.  And not easy to find when Github avoids showing old things by default.

doc/sphinx/language/gallina-extensions.rst
jfehrle: > it would seem that Printing Records (with an "s") is a flag whereas Printing Record (without an "s") is a table.

Confusingly similar, I'm sure others have gotten confused about this, too.

doc/sphinx/language/gallina-extensions.rst
jfehrle: Done

jfehrle:
Just waiting for agreement on "pattern-matching" now before I push again.

Zimmi48:
>Just waiting for agreement on "pattern-matching" now before I push again.

I believe I can talk for both Hugo and me by saying we trust you when it comes to deciding which form is best.

herbelin:
Usage seems indeed that "pattern matching" do not take an hyphen even when in "pattern matching problem".

Record/Records: indeed. Sorry for my confusion.

In any case, good to me.

Zimmi48:
Please push ASAP @jfehrle, I can't wait to close the 8.8.2 milestone (20 days later than I anticipated).

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -3195,10 +3197,10 @@ hints of the database named core.
...
 the :tacn:`auto` and :tacn:`trivial` tactics:
 
-.. opt:: Info Auto
-.. opt:: Debug Auto
-.. opt:: Info Trivial
-.. opt:: Debug Trivial
+.. flag:: Info Auto
+.. flag:: Debug Auto
+.. flag:: Info Trivial
+.. flag:: Debug Trivial
```
Zimmi48: When there are such a series of same objects you can group them with:

```
.. flag:: Info Auto
          Debug Auto
          Info Trivial
          Debug Trivial
   :undocumented:
```
The last `:undocumented:` bit signaling that the body is missing "on purpose".

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -3228,8 +3230,8 @@ Note that ``ex_intro`` should be declared as a hint.
 
 :tacn:`eauto` also obeys the following options:
 
-.. opt:: Info Eauto
-.. opt:: Debug Eauto
+.. flag:: Info Eauto
+.. flag:: Debug Eauto
```
Zimmi48: Same remark.

doc/sphinx/user-extensions/proof-schemes.rst
```diff
@@ -103,26 +103,26 @@ induction for objects in type `identᵢ`.
...
+.. flag:: Case Analysis Schemes
 
    This flag governs the generation of case analysis lemmas for inductive types,
    i.e. corresponding to the pattern-matching term alone and without fixpoint.
 
-.. opt:: Boolean Equality Schemes
+.. flag:: Boolean Equality Schemes
 
-.. opt:: Decidable Equality Schemes
+.. flag:: Decidable Equality Schemes
```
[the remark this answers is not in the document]
Zimmi48: Same.

Zimmi48:
In the end, you didn't change `.. cmd:: Unset Printing Records` into `.. flag:: Printing Records`?

jfehrle:
> you didn't change `.. cmd:: Unset Printing Records` into `.. flag:: Printing Records`?

Got it (and the others).  Not sure I saw it or maybe I got confused.

jfehrle:
Thanks for the reviews, gentlemen!

Thinking about revising the tactics chapter next...

PR #8443, opened 2018-09-10
[opam] [dune] Fix typo + set prefix for configure.

### LINKED ISSUE #8431: opam package `coq.dev` broken in opam 1
#### Version

master

#### Operating system
Linux

`opam install coq.dev` is broken on our CI:
```
[coq: dune build] Command started
[ERROR] The compilation of coq failed at "dune build @vodeps".
[WARNING] Directory /builds/FP/LambdaRust-coq/opamroot/system/lib/coq is not
          empty, not removing

#=== ERROR while installing coq.dev ===========================================#
# opam-version         1.2.2
# os                   linux
# command              dune build @vodeps
# path                 /builds/FP/LambdaRust-coq/opamroot/system/build/coq.dev
# compiler             system (4.02.3)
# exit-code            1
# env-file             /builds/FP/LambdaRust-coq/opamroot/system/build/coq.dev/coq-2451-736596.env
# stdout-file          /builds/FP/LambdaRust-coq/opamroot/system/build/coq.dev/coq-2451-736596.out
# stderr-file          /builds/FP/LambdaRust-coq/opamroot/system/build/coq.dev/coq-2451-736596.err
### stderr ###
# => no CoqIde will be built.
# [...]
#   Bytecode VM enabled         : true
#   Native Compiler enabled     : true
# 
#   Local build, no installation...
# 
# If anything is wrong above, please restart './configure'.
# 
# *Warning* To compile the system for a new architecture
#           don't forget to do a 'make clean' before './configure'.
```
Seems like the transition to dune was not done properly in all places?

ejgallego:
Fixes #8431 and produces working OPAM packages.

### DIFF
--- a/config/dune
+++ b/config/dune
@@ -9,5 +9,5 @@
 (rule
  (targets coq_config.ml)
  (mode fallback)
- (deps %{project_root}/configure.ml %{project_root}/dev/ocamldebug-coq.run)
- (action (chdir %{project_root} (run %{ocaml} configure.ml -local -warn-error yes -native-compiler yes))))
+ (deps %{project_root}/configure.ml %{project_root}/dev/ocamldebug-coq.run (env_var COQ_CONFIGURE_PREFIX))
+ (action (chdir %{project_root} (run %{ocaml} configure.ml -native-compiler no))))
--- a/coq.opam
+++ b/coq.opam
@@ -15,6 +15,10 @@ depends: [
   "camlp5"
 ]
 
+build-env: [
+  [ COQ_CONFIGURE_PREFIX = "%{prefix}" ]
+]
+
 build: [
   [ "dune" "build" "@vodeps" ]
   [ "dune" "exec" "coq_dune" "_build/default/.vfiles.d" ]
--- a/coq.opam
+++ b/coq.opam
@@ -18,5 +18,5 @@ depends: [
 build: [
   [ "dune" "build" "@vodeps" ]
   [ "dune" "exec" "coq_dune" "_build/default/.vfiles.d" ]
-  [ "dune" "build" "-p" package "-j" jobs ]
+  [ "dune" "build" "-p" name "-j" jobs ]
 ]
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,7 +9,7 @@ stages:
 variables:
   # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
   # for reference]
-  CACHEKEY: "bionic_coq-V2018-09-20-V2"
+  CACHEKEY: "bionic_coq-V2018-09-21-V2"
   IMAGE: "$CI_REGISTRY_IMAGE:$CACHEKEY"
   # By default, jobs run in the base switch; override to select another switch
   OPAM_SWITCH: "base"
--- a/Makefile.dune
+++ b/Makefile.dune
@@ -7,6 +7,9 @@
 # DUNEOPT=--display=short
 
 BUILD_CONTEXT=_build/default
+COQ_CONFIGURE_PREFIX?=_build/install/default
+
+export COQ_CONFIGURE_PREFIX
 
 help:
 	@echo "Welcome to Coq's Dune-based build system. Targets are:"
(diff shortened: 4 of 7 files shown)


coq.opam
```diff
@@ -16,7 +16,8 @@ depends: [
 ]
 
 build: [
+  [ "./configure" "-prefix" prefix "-mandir" man "-libdir" "%{lib}/coq" "-coqide" "no" ]
```
Zimmi48: Why add this line? Isn't it redundant with the next steps?

ejgallego: Yeah sorry, I should have marked this PR as in-progress.

The thing is that I wasn't expecting people to use the opam file yet, it is not ready as we need a few changes to implement the build scheme properly, anyways I will turn this PR into a fix.

ejgallego:
Ok, so this seems to produce working OPAM packages at least for me.

I hit a strange problem where the old Coq was not properly uninstalled and I got some weird dynlink errors.

The approach used here clearly needs more work on coqlib location and configure, but I guess this could be good for now.

ejgallego:
In particular, this code will get better when we remove the `-local` flag from configure.

Zimmi48:
Time to change the PR title I suppose.

ejgallego:
> Time to change the PR title I suppose.

Changed; now a problem we face is whether to stay in OPAM 1.2 or move to OPAM 2.0. I guess the coq repos will have to stay in 1.2 for a while.

config/dune
```diff
@@ -9,5 +9,5 @@
 (rule
  (targets coq_config.ml)
  (mode fallback)
- (deps %{project_root}/configure.ml %{project_root}/dev/ocamldebug-coq.run)
- (action (chdir %{project_root} (run %{ocaml} configure.ml -local -warn-error yes -native-compiler yes))))
+ (deps %{project_root}/configure.ml %{project_root}/dev/ocamldebug-coq.run (env_var COQ_CONFIGURE_PREFIX))
+ (action (chdir %{project_root} (run %{ocaml} configure.ml -native-compiler no))))
```
SkySkimmer: Why not use -prefix? Why drop native-compile?

ejgallego: > Why not use -prefix?

That's a good question, the only way I've found to pass info to Dune is using configure variables. That being said, in Dune 1.3.0 we should be able to do `-prefix %{env:COQ_PREFIX}` and that will work, but current version cannot expand env vars as the environment needs to be threaded along the rules.

Anyways I am just being cautious here, maybe we could just set a fixed prefix here and that would work? The coqlib stuff is hard to understand for me.

> Why drop native-compile?

I can revert that, but I've found that it is misleading to compile Coq with native support enabled for now as `coq_dune` doesn't support it yet [support coming soon tho].

ejgallego: Note that I am not excessively happy with this solution, which I see more of a workaround than a real solution until we can indeed mostly drop `-prefix` from coqlib location.

SkySkimmer: Why does coq_dune need to do something wrt to native_compute?

ejgallego: It needs to do the same than the makefile does. By default, even with native-compute on `coqc` won't generate native versions of the files, unless it is passed a special flag. Makefiles do setup this special flag, so they produce the `.coq_native` stuff, etc...

That should be easy to add, just need a bit of time.

SkySkimmer (approved):

config/dune
ejgallego: In fact I prefer to wait until `coq_dune` can read library config files so we can enable it on a per basis with `(mode native)`.

gares:
can you remind us the difference between the `name` and the `package` variable in opam?

ejgallego:
> can you remind us the difference between the `name` and the `package` variable in opam?

`name` is a variable containing the package name, `package` is just not defined. Thus the original opam file had a typo.

gares (approved):

Zimmi48:
Merged but maybe we should have a `pkg:opam` job to make sure that we don't break it in the future.

ejgallego:
@zimmi48 we do already have it, even if it is a placeholder for now (doesn't call opam but the target opam uses) I guess we can enable the full test once we land opam 2.0 support, as the typical strategy to test this is to pin to a local repos, and that has considerably changed in 2.0

Zimmi48:
Sounds good.

PR #8135, opened 2018-07-24
Sphinx fixing of the beginning of the Tactics chapter.

Zimmi48:
**Kind:** documentation fix

Anecdotally this fixes #8129.

Issue #6764, opened 2018-02-14
Printing Notation regressed compared to 8.7


RalfJung:
#### Version

git master

#### Operating system
Linux

#### Description of the problem

In the `gen_proofmode` branch of https://gitlab.mpi-sws.org/FP/iris-coq/, printing notations regressed compared to Coq 8.7. This is visible for example in `theories/bi/update.v`: What used to be printed as `(|==> P) ⊢ P` is now printed as `(sbi_bi PROP).(bi_entails) (|==> P)%I P`.

The relevant notation `⊢` is defined in `theories/bi/interface.v`:
```
Notation "P ⊢ Q" := (bi_entails P%I Q%I) : stdpp_scope.
```
`bi_entails` is a primitive projection, so #6651 could be related. Or maybe this is caused by one of the recent scope handling changes?

RalfJung:
Small example:
```coq
Set Primitive Projections.

Record foo := Foo { foo_n : nat }.

Notation "'■' x" := (foo_n x%type) (at level 50).

Goal forall (f:foo), ■ f = ■ f.
```
With 8.7, this prints what I typed (`■ f = ■ f`), with master it prints `f.(foo_n) = f.(foo_n)`

JasonGross:
This is almost certainly #6651, because the printing worked fine on the only-slightly-outdated version of master I was previously using (47e43e229ab02a4dedc2405fed3960a4bf476b58)

RalfJung:
Ah, you're right. I thought the `%type` was the problem, but actually the problem persists without it.

ppedrot:
Yes, the problem is that notation matching is completely unaware of primitive projections. The input term to be matched is pretyped to a primitive projection and then back to a `GProj`, but the notation constr doesn't know anything about that as it is a `NApp` node. Obviously matching then fails because the two terms are unrelated.

Either we need to handle specially primitive projections in notation matching, or we need to also generate a notation rule for applied primitive projections. I don't know which one is better.

ppedrot:
I have a potential patch at ppedrot/coq@189528e253df092ac5d542cb945139649e5edce0 but it breaks due to pretyping being stupid in the handling of `GProj`, as hinted by the following comment.
```
    (* TODO: once GProj is used as an input syntax, use bidirectional typing here *)
```
On @RalfJung's example I get a pretype error saying that the hole cannot be inferred.

maximedenes:
[...]

PR #8444, opened 2018-09-10
Bump version number to 8.10+alpha.

silene:


ejgallego (approved):
Please merge yourself when you want this to land in master.

Zimmi48:
>Please merge yourself

This goes against the only rule written in all caps in the merging doc. We can debate if this rule is right (see also the impromptu discussion at https://github.com/coq/coq/pull/8135#issuecomment-416668326) but until this rule is changed, I think we should keep self-merging by the RM restricted to stable branches only.

ejgallego:
> This goes against the only rule written in all caps in the merging doc. We can debate if this rule is right (see also the impromptu discussion at #8135 (comment)) but until this rule is changed, I think we should keep self-merging by the RM restricted to stable branches only.

Well I felt that this was a clear special case due to wanting to put this as the first commit of the `master` branch; as 8.9 is created, merging anyways.

Zimmi48:
Indeed, that won't be the first commit on master. v8.9 is already well behind.

I don't understand what's going on with Iris but it's failing very consistently and not anywhere else it seems (cc @RalfJung).

RalfJung:
Ah. Coq 8.9 is whitelisted in Iris's test suite (which also checks the output printed by Coq, to make sure we do not regress our notation or error messages), it is allowed to have broken output because of https://github.com/coq/coq/issues/6764. I guess I can mark Coq 8.10 as equally broken?

Though it seems that is not the only change, the diff seems to indicate that Ltac backtraces also got worse in some places?

Zimmi48:
OK, so because #6764 is making Iris unusable, we missed other regressions? That's pretty bad, maybe it's time to fix this bug by reverting the feature *in master as well*. Unless the patch is ready of course.

RalfJung:
Well, we missed other changes. I consider them regressions as far as Iris is concerned, but they may have been deliberate. Once #6764 is fixed I can investigate, but currently there is just too much noise in the diff.

RalfJung:
I marked Coq 8.10 as broken in Iris, and bumped Iris in lambdaRust. In <30min, the new lambdaRust should appear on opam, fixing CI here.

ybertot:
coq-dpdgraph should be fixed now.  The problem was not about bumping versions, but about the fact that version 8.10 was recognized as a variant of version 8.1

ejgallego:
> I will fix Fiat tomorrow

Great @JasonGross , please hit the retry button on the CI when you do this.

JasonGross:
Fiat succeeded

ejgallego:
Great, will merge soon unless people oppose.

PR #8447, opened 2018-09-10
Cleaning in the retroknowledge

vbgl:
This is based on #8436 [merged].

Currently, the retroknowledge maps constr to roles; but these constrs really are names. This PR builds on the availability of maps indexed by `GlobRef.t`s in the kernel to implement the “reactive” map.

This PR also simplifies the syntax of the `Register` command, so as to ease a later extension of this command for Coqlib (#186).

This reveals a discrepancy between the VM and native compilations in the handling of the universes for the `int31` operators; I thus added a FIXME comment where appropriate.


kernel/retroknowledge.ml
```diff
@@ -55,6 +55,33 @@ type int31_field =
...
+  | "div" -> Int31Div
+  | "diveucl" -> Int31Diveucl
+  | "addmuldiv" -> Int31AddMulDiv
+  | "compare" -> Int31Compare
+  | "head0" -> Int31Head0
+  | "tail0" -> Int31Tail0
+  | "lor" -> Int31Lor
+  | "land" -> Int31Land
+  | "lxor" -> Int31Lxor
+  | _ -> raise Not_found
```
ppedrot: I am not sure whether this is intended, but this will raise an uncaught exception anomaly if people write random Register commands.

plugins/ltac/extratactics.ml4
```diff
@@ -544,21 +544,6 @@ END
...
-           [ let env = Global.env () in
-             let evd = Evd.from_env env in
-             let tc,_ctx = Constrintern.interp_constr env evd c in
-             let tc = EConstr.to_constr evd tc in
-             let f = Retroknowledge.(KInt31 (int31_field_of_string n)) in
-             Global.register f tc ]
-END
-
-
-
```
ppedrot: Why moving this into the hardcoded vernac AST?

maximedenes: @ppedrot where would you put it? It clearly does not belong to `extratactics`.

ppedrot: Dunno, but I think I prefer using extensions rather than hardcoding vernaculars. The former tends to lead to a more uniform syntax, as there is no way to cheat the parser or the printer, which usually happens spontaneously with hardcoded ASTs.

maximedenes: I'm not sure what you mean exactly. But in any case the priority is to fix the location. This command has nothing to do with Ltac. If you want to change the "hard coded" AST for one based on extensions (although I'm not sure it would be a good change), it should be done separately.

Now, if there's a place where this command could belong under its extension form, just let us know and we'll move it there.

ppedrot: That's not a hard requirement, essentially nitpicking. There is no point in delaying this welcome cleanup.

kernel/retroknowledge.ml
maximedenes: Good point, I think this should be addressed. Maybe it was a mistake to do the dispatch after the parser.

ppedrot: I think a proper error would suffice. It is definitely better to translate a string after the parsing phase.

vbgl: Done. Thanks.

maximedenes (approved):
Looks very good to me, but I'll let @ppedrot merge because I co-authored this patch.

maximedenes:
@ppedrot ping

PR #8448, opened 2018-09-10
Have `gitFull` in the nix-shell

CohenCyril:
**Kind:** infrastructure.

Update `default.nix` so that a `nix-shell` brings `gitFull` into the context, instead of just `git`, which was potentially masking extra git commands.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -75,7 +75,7 @@ stdenv.mkDerivation rec {
     ++ [ ocamlPackages.ounit rsync which ]
   )
   ++ optionals shell (
-    [ jq curl git gnupg ] # Dependencies of the merging script
+    [ jq curl gitFull gnupg ] # Dependencies of the merging script
     ++ (with ocamlPackages; [ merlin ocp-indent ocp-index utop ]) # Dev tools
   );
 


default.nix
```diff
@@ -75,7 +75,7 @@ stdenv.mkDerivation rec {
     ++ [ ocamlPackages.ounit rsync which ]
   )
   ++ optionals shell (
-    [ jq curl git gnupg ] # Dependencies of the merging script
+    [ jq curl (if shell then gitFull else git) gnupg ] # Dependencies of the merging script
```
vbgl: The `shell` variable is always true here.

See the definition of the [optionals](https://github.com/NixOS/nixpkgs/blob/f38e8c0e96a75862daccc7f978496631dfcf6037/lib/lists.nix#L209) function.

vbgl (changes requested):

default.nix
Zimmi48: Yeah, you can just replace `git` with `gitFull`.

CohenCyril: oh ok cool

Zimmi48 (approved):

vbgl (approved):

PR #8451, opened 2018-09-10
Print Universes Subgraph

SkySkimmer:
This adds an optional [Subgraph] part to the Print Universes command,
eg [Print Universes Subgraph(i j)] to print only constraints related
to i and j (and Prop/Set).

Hopefully this is helpful when researching universe inconsistencies in a monomorphic context.

I feel like there should also be some syntax to include all universes related to some constant or module but I'm not sure about the details. Notably it's currently impossible to tell it to include any unnamed universes making it not so useful.

### DIFF
--- a/doc/sphinx/language/gallina-extensions.rst
+++ b/doc/sphinx/language/gallina-extensions.rst
@@ -2155,6 +2155,12 @@ If `string` ends in ``.dot`` or ``.gv``, the constraints are printed in the DOT
 language, and can be processed by Graphviz tools. The format is
 unspecified if `string` doesn’t end in ``.dot`` or ``.gv``.
 
+.. cmdv:: Print Universes Subgraph(@names)
+
+Prints the graph restricted to the requested names (adjusting
+constraints to preserve the implied transitive constraints between
+kept universes).
+
 .. _existential-variables:
 
 Existential variables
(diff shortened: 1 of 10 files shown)


SkySkimmer:
The order of printing is different in 32bits? How?

ppedrot:
A difference in hashing, probably.

SkySkimmer:
The dependency on hashing order should be gone since #8475 so let's see what a rebase tells us.

SkySkimmer:
I was wrong, the hash order is used by the map fold used when printing.

SkySkimmer:
Should work now.

ppedrot (changes requested):
It's a feature, so you should advertise it in the CHANGES.

SkySkimmer:
Added CHANGES and a simple fix for `Print Universes "foo.dot".`

ppedrot (approved):
Will merge quickly if nobody complains.

doc/sphinx/language/gallina-extensions.rst
```diff
@@ -2155,6 +2155,12 @@ If `string` ends in ``.dot`` or ``.gv``, the constraints are printed in the DOT
 language, and can be processed by Graphviz tools. The format is
 unspecified if `string` doesn’t end in ``.dot`` or ``.gv``.
 
+.. cmdv:: Print Universes Subgraph(@names)
+
+Prints the graph restricted to the requested names (adjusting
```
Zimmi48: This block should have been properly indented.

PR #8452, opened 2018-09-10
Fix #8358: circular make dependency for camldevfiles

### LINKED ISSUE #8358: circular dependency in make
```
make: Circular plugins/.merlin.in <- submake dependency dropped.
make: Circular .merlin.in <- submake dependency dropped.
make: Circular ide/.merlin.in <- submake dependency dropped.
make: Circular META.coq.in <- submake dependency dropped.
```

SkySkimmer:


gares (approved):

gares:
I believe you are fixing the warning, but I'd like to have a clean picture of these dependencies

SkySkimmer:
The `%:: submake` line adds a dependency on submake for everything which doesn't have a rule. This includes already existing files (otherwise `make bin/coqtop` wouldn't work properly in rebuilds).
So we put an empty rule to tell make these files should be considered roots.

PR #8456, opened 2018-09-11
Revert #6651: Use r.(p) syntax to print primitive projections

### LINKED ISSUE #6764: Printing Notation regressed compared to 8.7

maximedenes:
Fixes #6764: Printing Notation regressed compared to 8.7

### DIFF
--- a/dev/ci/user-overlays/08456-fix-6764.sh
+++ b/dev/ci/user-overlays/08456-fix-6764.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+if [ "$CI_PULL_REQUEST" = "8456" ] || [ "$CI_BRANCH" = "fix-6764" ]; then
+    Elpi_CI_REF=overlay/8456
+fi
--- a/interp/constrexpr.ml
+++ b/interp/constrexpr.ml
@@ -114,7 +114,6 @@ and constr_expr_r =
   | CGeneralization of binding_kind * abstraction_kind option * constr_expr
   | CPrim of prim_token
   | CDelimiters of string * constr_expr
-  | CProj of qualid * constr_expr
 and constr_expr = constr_expr_r CAst.t
 
 and case_expr = constr_expr                 (* expression that is being matched *)
--- a/interp/constrexpr_ops.ml
+++ b/interp/constrexpr_ops.ml
@@ -177,12 +177,10 @@ let rec constr_expr_eq e1 e2 =
     | CDelimiters(s1,e1), CDelimiters(s2,e2) ->
       String.equal s1 s2 &&
       constr_expr_eq e1 e2
-    | CProj(p1,c1), CProj(p2,c2) ->
-      qualid_eq p1 p2 && constr_expr_eq c1 c2
   | (CRef _ | CFix _ | CCoFix _ | CProdN _ | CLambdaN _ | CLetIn _ | CAppExpl _
      | CApp _ | CRecord _ | CCases _ | CLetTuple _ | CIf _ | CHole _
      | CPatVar _ | CEvar _ | CSort _ | CCast _ | CNotation _ | CPrim _
-     | CGeneralization _ | CDelimiters _ | CProj _), _ -> false
+     | CGeneralization _ | CDelimiters _ ), _ -> false
 
 and args_eq (a1,e1) (a2,e2) =
   Option.equal (eq_ast explicitation_eq) e1 e2 &&
@@ -359,8 +357,6 @@ let fold_constr_expr_with_binders g f n acc = CAst.with_val (function
             (fold_local_binders g f n acc t lb) c lb) l acc
     | CCoFix (_,_) ->
       Feedback.msg_warning (strbrk "Capture check in multiple binders not done"); acc
-    | CProj (_,c) ->
-        f n acc c
   )
 
 let free_vars_of_constr_expr c =
@@ -439,8 +435,6 @@ let map_constr_expr_with_binders g f e = CAst.map (function
           let e'' = List.fold_left (fun e ({ CAst.v = id },_,_,_) -> g id e) e' dl in
           let d' = f e'' d in
           (id,bl',t',d')) dl)
-    | CProj (p,c) ->
-      CProj (p, f e c)
   )
 
 (* Used in constrintern *)
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -958,9 +958,6 @@ let rec extern inctx scopes vars r =
   | GCast (c, c') ->
       CCast (sub_extern true scopes vars c,
              map_cast_type (extern_typ scopes vars) c')
-  | GProj (p, c) ->
-    let pr = extern_reference ?loc Id.Set.empty (ConstRef (Projection.constant p)) in
-    CProj (pr, sub_extern inctx scopes vars c)
 
   in insert_coercion coercion (CAst.make ?loc c)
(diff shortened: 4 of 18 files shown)


ppedrot:
Why master? There is still a hope of a proper fix, isn't it?

maximedenes:
> Why master? There is still a hope of a proper fix, isn't it?

I've exhausted the time I could spend on it, but if someone has a fix now, it is still not too late to push it to master indeed. But there's little point in keeping this in master forever if nobody is making progress on cleaning up the primitive projection / compatibility constants mess. Is anybody working on it?

ppedrot:
@ejgallego The patch doesn't break the compilation, it just makes Iris essentially not usable because the printing is all messed up. I don't know whether this qualifies as unstability...

ejgallego:
> @ejgallego The patch doesn't break the compilation, it just makes Iris essentially not usable because the printing is all messed up. I don't know whether this qualifies as unstability...

This qualifies 100%, CI for master has a lot of testing disabled. We either drop Iris from the CI or do the testing properly.

gares:
> It is easy to revert the revert if someone has a fix.

I agree with @ejgallego wrt the severity, I don't agree with the fact that it is an easy revert. The API changes as CI shows, and if we re-revert then it will change again. I'd like an estimation of the cost of the proper fix of #6764  by @maximedenes

ejgallego:
> I don't agree with the fact that it is an easy revert. The API changes as CI shows,

Oh indeed I has assumed that the revert would have tried to be a bit more cautions API-wise [if just to save effort] as this is a printing issue.

Anyways this is the kind of issue thankfully we are facing less and less as the CI testing becomes more complete, we didn't properly test the original PR by lack of test cases, but this should now be fixed. It would be also good if we could add some test case to the test suite extracted from the Iris one.

A big black hole in our testing is performance tho; hopefully we can fix it some day.

maximedenes:
> Oh indeed I has assumed that the revert would have tried to be a bit more cautions API-wise

Initially, I thought we could just unplug the production of `GProj` nodes, while keeping the same AST type. However, it would be an intermediate state that should not stay as it is, and since I'm not aware of any work done to do the proper clean-up, I don't favor that option anymore.

If, on the other hand, someone commits to doing the clean-up, we could go for it.

SkySkimmer:
Do we actually know what the proper fix is? Should `Notation foo := (p r)` also match the primitive projection `r.(p)`?

RalfJung:
> Should Notation foo := (p r) also match the primitive projection r.(p)?

That would be the backwards compatible thing to do. Given that `p r` parses as `r.(p)` when used as a term, it seems rather surprising that the same does not happen when it is used as notation.

ppedrot:
IIRC, the problem is that desugaring of primitive projections happens at pretyping, which happens after translation notation.

ejgallego:
I am not sure if this should be considered pretyper or interp stuff.

If the second case applies, then @herbelin (or I with Hugo's blessings) could be the assignee.

ejgallego:
Anyways for a start the CI looks very very red.

maximedenes:
The test suite fails on `UnivBinders` it seems...

herbelin:
> I am not sure if this should be considered pretyper or interp stuff.
> If the second case applies, then @herbelin (or I with Hugo's blessings) could be the assignee.

It can probably be both. Please assign if you wish.

> Do we actually know what the proper fix is? Should Notation foo := (p r) also match the primitive projection r.(p)?

> Given that p r parses as r.(p) when used as a term, it seems rather surprising that the same does not happen when it is used as notation.

I'm unsure to be aware of all the changes made in the last months, but If we succeed that for a given type, there is no need to have both a "compatibility" and a regular representation of the projection, the "fix" is not needed any more, right? In all cases, both `(p r)` and `r.(p)` will be possible notations for the corresponding projection.

herbelin (approved):

herbelin:
The failure of `UnivBinders.out` is probably the `p.(punwrap)` to be changed in `punwrap p`. As for elpi, @gares, do we need synchronization and if yes what are your constraints? Otherwise, I shall merge (in master) a few hours after `UnivBinders.v` is fixed.

Zimmi48:
Note to @silene: backporting this to v8.9 will require updating the hash of elpi.

maximedenes:
> wrt elpi, it is for sure the constructor being removed. The patch is simple, so if this gets merged in v8.9 I can just make a v8.9 branch on elpi, do the fix myself, and let you point ci there.

@gares An elpi fix would be welcome, indeed.

silene:
I am a bit lost. Is that the pull request without which Iris is unusable with 8.9? Or is that something else? I am trying to understand the blocker tag here.

maximedenes:
> I am a bit lost. Is that the pull request without which Iris is unusable with 8.9?

Yes.

maximedenes:
@gares ping

maximedenes:
@herbelin should be good now, modulo CI

herbelin:
Question is passing (before merging): this PR has an overlay. It will have no role once merged, so do I remove it before merging?

gares:
The overlay automatically becomes inactive once the PR is merged.
The files are then removed from time to time...
So please go ahead.

herbelin:
> The overlay automatically becomes inactive once the PR is merged.
The files are then removed from time to time...

Yes, but that's maybe non optimal to commit them and to remove them afterwards.

Zimmi48:
It was argued that this serves some historical documentation purpose. I don't have a strong opinion on this but I suppose it's only limited noise since it's contained in a very specific folder.

herbelin:
> It was argued that this serves some historical documentation purpose.

In this case, I did it wrong by removing from PRs those of my overlays which were merged upstream.

maximedenes:
> In this case, I did it wrong by removing from PRs those of my overlays which were merged upstream.

That's not really wrong, IMHO each developer should do as they like for this matter.

Zimmi48:
When I said "it was argued", I didn't mean that I did. I really think this is not an important matter but personally, I leave the overlays around.

ejgallego:
Yeah I feel like there is not a lot to see here, these a "bureaucratic" files that as of today we have no better way to handle.

silene:
@gares 
> wrt elpi, it is for sure the constructor being removed. The patch is simple, so if this gets merged in v8.9 I can just make a v8.9 branch on elpi, do the fix myself, and let you point ci there.

Could you do so? Otherwise, it seems I would also have to backport #8552, right?

Zimmi48:
@silene Indeed, these are the only two options. If you go for the second route, don't forget that you have to run a pipeline with `SKIP_DOCKER=false`.

Zimmi48:
@silene Enrico has created such branch but it still requires to backport #8552 (see discussion in #6764).

Issue #8234, opened 2018-08-10
Qed and Defined treat abstract proofs differently


andres-erbsen:
#### Version

~~~~~
The Coq Proof Assistant, version 8.8.1 (July 2018)
compiled on Jul 5 2018 7:25:50 with OCaml 4.06.0
~~~~~

#### Description of the problem

Using `abstract` in a proof and ending it with `Qed` inlines the `abstract`-ed proofs (and spends a bunch of time doing that). Using `abstract` in a proof ended with `Defined` does not inline the subproofs but pollutes the namespace where the proof is used with the names of the `abstract`ed subterms (and disallows async compilation of reverse-dependencies), so it is not a real alternative. Further, this strange action at a distance does not seem to be documented anywhere.

As an immediate workaround, `Qed` should stop inlining the `abstract`-ed subterms and instead rename them into totally unpredictable names starting with `__do_not_use_this_`. If it's easy, `-mangle-names` should affect them. A more principled solution would probably be somewhat similar to `Local Definition`s.

JasonGross:
cc @gares (and maybe @ejgallego?)

JasonGross:
See also #3442 (@andres-erbsen is this a duplicate of #3442?)

JasonGross:
See also #4357

andres-erbsen:
#3442 and this issue are indeed about the same problem. I would prefer to keep this one open and close #3442 because I think the solution requested in #3442 does not go far enough.

gares:
The reason why abstract does not play well with asynchronous proofs is, well, clear I hope.
All that follows the Qed has already been processed before one discovers that the ltac scripts calls abstract, hence things are type checked in an environment that does not contain the abstracted sub proofs.

What is the real problem? I see what the problem of using Defined is. I don't see what is the problem of having these abstracted proofs being inlined (given that #4357 is solved I hope).

andres-erbsen:
[...]

Issue #6631, opened 2018-01-22
Derive Plugin gives "Error: Anomaly "more than one statement."" on abstract


JasonGross:
#### Version

8.8+alpha

#### Description of the problem

```coq
Require Import Coq.derive.Derive.

Derive f SuchThat (f = 1 + 1) As feq.
Proof.
  transitivity 2.
  2:abstract vm_cast_no_check (eq_refl 2).
  (* Error: Anomaly "more than one statement."
Please report at http://coq.inria.fr/bugs/. *)
```

JasonGross:
No, `vm_cast_no_check` is fine.  Note that it also breaks with
```coq
Require Import Coq.derive.Derive.
Require Import Coq.Compat.AdmitAxiom.

Derive f SuchThat (f = 1 + 1) As feq.
Proof.
  transitivity 2.
  2:admit.
```

The anomaly with `abstract` is already present in 8.5beta1, and the derive plugin does not seem to be present in 8.4pl6...

aspiwack:
Right. This doesn't sound terribly hard to fix. Let's just take the first open thingy as the name instead of requiring that there is only one.

JasonGross:
Great!  Shall I expect a PR from you fixing this soon?  (Or are you wanting me to make one?)

PR #8457, opened 2018-09-11
Refactoring of Micromega plugin (including new Simplex based solver)

fajb:
This PR aims at improving the Micromega plugin - more specifically the proof procedures.
To get there, some non trivial refactoring has been performed.

### New feature
There is now a Simplex based linear prover enabled by default. 
The Fourier based algorithm can be obtained by Unset Simplex.
The Simplex probably requires some polishing but should be more scalable.
Eventually, the objective would be to dump the Fourier code (mfourier.ml)

### Completeness improvement
lia and nia were known to have certain shortcomings regarding completeness.
The following goals (added to the test-suite) were not solvable.
Each of them requires some orthogonal modification to the proof-procedure.
- Lemma cutting_plane is solved by finding cutting planes proofs -- the Simplex is instrumental for doing that.
- Lemma linear_subst is solved by performing some non-linear pivoting before solving the goal.
- Lemma negative_square is solved by performing some more aggressive non-linear pre-processing 

`Lemma cutting_plane:
  forall (m : Z)   (M : Z) (x : Z) (i : Z)  (e1 : Z)  (e2 : Z)  (e5 : Z) (e6 : Z)
         (H2 : e5 >=  M)  (H11 :  i <  m)
         (H5 : 0 <=  i)  (H15 :  m < 4294967296)
         (H7 : 0 <=  x) (H26 : e5 < 4294967296)
         (H21 :  x +  i = 4294967296 * e6 + e5)
         (H9 :  x +  m = 4294967296 * e2 + e1)
         (H12 :  x < e1) (H13 : e1 <  M),
    False.`

`Lemma linear_subst :  forall (x y : Z), 2*x = 0 -> x * y = 0.`

`Lemma negative_square :  forall (x y: Z), - x*x >= 0 -> x * y =  0.`


ejgallego:
That's great @fajb , however the patch is based on a quite old version of the CI system, would you mind rebasing wrt current master? Thanks.

plugins/micromega/Lia.v
```diff
@@ -32,9 +32,9 @@ Ltac zchange :=
 
 Ltac zchecker_no_abstract := zchange ; vm_compute ; reflexivity.
 
-Ltac zchecker_abstract := zchange ; vm_cast_no_check (eq_refl true).
+Ltac zchecker_abstract := abstract (zchange ; vm_cast_no_check (eq_refl true)).
 
-Ltac zchecker := zchecker_no_abstract.
+Ltac zchecker := zchecker_abstract.
```
JasonGross: This breaks the use of `lia` inside `Derive` lemmas.  I seem to recall there also still being speed issues with inlining abstract inside Qed, but perhaps I am misremembering and @ppedrot or someone else has fixed this.  I guess benchmarking will tell.

plugins/micromega/ZMicromega.v
```diff
@@ -162,7 +162,7 @@ Declare Equivalent Keys psub RingMicromega.psub.
 Definition padd  := padd Z0  Z.add Zeq_bool.
 Declare Equivalent Keys padd RingMicromega.padd.
 
-Definition norm  := norm 0 1 Z.add Z.mul Z.sub Z.opp Zeq_bool.
+Definition normZ  := norm 0 1 Z.add Z.mul Z.sub Z.opp Zeq_bool.
 Declare Equivalent Keys norm RingMicromega.norm.
```
JasonGross: Shouldn't `norm` become `normZ ` here?  Also, can there be a test in the test-suite for this?

JasonGross (dismissed):
I think this deserves an entry in `CHANGES`.  Also, it sounds like the `Simplex` option is a compatibility setting, so it should be unset in `theories/Compat/Coq89.v` after a `Declare ML Module` to make the option available.

plugins/micromega/Lia.v
fajb: Ok, this confirms that 'abstract' is not to be used.
I will revert this.

plugins/micromega/ZMicromega.v
fajb: You're right, this should be normZ.

I am not sure a test is relevant. This is pure .v. The purpose of the renaming was to avoid a clash at extraction.

plugins/micromega/Lia.v
JasonGross: Perhaps this deserves a comment pointing to https://github.com/coq/coq/issues/8234#issuecomment-412332294 (the slowdown in `abstract` in `Qed` due to hashconsing large terms).  And it seems that #6631 has been solved, so I mis-spoke about `Derive` lemmas.

plugins/micromega/ZMicromega.v
fajb: Before you explained it I had no idea about the meaning of those declarations! This may have been useful in some previous version of Coq. 

Anyhow, they do not seem necessary (at least, I see no observable differences for the test-suite). I am open to any suggestion: either keeping or removing is fine by me.

gares:
Looks amazing, but the title of the PR only cover the less interesting aspect of the patch as far as I can tell. It should be updated, since it appears in the merge commit. It should contain the word simplex, IMO.

fajb:
@JasonGross, I have updated CHANGES. Yet, I really expect the Simplex option to be deprecated as soon as possible. I see this as a way to investigate regressions.
 
@gares, the PR title is updated. It was indeed missing an important point - refactoring can lead farther than initially expected.

ejgallego:
Note that we are hoping to raise the minimal ocaml version to 4.05 , so there is some chance you could remove some cruft from this pr

plugins/micromega/Lia.v
fajb: If 'abstract' proves to have a clear benefit, I will be happy to include it by default.
Otherwise, I propose the keep the existing (without abstract) behaviour.

gares (commented):
I'm supposed to review the test suite.

I don't get the following pattern:
- there is already a test sign_pos_neg (uses psatz)
- you add a test sign_pos_neg_nia (uses nia)

But in some case, eg sign_zero_zero:
- you replace psatz with nia

Can you explain that?

fajb:
@coq/micromega-maintainers , my take is that the code is almost ready to be integrated -- at least in terms of deductive power, I did not check performance.

Yet, there is a remaining issue found in fiat-crypto due to lia using an hypothesis not mentioned in a 
'Proof using xxx' construct.
This is unfortunate but probably sheer luck that it was not picked up by the previous implementation.

I can think of two approaches:
1) Ignore the issue and fix the code by explicitly clearing the unwanted hypotheses.
2) Make lia aware of the 'Proof using' annotations. For that I would need help as I have no idea
whether this is even possible. Can this information be obtained from a goal?

Zimmi48:
I don't think tactics are supposed to be aware of these annotations, I think you simply discovered a bug in `Proof using`. @gares?

doc/sphinx/addendum/micromega.rst
```diff
@@ -27,6 +27,12 @@ rationals ``Require Import Lqa`` and reals ``Require Import Lra``.
   generating a *proof cache* which makes it possible to rerun scripts
   even without `csdp`.
 
+  opt:: Simplex
```
Zimmi48: Use `flag`. The refman now distinguishes between non-boolean `option` and `flag`.

doc/sphinx/addendum/micromega.rst
```diff
@@ -27,6 +27,12 @@ rationals ``Require Import Lqa`` and reals ``Require Import Lra``.
...
+
+  This option (set by default) instructs the decision procedures to use the Simplex method for solving linear goals.
```
Zimmi48: Indent the description.

Zimmi48 (commented):
The doc of `lra` should also be updated since it says "This tactic is searching for linear refutations using Fourier elimination".

ejgallego:
> I can think of two approaches:
> 
> 1. Ignore the issue and fix the code by explicitly clearing the unwanted hypotheses.
> 2. Make lia aware of the 'Proof using' annotations. For that I would need help as I have no idea
>    whether this is even possible. Can this information be obtained from a goal?

@fajb , indeed you could also update the `Proof using` annotation to include the new hypotheses.

doc/sphinx/addendum/micromega.rst
fajb: fixed in 7133b24a2

fajb:
@ejgallego , I think the Proof using is right. However, as lia being unaware of it, it violates the contract.
There is actually a correct proof abiding to the Proof using annotation.

ejgallego:
@fajb the whole notion of `Proof using`, is, IMVHO broken; we are requiring the programmer to manually add information that is often generated automatically and worse, dynamic. To make matters worse, Coq can infer that information precisely. Hopefully a better implementation of sections means that we can remove proof using in the future.

Thus, breakage is not strange, and to me it is hard to blame the tactic for that. IMHO, tactics should not be aware of `Proof using` at all.

herbelin:
@ejgallego: I don't understand your view.
1. How can you check a document skipping the proof if you don't tell explicitly which section variables a constant in a section should be generalized over?
2. The information can be generated by processing the proof and Coq can do it automatically (`Set Suggest Proof Using`) so that, next time, the document can be parsed without processing the proof.
3. IMHO, this is the _natural expectation_ that tactics are aware of `Proof using`. The reason we don't do it is that we have still no way to remember whether an occurrence of `x:A` comes from the section or from a `clear x; intro x` with `x` of same type. Let me reevaluate if we can make the leap once for all without too much delay, since this issue comes regularly as blocking other PRs.

ejgallego:
> @ejgallego: I don't understand your view.

That's offtopic for this thread, anyways we have discussed this a few times before. Let's move the discussion to the proper issue.

JasonGross:
> Yet, there is a remaining issue found in fiat-crypto due to lia using an
hypothesis not mentioned in a
'Proof using xxx' construct.

>
Without looking at the code, my guess is that either there are multiple
proofs, and `lia` just picked a different one, or else `zify` is now adding
an equation about a let-bound section variable which was previously unused
(is it this PR that changes done := things to equations?). In any case, I'd
be fine with merging a patch to fiat-crypto clearing the unused variables
in that proof.  It'd be nice to have an exact diagnosis of what changed,
though, for the CHANGELOG if nothing else.

Zimmi48 (commented):
Thanks for the additional changes you did to the doc as part of this PR. However, note that you still have an indentation problem after `.. flag:: Simplex`.

doc/sphinx/addendum/micromega.rst
```diff
@@ -189,10 +194,10 @@ a proof.
 .. tacn:: nra
    :name: nra
 
-This tactic is an *experimental* proof procedure for non-linear
-arithmetic. The tactic performs a limited amount of non-linear
-reasoning before running the linear prover of `lra`. This pre-processing
-does the following:
+   This tactic is an *experimental* proof procedure for non-linear
+   arithmetic. The tactic performs a limited amount of non-linear
+   reasoning before running the linear prover of `lra`. This pre-processing
```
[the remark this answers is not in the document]
fajb: fixed in 5cc5d6d05

doc/sphinx/addendum/micromega.rst
```diff
@@ -27,6 +27,11 @@ rationals ``Require Import Lqa`` and reals ``Require Import Lra``.
   generating a *proof cache* which makes it possible to rerun scripts
   even without `csdp`.
 
+.. flag:: Simplex
+
+   This option (set by default) instructs the decision procedures to use the Simplex method for solving linear goals.If it is not set, the decision procedures are using Fourier elimination.
```
Zimmi48: You are missing a space after `goals.`

doc/sphinx/addendum/micromega.rst
```diff
@@ -112,11 +116,11 @@ and checked to be :math:`-1`.
 .. tacn:: lia
    :name: lia
 
-This tactic offers an alternative to the :tacn:`omega` and :tacn:`romega`
-tactics. Roughly speaking, the deductive power of lia is the combined deductive
-power of :tacn:`ring_simplify` and :tacn:`omega`. However, it solves linear
-goals that :tacn:`omega` and :tacn:`romega` do not solve, such as the following
-so-called *omega nightmare* :cite:`TheOmegaPaper`.
+   This tactic offers an alternative to the :tacn:`omega` and :tacn:`romega`
```
Zimmi48: Just a heads-up, #8419 removes `romega` and will be merged soon and it will create a conflict here.

vbgl:
There seem to be two minor issues left:

 - This breaks `fiat-crypto`: will someone fix it?
 - Does the `Unset Simplex.` incantation deserves to be cast in the `Coq89` compatibility module?

Zimmi48:
The answer to the second question is yes indeed.

fajb:
For fiat-crypto, I will investigate a bit but would be glad if someone has some  insight about  what is going on.

Zimmi48 (approved):

andres-erbsen:
@fajb re fiat-crypto: I think this [call to lia](https://github.com/mit-plv/fiat-crypto/blob/6d3702edad1a69a08565a288f1153b4853ba3b25/src/Specific/Framework/ArithmeticSynthesis/Montgomery.v#L134) now succeeds without using the context variable [`m_big`](https://github.com/mit-plv/fiat-crypto/blob/6d3702edad1a69a08565a288f1153b4853ba3b25/src/Specific/Framework/ArithmeticSynthesis/Montgomery.v#L43), and thus the Ltac caller of this lemma still passing in `m_big` [here](https://github.com/mit-plv/fiat-crypto/blob/6d3702edad1a69a08565a288f1153b4853ba3b25/src/Specific/Framework/ArithmeticSynthesis/Montgomery.v#L331) causes an error. If so, with the new `lia`, it should be possible to just not pass `m_small`, but a backwards-compatible fix would probably involve somehow "using" `m_big` in the original lemma even if `lia` doesn't use it.

plugins/micromega/polynomial.ml
```diff
@@ -10,667 +10,901 @@
 (*                                                                      *)
 (* Micromega: A reflexive tactic using the Positivstellensatz           *)
 (*                                                                      *)
-(*  Frédéric Besson (Irisa/Inria) 2006-20011                            *)
+(*  Frédéric Besson (Irisa/Inria) 2006-20018                            *)
```
vbgl: This change is very funny ;-)

vbgl (approved):

vbgl:
@JasonGross I think your comments have been addressed. Do you have any objection to merging?

Also, it seems that the `Simplex` option is not a compatibility option but a debugging feature.

Zimmi48:
I can't see this:

>Does the Unset Simplex. incantation deserves to be cast in the Coq89 compatibility module?

being implemented. But I think it was asked by Jason and also by me.

Zimmi48:
>Also, it seems that the Simplex option is not a compatibility option but a debugging feature.

Ah, I hadn't missed this part of your message. So this idea is that normally the behavior should not change and thus the option not be needed?

fajb:
@Zimmi48 , for most of the cases, the behaviour will be the same.
What can change is the set of hypotheses used by a proof.
 In the CI, this has manifested in fiat-crypto.

When no other regression is found, the goal is to remove this option.

vbgl:
Frédéric, can you please rebase so as to fix the `CHANGES` file; then I’ll merge. Thanks.

fajb:
@vbgl rebase done. Thanks for taking care of the merge.

vbgl:
[Benchmark results](https://ci.inria.fr/coq/job/benchmark-part-of-the-branch/518/console)

```
│               coq-sf-vfa │   45.91   46.28 -0.80 % │   126923496387   127287091232 -0.29 % │   201801036949   201920772893 -0.06 % │  536212  534944  +0.24 % │   28    5 +460.00 % │
│                coq-gpfsl │ 1194.46 1201.86 -0.62 % │  3328297021142  3347920485805 -0.59 % │  4375643012786  4412360037465 -0.83 % │ 1922356 1707752 +12.57 % │   19   11 +72.73 % │
│              coq-bignums │   96.81   96.77 +0.04 % │   268160783473   267577219986 +0.22 % │   381354045758   381261217503 +0.02 % │  537280  533668  +0.68 % │    6    0  +nan % │
│                coq-flocq │  585.12  584.84 +0.05 % │  1626760932676  1625351157779 +0.09 % │  2360279645247  2355732360180 +0.19 % │ 1714236 1712344  +0.11 % │  176   20 +780.00 % │
│       coq-mathcomp-field │  465.50  464.64 +0.19 % │  1295674879169  1291856412326 +0.30 % │  2115220170176  2115519266743 -0.01 % │  807696  803440  +0.53 % │    0  212 -100.00 % │
│                  coq-vst │ 1684.59 1681.46 +0.19 % │  4680449021173  4673894764169 +0.14 % │  6419173834660  6421869185122 -0.04 % │ 2226996 2227004  -0.00 % │  374  186 +101.08 % │
│              coq-unimath │ 1324.78 1322.23 +0.19 % │  3677306879780  3663059316110 +0.39 % │  6220333290772  6216566887491 +0.06 % │ 1076300 1072828  +0.32 % │  240   21 +1042.86 % │
│          coq-fiat-crypto │ 6282.72 6270.32 +0.20 % │ 17477500871123 17435666663143 +0.24 % │ 28490054885928 28479544451172 +0.04 % │ 2466256 2481344  -0.61 % │ 1093 1099 -0.55 % │
│ coq-mathcomp-real-closed │  186.60  186.22 +0.20 % │   517828381241   516674086546 +0.22 % │   784449325326   784331452136 +0.02 % │  837488  831624  +0.71 % │    0    8 -100.00 % │
│         coq-fiat-parsers │  712.47  710.91 +0.22 % │  1977382336703  1971992320622 +0.27 % │  3034373511520  3031247617415 +0.10 % │ 2858796 2847208  +0.41 % │    4  217 -98.16 % │
│   coq-mathcomp-character │  291.62  290.77 +0.29 % │   809278622075   806740496391 +0.31 % │  1175166405093  1175903635028 -0.06 % │ 1100640 1119852  -1.72 % │    0    0  +nan % │
│         coq-math-classes │  245.32  244.56 +0.31 % │   676281999035   673120965823 +0.47 % │   883148940972   881253384162 +0.22 % │  526648  524848  +0.34 % │   83   83 +0.00 % │
│               coq-sf-plf │   72.97   72.73 +0.33 % │   201675971033   201320693904 +0.18 % │   291994683610   292162271003 -0.06 % │  517096  515684  +0.27 % │   15    0  +nan % │
│    coq-mathcomp-solvable │  218.80  217.94 +0.39 % │   607331307079   604271720803 +0.51 % │   865077783523   863970151823 +0.13 % │  861244  861564  -0.04 % │    1   44 -97.73 % │
│                 coq-hott │  329.99  328.66 +0.40 % │   918751820045   916391143141 +0.26 % │  1383554900327  1383415170418 +0.01 % │  641552  598524  +7.19 % │   61    0  +nan % │
│             coq-compcert │  888.38  884.74 +0.41 % │  2463340519084  2451104766873 +0.50 % │  3534693360002  3544171619024 -0.27 % │ 1345212 1349540  -0.32 % │  102   32 +218.75 % │
│    coq-mathcomp-fingroup │   83.53   83.15 +0.46 % │   231172730781   229706593649 +0.64 % │   342129752862   341203937777 +0.27 % │  591724  587856  +0.66 % │   51    0  +nan % │
│                coq-sf-lf │   43.18   42.98 +0.47 % │   118435858370   118331727843 +0.09 % │   191600758269   191403637684 +0.10 % │  425144  422224  +0.69 % │  129    0  +nan % │
│   coq-mathcomp-odd-order │ 1411.50 1404.66 +0.49 % │  3928588252719  3913101070567 +0.40 % │  6597440627897  6592367881629 +0.08 % │ 1362848 1354808  +0.59 % │    0    0  +nan % │
│     coq-mathcomp-algebra │  204.60  203.54 +0.52 % │   565525978362   562710905646 +0.50 % │   792990022337   791897383765 +0.14 % │  647968  644172  +0.59 % │  176    5 +3420.00 % │
│            coq-fiat-core │  133.26  132.54 +0.54 % │   371634099141   369524887824 +0.57 % │   494136555625   493539663948 +0.12 % │  503896  500328  +0.71 % │   67   56 +19.64 % │
│               coq-geocoq │ 2831.38 2812.86 +0.66 % │  7870580666678  7819337125032 +0.66 % │ 12531121501649 12527252327483 +0.03 % │ 1391304 1390120  +0.09 % │  125   84 +48.81 % │
│                 coq-corn │ 1545.27 1534.66 +0.69 % │  4287062364537  4260973935395 +0.61 % │  6345265660369  6336355184377 +0.14 % │  841692  846640  -0.58 % │   44   44 +0.00 % │
│                coq-color │  732.92  727.28 +0.78 % │  2029265627611  2017285501914 +0.59 % │  2443821663827  2439598250129 +0.17 % │ 1391412 1389416  +0.14 % │   38  226 -83.19 % │
│           coq-coquelicot │  102.64  101.59 +1.03 % │   283034052842   281454751796 +0.56 % │   379235498090   377382267630 +0.49 % │  715880  717128  -0.17 % │  280   13 +2053.85 % │
│   coq-mathcomp-ssreflect │   64.53   63.84 +1.08 % │   176926239299   175645112850 +0.73 % │   250292904298   249407860867 +0.35 % │  541224  535144  +1.14 % │    2    0  +nan % │
│      coq-formal-topology │   56.83   56.21 +1.10 % │   154427768035   154063457029 +0.24 % │   232351047970   231863084608 +0.21 % │  480620  479788  +0.17 % │    0    0  +nan % │
└──────────────────────────┴─────────────────────────┴───────────────────────────────────────┴───────────────────────────────────────┴──────────────────────────┴───────────────────┘
```

fajb:
@vbgl  the tests are fairly dispointing -- to say the least.
Would it be possible to get finer grained results? On a goal basis? And wether it is lia or nia.

I would like to understand better before starting optimising...

ejgallego:
Indeed @fajb you can access the per-goal data here. I'd dare to say that the tests show now same timing, this kind of 1% noise is normal within the current infrastructure.

https://ci.inria.fr/coq/job/benchmark-part-of-the-branch/ws/518/html/

Well, it shows "not found" which I dunno why. Maybe you can rerun and see?

PR #8462, opened 2018-09-11
Zify: replace local definitions by equations

vbgl:
This also makes the `lia` tactic stronger.

### DIFF
--- a/plugins/omega/PreOmega.v
+++ b/plugins/omega/PreOmega.v
@@ -85,6 +85,7 @@ Ltac zify_binop t thm a b:=
 
 Ltac zify_op_1 :=
   match goal with
+   | x := ?t : Z |- _ => let h := fresh "heq_" x in pose proof (eq_refl : x = t) as h; clearbody x
    | |- context [ Z.max ?a ?b ] => zify_binop Z.max Z.max_spec a b
    | H : context [ Z.max ?a ?b ] |- _ => zify_binop Z.max Z.max_spec a b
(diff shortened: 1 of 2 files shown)


Zimmi48:
This enhancement should be backported to 8.9 since this is to cover a use case that was covered by `romega`.

fajb (approved):
I can't think of any situation where this could produce a regression.

For me, this is fine.

vbgl:
CompCert and VST have been updated to accommodate for this change; this PR is thus ready to merge.

herbelin:
Assignee should normally be @fajb...

Zimmi48:
@thery Please do not use GitHub's merge button. It breaks automation that is assuming that you are using the merge script. If you have forgotten about it, re-reading https://github.com/coq/coq/blob/master/dev/doc/MERGING.md cannot hurt.

ejgallego:
@Zimmi48 can't we just disable the GitHub merge button?

thery:
@Zimmi48 sorry about that, got already scolded by @maximedenes ;-)

Zimmi48:
>can't we just disable the GitHub merge button?

@ejgallego Unfortunately, that is not something that GitHub allows even if it is a very common feature-request. I invented a trick to still make it impossible to use it though: https://github.com/Zimmi48/certified-origin
But it is quite complex to use as it would require a push script that needs to be parametrized with a personal access token.
So in fact, the easiest way out, I think would be to make @coqbot the only one allowed to merge and the developers call it through a comment. This would anyways be a good first step in putting in place a bors-like merging process (coq/bot#18) so I'll work on this.

PR #8463, opened 2018-09-11
Remove Dischargedhypsmaps

maximedenes:
This looks like dead code and is getting in the way of a section path
clean-up.


ppedrot:
If I believe the git log, this was introduced to support Mowgli. I remember having read about it, but never saw the actual code since I am a Coq dev.

silene (approved):
I suppose that the fact there was no user of `get_discharged_hyps` makes the patch trivially correct. Let us just hope no external plugin was making use of it.

herbelin:
Indeed, it was used in the xml plugin. Didn't Claudio make a plugin out of it?

ejgallego:
Seems like CI needs a rebase due to overlays

maximedenes:
Can this get an assignee? @ppedrot ?

ppedrot:
@herbelin I think that the plugin was not maintained, but there is legacy code that is publicly available somewhere IIRC. Are you fine with the merge?

herbelin:
> Are you fine with the merge?

From the point of view of the Coq archive, yes. Otherwise, same remark as @silene. Maybe the commit could just add a reference to the xml plugin and tell to take contact with us if the feature was ever needed.

maximedenes:
Please let me know what to write and I'll do it. Thanks!

maximedenes:
> Before answering your question, can you give indications about the section-path clean-up that the removal of `Dischargedhypsmaps` will allow?

I'm trying to remove section paths from kernel names. But this clean-up makes sense independently.

herbelin:
> I'm trying to remove section paths from kernel names.

Do you mean a view where sections add no structure to the name space, even locally while inside the section? If yes, the point of Claudio with `Dischargedhypsmaps` was the opposite. The idea is that there should be a way to decompile a compiled state (or a `.vo`) to restitute the source document as it was. For instance, in your view, do you mean that in the following example:
```coq
Section S.
Variable n : nat.
Definition p := n+n.
About p.
```
the answer should be `Top.p` of type `forall n:nat, n=n` rather than `Top.S.p` of type `n=n`, as it is now?

I mean, this is a possible model (which tends to see sections as pure sugar) but there are other models (which see structure in sections).

> But this clean-up makes sense independently.

I feel that, in the balance, there is the interrogation between why to keep such a code which is useless, slightly cumbersome and somehow obfuscating the whole understanding of the code, and the idea that, virtually, this can still be useful if someone wants to write a `.vo` decompiler.

So, let me propose the following commit message:
```
Removing Dischargedhypsmap which is unused internally.

The Dischargedhypsmap table collected the segment of section variables
that constants defined in a section were originally depending on. It
was useful to reconstruct the structure of sections as originally
given in a source file. In particular this was used in Sacerdoti
Coen's plugin for exportation of Coq files to xml. There is no
information that this plugin, moved out of Coq in September 2014, was
finally maintained, even as an external plugin. So it seems that the
Dischargedhypsmap table is virtually not used anymore in the wild.

Please contact the developers if ever the need for such a table 
happens to be necessary for your work.
```

maximedenes:
@ppedrot can you merge this one? Thanks!

maximedenes:
@ppedrot ping

maximedenes:
Thanks!

PR #8407, opened 2018-09-05
[WIP] Fix numeral notations

gares:
See #8401 

@JasonGross feel free to push to this branch, I'm not doing the refactoring step myself.

PR #8464, opened 2018-09-11
Fix numeral notations

### LINKED ISSUE #8401: Incorrect scope of notations in coqide
#### Version

8.9

```
Require Import ZArith.
Lemma foo : True.
Proof.
set (w := Z.abs 0%Z).
exact I.
Qed.
```

The above script works fine when run line by line, but it breaks when moving through the document (e.g., with "go to end"). This makes Coqide unusable.

```
The term "0" has type "nat" while it is expected to have type "Z".
```

JasonGross:
**Kind:** bug fix

Fixes / closes #8401
Supersedes / closes #8407

Vernacular-command-registered numeral notations now live in the summary,
and the interpretation function for them is hard-coded.

Plugin-registered numeral notations are still unsynchronized, and only
the UIDs of these functions gets synchronized.  I am not 100% sure why
this is fine, but the test-suite file working suggests that it is fine.

<strike>I don't know the right way of printing constrs in warnings / error messages in Notation, so I have introduced a terrible kludge that I hope someone else will tell me how to fix (or fix for me).</strike>


ejgallego:
The kludges should go before integration, at least for error handling you can follow what's done in pretype error.

Dunno about warnings.

ejgallego:
So unless I'm mistaken all that remains in the warning is to print a `GlobRef.t`. I think this should be possible. Note that we hope to introduce a unified warning / error registration and printing mechanism in 8.10, once some goodies of OCaml 4.05.0 become available.

JasonGross:
@ejgallego Thanks, updated with something analogous to pretype_errors.

JasonGross:
And updated with @SkySkimmer 's pointer to `Nametab.pr_global_env`.  Kludges now gone.

gares (commented):
changes to the test suite are ok, I've no time to review the rest

gares:
> Plugin-registered numeral notations are still unsynchronized, and only
> the UIDs of these functions gets synchronized. I am not 100% sure why
> this is fine, but the test-suite file working suggests that it is fine.

The system state contains the list of plugins loaded.
When a worker receives the system state it loads the plugins listed there.

At least on paper your approach is OK.

ppedrot (commented):
I don't see anything outright wrong, but this is subtle code. As long as the CI succeeds it's OK for me.

herbelin:
Unless someone else volunteers, I can be the assignee considering that it is about notations.

>  I can split this into four commits:

As far as I'm concerned, making these four points explicit in the commit message seems enough

JasonGross:
Rebased to fix conflicts, and updated the commit message.

JasonGross:
Status is: awaiting review from @mattam82 and @herbelin ?

ejgallego:
> Status is: awaiting review from @mattam82 and @herbelin ?

I think a review from @herbelin should suffice. I could also take over but not very soon, I am already late with other reviews.

herbelin:
@jasongross: I'm sorry to backtrack on what I said, but if you could without too much pain make point 1 a separate commit, that would help to see the diff in the history...

JasonGross:
@herbelin No worries.  I have split it into the 4 commits (really, it is step 3 that does most of the moving).

herbelin (approved):
Seems ok, as far as I can judge.

herbelin:
OK, thanks very much for splitting. @mattam82, do you want to say something? @ejgallego, should I understand that it is ok for you on the basis that I already read the main parts?

ejgallego:
> @ejgallego, should I understand that it is ok for you on the basis that I already read the main parts?

I am good @herbelin ; this is ready to merge AFAICS.

Note that I am not sure I understand how workers work, but that's a separate issue and to the extend of my understanding this PR looks OK in that regard.

herbelin:
Let's target a merge on Monday then.

mattam82:
Nothing to add here
Le sam. 22 sept. 2018 à 14:52, Hugo Herbelin <notifications@github.com> a
écrit :

samuelgruetter:
Here's a bedrock2 build which works in with Coq 8.8, but fails with Coq master:
https://travis-ci.com/mit-plv/bedrock2/builds/86077603

The failure seems to be related to notations for `nat` and `Z`, and looking through the Coq commits of the past few days, I found that it might be related to this, so I thought I'd post this here. If you think it's not related to this PR, I'll be happy to open a separate issue.

samuelgruetter:
I see, so does that mean that this small breaking change between 8.8 and master is actually intended?

JasonGross:
Yes, it is intended.  See
https://github.com/coq/coq/blob/71296e6e91de5b3fea08ef4b34426a02304f00b8/CHANGES#L113-L129

samuelgruetter:
I see. Note to myself: Read the `CHANGES` file :wink:
And thanks for your help @JasonGross, the fix was just one line: https://github.com/mit-plv/bedrock2/commit/cbf50671ba2d9235060d4e4747d0860f127f3995

Zimmi48:
It seems that we are going to have many such questions following the release (like for axiomatic type class instances in 8.8). We should make sure that the release notes put emphasis on this breaking change and encourage people to read CHANGES.

PR #8465, opened 2018-09-11
Small cleanup of summary uses

SkySkimmer:
- Remove Hints.add_hints_init
    
   It was only used in ltac/rewrite which as a plugin is too late to
   affect init.

- Replace trivial uses of declare_summary with summary.refs
    
   NB: The one in notation.ml looks trivial but is needed to do
   with_notation_protection (used by inductive/fixpoint local notations).

### DIFF
--- a/library/nametab.ml
+++ b/library/nametab.ml
@@ -279,10 +279,10 @@ module ExtRefTab = Make(FullPath)(ExtRefEqual)
 module MPTab = Make(FullPath)(MPEqual)
 
 type ccitab = ExtRefTab.t
-let the_ccitab = ref (ExtRefTab.empty : ccitab)
+let the_ccitab = Summary.ref ~name:"ccitab" (ExtRefTab.empty : ccitab)
 
 type mptab = MPTab.t
-let the_modtypetab = ref (MPTab.empty : mptab)
+let the_modtypetab = Summary.ref ~name:"modtypetab" (MPTab.empty : mptab)
 
 module DirPath' =
 struct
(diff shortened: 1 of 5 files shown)


ppedrot (approved):

SkySkimmer:
>Remove Hints.add_hints_init

>It was only used in ltac/rewrite which as a plugin is too late to
affect init.


Looking into this some more, it seems like something resurrected its usage after b594d3f6f2c5ae36dd1da727bb1707271b0b784d (probably a merge/rebase issue). That explains why things worked even when making ltac a plugin made it run too late to be useful.

library/nametab.ml
```diff
@@ -279,10 +279,10 @@ module ExtRefTab = Make(FullPath)(ExtRefEqual)
 module MPTab = Make(FullPath)(MPEqual)
 
 type ccitab = ExtRefTab.t
-let the_ccitab = ref (ExtRefTab.empty : ccitab)
+let the_ccitab = Summary.ref ~name:"ccitab" (ExtRefTab.empty : ccitab)
```
ejgallego: Very minor nit, maybe a structured naming `tab:cci` would be nicer. Understanding the summary is essential in work with workers so every bit helps.

ejgallego (approved):
Nice cleanup.

ppedrot:
It seems that this needs a rebase.

ppedrot:
@mattam82 @silene Last round to complain. That's simple enough that if I don't get news from you guys I'll merge that soon.

PR #8467, opened 2018-09-12
Manual: fix documentation of the “fresh” tactic

vbgl:
-  The grammar is now accurate.
 -  The markup is now correct (w.r.t. the expected rendering).

### DIFF
--- a/doc/sphinx/proof-engine/ltac.rst
+++ b/doc/sphinx/proof-engine/ltac.rst
@@ -107,7 +107,7 @@ mode but it can also be used in toplevel definitions as shown below.
                      : | solve [ `expr` | ... | `expr` ]
                      : | idtac [ `message_token` ... `message_token`]
                      : | fail [`natural`] [`message_token` ... `message_token`]
-                     : | fresh | fresh `string` | fresh `qualid`
+                     : | fresh [ `component` … `component` ]
                      : | context `ident` [`term`]
                      : | eval `redexpr` in `term`
                      : | type of `term`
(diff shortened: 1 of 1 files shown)


doc/sphinx/proof-engine/ltac.rst
```diff
@@ -107,7 +107,7 @@ mode but it can also be used in toplevel definitions as shown below.
                      : | solve [ `expr` | ... | `expr` ]
                      : | idtac [ `message_token` ... `message_token`]
                      : | fail [`natural`] [`message_token` ... `message_token`]
-                     : | fresh | fresh `string` | fresh `qualid`
+                     : | fresh [ `component` | … | `component` ]
```
Zimmi48: I don't understand the use of the vertical bars. Shouldn't it be like for `idtac` above?

vbgl: Indeed… I can’t remember where the meta-syntax is documented…

Zimmi48:
>I can’t remember where the meta-syntax is documented…

Here is the (minimal) doc of the Sphinx `.. productionlist::` directive: http://www.sphinx-doc.org/en/stable/markup/para.html#directive-productionlist
and here is the Coq doc of the grammars in the refman: https://coq.inria.fr/refman/language/gallina-specification-language.html#about-the-grammars-in-the-manual

vbgl:
Thanks for the pointers, Théo. I’ve thus addressed your comment.

I don’t understand the CI failure: seems spurious.

Zimmi48:
Indeed, AppVeyor is a frequent source of spurious failures. It's just too close to the 1h timeout and in this case it went over the limit. Nothing to do with the doc which is not built in this job.

Zimmi48 (approved):

PR #8472, opened 2018-09-13
Remove "support" for universe polymorphic lists from romega

SkySkimmer:
I'm pretty sure it's impossible to use omega with polymorphic `list` (Coq currently fails in Relation_Operators) so let's not pretend.
The removed `new_univ_level` would probably cause an undeclared universe error if it were ever used.


SkySkimmer:
Looks like romega has no code owners so it's first come first serve for assignment ^^

PR #8475, opened 2018-09-13
Centralize the reliance on abstract universe context internals

ppedrot:
This PR reduces the number of calls to the internal-inspecting function `Univ.AUContext.instance` to one. This function allows to access data that is irrelevant, namely the global name of universes that were used at some point before getting quantified over. It used to be spread over various places that were only printing-related, but the current PR centralizes its use in `UnivNames`.

This is a preliminary step before completely getting rid of this function, but it requires more work to adapt the callers of the registering function in `UnivNames` that would echo across the code base.

Albeit the current mechanism is quite fragile (and this PR neither improves nor makes the situation worse), this should not disrupt printing of universes, apart for one specific instance that had to be tweaked in the test-suite. Namely, when writing an explicitly anonymous name in an instance, e.g.
```
About foo@{_}.
```
where `foo` is universe polymorphic, the old code would print the name of the global universe (e.g. `Top.1234`) while the new code would produce `Var(i)` where `i` is the index of the variable. I am not sure what is expected in that case, as the arguments are not real universes but mere names, but the new behaviour seems slightly more principled to me.

The PR fixes a few oddities of the name handling mechanism e.g. regarding sections, and one bug in printing.

### DIFF
--- a/engine/univNames.ml
+++ b/engine/univNames.ml
@@ -53,7 +53,7 @@ let empty_binders = Id.Map.empty
 
 let universe_binders_table = Summary.ref GlobRef.Map.empty ~name:"universe binders"
 
-let universe_binders_of_global ref : Name.t list =
+let universe_binders_of_global ref : Id.t list =
   try
     let l = GlobRef.Map.find ref !universe_binders_table in l
   with Not_found -> []
@@ -68,24 +68,25 @@ let subst_ubinder (subst,(ref,l as orig)) =
 let name_universe lvl =
   (** Best-effort naming from the string representation of the level. This is
       completely hackish and should be solved in upper layers instead. *)
-  try Name (Id.of_string_soft (Level.to_string lvl)) with _ -> Anonymous
+  Id.of_string_soft (Level.to_string lvl)
 
 let discharge_ubinder (_,(ref,l)) =
   (** Expand polymorphic binders with the section context *)
   let info = Lib.section_segment_of_reference ref in
   let sec_inst = Array.to_list (Instance.to_array (info.Lib.abstr_subst)) in
-  let map i lvl = match Level.name lvl with
-  | None -> Anonymous
-  | Some na ->
-    try
-      let qid = Nametab.shortest_qualid_of_universe na in
-      Name (snd (Libnames.repr_qualid qid))
-    with Not_found -> name_universe lvl
+  let map lvl = match Level.name lvl with
+    | None -> (* Having Prop/Set/Var as section universes makes no sense *)
+      assert false
+    | Some na ->
+      try
+        let qid = Nametab.shortest_qualid_of_universe na in
+        snd (Libnames.repr_qualid qid)
+      with Not_found -> name_universe lvl
   in
-  let l = List.mapi map sec_inst @ l in
+  let l = List.map map sec_inst @ l in
   Some (Lib.discharge_global ref, l)
 
-let ubinder_obj : GlobRef.t * Name.t list -> Libobject.obj =
+let ubinder_obj : GlobRef.t * Id.t list -> Libobject.obj =
   let open Libobject in
   declare_object { (default_object "universe binder") with
     cache_function = cache_ubinder;
@@ -103,7 +104,7 @@ let register_universe_binders ref ubinders =
   let univs = AUContext.instance (Global.universes_of_global ref) in
   let revmap = Id.Map.fold (fun id lvl accu -> LMap.add lvl id accu) ubinders LMap.empty in
   let map lvl =
-    try Name (LMap.find lvl revmap)
+    try LMap.find lvl revmap
     with Not_found -> name_universe lvl
   in
   let ubinders = Array.map_to_list map (Instance.to_array univs) in
@@ -112,21 +113,19 @@ let register_universe_binders ref ubinders =
 type univ_name_list = Names.lname list
 
 let universe_binders_with_opt_names ref names =
+  let orig = universe_binders_of_global ref in
   let udecl = match names with
-  | None -> universe_binders_of_global ref
-  | Some udecl -> List.map (fun na -> na.CAst.v) udecl
-  in
-  let () =
+  | None -> orig
+  | Some udecl ->
     try
-      let ctx = Global.universes_of_global ref in
-      let len = AUContext.size ctx in
-      if not (Option.is_empty names || Int.equal len (List.length udecl)) then
-        CErrors.user_err ~hdr:"universe_binders_with_opt_names"
-          Pp.(str "Universe instance should have length " ++ int len)
-    with Not_found -> ()
+      List.map2 (fun orig {CAst.v = na} ->
+          match na with
+          | Anonymous -> orig
+          | Name id -> id) orig udecl
+    with Invalid_argument _ ->
+      let len = List.length orig in
+      CErrors.user_err ~hdr:"universe_binders_with_opt_names"
+        Pp.(str "Universe instance should have length " ++ int len)
   in
-  let fold i acc = function
-  | Anonymous -> acc
-  | Name na -> Names.Id.Map.add na (Level.var i) acc
-  in
+  let fold i acc na = Names.Id.Map.add na (Level.var i) acc in
   List.fold_left_i fold 0 empty_binders udecl
--- a/engine/univNames.ml
+++ b/engine/univNames.ml
@@ -53,10 +53,10 @@ let empty_binders = Id.Map.empty
 
 let universe_binders_table = Summary.ref GlobRef.Map.empty ~name:"universe binders"
 
-let universe_binders_of_global ref : universe_binders =
+let universe_binders_of_global ref : Name.t list =
   try
     let l = GlobRef.Map.find ref !universe_binders_table in l
-  with Not_found -> Names.Id.Map.empty
+  with Not_found -> []
 
 let cache_ubinder (_,(ref,l)) =
   universe_binders_table := GlobRef.Map.add ref l !universe_binders_table
@@ -68,25 +68,19 @@ let subst_ubinder (subst,(ref,l as orig)) =
 let discharge_ubinder (_,(ref,l)) =
   (** Expand polymorphic binders with the section context *)
   let info = Lib.section_segment_of_reference ref in
-  let sec_inst = info.Lib.abstr_subst in
-  let shift = Instance.length sec_inst in
-  let map lvl = match Level.var_index lvl with
-  | None -> lvl
-  | Some n -> Level.var (n + shift)
-  in
-  let l = Id.Map.map map l in
-  let fold i accu lvl = match Level.name lvl with
-  | None -> accu
+  let sec_inst = Array.to_list (Instance.to_array (info.Lib.abstr_subst)) in
+  let map i lvl = match Level.name lvl with
+  | None -> Anonymous
   | Some na ->
     try
       let qid = Nametab.shortest_qualid_of_universe na in
-      Id.Map.add (snd (Libnames.repr_qualid qid)) (Level.var i) accu
-    with Not_found -> accu
+      Name (snd (Libnames.repr_qualid qid))
+    with Not_found -> Anonymous
   in
-  let l = Array.fold_left_i fold l (Instance.to_array sec_inst) in
+  let l = List.mapi map sec_inst @ l in
   Some (Lib.discharge_global ref, l)
 
-let ubinder_obj : GlobRef.t * universe_binders -> Libobject.obj =
+let ubinder_obj : GlobRef.t * Name.t list -> Libobject.obj =
   let open Libobject in
   declare_object { (default_object "universe binder") with
     cache_function = cache_ubinder;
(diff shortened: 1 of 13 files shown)


SkySkimmer:
I was just looking at the universe naming system (I was looking into reducing universe redeclarations further after #8341 and universe naming has some), it feels a bit fragmented. I think I'll try to do a cleanup / remind myself how it works before a full review.

I guess for `Print foo@{_}.` it would make sense to print the normal name.
ie `Print foo@{u _}.` would mean "rename the first universe to `u` and use whatever name is normal for the second" where currently it means "rename the first universe to `u` and use the raw name that was there before abstracting for the second".
~~~coq
Definition foo@{i j +} : Type := Type@{i} -> Type@{j}.
Print foo@{u _ _}. 
(* now: foo : Type@{Top.3} = Type@{u} -> Type@{Top.2}
this PR: foo : Type@{Var 2} = Type@{u} -> Type@{Var 1}
proposed: foo : Type@{Top.3} = Type@{u} -> Type@{j} *)
~~~
Not sure what I was thinking when I decided on the current behaviour. I doubt anyone relies on it.

SkySkimmer:
I'm going to assign to myself, complain if you disagree.

ppedrot:
I agree that the universe naming system is all over the place. My plan is to move the bound names into the kernel (more precisely in the abstract contexts themselves), similarly to what we do for rel naming.

engine/univNames.ml
```diff
@@ -65,10 +66,27 @@ let subst_ubinder (subst,(ref,l as orig)) =
   let ref' = fst (Globnames.subst_global subst ref) in
   if ref == ref' then orig else ref', l
 
+let name_universe lvl =
+  (** Best-effort naming from the string representation of the level. This is
+      completely hackish and should be solved in upper layers instead. *)
+  try Name (Id.of_string_soft (Level.to_string lvl)) with _ -> Anonymous
```
SkySkimmer: What exception can be raised here?

engine/univNames.ml
```diff
@@ -65,10 +66,27 @@ let subst_ubinder (subst,(ref,l as orig)) =
...
+  let sec_inst = Array.to_list (Instance.to_array (info.Lib.abstr_subst)) in
+  let map i lvl = match Level.name lvl with
+  | None -> Anonymous
+  | Some na ->
+    try
+      let qid = Nametab.shortest_qualid_of_universe na in
+      Name (snd (Libnames.repr_qualid qid))
+    with Not_found -> name_universe lvl
+  in
+  let l = List.mapi map sec_inst @ l in
```
SkySkimmer: The `i` in `mapi` does not seem to be used here.

engine/univNames.ml
```diff
@@ -79,28 +97,37 @@ let ubinder_obj : GlobRef.t * universe_binders -> Libobject.obj =
...
+  | Some udecl -> List.map (fun na -> na.CAst.v) udecl
+  in
+  let () =
+    try
+      let ctx = Global.universes_of_global ref in
+      let len = AUContext.size ctx in
+      if not (Option.is_empty names || Int.equal len (List.length udecl)) then
+        CErrors.user_err ~hdr:"universe_binders_with_opt_names"
+          Pp.(str "Universe instance should have length " ++ int len)
+    with Not_found -> ()
```
SkySkimmer: When do we get Not_found here?

vernac/himsg.ml
```diff
@@ -896,7 +896,8 @@ let explain_not_match_error = function
       quote (Printer.safe_pr_lconstr_env env (Evd.from_env env) t2)
   | IncompatibleConstraints cst ->
     str " the expected (polymorphic) constraints do not imply " ++
-      let cst = Univ.AUContext.instantiate (Univ.AUContext.instance cst) cst in
+      let cst = Univ.UContext.constraints (Univ.AUContext.repr cst) in
+      (** FIXME: provide a proper naming for the bound variables *)
```
SkySkimmer: I guess we should look it up from the module field. If we're comparing functors that won't work though.

SkySkimmer:
Pushed a commit with improvements and a commit with a new test for Inductive in sections (this used to fail before this PR).

Needs rebasing, I'll leave that to you.

SkySkimmer:
(the conflict is trivial stuff from the blogrefmap rename)

ppedrot:
I don't agree with universe names being identifiers. They can definitely be anonymous, in which case we should generate some fresh name in the context for printing, just as what we do for rels. We really ought to treat this binders as we do for term binders.

mattam82:
I guess we should not confuse global universe names which one can’t just
alpha-rename and local universe binders of polymorphic constructs.

SkySkimmer:
The universes I get when I `Show Proof` / `Show Universes` during a proof should be the same as the ones I get when I `Print foo` afterwards (modulo universes that get minimised away etc). As such we are not allowed to use Anonymous.

engine/univNames.ml
SkySkimmer: If some code generates universes with invalid unicode I think it deserves to explode.

ppedrot:
I still don't agree. Not all definitions get to have universe names, and instead of generating random names at registration time, we should have some mechanism generating fresh names at printing time. 

We already talked about that, and we agreed that debugging universe polymorphic code is a hell. I don't think that adding to it a confusion between unnamed locally bound polymorphic variables printed as their internal name in the original proof and unnamed global monomorphic variables which are still in the environment helps a lot...

mattam82:
So your proposal @ppedrot is to have Anonymous in abstract context “instances” and pretty-print them by generating fresh names, is that the idea?

SkySkimmer:
I think my patch produces a better result than raw `Var`, and it doesn't go out of its way to do that (if anything it makes the code a bit simpler). So until we generate names it should be fine.

SkySkimmer:
That printing difference in 32bit seems quite annoying. I guess when the underlying universe are `Var` (with only a couple vars so small indices) the hash doesn't overflow so it's stable but in 32bit it does overflow with dirpath. However how did we not hit this issue with monomorphic universes?
Do we know what kind of perf consequences we get from ordering universes in algebraics using `natural_compare`? If it's bad then I guess we have to order when printing.

SkySkimmer:
I tried removing the hash based compare for universe levels and got this benchmark:
~~~

│                coq-gpfsl │ 1186.23 1200.20 -1.16 % │  3305500846647  3344857960023 -1.18 % │  4353331532033  4396092974764 -0.97 % │ 2158368 1713796 +25.94 % │   20    0  +nan % │
│   coq-mathcomp-odd-order │ 1407.15 1418.01 -0.77 % │  3917294700676  3949494134431 -0.82 % │  6604050291268  6600541868628 +0.05 % │ 1358844 1361700  -0.21 % │    0    0  +nan % │
│              coq-unimath │ 1321.93 1328.38 -0.49 % │  3669139648211  3689035795385 -0.54 % │  6228656641162  6233728262401 -0.08 % │ 1024228 1023876  +0.03 % │  233   21 +1009.52 % │
│                  coq-vst │ 1652.47 1660.38 -0.48 % │  4593308684708  4616327314009 -0.50 % │  6303407200484  6306190007334 -0.04 % │ 2226832 2228116  -0.06 % │  256  194 +31.96 % │
│    coq-mathcomp-fingroup │   83.20   83.58 -0.45 % │   229787867751   230458724882 -0.29 % │   340652168520   341235988733 -0.17 % │  587908  587884  +0.00 % │   50    0  +nan % │
│      coq-formal-topology │   56.02   56.27 -0.44 % │   153654962786   153645820802 +0.01 % │   231116574541   231477301975 -0.16 % │  479696  479796  -0.02 % │    0    0  +nan % │
│               coq-geocoq │ 2812.73 2824.25 -0.41 % │  7816805138971  7852183452049 -0.45 % │ 12531106307058 12532059028653 -0.01 % │ 1390156 1390028  +0.01 % │  112   84 +33.33 % │
│              coq-bignums │   97.54   97.92 -0.39 % │   269269008797   270027522808 -0.28 % │   384266910818   384627157871 -0.09 % │  537792  537564  +0.04 % │    5    0  +nan % │
│          coq-fiat-crypto │ 6251.36 6269.12 -0.28 % │ 17400229985137 17437097250022 -0.21 % │ 28483578318062 28516199951720 -0.11 % │ 2562612 2524872  +1.49 % │ 1245 1370 -9.12 % │
│                 coq-hott │  328.86  329.73 -0.26 % │   913717665912   918611425559 -0.53 % │  1377980791781  1379150628635 -0.08 % │  598436  598372  +0.01 % │   41    0  +nan % │
│                 coq-corn │ 1537.00 1540.60 -0.23 % │  4266629325086  4274118673290 -0.18 % │  6335708880852  6336389098963 -0.01 % │  870296  849124  +2.49 % │   44   44 +0.00 % │
│             coq-compcert │  887.80  889.04 -0.14 % │  2459910586595  2466132302645 -0.25 % │  3547645398985  3551379317055 -0.11 % │ 1340780 1352876  -0.89 % │  215   33 +551.52 % │
│   coq-mathcomp-character │  290.09  290.44 -0.12 % │   806472778872   807157309173 -0.08 % │  1175591964812  1175853998318 -0.02 % │ 1114248 1116448  -0.20 % │    0    0  +nan % │
│       coq-mathcomp-field │  463.90  464.39 -0.11 % │  1291197539722  1291953380883 -0.06 % │  2117187057913  2115659377062 +0.07 % │  807692  807736  -0.01 % │   47  226 -79.20 % │
│         coq-math-classes │  244.79  245.04 -0.10 % │   673989644262   675186637006 -0.18 % │   881236899453   881225599404 +0.00 % │  525008  525940  -0.18 % │   83   83 +0.00 % │
│    coq-mathcomp-solvable │  218.01  218.17 -0.07 % │   604882379921   605666816857 -0.13 % │   864598813806   864802881567 -0.02 % │  828252  862072  -3.92 % │    1   10 -90.00 % │
│            coq-fiat-core │  133.05  133.12 -0.05 % │   369841716626   370694626657 -0.23 % │   492886034057   493702177987 -0.17 % │  499852  500112  -0.05 % │   78   55 +41.82 % │
│         coq-fiat-parsers │  714.70  714.28 +0.06 % │  1983277163704  1980804411213 +0.12 % │  3044184115029  3042071733732 +0.07 % │ 2848028 2847288  +0.03 % │    5  201 -97.51 % │
│           coq-coquelicot │  101.86  101.79 +0.07 % │   281040444730   281789079781 -0.27 % │   377356972240   377619028789 -0.07 % │  717108  717148  -0.01 % │  278   13 +2038.46 % │
│               coq-sf-vfa │   45.84   45.80 +0.09 % │   126345544616   126832342776 -0.38 % │   201529690591   201733298561 -0.10 % │  534700  534624  +0.01 % │   28    5 +460.00 % │
│                coq-color │  729.50  728.69 +0.11 % │  2021939934321  2020780727935 +0.06 % │  2440010739558  2441059860528 -0.04 % │ 1389400 1339480  +3.73 % │   38  260 -85.38 % │
│     coq-mathcomp-algebra │  203.57  203.34 +0.11 % │   563321710994   564044183632 -0.13 % │   791290492808   791499138354 -0.03 % │  643200  643356  -0.02 % │  176    5 +3420.00 % │
│                coq-flocq │  569.93  569.00 +0.16 % │  1583534514333  1583014170584 +0.03 % │  2297039232887  2295069378178 +0.09 % │ 1532800 1712104 -10.47 % │  158    9 +1655.56 % │
│ coq-mathcomp-real-closed │  186.13  185.80 +0.18 % │   517572839040   517190479713 +0.07 % │   783340414991   783800723349 -0.06 % │  832724  832416  +0.04 % │    0    0  +nan % │
│               coq-sf-plf │   72.74   72.61 +0.18 % │   200835243792   201424140468 -0.29 % │   291869543661   292114761976 -0.08 % │  515768  515692  +0.01 % │   15    0  +nan % │
│                coq-sf-lf │   42.78   42.67 +0.26 % │   117359076583   117700948096 -0.29 % │   190988603277   191140440046 -0.08 % │  422220  422004  +0.05 % │  130    0  +nan % │
│   coq-mathcomp-ssreflect │   63.94   63.60 +0.53 % │   175608740474   175338211520 +0.15 % │   248821021761   248976351408 -0.06 % │  533268  533352  -0.02 % │    2    0  +nan % │
└──────────────────────────┴─────────────────────────┴───────────────────────────────────────┴───────────────────────────────────────┴──────────────────────────┴───────────────────┘
~~~
So I added it to this PR, it should fix the printing issue on 32bit.

ppedrot:
I think that we should remove hash comparison in the `compare` function and rely on `HMap` rather than `Map` for level-indexed maps. Then we bench to compare this.

SkySkimmer:
Isn't that what I did?

ppedrot:
My bad, I though `LMap` was not defined through `HMap`. Sorry, you're definitely right and your patch is exactly what I wanted.

ppedrot:
Who's merging that one? @SkySkimmer has written part of the PR so this is probably not allowed for him to do it. Also, I insist on reverting part of 1197f045617cd32f4ba2e8e43840bd5ff75d2f26, so that I won't have to backtrack on the API later on when abstract universe contexts use names. Are you fine with it @SkySkimmer?

engine/univNames.mli
```diff
@@ -27,15 +27,14 @@ type universe_binders = Univ.Level.t Names.Id.Map.t
...
-    If [l] is [Some univs] return the universe binders naming the levels of [u] by [univs] (skipping Anonymous).
-    May error if the lengths mismatch.
+    If [l] is [Some univs] return the universe binders naming the bound levels
+    of [ref] by [univs] (skipping Anonymous). May error if the lengths mismatch.
 
-    Otherwise return [universe_binders_of_global ref]. *)
+    Otherwise return the bound universe names registered for [ref]. *)
 val universe_binders_with_opt_names : Names.GlobRef.t ->
-  Univ.Level.t list -> univ_name_list option -> universe_binders
+  univ_name_list option -> universe_binders
```
SkySkimmer: >Also, I insist on reverting part of 1197f04, so that I won't have to backtrack on the API later on when abstract universe contexts use names.

This still takes names, so what API do you mean?

ppedrot:
@mattam82 ping

ppedrot:
*Completely not with a gun on my temple:* Erm, erm, "@SkySkimmer's commit are the best I've seen in my life!".

SkySkimmer:
The idea discussed on gitter is that we respectively approve the other's commit and then I merge.
I'll leave a day or so for @mattam82 or anyone else to oppose.

SkySkimmer (approved):

Issue #8027, opened 2018-07-09
Design of attributes needs to be refined


maximedenes:
#7920 has raised many questions on:
1. concrete syntax for attributes
1. intended use cases
1. further extensions in sight

It would be good to summarize and discuss them here before preparing PRs addressing those points.

ejgallego:
We also agreed that discussion will mostly resume on September due to holidays; a few points off of my head:
- scope, prefix vs suffix: What is the scope of attributes? Are we happy with vernacular-only attributes, or do we want a more refined scope as in OCaml. Also, prefix vs suffix; what do we prefer? Is that easy to implemented with the parser.
- scope II: what should `#[foo] Module Foo.` do? Do we support some form of attribute inheritance?
- payload: How do we write attributes with payload? What about special symbols [attributes may be used for some translation effort] How do we reconcile the syntax of payload with options [it seems bad to have `Set Warnings +foo` and `#[enable(foo_warning)]` as both valid commands.
- Dissonance with the actual syntax: indeed that's a concern, attributes look unnatural in the current style and that may be a problem for users.
- use cases? What attributes do we expect to have. So far, we haven't proposed many new things in this area; what about deriving-style definitions? etc...

maximedenes:
> scope, prefix vs suffix: What is the scope of attributes? Are we happy with vernacular-only attributes, or do we want a more refined scope as in OCaml. Also, prefix vs suffix; what do we prefer? Is that easy to implemented with the parser.

The scope of attributes is currently one vernac command. I think it is good as a first step. Do you think we should aim for more immediately?

About prefix vs suffix, I'm not sure what the question is. Do you want a poll among the team? We could certainly do that.

> scope II: what should #[foo] Module Foo. do? Do we support some form of attribute inheritance?

This passes the `foo` attribute to the module vernacular. For example, scoping of `Module` will probably be relevant, as well as deprecation. We may think about a form of inheritance, but I'm not sure it is required in the first step. Again, feel free to comment on what you think is necessary.

> payload: How do we write attributes with payload? What about special symbols [attributes may be used for some translation effort]

I'm not sure I understand what you mean here. Attributes can be written e.g. `#attribute(foo = val)`. Is that what you mean by payload?

> How do we reconcile the syntax of payload with options [it seems bad to have Set Warnings +foo and #[enable(foo_warning)] as both valid commands.

`#[enable(foo_warning)]` is not a valid command.

[...]

PR #8488, opened 2018-09-17
Warning when using automatic template polymorphism

SkySkimmer:
I also annotated every spot in the stdlib/test suite where it triggered. Probably most of them don't need it. Should I try doing some filtering?
I also disabled template polymorphism for types declared by funind, this seems to work fine.

I'm not sure that we actually want this. Thoughts? What's the next step for template polymorphism?
Should we try disabling the automatic decision by default (and make PRs to 3rd parties to add the option or annotate with attributes)?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -135,6 +135,17 @@ Universes
   for the "Private Polymorphic Universes" option (and Unset it to get
   the previous behaviour).
 
+Inductives
+
+- An option and attributes to control the automatic decision to
+  declare an inductive type as template polymorphic were added.
+  Warning "auto-template" will trigger when an inductive is
+  automatically declared template polymorphic without the attribute.
+
+Funind
+
+- Inductive types declared by Funind will never be template polymorphic.
+
 Misc
 
 - Option "Typeclasses Axioms Are Instances" is deprecated. Use Declare Instance for axioms which should be instances.
--- a/plugins/funind/glob_term_to_relation.ml
+++ b/plugins/funind/glob_term_to_relation.ml
@@ -1494,7 +1494,7 @@ let do_build_inductive
   let _time2 = System.get_time () in
   try
     with_full_print
-      (Flags.silently (ComInductive.do_mutual_inductive ~template:None None rel_inds false false false ~uniform:ComInductive.NonUniformParameters))
+      (Flags.silently (ComInductive.do_mutual_inductive ~template:(Some false) None rel_inds false false false ~uniform:ComInductive.NonUniformParameters))
       Declarations.Finite
   with
     | UserError(s,msg) as e ->
--- a/plugins/micromega/EnvRing.v
+++ b/plugins/micromega/EnvRing.v
@@ -118,6 +118,7 @@ Section MakeRingPol.
     - (Pinj i (Pc c)) is (Pc c)
  *)
 
+ #[universes(template)]
  Inductive Pol : Type :=
   | Pc : C -> Pol
   | Pinj : positive -> Pol -> Pol
@@ -939,6 +940,7 @@ Qed.
 
  (** Definition of polynomial expressions *)
 
+ #[universes(template)]
  Inductive PExpr : Type :=
   | PEc : C -> PExpr
   | PEX : positive -> PExpr
--- a/plugins/micromega/RingMicromega.v
+++ b/plugins/micromega/RingMicromega.v
@@ -289,6 +289,7 @@ destruct o' ; rewrite H1 ; now rewrite  (Rplus_0_l sor).
  now apply (Rplus_nonneg_nonneg sor).
 Qed.
 
+#[universes(template)]
 Inductive Psatz : Type :=
 | PsatzIn : nat -> Psatz
 | PsatzSquare : PolC -> Psatz
@@ -685,6 +686,7 @@ end.
 Definition  eval_pexpr : PolEnv -> PExpr C -> R :=
  PEeval rplus rtimes rminus ropp phi pow_phi rpow.
 
+#[universes(template)]
 Record Formula (T:Type) : Type := {
   Flhs : PExpr T;
   Fop : Op2;
--- a/plugins/micromega/Tauto.v
+++ b/plugins/micromega/Tauto.v
@@ -21,6 +21,7 @@ Require Import Bool.
 Set Implicit Arguments.
 
 
+  #[universes(template)]
   Inductive BFormula (A:Type) : Type :=
   | TT   : BFormula A
   | FF   : BFormula A
--- a/plugins/micromega/VarMap.v
+++ b/plugins/micromega/VarMap.v
@@ -30,6 +30,7 @@ Section MakeVarMap.
   Variable A : Type.
   Variable default : A.
 
+  #[universes(template)]
   Inductive t  : Type :=
   | Empty : t
   | Leaf : A -> t
--- a/plugins/rtauto/Bintree.v
+++ b/plugins/rtauto/Bintree.v
@@ -81,10 +81,12 @@ Section Store.
 
 Variable A:Type.
 
+#[universes(template)]
 Inductive Poption : Type:=
   PSome : A -> Poption
 | PNone : Poption.
 
+#[universes(template)]
 Inductive Tree : Type :=
    Tempty : Tree
  | Branch0 : Tree -> Tree -> Tree
@@ -177,6 +179,7 @@ generalize i;clear i;induction j;destruct T;simpl in H|-*;
 destruct i;simpl;try rewrite (IHj _ H);try (destruct i;simpl;congruence);reflexivity|| congruence.
 Qed.
 
+#[universes(template)]
 Record Store : Type :=
 mkStore  {index:positive;contents:Tree}.
 
@@ -191,6 +194,7 @@ Lemma get_empty : forall i, get i empty = PNone.
 intro i; case i; unfold empty,get; simpl;reflexivity.
 Qed.
 
+#[universes(template)]
 Inductive Full : Store -> Type:=
     F_empty : Full empty
   | F_push : forall a S, Full S -> Full (push a S).
--- a/plugins/setoid_ring/Field_theory.v
+++ b/plugins/setoid_ring/Field_theory.v
@@ -730,6 +730,7 @@ Qed.
 
 (* The input: syntax of a field expression *)
 
+#[universes(template)]
 Inductive FExpr : Type :=
  | FEO : FExpr
  | FEI : FExpr
@@ -762,6 +763,7 @@ Strategy expand [FEeval].
 
 (* The result of the normalisation *)
 
+#[universes(template)]
 Record linear : Type := mk_linear {
    num : PExpr C;
    denum : PExpr C;
@@ -944,6 +946,7 @@ induction e2; intros p1 p2;
   now rewrite <- PEpow_mul_r.
 Qed.
 
+#[universes(template)]
 Record rsplit : Type := mk_rsplit {
    rsplit_left : PExpr C;
    rsplit_common : PExpr C;
--- a/plugins/setoid_ring/InitialRing.v
+++ b/plugins/setoid_ring/InitialRing.v
@@ -740,6 +740,7 @@ Ltac abstract_ring_morphism set ext rspec :=
   | _ => fail 1 "bad ring structure"
   end.
 
+#[universes(template)]
 Record hypo : Type := mkhypo {
    hypo_type : Type;
    hypo_proof : hypo_type
--- a/plugins/setoid_ring/Ncring_polynom.v
+++ b/plugins/setoid_ring/Ncring_polynom.v
@@ -32,6 +32,7 @@ Variable phiCR_comm: forall (c:C)(x:R), x * [c] == [c] * x.
    with coefficients in C :
  *)
 
+#[universes(template)]
  Inductive Pol : Type :=
   | Pc : C -> Pol
   | PX : Pol -> positive -> positive -> Pol -> Pol. 
@@ -43,6 +44,7 @@ Definition cI:C . exact ring1. Defined.
  Definition P1 := Pc 1.
 
 Variable Ceqb:C->C->bool.
+#[universes(template)]
 Class Equalityb (A : Type):= {equalityb : A -> A -> bool}.
 Notation "x =? y" := (equalityb x y) (at level 70, no associativity).
 Variable Ceqb_eq: forall x y:C, Ceqb x y = true -> (x == y).
--- a/plugins/setoid_ring/Ring_polynom.v
+++ b/plugins/setoid_ring/Ring_polynom.v
@@ -121,6 +121,7 @@ Section MakeRingPol.
     - (Pinj i (Pc c)) is (Pc c)
  *)
 
+ #[universes(template)]
  Inductive Pol : Type :=
   | Pc : C -> Pol
   | Pinj : positive -> Pol -> Pol
@@ -908,6 +909,7 @@ Section MakeRingPol.
 
  (** Definition of polynomial expressions *)
 
+ #[universes(template)]
  Inductive PExpr : Type :=
   | PEO : PExpr
   | PEI : PExpr            
--- a/plugins/setoid_ring/Ring_theory.v
+++ b/plugins/setoid_ring/Ring_theory.v
@@ -540,6 +540,7 @@ Section AddRing.
  Variable (rO rI : R) (radd rmul rsub: R->R->R) (ropp : R -> R).
  Variable req : R -> R -> Prop. *)
 
+#[universes(template)]
 Inductive ring_kind : Type :=
 | Abstract
 | Computational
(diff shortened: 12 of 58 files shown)


SkySkimmer:
>I'm not sure that we actually want this.

Essentially, I do think we want to move away from silent template polymorphism but when I made this PR it seemed somehow heavy (and I abandoned annotating the non-output test suite files, which trigger the warning about 300-400 times).

ppedrot (approved):
I like the concept, and the code looks OK.

gares (commented):
I like the idea.
I think the name of the attribute could be better.
Either `universes-template-polymorphic` or `universes(template)` if `universes(polymorphic)` enables the other kind of polymorphism.
Just `template` is too ambiguous (template coq?).
Finally, I still hope template polymorphism goes away, hence `universes(template)` seems a better option.

gares:
FTR, the `polymorphic` attribute is also badly named, since polymorphism means another thing to most of us, so maybe this is the right occasion to introduce an attribute that describes how *universes* are interpreted.

SkySkimmer:
Now using `#[universes(template)]`

herbelin:
> Now using #[universes(template)]

In passing, don't you want to have a true discussion on the syntax of attributes? No need to move Coq to jargon-like syntaxes. At least asking users if they really prefer this kind of syntax over, say, a minima, `#[Template Polymophic]`?

maximedenes:
> In passing, don't you want to have a true discussion on the syntax of attributes? No need to move Coq to jargon-like syntaxes. At least asking users if they really prefer this kind of syntax over, say, a minima, `#[Template Polymophic]`?

Of course, we can discuss. I personally find "template polymorphism" more jargon than parentheses...

Zimmi48:
I think it is important that the discussion happens, otherwise it will always feel like it forced its way through. Here is not the place to have the discussion of course. We should probably switch back to #8027.

maximedenes:
> I think it is important that the discussion happens

Sure, I really wasn't trying to suggest otherwise. My point was simply that for the scope of this PR, the terminology and concepts we use (template polymorphism) seems way more intimidating to me than the syntax.

herbelin:
My remark was on the attribute syntax. On the PR, rephrasing @skyskimmer himself, I'm not sure that we actually want this.

Even if we want to forbid `Prop` <= `Set`, I don't understand well why (the accordingly intimidating) so-called template polymorphism could not be transparently simulated from universe polymorphism and cumulativity?

SkySkimmer:
>I don't understand well why (the accordingly intimidating) so-called template polymorphism could not be transparently simulated from universe polymorphism and cumulativity?

The behaviour wrt module subtyping is especially different. I'm sure there are other differences.

herbelin:
> My point was simply that for the scope of this PR, the terminology and concepts we use (template polymorphism) seems way more intimidating to me than the syntax.

I hope that you will agree that it may be worth to be honest and not to mix the two questions though...

mattam82:
@herbelin @SkySkimmer on the topic of template-polymorphism versus poly+cumulativity, yes we should be able now to emulate it. I think we should rather try to turn template-polymorphic inductives into polymorphic cumulative ones for a start and see where this goes. This is not my top priority right now but anyone is welcome to try. If that goes well, we can think of deprecating. What is the issue with module subtyping you are refering to @SkySkimmer ?

SkySkimmer:
Subtyping universe polymorphic stuff is generally a pain as instances need to match, this simply does not happen with template polymorphism.

mattam82:
But it's rare enough to use a non-trivial subtyping of inductives no? Does it happen in the stdlib for example? Note that I'm proposing to make only the Inductives polymorphic, not the associated definitions yet.

mattam82:
@herbelin it might be necessary, however the internal (global) universe will have to be fixed anyway in both cases so I'm not sure it changes much.

herbelin:
> the internal (global) universe will have to be fixed anyway 

I'm sorry, I don't get what you mean by "fixed"?

mattam82:
I mean in unifications of T@{outer inner} = T@{outer' inner'}, the inner universes will have to be equal, while in the template polymorphic mode you'd have a single representative of the inner universe.

JasonGross:
> I think we should rather try to turn template-polymorphic inductives into polymorphic cumulative ones for a start and see where this goes

I frequently take advantage of the fact that `sig` lands in `Prop` when both of it's arguments do, but unlike `ex` can be eliminated when the goal lives in `Type`.  This would be lost with universe polymorphism, right?

mattam82:
@JasonGross that's right indeed, we'd need to allow Prop to instantiate universe binders for this. Why don't you use a proper sigma type in Prop instead?

JasonGross:
> Why don't you use a proper sigma type in Prop instead?

Mostly convenience.  Because the standard library doesn't provide one (let alone have nice notation for it), and I have a decent amount of automation infrastructure around the two existing sigma types.

ppedrot:
It looks like the comments of this PR have been hijacked to a mildly unrelated topic, which is emulation of template polymorphism using cumulativity. Notwithstanding this problem, what is the future of this PR? Do we still want the explicit attribute or what?

mattam82:
Sorry for the hijack. In the current situation, template is not subsumed by cumulative polymorphism, so I'm not sure we want to activate the warning by default for users just yet but certainly the functionality is useful.

SkySkimmer:
I think we do want to move towards making use of template polymorphism explicit, even if we're not ready to deprecate template polymorphism itself.

mattam82:
Hmm, shouldn't there be a notemplate attribute as well here? Users should be able to avoid it, at least to prevent breaking extraction.

SkySkimmer:
notemplate already exists.

mattam82:
Ok, my bad! I guess I'd like to have opinions of users here, given that the warning will impact every definition of an inductive in Type in the wild, I guess that will come already from users of master when me merge. One last thing, should the warning be disabled in compat/Coq89.v?

SkySkimmer:
The warning is meant to be used to transition away from automatic 
template polymorphism, so after putting #[template] on the ones where 
it's wanted they can turn off the option.

I don't know if we have any policy on warnings wrt compat settings.

Gaëtan Gilbert

mattam82:
@herbelin Currently we cannot implement template using cumulative polymorphism, so this is only about making explicit when template polymorphism is used. The idea is that users should now be able to specify how their inductive is interpreted w.r.t. universes, and that information appears in the source. I think making this explicit is a plus. With respect to future plans, we'd like template to be replaced with cumulative polymorphic inductive types in the future, and in that case, it shall only be a matter of moving from universes(template) to universes(polymorphic, cumulative). I don't think the default should become polymorphic, cumIulative though, as that would change the meaning of existing definitions which use global universes. Similarly, if we decide to implement your suggestion to emulate template with a restricted kind of polymorphic, cumulative (being polymorphic only on the universes in the parameters that appear in the conclusion) then we'll still want to distinguish those two modes.

In short, this is about being explicit rather than implicit. And quoting you, I feel the consensus is that no form of polymorphism should be the default indeed, to avoid surprises. The proposed warning acts as a kind of `Set Template Polymorphic Inductives` option, when explicitly disabled.

herbelin:
@mattam82: Thanks for clarifying your plans, that seems good.

SkySkimmer:
So is this getting merged?

ppedrot:
I guess this is essentially waiting for a green light from @herbelin.

herbelin:
As far as the objective announced by @mattam82 in his comment above is to have no form of polymorphism at all on by default, a warning (and the PR) is ok to me.

ppedrot:
@SkySkimmer Can you rebase to run the CI on a fresh master?

ppedrot:
AAC-tactics is broken by this PR, this needs an overlay.

SkySkimmer:
Backward compatible overlay waiting for merge at https://github.com/coq-community/aac-tactics/pull/27

ppedrot:
@SkySkimmer can you rebase to get a fresh CI run?

SkySkimmer:
Rebase again and again...

gares (approved):

Zimmi48:
I too have realized after the fact that I had a few problems with this PR. The first one is that things are seriously badly documented. The second one is that it should be possible to use the option to say I want template polymorphism everywhere and I don't want the warning (currently the solution is to silence the warning). I've tried to come up with a solution with a non-boolean option: people could do `Set Auto Template Polymorphism "enabled"`, `Set Auto Template Polymorphism "disabled"`, or `Unset Auto Template Polymorphism` (default, which emits a warning if there is no attribute). However, I've hit strange failures...

herbelin:
Coming back to this PR, wouldn't there be a way to make the warning more precise? For instance, I get a warning `Automatically declaring True as template polymorphic` if I do
```coq
Inductive True := I.
```
But the warning seems to come here only from putting `True` in `Prop`, not from using any kind of polymorphism at all.

As far as I understand the objectives, it shall be either (1) to ask users to write explicitly
```coq
Inductive True : Prop := I.
```
to have `True` in `Prop`, or (2) to write:
```coq
Inductive unit : Type := tt.
```
or
```coq
Inductive unit : Set := tt.
```
to have `unit` not formally put in `Prop`.

I guess the warning could be something like `Automatically declaring True in Prop but the default shall change. Use explicitly Prop or Set to avoid the warning.`. Then, if the objective is (1) to have the warning whenever `Type` or no explicit type is given, and, if the objective is (2), to have it displayed only when `Type` is given.

What do you think?

SkySkimmer:
>Coming back to this PR, wouldn't there be a way to make the warning more precise?

Indeed this is a tricky case as thinking about it led me to #9294 so I'm not sure if we can detect it. Maybe if we do like with the "could not define as primitive record" warning where we look at what the kernel inferred.
We would also want to avoid warning when there is literally no difference between template and monomorphic, eg `Inductive Foo := foo : Type -> Foo.`, again this may be difficult to detect though.

herbelin:
> Indeed this is a tricky case as thinking about it led me to #9294 so I'm not sure if we can detect it. Maybe if we do like with the "could not define as primitive record" warning where we look at what the kernel inferred.

It seems to me that we should first clarify what is the expected definition of template polymorphism. Is the discriminating feature of template polymorphism only that level parameters are instantiated by the default levels inferred from typing the parameters of the inductive type rather than using fresh levels together with appropriate constraints (and cumulativity to tell that the exact level of these fresh universes do not matter at the end)?

Another remark: it is surprising that we are allowed to define
```coq
Inductive Foo@{i} (A:Type@{i}) : Type := foo : A -> Foo A.
```
but then, `Check Foo@{Set}` fails with `Universe instance should have length 0`.

If the difference between template polymorphism vs full polymorphism is about automatic inference of "canonical" values for level parameters vs using fresh variables by default when no `@{}` is used, why to allow an explicit `@{}` only for full polymorphism?

SkySkimmer:
>Is the discriminating feature of template polymorphism

There's no such thing, what there is is a discriminating implementation.

herbelin:
> ... rather than using fresh levels together with appropriate constraints ...

Actually, minimization in "full" polymorphism is already minimizing the fresh levels as template polymorphism is doing. Let's for instance consider
```
Polymorphic Inductive Foo (A : Type) : Type :=  foo : A -> Foo A.
Set Printing Universes.
Check Foo nat.
(* Foo@{Set} nat : Set *)
```
where we can see that minimization has indeed played the rôle of template polymorphism (with the additional constraint to be at least >= `Set`).

I had a quick look to the implementation. Apparently, the first step of template polymorphism (i.e. the computation of a `max` of constraints over the levels of the conclusion of the parameters plus impact of isolated occurrences of `Type` in the constructors) is still in used even for full polymorphism (function `ComInductive.inductive_levels`).

It is the second step of template polymorphism (i.e. the instantiation of the levels of the conclusion by inferred levels) which is (more or less?) performed by minimization.

> There's no such thing, what there is is a discriminating implementation.

I don't know well the specification of the minimization algorithm but, would it then makes sense to replace the second step of template polymorphism (i.e. `type_of_inductive_knowing_parameters`) by a variant of minimization which accepts on demand to minimize a level without lower constraints to `Prop`.

The latter would fix #9294. (Alternatively, a solution to fix #9294 would be to take into account the graph of constraints in `type_of_inductive_knowing_parameters`).

Side question: How complicated would it be to display the graph of constraints of an inductive type when doing `Print` in `Set Printing Universes` mode, as we do for `Definition`?

Back to minimization, if `type_of_inductive_knowing_parameters` could be replaced by minimization, one would not need anymore to have a specific warning for implicit template polymorphism. The relevant warnings would be:
- a warning for implicit use of polymorphism (since there would be only one kind of polymorphism)
- a warning for allowing minimization to minimize to `Prop`
- a warning for restricting polymorphism to only parameters.

(Then, "template" would be a short-hand for standard polymorphism with polymorphism restricted to the parameters of the inductive type and minimization to `Prop` allowed.)

Again, I don't know well enough minimization to understand if the above can be made precise so that template can be subsumed by minimization, but assuming that minimization is made formal enough and that this becomes possible, what would be missing to reduce the concept of template polymorphism to the one of regular polymorphism and to get rid of `type_of_inductive_knowing_parameters`?

SkySkimmer:
>Side question: How complicated would it be to display the graph of constraints of an inductive type when doing Print in Set Printing Universes mode, as we do for Definition?

Looks like some weird codepaths there, About prints constraints for any polymorphic thing whereas Print prints constraints for any definition regardless of polymorphism.

SkySkimmer:
>what would be missing to reduce the concept of template polymorphism to the one of regular polymorphism

Instantiation by Prop mostly I think

herbelin:
> Instantiation by Prop mostly I think

Not technically difficult then. Could we make a program of what the next steps are?

JasonGross:
As a moderately heavy user of both universe polymorphism and template polymorphism (moreso in the past when I was doing category theory than now), I will chime in to say that in my mental model, the differences between universe polymorphism and template polymorphism are:
- template polymorphism is ~broken for partially applied inductives; template-polymorphic inductives should always be in eta-long form (whereas universe polymorphic inductives do not have this restriction)
- template polymorphism allows things like `prod True True : Prop`
- template polymorphism is only polymorphic over the parameters of an inductive type, not also the fields

(Question: Does `constr_eq` consider different universe-instantiations of polymorphic `prod` to be the same?  Different universe-instantiations of template-polymorphic `prod` to be the same?)

I am not currently aware of any developments that break when replacing template-polymorphic inductives with full universe-polymorphic inductives, unless they take advantage of `Prop`-instantiation.

So my current view is that we should definitely do your Option 1 for Minimization à la template to `Prop`, and that there should be a compatibility setting (even if only for the compat files) for automatically turning on polymorphism for inductives, even if it is off for everything else.  I am indifferent to whether or not there is an option for making an inductive polymorphic over only its parameters, and not its constructors.

mattam82:
I think your option 1 could work, doing minimization to Prop on levels which have no explicit lower bound. That would mean that flexible universe variables would be allowed to be unified with Prop I suppose. Appart from being careful that we’re not breaking invariants in the uGraph about the lower bounds on universes and removing the restriction that instances cannot contain Prop I guess it’s not hard to implement. I also believe with this done you could see template as a restriction of full polymorphism to the apparent flexible universe variables in the parameters. One thing that might differ is how  `Ind (forall x : A, B x)` is treated, I.e, if the inferred level of the parameter is an algebraic, what happens?

herbelin:
> I think your option 1 could work, doing minimization to Prop on levels which have no explicit lower bound. That would mean that flexible universe variables would be allowed to be unified with Prop I suppose. Appart from being careful that we’re not breaking invariants in the uGraph about the lower bounds on universes and removing the restriction that instances cannot contain Prop I guess it’s not hard to implement. I also believe with this done you could see template as a restriction of full polymorphism to the apparent flexible universe variables in the parameters.

Yes. (But to be clear, I'm just talking about an optional behavior of `Polymorphic`, not the default behavior.)

> One thing that might differ is how Ind (forall x : A, B x) is treated, I.e, if the inferred level of the parameter is an algebraic, what happens?

You mean, say, in:
```coq
Polymorphic Inductive list A := nil | cons : A -> list A -> list A.
Check fun A B => list (A -> B).
(*
fun (A : Type@{Top.16}) (B : Type@{Top.17}) =>
list@{Top.15} (A -> B)
     : Type@{Top.16} -> Type@{Top.17} -> Type@{max(Set,Top.15)}
   {Top.17 Top.16 Top.15} |= Top.16 <= Top.15
                             Top.17 <= Top.15 *)
```
I don't know well minimization, but could it be inferred that `Top.15` can be minimized to `max(Top.16,Top.17)`?

SkySkimmer:
>I don't know well minimization, but could it be inferred that Top.15 can be minimized to max(Top.16,Top.17)?

It can't because it's used in `list@{Top.15}` and in that position algebraics aren't allowed.

mattam82:
Indeed, and lifting that restriction would be quite challenging I think. But it might be very rarely used. In the really relevant case where the application of the inductive to the forall... parameter would fall down in Prop, the product itself would have type Prop.

herbelin:
> It can't because it's used in list@{Top.15} and in that position algebraics aren't allowed.

I see. Then, if not allowed, I guess it is to prevent possible having to deal with equations of the form `u < max(v,w)`?

In template polymorphism, I'd have to check precisely but I don't see how lower constraints on the level of the conclusion of a parameter could arrive (if disallowing the kind of sharing that #9294 uses). Would it be possible to allow algebraics in situations where template polymorphism was allowing them? Or would it break somewhere else?

PR #8490, opened 2018-09-17
Fix printing of raw terms for lambda's without parameters

LasseBlaauwbroek:
When parsing terms, sometimes lambda's without any variables are
inserted. An example of such a term is
```coq
let x := 1 in True
```
This term gets parsed in to the following pseudo-term:
```coq
let x := fun => 1 in True
```
This means that during parsing, a lambda without parameters gets inserted.
These lambdas are detected and removed when globalizing terms
from ```Constrexpr.constr_expr``` to ```Glob_term.glob_constr``` in function ```Constrintern.internalize```:
https://github.com/coq/coq/blob/dd84c113a154742dff86328ebc758097e9aac8eb/interp/constrintern.ml#L1869-L1871

This works, but I have a usecase where I have to print terms that have not been internalized.
The printing function assumes that all lambdas have at least one parameter, however, as can be seen from the following two snippets:
https://github.com/coq/coq/blob/dd84c113a154742dff86328ebc758097e9aac8eb/printing/ppconstr.ml#L530-L537
https://github.com/coq/coq/blob/dd84c113a154742dff86328ebc758097e9aac8eb/printing/ppconstr.ml#L391-L398

This pull requests updates the printing function to ignore empty lambdas. Note that there are probably much better solutions for this problem. If so, please let me know.

This change does not affect anything in coq directly because raw terms are never printed, this is only for my usecase (and anyone else's in the future). I ran the tests on the v8.8 branch, which passed. I am currently unable to compile master.

**Kind:** bug fix

### DIFF
--- a/interp/constrintern.ml
+++ b/interp/constrintern.ml
@@ -1866,9 +1866,7 @@ let internalize globalenv env pattern_mode (_, ntnvars as lvar) c =
     | CProdN (bl,c2) ->
         let (env',bl) = List.fold_left intern_local_binder (env,[]) bl in
         expand_binders ?loc mkGProd bl (intern_type env' c2)
-    | CLambdaN ([],c2) ->
-        (* Such a term is built sometimes: it should not change scope *)
-        intern env c2
+    | CLambdaN ([],c2) -> anomaly (Pp.str "the AST is malformed, found lambda without binders")
     | CLambdaN (bl,c2) ->
         let (env',bl) = List.fold_left intern_local_binder (reset_tmp_scope env,[]) bl in
         expand_binders ?loc mkGLambda bl (intern env' c2)
--- a/parsing/g_constr.mlg
+++ b/parsing/g_constr.mlg
@@ -261,7 +261,7 @@ GRAMMAR EXTEND Gram
         { let ty,c1 = match ty, c1 with
           | (_,None), { CAst.v = CCast(c, CastConv t) } -> (Loc.tag ?loc:(constr_loc t) @@ Some t), c (* Tolerance, see G_vernac.def_body *)
           | _, _ -> ty, c1 in
-          CAst.make ~loc @@ CLetIn(id,mkCLambdaN ?loc:(constr_loc c1) bl c1,
+          CAst.make ~loc @@ CLetIn(id,(if bl = [] then c1 else mkCLambdaN ?loc:(constr_loc c1) bl c1),
                  Option.map (mkCProdN ?loc:(fst ty) bl) (snd ty), c2) }
       | "let"; fx = single_fix; "in"; c = operconstr LEVEL "200" ->
           { let fixp = mk_single_fix fx in
--- a/printing/ppconstr.ml
+++ b/printing/ppconstr.ml
@@ -395,7 +395,7 @@ let tag_var = tag Tag.variable
         kw n ++ pr_binder false pr_c (nal,k,t)
       | (CLocalAssum _ | CLocalPattern _ | CLocalDef _) :: _ as bdl ->
         kw n ++ pr_undelimited_binders sep pr_c bdl
-      | [] -> assert false
+      | [] -> anomaly (Pp.str "the ast is malformed, found lambda without binders")
 
   let pr_binders_gen pr_c sep is_open =
     if is_open then pr_delimited_binders pr_com_at sep pr_c


herbelin:
@LasseBlaauwbroek: Thanks a lot for the careful analysis of the situation. There are indeed several solutions to this issue, see #7867. In #7867, we decided to work at the level of the parser, which would mean to add a `if bl = [] then c1 else` in front of `mkLambdaC` in `parsing/g_constr.mlg` in the right-hand side for parsing `let`. Would you be ok with following this path?

LasseBlaauwbroek:
That looks like it is indeed a better solution than mine. I had not noticed that this was already fixed in master. I think this makes my PR superfluous. However, might I suggest that the following lines are removed since they should not serve a purpose anymore:
https://github.com/coq/coq/blob/dd84c113a154742dff86328ebc758097e9aac8eb/interp/constrintern.ml#L1869-L1871

herbelin:
> I had not noticed that this was already fixed in master. I think this makes my PR superfluous.

This is actually only partially fixed in master. Your example shows a second instance of `mkLambdaC` which we failed to fix in PR #7867 and which remains to be fixed.

> However, might I suggest that the following lines are removed since they should not serve a purpose anymore:
>  | CLambdaN ([],c2) -> 

Yes, I agree and I would suggest to replace the right-hand side with an `assert false`.

LasseBlaauwbroek:
Oke, I made the changes we discussed. Should be better now. There is still one unguarded `mkLambdaC` left, but that one should be fine.

herbelin:
> There is still one unguarded mkLambdaC left, but that one should be fine.

Yes, this one should be fine. Thanks.

@ejgallego: ok for you?

ejgallego:
> @ejgallego: ok for you?

Sure @herbelin !

herbelin:
OK, shall merge tomorrow end of afternoon in Paris.

herbelin:
@LasseBlaauwbroek: I'm sorry, I did not see that the compilation fails (in particular in the test-suite). I suspect that this is because of the presence of `mkCLambdaN` in other parts of the code, like in `def_body` in `vernac/g_vernac.mlg`, but in other places also.

Considering the relatively high number of `mkCLambdaN` around, I would lean towards going back to a fix I proposed in #8286's fix of #7867, but which @ejgallego did not like very much. So, @ejgallego, would you be ok to go back to moving the `if l = [] then c else` things directly in the code of `mkCLambdaN` (the consequence would be that the absence of degenerate `CLambda` is enforced in the `constr_expr` structure)? I would find strange that parsing and printing care about not showing degenerate `CLambda` but that the internal structure allows it. (This would be a bit similar to our forbidding of `App(f,[||])` in `constr`). I'm afraid that allowing degenerate `CLambda` would otherwise be a source of worries for plugins which may forget to care that their code is compatible with the degenerate case (as we ourselves forgot to care for printing). @ejgallego, it is maybe also that I did not understand your sentence "keep the code as it, however document the functions and the special pattern/case for the empty list" [here](https://github.com/coq/coq/pull/8286#pullrequestreview-151591899): in particular was "it" referring to the PR's code or to the master's code, and was "function" referring to `mkCLambdaC` or to some other function?

ejgallego:
> So, @ejgallego, would you be ok to go back to moving the `if l = [] then c else` things directly in the code of `mkCLambdaN` (the consequence would be that the absence of degenerate `CLambda` is enforced in the `constr_expr` structure)?

Well, it seems strange to me that `mkCLambda` would silently try to fix caller that are violating its precondition. Code calling `mkCLambda` with 0 arguments seems broken, don't you think?

I know the `mkCLambda` pattern you propose can be convenient in some functional programming styles, however maybe such variations should be made local, instead of assigning a counter-intuitive behavior to a core AST function.

ejgallego:
Anyways my point of view is not hard in any of the choices, as I understand each proposal seems to have its own advantages and disadvantages, I wonder what do other people think.

ejgallego:
> @ejgallego: Would a change of the name of `mkCLambda` into `mkNonDegenerateCLambda` or `mkCLambdaSmart`, ... makes things less strange.

@herbelin indeed `mkCLambdaSmart` seems like a very good idea and IMHO leaves clear the purpose of the function.

herbelin:
> @herbelin indeed mkCLambdaSmart seems like a very good idea and IMHO leaves clear the purpose of the function.

Since this gives little interest left for a function `mkCLambda` (compare to using `CLambda` directly), I guess one would go for a deprecation of it?

ejgallego:
> Since this gives little interest left for a function `mkCLambda` (compare to using `CLambda` directly), I guess one would go for a deprecation of it?

I dunno, it could be well that in the future we find a use for it. Note that the same could be said then for all the other `mkFoo` functions in constrexpr_ops, couldn't it?

herbelin:
> I dunno, it could be well that in the future we find a use for it. Note that the same could be said then for all the other mkFoo functions in `constrexpr_ops`, couldn't it?

More or less: `mkAppC` (in passing, why `mkAppC` and not `mkCApp`?) is already smart in that it compacts several `App` into one. In addition to `mkCLambdaN` and `mkCProdN`, only `mkCastC` and `mkLetInC` (note again the trailing `C`s) are only applying the constructor w/o doing anything special (each are used exactly once in the archive).

So, certainly a uniformisation to do (but which would go beyond the scope of the PR). Maybe we shouldn't actually mix a needed uniformization/documentation of the `constrexpr_ops` API with the purpose of this PR?

ejgallego:
Indeed I am not sure what to say; it seems to me that what to do here is very subjective. I am open to any approach you deem fit, all my comments were in the category of "soft" suggestions.

LasseBlaauwbroek:
Ah, I was assuming that the failed build was due to unrelated issues. (I can't seem to build master locally at all.) The discussion above is a little beyond my knowledge of the code base, so I will let you hash it out. But I don't think that I will be able to make the required code modifications.

herbelin:
@ejgallego. I think indeed that our discussion went a bit far...

@LasseBlaauwbroek I can then to take over considering how far this discussion went.

@ejgallego: I propose to keep `mkCLambda`, to add `mkCLambdaSmart` and then conclude this PR. OK?

ejgallego:
> @ejgallego: I propose to keep `mkCLambda`, to add `mkCLambdaSmart` and then conclude this PR. OK?

Sounds great to me, thanks!

herbelin:
Continued in #8662. @LasseBlaauwbroek, @ejgallgo, I'm then closing this PR.

LasseBlaauwbroek:
Thank you for the fix. May I suggest that the error checking that is in this PR still be merged? It seems to me that it is still theoretically possible for a malformed AST to be created, so why not build in some defensive code?

herbelin:
OK, I made #8680 for the malformed AST check.

PR #8497, opened 2018-09-18
Use "rm -rf" in "make clean" so .coq-native directories are removed

### LINKED ISSUE #8496: After #8243 .native directories are not removed by make clean
looks like `find .. -delete` kills directories while `-exec rm -f {}` does not, and #8243 switched to the latter.
I guess a `-r` should be added to `rm`.

jfehrle:
**Kind:** infrastructure.

Fixes / closes #8496


gares:
I think .native is the only directory. I would have an extra invocation of find just for that, and only there use `-rf`. It would make the patch way more obvious, I believe.

gares:
Oh, thanks for the quick fix!

jfehrle:
Updated as you suggested.

I didn't see your comment until just now because many of the Github notification emails were ending up in my spam folder.

gares:
@maximedenes I'm on VAC today and at a conference next week...
This PR is ow a trivial fix for an issue you found. Feel free to merge.

gares (approved):

PR #8499, opened 2018-09-18
[api] Deprecate constructors of deprecated datatypes.

### LINKED ISSUE #8498: [api] Deprecation of datatype constructors is incomplete.
Cleanup and modularization commits deprecated a few types that were either aliases or have been moved to a different place in the hierarchy. Unfortunately, deprecating the type doesn't deprecate its constructors. That means that plugins cannot just fix all the deprecation warnings and still work when the types are removed.

Thus, deprecation warning for most constructors whose types are deprecated must be added before the release.

@gasche , is there a way to place an attribute in a `type = ...` declaration such that the type and its constructors will be deprecated simultaneously?

ejgallego:
When deprecating some type alias [due to code refactoring] we forgot
to deprecate the constructors too. Closes #8498.

### DIFF
--- a/plugins/funind/indfun.ml
+++ b/plugins/funind/indfun.ml
@@ -98,7 +98,7 @@ let functional_induction with_clean c princl pat =
       List.map2
         (fun c pat ->
           ((None,
-            Ltac_plugin.Tacexpr.ElimOnConstr (fun env sigma -> (sigma,(c,NoBindings)))),
+            Tactics.ElimOnConstr (fun env sigma -> (sigma,(c,NoBindings)))),
            (None,pat),
            None))
         (args@c_list)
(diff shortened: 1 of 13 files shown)


ppedrot (approved):

ejgallego:
@ppedrot you want to be the assignee for this?

ejgallego:
Ping @ppedrot , it would be good to have this soonish in as IMHO it should absolutely go in the beta. Merci !

ppedrot:
I was waiting for others to comment, but the change is trivial enough that the PR can be merged soon. Let's say tomorrow if nobody objects.

ejgallego:
Yeah I guess the PR has not a lot to be said about, if people is unhappy with the current placement we should rather handle that in a separate issue.

ejgallego:
Thanks @ppedrot

PR #8503, opened 2018-09-19
[dune] [doc] Support for installing the reference manual with Dune.

ejgallego:
This is a preliminary attempt to generate install rules for the Sphinx-generated reference manual. Almost for sure it needs support from Dune usptream.


gares:
> Install? Should this live in a `coq-doc.opam` package?

I don't know if it was fixed in opam 2.0, but in opam 1.2 you cannot have multiple packages being built out of the same source tree. For coq and coqide it was still ok, since you need only one file of Coq to compile coqide, recompiling it is ok. For sphinx you need coqtop and the standard library in order to run the coq snippets, so having a separate package requires you to recompile almost everything.

ejgallego:
> I'm sorry, it may be enough to make the doc package depend on coq. Forgive the noise.

Indeed. The dependencies would have to be refined tho as sphinx should really depend on `coq.stdlib.prelude` or something like that for incremental builds to work properly in compositional-dev/release mode.

gares:
It depends on more than Prelude, I'm sure I've added snippets using `Bool` or `ssreflect`.

ejgallego:
Yeah that's item 3 on the to-do list.

ejgallego:
@coq/doc-maintainers , we will likely need some Dune 1.3 features to implement this properly, but what do you think of the current rule:
https://github.com/coq/coq/blob/267870030e306779b39a7bdbb5ed8c7a97442491/dune#L13-L28

I've seen that there seems to be some issues with the makefiles, so I'd rather be sure about what I am trying to do here.

Zimmi48:
The dependency on just the Prelude is certainly wrong since as of today the refman depends on the whole stdlib.
Regarding `_static`, we should look into renaming this directory. I don't know if that's supported by Sphinx.
As for the destination directory, why not use the base `_build` directory? It would be simpler for the install rules.

ejgallego:
> The dependency on just the Prelude is certainly wrong since as of today the refman depends on the whole stdlib.

Why does it need the whole stdlib?

> Regarding `_static`, we should look into renaming this directory. I don't know if that's supported by Sphinx.

That should be fixed in the Dune side, IMO.

> As for the destination directory, why not use the base `_build` directory? It would be simpler for the install rules.

I am using the same target that `Makefile.doc` uses, I'd rather try to keep them in sync I guess.

Zimmi48:
>Why does it need the whole stdlib?

Because people load parts of stdlib in the examples. I'm not defending this, just stating a fact.

>I am using the same target that Makefile.doc uses, I'd rather try to keep them in sync I guess.

Not sure I see the point. Dune introduced a `_build` directory, so we might as well use it consistently.

anton-trunov:
> Because people load parts of stdlib in the examples. I'm not defending this, just stating a fact.

And I guess it's sort of unavoidable, if we want examples for typeclasses, `Program` / `Function` and other advanced features.

ejgallego:
> Because people load parts of stdlib in the examples. I'm not defending this, just stating a fact.

I see, what I mean is "why do we need the _whole_ stdlib"? Instead, I'd like to have more granular dependencies here, that is to say, on the modules that are actually used.

Dune provides reasonable support for having an autogenerated list of dependencies, using promotion. Thus, we could have `refman-modules.inc` which would contain the modules actually used by the refman and `(include)`'d in by the build rule, and Sphinx would output this file for each build. Dune would complain if the list changes thanks to promotion, requiring to check the new version in.

> Not sure I see the point. Dune introduced a `_build` directory, so we might as well use it consistently.

Umm, so what's your proposed path? Note that the build cannot see below its own context, and in fact this will be forced by sandboxing.

ejgallego:
> I was thinking the result of the build could be stored in `_build/default/share/doc/coq/sphinx`

`_build/default` location is not so important, once we install the files I do agree that they should somewhere such as `_build/install/default/share/doc/refman`

ejgallego:
So I think we can move forward with this, but I do have a question, it seems to me that "OPAM-wise", this should live in a different package as building the docs is quite far from trivial.

So @coq/doc-maintainers , would you like to have a `coq-doc` or `coq-documentation` package?

Another possibility is to distribute an already-built version of the manual in the tar.gz. This is possible to do automatically if we move to `dune-release` in the future.

Zimmi48:
I think yes, we want to have `coq-doc` be a separate package. Maybe call it `coq-refman` actually, as long as it only includes the reference manual. Unless you want to add the stdlib doc as well in this.

cc @RalfJung, cc #4855.

ejgallego:
> I think yes, we want to have `coq-doc` be a separate package. Maybe call it `coq-refman` actually, as long as it only includes the reference manual. Unless you want to add the stdlib doc as well in this.

`coq-refman` package added. Tho proper support will have to wait for a new Dune feature [namely, support for directory targets https://github.com/ocaml/dune/issues/1377], but I think we could live with that.

I think that the documentation of the stdlib should be shipped in its own package, as it depends on `coqdoc` not a big deal [as opposed to this one].

So far this PR is mostly ready [apart from a strange problem with the `coqdoc.css` path], all that it misses is the install target.

Zimmi48:
>I think that the documentation of the stdlib should be shipped in its own package, as it depends on coqdoc not a big deal [as opposed to this one].

This is supposed to depend on coqdoc as well (for the colors of the input to coqtop).

coq-refman.opam
```diff
@@ -0,0 +1,22 @@
+opam-version: "1.2"
+maintainer: "The Coq development team <coqdev@inria.fr>"
+authors: "The Coq development team, INRIA, CNRS, and contributors."
+homepage: "https://coq.inria.fr/"
+bug-reports: "https://github.com/coq/coq/issues"
+dev-repo: "https://github.com/coq/coq.git"
+license: "LGPL-2.1"
+
+available: [ ocaml-version >= "4.02.3" ]
```
Zimmi48: This line needs to be fixed.

dune
```diff
@@ -39,3 +39,35 @@
...
+  ; %{bin:coqdoc}
+  ; Must find a solution for the _static folder
+  ; doc/sphinx/_static/
+  (package coq)
+  (source_tree doc/sphinx)
+  (source_tree doc/tools))
+ (action (progn
+          ; This will force us to build coqdoc for now
+          (run coqdoc)
+          (run sphinx-build -j4 -b html -d doc/sphinx/_build/doctrees doc/sphinx doc/sphinx/_build/html))))
```
Zimmi48: Looking at this line, it gives the impression that the PDF format of the manual is not built. Can this be fixed easily? Can we get the same semantics for the refman targets than in `Makefile.doc` (with `refman` being the general target, and `refman-pdf` and `refman-html` being one format only)?

Zimmi48 (commented):
If we want to be able to get rid of the legacy makefiles, we need also a way to build the doc for the standard library (but this can be done in another PR).

Zimmi48:
Is the "needs: progress" label really still relevant here?

dune
ejgallego: Sure, we can, depending on the timeline we can use a more advanced solution or we can keep a bit of makefile for this purpose. Dune still don't supports well rules that generate a whole dir.

ejgallego:
> Is the "needs: progress" label really still relevant here?

Yup, the doc is only build, but not installed in the proper package. I need either to find a minute to fix it or maybe upstream adds support for it natively.

Zimmi48:
@ejgallego I would very much like this to get finished because it is IIRC the only thing blocking the use of Dune in the Nix package as defined by `default.nix`.

ejgallego:
> @ejgallego I would very much like this to get finished because it is IIRC the only thing blocking the use of Dune in the Nix package as defined by `default.nix`.

Ok I'll raise priority. I tried a few approaches but I didn't manage to get anything that I was satisfied with, it seems we'll have to wait for the Dune feature I'm afraid.

ejgallego:
Building is fine, however it is the install step that fails. Of course if we do the install by hand we could add this.

Zimmi48:
I would also be OK with a test-suite style solution: i.e. use `Makefile.doc` but ensures it can work in combination with a dune build.

ejgallego:
The thing is that I'd like to generate the `coq-refman.install` file so things will work out of the box for package managers. I am also willing to fix this, because with the windows port is working well now, this is almost the last roadblocker indeed to have a complete Dune-based build.

cpitclaudel:
Thanks for your work on this Emilio. Let me know if there are bits I can help with (I don't know anything about Dune, but I have some experience with Sphinx)

Zimmi48:
Dune 1.6.0 has been released. Does it contain what was missing to finish this PR?

ejgallego:
> Dune 1.6.0 has been released. Does it contain what was missing to finish this PR?

In principle yes, but in experimental form; I'd like to discuss about this a bit more with the Dune devs; that being said we shall push the part of this PR that is ready I think.

ejgallego:
> Great, because given my focus on documentation. Not having this in has been a major drawback for me.

Ok, raising the priority. Note that we could benefit from slightly faster builds if there was a way to output the `.v` dependencies from the `.rst` file. Of course that is very low priority now but just letting the doc maintainers now [the sphinx Coq addon could indeed run `coqdep` on the examples at some point.

ejgallego:
This PR is stale and this approach won't work like it is now. We should either:

- wait for upstream to support install rules for dir targets
- use a small makefile / postprocessing script to generate the refman package.

Alizter:
@ejgallego We now have dir-targets in dune.

PR #8504, opened 2018-09-19
Allow successive attributes #[foo] #[bar]

SkySkimmer:
Since we're based on the rust syntax we should do it this way barring a reason not to (I wasn't watching super closely so please speak up if there was one).

### DIFF
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -522,7 +522,7 @@ The Vernacular
 ==============
 
 .. productionlist:: coq
-   decorated-sentence : [`decoration`] `sentence`
+   decorated-sentence : [ `decoration` … `decoration` ] `sentence`
    sentence           : `assumption`
                       : | `definition`
                       : | `inductive`
@@ -1438,7 +1438,7 @@ Attributes
 
 Any vernacular command can be decorated with a list of attributes, enclosed
 between ``#[`` (hash and opening square bracket) and ``]`` (closing square bracket)
-and separated by commas ``,``.
+and separated by commas ``,``. Multiple space-separated blocks may be provided.
 
 Each attribute has a name (an identifier) and may have a value.
 A value is either a :token:`string` (in which case it is specified with an equal ``=`` sign),
--- a/vernac/g_vernac.mlg
+++ b/vernac/g_vernac.mlg
@@ -83,11 +83,10 @@ GRAMMAR EXTEND Gram
     ]
   ;
   decorated_vernac:
-    [ [ a = attributes ; fv = vernac -> { let (f, v) = fv in (List.append a f, v) }
-      | fv = vernac -> { fv } ]
-    ]
+    [ [ a = LIST0 quoted_attributes ; fv = vernac ->
+        { let (f, v) = fv in (List.append (List.flatten a) f, v) } ] ]
   ;
-  attributes:
+  quoted_attributes:
     [ [ "#[" ; a = attribute_list ; "]" -> { a } ]
     ]
   ;
--- a/test-suite/success/attribute-syntax.v
+++ b/test-suite/success/attribute-syntax.v
@@ -1,4 +1,4 @@
-From Coq Require Program.
+From Coq Require Program.Wf.
 
 Section Scope.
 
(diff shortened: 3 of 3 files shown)


vbgl:
The printer (`vernac/ppvernac.ml`) should probably be updated accordingly.

herbelin:
> Since we're based on the rust syntax

I still could not understand if the current implementation was using rust syntax in order to avoid thinking at another syntax, or for intrinsic reasons. If I remember correctly, I saw noone defending Rust syntax. I saw several voices defending a well-parenthesized simple syntax. What I'm afraid of, but I don't know how is this view shared, is that we make an unsatisfactory choice which engages us for decades only by lazyness and facility.

gares:
Funnily enough Rust manual says it comes from C#, and there I read:

> For  instance, the attribute specifications  [A][B] ,  [B][A] ,  [A, B], and [B, A] are equivalent

See [page 394](http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-334.pdf)

Hence, to me, the old and the new syntax should be supported, with the semantics above.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -524,7 +524,7 @@ The Vernacular
    proof              : Proof . … Qed .
                       : | Proof . … Defined .
                       : | Proof . … Admitted .
-   decoration : #[ `attributes` ]
+   decoration : #[ `attribute` ]
    attributes : [`attribute`, … , `attribute`]
```
gares: The `attributes` non terminal has to go, unless we decide to keep both syntax `#[a] #[b]` and `#[a, b]` in which case it has to stay. I prefer the latter option.

ejgallego:
> If I'm not mistaken @ejgallego was defending that attributes should look (and behave) more closely to OCaml ones, so he is also defending borrowing a syntax from another language, albeit not the same one.

Indeed I think based on heritage and programming style, I'd rather become closer to OCaml than to Rust. I am also very sympathetic to Hugo's views, however I am not sure I get a complete view of regular the proposal is. [regular as in parsing]

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1409,9 +1409,8 @@ using the keyword :cmd:`Qed`.
 Attributes
 -----------
 
-Any vernacular command can be decorated with a list of attributes, enclosed
-between ``#[`` (hash and opening square bracket) and ``]`` (closing square bracket)
-and separated by commas ``,``.
+Any vernacular command can be decorated with a list of attributes, each enclosed
+between ``#[`` (hash and opening square bracket) and ``]`` (closing square bracket).
```
Zimmi48: Should you add "or separated by commas inside a single `#[` `]` block for the sake of completeness?

SkySkimmer:
@vbgl 
>The printer (vernac/ppvernac.ml) should probably be updated accordingly.

The new version of this PR doesn't remove the old syntax. We can either keep the structure of how attributes while parsing (ie change `_ list` to `_ list list`) in order to be able to print as it was input, or we pick one way to print everything (current version of the PR, printing `#[foo] #[bar] ...` as `#[foo, bar] ...`).

vbgl:
It’s fine then not to update the printer: no need to remember the precise input. `#[a] #[b]` should be equivalent to `#[a, b]`.

vernac/g_vernac.mlg
```diff
@@ -83,11 +83,11 @@ GRAMMAR EXTEND Gram
     ]
   ;
   decorated_vernac:
-    [ [ a = attributes ; fv = vernac -> { let (f, v) = fv in (List.append a f, v) }
+    [ [ a = LIST0 quoted_attributes ; fv = vernac -> { let (f, v) = fv in (List.append (List.flatten a) f, v) }
       | fv = vernac -> { fv } ]
```
vbgl: I think this second grammar rule is redundant, as the `a` list may be empty.

vbgl (dismissed):

vernac/g_vernac.mlg
SkySkimmer: Changed.

vbgl (approved):

Zimmi48 (approved):

PR #6739, opened 2018-02-12
Tentative fix for #6520: camlcity.org unresponsive makes AppVeyor fail.

Zimmi48:
**Kind:** infrastructure fix

Fixes / closes #6520 

Let's hope that this works! BTW, I don't like the many unsafe HTTP downloads that this script contains. Especially, given that we subsequently sign and distribute the resulting packages.

PR #8505, opened 2018-09-19
Fix Windows builds: OPAM has changed its URL schema.

Zimmi48:
**Kind:** infrastructure fix.

### DIFF
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -878,7 +878,7 @@ function make_num {
 
 function make_findlib {
   make_ocaml
-  if build_prep https://opam.ocaml.org/archives ocamlfind.1.8.0+opam tar.gz 1 ; then
+  if build_prep https://opam.ocaml.org/1.2.2/archives ocamlfind.1.8.0+opam tar.gz 1 ; then
     logn configure ./configure -bindir "$PREFIXOCAML\\bin" -sitelib "$PREFIXOCAML\\libocaml\\site-lib" -config "$PREFIXOCAML\\etc\\findlib.conf"
     # Note: findlib doesn't support -j 8, so don't pass MAKE_OPT
     log2 make all


MSoegtropIMC (approved):
I did some search and concluded that the official URL for this package is
```
http://download.camlcity.org/download/findlib-1.8.0.tar.gz
```
from this page:
```
http://projects.camlcity.org/projects/findlib.html
```
The contents is identical to the opam package, except that the opam package contains two additional files we don't need.

If it is urgent, the new opam URL is fine, otherwise I would prefer to use the URL mentioned above.

Zimmi48:
@MSoegtropIMC The reason to use another one in #6739 is that the site camlcity was frequently down. This change has proved quite efficient until now.

ejgallego:
Indeed; this is quite urgent as the CI is broken; I will take over a bit later if @MSoegtropIMC cannot do the merge today.

ejgallego:
@MSoegtropIMC , note that `camlcity.org` seems to be hosted in some kind of home-server. As @Zimmi48 pointed out, we had very frequent outages in the past. The OPAM archive change is a kind of exceptional event so we should keep using that one.

MSoegtropIMC:
Ok, if there is a good reason to use a secondary source, I am fine with it.

My plan was to do the merge today around 10:30 today, after CI went through, but fore sure if someone is waiting for it, I don't mind if it is done earlier by someone else.

MSoegtropIMC:
@Zimmi48, @maximedenes : one thing which went slightly wrong with this merge: I recently changed my committer email to a git email. I updated my certificates but forgot to do so on my laptop from which I did run this merge, so the signature doesn't match the email and the commit appears as unverified. This is something one might want to check in the merge-pr.sh script. (I could likely quickly do a PR for this).

Zimmi48:
Sure, if you know how to do so.

MSoegtropIMC:
I created PR #8512.

Since this merge is still on top of the history, shall I quickly redo it signed?

PR #8506, opened 2018-09-19
[ssr] use the right environment in ssrpattern (fix #8454)

### LINKED ISSUE #8454: [ssr] ssrpattern failure: variable n unbound
#### Version
8.7.2 and 8.8.1

#### Operating system
Ubuntu 18.04

The following fails with `Anomaly "Evar ?X5 was not declared`.  On an unrelated note, I don't understand this tactic, and there seems to be no documentation about it.

```
Require Import ssrmatching.ssrmatching.
Goal (if true then true else false) = true.
Proof.
  ssrpattern (if true then true else _).
```

gares:


gares:
ping @maximedenes (ci failure is spurious)

maximedenes (approved):

Issue #8492, opened 2018-09-18
DestKO on Scheme Equality


gares:
```coq
Require Import List.
Inductive tree (A : Type) := Node (siblings : list (tree A)).
Scheme Equality for tree.
```

backtrace:
```
Unexpected error during scheme creation: Anomaly "Uncaught exception Constr.DestKO."
Please report at http://coq.inria.fr/bugs/.
frame @ file "ide/xmlprotocol.ml", line 667, characters 29-46
frame @ file "ide/ide_slave.ml", line 406, characters 12-15
frame @ file "ide/ide_slave.ml", line 208, characters 13-28
frame @ file "stm/stm.ml", line 2646, characters 11-49
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "stm/stm.ml", line 2635, characters 4-55
frame @ file "stm/stm.ml", line 2525, characters 4-100
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "stm/stm.ml", line 934, characters 6-10
frame @ file "stm/stm.ml", line 2398, characters 16-43
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "stm/stm.ml", line 1087, characters 12-91
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "vernac/vernacentries.ml", line 2343, characters 4-36
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "lib/flags.ml", line 17, characters 14-17
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "vernac/vernacentries.ml", line 2310, characters 8-666
frame @ file "lib/flags.ml", line 100, characters 20-44
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "lib/flags.ml", line 17, characters 14-17
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "vernac/indschemes.ml", line 142, characters 6-25
frame @ file "vernac/indschemes.ml", line 267, characters 11-70
frame @ file "tactics/ind_tables.ml", line 167, characters 23-34
frame @ file "vernac/auto_ind_decl.ml", line 980, characters 22-193
frame @ file "proofs/pfedit.ml", line 157, characters 4-64
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "proofs/pfedit.ml", line 142, characters 17-23
[...]

PR #8509, opened 2018-09-19
Fix numerous anomalies with Scheme Equality

herbelin:
**Kind:** bug fix

We protect `Scheme Equality` against a few anomalies raised when out of its domain of specification. The case of inductive types with indices seems difficult but the cases of nested or mutual inductive types should be eventually implementable without too much pain (even if maybe with another methodology than using tactics).

This fixes / closes #4859, #2550, #8492, #4612.

PS: I had to decapsulate exceptions out of `TacticFailure` in `indschemes.ml`. I found this strange so I wonder if I did not do something wrong in `auto_ind_decl.ml` at the time of raising the exception. Maybe should I have used another primitive than `raise`??

### DIFF
--- a/test-suite/bugs/closed/4612.v
+++ b/test-suite/bugs/closed/4612.v
@@ -1,0 +1,7 @@
+(* While waiting for support, check at least that it does not raise an anomaly *)
+
+Inductive ctype :=
+| Struct: list ctype -> ctype
+| Bot : ctype.
+
+Fail Scheme Equality for ctype.
--- a/test-suite/bugs/closed/4859.v
+++ b/test-suite/bugs/closed/4859.v
@@ -1,0 +1,7 @@
+(* Not supported but check at least that it does not raise an anomaly *)
+
+Inductive Fin{n : nat} : Set :=
+| F1{i : nat}{e : n = S i}
+| FS{i : nat}(f : @ Fin i){e : n = S i}.
+
+Fail Scheme Equality for Fin.
--- a/vernac/auto_ind_decl.ml
+++ b/vernac/auto_ind_decl.ml
@@ -389,13 +389,9 @@ let do_replace_lb mode lb_scheme_key aavoid narg p q =
     | Const (cst,_) ->
       (* Works in specific situations where the args have to be already declared as a
          Parameter (see example "J" in test file SchemeEquality.v) *)
-      (
-        let mp,dir,lbl = Constant.repr3 cst in
-          mkConst (Constant.make3 mp dir (Label.make (
-          if Int.equal offset 1 then ("eq_"^(Label.to_string lbl))
-                       else ((Label.to_string lbl)^"_lb")
-        )))
-      )
+        let lbl = Label.to_string (Constant.label cst) in
+        let newlbl = if Int.equal offset 1 then ("eq_" ^ lbl) else (lbl ^ "_lb") in
+        mkConst (Constant.change_label cst (Label.make newlbl))
     | _ -> raise (ConstructorWithNonParametricInductiveType (fst hd))
 
   in
@@ -453,11 +449,9 @@ let do_replace_bl mode bl_scheme_key (ind,u as indu) aavoid narg lft rgt =
     | Const (cst,_) ->
       (* Works in specific situations where the args have to be already declared as a
          Parameter (see example "J" in test file SchemeEquality.v) *)
-        let mp,dir,lbl = Constant.repr3 cst in
-          mkConst (Constant.make3 mp dir (Label.make (
-          if Int.equal offset 1 then ("eq_"^(Label.to_string lbl))
-                       else ((Label.to_string lbl)^"_bl")
-        )))
+        let lbl = Label.to_string (Constant.label cst) in
+        let newlbl = if Int.equal offset 1 then ("eq_" ^ lbl) else (lbl ^ "_bl") in
+        mkConst (Constant.change_label cst (Label.make newlbl))
     | _ -> raise (ConstructorWithNonParametricInductiveType (fst hd))
   in
 
--- a/vernac/auto_ind_decl.ml
+++ b/vernac/auto_ind_decl.ml
@@ -200,6 +200,7 @@ let build_beq_scheme mode kn =
         match Constr.kind c with
         | Rel x -> mkRel (x-nlist+ndx), Safe_typing.empty_private_constants
         | Var x ->
+          (* Support for working in a context with "eq_x : x -> x -> bool" *)
           let eid = Id.of_string ("eq_"^(Id.to_string x)) in
           let () =
             try ignore (Environ.lookup_named eid env)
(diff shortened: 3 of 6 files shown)


vernac/indschemes.ml
```diff
@@ -176,6 +177,14 @@ let try_declare_scheme what f internal names kn =
     | NoDecidabilityCoInductive ->
          alarm what internal
            (str "Scheme Equality is only for inductive types.")
+    | DecidabilityIndicesNotSupported ->
+         alarm what internal
+           (str "Inductive types with annotations not supported.")
+    | ConstructorWithNonParametricInductiveType ind ->
+         alarm what internal
+           (strbrk "Unsupported constructor with an argument of type a non-parametric inductive type." ++
```
JasonGross: Grammar: "an argument of type" -> "an argument whose type is"

vernac/auto_ind_decl.ml
```diff
@@ -435,16 +442,18 @@ let do_replace_bl mode bl_scheme_key (ind,u as indu) aavoid narg lft rgt =
                    (str "Var " ++ Id.print s ++ str " seems unknown.")
       )
     in mkVar (find 1)
-  with e when CErrors.noncritical e ->
+(* This code has no reason to work in general for an arbitrary constant *)
+(*
```
JasonGross: Are there any cases that used to be correctly supported with this code that are no longer supported?

test-suite/bugs/closed/4612.v
```diff
@@ -0,0 +1,7 @@
+(* While waiting for a support, check that it at least not raise an anomaly *)
```
JasonGross: Grammar:
- "for a support" -> "for support"
- "at least not" -> "at least does not"

test-suite/bugs/closed/4859.v
```diff
@@ -0,0 +1,7 @@
+(* Not supported but check that it at least not raise an anomaly *)
```
JasonGross: Grammar: "at least not" -> "at least does not"

JasonGross (dismissed):

vernac/auto_ind_decl.ml
herbelin: I just tried to find one, but without success. For instance, it would have been to support an example like:
```coq
Parameter A:Type.
Parameter eq_A:A->A->bool.
Inductive I := C : A -> I.
Scheme Equality for I.
```
but it fails (with an incorrect message) due to other limitations of the code.

I tried the following which had code partially written to support it:
```coq
Section A.
Variable A:Type.
Variable eq_A:A->A->bool.
Inductive I := C : A -> I.
Scheme Equality for I.
End A.
```
But it fails also (with an anomaly).

At the end, I just preferred to add a commit to deactivate the partial support for variables and parameters.

herbelin:
I fixed the English issues. Thanks for checking.

mattam82:
Nested and mutual require at least as much support as parameters. I wonder
if for example the elpi version of this supports them, @gares ?
Le jeu. 20 sept. 2018 à 13:45, Hugo Herbelin <notifications@github.com> a
écrit :

herbelin:
@mattam82: Are you suggesting that scheme generation is yet another topic where communication and coordination between developers should be useful?

vernac/auto_ind_decl.ml
herbelin: Congratulations in finding what is the exact way to make this working as little changes (different types, A_lb missing, A_bl missing) provokes an anomaly.

Would the following not be good enough for you:
```coq
Inductive I (A:Type) := C : A -> I A.
Scheme Equality for I.
```

JasonGross: I think I originally found it by looking at the proof term that `Scheme Equality` generates, and the types of the lemmas that it generates, and mimicking that.
I'd adjust to this not working if I have to, but sometimes it's nice to be able to collect the arguments together in a section where I'm doing many things.  Also, with the new changes to type inference of inductive parameters (#7254), where the suggested fix is to make the parameters context variables, it seems kind-of annoying to have to open the section to define the inductive, close the section to define the equality scheme, and then open a new section if I want a bunch of lemmas to share the same parameters.

JasonGross: Also, I can't seem to get the anomalies you're talking about with 8.8.  I just get "Error: Unexpected error during scheme creation: Attempt to save an incomplete proof" when I change the types so that things do not work.  (And there are a bunch of small changes that work, e.g., adding extra hypotheses of type `True`, using `true = eq_A x y` rather than `eq_A x y = true`, etc.  However, I do get an anomaly if I change `forall x y, x = y -> eq_A x y = true` to end with `eq_A x x = true`.)

gares:
@mattam82 as of today elpi has no support for mutual inductive/fix. My impression is that mutual inductive play the role of a workaround. Most of the times one would like to just use a container, but given that the positivity checker is weak one defines a mutual inductive type (duplicating the container).

Another way to put it: I don't like the theory of today on that specific detail. If tomorrow we have more induction/recursion then it should be it (telescopes are easier than n-ary binders). For now, then best would be to fix the positivity checker and stay with single inductive/recursive types.

herbelin:
> My impression is that mutual inductive play the role of a workaround.

> For now, then best would be to fix the positivity checker and stay with single inductive/recursive types.

Are you talking about mutual inductive types in the Calculus of Inductive Constructions? If yes, I was thinking the opposite. Mutual inductive types are natural in many situations of formalizing data structures or predicates (starting with `odd`/`even`, `tree`/`forest`, `value`/`neutral`, etc.).

Or maybe is it that I just don't understand in which sense you are saying that `mutual inductive play the role of a workaround`, so what do you mean by `workaround`?

mattam82:
It is true that in many cases one uses mutual instead of nested types but
that’s not the only use of mutual inductives. As @herbelin mentions, even
and odd is one natural example. And inductive inductive types are a
generalization of mutual inductive types. I agree that the more general
presentation of inductive inductive types, seeing the context of an
inductive block arities as a real telescope is more pleasing than the
current representation where the arities are independent and one needs to
apply lifting to turn that into a context.
Le sam. 22 sept. 2018 à 00:14, Hugo Herbelin <notifications@github.com> a
écrit :

herbelin:
@JasonGross' request granted. @ppedrot: would you be ok to review and be the assignee?

vernac/auto_ind_decl.ml
```diff
@@ -232,9 +233,17 @@ let build_beq_scheme mode kn =
...
-	    (match Environ.constant_opt_value_in env (kn, u) with
-	      | None -> raise (ParameterWithoutEquality (ConstRef kn))
-              | Some c -> aux (Term.applist (c,a)))
+             (match Environ.constant_opt_value_in env (kn, u) with
+              | Some c -> aux (Term.applist (c,a))
+              | None ->
+                 (* Support for working in a context with "eq_x : x -> x -> bool" *)
+                 (* Needs Hints, see test suite *)
+                 let mp,dir,lbl = Constant.repr3 kn in
+                 let kneq = Constant.make3 mp dir (Label.make ("eq_"^(Label.to_string lbl))) in
```
ppedrot: This look quite ad-hoc. A principled approach would be definitely better but for now I guess this is good, as other parts of the code rely on the same kind of trick.

ppedrot (approved):

vernac/auto_ind_decl.ml
SkySkimmer: Why not use change_label? It's still hacky but less adhoc.

ppedrot: I was complaining about the hardwired name scheme, not the way the name was obtained. But you're right.

ppedrot:
@herbelin If you use `Constant.change_label` I will perform the merge.

herbelin:
@ppedrot: the whole code of `auto_ind_decl.ml` is rather fragile. My objective with the PR was merely to prevent one more bug report about the anomalies it was raising. We could work at improving the code or extent of `auto_ind_decl.ml` but if there are possibly strictly better replacements from @mattam82 or @gares, it is probably not so worth.

Anyway, I used `change_label` in the 3 places where it was relevant.

ppedrot:
@herbelin Yes, this was a mere remark, not a request for change.

herbelin:
@ppedrot: CI is green.

PR #8510, opened 2018-09-19
Extensible interpretation for attributes

SkySkimmer:
This PR adds a way to define new attributes without touching the attribute type or its interp function. It should also work from plugins although that's untested. 
Pre-existent attributes are moved to the new system, with some special casing for polymorphism as we do silly things with it.
As a bonus the classifier stops ignoring local `Polymorphic`/`#[polymorphic]` attributes.

The API looks like this (in `Attributes`):
~~~ocaml
type t (* parsed attributes *)
type 'a attribute (* typed key for some attribute, eg we can have [val program : bool attribute] *)

val read : 'a attribute -> t -> 'a option
(* was the attribute set, and if so what's the value? *)

type 'a flag_parser = 'a option -> ?loc:Loc.t -> Vernacexpr.vernac_flag_value -> 'a
(** How to parse some specific attribute key (eg "polymorphic" and
   "monomorphic" are separate keys for the same attribute). The
   previous value of the attribute if set is also passed, allowing
   handling of multiply set attributes as you desire (error, make a
   list, etc). *)

val register_attribute : name:string -> 'a flag_parser CString.Map.t -> 'a attribute
(** You can register multiple keys for the same attribute, eg
   "template" and "notemplate" keys produce values for the same
   "templateness" attribute. The [name] is currently not used. *)

(* type t is inhabited by these 2, whose use is centralised. users don't care about them. *)
val mk_atts : ?polymorphic: bool ->
  unit -> t

val attributes_of_flags : Vernacexpr.vernac_flags -> t ->
  bool option (* polymorphism attr *) * t
~~~

Underneath `t` is just a store, and `register_attribute` updates some ref with ways to fill the store.

The magic we do with polymorphism is too much for me so I gave up on removing it in this PR (there is also some weird stuff still done with program mode, but at least I got that out of the attribute stuff).
Aside from the magic around VernacLoad or whatever this stuff is, if we make polymorphism a regular attribute we can only get a `bool option` from the attribute type, so we would have to use `is_universe_polymorphism ()` all over the place again for the default case. I feel like this is an indicator that we're doing something suboptimally but not sure of the solution. It seems related to the issue of detecting unsupported attributes: each command wants not the type `t` of the collected attributes but instead some subset of it, or even a transformation of that subset (transformation which can eg insert default values). That however is beyond this PR or orthogonal to it.

3rd party plugins using attributes are broken: since the system is generic there are no more projections to access the attributes, you need to use functions. Overlay will come later.

### DIFF
--- a/plugins/firstorder/g_ground.ml4
+++ b/plugins/firstorder/g_ground.ml4
@@ -67,9 +67,9 @@ let (set_default_solver, default_solver, print_default_solver) =
 
 VERNAC COMMAND FUNCTIONAL EXTEND Firstorder_Set_Solver CLASSIFIED AS SIDEFF
 | [ "Set" "Firstorder" "Solver" tactic(t) ] -> [
-    fun ~atts ~st -> let open Attributes in
+    fun ~atts ~st ->
       set_default_solver
-        (Locality.make_section_locality atts.locality)
+        (Locality.make_section_locality (Attributes.locality atts))
         (Tacintern.glob_tactic t);
         st
   ]
--- a/plugins/firstorder/g_ground.ml4
+++ b/plugins/firstorder/g_ground.ml4
@@ -67,7 +67,7 @@ let (set_default_solver, default_solver, print_default_solver) =
 
 VERNAC COMMAND FUNCTIONAL EXTEND Firstorder_Set_Solver CLASSIFIED AS SIDEFF
 | [ "Set" "Firstorder" "Solver" tactic(t) ] -> [
-    fun ~atts ~st -> let open Vernacinterp in
+    fun ~atts ~st -> let open Attributes in
       set_default_solver
         (Locality.make_section_locality atts.locality)
         (Tacintern.glob_tactic t);
--- a/plugins/ltac/extratactics.ml4
+++ b/plugins/ltac/extratactics.ml4
@@ -27,7 +27,6 @@ open Equality
 open Namegen
 open Tactypes
 open Proofview.Notations
-open Attributes
 
 DECLARE PLUGIN "ltac_plugin"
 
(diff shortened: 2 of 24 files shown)


ejgallego:
Need to have a deeper look but I can't say I am happy to see more imperative APIs added to the core. These tend to be problematic in terms of implementing multi-process as marshalling is not possible, and also will have the typical problems with regards to introspection. Of course we can implement the proper interfaces inside the attribute module, but for each of these patterns we are going to have to duplicate the whole management code.

The OCaml way, or even a reification of the attribute type would do the same work while being less tricky to work with. Maybe this should be a good start for a CEP-like discussion?

ejgallego:
In particular a key point is the type of the vernac interpretation function. I'd dare to say that the tendency there would be to turn `atts` into a row type and have `coqpp` generate the proper type constraints for each interpretation function based on the `atts` it has declared. This PR indeed goes a bit into the opposite direction as now the type is more opaque for the type system.

SkySkimmer:
>The OCaml way, or even a reification of the attribute type would do the same job while being less tricky to work with. Maybe this should be a good start for a CEP-like discussion?

What is the ocaml way?

>In particular a key point is the type of the vernac interpretation function. I'd dare to say that the tendency there would be to turn atts into a row type and have coqpp generate the proper type constraints for each interpretation function based on the atts it has declared. This PR indeed goes a bit into the opposite direction as now the type is more opaque for the type system.

I can kind of see that. So eg you could write 
~~~
VERNAC COMMAND EXTEND Foo CLASSIFIED AS WHATEVER
| #[ locality = Somewhere.locality_attribute ] [ "Foo" ] -> [
  foo (Locality.make_section_locality locality)
]
END
~~~
and it would figure out that `Foo` supports the locality attribute as defined in `Somewhere`, and only that attribute?
I don't have the parser skills to make it happen though.

ppedrot:
Please use Dyn rather than Store. The only interest of the latter is that it is slightly more effective, which can matter with e.g. evar-local data, but it is also potentially unsound.

herbelin:
> if we make polymorphism a regular attribute we can only get a bool option from the attribute type, so we would have to use is_universe_polymorphism () all over the place again for the default case (...) each command wants not the type t of the collected attributes but instead some subset of it, or even a transformation of that subset (transformation which can eg insert default values).

This seems related to the difficulties I met when trying to support optional features in tactics: when to take into account the global state of an option (say [`Set Structural Injection`](https://github.com/coq/coq/blob/af0a04b8e16c2554e0c747da6d625799b332f5fe/tactics/equality.ml#L68))?

Let me digress for a while.

Some callers to the tactic want the parameterization done by the tactic. Other want to enforce the parameterization at call time. Additionally, when the tactic is a tactical, some other callers want the default behavior to be taken dynamically (e.g. by dynamically passing a `set Structural Injection in tac`).

A solution, and the question is whether this is overkill or just what we need, is to provide two variants of the tactic and three levels of use
1. An inner tactic take a `bool option` and, if `None` uses the current value in the global state
2. An outer tactic takes a `Static of bool | SetDefaultAtCallingTime | SetDefaultAtUsingTime` and calls the inner tactic using the value in the global state to replace `SetDefaultAtCallingTime` by a `Some`

Back to the polymorphism flag, what are the use cases? For instance, would there be an interest in supporting a local use of polymorphism? If yes, it seems that we are in the same case as what I had with tactics. If no, it would then be indeed enough that the attiribute manager passes the default value at calling time and that the commands supporting polymorphism only take a `bool`.

SkySkimmer:
Closing in favor of #8515.

Issue #8511, opened 2018-09-20
merge-pr.sh does not check if user.email and user.signingkey match


MSoegtropIMC:
The PR merge script does not check, if the git configurations

- user.email
- user.signingkey

match. If the don't match, the signing will not report an error, but the merge will be unverified. It is easy to check this.

PR #8512, opened 2018-09-20
Fix issue #8511 merge-pr.sh does not check if user.email and user.sig…

MSoegtropIMC:
This PR adds a check to merge-pr.sh if the ID of the git configured user signing key matches the git configured user email.

On the way I made two more minor changes:

- made a missing signing key an error (one can still override errors, but afaik signing is now mandatory, so this should be an error)

- gave symbolic names to the style color codes and changed the highlight style from blue to bold, because blue is unreadable on black background terminals (the default for cygwin and hard to change)

### DIFF
--- a/dev/tools/merge-pr.sh
+++ b/dev/tools/merge-pr.sh
@@ -18,19 +18,21 @@ else
     QUICK_CONF=""
 fi
 
-RED="\033[31m"
-RESET="\033[0m"
-GREEN="\033[32m"
-BLUE="\033[34m"
-YELLOW="\033[33m"
+# Style terminal codes
+SRESET="\033[0m"
+SHIGHLIGHT="\033[1m"     # bold (blue is unredable with black background terminals)
+SINFO="\033[32m"         # green
+SWARNING="\033[33m"      # yellow
+SERROR="\033[31m\033[1m" # red bold
+
 info() {
-  echo -e "${GREEN}info:${RESET} $1 ${RESET}"
+  echo -e "${SINFO}info:${SRESET} $1 ${SRESET}"
 }
 error() {
-  echo -e "${RED}error:${RESET} $1 ${RESET}"
+  echo -e "${SERROR}error:${SRESET} $1 ${SRESET}"
 }
 warning() {
-  echo -e "${YELLOW}warning:${RESET} $1 ${RESET}"
+  echo -e "${SWARNING}warning:${SRESET} $1 ${SRESET}"
 }
 
 check_util() {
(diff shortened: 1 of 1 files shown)


gares (approved):

MSoegtropIMC:
Can someone please merge this?

dev/tools/merge-pr.sh
```diff
@@ -18,19 +18,21 @@ else
     QUICK_CONF=""
 fi
 
-RED="\033[31m"
-RESET="\033[0m"
-GREEN="\033[32m"
-BLUE="\033[34m"
-YELLOW="\033[33m"
+# Style terminal codes
```
maximedenes: These changes seem completely unrelated to the commit message.

dev/tools/merge-pr.sh
```diff
@@ -193,9 +195,26 @@ if [ "$KIND" = "[]" ]; then
...
+# Sanity check: user.signingkey exists in gpg and matches user.email
+# Get data from GIT configuration
+GIT_EMAIL="$(git config user.email | tr '[:upper:]' '[:lower:]')"
+# Ask GPG for the email for this key
+# See usr\share\doc\gnupg2\DETAILS for documentation of the --with-colons format
+# The returned ID is assumed to be of the form "name <email>", from which the email is extracted
+# In case the ID ist just an email, the sed is a NOP
+GPG_SIGNINGKEY_TYPE="$(gpg --list-secret-keys --with-colons "$GIT_SIGNINGKEY" | cut --delimiter=: --fields=1)"
+GPG_SIGNINGKEY_EMAIL="$(gpg --list-secret-keys --with-colons "$GIT_SIGNINGKEY" | cut --delimiter=: --fields=10 | sed -e 's|.*<\(.*\)>|\1|' | tr '[:upper:]' '[:lower:]')"
+if [ "$GPG_SIGNINGKEY_TYPE" != "sec" ] || [ "$GPG_SIGNINGKEY_EMAIL" != "$GIT_EMAIL" ]; then
```
maximedenes: Does this check work if the key has several e-mail addresses attached?

dev/tools/merge-pr.sh
MSoegtropIMC: @maximedenes : I mentioned in the PR text that I made two rather cosmetic changes on the way - not sure if such cosmetic changes are worthwhile to mention in a commit message.

maximedenes: Sorry, I missed it. In principle, a separate commit would have been good (imagine we have to revert the checks because we find out they are too restrictive, we may want to keep the cosmetic changes), but no need to redo it this time.

dev/tools/merge-pr.sh
maximedenes: In fact, maybe I can test that.

MSoegtropIMC: I didn't know that this is possible, but some googling revealed how to add a second email to an existing private key. It will likely not work. I will fix this.

maximedenes: Indeed, I have such a setup and the check fails on my machine.

MSoegtropIMC: Would you mind sending me the output of the below command via private mail:

```
gpg --list-secret-keys --with-colons <yourkeyid> | cut --delimiter=: --fields=10
```

maximedenes: done!

SkySkimmer:
@MSoegtropIMC if you don't have time to work on this shall we close the PR?

MSoegtropIMC:
Sorry I was way busy the last months, but I am about to start closing all my old issues and PRs.

Best regards,

Michael

vbgl:
@MSoegtropIMC Is someone working on this PR? Should it be closed? or merged?

MSoegtropIMC:
As discussed I have a fix but cause of the nature of the merge script and the issue reported, I can't test it. As discussed, I will push the change anyway and if it doesn't work, we simply fix it then (in the end one isn't forced to use the current merge script to do a merge).
Didn't do it in the last weeks since the GTK and incompatible cygwin changes came in my way.

Zimmi48:
The easiest way to test this is to push your fix and to ask @maximedenes to just try merging this PR with your new proposed version of the merge script.

ppedrot:
Shouldn't we close this?

MSoegtropIMC:
I am still on it, it is just lower prio than other things.

SkySkimmer:
merge-pr.sh isn't especially important now that we have coqbot, and this PR is dead. Let's close.

MSoegtropIMC:
@SkySkimmer : yes, I still didn't have time to test the PR with multi-email signatures.

I read in the contribution guide about coqbot, but I wonder what it does, so I am still using merge-pr.sh. I somehow like this extra safety. What would coqbot do in case merge-pr.sh would throw a warning or an error? Maybe it has the same functionality, but I couldn't find documentation for it.

SkySkimmer:
coqbot is stricted IIUC, for instance it won't let you merge without a `kind:` label etc

Zimmi48:
The only thing that it doesn't check is the CI status. It's assumed that the assignee has checked this before calling coqbot.

MSoegtropIMC:
And for everything for which merge-pr.sh produced a warning or error, coqbot will post back a message and refuse the merge?

Zimmi48:
Yes.

PR #8515, opened 2018-09-20
Command driven attributes

SkySkimmer:
In this model commands need to request the attributes they use (instead of getting some preparsed record), with the API encouraging them to error on unsupported attributes.
The "core" commands (non vernacextend commands) still use the record type to reduce code movement, we can change that later to be more precise about non supported attributes.

An attribute is first defined:
~~~ocaml
let template = bool_attribute ~name:"Template" ~on:"template" ~off:"notemplate"
~~~
(more complicated attributes are also possible, see eg the deprecation one)
The one the command side attributes are used. The following is a command supporting deprecation and templateness, and will error with other attributes:
~~~ocaml
VERNAC COMMAND FUNCTIONAL EXTEND Foo CLASSIFIED AS Whatever
| [ "Foo" ] =>
  [ fun ~atts ~st -> let open Attributes in
    let deprecation, template = Runner.((!) (deprecation ++ template) atts) in
    (* deprecation : Attributes.deprecation option, template : bool option *)
    foo deprecation template;
    st
  ]
END
~~~
Without support from VERNAC EXTEND we can't stop FUNCTIONAL EXTEND commands from ignoring the `atts`, but the situation wrt silently ignoring attributes should be better than before.

Unlike #8510 we're not adding some global ref.

### DIFF
--- a/vernac/attributes.mli
+++ b/vernac/attributes.mli
@@ -8,14 +8,19 @@
 (*         *     (see LICENSE file for the text of the license)         *)
 (************************************************************************)
 
-(** Interpretation of extended vernac phrases. *)
+type deprecation = { since : string option ; note : string option }
 
-type 'a vernac_command = 'a -> atts:Attributes.t -> st:Vernacstate.t -> Vernacstate.t
+val mk_deprecation : ?since: string option -> ?note: string option -> unit -> deprecation
 
-type plugin_args = Genarg.raw_generic_argument list
+type t = {
+  loc : Loc.t option;
+  locality : bool option;
+  polymorphic : bool;
+  template : bool option;
+  program : bool;
+  deprecated : deprecation option;
+}
 
-val vinterp_init : unit -> unit
-val vinterp_add : bool -> Vernacexpr.extend_name -> plugin_args vernac_command -> unit
-val overwriting_vinterp_add : Vernacexpr.extend_name -> plugin_args vernac_command -> unit
-
-val call : Vernacexpr.extend_name -> plugin_args -> atts:Attributes.t -> st:Vernacstate.t -> Vernacstate.t
+val mk_atts : ?loc: Loc.t option -> ?locality: bool option ->
+  ?polymorphic: bool -> ?template:bool option ->
+  ?program: bool -> ?deprecated: deprecation option -> unit -> t
--- a/vernac/attributes.mli
+++ b/vernac/attributes.mli
@@ -23,3 +23,6 @@ type t = {
 val mk_atts : ?locality: bool option ->
   ?polymorphic: bool -> ?template:bool option ->
   ?program: bool -> ?deprecated: deprecation option -> unit -> t
+
+val attributes_of_flags : Vernacexpr.vernac_flags -> t ->
+  bool option (* polymorphism attr *) * t
--- a/vernac/attributes.mli
+++ b/vernac/attributes.mli
@@ -13,7 +13,6 @@ type deprecation = { since : string option ; note : string option }
 val mk_deprecation : ?since: string option -> ?note: string option -> unit -> deprecation
 
 type t = {
-  loc : Loc.t option;
   locality : bool option;
   polymorphic : bool;
   template : bool option;
@@ -21,6 +20,6 @@ type t = {
   deprecated : deprecation option;
 }
 
-val mk_atts : ?loc: Loc.t option -> ?locality: bool option ->
+val mk_atts : ?locality: bool option ->
   ?polymorphic: bool -> ?template:bool option ->
   ?program: bool -> ?deprecated: deprecation option -> unit -> t
--- a/vernac/attributes.mli
+++ b/vernac/attributes.mli
@@ -20,9 +20,9 @@ type t = {
   deprecated : deprecation option;
 }
 
-val mk_atts : ?locality: bool option ->
-  ?polymorphic: bool -> ?template:bool option ->
-  ?program: bool -> ?deprecated: deprecation option -> unit -> t
+val mk_atts :
+  ?polymorphic: bool ->
+  ?program: bool -> unit -> t
 
 val attributes_of_flags : Vernacexpr.vernac_flags -> t ->
   bool option (* polymorphism attr *) * t
--- a/vernac/attributes.mli
+++ b/vernac/attributes.mli
@@ -8,9 +8,33 @@
 (*         *     (see LICENSE file for the text of the license)         *)
 (************************************************************************)
 
+open Vernacexpr
+
+type 'a attribute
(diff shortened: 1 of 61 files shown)


grammar/vernacextend.mlp
```diff
@@ -89,6 +89,7 @@ EXTEND
   rule:
     [ [ "["; OPT "-"; l = LIST1 args; "]";
         d = deprecation; c = OPT classifier; "->"; "["; e = Pcaml.expr; "]" ->
+        (** XXX Error on nonempty attributes? *)
       let b = <:expr< fun ~{atts} ~{st} -> ( let () = $e$ in st ) >> in
```
SkySkimmer: I'll try adding an atrribute check here later, see if anything breaks.

vernac/attributes.mli
```diff
@@ -0,0 +1,102 @@
...
+  (** Notations to combine and run attributes. *)
+
+  val (++) : 'a attribute -> 'b attribute -> ('a * 'b) attribute
+  (** Combine 2 attributes. If any keys are in common an error will be raised. *)
+
+  val run : 'a attribute -> vernac_flags -> 'a * vernac_flags
+  (** Returns the unsupported attributes as well as the parsed value. *)
+
+  val (!) : 'a attribute -> vernac_flags -> 'a
+  (** Errors on unsupported attributes. *)
```
SkySkimmer: This may be over-use of notations. Maybe this should be called `run` or `parse` and the `run` above should have some other name.

Zimmi48: I think so too.

SkySkimmer: Changed.

vernac/attributes.mli
```diff
@@ -0,0 +1,100 @@
...
+module Notations : sig
+  (** Notations to combine attributes. *)
+
+  val (++) : 'a attribute -> 'b attribute -> ('a * 'b) attribute
+  (** Combine 2 attributes. If any keys are in common an error will be raised. *)
+end
+
+(** Definitions for some standard attributes. *)
+
+type deprecation = { since : string option ; note : string option }
```
herbelin: Are there some use for deprecation without a `since` flag?

herbelin (commented):
The `parse` and `++` idea is very nice. One thing I'm however worrying is that the risk we forget to add a `parse` line to command which takes no attribute is high. So, it seems to me that the check that all supported attributes have been consumed should rather be on the side of `vernacentries`.

gares (approved):

SkySkimmer:
I'm looking into making `universes(template)` and `universes(polymorphic)` work (see https://github.com/coq/coq/pull/8488#pullrequestreview-157663174).
I'm not sure that the current design is quite right for it so this PR may yet change.

grammar/vernacextend.mlp
SkySkimmer: Seems like nothing breaks.

SkySkimmer: @ppedrot told me the `atts` variable is in scope in the expression so nevermind.

SkySkimmer:
Here's some further work making polymorphism a mostly normal attribute. Note the way that uses the option is abstracted inside the Attributes module (except for a leak to the stm).
I'm wondering about how attributes and options should interact. For instance program mode is very pervasive and I'm pretty sure that some commands where it would have an impact reject the `#[program]` attribute. What's a principled way to handle this?
It would also make sense to set options in a generic way through attributes, eg `#[set_option="Print Universes", set_option="Print Projections"]`.

ejgallego:
> It would also make sense to set options in a generic way through attributes, eg `#[set_option="Print Universes", set_option="Print Projections"]`.

This needs indeed thinking; recall that options are set by the summary so it is very easy to get unexpected interactions when modifying them as they will be reset to the previous value in maybe unexpected ways [for example, the kernel could force some kind of closure]

SkySkimmer:
I think this PR needs to modify the parsing side to be complete.
ie instead of 
~~~
VERNAC COMMAND EXTEND AddRelation CLASSIFIED AS SIDEFF
  | [ "Add" "Relation" constr(a) constr(aeq) "reflexivity" "proved" "by" constr(lemma1)
        "symmetry" "proved" "by" constr(lemma2) "as" ident(n) ] ->
      [ declare_relation (Attributes.parse rewrite_attributes atts) a aeq n (Some lemma1) (Some lemma2) None ]
END
~~~
we should have
~~~
VERNAC COMMAND EXTEND AddRelation CLASSIFIED AS SIDEFF
  | #[ rwatts = rewrite_attributes ] [ "Add" "Relation" constr(a) constr(aeq) "reflexivity" "proved" "by" constr(lemma1)
        "symmetry" "proved" "by" constr(lemma2) "as" ident(n) ] ->
      [ declare_relation rwatts a aeq n (Some lemma1) (Some lemma2) None ]
END
~~~
or some such syntax, such that the `atts : vernac_flags` variable is not available and we have to go through `Attributes.parse`.
The syntax I imagine looks like `#[ pattern = attribute ]` where `attribute : 'a attribute` and `pattern` is a destructuring pattern for `'a` whose variables bind in the expression.

However I do not know how to implement this. Can someone guide me in modifying the ml4 preprocessor?

SkySkimmer:
I managed to do the VERNAC EXTEND change I wanted in coqpp, have a look.

ppedrot:
That looks reasonable to me. The parsing rule is a bit ad-hoc, you could probably factor out the SEMICOLON so as to allow not to finish the list by one.

SkySkimmer:
Changed

Gaëtan Gilbert

SkySkimmer:
TODO change the way polymorphism is transformed into `universes(polymorphic)` so that the error message when unsupported makes sense.

ppedrot (approved):
This PR would probably benefit from being split in smaller pieces, but I like the implementation nonetheless.

SkySkimmer:
>TODO change the way polymorphism is transformed into universes(polymorphic) so that the error message when unsupported makes sense.

In the end I can't see how to treat `Polymorphic` as a regular attribute, say "unsupported Polymorphic" on `Polymorphic Ltac foo := idtac.` and not allow `#[Polymorphic] Definition foo := Prop.` without even warning. So let's forget about it.

SkySkimmer:
So I think this is basically ready. Handling of program mode attributes may need to be refined further with experience but otherwise it should work.
We then need an assignee and reviews.

ejgallego (commented):
I'd dare to say that this is looking very good.

vbgl (approved):
Many thanks Gaëtan for all this work. It looks very good.

ppedrot:
OK let's merge not to delay this PR further.

herbelin:
Very nice PR by the way.

PR #8518, opened 2018-09-20
[dune] Add "watch" target for continuous build mode.

ejgallego:
`make -f Makefile.dune watch` will now watch for files changes
and rebuild what is needed.

This feature requires `fswatch` or `inotifywait`, and Dune >= 1.2.0.

Note that the current CI image ships an older Dune version.


SkySkimmer:
Why do we need to update the CI image? It doesn't test this feature.

ejgallego:
> Why do we need to update the CI image? It doesn't test this feature.

Ah well, good point indeed. I guess the image should be kept functional in case anyone does want to pull it?

Anyways if that's a burden we can remove the update; but in the end we will need a newer version sooner or later.

Zimmi48:
I think the same way as @SkySkimmer. This also doesn't include a bump of the version used in the Nix package, but that is not an issue since it is mostly a developer thing and developers can use more recent versions of nixpkgs and of Dune.

Zimmi48:
If the CI update is removed, I can test and merge. Otherwise, I'll need @SkySkimmer approval indeed.

ejgallego:
> This also doesn't include a bump of the version used in the Nix package, but that is not an issue since it is mostly a developer thing and developers can use more recent versions of nixpkgs and of Dune.

How can I do that? As the version is not local to the nix config file I am a bit lost.

Regarding the image update, I don't mind either way. Just for laziness I would apply the update as it is already done and we will use it in the future, [also it is convenient to be able to pull the CI image], but if you object I will remove it, tho after thinking a bit more it feels weird we ship some features our base image doesn't support.

Zimmi48:
>How can I do that? As the version is not local to the nix config file I am a bit lost.

A version of nixpkgs is pinned in `default.nix` but anyways I'm not requesting any action here.

>Regarding the image update, I don't mind either way. Just for laziness I would apply the update as it is already done and we will use it in the future, [also it is convenient to be able to pull the CI image], but if you object I will remove it, tho after thinking a bit more it feels weird we ship some features our base image doesn't support.

- CI hasn't passed yet though.
- I don't have a strong objection, just need @SkySkimmer approval then.
- As it is a dev feature, it does make sense that it is not supported in our base image.
- If you find it convenient like this, keep it like this.

ejgallego:
> * If you find it convenient like this, keep it like this.

Whatever @SkySkimmer says.

SkySkimmer (dismissed):
Please avoid unused changes.

Zimmi48 (approved):

Issue #8514, opened 2018-09-20
New Windows spurious CI failures.


Zimmi48:
I've seen two recent spurious failures of the Windows CI builds:
- https://gitlab.com/coq/coq/-/jobs/98687333
- https://gitlab.com/coq/coq/-/jobs/99619726

One is on win32 and the other on win64.
In the two cases, it fails during the installation step of ocamlbuild. I have no idea what could be going wrong. The logs that were uploaded as artifacts do not show any error...

cc @MSoegtropIMC

MSoegtropIMC:
I would say this is caused by not deleting cygwin. If a cygwin is reused, ocamlbuild already exists and the installation of it fails.

Commit 6a8c37c02504463afaa677641d75d9977020edf6 should fix this and also help in debugging such issues. This is very easy to port from v8.8 to master, so I think it would make sense to do this as a separate PR (the addons might take a bit longer).

MSoegtropIMC:
Since the logs show this behavior

- cygwin is deleted and freshly installed
- ocaml, sed and flexdll are build and installed
- ocamlfind is build

Then magically, before installing ocamlfind, this folder exists:
```
C:/cygwin64/usr/x86_64-w64-mingw32/sys-root/mingw/libocaml/ocamlbuild
```
exists. The only explanation for this I have, besides ocaml randomly installing ocamlfind, is that the cygwin folder is used by someone else.

Shall I switch to randomly generated or otherwise unique folder names? Cleaning up might be a bit tricky ...

MSoegtropIMC:
@Zimmi48 : just wanted to ask if you experienced any spurious windows CI failures after the fix. Should we doe more testing?

Zimmi48:
No more failures indeed.

PR #8519, opened 2018-09-21
Issue #8514 windows ci failures

MSoegtropIMC:
This is an attempt to fix issue #8514.

The PR merges the enhancements already done in v8.8.
On top it uses unique folder names for cygwin and coq dest to make sure that there is no dual use of the same folder. Also the folders are cleaned up after the CI run.

### DIFF
--- a/dev/ci/gitlab.bat
+++ b/dev/ci/gitlab.bat
@@ -1,5 +1,8 @@
 @ECHO OFF
 
+ECHO "Start Time"
+TIME /T
+
 REM This script builds and signs the Windows packages on Gitlab
 
 if %ARCH% == 32 (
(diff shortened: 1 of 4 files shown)


dev/ci/gitlab.bat
```diff
@@ -43,8 +70,46 @@ IF DEFINED WIN_CERTIFICATE_PATH (
...
+
+CALL :CleanupFolders
+
+ECHO "Finished Cleanup"
+TIME /T
+
 GOTO :EOF
 
+:ZipLogfiles
+  ECHO Zipping logfiles for artifact upload
```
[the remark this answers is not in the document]
MSoegtropIMC: Originally I did it like this to not clutter up the git folder more than necessary (artifacts can only be uploaded from the git folder), but I meanwhile found it a bit inconvenient. I will create an artifacts folder and put everything in there.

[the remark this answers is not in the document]
MSoegtropIMC: @SkySkimmer : I changed this and the files are more accessible now. One disadvantage is that the file change dates are gone this way. This can be useful information. Any ideas?

I could zip the flagfiles, which mostly exist to serve as timestamps, and keep the rest as copied files or I could dump a time stamp in the flagfiles. Of cause one has the information now also in the main log.

Another thing is that I probably I should rename the extensions to .log.txt and .err.txt. Otherwise they cannot be viewed in the browser.

dev/ci/gitlab.bat
```diff
@@ -43,8 +71,49 @@ IF DEFINED WIN_CERTIFICATE_PATH (
...
+
+CALL :CleanupFolders
+
+ECHO "Finished Cleanup"
+TIME /T
+
 GOTO :EOF
 
+:CopyLogFiles
+  ECHO Copy log fils for artifact upload
```
SkySkimmer: typo

dev/ci/gitlab.bat
```diff
@@ -43,8 +71,49 @@ IF DEFINED WIN_CERTIFICATE_PATH (
...
+  MKDIR artifacts\buildlogs
+  COPY %CYGROOT%\build\buildlogs\* artifacts\buildlogs
```
SkySkimmer: Is it not possible to move the files preserving times?

MSoegtropIMC: Actually it is fine. One just can't see the timestamp when browsing files, but when downloading the zip, the timestamps are correct. Since this is not needed that frequently, I think this is fine.

Zimmi48 (commented):
This seems like a good move. That being said, I've never seen the failure I reported for v8.8 on master so it is not impossible that it comes only after the commits that were merged into v8.8 and not into master yet.

Zimmi48:
Regarding your reliability tests. Note that another way is to start tests manually from the GitLab CI pipeline interface.

MSoegtropIMC:
@Zimmi48 : the additional addons change the timing and this can have a strong influence on the reproducability of intermittent tests. CI started to fail randomly after I added addons. The changes to the scripts I did afterwards to debug this. It e.g. could be that the long run time on v8.8 leads to reuse of runners, which almost never happens with short run time. It is really hard to tell why this folder pops into existence out of thin air. Anyway I will exclude one possible root cause after the other until it is fixed. If this change doesn't fix it, I will start a background shell script which supervises the existence of the folder every 10 seconds.

MSoegtropIMC:
Btw.: Shall I make a PR with these changes for v8.8?

Zimmi48:
>Btw.: Shall I make a PR with these changes for v8.8?

Sure, if it's not too difficult.

MSoegtropIMC:
I created PR #8523 to backport this on v8.8

Zimmi48:
Your reliability tests are pretty successful, although doing the same on v8.8 will be the decisive test.

MSoegtropIMC:
I feel not responsible for the travis Mac errors. I restarted it, but no luck. If this is normal, I would suggest merge this PR anyway (it is already merged in 8.8). Otherwise I need some help with the travis failures.

Zimmi48:
@MSoegtropIMC It is unrelated indeed. It was fixed in master. But yes, I'll merge this PR now.

PR #8520, opened 2018-09-21
[ci] [docker] Move to OPAM 2.0

ejgallego:
The OPAM 1.2 repository has been frozen, thus we must use OPAM 2.0 if
we want to get newer versions of packages / compilers.

For now we must perform a manual install of OPAM as no packages for
Ubuntu 18.04 exist.

Note that this means nothing about the installability of Coq itself,
whose OPAM repository should remain in 1.2 format for quite a long
period.

### DIFF
--- a/dev/ci/docker/bionic_coq/Dockerfile
+++ b/dev/ci/docker/bionic_coq/Dockerfile
@@ -1,4 +1,4 @@
-# CACHEKEY: "bionic_coq-V2018-09-21-V2"
+# CACHEKEY: "bionic_coq-V2018-09-23-V01"
 # ^^ Update when modifying this file.
 
 FROM ubuntu:bionic
@@ -8,7 +8,7 @@ ENV DEBIAN_FRONTEND="noninteractive"
 
 RUN apt-get update -qq && apt-get install --no-install-recommends -y -qq \
         # Dependencies of the image, the test-suite and external projects
-        m4 automake autoconf time wget rsync git gcc-multilib opam \
+        m4 automake autoconf time wget rsync git gcc-multilib build-essential unzip \
         # Dependencies of lablgtk (for CoqIDE)
         libgtk2.0-dev libgtksourceview2.0-dev \
         # Dependencies of stdlib and sphinx doc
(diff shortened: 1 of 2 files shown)


dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -18,9 +18,12 @@ RUN apt-get update -qq && apt-get install --no-install-recommends -y -qq \
         texlive-science tipa
 
 # More dependencies of the sphinx doc
-RUN pip3 install sphinx==1.7.8 sphinx_rtd_theme==0.2.5b2 \
+RUN pip3 install wheel sphinx==1.7.8 sphinx_rtd_theme==0.2.5b2 \
```
SkySkimmer: What is `wheel`?

SkySkimmer: Also all the other pip packages have explicit versions, so this one should probably follow.

Zimmi48: If you want to install wheel, you should probably do it with `apt`.

ejgallego: I thought it was already installed, but I see it is not.

ejgallego:
I wonder whether we should set njobs in OPAM 2, it is supposed to detect the available processors automatically.

ejgallego:
So OK @SkySkimmer , this is ready technically. Now, what should we do ?

Zimmi48:
@ejgallego You still have the "needs: progress" and the `true_disabled`. Shouldn't you also update Travis to use OPAM 2.0 as part of this PR?

ejgallego:
Indeed, the hack to help me test this PR is still on, but it shouldn't affect the decision about it; of course I will prepare a proper version for merge.

> Shouldn't you also update Travis to use OPAM 2.0 as part of this PR?

Good question. I tend to see Travis as more and more unsupported, so maybe this is the moment we want to remove support for it?

Zimmi48:
>I tend to see Travis as more and more unsupported, so maybe this is the moment we want to remove support for it?

:+1: 

We can keep only the macOS part and the linter until this moves to GitLab CI.

SkySkimmer (approved):
Maybe do travis in another PR so that this one can just go in once true_disabled is changed.

ejgallego:
> Maybe do travis in another PR so that this one can just go in once true_disabled is changed.

OK, submitting definitive version.

PR #8527, opened 2018-09-22
dev/doc/profiling.txt: per-component flame graphs

andres-erbsen:
**Kind:** infrastructure, documentation.

Add an example of generating flame graphs of coq execution time. [Example profile (5MB svg)](https://cdn.rawgit.com/andres-erbsen/b29b29cb6480dfc6a662062e4fcd0ae3/raw/304fc3fea9630c8e453929aa7920ca8a2a570d0b/stdlib_categorized_outermost.svg) of master after 3781e50331d563b47f1ea1ce1773a71db68fdb2a building stdlib. I find this significantly nicer to look at than `perf report`.

I am guessing that @ppedrot may be the appropriate person to decide whether something like this should be in `dev/doc`.

### DIFF
--- a/dev/doc/profiling.txt
+++ b/dev/doc/profiling.txt
@@ -21,6 +21,58 @@ and plug into the process
 
 perf record -g -p PID
 
+### Per-component [flame graphs](https://github.com/brendangregg/FlameGraph)
+
+I (Andres Erbsen) have found it useful to look at library-wide flame graphs of
+coq time consumption.  As the Ltac interpreter stack is reflected in the OCaml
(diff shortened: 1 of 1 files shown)


Zimmi48:
While leaving to @ppedrot the task of reviewing this, I'm noticing that this `.txt` file is already pretty close to being a Markdown file (and getting closer with this addition), so what about changing the file extension now?

ppedrot (approved):
Development documentation is always good to write somewhere, so I'm perfectly fine with merging this.

PR #8534, opened 2018-09-23
Checking if low-level name printers are used on purpose or not

herbelin:
This is the part of #7817 which fix cases of use of a low-level printer in user-level messages. It also clarifies some cases of a printer used for debugging but not explicitly calling the corresponding `debug_printer` function in `names.ml`.

**Kind:** bug fix

### DIFF
--- a/checker/environ.ml
+++ b/checker/environ.ml
@@ -183,7 +183,7 @@ let lookup_mind kn env =
 
 let add_mind kn mib env =
   if Mindmap_env.mem kn env.env_globals.env_inductives then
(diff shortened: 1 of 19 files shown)


ejgallego (commented):
I am not an expert here but so far what I understand looks OK to me. Who would be the proper assignee in your opinion @herbelin ?

ppedrot (approved):
No specific comment, everything looks in order.

maximedenes (approved):
Superfically looks ok, although the history would have benefited from separating independent changes, so that reverting just one (if we ever need to) is easy.

PR #7193, opened 2018-04-07
Fixes #7192: Print Assumptions does not enter implementation of submodules.

herbelin:
**Kind:** bug fix

Fixes / closes #7192

This was relatively easy to fix: we look manually for the implementation at each level of nesting rather than using the signature for the n first levels and looking for the implementation only in the n+1-th level.

Review from @letouzey needed.

PR #8535, opened 2018-09-23
Fixing #8532: regression in Print Assumptions within a functor.

### LINKED ISSUE #8532: Print Assumption anomaly (Assumptions of a parameter in a Module). kind: anomaly
#### Version
The Coq Proof Assistant, version 8.8.1 (September 2018)
compiled on Sep 5 2018 13:38:41 with OCaml 4.02.3

#### Operating system
Debian GNU/Linux 9 (stretch) 64-bit

This code raises an anomaly.
```
Module Type Typ.
  Parameter Inline(10) t : Type.
End Typ.
Module  Terms_mod (SetVars : Typ).
Print Assumptions SetVars.t.
End Terms_mod.
```
It was discovered during the use of Coq.Structures.Equalities library.

> Anomaly
> "File "vernac/assumptions.ml", line 74, characters 19-25: Assertion failed."
> Please report at http://coq.inria.fr/bugs/.
>

Maybe this tiny example has no sense, but this anomaly leads to impossibility of showing the assumptions of the soundness theorem in my project:
https://github.com/georgydunaev/VerifiedMathFoundations/commit/a2e7ab6269c60cc172ddf49fd4dfafc04668bc71
Shall I do it in a different way?

herbelin:
The regression was introduced in 1522b989 (PR #7193) which itself was fixing bug #7192. (Note another regression introduced by the same commit which has been fixed in #8416 :worried:.)

**Kind:** bug fix

Fixes / closes #8532

### DIFF
--- a/test-suite/bugs/closed/8532.v
+++ b/test-suite/bugs/closed/8532.v
@@ -1,0 +1,8 @@
+(* Checking Print Assumptions relatively to a bound module *)
+
+Module Type Typ.
+  Parameter Inline(10) t : Type.
+End Typ.
+Module  Terms_mod (SetVars : Typ).
+Print Assumptions SetVars.t.
+End Terms_mod.
--- a/vernac/assumptions.ml
+++ b/vernac/assumptions.ml
@@ -71,7 +71,7 @@ let rec fields_of_functor f subs mp0 args = function
 let rec lookup_module_in_impl mp =
     match mp with
     | MPfile _ -> Global.lookup_module mp
-    | MPbound _ -> assert false
+    | MPbound _ -> Global.lookup_module mp
     | MPdot (mp',lab') ->
        if ModPath.equal mp' (Global.current_modpath ()) then
          Global.lookup_module mp


ppedrot:
I think that this shows that nobody really understands how `Print Assumptions` really works...

herbelin:
> I think that this shows that nobody really understands how Print Assumptions really works...

I would not be so extreme, but this was indeed here trial-and-errors. It did not occur to me that the `MPBound` case (i.e. a `X` in `Module F(X:T)`) could have happened and that the `lookup` which existed before #7193 was useful for it.

So, yes, I don't have a global view at how `Print Assumptions` works and I made the initial fix by starting from the symptom.

herbelin:
> it must be merged ASAP.

To whom are you saying it?

ppedrot:
Since @mattam82 is not quite responsive, I am assigning myself and I am going to merge it, say tomorrow morning. The PR consists in partially reverting a change, so we shouldn't really bicker about it.

Zimmi48:
@ppedrot Would be even better if you could merge tonight. CI should have passed fully in less than an hour.

ppedrot:
Will do. I don't think that the CI is really testing this, but well...

ppedrot (approved):

Issue #8528, opened 2018-09-22
Addon dependency management in Windows installer


MSoegtropIMC:
This is a follow up on the discussion in PR #8322.

Currently there is no dependency checking in the installer, so one can end up with an installation where addons required by another addon are not installed.

The goal is to automatically check dependencies in the NSIS installer.

MSoegtropIMC:
I reopened it, because it also needs to go into master.
This will take a short while, because I must first properly add the addons in master (#8322).

Zimmi48:
Well, I was assuming that you would include it in master from the beginning. Anyways, if this must be kept open, please remove the 8.8.2 milestone.

MSoegtropIMC:
Fixed (in master) by #8322

PR #8542, opened 2018-09-24
Add addon dependency management to windows installer

MSoegtropIMC:
This PR is a fix for issue #8528 "Addon dependency management in Windows installer"

The declaration of the dependencies in the make file is not terribly elegant because the installer section name has to be given twice, once inside of the build_prep block and once outside. The reason is that the installer section must be declared inside the build_prep block, because it depends on installer delta file lists, and these can only be generated when something is actually build. The dependency management is better done outside of the build_prep blocks, because several addons might depend on one (e.g. ssreflect) and if the dependency management would be inside of the build_prep blocks, it would be processed only once.

I think the provided solution is reasonably elegant and robust. At least the duplications of the section names are very close in the code.

In the installer dependencies are strictly enforced and whenever a section is selected cause of a dependency a message box pops up. If a required section is unselected, it is immediately selected again and also a message box pops up explaining why.

### DIFF
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -611,6 +611,46 @@ function installer_addon_section {
 }
 
 # ------------------------------------------------------------------------------
+# Start an installer addon dependency group
+#
+# parameters
+# $1 identifier of the section which depends on other sections
+# The parameters must match the $1 parameter of a installer_addon_section call
+# ------------------------------------------------------------------------------
+
+dependencysections=
+
+function installer_addon_dependency_beg {
+  installer_addon_dependency "$1"
+  dependencysections="$1 $dependencysections"
+}
+
+# ------------------------------------------------------------------------------
+# End an installer addon dependency group
+# ------------------------------------------------------------------------------
+
+function installer_addon_dependency_end {
+  set -- $dependencysections
+  shift
+  dependencysections="$*"
+}
+
+# ------------------------------------------------------------------------------
+# Create an nsis installer addon dependency entry
+# This needs to be bracketed with installer_addon_dependencies_beg/end
+#
+# parameters
+# $1 identifier of the section on which other sections might depend
+# The parameters must match the $1 parameter of a installer_addon_section call
+# ------------------------------------------------------------------------------
+
+function installer_addon_dependency {
+  for section in $dependencysections ; do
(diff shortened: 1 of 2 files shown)


Zimmi48:
Note that while downloading the installer, I have noticed that the open source archive was still present in the artifacts. I thought you had removed it unless it was in "signing" mode. Anyways, not a big deal, and certainly no urgency to change this.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -1696,8 +1746,10 @@ function make_addon_menhirlib {
 # CompCert
 
 function make_addon_compcert {
+  installer_addon_dependency_beg compcert
   make_menhir
   make_addon_menhirlib
+  installer_addon_dependency_end
```
Zimmi48: For some reason, I can select CompCert and it won't select menhir. Am I right to think that this is not expected?

Zimmi48 (commented):
I am not really able to review this so I tested it instead. It generally looks good to me, except the note on the menhir dependency.

dev/build/windows/makecoq_mingw.sh
MSoegtropIMC: There is a module "make_menhir" and a module "make_addon_menhir". The first just compiles menhir and makes it available at build time. CompCert needs menhir for building. The addon  actually installs menhir with Coq.

As far as I can tell for proving code correct against CompCert (like VST does) one does (or might) need menhirlib (to reason about the generated parser files), but one does not need menhir. At least I think some of the .vo files installed with CompCert depend on menhirlib.

So this works as intended. A dependency would be generated if compcert would call "make_addon_menhir" instead of "make_menhir", but as far as I understand this, this dependency is not required.

Zimmi48: OK perfect then.

Zimmi48 (approved):

MSoegtropIMC:
> I am not really able to review this

I can imagine. The NSIS language is really bizarre, a kind of macro assembler with string valued registers and a stack. And most of the examples on the NSIS Wiki page are about 10x as complicated as required to get the job done. But meanwhile I got the basic ideas and am quite fast in writing NSIS code.

I must say I prefer WIX or even Orca over NSIS, but I can live with it and meanwhile it is probably not worthwhile to change it. Also for Coq MSI installers might be slow and MSI is not intended for use cases like installing 3 variants in 3 different folders and then deleting 2 of them. MSI installers would leave quite a broad trace in the registry and the system component database if this is done. A dumb NSIS installer is more appropriate for this. So under the line I am meanwhile quite ok with it - it gets the job done.

Issue #8533, opened 2018-09-23
[ci] Bedrock is routinely failing in the CI system.


ejgallego:
Since a few days, `ci-bedrock2` is routinely failing in the CI system, due to what it looks like a memory exhaustion problem. Example: https://gitlab.com/coq/coq/-/jobs/100472191

This should be mitigated or `ci-bedrock` will be put in the allow-failure category. Possible mitigation strategies include: removing the problematic files from the ci build, forcing `-j1`, removing parallelism from the Coq tactics, or using a runner tag to request runners with more memory.

cc: @JasonGross @samuelgruetter

Zimmi48:
What has been done previously when fiat-crypto was hitting a similar wall was to split the target into two, run the first with `-j2` and the second with `-j1`. This did the trick.

ejgallego:
Indeed, -j2 may be the culprit here but however some workers only have 2Gb of RAM so that's not a lot.

samuelgruetter:
@ejgallego does `-j1` fix the problem?
(Another culprit might be the `par:` in [DecodeEncode.v](https://github.com/mit-plv/riscv-coq/blob/21993a3401647c7bc6da9b6f7bd0fa978d61001b/src/proofs/DecodeEncode.v#L267), but that one is critical to keep the development "round trip time" short)
And btw, can't you just increase the RAM limit?

samuelgruetter:
> I wonder if it could be disabled just for the CI.

This would be bad:
*    It makes the claim "it builds locally if and only if it builds on CI" less true
*    It complicates the Makefile
*    It means we miss an opportunity to have a bigger-than-toy-example regression test for Coq's `par:` feature

> How can we do that? As far as I can tell runners come with a defined amount of RAM, which is usually enough for the whole test-suite [which is not precisely a toy these days]

According to the [CircleCI docs](https://circleci.com/docs/2.0/configuration-reference/#resource_class), this is possible.

I want to use Coq as a tool to do research on what will be possible in software verification in a few years from now. I hope that by then, Coq's memory usage will be lower, and RAM will be even cheaper than now. Therefore, we should really not be held back by low RAM limits today!

ejgallego:
@samuelgruetter a few points:
- first, we are not using CircleCI, but GitLab free runners, these come with those limitations,
[...]

PR #8546, opened 2018-09-24
[ci] Allow bedrock to fail.

ejgallego:
Until #8533 is mitigated.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -348,6 +348,7 @@ validate:edge+flambda:
 
 ci-bedrock2:
   <<: *ci-template
+  allow_failure: true
 
 ci-bignums:
   <<: *ci-template


SkySkimmer:
@Zimmi48 does the bot put a failing status if an allow_failure job fails?

ejgallego:
> @Zimmi48 does the bot put a failing status if an allow_failure job fails?

Indeed it seems we need to improve the bot here.

ejgallego:
https://github.com/coq/bot/issues/22

Zimmi48:
Thanks for opening the issue. I'll look into what is possible to do (whether I can support this generically or I have to do a special tweak).

ejgallego:
Thanks @zimmi48

Zimmi48:
I haven't yet looked into how @coqbot can support allow_fail because it seemed that there was a way to reduce the number of workers, but I can still do that.

SkySkimmer:
I think there's no point merging this until the bot supports allow_failure. Either we live with the failures or we disable the job entirely.

ejgallego:
> I think there's no point merging this until the bot supports allow_failure. Either we live with the failures or we disable the job entirely.

@SkySkimmer note that the PR has been updated to fully disable Bedrock. IMHO I would ping the folks there and merge in a couple of days.

ejgallego:
Correcting the tag.

Zimmi48:
I'm adding support for allow_failure in @coqbot right now.

ejgallego:
Thanks @Zimmi48 , but I don't see the point in keeping bedrock as is now. We will welcome it back when upstream updates their branch to fix #8533

SkySkimmer (approved):

Zimmi48:
I haven't seen any recent failure of bedrock, to the point I thought this PR had already been merged (confirmed by looking at 10 pages of most recent finished jobs).

ejgallego:
Indeed I was wondering too, let me investigate and I'll come back to you.

ejgallego:
Ok, so far indeed the reason is that all runners taking bedrock are indeed of the GCE type, they have more memory and the build works.

However, we cannot be sure all runners are going to be of this kind, and mitigation of #8533 has not happened. So I suggest to wait, but upstream should really perform some mitigation.

ejgallego:
Let's close this for now then.

ejgallego:
Reopened. Let's test the coqbot feature then.

Zimmi48:
Unfortunately bedrock2 succeeded :laughing: 
I've restarted the bedrock2 build to see what happens when it fails.

SkySkimmer:
Looks like it succeeded again.

Zimmi48:
Let's wait until we see it failing again in another PR.

ejgallego:
Plenty of recent failures, for example:
- https://gitlab.com/coq/coq/pipelines/32302860
- https://gitlab.com/coq/coq/pipelines/32307809

Zimmi48:
OK then let's try again to see a failure here.

ejgallego:
> OK then let's try again to see a failure here.

I just cancel as it was running on pyrolise. Actually we can predict if its going to fail from the first line of the log, so I will restart until we get the right worker.

ejgallego:
Ok, it is running on a "bad runner" now. [took 3 restarts] let's see...

https://gitlab.com/coq/coq/-/jobs/105829111

ejgallego:
Worked beautifully! I wonder if the github API allows for a kind of "yellow" state [passed with warnings]

I guess @Zimmi48  is aware of this , but we have the "checks" tab that for sure could provide a better overview.

Zimmi48:
Yeah, the checks tab (see very basic example here: https://github.com/coq-community/aac-tactics/pull/11/checks) opens possibilities for detailed reporting. We could explore doing this with our GitLab CI results (through the bot).

JasonGross:
Does bedrock2 not build with -j1 and replacing `par` with `all`?

Zimmi48:
>We could explore doing this with our GitLab CI results (through the bot).

But this checks feature requires going through a GitHub App (which @coqbot isn't yet) which makes it harder and not worth exploring if there isn't more at stake (such as other reasons to go to a GitHub App form).

SkySkimmer:
So do I merge?

ejgallego:
> So do I merge?

Lets wait a bit more, it turns out there could be a bug in `coqc`.

SkySkimmer:
I keep seeing bedrock die, so merging.

SkySkimmer:
The merge script completely died btw:
~~~
info: title for PR 8546 is null 
info: PR 8546 targets branch null 
info: you are merging in master 
info: remote for master is upstream 
info: fetching from upstream the PR 
Fetching upstream
From github.com:coq/coq
 * branch                  refs/pull/8546/head -> FETCH_HEAD
info: commit for PR 8546 is f76664af6b193193e83c00721ea2fd288e193154 
error: PR requests merge in null but you are merging in master 
Continue anyway? [y/N] y
fatal: bad revision 'null'
warning: This PR didn't run with the latest version of CI. 
warning: It is probably a good idea to ask for a rebase. 
Do you want to see the diff? [Y/n] n
Continue anyway? [y/N] y
jq: error (at <stdin>:4): Cannot iterate over null (null)
error: CI unsuccessful on  
Continue anyway? [y/N] y
jq: error (at <stdin>:4): Cannot iterate over null (null)
~~~

Zimmi48:
Looks like the merge script is not ready to handle any trouble on the GitHub API side...

silene:
That is why I have kept it around until now, but since it has never failed even once, there is no point anymore. Did anyone investigate what commit broke bedrock on master?

ejgallego:
> Did anyone investigate what commit broke bedrock on master?

We have a good understanding of the situation in #8533 ; the problem is not a particular commit, but that sometimes the CI system get workers with less resources so bedrock cannot properly build.

This is due to their use of the `par: ` tactic; they tried to limit Coq's resource use however there is a bug in `coqc` that doesn't allow them to pass the right argument.

I tried to fix the bug but IMHO the proper solution is to split `coqc/coqtop` as we discussed today, so obviously not 8.9 material, etc...

I think the current status is fine.

silene:
>  the problem is not a particular commit, but that sometimes the CI system get workers with less resources so bedrock cannot properly build.

I have a hard time believing that theory. Let us call probability theory to the rescue. If the failure is unrelated to the branch, then we can reasonably assume that getting a defective runner follows a binomial distribution. On the last 50 runs of the bedrock test on master and related branches, there was 8 failures. By virtue of a binomial distribution, this means that the number of defective runners is more than 2.7% with a 99% confidence. That is a low probability, yet it is high enough so that the test should have failed on 8.9 and related branches with a high probability already.  It never did.

Zimmi48:
Interesting info! Given that @ejgallego has identified which runners make Bedrock run out of memory (and GitLab gives this info for each job), we can further confirm this theory by finding Bedrock jobs on v8.9 which did not run out of memory even though they were running on the problematic runner. Then we could properly label this issue as an 8.10 regression.

ejgallego:
I'd dare to say that we are fairly confident on our analysis of the problem. Maybe 8.9 is testing a too old bedrock2 version where not enough `par: ` calls have been added?

Also, let me note that I'm afraid that a simple probabilistic analysis will not work for the problem or worker selection as for one, it is highly correlated with load and other complex constraints.

PR #8550, opened 2018-09-24
Don't use dune-template for apidoc

SkySkimmer:
dune-template works for the build jobs but followup jobs are different
enough to make reuse more confusing than useful IMO.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -288,13 +288,12 @@ doc:ml-api:ocamldoc:
       - dev/ocamldoc
 
 doc:ml-api:odoc:
-  <<: *dune-template
   stage: test
   dependencies:
     - build:egde:dune:dev
+  script: make -f Makefile.dune apidoc
   variables:
     OPAM_SWITCH: edge
-    DUNE_TARGET: apidoc
   artifacts:
     name: "$CI_JOB_NAME"
     paths:


ejgallego:
Well, dune-template is just a placeholder indeed until we develop a better system, but on the other hand I this PR has 0 effect, doesn't it?

SkySkimmer:
>Well, dune-template is just a placeholder indeed until we develop a better system

??

>this PR has 0 effect, doesn't it?

It's a cleanup, we replace importing the template just to override almost all fields by a simple key definition.

ejgallego:
> ??

In the future we will directly call dune instead of a makefile, and we may even have only one target, I dunno. Certainly the dune implementation, while functional, is far from definitive; more details should arrive soon but until Dune 1.4 / 1.5 we won't see something resembling full Coq support IMHO.

I put Dune in the CI for two primary reasons:
- to ensure integrity of the library boundaries [and it has already worked pretty well for that]
- to ensure that odoc can keep building the API docs.

> It's a cleanup, we replace importing the template just to override almost all fields by a simple key definition.

Yeah, no functional change I mean, so I was wondering about the motivation, the change seems like just a detail to me.

But sure, if you want this in I'll merge.

ejgallego (approved):

PR #8555, opened 2018-09-25
Remove section paths from kernel names

maximedenes:
Let's see what CI says.

API documentation needs to be updated, and `dev/doc/changes.md` to mention a few renamings.


ppedrot:
Looks broken...

gares:
you need to fix dev/*ml stuff to :-/

herbelin:
Can you explain your goals? Is it intended to be a purely internal cleaning or with a change of semantics in mind? For instance, does this have an impact in what `About` reports for constants in sections?

maximedenes:
> Can you explain your goals? Is it intended to be a purely internal cleaning or with a change of semantics in mind? For instance, does this have an impact in what `About` reports for constants in sections?

It is not fully internal, because it prevents for example:
```coq
Ltac f := idtac.
Section foo. Ltac f := idtac. End foo.
```

which was previously allowed. I didn't find an occurrence, except one in the test suite. But overall, it does not change much the semantics of names at the user level.

After this patch, `About` still reports the same thing, and you can still qualify constants using a section path, since this relies on the `full_path` / `Nametab` thing, not the kernel names that this patch cleans.

However, I was very surprised to discover one can qualify using section paths (and not just module paths), I'm not sure how aware users are of this feature. We may want to discuss it separately.

maximedenes:
> Can you explain your goals?

So the answer is I'm trying to clean (parts of) the section-related code, and save some memory.

kernel/names.mli
```diff
@@ -317,15 +316,12 @@ sig
...
-  val make3 : ModPath.t -> DirPath.t -> Label.t -> t
-  (** Shortcut for [(make1 (KerName.make ...))] *)
-
   (** Projections *)
 
   val user : t -> KerName.t
   val canonical : t -> KerName.t
 
-  val repr3 : t -> ModPath.t * DirPath.t * Label.t
+  val repr2 : t -> ModPath.t * Label.t
```
ppedrot: Should probably be called `repr`.

kernel/term_typing.ml
```diff
@@ -539,8 +539,8 @@ let translate_local_assum env t =
 let translate_recipe env kn r =
   (** We only hashcons the term when outside of a section, otherwise this would
       be useless. It is detected by the dirpath of the constant being empty. *)
-  let (_, dir, _) = Constant.repr3 kn in
-  let hcons = DirPath.is_empty dir in
+(*  let (_, dir, _) = Constant.repr3 kn in *)
+  let hcons = false (* XXX DirPath.is_empty dir *) in
```
ppedrot: Isn't that preventing section definitions from being hashconsed when they exit it?

ppedrot (approved):
I am supporting this change. It makes the kernel cleaner and reduces many places where we can break invariants while not removing any user-facing feature.

library/keys.ml
```diff
@@ -92,8 +92,7 @@ let subst_keys (subst,(k,k')) =
   (subst_key subst k, subst_key subst k')
 
 let discharge_key = function
-  | KGlob g when Lib.is_in_section g ->
-    if isVarRef g then None else Some (KGlob (pop_global_reference g))
+  | KGlob (VarRef _ as g) when Lib.is_in_section g -> None
```
SkySkimmer: Is there still a need for the `when`?

kernel/term_typing.ml
SkySkimmer: In fact now that we don't have to change the names I think we could just hashcons when the `cooking_info` is empty.

ppedrot: Indeed.

herbelin:
> After this patch, About still reports the same thing, and you can still qualify constants using a section path, since this relies on the `full_path` / `Nametab` thing, not the kernel names that this patch cleans.

OK, I think I now understand.

One reason I'm worrying is that I feel that there is a general confusion about where to go with discharge and with overwriting the name of section variables in goal, a sea snake which blocks us to make progresses (e.g. with `Proof using`, with treating the copy of section variables in goal as goal variables, etc.). I'd like to hear a vision on this latter point, knowing that in the absence of a vision, I shall propose to have the discharge in the kernel and the section variables locally-global objects.

maximedenes:
> One reason I'm worrying is that I feel that there is a general confusion about where to go with discharge and with overwriting the name of section variables in goal, a sea snake which blocks us to make progresses (e.g. with `Proof using`, with treating the copy of section variables in goal as goal variables, etc.). I'd like to hear a vision on this latter point, knowing that in the absence of a vision, I shall propose to have the discharge in the kernel and the section variables locally-global objects.

I'm pretty sure discussion on these points is needed indeed, and you can open an issue about it. But as far as I can tell, this patch is completely orthogonal to these questions.

kernel/names.mli
maximedenes: I doubt it, because `repr` should return the two names...

kernel/term_typing.ml
maximedenes: > Isn't that preventing section definitions from being hashconsed when they exit it?

Yeah, I completely forgot to come back to this part of the code, sorry.

maximedenes: > Isn't that preventing section definitions from being hashconsed when they exit it?

In fact, I don't think so because the only call to `translate_recipe` already hash conses everything (body, type, universes)...

Can you confirm?

library/keys.ml
maximedenes: I don't know, the logic is pretty mysterious to me, so I kept the behavior of the original code.

maximedenes:
This is blocking on a fiat-crypto-legacy failure, but building it on my machine takes ~2h before I see the first `coqc` invocations...

gares (approved):

maximedenes:
All green! @ppedrot can you be the assignee and merge? Thanks!

ppedrot:
Will merge soon. Last round!

maximedenes:
Any particular reason for delaying this more?

ejgallego:
Well, for one I find the commit message way too bare for what amounts to a critical kernel change. I guess changes to the kernel of this calibre should detail in the commit log:
- why is the change needed?
- how things used to work before?
- how do the work now?
- what is an example of a difference in behavior.

maximedenes:
> why is the change needed?

It is not needed.

> how things used to work before?

The section paths were essentially unused.

> how do the work now?

The same way as before, except for the example documented in `CHANGES`.

> what is an example of a difference in behavior.

See `CHANGES`.

I feel like you are asking me to repeat in the commit message what I put in the docs. Is that correct? If so, I can do it.

ejgallego:
> It is not needed.

It is a cleanup; then you should say so. It makes the code simpler.
 
> The section paths were essentially unused.

Good, state what "essentially" means, like "they were used only in this case"
 
> The same way as before, except for the example documented in `CHANGES`.

CHANGES is user-facing, I am talking about dev-facing here.

> I feel like you are asking me to repeat in the commit message what I put in the docs. Is that correct? If so, I can do it.

I had read the docs before submitting my comment. So far in `/dev/doc/changes.md` they only statement is "kernel names don't take a section path" but IMHO rationale is missing.

Indeed, I do think that a commit message should give me a good understanding of the change, specially for the kernel, I dunno what others things. To me, with the current message, I have to do a lot of work to understand what this change is about and what the implications are.

maximedenes:
> Good, state what "essentially" means, like "they were used only in this case"

Ok, you're asking me to describe the previous design but I failed to understand it, so I'll let you amend the commit message in the way you like. I anyway exhausted the time I was planning to spend on this patch, and as mentioned above, I don't need it at all.

ejgallego:
> Ok, you're asking me to describe the previous design but I failed to understand it

Does that mean that you don't understand what this patch does?

> I'll let you amend the commit message in the way you like.

I am very puzzled with this statement, which does seem wrong at several levels.

If you, the author of the patch, and maintainer of the subsystem in question, has deep trouble writing a short paragraph that documents the change for those doing `git` diving in the future, how am I supposed to come up with a description, precisely that I have no idea of what the patch does, nor what it does imply nor how does the kernel work.

maximedenes:
> Does that mean that you don't understand what this patch does?

Exactly.

ppedrot:
Well, it's pretty clear: there is legacy code in the kernel that is essentially dead in a non-trivial way. So it doesn't hurt removing it, it makes the kernel more robust and reduces its attack surface.

gares:
> I dunno what others things

I'm fine with this patch.

ppedrot:
@ejgallego are you reassured and OK for the PR?

ejgallego:
Sure, I have nothing about the PR, I just said that the commit message could be a bit better than empty IMO. Give me a few minutes and I'll prepare a commit message then, given that I am supposed to do it, it seems.

gares:
> Give me a few minutes

hem...

ejgallego:
> > Give me a few minutes
> 
> hem...

Can you explain please what is the objective / purpose of this comment?

ejgallego:
Sorry for the delay, here it goes:

```
[kernel] Remove section paths from `KerName.t`

We remove sections paths from kernel names. This is a cleanup as most of the times this information was unused. This implies a change in the Kernel API and small user visible changes with regards to tactic qualification. In particular, the removal of "global discharge" implies a large cleanup of code.

Additionally, the change implies that some machinery in `library` and `safe_typing` must now take an `~in_section` parameter, as to provide the information whether a section is open or not.
```

maximedenes:
> Sorry for the delay, here it goes:

Thanks, commit updated.

ppedrot:
CI failures are spurious, right? Iris is broken in master already...

ejgallego:
> CI failures are spurious, right? Iris is broken in master already...

Yup, CI here is good. However, this got bitten by the `CHANGES` disease so a new rebase seems to be needed.

maximedenes:
lol

gares:
No rebase, the guy that merges should take care of the conflict.

ejgallego:
I am confused or was this PR merged without the overlays having been properly submitted as PR @ppedrot ?

Anyways, CI in master is broken so please @maximedenes submit the corresponding PRs and @mattam82  @ppedrot merge, thanks.

ejgallego:
Also not sure about the deprecation policy followed here.

ppedrot:
> I am confused or was this PR merged without the overlays having been properly submitted

Strange, the merge script didn't warn me that there were overlays, and I forgot to check that there were there indeed. I can merge the Ltac2 one quickly, and submit one in equations.

> Also not sure about the deprecation policy followed here.

API changed in a non-backward compatible way, I don't think we can easily provide a deprecated interface.

ppedrot:
PR submitted on [equations](https://github.com/mattam82/Coq-Equations/pull/138) and already merged in Ltac2.

maximedenes:
> > Also not sure about the deprecation policy followed here.
> 
> API changed in a non-backward compatible way, I don't think we can easily provide a deprecated interface.

Yep, I did provide one deprecation warning for the construct for which I could see a transition path.

PR #6705, opened 2018-02-07
[ssr] extended intro patterns

gares:
# extended intro patterns

This PR implements the following intro patterns for ssr:
  - temporary introduction: `=> +`
  - block introduction: `=> [^ prefix ] [^~ suffix ]`
  - fast introduction: `=> >H`
  - tactics as views: `=> /ltac:mytac`

See the reference manual for the actual documentation.

## Status

- [x] Corresponding documentation was added / updated.
- [x] Entry added in CHANGES.
- [x] Check all bugs reported by Cyril are fixed

PR #8556, opened 2018-09-25
[ssr] use tclDISPATCH for IPatDispatch (fix #8544)

### LINKED ISSUE #8544: [ssr] dispatch intro pattern is too permisive
#### Version
 ```
The Coq Proof Assistant, version 8.8.1 (September 2018)
compiled on Sep 1 2018 9:56:03 with OCaml 4.06.1
```

#### Operating system
```Linux [...] 4.14.67 [...]-NixOS SMP Fri Aug 24 11:09:23 UTC 2018 x86_64 GNU/Linux```

The following should fail because there should be two members in the dispatch.
Now the intro pattern between brackets is applied to the first subgoal.
```coq
Require Import ssreflect.
Goal True \/ True -> False.
case => [a].
```

gares:


gares:
@silene I'd really like to have this PR in 8.9. It restores the check that `=> [....|....]` has the right number of branches (when `[]` is the first ipat after a `case` or `elim`). The bug was introduces in 8.8 I believe. I don't know if it qualifies as critical/blocking, since the language is fully functional. Hence I'd like your opinion on setting a severity and getting this code in the release.

To understand the bug/patch: ssr has an exception on `[]`, when it is the first ipat after case or elim it does not perform a new case, but just dispatches. When intro patterns were ported to the tactic monad some machinery for #6705 was put in place, eg the AST (bot not the parsing rules) for an intro pattern that dispatches. The exception above was implemented by rewriting `IPatCase` to `IPatDispatch`, but the implementation of dispatch was to not check the number of goals (a bug, my bad).
On top of that there is another exception (see `tclIORPAT`) that says that `=> .. [] ..` does not fail (eg, you just case and don't check that the number of goals matches the number of branches). If `[]` id the first one, then I map it to `IpatDispatch` and I set the mode to "non strict" (this is the boolean). In #6705 the dispatching ipat will get a strict mode when being parsed.
Hope it helps.

plugins/ssr/ssripats.ml
```diff
@@ -215,8 +215,11 @@ let rec ipat_tac1 ipat : unit tactic =
   | IPatView (clear_if_id,l) ->
       Ssrview.tclIPAT_VIEWS ~views:l ~clear_if_id
         ~conclusion:(fun ~to_clear:clr -> intro_clear clr)
-  | IPatDispatch ipatss ->
-      tclEXTEND (List.map ipat_tac ipatss) (tclUNIT ()) []
+
+  | IPatDispatch(false,[[]]) ->
+      tclUNIT ()
+  | IPatDispatch(_,ipatss) ->
```
maximedenes: This seems wrong. Why doesn't the behavior depend on the value of the "strict" flag?

gares: Hum, I guess the "name" of the boolean is misleading.
My intention is to distinguish Dispatch (number branches = number goal) and DispatchWhenUsedAsFirstCase (number branches = number goals OR number branches = 0).

So strict means "disable the second part of the OR".

I believe the code is correct. Suggestion on how to make it more readable are welcome.

gares: Another point here is: if ssr first case exception than I map Case -> Dispatch, but there are scripts (old ones) where one does `case=> [] []` and the first case is a no-op, today I'd write this `case=> - []`.
This prevents compiling the first `[]` as Dispatch, since Dispatch is strict now.
I did not check how many occurrences are there. We may fix the scripts and simplify this patch.

maximedenes: Can you change the name to make the code slightly more understandable. I'm afraid it looks very messy.

gares: Are you asking to have IPatDispatchLoose and IPatDispatchStrict ? Not sure I get what you want me to change?

gares: @maximedenes ping

maximedenes: I meant to change the comment saying the boolean means "strict". But in fact, I think comments with what you wrote above here should make it easy enough to understand the code.

maximedenes:
@gares as mentioned above, could you comment the code (at least with the explanation you wrote in the PR). Otherwise I'm afraid it will be impossible to follow what it does when we try to clean it. If you are at the office this afternoon, we can discuss it if you like.

gares:
hope it is OK now

plugins/ssr/ssrast.mli
```diff
@@ -84,7 +84,7 @@ type ssripat =
   | IPatId of (*TODO id_mod option * *) Id.t
   | IPatAnon of anon_iter (* inaccessible name *)
 (* TODO  | IPatClearMark *)
-  | IPatDispatch of ssripatss (* /[..|..] *)
+  | IPatDispatch of bool (* check number of branches and number of goals match *) * ssripatss (* (..|..) *)
```
maximedenes: I'm lost. I thought the discussion below showed that this was not the real meaning of the boolean. Does `false` mean we don't check the number of branches and number of goals match?

gares:
we will eventually get there...

gares:
@maximedenes green flag from CI

maximedenes:
@gares rebased and clarified the meaning of the boolean. Does it look good to you? The parsing code will need to be updated to put `false` instead of `true`.

maximedenes (approved):

gares:
thanks

PR #8559, opened 2018-09-25
[api] Two more missing deprecations.

ejgallego:
We missed two more constructor / record deprecation.

### DIFF
--- a/vernac/vernacexpr.ml
+++ b/vernac/vernacexpr.ml
@@ -114,8 +114,8 @@ type hint_mode = Hints.hint_mode =
 [@@ocaml.deprecated "Please use [Hints.hint_mode]"]
 
 type 'a hint_info_gen = 'a Typeclasses.hint_info_gen =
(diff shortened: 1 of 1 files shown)


ejgallego:
Also this is trivial enough that some good samaritan such as @ppedrot may want to take it.

ppedrot (approved):
Trivial, will merge promptly if no objection.

ejgallego:
Merci !

ejgallego:
Merci !

silene:
Does that even work? As far as I know, OCaml ignores deprecation attributes on record fields, as can be seen on the following example:
```ocaml
type t = { x : int; [@ocaml.deprecated] }
let f () = let ({ x } as y) = { x = 1 } in x + y.x
```

ejgallego:
Indeed, unfortunately for the record fields we may be out of luck, at least until 4.08. So that part of the PR is basically documentation.

Issue #7055, opened 2018-03-22
CoqIDE will be soon impossible to build in Debian derivatives


ejgallego:
The GTK libraries CoqIDE depends will be soon removed from Debian. This means CoqIDE will become much harder to build in these distributions.

For more details, see: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=885677

gares:
Thanks for the heads up.
Maybe it is time to really think about UIs ;-)

Zimmi48:
So what about choosing to drop CoqIDE completely and make VSCode with VSCoq the main interface for people who don't like Emacs?

I didn't personally tested VSCode yet (that's on my TODO list) but it has become the most popular editor among developers according to StackOverflow's 2018 survey (cf. https://insights.stackoverflow.com/survey/2018#technology-most-popular-development-environments).

Moreover, I think it would be saner that the Coq development team stop maintaining, building and distributing (with headaches) an editor on its own. We should focus on what we do best: Coq.

cc @MSoegtropIMC who proposed it on Coq-Club (or Coqdev, I can't remember)
and @siegebell who is the author of VSCoq

PS: @ejgallego you should read the description of the "priority: critical" label (just hover on it). Did you find a proof of False?

maximedenes:
One thing I wanted to explore is to develop first class support for LSP (https://langserver.org/) on the Coq side.

ppedrot:
It was pretty clear this would happen someday... As mentioned in the Debian thread, we're not the only ones in trouble here, we might want to ask LablGTK devs what they believe to be a decent solution (@garrigue ?). Maybe this simply means dropping LablGTK altogether?

garrigue:
Last time we had a discussion on the lablgtk list, the feeling was that gtk3 itself was not an interesting target.
So the question is whether there is sufficient compatibility support in gtk3 so that we could move at little cost.
I haven't checked yet.
If the required effort is the same as moving from gtk1 to gtk2, I'm not sure anybody is ready to do it.

ejgallego:
The best option is without any doubt to port lablgtk to gtk3.

LSP by itself is far from being useful to Coq.

maximedenes:
> LSP by itself is far from being useful to Coq.

Can you say about more?

Zimmi48:
> The best option is without any doubt to port lablgtk to gtk3.

Hum. Are you willing to do it? If there is no volunteer to do it, this is ridiculous to pretend it's the best option.

[...]

PR #8560, opened 2018-09-26
Unicode bindings for CoqIDE that works out of the box

charguer:
**Kind:** infrastructure.

This pull request aims at providing integrated support for input of unicode characters in CoqIDE.

System-wide support and GTK-support for Unicode input do exist, however there are not sufficiently easy to install and configure by students who just want to take on course on Coq. At the moment, due to this problem, Software Foundations (SF) is using ASCII only, but this restriction is not satisfying as it results in rather ugly scripts. As SF evolves and grows in popularity and contents, there is demand for unicode support that works out of the box for students. Here it is.

I am proposing a relatively lightweight patch, with the following features : 
- A shortcut (by default "Shift+Space") parses the word between the cursor and the immediately-preceding backslash, and replaces it with a corresponding unicode symbol. For example "\alpha" gets converted to the greek letter, and in fact typing just "\a" then Shift+Space suffices.
- If no corresponding binding applies, a space is inserted.
- A default table of latex-to-unicode bindings is provided, but it can be easily customized by adding bindings either in the configuration file directly, or by having a local (e.g. per-project) file with additional bindings (in which case this filename is passed as argument to CoqIDE using "-unicode-bindings f", to either augment or replace the default file). 
- Priorities can be assigned to bindings in order to determine which token to consider when only a prefix is provided, e.g. to ensure that \a goes to \alpha rather than \ast.

Question:
- I do not know what's the right way to get CoqIDE to copy the default bindings files into the .config/coq folder when such binding file does not exist yet. Should the contents of that file be distributed in /usr/lib? or be part of the CoqIDE binary? I could use some help there.

Remark:
- Corresponding documentation was NOT YET added, I first want to get feedback on this pull request.
- Entry was NOT YET added in CHANGES, for similar reasons.


ppedrot:
> System-wide support and GTK-support for Unicode input do exist, however there are not sufficiently easy to install and configure by students who just want to take on course on Coq

Why isn't it possible to provide a bundle with a decent IME for those users? I am personally using ibus-latex-tables and I am very happy about it. Reinventing the wheel has often nasty consequences.

Also, note that it is unluckily possible that CoqIDE is going to die out soonish, due to the lack of support for lablgtk on e.g. Debian, see #7055, in which case this endeavour will be moot.

Zimmi48:
This PR was made on an outdated version of master (and @coqbot apparently didn't detect it). Please rebase.

herbelin:
Hi, I just tried the feature (on MacOS X). This is absolutely great. It is indeed particularly light-weight and shall save my time.

The rebasing shall be trivial (just a `Flags.ideslave_coqtop_flags` to be changed into `Coq.ideslave_coqtop_flags`).

I shall try to give an answer later on the policy regarding the configuration file.

herbelin:
> I do not know what's the right way to get CoqIDE to copy the default bindings files into the .config/coq folder when such binding file does not exist yet. Should the contents of that file be distributed in /usr/lib? or be part of the CoqIDE binary? I could use some help there.

The default `coqide.keys` file is installed in `$(FULLCONFIGDIR)` (in the case of MacOS X, see `Makefile.ide`) and it seems consistent to add a similar line there for the default bindings file. (I tentatively submitted PR #8567 for generalizing this principle to all platforms.)

Otherwise, if following the policy for other configuration files is ok for you, I would suggest to replace the error by an `Ideutils.flash_info` when the binding file is not found.

Also, I would have better seen the `LaTeX-to-unicode` item in the Edit menu (next to `Complete Word`) but I don't know so much the usages.

Finally, since we already have the spelling `LaTeX` in the menu `File`/`Export`, better use the same spelling for `LaTeX-to-unicode`.

charguer:
> The default `coqide.keys` file is installed in `$(FULLCONFIGDIR)`

My understanding is that the contents of `coqide.keys` is generated, in other words it is not the result of a copy-paste from an existing file (am i correct?)

> Otherwise, if following the policy for other configuration files is ok for you, I would suggest to replace the error by an `Ideutils.flash_info` when the binding file is not found.

My suggested policy is the following:
- if the unicode bindings config  file does not exist, then coqide should install one
- if the user pass specific arguments to -unicode-bindings then errors should be reported in case of missing flags (not warnings).

> Also, I would have better seen the `LaTeX-to-unicode` item in the Edit menu (next to `Complete Word`) but I don't know so much the usages. Finally, since we already have the spelling `LaTeX` in the menu `File`/`Export`, better use the same spelling for `LaTeX-to-unicode`.

Ok for both.

herbelin:
> My understanding is that the contents of `coqide.keys` is generated, in other words it is not the result of a copy-paste from an existing file (am i correct?)

Except on MacOS where, if the `coqide.keys` does not exist in the config path of the user, it is loaded from a default file bundled with the distribution of Coq. Then #8567 is generalizing this to other platforms. Without #8567, indeed, on non-MacOS, no default file exists and the contents is generated by Gtk (normally with every possible option prefixed by `;` and thus not being modified).

> My suggested policy is the following:
> - if the unicode bindings config file does not exist, then coqide should install one
> - if the user pass specific arguments to -unicode-bindings then errors should be reported in case of missing flags (not warnings).

OK.

ejgallego:
Installing stuff seems like a bad idea, as it will force you to provide an upgrade path for users if such a mechanism needs changes in the future, and worse, you won't be able to distinguish if the file was auto-installed or not.

I suggest you just load a default file if the user file is not present.

charguer:
If we follow @ejgallego suggestion, then the semantics would be:
- if there is no file and no cmdline option passed, then use the default file distributed with the package
- if using "-unicode foo.bindings", then use only foo.bindings
- if using "-unicode foo.bindings,default", then use only foo.bindings in addition to default bindings.
I think I like that.

rudi-cilibrasi:
regarding config file replacement, it is usual e.g. in Debian to keep a history of the hashes of the contents of all previously distributed config files to be able to detect if the user has done any custom modifications or not. see e.g. https://manpages.debian.org/jessie/ucf/ucf.1.en.html   c.f.    ```  Specifically, questions should not be asked if the file in question is an unmodified version that was one shipped in a previous version of this package; and the maintainer can help by telling the script about the historical md5sums that published versions of this file contained.
The way to do this is to either create a file called <New file>.md5sum, with one md5sum on each line, (the file names you use are ignored, except for the entry named default), or create a directory, called <New file>.md5sum.d, which should contain any number of files, each containing a single line, namely, the md5sum of a previous version of <New file>. The names of these files are not important, with one exception: The file called default is treated specially. For example, the author personally uses either package version numbers or release code names, like 7.6.3, or potato. If none of the historical md5sums match, we are almost certain that either the historical record of md5sums is not complete, or the user has changed the configuration file.
 ```

ejgallego:
I am not sure what UCF has to do in this discussion, but certainly we as upstream should not care at all about it, and of course we should not require the user to modify system files in order to achieve a fully functioning Coq.

ppedrot (approved):
I am not convinced by the fact CoqIDE should be responsible for this feature, but the code looks OK to me, thus it can go in.

ejgallego:
> I am not sure what UCF has to do in this discussion, but certainly we as upstream should not care at all about it, and of course we should not require the user to modify system files in order to achieve a fully functioning Coq.

@Zimmi48 let me try to explain more. UCF (as pointed by @rudi-cilibrasi ) is a system used by Debian / Ubuntu to manage files in `/etc`. So for example, when the user modified `/etc/apache2/base.conf` which is also shipped by the `apache2` package, UCF helps in reconciling the changes.

UCF doesn't deal with user files. So what I mean in the comment is that `CoqIDE` should not require the user to modify things in `/etc`.

charguer:
Here is an updated proposal for the semantics of the default binding file:
- If an option "-unicode-bindings"  is passed with a list of files, then used the concatenation of the contents of these files. Recognize "default" as a special token that is an alias for the "default.bindings" file packaged with coqide.
- If no option is passed, then if a file "coqide.bindings"  exists in .config/coq, then use this one, else use "default.bindings" from the package.

Benefits:
- no installation of a config file is required upon installation
- if the user never put his own binding file, then he gets benefits of upgrades on new releases of coqide
- if the user has set up his own global binding file, then his file is never overwritten
- the user has the ability to easily exploit per-project additional bindings.

If I don't hear further complain about this plan, i'll go at implementing it.

gares:
I missed this PR but it was mentioned in the wiki page for the next WG. I just want to make you aware of a file we produced a long ago for Matita using various tables from latex, mathml, unicode:

    https://github.com/LPCIC/matita/blob/master/matita/matita/predefined_virtuals.ml

The file contains 3 things:
- a classification of symbols based on shape and/or usage (to present the data base to the user in a structured way)
- a mapping from \latex name or ASCII art (eg "->") to unicode
- a "home grown" collection of equivalence classes (the idea being that to type a funny arrow you can start by typing a simple arrow and the cycle trough the class, if you don't recall the exact name of the funny arrow)

FTR in matita there was no activation key, it was just space, eg typing `\forall ` one would get the flipped A symbol.

ppedrot:
@charguer This has been accepted at the WG. Can you rebase on a recent master for CI to run?

ppedrot:
@charguer ping.

charguer:
@ppedrot c'est les vacances !

ppedrot:
@charguer c'est plus les vacances! (at least here).

charguer:
I did the rebase (or at least, so I hope...).
I still need to implement the proposed behavior for storing the bindings files.
I also would like to see how to integrate Matita's bindings into the default file.

SkySkimmer:
Reset to 1f8cc0efd1 and try again.

charguer:
Sorry I don't know what you're asking me to do, precisely.

--If you're complaining about the "empty commit" (159ed81dbfc29d560287f0e1dd3ea4e60ff36833), I was asked by git to resolve a conflict because one line I edited was just next to a line that someone else edited, but there was nothing to do, so the diff was empty.

SkySkimmer:
I'm complaining about the 250+ new commits github is showing.
BTW I read the log graph wrong above, 770282e523 is the commit you should reset to and there's no need to rebase after that.
```

159ed81dbf *   merged
           |\  
540eb4d580 | *   Merge branch 'master' into coqide_unicode_bindings
           | |\  
1f8cc0efd1 | * | cosmetic changes
39520839e2 | * | support for coqide commande line arguments
74aa8fb445 | * | working set of bindings
d9c646001d | * | latex to unicode in coqide
770282e523 * | | cosmetic changes
11b5c6bb68 * | | support for coqide commande line arguments
67c64d13a3 * | | working set of bindings
496a21178e * | | latex to unicode in coqide
108c15b51a * | |   origin/master upstream/master Merge PR #8944: Revert PR #8923 (require camlp5 >=7.06)

```
the "cosmetic changes" which directly descends from master.

By reset I mean `git reset --hard`.

charguer:
Ok, thanks, i'll try the reset. And I will reuse @gares' bindings from Matita, they look great.

charguer:
I've made good progress. 
- There is a comprehensive list of bindings available by default. 
- The user can conveniently use custom bindings, in addition or in place of the default. 

 One last thing remains: we need the generated file `ide/default.bindings` to be installed by `make install-coqide` in the appropriate folder (which one?). @herbelin could you help me out for this?

gares:
This line is the one you should probably patch.
```
Makefile.ide:   $(INSTALLLIB) ide/coq.png ide/*.lang ide/coq_style.xml $@/coq/
```

charguer:
@maximedenes @gares @herbelin @jhjourdan @ejgallego I have implemented the installation of the default bindings file, and written the documentation. I think the PR is ready for final review and merge. Thanks.

charguer:
@herbelin: if you want the LaTeX-to-unicode item in the Edit menu, then you should also move the comment/uncomment commands to the "Edit" menu. And get rid of the nearly-empty "Tools" menu. This can be discussed later. But I think nobody really cares much anyway.

cpitclaudel:
Small comment: did you consider binding this to "TAB"? AFAICT, tab isn't used much except at the beginning of a line.

gares:
Is tab what PG uses for that?

herbelin:
@charguer: ok for postponing the menu question.

cpitclaudel:
> Is tab what PG uses for that?

No, Ctrl+RET (I think that binding came from my Visual Studio days).  In Emacs TAB reindents the current line, so it doesn't work very well to make it expand an abbreviation.   Plain Emacs uses M-/ for abbreviation expansion.

gares (approved):
I think it can be merged as such, but if you have time give a look at the changes I propose.

gares:
> No, Ctrl+RET

The I think we should consider using this combo and not TAB, it would help users of CoqIDE move to PG or VS, eventually. Why do you propose TAB?

ide/preferences.ml
```diff
@@ -1075,3 +1088,124 @@ let configure ?(apply=(fun () -> ())) () =
...
+  let selected_filenames = ref [] in
+  let add f =
+    selected_filenames := f::!selected_filenames in
+  if filenames = [] then begin
+    (* If no argument is provided using [-unicode-bindings],
+       then use the default file and the local file, if it exists *)
+    begin match get_bindings_default_file() with
+    | Some f -> add f
+    | None -> output_string stderr (Printf.sprintf "Warning: the file ide/default.bindings was not found in %s.\n" (Envars.coqlib()))
+    (* TODO: flush stderr does not seem to eagerly display the output message *)
```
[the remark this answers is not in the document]
charguer: Ok, I'm going to try and fix this

ide/wg_ScriptView.ml
```diff
@@ -405,6 +407,72 @@ object (self)
...
+        let prefix = backslash#get_text ~stop:insert in
+        let word =
+          match lookup prefix with
+          | None -> raise Abort
+          | Some word -> word
+          in
+        let was_deleted = buffer#delete_interactive ~start:backslash ~stop:insert () in
+        if not was_deleted then raise Abort;
+        word
+      with Abort -> " "
```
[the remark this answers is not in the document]
charguer: There is a good reason to put a space. If you write normal text, between two sentences you would press DOT then SPACE then a Capitalized letter. In french, the dot is accessed using SHIFT. Thus, many people would keep the SHIFT key down for the 3 keystrokes (DOT, SPACE, First letter of next sentence). Hence the need to insert a space.

Besides, even for users with english keyboards, it is good to know that your shift+space was taken into account, but did not match any registered binding, rather than giving the feeling that the keystroke is just ignored. 

I'll had these comments in the source to make sure they don't get lost.

charguer:
Regarding the choice of Shift+Space:
- Ctrl+Space is standard for autocompletion, e.g. in VS and Eclispse, so I don't want to bind it.
- Ctrl+Alt+Space (or Algr+Space) is introducing non-breakable spaces in Ubuntu, so I don't want to bind it.
- Tab is used for indenting, including blocks of lines, I would find it very confusing use it for modifying the text.

I wanted a shortcut in the same vein as autocompletion, that is very easy to type, and that wouldn't conflict. Shift+space appeared to me as the binding of choice.

charguer:
Ok done. To implement @gares suggestion to use GUI warnings, I had to move the bindings related stuff into a separate file. I think it's in fact a good thing, which I should have considered from the start. Please check final diff and merge.

cpitclaudel:
> Why do you propose TAB?

That's how most shells I use do completion

vbgl:
Has the generation of the bindings file been added to the `dune`-based build-system?

ppedrot:
I just had a sudden realization regarding an enhancement of the PR that makes the choice of hotkey moot: why don't we reuse the autocompletion feature of CoqIDE? My IME uses an interface that is very close to the autocompletion window, we could do the same. Essentially, when one starts writing `\` and a small prefix, a popup appears with the list of LaTeX / unicode characters available that start with that prefix.

WDYT? When the PR is merged I could open a new one implementing that change.

ppedrot:
@charguer If I'm right, this PR only needs a rebase before being merged.

doc/sphinx/practical-tools/coqide.rst
```diff
@@ -252,37 +253,66 @@ use antialiased fonts or not, by setting the environment variable
 Defining an input method for non-ASCII symbols
```
Zimmi48: You should probably change the title of the section as well.

doc/sphinx/practical-tools/coqide.rst
```diff
@@ -252,37 +253,66 @@ use antialiased fonts or not, by setting the environment variable
...
+For several symbols, ASCII art is also recognized, e.g. ``\->`` for a
+right arrow, or ``\>=`` for a greater than or equal sign.
+
+A larger number of latex tokens are supported by default. The full list
+is available here:
+  https://github.com/coq/coq/blob/master/ide/default_bindings_src.ml
+
+Custom bindings may be added, as explained further on.
+
+Remark: it remains possible to input non-ASCII symbols using system-wide
```
Zimmi48: Use `.. remark::`

doc/sphinx/practical-tools/coqide.rst
```diff
@@ -252,37 +253,66 @@ use antialiased fonts or not, by setting the environment variable
...
+approaches independent of |CoqIDE|.
+
+
+Adding custom bindings
+~~~~~~~~~~~~~~~~~~~~~~
+
+To extend the default set of bindings, create a file named `coqide.bindings`
```
Zimmi48:
```suggestion
To extend the default set of bindings, create a file named ``coqide.bindings``
```

Same problem on the next lines.

doc/sphinx/practical-tools/coqide.rst
```diff
@@ -252,37 +253,66 @@ use antialiased fonts or not, by setting the environment variable
...
-
-    $ export GTK_IM_MODULE=uim
-    $ coqide
-
-
-Activate the ELatin Input Method with Control-\\, then type the
-sequence `\\Gamma`. You will see the sequence being replaced by Γ as
-soon as you type the second "a".
-
-.. _character-encoding-saved-files:
```
Zimmi48: You remained too much. This last line should have been kept as it is part of the next section.

Zimmi48 (changes requested):

Zimmi48:
>Has the generation of the bindings file been added to the dune-based build-system?

This question hasn't been answered.

charguer:
Regarding the autocompletion pop-up, I personally find it very disruptive. If you like it, you can use the system wide latex input method, which features such a pop-up. If you like to reuse CoqIDE's box, you can implement it in addition to my patch. But I'd like to keep the option to work without a pop-up that raises on every other character that I type.

vbgl:
I can give a try to the `dune` related change. However, if the `dune` build is not broken, this will wait an other PR.

herbelin:
> Regarding the autocompletion pop-up, I personally find it very disruptive. If you like it, you can use the system wide latex input method, which features such a pop-up. If you like to reuse CoqIDE's box, you can implement it in addition to my patch. But I'd like to keep the option to work without a pop-up that raises on every other character that I type.

As far as I'm concerned, I'm fine without a pop-up too. An advantage I saw of displaying the possible completions was for the learning phase, until we know which bindings are available and what is the minimum number of letters we have to type before completion makes it be chosen by default.

charguer:
@vbgl the file `default.bindings` should go in `prefix/share/coq` (where is currently going coq.png and coq_style.xml). I added you as contributor to charguer/coq so that you can push your patches directly there in the branch `coqide_unicode_binding`. Many thanks for fixing dune support.

charguer:
@herbelin the question of how to help people learn the bindings is imo related to the question of "what is the best way to advertise for the file `ide/ default_bindings_src.ml` which contains the list of bindings in an organized way. One possibility is to register a link in the Help menu, like is done for the reference manual and the library documentation. I would suggest that the menu points to the section of the reference manual that deals with Coqide bindings, and ensure that on that page one cannot miss the link that points to the file with the list of default bindings. Thoughts?

vbgl:
I’ve just pushed my changes. Who should take care of the merge conflicts with `master` (how are we supposed to collaborate on a github PR)?

Zimmi48:
If I were you I would have rebased *before* adding your commit since it creates more conflict. It is still time to do it. GitHub keeps track nicely of commits before and after a force-push so that shouldn't be too much of a problem (as long as the one who did not force-push is ready to do a `git fetch` followed by a `git reset --hard`).

vbgl:
> If I were you I would have rebased _before_ adding your commit since it creates more conflict.

Thanks for the suggestion. But I’m not going to rebase everything and deal with conflicts in code that I’ve never heard about.

I’ve taken care of conflicts in `ide/dune`; Emilio’s commit should disappear when rebasing on `master`.

herbelin:
@charguer: It is not clear that I would have had the idea myself as a user to look at the help menu to know about bindings but that may indeed be a good default place to look for help when needing help. So, why not.

ppedrot:
I am a bit lost. What is missing for this PR?

gares:
I think the PR, but for the new conflicts, is OK.
@Zimmi48 do you agree?

doc/sphinx/practical-tools/coqide.rst
```diff
@@ -249,38 +250,75 @@ use antialiased fonts or not, by setting the environment variable
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+To extend the default set of bindings, create a file named ``coqide.bindings``
+and place it in the same folder as ``coqide.keys``. On Linux, this would be
+the folder ``~/.config/coq``. The file `coqide.bindings` should contain one
+binding per line, in the form ``\key value``, followed by an optional priority
+integer. (The key and value should not contain any space character.)
+Here is an example configuration file:
+
+```
```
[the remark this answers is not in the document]
Zimmi48: Reminder: this has to be fixed. This is Markdown syntax, not Sphinx.

doc/sphinx/practical-tools/coqide.rst
```diff
@@ -249,38 +250,75 @@ use antialiased fonts or not, by setting the environment variable
...
+It can be useful to work with per-project binding files. For this purpose
+|CoqIDE| accepts a command line argument of the form
+``-unicode-bindings file1,file2,...,fileN``.
+Each of the file tokens provided may consists of one of:
+- a path to a custom bindings file,
+- the token ``default``, which resolves to the default bindings file,
+- the token ``local``, which resolves to the `coqide.bindings` file
+  stored in the user configuration directory.
+
+.. remark::
```
Zimmi48: This could use `.. warning::` too.

Zimmi48 (approved):
There's still the problem I mentioned earlier with the doc and the CHANGES entry is missing.

Zimmi48:
The remaining "needs" flags reflect actions that actually need to be taken.

charguer:
Voilà, j'ai corrigé ce que @Zimmi48 m'a suggéré, merci.
Pour le rebase, je veux bien gérer les conflits, mais il faudra m'aider un peu pour la manip :)

CHANGES.md
```diff
@@ -66,6 +66,16 @@ Tools
 - A new `-bytecode-compiler` flag for `coqc` and `coqtop` controls whether
   conversion can use the VM. The default value is `yes`.
 
+- CoqIDE now supports input for Unicode characters. For example, typing
+  "\alpha" then the "Shift+Space" will insert the greek letter alpha.
+  In fact, typing the prefix string "\a" is sufficient suffices.
```
Zimmi48: Repetition in "sufficient suffices".

CHANGES.md
```diff
@@ -66,6 +66,16 @@ Tools
...
+  A larger number of default bindings are provided, following the latex
+  naming convention. Bindings can be customized, either globally, or on a
+  per-project basis, with the requirement is that keys must begin with a
+  backslash and contain no space character. Bindings may be assigned custom
+  priorities, so that prefixes resolve to the most convenient bindings.
+  The documentation pages for CoqIDE provides further details.
```
Zimmi48: The CHANGES entry should preferably be much shorter and refer to the reference manual for details.

doc/sphinx/practical-tools/coqide.rst
```diff
@@ -249,38 +250,78 @@ use antialiased fonts or not, by setting the environment variable
...
+binding per line, in the form ``\key value``, followed by an optional priority
+integer. (The key and value should not contain any space character.)
+
+.. example::
+
+   Here is an example configuration file:
+
+   ::
+
+   \par ||
```
Zimmi48: I think this should be indented two more characters compared to the `::`.

vbgl:
I took the liberty to rebase and address two of the comments by Théo. I hope that qualifies as “aider un peu pour la manip”.

charguer:
It's great, thanks a lot!
I will try to find time this week to compile and test the current code.

vbgl:
I think that this PR is ready. Are we waiting for something in particular?

ppedrot:
I don't think so, this looks ready to me too. The linter is complaining, is that a legit error though?

vbgl:
The linter complains because one commit adds an end-of-line byte at the end of a file. The touched line is unfortunately indented with tabs. Fixing the indentation in an unrelated part of the code might not belong to this PR.

SkySkimmer:
>The linter complains because one commit adds an end-of-line byte at the end of a file.

Actually one commit removes the end line, then another puts it back. So remove the spurious change in the first commit to make this a noop.

vbgl:
Done. Thanks for the suggestion.

PR #8561, opened 2018-09-26
Fix votour compilation after #8102.

ppedrot:
Quite urgent.

### DIFF
--- a/checker/votour.ml
+++ b/checker/votour.ml
@@ -152,6 +152,7 @@ let rec get_name ?(extra=false) = function
   |String -> "string"
   |Annot (s,v) -> s^"/"^get_name ~extra v
   |Dyn -> "<dynamic>"
+  | Proxy v -> get_name ~extra !v
 
 (** For tuples, its quite handy to display the inner 1st string (if any).
     Cf. [structure_body] for instance *)
@@ -255,6 +256,7 @@ let rec get_children v o pos = match v with
     | _ -> raise Exit
     end
   |Fail s -> raise Forbidden
+  | Proxy v -> get_children !v o pos
 
 let get_children v o pos =
   try get_children v o pos


ejgallego:
This is ready to merge, any volunteer?

JasonGross (dismissed):
Perhaps the CI should be updated to build votour, so this doesn't happen again?

ppedrot:
I am no CI specialist, so I'll defer that to somebody else.

ejgallego (approved):
CI passes, no idea about if the patch itself is correct.

SkySkimmer:
@JasonGross 
CI does test votour https://gitlab.com/coq/coq/-/jobs/101817168

SkySkimmer:
Although it's kinda in a fragile way it seems.

ejgallego:
> Although it's kinda in a fragile way it seems.

Why?

JasonGross:
I would suggest adding a line to the beginning of the `script` part of `test-suite-template` in `.gitlab.yml ` that says
```
- make -j "$NJOBS" bin/votour # test the success of extra makefile targets that aren't built by default
```

SkySkimmer:
Not in the test suite template, it would need to rebuild the 
dependencies of votour

It seems fragile because only one job failed, even dune+flambda didn't 
see the issue

Gaëtan Gilbert

ejgallego:
@JasonGross note that this problem occurred by a transitory state of the CI system where it indeed was incomplete wrt the binaries.

JasonGross:
Oh, is the issue that the breaking PR was tested before the dune PR was merged, and so the failure of votour wasn't caught?

ejgallego:
> Oh, is the issue that the breaking PR was tested before the dune PR was merged, and so the failure of votour wasn't caught?

Yeah, `votour` used to be tested before in the normal CI, we will likely restore that by forcing `make coqocaml` in the main builds.

PR #8563, opened 2018-09-26
Fix checker soundness bug with polymorphism capturing global univs

SkySkimmer:
Followup for #8341.

Not making a test as that's too difficult with our current infrastructure.

### DIFF
--- a/checker/checker.ml
+++ b/checker/checker.ml
@@ -243,7 +243,9 @@ let explain_exn = function
   | Invalid_argument s ->
       hov 0 (anomaly_string () ++ str "uncaught exception Invalid_argument " ++ guill s ++ report ())
   | Sys.Break ->
-      hov 0 (fnl () ++ str "User interrupt.")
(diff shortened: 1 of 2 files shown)


SkySkimmer:
>The comment that precedes your change seems directly related. Should it be removed?
>A few lines later, the exact same code appears. Why is it not a mistake there too?

The comment applies to the second function. It's not a mistake because we still double-declare universes in Include and especially side effects.

>Is the exception ever caught?

Let me add it to `Checker.explain_exn`.

ppedrot (approved):
The checker does fancy stuff w.r.t. module inclusion and the like, and it adds global monomorphic constraints for no good reason. This should be fixed, but in the meantime the following PR makes the checker more robust. It might also reveal other bugs, so that's very good.

ppedrot:
@SkySkimmer will you adapt the exception explanation or should I merge?

SkySkimmer:
merge

Gaëtan Gilbert

PR #8565, opened 2018-09-26
Make coqpp handle OCaml locations

### LINKED ISSUE #8018: coqpp: locations and tooling

ppedrot:
We dump the line number of the quotation and pad it enough so that the error message location corresponds to the one from the code to be preprocessed.

Fixes #8018.

### DIFF
--- a/coqpp/coqpp_main.ml
+++ b/coqpp/coqpp_main.ml
@@ -21,6 +21,13 @@ let pr_loc loc =
   let epos = loc.loc_end.pos_cnum - loc.loc_start.pos_bol in
   Printf.sprintf "File \"%s\", line %d, characters %d-%d:" file line bpos epos
 
+let print_code fmt c =
+  let loc = c.loc.loc_start in
+  (** Print the line location as a source annotation *)
+  let padding = String.make (loc.pos_cnum - loc.pos_bol + 1) ' ' in
+  let code_insert = asprintf "\n# %i \"%s\"\n%s%s" loc.pos_lnum loc.pos_fname padding c.code in
+  fprintf fmt "@[@<0>%s@]@\n" code_insert
+
 let parse_file f =
   let chan = open_in f in
   let lexbuf = Lexing.from_channel chan in
@@ -181,8 +188,7 @@ let print_fun fmt (vars, body) =
   in
   let () = fprintf fmt "fun@ " in
   let () = List.iter iter vars in
-  (* FIXME: use Coq locations in the macros *)
-  let () = fprintf fmt "loc ->@ @[%s@]" body.code in
+  let () = fprintf fmt "loc ->@ @[%a@]" print_code body in
   ()
 
 (** Meta-program instead of calling Tok.of_pattern here because otherwise
@@ -304,8 +310,8 @@ let rec print_binders fmt = function
   fprintf fmt "%s@ %a" id print_binders rem
 
 let print_rule fmt r =
-  fprintf fmt "@[TyML (%a, @[fun %a -> %s@])@]"
-    print_clause r.tac_toks print_binders r.tac_toks r.tac_body.code
+  fprintf fmt "@[TyML (%a, @[fun %a -> %a@])@]"
+    print_clause r.tac_toks print_binders r.tac_toks print_code r.tac_body
 
 let rec print_rules fmt = function
 | [] -> ()
@@ -338,7 +344,7 @@ let declare_plugin fmt name =
   fprintf fmt "let _ = Mltop.add_known_module %s@\n" plugin_name
 
 let pr_ast fmt = function
-| Code s -> fprintf fmt "%s@\n" s.code
+| Code s -> fprintf fmt "%a@\n" print_code s
 | Comment s -> fprintf fmt "%s@\n" s
 | DeclarePlugin name -> declare_plugin fmt name
 | GramExt gram -> fprintf fmt "%a@\n" GramExt.print_ast gram
--- a/coqpp/coqpp_ast.mli
+++ b/coqpp/coqpp_ast.mli
@@ -11,7 +11,7 @@ type loc = {
   loc_end : Lexing.position;
 }
 
-type code = { code : string }
+type code = { code : string; loc : loc; }
 
 type user_symbol =
 | Ulist1 of user_symbol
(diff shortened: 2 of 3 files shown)


coqpp/coqpp_main.ml
```diff
@@ -21,6 +21,20 @@ let pr_loc loc =
   let epos = loc.loc_end.pos_cnum - loc.loc_start.pos_bol in
   Printf.sprintf "File \"%s\", line %d, characters %d-%d:" file line bpos epos
 
+let print_code fmt c =
+  let loc = c.loc.loc_start in
+  (** Print the line location as a source annotation *)
+  let () = pp_print_newline fmt () in
+  let annot = Printf.sprintf "# %i \"%s\"" loc.pos_lnum loc.pos_fname in
```
ejgallego: Please use `Format` and a readable format string instead of a more messy printing sequence calls.

ejgallego (changes requested):
`coqpp` has no codeowners, please add so we can properly review and merge this.

Also, please fix the printing sequence to use a format string.

coqpp/coqpp_main.ml
ppedrot: I did that precisely not to use Format, otherwise it uses the boxing system which I want to deactivate locally.

ppedrot:
Do you want to be the secondary owner of coqpp? I can open a PR to add myself as a primary owner and you as a secondary one.

ejgallego:
> Do you want to be the secondary owner of coqpp?

I am OK.

coqpp/coqpp_main.ml
ejgallego: Umm, that seems certainly strange, that whole routine should just be a one `printf` line.

Will have a look later, but note that you can control indentation globally in `Format` using `\n` for example. It is _not_ a good idea to do what you are doing now as the formatters will get confused.

ejgallego: ?? I am not sure what you are talking about, but the current output of this PR looks fairly "bad". This one for example seems to do the job better:

```ocaml
  let loc = c.loc.loc_start in
  fprintf fmt "\n# %i \"%s\"@]@\n @[%s@]" loc.pos_lnum loc.pos_fname c.code
```

ppedrot: It is on purpose that the output is ugly, it is done to ensure correct placement of the metadata and character count of the quoted code.

ejgallego: Oh I see, however you should not reset the formatter in the way you do, let me cook a new rule.

ejgallego: You can do this for example that gives slightly better output after the inline code as the formatter is not reset:
```ocaml
let print_code fmt c =
  let loc = c.loc.loc_start in
  (** Print the line location as a source annotation *)
  let padding = String.make (loc.pos_cnum - loc.pos_bol + 1) ' ' in
  fprintf fmt "@[\n# %i \"%s\"@]@[\n%s%s@]@\n" loc.pos_lnum loc.pos_fname padding c.code
```
there are many variations tho, indeed you can build the whole string if you want and use `@<0>`, but that is only if you are not forcing a flush on the upstream `print_code` call

ppedrot: Your variant generates trailing whitespaces before line annotations. I am not sure it is really nice, it might trigger the wrath of linters.

ejgallego: Well linters should not deal with these files anyways but I agree that's not pretty. I can cook up a new rule that will indeed make the box transparent, anyways let's not delay this a lot. The original code indeed resetted too much so hence after the insertion you see the positions was also lost.

ejgallego: Note that Format choose to break there due to the top-level caller of `print_code` having had specified a `@ ` hint. but indeed we may want to make the box zero-sized.

ejgallego: Something like this maybe?
```ocaml
let print_code fmt c =
  let loc = c.loc.loc_start in
  (** Print the line location as a source annotation *)
  let padding = String.make (loc.pos_cnum - loc.pos_bol + 1) ' ' in
  let code_insert = asprintf "\n# %i \"%s\" \n%s%s" loc.pos_lnum loc.pos_fname padding c.code in
  fprintf fmt "@[@<0>%s@]@\n" code_insert
```

ppedrot: Still doesn't work, there are still many trailing spaces... What are we aiming for, exactly?

ejgallego: I am not sure what do you mean, as this version produces the same trailing whitespaces than the one originally in this PR, which is logical given that you insert the whitespace in the called of `print_code`.

Anyways, as I wrote in the review, "please fix the printing sequence to use a format string." This should have the side effect of not calling `pp_print_newline` incorrectly.

ppedrot: No, I mean that there is whitespace inserted for god knows which reason that is not part of the code. The whitespace appears seemingly on the line before the preprocessor line annotation in certain cases yet to be discovered.

Anyways, if the only thing you complain about is the call to `pp_print_newline` I can indeed just dump an ad-hoc string that contains newlines. Hopefully this will not disrupt the formatting.

ejgallego: > No, I mean that there is whitespace inserted for god knows which reason that is not part of the code. The whitespace appears seemingly on the line before the preprocessor line annotation in certain cases yet to be discovered.

This happens both with the above code and with the original one, and it appears because you request it when you do:
```ocaml
  let () = fprintf fmt "loc ->@ @[%a@]" print_code body in
```
obviously. The whitespace doesn't bother me at all tho.

My request is very clear, "use a format string" such as any of the ones I have proposed [to your taste].

ppedrot:
OK, I've pushed your last version.

ejgallego (approved):
LGTM, will merge as soon as the CI is happy.

PR #8566, opened 2018-09-26
[dune] [test-suite] Support for running the test suite with Dune.

ejgallego:
This patch adds support to run the test suite from Dune. We depend on the whole `coq` build for now, but ideally, we would run `coqdep` over the test suite to extract proper dependencies.

The PR depends on a few cleanups, but nothing substantial, and IMO it should be fairly close to be ready.

There is indeed the possibility to port some parts of the test suite to Dune or to a different testing framework, but that's not for the immediate future.

### DIFF
--- a/Makefile.dune
+++ b/Makefile.dune
@@ -1,7 +1,7 @@
 # -*- mode: makefile -*-
 # Dune Makefile for Coq
 
-.PHONY: voboot states world watch release apidoc ocheck ireport clean help
+.PHONY: help voboot states world watch test-suite release apidoc ocheck ireport clean
 
 # use DUNEOPT=--display=short for a more verbose build
 # DUNEOPT=--display=short
@@ -13,6 +13,7 @@ help:
 	@echo "  - states:  build a minimal functional coqtop"
 	@echo "  - world:   build all binaries and libraries"
 	@echo "  - watch:   build all binaries and libraries [continuous build]"
+	@echo "  - test-suite: run Coq's test suite"
 	@echo "  - release: build Coq in release mode"
 	@echo "  - apidoc:  build ML API documentation"
 	@echo "  - ocheck:  build for all supported OCaml versions [requires OPAM]"
@@ -32,6 +33,9 @@ world: voboot
 
 watch: voboot
 	dune build $(DUNEOPT) @install -w
+
+test-suite: voboot
+	dune $(DUNEOPT) runtest
 
 release: voboot
 	dune build $(DUNEOPT) -p coq
--- a/configure.ml
+++ b/configure.ml
@@ -1038,7 +1038,7 @@ let do_one_instdir (var,msg,uservalue,selfcontainedlayout,unixlayout,locallayout
           try Some (Sys.getenv "COQ_CONFIGURE_PREFIX")
           with
           | Not_found when !prefs.interactive -> None
-          | Not_found -> Some "_build/install/default"
+          | Not_found -> Some Sys.(getcwd () ^ "/../install/default")
         end
       | p -> p
     in match uservalue, env_prefix with
--- a/ide/dune
+++ b/ide/dune
@@ -10,6 +10,8 @@
 
 (executable
  (name fake_ide)
+ (public_name fake_ide)
+ (package coqide-server)
  (modules fake_ide)
  (libraries coqide-server.protocol coqide-server.core))
 
--- a/test-suite/dune
+++ b/test-suite/dune
@@ -1,0 +1,73 @@
+(rule
+ (targets summary.log)
+ (deps
+   ; File that should be promoted.
+   misc/universes/all_stdlib.v
+   ; Dependencies of the legacy makefile
+   ../Makefile.common
+   ../config/Makefile
+   ; Stuff for the compat script test
+   ../dev/header.ml
+   ../dev/tools/update-compat.py
+   ../doc/stdlib/index-list.html.template
+   (package coq)
+   ; For fake_ide
+   (package coqide-server)
+   (source_tree .))
+   ; Finer-grained dependencies look like this
+   ; ../tools/CoqMakefile.in
+   ; ../theories/Init/Prelude.vo
+   ; ../theories/Arith/Arith.vo
+   ; ../theories/Arith/Compare.vo
+   ; ../theories/PArith/PArith.vo
+   ; ../theories/QArith/QArith.vo
+   ; ../theories/QArith/Qcanon.vo
+   ; ../theories/ZArith/ZArith.vo
+   ; ../theories/ZArith/Zwf.vo
+   ; ../theories/Sets/Ensembles.vo
+   ; ../theories/Numbers/Natural/Peano/NPeano.vo
+   ; ../theories/Numbers/Cyclic/Int31/Cyclic31.vo
+   ; ../theories/FSets/FMaps.vo
+   ; ../theories/FSets/FSets.vo
+   ; ../theories/MSets/MSets.vo
+   ; ../theories/Compat/Coq87.vo
+   ; ../theories/Compat/Coq88.vo
+   ; ../theories/Relations/Relations.vo
+   ; ../theories/Unicode/Utf8.vo
+   ; ../theories/Program/Program.vo
+   ; ../theories/Classes/EquivDec.vo
+   ; ../theories/Classes/DecidableClass.vo
+   ; ../theories/Classes/SetoidClass.vo
+   ; ../theories/Classes/RelationClasses.vo
+   ; ../theories/Logic/Classical.vo
+   ; ../theories/Logic/Hurkens.vo
+   ; ../theories/Logic/ClassicalFacts.vo
+   ; ../theories/Reals/Reals.vo
+   ; ../theories/Lists/Streams.vo
+   ; ../plugins/micromega/Lia.vo
+   ; ../plugins/micromega/Lqa.vo
+   ; ../plugins/micromega/Psatz.vo
+   ; ../plugins/micromega/MExtraction.vo
+   ; ../plugins/nsatz/Nsatz.vo
+   ; ../plugins/omega/Omega.vo
+   ; ../plugins/ssr/ssrbool.vo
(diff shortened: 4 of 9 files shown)


tools/coq_makefile.ml
```diff
@@ -124,6 +124,7 @@ let quote s = if String.contains s ' ' || CString.is_empty s then "'" ^ s ^ "'"
 
 let generate_makefile oc conf_file local_file args project =
   let makefile_template =
+    Format.eprintf "%s@\n%!" Sys.(getcwd ());
```
gares: Is this a leftover?

ejgallego: Yes sorry, I was trying out to figure what was going on with path handling.

ejgallego: Indeed I am very close to factorize path handling a bit across all Coq, however I still need to figure out how to act on installed / built targets.

test-suite/misc/universes/dune
```diff
@@ -0,0 +1,9 @@
+(rule
+ (targets all_stdlib.v)
+ (deps
+   (source_tree ../../../theories)
+   (source_tree ../../../plugins))
+ (action
+   (write-file all_stdlib.v "Require Coq.Init.Prelude.")))
+   ; (with-outputs-to all_stdlib.v
+   ;  (run find ../../../theories ../../../plugins -type f -name *.v))))
```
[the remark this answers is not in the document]
ejgallego: This is indeed an in-progress adaptation of the recipe in `Makefile.vofiles`.

test-suite/dune
```diff
@@ -0,0 +1,64 @@
+(rule
+ (targets summary.log)
+ (deps
+   misc/universes/all_stdlib.v
```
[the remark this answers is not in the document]
ejgallego: Indeed this is work in progress, I'd like to have `coqdep` run on each file so the list below should disappear.

ejgallego:
Ok, so this mostly works but basically is blocked on having a richer dependency specification in Dune itself (cc: https://github.com/ocaml/dune/issues/1155 )

ppedrot (commented):
Not qualified to review this.

ejgallego:
This is now in a state I am satisfied with for the moment. Many more cleanups could happen in the future.

ide/dune
```diff
@@ -10,6 +10,8 @@
 
 (executable
  (name fake_ide)
+ (public_name fake_ide)
```
Zimmi48: What is this? I thought you had backtracked on this? (This seems confirmed by the way you have written the dune file BTW.

test-suite/misc/universes/build_all_stdlib.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+echo "Require $(find ../../../theories ../../../plugins -type f -name "*.v" | \
+        sed 's/^.*\/theories\///' | sed 's/^.*\/plugins\///' | sed 's/\.v$//' | sed 's/\//./g') ."
```
Zimmi48: I'm surprised that you don't make the `Makefile` call this script as well like you usually do.

test-suite/Makefile
```diff
@@ -102,7 +102,7 @@ VSUBSYSTEMS := prerequisite success failure $(BUGS) output \
   coqdoc ssr
 
 # All subsystems
-SUBSYSTEMS := $(VSUBSYSTEMS) misc bugs ide vio coqchk coqwc coq-makefile tools unit-tests
+SUBSYSTEMS := $(VSUBSYSTEMS) misc bugs ide vio coqchk coqwc coq-makefile unit-tests #tools
```
Zimmi48: Not sure what this means...

ide/dune
ejgallego: I am still undecided on that one, binaries on path have certain advantages, and the main makefile expects fake_ide to be in path. I still got some time tho as this PR depends on other 3 so we can implement the solution we think best.

test-suite/Makefile
ejgallego: See the comment on the last commit. Working with this or I have discovered some tests that are either broken or too fragile. The current one in tools certainly seems very suspicious, in particular due to the way it access the documentation, and in how many path deps it introduces, I guess I'll file a bug.

test-suite/misc/universes/build_all_stdlib.sh
ejgallego: What do you mean? This is just a shell version of what's in makefile.vofiles

Zimmi48: Yes indeed, but usually you would take out small pieces of Makefile, put them in shell scripts and have both the Makefile and dune call the script.

test-suite/Makefile
Zimmi48: Yep, I had read the commit message but it was still not clear to me. Anyways, this change should be reviewed by @gares.

Zimmi48: In fact I finally understood thanks to your comment in another thread:

>For example, the update-compat test is quite fragile and it even depends on the stdlib documentation having been built.

test-suite/misc/universes/build_all_stdlib.sh
ejgallego: It is more complicated than that for this case, if you look at the way the make rules work. Anyways this should be a promoted file and thus checked in the CI I think.

test-suite/Makefile
ejgallego: Yeah, sorry if things are not very clear. As this depends on a few PR I hesitated to improve more as to avoid spam.

ejgallego:
Thanks to Jason who took me out of my mistake, this now runs the full test-suite. The depending PR has been approved, so IMHO this is looking in good shape.

We only need to see what to do with `fake_ide`, having it installed is more convenient but anyways that's a minor concern.

ejgallego:
Ready for review / merge

test-suite/misc/universes/build_all_stdlib.sh
gares: I don't get why things are done like that in Makefile.build, and consequently why you are duplicating the logic here.
One wants the target "test-suite" to depend on the "stdlib" one. What is the point of generating a file that loads the entire stdlib? Rather than duplicating this logic, you should aim at the most natural way of expressing dependencies among targets. Do you have a good reason for doing things this way in dune?

gares:
I quite don't see what this PR brings. After all it just runs `make` inside test-suite. What is the motivation for that?

test-suite/misc/universes/build_all_stdlib.sh
Zimmi48: Indeed @gares, I think generating this file is necessary to check that we can give consistent universes to the full stdlib.

Zimmi48:
This is a first step to allow people to stop using `make` and use `dune` instead. Of course, once we abandon make for real, we can go further and replace the Makefile of the test-suite itself. But there is no urgency.

configure.ml
```diff
@@ -1038,7 +1038,7 @@ let do_one_instdir (var,msg,uservalue,selfcontainedlayout,unixlayout,locallayout
           try Some (Sys.getenv "COQ_CONFIGURE_PREFIX")
           with
           | Not_found when !prefs.interactive -> None
-          | Not_found -> Some "_build/install/default"
+          | Not_found -> Some Sys.(getcwd () ^ "/../install/default")
```
Zimmi48: That's a weird parenthesizing but well, why not?

ide/dune
Zimmi48: If you are still undecided, then keep it uninstalled.

ejgallego:
> If you are still undecided, then keep it uninstalled.

Ok, but it adds a bit more complexity as then the binary is named differently in `make`-based builds and in Dune-based builds, this is why I hesitated. Another option is to install it for now, and if we move to a Dune-only setup make it private again.

Zimmi48:
That could work as well. As long as we don't release, nothing is set in stone anyways.
That being said, the current patch seems to be taking both options at once (it reinstalls fake_ide but also tells the test-suite where to find it).

ejgallego:
> also tells the test-suite where to find it

Where does the patch do that? If you are thinking about `BIN=` note that this is used for all binaries `coqtop, etc...`

test-suite/dune
```diff
@@ -0,0 +1,73 @@
...
+   ../Makefile.common
+   ../config/Makefile
+   ; Stuff for the compat script test
+   ../dev/header.ml
+   ../dev/tools/update-compat.py
+   ../doc/stdlib/index-list.html.template
+   (package coq)
+   ; For fake_ide
+   (package coqide-server)
+   (source_tree .))
```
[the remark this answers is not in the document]
Zimmi48: Thanks!

Zimmi48 (approved):

test-suite/dune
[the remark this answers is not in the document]
ejgallego: Thanks to you!

SkySkimmer (approved):
CI part seems fine

Makefile.dune
```diff
@@ -1,7 +1,7 @@
 # -*- mode: makefile -*-
 # Dune Makefile for Coq
 
-.PHONY: voboot states world watch release apidoc ocheck ireport clean help
+.PHONY: help voboot states world watch test-suite release apidoc ocheck clean
```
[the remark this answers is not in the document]
ejgallego: Lost on the merge, good catch.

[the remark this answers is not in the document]
Zimmi48: Could you put each target on a separate line to avoid merging problems in the future?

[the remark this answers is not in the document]
ejgallego: Fixed.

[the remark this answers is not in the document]
ejgallego: > Could you put each target on a separate line to avoid merging problems in the future?

Seems like a good idea indeed.

ejgallego:
> Failed in AppVeyor but due to a timeout _after_ the test-suite passed.

Indeed maybe we could disable `coqchk` on Appveyor? Seems a bit overkill.

PR #8567, opened 2018-09-26
More general support for installation of coqide keys

### LINKED ISSUE #8269: Modifier configuration take precedence on accelerator keymap since Coq 8.6
#### Version

8.8.1

#### Operating system

macOS 10.13.6

```coq
Require Import Arith List.

Fixpoint insert (cmp: nat -> nat -> bool)(x : nat)(l : list nat) :=
  match l with
    nil => x::nil
	| a ::l' => if cmp x a then x::a::l' else a::insert cmp x l'
  end.

Fixpoint sort (cmp : nat -> nat -> bool) (l : list nat) :=
  match l with nil => nil | x :: l' => insert cmp x (sort cmp l') end.

Definition sort_increasing := sort leb.

Compute sort_increasing (3::4::2::1::11::nil).
```
 is displayed like this with coqIde 8.8.1

```
     = @cons nat (S O)
         (@cons nat (S (S O))
            (@cons nat (S (S (S O)))
               (@cons nat (S (S (S (S O))))
                  (@cons nat
                     (S
                        (S
                           (S
                              (S
                                 (S
                                 (S
                                 (S
                                 (S (S (S (S O)))))))))))
                     (@nil nat)))))
     : list nat
```
and like this with coqIde 8.8.0
```
     = 1 :: 2 :: 3 :: 4 :: 11 :: nil
     : list nat
```

herbelin:
**Kind:** installation, fix, enhancement

This PR:
- fixes an obsolete installation line in `Makefile.ide`
- add support for the `-local` mode to automatically find the Coq-provided default `coqide.keys`
- ~~add support for specific default `coqide.keys` files under Win32 or Unix (on MacOS X, we keep the existing file)~~ Changed: platform-dependent modifiers are now hard-wired in `preferences.ml` ([see proposal below](https://github.com/coq/coq/pull/8567#issuecomment-435661660))

On the later point, it might be useful to double-check with Win32, MacOS X, Unix users what should be the optimal default menu bindings.

Added: also fixes #8269 ~~by moving hard-wired configuration of menu keys to the platform-dependent configuration files~~ by delaying `attach_modifiers` callback to after loading preferences.


configure.ml
```diff
@@ -1391,3 +1391,23 @@ let write_configpy f =
...
+    while true do
+      let s = input_line i in
+      output_string o s;
+      output_char o '\n'
+    done
+  with End_of_file ->
+    close_in i;
+    close_out o
+
+let _ =
```
ejgallego: I am not sure what you are trying to do, but certainly configure is not the place to do this. Move to CoqIDE makefile, or even better make CoqIDE smarter so it can choose the file.

ejgallego (changes requested):

configure.ml
herbelin: I need the platform-dependent version of the file `ide/coqide.keys` to exist also in `-local` mode, this is why I put it in `configure`. This file has to be "generated" before any form of `make install` because `-local` does not install. The `cp` could be done in, say, `make  bin/coqide`, but e.g. `dev/ocamldebug-coq` is generated by `configure.ml` and not by `make`, so I did the same.

ejgallego: First, no reason why you can't have coqide select the right file based on the platform and you remove a problem. Modify `preferences.ml` so the name is selected in base of the platform.

Would you find a reason not to do that [bad idea], then `coqide.keys` would be an auto-generated file depending on other sources files, thus `configure` is the wrong place to generate it as it won't track the dependencies correctly.

herbelin: > First, no reason why you can't have coqide select the right file based on the platform and you remove a problem. 

What do you mean? That we should install all platform-dependent variants of the default `coqide.keys` template whatever the installation target is? (So that `coqide`, is dynamically able to select the default template based on the platform.)

> Would you find a reason not to do that [bad idea], then coqide.keys would be an auto-generated file depending on other sources files, thus configure is the wrong place to generate it as it won't track the dependencies correctly.

I don't understand which other dependencies we would have to track. I'm talking here about the template which serves to generate the `coqide.keys` file the first time `coqide` is used.

ejgallego: > What do you mean? That we should install all platform-dependent variants of the default `coqide.keys` template whatever the installation target is? (So that `coqide`, is dynamically able to select the default template based on the platform.)

I do mean that indeed.

> I don't understand which other dependencies we would have to track. I'm talking here about the template which serves to generate the `coqide.keys` file the first time `coqide` is used.

You need to track the source file. For example when cross building, or if you update the source file the copy will become stale unless you call configure again.

ejgallego: Configure is not a build system, it's purpose is to query the few things the build system cannot and then let the build system proceed. In the case of Coq, it means that we can get rid of configure altogether [and we hope we will].

herbelin: > I do mean that indeed.

OK, I can do that, but this would be inconsistent, as I said, with e.g. the way `dev/ocamldebug-coq` is built (which indeed is not updated automatically if the `configure.ml` is updated). Also, the current way is not impeding cross building: If you want to build for platform X on platform Y, you will tell the configure to prepare the `coqide.keys` appropriate for platform X.

Then, if you have in mind a view where `configure` disappears, that's another story. Can you tell more? Is that a standard enough approach in the ecosystem that users would not get disoriented?

ejgallego: I'd say that no configure is fairly standard these days, anyways users are not mean to see configure even today.

Building-wise remembering when to run configure is generally a pain, anyways, this kind of files are better suited for make.

herbelin: @ejgallego: I have nothing against your proposal but I remain skeptical that what I did is non standard or "bad" per se. For instance, we install the MacOS specific stuff (like `coqide.icns`) only on MacOS. Do you mean that we could want, say, to install coq on Linux but use this installation to run CoqIDE in a MacOS image and thus need the MacOS file even in the Linux installation? Otherwise said, why not doing what you suggest systematically then, for all platform-dependent files?

If the point is to get rid of `configure` why not to do all these getting-rid-of-configure at the same time, but in the meantime preserving some consistency.

So, shortly, I feel a bit lost in how imperative your request is regarding a future which you seem to see clearly but which I have some difficulty myself to foresee. Why should I change the way coqide is looking for its files just to avoid running `configure` which we anyway need to run at the current time, and which I don't see how we could avoid having to run.

ejgallego: I didn't mean such esoteric stuff, it was just a suggestion. It is OK if you want to add more complexity to the build system instead of the particular application is affected, but auto-generated files should be handled by make not by configure. Just add make rule for the file and dependencies will be properly tracked.

In the end, what matters is where do you push the complexity of maintaining consistency of the build, and what is the effort of doing so.

ejgallego: My request about this not going in configure is pretty strong. Regarding using `make` vs having CoqIDE load the proper file, I do understand that using OCaml to implement the second is far easier than having to write fairly tricky make scripts, however there I shall remain neutral if you feel writing make is easier than OCaml.

ejgallego: Also, why is configure a problem? Because you never know when you have to re-run it. In fact, in Dune, running configure is not needed as in fact the files that are produced are handled in the standard dependency chain. So you just do `dune build` and forget about what has to be done.

ejgallego: A good example of how painful configure is the problems we have when changing OCaml compiler switch. If you change your compiler and forget to do configure you get weird problems; this is better managed by the build system which can properly track a dependency on the compiler.

ejgallego: So this is not about "removing" configure so much, but moving its functionality to a system that manages it automatically.

herbelin: > I do understand that using OCaml to implement the second is far easier than having to write fairly tricky make scripts, however there I shall remain neutral if you feel writing make is easier than OCaml.

I'm very sorry @ejgallego, I don't feel at writing fairly tricky scripts now that I have a version which works and which seems compatible with the spirit of what exists in configure.ml currently. If you know exactly what you want and how to implement it (and incidentally how to make it compatible with dune), would you be ok to force-push what you want to this branch?

ejgallego: I am afraid that I don't have the time now to understand what this file is supposed to do; I am afraid that I cannot ACK the addition you propose to configure.

herbelin: Looks like a deadlock!

ejgallego: An additional and stronger reason is that up to this point `ide` is fairly independent [that is, it can build in isolation, which is important for packages]; after you patch, such property would be lost.

herbelin: How much time would you need to remove `configure.ml`, or at least to move the building of `dev/ocamldebug-coq` out of `configure.ml`?

herbelin: Maybe a third opinion could help me to understand what is the right direction to go? Is there something wrong with my approach, considering the current situation? What would e.g. a nix expert do?

Makefile.ide
```diff
@@ -159,6 +159,18 @@ $(IDETOPBYTE): ide/idetop.ml $(LINKCMO) $(LIBCOQRUN) $(IDETOPCMA)
...
+KEYSRC:=ide/config_files/MacOS/default_accel_map
+else
+ifeq ($(ARCH),win32)
+KEYSRC:=ide/config_files/Win32/default_accel_map
+else
+KEYSRC:=ide/config_files/Other/default_accel_map
+endif
+endif
+ide/coqide.keys: $(KEYSRC)
+	cp $< $@
```
SkySkimmer: Here's a makefile using commit. Should do the same thing as the configure code. WDYT?

SkySkimmer: `ARCH` is put by configure in config/Makefile btw.

ejgallego: > `ARCH` is put by configure in config/Makefile btw.

`make` has always `OS` in the env, right?

SkySkimmer:
I don't feel like figuring out if there are corner cases with it

Gaëtan Gilbert

Makefile.ide
```diff
@@ -159,6 +159,18 @@ $(IDETOPBYTE): ide/idetop.ml $(LINKCMO) $(LIBCOQRUN) $(IDETOPCMA)
 			 $(SYSMOD) -package camlp5.gramlib \
 			 $(LINKCMO) $(IDETOPCMA) $(BYTEFLAGS) $< -o $@
 
+ifeq ($(ARCH),Darwin)
+KEYSRC:=ide/config_files/MacOS/default_accel_map
```
ejgallego: You could maybe do `KEYSRC:=ide/config_files/$(ARCH)/default_accel_map` , but I am not expert.

ejgallego (approved):
Configure is not affected anymore, so I don't see a problem with this. I still believe that writing the selection in OCaml provides for more robust code, but using `make` to generate the file is also OK; down to the CoqIDE maintainers I guess.

ppedrot (commented):
Don't feel too confident reviewing that one so I'll trust the other. I can assign it to myself if needed though.

herbelin:
BTW, would say that it would be good for coqide users that it goes to 8.9, and preferably in the beta.

herbelin:
@ppedrot: do you assign as maintainer?

ppedrot:
I second @silene's comments, it does look like you just removed the ability to change modifiers.

herbelin:
>  it does look like you just removed the ability to change modifiers.

Yes, something is not clear. I was trying to fix #8269 (see this [comment](https://github.com/coq/coq/issues/8269#issuecomment-414704047) and the subsequent one by @newcat12).

There is apparently a confusion between the coqide preference file (`coqiderc`) and the accel_map file (`coqide.keys`). The former is menu-based, the latter is menu-item-based and there seems to be a competition between them. I suspect @newcat12 to be right and that loading the preference file is interpreted by `attach_modifiers` as a change which overwrites what the `coqide.keys` had decided.

Without this PR, try the following:
- open CoqIDE
- in (e.g.) the Navigation menu, go to Forward and type "Shift"+"Down" altogether so that the key is changed from Ctrl-down to Shift-Down
- quit CoqIDE, this changes the `coqide.keys` to indicate `(gtk_accel_path "<Actions>/Navigation/Forward" "<Shift>Down")`
- relaunch CoqIDE: the key for Forward has been changed back to `<Ctrl>Down` because of `attach_modifiers`

So, there is a conflict between the global modifier in `coqiderec` (and the preference panel) and the accep map `coqide.keys`, and I see two things to do:
- explain in the preference window that changing a modifier which change all items of the menu
- ensure that `attach_modifiers` is called after loading the preferences

I modified the PR accordingly. Do you agree with the analysis and change?

silene:
I am not sure. It feels like `attach_modifiers` should never be called after loading the acceleration map. Otherwise, we will just overwrite the user settings. It should only be called once, when the user changes the modifiers using the dialog box. Then, the modified acceleration map will be saved when closing Coqide.

I wonder if it would not be better to scrap all this code. Instead, the preference box would just let the user choose an acceleration map among several predefined ones. Then we never touch the map again, until the user decides to reset the map to some other predefined one.

silene:
> it should morally be called only once, but `load_prefs` is also called only once

No, `load_prefs` is called each time Coqide is started. But `attach_modifiers` should only be called when the user modifies the preferences. If `attach_modifiers` is called by `load_prefs` each time Coqide is started, then any user modification will be overwritten.

herbelin:
> > it should morally be called only once, but load_prefs is also called only once
> 
> No, load_prefs is called each time Coqide is started.

Sorry, I don't understand. Isn't "each time CoqIDE is started" precisely "once"?

> But `attach_modifiers` should only be called when the user modifies the preferences. If attach_modifiers is called by `load_prefs` each time Coqide is started, then any user modification will be overwritten.

Are you saying that my interpretation of what `attach_modifiers` does, namely that it does not register a modifier, but register a callback for when the users modifies the menu preferences, is wrong? If I'm not wrong, there is no problem in calling `attach_modifiers` since what matters is if and when the user effectively asks for a change. If the user does not change the preferences, the callback set by `attach_modifiers` will never be called, no?

silene:
> Are you saying that my interpretation of what `attach_modifiers` does, namely that it does not register a modifier, but register a callback for when the users modifies the menu preferences, is wrong?

Sorry, I should not have answered without looking at the code. My first post was correct, but then I swapped the names `attach_modifiers` and `modifier_for`. What I meant is that, at each startup, the following code is executed:

```
  let navitem (text, label, stock, callback, tooltip, accel) =
    let accel = modifier_for_navigation#get ^ accel in
    item text ~label ~stock ~callback ~tooltip ~accel
  in
  ... List.map navitem [
    ("Forward", "_Forward", `GO_DOWN, Nav.forward_one, "Forward one command", "Down");
```

So, any custom accelerator the user might have chosen is overwritten, since this code is executed after `GtkData.AccelMap.load loaded_accel_file`.

herbelin:
> So, any custom accelerator the user might have chosen is overwritten, since this code is executed after GtkData.AccelMap.load loaded_accel_file.

I just tested and it is apparently not the case.

The accel_map is registered even though the ui is not yet built. Apparently, when `navitem` is called and an "accel" already exists, prior to the creation of the item, the value registered in the accel_map is used.

I also tested to load the pref and accel_map files _after_ the ui is built, i.e. after `navitem` is called and it works the same. So, loading the accel map has an effect both by anticipation (when loaded before) and retroactively (when loaded after).

Don't know when loading the pref file is better though! Would you prefer it to be loaded after the interface is built? Or to keep as it is, which seems to work?

silene:
> The accel_map is registered even though the ui is not yet built. Apparently, when navitem is called and an "accel" already exists, prior to the creation of the item, the value registered in the accel_map is used.

Strange, that is not what I am experiencing. If I modify `coqide.keys`, my changes are overwritten by the value passed as `~accel`. Moreover, if I replace `~accel` by `~accel:"<Primary>Down"`, then I get a rather frustrating Coqide.

herbelin:
@silene: how do you modify `coqide.keys`?

silene:
> how did you modify `coqide.keys`?

By hand. For some reason, on my (new) laptop, modifying the accelerators on the fly inside Coqide no longer seems to work, so that is the only way I can imagine to modify an accelerator.

herbelin:
Sorry to ask this trivial question but I got trapped by this. Did you remove the semi-colon in front of the line you modified?

silene:
> Did you remove the semi-colon in front of the line you modified?

Good catch. Sorry for the noise. (And to add to the confusion, I was modifying the first Navigation entry of the file, i.e., Force, which happens to not be mapped to the menu, for some reason.)

That said, I don't understand the changes to `attach_modifiers` anymore. As you explained, it only registers a callback. So, it seems like all these changes were unneeded. Sorry.

Also, I now have a number of questions regarding the `default_accel_map` files you ship.

The one for MacOSX seems mostly fine, but the two other ones are a bit strange. First, none of their entries are commented. It feels like almost all their entries should be commented. If not, then it probably means that the default values in the OCaml code should be modified. For example, the default value for `<Actions>/Templates/Definition` is `<Primary><Shift>e`, but all three acceleration maps redefine it to `<Primary><Shift>d`. So, it should certainly be the default value in the OCaml code.

Another dubious thing with these files is that some entries do not make sense. For instance, Gtk maps `<Primary>`to `<Command>` on MacOSX, and to `<Control>` elsewhere. So, having entries using `<Primary><Control>` in the non-MacOSX acceleration maps feels wrong.

Another thing not making sense is that some entries are using `<Shift><Primary>`, since `<Primary>` is always the first modifier read/written by Gtk.

I guess all these issues could be fixed by having Coqide load and save the `default_accel_map` files at least once. If you need me to do so for Linux, just tell.

herbelin:
The `accep_map` are indeed dubious. I'll try to understand what generated them as they are.

silene:
> But loading the coqiderc preference file means a change of all the `modifier_for_*` map, and this calls the callback, thus overwriting what the accel map says.

I see.

> If GTK+ writes Control for control on non-MacOSX, I agree that there is no point in having a Primary in non-MacOSX accel maps.

No, it is the opposite. `<Control>` cannot appear on non-MacOSX accelerator maps, because that is superseded by `<Primary>`. The latter was introduced in Gtk to mean Command on MacOSX and Control elsewhere.

gares (approved):

herbelin:
I apparently forgot to push the commits I was talking about in this [comment](https://github.com/coq/coq/pull/8567#issuecomment-435661660). This is a completely different approach, closer from what @ejgallego proposed. No more platform-dependent installation file but a few hard-coded modifications of the default modifiers directly in `preferences.ml`. In particular, this is easier to maintain synchronous.

However, this is a slightly different mapping than what Pierre proposed (see [changes](https://github.com/coq/coq/pull/8567#issuecomment-435661660)).

@gares already approved. If new approach ok (@silene, @ppedrot, @skyskimmer, @ejgallego, ... ?), I'll squash before merging.

vbgl:
I’ll attempt to rebase.

vbgl:
Rebased. Can someone double-check, in particular that I did the right choice there: https://github.com/coq/coq/pull/8567/commits/bb7af1ed2d49b750596ab4d4fdeddaaee4baba64#diff-51f756aa3a4f676b4e8bfdb6158902a5R242 (the `get_config_file` function now gets a list of directories as argument).

herbelin:
Thanks a lot for the rebasing! For the `get_config_file` on unicode bindings, what I understand is that the unicode configuration file `unicode.bindings` lies at the first invocation in `Envars.configdir` (= `Minilib.coqide_default_config_dir ()`). Also, I would assume that having a system-wide unicode bindings file is allowed. So, with the confirmation of @charguer, I would look in order in the following list: `[Minilib.coqide_config_home (); Minilib.coqide_system_config_dirs (); Minilib.coqide_default_config_dir ()]`.

This list is basically `coqide_config_dirs ()` that this PR removed. Maybe this means that this alias should not be removed after all.

herbelin:
I just pushed the change so that the unicode bindings are searched in all config dirs.

I seized the opportunity to add a bit of documentation in Minilib and to move a trivial but common string function which was there in Minilib to CString.

ejgallego:
@herbelin seems to have a build problem.

herbelin:
@ejgallego: I fixed the build but I discovered dune problems in installing lablgtk3 using opam.

Apparently, `src/dune_config.ml` in the current version of `lablgtk3` (e.g. beta4 or beta5) is not anymore consistent with your proposed changes of the query API which have been integrated in the dune configurator 1.8.0. The function `gen_pkg` calls `query_pkg` with a package name of the form `gtk+-3.0 >= 3.18` which has become illegal with dune >= 1.8.0.

The following patch seems to respect the spirit of the query API change and work with dune 1.8:
```
*** src/dune_config.ml	2019-04-08 00:03:21.000000000 +0200
--- /Volumes/HFS/PACKAGES/lablgtk/src/dune_config.ml	2019-04-08 00:16:16.000000000 +0200
***************
*** 10,19 ****
      | None   -> error message
      | Some v -> v
  
-   let require_ok = function
-     | Error message -> error message
-     | Ok v -> v
- 
    let cata ~f d = function
      | None -> d
      | Some x -> f x
--- 10,15 ----
***************
*** 30,44 ****
    | _ ->
      package, []
  
! let query_pkg p ~package ~expr =
!   Option.require_ok C.Pkg_config.(query_expr_err p ~package ~expr)
  
  let gen_pkg p ~package ~version =
    let file kind = kind ^ "-" ^ package ^ ".sexp" in
    let package, extra_flags = platform_subst p ~package in
!   let expr =
      Option.cata ~f:(fun version -> Format.sprintf "%s >= %s" package version) package version in
!   let c_g = query_pkg p ~package ~expr in
    C.Flags.write_sexp (file "cflag") @@ c_g.C.Pkg_config.cflags @ extra_flags;
    C.Flags.write_sexp (file "clink") c_g.C.Pkg_config.libs
  
--- 26,40 ----
    | _ ->
      package, []
  
! let query_pkg p ~package =
!   Option.require ~message:(package ^ " not found") C.Pkg_config.(query p ~package)
  
  let gen_pkg p ~package ~version =
    let file kind = kind ^ "-" ^ package ^ ".sexp" in
    let package, extra_flags = platform_subst p ~package in
!   let package =
      Option.cata ~f:(fun version -> Format.sprintf "%s >= %s" package version) package version in
!   let c_g = query_pkg p ~package in
    C.Flags.write_sexp (file "cflag") @@ c_g.C.Pkg_config.cflags @ extra_flags;
    C.Flags.write_sexp (file "clink") c_g.C.Pkg_config.libs
```
Additionally, I had to add `libffi` manually to `PKG_CONFIG_PATH` under MacOS, but that seems to be a brew issue.

ejgallego:
> @ejgallego: I fixed the build but I discovered dune problems in installing lablgtk3 using opam.

Thanks @herbelin for the report; I am not sure tho the reason are my changes tho, but some other changes.

What Dune version you have?

herbelin:
Actually, I tested again without the above patch and it works (assuming `libffi` added manually to `PKG_CONFIG_PATH`). The only problem was that without the patch, the error about missing `libffi` was not reported and I did not know what to do.

The hiding of the error on missing `libffi` was indeed due not to your change on `query_expr_err` but to ocaml/dune#1842 (0e8655f4d) (in answer to ocaml/dune#1833).

I shall report it to lablgtk.

charguer:
> So, with the confirmation of @charguer, I would look in order in the following list: [Minilib.coqide_config_home (); Minilib.coqide_system_config_dirs (); Minilib.coqide_default_config_dir ()].

What would make sense to me is that `coqide.bindings` is searched for in the same places and in the same order as for `coqide.keys` and `coqiderc`. 

Beware that `default.bindings` is a different file, which is distributed with Coq binaries and should not be modified by the user.

As I wrote in the documentation source file `coqide.rst`:

> |CoqIDE| accepts a command line argument of the form
> ``-unicode-bindings file1,file2,...,fileN``.
> Each of the file tokens provided may consists of one of:
>  -  a path to a custom bindings file,
>  -  the token ``default``, which resolves to the default bindings file,
>  -  the token ``local``, which resolves to the `coqide.bindings` file
>     stored in the user configuration directory.

vbgl:
Ping @herbelin 

Will someone implement Arthur’s suggestion? Will this PR be ready by the end of this week? Thanks.

herbelin:
@charguer:
> What would make sense to me is that `coqide.bindings` is searched for in the same places and in the same order as for `coqide.keys` and `coqiderc`.

For `coqide.keys` and `coqiderc`, there is a notion of system-wide configuration and user-side configuration, with priority given to user-side configuration if existing.

Nothing was done in #8560 ("unicode bindings working out of the box") to possibly load a system-wide unicode bindings file, so I propose by simplicity to keep things as you did, i.e. to concatenate by default the local file and the default file. Therefore, I reverted to @vbgl revision.

Incidentally, the doc for locating `coqiderc` was wrong and I updated it with an extra commit.

I also "eliminated cuts" by squashing commits that were cancelling themselves.

herbelin:
@ppedrot: this one have CI finally green. I updated the initial comment. Seems ok to me.

doc/sphinx/practical-tools/coqide.rst
```diff
@@ -209,10 +209,13 @@ The last section is for miscellaneous boolean settings, such as the
...
 
-Notice that these settings are saved in the file `.coqiderc` of your
-home directory.
+Notice that these settings are saved in the file ``coqiderc`` in the
+``coq`` subdirectory of the user configuration directory which
+is the value of ``$XDG_CONFIG_HOME`` if this environment variable is
+set and which otherwise is ``$HOME/.config/``.
 
-A Gtk2 accelerator keymap is saved under the name `.coqide.keys`. It
+A Gtk2 accelerator keymap is saved under the name ``coqide.keys`` in
```
Zimmi48: Is it still "gtk2"?

Zimmi48 (approved):

doc/sphinx/practical-tools/coqide.rst
herbelin: No... 

Then, I just changed it to "GTK+ accelerator keymap".

ppedrot:
IIUC this has been ready for merging for a long time. @herbelin can you rebase?

herbelin:
I rebased. CI bedrock2 is failing but I imagine this is unrelated.

PR #8568, opened 2018-09-26
[dune] [coqide] Turn CoqIDE into a library.

ejgallego:
As noted by @anton-trunov, this is more useful for development and
debug.

### DIFF
--- a/ide/dune
+++ b/ide/dune
@@ -1,10 +1,11 @@
-(executable
- (name idetop)
- (public_name coqidetop.opt)
- (package coqide)
- (modules idetop)
- (libraries coq.toplevel coqide.protocol)
- (link_flags -linkall))
+(ocamllex utf8_convert config_lexer coq_lex)
+
+(library
+ (name core)
+ (public_name coqide.core)
+ (wrapped false)
+ (modules (:standard \ idetop coqide_main))
+ (libraries threads str lablgtk2.sourceview2 coq.lib coqide.protocol))
 
 (rule
  (targets coqide_main.ml)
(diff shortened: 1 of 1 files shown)


Zimmi48 (approved):
Looks reasonable. I can't do very convincing reviews yet: at some point I'll read the Dune doc to understand things better. Anyways, this has been tested by @anton-trunov so this is good to go.

ejgallego:
> Looks reasonable.

The patch should be self-explanatory, please ask any question you may have.

Instead of linking all the modules in ide in an executable, we put them in a cma and link the main file to it. The only advantage is that you can then load most of the IDE as a library for debug [or you could indeed link to ide functionality would you like to do so)

ide/dune
```diff
@@ -15,7 +16,13 @@
...
+ (modules coqide_main)
+ (libraries coqide.core))
 
-(ocamllex utf8_convert config_lexer coq_lex)
+(executable
+ (name idetop)
+ (public_name coqidetop.opt)
+ (package coqide)
+ (modules idetop)
+ (libraries coq.toplevel coqide.protocol)
```
Zimmi48: Why does `coqide_main` depends on the library `coqide.core` but `idetop` does not?

ejgallego: `coqide.core` contains all the "client-side" graphical routines, `idetop` is just `coqtop` with some extra capabilites, acting as the server. The only common part they have is the protocol, which is gtk-free.

Zimmi48: `coqide.core` is also said to contain module `idetop` though.

ejgallego: the line
```
(:standard \ idetop coqide_main)
```
means "all modules but `idetop` and `coqide_main`, see https://dune.readthedocs.io/en/latest/dune-files.html#ordered-set-language

PR #8571, opened 2018-09-26
[stdlib] Fix warning due to missing Declare Scope in Bvector

ejgallego:
This broke the build so it should be merged quickly. cc: #8171


herbelin:
Sorry, I was apparently too optimistic on accepting an outdated CI.

Issue #7149, opened 2018-04-01
Reducing configure


ejgallego:
IMHO we have too many configure options. I am aiming at a day where Coq can be configure-less given a standard OCaml/findlib environment is present, so this issue is a first step to see which options we could deprecate/remove. Options for the compiler/linker can be implemented using _workspaces_ that is a file containing different configurations, that can be manually edited. Current options are:

```
  -prefix <dir>               Set installation directory to <dir>
```
**REMOVE:** IMO, the build should not depend on the install path. This requires a different build system, and also some tweaks to options.

```
  -local                      Set installation directory to the current source tree
```
**REMOVE:** This requires a different build system that allows out-of-tree builds and provides a Nix-like PATH handling style. Dune #6857 can do both.
```
  -vmbyteflags <flags>        Comma-separated link flags for the VM of coqtop.byte
```
**REMOVE:** See #6475, so indeed we cannot `-custom` the bytecode executable, so we'll have to be careful with the path, but once `-local` is removed this can be done.
```
  -custom                     Build bytecode executables with -custom (not recommended)
  -no-custom                  Do not build with -custom on Windows and MacOS
```
**MOVE:** This option is useful but should indeed be moved to a workspace flags.
```
  -bindir <dir>               Where to install bin files
  -libdir <dir>               Where to install lib files
  -configdir <dir>            Where to install config files
  -datadir <dir>              Where to install data files
  -mandir <dir>               Where to install man files
  -docdir <dir>               Where to install doc files
  -emacslib <dir>             Where to install emacs files
  -coqdocdir <dir>            Where to install Coqdoc style files
  -ocamlfind <dir>            Specifies the ocamlfind command to use
  -lablgtkdir <dir>           Specifies the path to the Lablgtk library
  -camlp5dir <dir>            Specifies where is the Camlp5 library and tells to use it
  -camldir <dir>              Specifies path to 'ocaml' for running configure script
```
**REMOVE:** Libraries / tools should be located using ocamlfind.
```
  -arch <arch>                Specifies the architecture
```
**??:** No idea what this option does.
```
  -natdynlink (yes|no)        Use dynamic loading of native code or not
```
**UNDECIDED**
```
  -with-doc (yes|no)          Compile the documentation or not
  -coqide (opt|byte|no)       Specifies whether or not to compile CoqIDE
```
**REMOVE:** coqide should be turned into a standard package and build compositionally if specified by the build target. Similarly for doc.
```
[...]

PR #8572, opened 2018-09-27
[config] Miscellaneous cleaning of configuration variables.

ejgallego:
We perform a cleanup in `Coq_config` / `Envars`, consolidating handling of coqlib and removing unused / deprecated variables.

cc: #7149


ide/preferences.ml
```diff
@@ -359,14 +366,22 @@ let text_font =
   in
   new preference ~name:["text_font"] ~init ~repr:Repr.(string)
 
+let is_standard_doc_url url =
+  let wwwcompatprefix = "http://www.lix.polytechnique.fr/coq/" in
```
Zimmi48: Since you are moving this function around, I think it is time to remove this compat thing and simplify the code.

ejgallego: Sure, I planned to do that however I got scared by the logic there.

ejgallego:
@ppedrot you are the secondary for `lib` so maybe you could take this? `config/` has no owner, but it should be I guess the same than `configure` which is me, however there is no secondary.

ppedrot:
I am not sure I am really proficient, but let's take ownership whatsoever because otherwise nobody will.

kernel/nativelib.ml
```diff
@@ -37,7 +36,7 @@ let ( / ) = Filename.concat
 (* We have to delay evaluation of include_dirs because coqlib cannot be guessed
 until flags have been properly initialized *)
 let include_dirs () =
-  [Filename.get_temp_dir_name (); coqlib () / "kernel"; coqlib () / "library"]
+  [Filename.get_temp_dir_name (); Envars.coqlib () / "kernel"; Envars.coqlib () / "library"]
```
maximedenes: Why these changes?

ejgallego: I remove all `open Envars` so I can grep for who's using them, as all users need work.

For example `nativelib` will have to use a less adhoc compilation model to support Dune properly.

ppedrot (approved):

ppedrot:
Is this ready? If so I can merge quickly.

ejgallego:
> Is this ready? If so I can merge quickly.

@ppedrot seems ready to me, I don't think there is a lot to see here tho, most of the PR is trivial, and would we discover some problem straightforward to revert.

ejgallego:
Merci !

PR #8573, opened 2018-09-27
Backport #8247 and #8394 to 8.9

silene:

### DIFF
--- a/ide/idetop.ml
+++ b/ide/idetop.ml
@@ -204,27 +204,35 @@ let export_pre_goals pgs =
     Interface.given_up_goals = pgs.Proof.given_up_goals
   }
 
-let add_diffs oldp newp intf =
-  let open Interface in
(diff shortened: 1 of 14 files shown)


silene:
Sorry for the spam, I do not expect any review, that is just for continuous integration.

Zimmi48:
@silene We could kill the `CODEOWNER` file in v8.9 to avoid this issue in the future. WDYT? But what I used to do to avoid opening PRs when backporting to v8.8, was just to push to a special `staging-v8.8` branch to GitLab so that GitLab CI would run. Travis and AppVeyor are mostly redundant nowadays so I would only run them before pushing in case of doubt.

Zimmi48:
@silene If you want I can push a commit on top of this branch that would pin the tested versions of Elpi, HoTT and Ltac2 so that they pass CI (currently this is testing their master branch which contains breaking changes for commits that are only in master).
I can also push the commit fixing the Travis build by pinning OPAM to OPAM 1.2.

Zimmi48:
See #8574.

PR #8552, opened 2018-09-25
[ci] [docker] elpi version 1.1.0

gares:
I've not updated yet the coq-master branch of coq-elpi yet, since it would break things unless this is merged.
Tell me how/when to do it.

PR #8574, opened 2018-09-27
[ci] [v8.9] Test pinned commits for all external projects (and pin OPAM to 1.2 in Travis).

Zimmi48:
**Kind:** infrastructure.

cc @MSoegtropIMC: you were talking about pinning on v8.9. I just did it using the latest possible commits for all projects. This could be refined to using the latest tags for some projects (I guess this is what you had in mind).


Zimmi48:
I thought removing `CODEOWNERS` would be enough to avoid spamming in this PR but apparently it needs to be merged first, and only then the subsequent PRs won't spam.

Zimmi48:
@silene Note that the latest coq-elpi is technically still compatible with Coq 8.9 but testing it in CI would require backporting #8552. As for HoTT of course, the pinned commit will need to be updated if the numeral notation PR is backported.

ejgallego:
Indeed we should refine our CI policy wrt 3rd party developments branching. It should become mandatory projects follow a branch pattern so this kind of pr is not needed (by default the CI picks a branch name based on the declared coq version)

Zimmi48:
>Indeed we should refine our CI policy wrt 3rd party developments branching. It should become mandatory projects follow a branch pattern so this kind of pr is not needed (by default the CI picks a branch name based on the declared coq version)

That's what we did before, but in fact I really think that this approach (pinning) is better because it guarantees maximum CI predictability on stable branches.

Zimmi48:
CI failed: There is another commit I need to backport to fix the Windows build. Will do in a minute.

ejgallego:
Sure, pinning is not incompatible with what I mentioned, but auto branch selection may save some work and give us a more continuous system.

MSoegtropIMC (approved):
I would prefer tags where available instead of commit hashes, but for the time being this is fine. I can refine this later.

MSoegtropIMC:
@ejgallego : a release version tag of Coq should refer to a specific version of code. As soon as we include addons in the installer, we should use tags or commit hashes for all these addons. If we use a branch the code included in a release Coq installer could change. IMO this is not acceptable. So the idea is to use branches in the master branch but use tags in the release branches. One could use branches in the release branches as well but then one would have to change everything to tags before doing a release tag. This is also a viable option. I don't know which is better in practice, but I think using tags in the branches and possibly changing the tags when appropriate is safer (one could forget to do the tag change before doing a release tag).

There might be other options, e.g. retrieving branch versions at a specific point in time and define the time in the tag, but I find all these more difficult / error prone.

PR #8575, opened 2018-09-27
Remove {Safe_typing,Global}.push_context

SkySkimmer:
Adding a ucontext to the global environment only makes sense
internally when checking a polymorphic constant.

### DIFF
--- a/kernel/safe_typing.ml
+++ b/kernel/safe_typing.ml
@@ -273,7 +273,6 @@ let add_constraints_list cst senv =
   List.fold_left (fun acc c -> add_constraints c acc) senv cst
 
 let push_context_set poly ctx = add_constraints (Now (poly,ctx))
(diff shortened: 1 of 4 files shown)


ppedrot (approved):
Definitely agree.

ppedrot:
Actually, even `UContext.t` should be moved out of the kernel.

SkySkimmer:
If you mean abstracting before sending to the kernel I tried not so long ago but getting section letins to work was too hacky. Maybe it can be done but it requires actual thought.

ppedrot:
Sections are not the only problem, there are also issues with side-effects. But indeed, my long-term plan is to abstract outside of the kernel, and also separate the declaration of global constraints from term definitions.

maximedenes (approved):

PR #7820, opened 2018-06-15
[hints] Add Hint Variables/Constants Opaque/Transparent commands

mattam82:
This gives user control on the transparent state of a hint db. Can
override defaults more easily (report by J. H. Jourdan).

For "core", declare that everything is opaque

Document Hint Variables

**Kind:** feature

- [x] Corresponding documentation was added / updated (including any warning and error messages added / removed / modified).
- [x] Entry added in CHANGES.

PR #8576, opened 2018-09-27
[api] Remove unnecessary type alias introduced in 8.9

ejgallego:
This was introduced in #7820 and it is not needed indeed. As 8.9 was
not released we don't need to perform a deprecation phase.


ppedrot (approved):

ppedrot:
This is trivial, I'll merge quickly in absence of opposition.

PR #8578, opened 2018-09-27
[dune] Allow to build CI after a Dune build.

ejgallego:
This PR allows to perform `make ci-$pkg` after building Coq with Dune.

It surely needs refinement and there are some plugins that won't build [due to makefile hacks?] but it should be a good base to start with.

### DIFF
--- a/dev/ci/ci-common.sh
+++ b/dev/ci/ci-common.sh
@@ -8,6 +8,7 @@ export NJOBS
 
 if [ -n "${GITLAB_CI}" ];
 then
+    # Gitlab build, Coq installed into `_install_ci`
     export OCAMLPATH="$PWD/_install_ci/lib:$OCAMLPATH"
     export COQBIN="$PWD/_install_ci/bin"
     export CI_BRANCH="$CI_COMMIT_REF_NAME"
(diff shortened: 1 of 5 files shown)


SkySkimmer (changes requested):

dev/ci/ci-common.sh
```diff
@@ -15,18 +15,25 @@ then
     then
         export CI_PULL_REQUEST="${CI_BRANCH#pr-}"
     fi
+elif [ -n "${TRAVIS}" ];
+then
+    export CI_PULL_REQUEST="$TRAVIS_PULL_REQUEST"
+    export CI_BRANCH="$TRAVIS_BRANCH"
```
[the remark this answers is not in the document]
ejgallego: Oh indeed that was a mistake, fixed, thanks.

dev/ci/ci-common.sh
```diff
@@ -15,18 +15,29 @@ then
     then
         export CI_PULL_REQUEST="${CI_BRANCH#pr-}"
     fi
+# Travis build in -local mode
+elif [ -n "${TRAVIS}" ];
```
SkySkimmer: Do you mind putting the comments after the condition they apply to? Currently it feels like they have half a foot in the previous block.

SkySkimmer (approved):
Approving for ci-common.sh, I don't feel competent for the ml/dune changes

ejgallego:
> Approving for ci-common.sh, I don't feel competent for the ml/dune changes

Unfortunately `coqc` has no owner.

dev/ci/ci-common.sh
ejgallego: Sure, I never know where to place them, I usually prefer to read first the comment then the condition but will update.

dev/ci/ci-common.sh
```diff
@@ -8,25 +8,37 @@ export NJOBS
...
+    export CI_PULL_REQUEST="$TRAVIS_PULL_REQUEST"
+    export CI_BRANCH="$TRAVIS_BRANCH"
+elif [ -d "$PWD/_build/install/default/" ];
+then
+    # Dune build
+    export OCAMLPATH="$PWD/_build/install/default/lib/"
+    export COQBIN="$PWD/_build/install/default/bin"
+    export COQLIB="$PWD/_build/install/default/lib/coq"
+    CI_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
+    export CI_BRANCH
```
Zimmi48: There's no reason to keep the `export` on a separate line AFAICT. The quoting around `$( )` is probably excessive too. @SkySkimmer should know better though.

Zimmi48 (approved):
I would like to get @herbelin's review on the `coqc` files.

dev/ci/ci-common.sh
ejgallego: Note it was that way before, I ignore why. I can change if you want.

SkySkimmer:
>the |export| on a separate line

linter stuff (if git fails it won't be absorbed by export)

 > quoting around |$( )| is probably excessive too.

sure


Gaëtan Gilbert

ejgallego:
> I would like to get @herbelin's review on the `coqc` files.

The change in `coqc` is to use the centralized version for locating binaries, that works when these are symlinks [current one doesnt'].

> the |export| on a separate line

I am letting things as they are in that file then.

herbelin:
The change in `coqc` made on the assumption that `coqc` and `coqtop` lives in the same repository seems indeed reasonable.

Zimmi48:
@SkySkimmer You have self-assigned this PR, so I'll let you merge it.

PR #8582, opened 2018-09-27
[api] Deprecate `evar_map` ref combinators.

### LINKED ISSUE #6342: `e_` cleanup: remove all passing of `evar_map`s by using references.
Some parts of the codebase still manipulate `evar_map`s by using references, such as:

```ocaml
val pretype_type :
  int -> bool -> val_constraint -> env -> evar_map ref ->
  ltac_var_map -> glob_constr -> unsafe_type_judgment
```

We should indeed deprecate all the functions using this style and replace them with the equivalent functional version. It would be great to do the deprecation in 8.8.

A tree marking all the functions to be deprecated is at https://github.com/ejgallego/coq/tree/api%2Bevar_map_no_refs , so it could serve as a starting point. The number of warnings generated does indeed indicates that quite a bit of progress should be doable for 8.8.

A particular question is, what should we recommend for functions returning a pair, `Evd.sigma` ?

c.f. #946

ejgallego:
All the `evar_map` APIs were deprecated in 8.9, thus we deprecate the
combinators to discourage this style of programming.

Still a few places do use imperative style, but they are pretty
localized and should be cleaned up separately.

As these are the last bits of `e_` API remaining this PR closes #6342.

### DIFF
--- a/engine/evarutil.mli
+++ b/engine/evarutil.mli
@@ -258,8 +258,11 @@ val generalize_evar_over_rels : evar_map -> existential -> types * constr list
 (** Evar combinators *)
 
 val evd_comb0 : (evar_map -> evar_map * 'a) -> evar_map ref -> 'a
+[@@ocaml.deprecated "References to [evar_map] are deprecated, please update your API calls"]
 val evd_comb1 : (evar_map -> 'b -> evar_map * 'a) -> evar_map ref -> 'b -> 'a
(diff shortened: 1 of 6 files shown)


ejgallego:
This is ready for review (and merge).

ppedrot (approved):

ppedrot:
This is trivial, let me merge this right now.

PR #8584, opened 2018-09-28
[dune] Build also API documentation for plugins, improve ugly hack.

ejgallego:
IMHO we'd like to build documentation for plugins too, so we need to
run `voboot` before `apidoc` then.

We also try to make the horrendous hack to avoid `odoc` dying more
robust; suggestions on how to improve it welcome.

### DIFF
--- a/Makefile.dune
+++ b/Makefile.dune
@@ -34,10 +34,10 @@ watch: voboot
 release: voboot
 	dune build $(DUNEOPT) -p coq
 
-apidoc:
+apidoc: voboot
         # Ugly workaround for https://github.com/ocaml/odoc/issues/148
 	mv checker/dune checker/dune.disabled || true
-	dune build $(DUNEOPT) @doc
+	dune build $(DUNEOPT) @doc || ( mv checker/dune.disabled checker/dune && false )
 	mv checker/dune.disabled checker/dune || true
 
 clean:


Makefile.dune
```diff
@@ -34,10 +34,10 @@ watch: voboot
 release: voboot
 	dune build $(DUNEOPT) -p coq
 
-apidoc:
+apidoc: voboot
         # Ugly workaround for https://github.com/ocaml/odoc/issues/148
 	mv checker/dune checker/dune.disabled || true
-	dune build $(DUNEOPT) @doc
+	dune build $(DUNEOPT) @doc || mv checker/dune.disabled checker/dune
```
SkySkimmer: Add some exit 1 so it'll fail if there's a problem.

ejgallego: Done thanks, I hope I got the precedence right.

Makefile.dune
```diff
@@ -34,10 +34,10 @@ watch: voboot
 release: voboot
 	dune build $(DUNEOPT) -p coq
 
-apidoc:
+apidoc: voboot
         # Ugly workaround for https://github.com/ocaml/odoc/issues/148
 	mv checker/dune checker/dune.disabled || true
-	dune build $(DUNEOPT) @doc
+	dune build $(DUNEOPT) @doc || mv checker/dune.disabled checker/dune && exit 1
```
SkySkimmer: You can always put `{ }` around if you're not sure (I'm not either)

ejgallego: Umm, I think what we have now doesn't work, and `{ }` gives me a syntax error, `( )` doesn't work with exit for obvious reasons. I love `sh`.

SkySkimmer:
>() doesn't work with exit

Then use false

Gaëtan Gilbert

ejgallego:
> Then use false

well actually both produce the same return status, which is the important thing here.

Zimmi48 (approved):

ejgallego:
This is superseeded by #8588

PR #8585, opened 2018-09-28
Simplify declaration of universe names

SkySkimmer:
Declaring the universe to the kernel/section mechanism is centralized
to [Declare.declare_universe_context].
Then the universe name object really is only about the user visible
names.

### DIFF
--- a/interp/declare.ml
+++ b/interp/declare.ml
@@ -479,20 +479,7 @@ type universe_source =
   | QualifiedUniv of Id.t (* global universe introduced by some global value *)
   | UnqualifiedUniv (* other global universe *)
 
-type universe_decl = universe_source * Nametab.universe_id
-
-let add_universe src (dp, i) =
-  let level = Univ.Level.make dp i in
-  let optpoly = match src with
-    | BoundUniv -> Some true
-    | UnqualifiedUniv -> Some false
-    | QualifiedUniv _ -> None
-  in
-  Option.iter (fun poly ->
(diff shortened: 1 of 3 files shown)


interp/declare.ml
```diff
@@ -479,20 +479,7 @@ type universe_source =
...
-    | BoundUniv -> Some true
-    | UnqualifiedUniv -> Some false
-    | QualifiedUniv _ -> None
-  in
-  Option.iter (fun poly ->
-      let ctx = Univ.ContextSet.add_universe level Univ.ContextSet.empty in
-      Global.push_context_set poly ctx;
-      if poly then Lib.add_section_context ctx)
-    optpoly
+type universe_name_decl = universe_source * Nametab.universe_id
```
ppedrot: I think that while you're at it, you should store those as lists. Otherwise you end up declaring a crapload of libojects that take memory space for no good reason.

SkySkimmer: That's nontrivial because I use `add_leaf id` where `id` is the user name of the universe.
Do you still want to push for it?

ppedrot: Yes. Don't rely on the name of the object for the universe, simply add an anonymous leaf. In general it is a bad idea to do so, there is no real reason to use the same representation for libobject names and user-facing data.

SkySkimmer: Changed.

ppedrot (approved):

ppedrot:
CI failure is spurious, I'll merge tomorrow if no objections.

PR #8586, opened 2018-09-28
Pack the checker.

SkySkimmer:
Not sure how useful this is.
Note: because stuff put into a pack need to be compiled with -for-pack we can't just put select (c)lib/ modules in it, so we end up linking with all lib/ and clib/.
I don't know to modify dune, I think it's still non-packed.


SkySkimmer:
Ideally we could use this to remove the hack in dune apidoc target but that will require dune modifications.

ejgallego:
Well, does this have any effect? No source file is changed to use the pack so what is the point?

ejgallego:
I would expect at least `checker/main.ml` to have to do `open Checkerlib` Also, please use a pack name that corresponds to the directory name if possible [not a hard requirement].

Anyways, if we go this route I can do the same for Dune without needing this PR, see #8588 which would override this PR.

ejgallego:
We can close this tho.

PR #8588, opened 2018-09-28
[dune] Pack checker to avoid [odoc] problems + build doc for plugins.

ejgallego:
We also build documentation for plugins, for example `Ltac_plugin` is often used in other plugins.

### DIFF
--- a/checker/dune
+++ b/checker/dune
@@ -3,24 +3,30 @@
 (rule (copy %{project_root}/kernel/esubst.ml esubst.ml))
 (rule (copy %{project_root}/kernel/esubst.mli esubst.mli))
 
+; Careful with bug https://github.com/ocaml/odoc/issues/148
+;
+; If we don't pack checker we will have a problem here due to
+; duplicate module names in the whole build.
 (library
- (name checker)
- (public_name coq.checker)
+ (name checklib)
+ (public_name coq.checklib)
  (synopsis "Coq's Standalone Proof Checker")
- (modules values analyze names esubst)
- (wrapped false)
+ (modules :standard \ main votour)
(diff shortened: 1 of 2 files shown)


checker/dune
```diff
@@ -3,24 +3,30 @@
...
  (libraries coq.lib))
 
 (executable
  (name main)
  (public_name coqchk)
- (modules :standard \ votour values analyze names esubst)
- (modules_without_implementation cic)
- (libraries coq.checker))
+ (modules main)
+ (flags :standard -open Checklib)
```
Zimmi48: So we pack in Dune but not in make-based build, and thus we must open Checklib with a flag instead of modifying a source file. Is that correct? Why this choice?
When we start packing other parts like `kernel/` to have more control on the public API, we are going to have to pack it with both Dune and the make-based build (unless we wait to get rid of the make-based build first). And then, there will be an inconsistency.

Zimmi48 (approved):
My comment being a mere remark, I'll merge this PR if you confirm that you are positive about this approach.

checker/dune
ejgallego: That is correct. This choice is just an implementation of #8586 in Dune. As indeed packing is needed to workaround the `odoc` bug, it makes sense to do only so in Dune. Anyways would the make-based maintainers wanted to pack any library I am OK with it.

I am not sure why do you say there will be an inconsistency, can you explain more? [Checker is really a separate component]

I tried to pack indeed `kernel` using the make-based system, however I ran into problems so I prefer to wait for the make-based system to go away, as in particular using the dune version of packing will provide an easy migration path for plugins thanks to `(wrapped (transition ...))`.

Issue #4781, opened 2016-06-04
[abstract] typeclass resolution hints give very strange behavior (and an abundance of Anomaly: Uncaught exception Not_found. Please report.) (regression since 8.4)


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#4781
From: @JasonGross
Reported version: 8.6

See also: [BZ#4496](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%234496%22)
See also: [BZ#4493](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%234493%22)
See also: [BZ#4863](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%234863%22)

tchajed:
This bug has (at least partially) regressed between v8.6 and v8.7.1 as of d0e05a1964f.

SkySkimmer:
That commit is in 8.8, not 8.7 though.

tchajed:
Fair enough - I'll bisect to more precisely say where the regression is soon.

SkySkimmer:
Was it actually fixed? We have `test-suite/bugs/opened/4781.v` since 633ed9c52.

tchajed:
Nope, looks like the bug is in v8.6.1 and v8.7.1 as well, so it seems unlikely it was ever fixed?

SkySkimmer:
It seems the following 2 still get uncaught `Not_found` but the rest succeed or fail with catchable errors
```coq
  Fail let term := constr:(I) in
  let T := type of term in
  let x := constr:((_ : abstract_term term) : T) in
  let x := type of x in
  pose x. (* should succeed *)
  Fail let term := constr:(I) in
  let T := type of term in
  let x := constr:(_ : abstract_term term) in
  let x := type of x in
  pose x. (* should succeed *)
```

Issue #4674, opened 2016-04-17
instantiate error message could be more helpful


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#4674
From: @tchajed
Reported version: 8.5
CC:   @tchajed

SkySkimmer:
Attached file is dead

PR #8590, opened 2018-09-28
Functionalizing calls to the environment in Himsg

herbelin:
This shall eventually allow to call Himsg at any point of the execution, independently of the exact current global environment.

This is a small step in direction of better controlling the flow of exceptions in error reporting.

For instance, we turn exceptions into messages in at (presumably) a very precise point of the execution while we could process all kinds of errors, and not only the one dealt with `process_vernac_interp_error` at the same phase (i.e. typically in the `CErrors.iprint` of `vernac_loop`, as far as I understand at the current time).

**Kind:** infrastructure.

### DIFF
--- a/pretyping/indrec.ml
+++ b/pretyping/indrec.ml
@@ -40,7 +40,7 @@ type recursion_scheme_error =
   | NotMutualInScheme of inductive * inductive
   | NotAllowedDependentAnalysis of (*isrec:*) bool * inductive
 
-exception RecursionSchemeError of recursion_scheme_error
+exception RecursionSchemeError of env * recursion_scheme_error
 
 let named_hd env t na = named_hd env (Evd.from_env env) (EConstr.of_constr t) na
 let name_assumption env = function
@@ -86,7 +86,7 @@ let mis_make_case_com dep env sigma (ind, u as pind) (mib,mip as specif) kind =
     if not (Sorts.List.mem kind (elim_sorts specif)) then
       raise
 	(RecursionSchemeError
-           (NotAllowedCaseAnalysis (false, fst (UnivGen.fresh_sort_in_family kind), pind)))
+           (env, NotAllowedCaseAnalysis (false, fst (UnivGen.fresh_sort_in_family kind), pind)))
   in
   let ndepar = mip.mind_nrealdecls + 1 in
 
@@ -490,7 +490,7 @@ let mis_make_indrec env sigma ?(force_mutual=false) listdepkind mib u =
 let build_case_analysis_scheme env sigma pity dep kind =
   let (mib,mip) = lookup_mind_specif env (fst pity) in
   if dep && not (Inductiveops.has_dependent_elim mib) then
-    raise (RecursionSchemeError (NotAllowedDependentAnalysis (false, fst pity)));
+    raise (RecursionSchemeError (env, NotAllowedDependentAnalysis (false, fst pity)));
   mis_make_case_com dep env sigma pity (mib,mip) kind
 
 let is_in_prop mip =
@@ -550,9 +550,9 @@ let check_arities env listdepkind =
        let kelim = elim_sorts (mibi,mipi) in
        if not (Sorts.List.mem kind kelim) then raise
 	 (RecursionSchemeError
-          (NotAllowedCaseAnalysis (true, fst (UnivGen.fresh_sort_in_family kind),(mind,u))))
+          (env, NotAllowedCaseAnalysis (true, fst (UnivGen.fresh_sort_in_family kind),(mind,u))))
        else if Int.List.mem ni ln then raise
-	 (RecursionSchemeError (NotMutualInScheme (mind,mind)))
+         (RecursionSchemeError (env, NotMutualInScheme (mind,mind)))
        else ni::ln)
 	    [] listdepkind
   in true
@@ -561,7 +561,7 @@ let build_mutual_induction_scheme env sigma ?(force_mutual=false) = function
   | ((mind,u),dep,s)::lrecspec ->
       let (mib,mip) = lookup_mind_specif env mind in
       if dep && not (Inductiveops.has_dependent_elim mib) then
-        raise (RecursionSchemeError (NotAllowedDependentAnalysis (true, mind)));
+        raise (RecursionSchemeError (env, NotAllowedDependentAnalysis (true, mind)));
       let (sp,tyi) = mind in
       let listdepkind =
 	((mind,u),mib,mip,dep,s)::
@@ -572,7 +572,7 @@ let build_mutual_induction_scheme env sigma ?(force_mutual=false) = function
                 let (mibi',mipi') = lookup_mind_specif env mind' in
 		((mind',u'),mibi',mipi',dep',s')
 	      else
-		raise (RecursionSchemeError (NotMutualInScheme (mind,mind'))))
+                raise (RecursionSchemeError (env, NotMutualInScheme (mind,mind'))))
 	   lrecspec)
       in
       let _ = check_arities env listdepkind in
@@ -582,7 +582,7 @@ let build_mutual_induction_scheme env sigma ?(force_mutual=false) = function
 let build_induction_scheme env sigma pind dep kind =
   let (mib,mip) = lookup_mind_specif env (fst pind) in
   if dep && not (Inductiveops.has_dependent_elim mib) then
-    raise (RecursionSchemeError (NotAllowedDependentAnalysis (true, fst pind)));
+    raise (RecursionSchemeError (env, NotAllowedDependentAnalysis (true, fst pind)));
   let sigma, l = mis_make_indrec env sigma [(pind,mib,mip,dep,kind)] mib (snd pind) in
     sigma, List.hd l
 
--- a/pretyping/indrec.mli
+++ b/pretyping/indrec.mli
@@ -20,7 +20,7 @@ type recursion_scheme_error =
   | NotMutualInScheme of inductive * inductive
   | NotAllowedDependentAnalysis of (*isrec:*) bool * inductive
 
-exception RecursionSchemeError of recursion_scheme_error
+exception RecursionSchemeError of env * recursion_scheme_error
 
 (** Eliminations *)
 
--- a/tactics/inv.ml
+++ b/tactics/inv.ml
@@ -495,7 +495,7 @@ let raw_inversion inv_kind id status names =
 (* Error messages of the inversion tactics *)
 let wrap_inv_error id = function (e, info) -> match e with
   | Indrec.RecursionSchemeError
-      (Indrec.NotAllowedCaseAnalysis (_,(Type _ | Set as k),i)) ->
+      (_, Indrec.NotAllowedCaseAnalysis (_,(Type _ | Set as k),i)) ->
       Proofview.tclENV >>= fun env ->
       Proofview.tclEVARMAP >>= fun sigma ->
       tclZEROMSG (
--- a/vernac/explainErr.ml
+++ b/vernac/explainErr.ml
@@ -76,8 +76,8 @@ let process_vernac_interp_error exn = match fst exn with
       wrap_vernac_error exn (Himsg.explain_module_error e)
   | Modintern.ModuleInternalizationError e ->
       wrap_vernac_error exn (Himsg.explain_module_internalization_error e)
-  | RecursionSchemeError e ->
-      wrap_vernac_error exn (Himsg.explain_recursion_scheme_error e)
+  | RecursionSchemeError (env,e) ->
+      wrap_vernac_error exn (Himsg.explain_recursion_scheme_error env e)
   | Cases.PatternMatchingError (env,sigma,e) ->
       wrap_vernac_error exn (Himsg.explain_pattern_matching_error env sigma e)
   | Tacred.ReductionTacticError e ->
--- a/vernac/himsg.ml
+++ b/vernac/himsg.ml
@@ -601,12 +601,12 @@ let explain_var_not_found env id =
   spc () ++ str "in the current" ++ spc () ++ str "environment" ++ str "."
 
 let explain_wrong_case_info env (ind,u) ci =
-  let pi = pr_inductive (Global.env()) ind in
+  let pi = pr_inductive env ind in
   if eq_ind ci.ci_ind ind then
     str "Pattern-matching expression on an object of inductive type" ++
     spc () ++ pi ++ spc () ++ str "has invalid information."
   else
-    let pc = pr_inductive (Global.env()) ci.ci_ind in
+    let pc = pr_inductive env ci.ci_ind in
     str "A term of inductive type" ++ spc () ++ pi ++ spc () ++
     str "was given to a pattern-matching expression on the inductive type" ++
     spc () ++ pc ++ str "."
@@ -1156,24 +1156,24 @@ let error_large_non_prop_inductive_not_in_type () =
 
 (* Recursion schemes errors *)
 
-let error_not_allowed_case_analysis isrec kind i =
+let error_not_allowed_case_analysis env isrec kind i =
   str (if isrec then "Induction" else "Case analysis") ++
   strbrk " on sort " ++ pr_sort Evd.empty kind ++
   strbrk " is not allowed for inductive definition " ++
-  pr_inductive (Global.env()) (fst i) ++ str "."
+  pr_inductive env (fst i) ++ str "."
 
-let error_not_allowed_dependent_analysis isrec i =
+let error_not_allowed_dependent_analysis env isrec i =
   str "Dependent " ++ str (if isrec then "induction" else "case analysis") ++
   strbrk " is not allowed for inductive definition " ++
-  pr_inductive (Global.env()) i ++ str "."
+  pr_inductive env i ++ str "."
 
-let error_not_mutual_in_scheme ind ind' =
+let error_not_mutual_in_scheme env ind ind' =
   if eq_ind ind ind' then
-    str "The inductive type " ++ pr_inductive (Global.env()) ind ++
+    str "The inductive type " ++ pr_inductive env ind ++
     str " occurs twice."
   else
-    str "The inductive types " ++ pr_inductive (Global.env()) ind ++ spc () ++
-    str "and" ++ spc () ++ pr_inductive (Global.env()) ind' ++ spc () ++
+    str "The inductive types " ++ pr_inductive env ind ++ spc () ++
+    str "and" ++ spc () ++ pr_inductive env ind' ++ spc () ++
     str "are not mutually defined."
 
 (* Inductive constructions errors *)
@@ -1194,12 +1194,12 @@ let explain_inductive_error = function
 
 (* Recursion schemes errors *)
 
-let explain_recursion_scheme_error = function
+let explain_recursion_scheme_error env = function
   | NotAllowedCaseAnalysis (isrec,k,i) ->
-      error_not_allowed_case_analysis isrec k i
-  | NotMutualInScheme (ind,ind')-> error_not_mutual_in_scheme ind ind'
+      error_not_allowed_case_analysis env isrec k i
+  | NotMutualInScheme (ind,ind')-> error_not_mutual_in_scheme env ind ind'
   | NotAllowedDependentAnalysis (isrec, i) ->
-     error_not_allowed_dependent_analysis isrec i
+     error_not_allowed_dependent_analysis env isrec i
 
 (* Pattern-matching errors *)
 
--- a/vernac/himsg.mli
+++ b/vernac/himsg.mli
@@ -29,7 +29,7 @@ val explain_mismatched_contexts : env -> contexts -> Constrexpr.constr_expr list
 
 val explain_typeclass_error : env -> typeclass_error -> Pp.t
 
-val explain_recursion_scheme_error : recursion_scheme_error -> Pp.t
+val explain_recursion_scheme_error : env -> recursion_scheme_error -> Pp.t
 
 val explain_refiner_error : env -> Evd.evar_map -> refiner_error -> Pp.t
 
--- a/vernac/indschemes.ml
+++ b/vernac/indschemes.ml
@@ -330,11 +330,10 @@ let declare_sym_scheme ind =
 (* Scheme command *)
 
 let smart_global_inductive y = smart_global_inductive y
-let rec split_scheme l =
- let env = Global.env() in
+let rec split_scheme env l =
  match l with
   | [] -> [],[]
-  | (Some id,t)::q -> let l1,l2 = split_scheme q in
+  | (Some id,t)::q -> let l1,l2 = split_scheme env q in
     ( match t with
       | InductionScheme (x,y,z) -> ((id,x,smart_global_inductive y,z)::l1),l2
       | CaseScheme (x,y,z) -> ((id,x,smart_global_inductive y,z)::l1),l2
@@ -345,7 +344,7 @@ let rec split_scheme l =
 requested
 *)
   | (None,t)::q ->
-      let l1,l2 = split_scheme q in
+      let l1,l2 = split_scheme env q in
       let names inds recs isdep y z =
         let ind = smart_global_inductive y in
         let sort_of_ind = inductive_sort_family (snd (lookup_mind_specif env ind)) in
@@ -408,12 +407,12 @@ let do_mutual_induction_scheme ?(force_mutual=false) lnamedepindsort =
   let _ = List.fold_right2 declare listdecl lrecnames [] in
   fixpoint_message None lrecnames
 
-let get_common_underlying_mutual_inductive = function
+let get_common_underlying_mutual_inductive env = function
   | [] -> assert false
   | (id,(mind,i as ind))::l as all ->
       match List.filter (fun (_,(mind',_)) -> not (MutInd.equal mind mind')) l with
       | (_,ind')::_ ->
-	  raise (RecursionSchemeError (NotMutualInScheme (ind,ind')))
+          raise (RecursionSchemeError (env, NotMutualInScheme (ind,ind')))
       | [] ->
 	  if not (List.distinct_f Int.compare (List.map snd (List.map snd all)))
           then user_err Pp.(str "A type occurs twice");
@@ -422,7 +421,8 @@ let get_common_underlying_mutual_inductive = function
             (function (Some id,(_,i)) -> Some (i,id.CAst.v) | (None,_) -> None) all
 
 let do_scheme l =
-  let ischeme,escheme = split_scheme l in
+  let env = Global.env() in
+  let ischeme,escheme = split_scheme env l in
 (* we want 1 kind of scheme at a time so we check if the user
 tried to declare different schemes at once *)
     if not (List.is_empty ischeme) && not (List.is_empty escheme)
@@ -431,7 +431,7 @@ tried to declare different schemes at once *)
     else (
       if not (List.is_empty ischeme) then do_mutual_induction_scheme ischeme
       else
-	let mind,l = get_common_underlying_mutual_inductive escheme in
+        let mind,l = get_common_underlying_mutual_inductive env escheme in
 	declare_beq_scheme_with l mind;
 	declare_eq_decidability_scheme_with l mind
     )


ejgallego (commented):
Looks very good, I will review more carefully later. 

Out of curiosity, did you find a problem with the errors messages in the wild?

herbelin:
> Looks very good, I will review more carefully later.

This one is simple. The next step which is abstracting over the tables used for printing (implicit arguments, notations, ...) seems more involved.

> Out of curiosity, did you find a problem with the errors messages in the wild?

Not in this PR. An experience at avoiding the early call to `process_vernac_interp_error` seems to have a good impact on #4781 failing less though.

Nevertheless, something like `process_vernac_interp_error` is necessary. The errors from the kernel are purely logical but we need to know the current state of notations, implicit arguments, scopes, etc. to print them. When to attach the state to the kernel exception? How to statically ensure that the appropriate state has been attached?

For the first question, maybe `Stm.stm_purify` is a good place?? For the second question, I don't see a way to it. Or, but this is a bit radical, forbidding direct calls to `Safe_typing` and ensures that all such calls goes through an intermediate level lying after `state.ml` so that a state can be attached?? Or, other alternative, that `Safe_typing` takes an abstract `state` as argument and add this state to each exception it raises?? Any opinion?

ejgallego:
> When to attach the state to the kernel exception? How to statically ensure that the appropriate state has been attached?

Oh, I see what you mean. Indeed the document manager has no problem in knowing the proper input state. So I guess that for exceptions that come without state the document manager you print with the input state, and for exceptions that did modify the `evar_map` etc... they should attach it.

I think at some point indeed we'd love functionalize the state fully and these flow problems would be clear. for now, we ought to remove`Global.env ()` as much as possible first. Then we will put the summary.

So I think we want to do 2 things here:
- Make the STM / document manager use a different traversal strategy that is more clear, thread the state.
- Move more and more state to be functional, for example `Global.env ()`, and at last the summary.

Once we have that in place, we can indeed discuss more about error handling.

ejgallego (approved):
Nice cleanup / improvement.

ppedrot (approved):
Very welcome change.

herbelin:
> I think at some point indeed we'd love functionalize the state fully and these flow problems would be clear. for now, we ought to removeGlobal.env () as much as possible first. Then we will put the summary.

> Move more and more state to be functional, for example `Global.env ()`, and at last the summary.

Well, I'm already now continuing with adding an argument `States.state` to every printing function and I'm wondering how far I should go with threading the state deeper and deeper. For instane, do I go up to change `exception PretypeError of env * Evd.evar_map * pretype_error` into `exception PretypeError of States.state * env * Evd.evar_map * pretype_error`? Do I go up to adding a `States.state` to, say, `w_unify` and similar functions, just because these functions are liable to raise `PretypeError`.

Personally, I'm also ready to go in this direction, but is everyone ok if I push state threading so deeply. @ppedrot, @mattam82, @gares, ... what do you think?

herbelin:
Actually, why not to move `States` and `Summary` before the kernel, so that the kernel can attach the state to every `TypeError` exception it raises?

ejgallego:
> I'm already now continuing with adding an argument `States.state` to every printing function

Umm, I am not sure that is the right way to proceed @herbelin ; certainly printing shouldn't mess with the summary IMHO (yet). In order to do so you will have to call `freeze` and that is not something you want lower-level layers to take care of.

I'd be great to see first what the actual problem is, because as of now the architecture should work fine, except for a few places that may do some strange stuff.

Personally, I'd prefer we first make the upper layers more aware of functional state than the other way around. And here we need to take care of `Proof_global` which is just pending on the STM here #6171

ejgallego:
> Actually, why not to move `States` and `Summary` before the kernel, so that the kernel can attach the state to every `TypeError` exception it raises?

I think we want to go the total opposite way IMO, that is to say, to move state the further away from the kernel we can, so lower layers are functional and the uppers layers take care of the transition.

The kernel is mostly functional today and that is a good thing.

Note that "state" handling is really something the upper layers do care about; indeed as I mentioned previously the current scheme should be correct, apart from STM bugs. Maybe when I have time to push the new scheduler things should become clearer, as today the traversal is backwards so state handling is a bit messy.

ejgallego:
So summarizing, this is how I would proceed [in order]

1. Remove `Proof_global` #6171 , and `Global.env ()`
2. Thread state in the STM.
3. Push `Summary` one layer up (to `engine`) that would mean that mainly `Nametab` , `GOptions`, and library state would have to be handled functionally.
4. `engine` is mostly clear of `Summary`, so next step would be pretyper, which should also be easier, `interp` is also doable.

Once we are there we'd be almost done IMHO, the real work are 1, 2, 3, in particular `Global.env ()` and 3. will be loads of work.

herbelin:
> I'd be great to see first what the actual problem is, because as of now the architecture should work fine, except for a few places that may do some strange stuff.

Initially, I wanted to be able to call `CErrors.print exn` from somewhere in `proofs` and it was not working because it was raising a `PretypeError` which was not registered to `CErrors`. The reason `PretypeError` is not registered is that it is treated specially in `process_vernac_interp_error`. The reason for `process_vernac_interp_error` is that exceptions such as `PretypeError` are state-dependent (i.e. dependent on notations, et>c.), i.e. that `CErrors.print` is not able to print an error without knowing in which state the exception makes sense, which is a strong limitation.

> > Actually, why not to move States and Summary before the kernel, so that the kernel can attach the state to every TypeError exception it raises?

> I think we want to go the total opposite way IMO, that is to say, to move state the further away from the kernel we can, so lower layers are functional and the uppers layers take care of the transition.

I think that there is a misunderstanding. When I'm saying moving `States` earlier, I'm not saying being imperative. The point of `States` is not be imperative, it is to _glue together_ all local (functional) states into one big record, as you did with `modify_summary` and `project_from_summary`.

> I think we want to go the total opposite way IMO, that is to say, to move state the further away from the kernel we can, so lower layers are functional and the uppers layers take care of the transition.

If you make the lower layers functional, you will have the kernel with depends on an `env`, you will have implicit arguments which depend on some `type impargs_state = implicits_list list GlobRef.Map.t`, you will have notations which depend on some type `notation_state` which is the type of  `Notations.freeze()`. If you have a function which manipulate an `env`, an `impargs_state`, a `notation_state`, and so on, you certainly don't want to pass it as many arguments as the number of individual types of tables it depends on. So, you want to collect all of this into a `States.state` and this is why I find that it is good to have the code dealing with the functional registering of states low in the hierarchy.

For instance, printing functions are terrible, they depend on a lot of individual frozen (functionally-threaded) states. So, what typical functional type would you give to `pr_constr_env` if not something of the form `States.state -> env -> evar_map -> constr -> Pp.t`?

> So summarizing, this is how I would proceed [in order]

> 1. Remove `Proof_global` #6171 , and `Global.env ()`
> 2. Thread state in the STM.
> 3. Push `Summary` one layer up (to `engine`) that would mean that mainly `Nametab` , `GOptions`, and library state would have to be handled functionally.
> 4. `engine` is mostly clear of `Summary`, so next step would be pretyper, which should also be easier, `interp` is also doable.

Looks good (but I guess that you intend that `modify_summary` and `project_from_summary` remain in a lower layer?).

ejgallego:
> I wanted to be able to call `CErrors.print exn` from somewhere in `proofs` and it was not working because it was raising a `PretypeError` which was not registered to `CErrors`

Why did you need to print an error in `proofs?



> you certainly don't want to pass it as many arguments as the number of individual types of tables it depends on.

On the contrary, I think I'd like very much for the types of functions to be informative about what the need/produce. This enables for example unit testing.

> I find that it is good to have the code dealing with the functional registering of states low in the hierarchy.

Umm, I am not sure I see why you would like that to be low in the hierarchy, the only component that knows about state accurately is the document manager which is the one properly executing the commands.

Maybe you could sketch a more concrete proposal, such as how the module would look and what would be the interface? This is a large change so I think it makes sense to be sure we all agree on the design before spending what could be weeks of coding.



> they depend on a lot of individual frozen (functionally-threaded) states.

They are not bad right? AFAICS they mostly depend on the env, nametab and some stuff about implicits plus the printing options themselves, right?



> if not something of the form `States.state -> env -> evar_map -> constr -> Pp.t`?

I would be already happier with this type I guess, but I'd even be more happy with a type such as:
```
type printing_env = {
   env : ...;
   nametab : ...;
   implicit_state: ...;
...
}
```
which leads us to

> Looks good (but I guess that you intend that `modify_summary` and `project_from_summary` remain in a lower layer?).

At first they'll have to be there, but if we could that plan they would indeed live in `stm`, and the document manager would be the one to project / compose state.

The thing is that you want to be finer for example when working with threads / parallel checking, so indeed vernacular commands could be refined as we have discussed a few times:
```
type vernac_command =
  | Printing command (printing_state -> Pp.t)
  | Parsing_command (parsing_state -> parsing_state)
....
```

ejgallego:
> the kernel doesn't know about the nametab so the nametab must be added to the exceptions in whatever higher layer is calling the kernel and has access to the nametab.

Umm, I think that for the moment I'd be happy for the error handler [that is, the document manager] to fill in the missing state with the input one [as it has access to all state]. That should work pretty well IMO, tho it could be the case that very complex compound commands need indeed to re-process errors. I guess this is quite related to how we'd like the error registration API to be.

On the other hand, I guess that our printing process feels a bit heavy, in particular it seems to me that we have a lot of different kinds of names [I've never been able to follow all of them] and maybe the right solution to this question would be to reduce the dependency on the `Nametab` itself. But indeed OMMV.



> For the kernel calls are mostly centralised through Declare stuff so it could be done but eg the pretyper (unification) shouldn't know about notations but will raise errors which need to be printed using notations, and it gets called all over the place.

Yeah, but in this case the upper layer [document manager should be able to fill that in, unless indeed if you modify the notations locally and then call the pretyper, but we could provide a proviso for this] Currently we have this `protect` stuff, but I am not sure how well it can work in the general case as reasoning gets quite complex due to impurity.

ejgallego:
In fact, this is the problem we have, the `kernel` has no way to register a printer for its own errors.

herbelin:
> Why did you need to print an error in `proofs?

For `instantiate`, in `Evar_refiner`, I want to report a `PretypeError` in case the instance is ill-typed (see wish #4674).

herbelin:
> I would become add_anonymous_leaf: obj -> lib_state -> lib_state.

Ah, ok, so you mean decoupling the modification of the `lib_state` and the modification of a `summary`. So you would `cache` and register the `load` and `open` functions separatedly from registering to the `lib_state`. That seems possible too.

herbelin:
> And why do you need to print the error, you could just forward it right?

I could forward it, but to forward it, I would need to declare a new error name, say `InstantiateTacError of pretype_error` and I would need to add an interpreter for this new error in `process_vernac_interp_error` which is currently the only handler to be aware of `PretypeError`.

That `PretypeError` is known only from `process_vernac_interp_error` due to imperative state synchronization problems is actually defeating the use of an extensible `CErrors` error handling mechanism.

Note: another approach would be to move the part of `Himsg` which prints `PretypeError` in `pretyping` so that `Evar_refiner` can use it and raise a `UserError` without having to declare a new error.

herbelin:
> I think that error printing by the intermediate layers should not be supported, it is an anti-pattern and if we do it we will encourage others to do so. After all you are registering a new error.

At the beginning, I liked `Himsg` which provides with a global overview of error message, and I still like it for this reason. However, in practice, having to define new errors for any tactic has been a brake to fast prototyping, leading me to prefer raising errors which do not have to print terms just because the cost of adding a new error and a new handler in `Himsg`, and changing many files, including `mli`, was discouraging.

ejgallego:
> However, in practice, having to define new errors for any tactic has been a brake to fast prototyping, leading me to prefer raising errors which do not have to print terms just because the cost of adding a new error and a new handler in `Himsg`, and changing many files, including `mli`, was discouraging.

Well adding a new error to the codebase is not something that should be too light IMHO; in principle all errors should be documented for example.

herbelin:
> So summarizing, this is how I would proceed [in order]
> 
> 1. Remove `Proof_global` #6171 , and `Global.env ()`
> 2. Thread state in the STM.
> 3. Push `Summary` one layer up (to `engine`) that would mean that mainly `Nametab` , `GOptions`, and library state would have to be handled functionally.
> 4. `engine` is mostly clear of `Summary`, so next step would be pretyper, which should also be easier, `interp` is also doable.
> 
> Once we are there we'd be almost done IMHO, the real work are 1, 2, 3, in particular `Global.env ()` and 3. will be loads of work.

Let me come back on this program. There are 376 instances of `Global.env`. In #8598, I added a `state` argument to 291 calls to `pr_*constr_env` in a few hours. So, if we know exactly how we want to replace the calls to `Global.env` by calls to, say, a `global_env` parameter, a few hours shall probably be enough too.

So, to me, the only question is to take a decision on how we want to thread `Global.env`.

Do we want to pass `Global.env` separately from the rest of the state, e.g. because it is "environment-passing" and not "state-passing"-style? Or do we want to pass `Global.env` as part of a global state? Or do we want to split the state into small states, which is something that you seem to have in mind.

Let's take an example, again from `impargs.ml`. The function `declare_implicits_gen`:
- indirectly does a call to `Global.env` (via `compute_global_implicits`)
- read the state of implicit arguments
- write an extended state of implicit arguments
- read the `lib_state`
- write the `lib_state`

The original code is:
```ocaml
(* val compute_global : implicits_flags -> implicits_list *)
(* val add_anonymous_leaf : obj -> unit *)
(* val declare_implicits_gen : 
    implicit_discharge_request -> implicits_flags -> global_reference -> unit *)

let declare_implicits_gen req flags ref =
  let imps = compute_global_implicits flags ref in
  add_anonymous_leaf (inImplicits (req,[ref,imps]))
```
How to you see its translation:
```ocaml
(* val compute_global : global_env:env -> implicits_flags -> implicits_list *)
(* val add_anonymous_impargs_leaf :
    impargs_state:impargs_state -> lib_state:lib_state -> obj -> impargs_state * lib_state *)
(* val declare_implicits_gen :
    impargs_state:impargs_state -> lib_state:lib_state -> global_env:env ->
    implicit_discharge_request -> implicits_flags -> global_reference -> impargs * lib_state *)

let declare_implicits_gen ~impargs_state ~lib_state ~global_env req flags ref =
  let imps = compute_global_implicits ~global_env flags ref in
  add_anonymous_impargs_leaf ~impargs_state ~lib_state (inImplicits (req,[ref,imps]))
```
or
```ocaml
(* val compute_global : global_env:env -> implicits_flags -> implicits_list *)
(* val add_anonymous_impargs_leaf : impargs_state:impargs_state -> lib_state:lib_state -> obj ->
    impargs_state -> lib_state *)
(* val declare_implicits_gen : state:state ->
    implicit_discharge_request -> implicits_flags -> global_reference -> state *)

let declare_implicits_gen ~state req flags ref =
  let imps = compute_global_implicits ~global_env:(project_state global_env_tag state) flags ref in
  let impargs_state,lib_state = add_anonymous_impargs_leaf
     ~impargs_state:(project_state impargs_tag state)
     ~lib_state:(project_state lib_state_tag state)
    (inImplicits (req,[ref,imps])) in
  modify_state lib_state_tag (modify_state impargs_tag state)
```
or, moving the use of projections/mutations in the final place of use in the subfunctions,
```ocaml
(* val compute_global : state:state -> implicits_flags -> implicits_list *)
(* val add_anonymous_leaf : state:state -> obj -> state *)
(* val declare_implicits_gen : state:state ->
    implicit_discharge_request -> implicits_flags -> global_reference -> state *)

let declare_implicits_gen ~state req flags ref =
  let imps = compute_global_implicits ~state flags ref in
  ladd_anonymous_leaf ~state (inImplicits (req,[ref,imps])) in
```  
or something else?

If we add replacing the field `Proofview.Goal.env : Environ.env` by `Proofview.Goal.state : state`, so that all changes to the state made within a goal remains local the proof (until the proof ends and the "side effects" are propagated), and if we add adding a `state` argument to all errors referring to a `constr` (e.g. in `PretypeError`), then, all in all, I believe that the whole functionalisation of the state can be done methodically in less than 20 hours, and even probably less.

What do you think?

SkySkimmer:
Passing around a big `state` type seems unsatisfying. To me the main draw of functionalizing is controlling the input and output of functions, so if we're just passing everything around what's the difference with keeping the global refs and saying "we're implicitly passing the entire memory around" (IIRC that's how the haskell IO monad is explained)?
For functions which don't modify the state it would let us statically guarantee that but for those which do modify the state I don't see the benefit.

Then if we try to control how much state we pass to functions based on what they need it seems wrong to pass implicit argument, notation, nametab, etc information to the pretyper just for printing. Maybe what we want is to abstract over printing and then pass the appropriate subset of the state.

I guess that would look like some module (which I'm going to call `Printy`):
~~~ocaml
type t (* morally a collection of closures like (env -> evar_map -> constr -> Pp.t), (globref -> Pp.t), etc *)

val pr_constr : t -> env -> evar_map -> constr -> Pp.t
(* etc, basically Printer but with extra [t] arguments *)

val make : (* some functions sufficient to define pr_constr, pr_lconstr and so on *) -> t

val debug : t (* for anomalies? maybe debug printer is not necessary *)
~~~
Because we need evar_map for printing it seems this would have to be in `engine` somewhere after evd.

Then for instance a pretyper function would look like
~~~ocaml
type pretype_env = { env : Environ.env; coercions : Coercion.env; ??? }
(* maybe include the Printy.t in pretype_env? *)
val understand_tcc :  Printy.t -> pretype_env -> evar_map -> glob_constr -> evar_map * constr
~~~
and we would register exceptions with access to printy:
~~~ocaml
val register_exn_printy : (Printy.t -> exn -> Pp.t) -> unit
val print_exn_printy : Printy.t -> exn -> Pp.t
~~~
(all calls to `CErrors.print` are after Printy except for 
- https://github.com/coq/coq/blob/1bde8c0912ed1129e71ffe20299ac89299492ba5/kernel/nativelib.ml#L155
- https://github.com/coq/coq/blob/1bde8c0912ed1129e71ffe20299ac89299492ba5/kernel/safe_typing.ml#L821
- https://github.com/coq/coq/blob/1bde8c0912ed1129e71ffe20299ac89299492ba5/checker/checker.ml#L310

so I think this would be feasible)

The kernel doesn't have access to `evar_map` so either it would have some cut down `KPrinty` module (which can be gotten from the normal `Printy`) or maybe we should just make the error reporting exception-less for the kernel (?!) ala `val infer : env -> constr -> (unsafe_judgement, type_error) result`. Errors without exceptions everywhere seems like it would be syntax-heavy but maybe just the kernel would be fine.

Anyways to recap I'm very dubious about passing blob states around. The rest is just trying to figure out how to progress without the blobs.

ejgallego:
> So, if we know exactly how we want to replace the calls to `Global.env` by calls to, say, a `global_env` parameter, a few hours shall probably be enough too.

That would be great, however last time we looked situation was a bit more subtle, due to different way subsystems use [and update] different envs. @ppedrot may want to say something here, but IMHO threading the `env` looks more prioritary than for example the printing info.

ejgallego:
> Or do we want to pass `Global.env` as part of a global state?

The env is not always exactly synchronized with the state, thus I do favor handling it on its own.

ejgallego:
> What do you think?

I really like the first option, even if a bit verbose [which could be fixed later], makes really clear what is going on, which is the main advantage of making the code functional, in the codebase so large we cannot really track what's being modified anymore.

> it seems wrong to pass implicit argument, notation, nametab, etc information to the pretyper just for printing

Indeed, the pretyper should not have to print elaborated terms, and errors should be done by some kind of closure towards the upper layers IMHO.

> Anyways to recap I'm very dubious about passing blob states around. The rest is just trying to figure out how to progress without the blobs.

I do agree. As for the proposal about error printing it looks interesting. The key question is how we extend from the upper components, maybe after all we should do as you say an assume a fixed set.

herbelin:
> That would be great, however last time we looked situation was a bit more subtle, due to different way subsystems use [and update] different envs. @ppedrot may want to say something here, 

I just looked at `pretyping/typeclasses.ml` and the situation is difficult. `Global.env()` is used in the `discharge_function` and `rebuild_function` which are not assumed to have access to the global env. So I see two solutions:
- either we precompute the informations that the accesses to the global env are supposed to give (this seems to be to know which classes remain at section closing but I'm suffering in reading the code)
- either we pass the global env to the discharge and rebuild functions (the one at the opening of the section); this would not be absurd per se if the first option is too difficult

Also, `typeclasses.ml` mixes typing (e.g. `build_subclasses`) and library management (e.g. `declare_instance`). Maybe `pretyping/` itself should make a difference between what is typing (`pretyping.ml`, ...) and what is library/command management (`recordops.ml`, ...).

> but IMHO threading the env looks more prioritary than for example the printing info.

It depends what you call prioritary. If it has to be done and someone is ready to do it, it does not have to wait (but of course, we need to agree on the global model).

SkySkimmer:
>which are not assumed to have access to the global env

Aren't they?

ejgallego:
Thanks for the very nice analysis @herbelin , I really hope others can give some advice here.

In general, I tend to prefer "stronger" types, that is to say, option 1 in this case I believe; but it is hard to tell.

What you say about pretyping looks sensible, but I am not familiar enough to say.

mattam82:
When is the env not in sync with the state? I understand many lower level functions don’t need the state but when is there a desync ?

herbelin:
> I really like the first option, even if a bit verbose [which could be fixed later], makes really clear what is going on, which is the main advantage of making the code functional, in the codebase so large we cannot really track what's being modified anymore.

Imagine already for `declare_mind`: it returns projections, nametab, impargs, argument scopes, lib_state. It additionally takes `env` as arguments.

Then, at the upper level, `do_mutual_inductive` shall additionally returns notations, coercions, schemes

Aren't you afraid that we get then very huge numbers of arguments or tuples. We already have an octuple. Why not, but is it really reasonable? If the objective is to track what effects are done at each step, maybe a way to annotate a single GADT-defined `state` with a list of the effects would be more generic?

Moreover, what is your model for `add_anonymous_leaf`? The first option requires an ad hoc `add_anonymous_leaf` for each object. Which type will you then give to `load_object`, `import_object`, etc?

herbelin:
@SkySkimmer 
> > which are not assumed to have access to the global env
> 
> Aren't they?

Well, it is a matter of convention. Let me rephrase my sentence by saying that if we want `discharge` and `rebuild` to have possible access to the environment/state, this means option 2 of my 2 points. Why not, indeed.

ejgallego:
> When is the env not in sync with the state? I understand many lower level functions don’t need the state but when is there a desync ?

I mean that so far there seem to be cases where we start from `st, env` and arrive to `env, env', st` [which all 3 used] or `env, st, st'` [as for example setting some option for a tactic as @herbelin tried to do.

> Aren't you afraid that we get then very huge numbers of arguments or tuples. We already have an octuple. Why not, but is it really reasonable?

I guess we should be able to group related state in records, a middle ground seems to be the reasonable part here. But my point is that we already are returning such an octuple, but in a hidden way; then in other layers we need to spend a lot of time thinking what could have possibly happened.

> Moreover, what is your model for `add_anonymous_leaf`? The first option requires an ad hoc `add_anonymous_leaf` for each object. Which type will you then give to `load_object`, `import_object`, etc?

Indeed that is a problem [which is in parallel with the libobject tweaks being done], so for example this is why I think starting with `env` could be easier, as it would provide a good overview of the data-flow, and solve quite a few bugs related to lack of threading IIANM.

mattam82:
It makes sense to me to try first threading env for sure. Indeed I didn't think that rebuild_* should not see then env, actually the point of rebuild is to do an action on the environment after everything has been discharged. It's basically env * state -> state in this case, reregistering hints.

About the "separation of states" it could be useful to look into Lenses (as in the Haskell lens library) which aims to tame access/update of functional state in general.

ejgallego:
> About the "separation of states" it could be useful to look into Lenses (as in the Haskell lens library) which aims to tame access/update of functional state in general.

That's a very good idea. Maybe we are reaching the point where we could start thinking of depending on some extra libraries in order to make our life easier [we could for example have also let for monads, etc...]

herbelin:
@ppedrot: do you have something to say on your former experiences in threading the state? (I'm preparing a CEPS and shall try to collect as much issues as possible). Same question to everyone actually.

ppedrot:
My experience is that in 95% of the code it is trivial, but then you stumble upon some piece of code that works by sheer luck or deep arcana, and relies in intricate ways on the semantics of global mutable state. If you throw in closures atop of that, all hell breaks loose. (I am looking at you, @mattam82.)

mattam82:
Which code are you referring to specifically? I think the most intricate I
can think of is handling of global universes and abstract in the STM where
surgical updates of the evar_map are needed. But there are no closures
there AFAI remember :)

ppedrot:
@mattam82 IIRC the old implementation of rewrite was a marvel of entanglement. Also, not quite Coq yet but maybe one day: I had the unfortunate obligation to have a look at Equations to port it to #8601. Net result is that it is still not ported, and that I now have PTSD.

herbelin:
@mattam82 
> It makes sense to me to try first threading env for sure. Indeed I didn't think that rebuild_* should not see then env, actually the point of rebuild is to do an action on the environment after everything has been discharged. It's basically env * state -> state in this case, reregistering hints.

OK, I shall write the ceps in this direction.

@ejgallego: I'm continuing thinking at the flow of states and substates. Currently, the model is:

```
let Vernacentries.interp = function
| VernacFixpoint fix ->
       ...
       (* calling some code eventuall calling *)
       global_env := ...;
       (* calling some code eventuall calling *)
       add_anonymous_leaf (inImplicits cst);
       (* calling some code eventuall calling *)
       add_anonymous_leaf (inNotation ntn);
...
```
If we do naive state-passing-style, we get:
```
let Vernacentries.interp state = function
| VernacFixpoint fix ->
       ...
       let state = modify_global_env ... state in
       let state = add_anonymous_leaf (inImplicits cst) state in
       let state = add_anonymous_leaf (inNotation ntn) state in
       state
...
```
Are you suggesting a model where we do:
```
let Vernacentries.interp state = function
| VernacFixpoint fix ->
       ...
       let env = modify_global_env ... (project_globenv state) in
       (* some code applied to [env] eventually doing *)
       let notation_state = add_anonymous_leaf (inNotation ntn) (project_notation state) in
       (* some code applied to [env] eventually doing *)
       let implicits_state = add_anonymous_leaf (inImplicits cst) (project_implicits state_ in
       join_env env (join_implicits implicits_state (join_notation notation_state state))
...
```
I don't feel very comfortable with that: this will leave a large freedom in when doing the join, there will be the risk of forgetting to join, there will be the risk of inavertently modifying the flow of state in the transformation. On the other side, if every "side effect" is exclusively controlled in `add_anonymous_leaf` and similar, this shall ensure that we preserve the exact behavior as of today, and we shan't have to wondering whether this or that commutation between the different `write` in different components is safe or not. (Even if I agree that being aware of which commutations between side-effects are correct is valid, but learning that would take much more time than just a trivial state-passing-style transformation of the code.)

I'm at least talking for the state-passing-style transformation. For functions which only read the state, it is no problem I think to consider substates.

SkySkimmer:
I would use individual states as much as possible. We won't know how much is possible without trying though.

herbelin:
> I would use individual states as much as possible. We won't know how much is possible without trying though.

We can try several things. I made a proposal at coq/ceps#38 so that it is possible to have `add_anonymous_leaf` returns only individual states. Please comment.

herbelin:
@ejgallego, @skymmer, let me comment on individual states and make a proposition.

### Individual states vs "standard" states

Somehow, when we make a function dependent on `Global.env`, we often do it for "simplicity". I'm pretty sure that there are many situations where we pass a full `env` to a function which in practice only need the `named_context`, or the `named_context` and the `rel_context`. If we pass it the `Global.env`, it is because it is simpler to have a well-identified collection of functions working on `env` (e.g. `push*`, `lookup*`, `it_mkLambda`, ...) rather than having to provide the same collection for each possible smaller part of the environment. This is simple for the user of `env` which does not have to know all the subtleties of the structure of an environment.

Using individual states would be more informative about what a given function is exactly requiring, but there would similarly be a cost. The caller will have to know how to get from its own state of information the exact state of information needed by the function. The caller will have to know what are the methods to project from its own state. It will have to know how to assemble the result of the various projections to build the exact kind of state expected by the given function. On the other side, with a limited notions of global data, typically `env`, `evar_map` and `state` (of type `States.state`), it is clear for the caller, which is not a specialist of the feature, what to pass to the function, without thinking, and it shall be the rôle of the function, which is contrastingly the specialist of the feature that it provides, to use the appropriate projections to get out of the global state what it needs itself.

I think that with a limited number of key data, i.e. again, `env`, `evar_map`, `state`, maintenance will be easier. If a function is taking as argument exactly what it needs, this means that when we make a little extension of the function, we shall need to change its type, and all callers will have to reconsider how to build the new kind of extended state that the function now needs. Something which would not be needed with using standardized data types for communication, such as the 3 listed above.

Again, we lose in precision, we lose in ability to easily know if we can safely commute two lines of the form `let state = f state in let state = g state in ...` but I would tend to believe that we gain in simplicity.

## Making a plan

Anyway, if I did not convince you, I'm ready to experiment having fine-grained states, but it is a too big objective for a single person. As much threading the code around a single notion of `state` can be made rather mechanically, as much finding the optimal organization of individual states which shall make things overall smooth requires thinking. So, I propose that the 3 of us take an appointment, that we reserve 2 or 3 three days sometimes in this month, days exclusively devoted to move the whole of Coq in state-passing-style, each of us working on a particular subset of commands, and, in the meantime, making a plan of exactly what we are aiming. What do you think?

ejgallego:
> when we make a function dependent on `Global.env`, we often do it for "simplicity". I'm pretty sure that there are many situations where we pass a full `env` to a function which in practice only need the `named_context`, or the `named_context` and the `rel_context`.

That indeed is a good point, and there was quite a bit of discussion in some other issue about trying to refine the notion of `env`. However it seems nobody is bold enough as to try that.

> but there would similarly be a cost. [ ...] , it is clear for the caller, which is not a specialist of the feature, what to pass to the function, without thinking, and it shall be the rôle of the function, which is contrastingly the specialist of the feature that it provides, to use the appropriate projections to get out of the global state what it needs itself.

There is a cost but also there is a benefit; the point here is what the tradeoff is. But however, the whole purpose of having state in this way is to have the _caller_ think! The caller is the responsible of composing several calls, so the more information it has the better job it can do about the composition. Lower level layers have no reason to witness a fat state that they know nothing about.

> I'm ready to experiment having fine-grained states, but it is a too big objective for a single person.

Indeed it is a difficult one. Why don't we attempt the plan I posted above and leave `State.state` for later? As I mentioned before, I think that we would fix more bugs and prepare the terroir better would we focus on having less `Global.env`. There are many tricky things going on with that one, whereas `State.state` has a simpler semantics [right now set at the vernacentries barrier].

> So, I propose that the 3 of us take an appointment, that we reserve 2 or 3 three days sometimes in this month, days exclusively devoted to move the whole of Coq in state-passing-style, each of us working on a particular subset of commands, and, in the meantime, making a plan of exactly what we are aiming. What do you think?

Sounds good however this month due to travel and other commitments it is going to be hard for me. Maybe we could meet one day when I'm back to Paris as to a start?

herbelin:
> Lower level layers have no reason to witness a fat state that they know nothing about.

Except that the `fat` of the `fat state` is invisible from the lower layers. There can be a lot of indepent code, a lot of plugins, declaring a lot of summaries and the specialized lower levels will not be aware of it. Each lower level will only see an abstract `States.state` and it will be able to access this state only with the summary tag that it has registered to the summary manager. So, from the point of view of a lower level, the "fat state" is only an abstract row variable in a tuple, and, from the tuple itself, it shall be able to see only what _it_ has explicitly declared (or what the even lower levels it depends on have made public).

> That indeed is a good point, and there was quite a bit of discussion in some other issue about trying to refine the notion of env. However it seems nobody is bold enough as to try that.


I'm not sure I'm bold enough myself, but if someone wants to try (e.g. to isolate the goal contexts and rel contexts), I shall be interested by the experience.

> Why don't we attempt the plan I posted above and leave State.state for later? As I mentioned before, I think that we would fix more bugs and prepare the terroir better would we focus on having less Global.env. There are many tricky things going on with that one, whereas State.state has a simpler semantics [right now set at the vernacentries barrier].

I'm actually working on the plan of removing the calls to `Global.env`. I found examples where we could remove the global env by reorganizing the code a bit but there are examples which still look complicated to me.

For instance, In the rebuild method of type classes, we have a goal to the `Global.env`. A first approach would be to understand why there is this call and to try to instead precompute in the "object" what this call does so that we don't need it (my first approximation about what this call does is to check which of the classes on which the current class definition depends have not survived the section, but that's to be confirmed).

A second approach is to accept that using the state while in the `rebuild_method` is legitimate. In this case, to get rid of the `Global.env`, we need to pass it functionally to the `rebuild_method`. (I have a branch which does that [here](https://github.com/herbelin/github-coq/tree/master%2Brebuild-functional-application-typeclasses)).

Another example is the hook used after a declaration to tell to register the latter as a coercion, a canonical structure, etc. After registering `c:=body`, we look in the newly updated global env to find the `body` which the kernel has processed. If we had the invariant that the `body` which the kernel registered is exactly the same as the one which was to it, we could keep the pre-kernel copy and use this in the hook. Are we ready to accept this invariant?

We could alternatively do that the kernel returns the body such as it is been processed and the hook uses it. But if we don't have the invariant above, or the kernel returning the exact body it registered, we need to look up the new body in the new global env, and such kind of lookup is pervasive in the code.

> Sounds good however this month due to travel and other commitments it is going to be hard for me. Maybe we could meet one day when I'm back to Paris as to a start?

Yes, meeting at some time, say in November, would be nice. @skyskimmer, will you come to Scalp?

ejgallego:
> Except that the `fat` of the `fat state` is invisible from the lower layers. There can be a lot of independent code, a lot of plugins, declaring a lot of summaries and the specialized lower levels will not be aware of it. **Each lower level will only see an abstract `States.state` and it will be able to access this state only with the summary tag that it has registered to the summary manager.** So, from the point of view of a lower level, the "fat state" is only an abstract row variable in a tuple, and, from the tuple itself, it shall be able to see only what _it_ has explicitly declared (or what the even lower levels it depends on have made public).

Note what I made bold. Indeed, then if that is the case, why do we send the full state in the first place? I still have the feeling that composition should happen at the upper layer as this is the only way you get static guarantees for the writers. For the readers I agree you don't care so much but composition quite weakens which can be a problem in some scenarios [for example parsing vs executing].



> Are we ready to accept this invariant?

I am not sure how this part of the system works indeed.

> November, would be nice.

November would be nice indeed!

herbelin:
> Note what I made bold. Indeed, then if that is the case, why do we send the full state in the first place?

We indeed don't need to send the full state in the first place. I was only suggesting that it would be simpler to do so: more systematic way to proceed, less complex thinking, less worries at cutting states into pieces and rearrange these pieces into pieces of a substate.

> I still have the feeling that composition should happen at the upper layer as this is the only way you get static guarantees for the writers.

Yes, sure. (Maybe we could also imagine a smart type system which tells what field of a global state a function has read or write, but I must confess that I don't have a simple solution for that yet.)

But why would we need to get static guarantees? To commute some code? Then, we could temporarily move to the caller the projections and updates made in the lower level. We then know in the upper level which projections and updates are done and we can make some code restructuration based on that. Again, I see it is a trade-off between simplicity and fine-grained static information, and my worry is that if we want the latter, it might need significantly more time than just doing the straightforward state-passing transformation.

> For the readers I agree you don't care so much but composition quite weakens which can be a problem in some scenarios [for example parsing vs executing].

Yes, I'm convinced that we want to be able to parse without executing, but I don't think this implies to have to build the whole system on very fine-grained states, even more that it would require a lot of time to be fine-grained everywhere.

ejgallego:
> less worries at cutting states into pieces and rearrange these pieces into pieces of a substate.

Indeed that would be simpler, however at some point you have to reason about composition; I suggest we allow both styles and we see as we go.

> But why would we need to get static guarantees?

The main motivation for making the state more functional indeed supporting "out-of-order" execution such as abstract or parallel proof checking. At that point, you either depend on a (usually) imprecise static analysis tool [classifier] or you want types. For UI functionality is a must. But note that indeed that has to happen in the upper layers, so for example I will be happy once we can fix some of the issues with `Proof_global`, which is so far the only point where upper layers

SkySkimmer:
I'm getting a bit lost in the github issues about state, I propose we centralise discussion in the cep.

PR #8594, opened 2018-09-29
Miscellaneous refinements and cleaning of module printing

herbelin:

This PR does the following:

- Remove the fallback printer used in case of error while printing; otherwise said, we make the assumption that the non-short mode of the module printer is not buggy.
- Preparing functionalization of printers by always passing env rather than setting env to None in short mode. This is not strictly necessary for the env which is not used for printing global references but it shall be more consistent in style when passing the nametab functionally.
- Refine the fix to #2169 by distinguishing the short and non-short printing modes.

There are still some decisions to take with respect to module printing:
- There is a flag `is_impl` which is supposed to support printing of the body of a definition but it is always false. Do we try to allow a mode for `Full Module Printing` which prints bodies?
- The syntax for printing module is hybrid between OCaml and Coq, using `Struct` and `Sig` on one side, but period-separated items. I propose to use instead a pure Coq syntax. Any objections?
- In the longer term, maybe declarations of notation should be printed and notations be used?

### DIFF
--- a/printing/printmod.ml
+++ b/printing/printmod.ml
@@ -23,8 +23,6 @@ open Goptions
     - The "rich" one, that also tries to print the types of the fields.
     The short version used to be the default behavior, but now we print
     types by default. The following option allows changing this.
-    Technically, the environments in this file are either None in
-    the "short" mode or (Some env) in the "rich" one.
 *)
 
 module Tag =
@@ -38,6 +36,8 @@ end
 let tag t s = Pp.tag t s
 let tag_definition s = tag Tag.definition s
 let tag_keyword s = tag Tag.keyword s
+
+type short = OnlyNames | WithContents
 
 let short = ref false
 
@@ -282,7 +282,7 @@ let nametab_register_modparam mbid mtb =
       List.iter (nametab_register_body mp dir) struc;
       id
 
-let print_body is_impl env mp (l,body) =
+let print_body is_impl extent env mp (l,body) =
   let name = Label.print l in
   hov 2 (match body with
     | SFBmodule _ -> keyword "Module" ++ spc () ++ name
@@ -293,9 +293,9 @@ let print_body is_impl env mp (l,body) =
 	| Def _ -> def "Definition" ++ spc ()
 	| OpaqueDef _ when is_impl -> def "Theorem" ++ spc ()
 	| _ -> def "Parameter" ++ spc ()) ++ name ++
-      (match env with
-	  | None -> mt ()
-	  | Some env ->
+      (match extent with
+         | OnlyNames -> mt ()
+         | WithContents ->
             let bl = UnivNames.universe_binders_with_opt_names (ConstRef (Constant.make2 mp l)) None in
             let sigma = Evd.from_ctx (UState.of_binders bl) in
 	    str " :" ++ spc () ++
@@ -308,10 +308,10 @@ let print_body is_impl env mp (l,body) =
 	      | _ -> mt ()) ++ str "." ++
             Printer.pr_abstract_universe_ctx sigma ctx)
     | SFBmind mib ->
-      try
-	let env = Option.get env in
+      match extent with
+      | WithContents ->
         pr_mutual_inductive_body env (MutInd.make2 mp l) mib None
-      with e when CErrors.noncritical e ->
+      | OnlyNames ->
         let keyword =
           let open Declarations in
           match mib.mind_finite with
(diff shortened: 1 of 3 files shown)


printing/printmod.ml
```diff
@@ -20,11 +20,9 @@ open Goptions
 
 (** Note: there is currently two modes for printing modules.
     - The "short" one, that just prints the names of the fields.
-    - The "rich" one, that also tries to print the types of the fields.
+    - The "rich" one, that also try to print the types of the fields.
```
ejgallego: Is this correction right?

ejgallego:
Adding @mattam82 as a secondary owner. If he's busy @herbelin I could try to give a hand to merge PRs related to `printing`.

printing/printmod.ml
```diff
@@ -347,19 +346,23 @@ let rec print_typ_expr env mp locals mty =
...
-      (* XXX: What should env and sigma be here? *)
-      let env = Global.env () in
-      let sigma = Evd.from_env env in
-      hov 2 (print_typ_expr env' mp locals me ++ spc() ++ str "with" ++ spc()
-             ++ def "Definition"++ spc() ++ str s ++ spc() ++ str ":="++ spc()
-             ++ Printer.pr_lconstr_env env sigma c)
+      let body = match extent with
+        | WithContents ->
+            let sigma = Evd.from_env env in
+            spc() ++ str ":=" ++ spc() ++ Printer.pr_lconstr_env env sigma c
```
ejgallego: Small comment: this is down to style but indeed I find that using:
```
Pp.seq [spc (); str ":=; spc (); Printer....]
```
maybe be more future-proof if we move towards a box based model.

ejgallego (approved):
LGTM.

printing/printmod.ml
herbelin: Indeed no. Reverted.

ppedrot (approved):

ppedrot:
@ejgallego do you want to be the assignee of this PR? I guess it's ready to merge.

ejgallego:
> @ejgallego do you want to be the assignee of this PR? I guess it's ready to merge.

I can take it but note that the CI failure is real so it needs fixing (cc: @herbelin )

herbelin:
The CI failure is indeed related to this PR. More generally, the current implementation is an approximation of module printing and it is difficult to understand what we exactly expect from module printing.

1. The current implementation struggles to provide an approximation of the nametab to print the types or bodies of fields. Sometimes it fails and renounces to print the type. In particular, it does support referring to a submodule.

2. What level of precision of a module do we want to print? Somehow, the current policy, which is not to print the contents of submodules, is relatively satisfactory. Do we want more? Do we want to sometimes print submodules? Do we want to print libraries which are not modules (e.g. `List`)? Do we want to use notations, coercions, implicit arguments when printing fields?

After spending a few hours on the current algorithm, I feel that, if ever we want more, we need a different algorithm which is instead to instrument `Declaremods.import_module` to print a relevant information at each `import_object`, i.e., typically, to add an optional `print_object` method which tells what to do in `Print Module`.

Any opinions?

In any case, at the current time, I'm staying with @letouzey's implementation which seems to be the best we can do with the current approach.

ejgallego:
> `Declaremods.import_module` to print a relevant information at each `import_object`, i.e., typically, to add an optional `print_object` method which tells what to do in `Print Module`.
> 
> Any opinions?

I am not familiar with modules at all but that looks pretty heavy, doesn't it? Why such a low level component such as `Declaremods` has to know about printing? Shouldn't instead provide a safe way to enumerate the module's contents.

herbelin:
> Shouldn't instead provide a safe way to enumerate the module's contents.

Enumerating is easy. What is not easy is to print. First option: you print only full names, no notations, no coercions, etc. That would be easy, just hack the printer so that it does not use any elaborated printing.

> Why such a low level component such as Declaremods has to know about printing? 

`Declaremods` already contains all the infrastructure to browse all objects of a module, including nametab, notations, etc. Why not to reuse it rather than reimplementing the same algorithm in an upper level?

ppedrot:
> Declaremods already contains all the infrastructure to browse all objects of a module, including nametab, notations, etc. Why not to reuse it rather than reimplementing the same algorithm in an upper level?

IIRC it does not handle properly functors, even though this is theoretically possible by declaring the objects living in the codomain module type.

printing/printmod.ml
```diff
@@ -440,19 +443,21 @@ let print_module with_body mp =
...
-	print_signature' true None kn mtb.mod_type))
-
-
+     try
+      if !short then raise ShortPrinting;
+      print_signature' true WithContents
+        (Global.env ()) kn mtb.mod_type
+     with e when CErrors.noncritical e ->
+      print_signature' true OnlyNames
+        (Global.env ()) kn mtb.mod_type)
```
ejgallego: Umm, this seems like the wrong place to handle this case, shouldn't instead be handled in the proper place the contents are not found?

herbelin: Assuming that I understand your question, this was the choice of @letouzey. If at least one field cannot be printed, then no fields are printed at all. So, it is decided at the toplevel, not locally, and, with that, you get a homogeneous display.

ejgallego: Umm, I'd be OK with this choice [tho it seems restrictive], however I am not sure I like the control flow here, with a super exception catch-all.

Indeed, where is the exception this captures being generated?

herbelin: I'm not sure of all exceptions which can be raised when printing in an incorrect state, but at least `Not_found`.

ejgallego: The question is "where is it raised". This is why I don't this style of control-flow, because even after looking at the code for a long time I still don't see where the jump comes from.

herbelin: > The question is "where is it raised".

In one or the other of `pr_constr` occurring in the scope of the `try`... We could add a comment...

ejgallego: It would be good to really localize this and remove this wacky exception-based control. However given that it seems this PR just preserves the old behavior I don't think it is fair to request the cleanup to be done here.

ejgallego (approved):
After further review I do not feel very comfortable with the current printing strategy here, in particular with the way control-flow is handled using "catch-all" clauses.

However this was a prior problem and in fact this PR could be seen as a step to improve this; so I will merge soon if no more comments do arrive.

PR #8596, opened 2018-09-29
[api] Cleanup `Decls`: remove unused function, move vernac helper.

ejgallego:
It seems these two functions don't belong there. We can remove one,
and place the other actually next to whether their semantics are
necessary.

Note that indeed the whole `Decls` file seems a bit suspicious, why we
do we register this information in a separate table instead of in the
main ones in `Lib` ?

### DIFF
--- a/library/decls.mli
+++ b/library/decls.mli
@@ -34,12 +34,3 @@ val variable_exists : variable -> bool
 
 val add_constant_kind : Constant.t -> logical_kind -> unit
 val constant_kind : Constant.t -> logical_kind
-
-(* Prepare global named context for proof session: remove proofs of
-   opaque section definitions and remove vm-compiled code *)
-
-val initialize_named_context_for_proof : unit -> Environ.named_context_val
-
-(** Miscellaneous functions *)
-
-val last_section_hyps : DirPath.t -> Id.t list
--- a/library/decls.ml
+++ b/library/decls.ml
@@ -11,12 +11,9 @@
 (** This module registers tables for some non-logical informations
      associated declarations *)
 
-open Util
 open Names
 open Decl_kinds
 open Libnames
-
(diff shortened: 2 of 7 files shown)


SkySkimmer:
It seems like a remnant from before lib add information about section variables (before Lib.add_section_variable).

herbelin:
> Note that indeed the whole Decls file seems a bit suspicious, why we do we register this information in a separate table instead of in the main ones in `Lib` ?

There is no summary associated to the objects `CONSTANT` and `VARIABLES` (in `Declare`). So, if we want to store non-kernel information, we need such a proper summary, no?

SkySkimmer:
It seems redundant with the sectab though.
eg "is this variable polymorphic" is available through the Lib.sectab and through Decls.

library/decls.mli
```diff
@@ -34,12 +34,3 @@ val variable_exists : variable -> bool
 
 val add_constant_kind : Constant.t -> logical_kind -> unit
 val constant_kind : Constant.t -> logical_kind
```
SkySkimmer: This `constant_kind` seems to be used only for `Typeclasses.is_instance` which has no caller and really weird semantics (look at if if you're curious). Maybe we should just remove them (constant_kind and is_instance).

SkySkimmer: Oh constant_kind is also used in dumpglob sorry. is_instance should definitely be removed though.

maximedenes (approved):
LGTM

ejgallego:
Assignee needed.

ppedrot (approved):

ppedrot:
Without consideration for the potential future changes, are you OK with this PR @herbelin? If so I'll merge it.

herbelin:
I was ok when I looked at it a few days ago (and I assume I'll do again if I'd look at it again now).

PR #8595, opened 2018-09-29
Replacing Refine.pr_constr by Termops.Internal.print_constr.

herbelin:
**Kind:** infrastructure.

This is a very short PR, removing the redundant `Refine.pr_constr` hook.

PR #8598, opened 2018-09-29
Making printing functions dependent on global state + making impargs table functional

herbelin:
**Kind:** cleanup

This is an experiment in threading the state through printing functions.

There are different ways to do it. Here, we experiment giving to printing functions a type of the form:
```coq
val pr_lconstr_env : States.state -> env -> evar_map -> constr -> Pp.t
```
but alternatives have been proposed (e.g. [here](https://github.com/coq/coq/pull/8590#issuecomment-425659103)). Since the type `States.state` is easy to grep, a global replace could be possible at some time.

We experiment using this state for implicit arguments (which are then threaded functionally to the printer and other clients of `implicits_of_global`). Depending on the direction to which we decide to go, it would be quite easy to do it for notations, argument scopes, coercions, and even probably `Nametab`.

This is on top of #8590, #8594, #8595. To see the diff specific to this PR, look at the main (huge) commit "Passing functionally a state to printing functions". The diff for `Impargs` is then shorter, see "Dealing with the table of implicit arguments functionally".

Basically all features are concerned since the need for printers is pervasive.

I tried to follow the following general adaptation scheme:
- when the `env` and `sigma` are obtained by side-effects (`Pfedit.get_current_context ()`) I get the `state` by side-effects
- when the `env` and `sigma` come from a goal, I also use a side effect (by lack of another possibility), but ideally, if the state is attached to the goal, it will be just a `Proofview.Goal.state`

Added: the re-computation of the state is costly: in O(N) for N the number of declared objects, i.e. apparently 67 (and even worth if some tables are not themselves functional). So, optimizations shall have to be considered to avoid useless recomputations. In particular this is the cause for test `4429` to fail.

### DIFF
--- a/interp/impargs.ml
+++ b/interp/impargs.ml
@@ -508,11 +508,15 @@ type implicit_discharge_request =
   | ImplInteractive of GlobRef.t * implicits_flags *
       implicit_interactive_request
 
-let implicits_table = Summary.ref GlobRef.Map.empty ~name:"implicits"
+type impargs_state = implicits_list list GlobRef.Map.t
 
-let implicits_of_global ref =
+let _, implicits_tag = Summary.ref_tag GlobRef.Map.empty ~name:"implicits"
+
+let project_impargs st = Summary.project_from_summary (States.summary_of_state st) implicits_tag
+
+let implicits_of_global state ref =
   try
-    let l = GlobRef.Map.find ref !implicits_table in
+    let l = GlobRef.Map.find ref state in
     try
       let rename_l = Arguments_renaming.arguments_names ref in
       let rec rename implicits names = match implicits, names with
@@ -527,7 +531,7 @@ let implicits_of_global ref =
   with Not_found -> [DefaultImpArgs,[]]
 
 let cache_implicits_decl (ref,imps) =
-  implicits_table := GlobRef.Map.add ref imps !implicits_table
+  States.modify_state implicits_tag (fun s -> GlobRef.Map.add ref imps s)
 
 let load_implicits _ (_,(_,l)) = List.iter cache_implicits_decl l
 
--- a/library/states.ml
+++ b/library/states.ml
@@ -30,6 +30,12 @@ let intern_state s =
   unfreeze (with_magic_number_check (System.intern_state Coq_config.state_magic_number) s);
   Library.overwrite_library_filenames s
 
+let get_state () = freeze ~marshallable:`No
+
+let modify_state tag f =
+  let (fl,fs) = freeze ~marshallable:`No in
+  unfreeze (fl,Summary.modify_summary fs tag (f (Summary.project_from_summary fs tag)))
+
 (* Rollback. *)
 
 let with_state_protection f x =
--- a/dev/top_printers.ml
+++ b/dev/top_printers.ml
@@ -61,21 +61,26 @@ let prrecarg = function
 let ppwf_paths x = pp (Rtree.pp_tree prrecarg x)
 
 (* term printers *)
-let envpp pp = let sigma,env = Pfedit.get_current_context () in pp env sigma
+let envpp pp =
+  let sigma,env = Pfedit.get_current_context () in
+  let state = States.get_state () in
+  pp state env sigma
 let rawdebug = ref false
 let ppevar evk = pp (Evar.print evk)
 let pr_constr t =
   let sigma, env = Pfedit.get_current_context () in
-  Printer.pr_constr_env env sigma t
+  let state = States.get_state () in
+  Printer.pr_constr_env state env sigma t
 let pr_econstr t =
   let sigma, env = Pfedit.get_current_context () in
-  Printer.pr_econstr_env env sigma t
+  let state = States.get_state () in
+  Printer.pr_econstr_env state env sigma t
 let ppconstr x = pp (pr_constr x)
 let ppeconstr x = pp (pr_econstr x)
 let ppconstr_expr x = pp (Ppconstr.pr_constr_expr x)
 let ppsconstr x = ppconstr (Mod_subst.force_constr x)
 let ppconstr_univ x = Constrextern.with_universes ppconstr x
-let ppglob_constr = (fun x -> pp(pr_lglob_constr_env (Global.env()) x))
+let ppglob_constr = (fun x -> pp(pr_lglob_constr_env (States.get_state()) (Global.env()) x))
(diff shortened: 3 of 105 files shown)


ejgallego:
I am not sure what we gain here, in fact it may be even counterproductive to call `freeze` so often, because as for the rest `get_state` will make no difference.

IMHO the state has to be first threaded in the upper layers [STM], then it can be properly propagated downwards and clients can be aware.

ejgallego:
What I'd suggest is that you don't pass the `State.state`, but isolate the needed bits on a `printing_state` record.

Then, start propagating the state from `vernacentries`, to the commands that need such state, add it as a parameter; there, you have the exact point where the imperative state is first available as a value. You can in fact do this progressively and in an less invasive way.

In fact note that you call `get_state` in `Vernacentries` when you already have the state just in one level up in the call stack.

herbelin:
> What I'd suggest is that you don't pass the `State.state`, but isolate the needed bits on a `printing_state` record.

Yes, we can do that, but, at first view, this looks already like a lot of tables: the various notation tables (notations, delimiters, scopes, argument scopes, custom entries), the various name tables, the various record/structure tables (coercions, primitive projections), general options, plus morally the renaming table, the implicit types table, the generalized ident table.

Anyway, virtually any command shall need to print something, at least to report about errors in using the command. So, virtually any command shall need at least a `printing_env`. So, what I did can be reused: I called the state `States.state` but it can easily be changed into another type with a `sed`.

> Then, start propagating the state from `vernacentries`, to the commands that need such state, add it as a parameter; there, you have the exact point where the imperative state is first available as a value. You can in fact do this progressively and in an less invasive way.

We have a huge dag of functions coming from `Vernacentries` and what I did seems entirely complementary to what you are proposing.  We have to join the top (`Vernacentries`) to the leaves (e.g. `declare_implicits`) and we can go stepwise in either order I think. The only price is to be able to locally switch back and forth the functional view and the imperative view of the states until the whole migration is completed.

What I did is also for tactics. There are a lot of calls to the printer which come from tactics and tactics will carry their own state, so this is somehow something to do anyway which is independent from the graph of dependency in the state coming from `Vernacentries`.

I think that what is important is that we can collaborate on complementary tasks. For instance, #6171 is hard. But, if we arrive, say, to the level of having all `vernac/com*.ml` files being functional, it seems that what shall remain before fully threading the state in the `stm` will be easier.

> In fact note that you call get_state in Vernacentries when you already have the state just in one level up in the call stack.

Ah, indeed, I did not see it. Shall do.

By the way, what kind of naming scheme can we use to select sub-states? For instance, how would you call a function which goes from `States.state` to `printing_state`? Also, would we have all functions giving implicit arguments necessarily transiting via `printing_state` or could we have a more complex dag of coercions from the full global state to the individual states (using e.g. mixins?)

ejgallego:
> Anyway, virtually any command shall need to print something, at least to report about errors in using the command.

This is the key point, printing in Coq is a very complicated operation so indeed few commands should print stuff. In fact, today commands don't print to report errors, but raise an exception and the printing is done somewhere else.

ejgallego:
> what kind of naming scheme can we use to select sub-states?

You can already project from the summary, see the `project_from_summary` fct.

ejgallego:
Anyways I think printing is [for the reasons you mention] the first thing we should do. We should indeed first thread the `env`, that will prepare the code to indeed have an "extended env" that could carry for example printing information.

ejgallego (commented):
After a closer look at the patch, this is a welcome step; however I think there is a blocker for merging as-is: `get_state` in so many places. We should at least try to thread the state from the current owner [vernacentries] to the functions needing it so we can indeed figure out why functions do need to print stuff [they should not, except for the `Print` family of commands].

A few `get_state` could remain as a transition, but certainly not the number we have now.

library/states.ml
```diff
@@ -30,6 +30,12 @@ let intern_state s =
   unfreeze (with_magic_number_check (System.intern_state Coq_config.state_magic_number) s);
   Library.overwrite_library_filenames s
 
+let get_state () = freeze ~marshallable:`No
```
ejgallego: This is the same as `freeze`, so I don't see the point on introducing it.

interp/impargs.ml
```diff
@@ -527,7 +531,7 @@ let implicits_of_global ref =
   with Not_found -> [DefaultImpArgs,[]]
 
 let cache_implicits_decl (ref,imps) =
-  implicits_table := GlobRef.Map.add ref imps !implicits_table
+  States.modify_state implicits_tag (fun s -> GlobRef.Map.add ref imps s)
```
ejgallego: That certainly looks strange [and likely wrong], as this module is not the owner of the state.

What is the point of doing that indeed?

herbelin: This is correct afaij but a convoluted (and terribly inefficient) way to do the same as the previous line, since, instead of modifying the `implicits_table` reference directly, we freeze the whole state (which includes reading the value `implicits_table`), we modify the `implicits_table` field of this whole state, and we unfreeze, i.e. assign the new value back to the reference `implicits_table`.

There is no point to do that at the current time. It was an experiment to test how much the individual states could be seen as fields of a global state (rather than the global state as a collection of the freeze of the private states).

Note that in #8600, it is not anymore like this.

ejgallego: Indeed I see no point on doing that until you return the modified state to the upper layers. Owner is vernacentries so it must reach that part I think.

maximedenes (changes requested):
I agree with @ejgallego that functionalization should be done top-down. I don't know if this means delaying the PR or closing it, but it cannot be merged until the state can be threaded to these components IMHO.

herbelin:
> I agree with @ejgallego that functionalization should be done top-down. I don't know if this means delaying the PR or closing it, but it cannot be merged until the state can be threaded to these components IMHO.

I see this PR more as an experience and a way to trigger discussions than as _the_ solution to the desynchronization between error raising time and error printing time.

About functionalization going top-down or bottom-up, I don't think that it matters much, but I'm ok to go top-down. As far as I can judge, what is currently blocking the functionalization is the question of having a single `state` or a collection of specialized `state`s such as printing state, pretyping state, etc. The former is mechanical and thus easy to do. The latter requires thinking function per function and can thus take an arbitrary long time. If ever you have some ideas on how to organize ourselves to make progresses on this question, don't hesitate.

SkySkimmer:
Further discussion in the CEP

PR #8599, opened 2018-09-30
Typo in top_printers.ml

herbelin:
**Kind:** straightforward internal fix


ppedrot (approved):
Trivial, let's merge quickly. I don't think backporting is necessary, this is just for debug and unlikely to be useful to any user of the release.

PR #8600, opened 2018-09-30
Experimenting a functional variant of add_anonymous_leaf

herbelin:
**Kind:** experiment

[skip ci]

This is an experiment in providing a functional version of `add_anonymous_leaf`.

The idea is to provide three new methods for functional modification of the state (sps = state-passing-style):
```ocaml
val cache_sps_function : object_name * 'a -> Summary.frozen -> Summary.frozen
val load_sps_function : int -> object_name * 'a -> Summary.frozen -> Summary.frozen
val open_sps_function : int -> object_name * 'a -> Summary.frozen -> Summary.frozen
```
Then, this allows to provide a state-free function:
```ocaml
val add_anonymous_sps_leaf : Libobject.obj -> state -> state
```
and, consequently, functional variants of the various functions to add objects to the environment.

As an application, the table of implicit arguments is made fully functional, in both input and output.

At the current time, this is dramatically slow because of intensive useless freezing/unfreezing of the state (so as to synchronize the functional part and the effectful part).

### DIFF
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -299,8 +299,8 @@ let add_cpatt_for_params ind l =
   if !Flags.in_debugger then l else
     Util.List.addn  (Inductiveops.inductive_nparamdecls ind) (DAst.make @@ PatVar Anonymous) l
 
-let drop_implicits_in_patt cst nb_expl args =
-  let impl_st = (implicits_of_global cst) in
+let drop_implicits_in_patt state cst nb_expl args =
+  let impl_st = implicits_of_global (project_impargs state) cst in
   let impl_data = extract_impargs_data impl_st in
   let rec impls_fit l = function
     |[],t -> Some (List.rev_append l t)
@@ -367,8 +367,8 @@ let mkPat ?loc qid l = CAst.make ?loc @@
   (* Normally irrelevant test with v8 syntax, but let's do it anyway *)
   if List.is_empty l then CPatAtom (Some qid) else CPatCstr (qid,None,l)
 
-let pattern_printable_in_both_syntax (ind,_ as c) =
-  let impl_st = extract_impargs_data (implicits_of_global (ConstructRef c)) in
+let pattern_printable_in_both_syntax state (ind,_ as c) =
+  let impl_st = extract_impargs_data (implicits_of_global (project_impargs state) (ConstructRef c)) in
   let nb_params = Inductiveops.inductive_nparams ind in
   List.exists (fun (_,impls) ->
     (List.length impls >= nb_params) &&
@@ -433,12 +433,12 @@ let rec extern_cases_pattern_in_scope state (custom,scopes as allscopes) vars pa
 		Not_found | No_match | Exit ->
                   let c = extern_reference Id.Set.empty (ConstructRef cstrsp) in
                   if !asymmetric_patterns then
-		    if pattern_printable_in_both_syntax cstrsp
+                    if pattern_printable_in_both_syntax state cstrsp
 		    then CPatCstr (c, None, args)
 		    else CPatCstr (c, Some (add_patt_for_params (fst cstrsp) args), [])
 		  else
 		    let full_args = add_patt_for_params (fst cstrsp) args in
-		    match drop_implicits_in_patt (ConstructRef cstrsp) 0 full_args with
+                    match drop_implicits_in_patt state (ConstructRef cstrsp) 0 full_args with
 		      | Some true_args -> CPatCstr (c, None, true_args)
 		      | None           -> CPatCstr (c, Some full_args, [])
           in
@@ -472,7 +472,7 @@ and apply_notation_to_pattern ?loc state gr ((subst,substlist),(nb_to_drop,more_
             let l2 = List.map (extern_cases_pattern_in_scope state allscopes vars) more_args in
             let l2' = if !asymmetric_patterns || not (List.is_empty ll) then l2
 	      else
-		match drop_implicits_in_patt gr nb_to_drop l2 with
+                match drop_implicits_in_patt state gr nb_to_drop l2 with
 		  |Some true_args -> true_args
 		  |None -> raise No_match
 	    in
@@ -489,7 +489,7 @@ and apply_notation_to_pattern ?loc state gr ((subst,substlist),(nb_to_drop,more_
       let l2 = List.map (extern_cases_pattern_in_scope state allscopes vars) more_args in
       let l2' = if !asymmetric_patterns then l2
 	else
-	  match drop_implicits_in_patt gr (nb_to_drop + List.length l1) l2 with
+          match drop_implicits_in_patt state gr (nb_to_drop + List.length l1) l2 with
 	    |Some true_args -> true_args
 	    |None -> raise No_match
       in
@@ -537,7 +537,7 @@ let extern_ind_pattern_in_scope state (custom,scopes as allscopes) vars ind args
     with No_match ->
       let c = extern_reference vars (IndRef ind) in
       let args = List.map (extern_cases_pattern_in_scope state allscopes vars) args in
-      match drop_implicits_in_patt (IndRef ind) 0 args with
+      match drop_implicits_in_patt state (IndRef ind) 0 args with
 	   |Some true_args -> CAst.make @@ CPatCstr (c, None, true_args)
 	   |None           -> CAst.make @@ CPatCstr (c, Some args, [])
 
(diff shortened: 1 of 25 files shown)


ejgallego:
I think that what you want is precisely get things out `Summaries.frozen`, not make it pervasive all over the codebase.

herbelin:
> I think that what you want is precisely get things out Summaries.frozen, not make it pervasive all over the codebase.

Maybe we can get rid of `Summary.frozen` eventually, but I'm trying at the current time to find ways to stepwise move to a purely functional implementation. By providing a functional variant of `add_anonymous_leaf`, we can move the `object` tables progressively, until we are able, say, to get a purely functional variant of `do_definition`, `vernac_declare_implicits`, etc., etc.

In the process we can also try to make functions more fine-grained, e.g. `do_definition` manipulating a sub-state composed only of the global env and of the elaboration-level data, or `vernac_declare_implicits` manipulating only the sub-state of elaboration-level data, etc.

In the process, we can also decouple the modifications made to the `lib_state` and to the private (functional) states, so that a summary of all individual states is not needed anymore.

What would worry me is that we fail to make progress because of the size of the changes to do. By allowing a stepwise move to a purely functional treatment of the global tables, we can work in parallel and at times which suit us.

ejgallego:
> What would worry me is that we fail to make progress because of the size of the changes to do. By allowing a stepwise move to a purely functional treatment of the global tables, we can work in parallel and at times which suit us.

Indeed, but the question maybe is whether you want to proceed bottom-up or top-down. I do think top-down makes sense because in the end the top level layer is the one that has the control over the state, you that way you never risk being in the wrong. This is not so much about what parts you put but about who owns the state IMHO.

Whenever I've tried to go bottom-up I've found significant difficulties. Thus for example, the way I'd go about functions that modify the libstack is indeed to identify the vernaculars that can access such state and pass them the proper handle.

herbelin:
An alternative to what I did here is, in the transitory period, to distinguish two kinds of states, and have `val add_anonymous_sps_leaf : Libobject.obj -> functional_state -> functional_state`. In the case of implicit arguments, it would not be too costly to do the same as what you did in #6171 and to propagate the dependency of the dozen of callers to `declare_implicits` into `functional_state` up to `Vernacentries`. The latter would then be like in #6171 but of type `functional_state -> functional_state` rather than `Proof_global.t option -> Proof_global.t option`.

A current limitation is `VERNAC EXTEND` which is not able to take and return a state yet.

ejgallego:
> A current limitation is `VERNAC EXTEND` which is not able to take and return a state yet.

It used to be able to do so, with the `VERNAC FUNCTIONAL EXTEND` variant. However we had no users yet, but the infrastructure should be ready.

herbelin:
Trivial question but which matters in practice... what convention to use for a function moved in state-passing-style... 

- We should choose a name for the type of states which is short (`state`?)
- We should adopt a convention in the position of the `state` argument as input. What about before the non first optional argument?
- We should adopt  a convention in the position of the `state` argument in output. What about on the left (like for `evar_map`), and, if the function returns a pair, we don't return a triple but a pair of a state and of the pair?
- If the function already takes `env` (and possibly `sigma`) we give the state just before the `env` so as to have a regular subpattern `state -> env -> evar_map ->`?
- What name for the argument? `state` also, or `st` to be shorter (but more cryptic)?

For instance, if I have:
```ocaml
val f : ~a:int -> string -> ~b:bool -> int -> string * int
let f ~a x ~b y = if b then (x,a) else (x,y)
```
it becomes:
```ocaml
val f : ~a:int -> state -> string -> ~b:bool -> int -> state * (string * int)
let f ~a state x ~b y = (state, if b then (x,a) else (x,y))
```
Any opinion? Should we instead prefer a combinator, like `'a with_state = state -> state * 'a`? Should we prefer instead to have functions which behave well with `List.fold_left` (but then, it would not necessarily work for other combinators anyway)? Etc.

ejgallego:
> * We should adopt a convention in the position of the `state` argument as input. What about before the non first optional argument?

I think this is a good use case to use named parameters, that is `f ~st`. In fact, I wouldn't mind we use also `~env` and `~evd` (or `~evm`) which I prefer by far to `~sigma`.

ppedrot:
@herbelin FTR, I'd rather have some opaque monadic API, this would precisely alleviate the fact we have to stick to some code discipline by mere typing.

herbelin:
Personally, I find `st` a bit too cryptic, and it is already used for "transparent_state", for "style" (`terminal.ml`), for "size" (?), for "status" (`unicode.ml`, `constrintern.ml`, ...), for "stack" (`cbv.ml`), ...

That's true that an opaque API would e.g. avoid forgetting to return the state (as it happened to me in some local experiments), but wouldn't an opaque API force to insert some heavy `enter`-style boilerplate everywhere? Well, you will say: yes, that's the price for a strong invariant, and I shall not have counterarguments.

Anyway, I would be ok with either a `~state` (but in which position?) and with an opaque API.

ejgallego:
> but in which position

position doesn't matter when you use named arguments.

> that's the price for a strong invariant

using a monad provides no extra invariant, other than some convenience [and some potentially very large costs too] IMHO that's 100% orthogonal to state handling.

herbelin:
> That's true that an opaque API would e.g. avoid forgetting to return the state (as it happened to me in some local experiments), but wouldn't an opaque API force to insert some heavy enter-style boilerplate everywhere? Well, you will say: yes, that's the price for a strong invariant, and I shall not have counterarguments.

Actually, there are cases where we only need to read the state (e.g. the printers), so one would need two flavours of the API, right?

Still, an advantage of an API is that it would make the question of the position of the argument moot. On the other side, merlin is able to tell when a variable is unused, so the risk of forgetting to return a modified state is weak.

With an API, how would we translate typical patterns like (in `declare.ml`):
```ocaml
Array.iter2_i declare_one_projection labs projs
```
One would need to provide an ad hoc combinator `State.iter2_i`, right? To be compared with using `List.fold_left2_i` if without API?

Back to using a label, how would we translate the `Array.iter2_i` with a label? One would have to mandatorily eta-expand, as in
```ocaml
List.fold_left2_i (fun state -> declare_one_projection ~state) labs projs
```
right? On the other side, if the state is not the first argument of `declare_one_projection`, we would have to eta-expand anyway, so eta-expansion is somehow unavoidable in general.

Monad combinators, we already have a couple of them in `clib/monad.mli`, so the extra cost is little. On the other side, it seems a bit overkill to return `unit with_state` to comply with the monadic framework while we could simply return, more naturally, `state`.

ejgallego:
> how would we translate the `Array.iter2_i` with a label? One would have to mandatorily eta-expand, as in

I guess we could provide `fold_with_state` etc... to be made aware of the labels.

IMHO discussion of the API feels quite premature in the sense we still have not properly understood even which parts of the state we want to propagate. I'd say that an API should be introduced once the state-handling itself starts to feel mature.

herbelin:
> position doesn't matter when you use named arguments.

So, you can see how unfamiliar with named arguments I am! But if it does not matter at application time, it still matters, visually speaking, at definition time, so one would still need a convention, so that we don't write sometimes `~state env` and other times `env ~state`.

ejgallego:
> So, you can see how unfamiliar with named arguments I am! But if it does not matter at application time, it still matters, visually speaking, at definition time, so one would still need a convention, so that we don't write sometimes `~state env` and other times `env ~state`.

Sure, I guess most people places the named arguments first, but indeed. Note that you have the same problem with records, ordering doesn't matter. That's a minor issue tho.

A useful pattern tho is when you want to partially apply but only the 3rd argument.

p.s: It is not 100% true that position doesn't matter, but for the purpose of this discussion it does.

herbelin:
> IMHO discussion of the API feels quite premature in the sense we still have not properly understood even which parts of the state we want to propagate.

The WE, I focussed on implicit arguments and threading of the state between `declare.ml` and `vernacentries.ml` as a case study. In `declare.ml`, we return at least the global env, argument scopes, the nametab, implicit arguments (at least currently). In `comInductive.ml`, we additionally return at least notations and coercions . In `record.ml`, we additionally return projections.

How to organize these pieces of state? My solution would be to have a single notion of state, or at least only few notions of substates. If we start to have some state belonging to several larger classes of states, I feel that we shall enter a hell of subtyping and union of non disjoint states.

You said that you would like to have more differentiated classes of states and that's also possible, but that does not seem to me to be a necessary precondition for threading the state. This can be done in parallel, precisely when the dependencies will be made explicit.

> I'd say that an API should be introduced once the state-handling itself starts to feel mature.

OK, we can postpone the question of API, but I still feel that we need to do experiments, so that a general pattern come out. I'm afraid that if we don't set up objectives, we shall not progress. It is more than 3 years already that @gares explained that we have to make the spaghetti functional to get a more robust modularity.

Or, maybe, at least one priority would be that the proof state include the (whole) global state and that any action done on the global state during a proof is done instead on the local copy of the global state, with the global state changed only when the proof is done.

SkySkimmer:
Discussion is in the CEP

PR #8601, opened 2018-09-30
Move bound universe names to abstract contexts

### LINKED ISSUE #8939: Section universe names lost in discharge
~~~coq
Set Universe Polymorphism.
Section SomeSec.
  Universe u.
  Definition insec@{} := Prop.
End SomeSec.
Print insec.
(* Polymorphic insec@{Var(0)} = Prop
     : Type
*)
~~~
This is because we only add a universe binder object when it's nonempty, even though discharge may make it nonempty.
#8601 will fix this.

ppedrot:
We remove the global registration mechanism from `UnivNames` and set instead the bound universe names directly into the abstract universe contexts.

This has various advantages, e.g. guaranteeing that the bound names are meaningful (length must coincide), that no constants are forgotten (which used to be the case for schemes) and is more consistent with the handling of bound universe names as proper bound variables as e.g. done in bound term variables. It shouldn't change the output of printing, except maybe for cases that were clearly bugs, i.e. that forgot to register bound names.

This required adding name information to the polymorphic universe entries of constants and inductive types, and extending the API to abstract contexts to also require a list of bound names. 

This PR highlights the fact that a few places are still suspicious, e.g. that section universes use unordered and unnamed universe contexts, which is definitely wrong. Fixing this is left to another PR though.

Fixes #8939.

### DIFF
--- a/kernel/univ.ml
+++ b/kernel/univ.ml
@@ -1150,8 +1150,9 @@ let make_inverse_instance_subst i =
 let make_abstract_instance (ctx, _) = 
   Array.mapi (fun i _l -> Level.var i) ctx
 
-let abstract_universes ctx =
+let abstract_universes nas ctx =
   let instance = UContext.instance ctx in
+  let () = assert (Int.equal (List.length nas) (Instance.length instance)) in
   let subst = make_instance_subst instance in
   let cstrs = subst_univs_level_constraints subst 
       (UContext.constraints ctx)
@@ -1159,8 +1160,8 @@ let abstract_universes ctx =
   let ctx = UContext.make (instance, cstrs) in
   instance, ctx
 
-let abstract_cumulativity_info (univs, variance) =
-  let subst, univs = abstract_universes univs in
+let abstract_cumulativity_info nas (univs, variance) =
+  let subst, univs = abstract_universes nas univs in
   subst, (univs, variance)
 
 let rec compact_univ s vars i u =
--- a/kernel/univ.ml
+++ b/kernel/univ.ml
@@ -937,29 +937,29 @@ let hcons_universe_context = UContext.hcons
 
 module AUContext =
 struct
-  type t = Names.Name.t list constrained
+  type t = Names.Name.t array constrained
 
   let repr (inst, cst) =
-    (Array.init (List.length inst) (fun i -> Level.var i), cst)
+    (Array.init (Array.length inst) (fun i -> Level.var i), cst)
 
   let pr f ?variance ctx = UContext.pr f ?variance (repr ctx)
 
   let instantiate inst (u, cst) =
-    assert (List.length u = Array.length inst);
+    assert (Array.length u = Array.length inst);
     subst_instance_constraints inst cst
 
   let names (nas, _) = nas
 
   let hcons (univs, cst) =
-    (List.map Names.Name.hcons univs, hcons_constraints cst)
+    (Array.map Names.Name.hcons univs, hcons_constraints cst)
 
-  let empty = ([], Constraint.empty)
+  let empty = ([||], Constraint.empty)
 
-  let is_empty (nas, cst) = List.is_empty nas && Constraint.is_empty cst
+  let is_empty (nas, cst) = Array.is_empty nas && Constraint.is_empty cst
 
-  let union (nas, cst) (nas', cst') = (nas @ nas', Constraint.union cst cst')
+  let union (nas, cst) (nas', cst') = (Array.append nas nas', Constraint.union cst cst')
 
-  let size (nas, _) = List.length nas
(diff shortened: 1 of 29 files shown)


kernel/univ.ml
```diff
@@ -942,17 +942,30 @@ let hcons_universe_context = UContext.hcons
...
 
   let instantiate inst (u, cst) =
-    assert (Array.length u = Array.length inst);
+    assert (List.length u = Array.length inst);
     subst_instance_constraints inst cst
 
+  let names (nas, _) = nas
+
+  let hcons (univs, cst) =
+    (List.map Names.Name.hcons univs, hcons_constraints cst)
```
SkySkimmer: Shouldn't this do something on the list as a whole too? Although the way we use hcons is so adhoc it's hard to know anything.

library/lib.ml
```diff
@@ -483,7 +483,22 @@ let add_section_replacement f g poly hyps =
...
+      | Some na ->
+        try
+          let qid = Nametab.shortest_qualid_of_universe na in
+          Name (Libnames.qualid_basename qid)
+        with Not_found ->
+          (** Best-effort naming from the string representation of the level.
+              See univNames.ml for a similar hack. *)
+          Name (Id.of_string_soft (Univ.Level.to_string lvl))
+    in
+    let nas = Array.map_to_list map (Univ.Instance.to_array inst) in
```
SkySkimmer: Maybe move this out as its own function `instance -> name list`. It would avoid the name hacking distracting from the logic of `add_section_replacement`.

SkySkimmer (approved):
Any specific reason to use a list over an array?

library/lib.ml
ppedrot: I was thinking about this, but I don't know where such function should belong. It is really a user-facing feature so it would naturally be somewhere up, but the current wart forces it to be placed quite low. For the time being it might be better to leave it here, I will fix this code soon enough.

kernel/univ.ml
ppedrot: Indeed, but that would require writing yet another table. Note that this is not the only case where we don't properly hashcons intermediate structures for the sake of simplicity.

library/lib.ml
SkySkimmer: I was thinking just 
~~~ocaml
let name_instance = ...
let add_section_replacement = ...
~~~

ppedrot: Ah, OK, indeed. Let's do that.

kernel/univ.ml
SkySkimmer: There's some functor magic used by Universe, so it would just be ~2 lines.
Maybe we shouldn't hcons at all this part.

mattam82 (approved):
LGTM. I suppose you prefer a list of names for easier extension at e.g section closing time? Indeed section-polymorphic universes could be turned into arrays/lists right away, I guess at the beginning I was hoping to be able to shrink the contexts but that’s not safe. What happens to unnamed universes in the end, with Id.of_string_soft you get qualified numbered universe names like foo.12 ?

ppedrot:
@mattam82 I don't think it is very important to use lists for section extension, this is done only once and sharing is not of utmost importance in this case. Compactness of the representation might matter most. So, all in all it is just a matter of taste, I'll push a patch atop of this PR to use arrays instead if people prefer so.

Unnamed universes are given their own name through the `Id.of_string_soft` hack indeed, but this should be removed in favour of a fresh name printing in the instance printers. At least the current code prevents one to observe easily the internal global names of the universes generated during the proof.

ppedrot:
@mattam82 I need you help porting equations. There is only one use of the bad function it seems, and the root is `ucontext_of_aucontext` which as the names indicates breaks the abstraction barrier. The way this function is used is completely incomprehensible to me, so...

SkySkimmer:
@mattam82 how is it going?

ppedrot:
@mattam82 ping.

mattam82:
@SkySkimmer @ppedrot I'll give it a try this afternoon

mattam82:
Ok. I managed to fix things and get cleaner code. It's not simple though, I had to extend the hook of program to return information on the obligations (by which term they are defined) in PR #8889.
Using that, Equations can be cleaned-up (mattam82/Coq-Equations#147)

SkySkimmer:
@ppedrot could you add a test for #8939?

ppedrot:
Test added.

SkySkimmer:
Looks like quickchick is broken. The error doesn't look related but I'd like to be a bit more sure about it.

Meanwhile @herbelin @silene do you want to comment on the PR before I merge?

SkySkimmer:
I'm not waiting for appveyor.

PR #8602, opened 2018-09-30
[test-suite] Use ocamlfind to locate Coq libraries in unit tests.

ejgallego:
This is slightly more robust and allows to run the test suite with
Dune which may place OCaml objects differently.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -93,6 +93,7 @@ stdenv.mkDerivation rec {
   preInstallCheck = ''
     patchShebangs tools/
     patchShebangs test-suite/
+    export OCAMLPATH=$OCAMLFIND_DESTDIR:$OCAMLPATH
   '';
 
   installCheckTarget = [ "check" ];
--- a/default.nix
+++ b/default.nix
@@ -75,12 +75,18 @@ stdenv.mkDerivation rec {
         (path: _:
            !elem (baseNameOf path) [".git" "result" "bin" "_build" "_build_ci"]) ./.;
 
+  preConfigure = "patchShebangs kernel/";
+
   prefixKey = "-prefix ";
 
   buildFlags = [ "world" "byte" ] ++ optional buildDoc "doc-html";
 
   installTargets =
     [ "install" "install-byte" ] ++ optional buildDoc "install-doc-html";
+
+  createFindlibDestdir = !shell;
+
+  postInstall = "ln -s $out/lib/coq $OCAMLFIND_DESTDIR/coq";
 
   inherit doInstallCheck;
 
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -250,11 +250,11 @@ $(addsuffix .log,$(wildcard bugs/closed/*.v)): %.v.log: %.v
 # Unit tests
 #######################################################################
 
-OCAMLOPT := $(OCAMLFIND) opt $(CAMLFLAGS)
-SYSMOD:=-package num,str,unix,dynlink,threads
+ifeq ($(LOCAL),true)
+  export OCAMLPATH := $(patsubst "%",%,$(COQLIBINSTALL))
+endif
 
-COQSRCDIRS:=$(addprefix -I $(LIB)/,$(CORESRCDIRS))
-COQCMXS:=$(addprefix $(LIB)/,$(LINKCMX))
+OCAMLOPT := $(OCAMLFIND) opt $(CAMLFLAGS)
 
 # ML files from unit-test framework, not containing tests
 UNIT_SRCFILES:=$(shell find ./unit-tests/src -name *.ml)
@@ -278,10 +278,8 @@ unit-tests: $(UNIT_LOGFILES)
 # Build executable, run it to generate log file
 unit-tests/%.ml.log: unit-tests/%.ml
 	$(SHOW) 'TEST      $<'
-	$(HIDE)$(OCAMLOPT) -linkall -linkpkg -cclib -lcoqrun \
-		   $(SYSMOD) -package camlp5.gramlib,oUnit  \
-	     -I unit-tests/src $(COQSRCDIRS) $(COQCMXS) \
-	     $(UNIT_CMXS) $< -o $<.test;
+	$(HIDE)$(OCAMLOPT) -linkall -linkpkg -package coq.toplevel,oUnit \
+	     -I unit-tests/src $(UNIT_CMXS) $< -o $<.test;
 	$(HIDE)./$<.test
 
 #######################################################################
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -250,8 +250,11 @@ $(addsuffix .log,$(wildcard bugs/closed/*.v)): %.v.log: %.v
 # Unit tests
 #######################################################################
 
+# COQLIBINSTALL is quoted in config/make thus we must unquote it,
+# otherwise the directory name will include the quotes, see
+# see for example https://stackoverflow.com/questions/10424645/how-to-convert-a-quoted-string-to-a-normal-one-in-makefile
 ifeq ($(LOCAL),true)
-  export OCAMLPATH := $(patsubst "%",%,$(COQLIBINSTALL))
+  export OCAMLPATH := $(shell echo $(COQLIBINSTALL))
 endif
 
 OCAMLOPT := $(OCAMLFIND) opt $(CAMLFLAGS)
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -130,6 +130,7 @@ after_script:
     # careful with the ending /
     - BIN=$(readlink -f ../_install_ci/bin)/
     - LIB=$(readlink -f ../_install_ci/lib/coq)/
+    - export OCAMLPATH=$(readlink -f ../_install_ci/lib/):"$OCAMLPATH"
     - make -j "$NJOBS" BIN="$BIN" LIB="$LIB" all
   artifacts:
     name: "$CI_JOB_NAME.logs"


ejgallego:
> CI fails.

Yup, I need to figure out how to handle `-local` properly sight.

gares:
I think the `directory` field of the `META.coq` file is wrong (here, using -loca) since it is `""` while it should probably be `$PWD`

ejgallego:
@Zimmi48 another Coqbot spurious `needs: rebase`. Definitively it happens.

test-suite/Makefile
```diff
@@ -250,11 +250,11 @@ $(addsuffix .log,$(wildcard bugs/closed/*.v)): %.v.log: %.v
 # Unit tests
 #######################################################################
 
-OCAMLOPT := $(OCAMLFIND) opt $(CAMLFLAGS)
-SYSMOD:=-package num,str,unix,dynlink,threads
+ifeq ($(LOCAL),true)
+  export OCAMLPATH := $(patsubst "%",%,$(COQLIBINSTALL))
+endif
```
SkySkimmer: Why isn't that test around setting `LIB` and then an unconditional `export OCAMLPATH := ...`?

test-suite/Makefile
```diff
@@ -250,11 +250,11 @@ $(addsuffix .log,$(wildcard bugs/closed/*.v)): %.v.log: %.v
 # Unit tests
 #######################################################################
 
-OCAMLOPT := $(OCAMLFIND) opt $(CAMLFLAGS)
-SYSMOD:=-package num,str,unix,dynlink,threads
+ifeq ($(LOCAL),true)
+  export OCAMLPATH := $(patsubst "%",%,$(COQLIBINSTALL))
```
SkySkimmer: Please add a comment about what this patsubst is doing.

ejgallego: Is this really the correct thing to do?

Zimmi48:
It is ran after installation. But as soon as the test-suite and the documentation build are supported by Dune, I want to switch to using the Dune target (note though that the documentation is more than just Sphinx in principle).

ejgallego:
> It is ran after installation.

So does `nix` then don't properly update `OCAMLPATH` for Coq? Is that a bug in the nix Coq package then?

ejgallego:
> note though that the documentation is more than just Sphinx in principle

I see the refman and the coqdoc for the stdlib, anything else?

vbgl:
> So does `nix` then don't properly update `OCAMLPATH` for Coq?

What do you mean? What would be “proper”?

The packages are such that things like that work as expected: `nix-shell -p coq coq.camlp5 coq.findlib --pure --run 'ocamlfind query -r coq'`

ejgallego:
> What do you mean? What would be “proper”?

It seems that after the install step in the CI file, `ocamlfind query coq.toplevel` doesn't succeed. See the nix log.

ejgallego:
> Where do you see the bug?

Well, if the `check` test is supposed to run after Coq is installed, then the installed OCaml libraries should be properly registered.

So the bug [or weakness of the model] is that indeed the environment should be updated after install.

For example Dune, which likes to extend `PATH` and `OCAMLPATH` etc... in a similar fashion has to be very careful as to properly run the actions in the right environment.

vbgl:
I suggest to write two derivations, the second one to test the first one.

ejgallego:
> I suggest to write two derivations, the second one to test the first one.

That looks messy. Moreover the `nix` documentation [says](https://nixos.org/nixpkgs/manual/#ssec-installCheck-phase):
> The installCheck phase checks whether the package was installed correctly by running its test suite against the installed directories. The default installCheck calls make installcheck.

If so, sorry, this is a bug [I am not sure where]. If `ocamlfind query coq` doesn't work then the package was not properly installed.

ejgallego:
Well, it seems that Nix doesn't even update `PATH` after the build, so I am not sure how this was working before, but likely by doing the test using the local files, not the installed ones. Indeed, would we do `make clean` before the `check` I am not sure that would still work.

ejgallego:
Well, I tried to do `clean` before the `check` target to see if the thing fails, see 
https://gitlab.com/ejgallego/coq/-/jobs/104528914 but we get into a make-hell it seems ...

But I doubt it is using the installed Coq to run the test-suite indeed.

So I am not sure how to interpret the semantics of the Nix checkInstall target then. I dunno if the target should be fixed as to properly set the env vars, do 2 derivations, if you folks things the current semantics are OK?

vbgl:
My personal opinion is that the `installCheckPhase`, as documented, is meaningless.

If you want to test the thing you install, then do not attempt to replicate an environment in which your package is installed by modifying the environment in which the package is built (which may have nothing in common…).

ejgallego:
> If you want to test the thing you install, then do not attempt to replicate an environment in which your package is installed by modifying the environment in which the package is built (which may have nothing in common…).

Interesting, thanks for feedback. So how should we proceed here?

vbgl:
Let me give it a try.

Zimmi48:
Just to be clear. I used put the check in `installCheckPhase` only because it would fail in `checkPhase`. I believe that the build model of Dune will solve this and therefore at this point, I would happily put this back in `checkPhase`.

default.nix
```diff
@@ -75,13 +75,21 @@ stdenv.mkDerivation rec {
         (path: _:
            !elem (baseNameOf path) [".git" "result" "bin" "_build" "_build_ci"]) ./.;
 
+  enableParallelBuilding = true;
```
Zimmi48: The reason not to put this was because `world` and `byte` are not supposed to be run together in parallel. It would probably be more reasonable to wait to use Dune to build before enabling parallel building.

vbgl:
I just changed the `default.nix` so as to extend the `OCAMLPATH` environment variable before the tests are run. It seems to work on CI.

default.nix
```diff
@@ -75,18 +75,25 @@ stdenv.mkDerivation rec {
         (path: _:
            !elem (baseNameOf path) [".git" "result" "bin" "_build" "_build_ci"]) ./.;
 
+  preConfigure = "patchShebangs kernel/";
```
Zimmi48: This part should probably be backported

ejgallego:
Thanks a lot @vbgl , I need to think a bit more here.

default.nix
```diff
@@ -75,18 +75,25 @@ stdenv.mkDerivation rec {
...
+
   prefixKey = "-prefix ";
 
   buildFlags = [ "world" "byte" ] ++ optional buildDoc "doc-html";
 
   installTargets =
     [ "install" "install-byte" ] ++ optional buildDoc "install-doc-html";
 
+  createFindlibDestdir = !shell;
```
Zimmi48: What would this do if you simply put `true` instead?

Zimmi48 (approved):
@vbgl I see that this is a common pattern in nixpkgs, but what are the advantages of creating the folder and linking `lib/coq` in there compared to simply adding `lib/coq` to `OCAMLPATH` (like we do e.g. with `COQPATH`?)

default.nix
vbgl: Running `nix-shell` would then fail because it cannot create a directory in your store. (I did try and got bitten.)

ejgallego:
So @Zimmi48 @vbgl , what do you think then? Do you consider the changes acceptable? Should we proceed with this?

I hope we can clean this up more in the future; now it feels a bit messy, but when using Dune all the env is correctly set up so you could remove this kludge.

vbgl:
LGTM.

ejgallego:
Cool, thanks!

default.nix
Zimmi48: Isn't it a bug of nixpkgs' OCaml package or something like this?

Zimmi48:
Yep, LGTM as well.

ejgallego:
Great, so this needs an assignee I think.

test-suite/Makefile
gares: I don't understand it either, please explain what it does.

ejgallego:
Maybe @gares , or @jfehrle ? Who's in charge of `unit-tests`?

gares:
If you want me to be the assignee and merge the PR you should also answer the question I asked. Since I don't know what you do there.

ejgallego:
> If you want me to be the assignee and merge the PR you should also answer the question I asked. Since I don't know what you do there.

Sorry @gares I missed it with the discussion about Nix, you mean this comment right?

> I think the `directory` field of the `META.coq` file is wrong (here, using -loca) since it is `""` while it should probably be `$PWD`

I am not sure, I think the directory is relative to the location of `META` file itself.

It seems to work well here, just checked `./configure -profile devel && make test-suite`, have you found a problem?

gares:
Hum, I made two comments to which I did not have an answer.

The one about the `directory` being empty was a suggestion. The value is an absolute path, and IIRC, ocamlfind can fill it in when it is used to install a META, for example.
The seggestion was: in -local mode you could write in there the absolute path of the sources, and then by also setting OCAMLPATH you could help ocamlfind find that META and consequently find the .cm* fils using that absolute path.

The other one is about  `export OCAMLPATH := $(patsubst "%",%,$(COQLIBINSTALL))` that I don't get.  Gaetan is also puzzled... What does is do?

ejgallego:
> The one about the `directory` being empty was a suggestion. The value is an absolute path, and IIRC, ocamlfind can fill it in when it is used to install a META, for example.

The value is not absolute path, quoting the [ocamlfind reference manual](http://projects.camlcity.org/projects/dl/findlib-1.8.0/doc/ref-html/r759.html)
> The value of the "directory" variable is determined with an empty set of actual predicates. The value must be either: an absolute path name of the alternate directory, or a path name relative to the stdlib directory of OCaml (written "+path"), or a normal relative path name (without special syntax). 

I do think the current approach is simpler than the one you propose and has worked pretty well so far.

> The other one is about `export OCAMLPATH := $(patsubst "%",%,$(COQLIBINSTALL))` that I don't get. Gaetan is also puzzled... What does is do?

Oh, this is standard `make` fare to unquote a variable, `configure` writes `COQLIBINSTALL="dir"` to `config/Makefile`, however, `make` doesn't interpret quotes thus if you don't unquote you end with a quoted directory which is incorrect.

I am open to take suggestions from `make` experts, TTBOMK this is the simplest way to avoid this issue and don't fall into another problems.

SkySkimmer:
>Oh, this is standard make fare to unquote a variable, configure writes COQLIBINSTALL="dir" to config/Makefile, however, make doesn't interpret quotes thus if you don't unquote you end with a quoted directory which is incorrect.

Then please put a comment `unquote COQLIBINSTALL as make's export doesn't interpret quotes` or some such.

ejgallego:
> Then please put a comment `unquote COQLIBINSTALL as make's export doesn't interpret quotes` or some such.

OK. Is that OK with you too @gares ?

ejgallego:
> make's export doesn't interpret quotes

In general `make` doesn't interpret any kind of quoting I understand.

gares:
> The value is not absolute path, quoting ...
> > The value of the "directory" variable is determined with an empty set of actual predicates. The value must be either: an absolute path ...

I'm not so sure I follow you. The manual you quote says is can be an absolute path.

ejgallego:
> I'm not so sure I follow you. The manual you quote says is can be an absolute path.

What is the gain? It makes things much more complicated for no gain [you still have to extend `OCAMLPATH`]

gares:
I found the same stackoverflow post you found, but the solution suggested does not seem work..

```shell
gares@gargamel:/tmp$ cat Makefile 
A := "foo bar baz"
export B := $(shell echo $(A))
export C := $(patsubst "%",%,$(A))

all:
	echo $$B
	echo $$C

gares@gargamel:/tmp$ make
echo $B
foo bar baz
echo $C
"foo bar baz"
```

ejgallego:
> but the solution suggested does not seem work..

I dunno, it works here. [Remove it and you will witness the error]

ejgallego:
Well, try with `A:="foo_bar_baz"` , yes, when in spaces the thing is special. People wonder why we don't like make lol

ejgallego:
So basically I have no clue how such an operation should be performed in `make`.

SkySkimmer:
Maybe we should just not quote when outputting to make.
Maybe even error when there's a space in the path.

More practically instead of `if local then export ocamlpath = unquote bla` we could do `if local then ocamlfind = ocamlpath=bla ocamlfind` then the shell will unquote for us.

gares:
> What is the gain?

I'm fine if you leave things as they are if they work. I was just explaining my suggestion, and you argued it did not apply because the manual says it can't be an absolute path; and you quote a paragraph that says it can. Did you read the text you quoted?

> So basically I have no clue how such an operation should be performed in make.

You are right about the space. The solution with `echo` works in both cases, while the one with `patsubst` does not. I guess you should change your commit.

SkySkimmer:
also possible: export ocamlpath = $(shell echo bla)

Gaëtan Gilbert

ejgallego:
> I was just explaining my suggestion, and you argued it did not apply because the manual says it can't be an absolute path; and you quote a paragraph that says it can. Did you read the text you quoted?

You wrote "the value is an absolute path", whereas you meant "the value may be an absolute path" so hence I got confused; I have never said it cannot be an absolute path. I am OK changing to an absolute path if we see some advantage; but I find relative more robust here and we can just relocate the lib folder and it still works.

> You are right about the space. The solution with `echo` works in both cases, while the one with `patsubst` does not. I guess you should change your commit.

Ok I'll use `echo` if you prefer, I hope it doesn't introduce some other problems. With make you never know.

ejgallego:
PR updated, thanks for the help folks!

gares (approved):

jfehrle:
Since my name was mentioned:  Is this really specific to tests in `unit-tests`?  The existing tests there are calling specific functions (not very high level) in Coq.  None of them does anything nearly as complex as loading a library, though that may be a useful capability for the future.  Is there any plan to write such tests?

Will this require developers to use OPAM for libraries if they want to run unit tests locally?

ejgallego:
Grazie!

test-suite/Makefile
vbgl: This breaks local building of the test suite:

```
OCAMLOPT  unit-tests/src/utest.mli
ocamlfind: Package `oUnit' not found
```

ejgallego: I see what happened, with the latest change we are resetting the `OCAMLPATH` instead of appending to it, thus Nix breaks. Does:
```
ifeq ($(LOCAL),true)
  export OCAMLPATH := $(shell echo $(COQLIBINSTALL):$$OCAMLPATH)
endif
```
work? [The quoting stuff is crazy with `make`]

ejgallego: In fact I knew of this problem but somehow the proper rule got lost in the last commit.

vbgl: Works for me. Thanks.

ejgallego: Do you submit the PR? Otherwise I can do it.

ejgallego: Ok I got the PR ready so don't bother.

Issue #8610, opened 2018-10-01
Change important CI DOS batch files to CRLF


MSoegtropIMC:
Currently all Windows batch files used on CI use LF line endings. Unfortunately this can have the effect that jump/call labels are not found (but only if the jump label is more than 512 bytes away from the call and the moon is in a certain constellation to Saturn, otherwise it works fine - oh how we love MS). So I fear I have to change some important CI batch files to CRLF, although this makes diffing hard.

See https://web.archive.org/web/20160305063042/help.wugnet.com/windows/system-find-batch-label-ftopict615555.html

I actually run into this issue in CI job

https://gitlab.com/coq/coq/-/jobs/103241210

At the end it complains "The system cannot find the batch label specified - CopyLogFiles", but the label is definitely there.

MSoegtropIMC:
Fixed by #8612.

PR #8612, opened 2018-10-01
Fix issue 8610 - Change important CI DOS batch files to CRLF

MSoegtropIMC:
This fixed issue #8610 that LF line endings for DOS batch files can lead to unreliabilities in CI.


Zimmi48 (approved):
GitHub has a nice feature which allows me to confirm that this PR is only whitespace change.

PR #8615, opened 2018-10-01
[dune] Add Dune bootstrap target.

ejgallego:
This allows anyone to compile Coq easily without having to install
Dune.


ejgallego:
Once I fix the windows build, we will have no reason the keep the old make system in place so we can start the discussion about when to replace it.

Test suite and documentation do work well tho the PRs need updating.

Makefile.dune
```diff
@@ -1,22 +1,26 @@
 # -*- mode: makefile -*-
 # Dune Makefile for Coq
 
-.PHONY: help voboot states world apidoc
+.PHONY: help voboot states world watch release apidoc clean bootstrap bootstrap-build
```
Zimmi48: It seems that adding `watch` `release` and `clean` here is actually a bugfix. Can you push this to another PR?

Makefile.dune
```diff
@@ -37,6 +41,20 @@ release: voboot
...
 	dune build $(DUNEOPT) @doc
 
+DUNE_BOOT_VER=1.3.0
+DUNE_TBZ=dune-$(DUNE_BOOT_VER).tbz
+DUNE_BOOT_BIN=_dune/dune-$(DUNE_BOOT_VER)/_build/install/default/bin/dune
+bootstrap:
+	mkdir -p _dune
+	( cd _dune && wget https://github.com/ocaml/dune/releases/download/$(DUNE_BOOT_VER)/$(DUNE_TBZ) && tar xvfj $(DUNE_TBZ) )
+	( cd _dune/dune-$(DUNE_BOOT_VER)/ && make release )
+	@echo "A Local version of Dune is ready to be used, add _dune/dune-$(DUNE_TBZ)/_build/install/default/bin/ to your PATH"
```
Zimmi48: Typo: `DUNE_TBZ` should be `DUNE_BOOT_VER`.

Makefile.dune
```diff
@@ -37,6 +41,20 @@ release: voboot
 apidoc: voboot
 	dune build $(DUNEOPT) @doc
 
+DUNE_BOOT_VER=1.3.0
+DUNE_TBZ=dune-$(DUNE_BOOT_VER).tbz
+DUNE_BOOT_BIN=_dune/dune-$(DUNE_BOOT_VER)/_build/install/default/bin/dune
+bootstrap:
+	mkdir -p _dune
+	( cd _dune && wget https://github.com/ocaml/dune/releases/download/$(DUNE_BOOT_VER)/$(DUNE_TBZ) && tar xvfj $(DUNE_TBZ) )
```
Zimmi48: It is a bit weird to have a `wget` as part of a Makefile. Will it be OK with e.g. Debian policies? In any case, maybe this download step should be split up to a separate target so that package maintainers are allowed to do this step in their own pre-install script and not have the Makefile do this.

Zimmi48 (commented):
I'm OK with this but it makes sense only if the we agree to actually remove the Makefile-based system. I think it makes sense but we need to ensure first that all the usual targets work with Dune and all developers should switch to using Dune only for a few weeks first.

Makefile.dune
ejgallego: OK, will do.

Makefile.dune
ejgallego: Indeed, thanks.

Makefile.dune
ejgallego: It would be a violation of the Debian policy if we required to use the `wget` target to build the package, but indeed, in Debian the build would depend on Dune and use the system one so this is OK.

ejgallego:
> I'm OK with this but it makes sense only if the we agree to actually remove the Makefile-based system. 

Well, indeed having this in is a precondition to even start the discussion.

> I think it makes sense but we need to ensure first that all the usual targets work with Dune and all developers should switch to using Dune only for a few weeks first.

All the usual targets should be working shortly, the only blocker is Windows which is painful for me to debug; also I'd like to make some more progress on cleaning the location of libraries, `coqdep` etc...

I have a make-free already in my PC, and I will submit it as a PR to start the discussion as soon as windows is working.

SkySkimmer:
>Well, indeed having this in is a precondition to even start the discussion.

Why not just have people install dune?

ejgallego:
> Why not just have people install dune?

IMHO there are many users that install Coq "the old way", so I think we don't want to have them to install extra software when the solution is so easy. Note that long time ago, Coq used to ship with its own copy of `make`, so here we go, circle of reincarnation!

Makefile.dune
```diff
@@ -44,6 +48,22 @@ ocheck: voboot
 clean:
 	dune clean
 
+# Dune bootstrap targets:
+
+DUNE_BOOT_VER=1.3.0
```
gares: Not so sure I like hard coding the version here. I guess it may get out of sync with the one we use in CI for example.

Makefile.dune
```diff
@@ -44,6 +48,22 @@ ocheck: voboot
...
+DUNE_TBZ=dune-$(DUNE_BOOT_VER).tbz
+DUNE_BOOT_PATH=_dune/dune-$(DUNE_BOOT_VER)/_build/install/default/bin
+bootstrap:
+	mkdir -p _dune
```
gares: On OSX `curl` used to come by default, while `wget` did not. Not sure which is the best way to use the right one in this one liner.

gares:
Can you explain the purpose of this patch? Which is the use case?

FTR: Coq binary packages used to include make on windows, I have no memory of Coq embedding the sources of make.

Makefile.dune
ejgallego: How would you solve that problem? I am not sure we can do better as we are talking about a myriad of different devices envs.

Makefile.dune
ejgallego: Yeah I was doubting about that, I dunno, will take either of them.

ejgallego:
> Can you explain the purpose of this patch? Which is the use case?
> 
> FTR: Coq binary packages used to include make on windows, I have no memory of Coq embedding the sources of make.

Sorry I miswrote (I'm so tired I guess) what I meant is that used to provide instructions to compile make, see e16dafd4374edce11fe3b2a6778b5a058f44b87f

The purpose of this patch is exactly the same as of the above commit, provide a more streamlined install experience for users wanting to use Dune, also show that no additional dependencies are introduced.

Makefile.dune
gares: write it is a file that is "included" in all the others?

gares (changes requested):
I gave a look at the README in the tarball you download. It says:
```
Installation
------------

The recommended way to install dune is via the [opam package manager][opam]:

$ opam install dune
```
Sure, It also tells you how to compile it from sources but also tell you to run make install.

IMHO `Makefile.dune` should check if dune is available. If not point the user to installation instructions written and *maintained* by the dune developers. Not hack a local installation of a random version.

ejgallego:
> `Makefile.dune` should check if dune is available.
I am not sure I know how to do that. Personally I don't even use the makefile but just directly call Dune.

> If not point the user to installation instructions written and _maintained_ by the dune developers. Not hack a local installation of a random version.

The installation method this PR proposed is not a hack but supported by upstream; ditto about the version, there is nothing random about it, on the contrary it has deterministically been chosen to be 1.3.0 as of today.

gares:
> is not a hack but supported by upstream

So why they tell you do `make && make install` or preferably `opam install dune`? Is their README wrong?

I did say random, but what I really meant is arbitrary. As I point out in the comment you download version 1.3, but the rest of the file does not check that and woks with any dune that is in the path.

> I am not sure I know how to do that.

How do you check in your shell if dune is present? You run it! Come on...

Zimmi48:
I think it makes sense to just add Dune to the Coq dependencies and let people install it (they are forced, after all, to install camlp5 as well). Most people will install this through OPAM anyways. Others through other package managers (like Nix or many distribution package managers) that contain Dune. Given the opposition of @gares to this PR, I suggest to just close it. This shouldn't (IMO) be blocking to remove the make-based build system. Only the (potential) opposition of other developers could be.

ejgallego:
> This shouldn't (IMO) be blocking to remove the make-based build system. Only the (potential) opposition of other developers could be.

Well I think here we are underestimating how users are, IMHO requiring certain users to install Dune will be a lot of pain for them; they should be able to work like they do now.

Zimmi48:
>they should be able to work like they do now.

But how do they do now? They use OPAM or another package manager, don't they?

ejgallego:
> But how do they do now? They use OPAM or another package manager, don't they?

This is a good question, I guess that some of them are stuck in Coq 8.2 :)

I dunno, they may indeed download ocaml + camlp5 and compile it manually; old school is still there, but likely they are indeed moving to some form of package management.

ejgallego:
I do like "vendoring", but not by default of course. YMMV I guess.

Zimmi48:
>I dunno, they may indeed download ocaml + camlp5 and compile it manually; old school is still there, but likely they are indeed moving to some form of package management.

I know of a guy that still does it that way, but he was also the last one to use `coqtop` when everyone else had moved to Proof-General :wink:

ejgallego:
Well, going back on topic I don't this PR can harm anyone, and certainly I'd bet that this kind of kludge would appear in one form or another would we ever migrate to Dune.

That being said, feel free to close the PR @Zimmi48 if you deem it appropriate, that'll be easy to resurrect if ever needed.

Zimmi48:
Now one important question is whether Debian Buster will ship with OPAM 2. For now, it is still stuck at 1.2. Given that only the OPAM 2 package index is receiving updates, we won't be able to tell people to install a recent enough version of Dune with OPAM if they don't have OPAM 2...

gares:
The point is you don't want to maintain an installation script for dune in Coq.
We don't have such a thing for opam, for ocaml, for you name it.

ejgallego:
In #8421 you have an example of how useful it would have been to have this in the tree.

ejgallego:
What to do here? I keep collecting evidence that this is useful to people, specially in the testing stage.

So I propose to merge.

Zimmi48:
Let's discuss it at the next WG.

gares:
What was decided at the wg?

Zimmi48:
It was not discussed (when I said the next WG, I meant the next one not the one that was about to end -- Emilio had already left).

gares:
I don't know how to react to your message. How can you hope to convince someone this way?

ejgallego:
I am not hoping to convince anyone. It is a fact, whitout this or anything isomorphic in the CI scripts building on windows is just a pain for the users.

Have you seen the current windows build scripts? This pales in comparison.

Zimmi48:
The scripts already contains building logic for many dependencies of Coq. This PR could be integrated in these scripts. Or we could instead make this logic less specific so that it can be reused elsewhere. Both solutions are fine to me but we should try to aim for more consistency, not less.

ejgallego:
> The scripts already contains building logic for many dependencies of Coq. This PR could be integrated in these scripts. Or we could instead make this logic less specific so that it can be reused elsewhere. Both solutions are fine to me but we should try to aim for more consistency, not less.

I'd say that Dune is a bit special as it is the only "hard" dependency [along with OCaml] that we ship, and cannot be easily "vendored". [OCaml will be able to be vendored in the future tho !]

Regarding the rest of the dependencies, we should in fact remove all the build logic, and make them use a standard build system [Dune in this case]. This is of course a long-term plan but IMHO makes a lot of sense.

ejgallego:
Well another very useful PR for a particular demographic blocked for no good reason [that I can see]

No point on keeping this open, but for sure the need will come in the future.

PR #8616, opened 2018-10-01
Include the full Table of Contents document in the on-screen TOC, ...

jfehrle:
**Kind:** documentation

Made Introduction, Credits and License top-level items (TOC won't show subsections)
Fixed 2 incorrect option/table names in Flags, Options, Tables index.

I thought it more conventional to have the TOC first.  The indexes are next; this leaves them very readily accessible for readers (on-screen for any reasonable window size).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,7 +101,6 @@ doc/faq/axioms.pdf_t
 doc/faq/axioms.png
 doc/sphinx/index.rst
 doc/sphinx/zebibliography.rst
-doc/sphinx/credits.rst
 doc/stdlib/Library.out
 doc/stdlib/Library.ps
 doc/stdlib/Library.coqdoc.tex
--- a/Makefile.doc
+++ b/Makefile.doc
@@ -10,7 +10,7 @@
 
 # Makefile for the Coq documentation
 
-# Read INSTALL.doc to learn about the dependencies
+# Read doc/README.md to learn about the dependencies
 
 # The main entry point :
 
--- a/doc/sphinx/conf.py
+++ b/doc/sphinx/conf.py
@@ -193,8 +193,9 @@ html_theme = 'sphinx_rtd_theme'
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-#html_theme_options = {}
-
+html_theme_options = {
+    'collapse_navigation': False
+}
 html_context = {
     'display_github': True,
     'github_user': 'coq',
--- a/doc/sphinx/credits.html.rst
+++ b/doc/sphinx/credits.html.rst
@@ -1,7 +1,0 @@
-.. _credits:
-
--------
-Credits
--------
-
-.. include:: credits-contents.rst
--- a/doc/sphinx/credits.latex.rst
+++ b/doc/sphinx/credits.latex.rst
@@ -1,3 +1,0 @@
-.. _credits:
-
-.. include:: credits-contents.rst
--- a/doc/sphinx/credits.rst
+++ b/doc/sphinx/credits.rst
@@ -1,3 +1,7 @@
+-------
+Credits
+-------
+
 Coq is a proof assistant for higher-order logic, allowing the
 development of computer programs consistent with their formal
 specification. It is the result of about ten years of research of the
@@ -186,7 +190,7 @@ definitions of “inversion predicates”.
 |
 
 Credits: addendum for version 6.1
-=================================
+---------------------------------
 
 The present version 6.1 of |Coq| is based on the V5.10 architecture. It
 was ported to the new language Objective Caml by Bruno Barras. The
@@ -223,7 +227,7 @@ Barras.
 |
 
 Credits: addendum for version 6.2
-=================================
+---------------------------------
 
 In version 6.2 of |Coq|, the parsing is done using camlp4, a preprocessor
 and pretty-printer for CAML designed by Daniel de Rauglaudre at INRIA.
@@ -268,7 +272,7 @@ Loiseleur.
 |
 
 Credits: addendum for version 6.3
-=================================
+---------------------------------
 
 The main changes in version V6.3 were the introduction of a few new
 tactics and the extension of the guard condition for fixpoint
@@ -301,7 +305,7 @@ Monin from CNET Lannion.
 |
 
 Credits: versions 7
-===================
+-------------------
 
 The version V7 is a new implementation started in September 1999 by
 Jean-Christophe Filliâtre. This is a major revision with respect to the
@@ -390,7 +394,7 @@ J.-F. Monin from France Telecom R & D.
 |
 
 Credits: version 8.0
-====================
+--------------------
 
 Coq version 8 is a major revision of the |Coq| proof assistant. First, the
 underlying logic is slightly different. The so-called *impredicativity*
@@ -492,7 +496,7 @@ under the responsibility of Christine Paulin.
 |
 
 Credits: version 8.1
-====================
+--------------------
 
 Coq version 8.1 adds various new functionalities.
 
@@ -571,7 +575,7 @@ and Yale University.
 |
 
 Credits: version 8.2
-====================
+--------------------
 
 Coq version 8.2 adds new features, new libraries and improves on many
 various aspects.
@@ -665,7 +669,7 @@ the Coq-Club mailing list.
 |
 
 Credits: version 8.3
-====================
+--------------------
 
 Coq version 8.3 is before all a transition version with refinements or
 extensions of the existing features and libraries and a new tactic nsatz
@@ -739,7 +743,7 @@ Pierce for the excellent teaching materials they provided.
 |
 
 Credits: version 8.4
-====================
+--------------------
 
 Coq version 8.4 contains the result of three long-term projects: a new
 modular library of arithmetic by Pierre Letouzey, a new proof engine by
@@ -895,7 +899,7 @@ Eelis van der Weegen.
 |
 
 Credits: version 8.5
-====================
+--------------------
 
 Coq version 8.5 contains the result of five specific long-term projects:
 
@@ -1049,7 +1053,7 @@ Tankink. Maxime Dénès coordinated the release process.
 |
 
 Credits: version 8.6
-====================
+--------------------
 
 Coq version 8.6 contains the result of refinements, stabilization of
 8.5’s features and cleanups of the internals of the system. Over the
@@ -1189,7 +1193,8 @@ Dénès to put together a |Coq| consortium.
 |
 
 Credits: version 8.7
-====================
+--------------------
+
 |Coq| version 8.7 contains the result of refinements, stabilization of features
 and cleanups of the internals of the system along with a few new features. The
 main user visible changes are:
@@ -1294,8 +1299,7 @@ system, is now upcoming and will rely on Inria’s newly created Foundation.
 |
 
 Credits: version 8.8
-====================
-
+--------------------
 
 |Coq| version 8.8 contains the result of refinements and stabilization of
 features and deprecations, cleanups of the internals of the system along
--- a/doc/sphinx/index.html.rst
+++ b/doc/sphinx/index.html.rst
@@ -1,5 +1,3 @@
-.. _introduction:
-
 ==========================
 Introduction and Contents
 ==========================
@@ -81,9 +79,6 @@ Contents
    :caption: Reference
 
    zebibliography
-
-License
--------
 
 .. include:: license.rst
 
--- a/doc/sphinx/index.html.rst
+++ b/doc/sphinx/index.html.rst
@@ -1,13 +1,13 @@
 .. _introduction:
 
 ==========================
-Introduction
+Introduction and Contents
 ==========================
 
 .. include:: introduction.rst
 
-Table of contents
------------------
+Contents
+--------
 
 .. toctree::
    :caption: Indexes
--- a/doc/sphinx/index.latex.rst
+++ b/doc/sphinx/index.latex.rst
@@ -2,26 +2,22 @@
  The Coq Reference Manual
 ==========================
 
+------------
 Introduction
 ------------
 
 .. include:: introduction.rst
+
+.. include:: license.rst
 
 .. [#PG] Proof-General is available at https://proofgeneral.github.io/.
    Optionally, you can enhance it with the minor mode
    Company-Coq :cite:`Pit16`
    (see https://github.com/cpitclaudel/company-coq).
 
-Credits
--------
-
 .. include:: credits.rst
 
-License
--------
-
-.. include:: license.rst
-
+------------
 The language
 ------------
 
@@ -33,6 +29,7 @@ The language
    language/cic
    language/module-system
 
+----------------
 The proof engine
 ----------------
 
@@ -45,6 +42,7 @@ The proof engine
    proof-engine/detailed-tactic-examples
    proof-engine/ssreflect-proof-language
 
+---------------
 User extensions
 ---------------
 
@@ -53,6 +51,7 @@ User extensions
    user-extensions/syntax-extensions
    user-extensions/proof-schemes
 
+---------------
 Practical tools
 ---------------
 
@@ -62,6 +61,7 @@ Practical tools
    practical-tools/utilities
    practical-tools/coqide
 
+--------
 Addendum
 --------
 
--- a/doc/sphinx/introduction.rst
+++ b/doc/sphinx/introduction.rst
@@ -44,7 +44,7 @@ are processed from a file.
 .. seealso:: :ref:`thecoqcommands`.
 
 How to read this book
-=====================
+---------------------
 
 This is a Reference Manual, so it is not intended for continuous reading.
 We recommend using the various indexes to quickly locate the documentation
@@ -90,7 +90,7 @@ Nonetheless, the manual has some structure that is explained below.
    solvers and tactics. See the table of contents for a complete list.
 
 List of additional documentation
-================================
+--------------------------------
 
 This manual does not contain all the documentation the user may need
 about |Coq|. Various informations can be found in the following documents:
--- a/doc/sphinx/license.rst
+++ b/doc/sphinx/license.rst
@@ -1,3 +1,6 @@
+License
+-------
+
 This material (the Coq Reference Manual) may be distributed only subject to the
 terms and conditions set forth in the Open Publication License, v1.0 or later
 (the latest version is presently available at
--- a/doc/sphinx/proof-engine/proof-handling.rst
+++ b/doc/sphinx/proof-engine/proof-handling.rst
@@ -632,16 +632,15 @@ How to enable diffs
 ```````````````````
 
 .. opt:: Diffs %( "on" %| "off" %| "removed" %)
+   :name: Diffs
 
-  .. This ref doesn't work: :opt:`Set Diffs %( "on" %| "off" %| "removed" %)`
-
-  The “on” option highlights added tokens in green, while the “removed” option
-  additionally reprints items with removed tokens in red.  Unchanged tokens in
-  modified items are shown with pale green or red.  (Colors are user-configurable.)
+   The “on” option highlights added tokens in green, while the “removed” option
+   additionally reprints items with removed tokens in red.  Unchanged tokens in
+   modified items are shown with pale green or red.  (Colors are user-configurable.)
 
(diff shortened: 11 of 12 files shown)


jfehrle:
A couple things I noticed that I didn't try to fix (yet):

- index.latex.rst doesn't include the indexes anywhere in the document.  Seems like a bug.  I could change this but I'm not able to generate latex output, so I can't verify.
- index.html.rst changes should force rebuilding every html file since the short TOC is included in every html, but I noticed that incremental `make sphinx` doesn't check this dependency.  This may be in the Sphinx code, I didn't try to localize the problem.  I could document this fact somewhere (Makefile.doc or you tell me).
- It seems inelegant to repeat the HTML (60-80 lines) for the short table of contents in each HTML file.  Surely there is a way to share this info (which would solve the previous issue).
- Makefile.doc says `# Read INSTALL.doc to learn about the dependencies` but there is no INSTALL.doc.

cpitclaudel:
> index.latex.rst doesn't include the indexes anywhere in the document. Seems like a bug. I could change this but I'm not able to generate latex output, so I can't verify.

It's not a bug; the indices are in the PDF :)

> index.html.rst changes should force rebuilding every html file since the short TOC is included in every html, but I noticed that incremental make sphinx doesn't check this dependency. This may be in the Sphinx code, I didn't try to localize the problem.

This might be worth reproducing on a small example and reporting as a bug

> It seems inelegant to repeat the HTML (60-80 lines) for the short table of contents in each HTML file. Surely there is a way to share this info (which would solve the previous issue).

The only ways I know of are dynamic page generation (PHP templating), which is bad; frames, which are not great either (and might make pages slower to load, not faster); and Javascript content delivery (which adds complexity and will make the page less accessible).

jfehrle:
> It's not a bug; the indices are in the PDF :)

I'll take your word for it.  It's just that I see the following in index.html.rst but nothing that appears equivalent in index.latex.rst.

```
.. toctree::
   :caption: Indexes

   genindex
   coq-cmdindex
   coq-tacindex
   coq-optindex
   coq-exnindex
```

> frames, which are not great either (and might make pages slower to load, not faster)

It would be a little surprising if frames were noticeably slower.

I also observe that when you click on a top-level entry in the short TOC, the entire right hand panel gets loaded, which takes 3-4 seconds for the larger chapters when loading from my file system (Chrome on Windows 10 with an SSD).  You're not able to see any subentries in the short TOC until the new page is loaded.  It would be a better for the user if they could browse the subentries without this overhead.  Also, in this case, the TOC panel would not have to reload, only the right hand panel.  This might work better with a different HTML structure.  (I'm not saying this is urgent, just giving an analysis.)

Zimmi48:
>Fixed 2 incorrect option/table names in Flags, Options, Tables index.

This should have been part of another commit, even another PR. It would have ensured that this is merged faster. But well...

In any case, before we can properly review this (I intend to review mostly by looking at the rendered output), you have some fixing to do related to title levels (I'm not quite sure what is going on).

cpitclaudel:
> I also observe that when you click on a top-level entry in the short TOC, the entire right hand panel gets loaded, which takes 3-4 seconds for the larger chapters when loading from my file system (Chrome on Windows 10 with an SSD). You're not able to see any subentries in the short TOC until the new page is loaded. It would be a better for the user if they could browse the subentries without this overhead.

I don't understand this part :/

jfehrle:
> Indeed. I recently noticed this. Please fix. It should say ```Read [`doc/README.md`](doc/README.md)```

Sure, though in a Makefile comment it should just say `Read doc/README.md`, right?

jfehrle:
> This should have been part of another commit, even another PR.

It is a typo-level small cosmetic fix that doesn't affect readability.  I think separate PRs for every tiny detail is overkill--there's a fair bit of overhead just to put a PR through the process.  But if you prefer I will make it a separate commit/PR.

jfehrle:
> I don't understand this part :/

I'm happy to explain, but it's easier to do so if you give me an idea what part was unclear.

Here is the use case: I'm looking at the "Gallina specification language" chapter.  I want to see the subentries of "Extensions of Gallina" in the short (on-screen) table of contents.  Clicking on "Extensions of Gallina" takes 3-4 seconds before its subentries are visible because that loads the entire chapter as well.  That's pretty slow just for browsing the short TOC.

image


image

jfehrle:
> you have some fixing to do related to title levels (I'm not quite sure what is going on).

Ah, you must mean the job failure.  I wasn't seeing that error when I ran locally.

cpitclaudel:
> Here is the use case: I'm looking at the "Gallina specification language" chapter. I want to see the subentries of "Extensions of Gallina" in the short (on-screen) table of contents.

Oh, I see what you mean.  I think that's likely a theming issue. You could report this to the readthedocs people; that's where the theme is maintained.  I agree that it would be nice to load the full TOC on every page.

jfehrle:
I'll put in a request.  But I think it's not just an HTML theme issue because each HTML file only includes `toctree-l2` nodes for itself.

```
<li class="toctree-l1"><a class="reference internal" href="index.html">Table of contents</a></li>
<li class="toctree-l1"><a class="reference internal" href="introduction.html">Introduction</a></li>
<li class="toctree-l1 current"><a class="current reference internal" href="#">Credits</a><ul>
<li class="toctree-l2"><a class="reference internal" href="#credits-addendum-for-version-6-1">Credits: addendum for version 6.1</a></li>
<li class="toctree-l2"><a class="reference internal" href="#credits-addendum-for-version-6-2">Credits: addendum for version 6.2</a></li>
```

jfehrle:
I'm baffled by this error for latex.  The error gives the same filename twice.  The string "Introduction" only appears as a title once in introduction.rst and one in ssreflect-proof-language.rst.

Any ideas?

```
Warning, treated as error:
doc/sphinx/introduction.rst:4:duplicate label introduction, other instance in /builds/coq/coq/doc/sphinx/introduction.rst
Makefile.doc:63: recipe for target 'refman-latex' failed
```

jfehrle:
Ick.

The incremental Sphinx build gives misleading results.  For example, with SPHINXWARNERROR=1, it reports an error because it doesn't rebuild everything (doesn't track all the dependencies).  Then the failure causes Sphinx to reprocess every .rst file on the next run, which then may succeed after about 30 minutes on my system.

It also appears that there's some inconsistency in acceptable formats for chapter titles between html and latex.

Also the html and latex builds have different behavior for links.  The latex build reports a duplicate label for the following, which the html build accepts.  In Latex, apparently `License` creates a `_license` link, which is then a duplicate.  I have to remove the  `.. _license:` to get the latex build to succeed.

```
.. _license:

License
-------
```

Please take a look.

Zimmi48:
I agree that incremental builds can lead to inconsistent states. That can be pretty frustrating.

Zimmi48 (changes requested):
Looking at the rendered output, what I cannot accept for sure is that the Table of contents is the new index instead of the Introduction.

Regarding the move of the Indexes between the Preamble and The language part, I don't like this either because it makes finding the actual contents of the manual harder. With the indexes coming before the introduction, there is the advantage that you have the white highlight of the Introduction chapter actually marking the beginning of the manual.

cc @anton-trunov: You might want to give your opinion on this change.

anton-trunov:
@Zimmi48 Where can I see the rendered output?

Zimmi48:
https://coq.gitlab.io/-/coq/-/jobs/104051656/artifacts/_install_ci/share/doc/coq/sphinx/html/index.html
(this is in the checks section below)

anton-trunov:
I think that both of you have made good points. I agree with Jim that the indexes are still accessible. As a heavy user of the indexes I'm pretty happy with the current location of those (I mean before this PR), so I usually go straight to the top left corner of the screen. The new indexes location is not as clear, however, since the ToC shows the indexes right in the middle of the screen, I can get used to it.

I guess it all depends on who you optimize the layout for -- newcomers (Theo's introduction comment certainly makes sense in this case) or the every-day users, who wouldn't care about preliminaries.

While are on the topic, I'd like to mention that I find the Credits section of ToC sort of long -- I'd prefer it to be collapsed by default. Also, ToC could do without mentioning itself as the first entry -- it already has a link (the chain links symbol). It's just some random suggestions, feel free to ignore :)

anton-trunov:
> (this is in the checks section below)

thanks, I didn't notice it before :)

jfehrle:
> Looking at the rendered output, what I cannot accept for sure is that the Table of contents is the new index instead of the Introduction.

Can you elaborate?  I don't understand what you mean by "the Table of contents is the new index" or why you feel so strongly about this.

If you guys want to keep the index first, I can live with that.  I do think the Table of Contents should be a top-level entry in the small TOC so it's easy to discover.  I didn't realize that it was buried in the Introduction chapter after several months use of the new format.  Usually a Table of Contents appears before an Introduction, such as in most books, or, say, the OCaml Reference manual.

I did a lot of experiments trying to get subsections in the .rst file containing the `toctree` to appear as subentries in the short TOC, without success.  I could move the Credits and License sections under Introduction, which would make the short TOC a bit shorter.

jfehrle:
> While we are on the topic, I'd like to mention that I find the Credits section of ToC sort of long -- I'd prefer it to be collapsed by default. Also, ToC could do without mentioning itself as the first entry -- it already has a link (the chain links symbol).

Yes, the credits section has many subsections.  Are you suggesting that we hide or show menu subentries on a per-chapter basis?  Users might not like inconsistent behavior, and I don't know if that's easy to do.

If I remove "Table of Contents" from index.html.rst then it has no title at all, not good.  And it is conventional to have a title such as "Contents" above the text of the Table of Contents.

image

jfehrle:
BTW, there is an option that lets you browse the short TOC without having to reload each chapter. This includes all the menu entries in each html file, which takes about 870 lines.  I think it's worth the increased size.  I'll push updated code so you guys can take a look at it and try it.

Also, can we coqbot so it makes the pdf version of the doc available, too?  That could be helpful for debugging/verifying, especially for poor souls such as me who can't generate the pdf locally.

Zimmi48:
>Can you elaborate? I don't understand what you mean by "the Table of contents is the new index" or why you feel so strongly about this.

In this sentence I am talking about what is the main page people land on when opening the doc. It should be the introduction, not the TOC.

>I do think the Table of Contents should be a top-level entry in the small TOC so it's easy to discover.

I support this :+1: 

>Usually a Table of Contents appears before an Introduction, such as in most books, or, say, the OCaml Reference manual.

I am also OK with this, see my suggestion of putting it in the same part as the index, before the preamble.

>I could move the Credits and License sections under Introduction, which would make the short TOC a bit shorter.

License yes, but Credits no: I think it is better to keep as a separate chapter.

>Yes, the credits section has many subsections. Are you suggesting that we hide or show menu subentries on a per-chapter basis? Users might not like inconsistent behavior, and I don't know if that's easy to do.

What could actually be done would be to put the Credits and License at the end of the refman, close to bibliography. But I would need to have the opinion of other developers to validate this. I would also like this idea better if only the refman had fewer chapters.

>If I remove "Table of Contents" from index.html.rst then it has no title at all, not good. And it is conventional to have a title such as "Contents" above the text of the Table of Contents.

What gave you this idea?

>BTW, there is a config option that lets you browse the short TOC without having to reload each chapter. This includes all the menu entries in each html file, which takes about 870 lines. I think it's worth the increased size. I'll push updated code so you guys can take a look at it and try it.

:heart: 

>Also, can we update coqbot so it makes the pdf version of the doc available, too? That could be helpful for debugging/verifying, especially for poor souls such as me who can't generate the pdf locally.

I can point you to the link of the generated PDF manual but I don't think it would be useful for most PRs, so probably not worth having @coqbot do this.

Here you go (just replace `html/index.html` with `latex/CoqRefMan.pdf` in the URL):
https://coq.gitlab.io/-/coq/-/jobs/104477898/artifacts/_install_ci/share/doc/coq/sphinx/latex/CoqRefMan.pdf

Zimmi48:
BTW, the rendered PDF shows that indeed this PR is buggy regarding the LaTeX version (probably some inconsistencies between the updates to `index.html.rst` and `index.latex.rst`).

jfehrle:
> >If I remove "Table of Contents" from index.html.rst then it has no title at all, not good. And it is conventional to have a title such as "Contents" above the text of the Table of Contents.

> What gave you this idea?

I misread @anton-trunov's comment.  However if I remove "self" from toctree then the Table of Contents doesn't appear in the small TOC, either.

jfehrle:
> In this sentence I am talking about what is the main page people land on when opening the doc. It should be the introduction, not the TOC.

Got it.  The easiest way to do this is to use `.../intro-wrapper.html` as the top-level URL for the documentation on the Coq documentation page rather than `.../index.html`.  AFAICT, Sphinx automatically uses `index.html` for the table of contents, and subsections in the file containing `toctree` don't show up in the short or full TOC.

doc/sphinx/intro-wrapper.html.rst
```diff
@@ -0,0 +1,6 @@
+
+------------
+Introduction
+------------
+
+.. include:: introduction.rst
```
cpitclaudel: Quick question: why do you need this wrapper file?
Also, it would probably be better to rename this wrapper to introduction.rst and to include a file called `intro-contents.rst` instead.  This way, existing links won't break and the URL won't say `intro-wrapper.html`

jfehrle: The HTML and Latex versions require different formats for the title.  HTML:

```
------------
Introduction
------------
```

and Latex:

```
Introduction
------------
```

> Also, it would probably be better to rename this wrapper to introduction.rst

Will do.

Zimmi48:
Honestly, I'm not going to be competent enough to help you follow my requests. If you manage alone or with help from @cpitclaudel, good. I'm removing the milestone for now because it's clearly not going to be ready for 8.9+beta1. Maybe we can put it in 8.9.0.

jfehrle:
> Maybe we can put it in 8.9.0.

No worries, this change is not so urgent.

cpitclaudel:
Happy to help, but I've lost track.  Can I have a short summary? :innocent:

Zimmi48:
The main goal of this PR is to have a visible link to the TOC in the left-side panel. I would prefer if it were possible to do this while keeping everything else unchanged, especially the main page should continue to be the introduction and *I think* it would make more sense to keep the indexes before the preamble and include the TOC along them.
Regarding the PDF version, it is not clear what should be done. It seems to me that the most standard way of doing is to put the introduction *before* the TOC but maybe it makes sense to change this ordering as well. BTW @jfehrle disagrees on this statement:

>Usually a Table of Contents appears before an Introduction, such as in most books, or, say, the OCaml Reference manual.

Finally, @jfehrle noticed that:

>there is a config option that lets you browse the short TOC without having to reload each chapter

I am in favor of using it indeed. It helps with the browsing.

jfehrle:
OK guys, please take another look.  The PDF no longer has 2 Credits chapters.

Zimmi48:
>It's not under "INDEXES" because it's not an index. "PREAMBLE" is after "INDEXES".

What I was suggesting was to rename this part "Table of contents and indexes".

>References to the HTML documentation should link to ".../introduction.html" rather than ".../index.html". This effectively makes the introduction the starting page.

That doesn't work. It really should be the `index.html` page.

Anyways, I find this table of contents only marginally useful in the HTML version where there is also a TOC in the left panel (especially with your improvement allowing to browse this TOC without loading the corresponding chapters). Moreover, most of the browsing is through indexes anyways. So putting it as the main page really gives it too much importance.

Having a look at the 5 featured docs on readthedocs.io, I see that all of them include an introduction on the main page, which comes *before* the table of contents:

- https://bootstrap-datepicker.readthedocs.io/en/latest/
- http://docs.godotengine.org/en/3.0/
- https://packaging.python.org/ (this one is special, it doesn't show the TOC at all)
- http://docs.python-requests.org/en/master/
- https://solidity.readthedocs.io/en/v0.4.25/

Some of them are quite short so the TOC is not far, some are longer. Coq's own introduction is not the longest of them all. If you want, instead of all this infrastructure work, we can just work at making the introduction shorter.

jfehrle:
OK, it seems unlikely we'll get everything we want with a reasonable amount of effort.  I suspect that "index.html" is hardwired, unless @cpitclaudel knows otherwise.  How about if we keep the TOC in the Introduction, but change the page title to "Introduction and Contents"?  If that's agreeable, should it stay where it was under PREAMBLE?

> Anyways, I find this table of contents only marginally useful in the HTML version where there is also a TOC in the left panel (especially with your improvement allowing to browse this TOC without loading the corresponding chapters). Moreover, most of the browsing is through indexes anyways. So putting it as the main page really gives it too much importance.

I take your point about not having the TOC be the main page, but I think the full TOC is useful:
- It's not easy to skim the left hand panel--you have to do a lot of clicking to see the details.  A less-experienced user may want to skim the full TOC to get an overview or maybe search in it with the browser's built-in search.
- There's not (yet) a good index of topics and concepts.  It's hard to pick them out in the general index.
- The "Search docs" function is not (yet) that helpful--it doesn't show text snippets, only document titles, so there's a bunch of clicking to see what's what.  (I tend search in the .rst files.  The pdf is also useful for this purpose.)

Zimmi48:
Yep, I see your point indeed although we might argue that a new user will read the introduction and then will discover that the table of contents is there, while an experimented user will already know it is there.
Nonetheless, I don't have any opposition to renaming the "Introduction" page to "Introduction and Contents". We can also add a sentence mentioning this table of contents to the "How to read this book" section.

jfehrle:
> We can also add a sentence mentioning this table of contents to the "How to read this book" section.

That may be more trouble than it's worth.  Such a sentence only makes sense for the HTML output and it's in the middle of `introduction.rst`.

jfehrle:
@Zimmi48 Please take a look, thanks.

Zimmi48:
>That may be more trouble than it's worth. Such a sentence only makes sense for the HTML output and it's in the middle of introduction.rst.

I am talking about something like "(see the table of contents below)". I don't see the problem with such sentence in the PDF version.

Zimmi48 (approved):
Looks good but it would look better if the (preexisting) inconsistency in title levels between the introduction and the table of contents title could be fixed (use `-------` in `introduction.rst`).

jfehrle:
> I am talking about something like "(see the table of contents below)". I don't see the problem with such sentence in the PDF version.

In the PDF, the table of contents appears before the introduction, so to be correct the PDF should say "(see the table of contents **above**)"  You're OK with such an inconsistency?  I thought it was too much trouble to have different wording between the HTMl and the PDF.

jfehrle:
> (use ------- in introduction.rst)

Done.

Zimmi48:
>In the PDF, the table of contents appears before the introduction

OK, that's correct. I just got confused because I was used to the way it used to be in 8.7 and before (https://coq.inria.fr/distrib/V8.7.2/files/Reference-Manual.pdf).

But you just made me notice that there is something off with the current refactoring and it is due to the change in title levels I just asked you. Here is how it looked before: https://coq.gitlab.io/-/coq/-/jobs/105905462/artifacts/_install_ci/share/doc/coq/sphinx/latex/CoqRefMan.pdf
And here is how it looks now: https://coq.gitlab.io/-/coq/-/jobs/106286179/artifacts/_install_ci/share/doc/coq/sphinx/latex/CoqRefMan.pdf
"How to read this book" and "List of additional documentation" are now separate chapters! But that's of course due to the use of `-----` in `index.latex.rst`.
So it seems that the problem is that the PDF manual needs one more intermediate level of titles. I can only suggest changing this to use the level below. But also, following this realization I want to do a more in-depth review of the current situation regarding title levels. Please bear with me.

Zimmi48:
@jfehrle I added a commit on top of your branch that fixes the issue with title levels that I had noticed and I rebased the branch on latest master.

@cpitclaudel Since I'm now a co-author of this PR (and I have a commit undoing some of the stuff you did when preparing the PDF manual), can you be the assignee here? (BTW, I have two other open doc PRs that you could also merge :smiley: )

cpitclaudel (approved):

jfehrle:
Hi @cpitclaudel, will you be merging this soon?

cpitclaudel:
Done! Thanks for the reminder.
Btw, I noticed that one of the commits had a long title.  I like the format advocated at https://chris.beams.io/posts/git-commit/ :smiley:

jfehrle:
Fair enough.  But are 14 screen fulls in that page really necessary to advocate brevity and readability? :-)

jfehrle:
@Zimmi48 Your turn to add to 8.9-beta1.

PR #8618, opened 2018-10-02
[dune] [tools] Cleanup and extra install.

ejgallego:
We remove unused files in `tools` and install a few missing ones from the Dune build.


Zimmi48 (approved):
Welcome clean-up. Regarding the additional install, it looks good. I haven't checked that the location is the same as with the Makefile. I suppose you have tested this and this works. I also assume that this is required to make the test-suite run from Dune.

PR #8620, opened 2018-10-02
Docs: Missing backquote

nomeata:


doc/sphinx/language/gallina-extensions.rst
```diff
@@ -606,7 +606,7 @@ for several ways of defining a function *and other useful related
 objects*, namely: an induction principle that reflects the recursive
 structure of the function (see :tacn:`function induction`) and its fixpoint equality.
 The meaning of this declaration is to define a function ident,
-similarly to ``Fixpoint`. Like in ``Fixpoint``, the decreasing argument must
+similarly to ``Fixpoint``. Like in ``Fixpoint``, the decreasing argument must
```
Zimmi48: Thanks, but what would be even better would be to replace these two ``` ``Fixpoint`` ``` with ``` :cmd:`Fixpoint` ``` so that there is a link to the definition of the command.

nomeata:
Don't make perfect the enemy of good, this is justa drive-by bug fix :-)

Zimmi48:
>Don't make perfect the enemy of good, this is justa drive-by bug fix :-)

I'm always trying to make great contributors into excellent contributors by teaching them stuff :-)
But OK, merging as-is.

PR #8623, opened 2018-10-02
[dune] Fix PHONY target in Dune's helper makefile.

ejgallego:


Zimmi48 (approved):

Zimmi48:
In fact, none of these targets are telling me "missing PHONY" when I test them without the fix. So what is this fixing?

ejgallego:
> In fact, none of these targets are telling me "missing PHONY" when I test them without the fix. So what is this fixing?

What do you mean? Targets that are not a file should be declared `PHONY` as to preventively avoid problems.

Zimmi48:
OK it's just that I'm used to Makefile detecting these problems and raising a warning but here it wasn't the case.

SkySkimmer:
That's actually a custom Coq thing

%:
	@echo "Error: no rule to make target $@ (or missing .PHONY)" && false

in Makefile.build

Since Makefile.dune doesn't include it that doesn't happen.

Gaëtan Gilbert

PR #8624, opened 2018-10-02
Adding library object methods for functional updates.

herbelin:
**Kind:** architecture

This PR is a possible proof of concept for stepwise moving `declare_object` and `declare_summary` to a functional model.

Some summaries can be declared as functional using `declare_functional_summary`. In return, this gives a `tag` to project from the functionally-threaded state and to modify the functionally-threaded state.

For objects which come with functional summaries, new methods `cache_sps_object`, `load_sps_object` and `open_sps_object` are provided as functional alternatives to `cache_object`, `load_object` and `open_object`. Then `add_anonymous_sps_leaf` is used for functional modification of the state.

Note: `rebuild`/`discharge` remain to be done.

### DIFF
--- a/library/summary.mli
+++ b/library/summary.mli
@@ -91,5 +91,16 @@ val modify_summary : frozen -> 'a Dyn.tag -> 'a -> frozen
 val project_from_summary : frozen -> 'a Dyn.tag -> 'a
 val remove_from_summary : frozen -> 'a Dyn.tag -> frozen
 
+(** Functional summaries *)
+
+type functional_summaries
+
+val declare_functional_summary : name:string -> 'a -> 'a Dyn.tag
+val get_functional_summaries : unit -> functional_summaries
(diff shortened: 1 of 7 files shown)


library/summary.mli
```diff
@@ -91,5 +91,16 @@ val modify_summary : frozen -> 'a Dyn.tag -> 'a -> frozen
 val project_from_summary : frozen -> 'a Dyn.tag -> 'a
 val remove_from_summary : frozen -> 'a Dyn.tag -> frozen
 
+(** Functional summaries *)
+
+type functional_summaries
```
ejgallego: Why do you need a new type that seems like the same we had before?

SkySkimmer:
We've had time to look at it, let's close and focus discussion on the cep.

herbelin:
OK

PR #8630, opened 2018-10-02
Some cleaning in the test suite

vbgl:
Aim: generate .vo files for most test cases and run `coqchk` on them.


ejgallego:
Wouldn't it easier to allow `coqc/coqchk` handle modules that are a number [even if they cannot be referenced].

vbgl:
Files whose name is a number is the slightest issue: many other names are not valid but a lot of test files do not define a module (unclosed modules, proofs in progress…).

Zimmi48:
>Wouldn't it easier to allow coqc/coqchk handle modules that are a number [even if they cannot be referenced].

This is a recurring debate and given that it was never solved, I'm OK with either solution.

gares (approved):

PR #8617, opened 2018-10-01
[dune] Provide workspace file with all supported Ocaml builds.

ejgallego:

PR #8631, opened 2018-10-02
[dune] Fix couple of minor bugs in #8617

ejgallego:
I forgot to update PHONY and to put the proper flags fixed by other
PRs.

### DIFF
--- a/dev/dune-workspace.all
+++ b/dev/dune-workspace.all
@@ -2,7 +2,7 @@
 
 ; Add custom flags here. Default developer profile is `dev`
 (env
- (dev     (flags :standard -rectypes -w -9-27-50))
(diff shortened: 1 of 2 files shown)


dev/dune-workspace.all
```diff
@@ -2,7 +2,7 @@
 
 ; Add custom flags here. Default developer profile is `dev`
 (env
- (dev     (flags :standard -rectypes -w -9-27-50))
+ (dev     (flags :standard -rectypes -w -9-27-50+60))
```
Zimmi48: Is there any way to avoid this problem by "including" the default profile?

ejgallego: A future refactoring should take care of it

PR #8632, opened 2018-10-02
[ide] [dune] [test-suite] Reorganize `fake_ide` build.

ejgallego:
Even if `fake_ide` was under tools, it depended on libraries from
`ide`. Thus, we move `fake_ide` to `ide`.

In the Dune side, we reorganize libraries so `fake_ide` doesn't depend
on GTK anymore, this allows to run the test-suite when GTK is not
available.

In order to achieve this, we had to split the `coqide` package in a
server and client version.

EDIT: depends on #8657.

### DIFF
--- a/ide/dune
+++ b/ide/dune
@@ -1,11 +1,34 @@
+; IDE Server
 (ocamllex utf8_convert config_lexer coq_lex)
 
 (library
  (name core)
- (public_name coqide.core)
+ (public_name coqide-server.core)
  (wrapped false)
- (modules (:standard \ idetop coqide_main))
- (libraries threads str lablgtk2.sourceview2 coq.lib coqide.protocol))
+ (modules document)
+ (libraries coq.lib))
+
+(executable
+ (name fake_ide)
+ (modules fake_ide)
+ (libraries coqide-server.protocol coqide-server.core))
+
+(executable
+ (name idetop)
+ (public_name coqidetop.opt)
+ (package coqide-server)
+ (modules idetop)
+ (libraries coq.toplevel coqide-server.protocol)
+ (link_flags -linkall))
+
+; IDE Client
+(library
+ (name gui)
+ (public_name coqide.gui)
+ (wrapped false)
+ (modules (:standard \ document fake_ide idetop coqide_main))
+ (optional)
+ (libraries coqide-server.protocol coqide-server.core lablgtk2.sourceview2))
 
 (rule
  (targets coqide_main.ml)
@@ -17,12 +40,4 @@
  (public_name coqide)
  (package coqide)
  (modules coqide_main)
- (libraries coqide.core))
-
-(executable
- (name idetop)
- (public_name coqidetop.opt)
- (package coqide)
- (modules idetop)
- (libraries coq.toplevel coqide.protocol)
- (link_flags -linkall))
+ (libraries coqide.gui))
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -241,6 +241,7 @@ pkg:opam:
   script:
     - set -e
     - opam pin add coq .
+    - opam pin add coqide-server .
     - opam pin add coqide .
     - set +e
   variables:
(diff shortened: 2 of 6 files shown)


gares (approved):
makes sense to me

ppedrot (commented):
No particular opinion.

Zimmi48 (commented):
The separation between the server side and the actual ide makes sense to me but I'd rather suggest to put the server-part in the main Coq package instead of creating a new one. This is likely something that we want to ship every time (this or the protocol that replaces it in the future) because we want all clients to use such protocols.

ejgallego:
> The separation between the server side and the actual ide makes sense to me but I'd rather suggest to put the server-part in the main Coq package instead of creating a new one. This is likely something that we want to ship every time (this or the protocol that replaces it in the future) because we want all clients to use such protocols.

I'd rather keep the decoupling at this level and introduce a `coq` meta package that depends on `coq-code` `coq-language-server`. IMO the coupling of the server with Coq is not ideal as in general, would we ever develop a richer language server, it will almost for sure have to depend on some PPX/extra libraries.

ejgallego:
I guess this kind of organization is very subjective, for example I hope that soon I will place the `stdlib` in its own package too. I like to have smaller packages as for example you can update the stdlib or the server separatedly, but indeed it is not a settled science.

ide/coqide-server.opam
```diff
@@ -0,0 +1,19 @@
+opam-version: "1.2"
+maintainer: "The Coq development team <coqdev@inria.fr>"
+authors: "The Coq development team, INRIA, CNRS, and contributors."
+homepage: "https://coq.inria.fr/"
+bug-reports: "https://github.com/coq/coq/issues"
+dev-repo: "https://github.com/coq/coq.git"
+license: "LGPL-2.1"
+
+available: [ocaml-version >= "4.02.3"]
```
Zimmi48: This line must likely be updated. In fact, this is also the case in `coq.opam` and `coqide.opam`.

ide/dune
```diff
@@ -1,11 +1,36 @@
...
+ (public_name coqide-server.core)
  (wrapped false)
- (modules (:standard \ idetop coqide_main))
- (libraries threads str lablgtk2.sourceview2 coq.lib coqide.protocol))
+ (modules document)
+ (libraries coq.lib))
+
+(executable
+ (name fake_ide)
+ (public_name fake_ide)
```
Zimmi48: If this is installed, then I guess we should follow the convention of prefixing it with `coq`. Note that it isn't installed in the make-based build (at least not in `/bin`).

Zimmi48 (approved):
Apart from the small nits, this looks good to me.

ide/coqide-server.opam
ejgallego: Well, I didn't update this as I am not sure if we want to migrate to OPAM 2 or not.

ide/dune
ejgallego: Good point, I was indeed not sure whether to install it or not. If we install it, it'll live in `PATH` which is convenient, if we don't install it maybe indeed should be just moved to `test-suite/ide` and made private to it?

ide/coqide-server.opam
ejgallego: But indeed I guess I'll submit a PR fixing several OPAM issues, for 1.2 for now.

ide/dune
Zimmi48: Yes, it would make more sense IMHO to be made private to test-suite.

ejgallego: Done then, it is easy to change anyways.

ide/coqide-server.opam
Zimmi48: So now that you fixed the other opam files in #8650, please fix this one here.

ejgallego: Fixed, made this depend on that PR.

ejgallego:
Ready to merge, CI permitting.

Zimmi48:
@ppedrot You self-assigned this.

PR #8639, opened 2018-10-02
[api] Be more explicit about deprecation of debug printers.

ejgallego:
As suggested by @mattam82

### DIFF
--- a/engine/termops.mli
+++ b/engine/termops.mli
@@ -341,10 +341,10 @@ val print_constr : constr -> Pp.t
 end
 
 val print_constr : constr -> Pp.t
-[@@deprecated "use Internal.print_constr_env"]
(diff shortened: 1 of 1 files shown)


ppedrot (approved):
This might be slightly too emphatic, but that's not really important. Will merge tomorrow.

ejgallego:
> This might be slightly too emphatic, but that's not really important. Will merge tomorrow.

Let me try an alternate wording.

Zimmi48:
Sorry for the noise. I thought in this case the rebase label was spurious but in fact it was not...

ejgallego:
> Sorry for the noise. I thought in this case the rebase label was spurious but in fact it was not...

In fact it was not :) we are talking about a race of seconds :)

Issue #7510, opened 2018-05-14
Almost all uses of ** in the docs are wrong


cpitclaudel:
#### Description of the problem

Most uses of `**` in the docs seem wrong. Some examples:

```
./proof-engine/tactics.rst\01528:   is **IH**.
Should be ``

./proof-engine/tactics.rst\03566:**Variants:**
Should use .. cmdv::

./language/cic.rst:1436:1:**Example.**
./addendum/generalized-rewriting.rst\066:**Example 1 (Parametric relation):**
Should use .. example::

./addendum/micromega.rst\078:**Theorem (Psatz)**. Let :math:`S` be a set of polynomials.
Should use a custom theorem environment

./language/coq-library.rst:739:  * **Logic** : Classical logic and dependent equality
./language/module-system.rst\040:**Module definition.** A module definition is written :math:`\Mod{X}{S}{S'}`
Should be a definition list (maybe)

./user-extensions/syntax-extensions.rst\0693:**Syntax of notations**
./language/cic.rst\01520:**Definition of ι-reduction.**
./addendum/extraction.rst\0186:**Inlining and printing of a constant declaration:**
./practical-tools/utilities.rst\0832:**Contents options**
Should be subsection titles
```

Zimmi48:
Interesting grepping!

rudi-cilibrasi:
```
[doc] grep -n '[*][*]' **/*.rst  | wc                                   master 
     88     678    7896
```

```
sphinx/addendum/extended-pattern-matching.rst:90:to right. A match expression is generated **only** when there is at least
sphinx/addendum/extraction.rst:184:**Inlining and printing of a constant declaration:**
sphinx/addendum/extraction.rst:340:  * The ML code provided by the user is currently **not** checked at all by
sphinx/addendum/extraction.rst:345:    ad-hoc pattern matching) will often **not** be fully rigorously
sphinx/addendum/extraction.rst:351:  * Translating an inductive type to an arbitrary ML type does **not**
sphinx/addendum/micromega.rst:78:**Theorem (Psatz)**. Let :math:`S` be a set of polynomials.
sphinx/addendum/micromega.rst:175:**Theorem**. Let :math:`p` be an integer and :math:`c_1` and :math:`c_2`
sphinx/language/cic.rst:82:**Implementation issues** In practice, the Type hierarchy is
sphinx/language/cic.rst:146:**Free variables.**
sphinx/language/cic.rst:153:**Substitution.**
[...]

Issue #6254, opened 2017-11-27
Future of sections


SkySkimmer:
I think the discussion in #883 should be continued in an issue.

I'll quote @herbelin as an opening post:

> Hi, I remain unclear about how to make progresses on section variables (and, incidentally, on this PR). I did not re-read carefully all comments but the general idea seems to be well-represented by this comment from @silene: « the intuitive behavior is to clear these variables. It is just that, until all the bugs related to clearing section variables are fixed, we should refrain from automatically clearing section variables by default. »
> 
> So, how to step further and to eventually address the clearing section variables bugs, one way or another? Shall we think at a different model, e.g. based on contexts? With which transition strategy?
> 
> Shall we first move sections out of the kernel with on-the-fly generalization of the constants? But with which semantics for clearing a section variable? Forbidden? Allowed, but cutting all dependencies of the cleared variables in global objects?
> 
> More generally, is the dependency of notations, implicit arguments, hints, etc. in section variables a problem, or, on the opposite, we would like to eventually associate notations, or implicit arguments, etc. also to goal variables (think e.g. at a local polymorphic function which we just posed in the context and to which we would like to set implicit arguments and define an infix notation).
> 
> Depending on what we want, different models can be considered. For instance, we could invent a systematic way to refer to goal variables which were cleared (since after all, they are still there somewhere in the spine of lambdas of the current proof being constructed). Or, to control the effect of a clear, we could imagine that we maintain a table telling for each goal variable in which other objects it depends in.
> 
> So, shortly, it seems to me that it is first a lot about what user model we'd like to eventually propose, and that once the model is clear the technical constraints will resolve by themselves.
> 
> Any comments?

SkySkimmer:
Personally I don't understand what contexts are meant to be.

I think we want something in the environment to be able to decide if `A` is a section variable or an evar variable. This could be separating the name context into a section and an evar context, but I also considered the following idea:
Suppose we change constr so it goes `type (...) kind = ... | Var of variable | ...` where `variable = Id.t * tag` and `tag = int option`. Tag `None` means a section variable, tag `Some i` means an evar variable with the value of `i` being used to fail when a tactic does 
```coq
let x := constr(A (* Some i *)) in
  clear A (*, Some i *); 
  intros A (* , Some fresh j *);
  do_something_with_x (* (A, Some i) not in context (A, Some j) *)
```

[...]

PR #8642, opened 2018-10-03
Compiled interfaces with -vos and -vok options

charguer:
**Kind:** feature 

The goal is to propose an alternative solution to using .vio for compiling statements without compiling proofs, to enable recovering a working state as fast as possible when modifying a file in the dependency chain. 

In particular, this proposal fixes the two major issues associated with  .vio files: their huge size, and the fact that the choice of "the most recent file among .vio and .vo" interacts badly with "make".

- Corresponding documentation was added in doc/sphinx, and in --help output.
- Not yet done: updated list of warnings, entry in CHANGES.
- This feature has been tested so far on the compilation of TLC and CFML2 (my own projects).

Description of the feature: please read  doc/sphinx/addendum/compiled-interfaces.rst  (in the diff below).

On the way, I converted tabs to spaces in the files that I did edit. And I also optimize "coqdep" so that it won't traverse all the files twice, once for computing .vo dependencies and once again for .vio dependencies.

Two open questions:
- is "Include" handled properly? if not, what needs to be done?
- which are the practical use cases for which .vio batch compilation and verification would remain useful?


ejgallego (commented):
On a very quick look I'd say the patch looks pretty good; a few (non-binding) comments below.

I need to think more, but my main question is if we want to have so many modes or instead just have a single mode and have the ability to check for missing proofs.

In a sense, a missing proof is not so different from a proof that didn't arrive yet due to async processing, so I'd favor a more unified view even if of course this PR should not be responsible of enforcing that.

lib/flags.ml
```diff
@@ -50,6 +50,10 @@ let test_mode = ref false
 let async_proofs_worker_id = ref "master"
 let async_proofs_is_worker () = !async_proofs_worker_id <> "master"
 
+let load_vos_libraries = ref false
+
+let complain_on_proof_without_using_in_section = ref false
```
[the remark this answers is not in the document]
charguer: I investigated whether this was possible, but it seems that quite a few functions would need to receive extra arguments for propagating the information about load_vos_library flag all the way from coqtop.ml through stm.ml and library.ml. I'm happy to do the change if someone tells me what's a clean way to do it.

tools/coqdep_common.ml
```diff
@@ -239,15 +239,15 @@ let file_name s = function
...
     | None, Some mldir ->
-	let mlfile = file_name str mldir in
-	(" "^mlfile^".cmo"," "^mlfile^".cmx")
+        let mlfile = file_name str mldir in
+        (" "^mlfile^".cmo"," "^mlfile^".cmx")
     | Some mlidir, None ->
-	let mlifile = file_name str mlidir in
-	(" "^mlifile^".cmi"," "^mlifile^".cmi")
+        let mlifile = file_name str mlidir in
+        (" "^mlifile^".cmi"," "^mlifile^".cmi")
```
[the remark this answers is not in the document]
charguer: I am sorry but when I find a file with tabs in random places, mixed with spaces sometimes on the same line, I just can't work with that file, so I fix it.

doc/sphinx/addendum/compiled-interfaces.rst
```diff
@@ -0,0 +1,85 @@
...
+nontrivial proofs, or the files have many linear chains of dependencies, or
+one has many cores available, compilation should be overall faster.
+
+Need for `Proof using`
+----------------------
+
+When a theorem is part of a section, typechecking the statement of this
+theorem might be insufficient for deducing the type of this statement as
+of at the end of the section. Indeed, the proof of the theorem could make
+use of section variables or section hypotheses that do are not mentioned
```
Zimmi48: Spurious `do`.

doc/sphinx/addendum/compiled-interfaces.rst
```diff
@@ -0,0 +1,85 @@
...
+in the statement of the theorem.
+
+For this reason, proofs inside section should begin with `Proof using [..].`
+instead of `Proof`, where after the `using` one should provide the list
+of the names of the section variables that are required for the proof
+but are not involved in the typechecking of the statement. For uniformity,
+one is encouraged to use `Proof using` for all proofs, even those that do
```
Zimmi48: Use ```:cmd:`Proof using` ``` instead to have a link to the description of the command.

doc/sphinx/addendum/compiled-interfaces.rst
```diff
@@ -0,0 +1,85 @@
...
+not fall in a section.
```
Zimmi48: I don't get this advice.

doc/sphinx/addendum/compiled-interfaces.rst
```diff
@@ -0,0 +1,85 @@
...
+
+If |Coq| is invoked using the `-vos` option, whenever it finds the command
+`Proof.` inside a section, it will compile the proof, that is, refuse to
+skip it, and it will raise a warning. To disable the warning, one may pass
+the flag `-w -proof-without-using-in-section`.
```
Zimmi48: Document the warning using `.. warn::`

doc/sphinx/addendum/compiled-interfaces.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Limitations
+-----------
+
+The interpretation of `Include` for importing modules might not work
+properly with `-vos`, for the same reason as it might not work properly
+with `-quick`. Hopefully, this could be resolved in the near future.
```
Zimmi48: Seems like wishful thinking.

doc/sphinx/addendum/compiled-interfaces.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Compiling `.vok` files is not as trustworthy as building `.vo` files,
+because some universe constraints might get lost (?). So, it can be a
```
Zimmi48: Determine if this is trustworthy or not and then write a more affirmative claim.

Zimmi48 (commented):
General comments on the doc (not taking sides on the feature):

I personally do not like at all the current structure of the reference manual with a very long "Addendum", especially to put very short chapters there, so I would very much prefer if this documentation could be included in the existing chapter where it is most relevant (The Coq commands?). If this is not adapted to add to an existing chapter, then do create a new chapter but put it where it makes most sense (probably the Practical tools part). In any case, whatever is decided on the location, if this is a new chapter, it should also be added to the table of contents (`index.{html,latex}.tex`).

Another minor remark: single backticks are not to be used alone. For normal verbatim, use double backticks. Single-backticks can be combined with environments like `:cmd:`. Have a look at https://github.com/coq/coq/tree/master/doc/sphinx#documenting-coq-with-sphinx to learn more.

tools/coqdep_common.ml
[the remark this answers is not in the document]
ejgallego: Yeah that's a hard topic due to the tradeoff of noise history, which in a project such as Coq it is of critical importance, so the tendency is to try to minimize whitespace noise. I suggest as way to improve tho:
- Do `untabify` of the file first, put that in a separate commit; then rebase the current PR on top of that commit.

Then, in the final merge, we can choose to drop the untabify commit if deemed convenient.

library/library.ml
```diff
@@ -679,15 +721,20 @@ let error_recursively_dependent_library dir =
 (* Security weakness: file might have been changed on disk between
    writing the content and computing the checksum... *)
 
-let save_library_to ?todo dir f otab =
+(** Note: create_vos could be replaced by a direct read of
+    !Flags.compilation_mode, but this does not seem to be
+    the way things are done for .vio files. *)
+
+let save_library_to ~create_vos ?todo dir f otab =
```
gares: I think the API is misleading here. Passing `~todo:(Some ..)` and `~create_vos:true` seems meaningless.
I suggest crafting a data type along these lines
```
type todo_proofs = `NothingToDo (* for vo *) | `Taks of ... Future.t list (* for .vio *) | `EmptyTasks (* for .vos *)
```
It may also make sense to put together the opaque tables and the todo list while you are at it
```
type proofs = { proofs_done : Opaquetables.t ; proofs_todo = [ `Nothing | `Tasks ... | `EmptyTaks ... ] }
```
today the tasks are meant to fill in the missing items in the opaque table.

vernac/vernacentries.ml
```diff
@@ -2246,6 +2264,14 @@ let interp ?proof ~atts ~st c =
     let using = Option.append using (Proof_using.get_default_proof_using ()) in
     let tacs = if Option.is_empty tac then "tac:no" else "tac:yes" in
     let usings = if Option.is_empty using then "using:no" else "using:yes" in
+    if !Flags.complain_on_proof_without_using_in_section
```
SkySkimmer: This flag is redundant with the warning control system.

SkySkimmer (dismissed):

ejgallego:
@charguer could you push a rebase when you can? Thanks!

ppedrot (commented):
The core part of the implementation looks OK, but we need more feedback before going further.

ppedrot:
Who should be the assignee for this PR? @gares? @ejgallego? Myself?

ejgallego:
> Who should be the assignee for this PR? @gares? @ejgallego? Myself?

I dunno, I could try to help if you think that'd be fine, however I am strictly less knowledgeable about most parts of this PR than either both of you.

vernac/vernacentries.ml
charguer: I was not able to figure out a better way to integrate in the cwarning system. The behavior I want is the following:  when encouterning a "proof" (without "using") inside a section, a warning should only be triggered if the compilation mode is "-vos", which I keep track of using "Flags.complain_on_proof_without_using_in_section". When this mode is used, it should still be possible for the user to turn off such warnings. We could have the warning get triggered for every compilation mode (including .vo production), and ensure that the warning is "Disabled". However, this would cause two issues. First, it is not very clean because it generates many unnecessary calls to the warning system. Second, it is problematic if the user passes the option to turn off the warning in question before passing in the option -vos. In that case, the warning will be Enabled and not Disabled as requested.

charguer: If you find a better way to proceed, please suggest.

SkySkimmer: >a warning should only be triggered if the compilation mode is "-vos"

Why should -vos change the warnings used?

charguer: By default, in a -vo compilation, you don't want to report warnings for "Proofs" (without "using") in section. However, by default, in a -vos compilation, you do want to report such warnings.

library/library.ml
charguer: @gares I just tried implementing your suggestion, but there is a catch it seems. At the beginning of `save_library`, I need to compute the variable `except` which lists what proof contents should not be  exported by the global dump. However, with your datatype, I don't have at hand the list of futures of which to extract the identifiers to put into the exclude list. Unless you tell me it is safe to use `let except = Future.UUIDSet.empty` , I would need to  include the list of futures also in the constructor for the `.vio` mode. Please let me know.

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
...
+      by producing standard `.vo` files. (Technically, when using `-vos`,
+      fewer universe constraints are collected.)
+
+   .. warning::
+
+      In the current implementation, the use of the command `Include` for
+      importing modules compiled using `-vos` might not work properly.
+
+
+**Typical usage.**
```
Zimmi48: You can use a sub-section title level (cf. http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections). See also #7510.

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
   currently associated color and exit.
 :-h, --help: Print a short usage and exit.
 
+
+
+Compiled interfaces (produced using `-vos`)
+----------------------------------------------
+
+
+Compiled interfaces helps saving time while developing Coq formalizations,
```
Zimmi48:
```suggestion
Compiled interfaces help saving time while developing Coq formalizations,
```

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
   currently associated color and exit.
 :-h, --help: Print a short usage and exit.
 
+
+
+Compiled interfaces (produced using `-vos`)
```
Zimmi48:
```suggestion
Compiled interfaces (produced using ``-vos``)
```

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
...
+
+Compiled interfaces helps saving time while developing Coq formalizations,
+by compiling the formal statements exported by a library independently of
+the proofs that it contains.
+
+   .. warning::
+
+      Compiled interfaces should only be used for development purposes.
+      At the end of the day, one still needs to proof check all files
+      by producing standard `.vo` files. (Technically, when using `-vos`,
```
Zimmi48:
```suggestion
      by producing standard ``.vo`` files. (Technically, when using ``-vos``,
```

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
...
+      fewer universe constraints are collected.)
+
+   .. warning::
+
+      In the current implementation, the use of the command `Include` for
```
Zimmi48:
```suggestion
      In the current implementation, the use of the command ``Include`` for
```

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
...
+      importing modules compiled using `-vos` might not work properly.
```
Zimmi48:
```suggestion
      importing modules compiled using ``-vos`` might not work properly.
```

"might" or "does"?

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
...
+
+
+**Typical usage.**
+
+Assume a file ``foo.v`` that depends on two files ``bar1.v`` and ``bar2.v``. The
+command ``make foo.requires_vos`` will compile ``bar1.v`` and ``bar2.v`` using
```
Zimmi48: I'm not sure I understand this sentence. Is this something supported by `coq_makefile`?

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
...
+the option ``-vos`` to skip the proofs, producing ``bar1.vos`` and ``bar2.vos``.
+At this point, one is ready to work interactively on the file ``foo.v``, even
+though it was never needed to compile the proofs involved in the ``bar*.v``
+files.
+
+Assume a set of files ``f1.v ... fn.v`` with linear dependencies. The command
+``make vos`` enables compiling the statements (i.e. excluding the proofs) in all
+the files. Next, ``make -j vok`` enables compiling all the proofs in parallel.
```
Zimmi48: Again, that's something supported by `coq_makefile`?

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
...
+**Need for ``Proof using``**
+
+When a theorem is part of a section, typechecking the statement of this theorem
+might be insufficient for deducing the type of this statement as of at the end
+of the section. Indeed, the proof of the theorem could make use of section
+variables or section hypotheses that are not mentioned in the statement of the
+theorem.
+
+For this reason, proofs inside section should begin with :cmd:`Proof using`
+instead of ``Proof``, where after the ``using`` one should provide the list of
```
Zimmi48:
```suggestion
instead of :cmd:`Proof`, where after the ``using`` clause one should provide the list of
```

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
...
+the names of the section variables that are required for the proof but are not
+involved in the typechecking of the statement. Note that it is safe to write
+``Proof using.`` instead of ``Proof.`` also for proofs that are not within a
+section.
+
+.. warn:: You should use the “Proof using [...].” syntax instead of “Proof.” to
+enable skipping this proof which is located inside a section. Give as argument
```
Zimmi48: If I recall correctly, this directive does not allow splitting the error message on several lines.

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
...
+     refuse to skip it, and it will raise a warning. To disable the warning, one
+     may pass the flag ``-w -proof-without-using-in-section``.
+
+**Interaction with standard compilation**
+
+When compiling a file ``foo.v`` using |coqc| in the standard way (i.e., without
+``-vos`` nor ``-vok``), an empty file ``foo.vos`` is created in addition to the
+regular output file ``foo.vo``. If |coqc| is subsequently invoked on some other
+file ``bar.v`` using option ``-vos`` or ``-vok``, and that ``bar.v`` requires
+``foo.v``, if |Coq| finds an empty file ``foo.vos``, tyen it will load
```
Zimmi48:
```suggestion
``foo.v``, if |Coq| finds an empty file ``foo.vos``, then it will load
```

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,89 @@ and ``coqtop``, unless stated otherwise:
...
+``foo.vo`` instead of ``foo.vos``.
+
+The purpose of this feature is to allow users to benefit from the ``-vos``
+option even if they depend on libraries that were compiled in the traditional
+manner (i.e., never compiled using the ``-vos`` option).
```
Zimmi48: What about being able to read the `.vo` files even if the `.vos` file is not present? Or is that something that `make` wouldn't be able to support?

doc/sphinx/practical-tools/coq-commands.rst
charguer: I'm already in a subsection. I don't like subsubsection so much, it takes too much vertical space i find. Let me know if you think this is really important.

doc/sphinx/practical-tools/coq-commands.rst
charguer: I have update coq_makefile generator to support this, indeed.

doc/sphinx/practical-tools/coq-commands.rst
charguer: It might seem like a good idea at first, but I think it is going to raise a lot of confusion in case the user somehow screws up his Makefile and his lacking dependencies on .vos files. Then, he will have no clue what is going on when he gets errors of the form "interface foo is more recent than interface bla" in the compilation of a subsequent file. Having an explicit empty ".vos" file has some cost (it pollutes your "ls"), but it enables keeping the semantics simple.

ejgallego:
@charguer I ignore what you have exactly done but it is spectacular :) More seriously, the rebase seems seriously borked.

Zimmi48:
When people have "rebase" and "resolve conflicts" commits it usually is that they refused to actually rebase and tried (and failed) to do a merge instead.

doc/sphinx/practical-tools/coq-commands.rst
Zimmi48: It gives some structure. I don't particularly care but @cpitclaudel seemed to care.

doc/sphinx/practical-tools/coq-commands.rst
Zimmi48: Maybe it should be made explicit in the doc.

doc/sphinx/practical-tools/coq-commands.rst
Zimmi48: Indeed, I can see the point.

charguer:
I had some incomprehensible error messages. I had fixed the conflicts, but then the --continue wouldn't work, I had to do --skip, and then don't know what on earth happened; it's not by lack of good will, just lack of any experience in this rebase business.

ejgallego:
> I had some incomprehensible error messages. I had fixed the conflicts, but then the --continue wouldn't work, I had to do --skip, and then don't know what on earth happened; it's not by lack of good will, just lack of any experience in this rebase business.

I can only recommend to use `magit` or some kind of GUI always to rebase and avoid command line.

Also, enabling rerere tends to help a lot.

For fixing the mess, first identify the hash of the commits you would like to apply on top of master. Then, hard reset this branch to master, and cherry pick the identified commits one by one.

JasonGross:
>  I had fixed the conflicts, but then the --continue wouldn't work, I had to do --skip

You need to `git add` the files with conflicts that you fixed before you can --continue.  Using --skip drops the commit entirely, you don't want that in most cases.  You can use `git reflog ` to recover commits that you dropped (it will give you hashes for old commits).  You might want `git rebase -i ` if there are, in fact, commits you want to drop.     If you mess up in the middle of a rebase, you can go back with --abort.  Perhaps here you accidentally rebased onto an outdated version of master?

charguer:
I did "git add" but then still "--continue" wouldn't succeed, I searched google, and one guy said the only way out was to "--skip". Anyway, thanks for trying to help me out via the github messages, but I don't think I'll get through this without some live help hands on the terminal. 

Or, perphas simpler, is there a simple way to just create a new branch based on the contents of the diff:
https://github.com/coq/coq/compare/master...charguer:vos-files
which contains (as I write) exactly the contents of the patch that we're interested in?

SkySkimmer:
>Or, perphas simpler, is there a simple way to just create a new branch based on the contents of the diff:
master...charguer:vos-files

If you checkout the branch you want, `git reset master` (soft reset) then commit it should work.

SkySkimmer:
@charguer I reset your PRs to sane looking commits
fetch then `reset --hard origin/branch-name` to get the correct commit on your local branches

JasonGross:
> is there a simple way to just create a new branch based on the contents of the diff:

If you pipe the output of git diff to a file, you can git apply that patch to get the changes.

library/library.ml
charguer: I have found a way to do this relatively cleanly, by having the constructors for .vos an .vio directly take the list of except as argument.

charguer:
@SkySkimmer many thanks for the cleanup. I still have to finish implementing @gares suggestion. And also, do you have ideas for using the warning system, or is my proposal fine? i'll finish that next week.

SkySkimmer:
I don't agree that we want warnings to change when using -vos, but not enough to oppose merging.

charguer:
I have cleaned up the interface/code of library.ml following @gares ' suggestion.
I have added a specific test into the test suite.
I also checked that I can compile my TLC library using vos/vok files.

I think that everything should be ready for a merge. (I don't know if it'd be cleaner to collapse all my commits into a single (fresh) commit, but feel free to do so if you think it's a good idea.) 

Once the merge is done, I can take care of advertising on coq-dev/coq-club to encourage beta-testing.

Thanks to all for your valuable help!

vernac/vernacentries.ml
maximedenes: I also don't like this redundancy. Indeed, you don't have the bug related to disabling this warning, but you may have the dual with warnings as errors being ignored depending on the value of this flag.

> a warning should only be triggered if the compilation mode is "-vos", which I keep track of using "Flags.complain_on_proof_without_using_in_section"

That sounds strange. Why not check the compilation mode before emitting the warning?

maximedenes: (and not introduce a flag, then)

vernac/vernacentries.ml
```diff
@@ -886,6 +889,21 @@ let vernac_end_section {CAst.loc} =
 
 let vernac_name_sec_hyp {v=id} set = Proof_using.name_set id set
 
+let warn_proof_without_using_in_section =
+  CWarnings.create ~name:"proof-without-using-in-section" ~category:"vernacular"
+    (fun () -> strbrk "You should use the “Proof using [...].” syntax instead of “Proof.” to enable skipping this proof which is located inside a section. Give as argument to “Proof using” the list of section variables that are not needed to typecheck the statement but that are required by the proof.")
+
+let warn_arguments_assert =
```
maximedenes: Why did this appear here?

vernac/vernacentries.ml
```diff
@@ -886,6 +889,21 @@ let vernac_end_section {CAst.loc} =
 
 let vernac_name_sec_hyp {v=id} set = Proof_using.name_set id set
 
+let warn_proof_without_using_in_section =
+  CWarnings.create ~name:"proof-without-using-in-section" ~category:"vernacular"
+    (fun () -> strbrk "You should use the “Proof using [...].” syntax instead of “Proof.” to enable skipping this proof which is located inside a section. Give as argument to “Proof using” the list of section variables that are not needed to typecheck the statement but that are required by the proof.")
```
maximedenes: This line is too long, please split it.

vernac/vernacentries.ml
charguer: it's a warning related to sections. I defined it next to other functions that deal with vernac sections. Where would you like me to put it instead?

maximedenes: I meant `warn_arguments_assert`.

maximedenes:
> I think that everything should be ready for a merge.

Hi @charguer ! Indeed it looks like it's getting close, but I hope we can get rid of the ad-hoc warning-related flag, and you kept some spurious whitespace-related changes that should be removed from this PR. The current policy when you encouter mixed tabs and space is that it is ok to clean-up blocks you have to edit, but not the current file. IDEs will usually work fine.

vernac/vernacentries.ml
charguer: Throughout the file, it seems to be the convention that there is no line break in the middle of an error message formatted using str or strbrk. Please suggest the  formatting of your choice, because I wouldn't know what to go for to improve the situation.

maximedenes: Arf, we are not very good with coding conventions, sorry about that. I usually find long lines a pain to edit, so I'd do:

```ocaml
strbrk "An 80-character line where you start " ++
strbrk "the error message and then you keep " ++
strbrk "going."
```

But that can also be a pain to edit, so maybe you should just ignore this remark ;)

charguer:
@maximedenes For the warning, like I explained, I couldn't think of a better way to proceed. Help is welcome. For the spurious whitespaces, I don't know what is the best way to remove them. Should I create a separate commit just for those whitespace changes like @ejgallego suggested? Or should I create a patch with my current set of commits, and remove the whitespace changes from it? In both cases, I wouldn't know how to reflect the changes properly into this pull request.  (Perhaps we should simply call each other to fix these two remaining issues?)

SkySkimmer:
>@maximedenes For the warning, like I explained, I couldn't think of a better way to proceed. Help is welcome.

Let me reiterate that we probably shouldn't change warnings based on compilation mode.

vernac/vernacentries.ml
charguer: Somebody moved it before my rebase, and I screwed up in the merge.  Let me remove it just now.

vernac/vernacentries.ml
charguer: The warning is emitted from vernac.ml. In this file, I have no access to the compilation mode. All the information that I have is that passed to `load_vernac` and via the `Flags` module. So, I could choose to pass the information about the compilation mode in one way or the other, but I think that in the end that still need a boolean reference similar to the one that I use currently. If you confirm that you find it cleaner to pass an argument to `load_vernac`, I can make the change.

charguer (commented):
If you don't want to change warnings based on compilation mode, here are several possibilities:
- Issue a warning on all `Proof` without `using` inside sections, for every users. (That's not nice for all Coq users.)
- Don't issue any warning for `Proof` without `using` for `vos` users. (Silently, some proofs won't be skipped, not so nice for `vos` users).
- Interpret `Proof.` to mean `Proof using.` inside a section, whenever `vos` or `vok` mode is used. (A bit of a departure wrt what `vio` files used to do, but why not after all? This would impose on `vos` users to properly annotate all lemmas in sections, though.)

SkySkimmer:
>Don't issue any warning for Proof without using for vos users. (Silently, some proofs won't be skipped, not so nice for vos users).

That's a documentation issue.

charguer:
@SkySkimmer I'm not sure what you mean by "documentation" issue.

SkySkimmer:
That is, we should clearly say in the doc of `-vos` what causes proofs to be skipped/not skipped, and say that turning on warning `foo` will point them at the not skipped proofs.

charguer:
CONTRIBUTING.md reads: "Whitespace discipline (do not indent using tabs, no trailing spaces, text files end with newlines) is checked by Travis (using `git diff --check`). We ship a [`dev/tools/pre-commit`](/dev/tools/pre-commit) git hook which fixes these errors at commit time. "

Could be worth adding an explicit sentence to tell that one should not remove tabs outside of edited sections. (I believe different projects have different policies regarding polishing legacy code.)

ejgallego:
I'm not sure I do like the warning either but I am afraid that here we are talking about a serious problem of the Sections mechanism [that also impacts proof processing and other areas, see #6254] so I guess we'll have to live with that hack, provide this PR is approved.

@ggonthier mentioned once a possible way forward as to make statements in sections more robust wrt to proofs; I'd be great if he could comment here or in #6254 as to better understand how far we'd be from an improvement in that area [likely not very close indeed]

charguer:
@SkySkimmer let me try reformulate your proposal:
- warning "-report-proof-without-using" should be off by default
- all users of "-vos" are advised to use "-W +report-proof-without-using"

I think that's too dangerous, many people would just fail to benefit from "-vos" performance, and not know about it. One way out is to create yet a new warning: "you are using -vos but not  -W +report-proof-without-using".  (But isn't that overkill?)

SkySkimmer:
>I think that's too dangerous, many people would just fail to benefit from "-vos" performance

I don't think it's dangerous, how did they find out about -vos if they didn't read the manual?

SkySkimmer:
But is it actually how 99% of users will want to use it? Maybe we should 
ask on coq-club or some such thing.

Gaëtan Gilbert

ejgallego:
> But is it actually how 99% of users will want to use it? Maybe we should ask on coq-club or some such thing.

I'm not sure users are going to provide useful feedback when the feature is not even available. I have consistently seen this to happen.

We'd rather first release what we believe is sensible and then wait for feedback. A good thing about the `coq_dune` file format is that it is versioned, so we can have the flag enabled/disabled by default at `lang` version `0.x` and disable it at `0.x.+1`.

vernac/vernacentries.ml
charguer: @maximedenes I manually reverted the spaces to tabs. I did a commit with my changes. The commit hook killed my changes---for good reason, I believe. The hook called it a successful operation, and subsequently deleted the patch file. Hence, my changes are lost. Great.

Matafou:
Come on, people never read the manual to find ideas, they read it to solve
problems the are facing, generally after asking in coq-club.

People will hear about vos, try them naively and come ask in coq-club why
it does not speed proofs as expected. And we will answer "like it is said
in the manual, blablabla".

I think a single warning "Your file does not seem to be ready for vos,
don't expect a speed up until you use \"Proof using\" on every proof. See
manual section XX." would be nice. Would it be possible?

P.




Le lun. 12 nov. 2018 à 14:22, Gaëtan Gilbert <notifications@github.com> a
écrit :

charguer:
@Matafou 

You suggest  "Your file does not seem to be ready for vos, don't expect a speed up until you use \"Proof using\" on every proof." [+inside a section+].

The current warning is "You should use the “Proof using [...].” syntax instead of “Proof.” to enable skipping this proof which is located inside a section. Give as argument to “Proof using” the list of section variables that are not needed to typecheck the statement but that are required by the proof."

I think it's similar enough. I could mention "-vos" explicitly in the message if you'd like, but I felt that given  that this warning only comes up when activating -vos for the first time, and that it talks about "skipping" proofs, it should be clear that it's -vos related. 

(Btw, I agree with the claim that typical users don't read the manual. In fact, I think that a good tool would should not point users to the manual, but give them clear enough message that they may resolve the issue directly.)

charguer:
Hum, my message from yesterday about whitespaces seems to have vanished. Let me post it again:

I manually reverted spaces to tabs in the spurious changes. But when I commited these changes, the pre-commit hook just killed them all, furthermore leaving me with no backup of the changes. Hence, I have no clue how I'm supposed to contribute such an "undo" commit. 

Remark: there are only ~40 lines of spurious diff, which is quite less than the size of the patch itself. Moreover the whitespace are for the most part nearby actual changes. Given the benefit of cleaning up the file and making it a lot easier for people to read the files, would it be such a big deal to accept (just for this time) the patch as is?

SkySkimmer:
I would accept.

Gaëtan Gilbert

maximedenes:
Since I'm involved in the STM these days, I can take this PR as an assignee. Note that I believe it would be better to not make the two systems (vio files and this one) coexist, to keep complexity manageable. So if we go for this one, we probably want #8783.

charguer:
@maximedenes: if you remove support for vio files, there are a few places where my patch could be simplified. In which order would you recommend merging the two matches and performing the extra cleanup?

maximedenes:
> if you remove support for vio files, there are a few places where my patch could be simplified. In which order would you recommend merging the two matches and performing the extra cleanup?

Probably it would mean rebasing this on top of the `vio` removal, and making the extra cleanup before merging anything.

But of course, we have to make sure we all agree on the course of action.

charguer:
I have just looked at the diff of #8783, one issue is that it removes a few things that are needed for  `vos` support. At the same time, there are large amounts of removed material for vio checking that we still want to get rid of.  I think we would need to go through that diff line by line with @ejgallego in order to select which parts should be kept or not kept.  In any case, I would advise against merging #8783 before #8642.

ejgallego:
> I think we would need to go through that diff line by line with @ejgallego in order to select which parts should be kept or not kept.

Sure, let's discuss!

> In any case, I would advise against merging #8783 before #8642.

Well, of course even if we can do that it would defeat the goal of #8783 that was to measure how the features compare in order of code complexity. So even if we merge this PR first, I'd love to get a version of this PR on top of `vio`-clean Coq so we can clearly see which parts is touching and how we can optimize it.

tchajed:
Thanks for your work and patience on this @charguer!

Let me weigh in since FSCQ was a huge user of -quick: we did all of our development using quick builds, built the executable by extracting from a quick build, and compiled proofs once a day on a beefy server in a hacky CI setup that was added late in the project. Quick builds of FSCQ take about 15 minutes and full builds take 24 CPU hours, or about 6 real-time hours on our VM (which has a large number of cores and a lot of RAM).

We relied on quick builds, and it was not pleasant when they stopped working on Coq master (last I checked Coq CI never runs a quick build). I also discovered for myself by trial and error and thinking about the problem that Coq will run proofs in sections without a using statement and added some appropriate annotations, which significantly sped up our builds.

I fully agree with the argument above that users should not have to read the manual. The manual has poor discoverability and it's naive to assume users have to use the manual to discover features for a lot of reasons (a simple one being that the manual doesn't include all the features of Coq).

I like the idea of a warning that describes the problem with section variables. However, I think as with all warnings, it should offer a suggestion and _it should offer a local change to suppress the warning_. Currently, I don't think there's any simple advice for the user to follow to suppress the warning. This is especially problematic because the proof might be very short, so skipping it has no performance benefit and the warning is spurious. However, manually annotating what section variables to use has a real cost. For example, I have a library with 75 one-line proofs that takes about a second to compile; I really don't want to annotate them with section variables, nor do I want to add `Proof using.` to all of them since it would change the types.

The way I would use this feature is probably using some command-line configuration (set through a Makefile or _CoqProject), and I would like an option like `-vo` that sets a more complicatedly-named option and enables the appropriate warning. I'm a little disappointed this question wasn't asked on coq-club under the assumption that nobody would respond - wasn't it worth a try? It makes me sad every time the Coq development team discusses something without user input, even for people like myself that put an effort into participating.

Before this feature gets widely used, can we please come up with a better name than `vos`? Such an opaque name makes it much harder to understand the feature. It also looks like a plural of `vo`, which confused me for a while when reading about it.

I think compiling a `.vos` file for every build shouldn't be taken too lightly - Coq already builds `.Foo.aux`, `Foo.vo`, and `Foo.glob` and dumps them alongside the source, which is extremely untidy. LaTeX is almost as bad, except at least it supports redirecting the build output to a different directory.

gares:
> it was not pleasant when they stopped working on Coq master (last I checked Coq CI never runs a quick build).

Is it still broken? is there a bugreport? In the test-suite there are a few -quick related tests, but you are right we don't try to build with -quick everything. The main reason is that some constructs (eg Include for modules) is not well supported (it mostly works these days but not always).

SkySkimmer:
>(eg Include for modules) is not well supported (it mostly works these days but not always).

When does it fail?

gares:
@SkySkimmer you gave me the example here: https://github.com/coq/coq/pull/8787#issuecomment-438615221

SkySkimmer:
That should be fixed, at least according to the test added in #8995.

gares:
> That should be fixed

Indeed I could compile the stdlib with `-quick`. I'll prepare a PR compiling and validating the standard library using the quick compilation chain.

charguer:
@gares : once you have such a PR, I'd be interested to see whether the stdlib could also be compiled using vos/vok files.

ejgallego:
> Quick builds of FSCQ take about 15 minutes and full builds take 24 CPU hours, or about 6 real-time hours on our VM (which has a large number of cores and a lot of RAM).

@tchajed do you perform a "full build" by just disabling `-quick`, or do you go `.vio -> .vo`?

ejgallego:
Thanks for the answer @tchajed ; do you recall how much improvement the vio approach got you when compared to a regular vo build [with or without `-async-proofs` + aux metadata] ?

ejgallego:
@charguer I think that the status of the codebase towards 8.10 is reasonable for you to do a rebase and attempt a merge.

doc/sphinx/practical-tools/coq-commands.rst
gasche: (`required_vos` would make more sense.)

ejgallego:
@charguer something went wrong in the rebase, maybe you used an older ref?

This needs rebase again I'm afraid.

maximedenes:
> This needs rebase again I'm afraid.

@charguer and I had a meeting where we agreed on a few last things to do for this PRs. I'll post it here and do it ASAP. I'll take care of the rebase as well.

maximedenes:
> @charguer and I had a meeting where we agreed on a few last things to do for this PRs. I'll post it here and do it ASAP. I'll take care of the rebase as well.

Done in #9808.

ejgallego (commented):
I have many questions on this; but here is the first one:

- what kind of guarantees does the `-vok` flag provide?

As far as I can tell, it is generating a `.vo` file, however the `.vos` files will be missing some universe constraints, and IIUC, this could lead to unsoundness, right? [cc; @ppedrot ]

SkySkimmer:
Can we use some other string than `vos`? I always want to read it as the plural of `vo`.

tchajed:
@SkySkimmer yes, before it becomes too late I would really like better names for `.vos` and `.vok`. We don't need to use a single extension or limit ourselves to three letters.

I don't have an intuition for what these files mean. @charguer, if you can explain them and/or point me to documentation I'd propose new names.

Zimmi48:
Documentation's here: https://github.com/coq/coq/pull/8642/files#diff-fc6654a51ee663b14e195af01142e812
Preview is not available because the CI did not run because of the conflicts with master.

tchajed:
I guess I'd suggest `.v.defs` and `.v.checked`?

F* uses an empty `foo.fst.checked` to indicate that F* has run on `foo.fst`, and I think copying that idea is good. I kind of like `.vi` for `.v.defs` as well in analogy to `.mli` or F*'s `.fsti`, with the caveat that this is auto-generated so it's not quite the same thing.

One thing I realized is that I have no idea why the files use the extension `.v`; I guess `*.vo` is just "o" for "output", but what does the `v` stand for?

mattam82:
'.v' for vernacular and '.o' for object, I would think.

ejgallego:
> F* uses an empty `foo.fst.checked` to indicate that F* has run on `foo.fst`, and I think copying that idea is good.

I am OK with different naming, however I am not sure I am fond of using empty files as a communication device; IMO we should get proper build rules for this stuff, even if we don't implement them yet.

gares:
>     * guarantees does the `-vok` flag provide

The same as the .vio ones, some constraints are missing/not checked. It is documented in the PR.
Do you have other questions?

ejgallego:
> The same as the .vio ones, some constraints are missing/not checked. It is documented in the PR.

I read the documentation but I couldn't find that particular part, where it is? I am a bit worried then we produce `.vo` that are not "reliable". Does vio2vo have the same problem indeed?

ejgallego:
> Do you have other questions?

I have plenty but I'd rather have a rebased version.

ejgallego:
> I read the documentation but I couldn't find that particular part, where it is?

Yeah I am sorry but after a 3rd read I couldn't find any mention to this, mind pointing out where it is discussed @gares ? Thanks.

But it that's the case, then assume I have a `.vo` file, how can I know whether it was produced from a consistent set of constraints, is it marked somewhere in the lib?

gares:
> we produce `.vo` that are not "reliable". Does vio2vo have the same problem indeed?

vio2vo generates files in v format with extension vo that have all constraints in them, but the consistency of all vo together has not been checked. This is documented in the chapter about async proofs, and the countermeasure is to require all vo files generated this way (this checks the global consistency of universe). 

vok are not vo files, they are just file stamps. When a vok is generated out of a vos file (that are in vo file format) it means that the corresponding v file was fully compiled (proof included) but not its dependencies (that were loaded from the vos file). Hence the kind of guarantee they provide is the same of vio2vo: global consistency of universes is not checked. But once you have the vok stamp you don't have the corresponding vo file on disk. So, unless I'm mistaken, you can't require all vok files in order to check the consistency, something you can do with vo files generated by vio2vo.

ejgallego:
> So, unless I'm mistaken, you can't require all vok files in order to check the consistency, something you can do with vo files generated by vio2vo.

Thanks, definitively the current vok documentation should be improved to note this, but I'll wait until the rebase to see how the global doc looks [indeed they may want to link to the vio2vo part]

maximedenes:
I was hoping to get this in 8.10, but @ejgallego, you seem to say you still have blocking questions, is that right?

ejgallego:
> I was hoping to get this in 8.10, but @ejgallego, you seem to say you still have blocking questions, is that right?

I do have a few questions but not sure how can they be blocking, in any case is the answers to them that could be; to what me respects I didn't even start the review yet as I am waiting to have a current version.

ejgallego:
After a quick look of the stale version, the only real important question I have on this is how it should interact with the make system; it does not seem to me that the situation is blocking but I'd like to be sure the rules are properly documented before merge.

charguer:
> I'd like to be sure the rules are properly documented before merge.

I have just pushed additional text for the documentation. I think everything is written up now. If you see any specific item missing, please let me know, I'd be happy to complete it further.

SkySkimmer (dismissed):
The name should be changed to something which doesn't look like the plural of `vo`

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -226,6 +233,117 @@ and ``coqtop``, unless stated otherwise:
...
+the proofs that it contains.
+
+   .. warning::
+
+      Compiled interfaces should only be used for development purposes.
+      At the end of the day, one still needs to proof check all files
+      by producing standard ``.vo`` files. (Technically, when using ``-vos``,
+      fewer universe constraints are collected.)
+
+**Principle.**
```
Zimmi48: Trivial comment: I think we want these sub-parts to appear in the TOC. I.e. using a sub-section would be preferable to using bold.

gares:
> The name should be changed to something which doesn't look like the plural of `vo`

After 5 months in the pipelines, after 1 WG discussion, after the doc was written, I think a request like this one landing *one* week before the freeze is, well, inappropriate.

maximedenes:
I'll try to rebase this now.

vernac/vernacentries.ml
maximedenes: I'm still not convinced that the warning should depend on the compilation mode (and on which IDE you use, for example). Since it is more or less orthogonal to this PR, I'll remove the warning so that we can discuss it while not postponing the integration of the `.vos` file support.

maximedenes:
@charguer Rebase done, would you mind adding a changelog entry?

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -269,20 +268,52 @@ the proofs that it contains.
...
+and ``foo.vok`` also depend on ``bar.vos``.
 
-      In the current implementation, the use of the command ``Include`` for
-      importing modules compiled using `-vos` might not work properly.
+Note, however, that ``foo.vok`` does not depend on ``bar.vok``.
+Hence, as detailed further, parallel compilation of proofs is possible.
 
+In addition, ``coq_makefile`` generates for a file ``foo.v`` a target
+``foo.required_vos`` which depends on the list of ``.vos`` files that
+``foo.vos`` depends upon (excluding ``foo.vos` itself). As explained
```
maximedenes: Oh, there's a typo here, the closing tag is incorrect. Will fix.

maximedenes: Done.

charguer:
Sorry I don't have a checkout at hand in the coming days. But here is the message if you want to use it:

maximedenes:
> Sorry I don't have a checkout at hand in the coming days. But here is the message if you want to use it:

No worries, if you write here the changelog entry I can add it to the branch.

Zimmi48:
@maximedenes Arthur has edited the comment above with a message.

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -245,6 +252,117 @@ and ``coqtop``, unless stated otherwise:
...
+.. warn:: You should use the “Proof using [...].” syntax instead of “Proof.” to enable skipping this proof which is located inside a section. Give as argument to “Proof using” the list of section variables that are not needed to typecheck the statement but that are required by the proof.
+
+     If |Coq| is invoked using the ``-vos`` option, whenever it finds the
+     command ``Proof.`` inside a section, it will compile the proof, that is,
+     refuse to skip it, and it will raise a warning. To disable the warning, one
+     may pass the flag ``-w -proof-without-using-in-section``.
+
+**Interaction with standard compilation**
+
+When compiling a file ``foo.v`` using |coqc| in the standard way (i.e., without
```
Zimmi48: `|coqc|` is not defined. This is what makes the compilation of the refman fail.

maximedenes: Yes, I'll fix it and add the changelog entry.

maximedenes:
> No worries, if you write here the changelog entry I can add it to the branch.

@Zimmi48 where do you think the changelog entry belongs? "tools"? "infrastructure-and-dependencies"? somewhere else?

Zimmi48:
Arthur suggested "tools" and I agree that this is a good choice.

maximedenes:
> Arthur suggested "tools" and I agree that this is a good choice.

I was blind, sorry. Entry now added.

ejgallego (commented):
The code here seems very suspicious, for example the flag for require seems never used.

I think the problem is the same than in #9808 , some changes have been dropped in the rebase.

ejgallego:
Let's have someone test the PR. Also, I wonder if another assignee should be found [not sure how much Maxime has actually contributed] and about the list of reviewers.

maximedenes:
> The code here seems very suspicious, for example the flag for require seems never used.

Not impossible, indeed, the rebase was quite tricky with files having moved and being split. I'll have a look right after the Coq call. Are you joining btw, @ejgallego?

maximedenes:
> I wonder if another assignee should be found

I think it would be better, indeed. Maybe @gares or @ppedrot?

ejgallego:
> I think it would be better, indeed. Maybe @gares or @ppedrot?

I can also take care of it if you want; the changes in the PR look OK so far but incomplete?

maximedenes:
> the changes in the PR look OK so far but incomplete

You're right, I lost the loadpath-related changes in the process. Will add them back at the right place.

maximedenes:
> Will add them back at the right place.

Done, should be better now.

ejgallego:
I'll review in-depth before the end of the month, in the meantime it'd be great if someone could test the patch and check it works as expected as I have very little time.

maximedenes:
> I'll review in-depth before the end of the month

I thought this had already been reviewed in-depth (by you, in fact). We agreed at the WG to target 8.11 for this PR, but it would mean to get a review this week, to leave time to react.

So it is maybe a better strategy to have it reviewed by another pair of eyes, someone who'd hopefully have more time than you or me between now and the end of the month. That's why I suggested @gares or @ppedrot.

test-suite/vos/C.v
```diff
@@ -0,0 +1,15 @@
...
+
+Definition z := x + y.
+
+Lemma zeq : z = z.
+Proof. pose xeq. pose yeq. auto. Qed.
+
+Lemma yeq'' : y = y.
+Proof. apply yeq'. Admitted.
+
+(* To activate after rebase
```
gares: Shall we?

toplevel/coqcargs.mli
```diff
@@ -8,7 +8,22 @@
...
-type compilation_mode = BuildVo | BuildVio | Vio2Vo
+(** Compilation modes:
+  - BuildVo      : process statements and proofs (standard compilation),
+                   and also output an empty .vos file
+  - BuildVio     : process statements, delay proofs in futures
+  - Vio2Vo       : load delayed proofs and process them
+  - BuildVos     : process statements, and discard proofs,
+                   and load .vos files for required libraries
+  - BuildVok     : like BuildVo, but load .vos files for required libraries
+type color = [`ON | `AUTO | `EMACS | `OFF]
```
gares: `type color`? Is this an Easter egg?

toplevel/coqc.ml
```diff
@@ -30,6 +30,9 @@ coqc specific options:\
 \n                         into fi.vo\
 \n  -schedule-vio-checking j f1..fn   run up to j instances of Coq to check all\
 \n                         proofs in each fi.vio\
+\n  -vos                   process statements but ignore proofs, and produce a .vos file\
```
gares: can you write *opaque* proofs here?

stm/stm.ml
```diff
@@ -2801,13 +2801,21 @@ let handle_failure (e, info) vcs =
...
-    ldir long_f_dot_vo
-    (Global.opaque_tables ());
+  (* LATER: when create_vos is true, it could be more efficient to not allocate the futures; but for now it seems useful for synchronization of the workers,
+  below, [snapshot] gets computed even if [create_vos] is true. *)
+  let (tasks,counters) = dump_snapshot() in
+  let except = List.fold_left (fun e (r,_) ->
+     Future.UUIDSet.add r.Stateid.uuid e) Future.UUIDSet.empty tasks in
+  let todo_proofs =
+    if create_vos
+      then Library.ProofsTodoSomeEmpty except
```
gares: please rename this constructor, it sounds like `truefalse`. I don't get what except stands for, especially in the second case.

test-suite/vos/C.v
maximedenes: Oh, I had missed this. Will try.

ejgallego:
> I thought this had already been reviewed in-depth (by you, in fact). We agreed at the WG to target 8.11 for this PR, but it would mean to get a review this week, to leave time to react.

Too many things have changed since last time; also I'm naturally concerned as for example here and in #9808 critical code was left out and the test-suite / review didn't catch it at first.

I'm very afraid to play schedule chicken and ship a buggy implementation because of that; that means either a) 8.11 will be delayed when we discover problems or b) users will have a really bad experience.

It is not reasonable to ask for a review in 2 days for a PR as complex as this.

maximedenes:
> Too many things have changed since last time

Nothing has changed, it was just rebased.

> review didn't catch it at first

review did catch the missing code ;)

> It is not reasonable to ask for a review in 2 days for a PR as complex as this.

Oh, for the PR sure. But it 's been open for a year now, which is way more time. 2 days to review the rebase seems reasonable to me (if someone has some availability of course, which we can never guarantee).

gares:
I did some comments, but I need more time for the rest, hopefully tomorrow.

maximedenes:
> I did some comments, but I need more time for the rest, hopefully tomorrow.

Sure, I'll address these ones already, thanks!

stm/stm.ml
maximedenes: This is probably a question for @charguer.

test-suite/vos/C.v
maximedenes: Done. Let's see what CI says.

toplevel/coqcargs.mli
maximedenes: Fixed.

toplevel/coqc.ml
maximedenes: Done.

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -245,6 +252,117 @@ and ``coqtop``, unless stated otherwise:
...
+In addition, ``coq_makefile`` generates for a file ``foo.v`` a target
+``foo.required_vos`` which depends on the list of ``.vos`` files that
+``foo.vos`` depends upon (excluding ``foo.vos`` itself). As explained
+next, the purpose of this target is to be able to request the minimal
+working state for editing interactively the file ``foo.v``.
+
+**Typical compilation of a set of file using a build system.**
+
+Assume a file ``foo.v`` that depends on two files ``f1.v`` and ``f2.v``. The
+command ``make foo.requires_vos`` will compile ``f1.v`` and ``f2.v`` using
```
gares: Isn't this `make foo.required_vos`?

gares (approved):
modulo the constructor name and the glitch in the doc

gares:
I don't have time to test it

charguer:
I can test it tomorrow (Friday)

doc/sphinx/practical-tools/coq-commands.rst
charguer: indeed

stm/stm.ml
charguer: The variable name "except" has the same meaning as in the original code from `library.ml`, function `save_library_to`, and `kernel.ml` function `dump`. I figured it was better to reuse the same variable name for the same contents. If you want to rename it to something else, should do so throughout the codebase.

doc/sphinx/practical-tools/coq-commands.rst
charguer: i'll change it

charguer:
I have tested the patch, and fixed two bugs:
- One in coq_makefile that missed the COQLIB arguments on the command line for producing `.vos` and `.vok` target.
- One in that we forgot to enable loading `.vos` files from interactive sessions. I tentatively fixed that by adding `Flags.load_vos_libraries := true;` in the function `init_document` which appears to be specific to the creation of `Interactive` documents.

Recall that producing a `.vo` file always induces the creation of an empty `.vos` file, and that the policy when the `load_vos_library` flag is true is to search for `.vos` files upon `require`, and in case the file is empty then search for the corresponding `.vo`.

For the record, here is what I've been testing using TLC:

```
### Configure COQBIN to my path

export COQBIN=~/versions/coq-vos/coq/bin/


### Checkout TLC for Coq 8.10

git clone git@gitlab.inria.fr:charguer/tlc.git -b migration-to-coq-8.10 tlc
cd tlc


### Initialize _CoqProject and Makefile

make clean
echo "-R . TLC" > _CoqProject
~/versions/coq-vos/coq/bin/coq_makefile -f _CoqProject -o Makefile *.v
export VERBOSE=1


### Test maximally parallel compilation

# direct parallel compilation of all interfaces and files
make clean
make -j8 vok


### Test phased compilation: interface then proofs

# compile all interfaces
read -p "Press enter to continue..."
make clean
make -j8 vos

# compile all files
read -p "Press enter to continue..."
make -j8 vok


### Test incremental changes

# recompile modified interfaces of dependencies (LibList depends on LibEqual)
read -p "Press enter to continue..."
touch LibEqual.v
make -j8 LibList.required_vos

# check a single file, at a time its dependencies are ready
read -p "Press enter to continue..."
make LibList.vok


### Test interactive editing
  # (optional to restart all): make clean; make -j8 LibList.required_vos

# interactive editing of a file with its dependencies ready
read -p "Press enter to continue..."
${COQBIN}coqide LibList.v &


### Test mixing more involved scenario mixing vo files and vos files

# Here, we compile the chain up to LibEqual consists of true vo files
# (generating empty vos files). Beyond that point, we compile the chain
# for producing vos files. Then we can check a proof, producing LibList.vok.

read -p "Press enter to continue..."
make clean
make LibEqual.vo
make LibList.vok

# interactive editing of a file in that setting
read -p "Press enter to continue..."
${COQBIN}coqide LibList.v &
```

gares:
Something got broken, even in the test suite:
```
Error: Unable to locate library Prelude with prefix Coq. (While searching for a .vos file.)
```

gares:
>     * One in that we forgot to enable loading `.vos` files from interactive sessions.

The test suite uses, sometimes, coqtop or coqidetop (via fake_ide) but apparently CI does not copy over .vos to the test suite job. This is a wild guess of the bug we see, CC @coq/ci-maintainers

charguer:
I just commited a related bugfix to coq_makefile, to make sure it copies `.vos` files over.

As for the CI to copy the `.vos` files, does it use coq_makefile? If so it should be fixed now. If not, then something else needs to be patched in the CI-related files.

charguer:
@gares Thanks for your help, btw. I'm leaving for a week of vacations, so I soon won't be able to fix things further. If there are further issues, we might have to end up postponing :(

gares:
I believe the .gitlab ci file May need patches. I can have a look next week, if this is the only issue, but I can't allocate much time

gares:
This should do the trick, I hope.

gares:
OK, it worked but some tests still fail in coq-makefile/, in particular they assert which files are installed and now there are also .vos, so I've just updated the expect file list.

test-suite/Makefile
```diff
@@ -131,9 +131,10 @@ bugs: $(BUGS)
 
 clean:
 	rm -f trace .nia.cache .lia.cache output/MExtraction.out
+	rm -f vos/Makefile vos/Makefile.conf
 	$(SHOW) 'RM        <**/*.stamp> <**/*.vo> <**/*.vio> <**/*.log> <**/*.glob>'
```
SkySkimmer: We should update this

Alternatively we could stop using SHOW/HIDE since the command isn't especially noisy (unlike ocamlc commands which have tons of includes for instance)

test-suite/Makefile
```diff
@@ -748,3 +749,23 @@ tools/%.log : tools/%/run.sh
 	    $(FAIL); \
 	  fi; \
 	) > "$@"
+
+# vos/
+
+vos: vos/run.log
+
+vos/run.log: $(wildcard vos/*.sh) $(wildcard vos/*.v)
```
SkySkimmer: Shouldn't this be a misc/ test? It looks more special than it is when it lives in its own directory.

SkySkimmer: (or maybe a coq-makefile/ test)

vernac/library.ml
```diff
@@ -430,23 +430,33 @@ let error_recursively_dependent_library dir =
...
-        assert(Filename.check_suffix f ".vo");
-        Future.UUIDSet.empty
-    | Some (l,_) ->
-        assert(Filename.check_suffix f ".vio");
-        List.fold_left (fun e (r,_) -> Future.UUIDSet.add r.Stateid.uuid e)
-          Future.UUIDSet.empty l in
+type ('document,'counters) todo_proofs =
+ | ProofsTodoNone (* for .vo *)
+ | ProofsTodoSomeEmpty of Future.UUIDSet.t (* for .vos *)
+ | ProofsTodoSome of Future.UUIDSet.t * ((Future.UUID.t,'document) Stateid.request * bool) list * 'counters (* for .vio *)
```
SkySkimmer: How about using `TodoForVio` etc instead of `ProofsTodoSome` and having to explain in a comment?

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -245,6 +252,117 @@ and ``coqtop``, unless stated otherwise:
...
+correctly compiles, including all its opaque proofs. If the compilation
+succeeds, then the output is a file called ``foo.vok``, with empty contents.
+This file is only a placeholder indicating that ``foo.v`` has been successfully
+compiled. (This placeholder is useful for build systems such as ``make``.)
+
+When compiling a file ``bar.v`` that depends on ``foo.v`` (for example via
+a ``Require Foo.`` command), if the compilation command is ``coqc -vos bar.v``
+or ``coqc -vok bar.v``, then the file ``foo.vos`` gets loaded (instead of
+``foo.vo``). A special case is if file ``foo.vos`` exists and has empty
+contents, and ``foo.vo`` exists, then ``foo.vo`` is loaded.
```
SkySkimmer: coqtop and coqide behaviour should also be documented.

Also coqc in default mode producing the empty .vos should be documented.

SkySkimmer:
What's the plan for dune vs .vos? I pushed a commit to get the stdlib to work (ie stop dune from deleting the empty .vos produced by coqc). 

I guess we also will need to teach the coq module of dune itself about .vos so that it works for 3rd party developments.

This only gets normal full compilation to work, I don't know how hard it is to teach dune that 2 commands can produce .vos files (and unlike make we can't just forget to tell dune that default coqc produces a .vos).

gares:
Thanks for testing this with dune! I don't know dune enough to give an answer to your questions.

ejgallego:
> What's the plan for dune vs .vos?

Indeed last time I looked the situation was not optimal, `dune` does really want to be able to compute statically the targets a command call will produce so it can handle the dep graph more easily [you can have dynamic commands but that complicates things a lot]

I need to understand better what is produced and when to be able to say more.

ejgallego:
> Recall that producing a `.vo` file always induces the creation of an empty `.vos` file, and that the policy when the `load_vos_library` flag is true is to search for `.vos` files upon `require`, and in case the file is empty then search for the corresponding `.vo`.

These kind of hacks do indeed seem pretty undesirable to me. IIUC, now the dependency graph does depend on the _contents_ of a file, which itself is a build product.

It is not hard to see how potential build parallelism is reduced here as the scheduler won't have access to a static dep graph but instead must wait on a lot of commands to execute, unless of course you add a hack itself to scheduler to assume a dep cut-off.

gares:
I don't understand the argument about parallelism, but the first point about the empty file can be fixed by writing in the .vos file the md5sum of the .vo (the ml code need then to check the "magic number" rather than the size of the file in order to understand if it is just a time stamp). Would this ease supporting this in dune?

gares:
@maximedenes I think the rebase glitches were fixed, together with a few minor issues. It is now in your hands, or in the RM ones (CC @ppedrot @mattam82 ) since it was decided to get this in.

ppedrot:
@maximedenes is in vacations, right? This PR was definitely decided to be part of 8.11, so if the assignee doesn't react before the merge deadline, I'll take it over instead.

maximedenes:
> @maximedenes is in vacations, right?

Yep, feel free to take over now @ppedrot. Thanks!

ejgallego:
> I don't understand the argument about parallelism

The point is simple, the less of the dependency graph the build system knows, the less it can schedule. Given that this PR introduces a dynamic component it can decrease it.

> but the first point about the empty file can be fixed by writing in the .vos file the md5sum of the .vo (the ml code need then to check the "magic number" rather than the size of the file in order to understand if it is just a time stamp). Would this ease supporting this in dune?

The problem is not a problem of dune support, which can work, but about the general design of the system. Checking the md5 or the size is the same, the real problem is that a file `a.v` that requires `b`, will depend first on `b.vos` and only depending on the content of such file the dependency on `b.vo` will exist. This seems quite hacky to me.

> EDIT: I meant, if a.v requires b.v, then a.vo depends on b.vo and b.vos (and both outputs change when b.v does, if we write the hash in the .vos). So to me the graph is still static.

Is that correct? This is not how I understand it; if `b.vos` exists and it is not empty then the compilation of `a.v` only depends on `b.vos`, right? [at least in interactive mode for example]

> This PR was definitely decided to be part of 8.11, so if the assignee doesn't react before the merge deadline, I'll take it over instead.

I guess what was decided is that the PR would be part of 8.11 as long as it is ready. I'd suggest not to rush things and risk creating a mess that has to be solved later. For example there are a few comments by Gaëtan that need addressing.

If the PR is so important and it is late it can be backported as we have done for quite a few others in the past.

toplevel/coqtop.ml
```diff
@@ -271,6 +271,8 @@ let init_document opts =
      state before we take the first snapshot. This was not
      guaranteed in the past, but now is thanks to the STM API.
   *)
+  (* Next line allows loading .vos files when in interactive mode *)
+  Flags.load_vos_libraries := true;
```
ejgallego: This should rather be a command line flag, instead of the default.

gares:
> This is not how I understand it; if `b.vos` exists and it is not empty then the compilation of `a.v` only depends on `b.vos`, right?

Hum, I guess we need to reason with @charguer .
My impression is that there are two targets (classic and -vos), and in each case the dependencies are clear and statically determined. I'm expecting that dune should be able to at least do the two builds independently, as if they were two compilers.
The patch makes it so that the -vos chain can opportunistically use up to date .vo files. This may work with make, and may not work with dune (the optimization may not be an option in dune if the two builds are disjoint). If I'm not mistaken, I think the design is acceptable wrt dune.

charguer:
Sorry for latency and short answer, being out of office.

Either you are in real checking mode and you produce only vo files. The dependencies are the same as before, because vo files don't depend on Vos. 

Or you are in dev mode, and you work with vos and vok. There is no dependency on vo files, so you produce none. 

If there are vo files left around while on dev mode, then :
- if you use make and the vo file has been produced together with an up to date empty vos file, you can load the vo file.
- if you use dune you might need to either first clear empty vos files or teach dune how to deal with them (I lack expertise here, but clearing empty vos is certainly scriptable, and just ignoring the existence of the leftover vos files might even work). 

Recall that the hack of producing empty vos is meant to avoid libraries to export both vo and vos. 

I might have overlooked something. But please provide me with a specific counter example based on a practical scenario., so that I can think about it. Thanks

ppedrot:
Are the experts convinced by this explication, and can I proceed to the merge? Otherwise we'll have to postpone to 8.12...

doc/changelog/08-tools/08642-vos-files.rst
```diff
@@ -0,0 +1,7 @@
+- `coqc` now provides the ability to generate compiled interfaces.
+  Use `coqc -vos foo.v` to skip all opaque proofs during the
+  compilation of `foo.v`, and output a file called `foo.vos`.
+  This feature enables working on a Coq file without the need to
+  first compile the proofs contained in its dependencies
```
Zimmi48: The changelog entry should contain a warning that this is an `*experimental*` feature.

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -245,6 +252,117 @@ and ``coqtop``, unless stated otherwise:
...
+Compiled interfaces help saving time while developing Coq formalizations,
+by compiling the formal statements exported by a library independently of
+the proofs that it contains.
+
+   .. warning::
+
+      Compiled interfaces should only be used for development purposes.
+      At the end of the day, one still needs to proof check all files
+      by producing standard ``.vo`` files. (Technically, when using ``-vos``,
+      fewer universe constraints are collected.)
```
Zimmi48: Please add a second warning that this is an experimental feature.

Zimmi48 (commented):
If we decide to release this as an experimental feature, let's mark it experimental. Not like the attribute feature that we never marked experimental despite having claimed many times that we would.

ppedrot:
@Zimmi48 if @charguer does not react in time, can you push a minimal patch tweaking the documentation as you asked atop of this PR?

ppedrot:
Gaah, this needs a rebase now...

Zimmi48:
> @Zimmi48 if @charguer does not react in time, can you push a minimal patch tweaking the documentation as you asked atop of this PR?

Can someone else do it? I can't make any guarantee on my availability today (the quirks of having the branching date be set on a holiday).

ppedrot:
I'll do it, even though I am supposed to merge it after.

ppedrot:
@Zimmi48 done.

ppedrot:
CI is going to fail on fiat-crypto, let's merge before something gets in the way.

charguer:
Whoever was involved in this thread: please read my post describing a fix at https://github.com/coq/coq/issues/11057

Issue #8609, opened 2018-10-01
STM: sending back a fat state


silene:
If we ever decide to be less noisy about fat states, here is a small script to test it, as it causes the warning to appear systematically.

```coq
Goal Prop.
Proof.
set (a := Prop).
Qed.
```

gares:
I totally agree: #8643

gares:
This is the check https://github.com/coq/coq/blob/master/stm/stm.ml#L1587
This is the safe_env comparison that fails https://github.com/coq/coq/blob/master/stm/stm.ml#L921

PR #8643, opened 2018-10-03
[stm] remove warning about fat state

gares:
Fix: #8609

### DIFF
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -1587,7 +1587,6 @@ end = struct (* {{{ *)
         when is_tac expr && State.same_env o n -> (* A pure tactic *)
           Some (id, `ProofOnly (prev, State.proof_part_of_frozen n))
       | Some _, Some s ->
-          msg_debug (Pp.str "STM: sending back a fat state");
           Some (id, `Full s)
       | _, Some s -> Some (id, `Full s) in
     let rec aux seen = function


ejgallego (commented):
Why is a debug message a problem?

silene:
> Why is a debug message a problem?

Because it is displayed prominently by Coqide. Maybe that is a bug in Coqide, but that it is the only debug message that floods the user, as far as I know.

silene (approved):
If you consider that fat states are not an issue in the general case, then the commit is obviously fine.

gares:
A worker sends back a fat state when there is a change in the env (IIRC using == to check if there was one).

Two possibilities: the env started to be unshared for no deep reasons during tactic execution (it used to be the case for nested lemmas, or abstract, IIRC, dunno for broken proofs as the test case). The other one is that the warning used to be discarded and is not displayed.

In both cases it is pointless to show it to the final user. In the first case, something has to be investigated.

gares:
I made this PR just to let the RM choose. IMO silencing is fine. Finding the root cause is better. In any case, the only difference is performance (instead of the evar map, it sends back the env + evar map...).

silene:
So, #8609 is worth keeping open, right? Because I just can't see how the environment was changed there.

gares:
I think so, but note that there the proof is incomplete so maybe that plays too. I'll try to investigate

gares:
I digged a bit more, it is the `==` on safe env that fails (as if someone has added constants to it)... I need to dig more.

gares:
See #8712

PR #8644, opened 2018-10-03
[build] enable warnings on kernel/% in make based builds

gares:
This is to sync with the dune build system that sets these warnings

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -685,6 +685,8 @@ plugins/nsatz/%.cmi: plugins/nsatz/%.mli
 	$(SHOW)'OCAMLC    $<'
 	$(HIDE)$(OCAMLC) $(COND_BYTEFLAGS) -package unix,num -c $<
 
+kernel/%.cmi: COND_BYTEFLAGS+=-w +a-4-44-50
+
 %.cmi: %.mli
 	$(SHOW)'OCAMLC    $<'
 	$(HIDE)$(OCAMLC) $(COND_BYTEFLAGS) -c $<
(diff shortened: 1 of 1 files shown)


ejgallego (commented):
I don't understand well how the makefile system works, I'd say the change looks OK.

gares:
bedrock2 failure is OOM, I guess not related. Restarted.

gares:
I made this PR as a remedy to an issue @maximedenes reported to me that is due to a commit by @ejgallego (extra warning in the kernel are checked, via dune).

I can't assign this to myself. I guess one of you guys should merge it. All lights are green.

maximedenes (approved):

SkySkimmer:
Why not do like `COND_PRINTERFLAGS`? ie
~~~make
COND_KERNELFLAGS = $(if $(filter kernel/%,$<), -w ...,)

COND_BYTEFLAGS = ... $(COND_KERNELFLAGS)
COND_OPTFLAGS = ... $(COND_KERNELFLAGS)
~~~

gares:
Let's turn it around, what makes you think you solution is better? I've two arguments to support the current one:

- when you start to have 10 components with different flags you end up with a line
    ```
    COND_BYTEFLAGS = $(COND_C1) .... $(COND_Cn)
    ```
    with the invariant that one variable contains the right set of warning and all the others are empty (wrt warnings). 
While the current code looks more like tagging files by a pattern and setting some variables if match.
- I'm using the language feature designed for doing exactly what we are trying to do, eg add a few flags when a rule hits a specific subset of the targets.

It looks like you proved that cut is an admissible rule, and you are using this as an argument to make people no use it in their proofs ;-)

SkySkimmer:
>when you start to have 10 components with different flags you end up with a line 

I'll accept this answer.

PR #8650, opened 2018-10-04
[ci] [dune] [opam] Fixes to OPAM and CI target.

ejgallego:
The Dune `release` profile is used by OPAM so that should cover the
testing.

### DIFF
--- a/coq.opam
+++ b/coq.opam
@@ -6,13 +6,13 @@ bug-reports: "https://github.com/coq/coq/issues"
 dev-repo: "https://github.com/coq/coq.git"
 license: "LGPL-2.1"
 
-available: [ ocaml-version >= "4.02.3" ]
+available: [ ocaml-version >= "4.05.0" ]
 
 depends: [
-  "dune"      { build }
+  "dune"      { build & >= "1.2.0" }
   "ocamlfind" { build }
   "num"
-  "camlp5"
+  "camlp5"    {         >= "7.03"  }
 ]
 
 build-env: [
--- a/ide/coqide.opam
+++ b/ide/coqide.opam
@@ -6,14 +6,16 @@ bug-reports: "https://github.com/coq/coq/issues"
 dev-repo: "https://github.com/coq/coq.git"
 license: "LGPL-2.1"
 
-available: [ocaml-version >= "4.02.3"]
+available: [ocaml-version >= "4.05.0"]
 
 depends: [
-  "dune"      { build }
-  "ocamlfind" { build }
-  "num"
-  "camlp5"
+  "dune"               { build & >= "1.2.0" }
   "coq"
+  "conf-gtksourceview"
+  "lablgtk"            {         >= "2.18.5" }
 ]
 
+build-env: [
+  [ COQ_CONFIGURE_PREFIX = "%{prefix}" ]
+]
 build: [ [ "dune" "build" "-p" name "-j" jobs ] ]
--- a/ide/dune-workspace
+++ b/ide/dune-workspace
@@ -1,0 +1,6 @@
+(lang dune 1.2)
+
+; Add custom flags here. Default developer profile is `dev`
+(env
+ (dev     (flags :standard -rectypes -w -9-27-50+60))
+ (release (flags :standard -rectypes)))
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,7 +9,7 @@ stages:
 variables:
   # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
   # for reference]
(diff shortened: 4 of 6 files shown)


coq.opam
```diff
@@ -6,13 +6,13 @@ bug-reports: "https://github.com/coq/coq/issues"
...
-available: [ ocaml-version >= "4.02.3" ]
+available: [ ocaml-version >= "4.05.0" ]
 
 depends: [
-  "dune"      { build }
+  "dune"      { build & >= 1.2.0 }
   "ocamlfind" { build }
   "num"
-  "camlp5"
+  "camlp5"    {         >= 7.03  }
```
SkySkimmer: Aren't we accepting 7.01?

SkySkimmer (changes requested):

coq.opam
ejgallego: IIRC 7.02 has a critical bug making Coq fail so I'd rather play on the safe side here.

SkySkimmer (approved):
OK then

coq.opam
ejgallego: Or indeed maybe it was 7.02 the one with the bug, hard to tell....

ejgallego: Maybe the problem was with `Bytes`. Umm, I think I'm gonna up the version to 7.02 indeed.

ejgallego: Ok, I've settled in 7.02. I think that 7.01 had troubles working with 4.06.1 in the context of Coq, so IMHO that should be fine; we can keep 7.01 + 4.05.0 in the CI, but for OPAM it will save us headaches to move to 7.02.

ppedrot (commented):
The build-related files in `ide/` shouldn't be owned by me, I am utterly incompetent.

coq.opam
```diff
@@ -6,13 +6,13 @@ bug-reports: "https://github.com/coq/coq/issues"
...
-available: [ ocaml-version >= "4.02.3" ]
+available: [ ocaml-version >= "4.05.0" ]
 
 depends: [
-  "dune"      { build }
+  "dune"      { build & >= "1.2.0" }
   "ocamlfind" { build }
   "num"
-  "camlp5"
+  "camlp5"    {         >= "7.02"  }
```
Zimmi48: I don't like the indentation but I won't fight about it.

ide/dune-workspace
```diff
@@ -0,0 +1,6 @@
+(lang dune 1.2)
+
+; Add custom flags here. Default developer profile is `dev`
+(env
+ (dev     (flags :standard -rectypes -w -9-27-50+60))
+ (release (flags :standard -rectypes)))
```
Zimmi48: Isn't it inherited from `dune-workspace` if absent?

ide/coqide.opam
```diff
@@ -6,14 +6,19 @@ bug-reports: "https://github.com/coq/coq/issues"
...
 depends: [
-  "dune"      { build }
-  "ocamlfind" { build }
+  "dune"               { build & >= "1.2.0" }
+  "ocamlfind"          { build }
   "num"
-  "camlp5"
+  "camlp5"             {         >= "7.02"  }
+  "conf-gtksourceview"
+  "lablgtk"            {         >= "2.18.5" }
```
Zimmi48: This is inconsistent with what is documented in `INSTALL`.

Zimmi48:
What about replacing `build:edge:dune` with what you put in `pkg:opam`? Is the only additional thing that's tested the "edge" aspect of it?

ide/coqide.opam
ejgallego: Fixed, see new changes.

ide/dune-workspace
ejgallego: Not if `-p coqide` is used [release, isolated build]. Note that there is a problem indeed with how `-rectypes` are handled, it should not be necessary that we add them here.

I believe this problem is in both `ocamlfind` and Dune. I propose we implemented this while I discuss the `-rectypes` issue upstream.

ejgallego: @Zimmi48 so I had some discussion with Dune upstream and indeed they suggested a better layout that should avoid this redundancy problem. Will update once we converge I guess in a subsequent PR.

ejgallego:
Ok, I pushed some changed; as noted in the comment above we are discussing with upstream how to better handle a few things, this however should produce and test working OPAM packages so I recommend merging once CI passes.

Zimmi48 (approved):

SkySkimmer:
I'm going to assume the bedrock and lambdarust failures are unrelated and merge.

ejgallego:
> I'm going to assume the bedrock and lambdarust failures are unrelated and merge.

Yes they seem so, merci.

PR #8651, opened 2018-10-04
[dune] Provide an optimized build profile with inlining reports.

ejgallego:
This satisfies a wish by @ppedrot


Zimmi48 (commented):
From the Dune POV, it looks good, but I'll let @ppedrot test, comment and merge this.

ejgallego:
Failure is spurious, given the nature of this PR I won't push another build.

Makefile.dune
```diff
@@ -41,6 +42,9 @@ apidoc: voboot
 ocheck: voboot
 	dune build $(DUNEOPT) @install --workspace=dev/dune-workspace.all
 
+ireport: voboot
+	dune build $(DUNEOPT) @install --profile=ireport
```
rgrinberg: It might be worth mentioning that subsequent build commands (even in this profile) will regenerate the reports from scratch. Since dune doesn't know those files are targets, it will aggressively clean them. This behavior is usually fine, as it's nice to not to have to worry about stale targets of deleted rules, but here it would be annoying.

ejgallego: Thanks for the comment @rgrinberg but I am not sure I understood. So even if `foo.ml` does not need an update, its report will be deleted. That's fine; however, how does it then get built again?

rgrinberg: Here's another way of explaining it:

dune will delete any files in `_build/` that it does not consider as targets of the current rules on every invocation.

ejgallego: So indeed if we just build twice with no source change, the reports won't be regenerated but just deleted, right?

rgrinberg: I think so

ejgallego: Ok, thanks for the explanation. So that indeed deserved a Dune issue to update the built-in targets then.

ejgallego:
Updating to reflect and workaround a strong caveat this method has as of today.

ejgallego:
Note the CI failures are spurious. Given the nature of this PR I don't think a rerun is needed.

ppedrot (approved):
Looks like it's working.

PR #8656, opened 2018-10-04
Test Windows addons on v8.9

MSoegtropIMC:
(for the time being), this is just a test to check if the Windows addons available in master build fine on v8.9. This is not a trivial question since v8.9 picks different versions for the addons in ci-basic-overlays.sh.

This corresponds to PR #8322 on master, PR #8403 on v8.8 and issue #8321.

### DIFF
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -1647,7 +1647,7 @@ function make_addon_equations {
   installer_addon_dependency equations
   if build_prep_overlay Equations; then
     installer_addon_section equations "Equations" "Coq plugin for defining functions by equations" ""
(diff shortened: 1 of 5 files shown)


MSoegtropIMC:
@Zimmi48 , @silene : the Windows addon stuff went through smoothly on 8.9. You can either use the PR (which is based on v8.9) or you can merge #8322 from master. The only thing I changed from master is the comment and tag for coquelicot in dev/ci/ci-basic-overlay.sh.

Please note that this PR as is will include all supported addons in CI. This will increase build time. I will now prepare the PR for master to build extra addons only optionally. In released versions all addons should be in, but for CI this is not required.

Zimmi48:
@silene has backported #8322 in 8.9+beta1, so this PR can be closed.

PR #8657, opened 2018-10-04
[dune] Refactor files following advice from upstream.

ejgallego:
After some discussion upstream, I think that for the current a
multi-package setup in a single repository this setup saves a bit
work.

The most problematic bit is the `-rectypes` flag; its status is not
saved per-library so we must either duplicate the flags in the coqide
scope (scope == dune-project file), or unify its scope with Coq. We do
this last option as this seems like the easiest thing for now.

Changes:

- Move coqide.opam file to the root, use a single scope/`dune-project` file.
- Remove `dune-workspace`, this should be owned by the developer / user.


Zimmi48 (approved):
This makes a lot of sense to me but I am a bit worried that this isn't tested. What about just merging this with #8650?

ejgallego:
I guess we can just merge #8650 which is ready and the I can rebase this one as to run less CI, but I can indeed squash #8650 into this one too.

Zimmi48:
The strategy you propose seems OK.

ejgallego:
> Does the change require opam2?

Why?

gares:
IIRC ` name.opam` only works on opam 2

ejgallego:
I have no idea, I am not an expert and I've never understood how pin works, neither it seems to be documented enough.

Zimmi48:
I heard indeed that this model is only supported with OPAM 2. But this shouldn't be blocking, right?
I don't really understand well how the dev packages that we distribute in the OPAM archive work so it might be worth checking that we are not breaking them again. @RalfJung you are the only user of this package that I know, so could you confirm that this PR is OK? (BTW, I suppose you are still on OPAM 1.2, is this correct?)

ejgallego:
Can you folks point at the change you are referring to? Where did you see a variation on semantics?

gares:
Why did you move ide/coqide.opam in . and why do you now pin . ?

ejgallego:
> Why did you move ide/coqide.opam in . and why do you now pin . ?

This is a very good question. Dune has the notion of "scope", which determines which ones are individual projects on a composed tree. For compatibility reasons with `jbuilder`, Dune uses the existence of an `foo.opam` file to mark the directory as having a new scope. Nowadays this is done using the `dune-project` file. Each scope has its own set of flags.

Anyways, the story is long and we even discovered a bug in Dune ( https://github.com/ocaml/dune/pull/1408 ) , but for now, we think that the best strategy is to make the whole Coq repository a single unit [with several packages] versus before where the repos was the composition of two projects. This means that both Coq and CoqIDE share the same set of flags etc.. so we avoid having to repeat the flags twice.

Anyways, it turns out that some new features/changes will happen in Dune upstream so in the future we may want to put the IDE back on its own scope. But for now I think this is the solution that makes the most sense.

ejgallego:
> I heard indeed that this model is only supported with OPAM 2. But this shouldn't be blocking, right?

As far as I understand this should not be blocking at all indeed; in fact it is likely that the current solution in master is even less compatible with OPAM 1.2.

Anyways, I haven't been able to locate the specification of `opam pin` anywhere, so it is hard to know whether we are conforming or not.

ejgallego:
BTW I am not sure @RalfJung was in OPAM 1.2, I recall him having some OPAM 2.0 issues, but of course he can tell.

RalfJung:
I am usually pinning against the `coq.dev` package, which should not be affected at all by what you are doing here (except that opam 1.2 had a bug here, but opam 2 fixed that bug). This becomes relevant when one pins against the git URL directly, and then indeed I find it hard to predict where it gets the opam file from.

I don't know enough about opam to say of this change is correct or not.

Zimmi48:
Thanks for the clarification @RalfJung! So it seems indeed that supporting dev packages for opam 1.2 is likely going to be broken all the time due to this weirdness of opam 1.2 so I suppose we can just forget about it. @gares Do you mind?

SkySkimmer (approved):
CI change is trivial.
Who is assignee here?

Zimmi48:
Just waiting for @gares's green light before merging.

gares (approved):

ejgallego:
Merci!

PR #8660, opened 2018-10-05
Issue 8659 not always build all addons

### LINKED ISSUE #8659: Not always build extended set of addons for WIndows installer
PR #8322 added several addons to the Windows64 CI build, which substantially increases build time. It is not required to test all addons on every run. In a discussion in coqdev it was agreed to keep those which have been active before PR #8322 in master and just add mtac-2 for PR CI. The full set will be tested in a regular test, but not for each PR.

MSoegtropIMC:
This PR fixes #8659 Not always build extended set of addons for Windows installer.

I already added the variable WINDOWS_ALL_ADDONS to gitlab but didn't add CI runs with this enabled nor did I add machinery to enable it for release builds.

I though about using the same mechanism as in makecoq_mingw.sh to decide if the open source archive should be build to check if this is a release build, but I think using the WINDOWS_ALL_ADDONS variable this can be done more properly (and possibly a variable to build the open source archive can be added at the same time, so that I can remove this hack).

### DIFF
--- a/dev/ci/README.md
+++ b/dev/ci/README.md
@@ -175,13 +175,13 @@ If your repository has access to runners tagged `windows`, setting the
 secret variable `WINDOWS` to `enabled` will add jobs building Windows
 versions of Coq (32bit and 64bit).
 
+If the secret variable `WINDOWS` is set to `enabled_all_addons`,
+an extended set of addons will be added to the Windows installer.
+This leads to a considerable runtime in CI so this is not enabled
+by default for pipelines for pull requests.
+
 The Windows jobs are enabled on Coq's repository, where pipelines for
 pull requests run.
-
-If the additional secret variable `WINDOWS_ALL_ADDONS` is set to `enabled`,
-an extended set of addons will be added to the Windows installer.
-This leads to a considerable runtime in CI and is disabled by default for
-pipelines for pull requests.
 
 ### GitLab and Docker
 
--- a/dev/ci/README.md
+++ b/dev/ci/README.md
@@ -178,6 +178,11 @@ versions of Coq (32bit and 64bit).
 The Windows jobs are enabled on Coq's repository, where pipelines for
 pull requests run.
 
+If the additional secret variable `WINDOWS_ALL_ADDONS` is set to `enabled`,
+an extended set of addons will be added to the Windows installer.
+This leads to a considerable runtime in CI and is disabled by default for
+pipelines for pull requests.
+
 ### GitLab and Docker
 
 System and opam packages are installed in a Docker image. The image is
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -186,7 +186,7 @@ after_script:
     - call dev/ci/gitlab.bat
   only:
     variables:
-      - $WINDOWS == "enabled"
+      - $WINDOWS =~ /enabled/
 
 build:base:
   <<: *build-template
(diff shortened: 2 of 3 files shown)


dev/ci/README.md
```diff
@@ -178,6 +178,11 @@ versions of Coq (32bit and 64bit).
 The Windows jobs are enabled on Coq's repository, where pipelines for
 pull requests run.
 
+If the additional secret variable `WINDOWS_ALL_ADDONS` is set to `enabled`,
+an extended set of addons will be added to the Windows installer.
+This leads to a considerable runtime in CI and is disabled by default for
+pipelines for pull requests.
+
```
ejgallego: Two tags for a ternary switch may become problematic, I'd suggest you reuse the existing `WINDOWS` variable here.

MSoegtropIMC: @ejgallego : I am not sure if this would get easier, because both switches are possibly controlled in different places and it might get tricky to not overwrite one setting with another. But I don't know how the WINDOWS switch is really used, so I can't really tell if this would be a problem or not. If the point of control would be the same, for sure I can change this.

MSoegtropIMC: I mean where and how are these variables set? The default value in the gitlab settings might depend on environemnts and also on specific branch properties (protected branches ...). As far as I can tell if something is a signed release build or not is controlled here. But I guess most settings are controlled in coqbot. Not sure if there are other entities which control these variables.

One might want to control WINDOWS_ALL_ADDONS at least partly in a similar way as the release signign key variable. But I guess WINDOWS is always controlled by coqbot. You see my point?

ejgallego: I am not sure I follow, `WINDOWS` is set as a regular variable, AFAICT `coqbot` has nothing to do with it. It makes sense to me that the `WINDOWS` variable becomes `::= enabled | enabled_with_addons` etc...

MSoegtropIMC: I guess my concerns are a result of a lack of understanding of details on how these variables are set.  I will change it then.

ejgallego: TTOMK the variables are set from the GitLab CI panel, but ccing @Zimmi48 and @SkySkimmer to be more sure about that.

ejgallego: Anyways my comment is just a suggestion, we can use two variables if we think it is better [but I tend to dislike this 3 settings 2 binary variables because you find situations such as `WINDOWS=disabled` but `WINDOWS_ALL_ADDONS=enabled`.

In fact, maybe the variable should just be `WINDOWS_ADDONS=list_of_addons` so people could customize the build easily.

SkySkimmer: >TTOMK

What?

>the variables are set from the GitLab CI panel

yes

ejgallego: > TTOMK

To The Best of My Knowledge I meant ...

MSoegtropIMC: @ejgallego : I want to have the small and large set defined somewhere in GIT, because changes to this are not trivial. Usually changes to the set go together with other changes in GIT. Decoupling this would not be good.

It is by the way easy to do a local build with a custom set of addons (e.g. for teachers).

MSoegtropIMC: @ejgallego : I just changed it. Unfortunately gitlab doesn't allow boolean expressions in the .yml file, so I had to use regexp for the enable test. Please review if you really like this better.

Zimmi48 (approved):
Why is there still a progress tag?

Zimmi48:
Maybe this would deserve an update to `dev/doc/release-process.md` to remind about setting this variable in tag builds (although this doc needs a number of other updates).

Zimmi48:
What remains to be done (post-merge) is to create a pipeline schedule to run with all addons enabled: https://gitlab.com/coq/coq/pipeline_schedules

MSoegtropIMC:
@Zimmi48 : I removed the progress tag (sorry, thought I did so before)

PR #8661, opened 2018-10-05
CI: fix Iris and stdpp ref selection

SkySkimmer:
This was broken due to a typo when introducing the archive download
method.

We also remove default [master] values from basic-overlay which hid
this issue.

### DIFF
--- a/dev/ci/ci-basic-overlay.sh
+++ b/dev/ci/ci-basic-overlay.sh
@@ -49,11 +49,12 @@
 ########################################################################
 # Iris
 ########################################################################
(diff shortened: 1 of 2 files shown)


ejgallego (approved):
LGTM, merci.

silene:
8.9 does not use `master` for `stdpp_CI_REF` and `Iris_CI_REF`. Is this pull request meaningful for 8.9?

Zimmi48:
@silene Indeed, this PR is not useful for 8.9 because of the pinning I did.

PR #8662, opened 2018-10-05
Fix printing of raw terms for lambda's without parameters (continued, adding smart constr_expr binder constructors)

herbelin:
**Kind:** between fix and cleanup

This is a follow-up of #8490. I tried to make the API of `constr_expr` constructor a bit more uniform. I added "smart" versions for `CLambdaN` and `CProdN`.

@ejgallego: I finally did not call them `smart` but instead followed the existing naming pattern of having the `C` at the end. So, naming is less explicit but more consistent with the existing. If I had named them `mkSmartCLambdaN`, or even `mkSmartLambdaCN`, I feel I would have had to rename also `mkAppC` into `mkSmartAppC`.

### DIFF
--- a/interp/constrexpr_ops.ml
+++ b/interp/constrexpr_ops.ml
@@ -526,6 +526,14 @@ let mkAppC (f,l) =
   | CApp (g,l') -> CAst.make @@ CApp (g, l' @ l)
   | _           -> CAst.make @@ CApp ((None, f), l)
 
+let mkProdCN ?loc bll c =
+  if bll = [] then c else
+  CAst.make ?loc @@ CProdN (bll,c)
+
+let mkLambdaCN ?loc bll c =
+  if bll = [] then c else
+  CAst.make ?loc @@ CLambdaN (bll,c)
+
 let mkCProdN ?loc bll c =
   CAst.make ?loc @@ CProdN (bll,c)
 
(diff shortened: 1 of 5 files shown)


SkySkimmer:
What is the point of having the non-smart functions?

ejgallego:
> @ejgallego: I finally did not call them `smart` but instead followed the existing naming pattern of having the `C` at the end. So, naming is less explicit but more consistent with the existing. If I had named them `mkSmartCLambdaN`, or even `mkSmartLambdaCN`, 

@herbelin as you prefer, what does the `N` stand for? Maybe it would useful to explain. The "Smart" version has a small naming advantage IMHO [less prone to confusion, the name makes clear that there is some trick under the hood]

> I feel I would have had to rename also `mkAppC` into `mkSmartAppC`.

You could have done that and deprecate `mkAppC`. Anyways, regarding the naming I am OK with whichever scheme you prefer.

Taking this as `interp` 2nd, shall merge soon unless someone wants to discuss.

herbelin:
> What is the point of having the non-smart functions?

I think none, except to strictly follow a principle of compatibility.

herbelin:
> what does the N stand for?

`N` is like in `Constrexpr.CLambdaN`. It means n-ary binder. If to be documented, it should be in `constexpr.ml`, I guess.

ejgallego (approved):
LGTM, if @herbelin wants to further tweak the name as suggested he's welcome, otherwise merge is scheduled for the next couple of days.

ejgallego:
> `N` is like in `Constrexpr.CLambdaN`. It means n-ary binder. If to be documented, it should be in `constexpr.ml`, I guess.

@herbelin let me know if you wanna do another pass or you want this in as-is.

herbelin:
> let me know if you wanna do another pass or you want this in as-is.

I think it is ok. Other changes would need bigger restructuration. So, it is optimal in my eyes.

PR #6698, opened 2018-02-05
Use record of Printing options, issue #6560

psteckler:
This PR is an attempt to have `Set Printing All` work as the composition of Printing options, rather than working as its own option. That way, `Set Printing All` followed by, for example, `Unset Printing Implicit` means that implicits won't be printed.

The printing options are stored in a single globally-accessible record, rather than as refs scattered throughout the code, as before. The behavior of `Unset Printing All` changes: it negates the options of `Set Printing All`. Before, the `Unset` changed the `raw_print` flag. To compensate for that a bit, there's a new command, `Set Printing Defaults`, which restores the startup options.

`Set Printing All` also prints universes, which it didn't before.

`Set Printing Universes` also affected the printing of universe inconsistency explanations, which is somewhat different to printing universes in terms, so the PR adds the command `Set Printing Universe Explanations`, on by default. Maybe the explanations should always be printed, with no option involved.

One test-suite file, `output/Cases.v` and the corresponding `.out` file are changed a bit, because of the printing of universes.

This PR does not go as far as the request of @ejgallego to pass the record of printing options. That could be added, if there's agreement that the idea in this PR is OK. There were some places where it wasn't clear how the `raw_print` flag was representing other printing options.

An alternative to this PR would be to make clear in the documentation that `Set Printing All` is its own separate option, and in the code, make clear which options the `raw_print` flag represents. For example, instead of looking at `raw_print` directly when printing implicits, there could be a predicate `do_print_implicits`, which is the `or` of `raw_print` and the ref for the `Printing Implicit` option. If that cleanup is preferred, there are places in the code where the refs for particular options are aliased, which seems prone to confusion. Even if this PR is not adopted, it would still might be good to gather all the printing option refs in one place.

**Kind:** feature

Fixes #6560 .

- [x] Corresponding documentation was added / updated.
- [x] Entry added in CHANGES.

Issue #6560, opened 2018-01-06
Unset Printing Implicit should always disable printing of implicits


JasonGross:
#### Version

8.7.1

#### Description of the problem

If I `Set Printing All`, later doing `Unset Printing Implicit` should make implicit arguments disappear.  Worse, there are some things that seem to be unprintable without `Set Printing All` (in the sense that no other printing option changes them), so there's no way to print what I want to see without having to wade through things that I don't want to see.

psteckler:
The manual says:

> Set Printing All includes the effects of the commands Set Printing Implicit, Set Printing Coercions, Set Printing Synth, Unset Printing Projections and Unset Printing Notations.

Is it a bit odd that it does some `Set`s and some `Unset`s?  Why does it not include `Set Printing Universes`?

The description seems to leave open the possibility of other, unspecified effects. Which other things do you see that don't have their own option?

psteckler:
The operation of `Set Printing All` is a bit curious. It sets `Flags.raw_print`, rather than the flags associated with the other mentioned print options. That flag is examined directly at places in the code, so its operation can't be expressed as just the combination of other options.

Would it make more sense to have the `All` option work on those other flags, which would make its operation just the composition of those flags?

JasonGross:
> Would it make more sense to have the All option work on those other flags, which would make its operation just the composition of those flags?

Yes, I think this is better.  And anything that currently doesn't have a special flag should get one.

psteckler:
OK, it's a bit tricky. In many cases, the `raw_print` flag is used as a surrogate for the other flags in a boolean expression along with the other flags, so it can be just removed.

Sometimes, it's used on its own:

  https://github.com/coq/coq/blob/master/interp/constrextern.ml#L395

so it's not clear which flags it's shadowing (or if it's some other unspecified behavior).

Here's another tricky bit:

  https://github.com/coq/coq/blob/master/toplevel/coqtop.ml#L848

where the `raw_print` is temporarily set to `true`, some printing is done, then the old value is restored. With the modified setup, to do that correctly, you'd have to store and reset each of the component flags that `raw_print` represents. That can be done, of course, but it makes a grotty special case in the code.

SkySkimmer:
We could gather printing flags in a record.
ie
```ocaml
(* flags.mli *)
type printing_flags = { print_universes : bool; print_implicits : bool; ... }
val printing_flags : printing_flags ref
```
Then we could easily save and restore the whole value.
[...]

PR #8666, opened 2018-10-06
[printing] Collect Printing Options in a record (c.f. #6560)

ejgallego:
This PR was originally written by Paul Steckler (#6698), and updated,
rebased, and modified by me and others.

The main changes are the placement of all printing options in a
centralized record, thus the printing state is accessible in a more
convenient place. See CHANGES for printing semantic changes.

The most important change is that `Set Printing All` is not an ad-hoc
operation anymore, but just setting all options to "display".

There remain some important TODOs, apart from code review:

- Update the reference manual, diff in `pending_refman_update.diff`.
- Check IDE changes by an expert.

Co-authored-by: Paul Steckler <steck@stecksoft.com>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -192,6 +192,21 @@ Funind
 
 - Inductive types declared by Funind will never be template polymorphic.
 
+Printing
+
+- Printing All is no longer an option as such. Instead, the command
+  Set Printing All sets or unsets other Printing options to remove high-level
+  printing features. The new command Set Printing Sugared exactly reverses
+  those options, providing high-level printing features. Unset Printing All
+  restores the printing options in effect at the time of the last Set
+  Printing All.
+- The command Set Printing Defaults restores the default printing options.
+- The command Set Printing Let Binder Types prints the types of binders
+  for let-in expressions, disabled by default. Such printing had been
+  enabled by Set Printing All.
+- The Printing Records option affects the printing of record types, as well
+  as the instances of such types.
+
 Misc
 
 - Option "Typeclasses Axioms Are Instances" is deprecated. Use Declare Instance for axioms which should be instances.
--- a/dev/top_printers.ml
+++ b/dev/top_printers.ml
@@ -23,9 +23,14 @@ open Constr
 open Genarg
 open Clenv
 
-let _ = Detyping.print_evar_arguments := true
-let _ = Detyping.print_universes := true
-let _ = Goptions.set_bool_option_value ["Printing";"Matching"] false
+let _ = Printoptions.(
+    set
+      { (get ()) with
+        existential_instances = true;
+        universes = true;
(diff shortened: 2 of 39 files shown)


ejgallego:
This is a preliminary rebase, I shall do a larguish update when I have a free minute, but feel free to review/comment.

ejgallego:
It seems the rule in `g_vernac.mlg` doesn't work and `Set Printing All` is not recognized by the parser. I copied this rule from the original PR. @herbelin @ppedrot , what would be the right thing to do here?

In fact, why do we need to parse this specially, we should just set the handler properly I think.

herbelin:
> It seems the rule in g_vernac.mlg doesn't work and Set Printing All is not recognized by the parser. I copied this rule from the original PR. @herbelin @ppedrot , what would be the right thing to do here?

That's strange. Just looking at the code, I don't see what the problem could be.

> In fact, why do we need to parse this specially, we should just set the handler properly I think.

Yes, this is also a possibility.

SkySkimmer:
Even if the mlg code can be made to work I feel it would be cleaner to handle these special options in vernacentries, or even to declare them as regular options (with error on unset for Printing Sugared and Printing Defaults).

ejgallego:
Ok, trying to handle all options with the standard mechanism, let's see how this goes.

herbelin:
> Isn't there Hugo, as Gaëtan points out, an overlap?

Camlp5 was treating these overlaps sequentially. I don't know what `coqpp` does. Iirc, there are a couple of such overlaps in the parser, so, it should be checked that `coqpp` treats them the same.

ejgallego:
> Camlp5 was treating these overlaps sequentially. I don't know what `coqpp` does. Iirc, there are a couple of such overlaps in the parser, so, it should be checked that `coqpp` treats them the same.

Oh, indeed I think you are correct; there likely is a semantic discrepancy here cc: @ppedrot

ejgallego:
> Ok, trying to handle all options with the standard mechanism, let's see how this goes.

Bah, this interacts badly with the summary of course [that was Paul motivation] , as there is a race among the commands that try to modify group of options and commands that do modify one.

Will try to fix later in the week. Maybe it is time to introduce the notion of group of options in `GOptions` as to handle this case properly.

ejgallego:
There is also a problem in the original PR I think, which is a double registration with the `Summary`

It is not correct to register both the record and the option value in `Summary`, as they may not be in sync anymore.

ejgallego:
@herbelin I think the `Printoptions.t` record would be a great example to become part of `VernacState.t` and thus passed functionally if you want to try.

ejgallego:
I tried to reopen as a Draft PR, but it is not possible. This PR is not abandoned by the way, but it needs some work on the option handling to complete it.

Issue #7698, opened 2018-06-04
Improve plugin loader as to allow plugins that link to external dependencies.


ejgallego:
As we have discussed some other time (see #683), the current situation of linking libraries to plugins seems to be a bit dangerous. In particular, the problematic scenario is:

- plugins `A`, `B`, both depend on library `foo`, however `foo` is not linked into `coqtop`.

The correct behavior here is that Coq should indeed load `foo` before loading `A`, thus `B` can reuse this copy of `B`. Linking `foo` into `A` and `B`, which was the approach advocated until now will create problems and stop working in future OCaml versions starting with 4.08.

Implementing a plugin loader in Coq is not trivial, however thanks to work done in `findlib`, `Dune`, and `Frama-C`, we can use the `Findlib_dynload` API to achieve this without too much effort.

See the following links for more context:
- https://discuss.ocaml.org/t/dune-problems-using-dynlink-plugins/2874
- https://github.com/ocaml/dune/issues/1544
- http://projects.camlcity.org/projects/dl/findlib-1.8.0/doc/ref-html/lib/Fl_dynload.html
- https://dune.readthedocs.io/en/latest/advanced-topics.html?highlight=dynlink#dynamic-loading-of-packages

ejgallego:
@gares wrote in #8667 :

> My position is that it is OCaml that should implement dynamic linking properly. What it currently does is "delayed static linking" that is not at all the same thing. I'm against writing a loader inside coqtop (assuming it is possible). Given that, as of today, the scenario described in #7698 is not real, but just hypothetic, I would ignore the issue.

I'd like to add that this issue is not hypothetical even in today Coq's. In fact, the `micromega` plugin depends on `num`, and due to Coq not having its own loader we have to statically link `num` to the coqtop binary as to avoid the problem mentioned above.

So indeed that's a serious issue IMHO.

gares:
You could also make a num_plugin.cmxs out of num.cmxa, have a num.v Declare ML Module "num_plugin" and have extraction.v / nsatz.v Require num.
In this way the chain of dependencies is going to load things once and in the correct order.

What I'm against is to write code in coqtop that guesses that exaction_plugins.cmxs needs the Num module and that loads some num.cmxs on the fly (that would be a loader).

Having the dependency chain of cmxs re-use the one of .vo is boring, but way simpler. I can't tell if it is a complete solution or not.

ejgallego:
[...]

PR #8667, opened 2018-10-06
[RFC] Vendoring of Camlp5

ejgallego:
This is a proposal to vendor `camlp5` as `coq.gramlib`. The idea originated with @ppedrot , as we really don't use a lot of the features Camlp5 provides. This allows to drop the dependency from Camlp5 in order to run Coq, thus testing in newer OCaml versions would become easier. Also, we would be certainly able to optimize the integration with Coq.

The vendoring is complete, and it is possible to build Coq without camlp5 as long as you can bootstrap the `ml4` files. Once #8161 is fully implemented, Camlp5 won't be necessary at all [modulo `cLexer.ml4` and `coqide_main.ml4`]

For now, the regular make-based system uses `camlp5`, and the Dune build uses the vendored one.


herbelin:
> Also, we would be certainly able to optimize the integration with Coq.

Yes, that would then probably be easier to hack the parser to solve a few problems which bother us, like being able to continue to parse `3+-2` even if a new token `+-` is introduced in another orthogonal entry, or like removing this tolerance to parse `a < b = c` even though `<` and `=` are supposed to be non-associative, or a limitation we have with recursive notations `x sep .. sep y` when the `sep` is a list of symbols.

On the other side, we shall better watch regularly about updates of these files on camlp5 repository.

ejgallego:
@herbelin I think the base files have not changed a lot in the last few years, but I am not an expert here. Anyways, this PR is for the experts on parsing to debate. But certainly by doing this we are free from many constraints that camlp5 has to have due to its wider scope.

Despite the failure in the opam package, which is due to `promote` not being run in release mode, this PR is basically ready for merge now. Next steps would be to add a testing job for the `m̶a̶s̶t̶e̶r̶` `trunk` version of OCaml.

ppedrot:
> I think the base files have not changed a lot in the last few years

Indeed, camlp5 is pretty stable. Majority of the changes come from the necessity to adapt it to newer versions of OCaml, due to the magic of bootstrap.

> this PR is basically ready for merge now

I don't have anything against that, but it could still be pondered whether we should provide our minimal grammar library as a separate repo, there might be alternate users out there. Also, note that the current PR is probably a mid-term death penalty for camlp5, as Coq is one of the last big users of this library.

ppedrot:
Forgot to mention: we need more than #8161, as ARGUMENT EXTEND hasn't been ported yet. Also, the use of camlp5 in CLexer and Coqide is minimal, we could get rid of it with minimal hassle.

ppedrot (approved):
Here is my blessing for this PR.

ejgallego:
> I don't have anything against that, but it could still be pondered whether we should provide our minimal grammar library as a separate repo, there might be alternate users out there.

We can provide a gramlib package and release separately, but IMHO we want to keep things under the same repository. Having multiple repositories is a consistency hassle, specially when one strong reason to do this would be to allow to do some cleanups and improvements for Coq's use. But on the topic of users see below:

> Also, note that the current PR is probably a mid-term death penalty for camlp5, as Coq is one of the last big users of this library.

Indeed the motivation for this PR comes because I got tired of updating `camlp5` in OPAM, etc... For quite a time I've been alone in maintaining it, which is not surprising as it turns out that other than Coq, only 3 packages in OPAM depend on it https://opam.ocaml.org/packages/camlp5/

So indeed it seems there won't be any large campl5 user left in OPAM if we move.

ejgallego:
Another question is if we should switch the `make`-based system to vendoring? I guess so. So maybe we should add a test to be sure we still work with non-vendored camlp5 for a while?

herbelin:
> as Coq is one of the last big users of this library.

[Hol-Light](https://www.cl.cam.ac.uk/~jrh13/hol-light/) is using camlp5, and I'd bet there are more users than the ones we can see in our surrounding.

ejgallego:
Will more the 2 latest commits testing the ml4 build without camlp5 to new PR, as they may not be necessary if @ppedrot implements `VERNAC EXTEND`.

2 questions do remain:
- should we move the main make system to use the vendored lib?
- who should be the owner(s) of `gramlib`

ejgallego:
> I can volunteer if nobody does. Or we could set it to a new dedicated team or something.

I guess it makes sense to have a `parsing` team that integrates `coqpp`, `gramlib`, `parser`, and even maybe the AST / notations? WDYT @herbelin ?

ejgallego:
> That would be great. Unluckily I don't understand anything to make files so I can't help.

The main question here is if we want to keep compatibility with `camlp5` or not, API-wise. That is, if we want to support switching back.

Tho, Dune is almost ready so it may make sense to wait a bit.

ejgallego:
Regading camlp5, yeah, PPX basically removed a large part of the need for it; thus in Debian only hol-light and `ledit` [written by Daniel] depend on camlp5. `camlp4` has quite few more users tho, but it is on its way to be replaced by ppx in quite a few core users.

Regarding the revised syntax, Reason has taken that space, pity because I liked the pattern matching bracketing.

herbelin:
> I guess it makes sense to have a parsing team that integrates coqpp, gramlib, parser, and even maybe the AST / notations? WDYT @herbelin ?

Yes, you can put me in if you want.

ejgallego:
So how should we call it, `frontend-maintainers` ?

gares:
I've a problem with this patch: Coq_makefile says not to link camlp5.gramlib https://github.com/coq/coq/blob/master/tools/CoqMakefile.in#L100 since Coq already links it. Now, as of today we ship exactly the same files, hence things work by miracle, but I think the copy of camlp5 we ship should be put in a different namespace, so that the line in question can be removed from CoqMakefile and if someone happens to depend on camlp5 it just links it with no module name conflicts.

I'm planning to drop camlp5 in elpi, but I'm not ready yet. Of course elpi cannot depend on coq.gramlib (coq-elpi could, but elpi alone can't depend on Coq). So it will depend on camlp5 for some time. I this it is reasonable to let one link camlp5 proper to Coq (inside a plugin) without breaking linking safety.

Zimmi48:
@gares 
>Of course elpi cannot depend on coq.gramlib (coq-elpi could, but elpi alone can't depend on Coq).

This was I think the point of @ppedrot when he said that we can make this an independent package, thus depending on this does not mean depending on Coq (in fact it's going to be a dependency of Coq, and it can be a dependency of Elpi too).

@ejgallego replied:
>We can provide a gramlib package and release separately, but IMHO we want to keep things under the same repository. Having multiple repositories is a consistency hassle, specially when one strong reason to do this would be to allow to do some cleanups and improvements for Coq's use.

This is just defending the mono-repo / multi-packages view which I am also leaning towards lately: given that Coq will be the main user of this library, it makes sense to keep the source code near. But gramlib can still have its own OPAM package. Regarding the question of separating the release cycles, this is difficult as it could mean more hassle. Anyways, gramlib won't likely evolve enough between two Coq releases to justify releasing more frequently, and OTOH it wouldn't make sense releasing less frequently unless gramlib hasn't evolved at all between two Coq releases.

On other topics:
- If the plan is still to become Dune-only before 8.10, then it makes sense to make this change Dune-only. If not, then it makes sense to backport it to the main Makefile.
- frontend-maintainers is way too vague (it sounds like it includes UI). I recommend sticking to parsing-maintainers.

gares:
@Zimmi48 what I propose (rename the files) is orthogonal to the status of `coq.gramlib`. 

Be it maintained in Coq or outside Coq, the OCaml module names conflicts with camlp5, so it won't be possible to link both camlp5.gramlib and coq.gramlib (it is a limitation of OCaml) in the same application.
This may be a necessity (example of elpi).

I frankly don't care much how coq.gramlib is maintained, in both cases I wont to get rid of the extensible parser of elpi. I considered it acceptable to use camlp5 in the first place just because my target was Coq and Matita, and both already use it. FTR it was not me that wrote the parser of elpi ;-)

ejgallego:
> I've a problem with this patch: Coq_makefile says not to link camlp5.gramlib https://github.com/coq/coq/blob/master/tools/CoqMakefile.in#L100 since Coq already links it. Now, as of today we ship exactly the same files, hence things work by miracle, but I think the copy of camlp5 we ship should be put in a different namespace, so that the line in question can be removed from CoqMakefile and if someone happens to depend on camlp5 it just links it with no module name conflicts.

This is not by miracle I think, a "vendored" library is supposed to be a straight-up copy of the original one, so using the original one should work. Indeed we have also discussed _forking_ Camlp5 to our own parsing lib. For that to happen, vendoring would be indeed a first step, but I suggest we remain interface-compatible for now. Note tho that the CI is not run against this change [yet].

Regarding linking and plugins the situation is more complicated; issue #7698 has some more discussion but indeed plugins do work now by miracle IIUTC.

> I'm planning to drop camlp5 in elpi, but I'm not ready yet. Of course elpi cannot depend on coq.gramlib (coq-elpi could, but elpi alone can't depend on Coq). So it will depend on camlp5 for some time. I this it is reasonable to let one link camlp5 proper to Coq (inside a plugin) without breaking linking safety.

I agree, I was thinking of addressing the namespace issues on the whole tree [should we indeed put everything under Coq.Kernel, etc...] but we can wrap `gramlib` now for Dune-based builds; however, in the make case, it is more complicated.

> This was I think the point of @ppedrot when he said that we can make this an independent package, thus depending on this does not mean depending on Coq (in fact it's going to be a dependency of Coq, and it can be a dependency of Elpi too).

I'd bet that Elpi depends on far more stuff than what the vendored part does provide. I am not sure the subset we are shipping here will be of independent interest to many people, in fact, I think the current set of users is 0 [all of them need more than what we package here]

> * If the plan is still to become Dune-only before 8.10, then it makes sense to make this change Dune-only. If not, then it makes sense to backport it to the main Makefile.

I indeed suggest we wait a bit to merge this change.

> * frontend-maintainers is way too vague (it sounds like it includes UI). I recommend sticking to parsing-maintainers.

Well, `parsing is also not very precise as here we are talking about more than parsing. I dunno.

ejgallego:
> it is a limitation of OCaml

it would be great if someone could indeed clarify what is the status of the questions raised in #7698

ppedrot:
Be aware that end-users can still use camlp5 as a preprocessor, because it is not linked to the resulting code. So it's probably not a big deal...

gares:
My position is that it is OCaml that should implement dynamic linking properly. What it currently does is "delayed static linking" that is not at all the same thing. I'm against writing a loader inside coqtop (assuming it is possible). Given that, as of today, the scenario described in #7698 is not real, but just hypothetic, I would ignore the issue.

ejgallego:
> the scenario described in #7698 is not real, but just hypothetic, I would ignore the issue.

It is not hypothetic but very real. What is described there is the reason SerAPI couldn't be implemented as a plugin for example. Also, two plugins cannot use the same `ppx_deriving` or run into this problem, so it is just a matter of time. For example that prevented me from doing Talia's AST plugin with ppx.

gares:
> What is described there is the reason SerAPI couldn't be implemented as a plugin for example

It talks about A B and foo. It is not concrete. Can you be more explicit about which piece of coq serapi links and other plugins link?

ejgallego:
> It talks about A B and foo. It is not concrete. Can you be more explicit about which piece of coq serapi links and other plugins link?

In the plugin design SerAPI was designed as a set of plugins, so you had the `serlib` plugin, linking `sexplib` and the `serapi` [or serproto] plugin, linking to it too.

ejgallego:
Ok, `Gramlib` is now wrapped in the Dune build. This should avoid linking problems with `camlp5.gramlib`. Tho indeed it would be interesting to know if Elpi could use Coq's gramlib .

rgrinberg:
> Ok, Gramlib is now wrapped in the Dune build. This should avoid linking problems with camlp5.gramlib. Tho indeed it would be interesting to know if Elpi could use Coq's gramlib .

Just an FYI, there's a proposal in dune to help making this wrapping automatic:

https://github.com/ocaml/dune/issues/1016

As currently, you can do this easily only for unwrapped libraries and you still have to add `-open` everywhere. Of course, this kind of thing will only work for pure OCaml libraries. We're collecting feedback about vendoring support, so this is a useful data point for us.

ejgallego:
Actually that would be awesome @rgrinberg ; indeed, Coq may become a heavy user of vendored libs in some scenarios (for instance plugins) so that would help a lot.

gares:
> Elpi could use Coq's gramlib

No, it makes no sense to make elpi depend on coq. Coq-elpi does, but elpi alone is not related to Coq.

If gramlib was an external project it could, almost. We additionally use the `GRAMMAR EXTEND` and `parser` syntax extensions I believe. So elpi would also need coqpp to be a separate package, and eventually bootstrap the lexer.

In any case, don't think about Elpi specifically. I'm happy with the packing. This gives me the freedom to stay with or drop camlp5 when I want.

ejgallego:
> No, it makes no sense to make elpi depend on coq.

I agree with you, Coq's gramlib is going to be so minimal that won't likely be useful to others.

Note however that we can release `gramlib` as a fully independent package even if it lives in the Coq's repos.

ppedrot:
So, what's the conclusion? Should we merge this already?

ejgallego:
> So, what's the conclusion? Should we merge this already?

Only the Dune build supports that so there is some additional discussion to do:
- We could keep the make build using `camlp5` and the Dune build `coq.gramlib`. This means we'd be tied to maintain API-level compatibility among the two.
- We can move the make build to `gramlib`, however, we'd need to modify the makefile system as it doesn't support packing for that module out of the box.

ppedrot:
> We can move the make build to gramlib, however, we'd need to modify the makefile system as it doesn't support packing for that module out of the box.

I'd theoretically favour this option, but my makefile skills are not up to the task...

Zimmi48:
>If gramlib was an external project it could, almost.

What do you call an external project if there is a package that is fully independent of Coq? Do you mean that it should have a separate issue tracker? That you must be able to clone gramlib without getting the sources of Coq at the same time?

ejgallego:
An "external" or "vendored" project in the Coq tree is one that doesn't install libraries in the `coq.*` namespace, and optionally, that doesn't depend on any `coq.*` lib. So indeed `coq.gramlib` is not external but we could easily move it to `gramlib` if we add an opam file.

Remember that opam files will go soon away as a method to declare packages and instead we will add `(package ...)` declarations to `dune-project` files.

gares:
> What do you call an external project if there is a package that is fully independent of Coq?

elpi is an opam package, in the depends field I now have camlp5. I cannot put coq in place of camlp5, it would make no sense.
I could put, say, minicamlp6, a spin off of camlp5. So yes, if the files we are vendoring in this PR were a separate package (as in OPAM, but also as in github repo I guess) I could consider it.

But as I was saying, I hope to get rid of the extensible grammar altogether. So don't invest time in setting up this external project just for elpi.

Zimmi48:
>if the files we are vendoring in this PR were a separate package (as in OPAM, but also as in github repo I guess) I could consider it.

My point was precisely that OPAM (especially with Dune) supports doing a separate package (completely separate, no dependency on Coq) out of the same GitHub repo. But yeah, we are having more of a theoretical discussion here since the use case is not so clear.

ejgallego:
I consider this ready now @ppedrot , I think that for a start we can assume our vendored library to be really vendored, that is, compatible with the upstream `camlp5`. Would you like to be the assignee?

Removing camlp5 indeed does belong to a separate PR / discussion IMO. This is already very useful in itself as it allows to build without camlp5 for those interested using Dune.

Upcoming PR will add an OCaml `trunk` CI target.

ppedrot:
This is a highly non-trivial PR. Let's wait from all devs before merging. Ideally we could even discuss it next week at the WG.

ejgallego:
> This is a highly non-trivial PR. Let's wait from all devs before merging. Ideally we could even discuss it next week at the WG.

Well it doesn't touch the regular build path at all. But indeed let's wait.

Zimmi48:
>Upcoming PR will add an OCaml trunk CI target.

OK. But as "allow fail".

ejgallego:
> OK. But as "allow fail".

It is already set this way [for obvious reasons] see: https://gitlab.com/ejgallego/coq/-/jobs/110252857

maximedenes (approved):

ejgallego:
I understand we can remove the `needs: discussion` flag and merge. As noted before, this PR doesn't modify the standard build path, but it allows to use the vendored library in the Dune build.

ejgallego:
Files should get a codeowner maybe, even if for the moment we are bound to remain 100% compatible with camlp5's gramlib.

ppedrot:
@ejgallego You still need to add the license file to the gramlib folder before merging.

ejgallego:
> @ejgallego You still need to add the license file to the gramlib folder before merging.

OK, will do like that, but note that the license file of Coq has indeed problems, as Coq already contains many components licensed diferently. It should rather provide an overview [I use the Debian machine format myself] and point to the proper license bodies.

ejgallego:
Done, license added.

ppedrot:
I don't know who this really belongs to, but maybe asking @roglo for a relicensing would be an even better solution.

ejgallego:
> I don't know who this really belongs to, but maybe asking @roglo for a relicensing would be an even better solution.

Well, camlp5 license is very permissive so indeed it can already be distributed as LGPL 2.1 without a problem. I am unsure about relicensing, but if that's not possible it should be very close.

ppedrot:
En avant Guingamp!

PR #8668, opened 2018-10-07
Missing headers in two files

herbelin:
**Kind:** code cleanup/fix

This is straightforward PR adding missing headers in two files.


Zimmi48 (approved):
I don't suppose that code ownership matters so much here.

PR #8670, opened 2018-10-08
Document the issue with positive coinductive types.

ppedrot:
This is a subset of #7536, that adds the minimal amount of documentation to warn the user that the current implementation of positive coinductive types is wrong. As discussed in the aforementioned PR, we should ideally rewrite the whole documentation about primitive projections, but this will take way to much time to reach 8.9. The current PR is therefore a hotfix more than anything else, so that it can easily be backported to 8.9.

### DIFF
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -1112,6 +1112,59 @@ co-inductive definitions are also allowed.
    object of type :g:`(EqSt s1 s2)`. We will see how to introduce infinite
    objects in Section :ref:`cofixpoint`.
 
+Caveat
+++++++
+
+The ability to define co-inductive types by constructors, hereafter called
+*positive co-inductive types*, is known to break subject reduction. The story is
+a bit long: this is due to dependent pattern-matching which implies
+propositional η-equality, which itself would require full η-conversion for
+subject reduction to hold, but full η-conversion is not acceptable as it would
+make type-checking undecidable.
+
+Since the introduction of primitive records in Coq 8.5, an alternative
+presentation is available, called *negative co-inductive types*. This consists
+in defining a co-inductive type as a primitive record type through its
+projections. Such a technique is akin to the *co-pattern* style that can be
+found in e.g. Agda, and preserves subject reduction.
+
+The above example can be rewritten in the following way.
+
+.. coqtop:: all
+
+   Set Primitive Projections.
+   CoInductive Stream : Set := Seq { hd : nat; tl : Stream }.
+   CoInductive EqSt (s1 s2: Stream) : Prop := eqst {
+     eqst_hd : hd s1 = hd s2;
+     eqst_tl : EqSt (tl s1) (tl s2);
+   }.
+
+Some properties that hold over positive streams are lost when going to the
+negative presentation, typically when they imply equality over streams.
+For instance, propositional η-equality is lost when going to the negative
+presentation. It is nonetheless logically consistent to recover it through an
+axiom.
+
+.. coqtop:: all
+
+   Axiom Stream_eta : forall s: Stream, s = cons (hs s) (tl s).
+
+More generally, as in the case of positive coinductive types, it is consistent
+to further identify extensional equality of coinductive types with propositional
+equality:
+
+.. coqtop:: all
+
+   Axiom Stream_ext : forall (s1 s2: Stream), EqSt s1 s2 -> s1 = s2.
+
+As of Coq 8.9, it is now advised to use negative co-inductive types rather than
+their positive counterparts.
+
+.. seealso::
+   :ref:`primitive_projections` for more information about negative
+   records and primitive projections.
+
+
 Definition of recursive functions
 ---------------------------------
 


Zimmi48 (commented):
You should add a link to the section on primitive projections. Apart from this, this LGTM. (Although I would feel better if there was an issue opened about the need to refactor this part of the doc, and this issue had an assignee -- or at the minimum a help wanted label.)

herbelin:
I would rather tell that we don't know how to provide dependent case analysis for co-inductive types at the same time as decidable type checking without breaking subject reduction. That the problem is that dependent case analysis for coinductive types allows to prove a bit of η-expansion but η-equality is not decidable for recursive types in general (whether they are positive or negative, inductive or co-inductive). That a way to avoid the issue it to use negatively presented co-inductive types, which provide a form of dependent case analysis restricted enough not to derive η-equality. No?

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1112,6 +1112,50 @@ co-inductive definitions are also allowed.
...
+
+.. coqtop:: all
+
+   Axiom Stream_ext : forall (s1 s2: Stream), EqSt s1 s2 -> s1 = s2.
+
+As of Coq 8.9, it is now advised to use negative co-inductive types rather than
+their positive counterparts.
+
+See :ref:`primitive_projections` for more information about negative records and
+primitive projections.
```
Zimmi48: Since you put the link there, at the bottom, I recommend using a `.. seealso::` block.

ppedrot:
> I would rather tell that we don't know how to provide dependent case analysis for co-inductive types at the same time as decidable type checking without breaking subject reduction. 

This is probably a bit too much in-depth for this kind of remark. I really just wanted to signal that there is a known problem, without dwelving in the details. More explanation should be given when we rewrite the documentation.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1112,6 +1112,50 @@ co-inductive definitions are also allowed.
    object of type :g:`(EqSt s1 s2)`. We will see how to introduce infinite
    objects in Section :ref:`cofixpoint`.
 
+Caveat
+++++++
+
+The ability to define co-inductive types by constructors, hereafter called
+*positive co-inductive types*, is known to break subject reduction. It is a
+fundamental issue related to the duality of computational behavior between
```
Zimmi48: I guess that it is the "fundamental issue" sentence that @herbelin would like you to state differently...

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1112,6 +1112,50 @@ co-inductive definitions are also allowed.
...
+
+.. coqtop:: all
+
+   Axiom Stream_ext : forall (s1 s2: Stream), EqSt s1 s2 -> s1 = s2.
+
+As of Coq 8.9, it is now advised to use negative co-inductive types rather than
+their positive counterparts.
+
+.. seealso:: :ref:`primitive_projections` for more information about negative
+records and primitive projections.
```
Zimmi48: You most likely need some indentation before this second line.

herbelin:
@ppedrot: then I propose the following phrasing of the first paragraph:

"The ability to define co-inductive types by constructors, hereafter called
*positive co-inductive types*, is known to break subject reduction. The story is a bit long: this is due to dependent pattern-matching which implies propositional η-equality, which itself would require full η-conversion for subject reduction to hold, but full η-conversion is not acceptable as it would make type-checking undecidable."

And the following for the fourth paragraph:

"As indicated above, propositional η-equality is lost when going to the
negative presentation. It is possible to recover it through an axiom:

.. coqtop:: all
   Axiom Stream_eta : forall s: Stream, s = cons (hs s) (tl s).

More generally, as in the case of positive coinductive types, it is consistent to further identify extensional equality of coinductive types with propositional equality:

.. coqtop:: all
   Axiom Stream_ext : forall (s1 s2: Stream), EqSt s1 s2 -> s1 = s2.
"

As for the last sentence, independently of this PR, let me insist that one day, it should made clearer whether "primitive projections" means "with optimized representation of projections", or "formulated as records", or "defined by destructors". (Incidentally, I'm not convinced anymore that the distinction "defined by constructors"/"defined by destructors" is so important in an intuitionistic setting. It might be convenient to provide both dependent pattern-matching and native projections on one-constructor types.)

ppedrot:
@herbelin I integrated your proposal modulo a few changes. Should be ready now.

herbelin:
@ppedrot: I still don't understand your sentence "Essentially, negative streams behave similarly to functions and are very intensional in nature.". I don't understand in which way they would be more "intensional in nature" than positively-defined streams.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1112,6 +1112,60 @@ co-inductive definitions are also allowed.
...
 
+Caveat
+++++++
+
+The ability to define co-inductive types by constructors, hereafter called
+*positive co-inductive types*, is known to break subject reduction. The story is
+a bit long: this is due to dependent pattern-matching which implies
+propositional η-equality, which itself would require full η-conversion for
+subject reduction to hold, but full η-conversion is not acceptable as it would
+make type-checking undecidable."
```
gares: There is a trailing "

gares (approved):

gares:
Very clear explanation. It would be nice to add pointers to the literature, if any, especially for the consistency of the axioms you propose.

Zimmi48:
@ppedrot Are you intending to do any further change to this PR?

ppedrot:
> I don't understand in which way they would be more "intensional in nature" than positively-defined streams.

Less propositional equalities hold.

> Are you intending to do any further change to this PR?

If I manage to convince @herbelin about the phrasing, none. Otherwise I'll just remove the offending sentence.

Zimmi48:
What about the trailing `"` noticed by Enrico?

herbelin:
> > I don't understand in which way they would be more "intensional in nature" than positively-defined streams.
> 
> Less propositional equalities hold.

Yes, but these "less propositional equalities" is just eta. So, my conclusion would be more that the (traditional) version of positive inductive types with unrestricted case analysis is not the right one and the right one is the restricted version of your papers with Nicolas and @mattam82. In particular, if we are talking about the "nature" of positive types, the "nature" of positive inductive types is to have eta-extensional only optionally, exactly like the "nature" of negative types.

Zimmi48:
@ppedrot @herbelin What should be the conclusion here? Let's not wait too long before merging this. There must be a way to formulate this that is consistent with both of your views.

ppedrot:
OK, I removed the questionable sentence, I don't have parts in it. Should be OK now?

Zimmi48 (approved):
I'll merge this PR tomorrow.

ppedrot:
@Zimmi48 ping.

Zimmi48:
Yep, sorry about that.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -1112,6 +1112,59 @@ co-inductive definitions are also allowed.
...
+
+Some properties that hold over positive streams are lost when going to the
+negative presentation, typically when they imply equality over streams.
+For instance, propositional η-equality is lost when going to the negative
+presentation. It is nonetheless logically consistent to recover it through an
+axiom.
+
+.. coqtop:: all
+
+   Axiom Stream_eta : forall s: Stream, s = cons (hs s) (tl s).
```
aa755: How do we know that this axiom is consistent with Coq?
Also, with negative coinductive types, would this axiom be consistent with other commonly used axioms, e.g.:
function extensionality, proof irrelevance, forms of proof relevance (e.g. `((or_introl I):(True \/ True)) <> or_intror I`)?

ppedrot: Just as for the functional extensionality axiom. Usually mumbling something as *it's true in Set* is enough, as Coq tries to be consistent with the set-theoretic model (barring hottist axioms).

PR #8676, opened 2018-10-08
Fixes #8672: ill-formed pattern substitution in notation with "let"

### LINKED ISSUE #8672: Pattern notation leads to `Error: The variable pat was not found in the current environment.`
#### Version
8.8.1, master

The following code produces a somewhat cryptic (unless one knows about the encoding of patterns) error message. Additionally, I don't really know what behavior to expect from this notation. Should it work at all? I think so, but my intuition here is usually wrong.

```coq
Notation "'plet' x := e 'in' t" :=
  ((fun H => let x := id H in t) e) (at level 0, x pattern).
Definition bla := plet (pair x y) := pair 1 2 in x.
```

herbelin:
**Kind:** bug fix

The body of the `let` was mistakenly dropped (see example in #8672).

Fixes / closes #8672

### DIFF
--- a/interp/notation_ops.ml
+++ b/interp/notation_ops.ml
@@ -153,10 +153,12 @@ let protect g e na =
   if disjpat <> None then user_err (Pp.str "Unsupported substitution of an arbitrary pattern.");
   e',na
 
-let apply_cases_pattern ?loc ((ids,disjpat),id) c =
-  let tm = DAst.make ?loc (GVar id) in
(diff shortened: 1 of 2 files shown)


ejgallego (approved):
LGTM

ejgallego:
Sorry @herbelin, the CI failure was spurious and didn't stay more than a few minutes in `master`, but we were unlucky and it prevented the CI to run. Can you rebase? Thanks.

PR #8655, opened 2018-10-04
Test suite: use coqc and then coqchk

vbgl:
This is a bit rough. Nonetheless, it reveals 12 modules that are produced by coqc but cannot be validated by coqchk.

Blocked by:
- #8937 (fix waiting at #8996)

PR #8679, opened 2018-10-09
[test suite] Compile test cases, instead of only loading them

### LINKED ISSUE #7853: Test-suite should use file names that are valid for `coqc`
Today, we run the test suite by feeding files on the `coqtop` input. We would like instead (at least for some of the tests), to run `coqc`, to test the native compiler and the checker. However, the file names starting with numbers are not accepted by `coqc`, and we use a lot in the test suite.

What should be done? Allow such file names or renaming the files in the test suite? I imagine it can be confusing to have modules with number-only identifiers, like `x.1`. So maybe the second solution?

vbgl:
Fixes #7853.

One step towards #8655.

This also enforces that, in each test, all proofs, modules, sections, etc. are properly terminated (or aborted).


ejgallego:
Coqc is nothing but an annoying wrapper to coqtop, why not to set the proper options to coqtop instead?

vbgl:
Sorry, the commit message is misleading. What is actually run is `coqtop -compile`.

test-suite/Makefile
```diff
@@ -43,7 +43,7 @@ coqchk := $(BIN)coqchk -coqlib $(LIB) -R prerequisite TestSuite
 coqdoc := $(BIN)coqdoc
 coqtopbyte := $(BIN)coqtop.byte
 
-coqtopload := $(coqtop) -top Top -async-proofs-cache force -load-vernac-source
+coqtopload := $(coqtop) -async-proofs-cache force -load-vernac-source
 coqtopcompile := $(coqtop) -compile
```
gares: I'm not following here. Why you remove -top from the fils that are not compiled?
Also, I guess you want to pass the async proofs options to coqtopcompile too, or `Undo` and `Reset` are going to be expensive.

gares (changes requested):

test-suite/Makefile
vbgl: You’re right that the changes in `Top` is unrelated. I think that this argument does not serve any purpose, hence removing it.

gares (approved):

PR #8680, opened 2018-10-09
Check that lambda/prod ast's have proper binders during interning/printing

herbelin:
**Kind:** internal invariants

This is the part of #8490 which ensures stronger interning invariants on the form of Lambda/Prod AST, and reports a more meaningful error on the invariant not satisfied at printing.

I don't have a strong opinion on the Constrintern part (it seems better, but I don't know if plugins around enforce the invariant). The Ppconstr part is an improvement though. @ejgallego, are you ok?


ejgallego (approved):

ejgallego:
Note the CI problem due to an outdated tree @herbelin

ejgallego:
Ping for a rebase so the CI runs [should be quick]

herbelin:
The master branch is changing so quickly... now rebased to current.

PR #8681, opened 2018-10-09
[dune] Fix and improve flags:

ejgallego:
- We use the `_` pattern for flags that are common to all profiles.
- In `dev` profile, we add warning 40 as done in make.
- In `release` profile, we use an optimizing set of flags. This will
  only affect to people using the an FLambda-enabled OCaml.

### DIFF
--- a/dune
+++ b/dune
@@ -1,10 +1,15 @@
 ; Default flags for all Coq libraries.
 (env
  (dev     (flags :standard -rectypes -w -9-27-50+40+60))
- (release (flags :standard -rectypes))
- (ireport
-  (flags :standard -rectypes -w -9-27-50+40+60)
-  (ocamlopt_flags :standard -O3 -unbox-closures -inlining-report)))
+ (release (flags :standard -rectypes)
+          (ocamlopt_flags -O3 -unbox-closures))
+ (ireport (flags :standard -rectypes -w -9-27-50+40+60)
+          (ocamlopt_flags :standard -O3 -unbox-closures -inlining-report)))
+
+; The _ profile could help factoring the above, however it doesn't
+; seem to work like we'd expect/like:
+;
+; (_ (flags :standard -rectypes)))
 
 ; Rules for coq_dune
 (rule


ejgallego:
@RalfJung once this is merged this should compile an optimized version if under a `flambda` switch. You can also `opam pin add coq https://github.com/ejgallego/coq.git#dune+default_optimized_build`

dune
```diff
@@ -1,7 +1,10 @@
 ; Default flags for all Coq libraries.
+
 (env
- (dev     (flags :standard -rectypes -w -9-27-50+60))
- (release (flags :standard -rectypes)))
+ (dev     (flags :standard -rectypes -w -9-27-50+40+60))
+ (release (flags :standard -rectypes)
+          (ocamlopt_flags -O3 -unbox-closures))
```
rgrinberg: Would this work work if coq is installed via opam? `release` is the profile that will be selected for all switches, not just those that use flambda.

We don't have a great way of addressing this in dune unfortunately as there's no way to conditionally include flags in env.

Zimmi48 (commented):
I indeed don't see the point of adding `_` if that's not reused in the rest of the profiles. The rest looks good.

dune
```diff
@@ -1,10 +1,12 @@
...
 (env
  (dev     (flags :standard -rectypes -w -9-27-50+40+60))
- (release (flags :standard -rectypes))
+ (release (flags :standard -rectypes)
+          (ocamlopt_flags -O3 -unbox-closures))
  (ireport
   (flags :standard -rectypes -w -9-27-50+40+60)
-  (ocamlopt_flags :standard -O3 -unbox-closures -inlining-report)))
+  (ocamlopt_flags :standard -O3 -unbox-closures -inlining-report))
+ (_       (flags :standard -rectypes)))
```
Zimmi48: I still think that this line is useless if we have to repeat the same thing everywhere. You even repeat it in `(release)` which has exactly the same `flags`.
And what is this very weird alignment after the merge?

ejgallego: > And what is this very weird alignment after the merge?

The alignment is fine AFAICS.

Regarding using `(_ ...)` I dunno, I am not sure what it does, my understanding indeed is that it would help for profiles that are local to some part of the sub-tree.

Zimmi48: >Regarding using (_ ...) I dunno, I am not sure what it does, my understanding indeed is that it would help for profiles that are local to some part of the sub-tree.

If you are not sure, just don't add it for now. We have enough of code that we don't understand fully.

Zimmi48 (approved):

ejgallego:
Merci!

PR #8684, opened 2018-10-09
Remove a few circumvolutions around parameters of inductive entries

maximedenes:
This PR is trying to clean a bit the kernel interfaces, by making input and output data closer (output data should be input data that has been checked + derived data).

It is very much work in progress, there's only a tiny clean-up for now, but feedback on the general idea is welcome.

### DIFF
--- a/kernel/typeops.ml
+++ b/kernel/typeops.ml
@@ -469,25 +469,19 @@ let infer_v env cv =
 
 (* Typing of several terms. *)
 
-let infer_local_decl env id = function
-  | Entries.LocalDefEntry c ->
-      let () = check_wellformed_universes env c in
-      let t = execute env c in
-      RelDecl.LocalDef (Name id, c, t)
-  | Entries.LocalAssumEntry c ->
-      let () = check_wellformed_universes env c in
-      let t = execute env c in
-      RelDecl.LocalAssum (Name id, check_assumption env c t)
-
-let infer_local_decls env decls =
-  let rec inferec env = function
-  | (id, d) :: l ->
-      let (env, l) = inferec env l in
-      let d = infer_local_decl env id d in
-        (push_rel d env, Context.Rel.add d l)
-  | [] -> (env, Context.Rel.empty)
-  in
-  inferec env decls
+let check_context env rels =
+  let open Context.Rel.Declaration in
+  Context.Rel.fold_outside (fun d env ->
+    match d with
+      | LocalAssum (_,ty) ->
+        let _ = infer_type env ty in
+        push_rel d env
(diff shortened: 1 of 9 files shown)


kernel/typeops.ml
```diff
@@ -482,25 +482,19 @@ let infer_v env cv =
...
+    match d with
+      | LocalAssum (_,ty) ->
+        let _ = infer_type env ty in
+        push_rel d env
+      | LocalDef (_,bd,ty) ->
+        let j1 = infer env bd in
+        let _ = infer_type env ty in
+        conv_leq false env j1.uj_type ty;
+        push_rel d env)
+    rels ~init:env
```
SkySkimmer: I believe this drops the check_wellformed_universes.

SkySkimmer (approved):
So this is removing LocalFooEntry?
+1

SkySkimmer (changes requested):
Actually let me request changes over my previous comment, it shouldn't just be dropped ;)

kernel/typeops.ml
maximedenes: Isn't done as part of `infer_type`?

SkySkimmer: Oh you're using infer_type
OK then

SkySkimmer (approved):

mattam82 (approved):
Nice! I'm wondering why we ever wanted a separate type for the params.

herbelin (approved):
Had a quick look, seems good and relevant.

SkySkimmer:
By the way this means that 
~~~coq
Definition nat' := nat.
Inductive foo (A:=0:nat') := .
Print foo.
~~~
should now print `nat'` instead of `nat`.

herbelin:
@SkySkimmer: would you assign?

SkySkimmer:
>@SkySkimmer: would you assign?

I'm fine with doing it but why me?

herbelin:
> I'm fine with doing it but why me?

No strong reasons. First because you commented and made good remarks about this PR. Second because @mattam82 tends to be overbusy (but maybe that's changing). Third, for rotating the responsibilities, but maybe you think that you are already doing a lot.

SkySkimmer:
>maybe you think that you are already doing a lot.

I said I'm fine though.


In any case we have to wait for overlays.

gares:
I'm doing the one for elpi

gares:
There it is
  https://github.com/LPCIC/coq-elpi/tree/overlay/8684
but I think this PR needs a rebase since I had to "undo" the last PR being merged (monic hooks by Emilio)

PR #8687, opened 2018-10-09
Mini reorganization type of global constr of global

herbelin:
**Kind:** architecture

This moves `Global.constr_of_global_in_context` to `Typeops` and rename
`Global.type_of_global_in_context` to `Typeops.type_of_global`.

These functions are purely functional so I don't think they have to be in `Global`. Presumably, there were there only because `GlobRef.t` was not yet in the kernel.

I updated a bit the doc in the mli, which was apparently partly wrong (wrong copy-paste for `Global.constr_of_global_in_context`) or outdated (mention of a side effect which seems obsolete).

I would have like to rename `constr_of_global_in_context` into `constr_of_global` but the latter name is already massively used.

Don't know what the authors/experts of these functions would like to see for naming (@mattam82, @ppedrot, @skyskimmer, ...).

The PR is short to make it easier to review.

### DIFF
--- a/kernel/typeops.mli
+++ b/kernel/typeops.mli
@@ -54,11 +54,10 @@ val type_of_variable : env -> variable -> types
 val judge_of_variable : env -> variable -> unsafe_judgment
 
 (** {6 type of a constant } *)
-
+val type_of_constant_in : env -> pconstant -> types
 val judge_of_constant : env -> pconstant -> unsafe_judgment
 
 (** {6 type of an applied projection } *)
-
 val judge_of_projection : env -> Projection.t -> unsafe_judgment -> unsafe_judgment
 
 (** {6 Type of application. } *)
@@ -89,9 +88,7 @@ val judge_of_cast :
   unsafe_judgment
 
 (** {6 Inductive types. } *)
-
 val judge_of_inductive : env -> inductive puniverses -> unsafe_judgment
-
 val judge_of_constructor : env -> constructor puniverses -> unsafe_judgment
 
 (** {6 Type of Cases. } *)
@@ -99,7 +96,7 @@ val judge_of_case : env -> case_info
   -> unsafe_judgment -> unsafe_judgment -> unsafe_judgment array
     -> unsafe_judgment
 
-val type_of_constant_in : env -> pconstant -> types
+(** {6 Miscellaneous. } *)
 
 (** Check that hyps are included in env and fails with error otherwise *)
 val check_hyps_inclusion : env -> ('a -> constr) -> 'a -> Constr.named_context -> unit
--- a/kernel/typeops.mli
+++ b/kernel/typeops.mli
@@ -106,6 +106,14 @@ val type_of_global_in_context : env -> GlobRef.t -> types * Univ.AUContext.t
     usage. For non-universe-polymorphic constants, it does not
     matter. *)
 
+(** {6 Building a term from a global reference *)
+
+(** Map a global reference to a term in its local universe
+    context. The term should not be used without pushing it's universe
+    context in the environmnent of usage. For non-universe-polymorphic
+    constants, it does not matter. *)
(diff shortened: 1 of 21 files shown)


mattam82:
I’m not sure this goes well with the current refactorings of @ppedrot but indeed it should not be in global anymore. The purpose was to use it like is done in typeclasses, where we build a fresh instance for the polymorphic context that is returned, but most code should really use the higher level Evd.new_global etc to make constr from globals.

herbelin:
> I’m not sure this goes well with the current refactorings of @ppedrot 

Oups, I probably missed this. What do you suggest I should do?

herbelin:
By the way, @mattam82, is it ok to change `UnivGen.constr_of_global` so that it takes an env?

kernel/typeops.mli
```diff
@@ -89,17 +88,33 @@ val judge_of_cast :
...
+    usage. For non-universe-polymorphic constants, it does not
+    matter. *)
+
+(** {6 Building a term from a global reference *)
+
+(** Map a global reference to a term in its local universe
+    context. The term should not be used without pushing it's universe
+    context in the environmnent of usage. For non-universe-polymorphic
+    constants, it does not matter. *)
+val constr_of_global_in_context : env -> GlobRef.t -> types * Univ.AUContext.t
```
SkySkimmer: I think usually we put the comment below (as you did for type_of_global).
At least the 2 new ones should position their respective comments the same way.

herbelin: I can put them below. I put them above by consistency with the rest of the file but I can first put all comments below in the file, if there is a consensus in this policy.

SkySkimmer:
I don't think the name `type_of_global` is right for this one, I would rather use it for the one which takes an instance as argument.
These functions when called on universe polymorphic global will produce terms with debruijn universe variables, which cannot generally be used without further work. I'd like to have some reference to that in the name. the `_in_context` suffix does that although it's occult if you don't already know what it means. Maybe something like `_abstract_univs` would work too.
Someday we may have a type of universe abstracted constr which requires explicit casting to use as regular constr but that will require more work than just renaming some functions ofc.

Note all the callers of constr_of_global_in_context are highly suspicious:
- the one in pretype_global just throws away the constr, using only the size of the context (should be using something like Global.universes_of_global then getting the size I guess)
- the one in termops.vars_of_global_reference wraps the globref in a constr to send it to the kernel, nowadays the kernel can take globref so should be fixed
- 2 in funind which should be using the regular constr_of_global since I'm pretty sure it will just explode with undefined universe if it ever tries using it on universe polymorphic constants

The type_of_global_in_context are at least nontrivial to replace so I guess they can stay.

@herbelin do you want to make a commit and/or PR to remove the constr_of_global_in_context users or should I?

herbelin:
> do you want to make a commit and/or PR to remove the `constr_of_global_in_context` users or should I?

Please do, you seem much more qualified than me on this.

As for the naming of `type_of_global`, I agree. Maybe the `in_context` suffix is discriminating enough. Using `_abstract_univs` would maybe be a bit cryptic but, at the same time, the suffix `_in_context` can be confused with the suffix `_env`.

gares (approved):

ppedrot (approved):

SkySkimmer:
@Zimmi48 looks like the bot didn't figure out that https://gitlab.com/coq/coq/-/jobs/106508656 is a transient failure.

Zimmi48:
@SkySkimmer Thanks for reporting. coq/bot@01c6f60 should fix this.

ppedrot:
Needs another rebase...

herbelin:
Rebased.

ppedrot:
Alas, the rebase is now broken!

ppedrot:
OK, I'm really sorry but after #8724 this needs another rebase that is hopefully going to be the last one before direct merge.

herbelin:
@ppedrot: done

PR #8688, opened 2018-10-09
Generalizing the various evar_map printers in Termops over an environment

herbelin:
This is a small incremental step towards limiting calls to the global environment.

Don't hesitate to discuss design choices. Better now than latter.

**Kind:** architecture

### DIFF
--- a/dev/ci/user-overlays/08688-herbelin-master+generalizing-evar-map-printer-over-env.sh
+++ b/dev/ci/user-overlays/08688-herbelin-master+generalizing-evar-map-printer-over-env.sh
@@ -1,0 +1,6 @@
+if [ "$CI_PULL_REQUEST" = "8688" ] || [ "$CI_BRANCH" = "master+generalizing-evar-map-printer-over-env" ]; then
+
+    Elpi_CI_REF=master+generalized-evar-printers-pr8688
+    Elpi_CI_GITURL=https://github.com/herbelin/coq-elpi
+
+fi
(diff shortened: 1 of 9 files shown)


ejgallego (commented):
Something that was not clear to me in this file if the printers were supposed to be "debug", that is to say, to print the raw terms themselves or not.

herbelin:
> Something that was not clear to me in this file if the printers were supposed to be "debug"

That's a good question. I can't tell for all printers, but the `evar_map` printer is not very pretty and it was mainly intended for debugging. In practice, it is however used for the errors messages of `rewrite.ml` and also in `Himsg.explain_unsatisfiable_constraints`.

> that is to say, to print the raw terms themselves or not.

This is a different question I think. When I debug, I prefer in general to have "nice" terms, and I get nice terms with the tracer for instance. In ocamldebug, the printer of the debugger does not have access to the `Global.env` anyway (i.e. it runs in an empty `Global.env`), so it does not make a difference whether we pass it an `env` which in practice shall be empty, or if it takes dynamically a `Global.env` which shall also be empty.

In any case, it is not clear that this env-passing in evar-things printers is important. But this costs nothing to make them functional either.

ppedrot (approved):

ppedrot:
@herbelin Looks like elpi needs an overlay...

herbelin:
Overlay at LPCIC/coq-elpi#36.

I'm afraid there's going to be quite a lot of overlays to come on the `Global.env()` topic...

PR #8689, opened 2018-10-09
A few useless accesses to the global environment in pretyping and engine

herbelin:
**Kind:** architecture

This is a short PR to remove in `pretyping` and `engine` a couple of lookups in the global environement which could be done instead in an environment easily available.

----

Going further is not easy. The remaining accesses to the `Global.env` are difficult to remove.

- Several of them are done in the scope of an `add_anonymous_leaf` and thus require to change the interface of `declare_object`. This is the case in `recordops.ml`, `typeclasses.ml`, `classops.ml`, `heads.ml`.

- `Evarutil.new_global`: question to universe expert, would it be correct to pass it the env known by the caller, or does it have to be _the_ global environment?
- `Evd.evar_env`, `Evd.evar_filtered_env`, `Termops.pr_evar_suggested_name`: simplest solution seems to keep a copy of the `env` in the `evar_map` (w/o rels nor named variable)??
- `Global.lookup_named` in `Detyping.detype`: needs a way to distinguish section variables from other named variables in an env...

### DIFF
--- a/engine/evarutil.ml
+++ b/engine/evarutil.ml
@@ -579,7 +579,7 @@ let rec check_and_clear_in_constr env evdref err ids global c =
                      has dependencies in another hyp of the context of ev
                      and transitively remember the dependency *)
                     let check id _ =
-                      if occur_var_in_decl (Global.env ()) !evdref id h
+                      if occur_var_in_decl env !evdref id h
                       then raise (Depends id)
                     in
                     let () = Id.Map.iter check ri in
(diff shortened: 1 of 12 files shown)


SkySkimmer:
> question to universe expert, would it be correct to pass it the env known by the caller, or does it have to be the global environment?

Caller env should be fine.

ejgallego:
Cool @herbelin , let's see if this works. This stuff is very subtle as we have "hidden" dependencies in quite a few places.

ppedrot (approved):
I agree that there is subtle stuff going on in this removal, but CI looks happy so that's probably good enough. If we ever find bugs we'll have time to see them coming.

ppedrot:
All is green, will merge tomorrow.

ppedrot:
@herbelin Can you rebase? I'll merge after.

PR #8690, opened 2018-10-09
[toplevel] Split interactive toplevel and compiler binaries.

ejgallego:
We make `coqc` a truly standalone binary, whereas `coqtop` is
restricted to interactive use.

Thus, `coqtop -compile` will emit a warning and call `coqc`.

We have also refactored `Coqargs` into a common `Coqargs` module and a
compilation-specific module `Coqcargs`.

This solves problems related to `coqc` having its own argument
parsing, and reduces the number of strange argument combinations a
lot.

Depens on #9095


ejgallego:
Ok, this is IMO ready for review and comments.

I think the main question here is what to do with options that don't make sense for each of the tools, for example `coqtop -compile`. We could have a deprecation phase, however I am not sure it is worth the cost, as skipping the deprecation phase would allow us to start having some structure in coqargs.

maximedenes:
If this PR only needs a rebase, it would be good to do it now, so that we can merge #9095 and this one. I can review it after a rebase, if you like.

ejgallego:
> If this PR only needs a rebase, it would be good to do it now, so that we can merge #9095 and this one. I can review it after a rebase, if you like.

Last time I looked this PR seemed to be working pretty well, but I wouldn't be surprised if some problem arises.

IMO the deprecation and the actual changes to the tools seem like two separate issues.

maximedenes:
@ejgallego I'd like to help making this PR ready (so that we can merge #9095 and then this one). Can I rebase it? Do you know if there's a lot of work to be done to resolve the CI failures?

maximedenes:
> Can I rebase it?

I guess I meant to rebase #9095, since the conflict seems to come from there.

ejgallego:
Needs a few easy fixes.

ejgallego:
Ok, modulo a couple of nits this is ready for review and eventual merge.

maximedenes (commented):
How about putting `coqc.ml` and friend in `comp` folder, rather than `toplevel`? Incidentally, it would become closer to what they do in OCaml.

ejgallego:
> How about putting `coqc.ml` and friend in `comp` folder, rather than `toplevel`? Incidentally, it would become closer to what they do in OCaml.

That's only two files so I didn't see a lot of gain, but we can do it if you want. I'd like indeed to do even more refactoring on the organization, but I'd rather wait for Dune to be ready so we can create libraries in a more lightweight way.

ejgallego:
The particular hottest point of the refactoring is IMO to develop the boot library to manage commonly `-R` / `-Q`  options on toplevels / coqdep.

maximedenes:
> That's only two files so I didn't see a lot of gain, but we can do it if you want. I'd like indeed to do even more refactoring on the organization, but I'd rather wait for Dune to be ready so we can create libraries in a more lightweight way.

Ok.

maximedenes (approved):

maximedenes:
I'll rebase to keep history clean.

maximedenes:
Will merge after the last CI run.

ejgallego:
Sorry there was a problem with a stale depending commit, I've fixed it now.

ppedrot (commented):
Not competent on this PR.

Zimmi48:
Just FTR @ppedrot, one can always remove a review request from the list and this has the advantage of being silent for the others (in terms of notifications).

ppedrot:
Yes, but doing so doesn't explain why I removed myself. The comment is slightly more informative.

ejgallego:
Thanks!

hendriktews:
This commit seems to disables -schedule-vio2vo in coqtop. Was this intended? If yes, I would suggest to change the documentation appropriately. See all #10679.

ejgallego:
Yes that was intended, I suggest we just fix the documentation, a 8.10 patch with warning could be prepared.

PR #8691, opened 2018-10-10
Remove some dead code in nsatz and micromega plugins

vbgl:


thery (approved):

Zimmi48 (commented):
The dependency on csdp is technically an optional dependency, both of Coq itself, and of the test-suite. It can be useful to have it in `default.nix` for developers wanting to update the test-suite, but do we really want to put it in the Nix package by default given that Coq will anyways find csdp if it's on the PATH? What about adding a new argument to `default.nix` and having `shell.nix` set this argument to `true` (like it does for `shell`)?

vbgl:
There are a few files in the test-suite (`micromega/[qr]example.v`) which do need `csdp`. They usually rely on the cache file.

thery:
shouldn't the cache been automatically removed when testing?

Zimmi48:
>There are a few files in the test-suite (micromega/[qr]example.v) which do need csdp. They usually rely on the cache file.

This is what I meant by: csdp is an optional dependency for the test-suite. This was designed this way.

vbgl:
This makes sense. I’ve removed the change to `default.nix`.

vbgl:
Any volunteer to merge this approved PR?

cc @coq/micromega-maintainers.

thery:
I will but I first need to read the instruction, last time I got scolded ;-)

vbgl:
Thanks. [Here are the instructions](https://github.com/coq/coq/blob/master/dev/doc/MERGING.md).

thery:
@vbgl
warning: This PR didn't run with the latest version of CI. 
warning: It is probably a good idea to ask for a rebase.

thery:
@vbgl is it a good idea to ask for rebase?

vbgl:
Done. The “plugin tutorial” failure is unrelated.

PR #8692, opened 2018-10-10
[test-suite] Don't reset `OCAMLPATH`, but append to it.

ejgallego:
This fixes an obvious bug introduced in #8602.

### DIFF
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -254,7 +254,7 @@ $(addsuffix .log,$(wildcard bugs/closed/*.v)): %.v.log: %.v
 # otherwise the directory name will include the quotes, see
 # see for example https://stackoverflow.com/questions/10424645/how-to-convert-a-quoted-string-to-a-normal-one-in-makefile
 ifeq ($(LOCAL),true)
-  export OCAMLPATH := $(shell echo $(COQLIBINSTALL))
+  export OCAMLPATH := $(shell echo $(COQLIBINSTALL):$$OCAMLPATH)
 endif
 
 OCAMLOPT := $(OCAMLFIND) opt $(CAMLFLAGS)


gares (approved):

gares:
I'll merge this quickly

ejgallego:
> I'll merge this quickly

Note that I am not sure this is correct, if `OCAMLPATH` is not set we end up with a variable of the form: `dir:` 

Is that trailing `:` OK?

vbgl (approved):

vbgl:
Regarding the impact of a trailing separator, here is the source of truth: https://gitlab.camlcity.org/gerd/lib-findlib/blob/master/src/findlib/fl_split.ml#L82

gares:
yep, just checked: https://gitlab.camlcity.org/gerd/lib-findlib/blob/master/src/findlib/fl_split.ml#L82

gares:
@vbgl you win!

gares:
```
val path : string -> string list = <fun>
#   path "foo:";;
- : string list = ["foo"]
# path "foo::bar:";;
- : string list = ["foo"; "bar"]
```

Zimmi48:
Why not use the standard `${OCAMLPATH:+:$OCAMLPATH}`?

ejgallego:
> Why not use the standard `${OCAMLPATH:+:$OCAMLPATH}`?

I have no idea if that is better, I'll use whatever you folks that know more about make suggest.

Zimmi48:
My suggestion is standard shell for not putting the trailing `:`. I can't comment on make.

gares:
> Why not use the standard `${OCAMLPATH:+:$OCAMLPATH}`?

I did not know `:+` was posix! I'll merge it as it, but if you open a PR I'll merge that one too ;-)

ejgallego:
Thanks! I can do whatever you tell me to do, I dunno. I am not expert in this stuff.

rgrinberg:
FYI, I'm pretty sure that ocamlfind uses `;` as the separator on Cygwin/Windows. At least this is how we do it in dune (cc @diml or @nojb to confirm).

Thanks for the tip about the duplicate separators. Dune should probably handle those too.

vbgl:
This is right, the separator is platform-dependent. Is there a canonical way to address this issue?

nojb:
> FYI, I'm pretty sure that ocamlfind uses ; as the separator on Cygwin/Windows. At least this is how we do it in dune (cc @diml or @nojb to confirm).

Indeed: https://gitlab.camlcity.org/gerd/lib-findlib/blob/master/src/findlib/fl_split.ml#L72-79

gares:
@vbgl Am I right that "appending" to OCAMLPATH is needed only on nix?

vbgl:
That’s unrelated. It’s needed whenever you contemplate installing `oUnit` and Coq in distinct places.

gares:
Sorry, I forgot to add that we are talking about CI scripts, so we can assume we are not looking for troubles (users may mix and match different setups, but on CI we may just assume simple things, I hope)

PR #8694, opened 2018-10-10
Various cleanups of universe apis

SkySkimmer:
See commit messages.

### DIFF
--- a/kernel/typeops.ml
+++ b/kernel/typeops.ml
@@ -432,21 +432,8 @@ and execute_array env = Array.map (execute env)
 
 (* Derived functions *)
 
-let universe_levels_of_constr _env c =
-  let rec aux s c =
-    match kind c with
-    | Const (_c, u) ->
-       LSet.fold LSet.add (Instance.levels u) s
-    | Ind ((_mind,_), u) | Construct (((_mind,_),_), u) ->
-       LSet.fold LSet.add (Instance.levels u) s
-    | Sort u when not (Sorts.is_small u) ->
-      let u = Sorts.univ_of_sort u in
-      LSet.fold LSet.add (Universe.levels u) s
-    | _ -> Constr.fold aux s c
-  in aux LSet.empty c
-
 let check_wellformed_universes env c =
-  let univs = universe_levels_of_constr env c in
+  let univs = universes_of_constr c in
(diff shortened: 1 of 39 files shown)


ppedrot (commented):
I have a mixed feeling about this PR. While most of it is a very welcome cleanup, I am not so convinced by most of the moves of functions from upper layers into the kernel. There is no reason to inflate it with stuff that is only used by universe unification...

SkySkimmer:
I'll have another look when I rebase then.

ppedrot:
Maybe to be more precise: my main concern is `universes_of_constr` which is somewhat nonsense.

kernel/typeops.ml
```diff
@@ -432,21 +432,8 @@ and execute_array env = Array.map (execute env)
...
-    match kind c with
-    | Const (_c, u) ->
-       LSet.fold LSet.add (Instance.levels u) s
-    | Ind ((_mind,_), u) | Construct (((_mind,_),_), u) ->
-       LSet.fold LSet.add (Instance.levels u) s
-    | Sort u when not (Sorts.is_small u) ->
-      let u = Sorts.univ_of_sort u in
-      LSet.fold LSet.add (Universe.levels u) s
-    | _ -> Constr.fold aux s c
-  in aux LSet.empty c
```
SkySkimmer: universes_of_constr was already in the kernel, see this part of the diff

ppedrot: Bah, I missed that. I guess that's OK then.

ppedrot (approved):

kernel/typeops.ml
mattam82: It is used to check the universes are declared mainly, what's wrong with that?

ppedrot: I read it too quickly and thought that was the old version that was randomly picking universes declared in the constants. The current version is sane indeed.

ppedrot:
@mattam82, @maximedenes If that's fine with you guys I'll merge that.

mattam82 (approved):
Nice cleanups indeed, nothing to add. I think the move of globref manipulations to the kernel is harmless and useful.

ppedrot:
CI failure spurious, about to merge.

PR #8695, opened 2018-10-10
Adding a functional version of constant- and mind_of_delta_kn + functional version of is_polymorphic

herbelin:
**Kind:** architecture

This is a very short PR introducing a functional version of `constant_of_delta_kn` and `mind_of_delta_kn` in order to be able to call them without referring to the global environment.

We put these new functions in `Safe_typing` which seems to be the only reasonable place where to put them inbetween `Safe_typing` and `Global`.

I (reasonably) checked that the two "mod" and "param" resolvers are independent and removed the comment which was asking about whether they were independent.

Note that the two functions depend on the _full_ safe environment (with all the module structure) and not only on the `Environ.env`. This seems to be the only ones in this case. (Maybe that should be taken into account in organizing the global state wrt to its kernel components.) 

I don't know the exact projects of @ppedrot about kernel pairs, but I don't think it is a problem if `constant_of_delta_kn` and `mind_of_delta_kn` eventually become obsolete.

### DIFF
--- a/kernel/environ.ml
+++ b/kernel/environ.ml
@@ -680,6 +680,16 @@ let remove_hyps ids check_context check_value ctxt =
   in
   fst (remove_hyps ctxt)
 
+(* A general request *)
+
+let is_polymorphic env r =
+  let open Names.GlobRef in
+  match r with
+  | VarRef _id -> false
+  | ConstRef c -> polymorphic_constant c env
+  | IndRef ind -> polymorphic_ind ind env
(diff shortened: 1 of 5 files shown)


ppedrot:
> I don't know the exact projects of @ppedrot about kernel pairs, but I don't think it is a problem if constant_of_delta_kn and mind_of_delta_kn eventually become obsolete.

That's part of the plan indeed. Might take longer than expected though.

herbelin:
I extended the PR by adding a second function which is worth existing in a functional way: `is_polymorphic`.

ppedrot (approved):

PR #8697, opened 2018-10-10
[dune] Fix public name of tool: coq_tex -> coq-tex

ejgallego:


Zimmi48 (approved):

ejgallego:
Thanks a lot for the review @Zimmi48, I suggest you don't merge and I
amend this PR to incorporate your comments.

> However are `coqmlp5` and `coqp5` useful outside the build of Coq itself?

Yes, plugins should call them to perform the ml4 -> ml translation;
unless `coqpp` takes over before.

> What about `coq_dune` is it ready for consumption by external packages?

Not yet, but it may be soon so it makes sense installing it.

> * coqidetop (it installs coqidetop.opt only)
> I think it is a choice not to install coqidetop indeed.

Indeed, we could install it easily but I am not sure I see the point.

> * coqworkmgr
> What about coqworkmgr? I don't know what this is.

This is a STM tool, indeed good point, we should build it an install.

> * all .byte

This was also a choice, we install only `coqtop.byte`, but we could
install some others if people want.

tools/dune
```diff
@@ -31,25 +38,11 @@
...
-
-(executable
- (name coq_dune)
- (public_name coq_dune)
+(executables
+ (names coqwc coq_tex coq_dune)
+ (public_names coqwc coq-tex coq_dune)
  (package coq)
- (modules coq_dune)
+ (modules coqwc coq_tex coq_dune)
```
Zimmi48: Do you actually need to give the module names here (isn't Dune able to figure them out by itself)?

tools/dune
```diff
@@ -31,25 +38,11 @@
...
  (libraries str))
```
Zimmi48: coqwc didn't depend on str before.

tools/dune
```diff
@@ -22,6 +22,13 @@
  (modules coqc)
  (libraries coq.toplevel))
 
+(executable
+ (name coqworkmgr)
+ (public_name coqworkmgr)
+ (package coq)
+ (modules coqworkmgr)
```
Zimmi48: So isn't Dune smart here and able to locate the right module by itself?

ejgallego: > So isn't Dune smart here and able to locate the right module by itself?

As I mentioned in other PR the default rule is to use all modules in a dir, that makes sense for a very common setup of one dir, several modules, no library, one executable.

herbelin:
BTW, while the dynamics is clearly launched in preferring building a future on top of clearer and more uniform APIs, and clearer and smoother documentation, what about also stepwise renaming `coq_makefile` into `coq-makefile`?

ejgallego:
> what about also stepwise renaming `coq_makefile` into `coq-makefile`?

I am neutral to this.

Zimmi48:
If the goal is anyways to deprecate coq_makefile on the long term, maybe the renaming isn't worth it...

herbelin:
I guess I should emit the hypothesis that there are projects to use dune for compiling Coq projects!

Zimmi48:
Indeed, I think that given the close dialog between Emilio and Dune developers, the goal is even to have first-class support for Coq in Dune (eventually).

Issue #6446, opened 2017-12-16
Future of Coq's internal `Backtrace` module


ejgallego:
I am using OCaml's native backtraces for some stuff and indeed they seem better than what Coq does provide via it's private `Backtrace` module. In particular, I see inlined functions. I also see the logic monad operations.

Should we update this module to use newer features? Should we just get rid of it?

ppedrot:
This module was introduced to work around a missing API in OCaml. AFAIR we need to switch to a more recent version of OCaml to be able to remove it.

maximedenes:
What version would it require?

ppedrot:
I thought that it was OCaml 4.03, but actually I was wrong and it turns out from the documentation 4.02.0 is enough. So actually it's already possible to perform the move.

ejgallego:
Note also the point about using OCaml's native `reraise`, but if I am not mistaken there were some problems. It also requires 4.03 I think.

```
external reraise : exn -> 'a = "%reraise"
```

eponier:
There is a function `raise_with_backtrace` in `Printexc` that allows to raise an exception with a specific backtrace, but it requires 4.05.0 (see https://github.com/gasche/ocaml-releases-change-explanation/wiki/4.05.0-changes-explanation#debugging-and-profiling).

ejgallego:
Ok, so we are ready for this; how should be proceed to migrate?

Is is enough to change:
```ocaml
with exn ->
    let reraise = CErrors.push exn in
    (* cleanup *)
    iraise reraise
```
to
```ocaml
with exn ->
   (* cleanup *)
   reraise exn
```

gares:
I think so

PR #8700, opened 2018-10-10
Removing debugging warning when no exception handler is registered in futures

### LINKED ISSUE #8685: Failure of post-action to definition shows the `no fix_exn` message
#### Version

From Coq 8.5 to current master.

```
Fail Coercion c := eq_refl 0.
(* 
c is defined
Future: no fix_exn.
You have probably created a Future.computation from a value without passing the ~fix_exn argument.  You probably want to chain with an already existing future instead.
The command has indeed failed with message:
Error: c is not a function.
*)
```
I found this while working on unifying the code for declaration of lemmas and declaration of regular definitions. I wanted to backport the fix 2aafc6f1 of #3780 to `DeclareDef` but this started to generate this error. I then realized that I could reproduce a similar message with the above.

The code of `DeclareDef.declare_definition` already calls the coercion hook with a `fix_exn` and I have difficulties to understand what the exact problem is. Could it be a problem of propagating the `fix_exn` in chains of hooks? What is the right way to use `fix_exn`? Passed to the hook as in `DeclareDef.declare_definition` or applied to the final exception, as in `Lemmas.save`?

herbelin:
**Kind:** architecture

Fixes / closes #8685 

As far as I understood from @gares at #8685, this was useful for tine-tuning the stm but this is no longuer needed: it is ok not to have exception handler when a constant registration does not span over several commands (such as "Goal ... Qed" or obligations). Please correct me if this is wrong.

### DIFF
--- a/lib/future.ml
+++ b/lib/future.ml
@@ -33,7 +33,7 @@ let _ = CErrors.register_handler (function
   | _ -> raise CErrors.Unhandled)
 
 type fix_exn = Exninfo.iexn -> Exninfo.iexn
-let id x = prerr_endline "Future: no fix_exn.\nYou have probably created a Future.computation from a value without passing the ~fix_exn argument.  You probably want to chain with an already existing future instead."; x
+let id x = x
 
 module UUID = struct
   type t = int


ejgallego:
> As far as I understood from @gares at #8685, this was useful for tine-tuning the stm but this is no longuer needed: it is ok not to have exception handler when a constant registration does not span over several commands (such as "Goal ... Qed" or obligations). Please correct me if this is wrong.

@gares mentions that the origin of this message is because some exception was not re-raised properly.

This could be a more serious bug as location information is also there, do we have the backtrace that produces this error?

herbelin:
> What is the "hook" and why do we need it? What kind of workflow we are getting in the coercion that a future is forced?

A post-action called "hook" can be attached to a theorem/definition. This post-action is typically a coercion, an instance declaration, a canonical structure. The hook is called after the body of the definition or the theorem is pushed.

In the case of theorems, the hook is called at Qed/Defined time, well after the initial command, so, in a different state.

Any theorem/definition generates a "future". If there is a problem with this future, one has to treat the exception in the context in which the theorem was initially started. This is (afaiu) what the exception fix is doing: ensuring that the exception is printed in the correct state (the infamous `process_vernac_interp_error`).

> But IMVHO sweeping this problem below the carpet is not the right thing; this seems a good opportunity to cleanup all the mess there.

As far as I understand, but @gares can correct me if I'm wrong, we are not sweeping anything under the carpet. It is legitimate to let the exception flows normally, without fixing its state, when the exception is ensured to be raised in the exact same command as when the definition started. I.e., it is legitimate not to "fix" the exception in:
```
Definition a := 0
```
or 
```
Coercion a (n:nat) := Some n.
```
but it is not in
```
Definition a : nat.
exact  0.
Qed.
```

ejgallego:
> A post-action called "hook" can be attached to a theorem/definition. This post-action is typically a coercion, an instance declaration, a canonical structure. The hook is called after the body of the definition or the theorem is pushed.

Thanks a lot @herbelin , I see a bit better. Now my first question is:
- why do we need to create a Future for `Definition foo := a.` if the body is there?
- why do we need a hook for the above too? Is that due to program?

ejgallego:
> As far as I understand, but @gares can correct me if I'm wrong, we are not sweeping anything under the carpet.

I have looked at the code going on on `Lemma/Definition` save and certainly I am not sure what is going on there. Something seems fishy to me, but of course I am likely to be wrong, but I'd like at least we try to discuss it a bit more.

ejgallego:
The backtraces we get don't help, OCaml's native backtraces are quite better IMHO, I hope we can progress on #6446 some day.

ejgallego:
I've added this PR to the discussion on the WG page. Pity @gares won't be able to be there :( :(

herbelin:
> - why do we need to create a Future for `Definition foo := a.` if the body is there?

I'd say that we have to comply to the `Future` type, even if it is immediately computed. @gares could tell more probably.

> - why do we need a hook for the above too? Is that due to program?

I don't know if we "need" hooks for coercions. It is how this is implemented and we have to keep the information of what to do somewhere. Program's obligations also uses hooks but I did not spend time to understand the details. @mattam82 is the one to ask.

ejgallego:
> I don't know if we "need" hooks for coercions. It is how this is implemented and we have to keep the information of what to do somewhere. Program's obligations also uses hooks but I did not spend time to understand the details. @mattam82 is the one to ask.

Indeed I understand it now, however the current flow is a bit strange as hooks are captured by the terminators which are defined in the start scope but ran in the end scope with some partial capturing.

I succeeded in moving hooks to the commands themselves and that looks clean [I don't think this is a problem for Equations / Program], however, `terminators` are giving way more trouble in the sense they do quite complicated things so that's not easy to unify, not to say that they are part of the info of the proof that is sent across workers and thus lost due to its higher-order nature.

This will require quite a bit of thinking.

ejgallego (commented):
In principle I am fine removing the warning, however, as this PR is pretty trivial and the risk of conflict / painful rebase is basically non-existent I'd like to wait a bit more to see what more can we understand about the control-flow here.

gares:
> "terminators" that are not preserved by the workers

There is the entire point of the ephemerons: they wrap data that should not reach the workers. In this case the data is a closure (hard to marshal) and it is data needed at Qed time (that the worker is not supposed to do).

My impression is that the type of terminators is too generic.

> @gares mentions that the origin of this message is because some exception was not re-raised properly.

There are 2 reasons why the warning can be printed:
- one (written in the message that it prints) is that one creates a future without telling it on which sentence the error should be reported (this is what the ugly fix_exn is for)
- one created a future with the right information and then chained it with some code that does not let the (fixed) exception go trough, it catches it and raises another one without propagating the exninfo.

I've added the warning to see and fix the first kind of bugs. And at some point I was convinced I fixed all of them. Hence I speculate it is of the second form.

Note that the issue is only partially related to the messy control flow in the STM.
Some future computation goes outside the STM, eg it reaches the kernel as part of an opaque proof in the env. Then the join_safe_environment kernel call may force the future and raise an exception. Such an exception should be annotated with a sensible state_id or the UI will no be able to tell you where it applies.

ejgallego:
> My impression is that the type of terminators is too generic.

Indeed it looks like we could benefit from making them not attached to a proof, [and just tag the relevant proof metadata]. I have a tree doing that however it will have to wait due to other deadlines.

What I wonder tho is what is the current situation with the handling of terminators, am I confused or we just don't handle them properly then for proofs coming from the workers?

Does that mean that the system works because the classifier refuses to delay anything which doesn't carry a standard terminator?

herbelin:
> one created a future with the right information and then chained it with some code that does not let the (fixed) exception go trough, it catches it and raises another one without propagating the exninfo.

Then, I'm lost: should we track this rather than removing the fix_exn message?

herbelin:
@gares: what should then be done here? Should we track what leads to the fix_exn message or remove the message?

gares:
I'd say both. I re-read the thread, @ejgallego talks about discussing the issue at the WG but I was not there. Did it happen?

ejgallego:
Sorry folks it was not my intention to block this PR, which I'll leave at @gares discretion.

Something is strange tho wrt hooks, if the hook is declaring a coertion or class for example, and it is run a `Qed` time, it means that all the definitions with hooks can't just be processed asynchronously, right?

Is that enforced by the STM?

gares:
> Something is strange tho wrt hooks, if the hook is declaring a coertion or class for example, and it is run a `Qed` time, it means that all the definitions with hooks can't just be processed asynchronously, right?
> 
> Is that enforced by the STM?

This has nothing to do with this PR that suppresses a warning for developers.

gares:
> Sorry folks it was not my intention to block this PR, which I'll leave at @gares discretion.

Please merge this PR, and move the discussion about terminators elsewhere.

ejgallego:
> Please merge this PR, and move the discussion about terminators elsewhere.

Oh indeed sorry, I thought this PR was touching a file in the `stm` so didn't realize I had to assign.

ejgallego:
Will rebase to be sure the build works.

ejgallego:
> This has nothing to do with this PR that suppresses a warning for developers.

Oh well, it seems it does if you look at the original bug @herbelin meant to fix. We have a Coercion there, and it seems it is not chained with the right context, isn't it? Why that happens with the Coercion? Could it be the hook? :hotsprings: :hotsprings: 

I'll merge because I've lost hope that the `fix_exn` situation can be fixed without undergoing quite a deep re-structuring of the control flow, so it shouldn't matter so much whether we print the warning or not; but indeed, there are many questions (and bugs) open here and I disagree that this is not the right place to discuss.

gares:
The example in #8685 works fine, both with and without Fail, and it works in CoqIDE as well and without this patch. So to me the obscure print is what the bug report is about.

Since It prints an error message, someone is probably catching and re-raising the exception. It is surely in the code of the terminators, where the coercion is actually declared. This is another story.

The error message the STM is printing is not helping, so I think it should go, at least for the final user.

PR #8701, opened 2018-10-10
Fix build of Nix package with sandbox.

Zimmi48:
One more place where we must patch shebangs.

**Kind:** bug fix

Unfortunately, I don't know how to turn the sandbox on in CI so we cannot use it to make sure that the build doesn't break again.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -75,7 +75,10 @@ stdenv.mkDerivation rec {
         (path: _:
            !elem (baseNameOf path) [".git" "result" "bin" "_build" "_build_ci"]) ./.;
 
-  preConfigure = "patchShebangs kernel/";
+  preConfigure = ''
+    patchShebangs kernel/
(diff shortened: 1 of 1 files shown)


vbgl (approved):

vbgl:
I think this should be back-ported to 8.9. Currently, the test-suite fails with:

> logs/tools/update-compat.log
run.sh: dev/tools/update-compat.py: /usr/bin/env: bad interpreter: No such file or directory

cc/ @silene

silene:
This cannot be backported to 8.9, as there is no `preConfigure` on 8.9. My intuition tells me that the fix is obvious. But since I have no clue what `preConfigure` and `patchShebangs` do, I don't trust my intuition on this matter.

vbgl:
No problem. I’ll submit a patch against the 8.9 branch.

PR #8702, opened 2018-10-10
[coq_dune] Abstract path operations wrt directory separator.

ejgallego:
Even if cosmetic, this is useful for window-based hosts.

### DIFF
--- a/tools/coq_dune.ml
+++ b/tools/coq_dune.ml
@@ -72,6 +72,9 @@ module Aux = struct
 
   let sep fmt () = fprintf fmt "@;"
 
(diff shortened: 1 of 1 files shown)


tools/coq_dune.ml
```diff
@@ -72,6 +72,10 @@ module Aux = struct
 
   let sep fmt () = fprintf fmt "@;"
 
+  (* See what is the real problem *)
```
Zimmi48: Is this comment a left-over?

ejgallego: Oh indeed, sorry, will fix.

PR #8703, opened 2018-10-10
[vernac] Remove unused `start_hook` `save_hook` from Lemmas.

ejgallego:
These hooks are 10-years old and long unused I think.


ppedrot:
I think it used to be set by the XML plugin, if memory serves.

ejgallego:
> I think it used to be set by the XML plugin, if memory serves.

Could be, it was indeed used by Coq itself in the 2008 implementation, see e8ef565dadd110329f806fa3c281055fcd807440

ejgallego:
This is trivial I think and thus ready for merge.

herbelin:
OK, let me assign. Shall merge late tonight.

herbelin (approved):
It is simple enough to add it again if ever some plugin needs it.

ejgallego:
¡Gracias!

PR #8705, opened 2018-10-10
[vernac] [hooks] Refactor towards optional hooks.

ejgallego:
We make `declaration_hook`s optional arguments everywhere, and thus we
avoid some "fake" functions having to be passed.

This identifies positively the code really using hooks [funind,
rewrite, coercions, program, and canonicals] and helps moving toward
some hope of reification.

### DIFF
--- a/plugins/funind/functional_principles_types.ml
+++ b/plugins/funind/functional_principles_types.ml
@@ -310,7 +310,6 @@ let build_functional_principle (evd:Evd.evar_map ref) interactive_proof old_prin
       (Decl_kinds.Global,false,(Decl_kinds.Proof Decl_kinds.Theorem))
       !evd
       (EConstr.of_constr new_principle_type)
-      hook
   ;
     (*       let _tim1 = System.get_time ()  in *)
       let map (c, u) = EConstr.mkConstU (c, EConstr.EInstance.make u) in
@@ -326,7 +325,7 @@ let build_functional_principle (evd:Evd.evar_map ref) interactive_proof old_prin
       match entries with
       | [entry] ->
         discard_current ();
-        (id,(entry,persistence)), CEphemeron.create hook
+        (id,(entry,persistence)), hook
       | _ ->
         CErrors.anomaly Pp.(str "[build_functional_principle] close_proof returned more than one proof term")
   end
@@ -386,7 +385,7 @@ let generate_functional_principle (evd: Evd.evar_map ref)
   (* Pr  1278 :
      Don't forget to close the goal if an error is raised !!!!
   *)
-  save false new_princ_name entry g_kind hook
+  save false new_princ_name entry g_kind ~hook
   with e when CErrors.noncritical e ->
     begin
       begin
--- a/vernac/declareDef.mli
+++ b/vernac/declareDef.mli
@@ -14,8 +14,9 @@ open Decl_kinds
 val get_locality : Id.t -> kind:string -> Decl_kinds.locality -> bool
 
 val declare_definition : Id.t -> definition_kind ->
+  ?hook:Lemmas.declaration_hook ->
   Safe_typing.private_constants Entries.definition_entry -> UnivNames.universe_binders -> Impargs.manual_implicits ->
-  Lemmas.declaration_hook -> GlobRef.t
+  GlobRef.t
 
 val declare_fix : ?opaque:bool -> definition_kind ->
   UnivNames.universe_binders -> Entries.constant_universes_entry ->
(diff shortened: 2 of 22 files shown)


ejgallego:
Interestingly `call_hook` is only executed in 4 places:

- `Lemmas.save`
- `Lemmas.admit`
- `Declaredef.declare_definition`
- `Obligations.declare_{mutual_,}definition`
- `Indfun_common.save`

And the delayed ones need to have their exception "fixed", which seems a bit strange, as the command producing the error is in the right scope. Need to understand 289b2b9a1c3d3ebc3c1975663d16c04e88b6329b better [note to self].

plugins/funind/functional_principles_types.ml
```diff
@@ -320,7 +319,7 @@ let build_functional_principle (evd:Evd.evar_map ref) interactive_proof old_prin
     (* 	  let dur1 = System.time_difference tim1 tim2 in *)
     (* 	  Pp.msgnl (str ("Time to compute proof: ") ++ str (string_of_float dur1)); *)
     (* 	end; *)
-    get_proof_clean true, CEphemeron.create hook
+    get_proof_clean true, hook
```
herbelin: If there was a ephemeron, wasn't there a reason for it?

herbelin:
Looks good to me.

This changes the API of key functions but it seems that 8.10 (at least if `Global.env` is threaded) is going to introduce significant (arguable) API changes anyway.

I guess nothing is opposed to more pervasively use optional arguments?

Note that I'm actually working also on the API of vernac/com*.ml files and hooks. For instance, I'm unifying the `declaration_hook` and `univ_declaration_hook` (@mattam82, would you see a problem at doing that?). Eventually, one of the objective is to factorize the 3 copies of `declare_definition`, `save`, and `save_remaining_theorems` which do the same things but each with specificities which are worth to be unified.

plugins/funind/functional_principles_types.ml
ejgallego: As of today hooks are captured in the `terminators` thus already wrapped in an "ephemeron" by `Proof_global`. So indeed I dont' see why this is necessary, but I agree that a more careful review is needed.

ejgallego:
> This changes the API of key functions but it seems that 8.10 (at least if `Global.env` is threaded) is going to introduce significant (arguable) API changes anyway.

Indeed, so far this PR is more of an experiment to grab feedback than a real proposal for merging, we could introduce deprecation wrappers indeed if we deem so, but in the end the changes may turn out too deep (that is, in the model itself) so I prefer to wait a bit more on the compat history.



> I guess nothing is opposed to more pervasively use optional arguments?

Optional arguments have the risk that you forget to use them :) In this case they seemed like the right choice to me due the large use of empty hooks tho. Anyways I have a further cleanup where I move hooks out of the low-level declaration path so we'll see how it looks. [IMO a bit better]

> I'm unifying the `declaration_hook` and `univ_declaration_hook`

Indeed I think that makes sense; without too much of thinking of familiarity with the code yet, I'd say that unification among the `Obligations` / `Lemmas` path would be really good.



> Eventually, one of the objective is to factorize the 3 copies of `declare_definition`, `save`, and `save_remaining_theorems` which do the same things but each with specificities which are worth to be unified.

Indeed. Maybe the most complicated issue is the way the STM plugs into the declaration code; it does so in very tricky ways [see for example the handling of terminators], so indeed that may make the refactoring a bit harder.

plugins/funind/functional_principles_types.ml
herbelin: > As of today hooks are captured in the terminators thus already wrapped in an "ephemeron" by Proof_global

This is the case in `start_proof` but here, the hook is passed to `Indfun_common.save` which does no encapsulation in a terminator.

I mean, I don't undertand well what the impact of using ephemeron or not is but this precise hook is not encapulated in a terminator.

ejgallego: Indeed thanks, let me investigate more.

vernac/declareDef.mli
```diff
@@ -14,8 +14,9 @@ open Decl_kinds
 val get_locality : Id.t -> kind:string -> Decl_kinds.locality -> bool
 
 val declare_definition : Id.t -> definition_kind ->
```
ppedrot: Style remark: I prefer putting optional arguments before all the remaining ones, in the style of Coq implicit arguments. By no means a request for change.

ppedrot (approved):
Doesn't look hurtful.

ejgallego:
Should be ready, overlays:
- https://github.com/ybertot/plugin_tutorials/pull/20
- https://github.com/coq-community/paramcoq/pull/16
- https://github.com/LPCIC/coq-elpi/pull/43
- https://github.com/mattam82/Coq-Equations/pull/155

ejgallego:
This is ready, note the interaction with #8718 , IMHO this could go in or not.

Who wants to assign?

herbelin:
@ppedrot: you said it does not hurt. Do you want to take it? (And if not, assign it to me.)

ppedrot:
I'm in favour of merging it, but @ejgallego being the author he gets the last word. Are you fine with the merging, @ejgallego?

ejgallego:
Not the 8th wonder of the world but IMHO this patch will allow us to make some progress on reifying hooks, which still do remain a big problem in the async setup as they are ephemeral data.

ppedrot:
Banzai, then.

PR #8710, opened 2018-10-11
[omega, btauto] Remove some dead code

vbgl:


ppedrot (commented):
You should probably also add an mli file to coq_omega.ml, otherwise it is unclear whether people can depend on this code or not.

vbgl:
Excellent idea. This also revealed some more dead code.

ppedrot (approved):

Issue #8495, opened 2018-09-18
[api] [regression] Critical `pre_env` component `globals` has disappeared in 8.9


ejgallego:
In 8.8:
```
type globals = {
  env_constants : constant_key Cmap_env.t;
  env_inductives : mind_key Mindmap_env.t;
  env_modules : module_body MPmap.t;
  env_modtypes : module_type_body MPmap.t}
```
In 8.9/8.10:
```
type globals
```
this is a core API and breaks most of my developments.

ejgallego:
This is an example of a bug that would have a hard time following the rule "no milestone without assignee", isn't it?.

Another one is for example #8027

Zimmi48:
Indeed, this rule can bear a few exceptions.

silene:
> This is an example of a bug that would have a hard time following the rule "no milestone without assignee", isn't it?

What is the reason for not assigning @SkySkimmer ?

ejgallego:
There was a PR fixing this issue, it was however not accepted but the issue remains.

vbgl:
Is there someone willing to address this issue soon before the 8.10 release? Otherwise I’ll postpone it. Thanks.

ejgallego:
@vbgl so far we didn't agree on what the API for clients that need to use the environment should be, see discussion at #8228 , IMVHO we should bring back the old API until a new one is established [people is now using Obj.magic!] but if not I guess we can just postpone this, it's been already broken in 8.9

ppedrot:
This has been solved, right? Closing.

PR #8712, opened 2018-10-11
[stm] avoid unshare safe env to detect correctly env changing tactics

### LINKED ISSUE #8609: STM: sending back a fat state

gares:
Fix: #8609 

I believe it was introduced in de20a45db5d45154819f2ed4359effdbb8bf0292 where the option (part of the summary) is moved to the save env. By setting the summary, you unshare the safe env. Now we do that only if needed. The stm uses `==` on the safe env to detect tactics that alter the env, eg abstract.

### DIFF
--- a/kernel/environ.ml
+++ b/kernel/environ.ml
@@ -384,8 +384,26 @@ let set_engagement c env = (* Unsafe *)
   { env with env_stratification =
     { env.env_stratification with env_engagement = c } }
 
+(* It's convenient to use [{flags with foo = bar}] so we're smart wrt to it. *)
+let same_flags {
+     check_guarded;
+     check_universes;
+     conv_oracle;
+     share_reduction;
+     enable_VM;
+     enable_native_compiler;
+  } alt =
+  check_guarded == alt.check_guarded &&
+  check_universes == alt.check_universes &&
+  conv_oracle == alt.conv_oracle &&
+  share_reduction == alt.share_reduction &&
+  enable_VM == alt.enable_VM &&
+  enable_native_compiler == alt.enable_native_compiler
+[@warning "+9"]
+
 let set_typing_flags c env = (* Unsafe *)
-  { env with env_typing_flags = c }
+  if same_flags env.env_typing_flags c then env
+  else { env with env_typing_flags = c }
 
 (* Global constants *)
 
--- a/kernel/safe_typing.ml
+++ b/kernel/safe_typing.ml
@@ -192,7 +192,9 @@ let set_engagement c senv =
     engagement = Some c }
 
 let set_typing_flags c senv =
-  { senv with env = Environ.set_typing_flags c senv.env }
+  let env = Environ.set_typing_flags c senv.env in
+  if env == senv.env then senv
+  else { senv with env }
 
 let set_share_reduction b senv =
   let flags = Environ.typing_flags senv.env in


ejgallego:
> I believe it was introduced in [de20a45](https://github.com/coq/coq/commit/de20a45db5d45154819f2ed4359effdbb8bf0292) where the option (part of the summary) is moved to the save env.

Indeed an option modifying the global env seems pretty bad due to the reasons you pointed out. Isn't there any other way to do this?

ejgallego:
> Isn't there any other way to do this?

I don't mean about what the PR does, which seems fine, but about the setting of the flag by an option.

gares:
What is shaky, imo, is the ==, env should provide fast and sound comparison....

ejgallego:
I agree, I do diffs with ppx (and Obj.magic as coq doesn't want to provide a proper API in the name of ??)

library/global.ml
```diff
@@ -281,6 +281,8 @@ let set_strategy k l =
 let set_reduction_sharing b =
   let env = safe_env () in
   let flags = Environ.typing_flags (Safe_typing.env_of_safe_env env) in
-  let flags = { flags with Declarations.share_reduction = b } in
-  let env = Safe_typing.set_typing_flags flags env in
-  GlobalSafeEnv.set_safe_env env
+  if flags.Declarations.share_reduction = b then ()
+  else
+    let flags = { flags with Declarations.share_reduction = b } in
+    let env = Safe_typing.set_typing_flags flags env in
```
SkySkimmer: Would it work better to have set_typing_flags do this automatically?

gares:
Hum, yes I guess it would cover more cases... Feel free to push on this branch

SkySkimmer:
I made 7b35cb7c5190f552dc540fc7112d3b8e46e371e4 but not sure if we want to go that way (the `same_flags` I wrote feels a bit adhoc)
WDYT?

gares:
LGTM, feel free to supersede this PR with your commit.
(be sure you manage to run a proof in async mode and you don't get the fat state message, there is CI checking this)

SkySkimmer:
If CI checks it do I need to check it locally? (I don't know how to check locally)

ppedrot (approved):

ejgallego (approved):
Seems good as a workaround, even if we should have a better solution in the long-term due to the fragility of this.

maximedenes:
> there is CI checking this

@gares was this missing a negation?

gares:
rebased and tested. @maximedenes please merge

maximedenes:
> rebased and tested. @maximedenes please merge

Ok, will merge once CI passes.

gares:
@maximedenes CI passed

maximedenes:
@gares I recently introduced functions that look a lot like `set_typing_flags`. Did I break something?

maximedenes:
What I mean is that maybe we should add a comment explaining when it is critical to keep the env shared, WDYT? I personally have no idea of when it matters.

ejgallego:
> I personally have no idea of when it matters.

I'd dare to say than in general the whole approach is bound to create problems. STM or any other document manager cannot function properly without proper introspection of envs, as they must be serialized and sent incrementally among workers. As Enrico said using physical equality is just a horrible hack.

From the recent experience it seems people closer to the kernel don't understand the kind of needs upstream users such as the document manager have; #8495 for example is not looking like will be solved which will delay any attempt for another 6 months, or who knows by how much.

gares:
If you replace == by = of the logical env (indicatives + definitions/proofs) you get a slower implementation of what I'm doing. PMP put in the part of the environment I would not compare extra stuff breaking ==. I think a better API could be Environment.equal and that one would ignore some irrelevant bits and, if env is opaque, exploit a sharing invariant or use hashes or symgens.

maximedenes:
What I'm saying is that I added stuff in typing flags (see #8870), so you probably need to add fields in `same_flags`. Please add a comment saying what's going on here, so that people who extend `env` know what they are supposed to do.

maximedenes:
@gares did you have a look?

gares:
I think this PR should be fixed then, since the new fields are not taken into account (I guess we could locally enable the warning for matching over a record, and add a comment).

gares:
But, this PR is hopefully for 8.9, not for 8.10, so maybe it should be backported as such and we should re-fix the problem on master only.

maximedenes:
@gares I think it would be safer to write it correctly for master, then backporting should be easy (dropping the fields that don't exist). But we can handle this PR as @silene wishes.

gares:
CI green flag.

gares:
@maximedenes I think the milestone is appropriate. Given that the PR also needs to be backported I'd merge this soon rather than later.

PR #8713, opened 2018-10-11
[ltac] [rewrite] Remove deprecated Add Setoid / Morphism forms.

ejgallego:
Only CoRN/math-classes seem to need fixes.

@SkySkimmer asked about this so submitting. Seems doable as of today.


doc/sphinx/addendum/generalized-rewriting.rst
```diff
@@ -579,27 +579,8 @@ Printing relations and morphisms
...
-
-   This command is restricted to the declaration of morphisms
-   without parameters. It is not fully backward compatible since the
-   property the user is asked to prove is slightly different: for n-ary
-   morphisms the hypotheses of the property are permuted; moreover, when
-   the morphism returns a proposition, the property is now stated using a
-   bi-implication in place of a simple implication. In practice, porting
-   an old development to the new semantics is usually quite simple.
+Legacy ``Add Setoid`` and ``Add Morphism`` commands have been removed
+in Coq 8.10.
```
Zimmi48: No need to say this here.

Zimmi48 (commented):
I don't recall if this was already discussed when these commands were deprecated, but the replacement for `Add Setoid` seems much more complex according to the changes in the standard library!

herbelin:
> I don't recall if this was already discussed when these commands were deprecated, but the replacement for Add Setoid seems much more complex according to the changes in the standard library!

Could this be mitigated by adding support for an extra item `equivalence proved by term`?

ejgallego:
Indeed I have no idea what would the best path ot action be here. I guess that's up to the maintainers. I had this tree lying around due to being convenient for some other cleanups.

CoRN has around 6/8 deprecated declarations so that shouldn't be too hard to fix.

ppedrot (commented):
The code I am supposed to be an owner of looks correct, but indeed the question of the easiness of use is still there.

SkySkimmer:
Should we finish this or undeprecate? I don't know why it was deprecated originally but the warning was added in anticipation of removing the vtunknown in add morphism by removing the command, and it seems we're going for another approach now (#9530).

ejgallego:
Up to the maintainers; I keep this open as we tend to open the same PR every once in a while.

Zimmi48:
If there is no strong motive, I think we should clearly undeprecate. Anyways, it was unreasonable to deprecate in the first place if the alternative solution is horrible to use. This goes against our usual deprecation policies.

ejgallego:
The fact that #9129 was able to be fully implemented without needed this speaks IMO in favor of un-deprecation.

While the maintainers think what to do, I'll keep this PR open, but once #9129 is merged, I'll re-evaluate the current technical status and in all likeness close this PR.

PR #8716, opened 2018-10-12
Lemmas: Little simplification of artificially convoluted code

herbelin:
**Kind:** cleanup

This is a small cleanup local to `lemmas.ml` while on the way to unify and clean the API of the different variants of `save` and `declare_definition`.

### DIFF
--- a/vernac/lemmas.ml
+++ b/vernac/lemmas.ml
@@ -277,18 +277,9 @@ let save_remaining_recthms (locality,p,kind) norm univs body opaq i (id,(t_i,(_,
         let kn = declare_constant id ~local (DefinitionEntry const, k) in
         (locality,ConstRef kn,imps)
 
-let save_named ?export_seff proof =
-  let id,const,uctx,do_guard,persistence,hook = proof in
-  save ?export_seff id const uctx do_guard persistence hook
-
 let check_anonymity id save_ident =
   if not (String.equal (atompart_of_id id) (Id.to_string (default_thm_id))) then
     user_err Pp.(str "This command can only be used for unnamed theorem.")
-
-let save_anonymous ?export_seff proof save_ident =
-  let id,const,uctx,do_guard,persistence,hook = proof in
-  check_anonymity id save_ident;
-  save ?export_seff save_ident const uctx do_guard persistence hook
 
 (* Admitted *)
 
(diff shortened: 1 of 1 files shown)


vernac/lemmas.ml
```diff
@@ -277,19 +277,10 @@ let save_remaining_recthms (locality,p,kind) norm univs body opaq i (id,(t_i,(_,
         let kn = declare_constant id ~local (DefinitionEntry const, k) in
         (locality,ConstRef kn,imps)
 
-let save_named ?export_seff proof =
-  let id,const,uctx,do_guard,persistence,hook = proof in
-  save ?export_seff id const uctx do_guard persistence hook
-
 let check_anonymity id save_ident =
```
maximedenes: By the way, is there a plan to get rid of this hack and introduce a proper notion of unnamed theorems?

herbelin: Sorry, I don't understand your comment. What do you call a "proper notion of unnamed theorems"?

maximedenes: Well, `check_anonymity` is trying to figure out whether a theorem is anonymous based on the first letters of its name... Instead of this hack, I guess the name should be optional (or made inaccessible when the theorem is meant to be anonymous).

herbelin: Ah sure. I did not have plans but I can make plans if you want. (Or maybe you already have something like that in a branch?)

maximedenes: No, I don't have any design ready, in fact I heard about this function when @vbgl found it while working on other vernac clean-ups. It would be an interesting discussion to have. Sorry for hi-jacking this PR by the way (but it did move a call to `check_anonymity`, so it made me think about it).

herbelin: OK, waiting for comments of @vbgl.

vbgl: I agree that the handling of the “names” of anonymous theorems is a bit ugly and might deserve something more principled.

However, I don’t think it’s a reason to reject this cleaning PR. Thus, I’d let Hugo decide whether he wants to do some more work in this PR or to keep it as is and postpone/delegate the refactoring for unnamed theorems.

maximedenes: > However, I don’t think it’s a reason to reject this cleaning PR.

Oh sure, I didn't want to suggest that either.

ejgallego (approved):
Great.

PR #8704, opened 2018-10-10
[vernac] Remove unused abstraction from declaration_hook type.

ejgallego:
"Declaration" hooks can be polymorphic on their return type, however
this facility doesn't seem used in the codebase.

We thus remove the polymorphism with the hope to be able to reify the
control later on.

PR #8717, opened 2018-10-12
Lemmas, DeclareDef: internal reorganization of code highlighting what can be shared

herbelin:
**Kind:** cleanup

This is yet another small incremental PR, just factorizing code.

No change of semantics at this stage (except the removal of change of locality from `Discharge` to `Local` in `Lemmas.save` but without effect on the hooks which uses locality in `Vernacentries.vernac_definition_hook`).

No .mli change either.

`DeclareDef.ml` better to be seen in split view.

### DIFF
--- a/vernac/declareDef.ml
+++ b/vernac/declareDef.ml
@@ -33,34 +33,22 @@ let get_locality id ~kind = function
 | Local -> true
 | Global -> false
 
-let declare_global_definition ident ce local k pl imps =
-  let local = get_locality ident ~kind:"definition" local in
-  let kn = declare_constant ident ~local (DefinitionEntry ce, IsDefinition k) in
-  let gr = ConstRef kn in
+let declare_definition ident (local, p, k) ce pl imps hook =
+  let fix_exn = Future.fix_exn_of ce.const_entry_body in
+  let gr = match local with
+  | Discharge when Lib.sections_are_opened () ->
+      let _ = declare_variable ident (Lib.cwd(), SectionLocalDef ce, IsDefinition k) in
+      let () = if Proof_global.there_are_pending_proofs () then warn_definition_not_visible ident in
+      VarRef ident
+  | Discharge | Local | Global ->
+      let local = get_locality ident ~kind:"definition" local in
+      let kn = declare_constant ident ~local (DefinitionEntry ce, IsDefinition k) in
+      ConstRef kn in
   let () = maybe_declare_manual_implicits false gr imps in
   let () = Declare.declare_univ_binders gr pl in
   let () = definition_message ident in
-  gr
-
-let declare_definition ident (local, p, k) ce pl imps hook =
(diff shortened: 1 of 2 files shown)


mattam82 (approved):
Looks fine to me and clearer indeed.

SkySkimmer:
>change of locality from Discharge to Local

Anyone know why we did that?

ejgallego (approved):
Other than the question by @SkySkimmer the PR looks good!

herbelin:
> Anyone know why we did that?

It is because `Discharge`, which means `Variable`/`Hypothesis` or `Let` from the parser point of view, when not in a section, behaves as `Local` (which - enigmatically - means "do not modify the nametab even when imported").

So, at some time of the interpretation process, "`Discharge` not in section" has to be interpreted into "`Local`".

I hesitated to introduce a function to reinterpret "`Discharge` not in section" into "`Local`" early in the interpretation process, but I thought that there were more risk of forgetting calling this function than to keep it in the 3 `save` functions which I plan to merge into one.

Now, I realize while writing this comment, that such test could be centralized in `Locality.enforce_locality_exp`. Let me do that.

herbelin:
> Locality.enforce_locality_exp. Let me do that.

Actually, I have a more general question. It is well-known that the `Local` and `Global` keywords mean different things in different contexts, user-side, but let's try to clarify the role of  `Local`, `Global`, `Discharge`, `local=true`, `local=false` already in the code.

I'd be tempted to propose the following change:
1. In `Decl_kinds`, replacing:
    ```
    type locality = Discharge | Local | Global
    ```
    by
    ```
    type import_status = ImportAllowUnqualified | ImportEffectLess
    type locality = Discharge | Persistent of import_status
    ```
2. In `Declare.declare_constant`, replace the type of `~local` from `bool` to `import_status`. I don't know if we can go up to changing the name of the flag from `~local` to `~import`, that would be a too big extra incompatibility (??)
3. In other (non-logical) meanings of `local`, e.g. in `Classops.declare_coercion`, keep `~local` to mean discharge if in section, do not `vo`-export if not in section.

Currently, local and global have different meanings depending on whether the object is logical or not.

For logical objects, global currently means visible without qualifier after import beyond the scope of a module/file.

For non-logical objects, global means either surviving section (when in a section) or surviving module/files (when not in a section).

In particular, the terminology above avoids (in the code) to use `Global` and `Local` for logical objects, preferring `ImportAllowUnqualified` and `ImportEffectLess`, but this is a bit heavy, so I'm wondering if we can do better?

ejgallego:
> Does it sound good to you folks?

I may proceed later today if no comment.

herbelin:
> Does it sound good to you folks?

It does sound good to me.

ejgallego:
Needs a rebase @herbelin

ejgallego:
CI failures are due to #8704 overlays not being merged.

PR #8720, opened 2018-10-12
[Numeral notations] Use Coqlib registered constants

### LINKED ISSUE #8564: Numeral notations should maybe use Coqlib for constants
Is this the wrong way to do things?  If so, it should be fixed in master:
https://github.com/coq/coq/blob/7cc70b0df61718a946327d5bfb056b140eeb54ba/plugins/syntax/numeral.ml#L36-L40

vbgl:
Fixes #8564.

Waiting for HoTT release, see https://github.com/coq/coq/pull/8720#issuecomment-434597150

### DIFF
--- a/plugins/syntax/numeral.ml
+++ b/plugins/syntax/numeral.ml
@@ -36,19 +36,12 @@ let get_constructors ind =
 let qualid_of_ref n =
   n |> Coqlib.lib_ref |> Nametab.shortest_qualid_of_global Id.Set.empty
 
-let q_int () = qualid_of_ref "num.int.type"
-let q_uint () = qualid_of_ref "num.uint.type"
 let q_option () = qualid_of_ref "core.option.type"
 
 let unsafe_locate_ind q =
   match Nametab.locate q with
   | IndRef i -> i
   | _ -> raise Not_found
-
-let locate_ind q =
-  let q = q () in
-  try unsafe_locate_ind q
-  with Not_found -> Nametab.error_global_not_found q
 
 let locate_z () =
   let zn = "num.Z.type" in
@@ -64,8 +57,17 @@ let locate_z () =
   else None
 
 let locate_int () =
-  { uint = locate_ind q_uint;
-    int = locate_ind q_int }
+  let int = "num.int.type" in
+  let uint = "num.uint.type" in
+  if Coqlib.has_ref int && Coqlib.has_ref uint
+  then
+    let q_int = qualid_of_ref int in
+    let q_uint = qualid_of_ref uint in
+    Some ({
+        int = unsafe_locate_ind q_int;
+        uint = unsafe_locate_ind q_uint;
+    }, mkRefC q_int, mkRefC q_uint)
+  else None
 
 let has_type f ty =
   let (sigma, env) = Pfedit.get_current_context () in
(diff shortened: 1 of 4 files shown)


theories/Init/Decimal.v
```diff
@@ -40,7 +40,7 @@ Notation zero := (D0 Nil).
 
 (** For signed integers, we use two constructors [Pos] and [Neg]. *)
 
-Inductive int := Pos (d:uint) | Neg (d:uint).
+Variant int := Pos (d:uint) | Neg (d:uint).
```
SkySkimmer: Why this change?

vbgl: I agree it’s unrelated. But since I was already touching this file, I took the opportunity to tidy it a bit…

plugins/syntax/numeral.ml
```diff
@@ -33,50 +33,59 @@ let get_constructors ind =
...
   with Pretype_errors.PretypeError _ -> false
 
-let type_error_to f ty loadZ =
+let type_error_to f ty =
   CErrors.user_err
     (pr_qualid f ++ str " should go from Decimal.int to " ++
      pr_qualid ty ++ str " or (option " ++ pr_qualid ty ++ str ")." ++
-     fnl () ++ str "Instead of Decimal.int, the types Decimal.uint or Z could be used" ++
-     (if loadZ then str " (require BinNums first)." else str "."))
+     fnl () ++ str "Instead of Decimal.int, the types Decimal.uint or Z could be used.")
```
JasonGross: What is the reason for dropping the parenthetical?  You could instead augment it to say `require BinNums or Register a constant as ... first`.

Either way, please also update the documentation to match the updated error message.

JasonGross (changes requested):
Thanks!

vbgl:
I’ve just updated the error messages and the corresponding documentation.

Zimmi48 (approved):

plugins/syntax/numeral.ml
```diff
@@ -33,50 +33,59 @@ let get_constructors ind =
...
+     fnl () ++ str "Instead of Decimal.int, the types Decimal.uint or Z could be used (you may need to require BinNums or Decimal first).")
 
-let type_error_of g ty loadZ =
+let type_error_of g ty =
   CErrors.user_err
     (pr_qualid g ++ str " should go from " ++ pr_qualid ty ++
      str " to Decimal.int or (option Decimal.int)." ++ fnl () ++
-     str "Instead of Decimal.int, the types Decimal.uint or Z could be used" ++
-     (if loadZ then str " (require BinNums first)." else str "."))
+     str "Instead of Decimal.int, the types Decimal.uint or Z could be used (you may need to require BinNums or Decimal first).")
```
JasonGross: I think you should only display the part about `Require ` if the types don't resolve / aren't registered.  But I don't object strongly.

JasonGross (changes requested):

vbgl:
This PR breaks HoTT; the overlay is ready but would break the compatibility of HoTT with Coq 8.9. Thus we wait: 1. that Coq 8.9 is released (or at least the β); 2. that a version of HoTT compatible with said version of Coq is officially tagged; 3. that HoTT merges the overlay.

No need to hurry.

ppedrot (commented):
I'll let people more knowledgeable comment.

SkySkimmer:
>needs: merge of dependency

What's the dependency?

SkySkimmer:
I've only seen that tag used for other Coq PRs before but I guess nothing else fits better.

vbgl:
HoTT has been updated, so this PR is ready now.

JasonGross (approved):
My requests about error messages stand, but I'm not going to block this PR on that basis, and it would be nice to see this land.

ejgallego:
Taking this as part of @coq/parsing-maintainers

ejgallego (approved):
Look pretty straightforward, will merge soon if not further comments.

PR #8724, opened 2018-10-12
[universes] deprecate constr_of_global

mattam82:
In favor of a constr_of_monomorphic_global function. When people
move to the new Coqlib interface they will also see this deprecation
message encouraging them to think about the best move.

This commit changes a few references to constr_of_global and replaces
them with a constr_of_monomorphic_global which makes it apparent that
this is not the function to call to globalize polymorphic references.
The deprecation message for constr_of_global suggests the possible alternatives.

The remaining parts using constr_of_monomorphic_global are easily
identifiable using this: omega, btauto, ring, funind and auto_ind_decl
mainly (this fixes firstorder). What this means is that the symbols
registered for these tactics have to be monomorphic for now.

### DIFF
--- a/engine/univGen.ml
+++ b/engine/univGen.ml
@@ -77,17 +77,14 @@ let fresh_global_instance ?loc ?names env gr =
   let u, ctx = fresh_global_instance ?loc ?names env gr in
   mkRef (gr, u), ctx
 
-let constr_of_global gr =
-  let c, ctx = fresh_global_instance (Global.env ()) gr in
-    if not (Univ.ContextSet.is_empty ctx) then
-      if Univ.LSet.is_empty (Univ.ContextSet.levels ctx) then
-        (* Should be an error as we might forget constraints, allow for now
-           to make firstorder work with "using" clauses *)
-        c
-      else CErrors.user_err ~hdr:"constr_of_global"
-          Pp.(str "globalization of polymorphic reference " ++ Nametab.pr_global_env Id.Set.empty gr ++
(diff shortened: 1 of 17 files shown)


engine/univGen.ml
```diff
@@ -130,7 +130,7 @@ let fresh_inductive_instance env sp =
 let fresh_constructor_instance env sp =
   fresh_constructor_instance env sp None
 
-let constr_of_global gr =
+let constr_of_monomorphic_global gr =
   let c, ctx = fresh_global_instance (Global.env ()) gr in
     if not (Univ.ContextSet.is_empty ctx) then
       if Univ.LSet.is_empty (Univ.ContextSet.levels ctx) then
```
SkySkimmer: Should we keep accepting polymorphic values with empty instances or should we start testing with is_polymorphic?

mattam82:
You’re right, I’ll fix that.
Le sam. 13 oct. 2018 à 12:50, Gaëtan Gilbert <notifications@github.com> a
écrit :

ppedrot (approved):
Modulo @SkySkimmer remark this looks good.

gares (approved):

engine/univGen.ml
mattam82: Fixed

mattam82:
I adressed @SkySkimmer 's comment and rebased

ppedrot:
You missed a deprecated use it seems.

mattam82:
Right, in funind. Fixed.

PR #8718, opened 2018-10-12
Further reorganization/simplification of declaration hooks

herbelin:
**Kind:** architecture / cleanup

~~Depends on #8704 and #8717.~~

The main change is to unify the two kinds of hooks, the one depending on a `UState.t` and the other, leading to removal of redundancies (mainly in `lemmas.ml`).

Otherwise, it also starts a "hook API". Don't know yet if it is worth encapsulating it in a module.

It partially overlaps with #8705 in providing a `no_hook` where #8705 uses an optional argument.

- [ ] Entry added in `dev/doc/changes.md`? Probably to be added when stabilized

Better to skip the dependencies and see the diff only for the proper, last three, commits.

PR #8726, opened 2018-10-12
More robust treatment of the Discharge status

### LINKED ISSUE #8722: Inconsistent behaviors with respect to Local Definition
#### Version

Tested in 8.8

One can observe inconsistent behaviors of `admit` and automatic declaration of instances with respect to `Local Definition`.

1. `Admitted` on `Let` outside section and or `Local Definition` does not respect the "do not qualify at import" status (since 8.5)
```
Module TestAdmitVisibility.
  Module A.
    Let a1 : nat. Admitted. (* Suppose to behave like a "Local Definition" *)
    Local Definition b1 : nat. Admitted. (* Told to be a "Local Definition" *)
    Local Definition c1 := 0.
    Local Parameter d1 : nat.
    Section S.
      Let a2 : nat. Admitted. (* Told to be turned into a toplevel assumption *)
      Local Definition b2 : nat. Admitted. (* Told to be a "Local Definition" *)
      Local Definition c2 := 0.
      Local Parameter d2 : nat.
    End S.
  End A.
  Import A.
  Check a1. (* a toplevel Let thus should fail  [WRONG?] *)
  Check b1. (* a Local Definition thus should fail  [WRONG?] *)
  Fail Check c1.
  Fail Check d1.
  Check a2. (* a toplevel Let turn into an axiom, thus should fail [WRONG?] *)
  Check b2. (* a Local Definition, thus should fail [WRONG?] *)
  Fail Check c2.
  Fail Check d2.
End TestAdmitVisibility.
```
2. Automatic declaration of variables/axioms as instances does not work for variables automatically turned into axioms (since 8.6)
```
Module TestVariableAsInstances.
  Module Test1.
    Set Typeclasses Axioms Are Instances.
    Class U.
    Local Parameter b : U.
    Definition testU := _ : U. (* _ resolved *)

    Class T.
    Variable a : T.  (* warned to be the same as "Local Parameter" *)
    Definition testT := _ : T. (* _ resolved *)
  End Test1.

  Module Test2.
    Unset Typeclasses Axioms Are Instances.
    Class U.
    Local Parameter b : U.
    Fail Definition testU := _ : U. (* _ unresolved *)

    Class T.
    Variable a : T.  (* warned to be the same as "Local Parameter" thus should not be an instance *)
    Definition testT := _ : T. (* _ resolved [WRONG?] *)
  End Test2.
End TestVariableAsInstances.
```
3. `Local Axiom` is reported in the glob file as a variable rather as an axiom. This seems wrong.

### LINKED ISSUE #8725: Inconsistent warnings about toplevel Let turned into Local Definition, whether defined interactively or not
#### Version

In 8.8.

This is pretty minor, but the test checking that a toplevel `Let` is finally a `Local Definition` is not done in case of a proof.
```
Let x := 1. (* warning: x is declared as a local definition *)
Let y : nat.
exact 1.
Defined. (* no warning [WRONG?] *)
```
If a consensus to tell that this is not intended, a fix shall come.

herbelin:
**Kind:** cleanup / bug fix

Has overlays to be merged synchronously:
- LPCIC/coq-elpi#62
- mattam82/Coq-Equations#210
- Mtac2/Mtac2#210
- mit-plv/fiat#27
- coq-community/paramcoq#32

Fixes #8722

Update: It was intended to also fix #8725 but the latter has been fixed differently.

- [ ] Entry to be added in CHANGES.md? I have no idea whether one can fall on these side corner examples by chance, and rely on them. [Update: personally, I would not add a comment to the coq-level change log but I can if at least someone insist.]

See [this](https://github.com/coq/coq/pull/8717#issuecomment-429317576) comment for some questions on the design issues. The names of constructors are a bit long but I could not find one both short and non too ambiguous. [Update: I found a simplification of the names of constructors. See [below](https://github.com/coq/coq/pull/8726#issuecomment-497938838).]

See #8722 and #8725 to evaluate whether the reported behaviors have indeed to be fixed or not.

I changed the message `id is declared as a local definition` into the much longer `Interpreting this declaration as if a global declaration prefixed by "Local" i.e. as a global declaration which shall not be available without qualification when imported.`. The point is that the message is now given at the step going from parser to command interpreter. I drop the name of `id` but I can reintroduce it if there is a demand. Suggestions to simplify the message welcome. [Update: the message seems then ok.]

@ejgallego: Where do you suggest I put the locality type? Maybe in `declare.ml`? [Update: I kept them in `decl_kinds.ml` which seems ok.]

### DIFF
--- a/dev/ci/user-overlays/08726-herbelin-master+more-stable-meaning-to-Discharge-flag.sh
+++ b/dev/ci/user-overlays/08726-herbelin-master+more-stable-meaning-to-Discharge-flag.sh
@@ -1,0 +1,23 @@
+if [ "$CI_PULL_REQUEST" = "8726" ] || [ "$CI_BRANCH" = "master+more-stable-meaning-to-Discharge-flag" ]; then
+
+    fiat_parsers_CI_BRANCH=master+change-for-coq-pr8726
+    fiat_parsers_CI_REF=master+change-for-coq-pr8726
+    fiat_parsers_CI_GITURL=https://github.com/herbelin/fiat
+
+    elpi_CI_BRANCH=coq-master+fix-global-pr8726
+    elpi_CI_REF=coq-master+fix-global-pr8726
+    elpi_CI_GITURL=https://github.com/herbelin/coq-elpi
+
+    equations_CI_BRANCH=master+fix-global-pr8726
+    equations_CI_REF=master+fix-global-pr8726
+    equations_CI_GITURL=https://github.com/herbelin/Coq-Equations
+
+    mtac2_CI_BRANCH=master+fix-global-pr8726
+    mtac2_CI_REF=master+fix-global-pr8726
+    mtac2_CI_GITURL=https://github.com/herbelin/Mtac2
+
+    paramcoq_CI_BRANCH=master+fix-global-pr8726
+    paramcoq_CI_REF=master+fix-global-pr8726
+    paramcoq_CI_GITURL=https://github.com/herbelin/paramcoq
+
+fi
--- a/dev/doc/changes.md
+++ b/dev/doc/changes.md
@@ -5,6 +5,10 @@
 - Functions and types deprecated in 8.10 have been removed in Coq
   8.11.
 
+- Type Decl_kinds.locality has been restructured, see commit
+  message. Main change to do generally is to change the flag "Global"
+  to "Global ImportDefaultBehavior".
+
 ## Changes between Coq 8.9 and Coq 8.10
 
 ### ML4 Pre Processing
--- a/doc/plugin_tutorial/tuto1/src/simple_declare.ml
+++ b/doc/plugin_tutorial/tuto1/src/simple_declare.ml
@@ -7,7 +7,7 @@ let edeclare ?hook ident (_, poly, _ as k) ~opaque sigma udecl body tyopt imps =
   DeclareDef.declare_definition ident k ce ubinders imps ?hook_data
 
 let declare_definition ~poly ident sigma body =
-  let k = (Decl_kinds.Global, poly, Decl_kinds.Definition) in
+  let k = Decl_kinds.(Global ImportDefaultBehavior, poly, Definition) in
   let udecl = UState.default_univ_decl in
   edeclare ident k ~opaque:false sigma udecl body None []
 
--- a/interp/declare.ml
+++ b/interp/declare.ml
@@ -39,7 +39,7 @@ type constant_obj = {
   cst_decl : Cooking.recipe option;
   (** Non-empty only when rebuilding a constant after a section *)
   cst_kind : logical_kind;
-  cst_locl : bool;
+  cst_locl : import_status;
 }
 
 type constant_declaration = Safe_typing.private_constants constant_entry * logical_kind
@@ -63,8 +63,9 @@ let cooking_info segment =
 (* Opening means making the name without its module qualification available *)
 let open_constant i ((sp,kn), obj) =
   (* Never open a local definition *)
-  if obj.cst_locl then ()
-  else
+  match obj.cst_locl with
+  | ImportNeedQualified -> ()
+  | ImportDefaultBehavior ->
     let con = Global.constant_of_delta_kn kn in
     Nametab.push (Nametab.Exactly i) sp (ConstRef con)
 
@@ -137,7 +138,7 @@ let register_constant kn kind local =
   update_tables kn
 
 let register_side_effect (c, role) =
-  let () = register_constant c (IsProof Theorem) false in
+  let () = register_constant c (IsProof Theorem) ImportDefaultBehavior in
   match role with
   | Subproof -> ()
   | Schema (ind, kind) -> !declare_scheme kind [|ind,c|]
@@ -175,7 +176,7 @@ let define_constant ?role ?(export_seff=false) id cd =
   let kn, eff = Global.add_constant ?role ~in_section id decl in
   kn, eff, export
 
-let declare_constant ?(internal = UserIndividualRequest) ?(local = false) id ?(export_seff=false) (cd, kind) =
+let declare_constant ?(internal = UserIndividualRequest) ?(local = ImportDefaultBehavior) id ?(export_seff=false) (cd, kind) =
   let () = check_exists id in
   let kn, _eff, export = define_constant ~export_seff id cd in
   (* Register the libobjects attached to the constants and its subproofs *)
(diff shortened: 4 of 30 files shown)


ejgallego:
@herbelin I've moved the locality info to declaredef , let me finish that patch, then I look at this one, and I rebase on top.

Won't happen today tho.

ejgallego (commented):
I am not an expert on this area but I can confirm this PR goes well with other refactorings being done in `Proof_global` / `Lemmas`.

aspiwack (approved):
I'm marked as a code owner because of a trivial single-line change in the derive plugin.

But I'd just like to say that this makes the code overall simpler and clearer. I expect it to be entirely uncontroversial.

ppedrot (approved):

ppedrot:
@ejgallego Are you assigning this to you? Or should I?

ppedrot:
@ejgallego Are you assigning this to you? Or should I?

ppedrot:
Are the CI failures related? I don't know whether there has been some discrepancy between the devs and the master branch...

herbelin:
Failure of Mtac2 overlay seems similar to what @ppedrot reported [here](https://github.com/coq/coq/pull/7925#issuecomment-434262470):
```
Fatal error: exception Sys_error("$(COQMF_COQLIB)/user-contrib/Unicoq: 
No such file or directory")
```

ppedrot:
@herbelin I found out the problem, it is due to the Mtac2 branch for the CI. Instead of master, the PR needs to be based on master-sync.

herbelin:
@ppedrot: thanks.

So, this PR has 5 overlays. Obviously, with `Declare.declare_definition`, we are touching at a key function of the API for plugin.

Maybe worth to simplify further the API? I'm a bit worry by the complexity of a so-called "simple declare" in `tuto1`...

SkySkimmer:
>So, this PR has 5 overlays. Obviously, with Declare.declare_definition, we are touching at a key function of the API for plugin.

@herbelin did you intend to comment in #8718? This PR doesn't touch declare_definition's API AFAICT.

herbelin:
> This PR doesn't touch declare_definition's API AFAICT.

It does: one of argument has a different type.

herbelin:
This should be OK. Rebased and double-checking CI.

(In passing, if someone has an idea for a ligher name for the constructors of `type import_status = ImportAllowUnqualified | ImportNeedQualified`... otherwise, let's stay as it is.)

maximedenes:
> Fixes #8722, fixes #8725

The PR does not seem to contain test files in `bugs`. You can have a look at #9328, which has one.

vbgl:
Hugo, could you please rebase to fix the conflicts and add the missing test case? Or update the milestone if this change does not belong to version 8.10.

Thanks.

vbgl:
Rebased. Please double-check, in particular the only part where I had to be creative: https://github.com/coq/coq/pull/8726/commits/e1e1732574366ed0ab2509db1142bda7566e30b8#diff-e740c45ba76a0ca63da5a9b470a55ac3R106

The test case is still missing.

Overlays are probably out-of-date.

ejgallego:
Looks fine to me, even if that's a pity that the proof rework tree will have to undo some of the moves here due to the reification of the control flow.

Regarding overlays I dunno, I could always include this PR in the proof rework tree and take care of them this way.

vbgl:
Added a test-case.

vbgl:
Fixed the plugin-tutorial.

herbelin:
@vgbl: thanks for the rebasing and the test suite. About your question in `abstract.ml`, it should be `ImportAllowUnqualified` (I pushed the change).

Actually, I'm not proud of the terminology:
- `ImportAllowUnqualified` is supposed to mean: available at import without qualification (was the former meaning of `Global` which I found misguiding because not suggesting a difference of status after `Import`)
- `ImportNeedQualified` is supposed to mean: needs to be qualified even after `Import` (was the former meaning of `Local` which itself was somehow source of confusion with discharged)

Maybe `AvailableAfterImportWithoutQualification` and `NeedsQualificationEvenAfterImport` would be clearer? (But it is so long...).

Overlays: I updated the overlays for Equations, Mtac2, Elpi. I added an overlay for paramcoq (but w/o making a PR).

ejgallego:
> * `ImportNeedQualified` is supposed to mean: needs to be qualified even after `Import` (was the former meaning of `Local` which itself was somehow source of confusion with discharged)

Thanks for the explanations @herbelin , what are such cases, for Local inside sections?

SkySkimmer:
>Thanks for the explanations @herbelin , what are such cases, for Local inside sections?

In the v file syntax `Local Definition` is ImportNeedQualified
Recall that the meaning of user syntax Local/Global is highly dependent on the command it's attached to.

ejgallego:
> In the v file syntax `Local Definition` is ImportNeedQualified
> Recall that the meaning of user syntax Local/Global is highly dependent on the command it's attached to.

I see, thanks @SkySkimmer , that confuses me a bit, and I would think that `Local Definition` would mean "not import at all", à la Haskell.

ejgallego:
> I see, thanks @SkySkimmer , that confuses me a bit, and I would think that `Local Definition` would mean "not import at all", _à la_ Haskell.

Of course you need to close the dependency graph of definitions / proofs in our case; and that may be annoying.

ejgallego:
Is there some model of module systems with "hiding", I guess definition depending on private constants could be closed with an existential, in fact this is already done on purpose by some libraries to really hide stuff.

herbelin:
This PR was adding a new warning to be consistent with a missing case of the "definition-not-visible" warning (a `Let` made with tactics). But this warning was removed in #10188 so, I propose to remove the new warning (called `warn_statement_not_visible` in the PR) also to remain consistent with the new absence of warning. @skyskimmer, I guess this looks ok to you.

SkySkimmer:
OK with me

herbelin:
> OK with me

OK, I removed it.

I also made a last change. I was not really satisfied with the replacement of `Global` by the long and a bit headache-name `Persistent ImportAllowUnqualified`. I replaced it with `Global ImportDefaultBehavior` so that users who wants an easy migration path canonically replace `Global` by `Global ImportDefaultBehavior`. Of course, it depends on the intended meaning of the vernac command `Local Definition` and of how we want or can implement it, since it is not clear how to implement it in a stronger way than by only briddling the ability to qualify the name.

herbelin:
This seems technically ready: pending comments addressed, CI green, PRs for overlays submitted.

herbelin:
The possible criticism I see to this PR is the number of plugin overlays it needs. But based on the basis that API are still not stable, I'd assume that it is acceptable.

@ppedrot, you assigned apparently.

herbelin:
Note: once more time rebased.

Zimmi48:
> I changed the message `id is declared as a local definition` into the much longer `Interpreting this declaration as if a global declaration prefixed by "Local" i.e. as a global declaration which shall not be available without qualification when imported.`. The point is that the message is now given at the step going from parser to command interpreter. I drop the name of `id` but I can reintroduce it if there is a demand. Suggestions to simplify the message welcome. [Update: the message seems then ok.]

I had not seen this but the new message seems overly complex. I will open a PR to simplify it. Furthermore, the documentation should have been updated. I will do this as well.

JasonGross:
I think the identifier should appear in the warning, otherwise it is nearly useless on the command line

PR #8727, opened 2018-10-13
[options] New helper for creation of boolean options plus reference.

ejgallego:
This makes setting the option outside of the synchronized summary impossible.

### DIFF
--- a/engine/namegen.ml
+++ b/engine/namegen.ml
@@ -208,23 +208,16 @@ let it_mkLambda_or_LetIn_name env sigma b hyps =
 (* Introduce a mode where auto-generated names are mangled
    to test dependence of scripts on auto-generated names *)
 
-let mangle_names = ref false
-
-let () = Goptions.(
-    declare_bool_option
-      { optdepr  = false;
-        optname  = "mangle auto-generated names";
-        optkey   = ["Mangle";"Names"];
-        optread  = (fun () -> !mangle_names);
-        optwrite = (:=) mangle_names; })
+let get_mangle_names =
+  Goptions.declare_bool_option_and_ref
+    ~depr:false
+    ~name:"mangle auto-generated names"
+    ~key:["Mangle";"Names"]
+    ~value:false
 
 let mangle_names_prefix = ref (Id.of_string "_0")
-let set_prefix x = mangle_names_prefix := forget_subscript x
 
-let set_mangle_names_mode x = begin
-    set_prefix x;
-    mangle_names := true
-  end
+let set_prefix x = mangle_names_prefix := forget_subscript x
 
 let () = Goptions.(
     declare_string_option
@@ -238,7 +231,7 @@ let () = Goptions.(
                       with CErrors.UserError _ -> CErrors.user_err Pp.(str ("Not a valid identifier: \"" ^ x ^ "\".")))
                    end })
 
-let mangle_id id = if !mangle_names then !mangle_names_prefix else id
+let mangle_id id = if get_mangle_names () then !mangle_names_prefix else id
 
 (* Looks for next "good" name by lifting subscript *)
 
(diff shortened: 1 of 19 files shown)


SkySkimmer:
>This makes setting the option outside of the synchronized summary impossible.

What does that mean? Why do we care?

ejgallego:
> What does that mean? Why do we care?

As of today it doesn't mean a lot indeed so indeed I should be more clear about this being basically a cleanup. This is in fact part of a small refactoring trying to make group options to behave well [see the problems in #8666 .

Imagine we have `flag_foo = false`, then some code does: `flag_foo := true`. Now the `Summary` contains a different value for `foo`.

If that happens below the `Vernacentries.interp` freeze barrier, then the flag will be correctly saved upon, however, it that happens above, the change to `foo` will be lost as the summary will be reinstalled.

Thus, this API could provide a method for the upper layers to actually modify a setting properly, which is what I was thinking when I wrote that.

SkySkimmer:
>Thus, this API could provide a method for the upper layers to actually modify a setting properly

The API only restricts how it may be set compared to the preexisting method, do you mean that it forces to use the proper method?

ejgallego:
> The API only restricts how it may be set compared to the preexisting method, do you mean that it forces to use the proper method?

It is a step towards making the "proper method" work, as of today it forces you go to thru `Goptions` to set the value, however, `Goptions` still doesn't do the proper thing.

ejgallego:
So indeed I agree I should tweak the commit message.

ejgallego:
Umm, I think I want to think more on this one; maybe I should go further beyond and finish implementing proper handling of group options. It would be a lost of time if I have to change the interface again.

ppedrot (approved):
I am not sure this is the ideal API, but at least it looks better than what we have currently.

herbelin:
I'm curious. What would typically be a more ideal API?

gares (approved):

ppedrot:
@herbelin You should be able to also modify the 'reference' if you are in the right context, i.e. once we get a pure API you would also get a `state -> bool -> state` function. Maybe this can be made forward-compatible by returning a record with only one field for now...

herbelin:
>  You should be able to also modify the 'reference' if you are in the right context, i.e. once we get a pure API you would also get a `state -> bool -> state` function. Maybe this can be made forward-compatible by returning a record with only one field for now...

Yes, indeed, at some time, it should be that. What is unclear to me though is how to stepwise implement the transition. Threading the state of options in one go is maybe too much work, and we also have to give time to plugins for the translation. So, in the transitory period, we could maybe have the 3 variants:
- the current `declare_bool_option` (and variants) in deprecation mode
- the intermediate step done in this PR
- the fully-functional step with type `state -> bool -> state`

But, in the latter case, what shall be the definition of `state`? Will it be `state = { options : Goptions.option_state ; ... }` as you suggest, will it be `Goptions.option_state`, will it be `option_state = { printing_option_state : printing_option_state ; typeclass_option_state : typeclass_option_state ; ...}`, will it be `printing_option_state` selectively using a group of options depending on functions, will it be `'a Goptions.group_option_state` using a dynamic tag to enforce the `'a`?

I guess that in the transitory period, until everything is made functional, we shall need functions `save_state`/`get_state` or `save_option_state`/`get_option_state` to temporarily switch from the imperative to the functional view?

Is that what you have in mind?

ppedrot:
I'd personally be in favour of merging this PR as-is. Other improvements can be done afterwards, but letting this PR bitrot is a no-no. Otherwise, we should close it. @ejgallego ?

ejgallego:
I wouldn't say this PR will bitrot, IMHO it needs a further revision, unfortunately I've been quite busy so the slot for this fell back a bit behind.

ejgallego:
Indeed no point in delaying this more; let's merge and we can improve more later.

Issue #8052, opened 2018-07-12
dune improvements meta issue


ejgallego:
This issue documents the roadmap so Dune can be considered as the primary build system.

### TODO
- `configure.ml` refactoring #7149 ;
- `coq_config` refactoring ;
- `coqdep` refactoring [many PRs merged already, more needed]
- documentation targets #7155 #8503 ;
- native compute #8860 
- automatic version generation ocaml/dune#880, #483 ;
- `topkg` / `dune-release` / OPAM support; https://github.com/ocaml/dune/issues/1112 https://github.com/ocaml/dune/issues/1190 #17362 

### Done:

- support Dune's `recurse_subdirs` / `include_subdirs`;
- windows build #8421 ; tho test suite and paths are giving us headaches.
- c̶o̶q̶f̶i̶n̶d̶ ̶#̶7̶3̶1̶2̶ ̶
- f̶e̶e̶d̶b̶a̶c̶k̶ ̶t̶o̶ ̶t̶h̶e̶ ̶D̶u̶n̶e̶ ̶d̶e̶v̶e̶l̶o̶p̶e̶r̶s̶ ̶[̶m̶a̶i̶n̶l̶y̶ ̶t̶o̶ ̶e̶n̶a̶b̶l̶e̶ ̶o̶u̶r̶ ̶.̶v̶ ̶-̶>̶ ̶.̶v̶o̶ ̶r̶u̶l̶e̶s̶]̶ ̶o̶c̶a̶m̶l̶/̶d̶u̶n̶e̶#̶1̶4̶4̶6̶
- t̶e̶s̶t̶ ̶s̶u̶i̶t̶e̶ ̶t̶a̶r̶g̶e̶t̶ ̶#̶8̶5̶6̶6̶ ̶;̶
- P̶r̶o̶j̶e̶c̶t̶ ̶b̶u̶i̶l̶d̶i̶n̶g̶ ̶/̶ ̶`̶c̶o̶q̶_̶d̶u̶n̶e̶`̶ ̶f̶i̶l̶e̶ ̶s̶p̶e̶c̶i̶f̶i̶c̶a̶t̶i̶o̶n̶.̶

### Related Issues:
- #6137 : too many files at root.

### Related bugs that Dune may improve:

- #6921
- #7208 
- #6354
-  #8405
-  #7476
-  #8362
-  #8013
-  #7732
-  #7480
-  #6255
-  #7611
-  #7722
-  #6137
-  #6006
-  #5632
[...]

PR #8729, opened 2018-10-13
[build] [dune] Switch to a Dune-based build system.

ejgallego:
Over the last months we have added support to build Coq with Dune, I
feel we have reached a point where we can start to discuss if we
should replace the make-based system with a Dune-based one.

The *main rationale* for the switch is that maintaining two build
systems is hard, and Dune seems superior to make in almost every
aspect.

Indeed, I think it is going to be hard to find technical arguments to
defend the make-base system. Dune outperforms it on almost every
possible metric.

Additionally, the make-based system has grew organically over the
years, and these days we are spending significant developer resources
on it. The number of bugs that would be fixed by Dune is large [see #8052].
It is not a coincidence that few large projects do use `make` anymore,
but most have moved to `CMake`, `Ninja`, or some other alternative.

On this side, Dune provides a well-defined model that seems to git
Coq's present and future necessities well.

*Blockers*: the single blocker for a merge as of today, apart from the
depending PRs, is the lack of native-compute support. This is due to a
technical limitation wrt targets and can be solved in several
different ways. Also, it is likely that a smooth developer profile is
not possible until https://github.com/ocaml/dune/issues/1155 lands in
Dune itself. https://github.com/ocaml/dune/issues/1377 may be
convenient for the reference manual but we can have a small workaround
to generate the install file for now.

*Risk analysis*: it is important to understand the risks that such a
move would entail. After Dune support is merged we could always go
back to a make-based system, however we are very likely to depend on
Dune-specific features that would be hard to replicate. The main risks are:

- lock-in: indeed lock-in risk is significant and Coq's source code
  may depend on some Dune features. On the other hand, Dune is
  strongly poised to be the standard build tool for the OCaml platform
  and ecosystem, and more than 50% of OPAM packages use it.

  It is safe to say that if Dune would become unsupported in the
  future, there would be more important problems than Coq itself.

- lack of in-house knowledge: indeed, Dune requires some specific
  training in order to understand its build rules, which may become a
  problem. This risk is mitigated in 2 different ways and attenuated
  by an additional consideration: first, Dune rules are fairly simple
  and declarative and it is reasonable to expect developers get to
  know them without too much effort. This is one of the reasons for
  the high adoption numbers in the ecosystem. Second, Dune developers
  are very reactive and care about Coq, thus it is safe to assume that
  we would get external help if needed. The additional consideration
  that may make this less of a problem is that our in-house knowledge
  of make may be even worse than Dune's. A non-negligible number of
  developers have expressed discomfort with make and the amount of
  required knowledge to master make seems way higher than what you
  need to use Dune.

- lack of flexibility: indeed, Dune is way less flexible than
  make. Dune only knows how to do well two things: building OCaml
  executables and libraries. However, that is basically 99% of what
  building Coq entails; building other parts [documentation or Coq
  libraries] is fairly straightforward and seem to pose not a problem.

- lack of maturity: indeed, Dune develops fast, it is not free of
  bugs, and some amount of adaptation is expected from us. This risk
  can only be mitigated if there is continued developer interest.  In
  the worst case, Coq could become stuck in a particular Dune version.

- bootstrapping: this is not a problem as witnessed by #8615, Dune can
  be bootstrapped very easily as it depends only on OCaml and it is
  designed to do so.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -59,10 +59,28 @@ before_script:
 # - use artifacts between jobs                       #
 ######################################################
 
-# TODO figure out how to build doc for installed Coq
+# Template for building Coq + stdlib, typical use: overload the switch
 .build-template:
   stage: stage-1
   interruptible: true
+  dependencies: []
+  script:
+    - set -e
+    - make world
+    - set +e
+    - tar cfj _build.tar.bz2 _build
+  variables:
+    OPAM_SWITCH: base
+  artifacts:
+    name: "$CI_JOB_NAME"
+    when: always
+    paths:
+      - _build/log
+      - _build.tar.bz2
+    expire_in: 1 week
+
+.build-make-template:
+  stage: stage-1
   artifacts:
     name: "$CI_JOB_NAME"
     paths:
@@ -75,89 +93,31 @@ before_script:
     - set -e
 
     - echo 'start:coq.clean'
-    - make clean # ensure that `make clean` works on a fresh clone
+    - make -f Makefile.make clean # ensure that `make clean` works on a fresh clone
     - echo 'end:coq.clean'
 
     - echo 'start:coq.config'
-    - ./configure -warn-error yes -prefix "$(pwd)/_install_ci" ${COQ_EXTRA_CONF}"$COQ_EXTRA_CONF_QUOTE"
+    - ./configure -warn-error yes -prefix "$(pwd)/_install_ci"
     - echo 'end:coq.config'
 
     - echo 'start:coq.build'
-    - make -j "$NJOBS" byte
-    - make -j "$NJOBS" world $EXTRA_TARGET
-    - make test-suite/misc/universes/all_stdlib.v
+    - make -f Makefile.make -j "$NJOBS" byte
+    - make -f Makefile.make -j "$NJOBS" world $EXTRA_TARGET
+    - make -f Makefile.make test-suite/misc/universes/all_stdlib.v
     - echo 'end:coq:build'
 
     - echo 'start:coq.install'
-    - make install install-byte $EXTRA_INSTALL
-    - make install-byte
+    - make -f Makefile.make install install-byte $EXTRA_INSTALL
+    - make -f Makefile.make install-byte
     - cp bin/fake_ide _install_ci/bin/
     - echo 'end:coq.install'
 
     - set +e
 
-# Template for building Coq + stdlib, typical use: overload the switch
-.dune-template:
-  stage: stage-1
-  interruptible: true
-  dependencies: []
-  script:
-    - set -e
-    - make -f Makefile.dune world
-    - set +e
-    - tar cfj _build.tar.bz2 _build
-  variables:
-    OPAM_SWITCH: edge
-    OPAM_VARIANT: "+flambda"
-  artifacts:
-    name: "$CI_JOB_NAME"
-    when: always
-    paths:
-      - _build/log
-      - _build.tar.bz2
-    expire_in: 1 week
-
-.dune-ci-template:
-  stage: stage-2
-  interruptible: true
-  needs:
-    - build:edge+flambda:dune:dev
-  dependencies:
-    - build:edge+flambda:dune:dev
-  script:
-    - tar xfj _build.tar.bz2
-    - set -e
-    - echo 'start:coq.test'
-    - make -f Makefile.dune "$DUNE_TARGET"
-    - echo 'end:coq.test'
-    - set +e
-  variables:
-    OPAM_SWITCH: edge
-    OPAM_VARIANT: "+flambda"
-  artifacts:
-    when: always
-    name: "$CI_JOB_NAME"
-    expire_in: 2 months
-
 # every non build job must set dependencies otherwise all build
 # artifacts are used together and we may get some random Coq. To that
 # purpose, we add a spurious dependency `not-a-real-job` that must be
 # overridden otherwise the CI will fail.
-
-.doc-template:
-  stage: stage-2
-  interruptible: true
-  dependencies:
-    - not-a-real-job
-  script:
-    - SPHINXENV='COQBIN="'"$PWD"'/_install_ci/bin/"'
-    - make -j "$NJOBS" SPHINXENV="$SPHINXENV" SPHINX_DEPS= refman
-    - make install-doc-sphinx
-  artifacts:
-    name: "$CI_JOB_NAME"
-    paths:
-      - _install_ci/share/doc/coq/
-    expire_in: 2 months
 
 # set dependencies when using
 .test-suite-template:
@@ -166,18 +126,12 @@ before_script:
   dependencies:
     - not-a-real-job
   script:
-    - cd test-suite
-    - make clean
-    # careful with the ending /
-    - BIN=$(readlink -f ../_install_ci/bin)/
-    - LIB=$(readlink -f ../_install_ci/lib/coq)/
-    - export OCAMLPATH=$(readlink -f ../_install_ci/lib/):"$OCAMLPATH"
-    - make -j "$NJOBS" BIN="$BIN" COQLIB="$LIB" COQFLAGS="${COQFLAGS}" all
+    - make test-suite
   artifacts:
     name: "$CI_JOB_NAME.logs"
     when: on_failure
     paths:
-      - test-suite/logs
+      - _build/default/test-suite/logs
     # Gitlab doesn't support yet "expire_in: never" so we use the instance default
     # expire_in: never
 
@@ -188,7 +142,8 @@ before_script:
   dependencies:
     - not-a-real-job
   script:
-    - cd _install_ci
+    - tar xfj _build.tar.bz2
+    - cd _build/install/default
     - find lib/coq/ -name '*.vo' -fprint0 vofiles
     - xargs -0 --arg-file=vofiles bin/coqchk -o -m -coqlib lib/coq/ > ../coqchk.log 2>&1 || touch coqchk.failed
     - tail -n 1000 ../coqchk.log # the log is too big for gitlab so pipe to a file and display the tail
@@ -199,13 +154,15 @@ before_script:
       - coqchk.log
     expire_in: 2 months
 
+# Core CI template
 .ci-template:
   stage: stage-2
   interruptible: true
   script:
+    - tar xfj _build.tar.bz2
     - set -e
     - echo 'start:coq.test'
-    - make -f Makefile.ci -j "$NJOBS" "${CI_JOB_NAME#*:}"
+    - make -j "$NJOBS" "${CI_JOB_NAME#*:}"
     - echo 'end:coq.test'
     - set +e
   needs:
@@ -253,17 +210,19 @@ before_script:
     - git config --global user.name "coqbot"
     - git config --global user.email "coqbot@users.noreply.github.com"
 
-build:base:
-  extends: .build-template
+build:make:base:
+  extends: .build-make-template
   variables:
     COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
-    # coqdoc for stdlib, until we know how to build it from installed Coq
-    EXTRA_TARGET: "doc-stdlib"
-    EXTRA_INSTALL: "install-doc-stdlib-html install-doc-printable"
+
+build:base:
+  extends: .build-template
 
 # no coqide for 32bit: libgtk installation problems
 build:base+32bit:
   extends: .build-template
+  script:
+    - make coq coqide-server
   variables:
     OPAM_VARIANT: "+32bit"
     COQ_EXTRA_CONF: "-native-compiler yes"
@@ -276,17 +235,10 @@ build:edge+flambda:
   variables:
     OPAM_SWITCH: edge
     OPAM_VARIANT: "+flambda"
-    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt -flambda-opts "
-    COQ_EXTRA_CONF_QUOTE: "-O3 -unbox-closures"
-
-build:edge+flambda:dune:dev:
-  extends: .dune-template
 
 build:base+async:
-  extends: .build-template
-  stage: stage-1
+  extends: .build-make-template
   variables:
-    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
     COQUSERFLAGS: "-async-proofs on"
   timeout: 100m
   allow_failure: true # See https://github.com/coq/coq/issues/9658
@@ -295,7 +247,7 @@ build:base+async:
       - $UNRELIABLE =~ /enabled/
 
 build:quick:
-  extends: .build-template
+  extends: .build-make-template
   variables:
     COQ_EXTRA_CONF: "-native-compiler no"
     QUICK: "1"
@@ -366,6 +318,7 @@ pkg:opam:
     # because there is no --extra-trusted-public-key option.
     - nix-build -E "import (fetchTarball $CI_PROJECT_URL/-/archive/$CI_COMMIT_SHA.tar.gz) {}" -K --extra-substituters "$EXTRA_SUBSTITUTERS" --trusted-public-keys "$NIXOS_PUBLIC_KEY $EXTRA_PUBLIC_KEYS" | if [ ! -z "$CACHIX_SIGNING_KEY" ]; then cachix push coq; fi
   artifacts:
+    expire_in: 1 week
     name: "$CI_JOB_NAME.logs"
     when: on_failure
     paths:
@@ -411,29 +364,31 @@ pkg:nix:
     - master
     - /^v.*\..*$/
 
-doc:refman:
-  extends: .doc-template
-  dependencies:
-    - build:base
-  needs:
-    - build:base
-
-doc:refman:dune:
-  extends: .dune-ci-template
-  variables:
-    DUNE_TARGET: refman-html
+doc:apidoc:
+  extends: .ci-template
   artifacts:
+    when: on_success
+    name: "$CI_JOB_NAME"
+    expire_in: 1 week
     paths:
-      - _build/log
+      - _build/default/_doc/
+
+doc:refman-html:
+  extends: .ci-template
+  artifacts:
+    when: on_success
+    name: "$CI_JOB_NAME"
+    expire_in: 2 months
+    paths:
       - _build/default/doc/sphinx_build/html
 
-doc:stdlib:dune:
-  extends: .dune-ci-template
-  variables:
-    DUNE_TARGET: stdlib-html
+doc:stdlib-html:
+  extends: .ci-template
   artifacts:
+    when: on_success
+    name: "$CI_JOB_NAME"
+    expire_in: 2 months
     paths:
-      - _build/log
       - _build/default/doc/stdlib/html
 
 doc:refman:deploy:
@@ -445,13 +400,13 @@ doc:refman:deploy:
     variables:
       - $DOCUMENTATION_DEPLOY_KEY
   dependencies:
-    - doc:ml-api:odoc
-    - doc:refman:dune
-    - doc:stdlib:dune
+    - doc:apidoc
+    - doc:refman-html
+    - doc:stdlib-html
   needs:
-    - doc:ml-api:odoc
-    - doc:refman:dune
-    - doc:stdlib:dune
+    - doc:apidoc
+    - doc:refman-html
+    - doc:stdlib-html
   script:
     - echo "$DOCUMENTATION_DEPLOY_KEY" | tr -d '\r' | ssh-add - > /dev/null
     - git clone git@github.com:coq/doc.git _deploy
@@ -466,15 +421,6 @@ doc:refman:deploy:
     - git add api refman stdlib
     - git commit -m "Documentation of branch “$CI_COMMIT_REF_NAME” at $CI_COMMIT_SHORT_SHA"
     - git push # TODO: rebase and retry on failure
-
-doc:ml-api:odoc:
-  extends: .dune-ci-template
-  variables:
-    DUNE_TARGET: apidoc
-  artifacts:
-    paths:
-      - _build/log
-      - _build/default/_doc/
 
 test-suite:base:
   extends: .test-suite-template
@@ -503,35 +449,26 @@ test-suite:edge+flambda:
     OPAM_SWITCH: edge
     OPAM_VARIANT: "+flambda"
   only: *full-ci
-
-test-suite:edge:dune:dev:
-  stage: stage-2
-  dependencies:
-    - build:edge+flambda:dune:dev
-  needs:
-    - build:edge+flambda:dune:dev
-  script:
-    - tar xfj _build.tar.bz2
-    - make -f Makefile.dune test-suite
-  variables:
-    OPAM_SWITCH: edge
-    OPAM_VARIANT: "+flambda"
   artifacts:
     name: "$CI_JOB_NAME.logs"
     when: on_failure
     paths:
+      - _build/log
       - _build/default/test-suite/logs
     # Gitlab doesn't support yet "expire_in: never" so we use the instance default
     # expire_in: never
+    expire_in: 1 week
 
 test-suite:base+async:
-  extends: .test-suite-template
+  stage: stage-2
   dependencies:
-    - build:base
+    - build:make:base
   needs:
-    - build:base
+    - build:make:base
   variables:
     COQFLAGS: "-async-proofs on -async-proofs-cache force"
+  script:
+    - make -f Makefile.make test-suite
   timeout: 100m
   allow_failure: true
   only:
@@ -733,13 +670,6 @@ plugin:ci-paramcoq:
 
 plugin:ci-perennial:
   extends: .ci-template-flambda
-
-plugin:plugin-tutorial:
-  stage: stage-1
-  dependencies: []
-  script:
-    - ./configure -local -warn-error yes
-    - make -j "$NJOBS" plugin-tutorial
 
 plugin:ci-quickchick:
   extends: .ci-template-flambda
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -14,7 +14,7 @@ To compile Coq yourself, you need:
 
 - The [findlib](http://projects.camlcity.org/projects/findlib.html) library (version >= 1.8.0)
 
-- GNU Make (version >= 3.81)
+- [Dune](https://github.com/ocaml/dune) >= 2.4 _or_ GNU Make (version >= 3.81)
 
 - a C compiler
 
@@ -69,11 +69,14 @@ for more details.
 Build and Installation Procedure
 --------------------------------
 
-Coq offers the choice of two build systems, an experimental one based
-on [Dune](https://github.com/ocaml/dune), and the standard
-makefile-based one.
+Coq offers the choice of two build systems, one based on
+[Dune](https://github.com/ocaml/dune), the standard OCaml build tool,
+and one based on make. The default and recommended in the Dune-based
+one.
 
-Please see [INSTALL.make.md](dev/doc/INSTALL.make.md) for build and
-installation instructions using `make`. If you wish to experiment with
-the Dune-based system see the [dune guide for
-developers](dev/doc/build-system.dune.md).
+Please see [INSTALL.dune.md](dev/doc/INSTALL.dune.md) for build and
+installation instructions using `dune`. Also see the [dune guide for
+developers](dev/doc/build-system.dune).
+
+Instructions to install Coq with make can be found in
+[INSTALL.make.md](dev/doc/INSTALL.make.md).
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@
 ##########################################################################
 
 # The default build system is make-based one.
-ifndef COQ_USE_DUNE
+ifdef COQ_USE_MAKE
 include Makefile.make
 else
 include Makefile.dune
--- a/Makefile.dune
+++ b/Makefile.dune
@@ -116,6 +116,8 @@ ireport:
 clean:
 	dune clean
 
+include Makefile.ci
+
 # Other common dev targets:
 #
 # dune build coq.install
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -68,7 +68,7 @@ jobs:
       opam switch set ocaml-base-compiler.$COMPILER
       eval $(opam env)
       opam update
-      opam install -j "$NJOBS" num ocamlfind${FINDLIB_VER} ounit lablgtk3-sourceview3
+      opam install -j "$NJOBS" num ocamlfind${FINDLIB_VER} ounit dune lablgtk3-sourceview3
       opam list
     displayName: 'Install OCaml dependencies'
     env:
@@ -76,18 +76,17 @@ jobs:
       FINDLIB_VER: ".1.8.1"
       OPAMYES: "true"
 
+  # We don't install gtk for now
   - script: |
       set -e
-
       eval $(opam env)
       ./configure -prefix '$(Build.BinariesDirectory)' -warn-error yes -native-compiler no -coqide opt
-      make -j "$NJOBS" byte
-      make -j "$NJOBS"
+      make coq coqide-server coqide
     displayName: 'Build Coq'
 
   - script: |
       eval $(opam env)
-      make -j "$NJOBS" test-suite PRINT_LOGS=1
+      make NJOBS="$NJOBS" test-suite
     displayName: 'Run Coq Test Suite'
 
   - script: |
--- a/default.nix
+++ b/default.nix
@@ -84,10 +84,11 @@ stdenv.mkDerivation rec {
 
   enableParallelBuilding = true;
 
-  buildFlags = [ "world" "byte" ] ++ optional buildDoc "doc-html";
+  buildFlags = [ "world" ] ++ optional buildDoc "refman-html";
 
-  installTargets =
-    [ "install" "install-byte" ] ++ optional buildDoc "install-doc-html";
+  # Nix Dune:
+  # it should use "dune-install pkg" [opam-install for the refman]
+  # installTargets = [ "install" ] ++ optional buildDoc "install-html";
 
   createFindlibDestdir = !shell;
 
--- a/dev/build/osx/make-macos-dmg.sh
+++ b/dev/build/osx/make-macos-dmg.sh
@@ -9,10 +9,10 @@ VERSION=$(sed -n -e '/^let coq_version/ s/^[^"]*"\([^"]*\)"$/\1/p' configure.ml)
 APP=bin/CoqIDE_${VERSION}.app
 
 # Create a .app file with CoqIDE, without signing it
-make PRIVATEBINARIES="$APP" -j "$NJOBS" -l2 "$APP"
+make -f Makefile.make PRIVATEBINARIES="$APP" -j "$NJOBS" -l2 "$APP"
 
 # Add Coq to the .app file
-make OLDROOT="$OUTDIR" COQINSTALLPREFIX="$APP/Contents/Resources" install-coq install-ide-toploop
+make -f Makefile.make OLDROOT="$OUTDIR" COQINSTALLPREFIX="$APP/Contents/Resources/" install-coq install-ide-toploop
 
 # Create the dmg bundle
 mkdir -p "$DMGDIR"
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -1357,17 +1357,17 @@ function make_coq {
 
     # 8.4x doesn't support parallel make
     if [[ $COQ_VERSION == 8.4* ]] ; then
-      log1 make
+      log1 make -f Makefile.make
     else
       # shellcheck disable=SC2086
-      log1 make $MAKE_OPT
+      log1 make -f Makefile.make $MAKE_OPT
     fi
 
     if [ "$INSTALLMODE" == "relocatable" ]; then
       logn reconfigure ./configure -with-doc no -prefix "$PREFIXCOQ" -libdir "$PREFIXCOQ/lib/coq" -mandir "$PREFIXCOQ/man"
     fi
 
-    log2 make install
+    log2 make -f Makefile.make install
     log1 copy_coq_dlls
     log1 copy_coq_gtk
 
--- a/dev/ci/ci-common.sh
+++ b/dev/ci/ci-common.sh
@@ -8,9 +8,10 @@ export NJOBS
 
 if [ -n "${GITLAB_CI}" ];
 then
-    # Gitlab build, Coq installed into `_install_ci`
-    export OCAMLPATH="$PWD/_install_ci/lib:$OCAMLPATH"
-    export COQBIN="$PWD/_install_ci/bin"
+    # Gitlab build with Dune
+    export OCAMLPATH="$PWD/_build/install/default/lib/"
+    export COQBIN="$PWD/_build/install/default/bin"
+    export COQLIB="$PWD/_build/install/default/lib/coq"
     export CI_BRANCH="$CI_COMMIT_REF_NAME"
     if [[ ${CI_BRANCH#pr-} =~ ^[0-9]*$ ]]
     then
--- a/dev/ci/user-overlays/08729-ejgallego-dune+kill_make.sh
+++ b/dev/ci/user-overlays/08729-ejgallego-dune+kill_make.sh
@@ -1,0 +1,14 @@
+if [ "$CI_PULL_REQUEST" = "8729" ] || [ "$CI_BRANCH" = "dune+kill_make" ]; then
+
+    true
+
+    # ltac2_CI_REF=rm-section-path
+    # ltac2_CI_GITURL=https://github.com/maximedenes/ltac2
+
+    # Elpi_CI_REF=fix_ltac_packing
+    # Elpi_CI_GITURL=https://github.com/ejgallego/coq-elpi
+
+    # Equations_CI_REF=vernac+monify_hook
+    # Equations_CI_GITURL=https://github.com/ejgallego/Coq-Equations
+
+fi
--- a/dev/doc/INSTALL.dune.md
+++ b/dev/doc/INSTALL.dune.md
@@ -1,0 +1,20 @@
+# Installing with Dune
+
+The build and installation procedure using Dune is quite
+straightforward. We recommend you read Dune's user manual for
+questions, as Coq mostly implements a standard build workflow.
+
+## Quick Installation Procedure [Dune].
+=======================================
+
+1. COQ_CONFIGURE_PREFIX=$prefix make release
+2. dune install [--prefix=$prefix] coq
+
+If you don't pass the prefix argument to `dune install` it will choose
+your default OCaml library path.
+
+## Developer setup.
+===================
+
+For more information about the Dune build system just type `make`, you
+can also read `dev/doc/build-system.dune.md`
--- a/dev/doc/INSTALL.make.md
+++ b/dev/doc/INSTALL.make.md
@@ -2,8 +2,8 @@ Quick Installation Procedure using Make.
 ----------------------------------------
 
     $ ./configure
-    $ make
-    $ make install (you may need superuser rights)
+    $ make -f Makefile.make
+    $ make -f Makefile.make install (you may need superuser rights)
 
 Detailed Installation Procedure.
 --------------------------------
@@ -92,7 +92,7 @@ Detailed Installation Procedure.
 
 4. Still in the root directory, do
 
-        make
+        make -f Makefile.make
 
    to compile Coq in the best OCaml mode available (native-code if supported,
    bytecode otherwise).
@@ -115,17 +115,17 @@ Detailed Installation Procedure.
    defined at configuration time (step 3). Just do
 
         umask 022
-        make install
+        make -f Makefile.make install
 
    Of course, you may need superuser rights to do that.
 
 6. Optionally, you could build the bytecode version of Coq via:
 
-        make byte
+        make -f Makefile.make byte
 
    and install it via
 
-        make install-byte
+        make -f Makefile.make install-byte
 
   This version is much slower than the native code version of Coq, but could
   be helpful for debugging purposes. In particular, coqtop.byte embeds an OCaml
@@ -133,7 +133,7 @@ Detailed Installation Procedure.
 
 7. You can now clean all the sources. (You can even erase them.)
 
-        make clean
+        make -f Makefile.make clean
 
 Installation Procedure For Plugin Developers.
 ---------------------------------------------
@@ -206,7 +206,7 @@ So, in order to compile Coq for a new architecture, proceed as follows:
 * Omit step 7 above and clean only the architecture dependent files:
   it is done automatically with the command
 
-        make archclean
+        make -f Makefile.make archclean
 
 * Configure the system for the new architecture:
 
--- a/dev/doc/build-system.dune.md
+++ b/dev/doc/build-system.dune.md
@@ -220,6 +220,11 @@ useful to Coq, some examples are:
 - Automatic Generation of OPAM files.
 - Multi-directory libraries.
 
+## Tweaking `dune` files
+
+As a developer, you should not have to deal with `dune` configuration
+files on a regular basis, unless adding a new library or plugin.
+
 ## FAQ
 
 - I get "Error: Dynlink error: Interface mismatch":
--- a/tools/coqdep_common.ml
+++ b/tools/coqdep_common.ml
@@ -400,6 +400,8 @@ let add_caml_known phys_dir _ f =
   match suff with
     | ".mllib" -> add_mllib_known basename (Some phys_dir) suff
     | ".mlpack" -> add_mlpack_known basename (Some phys_dir) suff
+    (* Must add support to remove the mlpack files *)
+    (* | ".dune" -> add_mlpack_known basename (Some phys_dir) suff *)
     | _ -> ()
 
 let add_coqlib_known recur phys_dir log_dir f =


gares:
My take is that we Devs that did now work on the dune patch set have to test it during our daily activity in order to make up our mind. For this to happen the two build systems have to coexist for some time, eg a release cycle.

.gitlab-ci.yml
```diff
@@ -58,93 +58,40 @@ after_script:
   artifacts:
     name: "$CI_JOB_NAME"
     paths:
-      - _install_ci
-      - config/Makefile
-      - config/coq_config.py
-      - test-suite/misc/universes/all_stdlib.v
+      - _build
```
SkySkimmer: Doesn't this contain more than we install? We want to test install so that seems wrong.

.gitlab-ci.yml
```diff
@@ -191,38 +138,29 @@ after_script:
...
-    # coqdoc for stdlib, until we know how to build it from installed Coq
-    EXTRA_TARGET: "stdlib"
 
 # no coqide for 32bit: libgtk installation problems
 build:base+32bit:
   <<: *build-template
+  script:
+    - make voboot
+    - dune build coq.install
+    - dune build coqide-server.install
```
SkySkimmer: Does dune fail when doing make world if gtk isn't installed?

SkySkimmer (changes requested):
Why are the mlpack files empty but retained?

How do I do these with dune:
make coqocaml
make byte
make arith
make validate
make printers

use ocamldebug (dev/ocamldebug-coq is generated by configure)
?

.gitlab-ci.yml
ejgallego: As of today only libraries can be optional, so yes it complains it cannot build `coqide` as `world` is an alias today for "all the packages". Note that this is not a fundamental problem as we could well indeed do as we do here and only build the relevant packages.

Soon, libraries and even packages will be allowed to be declared optional so indeed the error will be just a warning when building all the packages.

ejgallego:
> Why are the mlpack files empty but retained?

This is due to `coqdep` needing fixing.

> How do I do these with dune:
> make coqocaml
> make byte
> make arith
> make validate
> make printers

For most of them you just do `dune $TARGET`, where `$TARGET` should be the right files. This will need a bit of tuning but nothing out of the ordinary.

> use ocamldebug (dev/ocamldebug-coq is generated by configure)

The file is still there in `_build/default/dev`.

.gitlab-ci.yml
SkySkimmer: Install isn't tested by the test suite, it's tested by the various 3rd party developments.

ejgallego:
> > How do I do these with dune:
> > make coqocaml
> > make byte
> > make arith
> > make validate
> > make printers

By the way, where are these documented?

SkySkimmer:
>For most of them you just do dune build $TARGET, where $TARGET should be the right files. This will need a bit of tuning but nothing out of the ordinary.

How's that supposed to work for `validate`? Also I'm pretty sure dune doesn't know how to build the debug printers, and I don't want to manually list the plugins to build them (which make byte/coqocaml do for me).

>By the way, where are these documented?

Nowhere.

ejgallego:
> How's that supposed to work for `validate`?

Like it works on the CI today? I dunno, there are so many ways to do it, including making Dune aware of validation.

> Also I'm pretty sure dune doesn't know how to build the debug printers, and I don't want to manually list the plugins to build them (which make byte/coqocaml do for me).

??? Of course Dune can build the debug printers, and in fact it can properly track the dependency of a debug session to them unlike the build system today. [Which by the way requires listing all modules in an `.mllib` file ironically] And by the way you had to manually list the printers in make too `DEBUGPRINTERS`

maximedenes (changes requested):
I'm not against the move in itself, but obviously you need to give developers a longer transition period (like one release cycle, as @gares mentioned). For now, the only change I ask if about the title of this PR. It should be something like "Remove Makefile-based build system". With the current title I thought it was only about switching the various builds we test to Dune.

ejgallego:
> Nowhere.

So here is another advantage the Dune-based setup has as it is better documented as of today.

ejgallego:
> but obviously you need to give developers a longer transition period

Why? It is not so obvious to me. In fact, you could switch today and few people would even notice, as they would keep doing the same `make $foo` where foo is a set of 4 target that could be / are emulated by the Dune makefile.

About the title, well, no point in switching and keeping the old thing around, untested, and full of bugs. What would be the point? The idea is that this PR will close a large number of currently opened bugs.

SkySkimmer:
>Like it works on the CI today?

Have you looked at what the CI does?
~~~yaml
.validate-template: &validate-template
  stage: test
  dependencies:
    - not-a-real-job
  script:
    - cd _install_ci
    - find lib/coq/ -name '*.vo' -print0 > vofiles
    - for regexp in 's/.vo//' 's:lib/coq/plugins:Coq:' 's:lib/coq/theories:Coq:' 's:/:.:g'; do sed -z -i "$regexp" vofiles; done
    - xargs -0 --arg-file=vofiles bin/coqchk -boot -silent -o -m -coqlib lib/coq/
~~~

>Why? It is not so obvious to me. In fact, you could switch today and few people would even notice, as they would keep doing the same make $foo where foo is a set of 4 target that could be / are emulated by the Dune makefile.

`make byte` `make coqocaml` `make world` `make test-suite`?
Except dune only knows 2 of those.

>??? Of course Dune can build the debug printers, and in fact it can properly track the dependency of a debug session to them unlike the build system today. [Which by the way requires listing all modules in an .mllib file ironically] 

So what's the command? `dune build dev/top_printers.cmo` (that doesn't work btw)?

>And by the way you had to manually list the printers in make too DEBUGPRINTERS

The point is that the dune system must do that listing too.

maximedenes:
> About the title, well, no point in switching and keeping the old thing around, untested, and full of bugs. What would be the point? The idea is that this PR will close a large number of currently opened bugs.

I understand your argument, so please make the title reflect what this PR does. Thank you.

ejgallego:
> Have you looked at what the CI does?

Yup, a re-implementation of `Makefile.common`, we have a few. Anyways, validate should really be part of test suite and in that sense you can do an automatic test for Coq lib. So in Dune it would work the same, Dune doesn't handle testing other than calling a particular test tool.

> `make byte` `make coqocaml` `make world` `make test-suite`?
> Except dune only knows 2 of those.

You got `states` there and a few others. I see what you mean now about `coqocaml` and `byte`, there are not here yet but they are fairly trivial to add, same for the dev printers. Should be added before the merge of course, my point is that at this moment these targets are basically nits.

I need more info about the dev printers, where are they loaded from? I thought the test suite was loading them but doesn't require its compilation it seems. A quick way to do it is just to add a `dune` file to `dev` [by the way the dev printers should be put in a separate dir].

> So what's the command? `dune build dev/top_printers.cmo` (that doesn't work btw)?

Indeed there is no `dune` file in `dev` so that folder doesn't exists for Dune [nor it is included from a recursive rule. That would work once you declare the dev_printers to be known to Dune.

> The point is that the dune system must do that listing too.

Well you could follow certain conventions and let recursive handling of dirs take care of it.

ejgallego:
> With the current title I thought it was only about switching the various builds we test to Dune.

It is like 99% of what the PR really does, so I think the title reflects better what is going on. Removing the Make files or not is just a detail IMHO [I could keep them and the PR would be the same].

maximedenes:
> It is like 99% of what the PR really does, so I think the title reflects better what is going on. Removing the Make files or not is just a detail IMHO [I could keep them and the PR would be the same].

I won't argue much more on this, but I suspect some developers might be interested in knowing that incidentally, this PR also removes the build system they use currently...

gares (changes requested):
As of today I can build Coq with make and I can call a bunch of targets like the ones mentioned by Gaetan. I can try to replace them with calls to dune in a progressive fashion, in my daily work, and open issues if some work flow is not supported. I don't want, from a day to another, be unable to build Coq as I'm used to. In case it is not clear, this PR removes the support for make (as in removing the Makefile.* files).

I think we should not merge this before we all managed to switch their workflow to the new build system. I also understand that putting no deadline does not work in practice. So I propose 1 release cycle. If at "freeze" time for 8.10 we are all happy with the new build system then let's do the switch.

ejgallego:
> I think the test suite loads the debug stuff for `Drop.` not the ones for ocamldebug.
> The usage is
> 
> ```shell
> $ dev/ocamldebug-coq bin/coqtop.byte # args...
> source db
> run / break / etc
> ```
> 
> where `db` is https://github.com/coq/coq/blob/master/dev/db

Thanks Gaëtan, first, am I correct in that having configure write this file seems like overkill? [The file works out of the box if called directly]. but indeed that setup is easy to fix.

mattam82:
Agreed with @gares and @maximedenes here, let’s have a transition phase until 8.10.

SkySkimmer:
>Thanks Gaëtan, first, am I correct in that having configure write this file seems like overkill? [The file works out of the box if called directly]. 

`source db` tells me
>Error during code loading: error while linking /home/gaetan/dev/coq/coq/parsing/parsing.cma.
Reference to undefined global `Ploc'

if I replace the generated ocamldebug-coq by the git tracked ocamldebug-coq.run

MSoegtropIMC:
One question on this: the Windows build system for the release installers (afaik) still uses the make based system. I guess this must be switched some time. Otherwise we would have to support 2 make systems. What time frame do you envision for this?

ejgallego:
> What time frame do you envision for this?

As soon as I get back to this, the switch should be pretty easy.

ejgallego:
> if I replace the generated ocamldebug-coq by the git tracked ocamldebug-coq.run

I see, that would be solved by either #8667 or by using ocamlfind to launch the debugger.

ejgallego:
Regarding the terminology I used "Switch" in the title as I consider it less negative than "remove" or "replace", both words that I don't like so much. I could use replace maybe, but I do prefer switch.

Indeed, Coq make system has served people well for many years and it took a large amount of work by top-level OCaml devs to produce something that could be faster.

maximedenes:
> Regarding the terminology I used "Switch" in the title as I consider it less negative than "remove" or "replace", both words that I don't like so much. I could use replace maybe, but I do prefer switch.

Switching in this context would mean making the main CI builds use Dune instead of Make, and have one specific job tracking the Make build, instead of the current Dune one.

You are not choosing a less negative synonym, you are using a title that is misleading. You should either change the title to match the content, or change the content to match the title.

ejgallego:
Regarding when to switch, I think that we should switch "whenever this is ready".

As of today, the Dune setup is obviously not ready yet, however we have reached the point where I don't see any further obstacle implementing the missing targets / bits [in many cases we can provide even better targets than before, like the new `check` that only builds the interfaces + merlin].

The idea of the PR is to see if anymore has arguments for not doing the switch, assuming of course that the Dune build is working properly.

maximedenes:
> "remove" or "replace", both words that I don't like so much

@vbgl often uses "archive" in this context, I guess because the artifact remains in the `git` history anyway. You can use that word instead, if you like it better.

Zimmi48:
I agree that we need some time for developers to confirm that they are working well with Dune only. The only problem I see with this approach is that we should ensure that developers are indeed using Dune in priority. So maybe some partial switch is required indeed before Makefiles are entirely removed. (E.g. `make` calls dune by default and you have to do `make -f Makefile.legacy` to continue using the legacy system.)

ejgallego:
> `make` calls dune by default and you have to do `make -f Makefile.legacy` to continue using the legacy system.)

> Switching in this context would mean making the main CI builds use Dune instead of Make, and have one specific job tracking the Make build, instead of the current Dune one.

Sounds like a sensible plan, once we agree the switch is OK.

Sorry I made this PR about the transition strategy when I meant it to be about whether to do the switch itself.

ejgallego:
> Sorry I made this PR about the transition strategy when I meant it to be about whether to do the switch itself.

I did like this because I thought that having "a demo" of how would Coq look once the switch is done would be a good basis to evaluate the switch itself. But indeed the contents of the PR still have to be refined, but the thing is not by a lot indeed.

Zimmi48:
>What do others think about the title of the PR? Do you find it "misleading"?

I think the title has not such a big importance and both the current one and the one proposed by Maxime are fine.

I am not sure it makes sense to keep this PR opened for several months anyways. It will be easy to recreate it, especially if the default Makefile is the Dune one. It will only be a matter of removing stuff.

ejgallego:
> I am not sure it makes sense to keep this PR opened for several months anyways. It will be easy to recreate it, especially if the default Makefile is the Dune one. It will only be a matter of removing stuff.

A small note, removing the `Makefile` files was important in order to be sure that things do work without Makefiles being there.

I think that I wouldn't close this PR just indeed amend it then as to keep the makefiles but perform the switch [and leave one job for testing, tho I am not sure how useful that is]

The thing is that if such switch is done, I don't think it matters much to keep the makefiles in place, it won't be long until they develop problems, but if people feels better that way so it be.

Zimmi48:
>The thing is that if such switch is done, I don't think it matters much to keep the makefiles in place, it won't be long until they develop problems, but if people feels better that way so it be.

But it won't be long until we notice what are the issues with Dune either and the longer we can do with the legacy Makefile, the less bitrotting becomes important because the less likely we are to switch back.

ejgallego:
Ok I've amended the PR so it doesn't remove the Makefiles, just they are not the default anymore. What tests to keep, that's a good question.

Zimmi48 (dismissed):
Find a way for git to understand the renaming so as to reduce the diff size (maybe rename `Makefile` in a first commit and `Makefile.dune` in a second one). Also, I think that `Makefile.legacy` is much clearer in intent than `Makefile.make` but that's just my own opinion.

Zimmi48:
>For the records, it won't work: git doesn't track individual diffs. It just computes the diff between initial and final commits (and basically counts the number of changes in a file and if it's below a threshold, it counts it as a move (the threshold can be configured, by the way).

I don't care about this. Since we don't squash PRs when we merge them, it will mean that `git log` can be used to give an accurate history of each file.

>I think this is unrealistic given that git doesn't support renaming and all that it has is heuristics that change from one version to the other [in fact Github has better heuristics than magit for example]. The two commit solution would leave a commit without a main makefile. Not a good idea IMHO.

I don't think that's a big problem. You may say "what about bisect scripts" but these should be made to use the same Makefile before and after the switch for better consistency anyways.

Zimmi48:
>I know some people don't like the word "legacy" so that's why I avoided it. Makefile.make is factual.

Indeed, `make` is clearly more factual. The point of `legacy` was to show the intent...

gares:
>  So I propose 1 release cycle. If at "freeze" time for 8.10 we are all happy with the new build system then let's do the switch.

How far are we from this being ready?
The freeze is in april IIRC, and I see quite a few issued on the dune build system still open.

Zimmi48:
@ejgallego recently said about the native Coq support in dune:

>hard to say, I am optimistic about having it done before the 8.10 freeze; Dune releases pretty often so that doesn't worry me a lot, however there is still some work to do to implement things properly, and I don't have a lot of free time.

:worried:

ejgallego:
> How far are we from this being ready?

I dunno what to say, from my point of view I'd have merged this long time ago, Dune provides to me strictly more functionality than make [and some of it essential such as the composed overlay build + check], but this PR doesn't even have an assignee so I am not sure what the criterion for merge would be.

I am happy with the current status quo tho, most of the heavy lifting was done and I can indeed benefit from the new features. Regarding the Dune implementation itself I think almost everything is functional except for #8860 [which could be fixed now if we wanted with a renaming of the output files] and some more needed workflow tweaks as in #8900.

I dunno, so far I have not a clear picture on who's using Dune, etc...

gares:
The question is: can we release with just the dune build system? Can we build packages for all platforms? These are prerequisites for merging the pr imo.

ejgallego:
> The question is: can we release with just the dune build system? Can we build packages for all platforms? These are prerequisites for merging the pr imo.

I think so pretty much; Win, OSX, and Linux do work. Some other very minor stuff missing is:
- generation of `install` files for the refman [can be done with a simple make rule]
- a rule for running `coqdoc`
- the stuff mentioned before

Note that this PR doesn't remove the make-based system, it just switches the default used in the CI to Dune. It is indeed getting smaller and smaller. I will rebase now.

So I dunno, this could happen for 8.11 if we want more polish and nothing would change, other than having to keep the make system for a bit longer.

Zimmi48:
The dune build system is also missing support for building the PDF version of the refman. I suppose this wouldn't be too hard to add.

Can you give some details about the kind of things that would need to go into an `install` file for the refman and how you would see a Makefile doing this?

Zimmi48:
FTR I'd be willing to be the assignee here (as dune secondary owner) but maybe @SkySkimmer (as CI maintainer) or @gares (as Makefile maintainer) would be more appropriate.

ejgallego:
> The dune build system is also missing support for building the PDF version of the refman. I suppose this wouldn't be too hard to add.

Indeed, that should be easy.

> Can you give some details about the kind of things that would need to go into an `install` file for the refman and how you would see a Makefile doing this?

`coq-refman.install` should contain the contents of the build directory. As we are already using some make plumbing make would build the script and then do the install. Hopefully `dune` will be able to do this soon anyways. IMHO it is in no way critical [we can survive without this package for now]

> FTR I'd be willing to be the assignee here (as dune secondary owner) but maybe @SkySkimmer (as CI maintainer) or @gares (as Makefile maintainer) would be more appropriate.

Thanks @Zimmi48 , as you want folks. As I've said I'm in no rush merging this as I can wait until the Dune support gets more and more refined, for now both systems can coexists fine tho it is true that Dune would IMO reduce the burden of maintenance.

Something that won't happen IMO is to reach point where the this PR is "perfect", but were the cutoff point is, I dunno; that's a very opinionated thing.

ejgallego:
I am removing some tags as this has been already properly tested, is IMHO reasonably documented, and sauf for NixOS [which should be easy] the rest is ready modulo some minor nits.

So indeed we may want to wait a bit more before merging this, but my point is that would be merge this nothing significant would happen I think.

maximedenes:
We still need to work on support for `native_compute`, right?

ejgallego:
> We still need to work on support for `native_compute`, right?

Yup, let me add `needs: progress` again, even if native compute is still available in the `build:make` job I guess people won't like to get this merged without support. It is very easy to do tho so I guess I'll get to it in the next weeks.

I am a bit concerned about the status of native compute by several reasons tho; a few comments:
- is it tested? Do we have any user of it in the CI beyond the stdlib?
- what is the performance gain relative to VM compute, are there benchmarks [using proper compiler flags]
- the tooling / linking situation in a bit messy; it could be improved by assuming Coq is built by Dune; but still there are some weird things such as the module naming scheme right?

More generally, vm compute and native compute should live independenly as libraries I think, and be plugged into the kernel, however the current design is the contrary [this made difficult for example the port to multicore OCaml]. So indeed a refactoring would be great however it depends on having a full Dune build.

yurug:
Just a comment about "native compute" : our compiler for Mtac will critically be based on this feature. Please do not break it.

maximedenes:
> * is it tested? Do we have any user of it in the CI beyond the stdlib?

Indeed, it is almost untested, which is bad.

> * what is the performance gain relative to VM compute, are there benchmarks [using proper compiler flags]

Of course, there are benchmarks, see the paper or my thesis. A typical speedup is 4-6x compared to VM IIRC.

> * the tooling / linking situation in a bit messy; it could be improved by assuming Coq is built by Dune; but still there are some weird things such as the module naming scheme right?

Yes, what makes it awkward is the way we interface with the OCaml compiler, basically.

ejgallego:
> A typical speedup is 4-6x compared to VM IIRC.

Does that still hold? I heard that there were improvements on the VM compilation chain since then?

> Yes, what makes it awkward is the way we interface with the OCaml compiler, basically.

Not only the compiler but ecosystem / linking.

> Just a comment about "native compute" : our compiler for Mtac will critically be based on this feature. Please do not break it.

Of course we will try not to break it, but you should be aware that you are critically relying on what is a very experimental feature so don't be surprised if you find yourselves in a lot of pain unless you have a fallback plan.

For example there are know issues, for example it is not possible to unlink modules, etc...

maximedenes:
> Does that still hold? I heard that there were improvements on the VM compilation chain since then?

What improvements do you have in mind?



> Of course we will try not to break it, but you should be aware that you are critically relying on what is a very experimental feature (IMVHO of course) so don't be surprised if you find yourselves in a lot of pain unless you have a fallback plan.

I think this is unnecessarily alarmist. `native_compute` has been used for year by various groups now. Of course, it is not one of the most mature features of the system, but I don't see why a lot of pain should be expected.

> 
> For example there are know issues, for example it is not possible to unlink modules, etc...

How is this a problem in practice?

ejgallego:
> I think this is unnecessarily alarmist. `native_compute` has been used for year by various groups now. Of course, it is not one of the most mature features of the system, but I don't see why a lot of pain should be expected.

Well, I differ here. So far we don't test it, and in my experience, not tested == very very bad. To me, if it is not tested as well it doesn't exists. See for example the `async` situation now that it has been added to the CI, it is failing every one build.

Let's have any of these groups add some of their developments to the CI and then we could see how it goes.

> What improvements do you have in mind?

I can think of a few [in fact you were who suggested them] but without having something that we can bench is useless.

> How is this a problem in practice?

We don't know as few people are testing it; thus the risk is large. Nothing could happen, or it could turn out that large developments using Mtac2 + native are unfeasible to do.

yurug:
I agree that every module of a software system should be continuously tested. Notice thought that this is not the only way to ensure robustness and `native_compute` seems quite robust to me (maybe because its dependencies on the rest of Coq are minimal). 

Our preliminary compiler prototype (for only a subset of Mtac2) uses the `native_compute` machinery and it simply worked out of the box. On our first benchmark, we get a 50x speedup with respect to Ltac.

gares:
Arguing that things that don't build with dune are broken, useless, don't exist is not leading anywhere.

Zimmi48:
To users of the native compute machinery: the opinions expressed above by Emilio are his own (and he's free to express them) but do not reflect a position of the Coq Development Team. If large redesign of native compute was deemed necessary, it would be done (like for any other component) while taking important care to understand users' needs and providing a good compatibility story (as virtually every redesign since the 8.6 release has demonstrated).

MSoegtropIMC:
> We still need to work on support for `native_compute`, right?

You mean native compute on Linux or Windows? It might be that dune can help with native compute on Windows. The main issue with native compute on WIndows is not compile or link tools, but that Coq/Ocaml calls these tools in a way which requires a reasonably Posix compliant shell. Maybe dune can help. On teh other hand I almost finished a "reasonably posix compliant" shell for Windows - just didn't have time to integrate it.

Zimmi48:
I think it was meant on Linux, since it's already supported there. But it would be great indeed if Dune could improve the situation on Windows too.

ejgallego:
> Arguing that things that don't build with dune are broken, useless, don't exist is not leading anywhere.

Inventing things that were not said also don't lead anywhere either. I am not sure if you do this intentionally or it is a problem of reading "too fast"; whatever the cause is I suggest you read again and discover that what you write was never said.

> To users of the native compute machinery: the opinions expressed above by Emilio are his own (and he's free to express them) but do not reflect a position of the Coq Development Team

Well I guess I was not careful to avoid replying to Yann [which somehow feared that this PR would break native].

Breaking/removing native has never been into the table, which does mean that it is still mostly untested thus "critically relying on it" is a risk. It is also good as it will bring some more much needed testing. Me removing the `needs: progress` task was indeed an oversight that I quickly corrected. But somehow my point was "we can switch the whole CI to a nativeless build and nothing happens". This is pretty bad.

In fact, I have (non-finished support for native) in Dune since long time ago. Taking advantage of the draft PR feature I've just submitted it so we can see what we do on Windows.

In fact I've discussed with Maxime many times on how Dune should support native, and I do believe that indeed Dune support will help getting native tested a long way given that today coq_makefile doesn't make it very discoverable [and also some strange things are done with deps, I dunno; it was easy to reproduce many problems in a parallel build due to missing intra-object dependencies]

> The main issue with native compute on WIndows is not compile or link tools, but that Coq/Ocaml calls these tools in a way which requires a reasonably Posix compliant shell.

Does it? I couldn't see where; indeed Dune + native should make the process maybe a bit easier [certainly a lot easier for object files, I dunno about the JIT]

@MSoegtropIMC in Dune 1.8 a truly bash-less build will be possible, if you really have a pressing need we can make it happen today very easily as Vincent removed the latest `bash` calls in #9624 , so all that we use bash for is recursive glob.

ejgallego:
On the other hand my point of view that we should not be afraid of evaluating maintenance costs whenever we are feeling like things are not clear.

It is natural to have questions, and very important to understand the tradeoff between maintenance cost and user benefit.

For example, in this case, native is pretty easy to support code-wise [it is fairly modular], however it introduces quite a few challenges infrastructure-wise [not all its fault as pointed out by Maxime] These permeate to the build system, command line arguments, library handling.

So now imagine that native provides a 1.8x speed-up when compared to the VM, would that justify maintaining it? Surely yes, however what if it was 1.1x? What if it was 20x? Thus IMO it is very important to have data at hand here and don't do maintenance decisions blindly.

gares:
Emilio this PR is about dune, and we all want it to happen. Then we check what is still missing, why it did not happen yet. Someone says it does not yet support X, and your answer to that is that X is maybe not that relevant, not tested hence broken, and finally you discourage actual users to use it.

Please, add support for X to dune or say why it can't be supported or ask for help in supporting X.
Removing X, for any X, is not a option for this PR, it is off topic.

ejgallego:
> your answer to that is that X is maybe not that relevant, not tested hence broken

Sorry that was not my answers, my answer was:

> Yup, let me add `needs: progress` again, even if native compute is still available in the `build:make` job I guess people won't like to get this merged without support. It is very easy to do tho so I guess I'll get to it in the next weeks.

How can you transform the above in "X is not relevant" I am not sure.

What I said next was related to the fact that we were able to get full green CI and forget about native. That obviously concerns me, as if it wasn't for Maxime reminder then I could have just merged this as I had forgotten. So as a CI maintainer, that's a fatal flaw. And I cannot recommend people to depend _critically_ [note here that's the key point] on untested features because that's a high risk, and we have plenty of technical debt, like people depending on `abstract`, `par:`, `derive`, `async, etc... all of them which do have non-trivial known issues and the dependency is introducing great problems now that we try to refactor the system to fix bugs. How can you transform "beware the thing is not tested" to "not tested hence broken" is also a mistery for me.

Native has at least a few critical issues:
- it seems it doesn't work on windows [so imagine how this fares for mtac2]
- it is not tested in the CI,
- it depends on dynamic linking which basically means that many issues could arise on interactive developments,
- tooling needs to be made more robust.

I hope this clears up things.

ejgallego:
I think that indeed the discussion got quite a bit entangled and off-topic, and it a lot my fault as the submitter of course; the PR itself is just waiting until the moment we feel we are ready to switch [and the remaining issues solved] but I don't think it needs a lot more discussion other than wait for that moment; the contents themselves are quite straightforward at this point.

I'd like indeed to take the opportunity of the presence of @yurug here to make clear my position as one of the devs working both in the build system and in trying to improve the integration with external developments:
- my comments about native should have been a private message for Maxime as these were mostly personal questions,
- regarding native itself, while I do think that caution is needed please don't take my words as "don't use it" but as "let's understand how can we have it better tested and supported",
- but indeed this brings into scope the topic of the "internal" view devs have about the system and what view "external" devs have, for example in terms of maturity / stability of features and interfaces.

This last point is critical and indeed I think it is very hard to navigate the current status of this huge and amazing system, as for example we lack a way to properly communicate what the current roadmap in terms of development is; or even worse, in some cases we don't have a roadmap, either because we have different views and we fail to converge, or because we are just exploring some new ideas [such as this PR] and we couldn't anticipate what the plan would be.

This seems like a hard problem, as for example the priorities of coq devs for example may not [and should not] be aligned with the priorities for example of the developers of MTAC [even if IMHO it would be great if unicoq / mtac where somehow part of the official build]. Let's try to communicate using dedicated channels so we are all up to date w.r.t. the plans.

I can try to share some of my experience when working on different parts of the ecosystem. For example, a tool that it has been usually positive is to open tracker issues. For example, for build systems and Coq, one of the first steps I did was to open "informative issues" in first the `ocamlbuild` repository and later in the Dune one outlining the main problems and needs I had w.r.t. using their tools in Coq. This has workerd pretty well and allowed us to keep in touch and understand expectations better. Also the OCaml bug tracker and mailing list have been effective medium for communication with upstream here.

I've also found meetings with the developers to be very useful. Indeed, I guess core plugin developers should try to join the Coq WG more often; encouraging this was under the plugin community task that admittedly it is sitting in my todo list a bit farther than I'd like, so I take the full blame for this.

Well, at least this time I didn't let this message get buried in some org file :) , at the cost of the hijack :)

ejgallego:
Given that we are 2 weeks from the branching point let's aim for the an early 8.11 merge.

Zimmi48:
I'd like to see this merged ASAP. Actually, now that we have branched, I would even remove the legacy Makefiles completely so that we don't have to test them / update them anymore. Since it is early in the release process, we could still restore them if we realize there are serious flaws. I'd like to understand what are the remaining blocking points now. It seems to me that most developers are using Dune now and are enjoying it. What I remember is missing is:
- native compute support
- porting the Windows scripts
- building the PDF version of the refman
- installing the refman

What else? And how easy is it to fix what I just listed?

ejgallego:
Hi @Zimmi48 , thanks for the message:

> I'd like to see this merged ASAP. Actually, now that we have branched, I would even remove the legacy Makefiles completely so that we don't have to test them / update them anymore. Since it is early in the release process, we could still restore them if we realize there are serious flaws. I'd like to understand what are the remaining blocking points now. It seems to me that most developers are using Dune now and are enjoying it.

Indeed it'd be lovely to push this soon; but I think that if something I've learned on this PR is that people get pretty upset when things are not "very polished"; in that sense I am a bit wary of making Dune the default yet. In my view we are missing:

-  to properly build the stdlib with Dune and get rid of the `coq_dune` hack. This is working already in my compose branch, but I prefer to wait for , even tho I'd like to use instead the solution of https://github.com/ocaml/dune/pull/2019 to bootstrap the prelude. Also, in the 1.0 version of the "coq" language, we need a special rule due to the implicit dependency on `stdlib`; that is to say, if a theory named `stdlib` is present in the local workspace, then we assume we are composing with it and we add it to the list of theories. This allows for developments to compose with Coq itself, and it is pretty cool [OCaml for example didn't got it yet sorted out]
- to have a solution for ProofGeneral / CoqIDE, that shouldn't be too hard but needs testing / validation

> * native compute support

that's already done, but needs #9882 or a Dune feature called "generalized domains" that will take quite a while to arrive.

> * porting the Windows scripts

that should be easy, the main CI scripts already use Dune on win.

> * building the PDF version of the refman

I don't see any problem here.

> * installing the refman

We can use the little hack, however it is likely that proper support gets added soon.

MSoegtropIMC:
Regarding the windows scripts: I can do this (I don't expect it to be difficult). Should I do it in this PR or is the plan to first merge it and then to this afterwards in a new PR? I would be fine with either. I have some time over the easter weekend but there are also other Coq PRs I need to work on.

ejgallego:
> Regarding the windows scripts: I can do this (I don't expect it to be difficult). Should I do it in this PR or is the plan to first merge it and then to this afterwards in a new PR? I would be fine with either. I have some time over the easter weekend but there are also other Coq PRs I need to work on.

Thanks @MSoegtropIMC , IMHO given that I'd like to solve the two remaining issues first; and that Dune is already tested on Windows by the CI I'd wait. IMO a plan could be:
- once upstream support is ready we merge this; we keep the makefiles,
- we migrate the rest of tooling / build scripts to Dune, in this transitional period,
- once everyone is happy, we can proceed to remove makefile support.

Matafou:
Sorry I did not follow this thread closely. If I understand well the dune file would be the only user-written file in a coq project

1) or would there be a configure.sh anyway?
2)  Should IDEs parse the dune file
2.a) if yes is this format simple and stable? 
3) Should dune generate a _CoqProject file instead?

ejgallego:
Hi @Matafou , this is more about the Coq build itself, dune + coq is handled upstream [in the dune repos], but indeed:

> If I understand well the dune file would be the only user-written file in a coq project

For a dune project you need to write at least 2 file, `dune-project` and `dune`; as of today you also need to write an OPAM file but that requirement will be (finally!) dropped in the next release.

> 1. or would there be a configure.sh anyway?

If you need to configure something yes, but most projects won't need it.

> 2. Should IDEs parse the dune file

No.

>    2.a) if yes is this format simple and stable?

The format is simple and with stability guarantees tho; however it doesn't provide IDEs the information they usually need.

> 3. Should dune generate a _CoqProject file instead?

This could be a possibility but I am not very much in favor of it, mainly due to a) `_CoqProject` being a very limited format [think of per-file options] and b) polluting the source tree.

What I will try to do in as a first step is to have a `dune coqtop foo.v` command. That will a) ensure all dependencies of `foo` are up to date; b) call coqtop with the right context for `foo`.

a) could be skipped with a flag. So for example for PG, if a dune file is detected, it would just ignore all parsing and just call coqtop that way. Ditto for CoqIDE.

MSoegtropIMC:
> Thanks @MSoegtropIMC , IMHO given that I'd like to solve the two remaining issues first; and that Dune is already tested on Windows by the CI I'd wait. IMO a plan could be:
> 
> * once upstream support is ready we merge this; we keep the makefiles,
> * we migrate the rest of tooling / build scripts to Dune, in this transitional period,
> * once everyone is happy, we can proceed to remove makefile support.

Fine. Please ping me when I can start.

Matafou:
> What I will try to do in as a first step is to have a `dune coqtop foo.v` command. That will a) ensure all dependencies of `foo` are up to date; b) call coqtop with the right context for `foo`.
> a) could be skipped with a flag. So for example for PG, if a dune file is detected, it would just ignore all parsing and just call coqtop that way. Ditto for CoqIDE.

Seems cool, but IDE want to keep track of dependencies for other things than compiling. For instance IDEs detect when a dependency of the currently scripted file is edited (e.g. to retract the currently scripted file). So providing a blind compiling tool is not enough, we need at least the list of dependencies.

Zimmi48:
@Matafou That's not something that PG does today, does it? In the future, shouldn't it be the role of the IDE protocol (i.e. LSP)?

ejgallego:
> @Matafou That's not something that PG does today, does it? In the future, shouldn't it be the role of the IDE protocol (i.e. LSP)?

Indeed, that is something that TTBOMK PG doesn't do today, so it is not in the TODO list of this PR.

The plan in this sense is to reuse whatever method OCaml will use for example in merlin lsp. As far as I know the exact details there are not yet decided, however I am not super up-to-date on what is going on.

Matafou:
Of course PG does it today! Just tick the option "coq/auto compilation/compile before require".
This is contribution from Hendrik Tews a few years ago.

It also recompiles what is needed when you script a "Require" line. It does this by analyzing the _CoqProject file + calls coq_dep, so at least it needs to know the -R options.

JasonGross:
Does dune have an install-coqlight target yet?  If you strip out the existing `Makefile`-based build, you will break HoTT's CI, which tries to build just enough of Coq to get the binaries and base plugins.

ejgallego:
Thanks for the pointer @Matafou , but if I understand that functionality correctly `dune coqtop` subsumes it, right?

Or for example let's say I have two buffers attached to files `a.v` and `b.v`, where in `b.v` we have `Require Import a`.

If I save `a.v`, does `b.v` gets retracted and reasserted?

That's the kind of functionality LSP provides and that I thought was out of scope for the initial merge.

ejgallego:
> Does dune have an install-coqlight target yet? If you strip out the existing `Makefile`-based build, you will break HoTT's CI, which tries to build just enough of Coq to get the binaries and base plugins.

@JasonGross yes there is something like that; you can try it out, and indeed the upcoming stdlib package will split `coq-core` from `coq-stdlib`.

But IMHO there is an even better solution, which is to Dunerize HoTT; this way, you have to do nothing, just place Coq and HoTT together and the build will compose and the exact amount of artifacts required to build HoTT will be the ones built.

Matafou:
> Thanks for the pointer @Matafou , but if I understand that functionality correctly `dune coqtop` subsumes it, right?

If dune coqtop is able to give me the list of dependencies, yes.

> Or for example let's say I have two buffers attached to files `a.v` and `b.v`, where in `b.v` we have `Require Import a`.
> 
> If I save `a.v`, does `b.v` gets retracted and reasserted?

1) As soon as you edit b (even if you don't save b), a gets retracted completely (because PG knows about dependencies thanks to coqdep -R ...).
2) As soon as you script the "Require b.v" in a.v, b.v gets recompiled first (blocking the scripting until the end of compilation) if it is newer than it .vo file. Recursive dependencies are considered.

It works well but it does not deal with vio files, which would speed up things greatly. 

> That's the kind of functionality LSP provides and that I thought was out of scope for the initial merge.

Maybe we just need a "dune coqdep"?

ejgallego:
> 1. As soon as you edit b (even if you don't save b), a gets retracted completely (because PG knows about dependencies thanks to coqdep -R ...).

Oh, I was not aware of this, it is very cool!

> Maybe we just need a "dune coqdep"?

Yeah, that could work; or maybe indeed we should just have PG to use the protocol merlin / LSP will have for this.

jfehrle:
> I'd like to see this merged ASAP. Actually, now that we have branched, I would even remove the legacy Makefiles completely so that we don't have to test them / update them anymore. Since it is early in the release process, we could still restore them if we realize there are serious flaws. I'd like to understand what are the remaining blocking points now. It seems to me that most developers are using Dune now and are enjoying it.

Don't remove the Makefiles yet.  I haven't been able to install Dune.  I spent a week trying to upgrade to 64-bit OPAM 2.0 without success.  See https://github.com/ocaml/opam/issues/3785.  The underIying issue is between cygwin and OCaml. In https://github.com/ocaml/opam/issues/3503#issuecomment-412279840, David Allsopp said "64-bit Cygwin is completely broken for OCaml at the moment, I'm afraid."

In https://github.com/ocaml/opam/issues/3785#issuecomment-478590370, David says

> I do really want to get the various patches in FlexDLL and OCaml dusted off and completed, but the work is not trivial (in particular the OCaml PR) - I doubt I will have time to get it ready in time for the OCaml 4.09 freeze, but hopefully it'll be done in OCaml 4.10 which will be frozen towards the end of this year.

I suggest we wait until that OCaml fix is available and we've verified that it permits building Coq with Dune before dropping the Makefiles.

And, in case anyone suggests using it, the Windows build script is not a viable option for development work.  The build script doesn't support incremental builds well; it takes 30-60 minutes to rebuild Coq after a trivial change (or, actually, even if there's no change).  In the past, I've used it when I had to build CoqIDE because CoqIDE won't build on Windows using `make` outside of the build script.  (I spent a couple weeks trying to make that work).  Also, the last time I tried running the build script a couple months ago, it failed and I was unable to work around the problem.

Getting things to work on Windows is often difficult.  (An unrelated example: I recently wanted to install emacs.  I discovered that the tar file I downloaded contained many binaries, but it didn't include the emacs binary!)

So let's not drop the Makefiles yet.

ejgallego:
@jfehrle have you tried https://fdopen.github.io/opam-repository-mingw/ 

You seem to be using like the worst possible method to install Dune ever, building opam from sources.

For the record I run dune + coq on several windows machines without any kind of trouble.

ejgallego:
> For the record I run dune + coq on several windows machines without any kind of trouble.

Including our CI ; I also think that @MSoegtropIMC didn't find any particular problem with Dune IIRC.

jfehrle:
> You seem to be using like the worst possible method to install Dune ever, building opam from sources.

No, that's not true at all.  I was trying to run the OPAM upgrade following the OPAM upgrade instructions, which triggers various builds.

> For the record I run dune + coq on several windows machines without any kind of trouble.

Do you mean "build" or just "run"?  It seems to depend on exactly how your system is set up, for which there is no guidance.

ejgallego:
> No, that's not true at all. I was trying to run the OPAM upgrade following the OPAM upgrade instructions, which triggers various builds.

First, OPAM is separate from Dune; you don't need OPAM at all to build Dune; I proposed in fact a patch that would have made the bootstrapping process transparent but it was rejected; now you have to suffer the pain but in that sense I am very sympathetic for you.

Second, the OPAM instructions you are following are just useless for Windows. I have pointed out to you several times what the right way to setup the windows toolchain is; I dunno what more I can do if you insist in trying the broken way.

Dune and OCaml do work well on Windows, in fact if OCaml is broken enough as not to be able to build Dune very likely it won't also work for Coq.

jfehrle:
> First, OPAM is separate from Dune; you don't need OPAM at all to build Dune

You recommended using OPAM on March 15: "You need to update to the opam 2 windows version."

On March 16, you wrote: "As Gaëtan pointed out using the quasi-official windows installer has
worked quite well for us so far: ..."  That didn't work either.

> Second, the OPAM instructions you are following are just useless for Windows.  I have pointed out to you _several times_ what the right way to setup the windows toolchain is; I dunno what more I can do if you insist in trying the broken way.

I beg to differ.  I raised the issue in an email chain to `coqdev`: `"make sphinx" on maser fails on Windows` that began on March 12 (excerpted above).  No solution was identified.  Why don't you post your recommendation somewhere on the Coq website?

> OCaml is broken enough as not to be able to build Dune very likely it won't also work for Coq.

I've not had any significant problems using OCaml with the makefiles over the last 18 months.  The issue seems to be specific to OPAM and 64-bit code that uses flexdll.

@gares suggestion far above made a lot of sense:

> we should not merge this before we all managed to switch their workflow to the new build system. I also understand that putting no deadline does not work in practice.  So I propose 1 release cycle. If at "freeze" time for 8.10 we are all happy with the new build system then let's do the switch.

Since this hasn't been achieved (I can't even run Dune), we should not rush into removing the Makefiles for 8.10.

I won't have time to do further experiments with Dune right away (I already spent a week or so on it). There are other things I want to complete for 8.10 first.

Zimmi48:
>we should not rush into removing the Makefiles for 8.10

I wasn't suggesting doing it in 8.10 anyway.

But OK, let's give you at least a few more weeks to leave you time to find a way to install Dune. Two ideas:
- if OCaml with cygwin64 is broken according to David Allsopp, what about using cygwin32?
- do you really need opam to install Dune?

MSoegtropIMC:
@ejgallego : I am currently working on setting up debug builds for MinGW (to get the GTK3 CoqIDE issues debugged). One thing I don't understand with dune (after reading most of the manual) is how to tell it to build a byte code stub dll for cairo and gtk3 or both bytecode and native DLLs. I guess I can add a (mode (byte)) line to all dune files but I also guess that there must be an easier way. The dune manual suggest something like dune build cairo2.bc.dll, but this doesn't work.

ejgallego:
> I guess I can add a (mode (byte)) line to all dune files but I also guess that there must be an easier way.

Umm, I am not sure I understand the problem, in Linux at least the bytecode stubs are always built and installed. `(mode byte opt)` is the default unless explicitly forbidden. The lib name should be something such as `dlllablgtk3_stubs.dll` or so [need to boot my windows machine to chekc]

MSoegtropIMC:
> Umm, I am not sure I understand the problem, in Linux at least the bytecode stubs are always built and installed. `(mode byte opt)` is the default unless explicitly forbidden. The lib name should be something such as `dlllablgtk3_stubs.dll` or so [need to boot my windows machine to chekc]

Indeed I compared (modes byte), (modes native) and (modes byte native). It seems to build and install the correct stuff in all cases including the stub dll. But it doesn't find teh stub DLL when I compile CoqIDE in bytecode. Can it be that pkgconfig is told that the stub DLL is only usable for native code? I checked that the created DLL is (modulo build date) identical in (modes byte) and (modes native), so I guess it is usable for both. But maybe pkgconfig doesn't know this.

ejgallego:
> But it doesn't find teh stub DLL when I compile CoqIDE in bytecode.

Oh indeed this is due to `LD_LIBRARY_PATH` and friends; you have two choices:
- install the packages that do provide the stubs,
- use `dune exec -- $binary` which will setup environment variables properly.

ejgallego:
> * install the packages that do provide the stubs,

You may have some problem with your install path indeed; check you `CAML_LD_LIBRARY_PATH`. 

Another choice is to use `-custom`, this is what is done in native mode by default.

MSoegtropIMC:
My CAML_LD_LIBRARY_PATH is empty. Can you tell me more about -custom?

I was assuming that ocamlfind asks pkgconfig where to find the DLLs. For gtksourceview a ocamlfind query returns the right stuff but I am not so sure where the cairo2 comes in. I am doing / getting this:
```
$ make SHELL="/bin/bash -x" coqide-byte

OCAMLC -o bin/coqide.byte.exe
+ 'C:\bin\cygwin_coq64_ms_coqide\usr\x86_64-w64-mingw32\sys-root\mingw\bin/ocamlfind.exe' ocamlc -thread -rectypes -w +a-4-9-27-41-42-44-45-48-58 -annot -bin-annot -safe-string -strict-sequence -I coqpp -I config -I clib -I lib -I kernel -I kernel/byterun -I library -I engine -I pretyping -I interp -I proofs -I gramlib/.pack -I parsing -I printing -I tactics -I vernac -I stm -I toplevel -I ide -I ide/protocol -package lablgtk3-sourceview3 -g -o bin/coqide.byte.exe -linkpkg -package str,unix,dynlink,threads,lablgtk3-sourceview3 -ccopt '-subsystem windows' config/config.cma clib/clib.cma lib/lib.cma ide/protocol/ideprotocol.cma ide/ide.cma ide/coqide_os_specific.cmo ide/coqide_main.mli ide/coqide_main.ml
[1mFile "[1mide/coqide_main.ml", line 1[0m[0m:
[1;31mError[0m: Error on dynamically loaded library: dllcairo_stubs.dll: The specified module could not be found.
```

ejgallego:
> I am doing / getting this:

That seems like an issue with the regular Makefiles.

MSoegtropIMC:
> That seems like an issue with the regular Makefiles.

I guess I shouldn't ask this to "Mr. Dune", but do you have an idea what is wrong? That an explicit reference to the cairo2 package is missing? The thing I don't understand is why it works in the native build. The command line looks very similar.

ejgallego:
> do you have an idea what is wrong?

As I said it seems that the Makefile system is not setting `CAML_LD_LIBRARY_PATH` properly, I am afraid there are quite a few bugs like that lurking in Make code.

But please let's keep this PR on topic, open an issue for that problem if it needs further discussion.

ejgallego:
> The thing I don't understand is why it works in the native build.

For the native build `-custom` is passed so the library is linked statically. I tried to do this for bytecode in the past but we decided against, c.f. https://github.com/coq/coq/issues/6475

It is not so important for me anymore as the build system I use handles all this stuff _almost_ properly [indeed the stuff is so tricky that even Dune has serious bugs in that regard]

gares:
> My CAML_LD_LIBRARY_PATH is empty

I think one needs to set it only if stuff is not installed in the standard paths (that is, you built them but you did not put them were ocamlfind expects them to be). Try `ocamlfind printconf` and then look at the `ld.conf` file that should contain the path of the stubs.

jfehrle:
> I wasn't suggesting doing it in 8.10 anyway.

Your earlier comment (below) sounds like you wanted to do it by the time 8.10 is released ("Since it is early in the release process ...").  You didn't say whether in master or the 8.10 branch.

> I'd like to see this merged ASAP. Actually, now that we have branched, I would even remove the legacy Makefiles completely so that we don't have to test them / update them anymore. Since it is early in the release process, we could still restore them if we realize there are serious flaws.

> if OCaml with cygwin64 is broken according to David Allsopp, what about using cygwin32?

I had thought about that.  I was concerned about how much stuff I would have to re-install and reconfigure.  The experimentation is time-consuming.  I've been doing everything 64-bit.  So far I haven't needed the larger address space, but the day I do need it may arrive sooner than I expect.

As I understand it, the flexdll problem is about mismanagement of library code shared between processes.  The failure appears to be intermittent and random.  I expect the more processes you have on your system, the more likely you are to encounter it.  Our CI systems are probably running only the build and little else; my system has all kinds of stuff on it: 359 processes as I write this.  Add on a parallel build.  After compiling a few thousand files, the odds of hitting the bug apparently get very high.  Failures occur at different points.

Users of 64-bit Windows Coq may run into this bug.  Hopefully not often.

> do you really need opam to install Dune?

For Dune and only Dune, maybe not.  As a short-term workaround, maybe.  But aren't we setting up a bunch of libraries in OPAM that require OPAM-installed Coq?  I would imagine that anything that relies on Coq would no longer be OPAM-installable without Dune.  So far I've (mostly) not needed to install Coq libraries, but if I do, it would be painful to lose package manager support.

Zimmi48:
>Your earlier comment (below) sounds like you wanted to do it by the time 8.10 is released ("Since it is early in the release process ...").

Sorry for being unclear: I meant it is early in the 8.11 release process.

>For Dune and only Dune, maybe not. As a short-term workaround, maybe. But aren't we setting up a bunch of libraries in OPAM that require OPAM-installed Coq? I would imagine that anything that relies on Coq would no longer be OPAM-installable without OPAM-installed Dune. So far I've (mostly) not needed to install Coq libraries, but if I do, it would be painful to lose package manager support.

Yes, but this is a question that I think should be viewed as fully independent of the Dune question since the Coq package index has already switched to opam 2. Therefore, if you have trouble installing opam 2 on Windows, this is already going to be a problem for you. But this shouldn't prevent moving the build system to Dune anyway, especially if you can manage to install Dune without opam.

jfehrle:
> this is a question that I think should be viewed as fully independent of the Dune question since the Coq package index has already switched to opam 2. Therefore, if you have trouble installing opam 2 on Windows, this is already going to be a problem for you.

Ah, it's not quite that independent.  I wasn't able to upgrade my existing OPAM 1 to OPAM 2, however I was able to install OPAM 2 as a Cygwin package (in a separate Cygwin installation).  The issue was that I wasn't able to build Dune+Coq under OPAM 2.

I'll let you know what I learn when I get back to this.

erikmd:
Hi all, I did not closely follow the discussion in this thread, but I've noted two approaches were mentioned here to get OPAM 2 working on Windows (Cygwin and MinGW).

FWIW, I've tested a third one this semester that might be of interest for some of you: It is based on **WSL** (the Windows Subsystem for Linux), and the installation of OPAM 2 + Merlin + Coq makes it possible to use the corresponding coqtop from a **native Windows version of GNU Emacs with ProofGeneral**.

We've been using that setup this year in 2 different teaching units in Univ. Toulouse III (Bachelor+Master), raising no particular issue.

To be more precise:
* It works only with Windows 64-bit;
* It allows one to natively run ELF executables under Windows thanks to WSL;
* these binaries can be installed from several distributions (e.g. Ubuntu, Debian), or compiled using make, etc.;
* the main advantage here is the capacity to smoothly run native GNU/Linux CLI programs in Windows;
* beyond CLI :
  * regarding the communication between WSL and ProofGeneral, I've been relying on [wsl-alias](https://github.com/leongrdic/wsl-alias)
  * but if need be (to be able to run CoqIDE instead), one can also install GNOME in WSL and install [VcXsrv](https://sourceforge.net/projects/vcxsrv/) (an implementation of Xorg for Windows)

For the moment I've just documented the approach [in this GitHub repo](https://github.com/erikmd/tapfa-init-win64.el) (in French for my students), but maybe I could add a related English translation in [proofgeneral.github.io](https://proofgeneral.github.io/) if it's worth it.

Zimmi48:
@erikmd We've derived far off the initial topic of this thread, so if we need to discuss this more, let's open a new one (possibly on [Discourse](https://coq.discourse.group/)), but let me say that I'm interested in having this documented, in the [Coq wiki](https://github.com/coq/coq/wiki/Installation%20of%20Coq%20on%20Windows) itself.
I had tried doing something similar to you, but with CoqIDE: my idea was to use the CoqIDE version provided in the Windows installer (native CoqIDE, no X server required) and to tell it to use the `coqidetop` command installed through WSL (either with opam or Nix) as a way to be able to easily install additional plugins and libraries. The brick I was missing was the `wsl-alias` trick.

ejgallego:
This PR has gone stale. Unfortunately was a huge failure and implied a total loss of time of many months (being conservative).

I still believe the original plan was sound and would have brought many benefits would we have been a bit more proactive; a summary of the ongoing effort can be found at https://github.com/coq/coq/wiki/Coq-Call-2021-02-24#dune-build-system-update

PR #8730, opened 2018-10-14
Add unicode category LM

mir-ikbch:
Since the Unicode character category LM contains some characters frequently used in Japanese words (e.g., U+30FC "ー" is used in words like データ, which means "data", オーバーフロー, which means "overflow", etc.,) so this change will be very helpful to write Coq codes in Japanese.

### DIFF
--- a/clib/unicode.ml
+++ b/clib/unicode.ml
@@ -86,6 +86,7 @@ let classify =
         Unicodetable.ll;           (* Letter, lowercase.                *)
         Unicodetable.lt;           (* Letter, titlecase.                *)
         Unicodetable.lo;           (* Letter, others.                   *)
+        Unicodetable.lm;           (* Letter, modifier.                 *)
       ];
     mk_lookup_table_from_unicode_tables_for IdentPart
       [


herbelin:
@mir-ikbch: Very good remark. But we have to think twice however as the "letter modifiers" table apparently brings 250 characters which would now be considered identifier parts, thus possibly breaking developments assuming these characters were lexed as symbols.

Apparently, many of these signs are for the U+02B0 - U+02FF block for IPA (33 letter modifiers out of 80 characters), the U+1D00 - U+1D7F block (62 letter modifiers out of 128 characters).

Many are legitimate signs for various languages, Greek, Arabic, Cyrillic, Thai, Japanese, ...

Among the ones liable to causing surprises, there are the "modified letter prime" `ʹ ` (U+02B9) or the "Greek numeral sign" `ʹ` (U+0374). On the other side, to have a symbol which does not collide with identifiers, we still have the "prime" `′` (U+2032). Also, the usual "apostrophe" `'` (U+0027) is not a letter but is already considered to be an ident part in Coq identifiers. [We could also argue that all these variants are confusing since they graphically look almost the same. But that's another, independent, story.]

In any cases, I did not make a systematic browsing, but are some of these letter modifiers intended to be used at the beginning of a word or are they intended to be used only after some characters (such as ー, ʹ, ...). In the latter case, it could instead be added to the `IdentPart` category.

To be noted: [Perl](https://docs.perl6.org/language/unicode_ascii#Alphabetic_characters) accepts any  Ll, Lu, Lt, Lm, Lo in identifiers, with Lm even in first position.

Well, let's take some time to collect comments about this proposal.

ppedrot (commented):
@herbelin said it better than me: this is a potentially far-reaching change that needs careful pondering. In the absolute, I think it is fine though, but let's wait from input by other people.

jashug (approved):
This seems fine to me, and indeed some of these characters are necessary for writing Japanese words.

vbgl (approved):

vbgl:
Pierre-Marie, do you think this needs more discussion? If this is the case, will it happen before next release? IMHO it’s fine to merge now.

Ping @ppedrot

ppedrot:
Fine, I'll merge this soon (when is the branching already?) if nobody complains.

ppedrot:
CI failure looks spurious, let's merge.

SkySkimmer:
I think this broke mtac2 (not detected due to old CI)

SkySkimmer:
I'll try making a PR there to fix it.

SkySkimmer:
see https://github.com/Mtac2/Mtac2/pull/180

Gaëtan Gilbert

Zimmi48:
I wonder what is the point of the merge script warning about old CI if people are so used to seeing it that they just don't care about it anymore. But it is actually useful to rebase an old PR before merging it. Maybe we should just remove this warning from the merge script and replace it by e.g., the bot adding a "needs: rebase" label to PRs whose CI has not been run in the last 2 weeks?

PR #8733, opened 2018-10-14
Implement ARGUMENT EXTEND in coqpp

ppedrot:
After the much acclaimed porting of all but one EXTEND macros to coqpp, now is the moment to finalize it! This PR implements the [VERNAC] ARGUMENT EXTEND macros in coqpp, and ports all remaining macro files to coqpp.

The only remaining ml4 files are in the test-suite and probably need a bit of fiddling with coq_makefile, and there are only two really remaining ml4 files containing code.

In the process a proper API for argument extension has been provided so as to reduce the quantity of code generated at compile-time for these EXTEND macros.


coqpp/coqpp_main.ml
```diff
@@ -97,9 +100,16 @@ let rec print_symbol fmt = function
   assert (e = "tactic");
   fprintf fmt "@[Extend.TUentryl (Genarg.get_arg_tag wit_%s, %i)@]" e l
 
+let print_string fmt s = fprintf fmt "\"%s\"" s
+
+let print_opt fmt pr = function
+| None -> fprintf fmt "None"
+| Some x -> fprintf fmt "Some@ (%a)" pr x
```
ejgallego: Nit: I'd recommend a box `@[(%a)@]`.

vernac/vernacentries.mli
```diff
@@ -63,6 +63,24 @@ val vernac_extend :
...
+    entries instead of ty_user_symbol and thus arguments as roots. *)
+
+type 'a vernac_argument = {
+  arg_printer : 'a -> Pp.t;
+  arg_parsing : 'a argument_rule;
+}
+
+val vernac_argument_extend : name:string -> 'a vernac_argument ->
+  ('a, unit, unit) Genarg.genarg_type * 'a Pcoq.Entry.t
+
```
ejgallego: I am not sure `Vernacentries` is the best file where to place this, after all is `Vernacinterp` that does the job. Maybe you should consider moving all the code after `let interp` to that file [or other]

ejgallego (approved):
Looks good to me.

ejgallego:
I am gonna rebase the vendoring PR on top of this one to see how things look.

vernac/vernacentries.mli
ppedrot: I put it there for uniformity with `vernac_extend`.

ejgallego:
> The only remaining ml4 files are in the test-suite and probably need a bit of fiddling with coq_makefile, and there are only two really remaining ml4 files containing code.

About this, indeed I can build Coq without `camlp5` at all using #8667. Note that there are two different kind of `.ml4` files, which are processed in Dune by `coqp5` and `coqmlp5`. None of the remaining files needs `coqp5` anymore thus `grammar` can go away completely.

ppedrot:
We probably need to support (French meaning of the word also applies) camlp5 for a full release cycle to provide a reasonable deprecation phase. This doesn't prevent being able to build Coq without it or even provide the macros as a standalone deprecated library.

ejgallego:
> We probably need to support (French meaning of the word also applies) camlp5 for a full release cycle to provide a reasonable deprecation phase. This doesn't prevent being able to build Coq without it or even provide the macros as a standalone deprecated library.

Yup, I just did `rm -rf grammar` as a sanity check [tree `b69972c596fb7bd7e9b8d82c6f8cc3cc6897c62b`] and indeed it works beautifully.

I dunno about `grammar`, supporting it was never a problem for me [using Dune] anyways.

ejgallego:
I think that what I will do in Dune is to declare `coqp5` an "optional" binary, that is to say, it is built if camlp5 is in scope. So that way things should work out of the box for both people with camlp5 and without not.

Of course plugins will only work fine if they have `coqp5` in scope, or they are ported to `.mlg`.

ppedrot:
@ejgallego Nitpick addressed, this is ready for me.

> cLexer.ml4 which I don't know why it's not a .mll

@SkySkimmer The camlp5 syntax is used for the stream parsing feature, which is orthogonal to ocamllex. Yet, it is very easy to expand it by hand without making the code horrible at all. I can open a quick PR for that.

ejgallego:
> @ejgallego Nitpick addressed, this is ready for me.

Do you want me to be the assignee? Last one was @maximedenes but indeed we didn't add codeowners to the `coqpp` stuff. I also think this is ready.

> Yet, it is very easy to expand it by hand without making the code horrible at all. I can open a quick PR for that.

Remember that `camlp5` can do the expansion for you.

ppedrot:
> Remember that camlp5 can do the expansion for you.

Yes, but the file is small enough that hand-expanding is fine and furthermore the code will be nicer-looking.

ejgallego:
> Do you want me to be the assignee?

I'm going offline, if you agree with me being the assignee assign this to me @ppedrot and I'll merge tomorrow morning.

ppedrot:
@ejgallego Also, you can be the assignee, I guess.

ejgallego:
> Yes, but the file is small enough that hand-expanding is fine and furthermore the code will be nicer-looking.

Sure, but I've found more effective to start from the machine-expanded version and then tweak by hand. But well, I was dealing with the revised syntax which is a bit more funny to translate by hand.

ejgallego:
> @ejgallego Also, you can be the assignee, I guess.

As you wish.

ejgallego:
> the remaining ml4 are

Couple of files in the test suite too.

Note that I am using the `promote` facility to keep a copy of the output files from preprocessing managed by the build system, that can be useful some times.

ejgallego:
Failure is spurious due to an overlay, it had passed earlier today.

PR #8734, opened 2018-10-14
Make diffs work for more input strings

### LINKED ISSUE #8120: Cleanup "To Do"s for Proof Diffs

Resolve "to do" issues in Proof Diffs (PR #6801):

1.  Add and use a lexer mode that makes every string lexable so that Pp_diff.diff_str can easily be used  on any input string.  Also make the tokens exactly match the input, e.g. include the quote characters for strings.
2.  Figure out if there's a good way to persist the diffs setting rather than having to specify it every time when invoking coqtop.  Also determine whether this is really a good idea.
3.  Document the feature in Sphinx.

jfehrle:
Diff code uses the lexer to recognize tokens in the inputs, which can be
Pp.t's or strings.  To add the highlights in the resulting Pp.t, the diff code
matches characters in the input to characters in the tokens.  Current
code fails for inputs containing quote marks or "(*" because the quote
marks and comments don't appear in the tokens.  Also some characters
such as "&" and UTF-8 characters that are considered `Symbol`s are rejected
by the lexer.  This commit adds a "diff mode" to the lexer to return those
characters, making the diff routine more robust.

Note: It will still be possible to get a Diff_Failure if the string contains ill-formed UTF-8 characters.

**Kind:** bug fix.

Fixes https://github.com/coq/coq/issues/8120 issue 1 (issue 3 has already been done, issue 2 is not important)

### DIFF
--- a/parsing/cLexer.ml
+++ b/parsing/cLexer.ml
@@ -548,20 +548,27 @@ let process_sequence loc bp c cs =
   aux 1 cs
 
 (* Must be a special token *)
-let process_chars loc bp c cs =
+let process_chars ~diff_mode loc bp c cs =
   let t = progress_from_byte loc None (-1) !token_tree cs c in
   let ep = Stream.count cs in
   match t with
     | Some t -> (KEYWORD t, set_loc_pos loc bp ep)
     | None ->
         let ep' = bp + utf8_char_size loc cs c in
-        njunk (ep' - ep) cs;
-	let loc = set_loc_pos loc bp ep' in
-        err loc Undefined_token
+        if diff_mode then begin
+          let len = ep' - bp in
+          ignore (store 0 c);
+          ignore (nstore (len - 1) 1 cs);
+          IDENT (get_buff len), set_loc_pos loc bp ep
+        end else begin
+          njunk (ep' - ep) cs;
+          let loc = set_loc_pos loc bp ep' in
+          err loc Undefined_token
+        end
 
 (* Parse what follows a dot *)
 
-let parse_after_dot loc c bp s = match Stream.peek s with
+let parse_after_dot ~diff_mode loc c bp s = match Stream.peek s with
   | Some ('a'..'z' | 'A'..'Z' | '_' as d) ->
       Stream.junk s;
       let len =
@@ -576,11 +583,11 @@ let parse_after_dot loc c bp s = match Stream.peek s with
           let len = ident_tail loc (nstore n 0 s) s in
           let field = get_buff len in
           (try find_keyword loc ("."^field) s with Not_found -> FIELD field)
-      | AsciiChar | Utf8Token _ | EmptyStream -> fst (process_chars loc bp c s)
+      | AsciiChar | Utf8Token _ | EmptyStream -> fst (process_chars ~diff_mode loc bp c s)
 
 (* Parse what follows a question mark *)
 
-let parse_after_qmark loc bp s =
+let parse_after_qmark ~diff_mode loc bp s =
   match Stream.peek s with
     | Some ('a'..'z' | 'A'..'Z' | '_') -> LEFTQMARK
     | None -> KEYWORD "?"
@@ -588,7 +595,7 @@ let parse_after_qmark loc bp s =
         match lookup_utf8 loc s with
           | Utf8Token (st, _) when Unicode.is_valid_ident_initial st -> LEFTQMARK
           | AsciiChar | Utf8Token _ | EmptyStream ->
-            fst (process_chars loc bp '?' s)
+            fst (process_chars ~diff_mode loc bp '?' s)
 
 let blank_or_eof cs =
   match Stream.peek cs with
@@ -598,20 +605,20 @@ let blank_or_eof cs =
 
 (* Parse a token in a char stream *)
 
-let rec next_token loc s =
+let rec next_token ~diff_mode loc s =
   let bp = Stream.count s in
   match Stream.peek s with
   | Some ('\n' as c) ->
       Stream.junk s;
       let ep = Stream.count s in
-      comm_loc bp; push_char c; next_token (bump_loc_line loc ep) s
+      comm_loc bp; push_char c; next_token ~diff_mode (bump_loc_line loc ep) s
   | Some (' ' | '\t' | '\r' as c) ->
       Stream.junk s;
-      comm_loc bp; push_char c; next_token loc s
+      comm_loc bp; push_char c; next_token ~diff_mode loc s
   | Some ('.' as c) ->
       Stream.junk s;
       let t =
-        try parse_after_dot loc c bp s with
+        try parse_after_dot ~diff_mode loc c bp s with
           Stream.Failure -> raise (Stream.Error "")
       in
       let ep = Stream.count s in
@@ -630,13 +637,13 @@ let rec next_token loc s =
       Stream.junk s;
       let t,new_between_commands =
         if !between_commands then process_sequence loc bp c s, true
-        else process_chars loc bp c s,false
+        else process_chars ~diff_mode loc bp c s,false
       in
       comment_stop bp; between_commands := new_between_commands; t
   | Some '?' ->
       Stream.junk s;
       let ep = Stream.count s in
-      let t = parse_after_qmark loc bp s in
+      let t = parse_after_qmark ~diff_mode loc bp s in
       comment_stop bp; (t, set_loc_pos loc bp ep)
   | Some ('a'..'z' | 'A'..'Z' | '_' as c) ->
       Stream.junk s;
@@ -670,12 +677,16 @@ let rec next_token loc s =
       Stream.junk s;
       begin try
         match Stream.peek s with
+        | Some '*' when diff_mode ->
+            Stream.junk s;
+            let ep = Stream.count s in
+            (IDENT "(*", set_loc_pos loc bp ep)
         | Some '*' ->
             Stream.junk s;
             comm_loc bp;
             push_string "(*";
-            let loc = comment loc bp s in next_token loc s
-        | _ -> let t = process_chars loc bp c s in comment_stop bp; t
+            let loc = comment loc bp s in next_token ~diff_mode loc s
+        | _ -> let t = process_chars ~diff_mode loc bp c s in comment_stop bp; t
       with Stream.Failure -> raise (Stream.Error "")
       end
   | Some ('{' | '}' as c) ->
@@ -683,7 +694,7 @@ let rec next_token loc s =
       let ep = Stream.count s in
       let t,new_between_commands =
         if !between_commands then (KEYWORD (String.make 1 c), set_loc_pos loc bp ep), true
-        else process_chars loc bp c s, false
+        else process_chars ~diff_mode loc bp c s, false
       in
       comment_stop bp; between_commands := new_between_commands; t
   | _ ->
@@ -695,14 +706,14 @@ let rec next_token loc s =
             comment_stop bp;
             (try find_keyword loc id s with Not_found -> IDENT id), set_loc_pos loc bp ep
         | AsciiChar | Utf8Token _ ->
-            let t = process_chars loc bp (Stream.next s) s in
+            let t = process_chars ~diff_mode loc bp (Stream.next s) s in
             comment_stop bp; t
         | EmptyStream ->
             comment_stop bp; (EOI, set_loc_pos loc bp (bp+1))
 
 (* (* Debug: uncomment this for tracing tokens seen by coq...*)
-let next_token loc s =
-  let (t,loc as r) = next_token loc s in
+let next_token ~diff_mode loc s =
+  let (t,loc as r) = next_token ~diff_mode loc s in
   Printf.eprintf "(line %i, %i-%i)[%s]\n%!" (Ploc.line_nb loc) (Ploc.first_pos loc) (Ploc.last_pos loc) (Tok.to_string t);
   r *)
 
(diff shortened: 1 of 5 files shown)


parsing/cLexer.ml4
```diff
@@ -513,16 +513,23 @@ let process_sequence loc bp c cs =
   aux 1 cs
 
 (* Must be a special token *)
-let process_chars loc bp c cs =
+let process_chars ?(diff_mode=false) loc bp c cs =
```
SkySkimmer: Using an optional argument here and for the other function makes me worry that we're going to forget to pass it in some call. Why not make it required? Especially since it has no impact on external callers (the mli didn't change).

jfehrle: I think `false` is the usual, "normal" setting and `true` will probably only ever be passed by the diff code.  Making it optional with a default implicitly documents this.  But I was not entirely consistent--`diff_mode` is a required argument for make_lexer but not the others.

I will change it if you still think I should.

SkySkimmer:
I do still think so.

Gaëtan Gilbert

jfehrle:
OK, I made the optional argument required.

parsing/cLexer.ml4
```diff
@@ -513,20 +513,27 @@ let process_sequence loc bp c cs =
   aux 1 cs
 
 (* Must be a special token *)
-let process_chars loc bp c cs =
+let process_chars diff_mode loc bp c cs =
```
SkySkimmer: While rebasing to deal with conflicts could you also make it a labelled argument? Thanks.

SkySkimmer (approved):

jfehrle:
> I'm trying to understand why lexing is needed for diffing.

Because there's no guarantee that a Ppcmd_string contains a single token.  I noticed some violations of this earlier in proof diffs such as spaces mixed with token(s) in a single string.  Also, I wanted to provide a routine Pp.diff_str that takes 2 strings with minimal limitations--to make the diff API more complete.

I acknowledge that it adds a bit of complexity.

The comments you point to were largely in reference to my using a global variable to control `diff_mode` in my initial quick prototype, which I have since removed.

I'm not eager to redesign the diff code after so much work--all of the code was written at least twice and extensively reviewed over a 9 month period.  Let's not make the perfect the enemy of the good.  Surely there are many other things to improve in Coq that would provide more value.

herbelin:
> Because there's no guarantee that a Ppcmd_string contains a single token. I noticed some violations of this earlier in proof diffs such as spaces mixed with token(s) in a single string.

Can you give an example of a `Ppcmd_string` which is built from more than a single token of a Coq expression (if that happens, that looks more hackish than recommended, and I would rather modify this non standard use of `Ppcmd_string`).

> Also, I wanted to provide a routine `Pp.diff_str` that takes 2 strings with minimal limitations--to make the diff API more complete.

What do you mean by minimal limitations? If you want the API to support arbitrary strings, you can still have the core `diff_str` working on list of strings, and a special variant of it which precomposes with lexing, without having to make a detour for the general case. Would that fit your need?

Also, you have a test with the string `(* comment *) () \"string\" \\ := Ж &gt; ∃ 'c' xx`. I tested it in Coq and the lexer seems to accept it. What is the problem which you had with the current lexer on this string?

jfehrle:
> Can you give an example of a `Ppcmd_string` which is built from more than a single token of a Coq expression

IIRC, the cases I saw were mostly single tokens plus spaces.  For example:

`  forall n : nat, ev n -> exists k : nat, n = double k`

has 
```
Ppcmd_string ' : '
Ppcmd_string ' ->'
Ppcmd_string ' : '
Ppcmd_string ' ='
```

Relying on the Pp.t to be grouped in tokens seems quite fragile--far too easy for someone modifying the printing code to forget about that requirement.  And how do you find and fix all of the non-token cases in the first place?  If diffs calls the lexer, it's much easier to believe that the code is correct.  And I think it matters to get the tokens exactly right--if they're not quite correct, it makes a poor impression to users.

> Also, you have a test with the string `(* comment *) () \"string\" \\ := Ж &gt; ∃ 'c' xx`. I tested it in Coq and the lexer seems to accept it. What is the problem which you had with the current lexer on this string?

1. The lexer raises an exception for the string "\&gt;" because "&" is neither alphanumeric nor defined as an operator.  Same thing with multibyte characters that are classified as "Symbol"s.  Look in the unit test `proof_diffs_test.ml` (in master, not this PR) below the comment "example of a limitation, not really a test".  Line 54.  

2. The lexer doesn't generate any output characters for comments, and for strings it drops the quotation marks.  After identifying the differences between the old and new values, the diff code has to add tags to the original Pp.t's to highlight the differences.  This part of the algorithm depends on matching the input string and the output string character by character (ignoring white space).  The diff can't be completed if this property doesn't hold.  Look at the output for the test at line 67 ("tokenize_string examples").  The input string `"(* comment *) \"string\" 'c' xx"` returns these tokens:

```
'string' ''' 'c'' 'xx' 
TEST SUCCEEDED: tokenize_string examples
```

> If you want the API to support arbitrary strings, you can still have the core `diff_str` working on list of strings, and a special variant of it which precomposes with lexing, without having to make a detour for the general case. Would that fit your need?

The special variant would need the lexer changes proposed in this PR.  If the lexer changes are available, what would be the benefit of making the general case operate differently?  At that point it would simpler to have them work the same way.  Lexing and diffing proof contexts should be fast enough generating user-visible output that the CPU time used will be imperceptible.

herbelin:
> IIRC, the cases I saw were mostly single tokens plus spaces.  For example:

`  forall n : nat, ev n -> exists k : nat, n = double k`

has 
```
Ppcmd_string ' : '
Ppcmd_string ' ->'
Ppcmd_string ' : '
Ppcmd_string ' ='
```
You're right. There are many such places which inserts spaces explicitly, which is not so bad after all.

> If diffs calls the lexer, it's much easier to believe that the code is correct.  And I think it matters to get the tokens exactly right--if they're not quite correct, it makes a poor impression to users.

So, you see the re-lexing as some kind of double-checking that the printer did not produce an unparsable string, so why not.

> The lexer doesn't generate any output characters for comments

There is a mode to collect comments, but first, I'd like to understand what kind of diff we can make with comments. Are you assuming for diffing that the inner of comments contains Coq expressions, or are you expecting nothing of the inner of comments? In the latter case, since the Coq lexer is designed to lex Coq expressions, I don't see what it could give useful for, say, natural language expressions (e.g. `(* top-down *)` would be lexed as `top`, `-`, `down`).

> 1. The lexer raises an exception for the string "\&gt;" because "&" is neither alphanumeric nor defined as an operator. 

If I do `Check &.` at toplevel, the lexer accepts the `&`, but, indeed, there are tokens which are not accepted. Coq lexer is dynamic and adds new tokens according to notations that you declare (for instance `Check $$.` is rejected by the lexer but `Notation "$$" := tt. Check $$.` is ok.) l So, if the lexer fails, it is that the printed sentence is not a correct Coq expression and the fault is not the one of the lexer but of the printer .

> and for strings it drops the quotation marks.

The lexer does not really drop the quotation mark, it is rather that it moves the information to the name of the token. To keep the quotes, you should modify the function `Tok.extract_string` so that it returns `"\"" ^ s ^ "\""` in the `STRING` case. Similarly, the `FIELD` case should return `"." ^ s`. The case `PATTERNIDENT` is obsolete apparently (replaced by `LEFTQMARK`) but at some times, it used to mean an identifier preceded by a `?`.

jfehrle:
> So, you see the re-lexing as some kind of double-checking that the printer did not produce an unparsable string, so why not.

Not quite.  I want to be sure that if a Ppcmd_string contains multiple tokens, such as "n:=" that the "n" and ":=" will be treated as indivisible units for diffing (or similarly if a token spans adjacent Ppcmd_strings).  Also, for " -> ", that the spaces are not included in the diff computation.  Whitespace gets highlighted only between adjacent tokens that are both differences, and never at the beginning or end of a line.

> Are you assuming for diffing that the inner of comments contains Coq expressions, or are you expecting nothing of the inner of comments? In the latter case, since the Coq lexer is designed to lex Coq expressions, I don't see what it could give useful for, say, natural language expressions (e.g. `(* top-down *)` would be lexed as `top`, `-`, `down`).

I'm not making assumptions about what comments contain.  It seemed reasonable to diff the comment with some granularity rather than as a single unit.  Comments can be very long.  Natural language words (unhyphenated) and numbers will be treated as units.  The one-line change I made to return comments seems reasonable since there's no unique way to tokenize and diff comments.  An example from my IDE:

image

> If I do `Check &.` at toplevel, the lexer accepts the `&`

In the unit test, I believe the notation `&` hasn't been loaded into the lexer, so the behavior is a little different.  I believe any symbol character (including Unicode) that isn't very basic such as `().` or defined as notation will not lex.  Maybe not the best example.

> So, if the lexer fails, it is that the printed sentence is not a correct Coq expression and the fault is not the one of the lexer but of the printer.

I think it's unlikely the printer will generate an unlexable string.  It is possible for a string not generated by the printer.

> To keep the quotes, you should modify the function `Tok.extract_string` so that it returns `"\"" ^ s ^ "\""` in the `STRING` case.

I'll do that.  Does the lexer support escape character sequences beginning with `\\`, possibly including unicode characters?  If so, is the mapping from the input string containing the escapes to `s` one-to-one?  (i.e., can I reliably recover the original characters from `s`?  Or would I have to change the code that accumulates `s`?  I thought it would be tolerable to break a string into multiple tokens because the code for that is just one line.  Behavior akin to comments.  AFAIK, proof contexts can't contain quoted strings--is that right?

> Similarly, the `FIELD` case should return `"." ^ s`. The case `PATTERNIDENT` is obsolete apparently (replaced by `LEFTQMARK`) but at some times, it used to mean an identifier preceded by a `?`.

For FIELD, yes.  But for inputs such as "?id" and "(* ?id *)", I am getting the question mark in the output.

Interestingly, `Proof_diffs.ml` has the only call to `Tok.extract_string`.  I can add `diff_mode` as an argument, or create an alternate routine to call.

jfehrle:
@herbelin I updated `Tok.extract_string` to handle the STRING and FIELD cases as you suggested.  I looked at turning on `Flags.beautify` and extracting the comment from `current_comment`.  That works, returning the entire comment as a single token.  On balance, I think tokenizing what's inside the comment instead of treating it as a single token will be more appealing for users, so I kept the latter behavior.

Please let me know your further comments, or if I've not fully responded to your earlier comments.

jfehrle:
> [String] escaping uses Pascal convention. Only " is escaped, and it is escaped by doubling.

Got this, thanks.

> I agree that it seems reasonable to diff the comment with some granularity rather than as a single unit. It is however unclear to me that it is better to parse them as Coq expressions rather than using an ad hoc lexer which, e.g. would parse words and punctuations.  (But maybe this gives roughly the same at the end.)

What I have seemed like the simplest change to do something reasonable.  If this turns out to be an important use case and the behavior less than ideal, we can improve it later on.

> I still don't understand why you treat process_chars in an ad hoc way. You don't want to trust that the printer shall print only registered symbols, right?

This is to handle the case where someone reuses the diff code on strings that were not generated by printer.

I pushed an updated version, ready to merge if you're OK with the changes.

jfehrle:
It occurs to me that this will fail for a malformed string, e.g. one missing its closing `"`.  Will get back to you.

ppedrot:
This PR seems to lack an assignee. It was once ready to merge, but now it needs a rebase. @herbelin are you assigning it to yourself? You were the most involved in the discussion.

herbelin:
I'm assigning and will try to find time for a quick review. Please ping me if I give no news by next Thursay evening.

jfehrle:
I just rebased but coqbot still says a rebase is needed.  (This page also says "This branch has no conflicts with the base branch".)  My local copy of master is current.  Not sure what the problem is.

Zimmi48:
>I just rebased but coqbot still says a rebase is needed. (This page also says "This branch has no conflicts with the base branch".) My local copy of master is current. Not sure what the problem is.

Sorry Jim. There were issues with the bot following the recent improvements made so that it does not request a rebase when there is no conflict with the base branch. In principle, all the issues were fixed, but if you do observe a problem with the bot in the future, do not hesitate to ping me.

jfehrle:
Thanks for the reviews and merge!  Hope you're having a great holiday.

Zimmi48:
The possibility of having this PR included in 8.9.0 or 8.9.1 has never been debated, has it?

jfehrle:
I certainly have no objection to putting this into 8.9 or 8.9.1.  I do think it's low risk.  My assumption has been that anything printed by Printer is going to lex--though that's not the same as being certain.  After including part of this PR in https://github.com/coq/coq/pull/9101, all of the proofs under `theories` diff successfully (there were only a couple failures).

This PR was aimed at making the code more bulletproof in case anyone tried to apply the base diff routines to something other than a proof context--clearly requiring a code change in OCaml.

(I saw some recent discussion about the current plan for 8.9.0, but I don't know the resolution--is it indeed delayed for a couple months?  And what are the criteria for something to go into 8.9 at this point?)

Zimmi48:
>Is it indeed delayed for a couple months?

In the end, it is not. Everything should be ready and the tag will probably be put early January.

>And what are the criteria for something to go into 8.9 at this point?

I'm not sure. But I wouldn't aim for 8.9.0 now (as I said, it's mostly ready) but rather for 8.9.1. So let's just add this to the milestone / project, and see what the RM says.

Zimmi48:
In fact, I confused this PR with #9101, which had already been merged in the 8.9.0 milestone but has not been backported yet. See my reply to the coqdev thread "Tagging 8.9.0 from Dec 17th".

PR #8737, opened 2018-10-15
Correctly report non-projection fields in records

### LINKED ISSUE #8736: Mixing in non-projection fields in a record results in an incorrect error message
#### Version

8.10+alpha

#### Operating system

macOS

When a record contains fields that are projections as well as fields that aren't projections, the error message is "This record contains fields of different records" when the non-projections should be reported instead.

```coq
Record t :=
  { foo: nat }.

Fail Check {| unit := 2 |}.
(* unit: Not a projection *)

Fail Check {| unit := 2;
              foo := 2 |}.
(* unit: Not a projection *)

Check {| foo := 2;
         unit := 2 |}.
(* "This record contains fields of different records"
should be "unit: Not a projection" *)
```

tchajed:
**Kind:** bug fix.

Fixes #8736

### DIFF
--- a/interp/constrintern.ml
+++ b/interp/constrintern.ml
@@ -122,6 +122,7 @@ type internalization_error =
   | BadPatternsNumber of int * int
   | NotAProjection of qualid
   | NotAProjectionOf of qualid * qualid
+  | ProjectionsOfDifferentRecords of qualid * qualid
 
 exception InternalizationError of internalization_error Loc.located
 
@@ -153,6 +154,10 @@ let explain_field_not_a_projection field_id =
 let explain_field_not_a_projection_of field_id inductive_id =
   pr_qualid field_id ++ str ": Not a projection of inductive " ++ pr_qualid inductive_id
 
+let explain_projections_of_diff_records inductive1_id inductive2_id =
+  str "This record contains fields of both " ++ pr_qualid inductive1_id ++
+  str " and " ++ pr_qualid inductive2_id
+
 let explain_internalization_error e =
   let pp = match e with
   | VariableCapture (id,id') -> explain_variable_capture id id'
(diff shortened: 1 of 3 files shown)


ppedrot (approved):
LGTM.

ppedrot:
That's quite a trivial fix. @herbelin if you don't complain I'll merge quickly.

tchajed:
I'd actually like to include the name of the inductive from the first projection in this error message (because it's known), but I'm not sure how to access it. Is there an easy to way to get it? If not it's ok, this is still an improvement.

ejgallego (changes requested):
Extend `internalization` error and display the error from the upper layers.

ejgallego (approved):
Ok, this looks good to me. It should help to have better error message in the way @ppedrot mentioned; but you don't need to do that in this PR indeed as it will involve a lot of surgery.

herbelin:
>  @herbelin if you don't complain I'll merge quickly.

I don't complain, well on the contrary.

Regarding the extension `internalization_error`, I approve, because it gives a better overall view to possible errors of the file.

We should however not restrict ourselves to write more informative error messages because of the cost induced by moving the printers to the upper layers. Moreover, there is an easy way to do it here. To get the user-level pretty-printed name, as hinted by @ppedrot, something like
 `shortest_qualid_of_global ?loc Id.Set.empty (IndRef (inductive_of_constructor record.Recordops.s_CONST))` should work.

tchajed:
Thanks for the tip @herbelin! I’ll update the pull request tomorrow and improve the error message

tchajed:
Done, and also added the conflicting records to the message "This record contains fields of different records".

ppedrot:
@tchajed CI is complaining for spurious reasons, can you rebase atop of a recent master?

ppedrot:
The CI failure is spurious, I'm merging the PR.

PR #8738, opened 2018-10-15
[clib] Deprecate string functions available in OCaml 4.05

ejgallego:
- `CString.strip -> String.trim`
- `CString.split -> String.split_on_char`

### DIFF
--- a/clib/cString.ml
+++ b/clib/cString.ml
@@ -18,6 +18,7 @@ sig
   val explode : string -> string list
   val implode : string list -> string
   val strip : string -> string
+  [@@ocaml.deprecated "Use [trim]"]
   val drop_simple_quotes : string -> string
   val string_index_from : string -> int -> string -> int
   val string_contains : where:string -> what:string -> bool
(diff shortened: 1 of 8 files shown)


ppedrot:
Note that the semantics of both variants are marginally different. For instance, OCaml `trim` also takes line feeds (LF) into account. Similarly, OCaml `split` never returns an empty list whereas Coq `trim` does on the empty string.

ejgallego:
> Note that the semantics of both variants are marginally different. For instance, OCaml `trim` also takes line feeds (LF) into account. Similarly, OCaml `split` never returns an empty list whereas Coq `trim` does on the empty string.

[You mean split not trim, right].

Noted in the commit message, IMHO the OCaml semantics are preferred, if only for the lower maintenance burden.

ppedrot:
If CI compiles I'll merge this quickly.

ppedrot (approved):

gares (approved):

ppedrot:
CI failure is unrelated.

PR #8740, opened 2018-10-16
Removing the Camlp5 macros from CLexer.

ppedrot:
We partially hand-translated so as to result in the minimal diff possible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -135,7 +135,6 @@ coqpp/coqpp_parse.mli
 g_*.ml
 
 lib/coqProject_file.ml
-parsing/cLexer.ml
 plugins/ltac/coretactics.ml
 plugins/ltac/extratactics.ml
 plugins/ltac/extraargs.ml
--- a/Makefile.build
+++ b/Makefile.build
@@ -406,7 +406,7 @@ grammar/%.cmi: grammar/%.mli
 
 .PHONY: coqbinaries coqbyte
 
-coqbinaries: $(TOPBINOPT) $(COQTOPEXE) $(CHICKEN) $(CSDPCERT) $(FAKEIDE)
+coqbinaries: $(TOPBINOPT) $(COQTOPEXE) $(CHICKEN) $(CSDPCERT) $(FAKEIDE) $(GRAMMARCMA)
 coqbyte: $(TOPBYTE) $(CHICKENBYTE)
 
 # Special rule for coqtop, we imitate `ocamlopt` can delete the target
--- a/parsing/cLexer.ml
+++ b/parsing/cLexer.ml
@@ -229,9 +229,11 @@ let unlocated f x = f x
 (*   try f x with Loc.Exc_located (_, exc) -> raise exc *)
 
 let check_keyword str =
-  let rec loop_symb = parser
-    | [< ' (' ' | '\n' | '\r' | '\t' | '"') >] -> bad_token str
-    | [< s >] ->
+  let rec loop_symb s = match Stream.peek s with
+    | Some (' ' | '\n' | '\r' | '\t' | '"') ->
+        Stream.junk s;
+        bad_token str
+    | _ ->
         match unlocated lookup_utf8 Ploc.dummy s with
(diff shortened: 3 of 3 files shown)


ppedrot:
It looks like no one triggers the compilation of grammar.cma now, which makes the installation phase fail... I'll add some hardcoded require rule I guess...

Also, @ejgallego you should probably be the assignee for that one.

Makefile.install
```diff
@@ -99,7 +99,7 @@ INSTALLCMX = $(sort $(filter-out checker/% ide/% tools/% dev/% \
 	configure.cmx toplevel/coqtop_byte_bin.cmx plugins/extraction/big.cmx, \
 	$(MLFILES:.ml=.cmx)))
 
-install-devfiles:
+install-devfiles: $(GRAMMARCMA)
```
SkySkimmer: Don't put it here, instead put it as a dependency of coqbinaries in makefile.build (I think that's the best place for this) (coqbinaries is a dependency of world)

ejgallego:
> Also, @ejgallego you should probably be the assignee for that one.

Ok, I can take this one, but I'd like to hear what @herbelin says. I'd also like the approval of some make-guru.

Also, as @SkySkimmer asked, would we benefit from moving this ocamlex in the future?

Makefile.install
ppedrot: Makes sense. As recommended by @ejgallego I'll wait for input from other Make specialists before proceeding further though.

ejgallego: Anyways the original dependency on this was a bug of the makefile as `CLexer` has never needed `grammar`. It is a regular camlp5 file.

ejgallego: The thing is that the whole pre-reprocessing story of the makefiles is a bit "botched".

ejgallego: The sanest option would be to migrate to Dune; the least insane option after that is imitate what Dune does and have the makefiles generate a `coqp5` binary to handle `ml4` processing.

gares: I agree with @SkySkimmer , install targets do not depend on the things they install... it is silly, but it is the way things are

ppedrot: Change applied.

ppedrot:
@ejgallego I don't know what ocamllex would do in the picture. It's not the same API at all, is not dynamically extensible and does not handle unicode natively...

ejgallego:
Ok, waiting to hear from @herbelin

herbelin:
I don't particularly find the new code more readable, nor easier to maintain and extend, since you now have to know when to manually insert `Stream.junk`, `Stream.count`, etc., but if no better compromise is available, I'll live with it.

ejgallego:
> if no better compromise is available, I'll live with it.

Indeed the new code is a concern. What compromise do you have in mind?

This file is the last user of `camlp5` in the [proper] coq repository, tho, so the dep is quite heavy IMHO for what it does.

ejgallego (approved):
This makes the lexer marginally harder to understand, but depending on the whole of `camlp5` just for this seems a bit heavy, thus I am OK, the tradeoff seems acceptable to me.

I'm also convinced the lexer could be greatly refactored and the downsides of this PR would diminish.

herbelin:
The only compromise I see would be to include the relevant parts of camlp5 (or a modified copy of it) but I have no idea how big it is. If too big, I don't see (as far as I'm concerned) better compromise than the current PR.

ejgallego:
I shall merge tomorrow I think unless someone feels we should wait more. Should be easy to revert if we find a problem or discover another solution.

ppedrot:
> correct me if I am wrong

You're right, indeed.

ppedrot:
@herbelin Making the stream code purely functional would make it way more palatable, but that requires changing the Camlp5 API (or use the fstream one, but that implies backtracking parsing iirc). When we use our own variant of camlp5 grammar engine this kind of change will be easily performed.

ejgallego:
> You're right, indeed.

Ok, I am merging this then.

PR #8746, opened 2018-10-16
Provide nat numeral syntax as a standalone import

tchajed:
**Kind:** feature.

### DIFF
--- a/theories/Init/NatNumeralNotation.v
+++ b/theories/Init/NatNumeralNotation.v
@@ -1,0 +1,15 @@
+Require Coq.Init.Datatypes.
+Require Coq.Init.Nat.
+
+Declare ML Module "numeral_notation_plugin".
+
+(* Parsing / printing of decimal numbers *)
+Arguments Nat.of_uint d%dec_uint_scope.
+Arguments Nat.of_int d%dec_int_scope.
+Numeral Notation Decimal.uint Decimal.uint_of_uint Decimal.uint_of_uint
+  : dec_uint_scope.
+Numeral Notation Decimal.int Decimal.int_of_int Decimal.int_of_int
+  : dec_int_scope.
+
+(* Parsing / printing of [nat] numbers *)
+Numeral Notation Datatypes.nat Nat.of_uint Nat.to_uint : nat_scope (abstract after 5000).
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,13 @@ Tactics
   Simplex-based proof engine. In case of regression, 'Unset Simplex'
   to get the venerable Fourier-based engine.
 
+Standard library
+
+- For users running Coq without the standard library ("-noinit" or "-nois"),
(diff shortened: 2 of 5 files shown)


tchajed:
@JasonGross am I missing some good reason why this shouldn't be provided? It seems to work as far as I can tell.

ppedrot:
Why not put this code directly in `Nat`? Is there some runtime dependency that will fail?

tchajed:
I think that would work, but keeping them separate means a user could use `Coq.Init.Nat` without the numeral notations plugin.

theories/Init/NatNumeralNotation.v
```diff
@@ -0,0 +1,15 @@
+Require Coq.Init.Datatypes.
+Require Coq.Init.Nat.
+
+Declare ML Module "numeral_notation_plugin".
+
+(* Parsing / printing of decimal numbers *)
+Arguments Nat.of_uint d%dec_uint_scope.
+Arguments Nat.of_int d%dec_int_scope.
+Numeral Notation Decimal.uint Decimal.uint_of_uint Decimal.uint_of_uint
```
JasonGross: Either this line should be somewhere else, or this file is misnamed; this is for numeral syntax for Decimal.

JasonGross (changes requested):
Is each ML Module supposed to be declared only once, or is it okay to declare them in multiple places (I know that some are declared in the compat files without much issue).  And does Declare ML Module follow Require, or Import?

It's not entirely clear to me why, say, the numeral notation ML isn't declared in the Decimal file.

My guess is that the numeral notations were put in Prelude because:
- Init.Datatypes is unsuitable because the conversions are defined in Init.Nat
- Init.Nat and Init.Decimal are unsuitable because they are not `Export`ed by the Prelude, only `Require`d, and numeral notations follow `Import`
- It was deemed unnecessary to add an extra file to the Prelude just to support numeral notations.
- They were not declared in their own module in Decimal / Nat and then re-exported in Prelude to avoid cluttering the module namespace.

Do you have a concrete use case for this in mind?  The only use cases I've seen for -noinit are HoTT and occasional bug test cases that depend on specific stdlib things.

I am inclined against this change unless there's a use-case in mind, but other than my comment about the file being misnamed, I will not stand in the way of this being merged.

ppedrot (commented):
Due to the fact I am the owner of Init I am supposed to be in charge of this PR, but I don't think it really has to do with what I am supposed to handle. So I'll follow @JasonGross advices on this topic, and also summon people knowledgeable about syntax.

tchajed:
I don't actually have a use case, I just saw the note in the changelog and it struck me as odd. If someone were to seriously start using Coq with an alternate standard library (along the lines of Haskell's [classy-prelude](https://hackage.haskell.org/package/classy-prelude) then it'd probably be handy to re-export nat syntax, but till then I'll just close the PR.

Thanks for taking a look @JasonGross!

PR #8747, opened 2018-10-16
Delay Universe Substitution in CClosure

### LINKED ISSUE #8721: Consider delaying universe substitutions during reduction.
@Janno and I did a quick implementation of this in the copy of cClosure being used for mTac and it reduced reduction time by about 40% for their test case.  Apparently this is because when dealing with heavily universe polymorphic code, by default Coq produces different universes for each constant invocation; therefore, the constant cache has a high miss rate.

What we did doesn't solve the high miss rate problem, but it does make adding universe polymorphic constants to the cache much less costly, by keeping a delayed universe substitution in every fterm and stack_member that currently has a regular delayed substitution (as opposed to the current implementation, which substitutes the universe instance through the whole constant body the first time it's encountered).  Actually substituting is deferred until mk_clos is called on something that needs it (i.e. a Const, Ind, Construct, or Sort), and otherwise we just propagate the universe substitution down.

This might lead to a bit of slowdown or increased memory usage in cases where there is no universe polymorphism, but in other cases I suspect it's usually faster, since we avoid substituting into subterms of the constant body that are never converted to fterms.

I can work with @Janno to provide a PR that applies to the real kernel if people are interested.

Janno:
This patch (written together with @pythonesque) addresses the simpler proposal from #8721. We tried the other, more complicated variant but currently that is slow and does not work.

In local benchmarks we see a slight slowdown (0.1%, could be noise) in building the standard library but HoTT got a 2.5% improvement (and Mtac2 also saw a measurable improvement with similar changes to its very outdated copy of cClosure.ml).

**Kind:** performance.

Fixes / closes #8721

### DIFF
--- a/kernel/cClosure.ml
+++ b/kernel/cClosure.ml
@@ -537,14 +537,14 @@ let destFLambda clos_fun t =
 
 (* Optimization: do not enclose variables in a closure.
    Makes variable access much faster *)
-let subst_if_not_empty subst u = if Univ.Instance.is_empty u then u else Univ.subst_instance_instance subst u
+let subst_if_not_empty subst u = if Univ.Instance.is_empty subst then u else Univ.subst_instance_instance subst u
 let mk_clos (e : fconstr fusubs) t =
   match kind t with
   | Rel i -> clos_rel (fst e) i
     | Var x -> { norm = Red; term = FFlex (VarKey x) }
-  | Const (c,u) -> { norm = Red; term = FFlex (ConstKey (c, subst_instance_instance (snd e) u)) }
+  | Const (c,u) -> { norm = Red; term = FFlex (ConstKey (c, subst_if_not_empty (snd e) u)) }
   | Meta _ -> { norm = Norm; term = FAtom t }
-  | Sort (Sorts.Type u) ->  { norm = Norm; term = FAtom (mkSort (Sorts.Type (subst_instance_universe (snd e) u))) }
+  | Sort (Sorts.Type u) ->  { norm = Norm; term = FAtom (mkSort (Sorts.Type (if Instance.is_empty (snd e) then u else subst_instance_universe (snd e) u))) }
   | Sort _ -> { norm = Norm; term = FAtom t }
   | Ind (i, u) -> { norm = Norm; term = FInd (i, subst_if_not_empty (snd e) u) }
   | Construct (c, u) -> { norm = Cstr; term = FConstruct (c, subst_if_not_empty (snd e) u) }
--- a/kernel/cClosure.ml
+++ b/kernel/cClosure.ml
@@ -284,16 +284,6 @@ let assoc_defined id env = match Environ.lookup_named id env with
 
 open Declarations
 
-let constant_value_in env (kn,u) =
-  let cb = lookup_constant kn env in
-  match cb.const_body with
-  | Def l_body ->
-      let b = Mod_subst.force_constr l_body in
-      (* subst_instance_constr u b *)
-      (b,u)
-  | OpaqueDef _ -> raise (NotEvaluableConst Opaque)
-  | Undef _ -> raise (NotEvaluableConst NoBody)
-
 let ref_value_cache ({i_cache = cache;_} as infos) tab ref =
   try
     Some (KeyTable.find tab ref)
@@ -313,7 +303,7 @@ let ref_value_cache ({i_cache = cache;_} as infos) tab ref =
           | LocalDef (_, t, _) -> lift n t, Instance.empty
           end
         | VarKey id -> assoc_defined id cache.i_env, Instance.empty
-	| ConstKey cst -> constant_value_in cache.i_env cst
+        | ConstKey (cst, u) -> constant_value_in cache.i_env (cst, Instance.empty), u
     in
     let v = cache.i_repr infos tab (body, u) in
     KeyTable.add tab ref v;
@@ -552,9 +542,9 @@ let mk_clos (e : fconstr fusubs) t =
   match kind t with
   | Rel i -> clos_rel (fst e) i
     | Var x -> { norm = Red; term = FFlex (VarKey x) }
-  | Const (c,u) -> { norm = Red; term = FFlex (ConstKey (c, Univ.subst_instance_instance (snd e) u)) }
+  | Const (c,u) -> { norm = Red; term = FFlex (ConstKey (c, subst_instance_instance (snd e) u)) }
   | Meta _ -> { norm = Norm; term = FAtom t }
-  | Sort (Sorts.Type u) ->  { norm = Norm; term = FAtom (mkSort (Sorts.Type (Univ.subst_instance_universe (snd e) u))) }
+  | Sort (Sorts.Type u) ->  { norm = Norm; term = FAtom (mkSort (Sorts.Type (subst_instance_universe (snd e) u))) }
   | Sort _ -> { norm = Norm; term = FAtom t }
   | Ind (i, u) -> { norm = Norm; term = FInd (i, subst_if_not_empty (snd e) u) }
   | Construct (c, u) -> { norm = Cstr; term = FConstruct (c, subst_if_not_empty (snd e) u) }
@@ -587,9 +577,9 @@ let rec to_constr lfts v =
         mkCase (ci, p, to_constr lfts c, ve)
       else
         let subs = comp_subs lfts env in
-        mkCase (ci, subst_constr subs p,
+        mkCase (ci, subst_constr (subs, snd env) p,
             to_constr lfts c,
-            Array.map (fun b -> subst_constr subs b) ve)
+            Array.map (fun b -> subst_constr (subs, snd env) b) ve)
     | FFix ((op,(lna,tys,bds)) as fx, e) ->
       if is_subs_id (fst e) && is_lift_id lfts && Instance.is_empty (snd e) then
         mkFix fx
@@ -597,8 +587,8 @@ let rec to_constr lfts v =
         let n = Array.length bds in
         let subs_ty = comp_subs lfts e in
         let subs_bd = comp_subs (el_liftn n lfts) (fstapp (subs_liftn n) e) in
-        let tys = Array.Fun1.map subst_constr subs_ty tys in
-        let bds = Array.Fun1.map subst_constr subs_bd bds in
+        let tys = Array.Fun1.map subst_constr (subs_ty, snd e) tys in
+        let bds = Array.Fun1.map subst_constr (subs_bd, snd e) bds in
         mkFix (op, (lna, tys, bds))
     | FCoFix ((op,(lna,tys,bds)) as cfx, e) ->
       if is_subs_id (fst e) && is_lift_id lfts && Instance.is_empty (snd e) then
@@ -607,8 +597,8 @@ let rec to_constr lfts v =
         let n = Array.length bds in
         let subs_ty = comp_subs lfts e in
         let subs_bd = comp_subs (el_liftn n lfts) (fstapp (subs_liftn n) e) in
-        let tys = Array.Fun1.map subst_constr subs_ty tys in
-        let bds = Array.Fun1.map subst_constr subs_bd bds in
+        let tys = Array.Fun1.map subst_constr (subs_ty, snd e) tys in
+        let bds = Array.Fun1.map subst_constr (subs_bd, snd e) bds in
         mkCoFix (op, (lna, tys, bds))
     | FApp (f,ve) ->
         mkApp (to_constr lfts f,
@@ -621,8 +611,8 @@ let rec to_constr lfts v =
         Term.compose_lam (List.rev tys) f
       else
         let subs = comp_subs lfts e in
-        let tys = List.mapi (fun i (na, c) -> na, subst_constr (subs_liftn i subs) c) tys in
-        let f = subst_constr (subs_liftn len subs) f in
+        let tys = List.mapi (fun i (na, c) -> na, subst_constr (subs_liftn i subs, snd e) c) tys in
+        let f = subst_constr (subs_liftn len subs, snd e) f in
         Term.compose_lam (List.rev tys) f
     | FProd (n,t,c)   ->
         mkProd (n, to_constr lfts t,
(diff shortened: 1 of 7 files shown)


Janno:
It would be very useful to run the benchmarks on this. It might introduce small slowdowns in many projects that do not use universe polymorphism.

SkySkimmer:
The stdlib uses universe polymorphism very little.
Bench running: https://ci.inria.fr/coq/view/benchmarking/job/benchmark-part-of-the-branch/547/console

kernel/cClosure.mli
```diff
@@ -179,7 +187,7 @@ val unfold_projection : 'a infos -> Projection.t -> stack_member option
    [create_clos_infos], inject the term to reduce with [inject]; then use
    a reduction function *)
 
-val inject : constr -> fconstr
+val inject : constr puniverses -> fconstr
```
SkySkimmer: The type of this function should probably not have been changed. In the same way that it assumes the identity substitution for variables it should assume no universe substitution.

kernel/cClosure.ml
```diff
@@ -520,22 +527,25 @@ let mk_lambda env t =
...
     | FLambda(n,(na,ty)::tys,b,e) ->
-        (na,clos_fun e ty,{norm=Cstr;term=FLambda(n-1,tys,b,subs_lift e)})
+      (na,clos_fun e ty,{norm=Cstr;term=FLambda(n-1,tys,b,fstapp subs_lift e)})
     | _ -> assert false
 	(* t must be a FLambda and binding list cannot be empty *)
 
 (* Optimization: do not enclose variables in a closure.
    Makes variable access much faster *)
-let mk_clos e t =
+let subst_if_not_empty subst u = if Univ.Instance.is_empty subst then u else Univ.subst_instance_instance subst u
```
SkySkimmer: This should probably be included in the Univ function.

kernel/cClosure.ml
```diff
@@ -561,32 +571,32 @@ let rec to_constr lfts v =
     | FInd op -> mkIndU op
     | FConstruct op -> mkConstructU op
     | FCaseT (ci,p,c,ve,env) ->
-      if is_subs_id env && is_lift_id lfts then
+      if is_subs_id (fst env) && is_lift_id lfts && Instance.is_empty (snd env) then
```
SkySkimmer: Make a function is_fusubs_id for these tests.

SkySkimmer:
Mixed results:
~~~

┌──────────────────────────┬─────────────────────────┬───────────────────────────────────────┬───────────────────────────────────────┬──────────────────────────┬──────────────────────┐
│                          │      user time [s]      │              CPU cycles               │           CPU instructions            │  max resident mem [KB]   │      mem faults      │
│                          │                         │                                       │                                       │                          │                      │
│             package_name │     NEW     OLD PDIFF   │            NEW            OLD PDIFF   │            NEW            OLD PDIFF   │     NEW     OLD  PDIFF   │  NEW  OLD    PDIFF   │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│                 coq-hott │  332.92  341.64 -2.55 % │   923345431138   946156146841 -2.41 % │  1406642863391  1440100985649 -2.32 % │  643632  666792  -3.47 % │  430    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│                coq-sf-lf │   44.21   44.42 -0.47 % │   121743835280   122546861357 -0.66 % │   199315596414   199501609977 -0.09 % │  423608  422796  +0.19 % │  128    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│                coq-flocq │  590.32  592.00 -0.28 % │  1640292423178  1647156759308 -0.42 % │  2411354196739  2405224361056 +0.25 % │ 1764632 1766556  -0.11 % │  606  867   -30.10 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│      coq-formal-topology │   58.08   58.24 -0.27 % │   158304951558   158865962983 -0.35 % │   239184450169   239985754037 -0.33 % │  477836  477416  +0.09 % │    0    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│   coq-mathcomp-odd-order │ 1417.46 1420.48 -0.21 % │  3948768967038  3956033844700 -0.18 % │  6699320818703  6658637427858 +0.61 % │ 1350972 1359328  -0.61 % │  261   49  +432.65 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│             coq-compcert │  889.88  891.51 -0.18 % │  2467456059069  2470265840463 -0.11 % │  3578096037308  3561391545432 +0.47 % │ 1343276 1349328  -0.45 % │  617  339   +82.01 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│         coq-math-classes │  249.50  249.83 -0.13 % │   687224056760   688445471139 -0.18 % │   900867275902   899805525097 +0.12 % │  523152  522136  +0.19 % │   84   81    +3.70 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│         coq-fiat-parsers │  719.68  719.13 +0.08 % │  1998538693108  1996036522423 +0.13 % │  3080457738017  3054736560975 +0.84 % │ 2880308 2898680  -0.63 % │  590  761   -22.47 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│               coq-sf-plf │   74.76   74.68 +0.11 % │   206677543407   206541556383 +0.07 % │   301500338389   300926704664 +0.19 % │  521084  518568  +0.49 % │   15    0     +nan % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│               coq-sf-vfa │   47.64   47.56 +0.17 % │   130990748145   131378232688 -0.29 % │   209993439783   209968706004 +0.01 % │  542076  533452  +1.62 % │   28    5  +460.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│            coq-fiat-core │  135.46  135.22 +0.18 % │   376724696562   377146921844 -0.11 % │   504668380099   504069509866 +0.12 % │  499292  497676  +0.32 % │  613  188  +226.06 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│   coq-mathcomp-ssreflect │   65.66   65.54 +0.18 % │   180023791228   180139505878 -0.06 % │   257095867228   256804442513 +0.11 % │  532368  532128  +0.05 % │   26  257   -89.88 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│          coq-fiat-crypto │ 6397.87 6379.31 +0.29 % │ 17792985556962 17744074236815 +0.28 % │ 29252536240733 29120658840711 +0.45 % │ 2723356 2450576 +11.13 % │ 1735 1347   +28.80 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│                coq-color │  740.67  738.32 +0.32 % │  2053660253546  2044831815430 +0.43 % │  2478563555420  2468606887351 +0.40 % │ 1371488 1375072  -0.26 % │   39  113   -65.49 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│    coq-mathcomp-fingroup │   85.43   85.12 +0.36 % │   236125046171   235251178827 +0.37 % │   350268845326   348848446344 +0.41 % │  594012  586388  +1.30 % │    4   48   -91.67 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│           coq-coquelicot │  104.28  103.85 +0.41 % │   287965548005   287163328551 +0.28 % │   388520908545   387043551167 +0.38 % │  725296  712312  +1.82 % │  282   13 +2069.23 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│                 coq-corn │ 1552.32 1545.75 +0.43 % │  4309382166395  4292619603160 +0.39 % │  6377647772888  6347727518034 +0.47 % │  926048  818152 +13.19 % │   45   60   -25.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│               coq-geocoq │ 2850.83 2838.13 +0.45 % │  7925675384464  7888923000596 +0.47 % │ 12703682815750 12579309110573 +0.99 % │ 1287824 1287704  +0.01 % │  606  190  +218.95 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│              coq-bignums │   99.47   98.88 +0.60 % │   274747249868   273697891152 +0.38 % │   394665541113   393829024474 +0.21 % │  536180  539500  -0.62 % │  175   43  +306.98 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│          coq-lambda-rust │ 2343.50 2329.46 +0.60 % │  6526599399602  6488604622730 +0.59 % │  9033098570801  8964156114014 +0.77 % │ 1544608 1450244  +6.51 % │  141   15  +840.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│   coq-mathcomp-character │  296.24  294.30 +0.66 % │   823620010282   817415400929 +0.76 % │  1197870057955  1191274757722 +0.55 % │ 1059976 1117428  -5.14 % │   45  302   -85.10 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│    coq-mathcomp-solvable │  222.11  220.18 +0.88 % │   617364103658   611819400697 +0.91 % │   882438288585   875349132855 +0.81 % │  807792  855168  -5.54 % │    2    1  +100.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│                  coq-vst │ 1728.61 1712.80 +0.92 % │  4806190843916  4762314463771 +0.92 % │  6562293639934  6513293300621 +0.75 % │ 2447808 2225288 +10.00 % │ 1166  868   +34.33 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│              coq-unimath │ 1343.67 1331.10 +0.94 % │  3731593916573  3693996753506 +1.02 % │  6317397679880  6240478754512 +1.23 % │ 1133008  971376 +16.64 % │  356   19 +1773.68 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│     coq-mathcomp-algebra │  208.10  205.23 +1.40 % │   577371493468   569418457780 +1.40 % │   810525479858   799120309546 +1.43 % │  648824  640676  +1.27 % │  293   14 +1992.86 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│ coq-mathcomp-real-closed │  190.92  188.10 +1.50 % │   531068217299   523597662881 +1.43 % │   804622480648   790375813147 +1.80 % │  853992  830204  +2.87 % │    1    1    +0.00 % │
├──────────────────────────┼─────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┼──────────────────────┤
│       coq-mathcomp-field │  475.01  467.97 +1.50 % │  1321459762740  1301256230599 +1.55 % │  2164466154596  2127074570257 +1.76 % │  805852  802264  +0.45 % │    2   19   -89.47 % │
└──────────────────────────┴─────────────────────────┴───────────────────────────────────────┴───────────────────────────────────────┴──────────────────────────┴──────────────────────┘
~~~

SkySkimmer:
I think unimath is the other big universe polymorphism user, it's interesting that they get no benefit.

Janno:
I reverted the change to `inject`'s type signature. Additionally, I added a fast path that skips empty variable substitutions and performs only universe substitution (if that is non-empty, of course). This latter change shows great potential in a local benchmark for UniMath and doesn't seem to slow down HoTT compared to the previous version of this patch.

The code for that is rather ugly right because it introduces a let binding for now. I already know how to write this differently but given my limited success with guessing the performance of ocaml code I would have to rerun my local benchmarks and that is rather annoying.

Could we have another run of the benchmark suite? I don't want to hog that resource too much so feel free to make me do more local benchmarks. I just don't know if I am doing them correctly so I put  limited trust in them.

kernel/cClosure.ml
mattam82: I guess subst_instance_instance subst u != u (using physical equality), but if the code is observing that we have a problem.

mattam82: Actually, subst_instance_instance uses a smart array map, so they should even be physically equal. Except if you call subst_instance_instance with an empty substitution while the instance has some free universe variables. Then you get an array out of bounds exception. Could your code be doing that?

kernel/cClosure.ml
```diff
@@ -595,43 +614,52 @@ let rec to_constr lfts v =
...
+      else if is_id1 then
+ subst_instance_constr (snd env) (t)
       else
         let subs = comp_subs lfts env in
-        subst_constr subs t
+        subst_constr (subs, snd env) t
     | FLOCKED -> assert false (*mkVar(Id.of_string"_LOCK_")*)
 
-and subst_constr subst c = match Constr.kind c with
+and subst_constr (subst, inst) c = match Constr.kind c with
```
SkySkimmer: I think you should curry this pair away.

kernel/cClosure.ml
mattam82: Ok, I see what you're doing now. When getting a constant body from the cache you pass Instance.empty so no universe substitution happens yet. Then you do the delayed substitutions with the actual requested instance. If some case of subst_instance_instance raises then it must be because somewhere you lost a substitution. I think you should define generalized compose_subs and lift_subs combinators to avoid all those fst / snd projections, that would make the change clearer.

kernel/cClosure.ml
Janno: it's used as an argument to some combinators (`map_with_binders, Array.Fun1.map`) so I would probably have to uncurry the function in these places, wouldn't I?

Janno:
No need to run the benchmarks. The improvements I saw locally come from other commits that I included via a rebase. Too bad. :(

ppedrot:
Should we consider this as a failed experiment?

pythonesque:
Probably.  I think we identified the reason for such large slowdowns (it is because we can actually do significantly *more* work [repeatedly reapplying the same substitutions] when we defer universe substitutions in this way), but it's not clear how to solve it, or whether such a solution would be complex enough that the extra overhead would outweigh its benefits.

ppedrot:
Let's archive this PR then. Hopefully we can find an alternative solution that proves to be more efficient on average.

PR #8750, opened 2018-10-17
[ci] [doc] Notes about branch names.

ejgallego:
I'd like to add this convention as it is very convenient for the
development of dev tools. Example, I can do `setup-coq-devs ltac
equations` and then get a fully composed tree. Similarly for preparing
overlays.

### DIFF
--- a/dev/ci/README.md
+++ b/dev/ci/README.md
@@ -26,7 +26,8 @@ our CI. This means that:
 
 On the condition that:
 
-- At the time of the submission, your development works with Coq master branch.
+- At the time of the submission, your development works with Coq's
+  `master` branch.
 
 - Your development is publicly available in a git repository and we can easily
   send patches to you (e.g. through pull / merge requests).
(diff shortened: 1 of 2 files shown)


SkySkimmer (approved):
I'm fine with it.

Zimmi48:
Since I don't know what is `setup-coq-devs`, it's hard to comment on this proposal but I already expressed my worries about the recommended convention here: https://github.com/ybertot/plugin_tutorials/issues/9

ejgallego:
> Since I don't know what is `setup-coq-devs`, it's hard to comment on this proposal but I already expressed my worries about the recommended convention here: [ybertot/plugin_tutorials#9](https://github.com/ybertot/plugin_tutorials/issues/9)

Well, the only worry I see there so far is recommending `coq-8.8` vs `v8.8`. IMHO that's not worth the gain as people then would likely get it wrong, name it differently. IMHO same name is what makes most sense.

Zimmi48 (approved):
OK, as long as it is only a recommendation.

Zimmi48:
Let me ping @gares here too since he is also one of the users of the coq-v8.8 convention...

ejgallego:
> It would be written down after this PR.

The whole point of the convention is not to have to look at docs, but only at the source code. I dunno, I am following this pattern based on what other convention-based system do [RoR lol, old times]

Of course is only a recommendation, but allows you to skip defining a branch name.

ejgallego:
Note that some people find conventions "obscure" or "too magical" sometimes, I dunno. But it does bothers me a bit the diverse variety of branches we have. [And despite pinning we still need a branch name to update the pinning].

Zimmi48:
>And despite pinning we still need a branch name to update the pinning

Absolutely, yes!

SkySkimmer:
>The whole point of the convention is not to have to look at docs, but only at the source code.

What source code?

ejgallego:
> What source code?

I mean Coq's git repos. For the branch name.

SkySkimmer:
Since @gares was pinged I'm going to wait for him to say something.

ejgallego:
Ping @gares

gares:
If you name the branch with a version it may clash with a branch of the sw itself. Hence I prefer to prefix it with coq-

SkySkimmer:
In order to avoid being stuck in bikeshed limbo I'll merge tomorrow (barring objections).

gares:
you can merge it now, for me

PR #8752, opened 2018-10-17
Enable fragile pattern warning in cclosure

SkySkimmer:
This file is already mostly in the required style so I wanted to see
what it looks like.

For a couple matches I locally disabled the warning as it was too
annoying otherwise (`when` clauses are especially annoying).

There are a couple places where I think it clearly looks better (eg
assoc_defined at the beginning of the file) but overall I'm not all
that convinced.
What do other people think?

### DIFF
--- a/kernel/cClosure.ml
+++ b/kernel/cClosure.ml
@@ -21,6 +21,8 @@
 (* This file implements a lazy reduction for the Calculus of Inductive
    Constructions *)
 
+[@@@ocaml.warning "+4"]
+
 open CErrors
 open Util
 open Pp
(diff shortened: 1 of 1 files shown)


ejgallego:
For this particular file and component I like the patch.

ppedrot (approved):
I am not a fan of locally disabling the flag, but I like the idea anyways.

maximedenes (approved):

maximedenes:
I don't think this requires a big discussion, it is kernel-only and makes maintainers happy. @SkySkimmer can I merge?

SkySkimmer:
yes

Gaëtan Gilbert

ppedrot:
@maximedenes ping.

maximedenes:
Failure is spurious, merging.

PR #8753, opened 2018-10-17
[build] Refactoring of config lib and ocamldebug tweaks.

ejgallego:
We make `config` into a proper library. This is more uniform and
useful for some clients. This also matches what was done in Dune.

Next step would be to push dependencies on `Coq_config` upwards, only
the actual toplevel binaries should depend on it.

We also remove the stale `camlp5.dbg` and refactor the dbg files a
bit, isolating the bits that are specific to the plugin / lib building
method used by makefile.


Makefile
```diff
@@ -261,7 +261,7 @@ cacheclean:
 	find theories plugins test-suite -name '.*.aux' -exec rm -f {} +
 
 cleanconfig:
-	rm -f config/Makefile config/coq_config.ml myocamlbuild_config.ml dev/ocamldebug-coq dev/camlp5.dbg config/Info-*.plist
+	rm -f config/Makefile config/coq_config.ml myocamlbuild_config.ml dev/ocamldebug-coq config/Info-*.plist
```
SkySkimmer: Do we still use this myocamlbuild thing by the way?

SkySkimmer (approved):

Makefile
ejgallego: > Do we still use this myocamlbuild thing by the way?

#8754

ejgallego:
Rebasing to solve a conflict, who would like to be the assignee?

ppedrot (commented):
Not skilled to review this PR...

ejgallego:
Maybe @SkySkimmer can be the assignee here?

SkySkimmer:
Still waiting for @gares review.

SkySkimmer:
@gares if you don't reply today I'll assume you don't care about this PR and I'll merge it.

PR #8756, opened 2018-10-17
[micromega] Build csdpcert

ejgallego:


Zimmi48 (approved):
This PR does indeed add csdpcert to the list of executables that are built and installed by Dune (cf. https://gitlab.com/coq/coq/-/jobs/109339931/artifacts/browse/_build/install/default/bin/). It strikes me that this was (again) an executable that was not built by default by the make-based target (cf. https://gitlab.com/coq/coq/-/jobs/109339927/artifacts/browse/_install_ci/bin/). Was it part of another target?
As far as I can tell, csdpcert is not mentioned in the documentation of micromega (cf. https://coq.inria.fr/refman/addendum/micromega.html), but this is not the case of votour either, BTW.
So what is this executable about?
I think I'll let micromega maintainers take care of merging this PR.

ejgallego:
Indeed @Zimmi48 how to handle this executable is an open question, we have the cache etc... also the way to locate it is wrong in Dune. But we will solve these issues in the `zarith` PR I think.

Zimmi48:
Not sure I understood how it was done with make, but never mind.

ejgallego:
> Not sure I understood how it was done with make, but never mind.

I think with make the binary was installed in `plugins/micromega`, so it was kind of a private one. Note that we could do the same with dune if we want. Just we don't do a public name, and then we add the rule:
```
(install
  (section lib)
  (files (csdpcert.exe as plugins/micromega/csdpcert%{exe}))
```

fajb:
To give a bit of context, csdpcert was kept separate for IP/License reasons because of sos_lib which is code originating from HOL Light.

ejgallego:
Thanks @fajb . I think a few devs [including me] are a bit confused about how it is supposed to be used, how the caches work, etc... There seems to be no documentation.

Would you mind adding a brief `README.md` to `plugins/micromega` explaining these topics for developers? [as opposed to what is in the manual]

Thanks.

fajb:
Here is some information. If you find it worth, feel free to include them in a readme.

## About 'csdpcert'

The micromega plugin is not stand-alone but is using a proof driver
(the binary 'csdpcert') which encapsulates the HOL Light code which is
solving sum of square problems using the semi-definite programming
solver 'csdp'.  When the 'psatz' tactic is run, the plugin spawns
'csdpcert' and communicate through pipes.

## About proof caches

In order to avoid re-doing over and over costly proof searches, the
provers are using memoisation. In order for the memoisation to be
reused across coq runs, the memo-tables are persistent and stored on
disk. The code managing caches is located in 'persistent_cache.ml'.
Currently, there is the following binding between tactics and caches:
- lia -> .lia.cache
- nia -> .nia.cache
- psatz -> .csdpcert.cache

As a last note, caches are directly using Ocaml Marshalling.

PR #8758, opened 2018-10-18
[api] Qualify access to `Nametab`

ejgallego:
In general, `Nametab` is not a module you want to open globally as it
exposes very generic identifiers such as `push` or `global`.

Thus, we remove all global opens and qualify `Nametab` access. The
patch is small and confirms the hypothesis that `Nametab` access
happens in few places thus it doesn't need a global open.

It is also very convenient to be able to use `grep` to see accesses to
the namespace table.

### DIFF
--- a/engine/namegen.ml
+++ b/engine/namegen.ml
@@ -21,7 +21,6 @@ open Constr
 open Environ
 open EConstr
 open Vars
-open Nametab
 open Nameops
 open Libnames
 open Globnames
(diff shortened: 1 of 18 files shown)


ppedrot (commented):
I don't really see the interest but I am not against this PR.

mattam82 (approved):
I'm fine with the qualification too.

ejgallego:
@herbelin would you like to be the assignee for this one?

herbelin:
No particular opinion, but I can be the assignee.

ejgallego:
Thanks @herbelin

herbelin:
Shall merge tomorrow (Friday) late afternoon, then.

herbelin (commented):
OK

PR #8761, opened 2018-10-18
[ci] Pin CI_REF to plugin_tutorial to use not yet merged commit.

Zimmi48:
This fixes the CI until this commit is merged into master.

**Kind:** infrastructure.

Following https://github.com/coq/coq/pull/8365#issuecomment-430963099.

### DIFF
--- a/dev/ci/ci-basic-overlay.sh
+++ b/dev/ci/ci-basic-overlay.sh
@@ -238,7 +238,7 @@
 ########################################################################
 # plugin_tutorial
 ########################################################################
-: "${plugin_tutorial_CI_REF:=master}"
+: "${plugin_tutorial_CI_REF:=14b2976cdf67db788b79d9421ce1e89bd15c7313}"
 : "${plugin_tutorial_CI_GITURL:=https://github.com/ybertot/plugin_tutorials}"
 : "${plugin_tutorial_CI_ARCHIVEURL:=${plugin_tutorial_CI_GITURL}/archive}"
 


ejgallego (approved):
Thanks a lot @Zimmi48 !

Zimmi48:
This can work thanks to https://github.com/ybertot/plugin_tutorials/archive/14b2976cdf67db788b79d9421ce1e89bd15c7313.tar.gz existing even before this commit (which is part of a PR) reaches master. Note that this could be a way to do overlays differently when merging changes with non-backward compatible overlays.

ejgallego:
You mean that we should require the pull request of the overlay to be submitted before the overlay?

IMO that could be fine, but not in all cases, as submitting the PR should happen when we are sure a change works.

I like more the convention that the overlay should be found automatically using the name of the PR submitter plus the name of PR branch.

Zimmi48:
We could still use the normal overlay system to test first, and then, once the PR is ready to merge, if the overlay is not backward compatible, submit the PR, change the CI_REF in `ci-basic-overlays.sh` then merge.

Said otherwise, we could keep the current overlay system for easy testing of a PR while it's being developed but it could be required to drop all overlays before merging (and update the CI_REF instead) in order to ensure that merging won't break CI on master.

SkySkimmer:
>This can work thanks to https://github.com/ybertot/plugin_tutorials/archive/14b2976cdf67db788b79d9421ce1e89bd15c7313.tar.gz existing even before this commit (which is part of a PR) reaches master.

Meaning it's broken in git clone mode (local builds)...

Zimmi48:
>Meaning it's broken in git clone mode (local builds)...

Correct. But then, even locally we can use the tarball mode during the (hopefully short) period of time until this commit is actually merged. That's a good use case for supporting `FORCE_GIT=0` after all! Cf. https://github.com/coq/coq/pull/8348#issuecomment-417598560

ejgallego:
@ybertot is around let's ping him to merge the PR.

PR #8762, opened 2018-10-18
[dune] [opam] Move to OPAM 2.0

ejgallego:
We need to update in Docker:

- dune to 1.4.0: as it honors `-p` on test stanzas
- dune-release to 1.0.1: support for OPAM 2.0

This makes `dune-release distrib` / `dune-release opam pkg` work.

TODO: we need to figure out what is going on with the
versioning. Should we do `dune subst` on `pinned`?

### DIFF
--- a/coq.opam
+++ b/coq.opam
@@ -1,18 +1,28 @@
-opam-version: "1.2"
+synopsis: "The Coq Proof Assistant"
+description: """
+Coq is a formal proof management system. It provides
+a formal language to write mathematical definitions, executable
+algorithms and theorems together with an environment for
+semi-interactive development of machine-checked proofs. Typical
+applications include the certification of properties of programming
+languages (e.g. the CompCert compiler certification project, or the
+Bedrock verified low-level programming library), the formalization of
+mathematics (e.g. the full formalization of the Feit-Thompson theorem
+or homotopy type theory) and teaching.
+"""
+opam-version: "2.0"
 maintainer: "The Coq development team <coqdev@inria.fr>"
 authors: "The Coq development team, INRIA, CNRS, and contributors."
 homepage: "https://coq.inria.fr/"
 bug-reports: "https://github.com/coq/coq/issues"
-dev-repo: "https://github.com/coq/coq.git"
+dev-repo: "git+https://github.com/coq/coq.git"
 license: "LGPL-2.1"
 
-available: [ ocaml-version >= "4.05.0" ]
-
 depends: [
-  "dune"      { build & >= "1.2.0" }
-  "ocamlfind" { build }
+  "ocaml"     {         >= "4.05.0" }
+  "dune"      { build & >= "1.4.0"  }
   "num"
-  "camlp5"    {         >= "7.03"  }
+  "camlp5"    {         >= "7.03"   }
 ]
 
 build-env: [
--- a/coqide-server.opam
+++ b/coqide-server.opam
@@ -1,15 +1,25 @@
-opam-version: "1.2"
+synopsis: "The Coq Proof Assistant"
+description: """
+Coq is a formal proof management system. It provides
+a formal language to write mathematical definitions, executable
+algorithms and theorems together with an environment for
+semi-interactive development of machine-checked proofs.
+
+This package provides the `coqidetop` language server, an
+implementation of Coq's [XML protocol](https://github.com/coq/coq/blob/master/dev/doc/xml-protocol.md)
+which allows clients, such as CoqIDE, to interact with Coq in a
+structured way.
+"""
+opam-version: "2.0"
 maintainer: "The Coq development team <coqdev@inria.fr>"
 authors: "The Coq development team, INRIA, CNRS, and contributors."
 homepage: "https://coq.inria.fr/"
 bug-reports: "https://github.com/coq/coq/issues"
-dev-repo: "https://github.com/coq/coq.git"
+dev-repo: "git+https://github.com/coq/coq.git"
 license: "LGPL-2.1"
 
-available: [ocaml-version >= "4.05.0"]
-
 depends: [
-  "dune"      { build & >= "1.2.0" }
+  "dune"      { build & >= "1.4.0" }
   "coq"
 ]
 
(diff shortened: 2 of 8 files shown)


coq.opam
```diff
@@ -1,18 +1,29 @@
...
 maintainer: "The Coq development team <coqdev@inria.fr>"
 authors: "The Coq development team, INRIA, CNRS, and contributors."
 homepage: "https://coq.inria.fr/"
 bug-reports: "https://github.com/coq/coq/issues"
-dev-repo: "https://github.com/coq/coq.git"
+dev-repo: "git+https://github.com/coq/coq.git"
 license: "LGPL-2.1"
-
-available: [ ocaml-version >= "4.05.0" ]
+doc: "https://coq.github.io/coq/api"
```
gares: this URL is a 404

ejgallego: This is required by the `topkg` / `dune-release` linter, it is not used in practice unless you do `dune-release publish doc`.

Dunno what to do about it.

ejgallego: I guess I will report a bug. If you modify it this happens:
```
$ dune-release distrib
Wrote archive _build/coq-V8.10+alpha-647-g5efe3fea6c.tbz
[ OK ] lint opam file coq.opam.
dune-release: [ERROR] Could not derive publication directory $PATH from opam
                      doc field value "https://coq.inria.fr/master/api";
                      expected the pattern
                      $SCHEME://$USER.github.io/$REPO/$PATH
[FAIL] lint _build/coq-V8.10+alpha-647-g5efe3fea6c failure: 1 errors.
```

Zimmi48: LOL that's a pretty aggressive expectation indeed.

coqide-server.opam
```diff
@@ -1,12 +1,23 @@
-opam-version: "1.2"
+synopsis: "The Coq Proof Assistant"
+description: """
+Coq is a formal proof management system. It provides
+a formal language to write mathematical definitions, executable
```
Zimmi48: What about putting a description of what coqide-server actually is?

coqide.opam
```diff
@@ -1,16 +1,26 @@
-opam-version: "1.2"
+synopsis: "The Coq Proof Assistant"
+description: """
+Coq is a formal proof management system. It provides
+a formal language to write mathematical definitions, executable
```
Zimmi48: Same remark for coqide.

coq.opam
gares: > Dunno what to do about it.

I don't know what dune-release is. But obliging people to put a dummy URL is silly.
Unless there is something like `--warn=-XX`, you should report a feature whish upstream.

ejgallego: > Unless there is something like `--warn=-XX`, you should report a feature whish upstream.

We can disable the linter but indeed it should be improved upstream. Dune release is the dune-aware for of `topkg`.

coqide-server.opam
ejgallego: Indeed, anyone does volunteer to write a version? [Current OPAM packages have no description]

coq.opam
ejgallego: Is not so silly when you consider the `topkg` release workflow; but indeed I do agree it has to be tweaked.

ejgallego: Anyways this should have minimal impact for Coq as of today as we are not going to use the `publish` command for now.

Zimmi48: Disabling the linter seems like the thing to do until the bug is fixed, yes.

SkySkimmer (changes requested):
Disable whatever linter is asking for a dummy URL.

ejgallego:
> Disable whatever linter is asking for a dummy URL.

Disable from where? The linter is only called if you try to upload packages, not by anything in this PR.

I'd suggest we discuss the API URL, indeed what the linter proposes is not so bad I think, does the current one have proper versioning?

SkySkimmer:
The point is that I don't want to have the dummy URL, and it sounds like there is a way to avoid it.

ejgallego:
> The point is that I don't want to have the dummy URL, and it sounds like there is a way to avoid it.

Ok, URL removed, but that will complicate the life of those willing to release packages.

Also, followed Théo advice and added some better descriptions for the packages.

coqide-server.opam
```diff
@@ -1,13 +1,24 @@
...
 homepage: "https://coq.inria.fr/"
 bug-reports: "https://github.com/coq/coq/issues"
-dev-repo: "https://github.com/coq/coq.git"
+dev-repo: "git+https://github.com/coq/coq.git"
 license: "LGPL-2.1"
 
-available: [ocaml-version >= "4.05.0"]
-
 depends: [
   "dune"      { build & >= "1.2.0" }
```
SkySkimmer: Are the coqide packages really still on dune 1.2?

ejgallego: Indeed let's raise them.

SkySkimmer (approved):

coqide-server.opam
SkySkimmer: I guess the CI pkg:opam job tests the opam file but without guaranteeing that the minimum versions are correct.

ejgallego: Indeed versions in OPAM packages are a subtle art. At least that particular one will be taken care when Dune is able to generate .opam files.

coqide-server.opam
```diff
@@ -1,15 +1,26 @@
...
+synopsis: "The Coq Proof Assistant"
+description: """
+Coq is a formal proof management system. It provides
+a formal language to write mathematical definitions, executable
+algorithms and theorems together with an environment for
+semi-interactive development of machine-checked proofs.
+
+This package provides the `coqidetop` language server, an
+implementation of Coq's (XML
+protocol)[https://github.com/coq/coq/blob/master/dev/doc/xml-protocol.md]
```
Zimmi48: If you meant to use Markdown's standard for link you have it backwards (should be `[ ]( )`). But what about not using Markdown? I'm not sure it's supported here.

Zimmi48 (approved):

ejgallego:
Merci !

Issue #7164, opened 2018-04-04
Identifier Completion API


ejgallego:
A key step in building a useful language server for Coq is to provide "code completion", which is usually a function of the form `completion_candidates : string -> string list` [other approaches are often possible and desirable].

#7156 reminded me to open this issue. As of today, Coq's particular name resolution design makes non-trivial to implement this feature if I am correct. [SerAPI](ejgallego/coq-serapi) doesn't even try to provide completion, and [company-coq](cpitclaudel/company-coq) uses a quite heavy-handed approach (a search in the global namespace).

As of today, thanks to #608 some internal bits of the namespace API are exposed, but I am not sure we would like to mess with that. What would a sensible roadmap to expose a convenient completion API? Some other points to consider:
- we may want to allow efficient prefix-search,
- completion may happen in different contexts, and involve also goal/proofs variables,

herbelin:
I don't know what would be a sensible roadmap wrt completion, but I'm rather happy with the "search in the global namespace" in company-coq and CoqIDE.

I'm otherwise missing being able to write things like `Def^T` to mean `Definition`, and so on.

Getting used to dynamic spell-checking with smartphones, I wonder how difficult would it be to have fixes for standard kinds of mispelling (e.g. `mathc` instead of `match`) automatically proposed?

Completion could indeed also involve goal/proofs variables.

ejgallego:
> I'm rather happy with the "search in the global namespace" in company-coq and CoqIDE.

Well, the only problem is that such global search takes several seconds [or even minutes sometimes], so it is not a very snappy interface to use for interactive typing. Not to say of the battery drain this incurs in laptops. That won't work at all for protocols such as the VSCode one that pings the API quite often.

One of my students in fact had to disable company Coq due to the large battery drain due to this.

So IMHO we want to provide prefix-efficient access to the identifiers somehow.

ejgallego:
I am also ccing @cpitclaudel , as the `company-coq` maintainer he may want to comment on the design of this.

cpitclaudel:
Why do you think searching in the global namespace is slow? And how did you measure the battery drain?
What's slow is getting the list of all identifiers, but then completion itself is rather fast.

ejgallego:
> Why do you think searching in the global namespace is slow?

[...]

PR #8766, opened 2018-10-18
[nametab] [api] Provide basic support for efficient completion.

ejgallego:
We provide a function `completion_candidates id` that will return a
list of candidates global references that have id as prefix.

This should be reasonably efficient, however UI still need to call
`shortest_qualid_of_global` which is a bit heavy. How to improve this
in the future is an open question.

cc: #7164

This is safe an useful enough for users that could be included in 8.9 if so we wish.

### DIFF
--- a/clib/cMap.ml
+++ b/clib/cMap.ml
@@ -35,6 +35,7 @@ sig
   val fold_left : (key -> 'a -> 'b -> 'b) -> 'a t -> 'b -> 'b
   val fold_right : (key -> 'a -> 'b -> 'b) -> 'a t -> 'b -> 'b
   val height : 'a t -> int
+  val filter_range : (key -> int) -> 'a t -> 'a t
   module Smart :
   sig
     val map : ('a -> 'a) -> 'a t -> 'a t
@@ -62,6 +63,7 @@ sig
   val fold_left : (M.t -> 'a -> 'b -> 'b) -> 'a map -> 'b -> 'b
   val fold_right : (M.t -> 'a -> 'b -> 'b) -> 'a map -> 'b -> 'b
   val height : 'a map -> int
+  val filter_range : (M.t -> int) -> 'a map -> 'a map
   module Smart :
   sig
     val map : ('a -> 'a) -> 'a map -> 'a map
@@ -85,8 +87,11 @@ struct
       if this happens, we can still implement a less clever version of [domain].
   *)
 
-  type 'a map = 'a Map.Make(M).t
-  type set = Set.Make(M).t
+  module F = Map.Make(M)
+  type 'a map = 'a F.t
+
+  module S = Set.Make(M)
+  type set = S.t
 
   type 'a _map =
   | MEmpty
@@ -163,6 +168,23 @@ struct
   let height s = match map_prj s with
   | MEmpty -> 0
   | MNode (_, _, _, _, h) -> h
+
+  (* Filter based on a range *)
+  let filter_range in_range m =
+    let rec aux m = function
+      | MEmpty -> m
+      | MNode (l, k, v, r, _) ->
+        let vr = in_range k in
+        (* the range is below the current value *)
+        if vr < 0 then aux m (map_prj l)
+        (* the range is above the current value *)
+        else if vr > 0 then aux m (map_prj r)
+        (* The current value is in the range *)
+        else
+          let m = aux m (map_prj l) in
+          let m = aux m (map_prj r) in
+          F.add k v m
+    in aux F.empty (map_prj m)
 
   module Smart =
   struct
--- a/clib/cMap.mli
+++ b/clib/cMap.mli
@@ -60,6 +60,12 @@ sig
   val height : 'a t -> int
   (** An indication of the logarithmic size of a map *)
 
+  val filter_range : (key -> int) -> 'a t -> 'a t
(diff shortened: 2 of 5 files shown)


ejgallego:
@herbelin what are your thoughts about this feature? Would you be the right assignee?

ppedrot:
How does this compare to https://github.com/coq/coq/blob/master/vernac/search.ml#L304?

ppedrot:
How does this compare to https://github.com/coq/coq/blob/master/vernac/search.ml#L304?

ppedrot:
How does this compare to https://github.com/coq/coq/blob/master/vernac/search.ml#L304?

ppedrot:
How does this compare to https://github.com/coq/coq/blob/master/vernac/search.ml#L304?

ppedrot:
How does this compare to https://github.com/coq/coq/blob/master/vernac/search.ml#L304?

ppedrot:
How does this compare to https://github.com/coq/coq/blob/master/vernac/search.ml#L304?

JasonGross:
Can this be exposed via a vernacular command for use in @cpitclaudel 's company-coq?

ejgallego:
> Can this be exposed via a vernacular command for use in @cpitclaudel 's company-coq?

Yes, it will.

This is only one small, but essential part of what is needed for a reasonable completion API. For example this doesn't deal with hypothesis and some other stuff.

Once, this gets in, I will first provide a prototype API in `coq-serapi`, once I am happy with it I will extract the useful part to a plugin.

ejgallego:
> How does this compare to https://github.com/coq/coq/blob/master/vernac/search.ml#L304?

@ppedrot they have so radically different behaviors that are difficult to compare.

`generic_seach` uses `Declaremods.iter_all_segments` which performs a brute-force iteration over _all_ the objects in `Lib`, then inspecting the tag and building a list of objects. This is `O(n)` in the number of objects in scope which for people working in large environments often mean a large delay (> 10 seconds) whenever the UI wants to refresh the list of identifiers. Filtering doesn't also account for open scopes, tho printing does as it tries to shorten the id up. This makes `Search` unsuitable for interactive completion use.

This PR indeed provides efficient matching over `Nametab` prefixes, it is on a different complexity class as it has efficient prefix matching, and it better suites interactive use. It will also allows to properly complete vs qualified identifiers.

ejgallego:
Does anyone volunteer to shepherd this?

@herbelin no opinion then?

ppedrot:
> This PR indeed provides efficient matching over Nametab prefixes, it is on a different complexity class

This is factually false, as this algorithm is still O(n) in the worst case. That said, I agree with the fact your algorithm is faster in practice.

clib/cMap.ml
```diff
@@ -164,6 +169,23 @@ struct
   | MEmpty -> 0
   | MNode (_, _, _, _, h) -> h
 
+  (* Filter based on a range *)
```
ppedrot: This function looks like it is violating the invariants of the binary map structure, as it can generate unbalanced trees IIUC (can it?). Couldn't use the map API instead?

ejgallego:
> This is factually false, as this algorithm is still O(n) in the worst case. That said, I agree with the fact your algorithm is faster in practice.

Well, it depends how you size the input for prefixes, but I see your point. It is extremely faster in practice, like with mathcomp loaded 10 seconds vs a few milliseconds for a prefix of size 3.

That being said, we will have to maintain a more complex trie structure if we want to provide good completion; prefixes are not enough. This PR does just add some API which is usable so we can begin to play with.

clib/cMap.ml
ejgallego: Can it? Do subtrees of a tree violate some invariant? Note the use of `F.add`.

I'm afraid it cannot use the tree API and still be fast, TTBOMK.

ppedrot: I don't see in you code how the call to filter_range is different to split, and furthermore the split function is going to be implemented by the OCaml stdlib in a similar way as filter_range... Anyways, if nobody complains much more I'll merge this PR soon.

ppedrot (approved):
Not completely happy about the shape of this PR, but this is for a greater good.

ejgallego:
Thanks @ppedrot . 

Now I'd like to request to @silene consideration for 8.9, let me summarize my rationale:

- **Summary**: this PR add a fast method for lookup of the main internalization symbol table.
- **Why is important to have this PR in 8.9**: current solutions for Coq completion use `Search`; unfortunately search takes a few seconds in large developments thus it is not suitable for real-time completion. The workaround is to cache the `Search`, but this is in general complex and annoying sometimes. [i.e.: will drain the battery of laptop users] 

  Having this PR in 8.9 would allow IDE users and toplevels to experiment with real-time completion strategies. IMHO this is a quite prioritary area.
- **What are the risks**: The main risk is that the proposed API is surely far from what we will like to have in the future. Likely, we will like to add hooks for a multi-trie data structure or even a text-based database that can run in a different thread. However, we can only know by starting to experiment.

Zimmi48:
Nope, changing the milestone *after* the PR has been merged is not yet supported by @coqbot (although I plan to add support for this). You need to both update the milestone and add the PR to the Request Inclusion column of the Coq 8.9. This will ensure that this gets Guillaume's attention.

ejgallego:
Ok, done, Thanks for the tips @Zimmi48

PR #8768, opened 2018-10-18
[stm] Add hooks for document actions.

ejgallego:
This arose after a question by Talia Ringer on how to log
user-interaction with a Coq document.

The hooks would allow a plugin to receive events about user data.

This is experimental and will need some tweaks to be useful for sure,
in particular w.r.t. errors.

[Note: this is safe enough as to be included in 8.9]


gares (changes requested):
The api already provides many hooks, why these 3 are packed in a record?

The exec hook is wrong: 1) observe is undocumented 2) there is already a hook that is called when a sentence is executed

SkySkimmer:
Why's this using Hook rather than pushing on a list? Do we care if more than one plugin is using these functions at a time?

ejgallego:
> The api already provides many hooks, why these 3 are packed in a record?

Why not? It makes more sense to me from the documentation point of view; they are also closely related as they are meant to capture user-behavior more than Coq behavior.

> The exec hook is wrong: 1) observe is undocumented 2) there is already a hook that is called when a sentence is executed

I am aware of it, however the current hook won't likely work in the final form: more information will need to be passed in order for the experiment to be useful, for example about failures, and I think the semantics will be a bit different. Hence the new hook.

> Why's this using Hook rather than pushing on a list? Do we care if more than one plugin is using these functions at a time?

I agree, I used `Hook` for stylistic reasons wrt file but I am not a fan of it. Feedback uses a hash-table, but I feel lazy to introduce yet another listener mechanism.

Note the API here is far from final IMHO, so adding the corresponding tag.

gares:
`observe` being documented, should not be used. Indeed CoqIDE never calls that API, hence your hook makes no sense for the purpose motivating this PR. Also your hook should be called observe, not exec.

IIRC you call it in serapi or jscoq because you opted for a pure lazy semantics and then you wanted to force some executions (the other option, requiring multi processes, has delayed tasks be picked up eventually).

In any case, the exec hook that exists is more abstract, since it is triggered when something does execute and is not tied to what caused the execution. A thunk may be run also because one calls Print, for example. exec triggers, observe not.

If you want to track what the UI does, add hooks to the UI.

About the record, why not putting all hooks in a record then? Either one (unpacked) or the other (packed).
I'm fine with both approaches. This PR is neither.

ejgallego:
> `observe` being UNdocumented, should not be used. Indeed CoqIDE never calls that API, hence your hook makes no sense for the purpose motivating this PR. Also your hook should be called observe, not exec.

`observe` is of course called by CoqIDE, by means of finish. Lets try to be a bit more factual before using words such "makes no sense", should we?

> Also your hook should be called observe, not exec.

It is `observe` that should be called `exec`, not the other way around.

> If you want to track what the UI does, add hooks to the UI.

Well, precisely what we are trying to avoid is to modify all Coq UIs in order to have a trace of the user. I think @tlringer will explain their use case shortly, so we should be able to refine this PR more.

Anyways I find this answer not very constructive. You dismiss our proposal without even understanding our use case it seems. Fortunately, times where this kind of development attitude was the norm in the OSS community are well-behind.

> About the record, why not putting all hooks in a record then? Either one (unpacked) or the other (packed).

What you are asking is like if I say "I should put all my functions in a module or else define one module per function". They are in a record because, as I pointed out, they do belong together to the functionality "trace editor interaction".

gares:
> `observe` is of course called by CoqIDE, by means of finish.

Indeed so why didn't you put the hook on finish, that the is the documented API for requesting some execution to happen?

> Anyways I find this answer not very constructive. You dismiss our proposal without even understanding our use case it seems. Fortunately, times where this kind of development attitude was the norm in the OSS community are well-behind.

I know, I'm sorry, the quality of our interactions are rotting at a speed I could not even imagine.

> What you are asking is like if I say "I should put all my functions in a module or else define one module per function". They are in a record because, as I pointed out, they do belong together to the functionality "trace editor interaction".

This kind of answers is the, silly, reason.

Note that also the other hooks are related, they notify when things happen, eg when a state is finally ready or can't be computed. Why this other group is not inside a record too? This PR introduces an asymmetry, this is what my comment is about.

tlringer:
What is the status on this?

If it helps to have a push from users: Two research tools, one at UCSD and one at UW, need a way to collect data from Coq users on the changes they make as they make those changes, rather than when they push (usually large) changes to Git. The data collected will help with development of both a proof patching tool and a machine learning tool to automate proofs. 

For me (proof patching tool, UW), Git commits contain changes to proofs which are too large. This means that it is impossible using just Git commits to observe how Coq users actually make changes to specifications and definitions and proofs as they are developing them. Without this data, it is difficult to make a tool that really helps Coq users.

For my collaborators (machine learning tool, UCSD), they need negative examples: changes to Coq proofs that don't actually work. People rarely commit these, so they essentially need the same kind of tooling I need.

Before using the data for our respective tools, we would like to deploy a plugin to collect this data to a number of proof engineers, then tag and analyze the data and see what it tells us about how proof engineers develop Coq proofs as specifications change. We expect this data would be useful to more than just our two projects, and would likely publish whatever analysis comes from it.

maximedenes:
Hi @tlringer,

You use case seems interesting and very legitimate. I'm sure @gares and @ejgallego will manage to reach an agreement as to how the hook should be called, and whether to put it in a record or not, which seems like a minor issue :)

@gares if you want more significant changes to this PR, what do you think would be the right way to instrument Coq to monitor user input?

gares:
If we forget about the way hooks are described in the API there is still the fact that the exec one should be attached to finish rather than observe, imo.

Btw I do understand why add and edit are useful to the two use cases above, but I fail to get why the exec one is. Tallia could you explain why you need to log that action?

tlringer:
@gares I don't fully understand the distinction between each of these hooks. But essentially, the information I would like is:

1. What changes do proof engineers make to specifications, definitions, and proofs without committing to Git?
2. Do those intermediate terms type-check or fail?

The second of these is more important for the machine learning tool, which is looking specifically for negative examples. But it is also interesting to me because it helps me observe the kinds of mistakes people make when adapting proofs to changes, which is valuable information for developing a tool that is trying to help people avoid having to make those mistakes to begin with.

gares:
This PR adds 3 hooks (`add`,`edit`,`exec`)
- `add` is called when a *parsed* sentence is appended to the document (what Coq knows about the text buffer the user sees)
- `edit` when part of the document is backtracked to a point
- `exec` when the UI demands Coq feedback, eg show me the current goal if any

In addition to that `Feedback` (a bus of messages, with an observer pattern) streams the following info https://github.com/coq/coq/blob/master/lib/feedback.mli#L30 of which
- `Error` on sentences that fail
- `Processed` on sentences that succeed

Unfortunately
- no present hook or feedback message signals when a sentence fails to parse.

To be more precise, UIs chop the text according to the terminator `.`, then send to Coq text, that is then parsed and `add`ed to the document. They may do this many times before requiring Coq to make any computations (sometimes execution may happen anyway because of our extensible parsing engine).
Then they typically ask for `exec`, report to the user and finally the user may `edit` to amend the text.

I imagine you also need to be notified about parsing errors.
I guess the last hook added bu this PR, `exec` is not relevant to you, correct me if I'm wrong.

Feel free to ask if my explanation is too obscure.

HazardousPeach:
Hey all,

I'm the researcher at UCSD that Talia mentioned. As far as I can tell, we probably don't need `exec` (I'm familiar with the `Observe` version from Serapi), but text that fails to parse would probably be useful (although this api would still have utility without it, since much of the more interesting failures happen after the parsing stage).

I'm a little confused about where the actual parsing happens. In the Serapi API that I'm more familiar with, `AddStmt` is called with any string, and Coq will attempt to parse the string, adding it as a statement if the parse is successful, or returning a `ParseError` (or `LexError`) object if it's not. But you're saying that in this API, the `add` in `stm/stm.ml` isn't called until after parsing? Is there an earlier place we can hook in to get the full string, or is the parsing somehow relegated to the UI code in coqide?

ejgallego:
Hey  Alex,

> As far as I can tell, we probably don't need `exec` (I'm familiar with the `Observe` version from Serapi),

You will need `exec` if you want to log precisely the moment the user press the "check" button. For PG this doesn't matter so much, but for CoqIDE may be relevant. Otherwise, listening to feedback may be OK.

> but text that fails to parse would probably be useful (although this api would still have utility without it, since much of the more interesting failures happen after the parsing stage).

You can hook parsing easily in `Stm.parse_sentence`.

ejgallego:
A further note @HazardousPeach @tlringer , IMHO the project feels too experimental (:)) as for this PR to work well as a development platform.

IMO, things would go smoother if we would fork for a while until things are running smoothly, then we would update this PR as to propose inclusion of the framework if you think that would be worth it.

WDYT?

If you dig this idea then I could provide a branch so that we could opam pin Coq and that would get an instrumented Coq including the plugin.

tlringer:
@ejgallego That sounds reasonable to me. Right now it seems like we basically want all of the hooks we can get, but we won't know for sure which ones we'll need access to until we play around a lot more and maybe deploy a beta version of the plugin to a few users.

I've added you as a collaborator to the plugin (which currently just outputs the state ID, but is setting hooks) so you can test it out with whatever Coq branch you cut if you'd like.

ejgallego:
Cool, thanks! I'll add the plugin to the branch and post and update here.

I don't see the plugin repos tho.

gares:
> the user press the "check" button.

Which button? If CoqIDE you have the ones with the wheels. This is does not call finish, but join.
CoqIDE calls finish after a sequence of add AFAIR, eg as the last action of a jump-to-point.

ejgallego:
> Which button?

I meant go to point, this calls `finish` which is basically `observe (get_cur_tip())`.

People using SerAPI do tend not to rely on implicit status so they call `observe sid` as to not to have to implement a synchronous event to call `finish`. This relies in using `add ~newtip` which is broken for certain side effects, but it is so convenient that usually simplifies IDE code by an order of magnitude.

[Note this schema was first proposed by Clément, but this is indeed what most protocols do use].

LSP Coq will also use a per sentence `Check`, that's a stronger requirement due to the full-document nature.

ejgallego:
In full-document models you have indeed no notion of "tip", [and most IDE protocols these day are full-document] so that's important to keep in mind.

tlringer:
@ejgallego It's here: http://github.com/uwplse/coq-change-analytics

ejgallego:
Great, thanks @tlringer ! 

In this case it may be more convenient start with a more coupled tree, what do you think of this?
https://github.com/ejgallego/coq/tree/coq-change-analytics

ejgallego:
I've tested the tree and indeed it works! However the hooks clearly need more work; for example add is not printing the correct thing, etc...

I'm gonna do a super-quick pass and I'll stop for today.

tlringer:
Does it make sense to move conversation about this elsewhere as we develop the prototype and determine what hooks make the most sense, and then check back on this later? Is there a procedure for that?

ejgallego:
> Does it make sense to move conversation about this elsewhere as we develop the prototype and determine what hooks make the most sense, and then check back on this later? Is there a procedure for that?

Sure, let's discuss over whatever medium is more convenient; we can revisit the PR later on.

Indeed, there are quite a few subtle things to fix as unfortunately the execution path is not as uniform as it should be, for example, the `Instance` vernacular is handled specially. Some commands may also be shown to be executed twice due to ghost nodes.

I guess the top-level output of the plugin should be in the lines of "diff [in textual form?] ok/error". Unfortunately again, modern IDE protocols pretty much work directly in this way, however in the case of Coq the plugin will have to do some aggregate as to process that.

ejgallego:
@tlringer @HazardousPeach after thinking a bit more about what you would ideally want for the kind of experiment you propose I think the current instrumentation is quite inconvenient.

After all, you don't care if something was added or not, but more about its actual execution, right?

I think a different instrumentation should be tried then, hooking in `stm_vernac_interp` + edit [which is really a cancellation of a part of the document]

That will still produce some noise in the case of parallel processing, as some sentences are scheduled to be checked twice, but that should not be too bad I think.

tlringer:
@ejgallego Getting back a result with parsing and execution errors sounds very useful. What type are you imagining `sentence_data` would have? 

Would it be possible to get back not only the current sentence/diff, but also the entire proof state at that point if we wanted it? I imagine that @HazardousPeach would want this for his negative examples. Similarly, most of my tooling works not with a syntactic diff over tactics but with a semantic diff over terms, so it would be useful for me to be able to inspect the old and new terms that correspond to changes in text. I imagine any automatic analysis we build over this data will also want access to proof state and to terms, and not just raw syntactic information.

tlringer:
We've determined that we do need exec after all, for the following reasons:

1. It tells us more information about the proof development process. It helps us, for example, to distinguish between rebuilding an entire project and simply stepping through a proof in an IDE.
2. If we send information to the server on every single hook, it ends up hanging for large execs, like building an entire project. Ideally, we want to log to a file on the other hooks, and only send that file to the server on exec.

ejgallego:
Hi @tlringer , I saw the issues thanks; I should be able to push the new hooks soon I hope; a couple of parts are indeed tricky due to having to map different codepaths to the same logical action, but IMHO it should be OK to use.

tlringer:
Hi, what is the status on this? We would like to release the beta version of our tool by the beginning of June. @gares, can we merge this change?

ejgallego:
Indeed, let's make some progress on this @tlringer ; I guess we should sort the conversation out and make a before-merge TODO list. For a start, I've just rebased the PR.

I've realized you were not a collaborator of the Coq organization @tlringer , that overlook should be fixed now and you can edit / amend the discussion, hide no more important things, etc...

ejgallego:
The first todo here is error handling, current "feedback" based system is unreliable in practice (see #5479 ) so we need to capture exceptions to the main commands.

Parsing errors should be captured this way without too much of a problem.

tlringer:
Can we get a summary of what's left on this, and what comments are no longer relevant? Specifically, are we addressing any of the feedback @gares left?

The hooks as-is are enough for our tool.

gares (changes requested):
The are already a few hooks in Stm.mli and I don't get why these 3 
  new ones are packaged (while the others are not, hence my request
  to unpackage the ones the PR is adding).

  Even if we decide to package "related hooks", here "add" and 
  "edit" are about editing the document  (adding/removing sentences), 
  while "exec" is rather about requesting the evaluation of a
  sentence, so they don't belong to the same package (that being 
  named document_edit_notifiers suggests it contains hooks for 
  document edition).
  
  It is about style and consistency, not a big deal, but it is 
  also a trivial to implement change. I hope it is not a problem.

tlringer:
@ejgallego, are you able to make only that change? It's the only one @gares thinks is necessary right now.

tlringer:
I can make the requested changes on a branch, but I'm not sure how to update the PR. Does anyone know the easiest way to do that?

gares:
You can surely open another PR, and we can merge that one and close this one. Or ask Emilio write writes on his clone.

tlringer:
I have write permission. I'm about to push a take of this on Emilio's branch.

I'm not sure what the other hooks are, so I'm not sure if my attempt at addressing your feedback is reasonable, but I tried to just follow the same exact style as the old hooks for the new hooks. If they should be further distinguished in some way, then let me know.

tlringer:
OK, the change is pushed to Emilio's branch. It implements the same hooks Emilio implemented, but tries to follow the same style as the existing hooks. Also, it's updated to work with the latest commits in master. 

Let me know if this is OK or if you'd prefer for these hooks to be distinguished more from the existing hooks.

gares (approved):

tlringer:
There's an ambiguous comment problem; I'll try to fix that. Not sure why it didn't break my local build.

tlringer:
I don't have permission to merge, can one of you do the merge?

gares:
Later today I will.

tlringer:
It's extremely minor, but I did realize just now I added my name to credits at the end rather than maintaining alphabetical order. It's 11:15 PM here and I'm browsing on my phone, but if you want you should change that to maintain the order before merging. If not, I'll move it next time I make a contribution.

gares:
It looks reasonable you fix it next time.

There are many minor issues in that file, the funniest is that INRIA is now written Inria and is not an acronym anymore... so if we are imprecise about our own employer name, I guess it is fine if your name is a few lines below ;-) Maybe also expand UW while you are at it, and looking forward to your next PR!

BTW, I've noticec @ppedrot is still at MPI according to that file... he may want to fix it

PR #8769, opened 2018-10-18
[library] Remove redundant re-addition of universe constraints.

ejgallego:
After some analysis this should be taken care of by
`Safe_typing.add_constant`

It was added in
https://github.com/coq/coq/commit/f7338257584ba69e7e815c7ef9ac0d24f0dec36c
, so maybe @gares can provide more context as to how is this stuff
supposed to work.

Note, I am not sure this is correct, I understand very little of this part of the code so this needs careful reviewing.

### DIFF
--- a/interp/declare.ml
+++ b/interp/declare.ml
@@ -60,14 +60,7 @@ let open_constant i ((sp,kn), obj) =
   if obj.cst_locl then ()
   else
     let con = Global.constant_of_delta_kn kn in
-    Nametab.push (Nametab.Exactly i) sp (ConstRef con);
-    match (Global.lookup_constant con).const_body with
-    | (Def _ | Undef _) -> ()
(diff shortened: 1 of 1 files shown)


ppedrot (approved):
Indeed this piece of code was definitely strange. I also believe it is correct to remove it.

gares (approved):

SkySkimmer (approved):
I also think it's correct. Also if we rely on this to avoid inconsistency we have serious problems.

herbelin:
> @herbelin are you the right assignee here?

Among the persons in the list of reviewers, I would rather say @ppedrot or @skyskimmer.

ejgallego:
@herbelin you are the codeowner of the file in question so that's why I asked first but indeed any of them can take of this one.

PR #8771, opened 2018-10-18
[core] Remove custom backtrace module in favor of OCaml's.

ejgallego:
We take advantage some newer primitives such as `reraise`.  This is
mainly an RFC and will need fine tuning, but it seems to provide
better backtraces and I do prefer the cleanup of the current API.

There is a catch, `reraise` won't work properly if another exception
is called in the exception handler itself, in this case we need to use
some extra protection.

I have tried to preserve places that mess with the `exninfo`, however
all that stuff is suspicious and I'm afraid that hack that could stop
being supported upstream at any time. IMHO we should migrate from it.

In particular, it is beneficial that `CErrors` doesn't not depend now
on `Exninfo`, as this used to give problems in the past.

Should fix some day #6446.


ejgallego:
I've also added some improvement on the support of the monad for backtraces, maybe it should go in a different PR but IMHO the results do seem quite better, see the commit message for an example.

ejgallego:
Note @maximedenes the last commit. IMHO worth it, bringing in half of `library` for a meager 20 lines seems like serious stuff, but review and do as you prefer.

maximedenes (approved):
I like the idea, but this should be reviewed by @ppedrot.

ppedrot:
Ah, sorry, forgot to review properly that one. First time I skimmed through it I found suspicious parts, so I deferred in-depth reviewing to a later time. I'll do this properly ASAP.

ejgallego:
Yup this needs more work for sure, consider the current PR as a proof of concept

ejgallego:
I am getting quite annoyed by the low-quality of our current backtraces, @ppedrot when are you gonna have time to work on this.

ppedrot (changes requested):
This is way too fragile, even compared to the current mess. There are places where you call reraise after calling arbitrary code, which means the backtraces will be completely wrong. I think that I'd prefer in a first step to get rid of the Coq-defined backtraces using the OCaml API that didn't exist at the time while keeping the explicit push after try-with blocks.

ejgallego:
It is a pity we didn't make a lot of progress on this; current backtraces are not very good :(

maximedenes:
@ejgallego do you plan to follow @ppedrot's suggestion?

ejgallego:
> @ejgallego do you plan to follow @ppedrot's suggestion?

I'd be happy to however this is quite tricky stuff so I need to understand the implications better.

ejgallego:
Also this patch adds better backtraces to the monad stuff, so indeed, it was more like a study; I guess that we will have to rewrite it in full.

ppedrot:
I propose to close this PR, whose rebase is going to be messy anyways, and ponder a bit about a proper cleanup. We'll open a fresh one when we come up with some good approximation of it.

PR #8773, opened 2018-10-19
[checker] Refactor by sharing code with the kernel

maximedenes:
For historical reasons, the checker was duplicating a lot of code of the
kernel. The main differences I found were bug fixes that had not been
backported.

With this patch, the checker uses the kernel as a library to serve the
same purpose as before: validation of a `.vo` file, re-typechecking all
definitions a posteriori.

We also rename some files from the checker so that they don't clash with
kernel files.

Based on #8684 [merged].

EDIT: now depends on #8842 [merged] and #8870 [merged].

Still TODO:
- [x] Fix the Dune build
- [x] Make sure we check the same things as before on modules (@ppedrot can you have a look?)
- [X] Disable VM (and native) for conversions triggered by the checker
- [X] Make sure the checker can't load plugins, as claimed in the manual


ejgallego:
Note @maximedenes the last commit. IMHO worth it, bringing in half of `library` for a meager 20 lines seems like serious stuff, but review and do as you prefer.

SkySkimmer:
This depends on #8684 IIUC.

SkySkimmer:
@ejgallego did you forget to remove the -I library from makefile in your commit or is it still somehow needed?

checker/checkTypes.ml
```diff
@@ -0,0 +1,36 @@
...
+  let rec check_p env pl params =
+    let open Context.Rel.Declaration in
+    match pl, params with
+        Some u::pl, LocalAssum (na,ty)::params ->
+          check_kind env ty u;
+          check_p (push_rel (LocalAssum (na,ty)) env) pl params
+      | None::pl,d::params -> check_p (push_rel d env) pl params
+      | [], _ -> ()
+      | _ -> failwith "check_poly: not the right number of params" in
+  check_p env pl (List.rev params)
```
SkySkimmer: Does this really deserve a file? It could live in kernel/typeops I think.

checker/mod_checking.ml
```diff
@@ -89,40 +102,41 @@ and check_structure_field env mp lab res = function
...
   | MEapply (f,mp) ->
       let sign = check_mexpr env f mp_mse res in
-      let farg_id, farg_b, fbody_b = destr_functor sign in
-      let mtb = module_type_of_module (Some mp) (lookup_module mp env) in
-      check_subtypes env mtb farg_b;
-      subst_signature (map_mbid farg_id mp) fbody_b
-  | MEwith _ -> error_with_module ()
+      let farg_id, farg_b, fbody_b = Modops.destr_functor sign in
+      let mtb = Modops.module_type_of_module (lookup_module mp env) in
+      let _ = check_subtypes env mtb farg_b in (* Should we stop inferring constraints? *)
```
SkySkimmer: Shouldn't we check that the constraints are satisfied instead of dropping them?

ppedrot:
Without presuming of the tradeoffs of this PR, I am a bit saddened by the disappearance of the `cic.mli` file that was a pretty good summary of the structures involved in the definition of the Coq kernel.

checker/safe_checking.ml
```diff
@@ -25,6 +25,18 @@ let pr_dirpath dp = str (DirPath.to_string dp)
 let genv = ref empty_env
 let get_env () = !genv
```
SkySkimmer: We should carry this global ref functionally at some point.

maximedenes: I'm taking care of this one, in fact almost the whole file can be refactored as well.

checker/mod_checking.ml
maximedenes: Right, what function should I call?

maximedenes: Although, isn't it already checked by the kernel at subtyping time? If not, when does the kernel check it?

ejgallego:
> @ejgallego did you forget to remove the -I library from makefile in your commit or is it still somehow needed?

I just don't use use `Makefile`. The current make system is anyways not well suited for the purpose of a containerized build, unless you specify all the rules by hand, that is.

ejgallego:
> I'm refactoring more with library, but we can evaluate again the tradeoff for the final patch.

Certainly I am convinced that bringing in `global/summary/nametab` etc... must be avoided.

checker/mod_checking.ml
SkySkimmer: The kernel adds the constraints to the environment. This part corresponds to some subcall of translate_module so the constraints are first added to the module_body that function produces, then when the module is added to the safe env the constraints are added.
When checking use Environ.check_constraints (returns bool).

maximedenes:
> Certainly I am convinced that bringing in `global/summary/nametab` etc... must be avoided.

Oh, not at all! In fact, it is a bug of the current patch that we are not using the global environment and instead have a local one that is badly kept in sync. Let me push the fixed patch so that you see.

ejgallego:
> Oh, not at all! In fact, it is a bug of the current patch that we are not using the global environment and instead have a local one that is badly kept in sync. Let me push the fixed patch so that you see.

well let me remove `global` from that list, but certainly `nametab` and `summary` must not be used.

maximedenes:
> but certainly `nametab` and `summary` must not be used.

Indeed, I don't think they are used. In fact, `coqchk` is like `coqc`, it is a toplevel that doesn't need interactive features, and what is special is that it takes as input a `.vo` which it validates.

maximedenes:
@ejgallego do you have a backup of your last commit? I'm going to have to force push the larger refactoring which removes it (but keeps the one for the Dune build, which may have to be adapted).

ejgallego:
> @ejgallego do you have a backup of your last commit? I'm going to have to force push the larger refactoring which removes it (but keeps the one for the Dune build, which may have to be adapted).

Umm why do I need a backup of it then?

IMHO you should not push until you make Dune satisfied, not especially with the current build rules you are using. Otherwise, the PR is useless as we cannot see what the new checker is using.

Zimmi48 (commented):
Reading again the checker's doc at https://coq.inria.fr/refman/practical-tools/coq-commands.html#compiled-libraries-checker-coqchk, I think it doesn't require any update after this patch. There is the word standalone but it is not meant in the sense that coqchk would be an independent implementation, but only a checker that does not load any plugin.

Zimmi48:
> Oh yes, I need to add code to ensure that, I forgot about it, thanks!

Wouldn't it be safer to do as Emilio suggested: do a kernel_lib from what is common to the checker and the kernel and use only this? This would avoid having to "disable" the VM and native compute, the module and section management, the plugins, etc. This would also avoid accidental violation of this invariant in future updates.

maximedenes:
> Wouldn't it be safer to do as Emilio suggested: do a kernel_lib from what is common to the checker and the kernel and use only this? This would avoid having to "disable" the VM and native compute, the module and section management, the plugins, the more efficient universe graph algorithm, etc. This would also avoid accidental violation of this invariant in future updates.

And how do I avoid the duplication of all the code from `library`, then?

SkySkimmer:
The checker should handle the environment functionally, not with library-style global refs.

maximedenes:
> The checker should handle the environment functionally, not with library-style global refs.

Sorry, but what is different between coqtop and coqchk, for this matter?

SkySkimmer:
coqtop has a ton of stuff on top of the kernel
coqchk doesn't have anything on top of the kernel

Gaëtan Gilbert

maximedenes:
> coqchk doesn't have anything on top of the kernel

that's not true, look at its library loading code...

SkySkimmer:
I may have exaggerated some but is it really that bad? If so I'm sorry 
for assuming but also a bit scared at the checker complexity that I 
wasn't aware of.

Gaëtan Gilbert

maximedenes:
> I may have exaggerated some but is it really that bad? If so I'm sorry for assuming but also a bit scared at the checker complexity that I wasn't aware of.

Well, it is not horrible, but it has dependency checking, checksum validation, logical paths mapping (I haven't refactored that part yet), etc.

Also, a lot of the added value code of the checker lives above the kernel, and validates the shape of `.vo` file. If you read the description of the checker in the manual, you'll realize it can be seen as an alternative toplevel which takes `vo` files as input and validates them.

I strongly believe that code duplication should be removed as much as possible. But I understand it would be better to not bring too much of the codebase in the checker at the same time. I think @ejgallego 's remark go in the direction of doing these code imports in a fine-grained way, so that we get the best of both worlds, maybe at the cost of reorganizing a bit the codebase.

The current patch is suboptimal in that respect, and I'll try to improve it once I fix the last remaining (universe-related) bugs.

gares (approved):
makefile related changes seem sensible to me.

maybe it is worth renaming the checker so that it is clear that its role is not the one it used to be (whatever that was really doing or perceived to be doing) but rather to perform a "vo roundtrip validation".

maximedenes:
> maybe it is worth renaming the checker so that it is clear that its role is not the one it used to be (whatever that was really doing or perceived to be doing) but rather to perform a "vo roundtrip validation".

I'm not sure I understand this comment. As @Zimmi48 mentioned, the role of the checker was rather precisely described in the manual and is still the same as before. I don't really know what else was perceived or where it was coming from, in particular I doubt the name played a big role there.

maximedenes:
> I'm not sure I understand this comment. As @Zimmi48 mentioned, the role of the checker was rather precisely described in the manual and is still the same as before. I don't really know what else was perceived or where it was coming from, in particular I doubt the name played a big role there.

But of course, I'm not opposed to a change of name for the checker, but it is fully orthogonal to this PR which does not change its role at all.

gares:
Well, the other day you told me that some people were thinking it was an independent checker, as in de bruijn criterium. If you think there is no confusion after all, then I'm fine with the name.

maximedenes:
> Well, the other day you told me that some people were thinking it was an independent checker, as in de bruijn criterium. If you think there is no confusion after all, then I'm fine with the name.

Yeah, I've heard that before, but I'm not sure it comes from the name. Honestly, I don't really know where the confusion comes from. Maybe simply the fact that Bruno was involved in both Coq-in-Coq and the checker.

ejgallego:
After another look it should be possible to have a checker with minimal duplication.

Proper organization could require virtual libraries (thus not possible today) or defining new interfaces, but on the other the amount of API we are talking about is not a lot.

dev/ci/docker/bionic_coq/Dockerfile
```diff
@@ -37,7 +37,7 @@ ENV COMPILER="4.05.0"
 # Common OPAM packages.
 # `num` does not have a version number as the right version to install varies
 # with the compiler version.
-ENV BASE_OPAM="num ocamlfind.1.8.0 dune.1.2.1 ounit.2.0.8 odoc.1.3.0" \
+ENV BASE_OPAM="num ocamlfind.1.8.0 dune.1.4.0 ounit.2.0.8 odoc.1.3.0" \
```
ejgallego: You should not need to bump the docker image anymore, 1.4.0 is in `master` now.

maximedenes:
This is ready for last reviews / merge. @ppedrot can you be the assignee? Thanks!

maximedenes:
Oh no, there is still two  FIXMEs, will address them now, sorry.

ppedrot (approved):
I know I'll regret this, but let's believe that this PR is OK (modulo FIXMEs).

checker/dune
```diff
@@ -1,7 +1,31 @@
...
+  %{project_root}/kernel/{cClosure,cPrimitives,csymtable,vconv,vm,uint31,cemitcodes,vmvalues,cbytecodes,cinstr,retroknowledge,copcodes}.ml{,i})
+
+(copy_files
+  %{project_root}/kernel/{cbytegen,clambda,nativeinstr,nativevalues,nativeconv,nativecode,nativelib,nativelibrary,nativelambda}.ml{,i})
+
+(copy_files
+  %{project_root}/kernel/{subtyping,term_typing,safe_typing,entries,cooking}.ml{,i})
+
+(copy_files
+  %{project_root}/library/{decl_kinds,declaremods,summary,libnames,nametab,loadpath,globnames,libobject,lib,library}.ml{,i})
```
ejgallego: As discussed in the comment and in the branch, I think this should be considerably thinned.

ejgallego (changes requested):

maximedenes:
@ejgallego I addressed your comments, the checker now depends only on kernel files. We may want to refactor the library code similarly, but I'll leave that for another PR.

checker/dune
```diff
@@ -1,7 +1,25 @@
...
+(copy_files
+  %{project_root}/kernel/{cClosure,cPrimitives,csymtable,vconv,vm,uint31,cemitcodes,vmvalues,cbytecodes,cinstr,retroknowledge,copcodes}.ml{,i})
+
+(copy_files
+  %{project_root}/kernel/{cbytegen,clambda,nativeinstr,nativevalues,nativeconv,nativecode,nativelib,nativelibrary,nativelambda}.ml{,i})
+
+(copy_files
+  %{project_root}/kernel/{subtyping,term_typing,safe_typing,entries,cooking}.ml{,i})
+
+; VM stuff
```
ejgallego: If you are disabling vm compute do you still to link this?

maximedenes: Good question, I thought we could provide a flag that enables VM / native for the checker, for cases when checking takes too much time otherwise. But I didn't expose such a flag yet (the machinery is there, though).

Do you think I should remove the runtime?

ejgallego (approved):
LGTM

checker/dune
ejgallego: I guess it is fine to leave it as is for now. Would you refine the VM/Native hooks indeed you could remove from the build a lot of files, which could be important for certification/review purposes, but that can happen later I guess.

ppedrot:
Last round of complains / comments.

Makefile.checker
```diff
@@ -34,22 +34,8 @@ CHECKMLLIBFILE := checker/.mllibfiles
...
-checker/names.mli: kernel/names.mli
-	cp -a $< $@
-	sed -i.bak '1i(* AUTOGENERATED FILE: DO NOT EDIT *)\n\n\n\n\n\n\n\n' $@ && rm $@.bak
-checker/names.ml: kernel/names.ml
-	cp -a $< $@
-	sed -i.bak '1i(* AUTOGENERATED FILE: DO NOT EDIT *)\n\n\n\n\n\n\n\n' $@ && rm $@.bak
-
 ifeq ($(BEST),opt)
-$(CHICKEN): config/config.cmxa checker/check.cmxa checker/coqchk.mli checker/coqchk.ml
+$(CHICKEN): $(CORECMA:.cma=.cmxa) checker/check.cmxa checker/coqchk.mli checker/coqchk.ml
```
SkySkimmer: Why are we linking all CORECMA here?

ejgallego: Indeed, this is a leftover of the original PR that used many files around the codebase. This should be able to be changed to `kernel.cma` now. @ppedrot wait until this is fixed please.

maximedenes: Fixed.

SkySkimmer:
It's somehow strange to me that in make we link kernel.cma but in dune we copy source files around. Why's it done this way?

maximedenes:
@ejgallego wanted to do it that way in Dune. The Makefile stuff can be seen as legacy anyway, although it also does not seem bad to me to link with libraries as opposed to individual files.

SkySkimmer (changes requested):
Dependencies in make are wrong, if I `make bin/coqchk` from clean state I get

```
OCAMLOPT -o bin/coqchk
/usr/bin/ld: cannot find -lcoqrun
collect2: error: ld returned 1 exit status
File "caml_startup", line 1:
Error: Error during linking
```
If I make coqtop first it magically works.

SkySkimmer:
I guess it needs `$(LIBCOQRUN)`

maximedenes:
> Dependencies in make are wrong, if I `make bin/coqchk` from clean state I get

Fixed, thanks for noticing!

Makefile.checker
```diff
@@ -22,7 +22,7 @@ CHICKEN:=bin/coqchk$(EXE)
 
 # The sources
 
-CHKLIBS:= -I config -I clib -I lib -I checker
+CHKLIBS:= -I config -I clib -I lib -I checker -I kernel -I kernel/byterun -I library -I engine
```
SkySkimmer: There should be no need to `-I library -I engine` now.

SkySkimmer (changes requested):

Makefile.checker
maximedenes: Right, it was a leftover, now removed.

SkySkimmer (approved):

checker/dune
```diff
@@ -1,7 +1,25 @@
-(rule (copy %{project_root}/kernel/names.ml names.ml))
-(rule (copy %{project_root}/kernel/names.mli names.mli))
-(rule (copy %{project_root}/kernel/esubst.ml esubst.ml))
-(rule (copy %{project_root}/kernel/esubst.mli esubst.mli))
+(copy_files
```
Zimmi48: Isn't `copy_files#` usually preferable? Cf. https://dune.readthedocs.io/en/latest/dune-files.html#copy-files

ejgallego: Good question, this is a bit of a special case but indeed `copy_files#` seems to be better.

maximedenes: I changed to `copy_files#`, thanks.

ppedrot:
*En avant Guingamp.*

PR #8776, opened 2018-10-19
Replace non-idiomatic "dead-alleys" with idiomatic "dead-ends"

ghost:
**Kind:** documentation.

Minor readability enhancement.

### DIFF
--- a/doc/sphinx/credits.rst
+++ b/doc/sphinx/credits.rst
@@ -120,7 +120,7 @@ G. Dowek, allowed hierarchical developments of mathematical theories.
 This high-level language was called the *Mathematical Vernacular*.
 Furthermore, an interactive *Theorem Prover* permitted the incremental
 construction of proof trees in a top-down manner, subgoaling recursively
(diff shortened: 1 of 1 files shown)


herbelin:
Well, I understand that it can make reading a bit strange, but the choice of "dead-alleys" was intended by the author, I think.

sampablokuper:
@herbelin wrote:
> I thought it would have been a pun but, browsing the web, it is not clear that it can be understood humoristically in English, more than as a confusion between dead end and blind alleys. (I'm not native English speaker.)

Given the context (a dry, tersely-worded passage in a technical manual), it reads as a mistranslation - a genuine confusion between  "dead end" and "blind alley" - rather than as a humorous pun. Either of those two idiomatic phrases would make more sense here than the current wording.

Zimmi48 (approved):

Zimmi48:
I am not too fond of changing this part of the manual (because there are dates and author names) but I think that for such minor rewording it is OK.

PR #8777, opened 2018-10-19
Move side-effects into Safe_typing

ppedrot:
Due to a cyclic dependency in data types, side-effects were inside Term_typing while semantically part of Safe_typing. This required exporting many internal functions in Term_typing that were hidden later on. We break this cycle by providing a higher-order API to Term_typing encapsulating the side-effect type and move it to the place it belongs. This cleans up the API a lot.


ejgallego (approved):
I was bothered by this in the past, cool to see it fixed!

[Didn't look at the type checking itself].

PR #7229, opened 2018-04-12
Deprecate implicit tactic solving.

ppedrot:
That's about it. Does it deserve a CHANGES entry or is that going to be automatic?

PR #8779, opened 2018-10-19
Remove the implicit tactic feature following #7229.

ppedrot:
We also remove the documentation and adapt the only place that still relies on the pretyping hook feature (guess what, it is Program).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -67,6 +67,8 @@ Vernacular commands
 - Binders for an `Instance` now act more like binders for a `Theorem`.
   Names may not be repeated, and may not overlap with section variable names.
 
+- Removed the deprecated `Implicit Tactic` family of commands.
+
 Tools
 
 - The `-native-compiler` flag of `coqc` and `coqtop` now takes an argument which can have three values:
--- a/dev/ci/user-overlays/08779-ppedrot-rm-implicit-tactic.sh
+++ b/dev/ci/user-overlays/08779-ppedrot-rm-implicit-tactic.sh
@@ -1,0 +1,8 @@
+_OVERLAY_BRANCH=rm-implicit-tactic
+
+if [ "$CI_PULL_REQUEST" = "8779" ] || [ "$CI_BRANCH" = "$_OVERLAY_BRANCH" ]; then
+
+    ltac2_CI_REF="$_OVERLAY_BRANCH"
+    ltac2_CI_GITURL=https://github.com/ppedrot/ltac2
+
+fi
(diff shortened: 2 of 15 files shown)


mattam82 (approved):
Fine with me. Maybe we should deprecate this program behavior of trying the default obligation tactic to solve holes in statements and force users to write a separated Program Definition for their statements instead? I don't think there are many uses of that feature in the wild but I might be wrong.

ppedrot:
Dunno, but this can be done at a later stage.

CHANGES.md
```diff
@@ -24,6 +24,9 @@ Tactics
   Simplex-based proof engine. In case of regression, 'Unset Simplex'
   to get the venerable Fourier-based engine.
 
+Vernacular Commands
+
+- Removed the `Implicit Tactic` family of commands.
```
Zimmi48: Maybe add the "deprecated" qualifier here so that people can see at a glance that we are following the normal process of deprecating first, then removing.

Zimmi48 (approved):

ejgallego (approved):

ejgallego:
What is blocking this?

ppedrot:
Nothing that I know of. I'll rebase the PR and the overlay, though.

gares (approved):

ejgallego:
@ppedrot it seems the overlay file has a problem.

ppedrot:
Doesn't look like this has to do with the overlay, but rather that the server had some connectivity issues. I tried to relaunch the CI, let's see if it goes better.

dev/ci/user-overlays/08779-ppedrot-rm-implicit-tactic.sh
```diff
@@ -0,0 +1,8 @@
+_OVERLAY_BRANCH=rm-implicit-tactic
+
+if [ "$CI_PULL_REQUEST" = "8779" ] || [ "$CI_BRANCH" = "$_OVERLAY_BRANCH" ]; then
+
+    ltac2_CI_BRANCH="$_OVERLAY_BRANCH"
+    ltac2_CI_REF=https://github.com/ppedrot/ltac2
```
SkySkimmer: should be CI_REF=$branch and CI_GITURL=https://...

ppedrot: Arf, stupid me. Let's fix it.

ppedrot:
@mattam82 Ready to merge.

PR #8783, opened 2018-10-20
[stm] [kernel] Remove support for `vio`, `vio -> vo`

ejgallego:
This is an experiment to assess what is the exact complexity of
supporting `vio -> vo` as opposed to a simpler mode where opaque
proofs are just not checked.

The patch is not intended to be merged in any form.

I did the removal in a very quick way, more refactoring may be
possible.

A good side-effect of this PR is that a few kernel interfaces stop
depending on `Future.t`, which IMHO is a good thing as they should
not. The inner workings of `library` are also greatly simplified.

Also, an ugly "hook" in `Opaqueproof` goes away, this seemed a bit
worrying as the kernel should not have such hooks set from the upper
layers. [note one extra hook do remain]


ejgallego:
This also raises the question about the `Indirect` loading of opaque proof objects. It seems to me that instead of having `Safe_typing` use the `Opaqueproof` hack with the callback to library, safe_typing should indeed get an opaque table with the lazy structures themselves; this would IMVHO make the code clearer and above all, provide more flexibility to the upper layers as to how to deal with opaque proof themselves.

ppedrot (commented):
It does clean quite a bit indeed. If we remove this feature, this essentially allows to provide #8642 as an efficient replacement for the quick stuff. I am refraining from approving the PR as it is for now a mere experiment and we need more feedback.

SkySkimmer:
Added a commit with a bit more simplification in opaqueproof.

gares:
@ejgallego pointed out at the last WG that there are users, eg @palmskog, of this feature so it is not going away (cfr https://dl.acm.org/citation.cfm?doid=3213846.3213877 )

Discussions lead to considering merging .vio with .vos (see #8642 ) and discuss with users in order to improve the feature. Some of us may meet in person at CoqPL.

ejgallego:
I think I will indeed close this PR very soon. I wrote on Gitter:

> @palmskog we did discuss a bit on the .vio removal and current API for proof checking, @gares and @maximedenes will be a CoqPL so more chat is possible; there were these points:
> - indeed removing vio would impact your work folks, so I think there are no immediate plans to proceed to remove this for now,
> - vos/vok should be merged, however it may be not declared stable, so the interface can change suddenly or disappear,
> - the general plan would indeed to try to have for 8.10/8.11 a unified API that would satisfy the needs of you folks; Coq developers said that it would be great if you could provide a bit of feedback on how your ideal API would look like.

PR #8784, opened 2018-10-20
[dune] Remove rule for cLexer.ml4 -> cLexer.ml

ejgallego:
When merging #8740 we didn't remove this rule.

The build didn't fails as Dune emits a warning for now [due to
compatibility with some schemes], but this will become an error in the
future.

### DIFF
--- a/parsing/dune
+++ b/parsing/dune
@@ -5,11 +5,6 @@
  (libraries proofs))
 
 (rule
- (targets cLexer.ml)
- (deps (:ml4-file cLexer.ml4))
- (action (run camlp5o -loc loc -impl %{ml4-file} -o %{targets})))
-
-(rule
  (targets g_prim.ml)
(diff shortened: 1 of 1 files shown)


ejgallego:
@herbelin what are your thoughts about this feature? Would you be the right assignee?

herbelin:
> @herbelin what are your thoughts about this feature?

Same as in #8740, mitigated, but no better alternative to propose.

> Would you be the right assignee?

I would rather see @ppedrot as assignee.

ejgallego:
Sorry @herbelin , this PR is trivial, and I cited you in mistake, I meant to post the message in #8766

Zimmi48 (approved):
I wonder about the remaining `coqmlp5` and `coqp5` in `grammar/dune` and `tools/coq_dune.ml`. Is it time to remove them?

Zimmi48 (approved):

PR #8787, opened 2018-10-21
[kernel] Move lazy object loading to the kernel.

ejgallego:
This fixes a key problem in regards of making the kernel
self-contained as it removes the "indirect_opaque_accessor" hook, used
to perform lazy-loading of opaque data.

Indeed, that part was a key component of the trusted base, however it
was living in library and made available to the kernel thru a "hook".

Instead, we add to the environment information about where the opaque
data lives and fetch it under demand. We need to be careful as to keep
the environments marshallable thus we add local cache to
`Opaqueproofs`.

This refactors quite a bit of code, and IMO opens the door to further
improvements in `Opaqueproofs`.

### DIFF
--- a/checker/check.ml
+++ b/checker/check.ml
@@ -59,7 +59,9 @@ type library_t = {
   library_opaques : seg_proofs option;
   library_deps : (compilation_unit_name * Safe_typing.vodigest) array;
   library_digest : Safe_typing.vodigest;
-  library_extra_univs : Univ.ContextSet.t }
+  library_extra_univs : Univ.ContextSet.t;
+  library_opaque_disk : Opaqueproof.disk_data;
+}
 
 module LibraryOrdered =
   struct
@@ -88,29 +90,6 @@ let library_full_filename dir = (find_library dir).library_filename
 let register_loaded_library m =
   libraries_table := LibraryMap.add m.library_name m !libraries_table
 
-(* Map from library names to table of opaque terms *)
-let opaque_tables = ref LibraryMap.empty
-let opaque_univ_tables = ref LibraryMap.empty
-
-let access_opaque_table dp i =
-  let t =
-    try LibraryMap.find dp !opaque_tables
-    with Not_found -> assert false
-  in
-  assert (i < Array.length t);
-  t.(i)
-
-let access_opaque_univ_table dp i =
-  try
-    let t = LibraryMap.find dp !opaque_univ_tables in
-    assert (i < Array.length t);
-    Some t.(i)
-  with Not_found -> None
-
-let () =
-  Opaqueproof.set_indirect_opaque_accessor access_opaque_table;
-  Opaqueproof.set_indirect_univ_accessor access_opaque_univ_table
-
 let check_one_lib admit senv (dir,m) =
   let md = m.library_compiled in
   let dig = m.library_digest in
@@ -121,11 +100,11 @@ let check_one_lib admit senv (dir,m) =
     if LibrarySet.mem dir admit then
       (Flags.if_verbose Feedback.msg_notice
          (str "Admitting library: " ++ pr_dirpath dir);
-       Safe_checking.unsafe_import senv md m.library_extra_univs dig)
+       Safe_checking.unsafe_import senv md m.library_extra_univs dig m.library_opaque_disk)
     else
       (Flags.if_verbose Feedback.msg_notice
          (str "Checking library: " ++ pr_dirpath dir);
-       Safe_checking.import senv md m.library_extra_univs dig)
+       Safe_checking.import senv md m.library_extra_univs dig m.library_opaque_disk)
   in
     register_loaded_library m; senv
 
@@ -278,14 +257,16 @@ type library_disk = {
   md_objects : library_objects;
 }
 
-let mk_library sd md f table digest cst = {
+let mk_library sd md f table digest cst dd = {
   library_name = sd.md_name;
   library_filename = f;
   library_compiled = md.md_compiled;
   library_opaques = table;
   library_deps = sd.md_deps;
   library_digest = digest;
-  library_extra_univs = cst }
+  library_extra_univs = cst;
+  library_opaque_disk = dd;
+ }
 
 let name_clash_message dir mdir f =
   str ("The file " ^ f ^ " contains library") ++ spc () ++
@@ -371,15 +352,12 @@ let intern_from_file ~intern_mode (dir, f) =
       sd,md,table,opaque_csts,digest
     with e -> Flags.if_verbose chk_pp (str" failed!]" ++ fnl ()); raise e in
   depgraph := LibraryMap.add sd.md_name sd.md_deps !depgraph;
-  Option.iter (fun table -> opaque_tables := LibraryMap.add sd.md_name table !opaque_tables) table;
-  Option.iter (fun (opaque_csts,_,_) ->
-    opaque_univ_tables :=
-      LibraryMap.add sd.md_name opaque_csts !opaque_univ_tables)
-    opaque_csts;
   let extra_cst =
     Option.default Univ.ContextSet.empty
       (Option.map (fun (_,cs,_) -> cs) opaque_csts) in
-  mk_library sd md f table digest extra_cst
+  (* XXX: FIXME *)
+  let empty_disk_data = Obj.magic 0 in
+  mk_library sd md f table digest extra_cst (empty_disk_data, None)
 
 let get_deps (dir, f) =
   try LibraryMap.find dir !depgraph
--- a/checker/safe_checking.ml
+++ b/checker/safe_checking.ml
@@ -11,14 +11,14 @@
 open Declarations
 open Environ
 
-let import senv clib univs digest =
+let import senv clib univs digest dd =
   let mb = Safe_typing.module_of_library clib in
   let env = Safe_typing.env_of_safe_env senv in
   let env = push_context_set ~strict:true mb.mod_constraints env in
   let env = push_context_set ~strict:true univs env in
   let env = Modops.add_retroknowledge mb.mod_retroknowledge env in
   Mod_checking.check_module env mb.mod_mp mb;
-  let (_,senv) = Safe_typing.import clib univs digest senv in senv
+  let (_,senv) = Safe_typing.import clib univs digest dd senv in senv
 
-let unsafe_import senv clib univs digest =
-  let (_,senv) = Safe_typing.import clib univs digest senv in senv
+let unsafe_import senv clib univs digest dd =
+  let (_,senv) = Safe_typing.import clib univs digest dd senv in senv
--- a/kernel/delayed.ml
+++ b/kernel/delayed.ml
@@ -1,0 +1,43 @@
+(************************************************************************)
+(*         *   The Coq Proof Assistant / The Coq Development Team       *)
+(*  v      *   INRIA, CNRS and contributors - Copyright 1999-2018       *)
+(* <O___,, *       (see CREDITS file for the list of authors)           *)
+(*   \VV/  **************************************************************)
+(*    //   *    This file is distributed under the terms of the         *)
+(*         *     GNU Lesser General Public License Version 2.1          *)
+(*         *     (see LICENSE file for the text of the license)         *)
+(************************************************************************)
+
+(************************************************************************)
+(** Serialized objects loaded on-the-fly *)
+
+exception Faulty of string
+
+type 'a t = {
+  del_file : string;
+  del_off : int;
+  del_digest : Digest.t;
+}
+
+let in_delayed f ch =
+  let pos = pos_in ch in
+  let _, digest = System.skip_in_segment f ch in
+  ({ del_file = f; del_digest = digest; del_off = pos; }, digest)
+
+(** Fetching a table of opaque terms at position [pos] in file [f],
+    expecting to find first a copy of [digest]. *)
+
+let raw_intern_library f =
+  System.with_magic_number_check
+    (System.raw_intern_state Coq_config.vo_magic_number) f
+
+let fetch del =
+  let { del_digest = digest; del_file = f; del_off = pos; } = del in
+  try
+    let ch = raw_intern_library f in
+    let () = seek_in ch pos in
+    let obj, _, digest' = System.marshal_in_segment f ch in
+    let () = close_in ch in
+    if not (String.equal digest digest') then raise (Faulty f);
+    obj
+  with e when CErrors.noncritical e -> raise (Faulty f)
--- a/kernel/delayed.mli
+++ b/kernel/delayed.mli
@@ -8,9 +8,8 @@
 (*         *     (see LICENSE file for the text of the license)         *)
 (************************************************************************)
 
-(*i*)
-open Safe_typing
-(*i*)
+exception Faulty of string
 
-val import : safe_environment -> compiled_library -> Univ.ContextSet.t -> vodigest -> Opaqueproof.disk_data -> safe_environment
-val unsafe_import : safe_environment -> compiled_library -> Univ.ContextSet.t -> vodigest -> Opaqueproof.disk_data -> safe_environment
+type 'a t
+val in_delayed : string -> in_channel -> 'a t * Digest.t
+val fetch : 'a t -> 'a
--- a/kernel/environ.ml
+++ b/kernel/environ.ml
@@ -95,6 +95,7 @@ type env = {
   env_typing_flags  : typing_flags;
   retroknowledge : Retroknowledge.retroknowledge;
   indirect_pterms : Opaqueproof.opaquetab;
+  opaque_disk_data : Opaqueproof.DiskData.t
 }
 
 let empty_named_context_val = {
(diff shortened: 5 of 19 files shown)


gares:
Ftr, the trusted code base does not coincide with the kernel directory, part of clib is also there for example. I'm not so sure the kernel should be  self contained either. In particular I'd prefer to have two components, both trusted, one doing type checking and one doing library management rather than one doing both. In other words putting code doing unmarshaling and checksums in the same pleace where type checking is does not seem an improvement to me.

ejgallego:
> Ftr, the trusted code base does not coincide with the kernel directory, part of clib is also there for example.

I don't think this is relevant to this PR, when we say kernel is self-contained I guess we assume kernel + its dependencies.

> I'm not so sure the kernel should be self contained either. In particular I'd prefer to have two components, both trusted, one doing type checking and one doing library management rather than one doing both. In other words putting code doing unmarshaling and checksums in the same pleace where type checking is does not seem an improvement to me.

The problem is that in the current state of affairs, type-checking won't fully work unless you link the library management component. (*) So they are not really two separate components, an in fact this PR makes the kernel into a component proper.

(*) that's not strictly true as indeed type checking can work to a extent with non-available opaque proofs.

ejgallego:
> In other words putting code doing unmarshaling and checksums in the same pleace where type checking is does not seem an improvement to me.

So do you prefer the current situation then?

gares:
I think that I prefer the current situation to what it would be after this PR. I see that a static hook makes the two components tied together, but maybe it would be sufficient to make the callback explicit, passing an higher order argument to the code requiring access to the opaque proofs.

Btw, as far as I remember, the only part of the kernel that may require loading opaque proofs is not type checking, but rather the implementation of `Include`. If I'm not mistaken, making include *point* *to* rather than *copy* the included (opaque) proofs would solve the dependency among the two components as well. Indeed for the type checking algorithm, on paper, opaque proofs are like axioms, they have no accessible body. Finally, the current implementation of `Include` is also what makes the `-quick` compilation chain (both the current one and the one proposed by @charguer) struggle, since in both cases the opaque proofs `Include` wants to copy are *not there* (but they have names, so they could be pointed to).

gares:
I think that I prefer the current situation to what it would be after this PR. I see that a static hook makes the two components tied together, but maybe it would be sufficient to make the callback explicit, passing an higher order argument to the code requiring access to the opaque proofs.

Btw, as far as I remember, the only part of the kernel that may require loading opaque proofs is not type checking, but rather the implementation of `Include`. If I'm not mistaken, making include *point* *to* rather than *copy* the included (opaque) proofs would solve the dependency among the two components as well. Indeed for the type checking algorithm, on paper, opaque proofs are like axioms, they have no accessible body. Finally, the current implementation of `Include` is also what makes the `-quick` compilation chain (both the current one and the one proposed by @charguer) struggle, since in both cases the opaque proofs `Include` wants to copy are *not there* (but they have names, so they could be pointed to).

gares:
I think that I prefer the current situation to what it would be after this PR. I see that a static hook makes the two components tied together, but maybe it would be sufficient to make the callback explicit, passing an higher order argument to the code requiring access to the opaque proofs.

Btw, as far as I remember, the only part of the kernel that may require loading opaque proofs is not type checking, but rather the implementation of `Include`. If I'm not mistaken, making include *point* *to* rather than *copy* the included (opaque) proofs would solve the dependency among the two components as well. Indeed for the type checking algorithm, on paper, opaque proofs are like axioms, they have no accessible body. Finally, the current implementation of `Include` is also what makes the `-quick` compilation chain (both the current one and the one proposed by @charguer) struggle, since in both cases the opaque proofs `Include` wants to copy are *not there* (but they have names, so they could be pointed to).

gares:
I think that I prefer the current situation to what it would be after this PR. I see that a static hook makes the two components tied together, but maybe it would be sufficient to make the callback explicit, passing an higher order argument to the code requiring access to the opaque proofs.

Btw, as far as I remember, the only part of the kernel that may require loading opaque proofs is not type checking, but rather the implementation of `Include`. If I'm not mistaken, making include *point* *to* rather than *copy* the included (opaque) proofs would solve the dependency among the two components as well. Indeed for the type checking algorithm, on paper, opaque proofs are like axioms, they have no accessible body. Finally, the current implementation of `Include` is also what makes the `-quick` compilation chain (both the current one and the one proposed by @charguer) struggle, since in both cases the opaque proofs `Include` wants to copy are *not there* (but they have names, so they could be pointed to).

gares:
I think that I prefer the current situation to what it would be after this PR. I see that a static hook makes the two components tied together, but maybe it would be sufficient to make the callback explicit, passing an higher order argument to the code requiring access to the opaque proofs.

Btw, as far as I remember, the only part of the kernel that may require loading opaque proofs is not type checking, but rather the implementation of `Include`. If I'm not mistaken, making include *point* *to* rather than *copy* the included (opaque) proofs would solve the dependency among the two components as well. Indeed for the type checking algorithm, on paper, opaque proofs are like axioms, they have no accessible body. Finally, the current implementation of `Include` is also what makes the `-quick` compilation chain (both the current one and the one proposed by @charguer) struggle, since in both cases the opaque proofs `Include` wants to copy are *not there* (but they have names, so they could be pointed to).

gares:
I think that I prefer the current situation to what it would be after this PR. I see that a static hook makes the two components tied together, but maybe it would be sufficient to make the callback explicit, passing an higher order argument to the code requiring access to the opaque proofs.

Btw, as far as I remember, the only part of the kernel that may require loading opaque proofs is not type checking, but rather the implementation of `Include`. If I'm not mistaken, making include *point* *to* rather than *copy* the included (opaque) proofs would solve the dependency among the two components as well. Indeed for the type checking algorithm, on paper, opaque proofs are like axioms, they have no accessible body. Finally, the current implementation of `Include` is also what makes the `-quick` compilation chain (both the current one and the one proposed by @charguer) struggle, since in both cases the opaque proofs `Include` wants to copy are *not there* (but they have names, so they could be pointed to).

ejgallego:
I have made this PR standalone. IMHO it can be still improved a bit, I am curious to know what the kernel maintainers think.

ejgallego:
I've deleted 10 duplicated comments due to the Github glitch.

gares:
I think that I prefer the current situation to what it would be after this PR. I see that a static hook makes the two components tied together, but maybe it would be sufficient to make the callback explicit, passing an higher order argument to the code requiring access to the opaque proofs.

Btw, as far as I remember, the only part of the kernel that may require loading opaque proofs is not type checking, but rather the implementation of `Include`. If I'm not mistaken, making include *point* *to* rather than *copy* the included (opaque) proofs would solve the dependency among the two components as well. Indeed for the type checking algorithm, on paper, opaque proofs are like axioms, they have no accessible body. Finally, the current implementation of `Include` is also what makes the `-quick` compilation chain (both the current one and the one proposed by @charguer) struggle, since in both cases the opaque proofs `Include` wants to copy are *not there* (but they have names, so they could be pointed to).

maximedenes:
> I am curious to know what the kernel maintainers think.

I think it would be worth thinking about solving the real problem related to `Include`, which would just remove the need for loading opaque proofs from the kernel IIUC. This is not unrelated to what @ppedrot is doing on canonical names (in particular, it makes `Include`d inductive types point to the definition, we could have something similar for opaque proofs).

ejgallego:
> I think it would be worth thinking about solving the real problem related to `Include`, which would just remove the need for loading opaque proofs from the kernel IIUC.

Umm, I need to understand more what the problem with `Include` is, however I don't think this loading code is only triggered by `Include` but some heavy code related to the side-effects is also using it.

ppedrot (commented):
I support @maximedenes and @gares point. This PR is probably an unsatisfying solution that side-steps the actual problem.

ejgallego:
> I support @maximedenes and @gares point. This PR is probably an unsatisfying solution that side-steps the actual problem.

Well, this certainly deserves more thinking / evaluation of the current situation as for one the kernel will just not work today unless linked to `library.cmxa`.

I feel that we are mixing here "this is how it should be in theory" vs "this is how it is today", and I'm afraid the two differ a bit. I am not sure putting the problem under the rug [as the hook does] is the best approach, but I will revisit this code soon tho.

ejgallego:
So what should we do here; I'm afraid the reasoning exposed here doesn't fully typecheck to me. We have two problems:

- problem 1: the kernel cannot work _today_ as an standalone component;
- problem 2: the kernel should not need to load this objects, `Include` problem, etc...

Problem 1 is real and fixing it doesn't preclude fixing 2, in fact it makes 2 more obvious so we may have a better chance to address it.

ejgallego:
I need to fix the checker.

gares:
Hum, I think things improved substantially from the last time I looked at include. With master `-quick` works with `Include`!!

```
gares@ollypat:~/COQ/coq/xxx$ cat a.v
Module X.
	Lemma x : True.
	Proof.
	trivial.
	Qed.
End X.

gares@ollypat:~/COQ/coq/xxx$ cat b.v
From xxx Require a.
Module Y. Include a.X. End Y.

gares@ollypat:~/COQ/coq/xxx$ ../bin/coqc -R . xxx -quick a.v
gares@ollypat:~/COQ/coq/xxx$ ../bin/coqc -R . xxx -quick b.v

gares@ollypat:~/COQ/coq/xxx$ ../bin/coqtop -R . xxx
Welcome to Coq ollypat:/home/gares/COQ/coq,ssr/fix-TC-case (f9e526bb936cc2c9cffe3995e35d976eb40014e9)

Coq < From xxx Require a b.

Coq < Print Module b.Y.
Module b.Y := Struct Definition x : True. End


Coq < Print b.Y.x.    
Monomorphic b.Y.x = a.X.x
     : True

b.Y.x is not universe polymorphic

> Print a.X.x.
> ^^^^^^^^^^^^
Error: The value you are asking for (x) is not available in this process. If
you really need this, pass the "-async-proofs off" option to CoqIDE to
disable asynchronous script processing and don't pass "-quick" to coqc.
```

So, going back to this PR: 2) was already solved by someone (not me). Hence the PR should try to *remove* from the kernel the hook and have `Extraction` and `Print (Assumption)` do query `Library` for the opaque bodies, instead of asking the kernel.

It is quite some refactoring I believe. So maybe we should think about it.
The point is that we probably want to keep in the kernel a handler to the future proofs, so that `join_safe_environment` can still be implemented as we do today. But we should provide no API to access the opaque bodies (the value inside the future proof), and re-route these queries to Library that stays in charge or storing them (and lazily fetch them on demand).

maximedenes:
@gares can you point to the `Include` code that was changed?

gares:
Indeed! There was a commit separating module bodies from module types.
Module types are still "copied", as @SkySkimmer suggests if you `sed s/Module/Module Type` you get:
```
gares@ollypat:~/COQ/coq/xxx$ ../bin/coqc -R . xxx -quick a.v
gares@ollypat:~/COQ/coq/xxx$ vim b.v
gares@ollypat:~/COQ/coq/xxx$ ../bin/coqc -R . xxx -quick b.v
Error:
The value you are asking for (x) is not available in this process. If you
really need this, pass the "-async-proofs off" option to CoqIDE to disable
asynchronous script processing and don't pass "-quick" to coqc.
frame @ file "kernel/safe_typing.ml", line 1091, characters 8-42
frame @ file "list.ml", line 117, characters 24-34
frame @ file "kernel/safe_typing.ml", line 343, characters 41-55
frame @ file "lib/future.ml", line 167, characters 10-18
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "toplevel/coqtop.ml", line 501, characters 8-38
frame @ file "list.ml", line 106, characters 12-15
frame @ file "toplevel/coqtop.ml", line 295, characters 2-44
frame @ file "toplevel/coqtop.ml", line 281, characters 17-58
frame @ file "stm/stm.ml", line 2771, characters 2-99
frame @ file "library/library.ml", line 693, characters 23-58
frame @ file "library/global.ml", line 146, characters 23-65
frame @ file "kernel/safe_typing.ml", line 1094, characters 6-55
raise @ file "lib/loc.ml", line 89, characters 16-23
```

gares:
Unfortunately the trace does not show where the future is copied, but rather where it is found to be empty.

gares:
Anyway, having in the interface of a module an opaque proof seems quite odd to me, why would one need that? Is there a use case? If we forbid that, then `Include` and `-quick` may become 100% compatible. (but this is a bit off topic, and this PR should stay focused on the way to remove the hook).

SkySkimmer:
It's not just module type, the same issue happens with the functor include (ie a.v is
~~~coq
Module Type E. End E.
Module X (A:E).
        Lemma x : True.
	Proof.
	trivial.
	Qed.
End X.
~~~

gares:
> functor include

Another thing that should be killed.

Can you say where this feature forces the body of the functor? Is it just functor application that is implemented this way?

SkySkimmer:
>Can you say where this feature forces the body of the functor?

It's the universe constraints. `add_field` at the end of `add_include` in safe typing will copy the constraints.
I'll try with 9b0305ac203f238695ed2dc85335aa15658eea5a, if it works it will be worth submitting.

SkySkimmer:
Yeah, it works.

ejgallego:
> Fixing 2 also fixes 1. This is why we don't like this pr: in order to fix 1 it makes 2 worse since it moves to the kernel more code that does not belong there.

The code is already there, by means of a hook, thus it is even worse and static analysis doesn't reveal what code is really being run.

ejgallego:
Of course I'd be great if we can remove the hook altogether, but indeed I think my point is valid in that exposing the current reality is going to be beneficial as people is gonna be more aware/motivated to kill the root cause.

maximedenes:
@SkySkimmer does the kernel still need to observe opaque bodies after #8995 ?

SkySkimmer:
>@SkySkimmer does the kernel still need to observe opaque bodies after #8995 ?

I don't know, it's not quite clear to me what "observe opaque bodies" means.
I think we break opacity for side-effects for instance, does that count? ie https://github.com/coq/coq/blob/f6a2d21b6c2a93cb70fde235fc897fb75ea51384/kernel/safe_typing.ml#L227-L234 and use through https://github.com/coq/coq/blob/f6a2d21b6c2a93cb70fde235fc897fb75ea51384/kernel/safe_typing.ml#L286-L304

gares:
I don't think it counts. The example about module types or functors does not use side effects (unless we use the make_eff function for them too, but I don't think so).

gares:
(side effects are for proofs that use `abstract`)

maximedenes:
> I don't think it counts. The example about module types or functors does not use side effects (unless we use the make_eff function for them too, but I don't think so).

I'm a bit lost. Does it mean the hook is not necessary anymore?

gares:
`get_opaque_body` is what makes the hook needed.

`make_eff` is not used for module types, so if it is the only user of the former, then the hook can be removed. I think it is not the only client.

gares:
My impression is that the approach proposed here is not OK, it is moving the code loading opaque proofs where it should not be.
The approach in CEP https://github.com/coq/ceps/pull/40 seemed to be consensual when it was discussed at the WG. (But has not implementation yet...).

What should we do with this PR? IMO we should close it, since it is not the code, but the whole approach that is problematic.

ejgallego:
> What should we do with this PR? IMO we should close it, since it is not the code, but the whole approach that is problematic.

I think the approach is as problematic as the current one; this only makes the reality explicit. I am waiting to rebase for other large PRs to come in [such as the vos / vok], I guess after that we can evaluate again.

ejgallego:
@ppedrot I updated this to your latest changes, however it seems to me that while we don't need to get bodies from the kernel anymore, universe constraints from disk are still needed ?

Do you expect to lift that restriction too?

ppedrot:
@ejgallego I'm about to submit a PR that removes all accesses to indirect opaques in the kernel, so, yes.

ejgallego:
Oh, I forgot about this case:
```
(** {6 Stm machinery } *)
let join_constant_body except otab cb =
  match cb.const_body with
  | OpaqueDef o -> Opaqueproof.join_opaque ~except otab o
  | _ -> ()
```
> @ejgallego I'm about to submit a PR that removes all accesses to indirect opaques in the kernel, so, yes.

ok cool, looking forward to it, then I may see if there's any change remaining here.

ppedrot:
@ejgallego the `join_opaque` function does not do anything for on-disk constants, so it's not a problem.

ejgallego:
> @ejgallego the `join_opaque` function does not do anything for on-disk constants, so it's not a problem.

It does not? At least on master it seems it grabs the body:
```
let join_opaque ?except { opaque_val = prfs; opaque_dir = odp; _ } = function
  | Direct (_,cu) -> join except cu
  | Indirect (_,dp,i) ->
      if DirPath.equal dp odp then
        let fp = snd (Int.Map.find i prfs) in
        join except fp
```

ppedrot:
@ejgallego The if branch is true only for the library being compiled, which is not stored on disk. You can see that, as you don't call the hooks in the `join_opaque` function.

ejgallego:
Oh indeed, the find is local, my bad, long time since I looked into this.

ejgallego:
Superseded by #10201

PR #8788, opened 2018-10-21
[kernel] Handle opaque bodies and constraints together.

ejgallego:
We simplify the opaque fetching API as to return a body and their
constraints together.

This IMHO makes sense as the underlying future has them together and
thus the whole entity is forced no matter what we'd like to retrieve.

Depends on #8787 but could be made independent.


ppedrot (commented):
This is a *cavalier législatif*, this PR also includes #8783... Can't we do that independently?

kernel/opaqueproof.mli
```diff
@@ -46,12 +46,8 @@ module DiskData : sig
...
 
 (* val mk_fetch : Constr.t Future.computation array Delayed.t -> opaque_disk_data *)
-val force_proof : DiskData.t -> opaquetab -> opaque -> constr
-val get_proof : DiskData.t -> opaquetab -> opaque -> constr Future.computation
-
-val force_constraints : opaquetab -> opaque -> Univ.ContextSet.t
-val get_constraints :
-  opaquetab -> opaque -> Univ.ContextSet.t Future.computation option
+val force_cst : DiskData.t -> opaquetab -> opaque -> constr * Univ.ContextSet.t
+val get_cst : DiskData.t -> opaquetab -> opaque -> (constr * Univ.ContextSet.t) Future.computation
```
gares: Why `_cst`? It is both the proof and the constraints, while cst seems to refer only to constraints

gares (changes requested):
It is a nice simplification, but I'm not so happy with the name of the new API.
I'd say `_proof` is ok (I could live with the proof being "implemented" by a term and a constraints set), `_proof_cst` also.

I also think there is a type combinator for the pair `('a * Cuniv.ContextSet.t)` somewhere.

I'm reviewing only the last commit.

ejgallego:
This is not needed anymore after some changes in master.

Issue #8793, opened 2018-10-22
Report on the 8.7 to 8.9 compatibility of packages in coq-contribs


herbelin:
This is a summary about porting "contribs" to 8.7 and 8.9, and to produce opam packages for them.

#### Porting

I discontinued the following "obsolete" packages: `jprover`, `legacy-ring` (`ring` version 2), `legacy-field` (`field` version 1), `rational` (`ring` version 1), string (source of current `string`), `micromega` (empty directory). If no objection from @ybertot, I would also remove `recursive-definition` which is subsumed by `Function` and `Equations`.

Five developments have moved from `coq-contribs` to `coq-community`: `aac-tactics`, `lemma-overloading`, `stalmarck`, `bertrand`, `qarith-stern-brocot`.

I failed to (easily) upgrade relatively to the following Coq changes:
- change of representation of real constants: `high-school-geometry` (I renounced)
- change of naming of hypotheses after removal of option `Standard Proposition Elimination Names`: `finger-tree` (I temporarily renounced)
- change in the API:
  - `counting`: due to removal of a declaration hook (this is a amusing plugin though, I'm probably going to submit a way to continue it)
  - `ergo`: due to difficulty in managing hash-tables on `econstr` (I temporarily renounced, but this is an interesting plugin)

Something should be done to remove `intuitionistic-nuprl` which is maintained upstream. Also for `relation-algebra` which is maintained upstream too.

#### Opam packages

There are scripts in `coq-contribs/coq-contribs/META` to create opam packages. I produced them for all contributions which compile in 8.6, 8.7 and 8.8. We could in theory weaken the constraints when a package is known to compile with several versions, but it is still to do.

I basically reused the encoding of coq-contribs style of metadata into opam which Matej did. Don't hesitate to make suggestions on conventions (keywords, format of authors, ...). I used 8.X.0 as version number but that's maybe not a very good idea. Something like 1.0+8.8 may be better?

Note: `lemma-overloading` (in coq-community) does not have an opam package.

#### Web page

I have a(n unpolished) tool to automatically build a web page for a package (with coqdoc-style file browsing). To be continued.

#### Upgrading to 8.7

The main difficulties to upgrade to 8.7 have been:
- the change of representation of constants in `R` (pretty difficult, was requiring thinking, sometimes gives simplification)
- adding `Require Extraction`/`Require FunInd` (relatively easy, but often)
- `Save` -> `Qed` (easy, sed-style)
- change of behavior of `{  }` in notations
[...]

PR #8795, opened 2018-10-22
Encapsulating declarations of primitive string syntax in a module

herbelin:
**Kind:** "bug fix"

This addresses @JasonGross' comment [here](https://github.com/coq/coq/issues/8793#issuecomment-431951074)

This is supposed to go together with @jasongross `fix.py` script for primitive string notations not being available by default anymore.

Drawback: This makes the fix needs two steps, a `Require` and an `Import`, with not so nice names.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -138,8 +138,9 @@ Standard Library
 
 - Syntax notations for `string`, `ascii`, `Z`, `positive`, `N`, `R`,
   and `int31` are no longer available merely by `Require`ing the files
-  that define the inductives.  You must `Import` `Coq.Strings.String`,
-  `Coq.Strings.Ascii`, `Coq.ZArith.BinIntDef`, `Coq.PArith.BinPosDef`,
+  that define the inductives.  You must `Import` `Coq.Strings.String.StringSyntax`
+  (after `Require` `Coq.Strings.String`), `Coq.Strings.Ascii.AsciiSyntax` (after
+  `Require` `Coq.Strings.Ascii`), `Coq.ZArith.BinIntDef`, `Coq.PArith.BinPosDef`,
   `Coq.NArith.BinNatDef`, `Coq.Reals.Rdefinitions`, and
   `Coq.Numbers.Cyclic.Int31.Int31`, respectively, to be able to use
   these notations.  Note that passing `-compat 8.8` or issuing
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -138,8 +138,9 @@ Standard Library
 
 - Syntax notations for `string`, `ascii`, `Z`, `positive`, `N`, `R`,
   and `int31` are no longer available merely by `Require`ing the files
-  that define the inductives.  You must `Import` `Coq.Strings.String`,
-  `Coq.Strings.Ascii`, `Coq.ZArith.BinIntDef`, `Coq.PArith.BinPosDef`,
+  that define the inductives.  You must `Import` `Coq.Strings.String.StringSyntax`
+  (after `Require` `Coq.Strings.String`), `Coq.Strings.Ascii.AsciiSyntax` (after
```
JasonGross: Is there any reason to do
```md
`Import` `Coq.Strings.String.StringSyntax`
```
rather than
```md
`Import Coq.Strings.String.StringSyntax`
```
?

JasonGross (approved):
Perhaps the relevant `Compat/Coq*.v` file should be updated to `Export` the syntax modules rather than re-declaring the ML modules.

In any case, LGTM.

CHANGES.md
```diff
@@ -138,8 +138,9 @@ Standard Library
 
 - Syntax notations for `string`, `ascii`, `Z`, `positive`, `N`, `R`,
```
JasonGross: This should be moved to the changes from 8.8 to 8.9 section, right?

Zimmi48:
@JasonGross You reviewed but did not self assign. It also seems that @herbelin took your comments into account but I'm not quite sure.

JasonGross (approved):

silene:
Do I understand correctly that, despite what the changes to `CHANGES.md` seem to say, the 8.9.0 recommended way of doing `Import Coq.Strings.String` still works fine after this patch?

JasonGross:
Yes, you can either `Import Coq.Strings.String` or `Import Coq.Strings.String.StringSyntax` after `Require Coq.Strings.String`

silene:
Actually, this PR depends on #7135, which is not 8.9 material, so not backporting. If this is nonetheless needed for 8.9, please make an adhoc pull request against 8.9.

JasonGross:
This PR does not depend on #7135, it's just that git cannot figure out the right place to apply the diff without #7135.  I've manually resolved the conflicts and made #9903

PR #8796, opened 2018-10-22
Elementary properties about IZR for generic use

herbelin:
**Kind:** library

These are three trivial lemmas useful for dealing with real constants when writing Ltac-style translators from `R` to another field (examples e.g. in [qarith-stern-brocot](https://github.com/coq-community/qarith-stern-brocot/blob/770eb8147b2d7ba765837d03655d8f3adba967e6/Q_to_R.v#L757-L770)).

### DIFF
--- a/theories/Reals/RIneq.v
+++ b/theories/Reals/RIneq.v
@@ -1774,6 +1774,28 @@ Proof.
   now rewrite <- INR_IPR, SuccNat2Pos.id_succ.
 Qed.
 
+Lemma IZR_NEG : forall p, IZR (Zneg p) = Ropp (IZR (Zpos p)).
+Proof.
+  reflexivity.
+Qed.
+
+(** The three following lemmas map the default form of numerical
+    constants to their representation in terms of the axioms of
+    [R]. This can be a useful intermediate representation for reifying
+    to another axiomatics of the reals. It is however generally more
+    convenient to keep constants represented under an [IZR z] form when
+    working within [R]. *)
+
+Lemma IZR_POS_xO : forall p, IZR (Zpos (xO p)) = (1+1) * (IZR (Zpos p)).
+Proof.
+  intro. unfold IZR, IPR. destruct p; simpl; trivial. rewrite Rmult_1_r. trivial.
+Qed.
+
+Lemma IZR_POS_xI : forall p, IZR (Zpos (xI p)) = 1 + (1+1) * IZR (Zpos p).
+Proof.
+  intro. unfold IZR, IPR. destruct p; simpl; trivial. rewrite Rmult_1_r. trivial.
+Qed.
+
 Lemma plus_IZR_NEG_POS :
   forall p q:positive, IZR (Zpos p + Zneg q) = IZR (Zpos p) + IZR (Zneg q).
 Proof.


herbelin:
@silene: you should be the assignee here. (I know that these are reflexivity lemmas but still useful I think as lemmas.)

herbelin:
> I am a bit annoyed by the name, though. Indeed, the name starts with opp, while Ropp does not even appear on the left-hand side.

What would you suggest? Do you think it would be better with the equality reversed? Or with name `IZR_NEG_opp_POS`?

> But I do not understand the point of the two other lemmas. Do you have a use for them? Also, why (1+1)%R rather than 2%R? Sure, they are convertible, but this choice seems strange.

The use was for automatic transfer to another implementation of `R` (see the link to qarith-stern-brocot in the initial comment).

silene:
> Do you think it would be better with the equality reversed? Or with name `IZR_NEG_opp_POS`?

The equality goes in the direction of computation (somehow), so it is fine with me. Thus, I prefer the later name.

> The use was for automatic transfer to another implementation of R.

This development predates the way literals are now represented in the `Reals` library, so I don't think it is a good source of inspiration. Nowadays, you should keep `IZR` around rather than trying to destruct it away.

herbelin:
> This development predates the way literals are now represented in the Reals library, so I don't think it is a good source of inspiration. Nowadays, you should keep IZR around rather than trying to destruct it away.

The foreign implementation (here qarith-stern-brocot) does not have an IZR' for its implementation R', so, to translate the `IZR ...` in `R` to `R'`, I don't see how to avoid mapping the `IZR` to what `R'` knows which is only `0`, `1`, `+`, `*`.

There is a similar automatic tactic for mapping from `R` to a custom `R'` in high-school-geometry (but there were too much changes to do there and I capitulate to port it).

But if you think that it is too specific to the kind of application I mentioned, I can remove them.

herbelin:
@silene: I added a comment to summarize when the last 3 lemmas may be useful and when they should not be used. Tell me if this reflects correctly your words.

ybertot:
I think that IZR_NEG_opp_POS could simply be named IZR_NEG or IZR_Zneg, under the philosophy that you only need to describe the left-hand side of the equality.

I think we really need to integrate this lemma and it is a pity that this merge is delayed by discussion on the other two.  It seems to me that @silene actually (reluctantly) accepts the whole 3 and I suggest that they should be merged.

For IZR_NEG_opp_POS (or whatever name we choose), I found another use case in porting [exact-real-arithmetic](https://github.com/coq-community/exact-real-arithmetic/).
When describing the sign of a real number, it is natural to express it as being either -1, 0, or 1, but -1 is actually (IZR (Zneg 1)) and there are no rewrite rules that work on this number (for example in multiplication).

I don't know why ci fails, yet.

herbelin:
@ybertot: No strong opinion on naming. It seems that `reals` adopts both style (cf `INR_IZR_INZ` but also `Int_part_INR`, `pow_IZR`). I made the change.

I left the two other lemmas (I may remove them if @silene prefers I remove them). Merging the first seems anyway prioritary.

For the CI, it is restarted, let's see. I guess you can assign.

SkySkimmer (approved):
No new comments have appeared for months.
@herbelin I'll merge once you've rebased.

PR #8803, opened 2018-10-23
Fix issue #8800 (gtk warning about gtk_scrolled_window_add_with_viewport)

### LINKED ISSUE #8800: gtk_scrolled_window_add(): cannot add non scrollable widget use gtk_scrolled_window_add_with_viewport() instead
Since commit 079e895cad0a205e22202da5ba1a919a820c863b

    coqide: queries from the query window are routed there (fix #5684)
    
    We systematically use Wg_MessageView for both the message panel and each
    Query tab; we register all MessageView in a RoutedMessageViews where the
    default route (0) is the message panel. Queries from the Query panel pick
    a non zero route to have their feedback message delivered to their MessageView

Coqide systematically displays the warning

    gtk_scrolled_window_add(): cannot add non scrollable widget use gtk_scrolled_window_add_with_viewport() instead

Looking at the commit, my guess is that the issue comes from the lines

    let result = Wg_MessageView.message_view () in
    router#register_route route_id result;
    r_bin#add (result :> GObj.widget);

silene:
GTK loudly complains that we are not using `gtk_scrolled_window_add_with_viewport`, so let us do.

Fixes #8800.

### DIFF
--- a/ide/wg_Command.ml
+++ b/ide/wg_Command.ml
@@ -98,7 +98,7 @@ object(self)
 	~packing:(vbox#pack ~fill:true ~expand:true) () in
     let result = Wg_MessageView.message_view () in
     router#register_route route_id result;
-    r_bin#add (result :> GObj.widget);
+    r_bin#add_with_viewport (result :> GObj.widget);
     views <- (frame#coerce, result, combo#entry) :: views;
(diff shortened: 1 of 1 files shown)


ppedrot (approved):
Typical stuff GTK complains about, never really understood the difference. As long as it works...

PR #8808, opened 2018-10-23
Extending support for mixing binders and terms in abbreviations

herbelin:
**Kind:** feature

By consistency with notations modifying the parser, add support for abbreviations which bind both a term and a pattern:
```
Notation U t := (t,fun t => t).
Check U tt.
(* gives "(tt, fun 'tt => tt)", printed "U tt" *)
Check fun x => U x.
(* gives "fun x : ?A => (x, fun x0 : ?T => x0)", printed "fun x : ?A => U x" *)
```

This is virtually a source of incompatibilities if a notation was already using a binder with same name as a notation variable.

Synchronous overlay: LPCIC/coq-elpi#130

### DIFF
--- a/doc/sphinx/user-extensions/syntax-extensions.rst
+++ b/doc/sphinx/user-extensions/syntax-extensions.rst
@@ -618,6 +618,41 @@ the next command fails because p does not bind in the instance of n.
    Notation "[> a , .. , b <]" :=
      (cons a .. (cons b nil) .., cons b .. (cons a nil) ..).
 
+Notations with expressions used both as binder and term
++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+It is possible to use parameters of the notation both in term and
+binding position. Here is an example:
+
+.. coqtop:: in
+
+   Definition force n (P:nat -> Prop) := forall n', n' >= n -> P n'.
+   Notation "▢_ n P" := (force n (fun n => P))
+     (at level 0, n ident, P at level 9, format "▢_ n  P").
+
+.. coqtop:: all
+
+   Check exists p, ▢_p (p >= 1).
+
+More generally, the parameter can be a pattern, as in the following
+variant:
+
+.. coqtop:: in reset
+
+   Definition force2 q (P:nat*nat -> Prop) :=
+     (forall n', n' >= fst q -> forall p', p' >= snd q -> P q).
+
+   Notation "▢_ p P" := (force2 p (fun p => P))
+     (at level 0, p pattern at level 0, P at level 9, format "▢_ p  P").
+
+.. coqtop:: all
+
+   Check exists x y, ▢_(x,y) (x >= 1 /\ y >= 2).
+
+This support is experimental. For instance, the notation is used for
+printing only if the occurrence of the parameter in term position
+comes in the right-hand side before the occurrence in binding position.
+
 .. _RecursiveNotations:
 
 Notations with recursive patterns
@@ -1382,6 +1417,17 @@ Abbreviations
    arguments and notation scopes of the constant. As an
    exception, if the right-hand side is just of the form :n:`@@qualid`,
    this conventionally stops the inheritance of implicit arguments.
+
+   Like for notations, it is possible to bind binders in
+   abbreviations. Here is an example:
+
+   .. coqtop:: in reset
+
+      Definition force2 q (P:nat*nat -> Prop) :=
+        (forall n', n' >= fst q -> forall p', p' >= snd q -> P q).
+
+      Notation F p P := (force2 p (fun p => P)).
+      Check exists x y, F (x,y) (x >= 1 /\ y >= 2).
 
 .. _numeral-notations:
 
--- a/vernac/metasyntax.ml
+++ b/vernac/metasyntax.ml
@@ -1073,12 +1073,12 @@ let make_internalization_vars recvars mainvars typs =
   let extratyps = List.map (fun (x,y) -> (y,List.assoc x maintyps)) recvars in
   maintyps @ extratyps
 
-let make_interpretation_type isrec isonlybinding = function
+let make_interpretation_type isrec isonlybinding default_if_binding = function
   (* Parsed as constr list *)
   | ETConstr (_,None,_) when isrec -> NtnTypeConstrList
-  (* Parsed as constr, but interpreted as a binder: default is to parse it as an ident only *)
+  (* Parsed as constr, but interpreted as a binder *)
   | ETConstr (_,Some bk,_) -> NtnTypeBinder (NtnBinderParsedAsConstr bk)
-  | ETConstr (_,None,_) when isonlybinding -> NtnTypeBinder (NtnBinderParsedAsConstr AsIdent)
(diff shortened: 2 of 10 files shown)


mattam82:
That looks quite strange to me, what is the use case? If I declare the x as
a pattern but use it as an arbitrary term in the notation body this is
allowed?
Le mar. 23 oct. 2018 à 17:00, Hugo Herbelin <notifications@github.com> a
écrit :

herbelin:
@mattam82, here is a typical example from Iris of notations _modifying the parser_ and using variables in both term and binder position:
```coq
Inductive ftele : Type :=
| fb {T:Type} : T -> ftele
| fr {T} : (T -> ftele) -> ftele.
Fixpoint args ftele : Type :=
  match ftele with
    | fb _ => unit
    | fr f => sigT (fun t => args (f t))
  end.
Definition fpack := sigT args.
Definition pack fp fa : fpack := existT _ fp fa.
Notation "'tele' x .. z := b" :=
  (fun x => .. (fun z =>
     pack (fr (fun x => .. ( fr (fun z => fb b) ) .. ) )
          (existT _ x .. (existT _ z tt) .. )
                ) ..)
  (at level 85, x binder, z binder).
```

@ejgallego
> I was under the impression that syndefs were frozen and to be replaced by notations. Isn't that the case?

Syndefs (i.e. abbreviations, i.e. notations of the form `Notation f args := ...`) are quite popular, especially because they don't modify the parser. I don't see how we could renounce to them.

herbelin:
Any other opinion on whether it is more useful or more troubling to allow abbreviations to bind variables as in the example above?

One could be more restrictive and that would be easy to implement (maybe no user is anyway really using the ability to bind variables in abbreviations). Or one could continue to support them (that would also be easy).

Whatever decision is taken, this PR is also doing some code factorization which is worth to be integrated. So, in the absence of strong opinion, the status quo would imply to merge this cleanup as it. But, @mattam82, if you wish, I can also remove the ability to write the strange things above.

herbelin:
@mattam82: I don't think this need a lot of thinking time. It is either about the status quo and merge the PR which improves over the status quo, or being stricter about binders in abbreviations and I can also implement this.

mattam82:
@herbelin Thanks for the explanations. I'm convinced by the Iris example that you want that feature for Notations, so I agree with your assessment, I'll merge when CI passes.

mattam82 (approved):

vernac/metasyntax.ml
```diff
@@ -1006,7 +1006,15 @@ let subentry_of_constr_prod_entry = function
   | ETConstr (InConstrEntry,_,_) -> InConstrEntrySomeLevel
   | _ -> InConstrEntrySomeLevel
 
-let make_interpretation_vars recvars allvars typs =
+let warn_not_interpreting_binding =
+  CWarnings.create ~name:"not-interpreting-binding-variable" ~category:"parsing"
+         (fun id ->
+          strbrk "Bound variable " ++ Id.print id ++
+          str " not bound by the notation variable.")
+
```
mattam82: This new warning should be documented.

mattam82 (dismissed):

vernac/metasyntax.ml
herbelin: Then I need help from @zimmi48 - I don't know where this documentation should go.

herbelin:
> I'm convinced by the Iris example that you want that feature for Notations

Note that Iris uses parsing-level notations for that (afaik), not abbreviations. In particular, it is unclear that use of binders in abbreviations (nametab-level notations) is actually common in the large (maybe in MathComp?).

vernac/metasyntax.ml
Zimmi48: It looks like this should be documented in the notation chapter: https://coq.inria.fr/refman/user-extensions/syntax-extensions.html#abbreviations
You should use the `.. warn::` directive to give the name of the warning.

mattam82:
@herbelin this needs a rebase apparently

mattam82:
@herbelin This should be easy to merge once rebased too.

herbelin:
I'm finally resurrecting this PR as part of an ongoing uniformization of notations and abbreviations.

I added a few words in the chapter user extensions (tell me @coq/doc-maintainers if it looks ok enough).

An overlay for elpi shall be needed.

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -589,6 +589,45 @@ the next command fails because p does not bind in the instance of n.
...
+   Check exists p, ▢_p (p >= 1).
+
+More generally, the parameter can be a pattern, as in the following
+variant:
+
+.. coqtop:: none
+
+   Reset Initial.
+
+.. coqtop:: in
```
Zimmi48, line 615: This is equivalent to the following:

```suggestion
.. coqtop:: in rest
```

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -1457,6 +1496,21 @@ Abbreviations
...
    this conventionally stops the inheritance of implicit arguments.
 
+   Like for notations, it is possible to bind binders in
+   abbreviations. Here is an example:
+
+   .. coqtop:: none
+
+      Reset Initial.
+
+   .. coqtop:: in
```
Zimmi48, line 1506:
```suggestion
   .. coqtop:: in reset
```

Zimmi48 (commented):
I am too ignorant to properly review the content of this documentation. But since you ask, I don't see what more the documentation would need to contain.

doc/sphinx/user-extensions/syntax-extensions.rst
herbelin: rest = reset ?? I would not have guessed!

herbelin: Thanks for the feedback and suggestions!

doc/sphinx/user-extensions/syntax-extensions.rst
```diff
@@ -608,11 +608,7 @@ binding position. Here is an example:
 More generally, the parameter can be a pattern, as in the following
 variant:
 
-.. coqtop:: none
-
-   Reset Initial.
-
-.. coqtop:: in
+.. coqtop:: in rest
```
Zimmi48, line 611: Looks like I made a typo in my suggestion:
```suggestion
.. coqtop:: in reset
```

herbelin: Ah, no problem.

herbelin:
May I sollicitate @ejgallego to be an alternative assignee for this PR?

Zimmi48:
@herbelin Looks like this still needs a changelog entry. I also suggest squashing a bit the documentation commits.

herbelin:
> Looks like this still needs a changelog entry. I also suggest squashing a bit the documentation commits.

Done.

Elpi overlay also rebased.

ejgallego (approved):
Indeed the semantics of capturing substitutions seems a bit surprising, but as pointed out by @herbelin this is not new. This PR seems OK.

PR #8812, opened 2018-10-24
[ssreflect] Better use of Coqlib

vbgl:
This is part of the work on stdlib2: it enables to load the `ssreflect` plug-in and to use some tactics without depending too much on the current Coq standard library and prelude.

### DIFF
--- a/library/coqlib.ml
+++ b/library/coqlib.ml
@@ -227,8 +227,7 @@ type coq_eq_data = {
 
 (* Leibniz equality on Type *)
 
-let build_eqdata_gen lib str =
-  let _ = check_required_library lib in {
+let build_eqdata_gen str = {
   eq    = lib_ref ("core." ^ str ^ ".type");
   ind   = lib_ref ("core." ^ str ^ ".ind");
   refl  = lib_ref ("core." ^ str ^ ".refl");
@@ -237,9 +236,9 @@ let build_eqdata_gen lib str =
   congr = lib_ref ("core." ^ str ^ ".congr");
   }
 
-let build_coq_eq_data       () = build_eqdata_gen logic_module_name "eq"
-let build_coq_jmeq_data     () = build_eqdata_gen jmeq_module_name  "JMeq"
-let build_coq_identity_data () = build_eqdata_gen datatypes_module_name "identity"
+let build_coq_eq_data       () = build_eqdata_gen "eq"
+let build_coq_jmeq_data     () = build_eqdata_gen "JMeq"
+let build_coq_identity_data () = build_eqdata_gen "identity"
 
 (* Inversion data... *)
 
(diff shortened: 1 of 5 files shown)


plugins/ssr/ssreflect.v
```diff
@@ -425,6 +440,9 @@ Lemma ssr_congr_arrow Plemma Pgoal : Plemma = Pgoal -> Plemma -> Pgoal.
 Proof. by move->. Qed.
 Arguments ssr_congr_arrow : clear implicits.
 
+Register nary_congruence as plugins.ssreflect.nary_congruence.
+Register ssr_congr_arrow as plugins.ssreflect.ssr_congr_arrow.
+
```
herbelin: Could you explain the technical role played by these constants?

library/coqlib.ml
```diff
@@ -227,8 +227,7 @@ type coq_eq_data = {
 
 (* Leibniz equality on Type *)
 
-let build_eqdata_gen lib str =
-  let _ = check_required_library lib in {
```
herbelin: This change was missing in the main coqlib PR right?

herbelin (commented):
Since the PR is talking about stdlib2, I'd like to seize this opportunity to ask for more details. In particular, I hardly see how stdlib2 can be something different than a collection, unification, improvement of existing contents providing standard results here and there (stdlib1, stdpp, hott, the many extensions of standard library existing here and there, e.g contents from mathcomp's `seq.v`, etc., etc.). Do you have the same view?

plugins/ssr/ssreflect.v
gares: they are used by `congr (pattern)` and `congr: hyp` respectively.

gares (approved):

library/coqlib.ml
gares: I guess it is a "bug" that prevented the rest of this PR to be tested. The whole point of the new coqlib is to be independent, the the removed code imposes a dependency on some library even if entry was rebound.

plugins/ssr/ssreflect.v
vbgl: May I suggest to discuss the general design of the ssreflect library in a different place (e.g., the [ssreflect mailing list](https://sympa.inria.fr/sympa/arc/ssreflect)), so as not to hijack this PR? Thanks.

herbelin: @vbgl: I don't think my question is out of topic. `congr` and `f_equal` are among the tactics which make little sense to have in double, so I think it is an important question to understand how to design a tactic which subsumes the advantage of both of them.

Discussions on ssreflect are most welcome on the media provided by this repository (which includes ssreflect). Why would you like the discussion to be elsewhere?

gares: Please ask here #8840 and I'll reply

PR #8815, opened 2018-10-24
NArith: add lemmas about numbers and vectors

liyishuai:
**Kind:** feature.

### DIFF
--- a/theories/NArith/BinNat.v
+++ b/theories/NArith/BinNat.v
@@ -422,6 +422,25 @@ Proof.
  unfold lt in *; simpl in *. now rewrite Pos.compare_xI_xO, H.
 Qed.
 
+Lemma double_lt_mono n m : n < m -> double n < double m.
+Proof.
+  destruct n as [|n], m as [|m]; intros H; try easy.
+Qed.
+
+Lemma double_le_mono n m : n <= m -> double n <= double m.
+Proof.
+  destruct n as [|n], m as [|m]; intros H; try easy.
+Qed.
+
+Lemma succ_double_lt_mono n m : n < m -> succ_double n < succ_double m.
+Proof.
+  destruct n as [|n], m as [|m]; intros H; try easy.
+Qed.
+
+Lemma succ_double_le_mono n m : n <= m -> succ_double n <= succ_double m.
+Proof.
+  destruct n as [|n], m as [|m]; intros H; try easy.
+Qed.
 
 (** 0 is the least natural number *)
 
--- a/theories/NArith/Ndigits.v
+++ b/theories/NArith/Ndigits.v
@@ -546,8 +546,10 @@ Fixpoint Bv2N (n:nat)(bv:Bvector n) : N :=
     | Vector.cons _ true n bv => N.succ_double (Bv2N n bv)
   end.
 
+Arguments Bv2N {n} bv, n bv.
+
 Definition ByteV2N {n : nat} : ByteVector n -> N :=
-  Bv2N (n * 8) ∘ to_Bvector.
+  Bv2N ∘ to_Bvector.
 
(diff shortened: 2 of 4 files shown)


theories/NArith/Ndigits.v
```diff
@@ -534,13 +536,19 @@ Definition N2Bv_sized (m : nat) (n : N) : Bvector m :=
...
+  of_Bvector ∘ N2Bv_sized (m * 8).
+
 Fixpoint Bv2N (n:nat)(bv:Bvector n) : N :=
   match bv with
     | Vector.nil _ => N0
     | Vector.cons _ false n bv => N.double (Bv2N n bv)
     | Vector.cons _ true n bv => N.succ_double (Bv2N n bv)
   end.
 
+Definition ByteV2N {n : nat} : ByteVector n -> N :=
```
liyishuai: I think `Bv2N` should also make `n : nat` argument implicit.
There could be another `Bv2N_sized {n : nat} (m : nat) : Bvector n -> N` that converts the first (*i.e.* lowest) `m` bits of given vector. I see no reason for providing `n` explicitly.

theories/NArith/BinNat.v
```diff
@@ -422,6 +422,25 @@ Proof.
...
+
+Lemma succ_double_lt_mono n m : n < m -> succ_double n < succ_double m.
+Proof.
+  destruct n as [|n], m as [|m]; intros H; try easy.
+Qed.
+
+Lemma succ_double_le_mono n m : n <= m -> succ_double n <= succ_double m.
+Proof.
+  destruct n as [|n], m as [|m]; intros H; try easy.
+Qed.
```
herbelin: With respect to naming, you apparently mix the model of `BinNat.succ_double_lt` and of `BinPos.succ_lt_mono`.

I don't have a good solution (even more if we also include the models `Ndec.Nleb_double_plus_one_mono`, the Frenchism `Zpow_facts.Zpower_le_monotone`, as well as `Rfunctions.Power_monotonic` and `Zorder.Zsucc_lt_compat`).

Personally, I find the suffix `_mono` a bit ambiguous. It could refer to "monotonic", but also e.g. to "monomorphic". On the other side, the full `monotonic` is long. So, somehow, I would avoid adding a new `_mono`, and stay with names such as `double_lt`, similar to what we already have in `BinNat`.

theories/NArith/Ndigits.v
herbelin: To preserve compatibility, you can give:
```coq
Arguments Bv2N [n] bv, n bv.
```
so that the first argument can freely be made implicit or explicit.

theories/Strings/ByteVector.v
```diff
@@ -8,7 +8,7 @@
 (*         *     (see LICENSE file for the text of the license)         *)
 (************************************************************************)
 
-Require Import Ascii Basics Bvector Psatz String Vector.
+Require Import Ascii Basics Bvector String Vector.
```
herbelin: This is unrelated?

herbelin (commented):
Looks good to me. See comments.

theories/Strings/ByteVector.v
liyishuai: Right. I shouldn’t have introduced this dependency.

theories/NArith/BinNat.v
liyishuai: I was trying to remove `_mono` from `succ_double_lt_mono` but saw this:
https://github.com/coq/coq/blob/fb56c4eb9cb4afe4365c05e0cc6ba2643064ee6b/theories/NArith/BinNat.v#L419

herbelin:
OK, let's keep `mono` in `succ_double_lt_mono`. Shall merge on Monday.

PR #8817, opened 2018-10-25
SProp: the definitionally proof irrelevant universe

SkySkimmer:
See also https://github.com/coq/ceps/pull/37

Overlay PRs:
https://github.com/coq-community/aac-tactics/pull/28
https://github.com/Karmaki/coq-dpdgraph/pull/56
https://github.com/lukaszcz/coqhammer/pull/12
https://github.com/LPCIC/coq-elpi/pull/45
https://github.com/mattam82/Coq-Equations/pull/161 (this one is fairly big)
https://github.com/ppedrot/ltac2/pull/93
https://github.com/unicoq/unicoq/pull/23
https://github.com/Mtac2/Mtac2/pull/136 (NB: they used SProp as an identifier so I had to rename it)
https://github.com/coq-community/paramcoq/pull/18
https://github.com/ybertot/plugin_tutorials/pull/21
https://github.com/QuickChick/QuickChick/pull/134
https://github.com/damien-pous/relation-algebra/pull/4


toplevel/coqtop.ml
```diff
@@ -456,6 +456,9 @@ let init_toplevel custom_init arglist =
       Flags.if_verbose print_header ();
       Mltop.init_known_plugins ();
       Global.set_engagement opts.impredicative_set;
+      Global.set_uip opts.uip;
+      (Global.set_allow_sprop opts.allow_sprop);
+      (if opts.cumulative_sprop then Global.make_sprop_cumulative ());
```
ejgallego: It would be cool if we could aggregate all the consistency-related options into a single record, right?

ejgallego (approved):
Changes in `clib/toplevel/coq_dune` are fine.

ejgallego:
@SkySkimmer have you tried seeing how much this patch loses when `-flambda` is enabled?

There is a possibility that the impact could be diminished.

Unfortunately you need to patch Coq so the right `-O3 -unbox-closures` are passed to Coq as the bench doesn't allow setting configure flags. Use 4.07.1+flambda.

SkySkimmer:
Added overlays and listed corresponding PRs in the PR header comment.

SkySkimmer:
Pushed some dev doc and some barebones refman doc.

doc/sphinx/language/cic.rst
```diff
@@ -336,14 +355,15 @@ following rules.
...
+   s \in \{\SProp, \Prop, \Set\}
    \WTE{\Gamma::(x:T)}{U}{\Set}
    ----------------------------
    \WTEG{\forall~x:T,U}{\Set}
 
 .. inference:: Prod-Type
 
-   \WTEG{T}{\Type(i)}
+   \WTEG{T}{s}
+   s \in \{\SProp, \Type{i}\}
```
Zimmi48: Did you really mean this change? Isn't it implied by universe cumulativity from the rule above as in the case of `Prop`?

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -218,13 +218,17 @@ numbers (see :ref:`datatypes`).
...
    single: Prop
    single: Type
 
 Sorts
 -----
 
-There are three sorts :g:`Set`, :g:`Prop` and :g:`Type`.
+There are four sorts :g:`SProp`, :g:`Prop`, :g:`Set`  and :g:`Type`.
+
+-  :g:`SProp`_is the universe of *definitionally irrelevant
```
Zimmi48:
```suggestion
-  :g:`SProp` is the universe of *definitionally irrelevant
```

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -235,7 +239,7 @@ There are three sorts :g:`Set`, :g:`Prop` and :g:`Type`.
    specifications by :production:`specif`. This constitutes a semantic subclass of
    the syntactic class :token:`term`.
 
--  :g:`Type` is the type of :g:`Prop` and :g:`Set`
+-  :g:`Type` is the type of the other sorts.
```
Zimmi48:
```suggestion
-  :g:`Type` is the type of sorts.
```

`Type` being also the type of `Type`...

dev/doc/SProp.md
```diff
@@ -0,0 +1,71 @@
...
+change is needed, eg if you have `match foo with Lambda (x, a, b) ->
+Prod (x, a, type_of (push_rel (LocalAssum (x,a)) env) b)`. Otherwise
+see `context.mli` for a few combinators on the `binder_annot` type.
+
+As a plugin you can generally just use `Relevant` everywhere, the
+kernel will fix it if needed when it checks the terms you produce. The
+only issue is that if you generate `Relevant` when it should have been
+`Irrelevant` you won't be able to use proof irrelevance on that
+variable until the kernel fixes it. See refman for examples as Coq
+also uses `Relevant` incorrectly in some places.
```
Zimmi48: I suppose this is meant to be added later.

dev/doc/SProp.md
```diff
@@ -0,0 +1,71 @@
...
+kernel is not (yet?) able to repair incorrect case annotations so your
+code may fail for `SProp` users.
+
+In general you should probably be using
+`Inductiveops.make_case_or_project` instead of producing the `is`
+annotation yourself, this will also handle primitive records for you.
+You still need to pass it the right `case_info` and so the right
+`ci_relevance` though.
+
+NB: make_case_or_project is already passed an env, maybe we should
```
Zimmi48:
```suggestion
NB: `make_case_or_project` is already passed an env, maybe we should
```

dev/doc/SProp.md
```diff
@@ -0,0 +1,71 @@
...
+have it retype to get `ci_relevance`. However currently the env is
+only for globals so this will break those who call it with incorrect
+rel_context (IIRC cases calls with incorrect rel_context).
```
Zimmi48: I assume this "IIRC" is a todo.

Zimmi48 (commented):
Obviously the doc would be better with examples but it looks like you have already planned to add some.

doc/sphinx/language/gallina-specification-language.rst
Zimmi48: This is what made the refman build fail BTW.

doc/sphinx/language/cic.rst
SkySkimmer: No cumulativity with SProp ;)

doc/sphinx/language/gallina-specification-language.rst
SkySkimmer: Oh I see, emacs displays the non breaking space (used below for Prop etc) as an underscore so I typed _ assuming sphinx would translate it.

dev/doc/SProp.md
SkySkimmer: The entire paragraph is a todo.

SkySkimmer: ie something to consider after the PR is merged

SkySkimmer:
>How much work was it preparing overlays for all the plugins (and opening the PRs, etc.).

It was OK. I expected to have to do more changes (except for the equations overlay which was the size I expected for all of them). I basically did equations today and the others during the WG.
Opening PRs is trivial (copy my fork repo's url from the overlay file, visit it in browser, click open PR, select branch, type in "please wait before merging", push button).

SkySkimmer:
As it feels increasingly unready to me I removed the implementation of uip in SProp.
The code is still available in the sprop-uip branch on my repo.

SkySkimmer:
>As it feels increasingly unready to me I removed the implementation of uip in SProp.

Without uip, case inversion doesn't provide much value. It would probably be simpler/less risky to implement outside the kernel (and it's possible, which is why it doesn't provide much value).
Also I don't know how to implement it for the VM/native and it's not implemented for the non kernel reductions either.
Considering the above I'm going to drop it from this PR.

dev/doc/SProp.md
```diff
@@ -0,0 +1,41 @@
+# Notes on SProp
+
+(ml API side, see refman for user side)
+
+## Relevance
+
+All kernel binders (`Prod`/`Lambda`/`LetIn`/`Context` elements) are
+now annotated with a value in `type Sorts.relevance = Relevant |
+Irrelevant`. It should verify that the binder's type lives in `SProp`
+iff the annotation is `Irrelevant`.
```
gares: It is not clear how a "value" could "verify". I guess you want to say that the kernel checks that.

Zimmi48 (commented):
I'm not quite sure what is causing the Sphinx timeout.

doc/sphinx/addendum/sprop.rst
```diff
@@ -0,0 +1,192 @@
...
+=====================================
+
+.. warning::
+
+   The status of strict propositions is experimental.
+
+This section describes the extension of |Coq| with definitionally
+proof irrelevant propositions (types in the sort :math:`\SProp`, also
+known as strict propositions). To use :math:`\SProp` you must pass
+``-allow-sprop`` to the |Coq| program.
```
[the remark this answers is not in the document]
SkySkimmer: I followed -impredicative-set and -indices-matter

SkySkimmer:
>I'm not quite sure what is causing the Sphinx timeout.

I guess it doesn't like coq comments

doc/sphinx/addendum/sprop.rst
```diff
@@ -0,0 +1,192 @@
+.. _sprop:
+
+SProp (proof irrelevant propositions)
+=====================================
```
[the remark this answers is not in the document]
SkySkimmer: >the exponential growth in the number of chapters

There's no such thing.

theories/Logic/StrictProp.v
```diff
@@ -0,0 +1,34 @@
...
+Record Box (A:SProp) : Prop := box { unbox : A }.
+Arguments unbox {_} _.
+
+Inductive Squash (A:Type) : SProp := squash : A -> Squash A.
+Arguments squash {_} _.
+
+Inductive sEmpty : SProp :=.
+Inductive sUnit : SProp := stt.
+
+Set Primitive Projections.
```
JasonGross: Why does `Box` not get primitive projections?

SkySkimmer: See https://github.com/coq/coq/pull/8817/files#diff-f68e219fa589e1ab7aa651953a417aafR107 ish

SkySkimmer:
How do I get sphinx to build sprop.rst?

doc/sphinx/addendum/sprop.rst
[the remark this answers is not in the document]
Zimmi48: What I meant is that we have too many chapters *already*, thus I'm very suspicious of any addition.

theories/Logic/StrictProp.v
JasonGross: Thanks!

Zimmi48:
>How do I get sphinx to build sprop.rst?

It needs to be listed in `index.*.rst`.

doc/sphinx/addendum/sprop.rst
```diff
@@ -0,0 +1,197 @@
...
+i.e. they can only be eliminated to construct proofs of other strict
+propositions. The exceptions are empty types and unit types.
+
+.. coqtop::
+
+   Inductive sEmpty : SProp := .
+   Check sEmpty_rect.
+
+   Inductive sUnit : SProp := stt.
+   Check sUnit_rect.
```
Zimmi48: This is a case where the reader really needs to see the output of `coqtop`. Note that I'm surprised that the output is not shown by default when not passing an option. Anyways, changing `.. coqtop::` into `.. coqtop:: all` should fix this.

doc/sphinx/addendum/sprop.rst
```diff
@@ -0,0 +1,197 @@
...
+would disallow instantiating it by ``sUnit : SProp``.
+
+In order to make the system usable without having to annotate every
+instance of :math:`\SProp`, we consider :math:`\SProp` to be a subtype
+of every universe during elaboration (i.e. outside the kernel). Then
+once we have a fully elaborated term it is sent to the kernel which
+will check that we didn't actually need cumulativity of :math:`\SProp`
+(in the example above, ``u`` doesn't appear in the final term).
+
+This means that some errors will be delayed until ``Qed``:
```
Zimmi48: There is no equivalent of `Guarded` then to use the kernel in the middle of a proof?

Zimmi48 (commented):
Good work on adding an `Issues` section and documenting the developer-only warning. This chapter is well written and structured, so OK for keeping it a separate chapter, at least until the all reference manual is refactored.

doc/sphinx/addendum/sprop.rst
```diff
@@ -0,0 +1,197 @@
...
+  Definition isquash A : A -> iSquash A
+    := fun a P f => f a.
+  Definition iSquash_sind A (P : iSquash A -> SProp) (H : forall x : A, P (isquash A x))
+    : forall x : iSquash A, P x
+    := fun x => x (P x) (H : A -> P x).
+
+.. note::
+
+   Eliminators to strict propositions are called ``foo_sind``, in the
+   same way that eliminators to propositions are called ``foo_ind``.
```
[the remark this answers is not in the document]
SkySkimmer: >Shouldn't this note appear later in the text?

Maybe. I put it there to explain why I was calling the example `iSquash_sind` but I guess it's not critical so can be put later.

doc/sphinx/addendum/sprop.rst
SkySkimmer: >There is no equivalent of Guarded then to use the kernel in the middle of a proof?

No. Maybe someday ^^

doc/sphinx/addendum/sprop.rst
SkySkimmer: @Zimmi48 I ended up removing relevant matches for irrelevant unit types (and the corresponding doc bit), is this comment still relevant?

doc/sphinx/addendum/sprop.rst
```diff
@@ -0,0 +1,197 @@
...
+----------------
+
+The purpose of :math:`\SProp` is to provide types where all elements
+are convertible:
+
+.. coqtop::
+
+   Definition irrelevance (A:SProp) (P:A -> Prop) (x:A) (v:P x) (y:A) : P y := v.
+
+Since we have definitional η-expansion :ref:`eta-expansion` for
```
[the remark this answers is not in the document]
SkySkimmer: Changed.

doc/sphinx/addendum/sprop.rst
mattam82: You mean you removed elimination from sUnit to Type?

SkySkimmer: >You mean you removed elimination from sUnit to Type?

Yes. It would need to always reduce (so needing so special handling in reduction) and it's useless (the identity function works fine) so I just removed it.

doc/sphinx/addendum/sprop.rst
Zimmi48: But wouldn't it be easy to add it?

doc/sphinx/addendum/sprop.rst
Zimmi48: >is this comment still relevant?

I suppose not, but if you never intend to show the output of `.. coqtop::` you should not waste computational resources and use `.. coqdoc::` instead.

SkySkimmer:
I messed with `coqtop all`/`coqdoc` some.

doc/sphinx/addendum/sprop.rst
SkySkimmer: >But wouldn't it be easy to add it?

Not so easy, or at least I don't see it. PRs welcome ;)

doc/sphinx/addendum/sprop.rst
```diff
@@ -0,0 +1,225 @@
...
+   Allows using :math:`\SProp` when set and forbids it when unset. The
+   initial value depends on whether you used the command line
+   ``-allow-sprop``.
+
+.. coqtop:: none
+
+   Set Allow SProp.
+
+Some of the definitions described in this document are available
+through ``Coq.Logic.StrictProp``, which see.
```
Zimmi48: I'm not sure I understand what "which see" means. Is that an unfinished sentence?

Zimmi48 (approved):

doc/sphinx/addendum/sprop.rst
SkySkimmer: It means "cf the thing I just mentioned"

SkySkimmer: If we could make the module name a link to the stdlib doc that would work, but I'm not sure how feasible that is.

SkySkimmer: (it seems to be used significantly by emacs lisp and C people, see language tab in https://github.com/search?q=%22which+see%22&type=Code)

SkySkimmer:
During the WG there was some worry about the elaboration mode accepting cumulativity.

I just added an option `Elaboration StrictProp Cumulativity` which you can disable to easily check what happens without it.

The StrictProp.v stdlib file and sprop.v test suite file are relatively heavily annotated on binders so you don't see much of what the doc says, but you get universe inconsistencies when you want to make a match with SProp return type. I have no clue why.

In the future we may try to handle non cumulative universes in a general way somehow, but I don't know how to do this with the graph setup and I don't want to delay this PR for it.

Can I get some reviews? @mattam82 @maximedenes @herbelin

ppedrot:
@SkySkimmer Something looks broken with dpdgraph for some reason, can you have a look?

@coq/core I think this should be merged as swiftly as possible into the master branch, so as to land in 8.10. There is probably a bit of polishing left here and there, but primitive integers also made it similarly, and this kind of feature needs to be more largely tested. The very small performance hit is likely to be more than largely hidden by the major speedup from reworking of pattern-matching type-checking, so I am not very wary there. Thus, given the lack of reaction from most people, I think **I'll merge this PR** (assuming the CI is fixed, probably a stupid reason) **one week from now**. Anybody willing to further review or opposing this merge is asked to **voice up now**.

kernel/indTyping.ml
```diff
@@ -122,73 +122,103 @@ let check_cumulativity univs variances env_ar params data =
...
+      (* relevant records must have at least 1 relevant argument *)
+      && (Univ.Universe.is_sprop info.ind_univ
+          || info.ind_has_relevant_arg)
+      && (match splayed_lc with
+          (* records must have 1 constructor with at least 1 argument, and no anonymous fields *)
+          | [|ctx,_|] ->
+            let module D = Context.Rel.Declaration in
+            List.exists D.is_local_assum ctx &&
+            List.for_all (fun d -> not (D.is_local_assum d)
+                                   || not (Name.is_anonymous (D.get_name d)))
```
maximedenes: Why enforce this here? Is it a new restriction?

kernel/indtypes.ml
```diff
@@ -458,14 +459,15 @@ let compute_projections (kn, i as ind) mib =
         (* from [params, x:I, field1,..,fieldj |- t(field1,..,fieldj)]
            to [params, x:I |- t(proj1 x,..,projj x)] *)
 	let fterm = mkProj (Projection.make kn false, mkRel 1) in
-        (i + 1, j + 1, lab :: labs, projty :: pbs, fterm :: letsubst)
-      | Anonymous -> raise UndefinableExpansion
+        (i + 1, j + 1, lab :: labs, r :: rs, projty :: pbs, fterm :: letsubst)
+      | Anonymous -> assert false (* checked by indTyping *)
```
maximedenes: I guess that answers my question above :)

kernel/declarations.ml
```diff
@@ -176,6 +177,8 @@ type one_inductive_body = {
 
     mind_recargs : wf_paths; (** Signature of recursive arguments in the constructors *)
 
+    mind_relevant : Sorts.relevance;
```
maximedenes: Why not `mind_relevance` to be uniform with`const_relevance`?

kernel/indTyping.ml
```diff
@@ -268,18 +300,38 @@ let typecheck_inductive env (mie:mutual_inductive_entry) =
...
 
+  let record = mie.mind_entry_record in
+  let data, record = match record with
+    | None | Some None -> data, record
+    | Some (Some _) ->
+      if check_record data then
+        data, record
+      else
+        (* if someone tried to declare a record as SProp but it can't
+           be primitive we must squash. *)
```
SkySkimmer: >Why enforce this here? Is it a new restriction?

Because of this bit we need to do the record downgrade earlier than previously.

ejgallego (approved):
`interp` / `parsing` / `plugin_tutorial` / `vernac` seem OK.

maximedenes (approved):
I'm ok with merging and refining later. One potential worry, though: this obviously introduces a path in conversion that could be dangerous (conversion errors are ignored for irrelevant things). So I was wondering: is none of what is done in `CClosure` relying on invariants that we know don't always hold? 

Specifically, I'm thinking about the fact that in practice we sometimes test conversion for terms that don't have convertible types. Is the handling of irrelevance marks in the reduction machine robust to that?

gares (dismissed):
I've listed a few minor changes that I guess can be easily fixed.

My main concern regarding this PR is the where the relevance info is placed in the terms. While it is clear that putting it there minimizes breakage in existing code, it seems error prone to me.
In particular a subterm that was typically ignored (a name, used only for pretty printing) now holds a logically relevant bit of information. The kernel may have been fixed so that it is not ignored by mistake, but I can hardly imagine the entire system was inspected (no compilation failure, no problem). I'm not sure which kind of bugs could arise from that. Eg a term with relevance annotations would compare equal to a term without (since eq tests typically skip these names), so these annotations seem a bit volatile to me.

I don't think this should be fixed now (before the merge, and if we decide to do so), but I'd like to point it out. Maybe there is no reason to be afraid of that, or maybe it is more severe than what I think...
I hope others have a clear picture about this.

gares:
I've a question about the behavior of ssr `=> +` now. The implementation does `Ssrcommon.tclINTRO` that introduces an hyp and saves his name (the name that was in the `Prod`  in the goal). Then at the end of the line `Ssripats.gen_astac` is called to revert back the hyp using the same name. The relevance mark on the original product in goal is lost (it seems the PR always puts `Relevant` when building a `Prod`).
What is the impact of that? The semantics of `=> +` is that it is a no-op. Is it really?
I also imagine the same happens with `let x := fresh "x" in intro x; revert x`, the new `Prod` is relevant no matter what the old one was. It this the case?

SkySkimmer:
Regular intros;revert should work. ssreflect does look heavy on the dummy Relevant. 
The likely involved mkProd are https://github.com/coq/coq/blob/a83eac8463787c13a2dbd3903baf2b59ca1a4635/plugins/ssr/ssrcommon.ml#L741-L744 and https://github.com/coq/coq/blob/a83eac8463787c13a2dbd3903baf2b59ca1a4635/plugins/ssr/ssrcommon.ml#L1132 which both get the domain by typing a term so we could just ask the relevance of the term.

You can test with something like 
~~~coq
(* -*- coq-prog-args: ("-allow-sprop"); -*- *)

Set Warnings "+bad-relevance".

(* example non tactic fail *)
Fail Definition id (A:SProp) x : A := x.

(* intros;revert is fine *)
Definition id : forall (A:SProp) (x:A), A.
Proof.
  intros A x;revert x;intros x;exact x.
Defined.

(* example tactic fail *)
Definition id' (A:SProp) (x:A) : A.
Proof.
  pose (y:=x). exact y.
  Fail Qed.
Abort.
~~~
I don't know what the ssr syntax would be, so I can't test if the proposed fix above works or even if it's needed.

SkySkimmer:
>Specifically, I'm thinking about the fact that in practice we sometimes test conversion for terms that don't have convertible types. 

Certainly code which asks for convertibility of irrelevant terms in different types is incorrect, eg
~~~coq
(* -*- coq-prog-args: ("-allow-sprop"); -*- *)

Inductive one : SProp := One.
Inductive two : SProp := Two.

Goal forall P, P One -> P One.
Proof.
  intros.
  set Two. (* t := Two |- P t *)
~~~
This should only be a concern for non kernel stuff though.

kernel/declarations.ml
SkySkimmer: Changed

kernel/context.ml
```diff
@@ -180,7 +201,7 @@ struct
 
   (** [extended_rel_list n Γ] builds an instance [args] such that [Γ,Δ ⊢ args:Γ]
       with n = |Δ| and with the local definitions of [Γ] skipped in
-      [args]. Example: for [x:T,y:=c,z:U] and [n]=2, it gives [Rel 5, Rel 3]. *)
+      [args]. Example: for [x:T,y:=c,z:U] and [n]=2, it gives [Rel 5el 3]. *)
```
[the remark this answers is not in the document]
SkySkimmer: Fixed

kernel/indTyping.ml
```diff
@@ -122,73 +122,103 @@ let check_cumulativity univs variances env_ar params data =
...
-
-let check_constructors env_ar_par params lc (arity,indices,univ_info) =
+  check_context_univs ~ctor:true env_ar_par info args
+
+let check_constructors env_ar_par isrecord params lc (arity,indices,univ_info) =
   let lc = Array.map_of_list (fun c -> (Typeops.infer_type env_ar_par c).utj_val) lc in
   let splayed_lc = Array.map (Reduction.dest_prod_assum env_ar_par) lc in
-  let univ_info = if Array.length lc <= 1 then univ_info
-    else check_univ_leq env_ar_par Univ.Universe.type0 univ_info
+  let univ_info = match lc with
```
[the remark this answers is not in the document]
SkySkimmer: Changed.

kernel/cooking.ml
```diff
@@ -159,6 +160,7 @@ type result = {
   cook_type : types;
   cook_universes : universes;
   cook_private_univs : Univ.ContextSet.t option;
+  cook_relevance : Sorts.relevance;
```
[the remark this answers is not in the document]
SkySkimmer: It's just passed through, there's nothing interesting about it. Maybe we could have a comment for the whole structure but not sure what would be the comment.

kernel/nativelambda.ml
```diff
@@ -518,8 +519,10 @@ let rec lambda_of_constr cache env sigma c =
 	else 
 	  match b with
 	  | Llam(ids, body) when Int.equal (Array.length ids) arity -> (cn, ids, body)
-	  | _ -> 
-	      let ids = Array.make arity Anonymous in
+          | _ ->
+              (** TODO relevance *)
```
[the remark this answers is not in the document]
SkySkimmer: >If native compute ignores SProp, then it should be documented

Indeed, I added a `.. warning::` to sprop.rst 

I'm not sure what the impact of this TODO relevance (and the 2 in clambda) is though. It looks like it's about eta-expanding contructors, and may be producing dummy stuff? ie
~~~coq
(* -*- coq-prog-args: ("-allow-sprop"); -*- *)

Set Warnings "+bad-relevance".

Inductive foo := C : (forall P:SProp, P) -> foo.

Definition c := Eval vm_compute in C.
Definition c' := Eval native_compute in C.

Print c.
Print c'. (* both: fun x => C x *)
~~~
this works without throwing a bad-relevance error. So I'm not sure where the produced dummy Relevant are going.
If I try vm/native on foo_rect or foo_sind it also works without bad relevance warnings.
Maybe they're just for conversion and don't matter when just reducing, in which case using dummies will only matter once we implement proof irrelevance in vm/native_conv.

ppedrot:
> This should only be a concern for non kernel stuff though.

Judging from the mess that kernel conversion is today, I wouldn't bet my life on that though.

gares:
It would be
```coq
Goal forall T : SProp, T -> True.
move=> T +.
```

Just to be clear: is it true that the current patch fixed any `mkProd/Lam/Let` by putting `Relevant`?
Do we have a CI job running with `+bad-relevance`?

gares:
> get the domain by typing a term so we could just ask the relevance of the term

yes please fix it, these are bugs to me (and the cost has 0 computational cost).

SkySkimmer:
>It would be

Fixed

>Just to be clear: is it true that the current patch fixed any mkProd/Lam/Let by putting Relevant?

For some of them I tried to be correct (also for some Relevant is correct, eg the equality schemes are invalid on irrelevant things since equality can't be applied to irrelevant things).

>Do we have a CI job running with +bad-relevance?

That doesn't really make sense until we know how to do the classical bad-relevance `Fail Definition id (A:SProp) x : A := x.` without post-hoc relevance repair. `+bad-relevance` is a debug warning for ocaml writers.

gares:
> That doesn't really make sense until we know how to do the classical bad-relevance `Fail Definition id (A:SProp) x : A := x.` without post-hoc relevance repair. `+bad-relevance` is a debug warning for ocaml writers.

I don't get what you mean. From your previous post I got the impression that `+bad-relevance` would detect, at run time, code that inserts `Relevant` by "mistake".

> Fixed

thanks

SkySkimmer:
>I don't get what you mean. From your previous post I got the impression that +bad-relevance would detect, at run time, code that inserts Relevant by "mistake".

We can't make it an error when it's intended to happen.

SkySkimmer:
Maybe I should add an "unknown" relevance to be able to differentiate between dummy Relevant and actually-intended Relevant. Then the kernel would error on incorrect Relevant and fix-with-debug-warning Unknown.

ppedrot:
> Then the kernel would error on incorrect Relevant and fix-with-debug-warning Unknown.

Please don't do that. That reminds me of the hacks for universe polymorphism to work around not fully ported plugins. Let's not make *once again* the kernel uglier to enhance the transition for our end users, we end up paying the interests of these technical debts Troika-style.

SkySkimmer:
>Please don't do that. That reminds me of the hacks for universe polymorphism to work around not fully ported plugins. Let's not make once again the kernel uglier to enhance the transition for our end users, we end up paying the interests of these technical debts Troika-style.

I already made the kernel "uglier" though

SkySkimmer:
also it's not about incomplete porting, see doc

Gaëtan Gilbert

gares:
> Please don't do that. That reminds me of the hacks for universe polymorphism

I don't think it would be worse, on the contrary. Let me explain and compare with polyuniv.

In polyuniv code that did not handle the new feature was supposed to do anyway and the kernel (even if it is still capable to repair (infer)) was just bailing out. The result is that users got proofs rejected by Qed. This is bad, very bad, it is blocking.

In this case the kernel does infer (thanks god) so putting "wrong" relevance values at proof construction time is not necessarily going to fail users. The fact that the kernel emits the warning only when the tactic layer put `Dummy`, rather than `Relevant` or `Irrelevant`, seems just a way to make the warning more precise. In particular the kernel would behave as follows:

1. `Dummy` means there is code to port, a TODO, and the user is calling it. I (kernel) can infer so nevermind. I can emit a warning to that developers can track this TODO on CI.
2. I (kernel) get a wrong `Relevant`/`Irrelevant` annotation. It means that there is a real bug in the tactic layer and should be fixed. In release mode, I can still accept that, and fix the relevance. In devel mode (or in CI) I should give an error.

When the feature matures, eg when `Dummy` does not exists anymore, we may decide to turn 2 into an error in release mode too (as polyuniv).
The *big* difference is that users get a working system from day one and don't need to report a ton of "tactic X fails with Set Univ Poly, I'm stuck!".
Of course we have to commit to get eventually rid of all `Dummy`.

ppedrot:
> This is bad, very bad, it is blocking

I am not convinced by this argument. SProp is experimental stuff, you should expect polishing issues when opting in. It would be another matter if this branch were to break non-user workflow, but thankfully it's not the case. (I should have written hopefully, but a bit of optimism doesn't hurt.)

> Of course we have to commit to get rid of all Dummy, eventually.

Everything is in the adverb, I'm afraid. Obviously at the end of the thermodynamical universe nothing will matter much, but in the meantime anybody trying to set up stronger invariants in the kernel is going to have a bad time.

SkySkimmer:
>I would like to be reasonably convinced that we are not introducing some unsoundness in the implementation. @SkySkimmer can you be more specific about why this comparison at non-convertible types is not a problem in kernel mode ?

The kernel should only call conversion on terms at the same type.

maximedenes:
> The kernel should only call conversion on terms at the same type.

Is it true today, though?

maximedenes:
> Is it true today, though?

I mean, it is probably true at toplevel, but does not seem guaranteed inside `CClosure` (due to the way it works on `App` and `Case` nodes).

SkySkimmer:
Up to some commutativity I think so
(ie `iter_left` and `iter_right` are the same if the only effect is raising NotConvertible)

gares:
> I am not convinced by this argument. SProp is experimental stuff, you should expect polishing issues when opting in.

The point is: you opt in. Then you call tactic x that is not ported. You are blocked. What do you do? You opt out (and report a bug). Time to wait for Coq vN+1 before enabling SProp again.

If the kernel was not doing inference I could understand your point. But here the code repairing the error is there already.

In any case, you do what you want. I was just suggesting a strategy to make people adopt SProp faster.
If you want the default to be off forever, just do as we did for polymorphic universes.

ppedrot:
> In any case, you do what you want. I was just suggesting a strategy to make people adopt SProp faster.

I'm mostly fine in practice with merging the PR as-is, I just love complaining. But Coq is definitely getting dirtier exponentially faster than we can clean it up.

mattam82:
Please leave me some time to do a review before merging. I hereby promise to do it by tuesday (I have Nicolas to remind me :).

mattam82:
> My main concern regarding this PR is the where the relevance info is placed in the terms. While it is clear that putting it there minimizes breakage in existing code, it seems error prone to me.
> In particular a subterm that was typically ignored (a name, used only for pretty printing) now holds a logically relevant bit of information. The kernel may have been fixed so that it is not ignored by mistake, but I can hardly imagine the entire system was inspected (no compilation failure, no problem). I'm not sure which kind of bugs could arise from that. Eg a term with relevance annotations would compare equal to a term without (since eq tests typically skip these names), so these annotations seem a bit volatile to me.

That's a good point, on the other hand it is quite likely that we will extend these annotations in the future with more logically relevant information (e.g. CIC* would require something similar for binders). And we could already easily pass "shallow" implicit status information there (so that we do not have to keep track of it externally), which would mix logical/non-logical status even further. Do you suggest splitting these annotations to a separate component of binders/context? In the particular case of relevance, thankfully it is uniquely determined by the type component of the binding, so it should not be too bad even when using imprecise equality tests outside the kernel.

There's also the related question of type-inference/unification. To allow setting the relevance mark correctly during unification/elaboration, we will need a new kind of metavariable for the annotation's relevance. E.g. something like

`Lambda of meta_binder_annot * types * types`
`meta_binder_annot = { binder_name : Name.t; binder_relevance: meta_relevance }`
where a `meta_relevance` can be an unknown relevance or an actual relevance. We would then have a set of constraints `?relevance_i = relevance_of(?Xi)` to solve, when `?Xi` is defined.

So I expect that we'll need to make `constr` ultimately polymorphic over the binder annotation structure to handle this without too much duplication of code. We could at the same time make the kernel polymorphic over the actually irrelevant `Name.t` in the binders. I don't think it's necessary to go this route yet but that's certainly some possibilities to keep in mind for further refinements.

The main issue I see with the current PR is all the places where the default `Relevant` annotations were put, but we can gradually improve the situation after the merge (which is long due, sorry @SkySkimmer !). About the warning / fixing of relevance annotations: If a user switches to SProp today, and uses a non-ported tactic which happens to work using the kernel-side fixing of annotations, 
then it won't be blocking, right? We as developers can enable the warning or turn it into an error to ensure that higher-level code gets ported, which is a bit less precise that Unknown but still workable I believe. Using Unknown doesn't really change the deal for users, unless we also make high-level code like unification repair/infer marks on the fly as well, right?

So I wouldn't go the Unknown way in the kernel but rather fix the root of the problem for code outside the kernel, which I think requires introducing Sort metavariables at the level of EConstr, that can be instantiated with SProp/Prop/Set/Type i later. E.g, `new_type_evar` (or `new_sort/univ_evar` to be more precise) would produce an existential `?Xi : Gamma |- Sort ?s` where `?s` could be later determined. Together with `meta_relevance`, higher-level features like the refiner could use this to handle uniformly constructions that can fall in `SProp/Prop/Set/Type i` due to typing constraints coming later. After discussion with @tabareau , it seems that we might get quite a handful of other universes to play with in the future, so better prepare for that as well while we're at it.

mattam82 (approved):
@maximedenes mentioned this:

> On the other hand, I am more concerned by the previous discussion on the correctness of the approach in the absence of some invariants.

Where is the previous discussion? Unless there is an issue there I'm fine with merging.

SkySkimmer:
>The main issue I see with the current PR is all the places where the default Relevant annotations were put, but we can gradually improve the situation after the merge

The idea with unknow is to make that gradual improvement easier. ie currently if you want to find places with incorrect Relevant you can't just grep for it as some are correct (eg tactics working with equalities).

ppedrot (approved):

ppedrot:
@SkySkimmer Failure of bedrock is spurious, right? This is ready to merge if so.

SkySkimmer:
It says "Aggregating timing log..." so it's spurious.

ppedrot:
Github definitely lacks the feature to play  the beginning "Also spoke Zarathustra" in the background while merging...

Zimmi48:
That's because we don't use the merge button.

ejgallego:
some aplay magic should be easy to add to the merge script!

Zimmi48:
@SkySkimmer This PR introduced a command-line flag `-sprop-cumulative` that was not documented except in `coqtop -help`. Contrary to `-allow-sprop`, it doesn't have a corresponding vernac flag AFAICT. What should be done with it?

Zimmi48:
Nevermind, I guess that's the same as option "Elaboration StrictProp Cumulativity". I don't know why I missed it at first.

SkySkimmer:
>Nevermind, I guess that's the same as option "Elaboration StrictProp Cumulativity". I don't know why I missed it at first.

It is not.

Zimmi48:
OK, would be great to document this further, and I'd also like to understand what makes `-sprop-cumulative` something that can only be set when launching Coq.

SkySkimmer:
@Zimmi48 see https://github.com/coq/coq/pull/12034

Issue #8819, opened 2018-10-25
Spurious name collisions with Automatic Introduction


jashug:
#### Version

Coq 8.8+beta1

#### Description of the problem

```
Theorem foo (_ : nat) (H : bool) : True.
(* Error: H is already used. *)
```

To perform automatic introduction, `Theorem` uses `intro` for anonymous parameters, which automatically assigns `H` to the `nat` parameter, and then proceeds to use `intro H` for the second, named, `bool` parameter. This is in `vernac/lemmas.ml` `start_proof_with_initialization`.

We could consider using `intros ? H` instead?

jashug:
After #8820 goes in, this behavior can be improved by tweaking `auto_intros_tac` to call `intros`

ejgallego:
LGTM

jashug:
Typeclass generalization ` ``{TypeClass A B}` (I can't figure out how to write a single backtick...) has the potential to add a complicating factor, in we need to make sure the names `A` and `B` are included as fixed binder names.
Note to self to investigate where the desugaring of the backtick binder happens.

PR #8820, opened 2018-10-25
Standardize handling of Automatic Introduction.

### LINKED ISSUE #8791: `Instance` refreshes parameter names in the global context
#### Version

8.9.dev

#### Operating system
Linux

Get an interactive Coq session running with `-mangle-names mangled_`, and run the following script:
```coq
Class Inhabited (A : Type) : Type := populate { inhabitant : A }.
Hint Mode Inhabited ! : typeclass_instances.
Arguments populate {_} _ : assert.

Global Instance foo (A: Type): Inhabited (list A).
```
This results in the following goal state:
```
  mangled_0 : Type
  ============================
  Inhabited (list mangled_0)
```
That seems rather wrong, because indeed I named type `A` explicitly!  It shouldn't get mangled.

jashug:
This fixes #8791.
We explicitly specify for `intro` the names of binders which are given by the user. This still can suffer from spurious collisions, see #8819. The implementation is factored out of `Lemmas.start_proof_with_initialization`.

**Kind:** bug fix

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -52,6 +52,9 @@ Vernacular commands
 - `Combined Scheme` can now work when inductive schemes are generated in sort
   `Type`. It used to be limited to sort `Prop`.
 
+- Binders for an `Instance` now act more like binders for a `Theorem`.
+  Names may not be repeated, and may not overlap with section variable names.
+
 Tools
 
 - The `-native-compiler` flag of `coqc` and `coqtop` now takes an argument which can have three values:
--- a/tactics/tactics.ml
+++ b/tactics/tactics.ml
@@ -1062,6 +1062,12 @@ let intros_replacing ids =
       (Tacticals.New.tclMAP (fun (id,pos) -> intro_move (Some id) pos) posl)
   end
 
+(* The standard for implementing Automatic Introduction *)
+let auto_intros_tac ids =
+  Tacticals.New.tclMAP (function
+      | Name id -> intro_mustbe_force id
+      | Anonymous -> intro) (List.rev ids)
+
 (* User-level introduction tactics *)
 
 let lookup_hypothesis_as_renamed env sigma ccl = function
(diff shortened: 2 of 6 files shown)


jashug:
A number of developments are now failing, mostly collisions with section variables as far as I can tell. I'll take a shot at writing overlays.

ppedrot (approved):
It's going to be hard to be backward compatible, but the new behaviour looks definitely better.

jashug:
4 overlays remaining. elpi and math-comp appear to be spurious (and I saw elpi work earlier, though math-comp could be hiding an incompatibility). Most of the changes are of one of two types:
```
Instance type1 (H1 : T) (H1 : T) : R.

Section sec.
Variable X : T.
Instance type2 (X : T) : R.
```
Where the first one is fixed by deduplicating the premise name (always an unintentional overlap, as far as I could tell) and the second one can be fixed by not abstracting over `X : T` on the `Instance`, but leaving the generalization until after the section closes. (This is not always possible, but none of the examples I've looked at were making use of the extra generality within the section)

Issue #8819 also cropped up twice.

For backwards compatibility, there should always be some alpha-renaming that is backwards compatible, though I'd struggle to define sufficient conditions.

jashug:
All the overlays have been merged. The builds failing on GitLab are the same as failed on the last run of master. In particular, `iris-lambda-rust` passed. I believe all the failures are spurious and this PR is ready to be merged.

ejgallego (approved):

Zimmi48:
There's now a conflict in CHANGES (although this is the one exception when it is OK to solve the merge conflict when merging).

ppedrot:
@jashug can you rebase to get a clean CI run? I'll merge after that.

ppedrot:
Fiat-crypto died of an out of memory error, I don't know if this is related. I think it used to compile with this PR, right?

jashug:
If I remember correctly, `fiat-crypto` was broken on master last time master was run. So I do not believe I have seen this PR compile with `fiat-crypto`. Unfortunately.

PR #8825, opened 2018-10-26
[libobject] Move object_name next to object definition.

ejgallego:
`object_name` is a particular choice of the implementation of
`Liboject`, thus it makes sense to tie it to that particular module.

This may prove useful in the future as we may want to modify object
naming.

### DIFF
--- a/interp/declare.mli
+++ b/interp/declare.mli
@@ -9,7 +9,6 @@
 (************************************************************************)
 
 open Names
-open Libnames
 open Constr
 open Entries
 open Decl_kinds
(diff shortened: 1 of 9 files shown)


ppedrot (approved):
LGTM. I think we even discussed getting rid of that type with @maximedenes because most of the time the kernel name makes no sense.

ejgallego:
> LGTM. I think we even discussed getting rid of that type with @maximedenes because most of the time the kernel name makes no sense.

I have indeed a in the works PR that may go in that direction.

ppedrot:
I am not supposed to be an owner but I can do it indeed.

ppedrot:
This is quite trivial, I'll merge tomorrow if no objections.

maximedenes (approved):

PR #8826, opened 2018-10-26
[toplevel] Allow to specify default options.

### LINKED ISSUE #8410: Welcome flood from Coqide
#### Version

8.9

Whenever a new tab is created, or whenever a buffer is retracted, Coqide displays on the standard error output the string `Welcome to Coq version (revision)`, which might amount to tens of lines at the end of a standard proof session.

ejgallego:
In some cases, toplevel ML clients may want to modify the default set
of flags that is passed to the main initalization routine. This is for
example useful for `idetop` to suppress some undesired printing at
startup.

I would say that clients ought to have more
control, but I do expect that PRs such as #8690 will help providing a
better separation thus a mode orthogonal API.

### DIFF
--- a/ide/idetop.ml
+++ b/ide/idetop.ml
@@ -537,5 +537,5 @@ let islave_init ~opts extra_args =
 
 let () =
   let open Coqtop in
-  let custom = { init = islave_init; run = loop; } in
+  let custom = { init = islave_init; run = loop; opts = Coqargs.default_opts } in
   start_coq custom
--- a/toplevel/coqargs.ml
+++ b/toplevel/coqargs.ml
@@ -90,7 +90,7 @@ type coq_cmdopts = {
 
 let default_toplevel = Names.(DirPath.make [Id.of_string "Top"])
 
-let init_args = {
+let default_opts = {
 
   load_init   = true;
   load_rcfile = true;
@@ -137,6 +137,8 @@ let init_args = {
 
   print_emacs = false;
 
+  (* Quiet / verbosity options should be here *)
+
   inputstate  = None;
   outputstate = None;
 }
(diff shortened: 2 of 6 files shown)


silene:
Is it possible for `Flags.quiet` to be true but not `print_header`, at the time of printing the header? If not, it seems like an unneeded complexity. Just testing `Flags.quiet` before calling `print_headers` would be a one-liner and it would work just as well.

ejgallego:
> Is it possible for `Flags.quiet` to be true but not `print_header`, at the time of printing the header? If not, it seems like an unneeded complexity. Just testing `Flags.quiet` before calling `print_headers` would be a one-liner and it would work just as well.

This was the old solution and it doesn't work because the IDE wants `quiet := false`.

Unless you want to do a hack such as `quiet := true; common_init; quiet := false` but that can also be problematic due to the summary I think.

ejgallego:
The likely best fix would be indeed to allow toplevel clients to control the feeders of the init process. Indeed, that would be nice as errors in the startup procedure are displayed in the console too.

We can go that way but it will be a bit invasive.

ejgallego:
> We can go that way but it will be a bit invasive.

Yeah I think I will go that route, at least for `master`.

ppedrot (approved):
Looks like quite a big change for such a small hack, but I'm fine with it. It seems that pidetop needs an overlay though.

ejgallego:
I will provide yet another solution, fixing more serious bugs that are still there since the console to feedback transitions.

ejgallego:
I like #8830 better than this even for `master`; adjusting the milestone. Still pondering what to do for the situation of the init feeder and CoqIDE.

ppedrot:
@ejgallego Shouldn't we close this PR then and simply port #8830 to master for the time being?

ejgallego:
There is some complex interaction among the ongoing toplevel PRs at this time. I suggest to keep this PR open for now, I will close/rectify once some of the others PRs are resolved, note however that some design issues are still open so it won't be immediate.

ejgallego:
The handling of `-quiet` flags needs more work but this can go as is for now.

ejgallego:
This is ready, failure is spurious.

PR #8829, opened 2018-10-26
Error when [foo.(bar)] is used with nonprojection [bar], warn if [bar] nonprimitive projection.

SkySkimmer:
Maybe it would make more sense to turn some of the stdlib records into primitive records rather than stop using the primitive projection syntax. Thoughts?

cf https://gitter.im/coq/coq?at=5bd31d8b600c5f64230f50a2

Overlay https://gitlab.mpi-sws.org/iris/lambda-rust/merge_requests/13

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,11 @@ Specification language, type inference
   solved by writing an explicit `return` clause, sometimes even simply
   an explicit `return _` clause.
 
+- Using non-projection values with the projection syntax is not
+  allowed. For instance "0.(S)" is not a valid way to write "S 0".
+  Projections from non-primitive (emulated) records are allowed with
+  warning "nonprimitive-projection-syntax".
+
 Kernel
 
 - Added primitive integers
--- a/dev/ci/user-overlays/08829-proj-syntax-check.sh
+++ b/dev/ci/user-overlays/08829-proj-syntax-check.sh
@@ -1,0 +1,5 @@
+if [ "$CI_PULL_REQUEST" = "8829" ] || [ "$CI_BRANCH" = "proj-syntax-check" ]; then
+    lambdaRust_CI_REF=proj-syntax-check
+    lambdaRust_CI_GITURL=https://github.com/SkySkimmer/lambda-rust
+    lambdaRust_CI_ARCHIVEURL=$lambdaRust_CI_GITURL/archive
+fi
--- a/interp/constrintern.ml
+++ b/interp/constrintern.ml
@@ -1033,7 +1033,7 @@ let intern_qualid ?(no_secvar=false) qid intern env ntnvars us args =
   | TrueGlobal (VarRef _) when no_secvar ->
       (* Rule out section vars since these should have been found by intern_var *)
       raise Not_found
-  | TrueGlobal ref -> (DAst.make ?loc @@ GRef (ref, us)), true, args
+  | TrueGlobal ref -> (DAst.make ?loc @@ GRef (ref, us)), Some ref, args
   | SynDef sp ->
       let (ids,c) = Syntax_def.search_syntactic_definition ?loc sp in
       let nids = List.length ids in
@@ -1043,7 +1043,6 @@ let intern_qualid ?(no_secvar=false) qid intern env ntnvars us args =
       let terms = make_subst ids (List.map fst args1) in
       let subst = (terms, Id.Map.empty, Id.Map.empty, Id.Map.empty) in
       let infos = (Id.Map.empty, env) in
-      let projapp = match c with NRef _ -> true | _ -> false in
       let c = instantiate_notation_constr loc intern intern_cases_pattern_as_binder ntnvars subst infos c in
       let loc = c.loc in
       let err () =
@@ -1067,33 +1066,60 @@ let intern_qualid ?(no_secvar=false) qid intern env ntnvars us args =
         user_err ?loc (str "Cannot change universe level of notation " ++ pr_qualid qid)
       | Some _, _ -> err ()
       in
-      c, projapp, args2
+      c, None, args2
 
-let intern_applied_reference intern env namedctx (_, ntnvars as lvar) us args qid =
+let warn_nonprimitive_projection =
+  CWarnings.create ~name:"nonprimitive-projection-syntax" ~category:"syntax" ~default:CWarnings.Disabled
+    Pp.(fun f -> pr_qualid f ++ str " used as a primitive projection but is not one.")
+
+let error_nonprojection_syntax ?loc qid =
+  CErrors.user_err ?loc ~hdr:"nonprojection-syntax" Pp.(pr_qualid qid ++ str" is not a projection.")
+
+let check_applied_projection isproj realref qid =
+  match isproj with
+  | None -> ()
+  | Some projargs ->
+    let is_prim = match realref with
+      | None | Some (IndRef _ | ConstructRef _ | VarRef _) -> false
+      | Some (ConstRef c) ->
+        if Recordops.is_primitive_projection c then true
+        else if Recordops.is_projection c then false
+        else error_nonprojection_syntax ?loc:qid.loc qid
+        (* TODO check projargs, note we will need implicit argument info *)
+    in
+    if not is_prim then warn_nonprimitive_projection ?loc:qid.loc qid
+
+let intern_applied_reference ~isproj intern env namedctx (_, ntnvars as lvar) us args qid =
   let loc = qid.CAst.loc in
   if qualid_is_ident qid then
-      try intern_var env lvar namedctx loc (qualid_basename qid) us, args
+      try
+        let res = intern_var env lvar namedctx loc (qualid_basename qid) us in
+        check_applied_projection isproj None qid;
+        res, args
       with Not_found ->
       try
-        let r, projapp, args2 = intern_qualid ~no_secvar:true qid intern env ntnvars us args in
+        let r, realref, args2 = intern_qualid ~no_secvar:true qid intern env ntnvars us args in
+        check_applied_projection isproj realref qid;
 	let x, imp, scopes, l = find_appl_head_data r in
 	  (x,imp,scopes,l), args2
       with Not_found ->
         (* Extra allowance for non globalizing functions *)
         if !interning_grammar || env.unb then
+          (* check_applied_projection ?? *)
           (gvar (loc,qualid_basename qid) us, [], [], []), args
         else Nametab.error_global_not_found qid
   else
-    let r,projapp,args2 =
+    let r,realref,args2 =
       try intern_qualid qid intern env ntnvars us args
       with Not_found -> Nametab.error_global_not_found qid
     in
+    check_applied_projection isproj realref qid;
     let x, imp, scopes, l = find_appl_head_data r in
     (x,imp,scopes,l), args2
 
 let interp_reference vars r =
   let (r,_,_,_),_ =
-    intern_applied_reference (fun _ -> error_not_enough_arguments ?loc:None)
+    intern_applied_reference ~isproj:None (fun _ -> error_not_enough_arguments ?loc:None)
       {ids = Id.Set.empty; unb = false ;
        tmp_scope = None; scopes = []; impls = empty_internalization_env}
       Environ.empty_named_context_val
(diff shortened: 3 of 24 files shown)


herbelin:
> Maybe it would make more sense to turn some of the stdlib records into primitive records rather than stop using the primitive projection syntax. Thoughts?

The meaning of "primitive projections" should be first defined. Strictly speaking, I don't see how to  define "primitive projections" differently than what @mattam82 initially did, i.e. as a compact representation of projections. Then, accepting this definition (rather than defining it as attached to "negative" types, i.e. with forbidden `match`), we don't have to wait to represent all projections (even the ones for positive types) in a compact way (up to the incompatibility in referring to subterms of the then-hidden parameters of the projections). By doing so, projections would become a synonymous of compactly-represented projections, as well as a synonymous of "native" projections.

In particular, I approve that we forbid being able to write `0.(S)` to mean `S 0`, as the system allows currently, but I would say that the warning should only be when the "ref" is not a "projection" (primitive or not in a first time, but without having to make a difference in the longer term when all full applications to a projection are eventually represented in a compact way).

SkySkimmer:
I prefer to have the syntax correspond to the internal representation.

ppedrot:
See also #8838, we might want to allow `t.(0)` to anonymously access the first field of a record.

herbelin:
> I prefer to have the syntax correspond to the internal representation.

Note, that, historically, the dot syntax was for `Record`'s projections, independently of the existence of a primitive internal representation of projections.

But, again, if we accept that all projections deserve a compact representation (independently of whether we think at this or that inductive type as `match`-based or projection-based, since, after all, `match`-based types can also be equipped with projections), this shall exactly put the superficial syntax in correspondence with the internal use of a `Constr.Proj` constructor.

> See also #8838, we might want to allow t.(0) to anonymously access the first field of a record.

What would be the advantages of not using a name? Will you skip the named fields in the numbering or will it be an alternative way to name fields?

ppedrot:
> will it be an alternative way to name fields?

Rather that option.

ppedrot (approved):
FTR, I like this.

ppedrot:
I think that this deprecation should be thoroughly discussed.

I am personally in favour of it, and if I read @herbelin's comment correctly he is not against, as this is somewhat orthogonal to the primitive projection implementation, as long as there is an actual notion of projection involved.

What about others? @mattam82 is the original designer of the primitive projection system so he should be involved in the discussion. @maximedenes has been suffering with them during the Great Notation Recession, and I think he would be happy if we could clean up the syntactic sugar.

This is probably also interesting for the few users of primitive projections out there, e.g. @RalfJung and @robbertkrebbers.

mattam82:
I like this too. However Hugo, I’m not sure it’s so easy to support the
notion that Proj(p,c) applies to any record, if you still allow `match` on
the same inductive as you seem to propose (while ideally yes, it would be
cool to allow compact representations as much as possible). For example,
you would have to unfold the primitive application to a `match` too, to
avoid introducing a morally equivalent neutral term. Unfolding the
primitive application to a match would require typing information to
retrieve the parameters, making reduction depend on type inference. This
problematic issue is relegated to the compatibility layer right now, that’s
basically what expand_projection does, and it’s not nice to look at.

Maybe you have another option in mind?
— Matthieu

SkySkimmer:
>For example, you would have to unfold the primitive application to a `match` too, to avoid introducing a morally equivalent neutral term.

How important is it to avoid having a morally equivalent neutral term?

herbelin:
In OCaml, tuples comes with both primitive destructing let and primitive projections (e.g. the code of `Pervasives.fst` is `external fst : 'a * 'b -> 'a = "%field0"`). Why couldn't we do the same?

In particular, I don't see an obstacle to have one reduction rule for `let (x,y) := (t,u) in c` and another distinct one for `fst (t,u)`. We shall want to relate the two forms, but this can be done in the conversion. There would be a priori different ways to do that, but, one of them could be to reduce (in the conversion) `let (x,y) := t in u` to `u[x:=fst t][y:=snd t]`. We already provide definitional eta on non-recursive non-unit primitive projections and this reduction does not seem to require more than that (details to check but I'm pretty confident).

Having both `match` and projections in the same type would help a lot for the compatibility since one of the problems for a straightforward move of the standard library to primitive projections is that the `match` becomes syntactic sugar and stops reducing the same. After all, from a computational point of view, both `match` and projections have their interest. The `match` is for instance able to provide a sharing blocking on neutral terms in situations where expanding it to `fst` and `snd` would increase the size of the term.

One may then wonder what becomes the notion of unique normal form and indeed this is what we would loose: reduction tactics would give different normal forms for terms intended to be equal. But after all, this is already the case with eta since eta is not considered in reduction tactics. And, similarly, we could justify these different normal forms by saying that they differ up to eta. Or we could also provide new reduction tactics which apply on demand eta-reduction, or eta-contraction, or hence also eta-expansion/reduction for match, or eta-expansion/reduction for record.

mattam82:
Indeed, if you go the other way and reduce match on a neutral to a substitution of projections then you can equate the let and projection variants, at the cost of changing the complexity but without depending on type inference. This should be fine even for recursive records. 

Just thinking out loud here, but there is also the case of records with non-definable projections due to sort restriction. If we expanded matches on those to projections we'd have to use "Anonymous" projection somehow? For example:
```
Record myRec : Prop := Rec_intro { somenat : nat; someprop : True }.
```
```
match t : myRec return P : Prop with Rec_intro x y => b
```
would reduce to
``` 
b[t.0/x, t.1/y]
```

SkySkimmer:
>Indeed, if you go the other way and reduce match on a neutral to a substitution of projections then you can equate the let and projection variants, at the cost of changing the complexity but without depending on type inference. This should be fine even for recursive records.

I think the idea is rather to have the conversion `match foo with {| bla = x |} => x end == foo.(bla)` without changing the reduction.
Then nondefinable projections don't come up.

mattam82:
I meant implementing that reduction during conversion, i.e. when finding a neutral term `n` of a record type as a discriminee of a match, expand it to `mkR r.0 ... r.n` and continue conversion on the weak-head normal forms.

SkySkimmer:
So in conclusion do we relax the warning or merge as-is?

herbelin:
> Just thinking out loud here, but there is also the case of records with non-definable projections due to sort restriction. 

>  Then nondefinable projections don't come up.

Right. For these types, surjective pairing does not hold and the only eta-conversion which we can think of is `E[let (...) := ... in ...] == let (...) := ... in E[...] end` for `E` with input and output in the allowed sorts. That seems to be less easy to deal with in the conversion. However, thinking aloud too, even if replacing `let` with `projections` and locally relying on untyped terms, I doubt that reducing `let` into projections could lead to wrong conversion problems in the sense that I conjecture that for any conversion problem going through such locally ill-typed terms, there is a conversion path which is well-typed.

> So in conclusion do we relax the warning or merge as-is?

My reconstruction of @mattam82 and @ppedrot's view (though I'm not 100% sure and it would be better that they express themselves) is that they don't care whether it is for primitive projections only or for all projections. Personally, I'm afraid that a strict warning causes more noise than needed in existing developments, even more if all kinds of projections are eventually turned into compactly represented projections. On the other side, the warning can help to identify what needs to be done in the long term. So, maybe it depends on your own plans?

SkySkimmer:
Maybe what we should do is error when it's not any kind of projection and warn when it's an emulated projection. Then developments which are OK with using `.(foo)` with emulated projections can disable the warning but we don't see total nonsense like `0.(S)`.

ppedrot:
We could probably do even better, i.e. by erroring on non-projections we can still bring the non-primitive and primitive projections closer, by saying that `t.(p)` is restricted to `p` that are registered projections, and forbid the presence of parameters there. Registering non-primitive projection is somewhat already done for other purposes, we could leverage this.

SkySkimmer:
Updated

SkySkimmer:
How do I make a PR on lambda-rust?

ejgallego:
> How do I make a PR on lambda-rust?

You need to go to their gitlab I think @RalfJung ?

RalfJung:
Yes, please create an account at https://gitlab.mpi-sws.org/users/sign_in and then tell me your username so I can unlock your account to create forks.

SkySkimmer:
@RalfJung https://gitlab.mpi-sws.org/SkySkimmer

RalfJung:
@SkySkimmer you can create forks now.

SkySkimmer:
@Zimmi48 Multiple spurious-looking failures in https://gitlab.com/coq/coq/pipelines/48021781 if you want to have a look (lambda-rust failure isn't spurious, I had a bad overlay)

SkySkimmer:
In https://gitlab.mpi-sws.org/iris/lambda-rust/merge_requests/13#note_33219 @RalfJung said 
>Hm, I have to admit I quite liked being able to use "fake projections", it communicates intent very well IMO. But I suppose you have good reasons for removing that ability.

So this syntax was to some extent deliberately used with non-projections.

RalfJung:
Yes, it was certainly deliberate. We were thinking of it as a "user-defined" projection, and at least I was quite happy to be able to use this syntax. I don't recall if @jhjourdan opposed or agreed. ;)

mattam82:
In an ideal situation compact projections could use the r.p syntax leaving the r.(f) syntax free... That requires disambiguating this from a module qualification at interpretation time but I don’t see why we couldn’t do that.

ppedrot:
> I don’t see why we couldn’t do that

Because then you wouldn't be able to use qualified name for projections, as there would be an ambiguity. This looks like a very bad idea to me.

mattam82:
Thanks, I had forgotten about the ambiguity.

ppedrot:
I am assigning this to myself because I think this is a preliminary but critical part of cleaning up the mess around the compatibility layer for primitive projections. In particular, I am strongly advocating for this to be part of 8.10, otherwise we'll just delay fixing the issues to a later time again.

My ultimatum mood leads me to believe that there is no opposition so far to this PR, and thus that I can safely propose a merge by next week if nobody opposes it more fiercely. The patch in itself is quite small, if more people want to review they're welcome.

mattam82:
Can we turn the warning off by default (leaving the error for nonprojections), until we have a clear path for moving everyone to a compact representation? I like the goal but I don't want to spam users like this until we have a beneficial alternative to offer.

ppedrot:
@SkySkimmer would you mind tweaking your PR? Thanks.

SkySkimmer:
changed

Gaëtan Gilbert

ppedrot:
Will merge this tomorrow if no opposition.

ppedrot:
Erf, forgot about this one. @SkySkimmer can you rebase so that this lands in 8.10?

SkySkimmer:
rebased

Gaëtan Gilbert

ppedrot:
Mtac2 broken: consequence or coincidence?

SkySkimmer:
It fails on master too https://gitlab.com/coq/coq/-/jobs/187407939

JasonGross:
Bedrock2 needs an overlay / patch; this was missed because we were previously masking errors via a pipe.  C.f. https://github.com/coq/coq/pull/9872#issuecomment-478534705

tchajed:
Argosy was actually broken by this PR, but it was added to the CI a day late to break the build (everything is fixed upstream now).

ejgallego:
Thanks @JasonGross @tchajed

cpitclaudel:
This patch broke our code too when updating to 8.10.  If the syntax restriction that this introduces is just a matter of taste, can we revert it and discuss it on somewhere more visible than the github tracker? The discussion mentions primitive projections a lot, but what bit us was the new syntactic restriction, and I don't get a clear picture of the connection between both from this discussion. Sorry if I miss something.

AFAICT, @RalfJung and @jhjourdan both expressed support for the old feature when merging the overlay, and this broke a bunch of other projects.  I too find the feature quite useful for writing more readable terms.

Additionally, what is part of a typeclass definition and what is a function derived from it is arguably an implementation detail when you're using the typeclass (as opposed to defining instances), so the previous syntax was quite convenient (and it provided a nice way to do backwards-compatible changes to records and typeclasses without forcing clients to update their syntax).

tchajed:
For the record, I also liked using the projection syntax for functions the way we were in Argosy. I agree with @cpitclaudel that it's an implementation detail whether something is part of a typeclass or is derived.

samuelgruetter:
Being able to use projection syntax for functions would be nice indeed, especially once https://github.com/coq/coq/issues/9814 is fixed.
How about this: Add a switch to the `Arguments` vernac to control whether a function uses projection syntax, which would apply to both parsing and printing, rather than hardcoding this criterion to be whether it's a primitive projection.

PR #8830, opened 2018-10-26
Fix #8410: Welcome flood from Coqide.

### LINKED ISSUE #8410: Welcome flood from Coqide

silene:
Fixes #8410 .

This commit is only for 8.9, since @ejgallego is working on a much more comprehensive pull request for master in #8826 .

### DIFF
--- a/toplevel/coqtop.ml
+++ b/toplevel/coqtop.ml
@@ -453,7 +453,6 @@ let init_toplevel custom_init arglist =
         prerr_endline "See -help for the list of supported options";
         exit 1
       end;
-      Flags.if_verbose print_header ();
       Mltop.init_known_plugins ();
       Global.set_engagement opts.impredicative_set;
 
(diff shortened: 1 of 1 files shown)


toplevel/coqtop.ml
```diff
@@ -519,6 +518,8 @@ type custom_toplevel = {
 let coqtop_init ~opts extra =
   init_color opts;
   CoqworkmgrApi.(init !async_proofs_worker_priority);
+  if CList.is_empty extra then
```
ejgallego: I think this check is not necessary; otherwise that's a good patch that could go even on `master`.

silene: Without this test, the behavior changes with respect to previous versions. Indeed, this would give the following output:
```
$ bin/coqtop -foo
Welcome to Coq machinette:/home/guillaume/src/coq,v8.9 (af858147b44b934f041f7d2f135046f41ef18577)
Don't know what to do with -foo
See -help for the list of supported options
```
instead of the expected output:
```
$ bin/coqtop -foo
Don't know what to do with -foo
See -help for the list of supported options
```

PR #8789, opened 2018-10-21
[coqtop] Tweak API to avoid abstraction breakage.

ejgallego:
We provide a principled mechanism to mask a particular feeder, and
modify the feeder API so it users can name them.

This allows to clean up the use of different feeders in `coqtop`, in
particular it avoids having to leak feeder ids to miscellaneous
subroutines.

PR #8833, opened 2018-10-27
[feedback] Add phase attribute.

ejgallego:
This is an alternative to #8789; I was not very happy with the
solution that we provided there, it seems wrong to mask the feeder
itself in order to reflect modal information.

This patch is not perfect either, but I do wonder what the best
trade-off would be here.

### DIFF
--- a/ide/protocol/xmlprotocol.ml
+++ b/ide/protocol/xmlprotocol.ml
@@ -12,7 +12,7 @@
 
 (** WARNING: TO BE UPDATED WHEN MODIFIED! *)
 
-let protocol_version = "20170413"
+let protocol_version = "20181023"
 
 type msg_format = Richpp of int | Ppcmds
 let msg_format = ref (Richpp 72)
@@ -942,17 +942,19 @@ let of_edit_or_state_id id = ["object","state"], of_stateid id
 let of_feedback msg =
   let content = of_feedback_content msg.contents in
   let obj, id = of_edit_or_state_id msg.span_id in
+  let phase = of_option of_string msg.phase in
   let route = string_of_int msg.route in
-  Element ("feedback", obj @ ["route",route], [id;content])
+  Element ("feedback", obj @ ["route",route], [id;phase;content])
 
 let of_feedback msg_fmt =
   msg_format := msg_fmt; of_feedback
 
 let to_feedback xml = match xml with
-  | Element ("feedback", ["object","state";"route",route], [id;content]) -> {
+  | Element ("feedback", ["object","state";"route",route], [id;phase;content]) -> {
       doc_id = 0;
       span_id = to_stateid id;
       route = int_of_string route;
+      phase = to_option to_string phase;
       contents = to_feedback_content content }
   | x -> raise (Marshal_error("feedback",x))
 
--- a/lib/feedback.ml
+++ b/lib/feedback.ml
@@ -41,6 +41,7 @@ type feedback = {
   doc_id   : doc_id;            (* The document being concerned *)
   span_id  : Stateid.t;
   route    : route_id;
+  phase    : string option;
   contents : feedback_content;
 }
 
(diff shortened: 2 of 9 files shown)


gares:
Both solutions look fine to me.

What I think is not well chose is the name of the new field: you call it "phase" but the comment says "the name of the component" and the usage seems to indicate it is a phase (eg initialization). I think both concepts can make perfect sense, indeed the same component may operate during different phases. Maybe you should think about two distinct fields?

ejgallego:
> Both solutions look fine to me.

Yup, but IMHO we should pick only one. I dunno which one I do prefer.

> What I think is not well chose is the name of the new field: you call it "phase" but the comment says "the name of the component" and the usage seems to indicate it is a phase (eg initialization). I think both concepts can make perfect sense, indeed the same component may operate during different phases. Maybe you should think about two distinct fields?

I agree, you want something for filtering / refinement but it is complicated to have a coherent view. I would rather not introduce two fields but think a bit more about what the semantics of this one should be.

ejgallego:
What should the way forward be here then:
- merge this PR as is
- merge #8789 
- rename the new field and merge [to what?]
?

gares:
The word phase does not suggest there can be a stack of phases, eg that when you leave one you go back to the previous one. The API does that, lets you nest phases. Is it intended? If so I'd rename the attribute part of the feedback message to "top_phase" or "current_phase". Or maybe the type should be just a list of phases, but in the case the XML encoding must be different.

ejgallego:
Yup, `phase` here is a synonym for `current_phase` in the same what that `id` means current `id` etc... I thought the "current" part was understood but could amend if you want.

I am not very happy with the naming either but I couldn't come with a better thing. For example `phase = "typechecking"` does make sense to me. I'll be happy to implement any proposal,

We "stack" phases as entering a phase is similar to a function call, thus we want a stack, however I am not sure this is very interesting to clients and worth the extra effort.

So while we discuss the field, do you agree that this solution is preferred to #8789 ? If so I can close that PR and properly rebase the others.

gares:
> So while we discuss the field, do you agree that this solution is preferred to #8789 ? If so I can close that PR and properly rebase the others.

ok



> Yup, `phase` here is a synonym for `current_phase` in the same what that `id` means current `id` etc... I thought the "current" part was understood but could amend if you want.

I'm not very happy here. A feedback msg applies to one and only one id. While phase is not unique, in the sense that the same command (the same id) may enter the (say) "typechecking" phase multiple times.
Eg `rewrite foo, bar` enter typechecking phase for `foo` then for `bar`.

So, even if there is a "current" implicit in "id" it is a different one in my view.

The problem this PR is trying to solve is simpler I believe but the API is more powerful, maybe too much.

Lets try a different angle. What if the extra field was "emitter"? It would still  be enough to solve the issue if I'm not mistaken. But we would not need a stack API (eg the API, even if implicitly keeps a stack, would have no reason to expose the overridden emitters).

ejgallego:
Umm, a message is emitted by a single "phase" so far. `phase` and `id` are associated to message, not to commands. For example `rewrite foo bar` can indeed produce messages with several `ids` if the stm decides to apply some `VtNow` classification forcing the execution of the previous commands.

I think the stack API is needed otherwise how can you perform bracketing?

But indeed there is a problem here, you can be in "loading init file" and then enter "typechecking". Boom! The current approach doesn't indeed work :( So I dunno; I'm not sure how to proceed.

gares:
> . For example `rewrite foo bar` can indeed produce messages with several `ids` if the stm decides to apply some `VtNow` classification forcing the execution of the previous commands.

Yes but these messages would refer to the other sentences, not the `rewrite` one. The entire point of these ids is to link output to a specific sentence. If the `rewrite` sentence got assigned id 23 by the STM, then all (and only) the messages about that sentence carry id 23. 

> But indeed there is a problem here, you can be in "loading init file" and then enter "typechecking". Boom! The current approach doesn't indeed work :( So I dunno; I'm not sure how to proceed.

If the messages contains the entire stack, not just the top frame, you could see if it contains "loading init".

ejgallego:
> Yes but these messages would refer to the other sentences, not the `rewrite` one. The entire point of these ids is to link output to a specific sentence. If the `rewrite` sentence got assigned id 23 by the STM, then all (and only) the messages about that sentence carry id 23.

So why it is a problem that a sentence gets a message for example related to parsing, and other different message related to typechecking?

> If the messages contains the entire stack, not just the top frame, you could see if it contains "loading init".

Yeah but even so I am not sure how the printer would have to react. In fact I am starting to think that the whole business of showing this decorative messages in toplevel is introducing more problems that it solves.

We could export the stack indeed, that's not a problem, but I am not sure I understand how feedback clients should interpret it.

ejgallego:
Closing this PR in favor of #8950 ; I agree with the issues @gares has raised so I prefer do a less invasive solution for now so we can move forward with the set of depending PRs.

PR #8835, opened 2018-10-28
[econstr] Move `pattern_of_constr` to EConstr.

ejgallego:
The function shouldn't need to observe defined evars, but CI will have
to confirm.

### DIFF
--- a/pretyping/patternops.ml
+++ b/pretyping/patternops.ml
@@ -20,7 +20,6 @@ open Pp
 open Mod_subst
 open Decl_kinds
 open Pattern
-open Environ
 
 let case_info_pattern_eq i1 i2 =
   i1.cip_style == i2.cip_style &&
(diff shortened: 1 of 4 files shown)


ejgallego:
Interesting failures.

pretyping/patternops.ml
```diff
@@ -175,15 +179,15 @@ let pattern_of_constr env sigma t =
...
+      pattern_of_constr env Retyping.(expand_projection env sigma p c [])
     | Evar (evk,ctxt as ev) ->
       (match snd (Evd.evar_source evk sigma) with
       | Evar_kinds.MatchingVar (Evar_kinds.FirstOrderPatVar id) ->
         PMeta (Some id)
       | Evar_kinds.GoalEvar | Evar_kinds.VarInstance _ ->
         (* These are the two evar kinds used for existing goals *)
         (* see Proofview.mark_in_evm *)
-         if Evd.is_defined sigma evk then pattern_of_constr env (Evd.existential_value0 sigma ev)
+         if Evd.is_defined sigma evk then pattern_of_constr env (existential_type sigma ev)
```
mattam82: This becomes impossible, and the right hand side does not make sense :)

mattam82:
I'm guessing this doesn't work because you will get more refined patterns where some implicit arguments for example are now more refined because of unification during typechecking. E.g. instead of [cons _ 0 l] you'll get a pattern [cons nat 0 l].

pretyping/patternops.ml
ejgallego: Indeed, will clean up. I was afraid of some subtle thing going on.

ppedrot (commented):
As already explained, I think that this frontal approach is doomed. We need to fix the pretyper in a way that makes it amenable to understand that something is not quite an evar but rather a syntactic hole. Or rewrite pattern compilation from scratch, maybe.

ppedrot:
I think we should close this PR. As an aside, I don't know how many people will try to fix this issue this way despite the comment explicitly stating it won't work. Making we need to put it in big, red blinking letters?

ejgallego:
Well, my idea to open this PR was to investigate / raise awareness of the current status here. 

If we are sure that the only answer is "full rewrite", then indeed we can proceed to close this PR.

ppedrot:
Let's close.

ejgallego:
I'm ok with closing, however this PR should be transformed into an issue and the discussion preserved.

PR #8836, opened 2018-10-28
[default.nix] Add coq-version and setupHook.

### LINKED ISSUE #8227: Nix build doesn't mirror that in nixpkgs
The `.nix` file in this repo is great and makes my life a lot easier, but there is one pain point: it doesn't closely match [the one in nixpkgs](https://github.com/NixOS/nixpkgs/blob/61bd4d7225742469ed0ada4b4fa27e1559b60cbc/pkgs/applications/science/logic/coq/default.nix).

For instance, the latter [defines some passthru variables](https://github.com/NixOS/nixpkgs/blob/61bd4d7225742469ed0ada4b4fa27e1559b60cbc/pkgs/applications/science/logic/coq/default.nix#L39-L40) that aren't present in the one in this repo:
```coq
    inherit coq-version camlp5;
    inherit (ocamlPackages) ocaml findlib num;
```
and [does something involving `findlib`](https://github.com/NixOS/nixpkgs/blob/61bd4d7225742469ed0ada4b4fa27e1559b60cbc/pkgs/applications/science/logic/coq/default.nix#L131).

Is there any reason for these builds to be different? Is there a way of sharing code between these two repos so that they are as close as possible? 

cc: @Zimmi48

Zimmi48:
Closes #8227 by solving remaining differences.

**Kind:** infrastructure.

I'm not fully sure what should be the value of the `coq-version` attribute. For now I put `git` which means that it won't have to be updated. It could also be some version number coming from a `VERSION` file (`builtins.readFile ./VERSION`) but it's not clear whether we are going to have such a file or if we are going to use tags and git describe instead.
Note that in nixpkgs, the `coq-version` attribute is a value of the form `8.X`. This makes sense because a Coq package compiled with Coq 8.X.Y should work with Coq 8.X.Z without issue. However, this makes less sense in the case of beta releases (we do change magic numbers between a beta and a final release, which means that a package compiled for Coq 8.X+betaY won't work for Coq 8.X.Z. Finally, there is no expectation of compatibility between packages built with various development versions.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -33,6 +33,7 @@
 , shell ? false
   # We don't use lib.inNixShell because that would also apply
   # when in a nix-shell of some package depending on this one.
+, coq-version ? "8.10-git"
 }:
 
 with pkgs;
@@ -101,7 +102,20 @@ stdenv.mkDerivation rec {
 
   installCheckTarget = [ "check" ];
 
-  passthru = { inherit ocamlPackages; };
+  passthru = {
+    inherit coq-version ocamlPackages;
+    dontFilter = true; # Useful to use mkCoqPackages from <nixpkgs>
+  };
+
+  setupHook = writeText "setupHook.sh" "
+    addCoqPath () {
+      if test -d \"$1/lib/coq/${coq-version}/user-contrib\"; then
+        export COQPATH=\"$COQPATH\${COQPATH:+:}$1/lib/coq/${coq-version}/user-contrib/\"
+      fi
+    }
+
+    addEnvHooks \"$targetOffset\" addCoqPath
+  ";
 
   meta = {
     description = "Coq proof assistant";
@@ -113,6 +127,7 @@ stdenv.mkDerivation rec {
     '';
     homepage = http://coq.inria.fr;
     license = licenses.lgpl21;
+    platforms = platforms.unix;
   };
 
 }
--- a/default.nix
+++ b/default.nix
@@ -23,8 +23,8 @@
 
 { pkgs ?
     (import (fetchTarball {
-      url = "https://github.com/NixOS/nixpkgs/archive/06613c189eebf4d6167d2d010a59cf38b43b6ff4.tar.gz";
-      sha256 = "13grhy3cvdwr7wql1rm5d7zsfpvp44cyjhiain4zs70r90q3swdg";
+      url = "https://github.com/NixOS/nixpkgs/archive/69522a0acf8e840e8b6ac0a9752a034ab74eb3c0.tar.gz";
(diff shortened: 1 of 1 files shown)


vbgl:
Did you try to use this setup-hook? How? Does that work well?

vbgl:
Thanks for the feedback. I was also worried about using `coqPackages`, i.e. already packaged third-party dependencies (e.g., `bignums`).

Zimmi48:
>I was also worried about using coqPackages, i.e. already packaged third-party dependencies (e.g., bignums).

Could you elaborate?

vbgl (changes requested):
Please also add a `meta.platforms` attribute.

Zimmi48:
@vbgl Done, and I changed the other things we said. In particular, the `coq-version` attribute is now `8.10-git`, but now I'm starting to wonder. Is this convention of installing the Coq packages in `lib/coq/${coq-version}/user-contrib` really right? Shouldn't we instead install them in `lib/coq/${coq.version}/user-contrib`? Or even go further and store the store path hash of coq in place of `coq-version`?

vbgl (approved):

Zimmi48:
Updated for `dontFilter` (see discussion at NixOS/nixpkgs#49456).

ejgallego (commented):
I'm not sure what is expected from @coq/ci-maintainers here. No comment from me.

default.nix
```diff
@@ -38,6 +38,8 @@
 with pkgs;
 with stdenv.lib;
 
+let coq-version = "8.10-git"; in
```
vbgl: This should be an argument rather than a local variable.

default.nix
```diff
@@ -101,7 +103,20 @@ stdenv.mkDerivation rec {
...
-  passthru = { inherit ocamlPackages; };
+  passthru = {
+    inherit coq-version ocamlPackages;
+    dontFilter = true; # Useful to use mkCoqPackages from <nixpkgs>
+  };
+
+  setupHook = writeText "setupHook.sh" ''
+    addCoqPath () {
+      if test -d "$1/lib/coq/${coq-version}/user-contrib"; then
+        export COQPATH="$COQPATH$\{COQPATH:+:}$1/lib/coq/${coq-version}/user-contrib/"
```
vbgl: The escaping looks wrong. Please double-check.

vbgl (changes requested):
Can you also please update the reference to `nixpkgs` (to e.g., 69522a0)?

default.nix
Zimmi48: Hum, I had fixed this already, so I must have messed up with the force-pushing at some point.

vbgl (approved):

PR #8842, opened 2018-10-29
Towards seeing Global purely as a wrapper on top of kernel functions

herbelin:
**Kind:** architecture

This is a minor PR preparing to functionalization of state: it provides with purely functional variants in `Environ` of functions with algorithmic contents in `Global`.

Incidentally, it renames `Termops.is_template_polymorphic` into `is_template_polymorphic_ind` to reflect that it is intended to work only for inductive types.

### DIFF
--- a/library/global.ml
+++ b/library/global.ml
@@ -128,19 +128,7 @@ let exists_objlabel id = Safe_typing.exists_objlabel id (safe_env ())
 
 let opaque_tables () = Environ.opaque_tables (env ())
 
-let instantiate cb c =
-  let open Declarations in
-  match cb.const_universes with
-  | Monomorphic_const _ -> c, Univ.AUContext.empty
-  | Polymorphic_const ctx -> c, ctx
-
-let body_of_constant_body cb =
-  let open Declarations in
-  let otab = opaque_tables () in
-  match cb.const_body with
-  | Undef _ -> None
-  | Def c -> Some (instantiate cb (Mod_subst.force_constr c))
-  | OpaqueDef o -> Some (instantiate cb (Opaqueproof.force_proof otab o))
+let body_of_constant_body ce = body_of_constant_body (env ()) ce
 
 let body_of_constant cst = body_of_constant_body (lookup_constant cst)
 
@@ -165,8 +153,6 @@ let import c u d = globalize (Safe_typing.import c u d)
 let env_of_context hyps =
   reset_with_named_context hyps (env())
 
-open Globnames
-
 let constr_of_global_in_context = Typeops.constr_of_global_in_context
 let type_of_global_in_context = Typeops.type_of_global_in_context
 
@@ -175,21 +161,9 @@ let universes_of_global gr =
 
 let is_polymorphic r = Environ.is_polymorphic (env()) r
 
-let is_template_polymorphic r = 
-  let env = env() in 
-  match r with
-  | VarRef id -> false
-  | ConstRef c -> false
-  | IndRef ind -> Environ.template_polymorphic_ind ind env
-  | ConstructRef cstr -> Environ.template_polymorphic_ind (inductive_of_constructor cstr) env
+let is_template_polymorphic r = is_template_polymorphic (env ()) r
 
-let is_type_in_type r =
-  let env = env() in
-  match r with
-  | VarRef id -> false
-  | ConstRef c -> Environ.type_in_type_constant c env
-  | IndRef ind -> Environ.type_in_type_ind ind env
-  | ConstructRef cstr -> Environ.type_in_type_ind (inductive_of_constructor cstr) env
+let is_type_in_type r = is_type_in_type (env ()) r
 
 let current_modpath () =
   Safe_typing.current_modpath (safe_env ())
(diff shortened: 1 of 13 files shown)


ppedrot (approved):

herbelin:
This PR mainly modifies `kernel/environ.ml` (@maximedenes) and `library/global.ml` (@silene). I feel that @maximedenes would be the one with most time available for being the assignee. So, would you be willing?

maximedenes:
Sure, will have a look.

library/global.ml
```diff
@@ -209,10 +183,3 @@ let register_inline c = globalize0 (Safe_typing.register_inline c)
 
 let set_strategy k l =
   GlobalSafeEnv.set_safe_env (Safe_typing.set_strategy (safe_env ()) k l)
-
-let set_reduction_sharing b =
```
maximedenes: Why remove this? Isn't it just a wrapper, precisely?

vernac/vernacentries.ml
```diff
@@ -1536,7 +1536,7 @@ let _ =
       optname  = "kernel term sharing";
       optkey   = ["Kernel"; "Term"; "Sharing"];
       optread  = (fun () -> (Global.typing_flags ()).Declarations.share_reduction);
-      optwrite = (fun b -> Global.set_reduction_sharing b) }
+      optwrite = (fun b -> Global.set_typing_flags { (Global.typing_flags ()) with Declarations.share_reduction = b }) }
```
maximedenes: I found the previous code clearer here.

maximedenes (changes requested):
Overall I like the idea, but:
- unrelated changes should go in separate commits (for example the renaming of `is_template_polymorphic`
- the removal of `set_reduction_sharing` seems to make the code more complex, I'm not sure what it brings

vernac/vernacentries.ml
herbelin: It can be moved just before the declare_bool_option though, and at the same time, we could add a `get_reduction_sharing` for the symmetry. It is a bit purely cosmetic but if you insist...

library/global.ml
herbelin: I removed it by consistency:
- it was only for the `share_reduction` flag and not for the other `typing_flag`
- there was an asymmetry between the `write` method in the `declare_option` and the `read` method

In conclusion, I evaluated that the benefit of keeping it was not worth the loss of regularity it was introducing.

vernac/vernacentries.ml
herbelin: I experimented the following:
```ocaml
let get_reduction_sharing () =
  (Global.typing_flags ()).Declarations.share_reduction

let set_reduction_sharing b =
  Global.set_typing_flags
    { (Global.typing_flags ()) with Declarations.share_reduction = b }

let _ =
  declare_bool_option
    { optdepr  = false;
      optname  = "kernel term sharing";
      optkey   = ["Kernel"; "Term"; "Sharing"];
      optread  = (fun () -> (Global.typing_flags ()).Declarations.share_reduction);
      optwrite = (fun b -> Global.set_typing_flags { (Global.typing_flags ()) with Declarations.share_reduction = b }) }
      optread  = get_reduction_sharing;
      optwrite = set_reduction_sharing }
```
I agree that this is clearer, but it is still a bit inconsistent with the rest of the series of `declare_bool_option` in `Vernacentries`. So, I would lean to say that this should go in an independent PR for someone who is willing to make the core more readbable as a whole. What do you think?

SkySkimmer: Maybe we should move this option to Global from Vernacentries. In fact why are random options define in vernacentries instead of next to their bool ref?

herbelin: > In fact why are random options define in vernacentries instead of next to their bool ref?

Iirc, when the `Goptions` mechanism was introduced, most (or all) existing options were put in `Vernacentries`. Then new options arrived and where declared locally to the place they were used. So, yes, it would make sense to now move the options in `Vernacentries` next to where they are relevant.

maximedenes:
Thanks for separating the commits. As for the typing_flags accessors, I think it would make sense to have one for each typing flag. So here, defining `Safe_typing.set_reduction_strategy` and `Global.set_reduction_strategy := globalize0 Safe_typing.set_reduction_strategy`.

I did something similar for VM/native flags in #8870. If you wish, I can do it on this PR too.

library/global.ml
```diff
@@ -209,10 +183,3 @@ let register_inline c = globalize0 (Safe_typing.register_inline c)
 
 let set_strategy k l =
   GlobalSafeEnv.set_safe_env (Safe_typing.set_strategy (safe_env ()) k l)
```
maximedenes: En passant (not introduced by this PR): `set_strategy` should use `globalize0`, shouldn't it?

herbelin:
> I did something similar for VM/native flags in #8870. If you wish, I can do it on this PR too.

If you did it for VM/native, then it is indeed worth to do it for all flags. So, yes, please push to this PR (or use the new github feature to propose a patch).

maximedenes:
@herbelin I rebased, added the setter for reduction sharing to make the code a bit more compact, and changed `set_strategy` to use the same combinators as others. The changes are minimal.

maximedenes (approved):

herbelin:
OK to me.

maximedenes:
@ppedrot can you give a last approval, since this now contains a few lines of code by me (although not a significant contribution to this PR) ? Thanks!

ppedrot (approved):
Not much added indeed, and still OK.

PR #8843, opened 2018-10-29
Remove coqide ml4

SkySkimmer:
Also remove ml4 file handling from the make based build system (coq_makefile still has it ofc, and dune because I'm not sure about removing it there).

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -86,7 +86,7 @@ byte: coqbyte coqide-byte pluginsbyte printers
 
 # This list of ml files used to be in the main Makefile, we moved it here
 # to avoid exhausting the variable env in Win32
-MLFILES := $(MLSTATICFILES) $(GENMLFILES) $(ML4FILES:.ml4=.ml)
+MLFILES := $(MLSTATICFILES) $(GENMLFILES)
 
 include Makefile.common
 include Makefile.vofiles
@@ -148,7 +148,7 @@ endif
 # This include below will lauch the build of all .d.
 # The - at front is for disabling warnings about currently missing ones.
 # For creating the missing .d, make will recursively build things like
-# coqdep_boot (for the .v.d files) or grammar.cma (for .ml4 -> .ml -> .ml.d).
+# coqdep_boot (for the .v.d files) or coqpp (for .mlg -> .ml -> .ml.d).
 
 VDFILE := .vfiles
 MLDFILE := .mlfiles
@@ -166,7 +166,7 @@ DEPENDENCIES := \
 # of include, and they will then be automatically deleted, leading to an
 # infinite loop.
 
-.SECONDARY: $(DEPENDENCIES) $(GENFILES) $(ML4FILES:.ml4=.ml)
+.SECONDARY: $(DEPENDENCIES) $(GENFILES) $(MLGFILES:.mlg=.ml)
 
 ###########################################################################
 # Compilation options
@@ -259,6 +259,7 @@ CAMLP5DEPS:=grammar/grammar.cma
 CAMLP5USE=pa_extend.cmo q_MLast.cmo pa_macro.cmo -D$(CAMLVERSION)
 
 PR_O := $(if $(READABLE_ML4),pr_o.cmo,pr_dump.cmo)
+# XXX unused but should be used for mlp files
 
 # Main packages linked by Coq.
 SYSMOD:=-package num,str,unix,dynlink,threads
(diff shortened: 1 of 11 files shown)


SkySkimmer:
The windows log is too verbose, I can't tell if the failure is spurious.

gares:
I've mixed feelings:
- I'm happy we are so close to not needing camlp5
- ifdefs are ugly but extremely useful
- the solution based on sed is frankly worse than the status quo

What I think would be reasonable is to add basic ifdef support to coqpp ( @ppedrot ).
If this is too hard, then I propose to generate the relevant piece of ml at configure time (or its equivalent for dune) by providing files such as `foo.ml.osx` and copying it to `foo.ml`.

SkySkimmer:
I don't think this quite fits coqpp.
Here's an alternative using m4 instead of custom sed stuff https://github.com/SkySkimmer/coq/commit/07de4a2c99fc3866119f61a88d7e8fc5c94be64f

> If this is too hard, or not relevant enough to others, then I propose to generate the relevant piece of ml at configure time (or its equivalent for dune) by providing files such as foo.ml.osx and copying it to foo.ml.

Not sure how to do such things in dune.

gares:
> Not sure how to do such things in dune.

We do have dune "rules" that copy files around already.
```
./checker/dune:(rule (copy %{project_root}/kernel/names.ml names.ml))
```
I did not test it but something like
```
(rule (copy %{project_root}/ide/foo.ml.%{ocaml-config:os_type} foo.ml))
```
could work

gares:
> Here's an alternative using m4

O_o

SkySkimmer:
When I searched for language agnostic preprocessors it's what came out.

Zimmi48:
@gares Emilio argued strongly about adding anything to configure. There is no equivalent of configure in Dune, only equivalents of steps in the Makefile. But this doesn't change anything right? You are just proposing to add a rule copying the right file instead of emulating `IFDEF`. It doesn't matter if this is done in Makefile or in configure.

gares:
>  You are just proposing to add a rule copying the right file instead of emulating `IFDEF`. It doesn't matter if this is done in Makefile or in configure.

Exactly. It should be trivial in both Make and Dune, and should be cleaner than sed or m4.

ejgallego (commented):
Indeed it is a bit unfortunate that we don't have a preprocessor in coq, that could be very useful [but also open a large can of worms]

I am fine with the m4 solution, I will make Dune do the right thing in other archs. 

But I am also wondering whether we should vendor cppo. As far as I understand that should be a quite clean solution [specially when we are Dune-only]

Zimmi48:
The question of it being at configure time or at make time is irrelevant to this discussion.
Both @ppedrot and @gares support a simple `cp`-based solution so I think this is the one which should get implemented.
Vendoring a pre-processor just for one file seems completely overkill.

ppedrot (changes requested):
Yep, please implement a cp-based mechanism instead.

gares (changes requested):
please use `cp`

SkySkimmer:
Here's the cp version.

ppedrot (approved):

Makefile.build
```diff
@@ -259,6 +259,7 @@ CAMLP5DEPS:=grammar/grammar.cma
 CAMLP5USE=pa_extend.cmo q_MLast.cmo pa_macro.cmo -D$(CAMLVERSION)
 
 PR_O := $(if $(READABLE_ML4),pr_o.cmo,pr_dump.cmo)
+# XXX unused but should be used for mlp files
```
gares: Can you remind me what mlp files are? Weren't they used just for camlp4 <-> camlp5 compatibility?

gares (approved):
LGTM

I let @ejgallego comment on dune. The code I proposed used `%{ocaml-config:os_type}` that is like `Sys.os_type` eg `Unix` or `Windows` ... here we use `X11` or `QUARTZ` or `WIN32`... so there is a mapping to be done that could be avoided by using the values `os_type` holds.

SkySkimmer:
CI is reliably failing on lib/hook.cmo but I can't reproduce locally. What is going on?

Makefile.build
SkySkimmer: camlp4 files that don't depend on VERNAC EXTEND &co (ie the ones used to define it)

gares:
did you try with -j1 on CI?

SkySkimmer:
OK it works now, the issue was a bad cherry pick of the second commit from the preprocessor branch.

ppedrot:
I'm assigning this PR and will merge promptly if nobody objects.

PR #8844, opened 2018-10-29
Move abstract out of tactics.ml

SkySkimmer:
tactics.ml is too big:
- tools slow down 
- it's hard to figure out how it's organized

This is a step towards reducing the problem.

### DIFF
--- a/dev/ci/user-overlays/08844-split-tactics.sh
+++ b/dev/ci/user-overlays/08844-split-tactics.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+if [ "$CI_PULL_REQUEST" = "8844" ] || [ "$CI_BRANCH" = "split-tactics" ]; then
+    Equations_CI_REF=split-tactics
+    Equations_CI_GITURL=https://github.com/SkySkimmer/Coq-Equations
+
+    ltac2_CI_REF=split-tactics
+    ltac2_CI_GITURL=https://github.com/SkySkimmer/ltac2
+
+    fiat_parsers_CI_REF=split-tactics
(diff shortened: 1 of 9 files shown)


ppedrot (approved):
Agreed.

herbelin:
`tactics.ml` is indeed too big. Alternative names could also be `abstracttac.ml`, `tacabstract.ml`, `tacAbstract.ml`, .... Maybe, the naming scheme shall also depend on which names shall be used for further steps of split of `tactics.ml` (which shall be a difficult task: for instance,`intro_patterns` is already internally using forward hooks),

ejgallego:
If we start to wrap libs that would be called `Tactics.Abstract` which IMHO looks pretty good.

herbelin:
> Tactics.Abstract 

Yes, that'll be good.

gares (approved):

JasonGross:
I think this also needs an overlay for fiat-parsers that updates the relevant `.ml{,4}.master` files

SkySkimmer:
How does the fiat-parsers versioning work? There are 810, trunk and master files around.

JasonGross:
trunk is outdated from back when the branch was called trunk.  The 810 and master files should be updated.

JasonGross:
(Currently 810 and master files are identical.  I think we use master as a fallback.)

JasonGross:
What's the policy on whether backwards-incompatible overlays should be merged before or after the relevant PR in Coq?  (I have a slight preference for after, to keep fiat's CI working, but I could be convinced merge before)

Zimmi48:
@JasonGross The policy is to merge the overlay right after.

PR #8845, opened 2018-10-29
Fix typos in the document about CIC

### LINKED ISSUE #8837: Simple typo in the document
### version
master

### Description of the problem
`t~0` in https://github.com/coq/coq/blob/master/doc/sphinx/language/cic.rst#normal-form may be a typo of `t_0`.

wkwkes:
Fixes #8837.


Zimmi48 (approved):
Thanks!
I wonder why this second part is using the unicode arrow instead of the LaTeX `\rightarrow` but your uniformization is obviously correct and an improvement.

PR #8846, opened 2018-10-29
[ci] Only build Docker image when Dockerfile changed.

Zimmi48:
**Kind:** infrastructure.

This allows to get rid of the trick with the `SKIP_DOCKER` variable.
This is using a feature that seems to have been added mainly for this use case:
https://docs.gitlab.com/ee/ci/yaml/README.html#only-changes


Zimmi48:
Actually, I'm stupid but this is not going to work because of this limitation:

>If you are pushing a new branch or a new tag to GitLab, the policy always evaluates to true and GitLab will create a job.

PR #8850, opened 2018-10-29
Private universes for opaque polymorphic constants.

SkySkimmer:
This lets opaque polymorphic constants have universes which are needed for the body but which do not need to be provided when using the constant (because we guarantee that some acceptable universe always exists, and since the constant is opaque it can't be observed).
For instance, 
~~~coq
Set Printing Universes.
Set Universe Polymorphism.
Lemma foo : Type.
Proof. exact (forall _ : Type, Type).
Qed.

Print foo.
(*
foo@{Top.1} = 
Type@{Top.2} -> Type@{Top.3}
     : Type@{Top.1}
(* Top.1 |= Prop < Set
            Set < Top.1
            local: {Top.3 Top.2} |= Top.2 < Top.1
                                    Top.3 < Top.1
                                     *)

foo is universe polymorphic
*)
Definition bar@{i|Set < i} := foo@{i}.
~~~

This is compatibility breaking for those using explicit universe instances (unless they have no opaque definitions I guess). I didn't make the others yet but see the HoTT overlay at https://github.com/SkySkimmer/HoTT/commit/2d4bf3814b1c2a72fb910f072e714c9c1e792b45 for what it gives us in the wild. For instance removing the need to pass around upper bound universes like https://github.com/SkySkimmer/HoTT/commit/2d4bf3814b1c2a72fb910f072e714c9c1e792b45#diff-e2b7e5e83439b62c3f9c1764d6f5eaf1L391 and removing lots of manually minimizing universes.

This can be disabled by unsetting `Private Polymorphic Universes`, such that `Qed` has no semantic difference other than blocking unfolding.

TODO:
- improve printing? `local` is indented strangely in the example above

### DIFF
--- a/doc/sphinx/addendum/universe-polymorphism.rst
+++ b/doc/sphinx/addendum/universe-polymorphism.rst
@@ -441,3 +441,60 @@ underscore or by omitting the annotation to a polymorphic definition.
    semantics that the first use declares it. In this mode, the universe
    names are not associated with the definition or proof once it has been
    defined. This is meant mainly for debugging purposes.
+
+.. flag:: Private Polymorphic Universes
+
+   This option, on by default, removes universes which appear only in
+   the body of an opaque polymorphic definition from the definition's
+   universe arguments. As such, no value needs to be provided for
+   these universes when instanciating the definition. Universe
+   constraints are automatically adjusted.
+
+   Consider the following definition:
+
+   .. coqtop:: all
+
+      Lemma foo@{i} : Type@{i}.
+      Proof. exact Type. Qed.
+      Print foo.
+
+   The universe :g:`Top.xxx` for the :g:`Type` in the body cannot be accessed, we
+   only care that one exists for any instantiation of the universes
+   appearing in the type of :g:`foo`. This is guaranteed when the
+   transitive constraint ``Set <= Top.xxx < i`` is verified. Then when
+   using the constant we don't need to put a value for the inner
+   universe:
+
+   .. coqtop:: all
+
+      Check foo@{_}.
+
+   and when not looking at the body we don't mention the private
+   universe:
+
+   .. coqtop:: all
+
+      About foo.
+
+   To recover the same behaviour with regard to universes as
+   :g:`Defined`, the option :flag:`Private Polymorphic Universes` may
+   be unset:
+
+   .. coqtop:: all
+
+      Unset Private Polymorphic Universes.
+
+      Lemma bar : Type. Proof. exact Type. Qed.
+      About bar.
+      Fail Check bar@{_}.
+      Check bar@{_ _}.
+
+   Note that named universes are always public.
+
+   .. coqtop:: all
+
+      Set Private Polymorphic Universes.
+      Unset Strict Universe Declaration.
+
+      Lemma baz : Type@{outer}. Proof. exact Type@{inner}. Qed.
+      About baz.
--- a/kernel/modops.ml
+++ b/kernel/modops.ml
@@ -338,7 +338,8 @@ let strengthen_const mp_from l cb resolver =
       | Polymorphic_const ctx -> Univ.make_abstract_instance ctx
     in
       { cb with
-	const_body = Def (Mod_subst.from_val (mkConstU (con,u)));
+        const_body = Def (Mod_subst.from_val (mkConstU (con,u)));
+        const_private_poly_univs = None;
 	const_body_code = Some (Cemitcodes.from_val (Cbytegen.compile_alias con)) }
 
 let rec strengthen_mod mp_from mp_to mb =
--- a/printing/printer.ml
+++ b/printing/printer.ml
@@ -143,7 +143,7 @@ let pr_sort sigma s = pr_glob_sort (extern_sort sigma s)
 let _ = Termops.Internal.set_print_constr
   (fun env sigma t -> pr_lconstr_expr (extern_constr ~lax:true false env sigma t))
 
-let pr_in_comment pr x = str "(* " ++ pr x ++ str " *)"
+let pr_in_comment x = str "(* " ++ x ++ str " *)"
 
 (** Term printers resilient to [Nametab] errors *)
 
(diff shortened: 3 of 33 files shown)


stm/stm.ml
```diff
@@ -2473,8 +2477,12 @@ let known_state ~doc ?(redefine_qed=false) ~cache id =
                           ~drop_pt exn_info block_stop, ref false
                     in
                     qed.fproof <- Some (fp, cancel);
+                    let opaque = match keep with
+                      | VtKeep opaque -> opaque
+                      | _ -> Proof_global.Opaque (* this seems strange *)
```
SkySkimmer: I don't know if using Opaque here makes sense. This branch is triggered by undo020 in the test suite, and probably by interactive use in coqide.

SkySkimmer:
>That being said, I am a bit sad that these local universes are still around internally. I suppose the checker would be unable to verify the consistency of the universe graph without them.

Maybe they could be put in the Opaque stuff but I don't understand how it works enough to do it.

gares:
I think it makes sense.

> add option to disable the feature

and deprecate it

I'll try to review the STM code

stm/stm.ml
gares: If it is not keep, then it is drop (eg Admitted). So I guess opaque is fine

SkySkimmer:
>and deprecate it

No, it's a legitimate option. 
For instance maybe you want to change a transparent lemma to opaque without breaking everyone using it with an explicit instance (at least in a first step, to separate issues from relying on the body and issues from relying on the universe info).

gares (changes requested):
I think it is mostly OK wrt the STM.

Given the way the code uses `keep` I'd suggest to change 
```
vernac_qed_type = VtKeep of Proof_global.opacity_flag | VtKeepAsAxiom | VtDrop
```
to something like
```
vernac_keep_as = VtKeepAxiom | VtKeepDefined | VtKeepOpaque
vernac_qed_type = VtKeep of vernac_keep_as | VtDrop
```
(eg it would avoid odd stuff like `is_keep keep || keep == KeepAsAxiom`)

gares:
> For instance maybe you want to change a transparent lemma to opaque without breaking everyone using it

Sorry if I'm naive, but can't you declare "phantom" universe parameters
```
Definition f@{i j} := Type@{i}.
```
(in order to keep the old arity)?

SkySkimmer:
You also have to write out fake constraints. I don't think it would be a fun time.

gares:
> You also have to write out fake constraints. I don't think it would be a fun time.

Sorry to insist, but I'd rather fix the pretty printer so that you can (more) easily copy past the constraints rather than provide yet another Set option (and the multiple code paths). Of course my position stands only If the reason for the option is just migrating... if there are other reasons I'm fine with the option.

SkySkimmer:
>Given the way the code uses keep I'd suggest to change

Changed.

SkySkimmer:
>Sorry to insist, but I'd rather fix the pretty printer so that you can (more) easily copy past the constraints rather than provide yet another Set option (and the multiple code paths). Of course my position stands only If the reason for the option is just migrating... if there are other reasons I'm fine with the option.

Mostly I don't want to force Qed to have different results than Defined other than opacity.

printing/printer.ml
```diff
@@ -211,16 +211,20 @@ let pr_universe_ctx sigma ?variance c =
...
-  if !Detyping.print_universes && not (Univ.AUContext.is_empty c) then
+let pr_abstract_universe_ctx sigma ?variance c local =
+  let open Univ in
+  if !Detyping.print_universes && not (Univ.AUContext.is_empty c && Univ.ContextSet.is_empty local) then
+    let local = if ContextSet.is_empty local then mt() else
+        str "local: " ++ Univ.pr_universe_context_set (Termops.pr_evd_level sigma) local
+    in
     fnl()++pr_in_comment (fun c -> v 0
-      (Univ.pr_abstract_universe_context (Termops.pr_evd_level sigma) ?variance c)) c
+      (Univ.pr_abstract_universe_context (Termops.pr_evd_level sigma) ?variance c) ++ v 0 local) c
```
SkySkimmer: Could someone who understands printing boxes have a look at this so that we don't get weird indentation (see OP)?

herbelin: The `0` will not give indentation, so that should be ok. However, the printing of universes could be improved. For instance, with `Set Printing Depth 10`, I get things like:
```
a
     : Type@{max(Top.1+1,Top.2+1,Top.3+1,Top.4+1,Top.5+1,Top.6+1,Top.7+1,Top.8+1,Top.9)}
(* Top.1 Top.2 Top.3 Top.4 Top.5 Top.6 Top.7 Top.8 Top.9 Top.10 |= 
   Top.10 < Top.9
    *)
```
To allow breaks, `UContext.pr` could use `hov` (or `hv`) rather than `h` (I also wonder whether the `h 0` in front of `v 0` is really useful). In `Ppconstr.pr_constr`, printing the `max` could use `str"," ++ cut()` to allow breaking.

Otherwise, back to this PR, the two steps printing introducing two `|=` is a bit difficult to understand at first. Alternative proposals I could think of includes:
```
Top.1 {Top.3 Top.2} |= Prop < Set
                       Set < Top.1
                       Top.2 < Top.1
                       Top.3 < Top.1
```
or
```
Top.1 {Top.3 Top.2} |= Prop < Set
                       Set < Top.1
                       {Top.2 < Top.1}
                       {Top.3 < Top.1}
```
or
```
Top.1 Top.Internal.3 Top.Internal.2 |= Prop < Set
                                       Set < Top.1
                                       Top.Internal.2 < Top.1
                                       Top.Internal.3 < Top.1
```
or (`Local` instead of `Internal`), or
```
Top.1 Top.3 (internal) Top.2 (internal) |= Prop < Set
                                           Set < Top.1
                                           Top.2 < Top.1 (internal)
                                           Top.3 < Top.1 (internal)
```
or
```
Public universes:
Top.1 |= Prop < Set
         Set < Top.1
Private universes:
Top.3 Top.2 |= Top.2 < Top.1
               Top.3 < Top.1
```

SkySkimmer:
Not sure about the last commit (29c649761fd05839ae8e30e37173fb5ac27ee9c7), but I do think it's better that way overall.

mattam82 (changes requested):
I like this patch, it should reduce the number of useless quantifications significantly, typically those generated by applying eliminators. I reviewed `merge_subgraph` and couldn't find an issue but a comment would be welcome in checker/merge_subgraph. This extends the TCB a bit in presence of the option (`constraints_for` used to be used only outside the kernel). I think this needs a test of the transitivity case `a <= b <= c` where `a` and `c` are kept which is missing it seems.

doc/sphinx/addendum/universe-polymorphism.rst
```diff
@@ -441,3 +441,54 @@ underscore or by omitting the annotation to a polymorphic definition.
...
+
+   Consider the following definition:
+
+   .. coqtop:: all
+
+      Lemma foo@{i} : Type@{i}.
+      Proof. exact Type. Qed.
+      Print foo.
+
+   The universe `Top.xxx` for the :g:`Type` in the body cannot be accessed, we
```
Zimmi48:
```suggestion
   The universe ``Top.xxx`` for the :g:`Type` in the body cannot be accessed, we
```

or

```suggestion
   The universe :g:`Top.xxx` for the :g:`Type` in the body cannot be accessed, we
```

doc/sphinx/addendum/universe-polymorphism.rst
```diff
@@ -441,3 +441,54 @@ underscore or by omitting the annotation to a polymorphic definition.
...
+   only care that one exists for any instantiation of the universes
+   appearing in the type of :g:`foo`. This is guaranteed when the
+   transitive constraint `Set <= Top.xxx < i` is verified. Then when
```
Zimmi48:
```suggestion
   transitive constraint ``Set <= Top.xxx < i`` is verified. Then when
```

or use math-mode (look at the doc or examples in the CIC chapter)

doc/sphinx/addendum/universe-polymorphism.rst
```diff
@@ -441,3 +441,54 @@ underscore or by omitting the annotation to a polymorphic definition.
    semantics that the first use declares it. In this mode, the universe
    names are not associated with the definition or proof once it has been
    defined. This is meant mainly for debugging purposes.
+
+.. flag:: Private Polymorphic Universes
+
+   Consider the following definition:
```
Zimmi48: This option is enabled by default. I think giving a brief description would be good too. Description by examples is usually not fully satisfying.

SkySkimmer:
Changed doc and test suite.

kernel/modops.ml
```diff
@@ -339,7 +339,8 @@ let strengthen_const mp_from l cb resolver =
       | Polymorphic_const ctx -> Univ.make_abstract_instance ctx
     in
       { cb with
-	const_body = Def (Mod_subst.from_val (mkConstU (con,u)));
+        const_body = Def (Mod_subst.from_val (mkConstU (con,u)));
+        const_private_poly_univs = None;
```
SkySkimmer: The checker caught the need for this in HoTT, so I fixed it. I added an Include to the new test file but it will have to wait for #8655 to be actually checked.

SkySkimmer:
Overlays should be good, can I get some reviews and an assignee?

SkySkimmer:
Bench https://ci.inria.fr/coq/view/benchmarking/job/benchmark-part-of-the-branch/578/console

mattam82 (approved):

Zimmi48 (approved):

kernel/modops.ml
mattam82: That looks right, it is actually changing the implementation of cb in the module signature with an alias to cb in the module implementation which itself holds the private universes.

mattam82:
@gares are you satisfied with @SkySkimmer 's answer?

maximedenes (approved):
I'm not competent, but no objection.

gares (approved):

mattam82:
I tried restarting equations and mtac2

PR #8849, opened 2018-10-29
Fix for bug #8848.

mattam82:
I had inverted by mistake the update_evar_source function.
Instead of updating the undefined evar's info with the defined one it did
the converse...

**Kind:** bug fix
Fixes / closes #8848 

- [x] Added / updated test-suite

I managed to minimize the bug and add a corresponding test-suite file.

PR #8852, opened 2018-10-29
Use the obligation evar flag

mattam82:
This stops using evar sources to determine what is an obligation and what isn't, rather relying on explicit declaration of obligation evars by Program/coercion. It should not change anything from the user point of view. Also added an [Evd.get_obligation_evars] to print the obligation evars associated to an evar_map for debugging purposes.

**Kind:** cleanup

EDIT: depends on ~~#8849~~.

### DIFF
--- a/engine/evd.ml
+++ b/engine/evd.ml
@@ -483,6 +483,8 @@ let is_typeclass_evar evd evk =
   let flags = evd.evar_flags in
   Evar.Set.mem evk flags.typeclass_evars
 
+let get_obligation_evars evd = evd.evar_flags.obligation_evars
+
 let set_obligation_evar evd evk =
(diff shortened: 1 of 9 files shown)


mattam82:
This is on top of PR #8849 currently but should be rebased when it is merged.

ppedrot:
Looks like it breaks a test file.

mattam82:
Indeed, one must be explicit when registering obligation evars now and I had missed one case in pretype_type.

ppedrot (approved):
This code still rely on the evar kinds to discriminate the obligations, but it does so somewhat before. I guess it's fine.

ppedrot:
Will merge promptly if nobody opposes.

PR #8856, opened 2018-10-30
[gitlab] Test Ocaml trunk.

ejgallego:
We add a job testing the build of Coq with OCaml 4.08 [AKA `trunk`]

CoqIDE is not supported in 4.08 due to missing `lablgtk`, also `oUnit`
cannot be currently installed, thus we have to add a switch to the
test suite to disable `unit-tests`.

Many deprecation warnings happened in 4.08 so we use the `release`
profile to make them not fatal. Using a 4.08 build profile would be an
option too.


maximedenes (approved):
Finally!

ejgallego:
> Finally!

Unfortunately the list of issues with 4.08 seems quite intimidating,
we will have to wait of course for the release to be stabilized. Among
others, we have the following problems:

1. deprecation of the non-qualified `Stdlib` which perform chaos in the build;
2. many spurious warnings including many useless ones about partial application;
  c.f. ocaml/ocaml#1974
3. ocaml/ocaml#1063 seems to produce all kind of linking problems with plugins.

It would be good to understand how to fix 1. and 2. in a compatible
way; for 3. the situation seems quite complicated; the make-based
system seems to perform better with the `-pack` approach however for
example it seems the linker now rejects plugins that have
interface-only files.

ejgallego:
> Oh I see.

I will make things run somehow, but not before I can reduce the number of depending PRs I'm dragging around here.

ejgallego:
Ok, so I've fixed Dune in 4.08, let's see what this says.

Last time it ran it used to almost complete the test suite, now the linker changes are gonna cause pain for sure before.

ejgallego:
Now that the make-based build doesn't depend on camlp5 I have added indeed a second build using make.

SkySkimmer:
For Pervasives we can do something dumb and simple like stick `module Pervasives = Pervasives [@ocaml.warning "-3"]` in util.ml(i) and then go through it.
What kind of code do we have which triggers partial application warnings? I know goptions.declare_option returns the writer for some reason, we should just make it return unit IMO.

ejgallego:
> I know goptions.declare_option returns the writer for some reason, we should just make it return unit IMO.

This, and some other stuff in `gramlib` at least. We can also do `let _ : _ = GOptions. ....` for example.

ejgallego:
> For Pervasives we can do something dumb and simple like stick `module Pervasives = Pervasives [@ocaml.warning "-3"]` in util.ml(i) and then go through it.

There are other tricks that tend to work well by using shadowing of modules. We will see, for now we can ignore this warning as this seems like a really unstable aspect upstream.

ejgallego:
Dependencies are merged, this should be ready I think.

ejgallego:
Yup, this is working !!

Zimmi48:
>I see you added the progress label? [I removed it in error] Why? As far as I see this PR is ready.

That's a GitHub bug I think: I only removed the "needs: merge of dependency" label at the same time as you removed both this one and "needs: progress".

ejgallego:
> That's a GitHub bug I think: I only removed the "needs: merge of dependency" label at the same time as you removed both this one and "needs: progress".

Indeed that was nice race condition :D

ejgallego:
Some excerpts of the fun that awaits:
```
File "plugins/micromega/coq_micromega.ml", line 1934, characters 14-28:
1934 |   let equal = Pervasives.(=)
                     <alert>^^^^^^^^^^^^^^</alert>
Alert deprecated: module Stdlib.Pervasives
Use Stdlib instead.

Warning 5: this function application is partial,
maybe some arguments are missing.
File "plugins/micromega/coq_micromega.ml", line 95, characters 9-70:
95 |  let _ = declare_int_option (int_opt ["Lia"; "Depth"] lia_proof_depth) in 
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Alert deprecated: Stdlib.Format.pp_open_tag
Use Format.pp_open_stag.
File "lib/pp.ml", line 202, characters 33-45:
202 |                                  pp_close_tag ft ()
                                       <alert>^^^^^^^^^^^^</alert>

File "./plugins/ssr/ssreflect.v", line 17, characters 0-37:
Error:
while loading /builds/coq/coq/plugins/ssr/ssreflect_plugin.cmxs:
no implementation available for Ssrast
```

maximedenes:
Is rejecting `.mli`-only files a desired change on the OCaml side? Or a regression?

ejgallego:
> Is rejecting `.mli`-only files a desired change on the OCaml side? Or a regression?

I dunno, we have been warning for a while that the semantics of `.mli` modules wrt linking are strange; so the dynamic linker may have a valid reason. Note that for example Dune will generate an implementation for all `.mli` only files as to avoid this problem. [As otherwise aliases don't work]

ejgallego:
I guess that many in the OCaml world (myself included) consider that .mli-only modules are a bug. So I would be inclined to view this as a bug in Coq.

I am not an expert tho.

SkySkimmer:
>I guess that many in the OCaml world (myself included) consider that .mli-only modules are a bug.

IMO they make more sense than ml only modules, and duplicating a file is unecessary since dune can do the copy for use with `(modules_without_implementation)`.
Some of our mli/ml only modules would make more sense merged with some functions operating on the types they declare ofc but that doesn't mean we can generalize.

ejgallego:
> IMO they make more sense than ml only modules, and duplicating a file is unecessary since dune can do the copy for use with `(modules_without_implementation)`.

I am talking at the pure compiler-level. In Coq + Dune the compiler sees both a ml and mli file?

What is the semantics you'd propose for adding an interface to a `cma` ?

Note that `.ml` only files have fully defined semantics, what is the problem you see with them?

.gitlab-ci.yml
```diff
@@ -336,6 +336,58 @@ test-suite:egde:dune:dev:
...
+  dependencies: []
+  script:
+    - opam switch create 4.08.0 --empty
+    - eval $(opam env)
+    - opam repo add ocaml-pr https://github.com/ocaml/ocaml-pr-repository.git
+    - opam update
+    - opam install ocaml-variants=4.08.0 num
+    - eval $(opam env)
+    # We use the release profile to avoid problems with warnings
+    - ./configure -profile devel -warn-error no
```
SkySkimmer: Comment seems to differ from code.

dune
```diff
@@ -4,7 +4,9 @@
  (release (flags :standard -rectypes)
           (ocamlopt_flags -O3 -unbox-closures))
  (ireport (flags :standard -rectypes -w -9-27-50+40+60)
-          (ocamlopt_flags :standard -O3 -unbox-closures -inlining-report)))
+          (ocamlopt_flags :standard -O3 -unbox-closures -inlining-report))
+ (ocaml408
+  (flags :standard -strict-sequence -strict-formats -short-paths -keep-locs -rectypes -w -9-27-50+40+60 -warn-error -5 -alert --deprecated)))
```
SkySkimmer: What does `-alert --deprecated` do?

dune-project
```diff
@@ -1,3 +1,3 @@
-(lang dune 1.4)
+(lang dune 1.5)
```
SkySkimmer: Why this change?

SkySkimmer (commented):
What's the plan for this PR? Do we try to get a successful job before merging?

ejgallego:
> What's the plan for this PR? Do we try to get a successful job before merging?

I think indeed we should get a successful job before merging, talking to upstream about the linking problems.

dune-project
ejgallego: The change in `test-suite/Makefile` requires 1.5.

dune
ejgallego: Alerts are a new feature in 4.08 , see OCaml's changelog.

.gitlab-ci.yml
ejgallego: Fixed.

ejgallego:
https://caml.inria.fr/mantis/view.php?id=7876

ejgallego:
https://github.com/ocaml/ocaml/pull/2176 fixes the problems with dynlink, so this should become buildable again soon.

.gitlab-ci.yml
```diff
@@ -336,6 +336,57 @@ test-suite:egde:dune:dev:
...
+    - eval $(opam env)
+    - opam repo add ocaml-pr https://github.com/ocaml/ocaml-pr-repository.git
+    - opam update
+    - opam install ocaml-variants=4.08.0 num
+    - opam pin add dune --dev # ounit lablgtk conf-gtksourceview
+    - opam install dune
+    - eval $(opam env)
+    # We use the release profile to avoid problems with warnings
+    - make -f Makefile.dune trunk
+    - export COQ_UNIT_TEST=
```
SkySkimmer: This doesn't actually work (outdated code?)

SkySkimmer:
Seems to work indeed (modulo my comment).

ejgallego:
> Seems to work indeed (modulo my comment).

Yup ! I wonder how to approach the current set of warnings; we are in a bit of delicate situation here as I am not sure what will OCaml devs do wrt to all those [pretty dramatic] deprecations.

Should we all try to agree that using polymorphic equality is bad?

SkySkimmer:
>Should we all try to agree that using polymorphic equality is bad?

I don't think that's the real issue.

ejgallego:
> I don't think that's the real issue.

Well that's an underlying issue too IMHO, we should avoid module hacks if possible.

A quick fix is indeed to drop qualification of `Pervasives` but this indeed reveals aliasing issues.

We could implement some hack like the one done in Dune with `open` but maybe it is better to address underlying issues first.

ejgallego:
Let's try again, this should be good now :crossed_fingers: 

One thing we are missing tho is the possibility to run benchs with `trunk` and PRs to the compiler, which was suggested by @chambart 

Unfortunately the bench system is still on OPAM 1.2 [and it works by a "miracle"], but as soon as we upgrade it that should mostly do the trick [modulo the annoying alerts].

ejgallego:
Success!!

ejgallego:
Ping

test-suite/Makefile
```diff
@@ -119,6 +117,9 @@ PREREQUISITELOG = prerequisite/admit.v.log			\
 all: run
 	$(MAKE) report
 
+# do nothing
+noop: ;
```
SkySkimmer: Why do we need this noop thing?
Shouldn't it be phony?

ejgallego: It should be phony indeed.

I introduced the noop as to avoid dealing with empty env vars, but maybe it can be avoided, let me see.

ejgallego: Well, the thing is that I do this:
```
   (run make -j 2 BIN= PRINT_LOGS=1 UNIT_TESTS=%{env:COQ_UNIT_TEST=unit-tests}))))
```
so if `COQ_UNIT_TEST` is not set, I enable them. So I pass `noop` to disable it. This is not so bad, as the other choice would be to disable by default, the point is that it is hard to unset a make variable if an envar is set.

SkySkimmer:
There are merge conflicts.

ejgallego:
Merci !

PR #8857, opened 2018-10-30
[library] Better sizing for libobject hashtbl.

ejgallego:
17 is a very small number as files in the stdlib will routinely pass
size 190.

As this is done only on startup, it seems wise to provide a bit more
space.

### DIFF
--- a/library/libobject.ml
+++ b/library/libobject.ml
@@ -65,7 +65,7 @@ type dynamic_object_declaration = {
 let object_tag (Dyn.Dyn (t, _)) = Dyn.repr t
 
 let cache_tab =
-  (Hashtbl.create 17 : (string,dynamic_object_declaration) Hashtbl.t)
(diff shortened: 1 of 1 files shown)


silene:
Where does the `673` value comes from?

I was curious, so I looked at the size of `cache_tab` for 8.9. That is what it looks like:

```
      1 158
      9 168
      1 169
    262 173
     38 175
      1 176
     76 180
      3 182
     66 183
      8 190
      8 192
      7 194
      8 197
      5 201
      1 204
```

So, most of the `.v` files require 173 slots. And the maximum is reached for only one file and it is 204. So, setting the initial size to 673 seems a bit overkill.

ejgallego:
Indeed that's a bit overkill. Current plugins in the CI declare at most 1/2 objects so I've adjusted the size to a more conservative value.

silene (approved):

ppedrot:
That's a trivial change, I'll merge to alleviate @silene's todo list.

PR #8863, opened 2018-10-30
Remove useless make_flexible from univ refresh

SkySkimmer:
In this branch we already know that it's flexible, so it's a
noop (make_flexible_variable ~algebraic:false will NOT turn algebraic
into non-algebraic).

### DIFF
--- a/pretyping/evarsolve.ml
+++ b/pretyping/evarsolve.ml
@@ -65,10 +65,7 @@ let refresh_universes ?(status=univ_rigid) ?(onlyalg=false) ?(refreshset=false)
 	    | UnivRigid ->
 	       if not onlyalg then refresh_sort status ~direction s
 	       else t
-	    | UnivFlexible alg ->
-	       if onlyalg && alg then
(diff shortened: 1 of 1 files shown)


ppedrot:
Is there somewhere a set of invariants about the universe unification algorithm? All of this seems very mysterious to me.

mattam82:
Dealing with algerbraics is a bit of a black art indeed. The purpose of refresh is to allow to move an inferred type, which can end in an algebraic, to a term on which type inference could run again, hence have no algerbraics nor variables that could be substituted by algerbraics. It seems we should always turn algebraic variables into non-algebraic ones actually in refresh.

mattam82:
Do you mean make_flexible_variable ~algebraic:false never actually removed the variable from the algebraic set?

SkySkimmer:
That's what the code says: `let avars' = if algebraic then ... else avars`

mattam82:
One simple fix is to remove the level from the algebraics, but I didn't check if it would impact other things.

PR #8864, opened 2018-10-30
Avoid passing empty environments

maximedenes:
Almost all calls to `Environ.empty_env` are suspicious in the codebase (except for intialization code). We tend to store more and more flags in the environment (type-in-type, reduction sharing, oracle, ...) which is good because it makes the code more functional, but it means that we should really try to get rid of the functions passing dummy environments to the kernel, otherwise these flags are reset in a hard-to-predict way (basically, we are breaking the reader monad).

I got rid of almost all suspicious calls, except:
- one in `Evarutil`
- one in `Tacentries`
- one in `Hints`

I'll need some help for the remaining ones, but if needed, it can be done in another PR.

### DIFF
--- a/plugins/ltac/tacentries.ml
+++ b/plugins/ltac/tacentries.ml
@@ -353,7 +353,7 @@ let extend_atomic_tactic name entries =
       let default = epsilon_value inj e in
       match default with
       | None -> raise NonEmptyArgument
-      | Some def -> Tacintern.intern_tactic_or_tacarg Tacintern.fully_empty_glob_sign def
+      | Some def -> Tacintern.intern_tactic_or_tacarg (Genintern.empty_glob_sign Environ.empty_env) def
     in
     try Some (hd, List.map empty_value rem) with NonEmptyArgument -> None
   in
--- a/plugins/ltac/tacintern.ml
+++ b/plugins/ltac/tacintern.ml
@@ -46,7 +46,6 @@ type glob_sign = Genintern.glob_sign = {
   extra : Genintern.Store.t;
 }
 
-let fully_empty_glob_sign = Genintern.empty_glob_sign Environ.empty_env
 let make_empty_glob_sign () = Genintern.empty_glob_sign (Global.env ())
 
 (* We have identifier <| global_reference <| constr *)
--- a/plugins/ltac/tacintern.ml
+++ b/plugins/ltac/tacintern.ml
@@ -83,7 +83,8 @@ let intern_hyp ist ({loc;v=id} as locid) =
   else if find_ident id ist then
     make id
   else
-    Pretype_errors.error_var_not_found ?loc id
+    CErrors.user_err ?loc Pp.(str "Hypothesis" ++ spc () ++ Id.print id ++ spc() ++
+                              str "was not found in the current environment.")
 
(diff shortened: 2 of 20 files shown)


maximedenes:
The occurrence in `Evarutil` is: https://github.com/coq/coq/blob/0ac673e562c34245e4e48efc428d808e917be79b/engine/evarutil.ml#L359

Maybe @ppedrot or @herbelin have a suggestion on how to improve the code there?

SkySkimmer:
>    one in Evarutil

The hdchar stuff could be changed to only take the rel_context (since that's all it cares about), then we would pass an empty one

> one in Tacentries

I can't see it.

> Hints

I don't know about this.

plugins/ltac/tacentries.ml
```diff
@@ -353,7 +353,7 @@ let extend_atomic_tactic name entries =
       let default = epsilon_value inj e in
       match default with
       | None -> raise NonEmptyArgument
-      | Some def -> Tacintern.intern_tactic_or_tacarg Tacintern.fully_empty_glob_sign def
+      | Some def -> Tacintern.intern_tactic_or_tacarg (Genintern.empty_glob_sign Environ.empty_env) def
```
maximedenes: I need help to get rid of this one (it was there before, this diff is just making it more obvious). @ppedrot, any suggestion?

maximedenes:
The one in `Hints` is here: https://github.com/coq/coq/blob/0ac673e562c34245e4e48efc428d808e917be79b/tactics/hints.ml#L321

Maybe @mattam82 or @ppedrot could help?

plugins/ltac/tacintern.ml
```diff
@@ -83,7 +82,8 @@ let intern_hyp ist ({loc;v=id} as locid) =
   else if find_ident id ist then
     make id
   else
-    Pretype_errors.error_var_not_found ?loc id
```
SkySkimmer: There's an env in the ist, maybe just pass that?

gares (approved):

mattam82:
I think for hints you can just pass the `env` variable that is already there.

plugins/ssr/ssrequality.ml
```diff
@@ -433,15 +433,15 @@ let lz_coq_prod =
 
 let lz_setoid_relation =
```
ppedrot: What is the purpose of this function? It looks like a cache but the implementation is a bit fishy and there are no comments...

plugins/ltac/tacentries.ml
ppedrot: This code is to be called as a subpart of a toplevel command, so I guess you can just call `Global.env`, or better take it as an argument and get the global environment at the verncular command level.

ppedrot (approved):
Haven't seen anything outrageously suspicious and it seems to work, so that's OK for me.

maximedenes:
@ppedrot Since this is a straightforward clean-up and passes all the tests, can we merge? I'll try to address the remaining occurrences in another PR. Thanks!

ppedrot:
Are you fine with this PR @mattam82? I'll merge by the end of the day if you don't complain.

mattam82 (approved):

PR #8867, opened 2018-10-30
Notations: fixing a bug when printing abbreviations in custom entries.

herbelin:
**Kind:** bug fix 

Coercions between entries were missing. See example in `output/Notation4.v` where `< ... >` was missing.


ejgallego (approved):

ejgallego:
Will merge soon. @silene, is the milestone correct here?

silene:
> is the milestone correct here?

No idea. I have a hard time evaluating how dangerous the patch is. In particular, the code now raises a `NoMatch` exception and I don't understand where it is caught.

herbelin:
> `No_match` is caught in `constrextern.ml` which is the caller to `extern_notation` then `match_notation_constr`. Raising `No_match` is the normal way to communicate with `constrextern.ml` to tell that the notation pattern-matching failed. The milestone is intended in the sense that this is a fix for a new feature.

PR #8868, opened 2018-10-31
[Docker] Update OCaml (4.07.1) and camlp5 (7.07-almost)

vbgl:
I’m hitting an OCaml compiler bug in CI. The issues seems unique to 4.07.0 (i.e., not there neither with 4.06.1 nor with 4.07.1), hence this update.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,7 +9,7 @@ stages:
 variables:
   # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
   # for reference]
-  CACHEKEY: "bionic_coq-V2018-10-23-V1"
+  CACHEKEY: "bionic_coq-V2018-10-30-V1"
   IMAGE: "$CI_REGISTRY_IMAGE:$CACHEKEY"
   # By default, jobs run in the base switch; override to select another switch
   OPAM_SWITCH: "base"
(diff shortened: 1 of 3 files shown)


silene:
> I’m hitting an OCaml compiler bug in CI.

Is it due to a specific PR? (Just so that I keep track that I need to regenerate the Docker image for 8.9 if I backport it.)

vbgl:
Thanks Guillaume. This is for PR #6914. I believe it is not going to be back-ported to 8.9.

ejgallego (commented):
That's fine, but please add details about the compiler bug you are facing. That could potentially impact users so we need to document it.

ejgallego:
@vbgl , note that a cool trick that I didn't try but in theory now possible is to compose the build of the OCaml compiler itself with the Coq build [using Dune for both]

That should allow you to perform a painless bisect on OCaml.

vbgl:
Said bug is unrelated to Coq: it’s valid OCaml code generated by Coq extraction.

ejgallego:
> Said bug is unrelated to Coq: it’s valid OCaml code generated by Coq extraction.

Even so I'd like to properly document the bug in the log please. The fact that is not a Coq bug doesn't mean it won't affect Coq users [such as the intel CPU bug did]

ejgallego:
Also, I've realized the commits do look wrong, submitting a new review.

ejgallego (changes requested):
Please:
- squash the commits: the imagine with 4.07.1 won't build due to wrong camlp5 version,
- update the `dev/dune-workspace.all` file,
- document in the commit log the OCaml bug you have found and how it affected Coq.

vbgl:
Emilio, I think I addressed your comments.

Did you expect more detail concerning the OCaml bug? It’s a bit tricky to minimize and report.

ejgallego (approved):
That's more than enough IMHO @vbgl , thanks! The idea is that if someone else finds the bug we have somehow a trace of it in the log.

Merging as soon as CI is OK.

PR #8871, opened 2018-10-31
[library] Move Nametab/Lib specific-names to Nametab

ejgallego:
We move `object_prefix` and `global_dir_reference` to `Nametab`. This highlights the coupling of
`Lib` and `Nametab` wrt naming.

This also thins `Libname`, which IMHO is a good thing as we are talking about "locally internal" naming here.

### DIFF
--- a/library/declaremods.ml
+++ b/library/declaremods.ml
@@ -139,7 +139,7 @@ let expand_sobjs (_,aobjs) = expand_aobjs aobjs
      Module M:SIG. ... End M. have the keep list empty.
 *)
 
-type module_objects = object_prefix * Lib.lib_objects * Lib.lib_objects
+type module_objects = Nametab.object_prefix * Lib.lib_objects * Lib.lib_objects
 
 module ModObjs :
  sig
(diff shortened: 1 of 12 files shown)


mattam82:
I don't feel competent enough on the nametab/libname interaction to review this. @herbelin might be better placed.

herbelin:
> I don't feel competent enough on the nametab/libname interaction to review this. @herbelin might be better placed.

I would rather move `object_prefix` to `Lib`. Otherwise no opinion. If the objective is to remove `Libnames` (and, indeed, there are so many files about names that I never know which one I need to open), then ok for moving `global_dir_reference` to `Nametab`.

ejgallego:
> I would rather move `object_prefix` to `Lib`.

@herbelin that was my first plan but `Lib` depends on `Nametab`.

I want to reify the `Lib` / `Nametab` interaction, and I feel I am close, but it is still too early to see how it will look.

herbelin:
> I want to reify the Lib / Nametab interaction, and I feel I am close, but it is still too early to see how it will look.

OK, then, let's go this way in the short term. I'm assigning.

ejgallego:
Thanks @herbelin

herbelin (approved):
An intermediate step towards simplifying the number of files about names.

Maybe some of the functions from `Libnames` can go to `Names`, seeing `Names` as a "name library", even if the functions from `Libnames` are not used by the kernel, strictly speaking.

herbelin:
Preferable to rebase to recheck elpi after the env of evar printer has been merged.

herbelin:
Note: to avoid time-consuming synchronisation steps, I finally double-check elpi myself.

ejgallego:
Thanks @herbelin

PR #8882, opened 2018-10-31
Fix #8881: validate fails to use inductive equivalence in case_info

### LINKED ISSUE #8881: validate fails to use inductive equivalence in case_info
~~~coq
Module Type ASIG.
  Inductive t : Set := a | b : t.
  Definition f := fun x => match x with a => true | b => false end.
End ASIG.

Module Type BSIG.
  Declare Module A : ASIG.
  Definition f := fun x => match x with A.a => true | A.b => false end.
End BSIG.

Module C (A : ASIG) (B : BSIG with Module A:=A).

  Lemma test : forall x, A.f x = B.f x.
  Proof.
    intro x. unfold B.f, A.f.
    destruct x; reflexivity.
  Qed.

End C.
(* Fatal Error: Type error: WrongCaseInfo *)
~~~

SkySkimmer:
See also 55dbe8e2fa7ed2053ecd54140f6bcbdf31981e0b

### DIFF
--- a/checker/inductive.ml
+++ b/checker/inductive.ml
@@ -388,7 +388,7 @@ let type_case_branches env (pind,largs) (p,pj) c =
 let check_case_info env indsp ci =
   let mib, mip as spec = lookup_mind_specif env indsp in
   if
-    not (eq_ind_chk indsp ci.ci_ind) ||
+    not (mind_equiv env indsp ci.ci_ind) ||
     (mib.mind_nparams <> ci.ci_npar) ||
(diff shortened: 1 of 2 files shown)


maximedenes:
@SkySkimmer do you know that all these problems are fixed by #8773 ? We can still merge this, but there may be better ways to spend time ;)

ppedrot:
Depends on the ETA of #8773... At least the changes are minimal enough that it cannot hurt to merge them.

ppedrot (approved):
Checking inductive names up-to quotient is already nonsense in the checker, I don't see how this can hurt in any worse way.

PR #8884, opened 2018-10-31
Improve rendering of the credits.

silene:
This mostly fixes text that was italicized instead of teletyped. When possible, tactic names have been made to point to their documentation.

Also, the date of the 8.9 release has been proactively changed to November.

### DIFF
--- a/doc/sphinx/credits.rst
+++ b/doc/sphinx/credits.rst
@@ -557,7 +557,7 @@ Pierre Courtieu, Julien Forest and Yves Bertot added extra support to
 reason on the inductive structure of recursively defined functions.
 
 Jean-Marc Notin significantly contributed to the general maintenance of
-the system. He also took care of `coqdoc`.
+the system. He also took care of ``coqdoc``.
 
 Pierre Castéran contributed to the documentation of (co-)inductive types
 and suggested improvements to the libraries.
@@ -642,7 +642,7 @@ improvements, Jean-Marc Notin provided help in debugging, general
 maintenance and coqdoc support, Vincent Siles contributed extensions of
 the Scheme command and of injection.
 
-Bruno Barras implemented the `coqchk` tool: this is a stand-alone
+Bruno Barras implemented the ``coqchk`` tool: this is a stand-alone
 type checker that can be used to certify .vo files. Especially, as this
 verifier runs in a separate process, it is granted not to be “hijacked”
 by virtually malicious extensions added to |Coq|.
@@ -817,12 +817,12 @@ expressions occurring in the goal by pattern in tactics such as set or
 destruct. Hugo Herbelin also relied on ideas from Chung-Kil Hur’s Heq
 plugin to introduce automatic computation of occurrences to generalize
 when using destruct and induction on types with indices. Stéphane Glondu
-introduced new tactics constr\_eq, is\_evar and has\_evar to be used
+introduced new tactics :tacn:`constr_eq`, :tacn:`is_evar`, and :tacn:`has_evar`, to be used
 when writing complex tactics. Enrico Tassi added support to fine-tuning
-the behavior of simpl. Enrico Tassi added the ability to specify over
+the behavior of :tacn:`simpl`. Enrico Tassi added the ability to specify over
 which variables of a section a lemma has to be exactly generalized.
 Pierre Letouzey added a tactic timeout and the interruptibility of
-vm\_compute. Bug fixes and miscellaneous improvements of the tactic
+:tacn:`vm_compute`. Bug fixes and miscellaneous improvements of the tactic
 language came from Hugo Herbelin, Pierre Letouzey and Matthieu Sozeau.
 
 Regarding decision tactics, Loïc Pottier maintained nsatz, moving in
@@ -866,7 +866,7 @@ Pierre Courtieu and Arnaud Spiwack contributed new features for using
 Coq through Proof General.
 
 The Dp plugin has been removed. Use the plugin provided with Why 3
-instead (http://why3.lri.fr).
+instead (http://why3.lri.fr/).
 
 Under the hood, the |Coq| architecture benefited from improvements in
 terms of efficiency and robustness, especially regarding universes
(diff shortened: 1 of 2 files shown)


doc/sphinx/credits.rst
```diff
@@ -1199,25 +1199,25 @@ Credits: version 8.7
...
-- New tactics: variants of tactics supporting existential variables eassert,
-  eenough, etc... by Hugo Herbelin. Tactics extensionality in H and
-  inversion_sigma by Jason Gross, specialize with ... accepting partial bindings
+- New tactics: variants of tactics supporting existential variables :tacv:`eassert`,
+  :tacv:`eenough`, etc... by Hugo Herbelin. Tactics ``extensionality in H`` and
+  :tacv:`inversion_sigma` by Jason Gross, ``specialize with ...`` accepting partial bindings
   by Pierre Courtieu.
 
-- Cumulative Polymorphic Inductive Types, allowing cumulativity of universes to
+- ``Cumulative Polymorphic Inductive Types``, allowing cumulativity of universes to
```
Zimmi48: I don't see why this should be displayed as teletyped.

Zimmi48 (commented):
Thanks! Just one minor comment.

doc/sphinx/credits.rst
```diff
@@ -817,12 +817,12 @@ expressions occurring in the goal by pattern in tactics such as set or
...
 when using destruct and induction on types with indices. Stéphane Glondu
-introduced new tactics constr\_eq, is\_evar and has\_evar to be used
+introduced new tactics :tacn:`constr_eq`, :tacn:`is_evar`, and :tacn:`has_evar`, to be used
 when writing complex tactics. Enrico Tassi added support to fine-tuning
-the behavior of simpl. Enrico Tassi added the ability to specify over
+the behavior of :tacn:`simpl`:. Enrico Tassi added the ability to specify over
 which variables of a section a lemma has to be exactly generalized.
 Pierre Letouzey added a tactic timeout and the interruptibility of
-vm\_compute. Bug fixes and miscellaneous improvements of the tactic
+:tacv:`vm_compute`:. Bug fixes and miscellaneous improvements of the tactic
```
Zimmi48: Use `tacn`, not `tacv`.

doc/sphinx/credits.rst
```diff
@@ -1199,25 +1199,25 @@ Credits: version 8.7
 and cleanups of the internals of the system along with a few new features. The
 main user visible changes are:
 
-- New tactics: variants of tactics supporting existential variables eassert,
-  eenough, etc... by Hugo Herbelin. Tactics extensionality in H and
-  inversion_sigma by Jason Gross, specialize with ... accepting partial bindings
+- New tactics: variants of tactics supporting existential variables :tacv:`eassert`,
```
Zimmi48: Same here/

doc/sphinx/credits.rst
```diff
@@ -1199,25 +1199,25 @@ Credits: version 8.7
...
+  :tacv:`eenough`, etc... by Hugo Herbelin. Tactics ``extensionality in H`` and
```
Zimmi48: Same here.

doc/sphinx/credits.rst
```diff
@@ -1199,25 +1199,25 @@ Credits: version 8.7
...
+  :tacv:`inversion_sigma` by Jason Gross, ``specialize with ...`` accepting partial bindings
```
Zimmi48: Same here.

Zimmi48 (commented):
For the purpose of referring to tactic names, the distinction between `tacv` and `tacn` does not exist. It should always be `tacn`. This is what makes the doc build fail.

doc/sphinx/credits.rst
silene: The documentation says:

    For example, you could link to the tactic variant above using :tacv:`simpl_at`

I naively thought the documentation was right.

doc/sphinx/credits.rst
silene: As far as I am concerned, the uppercased letters made it an obvious reference to the vernacular command `Cumulative Polymorphic Inductive`. (Putting Types inside the teletype part was an error of mine.)

Zimmi48: Indeed, and and types should probably have been lowercase.

doc/sphinx/credits.rst
Zimmi48: @cpitclaudel The documentation is wrong, right?

cpitclaudel: Indeed, the documentation is wrong :disappointed: `tacv` and `warn` are the only exceptions to the rule that `.. xxx::` can be linked to using `:xxx:`.

https://github.com/coq/coq/blob/2a93216a3851688dd29c06a29c6d1442898faab8/doc/tools/coqrst/coqdomain.py#L1023-L1037

cpitclaudel: This should be documented in README.rst. But I wonder whether we should fix the users or the code. Maybe `:tacv:` should be a synonym of `:tacn:`? The logic behind collapsing both `tacv` and `tacn` under the same role `tacn` was that it was a pain to have to distinguish when writing links. But it can also be confusing that `tacv` doesn't work :/

Zimmi48: Both solutions are OK to me. However, I'm quite surprised by the collapsing of `exn` and `warn`. These are not the same to me.

doc/sphinx/credits.rst
silene: Replacing ``:tacv:`inversion_sigma` `` by ``:tacn:`inversion_sigma` `` did not work. Despite `proof-engine/tactics.rst` containing `.. tacv:: inversion_sigma`, the script complains: `tacn reference target not found: inversion_sigma`. What am I missing?

Zimmi48: The other issue is that, by default, tactic variants do not get an index, unless you give them a `:name:`. @cpitclaudel I think this was a request from @maximedenes, right?

silene:
The compilation finally went through, here is the rendered version of the credits: https://coq.gitlab.io/-/coq/-/jobs/116901388/artifacts/_install_ci/share/doc/coq/sphinx/html/credits.html

Zimmi48 (approved):
Very nice!

doc/sphinx/credits.rst
cpitclaudel: Right.

Issue #8449, opened 2018-09-10
The “Register” command adds trusted assumptions — but nobody knows.


vbgl:
The “Register” vernacular command is not documented (#2409). It creates a link between a Coq constant and a primitive operation of the reduction machines (`vm_compute`, `native_compute`). It basically declares that the given Coq constant is soundly implemented by the given operation. The compilation then uses these links to generate efficient code (using the primitive operation rather than the Coq-defined one).

The validity of these links is trusted: nothing prevents a user to declare nonsensical links (e.g., that addition is soundly implemented by the subtraction operator) and deduce nonsensical facts from them. Only manual reviews can help understanding whether these “Register” declarations make sense or not. This is very similar to an axiom.

However, there is no easy way to know whether a term depends on such declarations. I think that the “Print Assumptions” command (or a similar one) should help understanding what declarations are used when type-checking a term.

JasonGross:
Abuse of this permits a proof of False, right?

@maximedenes is it possible to do stack-smashing by mismatching argument counts or universe polymorphic state with this?

andres-erbsen:
~~~~~
Register head031 as int31 div21 in "coq_int31" by True.
Register tail031 as int31 tail0 in "coq_int31" by True.

Compute tail031 (head031 T31).
(*
Program received signal SIGSEGV, Segmentation fault.
0x0000555556161308 in coq_interprete ()
(gdb) bt
#0  0x0000555556161308 in coq_interprete ()
#1  0x000055555606851f in camlCsymtable__eval_to_patch_102208 ()
#2  0x0000555555ed38d6 in camlVnorm__cbv_vm_205387 ()
#3  0x0000555555e5b231 in camlRedexpr__fun_405929 ()
#4  0x0000555555d69eb3 in camlVernacentries__redfun_1707559 ()
#5  0x0000555555e04573 in camlPrettyp__gallina_print_eval_1506529 ()
#6  0x0000555555d6c79f in camlVernacentries__interp_2208669 ()
#7  0x000055555606ac1c in camlFlags__with_modified_ref_inner_102094 ()
#8  0x0000555555d6d8ab in camlVernacentries__fun_2514679 ()
#9  0x0000555555d6d750 in camlVernacentries__aux_2308921 ()
#10 0x0000555555d6d9f6 in camlVernacentries__interp_2508937 ()
#11 0x0000555555cf5f37 in camlStm__aux_interp_706541 ()
[...]

PR #8886, opened 2018-10-31
[VM] Fix compilation of int31 eliminators

### LINKED ISSUE #8885: vm_compute crash on int31_rect
#### Version

Crash witnessed on Linux x86_64 with Coq versions 8.5pl3, 8.6.1, 8.7.2, 8.8.2, some recent master.

The following example crashes (segmentation fault).

```
From Coq Require Import Cyclic31.

Definition Nat `(int31) := nat.
Definition Zero (_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _: digits) := 0.

Check (eq_refl (int31_rect Nat Zero 1) : 0 = 0).
Check (eq_refl (int31_rect Nat Zero 1) <: 0 = 0).
```

Maxime pretends such a crash can be turned into a proof of False.

Spoiler: the compilation of the `int31_rect` constant occurs before the retroknowledge is informed that `int31` is the int31 type. The fix is trivial and coming soon.

vbgl:
The compilation to bytecode of the elimination schemes for int31 must
happen after the int31 type is registered to the retroknowledge.
Otherwise, the “decompint” instruction is not emitted.

Fixes #8885.

### DIFF
--- a/test-suite/bugs/closed/bug_8885.v
+++ b/test-suite/bugs/closed/bug_8885.v
@@ -1,0 +1,8 @@
+From Coq Require Import Cyclic31.
+
+Definition Nat `(int31) := nat.
+Definition Zero (_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _: digits) := 0.
+
+Check (eq_refl (int31_rect Nat Zero 1) : 0 = 0).
+Check (eq_refl (int31_rect Nat Zero 1) <: 0 = 0).
+Check (eq_refl (int31_rect Nat Zero 1) <<: 0 = 0).
--- a/theories/Numbers/Cyclic/Int31/Int31.v
+++ b/theories/Numbers/Cyclic/Int31/Int31.v
@@ -15,6 +15,8 @@ Require Import Wf_nat.
 Require Export ZArith.
 Require Export DoubleType.
 
+Local Unset Elimination Schemes.
+
 (** * 31-bit integers *)
 
 (** This file contains basic definitions of a 31-bit integer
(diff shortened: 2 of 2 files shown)


maximedenes (approved):
This is very well spotted!

maximedenes:
Note that this may be a critical bug.

herbelin:
@vgbl, could you explain more? What is the purpose of the color overlay? How to ensure that no user will not do the same error and, maybe put the vm into inconsistent state by issuing contradictory `Register` statement at the vernacular level?

maximedenes:
It is expected that one can introduce an inconsistency by using the `Register` command, just like `Axiom`. What is not expected is that some terms from the standard library are compiled incorrectly.

JasonGross:
@herbelin @ppedrot made a suggestion at https://github.com/coq/coq/issues/8449#issuecomment-422661062 about perhaps forbidding the `Register` command outside of `-boot` mode.

herbelin:
> @herbelin @ppedrot made a suggestion at #8449 (comment) about perhaps forbidding the Register command outside of -boot mode.

Thanks. I had missed the context of this issue. This helps to understand the current PR.

vbgl:
Rebased & removed the overlay, which is not needed any longer.

maximedenes:
@herbelin can I merge this?

ppedrot:
@maximedenes You should probably mark this as critical. It could be used to smash stacks and the like.

silene (approved):
Approved, but I agree with @JasonGross that `int31_rec` should also be generated for compatibility.

maximedenes:
@vbgl ok, can you add `int31_rec` as requested?

vbgl:
Done.

herbelin:
@herbelin can I merge this?

No objection on my side (up to adding `_rec` but that's already requested). As it is, I guess that the view is that the vm is after all more for speed than for safety.

silene:
@vbgl Do I understand correctly that, to backport this PR to 8.9, I just need to pin CoLoR to its current master?

vbgl:
I’m not sure to understand your question. Is it related to the fact that the MS-Windows installer ships a version of CoLoR? If so then yes, you need to refer to a compatible version of that library.

vbgl:
That’s right. Updating `CoLoR` should be enough.

PR #8325, opened 2018-08-26
proof that R is uncountable

VincentSe:
**Kind:** feature.

This is my first pull request to Coq, so it is probably wrong in many places. It is only a new proof, that R is uncountable, it doesn't modify Coq's OCaml code. I am not sure about the header I copied into Runcountable.v. I do accept it is released under the Lesser GPL.

I am not a researcher and I did this on my free time, so I did not put an organisation next to my name in the CREDITS file.

The first part of the proof is a slight extension of Coq.Logic.ConstructiveEpsilon. But I did not modify this file because I do not know yet the coding conventions of Coq.

I could not run the test suite. Even before my modification, make test-suite fails on commit 56b355e9e3e0394625cc40084f7cf11a949ebb29, with this error:

TEST      unit-tests/printing/proof_diffs_test.ml
TEST      unit-tests/clib/inteq.ml
TEST      unit-tests/clib/unicode_tests.ml
make[3]: Entering directory '/home/vincent/coq/test-suite'
BUILDING SUMMARY FILE
FAILURES
    coq-makefile/findlib-package/run.sh...Error!
    coq-makefile/timing/run.sh...Error!
Makefile:198: recipe for target 'report' failed
make[3]: *** [report] Error 1
make[3]: Leaving directory '/home/vincent/coq/test-suite'
Makefile:119: recipe for target 'all' failed
make[2]: *** [all] Error 2
make[2]: Leaving directory '/home/vincent/coq/test-suite'
Makefile.build:627: recipe for target 'test-suite' failed
make[1]: *** [test-suite] Error 2
make[1]: Leaving directory '/home/vincent/coq'
Makefile:165: recipe for target 'submake' failed
make: *** [submake] Error 2

The 2 failed tests I noticed are below, but make report fails so I don't know what is wrong with them:
FAILED    coq-makefile/timing.log
FAILED    coq-makefile/findlib-package.log

I did check that coqc accepts Runcountable.v.

PR #8888, opened 2018-10-31
Proof runcountable rebase

VincentSe:
Kind: feature

This is a reopening of #8325, after two bad merges.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -165,6 +165,7 @@ of the Coq Proof assistant during the indicated time:
   Nickolai Zeldovich (MIT 2014-2016)
   Théo Zimmermann (ORCID: https://orcid.org/0000-0002-3580-8806,
                    INRIA-PPS then IRIF, 2015-now)
+  Vincent Semeria (2018-now)
 
 ***************************************************************************
 INRIA refers to:
--- a/CREDITS
+++ b/CREDITS
@@ -165,7 +165,7 @@ of the Coq Proof assistant during the indicated time:
   Nickolai Zeldovich (MIT 2014-2016)
   Théo Zimmermann (ORCID: https://orcid.org/0000-0002-3580-8806,
                    INRIA-PPS then IRIF, 2015-now)
-  Vincent Semeria (2018-now)
+  Vincent Semeria (2018)
 
 ***************************************************************************
 INRIA refers to:
--- a/CREDITS
+++ b/CREDITS
@@ -152,6 +152,7 @@ of the Coq Proof assistant during the indicated time:
   Clément Renard (INRIA, 2001-2004)
   Claudio Sacerdoti Coen (INRIA, 2004-2005)
   Amokrane Saïbi (INRIA, 1993-1998)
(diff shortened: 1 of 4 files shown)


VincentSe:
I implemented the smallest element in constructive epsilon and used FinFun.Bijective in the final statement of the theorem.

However it is difficult to completely replace enumeration by Bijective, as sometimes we want to have the inverse function in Sort Set.

CREDITS
```diff
@@ -165,6 +165,7 @@ of the Coq Proof assistant during the indicated time:
   Nickolai Zeldovich (MIT 2014-2016)
   Théo Zimmermann (ORCID: https://orcid.org/0000-0002-3580-8806,
                    INRIA-PPS then IRIF, 2015-now)
+  Vincent Semeria (2018)
```
Zimmi48: You are supposed to insert your name in alphabetical order.

Zimmi48 (approved):
Doc change is trivial.

CREDITS
VincentSe: Done

VincentSe:
If I git rebase upstream/master, I cannot git push after, because 
To https://github.com/VincentSe/coq.git
 ! [rejected]            proofRuncountableRebase -> proofRuncountableRebase (non-fast-forward)
error: failed to push some refs to 'https://github.com/VincentSe/coq.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.

I now hesitate to git pull, it messed 2 previous pull requests of mine.

SkySkimmer:
That's normal, you need to push with force `-f` after rewriting history (which rebase does).

VincentSe:
So what do we do about this pull request ? It has been open for about 3 months now, and I think all problems are resolved.

VincentSe:
To prove that there is no surjection nat -> R, we could use that from any surjection nat -> A, we can (at least in classical logic) extract a bijection nat -> A (by skipping all natural numbers which images are already taken). But that theorem is more about natural numbers than real numbers : it should be proved in another file, and probably another pull request.

herbelin:
> Currently, the proof does use both surjectivity and injectivity : when it selects the next two indices in the real interval, it assumes they correspond to real numbers that were not seen before. Hacking the proof to workaround this is about the same as proving the bijection extraction above.

OK, I see. Thanks for the explanation.

> Concerning injections, we could use the current theorem "not bijective" plus the Cantor-Bernstein theorem. I don't know if Cantor-Bernstein is implemented in Coq. Anyway it should be in another file too.

I did that long time ago. I could unbury it if needed (I used classical logic of course).

I don't have more questions, thanks a lot. So, letting @silene assigning if he has time, and, otherwise, I'll do it.

herbelin:
Let me assign. I'll merge on next Thursday, giving some time to @silene to react if he is around.

VincentSe:
Thanks !

PR #8889, opened 2018-10-31
Program hook gives back an obligation substitiution

mattam82:
**Kind:** bug fix / cleanup

This is necessary to allow PR #8601 to go through with Equations.

[program] extend obligation to give back a mapping from obligations to
terms

This is necessary for programs like Equations that call add_definition
and want to later update in their hook some separate datastructures
which refer to the obligations that are defined by Program. We give back
a map from obligation name to a constr defined in the program's universe
state which the hook returns as well. (Obligation names also correspond
to undefined evars in the original terms through Obligations.eterm_obligations).

Using this, I can avoid ucontext_of_aucontext in Equations, allowing PR
 #8601 to go through.

### DIFF
--- a/dev/ci/user-overlays/08889-mattam-program-obl-subst.sh
+++ b/dev/ci/user-overlays/08889-mattam-program-obl-subst.sh
@@ -1,0 +1,6 @@
+if [ "$CI_PULL_REQUEST" = "8889" ] || [ "$CI_BRANCH" = "program-hook-obligation-subst" ]; then
+
+    Equations_CI_REF=program-hook-obligation-subst
+    Equations_CI_GITURL=https://github.com/mattam82/Coq-Equations
+
+fi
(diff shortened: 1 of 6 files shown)


ppedrot (approved):
I am not surprised that such a change is needed. If anything, it looks like the kind of code I'd have written myself trying to fix the Equation side.

ppedrot:
@mattam82 You should add the corresponding Equations PR as an overlay.

mattam82:
Yep. I'll have to fix it first, some parts around about # 8061. Next week, it's vacations now!

mattam82:
Equations overlay added

mattam82:
@ppedrot I think all is in order, bedrock2 and windows64 seem to fail for unrelated reasons.

ppedrot:
Indeed, those two development keep failing for no good reason. I'll merge.

ejgallego:
> Indeed, those two development keep failing for no good reason. I'll merge.

Windows seems to have problems with overlays. But yeah, to me is impossible to understand the logs.

PR #8890, opened 2018-11-01
Looking for notation both before or after removal of top coercion

herbelin:
**Kind:** experiment in prevision of further code simplification

For primitive tokens, in a situation like `Zpos some_constant : Z` and `Zpos` is a removable coercion, there is a heuristic which looks for a notation for `Zpos some_constant` and `some_constant` independently, then prefers the second one unless it requires a delimiter. (This was initially motivated by an example in the four color theorem, see e1c8e5ca2).

We apply this heuristic also for ordinary notations. I don't have realistic examples of application but this will help to do some factorization in the code.

### DIFF
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -716,20 +716,20 @@ let rec flatten_application c = match DAst.get c with
 (* one with no delimiter if possible)                                 *)
 
 let extern_possible_prim_token (custom,scopes) r =
-  try
-    let (sc,n) = uninterp_prim_token r in
-    match availability_of_entry_coercion custom InConstrEntrySomeLevel with
-    | None -> raise No_match
-    | Some coercion ->
-    match availability_of_prim_token n sc scopes with
-    | None -> None
-    | Some key -> Some (insert_coercion coercion (insert_delimiters (CAst.make ?loc:(loc_of_glob_constr r) @@ CPrim n) key))
-  with No_match ->
-    None
+   let (sc,n) = uninterp_prim_token r in
+   match availability_of_entry_coercion custom InConstrEntrySomeLevel with
+   | None -> raise No_match
+   | Some coercion ->
+   match availability_of_prim_token n sc scopes with
+   | None -> raise No_match
+   | Some key -> insert_coercion coercion (insert_delimiters (CAst.make ?loc:(loc_of_glob_constr r) @@ CPrim n) key)
 
-let extern_optimal_prim_token scopes r r' =
-  let c = extern_possible_prim_token scopes r in
-  let c' = if r==r' then None else extern_possible_prim_token scopes r' in
+let extern_possible extern r =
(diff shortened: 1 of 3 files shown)


herbelin:
@ejgallego: would you accept to be the assignee? (This is a minor PR for consistency of treatment of numerical notations and ordinary notations wrt to include a coercion or not while looking for the notation. It goes in the direction of removing the peculiarities of numerical notations wrt to other kinds of notations.)

ejgallego:
> @ejgallego: would you accept to be the assignee? (This is a minor PR for consistency of treatment of numerical notations and ordinary notations wrt to include a coercion or not while looking for the notation. It goes in the direction of removing the peculiarities of numerical notations wrt to other kinds of notations.)

Sure, when in this case, please add a review request to my username so I get notified a PR needs attention.

ejgallego:
This is unfortunately beyond my expertise in notations as I've seldom designed any; to my very limited knowledge it looks OK and I do trust @herbelin here. However, as this doesn't risk a bit problem with conflicts I will let a few days so expert notation users can comment.

@ggonthier @robbertkrebbers @RalfJung @JasonGross want to comment ?

RalfJung:
I cannot say anything about the code, but will this fix the problem of notations not being used when there is a coercion involved? IIRC that is something we ran into with Iris, though I cannot find the bug report now... @robbertkrebbers do you remember?

robbertkrebbers:
I'm afraid I cannot evaluate this MR. Do you have an example that demonstrate what changed exactly?

robbertkrebbers:
> I cannot say anything about the code, but will this fix the problem of notations not being used when there is a coercion involved? 

See below for an example:

```coq
Require Import String.

Inductive expr :=
  | Var : string -> expr
  | Lam : string -> expr -> expr
  | App : expr -> expr -> expr.

Notation Let x e1 e2 := (App (Lam x e2) e1).

Parameter e1 e2 : expr.

Check (Let "x" e1 e2).

Coercion App : expr >-> Funclass.

(* The notation is broken now *)
Check (Let "x" e1 e2).
```

ejgallego:
> See below for an example:

This example works fine with this PR. Is that a positive thing then?

robbertkrebbers:
Fine meaning the last `Check` printing `Let "x" e1 e2`?

If that's the case, could we add this example to the test suite?

Yes, that's a very positive thing.

ejgallego:
> Fine meaning the last `Check` printing `Let "x" e1 e2`?

Yes :)

herbelin:
> If that's the case, could we add this example to the test suite?

Done. Thanks for the example!

ejgallego:
Will merge on Friday if we get no more comments

ejgallego (approved):
Looks good to me. CI failure is spurious. Last 24 hours to comment for those interested.

PR #8894, opened 2018-11-02
Print full binders in subtyping incompatible polymorphism error.

### LINKED ISSUE #8891: "Signature components for label A do not match: polymorphic universe instances do not match." should be more informative
#### Version

8.8.1

Consider this code:
```coq
Set Universe Polymorphism.
Set Printing Universes.
Module Type A.
  Axiom A : Type.
End A.

Module a : A.
  Definition A := Set.
End a.                   

(* Error: Signature components for label A do not match:
  polymorphic universe instances do not match. *)
```

I would like to know the length of both the expected and actual universe instances.  Additionally, it would be nice if this error message would also print the expected and actual types and constraints (sort-of how type errors print the context).

SkySkimmer:
Close #8891

Note that this will print `Var(n)` until names are in aucontext.

### DIFF
--- a/engine/univNames.ml
+++ b/engine/univNames.ml
@@ -36,10 +36,6 @@ type universe_binders = Univ.Level.t Names.Id.Map.t
 
 let empty_binders = Id.Map.empty
 
-let universe_binders_of_global ref : Name.t array =
-  try AUContext.names (Environ.universes_of_global (Global.env ()) ref)
-  with Not_found -> [||]
-
 let name_universe lvl =
   (** Best-effort naming from the string representation of the level. This is
       completely hackish and should be solved in upper layers instead. *)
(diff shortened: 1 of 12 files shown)


ppedrot:
Test-suite looks broken.

ppedrot (approved):
LGTM. I am surprised that the removal of the catch-all error doesn't lead to strange anomalies when printing functors, but doing the latter is already quite dangerous so who cares.

ppedrot:
Last round of comments. @herbelin is probably the most interested one, so let's ping him.

herbelin:
OK to me, but a question. According to the top description, it is still printing `Var`. According to this [comment](https://github.com/coq/coq/pull/8983#issuecomment-438306458), it does not print `Var`. According to the code, "names are in aucontext", so, both comments are consistent (and names are used). Is that correct? And, if yes, what are the remaining situations where names are not used? Is it only when the names are `Anonymous`?

ppedrot:
@SkySkimmer ?

SkySkimmer:
The last commit makes it use names. I don't know when Anonymous happens, ask @ppedrot.

herbelin:
My question was confused sorry. There are two questions:
- Are there still situations where `Var` can mistakenly be used for printing?
- What are the projects wrt to always using names for printing local universes? If I understood correctly the discussion a week or two ago on gitter, @ppedrot said this was one of his long term plan, what answers the question.

PR #8896, opened 2018-11-02
Expose Typing.judge_of_apply

SkySkimmer:
This can be useful to avoid [Typing.type_of (App (f,args))] when
working with universe polymorphism.


ppedrot (approved):

ppedrot:
Trivial and will merge if @mattam82 doesn't complain very quickly.

mattam82 (approved):

Issue #8900, opened 2018-11-04
dune build system should provide a list of convenient targets


ppedrot:
This is discussed in #8729, but I couldn't find an issue that states it explicitly. Please close if I missed it.

In my daily development routine, I need aliases for the following:
- all plugins individually (I don't want to have to explicitly remember which vo file I have to compile)
- states (likewise having to spell out `theories/Init/Prelude.vo` is a big no)
- miniopt / pluginsopt (I don't want to wait for the thermodynamical death of the universe to check that ml files are ok)
- printers
- tools

Other people probably have different needs, we should gather them in this PR.

ejgallego:
Most of the targets should be able to do now [or are already present], however my typical way of working with Dune is a bit different than make, first, because the dev profile uses opaque by default so in most cases way less recompilation is needed, second, because a very typical setup is just a continuous build with `--watch` so I don't need to worry too much other than setting up your editor properly. Of course if battery life is a concern that's another thing.

In fact, I am not sure I like to have such a gazillion of fine-grained targets. They are a bit of a burden to properly maintain unless they correspond to truly logical installable units, and indeed the time gain vs doing a normal build is just not there.

SkySkimmer:
How many cores do you throw at compilation?

Gaëtan Gilbert

ejgallego:
> How many cores do you throw at compilation?

From 8 to 32.

ejgallego:
Personally I would rather prefer targets to be directed by actions, for example "build everything needed to run this part of the test suite" or "build everything so I can launch a toplevel + prelude" , or "build everything so I can load coq.stdlib.streams".

SkySkimmer:
Mostly I want "build every ml so I can fix type errors without wasting my time with the stdlib" ie byte/coqocaml. I tried dune for a bit but without that it was too annoying.

>indeed the time gain vs doing a normal build is just not there.

Disagree. At least with my 4thread laptop cpu.

ejgallego:
> Mostly I want "build every ml so I can fix type errors without wasting my time with the stdlib" ie byte/coqocaml. I tried dune for a bit but without that it was too annoying.

The thing is that defining "every ml file" in a modular build system is not so easy. Do you consider CoqIDE? All the plugins? All the ml files tested in the CI.

What happens if I add a random `kernel/foo.ml` that nothing is using? Why should it be build?

Anyways, I guess we will implement this kind of target by manually listing the set of `cma` to be build and be done with it.

[...]

PR #8901, opened 2018-11-05
[dune] Add "quick" and "check" targets for fast builds.

ejgallego:
In #8900 a quicker build target is requested, this PR does provide
targets towards that goal.

- `check`, introduced in Dune 1.5.0, will build all ml files in a fast way,
- `quick`, does build a set of hand-picked targets in a fast way.

Further improvements could come from having `coq_dune` use
`coqtop.byte` when appropriate, taking advantage from
https://github.com/ocaml/dune/issues/1073, and not generating rules
for `.vo` files.

### DIFF
--- a/Makefile.dune
+++ b/Makefile.dune
@@ -1,7 +1,7 @@
 # -*- mode: makefile -*-
 # Dune Makefile for Coq
 
-.PHONY: help voboot states world watch test-suite release apidoc ocheck ireport clean
+.PHONY: help voboot states world watch check quickbyte quickopt test-suite release apidoc ocheck ireport clean
 
 # use DUNEOPT=--display=short for a more verbose build
 # DUNEOPT=--display=short
@@ -13,6 +13,9 @@ help:
 	@echo "  - states:  build a minimal functional coqtop"
 	@echo "  - world:   build all binaries and libraries"
 	@echo "  - watch:   build all binaries and libraries [continuous build]"
+	@echo "  - check:   build all ML files as fast as possible [requires Dune >= 1.5.0]"
+	@echo "  - quickbyte:  build main ML files [coqtop + plugins + ide + printers] using the bytecode compiler"
+	@echo "  - quickopt:   build main ML files [coqtop + plugins + ide + printers] using the optimizing compiler"
 	@echo "  - test-suite: run Coq's test suite"
 	@echo "  - release: build Coq in release mode"
 	@echo "  - apidoc:  build ML API documentation"
@@ -33,6 +36,21 @@ world: voboot
 
 watch: voboot
 	dune build $(DUNEOPT) @install -w
+
+check: voboot
+	dune build $(DUNEOPT) @check
+
+COQTOP_FILES=ide/idetop.bc ide/coqide_main.bc checker/coqchk.bc
+PLUGIN_FILES=$(wildcard plugins/*/*.mlpack)
+PRINTER_FILES=dev/top_printers.cma dev/checker_printers.cma
+QUICKBYTE_TARGETS=$(COQTOP_FILES) $(PLUGIN_FILES:.mlpack=.cma) $(PRINTER_FILES) topbin/coqtop_byte_bin.bc
+QUICKOPT_TARGETS=$(COQTOP_FILES:.bc=.exe) $(PLUGIN_FILES:.mlpack=.cmxa) $(PRINTER_FILES:.cma=.cmxa) topbin/coqtop_bin.exe
+
+quickbyte: voboot
+	dune build $(DUNEOPT) $(QUICKBYTE_TARGETS)
+
+quickopt: voboot
+	dune build $(DUNEOPT) $(QUICKOPT_TARGETS)
 
 test-suite: voboot
 	dune $(DUNEOPT) runtest


SkySkimmer (changes requested):
Thanks!

AFAICT the docker update is unecessary (as CI doesn't test the new targets) and puts us out of sync with our declared dependencies (eg in .opam files).
I think you should remove the docker changes and put "requires dune >= 1.5.0" in the make help for check.
You could also bump our required dune version throughout, although I'm not sure it's worth doing here since only the new feature will break without updated dune.

gares:
> `quick`, does build a set of hand-picked targets in a fast way.

Can you explain which targets are built and why this set is relevant? Which is the rationale/use-case?
What can I do with these things (re)built? Can I run ocamldebug? Can I load plugins?

ejgallego:
> Can you explain which targets are built and why this set is relevant? Which is the rationale/use-case?

These targets are mainly derived from what it is already on the Makefiles in different forms. For example `pluginsopts`, etc... I added them because some developers requested it, I don't use them myself. AFAIU they compile most of the ML part of Coq.

> What can I do with these things (re)built?

Run them?

> Can I run ocamldebug? Can I load plugins?

Yup, that should work. Tho for ocamldebug you have the `dune-dbg` target that should work better. In the future Dune will support running ocamldebug directly I think.

ejgallego:
> AFAICT the docker update is unecessary (as CI doesn't test the new targets) and puts us out of sync with our declared dependencies (eg in .opam files).

Yup that's a problem but hopefully soon Dune will generate the opam files and take care of adding the right deps there.

> I think you should remove the docker changes and put "requires dune >= 1.5.0" in the make help for check.
> You could also bump our required dune version throughout, although I'm not sure it's worth doing here since only the new feature will break without updated dune.

I don't mind either way, Dune is released quite often so updates are expected.

gares:
> These targets are mainly derived from what it is already on the Makefiles in different forms. For example `pluginsopts`

pluginsopt builds native code, here I see `.bc` stuff.
Unfortunately, for my workflow, byte code is unusable (and ocamldebug too).

In any case I'm fine with a target for a minimal bytecode build. But you should find a name that says so. quick is a bit too vague, IMO. For sure it should mention "bytecode".

ejgallego:
> I don't mind either way, Dune is released quite often so updates are expected.

I have removed the 1.5.0 requirement.

> pluginsopt builds native code, here I see `.bc` stuff.
> Unfortunately, for my workflow, byte code is unusable (and ocamldebug too).
> 
> In any case I'm fine with a target for a minimal bytecode build. But you should find a name that says so. quick is a bit too vague, IMO. For sure it should mention "bytecode".

Oh indeed, I've added a `quickbyte` and `quickopt` then.

Makefile.dune
```diff
@@ -34,6 +37,21 @@ world: voboot
...
 	dune build $(DUNEOPT) @install -w
 
+check: voboot
+	dune build $(DUNEOPT) @check
+
+COQTOP_FILES=ide/idetop.bc ide/coqide_main.bc checker/coqchk.bc
+PLUGIN_FILES=$(wildcard plugins/*/*.mlpack)
+PRINTER_FILES=dev/top_printers.cma dev/checker_printers.cma
+QUICKBYTE_TARGETS=$(COQTOP_FILES) $(PLUGIN_FILES:.mlpack=.cma) $(PRINTER_FILES) topbin/coqtop_byte_bin.bc
+QUICKOPT_TARGETS=$(COQTOP_FILES:.bc=.exe) $(PLUGIN_FILES:.mlpack=.cmxa) $(PRINTER_FILES:.cma=.cmxa) topbin/coqtop_bin.exe
```
gares: I'm not so sure that PRINTER_FILES in opt are of any use. They are typically loaded the the ocaml toplevel for getting better `#trace` display.

ejgallego: They are included in `ocamlopt` which is what this target emulates. Maybe I should call it this way? Anyways, it is important to compile them and they take very little time.

SkySkimmer (approved):

Makefile.dune
```diff
@@ -1,7 +1,7 @@
 # -*- mode: makefile -*-
 # Dune Makefile for Coq
 
-.PHONY: help voboot states world watch test-suite release apidoc ocheck ireport clean
+.PHONY: help voboot states world watch check quick test-suite release apidoc ocheck ireport clean
```
Zimmi48: This has gone out of sync with the rest of the patch.

Zimmi48 (commented):
Arf, I thought that by agreeing to be secondary code owner for dune files, it would prevent me having to read any Makefile-syntax and already this is wrong. Can @gares or @SkySkimmer self-assign this PR?

SkySkimmer:
@Zimmi48 a bit of make won't kill you, but whatever I'll do it.

ejgallego:
> I thought that by agreeing to be secondary code owner for dune files, it would prevent me having to read any Makefile-syntax and already this is wrong.

This is a very good point, I could implement the rules using Dune syntax, but I didn't do so as I thought this would be more familiar to devs.

SkySkimmer:
>This is a very good point, I could implement the rules using Dune syntax, but I didn't do so as I thought this would be more familiar to devs.

I think implementing them in dune syntax would help devs learn from example how to write their own rules.

ejgallego:
> I think implementing them in dune syntax would help devs learn from example how to write their own rules.

A bit of the idea of Dune is that devs should not have to write any rules, so I am not sure we want to add a gazillion Dune targets, but indeed, I'll give it a try.

ejgallego:
> A bit of the idea of Dune is that devs should not have to write any rules, so I am not sure we want to add a gazillion Dune targets, but indeed, I'll give it a try.

Well, basically the Dune rule would list all the targets, so it is pretty boring:
```
(alias
  (name plugins)
  (deps plugin1 ... pluginn))
```
There is no mechanism to generate stuff like is done in make, and indeed there are technical reasons for that. When some dynamic rules are needed usually `include` is used with `promote`.

So I am not sure what to here at this stage of the Dune implementation. Maybe make would be the most convenient solution for now.

SkySkimmer:
OK then fix the PHONY then I'll merge.

ejgallego:
Fixed thanks!

PR #8904, opened 2018-11-05
Fix #5351: Intuition does not work with universe polymorphism.

### LINKED ISSUE #5351: Intuition does not simplify P -> Q, P in hypotheses with universe polymorphism

ppedrot:
Instead of using non-linear pattern-matching in tauto, we use constr_eq so as to add the missing universe constraints.

Fixes / closes #5351.


ppedrot:
So this is broken because it turns out Ltac non-linear pattern-matching also performs conversion rather than mere syntactic equality... Apart from the fact this is crazy, there is no way to access conversion directly in Ltac currently.

ppedrot:
Let's retry with this variant.

PR #8907, opened 2018-11-05
Cleanup camlp5 dead code

ppedrot:
This removes quite a bit of the grammar engine code that is not used in Coq. Most of it is related to quotation features of camlp5 that are meaningless as a mere grammar engine.

I am not sure of the policy regarding API compatibility w.r.t. camlp5. This does not change the API in the intersection, but it does remove code. As we already did that before (e.g. removing the backtracking parser) I figured it was OK.


ejgallego:
> I am not sure of the policy regarding API compatibility w.r.t. camlp5.

For now we should remain 100% compatible, but only API-wise / observationally. So I think this will be Ok.

ejgallego (approved):

Issue #6427, opened 2017-12-14
Undocumented options


SkySkimmer:
List after #6219:
```
Debug Ltac
Debug Ssreflect
Debug SsrMatching
Debug Tactic Unification
Debug Typeclasses
Debug Unification
Default Clearing Used Hypotheses
Default Proof Mode
Dependent Propositions Elimination (rm 8.8)
Discriminate Introduction (rm 8.8)
Dump Bytecode
Extraction AccessOpaque
Extraction File Comment
Extraction Flag
Extraction name
Extraction TypeExpand
Functional Induction Rewrite Dependent
Function_debug
Function_raw_tcc
Injection L2R Pattern Order
Inline Level
Keep Admitted Variables
Kernel Term Sharing
Keyed Unification
Lia Depth
Lia Enum
Lra Depth
NativeCompute Profile Filename
Omega OldStyle
Printing Goal Names
Printing Goal Tags
Program Mode
Rtauto Check
Rtauto Pruning
Rtauto Verbose
SimplIsCbn
Solve Unification Constraints
SsrHave NoTCResolution
SsrIdents
SsrMatchingProfiling
SsrOldRewriteGoalsOrder
SsrProfiling
SsrRewrite
Standard Proposition Elimination Names (rm 8.8)
Tactic Pattern Unification (rm 8.8)
Typeclasses Depth
Typeclasses Iterative Deepening
```

Zimmi48:
Duplicated in #8380.

PR #8911, opened 2018-11-06
Document undocumented flags and options

jfehrle:
**Kind:** documentation
See https://github.com/coq/coq/issues/6427 and https://github.com/coq/coq/issues/8380

As part of improving the Coq documentation, we'd like to document undocumented flags and options.

Here's the list of undocumented flags and options.  I hope that the authors or codeowners are willing and able to provide reasonable descriptions as comments on the PR, which I will edit for typos/format/language and add to the PR.  If you think some of these should be removed or kept undocumented, please say so.   (@Zimmi48 and @SkySkimmer favor documenting every flag and option; I'm less certain.)

I added documentation for the few undocumented flags and options that I could understand well enough from the code to describe.  These are in the second (smaller) table.

Name|File|Author (?)|Codeowners
-|-|-|-
Functional Induction Rewrite Dependent | plugins/funind/ indfun_common.ml | @forestjulien | @forestjulien @Matafou
Function_debug | plugins/funind/ indfun_common.ml | @forestjulien | @forestjulien @Matafou
Function_raw_tcc | plugins/funind/ indfun_common.ml | @forestjulien | @forestjulien @Matafou
Lia Depth @num | plugins/micromega/ coq_micromega.ml |@fajb | @coq/micromega-maintainers
Lia Enum | plugins/micromega/ coq_micromega.ml |@fajb | @coq/micromega-maintainers 
Lra Depth @num | plugins/micromega/ coq_micromega.ml |@fajb | @coq/micromega-maintainers
Omega OldStyle | plugins/omega/ coq_omega.ml | @backtracking | unassigned
Rtauto Check | plugins/rtauto/ refl_tauto.ml | @PierreCorbineau | @PierreCorbineau @herbelin
Rtauto Pruning | plugins/rtauto/ proof_search.ml | @PierreCorbineau | @PierreCorbineau @herbelin
Rtauto Verbose | plugins/rtauto/ refl_tauto.ml ml | @PierreCorbineau | @PierreCorbineau @herbelin
Default Clearing Used Hypotheses | tactics/tactics.ml | @herbelin | @ppedrot @mattam82
Dump Bytecode | vernac/vernacentries.ml | ? | @mattam82 @maximedenes
Dump Lambda | vernac/vernacentries.ml | ? | @mattam82 @maximedenes
Keep Admitted Variables | vernac/lemmas.ml | @ppedrot | @mattam82 @maximedenes
Kernel Term Sharing | vernac/lemmas.ml | @maximedenes | @mattam82 @maximedenes
Printing Goal Names | printing/printer.ml | @Matafou ? | @herbelin @mattam82
Printing Goal Tags | printing/printer. ml | @Matafou ? | @herbelin @mattam82
SimplIsCbn | proofs/redexpr.ml | @pirbo  | @ppedrot @Zimmi48
Default Proof Mode @string | proofs/proof_global.ml | ? | @ppedrot @Zimmi48
Inline Level @num	| vernac/vernacentries.ml | @letouzey | @mattam82 @maximedenes

The ones I understood and added text for are:

Name|File|Author (?)|Codeowners
-|-|-|-
Debug Ssreflect | plugins/ssr/ssrprinters.ml | ? | @gares @maximedenes
Debug SsrMatching | plugins/ssrmatching/ssrmatching.ml | @gares | @gares @maximedenes
Extraction File Comment @string | plugins/extraction/table.ml | @aspiwack  | @maximedenes
Extraction Flag @num | plugins/extraction/table.ml | @letouzey  | @maximedenes
Extraction TypeExpand | plugins/extraction/table.ml | @letouzey  | @maximedenes

The text for `Extraction Flag` and `Extraction TypeExpand` need some revision where I probably guessed poorly.

### DIFF
--- a/doc/sphinx/addendum/extraction.rst
+++ b/doc/sphinx/addendum/extraction.rst
@@ -422,7 +422,7 @@ Additional settings
    :name: Extraction Flag
 
    Controls which optimizations are used during extraction, providing a finer-grained
-   control than :flag:`Extraction Optimize`.  The bits of *num* are used as a bit mask.
+   control than :flag:`Extraction Optimize`.  The bits of :token:`num` are used as a bit mask.
    Keeping an option off keeps the extracted ML more similar to the Coq term.
    Values are:
 
@@ -454,7 +454,7 @@ Additional settings
 
 .. flag:: Extraction TypeExpand
 
-   If set, fully expand Coq types in ML.  **see type_expand in mlutil.ml**
+   If set, fully expand Coq types in ML.  See the Coq source code to learn more.
 
 Differences between |Coq| and ML type systems
 ----------------------------------------------
--- a/doc/sphinx/addendum/extraction.rst
+++ b/doc/sphinx/addendum/extraction.rst
@@ -164,7 +164,7 @@ The type-preserving optimizations are controlled by the following |Coq| options:
 .. cmd:: Extraction Inline {+ @qualid }
 
    In addition to the automatic inline feature, the constants
-   mentionned by this command will always be inlined during extraction.
+   mentioned by this command will always be inlined during extraction.
 
 .. cmd:: Extraction NoInline {+ @qualid }
 
@@ -409,6 +409,52 @@ It is possible to instruct the extraction not to use particular filenames.
 
 For |OCaml|, a typical use of these commands is
 ``Extraction Blacklist String List``.
+
+Additional settings
+~~~~~~~~~~~~~~~~~~~
+
+.. opt:: Extraction File Comment @string
+   :name: Extraction File Comment
+
+   Provides a comment that is included at the beginning of the output files.
+
+.. opt:: Extraction Flag @num
+   :name: Extraction Flag
+
+   Controls which optimizations are used during extraction, providing a finer-grained
+   control than :flag:`Extraction Optimize`.  The bits of *num* are used as a bit mask.
+   Keeping an option off keeps the extracted ML more similar to the Coq term.
+   Values are:
+
+   +-----+-------+----------------------------------------------------------------+
+   | Bit | Value | Optimization (default is on unless noted otherwise)            |
+   +-----+-------+----------------------------------------------------------------+
+   |   0 |    1  | Remove local dummy variables                                   |
+   +-----+-------+----------------------------------------------------------------+
+   |   1 |    2  | Use special treatment for fixpoints                            |
+   +-----+-------+----------------------------------------------------------------+
+   |   2 |    4  | Simplify case with iota-redux                                  |
+   +-----+-------+----------------------------------------------------------------+
+   |   3 |    8  | Factor case branches as functions                              |
+   +-----+-------+----------------------------------------------------------------+
+   |   4 |   16  | (not available, default false)                                 |
+   +-----+-------+----------------------------------------------------------------+
+   |   5 |   32  | Simplify case as function of one argument                      |
+   +-----+-------+----------------------------------------------------------------+
+   |   6 |   64  | Simplify case by swapping case and lambda                      |
+   +-----+-------+----------------------------------------------------------------+
+   |   7 |  128  | Some case optimization                                         |
+   +-----+-------+----------------------------------------------------------------+
+   |   8 |  256  | Push arguments inside a letin                                  |
+   +-----+-------+----------------------------------------------------------------+
+   |   9 |  512  | Use linear let reduction (default false)                       |
+   +-----+-------+----------------------------------------------------------------+
+   |  10 | 1024  | Use linear beta reduction (default false)                      |
+   +-----+-------+----------------------------------------------------------------+
+
+.. flag:: Extraction TypeExpand
+
+   If set, fully expand Coq types in ML.  **see type_expand in mlutil.ml**
 
 Differences between |Coq| and ML type systems
 ----------------------------------------------
--- a/doc/sphinx/addendum/type-classes.rst
+++ b/doc/sphinx/addendum/type-classes.rst
@@ -556,7 +556,8 @@ Settings
 
    Determines how much information is shown for typeclass resolution steps during search.
    1 is the default level.  2 shows additional information such as tried tactics and shelving
-   of goals.  Setting this option also sets :flag:`Typeclasses Debug`.
+   of goals.  Setting this option to 1 or 2 turns on :flag:`Typeclasses Debug`; setting this
+   option to 0 turns that option off.
 
 .. flag:: Refine Instance Mode
 
@@ -579,7 +580,7 @@ Typeclasses eauto `:=`
    resolution tactic. The semantics of the options are:
 
    + ``debug`` In debug mode, the trace of successfully applied tactics is
-     printed.  This value also be set with :flag:`Typeclasses Debug`.
+     printed.  This value can also be set with :flag:`Typeclasses Debug`.
 
    + ``dfs, bfs`` This sets the search strategy to depth-first search (the
      default) or breadth-first search.  This value can also be set with
--- a/doc/sphinx/addendum/type-classes.rst
+++ b/doc/sphinx/addendum/type-classes.rst
@@ -461,12 +461,12 @@ type, like:
 This is equivalent to ``Hint Transparent, Opaque ident : typeclass_instances``.
 
 
-Options
-~~~~~~~
+Settings
+~~~~~~~~
 
 .. flag:: Typeclasses Dependency Order
 
-   This option (on by default since 8.6) respects the dependency order
+   This flag (on by default since 8.6) respects the dependency order
    between subgoals, meaning that subgoals on which other subgoals depend
    come first, while the non-dependent subgoals were put before
    the dependent ones previously (Coq 8.5 and below). This can result in
@@ -475,7 +475,7 @@ Options
 
 .. flag:: Typeclasses Filtered Unification
 
-   This option, available since Coq 8.6 and off by default, switches the
+   This flag, available since Coq 8.6 and off by default, switches the
    hint application procedure to a filter-then-unify strategy. To apply a
    hint, we first check that the goal *matches* syntactically the
    inferred or specified pattern of the hint, and only then try to
@@ -483,13 +483,13 @@ Options
    improve performance by calling unification less often, matching
    syntactic patterns being very quick. This also provides more control
    on the triggering of instances. For example, forcing a constant to
-   explicitely appear in the pattern will make it never apply on a goal
+   explicitly appear in the pattern will make it never apply on a goal
    where there is a hole in that place.
 
 
 .. flag:: Typeclasses Limit Intros
 
-   This option (on by default) controls the ability to apply hints while
+   This flag (on by default) controls the ability to apply hints while
    avoiding (functional) eta-expansions in the generated proof term. It
    does so by allowing hints that conclude in a product to apply to a
    goal with a matching product directly, avoiding an introduction.
@@ -503,16 +503,16 @@ Options
 
 .. flag:: Typeclass Resolution For Conversion
 
-   This option (on by default) controls the use of typeclass resolution
+   This flag (on by default) controls the use of typeclass resolution
    when a unification problem cannot be solved during elaboration/type
-   inference. With this option on, when a unification fails, typeclass
+   inference. With this flag on, when a unification fails, typeclass
    resolution is tried before launching unification once again.
 
 
 .. flag:: Typeclasses Strict Resolution
 
-   Typeclass declarations introduced when this option is set have a
-   stricter resolution behavior (the option is off by default). When
+   Typeclass declarations introduced when this flag is set have a
+   stricter resolution behavior (the flag is off by default). When
    looking for unifications of a goal with an instance of this class, we
    “freeze” all the existentials appearing in the goals, meaning that
    they are considered rigid during unification and cannot be
@@ -528,15 +528,28 @@ Options
 
 .. flag:: Typeclasses Unique Instances
 
-   Typeclass declarations introduced when this option is set have a more
-   efficient resolution behavior (the option is off by default). When a
+   Typeclass declarations introduced when this flag is set have a more
+   efficient resolution behavior (the flag is off by default). When a
    solution to the typeclass goal of this class is found, we never
    backtrack on it, assuming that it is canonical.
+
+.. flag:: Typeclasses Iterative Deepening
+
+   When this flag is set, the proof search strategy is breadth-first search.
+   Otherwise, the search strategy is depth-first search.  The default is off.
+   :cmd:`Typeclasses eauto` is another way to set this flag.
+
+.. opt:: Typeclasses Depth @num
+   :name: Typeclasses Depth
+
+   Sets the maximum proof search depth.  The default is unbounded.
+   :cmd:`Typeclasses eauto` is another way to set this option.
 
 .. flag:: Typeclasses Debug
 
    Controls whether typeclass resolution steps are shown during search.  Setting this flag
-   also sets :opt:`Typeclasses Debug Verbosity` to 1.
+   also sets :opt:`Typeclasses Debug Verbosity` to 1.  :cmd:`Typeclasses eauto`
+   is another way to set this flag.
 
 .. opt:: Typeclasses Debug Verbosity @num
    :name: Typeclasses Debug Verbosity
@@ -547,7 +560,7 @@ Options
 
 .. flag:: Refine Instance Mode
 
-   This option allows to switch the behavior of instance declarations made through
+   This flag allows to switch the behavior of instance declarations made through
    the Instance command.
 
    + When it is on (the default), instances that have unsolved holes in
@@ -560,14 +573,17 @@ Typeclasses eauto `:=`
 ~~~~~~~~~~~~~~~~~~~~~~
 
 .. cmd:: Typeclasses eauto := {? debug} {? {dfs | bfs}} depth
+   :name: Typeclasses eauto
 
    This command allows more global customization of the typeclass
    resolution tactic. The semantics of the options are:
 
    + ``debug`` In debug mode, the trace of successfully applied tactics is
-     printed.
+     printed.  This value also be set with :flag:`Typeclasses Debug`.
 
    + ``dfs, bfs`` This sets the search strategy to depth-first search (the
-     default) or breadth-first search.
+     default) or breadth-first search.  This value can also be set with
+     :flag:`Typeclasses Iterative Deepening`.
 
-   + ``depth`` This sets the depth limit of the search.
+   + ``depth`` This sets the depth limit of the search.  This value can also be set with
+     :opt:`Typeclasses Depth`.
--- a/doc/sphinx/proof-engine/ltac.rst
+++ b/doc/sphinx/proof-engine/ltac.rst
@@ -1208,7 +1208,7 @@ Interactive debugger
 
 .. flag:: Ltac Debug
 
-   This option governs the step-by-step debugger that comes with the |Ltac| interpreter
+   This option governs the step-by-step debugger that comes with the |Ltac| interpreter.
 
 When the debugger is activated, it stops at every step of the evaluation of
 the current |Ltac| expression and prints information on what it is doing.
--- a/doc/sphinx/proof-engine/ssreflect-proof-language.rst
+++ b/doc/sphinx/proof-engine/ssreflect-proof-language.rst
@@ -167,8 +167,14 @@ compatible with the rest of |Coq|, up to a few discrepancies:
 
 .. flag:: SsrIdents
 
-   Controls whether reserved identifiers can appear in scripts.  The default
-   is "on".  "off" is compatible with other parts of Coq.
+   Controls whether tactics can refer to |SSR|-generated variables that are
+   in the form _xxx_.  Scripts with explicit references to such variables
+   are fragile; they are prone to failure if the proof is later modified or
+   if the details of variable name generation change in future releases of Coq.
+
+   The default is on, which gives an error message when the user tries to
+   create such identifiers.  Disabling the flag generates a warning instead,
+   increasing compatibility with other parts of Coq.
 
 |Gallina| extensions
 --------------------
@@ -3071,10 +3077,10 @@ An :token:`r_item` can be:
 
    Controls the order in which generated subgoals (side conditions)
    are added to the
-   proof context.  "off", the default value, puts subgoals generated
-   by conditional rules first, followed by the main goal.  "on" puts
-   the main goal first.  If your proofs are organized to complete
-   proving the main goal before side conditions, "on" will save you
+   proof context.  The flag is off by default, which puts subgoals generated
+   by conditional rules first, followed by the main goal.  When it is on,
+   the main goal appears first.  If your proofs are organized to complete
+   proving the main goal before side conditions, turning the flag on will save you
    from having to add :tacn:`last first` tactics that would be needed
    to keep the main goal as the currently focused goal.
 
--- a/doc/sphinx/proof-engine/ssreflect-proof-language.rst
+++ b/doc/sphinx/proof-engine/ssreflect-proof-language.rst
@@ -157,14 +157,18 @@ compatible with the rest of |Coq|, up to a few discrepancies:
   (see :ref:`pattern_conditional_ssr`).  To use the
   generalized form, turn off the |SSR| Boolean ``if`` notation using the command:
   ``Close Scope boolean_if_scope``.
-+ The following two options can be unset to disable the incompatible
-  rewrite syntax and allow reserved identifiers to appear in scripts.
++ The following flags can be unset to make |SSR| more compatible with
+  parts of Coq:
 
-  .. coqtop:: in
+.. flag:: SsrRewrite
 
-     Unset SsrRewrite.
-     Unset SsrIdents.
+   Controls whether the incompatible rewrite syntax is enabled (the default).
+   Disabling the flag makes the syntax compatible with other parts of Coq.
 
+.. flag:: SsrIdents
+
+   Controls whether reserved identifiers can appear in scripts.  The default
+   is "on".  "off" is compatible with other parts of Coq.
 
 |Gallina| extensions
 --------------------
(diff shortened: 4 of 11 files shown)


jashug:
The options `Mangle Names` and `Mangle Names Prefix` are documented in [the practical tools -> coq commands chapter](https://github.com/coq/coq/blame/9134d94d42bafd38dfcc6a09a99edd554e636b55/doc/sphinx/practical-tools/coq-commands.rst#L201), though not using the right sphinx syntax to generate an index entry. They are in a bit of an odd place, but perhaps someone more familiar with the doc system can tag them in the right way?

jfehrle:
@jashug I'll add the right formatting for the `Mangle` commands but they should be defined in another part of the manual.  If you have or @Zimmi48 have a good idea where, please let me know.

Zimmi48:
I don't have a good suggestion for a better location for the mangle documentation unfortunately...

Zimmi48:
- [`Printing Allow Match Default Clause`](https://coq.inria.fr/refman/language/gallina-extensions.html#coq:flag.printing-allow-match-default-clause) and [`Printing Records`](https://coq.inria.fr/refman/language/gallina-extensions.html#coq:flag.printing-allow-match-default-clause) are documented.
- `SsrIdents` and `SsrRewrite` are documented in https://coq.inria.fr/refman/proof-engine/ssreflect-proof-language.html#compatibility-issues but not in the proper format.
- [`Asymmetric Pattern`](https://coq.inria.fr/refman/addendum/extended-pattern-matching.html#coq:flag.asymmetric-patterns) is documented.
- Some of these options are documented only in CHANGES (which is not good enough) like `Typeclasses Axioms Are Instances`.
- [`Typeclasses Debug Verbosity @num`](https://coq.inria.fr/refman/addendum/type-classes.html#coq:opt.typeclasses-debug-verbosity) is documented.

doc/sphinx/addendum/type-classes.rst
```diff
@@ -527,7 +527,9 @@ Options
 
 .. flag:: Typeclasses Debug
 
-   Controls whether typeclass resolution steps are shown during search.  Setting this flag
+.. flag:: Debug Typeclasses
+
+   Two names for the same option. Controls whether typeclass resolution steps are shown during search.  Setting this flag
```
Zimmi48: One of these two names is marked as deprecated: remove it instead of documenting it.

doc/sphinx/proof-engine/ltac.rst
```diff
@@ -1200,7 +1200,10 @@ Interactive debugger
 
 .. flag:: Ltac Debug
 
-   This option governs the step-by-step debugger that comes with the |Ltac| interpreter
+.. flag:: Debug Ltac
+
+   Two names for the same option, which governs the step-by-step
+   debugger that comes with the |Ltac| interpreter
```
Zimmi48: One of these two names should probably get deprecated while the other is kept and documented. Given that `Ltac Debug` was documented, I'd favor deprecating `Debug Ltac`.

Zimmi48:
`Congruence Depth @num` is marked as deprecated and a comment indicate this is a no-op. So it should be removed.

gares:
`Set Debug Ssr*` enable some debug prints. Their quality and granularity varies. Prints are not addressed to users, eg a print like `c : bla bla` says that variable `c` holds `bla bla` (where `c` in a ML variable name). I like to keep this code around since it has little performance impact but helps debugging (without recompiling). Unless one takes care of making the prints understandable to non devs, I'm not so sure it is worth documenting these.

`Set SsrMatchingProfiling` should enable some profiling and the data be printed on stdout when Coq exits. I think this should be removed along with the profiling code.

`Set Proof Using Clear Unused` is not documented because it its effect is problematic, see # 883 and #6254.
I'm fine removing this option.

I think it is a good occasion to clean up flags.

About documenting each and every flag, I'm a bit skeptical. At least the ones that I'm supposed to document are largely dev only. Maybe we should have a way to tag them and don't this class of flags to be documented (maybe just listed as such).

Zimmi48:
`Typeclasses Iterative Deepening` sets the proof search strategy to breadth-first search instead of depth-first search; `Typeclasses Depth @num` changes the proof search depth (unbounded by default). These options can also be set through the more esoteric [`Typeclasses eauto := bfs | dfs @num`](https://coq.inria.fr/refman/addendum/type-classes.html#coq:cmd.typeclasses-eauto).

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -5428,6 +5428,18 @@ right hand side double , view hint declaration see :ref:`declaring_new_hints_ssr
...
+~~~~~~~~
+
+.. flag:: Debug Ssreflect
+
+   Print debug information on reflect to standard output.
+
+.. flag:: Debug SsrMatching
+
+   Print debug information on SSR matching to standard output.  Has no effect
+   unless the environment variable `SSRMATCHINGDEBUG` exists.
```
gares: This note about the env variable is just a bug in the code. I'm pretty sure I wanted to be able to enable the option via an env variable, I guess I should have used a disjunction rather than a conjunction. I'm fine removing this (broken) feature.

Also, I would not be so specific about where prints are performed, eg I see these debug prints in a CoqIDE window too.

doc/sphinx/addendum/extraction.rst
```diff
@@ -409,6 +409,54 @@ It is possible to instruct the extraction not to use particular filenames.
...
+   |   6 |   64  | Simplify case by swapping case and lambda                      |
+   +-----+-------+----------------------------------------------------------------+
+   |   7 |  128  | Some case optimization                                         |
+   +-----+-------+----------------------------------------------------------------+
+   |   8 |  256  | Push arguments inside a letin                                  |
+   +-----+-------+----------------------------------------------------------------+
+   |   9 |  512  | Use linear let reduction (default false)                       |
+   +-----+-------+----------------------------------------------------------------+
+   |  10 | 1024  | Use linear beta reduction (default false)                      |
+   +-----+-------+----------------------------------------------------------------+
```
gares: I believe this "API" was not meant for the user.
I've no idea if it is a good idea to expose this, but if we do I suggest we add to `goptions.ml` support for `string list` and we give names, rather than *bits*, to these options. I mean, even in `C` I would combined `#define`d name via `|`. Here I even have to do the sum by hand, in decimal.

doc/sphinx/addendum/type-classes.rst
jfehrle: Neither one is currenlty marked as deprecated.  I will remove "Debug Typeclasses" from the rst and mark it as deprecated.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
jfehrle: I was mistaken; the environment variable works correctly.  I'll keep it in the code but remove the mention in the doc unless you suggest otherwise.

jfehrle:
> `Set Ssr*Profiling` should enable some profiling and the data be printed on stdout when Coq exits. I think this should be removed along with the profiling code.

@Zimmi48 @gares It's OK to delete undocumented commands without first deprecating them for one release?

gares:
I'd say so

doc/sphinx/addendum/extraction.rst
jfehrle: I agree that having to compute a sum here is a poor design, but I'd rather not take on any more than the simplest code changes as part of this PR.  For this particular example, I have no idea what names would be good to use.

jfehrle:
@gares @Zimmi48 @jashug I addressed the comments made so far, please take a look.

jashug (approved):
LGTM.
I think referencing the command line `-mangle-names` option in the tactics chapter might be helpful, but it could be confusing as well. It seems fine the way you have it now.
Thanks for doing this work; it should be very helpful for users.

doc/sphinx/addendum/extraction.rst
```diff
@@ -409,6 +409,54 @@ It is possible to instruct the extraction not to use particular filenames.
...
+
+   **Reviewers: see flag_of_int in table.ml.  I'm guessing here, but still better than nothing.**
```
Zimmi48: The "needs: progress" flag is to reflect the presence of this comment in the code.

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -657,6 +657,12 @@ has type :token:`type`.
    :cmd:`Parameter` and :cmd:`Variable` (and their plural forms) in other cases
    (corresponding to the declaration of an abstract mathematical entity).
 
+.. flag:: Typeclasses Axioms Are Instances
```
Zimmi48: @SkySkimmer Did you deprecate this flag in another PR?

Zimmi48 (dismissed):

Zimmi48:
I am a bit disappointed that not more people answered to your request for help. We could consider merging this PR with what it has currently and opening an issue with the remaining list. WDYT?

Matafou:
I am sorry I don't have time currently to document Function related
options. They are all for debugging purpose only I think.

p.

Le mar. 20 nov. 2018 à 12:13, Théo Zimmermann <notifications@github.com> a
écrit :

mattam82:
- Debug Tactic Unification: prints the traces of unification steps of tactic unification (e.g. used in apply, rewrite)
- Debug Unification: prints the traces of unification steps of unification used during elaboration/typechecking and the `refine` tactic.
- Keyed Unification: makes higher-order unification used by `rewrite` rely on a set of keys to drive unification: the subterms considered as rewriting candidates must start with the same key as the lhs or rhs of the lemma given to rewrite, and the arguments are then unified up-to full reduction. This should be part of the rewrite tactic documentation.
- Program Mode: sets the program mode everywhere: typechecking allows subset coercions and the elaboration of pattern-matching of Program and Fixpoint/Definition act like Program Fixpoint/Definition, generating obligations if there are unresolved holes after typechecking. This should be in the Program chapter.
- Solve Unification Constraints: by default, after each tactic application, postponed typechecking unification problems generated are resolved using heuristics. When unset, this option disables this behavior. Hence tactics can leave unification constraints. One can manually ask for constraints to be solved at any point using the multi-goal tactic `solve_constraints`. This should be somewhere in the tactics chapter I think.

Zimmi48:
>I am sorry I don't have time currently to document Function related options. They are all for debugging purpose only I think.

The request was not to document per se, only to give a short description in a comment, which shouldn't take too long. But indeed, if they are all for debugging purpose, they can be documented as such. Do you think that all of them make sense to keep?

jfehrle:
> We could consider merging this PR with what it has currently and opening an issue with the remaining list. WDYT?

Yeah, I was thinking that myself.  Getting many comments on many settings in a single PR can make for confusion.  I have already removed settings that have been addressed from the table at the top.

jfehrle:
@gares I'm going to remove the `Proof using Clear Unused` option entirely rather than just deprecate it.  However, I'm not sure how much of `Proof_global.set_used_variables` and `Vernacentries.vernac_set_used_variables` should go away.  Please advise, perhaps send me what they should become?  Thanks.

gares:
> I'm not sure how much of `Proof_global.set_used_variables` and `Vernacentries.vernac_set_used_variables` should go away.

I think they should stay.

jfehrle:
If I remove the option, then `to_clear` is always `[]` and both those routines can be simplified: `let to_clear = if !proof_using_auto_clear then to_clear else [] in`.

Also, `aux` in `Proof_global.set_used_variables` can be simplified since the second return value is discarded.

I updated the code to remove `to_clear`.  Please let me know if that change is good and should be made.

jfehrle:
I added @mattam82's comments in a new commit.
@gares, would you check `Proof_global.set_used_variables` and `Vernacentries.vernac_set_used_variables` and approve?

@mattam82 and @ppedrot do you have any review comments?

I'll open a new PR to cover the remaining options, let's get this one completed with what it has.  Thanks.

doc/sphinx/addendum/extraction.rst
jfehrle: I removed the comment directed at reviewers and the "needs: progress" flag, leaving the table unchanged.  It is better than nothing even if may not be completely accurate.  I will take half a loaf here.

gares (approved):
The changes to proofglobal and vernacentries wrt automatic clear are OK to me

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -157,14 +157,20 @@ compatible with the rest of |Coq|, up to a few discrepancies:
...
-+ The following two options can be unset to disable the incompatible
-  rewrite syntax and allow reserved identifiers to appear in scripts.
++ The following flags can be unset to make |SSR| more compatible with
+  parts of Coq:
 
-  .. coqtop:: in
+.. flag:: SsrRewrite
+
+   Controls whether the incompatible rewrite syntax is required.  The default
+   is "on".  "off" makes the syntax compatible with other parts of Coq.
```
Zimmi48: Let me advise once more against using this "on" / "off" terminology which might confuse readers in thinking that these would be options taking strings as input.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -157,14 +157,20 @@ compatible with the rest of |Coq|, up to a few discrepancies:
...
+
+   **Reviewers: "required" or "allowed" above?**
```
Zimmi48: Rather "activated"... ? (thus closer to "allowed" or "supported")

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -264,6 +264,11 @@ Applying theorems
       This tactic behaves like :tacn:`simple refine` except it performs type checking
       without resolution of typeclasses.
 
+   .. flag:: Debug Unification
+
+   Enables printing traces of unification steps used during
```
Zimmi48: Indentation problem.

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -2534,6 +2543,13 @@ simply :g:`t=u` dropping the implicit type of :g:`t` and :g:`u`.
       unresolved bindings into existential variables, if any, instead of
       failing. It has the same variants as :tacn:`rewrite` has.
 
+   .. flag:: Keyed Unification
+
+   Makes higher-order unification used by :tacn:`rewrite` rely on a set of keys to drive
```
Zimmi48: Indentation problem.

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -4503,3 +4519,41 @@ user-defined tactics.
   significant changes in your theories to obtain the same result. As a
   drawback of the re-engineering of the code, this tactic has also been
   completely revised to get a very compact and readable version.
+
+Delaying solving unification constraints
+----------------------------------------
+
+.. tacn:: solve_constraints
+   :name: solve_constraints
```
Zimmi48: Add description or `:undocumented:` option.

doc/sphinx/proof-engine/tactics.rst
jfehrle: I indented it because it appears to apply only to `refine', which is the tactic before this line (about 2 pages).  There are 5 existing flags that have similar indentation, all in this file.  You want to change all of them to be at the left margin?

doc/sphinx/proof-engine/tactics.rst
jfehrle: Same idea as the previous comment.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
gares: On and off are the possible states of a switch. I don't find it confusing.

maximedenes: At least the quotes are confusing. And we avoided this terminology for other entries because it *did* create some confusion for some readers.

gares: Yes, the `"` should not be there.

doc/sphinx/proof-engine/tactics.rst
Zimmi48: I am not talking about the indentation of `.. flag:: Debug Unification` but the indentation of its documentation:
```suggestion
      Enables printing traces of unification steps used during
```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
jfehrle: I made it "enabled".  Presumably readers will grasp that enabled=set and disabled=unset.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
jfehrle: I avoid using phrases such as "when the flag is set" because they are ambiguous.  It could mean both "when the flag has the value `on`" or "when the flag is assigned any value (both `on` or `off``).  `Test` and `Print Options` show the value of flags as `on` or `off`.  As I mentioned, I changed the wording to `enabled`.

jfehrle:
Updated.  Please take another look.  @ppedrot, any comments?  Or ready to merge?

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -157,14 +157,18 @@ compatible with the rest of |Coq|, up to a few discrepancies:
...
+
+   Controls whether the incompatible rewrite syntax is enabled (the default).
+   Disabling the flag makes the syntax compatible with other parts of Coq.
 
-     Unset SsrRewrite.
-     Unset SsrIdents.
+.. flag:: SsrIdents
 
+   Controls whether reserved identifiers can appear in scripts.  The default
+   is "on".  "off" is compatible with other parts of Coq.
```
```suggestion
   is on.  Disabling the flag increases compatibility with other parts of Coq.
```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3063,6 +3067,17 @@ An :token:`r_item` can be:
         rewrite -[f y x]/(y + _).
 
 
+.. flag:: SsrOldRewriteGoalsOrder
+
+   Controls the order in which generated subgoals (side conditions)
+   are added to the
+   proof context.  "off", the default value, puts subgoals generated
```
```suggestion
   proof context.  It is off by default, in which case it puts subgoals generated
```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3063,6 +3067,17 @@ An :token:`r_item` can be:
...
+   by conditional rules first, followed by the main goal.  "on" puts
```
```suggestion
   by conditional rules first, followed by the main goal.  When this flag is on, it puts
```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3063,6 +3067,17 @@ An :token:`r_item` can be:
...
+   the main goal first.  If your proofs are organized to complete
+   proving the main goal before side conditions, "on" will save you
```
```suggestion
   proving the main goal before side conditions, turning this flag on will save you
```

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -606,6 +611,10 @@ Applying theorems
       when the instantiation of a variable cannot be found
       (cf. :tacn:`eapply` and :tacn:`apply`).
 
+.. flag:: Debug Tactic Unification
+
+   Enables printing traces of unification steps in tactic unification.
+   Tactic unification is used in tactics such as :tacn:`apply` and :tacn:`rewrite`.
```
Zimmi48: (also known as "the old unification engine")

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -2097,8 +2106,8 @@ and an explanation of the underlying technique.
    equalities between expressions that appeared in the hypothesis that is
    being processed. By default, no equalities are generated if they
    relate two proofs (i.e. equalities between :n:`@terms` whose type is in sort
```
Zimmi48:
```suggestion
   relate two proofs (i.e. equalities between :token:`term`\s whose type is in sort
```

Zimmi48 (commented):
I don't mind the use of "on" and "off" what I mind is the use of quotes that normally indicate string values. I made a few concrete proposals to get rid of quotes.

doc/sphinx/proof-engine/tactics.rst
jfehrle: > (also known as "the old unification engine")

You're referring to "tactic unification"?  Not sure what change you're suggesting or it's just a side comment.

doc/sphinx/addendum/extraction.rst
```diff
@@ -410,6 +410,52 @@ It is possible to instruct the extraction not to use particular filenames.
...
+   |   8 |  256  | Push arguments inside a letin                                  |
+   +-----+-------+----------------------------------------------------------------+
+   |   9 |  512  | Use linear let reduction (default false)                       |
+   +-----+-------+----------------------------------------------------------------+
+   |  10 | 1024  | Use linear beta reduction (default false)                      |
+   +-----+-------+----------------------------------------------------------------+
+
+.. flag:: Extraction TypeExpand
+
+   If set, fully expand Coq types in ML.  **see type_expand in mlutil.ml**
```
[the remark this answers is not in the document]
jfehrle: Will fix, was a comment directed at reviewers.

doc/sphinx/proof-engine/tactics.rst
Zimmi48: Yes, but this is just a side comment :wink:

doc/sphinx/addendum/type-classes.rst
```diff
@@ -528,15 +528,28 @@ Options
...
+
+   Sets the maximum proof search depth.  The default is unbounded.
+   :cmd:`Typeclasses eauto` is another way to set this option.
+
 .. flag:: Typeclasses Debug
 
    Controls whether typeclass resolution steps are shown during search.  Setting this flag
-   also sets :opt:`Typeclasses Debug Verbosity` to 1.
+   also sets :opt:`Typeclasses Debug Verbosity` to 1.  :cmd:`Typeclasses eauto`
+   is another way to set this flag.
```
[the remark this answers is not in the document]
jfehrle: What I have (below) is not specific enough?  Setting `Typeclasses Debug Verbosity` to 1 turns the option on, also setting to 0 turns it off (I'll add setting to 0).  (Not a very clear design, maybe we should get rid of `Typeclasses Debug`?)

```
Setting this option also sets :flag:`Typeclasses Debug`.
```

jfehrle:
I tried committing a few of your suggestions within Github.  I observe:
1. Each suggestion becomes a separate commit, which is less than ideal.  Solutions would be a) do a pull, squash, push again. or b) maybe github could make a squash function available.
2. You need to pull these before making more changes locally and especially
before force-pushing.  I haven't tried making local changes before pulling
a modified PR from github.  Hopefully that would become a merge rather than
losing the local changes.

jfehrle:
Updated, any more comments?

doc/sphinx/addendum/type-classes.rst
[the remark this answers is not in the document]
Zimmi48: Indeed, I was commenting on the design which I think is bad. `Set Typeclasses Debug Verbosity` should preferably have three levels, 0 (default), 1, and 2. Then `Set Typeclasses Debug` could be defined as a synonym for `Set Typeclasses Debug Verbosity 1.`

Zimmi48:
>maybe github could make a squash function available

Yes, see the "Add suggestion to batch" button.

>Hopefully that would become a merge rather than losing the local changes.

Correct, this is what would happen. `git` wouldn't let you silently lose changes.

Zimmi48 (approved):
Last comment @jfehrle. Sorry for not noticing earlier. After that, I promise to merge promptly. Thanks for the very useful work!

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -157,14 +157,18 @@ compatible with the rest of |Coq|, up to a few discrepancies:
...
+.. flag:: SsrRewrite
+
+   Controls whether the incompatible rewrite syntax is enabled (the default).
+   Disabling the flag makes the syntax compatible with other parts of Coq.
 
-     Unset SsrRewrite.
-     Unset SsrIdents.
+.. flag:: SsrIdents
 
+   Controls whether reserved identifiers can appear in scripts.  The default
```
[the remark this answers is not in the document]
jfehrle: Hmm, I wasn't able to construct an example of this.  `Theorem exists: True.` is not accepted with either setting.  `Theorem Print: True.` is always accepted.  What would be an example?  I suppose it would be in an SSR-specific line (which I'm unfamiliar with)?  I might clarify the wording a bit.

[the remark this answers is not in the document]
Zimmi48: The use of the term "reserved identifier" is probably confusing because it may be understood as "keywords" when in fact, it designate the way that SSReflect names variables that were not given an explicit name:

```coq
Require Import ssreflect.

Unset SsrIdents.

Goal forall (P : Prop), P -> P.
move => *.
apply: _Hyp_.
(* Warning: The name _Hyp_ fits the _xxx_ format used for anonymous variables.
            Scripts with explicit references to anonymous variables are fragile. *)
(* Without the Unset SsrIdents above, this tactic would fail with:
            The identifier _Hyp_ is reserved. *)
```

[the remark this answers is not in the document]
jfehrle: Got it, thanks.  "anonymous variables" is not a great term here because the "anonymous variables" are being assigned names.  So I used other language.  Please merge if agreeable.

I notice that `|Coq|` in the .rst file just becomes `Coq` in the final html document.  Does this construct still do something useful, e.g. uses a different font in the TeX version?  For `|SSR|` I see it replaces it with `SSReflect`.

`Small-scale reflection makes an extensive use of the programming
subset of |Gallina|, |Coq|’s logical specification language.`

Zimmi48:
The `| |` syntax uses the replace rules defined in https://raw.githubusercontent.com/coq/coq/master/doc/sphinx/refman-preamble.rst.

Zimmi48:
The CI failure is the same in master.

PR #8912, opened 2018-11-06
[dune] [coqide] Use copy action instead of (run cp ...)

ejgallego:
This is a bit more portable.

### DIFF
--- a/ide/dune
+++ b/ide/dune
@@ -35,7 +35,7 @@
 (rule
  (targets coqide_os_specific.ml)
  (deps (:in-file coqide_X11.ml.in)) ; TODO support others
- (action (run cp %{in-file} %{targets})))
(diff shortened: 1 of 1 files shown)


Zimmi48 (approved):
Could you go further and remove the targets and deps lines as explained in https://dune.readthedocs.io/en/latest/dune-files.html?highlight=copy#inferred-rules?

ejgallego:
> Could you go further and remove the targets and deps lines as explained in https://dune.readthedocs.io/en/latest/dune-files.html?highlight=copy#inferred-rules?

Yup, but I was way of doing so as this rule will become conditional in the future [I will copy a different file depending on an env variable (`ARCH`) thus it may fall on the case discussed in the doc.

PR #8915, opened 2018-11-06
More cleanup of vendored camlp5

ppedrot:
We remove:
- The non-functorial implementation of grammars (not used and error-prone)
- Almost of Plexing that is just defining types now. It used to ship a crapload of unused utility functions and wrappers.


ejgallego (approved):
LGTM, will merge as soon as CI passes.

PR #8918, opened 2018-11-06
Fix overlays on Windows CI

SkySkimmer:
"${foo+}" is always the empty string
"${foo-}" is "$foo" when foo is set and empty string when it's unset.

### DIFF
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -434,7 +434,7 @@ function build_prep_overlay {
 # ------------------------------------------------------------------------------
 
 function load_overlay_data {
-  if [ -n "${GITLAB_CI+}" ]; then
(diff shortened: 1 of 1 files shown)


SkySkimmer:
It works
~~~
+16:36:34 load_overlay_data
+16:36:34 '[' -n true ']'
+16:36:34 export CI_BRANCH=pr-8918
+16:36:34 CI_BRANCH=pr-8918
...
~~~

MSoegtropIMC (approved):
Thanks, well spotted!

I verified your fix against the Posix spec at

[http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html](url) Section 2.6.2.

Best regards,

Michael

Zimmi48:
@MSoegtropIMC Don't forget to self-assign and merge.

PR #8919, opened 2018-11-06
[vernac] Rename Vernacinterp to Vernacextend and move extension functions there

ejgallego:
This PR fixes an issue that was bugging me for some time, namely that
`Vernacinterp` really means `Vernacextend`.

We thus rename the file and move the associated functions there, which
were incorrectly placed in `Vernacentries`.

Note the beneficial effects on reducing the `.mli` API.


ejgallego:
@ppedrot what is the proper way to fix the `top_printers` ?

ppedrot:
What's the matter?

ejgallego:
> What's the matter?

Two printers there should be moved from vinterp_add to vernac_extend.

ejgallego:
See link https://github.com/coq/coq/blob/011d915ad22e5a97ab2177fbe1301655c2272c96/dev/top_printers.ml#L518

dev/top_printers.ml
```diff
@@ -515,7 +515,7 @@ open Egramml
 
 let _ =
   try
-    Vernacinterp.vinterp_add false ("PrintConstr", 0)
+    Vernacextend.vinterp_add false ("PrintConstr", 0)
```
SkySkimmer: TBH I don't know what this stuff is for, maybe we could just remove it and nobody would notice.

gares (approved):
stm change ok

dev/top_printers.ml
ejgallego: I guess it is used to Print a Constr from Coq? I'd guess @herbelin uses it.

ppedrot: I would support that. I don't use them. @herbelin probably does though.

ejgallego:
> @ejgallego Nah, they are a huge hack to add commands to the debugger. Not sure we actually want to do that.

I can remove them but I'd like to hear @herbelin blessing.

dev/top_printers.ml
ppedrot: (Nice race condition.)

ejgallego:
I've ported the commands nevertheless, but they should be tested.

ejgallego:
Seems ready.

ppedrot (approved):

herbelin:
I did not use "PrintConstr" for years. OK to me.

ejgallego:
> I did not use "PrintConstr" for years. OK to me.

It is still supported tho, if you think it should go away please submit a PR cleaning it up.

[Tho I like them as they are a good use case on the new API, maybe I'll add to the doc of `Vernacextend`]

herbelin:
> It is still supported tho, if you think it should go away please submit a PR cleaning it up.

I don't think it need to go away.

Then I maybe misunderstood the other question:

> >    @ejgallego Nah, they are a huge hack to add commands to the debugger. Not sure we actually want to do that.

> I can remove them but I'd like to hear @herbelin blessing.

What is the question to me?

herbelin:
You can assign it to me if you want.

A question though: there is already a file `vernacextend.mlp` (not linked to Coq though). That may create confusions, at least for human beings.

ejgallego:
> A question though: there is already a file `vernacextend.mlp` (not linked to Coq though). That may create confusions, at least for human beings.

Indeed, even if the belong to totally different "compilation" domains; I guess that's OK as we are gonna move `grammar` to its own package for 8.10, if possible to preserve compat.

ppedrot:
Indeed grammar is likely to be extruded in 8.10, so this shouldn't really matter. I'll merge.

ppedrot:
@ejgallego This PR didn't run with the last CI, just in case can you rebase?

ejgallego:
> @ejgallego This PR didn't run with the last CI, just in case can you rebase?

Done.

ppedrot:
OK, I was just worried by paramcoq, and it passed the CI. Other failures are spurious, I'll merge.

PR #8923, opened 2018-11-06
Bump camlp5 minimal version and use its safe API.

ppedrot:
This PR makes the Coq parsing API rely on the type-safe camlp5 API, that was first introduced in camlp5 7.06.

This requires first to bump up the minimal camlp5 version. I did that in the opam file but I am pretty sure the code base is littered with version checks I don't know about. Also not sure about the process of bumping up camlp5 versions, so let's ask here.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,7 +9,7 @@ stages:
 variables:
   # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
   # for reference]
-  CACHEKEY: "bionic_coq-V2018-10-30-V1"
+  CACHEKEY: "bionic_coq-V2018-11-07-V1"
   IMAGE: "$CI_REGISTRY_IMAGE:$CACHEKEY"
   # By default, jobs run in the base switch; override to select another switch
   OPAM_SWITCH: "base"
--- a/INSTALL
+++ b/INSTALL
@@ -39,7 +39,7 @@ WHAT DO YOU NEED ?
    - Findlib (version >= 1.4.1)
      (available at http://projects.camlcity.org/projects/findlib.html)
 
-   - Camlp5 (version >= 7.03)
+   - Camlp5 (version >= 7.06)
      (available at https://camlp5.github.io/)
 
    - GNU Make version 3.81 or later
--- a/configure.ml
+++ b/configure.ml
@@ -709,9 +709,9 @@ let check_camlp5_version camlp5o =
   let version_line, _ = run ~err:StdOut camlp5o ["-v"] in
   let version = List.nth (string_split ' ' version_line) 2 in
   match numeric_prefix_list version with
-  | major::minor::_ when s2i major > 6 || (s2i major, s2i minor) >= (6,6) ->
+  | major::minor::_ when s2i major > 7 || (s2i major, s2i minor) >= (7,6) ->
     cprintf "You have Camlp5 %s. Good!" version; version
-  | _ -> die "Error: unsupported Camlp5 (version < 6.06 or unrecognized).\n"
+  | _ -> die "Error: unsupported Camlp5 (version < 7.06 or unrecognized).\n"
 
 let config_camlp5 () =
     let camlp5mod = "gramlib" in
--- a/coq.opam
+++ b/coq.opam
@@ -22,7 +22,7 @@ depends: [
   "ocaml"     {         >= "4.05.0" }
   "dune"      { build & >= "1.4.0"  }
   "num"
-  "camlp5"    {         >= "7.03"   }
+  "camlp5"    {         >= "7.06"   }
 ]
 
 build-env: [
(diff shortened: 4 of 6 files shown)


SkySkimmer:
configure is still thinking that camlp5 6.06 is enough.
INSTALL should be updated.
I think that's it.

ejgallego:
The dockerfile too

ejgallego:
I am OK with the dependency bump for 8.10, this should be ok for most users. Debian / Ubuntu have basically abandoned their Coq packages so I don't think that's a problem.

ejgallego (changes requested):
Fix:
- configure
- docker
- INSTALL
- travis
- windows?

Zimmi48:
Windows IIRC is using the edge compiler, not the minimum version.

ppedrot:
I think I have updated the relevant parts. Note that the current master configure is completely outdated w.r.t. the supported camlp5 version, and Travis seems to be already using 7.06.

ppedrot:
It seems that the Gitlab environment is still out of date. I guess I have to push a docker instance myself, but I have no idea on how to do that (and I'm not too fond of having to install the Docker *usine-à-gaz* on my own computer). Anybody more skilled around?

ejgallego:
It is fairly well-documented in the README @ppedrot , you need to raise the image version and the request a build in the Gitlab web interface with `SKIP_DOCKER` set to `foo`.

ejgallego:
Basically grep for `CACHEKEY`, raise it, and the rest we can take care of, gitlab will build the Docker image.

Zimmi48 (approved):
Trivially OK for my part.

SkySkimmer:
>windows?

Windows says 
>  if build_prep https://github.com/camlp5/camlp5/archive rel706 tar.gz 1 camlp5-rel706; then

so should be OK.

ejgallego:
On a more general note about camlp5, we are ready to drop the dependency as soon as we do the porting of the CI contribs. I am not sure it is worth keeping the infrastructure in 8.10, given what our compat policy for plugins is, and given that I am implementing state handling in coqpp following the path we did on attributes, so commands have to declare what state they read/modify.

ppedrot:
@ejgallego I submitted a few PRs for the CI developments, but the least we can say is that maintainers (probably us, most of the time) are not quite reactive. Also, before removing camlp5 entirely, we should first provide the camlp5 grammar.cma macro as a separate library depending on camlp5.

SkySkimmer:
Manual pipeline running.

Zimmi48:
@ppedrot You only changed the CACHEKEY in the GitLab config and not in the Dockerfile! Pipeline cancelled.

SkySkimmer:
The one in the dockerfile doesn't actually do anything, I think we should just remove it. @ejgallego WDYT?

ejgallego:
> The one in the dockerfile doesn't actually do anything, I think we should just remove it. @ejgallego WDYT?

We discussed implementing a check. IMHO it is important to have this information as we have to keep consistency manually. It is fairly easy to forget which version the Dockerfile is.

ejgallego:
> but the least we can say is that maintainers (probably us, most of the time) are not quite reactive.

What do you mean? AFAICT your PRs got merged, right?

> Also, before removing camlp5 entirely, we should first provide the camlp5 grammar.cma macro as a separate library depending on camlp5.

Why do we need that? It is already a separate library depending on camlp5 AFAICT.

ejgallego (approved):
Looks good to me.

ppedrot:
> What do you mean? AFAICT your PRs got merged, right?

Not all of them yet, and I had to ping for this to happen.

> Why do we need that? It is already a separate library depending on camlp5 AFAICT.

Hum, how do we distribute it to users requiring the legacy macro?

ejgallego:
> Not all of them yet, and I had to ping for this to happen.

If CI maintainers are not reasonably reactive we can take some action indeed. Let us know if someone is unreasonably slow.

> Hum, how do we distribute it to users requiring the legacy macro?

IMHO we should just drop support for `.ml4` files. Porting is very easy, and anyways for ml developments intra-version compatibility is just not there. Also, I am preparing a `coqpp` patch that will add systematic handling of state inspired in the way we handle attributes; this will surely mean `.ml4` won't be able to function anymore.

JasonGross:
Please revert this change.  I don't want to package camlp5 myself, and there is *no* Ubuntu version that supports this version of OCaml (the only supported version is 7.01, in cosmic, as far as I can tell from https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=Camlp5&searchon=names).  So this change breaks the usage of the apt packages on Travis for HoTT, fiat, fiat-crypto, etc.

Also, I'd prefer to have more than just one day to notice and comment on PRs that bump version dependencies in the future.

JasonGross:
(Btw, if this is not reverted soon, I'm going to have to stop testing fiat-crypto against Coq master, because it's built in stages due to Travis timeouts and the failure of any one version of Coq in a state will prevent the other stages from running.)

ppedrot:
Let's revert indeed, I was not expecting this PR to get merged so quickly, and was thinking there would have been a little bit of discussion precisely to avoid this kind of situation.

ejgallego:
@ppedrot you should reopen this PR.

ppedrot:
@ejgallego I'm waiting for #8945 to go in, I don't like having chains of dependencies in my PRs.

PR #8929, opened 2018-11-06
Fix fixpoint related lifting in open recursors + related cleanups

SkySkimmer:
(the kernel doesn't use fold_with_full_binders so no risk of inconsistency in existing code)

Not sure how to get some observable differences. Maybe I should just try making some unit tests.

### DIFF
--- a/engine/eConstr.ml
+++ b/engine/eConstr.ml
@@ -296,6 +296,8 @@ let decompose_prod_n_assum sigma n c =
 
 let existential_type = Evd.existential_type
 
+let lift n c = of_constr (Vars.lift n (unsafe_to_constr c))
+
 let map_under_context f n c =
   let f c = unsafe_to_constr (f (of_constr c)) in
   of_constr (Constr.map_under_context f n (unsafe_to_constr c))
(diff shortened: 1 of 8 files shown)


ppedrot (approved):
Not very pretty to move substitution code to constr, but unluckily I don't think there is a way around...

herbelin:
Maybe the internal representation of `fix` should instead be changed and the term be lifted by default. Moreover, that would eventually allow to support mutual fixpoints with sequential dependencies, as in:
```
Fixpoint A (n:nat) : Type :=
...
with f (n:nat) : A n :=
...
```
which, currently, we can only (unfortunately) simulate by gluing the pair of fixpoints in a sigma-types.

ppedrot:
@herbelin I think this change is part of @mattam82's branch for induction-recursion, so this will (hopefully) eventually happen.

herbelin:
@skyskimmer, @ppedrot: looks like you are saying the same thing, no?

>  I think this change is part of @mattam82's branch for induction-recursion, so this will (hopefully) eventually happen.

Yes, I think so. (But changing the representation now would avoid the inclusion of lift in constr.ml.).

By the way, it is a long time I did not have news about induction-recursion. What is the status?

SkySkimmer:
ping @maximedenes or @ppedrot (as secondary kernel maintainer)

ppedrot:
@maximedenes are you fine with the patch? It's simple and fixes a long-lurking bug.

maximedenes (approved):
I'm ok with the patch, but please confirm @herbelin is too.

mattam82:
For the record I’m also fine with this, let’s make it apparent that currently fixpoint types are not a telescope and need to be lifted. We can go back when ind-ind lands.

ppedrot:
@herbelin ping.

herbelin:
I agree with the proposal (let's keep the convention as it is, which would prevent potential incompatibilities with code abroad if the purpose is only bugfixing and not providing a new feature).

PR #8930, opened 2018-11-06
Remove int31_syntax_plugin

JasonGross:
The error messages have changed a bit, and we now support parsing
negative numerals as int31.  (If we wanted to not do this, that would be
pretty easy, we'd just need to write a version of `phi_inv` that
returned `option int31` and rejected negative `Z` inputs.)

c.f. https://github.com/coq/www/pull/80#issuecomment-436409783

**Kind:** enhancement


ejgallego (approved):
Build stuff trivially OK

vbgl:
Interesting. Probably this PR should be based on #6914 and remove `int63_syntax_plugin` instead…

gares (approved):
make & test OK

vbgl:
Well. In fact we cannot use numeral notations for 63-bit native integers since `int` is not an inductive type.

maximedenes:
Oh! Are numeral notations restricted to inductive types? What about real numbers, then?

Zimmi48:
The initial version of the numeral notation extension by Daniel supported the reals as well. This was dropped by Pierre for... some reason.

maximedenes (changes requested):
Please make sure this does not rely on `int31` being an inductive type, otherwise it will be fundamentally incompatible with the upcoming #6914. Thanks!

JasonGross:
As far as I can tell, there is no way to make this compatible with #6914 without basing it on top of #6914.  The reason it is restricted to inductive types is because it needs to know how to recognize syntactically that an expression should be printed as a numeral; then we can just check for "is it headed by the correct sort of constructor".  I am not sure how the printing support for reals used to work, but I believe support for them was dropped in part because it would require running Ltac during printing and parsing, and this was saved as "to be added at a later date".  The support for primitive integers should be easier: relax the "is_inductive" check to "is_inductive || is_primitive_int_type", and augment the bit where it looks for constructor expressions of the right inductive type to also know how to find primitive int expressions.  (However, there's no way to make this change before the primitive integer types are added!)

JasonGross:
Also, note that of the 10 lines added, 4 are in CHANGES, 3 are in the test-suite, 1 is whitespace, and one of the remaining two is a duplicate of the other, just in Compat.  So really this PR just removes a bunch of stuff that the int63 PR will remove anyway, and adds one test (which is incompatible with the current form of the int63 PR; it checks that negative numerals can be parsed), and adds one line to replace the removed stuff.

As I suggest above, I think the syntax for int63 should be implemented differently, given that printing is hard-coded into the syntax tree of Coq.  So I think the changes here are mostly independent of the int63 PR (except the bit about parsing negative numerals; is there a reason that's been forbidden historically?), but of course this PR will be useless when the int63 PR is merged

maximedenes:
You're right that the parsing of integers in the primitive integers PR is strange, and should be made first class. With that view, this PR seems orthogonal, which is good.

> except the bit about parsing negative numerals; is there a reason that's been forbidden historically?

I don't know, I think this was what I was wondering when I wrote the FIXME you saw there.

maximedenes (approved):

JasonGross:
> I wrote the FIXME you saw there

The FIXME, AIUI, is about printing negative numerals, not parsing them.  This PR does not change the fact that primitive integers are always printed as positive.

JasonGross:
Rebased to fix conflict in CHANGES.md

maximedenes:
Ok, so after discussing more with @vbgl, here's our view:

- Efficient parsing of `int` literals does matter, so we don't want to have to go through `Z`. The current plugin uses `Bigint`, which is way more efficient.
- We don't want to introduce yet another (ad-hoc) mechanism to register `int_scope` when syntax plugins provide what we want.
- Allowing non-canonical literals seems unnatural: `Check -1` will return something like `max_int : int`, you will be able to prove theorems like `9223372036854775808 = 0`, etc. Is there a use case for it?

So I don't think it makes much sense to merge this PR, since we will bring back the plugin in #6914.

maximedenes:
> * Efficient parsing of `int` literals does matter, so we don't want to have to go through `Z`. The current plugin uses `Bigint`, which is way more efficient.

I think I was half asleep when I wrote that. We do care about efficient parsing for `Bignums`, but for `int63` literals, not so much. So we could further experiment porting the parsing of these integers to numeral notations.

Zimmi48:
The initial version of numeral notations also had some code allowing going through Bignums for efficient parsing instead of going through Z. This is something else that Pierre decided to drop.

maximedenes:
Interesting, so what is the current status for Bignums? Does it still have a plugin or has it been turned into an inefficient parser?

vbgl:
Also, going through `Z` for parsing primitive integers is a bit annoying: indeed, the `of_Z` and `to_Z` functions have a few constants (zero and one) in their code; they cannot be defined before the numeral notation is declared. That’s a chicken-and-egg issue.

JasonGross:
> also had some code allowing going through Bignums for efficient parsing instead of going through Z

What was the Coq side of this?  Note that the current version allows going through `Decimal.int`

Here is an alternate proposal: augment the numeral notation machinery to accept anything registered as the primitive integer type as a source/destination.  I assert that this is almost as little work as getting the parsing working properly in the syntax plugin.  Then you can register identity functions as the printing and parsing functions for int63.

vbgl:
Is there something in this PR that is not yet in master? I think all of it has been integrated as part of the primitive integers change (e43b1768d0f).

JasonGross:
Indeed, it has been integrated

PR #8938, opened 2018-11-07
[Plugins] Remove some dead code

vbgl:
Because maintaining dead code is always an annoyance.


maximedenes:
@ppedrot can you take this one? Thanks!

ppedrot (approved):
Yep, I was going to do that but it somehow went under my radar. Trivial and will merge right now.

PR #8944, opened 2018-11-08
Revert PR #8923 (require camlp5 >=7.06)

### LINKED ISSUE #8941: Coq no longer builds with the newest version of Ubuntu
#### Version

Master

#### Operating system

Ubuntu (cosmic)

Ubuntu cosmic ships camlp5 7.01.  See https://github.com/coq/coq/pull/8923#issuecomment-436812588 for more details

ppedrot:
It was probably a bit premature to rely on such a recent camlp5, see the discussion in #8923. Also, quite urgent.

Fixes #8941.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,7 +9,7 @@ stages:
 variables:
   # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
   # for reference]
-  CACHEKEY: "bionic_coq-V2018-11-07-V1"
+  CACHEKEY: "bionic_coq-V2018-10-30-V1"
   IMAGE: "$CI_REGISTRY_IMAGE:$CACHEKEY"
   # By default, jobs run in the base switch; override to select another switch
   OPAM_SWITCH: "base"
--- a/INSTALL
+++ b/INSTALL
@@ -39,7 +39,7 @@ WHAT DO YOU NEED ?
    - Findlib (version >= 1.4.1)
      (available at http://projects.camlcity.org/projects/findlib.html)
 
-   - Camlp5 (version >= 7.06)
+   - Camlp5 (version >= 7.03)
      (available at https://camlp5.github.io/)
 
    - GNU Make version 3.81 or later
--- a/configure.ml
+++ b/configure.ml
@@ -709,9 +709,9 @@ let check_camlp5_version camlp5o =
   let version_line, _ = run ~err:StdOut camlp5o ["-v"] in
   let version = List.nth (string_split ' ' version_line) 2 in
   match numeric_prefix_list version with
-  | major::minor::_ when s2i major > 7 || (s2i major, s2i minor) >= (7,6) ->
+  | major::minor::_ when s2i major > 6 || (s2i major, s2i minor) >= (6,6) ->
     cprintf "You have Camlp5 %s. Good!" version; version
-  | _ -> die "Error: unsupported Camlp5 (version < 7.06 or unrecognized).\n"
+  | _ -> die "Error: unsupported Camlp5 (version < 6.06 or unrecognized).\n"
 
 let config_camlp5 () =
     let camlp5mod = "gramlib" in
--- a/coq.opam
+++ b/coq.opam
@@ -22,7 +22,7 @@ depends: [
   "ocaml"     {         >= "4.05.0" }
   "dune"      { build & >= "1.4.0"  }
   "num"
-  "camlp5"    {         >= "7.06"   }
+  "camlp5"    {         >= "7.03"   }
 ]
 
 build-env: [
--- a/dev/ci/docker/bionic_coq/Dockerfile
+++ b/dev/ci/docker/bionic_coq/Dockerfile
@@ -1,4 +1,4 @@
-# CACHEKEY: "bionic_coq-V2018-11-07-V1"
+# CACHEKEY: "bionic_coq-V2018-10-30-V1"
 # ^^ Update when modifying this file.
 
 FROM ubuntu:bionic
@@ -41,7 +41,7 @@ ENV BASE_OPAM="num ocamlfind.1.8.0 dune.1.4.0 ounit.2.0.8 odoc.1.3.0" \
     CI_OPAM="menhir.20180530 elpi.1.1.0 ocamlgraph.1.8.8"
 
 # BASE switch; CI_OPAM contains Coq's CI dependencies.
-ENV CAMLP5_VER="7.06" \
+ENV CAMLP5_VER="7.03" \
     COQIDE_OPAM="lablgtk.2.18.5 conf-gtksourceview.2"
 
 # base switch
--- a/parsing/pcoq.ml
+++ b/parsing/pcoq.ml
@@ -83,9 +83,13 @@ module type S =
   *)
 
   type 'a entry = 'a Entry.e
+  type internal_entry = Tok.t Gramext.g_entry
+  type symbol = Tok.t Gramext.g_symbol
+  type action = Gramext.g_action
   type coq_parsable
 
   val coq_parsable : ?file:Loc.source -> char Stream.t -> coq_parsable
+  val action : 'a -> action
   val entry_create : string -> 'a entry
   val entry_parse : 'a entry -> coq_parsable -> 'a
 
@@ -96,6 +100,9 @@ end with type 'a Entry.e = 'a Grammar.GMake(CLexer).Entry.e = struct
   include Grammar.GMake(CLexer)
 
   type 'a entry = 'a Entry.e
+  type internal_entry = Tok.t Gramext.g_entry
+  type symbol = Tok.t Gramext.g_symbol
+  type action = Gramext.g_action
 
   type coq_parsable = parsable * CLexer.lexer_state ref
 
@@ -106,6 +113,7 @@ end with type 'a Entry.e = 'a Grammar.GMake(CLexer).Entry.e = struct
     state := CLexer.get_lexer_state ();
     (a,state)
 
+  let action = Gramext.action
   let entry_create = Entry.create
 
   let entry_parse e (p,state) =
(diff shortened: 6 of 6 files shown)


ppedrot:
FTR, I think it is better to simply get rid of camlp5 altogether, which should happen soon enough.

ejgallego:
Why is it urgent? I don't the urgency at all here.

IMHO the version bump is OK,

ejgallego:
> FTR, I think it is better to simply get rid of camlp5 altogether, which should happen soon enough.

I have a patch that does that, will submit; but indeed it does require porting the contribs.

ejgallego:
See #8945

ppedrot:
> Why is it urgent?

IIUC, @JasonGross said it is going to screw the Fiat CI.

ejgallego:
> IIUC, @JasonGross said it is going to screw the Fiat CI.

Maybe it is the time they switch to Docker for CI. What Erik did is amazing.

There is a reason we only support Nix and Docker for CI. Supporting Debian packages for CI is not a good idea, and I say that as a proud and happy user of Debian for more than 20 years.

vbgl:
I’m not sure why I have to take the decision, but here is my opinion.

I agree with Pierre-Marie: since depending on Camlp5 7.06 breaks the build on some relevant platform and since we have a better solution to propose in a not too far future, let’s move slowly towards this better solution.

ejgallego:
Yeah we cannot support every single package manager for `master` testing I'm afraid. At least that's my point of view. Of course, if someone wants to volunteer for Debian that's fine; but at the moment it is in bad shape so I cannot feel bad we broke them.

So far the supported systems are: OPAM, Nix, and Docker.

I'd be happy to add a `pkg:deb` job to the CI, that would have prevented the problem above, but we need a volunteer to maintain it.

ejgallego:
> I’m not sure why I have to take the decision, but here is my opinion.

I cced you as I think this ultimately fails in the domain of the release manager. Of course you don't have to take a decision if you don't want.

I did express my opinion above; I don't see the point of reverting as currently we have nobody maintaining Coq for Debian; so the PR will sit there waiting for what? Nobody is going to update `camlp5`.

I would only support reverting if someone volunteers to maintain a `pkg:deb` job.

ejgallego:
Another argument against revert is that #8945 would also fix the problem in Debian, but only for now.

As long as we don't support Debian officially problems like this will keep arising and IMHO there is little we can do about them in a `master` branch.

herbelin:
Can someone explain the issue with Debian? Is it about CI? About providing packages? Is it about interacting with the people listed at [Debian's coq page](https://packages.debian.org/sid/coq)? Thanks in advance.

ejgallego:
> Can someone explain the issue with Debian?

The main issue @herbelin is lack of manpower, that is to say, there seems to be nobody with time enough to keep core OCaml packages up to date. Coq itself is very outdated, and same for other tools.

Thus, my opinion is that we cannot recommend a Debian-based setup for CI of the `master` branch.

ejgallego:
> Thus, my opinion is that we cannot recommend a Debian-based setup for CI of the `master` branch.

Specially as we support 3 systems: Nix, Opam, and Docker.

But let me remark that adding a `pkg:deb` to our CI and thus making Debian officially supported would be great and I'd try to help as much as I could. However that's not going to be easy due to the status of OCaml there.

ppedrot:
@ejgallego The point is to let the Fiat people have a bit of time to port their infrastructure. We can re-revert after that, if you really insist. But getting rid of camlp5 is a matter of days if not hours, so I wouldn't be too concerned.

herbelin:
Does that "Debian is [maybe] not compatible with being a bleeding-edge testing platform" mean that noone uses OCaml and coq Debian's packages? Do you mean that OCaml and Coq pages at Debian (which I guess many Debian users rely on) should be rewritten so as to recommend installation via OPAM?

ejgallego:
> Does that "Debian is [maybe] not compatible with being a bleeding-edge testing platform" mean that noone uses OCaml and coq Debian's packages? Do you mean that OCaml and Coq pages at Debian (which I guess many Debian users rely on) should be rewritten so as to recommend installation via OPAM?

Indeed @gares just did this point here: https://github.com/coq/coq/issues/8941#issuecomment-436925984

So we either find someone to volunteer to fix Debian, or indeed we should recommend OPAM.

Next Debian is going to ship with a quite old [and inconsistent I think] Coq release.

SkySkimmer:
>Yeah we cannot support every single package manager for master testing I'm afraid. At least that's my point of view. Of course, if someone wants to volunteer for Debian that's fine; but at the moment it is in bad shape so I cannot feel bad we broke them.

I do feel bad that we broke them, and considering the docker system was only announced on coq-club 1 day ago I don't think it's fair to say they should just have been on docker.
Giving people more than 1 day to notice upcoming issues and adapt to them is not the same as supporting whatever system they use IMO.

ejgallego:
@ppedrot Fiat's CI is being broken because their CI depends on Debian packages.

Thus it is 100% related to the issue at stake here.

ejgallego:
> Thus it is 100% related to the issue at stake here.

At the bare minimum before reverting we should hear what @JasonGross plans are.

ppedrot:
> @ppedrot Fiat's CI is being broken because their CI depends on Debian packages.

I'm refraining from thanking Captain Obvious, but this is an epiphenomenon. The goal is not to fix Debian, which is out of our league. The goal is to have Fiat upgrade their CI to Docker.

ejgallego:
> The goal is to have Fiat upgrade their CI to Docker.

I am not sure they are going to like that solution; if I understand correctly they'd like to stay with their current setup, or at least they have a preference for doing so.

So indeed you are making a strong assumption here.

ppedrot:
Also, I don't understand the fuss around this PR. We break an external contrib we track in an involved way, the only sensible solution is reverting immediately, and **then** discussing cold-headedly. This is urgent, I repeat.

SkySkimmer:
>The key point now is what is the less effort for all involved:

Why not both? ie revert, then do whichever of fixing their CI or #8945 is ready first, then unrevert.

>Also, I don't understand the fuss around this PR. We break an external contrib we track in an involved way, the only sensible solution is reverting immediately, and then discussing cold-headedly. This is urgent, I repeat.

I agree, and as such I'll merge.

ppedrot:
> So indeed you are making a strong assumption here.

If this is hard, we can just wait for the removal of camlp5, as I advocated for above. But this is not on the same timescale.

ejgallego:
Sorry we didn't break anything we support; so this is not urgent.

You are assuming a lot about what the plans of the Fiat people are; they have never mentioned upgrading to docker of modifying their CI system, they asked to revert the change to camlp5 and they stated that they don't have a plan to fix the issue.

Thus no plan to fix the issue == no point on reverting.

ejgallego (dismissed):
Please block the merge until @JasonGross comments on what the plan for Fiat's CI should be.

ejgallego:
Also, on the ground of principles of development the rule about urgency is very clear to me. If you didn't care to get your stuff in Coq's CI, the issue is not urgent for us.

Period. We cannot rush to an urgent revert for any random 3rd party CI system that we are not aware of.

SkySkimmer:
No, in the spirit of 
>When a PR received lots of comments or if the PR has not been opened for long and the assignee thinks that some other developers might want to comment, it is recommended that they announce their intention to merge and wait a full working day (or more if deemed useful) before the actual merge, as a sort of last call for comments.

we should definitely revert.

In any case if this was @ejgallego's PR being reverted I would be more wary but considering @ppedrot is for reverting I'll just go ahead.

SkySkimmer (approved):

JasonGross:
I have been maintaining Ubuntu packages for Coq for use in CI for a while, and plan to continue doing that.  I do not want to have to package any of Coq's dependencies myself.

My preference is for this revert to be merged, and then to drop the dependency on camlp5.

If docker is really the only way of supporting Coq on CI moving forwards, I can start looking into that, but it'll take some time.  If it's possible, though, I'd like to keep using debian packages on Travis, for the time being.

ejgallego:
> In any case if this was @ejgallego's PR being reverted I would be more wary but considering @ppedrot is for reverting I'll just go ahead.

I don't care about the revert itself but about the policy precedent we are setting wrt to urgency of revert.

This is not urgent in any way and I want indeed to send the message out LOUD to everyone: if you don't bother to get your stuff in our CI [for example `pkg:deb`] then you are **very low priority** when things break.

Any other policy just doesn't make sense for us.

ejgallego:
Anyways go ahead if you want. @ppedrot when you resubmit you should update the cachekey again.

ejgallego:
> I have been maintaining Ubuntu packages for Coq for use in CI for a while, and plan to continue doing that. I do not want to have to package any of Coq's dependencies myself.

You have two choices: add a `pkg:deb` to Coq's CI or indeed move the CI to any of the supported systems.

I redid the Debian packaging sometime ago using `gbp` and it was nicer than the Debian 9 packages, YMMV.

JasonGross:
> Period. We cannot rush to an urgent revert for any random 3rd party CI system that we are not aware of.

If you want, I can possibly set up a script that checks the debian build dependencies on Coq's CI?  I haven't done this in the past because the last time I proposed a thing that was not a Coq development (my bug minimizer), I think I was told that that sort of thing was not the right sort of thing to be tested on Coq's CI.

ejgallego:
> I can possibly set up a script that checks the debian build dependencies on Coq's CI?

That won't suffice; we should add a job that actually builds the package [and it would have the nice side-effect of generating the artifact for people to use]

Anyways it is looking grim for Debian + Coq with the removal of gtk2. At least for the CoqIDE part.

ejgallego:
> Anyways it is looking grim for Debian + Coq with the removal of gtk2. At least for the CoqIDE part.

Note that adding a `pkg:deb` to the CI would mean the same than adding a regular contrib and fall under CI policy: that is to say, maintainers of the deb commit to implement changes in a reasonable period as to stay compatible with Coq.

For example that implies upgrading `dune`, `campl5`, `zarith / num`, etc... as needed, or else risk being dropped. This is a similar compromise to what Nix / OPAM people has done so far.

herbelin:
> > Does that "Debian is [maybe] not compatible with being a bleeding-edge testing platform" mean that noone uses OCaml and coq Debian's packages? Do you mean that OCaml and Coq pages at Debian (which I guess many Debian users rely on) should be rewritten so as to recommend installation via OPAM?

> Indeed @gares just did this point here: https://github.com/coq/coq/issues/8941#issuecomment-436925984

I just have a discussion with Ralf T. I'll try to make a summary of what we discussed (mixing with what @gares and Stéphane G said) within a couple of days.

ejgallego:
> I just have a discussion with Ralf T. I'll try to make a summary of what we discussed (mixing with what @gares and Stéphane G said) within a couple of days.

Cool, thanks @herbelin

PR #8945, opened 2018-11-08
[camlp5] Remove dependency on camlp5.

ejgallego:
depends #8985


ejgallego:
This is mostly ready but we need a few tweakes on the packing of `gramlib` [for the Make-based build] and likely some other things.

ejgallego:
I tried to pack gramlib in the style of OCaml's stdlib in https://github.com/ejgallego/coq/commit/308213b5e3a58632070b2f958625290eb8a47c3c

However it doesn't work, I don't understand why :(

SkySkimmer:
Give us a log or something to look at.

ejgallego:
```
File "gramlib/plexing.ml", line 1:
Error: Could not find the .cmi file for interface gramlib/plexing.mli.
```

ejgallego:
Failures:
- `coq-elpi` because of missing packing of `gramlib` This should be the only blocker for this PR. We could either pack or rename, but I'd be good to understand how to properly pack in `make`.
- `equations` https://github.com/mattam82/Coq-Equations/pull/148

ejgallego:
> BTW the ocaml manual on packing is around "Note the use of double underscores" in http://caml.inria.fr/pub/docs/manual-ocaml/extn.html

Umm, I am not sure what I am missing here: https://github.com/ejgallego/coq/tree/gramlib%2Bpack

SkySkimmer:
Have a look at SkySkimmer/coq@5d5c3ced92d64bf3310cc818dd09acd3f1f520e9 which tries to mimick dune
However dependencies from the rest of Coq on gramlib generate nonsense so it doesn't work

gares:
For the milestone, isn't he case that coqpp is not in 8.9? If so 8.10 will be the first release in which camlp5 can be deprecated, and removed in 8.11. Am I wrong?

JasonGross:
> If so 8.10 will be the first release in which camlp5 can be deprecated, and removed in 8.11. Am I wrong?

I think this is true of support for camlp5 in user plugins, but it should be possible to remove it as a build dependency of Coq itself in 8.10, right?

SkySkimmer:
We break plugins all the time, IMO it's fine not to have a deprecation release.

ejgallego:
8.9 has coqpp, but not will all the capabilities; in particular I guess the CoqMakefile changes are missing, plus some more stuff but that may be more esoteric.

But indeed, sadly we are not in a position to offer compatibility to most plugins anyways; so no point on having a deprecation phase for this, specially as 8.10 is going to introduce many incompatibilites due to making some interfaces more functional [I hope the pain is worth it]

ejgallego:
We can indeed make a coqp5 package that ships `grammar` if we want; but I am not sure it is going to be worth the work.

MSoegtropIMC:
Changes to the Windows build files are fine. As with your other PR: if I can help on the Windows side, please let me know.

Zimmi48 (approved):
We should aim at making it possible for a plugin to compile with two successive versions of Coq in the future. But it is true that we are not there yet. The work on packing the components that will be possible thanks to dune should help in this respect.

gares:
@ejgallego the failure in elpi could also be fixed by renaming `ploc` in `cPloc` and similarly the other files.
For packing in `make` you can look at what we do for .mlpack files in CoqMakefile.in.

SkySkimmer:
8a69bcb0cedf4a5fff144cebd011bf31bd86194f seems to work for packing in make, although the dependency handling is hacky. It breaks dune though...
We should get an ocaml build expert to look at this.

ppedrot:
That's super ugly though. Couldn't we just have some local makefile?

gares:
Can you explain what is this "packing dune caml 4.08 style" that does not use -pack and -for-pack?
Why is it called packing? Will -pack disappear?

ejgallego:
> @ejgallego the failure in elpi could also be fixed by renaming `ploc` in `cPloc` and similarly the other files.

I'd rather avoid this as we have the proper tools to handle the issue.

> For packing in `make` you can look at what we do for .mlpack files in CoqMakefile.in.

This is using the `-pack` flag method, which has several problems; we instead use the `-no-alias-method` which is what is used in the OCaml stdlib today and is the standard. However Coq's make system seems to have a hard time coping with it so it seems.

ejgallego:
> Why is it called packing? Will -pack disappear?

See the OCaml manual entry that @SkySkimmer posted [by the way the packing was done in 4.07]
[section 8.12 at https://caml.inria.fr/pub/docs/manual-ocaml/extn.html]

But indeed, I understand -pack has several problems and not even OCaml itself is using it.

ejgallego:
> It breaks dune though...

I guess it is because you didn't remove the `-open Gramlib` from the dune flags?

I dunno what todo; moving to the `__` packing method would allow for a large cleanup, note that `ocamldep` already has special support for the handling of old-style pack files.

I'd say let's port `ocamldep` and also pack plugins with the new method; however it is hard to justify the work when we already have a tool that can do that. I dunno.

rgrinberg:
> Can you explain what is this "packing dune caml 4.08 style" that does not use -pack and -for-pack? Why is it called packing? Will -pack disappear?

I don't what the future holds for packing (though I highly doubt it would get removed), I can explain why dune prefers module aliases to packing. Module aliases result in smaller executables and much better incremental compilation. There's an old blog post that explains this https://blog.janestreet.com/better-namespaces-through-module-aliases/ if you're interested.

ejgallego:
@JasonGross how do you want to approach the porting of Fiat to `mlg`?

AFAICS we would have to introduce a lot of custom logic in the makefiles to account for the different `ml4` vs `mlg` extension.

JasonGross:
@ejgallego The biggest issue is that we check `_CoqProject` into git, and `_CoqProject` has to be different in different versions.  The rest of the custom logic isn't an issue, I'm working on it in https://github.com/mit-plv/fiat/pull/18

ejgallego:
> @ejgallego The biggest issue is that we check `_CoqProject` into git, and `_CoqProject` has to be different in different versions. The rest of the custom logic isn't an issue, I'm working on it in [mit-plv/fiat#18](https://github.com/mit-plv/fiat/pull/18)

What other plugins do is to detect the extension based on the version and then generate `_CoqProject` using `sed` from a template. That's not so painful I guess.

SkySkimmer:
I think I figured out how to pack in Makefile, I was missing knowledge of the existence of `ocamldep -map`. See b77ba4e157b8150324cb9e51e9340dbdbc9306be. Dune does a copy dance (see commit) because in make we need the files to be `gramlib__Ploc.ml` etc, maybe there's a more generic way to do this than having one `(copy# ...)` per file but if there's not I think it's not too bad.

ejgallego:
Why you need the files to be `foo__bar.ml` ? Neither the OCaml stdlib nor Dune require this, as they are able to use `ocamlc bar.ml -o foo__bar.ml` etc...

ejgallego:
Thanks @JasonGross , relaunching the build.

SkySkimmer:
Because otherwise ocamldep produces dependencies for bar.cmo but what we 
need to produce is foo__bar.cmo
ocaml does some sed in the produce of ocamldep, see the end of 
stdlib/Makefile
dune does postprocessing of the output of `ocamldep -modules` (if you 
look eg in _build/default/parsing/.parsing.objs there's cLexer.ml.d 
produced by ocamldep containing Gramlib but also Plexing and Ploc even 
though dune will do `-o gramlib/.gramlib.objs/gramlib__Ploc.cmo`) (I 
don't know what postprocessing exactly)

Gaëtan Gilbert

ejgallego:
@SkySkimmer I suggest we submit the packing separatedly, do you think it does make sense?

I am a bit scared by the amount of hack we need to use tho, I could be better to wait for #8729 then.

SkySkimmer:
I don't think it's especially hacky now. The hardcoded dependencies internal to gramlib could probably be fixed too, I just can't be bothered right now.
I'm fine delaying the packing though.

ejgallego:
> I don't think it's especially hacky now. The hardcoded dependencies internal to gramlib could probably be fixed too, I just can't be bothered right now.
> I'm fine delaying the packing though.

Oh if you think that's OK then please submit a PR or push to this branch then; only Ltac2 remains a real problem so IMHO it'd be good if we can wrap this up then and move on.

SkySkimmer:
>Oh if you think that's OK then please submit a PR or push to this branch then; only Ltac2 remains a real problem so IMHO it'd be good if we can wrap this up then and move on.

OK pushed.

gramlib/dune
```diff
@@ -1,3 +1,16 @@
...
+; Makefile packing expects the packed modules to have file names like pack__Foo.ml
+; otherwise ocamldep has issues (dune uses ocamldep -modules then does postprocessing)
+(rule (targets ploc.ml)     (action (copy# gramlib__Ploc.ml ploc.ml)))
+(rule (targets ploc.mli)    (action (copy# gramlib__Ploc.mli ploc.mli)))
+(rule (targets plexing.ml)  (action (copy# gramlib__Plexing.ml plexing.ml)))
+(rule (targets plexing.mli) (action (copy# gramlib__Plexing.mli plexing.mli)))
+(rule (targets Gramext.ml)  (action (copy# gramlib__Gramext.ml Gramext.ml)))
+(rule (targets Gramext.mli) (action (copy# gramlib__Gramext.mli Gramext.mli)))
+(rule (targets Grammar.ml)  (action (copy# gramlib__Grammar.ml Grammar.ml)))
+(rule (targets Grammar.mli) (action (copy# gramlib__Grammar.mli Grammar.mli)))
```
ejgallego: please lowercase the module names.

ejgallego:
A nit that bothers me a bit is the spurious renaming [as we will rename the files back] and the fact that is the Dune part and git history who's paying the cost of lack of make support.

Maybe we should just ignore `gramlib` at all in the Makefile and instead have a pre-ocamldep rule that does generate a `gramlib_pack` directory and copies the prefixed files?

ppedrot:
> Maybe we should just ignore gramlib at all in the Makefile and instead have a pre-ocamldep rule that does generate a gramlib_pack directory and copies the prefixed files?

I support that.

SkySkimmer:
done

Gaëtan Gilbert

ejgallego:
> done

Taking the lock, I'm gonna rebase to solve the conflict.

I don't see the `_pack` change BTW, did you push the right commit?

ejgallego:
Lock released.

ejgallego:
> What do you mean?

Oh I thought you were talking about the `gramlib__pack` folder.

SkySkimmer:
I don't intend to work on this further, feel free to continue or merge.

ejgallego:
> I don't intend to work on this further, feel free to continue or merge.

Ok I will take care of this.

ejgallego:
Ok, so modulo small nits this is ready to go.

ejgallego:
Looking for an assignee; the natural one would be @ppedrot as the maintainer of gramlib.

ejgallego:
Failures are spurious TTBOMK, so once #8985 is merged this should be good to go.

maximedenes (approved):

ejgallego:
Dependency was finally merged, let's see if the CI is OK.

ppedrot:
In the end, the addition of coqpp + gramlib might have been somehow compensated by all the configure / Makefile cruft that was removed...

ppedrot (approved):
It is likely that there are still a few parts to fully clean up, but the PR seems to do a good job of removing all the now irrelevant mentions and uses of camlp5.

ppedrot:
This PR is the closing point of the camlp5 removal. I don't think there is much to discuss about it, we've been at it for quite a long time already, and it allows to do a lot of stuff, e.g. testing OCaml trunk. Thus I won't wait long to merge it once the CI goes through. If there are remaining dead code / documentation as I suspect this should be handled in a subsequent PR. *Parlez maintenant ou taisez-vous à jamais* as we say.

SkySkimmer (approved):

dev/build/windows/makecoq_mingw.sh
```diff
@@ -1386,7 +1386,7 @@ function make_coq {
   make_ocaml
   make_num
   make_findlib
-  make_camlp5
+  # make_camlp5
```
Zimmi48: I am not quite sure this is correct with respect to the building of addons (e.g. doesn't elpi still rely on camlp5?). Since the Windows build is currently broken apparently, we won't know for sure. Thus, I suggest to delay the changes to this file in a subsequent PR (e.g. the Dune + Windows one).

ejgallego: As far as I can see this is correct, the CI passed in the past, and elpi doesn't seem to be a supported addon.

Anyways we really need to do something about the mess `dev/build/windows` has become, starting by removing the huge amount of cruft living there.

I will indeed further additions of addons until we refactor the addon building code there; it will likely involve expressing the dependencies better (via a `CI_addon_DEPS` variable?) Then the script should be able to call the build procedure.

ejgallego: I confirm the windows failure is "spurious" to this PR, it is mtac2 that is failing but I cannot arrive to understand the error, for some reason `Unicoq` is not found but it should be.

ejgallego:
So indeed time to say _oui, je veux_ @ppedrot :)

ppedrot:
One thing less to care about.

PR #8947, opened 2018-11-08
Ensure termination of `file_exists_respecting_case`

### LINKED ISSUE #8943: infinite recursion in `file_exists_respecting_case` on absolute paths
#### Version

8.8.2, master

#### Operating system

Linux x86_64

Trying to run `coqchk` with an absolute path to a `.vo` file results in a crash instead of a proper error message.

~~~
$ coqchk -silent -norec $(coqc -where)/theories/Arith/Lt.vo
Fatal Error: Stack overflow
~~~

Here is a patch:

~~~
--- a/lib/system.ml
+++ b/lib/system.ml
@@ -83,7 +83,7 @@ let file_exists_respecting_case path f =
   let rec aux f =
     let bf = Filename.basename f in
     let df = Filename.dirname f in
-    (String.equal df "." || aux df)
+    (String.equal df "." || String.equal f df || aux df)
     && exists_in_dir_respecting_case (Filename.concat path df) bf
   in (!trust_file_cache || Sys.file_exists (Filename.concat path f)) && aux f
~~~

The issue is that `Filename.dirname` behaves as the identity function on the “root” path `"/"`.

vbgl:
Fixes #8943.

### DIFF
--- a/lib/system.ml
+++ b/lib/system.ml
@@ -83,7 +83,9 @@ let file_exists_respecting_case path f =
   let rec aux f =
     let bf = Filename.basename f in
     let df = Filename.dirname f in
-    (String.equal df "." || aux df)
+    (* When [df] is the same as [f], it means that the root of the file system
+       has been reached. There is no point in looking further up. *)
+    (String.equal df "." || String.equal f df || aux df)
     && exists_in_dir_respecting_case (Filename.concat path df) bf
   in (!trust_file_cache || Sys.file_exists (Filename.concat path f)) && aux f
 


herbelin:
Great. I let @ejgallego assigning (but tell me if I'm needed).

ejgallego:
I still fail to understand why do we need all this complex file handling code in Coq. Last time I looked it was because `library` was doing some inference of the module path based on the file, which should never be done.

Anyways if that makes the pain of the current setup easier I can merge.

PR #8949, opened 2018-11-08
Remove checker printers

SkySkimmer:
Now that the checker is using the regular kernel files it can also use
the normal printers.

Actually this is probably broken in dune since that copies the kernel files instead of linking the kernel. I think that's an argument for having dune build the checker by linking the kernel. @ejgallego @maximedenes WDYT?


ejgallego:
> Actually this is probably broken in dune since that copies the kernel files instead of linking the kernel. I think that's an argument for having dune build the checker by linking the kernel. @ejgallego @maximedenes WDYT?

It is likely broken indeed but the solution is instead to add the right `copy` rule for the printers too, and link against `Checklib`. So IMHO the printers are orthogonal as to whether to link vs the kernel or not.

SkySkimmer:
I do want to link the checker printers with extern so that we can have nice printers when debugging the checker though.

ejgallego (approved):
Looks fine to me, it is not essential we fix the printers for the checker in Dune right now.

SkySkimmer:
Test suite failure is spurious (non deterministic timeout).

PR #8950, opened 2018-11-08
[topfmt] Add phase attribute for toplevel printing.

ejgallego:
This is localized version of #8833, but instead of adding a phase
attribute which, as pointed by @gares has some problematic semantics,
we add a local one to the toplevel functions.

This moves the imperative part of the API to a better-delimited scope
and allows to progress with the separation of the interactive and
compilation API.

Note that still quite a few issues do remain in the "Feedback" path,
for example, idetop and other feedback clients cannot get a hold of
the feedback early enough as to direct init messages to the IDE part.

This is for example a serious issue of the API that shall be treated
separately.

### DIFF
--- a/toplevel/coqargs.ml
+++ b/toplevel/coqargs.ml
@@ -9,7 +9,7 @@
 (************************************************************************)
 
 let fatal_error exn =
-  Topfmt.print_err_exn Topfmt.ParsingCommandLine exn;
+  Topfmt.(in_phase ~phase:ParsingCommandLine print_err_exn exn);
   let exit_code = if CErrors.(is_anomaly exn || not (handled exn)) then 129 else 1 in
   exit exit_code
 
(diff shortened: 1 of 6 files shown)


ejgallego:
@gares I think this is a good compromise to be able to move forward with the rest of toplevel PRs.

ejgallego:
A question is whether we want to restore the "phase" on an exception; this PR has the consequence that feedbacks will be printed with the right phase, however not the errors; as they don't travel thru feedback actually.

ejgallego:
Ping! Maybe @herbelin would like to take this?

herbelin:
I can take it. Was @gares already involved in the PR. Or does he want to make comments?

gares (approved):

ejgallego:
So the main question here is if we should restore the previous phase when an exception happens; I think we may want both cases, so it may necessary to add a parameter.

ejgallego:
Any further thoughts @herbelin ? Are you OK with the current semantics, would you like a tweak? Maybe a `preserver_phase_on_error` optional parameter?

herbelin:
Let's go with this at the current time.

ejgallego:
Merci @herbelin !

herbelin:
OK, merging tomorrow.

ejgallego:
Merci !

PR #8955, opened 2018-11-09
[ssr] "case/elim: p" don't resolve TC in "p"

gares:
This fixes a bug reported by private communication.
It comes with a test case.

The bug was coming from the fact that elim unifies the pattern inferred from the (index in the) type of the inductive being eliminated and the one (optionally) provided by the user. Unifying `_` with `op _ _` was succeeding. Unification was performed via `Ssrmatching.unify_HO` that at some point in time was made to resolve TC too. We make this resolution optional.

### DIFF
--- a/plugins/ssrmatching/ssrmatching.ml
+++ b/plugins/ssrmatching/ssrmatching.ml
@@ -354,14 +354,16 @@ let nf_open_term sigma0 ise c =
   let c' = nf c in let _ = Evd.fold copy_def sigma0 () in
   !s', Evd.evar_universe_context s, EConstr.of_constr c'
 
-let unif_end env sigma0 ise0 pt ok =
+let unif_end ?(solve_TC=true) env sigma0 ise0 pt ok =
   let ise = Evarconv.solve_unif_constraints_with_heuristics env ise0 in
   let tcs = Evd.get_typeclass_evars ise in
   let s, uc, t = nf_open_term sigma0 ise pt in
   let ise1 = create_evar_defs s in
   let ise1 = Evd.set_typeclass_evars ise1 (Evar.Set.filter (fun ev -> Evd.is_undefined ise1 ev) tcs) in
   let ise1 = Evd.set_universe_context ise1 uc in
-  let ise2 = Typeclasses.resolve_typeclasses ~fail:true env ise1 in
+  let ise2 =
+    if solve_TC then Typeclasses.resolve_typeclasses ~fail:true env ise1
+    else ise1 in
   if not (ok ise) then raise NoProgress else
   if ise2 == ise1 then (s, uc, t)
   else
@@ -370,7 +372,7 @@ let unif_end env sigma0 ise0 pt ok =
 
 let unify_HO env sigma0 t1 t2 =
   let sigma = unif_HO env sigma0 t1 t2 in
-  let sigma, uc, _ = unif_end env sigma0 sigma t2 (fun _ -> true) in
+  let sigma, uc, _ = unif_end ~solve_TC:false env sigma0 sigma t2 (fun _ -> true) in
   Evd.set_universe_context sigma uc
 
 let pf_unify_HO gl t1 t2 =
--- a/plugins/ssr/ssrcommon.ml
+++ b/plugins/ssr/ssrcommon.ml
@@ -499,6 +499,22 @@ let pf_e_type_of gl t =
   let sigma, ty = Typing.type_of env sigma t in
   re_sig it sigma, ty
 
+let pf_resolve_typeclasses ~where ~fail gl =
+  let sigma, env, it = project gl, pf_env gl, sig_it gl in
+  let filter =
+    let evset = Evarutil.undefined_evars_of_term sigma where in
+    fun k _ -> Evar.Set.mem k evset in
+  let sigma = Typeclasses.resolve_typeclasses ~filter ~fail env sigma in
+  re_sig it sigma
+
+let resolve_typeclasses ~where ~fail env sigma =
+  let filter =
+    let evset = Evarutil.undefined_evars_of_term sigma where in
+    fun k _ -> Evar.Set.mem k evset in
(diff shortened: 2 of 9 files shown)


plugins/ssrmatching/ssrmatching.ml
```diff
@@ -354,28 +354,30 @@ let nf_open_term sigma0 ise c =
   let c' = nf c in let _ = Evd.fold copy_def sigma0 () in
   !s', Evd.evar_universe_context s, EConstr.of_constr c'
 
-let unif_end env sigma0 ise0 pt ok =
+let unif_end ?(solve_TC=true) env sigma0 ise0 pt ok =
```
maximedenes: Are you sure you want the flag to be optional? Shouldn't we look at each call and decide whether it makes sense to resolve TCs or not?

gares: If you want to have all ssr code be double checked for compatibility with TC, then yes, but it quite more ambitious.

I think that the current code is a bit messy, in particular `_end` means very little.
Historically `unif_end` was called at the end of a matching operation that also corresponded to the end of a tactic, and there were no TC. Then I started to use it as part as a higher level API unify_HO (since it ensure all remaining unif problems are solved, eg that unification did work for real). 
Then I think TC were added and it still makes sense to resolve TC at the end of a tactic (since the lemma you use may need these classes). But I don't think solving TC makes sense after a unification step.

After CI finishes (the failure seems unrelated, and works here), I'm pushing a commit with two different defaults:
- on for unif_end
- off for unify_HO
and let's see how it goes.

vbgl:
Thanks Enrico.

Here is a variant of the bug, which still fails:

~~~
From Coq Require Import Utf8 Bool ssreflect.

Set Printing All.
Set Debug Ssreflect.

Class Class := { sort : Type ; op : sort → bool }.
Coercion sort : Class >-> Sortclass.
Arguments op [_] _.

Lemma opP (A: Class) (a: A) : reflect True (op a).
Proof. Admitted.

Section Section.
  Context (A B: Class) (a: A).

  Goal is_true (op a).
    Fail by case: opP.
  Abort.

End Section.
~~~

In the debug trace, we can see that the `B` instance is inferred too early:

> Debug: SSR: globbing pattern: opP
> Debug: SSR: interpreting: opP
> Debug: SSR: decoded as: opP
> Debug: SSR: typed as: opP
> Debug: SSR: ==CASE==
> Debug: SSR: elim= match _ as r in (reflect _ b) return (_ b r) with
>                   | ReflectT _ x => _ x
>                   | ReflectF _ x => _ x
>                   end
> Debug: SSR: elimty= (_ _ _)
> Debug: SSR: c_is_head_p= true
> Debug: SSR: adding inf pattern _
> Debug: SSR: patterns=`{+}(opP _ _), {+}(@op B _)`
> Debug: SSR: inf. patterns=`_, (@op B _)`
> Debug: SSR: matching: `{+}(opP B _)`
> Debug: SSR: matching: `{+}(@op B _)`

gares:
I was lucky enough that the improved fix I proposed to maxime fixes this second declination of the bug ;-)

vbgl:
The current state of this PR does indeed addresses the real issue that triggered this discussion. Thanks. However, CI reports some problems, isn’t it?

gares:
Yep, I'll have to look at these since the behavior changed. My first fix was really local to elim, but the second one was more global... I guess it does not break mathcomp because there are no type classes.

gares:
I guess the trend is clear now. Before the patch resolve_typeclasses was called randomly (whenever a `unify_HO` was performed). After the patch it is not called anymore hence hence tactics do call it at the very end.

Most tactics end by calling `refine_with` but unfortunately it is not possible to put the call there easily.
The problem is that a few tactics (probably only elim) calls it with `~evars:false` that means it is using the "old" refiner. That one, IIRC, is smarter at HO things, but does not support goals containing evars (and I don't have an easy way to call resolve_typeclasses in this case). I tried removing that cruft, but some (frankly very shaky) proof script in Coqelicot was breaking. So I postpone this cleanup, I first want to know it Coqelicot is to be subsumed by mathcomp analysis or not.

Incidentally I came up with an API to resolve the type classes instances *of a term*.
I think this API is more natural than the current one, where `~filter` is an arbitrary predicate, maybe @mattam82 give a look at the commit touching ssrcommon and tell me what you think.

gares:
@CohenCyril I forgot to cite you, please answer about Coquelicot (if the answer is known).

mattam82:
The API for filtering the evars on which to launch resolution seems fine to me. Right now the ~filter is also used to filter based on the evar_source in Coq which should get a cleanup in the future.

gares:
@vbgl CI seems fine now, there are two failures but they don't depend on ssr hence I think they are spurious. @maximedenes please review the current patch and eventually merge.

CohenCyril:
> @CohenCyril I forgot to cite you, please answer about Coquelicot (if the answer is known).

Coquelicot will not be subsumed by mathcomp/analysis in the near future.

CohenCyril:
This bug is related to the one in this issue. Although the hint mode ensures that no typeclass resolution may happen before the pattern is resolved, case still fails.
Surprisingly for me, your patch also solves this...
Can you please add this test to the regression in your patch?

```coq
From Coq Require Import Utf8 Bool ssreflect.

Set Printing All.
Set Debug Ssreflect.

Class Class sort := { op : sort → bool }.
Arguments op {_ _}.
Hint Mode Class !.

Lemma opP A (C: Class A) (a: A) : reflect True (op a).
Proof. Admitted.
Arguments op {_ _}.

Section Section.
  Context A B (CA : Class A) (CB : Class B) (a: A).

  Goal is_true (op a).
    by case: opP.
  Abort.

End Section.
```

gares:
This should be merged. @maximedenes or @vbgl can you do that?

vbgl (approved):

CohenCyril:
@gares I messed up my script
It should be `Hint Mode Class ! : typeclass_instances.`

PR #8956, opened 2018-11-09
Fix dune runtest invocation

SkySkimmer:
`dune --display=short runtest` would error with "too many arguments", ie the --display option needs to be after the `runtest` subcommand.

### DIFF
--- a/Makefile.dune
+++ b/Makefile.dune
@@ -53,7 +53,7 @@ quickopt: voboot
 	dune build $(DUNEOPT) $(QUICKOPT_TARGETS)
 
 test-suite: voboot
-	dune $(DUNEOPT) runtest
+	dune runtest $(DUNEOPT)
 
 release: voboot
 	dune build $(DUNEOPT) -p coq


SkySkimmer:
Note however that it prints nothing (after the main build is done):
~~~
dune runtest --display=progress
Done: 11493/11496 (jobs: 1)
~~~
--display=short prints even less (literally nothing)
--display=verbose prints the make invocation but not its output

ejgallego:
Progressive display of the test suite is indeed work in progress; by default Dune waits for a command to complete before displaying its output [I believe this is due to avoid races that happen wrt to display otherwise]

I'd love to make Dune more aware of the test suite and test dependencies but no point on doing this while we have to keep make compat.

I cannot reproduce the lack of output in your case, in my setup, no matter how I run `runtest` I get the output of the test suite as soon as it is done.

ejgallego (approved):

SkySkimmer:
> I cannot reproduce the lack of output in your case, in my setup, no
 > matter how I run |runtest| I get the output of the test suite as soon as
 > it is done.

Probably because I didn't wait for it to be done ;)

Gaëtan Gilbert

ejgallego:
> Probably because I didn't wait for it to be done ;)

Yeah that's the likely cause; I vaguely recall that there was an option to avoid this but I cannot locate it now.

SkySkimmer:
man dune build
~~~
        --no-buffer
            Do not buffer the output of commands executed by dune. By 
default dune buffers the
            output of subcommands, in order to prevent interleaving when 
multiple commands are
            executed in parallel. However, this can be an issue when 
debugging long running tests.
            With --no-buffer, commands have direct access to the 
terminal. Note that as a result
            their output won't be captured in the log file. You should 
use this option in
            conjunction with -j 1, to avoid interleaving. Additionally 
you should use --verbose as
            well, to make sure that commands are printed before they are 
being executed.
~~~

Gaëtan Gilbert

ejgallego:
> man dune build

Oh indeed it was a cmdline option not an option in the stanza; thanks!

It may make sense to run `runtest` with this flag for the moment.

Issue #8953, opened 2018-11-09
[dune] installation of coqide is partial


gares:
![bug-dune](https://user-images.githubusercontent.com/1013846/48263009-efec5a80-e424-11e8-9a55-49999dea6720.png)

after doing (in a fresh checkout)
```
make -f Makefile.dune world
_build/install/default/bin/coqide
```

(BTW is there a way to get the binaries "closer" to where you are?)

gares:
You see the language combo is empty. It is made of the files in `ide/*.lang`, that I guess are not installed.

ejgallego:
Indeed this was a TODO; submitting a PR; however we may need tweak the rules for coqide to locate these files.

IMHO it would make sense first to install them in `share/coqide` instead of `share/coq`.

Also, I am not familiar with how CoqIDE will access the `share` path; usually some kind of helper library as the one proposed in https://github.com/ocaml/dune/issues/1185 would be useful.

PR #8959, opened 2018-11-09
[dune] [ide] Install data files.

ejgallego:
We should install the files in `share/coqide` instead of the current
`coq` location; but we defer this change until we are more advanced in
the make-phase out.

Fixes: #8953

### DIFF
--- a/ide/dune
+++ b/ide/dune
@@ -43,3 +43,14 @@
  (package coqide)
  (modules coqide_main)
  (libraries coqide.gui))
+
+; FIXME: we should install those in share/coqide. We better do this
+; once the make-based system has been phased out.
+(install
+ (section share_root)
+ (package coqide)
+ (files
+  (coq.png as coq/coq.png)
+  (coq_style.xml as coq/coq_style.xml)
+  (coq.lang as coq/coq.lang)
+  (coq-ssreflect.lang as coq/coq-ssreflect.lang)))


gares:
I'm trying to test this, is there a way to `@install` only coqide? (without installing and compiling the stdlib)

gares (approved):

ejgallego:
> I'm trying to test this, is there a way to `@install` only coqide? (without installing and compiling the stdlib)

`dune build coqide.install`

ejgallego:
Also see the docs about the `-p coqide` flag, but this won't do what you'd like.

ejgallego:
In general Dune has no so much magic, it is still a build system + some sugar. 

As I've noted in other msg the install file depends on all the installed files, then the installed file is build depending on a set of rules, etc...

When you do `(library (public_name foo) ...)` this in fact desugars to `(install ...)` stanzas. The ML code is straightforward.

gares:
> `dune build coqide.install`

I tired that, but does not install the idetop thing, so it is unusable.

ejgallego:
Oh you asked for the IDE. The XML language server is in the coqide-server package.

ejgallego:
As we want to build the server even it gtk is not available.

ejgallego:
I am convinced the decoupling makes sense, but it could be discussed.

gares:
To sum up, what should I type in order to get a coqide that works?

ejgallego:
> To sum up, what should I type in order to get a coqide that works?
```
$ make -f Makefile.dune world
```
or we could add a more refined target:
```
$ make -f Makefile.dune voboot # not needed on incremental builds
$ dune build coqide-server.install
$ dune build coqide.install # or any other more refined target
```

gares:
> or we could add a more refined target:

Yes please, call this target coqide (assuming it also builds the prelude)

ppedrot (commented):
Not qualified to review this, @gares seems more indicated.

Zimmi48:
@ejgallego: Don't you want to address Enrico's suggestion of adding a coqide alias.

ejgallego:
@Zimmi48 that was done in #9006

PR #8961, opened 2018-11-09
dune: link kernel in checker instead of copying files

SkySkimmer:
This allows to use the nice printers with constrextern etc, and since
we were copying everything we're not linking any more than previously.

Also the dune file is simpler without the copies.


ppedrot (approved):
Why not.

maximedenes (approved):

maximedenes:
@ppedrot can you be the assignee?

ppedrot:
@SkySkimmer this needs a rebase.

SkySkimmer:
Rebased.

PR #8966, opened 2018-11-10
coq_makefile: Fix ocamldep ignoring mlg files

SkySkimmer:
If you have file1.mlg and file2.ml, with file2 depending on file1,
ocamldep will be run before file1.ml exists so won't generate
[file2.cmx: file1.cmx] (ocamldep is silent on non-found dependencies).

This has been causing nondeterministic failures in quickchick
recently.

I guess it didn't come up in the past because ml4 files tend to be at
the end of the dependency chain.

### DIFF
--- a/test-suite/coq-makefile/camldep/run.sh
+++ b/test-suite/coq-makefile/camldep/run.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+export PATH=$COQBIN:$PATH
+export LC_ALL=C
+
+rm -rf _test
+mkdir _test
+cp _CoqProject _test/
+cd _test
+mkdir src
+
+echo '{ let foo = () }' > src/file1.mlg
+echo 'let bar = File1.foo' > src/file2.ml
+coq_makefile -f _CoqProject -o Makefile
+make src/file2.cmx
+[ -f src/file2.cmx ]
--- a/test-suite/coq-makefile/camldep/_CoqProject
+++ b/test-suite/coq-makefile/camldep/_CoqProject
@@ -1,0 +1,4 @@
+-Q . Foo
+-I src
+src/file1.mlg
+src/file2.ml
--- a/tools/CoqMakefile.in
+++ b/tools/CoqMakefile.in
@@ -719,6 +719,9 @@ endif
 
 redir_if_ok = > "$@" || ( RV=$$?; rm -f "$@"; exit $$RV )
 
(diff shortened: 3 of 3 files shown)


test-suite/coq-makefile/camldep/run.sh
```diff
@@ -0,0 +1,17 @@
...
+export LC_ALL=C
+
+rm -rf _test
+mkdir _test
+cp _CoqProject _test/
+cd _test
+mkdir src
+
+echo '{ let foo = () }' > src/file1.mlg
+echo 'let bar = File1.foo' > src/file2.ml
```
gares: Why don't you make the template cover this case?

gares (approved):

test-suite/coq-makefile/camldep/run.sh
SkySkimmer: I prefer to avoid the template as it confuses me.

SkySkimmer:
Can someone merge this quickly, quickchick keeps dying in CI because of the bug this fixes.

ejgallego:
Umm, is the fix you propose right? AFAICT `.mlg` and `ml4` files are not ML files so `-ml-synonym` seems incorrect here.

It seems like the bug here is that `ocamldep` should be ran _after_ pre-processing, shouldn't it?

SkySkimmer:
>It seems like the bug here is that ocamldep should be ran after pre-processing, shouldn't it?

Good point, changed.

ejgallego (approved):
Not enough expert in make but it looks good to me.

ejgallego:
I will take care of this as it seems it went under @gares radar and the CI failures are really creating a lot of noise [and wasted cycles and indeed we will merge something really breaking QuickChick and then we'll have a bigger mess]

silene:
This commit does not make much sense to me. In fact, its backport severely breaks the 8.9 branch, as can be seen here: https://gitlab.com/coq/coq/pipelines/39198401

Indeed, whenever there is a `foo.ml4` file, it makes the `*.ml.d` files depend on `foo.ml`. The thing is, there is no such thing, as file `foo.ml4` is directly compiled to `foo.cmo` and `foo.cmx`.

Also, there seems to have been quite a misconception earlier in this thread. `-ml-synonym .ml4` does not mean that `foo.ml4` is an OCaml file. It means that, if `bar.ml` performs `open Foo` and if `foo.ml4` exists, then `bar.cmx` should depend on `foo.cmx`.

PR #8967, opened 2018-11-11
Fix #8922 (uncaught pp_diff exception)

### LINKED ISSUE #8922: Uncaught exception Pp_diff.Diff_Failure
#### Version
master

#### Operating system
Fedora

I tried the new feature that colours the changes in the goal, and encountered the following bug:
```
Lemma eza : True.
Proof.
  eassert(_).
  refine (fun _ => _).
```

```
Anomaly
"Uncaught exception Pp_diff.Diff_Failure("Unable to match goals betwen old and new proof states (6)")."
```

jfehrle:
Fixed several problems exposed by #8922.  Without the change, diffs will omit some output lines when displaying the proof context, making the diff unacceptably unreliable.  While users can turn off diffs, this fix should go into 8.9.

Get hyps and goal the same way Printer does; don't omit info
Allow for new goals that don't map to old goals
Include background_goals in all_goals return value
Fix incorrect change to raw diffs in shorten_diff_span

**Kind:** bug fix

Fixes #8922

### DIFF
--- a/ide/idetop.ml
+++ b/ide/idetop.ml
@@ -212,25 +212,20 @@ let goals () =
     if Proof_diffs.show_diffs () then begin
       let oldp = Stm.get_prev_proof ~doc (Stm.get_current_state ~doc) in
       let diff_goal_map = Proof_diffs.make_goal_map oldp newp in
-      let map_goal_for_diff ng = (* todo: move to proof_diffs.ml *)
-        try Evar.Map.find ng diff_goal_map  with Not_found -> ng
-      in
 
       let process_goal_diffs nsigma ng =
         let open Evd in
-        let og = map_goal_for_diff ng in
         let og_s = match oldp with
         | Some oldp ->
           let (_,_,_,_,osigma) = Proof.proof oldp in
-          Some { it = og; sigma = osigma }
+          (try Some { it = Evar.Map.find ng diff_goal_map; sigma = osigma }
+          with Not_found -> raise (Pp_diff.Diff_Failure "Unable to match goals between old and new proof states (6)"))
         | None -> None
         in
         let (hyps_pp_list, concl_pp) = Proof_diffs.diff_goal_ide og_s ng nsigma in
         { Interface.goal_hyp = hyps_pp_list; Interface.goal_ccl = concl_pp; Interface.goal_id = Goal.uid ng }
       in
-      try
-        Some (export_pre_goals (Proof.map_structured_proof newp process_goal_diffs))
-      with Pp_diff.Diff_Failure _ -> Some (export_pre_goals (Proof.map_structured_proof newp process_goal))
+      Some (export_pre_goals (Proof.map_structured_proof newp process_goal_diffs))
     end else
       Some (export_pre_goals (Proof.map_structured_proof newp process_goal))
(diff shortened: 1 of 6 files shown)


ejgallego (approved):
Looks OK to me.

ide/idetop.ml
```diff
@@ -212,25 +212,19 @@ let goals () =
...
-      in
 
       let process_goal_diffs nsigma ng =
         let open Evd in
-        let og = map_goal_for_diff ng in
         let og_s = match oldp with
         | Some oldp ->
           let (_,_,_,_,osigma) = Proof.proof oldp in
-          Some { it = og; sigma = osigma }
+          (try Some { it = Evar.Map.find ng diff_goal_map; sigma = osigma } with Not_found -> None)
```
herbelin: Hi @jfehrle, by curiosity, is this `try with Not_found` needed because you want to possibly parse strings of the form `?X = ?Y` with unbound `?X` and `?Y`, or is this `try` needed to fix #8922?

jfehrle: Short answer: no, this is not related to parsing.

Proof_diffs.make_goal_map uses a heuristic to match new goals to old goals by looking at the old and new proof objects, which works fine for the cases I tried.  One of the problems uncovered by this bug report is a goal that could not be matched.  In the example below (after the `refine`), `?Goal1` doesn't appear in the `Show Proof`, so it couldn't be matched.  My initial fix was to add the `try with Not_found` here, which would treat any such goals as new, which would show them as new, fully highlighting them.

A day or two later I discovered that Proof.all_goals doesn't include shelved goals.  Including those goals fixed the matching problem of this bug in a cleaner way.  Although `?Goal1` seems a bit of a strange corner case to me (not appearing in the proof object at all), I'm not sure that change covers all the possibilities.

I think I will change the code so it raises Diff_failure instead as a way of surfacing other unhandled cases if they indeed exist.  (Note that users can easily work around such anomalies by turning off diffs for the proof steps getting anomalies.)

```
eza < Show Proof.
(let X := fun H : ?Goal0 => ?Goal2@{y:=H} in ?Goal)

eza < Show Existentials.
Existential 1 = ?Goal : [X : forall x : ?Goal0, ?Goal1 |- True]
Existential 2 = ?Goal0 : [ |- Type] (shelved)
Existential 3 = ?Goal1 : [x : ?Goal0 |- Type] (shelved)
Existential 4 = ?Goal2 : [y : ?Goal0 |- ?Goal1@{x:=y}]
```

herbelin: OK, thanks for the explanation.

ppedrot (approved):
The change to Proof is OK and minor btw.

herbelin:
> @herbelin and @ejgallego, which one of you will do the merge?

I let @ejgallego decide.

ejgallego:
> I let @ejgallego decide.

@herbelin I didn't have time to look into this PR sadly, please feel free to assign it.

herbelin:
> please feel free to assign it.

OK.

I did not look at all detals either but this is a fix which looks ok. So, I'll merge on Wednesday if no comments by then.

jfehrle:
Thank you for the review, gentlemen.

herbelin:
And thank you for your work!

PR #8968, opened 2018-11-11
Miscellaneous CoqIDE fixes

herbelin:
**Kind:** bug fix

This fixes the following issues:
- The nanoPG/emacs mode was preventing the ↓ and ↑ keys to work in the completion window
We fix it by not aliasing ↓ and ↑ in nanoPG (useless since they already have an action by default - can an expert confirm)
- The Ctrl-/ menu item for old-style completion was obsolete: we remove it.
- The preference and save windows had the annoying habit to get hidden by the main window resulting in a blocked main window w/o necessarily knowing that it was blocked by another window hidden behind, nor w/o necessarily being able to easily get back the hidden window in front.
 We ensure that the dialogue boxes stay in front, using the `parent` field of `GWindow.dialog` and cie.

I'm a gtk+ newbie, so maybe I did not follow the best road. Double-check by experts welcome.

Relatively to the ongoing discussions on GUIs, it might look anachronic to improve CoqIDE, but until there is no realistic alternatives to CoqIDE and PG, I don't see why to refrain ouselves to work at improving our _daily environments_ in the meantime. Moreover, by learning about GTK+, I'm sure that this shall be useful too when working later on other kinds of GUIs.

Also, should we bind CoqIDE completion to #8766? I'm a bit lost in the discussion around it.

### DIFF
--- a/ide/configwin.ml
+++ b/ide/configwin.ml
@@ -46,6 +46,6 @@ let modifiers = Configwin_ihm.modifiers
 
 let edit
     ?(apply=(fun () -> ()))
-    title ?width ?height
+    title ?parent ?width ?height
     conf_struct_list =
-  Configwin_ihm.edit ~with_apply: true ~apply title ?width ?height conf_struct_list
+  Configwin_ihm.edit ~with_apply: true ~apply title ?parent ?width ?height conf_struct_list
--- a/ide/configwin.mli
+++ b/ide/configwin.mli
@@ -158,6 +158,7 @@ val custom : ?label: string -> GPack.box -> (unit -> unit) -> bool -> parameter_
 val edit :
   ?apply: (unit -> unit) ->
   string ->
+  ?parent:GWindow.window ->
   ?width:int ->
   ?height:int ->
   configuration_structure list ->
--- a/ide/configwin_ihm.ml
+++ b/ide/configwin_ihm.ml
@@ -808,3 +808,40 @@ let custom ?label box f expand =
       custom_expand = expand ;
       custom_framed = label ;
     }
+
+(* Copying lablgtk question_box + forbidding hiding *)
+
+let question_box ~title ~buttons ?(default=1) ?icon ?parent message =
+  let button_nb = ref 0 in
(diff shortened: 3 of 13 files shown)


ppedrot (approved):

ppedrot:
CI failures are spurious, let's merge.

silene:
As the behavior of transient windows highly depends on the window manager, I don't feel like backporting this pull request to 8.9 just before the release. At this point, the only part I am confident with is the "complete word" commit, in case you would like to open a pull request for it.

herbelin:
> As the behavior of transient windows highly depends on the window manager

Hi @silene, I'm not well aware of all window manager policies. Do you see a possible problem in some situations? Is there more to fear than the window manager ignoring the relation of the "transient" window with the parent, which would not be worse than the situation of today?

In any case, ok for postponing. I'm going to make a 8.9 PR for the completion change.

herbelin:
By the way, there will be the issue of providing a lablgtk3-based coqide for the purpose of Debian freezing very soon. How do you see this happening?

Shall we plan a 8.8.3 with coqide based on gtk3? Shall we release 8.9 with coqide already based on gtk3?

The current status is that everything seems to work nicely with the exception of
- the progress bar being deactivated (need to move from pixmap to cairo)
- no more file chooser to find `coqtop` when the latter is not found
- the configuration for automatic tactics and user query shortcuts not working

So, nothing critical in any case, even if the progression bar is cool and maybe the next priority in finishing the port.

silene:
> Shall we plan a 8.8.3 with coqide based on gtk3?

No strong opinion on this one.

> Shall we release 8.9 with coqide already based on gtk3?

No. I want to release 8.9 as soon as possible. I am in favor of a 8.9.1 release that would use Gtk3, though.

> no more file chooser to find coqtop when the latter is not found

What is wrong with the file chooser? I don't remember encountering any issue when working on the port of Why3 to Gtk3.

herbelin:
> No. I want to release 8.9 as soon as possible. I am in favor of a 8.9.1 release that would use Gtk3, though.

OK.

> What is wrong with the file chooser? I don't remember encountering any issue when working on the port of Why3 to Gtk3.

The coqtop selection uses `file_selection` which does not exist anymore. It should use instead `file_chooser`.

Issue #8951, opened 2018-11-09
Kernel anomaly with universes and inlining


maximedenes:
I don't understand how univ poly and functor inlining work together, so I wrote one example:

```coq
Module Type T.
  Polymorphic Parameter Inline t@{i} : Type@{i}.
End T.

Module M.
  Polymorphic Definition t@{i} := nat.
End M.

Module Make (X:T).
  Include X.

End Make.

Module P := Make M.
```

and got:

```
Anomaly "File "kernel/mod_subst.ml", line 306, characters 15-21: Assertion failed."
Please report at http://coq.inria.fr/bugs/.
```

I need guidance on how to fix this.

ppedrot:
This most likely indicates some serious bug in the kernel... Luckily there was an assertion there.

SkySkimmer:
>I need guidance on how to fix this.

Was the stuff I said on gitter not enough?

maximedenes:
> ... but this is worse than that. Even `subst_constant` applies terms to empty instances... And that one is used by the VM.

Yeah, I saw that too. And outside of the kernel, there's more of it: `subst_global` etc.

ppedrot:
@maximedenes You can remove most of the calls to `subst_global` and use `subst_global_reference` instead. I am writing a patch for that, but you might already be doing the same.

maximedenes:
> I am writing a patch for that, but you might already be doing the same.

It's already done before I submitted this, let's synchronize.

JasonGross:
Can this be used to get a proof of False

SkySkimmer:
No, the assert catches attempts.

Gaëtan Gilbert

PR #8971, opened 2018-11-12
Clean-up around substitution of globals

maximedenes:
We clean up the kernel API to not expose unsafe combinators (that silently insert dummy universe instances).

This is related to #8951 but does not solve it per se. It just makes the problem more visible on the user side (in that case, `library/heads.ml`).

### DIFF
--- a/library/globnames.ml
+++ b/library/globnames.ml
@@ -29,6 +29,11 @@ let destConstRef = function ConstRef ind -> ind | _ -> failwith "destConstRef"
 let destIndRef = function IndRef ind -> ind | _ -> failwith "destIndRef"
 let destConstructRef = function ConstructRef ind -> ind | _ -> failwith "destConstructRef"
 
+let subst_constructor_reference subst (ind,j as ref) =
+  let ind' = subst_ind subst ind in
+    if ind==ind' then ref
+    else (ind',j)
+
 let subst_constructor subst (ind,j as ref) =
   let ind' = subst_ind subst ind in
     if ind==ind' then ref, mkConstruct ref
@@ -43,7 +48,7 @@ let subst_global_reference subst ref = match ref with
     let ind' = subst_ind subst ind in
       if ind==ind' then ref else IndRef ind'
   | ConstructRef ((kn,i),j as c) ->
-    let c',t = subst_constructor subst c in
+    let c' = subst_constructor_reference subst c in
       if c'==c then ref else ConstructRef c'
 
 let subst_global subst ref = match ref with
--- a/library/globnames.ml
+++ b/library/globnames.ml
@@ -54,7 +54,7 @@ let subst_global_reference subst ref = match ref with
 let subst_global subst ref = match ref with
   | VarRef var -> ref, mkVar var
   | ConstRef kn ->
-     let kn',t = subst_con_kn subst kn in
+     let kn',t = subst_con subst (Univ.in_punivs kn) in
       if kn==kn' then ref, mkConst kn else ConstRef kn', t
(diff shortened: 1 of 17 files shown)


mattam82 (approved):
Cleanup looks good to me. IIUC, there are places where we do the wrong thing in case inlining produces a term rather than a global?

maximedenes:
> Cleanup looks good to me. IIUC, there are places where we do the wrong thing in case inlining produces a term rather than a global?

Unfortunately, the most dangerous cases are when we don't inline, and `subst_con0` puts back the dummy universe instance...

maximedenes:
In order to resolve cleaning the universe poly vs inlining tension, I experimented removing functor inlining, and improving support for aliases in tactics instead. I went pretty far, but I'd need an answer to #5301, to make hints not rely on inlining.

library/globnames.ml
```diff
@@ -43,13 +48,13 @@ let subst_global_reference subst ref = match ref with
...
   | ConstructRef ((kn,i),j as c) ->
-    let c',t = subst_constructor subst c in
+    let c' = subst_constructor_reference subst c in
       if c'==c then ref else ConstructRef c'
 
 let subst_global subst ref = match ref with
   | VarRef var -> ref, mkVar var
   | ConstRef kn ->
-     let kn',t = subst_con_kn subst kn in
+     let kn',t = subst_con subst (Univ.in_punivs kn) in
```
SkySkimmer: This should have some `XXX` comment.

pretyping/classops.ml
```diff
@@ -192,7 +192,7 @@ let subst_cl_typ subst ct = match ct with
     let c' = subst_proj_repr subst c in
       if c' == c then ct else CL_PROJ c'
   | CL_CONST c ->
-      let c',t = subst_con_kn subst c in
+      let c',t = subst_con subst (Univ.in_punivs c) in
```
SkySkimmer: Should have a comment `XXX TODO FIXME univs`

pretyping/heads.ml
```diff
@@ -156,7 +156,7 @@ let cache_head o =
 
 let subst_head_approximation subst = function
   | RigidHead (RigidParameter cst) as k ->
-      let cst,c = subst_con_kn subst cst in
+      let cst,c = subst_con subst (Univ.in_punivs cst) in
```
SkySkimmer: `FIXME XXX TODO univs`

ppedrot (approved):
LGTM modulo the changes mentioned by @SkySkimmer. There is a call to `subst_global` in tacsubst.ml that can be removed as well, it triggers a warning that is not removable and not using the warning system, so it's better to just get rid of this check.

maximedenes:
I think I prefer #8974 , but even better if we can really fix the problems.

library/globnames.ml
herbelin: It is unclear what is the intended effect of the substitution wrt universes. If I have a module depending on a polymorphic name and the name is substituted by a name (which is the common case), do I want all the occurrences of the name to share the same universe instance or do I want each occurrence to have fresh universes?

Then, if I substitute a name by a term (due to `Inline`), do I want each occurrence to be substituted by an instance with new fresh universes? I would say yes. But if yes, this means that we should change the type of subst_global so that it returns not a `constr` but the closure of a `constr` over its universe parameters.

@mattam82, what do you think?

herbelin: I talked too quickly. Reading the discussions, #8974 seems a very good direction.

maximedenes:
Closing in favor of #8974.

PR #8972, opened 2018-11-12
Fix #4771: Substitution of inline global reference in tactics is broken

### LINKED ISSUE #4771: Substitution of inline global reference in tactics is broken

Original bug ID: BZ#4771
From: @ppedrot
Reported version: trunk
CC:   @JasonGross

maximedenes:
Fixes #4771. The behavior I chose is to not inline the reference in places where it does not make sense.

### DIFF
--- a/plugins/ltac/tacsubst.ml
+++ b/plugins/ltac/tacsubst.ml
@@ -88,20 +88,9 @@ let subst_reference subst =
 (*CSC: subst_global_reference is used "only" for RefArgType, that propagates
   to the syntactic non-terminals "global", used in commands such as
   Print. It is also used for non-evaluable references. *)
-open Pp
-open Printer
 
 let subst_global_reference subst =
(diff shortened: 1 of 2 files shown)


ppedrot (approved):

ppedrot:
Failures are unrelated.

ppedrot:
Also, note that I suspect this is still wrong when self-including functors, but this has little to do with this feature per se though.

maximedenes:
> Also, note that I suspect this is still wrong when self-including functors, but this has little to do with this feature per se though.

If there is still a problem, maybe we should confirm it and keep track of it? What kind of examples did you have in mind? Substitutions by your famous names which don't really exists?

ppedrot:
> Substitutions by your famous names which don't really exist?

Indeed. I don't really know how to generate them, but I suspect those Higgs-boson constants exist.

PR #8974, opened 2018-11-12
Fix mod_subst wrt universe polymorphism

SkySkimmer:
@maximedenes this is how I would do #8971 (and as such is incompatible with it).

Based on #8972.

### DIFF
--- a/interp/notation_ops.ml
+++ b/interp/notation_ops.ml
@@ -530,8 +530,10 @@ let rec subst_notation_constr subst bound raw =
   match raw with
   | NRef ref ->
       let ref',t = subst_global subst ref in
-	if ref' == ref then raw else
-	  fst (notation_constr_of_constr bound t)
+      if ref' == ref then raw else (match t with
+          | None -> NRef ref'
+          | Some t ->
+            fst (notation_constr_of_constr bound t.Univ.univ_abstracted_value))
 
   | NVar _ -> raw
 
--- a/kernel/mod_subst.ml
+++ b/kernel/mod_subst.ml
@@ -24,7 +24,7 @@ open Constr
    is the term into which we should inline. *)
 
 type delta_hint =
-  | Inline of int * (Univ.AUContext.t * constr) option
+  | Inline of int * constr Univ.univ_abstracted option
   | Equiv of KerName.t
 
 (* NB: earlier constructor Prefix_equiv of ModPath.t
@@ -164,7 +164,7 @@ let find_prefix resolve mp =
 
 (** Applying a resolver to a kernel name *)
 
-exception Change_equiv_to_inline of (int * (Univ.AUContext.t * constr))
+exception Change_equiv_to_inline of (int * constr Univ.univ_abstracted)
 
 let solve_delta_kn resolve kn =
   try
@@ -294,43 +294,34 @@ let subst_ind sub (ind,i as indi) =
 let subst_pind sub (ind,u) =
   (subst_ind sub ind, u)
 
-let subst_con0 sub (cst,u) =
+let subst_con0 sub cst =
   let mpu,l = Constant.repr2 cst in
   let mpc = KerName.modpath (Constant.canonical cst) in
   let mpu,mpc,resolve,user = subst_dual_mp sub mpu mpc in
   let knu = KerName.make mpu l in
   let knc = if mpu == mpc then knu else KerName.make mpc l in
   match search_delta_inline resolve knu knc with
-    | Some (ctx, t) ->
+    | Some t ->
       (* In case of inlining, discard the canonical part (cf #2608) *)
-      let () = assert (Int.equal (Univ.AUContext.size ctx) (Univ.Instance.length u)) in
-      Constant.make1 knu, Vars.subst_instance_constr u t
+      Constant.make1 knu, Some t
     | None ->
       let knc' =
         progress (kn_of_delta resolve) (if user then knu else knc) ~orelse:knc
       in
       let cst' = Constant.make knu knc' in
-      cst', mkConstU (cst',u)
+      cst', None
 
 let subst_con sub cst =
   try subst_con0 sub cst
-  with No_subst -> fst cst, mkConstU cst
+  with No_subst -> cst, None
 
-let subst_con_kn sub con =
-  subst_con sub (con,Univ.Instance.empty)
-
-let subst_pcon sub (_con,u as pcon) =
-  try let con', _can = subst_con0 sub pcon in
+let subst_pcon sub (con,u as pcon) =
+  try let con', _can = subst_con0 sub con in
 	con',u
   with No_subst -> pcon
 
-let subst_pcon_term sub (_con,u as pcon) =
-  try let con', can = subst_con0 sub pcon in 
-	(con',u), can
-  with No_subst -> pcon, mkConstU pcon
-
 let subst_constant sub con =
-  try fst (subst_con0 sub (con,Univ.Instance.empty))
+  try fst (subst_con0 sub con)
   with No_subst -> con
 
 let subst_proj_repr sub p =
@@ -351,7 +342,7 @@ let subst_evaluable_reference subst = function
 let rec map_kn f f' c =
   let func = map_kn f f' in
     match kind c with
-      | Const kn -> (try snd (f' kn) with No_subst -> c)
+      | Const kn -> (try f' kn with No_subst -> c)
       | Proj (p,t) -> 
           let p' = Projection.map f p in
 	  let t' = func t in
@@ -420,8 +411,14 @@ let rec map_kn f f' c =
       | _ -> c
 
 let subst_mps sub c =
+  let subst_pcon_term sub (con,u) =
+    let con', can = subst_con0 sub con in
+    match can with
+    | None -> mkConstU (con',u)
+    | Some t -> Vars.univ_instantiate_constr u t
+  in
   if is_empty_subst sub then c
-  else map_kn (subst_mind sub) (subst_con0 sub) c
+  else map_kn (subst_mind sub) (subst_pcon_term sub) c
 
 let rec replace_mp_in_mp mpfrom mpto mp =
   match mp with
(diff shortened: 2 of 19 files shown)


kernel/vars.mli
```diff
@@ -140,4 +140,6 @@ val subst_univs_level_context : Univ.universe_level_subst -> Constr.rel_context
 val subst_instance_constr : Instance.t -> constr -> constr
 val subst_instance_context : Instance.t -> Constr.rel_context -> Constr.rel_context
 
+val univ_unabstract_constr : Instance.t -> constr univ_abstracted -> constr
```
maximedenes: unabstract -> apply ?

maximedenes (commented):
I like it, but it is disappointing that we have to expose `univ_abstracted`. Is there any hope we can fix the problems (in clients), so that we can expose a safer API?

SkySkimmer:
>I like it, but it is disappointing that we have to expose univ_abstracted. Is there any hope we can fix the problems (in clients), so that we can expose a safer API?

I don't know, there's no point forcing people to provide an instance and substitute when we're just going to look for the head globref or whatever, no? And I'm not unhappy with the univ_abstracted, someday I'll try to add some in const_body.

kernel/vars.mli
SkySkimmer: Only if you insist.

maximedenes:
> I don't know, there's no point forcing people to provide an instance and substitute when we're just going to look for the head globref or whatever, no?

But then we should provide an API that inlines only to globrefs, I guess... We wouldn't go through `constr`.

kernel/vars.mli
mattam82: `instantiate` would fit even better with the rest of the naming scheme if I may.

mattam82 (approved):
The code looks good to me, I defer to @maximedenes about the cleanup of the mod_subst API otherwise.

kernel/vars.mli
maximedenes: Indeed!

SkySkimmer:
For the [eq -> eq t] inlining that won't work

Gaëtan Gilbert

maximedenes:
> For the [eq -> eq t] inlining that won't work

Why not? The head globref will still be the new `eq`, right?

SkySkimmer:
But it's not a globref -> globref inlining, it's globref -> constr.

maximedenes:
> But it's not a globref -> globref inlining, it's globref -> constr.

But precisely, if the client is concerned only about the head, like you were writing, we could encapsulate that as a globref -> globref inlining.

SkySkimmer:
That makes no sense to me TBH

Gaëtan Gilbert

tactics/hints.ml
```diff
@@ -1060,12 +1060,12 @@ let cache_autohint (kn, obj) =
...
-    let gr' =
-      (try head_constr_bound Evd.empty elab'
-       with Bound -> lab'')
-    in if gr' == gr then gr else gr'
+    let (gr', t) = subst_global subst gr in
+    match t with
+    | None -> gr'
+    | Some t ->
+      (try head_constr_bound Evd.empty (EConstr.of_constr t.Univ.univ_abstracted)
+       with Bound -> gr')
```
herbelin: An alternative could be to remove the hint if no head constant appears anymore.

kernel/univ.ml
```diff
@@ -963,6 +963,15 @@ struct
 
 end
 
+type 'a univ_abstracted = {
+  univ_abstracted : 'a;
+  bound_univs : Names.Name.t array
+}
```
herbelin: You could also use the conventional name of "closure". 

Also, at first, I did not get that the projection `univ_abstracted` meant the body. So, maybe `univ_closure_body` and `univ_closure_binders`?

kernel/univ.mli
```diff
@@ -349,6 +349,15 @@ sig
 
 end
 
+type 'a univ_abstracted = {
+  univ_abstracted : 'a;
+  bound_univs : Names.Name.t array
+}
+(** A value containing DeBruijn universe levels (with names bound
```
herbelin: The purists would say "de Bruijn"

kernel/univ.ml
herbelin: By the way, why a list of names rather than a `AUContext.t`? Even if you explain explicitly later that the bindings are à la de Bruijn, it is a bit misleading to see names, as if the universes were bound by names.

herbelin (commented):
Very nice.

pretyping/detyping.ml
```diff
@@ -935,10 +935,13 @@ let (f_subst_genarg, subst_genarg_hook) = Hook.make ()
...
-    if ref' == ref then raw else
-      let env = Global.env () in
-      let evd = Evd.from_env env in
-      DAst.get (detype Now false Id.Set.empty env evd (EConstr.of_constr t))
+    if ref' == ref then raw else (match t with
+        | None -> GRef (ref', u)
+        | Some t ->
+          let env = Global.env () in
+          let evd = Evd.from_env env in
+          let t = t.Univ.univ_abstracted in (* XXX This seems dangerous *)
```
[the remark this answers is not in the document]
SkySkimmer: However u is not an instance (it's a glob_instance ie a list of level names) so it's not easy to do.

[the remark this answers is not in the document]
herbelin: This code is basically for substituting the `glob_constr` occurring in LtacX code. At first view, in such code, the instance can only be made of global universes. Would using `Pretyping.interp_instance` be ok?

More upstream, it seems that Ltac code should detect universe errors earlier. For instance, the following leads to a wrong-instance-length runtime error:
```
Ltac f x := exact (id@{Type} x). 
```
and this would also break at the time of an `Inline` substitution of functor arguments.

kernel/univ.ml
SkySkimmer: Wouldn't a closure contain values for the universes rather than binder names?

kernel/univ.mli
```diff
@@ -349,6 +349,14 @@ sig
 
 end
 
+type 'a univ_abstracted = {
+  univ_abstracted : 'a;
+  nbound_univs : int;
+}
+(** A value with [nbound_univs] bound universe levels. *)
```
SkySkimmer: Changed to use just an `int` as we have no use for more, if that changes we can put whatever we need in the type later.

kernel/univ.ml
ppedrot: I support reusing the `AUContext.t` type. It is more informative, and that information might end up being needed, or allowing to thread data in a uniform way.

herbelin: > Wouldn't a closure contain values for the universes rather than binder names?

You're right. My confusion.

I still feel that the current terminology can be made more appealing though. So, maybe `univ_abstracted_body` and `univ_abstracted_binders`. Or `abstraction` in place of `abstracted`. Of course, that's pretty subjective, so maybe other opinions would help to move from subjective to more objective.

SkySkimmer:
Now using an `AUContext.t` and names `univ_abstracted_value` `univ_abstracted_binder`.

interp/notation_ops.ml
```diff
@@ -530,8 +530,10 @@ let rec subst_notation_constr subst bound raw =
   match raw with
   | NRef ref ->
       let ref',t = subst_global subst ref in
-	if ref' == ref then raw else
-	  fst (notation_constr_of_constr bound t)
+      if ref' == ref then raw else (match t with
+          | None -> NRef ref'
+          | Some t ->
+            fst (notation_constr_of_constr bound t.Univ.univ_abstracted_value))
```
ppedrot: This is wrong, but if you hit this bug you probably deserve it.

kernel/mod_subst.ml
```diff
@@ -294,43 +294,36 @@ let subst_ind sub (ind,i as indi) =
...
-let subst_con0 sub (cst,u) =
+let subst_con0 sub cst =
   let mpu,l = Constant.repr2 cst in
   let mpc = KerName.modpath (Constant.canonical cst) in
   let mpu,mpc,resolve,user = subst_dual_mp sub mpu mpc in
   let knu = KerName.make mpu l in
   let knc = if mpu == mpc then knu else KerName.make mpc l in
   match search_delta_inline resolve knu knc with
-    | Some (ctx, t) ->
-      (* In case of inlining, discard the canonical part (cf #2608) *)
```
ppedrot: You should probably preserve this comment.

kernel/mod_subst.mli
```diff
@@ -133,16 +133,13 @@ val subst_kn :
...
-  substitution -> pconstant -> Constant.t * constr
+  substitution -> Constant.t -> Constant.t * constr Univ.univ_abstracted option
 
 val subst_pcon :
   substitution -> pconstant -> pconstant
 
-val subst_pcon_term :
-  substitution -> pconstant -> pconstant * constr
-
 val subst_con_kn :
```
ppedrot: Remove this function.

kernel/vars.mli
```diff
@@ -140,4 +140,6 @@ val subst_univs_level_context : Univ.universe_level_subst -> Constr.rel_context
 val subst_instance_constr : Instance.t -> constr -> constr
 val subst_instance_context : Instance.t -> Constr.rel_context -> Constr.rel_context
 
+val univ_instantiate_constr : Instance.t -> constr univ_abstracted -> constr
```
ppedrot: This is a dangerous function because it drops the universe constraints, and nothing indicates this.

tactics/hints.ml
ppedrot: This is indeed a good idea.

ppedrot (changes requested):
Not all the comments are mandatory, but it's probably better to polish this.

kernel/vars.mli
SkySkimmer: >This is a dangerous function because it drops the universe constraints, and nothing indicates this.

Should we just go back to carrying an `int` then.

kernel/mod_subst.mli
SkySkimmer: Changed.

kernel/mod_subst.ml
SkySkimmer: Changed.

kernel/vars.mli
ppedrot: A comment is enough!

SkySkimmer:
Have your comment then.

Gaëtan Gilbert

SkySkimmer:
@ppedrot re ping

ppedrot (approved):

ppedrot:
I am not the one assigned to this PR, and my comments were not that mandatory to be taken into account, but here is your approval.

maximedenes (approved):

maximedenes:
Waiting for a green light from @herbelin.

herbelin:
I find it clearer now. Thanks. Ok to me

PR #8975, opened 2018-11-12
Minor update to micromega.rst

soraros:
**Kind:** documentation.

### DIFF
--- a/doc/sphinx/addendum/micromega.rst
+++ b/doc/sphinx/addendum/micromega.rst
@@ -35,7 +35,7 @@ rationals ``Require Import Lqa`` and reals ``Require Import Lra``.
 
 
 The tactics solve propositional formulas parameterized by atomic
-arithmetic expressions interpreted over a domain :math:`D` ∈ {ℤ, ℚ, ℝ}.
+arithmetic expressions interpreted over a domain :math:`D \in \{\mathbb{Z},\mathbb{Q},\mathbb{R}\}`.
 The syntax of the formulas is the following:
 
  .. productionlist:: `F`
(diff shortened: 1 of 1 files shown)


doc/sphinx/addendum/micromega.rst
```diff
@@ -245,11 +245,11 @@ proof by abstracting monomials by variables.
 As shown, such a goal is solved by ``intro x. psatz Z 2.``. The oracle returns the
 cone expression :math:`2 \times (x-1) + (\mathbf{x-1}) \times (\mathbf{x−1}) + -x^2`
 (polynomial hypotheses are printed in bold). By construction, this expression
-belongs to :math:`\mathit{Cone}({−x^2,x -1})`. Moreover, by running `ring` we
+belongs to :math:`\mathit{Cone}({−x^2,x -1})`. Moreover, by running :tacn:`ring` we
 obtain :math:`-1`. By Theorem :ref:`Psatz <psatz_thm>`, the goal is valid.
 
 .. [#] Support for `nat` and :math:`\mathbb{N}` is obtained by pre-processing the goal with
-  the `zify` tactic.
+  the :tacn:`zify` tactic.
```
Zimmi48: Unfortunately, it looks like `zify` is not documented which made the build of the manual fail with a fatal warning. You may fix this by adding a documentation stub for `zify`. This may be as little as:

```rst
.. tacn:: zify
   :undocumented:
```

Of course, the other solution is not to use `tacn` here.

Zimmi48 (commented):
Sorry in the delay for providing feedback. Note the small tweak required.

doc/sphinx/addendum/micromega.rst
soraros: not competent to write document for zify
just deleted :tacn:

PR #8976, opened 2018-11-12
CoqHammer CI

lukaszcz:
**Kind:** infrastructure.

CI for CoqHammer, done as described in dev/ci/README.md.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -414,6 +414,9 @@ ci-formal-topology:
 ci-geocoq:
   <<: *ci-template-flambda
 
+ci-coqhammer:
+  <<: *ci-template
+
 ci-hott:
   <<: *ci-template
 
--- a/Makefile.ci
+++ b/Makefile.ci
@@ -29,6 +29,7 @@ CI_TARGETS= \
     ci-flocq \
     ci-formal-topology \
     ci-geocoq \
+    ci-coqhammer \
     ci-hott \
     ci-iris-lambda-rust \
     ci-ltac2 \
--- a/dev/ci/ci-basic-overlay.sh
+++ b/dev/ci/ci-basic-overlay.sh
@@ -70,6 +70,13 @@
 : "${HoTT_CI_ARCHIVEURL:=${HoTT_CI_GITURL}/archive}"
 
 ########################################################################
+# CoqHammer
+########################################################################
+: "${coqhammer_CI_REF:=master}"
+: "${coqhammer_CI_GITURL:=https://github.com/lukaszcz/coqhammer}"
+: "${coqhammer_CI_ARCHIVEURL:=${coqhammer_CI_GITURL}/archive}"
+
+########################################################################
 # Ltac2
 ########################################################################
 : "${ltac2_CI_REF:=master}"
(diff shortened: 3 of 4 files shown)


SkySkimmer (changes requested):
Please use the full name `coqhammer` instead of `hammer` everywhere (I prefer the ci-foo script name etc to be the same name as the source repo for foo).

Also please set up CI on the CoqHammer side, have a look at https://github.com/coq-community/docker-coq/wiki and https://github.com/coq-community/paramcoq/blob/master/.travis.yml and ask on gitter if you need help.

lukaszcz:
I changed the name to coqhammer. I won't set up docker/travis anytime soon because I'm not familiar with this, would need to learn it, and I'm busy with other stuff at the moment.

ejgallego (commented):
why the discrepancy `CoqHammer` vs `coqhammer`? IMHO `coqhammer` works better.

lukaszcz:
I never allow commits that fail compilation. I don't realistically expect you to check anything beyond that (this would require the ATPs to be available, and an assumption that the machine is fast enough to execute 'make tests' -- most modern ones should be, but there may be some unexpected quirks for some tests -- whether the hammer fails is really machine-dependent).

As an aside, CoqHammer is designed in such a way that practically the only part that depends on Coq internals is hammer.ml4. The translation proper and everything else is performed on our own representation of Coq terms. So I never expect there to be too much work adapting it to new versions.

My branch policy from now on is that there is the coq8.8 branch which gets updated with new developments, and merged into. The branches e.g. v1.1-coq8.8 are for the releases which are in a stable state.

ejgallego:
> I never allow commits that fail compilation.

It is regardless a good idea to setup automated testing IMO; so far all the projects in the CI that did this way sooner or later committed code that didn't compile :)

> So I never expect there to be too much work adapting it to new versions.

Whether to include the hammer in the Coq's CI is up to you; but be aware that there will be significant API changes including even on term representation, the ml4 format is going away, etc... so YMMV.

lukaszcz:
> > > So I never expect there to be too much work adapting it to new versions.
> > 
> > 
> > Whether to include the hammer in the Coq's CI is up to you; but be aware that there will be significant API changes including even on term representation, the ml4 format is going away, etc... so YMMV.
> 
> Indeed, we do break a lot of stuff in ML APIs these days, but we try to provide the patches to adapt to the changes. So the responsibility of authors is usually limited to being reasonably responsive when we submit such patches.

Sure, my point was only that there is always practically just one file to be modified: hammer.ml4. This file converts internal Coq representation of terms to a "dumbed-down" representation we work on. After this step, the only point when we use Coq API is to invoke the reconstruction tactics (which is pretty simple).

SkySkimmer (approved):

lukaszcz:
OK, I squashed the commits, changed to lowercase, and did chmod.

SkySkimmer:
Coq's CI only runs for pushes which are up to date with master so you need to rebase.

ejgallego:
> OK, I squashed the commits, changed to lowercase, and did chmod.

Thanks! Indeed you need to rebase again as we have a strict policy on not running CI on outdated trees; sorry about that.

ejgallego (approved):

ejgallego (approved):
CI failure is spurious.

PR #8977, opened 2018-11-12
Update .mailmap.

Zimmi48:
**Kind:** infrastructure.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -133,5 +133,8 @@ Théo Zimmermann <theo.zimmermann@univ-paris-diderot.fr> Théo Zimmermann <theo.
 # Anonymous accounts
 
 anonymous < >                                      coq <coq@85f007b7-540e-0410-9357-904b9bb8a0f7>
-anonymous < >                                      (no author) <(no author)@85f007b7-540e-0410-9357-904b9bb8a0f7>
-anonymous < >                                      serpyc <serpyc@85f007b7-540e-0410-9357-904b9bb8a0f7>
+
+# Bot accounts
+
+cvs2svn < >                                        (no author) <(no author)@85f007b7-540e-0410-9357-904b9bb8a0f7>
+serpyc-bot < >                                     serpyc <serpyc@85f007b7-540e-0410-9357-904b9bb8a0f7>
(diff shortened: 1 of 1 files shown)


Zimmi48:
BTW, out of the three "anonymous accounts" one ("coq") has a total of 636 commits over many years (2000-07-19 to 2006-03-05) and is likely to correspond to commits from multiple authors, thus it is a true "anonymous" account while:
- the second one ("no author") is actually responsible for a single commit (9380f25b735834a3c9017eeeb0f8795cc474325b, the first one ever) whose commit message is "New repository initialized by cvs2svn", thus I would be tempted to have mailmap name the committer "csv2svn" instead of a plain "anonymous";
- the third one is "serpyc" and has also a single commit (88909c92cad0044dac83539b2b3d385242ed851e, "Removal of trailing spaces", in 2009). Searching "serpyc" in Google gives results on gforge (apparently this was a member of the Coq project named G. Serpyc) and on coq-contribs (so they did other commits outside of Coq *stricto sensu*), thus I think mailmap should not classify this user as anonymous (we don't do this either for GitHub users using pseudonyms).

Pushing a second commit doing this.

herbelin:
For the anecdote, G. Serpyc is an imaginary name, derived from the name of some server running (iirc) Coq benches at this time of the history - un arbre, si près -. Jean-Marc then added a G. but I don't remember if there was a particular reason for the G. rather than another letter. :smile:

herbelin:
> one ("coq") has a total of 636 commits over many years (2000-07-19 to 2006-03-05) and is likely to correspond to commits from multiple authors, thus it is a true "anonymous" account.

I can check in my archives (on another computer). There was about 3 or 4 users of the "coq" account. The main, significant, one was Jacek Chrząszcz who developed the module system. Iirc, @matafou used it for some while, as well as maybe a couple of other users for ponctual commits.

Matafou:
I remember that my first commits Coq have been actually done by Yves Bertot because I did not have rights. He probably used the coq account.

Zimmi48:
> For the anecdote, G. Serpyc is an imaginary name, derived from the name of some server running (iirc) Coq benches at this time of the history - un arbre, si près -. Jean-Marc then added a G. but I don't remember if there was a particular reason for the G. rather than another letter.

Thanks for the info. I should probably drop the second commit then, or at least update the commit message.

SkySkimmer (approved):
mailmap is unowned so I'm going to take this and merge unless someone complains within a couple days.

PR #8978, opened 2018-11-12
nix helpers for debugging external projects from CI

vbgl:
Use-case: a change in Coq breaks a third-party development (in particular, one of those that are tested on CI); how to fix that development. There are at least to issues that this PR aims to solve:

 1. What are the dependencies of that development?
 2. What is its the build-system?

This PR attempts to give a single interface to most developments tracked in CI, such that:

  1. Running `dev/ci/nix/shell foo` sets up an environment with all the required dependencies to build/edit development `foo`.
  2. In that environment, variables `$configure`, `$make`, and `$clean` can be `eval`-ed so that we can forget what build-system is used, what options have to be given to the configure script, what are the relevant targets of the `makefile`s, etc.

NB: this is not meant to work on Coq itself.

Caveat: the design is not perfect, but I find this rather usable and ready for sharing.


dev/ci/nix/unicoq.nix
```diff
@@ -0,0 +1,14 @@
+{ stdenv, fetchzip, coq }:
+
+stdenv.mkDerivation {
+  name = "coq${coq.coq-version}-unicoq-0.0-git";
+  src = fetchzip {
+    url = "https://github.com/vbgl/unicoq/archive/8b33e37700e92bfd404bf8bf9fe03f1be8928d97.tar.gz";
```
SkySkimmer: Why's this one using a static commit?

dev/ci/nix/shell
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env sh
```
SkySkimmer: This file should be set to belong to the nix team in codeowners.

ejgallego:
I am not happy to see this information duplicated twice, already having it for the CI system was bad, now having it twice seems horrible.

IMVHO projects upstream should get the `nix` file and then Coq's script should compose with it.

This is the plan for Dune, as CI developments get dunerized the CI system will have to know nothing about their build setup other than the repos name.

vbgl:
I totally agree that if we can have all third-party developments agree on a common build-system and a common format to describe their dependencies, that would be much simpler. Currently, I find it rather inconvenient to work on such a diverse set of settings, motivating this work. If you believe I should keep it for myself, that’s fine.

ejgallego:
What's the problem with submitting these `nix` files upstream?

ejgallego:
My main fear is that consistency will be hard to maintain. I have a similar script [not yet ready to submit] but I reuse the info already in the CI scripts.

ejgallego:
> If you believe I should keep it for myself, that’s fine.

I'm split, that's very useful, on the other I am wondering how the maintenance of this is going to work.

ejgallego (approved):
I am not too bothered by this given that it is self-contained; if it is useful for NiX people IMHO it can go in.

WDYT @SkySkimmer ?

SkySkimmer:
About the same.

dev/ci/nix/unicoq.nix
vbgl: I don’t know how to build the current version of `Unicoq`. See https://github.com/unicoq/unicoq/pull/17 for details.

Zimmi48:
An additional issue with forcing Nix files into external projects is that people won't know how to maintain them. I have done this in (some) coq-community projects but this was part of setting up CI (now it seems most projects will rather use Docker) and I am there to maintain them (and BTW they are far from polished).

@vbgl I would have appreciated a README file documenting this stuff.

ejgallego:
@Zimmi48 we are using Docker as it got ready with decent documentation, but I'd be very cool if we could provide NiX instructions too; in fact Nix is strictly better wrt to up to date builds so I'd prefer it if possible.

Do we have the docs on Travis + Nix somewhere ? I was expecting to find it on the manifesto website but I didn't, I guess this is https://github.com/coq-community/manifesto/issues/3 right?

vbgl:
Théo, you are right. Let me add a readme.

Emilio, there is a bit of documentation regarding Travis on the [NixOS wiki](https://nixos.wiki/wiki/Nix_on_Travis).

Zimmi48:
@ejgallego Yep, that's still WIP. The docs are not there yet because the final format for this setup wasn't clear yet. Thanks to @vbgl's help, it progressed a lot. The last hurdle is the integration with Cachix (to ensure e.g. that corn is built with the latest version of math-classes but this version is only built once). It's been a long time since I promised a specific issue tracking this. I'll make sure to open it next week.

PR #8981, opened 2018-11-12
[obligations] Restrict hack catching exceptions to tactics

maximedenes:
I hit a case where the error is silently dropped, leading to an
inconsistent state.

This logic seems to be a conscious choice, introduce for example by https://github.com/coq/coq/commit/07bd9c28f37b9ec390e5e3dcacdfd8183056be88 but I don't understand it. So I'm not sure if removing it is the right fix.

### DIFF
--- a/vernac/obligations.ml
+++ b/vernac/obligations.ml
@@ -827,17 +827,24 @@ let rec string_of_list sep f = function
 
 (* Solve an obligation using tactics, return the corresponding proof term *)
 
-let solve_by_tac name evi t poly ctx =
-  let id = name in
-  (* spiwack: the status is dropped. *)
-  let (entry,_,ctx') = Pfedit.build_constant_by_tactic
-    id ~goal_kind:(goal_kind poly) ctx evi.evar_hyps evi.evar_concl (Tacticals.New.tclCOMPLETE t) in
-  let env = Global.env () in
-  let entry = Safe_typing.inline_private_constants_in_definition_entry env entry in
-  let body, () = Future.force entry.const_entry_body in
-  let ctx' = Evd.merge_context_set ~sideff:true Evd.univ_rigid (Evd.from_ctx ctx') (snd body) in
-  Inductiveops.control_only_guard env ctx' (EConstr.of_constr (fst body));
-  (fst body), entry.const_entry_type, Evd.evar_universe_context ctx'
+let solve_by_tac loc name evi t poly ctx =
+  try
+    let id = name in
+    (* spiwack: the status is dropped. *)
+    let (entry,_,ctx') = Pfedit.build_constant_by_tactic
+        id ~goal_kind:(goal_kind poly) ctx evi.evar_hyps evi.evar_concl (Tacticals.New.tclCOMPLETE t) in
+    let env = Global.env () in
+    let entry = Safe_typing.inline_private_constants_in_definition_entry env entry in
+    let body, () = Future.force entry.const_entry_body in
+    let ctx' = Evd.merge_context_set ~sideff:true Evd.univ_rigid (Evd.from_ctx ctx') (snd body) in
+    Inductiveops.control_only_guard env ctx' (EConstr.of_constr (fst body));
+    Some ((fst body), entry.const_entry_type, Evd.evar_universe_context ctx')
+  with e when CErrors.noncritical e ->
+    let (e, _) = CErrors.push e in
+    match e with
+    | Refiner.FailError (_, s) ->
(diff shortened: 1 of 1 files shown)


ejgallego:
@maximedenes I tried to tweak this exact same code a few weeks ago; the problem we are facing here is one of callback with possibly the wrong printing (that is to say, notations, etc...) state; I didn't manage to understand the code enough as to move forward as the flow in this part is a very tricky jump-game from the stm to proof_terminators.

maximedenes:
I restricted the patch to what I needed. There are other suspicious pieces of error logic in obligations, though.

vernac/obligations.ml
```diff
@@ -827,17 +827,24 @@ let rec string_of_list sep f = function
...
+    let body, () = Future.force entry.const_entry_body in
+    let ctx' = Evd.merge_context_set ~sideff:true Evd.univ_rigid (Evd.from_ctx ctx') (snd body) in
+    Inductiveops.control_only_guard env ctx' (EConstr.of_constr (fst body));
+    Some ((fst body), entry.const_entry_type, Evd.evar_universe_context ctx')
+  with e when CErrors.noncritical e ->
+    let (e, _) = CErrors.push e in
+    match e with
+    | Refiner.FailError (_, s) ->
+      user_err ?loc ~hdr:"solve_obligation" (Lazy.force s)
+    | e -> None (* FIXME really ? *)
```
ejgallego: I guess the refactoring to option is fine, however I think you should strengthen the semantics here [CI permitting] and either:

- turn the `tclCOMPLETE` into a real exception.
- or even better remote `tclCOMPLETE` at all, it should be close_proof the one in charge of owning this logic. Let me have indeed a go at it.

ejgallego (changes requested):
Can you tell mor e why this change helps and what "case" are you getting this error?

I am not sure I can understand how your change does make a semantic difference.

maximedenes:
> I am not sure I can understand how your change does make a semantic difference.

Before this patch, the code incorrectly catches any exception raised by `declare_obligation` and treat it as if it had been raised by the tactic execution. This patch solves it by restricting the scope of the `try` block. I think it is a sound change. One can go further in the clean-up, but I'd like to get this in.

ejgallego:
> One can go further in the clean-up, but I'd like to get this in.

Ok, you should amend the PR title / description, I didn't realize the commit had the right message.

I can assign this PR if you want.

Out of curiosity, what kind of exceptions does `declare_obligation` raise?

ejgallego (approved):

maximedenes:
> Ok, you should amend the PR title / description, I didn't realize the commit had the right message.

Indeed, I forgot to update it when I changed the content.

> 
> I can assign this PR if you want.
> 
> Out of curiosity, what kind of exceptions does `declare_obligation` raise?

Well, any call to a user error there will show the problem. The error will not appear and put the obligations system in an inconsistent state.

In my case, `declare_obligation` was emitting a warning when it was declaring a hint database. With `-w +all`, the warning became an error, hence the exception...

maximedenes:
For the record, I think we should scan all unrestricted exceptions catching in the codebase. After loosing an hour on this one, I lost another hour because of https://github.com/coq/coq/blob/0b816e4df10d961cce082894f5e4087dc1c95f01/pretyping/pretyping.ml#L268

maximedenes:
Are the failures real?

ejgallego:
> In my case, `declare_obligation` was emitting a warning when it was declaring a hint database. With `-w +all`, the warning became an error, hence the exception...

Ah, I finally understand; this is what I was missing.

ejgallego:
> Are the failures real?

Spurious TTBOMK

maximedenes:
@mattam82 can you be the assignee? Thanks!

gares (approved):
but I've no idea what the code was trying to do

mattam82:
Fine with me. I think you analyzed the problem right, it’s about printing the error in the right global state, but indeed the catchall is too strong. I guess I didn’t know which exceptions should be fatal for Solve Obligations and which shouldn’t be. In retrospect, I think only the ones of complete / control_only_guard should be catched. Or we could use give_up and force the user to explicitly add an `|| give_up` tactic to signal that the tactic given might not solve all obligations.

ejgallego:
@mattam82 see also #8982 where indeed a stronger restriction is implemented for Solve Obligations.

maximedenes:
> @mattam82 can you be the assignee? Thanks!

ping

ppedrot:
Following the discussion in #8982, let's close this PR.

PR #8781, opened 2018-10-19
Remove primproj <-> constant dependency in Heads

SkySkimmer:
And following this remove a Projection.constant in recordops when declaring the association between primitive projections and constants.

Testing Heads is made difficult by it being used behind complicated old unification and coercion call stacks, but logically this should work and the test suite agrees with me.

PR #8982, opened 2018-11-13
[proof] Provide better control of "open proofs" exceptions.

ejgallego:
This is inspired and an alternative to #8981. We consolidate the "open
proof" exception, allowing clients to explicitly capture it and
removing some ugly duplicated code in the way.

The `Solve Obligation tac` semantics are then tweaked as to removed
the wide-scope "catch-all" and indeed will now relay errors in `tac`
as it will only absorb tactics that don't error but fail to close the
goal such as `auto`.

In particular, we also remove an unnecessary `tclCOMPLETE` call to
code that will actually call `close_proof`. In this case, it is better
to delegate error management to the core function.

Some error messages have changed [as we consolidate two error paths]
so this PR may require adjustment in that area.

### DIFF
--- a/proofs/logic.ml
+++ b/proofs/logic.ml
@@ -62,6 +62,8 @@ let is_unification_error = function
 
 let catchable_exception = function
   | CErrors.UserError _ | TypeError _
+  | Proof.OpenProof _
+  (* abstract will call close_proof inside a tactic *)
   | Notation.NumeralNotationError _
   | RefinerError _ | Indrec.RecursionSchemeError _
   | Nametab.GlobalizationError _
--- a/proofs/proof.ml
+++ b/proofs/proof.ml
@@ -335,28 +335,42 @@ let dependent_start goals =
   let number_of_goals = List.length (Proofview.initial_goals pr.entry) in
   _focus end_of_stack (Obj.repr ()) 1 number_of_goals pr
 
-exception UnfinishedProof
-exception HasShelvedGoals
-exception HasGivenUpGoals
-exception HasUnresolvedEvar
-let _ = CErrors.register_handler begin function
-  | UnfinishedProof -> CErrors.user_err Pp.(str "Some goals have not been solved.")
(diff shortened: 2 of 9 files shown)


ejgallego:
This broke omega, funny :)

ejgallego:
Oh, so indeed some developments do:
```
  Solve Obligations with coq_omega.
```
So the new, stricter semantics have that fail due to `omega`. Need to think about this one.

Another source of problems is `all: try abstract` as the `close_proof` error won't be in the monad; I have reverted this change.

ejgallego:
I guess we could introduce a warning for the case of tactics erroring in `Solve Obligations`, and then remove this behavior in the future and require tactics not to fail. WDYT @mattam82 @maximedenes ?

ejgallego:
Umm, very very interesting failure in VST; for sure there is something going on here with exn masking.

ejgallego:
Indeed this looks like a `tclCOMPLETE` bug, the implementation doesn't take into account shelved goals so it is not the case that if `tclCOMPLETE` succeeds a goal can be saved.

ppedrot (approved):
I have no specific opinion, this is higher-order spaghetti code, so I'll trust your PR. Nonetheless, you claim that this supersedes #8981, should we close that one in the meantime?

ppedrot:
OK, @maximedenes do you favour this PR over #8781?

maximedenes:
> OK, @maximedenes do you favour this PR over #8781?

As long as exceptions are no longer randomly caught, especially coming from `declare_instance`, I'm ok.

ejgallego:
> As long as exceptions are no longer randomly caught, especially coming from `declare_instance`, I'm ok.

In that sense this PR should behave exactly as yours.

ppedrot:
Last round of comments before I merge.

ejgallego:
Merci !

ejgallego:
@maximedenes please complain if this does indeed not help with your problem [the exception is handled in the same exact way as we discussed, but just in case]

PR #8984, opened 2018-11-13
Declare initial hint databases in prelude

maximedenes:
Previously, they were hard-wired in the ML code.

This is a preliminary step towards making database names qualified.

### DIFF
--- a/theories/Init/Tactics.v
+++ b/theories/Init/Tactics.v
@@ -332,3 +332,7 @@ Tactic Notation "assert_succeeds" tactic3(tac) :=
   assert_succeeds tac.
 Tactic Notation "assert_fails" tactic3(tac) :=
   assert_fails tac.
+
+Create HintDb rewrite discriminated.
+Hint Variables Opaque : rewrite.
+Create HintDb typeclass_instances discriminated.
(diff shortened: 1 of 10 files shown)


ppedrot (approved):

vbgl:
Is someone working on the overlays for this PR?

Zimmi48:
> Is someone working on the overlays for this PR?

It seems wrong that this code change requires overlay. It would seem that we should fix the code so that we can do this declaration without breaking anything (if that's not possible as of today).

theories/Init/Tactics.v
```diff
@@ -328,3 +328,7 @@ Tactic Notation "assert_succeeds" tactic3(tac) :=
   assert_succeeds tac.
 Tactic Notation "assert_fails" tactic3(tac) :=
   assert_fails tac.
+
+Create HintDb rewrite discriminated.
```
JasonGross: Maybe these should go at the top of the file rather than the bottom?  (So that if we ever get static checking of existence of databases for Ltac and also add a tactic that does rewrite or typeclasses, we won't have to reorder; this is a weak reason to put it at the top, and I see no reason to put them at the bottom.)

vbgl:
Regarding overlays, one failure is related to the move of the `typeclasses_db` OCaml value from the `Hints` module to the `Class_tactics` module. The two other failures are in developments that have their own prelude: these need to be updated to add the database declarations.

I’m not sure I understand your remark, Théo.

Zimmi48:
OK sorry @vbgl, I had misunderstood the need for overlays following a discussion related to this PR with Maxime.

vbgl:
No problem.

vbgl:
Overlays are OK, CI is green.

Ping assignee @ppedrot

PR #8986, opened 2018-11-13
Put -indices-matter in typing_flags

SkySkimmer:


ejgallego (approved):
Very good.

ppedrot (approved):
Makes sense. We should probably check consistency of separate modules though, as nothing prevents you from mixing both.

maximedenes:
> We should probably check consistency of separate modules though, as nothing prevents you from mixing both.

Is it a problem to mix both? If yes, I'm not so sure about this change, I thought it would have to follow `engagement`. If no, then this change is very good.

mattam82:
In principle you could mix but that might be a bit strange, what are use
cases for it? I guess HoTT users would want to prevent mixing both.
Le mar. 20 nov. 2018 à 12:06, Maxime Dénès <notifications@github.com> a
écrit :

maximedenes:
I see it the other way around: preventing mixing requires more work. We'd need a flavor of typing flags where consistency across files is guaranteed generically, like for `engagement`. If we don't care about mixing, this can get in as-is.

mattam82:
Right, it's already better than a `ref`. Currently we allow mixing, and I don't think that can be a source of inconsistency (Inductives defined with -indices-matter can be accepted with the same universes in non-indices-matter mode), so this can get in as is.

ppedrot:
@mattam82 you should probably merge this one.

PR #8987, opened 2018-11-13
Deprecate hint declaration/removal with no specified database

maximedenes:
Previously, hints added without a specified database where implicitly
put in the "core" database, which was discouraged by the user manual
(because of the lack of modularity of this approach).

This is another step towards making the migration to qualified database names easier.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,6 +47,10 @@ Tactics
   (e.g. `?[n]` or `?n` in terms - not in patterns) are now interpreted
   the same way as other variable names occurring in Ltac functions.
 
+- Hint declaration and removal should now specify a database (e.g. `Hint Resolve
+  foo : database`). When the database name is omitted, the hint is added to the
+  core database (as previously), but a deprecation warning is emitted.
+
 Vernacular commands
 
 - `Combined Scheme` can now work when inductive schemes are generated in sort
--- a/plugins/btauto/Algebra.v
+++ b/plugins/btauto/Algebra.v
@@ -10,7 +10,7 @@ end.
 
 Arguments decide P /H.
 
-Hint Extern 5 => progress bool.
+Hint Extern 5 => progress bool : core.
 
 Ltac define t x H :=
 set (x := t) in *; assert (H : t = x) by reflexivity; clearbody x.
@@ -147,7 +147,7 @@ Qed.
 
 (** * The core reflexive part. *)
 
-Hint Constructors valid.
+Hint Constructors valid : core.
 
 Fixpoint beq_poly pl pr :=
 match pl with
@@ -315,7 +315,7 @@ Section Validity.
 
 (* Decision procedure of validity *)
 
-Hint Constructors valid linear.
+Hint Constructors valid linear : core.
 
 Lemma valid_le_compat : forall k l p, valid k p -> (k <= l)%positive -> valid l p.
 Proof.
@@ -425,10 +425,10 @@ match goal with
 | [ |- (?z < Pos.max ?x ?y)%positive ] =>
   apply Pos.max_case_strong; intros; lia
 | _ => zify; omega
-end.
-Hint Resolve Pos.le_max_r Pos.le_max_l.
+end : core.
+Hint Resolve Pos.le_max_r Pos.le_max_l : core.
 
-Hint Constructors valid linear.
+Hint Constructors valid linear : core.
 
 (* Compatibility of validity w.r.t algebraic operations *)
 
--- a/plugins/btauto/Reflect.v
+++ b/plugins/btauto/Reflect.v
@@ -77,10 +77,10 @@ intros var f; induction f; simpl poly_of_formula; simpl formula_eval; auto.
   end.
 Qed.
 
-Hint Extern 5 => change 0 with (min 0 0).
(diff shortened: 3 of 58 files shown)


gares (approved):

ppedrot (approved):
I support this change a lot.

mattam82 (approved):
LGTM. Maybe you could also document the sed incantation that allows to port scripts somewhere?

ppedrot:
CI failure is spurious. This is both a simple and seemingly uncontroversial PR, so I am planning to merge quickly, but I'll wait for @herbelin comments nonetheless.

Zimmi48:
Two remarks after the fact:
- I agree the CHANGES entry should have hinted the sed command to use to fix script.
- The reference manual should have been updated as well to say that not providing the hint database is deprecated.

Zimmi48:
@maximedenes ping. I think at least updating the refman is required here.

charguer:
Hi.

The current implementation appears to be too restrictive.

I am using local hints everywhere, either in the form
```
Section Foo. 
Hint Resolve mylemma.
Lemma bla...
End Foo.
```
or (rarely, those I could get rid of) in the form:
```
Local Hint Resolve mylemma.
```

These patterns do not pollute the "core" database in non-modular way, because everything is local.
I would thus expect these forms to remain supported, in their current syntax.

maximedenes:
Hi @charguer! I'm not even convinced a `core` database should exist in the first place. Can you say more about the practical problem you are trying to solve?

charguer:
Ok, so concretely I have a bunch of 10 lemmas to prove, for which I need to provide a bunch of "hint resolve" and "hint extern". (If I don't use hints, the proofs would be horrible.) What I typically do is to open a section, load the hints, into the "core" database, but only within the scope of this section.

Providing for each hint a dummy database name, e.g. "mysection_hintdb", would be a lot of syntactic overhead. I would very much like to stick to the current syntax, that is, not need to come up with dummy hintdb names (btw, I'd also like to have anonymous sections). I would also not like to write "Local", because it's already implicit from being inside a section, so it seems redundant.

In summary, what I need is the ability to configure "auto" over a local scope to be extended with a bunch of hints, and to do so without lots of annotations (e.g. "Local" or ": mysection_hintdb"), and without the horribly heavy "auto with" or "auto using" syntax.

Zimmi48:
An issue should be opened to discuss this concrete use case and how it can be satisfied.

charguer:
@Zimmi48 Rebinding auto does not work, because there might be other tactics that I use that invoke "auto with foo" indirectly. These tactics would not be updated appropriately.

Not to mention that the syntax to achieve this is horrible.

maximedenes:
Thanks for the details, @charguer. As @Zimmi48 said, I believe what you need is indeed locally-bound implicit databases, rather than a hardwired `core` database like `core` is. We should remember to implement that before removing the deprecated syntax, and the best way to ensure it is to open an issue.

charguer:
Ok let me open up the issue and copy-paste some of this discussion, then.

In any case, the message accompanying the "deprecated use of the core database" should be very clear that the user should NOT modify his code, but instead wait for the next feature.

Even better, still, would be to postpone the depreciation of the use of core until an alternative solution does exist. 

Why is it urgent to raise the "deprecated use of core" warnings? Especially if we're going to provide a better solution in the future, which might very well use the same syntax as currently for hints local to a section. In that perspective, it seems better to wait, or to restrict the "deprecated" warning to hints that are not local to a section.

herbelin:
I experimented using a `sh` script to automatically add a database hint to all `Hint` and `auto`. This is the script (for BSD sed):
```
set HINTDB=foo
sed -i.bak -e "s/\(Hint [^:\.]*\)\./\1 : $HINTDB./g" \
-e "s/\([^a-zA-Z0-9_]auto *[0-9]* *with[ a-zA-Z0-9_]*\)/\1 $HINTDB/g" \
-e "s/\([^a-zA-Z0-9_]eauto  *[0-9]* *with[ a-zA-Z0-9_]*\)/\1 $HINTDB/g" \
-e "s/^\(auto  *[0-9]* *with[ a-zA-Z0-9_]*\)/\1 $HINTDB/g" \
-e "s/^\(eauto  *[0-9]* *with[ a-zA-Z0-9_]*\)/\1 $HINTDB/g" \
-e "s/\([^a-zA-Z0-9_]auto\)\( *[]).;|]\)/\1 with $HINTDB\2/g" \
-e "s/\([^a-zA-Z0-9_]eauto\)\( *[]).;|]\)/\1 with $HINTDB\2/g" \
-e "s/^auto *\([]).;|]\)/auto with $HINTDB\1/g" \
-e "s/^eauto *\([]).;|]\)/eauto with $HINTDB\1/g" *.v
```
Notes:
- With GNU sed, it should be `sed -i` instead of `sed -i.bak` (not tested).
- It requires all files to be ended with a newline (this can be enforced e.g. by calling `sed -i.bak -e "s/\(.\)/\1/g" *.v`).
- It does not support `auto` as the last word of a line.
- It does not support `auto using`, nor `auto NN` with `NN` a number and no `with`.
- It does not support `Hint` spanning over several lines.

I publish it because it is better than nothing but providing a diff from coqtop itself would be both either to write and more robust...

PR #8988, opened 2018-11-13
Towards unifying parsing/printing for universe instances and Type's argument

herbelin:
**Kind:** cleanup

Synchronous overlay: LPCIC/coq-elpi#63.

This is a mini-proposal to replace the encodings used to internally represent the universe level of a `Type` at the `constr_expr`/`glob_constr` levels by the syntax already used to internally represent instances of polymorphic definitions.

We consistently use:
- `UUnknown`: to mean a rigid anonymous universe
  (written `Type` in instances and `Type` as a sort; was called like this in instances and formerly encoded as `[]` in `Type`'s argument)
- `UAnonymous`: to mean a flexible anonymous universe
  (written `_` in instances and `Type@{_}` as a sort; was called like this in instances and formerly encoded as `[None]` in `Type`'s argument]
- `UNamed`: to mean a named universe or universe expression
  (written `id` or `qualid` in instances, and, as a sort, `Type@{id}` or `Type@{qualid}` or more generally `Type@{id+n}`, `Type@{qualid+n}`, `Type@{max(...)}`)

There is a little change of syntax: `_` in a `max` expression (e.g. `Type@{max(_,id+1)}` is not anymore allowed. But it was trivially satisfiable by unifying the flexible universe with a neighbor of the list and the syntax is not documented anyway.

Thanks to @SkySkimmer for helping.

Added: There is a little change of semantics: in `Notation x := Type. Check x@{Type}.`, a rigid variable is considered while it was formerly flexible. It is still flexible with `Check x@{_}.`.

### DIFF
--- a/vernac/comInductive.ml
+++ b/vernac/comInductive.ml
@@ -121,7 +121,7 @@ let mk_mltype_data sigma env assums arity indname =
 let rec check_anonymous_type ind =
   let open Glob_term in
     match DAst.get ind with
-    | GSort (GType UUnknown) -> true
+    | GSort (UAnonymous {rigid=true}) -> true
     | GProd ( _, _, _, e)
     | GLetIn (_, _, _, e)
     | GLambda (_, _, _, e)
@@ -495,7 +495,7 @@ let extract_params indl =
 let extract_inductive indl =
   List.map (fun ({CAst.v=indname},_,ar,lc) -> {
     ind_name = indname;
-    ind_arity = Option.cata (fun x -> x) (CAst.make @@ CSort (Glob_term.GType Glob_term.UUnknown)) ar;
+    ind_arity = Option.cata (fun x -> x) (CAst.make @@ CSort (Glob_term.UAnonymous {rigid=true})) ar;
     ind_lc = List.map (fun (_,({CAst.v=id},t)) -> (id,t)) lc
   }) indl
 
--- a/vernac/comInductive.ml
+++ b/vernac/comInductive.ml
@@ -121,7 +121,7 @@ let mk_mltype_data sigma env assums arity indname =
 let rec check_anonymous_type ind =
   let open Glob_term in
     match DAst.get ind with
-    | GSort (GType []) -> true
+    | GSort (GType UUnknown) -> true
     | GProd ( _, _, _, e)
     | GLetIn (_, _, _, e)
     | GLambda (_, _, _, e)
@@ -495,7 +495,7 @@ let extract_params indl =
 let extract_inductive indl =
   List.map (fun ({CAst.v=indname},_,ar,lc) -> {
     ind_name = indname;
-    ind_arity = Option.cata (fun x -> x) (CAst.make @@ CSort (Glob_term.GType [])) ar;
+    ind_arity = Option.cata (fun x -> x) (CAst.make @@ CSort (Glob_term.GType Glob_term.UUnknown)) ar;
     ind_lc = List.map (fun (_,({CAst.v=id},t)) -> (id,t)) lc
   }) indl
 
--- a/dev/ci/user-overlays/08988-herbelin-master+mini-uniformizaton-parsing-printing-univ-level-sort.sh
+++ b/dev/ci/user-overlays/08988-herbelin-master+mini-uniformizaton-parsing-printing-univ-level-sort.sh
@@ -1,0 +1,6 @@
+if [ "$CI_PULL_REQUEST" = "8988" ] || [ "$CI_BRANCH" = "master+mini-uniformizaton-parsing-printing-univ-level-sort" ]; then
+
+    elpi_CI_REF=coq-master+adapt-coq8988-type-internal-syntax
+    elpi_CI_GITURL=https://github.com/herbelin/coq-elpi
+
+fi
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -758,8 +758,9 @@ let extended_glob_local_binder_of_decl ?loc u = DAst.make ?loc (extended_glob_lo
 
 let extern_glob_sort = function
   (* In case we print a glob_constr w/o having passed through detyping *)
-  | GType _ when not !print_universes -> GType UUnknown
-  | u -> u
+  | UNamed [(GSProp,0) | (GProp,0) | (GSet,0)] as u -> u
+  | UNamed _ when not !print_universes -> UAnonymous {rigid=true}
+  | UNamed _ | UAnonymous _ as u -> u
 
 let extern_universes = function
   | Some _ as l when !print_universes -> l
@@ -1310,10 +1311,10 @@ let rec glob_of_pat avoid env sigma pat = DAst.make @@ match pat with
           Array.map (fun (bl,_,_) -> bl) v,
           Array.map (fun (_,_,ty) -> ty) v,
           Array.map (fun (_,bd,_) -> bd) v)
-  | PSort Sorts.InSProp -> GSort GSProp
-  | PSort Sorts.InProp -> GSort GProp
-  | PSort Sorts.InSet -> GSort GSet
-  | PSort Sorts.InType -> GSort (GType UUnknown)
+  | PSort Sorts.InSProp -> GSort (UNamed [GSProp,0])
+  | PSort Sorts.InProp -> GSort (UNamed [GProp,0])
+  | PSort Sorts.InSet -> GSort (UNamed [GSet,0])
+  | PSort Sorts.InType -> GSort (UAnonymous {rigid=true})
   | PInt i -> GInt i
 
 let extern_constr_pattern env sigma pat =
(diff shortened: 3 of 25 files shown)


SkySkimmer:
Interesting, I wonder if we should change it up a bit more
ie we have
~~~ocaml
type 'a universe_kind = UUnknown | UAnon | UNamed of 'a
type 'a glob_sort_gen = GProp | GSet | GType of 'a
type 'a expr = ('a * int) list (* inlined in real coq code for now *)
~~~
then currently we have 
~~~ocaml
type glob_level = qualid universe_kind glob_sort_gen
type glob_sort = (qualid * int) option list  glob_sort_gen (* the option (for _) makes it incompatible with expr *)

let just_Type = GType []
~~~
and the current proposal is
~~~ocaml
type glob_level = qualid universe_kind glob_sort_gen (* unchanged *)
type glob_sort = qualid expr universe_kind glob_sort_gen

let just_Type = GType UUnknown
~~~
but instead we could have 
~~~ocaml
type glob_sort = qualid glob_sort_gen expr universe_kind

let just_Type = UUnknown
~~~
This would allow representing `max(Set,i)` (as `UNamed [(GSet, 0); (GType i, 0))]`). (For uniformity we would also do `type glob_level = qualid glob_sort_gen universe_kind`, which is isomorphic to the current one.)

(NB: in both cases using `expr` the empty list should not be used)

SkySkimmer:
Also I think we should unify `UUnknown` and `UAnonymous` into one constructor `UAnonymous of { rigid : bool }` as it's not clear (at least for me) from the name which is which.

herbelin:
Both proposal sound good. Then changing the name of `UNamed` to `UExpr` or something like that?

SkySkimmer:
sure

Gaëtan Gilbert

herbelin:
Shall I do it now or shall I wait for @mattam82's opinion?

SkySkimmer:
Go ahead I think.

Gaëtan Gilbert

gares (approved):
changes to ssrcommon OK

herbelin:
I pushed an extra commit to experiment your proposal (I finally kept the name `UNamed`).

This simplifies some parts, such as constraints management, since constraints are not supposed to mention anonymous universes.

On the other side, this makes talking about `Prop` and `Set` a bit less natural since, in situations of parsing/printing a sort, or of converting to `constr`, it is encoded as `UNamed [GProp,0]` or `UNamed [GSet,0]`. Conversion to `Prop` and `Set` actually raises the question of the redundancy between the sort representations `Prop`/`Type [Prop,0]` and `Set`/`Type [Set,0]` in the kernel.

herbelin:
Besides the need for overlay, what does this PR inspire, especially @mattam82? A good direction or not?

mattam82 (approved):
It looks good to me, the representation you propose is more uniform and I like the changes to the code. 

About your second point, indeed we could use a view to move from the old representation (Prop/Set/Type) to the new one without loosing much in terms of ease of pattern-matching on sorts.

ppedrot:
Just to mention it, there is a desire to be able to extend Coq with alternative universe hierarchies, e.g. homotopy nTypes, Kerller-Lasson-style of parametric universes, effectful universes, etc. It'd be nice to take this kind of potential future extension into account in the design of universe syntax and handling.

herbelin:
@ppedrot: Certainly.

I don't have a clear idea of what each extension would require technically. What are the ideas?

Also, when you're taking about homotopy nTypes, are you talking about reinterpreting current `Type(n)` as `hSet(n)` and add a new, say, `Univ(n)`, for full-fledged Martin-Löf's universes, so that we can at the same time have univalence stated on `Univ(n)` and non-indices-matter singleton elimination remaining valid on `Type(n)`-as-`hSet(n)`? This would allow to get rid of the current hack of `-indices-matter` which changes non-indices-matter singleton elimination to indices-matter singleton elimination so that `Type(n)` can be interpreted as a full-fledged Martin-Löf's universe but looses the `eq` we know in `Prop`, with the drawback of needing two libraries with different definitions of `eq` while we could have instead a single library with two definitions of `eq`, one over `Type(n)`=`hSet(n)` in `Prop`, and one over `Univ(n)` in `Univ(n)` (hoping that I'm clear enough).

mattam82:
@herbelin I'm confused: we shouldn't mix homotopy level and universe levels in general. The spheres can live in Set for example. About Lasson's et al idea, I think it is closer to how we could extend the hierarchy by splitting Type into Set (for computational objects) and Types (for types, i.e. arities) and simplify extraction. I think the current PR is compatible with this idea. The idea of representing Prop and Set by using their representation in [Type u] would go against it though, I think.

SkySkimmer (approved):
I think this is ready, will merge soon.

herbelin:
> The spheres can live in `Set` for example

Yes, spheres can live in `Set` seen as a universe without other constraints than being a superset of `Prop` closed under products and sums (i.e. the current `Set` if I did not forget any constraint). What I'm proposing is to re-interpret the current names `Set` and `Type` as `hSet`, in which case spheres could not live any more in `Set` or `Type` of course. But they could live in`Univ` (which would be the new name for a `Type` with unrestricted homotopy structure).

My point was that for having `t = u :> A` in `Prop` for `A:Type` in a way compatible with both univalence and proof-irrelevance, it should be enough to restrict the interpretation of the name `Type(n)` to `hSet(n)` (*). By doing so, `bool = bool :> Type(0)` would not be in `Prop` because `Type(0)` = `hSet(0)` is not itself a `hSet`, so we would loose some expressivity but we would not loose the expressivity that we need for the compatibility of library: the libraries mostly requires `(t = u :> A) : Prop` for type variables `A` which are mostly expected to be instantiated  in practice by a type syntactically inferable to be `hSet`.

(*) By `hSet`, I mean in practice a syntactical underapproximation ensuring that for `A` to be in `hSet`, it has to satisfy IUP.

#### Further questions

The interplay between size and homotopy levels is indeed important. For instance, the discussion makes me wondering:
- Q1: Is it consistent to put in `U(0)` the equality of elements of `A : U(i)`, for arbitrary `i`? I.e., conversely, is it possible to build equality proofs which observationally have a large size? For instance, even with a HIT of the form
  ```
  Inductive I : Type(42) :=
  | C : Type(41) -> I
  | E : forall A : Type(41), C A = C A
  ```
  could we build a paradox from stating `(t = u :> I) : Type(0)`?
- If this is consistent, that would mean that the purpose of `-indices-matter` is not about size but about homotopy levels.
- This would also reduce the distance to answer to the question of whether univalence is consistent with axiom-free Coq. Since, in the absence of proof-irrelevance, `Prop` has an interpretation as impredicative `Set`, if it were clear that impredicativity is compatible with non-trivial homotopy level, Q1 would allow us to interpret `t = u :> A` as an element of an impredicative variant of `Type(0)`.

#### Keller-Lasson

> About Lasson's et al idea, I think it is closer to how we could extend the hierarchy by splitting Type into Set (for computational objects) and Types (for types, i.e. arities) and simplify extraction.

You mean a sort to mark what is extracted without having to expect this sort to be proof-irrelevant? Sounds good.

Would it be two hierarchies or all types in the same hiearchy, with a computational-vs-arity flag on each type?

> The idea of representing Prop and Set by using their representation in [Type u] would go against it though, I think.

Note that there are parts of the code which maps `Type[Prop,0]` to `Prop` already (e.g. `sort_of_univ`).

---

Sorry I was long. Just pick whatever you want from it.

SkySkimmer:
Looks like an overlay is needed for elpi.

SkySkimmer:
There's a conflict due to cff3c5a7148afc722852bd01658fe49ffec1d967

plugins/cc/ccalgo.ml
```diff
@@ -134,8 +134,8 @@ type cinfo=
      ci_nhyps: int}     (* # projectable args *)
 
 let family_eq f1 f2 = match f1, f2 with
-  | Set, Set
-  | Prop, Prop
+  | Set, Set | Prop, Prop -> true
+  | Prop, Type u | Type u, Prop -> Univ.is_type0m_univ u
   | Type _, Type _ -> true
```
SkySkimmer: ??

vernac/comInductive.ml
```diff
@@ -141,7 +141,8 @@ let make_conclusion_flexible sigma = function
      | _ -> sigma)
 
 let is_impredicative env u =
-  u = Prop || (is_impredicative_set env && u = Set)
+  (match u with Type s -> Univ.is_type0m_univ s | _ -> false)
+  || u = Prop || (is_impredicative_set env && u = Set)
```
SkySkimmer: Why's this change needed?

SkySkimmer:
Looks like the elpi overlay isn't working.

maximedenes (approved):
Looks good to me, module the elpi overlay and the last remarks by @SkySkimmer.

herbelin:
I removed the changes pointed by @SkySkimmer and rebased. I hope I did not forget anything important in the merge with SProp. At least CI is green.

SkySkimmer:
@herbelin please open a PR on alpi for the overlay.

herbelin:
OK, done: LPCIC/coq-elpi#63.

PR #8991, opened 2018-11-13
coqide: use correct toplevel name in files

### LINKED ISSUE #8989: CoqIDE does not pass the name of the current modpath
#### Version

The Coq Proof Assistant, version 8.10+alpha (October 2018)
compiled on Oct 25 2018 11:39:02 with OCaml 4.06.1

#### Operating system

Fedora 22

I am developing a plugin for logging changes in Coq as users make them. As part of this, I need to get the module path from which the plugin is loaded. I do this using `Lib.current_mp ()`.

Right now, I have a file `testdebug.v` with the following contents:

```
Require Import Analytical.Analytics. (* my plugin *)
Set Debug Analytics. (* print logging info locally *)
```

When I run this with `coqc`, I get the following output:

```
((time 1542142301.489689) (session-module: testdebug) (session 1542142300.542698)  (Control (StmObserve 3)))
```

Note the `session-module` there, which is taken from `Lib.current_mp ()` at load-time. However, when I open this in CoqIDE and execute it, I get the following instead:

```
((time 1542142348.179430) (session-module: Top) (session 1542142345.429123) (Control (StmObserve 3)))
```

Here, the `session-module` is `Top` instead of `testdebug`. Because of this, if users edit files in CoqIDE, it is currently impossible for my logging tool to tell what files they are editing. I am not sure if this has implications for other parts of the code.

@maximedenes said that this was due to a bug in CoqIDE, which is not passing the current modpath. I'm not sure exactly how this works, but I'm reporting an issue so that we can discuss this. He also pointed out that @ppedrot and @gares would probably know more.

I can provide my plugin code if you are interested, but it is a bit of a pain since it currently requires a branch of Coq not merged into master yet (that @ejgallego is developing) that has hooks into add, remove, and exec exposed to plugins. You can reproduce this by writing a simple plugin however that just logs the current modpath at loadtime:

```
let log =
  let session_module = ModPath.to_string (Lib.current_mp ()) in
  Feedback.msg_notice (Pp.str session_module)
```

SkySkimmer:
Fix #8989.

This adds an option -topfile taking a path so that inferring the right
dirpath is done by the toplevel after processing -Q/-R instead of the
client having to do it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,18 @@
 Changes from 8.9 to 8.10
 ========================
 
+Coqide
+
+- CoqIDE now properly sets the module name for a given file based on
+  its path, see -topfile change entry for more details.
+
+Coqtop
+
+- new option -topfile filename, which will set the current module name
+  (à la -top) based on the filename passed, taking into account the
+  proper -R/-Q options. For example, given -R Foo foolib using
+  -topfile foolib/bar.v will set the module name to Foo.Bar.
+
 OCaml
 
 - Coq 8.10 requires OCaml >= 4.05.0, bumped from 4.02.3 See the
--- a/toplevel/usage.ml
+++ b/toplevel/usage.ml
@@ -32,6 +32,7 @@ let print_usage_channel co command =
 \n  -R dir coqdir          recursively map physical dir to logical coqdir\
 \n  -Q dir coqdir          map physical dir to logical coqdir\
 \n  -top coqdir            set the toplevel name to be coqdir instead of Top\
+\n  -topfile f             set the toplevel name as though compiling f\
 \n  -coqlib dir            set the coq standard library directory\
 \n  -exclude-dir f         exclude subdirectories named f for option -R\
 \n\
--- a/ide/coqide.ml
+++ b/ide/coqide.ml
@@ -89,20 +89,29 @@ let make_coqtop_args fname =
     | Ignore_args -> !sup_args
     | Append_args -> !sup_args
     | Subst_args -> [] in
-  if read_project#get = Ignore_args then "", base_args
-  else
-    match !custom_project_file, fname with
-    | Some (d,proj), _ -> d, coqtop_args_from_project proj @ base_args
-    | None, None -> "", base_args
-    | None, Some the_file ->
-       match
-         CoqProject_file.find_project_file
-           ~from:(Filename.dirname the_file)
-           ~projfile_name:project_file_name#get
-       with
-       | None -> "", base_args
-       | Some proj -> 
-           proj, coqtop_args_from_project (read_project_file proj) @ base_args
+  let proj, args =
+    if read_project#get = Ignore_args then "", base_args
+    else
+      match !custom_project_file, fname with
+      | Some (d,proj), _ -> d, coqtop_args_from_project proj @ base_args
+      | None, None -> "", base_args
+      | None, Some the_file ->
+        match
+          CoqProject_file.find_project_file
+            ~from:(Filename.dirname the_file)
+            ~projfile_name:project_file_name#get
+        with
+        | None -> "", base_args
+        | Some proj ->
+          proj, coqtop_args_from_project (read_project_file proj) @ base_args
+  in
+  let args = match fname with
+    | None -> args
+    | Some fname ->
+      if List.exists (String.equal "-top") args then args
+      else "-topfile"::fname::args
+  in
+  proj, args
 ;;
 
 (** Setting drag & drop on widgets *)
--- a/library/library.ml
+++ b/library/library.ml
@@ -611,28 +611,6 @@ let import_module export modl =
 (************************************************************************)
 (*s Initializing the compilation of a library. *)
 
-let check_coq_overwriting p id =
-  let l = DirPath.repr p in
-  let is_empty = match l with [] -> true | _ -> false in
-  if not !Flags.boot && not is_empty && Id.equal (List.last l) coq_root then
-    user_err 
-      (str "Cannot build module " ++ DirPath.print p ++ str "." ++ Id.print id ++ str "." ++ spc () ++
-      str "it starts with prefix \"Coq\" which is reserved for the Coq library.")
-
-let start_library fo =
-  let ldir0 =
-    try
-      let lp = Loadpath.find_load_path (Filename.dirname fo) in
-      Loadpath.logical lp
-    with Not_found -> Libnames.default_root_prefix
-  in
-  let file = Filename.chop_extension (Filename.basename fo) in
-  let id = Id.of_string file in
-  check_coq_overwriting ldir0 id;
-  let ldir = add_dirpath_suffix ldir0 id in
-  Declaremods.start_library ldir;
-  ldir
-
 let load_library_todo f =
   let longf = Loadpath.locate_file (f^".v") in
   let f = longf^"io" in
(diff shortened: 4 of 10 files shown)


ejgallego (commented):
I do consider `Loadpath.find_load_path` as harmful, in particular this kind of reverse lookup has several pitfalls including forbidding sharing and aliasing of libraries. In fact we should go the other way and have Coq never infer the current logical path.

What's wrong with properly setting top ?

SkySkimmer:
I don't want to refactor the load path system for this PR.

 > What's wrong with properly setting top ?

What does that mean?

Gaëtan Gilbert

ejgallego:
> What does that mean?

Pass the proper -top setting.

SkySkimmer:
How do you get it?

Gaëtan Gilbert

ejgallego:
> How do you get it?

Can we computed from the current package in scope right?

I mean if we have no remedy we can do this invasive fix as is; however I'd rather no rely more on the low-level loadpath mechanism and try to refactor a bit; I could help a bit.

All in all the root of this problem [and of a few others] is that our current notion of _CoqProject has no structure at all, thus, the users knows stuff (flags, current package) but they are lost when transcribing to to _CoqProject, and must be inferred again.

Let me sleep over this and see if we can come with a slightly less invasive tweak.

SkySkimmer:
> Can we computed from the current package in scope right?

What is the "current package in scope"?
The way it's done in this PR is the way it's done in the compilation 
codepath (see Library.start_library).

Gaëtan Gilbert

SkySkimmer:
Here's a commit to make it clearer that we have the same behaviour in 
compile mode and -topfile mode.

Gaëtan Gilbert

ejgallego:
> The way it's done in this PR is the way it's done in the compilation codepath (see Library.start_library).

I'm aware of that, the thing is that this is already problematic so maybe we should think a bit about that.

> What is the "current package in scope"?

That's the thing, currently it is poorly defined as `Library` will determine it based on the filesystem while that part shouldn't really care but the upper parts should. So I am naturally way to perpetuate that pattern by adding more dependencies to the behavior.

In most cases the "current package" is really `-R Foo .`, but that is clearly not enough IMHO.

ejgallego:
> Here's a commit to make it clearer that we have the same behaviour in compile mode and -topfile mode.

That certainly looks much better, now I wonder, why do we need the option at all?

p.s: Note that I didn't click "Request Changes" but just comment; I understand as you say that we cannot afford to improve the situation in this commit; so I am just wondering what would the cleanest way to avoid relying in the reverse lookup too much.

SkySkimmer:
> That certainly looks much better, now I wonder, why do we need the
 > option at all?

How else would coqidetop (or coqtop for PG, etc) know what to do?

I guess we would need a toplevel "Top Module HoTT.Basics.Overture." or 
some such declaration in the file itself ala erlang.

Gaëtan Gilbert

ejgallego:
> How else would coqidetop (or coqtop for PG, etc) know what to do?

They know the file being open, don't they? Thus, they could set it up.

For example in the SerAPI protocol, `Init` does carry the name of the file.

I have an unfinished PR floating around that indeed introduces the notion of "file-backed" interactive document; and in fact allows even to save it.

But I think this PR should not be delayed a lot and thus we could merge as is but have the option not documented for now, just internal to `coqidetop`.

SkySkimmer:
I'd like to have proof general using correct top quickly too, without 
relying on the xml branch.

Gaëtan Gilbert

ejgallego:
> I'd like to have proof general using correct top quickly too, without relying on the xml branch.

Ok for the option then.

ejgallego (approved):
This PR fixes a bug and has several beneficial effects, including removing a use of `Flags.boot` in core.

This will be very useful for further improvements in the future, including compiling the stdlib as a regular lib.

The PR looks OK to me, I will merge soon but I'd like feedback from the IDE maintainers first.

toplevel/usage.ml
```diff
@@ -32,6 +32,7 @@ let print_usage_channel co command =
 \n  -R dir coqdir          recursively map physical dir to logical coqdir\
 \n  -Q dir coqdir          map physical dir to logical coqdir\
 \n  -top coqdir            set the toplevel name to be coqdir instead of Top\
+\n  -topfile f             set the toplevel name as though compiling f\
```
gares: typo, "though" is not what you want to say. I'd say "as if it was compiling f"

gares (changes requested):
well done!

toplevel/usage.ml
SkySkimmer: Doesn't seem to be actually different https://en.wiktionary.org/wiki/as_though
>as though
    As to suggest the idea that; as if, as would be true if.

ejgallego:
@ppedrot any feedback on this?

@gares should your review then be amended?

toplevel/usage.ml
gares: Oh, I learnt a thing ;-)

gares (approved):

ppedrot (approved):

ejgallego:
Good for merge then, however it seems to me that this change would deserve an entry in `CHANGES`, as it is user-visible.

WDYT @SkySkimmer ?

SkySkimmer:
CHANGEd

ejgallego:
Got hit by the rebase daemon.

SkySkimmer:
I have put the demon back in its place.

Gaëtan Gilbert

CHANGES.md
```diff
@@ -1,6 +1,15 @@
...
 
+Coqide
+
+- Automatically set -top appropriately for the current file and -R/-Q
+  options (including from _CoqProject)
+
+Coqtop
+
+- new option -topfile, acting as -top but taking a filesystem path.
+
```
ejgallego: I am not sure this is very understandable for the end-user. How about:

> new option `-topfile filename`, which will set the current module name (à la `-top`) based on the filename passed, taking into account the proper `-R/-Q` options. For example, given `-R Foo foolib` using `-topfile foolib/bar.v` will set the module name to `Foo.Bar`.

CHANGES.md
```diff
@@ -1,6 +1,15 @@
 Changes from 8.9 to 8.10
 ========================
 
+Coqide
+
+- Automatically set -top appropriately for the current file and -R/-Q
+  options (including from _CoqProject)
+
```
ejgallego: How about:

> CoqIDE now properly sets the module name for a given file based on its path, see `-topfile` change entry for more details.

ejgallego (changes requested):
I find the change entries hard to read.

SkySkimmer:
CHANGEd

Gaëtan Gilbert

ejgallego (approved):

ejgallego:
After understanding the situation better here I'd like to propose this for 8.9 as it is really a fix; freshness of keys in ltac do indeed depend on module name given to coqtop so problems may arise in a variety of situations, thus I would feel safer having a consistent module naming among all the frontend for 8.9.

What do you think @SkySkimmer ?

ejgallego:
I am now convinced anyways this is of critical importance for 8.9 so I'll try a backport and then we wait.

SkySkimmer:
>What do you think @SkySkimmer ?

That's more a question for the RM no?

ejgallego:
> That's more a question for the RM no?

Sure, but I guess he will like to have the opinion of the PR author too.

PR #8992, opened 2018-11-13
put protocol/ in ide/.merlin

SkySkimmer:


ppedrot (approved):
I don't understand merlin so I have no idea but it looks trivial so OK.

PR #8995, opened 2018-11-14
Don't redeclare constraints of fields in Include

SkySkimmer:
This avoids forcing foreign opaques when including a module type or functor into a regular module.


SkySkimmer:
Appveyor is some spurious hang.

ppedrot (approved):
I believe this is both correct and desired.

maximedenes (approved):
I can't evaluate the correctness w.r.t. universes, but if correct, I support this patch.

PR #8996, opened 2018-11-14
Fix #8937: inductive conversion in coqchk subtyping

### LINKED ISSUE #8937: coqchk fails on subtyping involving inductives
It's not failing on the inductive itself:
~~~coq
(* -*- coq-prog-args: ("-noinit"); -*- *)

Unset Elimination Schemes.
Module Type S.

Inductive foo : Prop :=.
Definition bar (x:foo) : Prop := match x with end.

End S.

Module M.

Inductive foo : Prop :=.
Definition bar (x:foo) : Prop := match x with end.

End M.

Module MS : S := M.
~~~
~~~
Fatal Error: Anomaly "Uncaught exception Modops.ModuleTypingError(_)."
~~~
(it's failing to convert the types of bar in the module implementation vs module type)

SkySkimmer:
As far as I can tell this is similar to what coqtop does. Delta
resolvers are complicated so I may be mistaken.

The important part is to avoid losing the modified delta resolver
returned by strengthen_and_subst in check_mexpr.

### DIFF
--- a/checker/mod_checking.ml
+++ b/checker/mod_checking.ml
@@ -3,7 +3,6 @@ open Util
 open Names
 open Reduction
 open Typeops
-open Subtyping
 open Declarations
 open Environ
 
@@ -65,17 +64,17 @@ let rec check_module env mp mb =
     check_signature env mb.mod_type mb.mod_mp mb.mod_delta
   in
   let optsign = match mb.mod_expr with
-    |Struct sign -> Some (check_signature env sign mb.mod_mp mb.mod_delta)
+    |Struct sign -> Some (check_signature env sign mb.mod_mp mb.mod_delta, mb.mod_delta)
     |Algebraic me -> Some (check_mexpression env me mb.mod_mp mb.mod_delta)
     |Abstract|FullStruct -> None
   in
   match optsign with
   |None -> ()
-  |Some sign ->
-    let mtb1 = mk_mtb mp sign mb.mod_delta
+  |Some (sign,delta) ->
+    let mtb1 = mk_mtb mp sign delta
     and mtb2 = mk_mtb mp mb.mod_type mb.mod_delta in
     let env = Modops.add_module_type mp mtb1 env in
-    let cu = check_subtypes env mtb1 mtb2 in
+    let cu = Subtyping.check_subtypes env mtb1 mtb2 in
     if not (Environ.check_constraints cu env) then
       CErrors.user_err Pp.(str "Incorrect universe constraints for module subtyping");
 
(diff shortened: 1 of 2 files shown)


ppedrot (commented):
OK, I think this is out my league. We probably need to gather people partially understanding the module system and hope that the kernel is a coherence space (pointwise coherence implies global coherence).

ppedrot:
@herbelin @maximedenes @mattam82 and other are kindly invited to ponder about this fix.

mattam82:
I'm not fluent enough in delta resolvers to review.

herbelin:
I'm not really fluent either, but if noone (besides @skyskimmer) is, I may eventually try to concentrate on it (at a later time).

SkySkimmer:
I'm not fluent in delta resolvers, I just followed along in ocamldebug 
until I found something different in the checker vs the kernel.

Gaëtan Gilbert

herbelin:
@skyskimmer: I agree with the analysis "The important part is to avoid losing the modified delta resolver returned by strengthen_and_subst in check_mexpr". Then, I compared further the code with what the kernel does and found another failing example. Just add the following to the test:
```
Module F (Z:S) := Z.
Module MS : S := F M.
```
which shows that we need also the `subst_codom_delta_resolver` to compose the functor substitution (i.e. `Z:=M`) with the delta-resolver (`MS=Z`) obtained from calling strengthening on the functor. The following seems to fix it:
```
-      Modops.subst_signature (Mod_subst.map_mbid farg_id mp Mod_subst.empty_delta_resolver) fbody_b, delta
+      let subst = Mod_subst.map_mbid farg_id mp Mod_subst.empty_delta_resolver in
+      Modops.subst_signature subst fbody_b, Mod_subst.subst_codom_delta_resolver subst delta
```
I stop the exploration here but the next step should be to understand if we need the delta-resolver attached to the type of the argument of the functor (obtained with `discr_resolver`) and if something has to be done with the inlining. (It is unclear to me at this stage that the checker is delta-resolver-ok.)

SkySkimmer:
Added @herbelin patch.

herbelin:
> I'm mostly concerned with being able to run the checker on the test suite to prevent further regressions.

Then, I believe that this is the best increment we can do at the current time. The PR is ok to me.

maximedenes:
@ppedrot Sorry, I'm not competent on this code.

SkySkimmer:
@herbelin or @ppedrot should merge this then

ppedrot:
@SkySkimmer I'll let @herbelin merge it, so that whenever an unsoundness bug is found I won't be the usual git blame suspect :smiling_imp:

herbelin:
OK, I'll merge it tomorrow. It is better than before, but I don't exclude that the treatment of delta-resolver is incomplete (and possibly be "exploited").

PR #8998, opened 2018-11-15
[legacy proof engine] Remove some cruft.

ejgallego:
We remove the `Proof_types` file which was a trivial stub, we also
cleanup a few layers of aliases.

This is not a lot but every little step helps.


dev/top_printers.mli
```diff
@@ -120,9 +120,9 @@ val ppclenv : Clenv.clausenv -> unit
 
 val ppgoalgoal : Goal.goal -> unit
 
-val ppgoal : Proof_type.goal Evd.sigma -> unit
+val ppgoal : Evar.t Evd.sigma -> unit
```
gares: This should be `Goal.goal`: the parameter of `sigma` is either `Goal.goal` or `Goal.goal list` depending if the `sigma` is the input or output of a legacy tactic.

This comment applies to all other mli files too.

gares (changes requested):

dev/top_printers.mli
ejgallego: I am more of the opinion of keeping this using `Evar.t` as I'd like to get rid of the `Goal` module too. WDTY @ppedrot ?

gares: Note that in the new engine goals and evars are different types. I don't care about the Goal module per se, it may be removed if it is useless, but the Evars.t there plays the role of a goal, hence the type alias `goal` makes the interface more readable. I thing goal should be an opaque alias of evars.t, but I guess that the organizational of files made it never happen.

ejgallego: The thing is that we, in theory, want to remove this interface. I'm sure making the alias opaque is not going to work. So indeed I prefer to expose the deprecated truth but no problem in using the transparent alias either.

ejgallego: so @ppedrot what would you favor, should we keep the `Goal.goal` alias or get rid of it too?

ppedrot: I don't have a strong opinion. We're already exporting functions breaching the abstraction, but the current model doesn't care. I'd be slightly leaning towards keeping Goal to clearly mark deprecated APIs, but that's not completely clear.

ejgallego: Ok, will amend to use `Goal.goal` where possible then.

ejgallego: Amended.

gares (approved):

ejgallego:
Overlays:
- https://github.com/coq-community/aac-tactics/pull/20
- https://github.com/mattam82/Coq-Equations/pull/152

ppedrot (approved):
I forgot to mention that I was OK with the overall changes.

ejgallego:
Rebased, needs assignee, updated overlays.

PR #9003, opened 2018-11-15
[vernacextend] Consolidate extension points API

ejgallego:
We group the extension API and datatypes under `Vernacextend`.

This means that the base plugin dependency is now `coq.vernac` from
`coq.stm`.

This is quite important as for example the LSP server won't like to
link the STM in.

LTAC still depends on the STM by means of the ltac_profile part tho.

The next step could be to move the extension point below `Vernacexpr`.

### DIFF
--- a/coqpp/coqpp_main.ml
+++ b/coqpp/coqpp_main.ml
@@ -374,9 +374,9 @@ let print_rules fmt rules =
 let print_classifier fmt = function
 | ClassifDefault -> fprintf fmt ""
 | ClassifName "QUERY" ->
-  fprintf fmt "~classifier:(fun _ -> Vernac_classifier.classify_as_query)"
+  fprintf fmt "~classifier:(fun _ -> Vernacextend.classify_as_query)"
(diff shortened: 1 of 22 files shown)


ppedrot (approved):
I don't think I fully grasp the motivations but AFAICT the code changes are OK.

ppedrot:
Also, despite the important number of codeowners this change is pretty trivial, so in order not to make @ejgallego wait for too long let's put an ultimatum on commenting this PR on Monday evening.

ejgallego:
> I don't think I fully grasp the motivations but AFAICT the code changes are OK.

Main motivation @ppedrot is to allow to use the vernac interpreter only linking `coq.vernac`; that is to say, it makes the extension points self-contained here.

This is convenient for example to new language servers.

ejgallego:
Ltac failure is real tho, submitting overlay.

ejgallego:
Ltac2 overlay: https://github.com/ppedrot/ltac2/pull/84

PR #9004, opened 2018-11-15
Controlling typing flags

SimonBoulier:
This PR allows to locally (with attributes) / globally (with Set/Unset commands) enable and disable:
  - the guard checking
  - the positivity checking
  - the check of universes (providing "type in type")

For this, the `check_guarded` typing flag is split  into `check_guarded` and `check_positive`.

For the moment the syntax of commands is: `Set Guard Checking`, `Unset Guard Checking`, `Set/Unset Positivity Checking`, `Set/Unset Universes Checking`.
And for attributes: `check_guarded`, `assume_guarded`, `check_positive`, `assume_positive`, `check_universes`, `type_in_type`.

This PR also adds a `Print Typing Flags` command which prints the status of those three flags, and improves a bit the output of `Print Assumptions`.

E.g.
```
#[type_in_type] Definition T := let t := Type in (t : t).

Unset Guard Checking.
Fixpoint e (n : nat) : nat := e n.

Print Typing Flags.
```

**Kind:** feature

Any comment welcome.

### DIFF
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -776,7 +776,7 @@ Simple inductive types
 
       The types of the constructors have to satisfy a *positivity condition*
       (see Section :ref:`positivity`). This condition ensures the soundness of
-      the inductive definition. The positivity checking can be disable using
+      the inductive definition. The positivity checking can be disabled using
       the command :cmd:`Unset Positivity Checking` or the attribute
       ``assume_positive`` (see :ref:`gallina-attributes`).
 
--- a/doc/sphinx/language/gallina-specification-language.rst
+++ b/doc/sphinx/language/gallina-specification-language.rst
@@ -776,7 +776,9 @@ Simple inductive types
 
       The types of the constructors have to satisfy a *positivity condition*
       (see Section :ref:`positivity`). This condition ensures the soundness of
-      the inductive definition.
+      the inductive definition. The positivity checking can be disable using
+      the command :cmd:`Unset Positivity Checking` or the attribute
+      ``assume_positive`` (see :ref:`gallina-attributes`).
 
    .. exn:: The conclusion of @type is not valid; it must be built from @ident.
 
@@ -1500,7 +1502,7 @@ the following attributes names are recognized:
     (see :ref:`programs`).
 
 ``global``, ``local``
-    Take no value, analogous to the ``Global`` and ``Local`` flags
+    Takes no value, analogous to the ``Global`` and ``Local`` flags
     (see :ref:`controlling-locality-of-commands`).
 
 ``deprecated``
@@ -1532,6 +1534,29 @@ the following attributes names are recognized:
         Proof.
           now foo.
         Abort.
+
+``check_guarded``, ``assume_guarded``
+    Takes no value. Enable/disable the guard checking of fixpoints during a
+    definition (see also :ref:`controlling-typing-flags`). Works with
+    :cmd:`Definition`, :cmd:`Fixpoint`, :cmd:`CoFixpoint`, :cmd:`Theorem`
+    (and its variants).
+
+``check_positive``, ``assume_positive``
+    Takes no value. Enable/disable the positivity checking for the declaration
+    of an inductive type or the productivity checking for a coinductive type
+    (see also :ref:`controlling-typing-flags`).
+
+``check_universes``, ``type_in_type``
+    Takes no value. Enable/disable the checking of universes during a definition
+    or the declaration of a (co)inductive type.
+    (see also :ref:`controlling-typing-flags`).
+
+.. example::
+
+   .. coqtop:: all reset
+
+        #[assume_guarded] Fixpoint f (n : nat) : False
+          := f n.
 
 .. [1]
    This is similar to the expression “*entry* :math:`\{` sep *entry*
--- a/doc/sphinx/proof-engine/vernacular-commands.rst
+++ b/doc/sphinx/proof-engine/vernacular-commands.rst
@@ -1202,6 +1202,66 @@ scope of their effect. There are four kinds of commands:
   occurs in a section.  The :cmd:`Set` and :cmd:`Unset` commands belong to this
   category.
 
+
+.. _controlling-typing-flags:
+
+Controlling Typing Flags
+----------------------------
+
+.. cmd:: Set Guard Checking
+.. cmd:: Unset Guard Checking
+
+   Enable/Disable the guard checking of fixpoints. Warning: this can break the
+   consistency of the system, use at your own risk. Decreasing argument can
+   still be specified but the decrease is not checked anymore. Unchecked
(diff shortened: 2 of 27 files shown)


printing/printer.ml
```diff
@@ -995,3 +998,8 @@ let print_and_diff oldp newp =
         pr_open_subgoals ~proof
     in
     Feedback.msg_notice output;;
+
+let pr_typing_flags flags =
+  str "check_guarded: " ++ bool flags.check_guarded
+  ++ str "\ncheck_positive: " ++ bool flags.check_positive
```
SkySkimmer: Use an `fnl()` instead of `\n`.

test-suite/success/typing_flags.v
```diff
@@ -0,0 +1,82 @@
...
+  Global Unset Universes Checking.
+  (* Disable Type In Type. *)
+  Definition T := let t := Type in (t : t).
+  (* Print Typing Flags. *)
+End b.
+(* Print Typing Flags. *)
+Import b.
+(* Print Typing Flags. *)
+(* About T. *)
+(* Print Assumptions T. *)
```
SkySkimmer: Remove the comments or make this an output test.

vernac/comFixpoint.ml
```diff
@@ -261,7 +261,7 @@ let declare_fixpoint local poly ((fixnames,fixdefs,fixtypes),pl,ctx,fiximps) ind
         fixdefs) in
     let evd = Evd.from_ctx ctx in
     Lemmas.start_proof_with_initialization (local,poly,DefinitionBody Fixpoint)
-      evd pl (Some(false,indexes,init_tac)) thms None (Lemmas.mk_hook (fun _ _ -> ()))
+      evd pl (Some(false,indexes,init_tac)) thms None (Lemmas.mk_hook (fun _ _ -> Global.set_typing_flags original_typing_flag))
```
SkySkimmer: This seems quite fragile. Let's try to come up with a better model for how to do this kind of thing.

SkySkimmer (changes requested):

vernac/comFixpoint.ml
SkySkimmer: Maybe it should be handled more like how Program Mode is done.
OTOH
~~~coq
Require Import Program.Tactics.
Program Lemma foo : Type.
Proof.
  Test Program Mode. (* off *)
~~~
so that doesn't quite work.

In general attributes which act by setting a global option seem difficult to get right especially with proof-starting commands like Lemma.

Does it even make sense to have these typing flags in the global environment? Maybe we should pass them through the constant_entry/mind_entry/etc.This would mean handling them more like universe polymorphism, with an option which is used only to get the default value of the attribute. start_proof takes a goal_kind, if we put typing flags in there we may be able to make it work in the proof too.

printing/printer.ml
SimonBoulier: Done. Thanks

vernac/comFixpoint.ml
SimonBoulier: Do other coq devs have any opinion?
Is it blocking or is it an "improvement request"?
Is `goal_kind` the right place for typing flags?

printing/printer.ml
herbelin: Additionally, you can encapsulate the code into a box (e.g. a `v 0 (...)` to ensure that the `fnl` will align the next `check` exactly under the first `check` and not under the beginning of the box which encapsulates the call to `pr_typing_flags` (even if they probably coincide as the code is currently).

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -776,7 +776,9 @@ Simple inductive types
 
       The types of the constructors have to satisfy a *positivity condition*
       (see Section :ref:`positivity`). This condition ensures the soundness of
-      the inductive definition.
+      the inductive definition. The positivity checking can be disable using
```
Zimmi48:
```suggestion
      the inductive definition. The positivity checking can be disabled using
```

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -776,7 +776,9 @@ Simple inductive types
...
+      the command :cmd:`Unset Positivity Checking` or the attribute
```
Zimmi48:
```suggestion
      the flag :flag:`Positivity Checking` or the attribute
```

doc/sphinx/language/gallina-specification-language.rst
```diff
@@ -776,7 +776,9 @@ Simple inductive types
...
+      ``assume_positive`` (see :ref:`gallina-attributes`).
```
Zimmi48: Would be good to show a small example, which may also help explaining what the use cases are. Actually, since you have put this example in another section, a link to this section would also be fine (use `.. seealso::`).

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -1202,6 +1202,66 @@ scope of their effect. There are four kinds of commands:
...
   category.
 
+
+.. _controlling-typing-flags:
+
+Controlling Typing Flags
+----------------------------
+
+.. cmd:: Set Guard Checking
+.. cmd:: Unset Guard Checking
```
Zimmi48: You should instead document this as:

```rst
.. flag:: Guard Checking
```

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -1202,6 +1202,66 @@ scope of their effect. There are four kinds of commands:
...
+
+   Enable/Disable the guard checking of fixpoints. Warning: this can break the
```
Zimmi48: Flag to control...

Use `.. warning::`

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -1202,6 +1202,66 @@ scope of their effect. There are four kinds of commands:
...
+.. cmd:: Unset Universes Checking
+
+   Enable/Disable the checking of universes, providing a form of "type in type".
+   Warning: this breaks the consistency of the system, use at your own risk.
+   Constants relying on "type in type" are printed by :cmd:`Print Assumptions`.
+
+.. cmd:: Print Typing Flags
+
+   Print the status of the three typing flags: check of guard, check of positivity
+   and check of universes.
```
Zimmi48: This command is not necessary as it redoes the job of `Test Universe Checking`, `Test Positivity Checking`, `Test Guard Checking`.

Zimmi48 (changes requested):

SimonBoulier:
The part without attributes has been merged.

vernac/comFixpoint.ml
ejgallego: The right place will be in `Declare.CInfo.t` once #12372 is finished.

ejgallego:
Rest implemented in #12586

PR #9005, opened 2018-11-15
More informative error on vo validation failure

SkySkimmer:
Now that the checker shares code with the kernel using md5 cic.mli
doesn't work. We could md5 declarations.ml but this would miss changes
to constr (and cic.mli already missed changes to names, univs).

Instead we just print a bit of information (the last seen type
name/annotation) when validate fails. This should help debugging when
forgetting to update values.ml without the full verbosity of -debug.

As such there is no need to -debug in the makefile validate
target (NB: CI has an independent implementation of the validate rule
since the vos are installed there).

### DIFF
--- a/checker/values.mli
+++ b/checker/values.mli
@@ -10,17 +10,36 @@
 
 type value =
   | Any
+  (** A value that we won't check, *)
+
   | Fail of string
+  (** A value that shouldn't be there at all, *)
+
   | Tuple of string * value array
+  (** A debug name and sub-values in this block *)
+
   | Sum of string * int * value array array
+  (** A debug name, a number of constant constructors, and sub-values
+     at each position of each possible constructed variant *)
+
   | Array of value
   | List of value
   | Opt of value
   | Int
   | String
+  (** Builtin Ocaml types. *)
+
   | Annot of string * value
+  (** Adds a debug note to the inner value *)
+
   | Dyn
+  (** Coq's Dyn.t *)
+
   | Proxy of value ref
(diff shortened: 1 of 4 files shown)


checker/values.mli
```diff
@@ -10,17 +10,44 @@
...
+
   | Tuple of string * value array
+  (** A debug name and sub-values in this block *)
+
   | Sum of string * int * value array array
+  (** A debug name, a number of constant constructors, and sub-values
+     at each position of each possible constructed variant *)
+
   | Array of value
+  (** Ocaml array *)
```
gares: If you really think the information convoyed by this comment is relevant you can rename the type "ocaml_value" or the constructor OCamlArray. I'd drop all all the comment on Array, List, String, Opt and Int. Maybe put a single comment "OCaml containers and primitive data types" and move these constructors under it.

gares (dismissed):
The Makefile change (for which I've been notified) is OK I guess.
But I did read the rest of the PR and I believe that many comments are mostly noise.

ppedrot (approved):
The changes look OK, and I don't care enough to complain about the tautological comments.

gares:
> I don't care enough to complain about the tautological comments.

Hey, but you commented here :P

checker/values.mli
gares: Feel free to ignore this, but maybe next time don't write the comments in the first place.

SkySkimmer:
Changed comments.

gares (approved):

PR #9006, opened 2018-11-15
[dune] Add quickide target for building of IDE.

ejgallego:
Note that `states` doesn't work reliably yet, but that is a separate
problem that will be fixed in Dune 1.6.

[Or we could also fix it improving the rules in envars.ml]

### DIFF
--- a/Makefile.dune
+++ b/Makefile.dune
@@ -3,7 +3,7 @@
 
 .PHONY: help voboot states world watch check          # Main developer targets
 .PHONY: coq coqide coqide-server                      # Package targets
-.PHONY: quickbyte quickopt                            # Partial / quick developer targets
+.PHONY: quickbyte quickopt quickide                   # Partial / quick developer targets
 .PHONY: refman-html stdlib-html apidoc                # Documentation targets
 .PHONY: test-suite release                            # Accesory targets
 .PHONY: ocheck trunk ireport clean                    # Maintenance targets
@@ -27,6 +27,7 @@ help:
 	@echo ""
 	@echo "  - quickbyte: build main ML files [coqtop + plugins + ide + printers] using the bytecode compiler"
 	@echo "  - quickopt:  build main ML files [coqtop + plugins + ide + printers] using the optimizing compiler"
+	@echo "  - quickide:  build main IDE files [client + server + prelude] using the optimizing compiler"
 	@echo ""
 	@echo "  - test-suite:  run Coq's test suite"
 	@echo "  - refman-html: build Coq's reference manual [HTML version]"
@@ -76,6 +77,11 @@ quickbyte: voboot
 
 quickopt: voboot
 	dune build $(DUNEOPT) $(QUICKOPT_TARGETS)
+
+quickide: states
+	dune build $(DUNEOPT) @topworkers
+	dune build $(DUNEOPT) coqide-server.install
+	dune build $(DUNEOPT) coqide.install
 
 test-suite: voboot
 	dune runtest --no-buffer $(DUNEOPT)
--- a/topbin/dune
+++ b/topbin/dune
@@ -36,3 +36,8 @@
  (modules :standard \ coqtop_byte_bin coqtop_bin coqc_bin)
  (libraries coq.toplevel)
  (link_flags -linkall))
+
+; Workers installed targets
+(alias
+ (name topworkers)
+ (deps %{bin:coqqueryworker.opt} %{bin:coqtacticworker.opt} %{bin:coqproofworker.opt}))


gares:
Thanks!
Just a minor remark: why "quickide" and not "ide" or "coqide"? What makes it quicker?

gares (approved):

ejgallego:
I named it in accordance to the other targets, the idea is that in the future we should be way faster and just build the minimal needed stuff. As coqide is built by default I thought to use a more marked naming. We can rename the rule of course..

ejgallego:
I named it in accordance to the other targets, the idea is that in the future we should be way faster and just build the minimal needed stuff. As coqide is built by default I thought to use a more marked naming. We can rename the rule of course..

Makefile.dune
```diff
@@ -1,7 +1,7 @@
 # -*- mode: makefile -*-
 # Dune Makefile for Coq
 
-.PHONY: help voboot states world watch check quickbyte quickopt test-suite release apidoc ocheck ireport clean
+.PHONY: help voboot states world watch check quickbyte quickopt quickide test-suite release apidoc ocheck ireport clean
```
Zimmi48: I think you should start splitting this list like in `Makefile.dune`.

Zimmi48 (approved):

ejgallego:
> Building from an empty state, the CoqIDE produced by this target cannot be run as is. Is this expected?

Oh indeed, there are two problems here:
- one is the same problem that `make states` has, [do we have an issue for it] and you must pass `-coqlib $(pwd)/_build/default` to coqide for now. It will be fixed when Dune 1.6 is released, other fixes would be possible as discussed on Gitter, namely adding a smarter logic to `lib/envars`.
- the other one is that the STM doesn't gracefully degrade if the workers are not available, indeed that is a bug on this PR, will fix.

ejgallego:
> * the other one is that the STM doesn't gracefully degrade if the workers are not available, indeed that is a bug on this PR, will fix.

Fixing this properly may require Dune 1.6 too, as we would need to depend on the installed version of the workers; will update the PR nonetheless but this will mean that you can only call coqide manually and not by using `dune exec`.

ejgallego:
New PR pushed, unfortunately it won't work 100% fine until either:
- Dune 1.6 is released,
- or we add smarter rules [not worth doing IMHO].

Makefile.dune
ejgallego: I will do the split in this PR indeed, one sec.

ejgallego: Well, on the other hand I like the global PHONY as it helps me keeps the help list in sync. I dunno, let me think about it. Sparse phony targets is a problem the current makefile system suffers a lot.

Zimmi48:
When is Dune 1.6 due? If this is soon, then it might make sense to wait instead of merging this now.

ejgallego:
> When is Dune 1.6 due? If this is soon, then it might make sense to wait instead of merging this now.

I'd say before 2019; so do as you prefer indeed, no problem on waiting.

Note these kind of targets [such as `states`] are still useful as you can usually do `world + quickide` and things will work fine. But indeed, that's an ugly hack.

I'd love to hear suggestions about how to handle the location of build-tree artifacts; so far I have been busy with other stuff as to propose a proper design, using the "installed" version seems sensible, but we should do better I think.

Zimmi48:
@gares Do you think we should merge the current patch with its limitations now or wait?

gares:
I think it would be better to wait.

BTW, the `.PHONY:` directive can be repeated, so you can still split the like and have all the PHONY targets be listed on a bunch on adjacent lines. Eg:
```make
.PHONY: a b c d
.PHONY: e f g h
```

ejgallego:
> BTW, the `.PHONY:` directive can be repeated, so you can still split the line and have all the PHONY targets be listed on a bunch on adjacent lines. Eg:

Indeed that is fine to me, I didn't predict the line would grow so long.

ejgallego:
IMO this is ready; the `states` target must be fixed but that'd happen in a separate PR.

Zimmi48 (changes requested):

topbin/dune
```diff
@@ -28,3 +28,8 @@
  (modules :standard \ coqtop_byte_bin coqtop_bin)
  (libraries coq.toplevel)
  (link_flags -linkall))
+
+; Should be changed to depend on the installed targets when Dune 1.6 is released
```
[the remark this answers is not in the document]
ejgallego: Oh indeed, thanks for pointing this out.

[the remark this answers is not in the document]
Zimmi48: ping @ejgallego 
Either fix this or add "needs: progress" / "needs: fixing"

ejgallego:
TTBOMK this is OK now, we still need a shim to get quicker running of the states target, à la #9047 but indeed if this is merged I could update #9047 to add a `coqide-prelude` shim.

Zimmi48 (approved):

Zimmi48:
Indeed, it would be good to have in addition a trick to be able to run `dune exec coqide` after `make -f Makefile.dune quickide` without having to also run `make -f Makefile.dune world` and kill it after a few seconds. Merging this one now anyway.

ejgallego:
Thanks!

PR #9010, opened 2018-11-15
Evd: align type interfaces

### LINKED ISSUE #9009: `evd.mli` does not reflect `evd.ml`
#### Version

8.10+alpha

#### Operating system
- Ubuntu 14.04.5 LTS
- macOS 10.13.3

In QuickChick [Build#748](https://travis-ci.org/QuickChick/QuickChick/builds/455734972):
> This expression has type Constr.t
but an expression was expected of type Evd.etypes = Evd.econstr

According to [evd.ml](https://github.com/SkySkimmer/coq/blob/master/engine/evd.ml#L24), these types are the same. However, such equivalence is not reflected in [evd.mli](https://github.com/SkySkimmer/coq/blob/master/engine/evd.mli#L31).

liyishuai:
**Kind:** bug fix

Fixes #9009 

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,6 +59,8 @@ Tactics
   (e.g. `?[n]` or `?n` in terms - not in patterns) are now interpreted
   the same way as other variable names occurring in Ltac functions.
 
+- Aligned type interfaces of `Evd` with their definitions.
+
 Vernacular commands
 
 - `Combined Scheme` can now work when inductive schemes are generated in sort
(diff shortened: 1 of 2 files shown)


ppedrot:
As explained in the corresponding bug report, the whole endeavour is precisely to separate those two types, by *definition*. I'll consequently close this PR. You might be rightfully complaining about the relative lack of documentation for EConstr, but even that is briefly described in that module. If so, please open a bug report about that instead.

PR #9013, opened 2018-11-16
Do not Export the init modules

SkySkimmer:
It seems we started doing the export silently in
47804492bd09c8b13b5aac45800d067dbdf04d00.

I think it's surprising and undesired behaviour so changing it back.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -93,6 +93,10 @@ Standard Library
 
 - Added `ByteVector` type that can convert to and from [string].
 
+- The prelude used to be automatically Exported and is now only
+  Imported. This should be relevant only when importing files which
+  don't use -noinit into files which do.
+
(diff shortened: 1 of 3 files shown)


ppedrot:
It was not observable except for the few people using `-noinit` right? Or were there some objects from the prelude imported in random orders due to Require semantics and hiding stuff defined later on?

SkySkimmer:
Appveyor failure is spurious (stm/ test).

ejgallego:
@ppedrot I let you handle this one as I am not sure I understand what the effects are here.

ejgallego (commented):
Delegating to more expert people.

ppedrot (approved):
I agree it is an undesired behaviour, but it's probably uncommon to hit an observable difference.

ppedrot:
Will merge quickly if nobody objects.

PR #9015, opened 2018-11-16
[Typeclasses] Warn when RHS of `:>` is not a class

### LINKED ISSUE #9014: In Class definitions, :> is silently ignored when the RHS is not a Class
When defining a class, using `:>` enables to declare “substructures”, as described in the [manual](https://coq.inria.fr/distrib/current/refman/addendum/type-classes.html?highlight=class#substructures). In the following example, `a_of_b` is declared as an instance, enabling to deduce an instance of `A` from an instance of `B`.

~~~
Variant T := mkT.
Class A := { p : Prop }.
Class B := { a_of_b :> A ; t_of_b :> T }.
~~~

However, when the right-hand-side of `:>` is not a class, as in `t_of_b` above, the “substructure” annotation is **silently** ignored.

In usual records, this annotation declares a coercion. Should this also happen with classes? At least, I suggest to emit a warning instead of only ignoring the annotation.

vbgl:
Fixes #9014.

Is there a better way of printing a `GlobRef.t`?

Should the projection be registered as a coercion instead?

NB: I also included in this PR an unrelated cleaning patch.

### DIFF
--- a/pretyping/typeclasses.ml
+++ b/pretyping/typeclasses.ml
@@ -434,28 +434,40 @@ let remove_instance i =
   Lib.add_anonymous_leaf (instance_input (RemoveInstance, i));
   remove_instance_hint i.is_impl
 
-let declare_instance info local glob =
+let warning_not_a_class =
+  let name = "not-a-class" in
+  let category = "typeclasses" in
+  CWarnings.create ~name ~category (fun (n, ty) ->
+      let env = Global.env () in
+      let evd = Evd.from_env env in
+      Pp.(str "Ignored instance declaration for “"
+          ++ Nametab.pr_global_env Id.Set.empty n
+          ++ str "”: “"
+          ++ Termops.Internal.print_constr_env env evd (EConstr.of_constr ty)
+          ++ str "” is not a class")
+    )
+
+let declare_instance ?(warn = false) info local glob =
   let ty, _ = Typeops.type_of_global_in_context (Global.env ()) glob in
   let info = Option.default {hint_priority = None; hint_pattern = None} info in
     match class_of_constr Evd.empty (EConstr.of_constr ty) with
     | Some (rels, ((tc,_), args) as _cl) ->
       assert (not (isVarRef glob) || local);
       add_instance (new_instance tc info (not local) glob)
-    | None -> ()
+    | None -> if warn then warning_not_a_class (glob, ty)
 
 let add_class cl =
   add_class cl;
   List.iter (fun (n, inst, body) ->
-	     match inst with
-	     | Some (Backward, info) ->
-	       (match body with
-	       | None -> CErrors.user_err Pp.(str "Non-definable projection can not be declared as a subinstance")
-	       | Some b -> declare_instance (Some info) false (ConstRef b))
-	     | _ -> ())
-  cl.cl_projs
+      match inst with
+      | Some (Backward, info) ->
+        (match body with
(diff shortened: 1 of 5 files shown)


pretyping/typeclasses.ml
```diff
@@ -434,14 +434,23 @@ let remove_instance i =
   Lib.add_anonymous_leaf (instance_input (RemoveInstance, i));
   remove_instance_hint i.is_impl
 
-let declare_instance info local glob =
+let warning_not_a_class =
+  let name = "not-a-class" in
+  let category = "typeclasses" in
+  CWarnings.create ~name ~category (fun n ->
+      Pp.(str "Ignored instance declaration: "
+          ++ Constr.debug_print (printable_constr_of_global n)
```
SkySkimmer: Use Nametab.shortest_qualid_of_global instead and print what it returns instead.

pretyping/typeclasses.mli
```diff
@@ -133,7 +133,7 @@ val remove_instance_hint : GlobRef.t -> unit
 val solve_all_instances_hook : (env -> evar_map -> evar_filter -> bool -> bool -> bool -> evar_map) Hook.t
 val solve_one_instance_hook : (env -> evar_map -> EConstr.types -> bool -> evar_map * EConstr.constr) Hook.t
 
-val declare_instance : hint_info option -> bool -> GlobRef.t -> unit
+val declare_instance : ?warn:bool -> hint_info option -> bool -> GlobRef.t -> unit
```
maximedenes: Why do you want to warn only sometimes?

vbgl: Because this function is called in other situations. E.g., when declaring section variables.

pretyping/typeclasses.ml
vbgl: Thus I have a similar question: how can I print a `types` (a.k.a. `Constr.t`)?

pretyping/typeclasses.mli
gares: I see. I think a comment (or the label name) should hint at when the warning is triggered

vbgl: Just added a comment.

pretyping/typeclasses.ml
SkySkimmer: Use Termops.Internal.print_constr_env

vbgl:
@mattam82 Matthieu, would you like to comment on this PR? Thanks.

mattam82 (approved):
Looks good to me. The warning will surely help highlight the ambiguity with coercions to users.

Zimmi48:
@mattam82 You're the only code owner so the natural assignee here.
That being said, @vbgl shouldn't this new warning get documented? (here: https://coq.inria.fr/refman/addendum/type-classes.html#substructures)

mattam82:
@Zimmi48 right. The CI failure is spurious I suppose. I agree it needs a doc entry. And I don't think the projection should be declared as a coercion instead, to answer your initial question. We might want to move to an unambiguous syntax though (sorry for my own mistake pushing this confusion in the codebase initially :/).

vbgl:
Just added documentation to the manual. Does it deserve a line in `CHANGES`?

mattam82:
@vbgl I don't think so. I'll merge once CI passes.

Zimmi48:
@mattam82 CI has passed.

Zimmi48 (approved):

mattam82:
It might be good to turn this into an error in the following version

PR #9017, opened 2018-11-16
Remove SSR profiling

jfehrle:
Deletes the SSR profiling feature, including the `SsrProfiling` and `SsrMatchingProfiling` options.
Requested by @gares in #8911. 

**Kind:** cleanup

**It's OK to delete these directly, rather than deprecate for one release?**
**Does this need a CHANGES.md entry?  These options were not documented**


gares (approved):

Zimmi48:
Just for the sake of deprecating, you could open a PR for 8.9.0 which deprecates these options.

jfehrle:
I think I'll skip a deprecation PR for 8.9 for undocumented options, including `Debug Ltac`, etc. in the main PR.  No need to mention in CHANGES.md, either?

Zimmi48:
Indeed. It makes sense. Given that these are debug options, this doesn't have any impact on compatibility anyways.

PR #9019, opened 2018-11-17
[STM] Simplify cache logic and interface

maximedenes:
Before this patch, functions fetching states from the cache were
expecting clients to check if the state was present and had the expected
shape (success state,...). Querying whether a state was cached could
itself trigger caching of states.

We simplify the interface by providing essentially two functions: one
for installing a state from the cache, and one for getting a state (to
be sent to a worker). The logic related to the current state is also
encapsulated in this interface, rather than carried out by the clients.


ppedrot (commented):
This is black magic, don't expect me to review this kind of change meaningfully.

maximedenes:
This is ready for review.

maximedenes:
@gares can I assign it to you?

library/summary.mli
```diff
@@ -11,17 +11,12 @@
 (** This module registers the declaration of global tables, which will be kept
    in synchronization during the various backtracks of the system. *)
 
-type marshallable =
-  [ `Yes       (* Full data will be marshalled to disk                        *)
-  | `No        (* Full data will be store in memory, e.g. for Undo            *)
-  | `Shallow ] (* Only part of the data will be marshalled to a slave process *)
-
```
ejgallego: You are removing documentation and IMO not properly explaining the semantics of the new boolean flag. Please improve.

stm/stm.mli
```diff
@@ -169,6 +169,8 @@ val finish_tasks : string ->
   Library.seg_univ -> Library.seg_discharge -> Library.seg_proofs ->
     tasks -> Library.seg_univ * Library.seg_proofs
 
+val get_success_state : doc:doc -> Stateid.t -> Vernacstate.t option
```
ejgallego: Adding a new API function and no documentation. Please document this properly.

stm/stm.mli
```diff
@@ -283,9 +285,6 @@ val forward_feedback_hook : (Feedback.feedback -> unit) Hook.t
 
 val get_doc : Feedback.doc_id -> doc
 
-val state_of_id : doc:doc ->
-  Stateid.t -> [ `Valid of Vernacstate.t option | `Expired | `Error of exn ]
```
ejgallego: How can clients now retrieve the reason of a missing state? As the new function has no documentation it is hard to tell.

ejgallego (dismissed):
New undocumented API needs to be documented.

stm/stm.mli
maximedenes: It is a bit hard to design the API since we never discussed the interaction model (we should do it in the second UI-related meeting). Can you give scenarios were that knowledge is useful to clients?

ejgallego: Well it seems pretty obvious, if a client gets none, what does that mean? Is it pending, was there an error? Anyways hard to tell as the function is not documented.

ejgallego: We could rely on other methods of course, but first we need a spec for get_sucess_state

library/summary.mli
maximedenes: Funnily enough, the boolean was already documented a few lines below. Probably a left there from before the commit which introduced the `Shallow` logic that this PR removes. What do you want me to improve?

maximedenes:
The fiat-crypto OOM may reveal that we are now caching more. Will investigate.

maximedenes:
@ejgallego One key design aspect of this State API is that `cur_id` is encapsulated, modulo the check in `parse_sentence`, that is going away in the parsing state PR. Then we can make `cur_id` private to `State`.

stm/stm.mli
maximedenes: Right, I added some documentation now.

ejgallego:
> @ejgallego One key design aspect of this State API is that `cur_id` is encapsulated, modulo the check in `parse_sentence`, that is going away in the parsing state PR. Then we can make `cur_id` private to `State`.

Umm, making `cur_id` private will change nothing, it used to private already; I only exposed it for debug info and it is still used only once.

library/summary.mli
ejgallego: IMHO the below comment as well as this one it does explain nothing. Ok, mashalled to disk, what's the difference?

stm/stm.ml
```diff
@@ -782,8 +782,19 @@ module State : sig
   exception Not_in_cache
 
   val install_state : cache:bool -> Stateid.t -> unit
+  (** Put the system in the state described by the input state id. If the state
+  is missing from the cache, raise [Not_in_cache], except if the requested state
+  is the current one, in which case we snapshot and cache it if [cache] is true,
+  and do nothing otherwise. *)
+
   val get_success_state : Stateid.t -> Vernacstate.t option
+  (** Return the success stat designated by the input id. If it is the current
```
ejgallego: typo `stat`

stm/stm.mli
ejgallego: Ah OK, you put the comments in the `.ml` file so I wasn't seeing it. I think the doc reads well, tho I am not sure why success states are more special than failed ones.

From the point of view of the client, I would expect that doing `(Check sid)` would eventually make `sid` to always have a state, either failed or successful. This extra logic complicates things for now reason IMHO.

Also, the magic going on about "if the current state is sid, do this" seems pretty weird to me, but that I guess you have a need for it. But I would expect states to be installed explicitly.

ejgallego:
Documentation seems OK, I will do another review on the second pass.

ejgallego:
> Querying whether a state was cached could
> itself trigger caching of states.

How could that happen? At least from the public `state_of_id` that was not possible, this function had a much simpler semantics than the replacement.

maximedenes:
> How could that happen?

See the former `cache` argument to `is_cached`.

> At least from the public `state_of_id` that was not possible, this function had a much simpler semantics than the replacement.

Indeed, but `state_of_id` had a semantics that was incompatible with what we do in `coqc`, because it was expecting to find the state in the cache even when it is the current one. This is incorrect in `coqc` mode, where we don't store all states in the cache.

The new API does have a correct equivalent of `state_of_id`, which is `get_state`. Initially I exposed it, but then realized all uses in STM cared only about success states, hence `get_success_state`.

In order to know what clients need (especially if they need `get_state` rather than just `get_success_state`), I'd need to know more about what they want to do with error states.

ejgallego:
> See the former `cache` argument to `is_cached`.

Well the same happens now in the patch, so this has not changed [just that cache=true now, right?]

> Indeed, but `state_of_id` had a semantics that was incompatible with what we do in `coqc`, because it was expecting to find the state in the cache even when it is the current one. This is incorrect in `coqc` mode, where we don't store all states in the cache.

Well, in "coqc" mode the clients assumed the cache was disabled so they expected to find nothing indeed. It seems strange you want to access the cache in `coqc` mode [there should be no cache].

> In order to know what clients need (especially if they need `get_state` rather than just `get_success_state`), I'd need to know more about what they want to do with error states.

I think that'd be useful indeed to provide error messages on failed states, but anyways that's not a big deal, should be easy to add later.

maximedenes:
> Well, in "coqc" mode the clients assumed the cache was disabled so they expected to find nothing indeed. It seems strange you want to access the cache in `coqc` mode [there should be no cache].

You need to put at least some state in the cache to be able to send them to workers.

> Well the same happens now in the patch, so this has not changed [just that cache=true now, right?]

What has changed is that it is the function which returns the state that can populate the cache. I find it more intuitive than when it was done by a function returning whether the state was in cache or not.

> I think that'd be useful indeed to provide error messages on failed states, but anyways that's not a big deal, should be easy to add later.

Note that it is already there (`State.get_state`), the question is whether we want to export it.

ejgallego:
> You need to put at least some state in the cache to be able to send them to workers.

That seems unfortunate, isn't it?

ejgallego:
> That seems unfortunate, isn't it?

I mean, it seems you are adding something in the cache just to get it back?

maximedenes:
> I mean, it seems you are adding something in the cache just to get it back?

It also serves for `Undo`.

ejgallego:
> It also serves for `Undo`.

well in that case you are in interactive mode, not in coqc mode.

maximedenes:
> well in that case you are in interactive mode, not in coqc mode.

Why? If you `coqc -async-proofs on` a file which contains `Undo`, you'll be in that scenario.

gares:
The fact that the STM uses the cache also as a storage for some states it knows will be needed is unfortunate but has always been the case, it is not changed by this pr

ejgallego:
> Why? If you `coqc -async-proofs on` a file which contains `Undo`, you'll be in that scenario.

That has nothing to do with async proofs, even without them, as of today you need to force the cache otherwise Undo will bite you.

ejgallego:
> The fact that the STM uses the cache also as a storage for some states it knows will be needed is unfortunate but has always been the case, it is not changed by this pr

My fear here is that this PR will store more things in the cache in "coqc" mode, right?

maximedenes:
> That has nothing to do with async proofs, even without them, as of today you need to force the cache otherwise Undo will bite you.

That's a very imprecise statement. It won't "bite" you, it will replay steps from the last cached states (typically the beginning of the proof). Coq used to work like that in interactive mode, so it does not seem crazy to do for the few cases where you want to store `Undo` in documents that you `coqc` (typically, tutorials).

maximedenes:
> My fear here is that this PR will store more things in the cache in "coqc" mode, right?

If it does, it is a bug. What makes you say that?

gares:
@maximedenes did you run a pendulum job?

maximedenes:
> did you run a pendulum job?

Not yet, and it should be done. But I probably need to investigate the fiat-crypto issue before.

maximedenes:
I'm going to split this PR in smaller patches to make it easier to review / integrate / base other patches on it.

ejgallego:
> I'm going to split this PR in smaller patches to make it easier to review / integrate / base other patches on it.

Does that mean this PR should be closed?

maximedenes:
This will be replaced by the new document manager, let's close.

PR #8924, opened 2018-11-06
Misc updates to codeowners

Zimmi48:
**Kind:** meta.

The main change is the new @coq/parsing-maintainers team (for now consisting of Emilio, Hugo and Pierre-Marie). This follows an idea emitted by @ejgallego and this fixes an absence of code owners for some recently added files.

PR #9024, opened 2018-11-19
[meta] Update CODEOWNERS for @parsing-maintainers

ejgallego:
`parsing` and `gramlib` are clear; metasyntax could be split up so
the parsing parts would live in parsing and the `Notation` command
parts would live in `vernac`.

For now it seems more appropriate to have it owned by @coq/parsing-maintainers.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -132,8 +132,14 @@
 
 ########## Parser ##########
 
-/parsing/         @herbelin
+/parsing/         @parsing-maintainers
 # Secondary maintainer @mattam82
+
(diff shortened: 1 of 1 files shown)


Zimmi48 (changes requested):
You need to use `@coq/parsing-maintainers`. But how is this not a duplicate of #8924?

Zimmi48:
If you have tweaks to make @ejgallego, I would prefer that you push to #8924 to adapt it.

ejgallego:
Oh sorry, somehow I lost track of that PR in the current sea. Let me reconcile.

PR #9030, opened 2018-11-20
CI: Run Mtac2 tests.

Janno:
`make` does not run tests in Mtac2. That requires `make test`.

**Kind:** infrastructure.

### DIFF
--- a/dev/ci/ci-mtac2.sh
+++ b/dev/ci/ci-mtac2.sh
@@ -9,4 +9,4 @@ git_download unicoq
 
 git_download mtac2
 
-( cd "${CI_BUILD_DIR}/mtac2" && coq_makefile -f _CoqProject -o Makefile && make )
+( cd "${CI_BUILD_DIR}/mtac2" && coq_makefile -f _CoqProject -o Makefile && make && make test )


ejgallego:
@Janno can you merge https://github.com/Mtac2/Mtac2/pull/126 ? Thanks!

ejgallego (approved):

Janno:
I am on it but I need to test it locally first since our CI is kinds of messed up right now and I don't want to risk making test suite failures the norm.

ejgallego:
> I am on it but I need to test it locally first since our CI is kinds of messed up right now and I don't want to risk making test suite failures the norm.

I suggest you use the Docker file from https://github.com/coq-community/docker-coq/wiki

https://github.com/erikmd/docker-coq-travis-ci-demo-1/blob/master/.travis.yml

It should take a couple of minutes to setup.

Janno:
Thanks, I already followed your recommendation but I can't get travis to rebuild existing PR's after changing the branch to be merged into.

Janno:
Yes, I see your point. For the time being, though, this seems like the most pragmatic approach. If we could rely on a more standardized notion of tests then I will be more than happy to separate the phases again.

As it stands right now, even we (Mtac developers) forget to run tests all the time and whenever we have intermittent CI failures we end with broken commits that only surface whenever somebody (else) remembers to run tests or fixes the CI. I suppose this could be fixed by tooling somehow but given my limited time to deal with infrastructure I will delay that problem a bit longer and go with the shortcut for now.

Regarding parallelism: I see how this could matter, theoretically, but in practice we have no opportunity for fan-out with Mtac2 at this moment.

PR #9032, opened 2018-11-20
 checker: check inductive types by roundtrip through the kernel. 

SkySkimmer:
I think this is more reliable than having the separate implementation. Note that the ocaml warnings will tell us if we start forgetting fields.

Secret motive: it should also be easier to work with for SProp (in which the checker is currently broken).

### DIFF
--- a/checker/checkInductive.ml
+++ b/checker/checkInductive.ml
@@ -8,264 +8,155 @@
 (*         *     (see LICENSE file for the text of the license)         *)
 (************************************************************************)
 
-open Sorts
-open Pp
 open Declarations
 open Environ
 open Names
-open CErrors
 open Univ
 open Util
-open Constr
 
-let check_kind env ar u =
-  match Constr.kind (snd (Reduction.dest_prod env ar)) with
-  | Sort (Sorts.Type u') when Univ.Universe.equal u' (Univ.Universe.make u) -> ()
-  | _ -> failwith "not the correct sort"
+[@@@ocaml.warning "+9+27"]
 
-let check_polymorphic_arity env params par =
-  let pl = par.template_param_levels in
-  let rec check_p env pl params =
-    let open Context.Rel.Declaration in
-    match pl, params with
-        Some u::pl, LocalAssum (na,ty)::params ->
-          check_kind env ty u;
-          check_p (push_rel (LocalAssum (na,ty)) env) pl params
-      | None::pl,d::params -> check_p (push_rel d env) pl params
-      | [], _ -> ()
-      | _ -> failwith "check_poly: not the right number of params" in
-  check_p env pl (List.rev params)
+exception InductiveMismatch of MutInd.t * string
 
-let conv_ctxt_prefix env (ctx1:rel_context) ctx2 =
-  let rec chk env rctx1 rctx2 =
-    let open Context.Rel.Declaration in
-    match rctx1, rctx2 with
-        (LocalAssum (_,ty1) as d1)::rctx1', LocalAssum (_,ty2)::rctx2' ->
-          Reduction.conv env ty1 ty2;
-          chk (push_rel d1 env) rctx1' rctx2'
-      | (LocalDef (_,bd1,ty1) as d1)::rctx1', LocalDef (_,bd2,ty2)::rctx2' ->
-          Reduction.conv env ty1 ty2;
-          Reduction.conv env bd1 bd2;
-          chk (push_rel d1 env) rctx1' rctx2'
-      | [],_ -> ()
-      | _ -> failwith "non convertible contexts" in
-  chk env (List.rev ctx1) (List.rev ctx2)
+let check mind field b = if not b then raise (InductiveMismatch (mind,field))
 
-(* check information related to inductive arity *)
-let typecheck_arity env params inds =
-  let nparamargs = Context.Rel.nhyps params in
-  let nparamdecls = Context.Rel.length params in
-  let check_arity arctxt = function
-    | RegularArity mar ->
-        let ar = mar.mind_user_arity in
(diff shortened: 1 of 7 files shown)


SkySkimmer:
Test suite is dying because we now check recargs which involves comparing inductive names so we get #8937. Not sure about compcert, will look once the first bug has been resolved.

SkySkimmer:
>Test suite is dying because we now check recargs which involves comparing inductive names so we get #8937. Not sure about compcert, will look once the first bug has been resolved.

Looks like that wasn't quite it. Let's see how it goes now.

checker/checkInductive.ml
```diff
@@ -8,264 +8,154 @@
...
+  | (RegularArity _ | TemplateArity _), _ -> false
+
+(* Use [eq_ind_chk] because when we rebuild the recargs we have lost
+   the knowledge of who is the canonical version.
+   Try with to see test-suite/coqchk/include.v *)
+let eq_recarg a1 a2 = match a1, a2 with
+  | Norec, Norec -> true
+  | Mrec i1, Mrec i2 -> eq_ind_chk i1 i2
+  | Imbr i1, Imbr i2 -> eq_ind_chk i1 i2
+  | (Norec | Mrec _ | Imbr _), _ -> false
```
SkySkimmer: This is a bit adhoc but better than before (where we just ignored the recargs field).

checker/checkInductive.ml
```diff
@@ -8,264 +8,155 @@
...
+  }
+
+let check_arity env ar1 ar2 = match ar1, ar2 with
+  | RegularArity ar, RegularArity {mind_user_arity;mind_sort} ->
+    Constr.equal ar.mind_user_arity mind_user_arity &&
+    Sorts.equal ar.mind_sort mind_sort
+  | TemplateArity ar, TemplateArity {template_param_levels;template_level} ->
+    List.equal (Option.equal Univ.Level.equal) ar.template_param_levels template_param_levels &&
+    UGraph.check_leq (universes env) template_level ar.template_level
+    (* template_level is inferred by indtypes, so functor application can produce a smaller one *)
```
SkySkimmer: new change

SkySkimmer:
Looks like it works now. See github comments for what changed.

SkySkimmer:
@ppedrot ping

ppedrot (commented):
Summoning @herbelin, @mattam82 and @maximedenes because this is dark magic.

checker/checkInductive.ml
```diff
@@ -8,264 +8,155 @@
...
-  let infos = Array.map (Indtypes.infos_and_sort env_params) lc in
-  let (small,unit) = small_unit infos in
-  (* We accept recursive unit types... *)
-  (* compute the max of the sorts of the products of the constructor type *)
-  let min = if Array.length lc > 1 then Universe.type0 else Universe.type0m in
-  let level = Array.fold_left (fun max l -> Universe.sup max l) min infos in
-  check_predicativity env_ar s small level;
-  allowed_sorts small unit s
+  ignore mind_typing_flags;
+  (* TODO non oracle flags *)
```
maximedenes: Do you plan to address this before merging?

test-suite/coqchk/inductive_functor_params.v
```diff
@@ -0,0 +1,15 @@
...
+
+Module F (A:T).
+  Inductive ind (n:nat) : nat -> Prop :=
+  | C : (forall x, x < n -> ind (A.foo n) x) -> ind n n.
+End F.
+
+Module A. Definition foo (n:nat) := n. End A.
+
+Module M := F A.
+(* Trick: M.ind can be seen as having 1 recursively uniform parameter
```
maximedenes: I think you mean non-uniform, right?

maximedenes (approved):
It's hard to review precisely, but this change was indeed on my list. Note that the complexity here is due to the strange gap between the kernel input and output data, especially because some universe-related operations should probably be performed outside the kernel.

checker/checkInductive.ml
SkySkimmer: >Do you plan to address this before merging?

No.

test-suite/coqchk/inductive_functor_params.v
SkySkimmer: No I do mean uniform.
F.ind has mind_nparams_rec = 0
M.ind could have mind_nparams_rec = 1 if module substitution did the necessary work (we don't want it to do that work at least for perf reasons though IMO).

maximedenes: Ok, so then I don't understand what the "trick" is. Maybe the comment could be made clearer, like:

```
Note: M.ind could be seen as having 1 recursively uniform parameter, but module substitution does not recognize it, so it is treated as a non-uniform parameter.
```

checker/checkInductive.ml
maximedenes: Were they already ignored before this PR?

SkySkimmer: Yes (look around `let env0 =` and its uses before the PR)

test-suite/coqchk/inductive_functor_params.v
SkySkimmer: >Ok, so then I don't understand what the "trick" is.

I was thinking "this asks the checker a trick question".

Comment changed.

herbelin:
> Summoning @herbelin, @mattam82 and @maximedenes because this is dark magic.

Since @maximedenes already took some time to review this PR, I'm trusting his review and I don't think I need to look at the PR myself.

ppedrot:
@mattam82 ?

ppedrot:
Let's merge, this is a mess anyways.

PR #9033, opened 2018-11-20
gitlab: Install stdlib doc in build:base

SkySkimmer:
Looks like we forgot to adapt this when we split off the sphinx job
and stopped using -with-doc yes.


ejgallego (approved):
That's fine I think.

I have a patch adding coqdoc support to coq_dune [the actual Dune PR] so it knows how to build documentation for a library, so that's my own plan regarding coq library doc building.

Issue #9037, opened 2018-11-20
anomaly in abstract with evars in the proof term


andres-erbsen:
#### Version

22c0b10f139d9a30fcbe4a5a489022e2b94130e9 (master from yesterday)

#### Operating system

Arch Linux, up to date as of Sunday.

#### Description of the problem

~~~~~
Goal False.
  Fail abstract (let pf := open_constr:(_) in exact pf).
  Fail abstract (idtac; let pf := open_constr:(_) in exact pf).
  Fail let pf := open_constr:(_) in abstract (refine pf).
(* Error: Anomaly "Uncaught exception "Anomaly: Evar ?X3 was not declared."." *)
  Fail let pf := open_constr:(_) in abstract (exact pf).
(* Error: Anomaly "Uncaught exception "Anomaly: Evar ?X3 was not declared."." *)
  Fail let pf := open_constr:(_) in abstract (exact_no_check pf).
(* Error: Anomaly "Uncaught exception Not_found." *)
Abort.
~~~~~

ejgallego:
This is normal as the `build_by_tac` family of functions don't allow you to send them an update `evar_map`. The below patch improves things but I am not sure is the right thing to do:
```diff
diff --git a/proofs/pfedit.ml b/proofs/pfedit.ml
index 81122e6858..cc66d979d5 100644
--- a/proofs/pfedit.ml
+++ b/proofs/pfedit.ml
@@ -130,11 +130,10 @@ open Decl_kinds
 
 let next = let n = ref 0 in fun () -> incr n; !n
 
-let build_constant_by_tactic id ctx sign ?(goal_kind = Global, false, Proof Theorem) typ tac =
-  let evd = Evd.from_ctx ctx in
+let build_constant_by_tactic id sigma sign ?(goal_kind = Global, false, Proof Theorem) typ tac =
   let terminator = Proof_global.make_terminator (fun _ -> ()) in
   let goals = [ (Global.env_of_context sign , typ) ] in
-  Proof_global.start_proof evd id goal_kind goals terminator;
+  Proof_global.start_proof sigma id goal_kind goals terminator;
   try
     let status = by tac in
     let open Proof_global in
diff --git a/proofs/pfedit.mli b/proofs/pfedit.mli
index 155221947a..908f87ee9f 100644
--- a/proofs/pfedit.mli
+++ b/proofs/pfedit.mli
[...]

PR #9038, opened 2018-11-20
[proof] [abstract] Pass evar_map to functions for constant building.

ejgallego:
This improves #9037, however it is not clear if we want to do this as
it would allow proof bodies to potentially depend on evars from
outside; anyways such proof bodies should be properly grounded before
abstraction so this could work.

### DIFF
--- a/proofs/pfedit.ml
+++ b/proofs/pfedit.ml
@@ -130,11 +130,10 @@ open Decl_kinds
 
 let next = let n = ref 0 in fun () -> incr n; !n
 
-let build_constant_by_tactic id ctx sign ?(goal_kind = Global, false, Proof Theorem) typ tac =
-  let evd = Evd.from_ctx ctx in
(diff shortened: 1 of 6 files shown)


ejgallego:
For this change to work we would need to restrict the `evar_map` in `build_constant_by_tactic` to the evars appearing in the term, @mattam82 gave recently an example for use by the Mtac developers; whether this pattern makes sense should indeed be discussed.

mattam82:
What does this achieve? This changes the anomaly "evar not found" with a different error when grounding the proof term?

ejgallego:
> What does this achieve? This changes the anomaly "evar not found" with a different error when grounding the proof term?

Yes.

Maybe the right fix here would be indeed to check that the input term is not going to be open for the context? That could be more effective maybe? But it would rule out some cases.

Hard to know what's best here, however indeed I think we should do something about the current situation where the term is indeed badly-scoped.

ejgallego:
#9315 was closed and I also feel that this one should be closed for now as it is too far from working.

PR #9041, opened 2018-11-21
[Windows CI] Do not build any addon if WINDOWS is not enabled_all_addons.

### LINKED ISSUE #9040: [ci] Windows CI builds are testing too much making them impractical
Current Windows CI builds are often in red, due to them testing not just Coq, but a wide variety of 3rd party contributions that are often broken by PRs.

This means that even breaking a small plugin will result in a red status for the Windows build, a so far they are red so often that we end up ignoring them; #8045 has a bit of blame here too.

Please, disable building of such artifacts in pull requests, or at least don't make the build fail if a 3rd party development does.

Zimmi48:
**Kind:** infrastructure.

Fixes / closes #9040.

### DIFF
--- a/dev/ci/gitlab.bat
+++ b/dev/ci/gitlab.bat
@@ -39,6 +39,10 @@ SET PATH=%PATH%;C:\Program Files\7-Zip\;C:\Program Files\Microsoft SDKs\Windows\
 
 IF "%WINDOWS%" == "enabled_all_addons" (
   SET EXTRA_ADDONS=^
+    -addon=bignums ^
+    -addon=equations ^
+    -addon=ltac2 ^
+    -addon=mtac2 ^
     -addon=mathcomp ^
     -addon=menhir ^
     -addon=menhirlib ^
@@ -56,10 +60,6 @@ IF "%WINDOWS%" == "enabled_all_addons" (
 call %CI_PROJECT_DIR%\dev\build\windows\MakeCoq_MinGW.bat -threads=1 ^
   -arch=%ARCH% -installer=Y -coqver=%CI_PROJECT_DIR_CFMT% ^
   -destcyg=%CYGROOT% -destcoq=%DESTCOQ% -cygcache=%CYGCACHE% ^
-  -addon=bignums ^
-  -addon=equations ^
-  -addon=ltac2 ^
-  -addon=mtac2 ^
   %EXTRA_ADDONS% ^
   -make=N ^
   -setup %CI_PROJECT_DIR%\%SETUP% || GOTO ErrorCopyLogFilesAndExit
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -229,7 +229,6 @@ windows64:
   <<: *windows-template
   variables:
     ARCH: "64"
-  allow_failure: true
 
 windows32:
   <<: *windows-template
@@ -237,7 +236,6 @@ windows32:
     ARCH: "32"
   except:
     - /^pr-.*$/
-  allow_failure: true
 
 pkg:opam:
   stage: test
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -1905,6 +1905,9 @@ function make_addon_quickchick {
 function make_addons {
   # Note: ':' is the empty command, which does not produce any output
   : > "/build/filelists/addon_dependencies.nsh"
+  : > "/build/filelists/addon_strings.nsh"
+  : > "/build/filelists/addon_descriptions.nsh"
+  : > "/build/filelists/addon_sections.nsh"
 
   for addon in $COQ_ADDONS; do
     "make_addon_$addon"


dev/ci/gitlab.bat
```diff
@@ -39,6 +39,11 @@ SET PATH=%PATH%;C:\Program Files\7-Zip\;C:\Program Files\Microsoft SDKs\Windows\
 
 IF "%WINDOWS%" == "enabled_all_addons" (
   SET EXTRA_ADDONS=^
+    -addon=aactactics ^
```
SkySkimmer: Sneakily start building aac in all_addons mode?

Zimmi48: Right. It was commented out because it was not working with 8.9 and master but it should now. Doing this change as part of this PR might not be a good idea though because this change should also be backported.

Zimmi48:
This job should not build any addon: https://gitlab.com/coq/coq/-/jobs/123445088
This job which is part of a manually started pipeline with `WINDOWS=enabled_all_addons` should build all addons successfully: https://gitlab.com/coq/coq/-/jobs/123448351

ejgallego (approved):
Thanks @Zimmi48 ; I am not competent on the code of the Windows setup; but I prefer to build no addons to the current situation where the whole build fails.

This has the downside of not testing the build of addons in the windows part, however the real problem here is that the amount of duplicated recipes in the Windows part is just not sustainable; the windows build should start to integrate better with current infrastructure IMHO, but it may be worth waiting for upstream build support changes as to simplify this.

Zimmi48:
>I prefer to build no addons to the current situation where the whole build fails.

The idea is that we can always run a manual pipeline if we want test addons on a PR. And we probably also want a cron job or something like this testing the full build from time to time.

>the amount of duplicated recipes in the Windows part is just not sustainable; the windows build should start to integrate better with current infrastructure

Sure. In fact, with the Dune setup, we should aim towards having almost zero Windows-specific code (the only specific code should be building the actual installer). The current infrastructure for building a bundle with addons would be very useful in other systems too and it will probably be much easier to maintain once it's all based on a Dune setup.

Zimmi48:
Anyways, we can't properly test this PR because the Windows build is actually broken.

Zimmi48:
cc @MSoegtropIMC (it seems the problem is occurring during the build of nsis, so probably not related to any change on Coq's side).

ejgallego:
> And we probably also want a cron job or something like this testing the full build from time to time.

+1 indeed.

I was going to open a different issue for the build problems, but I am not sure the problem is in NSIS, see the log:
```
+16:38:29 build_prep_overlay aactactics
+16:38:29 urlvar=aactactics_CI_ARCHIVEURL
+16:38:29 gitvar=aactactics_CI_GITURL
+16:38:29 refvar=aactactics_CI_REF
/cygdrive/c/ci/cygwin64_31227_26923\build\makecoq_mingw.sh: line 421: !urlvar: unbound variable
```
so it seems the `acc` addition to blame here.

ejgallego:
For other jobs, is mtac2 that is failing.

Zimmi48:
OK so the aac-tactics addition should definitely not be sneaked here. I'll remove it.

MSoegtropIMC:
Can someone explain what are the reasons that these plugins break? We deliberately decided to always put in bignums (because it used to be part of the standard library) and ltac2 and mtac2 because we said that it is important that people start experimenting with the new tactics languages. I am not so sure why equations is in the short list.

So is there a way to improve our infrastructure so that these plugins break less frequently, e.g. selecting specific approved tags for the plugins rather than the master branch?

If they break, what do we do when a release comes? Spend a few weeks to fix at least the most important plugins?

The NSIS failure is new to me, I will have a look at it.

ejgallego:
> If they break, what do we do when a release comes? Spend a few weeks to fix at least the most important plugins?

Indeed we should keep building the plugins for `master`, however it is in the pull requests that they break, and this makes the windows testing less useful.

It is expected that PR break plugins.

MSoegtropIMC:
@Zimmi48 : I can't see any failure in NSIS. The latest failure I saw in this PR was this:
```
+16:38:29 for addon in $COQ_ADDONS
+16:38:29 make_addon_aactactics
+16:38:29 installer_addon_dependency aac
+16:38:29 build_prep_overlay aactactics
+16:38:29 urlvar=aactactics_CI_ARCHIVEURL
+16:38:29 gitvar=aactactics_CI_GITURL
+16:38:29 refvar=aactactics_CI_REF
/cygdrive/c/ci/cygwin64_31227_26923\build\makecoq_mingw.sh: line 421: !urlvar: unbound variable
```
which probably means that aactactics_CI_ARCHIVEURL is undefined in the overlay files

Zimmi48:
@MSoegtropIMC We break APIs all the time and we fix plugins: this is part of the normal development process. The problem is that in the current setup, this will often result in Windows being red as well, and this pushes people to start ignoring it, which is not good. As long as we frequently test that all plugins compile fine, it's not a big deal not to do so on PRs (except when we suspect they might break the build).

Zimmi48:
Here the build failed even if we don't build any addon. Please have a look.

ejgallego:
```
SectionGroup Coq addons ->(Sec2)
!include: could not find: "..\..\..\filelists\addon_sections.nsh"
Error in script "coq_new.nsi" on line 132 -- aborting creation process
```

MSoegtropIMC:
```
!include: could not find: "..\..\..\filelists\addon_sections.nsh"
Error in script "coq_new.nsi" on line 132 -- aborting creation process
```
The addon_sections.nsh file is added on by each addon and if there is no addon, the file doesn't exist - a bug. One could simply touch the file in the build script, so that it does exist, but this would lead to an empty addon section in the installer. Better would be to check if the file exists and not add the addon section at all in case the addons file doesn't exist. I will fix it tomorrow afternoon.

Zimmi48:
In addition to this bug, as remarked by Emilio, the Windows build is currently failing everywhere else during the build of mtac2 which does not find UniCoq (see Mtac2/Mtac2#127).

MSoegtropIMC:
I added the following comment 6 days ago:

The addon_sections.nsh file is added on by each addon and if there is no addon, the file doesn't exist - a bug. One could simply touch the file in the build script, so that it does exist, but this would lead to an empty addon section in the installer. Better would be to check if the file exists and not add the addon section at all in case the addons file doesn't exist.

I see this comment in my mail inbox from github (Wed 11/21 11:39 PM) but not here. Does someone have an explanation for this?

I promised to fix it, but was then super busy with preparation of some sport events (a situation which will continue until this weekend).

How do we want to proceed? As I said a quick fix would be to just touch addon_sections.nsh somewhere in the makecoq_mingw.sh. Thinking about it I don't think it is much of an issue to have an empty installer section in CI builds. As far as I know we still plan to have all addons in release builds.

I would add this code in makecoq_mingw.sh
´´´
function make_addons {
  # Note: ':' is the empty command, which does not produce any output
  : > "/build/filelists/addon_dependencies.nsh"
NEW CODE STARTS HERE
  : > "/build/filelists/addon_strings.nsh"
  : > "/build/filelists/addon_descriptions.nsh"
  : > "/build/filelists/addon_sections.nsh"
´´´

The reason I use : > instead of touch is that when incremental builds are done, the files need to be cleared, because the addon logic is outside of the flag file protected build logic cause of the automated module dependency handling (modules with have build dependencies automatically get installer dependencies).

Zimmi48:
>I see this comment in my mail inbox from github (Wed 11/21 11:39 PM) but not here. Does someone have an explanation for this?

Your comment is just above: https://github.com/coq/coq/pull/9041#issuecomment-440832726

Do you want to push another commit implementing your simpler proposal to this branch?

MSoegtropIMC (commented):
Changes look good and I expect this to work, but I will wait until the windows64 CI pass and approve and merge then asap.

MSoegtropIMC (approved):
I checked the installer. As expected it has a useless empty plugin section, but for CI purposes this shouldn't be an issue. I will wait for the rest of the CI and merge then.

Zimmi48:
The rest of the CI is finished and the failures are unrelated (AppVeyor is timing out a lot these days / formal-topology has an overlay which is waiting to be merged).

MSoegtropIMC:
Is it normal that appveyor fails? I couldn't find anything in the logs what indicates the reason for the failure. Is it a timeout?

Should I merge anyway?

Zimmi48:
Yes, as I said in my comment above, this is a time out.

MSoegtropIMC:
> Yes, as I said in my comment above, this is a time out.

sorry, since a week or so I have issues that my browser doesn't show recent (or even not so recent) messages, even after a refresh and even though it shows my posts which came after. Also update of test results seems to be broken for me. This is really quite bizarre. I need to check if this is a proxy server thing.

Zimmi48:
I have had similar issues with GitHub sometimes: it looks like a (local storage) caching problem. I think you should reach out to GitHub support (since you see these problems very often you might be able to give them enough info). Are you using Firefox? (I am)

PR #9043, opened 2018-11-21
[windows] Cleanup cruft from `dev/build/windows`

ejgallego:
The amount of cruft we are carrying there is high enough as to even
difficult navigation.

More cleanup should be performed, but this is a first step.


ejgallego:
Ping!

MSoegtropIMC:
Since this is not terribly urgent, I would prefer if we could first fix the windows64 build issues (#9041) and rebase this to it. It all looks good but I don't feel comfortable doing cleanup work on an entity when this entitie's CI is not operational.

MSoegtropIMC:
@ejgallego : #9041 is merged. Can you please rebase?

MSoegtropIMC (changes requested):
I missed two things on my first quick browse of the PR:

The removal of the gtk related patches is not ok. The build script still has an option to build gtk from sources and this is useful for two reasons: first for debugging CoqIDE and second it is not sure that Cygwin will support GTK-2 until we move to GTK-3 (or away from CoqIDE).

Regarding the quickchick patch: did you check that quickchick does build now without this patch? I requested a change upstream (informally) but didn't check recently if it arrived. The reason we need a native build is that the Windows build comes only with a native Coq and cannot load bytecode plugins.

ejgallego:
> The removal of the gtk related patches is not ok

As far as I can see the GTK patch is still there. I've removed old patches for unsupported version.

> Regarding the quickchick patch: did you check that quickchick does build now without this patch? I requested a change upstream (informally) but didn't check recently if it arrived. The reason we need a native build is that the Windows build comes only with a native Coq and cannot load bytecode plugins.

The last version of the PR built well indeed; however I am not sure what the status is, maybe we'd like to relaunch the windows job with the proper variable set so quickchick gets built?

ejgallego:
Ping @MSoegtropIMC

Zimmi48:
Indeed, this PR probably build the version without the addon. I'll launch a new build with `WINDOWS=enabled_all_addons`.

MSoegtropIMC:
@Zimmi48 : please note that I didn't take care of the AAC fix as yet - possibly it was fixed by some changes in the overlays file already. The last few weeks was tournament season in the sports I am doing, so I have a bit of a backlog. Any ideas when the deadline for 8.9 is? I still have to fix the VST stuff, which is a bit of work It should compile, but doesn't work properly).

Zimmi48:
Windows build failed during Equations build but probably because of a recently merged overlay (given that it also failed for most plugins in CI). @ejgallego Can you rebase and then relaunch a pipeline with `WINDOWS=enabled_all_addons`?

ejgallego:
https://gitlab.com/coq/coq/-/jobs/130599802

Zimmi48:
@MSoegtropIMC aac-tactics is not built in this PR even with `enabled_all_addons` so it is orthogonal. I have no idea what the deadline for 8.9 is. IMO it should have released already but Guillaume is completely unresponsive and has not been backporting any PRs in a while (last backport was 27 days ago).

ejgallego:
@Zimmi48 mtac2 is still broken so the build will fail again; we may restart it once they merge the overlay at https://github.com/Mtac2/Mtac2/pull/128

ejgallego:
Ideally the build wouldn't fail if an addon fails tho.

MSoegtropIMC:
> Ideally the build wouldn't fail if an addon fails tho.

As I said, my plan is to separate the Windows installer build into these jobs

- build of Windows MinGW OCaml, tools and Coq (possibly separate out OCaml and tools as docker)
- parallel build of all plugins
- build of the installer

This way one gets a much better overview about what failed. But this would mean quite a bit of rework of the pipeline and I am not really sure how to do it. It would be good to have an online discussion on this topic some time.

MSoegtropIMC:
I am still more busy than expected, but I will definitely look over it until Monday. I anyway also have to finish VST and look over all plugins for 8.9 and there is a lot of overlap with this and general windows64 stability and cleanup considerations.

MSoegtropIMC:
@ejgallego : can you please rebase so that we can see if the windows64 build passes now?

MSoegtropIMC (changes requested):
I reviewed the list of removed patches once more. I think still required are
```
dev\build\windows\patches_coq\gtksourceview-2.11.2.patch
dev\build\windows\patches_coq\quickchick.patch
```
The rest can indeed be removed.

The quickckick patch is required because it is still not adopted in quickchick master (I checked today). And I am not sure they will adopt it soon, because for the authors a byte code build with debug capabilities is likely more suitable, but we require a native build. In the end they have to rewrite their makefile to support both.

The gtksourceview patch is still required because this is the version used (if GTK is build from sources which takes ages but is useful for debugging CoqIDE).

For testing this in CI one would need to either have a CI option to build GTK from sources or to have a CI option to add arbitrary additional command line arguments to MakeCoq_MinGW.bat (I would prefer the latter). If you want I could also add this to this PR.

ejgallego:
> The rest can indeed be removed.

Done thanks @MSoegtropIMC !

MSoegtropIMC (approved):

MSoegtropIMC:
The Windows build fails cause of Ltac-2 again ...

Would you mind if I add to this PR a change to allow passing arbitrary command line arguments to MakeCoq_MinGW, so that one can add arbitrary sets of plugins? That should be an easy change and would allow to easily get around this kind of CI problem.

ejgallego:
> Would you mind if I add to this PR a change to allow passing arbitrary command line arguments to MakeCoq_MinGW, so that one can add arbitrary sets of plugins?

Sure, I don't mind at all.

ejgallego:
Ping @MSoegtropIMC

MSoegtropIMC:
@ejgallego : thanks for the ping. I will do the option to test arbitrary plugin sets separately - otherwise the review stuff gets too complicated.

ejgallego:
Thanks!

Issue #9012, opened 2018-11-16
LSP server support meta issue


maximedenes:
This is a meta-issue to communicate around an on-going effort by @ejgallego, @gares, @siegebell, @vbgl and myself, trying to improve Coq's support for UI implementations.

The plan is to improve the STM API and provide an LSP server, probably based on https://github.com/rlepigre/lambdapi/tree/master/lp-lsp

We plan to connect it to @siegebell 's VsCoq client: https://github.com/siegebell/vscoq

Currently, @gares and I are working on #566 which should resolve blocking issues related to parsing state.

The corresponding project can be seen here: https://github.com/coq/coq/projects/30

herbelin:
Thanks for the news.

Among the questions I have, I'd be interested in understanding how to organize the code implementing CoqIDE features such that it is reusable for LSP. For instance, if I want to implement a graphical "search" widget for both CoqIDE and LSP (or VsCoq) what parts of the feature can I factorize?

herbelin:
See also some information on LSP and VScode at #7055.

lilred:
I'd like to help with the LSP effort. Do you know of any relatively self-contained tasks that still need doing?

Zimmi48:
It would be helpful to have some feedback on the current status of the effort. In particular, I recently read in SerAPI's Gitter room + on Twitter that @ejgallego has a branch implementing a working LSP protocol for Coq. It would helpful to share this branch with the world, even if in a completely unpolished state and with many unmerged dependencies, so that people can evaluate the work that has still to be done.

ejgallego:
Hi folks, I will publish the branch soon, job interviews are keeping me too busy so sorry for the delay.

I'm afraid that until I finish the ongoing rework there is not a lot that can be parallelized, I am working on the base architecture [which is extremely tricky due to inter-process communication issues] and that indeed requires care. I won't publish the branch until I am not sure I am not going to rewrite the whole thing next day.

brendanzab:
Just cross-linking the discussion here, which might be of interest: https://github.com/siegebell/vscoq/issues/167

I'd recommend posting public branches in general @ejgallego - sometimes people can give you advice that actually speeds stuff up! In saying that, I know it can be hard to do if you've been working on something for a while and it's still unfinished and unpolished. Anyway, good luck with the job hunt! 🙂

ejgallego:
[...]

PR #9044, opened 2018-11-21
Remove pidetop from CI

maximedenes:
pidetop relies on some rather low level API from STM, which we'd like to
change. It does not seem maintained much anymore, and still hasn't moved
to github.

If more information is required, I think @gares can describe the situation more precisely.


maximedenes:
The motivation for this change is to make upcoming work like #9019 easier, by not having to maintain pidetop on our side (by submitting overlays to bitbucket, etc).

On the other hand, I'll be happy to provide assistance to UI maintainers if they need to adjust code to API changes.

ejgallego (dismissed):
I don't think allow failures makes a lot of sense here; we rather support `pidetop` or we rather remove it in full. What is the point of having a zombie here?

Regarding what to do, @gares indeed has the word on that.

maximedenes:
> I don't think allow failures makes a lot of sense here; we rather support `pidetop` or we rather remove it in full. What is the point of having a zombie here?

Ok, removing it is also fine with me. Will update the patch.

maximedenes:
@ejgallego updated to remove rather than allowing failures

ejgallego (commented):
Up to @gares to decide, CI changes are OK.

gares:
The plan is to restore the Coqoon backend speaking the XML protocol, since the PIDE one is shaky.
The XML protocol never went away fully, and the code is parametrized by that, so it should be doable, and in the future it should be doable to plug LSP in there. @ale-f did not manage to update the XML protocol to the current variant yet. When it will be done I think we should probably unplug pidetop.

I'd personally like to see at least the failure we have today on pidetop while changing the STM API, so maybe allowed-failures is a reasonable compromise in the meanwhile.

ejgallego:
It was not clear from your message @gares if you approve this PR or not. What should we do?

Except for very specific, special cases (*), I'm afraid that `allow_failure: true` is not an allowed situation for a CI entry; if we know a CI is broken we are not gonna launch hundredths of useless jobs for them to fail.

If the plan is allow `pidetop` to be broken for some time and then to rewrite that's fine; we can remove it now, and add it back later on.

(*) So far we are allowing bedrock first because it mostly succeeds, and second because of a bug they cannot implement the proper fix [sadly the Coq PR has been complicated to get thu] Also, we will allow OCaml trunk to fail for obvious reasons, but even this job should mostly succeed once we polish the details, otherwise it should be removed.

gares:
Pidetop is broken by a pr of Maxime, if I can write a fix easily then it can stay. But I need to see the failure, the pr is not up to date so I don't know if it is fixable. Please don't merge it now

maximedenes:
> Pidetop is broken by a pr of Maxime, if I can write a fix easily then it can stay. But I need to see the failure, the pr is not up to date so I don't know if it is fixable. Please don't merge it now

I think there's a misunderstanding. I expect the failure to be easy to fix, it is not the motivation of this PR. The motivation is the cost (in terms of delay) of porting an unmaintained interface to a new API, especially by means of overlays.

But maybe I should reverse the question: why am I expected to do the porting or wait for somebody to do it each time I touch an STM API? What's the goal, since as you say, the whole thing is probably going to be replaced.

maximedenes:
I believe it would be a better strategy to save my time and let me help porting Coqoon to LSP when the time comes.

gares (approved):

ejgallego:
Ok, I understand @gares green tick as an indication to go ahead with this PR then. Will merge later today.

herbelin:
Actually, do we have an idea of how many people are using Coqoon? Is it used e.g. for teaching? Do we have an idea of the plans of the developers of Coqoon?

If needed, a concrete agreement could maybe then be made with them on helping them to rebind Coqoon on top of LSP?

gares:
> Do we have an idea of the plans of the developers of Coqoon?

I meant *their* plan in:

> The plan is to restore the Coqoon backend speaking the XML protocol, since the PIDE one is shaky.
> The XML protocol never went away fully, and the code is parametrized by that, so it should be doable, and in the future it should be doable to plug LSP in there. @ale-f did not manage to update the XML protocol to the current variant yet. When it will be done I think we should probably unplug pidetop.

ejgallego (approved):
Merging as per upstream-author.

herbelin:
@gares: Thanks.

I have many other questions which may however require to enter technical details. So, if ever you have time for explaining, is there a way to summarize the (conceptual) defects / advantages of pidetop wrt the xml protocol (independently of the cost of maintenance of this or that proposal)?

If too complicated, don't worry.

maximedenes:
> I have many other questions which may however require to enter technical details. So, if ever you have time for explaining, is there a way to summarize the (conceptual) defects / advantages of pidetop wrt the xml protocol (independently of the cost of maintenance of this or that proposal)?

Could we move this discussion somewhere else?

herbelin:
> Could we move this discussion somewhere else?

Sure, maybe on #9012? Or another issue? The UIs issues are so fascinating. There are so much to do, and very interesting things to do.

How to proceed? Do I let you open a new issue? Do we let @gares open an issue if he wants to contribute?

maximedenes:
> Sure, maybe on #9012? Or another issue? The UIs issues are so fascinating. There are so much to do, and very interesting things to do.

I guess it depends what you want to ask about. If your questions are related to the LSP server, #9012 seems like the right place. Otherwise, just open a new issue.

PR #9047, opened 2018-11-21
[dune] Shim for starting `coqtop/coqide` with minimal config.

ejgallego:
As requested by @Skyskimmer, we add shims

- `dev/shim/coqtop-prelude`
- `dev/shim/coqide-prelude`

that will build and start `coqtop` and `coqide` with just the prelude
loaded properly.

`dune exec dev/shim/coqtop-prelude` will build and execute this shim,
equivalent to doing `make states && bin/coqtop` under the old model.

This PR is just a bit of "a hack" until proper support for Coq
libraries arrives to Dune, however there is nothing wrong with it.

The long term plan (for now) is to have a command similar to `dune
utop $dir`, `dune coqtop $dir`, which would spawn a proper Coq shell
with the corresponding libraries on the path.

This will work for `dir=stdlib/Init/` for example, or for any other
combination.


SkySkimmer:
Won't this go in `_build/whatever/bin/coqtop` rather than `bin/coqtop`?

ejgallego:
> Won't this go in `_build/whatever/bin/coqtop` rather than `bin/coqtop`?

Indeed, in fact I see this as an advantage, don't you think? [tree is clean of autogenerated files, soon .merlin will be moved out of the way too]

SkySkimmer:
>Indeed, in fact I see this as an advantage, don't you think?

Having to enter longer paths is a pain IMO. Generally I am OK being impure in exchange for convenience in the build system.

>soon .merlin will be moved out of the way too]

How?

ejgallego:
> How?

Files will be placed in `_build` and `merlin` will know to look there.

gares:
> Having to enter longer paths is a pain IMO.

I could not agree more.

But this patch is a bit of a hack. Can't we just have the script tell us what to add to PATH?

gares:
(note that it is not trivial, since under _build I've many things I see)

Zimmi48:
I lack a bit of context to understand the reason for this patch. Is it a workaround to the `-coqlib` problem? It feels weird to call `coqtop` a script that actually calls `coqtop`.

SkySkimmer:
>But this patch is a bit of a hack. Can't we just have the script tell us what to add to PATH?

I don't really understand what you mean by that.

ejgallego:
I agree about the problems, I will update this PR ASAP and explain better the situation / try to address the comments.

ejgallego:
Updated, let's see how things look now.

gares:
Can the shim be in `bin/` so that if I build with make I can `bin/coqtop` and with dune `dune exec bin/coqtop`?

ejgallego:
> Can the shim be in `bin/` so that if I build with make I can `bin/coqtop` and with dune `dune exec bin/coqtop`?

Sure, I changed it as what @Zimmi48 suggested, note that as we putting the shim in the build tree we can even put it in the root tree and do `dune exec ct` for example.

I will implement `dune coqtop` soon tho.

dev/shim/dune
```diff
@@ -0,0 +1,11 @@
+(rule
+ (targets coqtop)
```
Zimmi48: This line seems to be in contradiction with the `(with-outputs-to coqtop-prelude` line. Moreover, I don't think you need this line in the presence of the other one.

dev/shim/dune
```diff
@@ -0,0 +1,11 @@
...
+ (deps
+  (:coqtop %{project_root}/topbin/coqtop_bin.exe)
+  (:prelude %{project_root}/theories/Init/Prelude.vo))
+ (action
+  (with-outputs-to coqtop-prelude
+   (progn
+    (echo "#!/usr/bin/env bash\n")
+    (bash "echo \"$(pwd)/%{coqtop} -coqlib $(pwd)/%{project_root}\"")
```
Zimmi48: It seems that additional arguments to `coqtop` such as `-R` won't be passed on to the real coqtop.

Zimmi48 (commented):
You may use `(mode promote-until-clean)` to please Gaëtan and Enrico and have the generated script be in the main tree. Also, given the clear name `coqtop-prelude`, I don't think it needs to be in sub-directory `shim`. `dev/coqtop-prelude` would look good to me.

dev/shim/dune
ejgallego: Fixed thanks!

dev/shim/dune
ejgallego: Oh indeed, good point, amended.

ejgallego:
What should we do with this?

It was meant more as a proof of concept, and I can understand the received reluctance. We should eventually find a better solution, but in my view requires declaring the stdlib as a Dune lib as this is still work in progress; there are some pitfalls in the current design of "recursive loadpaths" and we will need some time to adopt a model that we like.

On the other hand, Dune won't allow us to output to the build tree [I wish we could to drop `voboot` mais c'est la vie], so this is more or less what we can do for now, surely way more cleanly.

Another choice would be to turn this into an alias that could be called such as in `dune build @states` and that would launch coqtop.

Zimmi48:
>On the other hand, Dune won't allow us to output to the build tree [I wish we could to drop voboot mais c'est la vie], so this is more or less what we can do for now, surely way more cleanly.

Can you be more specific? Dune does allow outputting to the build tree with promote rules. I thought you were the one not liking this.

Zimmi48:
@ejgallego 
>Oh well promote files are supposed to be checked-in in the VCS

There is also `promote-until-clean` which is obviously not supposed to be checked-in.

SkySkimmer:
I dunno about the shim approach, I propose we close this PR (we have too many open PRs waiting already), wait until the ocaml/dune#1555 is done then reevaluate.

ejgallego:
Ok to me; this PR however has some documentation value IMO, as this is what `dune coqtop` will indeed do.

SkySkimmer:
??

ejgallego:
I am reopening this as it is the only way we have as of today to do a working "states" target, thus at least for me it is the only thing missing from the make setup.

> I dunno about the shim approach, I propose we close this PR (we have too many open PRs waiting already), wait until the [ocaml/dune#1555](https://github.com/ocaml/dune/pull/1555) is done then reevaluate.

That could take a while so no reason to delay this, which is pretty minimal and works OK.

We can add promote-until-clean as @Zimmi48 suggested.

ejgallego:
Ok, this is IMO ready and works; we can further tweak it; updated description to reflect that this also allows for a quick start of CoqIDE.

Makefile.dune
```diff
@@ -46,7 +46,7 @@ voboot:
 	dune exec ./tools/coq_dune.exe $(BUILD_CONTEXT)/.vfiles.d
 
 states: voboot
-	dune build $(DUNEOPT) theories/Init/Prelude.vo
+	dune build --display=short $(DUNEOPT) dev/shim/coqtop-prelude
```
Zimmi48: Is this `--display=short` on purpose or for debugging?

Zimmi48 (commented):
What is the point of `quickopt` / `quickbyte` compared to `states` after this PR?

I just tested this PR and it seems that it breaks `quickide` which builds the whole Coq package now.

Makefile.dune
ejgallego: A few devs told me in the last meeting that for states they'd like to get some more informative output by default; so I did this tweak; indeed it makes sense as a UI issue IMHO, but we can do as we want.

ejgallego:
> What is the point of `quickopt` / `quickbyte` compared to `states` after this PR?

I dunno, I always use `check` for byte and `states` for opt, I guess that's an usability thing for people migrating; we could maybe remove them. [In the long term for sure]

> I just tested this PR and it seems that it breaks `quickide` which builds the whole Coq package now.

Very strange, I don't understand what's going on here (adding `(package coqide-server)` pulls the whole of coq indeed), let's see what's going on here, it should not.

ejgallego:
So indeed I misunderstood what `(package ...)` does, should be fixed now, thanks @Zimmi48 for being as careful as always.

Makefile.dune
rgrinberg: This is configurable on a per user basis by the way. `$ man dune-config` for details. I'd encourage people to just set the output they like there because it's likely they're working on other OCaml projects as well.

Zimmi48:
There are CI failures that could be related. I didn't look at the details.

ejgallego:
Oh I likely used some dune 1.7 features indavertenly, indeed the versioning done by dune lang tends to be fairly imperfect.

ejgallego:
Umm, I cannot reproduce with dune 1.6.3, interesting.

ejgallego:
> Umm, I cannot reproduce with dune 1.6.2, interesting.

Ahh, this is the bootstrapping problem again. The error message dune does here is truly bad.

Resolving `%{bin:foo}` fails because `coq.plugins.ltac` is not yet in scope. Damn, I was hit by this problem before and now I am hit again by it.

ejgallego:
I think I've figured out a good way.

ejgallego:
Ok this should work reasonably well now.

Makefile.dune
ejgallego: User config is already documented in the Coq Dune readme. The choice here is in order to this particular legacy target "states" behaves more similarly for people migrating. This was a FAQ of `state` users: "I want it to display short mode".

Zimmi48 (approved):
Makes sense to me. I am just wondering if it shouldn't be added to the documentation of the Dune system (`dev/doc/build-system.dune.md`) that after running states one should run coqtop with `dune exec dev/shim/coqtop-prelude` or `_build/default/dev/shim/coqtop-prelude` but `dune exec coqtop` will still not work. Same for `dune exec dev/shim/coqide-prelude` vs `dune exec coqide`.

ejgallego:
> Makes sense to me. I am just wondering if it shouldn't be added to the documentation of the Dune system (`dev/doc/build-system.dune.md`) that after running states one should run coqtop with `dune exec dev/shim/coqtop-prelude` or `_build/default/dev/shim/coqtop-prelude` but `dune exec coqtop` will still not work. Same for `dune exec dev/shim/coqide-prelude` vs `dune exec coqide`.

Given that these targets are kind of "temporary", another choice is to add them as make targets, `make run-coqtop` would work, and would be discoverable in the make help.

Note that after the merge of #8690 , we can indeed have `coqtop` depend on the prelude, so that would make `dune exec coqtop` work! [At the cost of making coqtop's dep chain a bit more heavy]

So maybe that's the best choice?

ejgallego:
> dune exec coqtop

Well, actually not a lot of gain when compared with `dune exec coqtop-run`, so I think the tradeoff is not in favor of that solution.

Zimmi48:
Sorry, I didn't get what you proposed with `make run-coqtop` / `dune exec coqtop-run`.

ejgallego:
> Sorry, I didn't get what you proposed with `make run-coqtop` / `dune exec coqtop-run`.

No worries, I was just thinking out loud; IMO the best next step is to implement `dune coqtop` . We can live with the shims for now.

I've added some documentation. The whole Dune doc file could use a pass, but I'll do that in a separate PR once Dune's makefile becomes the default I think.

Zimmi48 (approved):

ejgallego:
Merci !

PR #9048, opened 2018-11-22
Fix vernac classification of `Fail Instance`

maximedenes:
AFAIK `Fail Instance` cannot open a goal, so this is probably a mistake by me when I wrote the `vernac_control` code.

### DIFF
--- a/stm/vernac_classifier.ml
+++ b/stm/vernac_classifier.ml
@@ -214,6 +214,6 @@ let classify_vernac e =
         | VtQed _, _ ->
             VtProofStep { parallel = `No; proof_block_detection = None },
             VtNow
-        | (VtStartProof _ | VtUnknown), _ -> VtUnknown, VtNow)
+        | (VtStartProof _ | VtUnknown), _ -> VtQuery, VtLater)
   in
   static_control_classifier e


SkySkimmer:
Why's Fail not always a VtQuery?

gares:
I think Gaetan is correct. Fail should always behave like a query (no trace left in the system state)

maximedenes:
@gares when classifying `Fail` as a query, I get more errors. You can open `test-suite/success/bteauto.v` in coqide to observe the problem. I probably need some help there.

maximedenes:
Ok, so it seems the STM makes a lot more assumptions on query (and in particular, distinguishing it from tactics) than just "it does not change the environment". This will require more work.

maximedenes:
@gares how do you think we should fix this? It seems the code is not consistent w.r.t. what it calls "queries".

gares:
We renounce to this

maximedenes:
> We renounce to this

Let's at least clarify what a query is, in the code. What's a query?

gares:
What a query should be is clear, so what this PR tries to do is morally OK. Unfortunately in the code the fact that something is labelled as a Query has other effects that, as we tried, makes Fail misbehave.

I consider that a sort of bug that requires a cleanup. Now I don't recall exactly what we found the other day, but it wound not be a documentation but rather an analysis of this bug report.

gares:
Another way to put it is that this PR should become a bugreport, or be rebased on top of the cleanup one that still does not exist.

gares (changes requested):
to signal I don't need to review this anymore and that it is not in a mergeable state

maximedenes:
@gares Could you list somewhere (in the code, around the classification I guess) the current expectations the STM has w.r.t. queries? i.e. what is a query, basically. At least, it would clarify the current status quo, and we could work on lifting some of these assumptions.

maximedenes:
@gares I reverted to the original version of the patch, which should be merged if CI passes, as it removes a code path generating `VtUnknown` (for no reason).

Of course, it would be better if we could implement @SkySkimmer 's suggestion, but it seems hard without understanding more the hidden invariants.

gares (approved):

gares:
I don't even compile it with quick (unless env variables are inherited)

maximedenes:
@gares ping

PR #9051, opened 2018-11-22
Camlp5 safe API strikes back

ppedrot:
Revival of the cleaning part of #8923 and further removal of all unsafe camlp5 exported API.

### DIFF
--- a/coqpp/coqpp_main.ml
+++ b/coqpp/coqpp_main.ml
@@ -139,7 +139,7 @@ let print_local fmt ext =
   match locals with
   | [] -> ()
   | e :: locals ->
-    let mk_e fmt e = fprintf fmt "%s.Entry.create \"%s\"" ext.gramext_name e in
+    let mk_e fmt e = fprintf fmt "Pcoq.Entry.create \"%s\"" e in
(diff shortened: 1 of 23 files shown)


gares:
Can you explain, high level, which are the unsafe API that gets removed? The patch does many things, eg `Gram.` disappears, but this is just distracting. I see a `delete_rule` gone. Is that it?

ppedrot:
The important commit is the last one, it removes the type-unsafe API from our grammar engine. Internally the latter still uses unsafe casts, but as long as we trust its implementation this means we statically ensure that no undefined behaviour could happen by merely parsing stuff.

ejgallego (approved):
Very good; note the need for overlays tho. Once these are ready I will merge.

ejgallego:
You can test `dev/tools/create-overlays.sh ppedrot 9051 equations ltac2` if you want @ppedrot ?

ppedrot:
That's very strange, I cannot reproduce the error locally. I rebased to see whether that was some quirk, but otherwise I guess something is wrong.

ppedrot:
OK, that was me being stupid. Should hopefully work now.

PR #9052, opened 2018-11-22
Add shims target to Makefile.dune

SkySkimmer:
Then while transitioning between make and dune there is no need to
change how one invokes Coq (for one off tests mostly for me).

This is what I meant by making some shims in bin/.


SkySkimmer:
See also #9047

Zimmi48:
Why not? At least, this avoids polluting the dune files with such hack. In any case, this should be documented as part of the developer documentation.

ejgallego (changes requested):
Unfortunately a strong invariant of the Dune-based setup I'd like to preserve is to consider the source tree a read-only. While this is not enforced now, it will soon be, so I am not eager to introduce something people will start depending on and that can go stale, and in particular may create problems with sandboxing. We can instead:

- make this file static;
- move this file into `_build`

Note that I will amend #9047 to correct many problems noted in review.

SkySkimmer:
>Unfortunately a strong invariant of the Dune-based setup I'd like to preserve is to consider the source tree a read-only. 

I don't share this interest.

>make this file static;

Impossible while we still have the normal makefile system

>move this file into _build

Too annoying to use.

ejgallego:
> Too annoying to use.

If the argument you have against `_build` is that it is too hard to define an emacs variable/macro that auto-detects the actual path, I am afraid I have problems considering it a serious problem. Manual input is annoying in general; either be 10 or 20 chars.

ejgallego:
> I don't share this interest.

In case you wonder I am not sure these kind of comments (like certain edits in ci.inria) are neither constructive or helpful. I certainly don't appreciate them, as they seem dismissive of goals and work that have a clear technical foundation.

A hygienic build is not "an interested", but a technical goal; a dirty tree creates problems, so please try to make technical argument when debating this kind of features.

gares:
> a dirty tree creates problems

I don't see why creating a directory `_build` is  clean, while creating a second one `bin/` is not. Is it the `_` that makes it clean? I'd be fine with `_bin/` too.

The point you fail to get @ejgallego is that (perfectly sound) technical choices made by the dune devs *do* have an impact on *our* workflow. Telling people to use emacs macros to work around that is very punning.
You had the same attitude the other day about the annoying behavior of coqbot w.r.t. needs:rebase. Someone, with no discussion that I can recall, made a choice that has an impact of the workflow of others. The answer was to make a macro. :-/

I don't believe that "a sound technical choice" is a sufficient justification to make someone's workflow painful. You have to show that there is no other "reasonably sound and less painful" way.

ejgallego:
> I don't see why creating a directory `_build` is clean, while creating a second one `bin/` is not. Is it the `_` that makes it clean? I'd be fine with `_bin/` too.

The build directory is indeed special and we could configure it to be `_bin` or whatever value we wanted; but indeed that is an opinionated choice, but it has a technical basis. Rules (will be soon) sandboxed so commands cannot write outside the build directory, usually `_build/default`.

> The point you fail to get @ejgallego is that (perfectly sound) technical choices made by the dune devs _do_ have an impact on _our_ workflow. Telling people to use emacs macros to work around that is very punning.

Indeed as I posted in the other thread Dune doesn't [and cannot] support every possible workflow. Sandboxing of builds is something that unfortunately cannot be relaxed so easily so I am afraid if I tell you we can support that I would be lying.

Maybe I have a different valuation, but changing from `bin/coqtop` to `_build/default/bin/coqtop` is a natural consequence of a hygienic build and I don't see a way around that. That is something that doesn't worry me too much, when I get this kind of request I feel they are very nit picky. For me is like complaining about having to call `ocamlfind opt` instead of `ocamlopt`.

> You had the same attitude the other day about the annoying behavior of coqbot w.r.t. needs:rebase. Someone, with no discussion that I can recall, made a choice that has an impact of the workflow of others. The answer was to make a macro. :-/

That belongs to a different topic, and I don't think your representation of my answer is fair; I intended to say that the current status quo seems better than the one we had before, at least from my point of view, of course it would be great if we could improve the bot to be smarter. I didn't do that change but I prefer the current tradeoff, that's all I said.

In this case, having a hygienic build may introduce some pain but it removes many other way larger pains we had. For example people having to run `git clean` every once in a while, etc...

> I don't believe that "a sound technical choice" is a sufficient justification to make someone's workflow painful. You have to show that there is no other "reasonably sound and less painful" way.

Changing some variable to point to `_build/default/foo` instead of `foo` is not painful to me; there are many ways to do that including defining some shell alias or whatever. It is way more painful having to spend dozen of hours in these discussions, when adding a shell alias `alias ct="dune exec dev/shim/coqtop-stdlib` does the job.

It is true that Dune imposes some workflow choices, and doesn't support many others. I like that, because in the end having less choice tends to provide a more consistent experience, but of course it is my view.

ejgallego:
Anyways I think I am pretty open to tweaking things / trying to make the experience better, but as of today I don't see a good way to break the sandbox :( For example a use case is package generation, `dune-release` will complain about a dirty tree, this makefile will make it dirty. You know Debian does also compare the tree against the orig.tar.gz, so indeed it is good we try to be respectful of the source tree == only modifications by the devs are allowed.

ejgallego:
The thing is that:
```
make -j 4 states && ./bin/coqtop
```
is not even less typing than:
```
dune exec bin/coqtop
```

SkySkimmer:
>For example a use case is package generation, dune-release will complain about a dirty tree, this makefile will make it dirty. You know Debian does also compare the tree against the orig.tar.gz, so indeed it is good we try to be respectful of the source tree == only modifications by the devs are allowed.

debian isn't going to run `make shims`. I don't know if there's a workflow where one runs dune-release in a dev tree but I imagine those who do would quickly see which files it complains about and would be free to just not use the shims.

>In this case, having a hygienic build may introduce some pain but it removes many other way larger pains we had. For example people having to run git clean every once in a while, etc...
>Changing some variable to point to _build/default/foo instead of foo is not painful to me

And running git clean isn't painful to me. The goal of this PR is to have a situation where we both have a non painful workflow.

Zimmi48:
On the unrelated bot discussion, see coq/bot#25 for context on the technical reasons that limited what the bot does today. But as mentioned there, there is actually a solution to fix this, it's just more complex.

SkySkimmer:
Needs further thought as mentioned in #9092.

PR #9054, opened 2018-11-22
[ci] [appveyor] Move Appveyor to OPAM 2.

ejgallego:
We update the Appveyor configuration so it uses OPAM 2.0, and thus it
can install newer packages.

### DIFF
--- a/dev/ci/appveyor.sh
+++ b/dev/ci/appveyor.sh
@@ -2,14 +2,15 @@
 
 set -e -x
 
-APPVEYOR_OPAM_SWITCH=4.07.0+mingw64c
+APPVEYOR_OPAM_VARIANT=ocaml-variants.4.07.1+mingw64c
 
-wget https://github.com/fdopen/opam-repository-mingw/releases/download/0.0.0.1/opam64.tar.xz
+wget https://github.com/fdopen/opam-repository-mingw/releases/download/0.0.0.2/opam64.tar.xz -O opam64.tar.xz
 tar -xf opam64.tar.xz
 bash opam64/install.sh
 
-opam init -a mingw https://github.com/fdopen/opam-repository-mingw.git --comp $APPVEYOR_OPAM_SWITCH --switch $APPVEYOR_OPAM_SWITCH
-eval "$(opam config env)"
+opam init default -a -y "https://github.com/fdopen/opam-repository-mingw.git#opam2" -c $APPVEYOR_OPAM_VARIANT --disable-sandboxing
+eval "$(opam env)"
 opam install -y num ocamlfind ounit
 
+# Full regular Coq Build
 cd "$APPVEYOR_BUILD_FOLDER" && ./configure -local && make && make byte && make -C test-suite all INTERACTIVE= # && make validate
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,8 +2,7 @@ version: '{branch}~{build}'
 clone_depth: 10
 
 cache:
-  - C:\cygwin64 -> dev\ci\appveyor.bat
-  - C:\cygwin64\home\appveyor\.opam -> dev\ci\appveyor.sh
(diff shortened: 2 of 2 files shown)


SkySkimmer:
Looks like it hung somehow. Doesn't look spurious.

ejgallego:
It turns out the problem is the cache; I don't have permissions in appveyor thus I'll have to workaround this.

dev/ci/appveyor.sh
```diff
@@ -2,14 +2,15 @@
...
+wget https://github.com/fdopen/opam-repository-mingw/releases/download/0.0.0.2/opam64.tar.xz -O opam64.tar.xz
 tar -xf opam64.tar.xz
 bash opam64/install.sh
 
-opam init -a mingw https://github.com/fdopen/opam-repository-mingw.git --comp $APPVEYOR_OPAM_SWITCH --switch $APPVEYOR_OPAM_SWITCH
-eval "$(opam config env)"
-opam install -y num ocamlfind ounit
+opam init default -a -y "https://github.com/fdopen/opam-repository-mingw.git#opam2" -c $APPVEYOR_OPAM_VARIANT --disable-sandboxing
+eval "$(opam env)"
+opam install -y num ocamlfind ounit dune
```
SkySkimmer: Why install dune here?

SkySkimmer:
Looks like it works now.

dev/ci/appveyor.sh
ejgallego: Oh that's a leftover of #8421 , from which this PR was spun.

ejgallego:
Let's see what happens with the cache again.

ejgallego:
So this is ready for merge as I works if the cache is not used; I think the best strategy is merge, then clean the caches.

ejgallego:
Working example [cache disabled]https://ci.appveyor.com/project/coq/coq/builds/20516158

maximedenes:
Oh! Am I still listed as an appveyor maintainer? Anyway, will take care of this one.

How should the appveyor be read?

ejgallego:
> Oh! Am I still listed as an appveyor maintainer? Anyway, will take care of this one.

Yes you are; feel free to move to me [tho this wouldn't help a lot as I am the one mainly caring] or to `ci-maintainers if @SkySkimmer agrees.

Note that a problem we have is that we are not admins in the appveyor instance so you should give us access.

> How should the appveyor be read?

What do you mean?

For this PR, there seems to be a problem with the cache that I cannot track, so you need to merge, then clean the caches and have this run I think.

maximedenes:
> For this PR, there seems to be a problem with the cache that I cannot track, so you need to merge, then clean the caches and have this run I think.

Ok, so you suggest I merge even if it failed? I'm ok with it, just asking to make sure I don't do a mistake.

ejgallego:
> Ok, so you suggest I merge even if it failed? I'm ok with it, just asking to make sure I don't do a mistake.

If you look a bit up you can see a build without cache that suceeded. I can't make this one to succeed as I don't have the proper permissions to clean the cache in appveyor.

So you need to merge, clean the caches, then rerun I think.

maximedenes (approved):

SkySkimmer:
This seems risky, unrebased branches still trigger appveyor so there will be cache resurrections for a while.
Shouldn't we make the `  - C:\cygwin64 -> dev\ci\appveyor.bat` cache depend on the appveyor.sh too?

ejgallego:
We already have:
```
  - C:\cygwin64\home\appveyor\.opam -> dev\ci\appveyor.sh
```
why is that not working?

SkySkimmer:
What's bash profile doing?

Gaëtan Gilbert

ejgallego:
This seems to be working now, failed due to overtime, rerunning again but if the worker is fast it should go thru.

Zimmi48:
Passed.

ejgallego:
Thanks!

PR #9055, opened 2018-11-22
[dev] fix create_overlay wrt branch names containing /

gares:


ejgallego (commented):
I actually have no clue about this Bash pattern. How is this supposed to work?

Shouldn't we instead escape `$OVERLAY_BRANCH` with ${///\/ ...}?

gares:
It replaces every occurrence of `/` with ``, that is it strips `/` that is otherwise allowed in git branch names. The syntax is horrible, I know.

I also have another minor point about the script: it uses "dangerous" names for the remotes.
It should use "origin" for the developer fork, and "upstream" for the upstream.
Currently it uses "gares" for my fork, and "origin" for the upstream.
The thing is that, by default, git pushes to "origin". If you have commit rights on the upstream repo you may push the branch there by mistake.

ejgallego:
> It replaces every occurrence of `/` with nothing, that is it strips `/` that is otherwise allowed in git branch names. The syntax is horrible, I know. The problem is that if the overlay file is called `123-foo/bar.sh` then the shell complains since `123-foo/` does not exists when you `>` to that path.

Wouldn't it better to replace `/` by `\/`, that is to say, to properly escape the branch name?

> The thing is that, by default, git pushes to "origin". If you have commit rights on the upstream repo you may push the branch there by mistake.

This is inherited from the CI scripts, so IMHO we should change it there. I was not aware that git pushes to `origin` by default tho, where is that specified?

I do propose however that we keep the branch name as is here, and indeed the CI scripts could checkout to upstream if that is convenient.

ejgallego:
So indeed we may want to pass `--origin upstream` to `git clone` in `dev/ci/ci-common.sh` , however I am not sure this won't set the default push branch to `upstream` so the problem would persist. I am not sure how to avoid that. Maybe we need to manually call `--set-upstream`?

I am fine with that change, @SkySkimmer @Zimmi48 does that look OK to you?

ejgallego:
Sorry, I didn't explain properly, what I meant is to make `1 = 2 = 3`, that is to say, escape the OVERLAY_BRANCH only for the filename.

gares:
You can't make 1 = 2 = 3 because 3 cannot contain / while 1 (and 2) can.
1 pr branch name (git)
2 pr overlay branch name (git)
3 overlay file name (file system)

ejgallego:
Oh yes of course you can't have / in a path unless you are in win / jfs :embarrassed:

Maybe lets substitute the slash with _ or + instead of a space ?

Regarding the remote name I agree itd be good to rename it but I am not sure I know the right git magic.

SkySkimmer:
> The thing is that, by default, git pushes to "origin". If you have commit rights on the upstream repo you may push the branch there by mistake.

We could do `git config remote.pushDefault $dev_name`, I think it would be enough for this issue.
If we want to be really sure that people won't push to origin we can do `git remote set-url --push origin no_push`.

gares:
+ is fine, just put it after the 3 / in the bash expansion

ejgallego (approved):

ejgallego:
> We could do `git config remote.pushDefault $dev_name`, I think it would be enough for this issue.
> If we want to be really sure that people won't push to origin we can do `git remote set-url --push origin no_push`.

Sounds good I think, should we make this work in the scope of the overlay script or in the main CI one?

Zimmi48:
git doesn't push to origin by default but many developers (I included) have set their `.gitconfig` to do so. I indeed try to clone most of the time with `-o upstream` and then setup origin to point to my fork.

PR #9064, opened 2018-11-24
[gramlib] Minor cleanups:

ejgallego:
- remove duplicate type definitions `gram_assoc`, `gram_position`,
- make global `warning_verbose` variable into a parameter,
- remove redundant `entry` type, make type-flow clearer.

### DIFF
--- a/coqpp/coqpp_main.ml
+++ b/coqpp/coqpp_main.ml
@@ -146,16 +146,16 @@ let print_local fmt ext =
     fprintf fmt "in@ "
 
 let print_position fmt pos = match pos with
-| First -> fprintf fmt "Extend.First"
-| Last -> fprintf fmt "Extend.Last"
-| Before s -> fprintf fmt "Extend.Before@ \"%s\"" s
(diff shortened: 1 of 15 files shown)


ppedrot (commented):
As explained in another PR, I am fine with this one, but the calls to grammar extension with the warn flag on are all oversights in the current implementation. Either we completely remove the warning system (and thus the flag) or we provide an API that doesn't print randomly on the terminal.

ejgallego:
Choices:
- make `gramlib` use `Feedback`
- pass the warning function as a parameter
- remove the warning [what was done in the other PR]

I am not expert enough to see what to do here; will do 2 for now.

ppedrot:
I think 2. is fine, but it should be done in an algebraic way, i.e. representing warnings using a structured type.

ejgallego:
> I think 2. is fine, but it should be done in an algebraic way, i.e. representing warnings using a structured type.

As opposed to the current `CWarnings` system?

Do prefer variants for warnings too tho.

ppedrot (approved):
Let's merge.

PR #9066, opened 2018-11-24
[parsing] Move pcoq-specific parts in extend to pcoq.

ejgallego:


herbelin:
> IMHO this is a step toward unifying Pcoq and Gramlib extension mechanisms.

If you feel like it, there is the backtracking mechanism to simplify. There are hacks around 'reinit' only to compensate that `grammar_delete` on a level with only one rule in it deletes not only the rule but also the level, which is not what we want.

ejgallego:
Ok, so I guess this is the case in `gramext.ml` `DeadEnd`. What should the correct thing to do to keep the updated level then?

ejgallego:
Overlay:
- https://github.com/ppedrot/ltac2/pull/89

ppedrot:
I don't think removing the well-typed `Extend` GADTs is a good thing, it is rather the opposite: they should be moved back to gramlib to remove calls to unsafe casts. Also, I have a few cleanup patches that make some changes here moot.

ejgallego:
> I don't think removing the well-typed `Extend` GADTs is a good thing, it is rather the opposite: they should be moved back to gramlib to remove calls to unsafe casts. Also, I have a few cleanup patches that make some changes here moot.

Note they are not removed, but placed into `Pcoq` as they are really Pcoq-specific and Pcoq doesn't see now the rest of `Extend`. So indeed that would be a step towards exposing them in `Gramlib` indeed.

So we can proceed as you wish.

herbelin:
> we should add a TODO list somewhere; indeed it'd be great to address that; I guess we need to sync with @ppedrot tho, I'm sure he has a lot of plans.

Yes and yes.

> you have an example of that level-deletion behavior happening?

This should be the test in `Notations.v` around lines 11-32.

ppedrot (commented):
I don't like this PR. Having a purely declarative interface for grammar extension is a net win, and furthermore Pcoq is already big enough not to be stuffed with this. If anything, I'd let the entry AST inside Extend, but move the remaining stuff to Gramlib.

ejgallego:
> Having a purely declarative interface for grammar extension is a net win

Sorry @ppedrot what does "purely declarative" means in this context, how does the patch makes extend not declarative?

> and furthermore Pcoq is already big enough not to be stuffed with this.

That's a good point, but we could refactor more later, tho I like this as is makes the API slightly more orthogonal == every user of extend had to import `Pcoq` anyways.

But yup, the idea is that the main consumer of these datatypes is `Pcoq` so it makes sense for it to own the type. As we have discussed a few times, `.mli` only files are very confusing from a documentation standpoint, as you cannot have an `odoc` chapter explaining the type properly as the consumer is in a different module.


> If anything, I'd let the entry AST inside Extend, but move the remaining stuff to Gramlib.

That's a possibility tho.

ejgallego:
> Having a purely declarative interface for grammar extension is a net win

Sorry @ppedrot what does "purely declarative" means in this context, how does the patch makes extend not declarative?

> and furthermore Pcoq is already big enough not to be stuffed with this.

That's a good point, but we could refactor more later, tho I like this as is makes the API slightly more orthogonal == every user of extend had to import `Pcoq` anyways.

But yup, the idea is that the main consumer of these datatypes is `Pcoq` so it makes sense for it to own the type. As we have discussed a few times, `.mli` only files are very confusing from a documentation standpoint, as you cannot have an `odoc` chapter explaining the type properly as the consumer is in a different module.


> If anything, I'd let the entry AST inside Extend, but move the remaining stuff to Gramlib.

That's a possibility tho.

ejgallego:
> Having a purely declarative interface for grammar extension is a net win

Sorry @ppedrot what does "purely declarative" means in this context, how does the patch makes extend not declarative?

> and furthermore Pcoq is already big enough not to be stuffed with this.

That's a good point, but we could refactor more later, tho I like this as is makes the API slightly more orthogonal == every user of extend had to import `Pcoq` anyways.

But yup, the idea is that the main consumer of these datatypes is `Pcoq` so it makes sense for it to own the type. As we have discussed a few times, `.mli` only files are very confusing from a documentation standpoint, as you cannot have an `odoc` chapter explaining the type properly as the consumer is in a different module.


> If anything, I'd let the entry AST inside Extend, but move the remaining stuff to Gramlib.

That's a possibility tho.

ejgallego:
> Having a purely declarative interface for grammar extension is a net win

Sorry @ppedrot what does "purely declarative" means in this context, how does the patch makes extend not declarative?

> and furthermore Pcoq is already big enough not to be stuffed with this.

That's a good point, but we could refactor more later, tho I like this as is makes the API slightly more orthogonal == every user of extend had to import `Pcoq` anyways.

But yup, the idea is that the main consumer of these datatypes is `Pcoq` so it makes sense for it to own the type. As we have discussed a few times, `.mli` only files are very confusing from a documentation standpoint, as you cannot have an `odoc` chapter explaining the type properly as the consumer is in a different module.


> If anything, I'd let the entry AST inside Extend, but move the remaining stuff to Gramlib.

That's a possibility tho.

ejgallego:
> Having a purely declarative interface for grammar extension is a net win

Sorry @ppedrot what does "purely declarative" means in this context, how does the patch makes extend not declarative?

> and furthermore Pcoq is already big enough not to be stuffed with this.

That's a good point, but we could refactor more later, tho I like this as is makes the API slightly more orthogonal == every user of extend had to import `Pcoq` anyways.

But yup, the idea is that the main consumer of these datatypes is `Pcoq` so it makes sense for it to own the type. As we have discussed a few times, `.mli` only files are very confusing from a documentation standpoint, as you cannot have an `odoc` chapter explaining the type properly as the consumer is in a different module.


> If anything, I'd let the entry AST inside Extend, but move the remaining stuff to Gramlib.

That's a possibility tho.

ppedrot:
> Sorry @ppedrot what does "purely declarative" means in this context, how does the patch makes extend not declarative?

A mli-only file, but we don't have those anymore, so that would be a 'ml-only' file. At any rate, something that does not duplicate a quite big ADT definition.

> as the consumer is in a different module.

But it's a free structure, there is no point in putting it inside another module! Your mind has been contaminated by OOP!

ejgallego:
> A mli-only file, but we don't have those anymore, so that would be a 'ml-only' file. At any rate, something that does not duplicate a quite big ADT definition.

Dune can support `mli`-only files just fine, it was the old make system that struggled with it, in particular when we added the `API` module based on zillions of aliases.

You know we can use the trick of having a private `foo_intf.mli` file and then importing that in `Pcoq` for example?

> But it's a free structure, there is no point in putting it inside another module! Your mind has been contaminated by OOP!

I guess Haskell and OCaml defining both `'a list` and `fold_left` in the same module are OOP then XDDDD

My motivation for this really is more to do with documentation / organization of the API / merlin discoverability tho, so I dunno, it will always be subjective.

ejgallego:
OK @ppedrot I've updated the PR which is now quite trivial.

Assuming we'd like to merge Gramlib's extension language with Pcoq's extension language, I still think that moving the datatype to Pcoq could have helped, but indeed we didn't gain a lot there tho, in fact the idea was to make the extend type private.

What would be the strategy to do such a move? Gramlib's extend vs Coq's extend do indeed use a different style, so that should be decided first.

ejgallego:
I've pushed a further cleanup, in particular the casts are not needed. That alone would compensate the original patch.

However, after revisiting this I'm more convinced that my original approach was sound.

In https://github.com/ejgallego/coq/commit/d84b6cab5af195b67c7c2f6812b8c3bbe90d0c8a I made the extend type private and indeed, except for some hackery in LTAC2, this seems like the way we would like to go for now, IMHO.

ejgallego:
In fact with the original approach there is no need to perform surgery in `Grammar` as to get the type out of the way, so indeed that indicates that the original intention on scoping of this PR was on the right path, I have a tree where I have removed this type entirely.

ejgallego:
Ping @ppedrot

ppedrot (approved):
LGTM. Why did we need the `CastedRNo` / `CasterRMay` in the first place?

parsing/extend.ml
```diff
@@ -79,8 +79,10 @@ type ('a,'b,'c) ty_user_symbol =
 
 (** {5 Type-safe grammar extension} *)
 
-type norec = NoRec    (* just two *)
-type mayrec = MayRec  (* incompatible types *)
+(* Should be merged with gramlib's implementation *)
+
+type norec = Gramlib.Grammar.ty_norec
+type mayrec = Gramlib.Grammar.ty_mayrec
```
ppedrot: Note that an alternative way would be to declare the Gramlib phantom types as open variants, it's a common design pattern. Those types are structural, so that we could declare the same here without ever referencing Gramlib.

ejgallego:
Thanks for the review @ppedrot.

> LGTM. Why did we need the `CastedRNo` / `CasterRMay` in the first place?

I think this is a leftover from the time we didn't have the typed api in gramlib, however it could have been the case that the only reason you needed the dependent cast was due to `norec/rec`.

> Those types are structural, so that we could declare the same here without ever referencing Gramlib.

Yup, that could work, not sure it is worth the hassle tho. I suggest we continue design discussion in the follow up PR to this one, which is _way more_ invasive and will force us to make some design choices.

ppedrot:
OK, let's merge in the meantime.

PR #9067, opened 2018-11-24
[gramlib] [pcoq] Don't delete levels on rule removal.

ejgallego:
As requested by @herbelin, this is an experiment to keep the level
when removing the last rule.

This should avoid the need to workaround the level deletion on rule
removal by re-creating it.

The changes in gramext are likely wrong as we may want to do some extra operation.

I don't still understand the error in the test suite [see below] but at least this is not a catastrophic failure so far:
```
File "./success/Notations.v", line 21, characters 0-58:
Error:
Level 8 is already declared right associative while it is now expected to be left associative.
```
Note that level 8 is one of the levels that where destroyed originally.

### DIFF
--- a/parsing/pcoq.ml
+++ b/parsing/pcoq.ml
@@ -85,12 +85,8 @@ end
     In [single_extend_statement], first two parameters are name and
     assoc iff a level is created *)
 
-(** Type of reinitialization data *)
-type gram_reinit = Gramlib.Gramext.g_assoc * Gramlib.Gramext.position
-
 type extend_rule =
 | ExtendRule : 'a Entry.t * 'a extend_statement -> extend_rule
-| ExtendRuleReinit : 'a Entry.t * gram_reinit * 'a extend_statement -> extend_rule
 
 module EntryCommand = Dyn.Make ()
 module EntryData = struct type _ t = Ex : 'b Entry.t String.Map.t -> ('a * 'b) t end
@@ -108,25 +104,13 @@ let camlp5_state = ref []
 let camlp5_entries = ref EntryDataMap.empty
 
 (** Deletion *)
-
 let grammar_delete e { pos; data } =
   List.iter
     (fun (n,ass,lev) ->
       List.iter (fun pil -> safe_delete_rule e pil) (List.rev lev))
     (List.rev data)
 
-let grammar_delete_reinit e reinit ({ pos; data } as d)=
-  grammar_delete e d;
-  let a, ext = reinit in
-  let lev = match pos with
-    | Some (Gramext.Level n) -> n
-    | _ -> assert false
-  in
-  let ext = { pos = Some ext; data = [Some lev,Some a,[]] } in
-  safe_extend e ext
-
 (** Extension *)
-
 let grammar_extend e ext =
   let undo () = grammar_delete e ext in
   let redo () = safe_extend e ext in
@@ -135,10 +119,6 @@ let grammar_extend e ext =
 
 let grammar_extend_sync e ext =
   camlp5_state := ByGrammar (ExtendRule (e, ext)) :: !camlp5_state;
-  safe_extend e ext
-
-let grammar_extend_sync_reinit e reinit ext =
-  camlp5_state := ByGrammar (ExtendRuleReinit (e, reinit, ext)) :: !camlp5_state;
   safe_extend e ext
 
 (** Remove extensions
@@ -150,10 +130,6 @@ let rec remove_grammars n =
   if n>0 then
     match !camlp5_state with
        | [] -> anomaly ~label:"Pcoq.remove_grammars" (Pp.str "too many rules to remove.")
-       | ByGrammar (ExtendRuleReinit (g, reinit, ext)) :: t ->
-           grammar_delete_reinit g reinit ext;
-           camlp5_state := t;
-           remove_grammars (n-1)
        | ByGrammar (ExtendRule (g, ext)) :: t ->
            grammar_delete g ext;
            camlp5_state := t;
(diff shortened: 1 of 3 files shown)


ejgallego:
So this actually fails on:
```coq
Notation "x + y" := (Add x y) (in custom expr at level 2, left associativity).
```
with
```
Error:
Level 2 is already declared right associative while it is now expected to be left associative.
```

gramlib/gramext.ml
```diff
@@ -469,55 +469,65 @@ and decr_keyw_use_in_tree gram =
...
+            let lev = { lev with assoc = NonA; lsuffix = t } in
+            lev :: levs
           | _ ->
-              let lev =
-                {assoc = lev.assoc; lname = lev.lname; lsuffix = t;
-                 lprefix = lev.lprefix}
-              in
-              lev :: levs
-          end
+            let lev =
```
herbelin: I suspect that you should remove this whole `DeadEnd` case. The bug in the test-suite is probably because `NonA` is used instead of `lev.assoc` (as in the non-`DeadEnd` case).

Idem in `delete_rule_in_prefix`.

herbelin (commented):
Very nice, thanks!

herbelin:
> So this actually fails on:
>
> Notation "x + y" := (Add x y) (in custom expr at level 2, left associativity).

I think I see. Coq was checking the consistency of associativity when adding a new rule in a level. This check was deactivated when the level was empty since no associativity is de facto forced yet.

Gramlib also does this check (in `Gramext.change_lev`), though with a warning and not an error. So, we could do the check only in Gramlib and make it an error rather than a warning.

However, the treatment of associativity is another place where Coq model slightly differs from Camlp5 model. I.e. gramlib itself should check deactivate the check if the level is empty.

Besides this problem, we should be careful about the symmetry between adding and deleting. In particular, we need sometimes to delete a level as the reverse operation of adding a level with no rule (as a result e.g. of `register_empty_levels`, as in the example `Notation "# x" := (S x) (at level 0, x at level 23).` where `23` is new and must be registered with no rules).

Previously, removing the last rule was always removing the level if the rule was unique. Thus the need to compensate that by calling again `G.safe_extend ` in `grammar_delete` which your PR allows to remove. When it is not the addition of a rule but the addition of a level which has to be reverted, `grammar_delete` has to do the appropriate revert action.

I don't know how to implement this. We could adopt the convention that when `rls` is empty in `grammar_delete` this means removing the level rather than removing the rule. But as a counterpart, this means that whenwe  create rules and a level in one go (in `Egramcoq.extend_constr`), we now need to split it into two `ExtendRule` (one with no rules to mean create the level and one with rules to mean add rules). Or we keep a flag in the `ExtendRule` to mean that it also creates, so that `grammar_delete` knows if it has to delete the level or not.

In short:
- There is a diversity of design spaces, depending on how we make the Coq and gramlib model closer or not. But, a minima, gramlib should provide a way to remove rules without necessarily removing the level when the level happens to be empty, so that the symmetry add/delete is ensured.
- For the current error in the test suite, either it is Coq who does the job of checking consistency of associativity but the `reinit` boolean should then be kept, or it is gramlib which does it, but gramlib should now raise an error rather than a warning, and only when the level is not empty.

ejgallego:
Anyways I'll wait for _le chef de chantier_ @ppedrot to outline his plans wrt gramlib before moving more.

ppedrot:
My short-terms plans are twofold: first, removing as much cruft as possible from camlp5, and second, providing a type-safe implementation so that we can more easily fiddle with it. In a second step, I would like to move the code towards a more functional implementation, not necessarily on the parsing API itself  (i.e. imperative streams) but at least the extension one (i.e. explicit state-passing when modifying the lexer and parser).

ppedrot:
Answering to myself: this is a actually dead code, I'll update #9069 accordingly.

ppedrot (commented):
I forgot to mention I am leaving this PR and its appreciation to @herbelin.

ppedrot:
@ejgallego shouldn't be that merely rebased since the merge of the depending PR?

ejgallego:
> @ejgallego shouldn't be that merely rebased since the merge of the depending PR?

I have not even yet submitted the depending PR as the depending PR depended on yet another submitted PR.

ppedrot:
@ejgallego status of this?

ejgallego:
> @ejgallego status of this?

Actually this PR depends on another which is almost ready, I have a single error due to some GADT magic; the PR on which this depends itself is pretty near in my TODO list so I'm hopeful this won't be blocked for long.

ejgallego:
Status is: couple of depending PRs submitted, once these are merged I'll give it a try to the solution @herbelin proposed. In particular a lot of duplicity has been removed by #11647 so this should be less painful to implement.

JasonGross:
Why does it matter what levels exist and what levels don't exist?  Said another way, why can't I model the initial state as if all levels exist but have no rules?

ppedrot:
@JasonGross It's important because existing levels have an implicit associativity and relative ordering.

Rediscovering this interesting PR. I think we should proceed in several steps, first splitting entry level creation / deletion from the mere rule definition and adding them at the relevant place. The API for rule insertion is already completely crazy because we take redundant information (position vs. level) and may output some warnings when there is a discrepancy. Instead we ought to first define levels and second populate them, and dually for deletion.

Then we can start pondering if this removal makes sense in Gallina / Ltac / Ltac2 notations and tweak the code accordingly, but this requires more work.

ejgallego:
> Why does it matter what levels exist and what levels don't exist? Said another way, why can't I model the initial state as if all levels exist but have no rules?

@JasonGross IIUC this is because levels do contain information such as associativity, so this "not initialized state" is modeled now as "the level doesn't exist".

ppedrot:
I believe this PR lost its important content during some rebase, since it does not touch gramlib at all. I'll try to come up with an alternative PR on a recent master which follows the plan I explained above.

ejgallego:
> The "needs: rebase" label was set more than 30 days ago. If the PR is not rebased in 30 days, it will be automatically closed.

Right now this PR has no value code-wise, however the discussion has very important information [thanks Hugo]

Ideally we would turn this into an issue, but I understand that's not possible yet, so I dunno how to proceed.

Zimmi48:
With current GitHub features, what you can do is to create an issue from any comment in the PR (using the menu). Then, it's up to you to summarize or refer to other comments in the issue body.

Issue #9008, opened 2018-11-15
About strictly enforcing parsing levels in camlp5/gramlib


herbelin:
Camlp5 has a "recovery" mode which bypasses level information when an expression cannot be parsed at the expected level.

One instance of it was shown in report #8990. Typical working instances in the large are of the form:
```
Check forall n, n > 1 /\ forall p, p >= n.
Check forall b, (if b then 0 else 1) = if b then 1 else 0.
Check forall n, n > 1 /\ forall p, p >= n /\ p <= n+1. 
```
where `/\` (resp. `=`) is waiting for a right-hand side at level 80 (resp. 70) but the `forall` and `if` are at a higher level, namely 200. Note that for the third example, the ambiguity is resolved into `forall n, (n > 1 /\ (forall p, (p >= n /\ p <= n+1)))`.

Besides changing priorities in sometimes unexpected ways (as in #8990), one of the consequence of this tolerance is that it is difficult to simulate the behavior of parsing in the printer so that they are inverse of each other.

Changing the default is possible but would introduce a rather important incompatibility (for instance, probably a couple of dozens of instances in the standard library).

An alternative would be to accept the fact and to make official than any notation with an open right-hand side (as `/\` above), tolerates the right-hand side to be at level 200. (I suspect that using a subparser `[ c = SELF -> c | c = binder_constr -> c]` on the right would work.) Then, the printer could be made aware of this convention so that they can be kept consistent.

Any opinion on going into one or the other direction?

ejgallego:
cc: @coq/parsing-maintainers

maximedenes:
Maybe @ggonthier has an opinion on this?

maximedenes:
@herbelin do you have a branch that implements stricter levels? It would be nice to see the impact outside stdlib (e.g. on the whole CI).

herbelin:
Sorry @maximedenes, I'm seeing your message only now. I pushed the corresponding experimental branch [here](https://github.com/herbelin/github-coq/pull/new/master+camlp5-mid-strict-mode).

SkySkimmer:
I don't see a level-tolerance warning with the given examples, is that expected?

PR #9069, opened 2018-11-25
[gramlib] Remove dead code

ppedrot:
This removes dead code from gramlib, that is sometimes non-trivially dead. Notably, some dynamic checks are made useless by the use of the functorial API, and some dynamic flags are useless, e.g. `warning_verbose` that should always be turned off in Coq (non-wrapped uses are oversights).

### DIFF
--- a/gramlib/gramext.ml
+++ b/gramlib/gramext.ml
@@ -284,39 +284,6 @@ let get_level ~warning entry position levs =
         lev :: levs -> [], change_lev ~warning lev "<top>", levs
       | [] -> [], empty_lev, []
 
-let rec check_gram entry =
-  function
-    Snterm e ->
-      if e.egram != entry.egram then
-        begin
-          eprintf "\
-Error: entries \"%s\" and \"%s\" do not belong to the same grammar.\n"
-            entry.ename e.ename;
-          flush stderr;
-          failwith "Grammar.extend error"
-        end
-  | Snterml (e, _) ->
-      if e.egram != entry.egram then
-        begin
-          eprintf "\
-Error: entries \"%s\" and \"%s\" do not belong to the same grammar.\n"
-            entry.ename e.ename;
-          flush stderr;
-          failwith "Grammar.extend error"
-        end
-  | Slist0sep (s, t, _) -> check_gram entry t; check_gram entry s
-  | Slist1sep (s, t, _) -> check_gram entry t; check_gram entry s
-  | Slist0 s -> check_gram entry s
-  | Slist1 s -> check_gram entry s
-  | Sopt s -> check_gram entry s
-  | Stree t -> tree_check_gram entry t
-  | Snext | Sself | Stoken _ -> ()
-and tree_check_gram entry =
-  function
-    Node {node = n; brother = bro; son = son} ->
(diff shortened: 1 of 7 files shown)


parsing/pcoq.ml
```diff
@@ -297,14 +289,14 @@ let grammar_extend e reinit ext =
   let ext = of_coq_extend_statement ext in
   let undo () = grammar_delete e reinit ext in
   let ext = fix_extend_statement ext in
-  let redo () = camlp5_verbosity false (uncurry (G.safe_extend e)) ext in
+  let redo () = uncurry (G.safe_extend e) ext in
```
ejgallego: Remove the `uncurry` here as done in #9064

ejgallego (commented):
This could be going a bit far in some removals, @herbelin are these modifications OK to you?

herbelin:
Using the type safe API to remove dynamic check is excellent.

On the other side, the  removal of `warning_verbose` can indeed be discussed. If not camlp5, we shall have to check that Coq is raising the warning/error we want to have. What was the motivation?

ppedrot:
> On the other side, the removal of warning_verbose can indeed be discussed

I am also fine with providing some cleaner API for the warning. AFAIU, modulo what looks like oversights like in `Pcoq.eoi_entry`, we tried to deactivate it in all Coq calls because it would print the warning on stderr. I think that if we provide a callback function to print the warning in a reasonable way, we're also good.

ejgallego:
> I am also fine with providing some cleaner API for the warning. AFAIU, modulo what looks like oversights like in `Pcoq.eoi_entry`, we tried to deactivate it in all Coq calls because it would print the warning on stderr. I think that if we provide a callback function to print the warning in a reasonable way, we're also good.

Have you seen what is done in #9064

ppedrot:
> Have you seen what is done in #9064

Yes, but it does not solve the problem, the warning is still printed on stderr. If anything, I'd rather *always* print the warning, but let the client decide how to do so.

ejgallego:
> Yes, but it does not solve the problem, the warning is still printed on stderr. If anything, I'd rather _always_ print the warning, but let the client decide how to do so.

That's easy as to amend the type of `~warning`.

ppedrot:
> That's easy as to amend the type of ~warning.

Indeed, but we should also reify warnings as an algebraic datatype, IIRC there are essentially three cases. And not sure if doable or worthwhile, but we could also return them functionally not to commit to a imperative interface.

ppedrot:
In order not to delay this otherwise pretty straightforward PR, I removed the contentious commit from it. We will be able to discuss the proper solution in #9064.

ejgallego:
I am OK with this but letting @herbelin take charge here as he knows for sure better than me what the needs of the grammar writer are.

ppedrot:
@herbelin This should be ready.

ppedrot:
@herbelin ping.

herbelin:
@ppedrot: Sorry, I've been sollicitated for non-Coq business in the last days. Assigning.

Remark in passing: I have work in progress on the `strict_parsing` mode (see #9008). So, let's remove it at the current time but I'll probably have to put it back again to eventually move towards a stricter parsing mode (i.e. a mode more compliant to precedences).

Will merge tomorrow.

PR #6795, opened 2018-02-20
[ssreflect] Export parsing witnesses in mli file.

ejgallego:
This is needed in order to manipulate/serialize SSR's AST.

A quicker [and maybe better] alternative would be to just remove
`ssrparser.mli`, as there are many grammar entries that still need
exporting.

Depends on #6700

PR #9070, opened 2018-11-26
[ssreflect] Export more parsing witnesses.

ejgallego:
This is needed in order to serialize ssreflect programs properly,
similar to #6795.

### DIFF
--- a/plugins/ssr/ssrparser.mlg
+++ b/plugins/ssr/ssrparser.mlg
@@ -576,6 +576,8 @@ END
 
 {
 
+type ssrfwdview = ast_closure_term list
+
 let pr_ssrfwdview _ _ _ = pr_view2
 
 }
@@ -637,6 +639,7 @@ let rec map_ipat map_id map_ssrhyp map_ast_closure_term = function
   | IPatView (clr,v) -> IPatView (clr,List.map map_ast_closure_term v)
   | IPatTac _ -> assert false (*internal usage only *)
 
+type ssripatrep = ssripat
 let wit_ssripatrep = add_genarg "ssripatrep" pr_ipat
 
 let pr_ssripat _ _ _ = pr_ipat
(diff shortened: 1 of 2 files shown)


gares (approved):

gares:
I'm  not sure the milestone is appropriate

ejgallego:
Thanks @gares, see #9071 , the truth is that I am not sure how to handle 8.9 requests

gares:
My take is:
- an existing, released, project needs an extra API on the stable release: we add it
- a promising, clever, cool, experiment needs an extra API: we add it in on master only

According to my understanding, the extra thing is not needed for a stable, released, project.
Am I wrong?

maximedenes:
> My take is:
> 
> * an existing, released, project needs an extra API on the stable release: we add it
> * a promising, clever, cool, experiment needs an extra API: we add it in on master only
> 
> According to my understanding, the extra thing is not needed for a stable, released, project.
> Am I wrong?

I think this is sound reasoning when *changing* an API. When *extending* it, we may want to be more flexible.

In this specific case, I was the one to write those `.mli` files, so I can assure you that not a lot of though went into the design. I basically sent 5 minutes adding what was strictly needed for the few things I wanted to compile. I wouldn't be shocked that we expose a bit more, even for a minor release.

ejgallego:
Well, what is going wrong here in my opinion is that we are leaving aside technical merit to judge the PR. Would we stick to technical merit there is no question, as @maximedenes pointed out that the current `.mli` file is fully incorrect and this fix should bear no discussion. The next measuring criteria should be what is the impact on users, and in this case it is very important.

However, the reply we get does a lot of things that I consider are not very efficient:
- first, it creates some new rules that we have never discussed,
- moreover, the set of rules is very questionable in itself [wouldn't we like to encourage, cool, new stuff],
- and even worse IMVHO is that the rules require us to pass judgement on projects, and determine if they are cool, promising, etc... which is obviously a trap.

PR #9071, opened 2018-11-26
[v8.9] [ssreflect] Export more parsing witnesses.

ejgallego:
This is needed in order to serialize ssreflect programs properly,
similar to #6795.

Backport of #9070 , this should be OK even for a minor release; it allows 8.9 SerAPI users to serialize more ssr scripts.

### DIFF
--- a/plugins/ssr/ssrast.mli
+++ b/plugins/ssr/ssrast.mli
@@ -140,6 +140,9 @@ type 'tac ssrhint = bool * 'tac option list
 type 'tac fwdbinders =
         bool * (ssrhpats * ((ssrfwdfmt * ast_closure_term) * 'tac ssrhint))
 
+type 'tac ffwbinders =
+  (ssrhpats * ((ssrfwdfmt * ast_closure_term) * 'tac ssrhint))
+
 type clause =
   (ssrclear * ((ssrhyp_or_id * string) *
               Ssrmatching_plugin.Ssrmatching.cpattern option) option)
--- a/plugins/ssr/ssrparser.ml4
+++ b/plugins/ssr/ssrparser.ml4
@@ -342,7 +342,7 @@ let interp_index ist gl idx =
 
 open Pltac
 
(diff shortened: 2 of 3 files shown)


gares (approved):

maximedenes (approved):

plugins/ssr/ssrparser.ml4
```diff
@@ -342,7 +342,7 @@ let interp_index ist gl idx =
 
 open Pltac
 
-ARGUMENT EXTEND ssrindex TYPED AS ssrindex PRINTED BY pr_ssrindex
+ARGUMENT EXTEND ssrindex PRINTED BY pr_ssrindex
```
gares: Why is this needed? I did not notice this before. The semantics is not the same if ssrindex has non trivial functions attached to it.

ejgallego: This is needed as to remove a warning "ignored TYPED AS" annotation.

ejgallego: Maybe the right thing to do is to remove the `mli` file? I dunno, in 8.10 certainly things are different now.

gares: ok then.

plugins/ssr/ssrast.mli
```diff
@@ -140,6 +140,9 @@ type 'tac ssrhint = bool * 'tac option list
 type 'tac fwdbinders =
         bool * (ssrhpats * ((ssrfwdfmt * ast_closure_term) * 'tac ssrhint))
 
+type 'tac ffwbinders =
+  (ssrhpats * ((ssrfwdfmt * ast_closure_term) * 'tac ssrhint))
```
SkySkimmer: Doesn't this factor with the above `fwbinders` as `type 'a fwdbinders = bool * 'tac ffwbinders`?

Zimmi48:
8.9.1 has been tagged without this PR

PR #9072, opened 2018-11-26
Clean stm flags

maximedenes:
This PR is a small clean-up on user-visible STM flags:
- it removes the `-async-proofs-full` flag
- it makes `async-proofs on` have an effect on `coqc`

The second part was probably a bug, fixing it will making it possible to test proof delegation automatically. The first flag had an unclear semantics, tweaking many unrelated STM options. We keep some infrastructure in place (like async queries), but hope to enable them in the default `-async-proofs on` mode, rather than a separate flag. If we fail to do so, we will provide an `-async-proofs full`, which will be cleaner than what we have today, although still not very clear.

### DIFF
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -44,7 +44,6 @@ module AsyncOpts = struct
     async_proofs_mode : async_proofs;
 
     async_proofs_private_flags : string option;
-    async_proofs_full : bool;
     async_proofs_never_reopen_branch : bool;
 
     async_proofs_tac_error_resilience : tac_error_filter;
@@ -61,7 +60,6 @@ module AsyncOpts = struct
     async_proofs_mode = APoff;
 
     async_proofs_private_flags = None;
-    async_proofs_full = false;
     async_proofs_never_reopen_branch = false;
 
     async_proofs_tac_error_resilience = `Only [ "curly" ];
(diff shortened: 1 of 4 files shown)


ejgallego (commented):
toplevel parts are OK, I leave the rest to @gares; I am not sure if this is a good idea as the full lazy mode used to have some problems?

Also, the `~check` change seems like a big one, in particular maybe you should insert a call to `Stm.finish` otherwise the file is not forced? But I guess that's OK for `master`.

stm/stm.ml
```diff
@@ -1442,7 +1440,6 @@ end = struct (* {{{ *)
   let task_match age t =
     match age, t with
     | Fresh, BuildProof { t_states } ->
-        not !cur_opt.async_proofs_full ||
```
gares: This Inverts the normal logic

gares (changes requested):

stm/stm.ml
maximedenes: Well spotted! Actually, I had fixed it but forgot to push.

maximedenes:
@gares should be good now, thanks!

stm/stm.ml
gares: I think I'd like a trace about the perspective to stay in the code.
Maybe define a "is_inside_perspective" that ccontains (commented out) the code below and is constant to true.

toplevel/ccompile.ml
```diff
@@ -119,7 +120,9 @@ let compile opts ~echo ~f_in ~f_out =
       Dumpglob.start_dump_glob ~vfile:long_f_dot_v ~vofile:long_f_dot_vo;
       Dumpglob.dump_string ("F" ^ Names.DirPath.to_string ldir ^ "\n");
       let wall_clock1 = Unix.gettimeofday () in
-      let state = Vernac.load_vernac ~echo ~check:true ~interactive:false ~state long_f_dot_v in
+      let open Stm.AsyncOpts in
+      let state = Vernac.load_vernac ~echo
+          ~check:(stm_options.async_proofs_mode = APoff) ~interactive:false ~state long_f_dot_v in
```
ejgallego: This is duplicated with the above and makes lines very long; please factor the check parameter to a `let check = ....` instead.

maximedenes: The other change is a mistake. I'll remove it, so no duplication. Will define `check` for this one, thanks!

maximedenes: Fixed, thanks!

ejgallego (approved):
Toplevel part is fine; under the assumptions that `Stm.join` will correctly finalize everything.

gares (approved):

gares:
Failure in formal-topology is spurious, isn't it ?

maximedenes:
> Failure in formal-topology is spurious, isn't it ?

It seems so.

PR #9073, opened 2018-11-26
[sphinx] No more undocumented objects.

### LINKED ISSUE #7602: Plan to get rid of objects without body

Zimmi48:
**Kind:** documentation fix

Fixes / closes #7602 

@cpitclaudel The change in the `coqrst` code remove the requirement for "documentation" for `prodn` objects.

@gares The SSReflect chapter is always the hardest to fix to make it closer to Sphinx requirements because it is generally written in a style that's not the expected one for this manual. I did what I could to reformulate in a way that is closer to what is expected (please check) and when I really couldn't without changing too much, I just put an `:undocumented:` option instead, but longer term these options should go away.

### DIFF
--- a/doc/sphinx/conf.py
+++ b/doc/sphinx/conf.py
@@ -61,7 +61,7 @@ extensions = [
 
 # Change this to "info" or "warning" to get notifications about undocumented Coq
 # objects (objects with no contents).
-report_undocumented_coq_objects = None
+report_undocumented_coq_objects = "warning"
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/doc/sphinx/proof-engine/ssreflect-proof-language.rst
+++ b/doc/sphinx/proof-engine/ssreflect-proof-language.rst
@@ -418,30 +418,29 @@ each point of use, e.g., the above definition can be written:
      Prenex Implicits null.
      Definition all_null (s : list T) := all null s.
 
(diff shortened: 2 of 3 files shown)


cpitclaudel (approved):
LGTM, though I'm not qualified wrt the SSReflect chapter

Zimmi48:
Let's give a few days to @gares to react before merging. @amahboubi or any member of @coq/ssreflect-maintainers could be alternative reviewers.

Zimmi48:
No comment after more than two weeks. Let's not delay this merge further. I'll rebase first just to be sure that no new warning has been introduced in the meantime.

Zimmi48:
Good. No new warning. The CI failure is irrelevant: math-comp is currently broken.

cpitclaudel:
Thanks!

Issue #9029, opened 2018-11-19
small manual proof takes 1min to Defined and 3min to print


andres-erbsen:
#### Version

~~~~~
The Coq Proof Assistant, version 8.8.2 (October 2018)
compiled on Oct 1 2018 0:22:50 with OCaml 4.07.0
~~~~~

#### Operating system

Arch Linux, up to date as of yesterday.

#### Description of the problem

I made a (relatively) self-contained example of a slow proof that I hope will be instructive for hunting down performance issues that affect bedrock2 and fiat-crypto (and perhaps other projects in the academic geneology of Adam Chlipala and @JasonGross).

The statement of the lemma applies a verification condition generator fixpoint to an imperative implementation of the binary exponentiation algorithm. The fully  normalized verification condition would contain a bunch of ∃, /\\, some ∀, and leaves which I think are relatively fast. The proof script is written very carefully to avoid generating large proof terms, in hope that it would improve performance. In particular, the size of the goal is kept modest at all times. The individual proofs steps correspond roughly to variants of `destruct`, `eexists`, and `apply`, and take up less than 200 lines when written out by hand. Proofs of nontrivial leaves are factored out using `abstract`, and the proof ends in `Defined` to avoid potential slowdown from recombination and hashconsing. Actually, the leaves could probably be admitted, but I haven't tried that yet. I took care to keep the proof term relatively readable, although adding notations would probably help even more. In particular, order of subproofs in the proofterm should match the order of parts of the imperative program whose safety they argue.

Executing the proof script takes around 10 seconds. Defined takes 1 minute. Printing the proof term takes 3 minutes.

The proof is at
https://github.com/mit-plv/bedrock2/blob/24abd309a46c039e22e732f9390aabeba4516fce/bedrock2/src/Examples/ipow.v#L134 and relies on lightweight tactics defined earlier in the same file. https://github.com/mit-plv/bedrock2/blob/24abd309a46c039e22e732f9390aabeba4516fce/bedrock2/src/WeakestPrecondition.v#L50 is probably relevant to understand what the goal is. The 500-line proof term is at eac425acac9139580fa15be1e363b603 using notations from https://github.com/mit-plv/bedrock2/blob/24abd309a46c039e22e732f9390aabeba4516fce/bedrock2/src/Examples/ipow.v#L116.

The proof really isn't that big. Some statistics:

~~~~~
browse@ashryn ~ % cat /tmp/a | grep -c let
71
[...]

PR #9078, opened 2018-11-26
Provide a faster bound name generation algorithm through a flag

### LINKED ISSUE #8909: Printing is nonlinear (quadratic?) in the size of the term
#### Version

8.8.1

#### Operating system

Linux

Consider this code:
```coq
Reserved Notation "'dlet' x .. y := v 'in' f"
         (at level 200, x binder, y binder, f at level 200, format "'dlet'  x .. y  :=  v  'in' '//' f").
Definition Let_In {A P} (x : A) (f : forall a : A, P a) : P x := let y := x in f y.
Notation "'dlet' x .. y := v 'in' f" := (Let_In v (fun x => .. (fun y => f) .. )).

Fixpoint lets (n : nat)
  := match n with
     | O => tt
     | S n' => dlet x := tt in lets n'
     end.

Time Definition x10 := Eval cbv [lets] in lets 10. (* Finished transaction in 0.001 secs (0.004u,0.s) (successful) *)
Time Definition x100 := Eval cbv [lets] in lets 100. (* Finished transaction in 0.006 secs (0.004u,0.004s) (successful) *)
Time Definition x1000 := Eval cbv [lets] in lets 1000. (* Finished transaction in 0.158 secs (0.124u,0.036s) (successful) *)
Time Definition x10000 := Eval cbv [lets] in lets 10000. (* Finished transaction in 8.835 secs (8.42u,0.415s) (successful) *)
Time Print x10. (* Finished transaction in 0. secs (0.u,0.s) (successful) *)
Time Print x100. (* Finished transaction in 0.016 secs (0.016u,0.s) (successful) *)
Time Print x1000. (* Finished transaction in 0.722 secs (0.711u,0.008s) (successful) *)
Time Print x10000. (* Finished transaction in 61.083 secs (60.895u,0.008s) (successful) *)
```
What's `Print` spending all of its time doing?

Real example where this came up, where jumping from 2253 lines/binders to 7575 lines/binders made the time of printing go from about 161 s to about 3653 s: https://github.com/mit-plv/fiat-crypto/pull/395 / https://github.com/mit-plv/fiat-crypto/pull/395/commits/4c3405eb1eea7f35e77d1cf2916f54698ca82d4d (or https://web.archive.org/web/20181106002633/https://github.com/mit-plv/fiat-crypto/pull/395/commits/4c3405eb1eea7f35e77d1cf2916f54698ca82d4d if that link goes dead)

cc @ppedrot @herbelin @ejgallego

ppedrot:
This PR adds a `Fast Name Printing` option, which once turned on removes a source of algorithmic explosion in detyping. The tradeoff is that while much faster asymptotically, the names that are picked up are less clever and thus less pretty to the user. Intuitively, we simply remember the highest subscript used for every base identifier and bump it up every time we go under a new bound variable.

The option is not turned on by default, because Ltac observes bound names, so changing the printer actually breaks proof scripts.

I did not document the option apart from CHANGES because I have no idea where to put it. Help welcome.

Fixes #8909.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -175,6 +175,10 @@ Vernacular commands
 - `Coercion` does not warn ambiguous paths which are obviously convertible with
   existing ones.
 
+- A new flag `Fast Name Printing` has been introduced. It changes the
+  algorithm used for allocating bound variable names for a faster but less
+  clever one.
+
 Tools
 
 - The `-native-compiler` flag of `coqc` and `coqtop` now takes an argument which can have three values:
--- a/doc/sphinx/proof-engine/vernacular-commands.rst
+++ b/doc/sphinx/proof-engine/vernacular-commands.rst
@@ -510,6 +510,20 @@ Requests to the environment
 
 .. seealso:: Section :ref:`locating-notations`
 
+.. _printing-flags:
+
+Printing flags
+-------------------------------
+
+.. flag:: Fast Name Printing
+
+   When turned on, |Coq| uses an asymptotically faster algorithm for the
+   generation of unambiguous names of bound variables while printing terms.
+   While faster, it is also less clever and results in a typically less elegant
+   display, e.g. it will generate more names rather than reusing certain names
+   across subterms. This flag is not enabled by default, because as Ltac
+   observes bound names, turning it on can break existing proof scripts.
+
 
 .. _loading-files:
 
--- a/engine/nameops.mli
+++ b/engine/nameops.mli
@@ -24,7 +24,41 @@ val add_prefix : string -> Id.t -> Id.t
 
 (** Below, by {i subscript} we mean a suffix composed solely from (decimal) digits. *)
 
+module Subscript :
+sig
+  type t
+  (** Abstract datatype of subscripts. Isomorphic to a string of digits. *)
+
+  val zero : t
+  (** Empty subscript *)
+
+  val succ : t -> t
+  (** Guarantees that [x < succ x], but [succ x] might not be the smallest
+      element strictly above [x], generally it does not exist. Example mappings:
+      ""   ↦ "0"
+      "0"  ↦ "1"
+      "00" ↦ "01"
+      "1"  ↦ "2"
+      "01" ↦ "02"
+      "9"  ↦ "10"
+      "09" ↦ "10"
+      "99" ↦ "100"
+  *)
+
+  val compare : t -> t -> int
+  (** Well-founded order. *)
+
+  val equal : t -> t -> bool
+
+end
+
 val has_subscript       : Id.t -> bool
+
+val get_subscript : Id.t -> Id.t * Subscript.t
+(** Split an identifier into a base name and a subscript. *)
+
+val add_subscript : Id.t -> Subscript.t -> Id.t
+(** Append the subscript to the identifier. *)
 
 val increment_subscript : Id.t -> Id.t
 (** Return the same identifier as the original one but whose {i subscript} is incremented.
--- a/engine/nameops.ml
+++ b/engine/nameops.ml
@@ -13,6 +13,51 @@ open Names
 
 (* Utilities *)
 
+module Subscript =
+struct
+
+type t = {
+  ss_zero : int;
+  (** Number of leading zeros of the subscript *)
+  ss_subs : int;
+  (** Digital value of the subscript, zero meaning empty *)
+}
+
+let rec overflow n =
+  Int.equal (n mod 10) 9 && (Int.equal (n / 10) 0 || overflow (n / 10))
+
+let zero = { ss_subs = 0; ss_zero = 0 }
+
+let succ s =
+  if Int.equal s.ss_subs 0 then
+    if Int.equal s.ss_zero 0 then
+      (* [] -> [0] *)
+      { ss_zero = 1; ss_subs = 0 }
+    else
+      (* [0...00] -> [0..01] *)
+      { ss_zero = s.ss_zero - 1; ss_subs = 1 }
+  else if overflow s.ss_subs then
+    if Int.equal s.ss_zero 0 then
+      (* [9...9] -> [10...0] *)
+      { ss_zero = 0; ss_subs = 1 + s.ss_subs }
+    else
+      (* [0...009...9] -> [0...010...0] *)
+      { ss_zero = s.ss_zero - 1; ss_subs = 1 + s.ss_subs }
+  else
+    (* [0...0n] -> [0...0{n+1}] *)
+    { ss_zero = s.ss_zero; ss_subs = s.ss_subs + 1 }
+
+let equal s1 s2 =
+  Int.equal s1.ss_zero s2.ss_zero && Int.equal s1.ss_subs s2.ss_subs
+
+let compare s1 s2 =
+  (* Lexicographic order is reversed in order to ensure that [succ] is strictly
+      increasing. *)
+  let c = Int.compare s1.ss_subs s2.ss_subs in
+  if Int.equal c 0 then Int.compare s1.ss_zero s2.ss_zero else c
+
+end
+
 let code_of_0 = Char.code '0'
 let code_of_9 = Char.code '9'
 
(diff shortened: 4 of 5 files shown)


JasonGross:
Interestingly, this does not speed up printing in #9029 by that much (from 107 seconds to 50 seconds).  I haven't tracked down that slowness.  I'm very much in favor of seeing this land, though.

ppedrot:
Summoning @jfehrle and @Zimmi48 for the documentation part.

Zimmi48:
Indeed, it's really not clear to me where to add the documentation for this option. I can only suggest to add a new section to the Ltac chapter since it has an impact on the behavior of the proof engine. Anyways, wherever you put this, it'll be always possible to move it somewhere else after a general refactoring of the reference manual (if that ever happens).

ppedrot:
I added documentation around the `Print` and `Set` commands.

Zimmi48:
Looks good but note that there was already a section named "Controlling display" in this chapter. I suppose it would make sense to merge both, wouldn't it? I tend to prefer the section name you gave ("Printing flags") but maybe it is more cryptic... I wouldn't mind keeping your name and even your location for this section and bumping up the rest of the printing flags from the "Controlling display" section to this new section...

Zimmi48:
(This can also be done later on if you prefer.)

ppedrot:
> (This can also be done later on if you prefer.)

I prefer delaying this to another unrelated PR.

ppedrot:
Who is handling this PR? @mattam82?

CHANGES.md
```diff
@@ -85,6 +85,10 @@ Vernacular commands
 - The `Automatic Introduction` option has been removed and is now the
   default.
 
+- A new option `Fast Name Printing` has been introduced. It changes the
```
jfehrle:
```suggestion
- A new flag`Fast Name Printing` has been introduced. It changes the
```

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -510,6 +510,19 @@ Requests to the environment
...
 
+.. _printing-flags:
+
+Printing flags
+-------------------------------
+
+.. flag:: Fast Name Printing
+
+   When turned on, |Coq| uses an asymptotically faster algorithm for the
+   attribution of non-ambiguous names of bound variables while printing terms.
```
jfehrle:
```suggestion
   attribution of unambiguous names of bound variables while printing terms.
```

jfehrle: "attribution of non-ambiguous names of bound variables" is a mouthful.  Is there a simpler way to say this?  Is "attribution" the best word--not something like "naming"?  I have trouble guessing what "attribution" means here.  (Caveat: still lots of things I don't know about Coq.)

It would be helpful to the reader to give a short concrete example, for example: "ABC" will be printed as "DEF" when the flag is enabled, but otherwise would be printed as "GHI".  This could also illustrate what you mean by "less clever" and "less elegant" which are vague terms.  Is the less-good appearance perhaps related to generating more bound variable names rather than reusing them?

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -510,6 +510,19 @@ Requests to the environment
...
+   While faster, it is also less clever and results in a typically less elegant
+   display. This flag is not enabled by default, because as Ltac observes bound
+   names, turning it on can break existing proof scripts.
+
```
jfehrle: Perhaps you could (briefly) be more explicit about the benefit of and when to use this feature.  It's unexpected to me that there is a printing operation where there is a performance issue.  Maybe you could include/adapt more of the info from the PR description?

> "because as Ltac observes bound names, turning it on can break existing proof scripts."

I don't quite get this.  Do you mean that the bound names appear in the proof scripts?  How the names "observed"?  Is there a simple example of the problem? (BTW, I've not used Ltac.)

jfehrle (dismissed):

doc/sphinx/proof-engine/vernacular-commands.rst
ppedrot: It's not quite naming, as the name hint is already there. It's mostly about desambiguation.

Regarding cleverness, I don't have a short insightful example, but indeed in general more names are generated, e.g. if a constant name clashes in one subterm, the slow algorithm will forbid this name only in the subterm while it will be globally forbidden with the fast one.

doc/sphinx/proof-engine/vernacular-commands.rst
ppedrot: > It's unexpected to me that there is a printing operation where there is a performance issue. 

It's a pervasive issue. I keep seeing this hotspot in virtually any non-trivial printing. The algorithm is at worst cubic and quadractic on average, you don't need a lot of bound variables to see the effects.

> Is there a simple example of the problem?

Typical tactic observing printing is `destruct n` where `n` is a bound name, e.g.
```
Goal forall n : nat, n = n.
Proof.
destruct n.
```


Change the bound name and this will break the script.

jfehrle: I didn't try your example with the new code, but how about adding "You can see the difference by running the following script with and without the option: ..." if that's accurate.

doc/sphinx/proof-engine/vernacular-commands.rst
jfehrle: Maybe add "It will generate more names rather than reusing certain names across subterms."  That explains less clever/elegant a bit without going into detail.

And maybe "generation" would be clearer than "attribution".

doc/sphinx/proof-engine/vernacular-commands.rst
jfehrle: "because as Ltac observes bound names, turning it on can break existing proof scripts"

It can break any script that applies an Ltac tactic, such as "easy"?  Is that the only way it breaks?  Seems like this isn't enough info for the user to know whether they can safely use the flag.

ppedrot: Unluckily, I believe that there is no way to know whether you can safely use the flag apart from calling your tactic and seeing whether it broke. Most tactics rely ultimately on some mechanism that might observe bound names, e.g. even the lower-level variants of intro used virtually everywhere.

maximedenes:
I'm not sure I like this as an option (not criticizing the change itself). I think it should be discussed.

ppedrot:
@maximedenes What is the alternative?

ppedrot:
This is getting stale. When should we discuss what to do?

JasonGross:
I personally would like to see this merged

ppedrot:
@maximedenes can you voice more precisely your concerns? Otherwise this PR should go in, it's a pretty important one for performance.

ppedrot:
@maximedenes ping.

maximedenes:
> @maximedenes can you voice more precisely your concerns? Otherwise this PR should go in, it's a pretty important one for performance.

My concern is: what is the transition strategy that will lead to not keeping the option around?

Sorry if this was explained before, I'm in a meeting so I can't re-read the whole thread.

ppedrot:
> My concern is: what is the transition strategy that will lead to not keeping the option around?

Getting rid of `intros until n` and similar haram tactics that observe bound names for control?

maximedenes:
> Getting rid of `intros until n` and similar haram tactics that observe bound names for control?

I'm personally in favor of it, but I don't know if this view is shared. @herbelin do you also think we should try to remove such constructions (on the long term) from the tactic language, to gain abstraction?

engine/nameops.mli
```diff
@@ -24,8 +24,42 @@ val add_prefix : string -> Id.t -> Id.t
...
+  (** Empty subscript *)
+
+  val succ : t -> t
+  (** Guarantees that [x < succ x], but [succ x] might not be the smallest
+      element strictly above [x], generally it does not exist. Example mappings:
+      ""   ↦ "0"
+      "0"  ↦ "1"
+      "00" ↦ "01"
+      "1"  ↦ "2"
+      "01" ↦ "01"
```
vbgl: This example seems to violate the [x < succ x] guarantee. Typo?

vbgl (approved):
This looks like a change in the right direction.

engine/nameops.mli
ppedrot: Looks like it.

maximedenes:
> Getting rid of `intros until n` and similar haram tactics that observe bound names for control?

I think I'd be more confident in the approach taken by this PR if we had a list of things that need to be deprecated, to ensure a smooth transition to enabling this option by default.

Zimmi48:
> > Getting rid of `intros until n` and similar haram tactics that observe bound names for control?
> 
> I'm personally in favor of it, but I don't know if this view is shared. @herbelin do you also think we should try to remove such constructions (on the long term) from the tactic language, to gain abstraction?

Well, @ppedrot took as example a tactic that was little used, but a much more used tactic that suffers from the exact same problem is `induction n` on a goal like `forall n : nat, ...`. So, unless I misunderstood something, the transition plan proposed by @ppedrot is BS.

That being said, I don't think that this should prevent the merging of this option @maximedenes. For instance, if there are no SSR tactics suffering from this problem, or if there are some, but they can be removed, then this flag would be useful for their users. There are also chances that many projects using vanilla tactics can turn on this flag without any breakage. The breakage that @ppedrot mentioned is possible in theory, but it doesn't mean it will be frequent. I'm not even sure @ppedrot tried to turn the flag on and run the CI.

ppedrot:
AFAIC, this is ready to go. Ideally this should go in 8.10 and thus this needs an assignee quickly. Who's up to the task?

vbgl:
I don’t mind assigning.

vbgl:
Pierre-Marie, can you please address the linting error:
> Checking 649b0e02f43a498b08444b5dcdf156b8f562dc2d
> pretyping/detyping.ml:390: tab in indent.
> `+	| LetIn (na,b,t,c),true ->`

PR #9081, opened 2018-11-26
[dune] A new Makefile.dune target for each package (coq, coqide-server and coqide).

ejgallego:


ejgallego:
As to help @mattam82

ejgallego:
We hope to be smarter with optional libs in the future.

gares:
I think you should consider using the dune equivalent of configure for this. Point is if gtk is optional and native compute (will be) optional we can't add all combinations as targets. It works for now, but can't scale.

ejgallego:
What's the dune equivalent of configure?

Ummm, native compute is a flag so it is handled by profiles. It seems to me that using the install targets scales better than adding adding additional programs and basically exponential logic. The build of the binary should be conditional tho as is the one of the library, I guess that should be possible to specify soon.

Note that the makefile exists mainly , from my point of view, as documentation. I never use it for example, and I would expect it to go away.

Anyways, it is still far from the number of (undocumented) targets the make-based system has so I don't see a reason to worry as we still have some more space in the terminal.

Makefile.dune
```diff
@@ -1,7 +1,7 @@
 # -*- mode: makefile -*-
 # Dune Makefile for Coq
 
-.PHONY: help voboot states world watch check quickbyte quickopt test-suite release apidoc ocheck ireport clean
+.PHONY: help voboot states world nogtk watch check quickbyte quickopt test-suite release apidoc ocheck ireport clean
```
Zimmi48: You were splitting this in another PR. I think it's a good idea to stop this line from growing forever.

Zimmi48 (commented):
I don't like the `nogtk` name and I don't like the underlying concept either.
May I suggest that you remove the `world` target instead and replace it with a `coq`, a `coqide`, and a `coqide-server` targets? Indeed, I think we should strive to keep this `Makefile.dune` as close as possible to the underlying `dune` commands.

As for the configure solution, I reject it as well. lablgtk is not an optional dependency of Coq, it is a required dependency of CoqIDE.

SkySkimmer:
>Indeed, I think we should strive to keep this Makefile.dune as close as possible to the underlying dune commands.

The world target is just 1 dune command though (modulo voboot).

Is it possible to make coqide an optional package?

gares:
> What's the dune equivalent of configure?

https://jbuilder.readthedocs.io/en/latest/configurator.html

> As for the configure solution, I reject it as well. lablgtk is not an optional dependency of Coq, it is a required dependency of CoqIDE.

I agree that in the dune philosophy CoqIDE would be a separate module, if you put it there you want to compile it and it needs gtk; if you don't put it in your source tree you simply don't compile it. But today this is not how the sources are organized. I don't know if the configurator above can easily skip a subdirectory when the discovery program fails.

Zimmi48:
>I agree that in the dune philosophy CoqIDE would be a separate module

Separate package. Dune supports very well having separate packages in a single "monorepo".

ejgallego:
@Zimmi48 this was recently requested on gitter thus updating this PR following your suggestion: one target per package.

Ideally we would use https://github.com/ocaml/dune/issues/1690 so `world` would still work fine in this case [emitting a warning]

Zimmi48 (approved):

PR #9087, opened 2018-11-27
Add CI job building stdlib with `async-proofs on`

maximedenes:
This should be useful to make sure we don't introduce regressions while working on STM improvements.

Based on #9072.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -225,6 +225,13 @@ build:egde:dune:dev:
     OPAM_SWITCH: edge
     DUNE_TARGET: world
 
+build:base+async:
+  <<: *build-template
+  stage: test
+  variables:
+    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
+    COQUSERFLAGS: "-async-proofs on"
+
 windows64:
   <<: *windows-template
   variables:
--- a/Makefile.build
+++ b/Makefile.build
@@ -44,6 +44,9 @@ NO_RECALC_DEPS ?=
 # Non-empty runs the checker on all produced .vo files:
 VALIDATE ?=
 
+# When non-empty, passed as extra arguments to coqtop/coqc:
+COQUSERFLAGS ?=
+
 # Output file names for timed builds
 TIME_OF_BUILD_FILE               ?= time-of-build.log
 TIME_OF_BUILD_BEFORE_FILE        ?= time-of-build-before.log
@@ -191,7 +194,7 @@ TIMER=$(if $(TIMED), $(STDTIME), $(TIMECMD))
 # the output format of the unix command time. For instance:
 #   TIME="%C (%U user, %S sys, %e total, %M maxres)"
 
-COQOPTS=$(NATIVECOMPUTE) $(COQWARNERROR)
+COQOPTS=$(NATIVECOMPUTE) $(COQWARNERROR) $(COQUSERFLAGS)
 BOOTCOQC=$(TIMER) $(COQTOPBEST) -boot $(COQOPTS) -compile
 
 LOCALINCLUDES=$(addprefix -I ,$(SRCDIRS))
--- a/theories/Strings/BinaryString.v
+++ b/theories/Strings/BinaryString.v
@@ -48,7 +48,7 @@ Module Raw.
               end
        end.
 
-  Fixpoint to_N_of_pos (p : positive) (rest : string) (base : N)
+  Fixpoint to_N_of_pos (p : positive) (rest : string) (base : N) {struct p}
     : to_N (of_pos p rest) base
       = to_N rest match base with
                   | N0 => N.pos p
--- a/theories/Strings/HexString.v
+++ b/theories/Strings/HexString.v
@@ -120,7 +120,7 @@ Module Raw.
               end
        end.
 
-  Fixpoint to_N_of_pos (p : positive) (rest : string) (base : N)
+  Fixpoint to_N_of_pos (p : positive) (rest : string) (base : N) {struct p}
     : to_N (of_pos p rest) base
       = to_N rest match base with
                   | N0 => N.pos p
--- a/theories/Strings/OctalString.v
+++ b/theories/Strings/OctalString.v
@@ -78,7 +78,7 @@ Module Raw.
               end
        end.
 
-  Fixpoint to_N_of_pos (p : positive) (rest : string) (base : N)
+  Fixpoint to_N_of_pos (p : positive) (rest : string) (base : N) {struct p}
     : to_N (of_pos p rest) base
       = to_N rest match base with
                   | N0 => N.pos p


maximedenes:
I deferred the test-suite part to #9088, because it will require fixing some issues.

.gitlab-ci.yml
```diff
@@ -77,7 +77,7 @@ after_script:
 
     - echo 'start:coq.build'
     - make -j "$NJOBS" byte
-    - make -j "$NJOBS" world $EXTRA_TARGET
+    - make COQUSERFLAGS="${COQ_USER_FLAGS}" -j "$NJOBS" world $EXTRA_TARGET
```
SkySkimmer: Remove the configure change, do `COQUSERFLAGS?=` in makefile.build before using the variable, and do nothing here (pass as environment variable).

SkySkimmer (changes requested):

.gitlab-ci.yml
maximedenes: Done, thanks for the suggestion!

maximedenes:
Running a CI build that is expected to fail, to make sure we test the right thing.

.gitlab-ci.yml
```diff
@@ -225,6 +225,13 @@ build:egde:dune:dev:
     OPAM_SWITCH: edge
     DUNE_TARGET: world
 
+build:asyncproofs:
+  <<: *build-template
+  variables:
+    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
+    COQUSERFLAGS: "-async-proofs on"
+    EXTRA_TARGET: "stdlib"
```
SkySkimmer: stdlib is the doc target, you shouldn't need to use extra_target here.

SkySkimmer:
Sorry about the review request churn, I got confused about which buttons to click on.

.gitlab-ci.yml
```diff
@@ -225,6 +225,13 @@ build:egde:dune:dev:
     OPAM_SWITCH: edge
     DUNE_TARGET: world
 
+build:asyncproofs:
```
ejgallego: Maybe you can just reuse the edge build for this?

Anyways, the naming is not good, it should be `build:base+async`

ejgallego (changes requested):

maximedenes:
Comments addressed, but the `base+async` build is a bit slow. Does it mean I should move it somewhere else? If so, what do you guys recommend to do, that still makes #9088 possible?

ejgallego:
> Comments addressed, but the `base+async` build is a bit slow.

Why is so slow? IMHO it looks too slow, right? Maybe you need `-j1`?

If possible we should try to understand what's wrong a bit better, but if you want that it I guess you can just move it to the test phase.

ejgallego:
Well in fact it doesn't even build so it seems

ejgallego (changes requested):
I am OK with the CI setup however the build doesn't seem to work as of today, so I think a bit more tweaking is needed, it seems that some quadratic behavior is happening, or maybe too much of a worker churn.

In particular we have these strange errors about parsing that should go away IMO before we allow this in.

SkySkimmer:
>Comments addressed, but the base+async build is a bit slow. Does it mean I should move it somewhere else? If so, what do you guys recommend to do, that still makes #9088 possible?

We could move to the test stage and run the -async test suite either in the same job or using a regular build as a dependency.
We should also look into running validate on the output of async built stdlib.

SkySkimmer (approved):

ejgallego:
> We could move to the test stage and run the -async test suite either in the same job or using a regular build as a dependency.

The async build is too broken so moving to test won't work I'm afraid.

maximedenes:
> Why is so slow? IMHO it looks too slow, right? Maybe you need `-j1`?

I think it is expected. It is slow because in the absence of `.aux` files, it delegates all the proofs. But that's also good to get more testing.

The build is currently broken because of a race. I'll need @gares' help to fix it.

maximedenes:
Ok, so it seems we successfully addressed the race condition, which revealed an incorrect handling of universes in proofs delegated to workers.

What remains to be done:
- [ ] move the job to the test stage
- [ ] check the results of bench (jenkins num 598)
- [ ] clean-up the history

ejgallego:
So this still takes like more than a 5x overhead in the compilation; not a good thing, as this even questions the feature at all. Why would anyone want to use it?

Maybe you could try to do a two pass compilation where you generate the `.aux` files?

maximedenes:
> So this still takes like more than a 5x overhead in the compilation; not a good thing, as this even questions the feature at all. Why would anyone want to use it?

I agree that this is a concern for real-world use, although `.aux` files help there. Regarding CI (this PR), I would prefer that all proofs are delegated since it then tests many more exotic cases.

Said otherwise, we could discuss removing parallel processing altogether, if this is what you mean. But if we want to keep it, I think this test should be there, even if a bit heavy (but of course, moved to the right stage, then).

ejgallego:
CI-wise we can assume this without too much problem, I was more like thinking if we are testing what the users really use.

Maybe you would like to this first pass with no `.aux` and then a second one in the same job with `.aux` on?

I don't think we should remove async, however the whole business needs a lot of improvement.

ejgallego:
> * check the results of bench (jenkins num 598)

Why are the results of the bench relevant here?

Shouldn't you spun off the fix to a different PR? (=> I tend to strongly think so)

maximedenes:
> Shouldn't you spun off the fix to a different PR? (=> I tend to strongly think so)

Oh sure, will do.

maximedenes:
> Maybe you would like to this first pass with no `.aux` and then a second one in the same job with `.aux` on?

Ah ok, I misunderstood, I thought you meant a first pass without async, just to generate the `.aux` file. Two async passes indeed make sense.

ejgallego:
> Ah ok, I misunderstood, I thought you meant a first pass without async, just to generate the `.aux` file. Two async passes indeed make sense.

Yup, I guess you'll need to do a little hack, something such as `find -name *.vo -exec rm {}`

ejgallego:
> I thought you meant a first pass without async, just to generate the `.aux` file.

I meant indeed both possibilities, I am OK with the 2 passes being on async.

SkySkimmer:
>Yup, I guess you'll need to do a little hack, something such as find -name *.vo -exec rm {}

We have `make voclean` for that.
Is there anything to do when one wants to use .aux stuff with dune?

SkySkimmer:
BTW @ejgallego maybe you should take the assignment from me on this one.

maximedenes:
> Note that by default you have just one worker, it is really tailored toward interactive use where you mostly care about mater giving quick feedback and proof being eventually checked in the background. If you want to test parallel compilation you go through vio files since all vio can be turned into vo in parallel, only the v to vio compilation needs to honour topological order.

I would rather not have this depend on the `-quick` chain, as we may want to remove it. On the other hand, we could set the number of workers to something higher than 1.

gares:
This was not a suggestion for this PR.

It was just explaining how parallelism can actually be exploited (that is not by just passing `-async-proofs on`).

Indeed the name does not suggest any speedup to me, but rather the fact that the document is checked out-of-order, that is not from the first to the last line since some lines are evaluated asynchronously).

gares (approved):

ejgallego (approved):
I am OK with this even if I am not sure I do see a lot of benefit in testing a setup that users will never use.

I'd like to hear some comment about the problem with body-less fixpoints, maybe @SkySkimmer @ppedrot @herbelin want to add anything? Otherwise I'll merge soon.

ejgallego:
Well, I see this still have a progress tag, thus waiting.

SkySkimmer:
I don't see anything fundamentally wrong with bodyless fixpoints. The implementation may do buggy things though (I'm not familiar with it).

maximedenes:
> I'd like to hear some comment about the problem with body-less fixpoints

Bodyless fixpoints are ok. What is not right is to expose a construction that performs an (interactive) induction, without asking on what the induction is to be done.

I have a plan to fix that design (require a `struct` annotation for bodyless fixpoints / lemmas by induction), but I think it is fairly orthogonal to STM problems. For example, even in sequential mode, `Guarded` can report false positives, and `Show` will display a proof term by induction on an argument that won't be the final one...

So I'd suggest to merge the quick workaround for now, and have a more complete discussion in a separate issue.

SkySkimmer:
>What is not right is to expose a construction that performs an (interactive) induction, without asking on what the induction is to be done.

I don't think it's the real problem.
Basically `Fixpoint f : nat -> nat.` wants to fill a hole `f : nat -> nat |- nat -> nat` by tactics and then do some postprocessing on it to turn it into a fixpoint. Guarded/Show are not aware of this postprocessing but that seems like a problem with the start_proof API forcing to pretend that we won't need postprocessing somehow (unless the API doesn't actually force that, in which case it's just a problem of how Fixpoint calls it).

maximedenes:
> I don't think it's the real problem.
> Basically `Fixpoint f : nat -> nat.` wants to fill a hole `f : nat -> nat |- nat -> nat` by tactics and then do some postprocessing on it to turn it into a fixpoint.

I'm not sure I follow. The first step it does is to refine the proof term to a partial `fix` term. The kernel API makes it mandatory to specify a recursive argument, so vernac passes a dummy one. And that's already looking wrong IMHO.

SkySkimmer:
>The first step it does is to refine the proof term to a partial |fix| 
term.

Doesn't seem like a workaround to running ltac on whatever goals that 
would produce and putting them in the correctly annotated fix afterwards?

Gaëtan Gilbert

ejgallego:
Indeed I think the right fix is to warn, then forbid, the bodyless construction without the proper annotation.

maximedenes:
> Indeed I think the right fix is to warn, then forbid, the bodyless construction without the proper annotation.

Yes, but it requires a bit of work because `Lemma ... with ...` does not even support the annotation currently... #9165 is a preliminary clean-up before I implement it.

But for the scope of this PR, I think a workaround is ok. I'm not even sure the annotations were omitted on purpose. There's really one occurrence in stdlib, it is just that we duplicate code...

maximedenes:
Thanks!

Issue #3934, opened 2015-01-18
coqc -time -quick gives unreadable output


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#3934
From: @JasonGross
Reported version: trunk
CC:   @ejgallego

PR #9088, opened 2018-11-27
Add CI job running test suite with `async-proofs on`

maximedenes:
This is based on #9072 and #9087.

Will probably require fixing some STM problems revealed here.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -151,7 +151,7 @@ after_script:
     - BIN=$(readlink -f ../_install_ci/bin)/
     - LIB=$(readlink -f ../_install_ci/lib/coq)/
     - export OCAMLPATH=$(readlink -f ../_install_ci/lib/):"$OCAMLPATH"
-    - make -j "$NJOBS" BIN="$BIN" LIB="$LIB" all
+    - make -j "$NJOBS" BIN="$BIN" LIB="$LIB" COQFLAGS="${COQFLAGS}" all
   artifacts:
     name: "$CI_JOB_NAME.logs"
     when: on_failure
@@ -426,6 +426,13 @@ test-suite:edge+trunk+dune:
       - _build/default/test-suite/logs
     expire_in: 1 week
   allow_failure: true
+
+test-suite:base+async:
+  <<: *test-suite-template
+  dependencies:
+    - build:base
+  variables:
+    COQFLAGS: "-async-proofs on"
 
 validate:base:
   <<: *validate-template
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -36,9 +36,10 @@ include ../Makefile.common
 # easily overridden
 LIB := ..
 BIN := $(shell cd ..; pwd)/bin/
+COQFLAGS?=
 
-coqtop := $(BIN)coqtop -coqlib $(LIB) -boot -q -batch -test-mode -R prerequisite TestSuite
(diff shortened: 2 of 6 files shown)


test-suite/Makefile
```diff
@@ -36,15 +36,16 @@ include ../Makefile.common
 # easily overridden
 LIB := ..
 BIN := $(shell cd ..; pwd)/bin/
+COQFLAGS := ""
```
SkySkimmer: Pretty sure this is broken. Use `COQFLAGS?=` instead.

maximedenes:
The current build trace seems to reveal a race, maybe @gares or @ejgallego know about it. Is it the one you wrote a comment about in the code, Emilio?

```
31358:master:4] Uncaught exception in worker manager: 
                Anomaly
                "The global environment cannot be accessed during parsing."
```

ejgallego:
> Is it the one you wrote a comment about in the code, Emilio?

We had a long discussion about this check at Nice, this check was introduced by @gares .

.gitlab-ci.yml
```diff
@@ -132,7 +132,7 @@ after_script:
     - BIN=$(readlink -f ../_install_ci/bin)/
     - LIB=$(readlink -f ../_install_ci/lib/coq)/
     - export OCAMLPATH=$(readlink -f ../_install_ci/lib/):"$OCAMLPATH"
-    - make -j "$NJOBS" BIN="$BIN" LIB="$LIB" all
+    - make -j "$NJOBS" BIN="$BIN" LIB="$LIB" COQFLAGS="${COQFLAGS}" all
```
ejgallego: Drop the `{}` from the quoting.

ejgallego (approved):
Seems OK, but needs to work.

maximedenes:
I need help (probably from @gares and @SkySkimmer) to fix the universe issues. It is probably revealing some serious problems.

maximedenes:
That is, we need to fix #9093.

maximedenes:
Hum the last run shows more problems, something must be off.

maximedenes:
@SkySkimmer when I download the artifact, I don't see the `bugs/opened` files (I do see the other ones, though). Is this an infrastructure bug?

SkySkimmer:
Yeah we explicitly exclude the opened/ logs. I think I had a misunderstanding about how they work when I wrote the script or some such issue. If you just remove `'(' -path ./bugs/opened -prune ')' -o` from report.sh it should work (maybe? I still don't fully understand the opened/ tests)

maximedenes:
@gares the regression in the test suite was introduced by 4838146ee993a3bafb6a359dff529aae815cfff6

Some files contain non-terminated proofs that trigger anomalies. But since that commit, the anomaly has been replaced by an error saying that the proof is not closed.

What seems strange is that `coqc -async-proofs on` now has a behavior which is different from `coqc` but also different from opening `coqide` and jumping to the end of the file. Is that expected?

maximedenes:
I made the tests more robust, it should go through now.

Dear coqbot, can you trigger a CI build even if you are not in a good mood today? I'll send you chocolates for Christmas.

maximedenes:
The last remaining problem is #3934. Working on a fix.

maximedenes:
Finally, everything passed! @ejgallego can we merge this, to make sure we don't introduce STM-related regressions? Thanks a lot!

maximedenes:
Thanks!

PR #9092, opened 2018-11-27
Modify coqdev-setup-proofgeneral to work with dune compiled Coq.

SkySkimmer:
It seems to work.

### DIFF
--- a/dev/tools/coqdev.el
+++ b/dev/tools/coqdev.el
@@ -72,18 +72,19 @@ Specifically `camldebug-command-name' and `ocamldebug-command-name'."
 (make-variable-buffer-local 'coq-prog-args)
 (setq-default coq-prog-args nil)
 
+(defvar coqdev-dune-context "_build/default/")
+
 (defun coqdev-setup-proofgeneral ()
   "Setup Proofgeneral variables for Coq development.
 
 Note that this function is executed before _Coqproject is read if it exists."
-  (let ((dir (coqdev-default-directory)))
+  (let* ((dir (coqdev-default-directory))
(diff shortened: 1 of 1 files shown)


ejgallego (approved):
I have a bit of different setup that I should reconcile; I am OK merging this if that's useful for you tho.

Something I avoid is having to type make as the idea is to use an alias so executing coqtop would make sure the right libraries are rebuilt.

Also, I bind make to `dune build @check`, but I am still not happy with that until we get rid of the `voboot` target which is a pain in the ass.

Zimmi48:
Looks good. This supersede the two "shim" PRs, right?

SkySkimmer:
No, coqdev only works within the Coq repo. Calling Coq from somewhere unrelated in the filesystem is still a pain.

ejgallego:
@SkySkimmer what kind of progress do you want here?

SkySkimmer:
>I have a bit of different setup that I should reconcile

I am very interested in your setup and general workflow with dune.
I'm not sure what exactly is missing for me to start using dune by default, I'll have to try again sometime. Maybe I'll wait for ocaml/dune#1555 though.

SkySkimmer:
Needs further thought.

ejgallego:
> I am very interested in your setup and general workflow with dune.

Unfortunately I feel like my current setup is still very unstable thus not suitable for further release; but indeed I hope we can finish the Dune Coq issue before Christmas [almost have a proper setup for recursive modules] and then things should be way smoother.

We will still have a hard time managing libs tho until we can add lib support.

PR #9095, opened 2018-11-27
[toplevel] Deprecate `-compile` flag in favor of `coqc`

ejgallego:
This PR deprecates the use of `coqtop` as a compiler.

There is no point on having two binaries with the same purpose; after
the experiment in #8690, IMHO we have a lot to gain in terms of code
organization by splitting the compiler and the interactive binary.

We adapt the documentation and adapt the test-suite.

Note that we don't make `coqc` a true binary yet, but here we take
care only of the user-facing part.

The conversion of the binary will take place in #8690 and will bring
code simplification, including a unified handling of arguments.

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -198,7 +198,7 @@ TIMER=$(if $(TIMED), $(STDTIME), $(TIMECMD))
 #   TIME="%C (%U user, %S sys, %e total, %M maxres)"
 
 COQOPTS=$(NATIVECOMPUTE) $(COQWARNERROR) $(COQUSERFLAGS)
-BOOTCOQC=$(TIMER) $(COQTOPBEST) -boot $(COQOPTS) -compile
+BOOTCOQC=$(TIMER) $(COQTOPBEST) -boot $(COQOPTS) -w -deprecate-compile-arg -compile
 
 LOCALINCLUDES=$(addprefix -I ,$(SRCDIRS))
 MLINCLUDES=$(LOCALINCLUDES)
--- a/doc/sphinx/addendum/parallel-proof-processing.rst
+++ b/doc/sphinx/addendum/parallel-proof-processing.rst
@@ -52,7 +52,7 @@ in interactive mode.
 It is not strictly mandatory in batch mode if it is not the first time
 the file is compiled and if the file itself did not change. When the
 proof does not begin with Proof using, the system records in an
-auxiliary file, produced along with the `.vo` file, the list of section
+auxiliary file, produced along with the ``.vo`` file, the list of section
 variables used.
 
 Automatic suggestion of proof annotations
@@ -154,22 +154,22 @@ to a worker process. The threshold can be configured with
 Batch mode
 ---------------
 
-When |Coq| is used as a batch compiler by running `coqc` or `coqtop`
--compile, it produces a `.vo` file for each `.v` file. A `.vo` file contains,
-among other things, theorem statements and proofs. Hence to produce a
-.vo |Coq| need to process all the proofs of the `.v` file.
+When |Coq| is used as a batch compiler by running ``coqc``, it produces
+a ``.vo`` file for each ``.v`` file. A ``.vo`` file contains, among other
+things, theorem statements and proofs. Hence to produce a .vo |Coq|
+need to process all the proofs of the ``.v`` file.
 
 The asynchronous processing of proofs can decouple the generation of a
-compiled file (like the `.vo` one) that can be loaded by ``Require`` from the
+compiled file (like the ``.vo`` one) that can be loaded by ``Require`` from the
 generation and checking of the proof objects. The ``-quick`` flag can be
-passed to `coqc` or `coqtop` to produce, quickly, `.vio` files.
-Alternatively, when using a Makefile produced by `coq_makefile`,
+passed to ``coqc`` or ``coqtop`` to produce, quickly, ``.vio`` files.
+Alternatively, when using a Makefile produced by ``coq_makefile``,
 the ``quick`` target can be used to compile all files using the ``-quick`` flag.
 
-A `.vio` file can be loaded using ``Require`` exactly as a `.vo` file but
+A ``.vio`` file can be loaded using ``Require`` exactly as a ``.vo`` file but
 proofs will not be available (the Print command produces an error).
 Moreover, some universe constraints might be missing, so universes
-inconsistencies might go unnoticed. A `.vio` file does not contain proof
+inconsistencies might go unnoticed. A ``.vio`` file does not contain proof
 objects, but proof tasks, i.e. what a worker process can transform
 into a proof object.
 
@@ -177,52 +177,52 @@ Compiling a set of files with the ``-quick`` flag allows one to work,
 interactively, on any file without waiting for all the proofs to be
 checked.
 
-When working interactively, one can fully check all the `.v` files by
-running `coqc` as usual.
+When working interactively, one can fully check all the ``.v`` files by
+running ``coqc`` as usual.
 
-Alternatively one can turn each `.vio` into the corresponding `.vo`. All
+Alternatively one can turn each ``.vio`` into the corresponding ``.vo``. All
 .vio files can be processed in parallel, hence this alternative might
 be faster. The command ``coqtop -schedule-vio2vo 2 a b c`` can be used to
-obtain a good scheduling for two workers to produce `a.vo`, `b.vo`, and
-`c.vo`. When using a Makefile produced by `coq_makefile`, the ``vio2vo`` target
-can be used for that purpose. Variable `J` should be set to the number
+obtain a good scheduling for two workers to produce ``a.vo``, ``b.vo``, and
+``c.vo``. When using a Makefile produced by ``coq_makefile``, the ``vio2vo`` target
+can be used for that purpose. Variable ``J`` should be set to the number
 of workers, e.g. ``make vio2vo J=2``. The only caveat is that, while the
-.vo files obtained from `.vio` files are complete (they contain all proof
+.vo files obtained from ``.vio`` files are complete (they contain all proof
 terms and universe constraints), the satisfiability of all universe
 constraints has not been checked globally (they are checked to be
 consistent for every single proof). Constraints will be checked when
-these `.vo` files are (recursively) loaded with ``Require``.
+these ``.vo`` files are (recursively) loaded with ``Require``.
 
 There is an extra, possibly even faster, alternative: just check the
-proof tasks stored in `.vio` files without producing the `.vo` files. This
+proof tasks stored in ``.vio`` files without producing the ``.vo`` files. This
 is possibly faster because all the proof tasks are independent, hence
 one can further partition the job to be done between workers. The
 ``coqtop -schedule-vio-checking 6 a b c`` command can be used to obtain a
-good scheduling for 6 workers to check all the proof tasks of `a.vio`,
-`b.vio`, and `c.vio`. Auxiliary files are used to predict how long a proof
+good scheduling for 6 workers to check all the proof tasks of ``a.vio``,
+``b.vio``, and ``c.vio``. Auxiliary files are used to predict how long a proof
 task will take, assuming it will take the same amount of time it took
 last time. When using a Makefile produced by coq_makefile, the
-``checkproofs`` target can be used to check all `.vio` files. Variable `J`
+``checkproofs`` target can be used to check all ``.vio`` files. Variable ``J``
 should be set to the number of workers, e.g. ``make checkproofs J=6``. As
-when converting `.vio` files to `.vo` files, universe constraints are not
+when converting ``.vio`` files to ``.vo`` files, universe constraints are not
 checked to be globally consistent. Hence this compilation mode is only
 useful for quick regression testing and on developments not making
-heavy use of the `Type` hierarchy.
+heavy use of the ``Type`` hierarchy.
 
 Limiting the number of parallel workers
 --------------------------------------------
 
 Many |Coq| processes may run on the same computer, and each of them may
-start many additional worker processes. The `coqworkmgr` utility lets
+start many additional worker processes. The ``coqworkmgr`` utility lets
 one limit the number of workers, globally.
 
 The utility accepts the ``-j`` argument to specify the maximum number of
-workers (defaults to 2). `coqworkmgr` automatically starts in the
+workers (defaults to 2). ``coqworkmgr`` automatically starts in the
 background and prints an environment variable assignment
 like ``COQWORKMGR_SOCKET=localhost:45634``. The user must set this variable
 in all the shells from which |Coq| processes will be started. If one
 uses just one terminal running the bash shell, then
 ``export ‘coqworkmgr -j 4‘`` will do the job.
 
-After that, all |Coq| processes, e.g. `coqide` and `coqc`, will respect the
+After that, all |Coq| processes, e.g. ``coqide`` and ``coqc``, will respect the
 limit, globally.
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -38,14 +38,15 @@ LIB := ..
 BIN := $(shell cd ..; pwd)/bin/
 COQFLAGS?=
 
-coqtop := $(BIN)coqtop -coqlib $(LIB) -boot -q -batch -test-mode -R prerequisite TestSuite $(COQFLAGS)
+coqc_boot := $(BIN)coqc -coqlib $(LIB) -boot -q -test-mode -R prerequisite TestSuite $(COQFLAGS)
 coqc := $(BIN)coqc -coqlib $(LIB) -R prerequisite TestSuite $(COQFLAGS)
 coqchk := $(BIN)coqchk -coqlib $(LIB) -R prerequisite TestSuite
 coqdoc := $(BIN)coqdoc
+coqtop := $(BIN)coqtop -batch -coqlib $(LIB) -boot -q -test-mode -R prerequisite TestSuite
 coqtopbyte := $(BIN)coqtop.byte
 
-coqtopload := $(coqtop) -async-proofs-cache force -load-vernac-source
-coqtopcompile := $(coqtop) -async-proofs-cache force -compile
+coqc_interactive := $(coqc) -async-proofs-cache force
+coqc_boot_interactive := $(coqc_boot) -async-proofs-cache force
 coqdep := $(BIN)coqdep -coqlib $(LIB)
 
 VERBOSE?=
@@ -60,11 +61,7 @@ SINGLE_QUOTE="
 #" # double up on the quotes, in a comment, to appease the emacs syntax highlighter
 # wrap the arguments in parens, but only if they exist
 get_coq_prog_args_in_parens = $(subst $(SINGLE_QUOTE),,$(if $(call get_coq_prog_args,$(1)), ($(call get_coq_prog_args,$(1)))))
-# get the command to use with this set of arguments; if there's -compile, use coqc, else use coqtop
-has_profile_ltac_or_compile_flag = $(filter "-profile-ltac-cutoff" "-profile-ltac" "-compile",$(call get_coq_prog_args,$(1)))
-get_command_based_on_flags = $(if $(call has_profile_ltac_or_compile_flag,$(1)),$(coqtopcompile),$(coqtopload))
 get_set_impredicativity= $(filter "-impredicative-set",$(call get_coq_prog_args,$(1)))
-
 
 bogomips:=
 ifneq (,$(wildcard /proc/cpuinfo))
@@ -209,7 +206,7 @@ $(addsuffix .log,$(wildcard bugs/opened/*.v)): %.v.log: %.v
 	@echo "TEST      $<  $(call get_coq_prog_args_in_parens,"$<")"
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
-	  $(coqtopcompile) "$<" $(call get_coq_prog_args,"$<") 2>&1; R=$$?; times; \
+	  $(coqc) "$<" $(call get_coq_prog_args,"$<") 2>&1; R=$$?; times; \
 	  if [ $$R = 0 ]; then \
 	    echo $(log_success); \
 	    echo "    $<...still active"; \
@@ -231,7 +228,7 @@ $(addsuffix .log,$(wildcard bugs/closed/*.v)): %.v.log: %.v
 	@echo "TEST      $< $(call get_coq_prog_args_in_parens,"$<")"
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
-	  $(coqtopcompile) "$<" $(call get_coq_prog_args,"$<") 2>&1; R=$$?; times; \
+	  $(coqc) "$<" $(call get_coq_prog_args,"$<") 2>&1; R=$$?; times; \
 	  if [ $$R = 0 ]; then \
 	    echo $(log_success); \
 	    echo "    $<...Ok"; \
@@ -297,7 +294,7 @@ $(addsuffix .log,$(wildcard prerequisite/*.v)): %.v.log: %.v
 	@echo "TEST      $< $(call get_coq_prog_args_in_parens,"$<")"
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
-	  $(coqtopcompile) "$<" $(call get_coq_prog_args,"$<") 2>&1; R=$$?; times; \
+	  $(coqc) "$<" $(call get_coq_prog_args,"$<") 2>&1; R=$$?; times; \
 	  if [ $$R != 0 ]; then \
 	    echo $(log_failure); \
 	    echo "    $<...could not be prepared" ; \
@@ -316,7 +313,7 @@ $(addsuffix .log,$(wildcard ssr/*.v success/*.v micromega/*.v modules/*.v)): %.v
 	$(HIDE){ \
 	  opts="$(if $(findstring modules/,$<),-R modules Mods)"; \
 	  echo $(call log_intro,$<); \
-	  $(coqtopcompile) "$<" $(call get_coq_prog_args,"$<") $$opts 2>&1; R=$$?; times; \
+	  $(coqc) "$<" $(call get_coq_prog_args,"$<") $$opts 2>&1; R=$$?; times; \
 	  if [ $$R = 0 ]; then \
 	    echo $(log_success); \
 	    echo "    $<...Ok"; \
@@ -342,7 +339,7 @@ $(addsuffix .log,$(wildcard stm/*.v)): %.v.log: %.v
 	@echo "TEST      $< $(call get_coq_prog_args_in_parens,"$<")"
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
-	  $(coqtopcompile) "$<" $(call get_coq_prog_args,"$<") -async-proofs on \
+	  $(coqc) "$<" $(call get_coq_prog_args,"$<") -async-proofs on \
 	    $$opts 2>&1; R=$$?; times; \
 	  if [ $$R = 0 ]; then \
 	    echo $(log_success); \
@@ -367,7 +364,7 @@ $(addsuffix .log,$(wildcard failure/*.v)): %.v.log: %.v $(PREREQUISITELOG)
 	@echo "TEST      $< $(call get_coq_prog_args_in_parens,"$<")"
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
-	  $(coqtopcompile) "$<" $(call get_coq_prog_args,"$<") 2>&1; R=$$?; times; \
+	  $(coqc) "$<" $(call get_coq_prog_args,"$<") 2>&1; R=$$?; times; \
 	  if [ $$R = 0 ]; then \
 	    echo $(log_success); \
 	    echo "    $<...Ok"; \
@@ -392,7 +389,7 @@ $(addsuffix .log,$(wildcard output/*.v)): %.v.log: %.v %.out $(PREREQUISITELOG)
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
 	  output=$*.out.real; \
-	  $(call get_command_based_on_flags,"$<") "$<" $(call get_coq_prog_args,"$<") 2>&1 \
+	  $(coqc_boot_interactive) "$<" $(call get_coq_prog_args,"$<") 2>&1 \
 	    | grep -v "Welcome to Coq" \
 	    | grep -v "\[Loading ML file" \
 	    | grep -v "Skipping rcfile loading" \
@@ -431,7 +428,7 @@ $(addsuffix .log,$(wildcard output-modulo-time/*.v)): %.v.log: %.v %.out
 	  echo $(call log_intro,$<); \
 	  tmpoutput=`mktemp /tmp/coqcheck.XXXXXX`; \
 	  tmpexpected=`mktemp /tmp/coqcheck.XXXXXX`; \
-	  $(call get_command_based_on_flags,"$<") "$<" $(call get_coq_prog_args,"$<") 2>&1 \
+	  $(coqc_boot_interactive) "$<" $(call get_coq_prog_args,"$<") 2>&1 \
 	    | grep -v "Welcome to Coq" \
 	    | grep -v "\[Loading ML file" \
 	    | grep -v "Skipping rcfile loading" \
@@ -486,7 +483,7 @@ $(addsuffix .log,$(wildcard complexity/*.v)): %.v.log: %.v $(PREREQUISITELOG)
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
 	  true "extract effective user time"; \
-	  res=`$(call get_command_based_on_flags,"$<") "$<" $(call get_coq_prog_args,"$<") 2>&1 | sed -n -e "s/Finished transaction in .*(\([0-9]*\.[0-9]*\)u.*)/\1/p" | head -1`; \
+	  res=`$(coqc_boot_interactive) "$<" $(call get_coq_prog_args,"$<") 2>&1 | sed -n -e "s/Finished transaction in .*(\([0-9]*\.[0-9]*\)u.*)/\1/p" | head -1`; \
 	  R=$$?; times; \
 	  if [ $$R != 0 ]; then \
 	    echo $(log_failure); \
@@ -517,7 +514,7 @@ $(addsuffix .log,$(wildcard ideal-features/*.v)): %.v.log: %.v $(PREREQUISITELOG
 	@echo "TEST      $< $(call get_coq_prog_args_in_parens,"$<")"
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
-	  $(coqtopcompile) "$<" $(call get_coq_prog_args,"$<") 2>&1; R=$$?; times; \
+	  $(coqc) "$<" $(call get_coq_prog_args,"$<") 2>&1; R=$$?; times; \
 	  if [ $$R != 0 ]; then \
 	      echo $(log_success); \
 	      echo "    $<...still wished"; \
@@ -531,7 +528,7 @@ $(addsuffix .log,$(wildcard ideal-features/*.v)): %.v.log: %.v $(PREREQUISITELOG
 # Additional dependencies for module tests
 $(addsuffix .log,$(wildcard modules/*.v)): %.v.log: modules/Nat.vo modules/plik.vo
 modules/%.vo: modules/%.v
-	$(HIDE)$(coqtop) -R modules Mods -compile $<
+	$(HIDE)$(coqc) -R modules Mods $<
 
 #######################################################################
 # Miscellaneous tests
@@ -550,7 +547,7 @@ $(patsubst %.sh,%.log,$(wildcard misc/*.sh)): %.log: %.sh $(PREREQUISITELOG)
 	  echo $(call log_intro,$<); \
 	  export BIN="$(BIN)"; \
 	  export coqc="$(coqc)"; \
-	  export coqtop="$(coqtop)"; \
+	  export coqtop="$(coqc_boot)"; \
 	  export coqdep="$(coqdep)"; \
 	  export coqtopbyte="$(coqtopbyte)"; \
 	  "$<" 2>&1; R=$$?; times; \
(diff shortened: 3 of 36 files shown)


gares:
Can you explain that `-_option` means? Looking at the diff I see also `-o` and `-image` were changed.

I think the change makes sense, at least in principle, but all changes should be summed up here to ease reviewing.

gares:
Note that `-o` was added to make it possible to keep the source tree clean.
In particular `coqc a.v` generates `a.vo` next to `a.v` while with `-o` you can put into another directory of your choice (like `_build` for dune). It was added, after a long time, to help the integration with Eclipse, that *forces* an hygienic build.

ejgallego:
> Can you explain that `-_option` means? Looking at the diff I see also `-o` and `-image` were changed.

That's an internal option, `coqc` still needs to call `coqtop` with `-compile`, `-o` etc... so with this hack we avoid the deprecation warning. They should go away once `coqc` is a real binary not a a shim.

We could do a way more heavy choice here but IMHO it is not worth it.

ejgallego:
> Note that `-o` was added to make it possible to keep the source tree clean.

`-o` is still supported in `coqc`, deprecated in `coqtop`.

ejgallego:
So `-compile`, `-o`, `-batch`, etc... are deprecated in coqtop.

`-_compile`, `-_o` are internal alias for use by `coqc` until it stops being a shim [see the updated referenced PR].

SkySkimmer:
I feel like it would be cleaner to have an option `-is-really-coqc` for coqtop which disables the warning.

ejgallego:
> I feel like it would be cleaner to have an option `-is-really-coqc` for coqtop which disables the warning.

Maybe, I don't think it makes a lot of difference tho, and adds an extra ref / field.

ejgallego:
> Maybe, I don't think it makes a lot of difference tho, and adds an extra ref / field.

Note that the hack goes away in #8690 , and that PR seems pretty ready to me! [Only a minor nit remaining plus the merge of this one of course]

doc/sphinx/addendum/parallel-proof-processing.rst
```diff
@@ -154,10 +154,10 @@ to a worker process. The threshold can be configured with
...
 ---------------
 
-When |Coq| is used as a batch compiler by running `coqc` or `coqtop`
--compile, it produces a `.vo` file for each `.v` file. A `.vo` file contains,
-among other things, theorem statements and proofs. Hence to produce a
-.vo |Coq| need to process all the proofs of the `.v` file.
+When |Coq| is used as a batch compiler by running `coqc`, it produces
+a `.vo` file for each `.v` file. A `.vo` file contains, among other
+things, theorem statements and proofs. Hence to produce a .vo |Coq|
+need to process all the proofs of the `.v` file.
```
Zimmi48: Use double backticks throughout this paragraph. Same thing in the other chapter.

ejgallego: Thanks @zimmi48, updated.

ejgallego:
Travis failure is totally spurious, it passed in a previous version.

ejgallego:
Anyone would like to assign?

After porting some upstreams to use `coqc`, maybe we could leave the `-batch` undeprecated for `coqtop`, at least while we keep the `load-vernac-source` options available.

Zimmi48 (approved):

maximedenes:
> Anyone would like to assign?

I can take it.

ejgallego:
Thanks @maximedenes

herbelin:
To play the negative role, isn't there a risk to upset users by modifying too much the way basic commands used to work? How do we motivate the need for the change (rather than ensuring by ourselves that `coqc` and `coqtop` take consistent arguments on the command line)?

(But maybe `coqtop -compile` is not known a lot after all, and my remark is then not really relevant.)

ejgallego:
> To play the negative role, isn't there a risk to upset users by modifying too much the way basic commands used to work? How do we motivate the need for the change (rather than ensuring by ourselves that `coqc` and `coqtop` take consistent arguments on the command line)?
> 
> (But maybe `coqtop -compile` is not known a lot after all, and my remark is then not really relevant.)

My point of view is that current situation is more upsetting, in particular as we have quite a few bugs that #8690 solves and we have evidence users have hit them.

In fact, if you are a user now you would be quite confused with the sea of options we have, and even worse many possible combinations cannot work well together IMO. I think #8690 speaks by itself, but indeed the current codepath is at least questionable.

I don't think users will be impacted a lot; CI passed without too much trouble and the very few remaining nits are due indeed to problems of the current setup IMO.

ejgallego:
Thus my opinion after the experiment is that I believe this change is positive both in terms of code and in terms of usability.

ejgallego:
What options to support in `coqtop` is indeed a bit open. We could support batch, or have `-load-source` imply it. There is a bit of design space mainly due to the need to test `coqtop`.

But that's IMO a minor design issue, even if we should get it right.

herbelin:
Having a robust and consistent set of command line options per executable is indeed a convincing argument. There are uses of `-compile` and `-batch` around though (e.g. in HoTT).

ejgallego:
Also in `iris`. HoTT used `coqtop -compile` just because Coq did for the Prelude, but there is already a PR removing that. `-batch` may be indeed worth keeping. `-compile` is the flag that gives us all the gain.

ejgallego:
Ok, I have undeprecated `-batch` as it may be convenient and creates no problem.

This is ready to go IMO.

Zimmi48:
>-compile is the flag that gives us all the gain.

Did you mean pain? :laughing:

ejgallego:
> Did you mean pain?

Indeed I mean removing it.

ejgallego:
IMO we should get this and the depending PR in soon as it will make quite a few things clearer for example in regards to recent work on STM caching, etc...

ejgallego:
Ping!

Zimmi48:
This needs a rebase now.

maximedenes (approved):

maximedenes (dismissed):
`make states` should not trigger the warning...

ejgallego:
> `make states` should not trigger the warning...

Oh indeed, good catch! Should be fixed now.

maximedenes:
@ejgallego After thinking more, I'm not sure I understand exactly what this PR brings. I understand what #8690 can bring, and I support its principle.

Can you say again what is the motivation of this one? I reread the description above, but it seems to explain the merits of #8690 instead, so I'm simply a bit lost.

ejgallego:
What this PR does is to split the migration from `coqtop -compile` to `coqc` in two steps.

It makes sense as to better separate two kind of regressions, and ease reviewing. Regressions in class 1 are the ones that happen by moving from `coqtop -> coqc`, then, regression in class 2 are the ones that happen when `coqc` code is actually modified.

This PR used to be included in #8690 but some experience showed that this is better done in 2 steps.

maximedenes:
> It makes sense as to better separate two kind of regressions, and ease reviewing

I'm not fully convinced by that, since one can review / test intermediate commits. On the other hand, this methodology is not a problem, as long as we make sure we integrate the two PRs and not just this one. Similary undesirable intermediate states have happened in the past.

So I guess the question is whether #8690 is ready, so that we can merge the two.

ejgallego:
This is a matter of risk; it makes sense to deprecate the options now; the state this PR introduces is not "undesirable". Having the deprecation in place is IMO good, regardless if we can yet reap the benefits from it.

It looks like #8690 is ready or very close to be, but indeed, it may need more testing / feedback from the few other issues. For example discussion with IRIS devs resulted in changes to both PRs.

maximedenes:
> Having the deprecation in place is IMO good, regardless if we can yet reap the benefits from it.

I would tend to think otherwise. If we can't actually take advantage of the deprecation, I see only disadvantages to this change.

> It looks like #8690 is ready or very close to be, but indeed, it may need more testing / feedback from the few other issues.

I'm confident we can make it work, and maybe that's were we should take a small risk (after some testing of course).

ejgallego:
We will eventually take advantage of the deprecation, that's the whole point of a deprecation.

I am afraid I don't understand what you mean, sorry. Deprecation is one thing; modification of the code is other, which usually done later.

ejgallego:
Like when we deprecate an API we remove it on the next release.

We could well deprecate this setting in 8.10, but only merge #8690 in 8.11.

ejgallego:
Even tho I think #8690 provides enough backward compatibility as to be shipped in 8.10, but that's still to see.

maximedenes:
> Deprecation is one thing; modification of the code is other, which usually done later.

I understand the analogy, but the situation is a bit more complex here, since we internally still rely on the deprecated API. That's why you had to introduce `-_compile`.

I think it would be highly undesirable that this intermediate state stays. What would we do next, if not #8690? Deprecate `-_compile`?

Essentially here, you are deprecating an API by aliasing it to something undocumented and deprecating the former name. I find the approach wrong, and I don't see what it brings other getting #8690 ready first, since it can still emulate the old API (that we can then deprecate).

ejgallego:
> I understand the analogy, but the situation is a bit more complex here, since we internally still rely on the deprecated API. That's why you had to introduce `-_compile`.

Which is a very common thing to do as long as the API cannot be removed, don't you think?

That's the whole point of deprecation, to mark a feature as don't use _while not removing it_.

Again, nothing guarantees that #8690 is going to be compatible enough to be in 8.10; we just hope so, but imagine it has to be reverted? That would be fine, but the deprecation would have to stay.

maximedenes:
> That would be fine, but the deprecation would have to stay.

So you are saying that we should judge this change (this PR) as if it could stay?

maximedenes:
> Which is a very common thing to do as long as the API cannot be removed, don't you think?

I thought we never deprecated APIs we still relied on. Is that incorrect?

ejgallego:
> I thought we never deprecated APIs we still relied on. Is that incorrect?

That is incorrect, as unfortunately it is not always possible to keep the deprecated stuff working. A recent example is the `Coqlib` PR, which was merged with not all the codebase ported.

Note that we don't rely on the deprecated API in the places that matters, that is, not call to `coqtop -compile` is done in Coq, except for the internal implementation detail.

In this case, as we are talking about a binary name things are ugly, if you have a better idea on how to deprecate `-compile` please tell so and we can implement it.

maximedenes:
> So you are saying that we should judge this change (this PR) as if it could stay?

So I guess I will follow others' opinion. If @gares is ok with the intermediate state and @Zimmi48 is ok with having two undocumented flags (`-_compile` and `-_compile-verbose`) that may stay, I'll merge.

ejgallego:
I mean, this is the way I came up to implement deprecation of `-compile` while actually keeping the old semantics for people that use; it you have folks a way to do that that is less ugly I'm all for it.

ejgallego:
BTW calling this an "intermediate state" seems misleading; this is really the same state than `master`, just with an added warning.

maximedenes:
> BTW calling this an "intermediate state" seems misleading; this is really the same state than `master`, just with an added warning.

and two undocumented options.

I called it intermediate states because of the motivation you gave: "What this PR does is to split the migration from `coqtop -compile` to `coqc` in two steps."

Anyway, no point on arguing more, I'll wait to hear from @gares and @Zimmi48 and follow their opinions.

ejgallego:
> I still don't see what is wrong with "intermediate", sorry.

Well you said "Similary undesirable intermediate states have happened in the past." The wrong part is here is the _undesirable_

This deprecated state is undesirable as much as any other deprecation state is not bad.

I think you are missing the point of `-_compile`, this is a flag that is forbidden to use by any client except `coqc`, so of course it needs no deprecation phase as you propose.

Zimmi48:
I'm afraid I side with @maximedenes on this one. Since the only reason this PR could have to be merged independently from #8690 is that there is a risk the later cannot be merged without a deprecation phase first, then let's wait that #8690 is complete before taking any decision on this one. Ideally, #8690 can be merged in 8.10 and we can close this one and never introduce these two undocumented options. Otherwise, it will always be possible to reconsider merging the present PR when 8.10 gets closer.

Zimmi48:
>this is a flag that is forbidden to use by any client except coqc

Is there any mechanism in place to enforce that?

ejgallego:
@Zimmi48 I think you are missing that problems can be discovered much after merging. So if we don't have a deprecation phase, and then we discover a fatal compatibility flaw on the compat layer of #8690 let's say in the beta, things become harder.

I don't see the harm in having a deprecation phase other than the ugly internal implementation detail. But I'm afraid that the really ugly (and buggy) thing was the implementation of `coqc` in the first place, so this PR can do little to avoid that while remaining faithful.

> Is there any mechanism in place to enforce that?

Common sense? I mean, you would have to read the code to discover this flags and do _against_ what is written which is "do not use"

gares:
I think you should merge the two PR and avoid the intermediate state because, as you say

> other than the ugly internal implementation detail.

and I don't want' this to happen.

ejgallego:
> and I don't want' this to happen.

Well given the tradeoffs I think having this is not so bad, after all, the ugly implementation detail is already in `master`, the PR just hides the `coqtop` part from the user.

IMHO no harm on merging the two PRs and having indeed better debugging possibilities.

toplevel/coqargs.ml
```diff
@@ -322,6 +322,9 @@ let usage batch =
   then Usage.print_usage_coqc ()
   else Usage.print_usage_coqtop ()
 
+let deprecated_coqtop opt_name =
+  Feedback.msg_warning Pp.(seq [str "The option "; str opt_name; str" is deprecated, please use coqc"])
```
SkySkimmer: Why isn't this using cwarnings? Then we could just pass `-w -deprecate-compile-arg` before -compile in coqc.

ejgallego: I didn't use `cwarnings` as in a sense, this warning should not be allowed to be disabled; but indeed no problem in doing what you propose.

maximedenes:
> So @maximedenes let's stop bikeshedding and let's go ahead with this PR as it is blocking important things. It seems 3 options have been proposed:
> 
> 1. Merge this PR, then #8690 ,
> 2. Remove the deprecation phase and thus implement #8690 directly ,
> 3. A different deprecation method?
> 
> What should we do? As the main toplevel maintainer I strongly support 1 , as it has no downsides.
> 
> I am OK with 3 too, but then please specify what the different method should be.

I like 2, since IIUC no intentional semantic changes are made in #8690. Is that incorrect?

ejgallego:
> I like 2, since IIUC no intentional semantic changes are made in #8690. Is that incorrect?

#8690 does have semantic changes, how many it is hard to predict, as the execution model changes.

But for example, it fixes some bugs related to command line flags, and it disallows some combinations that before were allowed but with nonsensical or undefined behavior.

So indeed it only provides compatibility for the most common cases [that should hopefully cover 100% of the users]

ejgallego:
Today you can do very weird things if you want to use weird command line flags combinations.

gares:
What is wrong in merging #8690 immediately? I'd like to be sure that we will not end up with this one merged and the other one not.

ejgallego:
> What is wrong in merging #8690 immediately?

Apart from the problems outlined above, now such PR has bitrotten [due to the delay] and while it used to pass CI it does not now and will require _significant_ work to make it work again.

> I'd like to be sure that we will not end up with this one merged and the other one not.

I still don't see what is the problem with having this PR in, it could be in a release and it wouldn't be a problem at all, tho likely we would like to ship the more elegant scheme proposed by @SkySkimmer

gares:
> due to the delay

I believe that you have a perception of the PR procedure that is just wrong, and makes the process exhausting for both you and the reviewers.

You cannot expect to be able to dismiss the reviews like that, accusing of bike shedding, repeating over and over your arguments as if the solution you propose was perfectly fine while 4 people are suggesting otherwise, and now claiming that the delay they provoked made the option that they propose not viable.

It just does not work like that.

ejgallego:
> I believe that you have a perception of the PR procedure that is just wrong, and makes the process exhausting for both you and the reviewers.

My view of the PR procedure is exclusively based on how much of my time is perceived to be spent on non-productive discussion / waiting for action. There are several causes for non-productive discussion including getting feedback that is wrong / inaccurate, or insistence on implementing solution with a bad trade-off on harm/benefit, or discussion on non-technical grounds.

I do highly value my time and the time of other developers so I try to find the best tradeoff to help them work efficiently, and more importantly, I trust them to know what they are doing.

Now going back to the technical topic, you review has not been dismissed, it just that your commentary is not technically accurate as explained above. You don't want this PR in, however I am still waiting for the _why_ this PR should not be in [maybe with @SkySkimmer modifications].

ejgallego:
Summarizing your arguments, you don't want "something ugly to happen" [even if @SkySkimmer proposed a more elegant way]

So I should lose 20 hours of my time because you find something that is already used in Coq "ugly". Sorry it doesn't work like that.

Well, I could buy it if the solution you proposed could work, but indeed, you propose we skip a deprecation commit and directly implement a solution that has quite different semantics in many fronts; I would be OK doing that if it wasn't for the reasons explained above including that the PR has bitrotten.

Now, I have to invest maybe 2 hours on fixing a PR that was ready to be merged [it is what you propose in fact], however it was not merged due to inaction from others.

So if that is your PR model then let me state very clearly I disagree with it.

maximedenes:
> that has quite different semantics in many fronts

This is what I don't understand. What are these *intended* semantic changes that are *not* bug fixes, or assigning meaning to previously meaningless combination of flags?

We usually don't deprecate if we don't have such changes in or ready to get in.

ejgallego:
> This is what I don't understand. What are these _intended_ semantic changes that are _not_ bug fixes, or assigning meaning to previously meaningless combination of flags?
> 
> We usually don't deprecate if we don't have such changes in or ready to get in.

There are 2 things I think:

- the semantics are different in some cases [is that clear?]
- what are the effects? This seems very hard to predict to me until we get #8690 in; I already saw very strange calls to coqtop / coqc from a few developments [see the IRIS / HoTT] issue so it is hard to tell how people is using the options.

So this approach seems to have a marginal benefit over going back to merging the 2, it could even be made pleasant using Gaëtan proposal. I wouldn't even mind re-merging with #8690 if it hadn't bitrotten.

However what I lack is a reason to actually do more work than needed. Requiring a developer to do many hours of work for no reason is just disrespectful and at best unconsiderate of the developer time IMHO.

From my point of view, I will save a considerable amount of time following the path that has been already set up. And so far none of you have provided a single valid technical reason on why I should spend an extra 2 hours to go with the other approach when in fact it is inferior from a engineering point of view.

I apologize if indeed you provided a reason and I am failing to understand it; it could be well the case; but after re-reading the whole thread I fail to see it.

maximedenes:
> so far none of you have provided a single valid technical reason
[...]
> when in fact it is inferior from a engineering point of view

At least we agree on the fact that we are wasting our time. I don't think it makes a lot of sense to ask for reviews when you have such postulates.

Just do what you want.

maximedenes:
> Regarding your reaction I must say that I find this kind of "retaliation" (or should I call it punishment) not the best thing to proceed, of course you are free to do so, but I wouldn't mind having an environment where I can be honest and express what I think without being afraid of being bad-mouthed or punished to be abandoned.

Just to be clear, there is no "retaliation" of any sort, I'm not even sure what you had in mind.

I'm just drawing the conclusion that I can't constructively handle this PR, so I'll let others do so.

ejgallego:
Ok, thanks for you help anyways. Any volunteers? @herbelin, maybe this falls into your are of knowledge?

I will implement Gaetan suggestion and I think this should be ready to go in, or if you also think that a deprecation phase is counterproductive I guess we should close .

ejgallego:
> Just to be clear, there is no "retaliation" of any sort, I'm not even sure what you had in mind.

Whether your intention was to retaliate or just are tired of this very boring PR [which should have been merged long time ago], the net effect is harm on me as `a)` I must spend now time finding another person and explaining the delicate details of this kind of work, including my experiments on porting the full CI to `coqc` and what I found while doing so, activity that if you read the original PR you will discover is what motivated this one; `b)` I was still not given an argument on why this solution is not the best.

Not that I would like to scare assignees more than they seem to be, but I would recommend to discourage people undertaking PR assignments if a certain level of commitment is not available.

Makefile.build
```diff
@@ -195,7 +195,7 @@ TIMER=$(if $(TIMED), $(STDTIME), $(TIMECMD))
 #   TIME="%C (%U user, %S sys, %e total, %M maxres)"
 
 COQOPTS=$(NATIVECOMPUTE) $(COQWARNERROR) $(COQUSERFLAGS)
-BOOTCOQC=$(TIMER) $(COQTOPBEST) -boot $(COQOPTS) -compile
+BOOTCOQC=$(TIMER) $(COQTOPBEST) -boot $(COQOPTS) -_compile
```
SkySkimmer: Why not use coqc here btw?

ejgallego: That's delayed to #8690 as to avoid modifying the make rules too much and thus introducing a regression here. In current `master`, `COQC` doesn't depend on `coqtop` (!!) so if we modify the rule `coqc` will be called right away and fail as `coqtop` has not been build.

As I wanted to introduce the least amount of functional changes in this PR I didn't perform this deep make modification here; if we end up skipping this PR then this would be solved.

ejgallego: That's indeed another bug of the make system I think.

ejgallego: Same happens for the Dune rules, etc...

ejgallego: Done.

ejgallego:
> You are right, as usual. There is no point in reviewing your PRs. I said that before, then I started doing it again because I consider it my job. I was wrong in doing that, sorry for the time I made you lose.

I wish you would, indeed, have spared me from your comment. As I said in some other thread I do not welcome this kind of personal attacks, and as far as I understand this ironic tone is not something encouraged by Coq's CoC, neither a good thing for the community.

"Los trapos sucios se lavan en casa"

It seems to me that reviews are for technical feedback and commentary, and I do very much welcome any such feedback from anyone. However, I am very concerned by the personal vs technical ratio of your comments.

Zimmi48:
>the more I repeat the point the more frustrated I become

What you may be missing is that it is not just you that becomes frustrated: it is everyone. Repeating your point *ad nauseam* has never been an efficient way of convincing people. However, in the context of GitHub, that makes for very long threads, very hard to follow and discourage people from helping. Sometimes they may end the discussion with "fine, I don't care too much and I'll merge your PR". But that's not to be seen as winning the argument.

You are both criticizing the people that fail to review PRs promptly and the ones that do comments you don't agree with / don't understand. There is a kind of contradiction here. Furthermore, we cannot have our cake and eat it too: PR reviewing most often improves code quality and is also a very valuable tool for sharing knowledge; if in some rarer cases, reviews are deemed nonconstructive, this is no good reason to attack reviewers. I agree that Enrico might not be the most patient person with you and this regularly leads to unpleasant outburst. But I too find your behavior exhausting from time to time. This is compensated by the very high value of your contributions, your enthusiasm and the pleasant interactions I have with you, so I try to forget about the part of your behavior that I don't like. But honestly, I can understand Enrico's lack of patience. (All of this while considering a possibility that you may be the one that is right both in this particular discussion and in all the others.)

ejgallego:
Thanks for your feedback @Zimmi48 ; I do agree that I do have to improve, and this is mostly my fault as I like to undertake really disturbing PRs [such as replacing the build system] that are not easy to review.

> What you may be missing is that it is not just you that becomes frustrated: it is everyone.

But I am the one to pay the highest price by far; as I have to invest _significantly more time_ to make the PR ready; I am sorry but here a lot of burden is put on the PR author to justify his decision in front of people who is less informed about the topic / subsystem. On top of that, the burden is then payed back with punishment. It doesn't seem fair to me.

> Repeating your point _ad nauseam_ has never been an efficient way of convincing people.

I am afraid I cannot share this view wrt what's happened in this PR. I'd say that the FUD spread here had its effect; after review of the PR, I can summarize what happened here as:
- based on my understanding, and analysis of contribs, I propose a two-steps approach;
- the approach is questioned on wrong assumptions, I do clarify all of them [this is the bulk of the discussion]
- after that, the approach is still rejected , overriding _trust_ on the maintainer of the subsystem itself.
- I _do_ complain that nobody produced a technical argument as to why the one-step approach is better

So indeed, if I ask for a justification, and the justification is not produced, I can hardly move my position. Note the justification has not been produced yet. Surprisingly, when I insisted on the _why_, people disappeared. This is a pattern it seems.

This is not about winning, is about technical review.

I do agree that these long discussions are tiring and I do share a big part of responsibility for it; but on one hand I am not gonna allow people spreading false FUD about me or about my work, and on the other I am not gonna accept an inferior solution just because I am told so. Produce a justification and I will happily modify the approach.

> You are both criticizing the people that fail to review PRs promptly and the ones that do comments you don't agree with / don't understand.

I don't have the view that I do criticize people, but indeed I cannot support a model where people have very little consideration/respect for developer's time. It seems we have forgotten that the goal of PR review is to _help_ the author, not to block them / make their life hard. So of course I am very critic with unhelpful or inaccurate reviews, it is a matter of respect. [And I do have quite a few examples]

My main critic here is that I asked questions, and such questions were never answered, instead, I got served by inappropriate behavior. Do you consider that as an unfair critic? In particular it is disappointing my questions were not answered after I had to lose a lot of time answering the ones given to me.

This is not a new situation as indeed for example Hugo has repeatedly complained about his questions not being answered and retaliation. IMHO the affirmation "you are both criticizing ..." is a mischaracterization.

I am surprised you go to great lengths to justify others inappropriate behavior, including personal attacks and selective, bad-faith quoting. IMO this kind of behavior should not be encouraged. If you cannot work on something there are other ways to get out.

Also I am surprised by the "compensated by" line of thinking. I am failing to work properly that cannot be justified by other good things I do. I should indeed correct the thing. But I don't think it was me who derailed this PR by not listening to the author's arguments.

> All of this while considering a possibility that you may be the one that is right both in this particular discussion and in all the others

I don't care about being right, I like to be wrong more, as that means I am gonna learn something [imagine playing chess and always winning].

Anyways, thanks for trying to mediate, even if I am afraid I cannot embrace most of the views you expose I do really appreciate you spending some minutes trying to help.

Zimmi48:
You are talking about respect, and I was precisely coming back to this thread to add a word about respect. I agree with you that the sarcastic tone that was employed earlier is strongly discouraged by the code of conduct. However, something else that the code of conduct says is "respect the work of others". This should be understood in the large sense. Work does not just include past and present code contributions, it also includes reviews (and I'm sure you agree with me as you have raised the problem of lack of credit for review work). You may think that there is a lack of respect for developer's time but realize that quickly dismissing a review is also showing lack of respect for reviewer's time. Correlated with the lack of credit for review work mentioned earlier, this is the kind of things that can discourage reviewers. **Reviewing is a hard task**: you have to switch context and try to understand a part of the code that you sometimes know less well than the PR author. It can happen that the review resulting from this effort is bad. But it is not a good reason to dismiss it quickly with disdain.

---

BTW, you were talking about bikeshedding, but sometimes it is also OK to realize that we don't know what the best course of action is and that delaying a decision might be the best option available at the current time.

ejgallego:
I don't think review work was disrespected here, quite the contrary, I did an effort to please the reviewers _way_ far beyond of what the triviality of the PR deserves, sorry. Neither any review was quickly dismissed and much less with disdain, so I have a problem mapping your comment to the actual facts in this PR.

I highly respect and appreciate the work of the reviewers, however I think I can call a horse a horse, I insist here the PR was blocked but the proper technical arguments were not given.

Note that I don't consider "bikeshedding" a pejorative word, the way I understand it is more on a funny way "we were bikeshedding the whole night XD" but I could be of course wrong, thus in this case I apologise.

I have to thank you again for willing to discuss and provide feedback, which is very appreciated, however I cannot yet share / relate some of your views with the actual facts in this PR. In no case delaying this trivial PR whose parents has been open for months is the best course of action, much less without providing a reason.

Zimmi48:
Right. Bikeshedding it is. There was no technical argument for not merging this PR first except that it looks like giving up to bad practice (introducing undocumented options in particular). So it would have been easy to just avoid the bikeshedding and merge this PR.

On the other hand, you did not provide much technical justification against the alternative of merging only #8690 either. Your two arguments against it were: it has bitrotten due to long review process (or actually the lack of review), and it would be good to have #9095 in even if #8690 does not make it into 8.10 (which is not clear for everyone else). Even though you only removed the “needs: progress” flag on #8690 three weeks ago when you created this PR, I can only agree that the review process has been too long and thus I can perfectly understand your frustration. Regarding the bitrotting aspect, there is no syntactic conflicts, only semantic ones (I suppose) which make the PR fail everywhere in CI. However, it's the case of both #8690 and #9095. How do you assess that #8090 will need significantly more work while #9095 will be easy to fix and merge?

(Often, misunderstanding and bikeshedding come from a lack of outsiders' knowledge of your internal thought process. For instance, in #8690 you initially said “We could have a deprecation phase, however I am not sure it is worth the cost”. You changed your mind afterwards but you never wrote down the reason that had pushed you to change your mind, so no one else that you really knows.)

ejgallego:
Thanks again @Zimmi48 , I think you did a pretty good summary of what I consider the facts to be, including serious problems in providing coherent information on PRs. Indeed I should be more careful on information presentation, but yeah Github still doesn't seem to have the optimal interface for that.

I do agree that my main reason to have this was a reason of developer efficiency; given that this solution is weakly better than a single commit; I did prefer the one that optimized my time use as I didn't perceive an improvement; also, the more I thought about the implications of the deprecation the less comfortable I was with no having it.

ejgallego:
Ok so this PR should be ready [with @SkySkimmer suggestion implemented]

Given the comment at https://github.com/coq/coq/pull/9095#issuecomment-447948288 it seems to me that I should update the `toplevel` entry in the `CODEOWNERS` file, does anyone volunteer to take assign here / help with toplevel stuff?

ejgallego:
Pending on https://github.com/coq/coq/issues/9419 , which seems like a recent regression.

ejgallego:
For the rest, very interesting failures on the async part:
```
==========> TESTING bugs/closed/bug_4366.v <==========
5291:proofworker:0:0] Slave: critical exception: Timeout!
Error: Worker dies or task expired

0m0.000000s 0m0.000000s
0m1.210000s 0m0.130000s
    bugs/closed/bug_4366.v...Error! (bug seems to be opened, please check)

==========> TESTING output/FunExt.v <==========
--- output/FunExt.out	2018-11-26 21:14:04.361602651 +0100
+++ output/FunExt.out.real	2019-01-28 01:47:57.086036375 +0100
@@ -9,10 +9,10 @@
 Tactic failure: Not an extensional equality.
 The command has indeed failed with message:
 Ltac call to "extensionality in (var)" failed.
-Tactic failure: Not an extensional equality.
+Tactic failure: Already an intensional equality.
 The command has indeed failed with message:
 Ltac call to "extensionality in (var)" failed.
-Tactic failure: Already an intensional equality.
+Tactic failure: Not an extensional equality.
 The command has indeed failed with message:
 In nested Ltac calls to "extensionality in (var)" and
 "clearbody (ne_var_list)", last call failed.
0m0.000000s 0m0.000000s
0m2.450000s 0m1.200000s
    output/FunExt.v...Error! (unexpected output)

==========> TESTING output/Show.v <==========
--- output/Show.out	2018-11-26 21:14:04.368602728 +0100
+++ output/Show.out.real	2019-01-28 01:47:55.383978759 +0100
@@ -8,3 +8,4 @@
  1 = S (S m')
 subgoal 3 (ID 22) is:
  S (S n') = S m
+Debug: 7331:proofworker:0:0 STM: sending back a fat state
0m0.000000s 0m0.000000s
0m0.240000s 0m0.150000s
    output/Show.v...Error! (unexpected output)

==========> TESTING output/unifconstraints.v <==========
--- output/unifconstraints.out	2018-12-30 16:49:16.119751131 +0100
+++ output/unifconstraints.out.real	2019-01-28 01:47:50.347807592 +0100
@@ -1,3 +1,11 @@
+The command has indeed failed with message:
+In environment
+P : nat -> Type
+x : nat
+h : P x
+Unable to unify "P x" with "?P x"
+(unable to find a well-typed instantiation for "?P": cannot ensure that
+"nat -> Type" is a subtype of "nat -> Prop").
 3 focused subgoals
 (shelved: 1)
   
@@ -13,6 +21,7 @@
    True /\ True /\ True \/
    veeryyyyyyyyyyyyloooooooooooooonggidentifier =
    veeryyyyyyyyyyyyloooooooooooooonggidentifier
+Debug: 6269:proofworker:0:0 STM: sending back a fat state
 3 focused subgoals
 (shelved: 1)
   
@@ -63,11 +72,6 @@
    True /\ True /\ True \/
    veeryyyyyyyyyyyyloooooooooooooonggidentifier =
    veeryyyyyyyyyyyyloooooooooooooonggidentifier
-The command has indeed failed with message:
-In environment
-P : nat -> Type
-x : nat
-h : P x
-Unable to unify "P x" with "?P x"
-(unable to find a well-typed instantiation for "?P": cannot ensure that
-"nat -> Type" is a subtype of "nat -> Prop").
+Debug: 6410:proofworker:1:0 STM: sending back a fat state
+Debug: 6410:proofworker:1:0 STM: sending back a fat state
+Debug: 6410:proofworker:1:0 STM: sending back a fat state
0m0.000000s 0m0.000000s
0m0.420000s 0m0.200000s

==========> TESTING complexity/constructor.v <==========
0m0.000000s 0m0.000000s
0m0.560000s 0m0.080000s
    complexity/constructor.v...Error! (couldn't find a time measure)
```

gares:
Looking at `output/FunExt.v` it seems the output is non deterministic since each proof prints one of these lines and depending on:
- how many workers
- which proof is delegated or run in master
you may get two proofs run in any order.

I guess we should disable -async-proofs on these.

gares:
The messages `Debug: 7331:proofworker:0:0 STM: sending back a fat state` is a "bug" since in stm.ml around line 1600 I have
```          msg_debug (Pp.str "STM: sending back a fat state");
```
and now that you don't pass `-batch` `Flags.quiet` is now set to false and we get this printed out.
I suggest you guard the `msg_debug` call with a test on `-debug`.

gares:
Test `output/unifconstraints.v` seems to show both problems.

ejgallego:
> I suggest you guard the `msg_debug` call with a test on `-debug`.

Indeed I think it should be that way. I guess we want to have this behavior on the `-quiet` flag, right?

ejgallego:
Results are slightly better after correcting the debug output:
```
    output/FunExt.v...Error! (unexpected output)
    output/unifconstraints.v...Error! (unexpected output)
    complexity/constructor.v...Error! (couldn't find a time measure)
    complexity/f_equal.v...Error! (couldn't find a time measure)
    complexity/injection.v...Error! (couldn't find a time measure)
    complexity/ring.v...Error! (couldn't find a time measure)
    complexity/ring2.v...Error! (couldn't find a time measure)
    complexity/setoid_rewrite.v...Error! (couldn't find a time measure)
    complexity/unification.v...Error! (couldn't find a time measure)
```
dunno what's going on with the complexity tests, is that maybe #9422

gares:
I think you fixed the fat state issue too hard, since the "no time measure" error is that test-suite/Makefile cannot grep `Finished transaction ..` that is what `Time` prints i guess.

ejgallego:
> I think you fixed the fat state issue too hard, since the "no time measure" error is that test-suite/Makefile cannot grep `Finished transaction ..` that is what `Time` prints i guess.

This was happening before the fat message change, anyways I only guarded that particular call to msg_debug o it shouldn't matter for that case.

ejgallego:
So indeed maybe the verbosity options are badly set on the workers? But then, why it would work on the regular test-suite?

gares:
> So indeed maybe the verbosity options are badly set on the workers?

I think you removed `-batch` that was setting some verbosity options.
IIRC the options of master are passed on to workers, but there is some mangling there:
  https://github.com/coq/coq/blob/master/stm/asyncTaskQueue.ml#L121
(it seems `-batch` would have been passed, but `-compile` would be correctly dropped).

gares:
So yes, `-batch` was having an effect on master only, and you have removed it.
Can this explain the difference?

ejgallego:
> Can this explain the difference?

Yeah I guess that's it, but indeed the previous situation looks like a mess, in theory `coqc` should not be different from `coqtop -batch` but however it is.

ejgallego:
Yeah I dunno how to fix this, the filtering on the workers is not very consistent due to the coupling of `-batch` and `-verbose` [one is filtered the other is not]

I guess I'll just special-case complexity.

ejgallego:
Umm the complexity tests are still failing; I wonder what's going on.

test-suite/Makefile
```diff
@@ -479,14 +476,15 @@ $(addsuffix .log,$(wildcard interactive/*.v)): %.v.log: %.v $(PREREQUISITELOG)
...
+# time is a 6120 bogomips cpu. We need to pass -verbose as in the
+# async case workers won't print the Timeout info otherwise.
 ifneq (,$(bogomips))
 $(addsuffix .log,$(wildcard complexity/*.v)): %.v.log: %.v $(PREREQUISITELOG)
 	@echo "TEST      $< $(call get_coq_prog_args_in_parens,"$<")"
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
 	  true "extract effective user time"; \
-	  res=`$(call get_command_based_on_flags,"$<") "$<" $(call get_coq_prog_args,"$<") 2>&1 | sed -n -e "s/Finished transaction in .*(\([0-9]*\.[0-9]*\)u.*)/\1/p" | head -1`; \
+	  res=`$(coqc_boot_interactive) -verbose "$<" $(call get_coq_prog_args,"$<") 2>&1 | sed -n -e "s/Finished transaction in .*(\([0-9]*\.[0-9]*\)u.*)/\1/p" | head -1`; \
```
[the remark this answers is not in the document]
ejgallego: So what happens now is that the test-suite is just ignoring the -async stuff for these targets.

test-suite/Makefile
```diff
@@ -307,7 +304,7 @@ $(addsuffix .log,$(wildcard prerequisite/*.v)): %.v.log: %.v
 	    echo "    $<...correctly prepared" ; \
 	  fi; \
 	} > "$@"
-	@echo "CHK $(shell basename $< .v)"
+	@echo "CHCK      $(shell basename $< .v)"
```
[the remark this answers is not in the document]
ejgallego: It is a shorter name for `CHECK`, but indeed we can use `CCHK` or `QCHK`, I dunno.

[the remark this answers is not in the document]
SkySkimmer: Regardless please don't put this change in an unrelated PR

ejgallego:
Finally green again cc @maximedenes

stm/vernac_classifier.ml
```diff
@@ -161,7 +161,6 @@ let classify_vernac e =
     | VernacExistingClass _ | VernacExistingInstance _
     | VernacRegister _
     | VernacNameSectionHypSet _
-    | VernacDeclareCustomEntry _
```
maximedenes: Spurious change?

ejgallego: Oh, this is coming from #9421 which was necessary for the CI to pass, but it is now merged on master.

ejgallego: I've rebased just in case. By the way I saw a few spurious failures of the test suite, mainly on async.

gares: I sometimes see `auto.v`, did you see another one?

ejgallego: Some `open_3xxx` I think.

maximedenes (approved):

ejgallego:
I did a last review myself and found a small stale change in `CoqMakefile.in` [removal of `-batch`] which shouldn't be included. Fixed and re-pushed.

maximedenes:
Seems ready now. Will merge this and #8690.

PR #9101, opened 2018-11-27
Fix 8922 again

### LINKED ISSUE #8922: Uncaught exception Pp_diff.Diff_Failure

### LINKED ISSUE #9091: Proof diff in mode "added and removed" duplicates some hypotheses
#### Version

master

#### Operating system

Fedora 29

When multiple hypotheses have the same type and they all disappear using one tactic, they are wrongly duplicated in mode "added and removed".

For instance,

```
Goal forall n m : nat, n = m.
Proof.
  intros. revert n m. (* here n m : nat is shown twice *)
```

jfehrle:
**Kind:** bug fix

Fixes #8922
Fixes #9091 
Fixes incorrect handling of STRING and FIELD tokens, which would generate errors

In diffs, goals that are unmatched because they don't appear in the Show Proof structure will be highlighted as if they were new.

NOTE: This should be merged into 8.9 after https://github.com/coq/coq/pull/8967.

### DIFF
--- a/ide/idetop.ml
+++ b/ide/idetop.ml
@@ -219,7 +219,7 @@ let goals () =
         | Some oldp ->
           let (_,_,_,_,osigma) = Proof.proof oldp in
           (try Some { it = Evar.Map.find ng diff_goal_map; sigma = osigma }
-          with Not_found -> raise (Pp_diff.Diff_Failure "Unable to match goals between old and new proof states (6)"))
+          with Not_found -> None) (* will appear as a new goal *)
         | None -> None
         in
         let (hyps_pp_list, concl_pp) = Proof_diffs.diff_goal_ide og_s ng nsigma in
--- a/parsing/tok.ml
+++ b/parsing/tok.ml
@@ -36,12 +36,24 @@ let equal t1 t2 = match t1, t2 with
 | EOI, EOI -> true
 | _ -> false
 
-let extract_string = function
+let extract_string diff_mode = function
   | KEYWORD s -> s
   | IDENT s -> s
-  | STRING s -> s
+  | STRING s ->
+   if diff_mode then
+    let escape_quotes s =
+      let len = String.length s in
+      let buf = Buffer.create len in
+      for i = 0 to len-1 do
+        let ch = String.get s i in
+        Buffer.add_char buf ch;
+        if ch = '"' then Buffer.add_char buf '"' else ()
+      done;
+      Buffer.contents buf
+    in
+    "\"" ^ (escape_quotes s) ^ "\"" else s
   | PATTERNIDENT s -> s
-  | FIELD s -> s
+  | FIELD s -> if diff_mode then "." ^ s else s
   | INT s -> s
   | LEFTQMARK -> "?"
   | BULLET s -> s
--- a/parsing/tok.mli
+++ b/parsing/tok.mli
@@ -22,7 +22,8 @@ type t =
   | EOI
 
 val equal : t -> t -> bool
-val extract_string : t -> string
+(* pass true for diff_mode *)
+val extract_string : bool -> t -> string
 val to_string : t -> string
 (* Needed to fit Camlp5 signature *)
 val print : Format.formatter -> t -> unit
--- a/printing/printer.ml
+++ b/printing/printer.ml
@@ -722,11 +722,11 @@ let pr_subgoals ?(pr_first=true) ?(diffs=false) ?os_map
   let get_ogs g =
     match os_map with
     | Some (osigma, _) ->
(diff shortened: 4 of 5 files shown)


ppedrot (commented):
Not confident to comment on that code.

jfehrle:
Thank you, sir!

eponier:
I confirm that https://github.com/coq/coq/issues/9091 is solved by this PR. There is still a slightly odd behaviour illustrated by the following example:
```

Goal True.
Proof.
  remember 2 as m. remember 3 as n.
  move n at top.
```
In normal view mode, we get one line `n, m : nat`, while in diff modes, we get two separate lines `n : nat` and `m : nat` that fusions after the next tactic. It's a minor point, and I think I have already seen such a behaviour elsewhere in Coq, so we can probably ignore it.

For https://github.com/coq/coq/issues/8922, there is still a bug, but a different one.

```
Goal True.
Proof.
  eassert (_).
  intros x. intros y.
```
gives:
`Anomaly "Uncaught exception Invalid_argument("List.iter2")."`

To be clear, all these experiments were done in CoqIDE.

herbelin:
@jfehre: do I merge this as it is now or would you prefer to add something to this PR in relation to @eponier's comment?

jfehrle:
I'm looking at it now.  Please wait, thanks.

jfehrle:
For your second example (the bug), I made the exception go away.  The heuristic doesn't match goals (existential variables) that don't appear in Show Proof.  And I have no ideas on how to extend it.  Unmatched goals will be treated if they are new, i.e. highlighted.  BTW, I did diffs on all the .v files without encountering this new bug.  The original bug appeared only 1 or 2 proofs across all the .v files.  @eponier, do your proofs normally generate such existential variable, or is it more that you are trying to comprehensively test the diff code?

The `n, m : nat` should be fixable but it seems rather minor and the logic is a bit complex--while there are many other things to improve in Coq.  The 2 separate lines will appear in one line after the next tactic, e.g. "exfalso".

@herbelin, please take a look.  The changes are in a new commit.

eponier:
@jfehrle I would not say I am doing "comprehensive" testing, but indeed, my examples are crafted to make the diff mode crash. I have only tested the initial version of the diff mode on a few real .v files, and did not encounter any strange behaviour.

I don't know why you are worried by the fact that unmatched goals are treated as new, it seems to me a sensible default choice. In any case, my second example does not raise an anomaly any more, so I am happy with the current state of the PR.

herbelin:
> @herbelin, please take a look. The changes are in a new commit.

As an external observer, it is not clear to me why these `iter2` should be done only when the two lists have same length. What does it mean when the lists do not have the same length? This is a superficial remark, but why not e.g. to iter on a prefix or suffix of the longest list, or on the two lists up to some substitution?

>  In any case, my second example does not raise an anomaly any more, so I am happy (as a user) with the current state of the PR.

In any case, as assignee, both options are ok to me. I can merge as is (and leave better control of why lists have different lengths for latter), or wait.

jfehrle:
>  I don't know why you are worried by the fact that unmatched goals are
treated as new, it seems to me a sensible default choice.

Not worried, just trying to make the feature as accurate as reasonably
possible when it's released.

> As an external observer, it is not clear to me why these iter2 should be
done only when the two lists have same length. What does it mean when the
lists do not have the same length? This is a superficial remark, but why
not e.g. to iter on a prefix or suffix of the longest list, or on the two
lists up to some substitution?

I did look at using the prefix, which seemed reasonable for this particular
example.  I would need to see more examples, perhaps many more, to be
confident in choosing a prefix, suffix or up to some substitution (for
substitution, I've no idea what that might mean).  And it appears that this
is not a common use case.

Given the choice between not matching goals in some cases versus possibly
matching them incorrectly, I think not matching would have less impact on
users.  On the other hand, if goals are mismatched, that might be more
likely to generate a bug report, giving me more examples to examine.

So I would leave it this way for now.

herbelin:
> Given the choice between not matching goals in some cases versus possibly matching them incorrectly, I think not matching would have less impact on users (...)

OK, that clarifies and I think that I understand your approach. Fine to me.

So rescheduling a merge for tomorrow, unless someone (@eponier?) adds something by then.

eponier:
@jfehrle Thanks for adding the feature.

PR #9102, opened 2018-11-28
[ltac] Remove aliases already present in the lower layers.

ejgallego:
We remove a few aliases present in the lower layers
[`Genintern/Tactypes`] from `Tacexpr`.

IMHO this enlarges the API for no good purpose, and difficults
analysis.


gares (approved):
SSR ok

ppedrot (approved):

Issue #8180, opened 2018-07-28
Adopting a standard format for the CHANGES file?


herbelin:
#### Version

Targetting 8.9.

#### Description of the problem

When adding a line to the `CHANGES` file, I'm always confused about which layout and style to adopt, about which head section to use, and various other questions. Could we try to adopt a common guideline?

Here are a couple of questions to start with:

- Length of lines?
- Use of `` ` ``, or of `"`, or of `[...]`, etc. to refer to Coq syntax?
- Skipping a line between items of a list or not?
- Use of capitals in acronyms, for instance ocaml, OCaml, ..., ascii, ASCII, ...? Do we adopt official spellings as much as possible? Do we prefer uniformity of style within the `CHANGES` file itself?
- Choice of headers? Should features be highlighted by having their own headers? Should features enter the category they belong too. We can take the "Display diffs between proof steps" feature as an example.
- do we assign someone the responsibility to check at the syntactic consistency of `CHANGES`? (I think I missed the discussion about whether `CHANGES` has an official maintainer or not.)

Zimmi48:
I personally started using markdown-styled quoting in CHANGES in the hope that we would move at some point to a markdown file and that it would make transitioning easier.

I would approve having (short) guidelines on the style to use. These guidelines should be added to the contributing guide. It is not uncommon that contributing guides include such guidelines (and also how to write commit messages for instance).

> Do we adopt official spellings as much as possible?

:+1: 

> Choice of headers?
 
IMHO we should have a standard set of headers that we reuse every time in the same order. This info could be documented in the release documentation ([`dev/doc/release-process.md`](https://github.com/coq/coq/blob/master/dev/doc/release-process.md)) and it could be the role of the RM to create such (empty) sections just after branching.

> I think I missed the discussion about whether CHANGES has an official maintainer or not.

There is no code owner to avoid spamming. I guess we can say it is the responsibility of the RM to check the consistency of CHANGES before a release.

JasonGross:
I would support renaming `CHANGES` to `CHANGES.md` to get GitHub to render it more nicely.

herbelin:
#8151 shows a case of modification to `CHANGES` where both `` ` `` ... `` ` `` and `[` ... `]` are used. As a reviewer, do I recommend the author to use only `` ` `` ... `` ` ``?

Is there someone who cares about the subject of the discussion and who did not express an opinion?

Zimmi48:
[...]

PR #9103, opened 2018-11-28
[contributing] [changes] Improve accuracy of credits.

ejgallego:
It has been a while that I am not very happy with our current system
to record changes and credits. I see several problems:

- imprecise credit for many features, for example, reviewers are not
  acknowledged [and are sometimes crucial for the design, etc...],
- dissociation of credits and change entries,
- lack of categorization / systematization.

This is indeed a hard problem to solve (cc: #8180) ; in this PR I have
a very modest proposal improve accuracy and traceability by adapting
the model used in OCaml.

In this model, every change entry includes the name of the author, the
PR numbers, and a list of additional acknowledgements, where revievers
are usually mandatory.

I have filled in the information for a few entries I am responsible
of; please let me know what you think.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ OCaml and dependencies
 
 - Coq 8.10 requires OCaml >= 4.05.0, bumped from 4.02.3 See the
   INSTALL file for more information on dependencies.
+  (@ejgallego #7522, review by @gares, @Skyskimmer, and @Zimmi48)
 
 - Coq 8.10 doesn't need Camlp5 to build anymore. It now includes a
   fork of the core parsing library that Coq uses, which is a small
@@ -14,11 +15,13 @@ OCaml and dependencies
 
   The Coq developers would like to thank Daniel de Rauglaudre for many
   years of continued support.
+  (@ejgallego @ppedrot @Skyskimmer #8667 #8945 #8985, review by @gares)
 
 Coqide
 
 - CoqIDE now properly sets the module name for a given file based on
   its path, see -topfile change entry for more details.
+  (@Skyskimmer #8991, review by @ejgallego, @gares, and @ppedrot)
 
 Coqtop
 
@@ -26,6 +29,7 @@ Coqtop
   (à la -top) based on the filename passed, taking into account the
   proper -R/-Q options. For example, given -R Foo foolib using
   -topfile foolib/bar.v will set the module name to Foo.Bar.
+  (@Skyskimmer #8991, review by @ejgallego, @gares, and @ppedrot)
 
 Specification language, type inference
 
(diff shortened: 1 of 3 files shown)


ejgallego:
> * some of the developers/contributors have rather cryptic github nicknames, maybe we could use real names when they are known from us

I do agree, however with real names the credits entry tends to get very large and outshine the change entry itself, so I am not sure what the right tradeoff here is.

> * it seems to me that only user-visible changes will be credited, whereas we may want to also credit large clean-ups and redesign. It is not so much because I spend most of my time on this kind of activity ;) but rather that I'd like to see it incentivized for other developers

Indeed that's a problem I'm well aware of :) We should also do this to `dev/doc/changes` or course.

Maybe we should have `NEWS.md` where the main relevant user news are, and `CHANGES.md` should be a complete list of all changes / PR sorted in some order.

I dunno, after all this is not an exact science.

Zimmi48:
For a new project I really like this format for the changelog (it is the one used by Dune IIRC). However, in the case of Coq, I am unsure that this change on its own constitutes an improvement because it seems to me that we already have a lot of duplication of information (between `CREDITS`, `CHANGES`, the Credits chapter and the GitHub release notes) and this change seems to increases the duplication instead of reducing it. I would like this change to be part of a more general plan.

Furthermore, I agree with Maxime's second concern. `CHANGES` was a place for users to look at to learn about the most important information while upgrading. If we start crediting in this file, then there is no reason for not including everything that deserves credit, including large clean-ups, that are not necessarily of direct interest to the reader (no reason to bury this into `dev/doc/changes.md` either). In a sense, the separation between the Credits chapter (for credit and historical purposes) and the `CHANGES` file (for minimum information at the time of upgrade) seems to work well.

That being said, I would also be OK with a full consolidation of all this information in a single place (this is probably the only way to avoid duplication entirely): as long as we give it enough structure, like OCaml tries to do, it should be reasonable.

I generally agree with your initial considerations, including the one about the lack of credits for PR reviewers.

ejgallego:
Should we then use `CHANGES` for all changes + credits, and `NEWS` for user-related information?

Zimmi48:
Whichever solution is chosen must also include a strategy for the future of the Credits chapter.

ejgallego:
> @ejgallego Some unrelated commits are included (at least according to the github UI)

Indeed thanks, I guess I did rebase against the wrong master.

gares (changes requested):
Please resolve GitHub logins to real names.
We may move to another platform and these files would lose meaning. Also users do not necessarily know  these nicks.

ejgallego:
@Zimmi48 let me know how should I proceed with this.

Zimmi48:
@ejgallego I think I have mostly taken your recommendations into account in my work to consolidate changes and credits (#9668, #9939, #9964), so this PR could be closed. Let me know if I've missed something.

ejgallego:
Thanks @Zimmi48 for all the great and time-consuming work. IMHO this is good to close; I hope I can have a deeper look some of these days.

PR #9105, opened 2018-11-28
Add indexes for coercion / substructure symbol :>.

Zimmi48:
**Kind:** documentation

Note that there are more uses of `:>`, in particular the program chapter says:

>Also, `term :>` explicitly asks the system to coerce term to its support type. It can be useful in notations, for example: 
>```coq
>Notation " x `= y " := (@eq _ (x :>) (y :>)) (only parsing).
>```
>
>This notation denotes equality on subset types using equality on their support types, avoiding uses of proof-irrelevance that would come up when reasoning with equality on the subset types themselves.

But as this is mentioned in passing (https://coq.inria.fr/refman/addendum/program.html#syntactic-control-over-equalities), I wouldn't call this proper documentation.

cc @mattam82

### DIFF
--- a/doc/sphinx/addendum/implicit-coercions.rst
+++ b/doc/sphinx/addendum/implicit-coercions.rst
@@ -285,19 +285,21 @@ Activating the Printing of Coercions
 Classes as Records
 ------------------
 
+.. index:: :> (coercion)
+
 We allow the definition of *Structures with Inheritance* (or classes as records)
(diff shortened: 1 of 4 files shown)


cpitclaudel (approved):
Hey, that's very nice :)

cpitclaudel:
I'll merge as soon as you rebase.

Zimmi48:
Done.

PR #9106, opened 2018-11-28
[dune] Install coq libraries in `%{lib_root}/coq` instead of in `lib

ejgallego:
This is what the native Dune Coq version already does, but it is a
problem for Coq Dune too as noted by @vbgl .

### DIFF
--- a/tools/coq_dune.ml
+++ b/tools/coq_dune.ml
@@ -193,8 +193,8 @@ let pp_dep dir fmt oo = match oo with
 let out_install fmt dir ff =
   let itarget = String.concat "/" dir in
   let ff = pmap (function | VO vo -> Some vo.target | _ -> None) ff in
-  let pp_ispec fmt tg = fprintf fmt "(%s as %s)" tg (bpath [itarget;tg]) in
-  fprintf fmt "(install@\n @[(section lib)@\n(package coq)@\n(files @[%a@])@])@\n"
+  let pp_ispec fmt tg = fprintf fmt "(%s as coq/%s)" tg (bpath [itarget;tg]) in
+  fprintf fmt "(install@\n @[(section lib_root)@\n(package coq)@\n(files @[%a@])@])@\n"
     (pp_list pp_ispec sep) ff
 
 (* For each directory, we must record two things, the build rules and


ejgallego:
@vbgl don't hesitate to communicate your use case as the only way we can get things right is if we hear from them.

vbgl:
References: [opam manual](https://opam.ocaml.org/doc/Manual.html#installfield-lib_root) and [dune manual](https://dune.readthedocs.io/en/latest/usage.html?highlight=dune%20install#destination).

Quizz: I give to `dune install` both arguments `--prefix=/a` and `--libdir=/b`, where do the files in the `lib_root` section go?

The [configure](https://github.com/coq/coq/blob/8d765b25d1eb9a3ccf809eb423593d3612920820/configure.ml#L890) script hardcodes that the standard library goes into `<prefix>/lib/coq`. Thus I’d guess that `.vo` files belong neither to the `lib` section nor to the `lib_root` section.

ejgallego:
> References: [opam manual](https://opam.ocaml.org/doc/Manual.html#installfield-lib_root) and [dune manual](https://dune.readthedocs.io/en/latest/usage.html?highlight=dune%20install#destination).
> 
> Quizz: I give to `dune install` both arguments `--prefix=/a` and `--libdir=/b`, where do the files in the `lib_root` section go?

I am sorry @vbgl I am not sure I understand what the problem is here, can you be more explicit?

> The [configure](https://github.com/coq/coq/blob/8d765b25d1eb9a3ccf809eb423593d3612920820/configure.ml#L890) script hardcodes that the standard library goes into `<prefix>/lib/coq`. Thus I’d guess that `.vo` files belong neither to the `lib` section nor to the `lib_root` section.

The do belong to `$lib_root/coq`, as this PR does; in the future hopefully they will belong to `$lib` or to `$lib_root/coq/$lib`, etc...

vbgl:
One problem is: how to read the doc?

The path for the standard library (aka theories) is `<prefix>/lib/coq`,
thus, in the above example, it should be `/a/lib/coq`. However, with and
without this PR, they end up in `/b/coq` which is wrong (`coqtop` cannot
find the library).

ejgallego:
> One problem is: how to read the doc? The path for the standard library (aka theories) is `<prefix>/lib/coq`, thus, in the above example, it should be `/a/lib/coq`. However, with and without this PR, they end up in `/b/coq` which is wrong (`coqtop` cannot find the library).

Well that seems correct, right? You are remapping `lib` to `b`, aren't you?

What is the concrete layout in [Nix I assume]? What is `b` here?

ejgallego:
Maybe Coq's configure can allow to specify the library dir better if that's needed.

vbgl:
There are two library directories here. One for the Coq standard
library. One for Coq as an OCaml library. What I try to achieve is to
separate them.

ejgallego:
> There are two library directories here. One for the Coq standard library. One for Coq as an OCaml library. What I try to achieve is to separate them.

Umm, doesn't the install file section `lib_root:` allows you to do that? With this setup:
- `lib:` ocaml files
- `lib_root` rest of architecture dependent files

ejgallego:
We could add a "libcoq" section to Dune by the way, but IMHO that seems a bit overkill, `lib_root` should be able to do the job.

vbgl:
Great. So back to the documentation-reading issue: how do I tell `dune
install` what is the `lib_root` directory?

ejgallego:
> I agree. Then let me reformulate: how do I tell coqtop/configure what is the lib directory?

Is not `-libdir` enough?

It feels like I am missing something here. Maybe you are having problems due to the ML part of `plugins` being in `lib` and the `.vo` part in `lib_root`?

Indeed that could be a serious problem depending on your layout as Coq doesn't follow OCaml conventions when it comes to plugins.

Tho, would we solve #7698 this could be fixed too [and maybe we could even deprecate -I ?]

ejgallego:
> Do you mean `configure`’s argument? I think I don’t have access to it when building with `dune`.

It is not documented yet as it is not tested but indeed you can call `./configure` and Dune will respect the settings, see the `(mode fallback)` setting in `config/dune`.

ejgallego:
Irregardless of the valid questions opened by @vbgl , this should go in as indeed it is a fix cc: @Zimmi48 @vbgl

vbgl (approved):

PR #9116, opened 2018-11-30
[assumptions] Some refactoring on the printing code.

ejgallego:
This arose in the context of ejgallego/coq-serapi#86

Note that I am not sure this is the best way to it, as it feels like
we have quite a bit of duplication between the new `AssumptionCtx`
structure and `ContextObjectMap` [which by the way, shouldn't be a
set]?

Thus, I am submitting this PR as to gather more feedback and design
comments, instead of letting it rot in my SSD.

### DIFF
--- a/printing/printer.ml
+++ b/printing/printer.ml
@@ -892,9 +892,50 @@ end
 module ContextObjectSet = Set.Make (OrderedContextObject)
 module ContextObjectMap = Map.Make (OrderedContextObject)
 
-let pr_assumptionset env sigma s =
-  if ContextObjectMap.is_empty s &&
-       engagement env = PredicativeSet then
+module AssumptionCtx = struct
+
+  type t =
+    { predicative : Declarations.set_predicativity
+    ; type_in_type : bool
+    ; vars   : (Id.t * types) list
+    ; axioms : (axiom * types * (Label.t * Constr.rel_context * Constr.t) list) list
+    ; opaque : (Constant.t * types) list
+    ; trans  : (Constant.t * types) list
+    }
+
+  let build env ctxmap =
+    let fold t typ accu =
+      let (v, a, o, tr) = accu in
+      match t with
+      | Variable id ->
+        ((id,typ) :: v, a, o, tr)
+      | Axiom (axiom, l) ->
+        (v, (axiom,typ,l) :: a, o, tr)
+      | Opaque kn ->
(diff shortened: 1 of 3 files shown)


ejgallego:
Ping @herbelin @mattam82

ejgallego:
> Maybe we could build the new object directly.

Agree, in fact I am not sure the new object is really an improvement over the `ContextObjectMap` itself.

> Also, I don't know why ContextObjectMap lives in Printer, wouldn't it fit better in Assumptions?

Agree also, maybe we should slice this part of the printer?

SkySkimmer:
@ejgallego so wrt the last couple comments do you still want this PR merged / do you want to work on it more / just close it?

ejgallego:
I'd like first to get an assignee so we can discuss what to do. Definitively improving the assumptions interface is needed.

ejgallego:
No assignee and this is low priority for me, no point on adding to 8.10 load IMO.

ppedrot (approved):
I am fine with the current status of this PR. A meager cleanup is better than no cleanup at all. If needed, I can assign (and merge).

ejgallego:
Thanks @ppedrot , please assign, but now that I am at it will try to implement some of the suggestions made in the review. I'll ping again when ready.

herbelin (commented):
No strong opinion. From a superficial reading, that looks good but maybe not crucial either.

ejgallego:
> No strong opinion. From a superficial reading, that looks good but maybe not crucial either.

The original motivation for this change @herbelin was that some users [for example Armael] wanted to have a slightly different API for programatically checking assumptions.

herbelin:
@ejgallego: OK

ejgallego:
> @ejgallego: OK

I will produce and update version tho, so I will ping you again when the updated PR is in.

ppedrot:
This PR is lingering for no good reason. It is a small code change, so either it gets merged quickly (but now it needs rebasing :zap:) or @ejgallego should close it and reopen a fresh one with the desired enhancements. In both cases, there is no point in keeping this open, I believe.

ejgallego:
Yeah I will update it when I get some free cycles; the PR is actually needed as there are users of this feature [now SerAPI just duplicate the code]

ppedrot:
In the meantime I am closing this. @ejgallego please reopen it when ready.

PR #9117, opened 2018-11-30
Accept argument names for extra arguments with "extra scopes"

maximedenes:
The checks were unnecessarily restrictive (since names can be used for
documentation purposes), and the error message was a bit wrong (it
mentioned a restriction on the explicit status of arguments).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -88,6 +88,8 @@ Vernacular commands
 - The `Automatic Introduction` option has been removed and is now the
   default.
 
+- `Arguments` now accepts names for arguments provided with `extra_scopes`.
+
 Tools
 
 - The `-native-compiler` flag of `coqc` and `coqtop` now takes an argument which can have three values:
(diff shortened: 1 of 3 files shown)


maximedenes:
This is a wish discussed here: https://github.com/math-comp/math-comp/pull/253#discussion_r237857556

maximedenes:
@mattam82 ping

herbelin:
Looks good. Can you add also an example in the test suite?

maximedenes:
@herbelin added a test case and a CHANGES entry

@mattam82 can you be the assignee?

maximedenes:
> hat looks like a cool feature to have, a bit dual in spirit to `'pat`, using instead coercions.

Why not.

@mattam ping, this is ready. Thanks!

maximedenes:
> That looks like a cool feature to have, a bit dual in spirit to `'pat`, using instead coercions.

Why not.

@mattam82 ping, this is ready. Thanks!

mattam82 (approved):
Alright with me, I’ll merge ASAP

maximedenes:
Thanks!

PR #9121, opened 2018-11-30
Closes #9118: single backticks are made equivalent to double backticks; try to fix all misuses.

### LINKED ISSUE #9118: Sphinx warning for use of single backticks.
@cpitclaudel If I'm not mistaken, all uses of single backticks without a preceding `:foo:` in the manual are wrong (and people continue to introduce some because they are used to Markdown syntax). Can we get a warning for this? (We should really strive for a CI that catches as much as possible by itself. Nowadays, most of my documentation reviewing is to tell people they are not using the format correctly.)

Zimmi48:


doc/sphinx/language/gallina-extensions.rst
```diff
@@ -1803,10 +1803,10 @@ Explicit applications
 In presence of non-strict or contextual argument, or in presence of
 partial applications, the synthesis of implicit arguments may fail, so
 one may have to give explicitly certain implicit arguments of an
-application. The syntax for this is ``(`` `ident` ``:=`` `term` ``)`` where `ident` is the
+application. The syntax for this is :n:`(@ident := @term)` where :token:`ident` is the
```
jfehrle: The `@ident` is OK here?  Doesn't have to be like this one?

```
:token:`ident`
```

jfehrle:
```
If ":n:`@ident`" is equivalent to ":token:`ident`", would it be easier just to use ":n:" everywhere
(get rid of one syntax variation)?  (Although we have a vast number of ":token:"s, may not be 
orth replacing them all.
```

jfehrle (commented):
Didn't see anything obviously wrong.

doc/sphinx/language/gallina-extensions.rst
Zimmi48: `:token:` is a primitive object of the rst format while `:n:` is something specific to Coq's Sphinx. But indeed, you are raising a good question.

doc/sphinx/language/gallina-extensions.rst
```diff
@@ -699,21 +698,22 @@ used by ``Function``. A more precise description is given below.
...
-   Defines the not recursive function `ident` as if declared with `Definition`. Moreover
-   the following are defined:
+   Defines the not recursive function :token:`ident` as if declared with
+   :cmd:`Definition`. Moreover the following are defined:
 
-    + `ident_rect`, `ident_rec` and `ident_ind`, which reflect the pattern
-      matching structure of `term` (see :cmd:`Inductive`);
-    + The inductive `R_ident` corresponding to the graph of `ident` (silently);
-    + `ident_complete` and `ident_correct` which are inversion information
+    + :n:`@ident_rect`, :n:`@ident_rec` and :n:`@ident_ind`, which reflect the pattern
```
Zimmi48: @cpitclaudel This change is actually incorrect as I've just discovered because Sphinx thinks `ident_rect` is the token here. Can we get a way of indicating the token stops after `ident`? (For instance `:n:`@ident\_rect`.)

cpitclaudel: Does ``` :token:`ident`\ ``_rect`` ``` work?

doc/sphinx/addendum/extraction.rst
```diff
@@ -47,30 +47,30 @@ extraction. They both display extracted term(s) inside |Coq|.
 All the following commands produce real ML files. User can choose to
 produce one monolithic file or one file per |Coq| library.
 
-.. cmd:: Extraction "@file" {+ @qualid }
+.. cmd:: Extraction @string {+ @qualid }
 
    Recursive extraction of all the mentioned objects and all
-   their dependencies in one monolithic `file`.
+   their dependencies in one monolithic file :token:`string`.
```
cpitclaudel: Hmm, `@file` was nice for documentation purposes though

doc/sphinx/addendum/extended-pattern-matching.rst
```diff
@@ -31,9 +31,9 @@ A variable pattern matches any value, and the identifier is bound to
 that value. The pattern “``_``” (called “don't care” or “wildcard” symbol)
 also matches any value, but does not bind anything. It may occur an
 arbitrary number of times in a pattern. Alias patterns written
-:n:`(@pattern as @identifier)` are also accepted. This pattern matches the
-same values as ``pattern`` does and ``identifier`` is bound to the matched
-value. A pattern of the form :n:`pattern | pattern` is called disjunctive. A
+:n:`(@pattern as @ident)` are also accepted. This pattern matches the
+same values as :token:`pattern` does and :token:`identifier` is bound to the matched
```
cpitclaudel: should this be `ident` instead of `identifier`?

doc/sphinx/addendum/extraction.rst
Zimmi48: I agree but this one the only use of this token in the all manual. Not enough to define a new grammar token.

doc/sphinx/language/gallina-extensions.rst
Zimmi48: It does but I was hoping we would try to avoid this...

doc/sphinx/addendum/extended-pattern-matching.rst
Zimmi48: Good catch.

doc/sphinx/language/gallina-extensions.rst
Zimmi48: @cpitclaudel What is your take on this?

cpitclaudel: No strong feelings either way :)

doc/sphinx/language/gallina-extensions.rst
cpitclaudel: I agree, it's not ideal. Do we have token names with underscores in them? Otherwise underscores could be a separator.

cpitclaudel: I think making the rendering uniform would be good.

Zimmi48: Unfortunately yes, we have many.

Zimmi48: >I think making the rendering uniform would be good.

If possible, I would make ``` :token:`foo` ``` look like ``` :n:`@foo` ```. But is that possible easily? It seems to me that this is something we might not have control over as it likely comes from the ReadTheDocs theme.

cpitclaudel: It's easy to override parts of the theme, using our own CSS.  In `notations.css` you can change `.notation .hole {` into `.notation .hole, .std-token .pre {`; that should do it.

Zimmi48: For some strange reason I don't understand, your suggested change has the opposite effect (it uses the style of the grammar token - blue, non italic - everywhere). I don't really mind: this is pretty as well. But it feels weird that such change to the CSS would have such effect.

cpitclaudel: Could it be that you broke the CSS, and thus that rule doesn't apply anymore? That change works for me when I live-test it in the firefox debugger :)

Zimmi48: Good guess. I had missed the comma.

Zimmi48: Fixed.

Zimmi48:
If you have no more comments, this PR is ready from my point of view. In particular, it contains a commit uniformizing the rendering of `:token` and `:n:` to the green, italic, bold version (which was already the most common).

cpitclaudel (approved):

cpitclaudel:
Thanks!

PR #9122, opened 2018-12-01
[genarg] Remove some deprecated genargs.

ejgallego:
No point on having these aliases; note that there are issues with the
current setup for `wit_preident/wit_pre_ident` and `wit_ref`.

[`ref` is taken as the keyword for example, the entry for preident
doesn't match the alias]


gares:
I'm ok removing aliases but the old names are better than the new ones imho. Eg quant_hyp.

ejgallego:
> I'm ok removing aliases but the old names are better than the new ones imho. Eg quant_hyp.

I agree, I think I preferred some old ones too.

ppedrot (approved):
No specific comment for naming, it is not an exact science.

ejgallego:
Ok thanks! Then I propose we follow @gares suggestions. Would that be ok with you?

ejgallego:
@ppedrot ?

ppedrot:
Yes, go ahead.

maximedenes (approved):
LGTM module the remarks on @gares about the naming.

ejgallego:
Should be ready modulo overlays, I think I got the naming right.

herbelin (commented):
In theory, the change `preident` -> `pre_ident` in the witness name may be a source of incompatibility in third-party `TACTIC EXTEND` (and virtually `Tactic Notation`, though I was not able to build an interesting tactic notation using `preident`).

A unification `wit_pre_ident` -> `wit_preident` might be safer.

About `reference`/`global` the former is what is documented (in the syntax chapter), so, even if maybe less intuitive, this should probably indeed be preferred.

ejgallego:
> A unification `wit_pre_ident` -> `wit_preident` might be safer.

Ok, sounds good @herbelin .

> About `reference`/`global` the former is what is documented (in the syntax chapter), so, even if maybe less intuitive, this should probably indeed be preferred.

We could change the documentation; IMHO this PR is a good opportunity to converge to a single set of identifiers for the arguments, so if you think that's better let's go for it.

Personally I don't have much of a preference, it is just the duplication that does bother me.

ppedrot:
@ejgallego any news on that PR?

ejgallego:
> @ejgallego any news on that PR?

It is basically ready [modulo a renaming / rebase] I'll try to take care when of it when I have some free time.

ppedrot:
@ejgallego ping

ejgallego:
> @ejgallego ping

Indeed I already did a rebase [but didn't push as some things broke] sorry for the delay, I hope I can push this soon.

ejgallego:
Hopefully I got this right finally; let's see what the CI says.

ejgallego:
Once there is an assignee and approval I'll do the overlays.

interp/stdarg.ml
```diff
@@ -51,13 +51,4 @@ let wit_uconstr = make0 "uconstr"
...
-  make0 "clause_dft_concl"
-
-(** Aliases for compatibility *)
-
-let wit_integer = wit_int
-let wit_preident = wit_pre_ident
-let wit_reference = wit_ref
-let wit_global = wit_ref
-let wit_clause = wit_clause_dft_concl
+let wit_clause_dft_concl = make0 "clause_dft_concl"
```
ppedrot: I am fine with the other changes, but I think that particular one is the wrong way. `clause_dft_concl` looks like a much uglier name than just `clause`.

ejgallego: Amended, thanks!

doc/changelog/05-tactic-language/9122-genarg_cleanup.rst
```diff
@@ -0,0 +1,10 @@
+- **Changed:**
+  Some duplicated grammar binders in tactic notations have been
+  removed, for example in ``Tactic Notation "foo" ref(x) := idtac.``
+  you now must use ``reference(x)``, etc.... The full list is:
+  + `integer` -> `int`
```
JasonGross: Does `int` even work with `Tactic Notation`?  I was under the impression that only `integer` worked.

JasonGross: Also, it seems wrong to change `integer` to `int` but `ref` to `reference`.  All the Ltac code in the wild I'm aware of uses `integer` and `reference`, I think we should standardize on that

ejgallego: Actually there are some extra compatibility notations registered in `extraargs.mlg` which indeed is making this work; I'm unsure what the relation is here, @ppedrot ? I've only updated the "core" ones.

ejgallego: > Does `int` even work with `Tactic Notation`? I was under the impression that only `integer` worked.

At least today both work, one from the core arg mechanism and the other from the compat layer I mentioned. I'm ok switching everything to `integer` if that's the preferred solution, but I'd like to get rid of duplicates. I got some users in the ML world that indeed would like the syntax to be more canonical.

JasonGross: For the ones accessible via `Tactic Notation`, let's standardize on the ones [currently document](https://coq.inria.fr/refman/user-extensions/syntax-extensions.html#tactic-notations), that is:
`ident`, `simple_intropattern`, `reference`, `hyp`, `hyp_list`, `ne_hyp_list`, `constr`, `uconstr`, `constr_list`, `ne_constr_list`, `integer`, `integer_list`, `ne_integer_list`, `int_or_var`, `int_or_var_list`, `ne_int_or_var_list`, `tactic`, `tactic0`, `tactic1`, `tactic2`, `tactic3`, `tactic4`, `tactic5`

ejgallego:
I'm afraid this is late for 8.12

maximedenes (commented):
This PR looks broken or missing some overlays, right?

ejgallego:
The PR was a bit of a headache, needs some progress indeed with the overlays and maybe with Jason's comments.

PR #9132, opened 2018-12-04
Refresh universes before trying to coerce.

### LINKED ISSUE #6682: Universe Polymorphism and Canonical Structures containing `Type`
#### Version
8.8+alpha (and 8.7.1)

The following snippet fails. I expected it to succeed (after https://github.com/coq/coq/issues/5368 was fixed, at least).

```Coq
Set Universe Polymorphism. Unset Universe Minimization ToSet.
Structure NoProof := no_proof { np_type : Type }.
Canonical Structure NP_Type := no_proof Type. 
Check (fun np (T : np_type np) => I) _ Type.
```

SkySkimmer:
Fix #6682.
Bench waiting jenkins 595.

### DIFF
--- a/pretyping/coercion.ml
+++ b/pretyping/coercion.ml
@@ -521,6 +521,13 @@ let rec inh_conv_coerce_to_fail ?loc env evd rigidonly v t c1 =
 
 (* Look for cj' obtained from cj by inserting coercions, s.t. cj'.typ = t *)
 let inh_conv_coerce_to_gen ?loc ~program_mode resolve_tc rigidonly env evd cj t =
+  let evd, cj =
+    let evd, uj_type = Evarsolve.refresh_universes
+        ~status:Evd.univ_flexible ~onlyalg:true (Some false)
+        env evd cj.uj_type
+    in
+    evd, {cj with uj_type }
+  in
   let (evd', val') =
     try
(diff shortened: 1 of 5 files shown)


pretyping/coercion.ml
```diff
@@ -521,6 +521,13 @@ let rec inh_conv_coerce_to_fail ?loc env evd rigidonly v t c1 =
 
 (* Look for cj' obtained from cj by inserting coercions, s.t. cj'.typ = t *)
 let inh_conv_coerce_to_gen ?loc resolve_tc rigidonly env evd cj t =
+  let evd, cj =
+    let evd, uj_type = Evarsolve.refresh_universes
+        ~status:Evd.univ_flexible ~onlyalg:true (Some false)
+        env evd cj.uj_type
+    in
+    evd, {cj with uj_type }
+  in
```
SkySkimmer: I don't know if this is the right place to put this, we could also go for `Pretyping.inh_conv_coerce_to_tycon` (with possibility of missing other callers).
Here seems OK though.

SkySkimmer:
Breaks developments, will investigate.

SkySkimmer:
I can't figure out where the incompatibility comes from exactly, don't know how to progress.
@mattam82 WDYT?

pretyping/coercion.ml
mattam82: It sounds quite strong to always refresh, while leq can handle cj.uj_type being an algebraic.

PR #9138, opened 2018-12-05
[funind] Apply automatic code formatting to the code base.

ejgallego:
This is an experiment to format Coq's code base automatically.

I have chosen `funind` as indeed, even if our policy of no format-only
changes is IMO a good thing, the formatting in this part of the
codebase is so bad that it was making my editing there painful, thus
indeed such measure could be justified here.

The command chosen is:
```
ocamlformat --module-item-spacing=compact --inplace --break-sequences --sequence-style=terminator --cases-exp-indent=2 --field-space=loose *.ml *.mli
```

Indeed the result is not perfect but so bad either. I like it. Let me
know what think folks.

### DIFF
--- a/plugins/funind/functional_principles_proofs.mli
+++ b/plugins/funind/functional_principles_proofs.mli
@@ -1,19 +1,26 @@
 open Names
 
 val prove_princ_for_struct :
-  Evd.evar_map ref ->
-  bool ->
-  int -> Constant.t array -> EConstr.constr array -> int -> Tacmach.tactic
-
+     Evd.evar_map ref
+  -> bool
+  -> int
+  -> Constant.t array
+  -> EConstr.constr array
+  -> int
+  -> Tacmach.tactic
 
 val prove_principle_for_gen :
-  Constant.t * Constant.t * Constant.t -> (* name of the function, the functional and the fixpoint equation *)
-  Indfun_common.tcc_lemma_value ref -> (* a pointer to the obligation proofs lemma *)
-  bool -> (* is that function uses measure *)
-  int -> (* the number of recursive argument *)
-  EConstr.types -> (* the type of the recursive argument *)
-  EConstr.constr -> (* the wf relation used to prove the function *)
-  Tacmach.tactic
-
-
+     Constant.t * Constant.t * Constant.t
+  -> (* name of the function, the functional and the fixpoint equation *)
+     Indfun_common.tcc_lemma_value ref
+  -> (* a pointer to the obligation proofs lemma *)
+     bool
+  -> (* is that function uses measure *)
+     int
+  -> (* the number of recursive argument *)
+     EConstr.types
+  -> (* the type of the recursive argument *)
+     EConstr.constr
+  -> (* the wf relation used to prove the function *)
+     Tacmach.tactic
 (* val is_pte  : rel_declaration -> bool  *)
--- a/plugins/funind/functional_principles_proofs.ml
+++ b/plugins/funind/functional_principles_proofs.ml
@@ -15,7 +15,6 @@ open Indfun_common
 open Libnames
 open Globnames
 open Context.Rel.Declaration
-
 module RelDecl = Context.Rel.Declaration
 
 (* let msgnl = Pp.msgnl *)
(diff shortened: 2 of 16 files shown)


ejgallego:
I cc people that could be interested @gares @maximedenes @SkySkimmer @herbelin @ppedrot @mattam82

plugins/funind/functional_principles_proofs.mli
```diff
@@ -1,19 +1,26 @@
...
-  Indfun_common.tcc_lemma_value ref -> (* a pointer to the obligation proofs lemma *)
-  bool -> (* is that function uses measure *)
-  int -> (* the number of recursive argument *)
-  EConstr.types -> (* the type of the recursive argument *)
-  EConstr.constr -> (* the wf relation used to prove the function *)
-  Tacmach.tactic
-
-
+     Constant.t * Constant.t * Constant.t
+  -> (* name of the function, the functional and the fixpoint equation *)
```
SkySkimmer: I'm not sure about the mli changes, some manual attention is needed because it makes the comments look misplaced at least.
Having the arrow at the start of the line is not bad but it doesn't match the rest of the code, is there an option to change it?

ejgallego: Not in the released version, you can use ` [@@@ocamlformat "disable=false"]` in that part.

I have switched to that style on my projects tho, it makes named arguments and `odoc` pretty convenient.

ejgallego: Note that ocamldoc style is as done by `ocamlformat`, comment below the argument.

SkySkimmer: I'd prefer the comment before the arrow though. I think it would produce something like
~~~ocaml
val bla : 
  foo
  (** it's a foo *)
  -> bar
  (** bla makes a bar *)
~~~

Zimmi48:
I'm generally in favor of using `ocamlformat`. I also think it makes sense to switch piece by piece (I would even go slower and do it file by file). However, we need first to make sure that we have the right tool support that goes with it. I.e. how to tell Emacs to apply `ocamlformat` on save but only for some files. Same question for Dune (which can be used to check the formatting).

plugins/funind/functional_principles_proofs.mli
ejgallego: Isn't this case a problem of the original code indeed? I guess ocmalformat cannot move comments thru arrow-boundaries.

ejgallego:
> However, we need first to make sure that we have the right tool support that goes with it. I.e. how to tell Emacs to apply `ocamlformat` on save but only for some files.

This could be a bit expensive in general; indeed I guess we could start with only files in `plugins/funind`.

The thing for `ocamlformat` is that you can disable it, but it is verbose. A problem of our `mlg` files not using ppx is that it won't work for them.

> Same question for Dune (which can be used to check the formatting).

Dune has reasonable support for formatting, tho it would need maybe a bit more. See https://dune.readthedocs.io/en/latest/formatting.html

Zimmi48:
The core of my thought is that using `ocamlformat` should make it *easier* to contribute and not *harder*. This is a condition *sine qua non*.

ejgallego:
> The core of my thought is that using `ocamlformat` should make it _easier_ to contribute and not _harder_. This is a condition _sine qua non_.

Well, it is an additional tool posing constraints on how code look so I am not sure it will make life easier to people. Would we adopt it in the full codebase indeed we could run `dune build @fmt` before commit but this would also have a cost in terms of time, Coq's codebase is large [even if you only run on changed files].

On the other hand you get a nice uniformity of style, which should also help ; so there is a tradeoff here.

Zimmi48:
In programming language community that are using automatic formatting systematically, this practice has indeed made contributing easier because people do not even have to think about whether their code is formatted correctly. I think that it could work for OCaml projects too as long as this is well documented. In the case of Coq,

I indeed fear a scenario *à la* linter (especially in the beginnings) when it made merging a PR significantly harder. I also cannot approve a PR that would apply a formatting commit but wouldn't ensure that this formatting is respected from then on.

ejgallego:
I understand your point of view; however achieving this may take a while as for example we may require to have a more unified tooling.

I dunno what to, this PR happened tho because of real pain with the files [see recent comments on other issues :D ] so it may be useful even if we don't add the `ocamlformat` tooling yet; which will require quite a bit of discussion I'm sure [and maybe even a newer `ocamlformat` that better adapts to Coq style]

Zimmi48:
A pure formatting PR that changes ~6000 LOC seems a bit much if we don't enforce this style from then on.
I would prefer that you modify just the bits that you are changing, at the same time that you are changing them. Or, just at the level of files that you are changing, if it's too complex at a smaller levels...

maximedenes:
> the formatting in this part of the
> codebase is so bad that it was making my editing there painful

I personally find the formatting issue almost negligible, compared to the fact that most of this part of the code makes no sense... So no opinion on this particular issue.

ejgallego:
> I would prefer that you modify just the bits that you are changing, at the same time that you are changing them. Or, just at the level of files that you are changing, if it's too complex at a smaller levels...

Yeah I could do at the file level, it is hard at the local level due to very strange identation. But see for example #6171 where every file is touched, that would imply this; so IMHO this is clearer.

I can try on enforcing the style indeed, but it would mean that every developer has to install ocamlformat, dune, etc... so the issue is not so easy.

@maximedenes the code makes much more sense once you can read it.

maximedenes:
> the code makes much more sense once you can read it.

I wish! I did reformat it locally, but didn't feel the same effect, unfortunately.

Zimmi48:
>I can try on enforcing the style indeed, but it would mean that every developer has to install ocamlformat, dune, etc... so the issue is not so easy.

Dune will soon be required anyways. And the Nix file could include ocamlformat. Those using opam shouldn't have any trouble either. But this would require at least very clear documentation.

ejgallego:
> Dune will soon be required anyways. And the Nix file could include ocamlformat. Those using opam shouldn't have any trouble either. But this would require at least very clear documentation.

Even so it seems to me like a high barrier of entry; IMHO the hook should be at most optional, imagine the casual contributor being put off by this. `ocamlformat` is not trivial to install at all [uses ppx, etc...]

ejgallego:
Let's not derail the discussion folks. I agree that "makes no sense" is a too strong term, also it is time for Santa to come with presents so herbelin  let's not use name calling either.

In my opinion the situation is as follows:
-we all would like to maintain and improve funind.
- this will likely be an incremental task
- I felt that formatting was getting too much in my way when working on that file,
- instead of doing a manual fix-up, I chose to use a tool, without willing to make a more general project-wide statement on automated formatting,
- I do agree with Theo's analysis of the risks of accepting this PR,
- even so I feel this is a bit of an exceptional case and this PR could help, Maxime does not think so,

So, it is clear we are far from having a complete solution, so the question is, do the risks posed by Theo outweight the short-term gain on readability?

maximedenes:
> * even so I feel this is a bit of an exceptional case and this PR could help, Maxime does not think so,

I'd to clarify that even if I don't find it so significantly helpful, I also don't think it hurts, so I'm not opposing this PR.

herbelin:
@ejgallego: thanks for your positivism.

I'd like to be positive but the kind of negative words I read above, maybe should I actually say self-negative words, because it is gratuitously negative about Coq itself, each time are *hurting* me. Over the past, I feel I used all kinds of attitudes to express that this was *hurting* me, and that this was *unfair*. I asked many times for elaboration of the given statements, mostly only receiving silence. At some times, I went even very upset. At other times, I tried to be nice, but, again without apparent success. I feel powerless on what to do to prevent such attitudes. @maximedenes, _do you understand that such unjustified wrong negative claim is unfair and hurts me_? Is there anything you can do about it?

About the PR, I'd rather be in favor of it as a general rule (assuming a clear formatting rule) but I can both live with and without.

About Santa, yes, I did not know it was today, but we could indeed come with gifts. After all, there are so many positive things which happen around Coq. Let me remotely share a beer with everyone around.

SkySkimmer:
>we all would like to maintain and improve funind.

I thought the plan was to replace it with equations at some point?

ejgallego:
> I thought the plan was to replace it with equations at some point?

Sure, but that's long term; maybe I didn't phrase what I meant properly, but indeed why not to improve it? It is low priority but along other refactorings we can keep improving it a bit, as we do for other parts of the system.

As of today funind is going to stay for a while and IMHO there is room to improve it, tho some stuff is undoubtedly hard.

SkySkimmer:
We should make a decision to merge or close this.

Zimmi48:
Related to #9572.

ejgallego:
> We should make a decision to merge or close this.

Yup; I find funind code unworkable that's why I submitted it; but anyways I had to do a very large funind refactoring without this so the pain was inflicted; not sure I'm looking forward to the rebase pain.

PR #9140, opened 2018-12-05
[ci] Add four color theorem proof to CI

ejgallego:


SkySkimmer (approved):
I wonder if we should split math-comp.sh into multiple files (ala formaltopology/corn/mathclasses).

ejgallego:
> I wonder if we should split math-comp.sh into multiple files (ala formaltopology/corn/mathclasses).

I don't think we'll be able to gain a lot with this setup, I think the splitting above stems from historical reasons [where they used to be tested separadtley]

But indeed it seems that `fourcolor` is broken in master :(

ppedrot:
> But indeed it seems that fourcolor is broken in master :(

A kind soul around willing to fix it?

ejgallego:
https://github.com/math-comp/fourcolor/pull/2

ejgallego:
Fix applied upstream, rerunning the CI to see if it is fine.

SkySkimmer:
Looks fine, shall I merge?

ejgallego:
Yes please.

PR #9144, opened 2018-12-05
Fixes #4633: clearer message unknown existential

### LINKED ISSUE #4633: "Error: Unknown existential variable." is not a great error message

Original bug ID: BZ#4633
From: @JasonGross
Reported version: 8.5

herbelin:
Hi, this is my contribution of the day to addressing open issues. It answers #4633 by providing the following alternative error message (mostly inspired from @JasonGross' proposal):
```
Goal exists x : nat, x = x.
eexists. (* ?x = ?x *)
set (v := ?x).
(*
          ^^
Error:
?x is a generated name. Only user-given names for existential variables can
be referenced. To give a user name to an existential variable, introduce it
with the ?[name] syntax.
*)
```

**Kind:** bug fix

Fixes / closes #4633

### DIFF
--- a/pretyping/pretyping.ml
+++ b/pretyping/pretyping.ml
@@ -486,8 +486,7 @@ let rec pretype k0 resolve_tc (tycon : type_constraint) (env : GlobEnv.t) (sigma
       let id = interp_ltac_id env id in
       let evk =
         try Evd.evar_key id sigma
-        with Not_found ->
-          user_err ?loc  (str "Unknown existential variable.") in
+        with Not_found -> error_evar_not_found ?loc !!env sigma id in
       let hyps = evar_filtered_context (Evd.find sigma evk) in
       let sigma, args = pretype_instance k0 resolve_tc env sigma loc hyps evk inst in
       let c = mkEvar (evk, args) in
--- a/engine/termops.ml
+++ b/engine/termops.ml
@@ -38,7 +38,7 @@ let set_print_constr f = term_printer := f
 
 module EvMap = Evar.Map
 
-let pr_evar_suggested_name evk sigma =
+let evar_suggested_name evk sigma =
   let open Evd in
   let base_id evk' evi =
   match evar_ident evk' sigma with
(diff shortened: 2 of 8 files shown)


pretyping/pretyping.ml
```diff
@@ -487,7 +487,15 @@ let rec pretype k0 resolve_tc (tycon : type_constraint) (env : GlobEnv.t) (sigma
       let evk =
         try Evd.evar_key id sigma
         with Not_found ->
-          user_err ?loc  (str "Unknown existential variable.") in
+          let all_undef_evars = Evar.Set.elements (Evar.Map.domain (Evd.undefined_map sigma)) in
+          if List.mem id (List.map (fun ev -> evar_suggested_name ev sigma) all_undef_evars) then
```
ppedrot: This is going to be horribly slow for a mere debugging feature.

ppedrot (dismissed):
Please use an exception and register a printer that does the name check at printing time, otherwise this makes using evar names algorithmically crazy.

pretyping/pretyping.ml
herbelin: To be on the safe side, I delayed the printing of the message to `himsg.ml` and made the test of belonging to the list of evar lazy (so that adding a number suffix does not systematically induce a n^2 complexity in case all evars have the same name, say `Goal`). But an LtacX funciton is (in principle) not supposed to use a `?x` with an `x` not bound by the LtacX function. So, the situations where the error would be raised and would not go directly to the toplevel seem rare (e.g a complex tactic expression with backtracking in strict mode?).

(Incidentally, I think there is a misunderstanding. It is not about debugging but about improving a user-level message.)

But you are right, let us be on the safe side. As far as I can judge, the worse we can have is `n` evars with same name `id` and asking for `?idN` for `N`<`n`, which shall be in complexity `N`^2.

ppedrot (approved):

ppedrot:
This changes the API and it can hardly be considered as a bugfix, rather an enhancement. Thus I am not marking it for backporting to 8.9.1.

PR #9145, opened 2018-12-05
Do so that an error message follows the "Error:" header on the same line

herbelin:
This PR is proposing to follow a common usage and to display the error message on the same line after the `Error:`  header (when in terminal mode).

This is probably mostly a subjective question, but if ever by chance, it'd happen that a consensus may emerge, why not...

Example:
```
Goal exists x : nat, x = x.
eexists. (* ?x = ?x *)
set (v := ?x).
```
produces
```
Error: ?x is a generated name. Only user-given names for existential
variables can be referenced. To give a user name to an existential variable,
introduce it with the ?[name] syntax.
```
rather than:
```
Error:
?x is a generated name. Only user-given names for existential variables can
be referenced. To give a user name to an existential variable, introduce it
with the ?[name] syntax.
```

Note that this implicitly means following the policy of not using an extra `hbox` when calling `UserError` (because there is already one in `Topfmt.make_body`), but the policy seems to be followed.

### DIFF
--- a/lib/cErrors.ml
+++ b/lib/cErrors.ml
@@ -112,7 +112,7 @@ let iprint_no_report (e, info) =
 
 let _ = register_handler begin function
   | UserError(s, pps) ->
-    hov 0 (where s ++ pps)
+    where s ++ pps
   | _ -> raise Unhandled
 end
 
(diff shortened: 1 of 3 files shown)


ejgallego:
This seems reasonable, the way we print errors has bugged me a bit. Tho we should be careful to check that PG and other tools using fragile `coqtop` parsing are OK with this change.

ejgallego (approved):

ejgallego:
> but the policy seems to be followed.

Yup, IMO it makes more sense to require users not to add any boxing and let the handler take care of it.

ejgallego:
Previsibly you need to adapt the test suite [recall we have promote target]

PR #9147, opened 2018-12-05
[dune] [doc] Support for building the reference manual with Dune.

ejgallego:
This is a reduced version of #8503 as to provide a way to build the
reference manual with Dune.

Dune 1.6 supports (experimentally) directories as targets, thus we
introduce a rule that will call `sphinx` to build the manual.

This only provides build, however generation of the `.install` file is
still incomplete, and it will be hopefully addressed in #8503.


doc/tools/coqrst/coqdoc/main.py
```diff
@@ -35,7 +35,7 @@
 
 def coqdoc(coq_code, coqdoc_bin=None):
     """Get the output of coqdoc on coq_code."""
-    coqdoc_bin = coqdoc_bin or os.path.join(os.getenv("COQBIN"), "coqdoc")
+    coqdoc_bin = coqdoc_bin or os.path.join(os.getenv("COQBIN") or "", "coqdoc")
     fd, filename = mkstemp(prefix="coqdoc-", suffix=".v")
     if platform.system().startswith("CYGWIN"):
         # coqdoc currently doesn't accept cygwin style paths in the form "/cygdrive/c/..."
```
cpitclaudel: You can pass a default to getenv instead of using `or ""`. But I'm not sure this is a good idea: do you really want to run the system-global coqdoc here?
This code was written with the assumption that callers would either set COQBIN or pass and explicit coqdoc_bin

ejgallego: > You can pass a default to getenv instead of using `or ""`. But I'm not sure this is a good idea: do you really want to run the system-global coqdoc here?

Yes, we want to run with the `coqdoc` in `PATH` here; using `COQBIN` has a problem as it doesn't properly compose when building several packages together.

Should I update and remove the `or` then?

ejgallego: Dune composes builds _à la Nix_ , that is to say, `PATH` is modified so you have the right scope. Starting in Dune 1.6 we can also scope private binaries, which helps with some problems we have with the test suite and `fake_ide`.

ejgallego:
Build by this PR https://coq.gitlab.io/-/coq/-/jobs/130461993/artifacts/_build/default/doc/sphinx_build/html/index.html

Reading it now to verify there are no oddities.

ejgallego:
> Reading it now to verify there are no oddities.

At a quick glance the build seems fine, thus this should be ready modulo the comments by Clément; IMHO as the refman will go on its own Debian/OPAM/Nix package having the possibility to use the installed `coqtop` does make sense to me.

doc/tools/coqrst/coqdoc/main.py
cpitclaudel: Yes, changing the or would be good, then. It's too bad that this will make mistakes when running the script on its own trickier to spot, but I don't think that's an important enough issue to delay this PR.

ejgallego: I see what you mean, on the other hand I'd dare to say running via path is a pretty common occurrence these days in sandboxed-style builds. For example I think that people on Nix should not have any other `coqtop` in scope.

Having to pass parameters becomes quickly a mess too if you context is large tho, and tends to have portability issues.

ejgallego: Amended; I am still open as how to improve the general setup for people that don't like to use `PATH` in general; this is an important issue as we have two core dev setups [Nix and Dune] that rely on it, so we must try to accommodate both options in the best possible way.

cpitclaudel: Thanks. I think the code is OK as is. The problem with picking up the wrong Coq version would only pop up if you ran the script standalone, not as part of the Makefile or through Dune. Do we have to worry about this? I think probably not. If we do want to worry, we could probably add another env var.

Zimmi48: Indeed, I don't think that we should have to worry about this. Soon Dune will be the only supported way of building BTW.

cpitclaudel: I meant that with this change, sphinx-build alone in the refman directory may yield surprising results. That's not a huge issue.

.gitlab-ci.yml
```diff
@@ -284,13 +284,27 @@ doc:refman:
...
+  variables:
+    OPAM_SWITCH: edge
+    DUNE_TARGET: refman
+  artifacts:
+    name: "$CI_JOB_NAME"
+    paths:
+      - _build/default/doc/sphinx_build/html
+
 doc:ml-api:odoc:
+  <<: *dune-template
```
SkySkimmer: This is not actually simplifying anything, don't use template when everything in it except for a trivial line gets overridden.

SkySkimmer (changes requested):

Makefile.dune
```diff
@@ -18,6 +18,7 @@ help:
 	@echo "  - quickopt:   build main ML files [coqtop + plugins + ide + printers] using the optimizing compiler"
 	@echo "  - test-suite: run Coq's test suite"
 	@echo "  - release: build Coq in release mode"
+	@echo "  - refman:  build Coq's reference manual"
```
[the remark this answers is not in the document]
ejgallego: Done, thanks.

.gitlab-ci.yml
ejgallego: Well, to me it servers a very important function here which is documentation; also it is a seed for future refactoring so I'd be in favor of keeping it as it doesn't create an obvious problem.

Makefile.dune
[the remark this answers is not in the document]
ejgallego: I've also moved from the other PR the refactoring of the PHONY line.

.gitlab-ci.yml
SkySkimmer: What does it document?
As for future refactoring, don't anticipate it by making things more complicated.

ejgallego: It documents this target is using a Dune-based build.

SkySkimmer: >It documents this target is using a Dune-based build.

Not worth it IMO. Please change.

Zimmi48: Let's not block this PR on this unrelated change. @ejgallego you can push it again once the benefits are clearer so that @SkySkimmer doesn't oppose.

ejgallego: I'm afraid we face a difference of opinion here. In particular I don't want to have N targets when we move from `make -f Makefile.dune` to `dune` itself; and doing a search for `dune-template` is important in my workflow.

SkySkimmer: Then move everything except `make -f Makefile dune` from dune-template to build:dune since it's specialised for it.

ejgallego: `dune-template` is already in `master`, thus apart from quite deep nitpicking indeed the comment is unrelated to this PR.

ejgallego: Ok, let's see now.

.gitlab-ci.yml
```diff
@@ -284,15 +300,23 @@ doc:refman:
...
+  <<: *dune-ci-template
   variables:
-    OPAM_SWITCH: edge
+    <<: *dune-ci-template-vars
+    DUNE_TARGET: apidoc
   artifacts:
-    name: "$CI_JOB_NAME"
+    <<: *dune-ci-template-artifacts
     paths:
       - _build/default/_doc/
```
SkySkimmer: NB: this change means this artifact now has an expire key. (maybe make a note in the commit message)

SkySkimmer (approved):

.gitlab-ci.yml
ejgallego: Ok. I do prefer to expire by default, specially in ephemeral builds; but of course this could be overriden if our deployment strategy changes.

ejgallego: Maybe one month could be a more sensible outcome here tho.

ejgallego:
Well, this has a problem that indeed I just realized:
- our default dune target is `dune build @install`,
- we now install the `coq-refman` package,
- thus the refman is built on the base target,

The `doc:refman:dune` target should not build anything, just upload the artifact.

There are different ways to solve this, maybe for now we should just guard the build of the refman alias somehow.

SkySkimmer:
I don't know about "people" in general, but for CI at least we can be more specific in what we want built/installed.

Makefile.dune
```diff
@@ -10,15 +13,20 @@ BUILD_CONTEXT=_build/default
...
-	@echo "  - quickbyte:  build main ML files [coqtop + plugins + ide + printers] using the bytecode compiler"
-	@echo "  - quickopt:   build main ML files [coqtop + plugins + ide + printers] using the optimizing compiler"
-	@echo "  - test-suite: run Coq's test suite"
-	@echo "  - release: build Coq in release mode"
-	@echo "  - apidoc:  build ML API documentation"
+	@echo ""
+	@echo "  - states: build a minimal functional coqtop"
+	@echo "  - world:  build all binaries and libraries"
+	@echo "  - watch:  build all binaries and libraries [continuous build]"
+	@echo "  - check:  build all ML files as fast as possible [requires Dune >= 1.5.0]"
```
Zimmi48: Well know the whole thing needs Dune >= 1.6, no?

Zimmi48 (commented):
@ejgallego I'm waiting on your green light before merging this.

rgrinberg:
> People should just not use the default target. WDYT?

Note that the default target can be changed to be whatever you want it to:

```
(alias
 (name default)
 (deps foo bar baz ..))
```

Zimmi48:
Thanks @rgrinberg. @ejgallego Could you change the default to `coq.install`? This would avoid confusion.

ejgallego:
I will tweak it, maybe I use a different method tho, like enabling the alias only iff an env var is set.

ejgallego:
Ok, after some thinking I am convinced now that the right strategy for this PR is not to enable the install target for now. Anyways the resulting install file was unusable. Thus, the refman will be only be built if the target is specified explicitly.

dune-project
```diff
@@ -1,3 +1,2 @@
-(lang dune 1.4)
-
+(lang dune 1.6)
```
Zimmi48: There are a few place in the doc that needs updating to reflect this: `Makefile.dune` and `dev/doc/build-system.dune.md`.

Zimmi48 (approved):
Last round of comments (totally optional ones):
What about updating `dev/doc/build-system.dune.md` to mention that there is experimental support for building the HTML version of the reference manual with Dune?

dev/doc/build-system.dune.md
```diff
@@ -10,7 +10,7 @@ Coq can now be built using [Dune](https://github.com/ocaml/dune).
 
 ## Quick Start
 
-Dune >= 1.5.0 is recommended, see `dune-project` for the minimum
+Dune >= 1.6.0 is recommended, see `dune-project` for the minimum
```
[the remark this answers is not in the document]
ejgallego: Umm, we need to be careful here, I'll had an "usually"

ejgallego:
> @ejgallego I have just pushed to your branch a commit updating `default.nix` so that it provides Dune 1.6 and it remains possible to build Coq with Dune for Nix users.

Oh thanks!

> What about updating `dev/doc/build-system.dune.md` to mention that there is experimental support for building the HTML version of the reference manual with Dune?

Sure.

ejgallego:
Thanks for all your help!

PR #9150, opened 2018-12-06
[doc] Enable Warning 50 [incorrect doc comment] and fix comments.

ejgallego:
This is a pre-requisite to use automated formatting tools such as
`ocamlformat`, also, there were quite a few places where the comments
had basically no effect, thus it was confusing for the developer.

p.s: Reading some comments was a lot of fun :)


aspiwack:
This is a trivial commit, it doesn't require 20 reviewers :slightly_smiling_face: you should probably remove most.

ejgallego:
Thanks Arnaud, no worries, I guess that as you say this is straightforward, maintainers interesting in docs will have a look, the rest may ignore it as I don't think we will require full approval from codeowners for this one.

interp/constrintern.mli
```diff
@@ -45,13 +45,15 @@ type var_internalization_type =
 
 type var_internalization_data =
     var_internalization_type *
-      (** type of the "free" variable, for coqdoc, e.g. while typing the
-	  constructor of JMeq, "JMeq" behaves as a variable of type Inductive *)
+      (* type of the "free" variable, for coqdoc, e.g. while typing the
```
SkySkimmer: Is there no way to put a comment on part of a tuple type?

ejgallego: No that I know of, other than using a record. Likely there are some mantis / gihub issues about this. Recall that `ocamldoc` has seen little to 0 development over the years.

SkySkimmer:
> No that I know of, other than using a record. Likely there are some
 > mantis / gihub issues about this. Recall that |ocamldoc| has seen little
 > to 0 development over the years.

What about in odoc?

Gaëtan Gilbert

ppedrot:
Why removing the double-star comments? It has a semantic meaning in ocamldoc.

ejgallego:
@ppedrot precisely this PR fixes those comments that have an incorrect meaning for ocamldoc, thus triggering warning 50.

@Skyskimmer I think that particular feature would need help from the OCaml parser, that being said 4.08 has a new parser so maybe they extend the AST. Odoc has been more focused over higher level issues, I feel.

ppedrot (approved):

ejgallego:
cc: #7155

ejgallego:
I think there is no point on delaying this a lot tho, I feel like we will gain little and it is IMO important that we enable the check so developers can start adapting to the new warning, which is always annoying.

Natural assignee would be somebody from the API doc team. @maximedenes?

SkySkimmer:
Should .merlin files be changed?

ejgallego:
> Should .merlin files be changed?

Indeed, tho I'm not sure merlin supports this particular warning.

ejgallego:
Asking for assignee and prompt merge as not point on delaying this as it will only make us all lose time.

ejgallego:
Ping !

maximedenes:
Will have a look soon.

ejgallego:
Thanks @maximedenes

maximedenes (approved):

PR #9155, opened 2018-12-06
Fix race condition triggered by fresh universe generation

maximedenes:
Remote counters were trying to build universe levels (as opposed to
simple integers), but did not have access to the right dirpath at
construction time. We fix it by constructing the level only at use time,
and we introduce some abstractions for qualified and unqualified level
names.

### DIFF
--- a/pretyping/detyping.ml
+++ b/pretyping/detyping.ml
@@ -589,8 +589,23 @@ let detype_cofix detype avoid env sigma n (names,tys,bodies) =
        Array.map (fun (_,_,ty) -> ty) v,
        Array.map (fun (_,bd,_) -> bd) v)
 
+(* TODO use some algebraic type with a case for unnamed univs so we
+   can cleanly detype them. NB: this corresponds to a hack in
+   Pretyping.interp_universe_level_name to convert Foo.xx strings into
+   universes. *)
+let hack_qualid_of_univ_level sigma l =
+  match Termops.reference_of_level sigma l with
+  | Some qid -> qid
+  | None ->
+    let path = String.split_on_char '.' (Univ.Level.to_string l) in
+    let path = List.rev_map Id.of_string_soft path in
+    Libnames.qualid_of_dirpath (DirPath.make path)
+
 let detype_universe sigma u =
-  let fn (l, n) = Some (Termops.reference_of_level sigma l, n) in
+  let fn (l, n) =
+    let qid = hack_qualid_of_univ_level sigma l in
+    Some (qid, n)
+  in
   Univ.Universe.map fn u
 
 let detype_sort sigma = function
@@ -611,7 +626,7 @@ let detype_anonymous = ref (fun ?loc n -> anomaly ~label:"detype" (Pp.str "index
 let set_detype_anonymous f = detype_anonymous := f
 
 let detype_level sigma l =
-  let l = Termops.reference_of_level sigma l in
+  let l = hack_qualid_of_univ_level sigma l in
   GType (UNamed l)
 
 let detype_instance sigma l = 
--- a/checker/values.ml
+++ b/checker/values.ml
@@ -93,9 +93,9 @@ let v_cons = v_tuple "constructor" [|v_ind;Int|]
 
 
 (** kernel/univ *)
-let v_level_qualid = v_tuple "Level.Qualid.t" [|v_dp;Int|]
+let v_level_global = v_tuple "Level.Global.t" [|v_dp;Int|]
 let v_raw_level = v_sum "raw_level" 2 (* Prop, Set *) 
-  [|(*Level*)[|v_level_qualid|]; (*Var*)[|Int|]|]
+  [|(*Level*)[|v_level_global|]; (*Var*)[|Int|]|]
 let v_level = v_tuple "level" [|Int;v_raw_level|] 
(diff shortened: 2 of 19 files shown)


SkySkimmer:
I don't really like the way it looked so I pushed a commit to change it. See commit message / diff.

SkySkimmer:
Also if this is a fix can we have a test suite file?

maximedenes:
I like your commit, but it seems to break the build, doesn't it?

> Also if this is a fix can we have a test suite file?

It seems very hard to test this race condition. However #9087 will hopefully serve as a test, since it revealed the problem.

gares (approved):

SkySkimmer:
>I like your commit, but it seems to break the build, doesn't it?

funind is hell
Let's see now.

pretyping/detyping.ml
```diff
@@ -589,8 +589,23 @@ let detype_cofix detype avoid env sigma n (names,tys,bodies) =
        Array.map (fun (_,_,ty) -> ty) v,
        Array.map (fun (_,bd,_) -> bd) v)
 
+(* TODO use some algebraic type with a case for unnamed univs so we
+   can cleanly detype them. NB: this corresponds to a hack in
+   Pretyping.interp_universe_level_name to convert Foo.xx strings into
+   universes. *)
```
mattam82: Indeed, both sides of this hack were made so that pretype (detype c) = c w.r.t universes, AFAIR. This was needed for funind but is not a bad property to have in general.

mattam82 (approved):
I’m fine with the patch. I’m still wondering why workers ever get to run code in a state where the modpath is wrong though (I suppose it contains DirPath.initial here), I spent last afternoon tracking such a problem in my IR branch. What is the rationale for this?

maximedenes:
> If the current module is not properly set in workers

This PR is not about workers, remote counters are used by the manager threads in the main process. The real problem is that the system is imperative, so the module path in the global `safe_env` may have changed before/while we emit the universes.

> I’m still wondering why workers ever get to run code in a state where the modpath is wrong though

If that's the case, it's probably a bug, but it is orthogonal to this PR, which does not touch workers.

maximedenes:
> What are threads used for? Aren't they overkill if their only role is input/ouput?

@gares can probably answer that, but this is not introduced by this PR ;)

I basically add an abstraction of global level names, which is useful for other purposes, as it reveals a number of hacks on universes (in pretyping/detyping, in `sort_universes`, etc).

maximedenes:
Can this get an assignee btw?

SkySkimmer:
I can assign if people don't mind that part of the code is from me.

gares:
Threads could be replaces by explicit control flow passing (plus async io calls, Unix.select style).
Explicit control flow amounts to writing *the entire code* in the lwt monad.
@ppedrot had a branch, but was never finalized because the task is huge.

I would not say overkill, if OCaml gets real multi thread support (as in multicore ocaml, that is getting quite serious these days) then the worker managers can take real advantage of parallelism, especially if we manage to structure the state and marshall it on demand.

ppedrot:
> @ppedrot had a branch, but was never finalized because the task is huge.

Maybe not so huge. I would set it on the same order of magnitude as EConstr, that is definitely not something you can do in one afternoon and without mental health support, but not out of the reach of a change between two major Coq versions.

maximedenes:
@gares can you be the assignee?

maximedenes:
(otherwise, I think it's fine if @SkySkimmer takes it)

ejgallego:
This may be a very silly question but why are we using a global reference instead of keeping track of this in the `env/evar_map` in a pure way.

mattam82:
@ejgallego that would be better, but unless you also prefix the name of the universe with the name of the thing you are producing (e.g. a Definition "U.foo" or a Variable (x : Type) in U) you would get unexpected aliasing.

ejgallego:
@mattam82 isn't the env threaded properly as to allow a global view? I am not familiar with the code so I am not sure where univ get created. I guess mainly at pretyping, right?

Indeed I am wondering, if univs can be created inside proofs how is this supposed to work with delegated proofs for example? In this case, the scoping you propose could maybe help a great deal.

gares:
If we prefix the name of the universes with the name of the constant *and* the name/id of the eventual worker in which it is generated, then there is no need to have a "remote counter", names would be unique by construction.

mattam82:
Ah right, I forgot about the worker id.

gares:
I don't know if "fresh" universes need to be declared somewhere. If it is not the case, then having unique univ ids by construction is probably doable. If they have to be registered in, say, the evar map or the env, then things may get hard.

SkySkimmer:
I think we could have local names for universes during a proof (in the evar map) and rename them at the end but it would take some work.
Also side effects may make it harder.

gares:
Promoting a fresh universe to a global one seems orthogonal to registering it in the evar map during proof construction.

Point is: if all fresh universes needs to be recorded in a single place, then distributing a computation among multiple entities requires communication and this is what we have today.

ejgallego:
Interesting questions folks, I'm afraid I don't know enough about this topic, however I hope I can learn a bit about how this is supposed to work for the opaque proof case. If I am not mistaken ppedrot asked some time ago but we found no documentation, maybe we could look at a concrete part of the implementation.

PR #9167, opened 2018-12-08
Fixes #9166: no deprecation warning on aliases used as pattern variables

### LINKED ISSUE #9166: Compatibility notation warning can be triggered by binding a pattern
#### Version

8.10+alpha

#### Operating system

macOS

The `[compatibility-notation,deprecated]` warning can be triggered by binding the notation's name in a pattern. The deprecated name is properly used as a pattern variable and not the notation, but the compatibility/warning system still thinks the notation is used.

As a workaround, one can simply change the bound name to anything not deprecated, but this comes from a real example in `hs-to-coq`'s [Enum.v](https://github.com/antalsz/hs-to-coq/blob/82a64710760156de0fa3720b302ae4d6a0eec41e/examples/base-src/manual/GHC/Enum.v#L319).

```
Require Import NArith.

Definition foo (x: nat) : nat :=
  match x with
  | 0 => 0
  | S Ngt => Ngt
  end.
```

herbelin:
**Kind:** bug fix 

Fixes / closes #9166

A rather ad hoc fix though.

### DIFF
--- a/interp/syntax_def.ml
+++ b/interp/syntax_def.ml
@@ -105,3 +105,10 @@ let search_syntactic_definition ?loc kn =
   let def = out_pat pat in
   verbose_compat ?loc kn def v;
   def
+
+let search_filtered_syntactic_definition ?loc filter kn =
+  let pat,v = KNmap.find kn !syntax_table in
+  let def = out_pat pat in
+  let res = filter def in
+  (match res with Some _ -> verbose_compat ?loc kn def v | None -> ());
+  res
--- a/test-suite/bugs/closed/bug_9166.v
+++ b/test-suite/bugs/closed/bug_9166.v
@@ -1,0 +1,9 @@
+Set Warnings "+deprecated".
+
+Notation bar := option (compat "8.7").
+
+Definition foo (x: nat) : nat :=
+  match x with
+  | 0 => 0
+  | S bar => bar
+  end.
--- a/interp/constrintern.ml
+++ b/interp/constrintern.ml
@@ -1527,8 +1527,8 @@ let drop_notations_pattern looked_for genv =
     try
       match Nametab.locate_extended qid with
       | SynDef sp ->
-	let (vars,a) = Syntax_def.search_syntactic_definition sp in
-	(match a with
+        let filter (vars,a) =
+        try match a with
 	| NRef g ->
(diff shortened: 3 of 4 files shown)


ppedrot:
@herbelin it seems this PR contains another unrelated PR about CoqIDE, probably due to some rebase issue. Can your remove it from the PR?

test-suite/bugs/closed/bug_9166.v
```diff
@@ -0,0 +1,9 @@
+Require Import NArith.
+
+Set Warnings "+deprecated".
```
ppedrot: This test is going to become invalid as soon as we remove the deprecated alias from the coq stdlib. You should rather define a local deprecated alias.

interp/syntax_def.ml
```diff
@@ -105,3 +105,10 @@ let search_syntactic_definition ?loc kn =
   let def = out_pat pat in
   verbose_compat ?loc kn def v;
   def
+
+let search_filtered_syntactic_definition ?loc filter kn =
+  let pat,v = KNmap.find kn !syntax_table in
+  let def = out_pat pat in
+  let res = filter def in
+  (match res with Some _ -> verbose_compat ?loc kn def v | None -> ());
+  res
```
ppedrot: Code duplication not quite pretty. Why not export the warning function instead and let the client decide what to do with the result?

herbelin: This was indeed the alternative but I thought it would be more intrusive. Do you prefer I change?

herbelin: My own arbitration was not to factorize, so, considering that a change would take time, I'd do it (only) if you push stronglier than what you're doing in the direction of factorizing.

ppedrot: Fine with me. But you should still fix the test as explained in the above review comment though.

herbelin: It is fixed I think: I'm now using a local notation.

ppedrot (approved):

Zimmi48:
This fix is more important for 8.9 than for 8.10 given that we are going to remove these compatibility aliases. The fix should be backportable since it only extends the API. Thus, I'm proposing to put this in 8.9.0.

Issue #8770, opened 2018-10-18
What are the shelves for?


mattam82:
I'd like to get a confirmation of my current understanding of the shelves in Proof and Proofview. 

It seems that the shelf in Proofview should only be of temporary use. Each run_tactic gathers the Proofview's shelf at the end of the tactic run and puts that in the Proof.t shelf. It resets the future_goals of the current Proofview evar_map, and its local shelf. Then the Unshelve command calls Proof.unshelve which empties its current (global) shelf and puts it as subgoals in the current proofview. It should be an invariant that run_tactic always starts with a proofview with an empty (local) shelf but that's unclear to me. @aspiwack @herbelin can you confirm that view?

I'm wondering if a different view could be used for the Proofview shelf, allowing tactics such as Proofview.unshelve to manipulate the local shelf, e.g. to create new subgoals from evars.
Imagine use cases like `evar (?a : nat); exact (some_complicated_term a); unshelve ?a`. It looks technically doable. Another use case I have is for typeclass resolution: say I apply a tactic t to a goal which generates some subgoals, on which resolution succeeds, but also shelved subgoals on which I would like resolution to be performed as well. In that case I need to call a kind of [unshelve_goals] tactic on the current proofview shelf. Does that sound reasonable?

There are also use cases for typeclass resolution where I want to leave the shelved goals shelved (typically this allows to implement forward hints).

ejgallego:
cc: #8386 , #6292

mattam82:
Haha. Well now I don't think anymore that they can be "merged" really. It's more like the Proofview shelf is a temporary receptacle for globally shelved goals in the Proof.t shelf, it seems. OR maybe we can keep everything in the Proofview's shelf and maintain it there, if that plays well with focusing.

aspiwack:
Sorry… I had forgotten about this :confused: 

> It should be an invariant that run_tactic always starts with a proofview with an empty (local) shelf but that's unclear to me.

This is correct. The shelve in the proofview was conceived of as as "writer" effect. Every tactic execution is supposed to start with an empty local shelf, and they would all be flushed to the global shelf at the end of the tactic execution.

> Imagine use cases like evar `(?a : nat); exact (some_complicated_term a); unshelve ?a`

[...]

Issue #8047, opened 2018-07-11
eauto can leave shelved evars


tchajed:
#### Version

```
The Coq Proof Assistant, version 8.9+alpha (July 2018)
compiled on Jul 10 2018 11:43:30 with OCaml 4.06.0
```

Coq b4572185522

#### Operating system

macOS 10.13.5

#### Description of the problem

In the following example, the proof of `P` goes through `Pred` and required some value of type `T` (which is unused). `eauto` cannot find such a value but it nonetheless creates an evar and surprisingly shelves and solves the goal. I would rather have `eauto` fail in this case. Interestingly, `exact ltac:(eauto)` _does_ fail and complains about not being able to find an instance for `T`, which is what makes me feel this is a bug.

```
Axiom T:Type.
Axiom x:T.

Inductive Pred (p:T) : Prop := PredPf.
Hint Resolve PredPf.

Axiom P:Prop.

Lemma P_from_Pred :
  forall p, Pred p -> P.
Proof.
Admitted.
Hint Resolve P_from_Pred.

Lemma buggy: P.
Proof.
  eauto.

  Grab Existential Variables.
  exact x.
Qed.
```

Zimmi48:
This was also mentioned two days ago by @fakusb on Gitter:

```
Goal forall A (P : A -> Prop), (forall a : A, P a) -> exists a, P a.
eauto
```

will behave as:
```
eexists.
apply H.
```

and thus leave evars on the shelf.

As mentioned there, @mattam82 wanted to have the new `typeclasses eauto` tactic fail if there are variables left on the shelf. This behavior was finally reverted because this can be used to instrument automatic tactics to do only some proof steps instead of solving the goal completely.

In the process of moving `eauto` to use the implementation of `typeclasses eauto` we could consider changing this behavior by default.

JasonGross:
You're proposing to define `eauto` as `try solve [ unshelve typeclasses eauto with core ]`, right?

Zimmi48:
Something close to this. Anyways, my intent is to keep many modifiable parameters but yes for the default.

Issue #6202, opened 2017-11-20
ltac:(...) loses evars from the shelf when it does not fully solve its hole


JasonGross:
#### Version

8.7.0 (also a recent master)

#### Operating system

Linux

#### Description of the problem

```coq
Inductive foo := F (a : forall var : Type -> Type, unit -> var unit) (_ : a = a).
Goal foo.
  eexists (fun var => fun u : unit => ltac:(clear u)).
  shelve.
  Unshelve.
  Fail all:[ > | ]. (* Error: Incorrect number of goals (expected 1 tactic, was given 2). *)
  shelve.
  Grab Existential Variables.
  all:[ > | ]. (* success *)
```

Or, a shorter example:
```coq
Goal True.
  refine (ltac:(idtac)).
  Unshelve.
  Fail all: [ > ]. (* Error: Incorrect number of goals (expected 0 tactics, was given 1). *)
  Grab Existential Variables.
  all: [ > ]. (* success *)
```

herbelin:
This seems to be fixed actually (suspectingly in 6f30f76d or close in time).

Issue #4509, opened 2016-01-22
Two shelves exist, and unshelve/Unshelve only reaches one


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#4509
From: @jonleivent
Reported version: 8.5
CC:   @ppedrot

herbelin:
This seems to have been solved (suspectingly in #6328). For instance, the following is working:
```
Goal exists n, n > 1.
Proof.
  unshelve eexists. (*2 goals, as expected*)
  Undo.
  unshelve (eexists; instantiate (1:=ltac:(idtac))). (*only 1 goal*)
  shelve.
  Undo.
  2:unshelve instantiate (1:=_).
```

PR #9170, opened 2018-12-09
[vernac] Remove obsolete Existential and Grab Existential variables.

ejgallego:
This is a proposal to eventually remove the commands `Existential` and
`Grab Existential`. To my understanding these commands are supposed to
be deprecated and scheduled to removal, but not 100% sure so we would
need to confirm.

This would of course need a deprecation phase, let's see the impact on
CI first.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,9 @@ Specification language, type inference
   solved by writing an explicit `return` clause, sometimes even simply
   an explicit `return _` clause.
 
+- The commands `Existential` and `Grab Existential` have been
+  removed. Use XXX in its place.
+
 Kernel
 
 - Added primitive integers
--- a/doc/sphinx/proof-engine/proof-handling.rst
+++ b/doc/sphinx/proof-engine/proof-handling.rst
@@ -247,25 +247,6 @@ Name a set of section hypotheses for ``Proof using``
 
          Collection Many := Fewer - (x y)
 
-
-
-.. cmd:: Existential @num := @term
-
-   This command instantiates an existential variable. :token:`num` is an index in
-   the list of uninstantiated existential variables displayed by :cmd:`Show Existentials`.
-
-   This command is intended to be used to instantiate existential
-   variables when the proof is completed but some uninstantiated
-   existential variables remain. To instantiate existential variables
-   during proof edition, you should use the tactic :tacn:`instantiate`.
-
-.. cmd:: Grab Existential Variables
-
-   This command can be run when a proof has no more goal to be solved but
-   has remaining uninstantiated existential variables. It takes every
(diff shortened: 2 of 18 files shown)


herbelin:
For `Grab Existentials`, my memories are that there are indeed cases where `Unshelve` misses some evars, as a consequence of cases where an evar fails to be correctly tracked in the shelf. I vaguely remember that @mattam82 reported a case once (but I don't know how to find it again). So, I guess the idea would indeed be to first ensure that all evars are tracked in either the shelf, the given_up list or the main list (@aspiwack, @ppedrot, is it correctly the idea?).

Zimmi48:
I wanted to do this too but the problem was indeed that we have a few bugs around that make it necessary to rely on `Grab Existential Variables`. This PR should be a good way of finding the remaining uses in external projects and whether `Unshelve` can be used instead.

aspiwack:
@herbelin it sounds correct, indeed.

ejgallego:
Ok, so I've tracked down a few issues that could be relevant:

https://github.com/coq/coq/issues/9171
https://github.com/coq/coq/issues/8047
https://github.com/coq/coq/issues/6202
https://github.com/coq/coq/issues/4509
https://github.com/coq/coq/issues/8770
https://github.com/coq/coq/issues/5570

CHANGES.md
```diff
@@ -39,6 +39,9 @@ Specification language, type inference
   solved by writing an explicit `return` clause, sometimes even simply
   an explicit `return _` clause.
 
+- The commands `Existential` and `Grab Existential` have been
```
gares: The command is `Grab Existential Variables`, and `xxx` should be fixed before merging

gares (changes requested):

gares:
I'm positive on this. Also, some users asked for a sort of "automatic unshelve and eventually solve" tactic to be run when one attempts a Qed and there are still evars on the shelf. At least this is what I remember of the discussions we had, CC @CohenCyril

gares:
I mean, it may be worth understanding is all use cases of `Unshelve` fall in this category (you unshelve at the very end and you run the same tactic).

gares:
Yes

ejgallego:
@ppedrot I was doing a large clenaup of proof global global etc... And I was wondering about this functions. I don't recall now if these functions pose a true problem but will update the PR based on your remarks next week.

The current or was maybe more of an experiment to see impact.

ejgallego:
Indeed I confirm this is not needed for any of the pending projects related to document / UI / proof cleanup.

However it does seem to me that the commands are deprecated so indeed we should discuss a bit more before closing IMHO.

ejgallego:
Closed in favor of #9545

PR #9172, opened 2018-12-09
[proof] Rework proof interface.

ejgallego:
- deprecate the old 5-tuple accessor in favor of a view record,

- move `name` and `kind` proof data from `Proof_global` to `Proof`,
  this will prove useful in subsequent functionalizations of the
  interface, in particular this is what abstract, which lives in the
  monads, needs in order no to access global state.

- Note that `Proof.t` and `Proof_global.t` are redundant anyways.

### DIFF
--- a/dev/ci/user-overlays/09172-ejgallego-proof_rework.sh
+++ b/dev/ci/user-overlays/09172-ejgallego-proof_rework.sh
@@ -1,0 +1,9 @@
+if [ "$CI_PULL_REQUEST" = "9172" ] || [ "$CI_BRANCH" = "proof_rework" ]; then
+
+    ltac2_CI_REF=proof_rework
+    ltac2_CI_GITURL=https://github.com/ejgallego/ltac2
+
+    mtac2_CI_REF=proof_rework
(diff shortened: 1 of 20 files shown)


ppedrot (approved):
I haven't seen anything suspicious, and it looked like a necessary cleanup indeed.

ejgallego:
Thanks for the review @ppedrot , rebasing to get a CI run.

ejgallego:
Overlays at:
- https://github.com/ppedrot/ltac2/pull/91
- https://github.com/Mtac2/Mtac2/pull/134

@ppedrot  the more I see this PR the more I wonder if we are doing the right thing here.

On one hand this PR helps to highlight some potential places where using abstract is an issue, on the other one, it seems to me that we should maybe fix abstract as not to need this information in the first place, in the lines @SkySkimmer pointed out [have side effects to be saved only on the save_proof path, thus we don't need to know polymorphic status / name until the constant is actually saved]

I am specially uncomfortable about how we handle interpretation hooks.

ejgallego:
Failure is spurious.

ejgallego:
Merci!

PR #9174, opened 2018-12-09
[vernac] Remove `Proof with tac.` idiom.

ejgallego:
It was mentioned on Gitter that we may want to deprecate this; adding
this PR to study impact.


ejgallego:
> You meant `Proof with`, no? You seem courageous. I suspect that this is used a lot. What would be the suggested replacement?

Indeed, sorry for the confusion. I had this on my list of notes, but I don't remember more context, maybe it was @ppedrot who suggested we could replace it?

I think there would be no suggested replacement for the general case, for the proof closing forms one could use `now tac` or `by tac` [which tends to be faster, fine tuning this things is interesting]

I will relaunch the CI as the results are bad due to transitory failures. [We are overloading the artifacts cache I think :)]

aspiwack:
I'm curious, what is the impetus to remove it?

ejgallego:
I was too optimistic to assume this would work; we'd have to reify the data attached to proof tho so it is safe to transmit to workers.

ejgallego:
> I was too optimistic to assume this would work; we'd have to reify the data attached to proof tho so it is safe to transmit to workers.

So the question is, should we deprecate it?

When we answer that question we can close this PR.

maximedenes:
> we'd have to reify the data attached to proof tho so it is safe to transmit to workers

Does the feature rely on non-algebraic data in the proof state (like the closures for terminators)? Can you give a pointer to the code?

ejgallego:
> Does the feature rely on non-algebraic data in the proof state (like the closures for terminators)? Can you give a pointer to the code?

It relies on the AST which indeed uses genargs; genargs are not so easily serializable [we do that in SerAPI] so indeed I tried this; however I guess the price is too large wrt gain; I dunno.

```
  endline_tactic : Genarg.glob_generic_argument option;
```

ejgallego:
In general all the parts of state that contain tactics are a bit problematic, if only because they are in a plugin so more advanced serialization strategies will be complex.

gares:
I don't remember any particular problem with the default tactic in the STM, and I don't see the link between "proof terminators" and "Proof with".

Can someone explain what would be the gain and where the problem is?

Note that marshaling genargs must work since the tactic AST is full of these, and we surely marshal the AST of tactics to delegate proof tasks.

ejgallego:
Precisely the idea would be to get rid or Marshal, but indeed this is a very minor step so given the difficulty we can forget about it now.

PR #9178, opened 2018-12-09
CoqIDE: Restoring configuration of default width/height of main window.

herbelin:
**Kind:** feature

While working on CoqIDE code, I discovered the existence of a code to configure the default width/height of main window in the preference file, something which I would have liked to have.

This short PR reactivates it. I don't know exactly what is the history of motivations for having or not having it, so I may not adopt the right approach for supporting that. Help welcome, in particular from @gares or @ppedrot who I realized made an impressive work at maintaining/extending CoqIDE in the past.

Incidentally removing dead code for configuring showing toolbar in preferences, since this is currently managed by a menu.

### DIFF
--- a/ide/preferences.ml
+++ b/ide/preferences.ml
@@ -815,33 +815,20 @@ let configure ?(apply=(fun () -> ())) parent =
     custom ~label box callback true
   in
 
-(*
-  let show_toolbar =
-    bool
-      ~f:(fun s ->
-	    current.show_toolbar <- s;
-	    !show_toolbar s)
-      "Show toolbar" current.show_toolbar
-  in
   let window_height =
     string
-    ~f:(fun s -> current.window_height <- (try int_of_string s with _ -> 600);
-	  !resize_window ();
-       )
-      "Window height"
-      (string_of_int current.window_height)
+    ~f:(fun s -> try window_height#set (int_of_string s) with _ -> ())
+      "Default window height at starting time"
+      (string_of_int window_height#get)
   in
+
   let window_width =
     string
-    ~f:(fun s -> current.window_width <-
-	  (try int_of_string s with _ -> 800))
-      "Window width"
-      (string_of_int current.window_width)
+    ~f:(fun s -> try window_width#set (int_of_string s) with _ -> ())
+      "Default window width at starting time"
+      (string_of_int window_width#get)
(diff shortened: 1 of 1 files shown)


gares:
Where is the code actually resizing the window?

herbelin:
> Where is the code actually resizing the window?

There is no code for resizing the window (hence the "needs restart" comment). If you know how to avoid the restart, I could integrate a fix.

The new values are otherwise taken into account in `Coqide.build_ui`.

gares (approved):

ppedrot:
I have an alternative patch that allows to dynamically set this value, but I am facing a semantic issue. Essentially, what does this option mean? GTK has both a notion of size and default size for windows, and they're not the same.

Setting dynamically the default size has no immediate effect, precisely because it is a *default* size. Meanwhile, setting the size changes the appearance of the window, but doesn't not persist as soon as the use changes it manually, which is expected as well. Thus, if the purpose of the this option is really about *default* size, it actually doesn't matter that this requires a reboot of CoqIDE, because *by design* default sizes are only relevant at startup, and therefore this patch is OK.

Contrarily, if the option is expected to change the size of the window, one needs to apply it dynamically. But in this case, I fail to see the usefulness of this behaviour, as the user can always resize it by hand and the value of the option is going to be out-of-sync as soon as this happens.

herbelin:
@ppedrot: your reasoning is perfectly ok to me.

gares:
One could record the size at quit time, and use it at next reboot.

ppedrot:
@herbelin So, this PR is implementing the 'default' semantics. If this is what we want, we can merge it as-is (optionally removing the "requires reboot" part and replacing it with a mention that this is the **default** window size).

herbelin:
> @herbelin So, this PR is implementing the 'default' semantics. If this is what we want, we can merge it as-is (optionally removing the "requires reboot" part and replacing it with a mention that this is the default window size).

Done.

> One could record the size at quit time, and use it at next reboot.

Personally, I prefer a default value which fits with my general habits rather than remembering the last value which may be specific to the last development I made, but if some people are more comfortable with remembering the last value, this can indeed be an extra possibility to propose in the preference panel.

ppedrot (approved):

Issue #8946, opened 2018-11-08
Roadmap to improve the documentation of tactics.


Zimmi48:
Here are some of the things that need to be done IMO:

- [ ] Change the general structure of the Tactics chapter:
  - [ ] Reorganize the sections "Applying theorems" and "Controlling the proof flow" in a part on "Concluding tactics" (`assumption`, `exact`, `reflexivity`, `discriminate`, `contradiction` and `easy`), a part on "Backward reasoning" (with `apply`, `enough`, `f_equal`?) and a part on "Forward reasoning" (consolidating the doc of `apply in`, `specialize`, `pose proof`, `assert`; we critically lack an explicit part on this alternative kind of proof organization that is perfectly supported by Coq but rarely used).
  - [ ] Consolidate the "Rewriting" and the "Performing computations" section in a single section on applying term transformations.
  - [ ] Rename the section "Classical tactics" in "Tactics for classical logic" or "Tactics requiring additional axioms" (and then consolidate with other such tactics).
  - [ ] Move the section "Checking properties of terms" in the Ltac chapter.
  - [ ] Move the section "Non-logical tactics" to the Proof handling chapter.
  - [ ] Remove the "Simple macro tactics" section.
  - [ ] Consolidate the "Decision procedure", "Automating" sections and the Omega, Micromega, ring and field, and nsatz chapters into a single "Decision procedures" chapter.
  - [ ] Consolidate the "Automation" and "Controlling automation" sections and the relevant parts of the Typeclass chapter into a single "User-controlled automation" chapter.
  - [ ] Merge the "Detailed examples of tactics" chapter in the Tactics chapter.
- [ ] Deprecate tactics that are not used anymore: `cut`, `elim`, `case` (vanilla versions), as well as `simple destruct`.
- [ ] Identify and deprecate useless tactics (example: `lapply`).
- [ ] Add as many cross-links as possible between various sections of the Tactics chapter and between the Tactics chapter and the SSReflect chapter.
- [ ] Add many examples.
- [ ] Add warnings about known inconsistencies and known problems so that users are not left wondering.

I am opening this issue after a discussion with @jfehrle. There was also @bollu (IIRC) who had asked me about writing some of the needs regarding documentation.
Feel free to contribute to part of this. Please state it in the issue thread so that we are aware you are working on it. Also feel free to mention additional changes that you feel would be necessary (regarding documentation). I'll open more similar documentation roadmap issues in the future and will add them all to the Documentation project.

[...]

PR #9179, opened 2018-12-10
Identify tacticals and Ltac constructs as "(tactical)" in index

jfehrle:
Add "tactical" as an index NotationObject
Include tacticals in tactics index
Change items in Ltac chapter to tactical::
Document is_* tacticals, move 2 of them to ltac chapter
Plus a few miscellaneous doc tweaks

**Kind:** documentation

### DIFF
--- a/doc/sphinx/README.rst
+++ b/doc/sphinx/README.rst
@@ -161,24 +161,32 @@ Here is the list of all objects of the Coq domain (The symbol :black_nib: indica
 ``.. tacn::`` :black_nib: A tactic, or a tactic notation.
     Example::
 
-       .. tacn:: do @num @expr
+       .. tacn:: apply @term
+          :name: apply
 
-          :token:`expr` is evaluated to ``v`` which must be a tactic value. …
+          This tactic applies to any goal. The argument term is a term well-formed in
+          the local context. …
 
-``.. tacv::`` :black_nib: A variant of a tactic.
+
+``.. tactical::`` :black_nib: A tactical or Ltac construct
     Example::
 
-       .. tacn:: fail
+       .. tactical:: fail
+          :name: fail (tactical)
 
           This is the always-failing tactic: it does not solve any goal. It is
           useful for defining other tacticals since it can be caught by
           :tacn:`try`, :tacn:`repeat`, :tacn:`match goal`, or the branching
           tacticals. …
 
-          .. tacv:: fail @natural
+``.. tacv::`` :black_nib: A variant of a tactic.
+    Example::
 
-             The number is the failure level. If no level is specified, it
-             defaults to 0. …
+       .. tacv:: simple apply @term
+          :name: simple apply
+
+          This is a shortcut for …
+
 
 ``.. thm::`` A theorem.
     Example::
--- a/doc/sphinx/coq-tacindex.rst
+++ b/doc/sphinx/coq-tacindex.rst
@@ -2,6 +2,6 @@
 
 .. hack to get index in TOC
 
--------------
-Tactic index
--------------
+--------------------------
+Tactic and tacticals index
+--------------------------
--- a/doc/sphinx/language/gallina-extensions.rst
+++ b/doc/sphinx/language/gallina-extensions.rst
@@ -603,7 +603,7 @@ The following experimental command is available when the ``FunInd`` library has 
    This command can be seen as a generalization of ``Fixpoint``. It is actually a wrapper
    for several ways of defining a function *and other useful related
    objects*, namely: an induction principle that reflects the recursive
-   structure of the function (see :tacn:`function induction`) and its fixpoint equality.
+   structure of the function (see :tacn:`functional induction`) and its fixpoint equality.
    The meaning of this declaration is to define a function ident,
    similarly to ``Fixpoint``. Like in ``Fixpoint``, the decreasing argument must
    be given (unless the function is not recursive), but it might not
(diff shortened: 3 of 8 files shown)


jfehrle:
What do you think about also identifying tactics that don't modify the proof state in the index (e.g. "swap")?  Not with "(tactical)", with something else.

Should I move the everything under "Checking properties of terms" from the tactics chapter to the Ltac chapter?

BTW, "tacticals" is not a great name; "tactical" is normally an adjective.  But I expect it's been used too long to think about changing it.  Right?

Zimmi48:
I think there was a confusion in our private discussion: tactical is not a generic term for everything that Ltac does but only for "tactic modifiers". Maybe @herbelin or someone else can explain what is generally meant by this terminology.
I think we also want to hear about @ppedrot's views on how to classify things as Ltac / tactics. It seems indeed that it would make sense to stop using the terminology "tactic" for something that *returns* a value instead of modifying the proof state.

jfehrle:
By all means, let's get input from others on how to group items and what the categories should be.  "Tactics and tacticals index" may not work as a title.  Do you have any thoughts on my first 2 questions?

Zimmi48:
>What do you think about also identifying tactics that don't modify the proof state in the index (e.g. "swap")? Not with "(tactical)", with something else.

In the index, I don't think so. Keeping things in alphabetical index is important for easy retrieval. However in the manual itself, yes. There is already a section about this: https://coq.inria.fr/refman/proof-engine/tactics.html#non-logical-tactics.

>Should I move the everything under "Checking properties of terms" from the tactics chapter to the Ltac chapter?

Yes, that was one of my suggestions in #8946.

jfehrle:
>  Keeping things in alphabetical index is important for easy retrieval.

Huh?  Adding a suffix like "(tactical)", "Ltac", etc. won't change the
order in the index.  (Though eventually we may want to add brief overview
summaries of tactics to the beginning of sections so you don't need to
read/skim the entire section as often.)

I'll look at doing some of the easier items you mention in
https://github.com/coq/coq/issues/8946, maybe as a separate commit/PR.

Zimmi48:
Sorry I think I had misread your question. Then, it's fine, although I don't know what to think of this whole business of adding things in parentheses as this is usually done only to distinguish between identically named index entries. I guess it depends how people use the index. I personally only use it to search for the documentation of a tactic I already know the existence of.

Zimmi48:
>(Though eventually we may want to add brief overview summaries of tactics to the beginning of sections so you don't need to read/skim the entire section as often.

Something like:

```
In this section, we will describe the tactics :tacn:`apply`, :tacn:`exact` and :tacn:`assumption`.
```

Yes, that would be great to have this!

jfehrle:
> I don't know what to think of this whole business of adding things in parentheses as this is usually done only to distinguish between identically named index entries. I guess it depends how people use the index. I personally only use it to search for the documentation of a tactic I already know the existence of.

Yeah, it's not perfect.  Better to have section summaries but that may take some time to create; IMHO the section summary should be more than just a list of names for tactics--there should be a one line description of each tactic so you can get the overview without reading every tactic in the chapter and maybe grouped by function if there are many entries.  Probably shown as a table.  Annotating the index is just a quick step.

Also, if we list things like `fresh` in the Tactics and tacticals index, how would the user discover that `fresh` is a tactical (or whatever category)?

JasonGross:
For me, a "tactical" is a(n unapplied) higher order tactic.  Any tactic which takes a tactic as an argument is as tactical.

I think the following are interesting categories of tactics:
- tactics that don't modify the proof term (`change`, `constr_eq`, `move ... at ...`, `is_evar`, `revgoals `, etc)
- tactics that don't change the appearance of the goal (I.e., don't make progress) (e.g., `idtac `, `fail `, `is_evar`)
- tactics/tactic expressions that can return constrs / that get run during tactic expression evaluation phase rather than tactics that are values (e.g., let, match, lazymatch)

cpitclaudel:
This is a good direction.  Thanks! :+1:  One question: do you actually need to add `(tactical)` to the name of all tacticals?

I think I agree with Jason's definition of a tactical.  And I wouldn't put tests like `is_constr` in there, since that feels mostly like a normal tactic to me.  numgoals, swap, and the like I would call goal-manipulation primitives, and I'd be fine with stuffing them with fresh and is_constr in a large bag of "ltac builtins", or something similar.  See my comments on adding tags to tactics in https://github.com/coq/coq/issues/8946#issuecomment-491292988

jfehrle:
> do you actually need to add `(tactical)` to the name of all tacticals?

I thought that made sense because the items would be in the section "Tactic Index" but they would tend to be documented in the Ltac chapter(s) rather than the "Tactics" chapter.  The manual uses "tactical" over 100 times, seems to me it must be a useful distinction.  Precision in terminology is very important for a reference manual.

samuelgruetter:
Today I was looking for documentation of the various `is_*` tactics in the manual, and I see that this PR would provide this, so here's one user who would have found this useful, FWIW

jfehrle:
They will eventually show up in the tactics index but probably without the "(tactical)" marking.  They are:

   is_cofix @one_term
   is_const @one_term
   is_constructor @one_term
   is_evar @one_term
   is_fix @one_term
   is_ground @one_term
   is_ind @one_term
   is_proj @one_term
   is_var @one_term

jfehrle:
@Zimmi48 I think we dropped the idea of making a distinction long ago, so closing.

PR #9182, opened 2018-12-10
Stop printing Monomorphic/Polymorphic in Print.

SkySkimmer:
You can tell which it is from the `@{}` if you really care, and seeing
`Monomorphic List (A:Type)` with no indication that `Monomorphic` is
about universes can confuse people.


gares (approved):

maximedenes (approved):
This is a very good change from the didactic point of view.

maximedenes:
Will merge soon if no comments.

Issue #9175, opened 2018-12-09
Windows: VST addon does not work if CompCert is also installed


MSoegtropIMC:
The VST addon copies certain files from CompCert. If both VST and CompCert are installed, VST picks up the wrong version of the files (the one from CompCert), which leads to errors.

I discussed various possible fixes with Andrew and his preference is to ensure via path settings that VST always loads the VST version of the CompCert files. One would have to recompile VST (regenerate .vo files) if one wants to change this but one can assume that people who are picky about which version of CompCert files VST uses anyway compile VST from sources. The bundled addon is more intended for teaching and to get started with VST, where this doesn't matter (in case the CompCert files VST uses are not identical to those CompCert is using, the proof chain breaks at the interface between VST and CompCert).

We also discussed to put the common files between VST and CompCert into a separate hidden installer module, but VST does changes to CompCert files from time to time and this might then lead to release scheduling issues in case the upstream push of the changes to CompCert is not yet done or taged.

ejgallego:
cc: #8052

MSoegtropIMC:
@ejgallego : can you elaborate a bit why #8052 is related? This is not a build issue, but an installer / usability issue.

ejgallego:
Dune will provide a reasonable clean solution to this kind of vensoring problems.

MSoegtropIMC:
@ejgallego : I guess this needs more explanation. The issue is that VST copies certain files from CompCert and can either use its own copies (which might or might not be modified to enable new VST features) or the original CompCert files (possibly reducing the functionality of VST).

Now the Windows installer installs precompiled .vo files. These .vo files are either based on the VST copy of the files or the original copy of the files, and this obviously can't be changed after compiling. So the options are more or less:

1.) always use the VST version of the files (even if CompCert is installed as well)
2.) always use the CompCert version of the files (which means they must be installed with VST in case CompCert is not installed and possibly that lastest VST features cannot be used)
3.) provide two versions of the .vo files and let the user choose somehow which to use, either during installation or by modifying some settings file later.

Andrew want's to go with option 1.

This is mostly an installer and path settings (after install) problem and I don't see how dune would help here. Handling the path settings during compilation is not difficult.

ejgallego:
Thanks @MSoegtropIMC , indeed I think I had understood the problematic here, and Dune is precisely designed to enable option 1 pretty seamlessly, and it can fallback to a system-installed compcert usually in a reasonable way.

[...]

PR #9187, opened 2018-12-11
Fix issue #9175 Windows: VST addon does not work if CompCert is also installed.

MSoegtropIMC:
This fixes issue #9175.

The solution chosen to resolve the name conflicts between VST and compcert is to use the name "VST.compcert" for the VST copy of compcert files. This is done by patching all VST files (with sed) and a small patch to the VST makefile.

I will discuss with the VST team if they take this patch upstream. This is a bit problematic, because e.g. the compcert references in .v files generated by compcert clightgen must then be patched. Also this makes it harder to switch between using the VST copy or the original compcert files harder, which researchers might want to do.

For the educational and evaluation purposes for which the Windows installer is intended, I think this is a good solution.

### DIFF
--- a/dev/ci/README.md
+++ b/dev/ci/README.md
@@ -172,6 +172,11 @@ If your repository has access to runners tagged `windows`, setting the
 secret variable `WINDOWS` to `enabled` will add jobs building Windows
 versions of Coq (32bit and 64bit).
 
+If the secret variable `WINDOWS` is set to `enabled_all_addons`,
+an extended set of addons will be added to the Windows installer.
+This leads to a considerable runtime in CI so this is not enabled
+by default for pipelines for pull requests.
+
 The Windows jobs are enabled on Coq's repository, where pipelines for
 pull requests run.
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -174,7 +174,7 @@ after_script:
     - call dev/ci/gitlab.bat
   only:
     variables:
-      - $WINDOWS == "enabled"
+      - $WINDOWS =~ /enabled/
 
 build:base:
   <<: *build-template
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -691,7 +691,7 @@ function installer_addon_end {
 # ------------------------------------------------------------------------------
 
 function coq_set_timeouts_1000 {
-  find . -type f -name '*.v' -print0 | xargs -0 sed -i 's/timeout\s\+[0-9]\+/timeout 1000/'
+  find . -type f -name '*.v' -print0 | xargs -0 sed -i 's/timeout\s\+[0-9]\+/timeout 1000/g'
 }
 
 ###################### MODULE BUILD FUNCTIONS #####################
@@ -1807,8 +1807,8 @@ function install_addon_vst {
     install_glob "progs" '*.v' "$VSTDEST/progs/"
     install_glob "progs" '*.c' "$VSTDEST/progs/"
     install_glob "progs" '*.h' "$VSTDEST/progs/"
-    install_glob "veric" '*.v' "$VSTDEST/msl/"
-    install_glob "veric" '*.vo' "$VSTDEST/msl/"
+    install_glob "veric" '*.v' "$VSTDEST/veric/"
+    install_glob "veric" '*.vo' "$VSTDEST/veric/"
 
     # Install VST documentation files
     install_glob "." 'LICENSE' "$VSTDEST"
@@ -1821,12 +1821,20 @@ function install_addon_vst {
     install_glob "." '_CoqProject-export' "$VSTDEST/progs"
 }
 
+function vst_patch_compcert_refs {
+  find . -type f -name '*.v' -print0 | xargs -0 sed -E -i \
+    -e 's/(Require\s+(Import\s+|Export\s+)*)compcert\./\1VST.compcert./g' \
+    -e 's/From compcert Require/From VST.compcert Require/g'
+}
+
 function make_addon_vst {
   installer_addon_dependency vst
   if build_prep_overlay VST; then
     installer_addon_section vst "VST" "ATTENTION: SOME INCLUDED COMPCERT PARTS ARE NOT OPEN SOURCE! Verified Software Toolchain for verifying C code" "off"
-    log1 coq_set_timeouts_1000
-    log1 make IGNORECOQVERSION=true $MAKE_OPT
+    # log1 coq_set_timeouts_1000
+    log1 vst_patch_compcert_refs
+    # The usage of the shell variable ARCH in VST collides with the usage in this shellscript
+    logn make env -u ARCH make IGNORECOQVERSION=true $MAKE_OPT
     log1 install_addon_vst
     build_post
   fi
--- a/dev/build/windows/patches_coq/VST.patch
+++ b/dev/build/windows/patches_coq/VST.patch
@@ -1,0 +1,15 @@
+diff --git a/Makefile b/Makefile
+index 4a119042..fdfac13e 100755
+--- a/Makefile
++++ b/Makefile
+@@ -76,8 +76,8 @@ endif
+
+ COMPCERTDIRS=lib common $(ARCHDIRS) cfrontend flocq exportclight $(BACKEND)
+
+-COMPCERT_R_FLAGS= $(foreach d, $(COMPCERTDIRS), -R $(COMPCERT)/$(d) compcert.$(d))
+-EXTFLAGS= $(foreach d, $(COMPCERTDIRS), -Q $(COMPCERT)/$(d) compcert.$(d))
++COMPCERT_R_FLAGS= $(foreach d, $(COMPCERTDIRS), -R $(COMPCERT)/$(d) VST.compcert.$(d))
++EXTFLAGS= $(foreach d, $(COMPCERTDIRS), -Q $(COMPCERT)/$(d) VST.compcert.$(d))
+
+ # for SSReflect
+ ifdef MATHCOMP
--- a/dev/ci/ci-basic-overlay.sh
+++ b/dev/ci/ci-basic-overlay.sh
@@ -125,8 +125,8 @@
 ########################################################################
 # VST
 ########################################################################
-# Latest commit on master as of Sep 27, 2018
-: "${VST_CI_REF:=7d41c087a20bd7937def5adf6fe562182de080d5}"
+# Latest commit on master as of Dec 10, 2018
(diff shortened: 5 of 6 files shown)


MSoegtropIMC:
gitlab pipeline with all addons enabled is running here: https://gitlab.com/coq/coq/pipelines/39729443

MSoegtropIMC:
@Zimmi48 : for me the main reason for adding compcert is that VST requires compcert's clightgen unless one just want's to do exercises where Andrew provides pre-generated clightgen files. clightgen is not included with VST for licensing reasons (clightgen is not free software while VST is).

I agree that the solution is not especially nice. I will have a discussion with Andrew and Lennart on the topic. As I wrote the main user group of the installer is education and evaluation, and I think for this it is the best solution, because it is absolutely stable. Coq will always find the files regardless of the path configuration (within certain limits ...). For researchers a different solution might be better, but it would also be an option to redirect VST.compcert to compcert. As I said I need to discuss it with Andrew.

Another solution would be to have a global path config file for coqc and coqide, where one could configure this. But this still would have the effect that if someone would want to use compcert files (rather than VST), the global path redirection would redirect compcert to VST.compcert, if this is what is configured.

In the end one needs some sort of crystal ball if compcert should be "the real compcert" or "VST.compcert" if both shall be accessible as compcert, so I concluded that one usage must use a different name or one needs to have configuration files for every usage.

Zimmi48:
As far as I understand, the compcert version shipped by VST is the real compcert, just an older version. If this is correct, then installing this older version of compcert (either standalone or as a dependency of compcert) instead of rebuilding the copy that can be found in the VST repository should work as well, no?

MSoegtropIMC:
> As far as I understand, the compcert version shipped by VST is the real compcert, just an older version

No VST just ships the (small) part of compcert which is LGPL. This is the front end of CompCert to which VST interfaces. The version was sometimes newer than compcert master, because during the days Andrews group worked heavily on enabling parallelism, they did also quite some changes to CompCert.

For effectively using VST, the non free part of CompCert is required. Xavier agreed to shipping this with Coq only if it is a separate module and the license file is displayed during installation if this module is selected.

But there are other issues. Even if I remove compcert, if the modules are prefixed with "compcert" in VST, they won't be found under "lib/user-contrib/VST/compcert" without an explicit path mapping configuration, which I want to avoid, since it substantially reduces usability and increases frustration (and questions on mailing lists).

Zimmi48:
>The version was sometimes newer than compcert master, because during the days Andrews group worked heavily on enabling parallelism, they did also quite some changes to CompCert.

OK I had missed this. Is it possible to find one specific version of VST that would be compatible with one unpatched version of CompCert? Then installing VST would install the free part of CompCert in the usual place. Installing CompCert would install the full thing. And the free parts would be the same.

MSoegtropIMC:
A quick question on the pipeline: https://gitlab.com/coq/coq/pipelines/39729443.

Apparently it does not run the windows tests at all - I guess I somehow messed up the WINDOWS setting. Can I see the environment for a running pipeline somewhere, so that I can see what went wrong before I restart it?

MSoegtropIMC:
Another question: how can I get around the gitlab " Pipeline did not run on GitLab CI because branch is not up-to date" error? I think in the past squashing the commit did help. I think these are separate things so I don't really want to squash them, but I don't have a strong opinion about this.

SkySkimmer:
rebase, not squash

Gaëtan Gilbert

MSoegtropIMC:
Note on the apveyor fail: I guess it is a time out. The last lines of the downloaded log are:
```
[00:59:57] TEST      bugs/closed/1859.v 
[00:59:58] TEST      bugs/closed/5347.v 
```

MSoegtropIMC:
@SkySkimmer : the branch is up to date vs. v8.9 which it references:
```
$ git remote -v
msoegtrop       git@github.com:MSoegtropIMC/coq.git (fetch)
msoegtrop       git@github.com:MSoegtropIMC/coq.git (push)
origin  git@github.com:coq/coq.git (fetch)
origin  git@github.com:coq/coq.git (push)

$ git fetch origin                                                                                                                                                                                                                              

$ git rebase origin/v8.9
Current branch issue-9175-VST-CompCert-common-files-v8.9 is up to date.

$ git status
On branch issue-9175-VST-CompCert-common-files-v8.9
nothing to commit, working tree clean

$ git push msoegtrop
Everything up-to-date
```
Any ideas?

Zimmi48:
Yes, this is indeed a bot bug which I was working on (cf. my comment at coq/bot#25).

MSoegtropIMC:
@Zimmi48 : can you please have a look into why gitlab doesn't run? See my communication with @SkySkimmer above - he thinks it is a bot bug.

MSoegtropIMC:
@Zimmi48 : ah ok - I didn't understand this - should have followed the link.

Zimmi48:
Sorry about that. The whole point of these recent changes was to avoid this rebase business but obviously the changes had bugs. I have pushed a temporary fix pending further investigation. Your pipeline is now running.

MSoegtropIMC:
@Zimmi48 : thanks for fixing the bot. I also started a new pipeline which (hopefully) builds all addons:

https://gitlab.com/coq/coq/pipelines/39756174

MSoegtropIMC:
The pipeline https://gitlab.com/coq/coq/pipelines/39756174 which has all addons enabled also passed. I downloaded the installer and tested VST by compiling one of the pairs of AST file and verification file in the progs folder (the .vo file for the AST file must be generated first, e.g. first generate the .vo file for bin_search.v and then run through verif_bin_search.v).

silene (approved):
This is fine for the sake of getting 8.9 out of the door. But this really should be tackled by VST upsteam, as there is nothing specific to the Windows build of Coq here, if I understand correctly.

MSoegtropIMC:
> But this really should be tackled by VST upsteam, as there is nothing specific to the Windows build of Coq here, if I understand correctly.

What is special here is that the .vo files are precompiled rather than compiled by the user. This leads to certain configuration problems one does not have if everything is compiled specifically for exactly the configuration a user has. My main goal with the installer is that people taking a course, or just exploring Coq on their own, can concentrate on the course content. As is, download and installation takes 2 minutes and then the stuff works.

Of cause Andrew also has an interest in making things simple for his students, but he might not want changes which make it harder for researchers to switch VST between different CompCert versions.

I have to look deeper into this. I am not sure this can be solved really nicely without some extensions to coqc and CoqIDE path management. I think something like a global path mapping configuration file, which is read from the installation folder rather than the users home directory might be required for a better solution.

Zimmi48:
>What is special here is that the .vo files are precompiled rather than compiled by the user. This leads to certain configuration problems one does not have if everything is compiled specifically for exactly the configuration a user has.

I'm not sure I get this. There probably is an issue as well for people installing VST + CompCert using opam for instance.

>I have to look deeper into this. I am not sure this can be solved really nicely without some extensions to coqc and CoqIDE path management. I think something like a global path mapping configuration file, which is read from the installation folder rather than the users home directory might be required for a better solution.

I think the only way to fix this would be on the contrary to impose best release management practices to the VST and CompCert developers. But never mind, this can be fixed later and independently.

@MSoegtropIMC Is there a PR for doing the same thing to the master branch? There should probably be one.

MSoegtropIMC:
> Is there a PR for doing the same thing to the master branch? There should probably be one.

Yes, I already created #9192.

> I'm not sure I get this. There probably is an issue as well for people installing VST + CompCert using opam for instance.

With Opam you have some of the problems, but not all of them. The problem that precompiled .vo files cannot be adjusted to different environments you only have if you have procompiled .vo files, which opam doesn't. With opam you configure VST in a specific way and compile the .vo files for this configuration. If you change the configuration I guess opam would recreate the .vo files. A Windows installer obviously can't do that. Please note that on Linux you have the same issue if you ever do a make install of a library.

The more general problem that it makes sense to use VST with either its own version of the CompCert frontend or the CompCert version exists of cause also in opam. I don't think this is a problem of release management. If Andrew's group develops a new feature in VST which is not supported in CompCert and does some experimental changes to CompCert which might or might not be the right way, it does make sense to do these changes at first in a VST local copy of the CompCert front end to give people a chance to experiment with the new high level features before it is finally sorted out how the required functionality is actually implemented in CompCert. In the past afaik this e.g. happened in the area of multithreading which was originally not supported at all by CompCert. Initially it was no way clear how to do this (I would say it took 2 years of research to figure this out), and I think it was not even clear if it could be done in a reasonable time and it was also not clear if such changes ever make it into the CompCert mainline.

Obviously using VST with such an experimental CompCert front end breaks the proof chain at the VST - CompCert interface and other people might not want this and use the CompCert files instead. This is the reason why Andrew decided he wants to have this switchable. On Windows one has to make this decision upfront cause of the precompiled .vo files. We decided that for students/teaching/learning/evaluation latest features are more important than the proof chain.

In conclusion I think that coqc and CoqIDE need some good way to say that VST.compcert either references lib/user-contrib/compcert or lib/user-contrib/VST/compcert, and since this is preinstalled library stuff this setting must be global. On Windows one would leave this mapping fixed because changing it would require to recompile the .vo files (with Dune we might support this one day on Windows). With opam one is able to change this configuration and recompile the .vo files.

Currently VST references compcert as compcert and I don't see any way to have a global configuration such that compcert sees lib/user-contrib/compcert when refernecing compcert while VST gets lib/user-contrib/VST/compcert when referencing compcert, so I concluded that VST needs to reference compcert under a private name (VST.compcert) which can then be switched around by configuration.

I thought about if this can be solved with local configuration files (in the VST and compcert folders) but can't think of a robust and elegant solution. Remember that we are talking about the configuration of pre installed libraries and not of user projects.

I am in discussion about this  with Andrew, but I hope you agree now that this is not trivial and can't be solved by "just doing it properly".

dev/ci/README.md
```diff
@@ -172,6 +172,11 @@ If your repository has access to runners tagged `windows`, setting the
 secret variable `WINDOWS` to `enabled` will add jobs building Windows
 versions of Coq (32bit and 64bit).
 
+If the secret variable `WINDOWS` is set to `enabled_all_addons`,
+an extended set of addons will be added to the Windows installer.
+This leads to a considerable runtime in CI so this is not enabled
+by default for pipelines for pull requests.
```
Zimmi48: In v8.9 this is actually not true: the extended set of addons is built in all cases.
See: https://github.com/coq/coq/blob/v8.9/dev/ci/gitlab.bat

MSoegtropIMC:
> In v8.9 this is actually not true: the extended set of addons is built in all cases.

Ah sorry, this slipped trough. I will prepare a PR soon.

Zimmi48:
>Ah sorry, this slipped trough. I will prepare a PR soon.

There is no issue leaving things as they are now though.

PR #9193, opened 2018-12-11
Tests for #4509, #6202 which happen to be fixed (was a lost of evars in shelf)

herbelin:
**Kind:** Tests

This mini-PR is a counterpart to closing #4509 #6202 which happen to be fixed. We add the corresponding regression tests to the test-suite.


SkySkimmer:
Why is this `needs: test-suite update`?

herbelin:
> Why is this `needs: test-suite update`?

I made a confusion. I did not see the `needs` and thought it was describing a type of PR. I'm sorry. I removed the label (no label then).

SkySkimmer (approved):
I'll take this.

PR #9085, opened 2018-11-26
Makefile: fix build on macOS

liyishuai:
**Kind:** bug fix

Fixes / closes #9084 

- [ ] Entry added in CHANGES.md.

PR #9194, opened 2018-12-11
Fixing uses of sed in #8985 which do not work on MacOS X

herbelin:
**Kind:** compilation fix

I don't know if this is the best fix we can imagine but it works for me.

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -572,11 +572,11 @@ module Gramext = Gramlib__Gramext \
 module Grammar = Gramlib__Grammar" > $@
 
 gramlib/.pack/gramlib__P%: gramlib/p% | gramlib/.pack
-	cp -a $< $@
-	sed -e "1i # 1 \"$<\"" -i $@
+	printf '# 1 "%s"\n' $< > $@
+	cat $< >> $@
 gramlib/.pack/gramlib__G%: gramlib/g% | gramlib/.pack
-	cp -a $< $@
-	sed -e "1i # 1 \"$<\"" -i $@
+	printf '# 1 "%s"\n' $< > $@
+	cat $< >> $@
 
 # Specific rules for gramlib to pack it Dune / OCaml 4.08 style
 GRAMOBJS=$(addsuffix .cmo, $(GRAMFILES))


ejgallego:
Note we already have https://github.com/coq/coq/pull/9085 which for some reason it has not been merged.

herbelin:
Sorry, I did not know about #9085.

Unfortunately, I tested #9085 and it does not work for me. It complains with `command i expects \ followed by text`.

ejgallego:
What OSX version are you using @herbelin ?

ejgallego:
> What OSX version are you using @herbelin ?

AFAICT the sed invocation used there should work.

herbelin:
I'm using 10.11 (El Capitan). I don't find a version number for `sed` though.

If I just do `sed -e "1i foo"` in a shell, it fails the same (`sed: 1: "1i foo
": command i expects \ followed by text`).

ejgallego:
Wow, that's quite strange; I am not `sed` OSX expert so I guess I'll let an expert handle this.

herbelin:
Without this PR I have to patch my archive for compiling. Can the first person who sees no problem with this PR merge it? This works for me. This works on CI (hence on linux I guess). Appveyor failure seems due to a timeout.

@mattam82 did not report anything and I'd be curious of why he does not meet the POSIX `sed` problem, but in any case, this is, formally speaking, "blocking" for me (and presumably for everyone working in a similar environement as mine).

mattam82:
In my setup the sed fails but the compilation goes through. `sed` is not portable, for sure BSD sed and gnu sed have a different handling of -i.

herbelin:
> In my setup the sed fails but the compilation goes through. 

For me, it goes through at the second time if I restart `make`. The first call to `make` fails (on a fresh archive). Is that the same for you?

mattam82:
Nope, I did a first `git clean -dfx` and the compilation succeeded at first, using the standard make (no dune)

herbelin:
With a `git clean -fxd` it seems indeed not to make the `make` failing (previously, I was doing `rm -r gramlib/.pack` and `make` was failing` afterwards).

In any case, the `sed` is failing and the `#` line is not added as expected. Is that what you observe too?

mattam82:
Your version works for me. I'd like to merge but I'd like to get an ack from @ejgallego, @SkySkimmer  or @gares , it seems this issue was already discussed at length.

mattam82 (approved):

Makefile.build
```diff
@@ -572,11 +572,11 @@ module Gramext = Gramlib__Gramext \
 module Grammar = Gramlib__Grammar" > $@
 
 gramlib/.pack/gramlib__P%: gramlib/p% | gramlib/.pack
-	cp -a $< $@
-	sed -e "1i # 1 \"$<\"" -i $@
+	printf "# 1 \"%s\"\n" $< > $@
```
SkySkimmer:
```suggestion
	printf '# 1 "%s"\n' $< > $@
```

Makefile.build
```diff
@@ -572,11 +572,11 @@ module Gramext = Gramlib__Gramext \
...
+	cat $< >> $@
 gramlib/.pack/gramlib__G%: gramlib/g% | gramlib/.pack
-	cp -a $< $@
-	sed -e "1i # 1 \"$<\"" -i $@
+	printf "# 1 \"%s\"\n" $< > $@
```
SkySkimmer:
```suggestion
	printf '# 1 "%s"\n' $< > $@
```

gares:
I'd recommend using `echo` (or `printf` as Gaetan suggests)

mattam82:
@herbelin if you fix the quoting as suggested by Gaëtan I can merge quickly.

herbelin:
Now using single quotes.

mattam82:
Perfect, thanks!

PR #9197, opened 2018-12-11
Fixing imports in debug printers: gramlib depends on Loc which is in lib.cma

herbelin:
**Kind:** debugger fix

At some time, `gramlib` started to depend on `Loc` but the order of imports in `core.dbg` was not updated.


ejgallego (approved):
Good catch, thanks @herbelin ; hopefully one day we can test this.

PR #9206, opened 2018-12-12
[stm] join the tip of the document even when fixing a proof (fix #9204)

### LINKED ISSUE #9204: [stm] Error resilience behaves strangely with `Stm.join`
Opening the following file https://github.com/ejgallego/coq-serapi/blob/b62ae5fc928fa2f098df1e2308d6e1244521d99e/tests/fail/assoc.v
in CoqIDE, then go to the end.

- The first time, it emits no error [exception], but a message in the buffer;
- the second time, click the gears, and CoqIDE tries to close the proof, an error "no such goal" is produced,
- and the third time, click the gears again, the CoqIDE says "All proof terms checked by the kernel", the error persists.

The problem is that `Stm.join` / `finish` returns successfully even if the bad proof is there; this seems a bad interaction with the error recovery mode. Disabling it fixes the problem.

Another variation of the bug is with `-async-proofs off` in CoqIDE, in this case clicking the gear will always report "All proof terms checked by the kernel".

gares:


ejgallego (approved):
I guess the fix is OK for CoqIDE and should likely be backported to 8.9.0

However I am not sure if the semantics we have now are the intended ones. Adding the example in the bug, then doing `join` still results in no error.

gares:
I'm working on a way to test the second commit, so please don't merge now

gares:
done

stm/stm.ml
```diff
@@ -2729,12 +2729,29 @@ let rec join_admitted_proofs id =
        join_admitted_proofs view.next
   | _ -> join_admitted_proofs view.next
 
+(* Error resiliency may have tolerated some broken commands *)
+let rec check_no_err_states ~doc visited id =
+  let open Stateid in
+  if Set.mem id visited then visited else
+  match state_of_id ~doc id with
+  | `Error e -> raise e
```
gares: @maximedenes heads up since IIRC in your PR about the cache we had a discussion about the representation of erroneous states and here I call the API you have changed.

ejgallego (changes requested):
After thinking a bit more about this I'd like to understand better what is going here.

I am a bit concerned about the situation, so it seems that we are calling the kernel and the kernel is telling us "everything is all right" despite it is not.

So in order to avoid this problem, we check if the document manager has seen some error. This seems a reason to be concerned, isn't it?

gares:
No, the broken lemma is not added to the env.
Try to print it and you will get an error.

It would be the same if it was a broken `Definition` (nothing to do with async).
It is just skipped.

ejgallego:
> No, the broken lemma is not added to the env.

Oh I see; so that seems like an interesting user interface issue. IMHO the current interface seems a bit misleading, as you would think Coq knows of the definition. If add a `Print` below things behave a bit weird, the thing feels a bit unstable with that combination of `EditAt` + resilience.

I was assuming that the error recovery strategy on proofs would be to add an empty proof, but indeed that strategy should be valid too. However the non-added lemma maybe should be highlighted then.

So indeed spans maybe should get an extra state for "kernel-checked", but we need to think more.

I guess we can then merge this solution for now.

ppedrot (commented):
Even though this has been assigned to me because it is a change in the IDE, this PR is actually about the STM, hence I am incompetent. I'll assign it to me though and because @ejgallego seems OK, I'll merge it soon. (@ejgallego can you officially turn your comment into an approval?)

ppedrot:
Oops, just realized it was assigned to @ejgallego already. OK, forget about my assignment then.

ejgallego (approved):
I am merging this, as this is a net improvement; but I will keep the original issue opened as there still remain some interaction issues.

gares:
thanks!

PR #9207, opened 2018-12-12
Addressing #9180 and other fine-tunings of notation printing, including proper support for generic notations of applications

### LINKED ISSUE #9180: Coq does not keep some notations
#### Version

8.10+alpha (cb287591722e4d44671eb8c90138cf063ac5b707)

#### Operating system

- macOS 10.14
- Debian 4.9.110-3+deb9u3 (2018-08-19) x86_64 

Here is some code from Mathcomp's `ssrnat`, illustrating the issue.
```coq
Notation succn := (Datatypes.S).

Notation "n .+1" := (succn n) (at level 2, left associativity,
  format "n .+1") : nat_scope.

Locate ".+1".
(* Notation *)
(* "n .+1" := S n : nat_scope (default interpretation) *)
(** so Coq does not apply succn notation *)

Goal forall x : nat, x.+1 = x.+1.
(* forall x : nat, succn x = succn x *)
(** here Coq should keep ".+1" notations *)
```

Reported by @mr-ohman.

herbelin:
**Kind:** misc bug fix / fine-tuning of #8187

Fixes / closes #9180

This PR implements "fine-tuning" changes to notation printing in relation with the discussion at #9180:
- on applicative nodes, be consistent with the subterm order (applicative nodes are treated specially to accomodate implicit arguments and there was no effort to browse notations in subterm order); this has two consequences:
  - in e.g. `S n`, if there is a notation both for `S n` as a whole and for `S` isolated, the one for `S n` is used first (this addresses #9180)
  - in case of generic notations for an	applicative node (i.e. a notation with a variable at the head of the application), the notation does not	apply any more only to the head of the full (flattened) applicative node; it instead matches the application stepwise as if a nested application (this addresses this [comment](https://github.com/coq/coq/issues/9180#issuecomment-446371111) at #9180)
- in situations like `f n l` where a notation bound to `@f` exists, and no notation capturing a larger subterm exists, use the notation (formerly, this was the case only if `f` had no arguments); the overlook was apparently involontary; a similar uniformization of behavior for notations is done when in position of pattern.

I suspect that this shall be considered as strict improvements. In addition to @anton-trunov and @ggonthier already involved in #9180, I'm informing also other users using notations heavily @beta-ziliani, @JasonGross, @ralfjung, @robbertkrebbers, in case they see a problem with these changes.

I suspect that I should include the above level of precision in the current documentation of notation printing rules (??).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,6 +49,9 @@ Notations
 - New command `String Notation` to register string syntax for custom
   inductive types.
 
+- Proper support for notations of generic applications (i.e. with
+  flexible head, as in "Notation app f x := (f x).")
+
 Plugins
 
 - The quote plugin (https://coq.inria.fr/distrib/V8.8.1/refman/proof-engine/detailed-tactic-examples.html#quote)
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -464,7 +464,11 @@ and apply_notation_to_pattern ?loc gr ((subst,substlist),(nb_to_drop,more_args))
 		substlist in
 	    let l2 = List.map (extern_cases_pattern_in_scope allscopes vars) more_args in
             let l2' = if Constrintern.get_asymmetric_patterns () || not (List.is_empty ll) then l2
-	      else
+              else if nb_to_drop = Some 0 then
+                (* indicates that notation is bound to some [@ref] *)
+                l2
+              else
+                let nb_to_drop = match nb_to_drop with Some nb_to_drop -> nb_to_drop | None -> 0 in
 		match drop_implicits_in_patt gr nb_to_drop l2 with
 		  |Some true_args -> true_args
 		  |None -> raise No_match
@@ -485,7 +489,11 @@ and apply_notation_to_pattern ?loc gr ((subst,substlist),(nb_to_drop,more_args))
           subst in
       let l2 = List.map (extern_cases_pattern_in_scope allscopes vars) more_args in
       let l2' = if Constrintern.get_asymmetric_patterns () then l2
-	else
+        else if nb_to_drop = Some 0 then
+          (* indicates that notation is bound to some [@ref] *)
+          (assert (l1=[]); l2)
+        else
+          let nb_to_drop = match nb_to_drop with Some nb_to_drop -> nb_to_drop | None -> 0 in
 	  match drop_implicits_in_patt gr (nb_to_drop + List.length l1) l2 with
 	    |Some true_args -> true_args
 	    |None -> raise No_match
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -29,6 +29,7 @@ open Pattern
 open Notation
 open Detyping
 open Decl_kinds
+open Notation_term
 
 module NamedDecl = Context.Named.Declaration
 (*i*)
@@ -464,11 +465,9 @@ and apply_notation_to_pattern ?loc gr ((subst,substlist),(nb_to_drop,more_args))
 		substlist in
 	    let l2 = List.map (extern_cases_pattern_in_scope allscopes vars) more_args in
             let l2' = if Constrintern.get_asymmetric_patterns () || not (List.is_empty ll) then l2
-              else if nb_to_drop = Some 0 then
-                (* indicates that notation is bound to some [@ref] *)
-                l2
-              else
-                let nb_to_drop = match nb_to_drop with Some nb_to_drop -> nb_to_drop | None -> 0 in
+              else match nb_to_drop with
+              | RefDeactivatingImpls | NotAppliedRef -> l2
+              | NAryApplication nb_to_drop ->
 		match drop_implicits_in_patt gr nb_to_drop l2 with
 		  |Some true_args -> true_args
 		  |None -> raise No_match
@@ -489,11 +488,9 @@ and apply_notation_to_pattern ?loc gr ((subst,substlist),(nb_to_drop,more_args))
           subst in
       let l2 = List.map (extern_cases_pattern_in_scope allscopes vars) more_args in
       let l2' = if Constrintern.get_asymmetric_patterns () then l2
-        else if nb_to_drop = Some 0 then
-          (* indicates that notation is bound to some [@ref] *)
-          (assert (l1=[]); l2)
-        else
-          let nb_to_drop = match nb_to_drop with Some nb_to_drop -> nb_to_drop | None -> 0 in
+        else match nb_to_drop with
+        | RefDeactivatingImpls | NotAppliedRef -> l2
+        | NAryApplication nb_to_drop ->
 	  match drop_implicits_in_patt gr (nb_to_drop + List.length l1) l2 with
 	    |Some true_args -> true_args
 	    |None -> raise No_match
@@ -1077,12 +1074,9 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
         if is_inactive_rule keyrule then raise No_match;
 	(* Adjusts to the number of arguments expected by the notation *)
 	let (t,args,argsscopes,argsimpls) = match DAst.get t ,n with
-          (* Convention: [Some 0] means a notation of the form @ref *)
-          (* This deactivates the use of implicit arguments or scopes *)
-          | GRef (ref,us), Some 0 -> DAst.make @@ GApp (t,[]), [], [], []
-          | GApp (f,args), Some 0 -> DAst.make @@ GApp (f,[]), args, [], []
-          (* This deactivates the use of implicit arguments or scopes *)
-	  | GApp (f,args), Some n
+          | GRef (ref,us), RefDeactivatingImpls -> DAst.make @@ GApp (t,[]), [], [], []
+          | GApp (f,args), RefDeactivatingImpls -> DAst.make @@ GApp (f,[]), args, [], []
+          | GApp (f,args), NAryApplication n
 	      when List.length args >= n ->
                 (match DAst.get f with
                 | GRef (ref,us) ->
@@ -1098,23 +1092,13 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
 		      select_impargs_size
 		        (List.length args) (implicits_of_global ref) in
 		    try List.skipn n impls with Failure _ -> [] in
-                  DAst.make @@ GApp (f,args1), args2, subscopes,impls
+                  (if n = 0 then f else DAst.make @@ GApp (f,args1)), args2, subscopes,impls
                 | _ ->
                   (* Case of a notation with no ref at head *)
                   (* We align on the size of arguments of the notation *)
                   t, [], [], [])
-	  | GApp (f, args), None ->
-            begin match DAst.get f with
-            | GRef (ref,us) ->
-              (* The notation is a reference w/o @: we use impl/scope info *)
-	      let subscopes = find_arguments_scope ref in
-	      let impls =
-		  select_impargs_size
-		    (List.length args) (implicits_of_global ref) in
-	      f, args, subscopes, impls
-            | _ -> t, [], [], []
-            end
-          | _, None -> t, [], [], []
+          | GRef _, NAryApplication 0 -> t, [], [], []
+          | _, NotAppliedRef -> t, [], [], []
           | _ -> raise No_match in
 	(* Try matching ... *)
         let terms,termlists,binders,binderlists =
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -1069,6 +1069,11 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
         if is_inactive_rule keyrule then raise No_match;
 	(* Adjusts to the number of arguments expected by the notation *)
 	let (t,args,argsscopes,argsimpls) = match DAst.get t ,n with
+          (* Convention: [Some 0] means a notation of the form @ref *)
+          (* This deactivates the use of implicit arguments or scopes *)
+          | GRef (ref,us), Some 0 -> DAst.make @@ GApp (t,[]), [], [], []
+          | GApp (f,args), Some 0 -> DAst.make @@ GApp (f,[]), args, [], []
+          (* This deactivates the use of implicit arguments or scopes *)
 	  | GApp (f,args), Some n
 	      when List.length args >= n ->
 	      let args1, args2 = List.chop n args in
@@ -1086,11 +1091,11 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
                   subscopes,impls
                 | _ ->
                   [], [] in
-	      (if Int.equal n 0 then f else DAst.make @@ GApp (f,args1)),
-	      args2, subscopes, impls
+              DAst.make @@ GApp (f,args1), args2, subscopes, impls
 	  | GApp (f, args), None ->
             begin match DAst.get f with
             | GRef (ref,us) ->
+              (* The notation is a reference w/o @: we use impl/scope info *)
 	      let subscopes = find_arguments_scope ref in
 	      let impls =
 		  select_impargs_size
@@ -1098,7 +1103,6 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
 	      f, args, subscopes, impls
             | _ -> t, [], [], []
             end
-	  | GRef (ref,us), Some 0 -> DAst.make @@ GApp (t,[]), [], [], []
           | _, None -> t, [], [], []
           | _ -> raise No_match in
 	(* Try matching ... *)
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -261,11 +261,6 @@ let rec insert_pat_coercion ?loc l c = match l with
   | [] -> c
   | ntn::l -> CAst.make ?loc @@ CPatNotation (ntn,([insert_pat_coercion ?loc l c],[]),[])
 
-let add_lonely keyrule seen =
-  match keyrule with
-  | NotationRule (None,ntn) -> ntn::seen
-  | SynDefRule _ | NotationRule (Some _,_) -> seen
-
 (**********************************************************************)
 (* conversion of references                                           *)
 
@@ -390,7 +385,7 @@ let rec extern_cases_pattern_in_scope (custom,scopes as allscopes) vars pat =
   with No_match ->
     try
       if !Flags.in_debugger || !Flags.raw_print || !print_no_symbol then raise No_match;
-      extern_notation_pattern allscopes [] vars pat
+      extern_notation_pattern allscopes vars pat
         (uninterp_cases_pattern_notations scopes pat)
     with No_match ->
     let loc = pat.CAst.loc in
@@ -444,14 +439,14 @@ let rec extern_cases_pattern_in_scope (custom,scopes as allscopes) vars pat =
       insert_pat_coercion coercion pat
 
 and apply_notation_to_pattern ?loc gr ((subst,substlist),(nb_to_drop,more_args))
-    (custom, (tmp_scope, scopes) as allscopes) lonely_seen vars =
+    (custom, (tmp_scope, scopes) as allscopes) vars =
   function
     | NotationRule (sc,ntn),key,need_delim ->
       begin
         match availability_of_entry_coercion custom (fst ntn) with
         | None -> raise No_match
         | Some coercion ->
-            let key = if need_delim || List.mem ntn lonely_seen then key else None in
+            let key = if need_delim then key else None in
             let scopt = match key with Some _ -> sc | _ -> None in
 	    let scopes' = Option.List.cons scopt scopes in
 	    let l =
@@ -497,7 +492,7 @@ and apply_notation_to_pattern ?loc gr ((subst,substlist),(nb_to_drop,more_args))
       in
       assert (List.is_empty substlist);
       insert_pat_coercion ?loc coercion (mkPat ?loc qid (List.rev_append l1 l2'))
-and extern_notation_pattern allscopes lonely_seen vars t = function
+and extern_notation_pattern allscopes vars t = function
   | [] -> raise No_match
   | (keyrule,pat,n as _rule,key,need_delim)::rules ->
     try
@@ -507,27 +502,25 @@ and extern_notation_pattern allscopes lonely_seen vars t = function
         | PatCstr (cstr,args,na) ->
           let t = if na = Anonymous then t else DAst.make ?loc (PatCstr (cstr,args,Anonymous)) in
 	  let p = apply_notation_to_pattern ?loc (ConstructRef cstr)
-            (match_notation_constr_cases_pattern t pat) allscopes lonely_seen vars
+            (match_notation_constr_cases_pattern t pat) allscopes vars
             (keyrule,key,need_delim) in
 	  insert_pat_alias ?loc p na
 	| PatVar Anonymous -> CAst.make ?loc @@ CPatAtom None
         | PatVar (Name id) -> CAst.make ?loc @@ CPatAtom (Some (qualid_of_ident ?loc id))
     with
       No_match ->
-      let lonely_seen = add_lonely keyrule lonely_seen in
-      extern_notation_pattern allscopes lonely_seen vars t rules
+      extern_notation_pattern allscopes vars t rules
 
-let rec extern_notation_ind_pattern allscopes lonely_seen vars ind args = function
+let rec extern_notation_ind_pattern allscopes vars ind args = function
   | [] -> raise No_match
   | (keyrule,pat,n as _rule,key,need_delim)::rules ->
     try
       if is_inactive_rule keyrule then raise No_match;
       apply_notation_to_pattern (IndRef ind)
-        (match_notation_constr_ind_pattern ind args pat) allscopes lonely_seen vars (keyrule,key,need_delim)
+        (match_notation_constr_ind_pattern ind args pat) allscopes vars (keyrule,key,need_delim)
     with
       No_match ->
-      let lonely_seen = add_lonely keyrule lonely_seen in
-      extern_notation_ind_pattern allscopes lonely_seen vars ind args rules
+      extern_notation_ind_pattern allscopes vars ind args rules
 
 let extern_ind_pattern_in_scope (custom,scopes as allscopes) vars ind args =
   (* pboutill: There are letins in pat which is incompatible with notations and
@@ -539,8 +532,8 @@ let extern_ind_pattern_in_scope (custom,scopes as allscopes) vars ind args =
   else
     try
       if !Flags.raw_print || !print_no_symbol then raise No_match;
-      extern_notation_ind_pattern allscopes [] vars ind args
-          (uninterp_ind_pattern_notations scopes ind)
+      extern_notation_ind_pattern allscopes vars ind args
+          (uninterp_ind_pattern_notations scopes ind args)
     with No_match ->
       let c = extern_reference vars (IndRef ind) in
       let args = List.map (extern_cases_pattern_in_scope allscopes vars) args in
@@ -780,7 +773,7 @@ let rec extern inctx (custom,scopes as allscopes) vars r =
     let r'' = flatten_application r' in
     if !Flags.raw_print || !print_no_symbol then raise No_match;
     extern_optimal
-      (fun r -> extern_notation allscopes [] vars r (uninterp_notations scopes r))
+      (fun r -> extern_notation allscopes vars r (uninterp_notations scopes r))
       r r''
   with No_match ->
   let loc = r'.CAst.loc in
@@ -1066,7 +1059,7 @@ and extern_eqn inctx scopes vars {CAst.loc;v=(ids,pll,c)} =
   let pll = List.map (List.map (extern_cases_pattern_in_scope scopes vars)) pll in
   make ?loc (pll,extern inctx scopes vars c)
 
-and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
+and extern_notation (custom,scopes as allscopes) vars t = function
   | [] -> raise No_match
   | (keyrule,pat,n as _rule,key,need_delim)::rules ->
       let loc = Glob_ops.loc_of_glob_constr t in
@@ -1076,14 +1069,14 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
 	let (t,args,argsscopes,argsimpls) = match DAst.get t ,n with
           | GRef (ref,us), RefDeactivatingImpls -> DAst.make @@ GApp (t,[]), [], [], []
           | GApp (f,args), RefDeactivatingImpls -> DAst.make @@ GApp (f,[]), args, [], []
-          | GApp (f,args), NAryApplication n
-	      when List.length args >= n ->
+          | GApp (f,args), NAryApplication n ->
+                assert (List.length args >= n);
+                let args1, args2 = List.chop n args in
                 (match DAst.get f with
                 | GRef (ref,us) ->
                   (* Case of a notation with a ref at head *)
                   (* We align on the ref and apply scope/impl *)
                   (* info to extra arguments *)
-                  let args1, args2 = List.chop n args in
 	          let subscopes =
 		    try List.skipn n (find_arguments_scope ref)
                     with Failure _ -> [] in
@@ -1096,7 +1089,7 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
                 | _ ->
                   (* Case of a notation with no ref at head *)
                   (* We align on the size of arguments of the notation *)
-                  t, [], [], [])
+                  (if n = 0 then f else DAst.make @@ GApp (f,args1)), args2, [], [])
           | GRef _, NAryApplication 0 -> t, [], [], []
           | _, NotAppliedRef -> t, [], [], []
           | _ -> raise No_match in
@@ -1110,7 +1103,7 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
              (match availability_of_entry_coercion custom (fst ntn) with
              | None -> raise No_match
              | Some coercion ->
-                  let key = if need_delim || List.mem ntn lonely_seen then key else None in
+                  let key = if need_delim then key else None in
                   let scopt = match key with Some _ -> sc | None -> None in
                   let scopes' = Option.List.cons scopt (snd scopes) in
 	          let l =
@@ -1148,8 +1141,7 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
 	  CAst.make ?loc @@ explicitize false argsimpls (None,e) args
       with
         No_match ->
-        let lonely_seen = add_lonely keyrule lonely_seen in
-        extern_notation allscopes lonely_seen vars t rules
+        extern_notation allscopes vars t rules
 
 and extern_recursion_order scopes vars = function
     GStructRec -> CStructRec
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -1084,10 +1084,12 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
           (* This deactivates the use of implicit arguments or scopes *)
 	  | GApp (f,args), Some n
 	      when List.length args >= n ->
-	      let args1, args2 = List.chop n args in
-              let subscopes, impls =
-                match DAst.get f with
+                (match DAst.get f with
                 | GRef (ref,us) ->
+                  (* Case of a notation with a ref at head *)
+                  (* We align on the ref and apply scope/impl *)
+                  (* info to extra arguments *)
+                  let args1, args2 = List.chop n args in
 	          let subscopes =
 		    try List.skipn n (find_arguments_scope ref)
                     with Failure _ -> [] in
@@ -1096,10 +1098,11 @@ and extern_notation (custom,scopes as allscopes) lonely_seen vars t = function
 		      select_impargs_size
 		        (List.length args) (implicits_of_global ref) in
 		    try List.skipn n impls with Failure _ -> [] in
-                  subscopes,impls
+                  DAst.make @@ GApp (f,args1), args2, subscopes,impls
                 | _ ->
-                  [], [] in
-              DAst.make @@ GApp (f,args1), args2, subscopes, impls
+                  (* Case of a notation with no ref at head *)
+                  (* We align on the size of arguments of the notation *)
+                  t, [], [], [])
 	  | GApp (f, args), None ->
             begin match DAst.get f with
             | GRef (ref,us) ->
--- a/interp/notation.ml
+++ b/interp/notation.ml
@@ -1260,13 +1260,31 @@ let interp_notation ?loc ntn local_scopes =
     user_err ?loc 
     (str "Unknown interpretation for notation " ++ pr_notation ntn ++ str ".")
 
+let less_args n n' = match n,n' with
+  | Some _, None -> true
+  | Some n, Some n' -> n < n'
+  | _, _ -> false
+
+let rec insert_by_number_of_arguments ((_,_,n),_,_ as fullrule) = function
+  | [] -> [fullrule]
+  | ((_,_,n'),_,_ as x) :: rest as all ->
+     if less_args n n' then x :: insert_by_number_of_arguments fullrule rest
+     else fullrule :: all
+
+let rec append_by_number_of_arguments l l' =
+  match l with
+  | [] -> l'
+  | x :: l -> insert_by_number_of_arguments x (append_by_number_of_arguments l l')
+
 let extract_notations (istype,scopes) keys =
   if keys == [] then [] (* shortcut *) else
   let scope_map, global_map = !notations_key_table in
   let rec aux scopes seen =
   match scopes with
-  | UninterpScope sc :: scopes -> keymap_extract istype keys sc scope_map @ aux scopes (String.Set.add sc seen)
-  | UninterpSingle rule :: scopes -> (rule,None,false) :: aux scopes seen
+  | UninterpScope sc :: scopes ->
+      append_by_number_of_arguments (keymap_extract istype keys sc scope_map) (aux scopes (String.Set.add sc seen))
+  | UninterpSingle rule :: scopes ->
+      insert_by_number_of_arguments (rule,None,false) (aux scopes seen)
   | [] ->
       let find key = try KeyMap.find key global_map with Not_found -> [] in
       keymap_extract_remainder istype seen (List.flatten (List.map find keys))
--- a/interp/notation.ml
+++ b/interp/notation.ml
@@ -184,7 +184,12 @@ type interp_rule =
   | NotationRule of scope_name option * notation
   | SynDefRule of KerName.t
 
-type scoped_notation_rule_core = scope_name * notation * interpretation * notation_applicative_structure
+type key_rigidity =
+  | RigidHeadKey of notation_applicative_structure
+  | FlexHeadKey of int
+
+type scoped_notation_rule_core = scope_name * notation * interpretation * key_rigidity
+type notation_rule_core_with_rigidity = interp_rule * interpretation * key_rigidity
 type notation_rule_core = interp_rule * interpretation * notation_applicative_structure
 type notation_rule = notation_rule_core * delimiters option * bool
 
@@ -203,7 +208,7 @@ module InterpRuleSet = Set.Make(struct
 
 type uninterp_scope_elem =
   | UninterpScope of scope_name
-  | UninterpSingle of notation_rule_core
+  | UninterpSingle of notation_rule_core_with_rigidity
 
 let uninterp_scope_eq_weak s1 s2 = match s1, s2 with
 | UninterpScope s1, UninterpScope s2 -> String.equal s1 s2
@@ -394,7 +399,39 @@ let keymap_add key sc interp (scope_map,global_map) =
   | (NotationRule (None,_) | SynDefRule _), _, _ -> global_map in
   (scope_map, global_map)
 
-let keymap_extract istype keys sc map =
+(* We order the notations by number of arguments: for each arity come
+     first those with rigid arity, then those with a flexible arity
+     instantiable to the given arity; RefDeactivatingImpls] is 0-ary *)
+
+let arity_rule = function
+  | NAryApplication n -> n
+  | RefDeactivatingImpls | NotAppliedRef -> 0
+
+let add_depending_on_rigidity (rule,interp,rigidity) delim need_delim n l =
+  List.map_i (fun i (rigid,flex as x) ->
+    match rigidity with
+    | RigidHeadKey c ->
+      if i = arity_rule c then
+        let rule = (rule, interp, c) in
+        ((rule,delim,need_delim)::rigid,flex)
+      else x
+    | FlexHeadKey n' ->
+      if i >= n' then
+        let rule = (rule, interp, NAryApplication i) in
+        (rigid,(rule,delim,need_delim)::flex)
+      else x) 0 l
+
+let add_lonely (keyrule,_,_) seen =
+  match keyrule with
+  | NotationRule (None,ntn) -> ntn::seen
+  | SynDefRule _ | NotationRule (Some _,_) -> seen
+
+let is_seen (rule,_,_) seen =
+  match rule with
+  | NotationRule (sc,ntn) -> List.mem ntn seen
+  | _ -> false
+
+let keymap_extract_and_add n istype keys sc lonely_seen map rest =
   let keymap =
     try ScopeMap.find (Some sc) map
     with Not_found -> KeyMap.empty in
@@ -405,8 +442,16 @@ let keymap_extract istype keys sc map =
     else
       (* Pass the delimiter so that it can be used if ever the notation is masked *)
       (String.Map.find sc !scope_map).delimiters in
-  let add_scope rule = (rule,delim,false) in
-  List.map_append (fun key -> try List.map add_scope (KeyMap.find key keymap) with Not_found -> []) keys
+  let l = List.map_append (fun key -> try KeyMap.find key keymap with Not_found -> []) keys in
+  List.fold_right (fun rule ->
+      (* There is a case when we need to enforce a delimiter, it is
+         when there is a lonely notation which would be used prioritary
+         at parsing time and hide the given interpretation, as in:
+           Notation "# x" := (S x) (at level 20) : nat_scope.
+           Notation "# x" := (Some x).
+           Check fun x => (# x)%nat. *)
+      let needs_delim = is_seen rule lonely_seen in
+      add_depending_on_rigidity rule delim needs_delim n) l rest
 
 let find_with_delimiters istype = function
   | None ->
@@ -420,21 +465,22 @@ let find_with_delimiters istype = function
         | Some key -> Some (Some key)
         | None -> None
 
-let rec keymap_extract_remainder istype scope_seen = function
-  | [] -> []
-  | (sc,ntn,interp,c) :: l ->
-      if String.Set.mem sc scope_seen then keymap_extract_remainder istype scope_seen l
+let rec keymap_extract_remainder n istype scope_seen = function
+  | [] -> List.make (n+1) ([],[])
+  | (sc,ntn,interp,rigidity) :: l ->
+      if String.Set.mem sc scope_seen then keymap_extract_remainder n istype scope_seen l
       else
         match find_with_delimiters istype (Some sc) with
-        | None -> keymap_extract_remainder istype scope_seen l
+        | None -> keymap_extract_remainder n istype scope_seen l
         | Some delim ->
-           let rule = (NotationRule (Some sc, ntn), interp, c) in
-           (rule,delim,true) :: keymap_extract_remainder istype scope_seen l
+          let rule = NotationRule (Some sc, ntn) in
+          add_depending_on_rigidity (rule,interp,rigidity) delim true n
+            (keymap_extract_remainder n istype scope_seen l)
 
 (* Scopes table : interpretation -> scope_name *)
 let notations_key_table =
   ref ((ScopeMap.empty, KeyMap.empty) :
-       notation_rule_core list KeyMap.t ScopeMap.t *
+       notation_rule_core_with_rigidity list KeyMap.t ScopeMap.t *
        scoped_notation_rule_core list KeyMap.t)
 
 let glob_prim_constr_key c = match DAst.get c with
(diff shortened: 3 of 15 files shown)


herbelin:
Hi  @ggonthier, and thank you very much for contributing to the discussion.

> I would think that implicits should not matter in deciding which notation to use for a given (kernel) application sub term (...). The implicit argument signature only comes into play once we've decided to display the term as the application of a constant or abbreviation.

I think that it is correct yes. However, you seem to deduce from that a possible simplification and I'm unsure to see what you have in mind. To take into account implicit arguments, we would still have to remember the extra arguments of the matched expressions, i.e. the arguments `t_{i+1} ... tn` if the first matching notation for `f t1 ... tn` matches only `f t1 ... ti`. Or am I missing your point?

> which, btw, states incorrectly that this also hold for coercions - this can't be true as the RHS is not type-checked

Indeed. Fixed in #9211.

> A separate issue is that if I understand the code correctly, flexible applications (where the head of the application is parameter of the notation) are prioritised by their apparent arity, whereas they should match any larger arity.

Your question makes me realize that there is still a problem with "flexible applications" even with the PR, which is that a recurrence step is missing in case of a n-ary application which does not match at the toplevel, but would match on a partial sub-application.

Regarding the question of priority, let me consider an example to try to visualize what you have in mind:
```coq
Parameter add : nat -> nat -> nat.
Notation "f '.s' x" := (f (fst x)).
Notation "add1 x" := (add (fst x)).
Notation "add2 x y" := (add (fst x) (fst y)).
Check add (fst x) (fst y).
```
Are you saying that we should consider giving to `"f '.s' x"` the same level of priority as `"add2 x y"` when applied to `add (fst x) (fst y)`?

> Finally, you might want to also use this PR to document the way the implicits signature of abbreviations is derived, or indeed that abbreviations can have implicits. This is definitely a useful feature, but I'm unsure of how it works, and how it can be controlled 

Basically, abbreviations to an applied constant inherit the implicit arguments of the underlying constant (unless the abbreviation is exactly a `@ref`, in which case this deactivates the implicit arguments - didn't we have discussions about that long time ago?).

> (Does `Arguments` apply to abbreviations? Can abbreviations have `{}` arguments?).

Do I interpret correctly that you are asking these questions because you see an interest in supporting these kinds of possibility?

> And would it be possible not to separate an applied abbreviation from its extra args with redundant parentheses?

I'm not fully sure I understand what you mean. Are you thinking about using abbreviations as an alternative to using parentheses in their rôle of stopping the effect of implicit arguments? Would you have an example of what you have in mind?

anton-trunov:
@herbelin Thank you very much for addressing the issue. I tried this PR locally and found one thing I was hoping you could look into:

```coq
From Coq Require Import ssreflect ssrbool.

Definition leq (m n : nat) := true.
Notation "m <= n" := (leq m n) : nat_scope.
Notation "m <= n <= p" := ((m <= n) && (n <= p)) : nat_scope.

Goal 0 <= 0 <= 0.
(* This is what I get back: (0 <= 0) && (0 <= 0) *)
(* Coq 8.8.2 keeps the goal  as `0 <= 0 <= 0`. *)
```

ggonthier:
> To take into account implicit arguments, we would still have to remember the extra arguments of the matched expressions, i.e. the arguments `t_{i+1} ... tn` if the first matching notation for `f t1 ... tn` matches only `f t1 ... ti`.

Correct, but those only come back into play once we've decide which notation or abbreviation to use to display the application, if any - we print that, followed by its arguments if it's an abbreviation, then the extra args, possibly filtered by an implicits signature if it's  a constant or abbreviation. My point was you shouldn't need a `RefDeactivatingImpls` tag (and I meant to write "Bottom line, abbreviations of unapplied contants should _not_ need special handling", silly me!).

> Are you saying that we should consider giving to `f '.s' x` the same level of priority as `add2 x y` when applied to `add (fst x) (fst y)`?

I think so; conceivably you could break the tie in favour of the non-flexible notation in this case, but you certainly would not want a `Notation gxy := (g x y)` (arity 2) to mask `f .s x` (apparent arity 1) in displaying `g x y (fst z)`. And the logic for breaking the tie would be to favour the more precisely matching notation, which is not otherwise used for prioritising notation, as well as dubious: a flexible notation could give a more specific pattern for its arguments than a longer fixed-arity one.

> Basically, abbreviations to an applied constant inherit the implicit arguments of the underlying constant (unless the abbreviation is exactly a `@ref`, in which case this deactivates the implicit arguments - didn't we have discussions about that long time ago?).

Fine - but this should be in the manual, as well as what happens when some of the arguments are specified, directly or with the `(a := v)` syntax, or when the term is `(f x) y`, or the user writes `@abbrev`, or if `Arguments` is used later to modify the implicits signature of the head constant...

> > (DoesArguments apply to abbreviations? Can abbreviations have {} arguments?)

That's a bit of an idle feature wish, and I don't expect you or @gares to take it up with any urgency, on the principle that I prefer to have the means to redress any choices Coq may have made for me that I'm uncomfortable with. Being able to have implicit abbreviation parameters would make it possible to specify non-linear patterns (e.g., `matrix _ ?n ?n`) with just one layer of abbrevs.

> Are you thinking about using abbreviations as an alternative to using parentheses in their rôle of stopping the effect of implicit arguments? Would you have an example of what you have in mind?

In my current version (8.8.1), I'm getting that
```
Definition myAnd A B := A /\ B.
Notation myAnd1 A := (myAnd A).
Goal myAnd1 True True.
```
displays `(myAnd1 True) True` with the undesirable parens; I'm guessing this is an echo of a similar issue with `fun_class` coercions.

While I've used the redundant parens idiom to stop the insertion of (on-demand) implicits, I'm not a big fan of it; for one thing, half of the time I need this is in the nullary case (viz. `(f) x`) where it doesn't work, forcing me to either resort to the `(x31 := v)` syntax, use a `let` eta-expansion, or use `@` and write in wildcards for all the _maximal_ implicits that precede the on-demand ones I want to specify.

herbelin:
@anton-trunov:

As far as I can judge, your example falls exactly in a situation affected by the priority given to open scopes rather than to the order of declaration of notations. Maybe someone would have a luminous idea about how to have your example working the same as before without any change but I'm afraid otherwise that it would require an adaptation to the new model, which, if ever different, reminds however more controllable as far as I can judge.

Possible changes are to put `"m <= n <= p"` in `bool_scope`, or to have `nat_scope` prioritary over `bool_scope` (but then having to take care of the other possible impacts such a change would have), or to have a new scope, say `nat_in_bool_scope` containing `"m <= n <= p"` and opened after `bool_scope`.

Are there other such changes that you could observe (besides the ones you already reported)? It would be useful to have an evaluation of the typical consequences of the change of priority model, specifically on math-comp which is a "heavy user", and on how easy or not it is to adapt.

PS: I realize that `Print Visibility` needs to be adapted to the new model. Actually, there are a couple of reports about how to improve information retrieving for notations. Some synthesis of needs should be done at some time.

herbelin:
> My point was you shouldn't need a RefDeactivatingImpls tag

Maybe there is something I don't see but it is at some time needed to remember the difference between `Notation foo := @bar` and `Notation foo' := bar`. The first one deactivates the implicit arguments and the second does not. It could be possible to retrieve it from the encoding of `@bar` as `App (Ref bar)` and of `bar` as `Ref bar`, but this is precisely the point of using an explicit name `RefDeactivatingImpls` rather than an encoding which even for myself requires a phase of deciphering every time I look again at it.

> Fine - but this should be in the manual, as well as what happens when some of the arguments are specified, directly or with the `(a := v)` syntax, or when the term is `(f x) y`, or the user writes `@abbrev`, or if Arguments is used later to modify the implicits signature of the head constant...

Well... maybe Coq is victim of the success of those who developed so extraordinarily impressive applications with it, but, up to say 6 years ago, the idea was that Coq was in large parts experimental and it was not clear that documenting such somehow marginal semantics was important. Even now, I'm not sure that the `(f x) y` is worth being documented (assuming it is not already, I did not check).

In the context of today, the `(a:=v)` (which works as if the abbreviation had been expanded) or the effect of `Arguments`, or `@abbrev`, yes, they should certainly be documented. But somehow, help would also be useful on this matter.

> I think so; conceivably you could break the tie in favour of the non-flexible notation in this case, but you certainly would not want a Notation gxy := (g x y) (arity 2) to mask f .s x (apparent arity 1) in displaying g x y (fst z). And the logic for breaking the tie would be to favour the more precisely matching notation, which is not otherwise used for prioritising notation, as well as dubious: a flexible notation could give a more specific pattern for its arguments than a longer fixed-arity one.

Yes, I agree. I shall think about it.

> In my current version (8.8.1), I'm getting that
>  ```coq
>  Definition myAnd A B := A /\ B.
>  Notation myAnd1 A := (myAnd A).
>  Goal myAnd1 True True.
>  ```
>  displays `(myAnd1 True) True` with the undesirable parens; I'm guessing this is an echo of a similar issue with fun_class coercions.

Thanks for reporting. I made #9214 for it.

For the rest of your comments, I'm a bit perplex about what to do. If you want to contribute yourself to the development of Coq, I believe that your clear-sightedness and experience will be most welcome. If you prefer to work on your side and "redress" the choices made by Coq which you are uncomfortable with, that's of course also possible. Alternatively, if people familiar with the needs of math-comp in term of notations are willing to contribute and enrich/fix/improve/document the notation subsystem for the general benefit of all users, they are also welcome.

ejgallego:
> For the rest of your comments, I'm a bit perplex about what to do. If you want to contribute yourself to the development of Coq, I believe that your clear-sightedness and experience will be most welcome. If you prefer to work on your side and "redress" the choices made by Coq which you are uncomfortable with, that's of course also possible. Alternatively, if people familiar with the needs of math-comp in term of notations are willing to contribute and enrich/fix/improve/document the notation subsystem for the general benefit of all users, they are also welcome.

As the assignee of this very complex PR I am trying to decipher what the status is, however I am not sure where is this bit coming from. I did not observe Georges making any kind of comment in this style, what you refer "the rest of the comments", in particular about whether he is "on his side" or "in the side of Coq".

So I am not sure why the commentary is going away from the technical discussion suddenly.

herbelin:
@ejgallego:

@ggonthier said:
> That's a bit of an idle feature wish, and I don't expect you or @gares to take it up with any urgency, on the principle that I prefer to have the means to redress any choices Coq may have made for me that I'm uncomfortable with. Being able to have implicit abbreviation parameters would make it possible to specify non-linear patterns (e.g., matrix _ ?n ?n) with just one layer of abbrevs.

I felt unable to give a concrete answer this paragraph. On one side, my own time is limited. On another side, it also seems to me that it would make many people happy if @gares had the opportunity to do something. On yet another side, I feel that we would have a lot to gain if @ggonthier, with his expertise, could directly contribute. On yet yet another side, @ggonthier said he wanted to "redress any choices Coq may have made for me that I'm uncomfortable with" which I interpreted as not being willing to work at the level of Coq so that it makes better choices but rather to "bypass" the choices made by Coq (did I overinterpret?). So, at the end, what I wrote was an attempt to express my perplexity about what to do to make progresses.

Does it explain clearlier what I was trying to say?

> As the assignee of this very complex PR

Maybe I can explain it into more details, but it is simpler than #8187 itself.

By the way, I'll push soon a patch to give a proper support to generic notations for applications (i.e. application with a flexible head), answering in particular to the cases mentioned above. I have a question on the way. If I have:
```coq
Parameter g : nat -> forall {A}, A -> A.
Notation "f '.s' x" := (f (fst x)) (at level 9).
Check g (fst (0,0)) 0.
```
Do we consider that the notation propagates the implicit arguments of `g` or not? I would say yes, but would be glad to have more examples at disposal to be sure that it makes more sense than deactivating the implicit arguments.

ejgallego:
Thanks a lot for the explanation @herbelin ! Indeed it seems to me that you over-interpreted what @ggonthier wanted to say a bit; [he can of course clarify] Many times you are right about people using too strong words when referring to Coq, IMO here "redress" cannot be given a bad meaning.

My understanding of the comment was more of that choice should still be allowed so that library authors can override Coq defaults easily; that IMHO makes sense for two reasons: first, libraries seems like the right territory to experiment as you never know whether a particular choice is going to work at scale; second, Coq itself has many other limits [such as backwards compat, release cycles, etc...] thus some changes may be hard to make.

It seems to me that George is willing to provide very valuable feedback so the core Coq implementation can better work with heavy notation users, but of course that doesn't mean lib authors should give away some flexibility.

As for the complexity comment, I didn't mean the PR itself, but whole notations situation feels pretty complex to me and I need to learn a lot more about the stuff in order to be an effective 2nd maintainer for interp [as I am assigned now]. I'd like at least to be able to provide some shepherding but it is going to take time.

Likely the 2nd in charge should be @ggonthier :) :) not me. [by the way Georges you should have an invitation to join the Coq GitHub org]

ggonthier:
@ejgallego did interpret my remark correctly - @herbelin, sorry I was not clear enough. I did mean to say that as abbreviations are used in a variety of ways, it is going to be very difficult to coin a default arguments signature that works for all cases - it might even be less work to give the user a way to depart from the default than to craft and document precisely a finely dovetailed universal default.
   For instance, I sometimes use abbreviations to partially specialise polymorphic global constants; in such cases it would be useful to retain the same signature for the arguments not specified. I also use abbreviations to compress qualified names. Ideally I'd want to retain all of the signature in that case,
but I'm not sure how to do this for constants with maximal implicits (unless there's a special case for this, they will get inserted in the Notation RHS unless I put an `@` in there, which would drop the signature entirely).

herbelin:
> @ejgallego did interpret my remark correctly - @herbelin, sorry I was not clear enough. I did mean to say that as abbreviations are used in a variety of ways, it is going to be very difficult to coin a default arguments signature that works for all cases - it might even be less work to give the user a way to depart from the default than to craft and document precisely a finely dovetailed universal default.
   For instance, I sometimes use abbreviations to partially specialise polymorphic global constants; in such cases it would be useful to retain the same signature for the arguments not specified. I also use abbreviations to compress qualified names. Ideally I'd want to retain all of the signature in that case, but I'm not sure how to do this for constants with maximal implicits (unless there's a special case for this, they will get inserted in the Notation RHS unless I put an `@` in there, which would drop the signature entirely). 

I understand your technical point and I don't disagree with the sentence "it might even be less work to give the user a way to depart from the default than to craft and document precisely a finely dovetailed universal default". In particular, we can imagine a syntax for telling I want an abbreviation which retains the signature but does not include maximal implicits now (assuming I got your point).

About the non-technical remark of @ejgallego, and to be sure that there is no misunderstanding about your phrasing (I learned along the years that it is better to clarify misunderstandings early than accumulating them, and since I'd be glad to collaborate with you, I'd prefer things to be clear), and sorry for the triviality of the question if ever you find it trivial, but, when you used the word "redress", did you mean something about Coq doing something wrong according to you, since, usually the word "redress" (as far as I can judge from the dictionary, not being a native English speaker) has a connotation of repairing something wrong or a grievance? I'm not asking for myself as I don't feel concerned by these particular choices of semantics for abbreviations in Coq. I'm just asking in the name of the various people who contributed to Coq and who I'm feeeling united and who probably did not intend to do something wrong. And again, maybe I'm attaching too much importance to a use of words which looks maybe anecdotic to you, but, precisely, if it is anedoctic, it should be easy to answer, and then we close the point.

(Note that if I'm asking, it is also because I feel that the development of MathComp precisely maybe suffered from similar misunderstandings and/or from a lack of good interaction with the Coq development. Fortunately, I'm glad to now see convergent and united work e.g. on the design of `Arguments`, or in removing the [`_:Type`](https://github.com/math-comp/math-comp/pull/248) thing in packed classes. So, I'm just trying to understand what is your general positioning on this kind of issues, in full simplicity, for the common benefit of all of us.)

PS: Feel free to reply privately if you think that the discussion should rather be private.

ggonthier:
@herbelin you are correct, this is anecdotic. I did mean "correct something wrong", but in the sense of "wrong for the purpose at hand", not necessarily "fundamentally wrong". There are many ways a feature as basic as abbreviations can be used, so it is after all unlikely that all can be anticipated or even consistent with a fixed strategy for inferring an `Arguments` signature. Just consider we still need to use `Arguments` extensively, even with the numerous flags supported by the implicits and scope inference procedures.

herbelin:
Thanks. I'm still wondering whether we can make a concrete proposal emerge from your remarks, but your new phrasing is clearer, and I'm grateful to you for it.

ggonthier:
@herbelin wrote
> I'm still wondering whether we can make a concrete proposal emerge from your remarks
I'll try to write down my ideas, so that you and @gares can decide what you'd like to pick from them, either now or later.

- My main point is that the treatment of constants and abbreviations should be factored as much as possible, and in particular that they should use the same data and functions to represent and exploit their syntactic argument signature. Namely, this data should be given default values by the constant or abbreviation  declaration, amended using the `Arguments` command, displayed by `About`, and used to elaborate and display (possibly applied) instances.
- The default for an abbreviation should consist of the names and scopes of its explicit arguments, followed by the data for the extra arguments of its head constant, if the pattern is a partial application or constant, regardless of how this application is written (including the use of `@`: it will be much clearer to specify `Arguments ... : clear implicits` to disable implicits).
- Elaboration of applications of constants and abbreviations should be handled by the same code, using the signature data and annotations within the application (i.e., `@` or `(x := v)`) to insert implicit arguments, and determine the scope stacks under which these arguments are elaborated (and check syntax class and partial application restrictions, see below). Only then does the the code need to branch for abbreviations by substituting in arguments into its pattern (though even that might be factored by considering a constant also denotes a nullary pattern - itself).
- Display can be similarly factored: once it's been determined that a kernel term should be displayed as an abbreviation or constant possibly applied to a sequence of arguments (including removing head coercions, if applicable), then the display of that application, and the interpretation of the arguments, should be done by the same function of the signature data.
- There are a few items of signature data that will be specific to constants and abbreviations, respectively. The `!` and `/` annotations for `simpl` only make really make sense for constants because they actually apply to the kernel constant, not the displayed qualified identifier (`About` could mention any such annotations as they apply to extra args, though). Abbreviation arguments may have syntactic restrictions (currently only `ident x`) that only make sense for `Notation` pattern-matching; also, they may need additional syntax scope information (see just below).
- `Notation` parameters may need to be elaborated with one or more scopes permanently pushed on the scope stack, possibly in addition to a transient scope (the latter is the only possibility for parameters of constants). `About` should mention this, we should have `Arguments` syntax for this - `%(d_1 ... d_n)` seems the obvious choice. It would make sense to support this for constant arguments, for instance we might want to specify a persistent `nat_scope` for a `T -> nat` argument so that when passing `(fun x => x + x)` the `+` gets interpreted in `nat_scope`.
- Besides `only parsing` and the implicitly supported `indent x`, there are many `Notation` annotations that make sense for abbreviations: `only printing`, `global x`, `bigint x`, `binder x`, `pattern x`, all make sense, could be useful, and should be easy to support. It's somewhat paradoxical that we should push users towards disruptive general notation creating new keywords when all they want is vanilla prefix application syntax.
- Indeed it would be desirable to support overloading abbreviations in different syntax scopes, for the reason above; it seems moderately easy to implement (mod determining the appropriate scope stack while allowing for different arity / `Arguments` for the various overloads).
- It makes sense to allow for implicit arguments for abbreviations, including using `{}` in the definition. For example `ssreflect` essentially provides
```
Notation id {T} := (fun x : T => x).
```
where the `T` can usually be inferred form context, but may need to be supplied with `@id T` - but I needed to hack this with a separate `Notation "@id T"`.
- Supplying `_` to a notation or an abbreviation is somewhat tricky as one should keep track on non linearities if the argument is used multiple times. The current Coq gets this wrong, e.g.,
```
Notation bar T := (@fst T T).
Notation f1 x := (bar _ x).
Fail Goal forall x, f1 x = 2.
(* Cannot infer this placeholder of type "Type".*)
Goal forall x : nat * bool, x.1 = 2.
(* Displays forall x : nat * bool, f1 x = 2 *)
```
and I'm afraid this may require some work to fix if the raw terms don't allow linking placeholders...
- Matching bound variables with a pattern, and terms that appear both inside and outside of a binder, is another tricky procedure, which Coq apparently does not get quite right:
```
Notation foo0 x t := (x = t \/ (forall x, x = t) /\ x + x = t)%N.
Goal forall x : nat, x = 2 /\ foo0 x (x + 2).
(* Displays forall x : nat, x = 2 /\ foo0 x (x + 2) as expected. *)
Abort.
Notation foo1 x := ((forall x, x = 3) /\ x = 2).
Goal forall n, foo1 n.
(* Displays forall n : nat, (forall n0 : nat, n0 = 3) /\ x = 2. *)
Abort.
Notation foo2 x := (forall x, x = 2).
Goal forall y, y = 2 /\ foo2 y.
(* Displays forall y : nat, y = 2 /\ foo2 y0 *)
Abort.
```
- The display function should take a list of extra args in order to properly handle coercions to `Funclass`. Contrary to the comment I wrote that lead to #9214 Coq still adds redundant parentheses around such coercions, which interfere with `Arguments` declarations.
```
Record Pnat := {inPnat :> nat -> Prop}.
Goal forall r : nat -> Pnat, r 2 3.
(* Displays forall r : nat -> Pnat, (r 2) 3. *)
```
I'm not sure what the point of #2044 was - a spurious `""` in the display, or the `%nat` scope delimiter. The former is clearly a bug, but the latter may be proper if `"0"` is overloaded in some other scope that has been pushed on top of `nat_scope`. One can't rely on the `Arguments` data of the coercion constant, since is will not be present in the displayed term. I also have to question the logic the code in which the comment is embedded: coercion elision should be done during externing, not as a separate pass, so that `Notation` for the coerced term can be adequately recognised. (The proper way to address scopes of extra args is though `Arguments`, though for this to work for this example we'd have to either extend `Arguments` to cover general notations, or allow an `arguments ...` annotation for Notations - maybe that's pushing it a bit).
- A final, rather unrelated nitpick: it seems `About` gets confused by `Import` when displaying `Notation`:
```
Module Import Mfoo.
Module Foo.
Definition FooCn := 2.
Module Bar.
Notation Cn := FooCn.
End Bar.
End Foo.
Export Foo.Bar.
About Cn. (* Displays Notation Cn := Foo.FooCn *)
End Mfoo.
About Cn. (* Displays Notation Cn := Cn !! *)
```

ggonthier:
@anton-trunov :
I am aware the issues with the `x < y < z` syntax and more generally with the overloading of relation operator notation. Their root cause is that Coq only does top-down syntactic overloading resolution, rather than the (complete) usual bottom-up-then-top-down resolution. So although one can put notations for the `X` relations in `X_scope`, that scope will not be added to the scopes stack in contexts that expect relations - what will be added, if anything, is `bool_scope`.
  We can thus only weakly select `X` notations by a global `Open X_scope`, or point wise with an immediate scope delimiter. For `x < y < z` you have the added problem of the `andb` notation taking precedence, which we can and should improve by opening `nat_scope` in `ssrnat`, but which will never truly disappear as the `&&` will reappear in any context that imposes `bool_scope`,
```
Check 1 < p < 11 -> if 2 < p < 4 then true else false.
(* 1 < p < 11 -> if (2 < p) && (p < 4) then true else false. *)
```
One possible alternative would be to hide the `andb` in `x < y < z` behind an alias. It would be convertible, but we'll still have to adjust quite a few `rewrite`s in the `mathcomp` library.

anton-trunov:
I replaced all `Proof.` with `Proof. Show.` in `ssreflect` subdirectory of `mathcomp` and recorded the compilation output. Here you can find the diff between Coq 8.8.2 and this PR: https://github.com/anton-trunov/coq-notations-diff/commit/4c3d1b06107ca4d7c703cd4a11f7f4e31ee101b9.

Of course, this does not show every observable proof state, but I hope it is still useful. I'm wondering if there was more elegant way and if it's easy to capture every proof state as if the user stepped through the files in ProofGeneral of CoqIDE.

herbelin:
@ggonthier: thanks for your detailed comment.

With respect to your first five paragraphs, I would not have spontaneously treat abbreviations as the same kind of objects as ordinary references, but your proposal makes sense and I have no objections against it. In particular, if it happens to be useful in your own practice or in the practice of other users, that is an excellent reason to propose a PR on it. And, indeed, that would also provide with a nice way to get rid of the ad hoc treatment for `Notation a := @b`.

> Notation parameters may need to be elaborated with one or more scopes permanently pushed on the scope stack, possibly in addition to a transient scope (the latter is the only possibility for parameters of constants). About should mention this, we should have `Arguments` syntax for this - `%(d_1 ... d_n)` seems the obvious choice. It would make sense to support this for constant arguments, for instance we might want to specify a persistent `nat_scope` for a `T -> nat` argument so that when passing `(fun x => x + x)` the `+` gets interpreted in `nat_scope`.

You may then be interested to comment on #9123 which is related. Another "obvious" choice could also be `x%d1...%dn` which would have the virtue of matching the syntax in terms (I'm using here the semantics of `%` proposed in #9123). To get `nat_scope` on the body of an argument of type `T -> nat`, I considered at some time to make temporary scopes spanning over lambdas. I was however missing examples to evaluate whether this was an interesting option or not. I'd be interested to know if such strategy would for instance be useful for MathComp...

> Besides `only parsing` and the implicitly supported `ident x`, there are many `Notation` annotations that make sense for abbreviations:

> `only printing`

Absolutely.

> `global x`, `bigint x`, `binder x`, `pattern x`

Sorry, I fail to see the semantics you give to this. Abbreviations are parsed using the general rule for parsing applications. Do you mean changing this?

> It's somewhat paradoxical that we should push users towards disruptive general notation creating new keywords when all they want is vanilla prefix application syntax.

Sorry, I fail to see in your sentence who are the "users", and who are the "we" who are pushing these users. In particular, are Coq developers concerned in the "we"? Did Coq developers push you in the past, or did they disregard your interest in abbreviations? If so and that I was some of those, or that some words from me were interpreted as such, I'm sorry and I regret it. If not, I don't understand why you are saying that.

> Indeed it would be desirable to support overloading abbreviations in different syntax scopes, for the reason above; it seems moderately easy to implement (mod determining the appropriate scope stack while allowing for different arity / `Arguments` for the various overloads).

An obstacle I had in trying to add scopes to abbreviations is that one would need to have the name resolution (nametab) depending on scopes, and actually even interleaving with scopes. If you see how to do that in a reasonably simple way, I'm interested.

> but may need to be supplied with `@id T` - but I needed to hack this with a separate `Notation "@id T"`.

Just to be sure that there is no misunderstanding about your use of the word "needed": do you see a responsiblity of "Coq" here? If you decided to use a hack, rather than, say, propose a patch to Coq, it was your choice and "Coq" is in no way responsible of your choice, right? (Again, I'm asking because there were many misunderstandings between ssreflect/MathComp developers and Coq developers in the past and I prefer to avoid new misunderstandings.)

> Supplying `_` to a notation or an abbreviation is somewhat tricky as one should keep track on non linearities if the argument is used multiple times. The current Coq gets this wrong, e.g.,
> ```
> Notation bar T := (@fst T T).
> Notation f1 x := (bar _ x).
> Fail Goal forall x, f1 x = 2.
> (* Cannot infer this placeholder of type "Type".*)
> Goal forall x : nat * bool, x.1 = 2.
> (* Displays forall x : nat * bool, f1 x = 2 *)
> ```
> and I'm afraid this may require some work to fix if the raw terms don't allow linking placeholders...

I would see this more as a feature request than as a bug. If you write `Notation f1 x := (bar _ x).`, it is expected to be the same as `Notation f1 x := (@fst _ _ x).` and the sharing present in `bar` is normally lost. So, I see this as a wish to be able to express a constraint between subterms which are unbound in the left-hand side of a notation (whether it is a parsing-level notation or an abbreviation). That's a very interesting idea.

> Matching bound variables with a pattern, and terms that appear both inside and outside of a binder, is another tricky procedure, which Coq apparently does not get quite right:
> ```
> Notation foo0 x t := (x = t \/ (forall x, x = t) /\ x + x = t)%N.
> Goal forall x : nat, x = 2 /\ foo0 x (x + 2).
> (* Displays forall x : nat, x = 2 /\ foo0 x (x + 2) as expected. *)
> Abort.
> Notation foo1 x := ((forall x, x = 3) /\ x = 2).
> Goal forall n, foo1 n.
> (* Displays forall n : nat, (forall n0 : nat, n0 = 3) /\ x = 2. *)
> Abort.
> Notation foo2 x := (forall x, x = 2).
> Goal forall y, y = 2 /\ foo2 y.
> (* Displays forall y : nat, y = 2 /\ foo2 y0 *)
> Abort.
> ```
Yes, this is indeed tricky and I guess that you noticed that this is due to whether it is the free variable or the binding variable which is matched first. This is tricky because this interferes with alpha-conversion. In the first case, `x` is bound to `x` and alpha-conversion sees that `x0` can be renamed to `x` so that it matches. In the second case, `x` is bound to `x0` and it is too late to check that this binding can safely be alpha-converted to `x` when trying to match the free variable name. I somehow postponed this issue at some time. The code is in `notation_ops.ml` around the `bind_..._env` functions, if ever someone wants to have a look (see also `add_env` for the alpha-conversion).

Note that this is independent of whether it is an abbreviation or a regular notation. Note also the discussion at #8808 which you may want to participate in: is it useful to have abbreviations able to bind binding variables or is it more troublesome than useful?

Your remark also reminds me that thanks to pattern in notations, the standard library is able to provide an `if t is pat then u else v` which is recognized at printing time (up to some heuristics in re-factorizing pattern-matching clauses). On the other side, there is still an ad hoc non-printed `if t is pat then u else v` coming with `ssreflect.v`. Maybe a good time to address a couple of ssreflect things that ssreflect does not do "quite right"?

> ```
> Record Pnat := {inPnat :> nat -> Prop}.
> Goal forall r : nat -> Pnat, r 2 3.
> (* Displays forall r : nat -> Pnat, (r 2) 3. *)
> ```

Thanks for the explicit example. Then, good news, this is also fixed by #9214. (And, for the record, it is not related to #2044, i.e. not related to the function which removes coercions but to the function which selects which arguments to make implicit).

> One can't rely on the `Arguments` data of the coercion constant, since is will not be present in the displayed term.

See a proposal below.

> I also have to question the logic the code in which the comment is embedded: coercion elision should be done during externing, not as a separate pass, so that `Notation` for the coerced term can be adequately recognised.

In which sense are you saying that this is separate pass. As far as I can judge, it is done as [part](https://github.com/coq/coq/blob/3099be05553dab10b41d864f4981860eb105f145/interp/constrextern.ml#L766) of externing.

> (The proper way to address scopes of extra args is though `Arguments`, though for this to work for this example we'd have to either extend `Arguments` to cover general notations, or allow an `arguments ...` annotation for Notations - maybe that's pushing it a bit).

At some time, I was wondering whether it could be useful to indicate the coercion in the `Arguments` command. Can you give more details on your test case? If you indicate extra scopes in `Arguments` in prevision of a coercion being applied, I guess that this works because you know which coercion is expected to be inserted. So, I was wondering whether indicating in `Arguments` the name of the coercion to use in case of extra arguments would not be better than indicating the extra scopes themselves. By indicating the name of the coercion, one would inherit automatically not only the scopes but also the implicit arguments of the coercion. On the other side, this would not work if the coercion is polymorphic and that the information on extra scopes is relevant only for a particular monomorphic instance of the coercion. So, it may depend on the examples that you have in mind.

> A final, rather unrelated nitpick: it seems `About` gets confused by `Import` when displaying `Notation`

This looks indeed like a little bug probably related to "canonical names": in the presence of modules, constants have a user name and a canonical normalized name and I suspect that printing the r.-h. side of `foo` does not deactivate the right copy of the name `foo`. I'll try not to forget to look at this. By the way, I'm curious, does it come from a realistic example or did you find it by some kind of investigation?

ggonthier:
@anton-trunov : Thank you for digging into this - you've uncovered quite a few more instances of regression.
It would indeed be interesting to extend this compatibility check to intermediate script lines, once most of the problems you've unvcovered have been resolved (otherwise the output will be too cluttered). The math comp scripts follow a uniform layout, with at most one tactic phrase per line, and `Proof` and `Qed` on separate lines unless the entire proof fits on one line. Thus, you should be able to get away with inserting a `Show.` at the start of each line between `Proof.` and `Qed.` lines, which is preceded by a line ending with a `.`, and after a `Proof.` whose line ends with `Qed.`
   Besides the adverse effects of the new strictly-by-scope display notation priority policy which started this discussion, to which I'll come back below, your test uncovered three separate issues, which actually account for most of the lines:

1. Many instances of coercions, particularly `pred_of_set` and `fun_of_fin` are no longer elided. This is likely an unwanted consequence of recent PRs, either a change in visibility rules or a bug in the elision procedure.

2. It seems that Coq now performs eager eta-contraction, and this systematically inhibits comprehension notation which requires explicit lambdas. This should be reverted: the change assumes that `f` is always more readable than `fun x => f x`, but this is totally wrong for comprehensions (as well as for big operators, but the syntax used in math comp avoids this issue).

3. The `%coq_nat` delimiter is incorrectly elided, even though it is required after the scope reshuffling performed by `ssrnat`.

Your check shows that the new display priority causes regressions for two other classes of notation, in addition to the interval inequalities `x < y < z` you have been using: equality mod `m == n %[mod d]` and bigop specialisations for nat such as `\sum_(0 <= i < n) 'C(n, i)`.

Superficially, one could resolve most of these by a better management of the scope stack. It just happens that ssreflect/mathcomp open `bool_scope` and `big_scope` _after_ `nat_scope`, since the latter is opened in the Coq `Prelude`, and don't bother with the effect on priorities as they used to be irrelevant here. Closing then reopening `nat_scope` will indeed fix all of these, except for the right hand side of `chinese_remainder`, where the `&&` will prioritise `bool_scope` regardless of how the global scope stack is ordered.

Unfortunately, this one case is an instance of a much larger and, I'm afraid, unsurmountable problem. First, we should have seen instances of it for the `m = n %[mod d]` notation, which abbreviates `modn m d = modn n d`, because the notation for `eq` should have priority in any `%type` context, such as the goal and the type of assumptions... I'm not sure why that didn't show up. Second, extending the test to the whole of math comp would have uncovered _many_ more instances of notations for specialisations of generic concepts with notations (`&&`, `==`, `=`, `big[/]_() `, and also `\in` and `\subset`), several of which are defined in specific scopes that are _never opened_, such as `%MS` (matrices treated as row spaces), `%G` (subgroup structures), or `%A` (algebraic number divisibility). To summarise, the new policy is fundamentally incompatible with a significant fraction of the higher-level algebra notation in the math comp library.

The root cause of this conflict is that the new policy is to use the scope stack to prioritise the choice of notation for a given term whereas it is now used solely for choosing the term denoted by a given notation. As these two problems a nearly orthogonal, it's unsurprising that this works out poorly. The new policy amounts to greedily minimising the number of scope delimiters used in the display. This is not what I want as a user: I'm quite happy with delimiters, indeed I often prefer reusing common symbols with a new delimiter to inventing silly new ones. What I usually want `extern` to pick is the most specific notation (though not always); this is difficult to determine automatically (two notations may specialise different sub terms), and "most-recently-declared" can provide a reasonable approximation. @herbelin, I understand you'd prefer a more declarative system, but you'd really need to support a completely separate priority hierarchy, and I'm not sure it's worth the complexity...

ggonthier:
> You may then be interested to comment on #9123 which is related. Another "obvious" choice could also be x%d1...%dn which would have the virtue of matching the syntax in terms (I'm using here the semantics of % proposed in #9123). 

Unfortunately the syntax proposed in #9123 is incompatible with math comp, which uses `%%` for the modulo operator. Something like `%1` would be safer, and also suggest the scope is only used once. One way to get compatibility with the existing `%` would be to use `%!` to mean "persistent push", and say that `%` means to `%1` in `Arguments` and `%!` in terms, and go from there. I suspect that eventually this will get simplified to `%` meaning `%1` everywhere, because most instances of `%` in terms do not need to be deep (and , in fact, those introduced by the pretty-printer almost never are. On thing that this does seem to allow is pushing several transient scopes for a given sub term - but come to think of it `Arguments` does seem to allow this as well, so perhaps this could be implemented, which would also get rid of the "type T already bound to scope s" error.

OTOH, with my proposal one would get the functionality of #9123 with an abbreviation:
``
Notation inType x := x.
Arguments inType x%type.
Notation "{ x | P }" := (sig (fun x => inType P)) : type_scope.
```
and one can get almost the same making inType a definition. This is indeed used in several places in math comp and fourcolor, especially with non-opened non-delimited scopes.

**Comment by ggonthier** [comment]
> it seems About gets confused by Import when displaying Notation

> By the way, I'm curious, does it come from a realistic example or did you find it by some kind of investigation?

I have little merit here: this occurs in the packed classes coding pattern used throughout math comp, so I noticed it while checking what `About` did with a random abbreviation I had at hand.

**Comment by ggonthier** [comment]
> global x, bigint x, binder x, pattern x
> Sorry, I fail to see the semantics you give to this. Abbreviations are parsed using the general rule for parsing applications. Do you mean changing this?

No, sorry I was being a little terse here. All of these `x`s can be parsed as general terms, so the parsing rule would not be affected. As in the implicit `ident x` that is already supported, these would imply a membership check for a restricted syntax, possibly followed with by a reinterpretation for `binder` and `pattern`. Also, these would affect how such arguments could be used in the abbreviation pattern, e.g., `global` and `bigint` arguments can appear in different scopes, `binder` can only appear in n-ary binders, etc.

**Comment by ggonthier** [comment]
> I considered at some time to make temporary scopes spanning over lambdas. I was however missing examples to evaluate whether this was an interesting option or not. I'd be interested to know if such strategy would for instance be useful for MathComp...

MathComp usually provides some form of comprehension notation, which allows for writing an explicit delimiter if needed - but then this means a persistent rather than transient scope, which only #9123 would solve. On the other hand, with transient scopes span over lambdas (possibly also plain and destructuring lets) one no longer needs #9123 to handle either the ssreflect comprehensions or the `sig` example motivating #9123 - so you've had an example all along.

**Comment by ggonthier** [comment]
> It's somewhat paradoxical that we should push users towards disruptive general notation creating new keywords when all they want is vanilla prefix application syntax.
> Sorry, I fail to see in your sentence who are the "users", and who are the "we" who are pushing these users. 

Sorry, I was going into full design mode, so that "we" was meant as "us designers" and I really should have said "the current Coq design" - nothing personal was intended. But the observation does hold: it is much easier to overload symbolic notation that abbreviations, and emulating an abbreviation with symbolic notation means creating a keyword or at least adding punctuation. For example MathComp defines `Aut G` as the set of automorphisms of a group `G`, but it needs to introduce an alternative syntax `[Aut G]` in order to overload this for the corresponding permutation subgroup. @hivert recently ran into this for defining symmetry groups, so this is something of a recurring theme.

The overall point was that ideally the design shouldn't force users off a the less disruptive abbreviations by  arbitrarily restricting features to general notations. Then of course there's the developer bandwidth concern...

**Comment by ggonthier** [comment]
> An obstacle I had in trying to add scopes to abbreviations is that one would need to have the name resolution (nametab) depending on scopes, and actually even interleaving with scopes. If you see how to do that in a reasonably simple way, I'm interested.

I did realise this was an issue, even more so if one wants to be very precise about using only transient scopes when there are no extra args. One strategy would be to do one-shot nametab resolution first, and fall back on the function called by `Locate` to get all the relevant bindings if the former returned an overloaded scoped abbreviation (so as to minimise the performance cost).

**Comment by ggonthier** [comment]
> Just to be sure that there is no misunderstanding about your use of the word "needed": do you see a responsiblity of "Coq" here?

No there's no moral responsibility - no one is supposed to have perfect foresight! Coq at the time didn't implement this - abbreviations with maximal implicits - but I realised I could cobble together the behaviour I wanted with existing functionality and forged ahead. To my defence, this was a while ago, before Coq had `Arguments` or even maximal implicits, so I wouldn't have been able to formulate a simple extension proposition.

**Comment by ggonthier** [comment]
> I would see this more as a feature request than as a bug. If you write `Notation f1 x := (bar _ x).`, it is expected to be the same as `Notation f1 x := (@fst _ _ x)`. and the sharing present in bar is normally lost.

On second thoughts, you're right: I don't actually want this, and the current behaviour "Notation arguments are _interned_ only once" is more useful than what was suggesting here, which amounts to "Notation arguments are _elaborated_ only once". The latter would forbid polymorphic arguments from being used in different context, as in `{morph phi: x y / x * y}`, the math comp notation for `forall x y, phi (x * y) = phi x * phi y`, in which the implicit rings of the two `*` need not be the same.

The current behaviour does entail that while `'M_n` means "an `n` by `n` square matrix, `'M_(_)` means "an arbitrary rectangular matrix" rather than "some square matrix", which was what was bothering me, but that's a theoretical issue whereas the `morph` notation is really useful. I do note that externing non-linear notations with this rule is really hard, because one should in principle extern each parameter occurrence separately then compare to check whether different occurrences are identical and the notation pattern fits (as a result, the `morph` notation above is more often than not displayed differently).

**Comment by ggonthier** [comment]
> In the second case, x is bound to x0 and it is too late to check that this binding can safely be alpha-converted to x when trying to match the free variable name. I somehow postponed this issue at some time. 

I think this can be resolved by deferring the binding of x from the lambda to the end of the notation match, and bindings arguments that have appeared only under an x binder temporarily to a term with free Refs while x has not been bound, at the end either everything is resolved, or we can pick x to be any of the lambda names, alpha converting only if that name occurs free in one of the parameters that appears under an x binder (plus additional conditions if there are multiple ident parameters!).

**Comment by herbelin** [comment]
Some answers to previous comments:

- Recognition of binders up to alpha-conversion when the name is used also as a free occurrence.

  I finally took the "bull by the horns" and made a preliminary fix which I will publish soon.

- Sharing more features of abbreviations with constants

  >  it is much easier to overload symbolic notation that abbreviations, and emulating an abbreviation with symbolic notation means creating a keyword or at least adding punctuation.

  > The overall point was that ideally the design shouldn't force users off a the less disruptive abbreviations by arbitrarily restricting features to general notations. Then of course there's the developer bandwidth concern...

  Yes, I don't disagree at all with your point. For some, let's say historical reasons, my mind has some difficulties to like abbreviations. In particular, one of the things I have difficulties to accept is that abbreviations are in the nametab. This looks to me like defeating the purpose of a notation. If I would certainly also prefer `Aut G` to `[Aut G]`, I would not like to see `MyModule.Aut G`.

- Scopes spanning over binders

  > On the other hand, with transient scopes span over lambdas (possibly also plain and destructuring lets) one no longer needs #9123 to handle either the ssreflect comprehensions or the sig example motivating #9123 - so you've had an example all along.

  Yes, but I was not missing one example. I was missing examples. So, with respect to the needs of MathComp, is the answer that either #9123 or having scopes spanning over binders would be useful, with no need for both?

- Interpretation modifiers to abbreviations (`global x`, `bigint x`,  `binder x`, `pattern x`, ...)

  OK, I see your point but I'd like to see examples where this is really worth. As told in reference to PR #8808, I'm already not fully convinced that there are a lot of cases where supporting binders in abbreviations give interesting notations. For instance, can we be satisfied with something like `Fun x t u` to mean `fun x : t => u`?

- #9123 

  > Unfortunately the syntax proposed in #9123 is incompatible with math comp, which uses %% for the modulo operator. 

  I'm sure that, besides math comp, there are a lot of users which use `%%` actually. But a solution could be to resort to the "usual" trick and rely on an `%%ident` token. Isn't already MathComp recommending to separate tokens with spaces? So, that should be ok. In passing, we could also change `%id` so that it is a single token, releasing `%` for more notations as long as we accept to put spaces.

  > because most instances of `%` in terms do not need to be deep (and , in fact, those introduced by the pretty-printer almost never are.

  Most instances of `%` in terms do not need to be deep but some do. A common idiom in the standard library for instance is `(t = u)%scope`.

  > On thing that this does seem to allow is pushing several transient scopes for a given sub term - but come to think of it `Arguments` does seem to allow this as well, so perhaps this could be implemented, which would also get rid of the "type T already bound to scope s" error.

  I'm a bit lost in what you are suggesting, but supporting multiple temporary scopes in `Arguments` looks good.

I'm otherwise taking note of your other comments, thanks. I'll come back on #8187 (printing based on scope) and on @anton-trunov's very useful report later on.

**Comment by herbelin** [comment]
First, the conclusion of @anton-trunov's test is indeed clear: developing infrastructures to evaluate printing strategies against large repositories would be extremely helpful.

Second, printing based on scopes started as an answer to the needs of some users looking for a way to control the printing of subterms having several possible notations from the englobing notations themselves. It was not intended to remove possibilities of controlling the printing but to give more control. If it shall not be the case, another way to give more control shall be needed. In particular, it is evident that it should not make the edifice constructed in MathComp more complicated to manage. It should on the contrary offers more possibilities and the current situation is not acceptable.

I shall below give a couple of short term answers to the MathComp regressions but it seems that we should also contemplate altogether how to answer the initial needs for scope-based printing in a way which also brings something for everyone else, even if it possibly means radically changing the approach I adopted.

- Scope-based printing already mixes scope-based and notation-based selection, with the latter currently only for abbreviations. Interleaving open scopes with notations/scopes expected to be used with a delimiter should not be too complicated (keeping an eye on the complexity though).

- It seems that we shall not escape adding scopes to abbreviations as a general rule. I already saw the need a couple of time and the `'I_n'+1` is a new example which would need it.

- For the coercion regression, a possible suspect is #8890 which introduces a heuristic to take into account notations whose head symbol is a coercion. Would there be a way to produce a stand-alone reduced example showing the problem?

- I'm not aware of a change about eta-contraction. Is it in reference to the `image` change? Is it clear that it would not be the order of scope which is involved here?

- For the indeed missing scope `coq_nat`, I know from where the problem is coming. The fix is relatively easy to do.

- > First, we should have seen instances of it for the `m = n %[mod d]` notation, which abbreviates `modn m d = modn n d`, because the notation for eq should have priority in any `%type` context, such as the goal and the type of assumptions...

  This is because there is a special treatment for scope `type_scope` which does not take precedence over other scopes (see the discussion starting from this [comment](https://github.com/coq/coq/pull/8187#issuecomment-430001686)).

**Comment by herbelin** [comment]
Hi, please let me know what you think of the following.

Is it possible for Math Comp users in master to live with the current status for about one week? If yes, I propose by about one week to do the following:

- Add a notion of abbreviation in scope, slightly different from the current notion of abbreviation in that it would not be qualifiable (and thus would take precedence over any constant with same base name). This should address in particular the `'I_n` issue.

- Extend #9123 to support attaching several temporary scopes to an argument. This should allow e.g. to activate the scope where  `x < y < z` live at the same we temporarily activate `bool_scope`, keeping it in front.

- Add a new command, say `Promote Scope scope` to tell to add a scope in the priority stack for printing but without making it accessible without a delimiter. Suspectingly, this would allow to support non-opened scopes such as`%MS`, `%G`, or `%A` at the expected level of priority.

My feeling is that all these features are anyway interesting, independently of how notations are disambiguated for printing. Please tell what you think.

In parallel, we track the coercion regression and I'm fixing the `coq_nat` problem.

What other issues will have to be solved?

[In the longer term, designers and users of notations could try to identify the different sources of conflict between notations and how to disambiguate them. Why do I want this notation to be used? Because it matches a more specific expression? Because I want to use the notation relevant to the context I'm working in? Because this is the notation I was last talking about? How do I want to tell this to Coq? Etc. (Sorry if this looks like trivial remarks but it is to initiate the discussion.)]

**Comment by anton-trunov** [comment]
> For the coercion regression, a possible suspect is #8890 which introduces a heuristic to take into account notations whose head symbol is a coercion.

Here is the impact of #8890 on ssreflect (compared to Coq 8.8.2 as previously): 
https://github.com/anton-trunov/coq-notations-diff/commit/ce63895aa1538e3b56e8bdf9a3ab2bb4f293e6d4

> Would there be a way to produce a stand-alone reduced example showing the problem?

I hope this helps:

```coq
Definition unit_to_bool (_ : unit) : bool := true.
Notation abbrev := unit_to_bool.
Coercion abbrev : unit >-> bool.

Goal true = tt.
(* true = abbrev tt *)
```

ggonthier:
> MathComp side: for `fun_of_fin` for instance, the name seems to be mostly unused at least for parsing. Is the primary purpose of the alias to lighten the output when the coercion cannot be dropped? Is the purpose to save the very few occurrences where `fun_of_fin` would otherwise have to be qualified for parsing?

The `Notation` is part of the coding pattern for creating "hard" opaque constants by putting their definition in a submodule and hiding them via signature ascription. The `Notation` is used to change the apparent location of the definition so that it can be used as if it had been defined in the top-level module. This ensures this coding pattern can be transparently swapped with less cumbersome, but slightly less effective idioms for defining hard opaque constants using `locked_with`, or possibly the new `Opaque` declaration, if this is needed to speed up type checking.
   Exporting the submodule while making `Local` its other components (equalities that reveal the hidden constant definitions) does not quite achieve the same thing, as if the enclosing module is not imported then the qualified names still refer to the submodule. 

At any rate the coding pattern makes it very clear that the abbreviation should be used as a coercion, because it actually declares the _abbreviation_ as a coercion, _after_ defining the abbreviation. Thus, there should be no conflict in this case. Another reason there should not be a conflict is that a coercion declaration for `C` in effect declares that `C a1 ... an x` should be abbreviated by `x` (where the `ai` are the coercion parameters). Thus according to the "longest match" rule we've been discussing all along this should have precedence over any notation of a partial application of `C`. To override the abbreviation of coercions one should have to define a notation for the fully applied coercion, and then given it precedence, whether by defining the notation after as in 8.8.2, or by a new mechanism to emerge from this conversation.

ggonthier:
> Is it possible for Math Comp users in master to live with the current status for about one week? If yes, I propose by about one week to do the following:

Certainly: MathComp is developed in loose sync with Coq, so most work is still done first in 8.8.2, then checked by CI for compatibility with older and newer editions. Since these checks are mostly for compilation they will not be affected by printing issues (that's a problem we'll have to address too!).

ggonthier:
> On the other side, there is still an ad hoc non-printed if t is pat then u else v coming with `ssreflect.v`. Maybe a good time to address a couple of ssreflect things that ssreflect does not do "quite right"?

Actually that syntax is defined in the ML plugin code, because of the intricacies of dual-purposing an `as`clause in the pattern (only the boolean `if` is defined in `ssreflect.v`, because that is not an issue). The bit of notation in the prelude does not address this (nor the `in` and `return` clauses either, for that matter), and also does not support the `isn't` variant. More work would be needed, including extensions for notation patterns that seem rather too specialised. Perhaps it would be preferable to provide hooks to allow plugins to flag their `match`es and provide their own externing code for them?
  This

ggonthier:
> coercion elision should be done during externing, not as a separate pass

As I've indicated above, for the purposes of externing a coercion should be treated as an abbreviation of `C ... x` by `x`. Thus, there should be no "recursive removal of coercions", because after applying the coercion abbreviation above, externing should simply recurse on `x`, to allow for other notations, especially if `x` is the application of a coercion - this does mean recursive calls to extern need to be passed a list of (kernel) extra args.

ggonthier:
> So, I was wondering whether indicating in Arguments the name of the coercion to use in case of extra arguments would not be better than indicating the extra scopes themselves.

Although the computation of default arguments could indeed look for known coercions to `Funclass` to populate extra args, this does not cover all possibilities - for instance there could be coercions to a type convertible to a function, but does not coerce to `Funclass` (such as the `ssreflect` `pred T`), the extra arity could come not from a coercion but from an expected value of a parameter on which the return type depends (think n-ary functions), and the correct scope and implicitness may depend on the specific constant at hand rather than the coercion itself, or may be different from the `Bind` defaults. For example MathComp sometimes put finite set arguments  of group operations in the finite group scope because they sometimes are the result of (lifted) group operations.

ggonthier:
> I'm not aware of a change about eta-contraction. Is it in reference to the `image` change? Is it clear that it would not be the order of scope which is involved here?

It is - `image` is an abbreviation, and the comprehension syntax is notation for its specialisation to an explicit lambda, so this can be partly explained by the new priorities. What got me confused is that there must also be new code that elides eta-expansions on display since otherwise we should not have seen `image f A` but `image (fun x => f x) A`, or rather `image [eta f] A` since MathComp has notation for explicit etas.

ggonthier:
> Is it possible for Math Comp users in master to live with the current status for about one week? If yes, I propose by about one week to do the following:
> Please tell what you think.

Your enthusiasm is much appreciated, but please don't slave over this during holidays!
In addition to what you propose, and possibly in lieu of it, I would like the ability to specify a separate  printing scope, which would be used solely for the purpose of determining display priority. Thus I could give a Notation specialising a previous one the same printing scope, so that it would be preferred if applicable (perhaps "precedence" would be a better  term than "scope", here). Lone notations and abbreviations could be given printing scopes without having to implement overloaded name tab lookup.

herbelin:
> Thus according to the "longest match" rule we've been discussing all along this should have precedence over any notation of a partial application of C

I'm buying the longest match rule as a general rule and thus to answer, to whom may want to use an abbreviation to force the display of a coercion, that extra arguments have to be added to the abbreviation.

While we are at #8890, the current criterion in the presence of a notation for both `c params (t v1..vm) u1..un` and `t v1..vm u1..un`, where `c` is a coercion and `t` is not an application, is to prefer the notation which does not need a delimiter, and the notation for the second if both are on a par for this criterion. The original motivation was for numeric notations, so that e.g. `Zpos xH` is printed `1` both  if `Z_scope` comes first or if `positive_scope` comes first.

This is rather ad hoc and a revised proposal could be to instead align to the general criterion we are currently discussing, i.e. to give priority to the longest match, then to open scopes (or promoted scopes if such notion is introduced), then to the most recently declared notation in the given scope (leaving pending the inclusion of other criteria yet to be discussed, e.g. like being more specific).

herbelin:
> Actually that syntax is defined in the ML plugin code, because of the intricacies of dual-purposing an `as`clause in the pattern (only the boolean `if` is defined in `ssreflect.v`, because that is not an issue). The bit of notation in the prelude does not address this (nor the `in` and `return` clauses either, for that matter), and also does not support the `isn't` variant. 

I did not propose the `isn't` myself because I was uncertain it was really worth (there was a working group where `isn't` was discussed but noone has been pushing strongly for it, iirc). The clauses `as` and `return` are not proposed either in the prelude but, as far as I could test, they are working for whoever wants to define them user-side. (The `in` clause and the implicit `as` when the term is a variable are not working though.)

Personally, I find the `if is` idea cool and useful and this is why I proposed it in the standard library. Congratulations for it. I'm a bit confused about what to do though. On one side, you seem very attached to your code, which I can understand, but on the other side, I guess that you would agree that it makes little sense to have code duplication, so why not to rather try to converge on a common implementation which satisfies all of us? Regarding `in` and the dual purpose of an `if` on a variable, I agree that it would be cool for completeness to add support for them too but, am I right that at the current time there is anyway no occurrences of `if is` or `if isn't` with the `in` clause in MathComp and no occurrences either of an implicit `as` clause?

> More work would be needed, including extensions for notation patterns that seem rather too specialised. Perhaps it would be preferable to provide hooks to allow plugins to flag their `match`es and provide their own externing code for them?

I'm sorry, I don't understand. Which extensions are your thinking about? Otherwise, regarding "open-ended" (plugin-extensible) externing code, this is a recurrent question in Coq development discussions but nothing concrete yet as far as I know.

herbelin:
> Although the computation of default arguments could indeed look for known coercions to `Funclass` to populate extra args, this does not cover all possibilities - for instance there could be coercions to a type convertible to a function, but does not coerce to `Funclass` (such as the `ssreflect` `pred T`), ...

I don't see at once why the `pred T` case would not be covered by what I was proposing. Would you mind explaining into more details?

> ... the extra arity could come not from a coercion but from an expected value of a parameter on which the return type depends (think n-ary functions), ...

In which sense are you considering that extra scopes in `Arguments` solves the n-ary function case? One can only provide a finite number of extra arguments but to cover the n-ary function case one would typically need to provide a recursive pattern of extra arguments.

> ... and the correct scope and implicitness may depend on the specific constant at hand rather than the coercion itself, or may be different from the `Bind` defaults. For example MathComp sometimes put finite set arguments  of group operations in the finite group scope because they sometimes are the result of (lifted) group operations.

Giving the possibility to override the argument scopes of the coercion in `Arguments` for this specific constant is fine to me. What I find difficult to explain in the current design is to give extra arguments to a constant and for these extra arguments to make sense one typically relies on an implicit remote invariant that the constant shall be combined with a particular coercion. Why not to make this dependency explicit?

herbelin:
> It is - `image` is an abbreviation, and the comprehension syntax is notation for its specialisation to an explicit lambda, so this can be partly explained by the new priorities. What got me confused is that there must also be new code that elides eta-expansions on display since otherwise we should not have seen `image f A` but `image (fun x => f x) A`, or rather `image [eta f] A` since MathComp has notation for explicit etas.

It is the other way round. When the printer expects a function and the term it gets is not syntactically a function, it eta-expands it on the fly so as to be able to apply a notation expecting a function. So, most probably, the `f` here was in eta-reduced form from the beginning.

herbelin:
> Your enthusiasm is much appreciated, but please don't slave over this during holidays!

I'd be glad to have users satisfied and my January is anyway already pretty busy, so next week seems to be a good time (unless the discussion did not converge enough to proceed).

What we can do is also to share the work. For instance, you may work at implementing your proposal of interleaving the elision of a chain of coercions with search for notations, if ok to you.

> Thus I could give a Notation specialising a previous one the same printing scope, so that it would be preferred if applicable (perhaps "precedence" would be a better term than "scope", here). Lone notations and abbreviations could be given printing scopes without having to implement overloaded name tab lookup.

It starts to be late and I'll concentrate on that later. Or, maybe, could you give a concrete example? (In particular, is it only about the `x < y < z` issue or also about the `%MS` kind of question?)

Wishing you a good night at the current time.

ggonthier:
> What we can do is also to share the work. For instance, you may work at implementing your proposal of interleaving the elision of a chain of coercions with search for notations, if ok to you.

It would be, but I'll be away myself two weeks for the holidays, so not anytime soon...

> It starts to be late and I'll concentrate on that later. Or, maybe, could you give a concrete example? (In particular, is it only about the `x < y < z` issue or also about the `%MS`  kind of question?)

Both, actually, though more acute for `%MS` notations because that scope is never opened. After sleeping on the matter I'm having some doubts that this will be an appropriate remedy. The rationale for my suggestion was this: Notations link the user term language to the kernel term language, and are used to translate in both directions. Users can say which notations can be used by selective `Import` (with an important caveat for qualified abbreviations, this is actually an old complaint of mine), where modules are used to group notations. Users can also amend the set of notations used for input by means of the scope stack, and, with that recent PR, this also affects display of system output; for this purpose notations are grouped by scopes. I fear this cannot work well because the conflicts for input and output are very different, so that we should not assume that the same groupings will work for both - hence the suggestion to allow for different grouping.

Conflicts for output are very different from conflicts for input because notation link a single grammar rule to multiple instances of a kernel term pattern - no overlap is possible on the LHS, while there may be arbitrary overlap on the RHSs. Indeed, all of the output conflicts I can think of in MathComp are specialisations: the head constant of the pattern of one notation happens to have some notation, often one that is used in the definition of the pattern. In this was giving precedence to the head constant notation amount to implicitly making the more specialised one `parsing only`, and this is silly (if the user wants that he would have said so). The "order of import" priority actually gets this right, because one can expect the notation used in the RHS of a more specialised notation to have been imported first.
I can imagine one could have output conflicts due to variants of notations for the same concept, e.g., ssreflect sequences versus coq lists. Is this the use case you had in mind for the new disambiguation strategy? In that case I would suggest using a separate scope stack, and ignoring the arguments scope - these are type-based and thus should be irrelevant here.

herbelin:
> It would be, but I'll be away myself two weeks for the holidays, so not anytime soon...

Great, and no problem. There is no urgency.

> I can imagine one could have output conflicts due to variants of notations for the same concept, e.g., ssreflect sequences versus coq lists. Is this the use case you had in mind for the new disambiguation strategy?

This originally came from a demand for Mtac2 which is using indeed `list` in different contexts with different expected rendering. I don't remember if the original request was somewhere on github. There is a related example at #873.

> Users can say which notations can be used by selective Import , where modules are used to group notations.

Grouping notations in module and using selective imports looks more constraining than grouping 
with scopes and using selective scope opening. Also, adding objects to a module in several steps is still not implemented.

> (with an important caveat for qualified abbreviations, this is actually an old complaint of mine)

What do you mean?

> In that case I would suggest using a separate scope stack, and ignoring the arguments scope - these are type-based and thus should be irrelevant here.

I'm unsure of what you have in mind. Which user API would come with this? Also, even if disambiguation for printing is on the RHS, it has to take into account how the LHS would be disambiguated at parsing time to know if a delimiter is needed.

Back to what I proposed above (ability to bind more than one scope to an argument, ability to "promote" a scope in the stack without opening it), would it be otherwise ok for your needs?

@ejgallego:
> The %1 proposal looks pretty good to me.

Wouldn't glueing a `1` with the name of the scope, as in `(x + y)%1nat` be more difficult to human-parse than if using a proper symbol?

ejgallego:
This got stuck I am not aware exactly when [as discussion also happened offline maybe?]

So @herbelin @ggonthier I guess given that the branching date is tomorrow @vbgl will postpone this to next release (which I also support as assignee) unless you consider it critical for 8.10.

I'd be great if we could go however make some progress on this at least for early inclusion in 8.11.

ejgallego:
Tentatively (re) setting the milestone just to help the RM, but dont' hesitate to complain if you think it is not right.

ggonthier:
My understanding is that this got put on hold because it was intended as a solution for #9180, but after further review it turned out not to be. It is still a worthwhile improvement to the notation selection logic, in particular it provides a nice solution to providing explicit notations to applied coercions, but it's not essential to have it in 8.10.

ejgallego:
I'm unsure if @herbelin is planning to revisit this as per comment in other PR; thus removing from my list of to-merge PRs for now; please ping me if you want to merge.

Zimmi48:

PR #9214, opened 2018-12-12
Notations: Removing useless parentheses on abbreviations shortening a strict prefix of an application

herbelin:
**Kind:** bug fix

We avoid useless parentheses when using abbreviations to shorten the prefix of an application. Example:
```
Definition myAnd A B := A /\ B.
Notation myAnd1 A := (myAnd A).
Check myAnd1 True True. (* used to print "(myAnd1 True) True" *)
```

Thanks to @ggonthier for reporting.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,6 +49,9 @@ Notations
 - New command `String Notation` to register string syntax for custom
   inductive types.
 
+- Various bugs have been fixed (e.g. PR #9214 on removing spurious
+  parentheses on abbreviations shortening a strict prefix of an application).
+
 Plugins
 
 - The quote plugin (https://coq.inria.fr/distrib/V8.8.1/refman/proof-engine/detailed-tactic-examples.html#quote)
--- a/interp/constrextern.ml
+++ b/interp/constrextern.ml
@@ -625,8 +625,13 @@ let explicitize inctx impl (cf,f) args =
 	  CApp ((ip,f),args1@args2)
     | None ->
       let args = exprec 1 (args,impl) in
-	if List.is_empty args then f.CAst.v else CApp ((None, f), args)
-  in
+        if List.is_empty args then f.CAst.v else
+          match f.CAst.v with
+          | CApp (g,args') ->
+              (* may happen with notations for a prefix of an n-ary
+                 application *)
+              CApp (g,args'@args)
+          | _ -> CApp ((None, f), args) in
     try expl ()
     with Expl -> 
       let f',us = match f with { CAst.v = CRef (f,us) } -> f,us | _ -> assert false in
--- a/test-suite/output/Notations4.out
+++ b/test-suite/output/Notations4.out
@@ -45,5 +45,9 @@ fun x : nat => (x.-1)%pred
      : Prop
 ##
      : Prop
+myAnd1 True True
+     : Prop
+r 2 3
+     : Prop
 Notation Cn := Foo.FooCn
 Expands to: Notation Top.J.Mfoo.Foo.Bar.Cn
--- a/test-suite/output/Notations4.v
+++ b/test-suite/output/Notations4.v
@@ -165,6 +165,22 @@ Check ##.
 
 End H.
 
+(* Fixing bugs reported by G. Gonthier in #9207 *)
+
+Module I.
+
+Definition myAnd A B := A /\ B.
+Notation myAnd1 A := (myAnd A).
+Check myAnd1 True True.
+
+Set Warnings "-auto-template".
+
+Record Pnat := {inPnat :> nat -> Prop}.
+Axiom r : nat -> Pnat.
+Check r 2 3.
+
+End I.
+
 (* Fixing a bug reported by G. Gonthier in #9207 *)
 
 Module J.


herbelin:
Hi, I'm a bit confused. How do we send pull requests to Iris? Do we need to register for an mpi-sws account? Thanks in advance for any help.

ppedrot:
@herbelin I guess that a simple way is to ping @RalfJung to set up your PR. (Note that I happen to still have an account there, but I don't want to mess with their process.)

RalfJung:
See https://gitlab.mpi-sws.org/FP/iris-coq/blob/master/CONTRIBUTING.md for how to contribute to Iris. @herbelin if you already have an account at https://gitlab.mpi-sws.org/, please let me know your handle so I can unlock you for creating forks.

herbelin:
Rebased + overlay submitted upstream. This shall need a synchronisation between the assignee and the Iris maintainers. @ejgallego, maybe would you be ok to assign?

ejgallego:
> @ejgallego, maybe would you be ok to assign?

Sure @herbelin ; it seems a few places need fixing, will wait for the tweaks and review/merge ASAP.

herbelin:
> it seems a few places need fixing

Was just a collision with the new template warnings. Now deactivated it.

ejgallego:
@herbelin for some reason the overlay failed, is your branch properly set?

herbelin:
> @herbelin for some reason the overlay failed, is your branch properly set?

Probably not. The CI file for Iris seems to force a specific `Iris_CI_REF` (line 12 of `ci-iris-lambda-rust.sh`) which overrides the name of the branch I gave in the userlay. I don't know what to do.

RalfJung:
I updated Iris and LambdaRust to not test Coq 8.10 output.  I think that should fix CI for this PR.

ejgallego:
> I updated Iris and LambdaRust to not test Coq 8.10 output. I think that should fix CI for this PR.

@RalfJung I am not sure this is the best thing to do. The Iris/LR test suite is very useful in order to track notation / printing regressions.

Wouldn't it be possible to add a few version-specific tests?

The idea for 8.10 is to make some notation changes [for the better] so IMHO we should allow them but on the other hand they should happen in a controlled way.

RalfJung:
> Wouldn't it be possible to add a few version-specific tests?

Do you have any proposal for how the reference files should be updated when adding a new test?  The only scheme I can think of would require us (Iris devs) to  run `make REF=1` with two different Coq versions (including recompiling Iris in between because Coq's build system stores the `.vo` files always in the same place)

ejgallego:
> Do you have any proposal for how the reference files should be updated when adding a new test? The only scheme I can think of would require us (Iris devs) to run `make REF=1` with two different Coq versions (including recompiling Iris in between because Coq's build system stores the `.vo` files always in the same place)

That's indeed a problem :( maybe the 8.10 tests should be ignored on IRIS CI, but fatal on Coq's CI?

ejgallego:
That wouldn't be so perfect unfortunately as it only covers test creation. Ummm, let's think for a bit.

ejgallego:
Let's think a bit about this, but IMHO we should devise a scheme that allows us to improve notations / printing while keeping the current coverage.

ejgallego:
@herbelin please rebase as it seems that's the only way to test the IRIS setup.

herbelin:
I rebased but should I removed the iris overlay? It seems that CI is looking for a specific commit hash in my branch of iris rather than the head of my iris branch.

RalfJung:
Is the Iris overlay the only thing blocking this?  If yes, we can coordinate to move it forward.  I'll temporarily disable notation checking for Coq 8.10 so that @herbelin's PR fixing the ref file can be merged, then CI here should pass, then this can get merged and then I can enable notation checking again on the Iris side.

herbelin:
OK, I rebased w/o the Iris overlay.

RalfJung:
I merged your PR on the Iris side, and disabled printing testing so that our CI remains green.  That means CI here should work now.

ejgallego:
Great, thanks a lot folks; your help is really appreciated and in particular Iris test-suite is of great use to Coq.

I am going to merge this then, I hope it doesn't create too much pain and we can keep evolving to a model that is painless for everybody.

ejgallego (approved):
LGTM. 

@herbelin , will merge this tomorrow, but a small question before we merge: it seems to me that this is an user-facing change, so maybe it would merit an entry in `CHANGES` under "Printing / Notations"?

RalfJung:
> will merge this tomorrow

Tomorrow has passed. What is blocking this?
(I'd like to merge https://gitlab.mpi-sws.org/FP/iris-coq/merge_requests/208 on our side which is blocked on this.)

herbelin:
> it seems to me that this is an user-facing change, so maybe it would merit an entry in CHANGES under "Printing / Notations"?

I could add a note but this a fix for a bug with only rare occurrences. That would be somehow strange to do it for this fix and not for all bug fixes. As a compromise, knowing that "we" do not report all bug fixes, I added a (non-exclusive) generic line mentioning in particular #9214.

RalfJung:
Aaand the notation tests in Iris are passing and enabled again.  Thanks all!

PR #9215, opened 2018-12-12
Set up CI with Azure Pipelines

SkySkimmer:
This seems to work except for a couple tests.

IMO appveyor is worse than useless due to how often it times out so I'm immediately dropping it.


SkySkimmer:
Not sure what exactly I did that got it to start but now it's running for this PR.

SkySkimmer:
So it's failing with
~~~
logs/coq-makefile/timing.log
+ cd ./001-correct-diff-sorting-order
+ /cygdrive/d/a/1/s/tools/make-both-time-files.py time-of-build-after.log.in time-of-build-before.log.in time-of-build-both.log
C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\python.exe: can't open file '/cygdrive/d/a/1/s/tools/make-both-time-files.py': [Errno 2] No such file or directory
~~~
and
~~~
logs/tools/update-compat.log
  File "dev/tools/update-compat.py", line 340, in <module>
    update_compat_notations(known_versions, new_versions, **args)
  File "dev/tools/update-compat.py", line 289, in update_compat_notations
    with open(os.path.join(root, fname), 'r') as f: contents = …:00:21.0065064Z   File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 1106: character maps to <undefined>
~~~

ejgallego:
> IMO appveyor is worse than useless due to how often it times out so I'm immediately dropping it.

appveyor was working fine until the recent test-suite PRs that made it too expensive to run. I will fix it shortly, so I suggest we treat its removal separately.

In particular, it is still to see what the performance and time limits of Azure are. At least with appveyor we have a clear idea of what it is offered.

SkySkimmer:
What PR is that?

ejgallego:
#8655 , anyways I have a PR that should improve things for AppVeyor close to be ready, I hope I can finish it before the WE.

SkySkimmer:
Appveyor has been unreliable for a while, that PR can't be the cause.

ejgallego:
> Appveyor has been unreliable for a while, that PR can't be the cause.

This PR put it from the occassional overtime to 100% overtime.

ejgallego:
The thing is that IMO @SkySkimmer we should add Azure, but no way we can remove appveyor before Azure is battle-tested for good. I've submitted #9217 which should ease the pain. I am confident we can put appveyor back on track.

Zimmi48:
>appveyor was working fine

AppVeyor was crap from the beginning. But sure, the AppVeyor removal can be postponed to a later PR. Let's just not wait for several months like we did for CircleCI after adding GitLab CI support.

Zimmi48:
For many reasons and I would have a hard time doing a comprehensive summary about my experience with AppVeyor so I'll just list a few things:
- no parallelism and what looks like the slowest workers of every CI platform we tested, making it always the last to terminate;
- no clear information that the build reached a time out instead of a normal failure;
- very hard to understand the configuration settings, how to give admin permission to a user, poor integration with GitHub in this respect.

ejgallego:
> * no parallelism and what looks like the slowest workers of every CI platform we tested, making it always the last to terminate;

Oh, workers are not so slow, in fact they have two cores [but we only use one, that'll be fixed in the PR]; in fact hardware seems way more powerful than the one that Travis for example uses, it is Windows which is _very slow_. So that's a bit unfair to blame on them.

On the other aspects I do agree, but for the config settings, indeed, we are talking about a tool meant for the windows world, things are very different there I'm afraid.

SkySkimmer:
Not sure how to fix. It looks like it wants to use the system installed python but that doesn't work out. I'm going to try the naive solution of adding `-P python` to cygwin.

MSoegtropIMC:
> Not sure how to fix. It looks like it wants to use the system installed python but that doesn't work out. I'm going to try the naive solution of adding -P python to cygwin.

I am busy today finishing something, so I didn't have a look. But what I would recommend is to delete pretty much the complete PATH inherited from Windows. Otherwise you don't have a reproducible environment. See [configure_profile.sh](https://github.com/coq/coq/blob/master/dev/build/windows/configure_profile.sh)
I call this script right after installing cygwin (from [MakeCoq_MinGW.bat](https://github.com/coq/coq/blob/master/dev/build/windows/MakeCoq_MinGW.bat)). Of cause you can do the same in your main script - I just do it in the profile because I sometimes use such cgwins for interactive development. You should also have a look at the notes on ORIGINAL_PATH in configure_profile.sh.

SkySkimmer:
BTW Azure also has OSX runners, I don't know how they compare with the travis ones.

SkySkimmer:
>I am busy today finishing something, so I didn't have a look. But what I would recommend is to delete pretty much the complete PATH inherited from Windows. Otherwise you don't have a reproducible environment. See configure_profile.sh

Thanks, it works now.

SkySkimmer:
Who should be codeowner on this stuff? ci-maintainers?

ejgallego:
I am OK with this.

I am leaning toward adding this so we start to battle-test azure; please if you have a problem with Azure comment ASAP.

ejgallego:
> Who should be codeowner on this stuff? ci-maintainers?

I am OK with this.

.gitattributes
```diff
@@ -53,3 +53,6 @@ tools/CoqMakefile.in whitespace=blank-at-eol
 
 # CR is desired for these Windows files.
 *.bat whitespace=cr-at-eol,blank-at-eol,tab-in-indent
+
+* eol=lf
+*.bat eol=crlf
```
ejgallego: :mask:

Zimmi48:
>34m 47s

I don't know how you did but that's quite an impressive speed up compared to AppVeyor.

ejgallego:
> I don't know how you did but that's quite an impressive speed up compared to AppVeyor.

Appveyor is using one core, this uses 2; as I pointed out the appveyor update will fix this by using reliable detection of cores, but we can also do like this and see https://github.com/coq/coq/pull/9239

.gitattributes
SkySkimmer: ??

SkySkimmer:
Should I add a make validate while I'm at it?

ejgallego:
> Should I add a make validate while I'm at it?

IMHO not sure how useful it'd be; I'd say not necessary but feel free to add it.

ejgallego:
> Appveyor is using one core, this uses 2; as I pointed out the appveyor update will fix this by using reliable detection of cores, but we can also do like this and see #9239

So indeed with 2 cores appveyor takes 47 mins, https://ci.appveyor.com/project/coq/coq/builds/21065378

I dunno how much difference is in the setup time? But indeed, azure performs better.

ejgallego (approved):

Zimmi48 (commented):
`dev/build/windows/MakeCoq_MinGW.bat` has whitespace-only changes (end of line characters I think). Is that on purpose?

ejgallego:
> `dev/build/windows/MakeCoq_MinGW.bat` has whitespace-only changes (end of line characters I think). Is that on purpose?

It is [see 2nd commit] Whether that's the right thing to do I have no idea.

SkySkimmer:
For some reason without the `* eol=lf` in azure bash scripts (eg ./configure) die with "unknown command \r".

ejgallego:
Well so that's ready, maybe we should merge after the WG?

Are we then sure that Azure is better than appveyor? [I don't know]

IMHO we should at least wait a month to see we don't overdo quotas, etc...

ejgallego:
What does this service can do in addition of the build; could it also be used to generate the Win images?

SkySkimmer:
>Well so that's ready, maybe we should merge after the WG?

Why wait? ;)
I think just the increase in runner counts should make it better than appveyor.

>What does this service can do in addition of the build; could it also be used to generate the Win images?

There seems to be a concept of artifact to I guess yes.

ejgallego:
> Why wait? ;)

Ok let's go!

But I think you should mention it, this potentially impacts the whole project.

ejgallego:
> But I think you should mention it

mention it in the WG

MSoegtropIMC:
A note on line endings in Windows batch files: it must be CR-LF. See https://github.com/coq/coq/issues/8610. One doesn't want to have such bizarre issues in CI.

PR #9217, opened 2018-12-13
[ci] [appveyor] Temporally disable test suite on Appveyor.

ejgallego:
Hopefully we will re-enable it back soon, I am preparing a refactoring
that should make it more robust w.r.t paths and changes on Windows
which will enable to use a faster build system.

But now it is timing out 100% of the times [due to #8655] so it is not
useful.

### DIFF
--- a/dev/ci/appveyor.sh
+++ b/dev/ci/appveyor.sh
@@ -13,4 +13,4 @@ eval "$(opam env)"
 opam install -y num ocamlfind ounit
 
 # Full regular Coq Build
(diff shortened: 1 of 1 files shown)


Zimmi48:
If the test-suite is not run, then there is no point of running AppVeyor at all. Can we instead select a subset of tests to run? Or disable the new checker run from #8655?

SkySkimmer (approved):
Given that appveyor currently fails 100% of the time this can't make things worse, so I'm immediately merging and refinement can happen in further PRs.

PR #9220, opened 2018-12-13
Move shallow state logic to the function preparing state for workers

maximedenes:
This patch turns the three-valued `marshallable` type into a boolean. The previous `Shallow` value is now handled directly by the code preparing the state for workers, rather than by clients of the `Summary` API.

The type `marshallable` was used for several different purposes (see the mapping from `Yes to `No ...), so I hope this makes the code clearer.

We may be able to remove the `marshallable` flag altogether, at some point.


gares:
> We may be able to remove the `marshallable` flag altogether, at some point.

Agreed. The introduction of the flag was before the ephemeron mechanism was added.
This latter mechanism supersedes the `~marshallable` flag (after this PR).

gares (approved):

gares:
I'm afraid you'll have to prepare an overlay for any plugin that registers a piece of state.
And do it again to remove marshall... maybe you should do both ;-)

maximedenes:
> I'm afraid you'll have to prepare an overlay for any plugin that registers a piece of state.
> And do it again to remove marshall... maybe you should do both ;-)

Hmm, I did so already. It broke only paramcoq, in very exotic piece of code (the comment just above reads "I have no idea what I'm doing").

This is ready for merge, modulo maybe AppVeyor.

ejgallego (approved):

maximedenes:
@gares this is ready for merge

stm/stm.ml
```diff
@@ -2133,14 +2136,14 @@ end (* }}} *)
...
-    doc:doc -> ?redefine_qed:bool -> cache:Summary.marshallable ->
+    doc:doc -> ?redefine_qed:bool -> cache:bool ->
       Stateid.t -> unit
 
 end = struct (* {{{ *)
 
 let async_policy () =
-  if Attributes.is_universe_polymorphism () then false
-  else if VCS.is_interactive () = `Yes then
+  if Attributes.is_universe_polymorphism () then false (* FIXME this makes no sense, it is the default value of the attribute *)
```
SkySkimmer: @maximedenes what did you mean with this comment?

Issue #9093, opened 2018-11-27
Test suite does not pass with `-async-proofs on`


maximedenes:
Running the test-suite with `-async-proofs on` (on top of #9072, otherwise proof delegation never occurs) reveals a few problems:

- success/polymorphism.v
- output-modulo-time/ltacprof_cutoff.v
- bugs/closed/bug_4306.v
- bugs/closed/bug_3612.v

I keep them here for reference. Will investigate a bit more, and probably call for help, especially on the two that triggers universe-related errors...

SkySkimmer:
>output-modulo-time/ltacprof_cutoff.v

I doubt it makes sense to expect this one to work. In fact how come the output tests work, shouldn't the order get messed up?

ejgallego:
Side effects are ordered, unless they happen inside opaque proofs.

ejgallego:
In fact likely an opaque proof with side effects is likely to be marked as synch.

SkySkimmer:
The issue in polymorphism.v can be reproduced with just
~~~coq
Unset Universe Polymorphism.

Ltac exact0 := let x := constr:(Type) in exact 0.

Lemma lemma_restrict_abstract@{} : nat.
Proof. abstract exact0. Qed.
(* Debug: 10237:proofworker:0:0 STM: sending back a fat state
Error: Universe {polymorphism.1} is unbound. *)
~~~

maximedenes:
> > output-modulo-time/ltacprof_cutoff.v
> 
> I doubt it makes sense to expect this one to work. In fact how come the output tests work, shouldn't the order get messed up?

Yes, this one is more of a design issue. What seems trickier is the universe-related ones.

ejgallego:
We have requested many times for info about the design of universe handling, it would be good if we could understand more !

SkySkimmer:
#9222 fixes bug_3612.v
#9223 fixes polymorphism.v and bug_4306.v

maximedenes:
Wow! You should make us pay for such an efficient hotline ;) Thanks!

gares:
What is left then? IIRC the plan was to simply run ltacprof in sync mode

maximedenes:
> What is left then? IIRC the plan was to simply run ltacprof in sync mode

How should I implement that?

gares:
many test-suite subfolders support `(* coqopts: bla bla *)`, I've recently added it to `ide/`, see the Makefile in the test suite

maximedenes:
[...]

PR #9222, opened 2018-12-14
Fix classification of Set Default Proof Mode.

SkySkimmer:
See #9093.


gares:
Hum, the fix is clearly OK, but I'd like to know what was wrong.

maximedenes:
> Hum, the fix is clearly OK, but I'd like to know what was wrong.

Ok, but can we merge the fix meanwhile? This is holding #9088.

ppedrot (commented):
STM trickery, out of my league.

PR #9223, opened 2018-12-14
Fix universe restriction in delayed mode.

SkySkimmer:
See #9093.

### DIFF
--- a/proofs/proof_global.ml
+++ b/proofs/proof_global.ml
@@ -339,6 +339,7 @@ let close_proof ~opaque ~keep_body_ucst_separate ?feedback_id ~now
     Proof.in_proof proof (fun m -> Evd.existential_opt_value0 m k) in
   let nf = UnivSubst.nf_evars_and_universes_opt_subst subst_evar
     (UState.subst universes) in
+
   let make_body =
     if poly || now then
       let make_body t (c, eff) =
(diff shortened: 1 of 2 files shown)


SkySkimmer:
Fix polymorphism.v and bug_4306.v in async mode.

SkySkimmer:
There may be a way to test the fix without using abstract or funind but I'm not going to bother looking for it.

gares (changes requested):
Thanks you so much for the fix.
I'm asking for a commit message that explains what/how things are fixed, or why the old code was broken.
The code already contains a comment, maybe you could amend it.

SkySkimmer:
I changed the commit message a bit but there isn't really anything to explain IMO. It's just a thing that we have to do and didn't before.

gares (approved):

gares:
Ok for the message, but I m not competent for the change itself

maximedenes:
The lack of global documentation of the universe API is making it very hard to find reviewers for such changes. Maybe @ppedrot could have a look?

ppedrot (commented):
This code is cursed. Those few lines of codes are always the epicenter of universe anomalies, and even when I believe that the current version is correct we always end up finding an issue. Thence I lost any hope of getting in right, and this is why as long as this seems to fix the bug and doesn't break any other test, this PR should be considered as correct.

gares:
Looking more at the code, I see various branches in which one takes the term, its type, computes the universes actually used and does the restrict/minimize mambo jumbo. Could we try to factor this out?

SkySkimmer:
I don't know, someone would have to try and see what happens.

gares:
Sorry I was not clear. I've the impression that there is literally repeated code
```
           let used_univs = Univ.LSet.union
              (Vars.universes_of_constr t)
              (Vars.universes_of_constr pt)
          in
          let univs = UState.restrict univs used_univs in
```
I see it https://github.com/coq/coq/pull/9223/files#diff-6c8b7174895e752db5efbccf6e900663R360 and https://github.com/coq/coq/pull/9223/files#diff-6c8b7174895e752db5efbccf6e900663R380 and https://github.com/coq/coq/pull/9223/files#diff-6c8b7174895e752db5efbccf6e900663R399 (roughly).

I believe the "invariant" is that in all but one branches you  want to collect all universes and restrict. The different branch seems to restrict the type and the body one at a time, I've no idea why, but it can stay like that.

If this sort of "invariant" is wrong, then there is no point in factoring and my argument is moot.
But here it seems you are fixing a bug caused by the fact that in one branch this operation was wrongly coded, so I'm suggesting to factor this operation in the 3 similar branches.

I'm not very knowledgeable, so feel free to ignore my comment.

ppedrot:
@gares Trying to sort out this cursed code doesn't preclude from merging the PR as-is, it's but a hotfix. That said, I really hope we can put a few more static invariants in there. I tried at a point with GADTs, but I gave up realizing I had no idea about what was supposed to happen, because of side-effects.

@SkySkimmer Is the CI failure spurious?

SkySkimmer:
> Is the CI failure spurious?

yes

ppedrot:
Except if @gares wants to refine the PR, I'll merge this soon.

gares:
go ahead

PR #9224, opened 2018-12-14
Move lint job to gitlab

SkySkimmer:
This changes the semantics a bit since we don't have
TRAVIS_COMMIT_RANGE anymore, instead we do per-commit linting for the
commits since the last merge commit.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -260,6 +260,18 @@ windows32:
   except:
     - /^pr-.*$/
 
+lint:
+  image: docker:git
+  stage: test
+  script:
+    - apk add bash
+    - dev/lint-repository.sh
+  dependencies: []
+  before_script: []
+  variables:
+    # we need an unknown amount of history for per-commit linting
+    GIT_DEPTH: ""
+
 pkg:opam:
   stage: test
   # OPAM will build out-of-tree so no point in importing artifacts
--- a/dev/lint-repository.sh
+++ b/dev/lint-repository.sh
@@ -4,33 +4,25 @@
 
 # lint-commits.sh seeks to prevent the worsening of already present
 # problems, such as tab indentation in ml files. lint-repository.sh
-# seeks to prevent the (re-)introduction of solved problems, such as
-# newlines at the end of .v files.
+# also seeks to prevent the (re-)introduction of solved problems, such
+# as newlines at the end of .v files.
 
 CODE=0
 
-if [ -n "${TRAVIS_PULL_REQUEST}" ] && [ "${TRAVIS_PULL_REQUEST}" != false ];
-then
-    # skip PRs from before the linter existed
-    if [ -z "$(git ls-tree --name-only "${TRAVIS_PULL_REQUEST_SHA}" dev/lint-commits.sh)" ];
-    then
-        1>&2 echo "Linting skipped: pull request older than the linter."
-        exit 0
-    fi
+# We assume that all merge commits are from the main branch
+# For Coq it is extremely rare for this assumption to be broken
+read -r base < <(git log -n 1 --merges --pretty='format:%H')
+head=$(git rev-parse HEAD)
 
-    # Some problems are too widespread to fix in one commit, but we
-    # can still check that they don't worsen.
-    CUR_HEAD=${TRAVIS_COMMIT_RANGE%%...*}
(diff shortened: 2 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -260,6 +260,18 @@ windows32:
   except:
     - /^pr-.*$/
 
+lint:
+  image: docker:git
```
ejgallego: Why to use a docker image here instead of the base one?

SkySkimmer: I don't remember how I picked it and don't want to search for images again.

ejgallego (approved):
I guess the approximation is good enough tho I am not sure why we cannot try to get the exact set of commits; I am afraid of losing time in the future due to confusion.

Also, the approximation may not be good enough in the future if we want to apply stricter coding rules.

Thus mergeable but worth thinking about it a bit more.

.gitlab-ci.yml
SkySkimmer: >I mean, why do you use an image at all?

I have no clue what happens if you don't use an image. Is it even possible to unset our toplevel `image: "$IMAGE"`?

ejgallego: Yeah, my comment was not very useful, your choice is fine.

ejgallego:
> Thus mergeable but worth thinking about it a bit more.

I will merge soon, but as this is not urgent / in risk of conflict I'd like to try to play with git a bit once a get a minute.

Zimmi48 (commented):
I approve the strategy of looking for a merge commit but it could be made even more robust, see inline comment.

dev/lint-repository.sh
```diff
@@ -4,33 +4,25 @@
...
-
-    # Some problems are too widespread to fix in one commit, but we
-    # can still check that they don't worsen.
-    CUR_HEAD=${TRAVIS_COMMIT_RANGE%%...*}
-    PR_HEAD=${TRAVIS_COMMIT_RANGE##*...}
-    MERGE_BASE=$(git merge-base "$CUR_HEAD" "$PR_HEAD")
-    dev/lint-commits.sh "$MERGE_BASE" "$PR_HEAD" || CODE=1
-fi
+# We assume that all merge commits are from the main branch
+# For Coq it is extremely rare for this assumption to be broken
```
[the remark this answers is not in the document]
SkySkimmer: I considered that but then it seemed like I should add something for backport branches. It also seems kinda fragile.
Plus I'm not sure it would even give the right commit when someone merges master in their branch.
In the end I prefer keeping it simple and just accept that some per-commit checks will be missed if people put a merge in their PR. IMO this is fine since we essentially never do that, when we do it's a pathological case like #8778 and the per-commit checks are just not that important.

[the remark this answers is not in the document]
ejgallego: It seems indeed fragile to me.

ejgallego (commented):
Will take me a while to process, so if urgent please anyone feel free to take over.

Wouldn't it make sense to keep both jobs for a while ?

SkySkimmer:
>Wouldn't it make sense to keep both jobs for a while ?

Pointless imo.

Zimmi48:
@SkySkimmer If we want, we can ask the bot to systematically create a merge commit by passing option `--no-ff` to the merge command in `make_ancestor.sh`. Wouldn't that be enough so that the linter script has all the information it needs?

SkySkimmer:
I also want the linter to run on my gitlab where pushes aren't from the bot though.

SkySkimmer:
Fundamentally I'm only aware of 1 PR with a merge commit (#8778 plugin tutorial import), so I don't see the point in making the script more complex to handle a never-happening case.

SkySkimmer:
Put another way, sure we could exploit the Coq specificity that we have standard messages for merges into mainline, or that we use a bot. But why do that instead of exploiting the specificity that PRs don't include merge commits?

ejgallego (approved):
LGTM

ejgallego:
@Zimmi48 are you OK with this?

Zimmi48:
>But why do that instead of exploiting the specificity that PRs don't include merge commits?

We could argue that we don't impose this convention as much as the PR naming scheme. Nonetheless, I'm quite happy with the PR as-is. It doesn't matter if there are a few PRs for which the linter is not run fully (i.e. we can leave with this under-approximation).

Zimmi48 (approved):

Issue #9212, opened 2018-12-12
Windows CI: cygwin install cache is not reused


MSoegtropIMC:
In the gitlab Windows CI the install cache for cygwin is not reused. This produces a lot of traffic.

Please note that the cache cannot result in an security issues, even if it is shared between runners. The cygwin setup checks the signature of the main index file and the main index file contains hashes for all modules. The only things we must make sure is not manipulated between runs is the cygwin setup program itself.

Does someone have preferences if we should put the setup program somewhere so that only admins can change it or if we should download it every time?

PR #9225, opened 2018-12-14
Fix issue: Windows CI: cygwin install cache is not reused #9212

MSoegtropIMC:
Use the same folder for the cygwin cache in all runs.

### DIFF
--- a/dev/ci/gitlab.bat
+++ b/dev/ci/gitlab.bat
@@ -26,12 +26,12 @@ if %ARCH% == 64 (
 
 SET CYGROOT=C:\ci\cygwin%ARCH%
 SET DESTCOQ=C:\ci\coq%ARCH%
+SET CYGCACHE=C:\ci\cache\cgwin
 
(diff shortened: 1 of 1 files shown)


MSoegtropIMC:
I just checked: with the new cygwin mirror downloading all setup files takes just 2 minutes, so it doesn't save that much in terms of CI runtime (with the old mirror it was more like 5 minutes).

But there are a few good arguments for caching:

- the setup files are required for reproduction of version XYZ of Coq and cygwin doesn't keep them. If one wants to debug a problem on an older version of Coq on Windows (might happen that this is requested by a consortium member some time) one needs these files. Keeping them on all runners should be sufficiently safe and one can also collect them from there and archive them after each release.

- caching might make the CI more reliable (less dependent on the mirror).

- we need less network traffic for INRIA and for the mirror

maximedenes (approved):
Not sure if this will be effective once we move to Azure, but I guess it can't hurt.

PR #9227, opened 2018-12-14
Provide scripts for CI runner maintenance, Issue #8622

MSoegtropIMC:
This PR just contributes a few scripts and documentation I am using for CI runner maintenance.

Not sure if this is the best imaginable place ...


dev/ci/tools/runner-maintanance/CiRunnerMaintenance_ConfigureCygwinProfile.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
SkySkimmer: Isn't this file the same as the dev/build/windows/configure_profile.sh?

MSoegtropIMC: Yes, the other file is also derived from there. But it serves a completely different purpose and I didn't think it makes a lot of sense to link them together. I am open to better ideas of handling this.

MSoegtropIMC: P.S.: the goal of this is to collect scripts and documentation to

- completely finish the setup of a runner from a fresh Windows VM
- efficiently maintain the operation of the runners

If someone has scripts and documentation to install and configure the gitlab runner elsewhere, this should go together.

SkySkimmer (commented):
Since the scripts live in a runner-maintanance (typo btw) directory why prefix them with CiRunnerMaintenance?

MSoegtropIMC:
> Since the scripts live in a runner-maintanance (typo btw) directory why prefix them with CiRunnerMaintenance?

Because they are intended to be copied to the runners and exist there outside of a GIT repo (in a folder /bin/tools) and then it is not so clear what these scripts are.

I will fix the typo (also I found that the login script has a wrong comment line in it).

ejgallego:
Hopefully we stop using INRIA unsecure runners in favor of Azure, thus this PR should not be needed.

MSoegtropIMC:
Azure is a good option - then this kind of maintenance work is done by others. I agree that this need not be merged - in case someone needs the scripts he/she can get them from the PR, so I will close the PR.

Btw.: as discussed via email I don't share your opinion on the security of INRIA runners.

PR #9231, opened 2018-12-16
Fixes #9229: Infix not robust wrt choice of variable names in right-hand side

### LINKED ISSUE #9229: Infix notations capture x and y.
#### Version
8.10+alpha
master (4ecbad30c77316294c8625ead722d469c1c7f79d)

```
Infix "#" := (fun x y => x + y)
  (at level 50, left associativity).
Check (3 # 5).
(*     ^          *)
(* Fails with: This expression should be a name. *)
```
This surprised me, I expected it to successfully check as `(fun x y => x + y) 3 5`.

Using `x` and `y` as lambda variables in an infix command seems to force them to be names.
I guess that this is literally translated as `Notation "x # y" := ((fun x y => x + y) x y).` where `x` and `y` have to be names to be used as binders. The [documentation](https://coq.inria.fr/refman/user-extensions/syntax-extensions.html#the-infix-command) gives this translation (`Infix "#" := f` becomes `Notation "x # y" := (f x y)`) but also says that `x` and `y` will be *fresh* names, so I was surprised by this behavior.

cc: @herbelin

herbelin:
**Kind:** bug fix

Fixes / closes #9229

Infix was not robust wrt to choice of variable names in the right-hand side, contrary to what was documented, Thanks to @jashug for noticing this fragility.

### DIFF
--- a/interp/constrexpr_ops.ml
+++ b/interp/constrexpr_ops.ml
@@ -366,6 +366,14 @@ let free_vars_of_constr_expr c =
     | c -> fold_constr_expr_with_binders (fun a l -> a::l) aux bdvars l c
   in aux [] Id.Set.empty c
 
+let names_of_constr_expr c =
+  let vars = ref Id.Set.empty in
+  let rec aux () () = function
+    | { CAst.v = CRef (qid, _) } when qualid_is_ident qid ->
+      let id = qualid_basename qid in vars := Id.Set.add id !vars
+    | c -> fold_constr_expr_with_binders (fun a () -> vars := Id.Set.add a !vars) aux () () c
+  in aux () () c; !vars
+
 let occur_var_constr_expr id c = Id.Set.mem id (free_vars_of_constr_expr c)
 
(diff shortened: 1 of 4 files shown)


interp/constrexpr_ops.ml
```diff
@@ -366,6 +366,14 @@ let free_vars_of_constr_expr c =
     | c -> fold_constr_expr_with_binders (fun a l -> a::l) aux bdvars l c
   in aux [] Id.Set.empty c
 
+let free_names_of_constr_expr c =
```
gares: Free is misleading here, it reminds me of free variables. The comment in the mli is right, you gather both free and bound variables, but also variables standing for global constants.

All_idents_of_... Maybe?

ppedrot: Just `names_of_constr_expr`?

herbelin: Indeed, this makes much more sense without "free"! Done.

ppedrot (approved):

Issue #9234, opened 2018-12-17
Windows v8.9: AAC_tactics has wrong require statements


MSoegtropIMC:
In the v8.9 branch there is a glitch in AAC_tactics. There used to be a patch to adjust import statements so that the Tutorial and other files would work out of the box (without adjusting Require commands). On December 6th this was corrected in AAC_tactics master:

https://github.com/coq-community/aac-tactics/commit/aead5cd61ec114d077d6128a87c908088292c36a#diff-19f9381828a2e9804938d9bab7e71f5e

and the patch in Coq was removed around the same time but now 8.9 has neither the patch nor the updated commit.

The easy fix is to update the version of AAC_tactics in ci-basic-overlay.sh

Zimmi48:
Fixed in #9235.

PR #9235, opened 2018-12-17
Change version of AAC_tactics plugin to include latest fixes for 8.9

MSoegtropIMC:
Fix Issue #9234 : AAC_tactics has wrong require statements in 8.9.

The master doesn't need this because it pulls in AAC_tactis master.

I decided to not take the latest version of AAC_tactics because the changes by @SkySkimmer there are labeled to target 8.10.

### DIFF
--- a/dev/ci/ci-basic-overlay.sh
+++ b/dev/ci/ci-basic-overlay.sh
@@ -282,7 +282,7 @@
 ########################################################################
 # aac-tactics
 ########################################################################
-# Latest commit on master as of Oct 8, 2018
-: "${aactactis_CI_REF:=c10636221af6324dfeea89830699dbb7488dfdde}"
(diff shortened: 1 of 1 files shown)


Zimmi48 (changes requested):
This commit is already one that is not compatible with Coq 8.9. Try using the tip of the `v8.9` branch instead: see https://github.com/coq-community/aac-tactics/commits/v8.9 (does it contain the patch you need).

Zimmi48:
I confirm that this branch has the commit you want so I suggest using commit c469b26e409b1bde6a64546df85226079796dbe7.

MSoegtropIMC:
@Zimmi48 : yes you are right - it was silly to not look into the v8.9 branch. Even simple changes shouldn't be done in a hurry ...

PR #9241, opened 2018-12-17
Fix #9240: Register for IDProp causes anomaly when non constant

### LINKED ISSUE #9240: Register for IDProp causes anomaly when nonconstant
~~~coq
Register unit as core.IDProp.type.
Register tt as core.IDProp.idProp.


Inductive vec (A : Type) : nat -> Type :=
| nil : vec A 0
| cons : forall n : nat, A -> vec A n -> vec A (S n).

Definition hd (A : Type) (n : nat) (v : vec A (S n)) : A :=
match v in (vec _ (S n)) return A with
| cons _ _ h _ => h
end. (* assertion failure in evarconv *)
~~~
Fix incoming.

SkySkimmer:


SkySkimmer:
TBH I'm not sure why we register IDProp instead of using the type of idprop.

SkySkimmer:
ping @mattam82 or @gares

gares (approved):

PR #9243, opened 2018-12-18
Fix line ending issues (azure related)

SkySkimmer:
Without this when one of these bat files is touched git will realise
that line ending conversion has not been done. Then the file will be
considered modified and it's a pain to get git to shut up about
it (you have to delete the eol= lines in gitattributes then run git
status, then you can undelete the lines).


SkySkimmer:
Looks like @jfehrle is hitting the same issue, so if we could accelerate this PR it would be nice.

MSoegtropIMC (approved):
I don't see a good reason to change the EOL of text files but don't have a strong opinion on this either. The batch files changes are good (hope this doesn't interfere to much with @ejgallego PR to remove many of these files).

MSoegtropIMC:
Sorry can't merge - see my post on coq-dev

SkySkimmer:
I'm going to try an alternate fix, should be saner.

.gitattributes
```diff
@@ -54,5 +54,5 @@ tools/CoqMakefile.in whitespace=blank-at-eol
 # CR is desired for these Windows files.
 *.bat whitespace=cr-at-eol,blank-at-eol,tab-in-indent
 
-* eol=lf
-*.bat eol=crlf
+*.sh eol=lf
+/configure eol=lf
```
SkySkimmer: When I saw that configure needed the eol=lf I was worried that I would have to list a bunch of files but this was unfounded. This way is definitely better.

SkySkimmer:
It almost works except for the update.py test which I'm not sure I can fix.
If this run doesn't work I'll disable it on azure and open an issue about having it support windows line endings.

MSoegtropIMC:
I am working on Windows since ever even using mixed MinGW and cygwin git on the same folder without much issues. I am using these settings:

For a MinGW / native Windows git:
```
git config --global -e
add:
[core]
        filemode = false
        autocrlf = false
```

For a Cygwin git:
```
git config --global -e
add:
[core]
        filemode = true
```

In both cases (as I said they work on the same folder) I use this local configuration:
```
git config --local -e
remove : 
[core]
        filemode = true
```

I am sure the same configuration can be achieved with CLI git commands (rather than calling an editor) or with .gitattributes, but I would need to find out how this would work.

I tend to use cygwin git for all CLI git operations but I am also using VSCode git integration and VSCode is easier to use with a separate MinGW git. So I am using both on the same folder and this works fine with these settings.

SkySkimmer:
@MSoegtropIMC I'm now trying to see if unsetting the text attribute on every file mimicks autocrlf=false properly.

JasonGross:
Why not just do
```
*.bat -text
```
?

Or am I misunderstanding the problem I've been called here to solve?  (The issue with update.py is that when python opens files in text mode I believe it uses `\n` to represent the system newline.  If you want to force it to deal with a different sort of line ending, the easiest way is probably to have it do its own line ending conversions when reading and writing files.  I can make it do this, but it seems strange to have the Coq repo override the user's preferences about line endings in all files.

SkySkimmer:
@JasonGross if we don't do any gitattributes it's not the .bat that fail, it's the .sh as bash dies on \r, then if we disable line ending translation for just the bash scripts the timing test and the update-compat test also die.

MSoegtropIMC (changes requested):
I would suggest to use

`* -text`

and commit batch files with CR-LF as they should be. On Linux this requires nothing special. On Windows this requires the settings I mentioned in the discussion. Otherwise those working on Windows end up in hell when they also use cygwin, because cygwin used LF line ending and Windows uses CR-LF line ending, so setting up git to do conversions is not really an option.

Regarding the python script issue: I guess you anyway have a cygwin, so just use a cygwin python. It won't produce CR.

SkySkimmer:
@Zimmi48 sf and dpdgraph had failures which disappeared after manual restart but not detected by the bot
https://gitlab.com/coq/coq/-/jobs/137261438 (dpdgraph, no clue what's going on there)
https://gitlab.com/coq/coq/-/jobs/137261469 (sf, network failure when fetching the tarball)

SkySkimmer:
>I would suggest to use `* -text` and commit batch files with CR-LF as they should be.

@MSoegtropIMC that's what I ended up doing. Looks like it works.

Zimmi48:
> @Zimmi48 sf and dpdgraph had failures which disappeared after manual restart but not detected by the bot
> https://gitlab.com/coq/coq/-/jobs/137261438 (dpdgraph, no clue what's going on there)
> https://gitlab.com/coq/coq/-/jobs/137261469 (sf, network failure when fetching the tarball)

The dpdgraph case seems to be a problem of non-determinism of their test-suite. I'd rather not address it through the bot (like our own non-deterministic tests).
The SF case I'd like to try to handle, but can't we add an option to `wget` so that network failures are made more explicit? I also add similar concerns with `wget` failures in Windows scripts.

SkySkimmer:
>what does this do:

That's for linting, we require that there's no space at the end of lines and without it CR counts as a space instead of being part of the line ending.

SkySkimmer:
To be clear the `whitespace` attribute causes no conversion to be applied. From `man gitattributes`:
~~~
   Checking whitespace errors
       whitespace
           The core.whitespace configuration variable allows you to define what diff and apply
           should consider whitespace errors for all paths in the project (See git-config(1)). This
           attribute gives you finer control per path.
~~~

MSoegtropIMC:
> To be clear the whitespace attribute causes no conversion to be applied

Thanks - I was about to ask :-)

MSoegtropIMC:
What I still don't get: what is the change to MakeCoq_MinGW.bat? As far as I can tell this file is CR-LF in master and CR-LF in the commit. Where does the diff display in gitub come from?

SkySkimmer:
>What I still don't get: what is the change to MakeCoq_MinGW.bat?

Because I did line conversion in a previous PR (#9215) it has to be undone now.

.gitattributes
```diff
@@ -54,5 +54,5 @@ tools/CoqMakefile.in whitespace=blank-at-eol
 # CR is desired for these Windows files.
 *.bat whitespace=cr-at-eol,blank-at-eol,tab-in-indent
 
-* eol=lf
-*.bat eol=crlf
```
SkySkimmer: This causes line conversion in master which is why you see mingw.bat as crlf.

MSoegtropIMC: I see. And because the .gitattribute in master also has *.bat eol=crlf I still see it as CR-LF when I do a pull - nasty. But I can see here:

https://raw.githubusercontent.com/coq/coq/master/dev/build/windows/MakeCoq_MinGW.bat

that in master it is indeed LF. So from my point of view the change is fine and I will approve.

MSoegtropIMC (approved):
This solution is fine and compatible with my Windows workflow.

MSoegtropIMC:
P.S.: merge might take a day (unless someone else merges) - I still need to debug this strange merge-pr.sh issue I have since today.

JasonGross:
@SkySkimmer 
> It almost works except for the update.py test which I'm not sure I can fix.
> If this run doesn't work I'll disable it on azure and open an issue about having it support windows line endings.

I'm confused.  According to https://docs.python.org/3/library/functions.html#open, 
> newline controls how universal newlines mode works (it only applies to text mode). It can be None, '', '\n', '\r', and '\r\n'. It works as follows:
>
> When reading input from the stream, if newline is None, universal newlines mode is enabled. Lines in the input can end in '\n', '\r', or '\r\n', and these are translated into '\n' before being returned to the caller. If it is '', universal newlines mode is enabled, but line endings are returned to the caller untranslated. If it has any of the other legal values, input lines are only terminated by the given string, and the line ending is returned to the caller untranslated.

This suggests that the update-compat test should work fine on all platforms.  I've made https://github.com/JasonGross/coq/tree/fix-bat-update which allows it to write Unix newlines on Windows, but that code isn't called in the test (which runs in read-only-checking-mode), so it shouldn't be needed.  Was Python2 different in handling line-endings on reading files?

SkySkimmer:
Please merge now. The python issues were only in previous versions of the PR.

Zimmi48:
>as they all need to be

They don't need to be anymore, I don't know if you have noticed (although of course it is still a good practice to rebase often).

PR #9245, opened 2018-12-18
[ssr] compile "=> {}id" as "=> {id}id"

gares:
This change lets one write `=> {}H` instead of `=> {H}H` in order to replace the context entry for `H`.
My main motivation for this patch is to make the idiom `have {H}H` a bit nicer: I was describing it in the MCB and I thought that the `{}` could be accepted there to avoid naming twice the hypothesis.

As of today `{}` is accepted in `rewrite {}H`, `move=> {}/H` with the meaning of "use and then discard". With this patch it is also accepted in `move=> {}H` with a meaning that is not exactly the same, but has some analogies (it more like "discard and then (re)use").

Do you think it is too confusing and not worth it? Do you think it is useful?
@ggonthier @CohenCyril @amahboubi

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -146,6 +146,7 @@ SSReflect
   - block introduction: => [^ prefix ] [^~ suffix ]
   - fast introduction: => >
   - tactics as views: => /ltac:mytac
+  - replace hypothesis: => {}H
   See the reference manual for the actual documentation.
 
 Changes from 8.8.2 to 8.9+beta1
--- a/doc/sphinx/proof-engine/ssreflect-proof-language.rst
+++ b/doc/sphinx/proof-engine/ssreflect-proof-language.rst
@@ -1559,7 +1559,7 @@ whose general syntax is
    i_view ::= {? %{%} } /@term %| /ltac:( @tactic )
 
 .. prodn::
-   i_pattern ::= @ident %| > %| _ %| ? %| * %| + %| {? @occ_switch } -> %| {? @occ_switch }<- %| [ {?| @i_item } ] %| - %| [: {+ @ident } ]
+   i_pattern ::= {? %{%} } @ident %| > %| _ %| ? %| * %| + %| {? @occ_switch } -> %| {? @occ_switch }<- %| [ {?| @i_item } ] %| - %| [: {+ @ident } ]
 
 .. prodn::
    i_block ::= [^ @ident ] %| [^~ @ident ] %| [^~ @num ]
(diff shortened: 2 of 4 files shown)


gares:
rebased on top of master since all dependencies were merged

Zimmi48 (approved):

gares:
ping @maximedenes @CohenCyril

maximedenes:
@ggonthier does the feature look good to you?

ggonthier:
I’m fine with the feature, but the documentation will have to point out that this only applies to the ‘{}’ intro pattern, contrary to rewrite, where the clear flag is tied to the following rule, and any clear flag (not just an empty one) attempts to clear the rewrite rule if it consists of a single identifier. Alternatively, we could try to implement this for intro patterns as well, adding a no-op ‘-‘ in cases where this causes a regression. Perhaps you could try that out? If there aren’t too many incompatibilities it would be better to have a more uniform interpretation. Also, does the clear apply to views preceding the pattern? Could you try that as well?

gares:
> any clear flag (not just an empty one) attempts to clear the rewrite rule if it consists of a single identifier.

I did not know that "not just an empty one" would have such an effect, and I cannot find it in the doc https://coq.inria.fr/distrib/current/refman/proof-engine/ssreflect-proof-language.html#an-extended-rewrite-tactic . Actually I cannot even find that `{}` has a meaning in rewrite, but I'm pretty sure it is written somewhere.

gares:
> Also, does the clear apply to views preceding the pattern? Could you try that as well?

Hum, I'm not sure I follow. `=> {}/v` compiles as `=> /v{v}` when `v` is an ident. `=> {H}/v` does not compile to `=> /v{H v}`. Is this you'd like me to try?

ggonthier:
You answered my second question in the second sentence; as for the first and third, yes, I think we should at least try to see if the regressions are infrequent enough that we can have the same semantics for rewrite and intro patterns. You’ve already implemented the more important change, which is to tie a clear switch to the pattern that follows it. Because inline clear switches are not that common, views are usually global constants, and introduced variables must be fresh, I would expect very few regressions.
   The extension of the clear to non-empty switches is used in the fairly common idiot for expanding a definition: rewrite {x}Dx with Dx : x = definition_of_x.

gares:
Imo this PR is ready to be merged, I've opened another one to make the behavior consistent with the one of rewrite (and also assessing its feasibility).

gares:
Ops, I've to ask @maximedenes to not merge this PR as it. The failures of #9341 shows that this implementation is too simplistic and does not scale.

gares:
closing in favor of #9341

Issue #9146, opened 2018-12-05
General issue about the `abstract` tactic.


ejgallego:
The `abstract` tactic is a notorious example of incomplete functionality in the Coq system. The tactic itself doesn't work in many contexts, and indeed this fact is not properly documented nor warned by the system.

I am opening this issue to continue and centralize discussion happening in other places such as #7905 and #9129 .

After a bit of thinking, and taking into account the global structure of the code, I think that a first proposal to start solving current problems with the current `abstract` tactic would be to change its type to that of a "super-tactic", and thus allow it only to appear at top-level positions.

Thus the only allowed forms would be:

```coq
abstract tactic.
```

This is for sure too radical but I don't have a better proposal as of today, and from a quick look it seems that would be doable without too much disruption.

We could still provide a tactic `abstract_tac` that would work in arbitrary positions but it would have different semantics and for example would take a few extra arguments (name of the abstracted term)

I finish the issue with commentary about a few strange idioms:
- `abstract (exact ltac:(abstract (exact A)))` What is this supposed to do?

ejgallego:
IMO we need to be a bit bold here, it doesn't seem sensible to me that we keep shipping `abstract` in their current state.

cc: @JasonGross as an abstract power user that for sure will be impacted. cc: @maximedenes @herbelin @ppedrot @vbgl @SkySkimmer @Zimmi48 @mattam82 , etc...

SkySkimmer:
>cc: @JasonGross as an abstract power user that for sure will be impacted. cc: @maximedenes @herbelin @ppedrot @vbgl @SkySkimmer @Zimmi48 @mattam82 , etc...

Do you want to @coq/some-team?

SkySkimmer:
Abstract is this side effect stuff right? I don't really understand how it works though. There's also scheme side effects (which I understand just as little).

JasonGross:
I think that `transparent_abstract` is also an important tactic.  It shares a lot of code with `abstract`, and is more tunable

ejgallego:
> Abstract is this side effect stuff right?

It does a bit more than that, but basically yes.

> I don't really understand how it works though.

In many cases, it doesn't.

> There's also scheme side effects (which I understand just as little).

These are a bit better as they are hardcoded, however abstract can appear in many strange places such as inside `ltac:()`, etc...

ejgallego:
> I think that `transparent_abstract` is also an important tactic. It shares a lot of code with `abstract`, and is more tunable

[...]

PR #9248, opened 2018-12-19
Fix #7904: update proofview env after ltac constr:()

### LINKED ISSUE #4781: [abstract] typeclass resolution hints give very strange behavior (and an abundance of Anomaly: Uncaught exception Not_found. Please report.) (regression since 8.4)

### LINKED ISSUE #7904: Strange error message "Error: Variable x should be bound to a term but is bound to a constr."
#### Version

8.8.0

I was taking another look at #4781, and discovered this:
```coq

Class abstract_term {T} (x : T) := by_abstract_term : T.
Hint Extern 0 (@ abstract_term ?T ?x) => force_clear; change T; abstract (exact x) : typeclass_instances.

Goal True.
  let term := constr:(I) in
  let T := type of term in
  let x := constr:((_ : abstract_term term) : T) in
  let x := match constr:(x) with ?y => y end in
  pose x. (* Error: Variable x should be bound to a term but is bound to a constr. *)
```

SkySkimmer:
(in case of side effects)

I hit the same code path in SProp (there it was a Not_found anomaly) so I need this fix. Hopefully someday we will have a more general solution for #9146 bugs though.

Fixes #4781.
Fixes #7904.

### DIFF
--- a/plugins/ltac/tacinterp.ml
+++ b/plugins/ltac/tacinterp.ml
@@ -1957,7 +1957,9 @@ let lifts f = (); fun ist x -> Ftactic.enter begin fun gl ->
   let sigma = Proofview.Goal.sigma gl in
   let (sigma, v) = f ist env sigma x in
   Proofview.tclTHEN (Proofview.Unsafe.tclEVARS sigma)
-  (Ftactic.return v)
+    (* FIXME once we don't need to catch side effects *)
+    (Proofview.tclTHEN (Proofview.Unsafe.tclSETENV (Global.env()))
+       (Ftactic.return v))
 end
 
 let interp_bindings' ist bl = Ftactic.return begin fun env sigma ->
--- a/test-suite/bugs/closed/bug_4781.v
+++ b/test-suite/bugs/closed/bug_4781.v
@@ -25,29 +25,29 @@ Goal True.
   let x := match constr:(Set) with ?y => constr:(y) end in
   pose x.
 (* This fails with an error: *)
-  Fail let term := constr:(I) in
+  let term := constr:(I) in
   let T := type of term in
   let x := constr:((_ : abstract_term term) : T) in
   let x := match constr:(x) with ?y => constr:(y) end in
   pose x. (* The command has indeed failed with message:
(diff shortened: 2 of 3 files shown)


plugins/ltac/tacinterp.ml
```diff
@@ -1957,7 +1957,8 @@ let lifts f = (); fun ist x -> Ftactic.enter begin fun gl ->
   let sigma = Proofview.Goal.sigma gl in
   let (sigma, v) = f ist env sigma x in
   Proofview.tclTHEN (Proofview.Unsafe.tclEVARS sigma)
-  (Ftactic.return v)
+    (Proofview.tclTHEN (Proofview.Unsafe.tclSETENV (Global.env()))
+       (Ftactic.return v))
```
ppedrot: Please add a FIXME comment here.

ppedrot (changes requested):

plugins/ltac/tacinterp.ml
SkySkimmer: Is there a need for a fixme? It seems like every tclSETENV will go away if we fix side effects, and won't if we don't.

ppedrot: `tclSETENV` is legit, the dreaded call is `tclSETENV + Global.env ()`.

SkySkimmer: OK, comment added.

ppedrot (approved):

ppedrot:
Should this be backported to 8.9?

PR #9250, opened 2018-12-19
coqchk: fix check for kelim with functors

SkySkimmer:


ppedrot (approved):
I am surprised that this is allowed by the kernel, but if so, the fix is fine.

SkySkimmer:
The kernel doesn't see it, it's the checker.

ppedrot:
Will wait a bit to let @maximedenes, @herbelin and @mattam82 to optionally comment.

mattam82 (commented):
I'm confused, this relaxes the checker check so that M.ind can be eliminated into Type?

SkySkimmer:
The checker sees M.ind after the functor application, so as far as it's concerned it can be eliminated into Type.

mattam82:
Ok. On the other hand the kernel does not see that and prevents such elimination, right?

SkySkimmer:
yes

Gaëtan Gilbert

herbelin:
> Will wait a bit to let @maximedenes, @herbelin and @mattam82 to optionally comment.

OK from my point of view.

This looks like dual to lowering a `Type` to `Prop` if a singleton type, i.e. unboxing a `Prop` to `Type` when eventually known to be a singleton. Ok to make it accepted by the kernel too if some interest to do so. Ok to make the `mind_kelim` a function dependent on the universes occurring in the type (or even dynamically recomputed) if it may help.

SkySkimmer:
@ppedrot can you merge?

PR #9260, opened 2018-12-20
Fix signals_are_pending linking error with ocaml trunk (4.08)

SkySkimmer:
`signals_are_pending` got renamed into `signals_might_be_pending` in ocaml/ocaml@1c82c481a5b33987e05012be61893c03f59ef361

### DIFF
--- a/dune
+++ b/dune
@@ -6,7 +6,8 @@
  (ireport (flags :standard -rectypes -w -9-27-40+60)
           (ocamlopt_flags :standard -O3 -unbox-closures -inlining-report))
  (ocaml408
-  (flags :standard -strict-sequence -strict-formats -short-paths -keep-locs -rectypes -w -9-27+40+60 -warn-error -5 -alert --deprecated)))
+  (flags :standard -strict-sequence -strict-formats -short-paths -keep-locs -rectypes -w -9-27+40+60 -warn-error -5 -alert --deprecated
+         -ccopt -Dcaml_signals_are_pending=caml_signals_might_be_pending)))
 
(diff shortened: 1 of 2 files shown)


dune
```diff
@@ -6,7 +6,8 @@
  (ireport (flags :standard -rectypes -w -9-27-40+60)
           (ocamlopt_flags :standard -O3 -unbox-closures -inlining-report))
  (ocaml408
-  (flags :standard -strict-sequence -strict-formats -short-paths -keep-locs -rectypes -w -9-27+40+60 -warn-error -5 -alert --deprecated)))
+  (flags :standard -strict-sequence -strict-formats -short-paths -keep-locs -rectypes -w -9-27+40+60 -warn-error -5 -alert --deprecated
+         -ccopt -Dcaml_signals_are_pending=caml_signals_might_be_pending)))
```
SkySkimmer: Not sure if this will actually work, I wanted to put a c_flags entry but that's not allowed (ocaml/dune#1124 / ocaml/dune#1073)

SkySkimmer:
If anyone knows a better way to do this, suggestions welcome.

ejgallego (changes requested):
I don't like this. Either:
- have upstream update `caml/compatiblity.h` and include it,
- add a conditional preprocessing line to coq_interp.c mathching on any of the values defined in `caml/version.h`

ejgallego:
What is the reason to close the PR?

SkySkimmer:
It looks like the ocaml change will be reverted.

ejgallego:
Thanks for the info!

PR #9261, opened 2018-12-21
Refine Auto Template Polymorphism option: allow auto template polymorphism without a warning.

Zimmi48:
**Kind:** enhancement.

The warning is now triggered if the option is not set and no attribute is used.
The option can be set to "enabled" or "disabled" and in both cases it will disable the warning.

### DIFF
--- a/vernac/comInductive.ml
+++ b/vernac/comInductive.ml
@@ -43,18 +43,32 @@ let warn_auto_template =
 
 let should_auto_template =
   let open Goptions in
-  let auto = ref true in
-  let () = declare_bool_option
+  let auto = ref None in
+  let () = declare_stringopt_option
       { optdepr  = false;
-        optname  = "Automatically make some inductive types template polymorphic";
+        optname  = "Auto Template Polymorphism";
         optkey   = ["Auto";"Template";"Polymorphism"];
-        optread  = (fun () -> !auto);
-        optwrite = (fun b -> auto := b); }
+        optread  =
+          (fun () ->
+            Option.map
+              (function true -> "enabled" | false -> "disabled") !auto);
+        optwrite =
+          (fun s ->
+            auto :=
+              Option.map
+                (function
+                 | "enabled" -> true | "disabled" -> false
+                 | _ -> user_err Pp.(str "Auto Template Polymorphism option only accept \"enabled\" and \"disabled\" values.")
+                )
+                s
+          );
+      }
   in
   fun id would_auto ->
-    let b = !auto && would_auto in
-    if b then warn_auto_template id;
-    b
+  match !auto, would_auto with
+  | None, true -> warn_auto_template id; true
+  | None, false -> false
+  | Some b, _ -> b && would_auto
 
 let rec complete_conclusion a cs = CAst.map_with_loc (fun ?loc -> function
   | CProdN (bl,c) -> CProdN (bl,complete_conclusion a cs c)


SkySkimmer (changes requested):
There's already `Set Warnings "-auto-template".` so I don't see the benefit.

Zimmi48:
The idea is that it doesn't make sense that asking explicitly for auto template polymorphism requires you to additionally disable a warning. In the future, the default value of this option is planned to change and the warning will be removed.

But anyways, I'm currently considering that it would be even better to go all the way towards the solution I recommended in https://github.com/coq/coq/issues/8027#issuecomment-441090197 and introduce `Universes` option with possible values, `"monomorphic"`, `"template"`, `"polymorphic"`, `"cumulative"` and same for the values of the `universes` attribute. In 8.10, we would emit the auto template warning if this option is unset and there are no attributes. In 8.11, we would make `"monomorphic"` the default value and stop emitting the warning.

SkySkimmer:
>In the future, the default value of this option is planned to change and the warning will be removed.

The whole option is going away, not just the default IMO.

Zimmi48:
>The whole option is going away, not just the default IMO.

Why should there be a way of making universe polymorphism automatic but none for template polymorphism?

SkySkimmer:
Because the option is actually "make template polymorphism some types chosen in an occult way"

Zimmi48:
I see...

herbelin:
I'm supportive of this PR.

Thanks for all the work aiming at clarifying how to work with universe polymorphism. A question though, @Skyskimmer, I don't understand what you mean by "chosen in an occult way"?

SkySkimmer:
Why should there be a special meaning in using `Type` for inductive types? ie why would `Inductive prod A B : Type := pair : A -> B -> prod A B.` be template polymorphic but not `Inductive prod@{i j} (A:Type@{i}) (B:Type@{j}) : Type@{max(i,j)} := ...`? Especially when `Print prod` with print universes on will show the later one.

herbelin:
> Why should there be a special meaning in using `Type` for inductive types?

Why `Type` has a special meaning in inductive types is somehow a convention. It was at a time where there was not the rich language of universes that we have now. It was at a time where we were not considering different forms of polymorphism but were instead looking for a default form of polymorphism obtained without extra syntax.

> ie why would `Inductive prod A B : Type := pair : A -> B -> prod A B.` be template polymorphic but not `Inductive prod@{i j} (A:Type@{i}) (B:Type@{j}) : Type@{max(i,j)} := ...`? Especially when `Print prod` with print universes on will show the later one.

Are you saying that there is no big difference between template polymorphism and the "full" polymorphism? If yes, I will agree. We can certainly reformulate the current implementation of template polymorphism as a particular case of full polymorphism where generalization is only on the conclusion of parameters (i.e. not on occurrences of`Type` in the types of constructors) and where instantiation of universes uses an ad hoc form of implicit arguments:

- `template_polymorphic_prod A B := prod@{inferred-level-of-A,inferred-level-of-B} A B`
- `template_polymorphic_prod A := prod@{inferred-level-of-A,prod.u1} A` (which can alternatively be seen as an eta-reduction of `fun B : Type@{prod.u1} => prod@{inferred-level-of-A,inferred-level-of-B} A B`)
- `template_polymorphic_prod :=prod@{prod.u0,prod.u1}` (which can alternatively be seen also as an eta-reduction of `fun (A : Type@{prod.u0}) (B : Type@{prod.u1}) => prod@{inferred-level-of-A,inferred-level-of-B} A B`)

With all the "uglyness" of having universe quantifiers only prenex and to use fixed levels (the `prod.u0` and `prod.u1`) to fill these quantifiers whenever `prod` is not fully applied.

But I'm unsure to exactly answer your comment. For instance, do you agree with the view I give above of "template polymorphism"?

SkySkimmer:
>Are you saying that there is no big difference between template polymorphism and the "full" polymorphism? 

No, I'm saying that using the type annotation to decide between monomorphic and template polymorphic is confusing (it certainly confused me for a while).

>But I'm unsure to exactly answer your comment.

The question is how should the choice between template poly and monomorphic be made. I want to move towards only explicit and specific annotations (currently using the attribute system), this PR goes in another direction.

Zimmi48:
I had missed the "lack of explicitness considered harmful" meaning in Gaëtan's PR. I will close this PR because it goes indeed against this spirit but I strongly recommend making the doc better.

First, the warning should be in the "deprecated" category, e.g. "Automatically declaring an inductive type template polymorphic is deprecated: make this explicit with an attribute or disable this behavior with option...".

Then, the documentation of the option should make the plan that this option will eventually be turned off by default explicit, and the warning should be documented alongside.

PR #566, opened 2017-04-15
stm: separate parsing state from system state

gares:
This PR builds on #459 (even if rewrites part of it) in order to make
Stm.parse_sentence a robust API.  In particular each document node
now holds a parsing state.

This PR also paves the way to quickly obtaining parsing states without executing
commands.  As of today this is done only for Lemma/Qed (that enable the
"Classic" mode for tactics).

In another PR (the one about proof modes) one could have "ProofMode" be
handled like that (today it is handled as Notation, i.e. it forces execution).

PR #9263, opened 2018-12-21
[STM] explicit handling of parsing states

maximedenes:
DAG nodes hold now a system state and a parsing state.
The latter is always passed to the parser.

This paves the way to decoupling the effect of commands on the parsing
state and the system state, and hence never force to interpret, say,
Notation.

Handling proof modes is now done explicitly in the STM, not by interpreting
VernacStartLemma.

Similarly Notation execution could be split in two phases in order to obtain a
parsing state without fully executing it (that requires executing all
commands before it).

Supersedes #566. In particular, @ejgallego's concerns about assigning a dummy parsing state have been addressed. The parsing state is now set after `process_transaction`.

The next steps (which should simplify the code a lot) will be to:
- Remove `VtUnknown` cases, making it possible to understand the structure of the document without executing commands.
- Separate the interpretation phase in a parsing effect phase and interpretation proper.

Depends on #9335


maximedenes:
Needs rebase?? I want my CI, coqbot!

maximedenes:
Working on fixing the remaining test-suite failures, should be easy.

maximedenes:
@gares we still have an issue with proofs in `Load`, since `Load` bypasses the STM, but the proof mode is now handled by the STM...

gares:
It seems the implementation of Load tried to use the old/broken way of setting the classic proof mode using Proof_global, and call the parser at a lower level. You should be able to patch vernacentries to call the functional parser and pass to it "Some classic" instead.

gares:
I think we should also deprecate Load.

maximedenes:
> You should be able to patch vernacentries to call the functional parser and pass to it "Some classic" instead.

But when do I know whether I should pass `Some "Classic"` or `None`? This is precisely managed by the STM...

gares:
I don't know when the compromise was made, but the code of load always passes classic or better tries...

maximedenes:
So you mean it parses tactics even outside of proof mode?

maximedenes:
> I think we should also deprecate Load.

I'm a bit surprised that it is so hard to support `Load`. After all, it is just an atomic transaction that groups several sentences, nothing too fancy. Compared to propagating side effects from slaves to master, or supporting navigation commands as part of the document, is seems much easier.

gares:
I think it does not get easy when the bunch of commands cannot be represented as a line in the dag. What if you Load in the middle of a proof? what if Load leaves a proof open? (while the former is not likely to be used in the wild, the latter is *the way* Load was used before PG existed.

I think deprecating Load is still a good option.

gares:
Let's put it like that. You write the vernac_classifier for Load, and I write the rest of the code :-)

maximedenes:
> What if you Load in the middle of a proof? what if Load leaves a proof open?

These two scenarios are already forbidden today.

ejgallego:
Thanks for the PR, will take a while to review the PR due to holidays; sorry for that.

maximedenes:
> You write the vernac_classifier for Load, and I write the rest of the code :-)

Ok, intuitively, it should be a max of all effects contained in the files being loaded (`VtNow` if there's at least a `VtNow` in the file, and `VtSideff` if there's a side effect). I'll give it a try (and ask for the rest of the code ;) )

maximedenes:
> due to holidays; sorry for that.

Don't be sorry for your holidays, enjoy them ;) This PR can wait a bit!

parsing/pcoq.mli
```diff
@@ -41,6 +41,16 @@ end
...
+
+   Note that parsing a Coq document is in essence stateful: the parser
+   needs to recognize commands that start proofs and use a different
+   parsing entry point for them.
+
+   We thus provide two different interfaces: the "raw" parsing
+   interface, in the style of camlp5, which provides more flexibility,
+   and a more specialize "parse_vernac" one, which will indeed adjust
+   the state as needed.
+
```
ejgallego: I am not sure I fully share the wording here. In a sense all that `parse_vernac` is doing is to choose a different parsing entry point; so indeed maybe `parse_vernac` should become pure and take an extra parameter "proof_mode" such that clients do indeed have to specify the parsing mode.

Long term, indeed, when inside a proof we should never ever have to support vernacs, but tactics [which could piggyback on vernac if wanted].

parsing/pcoq.mli
```diff
@@ -269,3 +279,7 @@ type any_entry = AnyEntry : 'a Entry.t -> any_entry
 
 val register_grammars_by_name : string -> any_entry list -> unit
 val find_grammars_by_name : string -> any_entry list
+
+(** Parsing state handling in the STM *)
```
ejgallego: Why is the STM mentioned here? AFAICT this module lives below it.

stm/stm.ml
```diff
@@ -569,9 +579,38 @@ end = struct (* {{{ *)
...
+      Hooks.(call state_ready ~doc:dummy_doc (* XXX should be taken in input *) id)
+
   let get_state id = (get_info id).state
+
+  let get_parsing_state id =
+    stm_pperr_endline (fun () -> str "retrieve parsing state state " ++ str (Stateid.to_string id) ++ str " }}}");
+    match (get_info id).state with
+    | FullState s -> s.Vernacstate.parsing
+    | ParsingState s -> s
+    | ErrorState _ | EmptyState -> anomaly Pp.(str"parsing state undefined")
```
ejgallego: It would be desirable that this doesn't raise an anomaly, but returns `None`.

vernac/lemmas.ml
```diff
@@ -335,12 +335,28 @@ let initialize_named_context_for_proof () =
...
+let proof_mode_opt_name = ["Default";"Proof";"Mode"]
+let () =
+  Goptions.declare_string_option Goptions.{
+    optdepr = false;
+    optname = "default proof mode" ;
+    optkey = proof_mode_opt_name;
+    optread = get_default_proof_mode;
+    optwrite = set_default_proof_mode;
+  }
+
```
ejgallego: It looks like this should live in `vernacentries`.

vernac/pvernac.ml
```diff
@@ -39,17 +52,33 @@ module Vernac_ =
...
   end
 
+let set_tactic_entry oname =
+  command_entry_ref := oname
+
 let main_entry = Vernac_.main_entry
 
-let set_command_entry e = Vernac_.command_entry_ref := e
-let get_command_entry () = !Vernac_.command_entry_ref
+type frozen_t = string option
```
ejgallego: I am not sure it is worth duplicating the state here twice. Indeed, `Vernacstate` contains this info so I think we should make the API here functional, something such as `parse_vernac_or_tactic : ?tmode:string -> ....` Clients can indeed be sure they properly thread the `Vernacstate.t`.

ejgallego (commented):
Mostly looks pretty good, however I think it is worth to explore the option that doesn't add a new summary entry.

stm/stm.ml
gares: I think that `EmptyState` should be anomaly (it is an invariant).

While for ErrorState I'm not convinced, in the end, that this is correct. Indeed we want to be able to parse a sentence after a state that was found to be erroneous. In particular this assertion, and the data carried by ErrorState makes "parse;parse;exec;exec" =/= "parse;exec;parse;exec" when the first execution fails, since the following parse is raising an anomaly.

gares (changes requested):
Fix the data carried by `ErrorState`

ejgallego:
Indeed `ErrorState`could include error recovery information, for example, the previously valid parsing state.

Something we saw when developing LSP is that parsing errors tend to be essentially different than post-parsing ones.

maximedenes:
`ErrorState` needs to be fixed indeed. In fact, I'm surprised this was not caught by tests.

@ejgallego Regarding `Load`, do you have any suggestion on how we should proceed?

ejgallego:
> @ejgallego Regarding `Load`, do you have any suggestion on how we should proceed?

Most uses for `Load` are to set options, etc... I'd say that we shouldn't support tactics inside files to be handled by load.

maximedenes:
> Most uses for `Load` are to set options, etc... I'd say that we shouldn't support tactics inside files to be handled by load.

Sounds reasonable indeed. What this patch is missing then is a nicer error message when opening a proof in a `Load`ed file.

ejgallego:
As you execute commands you can also indeed detect when a proof has been opened and thus thread the right parsing flag if you want.

vernac/pvernac.ml
maximedenes: > `Vernacstate` contains this info

Really? Can you point me to where the proof mode is stored in `Vernacstate`? Is it in the proof state?

ejgallego: Indeed it should be in the proof state, right?

maximedenes: Are you saying it is already in the proof state, or that we should put it there?

ejgallego: I think all the needed information is already there.

maximedenes: Sorry, I couldn't find the name of the proof mode there. Where is it, more precisely?

ejgallego: In `Proof_global`:
```
type pstate = {
  terminator : proof_terminator CEphemeron.key;
  endline_tactic : Genarg.glob_generic_argument option;
  section_vars : Constr.named_context option;
  proof : Proof.t;
  mode : proof_mode CEphemeron.key;
  universe_decl: UState.universe_decl;
  strength : Decl_kinds.goal_kind;
}

type t = pstate list
```

ejgallego: Of course this needs tweaking due to the Ephemeron, etc... but the info is already there.

maximedenes: Oh! So this PR was actually removing the mode from the proof state before I took it over. That's why I couldn't see it.

ejgallego: Yeah I guess where to place it is kinda open, but I think I do prefer that part of the parser to be functional and have the state living in upper layers.

ejgallego: That way, clients of the parser are indeed forced to declare the provenance of whether they think we are parsing proofs or not.

parsing/pcoq.mli
maximedenes: > so indeed maybe `parse_vernac` should become pure and take an extra parameter "proof_mode" such that clients do indeed have to specify the parsing mode.

In fact, this is what we have done, see `Vernacstate.Parser.parse`. It seems to me that the comment is now obsolete.

Did you want another API?

ejgallego: I am not sure the "is in essence stateful" is something that I agree with. The parser doesn't need to recognize proofs, it is the clients of the parser that need to do so.

Indeed as I mentioned below I would prefer if the state of `pvernac` is eliminated, and the `Vernacstate.Parser` module would become redundant.

ejgallego: To say in different words, for what `Vernacstate.Parser` does, it think it is better to reify the whole imperative state to a parameter.

maximedenes: I looked a bit more closely, and I think the current mess is more or less due (as you mention above) to the injection of vernac in tactics, which gives `tactic_entry : Vernacexpr.vernac_expr Entry.t`.

I'm afraid this prevents a clean encapsulation of proof modes at the `Proof_global` level (it has to be done in vernac). I will try to improve the current API for registering and passing proof modes, though.

maximedenes: I'm not sure what I should do about the use of `main_entry` in `g_toplevel`. It seems this one can't really use the functional interface, can it?

ejgallego: Oh indeed that one is a bit problematic :( I think parse_toplevel could indeed call the functional interface but that'll require a bit of massaging on the entry above.

gares:
if Load is VtNow, then it is OK (all scheduling affecting options are VtNow).

parsing/pcoq.mli
maximedenes: Meanwhile, I can expose the non-functional one (i.e. the real grammar entry) in `Vernac_`, and migrate all other clients to the functional one.

maximedenes:
> if Load is VtNow, then it is OK (all scheduling affecting options are VtNow).

Ah, right! But then it means I had forgotten about these options when I told you about the plan to get rid of execution at `Add` time. One more step in that plan, then...

maximedenes:
Ok, I'm almost there for proof modes. One question: @gares, why do you remove `VtProofMode` as part of this PR? It seems to make the `Proof Mode` command bypass the STM, which should be the only component in charge of setting the proof mode. I'm a bit lost.

maximedenes:
> why do you remove `VtProofMode` as part of this PR

@gares similarly, why does the proof mode no longer appear in `VtStartProof`?

parsing/pcoq.mli
maximedenes: Fixed.

stm/stm.ml
maximedenes: Fixed, `ErrorState` now carries a parsing state.

vernac/lemmas.ml
maximedenes: Done.

gares (approved):

gares:
> This currently breaks parsing of tactics by the toplevel. We probably need to discuss how to address it with @ejgallego.

This was fixed using a temporarily exported API to "implement" parametrized grammar entries over the non terminal for tactics (proof modes).
We discussed a few alternatives, none of them pertinent to this PR. Namely:
- support grammars with parametric non terminals (now that we embed gramlib)
- refactor the grammar and the AST in order to be able to easily reuse "decorators" of both commands/tactics and provide many initial symbols (instead of 1 with a parameter)
The latter seems the best since after all what one can write in proof mode needs not to be a strict superset of commands, eg we may want to just forbid problematic commands rather than twisting the system around the postulate that *anything* can be written anywhere.

ejgallego:
@maximedenes you mentioned this PR was ready, I'll be happy to review this prioritarily so ping me when you address @gares comments.

maximedenes:
> you mentioned this PR was ready, I'll be happy to review this prioritarily so ping me when you address @gares comments.

Thanks, will update this afternoon. But the mtac2 overlay is pending on #9347, if you can merge that one meanwhile, it would help.

maximedenes:
> 1. it is unclear how we can generate it and

I'm not sure why you say it is unclear. When we call `freeze_invalid`, we save the previously known parsing state for the same id, if it exists (which means that the error was triggered after the parsing phase for that id). If not (essentially a parsing error), we store `None`.

I'll make a comment about the expected precondition for `parse_sentence` indeed (the parsing state should be valid).

ejgallego:
Overlays at:
- https://github.com/Mtac2/Mtac2/pull/141
- https://github.com/ppedrot/ltac2/pull/94
- https://github.com/mattam82/Coq-Equations/pull/163

[we should specify better the overlay procedure BTW]

ejgallego (approved):
This is a great improvement on the status quo, and IMHO ready to merge. We shall indeed refine this API, so users of it should be beware of the instability, but the PR is more than ready to go.

Leaving 24 hours for comments as this as been open for a long time.

vernac/vernacstate.ml
```diff
@@ -36,11 +62,14 @@ let do_if_not_cached rf f v =
...
     proof  = update_cache s_proof (Proof_global.freeze ~marshallable);
-    shallow = marshallable }
+    shallow = marshallable;
+    parsing = Parser.cur_state ();
+  }
 
 let unfreeze_interp_state { system; proof } =
   do_if_not_cached s_cache States.unfreeze system;
   do_if_not_cached s_proof Proof_global.unfreeze proof
+  (* FIXME unfreeze parsing state? *)
```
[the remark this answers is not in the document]
gares: hem...

gares:
There is a FIXME that should be easy to address that was forgotten IMO

maximedenes:
Indeed, it was already mentioned in the discussion, but I forgot about it. Will try to add the missing `unfreeze`.

ejgallego:
Ok will merge when you guys tell me. In principle tomorrow if the FIXME is resolved.

maximedenes:
> Ok will merge when you guys tell me. In principle tomorrow if the FIXME is resolved.

The FIXME is resolved, let's hope CI goes fine!

maximedenes:
@ejgallego CI went fine, I think this is ready :)

PR #9265, opened 2018-12-21
Do not exclude "opened" bugs from report

maximedenes:
I don't know why they were excluded, but it prevents their storage as artifacts, which is inconvenient.


maximedenes:
It seems `report.sh` has no owners btw. @SkySkimmer do you agree to be one?

SkySkimmer:
>It seems report.sh has no owners btw. @SkySkimmer do you agree to be one?

Sure.

maximedenes:
I see you're an owner of `test-suite/save-logs` what has it become?

SkySkimmer (approved):

Issue #4638, opened 2016-03-21
nested Instance exits proof mode


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#4638
From: @siegebell
Reported version: 8.5
CC:   @gares, @JasonGross

See also: [BZ#3890](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%233890%22)

ejgallego:
@maximedenes can you see if this solved, and if not what remains to be done?

gares:
The bug is in the code of VtUnknown, when we will remove it (8.10 or 8.11) this bug will go.
One can avoid the bug by not turning on the Refine instance Mode option.
But the code is till there if you look for it.

ejgallego:
Still a problem cc : @maximedenes , the parsing state is not there:
```
egallego@lasserre:~/external/coq-master$ dune exec -- coqc k.v -debug
Error: Anomaly "Uncaught exception Option.IsNone."
Please report at http://coq.inria.fr/bugs/.
frame @ file "toplevel/coqc.ml", line 58, characters 4-16
frame @ file "toplevel/coqc.ml", line 37, characters 2-81
frame @ file "list.ml", line 106, characters 12-15
frame @ file "toplevel/ccompile.ml", line 188, characters 2-39
frame @ file "toplevel/ccompile.ml", line 127, characters 18-88
frame @ file "toplevel/vernac.ml", line 177, characters 30-88
raise @ file "clib/exninfo.ml", line 65, characters 2-15
frame @ file "toplevel/vernac.ml", line 117, characters 6-19
frame @ file "toplevel/vernac.ml", line 100, characters 6-89
frame @ file "stm/stm.ml", line 3121, characters 11-50
raise @ file "clib/option.ml", line 50, characters 9-21
```

maximedenes:
Is that after #9530?

ejgallego:
No, it is on master.

maximedenes:
This has been fixed by #9530.

Issue #3890, opened 2014-12-27
In refine instance mode, starting a new instance forgets the one in progress


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#3890
From: @mikeshulman
Reported version: 8.5
CC:   @gares, @JasonGross, @maximedenes

See also: [BZ#4638](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%234638%22)

ejgallego:
@maximedenes can you see if this solved, and if not what remains to be done?

maximedenes:
Fixed by https://github.com/coq/coq/pull/9530.

maximedenes:
In fact no, there is still an STM bug:

```
Class Foo.
Class Bar := b : Type.

Set Nested Proofs Allowed.
Instance foo : Foo.
(* 1 subgoals, subgoal 1 (ID 4)
  
  ============================
   Foo *)

Instance bar : Bar.
exact Type.
Defined.
(* bar is defined *)

About foo.
(* foo not a defined object. *)

Defined.
(** Error: No focused proof (No proof-editing in progress). *)

Backtrack 12 0 0.
(*
Error: Anomaly "edit_at 12: Uncaught exception Stm.Vcs_aux.Expired.."
Please report at http://coq.inria.fr/bugs/.
*)
```

Cc @gares

SkySkimmer:
Cannot reproduce

maximedenes:
Really? I get it consistently with coqtop, master at 5c5bd952e9c28c3acf740fcdced03b2b7145076d.

ppedrot:
Can still reproduce on a recent master 302334cd0e3961a8476e8071ca4482318e0dce6f. But the validity the new avatar of bug is arguable, in my opinion. `BackTo` is a coqtop-only command that cannot be used from UIs and is basically a pointy stick to poke at dangerous places. I'm inclined to close this issue since the original behaviour has been fixed.

ejgallego:
The bug is valid and it impacts all IDEs. In Maxime example, the proof state is threaded incorrectly and lost, resulting in an invalid configuration.

gares:
This bug is about the stm. If it has no more anything to do with BackTo it should ve affected to another component.

ejgallego:
Indeed, tags adjusted. The bug is reproduced without the STM with the original script:

```coq
Class Foo.
Class Bar := b : Type.

[...]

Issue #3632, opened 2014-09-16
feature request: require "Instance :=" to be fully solvable


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#3632
From: @mikeshulman
Reported version: 8.5
CC:   @ejgallego, @JasonGross, @ppedrot

See also: #3631

SkySkimmer:
I would call it `Refine Instance foo := bar.` 
The code is a bit too much a hairball for me to try though.

ejgallego:
Yup that looks good but indeed solving this is tricky... see also #1150 #1122

ejgallego:
For quite a few reasons I think that trying to clean up this is prioritary. In particular, `Instance` is not special case by a non-negligible part of the codebase due to its behavior.

We will have pain with this, but I don't see a pain-free path.

ejgallego:
So there is one problematic case:
```coq
Instance foo ... : A.
```
and the instance is fully resolved. Then, `Qed` is not needed! Thus, making the change in 8.8 so that
```coq
Instance foo ... : A.
Qed.
```
is required will break hell on earth, and worse, it is not backwards compatible.

So today we have:
```coq
Instance foo. ...      (* may open a proof, the flag has no effect, hard to make compatible. *)
```
how to proceed @mattam82 @SkySkimmer @herbelin @JasonGross @maximedenes @ppedrot ?

Should we suggest all the developments doing:
```coq
Instance foo : A.
```
to change to
```coq
Instance foo : A := {}.
```
?

JasonGross:
I consider `Instance foo : A.` to be bad practice when fully resolved.  I generally write `Instance foo : A := _.` to indicate something that is fully solved by typeclass resolution (which I am adding for caching purposes or whatever).

ejgallego:
Sadly the code is a bit funny and `Instance foo : A` seems to have different semantics from `Instance foo : A := _`.

SkySkimmer:
I think the only time `Instance foo : A.` can be auto solved is when it lands in this case https://github.com/coq/coq/blob/2c2a08083bc535397359299690d0bfb3523a9ee1/vernac/classes.ml#L229 ie when `A` is a trivial class `Class A args := {}.`
I think we should just remove this special case and make people use `:= {}.` or `:= _.` instead.

SkySkimmer:
[...]

PR #9270, opened 2018-12-21
Turn `Refine instance Mode` off by default

maximedenes:
This PR cleans up some code related to `Instance`, and turns the `Refine Instance Mode` option off by default.

This is the first of two steps to be able to tell statically which commands open proofs, which is a desirable property for the vernac IMHO.

The second step will be to make `Instance` without a body always open a proof.

### DIFF
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -3006,7 +3006,7 @@ let process_transaction ~doc ?(newtip=Stateid.fresh ())
               let bname = VCS.mk_branch_name x in
               let opacity_of_produced_term = function
                 (* This AST is ambiguous, hence we check it dynamically *)
-                | VernacInstance (false, _,_ , None, _) -> GuaranteesOpacity
+                | VernacInstance (_,_ , None, _) -> GuaranteesOpacity
                 | _ -> Doesn'tGuaranteeOpacity in
               VCS.commit id (Fork (x,bname,opacity_of_produced_term (Vernacprop.under_control x.expr),[]));
               let proof_mode = default_proof_mode () in
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -98,6 +98,9 @@ Vernacular commands
 
 - `Declare Instance` now requires an instance name.
 
+- Option `Refine Instance Mode` has been turned off by default, meaning that
+  `Instance` no longer opens a proof when a body is provided.
+
 Tools
 
 - The `-native-compiler` flag of `coqc` and `coqtop` now takes an argument which can have three values:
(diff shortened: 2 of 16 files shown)


maximedenes:
Let's see how CI looks to evaluate compatibility issues.

stm/stm.ml
```diff
@@ -3008,7 +3008,7 @@ let process_transaction ~doc ?(newtip=Stateid.fresh ())
               let bname = VCS.mk_branch_name x in
               let opacity_of_produced_term = function
                 (* This AST is ambiguous, hence we check it dynamically *)
-                | VernacInstance (false, _,_ , None, _) -> GuaranteesOpacity
+                | VernacInstance (_,_ , None, _) -> GuaranteesOpacity (* XXX really? can't we close it by Defined? *)
```
SkySkimmer: What does this mean?

maximedenes: Well, it probably means I don't understand the classification, but why would `Instance foo : C.` "guarantee opacity"?

SkySkimmer: IIRC GuaranteesOpacity means that if the proof is closed by Qed it's really opaque. Derive breaks this. I don't know why Instance would though.

maximedenes: Aaaah! Thanks for the clarification, I'll remove the comment before this is merge, then.

ejgallego:
cc: https://github.com/coq/coq/issues/3632 https://github.com/coq/coq/issues/3890

maximedenes:
also #4638

maximedenes:
@SkySkimmer @ejgallego @Zimmi48 any idea what the formal-topology failure could mean? It does not seem to find bignums...

SkySkimmer:
Looks like it skipped downloading artifacts for some reason. You can just reboot the job. We may want to report to gitlab too.

maximedenes:
> Looks like it skipped downloading artifacts for some reason. You can just reboot the job. We may want to report to gitlab too.

Could it be because I restarted the job some time after the first run? How long are the artifacts stored?

SkySkimmer:
build artifacts last 1 week

ejgallego:
We should keep in mind the problems Program Instance has tho; in particular with regards to async / terminators. IMHO the potential for regressions with this migration path is not zero.

maximedenes:
Can this get a review? @ppedrot maybe ?

maximedenes:
@gares would you mind merging this? Synchronization with downstream devs has been a bit difficult, and @ppedrot told me he'd be busy for a few more days. Thanks!

ppedrot (approved):

ppedrot:
Let's merge.

maximedenes:
Thanks a lot!

SkySkimmer:
Missed doc changes: default is off, also the doc has an example using Refine Instance Mode.

PR #9273, opened 2018-12-22
Fix #9272: `Unset Nested Proofs Allowed` does not capture nested `Ins…

### LINKED ISSUE #9272: `Unset Nested Proofs Allowed` does not capture nested `Instance` proofs.
On `master`, `Goal True. Goal True.` triggers an error, but:

```coq
Class C := { n : nat }.
Instance: C. Instance: C.
```

passes. I believe this is a bug, that will make migrating to `Unset Refine Instance Mode` more difficult. I'll try to fix it.

maximedenes:
…tance` proofs.

We forbid commands that may open proofs inside proofs.

TODO
- [ ] Add a test case

### DIFF
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -2993,7 +2993,14 @@ let process_transaction ~doc ?(newtip=Stateid.fresh ())
       (* Unknown: we execute it, check for open goals and propagate sideeff *)
       | VtUnknown, VtNow ->
           let in_proof = not (VCS.Branch.equal head VCS.Branch.master) in
-          let id = VCS.new_node ~id:newtip () in
+          if not (get_allow_nested_proofs ()) && in_proof then
+            "Commands which may open proofs are not allowed in a proof unless you turn option Nested Proofs Allowed on."
+            |> Pp.str
+            |> (fun s -> (UserError (None, s), Exninfo.null))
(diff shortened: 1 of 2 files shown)


SkySkimmer:
>We forbid commands that may open proofs inside proofs.

Is that really what we want? How difficult would it be to be more accurate to forbid only commands which actually open proofs?

maximedenes:
> Is that really what we want? How difficult would it be to be more accurate to forbid only commands which actually open proofs?

I think it is what we want because:
1- being more accurate is not trivial, since the test does not have the knowledge of whether the command did open a proof
2- the whole codepath (handling `VtUnknown` commands inside proofs) is very shaky, so better prevent it by default

maximedenes:
While writing a test case, I found another bug: the check for nested proofs seems to ignore control. So for example the following fails:

```coq
Goal True.
Fail Goal False.
```

as if `Fail` was not there.

But this is really a mess: what should the classification of `Fail Goal False` be? Today it is classified as `VtUnknown`, for unclear reasons, so the error is triggered. If we classify it as `VtQuery` instead, it won't error, but also won't test the nested proof restriction.

So I guess the more pragmatic way to make progress is to not include a test case.

gares (approved):

PR #9276, opened 2018-12-23
Remove dead code from CClosure.

ppedrot:
It seems that it was a remnant of a time where Reductionops would share the same data types.


maximedenes:
> It seems that it was a remnant of a time where Reductionops would share the same data types.

I hope one day they can share they same types again, but anyway this code won't be needed.

maximedenes (approved):

PR #9277, opened 2018-12-23
[dune] Build refman with fatal warnings like in the Makefile build.

Zimmi48:
That's a first step. The next one should be to support setting SPHINXWARNERROR=0 like in the Makefile build.

**Kind:** infrastructure.

### DIFF
--- a/doc/dune
+++ b/doc/dune
@@ -10,8 +10,10 @@
   ;   + tools/coqdoc/coqdoc.css
   (package coq)
   (source_tree sphinx)
-  (source_tree tools))
- (action (run sphinx-build -j4 -b html -d sphinx_build/doctrees sphinx sphinx_build/html)))
+  (source_tree tools)
+  (env_var SPHINXWARNOPT))
+ (action
+  (run sphinx-build -j4 %{env:SPHINXWARNOPT=-W} -b html -d sphinx_build/doctrees sphinx sphinx_build/html)))
 
 (alias
  (name refman-html)
--- a/doc/README.md
+++ b/doc/README.md
@@ -101,18 +101,21 @@ Alternatively, you can use some specific targets:
 Also note the `-with-doc yes` option of `./configure` to enable the
 build of the documentation as part of the default make target.
 
-If you're editing Sphinx documentation, set SPHINXWARNERROR to 0
-to avoid treating Sphinx warnings as errors.  Otherwise, Sphinx quits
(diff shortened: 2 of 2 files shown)


SkySkimmer:
Something went wrong in the gitlab windows job, I just cancelled it with "Duration: 1531 minutes 27 seconds"

Zimmi48:
The job only logged things for about twenty minutes then remained blocked without displaying anything. The job is supposed to be killed by GitLab after 4 hours, so there was a GitLab bug. Should we report it?

Zimmi48:
On another note, the Dune refman job was 13 minutes long while the normal refman job (which does more) was just 6 minutes long...

ejgallego:
Needs to be investigated indeed. It could be: a wrong -j setting, costly artifacts, bug in build.

doc/dune
```diff
@@ -11,7 +11,7 @@
   (package coq)
   (source_tree sphinx)
   (source_tree tools))
- (action (run sphinx-build -j4 -b html -d sphinx_build/doctrees sphinx sphinx_build/html)))
+ (action (run sphinx-build -j4 -W -b html -d sphinx_build/doctrees sphinx sphinx_build/html)))
```
ejgallego: I would prefer we go a bit further and unify this with the Makefile setting by using a variable `%{env:SPHINXOPTS=-j 2 -W}` so that users can disable warnings by overriding it.

Zimmi48: Sounds good.

Zimmi48:
PR updated. Dune did not accept spaces in the default value of the environment variable and it is not possible either to pass an empty string because of Dune's quoting, but the way that works as been documented.

doc/dune
```diff
@@ -10,8 +10,10 @@
   ;   + tools/coqdoc/coqdoc.css
   (package coq)
   (source_tree sphinx)
-  (source_tree tools))
- (action (run sphinx-build -j4 -b html -d sphinx_build/doctrees sphinx sphinx_build/html)))
+  (source_tree tools)
+  (env_var SPHINXWARNOPT))
```
ejgallego: This dependency should be automatically added by Dune as `%{env:...}` does appear in the run rule.

ejgallego (approved):
Looks good to me, thanks @Zimmi48 .

doc/dune
Zimmi48: Thanks, I didn't know that.

PR #9278, opened 2018-12-23
[ci] Annotate plugins and libraries.

Zimmi48:
This will allow @coqbot to automatically add appropriate needs labels (overlay or fixing) depending on whether it breaks a plugin or a library.
In practice, we will be able to add the "needs: fixing" labels as soon as this PR is merged, but the "needs: overlay" labels will have to wait for #6724 because we don't want to flood PRs with non-relevant labels every time a PR with overlays is merged.

Independently of the applications for @coqbot, this is also useful to developers (especially new contributors) as this allows to distinguish pretty quickly between the two categories (and thus whether it's OK to break or not). We currently test 18 libraries and 12 plugins.

**Kind:** infrastructure.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -174,20 +174,17 @@ after_script:
   script:
     - set -e
     - echo 'start:coq.test'
-    - make -f Makefile.ci -j "$NJOBS" ${TEST_TARGET}
+    - make -f Makefile.ci -j "$NJOBS" "${CI_JOB_NAME#*:}"
     - echo 'end:coq.test'
     - set +e
   dependencies:
     - build:base
-  variables: &ci-template-vars
-    TEST_TARGET: "$CI_JOB_NAME"
 
 .ci-template-flambda: &ci-template-flambda
   <<: *ci-template
   dependencies:
     - build:edge+flambda
   variables:
-    <<: *ci-template-vars
     OPAM_SWITCH: "edge"
     OPAM_VARIANT: "+flambda"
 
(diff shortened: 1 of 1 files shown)


Zimmi48:
After this PR, the only jobs which do not follow the convention of having a prefix with ":" are the Windows packaging jobs. I don't recall why we did not add "pkg:" in their name but I think there was a reason.

.gitlab-ci.yml
```diff
@@ -445,93 +445,98 @@ validate:edge+flambda:
...
-ci-elpi:
-  <<: *ci-template
+library:ci-fiat-crypto:
+  <<: *ci-template-flambda
 
-ci-equations:
+library:ci-fiat-crypto-legacy:
+  <<: *ci-template-flambda
+
+library:ci-fiat-parsers:
```
SkySkimmer: There's a plugin there though.

.gitlab-ci.yml
```diff
@@ -445,93 +445,98 @@ validate:edge+flambda:
...
 
-ci-aac_tactics:
-  <<: *ci-template
+# Libraries are by convention the projects that depend on Coq
+# but not on its ML API
 
-ci-bedrock2:
+library:ci-bedrock2:
   <<: *ci-template
   allow_failure: true
```
SkySkimmer: I guess this started working again at some point, if so then we should remove the allow_failure.

Zimmi48:
We can discuss whether adding "needs: overlay" automatically is useful or not. My main interest was rather to detect breakages in libraries. Ignoring the problem of libraries being broken by their authors (if this happens too much, we should start to do something about it), libraries broken by PRs are interesting to categorize PRs as breaking change or not. And even when the breakage is accidental, I also think we should start extracting test-suite additions from these "accidental regressions" that never get merged in master.

Zimmi48:
> >We currently test 18 libraries and 12 plugins.
>
>Did you count packed jobs as 1 thing? eg the formal-topology job runs mathclasses and corn too.

Indeed, I did. So counting the packed libraries as well, we should be closer to 23+ libraries.

.gitlab-ci.yml
```diff
@@ -180,7 +180,7 @@ after_script:
   dependencies:
     - build:base
   variables: &ci-template-vars
-    TEST_TARGET: "$CI_JOB_NAME"
+    TEST_TARGET: "${CI_JOB_NAME#*:}"
```
Zimmi48: This did not work in CI even if I tested and it works well in bash and dash. @SkySkimmer: do you have an idea on what the problem is?

Zimmi48: I did not understand what you meant but I realized that the `TEST_TARGET` indirection (inherited from Travis) is unnecessary.

SkySkimmer: Oops I misread and thought this was the artifact name key.

SkySkimmer: The point is that gitlab doesn't do any of the `${}` expansions, only the basic `$foo` one (I think it doesn't even have the `${foo}` without modifiers)

Zimmi48:
Updated. Hopefully it will work this time. Let me know if something needs to be done regarding artifact names (it doesn't look like it: all the artifact names seem already be based on the job name, even when it contains a colon).

.gitlab-ci.yml
Zimmi48: @ejgallego I'm removing that second commit then.

ejgallego (commented):
I am not sure I do follow the distinction between `library:` and `plugin:`; most things tagged `plugin` do also contain non-trivial libraries; and indeed some of the benches don't properly correspond to the right granular names. I would tend to classify developments instead on their dependencies; so for example a plugin can depend on the stdlib, ltac, ml API, etc... but this is tricky.

Also, I am not fond of using job naming as metadata; IMHO it would be more convenient that the CI job would output the meta-data itself in machine-readable format.

Automated tagging could be interesting, but I am not sure our setup is stable enough to work well yet.

A minor naming question is `lib:foo` vs `library:foo`.

Anyways, this change is trivial and I am happy to have it merged in the current form so @Zimmi48 can progress in his experiments.

Zimmi48:
Downloading and parsing the trace requires an extra step (although TBH this extra step is already done so that we can restart the build in case of spurious failure).

But another point in favor of the naming scheme was that it could be useful for humans (= developers) too: in a sense the naming "plugin" / "library" is a bit arbitrary but the point is to be able to distinguish between API breaking changes and user-side breaking changes.

When GitLab finally implements DAG-like workflows, we will be able to better separate builds into more granular jobs and thus the "plugin" / "library" segregation will make more sense. If you want, we can also wait for this to happen first.

ejgallego:
I was indeed pointing that out under the assumption that the bot already downloads the log.

IMHO it would be quite useful for the bot to detect errors and indeed build the link to the error message / source code [I could even contribute this, https://github.com/coq/bot/issues/29]

Whether to wait or not I am neutral here; I am happy to proceed as you think is best.

Regarding the plugin/library separation it is indeed a bit more subtle than that [some parts may form a single unit via a .v -> .cmo build dep and a .cmo -> .v runtime dep], but as long as don't rely too much on that definition I am OK with that.

Zimmi48:
I suggest we proceed to merge as I don't see how this can be hurtful, even in the current situation. Let me know if you are happy with "library" or if I should rename it to "lib".

ejgallego:
Ok will merge, I am happy with both, @skyskimmer any preference?

SkySkimmer:
library > lib IMO

PR #9279, opened 2018-12-23
CoqIDE on lablgtk3

### LINKED ISSUE #7055: CoqIDE will be soon impossible to build in Debian derivatives

herbelin:
**Kind:** feature

This is a version of CoqIDE building on top of lablgtk3. It ~~still needs a development version of lablgtk3 for a minor change and otherwise would~~ builds against the official beta3 version of lablgtk3 from @garrigue.

The following features are missing:
- the possibility to add new keys for user queries
- the possibility to add wizard tactics
- the progression bar

I don't know if the two first features are commonly used. Personally, I never used them.

The deactivation of the progression bar is a bit frustrating. The pixmap would need to be redone with cairo (with an extra compilation dependency as far as I understand - [cairo-ocaml](https://www.cairographics.org/cairo-ocaml/) or [ocaml-cairo](https://github.com/Chris00/ocaml-cairo)). Do we want to go in this direction? And if yes, I would probably need help. @ppedrot do you have a particular opinion?

Another minor loss is the stippling ("pointillé") of an incomplete `Qed`. To be honest, I had not noticed that the blue background was stippled. So, if I'm not the only one not to have noticed it, maybe it is not so critical. As far as I understood, it could be redone also using Cairo. What do you think (and especially @gares)?

Technically, it is unclear if we can make it for the next Debian freeze. In any case, @silene, @zimmi48, depending on the schedule of 8.9, depending on what is possible with Debian, I'm ready to do a port of this PR on top of 8.8.2 and release an ad hoc version of 8.8 specifically for gtk3-based CoqIDE if this is the condition to have it in the next Debian stable.

- [X] Entry added in CHANGES.md.
- [ ] Remove dead code in `configwin*.ml` if sure it will not be ported.
- [x] Remove temporary commit for building with `lablgtk3.0.beta3` when `3.0.0` is out.
- [X] Use a discriminating color to replace stippling (used `light sky blue`)
- [x] Update windows build files

Fixes #7055.


SkySkimmer:
I guess you need to update the dependencies in the dockerfile and windows scripts.

herbelin:
> I guess you need to update the dependencies in the dockerfile and windows scripts.

Is it `dev/ci/docker/Dockerfile` and `dev/build/windows/`? There is no opam package yet for lablgtk3 and the situation is changing fast. So, this shall have to be done in a later step. Putting a sticker in the meanwhile.

By the way, I made nothing for dune either (but I know that @ejgallego is following this closely, and, lablgtk3 may eventually have a dune-based building).

jashug:
I do all my Coq development in CoqIde. I haven't tried this new version (let me know if/when that would be helpful), but I can comment on the changes from the PR description:

- I also have never used the first two missing features.
- I found the progress bar a nice visual aid, and liked that you could click to move to a point within the file (to locate an axiom, for example). It is nice to have, but it's lack isn't a deal-breaker for me.
- I found it useful that an incomplete Qed was styled differently, but am not attached to the stippling effect; I expect a solid light blue/grey background would work just as well for me.

Zimmi48:
>Technically, it is unclear if we can make it for the next Debian freeze. In any case, @silene, @Zimmi48, depending on the schedule of 8.9, depending on what is possible with Debian, I'm ready to do a port of this PR on top of 8.8.2 and release an ad hoc version of 8.8 specifically for gtk3-based CoqIDE if this is the condition to have it in the next Debian stable.

I don't understand what you are hoping for / asking me exactly but Coq packages have always been completely outdated in Debian and there is no reason to hope that this would change. On the contrary, the best we can hope is that Debian maintainers actually remove the Coq package entirely so that there is no confusion anymore that Debian users *should not* install Coq using their distribution package manager, but prefer opam (or Nix) instead. Moreover, AFAICT the whole point of this PR was not so that we can have a Debian Coq package but so that CoqIDE continue to be **buildable** under Debian (otherwise, even with opam, it would have been a nightmare). If you are in contact with Debian maintainers, tell them it is important that opam 2.0 gets its way in before the freeze (maybe it's already the case, I have no idea). In any case, I am not fond of the idea of a new 8.8 release just for this.

Zimmi48:
This is an example of situation where the flexibility that developing CoqIDE in its own repository would bring would overcome the cost: it could allow to do a new CoqIDE major release, depending on gtk3 and possibly compatible with all recently released versions of Coq...

coqide.opam
```diff
@@ -20,7 +20,7 @@ depends: [
   "dune"               { build & >= "1.4.0" }
   "coqide-server"
   "conf-gtksourceview"
```
SkySkimmer: Is this the right version?

ide/configwin.ml
```diff
@@ -37,8 +37,10 @@ type return_button =
   | Return_cancel
 
 let string = Configwin_ihm.string
+(*
 let strings = Configwin_ihm.strings
 let list = Configwin_ihm.list
+*)
```
SkySkimmer: Why keep this commented?

ide/configwin.mli
```diff
@@ -119,6 +120,7 @@ val list : ?editable: bool -> ?help: string ->
 		('a -> string list) ->
 		  'a list ->
 		    parameter_kind
+*)
```
SkySkimmer: Why keep this commented?

ide/configwin_ihm.ml
```diff
@@ -279,10 +278,10 @@ class ['a] list_selection_box
       (* initialize the clist with the listref *)
       self#update !listref
   end;;
-
+*)
```
SkySkimmer: Why keep this commented? etc for any other commented blocks.

INSTALL
```diff
@@ -43,8 +43,8 @@ WHAT DO YOU NEED ?
 
    - a C compiler
 
-   - for CoqIDE, the lablgtk development files (version >= 2.18.5),
-     and the GTK 2.x libraries including gtksourceview2.
+   - for CoqIDE, the lablgtk development files (version >= 3.0.0),
+     and the GTK 3.x libraries including gtksourceview3.
```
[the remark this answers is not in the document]
herbelin: Yes. It could be possible to have a code compilable with both lablgtk2 and lablgtk3 with a couple of `ifdef` (or similar tricks).

coqide.opam
herbelin: I did not see an opam package for lablgtk3 yet. I guess that an opam package `conf-gtksourceview3` or similar may have to be prepared at some time.

garrigue: There is one!
It is up to date with the released version, lablgtk-3.0.beta3.

ide/configwin.ml
herbelin: If the outcome of the discussion is that the user query and wizard tactics preferences are not worth being preserved and/or that noone is motivated to adapt the code and use the recommanded `tree_store` feature in place of the current `clist`, I guess that the corresponding code, become dead, can indeed be removed completely.

Otherwise said, until there is some visibility on what this code will become, I preferred to keep it in the files at the current stage and to ask ourselves this question again before merging.

coqide.opam
garrigue: And there is a conf-gtksourceview3 too.

herbelin: Ah, ok!

Then, I upgraded the opam file (anticipating on the release number 3.0.0).

herbelin:
@Zimmi48: I feel rather powerless to answer your questions about Debian. I don't have a clear picture of how spread is Debian among Coq users nor how these Debian users install Coq. I don't have a clear picture of how important it is for Coq users in general to use the last version. I don't know how much opam or nix is used for installing.

I don't know why Coq is so outdated in Debian. If it is deliberate from Debian maintainers, it would be better to have a header on the [Coq Debian page](https://packages.debian.org/sid/coq) telling how to install Coq without Debian rather than giving the impression to Debian users that Coq is "dying" and to Coq users that Debian is "dying". Actually, Debian maintainers for OCaml have served for so long time that I would not be surprised if they say that they would be happy that a new team take over (and actually Ralf T. who is one of the maintainers, told me something like that). The coming freeze could be a good opportunity to upgrade testing at least to 8.8, if not to 8.9.

Yes, the ability to compile CoqIDE via gtk3 on the next Debian stable is already a, or even the, crucial step.

Note that opam 2.0.1 seems to be already in Debian [testing](https://packages.debian.org/testing/opam).

About releasing CoqIDE separately, it seems actually that this is the [case](https://packages.debian.org/buster/coqide) in Debian. So, maybe, it is enough to have a package named, say, coqide-on-gtk3-8.8.2???

herbelin:
@jashug: Thanks for feeback.

> I found it useful that an incomplete Qed was styled differently, but am not attached to the stippling effect; I expect a solid light blue/grey background would work just as well for me.

Maybe can I just add a new color configuration item for incomplete Qed, with default solid light blue or grey background?

A digression: I heard that black background are recommended for the eyes. Would it be worth to develop a default style conceived on top of a black background? (Not that I'm competent though)

gares:
Wrt QED any colour would do.

gares:
In Debian the source package coq generates many binary packages among which there is coqide.

Zimmi48:
>I don't know why Coq is so outdated in Debian. If it is deliberate from Debian maintainers, it would be better to have a header on the Coq Debian page telling how to install Coq without Debian

Even if it's not deliberate but is just a lack of commitment, this would be preferable. I don't get what makes Debian maintenance so hard. People managing other package managers are pretty capable of keeping up.

>About releasing CoqIDE separately, it seems actually that this is the case in Debian.

Releasing separately and having separate packages are two different things. We can generate many packages from a single release. This is the case in Debian, but also in opam, and someday it will probably be the case as well in Nix. However, having separate release cycles cannot be done while keeping the source of CoqIDE in the Coq repository, because this means that the version of CoqIDE and Coq are tied up together. In particular, what I was talking about was to release a CoqIDE version that would be compatible with lablgtk3 and a range of Coq versions.

gares:
On my side it is just lack of time, maintenance is not very hard. The deb package is in good shape, afaict (wrt recent policies). I think I should ask for removal unless someone steps up, but I did not ask Stephan, maybe he has à différent opinion.

herbelin:
> Even if it's not deliberate but is just a lack of commitment, this would be preferable. I don't get what makes Debian maintenance so hard. People managing other package managers are pretty capable of keeping up.

Someone from Debian could tell more. Possible hypotheses which maybe are completely wrong are: 1) being a Debian maintainer is demanding and involves a high responsibility: there is continuous integration made on several platforms which you have to watch, you have to follow closely the bug trackers of the corresponding packages, etc. 2) when the ratio pleasure/duties goes too low, it is more difficult to be motivated, and this ratio tends to naturally decrease with routine 3) Debian is somehow now more an institution than a fresh innovative project and people liable to take over happen to be more interested for instance in Nix or opam.

> In particular, what I was talking about was to release a CoqIDE version that would be compatible with lablgtk3 and a range of Coq versions.

Then, I'm not sure exactly how to do this with the current dependencies. For instance, there are currently 3 shared directories, `lib`, `clib`, `protocol`. At first view, we would have to versionize the files in `protocol` to know if a given `coqide` is compatible with a given `coqtop`.

herbelin:
Sorry @gares, I did not have seen your message which was answering my speculations.

> On my side it is just lack of time, maintenance is not very hard. (...)  I think I should ask for removal unless someone steps up (...)

So, let's start from this assumption, i.e. that it is either finding someone to take over or removing the package. Then, how to decide?

What I have difficulties to evaluate is the importance of being present in Debian. In the absolute, if we except the question of lack of time, would it better for the "community" to be in or better not? It seems to me that the question of how to organize ourselves at the level of the community so that the "community" gets what it needs should only come afterwards. In particular, if it makes sense to be in, I'm pretty sure that we shall find someone in the community to take over.

gares:
I agree with @Zimmi48 that distributing binaries using for example snap is the best option.

ejgallego:
@herbelin I did a complete distribution using https://github.com/garrigue/lablgtk/pull/38 , however I get:

```
File "ide/wg_ScriptView.ml", line 487, characters 38-65:
Error: This expression has type Pango.font_description
       but an expression was expected of type GPango.font_description
      ocamlc ide/.gui.objs/wg_ProofView.{cmo,cmt} (exit 2)
```

is that an error on my part?

ejgallego (approved):
I think this looks good, tho we may want to include both the GTK2 and GTK3 versions of CoqIDE for 8.10 [I could help with that], effectively freezing the gtk2-based one.

IMHO we should first sort out Dune+lablgtk 3, then merge this.

herbelin:
@ejgallego: I (hopefully) fixed the `Pango` error that you mentioned [here](https://github.com/herbelin/github-coq/tree/master%2Bcoqide-on-lablgtk3%2Bcairo-from-claudio).

At least, it works using `dune build bin/coqide`. Note that this fix is for the current head of the lablgtk3 branch (integrating CSC's cairo bindings merged after the beta3 release of lablgtk3). It would not work with the beta3 release of lablgtk3.

ejgallego:
Thanks Hugo, I think I'll aim to merge soon and add CI to the lablgtk repos too.

ejgallego:
I don't see the push to the branch tho @herbelin

ejgallego:
Also @herbelin, would you mind giving me a heads up on what's the way you use to install gtk in OSX [so we can setup the OSX CI in upstream properly]

herbelin:
I rebased the current PR. The extra commit for Cairo compatibility is [here](https://github.com/herbelin/github-coq/tree/master%2Bcoqide-on-lablgtk3%2Bcairo-from-claudio).

I don't remember the exact steps for MacOS X. I'm using brew and I probably just did:
- `brew install gtk+3`

ejgallego:
Thanks a lot @herbelin , I think we will target the `master` branch of lablgtk for this PR.

@ppedrot , would you mind helping here by being a co-assignee? I will help @herbelin fixing the build / CI setup, so as it will have some commits from me maybe it'd be better you merged after @coq/ci-maintainers give the blessing to my changes.

ejgallego:
So far it seems that even upstream the only working setup for building lablgtk3 in OSX for example is using Dune, so I'll base our setup on that [for now]

I may also add a job for building CoqIDE vs Coq 8.9

ejgallego:
So @herbelin I'm taking the lock to fix the build setup.

ejgallego:
@herbelin , this should compile as pass CI, however, both in `make` and `dune` builds, I cannot start CoqIDE, I get:
```
Fatal error: exception Failure("Gobject.unsafe_create : type GtkProgressBar is not yet defined")
```

ejgallego:
So indeed it looks like coqide now needs `linkall`

ejgallego:
So this seems ready to go!

I guess we should wait for the merge and OPAM update of garrigue/lablgtk#38 , but we could also bundle it.

dev/build/windows/makecoq_mingw.sh
```diff
@@ -837,7 +845,7 @@ function make_gdk-pixbuf {
   # CONFIGURE PARAMETERS
   # --with-included-loaders=yes statically links the image file format handlers
   # This avoids "Cannot open pixbuf loader module file '/usr/x86_64-w64-mingw32/sys-root/mingw/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache': No such file or directory"
-  build_conf_make_inst  http://ftp.gnome.org/pub/GNOME/sources/gdk-pixbuf/2.36  gdk-pixbuf-2.36.12  tar.xz  true  --with-included-loaders=yes
+  build_conf_make_inst  http://ftp.gnome.org/pub/GNOME/sources/gdk-pixbuf/2.8  gdk-pixbuf-2.38.0  tar.xz  true  --with-included-loaders=yes
```
MSoegtropIMC: This should be gdk-pixbuf/2.38 (a atypo)

MSoegtropIMC:
@herbelin : It looks like you are building a lot of GTK stuff from sources. Please note that building GTK from sources was required in the old days before cygwin had a GTK package. I kept the GTK 2 stuff mostly because I am not sure if cygwin will one day drop GTK 2 support and also I found it sometimes useful for debugging to be able to build this from sources. But I wanted to ask if your changes are intended to keep the option to build GTK from sources alive or if these packages really do not exist in cygwin?

Also please let me know if you need any help with the Windows build.

Zimmi48:
I believe that even though the author of the commits appears to be @herbelin, the Windows scripts were in fact modified by @ejgallego.

ejgallego:
Indeed the commit tweaking builds files is mine and properly attributed I think

ejgallego:
@MSoegtropIMC indeed I was following what is done today, which seems like rebuilding gtk2, right?

I don't mind, we can do in any way to install gtk3; we can indeed drop the gtk2 support at all for 8.10.

dev/build/windows/makecoq_mingw.sh
ejgallego: Thanks!

ejgallego:
@MSoegtropIMC I am unable to determine why glib fails, is really the glib package broken?

I'll to revert that part.

MSoegtropIMC:
> I'll to revert that part.

Since you already spent some time with it, I would prefer to keep it. But building from sources shouldn't be the default anyway. And building GTK from sources always needed a little bit of patching cause of incompatibilities between Linux and Cygwin.

If you want, I can look into this some time during this week.

ejgallego:
> If you want, I can look into this some time during this week.

That would be great @MSoegtropIMC , for some reason the updated dependencies don't work well. I am neutral as how to build GTK on windows. Most people seem to use `msys2` these days, but I am open to any option as long as we get a working `pkg-config`.

On other news, thanks to the hard work of @garrigue and @herbelin new beta packages have been uploaded to OPAM.

So except for the Windows part [where I wasn't able to build the deps with our scripts] this PR is ready for both Linux and OSX.

I suggest to merge the PR soon, and fix the windows build later on; IMHO it is of critical importance to start the testing now in case we deem the gtk3 port not ready and needing a revert for 8.10.

I don't think the risk is significant for windows users, CI for Windows lablgtk should be in place soon, but in the worst case we could indeed ship the gtk2 version for Win.

Zimmi48:
Do you mean that the PR in the present state does not break the CI even if the Windows support is not ready?

ejgallego:
> Do you mean that the PR in the present state does not break the CI even if the Windows support is not ready?

I mean we would disable building of CoqIDE in the Windows CI for now.

ejgallego:
The thing is that I cannot understand what's going on in the CI scripts, things fail in very strange ways; I can build lablgtk3 in my windows VM, should do some final testing soon to be sure. But IMHO if that's the case then I'd rather merge as testing of this early is urgent and leave the Windows expert to build gtk3 in the way they'd like.

Zimmi48:
If I understood @MSoegtropIMC correctly, he was suggesting to not build gtk3 at all but instead install it through cygwin... In theory, this should be much simpler. And if building from sources did not work, there is no reason to insist if there is a binary available.

ejgallego:
> If I understood @MSoegtropIMC correctly, he was suggesting to not build gtk3 at all but instead install it through cygwin... In theory, this should be much simpler. And if building from sources did not work, there is no reason to insist if there is a binary available.

This is a significant change beyond the scope of this PR, as currently the GTK2 system doesn't do that, but instead seems to build everything from source.

MSoegtropIMC:
> This is a significant change beyond the scope of this PR, as currently the GTK2 system doesn't do that, but instead seems to build everything from source.

This is not true. There is an option to build GTK from sources, and by default it is (or should be) off.

From makecoq_mingw.sh (a fairly recent version):

```
function make_gtk2 {
  # Cygwin packet dependencies: gtk-update-icon-cache
  if [ "$GTK_FROM_SOURCES" == "Y" ]; then
    make_glib
    make_atk
    make_pango
    make_gdk-pixbuf
    make_cairo
    build_conf_make_inst  http://ftp.gnome.org/pub/gnome/sources/gtk+/2.24  gtk+-2.24.32  tar.xz  patch_gtk2
  fi
}
```

Here $GTK_FROM_SOURCES defaults to N.

The batch file contains:
```
IF "%GTK_FROM_SOURCES%"=="N" (
  SET CYGWIN_OPT= %CYGWIN_OPT% -P mingw64-%ARCH%-gtk2.0,mingw64-%ARCH%-gtksourceview2.0
)
```
To add the cygwin binary packages in case GTK_FROM_SOURCES is N.

ejgallego:
Umm, indeed thanks @MSoegtropIMC , I am quite confused with the script it seems. Let me try again.

ejgallego:
Oh I see what happened, I removed the gtk2 build and lost the check.

MSoegtropIMC:
For the time being I would suggest to ignore this option or throw an error in case it is Y and always add the corresponding cygwin packages - hoping that they exist. In case they don't exist, please let me know, than I will try to fix the GTK3 sources build.

But please let the GTK 3 sources stuff in (possibly commented out) since you already put quite some work in there and for certain debugging stuff it is better to build GTK from sources (which is one of the reasons why this option exists). The other reasons why this options exists are:

- When I did this there was no cygwin GTK package
- I am afraid that GTK 2 will be discontinued some time in cygwin

These two reasons become obsolete with moving to GTK 3 (assuming the packages exist), but the CoqIDE debugging build reason remains.

ejgallego:
Yup, we'll do as you suggest, right know the from sources build is broken but that should be easy to fix.

Let's see if the lablgtk build goes OK.

ejgallego:
I'm afraid that I have reached my cycle limits for this, the OSX and Linux builds are OK, the Windows build needs a bit of work.

MSoegtropIMC:
> I'm afraid that I have reached my cycle limits for this, the OSX and Linux builds are OK, the Windows build needs a bit of work.

Can you write a short (5..10) lines wrap up on the status is (what you have tried, and what the issues have been)? I can take over then.

ejgallego:
> Can you write a short (5..10) lines wrap up on the status is (what you have tried, and what the issues have been)? I can take over then.

- attempt with the CI setup: flexlink fails to understand linking options.
- attempt with my VM: file `gdkwin32` not found

thanks @MSoegtropIMC

MSoegtropIMC:
I guess this is with the binary GTK3 package from cygwin?

ejgallego:
> I guess this is with the binary GTK3 package from cygwin?

Yes, so maybe the file is called `gdkwin64` now?

herbelin:
Note: For the record, I just learned that Debian [reverted](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=885677) the removal of gtksourceview2 in next stable (["buster"](https://wiki.debian.org/DebianBuster#Debian_Buster_Life_cycle)/Debian 10).

Zimmi48:
From what I got from the interaction with @bbarenblat, CoqIDE had been removed from Debian nonetheless. It would be great to make sure that it is brought back in Buster (in version 8.9.0) if it is not too late.

ejgallego:
IMHO we should take a decision now on what to do for 8.10, maybe @vbgl wants to weight in?

The PR seems ready to me, except for windows support that I'm afraid that windows maintainers will have to fix. That shouldn't be hard, but it has to be done.

I personally support merging the patch, as the cost of it going stale seems larger than the cost of doing a work that anyways it has to be done. Meanwhile, the windows build can be fixed later or still ship the GTK2 version. Building CoqIDE "2" apart from Windows is still easy. I've setup a tree here that the windows build could use: https://github.com/ejgallego/coqide2-split [tested to work]

Zimmi48:
>Meanwhile, the windows build can be fixed later or still ship the GTK2 version. Building CoqIDE "2" apart from Windows is still easy. I've setup a tree here that the windows build could use: https://github.com/ejgallego/coqide2-split [tested to work]

I'm pretty sure the Windows build problem can be fixed, so that doesn't seem like a blocker. However, if we want to preserve the 2.x version to ensure compatibility and do so in a split tree, then it makes even more sense to do the split for both the 2.x and the 3.x version, with a branch for each, and to have CI test both versions. For convenience, this would require killing right away make support for CoqIDE and depend entirely on Dune.

ejgallego:
> I'm pretty sure the Windows build problem can be fixed, so that doesn't seem like a blocker. However, if we want to preserve the 2.x version to ensure compatibility and do so in a split tree, then it makes even more sense to do the split for both the 2.x and the 3.x version, with a branch for each, and to have CI test both versions.

We can have the gtk3 version on the tree  and the gtk2 version out of the tree but on CI. I personally don't mind either. Anyways CI is not very useful for CoqIDE as it only builds it, but we have no tests. [The fake_ide tests don't test coqide but coqidetop]

> For convenience, this would require killing right away make support for CoqIDE and depend entirely on Dune.

This saves some effort indeed, but note that the build support is entirely findlib based so we can keep the makefiles without too much hassle too.

Anyways I'm OK with any plan that allows us to merge this PR soon as it is critical we have testing. So I dunno when is the moment to merge this. CI-wise it is ready [but no windows support for gtk3].

ejgallego:
Indeed I propose @vbgl to be the final assignee for this PR.

MSoegtropIMC:
I will give the Windows build a try this evening. I would prefer to do this as a separate PR (the review process is cleaner then), but don't have a strong opinion about this and would follow your preference.

ejgallego:
> Indeed I propose @vbgl to be the final assignee for this PR.

@vbgl do you agree?

herbelin:
No strong opinion either on the layout. Keeping the labgltk2 version in another repository or making the additional work of supporting both the lablgtk2 and lablgtk3 versions in the same archive are two options ok to me.

MSoegtropIMC:
I had a look at the lablgtk3 Windows build issues. At least the apparent flexlink error should be easy to fix but I am fighting a bit with the dune and sexp stuff. If this is the only issue, it should work tomorrow.

MSoegtropIMC:
I need help from a dune expert, maybe @ejgallego can help?

I would say the issue is the option `"-Wl,-luuid"` in `\build\lablgtk-3.0.beta4\_build\default\src\clink-gtk+-3.0.sexp`. I think the option should be either `"-ldopt "-link \"-Wl,-s\""` or just `"-link -Wl,-s"` so that flexlink forwards it to the system linker (see my comments in the lablgtk2 build in an old makecoq_mingw.sh). But the sexp file is auto generated on each build and I couldn't find out which magic leads to the effect that these options are in there. I searched the lablgtk and dune sources for uuid without success. So I don't know what to change in order to change these options.

ejgallego:
Thanks a lot @MSoegtropIMC ! I suggest we open an issue in the lablgtk3 repository https://github.com/garrigue/lablgtk ; the flags in this case come from `pkg-config` itself, thus it could be a gtk3 bug?

ejgallego:
The file is autogenerated by calling `dune_config "gtk+3.0 >= 3.18"` in the `dune` file, which basically calls `pkg-config`:
```
(rule
 (targets
   cflag-gtk+-3.0.sexp
   clink-gtk+-3.0.sexp)
 (action (run ./dune_config.exe -pkg gtk+-3.0 -version 3.18)))
```

ejgallego:
@MSoegtropIMC https://github.com/garrigue/lablgtk/issues/52

MSoegtropIMC:
> The file is autogenerated by calling `dune_config "gtk+3.0 >= 3.18"` in the `dune` file, which basically calls `pkg-config`:

Ah, ok that does the trick! Let me see if I can fix the build with this extra information.

One could also discuss if flexlink should detect more gnu linker command line commands and pass them through. I can't say I understand how flexlink works, but as far as I understand it tries to be command line wise compatible with the gnu linker, so that one doesn't see a difference between Linux and Windows. It does a fairly good job at this, but here it fails. I would think that flexlink just has a list of options it passes through to ld and I guess one could also fix this by adding these options to the pass through options list of flexlink. But this is just guessing from observation from a large distance without really understanding what is going on.

ejgallego:
> Ah, ok that does the trick! Let me see if I can fix the build with this extra information.

Great, thanks for the feedback @MSoegtropIMC .

IMVHO we should try to address this in such a way that we don't require any specific patching on the Coq side; both the maintainers of flexlink and lablgtk will be supportive, so I recommend we discuss with them what would the right choice be here.

MSoegtropIMC:
> IMVHO we should try to address this in such a way that we don't require any specific patching on the Coq side; both the maintainers of flexlink and lablgtk will be supportive, so I recommend we discuss with them what would the right choice be here.

Sure. I sent an email to Alain Frisch asking him to comment here on his opinion how this should be fixed.

ejgallego:
> Sure. I sent an email to Alain Frisch asking him to comment here on his opinion how this should be fixed.

Great thanks! AFAICT flexdll has a github repos so if you wanna open an issue there ping me and I'll follow it.

alainfrisch:
There is already some ad hoc logic in flexdll to preprocess the command line (https://github.com/alainfrisch/flexdll/blob/5f0011a3db1f3a499eeaaae2254e806cbe6b8035/cmdline.ml#L245-L258).  It wouldn't be hard to make it so that options like `-Wl,...`  are automatically prefixed with `-link`.  To avoid multiple iterations, it might be better if people here could try with a locally patched flexdll and send a PR.

ejgallego:
Thanks @alainfrisch , it makes sense indeed.

MSoegtropIMC:
> There is already some ad hoc logic in flexdll to preprocess the command line

@alainfrisch : thanks for the quick confirmation that it might make sense to fix this issue with a flexlink enhancement. I will see if I can patch flexdll so that the lablgtk3 build goes through without issues and send you a patch.

MSoegtropIMC:
@alainfrisch : I have one understanding question on flexlink: should it command line wise be compatible to gcc or to ld? I guess gcc? Or is it a misunderstanding that flexlink works as a wrapper for one of these and should be command line compatible for common cases?

alainfrisch:
flexlink is not designed to have a command-line compatible with either gcc, ld, cl, or link.  It just happens to be "compatible enough" to be easily used by the OCaml compiler.

MSoegtropIMC:
@alainfrisch : I see. Let me reformulate my question: what does the ocaml compiler / ocaml librarian call on Linux in the situations it calls flexlink on Windows? Is it gcc or ld or does it depend on the situation?

In order to fix the lablgtk build issues (which go back to rather strange pkg-config linker options supplied by GTK) I would need to add support for one gcc style command line options to flexlink. This is easy to do in the translator function you pointed me to.

MSoegtropIMC:
My summary from the Windows front: There are two issues:

- GTK gives a strange linker option in pkg-config. For libuuid it prefixes the -luuid option with a gcc linker path through prefix, which is fairly useless since the -lxxx options are anyway passed through by gcc. So the best fix would be to remove this gcc prefix from the linker options provied by GTK. Possibly there is some reason for doing this. It might be that there exists a universe in which -luuid means something else to gcc than "link libuuid" so that the linker forwarding prefix is required. A fix in GTK likely takes a while to trickle down (via cygwin), so for the time beeing I would patch the pkg-config file of gtk, but of cause I will trigger a fix upstream. An alternative fix would be to make flexlink understand the gcc linker pass through prefix, which is easy to do. I guess best would be to have both fixes.

- The labgtk library needs the same changes as I did for version 2 (see the patch file). The reason this is required is that on Linux some functions take socket or file handles, while on Windows one needs to call different functions for socket handles and file handles (at least if I remember right, it has been a few years). The reason I didn't push this change upstream is that it is a hack. It survived in CoqIDE for years without resulting in any issues, but it is still a hack. A real fix would require a medium size interface design change in lablgtk regarding socket and file interfaces. So I would suggest to also fix this in Coq with a patch rather than pushing a hack with limited usefulness for others upstream.

ejgallego:
Thanks @MSoegtropIMC for the very detailed summary.

> * The reason I didn't push this change upstream is that it is a hack. It survived in CoqIDE for years without resulting in any issues, but it is still a hack. A real fix would require a medium size interface design change in lablgtk regarding socket and file interfaces. So I would suggest to also fix this in Coq with a patch rather than pushing a hack with limited usefulness for others upstream.

On the other hand I was able to compile and run `CoqIDE` in the GTK2 version without this patch, so does that mean that my compiled version of `CoqIDE` had some kind of bugs?

IMHO we shouldn't keep this patch private to Coq, this'll be laughs for today but tears tomorrow. I would submit the path as a pull request upstream and we can discuss how to properly integrate it.

Lablgtk3 is in beta so this is the perfect moment to improve cross platform support IMHO.

MSoegtropIMC:
> On the other hand I was able to compile and run `CoqIDE` in the GTK2 version without this patch, so does that mean that my compiled version of `CoqIDE` had some kind of bugs?

I need to check - afaik there is also a patch in CoqIDE which uses modified functions on Windows, so it should not compile, but I forgot the details. As I said it is a while back. But essentially one can compile CoqIDE without this change. It will do interesting things then if file handles are passed to functions taking sockets and vice versa.

> Lablgtk3 is in beta so this is the perfect moment to improve cross platform support IMHO.

Indeed a good argument. In the lablgtk2 times I refrained from suggesting largish changes which would break compatibility, but here this argument does not count. Let me look into this. Still I would first just hack it for CoqIDE and do a proper patch later, unless it is less work than I expect right now.

ejgallego:
Thanks @MSoegtropIMC , maybe you could submit the current patch to https://github.com/garrigue/lablgtk as a PR, so we could discuss it. Many things have changed in the OCaml world and Win32 so it'd be good to discuss first and allow other experts to weight in.

ejgallego:
Note that the right branch in the upstream repos is `lablgtk3`

alainfrisch:
> Let me reformulate my question: what does the ocaml compiler / ocaml librarian call on Linux in the situations it calls flexlink on Windows? Is it gcc or ld or does it depend on the situation?

It's gcc.  AFAICT, OCaml will directly call ld only for "ocamlopt -pack" and "ocamlopt -output-obj" (i.e. to link into "compound" object files, not images), and this is done directly through link.exe, not flexlink.exe under Windows.

MSoegtropIMC:
@alainfrisch : thanks for the clarification! I will then add the gcc linker forwarding command prefix to flexlink.

MSoegtropIMC:
The Windows build is basically working, but I don't see the toolbar icons. Does someone know how the toolbar stuff does work? Do I need to copy some files? I don't think I had icon files for GTK2 - I always thought the icon files where xpm files compiled as strings into the C code.

I already adjusted all the DLL and GTK file copying scripts.

MSoegtropIMC:
Are there plans to merge this soon?

ejgallego:
> The Windows build is basically working, but I don't see the toolbar icons. Does someone know how the toolbar stuff does work? Do I need to copy some files? I don't think I had icon files for GTK2 - I always thought the icon files where xpm files compiled as strings into the C code.
> 
> I already adjusted all the DLL and GTK file copying scripts.

That's fantastic @MSoegtropIMC ! I am looking forward to add Windows CI to lablgtk so we can be sure it keeps working in the future.

I am not sure how the icons work, maybe @herbelin or @garrigue know?

MSoegtropIMC:
> I am looking forward to add Windows CI to lablgtk so we can be sure it keeps working in the future.

How shall we do this technically? Shall I commit to this PR or will this PR be merged soon, and I do a follow up PR for windows? I don't mind and can do it either way, but until Windows is final, it might take a few more iterations. E.g. I need to test the 32 bit build as well and also want to re-establish the GTK from sources build. But it would also be an option to first integrate only 64 bit binary GTK and do the rest later in a separate PR. I think how we proceed mostly depends on how quickly you want to integrate this PR. If you can wait say a good week, I would commit to this PR.

ejgallego:
> Are there plans to merge this soon?

Given the lack of response of the release manager I'd say yes, let's merge this soon. IMVHO before merging we'd like to:
- make the Win64 CI job build: please push a commit fixing it to this branch; I'll rebase as appropiate,
- add CI support for Windows on the lablgtk repos, this is IMHO important to be sure we don't introduce regressions in the next betas.

Both items should be doable in reasonable time I'd dare to say. WDYT @MSoegtropIMC ?

MSoegtropIMC:
> add CI support for Windows on the lablgtk

Should I take care of this?

ejgallego:
> Should I take care of this?

Sure @MSoegtropIMC any help would be really appreciated, thank you !

We have an open issue at https://github.com/garrigue/lablgtk/issues/52

vbgl (dismissed):
Please fix the `default.nix` file so that CoqIDE is still built. This should amount to changing the occurrence of `lablgtk` into `lablgtk3`. Thanks.

ejgallego:
> Please fix the `default.nix` file so that CoqIDE is still built. This should amount to changing the occurrence of `lablgtk` into `lablgtk3`. Thanks.

Thanks @vbgl for the pointer, I'm rebasing right now to try.

ejgallego:
@vbgl we'll have to update the nixos ref, CoqIDE needs beta4 and so far the current ref points to beta3.

They are not API compatible so this is a hard requirement.

vbgl:
You can try this ref then: b275d092249b6e2899d4bf93a87d69a41b965aae
And if you need sourceview, you should change `lablgtk3` into `lablgtk3-sourceview3`.

ejgallego:
> You can try this ref then: b275d092249b6e2899d4bf93a87d69a41b965aae
> And if you need sourceview, you should change `lablgtk3` into `lablgtk3-sourceview3`.

Thanks @vbgl , pushing an updated `default.nix`

vbgl:
Here is a way to get the correct hash for an *unpacked* tarball:
~~~
nix-prefetch-url --unpack https://github.com/NixOS/nixpkgs/archive/b275d092249b6e2899d4bf93a87d69a41b965aae.tar.gz
~~~

ejgallego:
> Here is a way to get the correct hash for an _unpacked_ tarball:

Thanks @vbgl , unfortunately there is no proper way to install Nix in my system [Ubuntu] , I'd appreciate if someone could tell me what to write exactly on the Nix file [or pushed to this branch]

ejgallego:
The build went thru! Thanks @vbgl !

@Zimmi48 or yourself may want to actually test whether the thing works. Also, updating to cairo2 0.6 may be useful.

MSoegtropIMC:
I tried to push my Windows changes to this PR but I don't understand how to do this. I tried to follow <https://help.github.com/en/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork> but failed at the first step "navigate to the main page of the fork". I looked at the github page of @herbelin but see only a relatively old fork coq-hh, which doesn't include this branch.

@ejgallego : you apparently managed to do this. Can you please explain how?

ejgallego:
> @ejgallego : you apparently managed to do this. Can you please explain how?

The fork is called `github-coq`, concretely 
`https://github.com/herbelin/github-coq.git master+coqide-on-lablgtk3`

MSoegtropIMC:
> The fork is called `github-coq`, concretely

sorry, this was silly - I only saw the "popular repositories" - not the full list.

I merged my stuff and do a local test run now.

How shall we handle the icon issue? Ignore it for the time being? CoqIDE works (in some quick test), just for all toolbar icons some default icon is shown. It might take me a while to figure out how toolbars and icons work in GTK so that I can fix this. One imaginable cause is that the icons are e.g. XPM and the image driver for XPM has been removed from the default image driver list in GTK3 (this is just a speculation).

ejgallego:
> I merged my stuff and do a local test run now.

Great @MSoegtropIMC , thanks a lot!

> How shall we handle the icon issue?

Would it be easy to do a very simple test case to reproduce the issue? If so, I'd suggest we do that an open an issue in the GTK3 repository.

MSoegtropIMC:
> Would it be easy to do a very simple test case to reproduce the issue? If so, I'd suggest we do that an open an issue in the GTK3 repository.

Not for me - I only used wxWidgets and Qt for my own apps in the past, but never GTK. But I have quite some experience with GUI apps and the frameworks are all somehow similar, so I think I can fix this at some point if nobody else does, just not right now.

Do the toolbar icons work in Linux?

ejgallego:
> Do the toolbar icons work in Linux?

They seem to. Note that I've rebased, attention when pushing.

MSoegtropIMC:
I pushed my changes (after a rebase - thanks for the hint!)

Regarding the two patches:

- the flexdll patch I will push upstream as soon as we agreed on the change here. The patch adds a bit more gcc command line compatibility to flexlink
- the lablgtk change I need to review first, remember why it was required in lablgtk2 and come up with a non hack change proposal for GTK3. As discussed I agree that we should look into getting this right in the lablgtk3 beta. Please note that this is a Windows only patch (the code is ifdefed away on Linux) which explains why you can build on Linux without this patch.

ejgallego:
Amazing, thanks Michael. Jacques will release beta5 likely today, so indeed, if CI goes well, and Vincent doesn't veto this, I will propose to merge this soon.

MSoegtropIMC:
> Amazing, thanks Michael.

I say thank you all!

One more note for the records: the lablgtk2 build was a source of occasional build failures of the Windows build (about 5..10% failure rate) - sometimes flexlink went wild when linking lablgtk2. The fix was to strip the lablgtk2 library by passing appropriate linker commands (stripping it after building also didn't work reliably). This is still in the lablgtk2 patch. I didn't add this to the lablgtk3 patch because up to now it seems to be reliable. But please watch out and let me know in case anything looks like the old lablgtk2 + flexlink issue. It should be easy to fix, but I might need some dune expert support.

MSoegtropIMC:
The Windows64 build went through and I did some quick test with the CoqIDE from the resulting installer. Some observations:

- icons don't work (known)
- file open does not work (crashes)
- Pasting a coq file and going through it step by step and to end works - proof state and highlighting look ok.
- If forward one line is done (either via toolbar or keyboard) the screen is scrolled to the end of the file rather than the current proof position

This was just a three minutes quick test - I don't have more time right now.

Can someone please test which if these issues also exist on Linux?

vbgl:
I observe similar behavior on NixOS (linux):

  - no icon.
  - stepping through an unsaved buffer seems to work; highlighting is random (but this is a feature…)
  - crash on file open / save

> 0x00007ffff71f5423 in g_log_structured_array () from /nix/store/6i3zqxizaliin2j4ag7fh7adwywhmhch-glib-2.58.2/lib/libglib-2.0.so.0

MSoegtropIMC:
> I observe similar behavior on NixOS (linux):

In this case I would suggest that someone with more experience in Ocaml GTK debugging than me has a look at this on Linux first.

Btw.: Does GTK have an event playback + window screen shot mechanism, so that one could do at least a little test suite - say load a file, step through it and take a screen shot every now and again and compare these.

vbgl:
Is CoqIDE on gtk3 working for someone around?

ejgallego:
I did some more extensive testing on the past. In Ubuntu 16.04 I get:

> * no icon.

I get icons.

> * stepping through an unsaved buffer seems to work; highlighting is random (but this is a feature…)

Highlighting seems to work here.

> * crash on file open / save
Works here.

GTK 3.22.30 / Sourceview 3.24.7

What is your gtk version folks, that is usually critical.

ejgallego:
Noting that I am running the locally-build copy with `dune exec coqide`.

MSoegtropIMC:
Since about 1 week (Feb 20th) the MinGW versions in Cygwin are:

- GTK: gtk3-3.22.28-1
- GTK: gtksourceview3.0-3.24.6-1

See

- <https://cygwin.com/packages/x86_64/mingw64-i686-gtk3/>
- <https://cygwin.com/packages/x86_64/mingw64-i686-gtksourceview3.0/>

This is slightly older than what you stated above (the -1 is the cygwin patch level on top of the source version). In case newer versions are required we could either try to update in cygwin or build it (for the time being) from sources. Note that for GTK2 it took about one hour to build GTK and its requirements, but we might get the requirements in binary and just build GTK3 from sources.

Can you easily test if indeed it does not work with the above versions (or do you already know?)

ejgallego:
> Can you easily test if indeed it does not work with the above versions (or do you already know?)

IMHO these versions should work just fine; I dunno what the problem is.

MSoegtropIMC:
@vbgl : can you please report the GTK / gtksourceview version you used for your NIX tests?

ejgallego:
Note that I am using the dune build, and indeed that could have something to do; @MSoegtropIMC , would you mind trying building CoqIDE with:
```
make -f Makefile.dune world
dune exec coqide
```

vbgl:
gtk+3-3.24.4
gtksourceview-3.24.9

ejgallego:
Versions should be fine, @vbgl were you using a make-built coqide, or a dune built one?

vbgl:
Same crash in both cases.

ejgallego:
Ummm...

> 0x00007ffff71f5423 in g_log_structured_array () from /nix/store/6i3zqxizaliin2j4ag7fh7adwywhmhch-glib-2.58.2/lib/libglib-2.0.so.0

Is there a possibility you could get a complete backtrace?

vbgl:
Backtrace: https://gist.github.com/vbgl/b84549f04834143304d9edbbd540a757

ejgallego:
Thanks @vbgl cc @garrigue

MSoegtropIMC:
The file open bug I can fix. Apparently there is a file missing, which can be created (in the windows case) via:

    glib-compile-schemas /usr/x86_64-w64-mingw32/sys-root/mingw/share/glib-2.0/schemas/

from the build cygwin console. The resulting `gschemas.compiled` file needs to be copied to

    <coqinstallroot>\share\glib-2.0\schemas

One can argue that this is a bug in the cygwin GTK mingw distribution. The corresponding file does exist for the cygwin native GTK.

@vbgl: can you check if something similar helps in the NIX case?

vbgl:
Thanks for the hint. I’ve thus fixed the icon & crash issue.

ejgallego:
Ok, so IMO it is time to pick a decision on this; not a lot more can be done without testing, so I'd say let's merge! We need a reasonable testing time.

The gtk2 version can be kept in a separate repos as pointed out above.

MSoegtropIMC:
I didn't push the open crash on Windows as yet and also didn't check if the NIX fix for the icon issue also works on Windows in a similar way. It might take until Mo.

ejgallego:
> I didn't push the open crash on Windows as yet and also didn't check if the NIX fix for the icon issue also works on Windows in a similar way. It might take until Mo.

Thanks @MSoegtropIMC , we may want to merge even if the setup is not perfect yet tho; this PR is getting gargantuan and reaching the limits of what the Github interface can do; so any fix can be done in a posterior PR if we want.

MSoegtropIMC:
> Thanks @MSoegtropIMC , we may want to merge even if the setup is not perfect yet tho; this PR is getting gargantuan and reaching the limits of what the Github interface can do; so any fix can be done in a posterior PR if we want.

Fine with me. Please create an issue and assign it to me in case this is merged without the fixes.

ejgallego:
> Fine with me. Please create an issue and assign it to me in case this is merged without the fixes.

Thanks, so indeed let's suggest @ppedrot to start looking at merging this, and if he finally processes this PR before you can get to fix the remaining issue I'll create a blocker issue.

MSoegtropIMC:
> and if he finally processes this PR before you can get to fix the remaining issue I'll create a blocker issue.

@ppedrot : please let me know if I should still try to fix this in this PR or if you would prefer that I do this in a separate PR. If I do it in this PR, it might take a few more CI cycles.

ppedrot:
Indeed, I am in favour of merging this now and fixing the small details later on. We can always backtrack in 8.10 if things go awry anyways.

ejgallego:
Sounds good @ppedrot , I've rebased, this should be ready to merge if CI doesn't surprise us.

ejgallego:
> this should be ready to merge if CI doesn't surprise us.

Indeed the windows build is broken again :(

XVilka:
@ejgallego if you check the log, seems problem is unrelated to the PR:
```
cp: cannot stat '/bin/x86_64-w64-mingw32-pkg-config.exe': No such file or directory
```
https://gitlab.com/coq/coq/-/jobs/175299403

ejgallego:
> @ejgallego if you check the log, seems problem is unrelated to the PR:

Yup, I saw it, but this was supposed to be fixed I think. The amount of hacks we have in the Windows build is a bit worrying.

MSoegtropIMC:
> Indeed the windows build is broken again :(

I see that CI is running. Did you do something, e.g. rebase to the recent pkg-config fix to fix this already or is this a new issue?

MSoegtropIMC:
> The amount of hacks we have in the Windows build is a bit worrying.

Can you be a bit more specific? I would say for the complexity it has,, the windows build is fairly clean, and I would also say that the amounts of changes which are required to handle changes in cygwin is very low (less than one per year).

ejgallego:
> I see that CI is running. Did you do something, e.g. rebase to the recent pkg-config fix to fix this already or is this a new issue?

The previous run was already on the fixed version so I dunno but this seems like a new problem.

> Can you be a bit more specific?

Sure, so far we seems to have:

- a somewhat custom build of OCaml,
- a custom build of findlib
- a patched + custom install of lablgtk
- some special patching of a few other binaries
- special build of addons
- plus all the Win code we have in Makefiles etc [much less in the Dune-based build]

MSoegtropIMC:
> Sure, so far we seems to have:
> 
> * a somewhat custom build of OCaml,
> * a custom build of findlib
> * a patched + custom install of lablgtk
> * some special patching of a few other binaries
> * special build of addons
> * plus all the Win code we have in Makefiles etc [much less in the Dune-based build]

Well after all Windows is a different operating system than all of this software is originally designed for and mostly maintained on. For that I would say that the number and complexity of the patches is very small. I find it a bit unfair to say "we have all the Windows specific stuff" in the build. If the builds would have originally be made for Windows, you would have a lot of Linux specific stuff in there. The point is that the systems are just different and this difference is not somehow evil - it is just there. E.g. the patch in lablgtk is required because on Linux file and socket handles are essentially the same type, while in Windows they are not. Of cause one can move this code upstream, but I already explained the rationale for not doing so (it is a hack of limited usefulness).

Of cause if Dune helps to substantially reduce differences and/or maintenance effort, yes let's switch. But I don't see how Dune can help to e.g. bridge differences in handle semantics between Linux and Windows. Where Dune might help is to handle CRLF / LF differences (this is the reason why I compile a patched sed) and the reason for quite a few patches here and there.

Regarding OCaml: I would say that the script pretty much follows the OCaml Windows install readme (or at least that was the case when I did it and the same procedure apparently still works). The instructions are just put into an automated script. I don't see in how far this is non standard. Actually most of the stuff in the script is just automated Readme instructions. Opam is conceptually not very different. Maybe it is now usable for MinGW cross builds, when I originally wrote the scripts this was not the case.

In the end it is all about minimization of effort for keeping it running. I think the effort is comparatively low. Maybe there are methods which are even lower effort, but first this has to be proven and second one has to see if the effort of switching is at least comparable to the maintenance effort one saves.

Zimmi48 (commented):
FTR the opam file requires the >= beta5 version, the README doesn't make any mention of a beta version, and the Nix file uses a version of nixpkgs that contains the beta4 version of lablgtk3.

MSoegtropIMC:
@ejgallego : regarding the Windows build: it fails in

    dune build cairo2.install

Can you have a look at the log:
<https://coq.gitlab.io/-/coq/-/jobs/176113751/artifacts/artifacts/buildlogs/ocaml_cairo2-0.6-dune-build_err.txt>
There is this in the log:
```
find_in_path: pkg-config

-> not found
```

Possibly this is an issue with the pkg-config changes of cygwin. pkg-config used to be an executable but it is a shell script now. This change has been done in most Linux distros as well recently but there it doesn't make much of a difference because both are executable and have the same name. On Windows one can't execute shell scripts and executables in the same way and also the shell script lacks the .exe extension, so possibly dune has issues with this.

Btw.: This would be solved by my "pico shell" replacement for the OCaml system command, which does check at least common she-bangs.

Another solution would be to create a simple executable version of the shell script (it just adds a parameter to the command line and then calls an excutable).

ejgallego:
Hi @MSoegtropIMC , unfortunately I lack the resources to debug this now [I don't have Windows machine at hand] Indeed that would be well the problem, `dune --verbose` should print a lot of debug information.

We use Dune's built-in support for pkg-config so there is not a lot we can do here. Tho, it is strange that it says "not found", instead of "cannot exec" or something like that, but I do agree that the move to a shell script looks very likely to be the culprit.

Dune may provide its own pkg-config implementation in the future, precisely to avoid this kind of problems.

MSoegtropIMC:
@ejgallego : I wonder what kind of fix you would prefer. One option, as I said, would be to create an .exe replacement for the one line shell script. It should be less than 10 lines of C code. Fixing this in dune in a nice way might be tricky since in the end you need some sort of shell to execute a shell script, and from what I learned from you about dune I figure that dune tries to avoid this. Another option would be to force an old version of pkg-config, but I guess such a solution wouldn't last that long.

Any other ideas?

ejgallego:
> Any other ideas?

I have no idea; have you checked indeed the verbose Dune log to see that the problem is the non-executable file?

ejgallego:
Dune upstream is informed of this so we should see what they said; but if that's the problem, indeed having a pkg-config-exe package would be the best solution. I wonder why pkg-config upstream did this really bad change.

ejgallego:
Umm, @MSoegtropIMC , the pkg-config repository hasn't seen a change in 2 years, so how come this change has happened? https://cgit.freedesktop.org/pkg-config/log/

In linux, `pkg-config` is still a binary:
```
$ file $(which pkg-config)
/usr/bin/pkg-config: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=09c70d0d6020d440400ea5c0d8dd63a24379e6cf, stripped
```
what is going on?

MSoegtropIMC:
@ejgallego : I created a new issue #9760 for this, since this starts to get to lengthy here. I put a link to the release message from Cygwin there.

MSoegtropIMC:
@ppedrot : Since recent changes in cygwin broke the Windows build for this PR in a more subtle way than on master (where this is fixed already), I will do an additional fix in this PR as discussed in issue #9760. If you want to merge this over the weekend, please let me know asap, then I won't do the fix here, but do a separate PR on master which has the disadvantage that it can't be tested there.

MSoegtropIMC:
@ppedrot , @ejgallego : I have a fix for the pkg-config issue with dune and dune calls pkg-config fine now. But there seems to be a bug in lablgtk-3: it seems to require 
```
lablgtk-3.0.beta5-dune-build_err.txt: Error: command has non-empty stderr: C:\bin\cygwin_coq64_gtk3\build\lablgtk-3.0.beta5\bin_special\pkg-config.exe --cflags gtk+-quartz-3.0
```
which as far as I can tell doesn't make a lot of sense on Windows since afaik quartz is the Mac backend of GTK, which obviously doesn't exist on Windows.

How shall I proceed? Push the pkg-config fix anyway and have someone else fix lablgtk3 or try to patch lablgtk3 as well? I didn't look into it as yet - I guess it should be a no brainer for a dune expert - not sure how long it would take me to teach dune platform specific dependencies.

ejgallego:
This doesn't seem a bug in Dune but a bug in the pkg-config wrapper script; Dune first queries for gtk+-quartz to see if it is available, which is a perfectly fine thing to do; it seems the call succeeded ?

MSoegtropIMC:
> This doesn't seem a bug in Dune but a bug in the pkg-config wrapper script;

Indeed you are right. The full log is:
```
lablgtk-3.0.beta5-dune-build_err.txt: run: C:\bin\cygwin_coq64_gtk3\build\lablgtk-3.0.beta5\bin_special\pkg-config.exe gtk+-quartz-3.0
lablgtk-3.0.beta5-dune-build_err.txt: -> process exited with code 0
lablgtk-3.0.beta5-dune-build_err.txt: -> stdout:
lablgtk-3.0.beta5-dune-build_err.txt: -> stderr:
lablgtk-3.0.beta5-dune-build_err.txt: run: C:\bin\cygwin_coq64_gtk3\build\lablgtk-3.0.beta5\bin_special\pkg-config.exe --cflags gtk+-quartz-3.0
lablgtk-3.0.beta5-dune-build_err.txt: -> process exited with code 0
lablgtk-3.0.beta5-dune-build_err.txt: -> stdout:
lablgtk-3.0.beta5-dune-build_err.txt: -> stderr:
lablgtk-3.0.beta5-dune-build_err.txt:  | Package gtk+-quartz-3.0 was not found in the pkg-config search path.
lablgtk-3.0.beta5-dune-build_err.txt:  | Perhaps you should add the directory containing `gtk+-quartz-3.0.pc'
lablgtk-3.0.beta5-dune-build_err.txt:  | to the PKG_CONFIG_PATH environment variable
lablgtk-3.0.beta5-dune-build_err.txt:  | Package 'gtk+-quartz-3.0', required by 'virtual:world', not found
lablgtk-3.0.beta5-dune-build_err.txt: Error: command has non-empty stderr: C:\bin\cygwin_coq64_gtk3\build\lablgtk-3.0.beta5\bin_special\pkg-config.exe --cflags gtk+-quartz-3.0
```
I guess the return code from the first call is wrong - should be easy to fix ...

MSoegtropIMC:
Ok, fixed - lablgtk3 builds fine locally now (coq build is still running).

I will **rebase and push** as soon as the Coq build went through locally and I did a quick test of CoqIDE unless someone has objections.

ejgallego:
Please go ahead @MSoegtropIMC , thanks!

MSoegtropIMC:
Btw.: the bugs in the resulting CoqIDE (icons missing, crash on open) are not yet fixed (we discussed that we do this later). My commit just fixes the Windows build issues introduced by the recent cygwin change. If there is still time, I could fix the bugs as well. Not sure what plans you have to merge this.

MSoegtropIMC:
OK, CI passes again :-)

Would you prefer to merge it now or shall I fix the remaining Windows CoqIDE runtime issues first (shouldn't be difficult).

ejgallego:
IMVHO this is ready to go, allez-y @ppedrot !

ppedrot:
This might be a little reckless, but let's merge indeed. We will fix the horrible bugs likely to be discovered on the way to 8.10.0. This is a highly non-trivial PR though, so, @coq/contributors just to be sure that everybody is fine, I'll leave until tomorrow for the merge.

vbgl:
I’ll fix the conflict on `default.nix`.

ejgallego:
> I’ll fix the conflict on `default.nix`.

Thanks @vbgl ; @ppedrot , of course this is hard to predict but I have a feeling that GTK3 will provide a better experience to people when compared to the ancient GTK2,

ejgallego:
CI is OK, failure is the typical async problem.

ppedrot:
Let's merge...

ppedrot:
BTW I am willing to reimplement the progress bar, does anybody knows what needs to be done?

ejgallego:
@ppedrot AFAICT you need to do the same as before but using Cairo [https://github.com/Chris00/ocaml-cairo]

TTBOMK the Cairo API should allow you to do quite interesting, and even fancy, stuff.

Note that indeed you may find that the lablgtk3 beta may want some better cairo interaction, but now you can do a quite interesting build setup, as all 3 cairo-ocaml, lablgtk3 and coq use Dune, you can just symlink the two libs into a vendor dir and hack the 3 projects at the same time [even getting merlin integration, but well I know you don't use merlin]

ejgallego:
By the way congrats to all involved here this was very hard work.

ppedrot:
@ejgallego Am I right that there is no Cairo bindings for lablgtk3 yet? So that I have to write them down first? Links to the documentation are broken and peeking at the API using OCaml toplevel is no fun...

ejgallego:
I think there are bindings indeed added by @sacerdot , there is also documentation, anyways it is pretty easy to generate using `dune build @doc`. Also, peeking at the API using `dune utop` is pretty pleasant.

sacerdot:
Hi @ppedrot ,
yes, I connected lablgtk3 with cairo and pango. You can see examples in lablgtk3/examples/{drawing,pango1,pango2,nd}.ml

Let me know if you have any question.

ppedrot:
Indeed thanks, I managed to hack a quick adaptation to Cairo, I'll submit a PR soon.

PR #9284, opened 2018-12-25
[dune] Fix states target to depend on installed libraries.

ejgallego:
Until proper Coq support comes to Dune, this is a workaround and an
alternative to #9047 so `dune build @states` will build a working
`coqtop`.

In the end the shim path seems better to me, and this is the way `dune
coqtop` will work [and it will allow to obtain a context with
arbitrary libraries], however this is an interesting option to
consider.

### DIFF
--- a/dune
+++ b/dune
@@ -42,3 +42,20 @@
  (name runtest)
  (package coqide-server)
  (deps test-suite/summary.log))
+
+; This is an ugly hack, we should use
+; %{lib_root:coq/theories/Init/Prelude.vo} when Dune supports it.
+(library
+ (name root)
+ (public_name coq.root))
+
+(alias
+ (name states)
+ (deps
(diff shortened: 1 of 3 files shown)


dune
```diff
@@ -42,3 +42,20 @@
  (name runtest)
  (package coqide-server)
  (deps test-suite/summary.log))
+
+; This is an ugly hack, we should use
+; %{lib_root:coq/theories/Init/Prelude.vo} when Dune supports it.
```
SkySkimmer: When will dune support it?

ejgallego: Likely next version, but indeed this PR may be closed in favor of the native solution before.

ejgallego:
It turns out that this is quite tricky to do as the `ltac` package is auto-generated thus the first dune invocation will fail. We will have to wait for a newer Dune with `coqlib` support.

PR #9269, opened 2018-12-21
Move and rewrite intro pattern section

jfehrle:
**Kind:** documentation.

Fixes / closes https://github.com/coq/coq/issues/9221

This is a work in progress.  It needs a fair bit of help from those who know more.

This PR:
- moves the description of intro patterns from under the intros tactic to its own subsection under "Tactics/Invocation of tactics"
- rewrites the material almost completely to make it simpler and accessible to the non-expert.  Description progresses from the most commonly to less commonly used parts.
- adds a grammar for intro patterns
- adds more (and simpler) examples, grouped at the end of the section

PR #164, opened 2016-04-29
A few tactics for 8.6

herbelin:
This PR offers various evolution of tactics. I did not make separate PRs because it was a bit difficult to rebase them into individual independent commits.

The first one is about pat%constr (or whatever form it takes). It adds eintros so that the underlying call to "apply in" calls "eapply in".

This is for consistency with the "e" model. If ever it happens to be a good move to have the "e" behavior a default, we would not need to have a distinction.

The second one is about implementing ssreflect-style "/constr" as discussed on Dec 20, if ever there is some demand for it as a complement of pat%constr (whatever the syntax of pat%constr is).

The third one is about adding an as clause to specialize, which might be convenient to have on the model of the as clause of "apply in" or "assert".

The commits about induction_arg are internal. They are here just because I found it too difficult to remove them from the sequence of commits.

The commit on injection is to provide a form of injection which is "pure" in the sense that it replaces an equation by a set of equations "in place", without interfering with the conclusion of the goal, following the model of destruct/induction/apply in/rewrite in/unfold in/etc. In particular, it invents names (but names can be chosen by the user using an "as" clause).

Incidentally, this variant of injection generate equations in the "natural" left-to-right order, rather the right-to-left order.

With this new behavior, we obtain a consistency of the "algebraic" pattern model and of the corresponding tactics, with an implicit "on H do pattern" behind the scene:

intros pat = intro H; on H do pat
destruct H as disj_pat = on H do disj_pat
injection H as eq_pat = on H do eq_pat (*)
apply c in H as pat = on H do pat%c

subst x = on H do ->  (when H has the form x=t or t=x)
rewrite H = on H do -> (otherwise)
destruct H = on H do []
injection H = on H do [=]
discriminate H = on H do [=]
apply c in H = on H do H%c
rename H into H' = on H do H'

(*) Incidentally, to complete the picture, this suggests to write "injection H as [= pat ... pat]" rather than "injection as pat ... pat".

The commit on ==> was already presented in #136, where its motivation was given. It is here because it is difficult to isolate it out.

[...]

PR #9288, opened 2018-12-27
Giving preference to syntax "injection ... as [= pat1 ... patn]".

herbelin:
**Kind:**  feature

This PR is possibly in relation with #9269 (rewriting the section on introduction patterns in the reference manual), considering that the syntax `injection ... as [= ...]` may be simpler to explain than the syntax `injection ... as pat1 ... patn` in the sense that it would follow the same pattern that we have in `destruct ... as [ ... ]`/`destruct ... as (...)` of using the same syntax as in `intros` . In particular, it would scale easier to a general action tactic `on H do action_pattern` as was discussed a few years ago.

We also deprecate the syntax `injection ... as pat1 ... patn`. Maybe this is overkill and both syntaxes could be accepted (i.e. maybe it is no harm to keep both `injection ... as [= pat1 ... patn ]` and `injection ... as pat1 ... patn`).

I adapted the files of the standard library using the following `sed`-like incantation:
```
sed -e "s/injection \([^ ]*\) as \([^;\.)]*\)\([;\.)]\)/injection \1 as [= \2]\3)/g"
```
This is fragile though. It works only if no parentheses are used in the intro patterns and no space in the main argument, and no space after the end of the tactic.

#### Typical possible future work around injection

Incidentally, there are cool things to implement regarding`injection` in the dependent case. For instance, currently, we have the following frustrating result:
```coq
Notation "( x ; .. ; y ; z )" := (existT _ x .. (existT _ y z) ..) (at level 0, format "( x ;  '/  ' .. ;  '/  ' y ;  '/  ' z )").
Import EqNotations.
Set Keep Proof Equalities.
Goal forall a b, (a ; b) = (0 ; eq_refl) -> {e : a = 0 & rew e in b = eq_refl }.
intros *. injection 1 as [=].
(* 
1 subgoal
a : nat
b : 0 = a
H : a = 0
H0 : (a; b) = (0; eq_refl)    (* The dependency is expressed back using a sigT type *)
=======================
{e : a = 0 & rew e in b = eq_refl }
*)
```
while we could typically get something more elaborated like:
```coq
1 subgoal
a : nat
b : 0 = a
H : a = 0
H0 : b =_[^H in 0 = H] eq_refl    (* Now rewritable *)
{e : a = 0 & rew e in b = eq_refl }
```
where `x =_[ e ] y` is a dependent equality, typically one defined by (but there are variants):
```coq
Declare Scope eq_scope.
Delimit Scope eq_scope with eq.
Reserved Notation "a =_[ e ] b" (at level 70, format "a  =_[ e ]  b").
Inductive eq_over (A:Prop) (a:A) : forall (B:Prop) (b:B), A = B -> Type :=
  refl_over : a =_[eq_refl] a
where "a =_[ e ] b" := (eq_over _ a _ b e%eq).
Notation "^ H 'in' P" := (f_equal (fun H => P) H) (at level 200) : eq_scope.
```

### DIFF
--- a/doc/changelog/04-tactics/09288-injection-as.rst
+++ b/doc/changelog/04-tactics/09288-injection-as.rst
@@ -1,0 +1,4 @@
+- Documented syntax :n:`injection @term as [= {+ @intropattern} ]` as
+  an alternative to :n:`injection @term as {+ @simple_intropattern}` using
+  the standard :n:`@injection_intropattern` syntax (`#09288
+  <https://github.com/coq/coq/pull/09288>`_, by Hugo Herbelin).
--- a/doc/sphinx/proof-engine/tactics.rst
+++ b/doc/sphinx/proof-engine/tactics.rst
@@ -131,16 +131,17 @@ include :tacn:`assert`, :tacn:`intros` and :tacn:`destruct`.
    simple_intropattern_closed : `naming_intropattern`
                               : _
                               : `or_and_intropattern`
-                              : `equality_intropattern`
+                              : `rewriting_intropattern`
+                              : `injection_intropattern`
    naming_intropattern     : `ident`
                            : ?
                            : ?`ident`
    or_and_intropattern     : [ `intropattern_list` | ... | `intropattern_list` ]
                            : ( `simple_intropattern` , ... , `simple_intropattern` )
                            : ( `simple_intropattern` & ... & `simple_intropattern` )
-   equality_intropattern   : ->
+   rewriting_intropattern  : ->
                            : <-
-                           : [= `intropattern_list` ]
+   injection_intropattern  : [= `intropattern_list` ]
    or_and_intropattern_loc : `or_and_intropattern`
                            : `ident`
 
@@ -2284,6 +2285,18 @@ and an explanation of the underlying technique.
       names are automatically generated to adjust the list of :n:`@simple_intropattern`
       to the number of new equalities. The original equality is erased if it
       corresponds to a hypothesis.
+
+   .. tacv:: injection @term {? with @bindings_list} as @injection_intropattern
+             injection @num as @injection_intropattern
+             injection as @injection_intropattern
+             einjection @term {? with @bindings_list} as @injection_intropattern
+             einjection @num as @injection_intropattern
+             einjection as @injection_intropattern
+
+      These are equivalent to the previous variants but using instead the
+      syntax :token:`injection_intropattern` which :tacn:`intros`
+      uses. In particular :n:`as [= {+ @simple_intropattern}]` behaves
+      the same as :n:`as {+ @simple_intropattern}`.
 
    .. flag:: Structural Injection
 
--- a/plugins/ltac/extratactics.mlg
+++ b/plugins/ltac/extratactics.mlg
@@ -146,6 +146,23 @@ let discrHyp id =
 let injection_main with_evars c =
  elimOnConstrWithHoles (injClause None None) with_evars c
 
+let isInjPat pat = match pat.CAst.v with IntroAction (IntroInjection _) -> Some pat.CAst.loc | _ -> None
+
+let decode_inj_ipat ?loc = function
+  (* For the "as [= pat1 ... patn ]" syntax *)
+  | [{ CAst.v = IntroAction (IntroInjection ipat) }] -> ipat
+  (* For the "as pat1 ... patn" syntax *)
+  | ([] | [_]) as ipat -> ipat
+  | pat1::pat2::_ as ipat ->
+  (* To be sure that there is no confusion of syntax, we check that no [= ...] occurs
+     in the non-singleton list of patterns *)
+  match isInjPat pat1 with
+  | Some _ -> user_err ?loc:pat2.CAst.loc (str "Unexpected pattern.")
+  | None ->
+  match List.map_filter isInjPat ipat with
+  | loc :: _ -> user_err ?loc (str "Unexpected injection pattern.")
+  | [] -> ipat
+
 }
 
 TACTIC EXTEND injection
@@ -158,15 +175,15 @@ TACTIC EXTEND einjection
 END
 TACTIC EXTEND injection_as
 | [ "injection" "as" intropattern_list(ipat)] ->
-    { injClause None (Some ipat) false None }
+    { injClause None (Some (decode_inj_ipat ipat)) false None }
 | [ "injection" destruction_arg(c) "as" intropattern_list(ipat)] ->
-    { mytclWithHoles (injClause None (Some ipat)) false c }
+    { mytclWithHoles (injClause None (Some (decode_inj_ipat ipat))) false c }
 END
 TACTIC EXTEND einjection_as
 | [ "einjection" "as" intropattern_list(ipat)] ->
-    { injClause None (Some ipat) true None }
+    { injClause None (Some (decode_inj_ipat ipat)) true None }
 | [ "einjection" destruction_arg(c) "as" intropattern_list(ipat)] ->
-    { mytclWithHoles (injClause None (Some ipat)) true c }
+    { mytclWithHoles (injClause None (Some (decode_inj_ipat ipat))) true c }
 END
 TACTIC EXTEND simple_injection
 | [ "simple" "injection" ] -> { simpleInjClause None false None }
--- a/theories/Arith/Peano_dec.v
+++ b/theories/Arith/Peano_dec.v
@@ -53,7 +53,7 @@ destruct le_mn1; intros le_mn2; destruct le_mn2.
   now destruct (Nat.nle_succ_diag_l _ le_mn0).
 + intros def_n0; generalize le_mn1; rewrite def_n0; intros le_mn0.
   now destruct (Nat.nle_succ_diag_l _ le_mn0).
-+ intros def_n0. injection def_n0 as ->.
++ intros def_n0. injection def_n0 as [= ->].
   rewrite (UIP_nat _ _ def_n0 eq_refl); simpl.
   assert (H : le_mn1 = le_mn2).
     now apply IHn0.
--- a/theories/FSets/FSetPositive.v
+++ b/theories/FSets/FSetPositive.v
@@ -1009,10 +1009,10 @@ Module PositiveSet <: S with Module E:=PositiveOrderedTypeBits.
       destruct o.
         intros x H. injection H; intros; subst. reflexivity.
         revert IHl. case choose.
-          intros p Hp x H. injection H as <-. apply Hp.
+          intros p Hp x [= <-]. apply Hp.
            reflexivity.
           intros _ x. revert IHr. case choose.
-            intros p Hp H. injection H as <-. apply Hp.
+            intros p Hp [= <-]. apply Hp.
             reflexivity.
             intros. discriminate.
   Qed.
@@ -1068,11 +1068,11 @@ Module PositiveSet <: S with Module E:=PositiveOrderedTypeBits.
    induction s as [| l IHl o r IHr]; simpl.
      intros. discriminate.
      intros x. destruct (min_elt l); intros.
-       injection H as <-. apply IHl. reflexivity.
+       injection H as [= <-]. apply IHl. reflexivity.
        destruct o; simpl.
-         injection H as <-. reflexivity.
+         injection H as [= <-]. reflexivity.
          destruct (min_elt r); simpl in *.
-           injection H as <-. apply IHr. reflexivity.
+           injection H as [= <-]. apply IHr. reflexivity.
            discriminate.
   Qed.
 
@@ -1096,15 +1096,15 @@ Module PositiveSet <: S with Module E:=PositiveOrderedTypeBits.
     induction s as [|l IHl o r IHr]; intros x y H H'.
       discriminate.
       simpl in H. case_eq (min_elt l).
-        intros p Hp. rewrite Hp in H. injection H as <-.
+        intros p Hp. rewrite Hp in H. injection H as [= <-].
         destruct y as [z|z|]; simpl; intro; trivial. apply (IHl p z); trivial.
         intro Hp; rewrite Hp in H. apply min_elt_3 in Hp.
         destruct o.
-          injection H as <-. intros Hl.
+          injection H as [= <-]. intros Hl.
           destruct y as [z|z|]; simpl; trivial. elim (Hp _ H').
 
           destruct (min_elt r).
-            injection H as <-.
+            injection H as [= <-].
             destruct y as [z|z|].
               apply (IHr e z); trivial.
               elim (Hp _ H').
@@ -1121,11 +1121,11 @@ Module PositiveSet <: S with Module E:=PositiveOrderedTypeBits.
    induction s as [| l IHl o r IHr]; simpl.
      intros. discriminate.
      intros x. destruct (max_elt r); intros.
-       injection H as <-. apply IHr. reflexivity.
+       injection H as [= <-]. apply IHr. reflexivity.
        destruct o; simpl.
-         injection H as <-. reflexivity.
+         injection H as [= <-]. reflexivity.
          destruct (max_elt l); simpl in *.
-           injection H as <-. apply IHl. reflexivity.
+           injection H as [= <-]. apply IHl. reflexivity.
            discriminate.
   Qed.
 
@@ -1149,15 +1149,15 @@ Module PositiveSet <: S with Module E:=PositiveOrderedTypeBits.
     induction s as [|l IHl o r IHr]; intros x y H H'.
       discriminate.
       simpl in H. case_eq (max_elt r).
-        intros p Hp. rewrite Hp in H. injection H as <-.
+        intros p Hp. rewrite Hp in H. injection H as [= <-].
         destruct y as [z|z|]; simpl; intro; trivial. apply (IHr p z); trivial.
         intro Hp; rewrite Hp in H. apply max_elt_3 in Hp.
         destruct o.
-          injection H as <-. intros Hl.
+          injection H as [= <-]. intros Hl.
           destruct y as [z|z|]; simpl; trivial. elim (Hp _ H').
 
           destruct (max_elt l).
-            injection H as <-.
+            injection H as [= <-].
             destruct y as [z|z|].
               elim (Hp _ H').
               apply (IHl e z); trivial.
--- a/theories/Lists/List.v
+++ b/theories/Lists/List.v
@@ -250,7 +250,7 @@ Section Facts.
     generalize (app_nil_r l); intros E.
     rewrite -> E; auto.
     intros.
-    injection H as H H0.
+    injection H as [= H H0].
     assert ([] = l ++ a0 :: l0) by auto.
     apply app_cons_not_nil in H1 as [].
   Qed.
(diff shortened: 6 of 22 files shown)


herbelin:
> In this view the PR adds redundancy, and not just uniformity. So I'm not so convinced that the deprecation warning is desirable.

Somehow, this is the indeed the question.

In `destruct`, seen as a tactic for splitting the structure of an inductive type, there is a redundancy between the syntax of the destruction pattern and the syntax of this destruction when used in `intros`. (This redundancy, combined with the maybe too general meaning of the word `destruct` makes that some people are actually surprised that `destruct ... as [=]` does not work...)

In vanilla `rewrite`, there is also a syntactic redundancy between the notations `->`/`<-` (as in `rewrite -> H`) and the introduction pattern notations `->`/`<-`.

There is no such redundancy in `apply foo1, foo2 in H`, and in particular no such thing as `apply %foo1 %foo2 in H` (or `apply /foo1 /foo2 in H`). But there is close to such a redundancy in `specialize (H args)` and ssreflect `/(_ args)` pattern.

There is also the general scheme that `as` expects one pattern, and `injection` is here an exception.

So, my question is somehow whether syntactic redundancy and uniformity is helpful to explain the introduction-patterns global picture to newcomers.

> It is your choice as an author of the text if the injection is a relevant proof step, and hence you name it, or it is a bookkeeping operation, and in that case you use the symbol.

Yes, this is basically also my view.

Zimmi48:
One thing is clear: the current situation is inconsistent with `destruct` requiring a destruct intro pattern while `injection` requires a list of intro patterns. We should fix this inconsistency, by changing either `destruct` or `injection`. While the redundant solution may be perceived as less elegant, it also as an advantage in terms of discoverability of the corresponding intro pattern. What this means then is that using `destruct` / `injection` is just about making things more explicit, at the price of redundancy.

Given the way the context is managed (in vanilla Coq), it is also the way of working with hypotheses which were already introduced. In this case, we cannot avoid the redundancy unless we introduce a new `on H do intropattern` as mentioned by Hugo. This seems to me like the next logical step, although I would prefer a syntax that keeps the "as" word, for instance `treat H as pattern` or `consider H as intropattern` (pushing the idea that tactics taking a single intro pattern do so through an "as" clause).

So overall, I'm favorable to this PR and the "next logical step" mentioned above.

gares:
I don't think it makes sense to trade redundancy for discoverability at the language level. Repeating things in the manual with cross links is one thing. Forcing every explicit call to injection to carry the "hyperlink" seems a bit excessive imho. Accepting it in the syntax so that one can use the syntax in a tutorial or something could make sense. Giving a warning if you don't always repeat yourself is another thing, the one I'm not happy about.

  In particular the warning tells our users to update their files. Imo you should have a good story to convince them that the new way of doing things is better, but I don't see strong arguments in this discussion.

maximedenes (commented):
No particular opinion (sending a review to remove the review requested tag).

Zimmi48:
I'm changing my mind on the deprecation, as I'll explain in the following, but even if I hadn't, I think it is stupid to hold this PR from being merged for this reason, so please @herbelin, remove the deprecation.

Today I taught the use of the `as` clause for `destruct`, `induction` and `injection`. As always, this is not something that I introduce from the start because it is non-trivial so I first introduce the tactics generating names, and only later shows how to choose the names for better readability and robustness. The use of the pair for the conjunction and the disjunctive intro-pattern is somehow ok although having a simpler variant would allow to introduce them earlier. In the case of `induction` in particular, the use of the disjunctive intro-pattern, that doesn't mean the same thing when used with `intros`, and that is quite complex when what you want is to give a name to the induction hypothesis is quite inconvenient to say the least. But with `injection` which doesn't behave well by default (put new hypothesis in the goal), this is important to introduce the `as` clause from the beginning.

Discoverability is also important but we must think of a mechanism that is not using a warning, at least not one that is activated by default...

herbelin:
Deprecation deactivated. Rebase done.

Maybe worth to take a couple of days to get extra new feedback. In the meantime, @gares, are you willing to assign?

gares:
Next week I'm busy, so I think you want someone more reactive. If the warning is gone, I'm ok with the pr being merged

Zimmi48:
I'll do it.

plugins/ltac/extratactics.mlg
```diff
@@ -146,6 +146,23 @@ let discrHyp id =
...
 
+let warn_deprecated_injection_intropattern_syntax = fun ?loc () -> ()
+(* Keeping code in case the general view evolves (e.g. when "on H do pat"
+   is introduced) *)
+(*
+  CWarnings.create ~name:"injection-arguments-flat-intropattern-syntax" ~category:"deprecated"
+    (fun () ->
+       strbrk "Syntax " ++ str "\"injection ... as pat1 ... patn\"" ++
+       strbrk " is deprecated. Use " ++ str "\" injection ... as [= pat1 ... patn ]\".")
+*)
```
Zimmi48: Please don't do this: the code is not complex enough to be useful to keep, and commented code is bound to rot. If you want to keep a comment noting that it was discussed whether the simpler intro-pattern should be deprecated, just keep it as a textual comment.

plugins/ltac/extratactics.mlg
```diff
@@ -146,6 +146,23 @@ let discrHyp id =
...
+
+let check_inj_ipat ?loc = function
+  | [{ CAst.v = IntroAction (IntroInjection ipat) }] -> ipat
+  | [] -> warn_deprecated_injection_intropattern_syntax (); []
```
Zimmi48: Moreover, this is confusing, because if you don't look further, you may think that this is actually deprecated. And it complexifies the code.

theories/Arith/Peano_dec.v
```diff
@@ -53,7 +53,7 @@ destruct le_mn1; intros le_mn2; destruct le_mn2.
   now destruct (Nat.nle_succ_diag_l _ le_mn0).
 + intros def_n0; generalize le_mn1; rewrite def_n0; intros le_mn0.
   now destruct (Nat.nle_succ_diag_l _ le_mn0).
-+ intros def_n0. injection def_n0 as ->.
++ intros def_n0. injection def_n0 as [= ->].
```
Zimmi48: What about simplifying this line to:

```suggestion
+ intros [= ->].
```

theories/FSets/FSetPositive.v
```diff
@@ -1009,10 +1009,10 @@ Module PositiveSet <: S with Module E:=PositiveOrderedTypeBits.
       destruct o.
         intros x H. injection H; intros; subst. reflexivity.
         revert IHl. case choose.
-          intros p Hp x H. injection H as <-. apply Hp.
+          intros p Hp x H. injection H as [= <-]. apply Hp.
```
Zimmi48: What about simplifying this line to:

```suggestion
          intros p Hp x [= <-]. apply Hp.
```

theories/FSets/FSetPositive.v
```diff
@@ -1009,10 +1009,10 @@ Module PositiveSet <: S with Module E:=PositiveOrderedTypeBits.
...
            reflexivity.
           intros _ x. revert IHr. case choose.
-            intros p Hp H. injection H as <-. apply Hp.
+            intros p Hp H. injection H as [= <-]. apply Hp.
```
Zimmi48:
```suggestion
            intros p Hp [= <-]. apply Hp.
```

theories/Lists/List.v
```diff
@@ -265,14 +265,14 @@ Section Facts.
...
+      injection H as [= H1 H0].
       apply app_cons_not_nil in H0 as [].
     - intros a b H.
-      injection H as H1 H0.
+      injection H as [= H1 H0].
       assert ([] = l ++ [a]) by auto.
       apply app_cons_not_nil in H as [].
     - intros a0 b H.
-      injection H as <- H0.
+      injection H as [= <- H0].
```
Zimmi48: Possible simplifications in this proof too.

theories/MSets/MSetPositive.v
```diff
@@ -910,10 +910,10 @@ Module PositiveSet <: S with Module E:=PositiveOrderedTypeBits.
       destruct o.
         intros x H. injection H; intros; subst. reflexivity.
         revert IHl. case choose.
-          intros p Hp x H. injection H as <-. apply Hp.
+          intros p Hp x H. injection H as [= <-]. apply Hp.
```
Zimmi48:
```suggestion
          intros p Hp x [= <-]. apply Hp.
```

theories/MSets/MSetPositive.v
```diff
@@ -910,10 +910,10 @@ Module PositiveSet <: S with Module E:=PositiveOrderedTypeBits.
...
            reflexivity.
           intros _ x. revert IHr. case choose.
-            intros p Hp H. injection H as <-. apply Hp.
+            intros p Hp H. injection H as [= <-]. apply Hp.
```
Zimmi48:
```suggestion
            intros p Hp [= <-]. apply Hp.
```

theories/MSets/MSetRBT.v
```diff
@@ -1950,7 +1950,7 @@ Module Make (X: Orders.OrderedType) <:
  generalize (fun x s' => @Raw.remove_min_spec1 s x s' Hs).
  set (P := Raw.remove_min_ok s). clearbody P.
  destruct (Raw.remove_min s) as [(x0,s0)|]; try easy.
- intros H U. injection U as -> <-. simpl.
+ intros H U. injection U as [= -> <-]. simpl.
```
Zimmi48:
```suggestion
 intros H [= -> <-]. simpl.
```

theories/Strings/String.v
```diff
@@ -123,7 +123,7 @@ intros H; generalize (H O); simpl; intros H1; inversion H1.
 case (Rec s).
 intros H0; rewrite H0; auto.
 intros n; exact (H (S n)).
-intros H; injection H as H1 H2.
+intros H; injection H as [= H1 H2].
```
Zimmi48:
```suggestion
intros [= H1 H2].
```

theories/Strings/String.v
```diff
@@ -290,14 +290,14 @@ intros n m s1 s2; generalize n m s1; clear n m s1; elim s2; simpl;
  auto.
 intros n; case n; simpl; auto.
 intros m s1; case s1; simpl; auto.
-intros H; injection H as <-; auto.
+intros H; injection H as [= <-]; auto.
```
Zimmi48:
```suggestion
intros [= <-]; auto.
```

theories/Strings/String.v
```diff
@@ -290,14 +290,14 @@ intros n m s1 s2; generalize n m s1; clear n m s1; elim s2; simpl;
...
 intros; discriminate.
 intros; discriminate.
 intros b s2' Rec n m s1.
 case n; simpl; auto.
 generalize (prefix_correct s1 (String b s2'));
  case (prefix s1 (String b s2')).
-intros H0 H; injection H as <-; auto.
+intros H0 H; injection H as [= <-]; auto.
```
Zimmi48:
```suggestion
intros H0 [= <-]; auto.
```

theories/Strings/String.v
```diff
@@ -323,15 +323,15 @@ intros n m s1 s2; generalize n m s1; clear n m s1; elim s2; simpl;
  auto.
 intros n; case n; simpl; auto.
 intros m s1; case s1; simpl; auto.
-intros H; injection H as <-.
+intros H; injection H as [= <-].
```
Zimmi48:
```suggestion
intros [= <-].
```

theories/Strings/String.v
```diff
@@ -323,15 +323,15 @@ intros n m s1 s2; generalize n m s1; clear n m s1; elim s2; simpl;
...
 intros p H0 H2; inversion H2.
 intros; discriminate.
 intros; discriminate.
 intros b s2' Rec n m s1.
 case n; simpl; auto.
 generalize (prefix_correct s1 (String b s2'));
  case (prefix s1 (String b s2')).
-intros H0 H; injection H as <-; auto.
+intros H0 H; injection H as [= <-]; auto.
```
Zimmi48:
```suggestion
intros H0 [= <-]; auto.
```

Zimmi48:
There is a linter problem. Should be fixed hopefully with a `git rebase --whitespace=fix`.

herbelin:
The point of this PR was to consider `injection t as [= pat1 ... patn]` as more in line with the syntax of `destruct` than `injection as pat1 ... patn` was. Accordingly, it adds symbol which can be seen as redundant information.

So, let's not have a deprecation. It remains two questions:
- At which level of importance do we document `injection t as [= pat1 ... patn]` and `injection as pat1 ... patn`? The conclusion (based on the expression of only 3 persons though) seems to say that we should document them at the same level, one justified by its uniformity, the other by its simplicity.
- At which level do we show how to use it in the files we release with Coq? @zimmi48 seems to say that it is even better to recommend `intros [= ...]` over `intros H; injection H as [= ...]`. In my (very) subjective opinion, I find the former (too) cryptic to read and the second (too) verbose to type. The dilemma is unresolvable, or, rather, it is solvable but with more tools... Ideally, there should be a way in `coqdoc`-rendered files, to show different levels of explicitation in scripts. The "pedestrian" button would typically give `intros H; injection H as [= ...]` and the "vanilla compact" button would typically give `intros [= ...]`. But in the meantime, I don't see how to address the issue in a way which satisfies all (reasonable) needs.

So, I'd like to propose that:
- This PR does not change anything in the stdlib (but, @zimmi48, other PRs can)
- To restrict this PR to documenting the `as [= ...]` syntax, either presented it on par with the `as ...` syntax, or as an alternative, explaining the reasons for two syntaxes. @gares, would that be ok? Extra opinions might also help... maybe what I feel as uniformity is not so important after all.

herbelin:
>  But with injection which doesn't behave well by default (put new hypothesis in the goal), this is important to introduce the as clause from the beginning.

Note that an alternative here is also to change the semantics of `injection` without `as` (+ some way to get the compatibility). [This was proposed as part of #164 actually]

gares:
I think your plan is OK, as I said i was just unhappy about the warning. Adding a new syntax for the sake of uniformity is OK. I don't have strong opinions about which one should be documented first. Wrt changing stdlib, my opinion is that it is like `fun x -> match x with ...` and `function ...` in OCaml. In some cases `x` can be a name that helps reading, but in most cases if `x` is only named to be immediately destructed then I strongly prefer the second form, because it tells the reader the input is used linearly (and if the code that follows is large, it is nice to know that IMO). The same holds for the two options you are discussing here, but I agree that change can go in another PR, and probably should.

Zimmi48:
I am fine with the plan you propose @herbelin and I agree with @gares about explicit linearity being the main benefit of merging `intros` with the next tactic. And I think the simpler syntax should be listed first (since it's simpler) but it should be immediately followed by the variant (because it helps discoverability).

> Note that an alternative here is also to change the semantics of `injection` without `as` (+ some way to get the compatibility).

This can also be done (separately) and I would be in favor of it. The plan for compatibility is quite simple: tell people to just use "as".

herbelin:
Here is an update on the plan:
- I documented `injection as [= ]` in second.
- I maintained a change of semantics which I forgot I had made in the initial PR: Previously, something like `injection term as [= pat1 pat2] pat3` was accepted and interpreted like `injection term as pat1 pat2 pat3`. I forbid it, as well as I forbid to use a `[= ...]` in a list of size > 1. This seems to me to prevent confusion (either we use exactly one `[= ...]`, or we don't need to use `[= ...]` at all since all equations are only in fully decomposed form).
- I added a test.
- I validated @zimmi48's proposed changes.

theories/Arith/Peano_dec.v
herbelin: I removed this one which does not work (`def_n0` used on the next line).

Zimmi48 (approved):
One minor issue with Sphinx markup.
Could you squash the changes to the stdlib in a single commit and add an entry in `doc/changelog/`?

Zimmi48:
ping @herbelin

herbelin:
> One minor issue with Sphinx markup.
Could you squash the changes to the stdlib in a single commit and add an entry in doc/changelog/?

Unless a mistake, the three requests have been satisfied. For the `changelog` I tried to follow existing model. Please confirm that's the way to do it.

doc/changelog/04-tactics/09288-injection-as.rst
```diff
@@ -0,0 +1,5 @@
+- Documented syntax :n:`injection term as [= pat1 ... patn ]` as an
+  alternative to :n:`injection term as pat1 ... patn` using the
+  standard syntax `[= pat1 ... patn ]` of the injection introduction
+  pattern (`#09288 <https://github.com/coq/coq/pull/09288>`_, by Hugo
```
Zimmi48: What about just:

```rst
using the standard :token:`injection_intropattern` syntax
```

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -2285,6 +2286,18 @@ and an explanation of the underlying technique.
...
+             injection @num as @injection_intropattern
+             injection as @injection_intropattern
+             einjection @term {? with @bindings_list} as @injection_intropattern
+             einjection @num as @injection_intropattern
+             einjection as @injection_intropattern
+
+      These are equivalent to the previous variants but using instead the
+      syntax :token:`injection_intropattern` which :tacn:`intros`
+      uses. In particular :n:`as [= {+ @simple_intropattern}]` behaves
+      the same as :n:`as {+ @simple_intropattern}`.
```
Zimmi48: Is the `[= ]` pattern supposed to contain a list of `simple_intropattern` or a list of `intropattern`? There seems to be a contradiction between the line above and the grammar definition.

Zimmi48 (commented):
Sorry for the nitpicking.

doc/sphinx/proof-engine/tactics.rst
herbelin: `[= ]` contains a list of `intropattern` but, in the other syntax, `injection term as simple_intropattern_list` these are restricted to `simple_intropattern` (not `intropattern`). So the story is a bit complicated.

Maybe can I add in the documentation that, in addition, `[= ]`  accepts `*` and `**` which span until the end of the block of injection equations under consideration?

Otherwise, an alternative would be to accept `*` and `**` in the `injection term as simple_intropattern_list` syntax. I'm a bit reluctant because I'm a bit afraid that it is not as obvious that the scope of `*` and `**` shall be the block of injection equations under consideration.

What would you suggest?

herbelin: What is the preference to give between `` `intropattern` ... `intropattern` `` and `` `{+ @intropattern}` ``? The doc seems to use both.

herbelin:
Thanks for your comment. I made some changes accordingly, but, somehow, changing the syntax of `injection term as simple_intropattern_list` into `injection term as intropattern_list` is probably the simplest way to resolve the mismatch you pointed out. (Even though we are speaking here about a feature which is not needed as long as `injection` does not support equations dependent on other equations.)

PS: To complicate things, internally, the `intropattern` entry is bound for `TACTIC EXTEND` and `Tactic Notation` to `simple_intropattern`... This is documented at the end of the syntax extension chapter but this is however misleading and maybe should we add a phase of deprecation, recommending to use `simple_intropattern`, then, when the deprecation phase is ended, to add `intropattern` with the semantics it has in the tactics grammar??? What do you think?

doc/sphinx/proof-engine/tactics.rst
Zimmi48: In `productionlist` we don't have a choice and are forced to use the first one, but generally we prefer the second when we can.

doc/changelog/04-tactics/09288-injection-as.rst
```diff
@@ -0,0 +1,4 @@
+- Documented syntax :n:`injection @term as [= {+ @intropattern} ]` as
+  an alternative to :n:`injection @term as {+ @simple_intropattern}` using
+  the standard :token:`@injection_intropattern` syntax (`#09288
```
Zimmi48: It can be a bit confusing, but `:token:` only takes the name of the token, like this:
```suggestion
  the standard :token:`injection_intropattern` syntax (`#09288
```

but if you want to always use the `@` syntax, this is the alternative (both render the same):

```suggestion
  the standard :n:`@injection_intropattern` syntax (`#09288
```

Zimmi48 (approved):
LGTM barring the needed fix

doc/changelog/04-tactics/09288-injection-as.rst
herbelin: Done.

Issue #2386, opened 2010-09-10
abstract cannot handle evars


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#2386
From: @braibant
Reported version: 8.5
CC:   @JasonGross, @jonleivent

SkySkimmer:
It works in 8.20

PR #9290, opened 2018-12-28
[stm] Remove `opacity_guarantee` from the classification.

ejgallego:
Semantics of this flag is really weird, as in particular, it is
overriding the core opacity setting in `Proof_global`.

However thanks to some recent changes in the core proof API we can use a more precise heuristic for delegation now, let's see.

### DIFF
--- a/stm/vernac_classifier.ml
+++ b/stm/vernac_classifier.ml
@@ -88,41 +88,27 @@ let classify_vernac e =
                       proof_block_detection = Some "curly" }
     (* StartProof *)
     | VernacDefinition ((DoDischarge,_),({v=i},_),ProveBody _) ->
-      VtStartProof(Doesn'tGuaranteeOpacity, idents_of_name i)
+      VtStartProof(idents_of_name i)
 
     | VernacDefinition (_,({v=i},_),ProveBody _) ->
-      let polymorphic = Attributes.(parse_drop_extra polymorphic atts) in
-      let guarantee = if polymorphic then Doesn'tGuaranteeOpacity else GuaranteesOpacity in
-      VtStartProof(guarantee, idents_of_name i)
+      VtStartProof(idents_of_name i)
     | VernacStartTheoremProof (_,l) ->
-      let polymorphic = Attributes.(parse_drop_extra polymorphic atts) in
       let ids = List.map (fun (({v=i}, _), _) -> i) l in
-      let guarantee = if polymorphic then Doesn'tGuaranteeOpacity else GuaranteesOpacity in
-      VtStartProof (guarantee,ids)
+      VtStartProof ids
     | VernacFixpoint (discharge,l) ->
-      let polymorphic = Attributes.(parse_drop_extra polymorphic atts) in
-       let guarantee =
-         if discharge = DoDischarge || polymorphic then Doesn'tGuaranteeOpacity
-         else GuaranteesOpacity
-       in
-        let ids, open_proof =
-          List.fold_left (fun (l,b) {Vernacexpr.fname={CAst.v=id}; body_def} ->
+      let ids, open_proof =
+        List.fold_left (fun (l,b) {Vernacexpr.fname={CAst.v=id}; body_def} ->
             id::l, b || body_def = None) ([],false) l in
-        if open_proof
-        then VtStartProof (guarantee,ids)
-        else VtSideff (ids, VtLater)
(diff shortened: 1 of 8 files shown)


gares (approved):

JasonGross:
We use the `derive` plugin in fiat-crypto and would like to continue using it

ejgallego:
I'm much afraid @JasonGross that as I mentioned before you recently started to use it, the status of the current implementation of derive is problematic and conflicts with some high priority ongoing work.

Moreover, derive functionality can be easily achieved by other means I think. We will see what others think but I would not hold my breath.

JasonGross:
@ejgallego I suppose replacing it with refine+transparent_abstract+abstract+Defined would not be terrible

ejgallego:
You could do a bit better using a projection. In fact, I guess that an alternative implementation would be maybe possible today, but I am not sure it is worth the effort.

Fixing refine would be also an option, but I am not sure we can do it without breaking the plug-in at least temporaly.

JasonGross:
Actually, to make my replacement tactic work nicely, #2386 (allowing `abstract` to succeed as long as it solves all evars) would need to be fixed; could you fix #2386 if you're going to be deprecating/breaking `Derive`?

gares:
@JasonGross if you have a minute please test #9315 against your ltac code to replace `Derive`

JasonGross:
I don't have time to test right now (and I'm not sure I'll get a chance until the 15th), but the tactic I had in mind was something like

```coq
Tactic Notation "derive" ident(val) "such" "that" constr(lemT) "as" ident(lem) "opaquely" "by" tactic3(tac) :=
  cut (sigT lemT); [ intros _ | ];
  [ |
  (* first make the body and resolve the evar; use abstract for efficiency; afterwards, we give a name to the definition and give the lemma the correct type *)
  let result := fresh "result" in
  simple refine (let val := _ in let result := existT lemT val _ in _);
  [ shelve | abstract tac | ];
  subst val;
  lazymatch (eval cbv delta [result] in result) with
  | existT _ ?body ?pf
    => clear result;
       simple refine (existT _ _ _);
       [ transparent_abstract exact body using val
       | abstract exact pf using lem ]
  end ].
```
Note that this is untested.

gares:
ping @JasonGross

JasonGross:
@gares I've now tested #9315 and gotten [an anomaly](https://github.com/coq/coq/pull/9315#issuecomment-456293383) and [a proof of `False` only rejected at `Qed`-time](https://github.com/coq/coq/pull/9315#issuecomment-456295006).  Not yet very usable.

stm/vernac_classifier.ml
```diff
@@ -95,36 +95,26 @@ let classify_vernac e =
     | VernacSetOption (_, ["Default";"Proof";"Using"],_) -> VtSideff [], VtNow
     (* StartProof *)
     | VernacDefinition ((Decl_kinds.DoDischarge,_),({v=i},_),ProveBody _) ->
-      VtStartProof(Doesn'tGuaranteeOpacity, idents_of_name i), VtLater
+      VtStartProof (idents_of_name i), VtLater
 
     | VernacDefinition (_,({v=i},_),ProveBody _) ->
-       let guarantee = if poly then Doesn'tGuaranteeOpacity else GuaranteesOpacity in
```
maximedenes: How do such definitions now work?

ejgallego: I guess that was used a hack to indicate "don't delegate poly proofs`. If so, that should be indeed accounted in the main classification path right?

ejgallego: This is the a part of the system that I (sadly) have no real idea, how delegation an universes work.

SkySkimmer: Do we have any tests that we do the right thing with polymorphic delegation?

ejgallego:
8.11 stuff

ejgallego:
Indeed the opacity_guarantee is not a good method to determine the delegation policy. Instead, we should rebase this on top of #9645 and then query the proof information for non-supported delegation information [poly, non-standard closing path]

However I am unsure how to accommodate that in the STM, so if anyone wants to give it a go here is this branch to push.

ejgallego:
Seems hard that someone will complete the work needed here, so closing for now, the PR stands as a reference in case anyone would like to check for proof data when deciding when to delay.

PR #9291, opened 2018-12-30
Do not take universes into account in lia reification.

### LINKED ISSUE #9268: lia works less well with universe polymorphism than omega
#### Version

8.8.2

```
Require Import Coq.ZArith.ZArith.
Require Import Coq.micromega.Lia.

Local Open Scope Z_scope.

Definition Register := Z%type.

Definition Opcode := Z%type.

Inductive InstructionI : Type
  := Lb : Register -> Register -> Z -> InstructionI
  |  InvalidI : InstructionI.

Inductive Instruction : Type
  := IInstruction : InstructionI -> Instruction.

Definition funct3_LB : Z := 0.

Definition opcode_LOAD : Opcode := 3.

Set Universe Polymorphism.

Definition MachineInt := Z.

Definition funct3_JALR := 0.

Axiom InstructionMapper: Type -> Type.

Definition apply_InstructionMapper(mapper: InstructionMapper Z)(inst: Instruction): Z :=
  match inst with
  | IInstruction   InvalidI   => 2
  | IInstruction   (Lb  rd rs1 oimm12) => 3
   end.

Axiom Encoder: InstructionMapper MachineInt.

Definition encode: Instruction -> MachineInt := apply_InstructionMapper Encoder.

Lemma foo: forall (ins: InstructionI),
    0 <= encode (IInstruction ins) ->
    0 <= encode (IInstruction ins) .
Proof.
  Set Printing Universes.
  intros.
```

opens the following goal:

```
  ins : InstructionI
  H : 0 <= encode@{Top.1295} (IInstruction ins)
  ==============================================
  0 <= encode@{Top.1296} (IInstruction ins)
 ```

On this goal, `lia` fails with "Cannot find witness", while `omega` and `assumption` succeed.

#### Desired behavior:

`lia` should solve the goal.

/cc @andres-erbsen

ppedrot:
This is slightly blunt, it might be the case that we get delayed constraints that cannot be solved resulting in a later universe inconsistency, but it looks highly unlikely on arithmetical statements.

Alternatively we would have threaded the unification state, but this would have required a much deeper change.

Fixes #9268.

### DIFF
--- a/plugins/micromega/coq_micromega.ml
+++ b/plugins/micromega/coq_micromega.ml
@@ -846,7 +846,7 @@ struct
      match env with
       | [] -> ([v],n)
       | e::l ->
-         if EConstr.eq_constr sigma e v
+         if EConstr.eq_constr_nounivs sigma e v
          then (env,n)
          else
           let (env,n) = _add l ( n+1) v in
--- a/test-suite/bugs/closed/bug_9268.v
+++ b/test-suite/bugs/closed/bug_9268.v
@@ -1,0 +1,46 @@
+Require Import Coq.ZArith.ZArith.
+Require Import Coq.micromega.Lia.
+
+Local Open Scope Z_scope.
+
+Definition Register := Z%type.
+
+Definition Opcode := Z%type.
+
+Inductive InstructionI : Type
+  := Lb : Register -> Register -> Z -> InstructionI
+  |  InvalidI : InstructionI.
+
+Inductive Instruction : Type
+  := IInstruction : InstructionI -> Instruction.
+
+Definition funct3_LB : Z := 0.
+
+Definition opcode_LOAD : Opcode := 3.
+
+Set Universe Polymorphism.
+
+Definition MachineInt := Z.
+
+Definition funct3_JALR := 0.
+
+Axiom InstructionMapper: Type -> Type.
+
+Definition apply_InstructionMapper(mapper: InstructionMapper Z)(inst: Instruction): Z :=
+  match inst with
+  | IInstruction   InvalidI   => 2
+  | IInstruction   (Lb  rd rs1 oimm12) => 3
+   end.
+
+Axiom Encoder: InstructionMapper MachineInt.
+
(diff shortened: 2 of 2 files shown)


fajb (approved):
It would be nice to eventually get a sound and precise reification code.
(BTW, other tactics e.g. 'rtauto' and 'congruence' also seem to be affected.)

That  said, the patch is fine by me.

maximedenes:
This sounds a bit like procrastination to me. @ppedrot can you give an evaluation of the amount of work needed to do things properly here?

ppedrot:
@maximedenes A few hours of work, I guess. That said, it is not clear to me that threading the universe state is sounder than just forgetting it at reification time, as in the latter case the constraints are recomputed at unification time. Furthermore computing them upfront might be quite costly. Universe unification has a big constant overhead.

maximedenes:
> as in the latter case the constraints are recomputed at unification time

Ah, you mean when we convert the goal to the evaluation of the reified form?

Indeed, so it could mean that reification should not distinguish universes. Was the same question already studied for some other reification code after universe constraint inference was moved out of the kernel?

fajb:
I am a bit confused by the discussion. I am no universe experts but, it seems to me that:
- 'EConstr.eq_constr' is a too strong relation (as shown by the current bug)
- 'EConstr.eq_constr_nounivs' is too weak (as mentioned by @ppedrot , this may trigger some universe inconsistencies.) 

My initial understanding was that threading the unification state (and applying the unification constraints on the fly) would only merge the relevant terms. In case of unification failure, 
a novel variable would be added to the environment. The reification would always succeed but the result would depend, in general,  on the traversal order of the arithmetic term.

Last remarks. I have no idea how this can be implemented... This also seem to put some burden on any reification code. Could there be some existing normalisation mechanism that could be of some help?

ppedrot:
I would still advocate for merging this code as-is. I am somewhat convinced that the alternative implementation threading the state would behave observationally the same, the only difference being the time at which the universe failure is raised internally (i.e. in particular it's the same to the tactic end user).

PR #9292, opened 2018-12-31
Fixing some wrong scopes of variables in the interpretation of the "in" clause of a "match"

herbelin:
**Kind:** bug fix

This fixes two small bugs I noticed in using a clause `in indname pat1 ... patn` in `match`:
- `indname` was considered as a binding name
- `pat as x` bindings in the patterns were ignored


ejgallego (approved):
LGTM tho not sure about the milestone.

PR #9293, opened 2019-01-01
Do not rename implicit argument names.

### LINKED ISSUE #6785: The initial implicit argument names can differ from the specified names
#### Version

8.8 master hash 8dd6d09 and 8.7.0

```
Definition foo (a a : nat) : nat := 3.
Fail Arguments foo a a.
(*
To rename arguments the "rename" flag must be specified.
Argument a0 renamed to a.
*)

Fail Arguments foo a a : rename.
(* Some argument names are duplicated: a *)

Arguments foo a a0 : assert.
Arguments foo a b : rename.
```

This behavior seems confusing and fragile. Would it be possible to allow repeated names for argument specifications, and use the first one on an implicit application `(a := 2)`? As opposed to actual binders, I don't see a problem with shadowing uses in later types. One issue I do see is that queries like `About` and `Print` will show the renamed varaibles, but that is the same situation as after an explicit `Arguments foo c d`, for which `About` gives
```
foo : nat -> forall a0 : nat, a0 = a0
Arguments are renamed to c, d
...
```
Having instead ```Arguments are named a, a``` always printed doesn't seem too different.

---

One effect of the option in #6582 is to also mangle default argument names, since they get refreshed (with respect to the local env. not the global env.). This caused problems working on #6781.
I think that changing the above behavior would be an improvement to the system, making it easier to understand and work with, but the argument could also be made that #6582 is too agressive here.

### LINKED ISSUE #8790: -mangle-names fails on inductive definition
#### Version

#### Operating system
Linux

Consider this example file:
```coq
From Coq Require Import List.
Export ListNotations.

```

Unfortunately, it fails:
```
```

jashug:
Automatic implicit arguments get the name given by the user,
if they are inferable and no previous binder used the same name.
Otherwise, they are marked uninferable, because the current system
requires all implicit arguments to have a name.

This changes the implicit status of the motive in `ex_ind`, for example,
because the motive is rigid but is not given a name (the most obvious
name, `P`, also conflicts with the second argument of `ex`).
The hope is that this sort of change is unusual enough that CI has few problems.

Allowing anonymous implicit arguments would be a step forward, fixing
this case and others, but is more disruptive and could introduce other
incompatibilities.

If necessary, a compatibility option could be introduced which uses the old naming algorithm in `find_displayed_name_in`, at the expense of keeping the `all` parameter around for a while.

**Kind:** bug fix

Fixes #8790
Fixes #6785

Removes the immediate need for #7254, which got stuck. I still think the basic idea is good, but I obviously underestimated how difficult it would be.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,10 @@ Specification language, type inference
   solved by writing an explicit `return` clause, sometimes even simply
   an explicit `return _` clause.
 
+- The automatic names for implicit arguments now never differ from those
+  given by the user, but we set fewer arguments to be implicit when using
+  the option `Implicit Arguments`.
+
 Notations
 
 - New command `Declare Scope` to explicitly declare a scope name
--- a/interp/impargs.ml
+++ b/interp/impargs.ml
@@ -21,7 +21,6 @@ open Libobject
 open EConstr
 open Reductionops
 open Constrexpr
-open Namegen
 
 module NamedDecl = Context.Named.Declaration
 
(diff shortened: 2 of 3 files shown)


jashug:
This didn't work as well as I hoped, lots of messages saying "Argument <> cannot be made implicit".

jashug:
Gitlab pipeline: https://gitlab.com/coq/coq/pipelines/41788995

PR #9297, opened 2019-01-02
Two fixes in printing notations with patterns

herbelin:
**Kind:** bug fix

This fixes two small notation printing bugs in the presence of patterns:
- A pattern construction was escaping its surrounding `try` because of lazyness, leading to a `Not_found` thrown too late: we force evaluation.
- Parameters were not taken in constructors when checking that a pattern used as a binder was the same as a pattern used as term (see example in test-suite).

### DIFF
--- a/pretyping/glob_ops.ml
+++ b/pretyping/glob_ops.ml
@@ -485,7 +485,11 @@ let is_gvar id c = match DAst.get c with
 | GVar id' -> Id.equal id id'
 | _ -> false
 
-let rec cases_pattern_of_glob_constr na = DAst.map (function
+let rec cases_pattern_of_glob_constr na c =
+  (* Forcing evaluation to ensure that the possible raising of
+     Not_found is not delayed *)
+  let c = DAst.force c in
+  DAst.map (function
   | GVar id ->
     begin match na with
     | Name _ ->
@@ -498,6 +502,8 @@ let rec cases_pattern_of_glob_constr na = DAst.map (function
   | GApp (c, l) ->
     begin match DAst.get c with
     | GRef (ConstructRef cstr,_) ->
+      let nparams = Inductiveops.inductive_nparams (fst cstr) in
+      let _,l = List.chop nparams l in
       PatCstr (cstr,List.map (cases_pattern_of_glob_constr Anonymous) l,na)
     | _ -> raise Not_found
     end
@@ -505,7 +511,7 @@ let rec cases_pattern_of_glob_constr na = DAst.map (function
      (* A canonical encoding of aliases *)
      DAst.get (cases_pattern_of_glob_constr na' b)
   | _ -> raise Not_found
-  )
+  ) c
 
 open Declarations
 open Term
--- a/interp/notation_ops.ml
+++ b/interp/notation_ops.ml
@@ -908,11 +908,8 @@ let bind_term_as_binding_env alp (terms,termlists,binders,binderlists as sigma) 
     (* TODO: look at the consequences for alp *)
     alp, add_env alp sigma var (DAst.make @@ GVar id)
 
-let force_cases_pattern c =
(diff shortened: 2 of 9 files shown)


pretyping/glob_ops.ml
```diff
@@ -484,7 +484,8 @@ let is_gvar id c = match DAst.get c with
 | GVar id' -> Id.equal id id'
 | _ -> false
 
-let rec cases_pattern_of_glob_constr na = DAst.map (function
+let rec cases_pattern_of_glob_constr na c =
+  DAst.make (match DAst.get c with
```
ejgallego: Why is the `map` removed here? Note that `map` preserves attributes in the AST whereas the current pattern does not.

ejgallego (commented):
LGTM however I am not sure I understand why the `DAst.map` was removed.

pretyping/glob_ops.ml
herbelin: I need to force evaluation and `DAst.map` keeps "thunks" unevaluated. But you are right that the loc is lost and I re-added it. Alternatively, we may have a `map_force` in `DAst` (in anticipation of more attributes being added)?

ejgallego: I guess we could add `DAst.force` instead.

herbelin: You mean writing `let rec cases_pattern_of_glob_constr na c = DAst.map (function ...) (DAst.force c)`?

ejgallego: For example, or even:
```
let rec cases_pattern_of_glob_constr na = DAst.map ...

let cases_pattern_of_glob_constr na c = cases_pattern_of_glob_constr na (force c)

(* or *)
let cases_pattern_of_glob_constr na c = force (cases_pattern_of_glob_constr na c)
```
etc...

ejgallego: The best would likely be to stop relying on exceptions [which seems like a field of mines in the lazy setting], but I guess that is pretty hard to do.

mattam82 (approved):
Looks fine to me too.

pretyping/glob_ops.ml
herbelin: The problem is that I need more than evaluation of the first level of `DAst`. There might be `Thunk` in the inner levels. Forcing only the first level was actually what I was doing before but this is not enough.

Exceptions are indeed a field of mines in a lazy setting.

ejgallego: > The problem is that I need more than evaluation of the first level of `DAst`.

Ah, I see, then I guess it is OK to do as you propose [a force on each recursive call]

ejgallego (changes requested):
Please:

- Preserve the `DAst.map` call and instead add a `DAst.force` function.
- Document a bit more the delicate interplay of exceptions and laziness.

ejgallego:
@herbelin let me know if you wanna tweak this patch, otherwise I could force-push the nits.

ejgallego:
Tagging as to reflect that this needs [very small] progress.

herbelin:
@ejgallego: is the new version using a `DAst.force` ok for you?

herbelin:
@ejgallego: is the new version using a `DAst.force` ok for you?

ejgallego (approved):
Looks good, thank @herbelin and sorry for the delay on this; will merge as soon as possible.

ejgallego:
@herbelin it looks like the test suite fails [ups]

herbelin:
Was a merge typo. Only unrelated output-modulo-time/ltacprof_cutoff.v now failing (and elpi for unrelated reasons).

ejgallego:
> Was a merge typo. Only unrelated output-modulo-time/ltacprof_cutoff.v now failing (and elpi for unrelated reasons).

Yup, thanks for the quick fix; will merge as soon as CI finishes.

Issue #9296, opened 2019-01-02
False positives in auto template polymorphism warning


SkySkimmer:
eg with `Inductive Foo := foo : Type -> Foo.` it doesn't matter whether it's declared template or not, so why warn?

See also https://github.com/coq/coq/pull/8488#issuecomment-450424437

ppedrot:
Nowadays,
```coq
#[universes(template)]
Inductive Foo := foo : Type -> Foo.
```
results in the error `Ill-formed template inductive declaration: not polymorphic on any universe.` Maybe we can consider this closed?

SkySkimmer:
There's still the Prop ones.

ppedrot:
Well, according to that point of view, #14471 would close this for good then?

SkySkimmer:
It should

SkySkimmer:
seems fixed

PR #9304, opened 2019-01-04
Default disable auto template warning.

SkySkimmer:
The situation is too unclear to make it of general use, plus it has
some issues (#9296)
I'm not deleting the warning as it can still be useful to find which
types are template for those who want to experiment.

### DIFF
--- a/doc/sphinx/language/cic.rst
+++ b/doc/sphinx/language/cic.rst
@@ -1041,6 +1041,12 @@ in :math:`\Type`.
    enabled it will prevail over automatic template polymorphism and
    cause an error when using the ``template`` attribute.
 
+.. warn:: Automatically declaring @ident as template polymorphic.
+
+   Warning ``auto-template`` can be used to find which types are
+   implicitly declared template polymorphic by :flag:`Auto Template
+   Polymorphism`.
+
 If :math:`A` is an arity of some sort and :math:`s` is a sort, we write :math:`A_{/s}`
 for the arity obtained from :math:`A` by replacing its sort with :math:`s`.
 Especially, if :math:`A` is well-typed in some global environment and local
--- a/vernac/comInductive.ml
+++ b/vernac/comInductive.ml
@@ -35,7 +35,7 @@ module RelDecl = Context.Rel.Declaration
 (* 3b| Mutual inductive definitions *)
 
 let warn_auto_template =
-  CWarnings.create ~name:"auto-template" ~category:"vernacular"
+  CWarnings.create ~name:"auto-template" ~category:"vernacular" ~default:CWarnings.Disabled
     (fun id ->
        Pp.(strbrk "Automatically declaring " ++ Id.print id ++
            strbrk " as template polymorphic. Use attributes or " ++


maximedenes (approved):
I also think that having this warning on by default is too confusing for beginners. Imagine that they have to learn what template polymorphism is, just to understand what the system replies when they define the list type...

mattam82 (approved):
I agree, while it is useful for experts it is puzzling for beginners, especially as it has an effect "at a distance", on other definitions. I'll merge quickly if nobody opposes.

herbelin:
Side question in passing: what happens if we turn non-template polymorphism on by default. At first view, the two cases where the behavior would differ from template polymorphism are rare and, luckily, we might get an almost transparent migration user-side?

mattam82:
I don't think they're that rare. If you use a container like `list : Type -> Type` to define a concrete term structure in `Set` (e.g. with a constructor `App : term -> list term -> term`), without template polymorphism the `term` definition fails.

herbelin:
Ah, interesting. Is this a kind of example which requires cumulativity? Do you have a stand-alone example so that we can better understand the issue?

mattam82:
Just defining `Inductive term : Set := var : nat -> term | app : term ->
list term -> term` is enough. I’m not sure what you mean by requiring
cumulativity, it certainly requires template(parameter) or full
polymorphism to get that `list term : Set`.
Le ven. 4 janv. 2019 à 12:16, Hugo Herbelin <notifications@github.com> a
écrit :

herbelin:
Apparently, I don't get it. If I do:
```coq
Polymorphic Inductive list A := nil | cons : A -> list A -> list A.
Polymorphic Inductive term : Set := var : nat -> term | app : term -> list term -> term.
```
it works.

SkySkimmer:
I think @mattam82 understood your question as "what breaks if we replace template by no polymorphism"

Zimmi48 (dismissed):
While I approve this PR, disabling this warning by default make it all the more important to document it (otherwise it will be useless). I request that this PR document this warning and an appropriate place to do so seems to be https://coq.gitlab.io/-/coq/-/jobs/141375851/artifacts/_install_ci/share/doc/coq/sphinx/html/language/cic.html#template-polymorphism.

herbelin:
> I think @mattam82 understood your question as "what breaks if we replace template by no polymorphism"

Ah, sorry for the misunderstanding.

>  I think you need `Polymorphic Cumulative` otherwise you would force some universe unifications that didn't appear in the polymorphic case, otherwise it's mostly for instantiations by `Prop` that it would break.

Then, wouldn't it be interesting, from the practical point of view, to switch the default from template to polymorphic cumulative, hoping that it would then be needed to add a, say, `Prop-inclusive` attribute in only a very few places, and that everything would go essentially smoothly otherwise?

Then, there would be no need to bother users with intricacies of the different variants of polymorphism in cases which do not matter for them.

mattam82:
Ideally, yes! I've no time to test this now but the experiment would be instructive.

herbelin:
> I've no time to test this now but the experiment would be instructive.

Apparently, making a very quick experiment of setting `poly` to `true` and `template` to false in `Vernacentries.vernac_inductive`, a first obstacle is that hints do not seem very universe-polymorphism-proof. If `eq` becomes normally polymorphic, then `auto` seems to fail to apply `eq_refl`. But I guess, that this is the kind of things which we would like to make working anyway.

mattam82:
It should work better with cumulativity on as well. Otherwise mixing monomorphic definitions with polymorphic inductives will surely result in inconsistencies. I believe eq’s universe should become irrelevant if made polymorphic.

herbelin:
> It should work better with cumulativity on as well. Otherwise mixing monomorphic definitions with polymorphic inductives will surely result in inconsistencies. I believe eq’s universe should become irrelevant if made polymorphic cumulative.

Having normally (cumulative) polymorphic inductive types by default seem indeed to require having e.g. `Hint` over constructors polymorphic by default too (which would solve the `eq_refl` limitation I mentioned previously).

So, continuing the thread started [here](https://github.com/coq/coq/pull/8488#issuecomment-451251210), what to target?

Do we continue to provide a strictly compatible notion of template polymorphism built on top of regular polymorphism? If yes, this means at least:
- implementing a `Prop-inclusive` feature in minimization
- implementing a `Parameter-only` polymorphism feature
- automatically turning polymorphic `Hint` about constructors of polymorphic types

Or do we consider that the concept of template polymorphism is not worth being maintained as a stand-alone concept, and that the only feature of the current implementation of template polymorphism which is worth being kept as a proper concept - as suggested [here](https://github.com/coq/coq/pull/8488#issuecomment-451288218) - is the one of `Prop-inclusive`?

Then, what polymorphism mode by default?

I would find tedious for users that they have to write explicitly for each inductive type that they want them polymorphic. Somehow, being polymorphic is the "natural" expectation and it is the reverse which should require an effort.

I would then personally lean towards having cumulative polymorphism on by default at least for inductive types, thus also for hints on constructors. Would there be reasons not to activate it by default also for definitions and all other hints? (Would it be e.g. too costly?)

If I understand @mattam82 correctly, you are open to various approaches but you are unsure that no surprise will come if we go towards a model with full cumulative polymorphism on by default.

Wrt warnings, we could have no warning at all: users who know what they want would simply add a `Prop-inclusive` attribute where they want polymorphism to range also over `Prop`. They would also use explicit `@{...}` annotations if they want a customization of polymophism which is not the default one.

Or we could also work in two steps and have a warning in the first phase for every situation where template polymorphism would not work like the mode which we plan to eventually have on by default in the second phase. If the latter is full cumulative polymorphism for all of inductive types, definitions, hints, ... this would mean warning where `Prop-inclusive` would matter as well as where `Parameter-only` would matter.

SkySkimmer:
@Zimmi48 changed

Zimmi48 (approved):

Zimmi48:
@herbelin As someone who is not used to think too much about universes (which must be the case of a good number of our users), I support any attempt to reduce the variety of options. Cumulative polymorphic inductive types (not just on parameters), with or without minimization to `Prop` by default, sounds right. Making everything universe polymorphic by default sounds right as well if it can be done without too much pain and without a too big performance cost. (Same for primitive projections BTW.)

palmskog:
Correct me if I'm wrong here, but wouldn't making everything universe polymorphic completely remove the possibility of fine-grained parallelism using `.vio` or `.vos`? 

For example, we tried to apply parallel fine-grained regression proving to John Wiegley's category theory [development](https://github.com/jwiegley/category-theory/) which heavily uses universe polymorphism, and basically no proof tasks at all showed up in the `.vio` files. Any kind of regression proving would just collapse into parallel `.vo` compilation, which is slow for most projects compared to `.vio`/`.vos`.

As a Coq user and project maintainer, I would prefer if template polymorphism was a completely silent option: not enabled by default and no "missing annotations" warnings.

SkySkimmer:
If we restrict "everything" to inductive types it's OK.

Zimmi48:
@SkySkimmer You forgot to update `CHANGES` in this PR...

PR #9306, opened 2019-01-04
Remove Printing Primitive Projection Compatibility

SkySkimmer:
The code to generate the legacy bodies is moved to its only user in
extraction.

It almost seems like we could remove it (ie no special extraction code
for primitive projection constants) but then we run into issues with
automatic unboxing eg `Record foo := { a : nat; b : a <= 5 }.` gets
extracted to `type foo = nat` and (if we remove the special code) `let
a = a`.

### DIFF
--- a/doc/sphinx/language/gallina-extensions.rst
+++ b/doc/sphinx/language/gallina-extensions.rst
@@ -246,11 +246,6 @@ Primitive Projections
    printing time (even though they are absent in the actual AST manipulated
    by the kernel).
 
-.. flag:: Printing Primitive Projection Compatibility
-
-   This compatibility option (on by default) governs the
-   printing of pattern matching over primitive records.
-
 Primitive Record Types
 ++++++++++++++++++++++
 
@@ -296,8 +291,8 @@ the folded version delta-reduces to the unfolded version. This allows to
 precisely mimic the usual unfolding rules of constants. Projections
 obey the usual ``simpl`` flags of the ``Arguments`` command in particular.
 There is currently no way to input unfolded primitive projections at the
-user-level, and one must use the :flag:`Printing Primitive Projection Compatibility`
-to display unfolded primitive projections as matches and distinguish them from folded ones.
+user-level, and there is no way to display unfolded projections differently
+from folded ones.
 
 
 Compatibility Projections and :g:`match`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -150,6 +150,8 @@ Misc
 
 - Option "Typeclasses Axioms Are Instances" is deprecated. Use Declare Instance for axioms which should be instances.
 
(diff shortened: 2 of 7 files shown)


ppedrot (approved):
I am supportive of this change. I regularly stumble on this piece of code, wish to remove it and realize the printing option is still there.

mattam82 (approved):

herbelin:
I'm ok with this PR, but native projections seem to still be under-exploited. A roadmap would probably be useful. For instance, it seems that it would be a good objective to move the most standard projections (e.g. `fst`, `snd`, `proj1_sig`, `proj2_sig`,`projT1_sig`, `projT2_sig`, ...) to a compact representation with minimal compatibility break. If we can do it, users will be able to do it for their own records with minimal efforts. Is it a kind of objective that we could target for 2019?

maximedenes (approved):

maximedenes:
This needs a `CHANGES` entry. Also, was the option undocumented?

SkySkimmer:
>was the option undocumented

It was incorrectly documented as on by default.

doc/sphinx/language/gallina-extensions.rst
```diff
@@ -296,8 +291,8 @@ the folded version delta-reduces to the unfolded version. This allows to
 precisely mimic the usual unfolding rules of constants. Projections
 obey the usual ``simpl`` flags of the ``Arguments`` command in particular.
 There is currently no way to input unfolded primitive projections at the
-user-level, and one must use the :flag:`Printing Primitive Projection Compatibility`
-to display unfolded primitive projections as matches and distinguish them from folded ones.
+user-level, and there is no way to display unfolded projections differently
+from folded ones.
```
SkySkimmer: I just noticed this change as sphinx was failing due to missing ref. This is pretty bad IMO, maybe we should delay the PR.

Zimmi48: Impressive what the user documentation (with fatal warnings) can do to help the developers :smiley:

Zimmi48 (approved):
Change looks OK beyond the issue noted by Gaëtan. In particular, since it is a printing option, and thus has no semantics impact, it feels acceptable to remove it without deprecating it first.

SkySkimmer:
If nobody decides to merge this within a week I'll take it as a decision to keep the option for now, and so I'll close the PR.

ppedrot:
@maximedenes it seems that there is no opposition, isn't there? If you don't have time to merge this in @SkySkimmer's window I'll do it.

maximedenes:
> it seems that there is no opposition, isn't there

Indeed, so let's merge. Sorry, I missed this PR because of the "needs: discussion" label. But IIRC, it was discussed during the last WG. Or at least the general plan around primitive projections was.

PR #9307, opened 2019-01-05
Handle local definitions in implicit arguments of Instance

### LINKED ISSUE #9300: Implicit binders of Instances can be shifted
#### Version

coq master 2a7992f75c86a15512568ac61ca4c43e23242b28

```coq
Existing Class True.

Instance foo {n : nat} (x := I) : forall {b : bool} (s : nat * nat), True. auto. Defined.

About foo.
(*
foo : nat -> let x : True := I in bool -> nat * nat -> True
Arguments n, s are implicit and maximally inserted
*)
Check foo (n := 3) true (s := (4 , 5)). (* foo true : True *)
```

Notice that b was marked implicit, and s was not, but they got switched.

The issue is at https://github.com/coq/coq/blob/0a5bbf347b5bbcb579f94eb3d0166778cd92cfdb/vernac/classes.ml#L307-L308 `len` should be using `nhyps` rather than counting local definitions as well.

jashug:
**Kind:** bug fix

Fixes #9300

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -93,6 +93,9 @@ Vernacular commands
 - The naming scheme for anonymous binders in a `Theorem` has changed to
   avoid conflicts with explicitly named binders.
 
+- Computation of implicit arguments now properly handles local definitions in the
+  binders for an `Instance`.
+
 Tools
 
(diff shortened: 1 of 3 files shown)


maximedenes (approved):
Looks good, thanks!

PR #9309, opened 2019-01-05
[ci] Add Verdi Raft with dependencies to CI

palmskog:
**Kind:** infrastructure.

Note that two of the dependencies (but not Verdi Raft itself) depends on positive co-inductive types. We hope to address this in the near future.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -506,6 +506,9 @@ library:ci-sf:
 library:ci-unimath:
   <<: *ci-template-flambda
 
+library:ci-verdi-raft:
+  <<: *ci-template-flambda
+
 library:ci-vst:
   <<: *ci-template-flambda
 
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -509,6 +509,9 @@ library:ci-unimath:
 library:ci-vst:
   <<: *ci-template-flambda
 
+library:ci-verdi-raft:
```
ejgallego: Please keep the entries in alphabetical ordering.

ejgallego (approved):
LGTM, in the future we may want to split some deps off; but for now things are good IMHO.

.gitlab-ci.yml
palmskog: Fixed.

silene:
I won't be backporting to 8.9 a CI test that depends on a master branch.

PR #9311, opened 2019-01-06
Printing notations with variables used both for binders and terms

herbelin:
**Kind:** bug fix

~~Depends on #9292 and #9297.~~ Merged.

This addresses item `foo1` in the middle of this [comment](https://github.com/coq/coq/pull/9207#issuecomment-448204714) in #9207.

This PR is partial work about a feature which I'm currently not completely sure that it is worth the additional implementation cost. It is about correctly printing a notation containing a variable which is used both in terms and in a binding structure.

Without this PR, the status is that such mixed binding/term notations are printed correctly only when the term comes before the binder. For instance, the following is already working:
```coq
Notation "'incr' x ; P" := (let x := x+1 in P) (at level 200).   
Check fun x => incr x; x=0.
```
but the following is not:
```coq
Notation "'incr' x ; P" := ((fun x => P) (x+1)) (at level 200).   
```
because the term comes after the binder (*).

Other known similar typical examples are:
- From Iris:
  ```coq
  Notation "'tele' x .. z := b" :=
    (fun x => .. (fun z =>
       pack (fr (fun x => .. ( fr (fun z => fb b) ) .. ) )
            (existT _ x .. (existT _ z tt) .. )
                  ) ..)
    (at level 85, x binder, z binder).
  ```
  but when we look more closely at it, we see that it is only about two binding structures intended     to share the same names: the `x` and `z` in `existT` are not free terms, they are bound by the `x` and `z` above them (this is a working case).
- [Example](https://github.com/coq/coq/pull/9207#issuecomment-448204714) `foo1` given by @ggonthier in #9207. I thought that it was a realistic example from MathComp but the discussion at math-comp/math-comp#265 suggests that the real example is different and I'm now a bit confused about whether there are real cases in MathComp.
- One of my other original motivation was to be able to define notations for expressions of the form `Notation "☐_ x P" := (forall x' >= x,  (fun x => P) x')`. But, to be satisfactory, one would like to avoid the beta-expansion and this would require yet another mechanism (e.g. to see `x >= t` as a stand-alone binder, so that, when writing `forall x >= t, P`, `x` binds in `P` but not in `t`).

There is also the question of supporting such mixed terms and binders in abbreviations. @mattam82 expressed a spontaneous perplexity in front of #8808. My answer was not quite right since I was arguing using Iris' example above but 1) Iris' example does finally not require mixing free terms and binders 2) Iris' example does not use an abbreviation but a proper parsing rule. I'm now perplex myself about what kind of interesting abbreviation involving binders, and a fortiori involving both binder and terms, we could do.

What does the PR is to make printing working also in (*), to the price of a non-trivial added complexity. Also, it does not work yet in the more general case where the binder is a pattern. Moreover, it would also require more testing.

### DIFF
--- a/interp/notation_ops.ml
+++ b/interp/notation_ops.ml
@@ -736,13 +736,15 @@ let is_bindinglist_meta id metas =
   try match Id.List.assoc id metas with _,NtnTypeBinderList -> true | _ -> false
   with Not_found -> false
 
-type term_matching_status =
+type binding_meta_subst = (Id.t * (Id.t * Id.t)) list
+
+type 'a term_matching_status =
   (* The term to which is associated the meta-variable cannot be changed *)
-  | Fixed
+  | Fixed of 'a glob_constr_g * binding_meta_subst
   (* The term to which is associated the meta-variable has been
      obtained by matching a binding structure; it can still be
      alpha-converted but avoiding the given list of names *)
-  | UpToAvoiding of Id.t list
+  | UpToAvoiding of Id.t * Id.t list
 
 exception No_match
 
@@ -754,7 +756,7 @@ let rec alpha_var id1 id2 = function
 
 let alpha_rename alpmetas v =
   if alpmetas == [] then v
-  else try rename_glob_vars alpmetas v with UnsoundRenaming -> raise No_match
+  else try rename_glob_vars (List.map snd alpmetas) v with UnsoundRenaming -> raise No_match
 
 let add_env (alp,alpmetas) (terms,termlists,binders,binderlists,fvs) var v =
   (* Check that no capture of binding variables occur *)
@@ -773,9 +775,8 @@ let add_env (alp,alpmetas) (terms,termlists,binders,binderlists,fvs) var v =
      have to check that "fun z => ... z ..." denotes the same term as
      "fun x => ... x ... ?var ... x" up to alpha-conversion when [var]
      is instantiated by [v] *)
-  let v = alpha_rename alpmetas v in
   (* TODO: handle the case of multiple occs in different scopes *)
-  ((var,(v,Fixed))::terms,termlists,binders,binderlists,fvs)
+  ((var,Fixed (v,alpmetas))::terms,termlists,binders,binderlists,fvs)
 
 let add_termlist_env (alp,alpmetas) (terms,termlists,binders,binderlists,fvs) var vl =
   if List.exists (fun (id,_) -> List.exists (occur_glob_constr id) vl) alp then raise No_match;
@@ -832,16 +833,16 @@ let rec pat_binder_of_term t = DAst.map (function
   | _ -> raise No_match
   ) t
 
-let unify_name_upto alp na na' =
+let unify_name_upto var alp na na' =
   match na, na' with
   | Anonymous, na' -> alp, na'
   | na, Anonymous -> alp, na
   | Name id, Name id' ->
      if Id.equal id id' then alp, na'
-     else (fst alp,(id,id')::snd alp), na'
+     else (fst alp,(var,(id,id'))::snd alp), na'
 
-let unify_pat_upto alp p p' =
-  try fold_cases_pattern_eq unify_name_upto alp p p' with Failure _ -> raise No_match
+let unify_pat_upto var alp p p' =
+  try fold_cases_pattern_eq (unify_name_upto var) alp p p' with Failure _ -> raise No_match
 
 let unify_term alp v v' =
   match DAst.get v, DAst.get v' with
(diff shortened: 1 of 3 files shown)


ggonthier:
There is indeed a legitimate use case for this in mathcomp: the notation `[max G | P(G)]` for `G` is a maximal (sub)group such that `P(G)` holds, for example `[max A | (A <| G) && abelian A]` means `A` is a self-centralising normal subgroup of `G` (there's also a `[min G | P(G)]` notation).
  This would always be printed with the more general `[max G as G0 | P(G0)]` notation; I attributed this to the `constr` to `glob` transformation doing too much alpha-conversion (to satisfy the Barendregt convention), but realised while investigating #9207 that swapping the arguments of the underlying definition (`maxgroup`) would make it work. I've since done this, though I yet have to propagate the change to the `finset` predicates on which these are based (there's no special notation for the latter, partly due to the problem we're discussing).
   So, strictly speaking this PR isn't absolutely needed for Mathcomp; but it would have to be replaced by documenting the argument order restriction, either in the manual or by emitting a self-explanatory warning, so "doing things right" might be preferable...
   There are some problems with the PR code as it stands - I'll point them out in a proper review once I've got a better grip on it.

ejgallego:
Adding a review tag to reflect @ggonthier 's comment.

ejgallego:
This also got sadly stuck on the pending review and now a needed rebase; as in #9207 I am tentatively delaying, but don't hesitate to complain if you believe this should be backported to 8.10.

ggonthier (commented):
Sorry about the review delay - when I last commented I noted some of the code logic seems off, but it took me several ons and offs to figure out what was really going on...
  First to answer @herbelin's initial question: yes, this does stem from a genuine use case. MathComp has a predicate transformer `max_group` that maps a property `P` of groups the the property "is maximal such that `P` holds".  This transformer is most often used to state "`G` is maximal subject to `P`". There's notation for that, `[max G | P] := max_group (fun G => P) G`, 
in which `G` is both bound and free. This is a fairly common pattern in classic maths (`x` is minimal/maximal/free/initial.... subject to ....). Now that notation was never recognised, because the notation matching algorithm fixes the value of a notation ident parameter at the first occurrence and uses this everywhere else, substituting bound occurrences as it goes. Because `detype` will always rename away the bound occurrence of `G`, and that occurs first, the notation never matches. I've since "fixed" this by swapping the `max_group` arguments, but this is unsatisfactory as `max_group` no longer has the type of a predicate transformer.
  This PR does fixes this use case by marking assignments of bound variables to parameters and overriding them if the need to assign a free variable arises later. Apart from a missing substitution noted below and fixed in @herbelin's local repository, the PR is correct and could be merged (and doing so in 8.10 would let MathComp "unkludge" `max_group` sooner).
   However reviewing this PR mad me stumble on several related notation matching issues, which I would like to point out here. The logic of the code is quite cunning, because it seems to omit all variable capture checks, where as these are in fact carried out in the `glob_ops` substitution code, which expects a, ordered substitution list, scans all terms under a scope, and raises `UnsoundRenaming` if it spots a capture. That approach creates several issues:
1. `rename_glob_vars` will throw if it spots a conflict between named bound variables, or _any_ conflict with a `match` variable; all of these could be resolved by renaming bound variables, and may be _caused_ by the renaming performed by `detype`.
2. The `notation_ops` also calls `rename_var` directly, and does _not_ trap `UnsoundRenaming` in that case.
```
Notation "'foo' x y" := ((forall y, y = tt) /\ (forall x y : unit, x = y))
                        (at level 10, x ident, y ident).
Lemma bar (x0 : unit) : foo x0 x1. 
(* Error: Anomaly "Uncaught exception Glob_ops.UnsoundRenaming." *)
```
3. Conversely the  `glob_ops` code does not check for the capture of global constants. In principle these can always be avoided by qualifying the constants, but the display code cannot do this as it does not keep track of bound variables in notation arguments.
```
Notation "'foo' x y" := ((forall x : unit, x = x) /\ (forall x : unit, x = y))
                        (at level 10, x ident, y at level 8).
Notation "'foo1' x y" := ((forall x : unit, x = x) /\ (forall y : unit, y = y))
                        (at level 10, x ident, y ident).
Lemma bar : ((forall tt : unit, tt = tt) /\ (forall x, x = tt)).
 (* foo tt tt *)
 assert (foo tt tt).
   (* foo1 tt tt !!! *)
```
4. The latter problem gets weirder with abbreviations, where a bound variable may be "captured" as a pattern:
```
Notation foo x y := ((forall x : unit, x = x) /\ (forall x : unit, x = y)).
Lemma bar : ((forall tt : unit, tt = tt) /\ (forall x, x = tt)).
(* foo tt tt as above *)
assert (foo tt tt).
(* (forall 'tt, tt = tt) /\ (forall 'tt, tt = tt) !!! *)
```
5. Lastly, bound variable parameters may be set by matching terms in which they occur; this requires implementing actual glob term unification...
```
Notation foo x y E := ((forall x, E) /\ (forall y, E)).
Lemma bar (x y : unit) : foo x y (x = y).
(* (forall x0 : unit, x0 = y) /\ (forall y0 : unit, x = y0) !!! *)
```
The upshot of all of this echoes the concern in this PR's initial commit: is the added complexity of matching notation up to renaming properly worth the functionality? I would contend that, barring obviously difficult problems such as factoring common `match` branches, there is a general expectation of users that Coq should be able to display notation as it was input. As this is spoiled by the aggressive renaming performed by `detype`, it's up to `notation_ops` to compensate for this. However, one can, and perhaps should, turn the argument around: is such aggressive renaming really worth the entailed code complexity in `notation_ops`? It seems difficult to argue that such renaming would make the displayed code easier to understand than the original one, when it seems so difficult to write code that undoes the renaming and recognises the expression as was first intended...

ejgallego:
> is such aggressive renaming really worth the entailed code complexity in `notation_ops`? It seems difficult to argue that such renaming would make the displayed code easier to understand than the original one, when it seems so difficult to write code that undoes the renaming and recognises the expression as was first intended...

Indeed, the renaming performed in `Detyping`, (and in other pars of Coq) is something that feels a tad heavy sometimes. I am unsure why so much renaming is needed in detyping vs just relying on scoping rules; but I'm sure I am missing some cases.

herbelin:
@ggonthier: thanks a lot for the careful review. I then also tend to think that another approach would be better than the complexity introduced in this PR.

Regarding the renamings made in `detyping.ml`, a change of the algorithm could indeed be the fastest path to the reversibility of notation mixing terms and binders of same name. Moreover, there are complexity issues with detyping which may be another reason worth a change. (I'm not sure I have the courage to think at another algorithm myself though.)

Maybe it is exaggerated, but I somehow care about the notation algorithm being able to work independently from assumptions made on the exact choices of bound names. Another a bit hackhish but probably easier direction would be to postpone the matching of any subterm starting with a binder so that the occurrences mentioning the given variables in position of term are found first... 
This does not prevent making a change in `detyping.ml` in the meantime.

ejgallego:
Dependencies are merged; maybe it is time we can revisit this.

herbelin:
> Dependencies are merged; maybe it is time we can revisit this.

Maybe... I'm not very courageous at concentrating again on it at the current time. It is a pity because I feel that we are wasting part of our energy. We would need to make a list of priorities over convergent objectives.

ejgallego:
> We would need to make a list of priorities over convergent objectives.

Indeed, I guess we should just meet a some point and discuss a plan?

I am removing this from my to-merge list for now then; please ping me if you want to revisit this and I could help with merging.

Zimmi48:

PR #9317, opened 2019-01-08
Restrict universes in records.

### LINKED ISSUE #8076: Universes of records aren't restricted
See also #7967.

SkySkimmer:
Fix #8076.

### DIFF
--- a/kernel/context.ml
+++ b/kernel/context.ml
@@ -134,6 +134,15 @@ struct
           let ty' = f ty in
           if v == v' && ty == ty' then decl else LocalDef (na, v', ty')
 
+    let map_constr_het f = function
+      | LocalAssum (na, ty) ->
+          let ty' = f ty in
+          LocalAssum (na, ty')
+      | LocalDef (na, v, ty) ->
+          let v' = f v in
+          let ty' = f ty in
+          LocalDef (na, v', ty')
+
     (** Perform a given action on all terms in a given declaration. *)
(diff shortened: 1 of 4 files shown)


engine/eConstr.mli
```diff
@@ -321,6 +321,8 @@ val of_rel_decl : (Constr.t, Constr.types) Context.Rel.Declaration.pt -> (t, typ
 
 val to_rel_decl : Evd.evar_map -> (t, types) Context.Rel.Declaration.pt -> (Constr.t, Constr.types) Context.Rel.Declaration.pt
 
+val map_to_rel_context : (t -> Constr.t) -> (t, types) Context.Rel.pt -> (Constr.t, Constr.types) Context.Rel.pt
```
ppedrot: I am not sure it is a great idea to export this specialized function. You should rather convert it and map, or the other way around, efficiency shouldn't matter here. Otherwise we'll end up with a bloated API.

SkySkimmer: Sure whatever

mattam82 (approved):
LGTM

ppedrot (approved):
I am not super happy with the naming scheme `_het` either (for some reason that intuitively reminds me of the Semitic letter rather than *heterogeneity*), but for the time being I can tolerate this little wart. Ideally we should rather call the monomorphic version `Smart` and let the other one have the default name.

PR #9318, opened 2019-01-08
Add Azure Pipelines build badge

kaylangan:
**Kind:** documentation

Adding a build badge for the `master` branch. I'm happy to update the CHANGES.md file, but wasn't sure if you wanted to clutter it since this is a pretty small doc change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Coq
 
+[![GitLab](https://gitlab.com/coq/coq/badges/master/pipeline.svg)](https://gitlab.com/coq/coq/commits/master)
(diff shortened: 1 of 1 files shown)


Zimmi48:
Thanks for the PR :sparkles: Could you move the badge in second position (after the GitLab CI badge, since this is the one which has the most tests)? Maybe the alt-text of the GitLab CI badge should be changed as well (this is the only one not mentioning the source). No CHANGES entry is needed.

BTW, I see that you are from Microsoft (Azure team I assume). Would it be alright if we were to @-mention you in the future if we have troubles with the Azure infrastructure or doc? (We are considering doing more on Azure: moving our package build and signing processes, which are currently on Travis for macOS and on GitLab CI with custom VMs for Windows.)

Zimmi48 (approved):
Cool!

PR #9324, opened 2019-01-09
Fix #9319: coq-makefile fails badly on missing v file in _CoqProject

### LINKED ISSUE #9319: Build fails unpredictably if _CoqProject file contains missing a file
#### Version

8.10+alpha(d58614c1adb94fbfbb75e385bb75f768022b9629)

#### Operating system

macOS 10.14.2 & Debian Stretch

E.g. this

```shell
touch a.v
echo 'Require a.' > b.v
printf "a.v\nb.v\n\no_such_file.v" > _CoqProject
coq_makefile -f _CoqProject -o Makefile
make VERBOSE=1
```

produces the following output:

```shell
/Applications/Xcode.app/Contents/Developer/usr/bin/make --no-print-directory -f "Makefile" pre-all
if [ "8.10+alpha" != "8.10+alpha" ]; then\
	  echo "W: This Makefile was generated by Coq 8.10+alpha";\
	  echo "W: while the current Coq version is 8.10+alpha";\
	fi
/Applications/Xcode.app/Contents/Developer/usr/bin/make --no-print-directory -f "Makefile" real-all
make: *** [all] Error 2
```

So it does not provide any pointers as to where to look for the source of error.

Also (this happened on a real project), when there is a missing file (basically a typo in the file name) in `_CoqProject` file, then some of dependencies get resolved incorrectly and the build is failing at an unrelated moment (too early to quickly understand what has caused it).

SkySkimmer:


SkySkimmer:
Nevermind, this doesn't work with generated files so bignums fails.

PR #9325, opened 2019-01-09
Stop [Print] from saying [is (not) universe polymorphic].

### LINKED ISSUE #9056: [Print] now always displays universe polymorphism information
#### Version

```
The Coq Proof Assistant, version 8.10+alpha (November 2018)
compiled on Nov 19 2018 12:21:48 with OCaml 4.06.0
```

#### Operating system

macOS

[Print] now always displays `Monomorphic` or `Polymorphic` as well as universe polymorphism information, which is line noise and confusing when universe polymorphism is disabled.

Looks like #8515 introduced this. There's no discussion in that PR about making universe polymorphism printed by default, although it required a number of changes to expected output in the test suite.

This behavior should at least require a printing option that is disabled by default.

```
Definition foo := 0.
Print foo.
(*
Monomorphic foo = 0
     : nat

foo is not universe polymorphic
*)
```

This used to print just
```
foo = 0
     : nat
```

SkySkimmer:
[About] still says it.

Close #9056.


ejgallego:
Ping @herbelin @mattam82 ? Are you Ok with this?

maximedenes (approved):

herbelin:
I'm ok with this PR.

In the longer term, I have a vague feeling though that telling whether an object is polymorphism or not will stop being relevant. Thanks to `@{ }` and to names for global universes, we are able to define arbitrary forms of quantifications over universes and arbitrary combinations of polymorphism and monomorphism, and that a monomorphic object will just be an object which refers only to global universes. And we may stop making a difference, the same way as we don't make a difference between `Definition T := nat -> nat` which would be monomorphic and `Definition T (X:Set) := X -> X` which would be polymorphic.

ejgallego:
> I'm ok with this PR.

Thanks @herbelin , IMHO you should be the assignee / merger here.

SkySkimmer:
>In the longer term, I have a vague feeling though that telling whether an object is polymorphism or not will stop being relevant. 

I'm actually working on that right now ^^

mattam82:
I'm ok with this PR as well and agree with the long term vision, it is unnecessary in most cases and About says it.

herbelin:
> assignee / merger here.

OK, can do.

> I'm ok with this PR as well

So, shall merge tomorrow evening.

> I'm actually working on that right now ^^

Cool.

PR #9326, opened 2019-01-09
[ci] compile with -quick & validate after vio2vo

gares:
I'd love CI maintainers to help me here, since I'm coding by copy/pasting.

What I think I've done:
- one extra build with -quick plus a call to -vio2vo
- one extra validate on top of the new build

What may be worth adding as an extra "chain":
1. regular build of vo files (could we reuse the artifacts of another job?)
2. clean the .vo, keep the .aux
3. build -quick
4. -vio2vo
5. validate

The interest would be that in the current job we only delay proofs that are in an empty section (if they don't have Proof using). If we have .aux files around we delay all proofs, hence we test more.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -245,6 +245,12 @@ build:base+async:
     COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
     COQUSERFLAGS: "-async-proofs on"
 
+build:quick:
+  <<: *build-template
+  variables:
+    COQ_EXTRA_CONF: "-native-compiler no"
+    QUICK: "1"
+
 windows64:
   <<: *windows-template
   variables:
@@ -453,6 +459,11 @@ validate:edge+flambda:
   variables:
     OPAM_SWITCH: edge
     OPAM_VARIANT: "+flambda"
+
+validate:quick:
+  <<: *validate-template
+  dependencies:
+    - build:quick
 
 # Libraries are by convention the projects that depend on Coq
 # but not on its ML API
--- a/stm/vio_checking.ml
+++ b/stm/vio_checking.ml
@@ -125,6 +125,7 @@ let schedule_vio_compilation j fs =
     | s :: rest -> s :: filter_argv b rest in
   let prog = Sys.argv.(0) in  
   let stdargs = filter_argv false (List.tl (Array.to_list Sys.argv)) in
+  let all_jobs = !jobs in
   let make_job () =
     let f, t = List.hd !jobs in
     jobs := List.tl !jobs;
@@ -141,6 +142,12 @@ let schedule_vio_compilation j fs =
     Worker.kill (Pool.find pid !pool);
     pool := Pool.remove pid !pool;
   done;
+  if !rc = 0 then begin
+    (* set the access and last modification time of all files to the same t
+     * not to confuse make into thinking that some of them are outdated *)
+    let t = Unix.gettimeofday () in
+    List.iter (fun (f,_) -> Unix.utimes (f^".vo") t t) all_jobs;
+  end;
   exit !rc
 
   
--- a/stm/vio_checking.ml
+++ b/stm/vio_checking.ml
@@ -95,6 +95,7 @@ let schedule_vio_checking j fs =
     done;
     let pid, ret = Unix.wait () in
     if ret <> Unix.WEXITED 0 then rc := 1;
+    Worker.kill (Pool.find pid !pool);
     pool := Pool.remove pid !pool;
   done;
   exit !rc
@@ -137,6 +138,7 @@ let schedule_vio_compilation j fs =
     done;
     let pid, ret = Unix.wait () in
     if ret <> Unix.WEXITED 0 then rc := 1;
+    Worker.kill (Pool.find pid !pool);
     pool := Pool.remove pid !pool;
   done;
   exit !rc
--- a/Makefile
+++ b/Makefile
@@ -268,7 +268,7 @@ cleanconfig:
 distclean: clean cleanconfig cacheclean timingclean
 
 voclean:
-	find theories plugins test-suite \( -name '*.vo' -o -name '*.glob' -o -name "*.cmxs" \
+	find theories plugins test-suite \( -name '*.vo' -o -name '*.vio' -o -name '*.glob' -o -name "*.cmxs" \
 	-o -name "*.native" -o -name "*.cmx" -o -name "*.cmi" -o -name "*.o" \) -exec rm -f {} +
 	find theories plugins test-suite -name .coq-native -empty -exec rm -rf {} +
 
--- a/Makefile.build
+++ b/Makefile.build
@@ -786,8 +786,6 @@ ifdef QUICK
 	$(SHOW)'COQC -schedule-vio2vo $(NJOBS) theories/**.vio plugins/**.vio'
 	$(HIDE)$(BOOTCOQC:-compile=-schedule-vio2vo) $(NJOBS) \
 		$(THEORIESVO) $(PLUGINSVO)
-	@# this should be done by -schedule-vio2vo, see #9334
-	$(HIDE)touch $(THEORIESVO:.$(VO)=.vo) $(PLUGINSVO:.$(VO)=.vo)
 endif
 
 coqlib.timing.diff: theories.timing.diff plugins.timing.diff
--- a/Makefile.build
+++ b/Makefile.build
@@ -57,6 +57,9 @@ TIME_OF_PRETTY_BUILD_EXTRA_FILES ?= - # also output to the command line
 BEFORE ?=
 AFTER ?=
 
+# Number of parallel jobs for -schedule-vio2vo
+NJOBS ?= 2
+
 ###########################################################################
 # Default starting rule
 ###########################################################################
@@ -543,7 +546,7 @@ $(CSDPCERTBYTE): $(CSDPCERTCMO)
 
 VALIDOPTS=$(if $(VERBOSE),,-silent) -o -m
 
-validate: $(CHICKEN) | $(ALLVO)
+validate: $(CHICKEN) | $(ALLVO:.$(VO)=.vo)
 	$(SHOW)'COQCHK <theories & plugins>'
 	$(HIDE)$(CHICKEN) -boot $(VALIDOPTS) $(ALLMODS)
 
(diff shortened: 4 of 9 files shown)


.gitlab-ci.yml
```diff
@@ -245,6 +245,15 @@ build:base+async:
...
     COQUSERFLAGS: "-async-proofs on"
 
+build:quick:
+  <<: *build-template
+  stage: test
+  variables:
+    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
+    COQUSERFLAGS: "-quick"
+  script:
+    - bin/coqc -boot -schedule-vio2vo 4 `find . -name \*.vio`
```
SkySkimmer: I think this overrides the script from the build-template.

.gitlab-ci.yml
```diff
@@ -245,6 +245,15 @@ build:base+async:
     COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
     COQUSERFLAGS: "-async-proofs on"
 
+build:quick:
+  <<: *build-template
+  stage: test
```
SkySkimmer: Putting this in the test stage makes it impossible to use it in the validate job.

.gitlab-ci.yml
gares: is there a way to *append* ?

.gitlab-ci.yml
gares: ok thanks I'll fix this.

.gitlab-ci.yml
gares: I need the regular build + this extra command.

gares:
CC: #8642

.gitlab-ci.yml
SkySkimmer: I don't think so.

SkySkimmer:
It may be better to just have a combined job in the test stage with a custom script
~~~bash
./configure -local -warn-error yes -native-compiler yes
COQUSERFLAGS=-quick make -j $NJOBS
bin/coqc -boot -schedule-vio2vo 4 $(find . -name '*.vio')
~~~

.gitlab-ci.yml
```diff
@@ -245,6 +245,14 @@ build:base+async:
     COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
     COQUSERFLAGS: "-async-proofs on"
 
+build:quick:
+  <<: *build-template
+  variables:
+    COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
```
SkySkimmer: Why coqide?

.gitlab-ci.yml
gares: I found `after_script:`

SkySkimmer: Using after_script breaks the bot.

.gitlab-ci.yml
gares: removed

SkySkimmer:
Does building with `-quick` work properly? It seems fragile considering that the make rules expect it to produce vo but it will actually produce vio. For instance I'm worried about the following `make validate` deciding to rebuild some vo: imagine b depends on a, and schedule-vio2vo finishes b.vo first, then `make validate` will see b as being outdated.

gares:
> then `make validate` will see b as being outdated.

You are right, indeed I did run coqchk by hand on my machine, I'll follow your suggestion.

gares:
wait, validate does not run `make validate`, so there is no rebuilding

SkySkimmer:
>wait, validate does not run make validate, so there is no rebuilding

Right, I forgot that.
I guess the install rules don't depend on the stuff they install so that's not an issue.

SkySkimmer (dismissed):
Using after_script causes problems for the bot.

Zimmi48:
@SkySkimmer 
>Using `after_script` causes problems for the bot.

This is not the case anymore. However, `after_script` should *not* be used to append to `script` because this is not its semantics (it will run even if `script` fails).

gares:
I think the PR is ok now @SkySkimmer

.gitlab-ci.yml
```diff
@@ -245,6 +245,32 @@ build:base+async:
...
+    - echo 'start:coq.config'
+    - ./configure -local
+    - echo 'end:coq.config'
+
+    - echo 'start:coqc.quick'
+    - COQUSERFLAGS="-quick" make -j "$NJOBS"
+    - echo 'end:coqc.quick'
+
+    - echo 'start:coqc.vio2vo'
+    - bin/coqc -boot -schedule-vio2vo "$NJOBS" `find . -name \*.vio`
```
ejgallego: All this custom logic doesn't belong into the CI scripts [think for example we would like to call from Azure]

ejgallego (changes requested):

.gitlab-ci.yml
SkySkimmer: We want to run -quick from azure?

.gitlab-ci.yml
```diff
@@ -245,6 +245,32 @@ build:base+async:
...
     COQUSERFLAGS: "-async-proofs on"
 
+quick:build+validate:
+  stage: test
+  script:
+    - set -e
+
+    - echo 'start:coq.config'
+    - ./configure -local
+    - echo 'end:coq.config'
```
SkySkimmer: I wouldn't bother with those echos, they just make the script harder to read. The ones in the build template are just there on inertia from travis.

.gitlab-ci.yml
ejgallego: Do we support `-quick` on Windows?

gares: > All this custom logic doesn't belong into the CI scripts

Do we support validate on Azure? (this line is very close to the one of validate)

gares:
> All this custom logic doesn't belong into the CI scripts [think for example we would like to call from Azure]

I'm happy to fix the PR but I need to understand what you mean by "custom logic".

.gitlab-ci.yml
gares: Which dependency path? I'm sorry but I don't understand. What would be the problem on Azure in the first place?

ejgallego: Ideally the first target would be produced by the makefile just by depending on `VOFILES:.vo:=.vio`, then, the second step should also be produced from a make rule that knows how to do `.vio -> .vo`.

gares:
@ejgallego I've pushed the logic to the Makefile and tested locally:
```
$ QUICK=1 make -j2 stdlib
COQC -quick theories/Init/Notations.vo
...
COQC -quick plugins/setoid_ring/Rings_R.v
COQC -schedule-vio2vo theories/**vio plugins/**vio

$ make validate
COQCHK <theories & plugins>
```
Instead of exposing a target to build all .vio and then another one to obtain the .vo I opted for `stdlib` meaning I want all vos, and if you are in QUICK mode I'll do the vio2vo step.

@coq/ci-maintainers I get an issue on gitlab I can't quite understand, help is welcome (where does the . come from?)

SkySkimmer:
>@coq/ci-maintainers I get an issue on gitlab I can't quite understand, help is welcome (where does the . come from?)

When COQ_EXTRA_CONF is unset configure gets an empty argument.
Use `COQ_EXTRA_CONF: "-native-compiler no"` or some such thing.

SkySkimmer:
We already use QUICK in makefile for doc stuff, I don't think it's a good idea to reuse the variable.

ejgallego (approved):
I am not an expert on Makefiles but this is OK to me. I wonder why the `touch` is needed tho.

gares:
> We already use QUICK in makefile for doc stuff, I don't think it's a good idea to reuse the variable.

Isn't the same semantics? "Do as fast as you can something with less quality guarantee?"

> When COQ_EXTRA_CONF is unset configure gets an empty argument.
> Use `COQ_EXTRA_CONF: "-native-compiler no"` or some such thing.

a HUGE thanks

gares:
> I wonder why the `touch` is needed tho.

it avoids that subsequent runs of make think .vos are outdated.

it should be probably something that should be done by vio2vo, that generates .vo files in random order (or better optimal scheduling order if you have aux file around). I'll add a comment and open an issue about that.

SkySkimmer:
>Isn't the same semantics? "Do as fast as you can something with less quality guarantee?"

Is it actually quicker?

gares:
It depends on the project and the number of cores (and the Proof using annotations/aux files present).
Here on CI it is faster, but I guess it is mostly because of native-compiler no.
The figures on the odd order theorem were decent: 10x on a 12 core (page 10 https://hal.inria.fr/hal-01135919/ )

ppedrot (commented):
Not qualified to review.

gares:
Ci failure is spurious, so I guess we could consider this PR for merge. It lets us detect regressions in -quick and -schedule-vio2vo/-vio2vo and, more importantly, that `Module Include` stays out of the opaque proofs business, and one can eventually start working on coq/ceps#40 (I guess after SProp and native integers are merged, to avoid conflicts)

gares:
I was preparing a separate PR fixing #9334 but then I've decided to push the commit here

stm/vio_checking.ml
```diff
@@ -141,6 +142,12 @@ let schedule_vio_compilation j fs =
     Worker.kill (Pool.find pid !pool);
     pool := Pool.remove pid !pool;
   done;
+  if !rc = 0 then begin
```
SkySkimmer: If there's an error do we end up writing some random successful vos? If so then shouldn't we still update their times?

gares: That is the faster review ever ;-)
This code does exactly what the makefile line was doing.

I frankly don't know how to improve things. Removing some .vo files that were successfully built seems to bring nothing. But I don't get how setting the times of some .vo could help. I mean if you then manage to obtain the remaining vos, theirs date will differ anyway, confusing make.

So unless you have some other argument, I'd stay like that.

SkySkimmer: >But I don't get how setting the times of some .vo could help

If a prefix succeeds it might make sense (eg everything in Init/ succeeds but somewhere else fails, then `make states` can tell nothing needs to be done)
Not a huge gain though

SkySkimmer:
Gitlab windows job failed to download sed for some reason, I'll retry it.

gares:
ping (CI failures seem totally unrelated)

.gitlab-ci.yml
```diff
@@ -245,6 +245,12 @@ build:base+async:
     COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
     COQUSERFLAGS: "-async-proofs on"
 
+build:quick:
+  <<: *build-template
+  variables:
+    COQ_EXTRA_CONF: "-native-compiler no"
```
SkySkimmer: Do we want -native-compiler yes here?

gares: I don't think so, it does not compile opaque proofs hence we would just slow down the build imo

SkySkimmer: OK, let's merge. If someone wants native compiler they can always make a subsequent PR.

SkySkimmer:
I caught a segfault in a CI run: https://gitlab.com/SkySkimmer/coq/-/jobs/147407854
~~~
Makefile.build:821: recipe for target 'theories/Reals/Cauchy_prod.vo' failed
make[1]: *** [theories/Reals/Cauchy_prod.vo] Segmentation fault (core dumped)
~~~

PR #9328, opened 2019-01-09
Fix #8725: warn on interactive Let outside section.

### LINKED ISSUE #8725: Inconsistent warnings about toplevel Let turned into Local Definition, whether defined interactively or not

SkySkimmer:


ejgallego:
Ping @herbelin @mattam82 @maximedenes

maximedenes:
Isn't this overlapping with #8726 ?

SkySkimmer:
Looks like it.

PR #9331, opened 2019-01-09
[STM] Fix semantics of `cur_id` w.r.t. error states

maximedenes:
Installing an error state should not update `cur_id`.


maximedenes:
@gares Can you be the assignee?

gares (approved):

PR #9332, opened 2019-01-09
Add OSX job to azure

SkySkimmer:


azure-pipelines.yml
```diff
@@ -1,31 +1,74 @@
...
+  - checkout: self
+    fetchDepth: 10
+
+  - script: |
+      set -e
+      brew update
+      brew unlink python
+      brew install gnu-time opam
+
+      opam init -y -a -j 2 --compiler=$COMPILER
```
ejgallego: All the opam commands should use the OPAM 2 format.

azure-pipelines.yml
```diff
@@ -1,31 +1,74 @@
...
+    env:
+      COMPILER: "4.07.0"
+      FINDLIB_VER: ".1.8.0"
+
+  - script: |
+      set -e
+
+      eval $(opam config env)
+      ./configure -local -warn-error yes -native-compiler no
+      make -j 2
```
ejgallego: Number of cores should not be hardcoded but the value of `NJBOS` [or the corresponding Azure var if available] should be used.

azure-pipelines.yml
```diff
@@ -1,31 +1,74 @@
...
+    displayName: 'Build Coq'
+
+  - script: |
+      eval $(opam config env)
+      make -j 2 test-suite
+    displayName: 'Test Coq'
```
ejgallego: Maybe this should read more specific? "Coq Test Suite" ["Test" seems too generic to me]

azure-pipelines.yml
SkySkimmer: >All the opam commands should use the OPAM 2 format.

I don't know what that means.

ejgallego: `Dockerfile` contains OPAM 2 format commands.

ejgallego: add `--disable-sandboxing`

azure-pipelines.yml
```diff
@@ -1,31 +1,74 @@
...
+    fetchDepth: 10
+
+  - script: |
+      set -e
+      brew update
+      brew unlink python
+      brew install gnu-time opam
+
+      opam init -y -a -j 2 --compiler=$COMPILER
+      opam switch $COMPILER
```
ejgallego: Should be `opam switch set` [or `create`.]

azure-pipelines.yml
```diff
@@ -1,31 +1,74 @@
...
+      eval $(opam config env)
```
ejgallego: Should be `eval $(opam env)`

azure-pipelines.yml
```diff
@@ -1,31 +1,74 @@
...
+      opam update
+      opam install -j 2 -y num ocamlfind${FINDLIB_VER} ounit
```
ejgallego: Remove `-y` and export `OPAMYES: true` instead.

azure-pipelines.yml
```diff
@@ -1,31 +1,74 @@
...
+      opam list
+    displayName: 'Install dependencies'
+    env:
+      COMPILER: "4.07.0"
+      FINDLIB_VER: ".1.8.0"
+
+  - script: |
+      set -e
+
+      eval $(opam config env)
```
ejgallego: Ditto.

azure-pipelines.yml
```diff
@@ -1,31 +1,74 @@
...
+      ./configure -local -warn-error yes -native-compiler no
+      make -j 2
+    displayName: 'Build Coq'
+
+  - script: |
+      eval $(opam config env)
```
ejgallego: Ditto.

SkySkimmer:
@Zimmi48 https://gitlab.com/coq/coq/-/jobs/143593562 looks like a spurious timeout, should the bot have caught it? (I manually restarted)

azure-pipelines.yml
SkySkimmer: >add --disable-sandboxing

Why? It works without.

ejgallego: I understand that makes the build faster, also it reduces the complexity of the dependency chain, but we can leave it there if you want.

SkySkimmer:
Changed

Zimmi48:
>https://gitlab.com/coq/coq/-/jobs/143593562 looks like a spurious timeout, should the bot have caught it? (I manually restarted)

Hum, since when are timeouts spurious? BTW, it timed out again!
(I mean, I don't think the bot should restart this test in the same way that it doesn't restart failed tests due to non-deterministic tests. We should fix these instead.)

ejgallego (approved):
Looks good, thanks, and sorry for the delay.

PR #9335, opened 2019-01-10
[STM] kill no_safe_id anomaly

gares:
This improves the backtraces on errors raised while the DAG is being built by the STM


maximedenes (approved):
Yay!

PR #9341, opened 2019-01-14
[ssr] uniform clear discipline

gares:
Superseeds #9245


test-suite/ssr/ipat_replace.v
```diff
@@ -0,0 +1,17 @@
...
+have H : True.
+  by [].
+have {}H : True.
+  by apply: H.
+by apply: H.
+Qed.
+
+Lemma test2 (H : True) : False -> False.
+Proof.
+Fail move=> {}W.
```
CohenCyril: On the other hand, does `Fail move=> {H}W` fail? I think it should not.

Zimmi48:
Since this is a breaking change, this should get at least a specific mention in CHANGES.

gares:
> Since this is a breaking change, this should get at least a specific mention in CHANGES.

I'm updating the semantics to make it more backward compatible...

test-suite/ssr/ipat_replace.v
gares: solved

gares:
OK, it works and the *minuscule* overlays are backward compatible, so I'm waiting them to be merged.
I guess I should eventually improve  the doc and we could merge this one directly (it includes #9350).

gares:
@maximedenes I think you should be the assignee and reviewer, since it is mostly a code cleanup (given we break almost nothing the semantics seems OK to me).

CohenCyril:
@gares unless I missed something, I assume you will be extending the doc for intro patterns and rewrite, and also detail what happens when the clearing is not possible (failure or quiet).

gares:
I did add some doc for intro patterns, but I did not touch rewrite (I forgot). I'll do that right now.
The doc is now built by CI, you can have a look and propose improvements:
  https://coq.gitlab.io/-/coq/-/jobs/148578587/artifacts/_install_ci/share/doc/coq/sphinx/html/index.html

The semantic is that the extra item to be cleared is added only if it is a existing non-section variable used naked (eg not within extra `()`, in case of views or rewrite rules).

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -1583,18 +1586,25 @@ An :token:`s_item` can simplify the set of subgoals or the subgoals themselves:
...
+Views
+`````
+
 The first entry in the :token:`i_view` grammar rule, :n:`/@term`,
 represents a view (see section :ref:`views_and_reflection_ssr`).
 It interprets the top of the stack with the view :token:`term`.
-It is equivalent to ``move/term``. The optional flag ``{}`` can
-be used to signal that the :token:`term`, when it is a context entry,
-has to be cleared.
+It is equivalent to ``move/term``.
```
```suggestion
It is equivalent to :n:`move/@term`.
```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -1583,18 +1586,25 @@ An :token:`s_item` can simplify the set of subgoals or the subgoals themselves:
...
+
+A :token:`clear_switch` that immediately precedes an :token:`i_view`
+that happen to be a :token:`ident` is extended with that
+:token:`ident` and the :token:`clear_switch` is executed after the view.
+Eg ``{}/v`` is equivalent to ``/v{v}``.
```
```suggestion
E.g. ``{}/v`` is equivalent to ``/v{v}``.
```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -1608,13 +1618,20 @@ Notations can be used to name tactics,  for example::
...
+``````````````
+
 |SSR| supports the following :token:`i_pattern`\s:
 
 :token:`ident`
   pops the top variable, assumption, or local definition into
   a new constant, fact, or defined constant :token:`ident`, respectively.
   Note that defined constants cannot be introduced when δ-expansion is
   required to expose the top variable or assumption.
+  A :token:`clear_switch` (even an empty one) immediately preceding a
```
```suggestion
  A :token:`clear_switch` (even an empty one) immediately preceding an
```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3030,13 +3056,20 @@ operation should be performed:
...
++ This optional term, or the :token:`r_item`, may be preceded by an
+  :token:`occ_switch` (see section :ref:`selectors_ssr`) or a
+  :token:`clear_switch` (see section :ref:`discharge_ssr`),
+  these two possibilities being exclusive.
+
+  An occurrence switch selects
   the occurrences of the rewrite pattern which should be affected by the
   rewrite operation.
 
+  A clear switch, even an empty one, is is performed *after* the
```
```suggestion
  A clear switch, even an empty one, is performed *after* the
```

Zimmi48 (commented):
This is not a deep review. I'm tired and I'm making GitHub comments to procrastinate the preparation of my slides for my talk tomorrow.

gares:
Thanks for the suggestions, I've applied them all.

CohenCyril:
> The semantic is that the extra item to be cleared is added only if it is a existing non-section variable used naked (eg not within extra `()`, in case of views or rewrite rules).

The doc is not completely clear about that.
For introduced variables it just says, 
> if the identifier  points to an existing context entry

but does not say what happens for a view or rewrite that cannot be cleared.

Moreover it does not detail what exactly are the corner cases: either because it is a section variable or global constant, or because there is a remaining dependency. And it does not say whether the behavior is the same in these different cases.

gares:
I tried to clarify by systematically using "simple proof context entry" and the defining it in a footnote as an "naked ident (not between parentheses) designating a context entry not being a section variable".

maximedenes (approved):

maximedenes:
@CohenCyril let me know if the doc looks ok to you now. Thanks!

CohenCyril:
@maximedenes 
One ambiguity remaining for me, which I cannot understand from the doc or even guess: what are the behaviors of:
-  first,
```coq
Goal forall m n, m = 3.
Proof. move=> n {}n. Abort.
```
This should fail for one of these two reasons, but which one?
1. `n is used in conclusion`
2. `n already used` 

- And In the same vein:
```coq
Goal forall (m n : nat), n = 3.
Proof. move=> m {m}n. Abort.
```
Will there be a failure `n is used in conclusion` ? 
- In which case, it would be a very mysterious message for a beginner, so I would like a message explaining that `{m}` also tries to clear `n` (and refer the doc?) and even suggest to write `=> {m} - n` instead.
- Otherwise, I would be a bit worried about the stability of the proof script. (But maybe my fears are not grounded)

CohenCyril:
Ok one more (inspired by @strub).

```coq
Lemma test (x := 0) (y := 1) : y = 0.
Proof. rewrite {x}/y. Abort.
```
I assume this clears both `x` and `y`, but what if I want to clear only `x`. Is the only of doing this `rewrite /y {x}`, since there is no "noop" for rewrite (as `-` is for intro patterns)?

gares:
Note that the pr only documents the semantics in rewrite, it does not touch the code.

gares:
@CohenCyril what should I do?

CohenCyril:
> @CohenCyril what should I do?

let's discuss that this afternoon?

gares:
@CohenCyril you should be good to go

gares:
@maximedenes @CohenCyril approved "by voice"...

CohenCyril (approved):
I approve formally.

gares:
This can be merged now

anton-trunov:
What should I do if I want _warning-free_ compilation for both Coq 8.9 _and_ Coq 8.10alpha? I'd prefer not to enable `Set Warnings "-duplicate-clear".` globally, because sometimes those can be fixed, e.g. `move=> {}/n` is supported in both versions of Coq.

I can only think of two workarounds:

1. Disabling and re-enabling the warning locally:
```coq
Set Warnings "-duplicate-clear".
move=> {n} n.   (* [move=> {} n.] does not compile for Coq 8.9 *)
Set Warnings "duplicate-clear".
```

2. Using a no-op view + clearing:
```coq
move=> {n}/id n.
```

Is there a better way?

ggonthier:
2. is less obtrusive and thus preferable. However the proper way of inserting a "no-op" in an intro pattern sequence is with the `-` switch, e.g., `move=> {n}-n`. The script you propose in 2. should also be an error, because the `{}` should affect the next simple intro item that follows. It would an error in the warning generation message if it doesn't.
  However, I'm somewhat sceptical of value of the `duplicate-clear` warning, except as a porting tool, so it should perhaps be off by default. What the "duplicate" clear does is explicitly document that the new intro shadows (replaces) a previous context variable, and it's never an error to document code.

anton-trunov:
@ggonthier Thank you for your explanation, I filed an issue #9798

PR #9343, opened 2019-01-16
Try allowing minimization to change Prop <= i into i := Set.

mattam82:
We relax minimization such that Prop <= i constraints coming from Typing for a fresh "i" will ultimately instantiate i to Set, as is done for Set <= i. Alternative solution would be i := Prop which can't be done yet due to the restriction of polymorphism to not instantiate universe binders with Prop. I first need some feedback from CI to see if that's reasonably compatible.

**Kind:**  bug fix / feature

Fixes / closes #????

### DIFF
--- a/test-suite/success/cumulativity.v
+++ b/test-suite/success/cumulativity.v
@@ -123,11 +123,11 @@ Definition bar := foo.
 
 (* The universe on mkfoo is flexible and should be unified with i. *)
 Definition foo1@{i} : foo@{i} := let x := mkfoo in x. (* fast path for conversion *)
-Definition foo2@{i} : bar@{i} := let x := mkfoo in x. (* must reduce *)
+Definition foo2@{i} : bar@{} := let x := mkfoo in x. (* must reduce *)
 
 (* Rigid universes however should not be unified unnecessarily. *)
 Definition foo3@{i j|} : foo@{i} := let x := mkfoo@{j} in x.
-Definition foo4@{i j|} : bar@{i} := let x := mkfoo@{j} in x.
+Definition foo4@{i j|} : bar@{} := let x := mkfoo@{j} in x.
 
 (* Constructors for an inductive with indices *)
 Module WithIndex.
--- a/test-suite/success/polymorphism.v
+++ b/test-suite/success/polymorphism.v
@@ -462,3 +462,9 @@ Module ObligationRegression.
   Require Import CMorphisms.
   Check trans_co_eq_inv_arrow_morphism@{_ _ _ _ _  _ _ _}.
 End ObligationRegression.
+
+Axiom poly@{i} : forall(A : Type@{i}) (a : A), unit.
+Set Printing Universes.
+
+Definition nonpoly := @poly True Logic.I.
+Definition check := nonpoly@{}.
--- a/test-suite/success/polymorphism.v
+++ b/test-suite/success/polymorphism.v
@@ -460,7 +460,7 @@ Module ObligationRegression.
   (** Test for a regression encountered when fixing obligations for
       stronger restriction of universe context. *)
   Require Import CMorphisms.
-  Check trans_co_eq_inv_arrow_morphism@{_ _ _ _ _  _ _ _}.
+  Check trans_co_eq_inv_arrow_morphism@{_ _ _ _ _  _ _}.
 End ObligationRegression.
 
 Axiom poly@{i} : forall(A : Type@{i}) (a : A), unit.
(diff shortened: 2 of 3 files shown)


mattam82:
@janno that would fix your example.

SkySkimmer:
Test suite, equations and mtac fail (fiat-crypto is broken everywhere)

mattam82:
That’s expected, I’ll investigate later (probably not before the 25th though :)

Janno:
The other two failures (dummylang.v and DepDestruct.v) in Mtac2 are a bit harder to debug for me but I can at least say that the failures go away with `Unset Universe Minimization ToSet.`. It is possible that we are seeing the flip side of minimization here. It might be too eager in some cases, although it is not clear to me that this is really the case in these specific examples. (My previous experience with universe minimization was a few versions ago and back then it seemed to constantly force universes to `Set` that I would later need to be larger universes.)

Given the failure in tauto.v it seems at least equally likely that a clever annotation somewhere could fix this but that would probably require a deep dive into these particular files.

Janno:
Is this blocked because of the Mtac2 failures? I would consider investing some time into this if that's the case. I would really like to have this in Coq—given how the alternative is manual annotation (or Coq grinding to a screeching halt due to the sheer number of universes).

maximedenes:
I think this is closely related to #9444 and #9918, as the description of this PR suggests. I had planned to work on the two with @mattam82 as soon as he's back (next week I believe). So I hope to be able to say more about this then.

maximedenes:
@mattam82 How to make progress on this? Is it related to the possibility of instantiating polymorphic universes to `Prop`?

mattam82:
It is related (ultimately we could set `i := Prop` in some cases). Notably, if you have some universes which don't appear in the type but only in the body and just have `Prop <= i` then you get just as general typings by setting `i := Set` or `i := Prop`.

Janno:
I suppose these universes which don't appear in the type are the major source of universe-related overhead for Mtac2. Even functions with just a handful of universes in their types end up with hundreds or even thousands of universes, most of which are for `Type` quantifiers that end up being instantiated with a `Prop` somewhere in the body.

maximedenes:
I'll start with rebasing this.

test-suite/success/polymorphism.v
```diff
@@ -460,5 +460,11 @@ Module ObligationRegression.
   (** Test for a regression encountered when fixing obligations for
       stronger restriction of universe context. *)
   Require Import CMorphisms.
-  Check trans_co_eq_inv_arrow_morphism@{_ _ _ _ _  _ _ _}.
+  Check trans_co_eq_inv_arrow_morphism@{_ _ _ _ _  _ _}.
```
maximedenes: @SkySkimmer I don't know what you were testing here. Does the change make sense or does it mean there's a regression here?

test-suite/success/cumulativity.v
```diff
@@ -123,11 +123,11 @@ Definition bar := foo.
 
 (* The universe on mkfoo is flexible and should be unified with i. *)
 Definition foo1@{i} : foo@{i} := let x := mkfoo in x. (* fast path for conversion *)
-Definition foo2@{i} : bar@{i} := let x := mkfoo in x. (* must reduce *)
+Definition foo2@{i} : bar@{} := let x := mkfoo in x. (* must reduce *)
```
maximedenes: @mattam82 `bar` now has an empty universe instance. Is it expected? I guess it makes sense.

maximedenes:
I had a look at the Equations failure, but I don't understand it. I'll need help from universe experts (cc @mattam82 @SkySkimmer).

test-suite/success/polymorphism.v
SkySkimmer: As long as there isn't an increase in the instance length it should be fine.

test-suite/success/cumulativity.v
SkySkimmer: Here it's an irrelevant universe on a cumulative inductive so it doesn't matter (although it changes what we're testing), but in general doesn't it mean if you have an unconstrained universe it will become fixed (I guess to Prop?)? That doesn't seem great.
I guess it doesn't break HoTT so it might not be too bad.

SkySkimmer:
>I had a look at the Equations failure, but I don't understand it. I'll need help from universe experts (cc @mattam82 @SkySkimmer).

What do you not understand?

maximedenes:
> What do you not understand?

The failure is an unbound universe failure. Adding an explicit `@{i j}` instance at the failure point fixes it. Where does the unbound universe come from?

SkySkimmer:
I guess some universe used to be unified with i or j and now doesn't. I don't know equations enough to tell why.

maximedenes:
> I don't know equations enough to tell why.

In fact, the failure is in the theories which don't seem to use Equations.

maximedenes:
Update after progress with @mattam82: the equations failure is a bug: we should transform `Prop <= i` constraints only when `i` is flexible.

SkySkimmer:
Replaced by #10331

Issue #8006, opened 2018-07-06
[feature wish] Associate a hint database to remaining existential variables at Qed.


CohenCyril:
In library https://github.com/math-comp/analysis, when I use the set of tactics to manipulate filters, it creates an evar, which is instanciated bits by bits. In the very end, at Qed. One can assign a default value.
I hence have a `Grab Existential Variables. by end_near. Qed.` to close many of my proofs.
_(Side note: Sometimes it is `Unshelve. by end_near. Qed.`, although I do not understand the difference between remaining existential variables and shelved ones...)_

I would like that Coq tries some tactics/automation to solve these. Either by letting me add tactics/theorems to call at Qed. Time, it could be done through a `HintDb qed` which can be populated with `Hint`s and `Hint Extern`s.

Zimmi48:
AFAIU existential variables which are not unshelved (and are not goals), i.e. which can be reached by `Grab Existential Variables` but not by `Unshelve` are highlighting bugs. I would appreciate confirmation by an expert.

As for your use case, I'd like to learn more about it because it seems to me that this should always be preventable (when it's not a bug in a tactic). More precisely, when solving the last goal to refer to a given existential variable, it should be the right time to instantiate it...

CohenCyril:
> AFAIU existential variables which are not unshelved (and are not goals), i.e. which can be reached by Grab Existential Variables but not by Unshelve are highlighting bugs. I would appreciate confirmation by an expert.

Me too

> As for your use case, I'd like to learn more about it because it seems to me that this should always be preventable (when it's not a bug in a tactic). More precisely, when solving the last goal to refer to a given existential variable, it should be the right time to instantiate it...

@Zimmi48 I like that alternative and I considered it, but any tactic may solve "the last goal to refer to a given existential variable", so it would require a hook. I thought it was easier to just wait for `Qed`, rather than having a garbage collection of evar on the fly, even if the latter looks more elegant to me.

tchajed:
@Zimmi48 I've run into this before and assumed it was expected behavior. Specifically, in our case it's triggered by `eauto`'s search creating evars and solving the goal without instantiating them. I just reported #8047 with a minimized example.

Also, both `Grab Existential Variables` and `Unshelve` have always worked for me.

maximedenes:
> which can be reached by `Grab Existential Variables` but not by `Unshelve` are highlighting bugs. I would appreciate confirmation by an expert.

@ppedrot can you say whether this is accurate? Thanks!

[...]

PR #9347, opened 2019-01-17
At Qed, if shelved goals remain, emit a warning instead of an error

maximedenes:
This error was more or less a debug tool (checking that no tactic breaks
the invariant). But some users may want to support other models, see
Mtac2/Mtac2#139 for an example discussion.

### DIFF
--- a/proofs/proof.ml
+++ b/proofs/proof.ml
@@ -351,19 +351,13 @@ let dependent_start ~name ~poly goals =
 
 type open_error_reason =
   | UnfinishedProof
-  | HasShelvedGoals
   | HasGivenUpGoals
-  | HasUnresolvedEvar
 
 let print_open_error_reason er = let open Pp in match er with
   | UnfinishedProof ->
     str "Attempt to save an incomplete proof"
-  | HasShelvedGoals ->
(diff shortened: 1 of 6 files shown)


Zimmi48:
Related to #8006.

But I don't understand. If `Qed` succeeds now, how do you fill the proof-term?

proofs/proof.ml
```diff
@@ -351,15 +351,12 @@ let dependent_start ~name ~poly goals =
 
 type open_error_reason =
   | UnfinishedProof
-  | HasShelvedGoals
   | HasGivenUpGoals
   | HasUnresolvedEvar
```
ppedrot: You should also remove this constructor.

gares:
> how do you fill the proof-term

If the var occurs in the proof term, then you get another error.
The "error" turning into a warning was a "note for developers" saying "hey, you've allocated an evar without needing it".

Zimmi48:
Ah, ok. Thanks for the explanation!

maximedenes:
@ppedrot ping

ppedrot (approved):

ppedrot:
CI failure looks spurious, I'll merge.

PR #9350, opened 2019-01-17
[ssr] ipat replace

gares:
sub pull request of #9341 (the part shat should not break existing code)

For the reviewers:
The cleanup commit introduces a new data type `ipatops` interpreted by the ipat machine.
So now you have `ssripats -- tclCompileIPat --> ssriops -- tclIPAT --> unit tactic`.
The need is motivated by the fact that the "compilation" phase takes into account multiple items, eg a clear followed by a intro.

This makes sense on the input of the user (ssripat), while it is not always desirable on intro patters generated internally. Eg. prefixing an `ipats` with a `IPatClear` changes  meaning when the first `ipat` is an `IPatId` (it now means replace, eg one extra clear) while prefixing an `iops` with a `IOpClear` does not have this effect. Hence internally tactics patch `iops`, so that they can do it compositionally.

### DIFF
--- a/doc/sphinx/proof-engine/ssreflect-proof-language.rst
+++ b/doc/sphinx/proof-engine/ssreflect-proof-language.rst
@@ -1559,7 +1559,7 @@ whose general syntax is
    i_view ::= {? %{%} } /@term %| /ltac:( @tactic )
 
 .. prodn::
-   i_pattern ::= @ident %| > %| _ %| ? %| * %| + %| {? @occ_switch } -> %| {? @occ_switch }<- %| [ {?| @i_item } ] %| - %| [: {+ @ident } ]
+   i_pattern ::= {? %{%} } @ident %| > %| _ %| ? %| * %| + %| {? @occ_switch } -> %| {? @occ_switch }<- %| [ {?| @i_item } ] %| - %| [: {+ @ident } ]
 
 .. prodn::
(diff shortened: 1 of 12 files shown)


gares:
Finally ready for review. The cleanup makes it trivial to extend it to #9431 if the regressions are considered to be acceptable. This PR introduces no regression and enables the `=> {}H` idiom.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -1615,6 +1615,10 @@ annotation: views are interpreted opening the ``ssripat`` scope.
   a new constant, fact, or defined constant :token:`ident`, respectively.
   Note that defined constants cannot be introduced when δ-expansion is
   required to expose the top variable or assumption.
+  An empty occurrence switch ``{}`` has the effect of clearing
```
Zimmi48: For more clarity, what about changing the line above:

```
  :token:`ident`
```

into

```
  :n:`{? %{%} } @ident`
```

Zimmi48 (approved):

doc/sphinx/proof-engine/ssreflect-proof-language.rst
gares: Hum, let's see how #9341 goes, since it generalizes it to non-empty clears...

maximedenes:
Superseded by #9341.

PR #9356, opened 2019-01-18
Prepare support for -Q semantics for plugins

maximedenes:
We make the computation of the load path a bit more fine grained, to prepare support for loading some plugins using `-Q` (notably Ltac2 if we ship it with Coq).

### DIFF
--- a/config/coq_config.mli
+++ b/config/coq_config.mli
@@ -38,6 +38,7 @@ val compile_date : string (* compile date *)
 val vo_magic_number : int
 val state_magic_number : int
 
+val plugins_dirs : string list
 val all_src_dirs : string list
 
 val exec_extension : string (* "" under Unix, ".exe" under MS-windows *)
--- a/configure.ml
+++ b/configure.ml
@@ -1117,11 +1117,11 @@ let write_configml f =
   Array.iter
     (fun f ->
       let f' = "plugins/"^f in
-      if Sys.is_directory f' && f.[0] <> '.' then pr "  %S;\n" f')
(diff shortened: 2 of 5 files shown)


ppedrot (approved):

ejgallego (changes requested):
I don't like having a configure-time list of plugins; why is that needed?

Other than the bug fix [which looks correct] the patch doesn't seem to change much, maybe if you say more about your use case we could understand what this PR brings.

maximedenes:
> I don't like having a configure-time list of plugins; why is that needed?

I'm not sure I understand. This PR does not introduce any configure-time list of plugins. It simply changes the way we iterate over the content of the `plugins` folder.

The point of the PR is to make it possible to use `-Q` to bring plugins in scope, rather than `-R`, for some plugins. Doing it for all is what we want in the long term, but it will break some developments.

Do you want me to change something in this PR?

maximedenes:
You're right that the approach needs to be refined. I'll separate the fix so that we can integrate it meanwhile.

ejgallego:
I will take this following up discussion at CoqPL.

ejgallego:
@maximedenes I'd say this PR can be closed, right? [I mean, none of the code here will resemble the solution we discussed about]

IMHO a `vendor/` directory could be the way to go.

ejgallego:
Closing as to avoid PRs as issues; this will need some design indeed.

maximedenes:
> IMHO a `vendor/` directory could be the way to go.

Yep, will explore that. Closing.

Issue #9080, opened 2018-11-26
`coqc` and `coqdep` don't agree on ambiguous modules, and `coqc` seems to pick a random one.


ejgallego:
Files [modules under `theories`]:
- `M.v`:
```coq
From L Require Import A.
Print a.
```
- `M/A.v`:
```coq
Definition a := 0.
```
- `N/A.v`
```coq
Definition a := 1.
```
- `_CoqProject`:
```
-R theories L
```
make will say:
```
COQDEP VFILES
*** Warning: in file theories/M.v, 
    required library A matches several files in path
    (found A.v in theories/N and theories/M; used the latter)
*** Warning: in file theories/M.v, 
    required library A matches several files in path
    (found A.v in theories/N and theories/M; used the latter)
COQC theories/N/A.v
COQC theories/M.v
a = 1
     : nat
COQC theories/M/A.v
```
despite using choosing `M` both in `coqdep` and in the runtime. Now do:
```
$ touch theories/N/A.v
$ make
```
and we get:
```
COQDEP VFILES
*** Warning: in file theories/M.v, 
    required library A matches several files in path
    (found A.v in theories/N and theories/M; used the latter)
*** Warning: in file theories/M.v, 
    required library A matches several files in path
    (found A.v in theories/N and theories/M; used the latter)
COQC theories/N/A.v
COQC theories/M.v
a = 0
     : nat
```
the whole system seems so broken that I am not sure what to even propose here.

ejgallego:
Note that `coqc` emits no warning whatsoever.

herbelin:
Indeed, both `coqdep` and `coqc`/`coqtop` are using `readdir` but they don't browse the list in the same order (as `cat .coqdeps.d` shows). Indeed, one more place needing factorization (or at least, as a quick fix, to fix `coqdep`).

Blaisorblade:
> the whole system seems so broken that I am not sure what to even propose here.

Just my 2 cents: As a requirement, I'd prefer a warning/error over making the behavior consistent; I understand the refactoring might be more urgent (per https://github.com/coq/coq/issues/9839#issuecomment-477168307), but a quickfix might be justified since that refactoring is not easy.

Alizter:
On 8.16 coqdep throws a warning and coqc rejects the ambiguous path. I would therefore consider this closed.

PR #9357, opened 2019-01-19
Fix recursive loadpath of ML files

maximedenes:
ML files at the root were not taken into account. Coqdep was already
doing the right thing.

### DIFF
--- a/vernac/mltop.ml
+++ b/vernac/mltop.ml
@@ -215,7 +215,7 @@ let add_vo_path ~recursive lp =
     let () = match lp.has_ml with
       | AddNoML -> ()
       | AddTopML -> add_ml_dir unix_path
-      | AddRecML -> List.iter (fun (lp,_) -> add_ml_dir lp) dirs in
+      | AddRecML -> List.iter (fun (lp,_) -> add_ml_dir lp) dirs; add_ml_dir unix_path in
     let add (path, dir) =
       Loadpath.add_load_path path ~implicit dir in
     let () = List.iter add dirs in


ejgallego (approved):
LGTM

herbelin:
That `coqdep` and `coqtop` behaves the same is indeed expected. Whether the current directory should be added or not by default should however be discussed. Spontaneously, considering e.g. the problems we had with CoqIDE being adding the current directory by default in the loadpath, I would rather have adjusted `coqdep` to `coqtop`. After all, typing `-I .` is not costly and what does `coqtop` counts more as an established usage than `coqdep`. (Assuming I understood correctly the PR.)

Question: would it be feasible for you to ensure that whatever policy is adopted and whatever change is possibly done in the future, `coqdep` and `coqtop` will continue to work the same with respect to this feature, e.g. by putting the relevant code in the shared file `Minisys`?

maximedenes:
@herbelin sorry, my wording was imprecise. This PR is not about adding an implicit `-I`, it is about adding `dir` when `-I dir` is added (or rather, the internal equivalent of a recursive `-I`), instead of only its strict subfolders.

Now, there is still a pending question related to your comment: why do we need to implictly add things in the loadpath at all? Why couldn't all the necessary flags be added by the caller of `coqtop`/`coqdep`?

We can try to answer the question, but I think this patch is a good fix even if we decide to remove all this logic in the longer term.

ejgallego:
I guess I'll take this. I can merge like now, however maybe it would be useful to add a test case in the `coq-makefile` folder so we don't regress on the future?

You let me know @maximedenes

herbelin:
> This PR is not about adding an implicit -I, it is about adding dir when -I dir is added (or rather, the internal equivalent of a recursive -I), instead of only its strict subfolders.

OK. I misunderstood. Then this is clearly a bug fix.

> why do we need to implictly add things in the loadpath at all? Why couldn't all the necessary flags be added by the caller of coqtop/coqdep?

I'm missing data. What remains implicit in the design of path-related flags as of today?

Also, quickly comparing `coqdep` and `coqtop`, I see e.g. an `Add Rec ML Path` in `coqtop` which I don't see treated in `coqdep`. Is it that I looked to quickly or should we open an issue to remember that some work has still to be done there, even, if some progresses seem to have been done in the last years at clarifying the API (use of datatypes such as `coq_path_spec` or `add_ml`).

ejgallego:
Indeed @herbelin there are some discrepancies with Coq / coqdep ; see for example #9080 .

We are trying to refactor things little by little, indeed opening issues / PR on concrete actions would help a lot.

Personally I am holding a few refactoring until Dune is the main build system, as sharing code among Coq / coqdep gets a bit complicated with the current make system.

herbelin:
> We are trying to refactor things little by little, indeed opening issues / PR on concrete actions would help a lot.

Indeed, trying to list all variants of the corresponding flags (`-I`, `-Q`, `-R`), all variants of the corresponding coqtop-level commands (`Add ML Path`, `Add Rec ML Path`, `Add LoadPath`, `Add Rec LoadPath`), all variants of the corresponding implementations (in `coqtop`/`coqc`, in `coqdep`, in `coqdoc`) seems to show other discrepancies. Ensuring a bijection between flags, coq commands, implemented semantics, documented semantics is required.

ejgallego:
So indeed as I told you in person I do have a patch that implements a `coq.boot` library unifying the handling of all these stuff, however it does depend on Dune as the current make system uses a complex setup that makes this a bit hard to do.

maximedenes:
> I guess I'll take this. I can merge like now, however maybe it would be useful to add a test case in the `coq-makefile` folder so we don't regress on the future?

I think a test case is a bit overkill, especially because we'll have to migrate it from `coq_makefile` to Dune soon.

I'd rather be in favor of adding more exhaustive testing once we design the new infrastructure, rather than adding just one ad-hoc test for an unlikely regression (in this particular case).

PR #9366, opened 2019-01-21
Tweak names of implicit arguments to be closer to user specification

### LINKED ISSUE #8790: -mangle-names fails on inductive definition

jashug:
**Kind:** bug fix

Fixes / closes #8790

This is a minimal fix for #8790, less aggressive than either of #7254 or #9293.


SkySkimmer:
Why close? Too much incompatibility? Misclick?

Issue #7483, opened 2018-05-11
4811.v non deterministic


Zimmi48:
Here is a recent spurious failure due to [4811](https://github.com/coq/coq/blob/master/test-suite/bugs/closed/4811.v): https://gitlab.com/Zimmi48/coq/-/jobs/67687015

FTR, the log is the following:

```
==========> TESTING bugs/closed/4811.v <==========
While loading initial state:
Warning: Cannot open src [cannot-open-path,filesystem]
While loading initial state:
Warning: Cannot open Bedrock [cannot-open-path,filesystem]
While loading initial state:
Warning: Cannot open coqprime-8.5/Coqprime [cannot-open-path,filesystem]
Finished failing transaction in 1. secs (0.7u,0.016s) (failure)
File "/builds/Zimmi48/coq/test-suite/bugs/closed/4811.v", line 1685, characters 2-25:
Error: Timeout!

0m0.000000s 0m0.000000s
0m2.010000s 0m0.490000s
==========> FAILURE <==========
    bugs/closed/4811.v...Error! (bug seems to be opened, please check)
```

Once again, the source of non-determinism is the use of the `Timeout` command. In light of #7408, #7430, shouldn't we remove all the tests that rely on this command?

ejgallego:
> In light of #7408, #7430, shouldn't we remove all the tests that rely on this command?

I think we should instead fix `Timeout`, I do think it is possible but I'll talk about that in the proper issue.

The good news is that indeed I think we have finally understood the reason why these bad failures happened.

Zimmi48:
I had a new occurrence: https://gitlab.com/Zimmi48/coq/-/jobs/73460133
Three already reported non-deterministic tests failed because of `Timeout` (see also #7388, #5982).

Zimmi48:
New occurrence: https://gitlab.com/coq/coq/-/jobs/93028883

Zimmi48:
And another one here: https://gitlab.com/coq/coq/-/jobs/93040622
@JasonGross could it be the sign of anything else than bad luck? (Note that the two failures I noted are from the v8.8 branch.)

JasonGross:
[...]

Issue #7388, opened 2018-04-29
3441.v non deterministic


SkySkimmer:
https://gitlab.com/coq/coq/-/jobs/65545311

Zimmi48:
I had a new occurrence: https://gitlab.com/Zimmi48/coq/-/jobs/73460133
Three already reported non-deterministic tests failed because of `Timeout` (see also #7383, #5982).

Zimmi48:
https://gitlab.com/coq/coq/-/jobs/136470246

Related to the discussion on timeout at: #7483

Issue #5982, opened 2017-10-19
bugs/closed/4366.v is non-deterministic


Zimmi48:
[`bugs/closed/4366.v`](https://github.com/coq/coq/tree/master/test-suite/bugs/closed/4366.v) failed on a merge commit [on Travis](https://travis-ci.org/coq/coq/jobs/233112145#L5193) but didn't afterwards.

#### Failure log (edit to complete)

- On Travis on 2017-05-17.

Zimmi48:
I had a new occurrence: https://gitlab.com/Zimmi48/coq/-/jobs/73460133
Three already reported non-deterministic tests failed because of Timeout (see also #7383, #7388).

Zimmi48:
New occurrence: https://gitlab.com/coq/coq/-/jobs/85934360

PR #9372, opened 2019-01-21
[thread] protect threads against sigalrm

gares:
Fix #7408 (when compiled with 4.08 if I got it right)

### DIFF
--- a/clib/cThread.ml
+++ b/clib/cThread.ml
@@ -97,6 +97,9 @@ let thread_friendly_input_value ic =
     end
   with Unix.Unix_error _ | Sys_error _ -> raise End_of_file
 
+(* On the ocaml runtime used in some opam-for-windows version the
+ * [Thread.sigmask] API raises Invalid_argument "not implemented",
+ * hence we protect the call and turn the exception into a no-op *)
 let protect_sigalrm f x =
   begin try ignore(Thread.sigmask Unix.SIG_BLOCK [Sys.sigalrm])
   with Invalid_argument _ -> () end;
--- a/clib/cThread.ml
+++ b/clib/cThread.ml
@@ -97,3 +97,10 @@ let thread_friendly_input_value ic =
     end
   with Unix.Unix_error _ | Sys_error _ -> raise End_of_file
 
+let protect_sigalrm f x =
+  begin try ignore(Thread.sigmask Unix.SIG_BLOCK [Sys.sigalrm])
+  with Invalid_argument _ -> () end;
+  f x
+
+let create f x =
+  Thread.create (protect_sigalrm f) x
--- a/clib/cThread.mli
+++ b/clib/cThread.mli
@@ -26,3 +26,6 @@ val thread_friendly_really_read :
   thread_ic -> Bytes.t -> off:int -> len:int -> unit
 val thread_friendly_really_read_line : thread_ic -> string
 
+(* Wrapper around Thread.create that blocks signals such as Sys.sigalrm (used
+ * for Timeout *)
+val create : ('a -> 'b) -> 'a -> Thread.t
--- a/configure.ml
+++ b/configure.ml
@@ -1001,6 +1001,7 @@ let print_summary () =
   pr "  Architecture                : %s\n" arch;
   if operating_system <> "" then
     pr "  Operating system            : %s\n" operating_system;
+  pr "  Sys.os_type                 : %s\n" Sys.os_type;
   pr "  Coq VM bytecode link flags  : %s\n" (String.concat " " vmbyteflags);
   pr "  Other bytecode link flags   : %s\n" custom_flag;
   pr "  OCaml version               : %s\n" caml_version;
--- a/lib/control.ml
+++ b/lib/control.ml
@@ -57,7 +57,7 @@ let windows_timeout n f x e =
     done
   in
   let init = Unix.gettimeofday () in
-  let _id = Thread.create thread init in
+  let _id = CThread.create thread init in
   try
     let res = f x in
     let () = killed := true in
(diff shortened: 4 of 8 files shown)


gares:
My hope is that this PR makes the `success/auto.v` file work reliably with `-async-proofs on` (as of today there is a Timeout generating a signal that is sometimes catch by the wrong thread

ejgallego (commented):
A few comments:
- I am not sure I like the `create` name; we should either name it `create_ignore_sigalarm` as it would be very confusing otherwise, or just do the ignore on the right places.
- How can we test this is working? Should you add a unit test?
- Hardcoding the signal name seems fragile, should timeout export a `timeout_signal` identifier?
- I am a bit confused how this is supposed to work. If you ignore a signal, it won't be redelivered to the right thread, but lost. So how is this supposed to work?

gares:
>     * I am a bit confused how this is supposed to work. If you ignore a signal, it won't be redelivered to the right thread, but lost. So how is this supposed to work?

I don't mask the signal in the main thread, that is the one doing the computations (both in master and in the workers).

gares:
> How can we test this is working? Should you add a unit test?

I really don't know how to reproduce it effectively, I was planning of running (by hand) the appropriate CI job a few times.

A unit test would just test what they do in caml, that is that `setsigmask` actually masks.
In the context of Coq I don't know... I'd just grep for `Thread.create` in order to be sure the entire codebase uses `CThread.create`.

gares:
> Hardcoding the signal name seems fragile, should timeout export a `timeout_signal` identifier?

I'm not sure I follow, the `alarm` system call is not changing the signal it raises, ever (it is a standard).

gares:
run 1 of `test-suite:base+async` passed, running again

gares:
run 3 of `test-suite:base+async` passed, running again

ejgallego:
Ok, thanks for the responses, I am indeed OK with the current status of the patch then, an in particular with using `create`.

However I am still confused about this:

> I don't mask the signal in the main thread, that is the one doing the computations (both in master and in the workers).

So does POSIX guarantee that the signal will be delivered to the main thread? Where is this particular case specified? I am wary because of the comment of Xavier a few weeks ago about the futility of implementing `Timeout` like this.

gares:
> So does POSIX guarantee that the signal will be delivered to the main thread

IIRC signal should be delivered to all threads (the main one is not special).

In any case, there are plenty of tests using Timeout, so I'm pretty sure it is delivered.
What I can't test is that we actually mask it in all threads (but the main one). I run it 5 times...

ejgallego:
> IIRC signal should be delivered to all threads (the main one is not special).

It is indeed delivered not to all threads but to an arbitrary one. After reading the documentation I do think indeed blocking should be enough as the wording says "the signal will be delivered to an arbitrary thread which didn't block the signal".

I am OK with the patch then, maybe you want to tweak the commit message a bit to use "main thread" instead of "computing thread" which confused me a bit.

CI is failing on Windows tho, what's going on there?

ejgallego:
> CI is failing on Windows tho, what's going on there?

So indeed it seems that this PR breaks Timeout on Windows.

gares:
>  I guess `Sys.os_type` returns "Cygwin"

No, it returns "Windows" hence we run a code path that should not be affected by this patch.
Indeed Windows jobs on appveyor and gitlab worked, so there must be something special about the azure job... 

Anything comes to mind? @coq/ci-maintainers

ejgallego:
> Anything comes to mind? @coq/ci-maintainers

It is not very likely but it could well be a different kernel version making a different threading choice?

gares:
I found it: `Thread 2 killed on uncaught exception Invalid_argument("Thread.sigmask not implemented")`

So I guess I should just make the syscall safe

gares:
Something weird is happening since the azure job is not running... ideas? 

A side note for @coq/ci-maintainers : the issue is that the ocaml runtime running on azure is not the same runtime running on gitlab since it has nonstandard external C functions: sigmask shoud not raise `Invalid_argument "not implemented"`, but it does.

SkySkimmer:
>Something weird is happening since the azure job is not running... ideas?

Indeed I don't know

gares:
I try a close + open to give azure a second chance

jhjourdan:
I am currently working on the OCaml patch to make Thread.sigmask. We had to revert the original patch, which created some regressions. It turns out it is more difficult than initially anticipated, and I now have a bug to debug on MacOS X.

See https://github..com/ocaml/ocaml/pull/2211, and stay tuned.

jhjourdan:
(That's said, this is no reason to suspend the integration to this patch in Coq, since the function already exists in previous OCaml versions.)

ejgallego:
> A side note for @coq/ci-maintainers : the issue is that the ocaml runtime running on azure is not the same runtime running on gitlab since it has nonstandard external C functions: sigmask shoud not raise `Invalid_argument "not implemented"`, but it does.

Can you tell more about what the above means? How they can differ, they are the same than appveyor for example?

Any link to an OCaml issue?

gares:
Ci is green now, also on azure. Can someone merge this?

gares:
The prebuilt opam for Windows thing has a heavily patched runtime. Iirc one can find the diff online, and be scared...

ejgallego:
> The prebuilt opam for Windows thing has a heavily patched runtime. Iirc one can find the diff online, and be scared...

Yes but what I mean is that in Appveyor we use the same exact runtime and still didn't see a difference?

ejgallego (approved):
The `try ... with ...` block on `cThread` likely deserves a comment, other than that the patch seems good.

gares:
comment added

Zimmi48:
@gares Do you expect this to solve #7483, #7388, and #5982?

PR #9373, opened 2019-01-21
Kernel: don't automatically downgrade ill-shaped primitive records

SkySkimmer:
This simplifies reasoning about the kernel code.
We still auto downgrade squashed Prop records as the code path to
avoid an error is more involved. Alternatively we could produce an
error forcing people to Unset Primitive Projections if they want a
squashed record.


SkySkimmer:
ping @maximedenes

ppedrot:
I can handle that one, if needed. The code looks OK to me.

ppedrot (approved):
Official seal of approval.

PR #9377, opened 2019-01-22
[CS] recognize applied primitive projections as keys (fix #9375)

### LINKED ISSUE #9375: Canonical structure inference breaks with applied primitive projections
#### Version

8.8.2

#### Operating system

Linux

In the following snippet, Coq fails to identify the primitive projection in `tata_toto` as a head constant. The second declaration works fine.

```
Set Primitive Projections.

Record toto : Type := Toto {
  toto1 : Type;
  toto2 : toto1 -> Type
}.

Record tata := Tata {
  tata1 : Type
}.

Canonical Structure tata_toto (x : toto) X :=
  Tata (toto2 x X).

(* Warning: Projection value has no head constant: toto2 x X *)

Canonical Structure tata_tata (x : tata) := Tata (tata1 x).

(* No problem here. *)
```

gares:
A branch of code was simply missing, resulting in a CS not being declared.

### DIFF
--- a/pretyping/recordops.ml
+++ b/pretyping/recordops.ml
@@ -202,7 +202,14 @@ let cs_pattern_of_constr env t =
       App (f,vargs) ->
 	begin
 	  try Const_cs (global_of_constr f) , None, Array.to_list vargs
-          with e when CErrors.noncritical e -> raise Not_found
+          with
+          | Not_found when isProj f ->
+              let p, c = destProj f in
+              let { Environ.uj_type = ty } = Typeops.infer env c in
+              let _, params = Inductive.find_rectype env ty in
+              Const_cs (ConstRef (Projection.constant p)), None,
+              params @ [c] @ Array.to_list vargs
+          | e when CErrors.noncritical e -> raise Not_found
 	end
     | Rel n -> Default_cs, Some n, []
     | Prod (_,a,b) when Vars.noccurn 1 b -> Prod_cs, None, [a; Vars.lift (-1) b]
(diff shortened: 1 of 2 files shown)


pretyping/recordops.ml
```diff
@@ -202,7 +202,14 @@ let cs_pattern_of_constr env t =
       App (f,vargs) ->
 	begin
 	  try Const_cs (global_of_constr f) , None, Array.to_list vargs
-          with e when CErrors.noncritical e -> raise Not_found
+          with
+          | Not_found when isProj f ->
+              let p, c = destProj f in
+              let { Environ.uj_type = ty } = Typeops.infer env c in
```
SkySkimmer: Why not retyping here?

gares: I don't know I've just adapted the code for `Proj` from below

SkySkimmer: Oh I see.
Maybe it would be more uniform to make the function recursive, then the App case would just add the arguments to the list returned by a recursive call on the head. This would change behaviour on applied Rel (from Not_found to returning something), I don't know what the consequence of that would be.

mattam82: It looks to me like Retyping would be enough, t is supposed to be typable here I believe.

gares:
If this code is correct, but not optimal in efficiency, I'd rather merge this and leave the optimization for another PR.

gares:
@mattam82 can you take care of this?

mattam82 (approved):
I'm ok with the fix, and indeed moving to retyping can be done later.

mattam82:
Will merge soon if nobody opposes.

pretyping/recordops.ml
ppedrot: I think this is because this is supposed to work on Constr when retyping works on EConstr.

gares:
@ppedrot @SkySkimmer are you OK with the current patch being merged?

SkySkimmer:
OK

PR #9380, opened 2019-01-22
Remove monomorphic/polymorphic distinction in the kernel

SkySkimmer:
We still have to track polymorphism outside the kernel for the following:
~~~coq
Polymorphic Inductive foo : Set := bar.
~~~
The universe of the predicate in `foo_rect` should be polymorphic. In current master this is done by imitating the polymorphism of the inductive. If we don't track polymorphism then there is no trace of having used `Polymorphic` (foo has no universes).
We could try passing the polymorphism attribute explicitly to scheme functions but that gets pretty heavy, see unfinished commit 24a59c4a262ab770e85e606dc5c8a3df147c143f
We could revert having to access the polymorphism option through an attribute, but that's kinda messy.
We could try just making all _rect universes polymorphic. If we just pass poly:true in schemes we get an assertion failure from https://github.com/SkySkimmer/coq/blob/bd96a3ff6e906f828530f3aa5f68a838dee533ec/kernel/term_typing.ml#L52 so I'm not sure how feasible it would be.

Some other places are calling is_polymorphic, they didn't have an issue with the implementation `= not (is_empty polymorphic_univs)` but I haven't reviewed them.

I put XXX/TODO in various places which need a look, I'll point them out later through the github UI.

Backward compat seems fine (aside from the usual plugin breaking).

This probably introduces bugs when the ability to combine monomorphic and polymorphic universes is used, especially in sections. Accordingly I'm not sure if we want to do this.

### DIFF
--- a/interp/declare.ml
+++ b/interp/declare.ml
@@ -16,6 +16,7 @@ open Util
 open Names
 open Libnames
 open Globnames
+open Univ
 open Constr
 open Declarations
 open Entries
@@ -84,7 +85,7 @@ let cache_constant ((sp,kn), obj) =
   assert (Constant.equal kn' (Constant.make1 kn));
   Nametab.push (Nametab.Until 1) sp (ConstRef (Constant.make1 kn));
   let cst = Global.lookup_constant kn' in
-  add_section_constant (Declareops.constant_is_polymorphic cst) kn' cst.const_hyps;
+  add_section_constant kn' cst.const_hyps;
   add_constant_kind (Constant.make1 kn) obj.cst_kind
 
 let discharge_constant ((sp, kn), obj) =
@@ -143,10 +144,14 @@ let declare_constant_common id cst =
   update_tables c;
   c
 
-let default_univ_entry = Monomorphic_const_entry Univ.ContextSet.empty
+let empty_univ_entry = {
+  entry_monomorphic_univs = ContextSet.empty;
+  entry_poly_univ_names = [| |];
+  entry_polymorphic_univs = UContext.empty;
+}
 let definition_entry ?fix_exn ?(opaque=false) ?(inline=false) ?types
-    ?(univs=default_univ_entry) ?(eff=Safe_typing.empty_private_constants) body =
-  { const_entry_body = Future.from_val ?fix_exn ((body,Univ.ContextSet.empty), eff);
+    ?(univs=empty_univ_entry) ?(eff=Safe_typing.empty_private_constants) body =
+  { const_entry_body = Future.from_val ?fix_exn (Safe_typing.make_proof body, eff);
     const_entry_secctx = None;
     const_entry_type = types;
     const_entry_universes = univs;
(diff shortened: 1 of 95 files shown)


ppedrot (changes requested):
I really find this PR is going the wrong way. I don't have much time to comment, but I am leaving this change request here so that it doesn't get merged without a *serious* discussion.

maximedenes:
Maybe a good topic the WG, then? It looks very technical, so some live explanations would be welcome :)

interp/declare.ml
```diff
@@ -205,46 +215,54 @@ let cache_variable ((sp,_),o) =
...
+         OK. *)
+      let univs = de.const_entry_universes in
+      let mono = ContextSet.union univs.entry_monomorphic_univs proof_univs in
+      let poly = ContextSet.union (ContextSet.of_context univs.entry_polymorphic_univs)
+          proof_priv_univs
       in
-      let univs = Univ.ContextSet.union uctx univs in
-      (* We must declare the universe constraints before type-checking the
-         term. *)
-      let () = Global.push_context_set (not poly) univs in
```
SkySkimmer: Why is this inverted from the usual?

kernel/mod_typing.ml
```diff
@@ -74,55 +74,46 @@ let rec check_with_def env struc (idl,(c,ctx)) mp equiv =
...
-          let c',cst = match cb.const_body with
-            | Undef _ | OpaqueDef _ ->
-              let j = Typeops.infer env' c in
-              let typ = cb.const_type in
-              let cst' = Reduction.infer_conv_leq env' (Environ.universes env')
-                  j.uj_type typ in
-              j.uj_val, cst'
-            | Def cs ->
-              let c' = Mod_subst.force_constr cs in
-              c, Reduction.infer_conv env' (Environ.universes env') c c'
```
SkySkimmer: Seems like this breaks the assumption that we only convert terms which have a common type.

vernac/obligations.ml
```diff
@@ -655,11 +653,11 @@ let declare_obligation prg obl body ty uctx =
       in
       if not opaque then add_hint (Locality.make_section_locality None) prg constant;
       definition_message obl.obl_name;
-      let body = match uctx with
-        | Polymorphic_const_entry (_, uctx) ->
-          Some (DefinedObl (constant, Univ.UContext.instance uctx))
-        | Monomorphic_const_entry _ ->
+      (* XXX why do we choose what to do based on polymorphism here? *)
+      let body = if Univ.UContext.is_empty uctx.entry_polymorphic_univs then
           Some (TermObl (it_mkLambda_or_LetIn_or_clean (mkApp (mkConst constant, args)) ctx))
```
SkySkimmer: Why do we branch on polymorphism here? @mattam82

SkySkimmer:
We probably want to work on detaching monomorphic universes from constants before doing this. I pointed out the XXX which may still have some interest aside from this PR.

JasonGross:
> We could try just making all _rect universes polymorphic. If we just pass poly:true in schemes we get an assertion failure from 

This interacts badly with the fact that it's currently forbidden to mix monomorphic and polymorphic variables in sections.

PR #9386, opened 2019-01-23
Pass some files to strict focusing mode.

SkySkimmer:
ie default goal selector `!`

Results look pretty OK IMO, do we want to go this way? We can't actually enforce it until we have per-project options or we change the default default goal selector (repetition intended ^^).

How to do this:
- change the default value of default goal selector in goal_select.ml
- eval the emacs code in this commit message
- compile Coq and in each erroring file repeatedly run
  [C-c f] (my/maybe-fix-buller-error) then [C-c C-b] (proof-process-buffer)
  until there are no errors (NB the first [C-c f] has no effect).

You need to watch for:
- overly deep proofs where the bullets need to go beyond the list in
  my/bullet-stack (6 layers is enough the vast majority of the time
  though). The system will give you an error and you need to finish
  the lemma manually.

~~~emacs-lisp
(defconst my/bullet-stack (list "-" "+" "*" "--" "++" "**")
  "Which bullets should be used, in order.")

(defvar-local my/bullet-count nil
  "The value in the car indicates how many goals remain in the
  bullet at (length-1), and so on recursively. nil means we
  haven't started bulleting the current proof.")

(defvar-local my/last-seen-qed nil)

(defun my/get-maybe-bullet-error ()
  "Extract the number of focused goals from the ! selector error message."
  (when-let* ((rbuf (get-buffer "*response*"))
              (str (with-current-buffer "*response*" (buffer-string)))
              (_ (string-match
                  (rx "Error: Expected a single focused goal but " (group (+ digit)))
                  str))
              (ngoals (string-to-number (match-string 1 str))))
    ngoals))

(defun my/bullet-fix-indent ()
  "Auto indent until the next Qed/Defined, and update my/last-seen-qed."
  ;;  (insert (format "(* %s -> %s *)\n" my/prev-count my/bullet-count))
  (when-let ((qed (save-excursion (search-forward-regexp (rx (or "Defined." "Qed.")) nil t))))
    (set-marker my/last-seen-qed qed)
    (indent-region (- (point) 1) qed)))

(defun my/nth-bullet (n)
  "Get nth bullet, erroring if n >= length my/bullet-stack"
  (or (nth n my/bullet-stack)
      (error "Too many bullets.")))

(defun my/is-at-indent ()
  (save-excursion
                (let ((cur-point (point)))
                  (back-to-indentation)
                  (= (point) cur-point))))

(defun my/maybe-fix-bullet-error (&optional arg)
  "Main function for porting a file to strict focusing.

Repeatedly process your file in proof general until you get a
focusing error, then run this function. Once there are no more
errors you're done.

Indentation commonly looks bad in the middle of fixing a proof,
but will be fixed.

This is a stateful process. The state is automatically reset when
you get to the next proof, but if you get an error or take manual
action which breaks the algorithm's expectation you can call with
prefix argument to reset."
  (interactive "P")
  (unless my/last-seen-qed
    (setq my/last-seen-qed (set-marker (make-marker) 0)))
  (when (or arg (> (point) my/last-seen-qed))
    (setq my/bullet-count nil)
    (set-marker my/last-seen-qed 0))

  (when-let ((ngoals (my/get-maybe-bullet-error)))
    (setq my/prev-count (format "%s %s" ngoals my/bullet-count))
    (unless (my/is-at-indent) (newline nil t))
    (if (= ngoals 0)
        (progn
          (while (and my/bullet-count (= (car my/bullet-count) 0))
            (pop my/bullet-count))

          (insert (concat (my/nth-bullet (- (length my/bullet-count) 1)) " "))
          (setq my/bullet-count (cons (- (car my/bullet-count) 1) (cdr my/bullet-count)))
          (my/bullet-fix-indent))

      (setq my/bullet-count (cons (- ngoals 1) my/bullet-count))
      (insert (concat (my/nth-bullet (- (length my/bullet-count) 1)) " "))
      (my/bullet-fix-indent))))

(bind-key "C-c f" #'my/maybe-fix-bullet-error coq-mode-map)

(defun my/maybe-fix-brace-error ()
  "Port to strict focusing using curly braces."
  (interactive)
  (when-let ((ngoals (my/get-maybe-bullet-error)))
    (if (= ngoals 0)
        (save-excursion
          (re-search-backward "[^[:space:]]" nil t)
          (forward-char 1)
          (insert " }"))
      (if (= ngoals 2) (insert "{ ") (error "Too many goals for braces.")))))

(bind-key "C-c g" #'my/maybe-fix-brace-error coq-mode-map)
~~~

### DIFF
--- a/theories/Bool/BoolEq.v
+++ b/theories/Bool/BoolEq.v
@@ -67,8 +67,8 @@ Section Bool_eq_dec.
   Proof.
     intros x y; case (exists_beq_eq x y).
     intros b; case b; intro H.
-      left; apply beq_eq; assumption.
-      right; apply beq_false_not_eq; assumption.
+    - left; apply beq_eq; assumption.
+    - right; apply beq_false_not_eq; assumption.
   Defined.
 
 End Bool_eq_dec.
--- a/theories/Bool/IfProp.v
+++ b/theories/Bool/IfProp.v
@@ -45,6 +45,6 @@ Qed.
 
 Lemma IfProp_sum : forall (A B:Prop) (b:bool), IfProp A B b -> {A} + {B}.
 destruct b; intro H.
-left; inversion H; auto with bool.
-right; inversion H; auto with bool.
+- left; inversion H; auto with bool.
+- right; inversion H; auto with bool.
 Qed.
--- a/theories/Classes/CMorphisms.v
+++ b/theories/Classes/CMorphisms.v
@@ -164,7 +164,11 @@ Section Relations.
 
   Lemma pointwise_pointwise {B} (R : crelation B) :
     relation_equivalence (pointwise_relation R) (@eq A ==> R).
-  Proof. intros. split. simpl_crelation. firstorder. Qed.
+  Proof.
+    intros. split.
+    - simpl_crelation.
+    - firstorder.
+  Qed.
   
   (** Subcrelations induce a morphism on the identity. *)
   
@@ -265,8 +269,8 @@ Section GenericInstances.
   Next Obligation.
   Proof with auto.
     assert(R x0 x0).
-    transitivity y0... symmetry...
-    transitivity (y x0)...
+    - transitivity y0... symmetry...
+    - transitivity (y x0)...
   Qed.
 
   Unset Strict Universe Declaration.
@@ -339,10 +343,11 @@ Section GenericInstances.
 
   Next Obligation.
   Proof with auto.
-    split. intros ; transitivity x0...
-    intros.
-    transitivity y...
-    symmetry...
+    split.
+    - intros ; transitivity x0...
+    - intros.
+      transitivity y...
+      symmetry...
   Qed.
 
   (** Every Transitive crelation induces a morphism by "pushing" an [R x y] on the left of an [R x z] proof to get an [R y z] goal. *)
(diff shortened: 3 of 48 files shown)


ejgallego:
Note that significant user developments [such as math-comp] are not going to adopt this "strict mode", so indeed I wonder what the default should be.

Zimmi48:
I'm favorable to merging this PR, not changing the default, and work on per-project options.
math-comp can start to use this mode as well if we provide them with significant indentation. I had said I would do this. Obviously I won't for a while. I should open an issue to at least summarize my ideas about it.

Zimmi48:
I'm not sure if you were trying to convince me but I'm well aware of these problems. That's why I said "if we provide them with significant indentation". What I meant by that is precisely a checker for the math-comp indentation, which indeed is just as meaningful as the use of bullets.

gares:
Oh, I thought you wanted to re-indent he scripts. Well, I'd love to see a "layout" checker for Coq.

Zimmi48:
Yeah, I've wanted to do that for the last one and a half year. Unfortunately, other things have taken the priority. That's why I should at least write down my thoughts in a new issue.

Zimmi48 (approved):
One of the stdlib maintainer should merge this IMO.

herbelin (approved):
A long-awaited wish!

ppedrot (approved):
That's cool indeed. Who's assigning?

PR #9389, opened 2019-01-23
Implement a method for manual declaration of implicits.

jashug:
This is intended to be separate from handling of implicit binders.
The remaining uses of `declare_manual_implicits` satisfy a lot of assertions, giving the possibility of simplifying the interface in the future.

Two disabled warnings are added for things that currently pass silently.

**Kind:** redesign

Should preserve behavior, except for rewording of some error messages.

Overlays https://github.com/mattam82/Coq-Equations/pull/181 https://github.com/Mtac2/Mtac2/pull/172

### DIFF
--- a/interp/impargs.ml
+++ b/interp/impargs.ml
@@ -674,7 +674,7 @@ let check_inclusion l =
 	  user_err Pp.(str "Sequences of implicit arguments must be of different lengths.");
 	aux nl
     | _ -> () in
-  aux (List.map (fun (imps,_) -> List.length imps) l)
+  aux (List.map snd l)
 
 let check_rigidity isrigid =
   if not isrigid then
@@ -685,6 +685,8 @@ let projection_implicits env p impls =
   CList.skipn_at_least npars impls
 
 let declare_manual_implicits local ref ?enriching l =
+  assert (List.for_all (fun (_, (max, fi, fu)) -> fi && fu) l);
+  assert (List.for_all (fun (ex, _) -> match ex with ExplByPos (_,_) -> true | _ -> false) l);
   let flags = !implicit_args in
   let env = Global.env () in
   let sigma = Evd.from_env env in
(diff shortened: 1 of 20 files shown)


interp/impargs.mli
```diff
@@ -119,6 +119,12 @@ val declare_manual_implicits : bool -> GlobRef.t -> ?enriching:bool ->
 val maybe_declare_manual_implicits : bool -> GlobRef.t -> ?enriching:bool ->
   manual_implicits -> unit
 
+(** [set_implicits local ref l]
+   Manual declaration of implicit arguments, for use by the Arguments command.
+   takes a list of possible sequences of implicit statuses, where
+   None = not implicit, Some false = non-maximal implicit, Some true = maximal implicit. *)
+val set_implicits : bool -> GlobRef.t -> bool option list list -> unit
```
SkySkimmer: Why do we pass a bool option here instead of a Vernacexpr.vernac_implicit_status?

jashug: That would probably be better, wouldn't it.
I think I was worried about layering issues, and also there was a point in drafting this PR when the type was `(Name.t * bool) option` and I didn't rethink the type when I removed the name.

jashug: There was actually a layering issue, so I pulled the definition of `vernac_implicit_status` down into `Impargs` and left an alias in `Vernacexpr`.

interp/impargs.mli
```diff
@@ -119,6 +119,14 @@ val declare_manual_implicits : bool -> GlobRef.t -> ?enriching:bool ->
 val maybe_declare_manual_implicits : bool -> GlobRef.t -> ?enriching:bool ->
   manual_implicits -> unit
 
+type implicit_kind = Implicit | MaximallyImplicit | NotImplicit
+
+(** [set_implicits local ref l]
+   Manual declaration of implicit arguments, for use by the Arguments command.
+   takes a list of possible sequences of implicit statuses, where
+   None = not implicit, Some false = non-maximal implicit, Some true = maximal implicit. *)
```
gares: Isn't this outdated? (you have `implicit_kind` now)

jashug:
Added a note in `dev/doc/changes.md`, and removed an outdated comment from `Vernacentries`.

vernac/vernacexpr.ml
```diff
@@ -237,7 +237,7 @@ type vernac_cumulative = VernacCumulative | VernacNonCumulative
 
 (** {6 The type of vernacular expressions} *)
 
-type vernac_implicit_status = Implicit | MaximallyImplicit | NotImplicit
+type vernac_implicit_status = Impargs.implicit_kind = Implicit | MaximallyImplicit | NotImplicit
```
SkySkimmer: Please deprecate this: add `[@ocaml.deprecated]` after each constructor and `[@@ocaml.deprecated]` after the whole declaration.


Or just remove it if you prefer, we give no guarantees of stability in the ml API.

interp/impargs.ml
```diff
@@ -690,6 +690,9 @@ let projection_implicits env p impls =
   CList.skipn_at_least npars impls
 
 let declare_manual_implicits local ref ?enriching l =
+  assert (List.for_all (List.for_all (fun (_, (max, fi, fu)) -> fi && fu)) l);
+  assert (List.for_all (List.for_all (fun (ex, _) -> match ex with ExplByPos (_,_) -> true | _ -> false)) l);
+  assert (List.length l = 1);
```
SkySkimmer: Please enforce this length invariant by typing instead.
Not sure what the other asserts mean, if reasonable to do please also enforce by typing.

interp/impargs.ml
```diff
@@ -698,28 +701,80 @@ let declare_manual_implicits local ref ?enriching l =
...
+    if is_local local ref then ImplLocal
+    else ImplInteractive(ref,flags,ImplManual (List.length autoimpls))
+  in add_anonymous_leaf (inImplicits (req,[ref,l']))
+
+let maybe_declare_manual_implicits local ref ?enriching l =
+  match l with
+  | [] -> ()
+  | _ -> declare_manual_implicits local ref ?enriching [l]
+
+(* TODO: either turn these warnings on and document them, or handle these cases sensibly *)
```
SkySkimmer: This needs to be resolved before the PR can be merged (let's try to avoid introducing new undocumented warnings, even disabled by default).

interp/impargs.ml
```diff
@@ -698,28 +701,80 @@ let declare_manual_implicits local ref ?enriching l =
...
+    | Name id :: autoimps, MaximallyImplicit :: manualimps ->
+       Some (id, Manual, (true, true)) :: aux (i+1) (autoimps, manualimps)
+    | Name id :: autoimps, Implicit :: manualimps ->
+       let imps' = aux (i+1) (autoimps, manualimps) in
+       let max = set_maximality imps' false in
+       if max then warn_set_maximal_deprecated i;
+       Some (id, Manual, (max, true)) :: imps'
+    | Anonymous :: _, (Implicit | MaximallyImplicit) :: _ ->
+       user_err ~hdr:"set_implicits"
+         (strbrk ("Argument number " ^ string_of_int i ^ " cannot be declared implicit."))
```
SkySkimmer: Maybe add an explanation "because it is originally anonymous" or some such?

SkySkimmer (changes requested):

SkySkimmer:
I'm going to assign, if some implicits expert wants it they can reassign.

SkySkimmer (approved):
I pushed the changes that I said should be done.

AFAICT the extra argument warning could not be triggered as number of arguments is checked earlier in the call stack so I replaced it with an assert false.

Someone else should assign and merge, maybe @herbelin? TBH This PR has waited a while so anyone can take it as far as I'ml concerned.

Zimmi48 (approved):
Doc is fine (couldn't review the rendered version though).
Build is failing because warnings are emitted by the stdlib and should be fixed before merging.

theories/Lists/List.v
```diff
@@ -1033,7 +1033,7 @@ Lemma map_ext_in_iff :
    forall (A B : Type)(f g:A->B) l, map f l = map g l <-> forall a, In a l -> f a = g a.
 Proof. split; [apply ext_in_map | apply map_ext_in]. Qed.
 
-Arguments map_ext_in_iff [A B f g l].
+Arguments map_ext_in_iff {A B f g l}.
```
SkySkimmer: Something strange is going on here, why is the immediately above `Arguments ext_in_map [A B f g l].` not triggering the warning?

jashug: Because `map f l = map g l` is the final argument, which is set to non-implicit. So `ext_in_map H` will expand to `ext_in_map _ _ _ _ _ H`, while `ext_in_map` will not expand to `ext_in_map _ _ _ _ _` because the arguments are not maximally implicit.

jashug:
Sorry for being unresponsive on this, and thanks SkySkimmer for taking the initiative here.

Zimmi48:
Just a heads-up, there is a semantic conflict between this PR and #9654. When the first one is merged, the other will need a trivial fix (adding the `warn` option to the new `coqtop` block).

Zimmi48:
@SkySkimmer You said you were going to assign but you didn't.

SkySkimmer:
@Zimmi48 see https://github.com/coq/coq/pull/9389#pullrequestreview-207949849

Zimmi48:
Ah right, I missed this. @herbelin @ppedrot @mattam82 One of you should self-assign.

ejgallego:
I will take this as per-policy, 10 days have passed.

ejgallego (approved):

Zimmi48:
Hum, as I announced, there was a semantics conflict with #9654 and now the two PRs have been merged so the build is broken on master. I will open a PR fixing this.

PR #9394, opened 2019-01-24
[nix-ci] Maintenance

vbgl:
A few improvements:

  - fix the build system of Unicoq
  - adds QuickChick
  - adds a convenience option to exclude Coq and Coq libraries from the dependencies.

### DIFF
--- a/dev/ci/nix/README.md
+++ b/dev/ci/nix/README.md
@@ -17,3 +17,10 @@ build-system of that project: `configure`, `make`, and `clean`. Therefore, after
 changing the working directory to the root of the sources of that project, the
 contents of these variables can be evaluated to respectively set-up, build, and
 clean the project.
+
+## Variant: nocoq
+
+The dependencies of the third-party developments are split into `buildInputs`
+and `coqBuildInputs`. The second list gathers the Coq libraries. In case you
+only want the non-coq dependencies (because you want to use Coq from your `PATH`),
+set the environment variable `NOCOQ` to some non-empty value.
--- a/dev/ci/nix/CoLoR.nix
+++ b/dev/ci/nix/CoLoR.nix
@@ -1,5 +1,5 @@
 { bignums }:
 
 {
-  buildInputs = [ bignums ];
+  coqBuildInputs = [ bignums ];
 }
--- a/dev/ci/nix/Corn.nix
+++ b/dev/ci/nix/Corn.nix
@@ -1,5 +1,5 @@
 { bignums, math-classes }:
 
 {
-  buildInputs = [ bignums math-classes ];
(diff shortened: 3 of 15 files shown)


maximedenes (approved):
Wonderful!

dev/ci/META
```diff
@@ -0,0 +1,2 @@
+archive(native) = "unicoq.cmxa"
+plugin(native) = "unicoq.cmxs"
```
SkySkimmer: What's this?

dev/ci/unicoq-num.patch
```diff
@@ -0,0 +1,44 @@
+commit f29bc64ee3d8b36758d17e1f5d50812e0c93063b
```
SkySkimmer: What stops this patch from going upstream?

vbgl: Nobody seems competent to write a proper install script for Windows:
https://github.com/unicoq/unicoq/pull/17#issuecomment-442302813

SkySkimmer: That PR is merged and seems to do the same thing as the patch, so why do we need the patch?

SkySkimmer (changes requested):
If dev/ci/META and dev/ci/unicoq-num.patch are nix specific then they should be moved to dev/ci/nix, otherwise please explain how they should be used.
Also the META should probably have a more specific name eg META.unicoq

dev/ci/unicoq-num.patch
maximedenes: IIRC it's been reverted.

vbgl:
I moved the files related to `Unicoq` to a dedicated directory.

SkySkimmer (approved):

ejgallego (approved):
Good from my part, I think that all the issues with Unicoq should be solved once we switch Nix to a Dune build. I am targeting the new Dune verison for next week.

dev/ci/nix/README.md
```diff
@@ -17,3 +17,10 @@ build-system of that project: `configure`, `make`, and `clean`. Therefore, after
 changing the working directory to the root of the sources of that project, the
 contents of these variables can be evaluated to respectively set-up, build, and
 clean the project.
+
+## Variant: nocoq
+
+The dependencies of the third-party developments are split into `buildInputs`
+and `coqBuildInputs`. The second list gathers the Coq libraries. In case you
+only want the non-coq dependencies (because you want to use Coq from you `PATH`),
```
Zimmi48:
```suggestion
only want the non-coq dependencies (because you want to use Coq from your `PATH`),
```

Zimmi48 (commented):
`NOCOQ` sounds useful

PR #9395, opened 2019-01-24
Global [open Univ] in UState

SkySkimmer:
This is cleaner IMO.


ppedrot (approved):
No specific preference, so why not.

PR #9399, opened 2019-01-24
[ide] fail on unavailable commands before adding to the document

### LINKED ISSUE #9385: Parsing at the tip of the document in CoqIDE
#### Version

Coq 8.8.2

#### Operating system

Red Hat Enterprise Linux Workstation 7.4

The following actions prevent further navigation forward.

* Open CoqIDE.
* Insert `Debug On.` into the buffer.
* Navigate forward.
* Ignore the error message.
* Insert `Set` before `Debug`.
* Replace `On` with `Unification`.
* Navigate forward.

```
Currently, the parsing api only supports parsing at the tip of the document.
You wanted to parse at: 2 but the current tip is: 3
```

The only way to recover is to navigate backward.

#### Notes

This is not a duplicate of #7142,
because triggering this bug does not require first triggering an anomaly.

gares:
not that not only the STM was getting out of sync, but also the toplevel (since it was calling `set_doc` before raising the error).

Fix #9385

### DIFF
--- a/ide/fake_ide.ml
+++ b/ide/fake_ide.ml
@@ -241,6 +241,9 @@ let eval_print l coq =
   | [ Tok(_,"ADD"); Top [Tok(_,name)]; Tok(_,phrase) ] ->
       let eid, tip = add_sentence ~name phrase in
       after_add (base_eval_call (add ((phrase,eid),(tip,true))) coq)
+  | [ Tok(_,"FAILADD"); Tok(_,phrase) ] ->
+      let eid, tip = add_sentence phrase in
+      after_fail coq (base_eval_call ~fail:false (add ((phrase,eid),(tip,true))) coq)
   | [ Tok(_,"GOALS"); ] ->
       eval_call (goals ()) coq
   | [ Tok(_,"FAILGOALS"); ] ->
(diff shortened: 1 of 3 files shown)


maximedenes:
Is there a way to get a test case?

gares:
Btw the bug was introduced in e1637bef0f38b46840f146f146ddd4f17bec129a 
I think that by splitting the checks into two groups I did not re-open the bug the commit above was closing.

ejgallego (approved):
Indeed I think this looks good.

gares:
I added a test, but I've also strengthened the patch, so you should look at it again.

The point is that the failure should also tell the UI which is the last safe id.
CoqIDE seems to have resilience in case this information is not provided.
Fake_ide does not, so I had to fix even more the code of idetop to enrich the exception with the correct piece of info.

This bug shows a deficiency in the API of the STM. In particular adding extra checks (and failures) is not easy, since one has to enrich exceptions by hand. The APIs of the lower layers let you do so, but one would expect it to be simpler.

ejgallego:
This seems OK too.

ppedrot (commented):
This is STM-specific, not competent.

PR #9400, opened 2019-01-24
Disable warning 40 Constructor or label name used out of scope.

SkySkimmer:
I stopped believing that it's worth bothering with.

### DIFF
--- a/.merlin.in
+++ b/.merlin.in
@@ -1,4 +1,4 @@
-FLG -rectypes -thread -safe-string -w +a-4-9-27-41-42-44-45-48
+FLG -rectypes -thread -safe-string -w +a-4-9-27-40-41-42-44-45-48
 
 S clib
 B clib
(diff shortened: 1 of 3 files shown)


ejgallego:
On the contrary, this warning prevents about a very dangerous situation and there is even discussion about making it an error upstream. Thus in principle I am against your change.

But I guess that if you provide some motivation or concrete use case for disabling the warning we could reconsider.

ejgallego:
cc: @gasche

ejgallego:
> What changed from [ocaml/ocaml#1333](https://github.com/ocaml/ocaml/pull/1333)?

It can make sense that the warning's default is off in OCaml given how annoying/hard to port some developments could be. However in Coq we already follow a stronger scoping discipline which I'm not sure we'd like to loose.

Anyways, a use case could convince me.

SkySkimmer:
An annoying case is 
~~~ocaml
match whatever with 
| Qualified.Foo -> bla
| Bar(* warning 40 + warning 42 *) -> bli
~~~

ejgallego:
Then I'd vote not to implement this change for now, we could discuss at the WG if @SkySkimmer thinks it would be interesting to gather more opinions.

PR #9405, opened 2019-01-25
Remove sed command which fails with MacOS's sed

mattam82:
**Kind:**  bug fix / infrastructure.

This patch removes sed invocations that are not valid with macOS's BSD sed and make the build fail depending on the make version.


RalfJung:
Ah yes, the fun of `sed` on macOS. This is the reason why I switched to `awk` for Iris' Makefile patching.

mattam82:
It's curious that appveyor reports an error while the log shows no error, is it a timeout thing?

SkySkimmer:
Should have a CHANGES I think.

mattam82:
Actually, the echo solution failed on nix and windows, so I reverted to removing the incriminated line

SkySkimmer:
Probably it interpreted \n badly, you want `printf '\n\n\n'` (with single quotes) or `echo; echo; echo` for reliable newlines I think.

vbgl (approved):

Issue #7724, opened 2018-06-06
The default for Firstorder Solver should be auto, not auto with *


JasonGross:
#### Version

8.8.0

#### Description of the problem

It's very unfortunate that `firstorder` means `firstorder auto with *` by default.  I think we should deprecate this behavior by setting it to `auto` by default and adding `Global Set Firstorder Solver auto with *` to the compat file.  (And similarly for `intuition`.)

PR #9406, opened 2019-01-25
[stdlib] Do not put INR_eq in the “real” hint database

vbgl:
According to Laurent (@thery) on the [mailing list](https://sympa.inria.fr/sympa/arc/coq-club/2019-01/msg00127.html), the `INR_eq` “has wrongly been put in the hint resolve database of reals.”

The test case is the example given by Caitlin D’Abrera (@caitlindabrera).

### DIFF
--- a/test-suite/output/NoAxiomFromR.out
+++ b/test-suite/output/NoAxiomFromR.out
@@ -1,0 +1,1 @@
+Closed under the global context
--- a/test-suite/output/NoAxiomFromR.v
+++ b/test-suite/output/NoAxiomFromR.v
@@ -1,0 +1,10 @@
+Require Import Psatz.
+
+Inductive TT : Set :=
+| C : nat -> TT.
+
+Lemma lem4 : forall (n m : nat),
+S m <= m -> C (S m) <> C n -> False.
+Proof. firstorder. Qed.
+
+Print Assumptions lem4.
--- a/theories/Reals/RIneq.v
+++ b/theories/Reals/RIneq.v
@@ -1692,7 +1692,6 @@ Proof.
   exact H.
   now apply not_INR in H.
 Qed.
-Hint Resolve INR_eq: real.
 
 Lemma INR_le : forall n m:nat, INR n <= INR m -> (n <= m)%nat.
 Proof.


thery:
A shorter test is the following:
```
Require Import Reals.
Lemma test : forall m, m + 1 <= m -> 2 = 1.
auto with *.
Qed.
Print Assumptions test.
```

herbelin:
That'd wonderful is someone has time to look at the hints in the `Real` library. For instance, another direction would be remove cycles. If I remember correctly, there was one around `ge` and `le`, or/and  `gt` and `lt`, while, in principle we could decide once for all that `ge` rewrite to `le` and have other hints apply only on `le`.

thery:
I am impressed no proof was broken, are flocq and coquelicot part of the checking process?

thery:
@herbelin 
I think there is no problem omega/lia. Here the problem comes from trying to solve a goal on nat
using auto with * that captures everything (related to https://github.com/coq/coq/issues/7724).
Here the patch just removes the possible ping-pong between (nat/R).

herbelin:
I indeed misunderstood the original issue about the detour via `R` causing an artificial dependency on the axioms for reals.

I also made a confusion on the direction `INR_eq` was used. 

Nevertheless, there are in principle (a few) cases where `INR_eq` might be useful as in proving `forall m n, INR m = INR n -> n = m`. Also, the strategy is not reasonable. If we have a goal mixing reals of the form `INR t` in the hypotheses and `nat` in the conclusion, it would seem wiser to get rid of `INR` in the hypotheses than lifting the conclusion in `R` as `INR_eq` does (e.g. in `forall m n, INR m = INR n -> n = m`, a strategy would be to rather simplify the hypothesis into `m = n` rather than complexifying the conclusion). But `auto` is not able yet to apply canonical simplifications on the hypotheses.

Interestingly if I do:
```
Require Import Reals.
Lemma test : forall m, m + 1 <= m -> 2 = 1.
auto with zarith real. (* rather than auto with real zarith *)
Qed.
```
the detour disappears. So we can also see the issue as a question of priority of hints.

Anyway, I have no particular opinion on whether it is good to have `INR_eq` as a hint or not in the absolute or on whether it would be better to play with priorities. Another possibility would be to have an `Hint Extern` with a `match goal` to apply it forwardly on the hypotheses but I'm afraid that this would be too costly (compared to a native support of `auto` to apply lemmas on the hypotheses).

vbgl:
Is this change welcome? If this is the case, can someone assign and merge?

Otherwise I shall close this PR. Thanks.

silene:
The change is welcome. As @herbelin remarks, applying `INR_eq` to the goal is almost always a mistake, but applying it to a hypothesis makes sense. So, merging this patch is a good first step, and we might later consider whether we want to add a dedicated `Hint Extern` for hypotheses.

silene (approved):

ejgallego:
What's holding this?

herbelin:
> What's holding this?

Maybe the question of adding a `Hint Extern` instead. But I feel that what we want is that it is applied to hypotheses once and for all (e.g. at the time an hypothesis is introduced) and that it would be costly to try the extern hint at each recursive step of the search.

So, @silene approved and I'd be ok to proceed to avoir further delays.

ejgallego:
> So, merging this patch is a good first step, and we might later consider whether we want to add a dedicated `Hint Extern` for hypotheses.

Thanks @herbelin , I inferred from the comment above that this PR could be merged and the Hint considered at a later stage, so indeed proceeding looks good.

ppedrot:
Do I understand correctly that the PR is mergeable as-is?

silene:
Yes.

ppedrot:
I assigned and rebased, will merge if CI agrees.

ppedrot:
CI failure spurious, let's merge.

PR #9408, opened 2019-01-25
Remove Keep objects

maximedenes:
This is a step towards simplifying the libobject API.

I'm not sure to grasp 100% of the subtleties of Keep vs Substitute, but for the part
I understand, this change should be ok. Let's rely on testing.

### DIFF
--- a/library/libobject.ml
+++ b/library/libobject.ml
@@ -13,7 +13,7 @@ open Pp
 module Dyn = Dyn.Make ()
 
 type 'a substitutivity =
-    Dispose | Substitute of 'a | Keep of 'a | Anticipate of 'a
+    Dispose | Substitute of 'a | Anticipate of 'a
 
 type object_name = Libnames.full_path * Names.KerName.t
 
@@ -34,7 +34,7 @@ let default_object s = {
   open_function = (fun _ _ -> ());
   subst_function = (fun _ ->
     CErrors.anomaly (str "The object " ++ str s ++ str " does not know how to substitute!"));
-  classify_function = (fun obj -> Keep obj);
+  classify_function = (fun obj -> Dispose);
   discharge_function = (fun _ -> None);
   rebuild_function = (fun x -> x)}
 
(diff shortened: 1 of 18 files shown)


gares (approved):
ssr changes are OK, for what that matters

library/libobject.ml
```diff
@@ -34,7 +34,7 @@ let default_object s = {
   open_function = (fun _ _ -> ());
   subst_function = (fun _ ->
     CErrors.anomaly (str "The object " ++ str s ++ str " does not know how to substitute!"));
-  classify_function = (fun obj -> Keep obj);
+  classify_function = (fun obj -> Dispose);
```
SkySkimmer: I'd prefer to make this anomaly "$s isn't classified" or some such thing.

ejgallego (approved):
Parsing part is trivial.

SkySkimmer (changes requested):
Please force people to pick a classification in default object by making the default an anomaly.

SkySkimmer:
@maximedenes ping

maximedenes:
> @maximedenes ping

I'm working on other parts of this simplification plan at the moment. Any reason why you ping this particular PR?

SkySkimmer:
I was looking through random unassigned PRs.

ppedrot:
I didn't comment at the time, but I am seizing the opportunity of @SkySkimmer's ping to do so.

The question is the following: without Keep objects, how are we going to be able to create libobjects tied to a functorized definition? Currently, you can define an object that is somewhat global respectively to a kernel entry living inside a functor. With only Substitute objects, their initialization only happens when the functor is instantiated.

A typical example, which is somewhat broken currently, is a definition-attached data that shouldn't go in the kernel, e.g. whether a one-constructor inductive should be treated as a record by the upper layers. Currently we store this info in the kernel, but the only real user, if we except user display, is extraction. Bloating the kernel with irrelevant data should be frowned upon, but without (fixed up) Keep objects, I don't see how to do that.

SkySkimmer:
Future of this is unclear, also dead PR so closing.

PR #9409, opened 2019-01-25
Move non-primitive-record warning to declare_mutual_inductive_with_eliminations

SkySkimmer:
This makes it print the warning before the definition message, so if
we run with +non-primitive-record we don't see
"""
defined foo
error could not define foo
"""


ppedrot (approved):
LGTM.

Issue #6216, opened 2017-11-22
Program Mode options


ejgallego:
Hi all,

there are a couple of places where `Program` introduces special behavior. I wonder what should be done about these places, in particular due that in a functional design of the core, flags would have to be propagated and that will be cumbersome.

I guess maybe making the program behavior the default would make sense, anyways I'm opening this issue for discussion.

In particular the two places that seem very low in the hierarchy are:

- evarutil.ml:
```ocaml
    if Flags.is_program_mode () then next_name_away na avoid
    else
```

The next part is in `coercion.ml`:
```ocaml
let inh_coerce_to_base ?loc env evd j =
  if Flags.is_program_mode () then
```

Next parts are already in `cases` and pretyping.

ppedrot:
The `Evarutil` use could be removed if we start using open recursion for the pretyper. Essentially, it's just that there are many ways to generate hypothesis names for evars and that Program has a specific, incompatible behaviour w.r.t. vanilla naming.

ejgallego:
My point is more of "do we want to foment such heterogeneous behavior"? Is it justified having several naming generation strategies? It seems to me like that is a recipe for problems.

herbelin:
About "several naming generation strategies", my conclusion at some time was that we should eventually provide a way to support configurable naming strategies at some time. There are many places where we can do better and to be able to do better while preserving the existing, such configurability would help.

ejgallego:
@herbelin a problem with that is that you will need to parameterize _really a lot_ of functions on the naming strategy then. Basically the whole system above `engine` would need to carry a parameter "naming strategy being used".

So IMVHO we should consider the tradeoff on this. Of course, a naming strategy that lives only at the tactic level would be much more reasonable I think.

ejgallego:
Fixed by #9410

PR #9410, opened 2019-01-25
Make `Program` a regular attribute

maximedenes:
This may break some scripts depending on name generation, but let's
first see how bad it is.

The motivation of the change is to turn Program into a regular attribute
(threaded instead of being accessed to through `Flags` directly). This
is especially important since it has an impact on proof modes, so on the
separation of parsing and execution phases.

This PR also makes attribute checking a bit stricter. In particular, `Inductive` and friends (`Class`, `Record`) don't accept locality attributes anymore. Previously, they were silently discarded.

### DIFF
--- a/vernac/vernacentries.ml
+++ b/vernac/vernacentries.ml
@@ -514,9 +514,9 @@ let () =
 (***********)
 (* Gallina *)
 
-let start_proof_and_print ?hook k l =
+let start_proof_and_print ~program_mode ?hook k l =
   let inference_hook =
-    if Flags.is_program_mode () then
+    if program_mode then
       let hook env sigma ev =
         let tac = !Obligations.default_tactic in
         let evi = Evd.find sigma ev in
@@ -536,7 +536,7 @@ let start_proof_and_print ?hook k l =
       in Some hook
     else None
   in
-  start_proof_com ?inference_hook ?hook k l
+  start_proof_com ~program_mode ?inference_hook ?hook k l
 
 let vernac_definition_hook p = function
 | Coercion ->
@@ -549,7 +549,6 @@ let vernac_definition_hook p = function
 
 let vernac_definition ~atts discharge kind ({loc;v=id}, pl) def =
   let open DefAttributes in
-  let atts = parse atts in
   let local = enforce_locality_exp atts.locality discharge in
   let hook = vernac_definition_hook atts.polymorphic kind in
   let () =
@@ -560,7 +559,7 @@ let vernac_definition ~atts discharge kind ({loc;v=id}, pl) def =
       | Discharge -> Dumpglob.dump_definition lid true "var"
       | Local | Global -> Dumpglob.dump_definition lid false "def"
   in
-  let program_mode = Flags.is_program_mode () in
+  let program_mode = atts.program in
   let name =
     match id with
     | Anonymous -> fresh_name_for_anonymous_theorem ()
@@ -568,7 +567,7 @@ let vernac_definition ~atts discharge kind ({loc;v=id}, pl) def =
   in
   (match def with
     | ProveBody (bl,t) ->   (* local binders, typ *)
-      start_proof_and_print (local, atts.polymorphic, DefinitionBody kind)
+      start_proof_and_print ~program_mode (local, atts.polymorphic, DefinitionBody kind)
         ?hook [(CAst.make ?loc name, pl), (bl, t)]
     | DefineBody (bl,red_option,c,typ_opt) ->
       let red_option = match red_option with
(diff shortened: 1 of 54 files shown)


SkySkimmer:
>The motivation of the change is to turn Program into a regular attribute

There are quite a few calls in pretyping, what's the plan for those?

maximedenes:
I don't know if they all have to stay, @mattam82 can probably comment on it.

At least, it is easier to thread the flag down to the pretyper than to unification (I hope).

mattam82:
In pretyping there is the coercion stuff and obligation handling, which has to stay I think, and the bidirectional typechecking hint for constructors of inductive types, which could be a separate flag actually.

maximedenes:
> In pretyping there is the coercion stuff and obligation handling, which has to stay I think, and the bidirectional typechecking hint for constructors of inductive types, which could be a separate flag actually.

I simplified a bit. I don't mean to remove the flags, but rather I was wondering if all code paths should lead to depending on the flag. For example, if an ltac construction calls the pretyper (can that happen?), should it really depend on the `Program` flag?

SkySkimmer:
>can that happen

`exact of_course`

maximedenes:
> exact of_course

exactly! :)

mattam82:
I see. One can use the subset coercion algorithm and pattern-matching compilation of Program when using `refine` for example, but I would bet it is seldom used. Would it cost much to carry this flag?

maximedenes:
Ok, in fact this PR is now more ambitious: it makes `Program` a regular attribute.

SkySkimmer:
Then the flags.program_mode ref should be moved next to the attribute definition and private to that module, like with polymorphism.

aspiwack (approved):

vernac/vernacentries.ml
```diff
@@ -675,9 +672,9 @@ let extract_inductive_udecl (indl:(inductive_expr * decl_notation list) list) =
     indicates whether the type is inductive, co-inductive or
     neither. *)
 let vernac_inductive ~atts cum lo finite indl =
-  let open DefAttributes in
   let atts, template = Attributes.(parse_with_extra template atts) in
-  let atts = parse atts in
+  let extra, poly = Attributes.(parse_with_extra polymorphic atts) in
+  unsupported_attributes extra;
```
SkySkimmer: Just do `let poly, template = Attributes.(parse Notations.(polymorphic ++ template) atts) in`

maximedenes: Done, thanks!

ejgallego:
@maximedenes does this fix #6216 , if so please make the commit message reflect it.

ejgallego (approved):
My parts are OK, great cleanup.

maximedenes:
I believe it does fix it, so I'll update the commit message once iris is updated (so that the CI run should hopefullly be successful).

maximedenes:
I also need to update `CHANGES` as there are some minor semantic changes for `Set Program Mode` (previously, the flag was doing much more than adding `Program` in front of commands which support it).

mattam82 (approved):
Great cleanup indeed. I guess we only lose the "globality" of program mode behavior of before where tactics like refine could use program's refinement mode when in Program Mode?

vernac/vernacentries.ml
```diff
@@ -2434,21 +2421,13 @@ let interp ?(verbosely=true) ?proof ~st {CAst.loc;v=c} =
...
       try
         vernac_timeout begin fun () ->
           if verbosely
           then Flags.verbosely (interp ?proof ~atts ~st) c
           else Flags.silently  (interp ?proof ~atts ~st) c;
-          (* If the command is `(Un)Set Program Mode` or `(Un)Set Universe Polymorphism`,
-             we should not restore the previous state of the flag... *)
-          if Option.has_some program then
-            Flags.program_mode := orig_program_mode;
           end
```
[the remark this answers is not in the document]
maximedenes: I was happy about this one ;)

maximedenes:
> I guess we only lose the "globality" of program mode behavior of before where tactics like refine could use program's refinement mode when in Program Mode?

Exactly. But CI did not reveal any incompatibility there. And also, the naming strategy used by `Program` is not propagated to tactics / TC resolution done when the obligation is opened. This showed up once in the CI, but was very easy to fix. `Program` pretyping itself should be 100% compatible.

mattam82:
Great, I didn't expect much use of it. And we can go from here if we ever want to reintroduce that feature, the activation model is obvious now.

maximedenes:
> Great, I didn't expect much use of it. And we can go from here if we ever want to reintroduce that feature, the activation model is obvious now.

In fact, I believe we'd like to enable some of the `Program` features unconditionally, like bidirectional typing. Of course, it will require a proper compatibility study / transition strategy.

mattam82:
Yes. For bidir I would expect a separate attribute/option. I imagine we could split case compilation, subset coercions and general inductive coercions also but that seems a bit overkill (I don't have use cases for that).

gares (approved):
respect

mattam82:
Everything seems to be in order, so I’ll merge quickly unless someone objects today

PR #9426, opened 2019-01-28
[test-suite] Fix display of check.

ejgallego:
After #8655

### DIFF
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -307,7 +307,7 @@ $(addsuffix .log,$(wildcard prerequisite/*.v)): %.v.log: %.v
 	    echo "    $<...correctly prepared" ; \
 	  fi; \
 	} > "$@"
-	@echo "CHK $(shell basename $< .v)"
+	@echo "CHECK     $(shell basename $< .v)"
 	$(HIDE)$(coqchk) -norec TestSuite.$(shell basename $< .v) > $(shell dirname $<)/$(shell basename $< .v).chk.log 2>&1
 
 ssr: $(wildcard ssr/*.v:%.v=%.v.log)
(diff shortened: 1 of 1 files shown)


SkySkimmer (changes requested):
We should either stick with "chk" to match coqchk, or fully expand to "check"

ejgallego:
> We should either stick with "chk" to match coqchk, or fully expand to "check"

I do prefer to keep the alignment with `TEST` which is 4 chars, and seems way more pleasant aesthetically, what do others think?

[I'm fine with any 4-letter combination tho, or then just `CHECK` indeed]

ejgallego:
> While you are at it, please also fix this discrepancy:

Sure, in which way tho? The basename there was added on purpose by @vbgl , I guess we'd like indeed to remove it right? [I do prefer full paths]

vbgl:
I can’t remember why I put this `basename` there. Feel free to remove them. Thanks.

gares:
```
TEST      failure/Case7.v 
CHECK     failure/Case7.vo
```

gares:
Actually, I quite don't get why we write `TEST` now that we run `COQC`...

SkySkimmer:
Rather I don't get why we write check

gares:
because we are running coqchk to check the .vo

gares:
Oh, I see, you want COQCHK... fine

SkySkimmer:
That could be viewed as part of testing no?

Gaëtan Gilbert

gares:
yes but we don't run it on all files... eg misc or coq-makefile stuff

ejgallego:
For me the current status of this comestic PR is OK now, and fixes an annoying problem. If you want to change `TEST` to other stuff that's fine, but maybe it does belong to a different change.

gares (approved):

ejgallego:
Thanks!

Issue #9429, opened 2019-01-29
Distributed PDF reference manual doesn't use calligraphic font


akr:
#### Version

https://github.com/coq/coq/releases/download/V8.9.0/coq-8.9.0-reference-manual.pdf

#### Operating system

% lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 9.7 (stretch)
Release:	9.7
Codename:	stretch

I use evince 3.22.1-3+deb9u1 to view the PDF file.

#### Description of the problem

I found the distributed PDF reference manual doesn't use a calligraphic font.
I guess the manual is formatted in a environment which lacks a font.

For example, there is a following paragraph in p.103 of coq-8.9.0-reference-manual.pdf : 

![no-calligraphic-font](https://user-images.githubusercontent.com/10538/51901507-2ff72400-23fb-11e9-9baa-8e0cb90d25f7.png)

The last two line describes the difference of two kind of "S".
But two "S" uses same font and they are not distinguishable.
The latter "S" should use calligraphic font.

The corresponding source in cic.rst is follows:

```
These inductive definitions, together with global assumptions and
global definitions, then form the global environment. Additionally,
for any :math:`p` there always exists :math:`Γ_P =[a_1 :A_1 ;…;a_p :A_p ]` such that
each :math:`T` in :math:`(t:T)∈Γ_I \cup Γ_C` can be written as: :math:`∀Γ_P , T'` where :math:`Γ_P` is
called the *context of parameters*. Furthermore, we must have that
each :math:`T` in :math:`(t:T)∈Γ_I` can be written as: :math:`∀Γ_P,∀Γ_{\mathit{Arr}(t)}, S` where
:math:`Γ_{\mathit{Arr}(t)}` is called the *Arity* of the inductive type t and :math:`S` is called
the sort of the inductive type t (not to be confused with :math:`\Sort` which is the set of sorts).
```

The latter "S" uses "\Sort" command defined in refman-preamble.sty:

```
\newcommand{\Sort}{\cal S}
```

\Sort command uses \cal command.
So, it should be formatted in calligraphic font.

Actually, it is formatted in calligraphic font in my environment when I built the reference manual from the source as : 

![uses-calligraphic-font](https://user-images.githubusercontent.com/10538/51901941-2e7a2b80-23fc-11e9-907a-31e132aa7c11.png)

So, I guess the distributed reference manual is built on a environment which lacks a font.

[...]

PR #9435, opened 2019-01-29
Use \mathcal instead of \cal

SkySkimmer:
Apparently it's deprecated / doesn't always work, see
https://tex.stackexchange.com/questions/84041/why-does-calm-n-give-m

Tested both on my machine and on my gitlab.

See #9429 (we also need to fix the distributed file on the server).

### DIFF
--- a/dev/doc/versions-history.tex
+++ b/dev/doc/versions-history.tex
@@ -271,7 +271,7 @@ Coq ``V7'' archive & August 1999 & new cvs archive based on J.-C. Filliâtre's \
    & & \feature{kernel-centric} architecture \\
    & & more care for outside readers\\
    & & (indentation, ocaml warning protection)\\
-Coq V7.0beta& released 27 December 2000 & \feature{${\cal L}_{\mathit{tac}}$} \\
+Coq V7.0beta& released 27 December 2000 & \feature{${\mathcal{L}}_{\mathit{tac}}$} \\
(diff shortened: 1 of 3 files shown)


Zimmi48 (approved):

Zimmi48:
Hum. That's strange. Many jobs were skipped on GitLab CI, including the refman. But you said you've already tested this somewhere else, correct?

Zimmi48:
BTW, the Azure failure is strange as well.

SkySkimmer:
You can get the pdf from https://gitlab.com/SkySkimmer/coq/-/jobs/152960546/artifacts/raw/_install_ci/share/doc/coq/sphinx/latex/CoqRefMan.pdf?inline=false
Very strange CI behaviour for sure

akr:
I think this also fixes an unintentional calligraphic font use in W-Global-Assum in HTML manual.

https://coq.inria.fr/distrib/current/refman/language/cic.html :

unintentional-calligraphy-in-html-manual

"E" in "c ∉ E" in W-Global-Assum is formatted in calligraphic font because
the effect of \cal retains after \Sort.

"E" in W-Global-Def doesn't use calligraphic font because it doesn't use \Sort.

So, it's better to update the HTML manual too.

Zimmi48:
>So, it's better to update the HTML manual too.

For this, I will wait until we set up continuous deployment of the manual.

PR #9438, opened 2019-01-30
Cleanup universe length for inductives in vconv

SkySkimmer:


kernel/vconv.ml
```diff
@@ -86,17 +86,11 @@ and conv_atom env pb k a1 stk1 a2 stk2 cu =
...
-	let ulen = 
-          match mib.Declarations.mind_universes with
-          | Declarations.Monomorphic_ind ctx -> Univ.ContextSet.size ctx
-          | Declarations.Polymorphic_ind auctx -> Univ.AUContext.size auctx
-          | Declarations.Cumulative_ind cumi -> 
-            Univ.AUContext.size (Univ.ACumulativityInfo.univ_context cumi)
-        in
+      let ulen = Univ.AUContext.size (Environ.mind_context env mi) in
+      if ulen = 0 then
+        conv_stack env k stk1 stk2 cu
```
SkySkimmer: Note that this causes a change such that we take the simple path also on polymorphic inductives which have no actual polymorphic universes. This shouldn't change observable behaviour.

SkySkimmer: We could also remove the simple path, although that may need a bench.

SkySkimmer:
Out of memory in async test suite success/auto https://gitlab.com/coq/coq/-/jobs/153472854
Network error in the gialb windows job https://gitlab.com/coq/coq/-/jobs/153499547

SkySkimmer:
Equations failure is spurious due to bad makefile on their side

kernel/vconv.ml
ppedrot: Could you add a test for this specific case (polymorphic without universes) in the test-suite?

ppedrot (approved):
After an offline discussion, it turns out that adding a test would be quite unrelated to this PR (much to my dismay). The code looks fine, so this is ready for merging.

PR #9442, opened 2019-01-30
Update pinned nixpkgs.

Zimmi48:
**Kind:** infrastructure.

Update pinned nixpkgs and use default OCaml version. This should avoid rebuilding things that are in the cache.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -23,10 +23,10 @@
 
 { pkgs ?
     (import (fetchTarball {
-      url = "https://github.com/NixOS/nixpkgs/archive/958a6c6dd39b0d6628e1408e798a8f1308f2f3e1.tar.gz";
-      sha256 = "0vs6k4jn0rbdfzaxmh3xh64q213326680i9g3cjgr7l9y6h6m5sy";
+      url = "https://github.com/NixOS/nixpkgs/archive/11cf7d6e1ffd5fbc09a51b76d668ad0858a772ed.tar.gz";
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -23,10 +23,10 @@
 
 { pkgs ?
     (import (fetchTarball {
-      url = "https://github.com/NixOS/nixpkgs/archive/958a6c6dd39b0d6628e1408e798a8f1308f2f3e1.tar.gz";
-      sha256 = "0vs6k4jn0rbdfzaxmh3xh64q213326680i9g3cjgr7l9y6h6m5sy";
+      url = "https://github.com/NixOS/nixpkgs/archive/11cf7d6e1ffd5fbc09a51b76d668ad0858a772ed.tar.gz";
+      sha256 = "0zcg4mgfdk3ryiqj1j5iv5bljjvsgi6q6j9z1vkq383c4g4clc72";
     }) {})
-, ocamlPackages ? pkgs.ocaml-ng.ocamlPackages_4_06
```
vbgl: Why this change?

IMHO, we prefer to choose the version of OCaml we use independently of what the default is in `nixpkgs`.

vbgl (approved):

default.nix
Zimmi48: The idea was that only the default version is available from NixOS official cache, and we might want to take advantage of this...

vbgl: We’ve got our own cache. We can store in it what we like or need.

PR #9443, opened 2019-01-30
Search: iter through the environment, not libobject

SkySkimmer:
The order of stuff produced by Search is changed, this is OK.

### DIFF
--- a/kernel/environ.mli
+++ b/kernel/environ.mli
@@ -46,8 +46,12 @@ type constant_key = constant_body * (link_info ref * key)
 
 type mind_key = mutual_inductive_body * link_info ref
 
-type globals
-(** globals = constants + projections + inductive types + modules + module-types *)
+type globals = {
+  env_constants : constant_key Cmap_env.t;
+  env_inductives : mind_key Mindmap_env.t;
+  env_modules : module_body MPmap.t;
+  env_modtypes : module_type_body MPmap.t;
+}
 
 type stratification = {
   env_universes : UGraph.t;
--- a/library/declaremods.ml
+++ b/library/declaremods.ml
@@ -969,28 +969,6 @@ let in_import : bool * ModPath.t -> obj =
 let import_module export mp =
   Lib.add_anonymous_leaf (in_import (export,mp))
 
-
-(** {6 Iterators} *)
-
-let iter_all_segments f =
-  let rec apply_obj prefix (id,obj) = match object_tag obj with
-    | "INCLUDE" ->
-      let objs = expand_aobjs (out_include obj) in
-      List.iter (apply_obj prefix) objs
-    | _ -> f (Lib.make_oname prefix id) obj
-  in
-  let apply_mod_obj _ (prefix,substobjs,keepobjs) =
-    List.iter (apply_obj prefix) substobjs;
-    List.iter (apply_obj prefix) keepobjs
-  in
-  let apply_node = function
-    | sp, Lib.Leaf o -> f sp o
-    | _ -> ()
-  in
-  MPmap.iter apply_mod_obj (ModObjs.all ());
-  List.iter apply_node (Lib.contents ())
-
-
 (** {6 Some types used to shorten declaremods.mli} *)
 
 type 'modast module_interpretor =
--- a/library/declaremods.mli
+++ b/library/declaremods.mli
@@ -124,14 +124,6 @@ val declare_include :
   'modast module_interpretor -> ('modast * inline) list -> unit
 
 (** {6 ... } *)
-(** [iter_all_segments] iterate over all segments, the modules'
-    segments first and then the current segment. Modules are presented
-    in an arbitrary order. The given function is applied to all leaves
-    (together with their section path). *)
-
-val iter_all_segments :
-  (Libobject.object_name -> Libobject.obj -> unit) -> unit
-
 
 val debug_print_modtab : unit -> Pp.t
 
--- a/test-suite/output/Search.out
+++ b/test-suite/output/Search.out
@@ -5,9 +5,9 @@ le_n_S: forall n m : nat, n <= m -> S n <= S m
 le_pred: forall n m : nat, n <= m -> Nat.pred n <= Nat.pred m
 le_S_n: forall n m : nat, S n <= S m -> n <= m
 min_l: forall n m : nat, n <= m -> Nat.min n m = n
-max_r: forall n m : nat, n <= m -> Nat.max n m = m
-min_r: forall n m : nat, m <= n -> Nat.min n m = m
 max_l: forall n m : nat, m <= n -> Nat.max n m = n
+min_r: forall n m : nat, m <= n -> Nat.min n m = m
+max_r: forall n m : nat, n <= m -> Nat.max n m = m
 le_ind:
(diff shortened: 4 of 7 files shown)


ejgallego:
Fixes #8495 closes #8228

ejgallego (approved):
IMHO this is a step in the right direction [as I've argued elsewhere] as we replace a hack with something more principled.

In the medium term tho `Search` should not access the kernel interface [which indeed it is a bit low-level], but better means to maintain a database should be devised.

ejgallego:
> I believe that one could get a much better API than just exposing the kind of "map type" that is used to store things. In particular this PR fights the fact that order cannot be respected, because we lack a timestamp that is otherwise implicit in libstack (here we have 4 maps, libstack is one list). In order to respect the order these map types need to change, or be accompanied by extra data structures: an API could hide that detail and make any related change break less code.
> 
> If we don't want to "bloat" the kernel with this API, I think we should put it elsewhere and expose the `globals` business into an `Internal` sumbmodule (so that it is clear we expose it just to be able to implement a better/safer/robust API on top of it elsewhere).

Indeed I think that environment is not the right API to be consumed by `Search`, as indeed you need quite a bit more information to have efficient indexing, however I do believe that `globals` should be just exposed as a low-level API to be consumed for example by the indexer. IMO not worth having the internal wrapper, just a proper documentation stating that this is a low-level interface so you are on your own.

SkySkimmer:
Here's a commit iterating through safe_env structures instead of the unsafe environment. This allows us to have some order in the iteration. I don't know where to find the dependency order between libraries so they're still unordered, but declarations from one cannot be interleaved with another.

SkySkimmer:
>declarations from one cannot be interleaved with another.

Well they shouldn't be, but it seems they still are for some reason.

SkySkimmer:
>Well they shouldn't be, but it seems they still are for some reason.

Because we stick them through a priority queue, so the order of the initial iteration is irrelevant.

ejgallego:
I am not sure I like this new interface, in fact I don't. The original map was much superior than the list-based one, and should used by the upper layers to build the database.

kernel/safe_typing.mli
```diff
@@ -195,6 +195,19 @@ val export :
...
 
+(** {6 Root extraction } *)
+
+type field =
+  | FModule of ModPath.t
+  | FModuleType of ModPath.t
+  | FConst of Constant.t
+  | FMind of MutInd.t
+
+val roots : safe_environment -> field list
```
gares: Two comments:
- with this approach I don't see the need of making `globals` transparent
- this API should take a filtering predicate, at least for the "module part". For example with `Search ... in Foo Bar` I limit my search to two modules, etc... That is, some filtering can be very cheap to do in this function (module name, label match) , while the more advanced one like "pattern matching" can be left to the caller. See https://coq.inria.fr/distrib/current/refman/proof-engine/ssreflect-proof-language.html#ssr-searching-tool

ejgallego: What you propose shouldn't be in the kernel, much less in safe typing; also the `roots` API is pretty suboptimal when compared the maps, which at least have log access time.

I guess a way to evaluate the API is to answer the question "what kind of efficient indexers can you build _outside_ of the kernel". With `globals`, I can build some for a class of prefix-range based-search queries, with `roots`, none.

SkySkimmer: Here's a change to providing a fold function, so you can do your filtering on the fly.
I don't know about indexing, I'm fine resetting to the first commit though.

ejgallego: > Here's a change to providing a fold function, so you can do your filtering on the fly.
> I don't know about indexing, I'm fine resetting to the first commit though.

Fold is still linear so not too good for me; note that current maps in Coq allow to filter by range and work better for small result sets.

IMVHO from the point of view of `Search` and completion the new API is basically as useless as the old one, so I dunno if that's worth the addition to safe_env.

SkySkimmer: >Fold is still linear so not too good for me; note that current maps in Coq allow to filter by range and work better for small result sets.

The hmaps that we have in the environment have a range filter? How?

>IMVHO from the point of view of Search and completion the new API is basically as useless as the old one, so I dunno if that's worth the addition to safe_env.

True

gares:
I thinkI changed my mind.

The catalog of available lemmas/inductives should be in library, eventually indexed and decorated with metadata such as a timestamp.
And I don't see why this data should be derived from the kernel, it does not even need to be trusted.

What is wrong with the old code?

kernel/safe_typing.mli
ejgallego: > The hmaps that we have in the environment have a range filter? How?

Oh indeed, I happened to look at the `MPmap` which is indeed a `CMap`.

SkySkimmer:
>What is wrong with the old code?

I don't like looking at the string tags on objects.

gares:
Then the right fix is to clean up libobject, possibly dropping the idea of a fully extensible but opaque storage and choosing a more rigid one but with a structure that can be traversed.

This may need coordination with @maximedenes

gares (changes requested):
not the right approach imo

SkySkimmer:
I don't understand why we wouldn't look in the kernel for a list of available things.

ejgallego:
> Then the right fix is to clean up libobject, possibly dropping the idea of a fully extensible but opaque storage and choosing a more rigid one but with a structure that can be traversed.

Why should libobject duplicate what's already in the env. If `libobject` is refactored, indeed that should not happen so this PR still would look fairly the same.

gares:
> I don't understand why we wouldn't look in the kernel for a list of available things.

Because the data is already in library, where adding indexing and other features is appropriate.
The fact that the data is in an ugly format is not a reason for looking for it in another place.
You don't want to inspect libobject tags, but you are happy to inspect another data structure that is even more opaque today.

For example, the kernel does not even know the order in which things are declared. You discovered that.
What do we do now? We add that data to the kernel?

I think we should make liboject less hard to inspect.

SkySkimmer:
I still don't like looking at the tags but I don't care enough to argue, 
let's move on.

Gaëtan Gilbert

ejgallego:
> For example, the kernel does not even know the order in which things are declared. You discovered that.
> What do we do now? We add that data to the kernel?
> 
> I think we should make liboject less hard to inspect.

Well, the kernel should know what a module is and that should provide this kind of info.

I don't advocate adding an indexed to the kernel. Clearly, any indexing for search should be performed outside. However, libobject should not contain duplicate information as it does today, and the indexer should work on the data the kernel know.

To that purpose, assume Δ is an update to the kernel env, then
the indexer should also get Δ to keep its database consistent.

In that sense, Gaëtan's indexer is just the identity one; which
admittedly is pretty poor, however that is IMHO better than what
we have today [and allows for more efficient search in some
cases]

ejgallego:
Sad to see this PR go, as it was a net improvement and would have allowed to have nice improvements in Serach for example in the upcoming language server.

Note that the way I see this is:
- kernel maintains the canonical structure:
- libobject maintains Deltas, that are applied to it;
- search listens to deltas and has an efficient engine.

PR #9444, opened 2019-01-30
Universe Polymorphism with Prop (WIP)

### LINKED ISSUE #9294: Template polymorphism breaks Prop/Set separation
ie we can get a Prop equivalent to any Set, in this example equivalent to `nat`.
~~~coq
Inductive Foo@{i} (A:Type@{i}) : Type := foo : (nat:Type@{i}) -> Foo A.
Arguments foo {_} _.

Check Foo True : Prop. (* should fail *)

Definition bar : Foo True -> nat := fun '(foo x) => x.

Definition foo_bar n : foo (bar n) = n.
Proof. destruct n;reflexivity. Qed.

Definition bar_foo n : bar (foo n) = n.
Proof. reflexivity. Qed.
~~~

I think we can't actually get a proof of false from this without the kind of axiom which breaks impredicative set but still marking as critical.

The root cause is that we typecheck the inductive with universes considered >= Set but then template polymorphism allows instantiating with Prop.

Not sure if there's a way to trigger this with neither `@{i}` or explicit `#[universes(template)]` syntax (if I don't use the `@{i}` universes unify differently) but plugins could probably do it since a long time ago (maybe since universe polymorphism, maybe before, I don't know how universes worked before).

mattam82:
This branch explores a redesign of universes supporting to instantiate polymorphic universe binders with Prop and using this to fix (not yet done) #9294 The easy part is allowing Prop, and this works already. The harder part is emulating template-polymorphism using cumulative inductives while keeping compatibility as much as possible.

**Kind:** bug fix / feature

Fixes / closes #????

Current setup:
- First, allow to set a global lower bound on universes (Prop for standard Coq, can be set to Set for HoTT). This allows i such that Prop < i < Set. At least without Set impredicative, I don't see any issue with that (Set is just a name for a specific level).
- Adapt minimization etc... to comply with this (done)
- Try and emulate template-polymorphic inductives with cumulative ones (WIP).

### DIFF
--- a/kernel/environ.ml
+++ b/kernel/environ.ml
@@ -59,6 +59,7 @@ type globals = {
 
 type stratification = {
   env_universes : UGraph.t;
+  env_universes_lbound : Univ.Level.t;
   env_engagement : engagement
 }
 
@@ -117,6 +118,7 @@ let empty_env = {
   env_nb_rel = 0;
   env_stratification = {
     env_universes = UGraph.initial_universes;
+    env_universes_lbound = Univ.Level.prop;
     env_engagement = PredicativeSet };
   env_typing_flags = Declareops.safe_flags Conv_oracle.empty;
   retroknowledge = Retroknowledge.initial_retroknowledge;
@@ -251,6 +253,7 @@ let deactivated_guard env = not (typing_flags env).check_guarded
 let indices_matter env = env.env_typing_flags.indices_matter
 
 let universes env = env.env_stratification.env_universes
+let universes_lbound env = env.env_stratification.env_universes_lbound
 let named_context env = env.env_named_context.env_named_ctx
 let named_context_val env = env.env_named_context
 let rel_context env = env.env_rel_context.env_rel_ctx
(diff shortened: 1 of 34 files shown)


kernel/environ.ml
```diff
@@ -59,6 +59,7 @@ type globals = {
 
 type stratification = {
   env_universes : UGraph.t;
+  env_universes_lbound : Univ.Level.t;
```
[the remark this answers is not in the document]
mattam82: That can be done as well.

mattam82:
Thinking a bit out loud here but. On top of the first commit, which is not too invasive,  we could implement a restriction of template polymorphism that would already give us soundness I believe: 
- only levels l that are introduced by the inductive declaration and on which no constraint except lbound <= l apply can be instantiated by template polymorphism.

This ensures that any substitution for l results in a typable copy of the inductive, which is how I interpret a template instantiation.

test-suite/bugs/closed/HoTT_coq_120.v
```diff
@@ -3,14 +3,14 @@ Require Import TestSuite.admit.
 Set Universe Polymorphism.
 Generalizable All Variables.
 Reserved Notation "g 'o' f" (at level 40, left associativity).
-Inductive paths {A : Type} (a : A) : A -> Type :=
+Inductive paths@{i} {A : Type@{i}} (a : A) : A -> Type@{i} :=
```
JasonGross: If I explicitly use `Type`, it should not be minimized to `Prop`.  Or, at least, there should be `Unset Universe Minimization ToProp ` to cause this behavior.  Then this line should not need to be changed.

test-suite/bugs/closed/bug_4287.v
```diff
@@ -104,15 +104,15 @@ Qed.
 End Hurkens.
 
 Polymorphic Record box (T : Type) := wrap {unwrap : T}.
-
+Set Printing Universes.
 (* Here we instantiate to Set *)
 
-Fail Definition down (x : Type) : Prop := box x.
+Definition down (x : Type) : Prop := box x.
```
JasonGross: This should fail, I think, at least with `Unset Universe Minimization ToProp ` or whatever.  Rigid universes should not be auto-minimized to `Prop`

test-suite/bugs/closed/bug_4301.v
```diff
@@ -9,5 +9,5 @@ Module Lower (X : Foo with Definition U := True : Type).
 End Lower.
 
 Module M : Foo.
-  Definition U := nat : Type@{i}.
+  Definition U := True : Type@{i}.
```
JasonGross: Why does this need to be changed?

test-suite/bugs/closed/bug_8951.v
```diff
@@ -3,7 +3,7 @@ Module Type T.
 End T.
 
 Module M.
-  Polymorphic Definition t@{i} := nat.
+  Polymorphic Definition t@{i} := True.
```
JasonGross: What's up with this change?

JasonGross (changes requested):

Janno:
I've been trying out this branch and I am reasonably sure that it would allow me to rip out at least one very annoying hack from Mtac2. And that's just some low-hanging fruit. I haven't even had time to see if we can remove our implementation of sort polymorphism with this branch.

I just want to say that I would very much appreciate the current stage of this PR landing soon (even without the whole template polymorphism business, although that would also be super awesome!).

gares:
Btw wasn't this pr, or a part of it, meant to solve the sounds issue of 8.9? Where are we with that?

maximedenes:
> Btw wasn't this pr, or a part of it, meant to solve the sounds issue of 8.9? Where are we with that?

I haven't made progress on it, but still hope to help @mattam82 with it.

gares:
My understanding of this PR, or better a part of it, is that it was supposed to fix the soundness bug that affects 8.9, but if we don't progress on this PR it will also affect 8.10 :-(

ppedrot:
Shouldn't we close this in favour of #9918?

ppedrot:
No answer from author, taking the liberty to close this PR. The various features introduced in this PR need to be reworked cleanly anyways.

PR #9434, opened 2019-01-29
Fix #9432: canonical structure and coercion accept universe binders.

SkySkimmer:
(when defining a new constant)

PR #9450, opened 2019-01-31
Fix #9432: canonical structure and coercion accept universe binders.

### LINKED ISSUE #9432: Canonical Structure command does not accept universe binders
#### Version

8.8.2

#### Operating system

Ubuntu

The `Canonical Structure` command does not accept universe binders, which is confusing. The following snippet demonstrates the issue.

```
Record box@{i} (T : Type@{i}) := Box { unbox : T }.
Canonical Structure box@{i} (T : Type@{i}) (x : T) := Box T x.
(* Toplevel input, characters 23-25: *)
(* > Canonical Structure box@{i} (T : Type@{i}) (x : T) := Box T x. *)
(* >                        ^^ *)
(* Error: Syntax error: '.' expected after [vernac:gallina_ext] (in [vernac_aux]). *)

```

gares:
This revives #9434

### DIFF
--- a/vernac/g_vernac.mlg
+++ b/vernac/g_vernac.mlg
@@ -693,13 +693,15 @@ GRAMMAR EXTEND Gram
           LIST1 [ v=strategy_level; "["; q=LIST1 smart_global; "]" -> { (v,q) } ] ->
             { VernacSetStrategy l }
       (* Canonical structure *)
-      | IDENT "Canonical"; IDENT "Structure"; qid = global ->
-          { VernacCanonical CAst.(make ~loc @@ AN qid) }
+      | IDENT "Canonical"; IDENT "Structure"; qid = global; ud = OPT [ u = OPT univ_decl; d = def_body -> { (u,d) } ] ->
+          { match ud with
+           | None ->
+             VernacCanonical CAst.(make ~loc @@ AN qid)
+           | Some (u,d) ->
+             let s = coerce_reference_to_id qid in
+             VernacDefinition ((NoDischarge,CanonicalStructure),((CAst.make (Name s)),u),d) }
       | IDENT "Canonical"; IDENT "Structure"; ntn = by_notation ->
           { VernacCanonical CAst.(make ~loc @@ ByNotation ntn) }
-      | IDENT "Canonical"; IDENT "Structure"; qid = global; u = OPT univ_decl; d = def_body ->
-          { let s = coerce_reference_to_id qid in
-          VernacDefinition ((NoDischarge,CanonicalStructure),((CAst.make (Name s)),u),d) }
 
       (* Coercions *)
       | IDENT "Coercion"; qid = global; u = OPT univ_decl; d = def_body ->
--- a/vernac/g_vernac.mlg
+++ b/vernac/g_vernac.mlg
@@ -693,14 +693,14 @@ GRAMMAR EXTEND Gram
           LIST1 [ v=strategy_level; "["; q=LIST1 smart_global; "]" -> { (v,q) } ] ->
             { VernacSetStrategy l }
       (* Canonical structure *)
-      | IDENT "Canonical"; IDENT "Structure"; qid = global; ud = OPT [ u = OPT univ_decl; d = def_body -> { (u,d) } ] ->
+      | IDENT "Canonical"; OPT [ IDENT "Structure" -> {()} ]; qid = global; ud = OPT [ u = OPT univ_decl; d = def_body -> { (u,d) } ] ->
           { match ud with
            | None ->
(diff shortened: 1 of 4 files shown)


vernac/g_vernac.mlg
```diff
@@ -654,13 +654,15 @@ GRAMMAR EXTEND Gram
...
+             let s = coerce_reference_to_id qid in
+             VernacDefinition ((NoDischarge,CanonicalStructure),((CAst.make (Name s)),u),d) }
       | IDENT "Canonical"; IDENT "Structure"; ntn = by_notation ->
           { VernacCanonical CAst.(make ~loc @@ ByNotation ntn) }
-      | IDENT "Canonical"; IDENT "Structure"; qid = global; u = OPT univ_decl; d = def_body ->
-          { let s = coerce_reference_to_id qid in
-          VernacDefinition ((NoDischarge,CanonicalStructure),((CAst.make (Name s)),u),d) }
 
       (* Coercions *)
       | IDENT "Coercion"; qid = global; u = OPT univ_decl; d = def_body ->
```
SkySkimmer: You probably need to do the same for Coercion too.

gares: The problem in you original commit is that  camlp5 is not able to factor line 661 with 657, dunno why, so I factored them by hand. For coercions there is only one line, so no factoring seems needed

gares: Feel free to push (more) tests on my branch

SkySkimmer: It doesn't need manual factoring with `IDENT "Coercion"; qid = global; ":"; s = class_rawexpr`?

gares: I don't think this command needs universe levels, it only declares a posteriori a coercion.

gares:
can someone take this PR?

ejgallego:
@maximedenes is secondary for `vernac`

vernac/g_vernac.mlg
SkySkimmer: >I don't think this command needs universe levels, it only declares a posteriori a coercion.

`Coercion foo := term.` defines foo so should allow universe binders.

gares:
Gaëtan is right this pr only fixes half of the issue. I'll try to improve it later today

ggonthier:
You should also port this to the SSReflect plugin grammar, to the rule that lets you drop `Structure` out of `Canonical Structure` (or take the opportunity to move this to the main Coq grammar?).

gares:
@ggonthier done.

The doc could be improved. @coq/doc-maintainers The point is: `Canonical Structure` is semantically wrong, since what you declare as `Canonical` is an instance (of a structure). So the shorted variant that was available in SSR, and that this PR moves to Coq proper, should be the preferred way to declare canonical instances. I'm not sure it is for this PR to change that.

@SkySkimmer `Coercion` already works fine IMO, I've added a little test.

Zimmi48 (approved):
Very cool to see some bits of SSReflect's Gallina extensions finally make it to the main grammar!
Furthermore, it looks like this extension was not even documented in the SSReflect chapter, so this also solves a documentation bug!

vernac/g_vernac.mlg
gares: Ok, I've added a test but there was no factoring needed since there were not 2 conflicting grammar rules for that.

maximedenes:
The failure is probably due to https://github.com/mit-plv/fiat/pull/19 not being merged yet.

vernac/g_vernac.mlg
```diff
@@ -693,18 +693,20 @@ GRAMMAR EXTEND Gram
           LIST1 [ v=strategy_level; "["; q=LIST1 smart_global; "]" -> { (v,q) } ] ->
             { VernacSetStrategy l }
       (* Canonical structure *)
-      | IDENT "Canonical"; IDENT "Structure"; qid = global ->
-          { VernacCanonical CAst.(make ~loc @@ AN qid) }
-      | IDENT "Canonical"; IDENT "Structure"; ntn = by_notation ->
+      | IDENT "Canonical"; OPT [ IDENT "Structure" -> {()} ]; qid = global; ud = OPT [ u = OPT univ_decl; d = def_body -> { (u,d) } ] ->
```
maximedenes: Why is `-> {()}` needed?

gares: Because `OPT IDENT "Structure "` does not parse and if you put `[` then it wants an `->`... Or at least this is my understanding of the errors I hot

maximedenes: Ok, thanks!

maximedenes (approved):

PR #9452, opened 2019-01-31
[proof] optimize proof always works on incomplete proofs

### LINKED ISSUE #9451: Anomaly "grounding a non evar-free term" during Optimize Proof
#### Version

bad 4b006833d6f866a33024e674d300f74592d24622, good 8.9.0

#### Operating system

Arch Linux, up to date as of yesterday.

```coq
Goal False. evar (x : True). Optimize Proof. Abort. 
(* Error: Anomaly "grounding a non evar-free term" *)
```

```coq
Goal False.
cut True.
assert False.
evar (x : True).
let v := open_constr:(_) in idtac. all: exfalso; clear.
Optimize Proof.
(* Error: Anomaly "grounding a non evar-free term" *)
Abort All.
```

Original finding: adding `Optimize Proof` on this bedrock2 line https://github.com/mit-plv/bedrock2/commit/48e2314623e8695df7f0bf2e390ce224baf7a161#diff-5c7e120c9281033af09d97fd489aa26aR74, where the line after which Optimize Proof starts anomalying is a call to [seprewrite](https://github.com/mit-plv/bedrock2/blob/master/bedrock2/src/Map/SeparationLogic.v#L228).

gares:
Fix #9451

### DIFF
--- a/engine/proofview.ml
+++ b/engine/proofview.ml
@@ -39,7 +39,7 @@ let proofview p =
 
 let compact el ({ solution } as pv) =
   let nf c = Evarutil.nf_evar solution c in
-  let nf0 c = EConstr.(to_constr solution (of_constr c)) in
+  let nf0 c = EConstr.(to_constr ~abort_on_undefined_evars:false solution (of_constr c)) in
   let size = Evd.fold (fun _ _ i -> i+1) solution 0 in
   let new_el = List.map (fun (t,ty) -> nf t, nf ty) el in
   let pruned_solution = Evd.drop_all_defined solution in
--- a/test-suite/bugs/closed/bug_9451.v
+++ b/test-suite/bugs/closed/bug_9451.v
@@ -1,0 +1,8 @@
+Goal False.
+cut True.
+assert False.
+evar (x : True).
+let v := open_constr:(_) in idtac. all: exfalso; clear.
+Optimize Proof.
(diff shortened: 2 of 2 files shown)


engine/proofview.ml
```diff
@@ -39,7 +39,7 @@ let proofview p =
 
 let compact el ({ solution } as pv) =
   let nf c = Evarutil.nf_evar solution c in
-  let nf0 c = EConstr.(to_constr solution (of_constr c)) in
```
ejgallego: Why is this not using the standard `nf` functions?

JasonGross (dismissed):
Add a file to the test suite to ensure this doesn't happen again?

engine/proofview.ml
gares: I've no idea, when I wrote this code EConstr did not even exist, not to talk about the anomaly about ground terms...

gares: It comes from c8c8ccdaaffefdbd3d78c844552a08bcb7b4f915

ppedrot (approved):
LGTM. This is only side-stepping the problem though.

engine/proofview.ml
SkySkimmer: nf_evars on constr is the awkwardly-named nf_evars_universes

PR #9454, opened 2019-01-31
Fix off-by-one error in nat syntax warnings

### LINKED ISSUE #9453: Check 5000 warns when it did not used to
#### Version

8.9.0

This code works on 8.8, but fails on 8.9.  The `5000` in the numeral notation for `nat` should be `5001`.

```coq
Set Warnings Append "+large-nat,+abstract-large-number".
Fail Check 5001.
Check 5000.
(* Error:
To avoid stack overflow, large numbers in nat are interpreted as applications of
Nat.of_uint. *)
```

JasonGross:
Fixes #9453

**Kind:** bug fix

- [ ] Entry added in CHANGES.md. (Do we need one?)


ppedrot (approved):
Looks like a very specific problem, but well, that's indeed a regression and the patch is fine.

PR #9456, opened 2019-02-01
The lowest universe level is 1.

akr:
Cic description doesn't describe the lowest universe level clearly.

- "Type(i) for any integer i" seems Type(-1) is possible
- "S = {Prop,Set,Type(i)| i ∈ ℕ }" depends on the definition of "ℕ"
  which is not described.
  It is well known that there are two definitions that ℕ includes 0 or not.
  In Coq, it is natural that ℕ  includes 0 because the inductive type
  nat includes 0.
- "Prop:Type(1), Set:Type(1)" suggests the lowest level is 1.
- AX-Prop and AX-Set describes Prop:Type(1) and Set:Type(1).
  So, Prop and Set are not belongs to Type(0).

Also, CPDT describes that "The type of Set is Type(0)".
http://adam.chlipala.net/cpdt/html/Universes.html

I think the lowest universe level is 1 because AX-Prop and AX-Set.

I'm not certain to fix this problem but
my idea to fix this problem is changing
"Type(i) for any integer i" to
"Type(i) for any integer i ≥ 1".

**Kind:** documentation.

### DIFF
--- a/doc/sphinx/language/cic.rst
+++ b/doc/sphinx/language/cic.rst
@@ -51,7 +51,7 @@ function types over these data types.
 Consequently they also have a type. Because assuming simply that :math:`\Set`
 has type :math:`\Set` leads to an inconsistent theory :cite:`Coq86`, the language of
 |Cic| has infinitely many sorts. There are, in addition to :math:`\Set` and :math:`\Prop`
-a hierarchy of universes :math:`\Type(i)` for any integer :math:`i`.
+a hierarchy of universes :math:`\Type(i)` for any integer :math:`i ≥ 1`.
 
 Like :math:`\Set`, all of the sorts :math:`\Type(i)` contain small sets such as
 booleans, natural numbers, as well as products, subsets and function


akr:
I found old reference manual mention this.
https://github.com/coq/coq/commit/0e1befcf9b15237220dfe2761484f361a963b952 : 

```
In the Reference
Manual of versions of Coq prior to 8.4, the level of {\Type} typing
{\Prop} and {\Set} was numbered $0$. From Coq 8.4, it started to be
numbered $1$ so as to be able to leave room for re-interpreting {\Set} in the hierarchy as {\Type$(0)$}.
```

SkySkimmer:
Set is a fancy name for Type0, unless you have -impredicative-set in which case it has the special property of being impredicative.
I think it would work to say that Set is Type0 in the part introducing Typei then in the -impredicative-set section say that it makes Set special.

akr:
I didn't expect such bigger rewrite.
My intent is minimum modification to clarify the lowest universe level for Type (without Set).
Now I'm certain that it is 1.
So the proposed patch itself is correct, I think.

Apart from that, I tried to modify the manual to explain Set = Type(0):
https://github.com/akr/coq/commit/68db1e22c1952f0e9c59691a32b95191ccd4f36e
It removes Ax-Set and bit simplify Prod-Set and the subtyping rule.
But I'm not sure that this modification is correct.

For example, I don't understand template polymorphism well.
I'm not sure that "Type hierarchy" of the sentence 
"In practice, the rule Ind-Family is used by Coq only when all the inductive types of the inductive definition are declared with an arity whose sort is in the Type hierarchy."
includes Set or not.
(I'm also not sure that it includes Prop or not.)

Zimmi48:
>For example, I don't understand template polymorphism well.

It seems that template polymorphism is going to be redesigned in 8.10 anyways, so let's merge this PR for now and hopefully the documentation of universes will be improved much more soon.

Zimmi48 (approved):

PR #9457, opened 2019-02-01
Correct W-Ind in Cic description of the reference manual.

akr:
Remove the 2nd premise "(E[Γ_P ] ⊢ A_j : s_j)_(j=1...k)".
Also, "Γ" in "c_i ∉ Γ ∪ E" is changed to "Γ_I".

"E[Γ_P ] ⊢ A_j : s_j" contradicts with a side condition
"A_j is an arity of sort s_j".
The latter means that A_j is ∀ x1:T1, ... ∀ xm:Tm, s_j.
So, "(∀ x1:T1, ..., s_j) : s_j)" is required.
Using Prod-{Prop,Set,Type}, it needs "s_j : s_j" which cannot be proved.

This problem is introduced at 2018-07-22:
https://github.com/coq/coq/commit/f25c1d252ad61b4dc4321e3a11f33b1e6d4e3dff
Before that, the premise describes "A_j : s'_j".
And "s'_j" is not explained anywhere.

I think "A_j : s'_j" describes that A_j is a type (a value of a sort).
Thus, "A_j is an arity of sort s_j" imply "A_j : s'_j".

So, I removed the 2nd premise.

Also, a side condition describes "c_i ∉ Γ ∪ E" but "Γ" is not explained.
I think that Γ should be Γ_I.

Γ should include Γ_I because a constructor, c_i, and inductive types are
both defined in global environment and it cannot accept duplicate name.
Actually, such definition fails:

```
Fail Inductive X := X : X.
(* The following names are used both as type names and constructor names: X.*)
```

Γ doesn't include Γ_P because parameters are not defined in global
environment.
Actually, a parameter can be same name as a constructor name:

```
Inductive I (X : nat) := X : I X.
Check X. (* X : forall X : nat, I X *)
```

**Kind:** documentation

### DIFF
--- a/doc/sphinx/language/cic.rst
+++ b/doc/sphinx/language/cic.rst
@@ -965,7 +965,7 @@ such that :math:`Γ_I` is :math:`[I_1 :∀ Γ_P ,A_1 ;~…;~I_k :∀ Γ_P ,A_k]`
    \WFE{Γ_P}
    (E[Γ_I ;Γ_P ] ⊢ C_i : s_{q_i} )_{i=1… n}
    ------------------------------------------
-   \WF{E;~\ind{p}{Γ_I}{Γ_C}}{Γ}
+   \WF{E;~\ind{p}{Γ_I}{Γ_C}}{}
    
 
 provided that the following side conditions hold:
@@ -975,7 +975,7 @@ provided that the following side conditions hold:
       context of parameters,
     + for :math:`j=1… k` we have that :math:`A_j` is an arity of sort :math:`s_j` and :math:`I_j ∉ E`,
     + for :math:`i=1… n` we have that :math:`C_i` is a type of constructor of :math:`I_{q_i}` which
-      satisfies the positivity condition for :math:`I_1 … I_k` and :math:`c_i ∉ Γ_I ∪ E`.
+      satisfies the positivity condition for :math:`I_1 … I_k` and :math:`c_i ∉  E`.
 
 One can remark that there is a constraint between the sort of the
 arity of the inductive type and the sort of the type of its
--- a/doc/sphinx/language/cic.rst
+++ b/doc/sphinx/language/cic.rst
@@ -963,7 +963,6 @@ such that :math:`Γ_I` is :math:`[I_1 :∀ Γ_P ,A_1 ;~…;~I_k :∀ Γ_P ,A_k]`
 .. inference:: W-Ind
 
    \WFE{Γ_P}
-   (E[Γ_P ] ⊢ A_j : s_j )_{j=1… k}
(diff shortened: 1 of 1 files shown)


Zimmi48:
ping @herbelin (or someone else that can review this)

Zimmi48:
I have a problem here because the CIC chapter is among the ones that I know the less well and this is a change to the actual content. So ideally, I would really prefer if someone better fitted could review this (if it is not @herbelin, maybe it can be @mattam82 or @ppedrot or @SkySkimmer?). But on the other hand, the contributor here has made a number of valuable contributions so far and I hate to see this PR pending for so long, so if there is no reaction, I will merge this without a real review.

SkySkimmer:
>This problem is introduced at 2018-07-22:
f25c1d2
Before that, the premise describes "A_j : s'_j".
And "s'_j" is not explained anywhere.

>I think "A_j : s'_j" describes that A_j is a type (a value of a sort).
Thus, "A_j is an arity of sort s_j" imply "A_j : s'_j".

>So, I removed the 2nd premise.

I guess whoever wrote it was worried that "is an arity" wouldn't clearly imply that it should be well-typed (because in the code you're supposed to know that something is well-typed before calling is_arity on it). Then it should be OK to remove the premise.

>Also, a side condition describes "c_i ∉ Γ ∪ E" but "Γ" is not explained.
I think that Γ should be Γ_I.

>Γ should include Γ_I because a constructor, c_i, and inductive types are
both defined in global environment and it cannot accept duplicate name.

Isn't that already handled by the side condition "all of Ij and ci are distinct names for j=1…k and i=1…n"?
Also, Γ is in the rule's conclusion `WF(E; Ind [p](ΓI := ΓC))[Γ]` which seems strange as it would allow arbitrary nonsense local contexts after declaring an inductive.

Meanwhile `Inductive X (X:nat) := C.` fails with a stupid error: 
>The expression "X0" of type "nat" cannot be applied to the term "X0" : "nat"

akr:
> Isn't that already handled by the side condition "all of Ij and ci are distinct names for j=1…k and i=1…n"?
> Also, Γ is in the rule's conclusion `WF(E; Ind [p](ΓI := ΓC))[Γ]` which seems strange as it would allow arbitrary nonsense local contexts after declaring an inductive.

Oh.  I overlooked that Γ is used in the conclusion.
I agree Γ in the conclusion is nonsense.
I think the conclusion should use empty local contexts as W-Global-Assum and W-Global-Def.

Now, I guess the intent of Γ of "c_i ∉ Γ ∪ E" is that c_i is not in the local context of the conclusion.
But the local context should be empty/
So, I think it should be changed to "c_i ∉ E".

I updated the branch.

SkySkimmer (approved):
Fine by me

PR #9133, opened 2018-12-04
Move README-V1-V5 to credits chapter

Zimmi48:
**Kind:** documentation

I have been bothered by the spread of documentation and especially release notes and credits in many places for a while. So this is a first step towards a consolidation.

I'm pretty sensitive to not rewording these parts of the doc too much since they are dated and have clearly identified authors. That's why this consolidation does not try to get rid of all duplication, but at the same time it tries to ensure that everything can be found in the same place, and the order is chronological as much as possible. Some notes were added to make the context more explicit.

I'm pretty happy with the result, it seems to make sense and be consistent. Have a look at the rendered version. If I may request help, it would be on the bibliography. I added articles to the `.bib` file by searching the title in Google Scholar and copy-pasting the bibtex that I could find. People who like TeX better than me might be able to improve the data for these new entries.

Issue #7075, opened 2018-03-26
Structure for CHANGES


ejgallego:
Previously, Coq used to ship a `COMPATIBILITY` file whose intent was to collect all the changes that could be a source of user-level incompatibility between versions.

Now, these changes are buried in the more detailed `CHANGES` file and it is indeed hard to spot what kind of changes may need immediate source-action.

IMHO, we should either resurrect the `COMPATIBILITY` file, or provide `CHANGES` with a more principled structure, maybe à la OCaml.

ejgallego:
@maximedenes points out that there is a `needs: CHANGES` tag. However, I am not sure we don't want a more refined approach so `needs: CHANGES` would be used for most PRs and `needs: COMPAT` for changes that introduce user-level observable changes, so they deserve even more attention.

Zimmi48:
FTR this file has been buried because anyways it was not consistently updated and CHANGES was updated instead. We could indeed structure CHANGES more but, contrary to OCaml, the changelog between two versions of Coq is not so long and reading it entirely is not such a difficult task.

So: not opposed to giving more structure to CHANGES if you have a proposal, but I wouldn't tend to worry about it unless some users complain.

@tchajed You are a user of Coq master branch and you indicated in the past that you rely on CHANGES to spot compatibility problems. What is your opinion on this question?

tchajed:
I agree with your assessment that the Coq CHANGES file is short enough to read.

I think it’s easy to get carried away with lots of structuring. I would recommend just three categories:

1) internal: API changes that only impact plugin developers and other Coq developers.
2) new features: things that shouldn’t impact any existing code but are good to know about
3) user-facing: any change that might affect users’ code

If in doubt, I would classify a change higher (since as a reader I might skip category 1 changes, for example). Drastic performance changes belong in category 3 as well, just in case. For simplicity I think bug fixes can be lumped into category 3 as well.

Zimmi48:
Note that category 1 is already well separated as it belongs to `dev/doc/changes.md`

Zimmi48:
For future reference, this automatic changelog generator was mentioned on GitHub: https://github.com/skywinder/Github-Changelog-Generator
We could consider using such a tool and a good language of labels to automatically generate a detailed and structured changelog. However, it has quite low priority to my eyes.

tchajed:
[...]

PR #9460, opened 2019-02-01
Split credits in two parts, one in reverse chronological order.

Zimmi48:
**Kind:** documentation

This is a proposal that is based on #9133 but goes further by applying @gares's proposal of using reverse chronological order for the recent credits, while keeping chronological order for older ones. The only way to do this that made sense to me was to split the credits chapter in two parts:

- the new "Early history of Coq" chapter is about versions 1-7 and is in chronological order;
- the new "Recent changes" chapter is in reverse chronological order as suggested by @gares (but also by @cpitclaudel).

As said before, this is on top of #9133, so this does not prevent merging this other one first. But we may alternatively want to merge only this new PR. Note that this gave me ideas on how to address #7058,  #7075, #8180, #9103 but I will open a separate (dependent) PR about this.

Note there is not much point in reviewing these changes as a block diff. Do it commit by commit or by looking at the rendered manual: https://zimmi48.gitlab.io/-/coq/-/jobs/154887049/artifacts/_install_ci/share/doc/coq/sphinx/html/index.html.


Zimmi48:
Given the lack of reaction, I'll close this in favor of an upcoming larger draft PR.

PR #9475, opened 2019-02-05
Automatic deployment of the user manual to GH-Pages

vbgl:
We can take advantage of Gitlab-CI to automatically deploy the current version of the user manual.

This PR adds a `deploy` stage with only one job. This job commits to the https://github.com/coq/manual/ repository the manual as built in the previous stage. For technical reasons, directories whose name starts with an underscore need to be renamed.

This automatic deployment only happen when the private key (allowed to write to that repository) is present.

Said key is dedicated to this task and currently stored on a private CI/CD mirror of this repository, and only made available to *protected* branches (e.g., `master`, as opposed to pull requests).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,6 +4,7 @@ stages:
   - docker
   - build
   - test
+  - deploy
 
 # some default values
 variables:
@@ -346,6 +347,35 @@ doc:refman:dune:
     <<: *dune-ci-template-artifacts
     paths:
       - _build/default/doc/sphinx_build/html
+
+doc:refman:deploy:
+  stage: deploy
+  environment:
+    name: deployment
+    url: https://coq.github.io/
+  only:
+    variables:
+      - $DOCUMENTATION_DEPLOY_KEY
+  dependencies:
+    - doc:refman:dune
+  before_script:
+    - which ssh-agent || ( apt-get update -y && apt-get install openssh-client -y )
+    - eval $(ssh-agent -s)
+    - echo "$DOCUMENTATION_DEPLOY_KEY" | tr -d '\r' | ssh-add - > /dev/null
+    - mkdir -p ~/.ssh
+    - chmod 700 ~/.ssh
+    - ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
+    - git config --global user.name "coqbot"
+    - git config --global user.email "coqbot@users.noreply.github.com"
+  script:
+    - git clone git@github.com:coq/doc.git _deploy
+    - rm -rf _deploy/$CI_COMMIT_REF_NAME/refman # NB: won’t work if branch name has ‘/’ in it
+    - mkdir -p _deploy/$CI_COMMIT_REF_NAME
+    - cp -rv _build/default/doc/sphinx_build/html _deploy/$CI_COMMIT_REF_NAME/refman
+    - cd _deploy/$CI_COMMIT_REF_NAME/refman
+    - git add .
+    - git commit -m "User manual of branch “$CI_COMMIT_REF_NAME” at $CI_COMMIT_SHORT_SHA"
+    - git push # TODO: rebase and retry on failure
 
 doc:ml-api:odoc:
   <<: *dune-ci-template


SkySkimmer (approved):
Cool

ejgallego (dismissed):
> For technical reasons, directories whose name starts with an underscore need to be renamed.

Indeed, if that's the case why not to fix it upstream instead of patching it here? I also don't like the underscores so much. cc: @coq/doc-maintainers

vbgl:
I think we cannot tell sphinx to write static and image files to specific directories. Please correct me if I’m wrong.

ejgallego:
> I think we cannot tell sphinx to write static and image files to specific directories. Please correct me if I’m wrong.

I am not expert on sphinx but:
```
html_static_path = ['_static']
```

vbgl:
This is to specify where are your inputs.

The names of output directories are hard-coded in a few places, e.g., https://github.com/sphinx-doc/sphinx/blob/bc805fb8bc0be321b88e0a07710aa69f6d14cffb/sphinx/builders/html.py#L814

vbgl:
Thanks for the hint, Emilio. I’ll give it a try.

.gitlab-ci.yml
```diff
@@ -347,6 +348,35 @@ doc:refman:dune:
...
+    - doc:refman:dune
+  before_script:
+    - 'which ssh-agent || ( apt-get update -y && apt-get install openssh-client -y )'
+    - eval $(ssh-agent -s)
+    - echo "$MANUAL_DEPLOY_KEY" | tr -d '\r' | ssh-add - > /dev/null
+    - mkdir -p ~/.ssh
+    - chmod 700 ~/.ssh
+    - ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
+    - git config --global user.name "Gitlab-CI"
+    - git config --global user.email "abc@example.com"
```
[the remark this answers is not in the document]
Zimmi48: @vbgl I suggest using "coqbot" as the name and "coqbot@users.noreply.github.com" as the e-mail.

Zimmi48:
>Said key is dedicated to this task and currently stored on a private CI/CD mirror of this repository, and only made available to _protected_ branches (e.g., `master`, as opposed to pull requests).

You will soon discover that this does not work because private repositories have a very low build time monthly limit. You should use a public fork with private logs, or just find a way to use the main repo and make this key accessible only to protected branches.

Zimmi48:
In fact, I suggest that you do the same thing for `REFMAN_DEPLOY_KEY` as we did for `CACHIX_SIGNING_KEY`: you use a protected variable restricted to a deployment environment. We might consider another protection scenario in the future, but there is no reason why the two deploy jobs should protect their private information differently.

Zimmi48:
I see that you are intending to store the various versions of the manual in various directories (on per branch). Would it be an option to push the various versions to distinct branches instead (the branching structure on https://github.com/coq/manual then reflecting the one on https://github.com/coq/coq)?

vbgl:
> I see that you are intending to store the various versions of the manual in various directories (on per branch). Would it be an option to push the various versions to distinct branches instead (the branching structure on https://github.com/coq/manual then reflecting the one on https://github.com/coq/coq)?

The fact that we are writing to a git directory is a technical accident.

We are not building a git repository but a website to be served through
HTTP. I don’t understand what “branches” mean in this context.

Zimmi48:
Indeed, the various branches wouldn't be published to GitHub pages at once. Sorry for the noise.

BTW, I suppose that the new URL would be then https://coq.inria.fr/manual/branch. Any reason to deviate from the traditional "refman" in the URL here?

Another remark: since it seems that URLs must change anyway, the choice of the name "manual" or "refman" for the deploy repo means that if you want to automatically deploy the api doc or the standard library doc, you will have to create other deploy repositories. An alternative would be to create a deploy repository "doc" and use it for everything. Then the URLs would be:

https://coq.inria.fr/doc/master/refman
https://coq.inria.fr/doc/master/stdlib
https://coq.inria.fr/doc/master/api

https://coq.inria.fr/doc/8.9.0/refman
https://coq.inria.fr/doc/8.9.0/stdlib
https://coq.inria.fr/doc/8.9.0/api

etc., or:

https://coq.inria.fr/doc/refman/master
https://coq.inria.fr/doc/stdlib/master
https://coq.inria.fr/doc/api/master

https://coq.inria.fr/doc/refman/8.9.0
https://coq.inria.fr/doc/stdlib/8.9.0
https://coq.inria.fr/doc/api/8.9.0

depending on the preferred tree structure (personally I prefer the first one).

.gitlab-ci.yml
```diff
@@ -347,6 +348,35 @@ doc:refman:dune:
...
+  stage: deploy
+  only:
+    variables:
+      - $MANUAL_DEPLOY_KEY
+  dependencies:
+    - doc:refman:dune
+  before_script:
+    - 'which ssh-agent || ( apt-get update -y && apt-get install openssh-client -y )'
+    - eval $(ssh-agent -s)
+    - echo "$MANUAL_DEPLOY_KEY" | tr -d '\r' | ssh-add - > /dev/null
```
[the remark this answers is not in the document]
vbgl: Don’t know.

Without the `tr`, I got the very enlightening:
> Error loading key "(stdin)": invalid format

vbgl:
The `.nojekyll` hack works. Thanks! There is no need to tamper with sphinx output, then.
Preview: https://coq.github.io/manual/deploy-master-documentation/

I agree that deployment to cachix and to gh-pages are very similar. Let’s do the same.

It seems to be a good idea to have a single repo gathering automatically deployed documentation (api, user manual, stdlib, something else?). I’ll rename the `manual` repository then.

SkySkimmer:
@vbgl ready for merge?

.gitlab-ci.yml
```diff
@@ -347,6 +348,35 @@ doc:refman:dune:
...
+    - echo "$DOCUMENTATION_DEPLOY_KEY" | tr -d '\r' | ssh-add - > /dev/null
+    - mkdir -p ~/.ssh
+    - chmod 700 ~/.ssh
+    - ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
+    - git config --global user.name "coqbot"
+    - git config --global user.email "coqbot@users.noreply.github.com"
+  script:
+    - git clone git@github.com:coq/doc.git
+    - rm -rf doc/$CI_COMMIT_REF_NAME/refman # NB: won’t work if branch name has ‘/’ in it
+    - mkdir -p doc/$CI_COMMIT_REF_NAME
```
Zimmi48: This command does not look future proof.

SkySkimmer: >This command does not look future proof.

How so?

Zimmi48: Sorry for the noise. I misread that in `rmdir`. I'm too tired.

vbgl:
> $ git clone git@github.com:coq/doc.git
fatal: destination path 'doc' already exists and is not an empty directory.

Let me fix this…

vbgl:
Alright. This seems now to be ready!
Preview: https://coq.github.io/doc/deploy-master-documentation/

Zimmi48:
> Preview: https://coq.github.io/doc/deploy-master-documentation/

Why does your preview use this URL instead of https://coq.github.io/doc/master/refman/ like the code in this PR suggests?

vbgl:
Because I shouldn’t write messages so early in the morning…
Here is the URL I meant: https://coq.github.io/doc/deploy-master-documentation/refman/

Zimmi48 (approved):

Issue #9466, opened 2019-02-04
Destruct fails when there are lets in the parameters of an inductive


SkySkimmer:
~~~coq
Inductive foo (x:=nat) : Type := C : foo.

Goal  foo -> True.
Proof.
  intros H. destruct H.
  (* Error: decompose_prod_n_assum: not enough assumptions *)
Abort.
~~~
cc @ppedrot

ppedrot:
FTR I discovered that changing the representation of constructor types in the kernel, and I am about to submit a PR fixing it.

SkySkimmer:
Fixed in 8.18

PR #9476, opened 2019-02-05
Constructor type information uses the expanded form.

ppedrot:
It used to simply remember the normal form of the type of the constructor. This is somewhat problematic as this is ambiguous in presence of let-bindings. Rather, we store this data in a fully expanded way, relying on rel_contexts.

Probably fixes a crapload of bugs with inductive types containing let-bindings, but it seems that not many were reported in the bugtracker.

Overlays:
- https://github.com/QuickChick/QuickChick/pull/145
- https://github.com/mattam82/Coq-Equations/pull/182

### DIFF
--- a/kernel/declareops.ml
+++ b/kernel/declareops.ml
@@ -214,7 +214,7 @@ let subst_mind_packet sub mbp =
     mind_consnrealdecls = mbp.mind_consnrealdecls;
     mind_consnrealargs = mbp.mind_consnrealargs;
     mind_typename = mbp.mind_typename;
-    mind_nf_lc = Array.Smart.map (subst_mps sub) mbp.mind_nf_lc;
+    mind_nf_lc = Array.Smart.map (fun (ctx, c) -> Context.Rel.map (subst_mps sub) ctx, subst_mps sub c) mbp.mind_nf_lc;
     mind_arity_ctxt = subst_rel_context sub mbp.mind_arity_ctxt;
     mind_arity = subst_ind_arity sub mbp.mind_arity;
     mind_user_lc = Array.Smart.map (subst_mps sub) mbp.mind_user_lc;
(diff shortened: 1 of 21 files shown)


SkySkimmer (approved):

gares:
Can you explain the advantage? I see many `Term.it_mkProd_or_LetIn` , so it seems that at least some code prefers the other form

SkySkimmer:
Once we have contexts in the representation of branches there may be less mkprod.

kernel/indtypes.ml
```diff
@@ -420,7 +420,9 @@ let compute_projections (kn, i as ind) mib =
   | Cumulative_ind acumi -> Univ.make_abstract_instance (Univ.ACumulativityInfo.univ_context acumi)
   in
   let subst = List.init mib.mind_ntypes (fun i -> mkIndU ((kn, mib.mind_ntypes - i - 1), u)) in
-  let rctx, _ = decompose_prod_assum (substl subst pkt.mind_nf_lc.(0)) in
+  let (ctx, cty) = pkt.mind_nf_lc.(0) in
+  let cty = it_mkProd_or_LetIn cty ctx in
+  let rctx, _ = decompose_prod_assum (substl subst cty) in
   let ctx, paramslet = List.chop pkt.mind_consnrealdecls.(0) rctx in
```
SkySkimmer: Interesting dance

ppedrot:
Also, this is a preliminary PR before replacing the representation of case nodes in the kernel. I stumbled upon the current issue while implementing the former.

gares (approved):

SkySkimmer:
Does this fix https://github.com/coq/coq/issues/9466 ?

ppedrot:
Unluckily, not, the bug is hit before reaching this (otherwise also buggy) part. I will write another patch for that one.

ppedrot:
AFAICT this is ready.

ppedrot:
@SkySkimmer do you want to assign this to yourself?

kernel/declareops.ml
```diff
@@ -299,9 +299,8 @@ let hcons_ind_arity =
 
 let hcons_mind_packet oib =
   let user = Array.Smart.map Constr.hcons oib.mind_user_lc in
-  let nf = Array.Smart.map Constr.hcons oib.mind_nf_lc in
-  (* Special optim : merge [mind_user_lc] and [mind_nf_lc] if possible *)
-  let nf = if Array.equal (==) user nf then user else nf in
+  let map (ctx, c) = Context.Rel.map Constr.hcons ctx, Constr.hcons c in
+  let nf = Array.Smart.map map oib.mind_nf_lc in
```
SkySkimmer: No point using smart map when you always re-pair is there?

ppedrot: Cannot hurt though... This is future-proof code.

PR #9478, opened 2019-02-05
Remove the comment fields of locations.

ppedrot:
They didn't seem to be used at all.

### DIFF
--- a/gramlib/ploc.ml
+++ b/gramlib/ploc.ml
@@ -6,17 +6,16 @@ open Loc
 
 let make_unlined (bp, ep) =
   {fname = InFile ""; line_nb = 1; bol_pos = 0; line_nb_last = -1; bol_pos_last = 0;
-   bp = bp; ep = ep; comm = ""; ecomm = ""}
+   bp = bp; ep = ep; }
 
 let dummy =
   {fname = InFile ""; line_nb = 1; bol_pos = 0; line_nb_last = -1; bol_pos_last = 0;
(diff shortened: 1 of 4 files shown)


ejgallego (commented):
Well, actually this can be useful in the future as it seems more sensible to attach comments to locations instead of the current solution with `CLexer.comments` which is quite fragile to relocation.

OCaml does something like that and they are able to handle reformatting of files without losing comments.

Thus I dunno, maybe we should go in the other direction and remove `CLexer.comments` instead? @herbelin ?

ppedrot:
I'm fine with closing this PR if we need a status quo, that just looked like dead code to me.

ejgallego:
> I'm fine with closing this PR if we need a status quo, that just looked like dead code to me.

Let's indeed wait for @herbelin feedback to see if we want to use this field or not.

herbelin:
I was not aware of these `comm` and `ecomm` fields. I have no idea at once whether this can be useful to reimplement `CLexer.comments`. Both merging (and possibly reintroducing it to reimplement `CLexer.comments` later on) or the status quo are ok to me.

ejgallego:
Ok, thanks Hugo. I suggest you let me have a look at the feasibility of reimplementing this (I want to look at how can ocaml do such a good job) and then I'll either close or merge.

ejgallego (approved):
Well let's merge this for now, as the refactoring is not going to happen soon; it should be easy to resurrect.

PR #9489, opened 2019-02-06
[ssr] avoid HO unification to perform truncation analysy in elim

gares:
This is a possible fix for math-comp/math-comp#279

### DIFF
--- a/plugins/ssr/ssrelim.ml
+++ b/plugins/ssr/ssrelim.ml
@@ -253,31 +253,49 @@ let ssrelim ?(is_case=false) deps what ?elim eqid elim_intro_tac =
   (* Here we try to understand if the main pattern/term the user gave is
    * the first pattern to be matched (i.e. if elimty ends in P t1 .. tn,
    * weather tn is the t the user wrote in 'elim: t' *)
-  let c_is_head_p, gl = match cty with
+  let c_is_head_p, gl =
+    match cty with
     | None -> true, gl  (* The user wrote elim: _ *)
     | Some (c, c_ty, _) ->
-    let res = 
-      (* we try to see if c unifies with the last arg of elim *)
-      if elim_is_dep then None else
-      let arg = List.assoc (n_elim_args - 1) elim_args in
-      let gl, arg_ty = pfe_type_of gl arg in
-      match saturate_until gl c c_ty (fun c c_ty gl ->
-        pf_unify_HO (pf_unify_HO gl c_ty arg_ty) arg c) with
-      | Some (c, _, _, gl) -> Some (false, gl)
-      | None -> None in
-    match res with
-    | Some x -> x
-    | None ->
-      (* we try to see if c unifies with the last inferred pattern *)
-      let inf_arg = List.hd inf_deps_r in
-      let gl, inf_arg_ty = pfe_type_of gl inf_arg in
-      match saturate_until gl c c_ty (fun _ c_ty gl ->
-              pf_unify_HO gl c_ty inf_arg_ty) with
-      | Some (c, _, _,gl) -> true, gl
-      | None ->
-        errorstrm Pp.(str"Unable to apply the eliminator to the term"++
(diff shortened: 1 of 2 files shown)


maximedenes (commented):
Can you add a comment in the code to understand the conditional?

gares:
> understand the conditional?

Let's see if it works as I expect ;-)

If I'm right, `is_rec` is misnamed. It should be "last prod is a forall (not an arrow)" and this is what I use if for in this PR. Elsewhere in the code I've the impression it is used to govern how to handle equation generation...

I'll wait for CI to tell me if I'm likely to be right, and in that case I'll rename it.

gares:
I was wrong, the new patch tries things in a different order... let's see

gares:
Ok, this time I got it right!

I'll add some comments next week

gares:
I did some code cleanup, but it is mostly unchanged but for the extra test for HO unif problems

gares:
@maximedenes please take this one

gares:
ping

gares:
ping

gares:
@maximedenes what do you want me to explain? (note that, quite surprisingly, the code has comments)

test-suite/ssr/elim_noquant.v
```diff
@@ -0,0 +1,29 @@
...
+Lemma test (aT rT : Type)
+           (pmap : (aT -> option rT) -> list aT -> list rT)
+           (perm_eq : list rT -> list rT -> Prop)
+           (f : aT -> option rT)
+           (g : rT -> aT)
+           (s t : list aT)
+           (E : forall T : list aT -> Type,
+               (forall s1 s2 s3 : list aT,
+                  T (s1 ++ s2 ++ s3) -> T (s2 ++ s1 ++ s3)) ->
+                 T s -> T t) :
```
gares: Here the type of the last argument is `?T s`, and this unifies with pretty much anything, including the type of the argument of `elim` that is `t` of type `list aT`. Since `?T` is the predicate of the eliminator, we can't go much far if we instantiate it to `fun _ => list aT`.

Indeed it is the job of synthesizing `?T` by carving out of the goal copied of `t`. It is not the code that tries to understand if the argument of `elim` stands for the last pattern (the last argument of of the conclusion of the eliminator) or the last argument of the eliminator itself (the one of type `?T s`). The bit of IA here tries to put `t` in both places, by checking its type matches. We don't try if the unification is too unconstrained.

gares: @maximedenes give that we could not meet this week I wrote this text. Does it shed some light?

gares:
ping

plugins/ssr/ssrelim.ml
```diff
@@ -253,31 +253,48 @@ let ssrelim ?(is_case=false) deps what ?elim eqid elim_intro_tac =
...
+            spc()++pr_econstr_env env (project gl) c++spc())
+        | x :: rest ->
+          match x () with
+          | None -> first rest
+          | Some (b,gl) -> b, gl
+      in
+      (* unify two terms if their heads are not applied unif variables, eg
+       * a HO unif problem *)
+      let pf_unify_HO_rigid gl a b =
+        let is_HO_unif x = match EConstr.kind (project gl) x with
```
maximedenes: `is_HO_unif` is a bit confusing, could we find a better name?

test-suite/ssr/elim_noquant.v
maximedenes: Ok, I get the picture now. The implem looks ok modulo naming.

I can't really comment on the heuristics, though. Maybe @ggonthier has an opinion?

plugins/ssr/ssrelim.ml
gares: Should be better now.

gares:
should be ready to go

maximedenes:
Ok, let's merge, we'll see later if the heuristics can be improved.

PR #9496, opened 2019-02-06
Avoid eqn when generating name in induction_gen.

### LINKED ISSUE #9494: destruct with eqn: can fail with Cannot create self-referring hypothesis

The following code

```
Set Mangle Names.

Lemma foo(a: nat): True.
Proof.
  let N := fresh in destruct a eqn: N.
```

fails with `Error: Cannot create self-referring hypothesis _0`.

This code never refers to autogenerated names, and it works if `Set Mangle Names.` is removed, so I believe it should also work with `Set Mangle Names.` (that's an invariant I'd expect to hold for this feature).

#### Coq Version

8.9

jashug:
Was failing with "Cannot create self-referring hypothesis" when the generated name equaled the `eqn`.

**Kind:** bug fix

Fixes / closes #9494

This shouldn't break compatibility, because it the only time a different name should be generated is if it would have failed before.

### DIFF
--- a/tactics/tactics.ml
+++ b/tactics/tactics.ml
@@ -4522,8 +4522,11 @@ let induction_gen clear_flag isrec with_evars elim
      declaring the induction argument as a new local variable *)
     let id =
     (* Type not the right one if partially applied but anyway for internal use*)
+      let avoid = match eqname with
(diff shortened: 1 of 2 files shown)


ppedrot (approved):
Not sure if the fix is fully correct, but as long as the bug is solved and no regression is observed, this looks good to me.

Issue #7353, opened 2018-04-26
Anomaly "error with no safe_id attached: Cannot undo." in coqtop


anton-trunov:
#### Version

Coq 8.8.0

#### Operating system

macOS 10.13.3 High Sierra

#### Description of the problem

Here is a `coqtop` session:
```
$  coqtop
Welcome to Coq 8.8.0 (April 2018)

Coq < Undo.
Toplevel input, characters 0-5:
> Undo.
> ^^^^^
Error: Anomaly "error with no safe_id attached: Cannot undo."
Please report at http://coq.inria.fr/bugs/.
```

anton-trunov:
This seems like a regression, since Coq 8.7.2 does not produce an anomaly:
```
$ coqtop
Welcome to Coq 8.7.2 (February 2018)

Coq < Undo.
Toplevel input, characters 0-5:
> Undo.
> ^^^^^
Error: Cannot undo
```

Zimmi48:
This problem is very annoying. I don't understand that it was completely ignored by STM maintainers...

gares:
As reported the bug gives an anomaly rather than an error. And the anomaly still display the error message. All this in a debug tool. To my eyes the priority on this is low.

Zimmi48:
@gares Why doesn't it work in the case that was reported in a duplicate bug report?

```
Definition foo := 1.
Definition bar := 2.
Undo.
(*Error: Anomaly "error with no safe_id attached: Cannot undo."
Please report at http://coq.inria.fr/bugs/.
*)
```

And did you notice the other duplicate bug report saying that it happens as well in CoqIDE?

Zimmi48:
OK, sorry I was confused on the expected behavior of `Undo`. For some reason, I thought `Undo` and `Back` were equivalent. I hadn't realized that `Undo` was only supposed to be used within a proof.

Note that `.. coqtop::` sphinx blocks can be passed the option `undo` which is supposed to allow them to have no effect on the rest of the document. It works by adding `Undo n.` at the end of the block where `n` is the number of sentences in the block. @cpitclaudel were you aware of this limitation? It can be very confusing...

cpitclaudel:
Nope, I wasn't aware of it.  We should change the `undo` flag to issue `Back n` instead of `Undo n`, if that works.

ejgallego:
[...]

PR #9501, opened 2019-02-07
Sphinx: fail when a command fails + other stuff

### LINKED ISSUE #7755: How to ensure that the Coq snippets in the manual are not broken?
It has been a long standing problem that the Coq examples in the manual can sometimes become outdated without us noticing; they can even start failing suddenly. The problem was there with coq-tex and still exists in the Sphinx based manual.

A solution that has been proposed and never implemented is to forbid failures (cc @silene @maximedenes). There are a few places where we want to show Coq failing, but it could still be done using `Fail` (although the output with `Fail` is less advanced, in particular it doesn't show the failure location).

Yet, I fear that this mechanism would still be too simple to catch a lot of problems: when we describe the supposed output in terms more complex than "this works" or "this fails". That's why I'm currently considering that we should instead have a system that is closer to how the test-suite `output/` system works, especially after @SkySkimmer's improvement in #7663. We would store the output along side the input. We could even use the stored outputs by default instead of regenerating them (it would allow the `sphinx` target not to depend on `world`) and have a distinct target (included in `test-suite`) that would regenerate outputs and check them against what was stored (plus an additional target to "approve" new outputs after careful review, also useful when we add a new example).

### LINKED ISSUE #9407: Manual examples broken with nested proofs
Code snippets like here:

https://coq.inria.fr/distrib/current/refman/language/gallina-extensions.html?highlight=generalizable#implicit-generalization

trigger an error about nested proofs. It's a pity that errors on these examples (which are meant to be successfully executed) is not caught by our documentation build.

SkySkimmer:
This uses a new coqtop-only option "Coqtop Fast Death", not sure where to put the doc for it.

Having it as an option allows us to disable it for a block with `coqtop:: fail`, but that's very little used so if needed it could become a command line flag. Those are heavier to implement though.

Closes #7755
Closes #9407.

Some notes:
- Undo is quite broken, eg `Definition x := 1. Undo 1.` says "cannot undo". It seems to work OK to undo tactics but not regular commands. Maybe sphinx should use Back instead? OTOH Back will see a lemma with 42 tactics as just 1 thing so sphinx basing the number on how many commands were sent is still broken. Can we come up with some more structured state reset system?
- ssreflect was using `coqtop:: in` instead of `coqdoc::` quite a lot
- there's a syntax error in ssreflect which seems like it's supposed to work, see FIXME in the rst (I turned it into coqdoc until someone looks at it), cc @gares 
- when a command fails coqtop does `exit 1` then sphinx fails with an exception, but it can be difficult to identify where the rst is broken (the exception doesn't mention the line number, it only says the last command sent to coqtop which can be something common like `Abort` and the last 100 characters of the error message). I don't know how to improve that.
- when a command failed, I fixed it and reran sphinx, it re-processed all previous files including successful ones wasting quite some time. I ended up cheating by doing `ln current-broken-file.rst canonical-structures.rst` (canonical-structures is the first to get processed) but is there a better way?
- sphinx wanted more recursion depth (breaking while processing ring.rst), so I doubled it. No clue what's going on.

Depends https://github.com/coq/coq/pull/9553

### DIFF
--- a/doc/sphinx/proof-engine/ssreflect-proof-language.rst
+++ b/doc/sphinx/proof-engine/ssreflect-proof-language.rst
@@ -4847,7 +4847,7 @@ For instance, the following lemma:
 
 .. coqdoc::
 
-     Lemma andP: forall b1 b2, reflect (b1 /\ b2) (b1 && b2).
+   Lemma andP: forall b1 b2, reflect (b1 /\ b2) (b1 && b2).
 
 relates the boolean conjunction to the logical one ``/\``. Note that in
 ``andP``, ``b1`` and ``b2`` are two boolean variables and the
@@ -4901,7 +4901,7 @@ Let us compare the respective behaviors of ``andE`` and ``andP``.
 
   .. coqtop:: none
 
-     Lemma test (b1 b2 : bool) : if (b1 && b2) then b1 else ~~(b1||b2).
+     Restart.
 
   .. coqtop:: all
 
--- a/doc/sphinx/proof-engine/ssreflect-proof-language.rst
+++ b/doc/sphinx/proof-engine/ssreflect-proof-language.rst
@@ -652,11 +652,7 @@ The tactic:
       Lemma test x :  f x + f x = f x.
       set t := f _.
 
-   .. coqtop:: none
-
-      Undo.
-
-   .. coqtop:: all
+   .. coqtop:: all restart
 
       set t := {2}(f _).
 
@@ -1989,19 +1985,17 @@ be substituted.
          Lemma test l : (length l) * 2 = length (l ++ l).
          case: (lastP l).
 
-      Applied to the same goal, the command:
-      ``case: l / (lastP l).``
-      generates the same subgoals but ``l`` has been cleared from both contexts.
+      Applied to the same goal, the tactc ``case: l / (lastP l)``
+      generates the same subgoals but ``l`` has been cleared from both contexts:
 
-      Again applied to the same goal, the command.
+      .. coqtop:: all restart
 
-      .. coqtop:: none
+         case: l / (lastP l).
 
-         Abort.
+      Again applied to the same goal:
 
-      .. coqtop:: all
+      .. coqtop:: all restart abort
 
-         Lemma test l : (length l) * 2 = length (l ++ l).
          case: {1 3}l / (lastP l).
 
       Note that selected occurrences on the left of the ``/``
@@ -2014,10 +2008,6 @@ be substituted.
    command:
 
    .. example::
-
-      .. coqtop:: none
-
-         Abort.
 
       .. coqtop:: all
 
@@ -2634,7 +2624,7 @@ Since the :token:`i_pattern` can be omitted, to avoid ambiguity,
 bound variables can be surrounded
 with parentheses even if no type is specified:
 
-.. coqdoc::
+.. coqtop:: all restart
 
    have (x) : 2 * x = x + x by omega.
 
@@ -2648,13 +2638,8 @@ copying the goal itself.
 
 .. example::
 
-  .. coqtop:: none
+  .. coqtop:: all restart abort
 
-     Abort All.
-
-  .. coqtop:: all
-
-     Lemma test : True.
      have suff H : 2 + 2 = 3; last first.
 
   Note that H is introduced in the second goal.
@@ -2675,10 +2660,9 @@ context entry name.
 
   .. coqtop:: none
 
-     Abort All.
      Set Printing Depth 15.
 
-  .. coqtop:: all
+  .. coqtop:: all abort
 
      Inductive Ord n := Sub x of x < n.
      Notation "'I_ n" := (Ord n) (at level 8, n at level 2, format "''I_' n").
@@ -2694,11 +2678,7 @@ For this purpose the ``[: name ]`` intro pattern and the tactic
 
 .. example::
 
-  .. coqtop:: none
-
-     Abort All.
-
-  .. coqtop:: all
+  .. coqtop:: all abort
 
      Lemma test n m (H : m + 1 < n) : True.
      have [:pm] @i : 'I_n by apply: (Sub m); abstract: pm; omega.
@@ -2711,11 +2691,7 @@ with have and an explicit term, they must be used as follows:
 
 .. example::
 
-  .. coqtop:: none
-
-     Abort All.
-
-  .. coqtop:: all
+  .. coqtop:: all abort
 
      Lemma test n m (H : m + 1 < n) : True.
      have [:pm] @i : 'I_n := Sub m pm.
@@ -2734,11 +2710,7 @@ makes use of it).
 
 .. example::
 
-  .. coqtop:: none
-
-     Abort All.
-
-  .. coqtop:: all
+  .. coqtop:: all abort
 
      Lemma test n m (H : m + 1 < n) : True.
      have [:pm] @i k : 'I_(n+k) by apply: (Sub m); abstract: pm k; omega.
@@ -2755,19 +2727,19 @@ typeclass inference.
 
   .. coqtop:: none
 
-     Abort All.
-
      Axiom ty : Type.
      Axiom t : ty.
 
      Goal True.
 
-  .. coqdoc::
+  .. coqtop:: all
 
      have foo : ty.
 
   Full inference for ``ty``. The first subgoal demands a
   proof of such instantiated statement.
+
+  .. A strange bug prevents using the coqtop directive here
 
   .. coqdoc::
 
@@ -2778,13 +2750,13 @@ typeclass inference.
   statement. Note that no proof term follows ``:=``, hence two subgoals are
   generated.
 
-  .. coqdoc::
+  .. coqtop:: all restart
 
      have foo : ty := t.
 
   No inference for ``ty`` and ``t``.
 
-  .. coqdoc::
+  .. coqtop:: all restart abort
 
      have foo := t.
 
@@ -2833,10 +2805,9 @@ The ``have`` modifier can follow the ``suff`` tactic.
 
   .. coqtop:: none
 
-     Abort All.
      Axioms G P : Prop.
 
-  .. coqtop:: all
+  .. coqtop:: all abort
 
      Lemma test : G.
      suff have H : P.
@@ -2900,10 +2871,6 @@ proof that quotient and reminder of natural number euclidean division
 are unique.
 
 .. example::
-
-  .. coqtop:: none
-
-     Abort All.
 
   .. coqtop:: all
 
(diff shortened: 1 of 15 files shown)


SkySkimmer:
>Some notes:

Almost forgot:
- the `coqtop::` in the micromega chapter fails due to lack of csdp, we should consider installing it on the CI images

ejgallego (changes requested):
Umm, I am not sure I like the option [which should be renamed anyways]

I suggest you instead use the `-l` option of `coqtop` if you can.

SkySkimmer:
>I suggest you instead use the -l option of coqtop if you can.

How would that work with the sphinx model? It's basically intended for interactive coqtop.

Zimmi48:
First, thanks for working on this. I would personally have preferred a solution comparing the output with a saved output and a promoting functionality, but what you propose is already a serious improvement on the current situation and solves the main issue.

Could you move the fixes to the manual to a separate PR so that they can be backported to 8.9?

>Undo is quite broken, eg `Definition x := 1. Undo 1.` says "cannot undo". It seems to work OK to undo tactics but not regular commands. Maybe sphinx should use Back instead? OTOH Back will see a lemma with 42 tactics as just 1 thing so sphinx basing the number on how many commands were sent is still broken. Can we come up with some more structured state reset system?

See the discussion in #7353.

>when a command fails coqtop does `exit 1` then sphinx fails with an exception, but it can be difficult to identify where the rst is broken (the exception doesn't mention the line number, it only says the last command sent to coqtop which can be something common like `Abort` and the last 100 characters of the error message). I don't know how to improve that.

@cpitclaudel might know...

>when a command failed, I fixed it and reran sphinx, it re-processed all previous files including successful ones wasting quite some time. I ended up cheating by doing `ln current-broken-file.rst canonical-structures.rst` (canonical-structures is the first to get processed) but is there a better way?

Unfortunately, Sphinx's caching is terrible. Both unsound and too weak.

ejgallego:
> How would that work with the sphinx model? It's basically intended for interactive coqtop.

`-l` basically simulates the interactive loading of a file, but however you are right, goal printing was removed from the loop in `Vernac` , let me think....

cpitclaudel:
Thanks for working on this, this is great work. I'll look at the details of the questions a bit later today (thanks for the ping also @Zimmi48)

ejgallego (approved):
Ok then with the option, but please rename to something closer to "Coqtop Exit On Error"

toplevel/coqloop.ml
```diff
@@ -410,7 +415,8 @@ let rec vernac_loop ~state =
     let loc = Loc.get_loc info in
     let msg = CErrors.iprint (e, info) in
     TopErr.print_error_for_buffer ?loc Feedback.Error msg top_buffer;
-    vernac_loop ~state
+    if fast_death () then exit 1
+    else vernac_loop ~state
```
ejgallego: I'd prefer:
```
if exit_on_error () then exit 1;
vernac_loop ~state
```
as to make the critical importance of tail-rec clear here.

SkySkimmer: As you like.

ejgallego: Thanks

gares:
>     * Undo is quite broken,

It is not broken, it is what Hugo wanted it to do.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -2997,7 +2996,7 @@ illustrated in the following example.
 
   .. coqtop:: none
 
-     Abort All.
+     Abort All. Reset Test.
```
gares: No, please, don't use Reset other than `Initial`

SkySkimmer: OK but why?

SkySkimmer:
>there's a syntax error in ssreflect which seems like it's supposed to work, see FIXME in the rst (I turned it into coqdoc until someone looks at it), cc @gares

WG: it isn't intended to be executed, leaving as `coqdoc::`

doc/sphinx/proof-engine/ssreflect-proof-language.rst
gares: Sorry, it was a reflex. If this is for coqtop, then it is fine

cpitclaudel:
> when a command failed, I fixed it and reran sphinx, it re-processed all previous files including successful ones wasting quite some time. I ended up cheating by doing `ln current-broken-file.rst canonical-structures.rst` (canonical-structures is the first to get processed) but is there a better way?

I opened a Sphinx bug to understand why caches are discarded upon any error (https://github.com/sphinx-doc/sphinx/issues/6035)

> when a command fails coqtop does `exit 1` then sphinx fails with an exception, but it can be difficult to identify where the rst is broken (the exception doesn't mention the line number, it only says the last command sent to coqtop which can be something common like `Abort` and the last 100 characters of the error message). I don't know how to improve that.

Here's a patch that should fix this: https://gist.github.com/cpitclaudel/8a9c40ff54485a240d063b02ad9d00f7 . I think it makes sense to merge it as part of this PR, but I can also submit it as a separate PR if preferred.  Now the output looks like this:

```
Extension error:
/build/coq/master/doc/sphinx/addendum/canonical-structures.rst:39: Error while sending the following sentence to coqtop: Module EQ.
```

SkySkimmer:
>Now the output looks like this:

Can we get it to include Coq's error message?

cpitclaudel:
> Can we get it to include Coq's error message?

I've updated the patch at https://gist.github.com/cpitclaudel/8a9c40ff54485a240d063b02ad9d00f7. Can you see if that helps?  It's tricky to capture the message after Coq crashes.

SkySkimmer:
> It's tricky to capture the message after Coq crashes.

See new commit, seems to work.

Zimmi48:
@SkySkimmer Would you mind using the "Co-authored-by" syntax for crediting Clément instead of the @-mention in the commit message?

SkySkimmer:
>@SkySkimmer Would you mind using the "Co-authored-by" syntax for crediting Clément instead of the @-mention in the commit message?

Like this?

SkySkimmer:
>Could you move the fixes to the manual to a separate PR so that they can be backported to 8.9?

https://github.com/coq/coq/pull/9553

Zimmi48:
>Like this?

Like this:
```
Co-authored-by: Clément Pit-Claudel <clement.pitclaudel@live.com>
```

doc/tools/coqrst/coqdomain.py
```diff
@@ -866,36 +868,58 @@ def make_rawsource(pairs, opt_input, opt_output):
...
+        for sentence in self.split_sentences(node.rawsource):
+            pairs.append((sentence, repl.sendone(sentence)))
+        if opt_fail:
+            repl.sendone("Set Coqtop Exit On Error.")
+        if opt_undo:
+            repl.sendone("Undo {}.".format(len(pairs) + 2 * opt_fail))
+
+        dli = nodes.definition_list_item()
+        for sentence, output in pairs:
+            # Use Coqdoq to highlight input
```
Zimmi48:
```suggestion
            # Use Coqdoc to highlight input
```

Zimmi48 (commented):
Could you please also remove the `undo` option in this PR? This would avoid me requesting once more than @vbgl rebases his commit after I merge this PR. Cf. c1ad06146504a12653fa771bdc9e5f4a5bb46cba.

Could you document the `Coqtop Exit on Error` option in the manual as an internal-only option? (The best location is likely in https://coq.inria.fr/refman/practical-tools/coq-commands.html#interactive-use-coqtop.)

While you are at it (and only if you don't mind), I'd like to add two more control options:
- `abort`, which runs `Abort All.` at the end of the block;
- `restart` which runs `Restarts` at the end of the block.

SkySkimmer:
>Could you please also remove the undo option in this PR? This would avoid me requesting once more than @vbgl rebases his commit after I merge this PR.

That got merged right?

>Could you document the Coqtop Exit on Error option in the manual as an internal-only option? 

Done.

>Coqdoq

Done.

>While you are at it (and only if you don't mind), I'd like to add two more control options:

Maybe some other time.

vbgl:
No, the PR that was merged did not include the removal of the `undo` option. It only included changes to the contents of the manual.

SkySkimmer:
>No, the PR that was merged did not include the removal of the undo option

OK, removed undo.

Zimmi48 (approved):

Zimmi48:
It looks like there is still an unexpected failure in the refman...

Zimmi48:
FTR I have fixed the error and intend to force-push to this branch.

SkySkimmer:
OK thanks

Zimmi48:
In addition to fixing the error, I used this PR to push additional commits adding support for the `abort` and `restart` options to the `coqtop` directive, I used these options throughout the manual, and I refactored (in a subsequent commit) the option system of the `coqtop` directive. In particular, in this last commit, I made the display option mandatory: before, it defaulted to `none` if a non-display option was passed and to `in` if no option was passed, and it was not documented at all.

@cpitclaudel Given what I just pushed, I removed my assignment and would appreciate if you could review and merge the PR.

Zimmi48:
Ideally, we would also like the `fail` option to raise a failure if the `coqtop` block did not fail but I don't think this can be done in the current setting (and it's not used often enough to be that important).

Zimmi48:
This has now a trivial conflict with master. @SkySkimmer do you want to fix it or shall I do it?

doc/tools/coqrst/coqdomain.py
```diff
@@ -827,22 +827,41 @@ def split_sentences(source):
...
+        options = node['coqtop_options']
+
+        # Behavior options
         opt_reset = 'reset' in options
-        opt_all, opt_none = 'all' in options, 'none' in options
-        opt_input, opt_output = opt_all or 'in' in options, opt_all or 'out' in options
+        opt_fail = 'fail' in options
+        opt_restart = 'restart' in options
+        opt_abort = 'abort' in options
+        options = options - set(('reset', 'fail', 'restart', 'abort'))
```
cpitclaudel: Sets are usually written as `{…}` instead of `set((…))`

doc/tools/coqrst/coqdomain.py
```diff
@@ -827,22 +827,41 @@ def split_sentences(source):
...
 
-        unexpected_options = list(set(options) - set(('reset', 'undo', 'all', 'none', 'in', 'out')))
+        unexpected_options = list(options - set(('all', 'none', 'in', 'out')))
```
cpitclaudel: Same here

cpitclaudel (approved):
LGTM. I will merge when the conflict is resolved.

cpitclaudel:
Oh, woops; the conflict is already resolved.

doc/tools/coqrst/coqdomain.py
Zimmi48: Sorry, I'm not an advanced Pythonista. I only copied what was already there.

PR #9506, opened 2019-02-07
Remove some non tailrec List.map from CList implementations

### LINKED ISSUE #9482: Type checking mutual recursion consumes way too much stack space

Type checking mutual recursion uses a lot of stack space. [Here](https://gist.github.com/xuanruiqi/3df52612b87867f4b5731fd5efe9c38a) is a Gist that contains a huge mutually recursive block (starting line 1827).

My stack was initially set to 8192KB, and trying to type check the code gives the error:

    File "./Semantics.v", line 2800, characters 6-29239:
    Error: Stack overflow.

After setting my stack size to 65535KB using `$ ulimit -s 65535`, the code successfully type checks (albeit after a long time).

We have since [applied a patch](https://github.com/hazelgrove/hazel/blob/dev/src/semantics/Semantics.v) that works around this by eliminating some of the mutual recursion, and the code now successfully type checks with less stack space. However, we are fairly sure that type checking mutually recursive code is consuming way too much stack space now (even 8192K is a quite large stack size).

Any insights on what is causing the stack overflows?

#### Coq Version
8.8.2

SkySkimmer:
Fix #9482

### DIFF
--- a/clib/cList.ml
+++ b/clib/cList.ml
@@ -780,7 +780,7 @@ let share_tails l1 l2 =
 
 (** {6 Association lists} *)
 
-let map_assoc f = List.map (fun (x,a) -> (x,f a))
+let map_assoc f = map (fun (x,a) -> (x,f a))
 
 let rec assoc_f f a = function
   | (x, e) :: xs -> if f a x then e else assoc_f f a xs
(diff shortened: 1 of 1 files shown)


ppedrot (approved):

ppedrot:
I don't think this is worth a test, stack overflow tests can by quite unreliable depending on the machine they are run on. I'll merge this quickly as-is if nobody objects.

PR #9513, opened 2019-02-07
Edit release-process.md to ease upcoming releases of Coq in Docker Hub

erikmd:
**Kind:** documentation / infrastructure.

This PR is a follow-up of coq/bignums#17.

The idea is that preparing a package for `coq-bignums` almost at the same time as for `coq` would make the process smoother for publishing Coq in Docker Hub for upcoming beta and stable releases. 

As this PR only touches the dev doc, I took the opportunity to add the `[ci skip]` tag in the commit message to avoid triggering the CI accordingly, but tell me if this was not a good idea…

### DIFF
--- a/dev/doc/release-process.md
+++ b/dev/doc/release-process.md
@@ -92,7 +92,11 @@
 ### These steps are the same for all releases (beta, final, patch-level) ###
 
 - [ ] Send an e-mail on Coqdev announcing that the tag has been put so that
-  package managers can start preparing package updates.
+  package managers can start preparing package updates (including a
+  `coq-bignums` compatible version).
+- [ ] Ping **@erikmd** to update the Docker images in `coqorg/coq`
+  (requires `coq-bignums` in `extra-dev` for a beta / in `released`
+  for a final release).
 - [ ] Draft a release on GitHub.
 - [ ] Get **@maximedenes** to sign the Windows and MacOS packages and
   upload them on GitHub.


SkySkimmer:
>As this PR only touches the dev doc, I took the opportunity to add the [ci skip] tag in the commit message to avoid triggering the CI accordingly, but tell me if this was not a good idea…

It's not really worth bothering IMO

ejgallego:
Should we move back `coq-bignums` to the Coq repository [even if as a separate package] ?

I am not sure I see the benefits of having splitted it.

Zimmi48 (approved):

Zimmi48:
>Should we move back `coq-bignums` to the Coq repository [even if as a separate package] ?
>
>I am not sure I see the benefits of having splitted it.

It depends on the future use of this package. If Bignums is made obsolete by primitive integers, then it wouldn't make more sense to keep this package close from Coq than any other contrib. In that case, the package would only be updated for a few more versions anyways, to leave some time for Coq projects to remove their dependency on Bignums. If it still makes sense as a front-end, then this could be an option.

What I want to avoid is: same repo -> split repo -> same repo -> split repo.

erikmd:
Thanks @Zimmi48 for the merge!

> If Bignums is made obsolete by primitive integers

I'm unsure this could happen actually… as Bignums and primitive integers are not two "orthogonal" implementations of the same spec: I'd summarize it by saying that primitive integers provide machine-efficient 63-bit integers while Bignums provide multiple-precision integers and rationals and are now based on primitive integers…

AFAIK Bignums had been removed from Coq's stdlib especially to speed-up the compilation of Coq, given that few Coq users rely on BigN/BigZ/BigQ… but maybe this was not the only reason.

ejgallego:
It could be compiled separately but still live on the repos.

Zimmi48:
>AFAIK Bignums had been removed from Coq's stdlib especially to speed-up the compilation of Coq, given that few Coq users rely on BigN/BigZ/BigQ… but maybe this was not the only reason.

This, and to reduce the size of the standard library, but indeed @ejgallego is right and it could be a separate package living in the same repository and not part of the main build target. This has become easier with Dune support. If Bignums is here to stay and given how its releases are going to be tied up with the ones of Coq that would make sense to put it back in the main repo.

"few Coq users rely on BigN/BigZ/BigQ" is relative when you see that there are two projects in our CI that depend on it (math-classes and color). Are these projects really using the specific features provided by Bignums or could they use only native integers instead? I.e. were they relying on multiple-precision integers only for efficiency reasons (in which case native integers seem to be enough) or for other reasons too?

erikmd:
@Zimmi48

> @ejgallego is right and it could be a separate package living in the same repository and not part of the main build target. This has become easier with Dune support. If Bignums is here to stay and given how its releases are going to be tied up with the ones of Coq that would make sense to put it back in the main repo.

It indeed seems there is a tight coupling between Coq and Bignums' releases, but I'd say this migration would be especially useful if it simplifies your maintenance (I don't know if this would be the case). Anyway, I have the following example in mind: if there is a breaking change in Coq that breaks Bignums, it might be easier to do a single PR that patches Bignums at the same time, rather than doing an overlay…

> "few Coq users rely on BigN/BigZ/BigQ" is relative when you see that there are two projects in our CI that depend on it (math-classes and color).
> Are these projects really using the specific features provided by Bignums
> or could they use only native integers instead? I.e. were they relying on multiple-precision integers only for efficiency reasons (in which case native integers seem to be enough) or for other reasons too?

I guess there are other reasons too, given that the semantics of unbounded integers and of 63-bit arithmetic is quite different… (but I'm not in the best position to answer, maybe @spitters or @fblanqui could confirm or disprove my guess on math-classes' and CoLoR's dependency on BigN/BigZ/BigQ…)

Zimmi48:
>Anyway, I have the following example in mind: if there is a breaking change in Coq that breaks Bignums, it might be easier to do a single PR that patches Bignums at the same time, rather than doing an overlay…

That doesn't happen as often as having to fix other external plugins, so I don't see why this only would justify the move back (if there is a problem with the way we currently patch plugins when preparing PRs, we should address it all at once, i.e. as proposed in https://github.com/coq/coq/issues/6724#issuecomment-443550002). However, moving Bignums back would mean that there is no need to put a tag on the Bignums repo every time there is a Coq release: the Coq tag would be enough.

ejgallego:
Many other things are broken with the current system, for example, Jenkins is not aware of overlays, git bisect, etc...

Zimmi48:
Yes, I agree but this is not specific to Bignums and thus I would like to address these issues in the same way for all the tested projects.

ejgallego:
Bignums is a bit different here as it is a repos under the Coq organization umbrella IMO.

fblanqui:
Le 08/02/2019 à 20:39, Erik Martin-Dorel a écrit :
>
> @Zimmi48 <https://github.com/Zimmi48>
>
>     @ejgallego <https://github.com/ejgallego> is right and it could be
>     a separate package living in the same repository and not part of
>     the main build target. This has become easier with Dune support.
>     If Bignums is here to stay and given how its releases are going to
>     be tied up with the ones of Coq that would make sense to put it
>     back in the main repo.
>
> It indeed seems there is a tight coupling between Coq and Bignums' 
> releases, but I'd say this migration would be especially useful if it 
> simplifies your maintenance (I don't know if this would be the case). 
> Anyway, I have the following example in mind: if there is a breaking 
> change in Coq that breaks Bignums, it might be easier to do a single 
> PR that patches Bignums at the same time, rather than doing an overlay…
>
>     "few Coq users rely on BigN/BigZ/BigQ" is relative when you see
>     that there are two projects in our CI that depend on it
>     (math-classes and color).
>     Are these projects really using the specific features provided by
>     Bignums
>     or could they use only native integers instead? I.e. were they
>     relying on multiple-precision integers only for efficiency reasons
>     (in which case native integers seem to be enough) or for other
>     reasons too?
>
> I guess there are other reasons too, given that the semantics of 
> unbounded integers and of 63-bit arithmetic is quite different… (but 
> I'm not in the best position to answer, maybe @spitters 
> <https://github.com/spitters> or @fblanqui 
> <https://github.com/fblanqui> could confirm or disprove my guess on 
> math-classes' and CoLoR's dependency on BigN/BigZ/BigQ…)
>

Hello. In CoLoR, we use the fact that various kinds of numbers form an 
ordered semi-ring. See Util/Algebra/OrdSemiRing.v. It is instantiated 
with nat, bigN, Z, bigZ, artic numbers (nat or Z plus -oo), tropical 
numbers (nat plus +oo). bigN and bigZ were introduced for efficiency 
reasons. But, still, the properties of ordered semi-rings need to be 
satisfied. Not sure it is the case with numbers modulo N...

PR #483, opened 2017-03-15
More precise version numbers for development versions.

maximedenes:
The goal is to ease the compilation of version-dependent scripts.

As discussed in WG, "trunk" is not precise enough and the sha1 shown by
coqtop is not always present and does not give an order on versions.

PR #9515, opened 2019-02-08
[Gitlab-CI] Automatic deployment of the standard library documentation to GH-pages

vbgl:
There are two objectives:
  - deploying the documentation of the standard library on *protected* branches
  - ease the merging of #9475 to branch 8.9

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -357,7 +357,7 @@ doc:refman:deploy:
     variables:
       - $DOCUMENTATION_DEPLOY_KEY
   dependencies:
-    - doc:refman:dune
+    - doc:refman
   before_script:
     - which ssh-agent || ( apt-get update -y && apt-get install openssh-client -y )
     - eval $(ssh-agent -s)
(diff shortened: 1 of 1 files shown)


ejgallego (approved):

ejgallego:
I couldn't see the pipelines in your fork @vbgl , merging then.

Zimmi48:
Something annoying is that every time the tests fail for some reason (e.g. overlay not yet merged), the deploying stage is not run. There is also the risk that if we relaunch a failed build, a more recent version of the documentation will be overwritten by an older one...

vbgl:
It is annoying, I agree. Thanks for raising the issue.

Do you think it’s a big deal?

Zimmi48:
No, I don't think so. I don't see any better solution than waiting for GitLab CI to implement DAG-style pipelines (see https://gitlab.com/gitlab-org/gitlab-ce/issues/47063). As for the problem of overwriting, let's just not restart failed jobs when a newer job on the same branch has run to completion. It would be good though to have a more precise version number show on the documentation page (in the style of #483).

vbgl:
What about allowing third-party developments to fail when running CI on `master` (or all protected branches)?

We would get the same information as before, but would allow the deployment stage to run.

Zimmi48:
That could be an option but I'm unsure we would actually be informed of the failures in the same way.
Ideally, I'd rather make sure that we reduce the frequency of these failures, e.g. by seriously considering proceeding with something like #6724.

SkySkimmer:
>What about allowing third-party developments to fail when running CI on master (or all protected branches)?

I don't think we can do that, allow_failure is unconditional.

ejgallego:
Failures should not be allowed, that's the real problem and we should not try to workaround it; mtac people is using bors , other plugins should do the same or give developers commit access.

PR #9522, opened 2019-02-08
Update link to refman for master branch.

Zimmi48:
**Kind:** documentation

cc @vbgl

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -9,8 +9,10 @@ The Coq documentation includes
 The documentation of the latest released version is available on the Coq
 web site at [coq.inria.fr/documentation](http://coq.inria.fr/documentation).
 
-Additionally, you can view the documentation for the current master version at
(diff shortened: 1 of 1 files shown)


vbgl (changes requested):
I think there should be also a link to the documentation of the standard library.

Zimmi48:
Sure. I did not put it yet because the deploy stage has never been reached since the PR that introduced the stdlib deployment because of fiat-crypto failing...
I don't mind delaying this PR until such a link can be added.

vbgl:
Here is the link: https://coq.github.io/doc/master/stdlib/

It seems that some CSS files are missing, but that irrelevant to this issue, I guess.

ejgallego:
> It seems that some CSS files are missing, but that irrelevant to this issue, I guess.

We have the same problem in the artifacts, so that's indeed a bug in the build system.

Zimmi48:
@vbgl Updated.

vbgl (approved):

PR #9524, opened 2019-02-08
Fix #7632: Change syntax of autoapply according to the documentation.

### LINKED ISSUE #7632: autoapply is incorrectly documented but maybe we should fix the syntax instead
#### Version

8.8.0

https://coq.inria.fr/refman/addendum/type-classes.html#coq:tacn.autoapply says the syntax is `autoapply term with ident` but the actual syntax is `autoapply term using ident`.
However, we can notice that the actual syntax is inconsistent with the general practice of having `with` as a keyword for hint databases and `using` to add specific lemmas.
So maybe instead of fixing the doc, we should change the syntax and deprecate the current one...

cc @mattam82

Zimmi48:
The documented syntax was using a `with` clause which is more standard with a hint database than the `using` clause that was actually implemented.

I deprecated the old syntax but I don't know how to make the deprecation warning behave as the deprecated attribute, i.e. only warn for direct uses of the tactic, not indirect ones.

Fixes #7632

**Kind:** bug fix

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -97,6 +97,9 @@ Tactics
 
 - Ltac backtraces can be turned on using the "Ltac Backtrace" option.
 
+- The syntax of the `autoapply` tactic was fixed to conform with preexisting
+  documentation: it now takes a `with` clause instead of a `using` clause.
+
 Vernacular commands
 
 - `Combined Scheme` can now work when inductive schemes are generated in sort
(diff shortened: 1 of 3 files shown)


ppedrot:
@Zimmi48 can you rebase?

Zimmi48:
Yep @ppedrot, sorry it took me a week to get down to this notification...

Zimmi48:
ping @ppedrot

ppedrot (approved):
The runtime warning is a bit ugly but IIRC there is no way to do this properly in the current preprocessor. (Strangely enough there is a `tac_deprecation` field in `Tacenv` but I don't see how to set it programmatically.)

PR #9529, opened 2019-02-08
Change Primitive message: "is registered" -> "is declared".

SkySkimmer:
"registered" sounds like it existed before the command.

This could use assumption_message which is currently the same, but I
don't think it has the right semantic.


vbgl (approved):
It is also more consistent with the error message when there are open sections.

SkySkimmer:
@vbgl this is more your code than the github codeowner's, do you want to merge?

PR #9530, opened 2019-02-08
Remove `VtUnkown` classification

### LINKED ISSUE #3632: feature request: require "Instance :=" to be fully solvable

maximedenes:
This is the clean-up we discussed during the working group. It removes the dependency of the current proof mode (and hence the parsing state) on unification.

The current proof mode can now be known simply by parsing and elaborating attributes. We give access to attributes from the classifier for this purpose.

We remove the infamous `VtUnknown` code path in the STM which is known to be buggy.

Fixes #3632 #3890 #4638.


maximedenes:
@Zimmi48 do you know what is a valid syntax for a multiple "Fixes" annotation?

plugins/ltac/g_rewrite.mlg
```diff
@@ -280,7 +280,7 @@ VERNAC COMMAND EXTEND AddSetoid1 CLASSIFIED AS SIDEFF
      }
   | #[ atts = rewrite_attributes; ] [ "Add" "Morphism" constr(m) ":" ident(n) ]
     (* This command may or may not open a goal *)
```
SkySkimmer: remove

maximedenes: Fixed, this was a rebase error. Thanks!

stm/vernac_classifier.ml
```diff
@@ -99,15 +99,18 @@ let classify_vernac e =
       VtStartProof(Doesn'tGuaranteeOpacity, idents_of_name i), VtLater
 
     | VernacDefinition (_,({v=i},_),ProveBody _) ->
-       let guarantee = if poly then Doesn'tGuaranteeOpacity else GuaranteesOpacity in
-        VtStartProof(guarantee, idents_of_name i), VtLater
+      let atts = DefAttributes.parse atts in
```
SkySkimmer: `DefAttributes.parse` can cause an unsupported attribute error, is that OK at classify time?

SkySkimmer: Actually I don't agree with using defattributes here, we should only read the attributes we need.

SkySkimmer (changes requested):

SkySkimmer:
ltac2 has a vtunkown.

gares:
Let me pick the chainsaw :-)

gares:
No worry, it should be sideff+now if I get it right since it is a notation

ppedrot:
The `VtUnknown` from Ltac2 has been removed.

Zimmi48:
>@Zimmi48 do you know what is a valid syntax for a multiple "Fixes" annotation?

I don't. Writing multiple "Fixes" on separate lines works though.

Zimmi48:
It would be good to know how needed / urgent this clean up is for 8.10 given that this goes against our usual compatibility policy.

maximedenes:
> It would be good to know how needed / urgent this clean up is for 8.10 given that this goes against our usual compatibility policy.

Sure, but that's exactly why I gave a talk at the WG. Do you want to have the discussion again?

maximedenes:
> Do you want to have the discussion again?

(or maybe you want a summary of the discussion to be posted here?)

Zimmi48:
I wouldn't mind a brief summary. The discussion at the WG was more general and did not stay long on the topic of removing this option in 8.10 (it quickly drifted to the libobject question). I may have missed some part though. Anyways, I'm not trying to block this clean up, just to get a better understanding of what is depending on it.

For later reference, here is the beginning of the full talk by @maximedenes about separating parsing and execution: https://youtu.be/9fof8dsdmG0?t=1945. And here is the specific point when removing this option in 8.10 is discussed: https://youtu.be/9fof8dsdmG0?t=2685.

maximedenes:
> I wouldn't mind a brief summary.

Sure!

Removing this options makes two things possible:
- Separation of parsing and execution, which IMHO are required for the kind of interaction models that we want to support, both in CoqIDE and LSP. Having a very costly parsing (when it needs to execute) prevents the system to acquire quickly some knowledge about the full document which is being sent.
- Removal of the `VtUnknown` STM codepath. It turns out that the current system handles `Instance` completely incorrectly when the option is on. As you'll see in the PR, we fix at least three major bugs by removing it. But in fact, @gares and I realized there are many more bugs that could have been reported as the dag itself is wrong in such scenarios.

The second item is done as part of this PR, the first will be the subject of subsequent PRs (we need to also clean libobject, etc).

stm/vernac_classifier.ml
maximedenes: Done, thanks!

maximedenes:
@SkySkimmer requested changes addressed

stm/vernac_classifier.ml
```diff
@@ -99,15 +98,18 @@ let classify_vernac e =
       VtStartProof(Doesn'tGuaranteeOpacity, idents_of_name i), VtLater
 
     | VernacDefinition (_,({v=i},_),ProveBody _) ->
-       let guarantee = if poly then Doesn'tGuaranteeOpacity else GuaranteesOpacity in
-        VtStartProof(guarantee, idents_of_name i), VtLater
+      let polymorphic = Attributes.(parse_drop_extra polymorphic atts) in
+      let guarantee = if polymorphic then Doesn'tGuaranteeOpacity else GuaranteesOpacity in
```
SkySkimmer: Maybe this code pattern should be lifted to some function `let guarantee_if_mono atts = if Attributes.(parse_drop_extra polymorphic atts) then Doesn'tGuaranteeOpacity else GuaranteesOpacity`

SkySkimmer (approved):

stm/vernac_classifier.ml
maximedenes: This was already there and its removal is being studied in another PR, so I'll leave it as it is for now.

ejgallego (dismissed):

plugins/ltac/g_rewrite.mlg
```diff
@@ -279,8 +279,7 @@ VERNAC COMMAND EXTEND AddSetoid1 CLASSIFIED AS SIDEFF
          add_setoid atts binders a aeq t n;
      }
   | #[ atts = rewrite_attributes; ] [ "Add" "Morphism" constr(m) ":" ident(n) ]
-    (* This command may or may not open a goal *)
-    => { VtUnknown, VtNow }
+    => { VtStartProof(GuaranteesOpacity, [n]), VtLater }
```
SkySkimmer: I think this is not quite correct, program_mode is included in rewrite_attributes and this command just forwards to the instance command so it has the same behaviour.
The current coqpp setup doesn't make the attributes available to the classify expression so we need to modify it to implement the correct thing. Otherwise we could put user_err "program mode not supported" in add_morphism_infer and leave the attribute issue for later (assuming nobody relies on program mode Add Morphism).

SkySkimmer: Never mind, it passes a partial term so never opens a proof.

SkySkimmer:
Bedrock is still using refine mode.

SkySkimmer:
Is there still a need for a test suite update? The mentioned bugs have tests and the test suite passes.

CHANGES.md
```diff
@@ -120,8 +120,12 @@ Vernacular commands
 
 - `Declare Instance` now requires an instance name.
 
-- Option `Refine Instance Mode` has been turned off by default, meaning that
-  `Instance` no longer opens a proof when a body is provided.
+- Option `Refine Instance Mode` has been turned off and removed, meaning that
+  `Instance` no longer opens a proof when a body is provided. This is a
+  breaking change. We recommend to port impacted developments by first
+  turning the option off in 8.9 (using `Unset Refine Instance Mode`). Note
```
SkySkimmer: I think it may be as easy to just run in master/8.10 and replace failing `Instance foo : Thing := bla. Proof. etc.` by `Instance foo : Thing. Proof. refine bla. etc.`

One may think that there's also the case where the type is gotten from the body, but 
~~~coq
Existing Class True.

Instance x : _ := I.
~~~
gives anomaly grounding non evar-free term so I guess there's no such case. (and `Instance x := I.` doesn't parse)

maximedenes:
I changed my mind a bit on this PR. I'll delay it until after the 8.10 branching.

On the other hand, I'll prepare a deprecation of the `Refine Instance Mode` option for 8.10.

maximedenes:
Have we branched yet? Can I dream of getting this in? @Zimmi48 do you know?

gares:
I can see a v8.10 branch on github, don't you see it?

maximedenes:
> I can see a v8.10 branch on github, don't you see it?

Yes, but I'm wonder if the branching point is now fixed, or if it is just a draft branch...

Zimmi48:
Yes, we have, 8.11 PRs can now be merged and indeed @vbgl should probably have announced it.

SkySkimmer:
Bedrock is still broken AFAICT

SkySkimmer:
Let's merge once bedorck is fixed

gares:
CC @JasonGross

JasonGross:
The issue is that there's no longer a refine instance mode?  Is there an overlay for bedrock2 that needs to be merged, or is there some other issue?
cc @samuelgrueter and @andres-erbsen, who are the people in the group actually working on bedrock2

vernac/classes.ml
```diff
@@ -31,15 +30,7 @@ module NamedDecl = Context.Named.Declaration
...
-let refine_instance = ref false
-
-let () = Goptions.(declare_bool_option {
-  optdepr  = true;
-  optname  = "definition of instances by refining";
-  optkey   = ["Refine";"Instance";"Mode"];
-  optread  = (fun () -> !refine_instance);
-  optwrite = (fun b -> refine_instance := b)
-})
+let typeclasses_db = "typeclass_instances"
```
SkySkimmer: Why's this appearing?

SkySkimmer (changes requested):
Doc needs to be changed, to remove the flag but also https://coq.gitlab.io/-/coq/-/jobs/210533166/artifacts/_install_ci/share/doc/coq/sphinx/html/addendum/type-classes.html#coq:cmd.instance
> Missing fields must be filled in interactive proof mode.

samuelgruetter:
> The issue is that there's no longer a refine instance mode? Is there an overlay for bedrock2 that needs to be merged, or is there some other issue?
> cc @samuelgruetter and @andres-erbsen, who are the people in the group actually working on bedrock2

I just [removed](https://github.com/mit-plv/bedrock2/commit/0c948cf58ac41d9aca0e4cdd4f583af45daefa97) all usages of `Set Refine Instance Mode`, does that solve the problem?

SkySkimmer:
I restarted the job, it'll probably work.

vernac/classes.ml
maximedenes: Probably a wrong rebase.

JasonGross (approved):
Compat file (which I'm code-owner on) looks fine to me, if we are actually removing this earlier than is typical for deprecated options.  No comment on the rest of it.

SkySkimmer:
Bedrock is still failing through riscv https://github.com/mit-plv/riscv-coq/blob/eb0f6c4885c348cb0312fef7beeccd34a875242f/src/Utility/Monads.v#L40-L48

samuelgruetter:
Whoops, let me try again: https://github.com/mit-plv/bedrock2/commit/11ce4fff7a17cdc55771d21e0620e312892faf05.

(The problem was that files which don't import `coqutil.sanity` had `Refine Instance Mode` on by default, because I was working in Coq 8.9, so I expected to get errors but didn't. This time I used Coq master which has `Refine Instance Mode` off by default.)

gares:
Ci green!

vernac/classes.ml
maximedenes: Fixed, thanks!

maximedenes:
Documentation fixed, should be ready modulo CI.

plugins/ltac/g_rewrite.mlg
```diff
@@ -278,7 +278,7 @@ VERNAC COMMAND EXTEND AddSetoid1 CLASSIFIED AS SIDEFF
      }
   | #[ atts = rewrite_attributes; ] ![ proof ] [ "Add" "Morphism" constr(m) ":" ident(n) ]
     (* This command may or may not open a goal *)
-    => { VtUnknown, VtNow }
+    => { VtStartProof(GuaranteesOpacity, [n]), VtLater }
```
SkySkimmer: AFAICT this is `if Lib.is_modtype() then sideff else start_proof`, see the add_morphism_infer code.

SkySkimmer (changes requested):
Sorry for catching the `Add Morphism` classification error so late.
A change entry may also be appropriate.

maximedenes:
@Zimmi48 `doc/sphinx/changes.rst` points to the `Refine Instance Mode` flag. Removing it causes a Sphinx build error. What should I do? Edit the 8.10 changes retroactively?

plugins/ltac/g_rewrite.mlg
maximedenes: Fixed, thanks!

maximedenes:
Classification fixed, changelog entry added.

SkySkimmer (approved):

SkySkimmer:
📣 Cheers 👏

gares:
:1st_place_medal:

Zimmi48:
> What should I do? Edit the 8.10 changes retroactively?

Yes, you did the right thing.

PR #9548, opened 2019-02-11
Almost fully type-safe gramlib implementation

ppedrot:
This PR annotates the gramlib implementation with GADTs in order to ensure type-safety. Most of it consists in moving code into the functor and locally adding type constraints that are systematically propagated.

A few places still rely on unsafe casts, because they require a bit of GADT tinkering I have no time to perform now. An almost type-safe implementation is always better than a fully type-unsafe one.

One place can still be the root of type unsoundness still, which is the handling of inner rules through the `s_rules` function. The API is to blame here, instead of a `ty_production` it should take some variant that is universally quantified over the type of the self entry. Luckily the upper layers of Coq enforce this invariant statically, so that in the end no unsoundness can be exploited.

These unsafe casts are to be fixed in a future PR, but for now I prefer pushing this first so as not to delay the introduction of more static guarantees in gramlib.

EDIT: this PR is better reviewed ignoring the first commit that moves a lot of code, the type-ensuring annotations are per se quite small.


ejgallego (approved):
I reviewed commit per commit; looks good to go. As you say more work is still needed but we have to go step by step.

ejgallego:
cc: @herbelin do you have any comment?

I see no point on delaying this more, as it is a net cleanup; some design questions are of course still open but IMO merging doesn't preclude them from being addressed at a later point.

ejgallego:
Ok will leave 24h more.

PR #9553, opened 2019-02-11
Sphinx various fixing of failing commands

SkySkimmer:
Doc-only #9501 (ie without the changes to coqtop which support checking that new failures don't get introduced) for easy backport.

### DIFF
--- a/Makefile.doc
+++ b/Makefile.doc
@@ -54,6 +54,7 @@ DOCCOMMON:=doc/common/version.tex doc/common/title.tex doc/common/macros.tex
 
 doc: refman stdlib
 
+SPHINX_DEPS ?=
 ifndef QUICK
 SPHINX_DEPS := coq
 endif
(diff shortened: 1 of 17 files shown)


doc/sphinx/addendum/type-classes.rst
```diff
@@ -564,12 +559,12 @@ Settings
    This flag allows to switch the behavior of instance declarations made through
    the Instance command.
 
-   + When it is on (the default), instances that have unsolved holes in
+   + When it is off (the default), they fail with an error instead.
```
Zimmi48: The commit fixing this cannot be backported (since it is not the same in master and v8.9) so this should be moved to another PR. Sorry about that :disappointed:

Zimmi48 (commented):
Thanks! I just have one small comment before I can merge this.

doc/sphinx/addendum/type-classes.rst
SkySkimmer: Removed

PR #9554, opened 2019-02-11
Don't save expected failure logs from opened/ bugs.

SkySkimmer:
Grepping for "Error!" is how we decide if we exit with failure or not.
I don't remember why I used the "==> FAILURE <==" string in the
save-logs script but it was an error.


SkySkimmer:
No secondary maintainer -> free for all PR.

maximedenes (approved):

PR #9557, opened 2019-02-11
[ssreflect] Export more parsing witnesses.

ejgallego:
This is the completion of #9070, needed in order to serialize
ssreflect programs properly. TTBOK this completes the interface for
all generic arguments.

### DIFF
--- a/plugins/ssr/ssrast.mli
+++ b/plugins/ssr/ssrast.mli
@@ -137,6 +137,9 @@ type 'tac ssrhint = bool * 'tac option list
 type 'tac fwdbinders =
         bool * (ssrhpats * ((ssrfwdfmt * ast_closure_term) * 'tac ssrhint))
 
+type 'tac ffwbinders =
+  (ssrhpats * ((ssrfwdfmt * ast_closure_term) * 'tac ssrhint))
+
 type clause =
   (ssrclear * ((ssrhyp_or_id * string) *
               Ssrmatching_plugin.Ssrmatching.cpattern option) option)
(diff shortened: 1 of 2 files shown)


ejgallego:
Ready to merge @coq/ssreflect-maintainers

ppedrot:
FTR, I think this is another instance of a tension between serialization and abstraction. After this PR we will not be able to ensure that nobody is relying on internal details of ssreflect in the wild...

gares:
> we will not be able

I think we were not able before, and are are not able after.

The PR makes things more uniform to me. If we find a way to hide the genarg handles by putting in the root object a serialization method (as people do since the 90s), then I'm all for it.

gares (approved):

ejgallego:
> FTR, I think this is another instance of a tension between serialization and abstraction. After this PR we will not be able to ensure that nobody is relying on internal details of ssreflect in the wild...

You are creating imaginary problems (for your head), and trying to create very very serious problems for the rest of us that like to get things done. Please stop this "abstraction" thing unless you can back it up with real problems the lack of it introduces.

First, because no `.mlg` file declares a `.mli` file, with ssreflect being the only instance so you are not so worried about for example LTAC of `funind`.

Second, all this does is to expose the AST, which while internal, is a perfectly sane thing to expose.

What's your alternative, should people re-implement Coq parser? I know a few folks that have d̶o̶n̶e̶ tried that and it is not something I am looking forward to.

ejgallego:
Thanks @gares

ejgallego:
Keep in mind @ppedrot that this "abstraction" thing [which I don't still understand] is breaking important real-world applications which are implemented in a sensible way; I am gonna defend my turf here as I cannot keep affording to break critical stuff for us as was done in 8.9.

See the tradeoff in pain here.

And this is not at all about serialization; serialization is just an application of it. What this patch does is to expose the AST.

If we want to improve the codebase in this front we should rather start by removing `genarg`.

PR #9560, opened 2019-02-12
[coqlib] Remove `-boot` option for setting the coqlib

ejgallego:
Instead, if the coqlib is special set it explicitly in the command
line, as Dune does.

This is a continuation of #9523.

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -198,7 +198,10 @@ TIMER=$(if $(TIMED), $(STDTIME), $(TIMECMD))
 #   TIME="%C (%U user, %S sys, %e total, %M maxres)"
 
 COQOPTS=$(NATIVECOMPUTE) $(COQWARNERROR) $(COQUSERFLAGS)
-BOOTCOQC=$(TIMER) $(COQC) -boot $(COQOPTS)
+# Beware this depends on the makefile being in a particular dir, we
+# should pass an absolute path here but windows is tricky
+# c.f. https://github.com/coq/coq/pull/9560
+BOOTCOQC=$(TIMER) $(COQC) -coqlib . -boot $(COQOPTS)
 
 LOCALINCLUDES=$(addprefix -I ,$(SRCDIRS))
 MLINCLUDES=$(LOCALINCLUDES)
--- a/doc/tools/coqrst/repl/coqtop.py
+++ b/doc/tools/coqrst/repl/coqtop.py
@@ -41,13 +41,10 @@ class CoqTop:
                            the ansicolors module)
         :param args:       Additional arugments to coqtop.
         """
-        BOOT = True
-        if os.getenv('COQBOOT') == "no":
-            BOOT = False
         self.coqtop_bin = coqtop_bin or os.path.join(os.getenv('COQBIN', ""), "coqtop")
         if not pexpect.utils.which(self.coqtop_bin):
             raise ValueError("coqtop binary not found: '{}'".format(self.coqtop_bin))
-        self.args = (args or []) + ["-boot"] * BOOT + ["-color", "on"] * color
+        self.args = (args or []) + ["-color", "on"] * color
         self.coqtop = None
 
     def __enter__(self):
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -132,7 +132,7 @@ after_script:
   dependencies:
     - not-a-real-job
   script:
-    - SPHINXENV='COQBIN="'"$PWD"'/_install_ci/bin/" COQBOOT=no'
+    - SPHINXENV='COQBIN="'"$PWD"'/_install_ci/bin/"'
     - make -j "$NJOBS" SPHINXENV="$SPHINXENV" SPHINX_DEPS= refman
     - make install-doc-sphinx
   artifacts:
--- a/Makefile.doc
+++ b/Makefile.doc
@@ -31,7 +31,7 @@ DVIPS:=dvips
 HTMLSTYLE:=coqremote
 
 # Sphinx-related variables
-SPHINXENV:=COQBIN="$(CURDIR)/bin/"
+SPHINXENV:=COQBIN="$(CURDIR)/bin/" COQLIB="$(CURDIR)"
 SPHINXOPTS= -j4
 SPHINXWARNERROR ?= 1
 ifeq ($(SPHINXWARNERROR),1)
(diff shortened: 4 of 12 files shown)


SkySkimmer:
Seems to break nix and windows.
I think nix is because sphinx uses -boot (although not in the other refman jobs).
I don't know what the issue is with windows.

ejgallego:
> I don't know what the issue is with windows.

It is actually very hard to get a reliable absolute path on Windows that will work with OCaml.

The problem here is that `pwd` returns something OCaml doesn't like [sight]

Running the test-suite on windows is also broken for Dune for this reason, as I tried to have it use an absolute path for coqlib which seems more robust. I dunno what to here. We could of course use relative paths, but that seems fragile to me, as things will break at the minimal layout change. And for example the test suite does indeed like to do `cd foo` a few times so paths need constant adjustment.

@MSoegtropIMC , do you know a way to get a reliable absolute path on Windows?

SkySkimmer:
>We could of course use relative paths, but that seems fragile to me, as things will break at the minimal layout change.

If you mean passing `-coqlib .` instead of `-coqlib $(PWD)` I don't see how it's more fragile.

ejgallego:
> If you mean passing `-coqlib .` instead of `-coqlib $(PWD)` I don't see how it's more fragile.

In the test suite is worse because quite a few commands do `cd foo`.

By the way I know a way to get a reliable multi-platform `pwd`: write it in OCaml, it will print the path in the right format.

MSoegtropIMC:
> @MSoegtropIMC , do you know a way to get a reliable absolute path on Windows?

You mean in OCaml on Windows?

ejgallego:
> You mean in OCaml on Windows?

Indeed sorry, I mean how to generate an absolute path on Windows that will work for OCaml.

Now if we do `coqc -coqlib $(pwd)` the generated path is incorrect and cannot be parsed by OCaml.

MSoegtropIMC:
If you are on cygwin, PWD will give you a cygwin path, which the MinGW coqc can't digest. Under cygwin this should work:
```
cygpath --absolute --type mixed .
````
type "mixed" means Windows path but with / instead of \. Most Windows software can handle this, since the Windows system commands can handle it.

From a DOS command shell, the PWD solution or just . should work.

If you don't know if you are on cygwin or not, you need a combination of the two.

**Comment by ejgallego** [comment]
> If you don't know if you are on cygwin or not, you need a combination of the two.

Thanks @MSoegtropIMC , unfortunately this would mean adding a lot of checks to see if we are in Windows or Linux, thus I think we'll have to use the more portable solution of using OCaml to print the right path.

**Comment by MSoegtropIMC** [comment]
> more portable solution of using OCaml to print the right path.

I am not sure if this will be portable between cygwin and plain Windows. All the Coq executables my scripts build are MinGW and a MinGW OCaml will create a path which is hard to use on cygwin (not unusable, but at least the path should have forward slashes and the colon in C: might create issues).

**Comment by ejgallego** [comment]
> I am not sure if this will be portable between cygwin and plain Windows. All the Coq executables my scripts build are MinGW and a MinGW OCaml will create a path which is hard to use on cygwin (not unusable, but at least the path should have forward slashes and the colon in C: might create issues).

Well the invariant here is that the path should work in OCaml, so far I have checked that a Windows path produced by OCaml can be used properly by an OCaml program.

**Comment by MSoegtropIMC** [comment]
> Well the invariant here is that the path should work in OCaml, so far I have checked that a Windows path produced by OCaml can be used properly by an OCaml program.

Indeed, this should work, because paths are shared between MinGW applications. But such paths are not necessarily meaningful to e.g. make.

**Comment by ejgallego** [comment]
Ok, let's try with the local path for now.

**Comment by ejgallego** [comment]
Windows is fixed with the relative path, however I am unsure what path to use for the doc build, can you help @coq/doc-maintainers ?

File: Makefile.build [hunk 1]
```diff
@@ -198,7 +198,10 @@ TIMER=$(if $(TIMED), $(STDTIME), $(TIMECMD))
 #   TIME="%C (%U user, %S sys, %e total, %M maxres)"
 
 COQOPTS=$(NATIVECOMPUTE) $(COQWARNERROR) $(COQUSERFLAGS)
-BOOTCOQC=$(TIMER) $(COQC) -boot $(COQOPTS)
+# Beware this depends on the makefile being in a particular dir, we
+# should pass an absolute path here but windows is tricky
+# c.f. https://github.com/coq/coq/pull/9560
```
SkySkimmer: I don't agree that we should pass an absolute path, `.` is fine.
It doesn't depend on the makefile being somewhere in particular, it depends on make being called somewhere in particular. But everything depends on that so it doesn't matter.

doc/tools/coqrst/repl/coqtop.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, coqtop_bin=None, color=False, args=None) -> str:
         self.coqtop_bin = coqtop_bin or os.path.join(os.getenv('COQBIN', ""), "coqtop")
         if not pexpect.utils.which(self.coqtop_bin):
             raise ValueError("coqtop binary not found: '{}'".format(self.coqtop_bin))
-        self.args = (args or []) + ["-boot"] * BOOT + ["-color", "on"] * color
+        self.args = (args or []) + ["-coqlib", "."] + ["-boot"] * BOOT + ["-color", "on"] * color
```
SkySkimmer:
```suggestion
        self.args = (args or []) + ["-coqlib", ".", "-boot"] * BOOT + ["-color", "on"] * color
```
This should fix CI.

ejgallego: Thanks!

Zimmi48:
It seems that we are almost there: only the dune refman build fails now.

ejgallego:
> It seems that we are almost there: only the dune refman build fails now.

We should use a variable I guess.

SkySkimmer:
I pushed a commit which I think fixes it (not tested though), have a look.

ejgallego:
> I pushed a commit which I think fixes it (not tested though), have a look.

Looks good to me, thanks @SkySkimmer ; I think this is ready to go in.

SkySkimmer (approved):
Approving my own changes ;)
Should be squashed since the 1st commit is broken in dune.

ejgallego:
> Should be squashed since the 1st commit is broken in dune.

Ok will squash.

ejgallego:
Needs an owner it seems.

ejgallego:
So main candidate for owner for this PR (and #9576) is @gares , would you mind being the assignee?

If not maybe @SkySkimmer would be the next logical choice? Also @vbgl may be interested in issues related on how we locate the libraries?

gares (approved):

ejgallego:
Thanks!

PR #9561, opened 2019-02-12
[dev] Add include versions for Dune builds.

### LINKED ISSUE #9537: dune and coqtop.byte

I was trying to use coqtop.byte built with dune to debug something in equations and failed.
First issue is that just running coqtop.byte failed, one has to 
```
export LD_LIBRARY_PATH=PATH_TO_COQ/_build/install/default//lib/stublibs:$LD_LIBRARY_PATH
```
For this to run, otherwise the dynamic library linker fails to find a DLL.

After this, processing works, and I can drop to the toplevel, however `#use "include";;` fails, and after adding the PATH_TO_COQ/dev directory to the source files it succeeds but cannot find the object files it needs. Adding #directory "PATH_TO_COQ/_build/default/dev" it still fails with:
```
File "/Users/sozeau/dev/coq/coq/_build/default/dev/top_printers.ml", line 13, characters 5-10:
Error: Unbound module Sorts
Hint: Did you mean Sort?
File "/Users/sozeau/dev/coq/coq/_build/default/dev/vm_printers.ml", line 2, characters 5-9:
Error: Unbound module Term
Unbound value ppid.
...
```

ejgallego:
Fixes #9537

This way, users can do:
```
dune exec coqtop.byte
> Drop.
 # #directory "dev";;
 # #use "include_dune";;
```

### DIFF
--- a/dev/include_dune
+++ b/dev/include_dune
@@ -2,8 +2,9 @@
 
 (* Typical usage :
 
-   $ coqtop.byte    # or even better : rlwrap coqtop.byte
+   $ dune exec coqtop.byte # or even better : rlwrap coqtop.byte
    Coq < Drop.
+   # #directory "dev";;
    # #use "include";;
 
    Alternatively, you can avoid typing #use "include" after each Drop
@@ -15,7 +16,7 @@
 *)
 
 #cd ".";;
-#use "incdir";;
+#use "incdir_dune";;
 #use "base_include";;
-#use "inc_ltac";;
+#use "inc_ltac_dune";;
 #use "include_printers";;
--- a/dev/base_include
+++ b/dev/base_include
@@ -1,23 +1,5 @@
-
 (* File to include to get some Coq facilities under the ocaml toplevel.
    This file is loaded by include *)
-
-#cd".";;
-#directory "parsing";;
-#directory "interp";;
-#directory "toplevel";;
-#directory "library";;
-#directory "kernel";;
-#directory "gramlib";;
-#directory "engine";;
-#directory "pretyping";;
-#directory "lib";;
-#directory "proofs";;
-#directory "tactics";;
-#directory "printing";;
-#directory "grammar";;
-#directory "stm";;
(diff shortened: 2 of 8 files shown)


dev/include_dune
```diff
@@ -0,0 +1,85 @@
...
+
+   Alternatively, you can avoid typing #use "include" after each Drop
+   by adding the following lines in your $HOME/.ocamlinit :
+
+   #directory "+compiler-libs";;
+   if Filename.basename Sys.argv.(0) = "coqtop.byte"
+   then ignore (Toploop.use_silently Format.std_formatter "dev/include")
+*)
+
+(* For OCaml 3.10.x:
```
Zimmi48: Seems like this was copied from somewhere else. You can probably delete this comment.

ejgallego: I will indeed refactor a bit more.

ejgallego:
That's ready, sorry I forgot to remove the tag.

dev/include_dune
```diff
@@ -0,0 +1,22 @@
...
+
+(* Typical usage :
+
+   $ dune exec coqtop.byte # or even better : rlwrap coqtop.byte
+   Coq < Drop.
+   # #directory "dev";;
+   # #use "include";;
+
+   Alternatively, you can avoid typing #use "include" after each Drop
+   by adding the following lines in your $HOME/.ocamlinit :
```
Zimmi48: Since we already ship a `.ocamlinit` file, wouldn't it make sense to include these lines there?

Zimmi48 (approved):
Looks better to me after this refactoring.

dev/include_dune
ejgallego: IMHO not as the included file may have other uses; for example utop will install printers differently [and automatically once we annotate them].

ejgallego:
Failure is spurious.

ejgallego:
Thanks @Zimmi48

Issue #9493, opened 2019-02-06
Problem: name conflict between Ltac argument and loaded library


olaure01:
#### Description of the problem

In the following code:

```
Ltac introsuntil h := intros until h.

Goal forall t : nat, t = t.
Proof.
introsuntil t.
reflexivity.
Qed.

Require Import VectorDef.

Goal forall x : nat, x = x.
Proof.
introsuntil x.
reflexivity.
Qed.

Goal forall t : nat, t = t.
Proof.
Fail introsuntil t.
intros until t.
reflexivity.
Qed.
```

The fact that ```t``` is defined in ```VectorDef``` seems to lead to a clash with the parameter binding of Ltac.
The problem does not really occur when defining directly ```Inductive t := .```, or even inside a module, since in both cases, the last statement is automatically converted into ```forall t0 : nat, t0 = t0```.

#### Coq Version

The Coq Proof Assistant, version 8.9.0 (January 2019)
compiled on Jan 30 2019 12:55:12 with OCaml 4.07.0

maximedenes:
I believe `VectorDef` was designed to not be `Import`ed. Any reason why you can't `Require Import Vector` instead, in your context?

olaure01:
It does not seem to make any difference concerning the problem above: same results with ```Require Import Vector.```

maximedenes:
Sorry, I meant `Require Vector`. I believe that library was designed to not be `Import`ed.

olaure01:
OK but it looks to be an unwanted behavior anyway.

maximedenes:
I think this issue has two levels:
- on the concrete example, I'd recommend not `Import`ing `Vector` and using qualified names, as is intended by the library
- on the more general issue of name clashes, this seems to me like yet another Ltac quirk, to which the answer is usually you should try Ltac2.

Maybe @ppedrot has more to say.

SkySkimmer:
The issue is that it's doing some evaluation of the argument before passing it to the ltac definition, and it gets evaluated to the fully qualified thing.
You can use `Tactic Notation "introsuntil" ident(x) := introsuntil x.` to inform the system about how it should evaluate this stuff.
In general name handling in ltac can get pretty hairy. Also "intros until name" is kinda weird since it relies on alpha-irrelevant names.

olaure01:
Independently of Ltac, there seems to be a difference on the impact of the names defined in a module from a library file and in a module in the current file:
```
Require VectorDef.
Import VectorDef.

[...]

Issue #9323, opened 2019-01-09
Names generated by `Instance` can clash


maximedenes:
Defining anonymous instances can fail at `Qed` because it generates a name which clashes:

in `A.v`
```coq
Class C := { x : True }.
Global Instance: C.
now constructor.
Qed.
```

in `B.v`
```coq
Require A.

Instance: A.C.
now constructor.
Qed.

Import A.

Instance: C.
now constructor.
Qed.
```

This is making the resolution of https://github.com/coq/coq/pull/8094 painful, because the warning suggests to insert an `Import`, but then we often fall in this case, where some instance definitions error.

maximedenes:
@mattam82 do you know where this could come from?

mattam82:
Certainly the fresh name generation for anonymous instances is not using the right set of identifiers to avoid.

mattam82:
Hmm, the Namegen.is_global is misleading, it treats imported refs as not clashing, which hides the fact that the identifier was also used for the first instance declaration in B.v

maximedenes:
Well, `is_global` has no written spec. What does "global" mean in this context? AFAIK, the only thing that matters w.r.t. the kernel is that the label should be fresh in the structure containing the definition.

mattam82:
I suppose for namegen.ml, the spec should be indeed what you propose: the identifier should be free in the current structure so that one can add an object labeled by it in the structure. @herbelin right?

herbelin:
Right, the `is_global` test is wrong and should be what you said. And this presumably applies to the name-generating command `Goal` as well.

PR #9565, opened 2019-02-12
Fix logic for which names are considered taken by globals

SkySkimmer:
See discussion in #9493.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,10 @@ Specification language, type inference
   solved by writing an explicit `return` clause, sometimes even simply
   an explicit `return _` clause.
 
+- Stopped alpha renaming variables to avoid imported globals
+  (previously collisions with globals imported from a file's toplevel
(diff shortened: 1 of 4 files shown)


maximedenes:
This looks closely related to https://github.com/coq/coq/issues/9323, can you check if it fixes it?

SkySkimmer:
>This looks closely related to #9323, can you check if it fixes it?

It doesn't.

SkySkimmer:
This breaks compatibility, do we go forward with overlays or drop it?

SkySkimmer:
If we're changing how to alpha rename things we should also put some thought into it. For instance why do we rename in
~~~coq
Definition x := Prop.
Goal forall x : Prop, x.
~~~
?
What about 
~~~coq
Definition x := Prop.
Module M.
  Goal forall x : Prop, x.
~~~
?

SkySkimmer:
See #9593

PR #9129, opened 2018-12-04
[proof] Removal of imperative state ; interpretation layers only.

ejgallego:
As discussed with @maximedenes , we submit the part of #6171 dealing only with the layers below the STM; this PR makes the proof state functional; thus functions that want to manipulate the global proof state have to explicitly declare it in the argument.

Fixes #6182

Overlays:
- https://github.com/mattam82/Coq-Equations/pull/178
- https://github.com/ppedrot/ltac2/pull/109
- https://github.com/Mtac2/Mtac2/pull/168
- https://github.com/coq-community/paramcoq/pull/22
- https://github.com/lukaszcz/coqhammer/pull/19
- https://github.com/LPCIC/coq-elpi/pull/48
- https://github.com/QuickChick/QuickChick/pull/144

PR #9566, opened 2019-02-12
[proof] Move proofs that have an associated constant to `Lemmas`

ejgallego:
The main idea of this PR is to distinguish the types of "proof object"
`Proof_global.t` and the type of "proof object associated to a
constant, the new `Lemmas.t`.

This way, we can move the terminator setup to the higher layer in
`vernac`, which is the one that really knows about constants, paving
the way for further simplification and in particular for a unified
handling of constant saving by removal of the control inversion here.

Terminators are now internal to `Lemmas`, as it is the only part of
the code applying them.

As a consequence, proof nesting is now handled by `Lemmas`, and
`Proof_global.t` is just a single `Proof.t` plus some environmental
meta-data.

We are also enable considerable simplification in a future PR, as this
patch makes `Proof.t` and `Proof_global.t` essentially the same, so we
should expect to handle them under a unified interface.

~~Depends on #9129~~ (merged)

Overlays:
- https://github.com/coq-community/aac-tactics/pull/47
- https://github.com/coq-community/paramcoq/pull/23
- https://github.com/mattam82/Coq-Equations/pull/196

### DIFF
--- a/dev/doc/changes.md
+++ b/dev/doc/changes.md
@@ -9,6 +9,17 @@
   message. Main change to do generally is to change the flag "Global"
   to "Global ImportDefaultBehavior".
 
+Proof state:
+
+  Proofs that are attached to a top-level constant (such as lemmas)
+  are represented by `Lemmas.t`, as they do contain additional
+  information related to the constant declaration.
+
+  Plugins that require access to the information about currently
+  opened lemmas can add one of the `![proof]` attributes to their
+  `mlg` entry, which will refine the type accordingly. See
+  documentation in `vernacentries` for more information.
+
 ## Changes between Coq 8.9 and Coq 8.10
 
 ### ML4 Pre Processing
@@ -62,6 +73,19 @@ Coqlib:
   vernacular `Register` command; it binds a name to a constant. The second
   command then enables to locate the registered constant through its name. The
   name resolution is dynamic.
+
+Proof state:
+
+- Handling of proof state has been fully functionalized, thus it is
+  not possible to call global functions such as `get_current_context ()`.
+
+  The main type for functions that need to handle proof state is
+  `Proof_global.t`.
+
+  Unfortunately, this change was not possible to do in a
+  backwards-compatible way, but in most case the api changes are
+  straightforward, with functions taking and returning an extra
+  argument.
 
 Macros:
 
--- a/doc/plugin_tutorial/tuto1/src/g_tuto1.mlg
+++ b/doc/plugin_tutorial/tuto1/src/g_tuto1.mlg
@@ -287,7 +287,7 @@ VERNAC COMMAND EXTEND ExploreProof CLASSIFIED AS QUERY
 | ![ proof_query ] [ "ExploreProof" ] ->
   { fun ~pstate ->
     let sigma, env = Pfedit.get_current_context pstate in
-    let pprf = Proof.partial_proof Proof_global.(give_me_the_proof pstate) in
+    let pprf = Proof.partial_proof (Proof_global.get_proof pstate) in
     Feedback.msg_notice
       (Pp.prlist_with_sep Pp.fnl (Printer.pr_econstr_env env sigma) pprf)
   }
--- a/plugins/derive/g_derive.mlg
+++ b/plugins/derive/g_derive.mlg
@@ -24,5 +24,5 @@ let classify_derive_command _ = Vernacextend.(VtStartProof (Doesn'tGuaranteeOpac
 
 VERNAC COMMAND EXTEND Derive CLASSIFIED BY { classify_derive_command } STATE open_proof
 | [ "Derive" ident(f) "SuchThat" constr(suchthat) "As" ident(lemma) ] ->
-  { Derive.(start_deriving f suchthat lemma) }
+  { Derive.start_deriving f suchthat lemma }
 END
--- a/proofs/proof_global.mli
+++ b/proofs/proof_global.mli
@@ -90,12 +90,9 @@ val close_future_proof : opaque:opacity_flag -> feedback_id:Stateid.t -> t ->
 
 val get_open_goals : t -> int
 
-(** Runs a tactic on the current proof. Raises [NoCurrentProof] is there is
-    no current proof.
-    The return boolean is set to [false] if an unsafe tactic has been used. *)
-val with_proof : (unit Proofview.tactic -> Proof.t -> Proof.t * 'a) -> t -> t * 'a
-val simple_with_proof : (unit Proofview.tactic -> Proof.t -> Proof.t) -> t -> t
-val modify_proof : (Proof.t -> Proof.t) -> t -> t
+val map_proof : (Proof.t -> Proof.t) -> t -> t
+val map_fold_proof : (Proof.t -> Proof.t * 'a) -> t -> t * 'a
+val map_fold_proof_endline : (unit Proofview.tactic -> Proof.t -> Proof.t * 'a) -> t -> t * 'a
 
 (** Sets the tactic to be used when a tactic line is closed with [...] *)
 val set_endline_tactic : Genarg.glob_generic_argument -> t -> t
--- a/proofs/proof_global.mli
+++ b/proofs/proof_global.mli
@@ -13,18 +13,16 @@
      environment. *)
 
 type t
-type stack
 
-val get_current_pstate : stack -> t
+(* Should be moved into a proper view *)
+val get_proof : t -> Proof.t
+val get_proof_name : t -> Names.Id.t
+val get_persistence : t -> Decl_kinds.goal_kind
+val get_used_variables : t -> Constr.named_context option
 
-val get_current_proof_name : t -> Names.Id.t
-val get_current_persistence : t -> Decl_kinds.goal_kind
-val get_all_proof_names : stack -> Names.Id.t list
+(** Get the universe declaration associated to the current proof. *)
+val get_universe_decl : t -> UState.universe_decl
 
-val discard : Names.lident -> stack -> stack option
-val discard_current : stack -> stack option
-
-val give_me_the_proof : t -> Proof.t
 val compact_the_proof : t -> t
 
 (** When a proof is closed, it is reified into a [proof_object], where
@@ -44,23 +42,7 @@ type proof_object = {
 
 type opacity_flag = Opaque | Transparent
 
-type proof_ending =
-  | Admitted of Names.Id.t * Decl_kinds.goal_kind * Entries.parameter_entry *
-                UState.t
-  | Proved of opacity_flag *
-              Names.lident option *
-              proof_object
-type proof_terminator
-type closed_proof = proof_object * proof_terminator
-
-val make_terminator : (proof_ending -> unit) -> proof_terminator
-val apply_terminator : proof_terminator -> proof_ending -> unit
-
-val push : ontop:stack option -> t -> stack
-
-val maybe_push : ontop:stack option -> t option -> stack option
-
-(** [start_proof ~ontop id str pl goals terminator] starts a proof of name
+(** [start_proof id str pl goals] starts a proof of name
    [id] with goals [goals] (a list of pairs of environment and
    conclusion); [str] describes what kind of theorem/definition this
    is; [terminator] is used at the end of the proof to close the proof
@@ -68,16 +50,22 @@ val maybe_push : ontop:stack option -> t option -> stack option
    morphism). The proof is started in the evar map [sigma] (which can
    typically contain universe constraints), and with universe bindings
    pl. *)
-val start_proof :
-  Evd.evar_map -> Names.Id.t -> ?pl:UState.universe_decl ->
-  Decl_kinds.goal_kind -> (Environ.env * EConstr.types) list  ->
-    proof_terminator -> t
+val start_proof
+  :  Evd.evar_map
+  -> Names.Id.t
+  -> ?pl:UState.universe_decl
+  -> Decl_kinds.goal_kind
+  -> (Environ.env * EConstr.types) list
+  -> t
 
 (** Like [start_proof] except that there may be dependencies between
     initial goals. *)
-val start_dependent_proof :
-  Names.Id.t -> ?pl:UState.universe_decl -> Decl_kinds.goal_kind ->
-  Proofview.telescope -> proof_terminator -> t
+val start_dependent_proof
+  :  Names.Id.t
+  -> ?pl:UState.universe_decl
+  -> Decl_kinds.goal_kind
+  -> Proofview.telescope
+  -> t
 
 (** Update the proofs global environment after a side-effecting command
   (e.g. a sublemma definition) has been run inside it. Assumes
@@ -86,8 +74,7 @@ val update_global_env : t -> t
 
 (* Takes a function to add to the exceptions data relative to the
    state in which the proof was built *)
-val close_proof : opaque:opacity_flag -> keep_body_ucst_separate:bool -> Future.fix_exn ->
-  t -> closed_proof
+val close_proof : opaque:opacity_flag -> keep_body_ucst_separate:bool -> Future.fix_exn -> t -> proof_object
 
 (* Intermediate step necessary to delegate the future.
  * Both access the current proof state. The former is supposed to be
@@ -99,27 +86,16 @@ type closed_proof_output = (Constr.t * Safe_typing.private_constants) list * USt
  * is allowed (no error), and a warn is given if the proof is complete. *)
 val return_proof : ?allow_partial:bool -> t -> closed_proof_output
 val close_future_proof : opaque:opacity_flag -> feedback_id:Stateid.t -> t ->
-  closed_proof_output Future.computation -> closed_proof
+  closed_proof_output Future.computation -> proof_object
 
-(** Gets the current terminator without checking that the proof has
-    been completed. Useful for the likes of [Admitted]. *)
-val get_terminator : t -> proof_terminator
-val set_terminator : proof_terminator -> t -> t
 val get_open_goals : t -> int
 
 (** Runs a tactic on the current proof. Raises [NoCurrentProof] is there is
     no current proof.
     The return boolean is set to [false] if an unsafe tactic has been used. *)
-val with_current_proof :
-  (unit Proofview.tactic -> Proof.t -> Proof.t * 'a) -> stack -> stack * 'a
-val simple_with_current_proof :
-  (unit Proofview.tactic -> Proof.t -> Proof.t) -> stack -> stack
-
 val with_proof : (unit Proofview.tactic -> Proof.t -> Proof.t * 'a) -> t -> t * 'a
+val simple_with_proof : (unit Proofview.tactic -> Proof.t -> Proof.t) -> t -> t
 val modify_proof : (Proof.t -> Proof.t) -> t -> t
-
-val with_current_pstate : (t -> t * 'a) -> stack -> stack * 'a
-val modify_current_pstate : (t -> t) -> stack -> stack
 
 (** Sets the tactic to be used when a tactic line is closed with [...] *)
 val set_endline_tactic : Genarg.glob_generic_argument -> t -> t
@@ -129,10 +105,3 @@ val set_endline_tactic : Genarg.glob_generic_argument -> t -> t
  * ids to be cleared *)
 val set_used_variables : t ->
   Names.Id.t list -> (Constr.named_context * Names.lident list) * t
-
-val get_used_variables : t -> Constr.named_context option
-
-(** Get the universe declaration associated to the current proof. *)
-val get_universe_decl : t -> UState.universe_decl
-
-val copy_terminators : src:stack -> tgt:stack -> stack
--- a/vernac/lemmas.ml
+++ b/vernac/lemmas.ml
@@ -26,7 +26,6 @@ open Decl_kinds
 open Declare
 open Pretyping
 open Termops
-open Namegen
 open Reductionops
 open Constrintern
 open Impargs
@@ -45,6 +44,46 @@ let call_hook ?hook ?fix_exn uctx trans l c =
     let e = CErrors.push e in
     let e = Option.cata (fun fix -> fix e) e fix_exn in
     iraise e
+
+(* Support for terminators and proofs with an associated constant
+   [that can be saved] *)
+
+type proof_ending =
+  | Admitted of Names.Id.t * Decl_kinds.goal_kind * Entries.parameter_entry * UState.t
+  | Proved of Proof_global.opacity_flag *
+              lident option *
+              Proof_global.proof_object
+
+type proof_terminator = proof_ending -> unit
+
+let make_terminator f = f
+let apply_terminator f = f
+
+(* Proofs with a save constant function *)
+type t =
+  { proof : Proof_global.t
+  ; terminator : proof_terminator CEphemeron.key
+  }
+
+let pf_map f { proof; terminator} = { proof = f proof; terminator }
+let pf_fold f ps = f ps.proof
+
+let set_endline_tactic t = pf_map (Proof_global.set_endline_tactic t)
+
+let with_proof f { proof; terminator } =
+  let proof, res = Proof_global.with_proof f proof in
+  { proof; terminator }, res
+
+let simple_with_proof f ps = fst @@ with_proof (fun t p -> f t p, ()) ps
+
+let by tac { proof; terminator } =
+  let proof, res = Pfedit.by tac proof in
+  { proof; terminator}, res
+
+(** Gets the current terminator without checking that the proof has
+    been completed. Useful for the likes of [Admitted]. *)
+let get_terminator ps = CEphemeron.get ps.terminator
+let set_terminator hook ps = { ps with terminator = CEphemeron.create hook }
 
 (* Support for mutually proved theorems *)
 
@@ -203,9 +242,6 @@ let save ?export_seff id const uctx do_guard (locality,poly,kind) hook universes
 
 let default_thm_id = Id.of_string "Unnamed_thm"
 
-let fresh_name_for_anonymous_theorem () =
-  next_global_ident_away default_thm_id Id.Set.empty
-
 let check_name_freshness locality {CAst.loc;v=id} : unit =
   (* We check existence here: it's a bit late at Qed time *)
   if Nametab.exists_cci (Lib.make_path id) || is_section_variable id ||
@@ -265,7 +301,6 @@ let check_anonymity id save_ident =
     user_err Pp.(str "This command can only be used for unnamed theorem.")
 
 (* Admitted *)
-
 let warn_let_as_axiom =
   CWarnings.create ~name:"let-as-axiom" ~category:"vernacular"
                    (fun id -> strbrk "Let definition" ++ spc () ++ Id.print id ++
@@ -312,7 +347,41 @@ let initialize_named_context_for_proof () =
       let d = if variable_opacity id then NamedDecl.drop_body d else d in
       Environ.push_named_context_val d signv) sign Environ.empty_named_context_val
 
-let start_proof id ?pl kind sigma ?terminator ?sign ?(compute_guard=[]) ?hook c =
+module Stack = struct
+
+  type lemma = t
+  type nonrec t = t * t list
+
+  let map f (pf, pfl) = (f pf, List.map f pfl)
+
+  let map_top ~f (pf, pfl) = (f pf, pfl)
+  let map_top_pstate ~f (pf, pfl) = (pf_map f pf, pfl)
+
+  let pop (ps, p) = match p with
+    | [] -> ps, None
+    | pp :: p -> ps, Some (pp, p)
+
+  let with_top (p, _) ~f = f p
+  let with_top_pstate (p, _) ~f = f p.proof
+
+  let push ontop a =
+    match ontop with
+    | None -> a , []
+    | Some (l,ls) -> a, (l :: ls)
+
+  let get_all_proof_names (pf : t) =
+    let prj x = Proof_global.get_proof x in
+    let (pn, pns) = map Proof.(function pf -> (data (prj pf.proof)).name) pf in
+    pn :: pns
+
+  let copy_terminators ~src ~tgt =
+    let (ps, psl), (ts,tsl) = src, tgt in
+    assert(List.length psl = List.length tsl);
+    {ts with terminator = ps.terminator}, List.map2 (fun op p -> { p with terminator = op.terminator }) psl tsl
+
+end
+
+let start_lemma id ?pl kind sigma ?terminator ?sign ?(compute_guard=[]) ?hook c =
   let terminator = match terminator with
   | None -> standard_proof_terminator ?hook compute_guard
   | Some terminator -> terminator ?hook compute_guard
@@ -323,7 +392,18 @@ let start_proof id ?pl kind sigma ?terminator ?sign ?(compute_guard=[]) ?hook c 
     | None -> initialize_named_context_for_proof ()
   in
   let goals = [ Global.env_of_context sign , c ] in
-  Proof_global.start_proof sigma id ?pl kind goals terminator
+  let proof = Proof_global.start_proof sigma id ?pl kind goals in
+  let terminator = CEphemeron.create terminator in
+  { proof ; terminator }
+
+let start_dependent_lemma id ?pl kind ?terminator ?sign ?(compute_guard=[]) ?hook telescope =
+  let terminator = match terminator with
+  | None -> standard_proof_terminator ?hook compute_guard
+  | Some terminator -> terminator ?hook compute_guard
+  in
+  let proof = Proof_global.start_dependent_proof id ?pl kind telescope in
+  let terminator = CEphemeron.create terminator in
+  { proof ; terminator }
 
 let rec_tac_initializer finite guard thms snl =
   if finite then
@@ -339,7 +419,7 @@ let rec_tac_initializer finite guard thms snl =
        | (id,n,_)::l -> Tactics.mutual_fix id n l 0
        | _ -> assert false
 
-let start_proof_with_initialization ?hook kind sigma decl recguard thms snl =
+let start_lemma_with_initialization ?hook kind sigma decl recguard thms snl =
   let intro_tac (_, (_, (ids, _))) = Tactics.auto_intros_tac ids in
   let init_tac,guard = match recguard with
   | Some (finite,guard,init_tac) ->
@@ -371,14 +451,14 @@ let start_proof_with_initialization ?hook kind sigma decl recguard thms snl =
         List.iter (fun (ref,imps) ->
 	  maybe_declare_manual_implicits false ref imps;
           call_hook ?hook ctx [] strength ref) thms_data in
-      let pstate = start_proof id ~pl:decl kind sigma t ~hook ~compute_guard:guard in
-      let pstate = Proof_global.modify_proof (fun p ->
+      let lemma = start_lemma id ~pl:decl kind sigma t ~hook ~compute_guard:guard in
+      let lemma = simple_with_proof (fun _ p ->
           match init_tac with
           | None -> p
-          | Some tac -> pi1 @@ Proof.run_tactic Global.(env ()) tac p) pstate in
-      pstate
+          | Some tac -> pi1 @@ Proof.run_tactic Global.(env ()) tac p) lemma in
+      lemma
 
-let start_proof_com ~program_mode ?inference_hook ?hook kind thms =
+let start_lemma_com ~program_mode ?inference_hook ?hook kind thms =
   let env0 = Global.env () in
   let decl = fst (List.hd thms) in
   let evd, decl = Constrexpr_ops.interp_univ_decl_opt env0 (snd decl) in
@@ -410,7 +490,7 @@ let start_proof_com ~program_mode ?inference_hook ?hook kind thms =
     else (* We fix the variables to ensure they won't be lowered to Set *)
       Evd.fix_undefined_variables evd
   in
-  start_proof_with_initialization ?hook kind evd decl recguard thms snl
+  start_lemma_with_initialization ?hook kind evd decl recguard thms snl
 
 (* Saving a proof *)
 
@@ -425,7 +505,7 @@ let () =
       optread  = (fun () -> !keep_admitted_vars);
       optwrite = (fun b -> keep_admitted_vars := b) }
 
-let save_proof_admitted ?proof ~pstate =
+let save_lemma_admitted ?proof ~(lemma : t) =
   let pe =
     let open Proof_global in
     match proof with
@@ -440,8 +520,8 @@ let save_proof_admitted ?proof ~pstate =
       let sec_vars = if !keep_admitted_vars then const_entry_secctx else None in
       Admitted(id, k, (sec_vars, (typ, ctx), None), universes)
     | None ->
-      let pftree = Proof_global.give_me_the_proof pstate in
-      let gk = Proof_global.get_current_persistence pstate in
+      let pftree = Proof_global.get_proof lemma.proof in
+      let gk = Proof_global.get_persistence lemma.proof in
       let Proof.{ name; poly; entry } = Proof.data pftree in
       let typ = match Proofview.initial_goals entry with
         | [typ] -> snd typ
@@ -453,10 +533,10 @@ let save_proof_admitted ?proof ~pstate =
       let universes = Proof.((data pftree).initial_euctx) in
       (* This will warn if the proof is complete *)
       let pproofs, _univs =
-        Proof_global.return_proof ~allow_partial:true pstate in
+        Proof_global.return_proof ~allow_partial:true lemma.proof in
       let sec_vars =
         if not !keep_admitted_vars then None
-        else match Proof_global.get_used_variables pstate, pproofs with
+        else match Proof_global.get_used_variables lemma.proof, pproofs with
           | Some _ as x, _ -> x
           | None, (pproof, _) :: _ ->
             let env = Global.env () in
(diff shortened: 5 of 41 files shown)


dune
```diff
@@ -42,3 +42,5 @@
  (name runtest)
  (package coqide-server)
  (deps test-suite/summary.log))
+
+(dirs (:standard _build_ci))
```
SkySkimmer: Make a PR with this already ;) it keeps creeping in

doc/plugin_tutorial/tuto1/src/g_tuto1.mlg
```diff
@@ -145,10 +145,12 @@ END
...
 
 VERNAC COMMAND EXTEND ExploreProof CLASSIFIED AS QUERY
-| [ "Cmd9" ] ->
-  { let p = Proof_global.give_me_the_proof () in
-    let sigma, env = Pfedit.get_current_context () in
-    let pprf = Proof.partial_proof p in
-    Feedback.msg_notice
-      (Pp.prlist_with_sep Pp.fnl (Printer.pr_econstr_env env sigma) pprf) }
+| ![ proof ] [ "Cmd9" ] ->
+  { fun ~pstate ->
```
SkySkimmer: Can't we provide a syntax like with attributes such that `![ proof ]` binds `proof` in the body?

ejgallego: We could, however it is quite practical to use named arguments here and for example imagine we have:
```ocaml
val foo : pstate:Lemmas.t -> int -> Lemmas.t
```
then you can put in a rule that parses `int_arg`:
```ocaml
  { foo int_arg }
```
by the way it was a large mistake not to use ppx for our preprocessor, it is so annoying not to have syntax support....

plugins/ssr/ssrelim.ml
```diff
@@ -321,11 +323,11 @@ let ssrelim ?(is_case=false) deps what ?elim eqid elim_intro_tac =
   let elim_pred, gen_eq_tac, clr, gl = 
     let error gl t inf_t = errorstrm Pp.(str"The given pattern matches the term"++
       spc()++pp_term gl t++spc()++str"while the inferred pattern"++
-      spc()++pr_constr_pat (EConstr.Unsafe.to_constr (fire_subst gl inf_t))++spc()++ str"doesn't") in
+      spc()++pr_econstr_pat env (project gl) (fire_subst gl inf_t)++spc()++ str"doesn't") in
```
gares: Don't call `fire_subst` here (it is just `nf_evar` and `EConstr` do it for you).

gares (dismissed):
I stopped reviewing when I relealized it depends on another PR and that the `env` is wrong (comment `XXX`) and I've pointed out there.

Still there is a minor thing to change here too.

ejgallego:
Ok, this is ready for review and hopefully merge [a few overlays will be needed tho]

Any takes?

dev/doc/changes.md
```diff
@@ -43,6 +43,19 @@ Coqlib:
   command then enables to locate the registered constant through its name. The
   name resolution is dynamic.
 
+Proof state:
+
+- Handling of proof state has been fully functionalized, thus it is
+  not possible to call global functions such as `get_current_context ()`.
+
+  The main type for functions that need to handle proof state is
+  `Lemmas.t`, for proofs that intend to associate a constant.
```
gares: I've nothing against having the proof state being define in the module Lemmas, but having it names `Lemmas.t` is not a good idea. Either put it in a sub module `ProofState` or call it `proof_state`.

gares: To clarify, `Lemmas.t` is also for "ongoing proofs" not just for finished ones.

ejgallego: Note that we are not just moving proof state here, but the complete information relevant for "Lemma-like" constants which can have obligations attached, etc...

If you keep that in mind I think that `Lemmas.t` is not so bad, indeed it is an object that contains a Lemma! I am open to other namings of course but I am not sure, with that in mind, that "proof state" is accurate or btter.

ejgallego: So we will have more or less:
- `Proof.t`: "proof objects", that as of today can have several entries but that's a bit of a mess.
- `Declaration.t`: Declaration-like objects, this can contain several constants, and hooks, they extend `Proof.t`
- `Lemma.t`: extends `Declaration.t` and contains a few more lemma-specific information; maybe indeed it should be merged with `Declaration.t`, however the codepaths are duplicated so even if the current series of PRs improve the situation it will take more effort.

ejgallego: I am defending thus the terminology where we reserve "proof" for proof objects attached or not to constants.

ejgallego: @gares I've fixed the changes entry tho, which indeed was confusing and incorrect.

ejgallego: > To clarify, `Lemmas.t` is also for "ongoing proofs" not just for finished ones.

Good point, `Lemmas.t` could contain a finished proof associated to a constant, but somehow not yet propagated to a certain part of the env.

gares: In `Lemmas.mli` there are API getting arguments of type `Lemma.t` with the label `~pstate`, eg the ones for saving a proof. I guess they should take a `~lemma` now?

Anyway, the doc is more clear now and it explains why `Lemma.t` is not just a renaming of `Proof_global.t`. Still it represent an ongoing proof (of a lemma). So to me the name could still be proof state, even if it is larger than the old proof state. Let's see what other think about the new names.

ejgallego:
Of all the chain of "proof handling" PRs, this is the only one that IMO should strongly go in 8.10, it is a stupid but very necessary API change; it'd be silly to have a temporary API in 8.10.

Also, it is IMO quite low risk.

ejgallego:
Regarding 8.10, we have the following chain depending on this:
- https://github.com/coq/coq/pull/9638
- https://github.com/coq/coq/pull/9641
- https://github.com/coq/coq/pull/9642
- https://github.com/coq/coq/pull/9645

at least the first one is a fix, and the rest also may fix or improve some tricky stuff w.r.t. async stuff; so I dunno; maybe it would be worth to try to include them, I see more to gain than to lose.

dev/doc/changes.md
ejgallego: Good point about the `pstate` name, I did not rename as to easy porting, but I agree that I should rename at least a few. In some cases it should be `~lemma`, in some others I guess `~ontop` is still the right choice.

ejgallego: I like the `Lemma.t` as for example we can provide `Lemma.equal` etc..., and it is shorter, but `Lemma.proof` etc... could also work; I dunno, indeed let's hear feedback.

ejgallego: The thing is that I understand `Lemma.proof` as a projection `proof : Lemma.t -> Proof.t`

ejgallego: Also `start_proof` and friends should likely be named `start_lemma`; I didn't rename as to ease diffs, but I am OK with renaming if we think it is worth it.

ejgallego: @gares I've done a pass on argument naming, I've also removed some spurious uses of `Lemma.t` where it should be `Proof_global.t`; before things where hard to see with the huge dependency.

ejgallego:
CI is green, naming was much improved after @gares comments [personally I find it is OK, not perfect, but OK]; waiting for an assignee, which could be anyone on the reviewer tab.

ejgallego:
Rebased, fixed conflict.

ejgallego:
@ppedrot you seem to be the principal codeowner here, would you like to be assignee?

ppedrot (approved):
I only had a cursory look at the patches, but haven't seen anything too suspicious. This is *gloubiboulga* code anyways, so that as long as it's apparently working as intended, I am fine with it. Furthermore I'm trusting @gares who is the expert on this kind of STM hackery.

ppedrot:
@gares If you don't complain by then, I'll merge this tomorrow evening.

gares:
I have not reviewed this code yet, I made only comments based on the doc.

gares:
I gave I quick look and I don't like the new code. I'll write a linger review later today.

gares:
90% of this PR is about renaming `~pstate` to `~ontop`. this makes the review hard and does not improve the code. Actually you declare a new command saying `![proof]` (that is, I need to see the proof) and you receive an argument called `~ontop`. In some API in lemmas, to support nesting, it may make sense to get the old state with the name `~ontop` but in the rest of the system the proof state (weather it is Lemma.t or Proof_global.t, or you name it) cannot be named `~ontop`. Take this example:
```diff
- let current_proof_name = Proof_global.get_current_proof_name pstate in
+ let current_proof_name = Lemmas.pf_fold Proof_global.get_current_proof_name ontop in
```
Taking the name of the current proof from `pstate` makes sense, taking it out of `ontop` not at all, at least to me.

The actual objective of PR seems reasonable, but the way it is done seems to maximize breakage in ML by renaming things in a way that does not clarify what data one is manipulating.

gares (dismissed):
avoid renaming pstate to ontop

ejgallego:
> avoid renaming pstate to ontop

I did rename in base of your last commentary; at first I was wary and I did an usage analysis and indeed it seems that `ontop` is the most used; for example the current version of the PR removes 7 occurrences of `~ontop:pstate` while it doesn't introduce spurious named renamings AFAICS.

I tried the renaming back to the original version [where you didn't like the naming either and I think it makes things worse]

> This is where problems begin: you declare that the command has to see the `![proof]` and it received a thing called `ontop`.

We could name the attribute `ontop` if that helps. I am not too fond of `ontop` but I cannot come with a different name. At least we still have margin as this API is introduced in 8.10 so we can improve.

> weather it is Lemma.t or Proof_global.t, or you name it

well these two objects are very different and they will differentiate more; ~pstate is reserved for Proof_global.

So I'm a bit stuck, more concrete suggestions welcome. Moving back to `pstate` makes things noticeable worse; renaming `![proof]` to `![ontop]` of `lemma` could work [but that's a nit].

The ontop name is not my favorite, but it makes for _vernaculars_; this is very important, note that vernaculars that are declaring this attribute are in fact running a command in a nested proof. That's already bad for itself, and there are so many because in #9129 it was asked to preserve the previous buggy behavior.

Many of these commands should not take a lemma state but instead use the global one.

This will be even more enforced once tactics are not considered vernaculars as such.

gares:
I know I've asked you to do some renaming, but I never said that `ontop` was a reasonable name for the proof/lemma state. I'm sorry you spent time on this renaming in reaction to a comment of mine.

I think I could leave with `start_proof ~inside:Lemmas.(nested_lemmas lstate)` (where `lstate:Lemmas.t`) and
`COMMAND ... ![nesting] ... (fun ~nested_lemmas -> add_morphism ~inside:nested_lemmas ...)`. I don't love it, but at least one gets that we are talking about nested lemmas and that are putting stuff inside the nesting. If `pstate` is not correct (since it holds more than a proof), then I suggest `lstate` for lemmas_state. I propose inside as a label, but I could live with ontop there.

ejgallego:
> I know I've asked you to do some renaming, I'm sorry you spent time on this renaming in reaction to a comment of mine.

On the contrary I'm glad you asked to rename as IMHO the original state was much worse than what we have now; the pass was worth it, specially as this API is not yet in a stable release.

> but I never said that `ontop` was a reasonable name for the proof/lemma state. 

I don't remember where I got the name, but more below.

> I think I could live with `start_proof ~inside:Lemmas.(nested_lemmas lstate)` (where `lstate:Lemmas.t`) and
> `COMMAND ... ![nesting] ... (fun ~nested_lemmas -> add_morphism ~inside:nested_lemmas ...)`. I don't love it, but at least one gets that we are talking about nested lemmas and that are putting stuff inside the nesting. If `pstate` is not correct (since it holds more than a proof), then I suggest `lstate` for lemmas_state. I propose inside as a label, but I could live with ontop there (as a label it is fine, you don't need to call the data the same way even if the language makes it easy).

Ok so let me try to split this in parts:
- `~ontop` -> `~inside`: no problem in doing that but IMO not an improvement here, and it takes more space.
- `~inside:Lemmas.(nested_lemmas lstate)` not sure what this means @gares, can you elaborate.
- `lstate:Lemmas.t`: for this I do prefer the current convention `~lemma`. In fact I hope we can remove `pstate` too; in the current view of the interface, `Lemmas.t` and `Proof.t` are constant and proof objects, there is nothing about state on them thus in this case I strongly prefer `lemma:Lemmas.t` and `proof:Proof.t`.
- `![proof]` -> `nesting` I don't mind but I am not sure this is an improvement.
- `(fun ~nested_lemmas -> add_morphism ~inside:nested_lemmas ...)` here I'd rather have the use and the coqpp label match so we can omit the `~inside:nested_lemmas` part. This is IMO important and we should fix the remaining discrepancies instead of adding more; but I am waiting on removing more of `Proof_global`.

So as I strongly prefer not to use something with `Xstate` in the name, I guess the main question is what to do with `ontop`; I dunno, I think `inside` is not so much of an improvement, maybe we can think something better. As for `![proof]` indeed maybe here using `lstate` would be better, I dunno, that's OK to change, at the very least we should rename it to `![lemma]` or `![lemma_data]` but that's too long.

ejgallego:
Maybe `ontop` -> `lemmas` is the best choice. Note that `Lemmas.t` came in as this was the existing module where this state fit, we could of course rename the module to something more fit [doesn't bother me too much on the other hand is not 100% perfect]

gares:
>     * in the current view of the interface, `Lemmas.t` and `Proof.t` are constant and proof objects,

I'm a bit lost, didn't I see code pulling out the current proof context or open proofs out of a Lemma.t ?

Wrt labels, my problem with ontop is not when you use it as an argument, but when you bind with a let.
The point is that `ontop` describes a role played by something (be the base of) but not what the something is.
Eg in the stm API you couw say `let oldid = ... in Stm.whatever ~ontop:oldid`, here you mix the two concepts to me.

ejgallego:
> I'm a bit lost, didn't I see code pulling out the current proof context or open proofs out of a Lemma.t ?

A lemma object contains a proof object indeed, but these are separated.

> Wrt labels, my problem with ontop is not when you use it as an argument, but when you bind with a let.

This I agree, let me review the code.

However indeed there are some parts that you want to access the inner proof object when you call the lower layers.

ejgallego:
> This I agree, let me review the code.

There is only a few cases like this however I am not sure how to fix them as I would like to shadow the old lemma stack structure for real; this is a problem when the functions can conditionally open lemmas, like in class which is the main offender.

The other offender is funind and I have not a lot of idea what is the intent there.

ejgallego:
Ok, so I've done a pass and I've addressed:

- improved documentation, the old one indeed was still far from good; I hope it is better now,
- renamed wrong uses of `ontop` as noted by @gares ; I've stuck to "ontop" vs "inside" as we are talking about a stack. The convention is now `ontop:Lemmas.t option` and `lemma:Lemmas.t` [without the option] It is not perfect but IMHO distinguishes the cases where you may have an open proof about a constants vs the case you _always_ have it. I hope that answers your question @gares (ontop == optional; lemma == always)
- rebased the overlays and the PR.

I think we are close to reaching the point of diminishing returns here on the API, in particular as the current one in `master` is much worse than this not yet perfect PR. Also 10 PRs depend on this one which is also a factor. I thus propose that unless we see a critical problem we merge.

WDTY @gares @ppedrot ?

plugins/derive/g_derive.mlg
```diff
@@ -24,5 +24,5 @@ let classify_derive_command _ = Vernacextend.(VtStartProof (Doesn'tGuaranteeOpac
 
 VERNAC COMMAND EXTEND Derive CLASSIFIED BY { classify_derive_command }
 | ![ proof ] [ "Derive" ident(f) "SuchThat" constr(suchthat) "As" ident(lemma) ] ->
-  { fun ~pstate -> Some Derive.(start_deriving f suchthat lemma) }
+  { fun ~ontop:_ -> Some Derive.(start_deriving f suchthat lemma) }
```
gares: given that `ontop` is unused, can `![proof]` be omitted? Alternatively if `Derive.` cannot be nested, we should error when `ontop isSome`.

ejgallego: > can `![proof]` be omitted?

Indeed that's a bug in derive, will fix as suggested thanks!

vernac/lemmas.ml
```diff
@@ -476,37 +557,38 @@ let save_proof_admitted ?proof ~pstate =
...
+  if Option.is_empty lemma && Option.is_empty proof then
     user_err (str "No focused proof (No proof-editing in progress).");
   let (proof_obj,terminator) =
     match proof with
     | None ->
       (* XXX: The close_proof and proof state API should be refactored
          so it is possible to insert proofs properly into the state *)
-      let pstate = Option.get pstate in
-      Proof_global.close_proof ~opaque ~keep_body_ucst_separate:false (fun x -> x) pstate
+      (* XXX uh! *)
```
gares: Can you make this comment more informative? What is the problem? What is the `_`?

ejgallego: I think the "uh!" should be just removed, it got there in some rebase, but the comment just before is the right one.

plugins/derive/g_derive.mlg
ejgallego: Fixed.

gares:
Apart for the few comments above, I still believe that renaming pstate into ontop is not improving the code and makes this PR extremely noisy for no good reason. Just look at vernacentries.

ejgallego:
> I still believe that renaming pstate into ontop is not improving the code and makes this PR extremely noisy for no good reason.

Indeed the improvement is quite small [only 7 cases of `ontop:pstate` eliminated] so I dunno; a few thoughts:
- noise would be a blocker IMHO if this API was in a stable release; as it just introduced IMHO we can live with it as it won't be observable.
- I can revert to `pstate` but that seems like a lot of work and I've never liked `pstate` either, `lemmas` could work but it becomes a bit confusing as it is an option type. Not sure what to do. At least now names are coherent with types.

:dizzy_face:

ejgallego:
CI green.

ejgallego:
So from my part, even if not perfect, I'm again happy pushing this @ppedrot .

@gares , any other change you would like to see? So far I have in my list:

> avoid renaming pstate to ontop

If you thing is crucial I'll be happy to do the renaming to the previous state; the first version of the patch used that scheme already, however after a lot of thinking I (very weakly) prefer `ontop`. These are the functions that take an ontop parameter:
```
val start_deriving        : ontop:Lemmas.t option
val do_generate_principle : ontop:Lemmas.t option
val make_graph            : ontop:Lemmas.t option
val declare_relation      : ontop:Lemmas.t option
val add_setoid            : ontop:Lemmas.t option
val add_morphism_infer    : ontop:Lemmas.t option
val add_morphism          : ontop:Lemmas.t option
val new_instance          : ontop:Lemmas.t option
val context               : ontop:Lemmas.t option
val do_assumptions        : ontop:Lemmas.t option
val declare_assumption    : ontop:Lemmas.t option
val do_definition         : ontop:Lemmas.t option
val do_fixpoint           : ontop:Lemmas.t option
val do_cofixpoint         : ontop:Lemmas.t option
val declare_fixpoint      : ontop:Lemmas.t option
val declare_cofixpoint    : ontop:Lemmas.t option
val declare_definition    : ontop:Lemmas.t option
val declare_fix           : ontop:Lemmas.t option
val start_lemma           : ontop:t option
```
many of them already take it on master, so IMHO it reads better "open proof on top" that "open proof with pstate".

gares:
> these are the functions that take an ontop parameter:

I'm ok with the parameter serving the purpose of nesting lemmas to be called ontop.

What I'm *not* OK with is renaming you seem to have applied in order to be able to write just `~ontop/~lemma` instead of `~ontop/lemma:pstate`. If you look at the diff of vernacentries https://github.com/coq/coq/pull/9566/files#diff-6ef67cbe9a5a78e16a501c71671ece92 you surely realize what I mean. Please do not operate the rename that pervasively (I suspect you did a global search/replace), it is not only generating "noise" in this PR but it is also making things worse. For example in vernacentries now `show_proof` takes a singe argument called `~ontop`: it does not make sense, the old `~pstate` is clearly better (and `~lemma` would work too), but the API in question is not showing the proof `~ontop` of anything, it is showing the proof of the thing you pass to it (be it a lemma handle or a proof state handle).

ejgallego:
> Please do not operate the rename that pervasively

Oh indeed, thanks for the careful review; I've fixed a few wrong renamings. Things should (hopefully) be finally consistent.

> you seem to have applied in order to be able to write just `~ontop/~lemma` instead of `~ontop/lemma:pstate`

I do consider having to write `~ontop:foo` almost always a bug, if we have to do that our labeling scheme is not consistent then, isn't it?

Actually @gares I think I've found a better solution for naming for 8.11, which does indeed come from ongoing work on refining the type of a vernac. This basically extends coqpp to work in a way similar to attributes, so we can finally declare "this vernac is pure", or "this vernac can nest" or "this vernac is a proof transformer" or "this vernac alters this part of the state". IMHO once we have that the naming becomes much better than the current too generic one; however that work in on top of 30 commits so it will take a while to reach master [and not yet fully done].

That should also have the good effect of making the classifier enforced by typing.

By the way @maximedenes , I am citing you here as to communicate my plans w.r.t. vernac, I dunno how I could do better here; I am not sure that submitting another 10 draft PRs on top of the 5 pending would help a lot.

maximedenes:
> I am citing you here as to communicate my plans w.r.t. vernac, I dunno how I could do better here

Thanks, I think I get the picture, sounds good!

ejgallego:
CI green.

gares:
> I do consider having to write `~ontop:foo` almost always a bug, if we have to do that our labeling scheme is not consistent then, isn't it?

I don't agree, and that is why I tried to make the distinction between what a thing is and the role it may play for a specific function.

Anyway, the current patch is still unsatisfactory, take again vernacentries, the first hunk is nonsensical.
```diff
- let there_are_pending_proofs ~pstate =
-   not Option.(is_empty pstate)
+ let there_are_pending_proofs ~ontop =
+  not Option.(is_empty ontop)
```
that is just noise and the change was made because of vernac_register that in turn was renamed because in 
```diff
- let rec interp_expr ?proof ~atts ~st c : Proof_global.t option =
-  let pstate = st.Vernacstate.proof in
+ let rec interp_expr ?proof ~atts ~st c : Lemmas.t option =
+  let ontop = st.Vernacstate.lemmas in
```
you decided to name the state `ontop` instead of `pstate` as it was previously named.
Revert this change please. If some API you call in vernac_register require an ontop argument now, just pass `~ontop:pstate` to them.

ejgallego:
> That change is just noise.

On the contrary the renaming of `pstate -> ontop` in the inner vernac loop was done on purpose, as vernacs are sometimes interpreted on top of a lemma stack.

> you decided to name the state `ontop` instead of `pstate` as it was previously named.

Yup, I consider my original naming to be a bug, so in this case I prefer the new schema as justified above by the use of `ontop`. The argument of noise IMHO doesn't hold as this is just refinement of a newly introduced API, which is a normal event in the development process. If the API was in 8.9 of course the renaming would not be justified.

> > I do consider having to write `~ontop:foo` almost always a bug, if we have to do that our labeling scheme is not consistent then, isn't it?
> 
> I don't agree, and that is why I tried to make the distinction between what a thing is and the role it may play for a specific function.

I'm afraid here we will have to disagree here, I see no benefit on reverting this and having to add dozens of `~ontop:pstate`; that's just bad naming IMHO.

maximedenes:
> On the contrary the renaming of `pstate -> ontop` in the inner vernac loop was done on purpose, as vernacs are sometimes interpreted on top of a lemma stack.

I agree that the change on `there_are_pending_proofs` feels bogus. This one is even worse:

```coq
let check_no_pending_proof ~ontop =
  if there_are_pending_proofs ~ontop then
    user_err Pp.(str "Command not supported (Open proofs remain)")
```

this certainly does not check if there is no pending proof on top of the current state, but in the current state.

ejgallego:
> I agree that the change on `there_are_pending_proofs` feels bogus.

Why? It checks if you are on top of some lemmas or for the contrary the stack is empty.

> this certainly does not check if there is no pending proof on top of the current state, but in the current state.

I am not sure I follow, this checks if ontop is different from none, that is to say, the the stack is not empty.

ejgallego:
Anyways the key point here is the naming of the state in `interp_expr`, the rest just follows after that choice is made.

maximedenes:
> the rest just follows after that choice is made.

I think this is were the problem is. Names should be picked in a way that describes their use by the function, not just to minimize `~foo:bar` occurrences IMHO.

maximedenes:
> I do consider having to write `~ontop:foo` almost always a bug

For example, does this mean that `there_are_pending_proofs` should not be used? Why so?

maximedenes:
> > I do consider having to write `~ontop:foo` almost always a bug
> 
> For example, does this mean that `there_are_pending_proofs` should not be used? Why so?

Ah no, sorry, I see what you mean, it is casting a `pstate` to `ontop` that you think reveals a problem.

But anyway, if we want meaningful name, I'd expect the inner loop to define `ontop`, and `there_are_pending_proofs` and co to take `pstate`.

ejgallego:
> I think this is were the problem is. Names should be picked in a way that describes their use by the function, not just to minimize `~foo:bar` occurrences IMHO.

If you read above you see that this choice does follow from the actual commands interpreting the vernacs:
```
val start_deriving        : ontop:Lemmas.t option
val do_generate_principle : ontop:Lemmas.t option
val make_graph            : ontop:Lemmas.t option
val declare_relation      : ontop:Lemmas.t option
val add_setoid            : ontop:Lemmas.t option
val add_morphism_infer    : ontop:Lemmas.t option
val add_morphism          : ontop:Lemmas.t option
val new_instance          : ontop:Lemmas.t option
val context               : ontop:Lemmas.t option
val do_assumptions        : ontop:Lemmas.t option
val declare_assumption    : ontop:Lemmas.t option
val do_definition         : ontop:Lemmas.t option
val do_fixpoint           : ontop:Lemmas.t option
val do_cofixpoint         : ontop:Lemmas.t option
val declare_fixpoint      : ontop:Lemmas.t option
val declare_cofixpoint    : ontop:Lemmas.t option
val declare_definition    : ontop:Lemmas.t option
val declare_fix           : ontop:Lemmas.t option
val start_lemma           : ontop:t option
```
whatever is named there, it should be used in the upper layers. I think that ontop is not great, however pstate is _wrong_

> But anyway, if we want meaningful name, I'd expect the inner loop to define `ontop`, and `there_are_pending_proofs` and co to take `pstate`.

pstate is wrong, as they don't take a proof stake, but a stack of open proofs about constants.

I am against using "state" in names as I said previously, I'd rather use `proof` than `pstate`.

ejgallego:
> Ah no, sorry, I see what you mean, it is casting a `pstate` to `ontop` that you think reveals a problem.
> 
> But anyway, if we want meaningful name, I'd expect the inner loop to define `ontop`, and `there_are_pending_proofs` and co to take `pstate`.

Not sure I follow here, if they have the same type why they should not use the same naming?

It is me who introduced `pstate` and I do regret is so much XD

maximedenes:
> if they have the same type why they should not use the same naming?

Do you give the same name to all integers?

ejgallego:
> Do you give the same name to all integers?

I do give the same name `~poly` to the boolean that represents polymorphic status for a constant, for example.

Here `ontop` means a stack in all cases, not a state. So you ask "are there open proofs in the stack"?

maximedenes:
> Here `ontop` means a stack in all cases, not a state. So you ask "are there open proofs in the stack"?

Then I find `ontop` a bad name for the current stack of proofs. How about `pstack`? Or `proofs` simply?

gares:
FTR, the large number of overlays is due to the renaming.
Please undo this renaming so that we can eventually what this PR is about.

ejgallego:
Going `ontop -> lemmas` in the interpretation loop would be (a very minor) improvement, but you would still have some ugly patches. So you gain, you lose.

IMHO the real problem here is that the vernac typing is not granular enough, so we should solve that problem and naming would become clearer.

maximedenes:
> IMHO the real problem here is that the vernac typing is not granular enough, so we should solve that problem and naming would become clearer.

So maybe then, as @gares suggested, let's keep this PR focused on the real change, and address the renaming separately (by fixing the underlying vernac typing issue you mention).

ejgallego:
> Then I find `ontop` a bad name for the current stack of proofs. How about `pstack`? Or `proofs` simply?

`lemmas` would work, but that doesn't go well with the role they play, for example in this signature:
```
let vernac_definition ~atts ~ontop discharge kind ({loc;v=id}, pl) def =
```
so here is the tension to resolve; you choose to name the stack `lemmas` for example, however it is always used as `ontop`.

SkySkimmer:
>If you read above you see that this choice does follow from the actual commands interpreting the vernacs:

Ultimately these commands won't be taking a proof stack / lemma argument right? It seems like they should just be returning one (or optionally returning) and something higher level should be pushing what it returns to the stack.
I think that's what "IMHO the real problem here is that the vernac typing is not granular enough, so we should solve that problem and naming would become clearer." means?

For instance `new_instance` just pushes the new lemma to the stack (if opening a proof) without really reading it (it only reads enough to push), and in fact is bugged: 
~~~coq
Set Nested Proofs Allowed. (* unset refine instance mode doesn't work but should morally *)
Class Bar := bar {}.
Module M.
Lemma foo : True.
Proof.
  Instance boz : Bar := {}.
  Fail exact I.
~~~

ejgallego:
> So maybe then, as @gares suggested, let's keep this PR focused on the real change, and address the renaming separately (by fixing the underlying vernac typing issue you mention).

This was what the original submission did, but as @gares pointed out the naming was a total mess; so I did agree with him and revised the PR.

ejgallego:
> For instance `new_instance` just pushes the new lemma to the stack (if opening a proof) without really reading it (it only reads enough to push), and in fact is bugged:

Yup @SkySkimmer , better typing would allow us to remove the `ontop` mess and have the pushing done in the top-level handler.

ejgallego:
> FTR, the large number of overlays is due to the renaming.

You are almost 100% wrong; why do you say false stuff like that? Undoing the renaming would only allow us to remove one overlay.

> Please undo this renaming so that we can eventually what this PR is about.

?

gares:
Do you mean:
> In Lemmas.mli there are API getting arguments of type Lemma.t with the label ~pstate, eg the ones for     saving a proof. I guess they should take a ~lemma now?

Fixing the issue I raise does not mean propagating the renaming all over the place.
It is your desire to never write `~foo:bar` that, I guess, provoked it.

gares:
> You are almost 100% wrong; why do you say false stuff like that? Undoing the renaming would only allow us to remove one overlay.

You re right, the elpi one just because of that while the other one are to account for `Proof_global.t` -> `Lemma.t`.

maximedenes:
So @gares, to be concrete, what would suggest @ejgallego to do w.r.t. `pstate` vs `ontop`, to minimize the diff noise?

gares:
Use `~ontop` in the API that needs to be changed (eg Lemmas), but don't rename `pstate` to `ontop` in the rest of the code (eg vernacentries) and live with `~ontop:pstate`.

ejgallego:
> For instance `new_instance` just pushes the new lemma to the stack (if opening a proof) without really reading it (it only reads enough to push), and in fact is bugged:

Most commands (if not all) are bugged except for the standard definition / lemma. They will be eventually fixed but a lot of work / PRs still lie ahead.

ejgallego:
Rebased and fixed conflicts.

Regarding the requested renaming I tried, but I have to digress on this one [for the reasons explained above, basically this was an interim API so I am not so concerned about the noise, IMHO this was an interim API and consistency is more important in my view] The result after renaming looks pretty bad IMHO so I'd rather avoid it. The problems with there are pending proofs etc... are fixed in subsequent PRs that refine types. I wish I could squash this commit with #9129 , however it is too late.

This PR takes an opinionated view indeed, is that a vernacular command does run "ontop" of a stack of lemmas.

A few more comments:
- contrary to what it was said, overlays are unavoidable, this was indeed a reason to aim for a more consistent naming as I had to to them anyways [I dunno why to say something that is not true in support of an argument, to me it seems like you have decided on the argument regardless of technical considerations and look for support post-facto];
- "Please undo this renaming so that we can eventually understand what this PR is about. (EDIT)" however you said before "The actual objective of PR seems reasonable" so I am not sure I follow anymore.

gares:
I was wrong about the overlays, I m sport I rushed to conclusions.

My request for changes is clear, and vernacentries is still noisy.

ejgallego:
> My request for changes is clear, and vernacentries is still noisy.

Is that the only file that you want changed?

gares:
vernacentries is the most visible example of a file that has many syntactic changes and very little semantic ones, if any. This it why I'm taking it ad an example.

To me the only changes this PR should show are in proof_global and lemmas.
All the rest of the changes should be minimized (no renaming other than what strictly necessary)

ejgallego:
Ok, so I find your request pretty fuzzy / unclear then; "minimized" and "naming" are subjective notions, this PR has undergone more than 40 revisions and thus I shall stop here. The changes you propose do have negative impacts in other parts of the code, I feel unable to improve the noise / quality balance. From my understanding of the code, the point of diminishing returns has been reached. 

Many useful comments on naming were implemented, and noise is statistically a small part of the changes; keeping into account that this PR is essentially a refactoring PR that moves a lot of things I do consider the noise to be within acceptable ranges, not taking into account that noise is minor concern when compared to code readability.

Significant work has been built on top of this PR, validating the current approach (~ 50 commits); I shall note that a few of them do significantly improve the naming by means of improved typing.

ejgallego:
I was not aware you moved the milestone here @vbgl without any discussion, for many reasons, not having this PR in 8.10 is the wrong thing to do.

maximedenes:
> I was not aware you moved the milestone here @vbgl without any discussion, for many reasons, not having this PR in 8.10 is the wrong thing to do.

I think we should discuss the milestone only if we manage to unblock this PR...

ejgallego:
> I think we should discuss the milestone only if we manage to unblock this PR...

The PR has been ready to merge for a while. I can't take "diff noise" to be a blocking concern over real code consistency. At least whitespace is something the programmer actually sees.

Anyways, I am just letting the RM know, as it is my "duty" I guess.

gares:
I think that the code in vernacentries before this patch is better than the one after this patch. Just look at the first two hunks in that file: the new name of the function argument is not meaningful to me,and I don't think it is necessary to operate these renamings in order to move some code in the Lemmas module, not to rename labels in its MLI file.

ejgallego:
> I think that the code in vernacentries before this patch is better than the one after this patch.

To me it is not; I do prefer the new code. Naming is a subjective matter, the news hunks are better to me.

Then we face person A think naming A is better person B thinks naming B is better. As in this case I am extremely short on time, and I think my naming is better and it is supported by lots of empirical evidence I have no incentive not to stick to it.

So if you are willing to block a critical PR for 8.10 [as it included among other things the relevant changes entry for the new API] and waste very precious developer time that's your call.

> move some code in the Lemmas module

The patch does more than that, as you can see if you follow the patch series.

maximedenes:
> So if you are willing to block a critical PR for 8.10 [as it included among other things the relevant changes entry for the new API] and waste very precious developer time that's your call.

I'm pretty sure we could split the renaming and the core critical change, if we wanted to make sure we merge the interesting bits quickly.

ejgallego:
> I'm pretty sure we could split the renaming and the core critical change, if we wanted to make sure we merge the interesting bits quickly.

You assume I have infinite amounts of free time which I do not. Also, the old naming was a bug, so I am not going to approve it.

The sad part is would we have processed PRs quicker; we would be already at a point in the refactoring where these labels would have gone away.

maximedenes:
> You assume I have infinite amounts of free time which I do not.

I don't assume that. I may even give a try to splitting the renaming next week.

ejgallego:
> I'm pretty sure we could split the renaming and the core critical change, if we wanted to make sure we merge the interesting bits quickly.

That's what the first version of the PR did, and it was bad. I am repeating myself all the time anyways, I feel like I have little to say here.

Perfect is the enemy of the good, and my impression based on the work done in that area is that it is futile to try to improve the status of the codebase.

There is a total disregard for the time of other developers, and people seem to enjoy blocking things and doing arbitrary demands, even when they are not the maintainers / they didn't participate on the roadmap of that particular subsystem; it is very hard to achieve even a bit of the amount of the required development autonomy to perform any significant evolution.

maximedenes:
> people seem to enjoy blocking things and doing arbitrary demands, even when they are not the maintainer

I'm the secondary maintainer of vernac, and I don't like the renamings I see there, as explained above already.

> There is a total disregard for the time of other developers, and people seem to enjoy blocking things and doing arbitrary demands

If this is your way of interpreting our feedback, I'd tend to think it is your problem. I don't see why we would enjoy blocking this change, which we agreed to from the start, if you exclude the renaming part.

ejgallego:
> I don't assume that. I may even give a try to splitting the renaming next week.

You are coming with a proposal that was already discussed here; I appreciate your role in trying to mediate on this but are IMHVO missing the issue here.

You spending time on re-doing the first version of this PR is IMHO a bad choice, first of all because the version is still available on the top of the page, and second because the version had significant problems many of the pointed out by Enrico, which led to the current one.

This is a recurring problem due I guess to the github interface, and is that people get lost on the discussions and things are rediscussed over and over again.

The real problem here is very different:
- a) we have a longstanding problem in the overall system architecture since the STM / Proof engine were introduced without any kind of coordination. Additionally to the own problems these subsystem have, their combination brought new ones.
- b) there has been an ongoing effort to untangle things and make the combination interact better since a few releases, for example, by removing layering problems [which created the `vernac` dir] etc...
- c) these are real problems that impact users, and cleaning up the codebase is a long-term project
- d) this PR is the second part of the critical #9129 , which was already a big pain to get in [6 months and hundreths of revisions] the PR is good, and nobody cares if a name is done one way or another. This doesn't impact users, and only marginally to developers.
- e) the blockage has a terrible tradeoff, in addition to all the stress and unnecessary discussion, it means that I could not progress with the cleanup plan. For example, I estimate that if this PR would not have been blocked I would already have reached the point where the API for constants with multiple proofs attached as suggested by @mattam82 would be in draft PR status.

And here you have the real problem, it was considered that aiming for some renamings (which by the way I think are worse) was a better tradeoff than over 20 PRs of improvements.

So with this kind of tradeoff function it is not a surprise that the project is at the point it is for example in terms of UI support. But indeed, I guess we are facing an extremely grave management problem and that is out of scope for the current PR.

ejgallego:
> if you exclude the renaming part.

This is where I have a problem with the cost/benefit ratio. There is no model as discussed in this PR where the benefit or redoing the renaming, which as I pointed out leads to worse code, justifies the cost of doing it.

And indeed you support my point of view, even if you found out the new name not to be very good [the old one was even worse but anyways], you _didn't block_ the PR on that.

maximedenes:
In any case, I'll try to improve the situation next week. That's all I can offer.

gares:
I feel like the aggressiveness in this thread has reached a level that is totally unprofessional, and I'm not spending any more time trying to point out what is wrong. I'm not the only reviewer / code owner, so someone else will eventually step in and review the code. Good luck.

ejgallego:
Look folks, the last thing I want is more fighting and name-calling.

On the contrary, I just want to make progress on these structural PRs that have been going on for 3 years, and are absolutely required in order to implement some high-priority features. So far, this work has been the most painful thing I have done in my programmer's life.

While for sure I could improve the implementation and communication strategy, I can safely say that the amount of negativity and rejection this kind of work has received has been staggering and pretty. It is very frustrating by a large variety of reasons.

@gares, I am sorry you felt attacked in this PR, rest assured that I was not referring specifically this particular PR when writing, but to the overall frustration accumulated over many years of having to fight negative and unfair/inaccurate reactions to my changes. My apologies, your feedback in this PR is appreciated.

Trust me folks: we are complicating our lives a lot and making hard the easy.

---

Going back to this PR, I expect it would take a week to merge or so; a large motivation was to ship a more coherent API in 8.10 and provide many interesting fixes to the overall proof management. I guess that point is moot now.

I would say that in 8.11 naming shouldn't be too relevant here as hopefully we would reify vernac interpretation, I don't dare to do any kind of plan as of now tho.

What to do? The only issue remaining is the one of naming, which is highly subjective; saying that something "is wrong" is IMO too strong in this case.

@maximedenes : I don't think it is worth your time to work on this PR to remove some internal renamings, specially if you consider that these names should be very short lived as many of them will go away in a future PR (you can look ahead at the PR stack if you want). Moreover, the current setup has a delicate balance: when you improve one part, others become worse.

mattam82:
(As a Code of Conduct enforcement team member)
Regardless of the technical issue being debated, and naming is included, avoid voicing general frustration while discussing a specific technical issue. You are very welcome to open a specific issue or contact the team coordinator to voice them. When technical conflicts arise and linger, please notify third party developers early on to widen the debate.

ejgallego:
@ppedrot , modulo overlays, this should be ready to merge, after #10050 the original issues on naming should not exist anymore. This PR is very critical as it is the enabling step to remove important proof data from being stored in closures. It also includes important API documentation missing from 8.10

In any case cc: @gares @maximedenes @SkySkimmer if they want to have a look.

proofs/proof_global.mli
```diff
@@ -13,18 +13,16 @@
...
 
 type t
-type stack
 
-val get_current_pstate : stack -> t
-
-val get_current_proof_name : t -> Names.Id.t
-val get_current_persistence : t -> Decl_kinds.goal_kind
-val get_all_proof_names : stack -> Names.Id.t list
+(* Should be moved into a proper view *)
```
SkySkimmer: Maybe we should expose the record `type t = private { ... }`. We already have getters for every field except endline_tactic, and it may be cleaner to access that through a getter or projection than passing it as argument to with_proof (which causes us to have to ignore the argument in most calls).

ejgallego: Something like this is done in a later pr, which specifically focuses on moving and cleaning up fields, including some duplication. If you don't mind I'd like to address this comment on the upcoming PR.

SkySkimmer: OK

doc/plugin_tutorial/tuto1/src/g_tuto1.mlg
SkySkimmer: @ppedrot might not like that partial application style.

proofs/proof_global.mli
ejgallego: Merci

ejgallego:
Ok, overlays update, two remaining failures also fail on master due to unmerged overlays.

ejgallego:
Ok, all overlays sorted out, remaining failure due to https://github.com/QuickChick/QuickChick/pull/164 not merged, all should be good when that gets merged.

ejgallego:
@ppedrot , all plugins complied finally, so once the heavier CI passes this should be ready!

ppedrot:
Will merge soon if I get no complains from @gares / @maximedenes.

ppedrot:
I have to merge both this PR and #8726. They'll most likely conflict, is there a preferred order of resolution? ping @herbelin.

ejgallego:
@ppedrot I'll be happy to rebase again, go ahead with @herbelin first.

ejgallego:
Indeed @herbelin 's PR interacts with all the chain of PRs refactoring `proof`, however it makes sense for his to go first as given the scope of changes it'll be much easier for the proof tree to adapt to the Discharge changes.

ejgallego:
Failures due to unmerged overlays for #8726 , must re-run once these are merged:
- https://github.com/LPCIC/coq-elpi/pull/62
- https://github.com/Mtac2/Mtac2/pull/210

ejgallego:
Ready now @ppedrot !

ejgallego:
Thanks to all! IMHO we are past the most painful parts of the `proof` refactoring!

maximedenes:
> Will merge soon if I get no complains from @gares / @maximedenes.

I didn't have time to look at it, so don't take this as a complaint against merging, but IMHO week-ends should not be counted as effective time to give feedback on PRs...

ejgallego:
@maximedenes I guess @ppedrot counted since this message https://github.com/coq/coq/pull/9566#issuecomment-499033136 but indeed it is not too late, please review and we can amend / revert if needed.

PR #9567, opened 2019-02-12
[vernac] Unify declaration hooks.

ejgallego:
Supersedes #8718.

### DIFF
--- a/doc/plugin_tutorial/tuto1/src/simple_declare.ml
+++ b/doc/plugin_tutorial/tuto1/src/simple_declare.ml
@@ -1,5 +1,5 @@
 (* Ideally coq/coq#8811 would get merged and then this function could be much simpler. *)
-let edeclare ident (_, poly, _ as k) ~opaque sigma udecl body tyopt imps hook =
+let edeclare ?hook ident (_, poly, _ as k) ~opaque sigma udecl body tyopt imps =
   let sigma = Evd.minimize_universes sigma in
   let body = EConstr.to_constr sigma body in
   let tyopt = Option.map (EConstr.to_constr sigma) tyopt in
@@ -9,16 +9,17 @@ let edeclare ident (_, poly, _ as k) ~opaque sigma udecl body tyopt imps hook =
      (Option.List.cons tyopt [body]) in
   let sigma = Evd.restrict_universe_context sigma uvars in
   let univs = Evd.check_univ_decl ~poly sigma udecl in
+  let uctx = Evd.evar_universe_context sigma in
   let ubinders = Evd.universe_binders sigma in
   let ce = Declare.definition_entry ?types:tyopt ~univs body in
-  DeclareDef.declare_definition ident k ce ubinders imps ~hook
+  let hook_data = Option.map (fun hook -> hook, uctx, []) hook in
+  DeclareDef.declare_definition ident k ce ubinders imps ?hook_data
 
 let packed_declare_definition ~poly ident value_with_constraints =
   let body, ctx = value_with_constraints in
   let sigma = Evd.from_ctx ctx in
   let k = (Decl_kinds.Global, poly, Decl_kinds.Definition) in
   let udecl = UState.default_univ_decl in
-  let nohook = Lemmas.mk_hook (fun _ x -> ()) in
-  ignore (edeclare ident k ~opaque:false sigma udecl body None [] nohook)
+  ignore (edeclare ident k ~opaque:false sigma udecl body None [])
 
 (* But this definition cannot be undone by Reset ident *)
--- a/plugins/funind/indfun_common.ml
+++ b/plugins/funind/indfun_common.ml
@@ -129,7 +129,7 @@ let get_locality = function
 | Local -> true
 | Global -> false
 
-let save with_clean id const ?hook (locality,_,kind) =
+let save with_clean id const ?hook uctx (locality,_,kind) =
   let fix_exn = Future.fix_exn_of const.const_entry_body in
   let l,r = match locality with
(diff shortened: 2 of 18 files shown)


ejgallego:
Overlays:
- https://github.com/mattam82/Coq-Equations/pull/173
- https://github.com/Mtac2/Mtac2/pull/155
- https://github.com/ppedrot/ltac2/pull/106

doc/plugin_tutorial/tuto1/src/simple_declare.ml
```diff
@@ -9,16 +9,17 @@ let edeclare ident (_, poly, _ as k) ~opaque sigma udecl body tyopt imps hook =
...
-  DeclareDef.declare_definition ident k ce ubinders imps ~hook
+  DeclareDef.declare_definition ident k ce ubinders imps ~hook_data:(hook,uctx,[])
 
 let packed_declare_definition ~poly ident value_with_constraints =
   let body, ctx = value_with_constraints in
   let sigma = Evd.from_ctx ctx in
   let k = (Decl_kinds.Global, poly, Decl_kinds.Definition) in
   let udecl = UState.default_univ_decl in
-  let nohook = Lemmas.mk_hook (fun _ x -> ()) in
+  let nohook = Lemmas.mk_hook (fun _ _ _ x -> ()) in
```
SkySkimmer: Should we keep messing around with empty hooks in the plugin tutorial? I'm inclined to just remove the hook argument to edeclare.

ejgallego: I was about to make it optional, but I hesitated due to the fact that people may forget them. As you wish.

ejgallego: Anyways if everything goes well we will greatly refactor hooks / terminators so this won't stay for long.

plugins/funind/indfun_common.ml
```diff
@@ -144,7 +144,9 @@ let save with_clean id const ?hook (locality,_,kind) =
 	(locality, ConstRef kn)
   in
   if with_clean then Proof_global.discard_current ();
-  Lemmas.call_hook ?hook ~fix_exn l r;
+  (* XXX: need help *)
+  let uctx = UState.empty in
```
SkySkimmer: Make the `save` function take a ustate? Or make the caller (there's only 1) call the hook themselves (ie 6e8ee9cd4c3da1d743664c994f7c61d2009926c4)

ejgallego: The whole `save` function should be indeed removed I think; but let's add the argument for now.

ejgallego:
Ready to go, any of the vernac maintainers may assign @mattam82 @maximedenes

vernac/lemmas.mli
```diff
@@ -13,23 +13,34 @@ open Decl_kinds
...
+ * etc...
+ *
+ * Here, we use an extended hook type suitable for obligations /
+ * equations.
+ *)
+(** [hook_type] passes to the client:
+    - [ustate]: universe constraints obtained when the term was closed
+    - [(n1,t1),...(nm,tm)]: XXXX
+    - [locality]: Locality of the original declaration
+    - [ref]: identifier of the origianl declaration
```
mattam82: [(n1, t1)] .. association list between obligation name and the corresponding defined term (might be a constant, but also an arbitrary term in the Expand case of obligations)
[ref] global reference of the declared definition

The first mapping could be generalized to a mapping between goal id (however we chose to encode it) and the defined constants corresponding to it (e.g. if Qed were to define all the initial goals as constants for example)

mattam82 (approved):
I like where this is going, the uniformity looks good too.

vernac/lemmas.mli
ejgallego: Thanks @mattam82 , updating. Indeed I think this can be made general-purpose.

ejgallego:
> I like where this is going, the uniformity looks good too.

Thanks for the review; rebased and ready to merge.

I think I'd like to push a couple more of extra PRs on cleanup before starting to sketch the API you discussed, in particular wrt the save path.

ejgallego:
The failure is spurious due to a missing overlay from other PR.

ejgallego:
Rebased PR and 4 overlays.

ejgallego:
Ready for merge again.

ejgallego:
@mattam82 it seems Maxime is busy, would you mind taking over this and merging? Thanks!

ejgallego:
By the way @mattam82 I have a patch that integrates the obligation terminators with the main ones, it depends on this and #9129 [so it will take a while] but IMHO it is a nice cleanup and should leave things in good state to attempt for a more flexible API.

ejgallego:
Rebased PR and 3 overlays.

ejgallego:
> Rebased PR and 3 overlays.

Ready for merge again, CI failure is spurious.

ejgallego:
Rebased PR.

ejgallego:
CI green.

maximedenes (approved):
I'm not sure I like the change of line-breaking style in the interfaces, but let's merge meanwhile and see that later.

ejgallego:
Thanks

ejgallego:
> I'm not sure I like the change of line-breaking style in the interfaces, but let's merge meanwhile and see that later.

I have no strong opinion on it, but it is a style that is much better than the one we have [which is no style and really messy lines], works well for diffs, and it is the one adopted by some other projects including `ocamlformat`, `dune`, etc... so given that it is becoming a short of "standard" in particular a style produced by automatic formatting tools I am OK using it.

ejgallego:
cc: #9572

Issue #8442, opened 2018-09-09
Use promote rule to generate doc/sphinx/README.rst


Zimmi48:
`doc/sphinx/README.rst` contains the following comment:

```rst
..
   README.rst is auto-generated from README.template.rst and the coqrst docs;
   use ``doc/tools/coqrst/regen_readme.py`` to rebuild it.
```

However, people usually miss this so the best way to avoid any sync problem would be to add a test to the test-suite like we did for some extracted ML files.

JasonGross:
I think `doc/sphinx/README.rst` should also be made read-only, if it is not already.

JasonGross:
`chmod +w; modify; chmod -w`
The important thing is that the checked-in version is read-only, so there's an extra barrier to manual modification.  I noticed that CompCert (or maybe it was VST) does this, and found it useful.

SkySkimmer:
This seems more like a lint thing than a test suite thing.

Zimmi48:
@JasonGross Why not? Should we do the same for the above mentioned extracted ML files? @SkySkimmer Sure, any solution is good as long as it ends up being run by CI.

JasonGross:
@Zimmi48 Yes, I'd suggest doing this for any file for which it would be a mistake to commit an edit-by-hand.

Zimmi48:
This seems like something that should be doable with dune, without going in the test-suite. The reason to put the test of the above mentioned extracted ML files in the test-suite is to avoid creating a too long dependency chain to build said ML files. This concern does not apply in the case of the README.

ejgallego:
Bug added to #8052

Zimmi48:
This has been addressed in #9568.

PR #9568, opened 2019-02-13
Add test that we regenerated doc/sphinx/README.rst to linter

SkySkimmer:
We change regen_readme such that when given an argument it outputs
there instead of overwriting the readme. Then the linter calls
regen_readme with a temporary file path and diffs the result with the
readme.

Prompted by me noticing I forgot to regen in #9553.

### DIFF
--- a/doc/sphinx/dune
+++ b/doc/sphinx/dune
@@ -1,1 +1,8 @@
 (dirs :standard _static)
+
+(rule (targets README.gen.rst)
+ (deps (source_tree ../tools/coqrst) README.template.rst)
+ (action (run ../tools/coqrst/regen_readme.py %{targets})))
+
+(alias (name refman-html)
+  (action (diff README.rst README.gen.rst)))
--- a/doc/sphinx/README.rst
+++ b/doc/sphinx/README.rst
@@ -214,7 +214,7 @@ In addition to the objects above, the ``coqrst`` Sphinx plugin defines the follo
 
     Example::
 
-       .. coqtop:: in reset undo
+       .. coqtop:: in undo
 
           Print nat.
           Definition a := 1.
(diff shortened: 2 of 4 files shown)


SkySkimmer:
The lint job doesn't work because it runs on a stripped-down image which doesn't have sphinx, and the regen script imports sphinx.
Should I try to figure out how to install sphinx on the small image, or just use the standard coq image?
cc @ejgallego

Zimmi48:
You could move this "lint" job in the refman build.
BTW, this PR fixes #8442.

ejgallego:
> You could move this "lint" job in the refman build.

Indeed you could move this to the refman build, doesn't it look more logical to do this test each refman build?

SkySkimmer:
I removed the linter code and added a dune rule (promotion capable). I'll only add a test to the makefile based refman build if someone insists.

Zimmi48:
@SkySkimmer The new file is picked up by the Sphinx build as a normal source file. You should add it here:

https://github.com/coq/coq/blob/77b454e5ab8698f0d87bdf2eb32b48ab998ba590/doc/sphinx/conf.py#L135-L146

doc/sphinx/dune
```diff
@@ -1 +1,8 @@
 (dirs :standard _static)
+
+(rule (targets README.gen.rst)
+ (deps (source_tree ../tools/coqrst) README.template.rst)
+ (action (run ../tools/coqrst/regen_readme.py %{targets})))
```
ejgallego: Not sure if this will run on Windows.

ejgallego (approved):
Leaving this to @coq/doc-maintainers

Zimmi48 (approved):

PR #9574, opened 2019-02-14
[Namegen] Redesign of name-avoidance strategy in `next_global_ident_away`

### LINKED ISSUE #9323: Names generated by `Instance` can clash

vbgl:
Fixes #9323.

Notice that the `is_global` function, used in `next_ident_away_in_goal` is unchanged. Changing it would break a few third-party developments which rely too much on generated names.

A few questions:

  - How to add an example with two files in the test-suite (the bug does not show up when the two modules are defined in the same file)?
  - Is the termination check in the loop efficient enough?

### DIFF
--- a/engine/namegen.ml
+++ b/engine/namegen.ml
@@ -320,6 +320,24 @@ let next_name_away_in_goal na avoid =
     | Anonymous -> default_non_dependent_ident in
   next_ident_away_in_goal id avoid
 
+(** Check whether [id] is already defined in the current namespaces,
+    i.e., in any of the enclosing sections in the current module path. *)
+let is_defined_here id =
+  (* Check if [path.id] is defined *)
+  let exists path =
+    match make_qualid (DirPath.make path) id |> Nametab.locate with
+    | exception Not_found -> false
+    | _ -> true
+  in
+  let short = Lib.cwd_except_section () |> DirPath.repr in
+  let long = Lib.cwd () |> DirPath.repr in
+  (* Test every path between [short], the current module path,
+     and [long], the current module-and-section path. *)
+  let rec loop path =
+    exists path
+    || not (List.equal Id.equal path short) && loop (List.tl path)
+  in loop long
+
 (* 3- Looks for next fresh name outside a list that is moreover valid
    as a global identifier; the legacy algorithm is that if the name is
    already used in the list, one looks for a name of same base with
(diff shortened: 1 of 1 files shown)


engine/namegen.ml
```diff
@@ -320,6 +320,24 @@ let next_name_away_in_goal na avoid =
     | Anonymous -> default_non_dependent_ident in
   next_ident_away_in_goal id avoid
 
+(** Check whether [id] is already defined in the current namespaces,
+    i.e., in any of the enclosing sections in the current module path. *)
+let is_defined_here id =
```
SkySkimmer: Why not use `Global.exists_objlabel`? Or move the logic in `Declare.exists_name` to here? I don't really understand the difference in behaviour though.

vbgl: Where can I find the documentation of these functions?

SkySkimmer: >Where can I find the documentation of these functions?

You look at the implementation

vbgl: Thanks for the pointers. The following patch seems good enough to address the issue:

~~~
--- a/engine/namegen.ml
+++ b/engine/namegen.ml
@@ -329,7 +329,7 @@ let next_name_away_in_goal na avoid =
 
 let next_global_ident_away id avoid =
   let id = if Id.Set.mem id avoid then restart_subscript id else id in
-  let bad id = Id.Set.mem id avoid || is_global id in
+  let bad id = Id.Set.mem id avoid || Global.exists_objlabel (Label.of_id id) in
   next_ident_away_from id bad
 
 (* 4- Looks for next fresh name outside a list; if name already used,
~~~

Let me run the CI to see if something would break.

herbelin:
> How to add an example with two files in the test-suite (the bug does not show up when the two modules are defined in the same file)?

If ever useful, the `test-suite/misc` directory was done for this kind of purpose (it already has stand-alone directory with custom compilation rules).

SkySkimmer:
We can also add stuff in prerequisites/

herbelin:
Indeed.

ppedrot:
Closing this PR in favour of #9577.

vbgl:
Closing. Subsumed by #9577.

PR #9576, opened 2019-02-14
[library] Remove `-boot` option.

### LINKED ISSUE #9575: -boot cleanup for test suite Makefile
After #9560 -boot should be exactly -q + allow defining modules in Coq. The test suite has `coqc_boot := $(BIN)coqc -coqlib $(LIB) -boot -q -test-mode -R prerequisite TestSuite $(COQFLAGS)` but the -boot isn't actually useful. Let's remove it.

ejgallego:
The `-boot` option was used to:

- suppress loading of the rc_file
- allow to save modules with prefix `Coq`

There is no good reason disable saving of modules with `Coq` prefix by
default, thus we remove this option.

Fixes: #9575
D̶e̶p̶e̶n̶d̶s̶ ̶o̶n̶:̶ ̶#̶9̶5̶6̶0̶ ̶

### DIFF
--- a/checker/checker.ml
+++ b/checker/checker.ml
@@ -192,7 +192,6 @@ let print_usage_channel co command =
 \n  -coqlib dir            set coqchk's standard library location\
 \n  -where                 print coqchk's standard library location and exit\
 \n  -v                     print coqchk version and exit\
-\n  -boot                  boot mode\
 \n  -o, --output-context   print the list of assumptions\
 \n  -m, --memory           print the maximum heap size\
 \n  -silent                disable trace of constants being checked\
@@ -320,8 +319,6 @@ let explain_exn = function
 let deprecated flag =
   Feedback.msg_warning (str "Deprecated flag " ++ quote (str flag))
 
-let boot_opt = ref false
-
 let parse_args argv =
   let rec parse = function
     | [] -> ()
@@ -358,7 +355,6 @@ let parse_args argv =
     | ("-?"|"-h"|"-H"|"-help"|"--help") :: _ -> usage ()
 
     | ("-v"|"--version") :: _ -> version ()
-    | "-boot" :: rem -> boot_opt := true; parse rem
     | ("-m" | "--memory") :: rem -> Check_stat.memory_stat := true; parse rem
     | ("-o" | "--output-context") :: rem ->
         Check_stat.output_context := true; parse rem
--- a/man/coqide.1
+++ b/man/coqide.1
@@ -100,15 +100,6 @@ Skip loading of rcfile.
 Set the rcfile to
 .IR f .
 .TP
-.B \-batch
-Batch mode (exits just after arguments parsing).
-.TP
-.B \-boot
-Boot mode (implies
-.B \-q
-and
-.BR \-batch ).
-.TP
 .B \-emacs
 Tells Coq it is executed under Emacs.
 .TP
--- a/toplevel/coqargs.ml
+++ b/toplevel/coqargs.ml
@@ -40,8 +40,6 @@ type native_compiler = NativeOff | NativeOn of { ondemand : bool }
 
 type t = {
 
-  boot : bool;
-
   load_init   : bool;
   load_rcfile : bool;
   rcfile      : string option;
@@ -92,8 +90,6 @@ let default_native =
   else NativeOff
 
 let default = {
-
-  boot = false;
 
   load_init   = true;
   load_rcfile = true;
@@ -178,6 +174,10 @@ let set_color opts = function
 let warn_deprecated_inputstate =
   CWarnings.create ~name:"deprecated-inputstate" ~category:"deprecated"
          (fun () -> Pp.strbrk "The inputstate option is deprecated and discouraged.")
+
+let warn_deprecated_boot =
+  CWarnings.create ~name:"deprecated-boot" ~category:"noop"
+         (fun () -> Pp.strbrk "The -boot option is deprecated, please use -q and/or -coqlib options instead.")
 
 let set_inputstate opts s =
   warn_deprecated_inputstate ();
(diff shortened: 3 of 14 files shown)


man/coqide.1
```diff
@@ -103,12 +103,6 @@ Set the rcfile to
 .B \-batch
 Batch mode (exits just after arguments parsing).
```
SkySkimmer: I think coqide doesn't actually work with -batch, at least I get "coqtop died badly" when I try ;)

test-suite/Makefile
```diff
@@ -578,7 +578,7 @@ ide : $(patsubst %.fake,%.fake.log,$(wildcard ide/*.fake))
 	@echo "TEST      $<"
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
-	  $(BIN)fake_ide $< "-coqlib $(LIB) -boot -async-proofs on -async-proofs-tactic-error-resilience off -async-proofs-command-error-resilience off $(call get_coq_prog_args,"$<")" 2>&1; \
+	  $(BIN)fake_ide $< "-coqlib $(LIB) -async-proofs on -async-proofs-tactic-error-resilience off -async-proofs-command-error-resilience off $(call get_coq_prog_args,"$<")" 2>&1; \
```
SkySkimmer: No -q needed here?

toplevel/coqargs.ml
```diff
@@ -459,7 +459,7 @@ let parse_args ~help ~init arglist : t * string list =
       { oval with batch = true }
     |"-test-mode" -> Flags.test_mode := true; oval
     |"-beautify" -> Flags.beautify := true; oval
-    |"-boot" -> { oval with boot = true; load_rcfile = false; }
+    |"-boot" -> warn_deprecated_boot (); oval
```
SkySkimmer: Shouldn't we still set -q while in the deprecation phase?

test-suite/Makefile
```diff
@@ -38,11 +38,11 @@ LIB := ..
 BIN := $(shell cd ..; pwd)/bin/
 COQFLAGS?=
 
-coqc_boot := $(BIN)coqc -coqlib $(LIB) -boot -q -test-mode -R prerequisite TestSuite $(COQFLAGS)
+coqc_boot := $(BIN)coqc -coqlib $(LIB) -q -test-mode -R prerequisite TestSuite $(COQFLAGS)
 coqc := $(BIN)coqc -coqlib $(LIB) -R prerequisite TestSuite $(COQFLAGS)
```
SkySkimmer: These variables are kind of awkward. _boot doesn't mean -boot (which doesn't exist anymore), it means -test-mode. Also why do we allow loading the rc file in the non _boot variable?

checker/checker.ml
```diff
@@ -191,7 +191,6 @@ let print_usage_channel co command =
 \n\
 \n  -where                 print coqchk's standard library location and exit\
 \n  -v                     print coqchk version and exit\
-\n  -boot                  boot mode\
```
SkySkimmer: Indeed after #9560 -boot on the checker is a noop.

SkySkimmer (approved):
Fine with me

test-suite/Makefile
ejgallego: Yep this RC-file stuff is strange, I'll add `-q`.

toplevel/coqargs.ml
ejgallego: Ok, why not?

ejgallego:
Comments addressed.

test-suite/Makefile
ejgallego: I'd be great if we could get rid of `-test-mode` indeed.

ejgallego:
Same owner situation than #9560 IMHO:
> So main candidate for owner for this PR (and #9576) is @gares , would you mind being the assignee?
>
> If not maybe @SkySkimmer would be the next logical choice? Also @vbgl may be interested in issues related on how we locate the libraries?

test-suite/Makefile
ejgallego: Solved.

test-suite/Makefile
ejgallego: In fact #9588 does improve this too.

toplevel/coqargs.ml
ejgallego: Resolved.

ejgallego:
Ok, rebased; should CI be happy we can merge I think. I guess @gares or @SkySkimmer would be the right people to assign.

toplevel/coqargs.ml
```diff
@@ -179,6 +175,10 @@ let warn_deprecated_inputstate =
   CWarnings.create ~name:"deprecated-inputstate" ~category:"deprecated"
          (fun () -> Pp.strbrk "The inputstate option is deprecated and discouraged.")
 
+let warn_deprecated_boot =
+  CWarnings.create ~name:"deprecated-boot" ~category:"noop"
+         (fun () -> Pp.strbrk "The -boot is a noop.")
```
gares: I think should be rephrased since -boot has still the effect of not loading the rc file (like passing `-q`).
The message should say to use -q instead.

gares (changes requested):
looks good but for the message of the warning that is incorrect

toplevel/coqargs.ml
ejgallego: Indeed great catch! The first version of this patch made `-boot` a noop but @SkySkimmer suggested indeed to keep it operational, however I didn't update the warning. Rebasing.

gares (approved):

gares (approved):

gares:
Sorry for the noise I was just trying to see what the new button up there does.

ejgallego:
Thanks!

PR #9577, opened 2019-02-14
[Namegen] Use Global.exists_objlabel in `next_global_ident_away`

### LINKED ISSUE #9323: Names generated by `Instance` can clash

vbgl:
Fixes #9323.

Much simpler than the previous attempt (#9574).

### DIFF
--- a/engine/namegen.ml
+++ b/engine/namegen.ml
@@ -329,7 +329,7 @@ let next_name_away_in_goal na avoid =
 
 let next_global_ident_away id avoid =
   let id = if Id.Set.mem id avoid then restart_subscript id else id in
-  let bad id = Id.Set.mem id avoid || is_global id in
+  let bad id = Id.Set.mem id avoid || Global.exists_objlabel (Label.of_id id) in
   next_ident_away_from id bad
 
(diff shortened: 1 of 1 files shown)


maximedenes:
Is `is_global` still used? The name is probably going to make other clients use it incorrectly...

vbgl:
Yes. It is still used is `next_ident_away_in_goal`. Fixing this other function is a different issue (see #9593).

maximedenes (approved):
Thanks!

maximedenes:
@ppedrot do you want to be the assignee?

ppedrot (approved):

PR #9578, opened 2019-02-14
Document primitive integers

### LINKED ISSUE #2409: What does "Register" do?

Original bug ID: BZ#2409
From: @jbapple
Reported version: trunk
CC:   @aspiwack, @pirbo

vbgl:
Closes #2409.

TODO:

  - [x] give details about efficiency and extraction
  - [x] document the `Primitive` vernacular command
  - [x] fix the bug and remove the joke

cc/ @maximedenes

### DIFF
--- a/doc/sphinx/language/gallina-extensions.rst
+++ b/doc/sphinx/language/gallina-extensions.rst
@@ -2260,3 +2260,52 @@ expression as described in :ref:`ltac`.
 This construction is useful when one wants to define complicated terms
 using highly automated tactics without resorting to writing the proof-term
 by means of the interactive proof engine.
+
+.. _primitive-integers:
+
+Primitive Integers
+--------------------------------
+
+The language of terms features 63-bit machine integers as values. The type of
+such a value is *axiomatized*; it is declared through the following sentence
+(excerpt from the :g:`Int63` module):
+
+.. coqdoc::
+
+   Primitive int := #int63_type.
+
+This type is equipped with a few operators, that must be similarly declared.
+For instance, equality of two primitive integers can be decided using the :g:`Int63.eqb` function,
+declared and specified as follows:
+
+.. coqdoc::
+
+   Primitive eqb := #int63_eq.
+   Notation "m '==' n" := (eqb m n) (at level 70, no associativity) : int63_scope.
+
+   Axiom eqb_correct : forall i j, (i == j)%int63 = true -> i = j.
+
+The complete set of such operators can be obtained looking at the :g:`Int63` module.
+
+These primitive declarations are regular axioms. As such, they must be trusted and are listed by the
+:g:`Print Assumptions` command, as in the following example.
+
+.. coqtop:: in reset
+
+   From Coq Require Import Int63.
+   Lemma one_minus_one_is_zero : (1 - 1 = 0)%int63.
+   Proof. apply eqb_correct; vm_compute; reflexivity. Qed.
+
+.. coqtop:: all
+
+   Print Assumptions one_minus_one_is_zero.
+
+The reduction machines (:tacn:`vm_compute`, :tacn:`native_compute`) implement
+dedicated, efficient, rules to reduce the applications of these primitive
+operations.
+
+These primitives, when extracted to OCaml (see :ref:`extraction`), are mapped to
+types and functions of a :g:`Uint63` module. Said module is not produced by
+extraction. Instead, it has to be provided by the user (if they want to compile
+or execute the extracted code). For instance, an implementation of this module
+can be taken from the kernel of Coq.
--- a/doc/sphinx/proof-engine/vernacular-commands.rst
+++ b/doc/sphinx/proof-engine/vernacular-commands.rst
@@ -1250,3 +1250,17 @@ Inlining hints for the fast reduction machines
 
    Due to its internal nature, this command is not for general use. It is meant
    to appear only in standard libraries and in support libraries of plug-ins.
+
+Registering primitive operations
+--------------------------------
+
+.. cmd:: Primitive @ident__1 := #@ident__2.
+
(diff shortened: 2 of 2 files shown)


doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -1227,3 +1227,26 @@ Exposing constants to OCaml libraries
...
    Due to its internal nature, this command is not for general use. It is meant
    to appear only in standard libraries and in support libraries of plug-ins.
+
+   As a special case, when the first segment of :n:`@qualid__2` is :g:`kernel`,
+   the constant is exposed to the kernel. For instance, the `Int63` module
+   features the following declaration:
+
+   .. coqdoc::
+
+   Register bool as kernel.ind_bool.
```
Zimmi48: Incorrect indentation.

maximedenes (approved):
Looks very good! I think we can address the todos later (in particular, when the story for extraction and linking becomes better).

Zimmi48:
>I think we can address the todos later

In that case, don't close the issue.

doc/sphinx/language/gallina-extensions.rst
```diff
@@ -2260,3 +2260,52 @@ expression as described in :ref:`ltac`.
...
+Primitive Integers
+--------------------------------
+
+The language of terms features 63-bit machine integers as values. The type of
+such a value is *axiomatized*; it is declared through the following sentence
+(excerpt from the :g:`Int63` module):
+
+.. coqdoc::
+
+   Primitive int := #int63_type.
```
Zimmi48: I wonder about the choice to name this type `int`. It doesn't convey the 63-bit nature. If it means that the module should not be imported (and used qualified), what about naming it `t` (as in standard OCaml conventions)?

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -1227,3 +1227,39 @@ Exposing constants to OCaml libraries
 
    Due to its internal nature, this command is not for general use. It is meant
    to appear only in standard libraries and in support libraries of plug-ins.
+
+   As a special case, when the first segment of :n:`@qualid__2` is :g:`kernel`,
+   the constant is exposed to the kernel. For instance, the `Int63` module
```
Zimmi48: Why doesn't the kernel reuse the standard mechanism? Because that would have increased the TCB?

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -1227,3 +1227,39 @@ Exposing constants to OCaml libraries
...
+   features the following declaration:
+
+   .. coqdoc::
+
+      Register bool as kernel.ind_bool.
+
+   This makes the kernel aware of what is the type of boolean values. This
+   information is used for instance to define the return type of the
+   :g:`#int63_eq` primitive.
```
Zimmi48: I suggest adding a `.. seealso::` directive with a ref to `primitive-integers`.

doc/sphinx/proof-engine/vernacular-commands.rst
```diff
@@ -1227,3 +1227,39 @@ Exposing constants to OCaml libraries
...
+
+Registering primitive operations
+--------------------------------
+
+.. cmd:: Primitive @ident__1 := #@ident__2.
+
+   Declares :n:`@ident__1` as the primitive operator :n:`#@ident__2`. When
+   running this command, the type of the primitive should be already known by
+   the kernel (this is achieved through this command for primitive types and
+   through the :cmd:`Register` command with the :g:`kernel` name-space).
```
Zimmi48: Did you mean something like this?

```suggestion
   through the :cmd:`Register` command with the :g:`kernel` name-space for functions).
```

Zimmi48 (approved):

doc/sphinx/proof-engine/vernacular-commands.rst
vbgl: What do you call a function? We are talking about types (`int`, `bool`, `comparison`…) some of which are primitive, some of which are not.

doc/sphinx/proof-engine/vernacular-commands.rst
vbgl: Indeed. Coqlib is not in the kernel.

doc/sphinx/language/gallina-extensions.rst
vbgl: The fact that there are 63 bits is an accident. Should this be reflected in the name of the type of “primitive machine integers”?

doc/sphinx/proof-engine/vernacular-commands.rst
vbgl: Done.

doc/sphinx/proof-engine/vernacular-commands.rst
vbgl: Updated. Thanks.

Zimmi48 (commented):
Sorry for thinking about this only now but is it necessary to have three distinct sections for the three documented commands? The vernacular commands chapter already has too many sections IMO. I would prefer if it was a single section for all three commands, e.g. "Internal registering commands" or something like this, and if you really want to keep the existing titles, make them sub-sections.

doc/sphinx/language/gallina-extensions.rst
Zimmi48: I don't really get what you are proposing. But my idea was indeed that this should be reflected, either as `Int63.t` or as `Int63.int63`, whichever you prefer. Anyway, this is outside of the scope of this PR.

vbgl:
The three sections at the end of the “Vernacular commands” chapter have just been merged.

Zimmi48 (approved):
Thanks @vbgl! And sorry for the delay.

PR #9588, opened 2019-02-17
[azure] [ci] Build on Windows using Dune.

ejgallego:
We may want to keep the make-based and Dune job, however the
make-based setup is tested by the INRIA workers so it may not be
needed.

In order for some test to run well, we always run in Dune with an
absolute path. The easiest way to get a portable absolute path is to
use OCaml itself so we introduce a small executable to do that.

While we are at it, we do some cleanup of the test-suite `dune` file,
in particular we remove useless comments, set `--no-buffer` so results
can be seen in real time, and recognize the `NJOBS` variable as we
have moved to a Dune version that supports env vars.

### DIFF
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -34,20 +34,24 @@ include ../Makefile.common
 
 # Default value when called from a freshly compiled Coq, but can be
 # easily overridden
-LIB := ..
+
 BIN := $(shell cd ..; pwd)/bin/
 COQFLAGS?=
 
-coqc_boot := $(BIN)coqc -coqlib $(LIB) -boot -q -test-mode -R prerequisite TestSuite $(COQFLAGS)
-coqc := $(BIN)coqc -coqlib $(LIB) -R prerequisite TestSuite $(COQFLAGS)
-coqchk := $(BIN)coqchk -coqlib $(LIB) -R prerequisite TestSuite
-coqdoc := $(BIN)coqdoc
-coqtop := $(BIN)coqtop -coqlib $(LIB) -boot -q -test-mode -R prerequisite TestSuite
-coqtopbyte := $(BIN)coqtop.byte
+ifeq ($(COQLIB),)
+  # This method of setting `pwd` won't work on win32 OCaml
+  COQLIB := $(shell cd ..; pwd)
+endif
+export COQLIB
 
-coqc_interactive := $(coqc) -async-proofs-cache force
-coqc_boot_interactive := $(coqc_boot) -async-proofs-cache force
-coqdep := $(BIN)coqdep -coqlib $(LIB)
+coqc := $(BIN)coqc -q -R prerequisite TestSuite $(COQFLAGS)
+coqchk := $(BIN)coqchk -R prerequisite TestSuite
+coqdoc := $(BIN)coqdoc
+coqtop := $(BIN)coqtop -q -test-mode -R prerequisite TestSuite
+coqtopbyte := $(BIN)coqtop.byte -q
+
+coqc_interactive := $(coqc) -test-mode -async-proofs-cache force
+coqdep := $(BIN)coqdep
 
 VERBOSE?=
 SHOW := $(if $(VERBOSE),@true,@echo)
@@ -398,7 +402,7 @@ $(addsuffix .log,$(wildcard output/*.v)): %.v.log: %.v %.out $(PREREQUISITELOG)
 	$(HIDE){ \
 	  echo $(call log_intro,$<); \
 	  output=$*.out.real; \
-	  $(coqc_boot_interactive) "$<" $(call get_coq_prog_args,"$<") 2>&1 \
+	  $(coqc_interactive) "$<" $(call get_coq_prog_args,"$<") 2>&1 \
 	    | grep -v "Welcome to Coq" \
 	    | grep -v "\[Loading ML file" \
 	    | grep -v "Skipping rcfile loading" \
@@ -437,7 +441,7 @@ $(addsuffix .log,$(wildcard output-modulo-time/*.v)): %.v.log: %.v %.out
 	  echo $(call log_intro,$<); \
 	  tmpoutput=`mktemp /tmp/coqcheck.XXXXXX`; \
 	  tmpexpected=`mktemp /tmp/coqcheck.XXXXXX`; \
-	  $(coqc_boot_interactive) "$<" $(call get_coq_prog_args,"$<") 2>&1 \
+	  $(coqc_interactive) "$<" $(call get_coq_prog_args,"$<") 2>&1 \
 	    | grep -v "Welcome to Coq" \
 	    | grep -v "\[Loading ML file" \
 	    | grep -v "Skipping rcfile loading" \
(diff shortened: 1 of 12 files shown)


test-suite/Makefile
```diff
@@ -657,6 +657,7 @@ coq-makefile/%.log : coq-makefile/%/run.sh
 	@echo "TEST      coq-makefile/$*"
 	$(HIDE)(\
 	  export COQBIN=$(BIN);\
+	  export COQLIB=$(LIB)/../../;\
```
SkySkimmer: ???
Maybe you want to use the absolute value for LIB?

test-suite/misc/poly-capture-global-univs.sh
```diff
@@ -7,7 +7,7 @@ export PATH=$COQBIN:$PATH
 
 cd misc/poly-capture-global-univs/
 
-coq_makefile -f _CoqProject -o Makefile
+coq_makefile -coqlib "$LIB"/../.. -f _CoqProject -o Makefile
```
SkySkimmer: coq_makefile doesn't understand -coqlib
Also why ../.. as in my other comment

test-suite/Makefile
ejgallego: There is no and easy way to generate an absolute path that will work well for Windows.

ejgallego: We had this discussion in the `-boot` PR and you seemed to favor relative paths, why the change?

test-suite/misc/poly-capture-global-univs.sh
ejgallego: It doesn't understand it but changes in 2.0 make it use the coqlib infrastructure [in a bit of ad-hoc way as it is not locating Coq libraries]

test-suite/Makefile
SkySkimmer: The issue is that here we're `cd`ing so the relative gets invalidated
But LIB isn't always relative, see gitlab.yml, so unconditionally adding ../.. doesn't work.

test-suite/misc/poly-capture-global-univs.sh
SkySkimmer: I mean you're getting an "unknown option -coqlib" error which needs fixing.

ejgallego: I mean `coq_makefile` should be fixed.

ejgallego: Anyways the point is that this kind of setup cannot both work with relative and absolute paths unless you do a hack; in case of a relative path you need to adjust the path, in case of an absolute one you don't.

Now it works just by chance.

ejgallego:
Uh, this took quite a bit of too much effort [including the old Dune PR with appveyor] however it should finally work.

test-suite/Makefile
```diff
@@ -34,20 +34,24 @@ include ../Makefile.common
...
-LIB := ..
+
 BIN := $(shell cd ..; pwd)/bin/
 COQFLAGS?=
 
-coqc_boot := $(BIN)coqc -coqlib $(LIB) -boot -q -test-mode -R prerequisite TestSuite $(COQFLAGS)
-coqc := $(BIN)coqc -coqlib $(LIB) -R prerequisite TestSuite $(COQFLAGS)
-coqchk := $(BIN)coqchk -coqlib $(LIB) -R prerequisite TestSuite
+ifeq ($(COQLIB),)
+  # This method of setting `pwd` won't work on win32 OCaml
```
SkySkimmer: I think when `$(LOCAL)` is `true` (from config/Makefile) it's fine to leave COQLIB unset and that would work on windows.

test-suite/Makefile
```diff
@@ -110,11 +114,14 @@ PREREQUISITELOG = prerequisite/admit.v.log			\
...
 .DELETE_ON_ERROR:
-.PHONY: all run clean $(SUBSYSTEMS)
+.PHONY: all run clean info $(SUBSYSTEMS)
 
-all: run
+all: info run
 	$(MAKE) report
 
+info:
+	echo "Coq library located at: $$($(BIN)coqc -where)"
```
SkySkimmer: To be removed?

test-suite/dune
```diff
@@ -1,3 +1,13 @@
+; The easiest way to generate a portable absolute path is to use OCaml
+; itself to print it
+(executable
+ (name ocaml_pwd)
+ (modules ocaml_pwd))
+
+(rule
+ (targets libpath.inc)
+ (action (with-outputs-to %{targets} (run ./ocaml_pwd.exe ../../install/default/lib/coq/ ))))
```
SkySkimmer: Why redirect stderr?

test-suite/Makefile
```diff
@@ -34,20 +34,24 @@ include ../Makefile.common
...
 # easily overridden
-LIB := ..
+
 BIN := $(shell cd ..; pwd)/bin/
 COQFLAGS?=
 
-coqc_boot := $(BIN)coqc -coqlib $(LIB) -boot -q -test-mode -R prerequisite TestSuite $(COQFLAGS)
-coqc := $(BIN)coqc -coqlib $(LIB) -R prerequisite TestSuite $(COQFLAGS)
-coqchk := $(BIN)coqchk -coqlib $(LIB) -R prerequisite TestSuite
+ifeq ($(COQLIB),)
```
SkySkimmer: Why not `COQLIB ?= ...`?

SkySkimmer (approved):
This path business looks like a pain, thanks for dealing with it.

test-suite/Makefile
ejgallego: I originally had a conditional export so I did it this way.

test-suite/Makefile
ejgallego: Not for dune I think, my plan is to try to remove local tho. IMHO the build layout is similar enough as not to be worth it.

test-suite/Makefile
ejgallego: Yup

test-suite/dune
ejgallego: Good point

test-suite/Makefile
SkySkimmer: >Not for dune I think

dune does `make COQLIB=...` so the `COQLIB := ...` in the makefile is ignored.

ejgallego: Yeah I am never so sure about that overwriting semantics, they seem to scare me; what happens if `COQLIB` is set by an envvar?

SkySkimmer: Setting inside the makefile overrides env vars.

test-suite/Makefile
ejgallego: Done.

ejgallego:
Comments addressed except for the `COQLIB` one which I am not sure what to do with it.

test-suite/dune
```diff
@@ -1,3 +1,13 @@
+; The easiest way to generate a portable absolute path is to use OCaml
+; itself to print it
+(executable
+ (name ocaml_pwd)
+ (modules ocaml_pwd))
+
+(rule
+ (targets libpath.inc)
+ (action (with-stdout-to %{targets} (run ./ocaml_pwd.exe ../../install/default/lib/coq/ ))))
```
SkySkimmer: Does dune always install to `default`?

test-suite/dune
```diff
@@ -1,3 +1,13 @@
+; The easiest way to generate a portable absolute path is to use OCaml
+; itself to print it
+(executable
+ (name ocaml_pwd)
+ (modules ocaml_pwd))
```
SkySkimmer: Does it make sense for this to live in test-suite? How about putting it in tools/?

test-suite/dune
ejgallego: This can be done using a variable I think so indeed it may be a good idea to use the variable.

test-suite/dune
ejgallego: IMHO is a private binary so it is more robust to put it here so we can move paths around, we could put it in tools and use the new binary scoping feature tho, but I felt it was a bit overkill.

test-suite/dune
ejgallego: Used `%{context_name}`

SkySkimmer (approved):
Should be fine

SkySkimmer:
Failed on gitlab windows job. I haven't looked at the logs.

ejgallego:
Yeah I rebased as this was likely to interfere with the boot changes.

ejgallego:
Failure was actually a transient network problem.

ejgallego:
Merci !

PR #9590, opened 2019-02-17
[gitlab] [docker] [ci] Remove "edge" compiler switch.

ejgallego:
Since a long time the compiler switch is not very useful as it is not
used to test any CI.

The `edge+flambda` version seems stable enough to carry out the `edge`
testing by itself, thus we remove the `egde` switch saving valuable
Docker image size and Gitlab runners.

We also tweak the `pkg:opam` job as to correctly set the version of
the pinned local package.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,7 +10,7 @@ stages:
 variables:
   # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
   # for reference]
-  CACHEKEY: "bionic_coq-V2019-01-28-V1"
+  CACHEKEY: "bionic_coq-V2019-02-17-V1"
   IMAGE: "$CI_REGISTRY_IMAGE:$CACHEKEY"
   # By default, jobs run in the base switch; override to select another switch
   OPAM_SWITCH: "base"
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -282,12 +278,14 @@ pkg:opam:
   dependencies: []
   script:
     - set -e
-    - opam pin add coq .
-    - opam pin add coqide-server .
-    - opam pin add coqide .
+    - opam pin add --kind=path coq.$COQ_VERSION .
+    - opam pin add --kind=path coqide-server.$COQ_VERSION .
+    - opam pin add --kind=path coqide.$COQ_VERSION .
```
SkySkimmer: How's this related?

SkySkimmer (approved):
I'm fine with removing edge but please explain the opam pin changes.

.gitlab-ci.yml
ejgallego: Oh I forgot to update the PR message, the commit message has the info.

ejgallego:
I was going to say I can split the `pkg:opam` fix to another PR if you want but this is merged so the issue is settled, thanks.

.gitlab-ci.yml
ejgallego: @Zimmi48 proposes here to use `dev` as `COQ_VERSION` ; I'm unsure if that would create problem, semantics of `pin` and `dev` are very tricky. Does anyone know?

PR #9591, opened 2019-02-18
Fix a problem with -mangle-names and inductive definitions.

### LINKED ISSUE #8790: -mangle-names fails on inductive definition

jashug:
We track separately the argument name and parameter name, and insert explicit applications of the form `(arg_1 := A)` rather than `(arg_1 := arg_1)`. Usually argument names agree with user-given parameter names in the absence of repeated parameter names, but `-mangle-names` breaks this invariant, so while default argument names need to be refreshed, this change is useful.

With `-mangle-names`, the default argument names will still be mangled, so need an explicit `Arguments` command to fix the names, but this at least removes a confusing error case.

**Kind:** bug fix

Fixes / closes #8790

### DIFF
--- a/interp/constrintern.ml
+++ b/interp/constrintern.ml
@@ -57,8 +57,8 @@ type var_internalization_data =
        constructor of JMeq, "JMeq" behaves as a variable of type Inductive *)
     var_internalization_type *
     (* impargs to automatically add to the variable, e.g. for "JMeq A a B b"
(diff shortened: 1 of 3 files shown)


jashug:
Suffers from needing to add the `rename` flag to `Arguments` command everywhere.

PR #9597, opened 2019-02-18
[ci] Resolve commit corresponding to branch when downloading tarball.

### LINKED ISSUE #8461: [ci] Download of branch tarballs loses commit hash information.
https://github.com/coq/coq/pull/8348 introduced a more convenient download method for CI testing; however, when downloading a branch [for example, `master`], the scripts will build a URL using the branch name so the CI log won't contain information about the exact hash being tested.

In order to improve this, we should instead use `git ls-remote` and download an archive corresponding to a concrete hash, as hinted in #8348

Zimmi48:
This ensures that the log will contain the commit hash that we tested.
This reuses the method from the Windows build script (we have a number of common functions, it would be interesting to start a bash shared library file).

**Kind:** infrastructure.

Fixes #8461

(Sorry for the delay in preparing this trivial fix.)

### DIFF
--- a/dev/ci/ci-common.sh
+++ b/dev/ci/ci-common.sh
@@ -62,27 +62,30 @@ git_download()
 {
   local PROJECT=$1
   local DEST="$CI_BUILD_DIR/$PROJECT"
+  local GITURL_VAR="${PROJECT}_CI_GITURL"
+  local GITURL="${!GITURL_VAR}"
+  local REF_VAR="${PROJECT}_CI_REF"
+  local REF="${!REF_VAR}"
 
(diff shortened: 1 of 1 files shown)


SkySkimmer (approved):
I guess it works with coquelicot since it's using force_git
Why did relation-algebra fail with "ERROR: Job failed: canceled"?

ejgallego (approved):

ejgallego:
> Why did relation-algebra fail with "ERROR: Job failed: canceled"?

Let's retry.

ejgallego:
Thanks Théo!

PR #9599, opened 2019-02-18
Remove undefined install_printer ppcumulativity_info

SkySkimmer:


ppedrot:
Didn't we have a test for that at some point?

PR #9600, opened 2019-02-18
CI: fix trunk jobs switch picking

SkySkimmer:
After #9590

Instead of this we could override before_script, either duplicating the debug
prints or just not doing debug prints for the trunk jobs.


ejgallego (approved):
Well spotted.

PR #9602, opened 2019-02-19
[kernel] termination checking: backtrack on stuck case, fix, proj

### LINKED ISSUE #9598: [kernel] [termination checking] reduction strategy incomplete
```coq
Inductive pair := K (n1 : nat) (n2 : nat).
Definition fst (p : pair) : nat := match p with K n _ => n end.
Definition mkK a b := K a b.
Fixpoint rec (x : nat) : nat := fst (K 0 (rec x)). (* works *)
Fixpoint rec (x : nat) : nat := fst (mkK 0 (rec x)). (* fails *)
```
The code in `inductive.ml` unfolds `mkK` then unfolds `fst` (but forgets it unfoleded `mkK`), hence it finds `match (mkK ...) with..` then it re-unfolds `mkK` but forgets that the obtained term is not neutral for the enclosing context.

gares:
Unfolding heuristic in termination checking made more complete.
In particular Coq is now more aggressive in unfolding constants
when it looks for a iota redex. Performance regression may occur
in Fixpoint declarations without an explicit {struct} annotation,
since guessing the decreasing argument can now be more expensive.

Fix #9598

### DIFF
--- a/test-suite/bugs/closed/bug_9598.v
+++ b/test-suite/bugs/closed/bug_9598.v
@@ -9,3 +9,15 @@ Module case.
   Fixpoint rec_ko (x : nat) : nat := fst (alias_K 0 (rec_ko x)).
 
 End case.
+
+Module fixpoint.
+
+  Inductive pair := K (n1 : nat) (n2 : nat).
+  Fixpoint fst (p : pair) : nat := match p with K n _ => n end.
+
+  Definition alias_K a b := K a b.
+
+  Fixpoint rec (x : nat) : nat := fst (K 0 (rec x)).
+  Fixpoint rec_ko (x : nat) : nat := fst (alias_K 0 (rec_ko x)).
+
+End fixpoint.
--- a/test-suite/bugs/closed/bug_9598.v
+++ b/test-suite/bugs/closed/bug_9598.v
@@ -21,3 +21,16 @@ Module fixpoint.
   Fixpoint rec_ko (x : nat) : nat := fst (alias_K 0 (rec_ko x)).
 
 End fixpoint.
+
+Module primproj.
+
+  Set Primitive Projections.
+
+  Inductive pair := K { fst : nat; snd : nat }.
+
+  Definition alias_K a b := K a b.
+
+  Fixpoint rec (x : nat) : nat := fst (K 0 (rec x)).
+  Fixpoint rec_ko (x : nat) : nat := fst (alias_K 0 (rec_ko x)).
+
+End primproj.
--- a/test-suite/bugs/closed/bug_9598.v
+++ b/test-suite/bugs/closed/bug_9598.v
@@ -1,0 +1,11 @@
+Module case.
+
+  Inductive pair := K (n1 : nat) (n2 : nat).
+  Definition fst (p : pair) : nat := match p with K n _ => n end.
+
+  Definition alias_K a b := K a b.
+
+  Fixpoint rec (x : nat) : nat := fst (K 0 (rec x)).
+  Fixpoint rec_ko (x : nat) : nat := fst (alias_K 0 (rec_ko x)).
+
+End case.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,13 @@ Kernel
 
 - Added primitive integers
 
+- Unfolding heuristic in termination checking made more complete.
+  In particular Coq is now more aggressive in unfolding constants
+  when it looks for a iota redex. Performance regression may occur
+  in Fixpoint declarations without an explicit {struct} annotation,
+  since guessing the decreasing argument can now be more expensive.
+  (PR #9602)
+
 Notations
 
 - New command `Declare Scope` to explicitly declare a scope name
--- a/dev/ci/user-overlays/09602-gares-more-delta-in-termination-checking.sh
+++ b/dev/ci/user-overlays/09602-gares-more-delta-in-termination-checking.sh
@@ -1,0 +1,6 @@
+if [ "$CI_PULL_REQUEST" = "9602" ] || [ "$CI_BRANCH" = "more-delta-in-termination-checking" ]; then
+
+    equations_CI_REF=more-delta-in-termination-checking
+    equations_CI_GITURL=https://github.com/gares/Coq-Equations
+
+fi
--- a/kernel/inductive.ml
+++ b/kernel/inductive.ml
@@ -1015,9 +1015,22 @@ let check_one_fix renv recpos trees def =
         | (Ind _ | Construct _) ->
             List.iter (check_rec_call renv []) l
 
-        | Proj (_p, c) ->
-           List.iter (check_rec_call renv []) l;
-           check_rec_call renv [] c
+        | Proj (p, c) ->
+            begin try
+              List.iter (check_rec_call renv []) l;
+              check_rec_call renv [] c
+            with (FixGuardError _ as exn) ->
+              let exn = CErrors.push exn in
+              (* we try hard to reduce the proj away by looking for a
+                 constructor in c (we unfold definitions too) *)
+              let c = whd_all renv.env c in
+              let hd, _ = decompose_app c in
+              match kind hd with
+              | Construct _ ->
+                  check_rec_call renv []
+                    (Term.applist (mkProj(Projection.unfold p,c), l))
+              | _ -> Exninfo.iraise exn
+            end
 
         | Var id ->
             begin
--- a/kernel/inductive.ml
+++ b/kernel/inductive.ml
@@ -921,16 +921,30 @@ let check_one_fix renv recpos trees def =
               end
 		
         | Case (ci,p,c_0,lrest) ->
-            List.iter (check_rec_call renv []) (c_0::p::l);
-            (* compute the recarg information for the arguments of
-               each branch *)
-            let case_spec = branches_specif renv 
-	      (lazy_subterm_specif renv [] c_0) ci in
-	    let stack' = push_stack_closures renv l stack in
-            let stack' = filter_stack_domain renv.env p stack' in
-              Array.iteri (fun k br' -> 
-			     let stack_br = push_stack_args case_spec.(k) stack' in
-			     check_rec_call renv stack_br br') lrest
+            begin try
+              List.iter (check_rec_call renv []) (c_0::p::l);
+              (* compute the recarg info for the arguments of each branch *)
+              let case_spec =
+                branches_specif renv (lazy_subterm_specif renv [] c_0) ci in
+              let stack' = push_stack_closures renv l stack in
+              let stack' = filter_stack_domain renv.env p stack' in
+              lrest |> Array.iteri (fun k br' ->
+                let stack_br = push_stack_args case_spec.(k) stack' in
+                check_rec_call renv stack_br br')
+            with (FixGuardError _ as exn) ->
+              let exn = CErrors.push exn in
+              (* we try hard to reduce the match away by looking for a
+                 constructor in c_0 (we unfold definitions too) *)
+              let c_0 = whd_all renv.env c_0 in
+              let hd, _ = decompose_app c_0 in
+              match kind hd with
+              | Construct _ ->
+                  (* the call to whd_betaiotazeta will reduce the
+                     apparent iota redex away *)
+                  check_rec_call renv []
+                    (Term.applist (mkCase (ci,p,c_0,lrest), l))
+              | _ -> Exninfo.iraise exn
+            end
 
         (* Enables to traverse Fixpoint definitions in a more intelligent
            way, ie, the rule :
--- a/kernel/inductive.ml
+++ b/kernel/inductive.ml
@@ -959,19 +959,33 @@ let check_one_fix renv recpos trees def =
            then f is guarded with respect to S in (g a1 ... am).
            Eduardo 7/9/98 *)
         | Fix ((recindxs,i),(_,typarray,bodies as recdef)) ->
-            List.iter (check_rec_call renv []) l;
-            Array.iter (check_rec_call renv []) typarray;
             let decrArg = recindxs.(i) in
-            let renv' = push_fix_renv renv recdef in
-	    let stack' = push_stack_closures renv l stack in
-              Array.iteri
-                (fun j body ->
-                   if Int.equal i j && (List.length stack' > decrArg) then
-		     let recArg = List.nth stack' decrArg in
-	             let arg_sp = stack_element_specif recArg in
-	             check_nested_fix_body renv' (decrArg+1) arg_sp body
-                   else check_rec_call renv' [] body)
-                bodies
+            begin try
+              List.iter (check_rec_call renv []) l;
+              Array.iter (check_rec_call renv []) typarray;
+              let renv' = push_fix_renv renv recdef in
+              let stack' = push_stack_closures renv l stack in
+              bodies |> Array.iteri (fun j body ->
+                if Int.equal i j && (List.length stack' > decrArg) then
+                  let recArg = List.nth stack' decrArg in
+                  let arg_sp = stack_element_specif recArg in
+                  check_nested_fix_body renv' (decrArg+1) arg_sp body
+                else check_rec_call renv' [] body)
+            with (FixGuardError _ as exn) ->
+              let exn = CErrors.push exn in
+              (* we try hard to reduce the fix away by looking for a
+                 constructor in l[decrArg] (we unfold definitions too) *)
+              if List.length l <= decrArg then Exninfo.iraise exn;
+              let recArg = List.nth l decrArg in
+              let recArg = whd_all renv.env recArg in
+              let hd, _ = decompose_app recArg in
+              match kind hd with
+              | Construct _ ->
+                  let before, after = CList.(firstn decrArg l, skipn (decrArg+1) l) in
+                  check_rec_call renv []
+                    (Term.applist (mkFix ((recindxs,i),recdef), (before @ recArg :: after)))
+              | _ -> Exninfo.iraise exn
+            end
 
         | Const (kn,_u as cu) ->
             if evaluable_constant kn renv.env then


SkySkimmer:
You don't need a PR to run jenkins though. Or do you intend some other meaning of benchmark?

gares:
Reworded

gares:
FTR, this commit makes the real example that was minimized to #9598 go trough. In the real example is not sufficient to unalias (delta expand by hand) `K`, one need to do more delta expansions by hand, so I can't work around the bug and hence this PR.

gares:
En passant, the current code does step-by-step, inside-out, delta unfolding *but* forges the unfolded (exactly as the `simpl` tactic does). So you get at least a quadratic blowup on terms where `f1 (f2 (f3 (rec_call wrong)))` since you unfold `f3` then `f2` in `(f2 (f3 ...))` then `f3` then `f1` in `(f2 (f3 ..))`... then `f3` then `f2` then `f3`...

With this PR delta unfolding is eager and outside-in. It may blow up in some cases where the lazy strategy is, well, lazy. But may also be much faster in examples like the one above. So let's see what the bench says.

SkySkimmer:
The bench was still running HoTT after 4h so I cancelled it. Not sure if it was an issue with an old version of the PR or what.

SkySkimmer:
Equations has a test for the old behaviour.

gares:
@mattam82 confirms that it was an incompleteness that this PR also fixes.

gares:
I did not write the overlay yet, but I think the PR can be reviewed now @maximedenes 
The bench is still running (so far no impact), but the current patch is way more conservative so I don't expect any surprise there.

ppedrot:
Why isn't there the same problem for other neutral terms? For instance, it might be the case that a fixpoint is blocked by a constructor hidden by a δ...

ppedrot:
More generally, why don't we perform full weak-head with δ included upfront? Is it for performance reasons?

gares:
> More generally, why don't we perform full weak-head with δ included upfront? Is it for performance reasons?

This is the first thing I tried and indeed it blows up.

I share your remark about neutral terms and their context.
I think it should be done fox fix too, but I first want the bench

gares:
```
┌────────────────────────┬──────────────────────────┬──────────────────────────────────────────────┬──────────────────────────────────────────────┬───────────────────────────────┬─────────────────────────────┐
│                        │      user time [s]       │                  CPU cycles                  │               CPU instructions               │     max resident mem [KB]     │         mem faults          │
│                        │                          │                                              │                                              │                               │                             │
│           package_name │     NEW     OLD  PDIFF   │               NEW               OLD  PDIFF   │               NEW               OLD  PDIFF   │        NEW        OLD PDIFF   │     NEW     OLD     PDIFF   │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│ coq-mathcomp-ssreflect │   38.45   38.91  -1.18 % │      106930087435      106973977575  -0.04 % │      128042685666      128044913363  -0.00 % │     523208     523116 +0.02 % │      14      65    -78.46 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│       coq-math-classes │  215.45  216.46  -0.47 % │      604634317621      603664760121  +0.16 % │      767262261165      767182924329  +0.01 % │     514596     514624 -0.01 % │      16      13    +23.08 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│     coq-mathcomp-field │  219.75  220.50  -0.34 % │      612150349747      614664068183  -0.41 % │      892466362513      892429985516  +0.00 % │     749956     749888 +0.01 % │       1       0      +nan % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│             coq-sf-plf │   43.29   43.41  -0.28 % │      121150035211      121193809816  -0.04 % │      153588604364      153594105689  -0.00 % │     490092     491380 -0.26 % │      44       0      +nan % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│              coq-flocq │  556.77  558.20  -0.26 % │     1549007451560     1551893876605  -0.19 % │     2280493274320     2280494287070  -0.00 % │    1718748    1718848 -0.01 % │     203     102    +99.02 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│  coq-mathcomp-fingroup │   54.08   54.21  -0.24 % │      151269594514      151262859467  +0.00 % │      197900004118      197921310232  -0.01 % │     573892     572956 +0.16 % │       0      28   -100.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│  coq-mathcomp-solvable │  189.94  190.20  -0.14 % │      529143008351      529790103760  -0.12 % │      711657197892      711631277140  +0.00 % │     848404     850756 -0.28 % │      22       0      +nan % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│ coq-mathcomp-character │  250.31  250.48  -0.07 % │      697319868989      697720066663  -0.06 % │      963269497652      963320718032  -0.01 % │    1069092    1068904 +0.02 % │       0       0      +nan % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│           coq-compcert │ 1002.81 1003.47  -0.07 % │     2795524183582     2795620239411  -0.00 % │     4363739414702     4363686897480  +0.00 % │    1313000    1313124 -0.01 % │     125      79    +58.23 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│ coq-mathcomp-odd-order │ 1359.91 1360.47  -0.04 % │     3790510988430     3789811827670  +0.02 % │     6367263608148     6367246553287  +0.00 % │    1348444    1348952 -0.04 % │       1       8    -87.50 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│   coq-mathcomp-algebra │  151.72  151.78  -0.04 % │      423121503156      423052432178  +0.02 % │      534323629829      534372379360  -0.01 % │     636912     636788 +0.02 % │     110       2  +5400.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│                coq-vst │ 3036.48 3037.36  -0.03 % │     8463758520431     8463018155480  +0.01 % │    11865749316421    11854369113611  +0.10 % │    1715872    1715696 +0.01 % │      53     119    -55.46 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│        coq-fiat-crypto │ 3907.01 3907.75  -0.02 % │    10852538643289    10855730443660  -0.03 % │    17509800717060    17505832194315  +0.02 % │    2457580    2551364 -3.68 % │    1051    1007     +4.37 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│         coq-verdi-raft │ 1367.28 1367.26  +0.00 % │     3813718351982     3813324494050  +0.01 % │     5220122391522     5220141611579  -0.00 % │    1802632    1802624 +0.00 % │       0       5   -100.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│        coq-lambda-rust │ 2062.87 2062.71  +0.01 % │     5750260340304     5748689241738  +0.03 % │     7929849410293     7933134620814  -0.04 % │    1375032    1374932 +0.01 % │      54       0      +nan % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│            coq-unimath │ 1944.89 1944.48  +0.02 % │     5415353337958     5413143950930  +0.04 % │     9426253752258     9425571317771  +0.01 % │     956320     903668 +5.83 % │     519     204   +154.41 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│              coq-verdi │  122.53  122.49  +0.03 % │      340493546002      340700209578  -0.06 % │      442434654559      442501232572  -0.02 % │     611928     605076 +1.13 % │      10       4   +150.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│             coq-geocoq │ 1982.15 1981.47  +0.03 % │     5531920225418     5530364964832  +0.03 % │     8059582543507     8060619868480  -0.01 % │    1196296    1190808 +0.46 % │     163      60   +171.67 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│       coq-fiat-parsers │  686.67  685.91  +0.11 % │     1920773825538     1920572084921  +0.01 % │     2950627612105     2949678572104  +0.03 % │    3022284    3034596 -0.41 % │     174     668    -73.95 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│              coq-color │  682.85  681.77  +0.16 % │     1908221697045     1908556226004  -0.02 % │     2285584755737     2285666295266  -0.00 % │    1379488    1385224 -0.41 % │       4      11    -63.64 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│          coq-fiat-core │  104.88  104.70  +0.17 % │      299559806343      298911370817  +0.22 % │      374520208317      374514257248  +0.00 % │     476428     476588 -0.03 % │     308     193    +59.59 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│    coq-formal-topology │   30.20   30.13  +0.23 % │       82345881612       82507075214  -0.20 % │      107792141948      107789226869  +0.00 % │     467396     467344 +0.01 % │      21       1  +2000.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│         coq-coquelicot │   73.96   73.71  +0.34 % │      202278276710      202319862473  -0.02 % │      244753108885      244780937470  -0.01 % │     687936     687936 +0.00 % │     214      11  +1845.45 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│               coq-hott │  337.92  336.69  +0.37 % │      915091553797      914779685103  +0.03 % │     1397609768871     1397613133315  -0.00 % │     479148     479104 +0.01 % │     391       1 +39000.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│            coq-bignums │   69.77   69.50  +0.39 % │      194186330781      194116323758  +0.04 % │      256890219856      256910439713  -0.01 % │     522600     522688 -0.02 % │     151       9  +1577.78 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼─────────────────────────────┤
│               coq-corn │ 1929.29 1484.99 +29.92 % │     5380525398448     4145705283146 +29.79 % │     8395145152750     6168273725279 +36.10 % │     854680     854708 -0.00 % │       1       1     +0.00 % │
└────────────────────────┴──────────────────────────┴──────────────────────────────────────────────┴──────────────────────────────────────────────┴───────────────────────────────┴─────────────────────────────┘
```
The bench shows a problem in corn

gares:
The issue on Corn is trivially solved. Coq now tries harder, if you omit `struct` and ask Coq to try blindly it may take more time to fail.

gares:
The linter strikes back:
```
dev/ci/user-overlays/09602-gares-more-delta-in-termination-checking.sh:1: trailing whitespace.
```
on a generated file... how useful

gares:
```
│               coq-corn │ 1487.14 1485.48 +0.11 % │    4151269587320    4147844115833 +0.08 % │    6168155633758    6168046144603 +0.00 % │     854720     854684 +0.00 % │      1      1    +0.00 % │
```
OK, bench is now good.

ppedrot:
@gares If I read the code correctly, now you also need to handle the `Fix` and `Proj` cases similarly.

gares:
Yes, now that I know the patch for match has no regression I can do that.

gares:
I've added the code for Fix and Proj along with test cases.

The funny thing is that *both* tests about Proj were failing, that is Proj was *never* reduced/unfolded before.

gares:
New bench at https://ci.inria.fr/coq/job/benchmark-part-of-the-branch/646/console

gares:
> Proj was _never_ reduced/unfolded before.

CC @maximedenes and @vbgl that are trying to enable primproj in stdlib2

gares:
After the 10th iteration I think I've finally made the white space master happy about all my commits.
It is now safe to review the patch without sunglasses.

gares:
ping

maximedenes:
I'm a bit concerned by the performance regressions, as they are hard to debug for non-experts. On the other hand, the patch looks sound. @ppedrot what's your take on this?

ppedrot:
If you write crazy code with transparent definitions all over the place and you don't specify `struct` annotations, you can only blame yourself. You could even get in definitional trouble if the algorithm picks up a different variable (e.g. when doing recursion over two `nat` arguments).

I support the patch, maybe with **AN EVEN MORE VISIBLE WARNING** in the `CHANGES` files regarding performance.

gares:
I did not have time to check which fixpoint(s) in flocq became slower.
I think we should be nice and tell them where they should add a struct annotation (help is welcome, since I won't be able to work on this for ~ 2 weeks). Weather this is blocking or now, it is not for me to say. 

Given that the first benchmark did not show any slowdown (back then I was backtracking only on `match`), and given that I doubt they use primitive projections, the fixpoint they don't annotate with `struct` must contain some other `fix` (slowdown appeared when I added the paragraph for backtracking on fixpoints). Dunno if it can help.

gares:
> I support the patch, maybe with **AN EVEN MORE VISIBLE WARNING**

I don't know how to do that without using bold characters. I hope that people read, at least, the kernel part of the CHANGES.

Maybe something to consider it to emit a warning saying "this fix lacks a struct annotation, please put `{struct x}` there". It would generate tons of warnings, but the current strategy to "guess" is very silly. One could imagine a termination checking code carrying a set of potential variables and when one that works for all calls is found we stop (instead of trying very hard to discover that the others don't work).

test-suite/bugs/closed/bug_9598.v
```diff
@@ -0,0 +1,36 @@
+Module case.
+
+  Inductive pair := K (n1 : nat) (n2 : nat).
+  Definition fst (p : pair) : nat := match p with K n _ => n end.
+
+  Definition alias_K a b := K a b.
+
+  Fixpoint rec (x : nat) : nat := fst (K 0 (rec x)).
```
mattam82: Wait a sec, this means that even more definitions are accepted that will loop in call by value?

JasonGross:
>I don't know how to do that without using bold characters. I hope that people read, at least, the kernel part of the CHANGES.

CHANGES is now an md file, so we could use bold characters in CHANGES.md...

test-suite/bugs/closed/bug_9598.v
gares: I think so.  The criteria has always been that it should terminate in when strategy, the one used by the kernel. Or at least, since 2010 :-) https://github.com/coq/coq/issues/1843

gares: Thinking more about it, I'm not sure CBV is necessarily the cause of potential divergence.
```
Definition bla := true.
Fixpoint f x := match bla with true => 1 | false -> f x  end.
```
Unless I'm mistaken this term converges to 1 in CBV, since the match is reduced first.

So, what is true, is that with this patch we accept more terms that would diverge by doing a strong reduction when there is an explicit match/fix/proj.

Another way to put it
```
Definition bla := true.
Fixpoint f x := match x with S x' => f (match bla with true => x' | false -> f x  end) | _ => 0 end .
```
This term was rejected before, should be accepted now, but converges in CBV (evaluating the argument of `f` gives `x'` without touching the `f x` subterm).

gares: I'm not sure I follow. `Fixpoint rec` at line 8 works without this patch. What this PR enables is the `rec_ko` example in the next line (that is identical, but `K` is hidden behind a delta expansion).

If we are on the same page, then please explain what do you mean by "linear whd that prevented reducing K bindings".

gares:
BTW, I could not reproduce the 1% slowdown locally on floq, and most Fixpoint do have a struct annotation.
So I'd say green flag there.

test-suite/bugs/closed/bug_9598.v
mattam82: We are on the same page. What I have in mind is an improvement of the guard checking using a  more fine-grained whd that prevents the reduction of K redexes that make unguarded recursive calls disappear, with the goal of establishing that guarded definitions are strongly normalizing (i.e. for any strategy). This would reject both `rec` and `rec_ko`, so I'm wondering if you have use cases for these kind of definitions. It would still allow cases where the recursive function is, partially applied before the `whd` but correctly applied after, for example.

gares: Rec_ko is a minimised version of a construction (I do it in elpi). I've a notion of trivial predicate, true on the entire domain and contractible. I prove contractibility by recursion. I alias/specialize K because the second component is not any P but a specific thing (being contractible) and I don't want to repeat it.

In the real case I have the rec call becomes fully applied, but the reduction does not unfold the alias. 

So yes, I think it would be ok to refine reduction to not drop rec calls that are partially applied.

But I don't see why a tricky language would diverge there, since partially applied functions (rec or not) are values. So I m lost again. :-)

gares:
@maximedenes are there other concerns on your side?
@mattam82 the discussion was interesting,  but I'm not sure I see clearly what its impact should be on this PR specifically.

test-suite/bugs/closed/bug_9598.v
mattam82: > Rec_ko is a minimised version of a construction (I do it in elpi). I've a notion of trivial predicate, true on the entire domain and contractible. I prove contractibility by recursion. I alias/specialize K because the second component is not any P but a specific thing (being contractible) and I don't want to repeat it.
> 
> In the real case I have the rec call becomes fully applied, but the reduction does not unfold the alias.

> 
> So yes, I think it would be ok to refine reduction to not drop rec calls that are partially applied.


Good, that's reassuring.
 
> But I don't see why a strict language would diverge there, since partially applied functions (rec or not) are values. So I m lost again. :-)

We agree, partial applications that ultimately reduce to guarded fully applied recursive calls are fine even with cbv evaluation. But the rec/rec_ko examples you added to the test-suite are not partially applied.

I'm fine with the PR as is.

mattam82 (approved):

gares:
Can we merge this one? I don't see any action needed on my side.

maximedenes (approved):

maximedenes:
> I support the patch, maybe with **AN EVEN MORE VISIBLE WARNING** in the `CHANGES` files regarding performance.

@gares do you want to address that?

gares:
I'm not so happy with bold or CAPS text, but if you have a better phrasing I'm OK with it.

maximedenes:
> I'm not so happy with bold or CAPS text, but if you have a better phrasing I'm OK with it.

Ok, so unless @ppedrot objects quickly, I'll merge as-is.

gares:
thanks!

tchajed:
For the record, this change caused a couple fixpoints without `{struct}` annotations to blow up and timeout on Travis; our fix is at mit-pdos/armada@da63e55ff8099602fceebb1b7c791f0caec15986.

gares:
Thanks. Was it hard to discover the culprit? Can we do something else to help users migrate?

tchajed:
@jtassarotti can comment more on how he figured it out, but his explanation was that he got it to work in proof mode, saw that the term had a `{struct}` annotation, and then added that. Arguably we should be checking `CHANGES.md` whenever something breaks, but I don't have that habit.

The most helpful thing would be for the termination checker to timeout and report that unfolding took too long and suggest adding a `{struct}` annotation. That said, I think the `CHANGES.md` entry for this issue is pretty good.

For our immediate problem, we should just get our current projects onto Coq's CI, but this won't work for everybody and I think it has high overhead (both the setup on Coq CI and then maintenance from the Coq developers).

Currently we do test against the docker Coq nightly version, so issues are caught quickly and we could easily review the Coq commits from the previous day to figure out where the regression is. The Docker Coq setup could make this easier by reporting the Coq commit hash to the Travis log, and providing some way to get a URL to a GitHub log between successive versions. If with a single click I could see all the modifications to `CHANGES.md` between Coq nightlies I would definitely start there.

ejgallego:
> Currently we do test against the docker Coq nightly version, so issues are caught quickly and we could easily review the Coq commits from the previous day to figure out where the regression is. The Docker Coq setup could make this easier by reporting the Coq commit hash to the Travis log, and providing some way to get a URL to a GitHub log between successive versions. If with a single click I could see all the modifications to `CHANGES.md` between Coq nightlies I would definitely start there.

I'm sure the docker images maintainer @erikmd would be happy to add something like that.

erikmd:
@ejgallego @tchajed Each `coqorg/coq:dev` image replaces the last-but-one each day at ~00:00 UTC, but it's true that the info giving the precise revision of the compiled `coq.dev` is not yet available in the image.

I've recently read related documentation on this topic and I confirm this is something easy to implement with Docker labels. I plan to take a look at this more closely soon (ideally next week).

In the meantime, you can have an *approximate* idea of the commits that have been integrated since last-but-one image by browsing one of these URLs:
* <https://github.com/coq/coq/commits/master/CHANGES.md>
* <https://github.com/coq/coq/compare/master@{2day}...master>

ejgallego:
Cool, thanks @erikmd ; maybe the Git hash could be just exported also as an env var, such as `COQ_DOCKER_GIT_HASH` etc...

tchajed:
Actually I more or less solved my problem. opam already knows what git hash Coq is at, so I just include a URL of the form https://github.com/coq/coq/commits/a796822e in the Travis logs. GitHub groups the log by date so it's pretty easy to browse it.

The change to our Travis config to add this output is at https://github.com/mit-pdos/armada/commit/7e9da81488448b5727d125fc43ee6a57c9131649.

erikmd:
Hi @tchajed @ejgallego, to follow-up [my earlier comment](https://github.com/coq/coq/pull/9602#issuecomment-477536834), I opened issue coq-community/docker-coq#1

PR #9604, opened 2019-02-19
Gramlib: Fixes #9358 (ensuring that requested locations are effectively computed at lexing time)

### LINKED ISSUE #9358: Anomaly "Uncaught exception Invalid_argument("Lexer: location function")." with custom entries
The following script:

```coq
Declare Custom Entry expr.
Notation "[ e ]" := e (e custom expr at level 2).
Check [ fun x => x ].
```

gives the anomaly in the title.

Tested in master 2e231b0eb3d107aedc7a26dc173d6ac00cb3473f

herbelin:
**Kind:** bug fix

Fixes / closes #9358

- [ ] Added / updated test-suite (no test added: I don't know at once how to test a parsing error - `Fail` does not capture them)

We take the shortest path. (I thought it would be possible to attach the locations to the tokens, but gramlib is requesting locations without having a token at hand, so that seems difficult to do and I renounce to investigate this further.)

### DIFF
--- a/gramlib/grammar.ml
+++ b/gramlib/grammar.ml
@@ -1368,6 +1368,9 @@ let parse_parsable entry p =
   let get_loc () =
     try
       let cnt = Stream.count ts in
+      (* Ensure that the token at location cnt has been peeked so that
+         the location function knows about it *)
+      let _ = Stream.peek ts in
       let loc = fun_loc cnt in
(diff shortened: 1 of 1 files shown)


ejgallego (approved):
Looks good as a fix, thanks @herbelin . I'm afraid we cannot have this for 8.9, right?

PR #9607, opened 2019-02-20
[ci] add metacoq

mattam82:
If you thought we didn't exercise the kernel's interfaces and extraction enough, here's a way to do it. This adds metacoq to the CI. The build is short (~6min) but a bit intricate (it separately builds various packed ocaml libraries depending on each other) and I only tested on Linux Travis instances so it might fail on other platforms.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -731,9 +731,23 @@ plugin:ci-elpi:
 
 plugin:ci-equations:
   extends: .ci-template
+  artifacts:
+    name: "$CI_JOB_NAME"
+    paths:
+      - _build_ci
 
 plugin:ci-fiat_parsers:
   extends: .ci-template
+
+plugin:ci-metacoq:
+  extends: .ci-template
+  stage: stage-3
+  needs:
+  - build:base
+  - plugin:ci-equations
+  dependencies:
+  - build:base
+  - plugin:ci-equations
 
 plugin:ci-mtac2:
   extends: .ci-template
--- a/dev/ci/ci-metacoq.sh
+++ b/dev/ci/ci-metacoq.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+ci_dir="$(dirname "$0")"
+. "${ci_dir}/ci-common.sh"
+
+git_download metacoq
+
+( cd "${CI_BUILD_DIR}/metacoq" && ./configure.sh local && make ci-local && make install )
--- a/Makefile.ci
+++ b/Makefile.ci
@@ -33,6 +33,7 @@ CI_TARGETS= \
     ci-iris-lambda-rust \
     ci-math-classes \
     ci-math-comp \
+    ci-metacoq \
     ci-mtac2 \
     ci-paramcoq \
     ci-perennial \
@@ -71,6 +72,8 @@ ci-fiat-crypto: ci-coqprime ci-rewriter
 
 ci-simple-io: ci-ext-lib
 ci-quickchick: ci-ext-lib ci-simple-io
+
+ci-metacoq: ci-equations
 
 # Generic rule, we use make to ease CI integration
 $(CI_TARGETS): ci-%:
--- a/dev/ci/ci-basic-overlay.sh
+++ b/dev/ci/ci-basic-overlay.sh
@@ -330,3 +330,10 @@
 : "${perennial_CI_REF:=master}"
 : "${perennial_CI_GITURL:=https://github.com/mit-pdos/perennial}"
 : "${perennial_CI_ARCHIVEURL:=${perennial_CI_GITURL}/archive}"
+
+########################################################################
+# metacoq
+########################################################################
+: "${metacoq_CI_REF:=master}"
+: "${metacoq_CI_GITURL:=https://github.com/MetaCoq/metacoq}"
+: "${metacoq_CI_ARCHIVEURL:=${metacoq_CI_GITURL}/archive}"
--- a/dev/ci/ci-equations.sh
+++ b/dev/ci/ci-equations.sh
@@ -5,4 +5,4 @@ ci_dir="$(dirname "$0")"
 
 git_download equations
 
-( cd "${CI_BUILD_DIR}/equations" && ./configure.sh coq && make ci)
+( cd "${CI_BUILD_DIR}/equations" && ./configure.sh coq && make ci && make install )


dev/ci/ci-metacoq.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+ci_dir="$(dirname "$0")"
+. "${ci_dir}/ci-common.sh"
+
+git_download metacoq
+
+( cd "${CI_BUILD_DIR}/metacoq" && make && make test-suite && make translations && make install)
```
SkySkimmer: Why not make everything in one go?

mattam82: This is a multi-make project, I didn't try but suspect it would not respect the order

ejgallego (commented):
Things look OK even if a bit heavy, I assume that you folks upstream have the proper CI to keep things glued properly. A couple of comments:

- I'd much prefer if you would fix all the OCaml warnings in the build, this is still under discussion but the plan is to require plugins to adhere to the same compiler flags than the main tree for being in the CI.
- What is the plan for having metaCoq depend on released versions of the embedded libraries so they could be built from upstream?

[Maybe @Zimmi48 this is an example where having multiple repos doesn't work so well for them as to maintain coherence]

mattam82:
Most ocaml warnings you see are generated by the compilation of extracted programs (unused opens), I can fix the rest. 

What do you mean by embedded libraries? The whole project depends on nothing else than Coq and Ocaml. We plan to release separate packages for separate functionalities but indeed, this will probably stay monorepo to maintain coherence, those are tightly knit developments.

ejgallego:
> Most ocaml warnings you see are generated by the compilation of extracted programs (unused opens), I can fix the rest.

Oh, the extracted warnings shouldn't matter indeed.

> What do you mean by embedded libraries?

I mean the "various packed ocaml libraries", somehow I thought you were grabbing them from different repos sorry.

Zimmi48:
>[Maybe @Zimmi48 this is an example where having multiple repos doesn't work so well for them as to maintain coherence]

I think the confusion has been cleared out but this is a case of a development using a monorepo, and since it is the same team maintaining the various packages, it is clearly a net win.

ejgallego:
For info, I am holding this for a bit as there are several large-scope PRs [Sprop, proof_global, etc...] that require coordination with this one as they would mess the CI status.

Given that the PRs we are talking about are ready and have been opened for quite a long time they do take priority over this one in my view, but of course we can coordinate in a different way.

ejgallego:
Note the rationale for the order works as follow: SProp, Proof_global, etc... are PR that should be merged way before the beta, so if this PR would delay them like 10 days that could be fatal.

On the other hand, this PR can be merged even in the beta phase so no such problem.

In fact, it would make sense to recommend new CI entries to be submitted in the beta period, but with 2 releases a year makes the window a bit narrow IMHO.

Zimmi48:
> In fact, it would make sense to recommend new CI entries to be submitted in the beta period, but with 2 releases a year makes the window a bit narrow IMHO.

I'd prefer that there be no criterion restricting when people can submit PRs. It is then up to the CI maintainers to decide when is the right time to merge.

ejgallego:
A recommendation is not restriction, but a hint, right? I guess indeed we may say nothing, I guess it will depend on each case.

ejgallego:
Just to be clear _I propose_ to delay a bit this PR as not to interfere with the ongoing 8.10 PRs, but it is just an idea, we can merge now if you folks think that's going to work better.

mattam82:
I'm fine with delaying. Anyway, we will continue developing MetaCoq on 8.9 branch until mid april I think. Indeed SProp should have priority. Proof_global will be far less work I anticipate :)

ejgallego:
Almost all the important stuff is in, only waiting on #9129

ejgallego:
No more stuff blocking this PR.

Zimmi48:
But a rebase is needed @mattam82 @ejgallego.

ejgallego:
> But a rebase is needed @mattam82 @ejgallego.

This is on the hands of @mattam82 and the rest of MetaCoq developers as they first need to make it work with master; I'll let them set the milestone.

mattam82:
We'll wait until after the branching, right now we're concentrating on the 8.8/8.9 versions anyway

ejgallego:
Cleaning up my "assigned" list, please ping when activity resumes.

dev/ci/ci-metacoq.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+ci_dir="$(dirname "$0")"
+. "${ci_dir}/ci-common.sh"
+
+git_download metacoq
+
+( cd "${CI_BUILD_DIR}/metacoq" && ./configure.sh local && make ci-local && make install )
```
SkySkimmer: It seems the `ci-local` target also does configure

mattam82: It shouldn't, I'll fix that upstream.

mattam82:
@SkySkimmer any idea why validate:quick failed?

mattam82:
So this is ready I believe.

mattam82:
@coq/ci-maintainers ?

.gitlab-ci.yml
```diff
@@ -722,10 +722,24 @@ plugin:ci-elpi:
 
 plugin:ci-equations:
   extends: .ci-template
+  artifacts:
+    name: "$CI_JOB_NAME"
+    paths:
+      - _build_ci
```
SkySkimmer: Had to add this

SkySkimmer:
The log seems a bit long for gitlab, can you make it quieter somehow?

mattam82:
Done, it should be reasonable now

mattam82:
I've relaunched the ci job to test that.

mattam82:
Funny enough, it seems that rebase has broken a proof in metacoq...

mattam82:
@SkySkimmer the pipeline actually succeeded https://gitlab.com/coq/coq/pipelines/128238914 and https://gitlab.com/coq/coq/-/jobs/479938862 I just canceled other unrelated jobs

SkySkimmer:
Still pretty long imo, do you really need to print all those terms?

mattam82:
I can fix the printings from the test-suite. Anyway, Coq broke it again... I'll ping when this is ready again.

mattam82:
@SkySkimmer This should be ok now, let's wait for CI one last time

SkySkimmer (approved):

PR #9609, opened 2019-02-20
Add check for inductive privacy in kernel

### LINKED ISSUE #9608: `Private Inductive` does not prevent generating a match
I believe the following should not be accepted:

```coq
Module M.
Private Inductive foo := C : foo -> foo.
End M.

Scheme Induction for M.foo Sort Prop.
```

I'm very surprised because privacy is seen by the kernel, so how come it can be circumvented?

SkySkimmer:
Backward compatible overlay https://github.com/HoTT/HoTT/pull/1006

Fix #9608

### DIFF
--- a/checker/checker.ml
+++ b/checker/checker.ml
@@ -268,6 +268,7 @@ let explain_exn = function
       | BadAssumption _ -> str"BadAssumption"
       | ReferenceVariables _ -> str"ReferenceVariables"
       | ElimArity _ -> str"ElimArity"
+      | CasePrivate _ -> str"CasePrivate"
       | CaseNotInductive _ -> str"CaseNotInductive"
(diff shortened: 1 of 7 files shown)


maximedenes:
Should fix #9608.

ppedrot:
FTR, rewrite rules in the Coq kernel might land sooner than what we think, and they would allow to implement HITs without relying on this kind of hacks...

maximedenes:
@SkySkimmer what's the status of this?

SkySkimmer:
Looks like CI failed spuriously, AFAIR it was ready
I need to look at the HoTT side discussion again though

maximedenes:
Sounds good, let me know when you want me to act on this.

SkySkimmer:
Don't have time/motivation to figure this out further.

PR #9612, opened 2019-02-20
Fix #9526: Registering inductives for primitive integers doesn't check enough

### LINKED ISSUE #9526: Registering inductives for primitive integers doesn't check enough
~~~coq
Primitive int := #int63_type.

Polymorphic Inductive badcarry (A:Type) : Type :=
| C0: A -> badcarry A
| C1: A -> badcarry A.
Register badcarry as kernel.ind_carry.
(* shouldn't be accepted, or kernel should support univ polymorphic carry *)

Primitive addc := #int63_addc.

Check addc _ _ : _.
(* Error: Conversion test raised an anomaly: Anomaly "Universe Var(0) undefined." *)
~~~
I don't know exactly how permissive the check should be, eg should we accept
~~~coq
(* currently accepted *)
Inductive badcarry (A:Type) : Prop :=
| C0: A -> badcarry A
| C1: A -> badcarry A.

(* currently rejected *)
Inductive badcarry (A:Set) : Set :=
| C0: A -> badcarry A
| C1: A -> badcarry A.

(* currently accepted *)
Inductive badcarry : Type -> Type  :=
| C0: forall A, A -> badcarry A
| C1: forall A, A -> badcarry A.
~~~

Of course this isn't limited to carry.

maximedenes:
@SkySkimmer I'd need your help to check that the sort is not `Prop`.

### DIFF
--- a/kernel/safe_typing.ml
+++ b/kernel/safe_typing.ml
@@ -1194,39 +1194,46 @@ let register_inline kn senv =
   let cb = {cb with const_inline_code = true} in
   let env = add_constant kn cb env in { senv with env}
 
-let check_register_ind (mind,i) r env =
-  let mb = Environ.lookup_mind mind env in
-  let check_if b s =
+let check_register_ind ind r env =
+  let (mb,ob as spec) = Inductive.lookup_mind_specif env ind in
+  let check_if b msg =
     if not b then
-      CErrors.user_err ~hdr:"check_register_ind" (Pp.str s) in
-  check_if (Int.equal (Array.length mb.mind_packets) 1) "A non mutual inductive is expected";
-  let ob = mb.mind_packets.(i) in
+      CErrors.user_err ~hdr:"check_register_ind" msg in
+  check_if (Int.equal (Array.length mb.mind_packets) 1) Pp.(str "A non mutual inductive is expected");
+  let is_monomorphic = function Monomorphic _ -> true | Polymorphic _ -> false in
+  check_if (is_monomorphic mb.mind_universes) Pp.(str "A universe monomorphic inductive type is expected");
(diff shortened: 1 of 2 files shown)


SkySkimmer:
Does the kernel actually care if the sort is Prop? As far as I can tell it only cares that it can instantiate the types and constructors, not that it can destruct them.

kernel/safe_typing.ml
```diff
@@ -1194,39 +1194,46 @@ let register_inline kn senv =
   let cb = {cb with const_inline_code = true} in
   let env = add_constant kn cb env in { senv with env}
 
-let check_register_ind (mind,i) r env =
-  let mb = Environ.lookup_mind mind env in
-  let check_if b s =
+let check_register_ind ind r env =
+  let (mb,ob as spec) = Inductive.lookup_mind_specif env ind in
```
SkySkimmer: If we do care you can ask that `List.mem Sorts.(family_eq InType) ob.mind_kelim` ie the inductive is not squashed.

BTW usually we call the one_inductive_body "mip" (p for packet since it comes from mind_packets) not "ob"

maximedenes:
> Does the kernel actually care if the sort is Prop? As far as I can tell it only cares that it can instantiate the types and constructors, not that it can destruct them.

Fair enough, let's drop the check.

maximedenes:
@vbgl can you be the assignee?

SkySkimmer:
For some reason you have unicode 0xFFFC "OBJECT REPLACEMENT CHARACTER" `￼` after the `:` in your commit message.

vbgl:
The failure of equations in CI seems unrelated.

vbgl (approved):

PR #9620, opened 2019-02-21
Stdlib HTML documentation: fix a few absolute URLs

vbgl:
This fixes the rendering of the HTML documentation of the stdlib when served from a different domain than `coq.inria.fr` (e.g., at https://coq.github.io/doc/master/stdlib/).


Zimmi48 (approved):
This is a good enough fix but not perfect either: it won't be of any use for an installed stdlib doc on a machine with no internet connection.

Zimmi48:
I'll merge this on Monday to leave time for comments.

vbgl:
I think the 8.9.1 milestone would be more appropriate (given the milestone of #9515).

PR #9621, opened 2019-02-21
[ide] only use Coq_config for the URL of the manual

### LINKED ISSUE #9617: Coq IDE 8.9.0 "Browse Coq manual" opens manual for 8.8.2
Coq IDE 8.9.0 menu item "Help -> Browse Coq manual" opens manual for 8.8.2 at https://coq.inria.fr/distrib/V8.8.2/refman/ .

gares:
Fix #9617

### DIFF
--- a/ide/ideutils.ml
+++ b/ide/ideutils.ml
@@ -458,15 +458,6 @@ let browse prerr url =
   in
   run_command (fun _ -> ()) finally com
 
-let doc_url () =
-  if doc_url#get = use_default_doc_url || doc_url#get = ""
-  then
-    let addr = List.fold_left Filename.concat (Envars.docdir ())
-      ["html";"refman";"index.html"]
-    in
-    if Sys.file_exists addr then "file://"^addr else Coq_config.wwwrefman
-  else doc_url#get
-
 let url_for_keyword =
   let ht = Hashtbl.create 97 in
     lazy (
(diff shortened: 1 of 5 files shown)


ppedrot:
Wasn't there a reason not to hardcode this link?

gares:
I can't find one different from "what feature could I add in the next 5 minutes and forget about it".
I mean, the code tries to use locally installed doc, hard codes the website at lix, ...
The current code is simple: you configure coq with a URL for the doc, this is the URL the user will get.

ppedrot:
I'm personally convinced, but I think I remember @herbelin taking great care of having a customizable URL. Let's summon him.

ejgallego (approved):

ppedrot:
Linting error looks spurious again, if nobody complains I'll merge as-is.

ide/ideutils.ml
```diff
@@ -458,15 +458,6 @@ let browse prerr url =
   in
   run_command (fun _ -> ()) finally com
 
-let doc_url () =
```
JasonGross: Removing this makes the documentation no longer work when disconnected from the internet, even if it's installed locally, right?  (I don't object, because I don't use CoqIDE, I just want to make sure this is deliberate.)

gares:
Yes

ide/ideutils.ml
ppedrot: I personally find it fairly nonsensical to have a link to a website inside a GUI. At best this helps total beginners who have no idea on where to find the doc, but apart from that, I really don't see why you wouldn't just open your favourite source of documentation by standard means.

gares: It makes (some) sense to have a link that points to the doc *for the version you are using*.
Or at least, this is the reason why I did not zap entirely this feature.

@JasonGross yes it is deliberate.

ppedrot:
Will probably merge by tomorrow, last round to save the feature.

ppedrot (approved):

gares:
thanks

PR #9622, opened 2019-02-21
Fix makefile deps for coqide

gares:
this dep was missing after the #6859

### DIFF
--- a/Makefile.ide
+++ b/Makefile.ide
@@ -70,7 +70,7 @@ SOURCEVIEWSHARE=$(shell pkg-config --variable=prefix gtksourceview-2.0)/share
 .PHONY: ide-toploop ide-byteloop ide-optloop
 
 # target to build CoqIde (native version) and the stuff needed to lauch it
-coqide: coqide-files coqide-opt theories/Init/Prelude.$(VO)
+coqide: coqide-files coqide-opt theories/Init/Prelude.$(VO) $(TOPBIN)
 
 # target to build CoqIde (in native and byte versions), and no more
 # NB: this target is used in the opam package coq-coqide
--- a/ide/coq.ml
+++ b/ide/coq.ml
@@ -119,7 +119,7 @@ let rec filter_coq_opts args =
 
 and asks_for_coqtop args =
   let pb_mes = GWindow.message_dialog
-    ~message:"Failed to load coqtop. Reset the preference to default ?"
+    ~message:"Failed to load coqidetop. Reset the preference to default ?"
     ~message_type:`QUESTION ~buttons:GWindow.Buttons.yes_no () in
   match pb_mes#run () with
     | `YES ->
@@ -130,7 +130,7 @@ and asks_for_coqtop args =
     | `DELETE_EVENT | `NO ->
       let () = pb_mes#destroy () in
       let cmd_sel = GWindow.file_selection
-	~title:"Coqtop to execute (edit your preference then)"
+        ~title:"coqidetop to execute (edit your preference then)"
 	~filename:(coqtop_path ()) ~urgency_hint:true () in
       match cmd_sel#run () with
 	| `OK ->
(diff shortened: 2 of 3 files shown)


Makefile.ide
```diff
@@ -70,7 +70,7 @@ SOURCEVIEWSHARE=$(shell pkg-config --variable=prefix gtksourceview-2.0)/share
 .PHONY: ide-toploop ide-byteloop ide-optloop
 
 # target to build CoqIde (native version) and the stuff needed to lauch it
-coqide: coqide-files coqide-opt theories/Init/Prelude.$(VO)
+coqide: coqide-files coqide-opt theories/Init/Prelude.$(VO) $(TOPBIN)
```
SkySkimmer: coqide doesn't depend on coqtop does it? I guess it's not too bad to be imprecise about it.

gares:
You are right, TOPBIN is coqtop + its dependencies, that happen to be the dependencies of all other binaries based on the STM, namely coqc and coqtop.

If you want to improve this fix you are welcome :-)

ppedrot (approved):

ppedrot:
@gares it seems the linter complains.

gares:
I don't care.

gares:
The effort to checkout, rebase, push, CI is, in terms of time and energy consumed, is simply not worth what we gain in listening the linter. Even the time  it takes to write this message is too much for this. I said already a thousands of times, the white space linter is silly, merge this as is, find someone that fixes it, or close it... I'm fine with it.

JasonGross:
Can we give the linter a `--fix` mode, and add a tag that will tell @coqbot to `rebase` the PR on top of the tip of its base branch and run the linter in fixup mode on each commit, and then force-push?  (Cc @Zimmi48?)  This way the only thing that needs to be done to fix most linter issues is to add a tag or write a comment to @coqbot?

ejgallego:
@JasonGross the default repos config already has a pre-commit hook that takes care of this.

It is fairly rare to run into problems unless you have opted out.

Zimmi48:
The pre-commit hook is installed by the configure script, which will soon disappear when we switch to dune only. I wonder what that will mean then. In this specific case, the linter detected a tab that was there before and after the commit. I wonder then, @gares do you have the pre-commit hook installed? If this is a fresh clone on which you have only used Dune to build, then you most likely haven't installed the pre-commit hook. You are complaining about the effort created by the linter, and I agree that it should be reduced to zero, but it is generally the case already.

ppedrot:
I personally don't care too much about the angry but pretty stupid gods of linting, the problem is more about social acceptance of merging PRs that do not pass the CI...

ejgallego:
> the problem is more about social acceptance of merging PRs that do not pass the CI...

we do that all the time [for example due to spurious failures]

> The pre-commit hook is installed by the configure script, which will soon disappear when we switch to Dune only.

Rather the configure script runs in a sandbox so it doesn't touch `.git`, but we can indeed have a method to ensure the repository is properly set up for Dune users.

gares:
I have the hook installed since day one. I don't have time to debug why it did not work.

Acceptance is away if you see what the point is. I m on strike.

maximedenes:
I'll fix the whitespace and force push.

maximedenes:
done

maximedenes:
> The effort to checkout, rebase, push, CI is, in terms of time and energy consumed, is simply not worth what we gain in listening the linter. Even the time it takes to write this message is too much for this. I said already a thousands of times, the white space linter is silly, merge this as is, find someone that fixes it, or close it... I'm fine with it.

Since I've been complaining about the linter too, let me add that I do think getting rid of tabs is useful (just look at the state of indentation in certain files). Also, it is trivial to fix it automatically (`git rebase --whitespace=fix`), so I don't see why we should ignore these errors.

End-of-line whitespace and end-of-file newlines are much more debattable IMHO, although I got used to being careful about it.

gares:
thanks all!

Issue #9199, opened 2018-12-11
Dune build: don't require a bash for building


MSoegtropIMC:
On Windows having a Posix compliant shell is difficult. @ejgallego pointed out that it wouldn't be difficult to have a Dune based build system for Coq projects which would not require aPosix shell. This would greatly simplify building Coq projects on Windows.

I Labeled this as feature because on Windows it would indeed be a valuable user facing feature to have a reliable build system, which does not require a bash.

SkySkimmer:
Posix shell and bash are not the same thing, which do you mean?

ejgallego:
Places to modify come from #8421 :+1:
```
(action (bash "./make_jumptbl.sh %{h-file} %{targets}")))
(action (bash "./make_opcodes.sh %{h-file} %{targets}")))
```
all that we need is to write an OCaml program doing the same; this should be easy [and even better than the sh solution]

The test suite is another story, the Windows build of Dune works quite well as of today, however the test-suite is having some path problems I need to solve.

MSoegtropIMC:
@SkySkimmer : of cause one should restrict one self to Posix and not using and bash specific stuff. But even having a reasonably small Posix subset on Windows is hard. I actually started to write a Posix sibset shell for system functions in programming languages cause it is such a pain.

@ejgallego : as I said I already started to write a shell replacement intended to replace the OCaml system function.

ejgallego:
I am not rewriting this couple of scripts for now; however that should be pretty easy for anyone willing to have the shell-free build on Win, but indeed that's better by a windows user.

By the way @MSoegtropIMC you may be interested on https://discuss.ocaml.org/t/preferred-ocaml-windows-environments-mainly-for-continuous-integration/3299/3

ejgallego:
@MSoegtropIMC so this was basically solved by #9624 , even if we have a couple of minor places that still use bash, these should be very easy to fix. In fact we won't depend on bash neither on a shell.

So the big question is, how do we actually test the bash-less build?

MSoegtropIMC:
> So the big question is, how do we actually test the bash-less build?

Well on Windows this is easy. My build scripts create a MinGW Coq in an folder outside of Cygwin so that it can be started without having a cygwin in the path. This folder than servers as a template for creating the installer. If Coq is started from this folder there there is no bash or other shell (besides CMD).

ejgallego:
Will ocaml even work in that case?

[...]

PR #9624, opened 2019-02-21
[Kernel] Simpler generation of opcode files

vbgl:
Files kernel/copcodes.ml, kernel/byterun/coq_instruct.h, and
kernel/byterun/coq_jumptbl.h are generated by a simple OCaml program
rather than a pipeline of sed and awk text processing.


ejgallego:
cc: #9199

kernel/byterun/dune
```diff
@@ -4,7 +4,12 @@
...
+ (targets coq_instruct.h)
+ (deps (:gen ../genOpcodeFiles.exe))
+ (action (run %{gen} "enum")))
+
 (rule
  (targets coq_jumptbl.h)
- (deps (:h-file coq_instruct.h) make_jumptbl.sh)
- (action (bash "./make_jumptbl.sh %{h-file} %{targets}")))
+ (deps (:gen ../genOpcodeFiles.exe))
+ (action (run %{gen} "jump")))
```
ejgallego: You can also do this in a single rule if you want, also gen is not strictly needed:
```
(rule
  (targets coq_instruct.h coq_jumptbl.h)
  (action (run ../genOpcodefiles.exe jump enum))
```
or
```
(action (progn (run ../... ) (run ...))
```
You can also scope the binary using the binary scoping feature, this allows you to extend path so you don't have to use `../gen`...

kernel/genOpcodeFiles.ml
```diff
@@ -0,0 +1,187 @@
+(** List of opcodes.
```
ejgallego: Missing header.

ejgallego (approved):
Looks pretty good!

vbgl:
Thanks Emilio. I’ve just added a header (also to the sibling `write_uint63.ml`) and merged two dune rules.

ejgallego:
> Thanks Emilio. I’ve just added a header (also to the sibling `write_uint63.ml`) and merged two dune rules.

Cool, I guess that the assignee here should be someone from the kernel team, however this could also be regarded as an infrastructure change I'd be happy to be the assignee here if kernel people think it's OK, just assign me in that case.

Makefile.build
```diff
@@ -316,11 +316,17 @@ $(LIBCOQRUN): kernel/byterun/coq_jumptbl.h $(BYTERUN)
...
+
+kernel/byterun/coq_jumptbl.h: kernel/genOpcodeFiles.ml
+	$(SHOW)'WRITE $@'
+	$(HIDE)(cd kernel/byterun && ocaml ../$(shell basename $<) jump)
 
-kernel/copcodes.ml: kernel/byterun/coq_instruct.h kernel/make_opcodes.sh kernel/make-opcodes
-	kernel/make_opcodes.sh $< $@
+kernel/copcodes.ml: kernel/genOpcodeFiles.ml
+	$(SHOW)'WRITE $@'
+	$(HIDE)(cd kernel && ocaml $(shell basename $<) copml)
```
gares: IIRC we were doing so for configure and ocaml would not abort on errors (just skip the broken toplevel sentence).
If it is the case, please use ocamlc first.

gares (commented):
please clarify what happens if the ml script breaks and the make based build system is used

ppedrot (approved):
FTR I like this change.

vbgl:
I’ve address Enrico’s remark for the case when the OCaml script fails at compile time. I don’t know what to do for run-time errors. `make` succeeds even if the OCaml program raises an exception and does not produce the target file…

ppedrot:
@vbgl Can't you test on the return value? It seems that failing OCaml programs return 2 on the shell.

vbgl:
Let me take a concrete example. I patch the generator program so that it fails to produce `copcodes.ml` and returns a non-zero value:

~~~
$ rm -f kernel/copcodes.ml
$ make kernel/cemitcodes.cmx
WRITE kernel/copcodes.ml
usage: ./genOpcodeFiles.exe [enum|jump|copml]
OCAMLOPT  kernel/cemitcodes.ml
File "kernel/cemitcodes.ml", line 15, characters 5-10:
Error: Unbound module Names

$ ./kernel/genOpcodeFiles.exe copml
usage: ./kernel/genOpcodeFiles.exe [enum|jump|copml]
$ echo $?
127
~~~

In the call to `make` above, the `OCAMLOPT` step should not be attempted as a dependency is missing. I believe the problem is neither in the generator program nor in the rule for `copcodes.ml`.

SkySkimmer:
Doesn't the same thing happen with the shell based stuff? That's just how using ocamldep to generate make dependencies dynamically works.

Zimmi48 (approved):
Nix change trivially OK. Otherwise, quite supportive of the idea.

kernel/genOpcodeFiles.ml
```diff
@@ -0,0 +1,197 @@
...
+      Format.fprintf fmt "let op%s = %d@.@." s n
+    ) opcodes
+
+let with_output_file f k =
+  let oc = open_out f in
+  oc |> Format.formatter_of_out_channel |> k;
+  close_out oc
+
+let usage () =
+  Format.eprintf "usage: %s [enum|jump|copml]@." Sys.argv.(0)
```
gares: you should exit 1 here, don't you?

gares:
> is neither in the generator program

It seems you don't exit 1 if the argument parsing goes wrong, I don't get how you see 127 there. IMO make sees 0.

vbgl:
In the example I reported about earlier, `make` saw 127.

gares:
What you print is the exit code of make, not the one of ocaml

vbgl:
Enrico, please read again the first two sentences of my comment above: https://github.com/coq/coq/pull/9624#issuecomment-466433067

What I print is the exit code of the broken code generator. No make and no OCaml involved.

kernel/byterun/dune
```diff
@@ -5,6 +5,6 @@
  (c_names coq_fix_code coq_memory coq_values coq_interp))
 
 (rule
- (targets coq_jumptbl.h)
- (deps (:h-file coq_instruct.h) make_jumptbl.sh)
- (action (bash "./make_jumptbl.sh %{h-file} %{targets}")))
+ (targets coq_instruct.h coq_jumptbl.h)
+ (deps (:gen ../genOpcodeFiles.exe))
+ (action (progn (run %{gen} enum) (run %{gen} jump))))
```
rgrinberg: I assume that `%{gen} enum` produces the first target, while `%{gen} jump` produces the second target. If that's the case, your rule is correct but it's less optimal than doing it in two separate rules. Now dune will fail to parallelize these 2 actions.

vbgl: Thanks for the tip.

kernel/genOpcodeFiles.ml
vbgl: Indeed. Done. Thanks.

kernel/byterun/dune
ejgallego: Yup indeed I was thinking of actually having a single `%{gen} jump gen` call [as to avoid a double process spawn but indeed two rules may be better in the end.

gares:
> Enrico, please read again the first two sentences

Sorry, I misread the code snippet. Should I still try to figure out why make does not bail out as expected when the codegen script fails?

vbgl:
> Sorry, I misread the code snippet.

No problem.

> Should I still try to figure out why make does not bail out as
> expected when the codegen script fails?

I’d say that you should not spend too much time investigating that
issue. At worse, one could get surprising error messages that should
vanish after running `git clean` or the like.

ejgallego:
Taking this as per merge policy [no assignee in 4 days] and given that I have been interested in build infrastructure for a while.

I think the current make situation is enough, however see my comment.

Makefile.build
```diff
@@ -316,11 +316,21 @@ $(LIBCOQRUN): kernel/byterun/coq_jumptbl.h $(BYTERUN)
...
+
+kernel/byterun/coq_instruct.h: kernel/genOpcodeFiles.exe
+	$(SHOW)'WRITE $@'
+	$(HIDE)(cd kernel/byterun && ../$(shell basename $<) enum)
 
-kernel/copcodes.ml: kernel/byterun/coq_instruct.h kernel/make_opcodes.sh kernel/make-opcodes
-	kernel/make_opcodes.sh $< $@
+kernel/byterun/coq_jumptbl.h: kernel/genOpcodeFiles.exe
+	$(SHOW)'WRITE $@'
+	$(HIDE)(cd kernel/byterun && ../$(shell basename $<) jump)
```
ejgallego: I am not make expert but a couple of comments on this rule:
- why do you spawn a subshell with `(...)` ? As far as I know make already does spawn a fresh shell for each line. In fact, this is likely the reason the output status was masked. AFAICT the right thing to do is to remove the `()`
- I'd also prefer if the `genOpcodeFiles` would output to stdout. This makes renaming and moving around files much easier, it should also simplify the code as you don't need to open a file with all the complex error handling than has to be done, and moreover you wouldn't need to use `cd` and `basename` as something like this should work:
` $< jump > $@`

kernel/write_uint63.ml
```diff
@@ -1,10 +1,18 @@
...
+(*         *   The Coq Proof Assistant / The Coq Development Team       *)
+(*  v      *   INRIA, CNRS and contributors - Copyright 1999-2018       *)
+(* <O___,, *       (see CREDITS file for the list of authors)           *)
+(*   \VV/  **************************************************************)
+(*    //   *    This file is distributed under the terms of the         *)
+(*         *     GNU Lesser General Public License Version 2.1          *)
+(*         *     (see LICENSE file for the text of the license)         *)
+(************************************************************************)
 
+(** Equivalent of rm -f *)
```
ejgallego: Indeed now that I see this [it slipped on the native review] you could also remove all this code and just use the primitive `with-stdout-to` to write the file, gaining in both robustness of the generated file layout and avoiding having to mess with files here.

ejgallego:
> I think the current make situation is enough, however see my comment.

Indeed, after a bit of thinking, if you think my suggestions make sense, then I'd prefer to merge a revised version.

vbgl:
I’ve updated the code generators to make them write to their standard output. Thanks Emilio for the hint regarding `with-stdout-to`!

ejgallego:
Cool @vbgl , thanks! I'll merge ASAP the CI is ready.

Very minor nit about the last commit, the commit message does indeed seem like too generic, right? Would you mind tweaking it so `git log --short` is readable?

Makefile.build
```diff
@@ -316,11 +316,21 @@ $(LIBCOQRUN): kernel/byterun/coq_jumptbl.h $(BYTERUN)
 	cd $(dir $(LIBCOQRUN)) && \
 	$(OCAMLFIND) ocamlmklib -oc $(COQRUN) $(notdir $(BYTERUN))
 
-kernel/byterun/coq_jumptbl.h : kernel/byterun/coq_instruct.h kernel/byterun/make_jumptbl.sh
-	kernel/byterun/make_jumptbl.sh $< $@
+kernel/genOpcodeFiles.exe: kernel/genOpcodeFiles.ml
+	$(SHOW)'OCAMLC    $<'
+	$(HIDE)cd $(dir $<) && $(OCAMLC) -o $(notdir $@) $(notdir $<)
```
SkySkimmer: Why this cd/notdir dance?

Makefile.build
```diff
@@ -349,7 +359,7 @@ $(COQPP): $(COQPPCMO) coqpp/coqpp_main.ml
 ###########################################################################
 kernel/uint63.ml: kernel/write_uint63.ml kernel/uint63_x86.ml kernel/uint63_amd64.ml
 	$(SHOW)'WRITE $@'
-	$(HIDE)(cd kernel && ocaml unix.cma $(shell basename $<))
+	$(HIDE)cd kernel && ocaml unix.cma $(shell basename $<) > $(shell basename $@)
```
SkySkimmer: Why cd/basename?

Makefile.build
ejgallego: Yup, indeed you can simplify that @vbgl if you want to just `ocamlc -o $@ $<`

Makefile.build
ejgallego: Indeed, same, I had missed the parts about uint63.

vbgl: No. That would break it.

The `write_uint63` programs looks for original source files in the current work directory.

SkySkimmer: Oh this program is just `cp`
So use `cp` don't reinvent the wheel

vbgl: Why are you guys hijacking this PR? I’ll drop the commit about write-uint63 and you can discuss that in a separate issue.

ejgallego: Another option is to pass the directory to look in as an argument; anyways, I am fine with this as-is.

ejgallego: > So use `cp` don't reinvent the wheel

The program does some testing that it is better done in OCaml, but indeed has a reimplementation of cp. We should be able to remove this once we are dune only with a conditional `copy_file` rule as Dune exposes `ocaml -config`, but not yet IMO.

ejgallego: > Why are you guys hijacking this PR? I’ll drop the commit about write-uint63 and you can discuss that in a separate issue.

We are not hijacking the PR, we are providing (optional as I noted) feedback on the current implementation.

ejgallego: > We are not hijacking the PR, we are providing (optional as I noted) feedback on the current implementation.

But I know how it feels as I've been in your position many times, with PRs blocked for a long time for what you call "hickjacking" reviews. [In fact I have a few now]

Just note that it is not the case here as a) I took this PR despite being really a task for kernel maints, b) the tick is still green so that means our request are not blocking.

I would tho fix the commit message, that seems maybe a bit more serious to me.

ejgallego:
Great, thanks a lot!

PR #9632, opened 2019-02-22
Fix #9631: Instance: anomaly grounding non evar-free term

### LINKED ISSUE #9631: Instance: anomaly grounding non evar-free term
`Instance x : _.` and variants

SkySkimmer:


vernac/himsg.ml
```diff
@@ -1041,8 +1041,7 @@ let explain_module_internalization_error = function
 
 let explain_not_a_class env c =
   let sigma = Evd.from_env env in
-  let c = EConstr.to_constr sigma c in
-  pr_constr_env env sigma c ++ str" is not a declared type class."
+  pr_econstr_env env sigma c ++ str" is not a declared type class."
```
ppedrot: @SkySkimmer just realized this was nonsense, but I'm writing the reminder: this should take an evarmap as an argument rather than creating a dummy one out of thin air, otherwise the fix won't work properly.

ejgallego: The detyping path is resistant to undefined vars so this kinda works.

SkySkimmer: Here it would be seen as an undeclared evar. Also if it's supposed to be defined it will say some nonsense.

SkySkimmer: Changed.

ejgallego:
18 days a no assignee.

ejgallego (approved):
LGTM

PR #9635, opened 2019-02-23
[ci] [ocaml] Support OCaml 4.08

ejgallego:
The main change is the renaming of `Pervasives -> Stdlib`.

Some `Format` functions have been deprecated, however we have no
backwards alternative so we need to disable the warnings.

Builds now depend on the `stdlib-shims` package in order to be
compatible with OCaml < 4.07.0.

### DIFF
--- a/gramlib/grammar.ml
+++ b/gramlib/grammar.ml
@@ -876,7 +876,7 @@ and print_level : type s. _ -> _ -> s ex_symbols list -> _ =
       (fun sep (ExS rule) ->
          fprintf ppf "%t%a" sep print_rule rule;
          fun ppf -> fprintf ppf "%a| " pp_print_space ())
-      (fun ppf -> ()) rules
+      (fun ppf -> ()) rules ppf
   in
   fprintf ppf " ]@]"
 
@@ -901,7 +901,7 @@ let print_levels ppf elev =
          fprintf ppf "@]@;<1 2>";
          print_level ppf pp_force_newline rules;
          fun ppf -> fprintf ppf "@,| ")
-      (fun ppf -> ()) elev
+      (fun ppf -> ()) elev ppf
   in
   ()
 
--- a/interp/dumpglob.ml
+++ b/interp/dumpglob.ml
@@ -12,13 +12,13 @@ open Util
 
 (* Dump of globalization (to be used by coqdoc) *)
 
-let glob_file = ref Pervasives.stdout
+let glob_file = ref Stdlib.stdout
 
 let open_glob_file f =
-  glob_file := Pervasives.open_out f
+  glob_file := Stdlib.open_out f
 
 let close_glob_file () =
-  Pervasives.close_out !glob_file
+  Stdlib.close_out !glob_file
 
 type glob_output_t =
     | NoGlob
@@ -37,7 +37,7 @@ let dump_to_dotglob () = glob_output := MultFiles
 
 let dump_into_file f =
   if String.equal f "stdout" then
-    (glob_output := StdOut; glob_file := Pervasives.stdout)
+    (glob_output := StdOut; glob_file := Stdlib.stdout)
   else
     (glob_output := File f; open_glob_file f)
 
@@ -45,7 +45,7 @@ let feedback_glob () = glob_output := Feedback
 
 let dump_string s =
   if dump () && !glob_output != Feedback then 
-    Pervasives.output_string !glob_file s
+    Stdlib.output_string !glob_file s
 
 let start_dump_glob ~vfile ~vofile =
   match !glob_output with
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,7 +10,7 @@ stages:
 variables:
   # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
   # for reference]
-  CACHEKEY: "bionic_coq-V2019-06-11-V1"
+  CACHEKEY: "bionic_coq-V2019-06-16-V1"
   IMAGE: "$CI_REGISTRY_IMAGE:$CACHEKEY"
   # By default, jobs run in the base switch; override to select another switch
   OPAM_SWITCH: "base"
--- a/INSTALL
+++ b/INSTALL
@@ -31,7 +31,7 @@ WHAT DO YOU NEED ?
 
    - OCaml (version >= 4.05.0)
      (available at https://ocaml.org/)
-     (This version of Coq has been tested up to OCaml 4.07.0)
+     (This version of Coq has been tested up to OCaml 4.08.0)
 
    - The Num package, which used to be part of the OCaml standard library,
      if you are using an OCaml version >= 4.06.0
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -67,7 +67,7 @@ jobs:
       opam list
     displayName: 'Install OCaml dependencies'
     env:
-      COMPILER: "4.07.1"
+      COMPILER: "4.08.0"
       FINDLIB_VER: ".1.8.0"
       OPAMYES: "true"
 
--- a/checker/analyze.ml
+++ b/checker/analyze.ml
@@ -395,8 +395,8 @@ end
 module IChannel =
 struct
   type t = in_channel
-  let input_byte = Pervasives.input_byte
-  let input_binary_int = Pervasives.input_binary_int
+  let input_byte = Stdlib.input_byte
+  let input_binary_int = Stdlib.input_binary_int
 end
 
 module IString =
--- a/checker/check.ml
+++ b/checker/check.ml
@@ -66,7 +66,7 @@ module LibraryOrdered =
   struct
     type t = DirPath.t
     let compare d1 d2 =
-      Pervasives.compare
+      Stdlib.compare
         (List.rev (DirPath.repr d1)) (List.rev (DirPath.repr d2))
   end
 
--- a/dev/ci/azure-opam.sh
+++ b/dev/ci/azure-opam.sh
@@ -2,7 +2,7 @@
 
 set -e -x
 
-OPAM_VARIANT=ocaml-variants.4.07.1+mingw64c
+OPAM_VARIANT=ocaml-variants.4.08.0+mingw64c
 
 wget https://github.com/fdopen/opam-repository-mingw/releases/download/0.0.0.2/opam64.tar.xz -O opam64.tar.xz
 tar -xf opam64.tar.xz
@@ -10,4 +10,4 @@ bash opam64/install.sh
 
 opam init default -a -y "https://github.com/fdopen/opam-repository-mingw.git#opam2" -c $OPAM_VARIANT --disable-sandboxing
 eval "$(opam env)"
-opam install -y num ocamlfind dune ounit
+opam install -y num ocamlfind dune stdlib-shims ounit
--- a/dev/ci/docker/bionic_coq/Dockerfile
+++ b/dev/ci/docker/bionic_coq/Dockerfile
@@ -1,4 +1,4 @@
-# CACHEKEY: "bionic_coq-V2019-06-11-V1"
+# CACHEKEY: "bionic_coq-V2019-06-16-V1"
 # ^^ Update when modifying this file.
 
 FROM ubuntu:bionic
@@ -22,7 +22,7 @@ RUN pip3 install sphinx==1.7.8 sphinx_rtd_theme==0.2.5b2 \
                  antlr4-python3-runtime==4.7.1 sphinxcontrib-bibtex==0.4.0
 
 # We need to install OPAM 2.0 manually for now.
-RUN wget https://github.com/ocaml/opam/releases/download/2.0.3/opam-2.0.3-x86_64-linux -O /usr/bin/opam && chmod 755 /usr/bin/opam
+RUN wget https://github.com/ocaml/opam/releases/download/2.0.4/opam-2.0.4-x86_64-linux -O /usr/bin/opam && chmod 755 /usr/bin/opam
 
 # Basic OPAM setup
 ENV NJOBS="2" \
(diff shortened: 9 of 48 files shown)


SkySkimmer:
Don't we already compile with 08? Is it worth adding a dependency just to get rid of some warnings?

ejgallego:
> Don't we already compile with 08?

With trunk, but it is close.

> Is it worth adding a dependency just to get rid of some warnings?

No other choice other than require OCaml >= 4.07.0

SkySkimmer:
>No other choice other than require OCaml >= 4.07.0

I was thinking of ignoring the warnings like you did for the format stuff.

ejgallego:
> I was thinking of ignoring the warnings like you did for the format stuff.

In the case of format there is no migration path yet, so there is not other choice, however for `Stdlib` the situation is different.

ejgallego:
We like or not, upstream has gone this way [no more `Pervasives`] so the sooner we accept it IMHO the better.

JasonGross:
Does this change which Debian packages I need to install to build Coq without opam?  (Is this going to break the launchpad packages I build?)

ejgallego:
> Does this change which Debian packages I need to install to build Coq without opam? (Is this going to break the launchpad packages I build?)

Likely yes, as this requires a new compatibility package. 

You may also use Dune to build Coq and vendor the shim.

ejgallego:
> You may also use Dune to build Coq and vendor the shim.

Would we have done the switch, so many things would be so easier, including this PR........

JasonGross:
Which Ubuntu package carries stdlib-shims?  I can't find it on package search (https://packages.ubuntu.com/search?keywords=Stdlib-shims&searchon=all&suite=all&section=all)

If I backport dune 1.6.2, is that new enough?  And what do you mean by "vendor the shim"?

ejgallego:
> Which Ubuntu package carries stdlib-shims?

There is no Ubuntu package for `stdlib-shims`, the package was just released a couple of days ago by upstream.

> If I backport dune 1.6.2, is that new enough?

Yes

> And what do you mean by "vendor the shim"?

You can just drop a copy of the shim package in `vendor/` and Dune will drop the external dependency and use the "vendored" one automatically.

Note that it is not sure we will merge this PR as-is. It could be well that we drop the dependency on the shim and just remove the warnings.

JasonGross:
> Note that it is not sure we will merge this PR as-is. It could be well that we drop the dependency on the shim and just remove the warnings.

That would be my strong preference.  I should really get a ci job that tests this so I don't have to track these things manually...

ejgallego:
> I was thinking of ignoring the warnings like you did for the format stuff.

@SkySkimmer ignoring the warnings is IMHO not an option. Removing them by dropping the qualification could work indeed.

JasonGross (changes requested):
Please do not add dependencies without Ubuntu packages for OCaml 4.05 (the newest available via package).  I will work on getting my launch pad job into the ci here so that this is a more automated check.

ejgallego:
> Please do not add dependencies without Ubuntu packages for OCaml 4.05 (the newest available via package). I will work on getting my launch pad job into the ci here so that this is a more automated check.

I am not sure I will support that CI test as discussed in some other issues. OCaml in Ubuntu is unmaintained so that poses a huge debt on the Coq developers.

Please open an issue to discuss this before submitting the CI PR.

ejgallego:
> Please open an issue to discuss this before submitting the CI PR.

From my point of view, I think there is already too much packaging in the core Coq repository and I don't like it. We cannot satisfy every packing system, and while for historical reasons we have respected for example Debian, this is not guaranteed to hold in the future.

For example, the LSP server or SerAPI have no hope on working in such an outdated environment.

We already ship a build system that provides vendoring for OCaml libs, thus packages should be able to take profit of it easily would they need it.

ejgallego:
After thinking a bit more about packaging, something that I'd indeed support is a bootstrapping mode, that could be very useful for packagers.

For example, with https://github.com/coq/coq/pull/8615 , you wouldn't even have to backport Dune @JasonGross 

So packages could do `make bootstrap WITH_PKGS=dune,shim,foo` and have the makefile take care of the rest in the manner of the above PR.

Unfortunately that approach got a strong opposition from some other developers on grounds that I cannot still understand, but IMO could be a way to go.

JasonGross:
I would be in favor of having a bootstrapping mode available.  Even if the bootstrapping mode was just a shell script in the test-suite (and there was a test to ensure that, after running this script, Coq could be built), I think that would be enough to satisfy my desires for my packaging.  @bbaren might have different desires for packaging that is supposed to be less kludgey

ejgallego:
I am all for adding a bootstrapping mode but lack the energy to do after a blockage of a few months in what I regard non-rational / technical grounds. Bootstrapping is very useful not only for packagers but also for users, for example for those on Windows where opam may be unstable.

You can take the contents of #8615 tho and that should work for your Deb.

ejgallego:
Regarding adding a deb build to the CI, I dunno; I remain scared of it: it should not be the responsibility of Coq developers to have to fix the Debian / Ubuntu packaging.

As of today, Coq CI supports only one official development environment and that is Docker + OPAM based. It is not perfect but it does what we need without a lot of fuss.

Given that a large part of the developers also use NixOS they do maintain the nix development but that needs quite a bit of manpower; without these key developers such build would have been out of the CI.

So I dunno what should we do for deb, it is certainly an important platform, but Debian is so behind in OCaml stuff...

bbarenblat:
As a Debian developer, it makes my work easiest if there is some mode I can put the build system into that ensures it will neither access the internet nor use any vendored dependencies. (By policy, Debian packages must not do either of those things.) I therefore have no quarrel with a bootstrapping mode as long as it remains optional; I think it’s unlikely that the Debian packaging will use it.

SkySkimmer:
>@SkySkimmer ignoring the warnings is IMHO not an option.

Why not?

gares:
> The main change is the renaming of `Pervasives -> Stdlib`.

There are some places where `Pervasives.compare -> compare` (not `Stdlib.compare`). Is it intended?

ejgallego:
> There are some places where `Pervasives.compare -> compare` (not `Stdlib.compare`). Is it intended?

Depends what we want to do; if we do this things work in all versions, if not we need the shim.

ejgallego:
> Why not?

This is a problem that needs to be addressed, not workarounded I'm afraid.

SkySkimmer:
>This is a problem that needs to be addressed, not workarounded I'm afraid.

That's just rephrasing that you don't want `[@ocaml.warning "-3"]`, not an explanation of why.

ejgallego:
> That's just rephrasing that you don't want `[@ocaml.warning "-3"]`, not an explanation of why.

We add that, and then at some point when `Pervasives` is removed we have to fix it again. So that's a double fix == double work.

gares:
> Depends what we want to do; if we do this things work in all versions, if not we need the shim.

My remark was about the inconsistency, I'm happy to let you chose the transition strategy (that is one or the other, picking both looks like a mistake to me).

ejgallego:
> My remark was about the inconsistency, I'm happy to let you chose the transition strategy (that is one or the other, picking both looks like a mistake to me).

This inconsistency is already here, in the sense that many things from `Pervasives` are used unqualified in the codebase. I dunno what to do.

gramlib/grammar.ml
```diff
@@ -720,7 +720,7 @@ and print_level : type s. _ -> _ -> s ex_symbols list -> _ =
       (fun sep (ExS rule) ->
          fprintf ppf "%t%a" sep print_rule rule;
          fun ppf -> fprintf ppf "%a| " pp_print_space ())
-      (fun ppf -> ()) rules
+      (fun ppf -> ()) rules ppf
```
SkySkimmer: I guess 08 detects the partial application? Why did this even work before?

interp/dumpglob.ml
```diff
@@ -37,15 +37,15 @@ let dump_to_dotglob () = glob_output := MultFiles
...
+    (glob_output := StdOut; glob_file := Stdlib.stdout)
   else
     (glob_output := File f; open_glob_file f)
 
 let feedback_glob () = glob_output := Feedback
 
 let dump_string s =
   if dump () && !glob_output != Feedback then 
-    Pervasives.output_string !glob_file s
+    Stdlib.output_string !glob_file s
```
SkySkimmer: None of the qualifications in this file seem necessary.

kernel/indtypes.mli
```diff
@@ -33,4 +33,4 @@ type inductive_error =
 [@@ocaml.deprecated "Use [Type_errors.inductive_error]"]
 
 exception InductiveError of Type_errors.inductive_error
-[@@ocaml.deprecated "Use [Type_errors.InductiveError]"]
+(* [@@ocaml.deprecated "Use [Type_errors.InductiveError]"] *)
```
SkySkimmer: ??

SkySkimmer (commented):
It seems qualification by Pervasives can be classified into
- compare and `(=)` qualified to make it more obvious that we're using the polymorphic ones (most common)
- accessing locally overloaded stuff (AFAICT this is just the ref operations overloaded by nonlogical monad, which we can deal with by using `.contents` instead of `Pervasives.(!)` etc) (most rare)
- not actually useful qualification (should just be removed)

To deal with explicit compare/(=) maybe it would make sense to define aliases `polymorphic_compare`/`polymorphic_equal` in lib/util and use those everywhere else. The alias definition doesn't need to use qualification.

Then we wouldn't need to use the stdlib shim. 

Or we can go with the shim, I'm lessening my opposition as I think about it.

SkySkimmer (changes requested):
The shim check should not fail if ocaml is recent enough to not need it, and then make shouldn't pass -package stdlib-shims.

gramlib/grammar.ml
ejgallego: It did not.

interp/dumpglob.ml
ejgallego: I'm OK removing them; I just didn't dare to do it without consensus first.

kernel/indtypes.mli
ejgallego: This is actually an error now, the attribute was a no-op but < 4.08 didn't complain.

ejgallego:
> The shim check should not fail if ocaml is recent enough to not need it, and then make shouldn't pass -package stdlib-shims.

I am afraid I am not sure what this means, we could make it conditional on the version indeed but not sure it is worth it. We could as well drop all qualification and just drop the shim.

JasonGross:
> To deal with explicit compare/(=) maybe it would make sense to define aliases polymorphic_compare/polymorphic_equal in lib/util and use those everywhere else. The alias definition doesn't need to use qualification.
>
> Then we wouldn't need to use the stdlib shim.

I would prefer this solution (or really any solution that doesn't require depending on the shim).

kernel/indtypes.mli
SkySkimmer: >the attribute was a no-op but < 4.08 didn't complain.

I just tested and I do get a deprecated warning if I put `raise (Indtypes.InductiveError (assert false))` somewhere, and no warning if I remove the attribute. This is on 4.07.1.

SkySkimmer:
>I am afraid I am not sure what this means

In INSTALL you say "For OCaml versions older than 4.07.0, you will need the stdlib-shim compatiblity package installed" but configure/make rules require it regardless of ocaml version.
See also CI failures in the trunk jobs.

kernel/indtypes.mli
ejgallego: Maybe a 4.08 bug?

Zimmi48 (commented):
It would be good to make this backportable to 8.9. For this purpose, the shim solution looks too invasive. But the solution could be different for v8.9 and master of course.

ejgallego:
> It would be good to make this backportable to 8.9. For this purpose, the shim solution looks too invasive. But the solution could be different for v8.9 and master of course.

I guess that for non-dev releases you can build with `-warn-error no`, dune already does this automatically when using the release mode.

Zimmi48:
In any case, if this PR is not marked for backporting, we should make sure that an issue is opened about turning the fatal warnings off in v8.9 if this is the only thing preventing compatibility with 4.08.

----

BTW, this is a bit off-topic, but since it was discussed in this thread, concerning the bootstrapping debate:
I think that the PR was blocked by Enrico mostly because the use case was not super clear at that time and even if I personally didn't mind, I had a hard time defending it because I couldn't see clearly the use case either. Since then, you have shown repeatedly a number of use cases in practice. In particular, you are right that this is the way the Windows build works and it would be better if as much as possible in this process could be made generic. So in this context, I would be more supportive if you were to resubmit a proposal to support a bootstrapping target, especially if it can be as generic as possible and usable for Windows. I don't think anyone would block it again then...

ejgallego:
> To deal with explicit compare/(=) maybe it would make sense to define aliases `polymorphic_compare`/`polymorphic_equal` in lib/util and use those everywhere else. The alias definition doesn't need to use qualification.

I think it does make sense to follow this route and drop the shim. The gain is not so large, IMHO the shim makes more sense for people that want to use the new `Float` `UChar` modules, etc...

I'll add these aliases to `Util`

mattam82:
I also prefer the aliasing solution. What is the motivation of OCaml to move from Pervasives to Stdlib?

kernel/indtypes.mli
SkySkimmer: IIUC https://github.com/ocaml/ocaml/pull/1705/ changed @@ on exceptions to handle more stuff, and the correct way to deprecate an exception is just 1 @ (like with constructors). Just 1 @ works for me at least.
We can either go with 1 @ or drop the re-export entirely.

SkySkimmer:
With 4.08 released, what's the plan for this PR?

ejgallego:
> With 4.08 released, what's the plan for this PR?

I'm not sure what to do; OPAM packages should be OK as [at least for Dune] they don't build with fatal warnings on.

I dunno if we want to go the route of requiring the `stdlib-shim` packages, or just remove references to pervasives?

More interesting tricks can be done with module aliases [by merit of Coq defining a `Stdlib` module [this is done in Dune's codebase for example] but I'm not sure how far we want to go.

gasche:
> IMHO the shim makes more sense for people that want to use the new Float UChar modules, etc...

`stdlib-shims` doesn't help you with that; when you install it from OCaml version N, you get only the stdlib modules that were available at version N in your Stdlib. If you want to have the new modules, you should use @thierry-martinez's https://github.com/thierry-martinez/stdcompat instead.

> What is the motivation of OCaml to move from Pervasives to Stdlib?

(I don't know.)

XVilka:
Also note that stdcompat was recently released with a couple important fixes: https://github.com/ocaml/opam-repository/pull/14365

PR #9636, opened 2019-02-23
[dune] [refman] Generate coq-refman.install file.

ejgallego:
We generate a coq-refman install file manually as Dune still doesn't
have install rules for directories.

This should be solved upstream soon tho, but it is useful for some
cases now.

### DIFF
--- a/Makefile.dune
+++ b/Makefile.dune
@@ -87,7 +87,7 @@ test-suite: voboot
 	dune runtest --no-buffer $(DUNEOPT)
 
 refman-html: voboot
-	dune build @refman-html
+	dune build $(DUNEOPT) @refman-html
 
 stdlib-html: voboot
 	dune build @stdlib-html
--- a/tools/dune
+++ b/tools/dune
@@ -46,3 +46,7 @@
  (package coq)
  (modules coq_tex coq_dune)
  (libraries str))
+
+(executable
+ (name gen_install)
+ (modules gen_install))
--- a/doc/dune
+++ b/doc/dune
@@ -15,6 +15,19 @@
  (action
   (run env COQLIB=%{project_root} sphinx-build -j4 %{env:SPHINXWARNOPT=-W} -b html -d sphinx_build/doctrees sphinx sphinx_build/html)))
 
+(include refman-install.inc)
+
+(rule
+ (targets refman-install.gen)
+ (action
+  (with-stdout-to %{targets}
+    (run %{project_root}/tools/gen_install.exe coq-refman doc sphinx_build/html refman/html))))
+
+(alias
+ (name refman-html)
+ (action
+  (diff refman-install.inc refman-install.gen)))
+
 (alias
  (name refman-html)
  (deps sphinx_build))
--- a/doc/refman-install.inc
+++ b/doc/refman-install.inc
@@ -1,0 +1,9 @@
+(rule
+ (targets sphinx_build/html/index.html)
+ (deps sphinx_build)
+ (action (run true)))
+
+(install
+ (section doc)
+ (package coq-refman)
+ (files ("sphinx_build/html/index.html" as "refman/html/index.html")))
--- a/tools/coq_dune.ml
+++ b/tools/coq_dune.ml
@@ -57,11 +57,6 @@ module Aux = struct
       | x::xs, y::ys -> let r = f x y in if r = 0 then lc xs ys else r
     in lc
 
-  let rec pp_list pp sep fmt l = match l with
-    | []  -> ()
-    | [l] -> fprintf fmt "%a" pp l
-    | x::xs -> fprintf fmt "%a%a%a" pp x sep () (pp_list pp sep) xs
-
   let rec pmap f l = match l with
     | []  -> []
     | x :: xs ->
@@ -70,7 +65,7 @@ module Aux = struct
         | Some r -> r :: pmap f xs
       end
 
-  let sep fmt () = fprintf fmt "@;"
+  let pp_sep fmt () = fprintf fmt "@;"
 
   (* Creation of paths, aware of the platform separator. *)
   let bpath l = String.concat Filename.dir_sep l
@@ -154,13 +149,13 @@ let gen_sub n =
 
 let pp_rule fmt targets deps action =
   (* Special printing of the first rule *)
-  let ppl = pp_list pp_print_string sep in
+  let ppl = pp_print_list ~pp_sep pp_print_string in
   let pp_deps fmt l = match l with
     | [] ->
       ()
     | x :: xs ->
-      fprintf fmt "(:pp-file %s)%a" x sep ();
-      pp_list pp_print_string sep fmt xs
+      fprintf fmt "(:pp-file %s)%a" x pp_sep ();
+      pp_print_list ~pp_sep pp_print_string fmt xs
   in
   fprintf fmt
     "@[(rule@\n @[(targets @[%a@])@\n(deps @[%a@])@\n(action @[%a@])@])@]@\n"
@@ -204,7 +199,7 @@ let out_install fmt dir ff =
   let ff = List.concat @@ pmap (function | VO vo -> Some (gen_coqc_targets vo) | _ -> None) ff in
   let pp_ispec fmt tg = fprintf fmt "(%s as coq/%s)" tg (bpath [itarget;tg]) in
   fprintf fmt "(install@\n @[(section lib_root)@\n(package coq)@\n(files @[%a@])@])@\n"
-    (pp_list pp_ispec sep) ff
+    (pp_print_list ~pp_sep pp_ispec) ff
 
 (* For each directory, we must record two things, the build rules and
    the install specification. *)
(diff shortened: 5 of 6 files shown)


ejgallego:
Note that this won't work with `dune install` as it will rebuild the `.install` file. However it does work with OPAM for example.

We could remove the `coq-refman` package temporarily to avoid this problem.

rgrinberg:
Is there a dune ticket that describes the issue? I'm not aware of one. An issue would certainly help solve this issue sooner rather than later.

ejgallego:
> Is there a dune ticket that describes the issue? I'm not aware of one. An issue would certainly help solve this issue sooner rather than later.

These ones are relevant:
- https://github.com/ocaml/dune/issues/1377
- https://github.com/ocaml/dune/pull/1547#issuecomment-441650546

Indeed Dune cannot properly install directory targets yet, do you think an extra issue should be opened? I was planning to give it a go once I finish the rewrite of the Coq PR, but of course it is great if someone wants to go ahead.

rgrinberg:
The issues you listed are related to providing first class support for directories as dependencies and targets. Actually, making directories installable has its own set of issues so I think it deserves a ticket.

The main issue is that if a rule produces a whole directory, then dune can no longer know if a path inside the directory is a target of the rule statically. I suppose this limitation is acceptable, but it might be painful to actually implement. So we should discuss it.

By the way, can you confirm that opam's install files support listing directories?

ejgallego:
> By the way, can you confirm that opam's install files support listing directories?

I think that's the main issue, last time I tried the `.install` format didn't support directories, so indeed the simplest install strategy that seems to work is to list all the files recursively; that's what this hack does.

Another possibility would be for `.install` to support directories, I am not sure how hard is that, and would indeed require a newer opam.

> The main issue is that if a rule produces a whole directory, then dune can no longer know if a path inside the directory is a target of the rule statically.

Any file inside it should be a target, right? Cannot this be a simple file containment check? [I am likely missing something]

> Actually, making directories installable has its own set of issues so I think it deserves a ticket.

Sounds good so let's open a ticket then.

rgrinberg:
> Another possibility would be for .install to support directories, I am not sure how hard is that, and would indeed require a newer opam.

Yeah, it's not a big deal. Just an implementation detail.


> Any file inside it should be a target, right? Cannot this be a simple file containment check? [I am likely missing something]

The problem is that dune currently knows if a file is target without executing any rules. The simple file containment check would have to be done after execution the rule.

tools/dune
```diff
@@ -41,8 +41,8 @@
 (ocamllex coqwc)
 
 (executables
- (names coq_tex coq_dune)
- (public_names coq-tex coq_dune)
+ (names coq_tex coq_dune gen_install)
+ (public_names coq-tex coq_dune gen_install)
  (package coq)
```
Zimmi48: Shouldn't this particular executable be part of the `coq-refman` package instead? And don't you want to put it in `doc/tools`, or even `doc/tools/coqrst` or did you put it here and in the main package because you expected that it would be useful for other applications?
Another question: shouldn't this executable be private? IIUC the default is to install it and it doesn't really make sense to me that it be the case.

Makefile.dune
```diff
@@ -78,8 +78,11 @@ quickopt: voboot
 test-suite: voboot
 	dune runtest --no-buffer $(DUNEOPT)
 
-refman-html: voboot
+coq-refman.install: voboot
 	dune build @refman-html
+	$(BUILD_CONTEXT)/tools/gen_install.exe $(BUILD_CONTEXT)/doc/sphinx_build/html doc $@
```
Zimmi48: I am bad at Makefiles, but don't you need to explicitly tell the Makefile that you need to build the tool first? Or is it because the refman target depends on the full coq package anyway?

tools/dune
ejgallego: Well spotted, that should be 100% private; sorry for the bug, fixing.

Makefile.dune
ejgallego: Indeed you are right; this solution is a bit messy as you can see :/

Makefile.dune
```diff
@@ -80,8 +80,12 @@ quickopt: voboot
...
 	dune runtest --no-buffer $(DUNEOPT)
 
-refman-html: voboot
-	dune build @refman-html
+coq-refman.install: voboot
+	dune build $(DUNEOPT) @refman-html
+	dune build $(DUNEOPT) tools/gen_install.exe
+	$(BUILD_CONTEXT)/tools/gen_install.exe $(BUILD_CONTEXT)/doc/sphinx_build/html doc $@
+
+refman-html: coq-refman.install
```
Zimmi48: I'm unsure of the benefit having a new target in `Makefile.dune` brings. You could copy-paste the content of this target here, couldn't you?

Zimmi48 (approved):
Thanks a lot! Is the stdlib doc already installed? If not, can you use the same trick to have it installed too? Then, we should finally be able to switch the Nix package to Dune.

Don't you want to remove the following lines now?

https://github.com/coq/coq/blob/8b42c73a6a3b417e848952e7510e27d74e6e1758/doc/dune#L22-L26

Zimmi48:
> Is the stdlib doc already installed? If not, can you use the same trick to have it installed too?

Once this is done, we could make the deploy job depend on the `_build/install/default/doc` files, rather than the `_build/default/doc` files...

Makefile.dune
ejgallego: The rationale is that we are overriding the default `coq-refman.install` target which happens to be a true file, so IMHO it is better to declare and have the file properly rebuilt, we could even add the right dependencies using a glob.

Anyways it seems that proper support in Dune is coming in 1.8 [pretty soon] so hopefully this hack will be very short-lived.

ejgallego:
> Once this is done, we could make the deploy job depend on the `_build/install/default/doc` files, rather than the `_build/default/doc` files...

Yes, tho there is a small problem here, we would need to introduce a `coq-doc-stdlib` doc package in order to this with the current setup. Dune has I think the same problem with the `odoc` target, docs cannot be installed.

In my view the stdlib docs should live in the main `coq` package, and then moved to `coq-stdlib` once that's in its own package, so I dunno if it is worth doing the hack now.

Zimmi48:
I can't see the problem. Why can't it be installed as part of the `coq` package until we introduce the `coq-stdlib` package?

Zimmi48:
Oh I get it! This is because you need a single `.install` file. Well, then the `coq-stdlib-doc` package option seems the way to go indeed...

ejgallego:
> Oh I get it! This is because you need a single `.install` file. Well, then the `coq-stdlib-doc` package option seems the way to go indeed...

We can also append to the `coq.install` file; just another variation of the hack.

Zimmi48:
We can? It was what I was going to propose at first. I would prefer this variation of the hack indeed.

ejgallego:
Done, the hack is not very nice but in some sense not so different from what the makefiles do today.

Zimmi48 (approved):
Wonderful! Don't you want to update the deploy job to use the install directories? This won't change much but it could allow us to detect early if we were to introduce problems in the future...

ejgallego:
Yup seems like a good idea.

ejgallego:
Not sure tho how we should install things and which job; will take me a while to figure.

Zimmi48:
I meant something simpler: just copy the files in `_build/install/default/doc`, rather than those in `_build/default/doc`... Obviously, running `opam-install` is possible too.

ejgallego:
> I meant something simpler: just copy the files in `_build/install/default/doc`, rather than those in `_build/default/doc`... Obviously, running `opam-install` is possible too.

Oh indeed note that we didn't place files in `_build/install`, the `.install` file instead refers the files in the build dir.

Maybe we should generate proper install rules, and store them in a promoted file, so we avoid these problems.

ejgallego:
Only reason I avoided this is as it'd require a person adding/remove a doc file to run the promote target.

Zimmi48:
That wouldn't be a big deal I think, especially given that as of today you already have to update manually:
- `doc/sphinx/index.*.rst` when you add a new file to the reference manual.
- `index-list.html.template` when you add a new file to the stdlib.

ejgallego:
Ok lets go for the promoted way which is way less hackish.

ejgallego:
After some more thinking, I think all the approaches I've tried are fatally flawed so no choice here but to wait for upstream Dune support.

Zimmi48:
@ejgallego I don't understand at all. I was about to merge this during the weekend. I just made a minor improvement suggestion, that led you to propose a new approach and then to close this PR :'( But I'm fine with going back to the previous hack of appending to the install file. I honestly don't think it is so bad.

ejgallego:
> But I'm fine with going back to the previous hack of appending to the install file. I honestly don't think it is so bad.

It is worse than it looks, for example it didn't work with `dune install` which may be used by Nix; so in the end, this was too hacky as not to create problems. I'd rather put the manpower on fixing this properly.

PR #9638, opened 2019-02-23
[save_proof] Have admitted drop the current proof.

ejgallego:
It was strange that before we didn't drop the proof. Indeed, this is a
bugfix for clients of the `Lemmas` API.

Maybe this could help remove some `discard_all` from the STM, but I am
not sure.

Submitting as a draft PR to try out the proposed policy given that this depends on #9566


ejgallego:
After #10050 and the new version of #9566 this bug doesn't exist anymore.

PR #9641, opened 2019-02-24
[proofs] Store hooks in the proof object.

ejgallego:
As of now, hooks were stored in the terminators as closures, we place
them instead in the proof object and are thus passed back at proof
closing time.

This helps towards the reification and unification of terminators.


ejgallego:
I am consolidating this PRs in #9645 as indeed it makes sense now [the original PR was submitted before the consolidating PR existed]

PR #9642, opened 2019-02-24
[proof] Unify obligation proof save path: Part I, declareObl

ejgallego:
We move obligation declaration-specific functions to their own
file. This way, `Lemmas` can access them, and in the next part we can
factorize common parts in the save proof.


ejgallego:
I am consolidating this PRs in #9645 as indeed it makes sense now [the original PR was submitted before the consolidating PR existed]

PR #9645, opened 2019-02-24
[proof] Remove terminator type, unifying regular and obligation ones.

ejgallego:
We radically redesign how proof closing information is stored. Instead
of a user-defined closure, we now reify control into a single data
structure containing the needed information.

In this scheme, the `Lemmas` module can get extra information with
obligation info when opening the proof, and will correspondingly call
the right closing function based on this. Another important advantage
is that essential proof data is not stored in closures anymore, which allow
clients to correctly thread / support it.

Note that some more essential data is still stored in a closure but #10316 should fix it.

This PR is the start of what will hopefully be a much bigger unification of all the existing proof saving paths.

Overlays:
- https://github.com/mattam82/Coq-Equations/pull/215
- https://github.com/coq-community/paramcoq/pull/33
- https://github.com/Mtac2/Mtac2/pull/213


ejgallego:
Ok, this PR is becoming ready except for some strange stuff in Equations; I'm not sure what the problem is I'll research more.

Note that now clients of the `vernac` layer [such as the STM] may test against the proof information and see whether it is safe to delegate or not. In the current setup, it is only safe when `proof_ending = Regular` and `hook = None`, I think.

ejgallego:
Finally ready for review and hopefully merge! Needs assignee, anyone interested on proof life cycle would be a good candidate.

ejgallego:
Review hints:
- the important changes are confined to `lemmas.ml`
- there is some refactoring due to control inversion no more, this for the most part is very boring [see individual commits]. Main changes are indeed w.r.t. hooks and obligations:
  + for hooks, now they are called directly by `DeclareDef`, so they have to live here, before they were called by hand from the terminator function [thus from lemmas]
  + for obligations, that's the same, the closing function was captured in the terminator, and thus above `Lemmas`, now it needs to live below it so `DeclareObl` was introduced.

aspiwack:
Just so I understand, the derive plugin was super simplified, because its logic was moved to `lemma.ml`. Am I correct? I just had a very quick look, for the moment.

ejgallego:
> Just so I understand, the derive plugin was super simplified, because its logic was moved to `lemma.ml`. Am I correct? I just had a very quick look, for the moment.

Indeed @aspiwack , clients have a hard time using the proof API as terminators can do arbitrary actions including saving constants [so for example the STM is in continuous tension with what plugins do] For example the STM cannot work with Derive as things are now.

So what we aim with this PR is to gather all proof saving paths in a direct style, and hopefully we can unify them later on. Already big chunks of `declareDef / lemmas` can be done, as well as obligations; then at a later step the multi proof API [derive and equations] should be handled too, but here we need to explore the design space more as hinted by @mattam82 in the last WG.

ejgallego:
@ppedrot ping, would you mind taking this PR?

ppedrot (commented):
The code changes look good to me, and even though the current implementation is a mess beyond comprehension, I think that I understood the main changes, and I am mostly OK with them.

That said, while I agree that making explicit closed behaviour assumptions and reducing inversion of control is desirable in the long run, it seems that the current PR is reducing the extensibility of the Coq declaration mechanism. This is pretty clearly witnessed by the fact that you had to add a dedicated constructor for the Equations code path.

So, my approval of the current PR boils down to the following question: should we allow external devs to write clever (read: *crazy*) plugins that can arbitrarily mess with proof outputs in order to design unexpected features, or should we instead restrict them to a few safe, finely hand-crafted terminators?

(And also, I think you need to be aware of the Castillan version of that one movie to understand the joke culturally veiled in the name of this branch.)

ejgallego:
Thanks for the comments @ppedrot ; my point of view is that something such as the proof saving path should be flexible enough so plugin writers don't have the need to override it. Indeed, for now I have just copied the existing code verbatim but the plan would be to unify all the current 3 ones.

I could even argue that in fact plugin writers doing this is not so good, as they will indeed find their own workarounds to Coq API defects [such as the `ref` in the equation termination data] instead of proposing improvements.

Note this PR solves a couple of critical problems:
- with terminators, important data for proofs is stored inside a closure, thus proofs cannot be delegated to workers and indeed stuff such as equations is buggy w.r.t. async processing [tho it tends to be very conservative w.r.t. "opacity data" but however that's an approximation whereas storage of data in direct form allows to analyze it.
- second, all document managers need to make assumptions on the closing proof behavior if they want to delay checking / allow incremental compilation. With the current terminators the upper layers just have no idea about what a proof will do.

So indeed I think that restricting behavior so plugin authors have to collaborate with Coq upstream if they want to do a fancy feature should be OK. Note that they still have a `hook` member that can do lots of extra stuff.

ejgallego:
> (And also, I think you need to be aware of the Castillan version of that one movie to understand the joke culturally veiled in the name of this branch.)

Indeed, Constantino Romero has been the best terminator ever :)

ejgallego:
> the plan would be to unify the 3 proof closing paths into one with a richer API.

Indeed I do believe that for example, the proof_ending for equations / derive should be the same, and correspond to the regular `start_dependent_proof` closing path.

Integration of obligations would take a bit more time, but I do believe that it should be possible. As of today it is not using the start_dependent_proof because of API problems but that's fixable too, and the API can allow indeed two refinement styles.

I forgot to mention that also a big motivation for this PR is to remove loads of duplicate code [for example in declaredef / lemmas / obligations / equations / etc...] while this PR doesn't tackle it [as I preserve code for now], it is a strong requirement in order to try to factorize more, for example the `proof+sayonara_baby+unify_save` branch does merge the common parts of `Lemmas` / `DeclareDef`.

ppedrot (approved):

vernac/declareDef.ml
```diff
@@ -14,6 +14,29 @@ open Entries
...
+      -> Names.GlobRef.t
+      -> unit
+  end
+
+  type t = S.t CEphemeron.key
+
+  let make hook = CEphemeron.create hook
+
+  let call ?hook ?fix_exn uctx trans l c =
+    try Option.iter (fun hook -> CEphemeron.get hook uctx trans l c) hook
```
gares: I'd like to have a specific error if the hook is not there (eg the ephemeron was sent over a wire)

gares: Maybe I did not get what this PR does wrt the ephemeron stuff but I guess there are 3 cases:
- no hook is set
- hook is set but CEphemeron.get returns None
- hook is set and we get Some
Are the first two paths mixed? I believe that the second should be detected and properly reported.

ejgallego: In the second case indeed the CEphemeron will raise `Invalid_key`, this is the standard reporting mechanism for "ephemerons", how would you like to be reported?

Note that this works with current clients as the new behavior is strictly a superset of the old one, before this PR you always get an invalid client, now you get it only if the hook was set.

ejgallego: The API is a bit more convenient like this, but the CEphemeron could be indeed placed in the proof info field.

ejgallego: > The API is a bit more convenient like this, but the CEphemeron could be indeed placed in the proof info field.

So the argument in favor of doing that is that now `DeclareDef` doesn't need to know about the Ephemeron, the argument against is that clients do need to be aware of the internal representation of the hook. I dunno.

ejgallego: Note @gares  that indeed after this and #10316 clients would have better options w.r.t. analyzing what kind of proof they have on hand [for example the `respects opacity` thing is not longer needed], however so far the PRs keep proof information private on purpose, analysis functions should be added as clients need I guess. In any case decisions such as the exact location of Ephemerons should evolve with use cases and IMHO are not a big deal in the sense that types keep us safe when doing such refactorings.

gares: I was suggesting to catch the exception here and give an ad-hoc anomaly mentioning "proof terminator" (and possibly the constant name).

ejgallego: Actually this doesn't to have to be fatal, clients may want to catch `InvalidKey` and do something else [like re-schedule the proof?]

gares: I don't think it can happen: declaredef is called *always* from the very same actor that started the definition (eg that passed the hook in the first place).

The hack in the STM that was putting hooks back was there because hooks were part of the proof state that is sent back and forth, and hence pruned wrt hooks. This should have gone now.

There was also another "exception" that I don't know if it is still there. The worker would "run Qed" not really to update the environment but to check the proof term is OK (free from tactic errors) before sending it back. In that case a dummy hook could be used (I don't recall how this is/was done). A more polite thing would be to call the kernel, but IIRC the API were not easy to call, so I just did as if Qed was run.

ejgallego: > I don't think it can happen: declaredef is called _always_ from the very same actor that started the definition (eg that passed the hook in the first place).

I guess it'd depend what the client does with the proof information.

But indeed the STM should have a better time [specially after #10316 ] but I am not so familiar with that it does [at least now it does indeed not preserve the lemma info, but after #10316 as you say the same actor could keep a hold on it.]

ejgallego: The Ephemeron is necessary yet by the way as the STM will try to marshall the full state for example on `par:` .

SkySkimmer:
>second, all document managers need to make assumptions on the closing proof behavior if they want to delay checking / allow incremental compilation. With the current terminators the upper layers just have no idea about what a proof will do. So flexibility is very low and the STM for example needs to be provided of extra meta-data which is often out of sync and which this PR makes redundant.

I think we could marshal by using some indirection but this would remain so I'm OK with the PR. Looking forward to code unifications.

ejgallego:
@ppedrot IMHO we can merge, this has been open in the same exact shape for months so IMVHO people who had interest on this part of the system did have time to comment.

ppedrot:
Waiting for an ACK from @gares, and optionally @maximedenes. I'll merge tomorrow if no answers by then.

ejgallego:
> Waiting for an ACK from @gares, and optionally @maximedenes. I'll merge tomorrow if no answers by then.

Sure, no particular rush; if you want some more extra time to look at this folks please say so, but I'd love to have some sort of rough feedback date so I can plan instead of having to check the status of this every day.

SkySkimmer:
>@ppedrot IMHO we can merge, this has been open in the same exact shape for months so IMVHO people who had interest on this part of the system did have time to comment.

Counting the time spent with needs:merge dependency is dubious IMO

gares:
looks ok

ejgallego:
Thanks @ppedrot , "no problemo"

PR #9646, opened 2019-02-24
[proof] Fix proof bullet error helper which was implemented as a hook

ejgallego:
We add the information on the proper layer by catching the low-level
exception.

This needs testing and review, likely we are not hooking into the right place.

### DIFF
--- a/engine/proofview.ml
+++ b/engine/proofview.ml
@@ -352,11 +352,12 @@ let _ = CErrors.register_handler begin function
   | _ -> raise CErrors.Unhandled
 end
 
-(** [tclFOCUS_gen nosuchgoal i j t] applies [t] in a context where
+(** [tclFOCUS ?nosuchgoal i j t] applies [t] in a context where
     only the goals numbered [i] to [j] are focused (the rest of the goals
     is restored at the end of the tactic). If the range [i]-[j] is not
     valid, then it [tclFOCUS_gen nosuchgoal i j t] is [nosuchgoal]. *)
-let tclFOCUS_gen nosuchgoal i j t =
+let tclFOCUS ?nosuchgoal i j t =
+  let nosuchgoal = Option.default (tclZERO (NoSuchGoals (j+1-i))) nosuchgoal in
   let open Proof in
   Pv.get >>= fun initial ->
   try
@@ -367,10 +368,9 @@ let tclFOCUS_gen nosuchgoal i j t =
     return result
   with CList.IndexOutOfRange -> nosuchgoal
 
-let tclFOCUS i j t = tclFOCUS_gen (tclZERO (NoSuchGoals (j+1-i))) i j t
-let tclTRYFOCUS i j t = tclFOCUS_gen (tclUNIT ()) i j t
+let tclTRYFOCUS i j t = tclFOCUS ~nosuchgoal:(tclUNIT ()) i j t
 
-let tclFOCUSLIST l t =
+let tclFOCUSLIST ?(nosuchgoal=tclZERO (NoSuchGoals 0)) l t =
   let open Proof in
   Comb.get >>= fun comb ->
     let n = CList.length comb in
@@ -384,7 +384,7 @@ let tclFOCUSLIST l t =
     in
     let l = CList.map_filter sanitize l in
     match l with
-      | [] -> tclZERO (NoSuchGoals 0)
+      | [] -> nosuchgoal
       | (mi, _) :: _ ->
           (* Get the left-most goal to focus. This goal won't move, and we
              will then place all the other goals to focus to the right. *)
@@ -401,7 +401,7 @@ let tclFOCUSLIST l t =
 
 
 (** Like {!tclFOCUS} but selects a single goal by name. *)
-let tclFOCUSID id t =
+let tclFOCUSID ?(nosuchgoal=tclZERO (NoSuchGoals 1)) id t =
   let open Proof in
   Pv.get >>= fun initial ->
   try
@@ -421,7 +421,7 @@ let tclFOCUSID id t =
         t >>= fun result ->
       Comb.set initial.comb  >>
         return result
-  with Not_found -> tclZERO (NoSuchGoals 1)
+  with Not_found -> nosuchgoal
 
 (** {7 Dispatching on goals} *)
 
--- a/engine/proofview.ml
+++ b/engine/proofview.ml
@@ -345,22 +345,11 @@ let tclBREAK = Proof.break
 
 exception NoSuchGoals of int
 
-(* This hook returns a string to be appended to the usual message.
-   Primarily used to add a suggestion about the right bullet to use to
-   focus the next goal, if applicable. *)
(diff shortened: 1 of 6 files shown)


maximedenes:
@Zimmi48 it would be nice if you could have a look at the fix suggested here, to understand if the risk of merging #9129 without this is low enough.

@ejgallego can you say what refinements you have in mind? What do all these CI errors mean? Is it just because the overlays need to be rebased?

maximedenes:
@Zimmi48 ideally, if you could provide some tests for this feature, it would be good to exercise this PR against them.

ejgallego:
> can you say what refinements you have in mind?

More thinking is needed about what is the right proof entry point for this feature to hook.

> What do all these CI errors mean? Is it just because the overlays need to be rebased?

Yes.

Zimmi48:
I can try to do so, but unfortunately I have very little time ATM so this should not delay #9129.

ejgallego:
Yeah, I'd say not to worry too much about this, it is after all just a suggestion for the user; we'll get that fixed or even improved easily.

Zimmi48:
So I did test the PR and it does not work:

```
Goal True /\ True
split.
- easy.
  idtac.
```

gives `Error: No such goal.`, whereas in Coq 8.9, it gave `Error: No such goal. Focus next goal with bullet -.`

ejgallego:
> I did test the PR and it does not work:

Yup, what is the right entry point to add the info then?

engine/proofview.ml
```diff
@@ -345,22 +345,12 @@ let tclBREAK = Proof.break
...
 
-(* This hook returns a string to be appended to the usual message.
-   Primarily used to add a suggestion about the right bullet to use to
-   focus the next goal, if applicable. *)
-let nosuchgoals_hook:(int -> Pp.t) ref = ref (fun n -> mt ())
-let set_nosuchgoals_hook f = nosuchgoals_hook := f
-
-
-
 (* This uses the hook above *)
```
Zimmi48: This comment is outdated and should be removed.

engine/proofview.ml
```diff
@@ -345,22 +345,12 @@ let tclBREAK = Proof.break
...
 let _ = CErrors.register_handler begin function
   | NoSuchGoals n ->
-    let suffix = !nosuchgoals_hook n in
-    CErrors.user_err 
-      (str "No such " ++ str (String.plural n "goal") ++ str "." ++
-       pr_non_empty_arg (fun x -> x) suffix)
-      | _ -> raise CErrors.Unhandled
+    CErrors.user_err
+      (str "No such " ++ str (String.plural n "goal") ++ str ".")
```
Zimmi48: But more generally, it seems wrong to have an handler for `NoSuchGoals` if the goal is to always catch this exception and re-raise it as `SuggestNoSuchGoal`.

Zimmi48: Hum, OK I see. There are cases where the exception is not supposed to be caught and this corresponds to previous cases where the hook was put back to the default value. Not the easiest code to read indeed :laughing:

engine/proofview.ml
ejgallego: Fixed thanks.

engine/proofview.ml
ejgallego: Indeed these codepaths are a bit tricky, they are getting a bit better after #9566 I hope.

Zimmi48:
I've spent more time looking into this than I actually had. It still doesn't work with the version where you also catch `NoSuchGoals` exceptions in `run_tactic`. I have no idea where the right entry point is, but I am still doubtful of the interest of registering an error handler for `Proofview.NoSuchGoals`.

ejgallego:
I'll look into this more calmly.

maximedenes:
So what should do about this? If we don't finish this fix, it means #9129 will have introduced a regression...

SkySkimmer:
>I am still doubtful of the interest of registering an error handler for Proofview.NoSuchGoals.

IMO it makes sense in this kind of scenario:
~~~coq
Goal True.
  - only 2: exact I.
~~~
In general the `only 2` would be in the middle of some tactic chain and we don't care what the next bullet is.
Without the `only` it might also make sense, not sure.

Zimmi48:
> If we don't finish this fix, it means #9129 will have introduced a regression...

The regression might be acceptable (up to the RM) but the documentation is currently incorrect because the error message is said to contain a hint on the next bullet, so at least this should be fixed.

SkySkimmer:
Pushed a fixed version (also rebased). AFAICT the change in proof_global wasn't useful. See commit for details.

ejgallego:
I am not sure about the fix here, indeed maybe it is better than the original one?

I had a fix done on top of the February 12th submission but as they didn't advance so didn't the fix, but indeed was based on enriching the error.

Zimmi48:
At least, it works with the test I provided while the original one didn't.

ejgallego:
> At least, it works with the test I provided while the original one didn't.

The updated version of the fix I had did work, but IMHO this one seems equal or better so if you `engine/proof` guys are OK with it IMO we should merge @SkySkimmer 's fix.

Zimmi48:
Ah OK, I had missed it then.

proofs/pfedit.ml
```diff
@@ -73,6 +73,7 @@ let solve ?with_end_tac gi info_lvl tac pr =
       | None -> tac
       | Some _ -> Proofview.Trace.record_info_trace tac
     in
+    let nosuchgoal = Proofview.tclZERO (Proof_bullet.SuggestNoSuchGoals (1,pr)) in
```
Zimmi48: I don't get the point of this first argument if it is always set to 1.

Zimmi48 (approved):
Barring my small remark I like very much this solution.

ejgallego:
> Ah OK, I had missed it then.

I didn't submit it as it depended on a lot of PRs, it did not plan well I guess.

ejgallego:
> Barring my small remark I like very much this solution.

I am not sure tho about using an optional argument here, I guess it may be worth making it explicit so we don't miss important cases.

SkySkimmer:
>I am not sure tho about using an optional argument here, I guess it may be worth making it explicit so we don't miss important cases.

I'm pretty sure the only case where we want a bullet suggestion is the one I changed.

proofs/pfedit.ml
SkySkimmer: >I don't get the point of this first argument if it is always set to 1.

I was incompletely following the previous system.
I'll take some time to think about it.

ejgallego:
In case it was not clear [as I cannot mark the PR green] this is ready for merge @maximedenes

maximedenes:
@ejgallego Thanks for letting me know, it wasn't clear to me indeed.

ejgallego:
> @ejgallego Thanks for letting me know, it wasn't clear to me indeed.

Yeah that's the pitfall of reviews-on-your-own PRs.

Zimmi48:
This leaves my small comment for later consideration then.

PR #9647, opened 2019-02-25
[default.nix] Enable parallel build

vbgl:
This should reduce the total build time.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -84,6 +84,8 @@ stdenv.mkDerivation rec {
 
   prefixKey = "-prefix ";
 
+  enableParallelBuilding = true;
+
   buildFlags = [ "world" "byte" ] ++ optional buildDoc "doc-html";
 
   installTargets =
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -301,6 +301,7 @@ pkg:opam:
   dependencies: [] # We don't need to download build artifacts
   before_script: [] # We don't want to use the shared 'before_script'
   script:
+    - cat /proc/{cpu,mem}info || true
     # Use current worktree as tmpdir to allow exporting artifacts in case of failure
     - export TMPDIR=$PWD
     # We build an expression rather than a direct URL to not be dependent on
--- a/dev/ci/nix/coq.nix
+++ b/dev/ci/nix/coq.nix
@@ -5,5 +5,4 @@ let coq = callPackage wd { buildDoc = false; doInstallCheck = false; coq-version
 coq.overrideAttrs (o: {
   name = "coq-local-${branch}";
   src = fetchGit "${wd}";
-  enableParallelBuilding = true;
 })


ejgallego:
Hard to compare if build times are affected as the nix job doesn't print the needed worker information.

vbgl:
What do you mean? Some hint would be welcome on how to address this issue.

ejgallego:
> What do you mean? Some hint would be welcome on how to address this issue.

Oh sorry, workers have different numbers of processors and memory, what we do in the standard gitlab job is:
```
  - cat /proc/{cpu,mem}info || true
```
you need that in order to understand build times with -j

default.nix
```diff
@@ -80,10 +80,13 @@ stdenv.mkDerivation rec {
   preConfigure = ''
     patchShebangs kernel/
     patchShebangs dev/tools/
+    cat /proc/{cpu,mem}info || true
```
Zimmi48: I don't like having this here, it isn't related to the actual build. Please move it to `.gitlab-ci.yml`

Zimmi48 (commented):
That's not the first time I'm explaining this but the reason for not enabling parallel building was because we build both `byte` and `world` targets and this could lead to races. The plan was to fix this by moving to Dune. With ongoing work by @ejgallego to build and install the HTML doc with dune, that should be possible soon.

vbgl:
I don’t remember ever observing such a race. If that’s an issue, then this should not be merged. No problem.

ejgallego:
> I don’t remember ever observing such a race. If that’s an issue, then this should not be merged. No problem.

Indeed there is a problem in the makefile so you need to call make twice.

ejgallego:
I tried like 5 times to get a worker with 2 processors, no luck today... So I guess I leave this up to the Nix maintainers.

ejgallego (approved):

Zimmi48:
We'll soon be able to switch the build to Dune anyways.

ejgallego:
We finally got a builder with more than one core https://gitlab.com/coq/coq/-/jobs/179101249 , it seems to sped the build fine, merging.

ejgallego:
Well, @Zimmi48 , I see you didn't green tick this, so I won't assign / merge until I get your approval.

Zimmi48:
> We finally got a builder with more than one core https://gitlab.com/coq/coq/-/jobs/179101249 , it seems to sped the build up fine, merging.

We would need a few more similar runs to be sure there is no race. I prefer the solution of #9689 anyway.

ejgallego:
> We would need a few more similar runs to be sure there is no race. I prefer the solution of #9689 anyway.

Ok, then let's close this I guess.

vbgl:
If #9689 is for 8.11, then maybe this one is good for 8.10…

Zimmi48 (approved):
OK well, we can still merge and see if problems start occurring. If it's the case, it will be easy to revert.

Zimmi48:
The Nix failure is due to an unreliable test in the test-suite that has been removed since then.

PR #9648, opened 2019-02-25
[Gitlab-CI] Deploy api documentation to GH-Pages

vbgl:
In the same way as we currently deploy the user manual and the stdlib documentation, this also deploys the OCaml API documentation.

Test result: https://coq.github.io/doc/deploy-ml-api/api/

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -355,25 +355,29 @@ doc:refman:deploy:
     variables:
       - $DOCUMENTATION_DEPLOY_KEY
   dependencies:
+    - doc:ml-api:odoc
+    - doc:refman:dune
     - doc:refman
   before_script:
     - which ssh-agent || ( apt-get update -y && apt-get install openssh-client -y )
     - eval $(ssh-agent -s)
-    - echo "$DOCUMENTATION_DEPLOY_KEY" | tr -d '\r' | ssh-add - > /dev/null
     - mkdir -p ~/.ssh
     - chmod 700 ~/.ssh
     - ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
     - git config --global user.name "coqbot"
     - git config --global user.email "coqbot@users.noreply.github.com"
   script:
+    - echo "$DOCUMENTATION_DEPLOY_KEY" | tr -d '\r' | ssh-add - > /dev/null
     - git clone git@github.com:coq/doc.git _deploy
+    - rm -rf _deploy/$CI_COMMIT_REF_NAME/api
     - rm -rf _deploy/$CI_COMMIT_REF_NAME/refman
     - rm -rf _deploy/$CI_COMMIT_REF_NAME/stdlib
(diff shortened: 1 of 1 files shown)


ejgallego:
Cool, you should sync with @maximedenes so that https://coq.inria.fr/api/master/ points to here or it is removed.

.gitlab-ci.yml
```diff
@@ -355,25 +355,29 @@ doc:refman:deploy:
     variables:
       - $DOCUMENTATION_DEPLOY_KEY
   dependencies:
+    - doc:ml-api:odoc
+    - doc:refman:dune
     - doc:refman
```
ejgallego: That seems suspicious, why the double dependency on the `refman`?

ejgallego: Is that because of the lack of `coqdoc` build for the stdlib? If so please open a tracker issue so we are reminded to do so and update the deploy rule.

The plan for coqdoc is to have their own built-in rules for libraries so that requires the native Dune support being worked on, but in fact we may use coq2html instead.

vbgl: You are right: the refman is taken from the dune build; the stdlib from the non-dune build.

When the dune build-system will be able to generate the stdlib documentation, I guess that the `doc:refman` job will be removed. That will already mandate to update this job. Do we need an additional issue?

ejgallego: I have opened #9649 so I think that'll work.

Zimmi48:
> Cool, you should sync with @maximedenes so that https://coq.inria.fr/api/master/ points to here or it is removed.

I think it can just be removed. It hadn't been documented anywhere yet, and I much prefer the new hierarchy `doc/master/api`.

ejgallego (approved):
Looks ready to me, merging later today.

.gitlab-ci.yml
SkySkimmer: >You are right: the refman is taken from the dune build; the stdlib from the non-dune build.

Stdlib is from build:base though. Or you could go with just the non-dune built refman no?

ejgallego:
@maximedenes can you take care of removing that URL?

Zimmi48:
> @maximedenes can you take care of removing that URL?

@vbgl It could make sense for you to do it as well if you know what needs to be done... In any case, the sooner it goes, and we show people the new URL, the better it is as it will avoid people starting to rely on it (even if undocumented yet).

vbgl:
Sorry. I don’t have this power. Maxime wil take care of this.

maximedenes:
Indeed, this is a big problem of our current infrastructure. We can easily give access to Vincent... We hope to improve that very soon.

Zimmi48:
We can or we can't? Aren't you both back within Inria? Or by easy do you mean, without having to send a support request to Inria's IT team?

maximedenes:
> We can or we can't? Aren't you both back within Inria? Or by easy do you mean, without having to send a support request to Inria's IT team?

Inria has separate SSH networks for its centers, so that's not so simple.

PR #9650, opened 2019-02-25
[proof] Allow only one proof term per proof object.

ejgallego:
We would like to drop this restriction in the future but first we need
to continue cleanup of the save proof path.


ejgallego:
@gares as being discussed in the new PR policy the idea is that draft PR don't need to use the `needs: merge of dependency` tag.

gares:
Which new PR policy?

This PR needs the merge of a dependency, it is orthogonal to begin a draft.

ejgallego:
> Which new PR policy?

https://coq.discourse.group/t/pull-request-policy/130/27 (cc: @Zimmi48 )

> This PR needs the merge of a dependency, it is orthogonal to begin a draft.

A draft PR is not ready for review or merge, so in a sense the tag is redundant, if we don't allow "depending" PRs to be submitted into a non-draft state [as to avoid review spam]

I do indeed think that we could deprecate the tag, even tho it may have some uses still.

gares:
> first, I am not sure we should move development discussion here; this is key policy and not all devs are here; IMVHO, this discussion belongs to a GitHub issue. It is not like Discourse is giving us any advanced stuff here, like threading.

Thanks for pointing that out. Nevertheless discussion went on there. Having many media 2 media (GH and coqdev) was bad, not we have three, not sure it is a win @Zimmi48 

Well, let me say here that `needs:` suff, but for `progress` is a metadata that I'd keep on drafts too, since it is a sort of TODO list.

ejgallego:
> Nevertheless discussion went on there. Having many media 2 media (GH and coqdev) was bad, now we have three, not sure it is a win @Zimmi48

Yup, I kind agree, I'd like discourse to be something more "optional", but for core dev issues let's focus on GH. We can continue discussion here when @Zimmi48 does a PR with the policy I guess.

Zimmi48:
> Having many media 2 media (GH and coqdev) was bad, now we have three, not sure it is a win @Zimmi48

We would only have two if people had not told me during the WG that we needed to wait before removing Coqdev. Anyway, I see that you are not following Coqdev either since I advertised this discussion there... GitHub is not a forum software. It provides a bug tracker. Bugs are things that can be fixed / closed. Open discussions do not belong there, as many past examples have shown: we never know when to close them. I'm talking about things like "Organization of a hot freezing day" and stuff like this. And finally, for many people, following all the activity on GitHub is very difficult (apparently this was the case even for our 8.9 RM). On the other hand, Coqdev-to-be-replaced-with-the-development-category-on-Discourse is low activity enough to be possible to follow by anyone.

@gares You can register on Discourse using your GitHub account (possibly after temporarily disabling Adblock+ on Firefox). If you have reasons not to register there, I would like to know (but then I would have preferred of course to know before creating the development category there...).

ejgallego:
> We would only have two if people had not told me during the WG that we needed to wait before removing Coqdev.

@Zimmi48 I understand your point of view but no way we can close coqdev; there is a very large audience there that use mail and are not going to use forums. In fact you could call it a generation gap if you want, but coqdev plays a very important role.

In fact, some of the things that hold discussion is the large number of people following.

> GitHub is not a forum software.

Well, in some sense it certainly is; I am afraid I have seen nothing that Discourse has than Github doesn't, in fact, it has less functionality. I am trolling, I am failing to see why discussing there is better.

And for the policy, indeed, that's an issue, you propose it, it is debated, and then accepted [closed]. With the advantage that having it on GH would allow to keep the critical discussion data where all the other is.

So IMHO critical dev data should be stored in GH.

Zimmi48:
>there is a very large audience there that use mail and are not going to use forums. In fact you could call it a generation gap if you want, but coqdev plays a very important role.

NixOS did close their mailing list. Discourse has a mailing list mode. Generation gaps are not unsolvable problems. Before 2002, the bug tracker was also a mailing list. Should we have stuck with it for generation gap reasons?

>I am afraid I have seen nothing that Discourse has than Github doesn't, in fact, it has less functionality

I can teach you about some functionality that Discourse has that GitHub hasn't. But the main difference is that in a forum, more posts is usually seen in a positive light while on GitHub more bugs is seen in a negative light, especially as we have now developers who refuse to create new issues.

>So IMHO critical dev data should be stored in GH.

This is not critical dev data, this is preliminary request for feedback.

Anyways, this is way off-topic, so I propose we stop this discussion there, and we talk about this at the next WG.

ejgallego:
It seems we won't remove the `start_dependent_lemma` API anytime soon; so closing this for now.

PR #9651, opened 2019-02-25
[ssr] Add tactics under and over

erikmd:
**Kind:** feature.

This PR introduces a new tactic `under` for proofs in math-comp style with the following syntax:

- `under {occs}[patt]lemma => [ipats | ...].` (for interactive mode)
- `under {occs}[patt]lemma => [ipats | ...] do [tac1 | ...]` (for one-liners)

This tactic can be viewed as a facility to rewrite under the binders of objects that come with an "extensionality lemma". This includes in particular math-comp `bigop`s, `polynomial`s, matrices, finite sets…

For example:
```
under eq_bigr => i do rewrite addnC.
```

(see also the PR math-comp/math-comp#292).

In *interactive mode* (if there is no `do …`), a terminator `over` *and* a lemma `over` are provided:
```
under eq_bigr => i.
  rewrite addnC; over.
  (* or alternatively *)
  by rewrite addnC over.
```

More example are available in the `test-suite/ssr/under.v` file.  

This feature has been co-developed with @gares and is a re-implementation of a previous prototype that I had developed in pure Ltac1 (https://github.com/erikmd/ssr-under-tac).

### ToDo

- [x] Add corresponding documentation (including any warning and error messages)
- [x] Add entry in CHANGES.md
- [x] Refactor the syntax w.r.t. the latest feedback
- [x] Update the documentation snippets once the syntax has been refactored
- [x] Fix the regression about `SsrOldRewriteGoalsOrder` that had been introduced at first


gares:
> The only potential breaking change

This is easy to avoid by passing to rewrite an extra argument.

I think I'll carefully review the code on Friday, after the meeting. For now I've seen some rough edges, but most of them are for me to fix (eg some API expose "old" tactics, while we should avoid that...).

gares:
In ssr `by [t1|t2]` is `solve first [t1;t2]` while here it seems that `under ... by [t1|t2]` is like `under ...; [t1|t2]`, that is dispatching `ti` to goal `i`, instead of trying each tactic in each goal. Am I right?

erikmd:
Hi @gares, thanks for your comments! The notation I chose
```
under ... by [t1 | t2].
```
indeed amounts to a dispatch, namely it is some syntactic sugar for
```
under ...; [t1; over | t2; over | move=>/=].
```

According to the reference manual I believed it was the same for Ssr:
https://coq.inria.fr/refman/proof-engine/ssreflect-proof-language.html#coq:tacn.by
> The Ltac expression by [tactic | [tactic | …] is equivalent to [by tactic | by tactic | ...] and this form should be preferred to the former.

but according to your comment I realize this is not the case:
```coq
From mathcomp Require Import ssreflect.
Goal forall b : bool, exists c : bool, b = c.
Fail case; [by fail |by exfalso |by eexists].
case; by [fail | exfalso | eexists]. (* ok, not a dispatch *)
Undo.
Fail case; by do [fail | exfalso | eexists].
case; do [by fail |by exfalso |by eexists]. (* ok *)
```
so maybe the documentation of `by` should be changed?

And regarding the `under` tactic, I guess you may want that I don't use the `by` token nor `do`… do you have suggestions for other tokens?
maybe no token at all before the special dispatch? (I'm unsure if it's possible, nor if this would be legible enough? see below)

```coq
(* Currently: *)
(* cf. test-suite/ssr/under.v:test_big_nested_2 *)
under i Hi: eq_bigr by under j Hj: eq_big by [rewrite muln1 | rewrite addnC].
(* cf. test-suite/ssr/under.v:test_big_occs *)
under i Hi: {2}[in RHS]eq_bigr by rewrite muln0.

(* With no token before the special dispatch: *)
(* cf. test-suite/ssr/under.v:test_big_nested_2 *)
under i Hi: eq_bigr under j Hj: eq_big [rewrite muln1 | rewrite addnC].
(* cf. test-suite/ssr/under.v:test_big_occs *)
under i Hi: {2}[in RHS]eq_bigr rewrite muln0.
```

What do you think?

erikmd:
> regarding the under tactic, I guess you may want that I don't use the `by` token nor `do`…

On the other hand, (1) the phrasing `under i: lem by tac` is similar to ssr tactics such as
```
have h : … by …
```
and (2) it conveys the intuition that `under i: lem by tac` won't leave extra goals opened (as this syntax is equivalent to `under i: lem; [ tac; over | move=>/= ]` where `over` is a terminator…)

So maybe it would be worth it to keep `by` despite the fact it is a dispatch and not a `solve first […]`?

Anyway, we'll be able to discuss this at the meeting :)

ggonthier:
It's a little difficult to judge this proposal fairly, as you've not written down precisely what it's intended to do and when you propose it is should be used.
  That said, from what I gather from the discussion, `do` would seem to be the appropriate keyword, since you intend the tactics to be executed under the binders introduced by a context lemma such as `eq_bigr`. As you point out, `by [t1 | t2]` means `do [done | t1; done | t2; done]` so as `under` doesn't call `done` using `by` would be confusing. I also think the casual reading of `do` is better.
   One caveat: at least in Coq 8.9 `do` is a reserved identifier, not a keyword. I don't think you need to make it into a keyword here, since the `<rewrite_item>` preceding it is not an open term.

erikmd:
@ggonthier

> from what I gather from the discussion, `do` would seem to be the appropriate keyword, since you intend the tactics to be executed under the binders introduced by a context lemma such as `eq_bigr`.
> […]
> One caveat: at least in Coq 8.9 `do` is a reserved identifier, not a keyword. I don't think you need to make it into a keyword here, since the `<rewrite_item>` preceding it is not an open term.

Thanks for your feedback! so I'll try this syntax instead.

> you've not written down precisely what it's intended to do and when you propose it is should be used.

Indeed, I had not elaborated on the behavior of the tactic. Basically, it applies to any rewriting lemma, and it is mostly useful for extensionality lemmas. To give more details on a typical example:
```
Lemma demo (P : nat -> bool) (F : nat -> nat) m : \sum_(0 <= i < m | P i) F i >= 0.
under i: eq_big do [tac1 | tac2].
```
will do the following:
- `rewrite eq_big` − without failing but generating evars.  
  3 subgoals are then created (the side-conditions for the pred and the general term + the main subgoal)
- For each subgoal created (except the main subgoal), if the goal's type is a product, it tries to introduce as many provided names as possible to the context (so it amounts to `move=> i` here) and if the conclusion is a Leibniz equality (e.g. `F1 i = ?Goal i`) then it massages the goal to obtain the provably-equivalent goal (but "locked" w.r.t. `done`) `@Under _ (F1 i) (?Goal i)`, conveniently displayed as `'Under[ F1 i ]`.
- Next step is a dispatch that apply `tac1; over` resp. `tac2; over` on the corresponding subgoals.
  `over` being a terminator that basically turns back `@Under` to `@eq` and instantiates the evar. 
- Finally, `move=>/=` is applied to the (only remaining) main subgoal so there is no spurious beta-redex.

erikmd:
I followed [@gares' suggestion](https://github.com/coq/coq/pull/9651#issuecomment-467501078) for the the backward-compatibility issue, and I fixed another issue related to `SsrOldRewriteGoalsOrder` after testing `under` with mathcomp-dev + math-comp/math-comp#292.

(But the change of `"by" ==> "do"` is still to be done.)

gares:
Notes from the discussion at the MC working group:
1. support `under i _ : ...` and/or `under i ?: ...` when the hyp name is irrelevant
2. in `under i Hi : eq_big` rather do `under [i Hi|i] : eq_big` so that you name what you see in each branch, pick the names from the first one to rename the resulting goal
3. after `over` remove the spurious beta redex (maybe internally do a change)
4. final syntax: `under [pat]lemma => [intros|...] do stuff`

In practice 1 and 2 are about using real intro patterns there, rather than ad hoc syntax (and you gain `Hi` could even be processed with a view!).

plugins/ssr/ssrfwd.ml
```diff
@@ -318,3 +318,130 @@ let sufftac ist ((((clr, pats),binders),simpl), ((_, c), hint)) =
...
+      | n :: names -> Ssripats.(tclIPAT @@ [Ssripats.IOpId n]) <*> intro_lock names
+      | [] ->
+          (* after quell intro patterns, use IpatTemp *)
+          let id =
+            match name with
+            | Name.Anonymous -> Id.of_string (Printf.sprintf "_under_%d_" !tmp)
+            | Name.Name n -> n in
+          incr tmp;
+          Ssripats.tclIPAT [
+            Ssripats.IOpId id;
```
gares: We can use the `+` ipat here (and get rid of the comment about quill and `tmp` as well): https://github.com/coq/coq/blob/master/plugins/ssr/ssripats.mli#L30

plugins/ssr/ssrfwd.ml
```diff
@@ -318,3 +318,130 @@ let sufftac ist ((((clr, pats),binders),simpl), ((_, c), hint)) =
...
+            Ssripats.IOpEqGen (intro_lock names);
+            Ssripats.IOpEqGen (Tactics.revert [id])]
```
gares: If we use `IOpTemporary` this last op can go away

plugins/ssr/ssrfwd.ml
```diff
@@ -318,3 +318,130 @@ let sufftac ist ((((clr, pats),binders),simpl), ((_, c), hint)) =
...
+      | [] ->
+          (* after quell intro patterns, use IpatTemp *)
+          let id =
+            match name with
+            | Name.Anonymous -> Id.of_string (Printf.sprintf "_under_%d_" !tmp)
+            | Name.Name n -> n in
+          incr tmp;
+          Ssripats.tclIPAT [
+            Ssripats.IOpId id;
+            Ssripats.IOpEqGen (intro_lock names);
```
gares: Hum, this is now misnomed, I think it is fine to use this IOp, bet then its name should be changed into IOpTactic (we had it before, but it was only used for the EqGen case, so I renamed it, my bad)

plugins/ssr/ssrfwd.ml
```diff
@@ -318,3 +318,130 @@ let sufftac ist ((((clr, pats),binders),simpl), ((_, c), hint)) =
...
+              let sigma, under_from_eq =
+                Ssrcommon.mkSsrConst "Under_from_eq" env sigma in
+              let ty = EConstr.mkApp (under,args) in
+              let sigma, t = Evarutil.new_evar env sigma ty in
+              sigma, EConstr.mkApp(under_from_eq,Array.append args [|t|]))
+      | _ -> Proofview.tclUNIT ()
+  in
+    aux c
+
+end
```
gares: This looks like a nice combinator (that funnily enough I'd name `under`).

plugins/ssr/ssrfwd.ml
```diff
@@ -318,3 +318,130 @@ let sufftac ist ((((clr, pats),binders),simpl), ((_, c), hint)) =
...
+           let name = Names.Name.mk_name varname in
+           (* Note: we don't explicitly check name \notin free_vars(res) here *)
+           EConstr.mkLambda (name, types, res)
+         end
+     with
+     | DestKO ->
+        ppdebug(lazy Pp.(str"under: cannot pretty-rename all bound variables with destLambda"));
+        term
+
+let overtac gl = ssr_n_tac "over" ~-1 gl
```
gares: Move here the cast from the old type of tactics to the new one, so that the .mli we export the new type.

plugins/ssr/ssrfwd.ml
```diff
@@ -318,3 +318,130 @@ let sufftac ist ((((clr, pats),binders),simpl), ((_, c), hint)) =
...
+        t
+    in
+    ppdebug(lazy Pp.(str"under: to:" ++ pr_econstr_env env evar_map new_t));
+    evar_map, new_t
+  in
+  let undertacs =
+    if hint = nohint then
+      Proofview.tclUNIT ()
+    else
+      let simpl = Ssrequality.simpltac (Simpl ~-1) in
```
gares: There must be a watic to perform `nf_betaiota`, it is done almost systematically after every `refine` but I can't find it now.

plugins/ssr/ssrparser.mlg
```diff
@@ -2650,6 +2650,11 @@ TACTIC EXTEND ssrgenhave2
     V82.tactic (wlogtac ist pats fwd hint false (`Gen id)) }
 END
 
+TACTIC EXTEND under
+  | [ "under" ne_ident_list(names) ":" ssrrwarg(r) ssrhint(hint) ] ->
```
gares: TODO @gares: 
```
  under {occ}[pat]ssrterm =>ipat do [t1 | tn ]
```
then extract from `ipat` the first names of the first branch. If the name is `_` or `?` or ... return `None` (list of `Id.t option`).

plugins/ssr/ssrfwd.ml
erikmd: Done.

plugins/ssr/ssrfwd.ml
erikmd: @gares Taking some inspiration from
https://github.com/coq/coq/blob/d31056ec924ef6e09b28bc3822b427b67a8a300b/plugins/ltac/rewrite.ml#L1567
I've just addressed this in a commit, but only with `nf_beta` which appears to suffice…

plugins/ssr/ssrparser.mlg
erikmd: I've discussed with @ggonthier regarding our question about `move=> *` and he agreed that it could be quite useful to make `under eq_bigr do …` a shortcut for `under eq_bigr => * do …`.

Let's recap the advantages:
* we'll never want that `under eq_bigr do …` be a no-op regarding the intros, because there will be at least one binder somewhere to introduce;
* if the desired behavior is `move=> *` on all goals, we earn 5 characters (` => *`), which can be useful in one-liner mode with several nested bigops or so;
* if we have several subgoals and we don't want to do `=> *` everywhere, we can always write `=>[i| ]`.

erikmd:
Hi @gares! I've just finished the things I promised to do after the meeting, including the documentation.
Tell me if it looks OK to you. I've also updated the checkboxes in the [initial comment](https://github.com/coq/coq/pull/9651#issue-255967233).
Kind regards, Erik

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3636,6 +3636,236 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+  pretty-printed as ``Under[ some_expr i ]``.
+  This change permits to avoid instantiating the evar by accident
+  (e.g., if one uses a proof script such as:
+  :n:`rewrite lem1 // lem2`).
++ Finally, a post-processing step is performed in the main (here, 3rd)
+  subgoal to try keeping (in the object after the rewrite) the same
+  bound variables as the identifers provided in the first
+  intro-pattern branch (here, ``i``).
+
+If the the intro-pattern is omitted, the Ltac expression:
```
```suggestion
If the intro-pattern is omitted, the Ltac expression:
```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3636,6 +3636,236 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+   which corresponds here to the term we want to obtain *after* the
+   rewriting step. We can thus do:
+
+   .. coqtop:: all abort
+
+      rewrite (@eq_G _ (fun _ : nat => 0)); last first.
+        by move=> m; rewrite subnn.
+
+   Otherwise, a more convenient approach consists in relying on the
+   ``under`` tactic that alleviates this bookkeeping:
```
Zimmi48: Everywhere where you put ``` ``under`` ```, it would be preferable to put ``` :tac:`under` ```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3636,6 +3636,236 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+   .. coqtop:: all abort
+
+      Lemma example_G (n : nat) : G (fun n => n - n) n >= 0.
+      under m: eq_G by rewrite subnn.
+
+  ``(* FIXME: replace with do... *)``
+
+The ``under`` tactics enjoys two different modes (the "interactive
+mode", and the "one-liner mode" that was mentioned in the example
+above). The corresponding syntax is as follows:
```
Zimmi48: Even if the SSR chapter still  contains that kind of phrasing, we prefer to avoid "The corresponding syntax is as follows:" and instead put the syntax first, and the full description in the body.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3636,6 +3636,236 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+
+.. tacn:: under {? @occ_switch} {? [ @r_pattern ] } @term {? => @i_item}
+   :name: under
+
+   rewrite under ``term`` in :ref:`interactive mode <under_interactive>`
```
Zimmi48: Use a capital letter, try to make this a full sentence / description.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3636,6 +3636,236 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+
+.. tacn:: under {? @occ_switch} {? [ @r_pattern ] } @term {? => @i_item} do ( @tactic | [ {*| @tactic } ] )
+   :name: under-do
```
Zimmi48: I don't think it needs a specific index entry given how close the two are. If it needs one, the practice would rather be `under ... do`. But can't you just merge the two grammars?

```
.. tacn:: under {? @occ_switch} {? [ @r_pattern ] } @term {? => @i_item} {? do ( @tactic | [ {*| @tactic } ] ) }

   Rewrite under :token:`term`. We distinguish two modes,
   :ref:`interactive mode <under_interactive>` without a `do` clause, and
   :ref:`one-liner mode <under_one_liner>` with a `do` clause, which are
   explained in more details below.
```

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3636,6 +3636,236 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+close intermediate subgoal ``Under[ … ]`` and instantiate the
+corresponding evar:
+
+.. tacn:: over
+   :name: over
+
+   terminator tactic to close goals of the form ``Under[ … ]``
+
+.. tacv:: by rewrite over
+   :undocumented:
```
Zimmi48: This `:undocumented:` option is useless since the tactic is described below.

Zimmi48 (commented):
I only had a cursory look at the doc (didn't try to review the content but only the form). I'll let the @coq/ssreflect-maintainers team take care of reviewing the documentation content.

Zimmi48:
You can fix the linter errors with `git rebase --whitespace=fix`. Do you have the pre-commit hook installed? If you are in a fresh clone and have never run `./configure`, the answer is probably no.

erikmd:
Hi @Zimmi48,
Thanks for your comments! that was very helpful, all the more as at it was my first PR involving some Sphinx documentation. I took your remarks into account, rebased and squashed several commits.

> Do you have the pre-commit hook installed? If you are in a fresh clone and have never run `./configure`, the answer is probably no.

Actually I had run `./configure` on my clone, as I had tested the output with `make doc`, but I guess the pre-commit hook is not triggered by `git rebase -i` alone, so the commits that raised linting issues were old commits, and they were fixed after I did `git rebase -i --whitespace=fix`. Hopefully the lint job will be happy now.

Cheers, Erik

erikmd:
Apparently the automated whitespace didn't work, but happily it was easy to identify the commits at stake and I just force-pushed a new version.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3636,6 +3636,231 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+involves the following steps:
+
++ It performs a full-blown :n:`rewrite {@occ_switch}[@r_pattern]@term`,
+  without failing like in the above example with ``eq_G``, but
+  creating evars.
++ As a 2-branches intro-pattern is provided, it checks that 3 subgoals
+  have been created (the last one being the initial, main subgoal that
+  now contains some evars).
++ In the first 2 subgoals, it introduces ``i``, resp. ``j``, and tests
+  whether these 2 subgoals are some (quantified) equalities involving
```
jfehrle: "some" seems unnecessary

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3636,6 +3636,231 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+  an evar. For example: ``some_expr(i) = ?Goal i``.
++ If this is the case, the subgoal is manipulated to get the
+  provably-equivalent goal ``Under _ (some_expr i) (?Goal i)``,
```
jfehrle: "provably equivalent"

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3636,6 +3636,231 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+  pretty-printed as ``Under[ some_expr i ]``.
+  This change permits to avoid instantiating the evar by accident
```
jfehrle: "permits to avoid" is somewhere between awkward and ungrammatical.  "permits avoiding" is the usual wording.  But then "permits avoiding instantiating" is not so graceful, either.  Maybe just use "This change avoids instantiating" or "This change lets you avoid instantiating",.

CHANGES.md
```diff
@@ -200,12 +200,22 @@ Misc
 
 SSReflect
 
+- New tactic `under` to rewrite under binders, given an extensionality lemma:
+  - interactive mode: `under lem`, associated terminator: `over`
+  - one-liner mode: `under lem do [tac1 | ...]`
+
+  It can take occurrence switches, contextual patterns, and intro-patterns:
```
jfehrle: I would make this "intropattern" here and elsewhere for consistency with the rest of the doc.

jfehrle (commented):
A few minor comments on wording.

CHANGES.md
erikmd: @jfehrle Thanks for your comments.

It seems `intropattern` is almost never used (except for Sphinx identifiers):
```
$ git grep -n -e "intropattern" --and --not -e '@intropattern\|_intropattern\|intropattern_\|`intropattern' "*.rst" | cat
doc/sphinx/proof-engine/tactics.rst:127:   intropattern            : *
doc/sphinx/proof-engine/tactics.rst:235:  If :tacn:`injection` is applicable, the intropattern
```

However, there are several occurrences of `intro-pattern` and `intro pattern`:
```
$ git grep -n -e "intro-pattern" "*.rst"
doc/sphinx/credits.rst:1077:   intro-patterns by Hugo Herbelin and others.
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1742:destructing intro-patterns.
doc/sphinx/proof-engine/ssreflect-proof-language.rst:3722:+ As a 2-branches intro-pattern is provided, it checks that 3 subgoals
doc/sphinx/proof-engine/ssreflect-proof-language.rst:3736:  intro-pattern branch (here, ``i``).
doc/sphinx/proof-engine/ssreflect-proof-language.rst:3738:If the intro-pattern is omitted, the Ltac expression:

$ git grep -n -e "intro pattern" "*.rst"
doc/sphinx/credits.rst:1452:  end of the intro pattern. In addition to that, the use-and-discard flag
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1506:   :n:`[: @ident ]` intro pattern (see section :ref:`introduction_ssr`).
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1577:The :token:`i_pattern`\s can be seen as a variant of *intro patterns*
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1614:from the :token:`i_view` with the ``-`` intro pattern or by putting
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1631:lets one write just ``/myop`` in the intro pattern. Note the scope
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1650:  from the :token:`ident` with the ``-`` intro pattern.
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1686:  of the intro pattern. For example ``move=> + y`` on a goal::
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1727:  does nothing, but counts as an intro pattern. It can also be used to
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1747:Clears are deferred until the end of the intro pattern.
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1763:If the cleared names are reused in the same intro pattern, a renaming
doc/sphinx/proof-engine/ssreflect-proof-language.rst:1806:  with :token:`ident`. If the intro pattern immediately follows a call
doc/sphinx/proof-engine/ssreflect-proof-language.rst:2555:Another frequent use of the intro patterns combined with ``have`` is the
doc/sphinx/proof-engine/ssreflect-proof-language.rst:2676:For this purpose the ``[: name ]`` intro pattern and the tactic
doc/sphinx/proof-engine/ssreflect-proof-language.rst:2689:When intro patterns for abstract constants are used in conjunction
doc/sphinx/proof-engine/ssreflect-proof-language.rst:2702:solved. Goals corresponding to intro patterns for abstract constants
doc/sphinx/proof-engine/ssreflect-proof-language.rst:2718:Last, notice that the use of intro patterns for abstract constants is
doc/sphinx/proof-engine/ssreflect-proof-language.rst:2890:followed by a comma before the usual intro pattern. The identifier
doc/sphinx/proof-engine/ssreflect-proof-language.rst:5366:Note that the NO-OP intro pattern ``-`` can be used to separate two views,
doc/sphinx/proof-engine/ssreflect-proof-language.rst:5518:intro pattern :ref:`introduction_ssr`
doc/sphinx/proof-engine/tactics.rst:120:multiple hypotheses or subgoals.  Common tactics that accept intro patterns
doc/sphinx/proof-engine/tactics.rst:147:Note that the intro pattern syntax varies between tactics.
doc/sphinx/proof-engine/tactics.rst:183:Note that there is no equivalent to intro patterns for goals.  For a goal :g:`A /\ B`,
doc/sphinx/proof-engine/tactics.rst:264:   similar to |SSR|'s intro patterns.
doc/sphinx/proof-engine/tactics.rst:304:   .. example:: intro pattern for /\\
doc/sphinx/proof-engine/tactics.rst:320:   .. example:: intro pattern for \\/
doc/sphinx/proof-engine/tactics.rst:336:   .. example:: -> intro pattern
doc/sphinx/proof-engine/tactics.rst:352:   .. example:: [=] intro pattern
doc/sphinx/proof-engine/tactics.rst:377:   .. example:: (A & B & ...) intro pattern
doc/sphinx/proof-engine/tactics.rst:393:   .. example:: * intro pattern
doc/sphinx/proof-engine/tactics.rst:415:   .. example:: compound intro pattern
doc/sphinx/proof-engine/tactics.rst:428:   .. example:: combined intro pattern using [=] -> and %
doc/sphinx/proof-engine/tactics.rst:1082:   intro patterns.  See the description in :ref:`intropatterns`.
doc/sphinx/proof-engine/tactics.rst:1385:      If :n:`simple_intropattern` is an intro pattern (see :ref:`intropatterns`),
doc/sphinx/user-extensions/syntax-extensions.rst:1739:        - an intro pattern
```

so I propose to change the occurrences I had added in my patch to the most frequent form, `intro pattern`.

erikmd: I did that, and I also reworded the very last occurrence of `intro-pattern` to `intro pattern` in `doc/sphinx/proof-engine/ssreflect-proof-language.rst:1742`.

jfehrle: Makes sense.  I didn't think to search for "intro pattern".

gares:
I've rebased locally and fixed the conflict. Can I push here @erikmd  ?
After that I'll try to fix the parser as I promised.

erikmd:
@gares
> I've rebased locally and fixed the conflict. Can I push here @erikmd ?

Yes, no problem!

> After that I'll try to fix the parser as I promised.

OK great! :+1:

BTW do you know when the 8.10 freeze will occur? (I don't remember the date I was told during the Sophia meeting)

gares:
I don't think we have a date, but it will happen very soon :-(

gares:
OK, the new syntax is in place, the test ported and the implementation slightly simplified.
TODO:
- [ ] `over` should beta-reduce *the other goals*, espeacially the one in which the finally closed evar occurs

Remark for @erikmd and @ggonthier : The new syntax works well for `eq_bigr`, but not very nicely for `eq_big`
```coq
under eq_bigr => i Hi do ..
under eq_big => [i|i Hi] do ..
```
factoring `i` there is not possible since then `[` would mean case, and not dispatch. We could resurrect the `(|)` proposal for explicit dispatching. Iterating `=>` here seems not a good options to me (it would be `under lem => i => [|Hi] do ...`, in the current code it does not even parse I guess).

Caveat: I did not turn `do` into a global keyword, but I did make `=>` not accept it as an ident in order to make the syntax non ambiguous. The alternative is to make `do` a real keyword.

gares:
I forgot to say that `=> ...` is the regular intro pattern, not a mock up. I thought that it may be useful to apply a view to `Hi` en passant.

erikmd:
@gares thanks a lot for your help in tweaking the syntax of the tactic! :+1:

I've just pushed two more commits and rebased on `master` to solve the conflict.
Here is some feedback regarding the remarks you raised, a summary of what I have just implemented, and a question that should be addressed before merging IMO:

### feedback on your questions

> TODO: `over` should beta-reduce the other goals, especially the one in which the finally closed evar occurs.

I tried implementing this in pure Ltac and directly in OCaml, and I am finally puzzled whether this would actually be possible. Indeed: the algorithm that we'd want here would do `current_over; all: cbv beta iota.` However, (1) the toplevel selector `all` can't be used here, and (2) if it were possible at OCaml level, the second tactic won't be executed anyway because the first tactic succeeding makes the second tactic (after `;`) to be ignored. A solution might be to do `cbv beta iota` on all goals before running `current_over`, but this couldn't work either, because **`cbv beta iota` should be executed afterwards** (after the instantiation of the evar).

Do you agree with my analysis or am I missing anything?

Anyway, even if this is not possible and there is a spurious beta-redex in bigop proofs in interactive mode, it doesn't matter very much IMHO, because it is very easy to call `simpl.` or `cbv beta.` in the final goal... and as had suggested @ggonthier during the MC meeting, we could view the interactive mode as a proof-script-elaboration phase, that could afterwards be refactored to the "one-liner" mode that doesn't have this small defect.

> The new syntax does not work very nicely for eq_big
> under eq_big => [i|i Hi] do ..

I guess that by "nicely" you refer to the heuristic to rename the bound variables involved in the bigop (that was indeed a no-op with the `eq_big` under lemma), so I've implemented a simple extension of your code to work-around this (basically, in case of a `=> […|…]` intro-pattern, the heuristic only looks at the identifiers in the first branch). Is this solution OK for you? or would you want to refine it? Anyway, it seems to work well with `eq_big`.

> I forgot to say that => ... is the regular intro pattern, not a mock up. I thought that it may be useful to apply a view to Hi en passant. The code adds an extra |] for the goal that is not "under" the binder.

Great!

### extra feature for `iff`

I recently discussed this with @proux01 who pointed out to me a possible use case with extensionality lemmas from the the Coquelicot library, which will typically have side-conditions with `iff` rather than `eq`.

It was easy to extend the `under` tactic to support this, so do you agree to include this feature as well?

Here are some examples involving lemma [Lub.Lub_Rbar_eqset](http://coquelicot.saclay.inria.fr/html/Coquelicot.Lub.html#Lub_Rbar_eqset) from Coquelicot, and an extra dedicated lemma `ex_iff` that involves an existential binder:

```coq
(* Example 1, cf. [under.v] *)
Parameters (R Rbar : Set) (R0 : R) (Rbar0 : Rbar).
Parameter Rbar_le : Rbar -> Rbar -> Prop.
Parameter Lub_Rbar : (R -> Prop) -> Rbar.
Parameter Lub_Rbar_eqset :
  forall E1 E2 : R -> Prop,
    (forall x : R, E1 x <-> E2 x) ->
    Lub_Rbar E1 = Lub_Rbar E2.

Lemma test_Lub_Rbar (E : R -> Prop)  :
  Rbar_le Rbar0 (Lub_Rbar (fun x => x = R0 \/ E x)).
Proof.
under Lub_Rbar_eqset => r.
over.
Abort.

(* Example 2 *)
Lemma ex_iff R (P1 P2 : R -> Prop) :
  (forall x : R, P1 x <-> P2 x) -> ((exists x, P1 x) <-> (exists x, P2 x)).
Proof.
by move=> H; split; move=> [x Hx]; exists x; apply H.
Qed.
Arguments ex_iff [R P1] P2 iffP12.

Require Import Setoid.
Lemma test_ex_iff (P : nat -> Prop) : (exists x, P x) -> True.
under ex_iff => n.
over.
Abort.
```

### question on a possible regression

As mentioned in the checklist of my [initial post](https://github.com/coq/coq/pull/9651#issue-255967233), there remains a TODO about the `SsrOldRewriteGoalsOrder` option:

I ran the example below in two different environments (latest `coqorg/coq:dev` image, and this rebased PR #9651) and it seems there is a *regression* with the handling of this option, due to this PR… I didn't investigate this yet, but do you think this could be something easy to fix?  

```coq
From Coq Require Import ssreflect.

Lemma test (H : forall n : nat, True -> n - n = 0) : 5 - 5 = 0.
rewrite H.
(* 2 subgoals (ID 261) *)
(*    *)
(*   H : forall n : nat, True -> n - n = 0 *)
(*   ============================ *)
(*   0 = 0 *)
(*  *)
(* subgoal 2 (ID 262) is: *)
(*  True *)
Abort.

(* REGRESSION in the order of subgoals above: it is the same as with: *)

Global Set SsrOldRewriteGoalsOrder.

Lemma test2 (H : forall n : nat, True -> n - n = 0) : 5 - 5 = 0.
rewrite H.
(* 2 subgoals (ID 261) *)
(*    *)
(*   H : forall n : nat, True -> n - n = 0 *)
(*   ============================ *)
(*   0 = 0 *)
(*  *)
(* subgoal 2 (ID 262) is: *)
(*  True *)
```

Sorry for the long message :)
Kind regards,
Erik

gares:
The doc needs fixing.

For the beta, I think I can do it in ML, but I need more time.

erikmd:
@gares thanks for your comment, indeed the CI job regarding the doc was failing; I pushed a new version that fix this, and mention the support of iff in side-conditions.

gares:
I've asked a consultancy to my boolean logic expert @maximedenes and fixed the first-goes-last bug

erikmd:
Awesome! Thank you @gares @maximedenes :+1:

I'm currently doing an interactive rebase to squash fixups and clean-up the branch.
I keep you informed ASAP.

erikmd:
Hi @gares , I've just finished working on my branch and force-pushed. To sum up:

* I did some cleanup with an interactive rebase, and made sure everything compiles.
* I fixed the doc compilation again and remove some whitespace to make the linter happy.
* I fixed a bug with the defective form `under eq_big do [t1|t2]` *that didn't work* because it has to be understood as `under eq_big => [*|*] do [t1|t2]`, that is `under eq_big; move => [*|*|]; …`, not `under eq_big; move => [*|]; …` so I needed to implement some kind of padding, whose parameter was happily known just by counting the number of "do" hints!
* I added a regression test for the previous issue, and I made sure the doc was precise about this
  (namely, `under eq_big => i do [t1|t2]` isn't accepted but `under eq_big => [i|i] do [t1|t2]` is).
* I added another simple output test to make sure that the `'Under[ _ ]` notation works in interactive mode.

I believe the feature is ready but hopefully the CI will be OK :)

Feel free to take a quick look at the documentation I wrote to see if it is OK for you.

gares:
Great! I'll look at it asap.

Quick question: Do we really need different notations for the iff case? I mean we could use the same 'Under[].

erikmd:
Thanks!
Regarding iff, I had tried doing what you suggest but it didn't seem possible to factor-out the notation, as the type of iff is not dependent.

Le 3 avril 2019 08:53:25 GMT+02:00, Enrico <notifications@github.com> a écrit :
>Great! I'll look at it asap.
>
>Quick question: Do we really need different notations for the iff case?
>I mean we could use the same 'Under[].
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/coq/coq/pull/9651#issuecomment-479363466

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

gares:
I think you can have two print only notations that display the same but are associated to very different terms.

gares:
Wrt the doc, I think it is a bit too Dev oriented: it explains (also)how things are implemented. I think you could shuffle the text and put forward how to use it, and keep the fact that it is an erewrite + this + that to a Dev only dedicated section.

gares:
Maybe I'll give the doc a try myself in order to improve its consistency with the rest of the manual

erikmd:
@gares 

> I think you can have two print only notations that display the same but are associated to very different terms.

Ah yes indeed... the only drawback is that it would make it a bit more difficult to easily explain in the doc the difference between the lemma version of the terminator `by rewrite over` and `by rewrite over_iff`:

> .. tacv:: by rewrite over
> 
>    This is a variant of :tacn:`over` in order to close ``Under[ … ]``
>    goals, relying on the ``over`` lemma.
> 
> .. tacv:: by rewrite over_iff
> 
>    This is a variant of :tacn:`over` in order to close ``Under_iff[ … ]``
>    goals, relying on the ``over_iff`` lemma.

erikmd:
> Wrt the doc, I think it is a bit too Dev oriented: it explains (also)how things are implemented. I think you could shuffle the text and put forward how to use it, and keep the fact that it is an erewrite + this + that to a Dev only dedicated section.

OK, I had tried my best to combine both examples and a detailed description of the semantics… but it's true that the `Interactive mode` section in particular, could give more intuition, e.g. with an extra paragraph before enumeration `1. 2. 3. 4. 5.`?

> Maybe I'll give the doc a try myself in order to improve its consistency with the rest of the manual.

OK, feel free to add commits to refine the doc.

Kind regards, Erik

gares:
I was thinking at making `over` a rewrite rule data base gathering both the `over` for `eq` and the `over_iff` for `iff` so that you can always do `rewrite over`. Do you see any problems with this?

erikmd:
You mean setting something like `Definition over := (over_eq, over_iff).` ? why not…
*If this works* (I have not tested it), it would be convenient indeed and one could at the same time replace the `'Under_iff[ _ ]` notation with a similar printing-only notation `'Under[ _ ]` as you suggested.
I need to be away for the keyboard now for ~1h so feel free to test implementing this rewrite rule if you have time.
Cheers, Erik

gares:
> You mean setting something like `Definition over := (over_eq, over_iff).` ?

exactly

gares:
I'm on the doc, feel free to hack the .v/.ml code, but don't update the doc yet plz

gares:
doc updated, please have a look

erikmd:
@gares 
> doc updated, please have a look

Thanks a lot for your improvements! the doc looks very good and your suggestion about the notation and the rewrite rule were successful.
I've just pushed a new commit that implements this, does minor changes in the doc, and implements a very last optimization… namely, beforehand we got:

```coq
Lemma test_big_2cond_all_implied (F : nat -> nat) (m : nat) :
  \sum_(0 <= i < m | odd (i + 1)) (i + 2) >= 0.
Proof.
Fail under eq_big do [rewrite addnC | rewrite addnC].
under eq_big do [move=> ?; rewrite addnC | rewrite addnC].
```
And with the last commit:
```coq
Lemma test_big_2cond_all_implied (F : nat -> nat) (m : nat) :
  \sum_(0 <= i < m | odd (i + 1)) (i + 2) >= 0.
Proof.
under eq_big do [rewrite addnC | rewrite addnC].

**Comment by erikmd** [comment]
Hi @gares, very last documentation question: I mentioned in the doc the last optimization that I added, but maybe it should be rephrased… because on second thought:

`Tactics.reduct_in_concl (Reductionops.whd_all, DEFAULTcast)`

does not exactly correspond to the `hnf` tactic… so feel free to add a commit or amend my last commit to fix this if you have a better phrasing in mind.

Kind regards, Erik

**Comment by gares** [comment]
In your last commit think you should call `Ssripats.ssrsmovetac`:
```
let ssrsmovetac = Goal.enter begin fun g ->
  let sigma, concl = Goal.(sigma g, concl g) in
  match EConstr.kind sigma concl with
  | Prod _ | LetIn _ -> tclUNIT ()
  | _ -> Tactics.hnf_in_concl
end
```
from ssripats (export it if it is not yet). It is the code that is run when you `move.` so I prefer to use the very same code (and in the doc one can say that a `move.` is performed if `do` is omitted).

**Comment by gares** [comment]
I mean, if you omit the ipat, then the default one is `move; move=> *`.

**Comment by erikmd** [comment]
Great! thanks Enrico, I'm going to amend my commit.

**Comment by erikmd** [comment]
Indeed `ssrsmovetac` was already exported :+1:
and if you agree I'll keep the idea of my first attempt of applying the extra tactic (so, `move.`) only on the n first subgoals, not on the (n+1)st one… OK?

**Comment by erikmd** [comment]
> I'm going to amend my commit.

Done… so let's wait for the CI :)

**Comment by gares** [comment]
> so let's wait for the CI :)

hum... ;-)

**Comment by erikmd** [comment]
Hi @gares 

> hum... ;-)

what do you mean? :)

BTW are the CI job failures spurious? should one re-start the pipeline?

**Comment by gares** [comment]
I don't think they are spurious:
```
FAILURES
    ssr/over.v...Error! (should be accepted)
    ssr/over.v...could not be checked (Error!)
```

**Comment by erikmd** [comment]
@gares Good catch!
my bad, I had systematically run the `under.v` and `output/ssr_under.v` tests, but not `over.v`… now fixed.
Cheers, Erik

**Comment by gares** [comment]
Dear @coq/ssreflect-maintainers this PR is finally ready for merge, I believe.

Given I've authored part of the Coq I think one among @maximedenes @amahboubi or @CohenCyril should take care of this as an assignee, while anybody is welcome to review stuff he is competent on (there is quite some new doc, .v files and ml code).

**Comment by CohenCyril** [comment]
I could be assigned, and review the doc and examples. @maximedenes could you review the ml?

**Comment by erikmd** [comment]
Thanks @CohenCyril !

BTW regarding the examples of use (beyond those mentioned in the doc), there are several other examples, more or less typical, in the file [test-suite/ssr/under.v](https://github.com/erikmd/coq/blob/ssr-under/test-suite/ssr/under.v)

**Comment by CohenCyril** [comment]
@erikmd and @gares is the failure still relevant?

**Comment by gares** [comment]
I don't think it is related

File: doc/sphinx/proof-engine/ssreflect-proof-language.rst [hunk 18]
```diff
@@ -3637,6 +3636,272 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+Let us redo the running example in interactive mode.
+
+.. example::
+
+   .. coqtop:: all abort
+
+      Lemma example_map l : sumlist (map (fun m => m - m) l) = 0.
+      under eq_map => m.
+        rewrite subnn.
+        by over.
```
erikmd: @gares during your refactoring of the doc, you had replaced `over.` with `by over.` here. I didn't comment on that because I know the math-comp convention to prepend final lines with the closing tactic `by`.

But reading again [the doc](https://coq.gitlab.io/-/coq/-/jobs/190788110/artifacts/_install_ci/share/doc/coq/sphinx/html/proof-engine/ssreflect-proof-language.html#under-ssr), it now seems to me that we'd better remove the `by` here and just put `over`… (because `over` is itself a terminator, and the readers could believe it isn't, by reading this example in the first place)

WDYT?

Cc @CohenCyril

doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -3637,6 +3636,272 @@ rewriting rule is stated using Leibniz equality (as opposed to setoid
...
+One-liner mode
+``````````````
+
+The Ltac expression:
+
+:n:`under @term => [ @i_item__1 | … | @i_item__n ] do [ @tac__1 | … | @tac__n ].`
+
+can be seen as a shorter form for the following expression:
+
+:n:`under @term => [ @i_item__1 | … | @i_item__n | ]; [ @tac__1; over | … | @tac__n; over | cbv beta iota ].`
```
erikmd: @CohenCyril @gares 
I've just realized this line could be improved to perfectly describe the semantics of the Ltac expr on previous [Line 3805](https://github.com/coq/coq/pull/9651/files#diff-d80ca9da497679f1161bfbf7e7ab7ce1R3805):

```suggestion
 :n:`under @term; [ move ..| ]; move => [ @i_item__1 | … | @i_item__n | ]; [ @tac__1; over | … | @tac__n; over | cbv beta iota ].`
```

but  I propose to commit this only later on (after Cyril's feedback on the doc itself)

doc/sphinx/proof-engine/ssreflect-proof-language.rst
gares: Ok

doc/sphinx/proof-engine/ssreflect-proof-language.rst
gares: I've mixed feeling s about that. What you write is correct, but it also makes the example harder to read and its message more complex: before it was about the extra branch, now it is also about the automatic `move`.
While the latter is useful in practice (when the ipat is omitted) it is just a detail (already documented), while the number of branches not matching the number of goals but still be accepted is way more surprising to me, and  IMO should be the main focus of the example.

BTW, is the extra move always performed or just when the ipat is omitted (and the default `=>*` added)? I think the call to `move` (to do an `hnf`) should only occur in the case the ipat is omitted.

doc/sphinx/proof-engine/ssreflect-proof-language.rst
CohenCyril: Agreed, and the name of the tactic is pretty explicit about it being a terminator!

doc/sphinx/proof-engine/ssreflect-proof-language.rst
erikmd: Hi @gares, thanks for your comment. I fully agree that the example was harder to read and the message was thereby more difficult to grasp. So I just did another push, replacing `under @term; [ move ..| ] => ...` with `(under @term) => ...` because without the parentheses, the initial example was wrong regarding the specification of the extra branch.

> BTW, is the extra move always performed or just when the ipat is omitted (and the default `=>*` added)? I think the call to `move` (to do an `hnf`) should only occur in the case the ipat is omitted.

Good question: actually, the extra (hnf-like) `move.` is always performed, and it has two interesting benefits:

1. it improve user-experience with lemmas such as `eq_big` (e.g. `under eq_big=>[*|i Hi] do [rewrite something|...]`) because if the `move.` was not performed in this case, due to the automatic protection with `'Under[ _ ]` on the conclusion of the goal, **some kind of hnf would basically be performed on the goal** (compare `erewrite eq_big.` with `under eq_big.`) but without the `; [move ..|]` the outcome of `=> [*|...]` would sound weird to the user...

2. thanks to this `move.` which is always done at the proper time (and already documented in the `_under_interactive` section as you pointed out), the final example I pushed is formally correct:
```
(under @term) => [ @i_item__1 | … | @i_item__n | ]; [ @tac__1; over | … | @tac__n; over | cbv beta iota ].
```

erikmd:
Hi @CohenCyril, following your comment I took the opportunity to assign you for this PR…
Hopefully @maximedenes will be able to have a look at the `*.ml` as well.
Have a nice WE!

erikmd:
Dear @CohenCyril,

Following the syntax change implemented in this PR − from `under eq_lem => i by tac.` to the current syntax `under eq_lem => i do tac.` − I remembered your suggestion on the earlier version of the tactic in erikmd/ssr-under-tac#1 :) so I've adapted the code of `under` accordingly − cf. the last commit entitled `"set under's tactic argument at LEVEL 3"`, which is exemplified in the first proof step of the lemma `test_big_andb` in `test-suite/ssr/under.v`:
```
under eq_bigl => i do [rewrite andb_idl; first by move/eqP->].
```
Now, `under eq_bigl => i do tac1; tac2` would amount to `(under eq_bigl => i do tac1); tac2`

(This phrasing was not salient beforehand, as the prior syntax `under eq_bigl => i by tac1; tac2.` was naturally suggesting the parenthesizing `under eq_bigl => i by (tac1; tac2).`)

Are you OK with this very last change?

(BTW I force-pushed because I have squashed 3 commits that were just doc fixups.)

Cc @gares

erikmd:
FYI the 1026 lines added by this PR can be summarized like this:
- `plugins/ssr/`:
  - Coqpp code (`ssrparser.mlg`) +63
  - Coq code (`ssreflect.v`) +97
  - `.mli` spec +26
  - `.ml` code +211
- Tests +340
- Doc +289

vbgl:
Is this ready?

Ping @CohenCyril

gares:
> Are you OK with this very last change? Cc @gares

I was not aware of the discussion, but indeed the new precedence is very consistent with the rest of the language, so I'm fine with it.

CohenCyril:
Hi. I am very happy with the PR including @erikmd's last addition about the precedence of the semi-colon.
However I am a bit puzzled about the choice to distinguish `iff` amongst other setoid equalities. AFAIK, and please correct me @gares if I am wrong, ssreflect dispatches between Leibniz equality (in order to use the inductive elimination principle) and Setoid equality (in order to default to Coq's typeclass based setoid rewrite), and makes no special case for `iff`. I was wondering if there was a particular reason and how difficult it would be to implement this feature in all it's generality.
Cheers

gares:
> implement this feature in all its generality

I think you are right, instead of recognizing `eq` and `iff` one could do as rewrite and call https://github.com/coq/coq/blob/master/plugins/ssr/ssrequality.ml#L459 (that amounts at refining `_ : Rewriteable c` or something like that) but I don't know if one can then easily make `over` cover all cases. @erikmd surely knows

CohenCyril:
I do not want to stall this PR any further, so I see three options, ordered by decreasing personal preference:
1. we wait until @erikmd implements the generalization
2. we keep only Leibniz equality (and open another PR comes to cover any setoid equality including `iff`)
3. we keep it as it is right now (and open another PR comes to generalize from `iff` to any setoid)

erikmd:
Dear Cyril and Enrico, thanks for your feedback; so I'm working on this generalization (option 1).
I let you informed ASAP…

erikmd:
Done − I also updated the doc and the file `test-suite/ssr/under.v`.

erikmd:
@CohenCyril @gares FYI I've just pushed a new version that generalizes the `.ml` code to parameterized setoid relations (i.e., testing `Array.length args >= 2` instead of `Array.length args = 2`)
The related test passed on my machine, but let's wait for the CI anyway :)

gares:
I think that the setoid part of the tactic should be moved out of ssreflect and punt inside the file (or be exported by the file) one usually loads in order to enable setoid rewriting. I don't thin making ssreflect.v depend on setoids/classes/whatever is desirable, esp given that @maximedenes and @vbgl are experimenting with a bootstrapped version of ssr in stdlib2.

erikmd:
@gares 

> I think that the setoid part of the tactic should be moved out of ssreflect and punt inside the file (or be exported by the file) one usually loads in order to enable setoid rewriting. I don't thin making ssreflect.v depend on setoids/classes/whatever is desirable, esp given that @maximedenes and @vbgl are experimenting with a bootstrapped version of ssr in stdlib2.

I understand the objective you describe, but I'm unsure it is feasible not to require `Relation_Definitions` and `RelationClasses` in `ssreflect.v`. Indeed the `over` *rewrite rule* depends on `Under_rel.under_rel_done`, which depends on constants `relation` and `Reflexive`…

The only refactoring I can come up with at first sight is: replacing `Require Import Relation_Definitions` with `Require Relation_Definitions.`

BTW, note that `ssreflect.v` does not require `Setoid` itself, the corresponding line is just in the file [test-suite/ssr/under.v](https://github.com/coq/coq/blob/9785d5437ef25f7c43cbdc2d61ab93ea35622c8b/test-suite/ssr/under.v#L230).

WDYT?

erikmd:
Also, I see another concern to remove default support of arbitrary `relation`s in the environment obtained with `From Coq Require Import ssreflect.`: the proofs depending on `iff` side-conditions, but not on setoid rewriting, wouldn't work anymore.
There is already [a related test in `under.v`](https://github.com/coq/coq/blob/9785d5437ef25f7c43cbdc2d61ab93ea35622c8b/test-suite/ssr/under.v#L217) (note that this test comes before `Require Import Setoid.`)

gares:
I meant moving the entire rel module elsewhere, eg in setoids. But I agree iff should work no matter what is loaded.

erikmd:
@gares
> I meant moving the entire rel module elsewhere, eg in setoids. But I agree iff should work no matter what is loaded.

Yes. So, does this mean you are OK with the current state of the PR?

BTW, if the underlying concern in your suggestion was to ease the porting of the ssr plugin to stdlib2, so it can work *without loading the stdlib1*, it's true that `under` relies on bits of stdlib that is not yet in stdlib2 (however I guess that `under` is not the only tactic or plugin in this case… see e.g. [that import](https://github.com/coq/coq/blob/8202a6394425bfd8db22966fe0ef83441d7b5d04/plugins/ssr/ssreflect.v#L15)).
To overcome this I only see two ways:
1. Copy the theory at stake from stdlib to stdlib2;
2. Move the required bits in both libraries (in the case of `under` this essentially amounts to the class `Coq.Classes.RelationClasses.Reflexive`) in another place "upper in the hierarchy" (e.g., in [stdlib / Relation_Definitions.v](https://github.com/coq/coq/blob/8202a6394425bfd8db22966fe0ef83441d7b5d04/theories/Relations/Relation_Definitions.v) resp. in [stdlib2 / prelude.v](https://github.com/coq/stdlib2/blob/master/src/prelude.v)) so one could get rid of the line `Require Import RelationClasses.` in `ssreflect.v`…

(maybe there is a third way :)
Anyway I guess one of these choices could be performed later on in another PR?

gares:
I was wondering if one could revert the latest change and go back where the code was only about iff and eq.
Then, in another PR, add to the ML code the test for any rewriteable relation and put in `theories/.../Setoid.v` the module your last commits were adding to ssreflect.v. Then, by just loading ssr one gets eq and iff, and if he also loads Setoid he gets support for other relations (the test for rewriteable relations could also check if Setoid is loaded). Do you think this is feasible/reasonable?

erikmd:
> Do you think this is feasible/reasonable?

This seems almost feasible to me: I could indeed resurrect the `Under_iff` module and move the `Under_rel` module to `Setoid`. However, there will be an issue with the following code defined in `ssreflect.v`:

```
Definition over := (over_eq, over_rel).

Ltac over :=
  by [ apply: Under_eq.under_eq_done
     | apply: Under_rel.under_rel_done
     | rewrite over
     ].
```

I don't see how to preserve the ability to do `over.` and `by rewrite over.` with any of the three supported relations (`eq`, `iff`, `RewriteRelation`) without having `Under_rel` in the file…
And, if we redefine this rule and this tactic in `Setoid`, this would need to add `Require Import ssreflect` in `Setoid`, which is probably yet a less wanted thing?

erikmd:
Hi @gares, to follow-up my last comment, it seems to me that the splitting of the definitions in `ssreflect.v` related to (`under` and) `over` is not feasible, and if it were possible with a hack, it might be simpler to keep it as is, and adapt the stdlib2 porting later on accordingly, rather than refactoring this PR "a priori" in a forward-compatible albeit not natural way (?)

I Cc @CohenCyril, @maximedenes and @vbgl in case they have a strong feeling on that question.

gares:
Ok, It is harder than what I thought... We can stay like that

CohenCyril (approved):
LGTM

gares:
@CohenCyril green flag

erikmd:
:tada: many thanks to @gares, @CohenCyril and all people that contributed to the review of this feature!
(either at the math-comp meeting or directly in this PR)

PR #9653, opened 2019-02-26
Fix gitattributes for Makefile.dune

SkySkimmer:
Since it matches *.dune and Makefile* the later needs to come second
in the gitattributes file.


ejgallego:
Thanks I was hit by this.

PR #9654, opened 2019-02-26
[sphinx] Add warn option to coqtop directive.

Zimmi48:
By default Coq warnings are made fatal when building the manual.
If you want to show a command resulting in a warning, use the warn option.

**Kind:** infrastructure.

### DIFF
--- a/doc/sphinx/addendum/canonical-structures.rst
+++ b/doc/sphinx/addendum/canonical-structures.rst
@@ -313,7 +313,9 @@ constructor ``*``. It also tests that they work as expected.
 Unfortunately, these declarations are very verbose. In the following
 subsection we show how to make them more compact.
 
-.. coqtop:: all
+.. FIXME shouldn't warn
+
+.. coqtop:: all warn
 
   Module Add_instance_attempt.
 
@@ -418,7 +420,9 @@ the reader can refer to :cite:`CSwcu`.
 
 The declaration of canonical instances can now be way more compact:
 
-.. coqtop:: all
+.. FIXME should not warn
+
+.. coqtop:: all warn
 
   Canonical Structure nat_LEQty := Eval hnf in Pack nat nat_LEQmx.
   
--- a/doc/sphinx/language/coq-library.rst
+++ b/doc/sphinx/language/coq-library.rst
@@ -606,7 +606,10 @@ Finally, it gives the definition of the usual orderings ``le``,
   single: ge (term)
   single: gt (term)
 
-.. coqtop:: in
+.. This emits a notation already used warning but it won't be shown to
+   the user.
+
+.. coqtop:: in warn
 
   Inductive le (n:nat) : nat -> Prop :=
   | le_n : le n n
--- a/doc/sphinx/proof-engine/ssreflect-proof-language.rst
+++ b/doc/sphinx/proof-engine/ssreflect-proof-language.rst
@@ -2902,6 +2902,7 @@ pattern will be used to process its instance.
 
      Axiom P : nat -> Prop.
      Axioms eqn leqn : nat -> nat -> bool.
+     Declare Scope this_scope.
      Notation "a != b" := (eqn a b) (at level 70) : this_scope.
      Notation "a <= b" := (leqn a b) (at level 70) : this_scope.
      Open Scope this_scope.
--- a/doc/sphinx/proof-engine/tactics.rst
+++ b/doc/sphinx/proof-engine/tactics.rst
@@ -378,7 +378,7 @@ Examples:
 
       .. coqtop:: reset none
 
-         Variables (A : Prop) (B: nat -> Prop) (C: Prop).
+         Parameters (A : Prop) (B: nat -> Prop) (C: Prop).
 
       .. coqtop:: out
 
@@ -730,15 +730,15 @@ Applying theorems
 
    .. coqtop:: reset in
 
-      Variable R : nat -> nat -> Prop.
+      Parameter R : nat -> nat -> Prop.
 
-      Hypothesis Rtrans : forall x y z:nat, R x y -> R y z -> R x z.
+      Axiom Rtrans : forall x y z:nat, R x y -> R y z -> R x z.
 
-      Variables n m p : nat.
+      Parameters n m p : nat.
 
-      Hypothesis Rnm : R n m.
+      Axiom Rnm : R n m.
 
-      Hypothesis Rmp : R m p.
+      Axiom Rmp : R m p.
 
    Consider the goal ``(R n p)`` provable using the transitivity of ``R``:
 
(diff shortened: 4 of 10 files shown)


doc/sphinx/proof-engine/ssreflect-proof-language.rst
```diff
@@ -2902,6 +2902,7 @@ pattern will be used to process its instance.
 
      Axiom P : nat -> Prop.
      Axioms eqn leqn : nat -> nat -> bool.
+     Declare Scope "this_scope".
```
Zimmi48: Currently stucked there.
Without it:

```
Extension error:
/home/theo/git/coq/_build/default/doc/sphinx/proof-engine/ssreflect-proof-language.rst:2771: Error while sending the following to coqtop:
    Notation "a != b" := (eqn a b) (at level 70) : this_scope.
  coqtop output:
    Toplevel input, characters 0-58:
    > Notation "a != b" := (eqn a b) (at level 70) : this_scope.
    > ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    Error: Declaring a scope implicitly is deprecated; use in advance an explicit
    "Declare Scope this_scope.".
```

With it:

```
Extension error:
/home/theo/git/coq/_build/default/doc/sphinx/proof-engine/ssreflect-proof-language.rst:2771: Error while sending the following to coqtop:
    Declare Scope "this_scope".
  coqtop output:
    Toplevel input, characters 8-13:
    > Declare Scope "this_scope".
    >         ^^^^^
    Error:
    Syntax error: 'Equivalent' 'Keys' expected after 'Declare' (in [vernac:command]).
```

Zimmi48: OK the trouble was with the quotes. But the syntax error message is really weird.

SkySkimmer: No quotes ;)

doc/tools/coqrst/coqdomain.py
```diff
@@ -895,14 +900,19 @@ def add_coq_output_1(self, repl, node):
         if options['reset']:
             repl.sendone("Reset Initial.")
             repl.sendone("Set Coqtop Exit On Error.")
+            repl.sendone("Set Warnings \"+default\".")
```
SkySkimmer: You should also send it at the start (L926).
Perhaps we should lift this (re)initialisation stuff to a function.

doc/tools/coqrst/coqdomain.py
```diff
@@ -895,14 +900,19 @@ def add_coq_output_1(self, repl, node):
...
+        if options['warn']:
+            repl.sendone("Set Warnings \"default\".")
         for sentence in self.split_sentences(node.rawsource):
             pairs.append((sentence, repl.sendone(sentence)))
         if options['abort']:
             repl.sendone("Abort All.")
         if options['fail']:
             repl.sendone("Set Coqtop Exit On Error.")
+        if options['warn']:
+            repl.sendone("Set Warnings \"+default\".")
```
SkySkimmer: Might be cleaner using single quotes to avoid the need for escaping (here and the other Set Warnings).

doc/tools/coqrst/coqdomain.py
```diff
@@ -581,6 +581,7 @@ class CoqtopDirective(Directive):
 
       - ``reset``: Send a ``Reset Initial`` command before running this block
       - ``fail``: Don't die if a command fails
+      - ``warn``: Don't die if a command emits a warning
```
SkySkimmer: Should say that fail implies warn.

doc/sphinx/language/coq-library.rst
```diff
@@ -178,7 +178,7 @@ Quantifiers
 
 Then we find first-order quantifiers:
 
-.. coqtop:: in
+.. coqdoc::
```
SkySkimmer: Why change all these? I'd prefer it to break if we change some syntax.

doc/sphinx/addendum/canonical-structures.rst
```diff
@@ -313,7 +313,9 @@ constructor ``*``. It also tests that they work as expected.
 Unfortunately, these declarations are very verbose. In the following
 subsection we show how to make them more compact.
 
-.. coqtop:: all
+.. FIXME shouldn't warn
```
Zimmi48: @vbgl Have you figured out the meaning of the canonical structure warnings?

vbgl: No sorry.

Zimmi48:
I think this is fine to merge even with the two remaining `FIXME` as this already improves the situation quite a lot by removing almost all existing unwanted warnings and preventing that we introduce new warnings in the future.

SkySkimmer:
If I have 
~~~
.. coqtop:: none

  Set Warnings "pedantic".

.. coqtop:: warn ...

  Variables foo : A. (* demonstrate pedantic warning *)
~~~
does it work?

Zimmi48:
It works but that's because the semantics of `Set Warnings` is really strange:

```
Set Warnings "+default".
Set Warnings "pedantic".
Set Warnings "default".
Parameters A : Prop.
Set Warnings "+default".
Set Warnings "-pedantic".
Parameters B : Prop.
```

You would have to think about disabling the pedantic warning at the end though, and using `Parameters` rather than `Variables` would be more efficient to demonstrate the warning without polluting the demonstration with another warning.

Zimmi48:
@cpitclaudel Can you look into reviewing and merging this?

doc/sphinx/proof-engine/tactics.rst
```diff
@@ -3683,11 +3683,7 @@ The general command to add a hint to some databases :n:`{+ @ident}` is
       Local is useless since hints do not survive anyway to the closure of
       sections.
 
-   .. cmdv:: Local Hint @hint_definition
-
-      Idem for the core database.
```
cpitclaudel: Any reason to remove this?  Sorry if I missed something in the discussion

cpitclaudel: If this is deprecated, we should probably still document it, and mention that it's deprecated, shouldn't we ?

doc/tools/coqrst/coqdomain.py
```diff
@@ -891,18 +896,22 @@ def add_coq_output_1(self, repl, node):
         pairs = []
 
         if options['restart']:
-            repl.sendone("Restart.")
+            repl.sendone('Restart.')
```
cpitclaudel: This seems a bit gratuitous

doc/tools/coqrst/coqdomain.py
```diff
@@ -891,18 +896,22 @@ def add_coq_output_1(self, repl, node):
...
+        if options['warn']:
+            repl.sendone('Set Warnings "default".')
         for sentence in self.split_sentences(node.rawsource):
             pairs.append((sentence, repl.sendone(sentence)))
         if options['abort']:
-            repl.sendone("Abort All.")
+            repl.sendone('Abort All.')
         if options['fail']:
-            repl.sendone("Set Coqtop Exit On Error.")
+            repl.sendone('Set Coqtop Exit On Error.')
```
cpitclaudel: Same here and above

cpitclaudel (approved):
LGTM. I left comments, but these don't need to be addressed before merging.

doc/sphinx/proof-engine/tactics.rst
Zimmi48: The main variant documents the option of not putting the database explicitly and says that this is deprecated.

doc/tools/coqrst/coqdomain.py
Zimmi48: Yes, I am a bit obsessed with consistency 😆

PR #9655, opened 2019-02-26
[interp] [numeral] Fix wrong unconditional call to VM reduction.

ejgallego:
PR #8464 introduced an unconditional call to the VM in the
interpretation code.

This is wrong when the VM is not available [such as in JsCoq] and may
create problems.

I am not sure anyways that this PR is the right fix for the issue, it
is my impression that `Vnorm.cbv_vm` should never be called directly,
expert in the reduction API is needed here.

### DIFF
--- a/interp/notation.ml
+++ b/interp/notation.ml
@@ -595,6 +595,7 @@ type numeral_notation_obj = (target_kind, numnot_option) prim_token_notation_obj
 type string_notation_obj = (string_target_kind, unit) prim_token_notation_obj
 
 module PrimTokenNotation = struct
+
 (** * Code shared between Numeral notation and String notation *)
 (** Reduction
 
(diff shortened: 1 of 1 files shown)


ejgallego:
The fix should strongly land in the 8.9 branch.

ejgallego:
> expert in the reduction API is needed here.

cc: @maximedenes @ppedrot

SkySkimmer:
Why not put the test in vnorm?

ejgallego:
> Why not put the test in vnorm?

There is already a check in `Vconv`, but the whole thing seems a bit messy as for example `cbv_vm` skips it, and there is a hook.

I am not sure but indeed it seems that the whole call structure should be tweaked.

ejgallego:
> Why not put the test in vnorm?

So as to concretely answer:
- this seems like the minimal fix for 8.9 [as to get jscoq work again]
- a proper fix may even go beyond adding a check in `vnorm`.

ppedrot (approved):
I am fine with this hotfix.

ejgallego:
Thanks @ppedrot , while the hotfix is fine it seems to me that @SkySkimmer has a point so let's wait for @maximedenes review.

ejgallego:
This is almost a critical fix for 8.9 , ping @maximedenes , maybe we should just do this hotfix and keep the mess as is.

maximedenes:
I believe the test should be moved as @SkySkimmer suggests. @ejgallego do you have time to address it? Otherwise I'll try but this week is difficult for me.

ejgallego:
> I believe the test should be moved as @SkySkimmer suggests. @ejgallego do you have time to address it? Otherwise I'll try but this week is difficult for me.

I can try, the thing is that the whole setup seems a bit messy; anyways this should be backported as it is a fix so no risk on losing the branching point.

JasonGross:
This is superseded by #9874.

PR #9660, opened 2019-02-27
Set COQLIB so the test suite runs locally on Windows.

jfehrle:

### DIFF
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -38,9 +38,9 @@ include ../Makefile.common
 BIN := $(shell cd ..; pwd)/bin/
 COQFLAGS?=
 
+COQLIB?=
 ifeq ($(COQLIB),)
-  # This method of setting `pwd` won't work on win32 OCaml
-  COQLIB := $(shell cd ..; pwd)
+  COQLIB := $(shell ocaml ocaml_pwd.ml ..)
 endif
 export COQLIB
 
--- a/test-suite/dune
+++ b/test-suite/dune
@@ -6,7 +6,7 @@
 
 (rule
  (targets libpath.inc)
- (action (with-stdout-to %{targets} (run ./ocaml_pwd.exe ../../install/%{context_name}/lib/coq/ ))))
+ (action (with-stdout-to %{targets} (run ./ocaml_pwd.exe -quoted ../../install/%{context_name}/lib/coq/ ))))
 
 (rule
  (targets summary.log)
--- a/test-suite/ocaml_pwd.ml
+++ b/test-suite/ocaml_pwd.ml
@@ -1,7 +1,7 @@
 let _ =
-  let ch_dir = Sys.argv.(1) in
+  let quoted = Sys.argv.(1) = "-quoted" in
+  let ch_dir = Sys.argv.(if quoted then 2 else 1) in
   Sys.chdir ch_dir;
   let dir = Sys.getcwd () in
-  (* Needed for windows *)
-  let dir = Filename.quote dir in
+  let dir = if quoted then Filename.quote dir else dir in
   Format.printf "%s%!" dir


jfehrle:
@ejgallego This works on windows now, thanks.

test-suite/Makefile
```diff
@@ -39,8 +39,7 @@ BIN := $(shell cd ..; pwd)/bin/
 COQFLAGS?=
 
 ifeq ($(COQLIB),)
```
vbgl: This variable is used before being defined, it seems.

> Makefile:41: warning: undefined variable `COQLIB'

What about adding a line like `COQLIB?=` before?

jfehrle: Sure, I'll add that.  make doesn't generate such warnings on my system.  Perhaps you're picking up some additional command-line settings?

vbgl: I’m trying to understand the failures of the CI. In both cases, I can see this warning that I find suspicious.

jfehrle: ... though I do see that message in the CI output.

jfehrle: Hmm, that warning should also have appeared without the changes of this PR.

vbgl:
I understand the issue: most test jobs run `cd test-suite; make`, whereas the Nix job runs: `make -C test-suite`.

test-suite/Makefile
```diff
@@ -38,9 +38,9 @@ include ../Makefile.common
 BIN := $(shell cd ..; pwd)/bin/
 COQFLAGS?=
 
+COQLIB?=
```
ejgallego: This is incorrect I think; welcome to the world of make where environment variables and make variables share the same namespace; so IIUC, you are unconditionally destroying the `COQLIB` envar here.

Please remove.

ejgallego (dismissed):

test-suite/Makefile
SkySkimmer: `?=` shouldn't override environment variables

ejgallego: Yet it does?

SkySkimmer: Why do you think it does? I just tested and it didn't.

ejgallego: Or else why it makes the build fail?

ejgallego: My question is more general, why `F=foo make`  and `make F=foo` are not the same? I learned that once, but I forgot. [I am indeed wrong about `Foo ?=` overriding envars]

SkySkimmer: in `make V=value` `$V` will be value unless it's set with a special keyword in the makefile. We never use the keyword so `make V=value` always forces `$V` to be `value`.
Values from the environment are overriden by any assignation except ?=

test-suite/Makefile
```diff
@@ -38,9 +38,9 @@ include ../Makefile.common
...
 ifeq ($(COQLIB),)
-  # This method of setting `pwd` won't work on win32 OCaml
-  COQLIB := $(shell cd ..; pwd)
+  COQLIB := $(shell ocaml ocaml_pwd.ml ..)
```
SkySkimmer: It looks like ocaml_pwd is quoting the filename, then the quotes survive throught the environment and Coq doesn't unquote back so it tries to read `'/bla/'theories`

SkySkimmer: Previously it was used only by dune which uses it through `make ... COQLIB=%{read:libpath.inc}` so the quotes get removed by the shell.

SkySkimmer: Why do we even quote in ocaml_pwd, shouldn't the caller be in charge of quoting as they need? @ejgallego

ejgallego: That's tricky business as far as I remember, the windows filename quoting rules are special so you cannot quote in bash for example AFAICT.

jfehrle: Adding the `echo` does the trick.  Unless one of you has a better fix (e.g. to ocaml_pwd) that you can make in, say, the next week, let's merge this fix.

ejgallego: Maybe we could just require the caller of the `Makefile` to set `COQLIB` else error?

Anyways, I am OK with this fix, thanks @jfehrle

jfehrle: I would avoid changing the requirements for running the test suite unless there's a strong reason to do so.  Less disruption to others.

@gares, would you review and merge if this is agreeable?

gares (commented):
Looks good to me, but I'd like to know why ocaml_pwd quotes. Who wrote it, btw? Imo it should not quote.

jfehrle:
@ejgallego wrote it.  Perhaps he would explain why it quotes.  The only commit for ocaml_pwd.ml is dated 2/16/2019, so it's new.  Not including the added use here, it is only used on one line in test-suite/dune.

ejgallego:
TTBOMK the quotes are necessary for using the path from a shell, windows paths have special quoting rules BTW.

gares:
Yes, quotes are for that but it is typically in the call to the shell that you quote.
Eg
```
DIR=c:\program files\whatever\
BINARY=foo.exe

all:
        '$(DIR)$(BINARY)' -bar 'arg w space'
```
This is the best way I know: you keep the data in raw format in the variables, and add quotes when you call the shell (or when you build a final path). Otherwise you have hard time manipulating the data, eg concatenating `DIR` and `BINARY`, since you have the data in quoted format but concatenation does not understand it, and you end up building `'path'/extra` that is the bug wee see up there, and that was worked around by unescaping before concatenation.

So the `ocaml_pwd` (that is meant to replace `pwd`) should spit the data in raw format (as pwd does), and let the user add quotes if/when he wants.

SkySkimmer:
We quote to send data from dune, through the shell, to make. Quoting in the shell seems unfeasible (it's too late). Maybe dune should be in charge of quoting (although there's no such functionality in current dune).

ejgallego:
As I understand Windows quoting rules are a bit strange so I dunno; all that I can say is that this method works, I dunno about others, I'm not an expert.

ejgallego:
https://github.com/ocaml/ocaml/blob/e0a11fe2e3b7284ffefb2f35930828e011e4928f/stdlib/filename.ml#L140-L163

gares:
> As I understand Windows filenames quoting rules are a bit strange

It depends for which shell you quote too. The makefile assumes, on windows, cygwin and hence a posix shell. If you were calling `cmd.exe`. I don't know what dune does, which shell it uses, but here (in make) we assume a posix shell. So maybe `ocaml_pwd` could take an argument `-quote` (or the inverse).

It is just silly to have a tool we control always quote even if we don't want it to, and then use `echo` to unquote. Let's just get the tool print what we need.

ejgallego:
> I don't know what dune does, which shell it uses

Dune doesn't use a shell; unless you use the `(bash ...)` action, which we do for this particular non-build call I think.

> It is just silly to have a tool we control always quote even if we don't want it to, and then use `echo` to unquote. Let's just get the tool print what we need.

Well the tool was designed to print with quotes, as this is what I needed. [note that where this problem shows is when using an absolute path on Win], so it is part of its spec.

I hope to get rid of the tool anyways once we have a native dune file for the test suite [pending on the Dune plugin PR].

Please feel free to do any change you want, but you may find surprises.

jfehrle:
Just for our education I'll remove the quoting and the echo and see what happens in CI.  If it works or there are simple changes that make it work, we can use those.  And if it's not so simple maybe we live with this small wart.  Eh?

jfehrle:
I suppose ocaml_pwd could take an argument indicating whether to include quotes.  @gares would that be better?

gares:
Yes, I'd prefer that. Thanks

jfehrle:
Done.  I took the liberty of renaming `ocaml_pwd` to `abspath`, I think that name matches its function better.

@ejgallego, I noticed `test-suite/dune` referred to "ocaml_pwd.exe".  The suffix seems Windows-specific but the filename `dune` doesn't seem Windows-specific.  Is this correct?

gares:
> The suffix seems Windows-specific

in the dune slang, `exe` is not windows specific (on linux you can name executable .exe as well, extensions have no special meaning).

ejgallego:
> I took the liberty of renaming `ocaml_pwd` to `abspath`, I think that name matches its function better.

Sorry but I don't like the new name, the function "prints the result of ocaml's `get_cwd`" thus the old name is much more informative. In particular, it prints paths in a format _OCaml can understand_. Please revert the name change.

ejgallego:
> The suffix seems Windows-specific but the filename `dune` doesn't seem Windows-specific. Is this correct?

Indeed the suffix is not windows-specific, for internal binaries dune prepends `exe` to the binaries as to have a more uniform naming.

test-suite/dune
```diff
@@ -1,12 +1,12 @@
 ; The easiest way to generate a portable absolute path is to use OCaml
 ; itself to print it
 (executable
- (name ocaml_pwd)
- (modules ocaml_pwd))
+ (name abspath)
```
ejgallego: Renaming is to be avoided unless here is a good reason; in this case the old name is better so this needs to be reverted.

test-suite/abspath.ml
```diff
@@ -0,0 +1,7 @@
+let _ =
+  let quoted = Sys.argv.(1) = "-quoted" in
```
ejgallego: How come this doesn't emit an `index out of bounds` when called without quote?

ejgallego: Oh indeed that was incorrect.

ejgallego:
Have you tested this works when the Windows path contains special chars / spaces?

jfehrle:
> Have you tested this works when the Windows path contains special chars / spaces?

Obviously this works:

```
$ mkdir "a b"
$ ocaml ocaml_pwd.ml "a b"
C:\users\Jim.DESKTOP-MMPNVGH\Documents\coqSource\coq\test-suite\a b
```

But it's not worth the trouble to test that the script will run under, say, "...\c o q\test-suite".  That may have issues far beyond just COQLIB.  Windows developers avoid using spaces and special characters in pathnames other than : . and \ because it's confusing and you frequently need to quote such names.  (Occasionally I use them in the name of a Word or Excel document--but never for software development.)  I think Linux developers also avoid spaces and special characters in pathnames, right?

ejgallego:
> But it's not worth the trouble to test that the script will run under, say, "...\c o q\test-suite". That may have issues far beyond just COQLIB.

IMHO things should be properly quoted and work, but anyways this whole infrastructure is temporary and will go away soon, as we remove the dependency of a shell to run the test suite.

> Windows developers avoid using spaces and special characters in pathnames other than

Oh well this is not my experience with windows, many folders have space names on them, such as "Program Files".

jfehrle:
> Oh well this is not my experience with windows, many folders have space names on them, such as "Program Files".

That was Microsoft's choice.  Alas.

jfehrle:
OK, it's back to `ocaml_pwd`, the build succeeds.  So I think ready to merge.

ejgallego (approved):

Issue #8019, opened 2018-07-08
native_compute (and vm_compute) is about 14x-16x slower (about 25x-27x, resp.) than extraction+ocamlopt


JasonGross:
#### Version

8.8.0

#### Operating system

Linux

#### Description of the problem

Here is an example, also at https://github.com/JasonGross/slow-coq-examples/blob/master/vm_native_slower_extraction.v
```coq
Require Import Coq.Lists.List.
Require Import Coq.ZArith.ZArith.
Import ListNotations.
Local Open Scope list_scope. Local Open Scope Z_scope.

Fixpoint update_nth {T} n f (xs:list T) {struct n} :=
	match n with
	| O => match xs with
				 | nil => nil
				 | x'::xs' => f x'::xs'
				 end
	| S n' =>  match xs with
				 | nil => nil
				 | x'::xs' => x'::update_nth n' f xs'
				 end
  end.

Definition Let_In {A P} (x : A) (f : forall a : A, P a) : P x := let y := x in f y.
Reserved Notation "'dlet' x .. y := v 'in' f"
         (at level 200, x binder, y binder, f at level 200, format "'dlet'  x .. y  :=  v  'in' '//' f").
Notation "'dlet' x .. y := v 'in' f" := (Let_In v (fun x => .. (fun y => f) .. )).

Definition encodemod := (fun (limbwidth_num limbwidth_den s : Z) (c : list (Z * Z)) (n : nat) =>
 (fun (weight : nat -> Z) (n0 : nat) (s0 : Z) (c0 : list (Z * Z)) (x : Z) =>
  (fun (weight0 : nat -> Z) (n1 : nat) (s1 : Z) (c1 : list (Z * Z)) (p : list Z) (idxs : list nat) =>
   fold_right
     (fun (a : nat) (b : list Z) =>
      (fun (weight1 : nat -> Z) (n2 : nat) (s2 : Z) (c2 : list (Z * Z)) (index : nat) (p0 : list Z) =>
       (fun (weight2 : nat -> Z) (n3 : nat) (p1 : list (Z * Z)) =>
        fold_right
          (fun (t : Z * Z) (ls : list Z) =>
           dlet p2 : nat * Z := (fun (weight3 : nat -> Z) (t0 : Z * Z) (i : nat) =>
                                 nat_rect (fun _ : nat => (nat * Z)%type) (0%nat, fst t0 * snd t0)
[...]

Issue #7273, opened 2018-04-16
vm_compute and cbv are 10x slower at (fun x => f (g x)) than at first doing g and then doing f


JasonGross:
#### Version

8.8+beta1

#### Operating system

Linux

#### Description of the problem

In the single .v file in [bug_slow_vm_PHOAS_06.zip](https://github.com/coq/coq/files/1917180/bug_slow_vm_PHOAS_06.zip), we have the following performance behavior:
```coq
  pose k as k'.
  pose (option_map ToFlat k') as k''.
  pose (option_map FromFlat k'') as k'''.
  pose (option_map (PartialEvaluate false) k''') as k''''.
  Time vm_compute in k''''. (* Finished transaction in 3.206 secs (3.207u,0.s) (successful) *)
  Undo.
  Time time vm_compute in k'''; time vm_compute in k''''.
  (* Tactic call ran for 0.087 secs (0.083u,0.s) (success)
Tactic call ran for 0.284 secs (0.288u,0.s) (success)
Finished transaction in 0.371 secs (0.371u,0.s) (successful)
   *)
  Undo.
  Time time vm_compute in k''; time vm_compute in k''''.
  (* Tactic call ran for 0.062 secs (0.064u,0.s) (success)
Tactic call ran for 3.216 secs (3.215u,0.s) (success)
Finished transaction in 3.278 secs (3.279u,0.s) (successful)
*)
```
Notice that `vm_compute`ing `(option_map FromFlat k'')` as `k'''` and then `vm_compute`ing `(option_map (PartialEvaluate false) k''')` is 10x faster than `vm_compute`ing `(option_map (PartialEvaluate false) (option_map FromFlat k''))`.  Using `cbv` rather than `vm_compute` gives times that are about 7x slower in all cases, and also show the same 10x disparity.  Using `lazy` is about 1.5x--2x faster than using `vm_compute` on the composition, but is a bit slower than `vm_compute` at each pass separately (which is still a bit more than 3x faster than `lazy` at the composition).  Presumably this has something to do with lambdas, because these are PHOAS passes, but what's going on here?

ppedrot:
Where can I find the `Crypto.Util.Option` module (or alternatively, can you inline it in the example)?

JasonGross:
[...]

Issue #7159, opened 2018-04-03
Requesting fine-grained control of evaluation order


andres-erbsen:
#### Description of the problem

In several cases it would be very useful to directly control relative evaluation order of subterms of a larger term. Here are two examples that would have been very helpful to me if they had been available, though it is not obvious that just adding these two is the best way to improve the situation.

1. Hand-picked ordering `reduce_and_then {A P} (x:A) (f: forall x, P x) := let v := x in f v.` should reduce `x` to its normal form before substituting it into `f`. Silly example: `reduce_and_then (fun (_:unit) => 1+1) (fun x => x tt*x tt)` should compute `1+1` once. Real example: in a PHOAS-based compiler, passing a polymorphic syntax tree into a compiler pass that requires two instantiations of its var type should not double-compute all previous compiler passes; currently the only way I know for avoiding it is to convert out of PHOAS and back (https://github.com/mit-plv/fiat-crypto/pull/340/files#diff-d6d787cba38e2097ad545be3240b32feR6047).
2. Avoiding reduction in a (proof) term `reduce_never {T} (x:T) := x` would be the same as Lean's old [`id_locked`](https://github.com/leanprover/lean/commit/47994fe14ec7982d5b727c4f8a4f29ae9abce95c). Canonical example `exist prime (2^127-1) (reduce_never ltac:(eauto with primes))`, followed by unsuspecting computation on that value. I hear @JasonGross may have requested this one before, but I failed to find the issue.

I am primarily interested in using these with `vm_compute`.

Zimmi48:
Are you aware of the `Arguments ... : simpl never.` command? It is documented here: https://coq.inria.fr/refman/tactics.html#hevea_tactic155. Sorry if this is not what you were requesting in your second point.

JasonGross:
I believe both points are requests for ways to modify the behavior of cbv, lazy, vm_compute, native_compute.  As far as I know, simpl never only applies to simpl and cbn, and only applies to an identifier, not to its arguments

JasonGross:
The lack of `reduce_and_then` is a severe problem for PHOAS passes.  I have a case where if I convert a de Bruijn AST into a PHOAS AST in `vm_compute` (which takes about 2-3 seconds), and then run `vm_compute` on a PHOAS pass on the AST, that takes about 23 seconds; if I run `vm_compute` in the composition of the conversion from de Bruijn and the PHOAS pass, it takes over 10x longer(!), about 313 seconds.

[...]

PR #9665, opened 2019-02-27
[dune] Enable optimization options in the compilation of the VM.

ejgallego:
As we were waiting for better control of the optimisation flags in
Dune [likely coming in 1.8], we didn't setup optimization flags for
the VM in the Dune build.

However time has come to experiment with such flags, we try -O3 and
enable -flto in the final binary build. This should provide pretty
good results, and hopefully be stable.

Does anyone have a good benchmark for the VM reduction machine?

Maybe @JasonGross ?

### DIFF
--- a/dune
+++ b/dune
@@ -8,6 +8,13 @@
  (ocaml409
   (flags :standard -strict-sequence -strict-formats -keep-locs -rectypes -w -9-27+40+60 -warn-error -5 -alert --deprecated)))
 
+; Information about flags for release mode:
+;
+; In #9665 we tried to add (c_flags -O3) to the release setup,
+; unfortunately the resulting VM seems to be slower [5% slower on
+; fourcolor, thus we keep the default C flags for now, which seem to
+; be -O2.
+
 ; The _ profile could help factoring the above, however it doesn't
 ; seem to work like we'd expect/like:
 ;
--- a/topbin/dune
+++ b/topbin/dune
@@ -26,6 +26,8 @@
  (package coq)
  (modules coqc_bin)
  (libraries coq.toplevel)
+ ; Adding -ccopt -flto to links options could be interesting, however,
+ ; it doesn't work on Windows
  (link_flags -linkall))
 
 (install


JasonGross:
> Does anyone have a good benchmark for the VM reduction machine?

Legacy fiat-crypto (https://github.com/mit-plv/fiat-crypto/tree/sp2019latest?files=1), `make display`.  Note that this also stresses many things that are not the vm (conversion, Ltac reification, printing), and also some of the .vo files there take hours and tens of gigabytes of RAM to build (though nothing in the critical path; if you first `make nonautogenerated-specific-display` or `make selected-specific-display `, that should not be too hard on your machine, and will build all the prerequisites; you can then build whichever files you want; in general, I believe the Montgomery ones stress conversion more than the vm (solinas ones stress conversion less), the `{solinas,montgomery}32_*` ones take longer than `{solinas,montgomery}64_*` ones, and the larger the first number after the first `2e`, the slower the file will be / the more it will stress things.  If you want a not terrible benchmark, just use the current targets in legacy fiat-crypto in the CI, or use `make nonautogenerated-specific-display selected-specific-display`.

JasonGross:
I believe fiat-crypto legacy is included in the Jenkins bench (if not, that should be fixed), so the standard bench should get you some amount of data.

ejgallego:
Thanks a lot @JasonGross , this is useful!

I do think however that given the nature of the patch, a micro-benchmark would be more useful, as to be able to properly measure the impact of such flags. Let me know if you can think of any, otherwise I'll to extract some complex VM call in fiat.

ejgallego:
No idea what is going on in Windows :S

JasonGross:
When I want to test slowness of reduction in Coq, I generally stick a bunch of copies of `Z.div_eucl` in a tuple (without any arguments) and reduce that term.  It gets big pretty quickly due to all of the bare fixpoints.  Of course, this stresses the vm on how it handles large terms, not big computations.  If you want to stress big computations, hm, maybe if you sort a list of strings where you compare two strings by converting each character to a `nat` (not `N`) and comparing the `nat`s?  (I seem to recall @ppedrot finding that some slowness in fiat-parsers was due to that.  Speaking of which, there might be some more vm tests in fiat-parsers ; I believe the CI does not test the files which stress the vm (and some other parts of the system) the most; the `parsers-all` target in fiat will build all of the fiat-parsers stuff the CI does, plus some monstrously slow examples.)

JasonGross:
Also, re microbenchmarks, there are a bunch of calls to the vm in https://github.com/mit-plv/fiat-crypto/blob/master/src/PushButtonSynthesis/SmallExamples.v (and a couple more in https://github.com/mit-plv/fiat-crypto/blob/master/src/SlowPrimeSynthesisExamples.v) in a way that's pretty easy to factor; you can take all of the things passed to `Compute` and make them definitions (so that reification by tactics in terms happens earlier), and then run them through `Compute`.  If you want something slower than any of the things in that file, you can use 

```coq
Compute
     (Pipeline.BoundsPipeline
        true None [64; 128]
        ltac:(let r := Reify (fun f => carry_mulmod 521 9 (2^521) [(1,1)] 9 (seq 0 9 ++ [0; 1])%list%nat f f) in
              exact r)
               (Some (repeat (@None _) 5), tt)
               ZRange.type.base.option.None).
```
I haven't tested it, but the extracted version of this can takes roughly 30 seconds.  (I can give you an even slower one, where the extracted version takes about 3 minutes, if you prefer.)

JasonGross:
Also, there are perhaps some useful microbenchmarks in https://github.com/coq/coq/issues/8019 and in https://github.com/coq/coq/issues/7273 (which is closed not because it is fixed, but because it is subsumed by the wish of https://github.com/coq/coq/issues/7159)

ejgallego:
Thanks a lot @JasonGross , that's a load of useful information.

SkySkimmer (approved):

ejgallego:
Not a good result on the bench tho: https://ci.inria.fr/coq/job/benchmark-part-of-the-branch/690/console
```
┌───────────────┬─────────────────────────┬───────────────────────────────────────────┬─────────────────────────────────────────────┬─────────────────────────────┬───────────────────────┐
│               │      user time [s]      │                CPU cycles                 │              CPU instructions               │    max resident mem [KB]    │      mem faults       │
│               │                         │                                           │                                             │                             │                       │
│  package_name │     NEW     OLD PDIFF   │              NEW              OLD PDIFF   │               NEW               OLD PDIFF   │       NEW       OLD PDIFF   │    NEW    OLD PDIFF   │
├───────────────┼─────────────────────────┼───────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────────────────┼───────────────────────┤
│ coq-fourcolor │ 2674.96 2546.54 +5.04 % │    7458073525689    7098829055235 +5.06 % │    14004766866277    12934609806969 +8.27 % │    971648    971676 -0.00 % │    100    109 -8.26 % │
└───────────────┴─────────────────────────┴───────────────────────────────────────────┴─────────────────────────────────────────────┴─────────────────────────────┴───────────────────────┘
```

ejgallego:
With `-O2`: https://ci.inria.fr/coq/job/benchmark-part-of-the-branch/691/console
```
┌───────────────┬─────────────────────────┬───────────────────────────────────────────┬─────────────────────────────────────────────┬─────────────────────────────┬────────────────────────┐
│               │      user time [s]      │                CPU cycles                 │              CPU instructions               │    max resident mem [KB]    │       mem faults       │
│               │                         │                                           │                                             │                             │                        │
│  package_name │     NEW     OLD PDIFF   │              NEW              OLD PDIFF   │               NEW               OLD PDIFF   │       NEW       OLD PDIFF   │    NEW    OLD  PDIFF   │
├───────────────┼─────────────────────────┼───────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────────────────┼────────────────────────┤
│ coq-fourcolor │ 2552.03 2553.82 -0.07 % │    7119240424984    7117124667816 +0.03 % │    12935213366163    12935063422055 +0.00 % │    971248    971248 +0.00 % │    100    179 -44.13 % │
└───────────────┴─────────────────────────┴───────────────────────────────────────────┴─────────────────────────────────────────────┴─────────────────────────────┴────────────────────────┘
```
not a good result then, I guess I'll just have this PR document the non-usefulness of the tweak.

Zimmi48:
Not sure I understand what the status of this is.

ejgallego:
> Not sure I understand what the status of this is.

@Zimmi48 , I've updated the PR. Adding -O2 to c_flags resulted in a slower VM overall so I have just documented that fact in the dune file.

IMHO we can merge this.

silene:
Can you test on something that actually makes heavy use of the VM? Because adding `-O2` originally gave a +130% speedup: 2887393cf817d0509caf7a2bb8f7850e2bc2d123

ejgallego:
Oh sorry indeed @silene , my comment in the PR was wrong. `-O2` is already the default for all Dune builds, it was -O3 which provided worse performance. Corrected.

Zimmi48 (approved):

PR #9666, opened 2019-02-28
[dune] Support compiling libraries with the native compiler.

ejgallego:
This is mostly a proof of concept/RFC; we likely want to refine this
more, so not ready for merge yet.

In particular it has the problem of dropping the `.coq_native` prefix,
which is not such a big deal for hygienic builds but it is cumbersome
for the rest.

This is due to a limitation of the build language, such limitation
doesn't exist on the native Coq Dune mode.

Ideally we would like to isolate native runtime dependencies on its
own library. That's not too hard to do with Dune's `copy_files`.

Another thing to do in the native Coq mode is to add a `(mode native)`
flag.


ejgallego:
This should stress native on Azure, let's see.

ejgallego:
@MSoegtropIMC the Azure build seems to work and it should have native compilation enabled.

So indeed this is not a problem of `bash` but it seems more like a problem of path handling. But maybe it even works today?

MSoegtropIMC:
> the Azure build seems to work and it should have native compilation enabled.

Could you give a bit more details on how you test this (on Windows?)

For windows it would for sure also mean to add quite a number of tools to the installer (an OCaml compiler most notably). I have all this ready - up to now the main issue was that ocamlbuild & Co do not really work on Windows without a bash.

ejgallego:
> Could you give a bit more details on how you test this (on Windows?)

I am not sure I understand what info you are missing? The setup is the standard azure one in the CI, documented in the `azure-pipelines` file, which is opam mingw AFAICT, it seems pretty straightforward to me.

> the main issue was that ocamlbuild & Co do not really work on Windows without a bash.

You mean ocamlfind? It so that can be also solved by Dune as it doesn't need ocamlfind to work.

MSoegtropIMC:
> I am not sure I understand what info you are missing

I mean what do you do to test if native compute on Windows works in this build. I guess I can neither login to Azure to do manual test on a CI build nor does it produce artifacts I could download for a local test.

> You mean ocamlfind?

When I looked into this last (which is a few years back) native compute was calling the OCaml compiler via ocamlbuild. It might also have been a different program but there was something in between Coq and ocamlc which did require fancy shell support, and back then this was the main issue with native compute. How does this work today?

ejgallego:
> I mean what do you do to test if native compute on Windows works in this build.

It should be producing the native files and the test suite should check basic things about native compilation. You could test locally by using this PR.

> When I looked into this last (which is a few years back) native compute was calling the OCaml compiler via ocamlbuild.

Umm, native has never used ocamlbulid, I think you mean `ocamlfind`. This is still used today but could be dropped. I have no idea on how's windows support in ocamlfind these days, but I would not be surprised if it has improved a lot.

I guess I can just enable it in the win64 build and we can see.

ejgallego:
So it failed with the regular error, https://coq.gitlab.io/-/coq/-/jobs/169371665/artifacts/artifacts/buildlogs/coq-local-make_err.txt let me fix the regular build.

ejgallego:
So indeed it seems to _almost work_ https://coq.gitlab.io/-/coq/-/jobs/169488781/artifacts/artifacts/buildlogs/coq-local-make_err.txt:

```
The command line is too long.
** Fatal error: Cannot run cygpath -m "libmoldname" 
```
the opam windows Job doesn't seem to suffer from this problem tho.

SkySkimmer:
We put way too much stuff in the environment, let me try a thing.

Makefile.build
```diff
@@ -11,6 +11,9 @@
 # This makefile is normally called by the main Makefile after setting
 # some variables.
 
+# Cleanup environment (avoids filling it up)
+unexport $(COQ_EXPORTED)
```
SkySkimmer: Looks like it worked.

ejgallego: > Looks like it worked.

Cool! So indeed this provides a path to enable native on Windows I think.

As for Dune, let's see what we do.

MSoegtropIMC:
> The command line is too long.

The mini-Posix shell I have doesn't suffer from this. In Windows it depends a lot on how you start a process. There are layers on layers on function to create processes. The most basic ones have fairly few restrictions, but this is not what most tools are using. My plan is to replace the OCaml system function with something that gives good Linux / Windows compatibility and doesn't suffer from such crappy restrictions. If you think this is indeed worthwhile, I could pick this up. It is almost finished and should be more than good enough for the purpose at hand.

ejgallego:
IMHO @MSoegtropIMC I am failing to see a strong motivation to replace OCaml internals, with a modern toolchain I think all these issues are taken care of. Indeed the problem in the win makefile was what @SkySkimmer fixed and I think there are no other problems around that would justify such effort.

ejgallego:
@SkySkimmer can you submit the Makefile fix [+ enabling the native compiler on windows] as a separate PR? IMO that's ready to merge, proper native support will take a bit more I think as we need to better define the linking mode we want.

Zimmi48:
This broke the test-suite in Nix: https://gitlab.com/coq/coq/-/jobs/169525384/artifacts/browse/nix-build-coq.drv-0/k6khgfql6w36r17p055yh3c0bim6l7yj-source/test-suite/logs/. Since it is not yet using Dune, it must be some changes in the Makefile...

maximedenes:
@ejgallego cleaning-up my list of assigned PRs, ping me again if some action is required on this PR.

ejgallego:
This has been superseded by a native implementation in dune and #11581

PR #9667, opened 2019-02-28
Use make -n in create_overlays instead of trying to compile.

SkySkimmer:
It's not perfect but I think it's better than the current state.

### DIFF
--- a/dev/tools/create_overlays.sh
+++ b/dev/tools/create_overlays.sh
@@ -61,9 +61,12 @@ do
     _CONTRIB_GITURL="https://github.com/$DEVELOPER_NAME/$_CONTRIB_GITSUFFIX"
     _CONTRIB_GITPUSHURL="git@github.com:$DEVELOPER_NAME/${_CONTRIB_GITSUFFIX}.git"
 
-    # This should work better: for example we should be able not to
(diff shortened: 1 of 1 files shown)


SkySkimmer:
Actually nevermind I feel like this is asking for bugs since the scripts don't expect to be run with -n.

SkySkimmer:
Instead we should use an environment variable `CHECKOUT_ONLY` and use it in the CI scripts.

ejgallego:
Yup this needs to be fixed, I was lazy. We could maybe use make targets for that?

SkySkimmer:
Alternatively we could tell people to use a workflow like
- while not `make ci-broken`; do fix
- run create-overlays broken
- commit in ci-broken (or create-overlays could auto commit and push)

and remove the `make` line from the script.

ejgallego:
Umm, I'm not sure I understand this workflow, seem pretty cumbersome. My current workflow is:
- do change, push to CI; observe broken,
- `create-overlays ejgallego XXX broken`
- make -f Makefile.dune check # this is M-c in my emacs and critically builds all the broken overlays at once.

Then I just wait for emacs to send me to the broken place, I do commit and push from emacs too.

PR #9675, opened 2019-03-01
Dune: remove -short-paths flag.

SkySkimmer:
Take this example:
~~~ocaml
module M = struct
  type a
  let f (x:a) = x
end

type b = M.a

let _ = M.f 2
~~~
With this PR get this error message:

    Error: This expression has type int but an expression was expected of type M.a

instead of the following one:

    Error: This expression has type int but an expression was expected of type b

The first error message is best because we have numerous aliases in Coq's source code (eg for backward compatibility, of for documentation, eg `module_ident = Id.t` in names). Of course you don't want to see `module_ident` wherever an `Id.t` is used.

See https://github.com/ocaml/dune/issues/1794 https://caml.inria.fr/mantis/view.php?id=7911

### DIFF
--- a/dune
+++ b/dune
@@ -1,12 +1,12 @@
 ; Default flags for all Coq libraries.
 (env
- (dev     (flags :standard -rectypes -w -9-27+40+60))
+ (dev     (flags :standard -rectypes -w -9-27+40+60 \ -short-paths))
  (release (flags :standard -rectypes)
           (ocamlopt_flags -O3 -unbox-closures))
  (ireport (flags :standard -rectypes -w -9-27-40+60)
           (ocamlopt_flags :standard -O3 -unbox-closures -inlining-report))
  (ocaml409
-  (flags :standard -strict-sequence -strict-formats -short-paths -keep-locs -rectypes -w -9-27+40+60 -warn-error -5 -alert --deprecated)))
+  (flags :standard -strict-sequence -strict-formats -keep-locs -rectypes -w -9-27+40+60 -warn-error -5 -alert --deprecated)))
 
 ; The _ profile could help factoring the above, however it doesn't
 ; seem to work like we'd expect/like:


ejgallego:
I am split on this as dropping this flag makes other errors much worse. I dunno, what do other devs think @coq/core ?

ejgallego:
As for myself, I am open to merge this and see if my life becomes much worse, so that's a yes in principle.

SkySkimmer:
What errors are worse?

Gaëtan Gilbert

ppedrot:
At worst this is easy to revert...

ejgallego:
> What errors are worse?

Module aliases [for example `core` and `stdlib` etc....] may get you into trouble here, printing stuff such as:

```
Core__foo.Bar.type
```

wrapping of libraries may introduce some annoying cases.

ejgallego:
Anyways indeed no harm to try, but I'd leave a few days so others devs can comment on the default.

gares:
It looks like automatic packing as done by dune needs this option is order to be hidden from the user.
But if the error message is not coming from automatically generated packing files, then the option is unlikely to make things simpler to read.

My take is that his is a bug in the automatic packing dune does. They should ask for upstream support and make error message nice another way, or in other words make the behavior enabled by this flag "local" to packing files. Maybe an attribute is sufficient, eg `[@ unfold_in_error_msgs ]`.

Re-using `private` or `Box of t [@@unboxed]` as proposed in the two threads seems ta bit hackish to me (they overload another meaning).

gares (approved):

SkySkimmer:
>My take is that his is a bug in the automatic packing dune does. They should ask for upstream support and make error message nice another way,

According to the doc and unless I misunderstand what you mean that's already done https://caml.inria.fr/pub/docs/manual-ocaml/extn.html
>Note the use of double underscores in Mylib__A and Mylib__B. These were chosen on purpose; the compiler uses the following heuristic when printing paths: given a path Lib__fooBar, if Lib.FooBar exists and is an alias for Lib__fooBar, then the compiler will always display Lib.FooBar instead of Lib__fooBar. This way the long Mylib__ names stay hidden and all the user sees is the nicer dot names. This is how the OCaml standard library is compiled.

ejgallego:
Note that the packing method is an ocaml packing method (used by stdlib for example) not exclusive to dune.

I am not very fond of the original reason tho, I am not so sure these type aliases we have are indeed useful in the first place.

ejgallego:
The more I think about it the less convinced I am about deviating from the standard ecosystem default.

Using type aliases for documentation seems like an idea doomed to fail. In the presence of aliases, printing is indeed done by an heuristic, so nothing guarantees that short paths will provide you the desired results.

In this context types seems like a very poor form of documentation, polluting the API. Instead, that should be rather done using named parameters and records.

SkySkimmer:
I don't care if we remove the aliases, but so long as they're there we shouldn't use short-paths.

ejgallego:
> the option is unlikely to make things simpler to read.

Why? There is consensus by OCaml experts that this option is preferred. Do you have an example?

For example, in the original report, I do prefer `M.a` as I find the aliases confusing. [Like all this aliases on bool we have, every time i need to use the API I need an extra lookup to see what `polymorphic_flag` is supposed to be]

Let's wait a bit more to see what other developers do say.

ejgallego:
> I don't care if we remove the aliases, but so long as they're there we shouldn't use short-paths.

I am failing to see the reasoning here indeed. Why would you ever prefer to get `module_ident` instead of `Id.t` ? In the first case, unless you have memorized what the alias is, you have to do an extra lookup to understand what is going on, however, if you we `Id.t` it is very clear what the problem is.

SkySkimmer:
>I am failing to see the reasoning here indeed. Why would you ever prefer to get module_ident instead of Id.t ?

There's a misunderstanding, it's only after this PR that you'll see Id.t. In current master you'll see module_ident.

ejgallego:
> There's a misunderstanding, it's only after this PR that you'll see Id.t. In current master you'll see module_ident.

That's the reverse of what's written in the issue right?

ejgallego:
Ok I think I got it wrong; thus I do indeed support trying the flag.

In fact I was surprised about this as I am bothered by the aliases now, so I was wondering how it could be the case.

ejgallego:
Let's leave around 1 week for discussion, then if no developer opposes I'll merge given the current support.

dune
```diff
@@ -1,12 +1,12 @@
 ; Default flags for all Coq libraries.
 (env
- (dev     (flags :standard -rectypes -w -9-27+40+60))
+ (dev     (flags :standard -rectypes -w -9-27+40+60 \ -short-paths))
```
ejgallego: IMHO it would be good if we could comment a bit more the reasoning for the flags [in the style of what we do in configure]. It is important to document this stuff well.

In particular, we may add something signalling that `short-paths` is needed due to aliases, but would we remove the remaining aliases we could also add the flag back.

gares:
It seems the summary of the pr is misleading. 

After re reading the doc I really don't know why they put the flag in the first place. Maybe it was useful before that heuristic got implemented in the compiler.

SkySkimmer:
The flag is more recent than that, it was added around dune 1.7.
I don't see how the PR summary is misleading so can't improve it, feel free to edit it though.

gares:
Done. BTW I don't see which kind of discussion we need here.

ejgallego:
Indeed not a lot of discussion needed anymore I think; anyways I'll leave a few more days in case some OCaml or Coq developer would like to comment. I'll be away anyways so merge target for this is Sunday/Monday.

gares:
I've tried to use dune last week and this Pr is clearly needed, error messages are super confusing.
We have so many alias that no base type get's printed as it is (I'm talking about string an int...) in error messages. The aliases in question, in tok/grammar, are reasonable to me, so I really think this PR is the right thing to do.

ejgallego:
Ok, let's try, but I have been warned by OCaml people that this is a bad step to make in general; indeed why are you using an alias to a base type if you want the error display the base type?

Seems very weird, if the alias are, as quoted above "reasonable" or "informative" why do we want them eliminated?

SkySkimmer:
We don't want them eliminated, we just don't want to get an alias out of nowhere.

ejgallego:
> We don't want them eliminated, we just don't want to get an alias out of nowhere.

I mean we have for example:
```
type allow_implicits = bool

val pretype : allow_implicits -> constr -> constr
```
so if we do:
```
pretype 3 c
```
now we get

"Error: expected `allow_implicits` but we got `int`"

right? However we want:
```
"Error, expected `bool` but we got `int`"
```
But that's a bit weird, if the alias has informative value, why would we like to lose it? Doesn't it mean that indeed the alias makes little sense?

SkySkimmer:
>"Error: expected allow_implicits but we got int"

This is what we get both before and after the PR, and also with make which doesn't use -short-paths (although I think it says "expected type allow_implicits = bool").

Actually I tested it by pasting the following into impargs and changing my .merlin to add/remove short-paths:
~~~ocaml
type al = bool
let f : al -> int = fun _ -> 2

let _ = f 2
(* with -short-paths: This expression has type metavariable
       but an expression was expected of type al *)
(* without -short-paths: This expression has type int but an expression was expected of type
         al = bool *)
~~~
I don't know why ocaml people think -short-paths gives better error but for the current Coq codebase it is certainly wrong.

ejgallego:
> I don't know why ocaml people think -short-paths gives better error but for the current Coq codebase it is certainly wrong.

The reason is module aliases, some important libraries such as `core` make heavy use of that feature and it seems they are going to play an important role in the future. For Coq this is no such a big problem yet, but for example we could have problems with plugins as they are packed this way.

Anyways let's go for it and see what we get.

PR #156, opened 2016-04-12
Numeral notation

roglo:
Added a new vernac command "Numeral Notation". It allows to add parsing and printing strings of decimal digits (such as "0", "-47", "223") from and to any datatype. It has been implemented in the Coq library for Nat, Positive, N, Z, R, int31, bigN, bigZ, bigQ. The OCaml versions in plugins/syntax have been removed. See the examples of usage by typing
   grep "Numeral Notation" theories/*/*.v theories/*/*/*/*.v
and look at the contents of these files.

A type named Z' have been added in Datatypes.v to deal with strings of digits. It is defined exactly like Z and positive, but with quotes. It could be Z if you decide to define Z here.

Syntax:
      Numeral Notation t f g : t_scope {(printing ...)} {(warning after w)}
- t is the datatype
- f is the parsing function with must be of type Z' -> option t, None means not recognized, like in the example "Check (-1)%nat.".
- g is for printing and is either a function of type t -> option Z' or a Ltac, depending on the datatype t and the printing option (see below)
- t_scope is the name of the scope of t
- the optional (printing ...) tells to consider the list ... of constructors or axioms to print. See examples in Rdefinitions.v. In that case, g must be a Ltac. If absent, t is supposed to be an inductive type and the printed list are its constructors. In that case, g must be a function.
- the optional (warning after w) where w is an integer tells to print a warning when using a too big number. See example in Nat.v. If you type e.g. Check 5001, you get the message.

Implementation: the main stuff is in tacinterp.ml. I previously put it in vernacentries.ml but in the present trunk, this would create a circular dependency, the Numeral Notation possibly using the execution of tactics. This code of about 300 lines could be in a separate file, named for example "numeral.ml", thing like that.

Small thing: in this file (tacinterp.ml), I added a function named "eval_tacexpr" taking a tacexpr and returning a value option, to allow to call a Ltac and taking its result, like a function. Perhaps it could be useful for other Coq developers.

PR #9678, opened 2019-03-01
Stop accessing proof env via Pfedit in printers

maximedenes:
This should make https://github.com/coq/coq/pull/9129 easier.

This is still work in progress, I'll need some input to complete it.


plugins/ssrmatching/ssrmatching.ml
```diff
@@ -725,13 +724,13 @@ let rec uniquize = function
       0, uniquize (instances ())
     | NoMatch when (not raise_NoMatch) ->
       if !failed_because_of_TC then
-        errorstrm (source ()++strbrk"matches but type classes inference fails")
+        errorstrm (source env (Evd.from_env env) (* XXX *) ++strbrk"matches but type classes inference fails")
```
maximedenes: @gares which `sigma` should I use here?

plugins/ssrmatching/ssrmatching.ml
```diff
@@ -725,13 +724,13 @@ let rec uniquize = function
...
       else
-        errorstrm (source () ++ str "does not match any subterm of the goal")
+        errorstrm (source env (Evd.from_env env) (* XXX *) ++ str "does not match any subterm of the goal")
```
maximedenes: @gares which `sigma` should I use here?

plugins/ssrmatching/ssrmatching.ml
```diff
@@ -725,13 +724,13 @@ let rec uniquize = function
...
     | NoProgress when (not raise_NoMatch) ->
         let dir = match upats_origin with Some (d,_) -> d | _ ->
           CErrors.anomaly (str"mk_tpattern_matcher with no upats_origin.") in
-        errorstrm (str"all matches of "++source()++
+        errorstrm (str"all matches of "++source env (Evd.from_env env) (* XXX *) ++
```
maximedenes: @gares which `sigma` should I use here?

plugins/ssrmatching/ssrmatching.ml
```diff
@@ -773,10 +772,10 @@ let rec uniquize = function
   if max_occ <= !nocc then p', u.up_dir, (sigma, uc, u.up_t)
   else errorstrm (str"Only " ++ int !nocc ++ str" < " ++ int max_occ ++
         str(String.plural !nocc " occurrence") ++ match upats_origin with
-        | None -> str" of" ++ spc() ++ pr_constr_pat p'
+        | None -> str" of" ++ spc() ++ pr_constr_pat (Global.env () (* XXX *)) sigma p'
```
maximedenes: @gares which `env` should I use here and below?

plugins/syntax/numeral.ml
```diff
@@ -78,7 +78,7 @@ let locate_int63 () =
   else None
 
 let has_type f ty =
-  let (sigma, env) = Pfedit.get_current_context () in
+  let (sigma, env) = Pfedit.get_current_context () in (* XXX *)
```
maximedenes: @JasonGross why are we using the proof-local environment here? Is this correct?

plugins/syntax/string_notation.ml
```diff
@@ -33,7 +33,7 @@ let q_list () = qualid_of_ref "core.list.type"
 let q_byte () = qualid_of_ref "core.byte.type"
 
 let has_type f ty =
-  let (sigma, env) = Pfedit.get_current_context () in
+  let (sigma, env) = Pfedit.get_current_context () in (* XXX *)
```
maximedenes: @JasonGross same question as for numeral, plus why do we duplicate the code here?

plugins/funind/recdef.ml
```diff
@@ -1417,7 +1419,7 @@ let com_terminate
     nb_args ctx
     hook =
   let start_proof ctx (tac_start:tactic) (tac_end:tactic) =
-    let evd, env = Pfedit.get_current_context () in
+    let evd, env = Pfedit.get_current_context () in (* XXX *)
```
maximedenes: @Matafou @forestjulien I'm not sure I follow the logic here: we pass the current proof state to `start_proof`? Can you explain a bit what this code is doing? Thanks!

plugins/ltac/extraargs.mlg
```diff
@@ -148,7 +148,7 @@ let pr_occurrences = pr_occurrences () () ()
 let pr_gen prc _prlc _prtac c = prc c
 
 let pr_globc _prc _prlc _prtac (_,glob) =
-  let _, env = Pfedit.get_current_context () in
+  let _, env = Pfedit.get_current_context () in (* XXX *)
```
maximedenes: @ppedrot What is the `glob` arg type used for?

maximedenes: Let me make my question more precise: this argument extension seems rarely used, although we have a lot of stuff dealing with `glob_constr` (e.g. ssr), so I was wondering: are there many ways to take glob constrs in the ML interface for tactics?

plugins/ltac/g_auto.mlg
```diff
@@ -64,10 +64,10 @@ let eval_uconstrs ist cs =
 
 let pr_auto_using_raw _ _ _  = Pptactic.pr_auto_using Ppconstr.pr_constr_expr
 let pr_auto_using_glob _ _ _ = Pptactic.pr_auto_using (fun (c,_) ->
-    let _, env = Pfedit.get_current_context () in
+    let _, env = Pfedit.get_current_context () in (* XXX *)
```
maximedenes: @ppedrot are we sure we want the proof-local env here? In particular, won't the same term print differently depending on whether we take it as a `raw` or as a `glob` (since the printer for `raw` does not take the `env`)?

plugins/ltac/g_auto.mlg
```diff
@@ -64,10 +64,10 @@ let eval_uconstrs ist cs =
...
     Printer.pr_glob_constr_env env c)
 let pr_auto_using _ _ _ = Pptactic.pr_auto_using
-    (let sigma, env = Pfedit.get_current_context () in
+    (let sigma, env = Pfedit.get_current_context () in (* XXX *)
```
maximedenes: AFAICT this is the only `PRINTED BY` stuff that needs the evar map. @ppedrot is there something special about it? Does it mean the evarmap should be threaded everywhere, just for this one?

plugins/ltac/pptactic.ml
```diff
@@ -1300,11 +1300,11 @@ let register_basic_print0 wit f g h =
 
 
 let pr_glob_constr_pptac c =
-  let _, env = Pfedit.get_current_context () in
+  let _, env = Pfedit.get_current_context () in (* XXX *)
```
maximedenes: Here again I'm not sure I understand the logic. Isn't the glob_constr printer going to do things differently (in terms of disambiguating names) from the raw printer? Is this really desired? @ppedrot

plugins/ssrmatching/ssrmatching.ml
gares: I guess `ise` but I don't think source should be abstrcted, since the term it prints is inside its closure (so it is always the same, no point in printing it under a different context). (Eg, In the first case of `source` the term is printed using `ise`)

plugins/ssrmatching/ssrmatching.ml
gares: uff, you don't have it at hand. It is the env the companion closure gets in input. You could make this function take an env rather than a unit, but the the user would need to pass the same env to both (that is the proof env).

I'm tempted to say that you could change the find_P type, make a pair adding an env, return it in the first closure, and bind it in the second one.

The old code was using the global one, that seems a but to me.

plugins/ssrmatching/ssrmatching.ml
gares: (source is just a thunk to not compute the error message if we don't need to)

plugins/ltac/extraargs.mlg
ppedrot: That's probably an attempt at a poor man's closure, as Ltac doesn't have a first-class support for value-returning functions. I don't really understand why you would need a glob at runtime though. It only seem to be used in rewrite strategies?

ppedrot: Ah, no, it's also used by `instantiate`, and from there by `w_refine`. So I guess that you have your answer, that's legacy stuff.

plugins/ltac/g_auto.mlg
ppedrot: That's only used in a few tactics, so you'll most likely never use the raw printer. `Print Ltac` uses the globalized one, raw printing would only be used by the beautifier.

plugins/ltac/g_auto.mlg
ppedrot: Most likely. Virtually all runtime values depend on the current evar map (think about terms for instance).

plugins/ltac/pptactic.ml
ppedrot: No idea, but see my comment above.

plugins/ltac/g_auto.mlg
ejgallego: Anyways a lot of entries are missing a `PRINTED BY` declaration at all so I wonder how much these printers are used as for the missing printers we get a dummy value.

plugins/ltac/pptactic.ml
ejgallego: Indeed you are asking a deep question on the design of printing; the current situation indeed has this problem, but it must be said that printing from elaborated terms is a quite hard problem.

None of the solutions I know are easy, what I do in some other compilers is to actually resolve names to some extent at parsing time, but yeah, this is harder here with proofs....

plugins/ssr/ssrcommon.ml
```diff
@@ -1223,7 +1223,7 @@ let abs_wgen keep_let f gen (gl,args,c) =
   let evar_closed t p =
     if occur_existential sigma t then
       CErrors.user_err ?loc:(loc_of_cpattern p) ~hdr:"ssreflect"
-        (pr_constr_pat (EConstr.Unsafe.to_constr t) ++
+        (pr_constr_pat env sigma (EConstr.Unsafe.to_constr t) ++
```
ejgallego: Should `pr_constr_pat` be switched to `EConstr`?

ejgallego:
Hard and non very rewarding work @maximedenes ; but indeed that should solve quite a few "bugs" or a least clarify some semantics of printing. It is still tricky to know what to do, would be great to have some more typing discipline.

I guess tho that this kinda settles the question of #9129 vs printing.

I can of course rebase that PR on this one, but we could also merge that one and continue work here, the  situation in practice wouldn't change a lot given the current state of the codebase. WDYT?

JasonGross:
> Anyways a lot of entries are missing a PRINTED BY declaration at all so I
wonder how much these printers are used as for the missing printers we get
a dummy value.

Is this related to Print Ltac failing to Print tactics in terms (instead
printing an underscore) and a couple other constructions?

ppedrot:
> Is this related to Print Ltac failing to Print tactics in terms (instead printing an underscore) and a couple other constructions?

No, Print Ltac is using the glob printer, the reason it fails on tactics-in-terms is rather due to the fact that we don't know how to externalize (glob -> raw) tactics.

plugins/ltac/g_auto.mlg
ppedrot: `PRINTED BY` is almost never used in practice, except for a few base types that have a harcoded variant like terms. For it to be called, you'd need to use a tactic notation with the corresponding argument and try e.g. calling `idtac` on the bound variable. If you only use it for ML tactics, you're most likely going to pass it directly to some ML code, which means that you won't ever put it inside a Ltac environment that would be used for debug printing.

ppedrot: And also, thinking about it, for this to be actually used, you need not to have the `TYPED AS` clause, otherwise the runtime representation of that generic argument will be the same at the one given in that clause, resulting in a call to that printer instead. I am not even sure we have an instance of that in the Coq code.

(Note that `PRINTED BY` would be used nonetheless if there is no `GLOB/RAW PRINTED BY` clause, in which case the argument printer is used for all three levels.)

plugins/ssrmatching/ssrmatching.ml
maximedenes: > The old code was using the global one, that seems a but to me.

@gares Does it mean my patch preserves the former semantics here? That's the main goal I'm trying to achieve.

gares:
@maximedenes this commit should replace your changes to ssrmatching
https://github.com/gares/coq/commit/159558c2ab554653c74c5379fe5ea5e667b70ac6

plugins/syntax/numeral.ml
maximedenes: @JasonGross ping

plugins/ltac/g_auto.mlg
maximedenes: So, what should I do here, in your opinion @ppedrot ?

plugins/ssr/ssrcommon.ml
maximedenes: Yes, but that's orthogonal to this PR.

plugins/syntax/numeral.ml
JasonGross: The original numeral notations PR (#156) by @roglo uses `Lemmas.get_current_context ()`, and the revised one (#496) by @letouzey uses `Pfedit.get_current_context ()`.  I have not modified this bit of code, only moved it from one file to another, and don't know the reasoning behind it.  I believe this function is only called when invoking the `Numeral Notation` vernacular / when adding new numeral notations.  I think it would be fine to use the global env, or to get the env from the vernacular command and thread it through to this function call.

plugins/syntax/string_notation.ml
JasonGross: Same answer as for numeral notation, and I duplicated this code because I didn't see a good place to export the common code, and didn't want to rename `numeral.ml `  to `numeral_and_string_notation.ml `.  Feel free to factor the common bits if you have a good place for them.

maximedenes:
Ok, we're getting there. @ppedrot would you mind having a look?

ppedrot (commented):
Haven't seen anything suspicious, this is mostly what I would've written myself.

plugins/funind/recdef.ml
maximedenes: I'll leave this part as it is.

maximedenes:
@ppedrot can you be the assignee? I'm going to prepare overlays.

Matafou:
Sorry Maxime I am not the author of this part, I am not sure of the
consequence. I think @jforest needs some strange hacks at some point to
define things in the right order, but I don't know if this is part of those
hacks.
P.

Le jeu. 14 mars 2019 à 20:28, Maxime Dénès <notifications@github.com> a
écrit :

plugins/funind/g_indfun.mlg
```diff
@@ -47,15 +47,15 @@ let pr_fun_ind_using_typed prc prlc _ opt_c =
...
 
 }
 
 ARGUMENT EXTEND fun_ind_using
   TYPED AS constr_with_bindings option
   PRINTED BY { pr_fun_ind_using_typed }
-  RAW_PRINTED BY { pr_fun_ind_using }
-  GLOB_PRINTED BY { pr_fun_ind_using }
+  RAW_PRINTED BY { pr_fun_ind_using env sigma }
+  GLOB_PRINTED BY { pr_fun_ind_using env sigma }
```
ejgallego: I am not sure I am very fond of this kind of hidden binding; I'd rather have the form take a function.

ejgallego: cc @ppedrot

maximedenes: This is for compatibility. With the current approach, only term-related printers need to be changed.

ejgallego: Indeed, OK with that then.

maximedenes:
I rebased the overlays (previous errors were due to the `SProp` merge), hopefully it should be ok now.

ejgallego:
CI is OK, failure is the usual async test-suite failure.

ejgallego:
@ppedrot I'd say let's merge; each time I look at the use of `get_current_context` I discover something really fishy [like for example plugins modifying the global environment and then relying on such function returning it as there is no proof open]

So indeed, improving the use of env in Coq is to be a long term project.

I've got almost 20 PRs depending on this one and some of them include what I regard as important bugfixes, and the branching point is just 2 weeks ahead.

maximedenes:
Hum, there's now a conflict, I'll rebase.

ppedrot:
I'll merge this PR after the rebase and CI passes.

maximedenes:
Rebased.

maximedenes:
I don't understand the windows64 failure. Where's the error?

maximedenes:
@ppedrot can you merge? Before rebase, the windows build did succeed, so I assume the current failure should be unrelated.

ppedrot:
This error happens from time to time, it is unrelated to your PR. I'll merge it right away.

ejgallego:
> I don't understand the windows64 failure. Where's the error?

Yeah logs do remain unreadable; it was a network error to sourceforge, the build in itself was fine.

maximedenes:
Thanks!

plugins/syntax/g_string.mlg
```diff
@@ -21,5 +21,6 @@ open Stdarg
 VERNAC COMMAND EXTEND StringNotation CLASSIFIED AS SIDEFF
   | #[ locality = Attributes.locality; ] [ "String" "Notation" reference(ty) reference(f) reference(g) ":"
       ident(sc) ] ->
-    { vernac_string_notation (Locality.make_module_locality locality) ty f g (Id.to_string sc) }
+    { let (sigma, env) = Pfedit.get_current_context () in
```
ejgallego: @maximedenes this is a good example of a dubious use of `get_current_context`, this function needs the global environment, and here it only happens to work because get_current_context will never be called inside a proof.

maximedenes: Yes, that's what I thought too, but I tried to limit this PR to making the threading explicit, without changing too much the existing code.

We should however keep track of the problematic occurrences, maybe.

ejgallego: My point of view here is that is safer [and easier to track] to make such functions not to take an env sigma, as they are not supposed to take it, and have them just use the global environment.

Having `env -> evar_map` is extremely confusing as it gives the user the wrong impression.

printing/ppconstr.mli
```diff
@@ -41,19 +41,19 @@ val pr_guard_annot : (constr_expr -> Pp.t) ->
...
 val pr_record_body : (qualid * constr_expr) list -> Pp.t
-val pr_binders : local_binder_expr list -> Pp.t
-val pr_constr_pattern_expr : constr_pattern_expr -> Pp.t
-val pr_lconstr_pattern_expr : constr_pattern_expr -> Pp.t
-val pr_constr_expr : constr_expr -> Pp.t
-val pr_lconstr_expr : constr_expr -> Pp.t
+val pr_binders : Environ.env -> Evd.evar_map -> local_binder_expr list -> Pp.t
+val pr_constr_pattern_expr : Environ.env -> Evd.evar_map -> constr_pattern_expr -> Pp.t
+val pr_lconstr_pattern_expr : Environ.env -> Evd.evar_map -> constr_pattern_expr -> Pp.t
+val pr_constr_expr : Environ.env -> Evd.evar_map -> constr_expr -> Pp.t
```
ejgallego: We shouldn't have changed this type, but instead we add a new, beautifier specific printer that does the round trip [even better would be to do the roundtrip at the AST level]

PR #9685, opened 2019-03-02
[vernac] Small cleanup to remove assert false.

ejgallego:
This is a fairly small cleanup on the `vernac_interp` function, which
makes code cleaner and for example would allow to have `Load` inside
`Load`. [Not that we would ever want that]

### DIFF
--- a/vernac/vernacentries.ml
+++ b/vernac/vernacentries.ml
@@ -2152,46 +2152,7 @@ let vernac_check_guard ~pstate =
       (str ("Condition violated: ") ++s)
   in message
 
-exception End_of_input
-
-(* XXX: This won't properly set the proof mode, as of today, it is
-   controlled by the STM. Thus, we would need access information from
-   the classifier. The proper fix is to move it to the STM, however,
-   the way the proof mode is set there makes the task non trivial
-   without a considerable amount of refactoring.
- *)
-let vernac_load ~st interp fname =
-  let pstate = st.Vernacstate.proof in
-  if there_are_pending_proofs ~pstate then
-    CErrors.user_err Pp.(str "Load is not supported inside proofs.");
-  let parse_sentence proof_mode = Flags.with_option Flags.we_are_parsing
-    (fun po ->
-    match Pcoq.Entry.parse (Pvernac.main_entry proof_mode) po with
-      | Some x -> x
-      | None -> raise End_of_input) in
-  let fname =
-    Envars.expand_path_macros ~warn:(fun x -> Feedback.msg_warning (str x)) fname in
-  let fname = CUnix.make_suffix fname ".v" in
-  let input =
-    let longfname = Loadpath.locate_file fname in
-    let in_chan = open_utf8_file_in longfname in
-    Pcoq.Parsable.make ~loc:(Loc.initial (Loc.InFile longfname)) (Stream.of_channel in_chan) in
-  let rec load_loop ~pstate =
-    try
-      let proof_mode = Option.map (fun _ -> get_default_proof_mode ()) pstate in
-      let pstate = interp ~st:{ st with Vernacstate.proof = pstate }
-          (parse_sentence proof_mode input).CAst.v in
-      load_loop ~pstate
-    with
-      End_of_input ->
-      pstate
-  in
-  let pstate = load_loop ~pstate in
-  (* If Load left a proof open, we fail too. *)
-  if there_are_pending_proofs ~pstate then
-    CErrors.user_err Pp.(str "Files processed by Load cannot leave open proofs.");
-  pstate
-
+(* Attributes *)
 let with_locality ~atts f =
   let local = Attributes.(parse locality atts) in
   f ~local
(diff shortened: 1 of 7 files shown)


ejgallego:
Sorry for the review spam, I submitted this originally on top of master, but I realized that the rebase would be very bad so I had to rebase on top of a dependency but github didn't allow me to change this to draft status.

ejgallego:
Ready for review / assignee.

ejgallego:
IMO worth merging in 8.10 as to easy backports / rebases; changes are pretty low risk I think.

vernac/vernacexpr.ml
```diff
@@ -405,11 +405,6 @@ type vernac_control =
      the flag is used to print differently in `-time` vs `Time foo` *)
   | VernacTime of bool * vernac_control CAst.t
   | VernacRedirect of string * vernac_control CAst.t
-  | VernacTimeout of int * vernac_control
-  | VernacFail of vernac_control
+  | VernacTimeout of int * vernac_control CAst.t
+  | VernacFail of vernac_control CAst.t
```
SkySkimmer: The linter wants you to remove this empty line.

vernac/vernacentries.ml
```diff
@@ -2557,121 +2624,17 @@ let () =
...
-
 (* Be careful with the cache here in case of an exception. *)
-let interp ?verbosely ?proof ~st cmd =
+let interp ?(verbosely=true) ?proof ~st cmd =
   Vernacstate.unfreeze_interp_state st;
-  try
-    let pstate = interp ?verbosely ?proof ~st cmd in
-    Vernacstate.Proof_global.set pstate;
-    Vernacstate.freeze_interp_state ~marshallable:false
+  try vernac_timeout ?timeout:!default_timeout (fun st ->
```
SkySkimmer: If we have 
~~~coq
Set Default Timeout 1.
Timeout 3 Check thing_taking_2s.
~~~
it seems like it won't work properly.

vernac/vernacentries.ml
```diff
@@ -2238,6 +2244,11 @@ let interp ?proof ~atts ~st c : Proof_global.t option =
   (* This one is possible to handle here *)
   | VernacAbort id    -> CErrors.user_err  (str "Abort cannot be used through the Load command")
 
+  (* Loading a file requires access to the control interpreter *)
+  | VernacLoad (verbosely,fname) ->
```
SkySkimmer: The comment was to explain the assert false so isn't relevant anymore.

vernac/vernacexpr.ml
ejgallego: Thanks!

vernac/vernacentries.ml
ejgallego: thanks

vernac/vernacentries.ml
ejgallego: Indeed that's a change of semantics, well spotted. That being said, doesn't the new behavior make more sense?

ejgallego: Like what should happen in `Timeout 1 (Timeout 2 foo))` ?

SkySkimmer: `Error: multiple timeouts not allowed`
I'm concerned about the specific one I posted with Default Timeout, where it makes sense for a local Timemout to override it.

ejgallego: Could make sense otherwise [the global taking priority over the local] but I'll try to restore behavior.

SkySkimmer: >the global taking priority over the local

Not with it being called Default imo

ejgallego: Ok I tried to amend the change.

SkySkimmer: >Ok I tried to amend the change.

Did you test something like the example I posted? USe something like
~~~coq
Set Default Timeout 1.

Lemma foo : True.
  Timeout 2 do 2000 pose I.
~~~

ejgallego: Yup it seems to work.

SkySkimmer: Cool.

vernac/vernacentries.ml
```diff
@@ -2211,18 +2172,69 @@ let with_def_attributes ~atts f =
...
+   caller's responsibility *)
+let with_fail ~st f =
+  try
+    (* If the command actually works, ignore its effects on the state.
+       * Note that error has to be printed in the right state, hence
+       * within the purified function *)
+    try let _ = f () in raise HasNotFailed
+    with
+    | HasNotFailed as e -> raise e
+    | e ->
```
SkySkimmer: How about `e when not e = HasNotFailed`?

vernac/vernacentries.ml
```diff
@@ -2211,18 +2172,69 @@ let with_def_attributes ~atts f =
...
+    f x
+
+(* Fail *)
+exception HasNotFailed
+exception HasFailed of Pp.t
+
+let test_mode = ref false
+
+(* XXX STATE: this type hints that restoring the state should be the
+   caller's responsibility *)
```
SkySkimmer: I don't understand this comment.

ejgallego: The comment was here before; I am not sure what it means anymore tho; see #9855 however that one is work in progress.

vernac/vernacentries.ml
ejgallego: I am trying to address this in #9855 however there is some tricky stuff.

vernac/vernacentries.ml
ejgallego: I'd say let's leave the comment as it was for this PR.

vernac/vernacentries.ml
ejgallego: I'd say let's leave the code of fail intact in this PR.

ejgallego:
Rebased, solved conflict.

ejgallego:
This could get an assignee maybe? I guess @maximedenes , @vbgl , or @SkySkimmer may be among the ones more familiar with this code here.

vbgl (approved):

ejgallego:
Thanks @vbgl

PR #9686, opened 2019-03-03
[error] Remove special error printing pre-processing

### LINKED ISSUE #5360: Lots of Anomaly reports during setoid_rewrite

Original bug ID: BZ#5360
From: @vzaliva
Reported version: 8.6
CC:   @ejgallego, frederic.loulergue@nau.edu, @JasonGross, @jashug, @psteckler, @vzaliva

ejgallego:
We remove the special error printing pre-processing in favor of just
calling the standard printers.

Error printing has been a bit complex for a while due to an incomplete
migration to a new printing scheme based on registering exception
printers; this PR should alleviate that by completing the registration
approach.

After this cleanup, it should not be ever necessary for normal
functions to worry a lot about catching errors and re-raising them,
unless they have some very special needs.

Should fix #5360


ejgallego:
This could alter async proof printing in some way, hard to tell as it seems state in not closed properly anyways [for example hooks seem to be ran in the wrong state, that could explain some of the async failure].

ejgallego:
This is almost ready, modulo the dependencies and a tweak needed for the "ltac" printing decorators which should register in a different way.

I am pretty confident that we can do away with that "explain" function for good.

ejgallego:
Ok, this should be ready for review. @ppedrot I'm not sure I understand the previous LTAC2 code raising `Exit`, can you comment on that?

gares:
This pr needs discussion. In particular I believe the mechanism bein deprecated is used to attach a locto errors (not only an stm related stated I'd). Also, errors were algebraically enriched before this mechanism was introduced and it was no a panacea. Hence the need for a wg discussion.

ejgallego:
Indeed the wording I used was misleading, @gares is the current one OK for you?

This PR doesn't deprecate `Exninfo` use, and it should not do it as the change is largely unrelated; I'll open its own issue for it.

gares:
I'm not so sure it is true either, @ppedrot knows better, but I believe re-raising is safe.
What is not "safe" is `... with e -> raise e'` that indeed looses all the info attached to `e`. If you re-raise `e` things should be fine.

ejgallego:
> I'm not so sure it is true either, @ppedrot knows better, but I believe re-raising is safe.
> What is not "safe" is `... with e -> raise e'` that indeed looses all the info attached to `e`. If you re-raise `e` things should be fine.

I think that depends on what primitive the compiler chooses, as of today that's a heuristic so indeed it can happen that `with e -> .....; raise e` doesn't use `%reraise`.

What I am pretty positive is that IMHO `Exninfo` is not a safe pattern for any non-debugging data; locations and backtraces are OK as nothing breaks if you lose them.

I'll be happy tho @gares to remove the comment or reword as you think, we can discuss this in its own issue.

gares:
I meant re-raising as in "raising the same exception without allocating a new one". IIRC the exninfo thing is implemented by linking the extra data to a pointer. So
```ocaml
try raise (Foo 1)
with e -> whatever; raise e
```
is OK while
```ocaml
try raise (Foo 1)
with e -> whatever; raise (Foo 1)
```
is not. But I wish @ppedrot to confirm.

For sure
```ocaml
try raise (Foo 1)
with e -> raise (Bar 1)
```
drops all the exninfo, and we have seen some code breaking the stm assumptions by doing this in the past.

ejgallego:
> is OK while

It may not OK, depending on what lower-level primitive the compiler chooses.

ejgallego:
The discussion diverged a bit to a topic fully orthogonal to this PR. IMO the PR itself is ready so an assignee would be appreciated.

I can remove the comment in exninfo if you want.

gares:
Yes please remove it.

ejgallego:
> Yes please remove it.

Done!

ejgallego:
IMHO this should be ready, thus an assignee would be great, any of @gares @ppedrot @SkySkimmer may be a good fit.

gares:
Next week I'm all busy, not for me

SkySkimmer (approved):
Looks OK

PR #9688, opened 2019-03-04
[stm] unfocus when edition exits the proof (fix #9431)

### LINKED ISSUE #9431: Asynchronous proofs crash proof workers in CoqIDE
#### Version

Coq 8.8.2

#### Operating system

Red Hat Enterprise Linux Workstation 7.4

The following actions crash proof workers.

* Open CoqIDE.
* Insert the following code into the buffer.
* Navigate to the end.
* Navigate to any point between `Proof` and `Qed` of the first goal.
* Navigate to any point between `Goal` and `Proof` of the first goal.
* Navigate to the end.

```
Goal False.
Proof. try constructor. Qed.

Goal True.
Proof. try constructor. Qed.
```

```
Coqtop died badly.
```

There are two ways to recover.
The more convenient one is to be more careful.

* After navigating to any point between `Goal` and `Proof` of the first goal...
* Before navigating to the end...
* Navigate to any point before `Goal` of the first goal.
* Navigate to the end.

The less convenient one is to disable asynchronous proofs altogether.

* Open the *Tools* menu.
* Open the *Coqtop arguments* dialog.
* Enter `-async-proofs off` into the text field.
* Push the `OK` button.
* Navigate to the end.

gares:
Closes #9431

### DIFF
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -3205,10 +3205,9 @@ let edit_at ~doc id =
   let vcs = VCS.backup () in
   let on_cur_branch id =
     let rec aux cur =
-      if id = cur then true
-      else match VCS.visit cur with
+      match VCS.visit cur with
       | { step = `Fork _ } -> false
-      | { next } -> aux next in
+      | { next } -> if id = cur then true else aux next in
     aux (VCS.get_branch_pos (VCS.current_branch ())) in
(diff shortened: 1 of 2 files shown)


ejgallego (approved):
Makes sense.

PR #9689, opened 2019-03-04
[default.nix] Use the dune-based build-system

vbgl:
The installation of the user manual does nothing; but that shall be fixed independently (see https://github.com/ocaml/dune/issues/1868).

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -35,6 +35,11 @@
 with pkgs;
 with stdenv.lib;
 
+let packages = [ "coq" "coqide-server" ]
+++ optional buildIde "coqide"
+++ optional buildDoc "coq-refman"
+; in
+
 stdenv.mkDerivation rec {
 
   name = "coq";
@@ -78,31 +83,46 @@ stdenv.mkDerivation rec {
            !elem (baseNameOf path) [".git" "result" "bin" "_build" "_build_ci" "nix"]) ./.;
 
   preConfigure = ''
+    patchShebangs tools/
     patchShebangs dev/tools/
+    configureFlagsArray+=("-libdir" "$OCAMLFIND_DESTDIR"coq "-native-compiler" "no")
+  '' + optionalString doInstallCheck ''
+    patchShebangs test-suite/
+  '' + optionalString buildDoc ''
+    patchShebangs doc/tools/
   '';
 
   prefixKey = "-prefix ";
 
-  enableParallelBuilding = true;
-
-  buildFlags = [ "world" "byte" ] ++ optional buildDoc "doc-html";
-
-  installTargets =
-    [ "install" "install-byte" ] ++ optional buildDoc "install-doc-html";
+  makefile = "Makefile.dune";
 
   createFindlibDestdir = !shell;
 
-  postInstall = "ln -s $out/lib/coq $OCAMLFIND_DESTDIR/coq";
+  buildFlags = [ "voboot" ];
+
+  postBuild = ''
+    for package in ${toString packages}
+    do
+      dune build $package.install
+    done
+  '';
+
+  installPhase = ''
+    runHook preInstall
+    for package in ${toString packages}
+    do
+      dune install --prefix=$out --libdir=$OCAMLFIND_DESTDIR $package
+    done
+    runHook postInstall
+  '';
 
   inherit doInstallCheck;
 
   preInstallCheck = ''
-    patchShebangs tools/
-    patchShebangs test-suite/
     export OCAMLPATH=$OCAMLFIND_DESTDIR:$OCAMLPATH
   '';
 
-  installCheckTarget = [ "check" ];
+  installCheckTarget = [ "test-suite" ];
 
   passthru = {
     inherit coq-version ocamlPackages;
--- a/test-suite/dune
+++ b/test-suite/dune
@@ -17,6 +17,7 @@
    ../Makefile.common
    ../config/Makefile
    ; Stuff for the compat script test
+   ../configure.ml
    ../dev/header.ml
    ../dev/tools/update-compat.py
    ../doc/stdlib/index-list.html.template


default.nix
```diff
@@ -78,29 +83,50 @@ stdenv.mkDerivation rec {
...
+  makefile = "Makefile.dune";
 
-  createFindlibDestdir = !shell;
+  buildFlags = [ "voboot" ];
 
-  postInstall = "ln -s $out/lib/coq $OCAMLFIND_DESTDIR/coq";
+  postBuild = ''
+    for package in ${toString packages}
+    do
+      dune build $package.install
```
ejgallego: If you are building for the final user you should instead use `dune build -p package` which enables the "release" profile. I am not sure how Nix handles Dune packages [coq will soon be 100% standard and don't require `voboot`], but you may want to have a flag for the build to happen in release or in dev mode

See https://dune.readthedocs.io/en/latest/usage.html?highlight=profile#invocation-from-opam for more details. In particular, `-p` requires that the depending packages are system-installed, so indeed that's a tricky tradeoff.

default.nix
```diff
@@ -78,29 +83,50 @@ stdenv.mkDerivation rec {
...
+    done
+  '';
+
+  installPhase = ''
+    runHook preInstall
+    for package in ${toString packages}
+    do
+      dune install --prefix=$out $package
+    done
+    ln -s $out/lib/coq $OCAMLFIND_DESTDIR/coq
```
ejgallego: Why is this still needed? Anything we could do so we get the right thing done?

default.nix
```diff
@@ -39,6 +39,11 @@
 with pkgs;
 with stdenv.lib;
 
+let packages = [ "coq" ]
+++ optionals buildIde [ "coqide" "coqide-server" ]
```
ejgallego: You can always build `coqide-server` if you want, it doesn't include any gui library, only a different toplevel.

default.nix
vbgl: You need to distinguish (in `dune-install`) the path where OCaml libraries are installed from the path where Coq libraries are installed.

ejgallego: I thought we had fixed this. Currently we do:
```
lib_root: [
  "_build/install/default/lib/coq/plugins/btauto/.Algebra.aux" {"coq/plugins/btauto/.Algebra.aux"}
  "_build/install/default/lib/coq/plugins/btauto/.Btauto.aux" {"coq/plugins/btauto/.Btauto.aux"}
  "_build/install/default/lib/coq/plugins/btauto/.Reflect.aux" {"coq/plugins/btauto/.Reflect.aux"}
  "_build/install/default/lib/coq/plugins/btauto/Algebra.glob" {"coq/plugins/btauto/Algebra.glob"}
  "_build/install/default/lib/coq/plugins/btauto/Algebra.vo" {"coq/plugins/btauto/Algebra.vo"}
```
how is this not the right place? OCaml libraries go in `lib` which can be mapped to `lib/ocaml` by distributions for example.

ejgallego: IMHO is very important we get this right as soon Dune itself will generate this rules and would be a pain to have to change them. In fact, the Dune version will place this stuff in `lib_root/coq/stdlib/...` as the lib will be named stdlib.

default.nix
vbgl: This means that the nix build should be split: one per package. Otherwise, we cannot build say `coqide-server` until `coq` is installed.

default.nix
vbgl: Can you please elaborate on how distributions can map these different lib directories to what they need?

default.nix
ejgallego: That's what OPAM does, if `coqide` depends on `coqide-server` it will build an install that first.

Note that Nix doesn't have to do that, the composed build is still sound, however you have to be careful to install all the stuff you've build [-p gives you this sanity check as it ignores the rest of packages in the tree].

What would be the best Nix model? I'm sure it would be nice to add native Nix workflow to dune.

default.nix
ejgallego: I am not sure what distros need so I cannot elaborate a lot. In this case, you can use the `--libdir` and `--prefix` parameter to relocate `lib`, so you mean that `lib_root` should be not affected by `--libdir`?

What should the right layout be on Nix?

ejgallego: Of course `dune-install` can be extended to accommodate Nix's needs.

vbgl: The issue here is that there are two `lib` directories. One for Coq, one for OCaml. I’m OK using `--libdir` to set one of them. What shall I do for the other one?

Currently, the layout used in nixpkgs is as follows (for non-standard libraries):
 - For OCaml: `$prefix/lib/ocaml/$ocaml-version/site-lib/`
 - For Coq: `$prefix/lib/coq/$coq-version/user-contrib`

ejgallego: Yup I think this layout is not supported by dune-install; `--libdir` will move both library paths. We ought to add a new option to dune-install so we can properly remap `lib_root` at least.

I'll open an issue.

ejgallego: > * For Coq: `$prefix/lib/coq/$coq-version/user-contrib`

The layout I propose for 8.11 [most likely is]:

`$prefix/lib/coq/$coq-version/$lib_name/`

does that sound good to you?

ejgallego: In coq dune there will be no library loaded by default, users of the stdlib will have to declare it in the deps.

vbgl: Does this means getting rid of `COQPATH`?

default.nix
Zimmi48: I would be very much in favor of splitting Coq in as many Nix packages as there are opam ones. Ideally, we would even split the source tree in a way that avoids rebuilding the main Coq package if only the doc changes.

default.nix
ejgallego: Both should coexist I think, as for example to keep the NIX use case, but a single nix package could install many coq packages in its private `lib/coq/` dir.

ejgallego: Note that there is something attractive on installing coq packages under just `lib`, and this is that plugins get to be more self-contained.

ejgallego:
c.f https://github.com/ocaml/dune/issues/1901

default.nix
ejgallego: Note that I think that for the CI build, we should build in the way this PR builds now. However for the Nix user, we may want to enable the optimizing profile.

ejgallego:
Umm, the log seems quite different from what we get on the regular makefile build, for example all these
```
wrong ELF type
```
are gone. What does this mean?

vbgl:
In means that native compute is disabled.

ejgallego:
> In means that native compute is disabled.

What's the problem with native compute, aren't these regular `.cmx` files?

default.nix
ejgallego: Note @zimmi48 that one big downside of the per-package split for developers is that composition is lost. For example, now you can modify a file in lib and coqide will rebuild just that single file and relink. I'm not sure how this is handled by Nix.

Zimmi48: @ejgallego Nix doesn't support fine-grained compositionality but only at the package level. So splitting packages (if we can ensure the sources are split as well) is actually in favor of compositionality.

vbgl:
> What's the problem with native compute, aren't these regular `.cmx` files?

Patchelf will issue one such notice for every installed executable
binary file it doesn’t know about.

Disabled native compute produces less such files. Therefore you see less
of these messages.

default.nix
ejgallego: So for a developer, if I modify a single file, say `lib/queue.ml`, and I want to build `coqide`, then only that file of the Coq package is rebuilt [as happens now with the regular Dune setup]?

default.nix
ejgallego: I am lost on one detail tho, what's the value of `$OCAMLFIND_DESTDIR` ?

ejgallego:
> Patchelf will issue one such notice for every installed executable binary file it doesn’t know about.

So the problem is that Coq is setting the executable bit on those files. That's a Coq bug I'd say.

default.nix
vbgl: `$prefix/lib/ocaml/$ocaml_version/site-lib/`

vbgl:
My bad. They are not executable.

Files that trigger this message are the `.coq-native/*.o`.

default.nix
Zimmi48: No the full package, but note that a developer using Nix wouldn't build with `nix-build`. They would use `nix-shell` to get all the dependencies and then would build locally using Dune like any other developer. The compositionality of having several Nix package would only get useful in CI, where `nix-build` is used. A fully nixified CI could support not rebuilding any of the external projects if only the documentation of Coq changed.

default.nix
ejgallego: So indeed shouldn't you just use `dune install --libdir="$OCAMLFIND_DESDIR" --prefix="$out"` instead of `ln` ?

vbgl: Do you mean running that in addition to the usual install? Are you suggesting to install Coq in full twice?

ejgallego: Umm, this should install Coq only once, but in the proper place, right?

ejgallego: I mean as the default install call; `lib` dir is already allowed to be relocated by distros, so here we have two things: OCaml libraries installed in the wrong place, and Coq libraries installed in the wrong place.

For 1) `--libdir` is standard and supported, for 2) we need to think / tweak `dune install` I think.

vbgl: What do you call Coq?

First, we install the binaries, the stdlib, the doc… under $prefix.

Then, this line puts the OCaml libraries (for plug-in writers) where they can be found by `ocamlfind`.

ejgallego: In fact I guess that all dune-built packages in Nix should follow this pattern; and indeed they do:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/ocaml/dune.nix#L24

So why should be Coq different here?

vbgl: You are right. As dune stands, since there is only one `lib` directory concept, you have to choose between OCaml libraries at the wrong place or Coq theories at the wrong place.

An other possibility could be to fix Coq so that the location of the theories is configurable (at configure time). With such a possibility, we could put both Coq and OCaml libraries in a single `lib` directory, where they can be found by `ocamlfind` as well as by `coqc`.

ejgallego: Isn't the Nix Coq package setting `COQPATH` anyways? I thought that in your setup the original Coq configure location didn't matter so much, right?

IMHO we should first install OCaml stuff properly [as done in dune.nix], and indeed we should fix Dune so Coq libraries can be installed properly, for now they will be in the wrong place but at least on Nix [and OPAM] things should be fine as they use `COQPATH`.

Note that Coq's configure already has libdir which will allow to do that, Dune supports calling configure before the build so you can do: `./configure -prefix $out -libdir $OCAMLFIND_DESTDIR -native-compiler no` and then build normally.

I'll fix this native compiler stuff by the way, it is the only real annoying difference we've got.

ejgallego:
> Files that trigger this message are the `.coq-native/*.o`.

That deserves some investigation, if regular `.o` files don't trigger the warning, "native" ones should not IMO.

default.nix
vbgl: Thanks for repeating that again. I thought I had tried but I was doing the quoting wrong…

default.nix
vbgl: Thanks for the tip.

ejgallego: In particular it is used for the test-suite so indeed it may be useful to have around.

ejgallego:
Cool, in fact I don't dislike the `lib/ocaml/4.06.1/site-lib/coq/theories/Numbers/Integer/Abstract/` path so much given that `.vo` files are produced by marshall and thus, 100% dependent on the OCaml version number.

So I guess the TODO from Dune part could be something such as:
- drop the need for `voboot` [then dune install -p can be used directly]
- setup `libdir` automatically from configure if in Dune mode and OCAMLFIND_DESTDIR is present.

vbgl:
Test-suite fails with:
~~~
logs/tools/update-compat.log
  File "dev/tools/update-compat.py", line 84, in <module>
    assert(os.path.exists(CONFIGURE_PATH))
AssertionError
~~~

Could it be a missing dependency in the `test-suite/dune` file?

ejgallego:
> Could it be a missing dependency in the `test-suite/dune` file?

I dunno, likely some assumption about paths in that script which is fairly ad-hoc and fragile.

Zimmi48:
This is a test-suite file that should typically become a promote target instead. Similar to what we have now for `doc/sphinx/README.rst` and also similar to the few ML files that are extracted from Coq files... Contrary to most tests in the test-suite, it doesn't really make sense for such a test to be run *post-install*.

vbgl:
Updated the `test-suite/dune` file: this fixes the issue locally.

dev/ci/nix/coq.nix
```diff
@@ -5,5 +5,4 @@ let coq = callPackage wd { buildDoc = false; doInstallCheck = false; coq-version
 coq.overrideAttrs (o: {
   name = "coq-local-${branch}";
   src = fetchGit "${wd}";
-  enableParallelBuilding = true;
```
Zimmi48: Didn't you forget to add this to `default.nix`?

vbgl: No. This affects the arguments of `make`. No parallelism between the various `Makefile.dune` targets is expected. `dune` has its own heuristics.

ejgallego (commented):
The parts I understand do look good to me, however I still have one major concern:

- when is `dune build -p` used? in other words, when do we build in release mode, and when in dev profile?

minor comments:
- we would like to fix configure so OCAMLFIND_DESTDIR is used, that is easy enough to do here, otherwise I'll do later,
- the current setup is a bit cumbersome, I dunno if we should wait to merge this until we remove the need for `voboot`, up to Nix people I think

Zimmi48:
>     * when is `dune build -p` used? in other words, when do we build in release mode, and when in dev profile?

My understanding is that using it will need to split the Nix package in several parts, which is future work.



> * 
> 
> * the current setup is a bit cumbersome, I dunno if we should wait to merge this until we remove the need for `voboot`, up to Nix people I think

That was the initial plan but it doesn't look like it is going to be so soon. In particular, we might decide to remove the legacy Makefile in 8.10, but we don't know if the native Dune support will be ready in time for this.

vbgl:
Maybe we should consider this PR as a fruitful experiment and delay a bit the real switch to dune (including split packages, fixed `voboot` and native-compute, …).

ejgallego:
> it doesn't look like it is going to be so soon.

I'm convinced it is gonna happen sooner than what everybody expects :) but indeed, from my part there is no need to rush this; all the things that really require a Dune build are as I mentioned 8.11 material in my list, and I can develop using Dune just fine these days.

gares:
What is the transition strategy for third party plugins? It occurred to me that unless #9735 is fixed, one cannot build (external) plugins using coq_makefile. This means all external projects containing ml code have to be ported to dune (assuming this works) as well.

Zimmi48:
> What is the transition strategy for third party plugins? It occurred to me that unless #9735 is fixed, one cannot build (external) plugins using coq_makefile. This means all external projects containing ml code have to be ported to dune (assuming this works) as well.

The switch of build system within Coq shouldn't affect people depending on Coq. In particular, even if the native Coq support in Dune should give rise to an alternative to coq_makefile, which is likely to be popular, coq_makefile should be maintained as long as there are projects that keep using it...

So this PR cannot be merged until at least #9739 is merged. But given this comment:

>Maybe we should consider this PR as a fruitful experiment and delay a bit the real switch to dune (including split packages, fixed voboot and native-compute, …).

We might decide to wait longer.

ejgallego:
> What is the transition strategy for third party plugins?

They should work without changes, the `gramlib` stuff happened because we did change some stuff there after #8729 , but indeed #8729 has been green for all plugins for a while.

vbgl:
Hi. I try to revive this PR. Is the dune setting ready for building the documentation? When I run `dune build -p coq-doc` (in the current `master` branch), I get the following error. I don’t know what it means.

> File "doc/dune", line 22, characters 0-386:
22 | (rule
23 |  (targets refman-html)
24 |  (alias refman-html)
....33 |   (env_var SPHINXWARNOPT))
34 |  (action
35 |   (run env COQLIB=%{project_root} sphinx-build -q %{env:SPHINXWARNOPT=-W} -b html sphinx %{targets})))
Error: No rule found for alias .coq-files

ejgallego:
> Hi. I try to revive this PR. Is the dune setting ready for building the documentation? When I run `dune build -p coq-doc` (in the current `master` branch), I get the following error. I don’t know what it means.

You may have to use `dune build -p coq,coq-doc` ?

`dune build -p ` builds in "release mode" so it will pick the system-installed Coq so instead some stuff will be missing. Let me try.

ejgallego:
The concrete error seems strange as I cannot find any `.coq-files` string in the Coq repos.

ejgallego:
Hi folks so indeed that may be a good moment to think what we are missing and whether we need to do something in dune, coq, or both. What is the status of this then?

Also note that https://github.com/ocaml/dune/issues/1901 is still open.

Zimmi48:
We might need a makefile wrapper to handle the build and installation of the documentation with a Dune-built Coq then.

ejgallego:
> We might need a makefile wrapper to handle the build and installation of the documentation with a Dune-built Coq then.

The build does work OK, installation indeed does not as of today, but actually I think that could be fixed. Let me have a look.

ejgallego:
> The build does work OK, installation indeed does not as of today, but actually I think that could be fixed. Let me have a look.

Note that I was not able to understand the error @vbgl got, neither to reproduce.

ejgallego:
We are going to add coqdoc support really soon, so the only thing is the manual installation, but as it is a single directory what we are talking about we can add glob support to `(install ...)` stanzas and that should work pretty well IIUC.

Zimmi48:
There's also the refman. In this case, there are also sub-directories. And globs still do not support referring to all files included in sub-directories. Otherwise, that could simplify:

https://github.com/coq/coq/blob/e583be62b74d71b5af159700e3a31f78fec9a7d2/doc/tools/docgram/dune#L16-L29

and

https://github.com/coq/coq/blob/e583be62b74d71b5af159700e3a31f78fec9a7d2/doc/tools/docgram/dune#L31-L42

a lot.

ejgallego:
> There's also the refman. In this case, there are also sub-directories. And globs still do not support referring to all files included in sub-directories.

There was progress fixing this, I need to check what's the status for Dune 2.8, maybe it is more efficient (and selfish) to ping @rgrinberg

rgrinberg:
Ack. It's not the highest priority feature since it doesn't unblock anyone. Just makes things more convenient.

vbgl:
Trying again today, the error is different (when running `dune build -p coq-doc`):

> File "doc/dune", line 29, characters 11-14:
29 |   (package coq)
                ^^^
Error: Package coq does not exist

ejgallego:
> Trying again today, the error is different (when running `dune build -p coq-doc`):

Indeed `coq-doc` requires `coq` to be built, is it installed?

[You can build them both at the same time in release mode using `-p coq,coq-doc`]

vbgl:
> Indeed |coq-doc| requires |coq| to be built, is it installed?

I think so. How do I check? (e.g., `ocamlfind query -r coq` is happy).

> [You can build them both at the same time in release mode using |-p 
> coq,coq-doc|]

I know, but I wished I could build the documentation alone.

ejgallego:
> I think so. How do I check? (e.g., `ocamlfind query -r coq` is happy).

Umm, let me have a look.

ejgallego:
> Umm, let me have a look.

Umm, so indeed `(package coq)` means "all files installed by package Coq", but in the current build context, so indeed that doesn't translate to a installed situation.

IIANM `sphinx` depends on `coqc` + some `.vo` libraries, so then we could indeed replace `(package coqc)` by `%{bin:coqc}` [this works and will correctly pick the system-wide Coq when `-p` is used` and some `%{lib:coq:theories/...}` or `%{lib-available:coq}`, here it is not clear yet how to best do it.

vbgl:
Can’t dune call “coqc -where” to determine where to look for the Coq standard library?

ejgallego:
> Can’t dune call “coqc -where” to determine where to look for the Coq standard library?

It could, that's not a big deal, Dune knows anyways but other means. It is more that the current language for dependencies doesn't allow to express that in a rule.

ejgallego:
For reference that's the recursive glob issue https://github.com/ocaml/dune/issues/1866

PR #9695, opened 2019-03-04
[make install] Do not install the sphinx doctrees

vbgl:
I don’t know what these doctree files are and what they are used for. They refer to `python` (the interpreter) hence make the nix closure of Coq include python, sphinx etc. This is annoying for CI jobs which must download all this closure.

Ping @cpitclaudel.

### DIFF
--- a/Makefile.doc
+++ b/Makefile.doc
@@ -209,10 +209,11 @@ install-doc-printable:
 
 install-doc-sphinx:
 	$(MKDIR) $(FULLDOCDIR)/sphinx
-	(for f in `cd doc/sphinx/_build; find . -type f`; do \
-		$(MKDIR) $$(dirname $(FULLDOCDIR)/sphinx/$$f);\
-		$(INSTALLLIB) doc/sphinx/_build/$$f $(FULLDOCDIR)/sphinx/$$f;\
-	done)
+	(for d in html latex; do \
+	for f in `cd doc/sphinx/_build/$$d && find . -type f`; do \
+		$(MKDIR) $$(dirname $(FULLDOCDIR)/sphinx/$$d/$$f);\
+		$(INSTALLLIB) doc/sphinx/_build/$$d/$$f $(FULLDOCDIR)/sphinx/$$d/$$f;\
+	done; done)
 
 # For emacs:
 # Local Variables:


Makefile.doc
```diff
@@ -208,11 +208,13 @@ install-doc-printable:
 	$(INSTALLLIB) doc/stdlib/Library.ps $(FULLDOCDIR)/ps
 
 install-doc-sphinx:
+	mv doc/sphinx/_build/doctrees .
 	$(MKDIR) $(FULLDOCDIR)/sphinx
 	(for f in `cd doc/sphinx/_build; find . -type f`; do \
 		$(MKDIR) $$(dirname $(FULLDOCDIR)/sphinx/$$f);\
 		$(INSTALLLIB) doc/sphinx/_build/$$f $(FULLDOCDIR)/sphinx/$$f;\
 	done)
+	mv doctrees doc/sphinx/_build/
```
ejgallego: This rule seems pretty weird and will leave the sphinx cache in non consistent status would the command fail. Instead, just install the `html` folder.

Makefile.doc
```diff
@@ -209,10 +209,11 @@ install-doc-printable:
...
 	$(MKDIR) $(FULLDOCDIR)/sphinx
-	(for f in `cd doc/sphinx/_build; find . -type f`; do \
-		$(MKDIR) $$(dirname $(FULLDOCDIR)/sphinx/$$f);\
-		$(INSTALLLIB) doc/sphinx/_build/$$f $(FULLDOCDIR)/sphinx/$$f;\
-	done)
+	(for d in html latex; do \
+	for f in `cd doc/sphinx/_build/$$d && find . -type f`; do \
+		$(MKDIR) $$(dirname $(FULLDOCDIR)/sphinx/$$d/$$f);\
+		$(INSTALLLIB) doc/sphinx/_build/$$d/$$f $(FULLDOCDIR)/sphinx/$$d/$$f;\
+	done; done)
```
ejgallego: Sorry to be annoying, but why to use another `for`?

I guess you can leave it like this as it is not worth the effort, however the whole rule seem bugged, for example w.r.t. quoting, certainly the `cd` shouldn't be there. Something such as:
```
find doc/sphinx/_build{html,latex} -type f -exec "install -D {} $(FULLDOCDIR)/sphinx/";
```
would be much preferred IMVHO.

[I forgot the exact find syntax]

ejgallego: Yeah i'd need a bit more tweaking, never mind, not worth the effort to fix this. The problem of for loops is that you usually need to re-quote.

vbgl:
Ping @coq/doc-maintainers 

Should the `doctrees` be installed? If not, can we merge this? Thanks.

Zimmi48:
@cpitclaudel You are the only one that can answer this. I personally do not even know what this doctrees are, so I wouldn't have anything against merging the PR.

cpitclaudel (approved):
No need to install doctrees, indeed.

cpitclaudel:
Can someone set an appropriate milestone? Then I (or they ^^) can merge.

Issue #9658, opened 2019-02-26
[ci] base+async tends to fail often


ejgallego:
As pointed out on gitter it is not rare to see this job randomly failing. Example:

- https://gitlab.com/coq/coq/-/jobs/167989714
  > `Error: Tactic failure: reversible in 1st order mode.`
- https://gitlab.com/coq/coq/-/jobs/168148638
  > `make[1]: *** [theories/Reals/Rpower.vo] Segmentation fault (core dumped)`
- https://gitlab.com/coq/coq/-/jobs/167922293
  > `Error: Illegal application: 
The term "@Logic.eq" of type "forall A : Type, A -> A -> Prop"
cannot be applied to the terms
 "Z" : "Set"
 "z" : "Z"
 "[|x|]" : "Z"
The 1st term has type "Set" which should be coercible to 
"Type".`

Let's collect failures here as they tend to be diverse, and let's collect also the test-suite:async failures here.

SkySkimmer:
People talk about our CI on twitter?

ejgallego:
I mean Gitter.

ejgallego:
Failures are happening at an alarming rate (cc: @maximedenes @gares ) thus I'd dare to say that the status of this job in the CI may be demoted to `allow_failure` if things continues like that.

ejgallego:
Failures continue and indeed there seem to be some state handling problems in the STM, for example it is not clear to me if declarations that may involved hooks are correctly determined to be ```Sync`.

ejgallego:
Plenty of failures https://gitlab.com/coq/coq/-/jobs/179101771 https://gitlab.com/coq/coq/-/jobs/179093502 https://gitlab.com/coq/coq/-/jobs/179053653 https://gitlab.com/coq/coq/-/jobs/179043712 https://gitlab.com/coq/coq/-/jobs/178917545

this is just on the first page of gitlab as of now; this misuse of resources is jeopardizing the whole CI infrastructure so I am much afraid that we cannot continue like this for long.

maximedenes:
@gares Any idea how we can debug this?

ppedrot:
The universe errors are not surprising, I believe this is just the non-commutativity of the inference algorithm at work. I wouldn't have thought that this would happen in practice though. Some test failures are probably an instance of this as well. The SEGFAULT is much more concerning in my opinion.

gares:
[...]

PR #9698, opened 2019-03-05
Remove last random failures

### LINKED ISSUE #5985: output-modulo-time/ltacprof_cutoff.v is non deterministic
[`output-modulo-time/ltacprof_cutoff.v`](https://github.com/coq/coq/blob/master/test-suite/output-modulo-time/ltacprof_cutoff.v) failed randomly [on GitLab](https://gitlab.com/SkySkimmer/coq/-/jobs/17569286).

#### Failure log (edit to complete)

- On GitLab on 2017-05-31

### LINKED ISSUE #9637: [ci] build:quick takes around 3 hours to complete sometimes.
Sometimes, the `build:quick` job takes around 3 hours to complete. Examples:
- https://gitlab.com/coq/coq/-/jobs/166438863
- https://gitlab.com/coq/coq/-/jobs/166471390

IMHO this makes the job fail the "reasonable build time" clause of the CI rules, thus we I suggest we place it on a 30-days quarantine, and if the problem is not fixed, remove.

Zimmi48:
**Kind:** infrastructure.

Closes #5985.

Let's not close #9637 and #9658 just yet because the final move should be to find the issue and fix them or remove them from CI entirely.


.gitlab-ci.yml
```diff
@@ -252,12 +252,14 @@ build:base+async:
   variables:
     COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
     COQUSERFLAGS: "-async-proofs on"
+  allow_failure: true
```
SkySkimmer: Maybe put `see #9658` as a comment.

Zimmi48: Yes, will do.

Zimmi48: Done.

ejgallego (approved):
Unfortunately failure rate of these entries seems to be really high (cc: @maximedenes @gares ) and in many cases they fail by timeout which will keep a Gitlab worker busy for 3 hours, so I see no other choice.

I am not sure how long we should keep this jobs in the CI tho, as even in this mode they still take quite a bit of resources. I'd say that if 3 weeks the maintainers haven't figured out what's causing the problem we should disable the jobs entirely.

gares:
My impression is that failures started to occur recently, when the jobs were added they were more reliable. So something has changed, even if we don't know why.

I don't see where the 3 weeks ultimatum comes from, I'm against removing the jobs before someone has time to look into it seriously. If they often hang, reduce their timeout to something closer to their expected time.

ejgallego:
> So something has changed, even if we don't know why.

Only change that I can say is that we tend to get way more workers with only 1 processor. But the failures have been since the very beginning I'm afraid.

> I don't see where the 3 weeks ultimatum comes from

It comes from the written down CI policy; and indeed, no point in holding on an allow_failure job for a long time. What's the point on doing the test if we don't pay attention to it?

Only exception so far is OCaml `master`, but that's also up for discussion, it was more of an experiment.

ejgallego:
> I'm against removing the jobs before someone has time to look into it seriously.

There has been plenty of time already to look at the failures, we don't have a hard policy, so if you folks tell us "we will solve this in around 3 months" that'd be OK.

But even so, removing is not such a big deal, we can remove and add back very easily if that's the worry; but getting red marks on unreliable works is bad and makes _everybody_ lose time, at least thanks to the detailed reporting we can see what's wrong quickly tho.

Regarding this particular PR, I do propose to leave a week to further observe how CI is doing; if the failures still persist then we will move the jobs to the allow_failure state, and re-evaluate again like in month, maybe for the beta so 8.10 can have reliable CI?

@SkySkimmer , what are your thoughts on this?

SkySkimmer:
@ejgallego full agreement with you so far

Zimmi48:
>If they often hang, reduce their timeout to something closer to their expected time.

I would be perfectly OK with keeping the jobs in allow failure mode with a short timeout. Unfortunately, this is not something that GitLab supports (https://gitlab.com/gitlab-org/gitlab-ce/issues/18183). So if you can find another way of doing the same thing, that would be good too.

>It comes from the written down CI policy; and indeed, no point in holding on an allow_failure job for a long time. What's the point on doing the test if we don't pay attention to it?

It could make sense if someone says it is useful to them, and they look at it regularly.

ejgallego:
> It could make sense if someone says it is useful to them, and they look at it regularly.

How can it be useful? I mean, what more are we going to learn that we haven't learned yet?

We can keep the thing there for a few weeks or even months, but in the end, if there is no active development I don't see the point on the resource use.

gares:
> Only change that I can say is that we tend to get way more workers with only 1 processor

I don't know how NJOBS works. If you are on these workers, is NJOBS=1 ?

gares:
>  I mean, what more are we going to learn that we haven't learned yet?

What does this mean?

gares:
> t could make sense if someone says it is useful to them, and they look at it regularly.

It is useful to me, it tests some code paths in the STM the otherwise would only be tested by using CoqIDE interactively.

> There has been plenty of time already to look at the failures

The oldest issue up there (not counting the one being closed by removing the file) is 10 days old.

As I told you, I'm pretty sure the failures were not there before, or at least I did not see them.
For sure when the async job was added, we tried it a few times on the PR before considering it.
We found a real race and it was fixed, after that it behaved reliably. (We = me and @maximedenes )

ejgallego:
> What does this mean?

I mean, all the failures fall in the same patterns, so unless there is active development, the info of the job is not very useful.

> I don't know how NJOBS works. If you are on these workers, is NJOBS=1 ?

No, we don't attempt to detect processors (as this some other problems) `NJOBS=2` is a pretty good default value.

gares:
>  unless there is active development, the info of the job is not very useful.

I really don't know what you are talking about. The last PR fixing an STM issue was opened yesterday (and merged by you), so I guess this counts for active development.

I'm sorry these failures got very frequent, they are probably due to a bug in the code I wrote six years ago, but I don't see how removing the job from CI is going to solve the problems. Yes, it is a red cross, and you have to press "y" to ignore the failure. I had to do the same a million times, and not because this job failing.

I'll do my best to look into the issue, assuming I can reproduce it locally (I couldn't, I asked how to access the machines in one of the issue, and the answer was roughly: you can't...). Maybe I'll try with a VM with just one core, now that I know there is a correlation.

Zimmi48:
You can also add as much debugging info to the CI log if that's helpful.

Zimmi48:
Again a failure: https://gitlab.com/coq/coq/-/jobs/174388603

I don't understand why this PR hasn't been already merged. The debate was not about marking as `allow_failure` but whether to remove (later), so that shouldn't be blocking.

ejgallego:
> I don't understand why this PR hasn't been already merged.

The idea was to gather a bit more data on the failure rate, but activity on the CI system has been very low so far so we couldn't gather a lot of data.

> The debate was not about marking as `allow_failure` but whether to remove (later), so that shouldn't be blocking.

With the current policy, setting allow failure implies that a removal count starts for the project, so indeed, both are related. I'm not sure what is @gares position with regards to setting the allow failure flag.

SkySkimmer:
>With the current policy, setting allow failure implies that a removal count starts for the project

That's only for 3rd party stuff (at least as far as explicit policy is concerned)

ejgallego:
@SkySkimmer I'll let you handle this one.

SkySkimmer:
Goodbye ltac_cutoff
I never had a clue what it was about

ejgallego:
Even if this merged, jobs hitting the timeout is still pretty bad as in particular it flags us to Gitlab as a project making improper use of resources which could mean non-compliance with their terms of service.

Thus, I propose to keep the jobs until the 8.10 branching point.

If we agree on that date, then if the timeouts / failures have not been fixed we shall deactivate the running of the async stuff until a fix has been applied.

SkySkimmer:
>Even if this merged, jobs hitting the timeout is still pretty bad as in particular it flags us to Gitlab as a project making improper use of resources which could mean non-compliance with their terms of service.

What are you basing this on?

ejgallego:
> What are you basing this on?

I didn't want to highlight our large resource to Gitlab by asking where Coq is in their list directly, but indeed all providers do monitor resource usage and sort resource use by project.

Haven't you noticed our worker allocation seems to have gotten worse in the last few weeks?

We could ask indeed but I'm afraid of having a bad effect.

ejgallego:
Anyways, as a matter of infrastructure principles, having jobs that often timeout and block a full worker for 3 hours it is just not acceptable, [that worker could process dozen of other jobs].

So IMHO the soft place of 2 months is fairly OK, if in two months we are not making progress then it can be removed and added back when ready.

SkySkimmer:
Gitlab already knows about us, they did that one PR about git depth.

 > Haven't you noticed our worker allocation seems to have gotten worse in
 > the last few weeks?

No, what are you talking about?

Gaëtan Gilbert

ejgallego:
> No, what are you talking about?

The number of workers we get is consistently lower these days.

> Gitlab already knows about us, they did that one PR about git depth.

Of course they know us, that's what I mean.

ejgallego:
Anyways the reason is that I am not very friendly to these particular jobs is that the whole setup seems _very broken_ to me.

Let's take this job https://gitlab.com/coq/coq/-/jobs/175253451 it is running on pyrolise that has plenty of cores, so it is so far taking more than 1h , whereas its sister job has taken 7 minutes 35 seconds.

So async is supposed to sped builds up, however it makes things way slower, like 10x slower? Please understand my not very optimistic position about the code then.

ejgallego:
Said otherwise, we are testing a codepath nobody would ever want to use. Why to build / run the test suite with a mode that makes things much slower.

So indeed let's either add realistic use cases of this stuff, or just let's stop wasting resources.

gares:
> So async is supposed to sped builds up

This is not at all what async is about.

This job was added to CI to test the code that delegates proofs to workers, something that happens when using CoqIDE. Sure, the code path is not exactly the same, but much more code paths gets tested this way than what we can test with fake_ide for example. The fact that errors are spot means that the resources have not been wasted to me.

I'll do my best to find out the root cause. As I said before, I'm fine giving this job a strict timeout. If it is not yet implemented by gitlab, that would be a nice discussion to have with them.

PR #9699, opened 2019-03-05
[ci] [windows] Use an install path requiring proper quoting.

ejgallego:
Current quoting in Coq's infrastructure is not as systematic as it
should; IMHO this is a serious problem as it will make for random failures
would the user work with Coq inside a folder requiring quoting.

This PR switches the CI system as to use a destination directory
containing spaces; while this is still not optimal [the build dir is
free of spaces], it should provide some more assurance of proper
working wrt quoting for the runtime.

### DIFF
--- a/dev/build/windows/MakeCoq_MinGW.bat
+++ b/dev/build/windows/MakeCoq_MinGW.bat
@@ -40,7 +40,7 @@ REM see -destcyg in ReadMe.txt
 SET DESTCYG=C:\bin\cygwin_coq
 
 REM see -destcoq in ReadMe.txt
-SET DESTCOQ=C:\bin\coq
+SET DESTCOQ=C:\bin\the coq proof assistant
 
 REM see -setup in ReadMe.txt
 SET SETUP=setup-x86_64.exe
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -60,7 +60,7 @@ after_script:
   artifacts:
     name: "$CI_JOB_NAME"
     paths:
-      - _install_ci
+      - "_install ci"
       - config/Makefile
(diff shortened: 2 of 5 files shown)


dev/build/windows/MakeCoq_MinGW.bat
```diff
@@ -40,7 +40,7 @@ REM see -destcyg in ReadMe.txt
 SET DESTCYG=C:\bin\cygwin_coq
 
 REM see -destcoq in ReadMe.txt
-SET DESTCOQ=C:\bin\coq
+SET DESTCOQ=C:\bin\the coq proof assistant
```
Zimmi48: Are you sure about this? It looks like something that goes beyond strengthening CI.

ejgallego: Yup we should make this conditional to running in PRs for example.

I anticipate having this PR working will be lots of "fun" anyways.

MSoegtropIMC: Afaik the default path is not used. On the INRIA runners we are using random paths.
The defaults are more intended to be convenient for users.
If the CI path needs to be changed, it needs to be done in the CI file.

One more note: this test will not be terribly effective because this path is only used for installation. During build a relative path is used. This is because a Coq which can be installed into random folders needs to have relative paths inside. The only place where this path is used is during make install.

If we really want to test this, we actually need to install Coq somewhere and this path should have spaces in it.

SkySkimmer:
If you want really fun quoting errors, try using `-prefix 'coq $install'`

ejgallego:
I guess I am going to close this, fixing all the current quoting problems look like a daunting task; we may revisit later.

Issue #3166, opened 2013-11-23
Anomaly: Not enough lambda/let's. Please report.


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#3166
From: @JasonGross
Reported version: 8.4
CC:   @silene, @herbelin, @maximedenes

herbelin:
Note: @silene example (in his first comment) now works, but initial bug not (i.e. the underlying CCI issue is still there).

PR #9710, opened 2019-03-07
Compact kernel representation of pattern-matching

ppedrot:
This is a blunt implementation of @herbelin's [CEP 34](https://github.com/coq/ceps/pull/34) that I started afresh after trying more progressive changes.

This draft PR removes all type and term annotations on both branches and return clauses of pattern-matching nodes. In order to keep the ability to construct the the old expanded form out of the new compact form, I had to also had to add data that was stealthily present in the return clause, namely universe instances and parameters of the inductive type being eliminated on.

The case node now looks like
```
Case of
  case_info *
  Instance.t * (* universe instances of the inductive *)
  constr array * (* parameters of the inductive *)
  case_return * (* erased return clause *)
  constr * (* scrutinee *)
  case_branch array (* erased branches *)
```
where
```
type case_branch = Name.t array * constr
type case_return = Name.t array * types
```

Note that let-bindings are not present in the term, which requires access to the environment in some cases, e.g. to compute branch reduction. (There is a fast-path for non-let-containing inductive types though, which are like 99.9% of them.)

There are two conversion function from and to the expanded form:
- `[Inductive, EConstr].expand_case` that goes from the compact to the expanded form and cannot fail (assuming the term was well-typed)
- `[Inductive, EConstr].contract_case` that goes the other way around and will raise anomalies if the expanded forms are not fully eta-expanded.

For now the prelude compiles, but I get to random anomalies and unification errors quickly after. I need help to try understanding what is going on. I already fixed a few calls that were generating non-eta expanded forms, but we should also get rid of the conversion function in the end.

As remarked on gitter, I discovered in the process that even disregarding application order, conversion doesn't preserve the invariant that it only compares terms living in a common supertype. The reason is that pattern-matching is not carrying enough information to check that, when converting to match nodes, both scrutinees have the same type. The only thing we know is that the two terms are living in the same inductive types with the same universes and parameters, but they could very well have different indices. We might solve that by also adding the indices in the match node.

### DIFF
--- a/kernel/constr.mli
+++ b/kernel/constr.mli
@@ -141,14 +141,29 @@ val mkRef : GlobRef.t Univ.puniverses -> constr
 
 (** Constructs a destructor of inductive type.
 
-    [mkCase ci p c ac] stand for match [c] as [x] in [I args] return [p] with [ac]
+    [mkCase ci params p c ac] stand for match [c] as [x] in [I args] return [p] with [ac]
     presented as describe in [ci].
 
-    [p] structure is [fun args x -> "return clause"]
+
+    [p] structure is [args x |- "return clause"]
 
     [ac]{^ ith} element is ith constructor case presented as
-    {e lambda construct_args (without params). case_term } *)
-val mkCase : case_info * constr * constr * constr array -> constr
+    {e construct_args |- case_term } *)
+
+type 'constr pcase_branch = Name.t Context.binder_annot array * 'constr
+(** Names of the indices + name of self *)
+
+type 'types pcase_return = Name.t Context.binder_annot array * 'types
+(** Names of the branches *)
+
+type ('constr, 'types, 'univs) pcase =
+  case_info * 'univs * 'constr array * 'types pcase_return * 'constr * 'constr pcase_branch array
+
+type case_return = types pcase_return
+type case_branch = constr pcase_branch
+type case = (constr, types, Univ.Instance.t) pcase
+
+val mkCase : case -> constr
 
 (** If [recindxs = [|i1,...in|]]
       [funnames = [|f1,.....fn|]]
@@ -232,7 +247,7 @@ type ('constr, 'types, 'sort, 'univs) kind_of_term =
 
   | Ind       of (inductive * 'univs)                 (** A name of an inductive type defined by [Variant], [Inductive] or [Record] Vernacular-commands. *)
   | Construct of (constructor * 'univs)              (** A constructor of an inductive type defined by [Variant], [Inductive] or [Record] Vernacular-commands. *)
-  | Case      of case_info * 'constr * 'constr * 'constr array
+  | Case      of case_info * 'univs * 'constr array * 'types pcase_return * 'constr * 'constr pcase_branch array
   | Fix       of ('constr, 'types) pfixpoint
   | CoFix     of ('constr, 'types) pcofixpoint
   | Proj      of Projection.t * 'constr
@@ -339,7 +354,7 @@ Ci(...yij...) => ti | ... end] (or [let (..y1i..) := c as x in I args
 return P in t1], or [if c then t1 else t2])
 @return [(info,c,fun args x => P,[|...|fun yij => ti| ...|])]
 where [info] is pretty-printing information *)
-val destCase : constr -> case_info * constr * constr * constr array
+val destCase : constr -> case
 
 (** Destructs a projection *)
 val destProj : constr -> Projection.t * constr
@@ -409,12 +424,6 @@ val lift : int -> constr -> constr
 (** {6 Functionals working on expressions canonically abstracted over
        a local context (possibly with let-ins)} *)
 
-(** [map_under_context f l c] maps [f] on the immediate subterms of a
-    term abstracted over a context of length [n] (local definitions
-    are counted) *)
-
-val map_under_context : (constr -> constr) -> int -> constr -> constr
-
 (** [map_branches f br] maps [f] on the immediate subterms of an array
    of "match" branches [br] in canonical eta-let-expanded form; it is
    not recursive and the order with which subterms are processed is
@@ -422,7 +431,7 @@ val map_under_context : (constr -> constr) -> int -> constr -> constr
    types and possibly terms occurring in the context of each branch as
    well as the body of each branch *)
 
-val map_branches : (constr -> constr) -> case_info -> constr array -> constr array
+val map_branches : (constr -> constr) -> case_branch array -> case_branch array
 
 (** [map_return_predicate f p] maps [f] on the immediate subterms of a
    return predicate of a "match" in canonical eta-let-expanded form;
@@ -431,16 +440,7 @@ val map_branches : (constr -> constr) -> case_info -> constr array -> constr arr
    the types and possibly terms occurring in the context of each
    branch as well as the body of the predicate *)
 
-val map_return_predicate : (constr -> constr) -> case_info -> constr -> constr
-
-(** [map_under_context_with_binders g f n l c] maps [f] on the
-    immediate subterms of a term abstracted over a context of length
-    [n] (local definitions are counted); it preserves sharing; it
-    carries an extra data [n] (typically a lift index) which is
-    processed by [g] (which typically add 1 to [n]) at each binder
-    traversal *)
-
-val map_under_context_with_binders : ('a -> 'a) -> ('a -> constr -> constr) -> 'a -> int -> constr -> constr
+val map_return_predicate : (constr -> constr) -> case_return -> case_return
 
 (** [map_branches_with_binders f br] maps [f] on the immediate
    subterms of an array of "match" branches [br] in canonical
@@ -452,7 +452,7 @@ val map_under_context_with_binders : ('a -> 'a) -> ('a -> constr -> constr) -> '
    occurring in the context of the branch as well as the body of the
    branch *)
 
-val map_branches_with_binders : ('a -> 'a) -> ('a -> constr -> constr) -> 'a -> case_info -> constr array -> constr array
+val map_branches_with_binders : ('a -> 'a) -> ('a -> constr -> constr) -> 'a -> case_branch array -> case_branch array
 
 (** [map_return_predicate_with_binders f p] maps [f] on the immediate
    subterms of a return predicate of a "match" in canonical
@@ -464,7 +464,7 @@ val map_branches_with_binders : ('a -> 'a) -> ('a -> constr -> constr) -> 'a -> 
    occurring in the context of each branch as well as the body of the
    predicate *)
 
-val map_return_predicate_with_binders : ('a -> 'a) -> ('a -> constr -> constr) -> 'a -> case_info -> constr -> constr
+val map_return_predicate_with_binders : ('a -> 'a) -> ('a -> constr -> constr) -> 'a -> case_return -> case_return
 
 (** {6 Functionals working on the immediate subterm of a construction } *)
 
@@ -479,13 +479,6 @@ val fold : ('a -> constr -> 'a) -> 'a -> constr -> 'a
    not specified *)
 
 val map : (constr -> constr) -> constr -> constr
-
-(** [map_user_view f c] maps [f] on the immediate subterms of [c]; it
-   differs from [map f c] in that the typing context and body of the
-   return predicate and of the branches of a [match] are considered as
-   immediate subterm of a [match] *)
-
-val map_user_view : (constr -> constr) -> constr -> constr
 
 (** Like {!map}, but also has an additional accumulator. *)
 
--- a/kernel/constr.mli
+++ b/kernel/constr.mli
@@ -466,25 +466,6 @@ val map_branches_with_binders : ('a -> 'a) -> ('a -> constr -> constr) -> 'a -> 
 
 val map_return_predicate_with_binders : ('a -> 'a) -> ('a -> constr -> constr) -> 'a -> case_info -> constr -> constr
 
-(** [map_under_context_with_full_binders g f n l c] is similar to
-    [map_under_context_with_binders] except that [g] takes also a full
-    binder as argument and that only the number of binders (and not
-    their signature) is required *)
-
-val map_under_context_with_full_binders : ((constr, constr) Context.Rel.Declaration.pt -> 'a -> 'a) -> ('a -> constr -> constr) -> 'a -> int -> constr -> constr
-
-(** [map_branches_with_full_binders g f l br] is equivalent to
-   [map_branches_with_binders] but using
-   [map_under_context_with_full_binders] *)
-
-val map_branches_with_full_binders : ((constr, constr) Context.Rel.Declaration.pt -> 'a -> 'a) -> ('a -> constr -> constr) -> 'a -> case_info -> constr array -> constr array
-
-(** [map_return_predicate_with_full_binders g f l p] is equivalent to
-   [map_return_predicate_with_binders] but using
-   [map_under_context_with_full_binders] *)
-
-val map_return_predicate_with_full_binders : ((constr, constr) Context.Rel.Declaration.pt -> 'a -> 'a) -> ('a -> constr -> constr) -> 'a -> case_info -> constr -> constr
-
 (** {6 Functionals working on the immediate subterm of a construction } *)
 
 (** [fold f acc c] folds [f] on the immediate subterms of [c]
@@ -492,10 +473,6 @@ val map_return_predicate_with_full_binders : ((constr, constr) Context.Rel.Decla
    the usual representation of the constructions; it is not recursive *)
 
 val fold : ('a -> constr -> 'a) -> 'a -> constr -> 'a
-
-val fold_with_full_binders :
-  (rel_declaration -> 'a -> 'a) -> ('a -> 'b -> constr -> 'b) ->
-    'a -> 'b -> constr -> 'b
 
 (** [map f c] maps [f] on the immediate subterms of [c]; it is
    not recursive and the order with which subterms are processed is
--- a/kernel/reduction.ml
+++ b/kernel/reduction.ml
@@ -56,7 +56,7 @@ let compare_stack_shape stk1 stk2 =
     | (_, Zapp l2::s2) -> compare_rec (bal-Array.length l2) stk1 s2
     | (Zproj _p1::s1, Zproj _p2::s2) ->
         Int.equal bal 0 && compare_rec 0 s1 s2
-    | (ZcaseT(_c1,_,_,_)::s1, ZcaseT(_c2,_,_,_)::s2) ->
+    | (ZcaseT(_c1,_,_,_,_,_)::s1, ZcaseT(_c2,_,_,_,_,_)::s2) ->
         Int.equal bal 0 (* && c1.ci_ind  = c2.ci_ind *) && compare_rec 0 s1 s2
     | (Zfix(_,a1)::s1, Zfix(_,a2)::s2) ->
         Int.equal bal 0 && compare_rec 0 a1 a2 && compare_rec 0 s1 s2
@@ -74,7 +74,7 @@ type lft_constr_stack_elt =
     Zlapp of (lift * fconstr) array
   | Zlproj of Projection.Repr.t * lift
   | Zlfix of (lift * fconstr) * lft_constr_stack
-  | Zlcase of case_info * lift * constr * constr array * fconstr subs
+  | Zlcase of case_info * lift * Univ.Instance.t * constr array * case_return * case_branch array * fconstr subs
   | Zlprimitive of
      CPrimitives.t * pconstant * lft_fconstr list * lft_fconstr next_native_args
 and lft_constr_stack = lft_constr_stack_elt list
@@ -109,8 +109,8 @@ let pure_stack lfts stk =
             | (Zfix(fx,a),(l,pstk)) ->
                 let (lfx,pa) = pure_rec l a in
                 (l, Zlfix((lfx,fx),pa)::pstk)
-            | (ZcaseT(ci,p,br,e),(l,pstk)) ->
-                (l,Zlcase(ci,l,p,br,e)::pstk)
+            | (ZcaseT(ci,u,pms,p,br,e),(l,pstk)) ->
+                (l,Zlcase(ci,l,u,pms,p,br,e)::pstk)
             | (Zprimitive(op,c,rargs,kargs),(l,pstk)) ->
                 (l,Zlprimitive(op,c,List.map (fun t -> (l,t)) rargs,
                             List.map (fun (k,t) -> (k,(l,t))) kargs)::pstk))
@@ -230,6 +230,9 @@ let sort_cmp_universes env pb s0 s1 (u, check) =
 let convert_instances ~flex u u' (s, check) =
   (check.compare_instances ~flex u u' s, check)
 
+let convert_instances_cumul pb var u u' (s, check) =
+  (check.compare_cumul_instances pb var u u' s, check)
+
 let get_cumulativity_constraints cv_pb variance u u' =
   match cv_pb with
   | CONV ->
@@ -289,8 +292,6 @@ let conv_table_key infos k1 k2 cuniv =
   | RelKey n, RelKey n' when Int.equal n n' -> cuniv
   | _ -> raise NotConvertible
 
-exception IrregularPatternShape
-
 let unfold_ref_with_args infos tab fl v =
   match unfold_reference infos tab fl with
   | Def def -> Some (def, v)
@@ -317,17 +318,6 @@ let push_relevance infos r =
 
 let push_relevances infos nas =
   { infos with relevances = Array.fold_left (fun l x -> Range.cons x.Context.binder_relevance l) infos.relevances nas }
-
-let rec skip_pattern infos relevances n c1 c2 =
-  if Int.equal n 0 then {infos with relevances}, c1, c2
-  else match kind c1, kind c2 with
-    | Lambda (x, _, c1), Lambda (_, _, c2) ->
-      skip_pattern infos (Range.cons x.Context.binder_relevance relevances) (pred n) c1 c2
-    | _ -> raise IrregularPatternShape
-
-let skip_pattern infos n c1 c2 =
-  if Int.equal n 0 then infos, c1, c2
-  else skip_pattern infos infos.relevances n c1 c2
 
 let is_irrelevant infos lft c =
   let env = info_env infos.cnv_inf in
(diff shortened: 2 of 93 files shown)


SkySkimmer:
I'd like to understand better where these are used
~~~
  Instance.t * (* universe instances of the inductive *)
  constr array * (* parameters of the inductive *)
~~~

kernel/constr.mli
```diff
@@ -227,7 +243,7 @@ type ('constr, 'types, 'sort, 'univs) kind_of_term =
 
   | Ind       of (inductive * 'univs)                 (** A name of an inductive type defined by [Variant], [Inductive] or [Record] Vernacular-commands. *)
   | Construct of (constructor * 'univs)              (** A constructor of an inductive type defined by [Variant], [Inductive] or [Record] Vernacular-commands. *)
-  | Case      of case_info * 'constr * 'constr * 'constr array
+  | Case      of case_info * 'univs * 'constr array * 'types pcase_return * 'constr * 'constr pcase_branch array
```
SkySkimmer: This may be a good time to consider using a record `Case of { case_info : case_info; ... }` WDYT?

ppedrot: Yeah but then the stupid warning for missing matched fields that is active on the kernel is going to be a PITA.

ppedrot:
> I'd like to understand better where these are used

They're present in the expanded form, as arguments to the "self" inductive type. If we want to have a compatibility layer, we need to store it somehow. Actually, as explained above that's not even enough we would also need the indices.

kernel/constr.mli
ejgallego: There's plenty of empirical evidence that the warning is quite useful. What's the problem in adding `; _`?

gares:
We had a discussion about some aspects of cep 34 at the latest math comp wg. Cc @ggonthier  @CohenCyril

kernel/constr.mli
ppedrot: It's not future-proof. Next time you add a field, the compiler is not going to help you. The alternative solution is to list all the fields but its really cumbersome.

gares: This is exactly why I think this warning on by default is a very silly choice.

If you use records, it is exactly because you don't want your code to break when you add a field.
Otherwise use a simple ADT with one constructor.

ppedrot:
I have fixed a few bugs, and an important chunk of the stdlib is now compiling. I am stuck on funind issues though, it seems that the printf-debugger (!) is calling the term printer in broken environments, leading to a failure. There is also another file that loops, but I'll defer investigations on this one.

kernel/constr.mli
maximedenes: While you are at it, you could use a record `Match of { ... }` instead ;)

ppedrot:
stdlib now compiles.

ppedrot:
Most of test-suite compiles, I ran a small bench to get an idea of what this would result in. Results are mixed.
```
┌────────────────────────┬──────────────────────────┬──────────────────────────────────────────────┬──────────────────────────────────────────────┬───────────────────────────────┬────────────────────────────┐
│                        │      user time [s]       │                  CPU cycles                  │               CPU instructions               │     max resident mem [KB]     │         mem faults         │
│                        │                          │                                              │                                              │                               │                            │
│           package_name │     NEW     OLD  PDIFF   │               NEW               OLD  PDIFF   │               NEW               OLD  PDIFF   │        NEW        OLD PDIFF   │     NEW     OLD    PDIFF   │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│               coq-corn │ 1389.74 1494.14  -6.99 % │     3880983145398     4169238105742  -6.91 % │     5329282847227     6168443959111 -13.60 % │     796880     855400 -6.84 % │       1      13   -92.31 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│   coq-mathcomp-algebra │  149.44  151.76  -1.53 % │      416073552834      423802839564  -1.82 % │      522476794700      534823451921  -2.31 % │     652460     637024 +2.42 % │      97       1 +9600.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│              coq-verdi │  121.63  122.56  -0.76 % │      338089304622      340211935037  -0.62 % │      433214237970      439567089919  -1.45 % │     606564     609152 -0.42 % │      10       5  +100.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│    coq-formal-topology │   29.95   30.15  -0.66 % │       82707667531       82793181886  -0.10 % │      107672834507      107992620849  -0.30 % │     467940     467796 +0.03 % │     108       0     +nan % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│ coq-mathcomp-character │  249.47  251.02  -0.62 % │      694766517449      698748415005  -0.57 % │      959779313768      963709315122  -0.41 % │    1065524    1067624 -0.20 % │       6      73   -91.78 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│           coq-compcert │ 1020.08 1022.70  -0.26 % │     2842094566310     2849588006409  -0.26 % │     4332339420432     4365515071354  -0.76 % │    1305080    1313168 -0.62 % │     461     377   +22.28 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│ coq-mathcomp-odd-order │ 1379.26 1380.74  -0.11 % │     3842392310442     3844816689021  -0.06 % │     6370022262314     6369146789068  +0.01 % │    1357980    1350216 +0.58 % │     120     115    +4.35 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│            coq-unimath │ 2052.55 2050.88  +0.08 % │     5713441901637     5710001946105  +0.06 % │     9968258813680     9988774110907  -0.21 % │     907768     956952 -5.14 % │     526     132  +298.48 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│       coq-math-classes │  217.31  216.58  +0.34 % │      607821172895      604564552471  +0.54 % │      765830059484      767247256435  -0.18 % │     515288     514492 +0.15 % │      75      17  +341.18 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│  coq-mathcomp-fingroup │   54.62   54.43  +0.35 % │      152147748604      151991298655  +0.10 % │      199241705987      198358817545  +0.45 % │     576696     574228 +0.43 % │       7      49   -85.71 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│             coq-sf-plf │   43.10   42.91  +0.44 % │      121097125199      120227042985  +0.72 % │      152613555954      151959980136  +0.43 % │     495176     490720 +0.91 % │      44       0     +nan % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│        coq-lambda-rust │ 2086.07 2076.46  +0.46 % │     5813762619287     5788247682723  +0.44 % │     7971084694665     7966666829792  +0.06 % │    1380140    1376840 +0.24 % │      54       5  +980.00 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│  coq-mathcomp-solvable │  191.68  190.73  +0.50 % │      534326078677      531751618937  +0.48 % │      716801301001      712410538477  +0.62 % │     852988     851444 +0.18 % │       5      15   -66.67 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│            coq-bignums │   69.96   69.56  +0.58 % │      194570670362      193974099211  +0.31 % │      256090434827      256830023959  -0.29 % │     518588     522952 -0.83 % │     181     349   -48.14 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│              coq-flocq │  566.20  562.22  +0.71 % │     1573096244500     1562812124935  +0.66 % │     2282295647036     2279755487562  +0.11 % │    1718344    1720084 -0.10 % │     525     603   -12.94 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│         coq-verdi-raft │ 1374.03 1364.19  +0.72 % │     3831611021406     3805021065215  +0.70 % │     5198613408413     5209682169883  -0.21 % │    1800372    1804952 -0.25 % │      20      32   -37.50 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│         coq-coquelicot │   73.43   72.80  +0.87 % │      201651746507      200313512782  +0.67 % │      241439550054      240665310551  +0.32 % │     688292     687880 +0.06 % │     214      11 +1845.45 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│ coq-mathcomp-ssreflect │   38.86   38.52  +0.88 % │      107371006106      107018493775  +0.33 % │      128472882014      128211245022  +0.20 % │     524908     523236 +0.32 % │      11      14   -21.43 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│     coq-mathcomp-field │  227.81  220.53  +3.30 % │      635184374001      614148640246  +3.43 % │      942156264746      892457328238  +5.57 % │     751008     750764 +0.03 % │       1      81   -98.77 % │
├────────────────────────┼──────────────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┼───────────────────────────────┼────────────────────────────┤
│        coq-fiat-crypto │ 5202.35 3945.34 +31.86 % │    14456057686179    10955693955750 +31.95 % │    21981366316992    17555499605820 +25.21 % │    2466872    2564480 -3.81 % │    1018    1022    -0.39 % │
└────────────────────────┴──────────────────────────┴──────────────────────────────────────────────┴──────────────────────────────────────────────┴───────────────────────────────┴────────────────────────────┘
```
I suspect that the fiat-crypto slowdown is due to the way I implemented `Reductionops`, there is probably an easy fix.

ppedrot:
Almost there, the test-suite has two remaining errors:
```
    bugs/opened/bug_3166.v...Error! (bug seems to be closed, please check)
    output/Cases.v...Error! (unexpected output)
```

ppedrot:
I need @herbelin's help to understand what needs to be done to fix #3166. I probably have to remove a pair of `expand_case` / `contract_case` somewhere in the matching algorithm but I don't know where, and I don't know how to adapt his patches on `Constr_matching` either.

kernel/constr.mli
```diff
@@ -138,14 +138,30 @@ val mkRef : GlobRef.t Univ.puniverses -> constr
...
     [ac]{^ ith} element is ith constructor case presented as 
-    {e lambda construct_args (without params). case_term } *)
-val mkCase : case_info * constr * constr * constr array -> constr
+    {e construct_args |- case_term } *)
+
+type 'constr pcase_branch = Name.t Context.binder_annot array * 'constr
+(** Names of the indices + name of self *)
+
+type 'types pcase_return = Name.t Context.binder_annot array * 'types
+(** Names of the branches *)
```
SkySkimmer: Is it OK to copy the relevances here? Maybe they should be gotten from the inductive_body like with the types.

ppedrot: No, otherwise you'd need to fetch them during reduction, always, and that would incur an important overhead for no good reason.

kernel/reduction.ml
```diff
@@ -645,11 +640,22 @@ and convert_stacks l2r infos lft1 lft2 stk1 stk2 cuniv =
             | (Zlfix(fx1,a1),Zlfix(fx2,a2)) ->
                 let cu2 = f fx1 fx2 cu1 in
                 cmp_rec a1 a2 cu2
-            | (Zlcase(ci1,l1,p1,br1,e1),Zlcase(ci2,l2,p2,br2,e2)) ->
+            | (Zlcase(ci1,l1,u1,pms1,p1,br1,e1),Zlcase(ci2,l2,u2,pms2,p2,br2,e2)) ->
                 if not (eq_ind ci1.ci_ind ci2.ci_ind) then
                   raise NotConvertible;
-                let cu2 = f (l1, mk_clos e1 p1) (l2, mk_clos e2 p2) cu1 in
-                convert_branches l2r infos ci1 e1 e2 l1 l2 br1 br2 cu2
+                let cu = cu1 in
```
SkySkimmer: Why not rename all of them to the usual `cuniv`?

ppedrot: Could do that but it would be a more line-touching patch just for uniformity...

SkySkimmer:
> No, otherwise you'd need to fetch them during reduction, always, and
 > that would incur an important overhead for no good reason.

Conversion not reduction right?
I guess the current lookup is only on polymorphic inductives (to check 
variance) which is still rare.

Gaëtan Gilbert

kernel/constr.mli
SkySkimmer: Conversion not reduction right? I guess the current lookup is only on polymorphic inductives (to check variance) which is still rare.

ppedrot:
@SkySkimmer Conversion indeed. For uniformity, I'd like to share the same representation for branches and return clause, so that if we remove the irrelevance marks from the clause, we also do it for branches. In which case we have an issue because branches need to know their annotations to perform conversion.

kernel/constr.mli
SkySkimmer: (I didn't mean specifically the _return one, I just had to pick one in the github interface)

ppedrot: You still need to push the relevance of binders in the environment during conversion though, so you need to have them somewhere.

ppedrot:
Bug #3166 fixed properly!

I now have a last problem, which is that `destruct` creates trivial β-redexes in the branches, e.g.
```
Goal True \/ True -> False.
Proof.
destruct 1.
Show Proof.

(fun H : True \/ True =>
 match H with
 | or_introl x => (fun H0 : True => ?Goal@{H:=H0}) x
 | or_intror x => (fun H0 : True => ?Goal0@{H:=H0}) x
 end)
```
I don't know where it comes from, this requires more in-depth debugging. It breaks an output test in the test-suite, so I deactivated it temporarily to see how much this PR affects the CI.

ppedrot:
Many of the CI devs are failing because mathcomp is broken in a way that is fixed by a one-liner. Due to the change of representation, occurrences in the branches and return clauses are not taken anymore into account. This breaks exactly one rewrite in the whole mathcomp library. I don't know whether this should be fixed there or whether we should revert the fix for #3166. @gares?

gares:
Open an issue on MC, I imagine the fix can be backward compatible.

ggonthier:
I've created PR math-comp/math-comp#298 to deal with this; it would be good to check if there are no further errors of this kind, before merging.
  Pointing out the obvious, `ssreflect` users are somewhat better armed to deal with this, in particular because they can usually (and are encouraged to) replace numeric selectors with contextual patterns - indeed that very issue of counting invisible occurrences prompted the development of the contextual pattern feature. The offending proof in `fintype` may be one of the few that got away...
  Other users may not be so lucky, so you might want to issue a call to test outside the CI base. Users that have painstakingly bisected to find they needed to select occurrence 33 out of 67 may be irked at a change that makes that 29 out of 57...

maximedenes:
After discussing with @ggonthier, I'm wondering what the motivation for this change is. CEP 34 mentions the discrepancy between user and internal syntax, which could be addressed by improving the user syntax instead, and not touching the kernel.

ppedrot:
@maximedenes I have several motivations for this PR, the major one being performance.

Storing redundant information in the proof term can lead to serious blowups, especially because of inductive parameters being duplicated in a lot of places. For instance, for sigma types, they appear twice as many times as needed, once in the return clause as arguments applied to the inductive type, the other time in the single match branch. This has observable consequences, e.g. for hashconsing. See the timing diffs of CoRN for instance. (N.B. : timings not yet optimal, there are several places where we could get rid of the on-the-fly expansion and contraction of type annotations, most notably in kernel type-checking).

The rest of the motivations can be summarized by "cleanliness". I find the current case representation to be really error-prone, both from the point of view of the kernel TCB and from the point of view of the upper layers.
- For the kernel, we have a bunch of alleged invariants that are not enforced, like the static η-expansion of branches. Sprinkle that with the ability to have let-bindings in both inductive arities and constructor types, and this is a recipe for soundness disaster.
- For the upper layers, we have a crapload of bugs occurring due to a too liberal representation, most notably when relying on occurrences, but not only. Often `simpl` explodes mid-flight because it starts unfolding types hidden in the return clause or in the branch, without any way to neither see it (which could be fixed with a better printer) nor control it (which is harder to fix). I remember a similar issue with `ring` calling its handrolled variant of `lazy` on a type that unfolds to a match and going bonkers, but I can't find it right now.

Relying on casts to work around this kind of foundational defect looks like a big no-no to me. Casts have no real status in proof terms, they're but an annotation that is often erased or changed in an uncontrolled fashion. Furthermore, the current situation is very strange: why would we add redundant information in case nodes, while we could also do the same for constants and, say, store their types in the `Const` node while we're at it?

gares:
I'm just curios: how is a proof term generated by a case/destruct that does *not* introduce names + a subsequent intro?
I imagine the first step generates `?p = match ... with ?e args end` and the second one `?e =fun args' : tys => ?b`. So unless one beta reduces eagerly the lambdas, their types are there nonetheless. In a sense the new representation seems to impose an eager introduction in the branches, and if you don't want to do that eager introduction you have to do a revert (eg eta expansion). Am I missing something? I did not look at the code: do I need to do something in order to "optimize" `case:` and `elim:`?

SkySkimmer:
Before this PR, in the user syntax (which is also what we talk about whenever we talk about the theory of Coq) you see `match discr as x in Ind _ (* <- params *) indices return P with C _ (* <- params *) args => body end` but in the ocaml code it's more like `match discr return fun (indices:index_type) (x:Ind params index_type) => P with C => fun (args:args_type) => body end` (also the order of discr and P is inverted so naive folds go in the wrong order, but that's unchanged in the PR)
Then doing something like `set (fun x => P)` is supposed to be forbidden, and either we detect that and error the tactic, or we miss it and get into a bugged state, with kernel errors, unprintability, etc.

After this PR the internal representation is more aligned with the user one (although the internal representation also carries params and universes, so maybe we should print `in Ind@{univs} params indices` instead of `in Ind _ indices`).

ppedrot:
@gares Indeed, as mentioned in a comment above, in its current state the PR introduces dummy β-redexes when calling `destruct`-like tactics. I think you were around at that time, but FTR with @herbelin we discussed about this behaviour in Rungis (... for lack of a better thing to do) and IIUC it is non-trivial to get rid of these additional cuts.

There are several entangled reasons for that, but in general getting rid of them also requires being smart about some commutative cuts that can be introduced via complex intro-patterns, and that break *a priori* the η-expansion invariant in `master`. I hope being able to fix this properly at some point, but this will need a bit of serious pondering. As for the ssr variants, I don't know which code they're based on, so I cannot comment on the proper course of events.

gares:
Maybe you can just pre process the proof term before qed since these redexs are easy to find, they are just after a match.

JasonGross:
Alternatively, can you make it so that `destruct ` is more primitive than `case `, and have `case ` be equivalent to `destruct in |- * ; revert `?  Then `destruct ` will not introduce β-redexes, and I think it is fine to say that `intros` always results in a λ?

ppedrot:
@JasonGross that's the idea, but the execution is not that easy as far as I can judge from the code.

herbelin:
@ppedrot: was the problem of beta-expansion blocking this PR in practice? Or can it be treated in a second step?

ppedrot:
@herbelin it introduces backwards incompatibility in a few developments, so this should be fixed before merging.

herbelin:
> it introduces backwards incompatibility in a few developments, so this should be fixed before merging.

Maybe could we plan a visio to work on it, say next week (??), and try to address the beta-expansion issue (I agree that `case` should be redefined from `destruct`)? Also, the azure logs have apparently been removed, so a rebase could be useful. 

The Coq-level incompatibilities are in bedrock2, fiat-crypto, fiat-parsers, and bugs/opened/bug_3396.v, is that right? Related to beta-expansion or to changes in occurrence numbering?

ppedrot:
AFAIR I fixed all the renumbering issues, so we're left with spurious β-expansions. We can discuss this next week indeed.

herbelin:
@ppedrot: 

Here is a preliminary proposal to change the introduction pattern engine, `case`, `destruct` and their variants so that they do not introduce beta-expansions:
- make a variant of `build_case_analysis_scheme` with `match` branches of the form `?x` (with typing `y1..yn : cstr_ctxt |- ?x : P (cstr y1..yn)` rather than of some form `f y1..yn` as of today; this variant shall take the names `y1`..`yn` as arguments;
- or, alternatively, simply bypass `build_case_analysis_scheme`
- change `general_case_analysis_in_context` so that it returns goals with the pattern variables in the context; this requires passing to it the expected names of the variables;
- add a revert in `general_case_analysis`, `simplest_ecase`, after they have called `general_case_analysis_in_context` (so as to simulate current variants of `case`); the ugly thing is that the variables, once requantified, need to have the same name as they would have before; I'm afraid some incompatibility may arrive here; at worse, we shall require a variant of `revert` which tells the name the variables should have once requantified;
- change `get_and_check_or_and_pattern` in `intro_or_and_pattern` to pre-compute introduction names for the `y1`..`yn`; I suspect that the `make_tmp_naming` strategy will work, but this has to be double-checked; then `get_and_check_or_and_pattern` will return `intro_action_pattern`s rather than `intro_pattern`s;
- the tricky part: now, the function `intro_pattern_action` does not necessarily apply to the last introduced hypothesis of the context; so a generalization might be needed; how to do the generalization over `y_i`? generalizing over `yi+1..yn` systematically or recomputing (with some cost) the minimal dependent closure?
- an advantage of recomputing the generalization would be that the `[...]` intropattern exactly implements `destruct`, what some users are asking for; is the extra cost in computing dependencies worth?

Side questions:
- How much are we ready to change the structure of `named_context`? For instance:
  - to split it into a map with log access and an associated ordering?
  - maybe to add a cache of the computation of (a superset of) what variables a given variable depends on (for quicker computation of dependencies in `destruct`)?
  - and, specially to deal with intro-patterns, to treat specially the set of `y_1`..`y_n` coming from a `match`: treating them as `Rel` until they are named? collecting dependencies for efficiency?

- One way wonder why we use by default a `_rect` scheme for induction and a `match` for case. Besides the compatibility consequences of a change, what would be the consequences of using a `fix f x := match` for `_rect` or, conversely, to use a `_case` for case analysis? One question is printing: `_rect` is a bit difficult to parse, but, maybe, this can just be addressed by using a specific rule? Another question is compactness of representation. Which one of `_rect` or `fix match` is more compact at the end?

What do you think? This would require some time, so I think it is better to have a good idea in advance about where we are going.

Side implementation question: does this interact with other works in progress? E.g. use of the new proof engine for `general_case_analysis_in_context`?

JasonGross:
> the ugly thing is that the variables, once requantified, need to have the same name as they would have before; I'm afraid some incompatibility may arrive here; at worse, we shall require a variant of revert which tells the name the variables should have once requantified;

As I understand it, the only case where `revert x; intro` changes the name of `x` is when there is a constant defined in the current module (?) with the name `x`.  However, these are exactly the times when `destruct`/`case` already gives a fresh name to relevant variable (i.e. as far as I'm aware, `destruct` will never pick a name automatically that shadows a name in the current module, and `case` will not let the user give names ahead of time).  Is my analysis right?

herbelin:
> Is my analysis right?

Probably yes. I'll check.

herbelin:
Hi,

> How much are we ready to change the structure of `named_context`

Solving the beta-expansion issue could be the opportunity to implement also this variant of `destruct` advocated by some users for many years:
- inform explicitly (and w/o hack) about which case of a case analysis we are currently working one
- give the pattern names at the time of solving the subgoal rather than using an `as` clause at the time of `destruct`.

Technically, this would mean to add a field of type `cases_pattern` to existential variables to remember to which case of a case analysis a subgoal corresponds. Then, it is at the name of addressing the subgoal that the pattern-matching name are set.

Concretely, the model I have in mind looks like this:
```
destruct n.
- [case (S n')] tactic.
- [case 0] tactic.
```
where, `destruct n` produces an evar for the case `S` and one for the case `0`. The first evar comes with pattern `0` and the second one with pattern `S n0` (for some auto-generated name `n0`).

At the time of writing `- [case (S n')]` the subgoal tagged with `S` is focussed and the `n0` is updated to `n'` as the user decides. We also remove the corresponding pending `cases_pattern` in the evar.

At the time of showing goals after `destruct n`, we would have the information available to be able to display in which case we are, as e.g. in:
```
2 subgoals
______________________________________(case 0)
0 = 0
______________________________________(case S)
S n = 0
```

To extend this to nested `destruct`, we may use a `cases_pattern list` instead of a `cases_pattern`. We then expect the focus on a goal to be e.g. given under the form `- [case (S n), (S n')]`.

To deal with `induction`, we would have an extra field for collecting the names of the induction hypotheses. [We would also need to find an extension of the `[case (S n)]` notation for indicating the name of induction hypotheses.]

For compatibility, we could still accept the `as` clause.

The pattern used for focussing could be arbitrary complex, e.g. `- [case (S n :: 0 :: l)]`, doing some form of case-split of the goal on the fly.

What do you think?

herbelin:
What about merging this right after the 8.2 branching? We can live with the spurious β-expansions for a while and it will be easier to take care of the β-expansions directly on master.

The alternative is that I turn my comments above into a CEP which we can then discuss?

Any opinion?

ppedrot:
Closing this for the time being.

PR #9711, opened 2019-03-07
Allow projections to be declared as classes.

### LINKED ISSUE #8994: Projections can be made classes but you can't define instances
#### Version

```shell
$ coqtop -v
The Coq Proof Assistant, version 8.8.2 (November 2018)
compiled on Nov 7 2018 21:54:20 with OCaml 4.07.0
```

#### Operating system

Ubuntu 18.04

```coq
Set Primitive Projections.

Class type (t : Type) : Type :=
  { bar : t -> Prop  }.

Instance type_nat : type nat :=
{ bar := fun _ => True }.

Existing Class bar.

Global Instance foo_bar {n : nat} : bar n := I.
(* with primitive projections, this line fails with "Error: bar is not a declared type class."
 * without primitive projections it succeeds
 *)
```

Expected behavior: `bar` should be a typeclass so there shouldn't be a problem with the definition of `foo_bar` as an instance.
If there is a reason to forbid supporting this, then there should be an error on the call to `Existing Class bar`.

### LINKED ISSUE #12975: Primitive projections breaks ability to declare field of record as Existing Class

Setting primitive projections breaks the ability to declare a field of a record as an `Existing Class`. Observe that this works on a record without Primitive Projections, but on a record with, it doesn't have the same behaviour. Coq just doesn't see that `myotherprop` is a class even though it has been declared so using `Existing Class`. 

```coq
Record foo :=
{
  myprop : Type ;
}.

Set Primitive Projections.

Record bar :=
{
  myotherprop : Type ;
}.


Existing Class myprop.
Existing Class myotherprop.

Instance : forall O, myprop O. Abort.
Fail Instance : forall (O : bar), myotherprop O.
```

#### Coq Version

Coq 8.12.0
Coq 8.13+alpha (September 2020)

mattam82:
**Kind:** feature

This fixes an arbitrary limitation of typeclasses, which didn't allow primitive projections
to be declared as class heads. They can usefully be used as proxys. In our use case:

```coq

Class UR A B := {
  ur : A -> B -> Type 
}.
```

We want to declare `ur` as a type class head, as in all instances of `UR`, `ur` is itself defined by a type-class. We can use a hint to unfold the `ur` projection when applied to a particular `UR` instance to simplify the goal to a new typeclass constraint. This allows a kind of `dependent` type-class resolution. 

As long as users did not try to do `Existing Class ur` for some primitive projection `ur`, which had no effect previously, this should be backwards compatible.

Fix #12975, fix #8994.

### DIFF
--- a/doc/changelog/02-specification-language/9711-primitive-projection-classes.rst
+++ b/doc/changelog/02-specification-language/9711-primitive-projection-classes.rst
@@ -1,0 +1,4 @@
+- **Added:**
+  Ability to declare primitive projections as class, for dependent typeclass resolutions
+  (fixes `#12975 <https://github.com/coq/coq/issues/12975>`_, `#9711 <https://github.com/coq/coq/pull/9711>`_,
+  by Matthieu Sozeau).
--- a/pretyping/typeclasses.ml
+++ b/pretyping/typeclasses.ml
@@ -107,14 +107,22 @@ let global_class_of_constr env sigma c =
     GlobRef.Map.find gr !classes, u
   with DestKO | Not_found -> not_a_class env sigma c
 
+let decompose_class_app env sigma c =
+  let hd, args = EConstr.decompose_app sigma c in
+  match EConstr.kind sigma hd with
+  | Proj (p, c) ->
+    let expp = Retyping.expand_projection env sigma p c args in
+    EConstr.decompose_app sigma expp
+  | _ -> hd, args
+
 let dest_class_app env sigma c =
-  let cl, args = EConstr.decompose_app sigma c in
+  let cl, args = decompose_class_app env sigma c in
     global_class_of_constr env sigma cl, (List.map EConstr.Unsafe.to_constr args)
 
 let dest_class_arity env sigma c =
   let open EConstr in
   let rels, c = decompose_prod_assum sigma c in
-    rels, dest_class_app env sigma c
+    rels, dest_class_app (push_rel_context rels env) sigma c
 
 let class_of_constr env sigma c =
   try Some (dest_class_arity env sigma c)
@@ -130,15 +138,11 @@ let rec is_class_type evd c =
     match EConstr.kind evd c with
     | Prod (_, _, t) -> is_class_type evd t
     | Cast (t, _, _) -> is_class_type evd t
+    | Proj (p, c) -> GlobRef.(Map.mem (ConstRef (Projection.constant p))) !classes
     | _ -> is_class_constr evd c
 
 let is_class_evar evd evi =
   is_class_type evd evi.Evd.evar_concl
-
-let is_class_constr sigma c =
-  try let gr, u = EConstr.destRef sigma c in
-    GlobRef.Map.mem gr !classes
-  with DestKO | Not_found -> false
 
 let rec is_maybe_class_type evd c =
   let c, _ = Termops.decompose_app_vect evd c in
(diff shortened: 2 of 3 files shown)


SkySkimmer:
@mattam82 is this PR dead?

mattam82:
It's been forgotten, I'll try to bring it back to life

ppedrot:
@mattam82 ping. I believe this is quite high priority actually.

mattam82:
@ppedrot I’m on holidays next week, so if you want to pick it up you’re welcome. Otherwise thanks for the reminder !

mattam82:
Rebased now, let's see.

mattam82:
@ppedrot this is ready now.

pretyping/typeclasses.ml
```diff
@@ -130,22 +138,19 @@ let rec is_class_type evd c =
     match EConstr.kind evd c with
     | Prod (_, _, t) -> is_class_type evd t
     | Cast (t, _, _) -> is_class_type evd t
+    | Proj (p, c) -> GlobRef.(Map.mem (ConstRef (Projection.constant p))) !classes
```
ppedrot, line 141: This is adding a new dependency to the legacy constant binding mechanism. I know we're a bit in a hurry w.r.t. the release, but instead of doing that we should have a type of global references including projections. I can validate the PR as is, but if you have time, could you do that the right way from the start?

mattam82: Hmm, I don't really have time right now, but indeed that should happen next. Note that we already had code using `Projection.constant` here, so it's not really worse than before.

doc/changelog/02-specification-language/9711-primitive-projection-classes.rst
```diff
@@ -0,0 +1,4 @@
+- **Added:**
+  Ability to declare primitive projections as class, for dependent typeclass resolutions
+  (fixes `#12975 <https://github.com/coq/coq/issues/12975`_, `#9711 <https://github.com/coq/coq/pull/9711>`_,
```
ppedrot, line 3: The first link is missing a closing chevron, leading to a CI failure.

ppedrot:
@mattam82 can you squash the last commit? It's useless to keep it.

ppedrot:
Erm, I would gladly have merged this PR without the additional comments as soon as the original CI would have finished... This needs to be merged before the branch, so you'll have to rebase again to remove the commits once your experiment is finished.

mattam82:
That's fine. I'm not sure this is the end just yet.

ppedrot:
Why is coqbot repeating the very same thing twice?

JasonGross:
It'll report again every time the overall pipeline status check is finished.  If a job fails and then someone restarts the job, when it finishes the second time, this will trigger a second comment.  Perhaps that's what happened?  (The coq-tools issue is on my side, I'll fix it shortly, sorry.)

mattam82:
What is coq-tools ?

mattam82:
Yes, I did retrigger it.

mattam82:
Do you main that the failure of the ci-coq_tools job is your fault?

JasonGross:
> Do you mean that the failure of the ci-coq_tools job is your fault?

Yes, I pushed what I thought was an innocuous commit working around an intermittent issue with the native compiler, not realizing that it would change the output log, and I didn't enforce the discipline of running the test-suite before pushing.  This broke about half of the output tests, which I'm now fixing; it should be fixed within 10 minutes, and sorry about this.

mattam82:
That's good news, both PRs should go green by tonight then!

JasonGross:
coq-tools should be fixed now, I've restarted it

ppedrot:
... still failing it seems.

JasonGross:
Yeah, I forgot that `-native-compiler ondemand` is required to work around native compiler issues on old versions of coq but produces warnings on new versions of Coq.  Hopefully now fixed

ppedrot:
@mattam82 in order not to lose time I have rebased and slightly tweaked the commit message, I'll merge when the CI finishes.

ppedrot (approved):

mattam82:
@ppedrot thanks a whole bunch!

PR #9720, opened 2019-03-08
[refman] Fix typo in local application of tactics

hawnzug:
**Kind:** documentation

The index of local tactic application `[> expr|* ]` should start from 1. It seems that this typo originates from the old doc for `expr ; [ expr|* ]`.


Zimmi48 (approved):
Thanks!

SkySkimmer:
Nit:
>[sphinx]

This doesn't have anything to do with the sphinx system AFAICT (if we were still using latex the PR would be the same), [doc] would be more appropriate.

Zimmi48:
> > 
> 
> This doesn't have anything to do with the sphinx system AFAICT (if we were still using latex the PR would be the same), [doc] would be more appropriate.

I'll edit the PR title, but don't bother changing the commit message.

Issue #9722, opened 2019-03-08
Windows packaging job failing consistently


Zimmi48:
```
+13:13:17 cd lablgtk-2.18.6
+13:13:17 mkdir bin_special
++13:13:17 pwd
+13:13:17 PATH=/build/lablgtk-2.18.6/bin_special:/usr/local/bin:/usr/bin:/cygdrive/c/ci/coq64_18771_13803/bin:/usr/x86_64-w64-mingw32/sys-root/mingw/bin:/cygdrive/c/Windows/system32:/cygdrive/c/Windows
+13:13:17 return 0
+13:13:17 cp /bin/x86_64-w64-mingw32-pkg-config.exe bin_special/pkg-config.exe
cp: cannot stat '/bin/x86_64-w64-mingw32-pkg-config.exe': No such file or directory
ERROR MakeCoq_MinGW.bat failed
```

e.g. in https://gitlab.com/coq/coq/-/jobs/174132485

MSoegtropIMC:
Thanks for reporting, I will fix it asap!

MSoegtropIMC:
This is caused by a package change in cygwin and needs to be back ported to all active branches. Shall I create a separate PR for 8.9?

Zimmi48:
@silene might be able to answer this. Generally speaking, small changes are easy to backport automatically, as soon as the PR is put in the right milestone.

PR #9728, opened 2019-03-08
Fix issue #9722 pkg-config not found

MSoegtropIMC:
This fixes issue #9722 that pkg-config is not found in Windows CI.
The root cause was change in cygwin - the mingw variant of pkg-config has been replaced by a shell script which calls a generic version of pkg-config.

This needs to be back merged to all active branches!

### DIFF
--- a/dev/build/windows/MakeCoq_MinGW.bat
+++ b/dev/build/windows/MakeCoq_MinGW.bat
@@ -373,7 +373,8 @@ IF "%RUNSETUP%"=="Y" (
     -P make,unzip ^
     -P gdb,liblzma5 ^
     -P patch,automake1.14 ^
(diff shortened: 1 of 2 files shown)


MSoegtropIMC:
P.S.: I checked in a local build that the resulting CoqIDE actually works and links with MinGW and not with Cygwin versions of GTK.

It is worth discussing if it would be a better fix to change the lablgtk configure script such that it picks up the correct version of pkg-config in a cross build, so that one doesn't need to hack around this. If the lablgtk configure script would do the right thing, our CI would have survived the change in Cygwin.

I guess it is best to fix it now as suggested in this PR and do a follow up PR to lablgtk.

MSoegtropIMC:
The Windows64 build in gitlab passed again, so please can someone fix this asap to restore the Windows CI?

I will create separate PRs for 8.9 and 8.10.

Zimmi48 (approved):
It works :tada: Thanks.

Zimmi48:
Putting in the 8.10 milestone since you have prepared a PR for v8.9.

PR #9736, opened 2019-03-11
Don't coqchk the test suite prerequisites

SkySkimmer:
This causes the makefile to break due to dependencies when it fails,
and it's not worth adding a whole mess of code to catch the failure
for these files.


gares:
> This causes the makefile to break due to dependencies when it fails

I'm a bit lost here. How can it fail to check prerequisites.v but have no problem in running the tests?

SkySkimmer:
We try to run all the tests before reporting errors, this issue prevents that (most of the tests don't depend on prerequisites).
Also it prevents logs from being visible on CI so we can't tell why prerequisites failed.

gares (approved):

PR #9738, opened 2019-03-11
[ci] [docker] Upgrade odoc to 1.4.0

ejgallego:
This is routine as for API doc writers to be able to access the newer
features.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,7 +10,7 @@ stages:
 variables:
   # Format: $IMAGE-V$DATE [Cache is not used as of today but kept here
   # for reference]
-  CACHEKEY: "bionic_coq-V2019-02-17-V1"
+  CACHEKEY: "bionic_coq-V2019-03-11-V1"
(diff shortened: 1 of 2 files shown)


SkySkimmer:
Should we do any other updates while we're at it? eg dune

ejgallego:
> Should we do any other updates while we're at it? eg dune

hehehe I was indeed going to do it but I thought you would tell me "do in a different PR" :)

Given that we still set `(lang dune 1.6)` let's keep using 1.6 for a bit more; we shall increase the version number to 1.9 I think next time, which will include quite a few nice features for Coq.

SkySkimmer:
>hehehe I was indeed going to do it but I thought you would tell me "do in a different PR" :)

That's pretty easy to avoid, just name the PR "[docker] update opam packages" or some such less specific thing ;)

SkySkimmer (approved):

ejgallego:
Seems to work fine, https://coq.gitlab.io/-/coq/-/jobs/175353544/artifacts/_build/default/_doc/_html/index.html

PR #9743, opened 2019-03-11
Relax the ambiguous path condition of coercion

### LINKED ISSUE #3219: Coercions should not report ambiguous paths in cases where the paths are judgmentally equal

Original bug ID: BZ#3219
From: @JasonGross
Reported version: 8.4

pi8027:
**Kind:** feature.

Fixes / closes #3219

The `Coercion` command did report many ambiguous paths when one declared multiple inheritances.  This change makes the `Coercion` command to do not report them when

1. all the coercion in the potentially ambiguous paths respect the uniform inheritance condition and
2. functional compositions of the potentially ambiguous paths are convertible to each other.

The first condition is not explicitly checked but is used to make the checking process of the second condition easy.

The key idea: let us consider a sequence of coercion
```
f_1 : C_1 >-> C_2,
f_2 : C_2 >-> C_3,
...,
f_n : C_n >-> C_(n+1)
```
which respect the uniform inheritance condition and where the user-defined classes C_i have m_i parameters respectively (i <= n). The functional composition f_1 ∘ ... ∘ f_n can be expressed as follows:
```coq
(fun (x_1 : _) ... (x_(m_1) : _) (y : _) =>
  @f_n _ ... _ (* m_n times repetition of holes *)
      (...
           (@f_2 _ ... _ (* m_2 times repetition of holes *)
                (@f_1 x_1 ... x_(m_1) y))...)),
```
and the contents of all the holes can be determined (inferred) without leaving any existential variables in them thanks to the uniform inheritance condition.

- I needed to put a ref `path_comparator` to avoid a circular dependency between `classops` and `pretyping`. But I guess splitting `classops` into 2 files is a better way to implement this.
- I don't know how to systematically test this kind of warnings in Coq.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -140,6 +140,9 @@ Vernacular commands
 
 - `Hypotheses` and `Variables` can now take implicit binders inside sections.
 
+- `Coercion` does not warn ambiguous paths which are obviously convertible with
+  existing ones.
+
 Tools
 
 - The `-native-compiler` flag of `coqc` and `coqtop` now takes an argument which can have three values:
--- a/pretyping/classops.ml
+++ b/pretyping/classops.ml
@@ -308,9 +308,16 @@ let install_path_printer f = path_printer := f
 
 let print_path x = !path_printer x
 
-let message_ambig l =
-  str"Ambiguous paths:" ++ spc () ++
-  prlist_with_sep fnl print_path l
+let path_comparator : (inheritance_path -> inheritance_path -> bool) ref =
+  ref (fun _ _ -> false)
+
+let install_path_comparator f = path_comparator := f
+
+let compare_path p q = !path_comparator p q
+
+let warn_ambiguous_path =
+  CWarnings.create ~name:"ambiguous-paths" ~category:"typechecker"
+    (fun l -> strbrk"Ambiguous paths: " ++ prlist_with_sep fnl print_path l)
 
 (* add_coercion_in_graph : coe_index * cl_index * cl_index -> unit
                          coercion,source,target *)
@@ -329,21 +336,15 @@ let add_coercion_in_graph (ic,source,target) =
   let ambig_paths =
     (ref [] : ((cl_index * cl_index) * inheritance_path) list ref) in
   let try_add_new_path (i,j as ij) p =
-    try
-      if Bijint.Index.equal i j then begin
-	if different_class_params i then begin
-	  let _ = lookup_path_between_class ij in
-          ambig_paths := (ij,p)::!ambig_paths
-	end
-      end else begin
-        let _ = lookup_path_between_class ij in
-        ambig_paths := (ij,p)::!ambig_paths
-      end;
+    if not (Bijint.Index.equal i j) || different_class_params i then
+      match lookup_path_between_class ij with
+      | q ->
+        if not (compare_path p q) then
+          ambig_paths := (ij,p)::!ambig_paths;
+        false
+      | exception Not_found -> (add_new_path ij p; true)
+    else
       false
-    with Not_found -> begin
-      add_new_path ij p;
-      true
-    end
   in
   let try_add_new_path1 ij p =
     let _ = try_add_new_path ij p in ()
@@ -364,9 +365,7 @@ let add_coercion_in_graph (ic,source,target) =
 	 end)
       old_inheritance_graph
   end;
-  let is_ambig = match !ambig_paths with [] -> false | _ -> true in
-  if is_ambig && not !Flags.quiet then
-    Feedback.msg_info (message_ambig !ambig_paths)
+  match !ambig_paths with [] -> () | _ -> warn_ambiguous_path !ambig_paths
 
 type coercion = {
   coercion_type   : coe_typ;
--- a/pretyping/pretyping.ml
+++ b/pretyping/pretyping.ml
@@ -1128,3 +1128,39 @@ let understand_tcc ?flags env sigma ?expected_type c =
 let understand_ltac flags env sigma lvar kind c =
   let (sigma, c, _) = ise_pretype_gen flags env sigma lvar kind c in
   (sigma, c)
+
+let path_convertible p q =
+  let open Classops in
+  let mkGRef ref          = DAst.make @@ Glob_term.GRef(ref,None) in
+  let mkGVar id           = DAst.make @@ Glob_term.GVar(id) in
+  let mkGApp(rt,rtl)      = DAst.make @@ Glob_term.GApp(rt,rtl) in
+  let mkGLambda(n,t,b)    = DAst.make @@ Glob_term.GLambda(n,Decl_kinds.Explicit,t,b) in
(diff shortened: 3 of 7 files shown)


gares:
> I don't know how to systematically test this kind of warnings in Coq.

The `test-suite/output/` directory contains tests for which an expected output is tested.
Alternatively you could put a test in `misc/` and turn warnings into errors there.

It would be interesting to know if the refined warning still fires in the standard library (and eventually in CI)

pi8027:
It was not possible to detect only the inconvertible ambiguous paths in the MathComp library because there was a bunch of "ambiguous paths" messages which are almost meaningless. This change makes this possible and I found 19 ambiguous paths.

pretyping/pretyping.ml
```diff
@@ -1128,3 +1128,39 @@ let understand_tcc ?flags env sigma ?expected_type c =
...
+  in
+  try
+    let e = Global.env () in
+    let m,tp = understand_tcc e (Evd.from_env e) (path_to_gterm p) in
+    let m',tq = understand_tcc e m (path_to_gterm q) in
+    if Evd.has_undefined m' then
+      false
+    else
+      let _ = Evarconv.the_conv_x e tp tq m' in true
+  with Evarconv.UnableToUnify _ | PretypeError _ | Not_found -> false
```
gares: The `Not_found` here is a bit fishy. When can it occur?

pretyping/pretyping.ml
```diff
@@ -1128,3 +1128,39 @@ let understand_tcc ?flags env sigma ?expected_type c =
...
+             mkGApp (mkGRef ic.coe_value,
+                     List.make ic.coe_param (mkGHole ()) @ [t]))
+          (mkGApp (mkGRef ic.coe_value, List.map (fun i -> mkGVar i) names))
+          p'
+    | [] -> anomaly (str "A coercion path shouldn't be empty.")
+  in
+  try
+    let e = Global.env () in
+    let m,tp = understand_tcc e (Evd.from_env e) (path_to_gterm p) in
+    let m',tq = understand_tcc e m (path_to_gterm q) in
```
gares: We call the evar_map either `sigma` or `evd` and we (strongly) advise to mask any previous copy of it (avoid the `'`, since using later on the old copy is very likely to be wrong).

gares:
> I found 19 ambiguous paths.

This is really nice, and a bugreport should go to the math-comp repo.

For the test suite of Coq, I was suggesting to add a silly test, just to be sure your code is does not print the warning in some cases.

gares:
> This is really nice, and a bugreport should go to the math-comp repo.

BTW, conversion is pretty dumb, since `f := fun x y t => t` and `g := fun y x t` are different. But `forall a b, f a b t = g b a t` may hold... I wonder how many "false positive" are still there.

ggonthier:
I suspect there may be a few false positives hinging on record eta.

pretyping/pretyping.ml
pi8027: I think this `try` clause never raises the `Not_found` exception. But when it happens (by a bug?, `understand_tcc` may raise the `Not_found`) an existing coercion path will be overridden by a `try` clause in `add_coercion_in_graph`. Should I put `Not_found -> anomaly ...` here, or rewrite `add_coercion_in_graph` to do not call `path_convertible` in a `try` clause?

pretyping/classops.ml
```diff
@@ -329,21 +336,17 @@ let add_coercion_in_graph (ic,source,target) =
...
-        let _ = lookup_path_between_class ij in
-        ambig_paths := (ij,p)::!ambig_paths
-      end;
+    if not (Bijint.Index.equal i j) || different_class_params i then
+      try
+        let q = lookup_path_between_class ij in
+        if not (path_convertible p q) then
+          ambig_paths := (ij,p)::!ambig_paths;
+        false
+      with Not_found -> begin
```
SkySkimmer: Do 
~~~ocaml
match lookup_path_between_class ij with 
| q -> if not (path_convertible p q) ...
| exception Not_found -> (add_new_path uj p; true)
~~~
that way you won't catch arbitrary exceptions from the path comparator.

pi8027: I didn't know this syntax. It's nice!

pi8027:
> BTW, conversion is pretty dumb, since f := fun x y t => t and g := fun y x t are different. But forall a b, f a b t = g b a t may hold... I wonder how many "false positive" are still there.

Such false positive examples do not hold the uniform inheritance condition. https://coq.inria.fr/refman/addendum/implicit-coercions.html#coercions

> I suspect there may be a few false positives hinging on record eta.

`Set Primitive Projections` would solve that if one really need it.

pretyping/pretyping.ml
```diff
@@ -1128,3 +1128,39 @@ let understand_tcc ?flags env sigma ?expected_type c =
...
+          (fun t ic ->
+             mkGApp (mkGRef ic.coe_value,
+                     List.make ic.coe_param (mkGHole ()) @ [t]))
+          (mkGApp (mkGRef ic.coe_value, List.map (fun i -> mkGVar i) names))
+          p'
+    | [] -> anomaly (str "A coercion path shouldn't be empty.")
+  in
+  try
+    let e = Global.env () in
+    let m,tp = understand_tcc e (Evd.from_env e) (path_to_gterm p) in
```
SkySkimmer: sigma, not m

pretyping/pretyping.ml
gares: I prefer the anomaly

Zimmi48 (approved):

ejgallego:
Failure on CI is the usual spurious async one.

Any assignee here?

gares:
I can take it

gares:
As far as I can tell, all issues were addressed. Unless @mattam82 wants more time for reviewing, I'm going to merge this.

maximedenes:
Great work, @pi8027 !

CHANGES.md
```diff
@@ -140,6 +140,9 @@ Vernacular commands
 
 - `Hypotheses` and `Variables` can now take implicit binders inside sections.
 
+- `Coercion` does not warn ambiguous paths which are obviously convertible with
```
maximedenes: I'm afraid people will wonder what "obviously convertible" means. Should it just be "convertible"?

pi8027: Some inheritance paths which do not respect the uniform inheritance condition may be reported as ambiguous paths even if they are convertible with existing ones.

RalfJung:
See https://github.com/coq/coq/issues/10247: in our library with multiple inheritance, this PR actually lead to *more* ambiguous paths warnings.

CHANGES.md
pi8027: I have provided a better description of this change in #10322.

pi8027:
I found that the convertibility of ambiguous paths has been mentioned as the *coherence* of inheritance graphs in [Saïbi 1997] and [Barthe 1995]. However, I wasn't aware of the first condition of coherence, that is, circular inheritance paths should be convertible with the identity function. I will implement this missing part soon.
- [Saïbi 1997] https://dl.acm.org/citation.cfm?id=263742
- [Barthe 1995] https://link.springer.com/chapter/10.1007/3-540-61780-9_58

Issue #9727, opened 2019-03-08
[dune] missing dependency on plugins


gares:
```
gares@ollypat:~/COQ/wt/master$ dune exec coqtop.byte
Welcome to Coq ollypat:/home/gares/COQ/wt/master/_build/default,quotations (e1e1fc7ea327bf3536b00260c394d0a219e45dbb)
The files /home/gares/COQ/wt/master/_build/install/default/lib/coq/plugins/ltac/ltac_plugin.cma
and /home/gares/COQ/wt/master/_build/install/default/bin/coqtop.byte
disagree over interface Tok
Error: Could not load plugin ltac_plugin.cma
```
I did that to be able to `Drop` as per `dev/doc/build-system.dune.md` but I got this error.

gares:
I was indeed editing tok.ml, this is why `Tok` is there.

gares:
I did remove `_build` and now I get
```
gares@ollypat:~/COQ/wt/master$ dune exec coqtop.byte
Fatal error: cannot load shared library dllbyterun_stubs
Reason: dllbyterun_stubs.so: cannot open shared object file: No such file or directory
```

ejgallego:
This is a know problem, you need to use the dev/shim for now. Indeed coqtop doesnt depend on the plugins as it is an optional dependency (you can load without the prelude)

This will be fixed hopefully in the next iteration by having a command to specify the lib to load.

gares:
The first error seems to match your description, but the second one is different. The instructions to run Drop seem to be outdated.

ejgallego:
Indeed the second error I think is a bug in Dune, I was aware of it but didn't have time to investigate more, we should add a workaround for now.

ejgallego:
@rgrinberg , indeed, given a bytecode binary `a/foo`, `dune exec a/foo.bc` won't properly install the necessary stublibs for it to run. Should I file a bug?

rgrinberg:
Yeah, we're aware of this issue. The issue is that the stublibs aren't actually necessary to build the binary but to _run_ it. There's no way to express that runtime dependency in dune at the moment.

ejgallego:
> There's no way to express that runtime dependency in dune at the moment.

This seems like a bug in `dune exec`, right? In the same way `dune utop` works, exec should indeed generate the extra rules, right?

Is there an issue already for that?

rgrinberg:
> In the same way dune utop works, exec should indeed generate the extra rules, right?

[...]

PR #9748, opened 2019-03-12
[dune] Add shim for coqtop.byte

ejgallego:
We add a shim for running the byte version of coqtop. This fixes the
Coq part of #9727 , the Dune part is still open at
https://github.com/ocaml/dune/issues/108 but this PR includes a
workaround.

Unfortunately we have to introduce a small inefficiency in the build
process as we build both byte and native versions of plugins for this
work reliable.

As this is a choice done during bootstrap it won't be easy to fix
until we have our own `dune coqtop` command and we can control the
rules depending on the final target.

This should affect the `check` target so still fast builds should be
possible, but if this is a problem we could add a `byteboot` target to
help.

### DIFF
--- a/dev/doc/build-system.dune.md
+++ b/dev/doc/build-system.dune.md
@@ -126,9 +126,9 @@ script again] This will be fixed in the future.
 
 ## Dropping from coqtop:
 
-The following sequence is recommended:
+After doing `make -f Makefile.dune voboot`, the following commands should work:
 ```
-dune exec coqtop.byte
+dune exec dev/shim/coqbyte-prelude
 > Drop.
 # #directory "dev";;
 # #use "include_dune";;
--- a/dev/shim/dune
+++ b/dev/shim/dune
@@ -11,6 +11,19 @@
     (run chmod +x %{targets})))))
 
 (rule
+ (targets coqbyte-prelude)
+ (deps
+  %{bin:coqtop.byte}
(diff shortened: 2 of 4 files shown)


gares (approved):
I've no idea what this does, but I guess it works

ejgallego:
> I've no idea what this does, but I guess it works

Questions are welcome! It is not very hard actually, we do:
```
(rule
 (targets coqbyte-prelude)
 (deps
  %{bin:coqtop.byte}
  %{lib:coq.kernel:../../stublibs/dllbyterun_stubs.so}
  %{project_root}/theories/Init/Prelude.vo)
 (action
  (with-outputs-to %{targets}
   (progn
    (echo "#!/usr/bin/env bash\n")
    (bash "echo \"$(pwd)/%{bin:coqtop.byte} -coqlib $(pwd)/%{project_root}\" \"$@\"")
    (run chmod +x %{targets})))))
```
so we depend on coqtop.byte, on the stubs to workaround the dune bug, and on the prelude. The rest is due to problems with `coqdep` as dune builds plugins as `cma` and makefile as `cmo`.

We generate a file that calls coqtop byte with the right path, etc... as a convenience.

Zimmi48 (approved):

mattam82:
@ejgallego it works very well, I'm a dune convert now!

ejgallego:
Thanks for the feedback @mattam82

Issue #9751, opened 2019-03-12
printing whitespace change on master


andres-erbsen:
#### Description of the problem

https://github.com/mit-plv/bedrock2/commit/9b427dc4cea9731a7869eb7c68a1de95047dd7be

prints an additional newline and space on master ([log](https://travis-ci.com/mit-plv/bedrock2/jobs/184001042#L1616)). I cannot find any CHANGES entry to explain this, and it seems undesirable anyway.

#### Coq Version

660732f bad, 8.9.0 good. ~~@JasonGross you said the first bit would tell us the coq commit hash?~~

See also #9744.

andres-erbsen:
This was found by adding a test requiring the 8.9 behavior to bedrock2 CI and seeing it fail on master. Currently the test in question is disabled, but we would like to re-enable it.

JasonGross:
That's from 2019-02-08, which was before I changed the behavior to include the commit hash.  The current package is 8.master~git~201903120218+20181-0~daily319-74534f8-74534f8~ubuntu14.04.1 which is commit 74534f8.

andres-erbsen:
Ok, hash updated above.

andres-erbsen:
More explanation: the newly printed newline is just a printing width break, it is the space that needs to be explained.

ejgallego:
This should be a silly thing, likely due to the heavy refactoring we did for the toplevel.

If 8.9 was OK here I'm positive we can fix this. A bisect would greatly help.

andres-erbsen:
I am working on minimizing, then bisect.

andres-erbsen:
minimized

```coq
Delimit Scope test_scope with test.
Notation "a b" := (@cons bool a%test b%test) (only printing, right associativity, at level 3, format "a b") : test_scope.
Undelimit Scope test_scope.

Notation "'A'" := (false) (only printing) : test_scope.
Notation "  " := 42 (only printing) : test_scope.
Notation "a" := (@cons bool a nil) (only printing, at level 3).
Notation "" := (@nil bool) (only printing, right associativity, at level 3, format "") : test_scope.

Local Open Scope test_scope.
Goal True.
  let cc := constr:(cons false nil) in
  idtac cc.
Abort.
```

expected output `A\n`, master gives `A \n`.

andres-erbsen:
@ejgallego is there a recommended way of bisecting coq? e.g. should I add this as a test somehow, or something else?

ejgallego:
[...]

PR #9752, opened 2019-03-12
[ci] [docker] Add bsdmainutils to fix bedrock2 build.

ejgallego:


SkySkimmer:
V8?
Why is CI cancelled?

ejgallego:
> Why V8?

That's the current conflict-free number of docker images today.

ejgallego:
Umm, what's going on with bedrock @samuelgruetter ? I thought the problem was a missing binary but I see this still fails otherwise [and in master too]

samuelgruetter:
bedrock2 [succeeds on our own Travis CI](https://github.com/mit-plv/bedrock2/runs/77110044), but [fails on Coq's CI](https://gitlab.com/coq/coq/-/jobs/176113494), and I guess the reason is that the file `bedrock2/bedrock2/src/Bytedump.v` contains non-ascii-nor-unicode characters, which seems to work fine on our CI, but makes `./tools/make-one-time-file.py` fail on Coq's CI.
I would actually prefer to avoid non-ascii-nor-unicode characters in `.v` files, do you think that would be possible @andres-erbsen ?

andres-erbsen:
1) bedrock2 does not *right now* depend on `hexdump`, and if something is to be added for our use, let it already be `xxd` which can convert in both directions.
2) I don't want to take responsibility for the failure in `Aggregating timing log...`. As `cocq` accepts the file just fine, it is the timing script that should be fixed.
3) I don't like having non-utf8 files either, but I prefer having them to having a plugin that implements the same functionality.
4) We may backtrack on this experiment anyway because the current ridiculous hack also fails on master due to #9751 (which I think should be considered a real issue regardless of how ridiculous the use case that found it is). I think we also prefer having two separate implementations of the same printing for proofgeneral output and external output to having a plugin.

ejgallego:
> bedrock2 [succeeds on our own Travis CI](https://github.com/mit-plv/bedrock2/runs/77110044), but [fails on Coq's CI](https://gitlab.com/coq/coq/-/jobs/176113494),

Indeed I've noticed this, this is likely due to some missing Python libs? We could add them to the CI without a problem.

> 1. bedrock2 does not _right now_ depend on `hexdump`,

we saw a few build logs where the failure was "hexdump" cannot be found; so indeed I thought you folks had added a dependency on it.

Regarding output I'd say indeed let's implement a real output primitive such as fprintf if you have needs for it, your live will be much better as of today, where you are using the user messaging infrastructure which is by design not very stable.

andres-erbsen:
We did use hexdump, I removed it from default build when @SkySkimmer commented on my commit that it made coq CI fail.

ejgallego:
Ok, closing this PR then; will open an issue tomorrow if the CI is still giving trouble so we can track solutions.

JasonGross:
> Indeed I've noticed this, this is likely due to some missing Python libs? We could add them to the CI without a problem.

It is not due to missing python libs, it is due to the fact that in Python3 to input a file as text (which I believe must be done to do, e.g., regexp matching), AIUI, you have to tell Python what encoding that file uses.  I figured UTF8 was a fine encoding to pick.  I was not aware that coqc supports non-utf8 input; how does coqc decide what encoding to use for its input?

ejgallego:
> I was not aware that coqc supports non-utf8 input; how does coqc decide what encoding to use for its input?

As far as I know this is custom-made stuff in `clib/unicode` plus some other parsing rules.

SkySkimmer:
>I was not aware that coqc supports non-utf8 input; how does coqc decide what encoding to use for its input?

Notations do special processing on the notation string, AFAICT it only cares about unicode when detecting whether something is a valid identifier.

JasonGross:
Note that it is not an issue with the file *containing* non-UTF-8, it is an issue with files that try to print non-UTF-8 to the console.  @andres-erbsen @samuelgruetter Are you willing to redirect all non-UTF-8 printing to a file so that it doesn't show up in the log?  I think the alternative here is to disable the timing log for bedrock2 in the CI (just requires locally unsetting the relevant environment variables in the bedrock2 ci script).

Anyway, I've created https://github.com/coq/coq/issues/9767 to track this.  I expect to definitely not get to this before April, and may or may not find time for it after.  (I'm also not convinced that I'm the best person to do this; I've never used `bytes` in python, and would be mostly guessing and coding in the dark.)

PR #9754, opened 2019-03-12
Don't lose the warning name when warning becomes error.

SkySkimmer:


ejgallego (approved):
7 days, no action, kind of trivial, taking over as per policy.

PR #9757, opened 2019-03-12
[refman] Add 'warn' option to coqtop directive.

Zimmi48:
Fix semantic conflict between #9389 and #9654.

**Kind:** fix

I suggest the first available developer to self-assign and merge once CI has passed on this PR.

### DIFF
--- a/doc/sphinx/addendum/extended-pattern-matching.rst
+++ b/doc/sphinx/addendum/extended-pattern-matching.rst
@@ -285,7 +285,7 @@ By default, implicit arguments are omitted in patterns. So we write:
 
 .. coqtop:: all
 
-   Arguments nil [A].
+   Arguments nil {A}.
    Arguments cons [A] _ _.
    Check
      (fun l:List nat =>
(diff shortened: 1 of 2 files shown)


cpitclaudel:
LGTM, will merge when CI passes

cpitclaudel (approved):

Zimmi48:
So it turns out that this newly introduced warning occurred in several other places in the manual. Now fixed. I had to re-read the documentation about the topic and discovered in passing a regression from the Sphinx translation (now fixed too). This time I actually built the manual locally, so this should pass CI without trouble.

ejgallego:
Thanks @Zimmi48 , and sorry for the oversight, I didn't understood at all what the problem was.

Will steal from @cpitclaudel as this is urgent.

cpitclaudel:
Great job all. Thanks!

JasonGross:
Note that the grammar of `Arguments ` here is still incomplete; you can put multiple arguments in the same delimiters, e.g., `Arguments map [A B] f l.`

Zimmi48:
Thanks Jason. If someone can open a PR fixing this that would be welcome. Otherwise, I'll do it at some point in the future.

PR #9766, opened 2019-03-14
Don't use print for sys.stdout

### LINKED ISSUE #9705: coq-makefile: print-pretty-timed-diff broken
When trying to use the print-pretty-timed-diff target, there is some problem with the invoked python code (at least when using Python 3.7.2):

```
> make -f Makefile.coq print-pretty-timed-diff 
  File "/home/fakusb/.opam/coq-dev/lib/coq//tools/make-both-time-files.py", line 16, in <module>
    print_or_write_table(table, args[3:])
  File "/home/fakusb/.opam/coq-dev/lib/coq/tools/TimeFileMaker.py", line 215, in print_or_write_table
    print(table.encode("utf-8"), file=binary_stdout)
TypeError: a bytes-like object is required, not 'str'
make: *** [Makefile.coq:337: print-pretty-timed-diff] Error 1
```

this occurs both on coq 8.8.2 and master (at least in a build from 2 weeks ago)

JasonGross:
This should fix #9705

I'm kind-of cargo-cult coding here, from things like
https://docs.python.org/3/library/sys.html#sys.displayhook and
https://github.com/coq/coq/issues/9705#issuecomment-471996313, but
hopefully this fixes the issue without breaking anything.  (I am really
a novice when it comes to the str/bytes distinction in python3.)

**Kind:** bug fix

Fixes / closes #9705

- [ ] Added / updated test-suite (not sure what the right test-case is, possibly not worth testing?)

### DIFF
--- a/tools/TimeFileMaker.py
+++ b/tools/TimeFileMaker.py
@@ -209,11 +209,10 @@ def make_table_string(times_dict,
 
 def print_or_write_table(table, files):
     if len(files) == 0 or '-' in files:
-        try:
-            binary_stdout = sys.stdout.buffer
-        except AttributeError:
-            binary_stdout = sys.stdout
-        print(table.encode("utf-8"), file=binary_stdout)
+        if hasattr(sys.stdout, 'buffer'):
+            sys.stdout.buffer.write(table.encode("utf-8"))
+        else:
+            sys.stdout.write(table)
     for file_name in files:
         if file_name != '-':
             with open(file_name, 'w', encoding="utf-8") as f:


vbgl:
Is it related to the current failure of bedrock?
~~~
  File "/builds/coq/coq/tools/TimeFileMaker.py", line 47, in get_times
    lines = f.read()
  File "/usr/lib/python2.7/codecs.py", line 314, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf8' codec can't decode byte 0x80 in position 52285: invalid start byte
~~~

jashug (changes requested):
You should also encode in the plain stdout case. Other than that, looks good to me.
I suppose that print must have been sending a string to the buffer?
Using write directly, you will also lose the newline that print appends, so may want to `write(encode(table + "\n").

JasonGross:
@vbgl it is related only in the sense that both of these are issues with needing to carefully track encodings of bytes in python3 in a way that was not necessary in python2.  This change will not impact that failure; see also https://github.com/coq/coq/issues/9767

@jashug thanks for the pointer about the trailing newline.  I'm not so sure regarding encoding in the else case; can you explain why https://docs.python.org/3/library/sys.html#sys.displayhook explicitly decodes bytes to text before writing to sys.stdout in a way that's compatible with us having to encode here?  (My model is that sys.stdout.buffer is opened in wb mode and is for bytes-writing, while sys.stdout is opened in w mode and is for text writing.)

JasonGross:
I will fix the trailing newline issue when I'm next at a computer

jashug:
In Python 3, the displayhook decodes bytes to string before writing to `sys.stdout` because in Python 3 you are correct, `stdout` is for text writing while `stdout.buffer` is for text writing. Since we always want to output a UTF8 encoded bytestream, in Python 3 we can directly write to `stdout.buffer`.

In Python 2, `stdout.buffer` does not exist, but that is OK because we can write a byte stream directly to stdout in Python 2. We can also write a unicode object (which is the type of `table` in Python 2) to `stdout` and let `stdout` do the encoding for us, but this becomes dependent on what encoding the python system guesses for `stdout`. If we always want to output a UTF8 bytestream, we need to do the encoding ourselves.

The displayhook psuedocode you linked to is only for Python 3, and first writes the text as text to `stdout`, using whatever encoding is guessed. If that encoding fails (not all encodings support all unicode characters), the text is encoded with the same encoding but using a backslash escaping system to support unsupported characters. Then it writes that bytestream to `stdout.buffer`. If, however, someone has sneakily replaced `stdout` with some other file-like object which doesn't have a binary interface buffer, then we turn those bytes back into a string and write that as text to `stdout`, this time with any invalid characters replaced with backslashes. So the whole dance is to support the full range of unicode even with restrictive encodings and even after `stdout` has been replaced by a mere file.

In our case, we don't expect anyone to replace stdout from within python (but they should be able to redirect it, in which case the encoding python guesses for stdout may change), and we always want to use UTF8, so we don't have to worry about invalid characters. Instead, we do have to worry about Python 2 behavior, under which we still get the most consistent results by specifying the encoding ourselves rather than leaving it up to a guess.

fakusb:
Why do we even try to support both python2 and python3? We could just add `#!/usr/bin/env python3` to the called `.py`-file?

vbgl:
@JasonGross Thanks for the clarification.

JasonGross:
Who should be assignee on this?

PR #9770, opened 2019-03-14
Correct dependencies in the micromega pack

### LINKED ISSUE #9768: Dependencies in the Micromega pack are wrong

In the file plugins/micromega/micromega_plugin.mlpack , the dependencies are wrong: the module Mutils depends on the module Micromega but is placed before. This prevent users of this pack to use the module Mutils.

#### Coq Version

At least since Coq-8.8 (did not check for previous versions).

ckeller:
**Kind:** bug fix

Fixes #9768

Remark: it would be great to backport it to Coq-8.9.1 (there are less modules in micromega_plugin but it still applies).

### DIFF
--- a/plugins/micromega/micromega_plugin.mlpack
+++ b/plugins/micromega/micromega_plugin.mlpack
@@ -1,8 +1,8 @@
+Micromega
 Mutils
 Itv
 Vect
 Sos_types
-Micromega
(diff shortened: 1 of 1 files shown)


ejgallego (approved):
Looks OK; taking over as per PR policy.

PR #9775, opened 2019-03-15
inferCumulativity: shortcut for all-Invariant inductives

SkySkimmer:
The 0 universe case is certainly useful, but if that's the only one that comes up in practice we could go with a much smaller commit.
Submitting so I don't forget about it.

### DIFF
--- a/pretyping/inferCumulativity.ml
+++ b/pretyping/inferCumulativity.ml
@@ -12,41 +12,44 @@ open Reduction
 open Declarations
 open Constr
 open Univ
+open Variance
 open Util
 
-(** Throughout this module we modify a map [variances] from local
-   universes to [Variance.t]. It starts as a trivial mapping to
-   [Irrelevant] and every time we encounter a local universe we
-   restrict it accordingly. *)
+type inferred = IrrelevantI | CovariantI
 
-let infer_level_eq u variances =
-  if LMap.mem u variances
-  then LMap.set u Variance.Invariant variances
+(** Throughout this module we modify a map [variances] from local
+    universes to [inferred]. It starts as a trivial mapping to
+    [Irrelevant] and every time we encounter a local universe we
+    restrict it accordingly.
+    [Invariant] universes are removed from the map.
+*)
+exception TrivialVariance
+
+let maybe_trivial variances =
+  if LMap.is_empty variances then raise TrivialVariance
   else variances
 
+let infer_level_eq u variances =
+  maybe_trivial (LMap.remove u variances)
+
 let infer_level_leq u variances =
-  match LMap.find u variances with
-  | exception Not_found -> variances
-  | varu -> LMap.set u (Variance.sup varu Variance.Covariant) variances
+  (* can only set Irrelevant -> Covariant so nontrivial *)
+  LMap.update u (function
+      | None -> None
+      | Some CovariantI as x -> x
+      | Some IrrelevantI -> Some CovariantI)
+    variances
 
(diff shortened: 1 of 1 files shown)


SkySkimmer:
ping @Janno

ejgallego:
@SkySkimmer why is this a draft? Should a tag be set?

SkySkimmer:
I guess we can merge it, it's not that complex.

ejgallego:
> I guess we can merge it, it's not that complex.

Cool, please set a milestone.

SkySkimmer:
Isn't it better to let the assignee set the milestone when they merge?

ejgallego:
> Isn't it better to let the assignee set the milestone when they merge?

I was thinking that setting a milestone may help finding an assignee sometimes tho.

pretyping/inferCumulativity.ml
```diff
@@ -12,41 +12,52 @@ open Reduction
...
+
+let maybe_trivial variances =
+  if variances.nontrivial_count = 0 then raise TrivialVariance
+  else variances
 
 let infer_level_eq u variances =
-  if LMap.mem u variances
-  then LMap.set u Variance.Invariant variances
-  else variances
+  match LMap.set u Variance.Invariant variances.v with
```
ppedrot: This looks pretty fragile to me. If ever we break the invariant that adding the same data in a map results in physical equality, this code is broken. I'd rather check the current value, even if that incurs one more access, or use the `modify` function.

pretyping/inferCumulativity.ml
```diff
@@ -12,41 +12,52 @@ open Reduction
...
-  match LMap.find u variances with
-  | exception Not_found -> variances
-  | varu -> LMap.set u (Variance.sup varu Variance.Covariant) variances
+  let v' = LMap.update u (function
+      | None -> None
+      | Some (Invariant| Covariant) as x -> x
+      | Some Irrelevant -> Some Covariant)
+      variances.v
+  in
+  if variances.v == v' then variances
```
ppedrot: Same remark.

ppedrot (changes requested):

ppedrot:
@SkySkimmer please don't use physical equality for control. You can use it as an approximation for fast-path, with the invariant that `x == y` implies `x = y`, but that's about it. Other uses are really sources of very subtle bugs you don't want to have to find out painfully.

SkySkimmer:
New version. The idea is that we only keep the non-Invariant universes in the map, so making a universe Invariant is just Map.remove and instead of keeping a count of non-invariant universes we just check if the map is empty.

ppedrot (approved):

ppedrot:
Can we get a bench on polymorphism-using devs?

SkySkimmer:
It's cumulativity not polymorphism, and it only has an effect when declaring inductives, so I don't think we would see any effect.

ppedrot:
No developments are using cumulativity?

mattam82:
Coq-HoTT will be using it soon, for equality, and I remember we had one case there where switching cumulativity on for everything made cumulativity inference  take ages for one record declaration (to finally show that all universes were invariant). Equations Type variant uses cumulativity for inductives heavily, but there are not so many inductive declarations. Amin Timany’s category-theory library uses it more intensively but is not part of CI I believe. Maybe MTac2 has some examples?

ppedrot:
OK. In absence of a comprehensive bench, I guess we can safely merge this. I can't be worse than what is implemented currently, so...

PR #9776, opened 2019-03-15
[kernel] Fix compare_head_gen_leq_with to use [leq] on applications

mattam82:
This fixes an incompleteness of subtyping on cumulative inductives,
where I@{i} A <= I@{j} A should imply i <= j, i = j or no relation
depending on the variance of I's universe, but leq_constr_universes
was incorrectly forcing i = j.

**Kind:** bug fix

I'm not sure this requires an entry in CHANGES.md, the problem affects only flexible universes during elaboration (as the kernel does only checking, without any flexible/rigid notion), so its super-technical.

### DIFF
--- a/test-suite/success/cumulativity.v
+++ b/test-suite/success/cumulativity.v
@@ -137,3 +137,12 @@ Module WithIndex.
   Monomorphic Constraint i < j.
   Definition bar : eq mkfoo@{i} mkfoo@{j} := eq_refl _.
 End WithIndex.
+
+Module CumulApp.
+
+  (* i is covariant here, and we have one parameter *)
+  Inductive foo@{i} (A : nat) : Type@{i+1} := mkfoo (B : Type@{i}).
+
+  Definition bar@{i j|i<=j} := fun x : foo@{i} 0 => x : foo@{j} 0.
+
+End CumulApp.
--- a/kernel/constr.ml
+++ b/kernel/constr.ml
@@ -860,7 +860,7 @@ let compare_head_gen_leq_with kind1 kind2 leq_universes leq_sorts eq leq nargs t
   | App (c1, l1), App (c2, l2) ->
     let len = Array.length l1 in
     Int.equal len (Array.length l2) &&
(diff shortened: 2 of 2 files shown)


test-suite/success/cumulativity.v
```diff
@@ -137,3 +137,26 @@ Module WithIndex.
...
+    set (foonat := foo 0).
+    match goal with
+      [ foonat := ?mytype |- _ ] =>
+      (* We need to capture the body, because it is only the syntactic check that was broken *)
+      (* x is at Set *)
+      refine (let x := mkfoo 0 nat : foo@{Set} 0 in
+              let y := x : mytype in (* This forced incorrectly that Set = 'i' *)
+              mkfoo@{i} 0 nat : mytype (* This prevents this otherwise correct instantiation *))
+    end.
+  Defined.
```
SkySkimmer: This test seems unnecessarily complex, just `Definition bar@{i j|i<=j} := fun x : foo@{i} 0 => x : foo@{j} 0.` exhibits the failure (using the same definition for `foo`)

test-suite/success/cumulativity.v
```diff
@@ -137,3 +137,26 @@ Module WithIndex.
   Monomorphic Constraint i < j.
   Definition bar : eq mkfoo@{i} mkfoo@{j} := eq_refl _.
 End WithIndex.
+
+Module IndCumul.
```
SkySkimmer: I'd call this sub test `EasyApp` or some such thing, `IndCumul` sounds too general to me.

SkySkimmer (approved):

test-suite/success/cumulativity.v
mattam82: Right thanks, I was misled thinking this was related to flexibles only.

test-suite/success/cumulativity.v
mattam82: Done

mattam82:
I think the failures are spurious, @SkySkimmer ?

PR #9777, opened 2019-03-15
[Nix] A few improvements

vbgl:
- The reference to `nixpkgs`  is stored in a file and is now shared by the main shell/build script and the “ci” script (as requested by Maxime).
 - Third-party libraries `coq-ext-lib` and `simple-io` now refer to the *moving* master version. This fixes the `QuickChick` build.

cc/ @maximedenes


dev/ci/nix/default.nix
```diff
@@ -20,8 +20,16 @@ let mathcomp = coqPackages.mathcomp.overrideAttrs (o: {
...
-let coq-ext-lib = coqPackages.coq-ext-lib; in
-let simple-io = coqPackages.simple-io; in
+
+let coq-ext-lib = coqPackages.coq-ext-lib.overrideAttrs (o: {
+    src = fetchTarball "https://github.com/coq-ext-lib/coq-ext-lib/tarball/master";
+  }); in
+
+let simple-io = coqPackages.simple-io.overrideAttrs (o: {
+    src = fetchTarball "https://github.com/Lysxia/coq-simple-io/tarball/master";
+    buildInputs = o.buildInputs ++ [ coq-ext-lib ];
```
Zimmi48: Why do you need to change this. Is it because the last packaged version did not depend on coq-ext-lib? Can we update the version in nixpkgs to avoid this change?

nixpkgs.nix
```diff
@@ -0,0 +1,4 @@
+import (fetchTarball {
```
Zimmi48: Please move this file under `dev` to avoid polluting the root given that it is not necessary that it is there.

Zimmi48 (commented):
LGTM modulo one question and one request for change.

nixpkgs.nix
vbgl: Done.

dev/ci/nix/default.nix
vbgl: Yes we can. See: https://github.com/NixOS/nixpkgs/pull/57708

Zimmi48: Cool!

vbgl: Fixed.

Zimmi48 (approved):

Zimmi48:
@vbgl Could we consider extending this work by providing a nixpkgs overlay that sets the version of every package to the development branch? This would make it easier to support building a package with the development branch of Coq and its dependencies, as discussed in coq-community/manifesto#37.

vbgl:
What kind of overlay? Where should it be stored?

Zimmi48:
The kind that is defined in https://nixos.org/nixpkgs/manual/#chap-overlays. I have never used such overlays so I'm not really sure what the setup would be like.

vbgl:
Indeed. We should do something like that.

I think this PR is ready for merging. Thanks.

PR #9783, opened 2019-03-15
[ci] [gitlab] Replace YAML grafting by `extends: ` declaration.

ejgallego:
This should in general be more robust as we don't need to remember to
graft subsections.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -55,7 +55,7 @@ after_script:
 ######################################################
 
 # TODO figure out how to build doc for installed Coq
-.build-template: &build-template
+.build-template:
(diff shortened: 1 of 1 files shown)


ejgallego:
After a brief look CI seems still OK [that is to say, doing what it is supposed to be doing]

SkySkimmer (approved):
I stuck before/after in https://gitlab.com/coq/coq/-/ci/lint and diffed the output, there are no differences.

ejgallego:
> I stuck before/after in https://gitlab.com/coq/coq/-/ci/lint and diffed the output, there are no differences.

Good trick!

Issue #9767, opened 2019-03-14
Timing diff scripts should support non-utf8


JasonGross:
#### Description of the problem

TimeFileMaker.py does not support files which emit non-utf-8 to stdout.  See [this job log](https://gitlab.com/coq/coq/-/jobs/176113494) and the conversation in https://github.com/coq/coq/pull/9752.  It would be nice for them to support this.  This requires using bytes/`rb` rather than `r`, etc.  I might do this on my own in about two weeks, though I'd love if someone more familiar with python3 and `bytes` would do this.

#### Coq Version

8.9

JasonGross:
@andres-erbsen or @samuelgruetter Can you point us at a particular file and commitish that prints non-utf-8 to the console, for inclusion in the test-suite, when I (or someone else) gets around to working on this?

samuelgruetter:
[Here](https://travis-ci.com/mit-plv/bedrock2/jobs/184001043#L1195)'s a log on Travis with some non-utf8 in the console. It's from commit https://github.com/mit-plv/bedrock2/commit/9b427dc4cea9731a7869eb7c68a1de95047dd7be, file [bedrock2/src/BytedumpTest.v](https://github.com/mit-plv/bedrock2/blob/9b427dc4cea9731a7869eb7c68a1de95047dd7be/bedrock2/src/BytedumpTest.v)

andres-erbsen:
an alternative to updating the script would be up change bedrock2 build to not output binary to console. it would probably require changing the build organization (currently we just call coq makefile). fixing the script is probably preferred anyway.

ejgallego:
I am much afraid that I don't understand what is going on here, however we must get the `bedrock2` build fixed in the CI, please point us to what the right workaround is.

JasonGross:
The right workaround for now is to redirect the output of `make ` to `/dev/null` for the bedrock2 ci, and cite this issue in a comment.

andres-erbsen:
Actually, it should work just to grep out lines with high bytes...

andres-erbsen:
`echo -ne 'ab\x80c\ndd\nas' | LANG=en_US.UTF-8 grep -a '^.*$'`
produces
`dd`
`as`

ejgallego:
> The right workaround for now is to redirect the output of `make ` to `/dev/null` for the bedrock2 ci, and cite this issue in a comment.

That seems pretty radical, if something else fails how are we going to debug it?

PR #9787, opened 2019-03-16
iconv bedrock2 CI output to UTF-8

andres-erbsen:
The timing diff script dies badly on non-utf8 input (https://github.com/coq/coq/issues/9767).

**Kind:** infrastructure

### DIFF
--- a/dev/ci/ci-bedrock2.sh
+++ b/dev/ci/ci-bedrock2.sh
@@ -6,4 +6,4 @@ ci_dir="$(dirname "$0")"
 FORCE_GIT=1
 git_download bedrock2
 
-( cd "${CI_BUILD_DIR}/bedrock2" && git submodule update --init --recursive && COQMF_ARGS='-arg "-async-proofs-tac-j 1"' make )
+( cd "${CI_BUILD_DIR}/bedrock2" && git submodule update --init --recursive && COQMF_ARGS='-arg "-async-proofs-tac-j 1"' make | iconv -t UTF-8 -c `#9767` )


SkySkimmer:
Doesn't work

ejgallego:
Yeah, how that could work ?

andres-erbsen:
Indeed doesn't work (I thought it would). @ejgallego if you understand why it doesn't work, please enlighten me. In the meantime I'm seeing whether `>/dev/null` makes the build pass at all in #9790.

andres-erbsen:
Works on my machine (https://gist.github.com/andres-erbsen/9cccbcadb23e54276ea52547e7377b60). Trying out a more aggressive `env` command...

JasonGross:
What about something like `grep -v [^[:print:]]`? (Exclude anything with a non-printable character.)

ejgallego:
I'm not sure how you think folks `grep` is gonna work here; likely the `/dev/null` solution worked as it made python think there was no terminal attached, I'm not sure grep will have the same effect.

JasonGross:
@ejgallego I think you're misunderstanding the setup.  The [wrapper script](https://github.com/coq/coq/blob/master/dev/ci/ci-wrapper.sh) pipes the output of `ci-bedrock2.sh` to a file via `tee`.  That lot is then fed to the python script.  If this output log contains anything other than valid utf8, the python script barfs.  The point of `grep ` is to ensure that no invalid utf8 characters end up in the log.  I suppose another stop-gap measure is to have the python script read things one line at a time, and discard any lines that raise a Unicode decode error

dev/ci/ci-bedrock2.sh
```diff
@@ -6,4 +6,4 @@ ci_dir="$(dirname "$0")"
 FORCE_GIT=1
 git_download bedrock2
 
-( cd "${CI_BUILD_DIR}/bedrock2" && git submodule update --init --recursive && COQMF_ARGS='-arg "-async-proofs-tac-j 1"' make )
+( cd "${CI_BUILD_DIR}/bedrock2" && git submodule update --init --recursive && COQMF_ARGS='-arg "-async-proofs-tac-j 1"' make | env -i LANG=en_US.UTF-8 grep -av '[^[:print:]]' `#9767` )
```
SkySkimmer: This grep doesn't work for my local testing, OTOH piping through `iconv -t UTF-8 -c` works (we already have iconv in the image)
Found through https://unix.stackexchange.com/a/6517

SkySkimmer (approved):
Please squash the commits, then we can merge.

andres-erbsen:
squashed

ejgallego:
Ok looks good then, thanks to all!

ejgallego:
@andres-erbsen you created a branch on the main Coq repos, beware of your remotes!

andres-erbsen:
@ejgallego ugh, the github web interface changed behavior, probably based on whatever was done to allow me to add labels to issues. It used to be that if I edited a coq file on github, it would fork the repo and create the branch there, and then make a PR. Now the exact same user interface creates a branch on coq/coq instead, with no indication of that as far as I can tell. In their defense, the previous interface explicitly said that the branch will be created on a fork...

Zimmi48:
Yes, this is sad, but once you're added to the contributors team, you can't use this interface anymore.

PR #9790, opened 2019-03-17
trying out bedrock2 make>/dev/null

andres-erbsen:
@JasonGross suggested this as a workaround for #9767. We may not want this, but I am curious whether it even makes the build pass.


andres-erbsen:
The bedrock2 build passed.

Issue #9351, opened 2019-01-17
Windows installer: and Flocq and Interval as packages to Windows installer


MSoegtropIMC:
I want to add Flocq (indepdendent of VST or CompCert) and Coq-Interval as Windows installer packages. On the way I would also add Interval to CI.

Configuring and building Flocq and Interval is straight forward, but there are a few issues with remake install:

- cause of the mix of MinGW and cygwin, the install path ends up to have a \r in it, which is added to the path - likely some issue with awk (it does not strip \r on cygwin)
- examples are not installed
- README is not installed
- quite a few (but not all) examples import Interval with Require Import Interval_tactic (without Interval.)

I think the last one should be fixed in the Interval sources, but for the install issues I am not so sure. Redoing the installer and including examples and README would be 3 lines in the shell script. On the other hand fixing the \r issue might be rather intricate since I think awk or whatever is called via some automake / configure magic, so that I can't easily add a strip (might have to patch automake). We had a lot of similar issues with sed and I ended up compiling a patched sed which strips \r.

@silene : First I hope that you don't have any objections against adding Flocq and Interval. Then I would like to have your opinion on how to get around the install issues:

a) simply write my own install procedure
b) Fix the /r issue and add installation of examples and README to the remake install target

Zimmi48:
@MSoegtropIMC @vbgl I am turning this issue into a blocker for 8.10.0 because given that it was "fixed" in 8.9.1, not doing it for 8.10.0 would be a serious regression.

MSoegtropIMC:
One missing thing here is Gappa tool, which I also added in 8.9.1. It originally crashed on 64 bit Windows cause of the int/pointer size difference there. I fixed it and provided a patch to @silene which he was not 100% happy with. I can quickly make a PR which is equivalent to what is in 8.9.1. I would say the risk is low. Worst case gappa tool (an external C++ app) crashes and fails to provide a proof. But I tested my patched version quite a bit and so far it did work fine, so I think it is better than not having it.

silene:
I have now released Gappa 1.3.5, which should hopefully fix the remaining issues on 64-bit Windows, hopefully.

Zimmi48:
@MSoegtropIMC This is a gentle reminder of this issue, now that you are no longer busy with fixing CoqIDE on Windows.

Zimmi48:
@MSoegtropIMC ping

MSoegtropIMC:
[...]

PR #9792, opened 2019-03-17
Add Flocq, CoqInterval, Gappa tool and Gappa plugin to the Windows Installer (Issue #9351)

MSoegtropIMC:
This PR Implements Issue #9351 and adds these new plugins to the Windows Installer

- Flocq (as independent library)
- CoqInterval
- Gappa stand alone executable for MinGW
- Gappa Coq plugin

This is work in progress - I don't plan to have it merged before @silene pushed the patches to the respective repos, but I want to start testing CI now.

A note on Flocq: this is already included as part of CompCert, but this version is quite different. I need to discuss this with @xavierleroy and @silene. In the end this makes it very hard to verify numerical C programs using VST if VST (via CompCert) uses a different version of Flocq than CoqInterval and Gappa.

### DIFF
--- a/dev/build/windows/MakeCoq_MinGW.bat
+++ b/dev/build/windows/MakeCoq_MinGW.bat
@@ -376,6 +376,7 @@ IF "%RUNSETUP%"=="Y" (
     -P pkg-config ^
     -P mingw64-%ARCH%-binutils,mingw64-%ARCH%-gcc-core,mingw64-%ARCH%-gcc-g++,mingw64-%ARCH%-windows_default_manifest ^
     -P mingw64-%ARCH%-headers,mingw64-%ARCH%-runtime,mingw64-%ARCH%-pthreads,mingw64-%ARCH%-zlib ^
+    -P mingw64-%ARCH%-gmp,mingw64-%ARCH%-mpfr ^
     -P libiconv-devel,libunistring-devel,libncurses-devel ^
     -P gettext-devel,libgettextpo-devel ^
     -P libglib2.0-devel,libgdk_pixbuf2.0-devel ^
@@ -383,6 +384,7 @@ IF "%RUNSETUP%"=="Y" (
     -P gtk-update-icon-cache ^
     -P libtool,automake ^
     -P intltool ^
+    -P bison,flex ^
     %EXTRAPACKAGES% ^
     || GOTO ErrorExit
 
--- a/dev/build/windows/MakeCoq_MinGW.bat
+++ b/dev/build/windows/MakeCoq_MinGW.bat
@@ -373,7 +373,8 @@ IF "%RUNSETUP%"=="Y" (
     -P make,unzip ^
     -P gdb,liblzma5 ^
     -P patch,automake1.14 ^
-    -P mingw64-%ARCH%-binutils,mingw64-%ARCH%-gcc-core,mingw64-%ARCH%-gcc-g++,mingw64-%ARCH%-pkg-config,mingw64-%ARCH%-windows_default_manifest ^
+    -P pkg-config ^
+    -P mingw64-%ARCH%-binutils,mingw64-%ARCH%-gcc-core,mingw64-%ARCH%-gcc-g++,mingw64-%ARCH%-windows_default_manifest ^
     -P mingw64-%ARCH%-headers,mingw64-%ARCH%-runtime,mingw64-%ARCH%-pthreads,mingw64-%ARCH%-zlib ^
     -P libiconv-devel,libunistring-devel,libncurses-devel ^
     -P gettext-devel,libgettextpo-devel ^
--- a/dev/build/windows/makecoq_mingw.sh
+++ b/dev/build/windows/makecoq_mingw.sh
@@ -242,6 +242,7 @@ fi
 # $5 [optional] module name (if different from archive)
 # $6 [optional] expand folder name (if different from module name)
 # $7 [optional] module base name (used as 2nd choice for patches, defaults to $5)
+# $8 [optional] extra tar arguments (unquoted!)
 # ------------------------------------------------------------------------------
 
 function get_expand_source_tar {
@@ -270,6 +271,20 @@ function get_expand_source_tar {
     basename=$name
   fi
 
+  if [ "$#" -ge 8 ] ; then
+    tarextra=$8
+  else
+    tarextra=""
+  fi
+
+  # INRIA GIT paths don't follow the naming scheme, but we still need the extension
+  # So as a hack, if the name is empty, $1 is taken as full url
+  if [ -n "$2" ] ; then
+    fullurl="$1/$2.$3"
+  else
+    fullurl="$1"
+  fi
+
   # Set logging target
   logtargetold=$LOGTARGET
   LOGTARGET=$name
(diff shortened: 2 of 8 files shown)


Zimmi48:
@silene said he's working on porting CompCert to the lastest version of Flocq. I don't know how it is going.

MSoegtropIMC:
> @silene said he's working on porting CompCert to the lastest version of Flocq. I don't know how it is going.

Thanks, good to know! I think we can integrate it as is then as soon as the patches are merged and the CI is finalized.

xavierleroy:
CompCert contains its own copy of Flocq, so you can install a newer version of Flocq and it will not conflict with CompCert.  

This said, an upgrade of CompCert's version of Flocq is about to be merged, indeed: https://github.com/AbsInt/CompCert/pull/273

MSoegtropIMC:
@xavierleroy : will this be the same version CoqInterval and Gappa are based on (which is the latest version of Focq as far as I can tell)? I ask because I am working on verification of numerical code with Princeton VST and want to use Gappa and CoqInterval to discharge proof obligations from C code, which only works if both use the same version of Flocq.

xavierleroy:
> will this be the same version CoqInterval and Gappa are based on?

Perhaps.  But in the current state of affairs, CompCert's copy of Flocq defines types and modules in a different namespace than that of the standalone Flocq library, precisely so that the two installations of Flocq can differ.  So, I'm pretty sure CompCert's float types will be incompatible with CoqInterval's and Gappa's float types.  But only @silene can tell.

Another option would be for CompCert to stop having its own copy of Flocq and rely on a separately-installed Flocq.  OPAM makes this easy.  But industrial users of CompCert may not agree.

Zimmi48:
@xavierleroy It is my understanding that Dune supports well a model where users can choose between using an opam packaged dependency, or a vendored one. @ejgallego is working on adding Coq support to Dune (ocaml/dune#1555). Once this is done, it would make sense to start relying on Dune to build CompCert and support these two models. [And at some point, it could be a good idea for us not to build Flocq three times in our CI (once for Flocq, once for CompCert, and once for VST), although at the current time, GitLab CI doesn't support the dependency model we need to do this.]

MSoegtropIMC:
> Another option would be for CompCert to stop having its own copy of Flocq and rely on a separately-installed Flocq.

I would prefer this. One of the major advantages of Coq is that so many useful "modules" exists based on the same tool (like CompCert, VST and arithmetic proof tools of @silene ). It is not good to put up walls between these modules unless there is a really good reason.

MSoegtropIMC:
I had a discussion on the patches with @silene. He said he wont have time to integrate them upstream soon since he is very busy with other topics. Since the patches are rather minor I would suggest to merge it like this. I am using the resulting installer since a while and Interval and Gappa work fine. I will remove the patches then later as soon as silene integrated them upstream.

I will add unit tests for the Linux versions of the added plugins this week. The rest of the PR can be reviewd as is - I don't plan any changes and, as I said, it works.

MSoegtropIMC:
Hmm ok, I just saw that this is a PR for 8.9.

I asked a while back on coqdev if there will be a 8.9.1 where this can be included. I guess not.

Should I rebase this PR to master or should I keep it (in case there will be an 8.9.1) and create a new PR for master? The files touched tend to be quite a bit different between 8.9 and master and cannot be merged trivially, so if there will be a 8.9.1, I would recommend to keep it.

Zimmi48:
There will be an 8.9.1. @silene told me that he planned to prepare it during the second week of April.
Yes, please open a corresponding PR for master so that it can be merged there, preferably, before the branching next week.

MSoegtropIMC:
> Yes, please open a corresponding PR for master so that it can be merged there, preferably, before the branching next week.

Perfect will do.

xavierleroy:
CompCert trunk now uses Flocq 3.1, but with its own copy of Flocq.

Using an already-installed version of Flocq is a simple edit to 3 source files, but I haven't found a way to do it entirely from configure and Makefile, without changing .v files.  

> It is my understanding that Dune supports well a model where users can choose between using an opam packaged dependency, or a vendored one

I am skeptical because I believe Coq's handling of name spaces does not support this scenario, but I'd be happy to be proved wrong.

MSoegtropIMC:
@xavierleroy : perfect. If you don't mind I will simply apply a small patch to compcert to use a separate Flocq in the Windows Installer.

Zimmi48:
>I am skeptical because I believe Coq's handling of name spaces does not support this scenario, but I'd be happy to be proved wrong.

Indeed, from recent discussion with @ejgallego, it seems that this model will not be supported in the first version of the Coq support in Dune. Later on, though, with some necessary modifications to Coq itself, we want to support such a model.

MSoegtropIMC:
> Later on, though, with some necessary modifications to Coq itself, we want to support such a model.

I guess what is needed is a file which maps Coq module paths to file paths. This would help also in many other cases, e.g. to switch between fast and safe implementations of computation (e.g. using BigZ or Z via a wrapper).

I guess it is agreeable to then have "yet another patch" in the Windows build until we switched to dune and such functionality is available.

ejgallego:
> Indeed, from recent discussion with @ejgallego, it seems that this model will not be supported in the first version of the Coq support in Dune. Later on, though, with some necessary modifications to Coq itself, we want to support such a model.

We could support composition / vendoring now without a lot of problems, the issue is more about compatibility with project using coq_makefile. Vendoring just sets the proper load paths and dependency rules, I don't see any need for namespacing, as the code itself cannot distinguish when it is linking with vendored code.

ejgallego (commented):
I am OK with this, please re-request a review when ready.

MSoegtropIMC:
@ejgallego : since @silene wants to push the patches upstream later and since I want to do the alignment of Flocq, CompCert and VST as a separate step in one go (which is likely a bit of work), this is ready as is from my point of view.

Zimmi48:
> > Yes, please open a corresponding PR for master so that it can be merged there, preferably, before the branching next week.
> 
> Perfect will do.

I suppose you didn't forget to do this and have this on your todo list, but just in case...

MSoegtropIMC:
> I suppose you didn't forget to do this and have this on your todo list, but just in case...

Yes, it is pretty on top of my to do list.

Zimmi48:
The milestone here is pretty weird. This should be in 8.9.1 given that this PR is v8.9.

MSoegtropIMC:
I was going to ask how to do it: I want it in both 8.9.1 and 8.10 and since @vbgl asked for everything we want in 8.10 be marked as such this week, I labeled it this way. Should I create 2 issues or does 8.9.1 imply 8.10beta?

Zimmi48:
>Should I create 2 issues or does 8.9.1 imply 8.10beta?

For a PR submitted to the master branch, 8.9.1 does imply 8.10+beta1. But this case is a bit special since you plan to have two separate PRs.

Since #9351 is already in the 8.10+beta1 milestone, it will be enough as a reminder to @vbgl until you have time to prepare the actual PR for master / 8.10. According to the schedule released by @vbgl, we still have one month before the beta release anyway.

MSoegtropIMC:
Can someone please review / merge?

As discussed @silene doesn't have time to push the patches upstream right now, and I want this in 8.9.1.

Zimmi48:
But @silene is the one who manages v8.9 and 8.9.1 so it is him who must merge.

MSoegtropIMC:
> But @silene is the one who manages v8.9 and 8.9.1 so it is him who must merge.

Yes, in a private email he mentioned that he thinks it is a design bug in gappa that it had to be fixed in the way I did, and he wants to do some changes in this area. Not sure what we do for 8.9.1 - since it works - even though the code is, well, not for small kids - I would vote for including it.

ejgallego (approved):
Refs additions are OK to me.

MSoegtropIMC:
@silene : CI passes now with the extended run time. It would also have passed with the old limit, but the run times are bit random so Zimmi48 extended the timeout in order to avoid issues.
So this should be ready to merge for 8.9.1.

PR #9796, opened 2019-03-18
[ci] Guard broken jobs under an `UNRELIABLE = enabled` variable.

ejgallego:
This way, we can opt out of these development in our personal
pipelines, so we don't risk being throttled by Gitlab.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -253,6 +253,9 @@ build:base+async:
     COQ_EXTRA_CONF: "-native-compiler yes -coqide opt"
     COQUSERFLAGS: "-async-proofs on"
   allow_failure: true # See https://github.com/coq/coq/issues/9658
+  only:
+    variables:
+      - $UNRELIABLE =~ /enabled/
 
 build:quick:
   extends: .build-template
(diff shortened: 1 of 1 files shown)


SkySkimmer:
How sure are you that gitlab is throttling, and that it's because of these jobs?

ejgallego:
Regarding the findlib problem, upstream is informed, a fix shoul appear any time soon.

ejgallego:
> How sure are you that gitlab is throttling, and that it's because of these jobs?

Gitlab won't confirm anything like that officially, go talk to them if you want; but we can do this without a lot of harm and then we can compare our personal pipeline with the one here.

ejgallego:
Or just look at Grafana and then here https://gitlab.com/coq/coq/-/jobs?scope=running [pyrolise is taking almost all of the jobs these days]

ejgallego:
Timeout is the worst offense you can do in terms of worker occupation.

SkySkimmer:
>Regarding the findlib problem, upstream is informed, a fix shoul appear any time soon.

What findlib problem?

ejgallego:
> What findlib problem?

The failure of the trunk builds, this is also raising our failure rate which is not good. Hope to have it fixed very soon.

SkySkimmer (approved):

SkySkimmer:
I'll let you flip the variable in the gitlab settings.

ejgallego:
Gitlab variable set to true.

PR #9797, opened 2019-03-18
[ci] Fix OCaml trunk build.

ejgallego:
This is a temporary pinning, we should amend it before merge in Coq
[once it is indeed merged upstream] , see
https://gitlab.camlcity.org/gerd/lib-findlib/merge_requests/22

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -464,7 +464,9 @@ test-suite:edge+trunk+make:
     - eval $(opam env)
     - opam repo add ocaml-beta https://github.com/ocaml/ocaml-beta-repository.git
     - opam update
-    - opam install ocaml-variants=4.09.0+trunk num
+    - opam install ocaml-variants=4.09.0+trunk
+    - opam pin add -n ocamlfind https://gitlab.camlcity.org/gerd/lib-findlib.git#gerd/optional-vmthreads
+    - opam install num
     - eval $(opam env)
     # We avoid problems with warnings:
     - ./configure -profile devel -warn-error no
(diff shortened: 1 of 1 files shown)


ejgallego:
Interesting, the make-based stuff seems to have a problem with the thread flags; I'll try a tweak.

ejgallego:
> Interesting, the make-based stuff seems to have a problem with the thread flags; I'll try a tweak.

The tweak didn't work at all, reverting. This needs more investigation, Dune manages to build Coq without a problem so it is certainly some problem with the make flags.

SkySkimmer:
dune goes through findlib in a different way though, when I use display=verbose I see `/absolute/path/to/ocamlc.opt bla bla` not a call to ocamlfind.

SkySkimmer:
I think the findlib proposed patch is just incorrect (it disabled -thread as well as -vmthread)

ejgallego:
> dune goes through findlib in a different way though, when I use display=verbose I see `/absolute/path/to/ocamlc.opt bla bla` not a call to ocamlfind.

Oh in fact it does not use findlib at all these days IIRC; it has its own implementation of `META` parsing.

ejgallego:
> I think the findlib proposed patch is just incorrect (it disabled -thread as well as -vmthread)

Yeah that is my impression too.

SkySkimmer:
The findlib MR just updated, let's see how it goes.

ejgallego:
> The findlib MR just updated, let's see how it goes.

Seems to be doing better indeed.

ejgallego:
@SkySkimmer may be can merge this as-is, findlib moves super slowly so indeed it could be months before a release / merge to master.

SkySkimmer:
Let's merge then

SkySkimmer (approved):

PR #9801, opened 2019-03-19
CI: Don't fail validate:quick when build:quick failed

SkySkimmer:
(since build:quick is allow_failure)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -167,7 +167,7 @@ after_script:
   dependencies:
     - not-a-real-job
   script:
-    - cd _install_ci
+    - cd _install_ci || exit 0 # maybe no artifacts from allow_failure dep on build:quick
     - find lib/coq/ -name '*.vo' -print0 > vofiles
     - for regexp in 's/.vo//' 's:lib/coq/plugins:Coq:' 's:lib/coq/theories:Coq:' 's:/:.:g'; do sed -z -i "$regexp" vofiles; done
     - xargs -0 --arg-file=vofiles bin/coqchk -silent -o -m -coqlib lib/coq/


.gitlab-ci.yml
```diff
@@ -167,7 +167,7 @@ after_script:
   dependencies:
     - not-a-real-job
   script:
-    - cd _install_ci
+    - cd _install_ci || exit 0 # maybe no artifacts from allow_failure dep on build:quick
```
ejgallego: Umm, I am not a fan of this given that it will apply to all validate jobs; shouldn't we instead mark that particular job as allow fail then?

Why is gitlab trying to start the validate job with a failed dependency?

SkySkimmer: >shouldn't we instead mark that particular job as allow fail then?

It's not allow_fail though, if build:quick succeeds then we expect this one to succeed too as it should only operate on fairly normal vo files
>Why is gitlab trying to start the validate job with a failed dependency?

Because that's how the semantics of allow_failure are 🤷

SkySkimmer:
Why's equations failing? @mattam82

.gitlab-ci.yml
ejgallego: I see, thanks; but still the scope of the patch seems to wide for me; we could have a problem with the general CI setup for validate and it would be undetected for days.

Can we restrict the fix [with a var] to the particular job involved?

In fact this illustrates the very high cost of carrying unreliable tests in the CI, we should just drop the tests responsible for this.

ejgallego:
CI passed, but with the usual async failure.

.gitlab-ci.yml
ejgallego: In fact none of the base builds should be in allow failure; again we are adding hack to the CI to accommodate for broken code; IMO the broken is the thing that should be fixed.

ejgallego:
Another choice which is acceptable to me is to merge both jobs and move the unreliable one out of build.

SkySkimmer:
Does build:quick ever fail recently? Let's forget about this PR.

Issue #9664, opened 2019-02-27
[meta] Organization of compatibility breaking changes.


ejgallego:
I've opened this issue to discuss a possible implementation of the request in #6043 , that is to say, to have a model where we can perform changes to core parts of the system that won't retain backwards compatibility.

What this issue proposes is to "version" full subsystems in such a way that they can undergo a major revision incorporating many changes but still be controlled by a single flag.

The phases a subsystem willing to declare a new major version would undergo would be:
- _alpha phase_: a new version `Set Notations Version 2.0` would be added to the system, nothing that this version is not stable. Bleeding-edge users can test the features, and new PRs and changes of behavior are the norm.
- _beta phase_: once the redesigned subsystem is declared stable, the beta phase is a feature freeze so people can start porting their developments to their new subsystem but only bug fixes will come in,
- _stable phase_: the new version is supported officially, however it is not the default,
- _takeover phase_: the old version is deprecated, and all users are forced to migrate, old subsystems are removed in the next 1 or 2 releases.

This proposal is of course a draft, and I suggest we collect issues and redesign requests on particular subsystems here as a start:

### Type classes:
- https://github.com/coq/coq/issues/9661
### Unification:
- https://github.com/coq/coq/issues/9663

gares:
I think bleeding edge users may need a very local way to enable a feature than a Set. I think it is a good exercise to apply this to primproj, polyuniv and sprop. In some cases alpha is already there, beta too. But no next phase seem reasonable to me.

In general this proposal makes sense to me, but I'm afraid there is no golden/general rule here. It depends too much on the feature. 
But maybe I'm mixing feature and subsystem, so you may want to clarify that.

ejgallego:
> But maybe I'm mixing feature and subsystem, so you may want to clarify that.

Indeed I think here I am thinking more of complete systems that need coherent design. For example notations or unification fall into that; these systems have many features and they interact in particular ways.

> I think bleeding edge users may need a very local way to enable a feature than a Set.

Set can be made local right? I'm not a big fan of `Set` but I used it as an example.

> think it is a good exercise to apply this to primproj, polyuniv and sprop.

I guess these look more like features than design decisions. For example, I could imagine we revise type classes introducing major changes to many parts, that would be the kind of change scope I am thinking of.

> In some cases alpha is already there, beta too.

Interesting, would you mind mentioning which subsystems or features do you think we have in alpha.

ejgallego:
[...]

PR #9185, opened 2018-12-10
Start of constructive measure theory

VincentSe:
**Kind:** feature

This is the beginning of constructive measure theory, following the article of Errett Bishop and Henry Cheng. It is more general than the Riemann and Newton integrals already in Coq, closer to the Lebesgue integral.

The pull request is already 3500 lines long, so I send it for review before developing it further.

PR #9803, opened 2019-03-19
Adding more trigonometry in Reals

thery:
**Kind:** infrastructure.

<!- [ ] Corresponding documentation was added / updated (including any warning and error messages added / removed / modified). -->

### DIFF
--- a/doc/changelog/10-standard-library/00000-title.rst
+++ b/doc/changelog/10-standard-library/00000-title.rst
@@ -1,3 +1,2 @@
 
 **Standard library**
-
--- a/doc/changelog/10-standard-library/00000-title.rst
+++ b/doc/changelog/10-standard-library/00000-title.rst
@@ -1,2 +1,3 @@
 
 **Standard library**
+
--- a/doc/changelog/10-standard-library/09803-trigo.rst
+++ b/doc/changelog/10-standard-library/09803-trigo.rst
@@ -1,2 +1,0 @@
-
-- Addition to the Reals theory (`#09803 <https://github.com/coq/coq/pull/09803>`_ by Laurent Théry):
--- a/doc/changelog/10-standard-library/09803-trigo.rst
+++ b/doc/changelog/10-standard-library/09803-trigo.rst
@@ -1,0 +1,2 @@
+
+- Addition to the Reals theory (`#09803 <https://github.com/coq/coq/pull/09803>`_ by Laurent Théry):
--- a/theories/Reals/R_sqr.v
+++ b/theories/Reals/R_sqr.v
@@ -72,7 +72,7 @@ Proof.
   rewrite Rinv_mult_distr.
   repeat rewrite Rmult_assoc.
   apply Rmult_eq_compat_l.
-  rewrite Rmult_comm. 
+  rewrite Rmult_comm.
   repeat rewrite Rmult_assoc.
   apply Rmult_eq_compat_l.
   reflexivity.
@@ -179,6 +179,38 @@ Proof.
   apply Rge_le in Hle;
   assert (0 <= y) by (apply Rle_trans with x; assumption).
   apply Rsqr_incr_1; assumption.
+Qed.
+
+Lemma neg_pos_Rsqr_lt : forall x y : R, - y < x -> x < y -> Rsqr x < Rsqr y.
+Proof.
+  intros x y Hneg Hpos.
+  destruct (Rcase_abs x) as [Hlt|HLe].
+  - rewrite (Rsqr_neg x); apply Rsqr_incrst_1.
+    + rewrite <- (Ropp_involutive y); apply Ropp_lt_contravar; exact Hneg.
+    + rewrite <- (Ropp_0). apply Ropp_le_contravar, Rlt_le; exact Hlt.
+    + apply (Rlt_trans _ _ _ Hneg) in Hlt.
+      rewrite <- (Ropp_0) in Hlt; apply Ropp_lt_cancel in Hlt; apply Rlt_le; exact Hlt.
+  - apply Rsqr_incrst_1.
+    + exact Hpos.
+    + apply Rge_le; exact HLe.
+    + apply Rge_le in HLe.
+      apply (Rle_lt_trans _ _ _ HLe), Rlt_le in Hpos; exact Hpos.
+Qed.
+
+Lemma Rsqr_bounds_le : forall a b:R, -a <= b <= a -> 0 <= Rsqr b <= Rsqr a.
+Proof.
+  intros a b [H1 H2].
+  split.
+  - apply Rle_0_sqr.
+  - apply neg_pos_Rsqr_le; assumption.
+Qed.
+
+Lemma Rsqr_bounds_lt : forall a b:R, -a < b < a -> 0 <= Rsqr b < Rsqr a.
+Proof.
+  intros a b [H1 H2].
+  split.
+  - apply Rle_0_sqr.
+  - apply neg_pos_Rsqr_lt; assumption.
 Qed.
 
 Lemma Rsqr_abs : forall x:R, Rsqr x = Rsqr (Rabs x).
--- a/theories/Reals/Ranalysis1.v
+++ b/theories/Reals/Ranalysis1.v
@@ -27,6 +27,7 @@ Definition div_fct f1 f2 (x:R) : R := f1 x / f2 x.
 Definition div_real_fct (a:R) f (x:R) : R := a / f x.
 Definition comp f1 f2 (x:R) : R := f1 (f2 x).
 Definition inv_fct f (x:R) : R := / f x.
+Definition mirr_fct f (x:R) : R := f (- x).
 
 Declare Scope Rfun_scope.
 Delimit Scope Rfun_scope with F.
@@ -40,6 +41,7 @@ Arguments opp_fct f%F x%R.
 Arguments mult_real_fct a%R f%F x%R.
 Arguments div_real_fct a%R f%F x%R.
 Arguments comp (f1 f2)%F x%R.
+Arguments mirr_fct f%F x%R.
 
 Infix "+" := plus_fct : Rfun_scope.
 Notation "- x" := (opp_fct x) : Rfun_scope.
@@ -92,7 +94,7 @@ exists (Rmin a a'); split.
 intros y cy; rewrite <- !q.
   apply Pa'.
   split;[| apply Rlt_le_trans with (Rmin a a');[ | apply Rmin_r]];tauto.
- rewrite R_dist_eq; assumption.   
+ rewrite R_dist_eq; assumption.
 apply Rlt_le_trans with (Rmin a a');[ | apply Rmin_l]; tauto.
 Qed.
 
@@ -499,7 +501,7 @@ Qed.
 
 (* Extensionally equal functions have the same derivative. *)
 
-Lemma derivable_pt_lim_ext : forall f g x l, (forall z, f z = g z) -> 
+Lemma derivable_pt_lim_ext : forall f g x l, (forall z, f z = g z) ->
   derivable_pt_lim f x l -> derivable_pt_lim g x l.
 intros f g x l fg df e ep; destruct (df e ep) as [d pd]; exists d; intros h;
 rewrite <- !fg; apply pd.
@@ -507,7 +509,7 @@ Qed.
 
 (* extensionally equal functions have the same derivative, locally. *)
 
-Lemma derivable_pt_lim_locally_ext : forall f g x a b l, 
+Lemma derivable_pt_lim_locally_ext : forall f g x a b l,
   a < x < b ->
   (forall z, a < z < b -> f z = g z) ->
   derivable_pt_lim f x l -> derivable_pt_lim g x l.
@@ -577,6 +579,124 @@ Qed.
 (** *                    Main rules                             *)
 (****************************************************************)
 
+(** ** Rules for derivable_pt_lim (value of the derivative at a point) *)
+
+Lemma derivable_pt_lim_id : forall x:R, derivable_pt_lim id x 1.
+Proof.
+  intro; unfold derivable_pt_lim.
+  intros eps Heps; exists (mkposreal eps Heps); intros h H1 H2;
+    unfold id; replace ((x + h - x) / h - 1) with 0.
+  rewrite Rabs_R0; apply Rle_lt_trans with (Rabs h).
+  apply Rabs_pos.
+  assumption.
+  unfold Rminus; rewrite Rplus_assoc; rewrite (Rplus_comm x);
+    rewrite Rplus_assoc.
+  rewrite Rplus_opp_l; rewrite Rplus_0_r; unfold Rdiv;
+    rewrite <- Rinv_r_sym.
+  symmetry ; apply Rplus_opp_r.
+  assumption.
+Qed.
+
+Lemma derivable_pt_lim_comp :
+  forall f1 f2 (x l1 l2:R),
+    derivable_pt_lim f1 x l1 ->
+    derivable_pt_lim f2 (f1 x) l2 -> derivable_pt_lim (f2 o f1) x (l2 * l1).
+Proof.
+  intros; assert (H1 := derivable_pt_lim_D_in f1 (fun y:R => l1) x).
+  elim H1; intros.
+  assert (H4 := H3 H).
+  assert (H5 := derivable_pt_lim_D_in f2 (fun y:R => l2) (f1 x)).
+  elim H5; intros.
+  assert (H8 := H7 H0).
+  clear H1 H2 H3 H5 H6 H7.
+  assert (H1 := derivable_pt_lim_D_in (f2 o f1)%F (fun y:R => l2 * l1) x).
+  elim H1; intros.
+  clear H1 H3; apply H2.
+  unfold comp;
+    cut
+      (D_in (fun x0:R => f2 (f1 x0)) (fun y:R => l2 * l1)
+        (Dgf no_cond no_cond f1) x ->
+        D_in (fun x0:R => f2 (f1 x0)) (fun y:R => l2 * l1) no_cond x).
+  intro; apply H1.
+  rewrite Rmult_comm;
+    apply (Dcomp no_cond no_cond (fun y:R => l1) (fun y:R => l2) f1 f2 x);
+      assumption.
+  unfold Dgf, D_in, no_cond; unfold limit1_in;
+    unfold limit_in; unfold dist; simpl;
+      unfold R_dist; intros.
+  elim (H1 eps H3); intros.
+  exists x0; intros; split.
+  elim H5; intros; assumption.
+  intros; elim H5; intros; apply H9; split.
+  unfold D_x; split.
+  split; trivial.
+  elim H6; intros; unfold D_x in H10; elim H10; intros; assumption.
+  elim H6; intros; assumption.
+Qed.
+
+Lemma derivable_pt_lim_opp :
+  forall f (x l:R), derivable_pt_lim f x l -> derivable_pt_lim (- f) x (- l).
+Proof.
+  intros f x l H.
+  apply uniqueness_step3.
+  unfold opp_fct, limit1_in, limit_in, dist; simpl; unfold R_dist.
+  apply uniqueness_step2 in H.
+  unfold limit1_in, limit_in, dist in H; simpl in H; unfold R_dist in H.
+  intros eps Heps; specialize (H eps Heps).
+  destruct H as [alp [Halp H]]; exists alp.
+  split; [assumption|].
+  intros x0 Hx0; specialize(H x0 Hx0).
+  rewrite <- Rabs_Ropp in H.
+  match goal with H:Rabs(?a)<eps |- Rabs(?b)<eps => replace b with a by (field; tauto) end.
+  assumption.
+Qed.
+
+Lemma derivable_pt_lim_opp_fwd :
+  forall f (x l:R), derivable_pt_lim f x (- l) -> derivable_pt_lim (- f) x l.
+Proof.
+  intros f x l H.
+  apply uniqueness_step3.
+  unfold opp_fct, limit1_in, limit_in, dist; simpl; unfold R_dist.
+  apply uniqueness_step2 in H.
+  unfold limit1_in, limit_in, dist in H; simpl in H; unfold R_dist in H.
+  intros eps Heps; specialize (H eps Heps).
+  destruct H as [alp [Halp H]]; exists alp.
+  split; [assumption|].
+  intros x0 Hx0; specialize(H x0 Hx0).
+  rewrite <- Rabs_Ropp in H.
+  match goal with H:Rabs(?a)<eps |- Rabs(?b)<eps => replace b with a by (field; tauto) end.
+  assumption.
+Qed.
+
+Lemma derivable_pt_lim_opp_rev :
+  forall f (x l:R), derivable_pt_lim (- f) x (- l) -> derivable_pt_lim f x l.
+Proof.
+  intros f x l H.
+  apply derivable_pt_lim_ext with (f := fun x => - - (f x)).
+  - intros; rewrite Ropp_involutive; reflexivity.
+  - apply derivable_pt_lim_opp_fwd; exact H.
+Qed.
+
+Lemma derivable_pt_lim_mirr_fwd :
+  forall f (x l:R), derivable_pt_lim f (- x) (- l) -> derivable_pt_lim (mirr_fct f) x l.
+Proof.
+  intros f x l H.
+  change (mirr_fct f) with (comp f (opp_fct id)).
+  replace l with ((-l) * -1) by ring.
+  apply derivable_pt_lim_comp; [| exact H].
+  apply derivable_pt_lim_opp.
+  apply derivable_pt_lim_id.
+Qed.
+
+Lemma derivable_pt_lim_mirr_rev :
+  forall f (x l:R), derivable_pt_lim (mirr_fct f) (- x) (- l) -> derivable_pt_lim f x l.
+Proof.
+  intros f x l H.
+  apply derivable_pt_lim_ext with (f := fun x => (mirr_fct f (- x))).
+  - intros; unfold mirr_fct; rewrite Ropp_involutive; reflexivity.
+  - apply derivable_pt_lim_mirr_fwd; exact H.
+Qed.
+
 Lemma derivable_pt_lim_plus :
   forall f1 f2 (x l1 l2:R),
     derivable_pt_lim f1 x l1 ->
@@ -602,28 +722,6 @@ Lemma derivable_pt_lim_plus :
   split.
   assumption.
   intros; rewrite H3; apply H8; assumption.
-  intro; unfold Rdiv; ring.
-Qed.
-
-Lemma derivable_pt_lim_opp :
-  forall f (x l:R), derivable_pt_lim f x l -> derivable_pt_lim (- f) x (- l).
-Proof.
-  intros.
-  apply uniqueness_step3.
-  assert (H1 := uniqueness_step2 _ _ _ H).
-  unfold opp_fct.
-  cut (forall h:R, (- f (x + h) - - f x) / h = - ((f (x + h) - f x) / h)).
-  intro.
-  generalize
-    (limit_Ropp (fun h:R => (f (x + h) - f x) / h) (fun h:R => h <> 0) l 0 H1).
-  unfold limit1_in; unfold limit_in; unfold dist;
-    simpl; unfold R_dist; intros.
-  elim (H2 eps H3); intros.
-  exists x0.
-  elim H4; intros.
-  split.
-  assumption.
-  intros; rewrite H0; apply H6; assumption.
   intro; unfold Rdiv; ring.
 Qed.
 
@@ -718,22 +816,6 @@ intros f x l a df;
 unfold Rdiv; rewrite Rmult_comm; apply derivable_pt_lim_scal; assumption.
 Qed.
 
-Lemma derivable_pt_lim_id : forall x:R, derivable_pt_lim id x 1.
-Proof.
-  intro; unfold derivable_pt_lim.
-  intros eps Heps; exists (mkposreal eps Heps); intros h H1 H2;
-    unfold id; replace ((x + h - x) / h - 1) with 0.
-  rewrite Rabs_R0; apply Rle_lt_trans with (Rabs h).
-  apply Rabs_pos.
-  assumption.
-  unfold Rminus; rewrite Rplus_assoc; rewrite (Rplus_comm x);
-    rewrite Rplus_assoc.
-  rewrite Rplus_opp_l; rewrite Rplus_0_r; unfold Rdiv;
-    rewrite <- Rinv_r_sym.
-  symmetry ; apply Rplus_opp_r.
-  assumption.
-Qed.
-
 Lemma derivable_pt_lim_Rsqr : forall x:R, derivable_pt_lim Rsqr x (2 * x).
 Proof.
   intro; unfold derivable_pt_lim.
@@ -748,41 +830,80 @@ Proof.
   ring.
 Qed.
 
-Lemma derivable_pt_lim_comp :
-  forall f1 f2 (x l1 l2:R),
-    derivable_pt_lim f1 x l1 ->
-    derivable_pt_lim f2 (f1 x) l2 -> derivable_pt_lim (f2 o f1) x (l2 * l1).
+(** ** Rules for derivable_pt (derivability at a point) *)
+
+Lemma derivable_pt_id : forall x:R, derivable_pt id x.
 Proof.
-  intros; assert (H1 := derivable_pt_lim_D_in f1 (fun y:R => l1) x).
-  elim H1; intros.
-  assert (H4 := H3 H).
-  assert (H5 := derivable_pt_lim_D_in f2 (fun y:R => l2) (f1 x)).
-  elim H5; intros.
-  assert (H8 := H7 H0).
-  clear H1 H2 H3 H5 H6 H7.
-  assert (H1 := derivable_pt_lim_D_in (f2 o f1)%F (fun y:R => l2 * l1) x).
-  elim H1; intros.
-  clear H1 H3; apply H2.
-  unfold comp;
-    cut
-      (D_in (fun x0:R => f2 (f1 x0)) (fun y:R => l2 * l1)
-        (Dgf no_cond no_cond f1) x ->
-        D_in (fun x0:R => f2 (f1 x0)) (fun y:R => l2 * l1) no_cond x).
-  intro; apply H1.
-  rewrite Rmult_comm;
-    apply (Dcomp no_cond no_cond (fun y:R => l1) (fun y:R => l2) f1 f2 x);
-      assumption.
-  unfold Dgf, D_in, no_cond; unfold limit1_in;
-    unfold limit_in; unfold dist; simpl;
-      unfold R_dist; intros.
-  elim (H1 eps H3); intros.
-  exists x0; intros; split.
-  elim H5; intros; assumption.
-  intros; elim H5; intros; apply H9; split.
-  unfold D_x; split.
-  split; trivial.
-  elim H6; intros; unfold D_x in H10; elim H10; intros; assumption.
-  elim H6; intros; assumption.
+  unfold derivable_pt; intro.
+  exists 1.
+  apply derivable_pt_lim_id.
+Qed.
+
+Lemma derivable_pt_comp :
+  forall f1 f2 (x:R),
+    derivable_pt f1 x -> derivable_pt f2 (f1 x) -> derivable_pt (f2 o f1) x.
+Proof.
+  unfold derivable_pt; intros f1 f2 x X X0.
+  elim X; intros.
+  elim X0; intros.
+  exists (x1 * x0).
+  apply derivable_pt_lim_comp; assumption.
+Qed.
+
+Lemma derivable_pt_xeq:
+  forall (f : R -> R) (x1 x2 : R), x1=x2 -> derivable_pt f x1 -> derivable_pt f x2.
+Proof.
+  intros f x1 x2 Heq H.
+  subst; assumption.
+Qed.
+
+Lemma derivable_pt_opp :
+  forall (f : R -> R) (x:R), derivable_pt f x -> derivable_pt (- f) x.
+Proof.
+  intros f x H.
+  unfold derivable_pt in H.
+  destruct H as [l H]; exists (-l).
+  apply derivable_pt_lim_opp; assumption.
+Qed.
+
+Lemma derivable_pt_opp_rev:
+  forall (f : R -> R) (x : R), derivable_pt (- f) x -> derivable_pt f x.
+Proof.
+  intros f x H.
+  unfold derivable_pt in H.
+  destruct H as [l H]; exists (-l).
+  apply derivable_pt_lim_opp_rev.
+  rewrite Ropp_involutive; assumption.
+Qed.
+
+Lemma derivable_pt_mirr:
+  forall (f : R -> R) (x : R), derivable_pt f (-x) -> derivable_pt (mirr_fct f) x.
+Proof.
+  intros f x H.
+  unfold derivable_pt in H.
+  destruct H as [l H]; exists (-l).
+  apply derivable_pt_lim_mirr_fwd.
+  rewrite Ropp_involutive; assumption.
+Qed.
+
+Lemma derivable_pt_mirr_rev:
+  forall (f : R -> R) (x : R), derivable_pt (mirr_fct f) (- x) -> derivable_pt f x.
+Proof.
+  intros f x H.
+  unfold derivable_pt in H.
+  destruct H as [l H]; exists (-l).
+  apply derivable_pt_lim_mirr_rev.
+  rewrite Ropp_involutive; assumption.
+Qed.
+
+Lemma derivable_pt_mirr_prem:
+  forall (f : R -> R) (x : R), derivable_pt (mirr_fct f) x -> derivable_pt f (-x).
+Proof.
+  intros f x H.
+  unfold derivable_pt in H.
+  destruct H as [l H]; exists (-l).
+  apply derivable_pt_lim_mirr_rev.
+  repeat rewrite Ropp_involutive; assumption.
 Qed.
 
 Lemma derivable_pt_plus :
@@ -794,15 +915,6 @@ Proof.
   elim X0; intros.
   exists (x0 + x1).
   apply derivable_pt_lim_plus; assumption.
-Qed.
-
-Lemma derivable_pt_opp :
-  forall f (x:R), derivable_pt f x -> derivable_pt (- f) x.
-Proof.
-  unfold derivable_pt; intros f x X.
-  elim X; intros.
-  exists (- x0).
-  apply derivable_pt_lim_opp; assumption.
 Qed.
 
 Lemma derivable_pt_minus :
@@ -843,28 +955,36 @@ Proof.
   apply derivable_pt_lim_scal; assumption.
 Qed.
 
-Lemma derivable_pt_id : forall x:R, derivable_pt id x.
-Proof.
-  unfold derivable_pt; intro.
-  exists 1.
-  apply derivable_pt_lim_id.
-Qed.
-
 Lemma derivable_pt_Rsqr : forall x:R, derivable_pt Rsqr x.
 Proof.
   unfold derivable_pt; intro; exists (2 * x).
   apply derivable_pt_lim_Rsqr.
 Qed.
 
-Lemma derivable_pt_comp :
-  forall f1 f2 (x:R),
-    derivable_pt f1 x -> derivable_pt f2 (f1 x) -> derivable_pt (f2 o f1) x.
+(** ** Rules for derivable (derivability on whole domain) *)
+
+Lemma derivable_id : derivable id.
 Proof.
-  unfold derivable_pt; intros f1 f2 x X X0.
-  elim X; intros.
-  elim X0; intros.
-  exists (x1 * x0).
-  apply derivable_pt_lim_comp; assumption.
+  unfold derivable; intro; apply derivable_pt_id.
+Qed.
+
+Lemma derivable_comp :
+  forall f1 f2, derivable f1 -> derivable f2 -> derivable (f2 o f1).
+Proof.
+  unfold derivable; intros f1 f2 X X0 x.
+  apply (derivable_pt_comp _ _ x (X _) (X0 _)).
+Qed.
+
+Lemma derivable_opp : forall f, derivable f -> derivable (- f).
+Proof.
+  unfold derivable; intros f X x.
+  apply (derivable_pt_opp _ x (X _)).
+Qed.
+
+Lemma derivable_mirr : forall f, derivable f -> derivable (mirr_fct f).
+Proof.
+  unfold derivable; intros f X x.
+  apply (derivable_pt_mirr _ x (X _)).
 Qed.
 
 Lemma derivable_plus :
@@ -872,12 +992,6 @@ Lemma derivable_plus :
 Proof.
   unfold derivable; intros f1 f2 X X0 x.
   apply (derivable_pt_plus _ _ x (X _) (X0 _)).
-Qed.
-
-Lemma derivable_opp : forall f, derivable f -> derivable (- f).
-Proof.
-  unfold derivable; intros f X x.
-  apply (derivable_pt_opp _ x (X _)).
 Qed.
 
 Lemma derivable_minus :
@@ -907,21 +1021,87 @@ Proof.
   apply (derivable_pt_scal _ a x (X _)).
 Qed.
 
-Lemma derivable_id : derivable id.
-Proof.
-  unfold derivable; intro; apply derivable_pt_id.
-Qed.
-
 Lemma derivable_Rsqr : derivable Rsqr.
 Proof.
   unfold derivable; intro; apply derivable_pt_Rsqr.
 Qed.
 
-Lemma derivable_comp :
-  forall f1 f2, derivable f1 -> derivable f2 -> derivable (f2 o f1).
+(** ** Rules for derive_pt (derivative function on whole domain) *)
+
+Lemma derive_pt_id : forall x:R, derive_pt id x (derivable_pt_id _) = 1.
 Proof.
-  unfold derivable; intros f1 f2 X X0 x.
-  apply (derivable_pt_comp _ _ x (X _) (X0 _)).
+  intros.
+  apply derive_pt_eq_0.
+  apply derivable_pt_lim_id.
+Qed.
+
+Lemma derive_pt_comp :
+  forall f1 f2 (x:R) (pr1:derivable_pt f1 x) (pr2:derivable_pt f2 (f1 x)),
+    derive_pt (f2 o f1) x (derivable_pt_comp _ _ _ pr1 pr2) =
+    derive_pt f2 (f1 x) pr2 * derive_pt f1 x pr1.
+Proof.
+  intros.
+  assert (H := derivable_derive f1 x pr1).
+  assert (H0 := derivable_derive f2 (f1 x) pr2).
+  assert
+    (H1 := derivable_derive (f2 o f1)%F x (derivable_pt_comp _ _ _ pr1 pr2)).
+  elim H; clear H; intros l1 H.
+  elim H0; clear H0; intros l2 H0.
+  elim H1; clear H1; intros l H1.
+  rewrite H; rewrite H0; apply derive_pt_eq_0.
+  assert (H3 := proj2_sig pr1).
+  unfold derive_pt in H; rewrite H in H3.
+  assert (H4 := proj2_sig pr2).
+  unfold derive_pt in H0; rewrite H0 in H4.
+  apply derivable_pt_lim_comp; assumption.
+Qed.
+
+Lemma derive_pt_opp :
+  forall f (x:R) (pr1:derivable_pt f x),
+    derive_pt (- f) x (derivable_pt_opp _ _ pr1) = - derive_pt f x pr1.
+Proof.
+  intros.
+  apply derive_pt_eq_0.
+  apply derivable_pt_lim_opp_fwd.
+  rewrite Ropp_involutive.
+  apply (derive_pt_eq_1 _ _ _ pr1).
+  reflexivity.
+Qed.
+
+Lemma derive_pt_opp_rev :
+  forall f (x:R) (pr1:derivable_pt (- f) x),
+    derive_pt (- f) x pr1 = - derive_pt f x (derivable_pt_opp_rev _ _ pr1).
+Proof.
+  intros.
+  apply derive_pt_eq_0.
+  apply derivable_pt_lim_opp_fwd.
+  rewrite Ropp_involutive.
+  apply (derive_pt_eq_1 _ _ _ (derivable_pt_opp_rev _ _ pr1)).
+  reflexivity.
+Qed.
+
+Lemma derive_pt_mirr :
+  forall f (x:R) (pr1:derivable_pt f (-x)),
+    derive_pt (mirr_fct f) x (derivable_pt_mirr _ _ pr1) = - derive_pt f (-x) pr1.
+Proof.
+  intros.
+  apply derive_pt_eq_0.
+  apply derivable_pt_lim_mirr_fwd.
+  rewrite Ropp_involutive.
+  apply (derive_pt_eq_1 _ _ _ pr1).
+  reflexivity.
+Qed.
+
+Lemma derive_pt_mirr_rev :
+  forall f (x:R) (pr1:derivable_pt (mirr_fct f) x),
+    derive_pt (mirr_fct f) x pr1 = - derive_pt f (-x) (derivable_pt_mirr_prem f x pr1).
+Proof.
+  intros.
+  apply derive_pt_eq_0.
+  apply derivable_pt_lim_mirr_fwd.
+  rewrite Ropp_involutive.
+  apply (derive_pt_eq_1 _ _ _ (derivable_pt_mirr_prem f x pr1)).
+  reflexivity.
 Qed.
 
 Lemma derive_pt_plus :
@@ -943,21 +1123,6 @@ Proof.
   assert (H4 := proj2_sig pr2).
   unfold derive_pt in H0; rewrite H0 in H4.
   apply derivable_pt_lim_plus; assumption.
-Qed.
-
-Lemma derive_pt_opp :
-  forall f (x:R) (pr1:derivable_pt f x),
-    derive_pt (- f) x (derivable_pt_opp _ _ pr1) = - derive_pt f x pr1.
-Proof.
-  intros.
-  assert (H := derivable_derive f x pr1).
-  assert (H0 := derivable_derive (- f)%F x (derivable_pt_opp _ _ pr1)).
-  elim H; clear H; intros l1 H.
-  elim H0; clear H0; intros l2 H0.
-  rewrite H; apply derive_pt_eq_0.
-  assert (H3 := proj2_sig pr1).
-  unfold derive_pt in H; rewrite H in H3.
-  apply derivable_pt_lim_opp; assumption.
 Qed.
 
 Lemma derive_pt_minus :
@@ -1027,13 +1192,6 @@ Proof.
   apply derivable_pt_lim_scal; assumption.
 Qed.
 
-Lemma derive_pt_id : forall x:R, derive_pt id x (derivable_pt_id _) = 1.
-Proof.
-  intros.
-  apply derive_pt_eq_0.
-  apply derivable_pt_lim_id.
-Qed.
-
 Lemma derive_pt_Rsqr :
   forall x:R, derive_pt Rsqr x (derivable_pt_Rsqr _) = 2 * x.
 Proof.
@@ -1042,28 +1200,8 @@ Proof.
   apply derivable_pt_lim_Rsqr.
 Qed.
 
-Lemma derive_pt_comp :
-  forall f1 f2 (x:R) (pr1:derivable_pt f1 x) (pr2:derivable_pt f2 (f1 x)),
-    derive_pt (f2 o f1) x (derivable_pt_comp _ _ _ pr1 pr2) =
-    derive_pt f2 (f1 x) pr2 * derive_pt f1 x pr1.
-Proof.
-  intros.
-  assert (H := derivable_derive f1 x pr1).
-  assert (H0 := derivable_derive f2 (f1 x) pr2).
-  assert
-    (H1 := derivable_derive (f2 o f1)%F x (derivable_pt_comp _ _ _ pr1 pr2)).
-  elim H; clear H; intros l1 H.
-  elim H0; clear H0; intros l2 H0.
-  elim H1; clear H1; intros l H1.
-  rewrite H; rewrite H0; apply derive_pt_eq_0.
-  assert (H3 := proj2_sig pr1).
-  unfold derive_pt in H; rewrite H in H3.
-  assert (H4 := proj2_sig pr2).
-  unfold derive_pt in H0; rewrite H0 in H4.
-  apply derivable_pt_lim_comp; assumption.
-Qed.
+(** ** Definition and derivative of power function with natural number exponent *)
 
-(* Pow *)
 Definition pow_fct (n:nat) (y:R) : R := y ^ n.
 
 Lemma derivable_pt_lim_pow_pos :
@@ -1141,6 +1279,8 @@ Proof.
   apply derivable_pt_lim_pow.
 Qed.
 
+(** ** Irrelevance of derivability proof for derivative *)
+
 Lemma pr_nu :
   forall f (x:R) (pr1 pr2:derivable_pt f x),
     derive_pt f x pr1 = derive_pt f x pr2.
@@ -1149,6 +1289,16 @@ Proof.
   apply (uniqueness_limite f x x0 x1 H0 H1).
 Qed.
 
+(** In dependently typed environments it is sometimes hard to rewrite.
+    Having pr_nu for separate x with a proof that they are equal helps. *)
+
+Lemma pr_nu_xeq :
+  forall f (x1 x2:R) (pr1:derivable_pt f x1) (pr2:derivable_pt f x2),
+    x1 = x2 -> derive_pt f x1 pr1 = derive_pt f x2 pr2.
+Proof.
+  intros f x1 x2 H1 H2 Heq.
+  subst. apply pr_nu.
+Qed.
 
 (************************************************************)
 (** *           Local extremum's condition                  *)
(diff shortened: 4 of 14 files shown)


ejgallego:
@thery please rebase your commits and provide a meaningful commit messages detailing the changes, as well a CHANGES entry if needed.

ejgallego:
Thanks a lot @thery , this needs some rebase / squah magic still tho.

thery:
@ejgallego I think it looks ok now

maximedenes:
cc @MSoegtropIMC

maximedenes:
@MSoegtropIMC has developped something similar, let's see how we should organize the merge.

MSoegtropIMC:
yes, I wrote a fairly complete library also including the derivatives of asin and acos and quite a few trogonometric identities. I didn't create a PR as yet because I wanted to discuss with @silene first how to add these to CoqInterval and do it. I guess we should merge this.

MSoegtropIMC:
@thery: I had a look at your definitions and think yours are more elegant than mine (I used the "natural" definition with atan and added the cases for -1 and 1, which gets a bit messy at some points). I would suggest we merge this and I add the additional stuff I have (derivatives of asin and acos and quite a few trig identities and other useful stuff) based on your definitions.

thery:
@MSoegtropIMC perfect! There was no claim at all that what I've being added is complete but it was definitively lacking...

MSoegtropIMC:
@thery : I think you can put me in as reviewer and I will approve and merge this. I also discussed possible implementation strategies with @silene last week in Paris especially regarding the question how one can add these functions to CoqInterval. I would think your definitions are compatible with what silene suggested. I don't think silene will have time to review this in the next few weeks.

MSoegtropIMC:
@thery : if you want me to approve this, please also request me as reviewer.

Zimmi48:
@MSoegtropIMC and @thery: Should we create a @coq/reals-library-maintainers team and put you in there in addition to @silene? @MSoegtropIMC It's great to hear that you had fruitful discussions with @silene. It would be good to have your thoughts about #9185, since @silene apparently hasn't had time to comment this PR yet.

MSoegtropIMC:
> Should we create a @coq/reals-library-maintainers team and put you in there in addition to @silene?

It would make sense, yes. I did work a lot with real analysis in the last 6 months and will continue to do so for at least 12 months,likely much longer, and am fairly familiar meanwhile with the setup of the standard library and the extensions (Coquelicot, CoqInterval, Gappa).

MSoegtropIMC:
Extending the discussion on what should go into the standard library in #9185 on this issue: I think it should somehow be "rounded up or closed". If there is a sin, cos, tan and atan function, there is no good reason for not having asin and acos. And if one has asin and acos, one should also have the derivatives.

If one should have a bunch of trigonometric identities in here is a different question. Here the issue is that maybe 1/3 to 1/2 of a "rounded up" set of these identities are required anyway to prove required properties of the functions. Then in the sense that it should be "rounded up" it would make sense to add a rounded up set, because everything else is way annoying for the user (unless one makes the identities required for the proofs local).

Yet another question is if e.g. trigonometric addition theorems should be in the standard library or in a separate trigonometry library. IMHO everything which can be found in a basic school formula booklet (what is used for high school exams) is basic mathematics and should be in the standard library. So I would add the basic addition theorem (like sin(a+b) = cos(a)*sin(b)+sin(a)*cos(b)) but nothing fancy (like sin(a)^3=(3*sin(a)-sin(3*a))/4)

Maybe this rule "what everybody has learned in high school belongs in the standard library" is a good first indication to make such decisions. I would say it fits pretty well to the current layout of the library: it contains what everyone learns in high school about numbers, functions, relations, algebra, ... and what one learns in basic (high school level) computer science courses and the infrastructure required to define all this in CIC. Geometry is currently missing - I am not sure if I like this.

In this sense IMHO #9185 doesn't belong into the standard library, since the Lebesgue integral is taught in elementary university courses.

thery:
@Zimmi48 @MSoegtropIMC I agree,  #9185 look more to me as a new library and should be added as a contribution.

Zimmi48:
@MSoegtropIMC You said you would merge this PR but you apparently forgot.
@thery This now needs a rebase to solve a conflict in `CHANGES`.

thery:
> @thery This now needs a rebase to solve a conflict in `CHANGES`.

done

MSoegtropIMC:
@Zimmi48 : I didn't forget, I was just too busy (mostly with the INRIA evaluation after work).

maximedenes:
@MSoegtropIMC I can merge, if it helps.

MSoegtropIMC:
> @MSoegtropIMC I can merge, if it helps.

I didn't have time to review it again after I discussed it with silene. I planned to do it tomorrow evening. The question is how easy it will be to add asin and acos with the provided definition to CoqInterval (and also how easy it will be to adjust my proofs of the derivatives). I didn't want to add a definition which later proves suboptimal in this regard. Not that I think it would be likely to happen, but I didn't think it through.

Do you think it would be n issue if we would later have to change the definitions of asin and acos, say in 2 weeks?

maximedenes:
> Do you think it would be n issue if we would later have to change the definitions of asin and acos, say in 2 weeks?

Well the problem with this scenario is that 8.10 will be released with the suboptimal definitions. So changing it afterwards might be a bit problematic.

Let's delay, then.

MSoegtropIMC:
> Well the problem with this scenario is that 8.10 will be released with the suboptimal definitions.

I hoped to add my additions (derivatives and some more useful lemmas) on top. I am not sure if the release process has to be that strict for library stuff which can be checked by Coq - in the end one can't really mess up such lemmas. I was hoping the timing for such stuff (or addon libraries in the installer for that matter) is a bit more relaxed compared to changes in the Coq OCaml code, but that we need to discuss now what shall go in.

maximedenes:
> I am not sure if the release process has to be that strict for library stuff which can be checked by Coq - in the end one can't really mess up such lemmas.

Well, the point of the process is not just to avoid messing up stuff. It is also to ensure convergence and stability. Being "more relaxed" usually means putting pressure on the release process and the release manager, in my experience.

MSoegtropIMC:
> It is also to ensure convergence and stability.

I understand the constraints and I know it is a tough job (thanks for all the release managers!). Still I think that not everything needs to be fixed at the same date depending on how likely it is to cause problems. Anyway I try to get the review done tomorrow and my stuff on top over the weekend. I hope this is ok then for 8.10. If I can get asin and acos into CoqInterval until then I can't tell - I expect that I can redirect the lemmas required by CoqInterval to the atan lemmas with the definition of asin and acos as is, but I am not sure how tough it will get at the splicing point. If I can manage I hope this is also ok for 8.10. if not CoqInterval support will come later.

maximedenes:
@MSoegtropIMC I certainly didn't mean to make you rush. In fact, since there are still design questions (what's the optimal definition) one day after the freeze is a pretty strong indicator that this should be moved to 8.11. And that's fine!

If you mean that 6 months seem like a lot to wait for, we should discuss the length of the release cycle, rather than starting to rush before freeze dates or even trying to push them back.

Zimmi48:
Indeed, I agree with Maxime. We should discuss shortening the release cycle even more, since people continue to view missing a freeze as a problem. But don't rush things, that's almost always a bad idea.

Zimmi48:
See also https://github.com/coq/coq/issues/9664#issuecomment-468211793.

MSoegtropIMC:
Well I can live with my own builds to add these functions for 6 months. I am not sure about @vbgl . I guess I don't want to be the "bad guy" holding his PR back (which was early enough, the release schedule just colided a bit with the INRIA evaluation in my time table).

But I still think that all the design questions can easily be answered by Coq in this case, so I don't really see a problem from rushing this a bit. If there is an elegant proof fine. If not, it needs to be delayed anyway.

thery:
As the owner of this pull request, just my two cents.
For  me it was an innocent pull request, just pushing what I think should be in the standard library
If it does not make it for 8.10, fine with me.
If it does and someone later changes some of my definitions for better  ones, fine with me too.

silene (dismissed):
Feature-wise, I am perfectly happy. I would like some renaming before merging:

- `atan_left_inv` -> `atan_tan`
- `asin_right_inv` -> `sin_asin`
- `asin_left_inv` -> `asin_sin`
- `asin_Vsqrt2` -> `asin_inv_sqrt2`
- `acos_right_inv` -> `cos_acos`
- `acos_left_inv` -> `acos_cos`
- `sin_is_inj` -> `sin_inj`
- `coq_is_inj` -> `cos_inj`

For uniformity, `tan_is_inj` should be renamed to `tan_inj` and a compatibility notation should be added. Similarly, `tan_right_inv` should be renamed to `tan_atan`.

(Note: this was discussed with @thery.)

thery:
@silene renaming done
@rnrand subsume #10279

MSoegtropIMC:
@thery : my apologizes for leaving this for months - I was simply too busy with CoqIDE and 8.10. I will pick this up now, that is see what needs to be done to make asin and acos computable in CoqInterval and also add the derivative and some trig lemmas from my version.

MSoegtropIMC:
P.S.: I plan separate PRs for this and my changes on top, but I want to upfront check if the things I want to add are easier with your or with my definition of asin and acos.

MSoegtropIMC:
I started out and compared what lemmas I have on top of yours. This are:
```
Lemma derivable_pt_asin : forall x, -1 < x < 1 -> derivable_pt asin x.
Lemma derivable_pt_acos : forall x, -1 < x < 1 -> derivable_pt acos x.
(plus a bunch of generic lemmas on deriviatives of falling functions)
Lemma derive_pt_asin : forall (x : R) (Hxrange : -1 < x < 1),
   derive_pt asin x (derivable_pt_asin x Hxrange) = 1/sqrt(1-x²).
Lemma derive_pt_acos : forall (x : R) (Hxrange : -1 < x < 1),
   derive_pt acos x (derivable_pt_acos x Hxrange) = -1/sqrt(1-x²).

Lemma sin2_bound: forall α:R, 0 <= (sin α)² <= 1.
Lemma cos2_bound: forall α:R, 0 <= (cos α)² <= 1.
Lemma cos_of_sin : forall α:R, cos α >=0 -> cos α = sqrt(1 - (sin α)²).
Lemma cos_of_sin_opp : forall α:R, cos α <=0 -> cos α = - sqrt(1 - (sin α)²).
Lemma cos_of_sin_Rabs : forall α:R, Rabs (cos α) = sqrt(1 - (sin α)²).
Lemma sin_of_cos : forall α:R, sin α >=0 -> sin α = sqrt(1 - (cos α)²).
Lemma sin_of_cos_opp : forall α:R, sin α <=0 -> sin α = - sqrt(1 - (cos α)²).
Lemma sin_of_cos_Rabs : forall α:R, Rabs (sin α) = sqrt(1 - (cos α)²).
Lemma tan_of_sin: forall α:R, 0 <= cos α -> tan α = sin α / sqrt (1-(sin α)²).
Lemma tan_of_sin_opp: forall α:R, 0 > cos α -> tan α = - (sin α / sqrt (1-(sin α)²)).
Lemma tan_of_cos: forall α:R, 0 <= sin α -> tan α = sqrt (1-(cos α)²) / cos α.
Lemma tan_of_cos_opp: forall α:R, 0 >= sin α -> tan α = - sqrt (1-(cos α)²) / cos α.
Lemma sin_of_tan: forall α:R, 0 < cos α -> sin α = tan α / sqrt (1+(tan α)²).
Lemma cos_of_tan: forall α:R, 0 < cos α -> cos α = 1 / sqrt (1+(tan α)²).
Lemma sin_of_atan: forall x:R, sin(atan(x))=x/sqrt(1+x²).
Lemma cos_of_atan: forall x:R, cos(atan(x))=1/sqrt(1+x²).
Lemma cos_of_atan: forall x:R, cos(atan(x))=1/sqrt(1+x²).
Lemma cos_of_asin: forall x : R, -1 <= x <= 1 -> cos (asin x) = sqrt (1 - x²).
Lemma tan_of_asin: forall x : R, -1 <= x <= 1 -> tan (asin x) = x / sqrt (1-x²).
Lemma tan_of_acos: forall x : R, -1 <= x <= 1 -> tan (acos x) = sqrt (1-x²) / x.
```
The derivative lemmas I put into Ratan.v.
The identities I put into a new file "Rtrigo_ident.v".
I thought about also adding half and double angle lemmas.

I will now check how easy or hard the derivative lemmas are to prove with your definitions of asin and acos. The other lemmas involving asin and acos should not depend (much) on the definition.

MSoegtropIMC:
@thery : I had some time to look at e.g. the derivative proofs with your and my definition. Btw, my definition of asin reads:
```
Definition asin_open s := atan(s/sqrt(1-s²)).

Definition pre_asin (y : R) : {x : R | -PI/2 <= x <= PI/2 /\ (-1 <= y <= 1 -> sin x = y)}.
Proof.
  destruct (Rcase_abs (-1-y)) as [Hylow|Hylow];
  destruct (Rcase_abs (y-1)) as [Hyhigh|Hyhigh].

  - exists (asin_open y); split.
    + pose proof asin_open_bound y; lra.
    + intros Hyrange; apply asin_open_right_inv; lra.

  - exists (PI/2); split.
    + pose proof PI_RGT_0; lra.
    + intros Hyrange; rewrite sin_PI2; lra.

  - exists (-(PI/2)); split.
    + pose proof PI_RGT_0; lra.
    + intros Hyrange; rewrite sin_antisym; rewrite sin_PI2; lra.

  - contradict Hylow. lra.
Defined.

Definition asin x := let (v, _) := pre_asin x in v.
```
For other readers, the definition of @thery is:
```
Definition asin x :=
  match total_order_T x 0 with
  | inleft (left _) => - PI/2 - atan (sqrt (1 - x * x) / x)
  | inleft (right _) => 0
  | inright _  => PI/2 - atan (sqrt (1 - x * x) / x)
  end.
```
In comparison my definition might look overly complicated, but at one point one has to do the case analysis for the full range sin_asin lemma anyway, so one doesn't gain much by a simpler definition. The proofs about asin which are valid for the full range are in summary of similar complexity with your and with my definition. The additional trigonometric identity proofs are identical.

But for proofs on derivatives and likely also on power series (still working on these) which are valid only in the reduced range `-1 < x < 1` my definition has a trivial case analysis because the special case of my definition, -1 and 1, are excluded. With your definition I end up with two non trivial cases (positive and negative). It is not that bad, but one needs quite a few one sided bounds results and cause of the division in the argument of atan these typically cannot be solved by lra and need manual derivation. Since my definition has no case distinction in the valid range, I don't need such results at all. As I said, not difficult but lengthy.

For acos it gets a bit worse because some derivative lemmas I use (e.g. `derivable_pt_recip_interv`) require increasing functions, so I did a trick and defined an argument sign swap function
```
Definition opp_arg (f : R -> R) := fun (x : R) => f (- x).
```
and first do the derivative lemmas for (opp cos). This doesn't really work with your definition. This is also not that bad, probably one should do it properly and prove a lemma like `derivable_pt_recip_interv` for decreasing functions.

In summary I think a definition which uses -1 and 1 as special cases and is continuous in between but has explicit tests like your definition rather than using a dependent definition like I did would be easiest to handle. The discontinuity at 0 in your definition might make things a bit tricky for the power series proofs around 0.

Please let me know, what you think.

thery:
Maybe another way
```
Definition Req1_spec x : {x = 1} + {x = -1} + {x <> 1 /\ x <> -1}.
Proof.
destruct (Req_EM_T (Rabs x) 1) as [H|H].
  apply inleft.
  split_Rabs; [right|left]; auto.
  now rewrite <-(Ropp_involutive x), H.
apply inright.
split_Rabs; split; auto with real.
  apply Rlt_not_eq; apply Rlt_trans with (r2 := 0); auto with real.
  now contradict H; rewrite H; unfold IZR; simpl; rewrite Ropp_involutive.
apply not_eq_sym.
apply Rlt_not_eq; apply Rlt_le_trans with (r2 := 0); auto with real.
Defined.

Definition asin x := 
  match Req1_spec x with
  | inleft (left _) => PI/2
  | inleft (right _) => -PI/2
  | inright _  => atan (x / sqrt(1 - x²))
  end.
```

thery:
> one should do it properly and prove a lemma like `derivable_pt_recip_interv`

this sounds better

MSoegtropIMC:
> Maybe another way

Yes, this is a slightly more elegant version of what I had in mind. I also thought about using 5 cases, following your style:
```
{x < -1} + {x = -1} + {-1 < x < 1} + {x = 1} + { x > 1}
```
because one has to distinguish these cases in most of the advanced proofs anyway, but I came to the conclusion that in the definition the 3 cases are better. One can still have a lemma for such a destruction and a lemma that the atan expression is 0 out of range.

How shall we proceed? May I change your PR/branch to the definition you suggested and adjust the proofs or would you prefer to do this yourself? I would suggest to merge it then and I do the additional lemmas as separate PR and final PR in CoqInterval to make it compute there.

P.S.: I think you might be underestimating the power of lra. The proof of the spec lemma can be done with:
```
Definition Req1_spec x : {x = 1} + {x = -1} + {x <> 1 /\ x <> -1}.
Proof.
  destruct (Req_EM_T (Rabs x) 1) as [H|H].
  - left; split_Rabs; [right|left]; lra. 
  - right; split_Rabs; split; lra.
Defined.
```

thery:
> May I change your PR/branch to the definition you suggested and adjust the proofs

If you can do, it it will be great. I will have a look before merging.

MSoegtropIMC:
> > one should do it properly and prove a lemma like `derivable_pt_recip_interv`
> 
> this looks better

Yes, I thought the `opp_arg` trick was super smart, but I needed quite a few lemmas on that in the end and getting the arguments of derivable_pt_recip_interv right gets tricky with the opp_arg hack. A direct equivalent of derivable_pt_recip_interv for decreasing functions is for sure easier to use and more generic.

MSoegtropIMC:
> If you can do, it it will be great. I will have a look before merging.

Fine, I will do the changes and push to your branch.

thery:
> in the definition the 3 cases are better

Initially I thought that having only one special case on 0 was better, but with 3 cases as you
have, we have anyway asin (- x) = - asin(x),  so one can always work with x >=0 and only take care on the problem on the left of 1.

> I think you might be underestimating the power of lra. The proof of the spec lemma can be done with:
> 
> ```
> Definition Req1_spec x : {x = 1} + {x = -1} + {x <> 1 /\ x <> -1}.
> Proof.
>   destruct (Req_EM_T (Rabs x) 1) as [H|H].
>   - left; split_Rabs; [right|left]; lra. 
>   - right; split_Rabs; split; lra.
> Defined.
> ```

Much better!!

thery:
> Definition Req1_spec x : {x = 1} + {x = -1} + {x <> 1 /\ x <> -1}.
> Proof.
>   destruct (Req_EM_T (Rabs x) 1) as [H|H].
>   - left; split_Rabs; [right|left]; lra. 
>   - right; split_Rabs; split; lra.
> Defined.



> Much better!!

one liner 

```
Definition Req1_spec x : {x = 1} + {x = -1} + {x <> 1 /\ x <> -1}.
Proof.
destruct (Req_EM_T (Rabs x) 1) as [H|H]; split_Rabs; intuition lra.
Defined.

**Comment by MSoegtropIMC** [comment]
> Definition Req1_spec x : {x = 1} + {x = -1} + {x <> 1 /\ x <> -1}.
> Proof.
> destruct (Req_EM_T (Rabs x) 1); split_Rabs; intuition lra.
> Qed.

Thanks, very nice! I thought about how to do that, but I couldn't find an elegant solution. `intuition` is not something I have on my radar - I guess I should!

It would be a nice student exercise to polish some of the proofs in Reals. It does make a difference for learning how to write similar proofs. Some of the proofs are not really inviting.

**Comment by thery** [comment]
This looks even better

```
Definition Rbetween_dec x y z : {z <= x} + {x <= y} + {y < x < z}.
Proof.
destruct (Rle_dec z x); destruct (Rle_dec x y); intuition lra.
Qed.

Definition asin x := 
  match Rbetween_dec x (-1) 1 with
  | inleft (left _) => PI/2
  | inleft (right _) => - (PI/2)
  | inright _  => atan (x / sqrt(1 - x ^ 2))
  end.
```

**Comment by thery** [comment]
It is kind of neverending game...
```
Definition Rbetween_dec x y z : {x <= y} + {y < x < z} + {z <= x} .
Proof.
destruct (Rle_dec z x); destruct (Rle_dec x y); intuition lra.
Qed.

Definition f_bextend x y z (f g h : R -> R) := 
  match Rbetween_dec x y z with
  | inleft (left _) => f x
  | inleft (right _) => g x
  | inright _  => h x
  end.

Variant f_bextend_Spec (x y z : R) (f g h : R -> R) : R -> Type :=
| f_bextend_le_spec : x <= y -> f_bextend_Spec x y z f g h (f x)
| f_bextend_bt_spec : y < x < z -> f_bextend_Spec x y z f g h (g x)
| f_bextend_ge_spec : z <= x -> f_bextend_Spec x y z f g h (h x).

Lemma f_bextend_spec x y z f g h : f_bextend_Spec x y z f g h (f_bextend x y z f g h).
Proof.
unfold f_bextend.
now case Rbetween_dec; [intros [H|H]|intros H]; constructor.
Qed.

Definition asin x := 
  f_bextend x (-1) (1) (fun _ => -PI/2) 
                     (fun _ => atan (x / sqrt(1 - x ^ 2))) 
                     (fun _ => PI/2).

Lemma asin_spec x : 
   f_bextend_Spec x (-1) (1) (fun _ => -PI/2) 
                           (fun _ => atan (x / sqrt(1 - x ^ 2))) 
                           (fun _ => PI/2) (asin x).
Proof. now apply f_bextend_spec. Qed.

Lemma asin_antisym x : asin (- x) = - (asin x).
Proof.
do 2 case asin_spec; intros H H1; try lra.
rewrite <- atan_opp.
replace ((-x) ^ 2) with (x ^ 2) by lra.
now unfold Rdiv; rewrite Ropp_mult_distr_l.
Qed.
```

**Comment by MSoegtropIMC** [comment]
> It is kind of neverending game...

I find this from a software engineering point of view already a bit over the top. It is not easier to read than your previous definition, and the reader has to ponder over questions like why it is
```(fun _ => atan (x / sqrt(1 - x ^ 2)))```
rather than
```(fun x => atan (x / sqrt(1 - x ^ 2)))```.

I would more go into the direction of *improving* readability (rather than improving writeability for a very limited audience) e.g. with:
```
Definition Rbetween_dec x lo up : {x <= lo} + {lo < x < up} + {up <= x}.
Proof.
destruct (Rle_dec up x); destruct (Rle_dec x lo); intuition lra.
Qed.

Notation "'sum3_1' p" := (inleft (left p)) (at level 0).
Notation "'sum3_2' p" := (inleft (right p)) (at level 0).
Notation "'sum3_3' p" := (inright p) (at level 0).

Definition asin x := 
  match Rbetween_dec x (-1) 1 with
  | sum3_1 _ => - (PI/2)
  | sum3_2 _ => atan (x / sqrt(1 - x ^ 2))
  | sum3_3 _ => PI/2
  end.
```
Note also the change in order of the sumbool entries, which I think is easier to read in the definition of Rbetween_dec as well as in the definition of asin.

I think one of the strength of Coq is that one can write down definitions and lemma statements in a way which is quite easy to read. For development of libraries this should be an important goal. Writing proofs in an understandable way is the cherry on the cake.

**Comment by herbelin** [comment]
That's a very interesting discussion. A comment in passing...

> Notation "'sum3_1' p" := (inleft (left p)) (at level 0).
Notation "'sum3_2' p" := (inleft (right p)) (at level 0).
Notation "'sum3_3' p" := (inright p) (at level 0).

If you are ready for notations, you could even go up to formulate `Rbetween_dec` using a ternary sum:
```
Inductive sum3 P Q R := sum3_1  (p : P) | sum3_2 (p : Q) | sum3_3 (p : R).
Notation "[ + P1 , P2 , P3 ]" := (sum3 P1 P2 P3) : type_scope.
Definition Rbetween_dec x lo up : [+ x <= lo , lo < x < up , up <= x].
```
or even reuse the existing synonym [`Datatypes.CompareSpecT`](https://coq.inria.fr/distrib/current/stdlib/Coq.Init.Datatypes.html#CompareSpecT).  (Notation here tentatively inspired from mathcomp's [scheme]( https://github.com/coq/coq/blob/4e679df3c15e5e554ff9ef85138f9c55396e9f0b/plugins/ssr/ssrbool.v#L449) for n-ary notations.)

**Comment by thery** [comment]
> `(fun x => atan (x / sqrt(1 - x ^ 2)))`.

You are right

> I find this from a software engineering point of view already a bit over the top.

I thought that extending a function is something that could be worth abstracting so
it could be reused for other functions than asin but maybe there is no other occurrence
in the library.

> Note also the change in order of the sumbool entries, which I think is easier to read in the definition of Rbetween_dec as well as in the definition of asin.

Perfect!

**Comment by thery** [comment]
> Notation "'sum3_1' p" := (inleft (left p)) (at level 0).
> Notation "'sum3_2' p" := (inleft (right p)) (at level 0).
> Notation "'sum3_3' p" := (inright p) (at level 0).

Unfortunately these notations must remain local

**Comment by thery** [comment]
Thinking over it.  Maybe using {}+{}+{} was not such a great idea,  `asin_spec`  gives us already
directly the 3 cases.

```
Definition asin x :=
  if Rle_dec x (-1) then - (PI / 2) else
  if Rle_dec 1 x then PI / 2 else 
  atan (x / sqrt(1 - x ^ 2)).

Variant asin_Spec (x : R) : R -> Type :=
| asin_p2_spec :  1 <= x -> asin_Spec x (PI / 2)
| asin_mp2_spec : x <= -1 -> asin_Spec x (- (PI / 2))
| asin_atan_spec : -1 < x < 1 -> asin_Spec x (atan (x / sqrt (1 - x ^ 2))).

Lemma asin_spec x : asin_Spec x (asin x).
Proof.
unfold asin; repeat case Rle_dec; intros; constructor; lra.
Qed.

Lemma asin_antisym x : asin (- x) = - (asin x).
Proof.
do 2 case asin_spec; intros H H1; try lra.
replace ((- x) ^ 2) with (x ^ 2) by lra.
now unfold Rdiv; rewrite <- atan_opp, Ropp_mult_distr_l.
Qed.
```

**Comment by MSoegtropIMC** [comment]
> Notation here tentatively inspired from mathcomp's [scheme](https://github.com/coq/coq/blob/4e679df3c15e5e554ff9ef85138f9c55396e9f0b/plugins/ssr/ssrbool.v#L449) for n-ary notations.

I agree that generic ternary sums (Datatypes.CompareSpecT is a bit too specialized in terms of naming) would be a useful feature. The match patterns for combinations of somor+sumbool are hard to read in my opinion. I agree that this is better solved with a separate definition rather than notations because notations only help for explicit matches, but e.g. not for destructing such a term. Do you think adding this would be agreeable?

Btw.: I thought about having a n-ary sum-vector, but it gets a bit involved to define it such that when destructed one ends up with n cases - one has to have constructors which just take a Prop and constructors which take a Prop and a witness and one has to make sure that each instance has exactly one constructor which takes a witness. I think this is best done with two inductives - as soon as one has one witness one switches from one type to the other. But half way through that I concluded that this likely wouldn't pass my criteria for a good compromise between understandability and flexibility. On the other hand this wouldn't if using such a construct has nice intuitive notations and works seamless with tactics like destruct.

**Comment by MSoegtropIMC** [comment]
> Definition asin x :=
>   if Rle_dec x (-1) then - (PI / 2) else
>   if Rle_dec 1 x then PI / 2 else 
>   atan (x / sqrt(1 - x ^ 2)).

this is what I had in mind originally. I agree that one should either use this or a true ternary sum as suggested by @herbelin. A true ternary sum has the advantage that destructs naturally deliver 3 cases and matches are easier to read. This advantage is a bit hidden in your proofs because they are well automated and the individual cases are not visible - you have just minor additional complexity in the case analysis. In more involved proofs this might be more visible. But if it is worth it under the line to use ternary sums is hard to tell - I think both solutions are good.

> Maybe using {}+{}+{} was not such a great idea

I agree that using sumor+sumbool doesn't give a substantial gain in readability - possibly even on the contrary because it is a bit harder to tell which case is which in a match. The two cases of Rle_dec  in a nested if might be more natural.

If it is agreeable to add a generic ternary sum to the library, I would go this way, otherwise I would follow your latest suggestion.

**Comment by thery** [comment]
Let's go with the Rle_dec definition for the moment.

**Comment by herbelin** [comment]
> If it is agreeable to add a generic ternary sum to the library, I would go this way, otherwise I would follow your latest suggestion.

The need for a primitive `or3` apparently came to MathComp developers too. The `or3` from `ssrbool.v` could just be moved to `Logic.v` and a `sum3` be added to `Datatypes.v`.

There is the question of notations though. From a "mathematician" point of view, I believe that we would like a notation like `A + B + C` to be treated up to associativity, which would mean as a ternary sum. So, the notation for `+` could be changed so that `A + B` (where `B` not a `+` any longer) is interpreted as binary, and `A + B + C` as ternary. We cannot of course define a n-ary sum for arbitrary n so we have to stop somewhere. Stopping at three seems reasonable, especially because trichotomy statements are very common. Then we need to parse `A₁ + A₂ + A₃ + ... + An` though. A solution would be to parse it as `sumn` for a n-ary definition of `sum` as you mention. Then if one wants explicit associativity, we explicitly use parentheses.

But then, I feel that we go beyond what can be done at the pure library level. To be able to reason over cases of a `sumn`, one would need notations say of the form:
```
Notation "`sum_` n x" := (inr .. (inr (inl x)) .._n).
```
with the meaning that `n`, parsed as a natural number, denotes the number of iteration of `inr`. (That's not the first time I see the need for such kind of notations though.)

Similarly, `destruct` should be taught that destructing a `sumn` should decompose it recursively.

Other design choices are possible: we could also have an option to `destruct` (say `destruct iteratively`) so that it recursively decomposes its argument on all its subterms which are in the same type.

We could also think at hard-wiring notions of isomorphisms and provide things like `destruct a seen as sum3` so that the isomorphism between `A + (B + C)` and ternary `A + B + C` is automatically applied.

Don't know which approaches would be the most promosing and general purpose enough though. Maybe proposing a `"`sum_` n x" := (inr .. (inr (inl x)) .._n)` kind of notation and a `destruct iteratively` as implementation projects would be a good first step.

**Comment by MSoegtropIMC** [comment]
> We cannot of course define a n-ary sum for arbitrary n so we have to stop somewhere

Well one can write 
```
Require Import Vector.

Inductive sumn (n : nat) (P : Vector.t Prop n) : Set :=
  sumi (i : Fin.t n) (p : Vector.nth P i) : sumn n P.

Import VectorNotations.
Import Fin.
Require Import Coq.Program.Equality.
Require Import ZArith.

(* Definition of a sumn lemma *)

Lemma lt_eq_lt_dec_1: forall n m : nat, sumn 3 [ n < m; n = m; m < n ].
Proof.
  intros n m.  
  pose proof Compare_dec.lt_eq_lt_dec n m as H.
  destruct H as [[H|H]|H].
  apply (sumi _ _ F1); assumption.
  apply (sumi _ _ (FS F1)); assumption.
  apply (sumi _ _ (FS (FS F1))); assumption.
Qed.

(* Using a sumn lemma in a Proof *)

Lemma lt_eq_lt_dec_2: forall n m : nat, {n < m} + {n = m} + {m < n}.
Proof.
  intros n m.  
  pose proof lt_eq_lt_dec_1 n m as H.
  destruct H.
  repeat dependent destruction i; simpl in p.
  - apply (inleft (left p)).
  - apply (inleft (right p)).
  - apply (inright p).
Qed.

(* Using a sumn lemma in a Definition *)

Definition ord (n m : nat) : Z :=
  match lt_eq_lt_dec_1 n m with
  | sumi _ _ F1 _ => -1
  | sumi _ _ (FS F1) _ => 0
  | sumi _ _ (FS (FS F1)) _ => 1
  | sumi _ _ _ _ => 0
  end.
```
This is not really nice, we would like to have only 3 cases. I actually don't understand why this typechecks:
```
Definition ord_typechecktest (n m : nat) : nat :=
  match lt_eq_lt_dec_1 n m with
  | sumi _ _ (FS (FS (FS F1))) _ => 1
  | sumi _ _ _ _ => 0
  end.
```
`lt_eq_lt_dec_1 n m` clearly returns a sumn 3 which expects a Fin.t 3 as 3rd argument and `(FS (FS (FS F1))` is clearly not a Fin.t 3 as this shows:
```
Definition Four_as_Fin3 : Fin.t 3 := FS (FS (FS F1)).```
```
If one could define a Fin.t such that Coq understands that a Fin.t 3 has only 3 cases, and if there would be a nice notation for Fin.t numbers, the vector based definition would not be that bad, except for the overhead of the dependent destruction.

P.S.: if other readers think that this gets to much off topic, please complain - we can continue elsewhere.

**Comment by MSoegtropIMC** [comment]
> Let's go with the Rle_dec definition for the moment.

Yes, I agree. Advanced definitions of ternary sums would get us to much off the road.

**Comment by MSoegtropIMC** [comment]
@thery : FYI: I forced push to rebase your branch to latest master. The discussed changes I will do as separate commit.

**Comment by herbelin** [comment]
> This is not really nice, we would like to have only 3 cases. I actually don't understand why this typechecks:
> 
> ```
> Definition ord_typechecktest (n m : nat) : nat :=
>   match lt_eq_lt_dec_1 n m with
>   | sumi _ _ (FS (FS (FS F1))) _ => 1
>   | sumi _ _ _ _ => 0
>   end.
> ```
> 
> `lt_eq_lt_dec_1 n m` clearly returns a sumn 3 (...).

This is the same issue as destructing a `Vector.t nat 0` which requires to consider the `cons` case too. The internal `match` construction of the Calculus of Inductive Constructions does not take into account the specific values of the indices of an inductive type (in your case `3`). Impossible cases should be handled by giving well-chosen pattern-matching clauses, what Equations is doing its way, what is also planned to be done with the standard pattern-matching algorithm, and what Agda supports directly. In your case, here is what the general recipe for compiling such a pattern-matching should give. It is a bit complex-looking terms, so I used notations to hide the complexity.:
```
Require Import Fin.

Notation "[[ n 'mustbe' p ]]" := (match n with p => nat | n => unit end) (p pattern).
Notation "[{ n = p => N }]" := (match n with p => N | n => tt end : [[ n mustbe p ]]) (n at level 0, p pattern at level 0).
Notation "[{ n < p => [ i ] kont }]" := (match n%nat as n return Fin.t n -> [[ n mustbe p ]] with p => fun i => kont | _ => fun _ => tt end i : [[ n mustbe p ]]) (n at level 0, p pattern).

Notation "'finmatch' i 'for' p 'with' | 'Fin_F1' => N | 'Fin_FS' i'' => kont 'end'" :=
  (match i in Fin .t n return [[ n mustbe (S p) ]] with
  | @F1 n' => [{ n' = p => N }]
  | @FS n' i' => [{ n' < p => [i'] (fun i'' => kont) i' }]
  end)
  (p pattern at level 0).

Definition impossible (i:Fin.t 0) : nat :=
  match i with end.

Definition fincase3 (i:Fin.t 3) N0 N1 N2 : nat :=
  finmatch i for 2 with
  | Fin_F1 => N0
  | Fin_FS i =>
    finmatch i for 1 with
    | Fin_F1 => N1
    | Fin_FS i =>
      finmatch i for 0 with
      | Fin_F1 => N2
      | Fin_FS i => impossible i
      end
    end
  end.
```

(I tried to make a notation which avoids repeating the integer at each step but the notation mechanism is not powerful enough.)

Again, Agda, or Equations, should be able to manage this automatically. The pattern-matching compilation algorithm is also in the process of being extended to insert all those `match` expansions hidden behind notations (the reference is Pierre Boutillier's PhD).

**Comment by MSoegtropIMC** [comment]
@herbelin : thanks for the detailed explanation! I experimented with your notations and managed to streamline it a bit. Also the resulting term is about only half the size with this notation and it works for any type:
```
Require Import Fin.

Declare Custom Entry finmatch_pattern.

Notation "'finmatch' i 'with' CN" := (CN i)
  (at level 0, CN custom finmatch_pattern at level 0).

Notation "| 'Fin' 'n' - r => C1 CN" :=
  ( fun i =>
    match i in Fin.t n return match n with S r => _ | _ => unit end with
    | @F1 r    => C1
    | @FS r i' => CN i'
    | _        => tt
    end
  )
  (in custom finmatch_pattern at level 0, C1 constr at level 0, r pattern at level 0, CN custom finmatch_pattern at level 0).

Notation "| 'Fin' 'n' - '0' => C1 'end'" :=
  ( fun i =>
    match i in Fin.t n return match n with 1 => _ | _ => unit end with
    | @F1 0    => C1
    | @FS 0 i' => match i' with end
    | _        => tt
    end
  )
  (in custom finmatch_pattern at level 0, C1 constr at level 0).

Definition fincase3nat (f : Fin.t 3) (N0 N1 N2 : nat) :=
  finmatch f with
  | Fin n-2 => N0
  | Fin n-1 => N1
  | Fin n-0 => N2
  end.
```
How about having this notation in the standard library?

**Comment by Zimmi48** [comment]
@coqbot is currently not working because Heroku is having some troubles: https://status.heroku.com/
I have manually started the pipeline for this PR but the results won't be reported back unless the troubles are solved by the time the pipeline has passed: https://gitlab.com/coq/coq/pipelines/79564504

**Comment by MSoegtropIMC** [comment]
@Zimmi48 : thanks, but this was anyway just a rebase - I keep getting interrupted with other things (last week the make file issues only visible on Windows). I expect that it takes a bit to finish this.

**Comment by MSoegtropIMC** [comment]
@thery : a style question: Ratan.v as is contains a mix of r² (a notation for Rsqr) and r^2. I wonder which should be preferred. I exclusively used ² / Rsqr.

**Comment by MSoegtropIMC** [comment]
@thery : I found that field can handle r^2 but not r², so I guess it is better to consistently use r^2.

**Comment by thery** [comment]
@MSoegtropIMC I think (_ ^ 2) is better but you may need to fix some proofs as there are some specific theorems for Rsqr.

**Comment by MSoegtropIMC** [comment]
Regarding x² and x^2 see #10723.

**Comment by MSoegtropIMC** [comment]
@thery: I originally had your suggested case Variant:
```
Variant asin_Spec (x : R) : R -> Type :=
| asin_p2_spec :  1 <= x -> asin_Spec x (PI / 2)
| asin_mp2_spec : x <= -1 -> asin_Spec x (- (PI / 2))
| asin_atan_spec : -1 < x < 1 -> asin_Spec x (atan (x / sqrt (1 - x ^ 2))).

Lemma asin_spec x : asin_Spec x (asin x).
Proof.
unfold asin; repeat case Rle_dec; intros; constructor; lra.
Qed.
```
but there were only two cases where this gave a very minor advantage over a simple `repeat case Rle_dec`, so I removed it during clean up. As it looks the straight forward definition:
```
Definition asin x :=
  if Rle_dec x (-1) then - (PI / 2) else
  if Rle_dec 1 x then PI / 2 else 
  atan (x / sqrt(1 - x^2)).
```
is easy to work with.

**Comment by thery** [comment]
> As it looks the straight forward definition:

Fine with me

**Comment by MSoegtropIMC** [comment]
@thery: I am thinking about how to prove the derivative of acos in an elegant way. In my old proof (for my definition) I used:
```
Definition opp_arg (f : R -> R) := fun (x : R) => f (- x).
```
and then proved
```
Lemma derivable_pt_acos_opp : forall x, -1 < x < 1 -> derivable_pt (opp_arg acos) x.
```
using the lemma for monotonously rising inverses `derivable_pt_recip_interv`. This works but requires quite a bit of (potentially useful elsewhere) boilerplate lemmas about `opp_arg`. As discussed it would probably make more sense to proof a variant of `derivable_pt_recip_interv` for falling functions.

I though now about going via:
```
Lemma acos_asin x : -1 <= x <= 1 -> acos x = PI/2 - asin x.
```
which should be trivial given the derivative for àsin`, but I don't see how I could use this in e.g.
```
derivable_pt acos x
```
I think I could define a function for `pihalfm x := PI/2 - x` and then go via `derivable_pt_comp`. Or I could use functional extensionality for the rewrite. Do you see a more elegant way of doing this?

**Comment by herbelin** [comment]
> ```
> Definition fincase3nat (f : Fin.t 3) (N0 N1 N2 : nat) :=
>  finmatch f with
>   | Fin n-2 => N0
>   | Fin n-1 => N1
>   | Fin n-0 => N2
>   end.
>  ```
> How about having this notation in the standard library?

Yes, that'll be very useful.

**Comment by MSoegtropIMC** [comment]
@thery : I pushed a preliminary version. What is missing is the derivative of acos and possibly some cleanup / moving of lemmas to different files.

**Comment by MSoegtropIMC** [comment]
@thery : just wanted to let you know that I am picking this up again and (hopefully) finish it soon. The only thing missing is the more elegant way to prove the cosine derivative.

**Comment by MSoegtropIMC** [comment]
@thery: I finally added the derivatives for the arccosine function.

I took the opportunity and tested a few ways of proving such things. In the end I decided to provide the generic lemmas for the inverse of a decreasing function. This in turn is proven by introducing a function modifier mirr_fct which mirrors a function along the x axis. There are many other ways of doing this. E.g. proving the derivative of acos using opp_fct and mirr_fct directly is only a few lines more than the provided proof, while the generic lemmas are quite lengthy. One can argue that it is not worthwhile to have the generic lemmas for decreasing functions if it is easy to use opp_fct and mirr_fct directly. The point is that the proofs are not lengthy but a bit tricky.

A note on the 3 variants of `derivable_pt_mirr`: the third variant `derivable_pt_mirr_prem` is required because this provides the form required by some lemmas using this as dependent type. Any minor modification on this makes the proofs quite tough. Of cause one can discuss if any of the other variants are required, but they are quite useful in the proofs. But I am open for discussions on this.

I am looking forward to your comments.

I guess this needs some change log entries?

File: doc/changelog/10-standard-library/00000-title.rst [hunk 1]
```diff
@@ -1,3 +1,2 @@
 
 **Standard library**
-
```
Zimmi48: This change is unexpected.

doc/changelog/10-standard-library/09803-trigo.rst
```diff
@@ -0,0 +1,2 @@
+
+- Addition to the Reals theory (`#09803 <https://github.com/coq/coq/pull/09803>`_ by Laurent Théry):
```
Zimmi48: Yes, this changelog entry should be made much more descriptive.

MSoegtropIMC:
I added a more informative chaneglog entry - I hope the splitting in two separate entries for changes and additions is what is expected - rebeased and squashed my commits into one.

MSoegtropIMC:
Since this PR is older than the "Real library maintainers group" I added the group as reviewers.

thery:
@MSoegtropIMC this looks great! I will try to have a closer look to it in a couple of days. Sorry about that.

MSoegtropIMC:
@thery: no issue - I have to apologise that it took so long.

A few things which are missing:

- integration into the reg tactic. This is a bit tricky because reg is used in various places in Ratan.v and removing this dependency is not so easy. I think one needs to make reg more modular with generic functions (like convert a (fun y:R => ...) into a fct construct) and functionality for specific functions like sin ... . The latter one can probably but in an extensible way in a hint database.

- integration with coq-interval. I simply hand no time to finalise this and didn't want to have this hanging around for any longer.

Btw.: I would enjoy some comments on the proof style. I think the proofs in the library should be more readable than they are, because people might try to understand them / take them as examples. I tried to write proofs such that the big lemma applied comes first and the rest is a fairly guided proof of premises for this lemma. Many existing proofs are the other way around with rather obscure asserts first and the big lemma at the end. But of course proof style and understandability is very subjective, so I would likt to know is you agree that the style I used is easy to understand. Of cause it is a good question how some of the harder lemmas in Ratan could be written in a more understandable way. With understandable I mean that it is fairly obvious why things are done as they done when one steps through the proof but also rough understanding of the proof concept when looking at the tactics without stepping through them.

theories/Reals/R_sqr.v
```diff
@@ -181,6 +181,38 @@ Proof.
   apply Rsqr_incr_1; assumption.
 Qed.
 
+Lemma neg_pos_Rsqr_lt : forall x y : R, - y < x -> x < y -> Rsqr x < Rsqr y.
```
thery: should be ```- y < x <  y ```
this works only for 0 < y
is this really needed, there is ```Rsqr_lt_abs_1``` and ```split_Rabs``` should be enough

theories/Reals/Ranalysis1.v
```diff
@@ -608,25 +610,104 @@ Qed.
...
+  intros x0 Hx0; specialize(H x0 Hx0).
+  rewrite <- Rabs_Ropp in H.
+  match goal with H:Rabs(?a)<eps |- Rabs(?b)<eps => replace b with a by (field; tauto) end.
+  assumption.
+Qed.
+
+Lemma derivable_pt_lim_opp_rev :
+  forall f (x l:R), derivable_pt_lim (- f) x (- l) -> derivable_pt_lim f x l.
+Proof.
+  intros f x l H.
```
thery: maybe simply:

```intros f x l H.
apply derivable_pt_lim_ext with (f := fun x => - - (f x)).
- now intros; rewrite Ropp_involutive.
- now apply derivable_pt_lim_opp_fwd.```

File: theories/Reals/Ranalysis1.v [hunk 5]
```diff
@@ -608,25 +610,104 @@ Qed.
...
+  match goal with H:Rabs(?a)<eps |- Rabs(?b)<eps => replace b with a by (field; tauto) end.
   assumption.
-  intros; rewrite H0; apply H6; assumption.
-  intro; unfold Rdiv; ring.
 Qed.
 
+Lemma derivable_pt_lim_mirr_fwd :
+  forall f (x l:R), derivable_pt_lim f (- x) (- l) -> derivable_pt_lim (mirr_fct f) x l.
+Proof.
+  intros f x l H.
```
thery: maybe
```  intros f x l H.
  change (mirr_fct f) with (comp f (- (fun x => x))).
  replace l with ((-l) * -1) by ring.
  apply derivable_pt_lim_comp; auto.
  apply derivable_pt_lim_opp.
  now apply derivable_pt_lim_id.```

File: theories/Reals/Ranalysis1.v [hunk 6]
```diff
@@ -608,25 +610,104 @@ Qed.
...
+  assumption.
+  ring_simplify(- x + - x0).
+  ring_simplify(- (x + x0)).
+  field; assumption.
+Qed.
+
+Lemma derivable_pt_lim_mirr_rev :
+  forall f (x l:R), derivable_pt_lim (mirr_fct f) (- x) (- l) -> derivable_pt_lim f x l.
+Proof.
+  intros f x l H.
```
thery:
```  intros f x l H.
  apply derivable_pt_lim_ext with (f := fun x => (mirr_fct f (- x))).
  - now intros; unfold mirr_fct; rewrite Ropp_involutive.
  - now apply derivable_pt_lim_mirr_fwd.
```

theories/Reals/Ranalysis5.v
```diff
@@ -884,6 +884,35 @@ intros f g lb ub x lb_lt_ub x_encad f_eq_g g_ok f_incr f_derivable.
  apply Left_inv ; intuition.
 Qed.
 
+Lemma derivable_pt_recip_interv_prelim1_decr :forall (f g:R->R) (lb ub x : R),
```
thery: space ```Lemma derivable_pt_recip_interv_prelim1_decr : forall```

thery: a bit puzzled by this one, there are some unnecessary hypothesis
```  intros f g lb ub x lb_lt_ub x_encad f_eq_g g_wf f_decr f_deriv.
  clear f_decr f_eq_g.
  apply f_deriv.
  apply g_wf; lra.
```

theories/Reals/R_sqr.v
MSoegtropIMC: The style is likely a bit influenced by how I do automation. I use this kind of lemma in automation in forwarding facts for lra. The forwarding should be as direct as possible and the preconditions separated so that they can be solved with `; assumption`. That's more or less the reason why this lemma is there and why it is written the way it is. But let me check how I used it here.

theories/Reals/Ranalysis1.v
MSoegtropIMC: The general idea behind having several of these lemmas is that such lemmas are sometimes used for forward and sometimes for backward reasoning. `derivable_pt_lim_opp_rev` is intended for backward reasoning and thus should have a goal which matches an as wide class of goals as possible, so it should not include -l or -f but l and f. Then there is a variant for forward reasoning which should have a generic hypothesis without -l and -f. And then (sorry for that) I needed a third variant for derive - it gets quite messy if one uses one of the two other forms.

Of cause one can live with only one lemma, but it generally means at least a bit of "intelligent" massaging of goals or hypothesis, which is hard to automate. Having specialized lemmas for different purposes usually makes automation much faster, because they either fit immediately or not at all.

But I fully agree that this is way inflationary. I mostly put it in this way to have a discussion on the topic. Btw.: this applies much more to lemmas like
```
Lemma derive_pt_plus :
  forall f1 f2 (x:R) (pr1:derivable_pt f1 x) (pr2:derivable_pt f2 x),
    derive_pt (f1 + f2) x (derivable_pt_plus _ _ _ pr1 pr2) =
    derive_pt f1 x pr1 + derive_pt f2 x pr2.
```
This is not very nice for automation because the derivable proof has to have the given specific form. For blind rewriting of derive_pt statements of function addition it would be more useful to make the derivable proof an implicit argument of the lemma. I mean if you have derive_pt (f + g) in a term, you must already have a proof that the derivative exists (ok it might be an evar, but usually it is not). Why should I care how it looks? If there are restrictions on this somewhere, it is much easier to sort these out after rewriting then to massage the goal accordnigly before rewriting.

On the other hand if you want to forward this, so that you don't have a proof of derivability of the sum, you of cause want this lemma in the form it is stated.

Comming to the point: I think in general we should have forward and backward versions of most of the derivable lemmas. As is working with this is a bit hard (without reg). And I think that the the inverse of atan is not in reg is more or less because it is a bit hard to prove it without reg, so there is a hen and egg problem with reg (or one needs an extensible modular version of reg). But introducing several variants of these lemmas would make it much easier to use them without reg.

What is you opinion on this?

theories/Reals/Ranalysis5.v
MSoegtropIMC: Indeed, I took the statement from the existing increasing variant `derivable_pt_recip_interv_prelim1` - as it looks not a smart idea to do the proof based on this. Should I simplify both (the increasing and decreasing variant) or do you think the increasing variant might be used outside of this file?

theories/Reals/Ranalysis1.v
MSoegtropIMC: @thery: can you give me a short feedback on how we shall continue here? As I said I am completely open to remove the slightly inflationary variants of lemmas - the only one really required is the one which matches the structure used in derive terms - but as I wrote I think there are good reasons for having separate lemmas for forward reasoning, backward reasoning and dependent type applications and I would like to drive the standard library more into this direction.

Do you think it would help the discussion if I modify some existing proofs to the style I propose? I think I gave a few examples of what I have in mind but maybe it becomes more obvious if I modify a more complex existing proof.

thery: Fine with me to have several variants my initial point was just that their proof should reflect that they are variant (i.e. be very short).

theories/Reals/Ranalysis5.v
thery: thinking about that I was wondering if instead of incr and decr we should have monotonic,
is the argument very different between incr and decr

MSoegtropIMC: Yes I thought about it but I think one has to change the interface quite a bit then because several of the premises depend on the incr/decr question. I think it is better to do this as a separate step.

theories/Reals/Ranalysis1.v
MSoegtropIMC: OK, a missunderstanding then. I will have a look at the proofs.

theories/Reals/Ranalysis5.v
thery: ok

theories/Reals/R_sqr.v
MSoegtropIMC: I reviewed this. The reason I did it this way is for symmetry with neg_pos_Rsqr_le, which already exists in the standard library. The variant with -y<x<y exists as Rsqr_bounds_lt. I added Rsqr_bounds_le and Rsqr_bounds_lt. These are the variants I actually use. So from my point of view one could remove neg_pos_Rsqr_lt, but I think for symmetry I would prefer to do it like this, and as I said for automation sometimes the split comparison variants are easier.

theories/Reals/Ranalysis1.v
MSoegtropIMC: Indeed derivable_pt_lim_ext is very useful here. I used your proof.

theories/Reals/Ranalysis1.v
MSoegtropIMC: Thanks, much more elegant. I had to move derivable_pt_lim_comp and derivable_pt_lim_id up a bit for this and did this consistently for the other variants. I also added a few subsection headlines.

theories/Reals/Ranalysis1.v
MSoegtropIMC: Thanks, I took your proof.

theories/Reals/Ranalysis5.v
MSoegtropIMC: Fixed

MSoegtropIMC: I removed the extra premises in the decr and also the original incr variant and adjusted the proves which use these lemmas.

MSoegtropIMC: P.S.: Interestingly thins changed what gets shelved by eapply in some cases.

MSoegtropIMC:
@thery: sorry that it took so long. I implemented your review comments. Thanks - I learned quite a bit. I didn't squash this as yet because the change of lemma order required by one of your suggestions leads to quite a few shift changes. I added a few subsection headers. Please especially review these.

thery:
thanks I will try to look at it omorrow

MSoegtropIMC:
@thery: I will rebase cause of the CI failures. I changed the two lemmas with superfluous premises, but I can't imagine someone outside is using these. Maybe one should make such lemmas local?

thery:
@MSoegtropIMC  this looks goot to me. I am not sure if @silene is around but if so he could approve his change request so we can (finally) merge... :relieved:

MSoegtropIMC:
@thery : did you hear from @silene? I would say if we don't hear back from him until Monday as an exception I approve your commit and you approve my commit. I want to continue with this (e.g. make acos and asin available in CoqInterval).

thery:
@MSoegtropIMC  perfect let's wait and see

MSoegtropIMC:
@thery : how shall we proceed? I implemented the name changes requested by @silene quite some time back, but I could double check.

thery:
apparently @silene is not very active lately (which is understandable).  If you think it is ok (I think it is) maybe @ejgallego  or @Zimmi48 can merge it.

Zimmi48:
@herbelin has also joined the @coq/reals-library-maintainers team a while ago, so I suppose he could take care of dismissing @silene's request for change and approving and merging the PR.

MSoegtropIMC:
@ejgallego , @Zimmi48 : a bit of background info: this PR is originally by @thery. I was working on the same thing at the same time and eventually we put our two developments together. So this PR consists of the original commit by thery and changes/extensions from me. I reviewed thery's commit and thery reviewed my commit.

Another solution would be that I first merge thery's commit and thery immediately after merges my commit - that is I remove my commit from this PR and do a separate PR.

The reason we didn't do this right away is that we agreed to changed the definition of the asin and acos function and we didn't want that people start to work with the old definition, e.g. integrate it into CoqInterval. But if we do both merges shortly after another, there is no such problem. Just months in between or even a release in between would have been a problem.

Zimmi48:
@MSoegtropIMC I know, I've followed (from afar) this PR history. Have you seen my comment above though? It looks like you didn't.

theories/Reals/Rtrigo_ident.v
```diff
@@ -0,0 +1,317 @@
...
+    all: pose proof Rsqr_pos_lt (cos α) ltac:(lra).
+    all: pose proof sqrt_lt_R0 (cos α)² ltac:(assumption).
+    all: lra.
+Qed.
+
+(*********************************************************)
+(** * Additional shift lemmas for sin, cos, tan          *)
+(*********************************************************)
+
+Lemma sin_pi_minus_x:
```
herbelin: I would a priori avoid using a variable name in the lemma name. The name `sin_pi_minus` seems as good.

Same remark for other statements below.

theories/Reals/Rtrigo_ident.v
```diff
@@ -0,0 +1,317 @@
+(************************************************************************)
```
herbelin: Used to program in Coq, the suffix `_ident` tends to mean `identifier` for me. Here, it is obviously supposed to be `identities`, which is also quite long to type. I'm ok with `_ident` but otherwise, maybe `_laws` or `_facts` might be less ambiguous.

herbelin:
> apparently @silene is not very active lately (which is understandable). If you think it is ok (I think it is) maybe @ejgallego or @Zimmi48 can merge it.

@silene told me that for urgent matters, his direct e-mail address should be used. He does not have the time to regularly look at his coq github notifications.

> @herbelin has also joined the @coq/reals-library-maintainers team a while ago, so I suppose he could take care of dismissing @silene's request for change and approving and merging the PR.

I can assign (and a quick review is coming).

herbelin (commented):
This is a very interesting PR with a lot of contents, thanks a lot. I made a quick review on the form and made a few comments.

thery:
@herbelin  thanks, @MSoegtropIMC  can you take care
 of making the changes suggested by Hugo?

theories/Reals/Rtrigo_ident.v
```diff
@@ -0,0 +1,317 @@
...
+Require Import Lra.
+Require Import Ranalysis_reg.
+
+Local Open Scope R_scope.
+
+(*********************************************************)
+(** * Preliminaries                                      *)
+(*********************************************************)
+
+(** ** Various generic lemmas which probably should go somewhere else *)
```
herbelin: The three lemmas below should go in `RIneq.v` I believe (which for long is not only about inequalities).

MSoegtropIMC: Thanks for the review! I will fix all the stuff - sorry for the non OCaml'isch syntax sometimes - I frequently switch between Coq and Maxima.

Regarding these lemmas: I have a larger set of such lemmas which are intended to manipulate equations as I learned it in school. Say add, multiply, subtract, divide both sides of an equation with something, take the log or asin of both sides of an equation, ... sometimes this is convenient and a natural way to manipulate equations - together with field and field_simplify. A good part of this can be done with f_equal and lambda expressions, but since a lot of this requires side conditions for the domain, this is usually less obvious (especially when reading a proof) than applying especially tailored lemmas. Would it be agreeable to have this in the standard library? And if so, should this go into a separate module or is RIneq.v the right place?

theories/Reals/Rtrigo_ident.v
```diff
@@ -0,0 +1,317 @@
...
+    all: pose proof sqrt_lt_R0 (cos α)² ltac:(assumption).
+    all: lra.
+Qed.
+
+(*********************************************************)
+(** * Additional shift lemmas for sin, cos, tan          *)
+(*********************************************************)
+
+Lemma sin_pi_minus_x:
+  forall x:R, sin(PI - x) = sin(x).
```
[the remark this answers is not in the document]
MSoegtropIMC: I agree, not very consistent. Would you prefer `α` or `x` throughout the module?

theories/Reals/Rtrigo_ident.v
MSoegtropIMC: I would prefer `_facts` then. Btw.: would you find it agreeable to collect *all* trigonometry lemmas here (as a separate PR)? Right now they are spread all over the place. The reason for this is that a lot of stuff is required earlier to prove other lemmas, like derivatives and Taylor series. But I think one could restate the lemmas here and search blacklist the original, so that one has a sort of trigonometry formulary in one place. Search is of cause your friend, but only if one expects that certain lemmas are there and for more advanced stuff there are many ways to state them and Search doesn't work that well.

herbelin:
> I would enjoy some comments on the proof style. 
> 
> I think the proofs in the library should be more readable than they are,

My view is that whatever we do, the proofs are difficult to read. My dream for readibility would be to have a UI/coqdoc tool which turns a proof script into a "human-readable" proofs, with explicitation of what is proved at each step, and with levels of details to tackle the complexity.

> because people might try to understand them / take them as examples.

Providing examples of proof style is indeed a common request. On my side, I renounced to the idea that a strong style can be imposed (on the contrary e.g. of MathComp which "succeeded" to have a rather consistent way of writing proofs). 

My observation is that even the style of a single individual evolves in time, with differences between a newcomer (who will tend to use simple tactics) and an intensive user (who will tend to use time-saving "key bindings" and to write compact proofs). Important points for me though are:
- structurate the proofs with bullets
- avoid using generated names on which there is no control (this is too fragile)

In particular, I'm not shocked by the proofs in this PR.

Reading them again, I may recommend a few formatting conventions though:
- consistently keep an initial two-space indentations in proofs, after the initial `Proof`
- add a space in `assert(foo)` and change things like `field_simplify( foo)` into `field_simplify (foo)`

> I tried to write proofs such that the big lemma applied comes first and the rest is a fairly guided proof of premises for this lemma. Many existing proofs are the other way around with rather obscure asserts first and the big lemma at the end.

I would classify this as subjective, like writing `let y = x+1 in let z = y * 2 in z ^ 3` vs `((x+1)*2)^3`. But maybe some people have strong arguments to discriminate between these two style.

theories/Reals/R_sqr.v
herbelin: > The style is likely a bit influenced by how I do automation.

Somehow, I would tend to prefer that the automation tools adapt to the style, or, conversely, that unity of style is not sacrified for technical reasons which are a priori unrelated, but we also have to be pragmatic and to sacrify automatisability would not be good either.

theories/Reals/Rtrigo_ident.v
herbelin: > would you find it agreeable to collect all trigonometry lemmas here (as a separate PR)?

From my point of view, certainly, if this seems good to @thery too.

theories/Reals/Rtrigo_ident.v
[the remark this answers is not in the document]
herbelin: If we had to choose, the whole library uses mostly `x` (5000 occurrences), then `y` (1900 occurrences), then `z`, and sometimes `r` (1000 occurrences + 900 occurrences `r1` + 800 occurrences `r2`), so I guess `x` should be the canonical choice (even if `α` has some charm...).

theories/Reals/Rtrigo_ident.v
herbelin: > Would it be agreeable to have this in the standard library? 

I would say yes from my side.

> And if so, should this go into a separate module or is RIneq.v the right place?

I would try `RIneq.v` in a first step (even if the conclusion is that it should eventually be structurated differently).

MSoegtropIMC:
> I would classify this as subjective, like writing `let y = x+1 in let z = y * 2 in z ^ 3` vs `((x+1)*2)^3`. But maybe some people have strong arguments to discriminate between these two style.

In my opinion it is a question of flow of intention. In your example in doesn't make much of a difference if you go inside out or outside in, but this is not always the case. Let's take as example `Ratan.derivable_pt_atan` and `Ratan.derivable_pt_asin` which prove a similar goal using the same main lemma in two different styles.

`Ratan.derivable_pt_asin` *starts* with:
```
  eapply (derivable_pt_recip_interv sin asin (-PI/2) (PI/2)); [shelve ..|].
```
then continues with the most complicated premise, that the derivative of f (g x) is not 0, and then proves all the other, rather trivial, premises of derivable_pt_recip_interv. I think this is close to how one would do a pencil and paper proof, say "We prove this using derivable_pt_recip_interv. For this we have to show ...".

`Ratan.derivable_pt_atan` *ends* with
```
apply (derivable_pt_recip_interv tan atan (-ub) ub x
      lb_lt_ub xint inv_p int_tan incr der).
exact df_neq.
```
All premises are proven upfront as assert (not sure why the last premise df_neq is not given as argument as well). I find this hard to read because one does not know why all these things are asserted before one comes to the end.

The advantage and at the same time disadvantage of the "assert premises upfront" approach is that one actually states the things to prove. This is more explicit but on the other hand also makes the proof much longer, because it states a lot of things eapply can tell us automatically.

The main advantage of the "eapply main lemma upfront" approach I see is that it states at the beginning the proof sketch - use `derivable_pt_recip_interv` and prove whatever this needs as premise.

A disadvantage is that the eapply approach is a bit fragile. Minor changes in the context can influence the order in which things are staged.

Another advantage of the "eapply" approach is that it makes it easer to automate things - in which case the fragility argument doesn't apply any more.

And another advantage of the "assert" approach is that it is a bit easier / more natural to reuse intermediate results required for several premises.

I just wonder if I am used to such a proof style cause I tend to automate things so that I just give a rough proof sketch and have machinery figure out the details.

herbelin:
I can only share your analysis!

MSoegtropIMC:
@herbelin: I did quite a bit of editorial cleanup as you suggested and pushed this as separate commit. One point: I experimented a bit and I am not sure which is more idiomatic / readable for lemmas:
```
Lemma sin_atan x : sin (atan x) = x / sqrt (1+x²).
```
or
```
Lemma sin_atan, forall x :
  sin (atan x) = x / sqrt (1+x²).
```
I think at least in case there are range premises, as in:
```
Lemma acos_atan : forall x, 0 < x ->
  acos x = atan (sqrt (1-x²) / x).
```
the forall is more readable, and after thinking about it I think for lemmas it is generally more readable, since it is closer to how I read a lemma loud. `sin_atan x` I would more read sin_aton *of* x.

I use lemmas frequently as functions. I even think that one weakness of the SF tutorial is that they don't use this as much as they could, e.g. control rewrite by giving arguments to the rewrite lemma rather than locating replacing points with at. Still when reading lemmas I prefer the forall. In the commit I did this pretty systematically. Just let me know if you like it this way. If not, I can also change it such that at least simple lemmas have no forall.

Btw.: Is this technically (from a lemma usage point of view) 100% the same?

theories/Reals/Rtrigo_ident.v
MSoegtropIMC: OK. I think a bit about if for now I just add the 3 lemmas required or more (depends on how clean the lemmas I have are).

herbelin:
> Still when reading lemmas I prefer the forall. In the commit I did this pretty systematically. Just let me know if you like it this way. If not, I can also change it such that at least simple lemmas have no forall.

Both style have their arguments, and, as far as I'm concerned, the "forall" style is fine.

> Btw.: Is this technically (from a lemma usage point of view) 100% the same?

From the lemma usage point of view, yes (as far as I know).

Otherwise, I lost a bit track of what has been changed, so tell me you're done with the changes.

MSoegtropIMC:
> Otherwise, I lost a bit track of what has been changed, so tell me you're done with the changes.

The moving of the 3 equation manipulation lemmas is still missing. I didn't have time to review my library. But I think I will just put in these 3 and do it as a separate PR - it is quite unrelated.

herbelin:
> But I think I will just put in these 3 and do it as a separate PR - it is quite unrelated.

Yes, let's do like this.

MSoegtropIMC:
@herbelin: I fixed the issue with the 3 equation rewrite lemmas by replacing their use with something else. The involved proofs get only very slightly longer. I think this is a better solution than having half done stuff in the library. I also rebased and squashed my main and my cleanup commits. So from my point of view this is ready to merge.

herbelin (approved):
> So from my point of view this is ready to merge.

OK, let's do.

PR #9804, opened 2019-03-19
CI: Use job-local timeout for build-template and test-suite-template

SkySkimmer:


ejgallego (commented):
Umm, I will think about this but I am not a fan at all of this hack.

I am not willing to support unrealiable jobs in the CI, sorry.

ejgallego (changes requested):
I am not OK with this; core jobs such as `build` and `test-suite` should never timeout; I am not willing to add more complexity to the CI setup in order to accommodate non-acceptable behavior.

Zimmi48:
I believe that this is useful nonetheless (even if we kill these unreliable jobs). Most of our jobs run quickly and it is useful to detect early when they are taking too much time. It happened to see PRs that were creating bugs that made the system loop. There is no point in letting these jobs run for four hours. The issue is that our default timeout should be much lower, and we should have exceptions for Windows jobs and a few libraries but GitLab doesn't support this, so we have a long timeout for every job.

ejgallego:
> I believe that this is useful nonetheless (even if we kill these unreliable jobs). Most of our jobs run quickly and it is useful to detect early when they are taking too much time. It happened to see PRs that were creating bugs that made the system loop. There is no point in letting these jobs run for four hours. The issue is that our default timeout should be much lower, and we should have exceptions for Windows jobs and a few libraries but GitLab doesn't support this, so we have a long timeout for every job.

I am having trouble to remember such looping cases; IMHO it was an extremely rare occurrence before the async jobs were added.

Indeed, calculating a good timeout is very challenging; some jobs may need 5 minutes some may need 2 hours [such as mathcomp]. This PR adds a) a new variable and b) a new dependency on a tool, so IMO it increases the already complex CI setup for no good reason.

I do agree that once Gitlab supports a timeout per job we should fine tune the stuff. In fact we should likely lower the current default to 2h, and increase it only when the full master build is being done.

Zimmi48:
Sure, this was quite rare. And I didn't look at the proposed implementation, so I can't really judge if it's hackish.

gares (approved):
It looks very clean and declarative to me

SkySkimmer:
TBH it's not as clean as I'd like, because we have the scripts inline in 
the yml so the timeout only applies to some individual commands.



Gaëtan Gilbert

gares:
I frankly don't get why this PR is blocked. It is true we have jobs that loop/hang, and they shall be fixed.
This PR makes things fail faster, improving our CI while we wait for a proper fix for the jobs above.

I don't see while making CI better means allowing jobs that fail, it just makes things better while we look for a fix. And the complexity of the fix is ridiculous, it is just adding the word `timeout` in front of `make`.

ejgallego:
I have explained my reasons; a) timeout is grave offense due to resource use, jobs in base are not supposed to do so, thus we are not adding a workaround for a thing that should not happen in the first place; b) the implementation is ad-hoc and doesn't transfer well to the complete job, it adds non-trivial complexity and dependencies for example about process use, etc...

This is a NACK from me; the right fix is to fuse the two quick jobs in `test`, as suggested in #9801

ejgallego:
From my point of view this PR can be closed, WDYT @SkySkimmer ?

SkySkimmer:
>This is a NACK from me; the right fix is to fuse the two quick jobs in test

That does nothing for timeouts

>thus we are not adding a workaround for a thing that should not happen in the first place

We put a workaround because we don't want it to happen, I don't understand your objection

>the implementation is ad-hoc and doesn't transfer well to the complete job, it adds non-trivial complexity and dependencies for example about process use, etc...

It's adhoc but not complex and I have no clue what dependencies you're talking about.

SkySkimmer:
@ejgallego as far as I can tell you just want the jobs gone, but until and unless that happens please let use reduce the damage from them breaking.

ejgallego:
> It's adhoc but not complex and I have no clue what dependencies you're talking about.

Now you have to reason about what timeout does for processes, portability, etc... it is yet another wrapper. If we would need it OK, but this is a GNU only dependency.

> That does nothing for timeouts

Put the timeout on that particular job so the core ones are not affected.

> We put a workaround because we don't want it to happen, I don't understand your objection

It is the other way around, jobs in base are not allowed to timeout; by adding this command, we are quite sending the message "oh, now your can add broken stuff to the base CI and we are OK with it"

I am not OK with it. If a job in base is not stable enough then it should not be on base.

ejgallego:
> @ejgallego as far as I can tell you just want the jobs gone, but until and unless that happens please let use reduce the damage from them breaking.

I don't have a particular opinion on the jobs, but certainly I am very set in that CI is not the place to put broken stuff; sorry.

At least not in base, move the job to `test` as I mentioned and we can keep them for a while.

Zimmi48:
I'm fine with what @ejgallego proposes (the move to test, and merging of the two jobs) but that doesn't go against merging this PR. In particular, and generally speaking, this PR is going to make us reach the GitLab timeout less often, which is good, and which does not provide any reason why the faulty jobs shouldn't be removed nonetheless.

ejgallego:
> In particular, and generally speaking, this PR is going to make us reach the GitLab timeout less often, which is good, and which does not provide any reason why the faulty jobs shouldn't be removed nonetheless.

I wish that was true, and like 99% is what would happen, but we have no idea how this untested timeout command is going to play out in general, and it affects _all the jobs_, so I dunno, I don't want to have to spend time figuring out some problem we introduced here.

IMHO what I say is easy to do [and reduces in fact useless bootstrap time as there is only 1 jobs that depends on quick].

ejgallego:
The thing is that this is trickier than it looks; make spawns many processes, coq spawns even more; timeout sends a signal. Not trivial at all.

ejgallego:
> I don't have a particular opinion on the jobs, but certainly I am very set in that CI is not the place to put broken stuff; sorry.

To be clear, my personal opinion regarding the unreliable jobs is that I would remove them now; however I do fully respect the wish of the maintainers to keep them for now so from my part they will remain in the CI until re-evaluation is wanted [August?]

Zimmi48:
August seems a bit far away to continue wasting resources like that.

ejgallego:
> August seems a bit far away to continue wasting resources like that.

I dunno, up to the async maintainers to set a date; they claim the jobs are useful to try to debug. Also, I have many changes in the queue wrt "terminators" that could help; I don't think so but you never know.

maximedenes:
I don't understand why we don't merge this as a temporary workaround, until we can investigate the real issues, which we all would like to see fixed.

ejgallego:
> I don't understand why we don't merge this as a temporary workaround, until we can investigate the real issues, which we all would like to see fixed.

As I stated above because I deem this change risky to regular processes and I don't want to spend time debugging things that work today.

gares:
CI is green, what do you need to debug?

SkySkimmer:
>CI is green, what do you need to debug?

We haven't seen what happens when the timeout triggers though.

ejgallego:
I am sorry but I am very set on this; already we are making a huge exception to the policy allowing jobs that fail often to remain in the CI; on top of that we are not going to implement workarounds for those broken jobs that affect the regular flow of the sane ones.

Confine the fix to the particular badly-behaving job and things will be OK then.

SkySkimmer:
Changed so that the timeout is `0` = none by default, `100m` for the async/quick jobs.

ejgallego:
> Changed so that the timeout is `0` = none by default, `100m` for the async/quick jobs.

Sorry if I was not clear on this; I'd prefer that the regular jobs don't get wrapped on timeout [which for example may modify signal masks]

gares:
> which for example may modify signal masks

What does this mean?

ejgallego:
> What does this mean?

My point precisely is that I want to avoid having to understand what the `timeout` command does, so we put the fix in the job that can timeout, and we leave the rest of the jobs as before.

ejgallego (approved):
Sorry I was not clear about what I meant: I was thinking of duplicating the script in the unreliable jobs and letting the template untouched.

I wouldn't like to support something that is not allowed [that is base jobs going out of time], as it may be tempting for people to abuse it.

Anyways, this is good enough for now, thanks, we'll clean up later.

ejgallego:
@SkySkimmer you may want to squash the commits tho?

SkySkimmer:
squ	ashed

Gaëtan Gilbert

ejgallego:
Uh there was this failure here https://gitlab.com/coq/coq/-/jobs/189251729:
```
COQC issues/issue143.v
File "./issues/issue129.v", line 9, characters 47-81:
Error: Timeout!
```
surely not related but re-rerunning.

PR #9808, opened 2019-03-20
Preliminary refactoring for `.vos` files support

maximedenes:
This is meant to make #8642 easier, in particular avoiding the introduction of a global flag governing the semantics of `Require`.

The approach we take is to make explicit the link between the compilation target and the semantics of `Require`.

We do introduce one semantic change: when producing a `.vo` file, we no longer look for `.vio` files. @gares can you confirm this is the desired behavior?

### DIFF
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -306,15 +306,6 @@ end (* }}} *)
 (*************************** THE DOCUMENT *************************************)
 (******************************************************************************)
 
-type interactive_top = TopLogical of DirPath.t | TopPhysical of string
-
-type compilation_mode = BuildVo | BuildVio | Vio2Vo
-
-(* The main document type associated to a VCS *)
-type stm_doc_type =
-  | Batch of compilation_mode * string
-  | Interactive of interactive_top
-
 (* Dummy until we land the functional interp patch + fixed start_library *)
 type doc = int
 let dummy_doc : doc = 0
@@ -335,10 +326,7 @@ module VCS : sig
 
   type vcs = (branch_type, transaction, vcs state_info, box) Vcs_.t
 
-  val init : stm_doc_type -> id -> Vernacstate.Parser.state -> doc
-  (* val get_type : unit -> stm_doc_type *)
-  val set_ldir : Names.DirPath.t -> unit
-  val get_ldir : unit -> Names.DirPath.t
+  val init : id -> Vernacstate.Parser.state -> doc
 
   val is_interactive : unit -> bool
   val is_vio_doc : unit -> bool
@@ -527,30 +515,22 @@ end = struct (* {{{ *)
   type vcs = (branch_type, transaction, vcs state_info, box) t
   let vcs : vcs ref = ref (empty Stateid.dummy)
 
-  let doc_type = ref (Interactive (TopLogical (Names.DirPath.make [])))
-  let ldir = ref Names.DirPath.empty
-
-  let init dt id ps =
-    doc_type := dt;
+  let init id ps =
     vcs := empty id;
     let info = { (default_info None) with state = ParsingState ps } in
     vcs := set_info !vcs id info;
     dummy_doc
 
-  let set_ldir ld =
-    ldir := ld
-
-  let get_ldir () = !ldir
-  (* let get_type () = !doc_type *)
-
   let is_interactive () =
-    match !doc_type with
-    | Interactive _ -> true
+    let open Declaremods in
+    match (get_current_library_info ()).library_mode with
+    | Interactive -> true
     | _ -> false
 
   let is_vio_doc () =
-    match !doc_type with
-    | Batch (BuildVio, _) -> true
+    let open Declaremods in
+    match (get_current_library_info ()).library_mode with
+    | Batch BuildVio -> true
     | _ -> false
 
   let current_branch () = current_branch !vcs
@@ -2570,12 +2550,16 @@ end (* }}} *)
 (********************************* STM API ************************************)
 (******************************************************************************)
 
-(* Main initialization routine *)
+type top_path = TopLogical of DirPath.t | TopPhysical of string
+
+(* Main initalization routine *)
 type stm_init_options = {
   (* The STM will set some internal flags differently depending on the
-     specified [doc_type]. This distinction should disappear at some
-     some point. *)
-  doc_type     : stm_doc_type;
+     specified [library_mode]. This distinction should disappear at some
+     point. *)
+  library_mode : Declaremods.library_mode;
+
+  top_path : top_path;
 
   (* Initial load path in scope for the document. Usually extracted
      from -R options / _CoqProject *)
@@ -2593,13 +2577,6 @@ type stm_init_options = {
 }
 (* fb_handler   : Feedback.feedback -> unit; *)
 
-(*
-let doc_type_module_name (std : stm_doc_type) =
-  match std with
-  | VoDoc mn | VioDoc mn | Vio2Vo mn -> mn
-  | Interactive mn -> Names.DirPath.to_string mn
-*)
-
 let init_core () =
   if !cur_opt.async_proofs_mode = APon then Control.enable_thread_delay := true;
   State.register_root_state ()
@@ -2616,7 +2593,7 @@ let dirpath_of_file f =
   let ldir = Libnames.add_dirpath_suffix ldir0 id in
   ldir
 
-let new_doc { doc_type ; iload_path; require_libs; stm_options } =
+let new_doc { library_mode ; top_path; iload_path; require_libs; stm_options } =
 
   let require_file (dir, from, exp) =
     let mp = Libnames.qualid_of_string dir in
@@ -2629,7 +2606,7 @@ let new_doc { doc_type ; iload_path; require_libs; stm_options } =
   (* We must reset the whole state before creating a document! *)
   State.restore_root_state ();
 
-  let doc = VCS.init doc_type Stateid.initial (Vernacstate.Parser.init ()) in
+  let doc = VCS.init Stateid.initial (Vernacstate.Parser.init ()) in
 
   (* Set load path; important, this has to happen before we declare
      the library below as [Declaremods/Library] will infer the module
@@ -2638,20 +2615,28 @@ let new_doc { doc_type ; iload_path; require_libs; stm_options } =
 
   Safe_typing.allow_delayed_constants := !cur_opt.async_proofs_mode <> APoff;
 
-  begin match doc_type with
-    | Interactive ln ->
-      let dp = match ln with
-        | TopLogical dp -> dp
-        | TopPhysical f -> dirpath_of_file f
-      in
-      Declaremods.start_library dp
+  let library_path = match top_path with
+    | TopLogical dir -> dir
+    | TopPhysical f -> dirpath_of_file f
+  in
 
-    | Batch (_, f) ->
-      let ldir = dirpath_of_file f in
-      let () = Flags.verbosely Declaremods.start_library ldir in
-      VCS.set_ldir ldir;
+  let library_info =
+    Declaremods.{ library_mode;
+                  library_path;
+                }
+  in
+
+  begin match library_mode with
+  | Declaremods.Interactive ->
+    Declaremods.start_library library_info
+  | Declaremods.Batch _ ->
+    Flags.verbosely Declaremods.start_library library_info
+  end;
+
+  begin match library_mode, top_path with
+    | Declaremods.Batch _, TopPhysical f ->
       set_compilation_hints f
-
+    | _, _ -> ()
   end;
 
   (* Import initial libraries. *)
@@ -3199,7 +3184,6 @@ let edit_at ~doc id =
         Exninfo.iraise (e, info)
 
 let get_current_state ~doc = VCS.cur_tip ()
-let get_ldir ~doc = VCS.get_ldir ()
 
 let get_doc did = dummy_doc
 
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -308,10 +308,11 @@ end (* }}} *)
 
 type interactive_top = TopLogical of DirPath.t | TopPhysical of string
 
+type compilation_mode = BuildVo | BuildVio | Vio2Vo
+
 (* The main document type associated to a VCS *)
 type stm_doc_type =
-  | VoDoc       of string
-  | VioDoc      of string
+  | Batch of compilation_mode * string
   | Interactive of interactive_top
 
 (* Dummy until we land the functional interp patch + fixed start_library *)
@@ -549,7 +550,7 @@ end = struct (* {{{ *)
 
   let is_vio_doc () =
     match !doc_type with
-    | VioDoc _ -> true
+    | Batch (BuildVio, _) -> true
     | _ -> false
 
   let current_branch () = current_branch !vcs
@@ -2645,17 +2646,12 @@ let new_doc { doc_type ; iload_path; require_libs; stm_options } =
       in
       Declaremods.start_library dp
 
-    | VoDoc f ->
+    | Batch (_, f) ->
       let ldir = dirpath_of_file f in
       let () = Flags.verbosely Declaremods.start_library ldir in
       VCS.set_ldir ldir;
       set_compilation_hints f
 
-    | VioDoc f ->
-      let ldir = dirpath_of_file f in
-      let () = Flags.verbosely Declaremods.start_library ldir in
-      VCS.set_ldir ldir;
-      set_compilation_hints f
   end;
 
   (* Import initial libraries. *)
--- a/stm/stm.mli
+++ b/stm/stm.mli
@@ -44,12 +44,13 @@ end
 
 type interactive_top = TopLogical of DirPath.t | TopPhysical of string
 
+type compilation_mode = BuildVo | BuildVio | Vio2Vo
+
 (** The STM document type [stm_doc_type] determines some properties
    such as what uncompleted proofs are allowed and what gets recorded
    to aux files. *)
 type stm_doc_type =
-  | VoDoc       of string       (* file path *)
-  | VioDoc      of string       (* file path *)
+  | Batch of compilation_mode * string (* file path *)
   | Interactive of interactive_top    (* module path *)
 
 (** Coq initialization options:
--- a/stm/stm.mli
+++ b/stm/stm.mli
@@ -42,20 +42,9 @@ module AsyncOpts : sig
 
 end
 
-type interactive_top = TopLogical of DirPath.t | TopPhysical of string
+(** Coq initalization options:
 
-type compilation_mode = BuildVo | BuildVio | Vio2Vo
-
-(** The STM document type [stm_doc_type] determines some properties
-   such as what uncompleted proofs are allowed and what gets recorded
-   to aux files. *)
-type stm_doc_type =
-  | Batch of compilation_mode * string (* file path *)
-  | Interactive of interactive_top    (* module path *)
-
-(** Coq initialization options:
-
- - [doc_type]: Type of document being created.
+ - [library_mode]: Type of document being created.
 
  - [require_libs]: list of libraries/modules to be pre-loaded at
    startup. A tuple [(modname,modfrom,import)] is equivalent to [From
@@ -64,11 +53,15 @@ type stm_doc_type =
    the module, [Some true] will additionally export it.
 
 *)
+type top_path = TopLogical of DirPath.t | TopPhysical of string
+
 type stm_init_options = {
   (* The STM will set some internal flags differently depending on the
-     specified [doc_type]. This distinction should disappear at some
-     some point. *)
-  doc_type     : stm_doc_type;
+     specified [library_info]. This distinction should disappear at some
+     point. *)
+  library_mode     : Declaremods.library_mode;
+
+  top_path : top_path;
 
   (* Initial load path in scope for the document. Usually extracted
      from -R options / _CoqProject *)
@@ -177,7 +170,6 @@ val finish_tasks : string ->
 
 (* Id of the tip of the current branch *)
 val get_current_state : doc:doc -> Stateid.t
-val get_ldir : doc:doc -> Names.DirPath.t
 
 (* This returns the node at that position *)
 val get_ast : doc:doc -> Stateid.t -> Vernacexpr.vernac_control option
--- a/toplevel/ccompile.ml
+++ b/toplevel/ccompile.ml
@@ -107,18 +107,19 @@ let compile opts copts ~echo ~f_in ~f_out =
     | NativeOff -> false | NativeOn {ondemand} -> not ondemand
   in
   match copts.compilation_mode with
-  | Stm.BuildVo ->
+  | Declaremods.BuildVo ->
       let long_f_dot_v, long_f_dot_vo =
         ensure_exists_with_prefix f_in f_out ".v" ".vo" in
 
       let doc, sid = Topfmt.(in_phase ~phase:LoadingPrelude)
           Stm.new_doc
-          Stm.{ doc_type = Batch(BuildVo, long_f_dot_vo);
+          Stm.{ library_mode = Declaremods.(Batch BuildVo);
+                top_path = TopPhysical long_f_dot_vo;
                 iload_path; require_libs; stm_options;
               } in
       let state = { doc; sid; proof = None; time = opts.config.time } in
       let state = load_init_vernaculars opts ~state in
-      let ldir = Stm.get_ldir ~doc:state.doc in
+      let ldir = Declaremods.((get_current_library_info ()).library_path) in
       Aux_file.(start_aux_file
         ~aux_file:(aux_file_name_for long_f_dot_vo)
         ~v_file:long_f_dot_v);
@@ -139,7 +140,7 @@ let compile opts copts ~echo ~f_in ~f_out =
       Aux_file.stop_aux_file ();
       Dumpglob.end_dump_glob ()
 
-  | Stm.BuildVio ->
+  | Declaremods.BuildVio ->
       let long_f_dot_v, long_f_dot_vio =
         ensure_exists_with_prefix f_in f_out ".v" ".vio" in
 
@@ -158,20 +159,21 @@ let compile opts copts ~echo ~f_in ~f_out =
 
       let doc, sid = Topfmt.(in_phase ~phase:LoadingPrelude)
           Stm.new_doc
-          Stm.{ doc_type = Batch(BuildVio, long_f_dot_vio);
+          Stm.{ library_mode = Declaremods.(Batch BuildVio);
+                top_path = TopPhysical long_f_dot_vio;
                 iload_path; require_libs; stm_options;
               } in
 
       let state = { doc; sid; proof = None; time = opts.config.time } in
       let state = load_init_vernaculars opts ~state in
-      let ldir = Stm.get_ldir ~doc:state.doc in
+      let ldir = Declaremods.((get_current_library_info ()).library_path) in
       let state = Vernac.load_vernac ~echo ~check:false ~interactive:false ~state long_f_dot_v in
       let doc = Stm.finish ~doc:state.doc in
       check_pending_proofs ();
       let () = ignore (Stm.snapshot_vio ~doc ~output_native_objects ldir long_f_dot_vio) in
       Stm.reset_task_queue ()
 
-  | Stm.Vio2Vo ->
+  | Declaremods.Vio2Vo ->
       let long_f_dot_vio, long_f_dot_vo =
         ensure_exists_with_prefix f_in f_out ".vio" ".vo" in
       let sum, lib, univs, tasks, proofs =
--- a/toplevel/ccompile.ml
+++ b/toplevel/ccompile.ml
@@ -107,13 +107,13 @@ let compile opts copts ~echo ~f_in ~f_out =
     | NativeOff -> false | NativeOn {ondemand} -> not ondemand
   in
   match copts.compilation_mode with
-  | BuildVo ->
+  | Stm.BuildVo ->
       let long_f_dot_v, long_f_dot_vo =
         ensure_exists_with_prefix f_in f_out ".v" ".vo" in
 
       let doc, sid = Topfmt.(in_phase ~phase:LoadingPrelude)
           Stm.new_doc
-          Stm.{ doc_type = VoDoc long_f_dot_vo;
+          Stm.{ doc_type = Batch(BuildVo, long_f_dot_vo);
                 iload_path; require_libs; stm_options;
               } in
       let state = { doc; sid; proof = None; time = opts.config.time } in
@@ -139,7 +139,7 @@ let compile opts copts ~echo ~f_in ~f_out =
       Aux_file.stop_aux_file ();
       Dumpglob.end_dump_glob ()
 
-  | BuildVio ->
+  | Stm.BuildVio ->
       let long_f_dot_v, long_f_dot_vio =
         ensure_exists_with_prefix f_in f_out ".v" ".vio" in
 
@@ -158,7 +158,7 @@ let compile opts copts ~echo ~f_in ~f_out =
 
       let doc, sid = Topfmt.(in_phase ~phase:LoadingPrelude)
           Stm.new_doc
-          Stm.{ doc_type = VioDoc long_f_dot_vio;
+          Stm.{ doc_type = Batch(BuildVio, long_f_dot_vio);
                 iload_path; require_libs; stm_options;
               } in
 
@@ -171,7 +171,7 @@ let compile opts copts ~echo ~f_in ~f_out =
       let () = ignore (Stm.snapshot_vio ~doc ~output_native_objects ldir long_f_dot_vio) in
       Stm.reset_task_queue ()
 
-  | Vio2Vo ->
+  | Stm.Vio2Vo ->
       let long_f_dot_vio, long_f_dot_vo =
         ensure_exists_with_prefix f_in f_out ".vio" ".vo" in
       let sum, lib, univs, tasks, proofs =
--- a/toplevel/coqargs.ml
+++ b/toplevel/coqargs.ml
@@ -47,7 +47,7 @@ type option_command = OptionSet of string option | OptionUnset
 type coqargs_logic_config = {
   impredicative_set : Declarations.set_predicativity;
   indices_matter    : bool;
-  toplevel_name     : Stm.interactive_top;
+  top_path          : Stm.top_path;
   allow_sprop       : bool;
   cumulative_sprop  : bool;
 }
@@ -112,7 +112,7 @@ let default_native =
 let default_logic_config = {
   impredicative_set = Declarations.PredicativeSet;
   indices_matter = false;
-  toplevel_name = Stm.TopLogical default_toplevel;
+  top_path = Stm.TopLogical default_toplevel;
   allow_sprop = false;
   cumulative_sprop = false;
 }
@@ -440,10 +440,10 @@ let parse_args ~help ~init arglist : t * string list =
       let topname = Libnames.dirpath_of_string (next ()) in
       if Names.DirPath.is_empty topname then
         CErrors.user_err Pp.(str "Need a non empty toplevel module name");
-      { oval with config = { oval.config with logic = { oval.config.logic with toplevel_name = Stm.TopLogical topname }}}
+      { oval with config = { oval.config with logic = { oval.config.logic with top_path = Stm.TopLogical topname }}}
 
     |"-topfile" ->
-      { oval with config = { oval.config with logic = { oval.config.logic with toplevel_name = Stm.TopPhysical (next()) }}}
+      { oval with config = { oval.config with logic = { oval.config.logic with top_path = Stm.TopPhysical (next()) }}}
 
     |"-main-channel" ->
       Spawned.main_channel := get_host_port opt (next()); oval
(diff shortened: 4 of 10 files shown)


library/library.mli
```diff
@@ -80,5 +76,10 @@ val locate_qualified_library :
 
 *)
 
+(** {6 ... }
+    Require = load in the environment + open (if the optional boolean
+    is not [None]); mark also for export if the boolean is [Some true] *)
+val require_library_from_dirpath : mode:require_mode -> (DirPath.t * string) list -> bool option -> unit
```
ejgallego: Please avoid these long lines, I like `ocamlformat` style which is become quite standard in the larger OCaml ecosystem.

ejgallego: Also the doc comment should be fixed I think.

library/library.ml
```diff
@@ -278,48 +278,45 @@ let warn_several_object_files =
...
       when Unix.((stat vo).st_mtime < (stat vi).st_mtime) ->
       warn_several_object_files (vi, vo);
-       resvi
+      resvi
     | Some resvo, Some _ -> resvo
-  in
+    end
+
+let locate_absolute_library ~mode dir =
+  (* Search in loadpath *)
```
gares: Is the comment true? The name of the function seems to suggest that there is no search, while the code does a limited one.

maximedenes: > The name of the function seems to suggest that there is no search

Why? Can you elaborate? "locate" seems like the description of a limited search to me, no?

gares (approved):
LGTM

library/library.ml
gares: Below you have another `locate_` function with a very similar comment.
The two comments should tell the difference.

library/library.mli
maximedenes: By "fixed" you mean to also say that "mode" defines what should be loaded? Maybe a better place for such a description would be near the declaration of `require_mode`, no?

ejgallego (commented):
Umm, I need to think more; some of the stuff here is good, some of the stuff is fishy [not saying it is the fault of the PR, I guess it was already fishy before]

I guess that a summary of how things are supposed to work [preferably in a comment] is extremely desirable.

library/library.ml
maximedenes: Well, as their name indicates, one takes an absolute name, and the other one a qualid. Isn't it clear from both the types and the names?

Note that this PR doesn't introduce any of this.

library/library.mli
ejgallego: By fixed I meant that the `ocamldoc` is not correct, shouldn't have a section and should use the standard presentation format [function arg1 arg2 ...]

But indeed after reading the PR I realize we also need proper documentation, otherwise the situation here un-understandable.

ejgallego (changes requested):
Please document the new flag as well as the behavior of the function.

The previous situation was already bad, as one couldn't figure which files would be loaded by looking at the documentation, but we cannot allow to make it worse.

gares:
`vos` needs yet another require mode.
This PR factors things so that the main PR about `vos` can "just" add one mode.
The mode is known to the toplevel (compilation mode) but has to reach library, hence the threading of the flag trough the STM.

library/library.mli
maximedenes: Note that this PR doesn't introduce or change the faulty comment.

I'll try to improve the documentation, though.

gares:
I frankly don't see how this PR makes things worse. The new require mode data type is a sort of documentation to me.

maximedenes:
> I frankly don't see how this PR makes things worse.

Well, if I kept thins as they were, we'd have only one `require_mode`. I introduced a new one because you mentioned compilation of `.vo` files should not load `.vio` files, which it used to do.

ejgallego:
I mean makes the non-existing documentation worse; documentation for this _is critical_ as otherwise there is no way to keep build system rules in synch.

library/library.mli
ejgallego: > Note that this PR doesn't introduce or change the faulty comment.

This PR modifies the function so for sure has to update the doc comment. I know it is a pain but it also modifies critical rules, in particular w.r.t. build system so I'm much afraid that the new rules have to be properly documented before we let this in.

It is true that the rules should be already be documented in the past, but the fact that in the past we had laxer standards doesn't mean today we have to keep them.

maximedenes: Sure, no pb, I'm already writing the new rules :)

maximedenes:
@ejgallego Should be better now, let me know if some things are still unclear.

library/library.mli
```diff
@@ -70,8 +65,17 @@ exception LibUnmappedDir
 exception LibNotFound
 type library_location = LibLoaded | LibInPath
 
+type require_mode = RequireOnlyVo | RequireMostRecent
+(** Describes what files should be looked for at require time.
+
+
+    - [RequireOnlyVo] means only look for `.vo` files
+    - [RequireMostRecent] means look for `.vio` and `.vo` files, and if both
+      exist, emit a warning and load the most recent of the two
```
gares: The warning is emitted only if the .vio is more recent than the vo.
Otherwise the .vo is loaded, and no warning is emitted.

library/library.ml
gares: Oops, then OK. leave it like that.

What puzzles me as a reader is that if a "path" is absolute, then I don't get why some "location" is needed.
If any, this "location" is a different from the "location" that resolve a, say, relative path to an absolute one. It is more like the location that resolve a file name to sectors and cylinders. You are not the author of the mess (mixing the two concepts).
If it is clear to you what lines 312-313 actually do, then please comment it (and prune the two silly comments).

ejgallego:
> @ejgallego Should be better now, let me know if some things are still unclear.

Things are much better, thanks. Now let's discuss the rules for a second:
```
    - [RequireOnlyVo] means only look for `.vo` files
    - [RequireMostRecent] means look for `.vio` and `.vo` files, and if both
      exist, emit a warning and load the most recent of the two
```
do we want to keep this second case as is? Should we just get rid of this flag and have a simpler rule:
"always try vo, and if vo doesn't exists, try vio"

maximedenes:
> do we want to keep this second case as is? Should we just get rid of this flag and have a simpler rule:
> "always try vo, and if vo doesn't exists, try vio"

Please discuss this somewhere else, it is not related to this PR which does not change the behavior there.

However, the comment is slightly incorrect as @gares mentionned, I need to update it.

ejgallego:
> Please discuss this somewhere else, it is not related to this PR which does not change the behavior there.

Sorry but no; this is very related to this PR, which is altering the require import rules. If my reasoning is correct, then this PR and a flag is not needed at all.

ejgallego (changes requested):
The new flag has to be introduced as to workaround a questionable behavior of the current implementation. However, this is not the only solution, choices are:
- keep the old behavior, add a new flag,
- correct the old behavior, don't introduce a flag.

The second option is much preferred as adding a parameter just for this to a criticla path such as `Vernacentries.interp` does signal we are doing something pretty wrong.

gares:
This PR does two things:
- fix the behavior
- factors some code and introduce a flag so that the main PR about vos can be implemented in a simpler way

It can be separated in two PR, but both then needs to be merged before the main vos one, since they touch the very same piece of code. Not worth it in my opinion.

ejgallego:
> It can be separated in two PR,

Indeed I don't think that is needed; IMO the PR would be good to go as is; my question is different, do we need the flag at all?

ejgallego:
When I talk about "old behavior" I mean the behavior after the fix this PR does.

Indeed "interface" files are really tricky from a build point of view. I learned about some of the problematics when I discovered the gimmicks Dune has to do in order to handle `cmi` properly.

This includes computing an updated list of targets for an incremental build and then wiping out all the rest of the build outputs including old interface files.

I guess we can start by listing the current use cases, let's see:
- regular vo compilation: we produce a vo, we never want to input a vio?
- regular vio compilation: we produce vio, we want to input always a vo, and if no vo a vio?
- vio 2 vo compilation: in this case we don't need to input anything, do we?

maximedenes:
> do we need the flag at all?

I think there are two steps:
1. with `.vos` files, it is pretty clear that the semantics of `Require` is not the same as when producing a `.vo` file. #8642 introduces a global flag for this purpose, this PR puts in place the infrastructure needed to avoid such a global flag by threading it.
1. without `.vos` files, @gares told me it would be better that compilation of `.vo` files does not pick up `.vio` files on `Require`.

In the current PR, you see only 2., so you get the feeling we could avoid it in principle, but as the title and description suggest, I'm interested in 1., where a flag seems unavoidable.

I agree with you that threading it to vernac seems heavy, but IMHO this is a consequence of managing dependencies via a vernac command. Hopefully we can improve on that later, but the threading is just making the current concepts and their complexity explicit. It is not making things worse per se.

ejgallego:
> 1. with `.vos` files, it is pretty clear that the semantics of `Require` is not the same as when producing a `.vo` file. #8642 introduces a global flag for this purpose, this PR puts in place the infrastructure needed to avoid such a global flag by threading it.

Do you mind reminding me of the `vos` require semantics? I'd like to have indeed a complete picture.

> I agree with you that threading it to vernac seems heavy, but IMHO this is a consequence of managing dependencies via a vernac command. Hopefully we can improve on that later, but the threading is just making the current concepts and their complexity explicit. It is not making things worse per se.

We already have other places where the information could be stored, for example at the time we do `Declaremods.start_library`

ejgallego:
> We already have other places where the information could be stored, for example at the time we do `Declaremods.start_library`

So for example that seems fishy to me, that we don't story the info about the current library in a single central place.

ejgallego:
> Do you mind reminding me of the `vos` require semantics? I'd like to have indeed a complete picture.

Indeed I am worried about keeping things in sync with the build system; which is not trivial. Once Dune is the single build system we will be able to factor out a lot of duplication for example w.r.t. coqdep / require, which are already inconsistent, but for example, are we checking that require is now in sync with what `coqdep` produces?

ejgallego:
Another point that concerns me quite a bit is the interactive case.

In modern UI protocols, you can edit a document, and when pressing save you should be able to write the `.vo` file; this use case seems to interact pretty badly with the a modal require flag.

The problem is indeed way deeper, and it seems to me that the only way to solve this is indeed to have a single `.vo` file format, but of course that's beyond to what this PR can do; however, not having the flag would help in the above case as the controller could still work as long as it determines the output format before starting the interactive session.

gares:
I don't think this is the place to discuss all this.
vos should only load vos, hence we need yet another mode. in the main PR it is implemented in the very same way, but instead of reading a locally passed "mode", it reads a global boolean in Flags.

ejgallego:
> vos should only load vos, hence we need yet another mode. in the main PR it is implemented in the very same way, but instead of reading a locally passed "mode", it reads a global boolean in Flags.

Then why for example is this information not provided to `start_library`?

maximedenes:
> Do you mind reminding me of the `vos` require semantics?

You can see it here: https://github.com/coq/coq/pull/8642/files#diff-66c3823693299fb996b3cb2838ef6e14R294 together with a comment from you saying we should pass it as an argument ;)

ejgallego:
> vos should only load vos, hence we need yet another mode.

The more I think the more I do believe modes are not the right choice for this.

How can it be that `vos` only load `vos`? So they cannot load modules from regular compiled libs?

ejgallego:
> You can see it here: https://github.com/coq/coq/pull/8642/files#diff-66c3823693299fb996b3cb2838ef6e14R294 together with a comment from you saying we should pass it as an argument ;)

Thanks, so indeed Enrico comment is not accurate, `vos` try to load `vos`, then if not present they do load the `vo`.

ejgallego:
> together with a comment from you saying we should pass it as an argument ;)

Indeed that helps to reason about what is going on; now the point is why do we need modes?

Let's focus on vos just a moment, why it isn't the case that rule should always be: "try vo, if not try vos".

maximedenes:
> Thanks, so indeed Enrico comment is not accurate, `vos` try to load `vos`, then if not present they do load the `vo`.

What should not happen is that `.vo` or `.vio` files load `.vos` files.

maximedenes:
> Let's focus on vos just a moment, why it isn't the case that rule should always be: "try vo, if not try vos".

Because that would make no sense when producing a `.vo`?

maximedenes:
> Let's focus on vos just a moment, why it isn't the case that rule should always be: "try vo, if not try vos".

At this point I think we'd need @charguer to describe why a specific strategy is needed for `Require` in the case of `.vos` files.

ejgallego:
> Because that would make no sense when producing a `.vo`?

Umm, well, that should work but indeed we may not want that.

Let's see the thing from a different point of view, what does `coqdep` produce on this file `B.v`, living on library `A`:
```
Require Coq.Init.Prelude.
Require A.foo.
```
then we need to fill:
```
B.vo: B.v ???
B.vos:  B.v ???
```

ejgallego:
So indeed the only thing that bothers me is the conditional require rules, as these maybe not properly be handled at the build level. Imagine this situation, we have our library `A` and two cases:

- `stdlib` is installed, as regular `.vo` files.
- `stdlib` is vendored, and unbuilt, and we want to do a global `vos` build.

So indeed here the build system can know the scope of the current `vos` build profile, however Coq may have a hard time; we could tho fix this by build hygiene I think.

ejgallego:
> Then why for example is this information not provided to `start_library`?

Also if the flag is OK, this is still a valid question, right?

maximedenes:
My memory must be failing me, but wasn't all this discussed during a WG already, as part of #8642, including the `.vok`, empty `.vos` stuff, and so on?

maximedenes:
> > Then why for example is this information not provided to `start_library`?
> 
> Also if the flag is OK, this is still a valid question, right?

I'd like to understand the question better. As it is, I'd be tempted to say we don't pass it to `start_library` because it does not need the information.

gares:
Let me just repeat that this PR is for a cleanup, the design of vos should be discussed elsewhere, for example in the main PR. Sure, if the design is questioned, then this PR is also to be questioned. But given that we agree on the design of vos I don't get what we are doing here.

ejgallego:
> Let me just repeat that this PR is for a cleanup, the design of vos should be discussed elsewhere, for example in the main PR.

This is not a cleanup, it introduces a new flag in core parts of the system; it just doesn't move an old flag. The new flag has to be justified.

ejgallego:
> I'd like to understand the question better. As it is, I'd be tempted to say we don't pass it to `start_library` because it does not need the information.

Well, information about the current library is there and this indeed is critical for loadpath, etc... so it seems to me that the kind of library under check should be rather stored centrally there.

ejgallego:
> My memory must be failing me, but wasn't all this discussed during a WG already, as part of #8642, including the `.vok`, empty `.vos` stuff, and so on?

Sure, we did discuss, however that doesn't mean the implementation should not undergo review.

ejgallego:
I do agree that most discussion does belong to the vos PR; I think a quick way indeed to get this PR thru is to remove the threading of the flag and add instead a parameter to `start_library`.

That should make the patch much more self-contained.

ejgallego (changes requested):
There is some more discussion to about require behavior in the vos PR, I suggest for now we store the mode of the library at `start_library` time, instead of threading the flag.

ejgallego:
Review updated, other than moving the flag this is OK to me.

gares:
@ejgallego If I understand it correctly, you want the flag to be passed to start_library, stored in an imperative field in the library module, and read from there at interpretation time, eg when a `Require` is executed.
This makes Coq unable to start more than one library at a time. Today Coq is not able to process more than one file at a time, but I recall some work of yours to handle multiple documents. In that scenario the imperative flag seems counterproductive. Am I missing something?

ejgallego:
> This makes Coq unable to start more than one library at a time. Today Coq is not able to process more than one file at a time, but I recall some work of yours to handle multiple documents. In that scenario the imperative flag seems counterproductive. Am I missing something?

The flag has to be handled similarly to `openmod_info` of course to work. In fact what I propose is way more robust w.r.t. multiple documents than the current proposal.

gares:
@ejgallego you are saying to put in `Declaremods` a `Summary.ref` set by `Declaremod.start_library` (called by `Stm.new_doc`) and accessed by `Library.locate_qualified_library` (called by `Stm.vernac_interp`), right?

I was not very happy with an internal ref in a module (namely `Library`); what you propose, if I got it right, is a ref internal to a component `library/`, that is not better. What is really out of place, to me, is that this setting has nothing to do with the library itself, but rather with how you are generating/loding it. This piece of info fits, at best, in the document manager. I don't see how it can be pertinent to the library component. I don't like the threading much, but I prefer `locate_qualified_library` to take I/O specific options (which files should be loaded, where, how...), rather than polluting the component with that piece of info. Finally the approach you propose fixes the flag at library creation time that is quite coincidental to me. Wrt robustness, I really don't know what you mean. Robust in face of what?

The current solution is OK to me, but the one proposed by @ejgallego (if I got it right) would also work, even if I find it worse than the current one. @ejgallego do you confirm that my understanding of your proposal is correct?

ejgallego:
> @ejgallego you are saying to put in `Declaremods` a `Summary.ref` set by `Declaremod.start_library` (called by `Stm.new_doc`) and accessed by `Library.locate_qualified_library` (called by `Stm.vernac_interp`), right?

Yup, the thing is that we need to discuss what is going on with `locate_qualified_library`, but is this PR is not the place I'd say we just plug the piece of info in the less disruptive place, we already have open_mod_info so indeed as long as we have that we can add the kind of the library there.

I agree it would be better to clean this stuff up, but on the other hand if we piggyback on `start_library` we can tell the right layer what kind of library is, after all this component is the one that makes the choice about what file to load.

Having to thread such flag in the top-level vernac interp is IMO wrong, at least in its current form.

> has nothing to do with the library itself, but rather with how you are generating/loding it

well the setting tells if the library is a vo, a vio, etc... so IMO it has a lot to do with the library; in fact, if you store that info at start_lib time you can also implement a uniform save, etc....

It is an ugly asymmetry now that file selection is done in library, whereas the choice of saving commands is done in the STM / toplevel.



> rather than polluting the component with that piece of info

You are facing "polluting" the component [adding one field to an existing variable] vs polluting the whole system. I rather keep the modifications local.


> Robust in face of what?

Robust in the face that the upper layers don't have to handle and thread a parameter they don't care about at all.

> This piece of info fits, at best, in the document manager.

Yup, we can discuss this, in fact I was trying to but you told me this PR was not the place, so I dunno. What I propose allows the PR to move foward.

ejgallego:
A good hint that threading this flag may not be the best way to proceed is that after document creation, the flag is never supposed to change.

Anyways, I'd say let's indeed let's not block this for a long time and aim for the less invasive patch for now which is quite trivial and move ahead to the real discussion which is about build rules and dependencies among files.

gares:
> well the setting tells if the library is a vo, a vio

No, the setting tells what should be loaded when building a vo, vio, vos.. not what the library will be.
Note that all these "libraries" have the same format, so you can load one or the other. The flag in question governs the policy (only this file extension, the most recent between this and that...). For example the current policy for vos is there to maximize chances of parallelism when building with make without incurring in races. This kind of policy has little to do with the library itself, it may even be the wrong one if you are building vos with, say, dune. To me it really belongs to the outer layers.

> the flag is never supposed to change.

Why? The current code does not change it, right. But why it should not be changed? It is true that we have no way (protocol message) to tell coq to "degrade" its safety requirements and load anything that fits (without restarting coq), but it is not hard to imagine a UI that proposes you that (exactly as browsers propose you to visit a website even if its SSL certificate is expired).

ejgallego:
> Why? The current code does not change it, right. But why it should not be changed? It is true that we have no way (protocol message) to tell coq to "degrade" its safety requirements and load anything that fits (without restarting coq), but it is not hard to imagine a UI that proposes you that (exactly as browsers propose you to visit a website even if its SSL certificate is expired).

That sounds a bit strange for example if we go ahead with the proposal of only allowing require in the header of the document, doesn't it? I am not seeing this use case well; I can imagine users confused seeing this kind of information [as they are with browsers]

> No, the setting tells what should be loaded when building a vo, vio, vos.. not what the library will be.

Indeed, it tells the loading requirements which is currently derived from the document type, right. I but I can't say more without getting in the discussion of how the current system works [I do think that no flag should ever exists on the first place], which we agreed was not to be done in this PR.

maximedenes:
I can put a ref in declaremods, as suggested by @ejgallego.

gares:
OK, so I don't touch the milestone

maximedenes:
@ejgallego In `vio2vo` mode, we don't call `start_library`, what did you have in mind for that case?

ejgallego:
> @ejgallego In `vio2vo` mode, we don't call `start_library`, what did you have in mind for that case?

Good question, I guess it does inherit the flag that was set when the vio was created, @gares ?

As of today without start library things just don't work [and something strange happens in the kernel that should not, as we recently saw in our trace replay experiment]

> I can put a ref in declaremods, as suggested by @ejgallego.

You can even reuse the one already there.

stm/stm.ml
```diff
@@ -1139,10 +1141,16 @@ let show_script ?proof () =
 
 end
 
+let doc_require_mode = function
+  | Interactive _ -> Library.RequireMostRecent
+  | Batch (Vio2Vo, _) -> Library.RequireOnlyVo
```
ejgallego: Umm, aren't we supposed in `vio2vo` to be able to run in parallel thus using `vio` files?

Oh it is just the case that vio2vo will never require a file?

gares: vio2vo does not execute `Require`, it only executes tactics (if there is a Require in the middle of a proof, then such proof is not delayed).

gares: @maximedenes the thing I asked you to change is in the v2vio phase, today we load the most recent between vo and vio, while I think we should only load vio

ejgallego: That's what I thought, @maximedenes , does it answer you question?

ejgallego: > while I think we should only load vio

With vo as a fallback right?

gares: This is what happens today. But I believe it should only load vio files (at least, this is what the dependencies coqdep generates dictate)

SkySkimmer: Internal dependencies are expected to be loaded from vio, but external ones (typically the stdlib) have only vo available right?

ejgallego: I'd be OK with that, however what about system-installed libs? As of today these are mostly installed as vo-only, right?

charguer:
[Catching up]

I have documented in my PR the desired behavior for loading a library when the compilation target is vos or vok.
```
   ---when compilation target is vos or vok---
   If the .vos file exists and is not empty, it describes the library.
   If the .vos file exists and is empty, then load the .vo file.
   If the .vos file is missing, then fail. 
```
The rational for this behavior is that we don't want to load the "most recent file", as this confuses make and leads to loading stale files. I can't recall exactly why it is bad to "load vo if vos does not exist" in general, but in the context of using a makefile you want to have a dependency "vos->vos", thus there is no reason to load a "vo" when you expect the "vos" to be ready. (I don't think it's a good idea to have rules of the form "vos" depends on "foo.vo" or "foo.vos" in a build system if you can avoid it.)

Remark: the idea of producing an empty "foo.vos" file when producing "foo.vo" is precisely to have a placeholder to indicate that the dependency on "foo.vos" is satisfied when "foo.vo" exists. I know it's a bit ugly to distribute bunch of empty files, but it is the simplest solution that I could think of that works with Makefiles. You could argue that we should not use Makefile, but the fact is that many people do use Makefile, and we don't want to make their life hard.

Regarding the behavior of library loading for other compilation targets, here is what I would suggest:

```
   ---when compilation target is vo or coqchk---
   Only accept to load a vo file.

   ---when compilation target is vio---
   Load the most recent of .vo or .vio or .vos
```

By comparison, the current policy is, as far as I understand:

```
   ---when compilation target is coqchk---
   Only accept to load a vo file.

   ---when compilation target is vo or vio---
   Load the most recent of .vo or .vio
```

ejgallego:
Thanks @charguer ,

> thus there is no reason to load a "vo" when you expect the "vos" to be ready. (I don't think it's a good idea to have rules of the form "vos" depends on "foo.vo" or "foo.vos" in a build system if you can avoid it.)

The main problem with this is w.r.t. to libs installed system wide. For example for `vio`, installing them does increase space by a lot, so it is not very viable.

`vos` do fare better so they could be always installed maybe.

> I don't think it's a good idea to have rules of the form "vos" depends on "foo.vo" or "foo.vos" in a build system if you can avoid it.

I don't think that rule is so hard to write, but not a big deal anyways; we could improve that later on.

> Regarding the behavior of library loading for other compilation targets, here is what I would suggest:
> 
> ```
>    ---when compilation target is vo or coqchk---
>    Only accept to load a vo file.
> ```

I am confused here, so how can you produce them the vo in parallel now?

charguer:
> vos do fare better so they could be always installed maybe.

It makes no sense in my opinion to distribute vos files with a proper contents for libraries.
They consist of just prefixes of the vo files that you would also distribute alongside.

Recall that vos files are only meant to improve the workflow of developing proofs. They serve no purpose for stable, distributed libraries.

> I am confused here, so how can you produce them the vo in parallel now?

vos files have never been meant to produce vo files in the end.

The vio-2-vo process could be used to produce vo files, however (as far as I understand) it lacks the checks of universe consistency, which is hard to do when proofs are done in parallel as you need some means of gathering all constraints at the very end. I have argued (and others have agreed) that a vo file should be an evidence that a .v file is correct, including w.r.t. universe consistency.

To me, it is not a problem in practice. During the day, you build only vok files, which you can do in parallel. Once every night, you build your vo files, respecting the dependency graph (with some amount of parallelism exploitable, but less than for vok files in general). Then you're happy because your day is efficient, and your night is trustworthy :)

ejgallego:
I see, I am not against what you propose, however that seems to disable important functionality of the vio system, so that'd need @gares approval [and modification of the docs]

> I have argued (and others have agreed) that a vo file should be an evidence that a .v file is correct, including w.r.t. universe consistency.

I do agree, however the situation is not so simple for example when we consider incremental compilation; but indeed that is orthogonal to this PR. I do believe that in the end, `.vo` files will need some metadata about their trust status.

Regarding the empty file check, while it seems uglier to me than a conditional check, it can be fixed separately.

maximedenes:
> > I can put a ref in declaremods, as suggested by @ejgallego.
> 
> You can even reuse the one already there.

@ejgallego which one do you mean?

I see `Declaremods.openmod_info`, but that's about the current module, not the current library.

ejgallego:
> I see `Declaremods.openmod_info`, but that's about the current module, not the current library.

Oh indeed sorry, maybe it is better to add a new ref after all.

gares:
What do we do with this one (and #8642)?

ejgallego:
> What do we do with this one (and #8642)?

This one I can take care of doing what I proposed if you want. That should be quite easy. However @gares :

> I see, I am not against what you propose, however that seems to disable important functionality of the vio system,

What's the deal with this? Do you agree to remove parallel compilation of `vio -> vo` ?

About the other PR I dunno, I didn't seen anything blocking for the merge but of course given that the branching point is scheduled for tomorrow and the code needs a rebase it is @vbgl who will say.

I could review the parts I care on #8642 quickly if needed.

maximedenes:
I plan to finish this PR later today or early tomorrow.

ejgallego:
> I don't see how this PR is related to that. Indeed I don't want to remove this feature, but I still want to merge this PR.

@gares this PR forces Coq in `vio2vo` mode to load only `.vo`, thus you lose most of the parallelism.

charguer:
W.r.t. vio2vo forcing vo loading, I think that we all agree (do we?) that we don't want "vo" files created if they have not checked universes.

So, one middle ground would be to continue allowing vio2vo loading vio file for 8.10, but warn that this will go away in the next release.

A simpler route for developers would be to remove this ability right now (for 8.10), under the guess that the number of users that need the parallelism AND that cannot use the new vos system is so few that we should not bother warning them one release ahead. Concretely, if these users can use the n-1 released version of Coq for the next 6 months, then it won't make a difference for them compared with the previous plan.

I suspect that the current PR is choosing to follow the second route, am I guessing right?

ejgallego:
I am personally neutral as to what to do with vio to vo ; but if the parallel generation disappears some other parts of the system may need adjustment [including the docs and the scheduler]

ejgallego:
> I think that we all agree (do we?) that we don't want "vo" files created if they have not checked universes.

The reason I am neutral is that I do believe that in the end we will need a different mechanism to check trust that can take into account some form of incremental compilation and file extension alone is not gonna be enough.

gares:
For the meaning of .vo I disagree. It is like that since 8.5, documented, and nobody complained. Also it is very easy to turn the current "incompletely checked" vo files into fully checked ones. Not to talk about you can already validate them with coqchk if you want.

In any case this pr was not blocked because of that, but because of the way a flag was threaded from the top-level component to the library one.

Wrt he issue about only loading vo during bio completion the current code is broken, my fault, Maxime knows it. But again this is not why the pr was blocked.

ejgallego:
I am unsure I understand what you mean @gares but there was ample discussion on both the threading and the semantic change. In fact the first question is why we do need a flag at all.

What is your position on the vo issue? If we take the patch as is certainly some other changes should be done.

gares:
The patch needs fixing wrt what is loaded when building vio files iirc. I will check the details tomorrow with Maxime.

Installing vos files just to avoid a flag is overkill.

maximedenes:
@ejgallego does it look better to you now?

gares (approved):

ejgallego (commented):
It looks OK, thanks! Don't forget to update the description of the PR as it seems out of date now.

I am not a fan of leaking "interactive/vo" etc... to `Declaremods` as it was not so long ago we moved it out of there, but we can live with that for now [I would just have added the flags and kept the require <-> file type logic in the document manager]

ejgallego (approved):

library/library.ml
```diff
@@ -278,48 +278,61 @@ let warn_several_object_files =
...
+    - [RequireOnlyVo] means only look for `.vo` files
+    - [RequireMostRecent] means look for `.vio` and `.vo` files, and if both
+      exist, emit a warning and load the most recent of the two
+*)
+
+let library_require_mode info =
+  let open Declaremods in
+  match info.library_mode with
+  | Interactive -> RequireMostRecent
+  | Batch (Vio2Vo) -> RequireMostRecent
```
ejgallego: the parenthesis here look weird.

maximedenes: lol, right, will fix.

maximedenes:
> I am not a fan of leaking "interactive/vo" etc... to `Declaremods` as it was not so long ago we moved it out of there

Yeah this is a bit delicate. My current understanding is that some logic that I put back in library should really be in the `coqc`/`coqtop` init code. I don't think the semantics of `Require` really belong to the STM. In particular, `start_library` should not be called by the STM, but by `ccompile` and the likes.

ejgallego:
> In particular, `start_library` should not be called by the STM, but by `ccompile` and the likes.

Umm, well this is a delicate point due to state which is owned by the document manager, see #1089 in which we moved this to the STM and it was the right thing to do IMO.

 In other designs indeed it is the case that `ccompile` doesn't exist at all, as it is equivalent to:
```
DM.start_doc
DM.submit_check
DM.save
```

ejgallego:
> I don't think the semantics of `Require` really belong to the STM.

Definitively they do; as it is for example the component that controls most of the `vio` logic.

`coqc/coqtop` should really be almost empty wrappers, if they are not is because the STM API is limited w.r.t. batch mode, but see for example a different toplevel [here](https://github.com/ejgallego/coq-serapi/blob/v8.10/sertop/sercomp.ml)

ejgallego:
In fact I am but convinced now that the STM/DM should be the component managing the file-part of `Require`, as for example, trigger recompilation, downloading a missing library, checking consistency status, etc...

I thought about that some time ago but it was too invasive, however now interpretation can take a type of handlers and do this properly; that would also clean the structure of the forward-traversal document manager a lot due to types.

maximedenes:
Maybe this is a matter of terminology, but I think there should be some separation between the STM (which schedules and delegate proofs) and the document manager which can also handle what you wrote (file handling and co).

ejgallego:
> > In particular, `start_library` should not be called by the STM, but by `ccompile` and the likes.

Going back to your comment @maximedenes , indeed after thinking a bit more I do agree with you, I was at first confused as here we are giving `start_library` a semantics that IMHO shouldn't have.

I have becomed convinced that the original approach you took was going on the right direction, I proposed the `start_library` thing as to have a quick way so we could think more of the problem.

Indeed, if something, I do believe your original PR should have go further and indeed removed all the logic at the `library` level in favor of a library resolver structure; some preliminary steps towards the deeper solution are now in #9895 [which should rebase cleanly on top of this] but maybe we want to merge the two.

> Maybe this is a matter of terminology, but I think there should be some separation between the STM (which schedules and delegate proofs) and the document manager which can also handle what you wrote (file handling and co).

Indeed, even if now in Coq both roles are done by the same component mainly; I kind of become confused because we had loadpath logic in the `library` layer, but IMHO it clearly does belong in the document manager.

In fact after doing #9895 , it is clear that moving the resolve logic would we like to `ccompile` for example wouldn't break the invariants that #1089 tried to enforce; modulo the `Add LoadPath` command which could be a bit problematic.

Maybe that command should be deprecated, or we could store the resolver status in `Vernacstate` [however this last choice may be less flexible w.r.t. alternative resolvers]

ejgallego:
Note that my comment is just informative I am still very OK if you want to merge this PR; I'm sorry my request to put the flag more inside instead of more outside was not the correct one.

Somehow I saw the original submission in the "middle" and I thought it would be easy to push it downwards, but indeed I was wrong.

gares:
@maximedenes I've restarted the jobs that failed (they seemed spurious).
If tests pass, should I merge this?

ejgallego (dismissed):
If this is targeting 8.11 then it should be done on top of #9895

gares:
@ejgallego my understanding of #9895 is that is moves library name resolution from Library to Vernac, and I don't think it belongs there, so I don't think it is ready. On the contrary you say " I am still very OK if you want to merge this PR" so this one seems to be ready and can unblock #8642. @maximedenes is this PR OK, of you discussed with @ejgallego on another channel about moving this PR on top of #9895 ?

ejgallego:
> I am still very OK if you want to merge this PR" so this one seems to be ready and can unblock #8642.

I was OK with this imperfect solution as not to miss 8.10, however given that it is not making it we should improve the approach.

ejgallego:
> @ejgallego my understanding of #9895 is that is moves library name resolution from Library to Vernac, and I don't think it belongs there, so I don't think it is ready.

IIRC it was @maximedenes who suggested that such move would make sense, at first I thought not, but then later on I changed opinion. So indeed if @maximedenes thinks this shouldn't be based on that PR we can merge as-is, but I think the architecture seems clearer the other way around [and it contains a nice refactoring w.r.t. the change here]

maximedenes:
> IIRC it was @maximedenes who suggested that such move would make sense, at first I thought not, but then later on I changed opinion. So indeed if @maximedenes thinks this shouldn't be based on that PR we can merge as-is, but I think the architecture seems clearer the other way around [and it contains a nice refactoring w.r.t. the change here]

I need to review the current status of both PRs before I can comment. However, with the release and ome prioritary issues I need to address, it probably won't be before next week. Sorry about that.

ejgallego:
> I need to review the current status of both PRs before I can comment. However, with the release and ome prioritary issues I need to address, it probably won't be before next week. Sorry about that.

Oh no worries, let us know if you need help with the .vos stuff ; I assume it is missing 8.10 right?

After a quick review this seemed like the principal obstacle but IMO either choice should do the job.

maximedenes:
Ok, I got a bit lost in the discussion, but I now rebased this. Does it look ok, @ejgallego?

ejgallego:
Give me a couple of days to review this; after a quick scan I think it looks OK, however I am surprised you didn't hook into `Loadpath` directly, I guess I need to do a bit of thinking and review of the above discussion again.

maximedenes:
@ejgallego @gares I'm afraid that if we don't merge this today, the support for `vos` files will miss 8.11, as I'll be on holidays on Thursday and until after the freeze...

Could I get a review?

ppedrot:
I am not sure what the standard policy is, but if the PR is accepted by reviewers during your holidays and the PR is mergeable as-is, we can merge it before the freeze.

gares:
I was fine even with the global flag. Now the imperative state is in vernac and is write once, so it is even better. As some point it was even functionally passed, but I guess I should be happy with a local maximum.

gares (approved):
I'm fine with this, as I was fine before. But I think @ejgallego is the one that was not OK with the code and should express his opinion again...

ejgallego (changes requested):
Either I am missing something, or I don't understand the patch at all.

As far as I can see the changes are a no-op, so what is the purpose? For example, second commit says "Make explicit the semantics of Require w.r.t. compilation target" but no actual require code is touched.

Maybe something was lost in the rebase?

ejgallego:
Note the comment on the original PR description "We do introduce one semantic change: when producing a .vo file, we no longer look for .vio files." seems outdated too. Also the stuff about `RequireOnlyVo` etc... seems missing.

IIRC it seems weird to have to add this bit of info in `Declaremods`, wasn't it a property of the library resolver? Summarizing [my memory] of the discussion above, some points we seemed to agree were:
- resolving a library is a responsibility of the build system / document manager. After this we moved `Loadpath` to vernac, but ideally it should be set above. I guess the easiest would be to have a flag in `Loadpath` which is controlled by the STM, in general you may want to pass a `resolve` structure to vernacular interpretation but that's for future work [we could even layer that]

maximedenes:
Ok, I probably got lost, I couldn't make sense of the discussion again. Let's close this and merge the other PR with the global flag, as was decided during the WG. I will let you prepare a PR that then removes the global flag.

Issue #6840, opened 2018-02-26
Compatibility notations do not have the same semantics as normal notations (or only parsing notations).


Zimmi48:
#### Version

8.8+alpha (current master)

#### Description of the problem

```coq
Notation "[ x ]" := (cons x nil) : list_scope.
Notation "[ x ; .. ; y ]" := (cons x .. (cons y nil) ..) (compat "8.6") : list_scope.

Variable Foo : nat -> nat.

Delimit Scope Foo_scope with F.

Notation " [ x ] " := (Foo x) : Foo_scope.

Check ([1] : nat)%F.
```

works.

```coq
Notation "[ x ]" := (cons x nil) : list_scope.
Notation "[ x ; .. ; y ]" := (cons x .. (cons y nil) ..) : list_scope.

Variable Foo : nat -> nat.

Delimit Scope Foo_scope with F.

Notation " [ x ] " := (Foo x) : Foo_scope.

Check ([1] : nat)%F.
```

doesn't.

```coq
Notation "[ x ]" := (cons x nil) : list_scope.
Notation "[ x ; .. ; y ]" := (cons x .. (cons y nil) ..) (only parsing) : list_scope.

Variable Foo : nat -> nat.

Delimit Scope Foo_scope with F.

Notation " [ x ] " := (Foo x) : Foo_scope.

Check ([1] : nat)%F.
```

doesn't either.

I thought the compatibility annotation was just about emitting warnings but apparently it changes also the semantics. I don't know if that's wanted or not.

JasonGross:
This is deliberate.  Compatibility notations are so that when you issue `-compat`, you get the notations.  I knew this when I added the compat list notations, and the effect is exactly what was intended; I don't know if other notations were added with the same knowledge.

Zimmi48:
OK thanks for the explanation. But why does the following works then? Isn't it an inconsistency?

```coq
Parameter bar : Prop.
Notation foo := bar (compat "8.6").
Check foo.
```

Zimmi48:
If I understand correctly, there are compatibility notations and compatibility aliases with the same syntax but not the same semantics and this is were my confusion comes from.

JasonGross:
Ah, that seems plausible.  Are there any compatibility notations other than the list ones which are not aliases?  (It's possible that I asked for the compatibility notation behavior when changing the list notations; the goal was to make `-compat` mimic 8.4 behavior even in the corner cases, as closely as possible, while allowing better behavior in 8.5.)

[...]

PR #9811, opened 2019-03-21
[stdlib] Remove deprecated module Zlogarithm

vbgl:
This module has been deprecated for quite a long time (since version 8.4). Removing it is a step towards #7878.

This still needs a simple overlay for CompCert/VST that I’ll submit if this PR gets some support.

Update: overlay for CompCert at https://github.com/AbsInt/CompCert/pull/286 and for VST at https://github.com/PrincetonUniversity/VST/pull/326


Zimmi48 (approved):
I don't mind the removal, especially if the porting is straightforward. Note that we don't have yet module-wide deprecation attributes, so some users could have been unaware of the deprecation if they hadn't read doc. But again, that doesn't matter too much if the porting is straightforward.

maximedenes (approved):

maximedenes:
@herbelin you're the owner here.

xavierleroy:
I removed the dependency on Zlogarithm in CompCert.  I still think the way you're going about this removal is wrong.

> Note that we don't have yet module-wide deprecation attributes, 

Then put deprecated attributes on every definition and theorem of the module.  How hard is this?

> so some users could have been unaware of the deprecation if they hadn't read doc.

Some users, yeah.  Once a proof works, nobody goes back to the comments of the stdlib modules at every release to make sure someone hasn't decided to deprecate it.  I don't see any mention of this deprecation in the CHANGES file either.  You're really expecting your users to be clairvoyant?

vbgl:
> Then put deprecated attributes on every definition and theorem of the module.  How hard is this?

I guess we don’t have deprecation attributes for definitions or for
theorems…

xavierleroy:
I'm pretty sure I've seen Coq emit warnings like "Foo.bar is deprecated".  I have no idea how those are implemented and generated, but the corresponding mechanism is what I meant by "deprecated attributes".

maximedenes:
> I guess we don’t have deprecation attributes for definitions or for theorems…

I just tested, and it seems worse than that: the `deprecated` attribute is accepted on definitions but apparently has no effect.

> I'm pretty sure I've seen Coq emit warnings like "Foo.bar is deprecated". I have no idea how those are implemented and generated, but the corresponding mechanism is what I meant by "deprecated attributes".

Those were probably so-called "syntactic definitions" (declared using the `Notation` command). I'll make sure we support deprecation on all main constructions (I did it for tactics already), so that we avoid this kind of rough transitions in the future.

Honestly, I'm not sure even developers were aware of this particular deprecation (at least, I wasn't).

ejgallego:
@xavierleroy you are right; we don't have the means to do well but IMHO we are aware of the many shortcomings we have w.r.t. compatibility, some of them are just not easy to fix due to real technical problems / lack of manpower [see for example https://github.com/coq/coq/issues/9664 ] For example we don't even have a unification test suite or a way to version the engine.

So for now we kind of piggyback on the CI testing and "continuous" development, which indeed is time-consuming and confusing for users; but rest assured that we are not satisfied with it.

Zimmi48:
> I'll make sure we support deprecation on all main constructions (I did it for tactics already), so that we avoid this kind of rough transitions in the future.

Thanks for this @maximedenes. See also my requests in #6840 (remove the infrastructure for "compat notations" and replace it with deprecation attributes) and in #8382. This would be two issues that you would be closing by working on this.

If this is easy enough to do in 8.10, then we could consider adding these deprecation warnings for all the definitions and theorems of `Zlogarithm` in 8.10 and merge the present PR only after the branching.

vbgl:
This PR is now ready. May someone please assign and merge? Thanks.

ppedrot:
@herbelin is the natural assignee, but just in case, I can also perform the merge.

herbelin:
Sorry, I'm coming late to the discussion. Was an agreement obtained?

I don't see a warning introduced in 8.10, right? So do we estimate (@xavierleroy included) that the number of incompatibilities in the wild, based on the CompCert CI sample, shall be too few to be worth a (at the current time not so cheap to implement) formal deprecation phase? Or do we add support for deprecating definitions and/or modules (what shall have to be done anyway at some time I guess) and use it for deprecating `Zlogarithm` and `Zsqrt_compat` in 8.10?

(Sorry in advance if I missed discussions which would already have given an answer to the above question.)

ejgallego:
@herbelin indeed I'd guess that the effort of introducing a deprecation phase wouldn't be worth delaying this PR for two releases, fortunately it seems that 8.11 will provide much better deprecation facilities so hopefully this is the last time a situation like this happens.

[by the way I see that math-comp has introduced a way to deprecate lemmas too]

ejgallego:
What's holding this? I propose @vbgl rebases and we merge; indeed the path was not ideal but IMHO the tradeoff is in favor of merge.

vbgl:
I’ve rebased and CI went through.

ejgallego:
> I’ve rebased and CI went through.

Thanks @vbgl , I am assigning and I propose to merge soon unless anyone would like to propose a different plan.

PR #9824, opened 2019-03-24
[doc] Add some pointers on how to start with Coq

ejgallego:
"How do I start using Coq" is a really frequent FAQ these days.

This work in progress tries to address that more prominently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,13 @@ Coq is a formal proof management system. It provides a formal language to write
 mathematical definitions, executable algorithms and theorems together with an
 environment for semi-interactive development of machine-checked proofs.
 
+## New to Coq?
+
+Please see the [new to coq] introductory page for links to tutorial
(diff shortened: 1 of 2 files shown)


Zimmi48:
Thanks for this initiative Emilio, but I'm quite strongly opposed to adding this as a doc file in the sources of Coq. I'd rather see it as one of the following options:
- a new section in the Introduction chapter of the reference manual;
- a new short chapter of the reference manual in its own right;
- a new page on the website;
- a new page on the Coq wiki.

Nonetheless, it makes a lot of sense to link to this new page from the README, given that lots of people stumble first on the GitHub repository these days.

ejgallego:
> * a new short chapter of the reference manual in its own right;

I am OK adding this to the reference manual, the most important bit for me is the README part, as we get a lot of questions "how do I start"

PR #9825, opened 2019-03-25
Deprecate `Refine Instance Mode` option

maximedenes:
This is in view of the 8.10 release, after which we will remove the
option and the `VtUnknown` classification.

See #9530.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -153,6 +153,9 @@ Vernacular commands
 
 - The `Show Script` command has been deprecated.
 
+- Option `Refine Instance Mode` has been deprecated and will be removed in
+  the next version.
+
 Tools
 
 - The `-native-compiler` flag of `coqc` and `coqtop` now takes an argument which can have three values:
--- a/doc/sphinx/addendum/type-classes.rst
+++ b/doc/sphinx/addendum/type-classes.rst
@@ -561,6 +561,8 @@ Settings
 
 .. flag:: Refine Instance Mode
 
+   .. deprecated:: 8.10
+
    This flag allows to switch the behavior of instance declarations made through
    the Instance command.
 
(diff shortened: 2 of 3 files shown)


Zimmi48 (commented):
Please add a `.. deprecated:: 8.10` in the documentation of the flag.

maximedenes:
@Zimmi48 can you be the assignee?

Zimmi48 (approved):

RalfJung:
When deprecating something it would be really useful (e.g. in the release notes) to indicate the intended replacement. I suppose in this case it's `Program Instance`, but I am not sure.

SkySkimmer:
Also https://github.com/coq/coq/issues/10993

RalfJung:
In particular, converting things to `Program` has quite a few pitfalls, that also would be good to have mentioned in the release notes so as to avoid everyone having to rediscover them when they are forced to do this change:
* All `match` should be equipped with `return _` or else `Program` will do unspeakable horrors to them.
* When there are sigma types involved, `Program` will do Weird Stuff (TM), and I do not know a systematic way to translate a normal Coq term into something that works with `Program`. In particular, sigma types seem to be forcefully coerced to their data part, i.e. if they had type `{ x : T  | ... }` before they will have just type `T` after adding `Program`.

SkySkimmer:
That's what https://github.com/coq/coq/pull/10996 is for, although it needs 8.12.
The expected porting path is not Program imo, you can just replace `:= term. Proof. bla.` by `Proof. refine term. bla.`

SkySkimmer:
Althoug if `term` uses the magic `{}` syntax that won't work I guess.

RalfJung:
> The expected porting path is not Program imo, you can just replace := term. Proof. bla. by Proof. refine term. bla.

Interesting. And that will have the same behavior wrt. which part of the term is sealed and which is not?

SkySkimmer:
Not sure what sealed means.

RalfJung:
I mean `Qed.` vs `Defined.`

SkySkimmer:
No difference then. There's no partial Qed.

PR #9826, opened 2019-03-25
[ssr] Two small improvements

vbgl:
1. Use Coqlib in the “abstract” tactic
 2. Give a slightly more detailed error message in the “rewrite” tactic


gares (approved):
thanks!

PR #9830, opened 2019-03-25
[parser] initialization based on Loc.t rather than Loc.source

gares:
In this way one can also set the current offsets in a file,
useful if you are parsing a Coq fragment within a file instead
of a full file starting from the first line.

### DIFF
--- a/parsing/cLexer.ml
+++ b/parsing/cLexer.ml
@@ -383,9 +383,6 @@ let rec string loc ~comm_level bp len s = match Stream.peek s with
      let loc = set_loc_pos loc bp ep in
      err loc Unterminated_string
 
-(* To associate locations to a file name *)
-let current_file = ref Loc.ToplevelInput
-
 (* Utilities for comments in beautify *)
 let comment_begin = ref None
 let comm_loc bp = match !comment_begin with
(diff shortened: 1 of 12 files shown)


gares:
This PR is orthogonal to #9733 but complements the feature it provides. In particular to parse anti quotations and get errors with the right Loc.t, one needs to tell Coq where the text begins.

parsing/cLexer.ml
```diff
@@ -754,9 +750,9 @@ let token_text = function
   | (con, "") -> con
   | (con, prm) -> con ^ " \"" ^ prm ^ "\""
 
-let func next_token cs =
+let func next_token ?loc cs =
   let loct = loct_create () in
-  let cur_loc = ref (Loc.create !current_file 1 0 0 0) in
+  let cur_loc = ref (Option.default Loc.(create ToplevelInput 1 0 0 0) loc) in
```
ejgallego: I suggest you factor the creation of the `1 0 0 0` location to a function in `Loc`.

parsing/pcoq.ml
```diff
@@ -59,7 +59,7 @@ module type S =
 
   type coq_parsable
 
-  val coq_parsable : ?file:Loc.source -> char Stream.t -> coq_parsable
+  val coq_parsable : ?loc:Loc.t -> char Stream.t -> coq_parsable
```
ejgallego: Just an idea, you could keep the `?file` arg and add another optional `?loc` arg as to be more compatbile; likely not worth it.

toplevel/vernac.ml
```diff
@@ -91,7 +91,8 @@ let load_vernac_core ~echo ~check ~interactive ~state file =
   let input_cleanup () = close_in in_chan; Option.iter close_in in_echo in
 
   let in_pa =
-    Pcoq.Parsable.make ~file:(Loc.InFile file) (Stream.of_channel in_chan) in
+    Pcoq.Parsable.make ~loc:(Loc.create (Loc.InFile file) 1 0 0 0)
```
ejgallego: Ditto, use `Loc.initial_from_file`.

vernac/vernacentries.ml
```diff
@@ -2111,7 +2111,8 @@ let vernac_load interp fname =
   let input =
     let longfname = Loadpath.locate_file fname in
     let in_chan = open_utf8_file_in longfname in
-    Pcoq.Parsable.make ~file:(Loc.InFile longfname) (Stream.of_channel in_chan) in
+    Pcoq.Parsable.make ~loc:(Loc.create (Loc.InFile longfname) 1 0 0 0)
```
ejgallego: Ditto, use `Loc.initial_from_file`.

ejgallego (approved):
This is without a doubt very useful; I think the current interface should play well with the state; even if `cLexer` is exposing unfortunately unsafe primitives.

The initial location should be factored to an utility function tho.

I'll let some time for @ppedrot @herbelin to comment, otherwise I'll merge.

parsing/cLexer.ml
gares: done

parsing/pcoq.ml
gares: Given that I've added an API `Loc.initial` with just one argument, the source, I think that creating a loc out of a source is simple enough that this API can just take the loc.

ejgallego: Sounds good then.

gares:
ping

ejgallego:
We got a conflict, please rebase; will merge this WE.

gares:
done.

The conflicts was trivial, I think we could avoid the ping pong between author and assignee in these cases.
I'll try to talk about that at the next WG.

PR #9833, opened 2019-03-26
Improve the backport script.

silene:
It now uses the origin/master branch rather than the master branch, thus
avoiding the need for local merges.

More importantly, it no longer creates a subshell in case of conflicts,
but instead gives control back to the user. Once conflicts are solved, it
suffices to relaunch the script (instead of exiting the subshell). The
reason is that, otherwise, there was no sane way of giving up a backport,
due to the infinite subshell loop.

### DIFF
--- a/dev/tools/backport-pr.sh
+++ b/dev/tools/backport-pr.sh
@@ -30,13 +30,15 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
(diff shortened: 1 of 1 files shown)


Zimmi48 (approved):
Thanks. You fixed what I most disliked about my script. :tada:

Zimmi48:
@silene I put the 8.9.1 milestone in case it was useful to you to backport this, but there is no other reason for doing this, so just do as you wish.

PR #9836, opened 2019-03-26
[schemes] Don't re-declare scheme side-effects that are already there.

ejgallego:
This is an experimental PR as I am not sure I can follow the reasoning
in e1ba72037191b1d4be9de8a0a8fc1faa24eeb12c

Note that in safe_typing we avoid re-declaring effects twice.

This removes a huge number of redeclaration of schemes side-effects that in fact
are already generated such as `eq_ind`.

Anyways we should deprecate the declaration of Schemes on-the-fly, I
don't see the point honestly; just make sure your theory has the right
ones.

Well, going to a more eager declaration scheme could be costly in
terms of size, so OMMV.

TODO: only declare side-effects if the scheme is generated on-the-fly,
add a parameter to the declaration function.

### DIFF
--- a/tactics/ind_tables.ml
+++ b/tactics/ind_tables.ml
@@ -181,10 +181,7 @@ let define_mutual_scheme kind mode names mind =
 
 let find_scheme_on_env_too kind ind =
   let s = String.Map.find kind (Indmap.find ind !scheme_map) in
-  s, Safe_typing.concat_private
-      (Safe_typing.private_con_of_scheme
-            ~kind (Global.safe_env()) [ind, s])
-      Safe_typing.empty_private_constants
+  s, Safe_typing.empty_private_constants
 
 let find_scheme ?(mode=InternalTacticRequest) kind (mind,i as ind) =
   try find_scheme_on_env_too kind ind


gares:
>  I am not sure I can follow the reasoning
> in [e1ba720](https://github.com/coq/coq/commit/e1ba72037191b1d4be9de8a0a8fc1faa24eeb12c)

reading the message of that famous commit I believe the example is
```
Proof.
...
abstract t.
abstract t.
...
Qed.
```
Where `t` is a tactic that generates a lemma by side effect (the same, in both branches)
The commit in question adds to the test suite some proofs ending with `Defined`.
At that time `Defined` was acting like `Qed`, eg inlining effects, IIRC.
Today `Defined` does not (to improve backward compat with some use cases).
So I think you should add a test with the scenario mocked up above in order to be sure the double declaration is not needed (that I think is still needed).

gares:
> Anyways we should deprecate the declaration of Schemes on-the-fly, I
> don't see the point honestly; just make sure your theory has the right
> ones.
> 
> Well, going to a more eager declaration scheme could be costly in
> terms of size, so OMMV.

I don't think so.

Also, "on the fly declared schemes" are looked up by name, IIRC, hence if you declare a scheme in file A and not Import it in B, the scheme is declared again and again. If you declare together with type Fullpath.T you can just look in Fullpath.scheme_T and be done with it (or use coqlib).

So I'm 100% in favor of killing the on-the-fly generation of schemas.

ejgallego:
> Where `t` is a tactic that generates a lemma by side effect (the same, in both branches)

Sure, I'd love to add a test, but after #9838 almost no tactic does generates schemes so it is tricky to add the test case. I have indeed a branch were on-the-fly declaration of schemes is killed, so maybe we should just merge that branch and don't do the test case.

> I think is still needed

Umm, shouldn't the second abstract do indeed have access to the stuff declared on the first?

> So I'm 100% in favor of killing the on-the-fly generation of schemas.

I am too in favor, even if that's a useful feature sometimes, IMO we'd like to implement it a bit differently, and the situation of the code in the scheme things is not good. I have a branch doing this, I just need to fix a bug in find_elim.

> > could be costly in
> > terms of size, so OMMV.
> 
> I don't think so.

That's a bit delicate, for example using `ppx` in opam to add deriving etc... has turned out to be bad in terms of size for binaries, so we should be careful about over generating schemes. Even if that's not a problem now we should IMHO keep it in mind.

gares:
> Umm, shouldn't the second abstract do indeed have access to the stuff declared on the first?

it depends where the effect is placed. At the time I wrote the commit, the effect was hidden inside the first opaque_subproof (generated by the first call to abstract), so not available to the second opaque_subproof.
I suspect that, in the test suite, the effect is hoisted once and forall (by the new semantics of Defined), but if you take the same test and put Qed in place of Defined (that was the same when I wrote the commit) you get to see the problem, making CI fail on this PR.

ejgallego:
> making CI fail on this PR.

Umm, do you mean it would fail at `Qed` time? I don't think so, at the time the second abstract is going to be checked, the scheme should have been properly put in place I think.

I guess we can have an example but it is not easy to cook due to few schemes being really generated.

gares:
I've just tried and you are right it works.
I guess as some point in time effects inside abstract were not propagated to the enclosing constant.
It seems they are now, so I think this PR is fine.

gares (approved):

SkySkimmer:
The PR title says `[kernel]` but only ind_tables is touched?

SkySkimmer:
@gares maybe you can merge without waiting for @ppedrot?

ppedrot:
Go ahead, as long as it looks like it works this is fine.

gares:
I'm fine merging but I don't think the milestone is appropriate since it brings no user visible change.

ejgallego:
I am OK with any milestone; this borders the bug fix tho.

PR #9838, opened 2019-03-27
[scheme side effects] Declare most schemes not as a side effect

ejgallego:
This PR is also an experimental draft and basically separates
declaration of schemes "on the fly" from those being declared as part
of a vernacular command, and thus not inside a proof.

Let's see what the CI says, if successful, we should reify the types
as to eliminate the extra `eff` parameter from many places and make
dynamic and static versions.

Current "real" dynamic effects for the stdlib are:
```
adding seff [scheme]: Coq.Logic.EqdepFacts.internal_eq_rew_r_dep
adding seff [scheme]: Coq.Logic.Hurkens.TypeNeqSmallType.internal_eq_rew_dep
adding seff [scheme]: Coq.setoid_ring.Ring_polynom.internal_True_rew_r
adding seff [scheme]: Coq.micromega.EnvRing.internal_True_rew_r
adding seff [scheme]: Coq.Logic.ChoiceFacts.internal_eq_rew_dep
adding seff [scheme]: Coq.setoid_ring.Field_theory.internal_True_rew_r
adding seff [scheme]: Coq.btauto.Algebra.internal_null_rew
```
Most of them seem to appear due to
`Indschemes.declare_rewriting_schemes` not triggering for the concrete
case at hand, this seems like a bug indeed in `Equality.find_elim`.

### DIFF
--- a/plugins/ltac/rewrite.ml
+++ b/plugins/ltac/rewrite.ml
@@ -958,7 +958,7 @@ let fold_match ?(force=false) env sigma c =
     in 
     let exists = Ind_tables.check_scheme sk ci.ci_ind in
       if exists || force then
-	dep, pred, exists, Ind_tables.find_scheme sk ci.ci_ind
+        dep, pred, exists, Ind_tables.find_scheme ~static:false sk ci.ci_ind
       else raise Not_found
   in
   let app =
--- a/tactics/elimschemes.ml
+++ b/tactics/elimschemes.ml
@@ -31,7 +31,7 @@ let optimize_non_type_induction_scheme kind dep sort _ ind =
     (* in case the inductive has a type elimination, generates only one
        induction scheme, the other ones share the same code with the
        apropriate type *)
-    let cte, eff = find_scheme kind ind in
+    let cte, eff = find_scheme ~static:true kind ind in
     let sigma, cte = Evd.fresh_constant_instance env sigma cte in
     let c = mkConstU cte in
     let t = type_of_constant_in (Global.env()) cte in
--- a/tactics/eqschemes.ml
+++ b/tactics/eqschemes.ml
@@ -247,9 +247,9 @@ let sym_scheme_kind =
 (*                                                                    *)
 (**********************************************************************)
 
-let const_of_scheme kind env ind ctx = 
-  let sym_scheme, eff = (find_scheme kind ind) in
-  let sym, ctx = with_context_set ctx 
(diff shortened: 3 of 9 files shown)


SkySkimmer:
>Most of them seem to appear due to
Indschemes.declare_rewriting_schemes not triggering for the concrete
case at hand, this seems like a bug indeed in Equality.find_elim.

Isn't it because they use the registered `core.eq` which isn't available when the inductives are defined? At least some of them. 
eg for eq itself the registration must be a command after the inductive.

ejgallego:
> Isn't it because they use the registered `core.eq` which isn't available when the inductives are defined? At least some of them.

What part of the code are you thinking about Gaëtan?

It seems to me that `is_inductive_equality` doesn't use the coqlib stuff, right?

SkySkimmer:
For instance the `_rew_r_dep` schemes use `get_coq_eq`, but `declare_rewriting_schemes` wraps with `ignore_error`
I guess we don't set the option to declare rewriting schemes in Init.Logic so it doesn't actually matter

ejgallego:
> For instance the `_rew_r_dep` schemes use `get_coq_eq`, but `declare_rewriting_schemes` wraps with `ignore_error`

Ah I see, thanks! This looks like a bug indeed. In fact that ignore_error is really suspicious! We should ignore it.

> I guess we don't set the option to declare rewriting schemes in Init.Logic so it doesn't actually matter

Indeed, should we? Is the on-the-fly generation worth it? @mattam82 ?

ppedrot:
I'd like to push this PR forward, it looks like a very good improvement. @ejgallego will you have time to work on that?

ejgallego:
> I'd like to push this PR forward, it looks like a very good improvement. @ejgallego will you have time to work on that?

I have been actually working on this quite a bit; however I have stumbled upon some other issues with side-effects which I'd like to address before this; in fact I booked some time in the meeting today to discuss.

I'll submit a couple of PRs soon which IMO should be merged before this; for example I believe the uses of `Safe_typing.push_private_constants` in `Refine` and `Lemmas` are wrong.

Anyways the more I think the more I am convinced we should just go for a more radical approach instead of spending our time in cleaning this likely flawed design.

ppedrot:
Is this PR still relevant? I think that #11269 cleaned part of the mess, and that after it this PR would basically require a full rewriting now?

SkySkimmer:
Please reopen (or open a fresh pr) if progress is made

PR #9842, opened 2019-03-27
[schemes] Don't ignore errors on scheme creation.

ejgallego:
This was likely a temporary measure to help with refactoring, but now
it seems counterproductive to do so.


SkySkimmer:
Not sure how much we test this code.

ejgallego:
> Not sure how much we test this code.

Indeed, but IMHO we should see the errors and properly handle them.

ejgallego:
ah ah
```
File ".\success/Scheme.v", line 26, characters 0-67:
Error: Constructors arguments must repeat the parameters.
```

PR #9849, opened 2019-03-27
[funind] Use new proof engine.

ejgallego:
Right in time for the spring!

Main user of the legacy engine now is ssreflect.

The PR is almost ready [modulo who knows what bugs, this was written quite a while ago so the rebases may have introduced strange stuff], however there are a few nits to solve, and I would enjoy help:

- missing `tclTHEN_i`
- `Recdef.mkDestructEq` had a strange type
- `pf_nf_betaiota` is missing
- `pf_nth_hyp_id`

Question: does the proof engine provide similar facilities to the
pervasive `observe` used here?


ejgallego:
This can be closed in favor of other smaller PRs such as #10525, and #10635

PR #9828, opened 2019-03-25
Fix syntax of Typeclasses eauto := in reference manual.

Zimmi48:
**Kind:** documentation fix

PR #9852, opened 2019-03-28
merge-pr.sh: Remove \r characters returned by jq on Windows

jfehrle:
**Kind:** infrastructure.

When running the script on Windows using cygwin in the Windows `command` window, `jq` returns values that contain carriage return characters `\r` as discussed in https://github.com/coq/coq/pull/9828.  This uses `tr -d "\r"` to remove those characters.

I couldn't figure out the syntax for one jq call--help appreciated.

### DIFF
--- a/dev/tools/merge-pr.sh
+++ b/dev/tools/merge-pr.sh
@@ -73,10 +73,10 @@ fi
 
 PRDATA=$(curl -s "$API/pulls/$PR")
 
-TITLE=$(echo "$PRDATA" | jq -r '.title')
+TITLE=$(echo "$PRDATA" | jq -r '.title' | tr -d "\r")
 info "title for PR $PR is ${BLUE}$TITLE"
 
-BASE_BRANCH=$(echo "$PRDATA" | jq -r '.base.label')
+BASE_BRANCH=$(echo "$PRDATA" | jq -r '.base.label' | tr -d "\r")
 info "PR $PR targets branch ${BLUE}$BASE_BRANCH"
 
 CURRENT_LOCAL_BRANCH=$(git rev-parse --abbrev-ref HEAD)
@@ -143,7 +143,7 @@ fi
 
 # Sanity check: PR has an outdated version of CI
 
-BASE_COMMIT=$(echo "$PRDATA" | jq -r '.base.sha')
+BASE_COMMIT=$(echo "$PRDATA" | jq -r '.base.sha' | tr -d "\r")
 CI_FILES=(".gitlab-ci.yml" "azure-pipelines.yml")
 
 if ! git diff --quiet "$BASE_COMMIT" "$LOCAL_BRANCH_COMMIT" -- "${CI_FILES[@]}"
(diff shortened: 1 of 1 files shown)


gares:
> I couldn't figure out the syntax for one jq call--help appreciated.

I'm not so sure I get what you ask. In case you look for the manual: https://stedolan.github.io/jq/manual/

BTW, are you sure it's `jq` that outputs `\r` and not `echo`?

jfehrle:
> I'm not so sure I get what you ask.

I had trouble figuring where to put `tr -d "\r"` in the line
`[ "$(jq -rc 'map(select(.user.login == "'"$1"'") | .state) | any(. == "'"$2"'")' <<< "$reviews")" = true ]`

I'll leave that fix for the future if I ever run into a problem.  Not so easy testing the script again since I've merged the one PR I intended to merge.

> BTW, are you sure it's jq that outputs \r and not echo?

Probably not echo.

```
$ echo foo|od -bc
0000000 146 157 157 012
          f   o   o  \n
0000004
```

SkySkimmer:
>Not so easy testing the script again since I've merged the one PR I intended to merge.

Just don't push after it merges, pretty easy imo

Zimmi48:
Indeed, merge a random PR, then `git reset --hard upstream/master`. I sometimes do this for testing a PR (maybe a bit dangerous though).

gares:
@jfehrle can you check this related issue? https://github.com/stedolan/jq/issues/363

jfehrle:
> can you check this related issue? stedolan/jq#363

I get return characters in one shell but not in the other.  I'm using the Windows 64 bit version from https://stedolan.github.io/jq/download/.  Very strange.

```
$ echo xxxx$(echo 'true' | jq '.')yy|od -bc
0000000 170 170 170 170 164 162 165 145 015 171 171 012
          x   x   x   x   t   r   u   e  \r   y   y  \n
0000014
```

jfehrle:
OK, updated.  I think it's good to go.

> merge a random PR

actually random, passing CI and reviewed...

FWIW Thinking about getting a new laptop this year with a bigger SSD that would let me install Linux for Windows.  Virtualizing that system or getting a Linux/Windows dual boot setup may be the most flexible, but I expect that would add complexity.

jfehrle:
@maximedenes or @gares, would you review this and approve if agreeable?  Thanks.

maximedenes:
@MSoegtropIMC did you experience similar issues with the merging script on Windows?

MSoegtropIMC:
> @MSoegtropIMC did you experience similar issues with the merging script on Windows?

No. I am using the cygwin version of jq, which is what I recommend here (it is available as cygwin package). IMHO it doesn't make a lot of sense to use a MinGW version of jq here.

Btw.: I picked up my efforts for a script to setup a nice development environment on Windows. I originally wanted to do this with a cygwin compile of coq, but I found that cygwin OCaml is way broken. Also we need to able to fix bugs in the MinGW builds and there is a recent quite well working cross MinGW opam. Coming back to the topic: I would say that this script need only support the dev environment provided by this upcoming script. My current working point is the Merlin/VSCode integration, but I could deliver what I have so far.

One more note: there are a few places where I have to strip \r, but this are places where the build scripts call the tools they build (like coqc -where). Since the goal is to build MinGW executables, it cannot be avoided that coqc is MinGW. But as I said, it doesn't make sense to use MinGW tools in the dev process without need.

maximedenes:
@jfehrle Any reason why you can't use a cygwin-compiled `jq`?

jfehrle:
> Any reason why you can't use a cygwin-compiled `jq`?

I was not aware of the differences between MinGW and Cygwin or that `jg` was available as a Cygwin package.  When I've needed tools I've generally tried to download Windows binaries.  IIRC, none of the websites I've downloaded from mentioned the MinGW vs. Cygwin distinction, so it's not so obvious what they are.  Maybe they're all MinGW.

@MSoegtropIMC, did you get the tools needed to build the documentation as Cygwin packages?

I'll try to look at using the Cygwin `jg` sometime.  For now, I'll just avoid doing any merges--which I wanted to do as a way to help out.  This is, I think, the 4th Windows-specific issue I've run into with the tool chain in the last 6 weeks.  It seems less and less a viable development environment.

MSoegtropIMC:
> @MSoegtropIMC, did you get the tools needed to build the documentation as Cygwin packages?

I usually don't build the documentation, but I won't expect issues. Cygwin has a quite complete tool set.

One major issue with using MinGW tools in the build process is that they don't understand cygwin paths, so one has to be very careful with specifying file names when using a cygwin bash. Cygwin tools can mostly handle both Windows path (if properly quoted or when using / slashes) and cygwin paths.

maximedenes:
> I'll try to look at using the Cygwin `jg` sometime.

Meanwhile, closing. Feel free to reopen if there's a stronger motivation for the change.

Zimmi48:
FWIW even with this patched version of the script, Jim still had trouble using it with cygwin jq. I'm working on supporting merging with @coqbot. This should be helpful for people not wanting / not managing to run the script.

PR #9853, opened 2019-03-28
Use only lowercase for unimath in CI scripts

### LINKED ISSUE #9845: unimath vs UniMath makes create-overlay.sh fail

Reproduce by doing this:
```
./dev/tools/create_overlays.sh Matafou 8347 unimath
```
After compilation (takes a while) the script fails because the directory created is `_build_ci/UniMath` instead of `_build_ci/unimath`.

Solution: either have the name of an "external projects" always match its git name, or make the infrastructure able to compute the name of the directory.

The problem comes from:
[this file](https://github.com/coq/coq/blob/master/dev/tools/create_overlays.sh) at line 69:
    

```
setup_contrib_git $_CONTRIB_DIR $_CONTRIB_GITPUSHURL
```
where `$_CONTRIB_DIR` is incorrect.

#### Coq Version
master

SkySkimmer:
Fix #9845

### DIFF
--- a/dev/ci/ci-basic-overlay.sh
+++ b/dev/ci/ci-basic-overlay.sh
@@ -24,9 +24,9 @@
 ########################################################################
 # UniMath
 ########################################################################
-: "${UniMath_CI_REF:=master}"
-: "${UniMath_CI_GITURL:=https://github.com/UniMath/UniMath}"
-: "${UniMath_CI_ARCHIVEURL:=${UniMath_CI_GITURL}/archive}"
+: "${unimath_CI_REF:=master}"
+: "${unimath_CI_GITURL:=https://github.com/UniMath/UniMath}"
+: "${unimath_CI_ARCHIVEURL:=${unimath_CI_GITURL}/archive}"
 
 ########################################################################
 # Unicoq + Mtac2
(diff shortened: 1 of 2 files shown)


Zimmi48 (approved):

dev/ci/ci-basic-overlay.sh
```diff
@@ -22,11 +22,11 @@
...
 
 ########################################################################
-# UniMath
+# unimath
 ########################################################################
-: "${UniMath_CI_REF:=master}"
-: "${UniMath_CI_GITURL:=https://github.com/UniMath/UniMath}"
-: "${UniMath_CI_ARCHIVEURL:=${UniMath_CI_GITURL}/archive}"
+: "${unimath_CI_REF:=master}"
+: "${unimath_CI_GITURL:=https://github.com/unimath/unimath}"
```
ejgallego: Umm, I'd rather respect the upstream capitalization. Assuming things are case-insensitive is not cool.

SkySkimmer:
As you wish

Gaëtan Gilbert

Matafou:
Hi, it seems Corn suffer the same problem. Maybe doing one PR for both would be better?

SkySkimmer:
There are other issues regarding - vs _, but I don't have motivation to 
go further on this PR so unless someone else does it this is as far as 
we go.

Gaëtan Gilbert

ejgallego (approved):

PR #9855, opened 2019-03-28
[Fail] Simplify `Fail` implementation.

ejgallego:
We can now implement fail in a direct style.

Depends on #10380

### DIFF
--- a/vernac/vernacentries.ml
+++ b/vernac/vernacentries.ml
@@ -2293,23 +2293,26 @@ let vernac_timeout ?timeout (f : 'a -> 'b) (x : 'a) : 'b =
 (* Fail *)
 let test_mode = ref false
 
-(* XXX STATE: the type hints that restoring the state should be the
-   caller's responsibility, see how we restore it in the two
-   branches. *)
-let with_fail ~st f =
+(* Restoring the state is the caller's responsibility *)
+let with_fail f : (Pp.t, unit) result =
   try
     let _ = f () in
-    Vernacstate.invalidate_cache ();
-    Vernacstate.unfreeze_interp_state st;
-    user_err ~hdr:"Fail" (str "The command has not failed!")
+    Error ()
   with
   (* Fail Timeout is a common pattern so we need to support it. *)
   | e when CErrors.noncritical e || e = Timeout ->
     (* The error has to be printed in the failing state *)
-    let msg = CErrors.(iprint ExplainErr.(process_vernac_interp_error (push e))) in
-    (* Restore the previous state; careful here with the cache! *)
-    Vernacstate.invalidate_cache ();
-    Vernacstate.unfreeze_interp_state st;
+    Ok CErrors.(iprint ExplainErr.(process_vernac_interp_error (push e)))
+
+(* We restore the state always *)
+let with_fail ~st f =
+  let res = with_fail f in
+  Vernacstate.invalidate_cache ();
+  Vernacstate.unfreeze_interp_state st;
+  match res with
+  | Error () ->
+    user_err ~hdr:"Fail" (str "The command has not failed!")
+  | Ok msg ->
     if not !Flags.quiet || !test_mode
     then Feedback.msg_info (str "The command has indeed failed with message:" ++ fnl () ++ msg)
 
--- a/vernac/vernacentries.ml
+++ b/vernac/vernacentries.ml
@@ -2291,36 +2291,27 @@ let vernac_timeout ?timeout (f : 'a -> 'b) (x : 'a) : 'b =
     f x
 
 (* Fail *)
-exception HasNotFailed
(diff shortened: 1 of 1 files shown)


maximedenes:
I can't resist: CI failed :)

ejgallego:
Yup I was aware of this case, but I wanted to flush my very large queue of PRs on disk. I'll fix when the dep is merged.

ejgallego:
There is something a bit weird going on here with the bug_3890 which produces an anomaly [@maximedenes this is indeed a bug I've seen on the parsing state PR]

Let me investigate.

ejgallego:
> Let me investigate.

Ok, this is working now, but the whole setup of `bugs/opened` is very strange.

The anomaly is a real bug by the way.

ejgallego:
So this was part of a much larger PR and I spun it off; IMHO it is good for 8.10, but of course up to the assignee / RM.

ejgallego:
Pfff, this needs further cleanup and likely to go hand-in-hand with #9686

ejgallego:
Ok, this is ready but depends on the cleanup of #10380

ejgallego:
Should be ready now.

vernac/vernacentries.ml
```diff
@@ -2291,36 +2291,30 @@ let vernac_timeout ?timeout (f : 'a -> 'b) (x : 'a) : 'b =
...
+    let _ = f () in
+    Error ()
+  with
+  (* Fail Timeout is a common pattern so we need to support it. *)
+  | e when CErrors.noncritical e || e = Timeout ->
+    (* The error has to be printed in the failing state *)
+    Ok CErrors.(iprint ExplainErr.(process_vernac_interp_error (push e)))
+
+(* We restore the state always *)
+let with_fail ~st f =
```
SkySkimmer: Isn't `st` unused?

vernac/vernacentries.ml
```diff
@@ -2291,36 +2291,30 @@ let vernac_timeout ?timeout (f : 'a -> 'b) (x : 'a) : 'b =
...
-      if not !Flags.quiet || !test_mode then Feedback.msg_info
-          (str "The command has indeed failed with message:" ++ fnl () ++ msg)
-    | _ -> assert false
+    let _ = f () in
+    Error ()
+  with
+  (* Fail Timeout is a common pattern so we need to support it. *)
+  | e when CErrors.noncritical e || e = Timeout ->
+    (* The error has to be printed in the failing state *)
+    Ok CErrors.(iprint ExplainErr.(process_vernac_interp_error (push e)))
```
SkySkimmer: What happens if this raises an exception?

vernac/vernacentries.ml
```diff
@@ -2291,36 +2291,30 @@ let vernac_timeout ?timeout (f : 'a -> 'b) (x : 'a) : 'b =
...
 (* Fail *)
-exception HasNotFailed
-exception HasFailed of Pp.t
-
 let test_mode = ref false
 
-(* XXX STATE: this type hints that restoring the state should be the
-   caller's responsibility *)
-let with_fail ~st f =
+(* Restoring the state is the caller's responsibility *)
```
SkySkimmer: This comment confused me considering we immediately shadow with a wrapper to restore state. Maybe it should indicate that this `with_fail` isn't the full implementation of Fail and mention the wrapper.

vernac/vernacentries.ml
ejgallego: An anomaly? Exception printing should not raise, right?

vernac/vernacentries.ml
ejgallego: See the first commit, maybe the second commit is not worth it and we should drop it?

I wanted to hint at a "functional" implementation, but indeed, maybe that just makes things more confusing.

vernac/vernacentries.ml
ejgallego: For the record I did a test and I got:
```
File "./k.v", line 1, characters 0-23:
Error: Anomaly "Uncaught exception Vernacentries.Koo."
```
where `Koo` is a hack so the fail printer raises.

ejgallego: The exact code flow if you are curious is actually than once the printing has raised an exception, the cache will be invalidtated and the document manager will try to process the printing exception.

Usually the document manager won't be able to process this exception [for example a new found] so hence the anomaly will happen, but it could be well the case that I'd like to process the printing exn and turn into an user message.

ejgallego:
Needs assignee; only question IMO is if we want the second commit.

ejgallego:
Ping for assignee.

SkySkimmer (approved):
I do like the second commit. The comment still bugs me somehow but I can't make a precise suggestion so let's move on.

PR #9859, opened 2019-03-28
[dune] Don't regenerate ltac/dune after bootstrapping.

ejgallego:
Once we have the good `plugins/ltac/dune` in place for bootstrapping,
we should not regenerate it. Thanks to @maximedenes for the report.

This fixes `make states` always rebuilding ltac's dependencies.


Zimmi48 (approved):

ejgallego:
Thanks @Zimmi48 !

PR #9861, opened 2019-03-28
[program] Allow evars in type of fixpoints.

### LINKED ISSUE #9163: [Program] Anomaly: grounding a non evar-free term
```coq
Require Import Program.
Program Fixpoint foo m {wf lt m} := foo m.
```

ejgallego:
This is the right thing to do until we refine the program architecture
a bit to use EConstr.

Closes #9163 .

### DIFF
--- a/vernac/comProgramFixpoint.ml
+++ b/vernac/comProgramFixpoint.ml
@@ -232,7 +232,7 @@ let build_wellfounded (recname,pl,n,bl,arityc,body) poly r measure notation =
   let hook = Lemmas.mk_hook (hook sigma) in
   (* XXX: Grounding non-ground terms here... bad bad *)
   let fullcoqc = EConstr.to_constr ~abort_on_undefined_evars:false sigma def in
-  let fullctyp = EConstr.to_constr sigma typ in
+  let fullctyp = EConstr.to_constr ~abort_on_undefined_evars:false sigma typ in
   Obligations.check_evars env sigma;
   let evars, _, evars_def, evars_typ =
     Obligations.eterm_obligations env recname sigma 0 fullcoqc fullctyp


ejgallego:
@ppedrot @SkySkimmer is you PR assigned list empty enough as to accommodate this little PR?

vernac/comProgramFixpoint.ml
```diff
@@ -232,7 +232,7 @@ let build_wellfounded (recname,pl,n,bl,arityc,body) poly r measure notation =
   let hook = Lemmas.mk_hook (hook sigma) in
   (* XXX: Grounding non-ground terms here... bad bad *)
   let fullcoqc = EConstr.to_constr ~abort_on_undefined_evars:false sigma def in
-  let fullctyp = EConstr.to_constr sigma typ in
+  let fullctyp = EConstr.to_constr ~abort_on_undefined_evars:false sigma typ in
   Obligations.check_evars env sigma;
```
SkySkimmer: I wonder if moving this check_evars earlier would solve the bug. I don't know enough about obligations to tell if check_evars can pass while the type still contains evars.

ejgallego: In the example it seems that the type does still contain an unsolved obligation / evar; check evars won't complain in this case so IMO that doesn't work.

ejgallego: I am not an expert of course, but in:
```
let check_evars env evm =
  Evar.Map.iter
    (fun key evi ->
       if Evd.is_obligation_evar evm key then ()
       else
         let (loc,k) = evar_source key evm in
         Pretype_errors.error_unsolvable_implicit ?loc env evm key None)
  (Evd.undefined_map evm)
```
things seem pretty clear that we still have evars.

SkySkimmer (approved):

PR #9864, opened 2019-03-29
Make Opaque/Transparent/Strategy import-time instead of require-time

SkySkimmer:
Problem: the test of `ring` in Ring31 loops (or takes a long time)
unless we import Ring_polynom (probably because of the `Strategy
expand [PEeval].` there).

Let's see what CI breaks.

This enables doing something like
~~~coq
Module Transparentify.
Global Transparent foo.
Global Transparent bar.
(* a bunch of others *)
End Transparentify.

Module Opacify
Global Opaque foo.
Global Opaque bar.
(* etc *)
End Opacify

Import Transparentify.
(* work transparently *)
Import Opacify.
(* work opaquly *)
Import Transparentify.
(* back to transparent *)

(* end of file: modules who require us aren't affected *)
~~~

### DIFF
--- a/tactics/redexpr.ml
+++ b/tactics/redexpr.ml
@@ -126,7 +126,7 @@ type strategy_obj =
 let inStrategy : strategy_obj -> obj =
   declare_object {(default_object "STRATEGY") with
                     cache_function = (fun (_,obj) -> cache_strategy obj);
-                    load_function = (fun _ (_,obj) -> cache_strategy obj);
+                    open_function = (fun i (_,obj) -> if i = 1 then cache_strategy obj);
                     subst_function = subst_strategy;
                     discharge_function = discharge_strategy;
                     classify_function = classify_strategy }
--- a/plugins/setoid_ring/Field_tac.v
+++ b/plugins/setoid_ring/Field_tac.v
@@ -11,6 +11,8 @@
 Require Import Ring_tac BinList Ring_polynom InitialRing.
 Require Export Field_theory.
 
+Strategy expand [PEeval].
+
  (* syntaxification *)
  (* We do not assume that Cst recognizes the rO and rI terms as constants, as *)
  (* the tactic could be used to discriminate occurrences of an opaque *)
--- a/plugins/setoid_ring/Ncring_tac.v
+++ b/plugins/setoid_ring/Ncring_tac.v
@@ -20,6 +20,7 @@ Require Export Ncring.
 Require Import Ncring_polynom.
 Require Import Ncring_initial.
 
+Strategy expand [PEeval].
 
 Set Implicit Arguments.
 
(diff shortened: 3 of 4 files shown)


tactics/redexpr.ml
```diff
@@ -126,7 +126,7 @@ type strategy_obj =
 let inStrategy : strategy_obj -> obj =
   declare_object {(default_object "STRATEGY") with
                     cache_function = (fun (_,obj) -> cache_strategy obj);
-                    load_function = (fun _ (_,obj) -> cache_strategy obj);
+                    open_function = (fun _ (_,obj) -> cache_strategy obj);
```
maximedenes: I believe this is incorrect, you need to check if the integer is `1`. With your current patch, you activate the strategy as soon as one enclosing module is imported (even if not directly the one containing the command).

SkySkimmer: Indeed changed

SkySkimmer:
What do we do wrt compatibility? It's pretty hard to tell how to fix scripts, especially when they just start diverging. It's similar to the require-time coercion issue except here we're in a performance-critical part of the kernel so it might not be so easy to provide a nice warning.

Would it make sense to keep a way to do require-time reduction strategy? Typically it seems like it's sensible like in ring where you define some internal constant and immediately set its reduction strategy, then some other module expects it to just work and shouldn't have to reexport stuff about the internals.
If we keep the require time behaviour we can offer the import time one as a new option without affecting the preexisting ones, ala `Export Set`. Then we have no compat issues.

Developers from the failing developments in CI may be interested:
- compcert failing with an error in some tactic script, cc @xavierleroy 
- stdpp failing by diverging, cc @robbertkrebbers @RalfJung 
- equations failing as some obligation solving is missing (so failure in some tactic script) @mattam82 
(NB: vst is failing in their compcert copy)

ejgallego:
> Would it make sense to keep a way to do require-time reduction strategy?

Maybe this is the moment to try `Import { reduction-hints } From Mod.`

ejgallego:
But well, I see your point with ring. Ummmm....

RalfJung:
I have no idea what the PR title even means, or what is being changed here and why, or how that can make std++ diverge. Could you provide some context?  Do you know which file/proof is divergeing?

JasonGross:
> Would it make sense to keep a way to do require-time reduction strategy?

I think this is a good idea, at least for the sake of compatibility.  Perhaps the most compatible thing would be to offer a setting either about whether or not you auto-import transparency, or whether or not to have a particular declaration follow require.  Alternatively, we could provide a script that does compatibility porting in a heavy way: it would need to do three things:
1. After every `End module.`, re-issue all non-`Local` `Strategy`, `Opaque`, and `Transparent` commands from that module.
2. At the end of every file, re-issue all non-`Local` `Strategy`, `Opaque` and `Transparent` commands from that file inside a module named, say, `OpacityCommandsForCompatWith9864`.  Additionally in these modules, export the corresponding modules from every file required in this one.
3. After every `Require` statement that is not also an `Import`, import the relevant opacity module

Moving forwards, I think the more controlled approach would be to allow setting transparency locally within a tactic, e.g., `with Strategy ... do ring_code`

SkySkimmer:
I don't have time/motivation to figure this out further.

PR #9869, opened 2019-03-29
[CI] Force caching when running test-suite in async mode

maximedenes:
I thought the test-suite infrastructure was always passing
`-async-proofs-cache force`, but in fact it does it only for interactive
tests.

This should speed up the tests quite a bit.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -516,7 +516,7 @@ test-suite:base+async:
   dependencies:
     - build:base
   variables:
-    COQFLAGS: "-async-proofs on"
+    COQFLAGS: "-async-proofs on -async-proofs-cache force"
   allow_failure: true
   only:
     variables:


SkySkimmer:
If it's useful why not set it for the async build too?

maximedenes:
> If it's useful why not set it for the async build too?

I don't expect many occurrences of `Undo` in stdlib, hopefully (please, don't spoil my week-end :) ). So the tradeoff there may be bad (more memory usage). For the test-suite however, it should be clearly better.

SkySkimmer:
Maybe we should remove Undo from the stdlib instead.

maximedenes:
You mean from the test suite?

SkySkimmer:
Yes

maximedenes:
Indeed, I think that should be the long term plan. But this is reasonable meanwhile, I believe.

ejgallego:
> This should speed up the tests quite a bit.

Umm, I thought that this option does affect equally async and non async proofs, doesn't it?

gares:
This option saves all states, as when being used interactively

ejgallego:
> This option saves all states, as when being used interactively

So what is the difference when using async then ? What happens to the states when the option is not set?

gares:
This option is used in the testsuite in sync mode, it was a bug that it was lost in async. There is no change semantically, it just get slower if you Undo, and you get traces with warnings. We were comparing logs to debug the async problems, and we noticed the discrepancy. That is all.

ejgallego:
Yeah the current fix seems to improve async, my question is why it doesn't also improve the other ones.

maximedenes:
> This option is used in the testsuite in sync mode

@gares Really, where?

> my question is why it doesn't also improve the other ones

Good question, but why merge if you didn't get an answer?

gares:
```
gares@ollypat:~/COQ/coq$ git grep cache test-suite/Makefile
test-suite/Makefile:coqtopload := $(coqtop) -top Top -async-proofs-cache force -load-vernac-source
```

maximedenes:
what about `coqc` (non-interactive tests)?

maximedenes:
Hum, what coq version are you looking at?

PR #9871, opened 2019-03-31
CI: add mit-pdos/argosy

tchajed:
Add Argosy, a system for verifying layered storage systems.

Argosy includes an example that extracts to Haskell. The default `make` target runs extraction, but to build the executable the CI machine needs Haskell stack installed. It would also benefit from caching the ~/.stack directory; otherwise all the Haskell dependencies will be re-compiled every build.

**Kind:**  infrastructure.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -553,6 +553,9 @@ validate:quick:
 # Libraries are by convention the projects that depend on Coq
 # but not on its ML API
 
+library:ci-argosy:
+  extends: .ci-template
+
 library:ci-bedrock2:
   extends: .ci-template
 
--- a/Makefile.ci
+++ b/Makefile.ci
@@ -10,6 +10,7 @@
 
 CI_TARGETS= \
     ci-aac_tactics \
+    ci-argosy \
     ci-bedrock2 \
     ci-bignums \
     ci-color \
--- a/dev/ci/ci-argosy.sh
+++ b/dev/ci/ci-argosy.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+ci_dir="$(dirname "$0")"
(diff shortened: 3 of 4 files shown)


Zimmi48:
@tchajed See the Dockerfile in https://github.com/coq/coq/blob/master/dev/ci/docker/bionic_coq/Dockerfile. You can change the Docker image to install haskell and stack and run some commands to populate the stack cache. See also https://github.com/coq/coq/blob/master/dev/ci/README-developers.md#gitlab-and-docker regarding updating the Docker image.

tchajed:
For now I think I'll leave off the Haskell build, but I'll get back to it when I have time. In the meantime at least we'll have something in Coq's CI, and our current project [mit-pdos/armada](https://github.com/mit-pdos/armada) is already pretty well tested by having Iris in CI (we also test on Coq nightly so at least I know within 24h when things break).

ejgallego (approved):
Cool, thanks!

ejgallego:
It didn't take long for you folks to break your own dev: https://gitlab.com/coq/coq/-/jobs/187636064

ppedrot:
@ejgallego isn't that due to a CI race condition after the merge of #8829?

ejgallego:
Indeed it could be, I have no idea why it was broken :)

tchajed:
@ppedrot is right, #8829 breaks Argosy. I've generally treated x.(f) as just another syntax for function application, and Argosy uses it all over the place.

I'll fix it soon.

ejgallego:
I was kind of kidding in case it was not clear, we all know about Murphy's law :)

tchajed:
Actually, to be more specific, I use x.(f) for projection-like functions f. It leads to uniform code that doesn't worry so much about what's in the record versus derived.

PR #9872, opened 2019-03-31
Fix timing diff script to support non-utf8

### LINKED ISSUE #9705: coq-makefile: print-pretty-timed-diff broken

### LINKED ISSUE #9767: Timing diff scripts should support non-utf8

JasonGross:

Fixes / closes #9767
Fixes / closes #9705
On top of #9766
Closes #9766

(Should this be backported to 8.9.1?)


JasonGross:
Who should be assignee on this?

jashug (approved):
The input and output handling changes look OK to me.
I don't see what was changed to support non-ASCII in filenames nor does that seem to be tested.

JasonGross:
Ah, oops.  Yeah, I realized afterwards that the change was actually to support python3 at all, I think.  I'll update the changes

JasonGross:
While I'm at it, I guess I can also test for non-ascii filenames

JasonGross:
> I don't see what was changed to support non-ASCII in filenames nor does that seem to be tested.

It is now tested, and I've updated the CHANGES to be more accurate

Zimmi48:
>(Should this be backported to 8.9.1?)

Seems reasonable to me. FTR, to request backporting to v8.9, you should use the 8.9.1 milestone instead.

ejgallego (approved):
Thanks a lot @JasonGross ; looks pretty straightforward to me so I'll merge tomorrow.

ejgallego:
@JasonGross @andres-erbsen @samuelgruetter bedrock2 is not failing with a missing `hexdump` https://gitlab.com/coq/coq/-/jobs/187636065

Could it be related to this PR?

JasonGross:
The missing `hexdump` is a red herring (it is not fatal due to `|| true `), the real error is:
```
COQC /builds/coq/coq/_build_ci/bedrock2/deps/riscv-coq/src/Spec/MetricPrimitives.v
File "/builds/coq/coq/_build_ci/bedrock2/deps/riscv-coq/src/Spec/MetricPrimitives.v", line 35, characters 41-47:
Error: getMem is not a projection.
```
This is caused by the merge of #8829, which collided in flight with the merge of this PR ; the CI of this PR succeeded because #8829 had not yet been merged, and the CI of #8829 erroneously succeeded because the pipe to iconv that this PR reverted was masking errors in bedrock2 (I guess we forgot to set pipefail in the PR that added iconv).  The solution is to prepare an overlay for bedrock2 for #8829.

ejgallego:
Oh, indeed sorry, thanks for the analysis @JasonGross

PR #9502, opened 2019-02-07
Remove nondeterministic tests

SkySkimmer:
Close #5982 #7388 #7483 #9497

Issue #4429, opened 2015-11-14
eauto with arith: 70x performance regression in Coq 8.5.


coqbot:
Note: the issue was created automatically with bugzilla2github tool

Original bug ID: BZ#4429
From: @asr
Reported version: 8.5
CC:   @JasonGross, @ppedrot, @Zimmi48

See also: [BZ#5555](https://github.com/coq/coq/issues?q=is%3Aissue%20%22Original%20bug%20ID%3A%20BZ%235555%22)

ppedrot:
There is a lot of philosophical stuff going on in this issue, but no clear bug I can identify. This is an old change, the eauto life went on, I think we can close.

PR #9873, opened 2019-03-31
Remove test file with Timeout that failed spuriously.

Zimmi48:
In #9502 it was agreed to remove test files that contain the `Timeout` command and fail spuriously. Here is another one (see failure at https://gitlab.com/coq/coq/-/jobs/187496964).

**Kind:** infrastructure.


maximedenes (changes requested):
I think it is a bad idea to remove such tests. Instead, they should be made reliable.

In this case, it should be easy, just bound the search depth, no?

Zimmi48:
If you bound the search depth, the test isn't testing anything useful anymore. I think these tests were added mostly by @JasonGross and merged without much consideration. We already have infrastructure for complexity test: this should be used instead.

maximedenes:
> If you bound the search depth, the test isn't testing anything useful anymore.

Why? I thought the initial bug report was about depth count having changed semantics.

Zimmi48:
OK maybe then, but we would need to check that the test file would have still caught the bug at that time. No point in keeping test files that don't test anything.

Zimmi48:
Yet another failure: https://gitlab.com/coq/coq/-/jobs/208454712

Zimmi48:
New instance: https://gitlab.com/coq/coq/-/jobs/210749660

SkySkimmer (approved):
Unless someone makes a PR with a better alternative soon (let's say this week) we should get on with it and merge.
I think we rely on the bench much more than tests to avoid perf regressions.

Zimmi48:
Yet another occurrence: https://gitlab.com/coq/coq/-/jobs/214414670

Zimmi48:
And yet another one! https://gitlab.com/coq/coq/-/jobs/214658761

SkySkimmer:
As announced, let's merge. @maximedenes if you still want to have a test for #4429 please open a PR with one, or if you don't have time you can make an issue.

PR #9876, opened 2019-04-01
Command-line setters for options

### LINKED ISSUE #9295: Project level option system
Currently if you want to set some options throughout your project (eg Primitive Projections) you either put `Set Primitive Projections.` at the top of every file, or put `Export Set Primitive Projections.` in some options file and `Require Import myoptions.` at the top of all other files. This is a pain and liable to lead to forgetting to do it for new files. You can also `Global Set` in your options file and require it whenever it isn't already transitively required but this is messy especially if you want people to use you as a library.
Alternatively you could put `-arg -l -arg myoptions.v` in your _CoqProject, but coq_makefile won't register the dependency and there are probably random things breaking.

Meanwhile if we want to set an option for the stdlib without setting it for everyone else we have no way to do it really (or maybe the -l hack would work).

Can we come up with some better system? Preferrably something that we can get to work with both coq_makefile and the upcoming dune build systems.
cf #8052, [some WG discussion about _CoqProject format](https://github.com/coq/coq/wiki/CoqWG-2018-12-19)

SkySkimmer:
Draft until documented

TODO coqproject handling (for now it can be done through -arg I guess)

Closes #9295.

### DIFF
--- a/doc/sphinx/practical-tools/coq-commands.rst
+++ b/doc/sphinx/practical-tools/coq-commands.rst
@@ -210,6 +210,13 @@ and ``coqtop``, unless stated otherwise:
   is intended to be used as a linter for developments that want to be robust to
   changes in the auto-generated name scheme. The options are provided to
   facilitate tracking down problems.
+:-set *string*: Enable flags and set options. *string* should be
+   ``Option Name=value``, the value is interpreted according to the
+   type of the option. For flags ``Option Name`` is equivalent to
+   ``Option Name=true``. For instance ``-set "Universe Polymorphism"``
+   will enable :flag:`Universe Polymorphism`. Note that the quotes are
+   shell syntax, Coq does not see them.
+:-unset *string*: As ``-set`` but used to disable options and flags.
 :-compat *version*: Attempt to maintain some backward-compatibility
   with a previous version.
 :-dump-glob *file*: Dump references for global names in file *file*
--- a/library/goptions.mli
+++ b/library/goptions.mli
@@ -172,6 +172,14 @@ type option_value =
   | StringValue of string
   | StringOptValue of string option
 
+val set_option_value : ?locality:option_locality ->
+  ('a -> option_value -> option_value) -> option_name -> 'a -> unit
+(** [set_option_value ?locality f name v] sets [name] to the result of
+    applying [f] to [v] and [name]'s current value. Use for behaviour
+    depending on the type of the option, eg erroring when ['a] doesn't
+    match it. Changing the type will result in errors later so don't do
+    that. *)
+
 (** Summary of an option status *)
 type option_state = {
   opt_depr  : bool;
--- a/toplevel/coqargs.ml
+++ b/toplevel/coqargs.ml
@@ -38,6 +38,8 @@ type color = [`ON | `AUTO | `OFF]
 
 type native_compiler = NativeOff | NativeOn of { ondemand : bool }
 
+type option_command = OptionSet of string option | OptionUnset
+
 type t = {
 
   load_init   : bool;
@@ -62,6 +64,8 @@ type t = {
   native_compiler : native_compiler;
   allow_sprop : bool;
   cumulative_sprop : bool;
+
+  set_options : (Goptions.option_name * option_command) list;
 
   stm_flags   : Stm.AsyncOpts.stm_opt;
   debug       : bool;
@@ -114,6 +118,8 @@ let default = {
   native_compiler = default_native;
   allow_sprop = false;
   cumulative_sprop = false;
+
+  set_options = [];
 
   stm_flags    = Stm.AsyncOpts.default_opts;
   debug        = false;
@@ -244,6 +250,16 @@ let get_native_name s =
     String.concat "/" [Filename.dirname s;
       Nativelib.output_dir; Library.native_name_from_filename s]
   with _ -> ""
+
+let to_opt_key = Str.(split (regexp " +"))
+
+let parse_option_set opt =
+  match String.index_opt opt '=' with
+  | None -> to_opt_key opt, None
+  | Some eqi ->
+    let len = String.length opt in
+    let v = String.sub opt (eqi+1) (len - eqi - 1) in
+    to_opt_key (String.sub opt 0 eqi), Some v
 
 (*s Parsing of the command line.
     We no longer use [Arg.parse], in order to use share [Usage.print_usage]
@@ -449,6 +465,16 @@ let parse_args ~help ~init arglist : t * string list =
           error_wrong_arg ("Error: (yes|no|ondemand) expected after option -native-compiler")
       in
       { oval with native_compiler }
+
+    | "-set" ->
+      let opt = next() in
+      let opt, v = parse_option_set opt in
+      { oval with set_options = (opt, OptionSet v) :: oval.set_options }
+
+    | "-unset" ->
+      let opt = next() in
+      let opt = to_opt_key opt in
+      { oval with set_options = (opt, OptionUnset) :: oval.set_options }
 
     (* Options with zero arg *)
     |"-async-queries-always-delegate"
--- a/toplevel/coqtop.ml
+++ b/toplevel/coqtop.ml
@@ -50,6 +50,41 @@ let print_memory_stat () =
 
 let _ = at_exit print_memory_stat
 
+let interp_set_option opt v old =
+  let open Goptions in
+  let err expect =
+    let opt = String.concat " " opt in
+    let got = v in (* avoid colliding with Pp.v *)
+    CErrors.user_err
+      Pp.(str "-set: " ++ str opt ++
+          str" expects " ++ str expect ++
+          str" but got " ++ str got)
+  in
+  match old with
+  | BoolValue _ ->
+    let v = match String.trim v with
+      | "true" -> true
+      | "false" | "" -> false
+      | _ -> err "a boolean"
+    in
+    BoolValue v
+  | IntValue _ ->
+    let v = String.trim v in
+    let v = match int_of_string_opt v with
+      | Some _ as v -> v
+      | None -> if v = "" then None else err "an int"
+    in
+    IntValue v
+  | StringValue _ -> StringValue v
+  | StringOptValue _ -> StringOptValue (Some v)
+
+let set_option = let open Goptions in function
+  | opt, OptionUnset -> unset_option_value_gen ~locality:OptLocal opt
+  | opt, OptionSet None -> set_bool_option_value_gen ~locality:OptLocal opt true
+  | opt, OptionSet (Some v) -> set_option_value ~locality:OptLocal (interp_set_option opt) opt v
+
+let set_options = List.iter set_option
+
 (******************************************************************************)
 (* Input/Output State                                                         *)
 (******************************************************************************)
@@ -194,6 +229,8 @@ let init_toplevel ~help ~init custom_init arglist =
   Global.set_native_compiler (match opts.native_compiler with NativeOff -> false | NativeOn _ -> true);
   Global.set_allow_sprop opts.allow_sprop;
   if opts.cumulative_sprop then Global.make_sprop_cumulative ();
+
+  set_options opts.set_options;
 
   (* Allow the user to load an arbitrary state here *)
   inputstate opts;
--- a/toplevel/usage.ml
+++ b/toplevel/usage.ml
@@ -74,6 +74,9 @@ let print_usage_common co command =
 \n  -indices-matter        levels of indices (and nonuniform parameters) contribute to the level of inductives\
 \n  -type-in-type          disable universe consistency checking\
 \n  -mangle-names x        mangle auto-generated names using prefix x\
+\n  -set \"Foo Bar\"       enable Foo Bar (as Set Foo Bar. in a file)\
+\n  -set \"Foo Bar=value\" set Foo Bar to value (value is interpreted according to Foo Bar's type)\
+\n  -unset \"Foo Bar\"     disable Foo Bar (as Unset Foo Bar. in a file)\
 \n  -time                  display the time taken by each command\
 \n  -profile-ltac          display the time taken by each (sub)tactic\
 \n  -m, --memory           display total heap size at program exit\
--- a/ide/idetop.ml
+++ b/ide/idetop.ml
@@ -57,9 +57,9 @@ let coqide_known_option table = List.mem table [
   ["Diffs"]]
 
 let is_known_option cmd = match Vernacprop.under_control cmd with
-  | VernacSetOption (_, o, BoolValue true)
-  | VernacSetOption (_, o, StringValue _)
-  | VernacUnsetOption (_, o) -> coqide_known_option o
+  | VernacSetOption (_, o, OptionSetTrue)
+  | VernacSetOption (_, o, OptionSetString _)
+  | VernacSetOption (_, o, OptionUnset) -> coqide_known_option o
   | _ -> false
 
 (** Check whether a command is forbidden in the IDE *)
@@ -366,12 +366,13 @@ let get_options () =
   Goptions.OptionMap.fold fold table []
 
 let set_options options =
+  let open Goptions in
   let iter (name, value) = match import_option_value value with
-  | BoolValue b -> Goptions.set_bool_option_value name b
-  | IntValue i -> Goptions.set_int_option_value name i
-  | StringValue s -> Goptions.set_string_option_value name s
-  | StringOptValue (Some s) -> Goptions.set_string_option_value name s
-  | StringOptValue None -> Goptions.unset_option_value_gen name
+  | BoolValue b -> set_bool_option_value name b
+  | IntValue i -> set_int_option_value name i
+  | StringValue s -> set_string_option_value name s
+  | StringOptValue (Some s) -> set_string_option_value name s
+  | StringOptValue None -> unset_option_value_gen name
   in
   List.iter iter options
 
--- a/plugins/ssr/ssrvernac.mlg
+++ b/plugins/ssr/ssrvernac.mlg
@@ -183,7 +183,7 @@ GRAMMAR EXTEND Gram
   GLOBAL: gallina_ext;
   gallina_ext:
    [ [ IDENT "Import"; IDENT "Prenex"; IDENT "Implicits" ->
-      { Vernacexpr.VernacUnsetOption (false, ["Printing"; "Implicit"; "Defensive"]) }
+      { Vernacexpr.VernacSetOption (false, ["Printing"; "Implicit"; "Defensive"], Vernacexpr.OptionUnset) }
    ] ]
   ;
 END
--- a/stm/asyncTaskQueue.ml
+++ b/stm/asyncTaskQueue.ml
@@ -139,7 +139,7 @@ module Make(T : Task) () = struct
         (* We need to pass some options with one argument *)
         | ( "-I" | "-include" | "-top" | "-topfile" | "-coqlib" | "-exclude-dir" | "-compat"
           | "-load-ml-object" | "-load-ml-source" | "-require" | "-w" | "-color" | "-init-file"
-          | "-profile-ltac-cutoff" | "-main-channel" | "-control-channel" | "-mangle-names"
+          | "-profile-ltac-cutoff" | "-main-channel" | "-control-channel" | "-mangle-names" | "-set" | "-unset"
           | "-diffs" | "-mangle-name" | "-dump-glob" | "-bytecode-compiler" | "-native-compiler" as x) :: a :: tl ->
           x :: a :: set_slave_opt tl
         (* We need to pass some options with two arguments *)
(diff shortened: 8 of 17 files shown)


ejgallego (commented):
I am bit wary of stuff like this because I do believe that the whole option system should rewritten instead of keep adding some more kludges such as this one that may end up being technical debt in the end. For example doing #8666 is not possible with the current system.

But if people want that I am OK with this patch. You need however:
- to add documentation,
- to check that the command handling stuff work correctly w.r.t. workers (see asynctaskqueue)

ejgallego:
So indeed maybe @SkySkimmer if you think this shouldn't go in 8.10 then I'd suggest we chat a bit about the Options API then we go ahead to merge this.

SkySkimmer:
>For example doing #8666 is not possible with the current system.

I don't understand.

>to check that the command handling stuff work correctly w.r.t. workers (see asynctaskqueue)

How do I do that?

ejgallego:
> I don't understand.

The current API doesn't allow atomic setup of several options and that is a big problem for use cases such as the one in the PR [printing all has to become a bad hack]

> How do I do that?

Good question, there is some filtering there, so I guess you should check that the options are correclty passed to the worker processes with `ps ax` or such.

SkySkimmer:
> Good question, there is some filtering there, so I guess you should
 > check that the options are correclty passed to the worker processes with
 > |ps ax| or such.

no need to ps, just launching coqide with a -set caused it to die ;) 
fixed now

Gaëtan Gilbert

Zimmi48:
Redesigning the whole option system would be definitely welcome (if someone has ideas, please start a CEP). However, in the meantime there are lots of low-hanging fruits (and TBH the option system works better than one could expect given the heavy use that we make of it).

In particular, this PR is very welcome, as are all incremental improvements until someone has a better proposal.

ejgallego (commented):
Preliminary review: this is looking good, I don't have any comment; I assigned to myself.

I will do in-depth review when the PR goes out of draft status.

SkySkimmer:
Added doc.
The doc for `-set-to` is quite verbose due to the complex behaviour, maybe it would be better to have separate `-set-int` and `-set-string` so that we don't have to implement any quoting behaviour.

ejgallego:
> The doc for `-set-to` is quite verbose due to the complex behaviour, maybe it would be better to have separate `-set-int` and `-set-string` so that we don't have to implement any quoting behaviour.

Can't you delay the decision on the type after parsing the option and querying for its type?

I'd rather have a more uniform syntax by the way, what do you think of just:
- `-set "Foo Bar=true"`
- `-set "Foo Bar=false"`
- `-set "Foo Ba=2"` ,etc...

ejgallego:
We need a choice on the milestone tho; this is the kind of change I'd even backport, but pinging @vbgl @Zimmi48 to see what you think.

Zimmi48:
I agree with you: having `-set-int` and `-set-string` would avoid any confusion and would make the feature easier to use for this very special case when you want to pass `"1"` as the value of your option.

SkySkimmer:
Why's appveyor complaining? And where's azure?

Zimmi48:
I hadn't seen @ejgallego's comment above.

>Can't you delay the decision on the type after parsing the option and querying for its type?

That sounds like a good solution too if it's not too heavy to implement.

I don't have any clear opinion on proposing integration in 8.10.

ejgallego:
> That sounds like a good solution too if it's not too heavy to implement.

Indeed it is easy for tools interacting with coqtop to expose single interface `key=value` so they don't have to be aware of types.

ejgallego:
> Indeed it is more convenient for tools interacting with coqtop to expose single interface `key=value` so they don't have to be aware of types.

Like imagine for example a project configuration in `yaml`, etc... this way you could write:
```yaml
coq_options:
  printing all: true
  unification keys: active
```
etc...

ejgallego (commented):
The PR looks good to me, however I do have a request:

- change the API to be of the form `-set "option=value"`

This would amount to:
- modifying the parser
- querying goptions for type after parsing

I strongly recommend we do this modification, unless @SkySkimmer opposes; the motivation is that I'd like to be able to map `key: value` stores to Coq options directly. IMO this is an important use case.

WDYT @SkySkimmer ? I can help pushing to be branch if you are short on time for this.

SkySkimmer:
OK
Do we keep -unset? If not there's ambiguity for stringopt options (currently only 2 users Default Proof Using and Dump Arith (from micromega)).

SkySkimmer:
Changed.

SkySkimmer:
The first commit is now unused for the purpose of this PR as the command line handling doesn't go through vernacentries, but I still like so leaving it as a rider.

Zimmi48 (approved):
Can you add a mention of this feature in the section about flags and options in the vernac chapter (with a link)?

ejgallego:
Thanks @SkySkimmer , will review ASAP.

> Do we keep -unset? If not there's ambiguity for stringopt options (currently only 2 users Default Proof Using and Dump Arith (from micromega)).

I'd say no point in keeping `-unset` neither the `stringopt` kind.

Zimmi48:
> I'd say no point in keeping `-unset` neither the `stringopt` kind.

:-1: you talk too fast Emilio, the integer and boolean options supporting `Unset` do make lots of sense. I'll try to give you some examples.

Zimmi48:
You can map this to your key-value model with value `null` BTW.

maximedenes:
> You can map this to your key-value model with value `null` BTW.

Yeah, `null` values are a well known example of a brilliant concept in computer science :)

ejgallego:
>  you talk too fast Emilio?

Well as much as I am a fan of intuitionistic type theory I'd rather keep boolean options bi-valued.

> the integer and boolean options supporting `Unset` do make lots of sense. I'll try to give you some examples.

Sure, I'm looking forward to it.

Zimmi48:
> Yeah, `null` values are a well known example of a brilliant concept in computer science :)

I think you are confusing with null pointers, are you not? Here we are talking about options, which are quite a fundamental concept of functional programming that actually allows better modeling.



> Well as much as I am a fan of intuitionistic type theory I'd rather keep boolean options bi-valued.

But we are not talking about boolean options (a.k.a. flags) here.

ejgallego:
> But we are not talking about boolean options (a.k.a. flags) here.

You wrote: "the integer and boolean options supporting Unset do make lots of sense"

So I am looking forward for examples, because I am quite convinced of the opposite.

maximedenes:
> I think you are confusing with null pointers, are you not?

Hum, no, thank you, I did mean `null` *values*. Like @ejgallego, I like when my booleans have only two values, my unit type only one inhabitant etc... But I may be old-fashioned.

ejgallego:
By the way you can have unset work for integers using `-set "foo="` and similarly `foo: ` in YAML-like languages, strings are the ones that create trouble here and IMHO they are too fringe of a case as to warrant yet another option. Note that Gaëtan's proposal didn't mention removing `int option` so maybe you read too fast.

In fact the number of bugs that "null" in environment variables has created is on par with its C counterpart. What's the difference between `FOO` being unset and `FOO` being set to `""`? Sounds like fun XD

ejgallego:
As not to make @SkySkimmer 's life bad, I've submitted https://github.com/coq/coq/pull/9953 to continue discussion over string option option.

toplevel/coqargs.ml
```diff
@@ -38,6 +38,8 @@ type color = [`ON | `AUTO | `OFF]
 
 type native_compiler = NativeOff | NativeOn of { ondemand : bool }
 
+type option_set = OptionSet of string option | OptionUnset
```
gares: This is a bad name, it reads "a set of options" to me. Maybe "option_status"?

toplevel/coqtop.ml
```diff
@@ -50,6 +50,40 @@ let print_memory_stat () =
...
+    let opt = String.concat " " opt in
+    let got = v in (* avoid colliding with Pp.v *)
+    CErrors.user_err
+      Pp.(str "-set: " ++ str opt ++
+          str" expects " ++ str expect ++
+          str" but got " ++ str got)
+  in
+  match old with
+  | BoolValue _ ->
+    let v = match v with
```
gares: In the parsing you allow `-set "This Option =true"` but not `-set "This Option = true"`.
IMO you should trim white spaces for all non-string options.

gares: (and check if int_of_string does that)

gares (approved):
I've some remarks, but none is blocking.

gares:
>  `null` _values_.

I know languages where `null` has type `null` and no over value has type `null`.
If you are dynamically typed (as this code) then `map[key] = null` is a perfectly sound way of expressing that the binding of `key` has to be removed (since no value in the map is allowed to be of type `null`).

toplevel/coqargs.ml
```diff
@@ -245,6 +251,19 @@ let get_native_name s =
       Nativelib.output_dir; Library.native_name_from_filename s]
   with _ -> ""
 
+let to_opt_key opt =
+  let opt = String.split_on_char ' ' opt in
+  let opt = List.filter (fun s -> s <> "") opt in (* allow the user to duplicate spaces *)
```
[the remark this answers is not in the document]
SkySkimmer: Didn't know about Str, thanks.

toplevel/coqargs.ml
SkySkimmer: Changed to `option_command`.

toplevel/coqtop.ml
SkySkimmer: done
int_of_string doesn't trim.

library/goptions.mli
```diff
@@ -172,6 +172,14 @@ type option_value =
   | StringValue of string
   | StringOptValue of string option
 
+val set_option_value : ?locality:option_locality ->
+  ('a -> option_value -> option_value) -> option_name -> 'a -> unit
```
ejgallego: I'd prefer if the return type of this function was `(unit, Pp.t) result` now that you are exporting it, but we can live with the current version if you don't feel like modifying it.

toplevel/usage.ml
```diff
@@ -74,6 +74,9 @@ let print_usage_common co command =
 \n  -indices-matter        levels of indices (and nonuniform parameters) contribute to the level of inductives\
 \n  -type-in-type          disable universe consistency checking\
 \n  -mangle-names x        mangle auto-generated names using prefix x\
+\n  -set \"Foo Bar\"       enable Foo Bar (as Set Foo Bar. in a file)\
+\n  -set \"Foo Bar=value\" set Foo Bar to value (value is interpreted according to Foo Bar's type)\
+\n  -unset \"Foo Bar\"     disable Foo Bar (as Unset Foo Bar. in a file)\
```
ejgallego: IMHO having `-unset` is not worth it; the type in `CoqArgs` would fit better if instead we have:
```
-set "Foo Bar="   => return None in coqargs
-set "Foo Bar=true"
```
For boolean options just have the user do `"Foo Bar=true/false"`.

ejgallego (commented):
Things look good however I dislike the special status of boolean options; they should be treated as regular options taking `true/false` or `on/off`.

`-unset` seems also redudant then and can be codified as `-set "Foo="`

library/goptions.mli
SkySkimmer: >I'd prefer if the return type of this function was (unit, Pp.t) result now that you are exporting it

Not seeing the point. Note the exception is raised in the callback, not by set_value itself.

SkySkimmer:
-unset is needed for stringopt as explained previously. Also I like having the shorthand forms.

SkySkimmer:
Timeout in romega3 in the nix job https://gitlab.com/coq/coq/-/jobs/195847533
Is this known unreliable? cc @Zimmi48

library/goptions.mli
ejgallego: Indeed you'd need to modify the callback too; anyways that's just "do if you want", but the mess with functions throwing exceptions is real.

ejgallego:
> -unset is needed for stringopt as explained previously. Also I like having the shorthand forms.

The shorthand forms are a disaster for outsider tooling as now the tooling has to learn the type of the options and cannot do
```
foo: true
bar: 3
```
Keeping unset just for that fringe case seems not worth the tradeoff, let's have `foo=` as unset.

SkySkimmer:
`foo: true` works perfectly fine though

ejgallego:
> `foo: true` works perfectly fine though

Oh indeed; so only question then is what to do with unset:
- IMHO we should at least accept `-set "Foo="` as unset;
- the type of `Coqargs.option_command` seems indeed too general [None should mean unset IMO].
- I would get rid of `-unset`, however this is not a hard request and I won't oppose merge if you would like to keep it.

ejgallego:
Regardless of the last 3 non-blocking suggestions, the PR seems ready; as the maintainer of the subsystem I'm suggesting this PR to target the 8.10 milestone due to its usefulness / risk ratio; but of course that ultimately boils to the RM.

doc/sphinx/practical-tools/coq-commands.rst
```diff
@@ -210,6 +210,13 @@ and ``coqtop``, unless stated otherwise:
   is intended to be used as a linter for developments that want to be robust to
   changes in the auto-generated name scheme. The options are provided to
   facilitate tracking down problems.
+:-set *string*: Enable flags and set options. *string* should be
+   ``Option Name=value``, the value is interpreted accoring to the
```
Zimmi48:
```suggestion
   ``Option Name=value``, the value is interpreted according to the
```

Zimmi48:
I think there is a confusion around what `Unset` means. For flags, it means set to false, which is not necessarily the default. For integers, it means set to the default, which can be "undefined" or can be some number. For strings, it can be unsupported (most common case) or it can be supported for some options that want to have an undefined case. But it would actually make sense to merge the two cases by following the lead of integer options and always have `Unset` reset to the default value.

Anyway, that's completely out of the scope of this PR, and out of the scope of 8.10 too.

>Timeout in romega3 in the nix job https://gitlab.com/coq/coq/-/jobs/195847533
>Is this known unreliable? cc @Zimmi48

I've seen it fail before but I haven't investigated that yet.

SkySkimmer:
>For integers, it means set to the default

I don't think that's true at least in principle? In practice it might be.
For instance it would make sense for `Unset Printing Depth` to mean unlimited printing depth even though the default isn't, otoh currently it means reset to default.

Zimmi48:
> I don't think that's true at least in principle? In practice it might be.
> For instance it would make sense for `Unset Printing Depth` to mean unlimited printing depth even though the default isn't, otoh currently it means reset to default.

This is what is documented *and* that's how all integer options work today. I'm not saying that other solutions wouldn't make sense. I'm just trying to clear the confusion around the current situation.

ejgallego:
> I think there is a confusion around what `Unset` means

Indeed but I think more that `Unset` is confusing in itself; the fact that what it does depends on the type is already a warning signal. IMHO it only makes sense for boolean options.

SkySkimmer:
In any case I'm not looking to change the semantics of v files for this PR.

ejgallego:
> In any case I'm not looking to change the semantics of v files for this PR.

@SkySkimmer I am waiting for an answer to https://github.com/coq/coq/pull/9876#issuecomment-483040321

SkySkimmer:
>IMHO we should at least accept -set "Foo=" as unset;

Done for bool/int options

>the type of Coqargs.option_command seems indeed too general [None should mean unset IMO].

It looks fine to me.

I'm keeping -unset

ejgallego:
@SkySkimmer  in:
```
type option_command = OptionSet of string option | OptionUnset
```
`OptionSet None` is just a special case for `OptionSet (Some "true")` IMO, you could remove the option type and just have parsing return `true`.

ejgallego:
The change is pretty trivial, I've just done it.

Zimmi48:
`-set "Foo="` for unset sounds more reasonable for options than for flags IMHO. I think this is related to the current confusion around the `Unset` command that has a different meaning for options and flags (set to default vs set to false). Overall, I'm wondering if we are not rushing things by wanting this in 8.10. If we were to delay this to 8.11, and still merge early, it would leave time for everyone to check if they are satisfied with the chosen solution.

SkySkimmer:
But then `-set foo` when foo is a string option means `-set foo=true` which is surprising to me, I prefer it being a type error.

ejgallego (approved):
LGTM, this is ready for merge an IMHO has a benefit/risk ratio high enough as to warrant inclusion in 8.10.

Semantics of `UnSet` and some other nuances with options predate this PR, it is a bit unfortunate we have to expose `-unset` in the command line as it will likely lead users towards `-unset Foo` instead of `-set Foo=false` which is IMO by far the preferred form.

This means this PR does introduce some technical debt at the cmdline level which is one of the worst parts of the system in terms of impact. I would still prefer if we could at least discourage the use of `-unset` in the documentation or restrict it somehow.

The note about improving the type of `option_command` does still apply indeed; I see no reason we should not tweak it before merge.

ejgallego:
> But then `-set foo` when foo is a string option means `-set foo=true` which is surprising to me, I prefer it being a type error.

Good point, no change then if you prefer. But for me this is indeed a warning that the shortcut `-set Foo` brings more ambiguity than benefits; I'm not sure allowing to omit the `=` sign is worth it.

Leaving a few hours more before merging so discussion can close.

ejgallego:
> Overall, I'm wondering if we are not rushing things by wanting this in 8.10. If we were to delay this to 8.11, and still merge early, it would leave time for everyone to check if they are satisfied with the chosen solution.

No problem in delaying this Theo if you think we should discuss more.

I think the first problem is that we don't agree on the distinction between "options" and "flags", I don't see why they are different; maybe I am influenced by my current choice of cmline handling in other projects. I dunno.

ejgallego:
> it would leave time for everyone to check if they are satisfied with the chosen solution.

I am personally OK with the current solution; I'd have chosen myself a more "regular" solution likely, but at some point stuff like this comes down to taste, and [when you write the code, you have a fair amount of power](https://twitter.com/btaylor/status/1099370172916195328) XD so I guess at least myself I'm not willing to delay this more [otherwise at some point we will just halt all development]

Zimmi48:
Not that I am not arguing to delay the merge. I'm perfectly OK with merging now. I'm just arguing to not squeeze this feature into 8.10, which seems like the normal thing to do when discussion has revealed that things are not that simple.

Zimmi48:
> I think the first problem is that we don't agree on the distinction between "options" and "flags", I don't see why they are different

Maybe, they should not be different. But they currently are (see documentation: https://coq.github.io/doc/master/refman/proof-engine/vernacular-commands.html#flags-options-and-tables).

ejgallego:
> I'm just arguing to not squeeze this feature into 8.10, which seems like the normal thing to do when discussion has revealed that things are not that simple.

Well, modulo a few nits, the current implementation seems reasonable to me; in the sense that semantics of the command line flags should match the ones of the vernaculars, is the current `-set Foo=` behavior a concern for you? My reasoning was to map to to the key/value system when `Foo: `  appears.

If so we could revert.

ejgallego:
I have been thinking about the milestone of this PR, and indeed it seems a bit complicated on the basis of what @Zimmi48 says:
- indeed, putting in to 8.10 seems a bit precipitate as we'd like maybe to think a bit more about what we do. On the other hand we seem to have reached a kind of fixpoint in the iterations, I'm not sure what more we'd like to change.
- on the other hand, putting this PR in 8.10 would greatly help us to get feedback from users, and thus help towards a more comprehensive re-design of the option / flag system.

So I dunno; I guess for me the best compromise in this case is to ship in 8.10, but declare the feature "experimental" , that is to say, 8.10 adopters should treat it like in beta / unstable, and thus only use it if they feel comfortable with having to tweak scripts in 8.11; on the other hand that's allow us to get welcome feedback and testing.

Zimmi48:
Sounds good to me.

ejgallego:
> Sounds good to me.

So is there some action to do on the docs then?

@SkySkimmer a last nit seems whether the non-uniformity of `-set Foo` vs `-set Foo=true` is worth the gain. That is of course a non-blocker for the PR but maybe we'd like to think twice.

IMVHO I'd at least add a note to the docs stating that `-set Foo=true` is preferred [and more future-proof]

Zimmi48:
>So is there some action to do on the docs then?

Well, at least the mention of the experimental status, no? Maybe also the link from the vernac section on options to the command-line flags section.

> @SkySkimmer a last nit seems whether the non-uniformity of `-set Foo` vs `-set Foo=true` is worth the gain. That is of course a non-blocker for the PR but maybe we'd like to think twice.
> 
> IMVHO I'd at least add a note to the docs stating that `-set Foo=true` is preferred [and more future-proof]

I disagree... The doc could say that `-set Foo` is more human-oriented and `-set Foo=true` is more machine-oriented.

ejgallego:
> I disagree... The doc could say that `-set Foo` is more human-oriented and `-set Foo=true` is more machine-oriented.

That's another choice; but it not so much about "human vs machine" IMHO, more about how data formats these days works, for example JSON or YAML.

ejgallego:
Leaving the question of the milestone and status of the PR open, IMHO no more benefit from delaying the actual merge; these issues could be addressed in additional PRs.

PR #9881, opened 2019-04-01
Protect some I/O routines from SIGALRM

maximedenes:
This is necessary to prevent Coq from sending ill-formed output in some
scenarios involving `Timeout`.

Co-authored-by: Enrico Tassi <Enrico.Tassi@inria.fr>

### DIFF
--- a/lib/control.mli
+++ b/lib/control.mli
@@ -29,3 +29,14 @@ val timeout : int -> ('a -> 'b) -> 'a -> exn -> 'b
    API and it is scheduled to go away. *)
 type timeout = { timeout : 'a 'b. int -> ('a -> 'b) -> 'a -> exn -> 'b }
 val set_timeout : timeout -> unit
+
+(** [protect_sigalrm f x] computes [f x], but if SIGALRM is received during that
+    computation, the signal handler is executed only once the computation is
+    terminated. Otherwise said, it makes the execution of [f] atomic w.r.t.
+    handling of SIGALRM.
+
+    This is useful for example to prevent the implementation of `Timeout` to
+    interrupt I/O routines, generating ill-formed output.
+
+*)
+val protect_sigalrm : ('a -> 'b) -> 'a -> 'b
--- a/clib/cThread.ml
+++ b/clib/cThread.ml
@@ -100,10 +100,10 @@ let thread_friendly_input_value ic =
 (* On the ocaml runtime used in some opam-for-windows version the
  * [Thread.sigmask] API raises Invalid_argument "not implemented",
  * hence we protect the call and turn the exception into a no-op *)
-let protect_sigalrm f x =
+let mask_sigalrm f x =
   begin try ignore(Thread.sigmask Unix.SIG_BLOCK [Sys.sigalrm])
   with Invalid_argument _ -> () end;
   f x
 
 let create f x =
-  Thread.create (protect_sigalrm f) x
+  Thread.create (mask_sigalrm f) x
--- a/ide/idetop.ml
+++ b/ide/idetop.ml
@@ -466,7 +466,7 @@ let print_xml =
   let m = Mutex.create () in
   fun oc xml ->
     Mutex.lock m;
-    try Xml_printer.print oc xml; Mutex.unlock m
+    try Control.protect_sigalrm (Xml_printer.print oc) xml; Mutex.unlock m
     with e -> let e = CErrors.push e in Mutex.unlock m; iraise e
 
 let slave_feeder fmt xml_oc msg =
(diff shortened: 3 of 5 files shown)


SkySkimmer (commented):
Why are the functions called `foo_sigalrm` instead of `foo_sigalarm`?

ejgallego:
I am supposed to be the assignee here?

ejgallego (approved):
LGTM; this problem is pervasive all over the codebase tho, for example when doing `Ctrl-C` in printing etc... see related issues.

gares:
You forgot to protect against invalid arg on win

maximedenes:
> Why are the functions called `foo_sigalrm` instead of `foo_sigalarm`?

Isn't it the standard Unix name for the signal? In French for you: https://fr.wikipedia.org/wiki/SIGALRM

> You forgot to protect against invalid arg on win

Right, should be fixed now, thanks!

lib/control.mli
```diff
@@ -29,3 +29,14 @@ val timeout : int -> ('a -> 'b) -> 'a -> exn -> 'b
    API and it is scheduled to go away. *)
 type timeout = { timeout : 'a 'b. int -> ('a -> 'b) -> 'a -> exn -> 'b }
 val set_timeout : timeout -> unit
+
+(** [protect_sigalrm f x] computes [f x], but if SIGALRM is received during that
+    computation, the signal handler is executed only once the computation is
+    terminated. Otherwise said, it makes the execution of [f] atomic w.r.t.
+    handling of SIGALARM.
```
SkySkimmer: >Isn't it the standard Unix name for the signal?

OK but then this one is incorrect

JasonGross (dismissed):
Please do not make printing uninterruptible.  It is quite counter-productive; if I cannot interrupt printing, then I will have to kill the process, because I do not want to wait minutes or hours for printing to finish.  And then I lose all of the computation I've done up to the current point, and have to restart coqtop, perhaps just because I didn't realize how long it would take to print my goal with `Set Printing All`.  (Though I suppose if this does not affect PG it is not so terrible, but I think there must be a way to interrupt printing.)

gares:
@JasonGross  this PR is only masking `SIGALRM`, I guess you try to interrupt Coq using `SIGTERM`.

gares:
Failure on Ltac2 is spurious, restarting.

lib/control.mli
maximedenes: Right, fixed, thanks.

maximedenes:
@ejgallego can you be the assignee?

ejgallego:
Sure @maximedenes

ejgallego:
IMHO this can be merged quickly, there was a problem with the opam certificate expiring so I will rerun the OSX job to be sure.

ejgallego:
Something very weird is going one, has math-comp been broken upstream?

gares:
It seems the install target tries to install a test-suite file that was recently merged (but that this job does not compile)

gares:
What I said is true, but not the cause of the failure (that I don't get since install never fails)

ejgallego:
> What I said is true, but not the cause of the failure (that I don't get since install never fails)

I propose we don't block this PR on this failure, as long as the non-mathcomp jobs do pass.

JasonGross:
@gares How does SIGALRM get sent?  Now that I think about it, Ctrl-c generally sends SIGINT, right?

ejgallego:
> @gares How does SIGALRM get sent? Now that I think about it, Ctrl-c generally sends SIGINT, right?

Timeout @JasonGross

ejgallego:
@gares @maximedenes note that the async test suite seems to be still broken, but I guess this time by a different PR. Can you confirm you are OK with merging as is?

gares:
Exactly.

This pr should fix some issues. Some are fixed by ocaml 4.08. some still to track down.

maximedenes:
> Can you confirm you are OK with merging as is?

Yes.

> . Some are fixed by ocaml 4.08.

Oh yes, this is something we should have mentioned explictly. This fix in this PR is only partial for OCaml < 4.08, as there is no way to mask signals per thread. So we'd expect `Timeout`-related issues to be fixed on OCaml >= 4.08 + this patch.

ejgallego:
Ups, the 4.08 stuff looks important enough as to be mentioned in the changelog? Do you mind amending the commit message?

I will merge ASAP.

maximedenes:
> Do you mind amending the commit message?

I don't think it is necessary, because it is unrelated to this PR. This PR does what it says it does, regardless of the OCaml version.

However, a previous patch did introduce `Thread.sigmask` to protect the worker manager against some signals. I don't know if a `CHANGES` entry was added at that time. But if we want to add something, I believe it'd belong to another PR anyway.

ejgallego:
OK then!

PR #9882, opened 2019-04-01
[native compiler] Allow to set the output directory for cmx objects

ejgallego:
This is useful in order to implement native support in Dune for
example, which as of today as strict target rules.

Hopefully this option could go away; it is really internal, but I've
chosen to document it.

### DIFF
--- a/kernel/nativelib.ml
+++ b/kernel/nativelib.ml
@@ -26,12 +26,15 @@ let open_header = ["Nativevalues";
 let open_header = List.map mk_open open_header
 
 (* Directory where compiled files are stored *)
-let output_dir = ".coq-native"
+let output_dir = ref (Some ".coq-native")
 
 (* Extension of genereted ml files, stored for debugging purposes *)
 let source_ext = ".native"
 
 let ( / ) = Filename.concat
+
+(* Conditional append *)
+let ( /+ ) = fun base -> function None -> base | Some dir -> Filename.concat base dir
 
 (* We have to delay evaluation of include_dirs because coqlib cannot be guessed
 until flags have been properly initialized *)
@@ -68,7 +71,7 @@ let error_native_compiler_failed e =
 
 let call_compiler ?profile:(profile=false) ml_filename =
   let load_path = !get_load_paths () in
-  let load_path = List.map (fun dn -> dn / output_dir) load_path in
+  let load_path = List.map (fun dn -> dn /+ !output_dir) load_path in
   let include_dirs = List.flatten (List.map (fun x -> ["-I"; x]) (include_dirs () @ load_path)) in
   let f = Filename.chop_extension ml_filename in
   let link_filename = f ^ ".cmo" in
@@ -119,7 +122,7 @@ let compile_library dir code fn =
   let fn = fn ^ source_ext in
   let basename = Filename.basename fn in
   let dirname = Filename.dirname fn in
-  let dirname = dirname / output_dir in
+  let dirname = dirname /+ !output_dir in
   let () =
     try Unix.mkdir dirname 0o755
     with Unix.Unix_error (Unix.EEXIST, _, _) -> ()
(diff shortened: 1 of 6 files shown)


maximedenes:
I need more time to test this, as problems may not be caught. Indeed, when compiled files are not found, the native compiler silently recompiles stuff. So I'd like to add a warning and then make sure this PR doesn't trigger it.

I suggest to delay to delay to 8.11 anyway.

ejgallego:
How can this PR trigger a warning? If you don't pass the option there should be no functional change.

> I suggest to delay to delay to 8.11 anyway.

Main impact here is whether you'll like Dune + Coq users to have `(mode native)` in 8.10 or not.

maximedenes:
> Dune + Coq users to have `(mode native)`

Can you describe a way I can test it?

ejgallego:
> Can you describe a way I can test it?

I'll have to prepare a Dune PR which depends on this unmerged Coq branch; that will take some time.

maximedenes (commented):
I've tested the flag introduced by this PR. It works ok, but only when not mixing libraries built with different values passed for `-native-output-dir`. Given this limitation, shouldn't this be a one time flag (à la `configure`)?

Also, I'm not sure I like the possibility of defining the directory as `none`. What's the use case for it?

ejgallego:
> Also, I'm not sure I like the possibility of defining the directory as `none`. What's the use case for it?

Thanks for the review @maximedenes , the use case is to output the OCaml object files in the same directory the sources files are.

> I've tested the flag introduced by this PR. It works ok, but only when not mixing libraries built with different values passed for `-native-output-dir`. Given this limitation, shouldn't this be a one time flag (à la `configure`)?

Actually we set the flag only for building, but then we place the object files in the right directory so things will work without the flag.

My point of view is that the flag is an ugly hack to overcome a Dune limitation [which will be lifted in future versions of Dune], so if we decide to merge it IMHO it should be undocumented / internal and forbidden to use by regular folks.

There are many other things to do before the Dune build works well, so maybe the best is to wait.

Zimmi48:
Document as internal is always better than leaving undocumented. Experience has proved times and times again that leaving stuff undocumented does not prevent users from relying on it. Big fat warnings may have a stronger effect.

maximedenes:
> the use case is to output the OCaml object files in the same directory the sources files are.

Well, that's a description of the effect rather than a use case. The native compiler produces a lot of files with strange names (that's unfortunate, but interfacing with OCaml is not so easy), so it looks like a very non-hygienic build.

> Actually we set the flag only for building, but then we place the object files in the right directory so things will work without the flag.

Ah, I see. I'm fine with exposing such a flag, especially if it helps setting up a hygienic build.

> so if we decide to merge it IMHO it should be undocumented / internal and forbidden to use by regular folks.

Please add a few words like "for internal use only", remove the `none` case except if you need it for special reasons, and we can merge IMHO.

> the flag is an ugly hack to overcome a Dune limitation

Out of curiosity, what's the limitation?

ejgallego:
> it looks like a very non-hygienic build.

Indeed we don't care so much about this because when the flag is used we are inside `_build` anyways.

> Please add a few words like "for internal use only", remove the `none` case except if you need it for special reasons, and we can merge IMHO.

Well the only case we need is actually `none`, see below.

> Out of curiosity, what's the limitation?

The limitation is that a `rule` cannot have targets in a different directory.

This is indeed used so Dune can remove stale targets [as you were experiencing with the test-suite]

This limitation will be lifted at some point by allowing finer domains for rules.

Anyways, I propose we wait for now on this, for example `coq_dune` could work without the flag by actually setting a `.coq_native/dune` file.

maximedenes:
Thanks for the explanation, @ejgallego, it makes sense now. Let me know if you need it, no objection from my part.

I see you added the "needs: testing" label, but in fact I just tested this PR (and it works fine), that's why I commented on it.

ejgallego:
Let's wait a bit for now, I put the testing flag as I liked your suggestion for validation. Indeed we should add the flag only when we have a prototype working. I don't recall but I guess I submitted this looking more for feedback.

In fact I'll just close the PR [seems more efficient] and reopen when native support is working fully.

Issue #9251, opened 2018-12-19
coqchk: skip structure validation for nonchecked libraries (-norec)


SkySkimmer:
It seems to take a significant amount of the computation time. eg when foo is a vo from an empty file (requires the prelude), `coqchk -norec foo` takes about 0.17s but with validate commented out it takes 0.09s.
In the test suite most files are small so structure validation of requires probably takes a good amount of time.

PR #9890, opened 2019-04-02
Improve coqchk -norec perf by not checking for ill-formed data in dependencies

SkySkimmer:
Overall on test-suite/arithmetic/mod (which requires a big chain through Int63) we go from about 7.5s to 0.5s (15 times faster).

Close https://github.com/coq/coq/issues/9251


ppedrot (approved):

ppedrot:
I'll let a bit of time for others to comment before merging, just in case, but it seems pretty straightforward.

ppedrot:
CI failure is spurious, let's merge.

PR #9895, opened 2019-04-03
[loadpath] Make loadpath handling self-contained and move to vernac

ejgallego:
[loadpath] Make loadpath handling self-contained and move to vernac

We consolidate loadpath handling as a single `Loadpath` module from
parts in `Library` and `Mltop`, placing it at the `vernac` level [as
`Mltop`]

This idea was first suggested in https://github.com/coq/coq/pull/9808
, and indeed it makes sense as library resolution tends to be business
of the upper layers: IDE / build tools.

Logic could be pushed upwards, but this is good enough for now.

This consolidation has enabled some good and long overdue
refactorings, and the module should become self-contained enough as to
allow the resolution logic to be shared with `coqdep` in the future.

The `Mltop` module only cares now about ML-level modules, and should
go away once we rewrite the loader using `findlib` to solve
https://github.com/coq/coq/issues/7698 .

Overlay:

QuickChick/QuickChick#152


ejgallego:
cc @gares @maximedenes , this is a more radical alternative to #9808 ; I placed the milestone in 8.11 as this is not urgent in any way, but if we prefer going this route this PR should be OK to retarget as it is just a code reorganization.

ejgallego:
For example this approach makes it a tad easier to attach a loading mode to a particular `LoadPath`, so for system libraries you always load `.vo`, and for the current binded one you only look at `.vio`, etc...

IMVHO this is a work in progress so we will likely keep refining.

ejgallego:
Remarkably enough the loadpath logic has been in its current shape since the first versions of 1999, and the current mapping system since 2005 basically unchanged!

gares:
If I understand it correctly you want the upper layers to be able to control library resolution. Why not leaving the current, default, logic in library and add a library.init API that lets one override it at system or better component initialization time?

To me the code you move to vernac does not belong there.

Imo functional programming in its pure form is not helping. In a oo language you would just subclass library in jscoq, and leave the default logic where it belongs .

ejgallego:
> Why not leaving the current, default, logic in library and add a library.init API that lets one override it at system or better component initialization time?

At first I thought that way, but now IMHO the logic doesn't below there, as @maximedenes noted; that's the point of this PR; the compilation manager for example seems like the component to know whether for example files in `Coq` should be loaded as `.vo` or what is going on with the others.

IMHO the way the patch looks [which cleans up a lot Mltop and would allow to minimize the public API there] does support this view.

> To me the code you move to vernac does not belong there.

In fact it belongs in the STM or even in `toplevel`. Why does library have to know about path resolution, and a lot of system/build specific stuff which seems like pretty high-level stuff to me.

> Imo functional programming in its pure form is not helping. In a oo language you would just subclass library in jscoq, and leave the default logic where it belongs .

I don't think the style used is a big deal here [we could of course install the resolver as reference, but for now the delta is very small and functional programming does indeed help to manage a system > 200.000 lines of OCaml]

But even in OO-style, subclassing the whole of library just to overload the resolver would seem quite overkill to me; I would personally just have the constructor of library take a reference to a `Resolver` object.

gares:
What is the purpose of library then? To me mapping logical paths to physical paths is part of its job as well as understanding the format of libraries, handle their loading and saving.

I'm fine having all its duties be separate, and "overrideable" if necessary without replacing the whole component. But, see below

> reference to a `Resolver` object.

Yes, this is what you implemented, but you also moved the standard resolver class/code to the upper layers and you don't need to. You could leave it in library, have it be the default "resolver" argument to start library, and still be able to implement another one in jscoq.

In other words you want the *choice* of the resolver to be done in the toplevel, so that if you implement a new one you can plug in a resolver that fits better (this is the abstraction part, and I like it), but you don't want the standard resolved to be in toplevel, otherwise if I build jet another toplevel that is fine with the default resolver what do I do? Link the standard Coq toplevel just to be able to use the standard resolver?

ejgallego:
> You could leave it in library, have it be the default "resolver" argument to start library, and still be able to implement another one in jscoq.

The argument for moving the resolver is that I think the current one does belong to the toplevel as @maximedenes pointed out; IMHO library is a very low-level componenent and should not learn about "this loadpath was compiled with -quick" but or course it could live there as you propose.

> Link the standard Coq toplevel just to be able to use the standard resolver?

Note that you already have to link at least Mltop [which this PR cleanups] otherwise the whole stuff doesn't work.

IMHO if you read the patch the changes in Mltop do justify the move, at least to `vernac`.

maximedenes:
> library is a very low-level componenent and should not learn about "this loadpath was compiled with -quick" but or course it could live there as you propose.

I think there are two approaches here, which amount to basically the same result.

The key question is: what should "library" be in charge of?

Depending on the answer, you may want to keep library where it is and move some logic out of it (like you did in this PR), or you may consider making library itself higher level (at the level of the document manager).

ejgallego:
What do you think library should be in charge of @maximedenes?

gares:
My impression is that the code in mltop should be in Library ;-)
After all .cmxs are very like vo files.

maximedenes:
> What do you think library should be in charge of @maximedenes?

Of course I need to think more about it, but as a first impression I wouldn't be shocked if `Library` stayed in charge of library loading and saving, loadpath handling and the likes. The question then becomes: why does `Library` need to be so low-level. Ideally, it should be above `Vernac` and the STM. I suspect the current status quo comes from the fact that `Require`, `Add LoadPath`, `Cd` and the likes were backed into the vernacular, making vernacular execution depend on `Library`. Maybe that's the design choice that should be reconsidered.

gares:
Note that library is (today) also in charge or holding data for the library being constructed, in particular it contains `Global`. This is the main reason why it cannot be lifted.

ejgallego:
> The question then becomes: why does `Library` need to be so low-level? Ideally, it should be above `Vernac` and the STM

I can say that after all these experiments I do share that view; however as @gares points out not all stuff can be easily moved upwards; but IMHO this PR is a step in the right direction then.

> Maybe that's the design choice that should be reconsidered.

Having this stuff in the vernacular is annoying but IMHO not the end of the world, we just need to provide the right callbacks.

ejgallego:
Anyways despite all the interesting discussion, this is not urgent, however it interacts / integrates with the work on #9808 , in fact it goes in the line the first PR went, and enables some other use cases such as per-loadpath `vo/vio` selection, so I guess it is up to @maximedenes to see what we should do for 8.10.

ejgallego:
We had some more discussion and indeed I do believe this PR is good to go.

Indeed at some point the loadpath module itself should become a more independent component, but personally I am waiting to refactor coqdep as to be able to use the flexibility that Dune offers when organizing libs.

IMO this is a good intermediate step for now, and does open the door to quite a few additional refactorings in mltop / loadpath.

ejgallego:
Psssss, I messed up with reviews, as my own master tree is ahead of master; sorry folks :(

ejgallego:
CI is green, needs assignee. Note that we may want to merge with #9808 in which case we can target 8.10, this PR should not modify any semantics.

ejgallego:
@maximedenes would you like to be the assignee?

ejgallego:
Seems @maximedenes is busy and this PR has been unhandled for close to 2 months,  maybe @herbelin or @SkySkimmer would be interested?

SkySkimmer:
I might have a look but don't rely on me for this one.

>The main remaining task problem
to handle is how to handle the callback from the Add LoadPath
vernacular. Maybe it should be forbidden at the vernac level, becoming
a toplevel / STM only command?

What's the use case for Add LoadPath? What if we just removed it?

ejgallego:
> What's the use case for Add LoadPath? What if we just removed it?

Main use so far has been for `.coqrc`, not so long ago it was the recommended way to add new libraries installed somewhere.

May indeed be tricky to remove without impacting quite a few users.

ejgallego:
The PR description was pretty bad, I've updated it.

maximedenes (approved):
Separating loadpath handling from the rest of `library` (which in fact handles the current session) is a very good thing.

This is only a preliminary step, as:
- Loadpath handling should probably be higher level than `vernac` on the long run
- Other parts of `library` should be moved upwards (for example, I don't think `intern_library` and the likes need to be very lowlevel, do they?). In fact, only the session management should stay there.

But I think it is a move in the right direction. I'll merge soon if no opposition.

ejgallego:
> * Other parts of `library` should be moved upwards (for example, I don't think `intern_library` and the likes need to be very lowlevel, do they?). In fact, only the session management should stay there.

Indeed, see #9917 [draft PR]

This should also make #9808 very easy on top.

Issue #9512, opened 2019-02-07
omega and lia fail to identify primitive projections and their compatibility constants


andres-erbsen:
#### Description of the problem

I would expect the following to succeed, but it fails:

```coq
Require Import Coq.ZArith.BinInt Coq.omega.Omega Coq.micromega.Lia.

Set Primitive Projections.
Record params := { width : Z }.
Definition p : params := Build_params 64.

Set Printing All.

Goal width p = 0%Z -> width p = 0%Z.
  intros .

  assert_succeeds (enough True; [omega|]).
  assert_succeeds (enough True; [lia|]).

(*   H : @eq Z (width p) Z0 *)
(*   ============================ *)
(*   @eq Z (width p) Z0 *)

  change tt with tt in H.
  
(*   H : @eq Z (width p) Z0 *)
(*   ============================ *)
(*   @eq Z (width p) Z0 *)

  assert_succeeds (enough True; [lia|]).
  (* Tactic failure: <tactic closure> fails. *)
  assert_succeeds (enough True; [omega|]).
  (* Tactic failure: <tactic closure> fails. *)

  omega.
  (* Error: Omega can't solve this system *)

  Undo.
  lia.
  (* Tactic failure:  Cannot find witness. *)
Qed.
```

It succeeds if `Set Primitive Projections` is removed.

Bigger example [code](https://github.com/mit-plv/bedrock2/commit/b28c5aaef13f45b1789237a39f9209930c4dd9f8), [8.9 log](https://builds.sr.ht/~andres/job/26786#task-get_bedrock2-173), [master log](https://builds.sr.ht/~andres/job/26791#task-bedrock2-368) (disregard the branch and file names, I chose them before I understood what this behavior was about).

Even larger context: goals with such superficial differences occur when proving that calculations done in bedrock2 programs running on 64-bit integers correspond to the same operations on mathematical integers by tracking ranges. We are trying to make these proofs as automated as possible, and this issue is currently a snag in that process. Any automatable workaround would be appreciated.

cc @samuelgruetter 

#### Coq Version

8.9 and e209341f6b9858279d2fe53d1eb795ff25925eed

PR #9897, opened 2019-04-03
Ignore the hidden boolean when comparing projections and fix micromega/lia/psatz wrt projections

### LINKED ISSUE #9892: omega and lia fail to recognize equal terms if primitive projections are on
With @JasonGross's help, I managed to minimize something similar to the follow-up problem I posted in https://github.com/coq/coq/issues/9268:

```
Require Import Lia.
Require Import Omega.
Set Primitive Projections.

Record foo(a: unit) := { field: nat }.

Goal forall x, field tt x = 1 -> field tt x = 1.
  intros.
  cbv [field].
  Set Printing Universes.
  Set Printing All.
  Fail omega.
  Fail lia.
```

Expected behavior: `omega` and `lia` should succeed.

Coq version: Coq master 424c1973e96dfbf3b2e3245d735853ffa9600373 of Mon Apr 1 19:40:53 2019 +0000.

Might be related to/the same as https://github.com/coq/coq/issues/9512.

vbgl:
Fixes #9892.

This breaks one test case, that was introduced there: https://github.com/coq/coq/commit/b6e39ade125862ba41ca17b06b8e35726b9b0d7d

However, since this is the only thing that seems to break, I’ve amended the test ;-) Ping @mattam82

### DIFF
--- a/doc/changelog/04-tactics/09897-fix-micromega-wrt-projections.rst
+++ b/doc/changelog/04-tactics/09897-fix-micromega-wrt-projections.rst
@@ -1,0 +1,3 @@
+- Micromega tactics (:tacn:`lia`, :tacn:`nia`, etc) are no longer confused by
+  primitive projections (`#9897 <https://github.com/coq/coq/pull/9897>`_,
+  by Vincent Laporte).
--- a/plugins/micromega/coq_micromega.ml
+++ b/plugins/micromega/coq_micromega.ml
@@ -745,7 +745,7 @@ struct
     (** [eq_constr gl x y] returns an updated [gl] if x and y can be unified *)
     let eq_constr gl x y =
       let evd = gl.sigma in
-      match EConstr.eq_constr_universes gl.env evd x y with
+      match EConstr.eq_constr_universes_proj gl.env evd x y with
       | Some csts ->
          let csts = UnivProblem.to_constraints ~force_weak:false (Evd.universes evd) csts in
          begin
@@ -769,15 +769,16 @@ struct
     ({vars=vars';gl=gl'}, CamlToCoq.positive n)
 
    let get_rank env v =
-     let evd = env.gl.sigma in
+     let gl = env.gl in
 
      let rec _get_rank env n =
        match env with
        | [] -> raise (Invalid_argument "get_rank")
       | e::l ->
-         if EConstr.eq_constr evd e v
-         then n
-         else  _get_rank l (n+1)  in
+        match eq_constr gl e v with
+        | Some _ -> n
+        | None -> _get_rank l (n+1)
+     in
      _get_rank env.vars 1
 
    let elements env = env.vars
--- a/test-suite/bugs/closed/bug_3656.v
+++ b/test-suite/bugs/closed/bug_3656.v
@@ -48,7 +48,3 @@ Goal setT = setT.
     | _ => idtac
   end. (* should not fail *)
 Abort.
-
-Goal forall h, setT h = setT h.
-Proof. intro. progress unfold setT.
-Abort.
--- a/doc/changelog/01-kernel/09897-constr_eq-primproj.rst
+++ b/doc/changelog/01-kernel/09897-constr_eq-primproj.rst
@@ -1,0 +1,3 @@
+- Term (constr) comparison ignores whether primitive projections are unfolded
+  or not (`#9897 <https://github.com/coq/coq/pull/9897>`_,
+  by Vincent Laporte).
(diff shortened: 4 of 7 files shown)


plugins/micromega/coq_micromega.ml
```diff
@@ -731,7 +731,7 @@ struct
     (** [eq_constr gl x y] returns an updated [gl] if x and y can be unified *)
     let eq_constr gl x y =
       let evd = gl.sigma in
-      match EConstr.eq_constr_universes gl.env evd x y with
+      match EConstr.eq_constr_universes_proj gl.env evd x y with
```
SkySkimmer: This change isn't mentioned by the PR title.

test-suite/bugs/closed/bug_3656.v
```diff
@@ -50,5 +50,5 @@ Goal setT = setT.
 Abort.
 
 Goal forall h, setT h = setT h.
-Proof. intro. progress unfold setT.
+Proof. intro. Fail progress unfold setT.
 Abort.
```
SkySkimmer: Just remove this test.

SkySkimmer (commented):
Should the unfold bool be ignored by syntactic equality? I guess we already ignore casts, so we may as well ignore this bool.
We probably want a Changes entry though.

plugins/micromega/coq_micromega.ml
vbgl: Updated.

test-suite/bugs/closed/bug_3656.v
vbgl: Done.

vbgl:
I’ve added a line in the CHANGES log.

CHANGES.md
```diff
@@ -62,6 +62,9 @@ Kernel
   since guessing the decreasing argument can now be more expensive.
   (PR #9602)
 
+- Term (constr) comparison ignores the internal boolean of primitive
+  projections.
```
SkySkimmer: Maybe also mention micromega being fixed here, tactic changes can be fragile.

maximedenes: This entry is impossible to understand for non-developers. Try to reexpress it in terms of unfolding of the projections, at least.

vbgl: Do non-developers understand the unfolding of primitive projections? Running `unfold p` where `p` is a primitive projections has no visible effect (beyond breaking a few tactics…).

CHANGES.md
```diff
@@ -131,7 +134,8 @@ Tactics
 - The syntax of the `autoapply` tactic was fixed to conform with preexisting
   documentation: it now takes a `with` clause instead of a `using` clause.
 
-
+- Micromega tactics (`lia`, `nia`, etc) are no longer confused by primitive
+  projections.
```
andres-erbsen: Does this PR also fix #9512?

vbgl: It fixes it for `lia`, not for `omega`.

SkySkimmer:
Bedrock is broken.

andres-erbsen:
rerunning bedrock2 ci might fix it, we disabled primitive projections a couple of days ago...

test-suite/bugs/closed/bug_3656.v
```diff
@@ -48,7 +48,3 @@ Goal setT = setT.
     | _ => idtac
   end. (* should not fail *)
 Abort.
-
-Goal forall h, setT h = setT h.
-Proof. intro. progress unfold setT.
```
JasonGross: It seems to me like `progress` should record changes such as this.  E.g., if `change tt with tt` or `unfold setT` is going to have any impact on possible behavior of subsequent tactics, `progress` should record it.  Maybe you can have the conversation algorithm take an optional Boolean about whether or not to consider projections the same as their compatibility constants, have it default to true, and pass false only for progress?

SkySkimmer:
Someone needs to decide if we want to do this change.

maximedenes:
> Someone needs to decide if we want to do this change.

It looks desirable, but we need a larger plan of how we want to deal with primitive projections unfolding (especially for tactics).

doc/changelog/04-tactics/09897-fix-micromega-wrt-projections.rst
```diff
@@ -0,0 +1,3 @@
+- Micromega tactics (`lia`, `nia`, etc) are no longer confused by primitive
```
Zimmi48:
```suggestion
- Micromega tactics (:tacn:`lia`, :tacn:`nia`, etc) are no longer confused by primitive
```

gares:
WG discussion: we target 8.12, we favor a global approach to solve this class of bugs (namely remove the compat layer but add to the upper layers support for primitive projections, eg a name for them) working group with @mattam82, @SkySkimmer and @maximedenes .

vbgl:
I think that the second commit in this PR nonetheless has good changes in it. Frédéric, do you confirm?

Ping @fajb

fajb:
I confirm. This cannot hurt to detect that a projection and an application represent the same term.

PR #9900, opened 2019-04-03
[native compiler] Fix critical bug with stuck primitive projections

### LINKED ISSUE #9684: Index out of bounds in native_compute

Running a native_compute in file examples/AlmostFull.v of this branch of Equations https://github.com/mattam82/Coq-Equations/tree/almost_full_native_compute_bug
on current master results in an "Index out of bounds" error. It's hard to minimize. I would bet we're doing something wrong with universes (there are a few polymorphic things there).

#### Coq Version

master

maximedenes:
Since e1e7888, stuck projections were not computed correctly.

Fixes #9684

### DIFF
--- a/pretyping/nativenorm.ml
+++ b/pretyping/nativenorm.ml
@@ -222,7 +222,12 @@ and nf_type_sort env sigma v =
   match kind_of_value v with
   | Vaccu accu -> 
       let t,s = nf_accu_type env sigma accu in
-      let s = try destSort s with DestKO -> assert false in
+      let s =
+        try
+          destSort (whd_all env s)
+        with DestKO ->
+          CErrors.anomaly (Pp.str "Value should be a sort")
+      in
       t, s
   | _ -> assert false
 
--- a/dev/doc/critical-bugs
+++ b/dev/doc/critical-bugs
@@ -207,6 +207,15 @@ Conversion machines
   GH issue number: ?
   risk: 
 
+  component: primitive projections, native_compute
+  summary: stuck primitive projections computed incorrectly by native_compute
+  introduced: 1 Jun 2018, e1e7888a, ppedrot
+  impacted released versions: 8.9.0
+  impacted coqchk versions: none
+  found by: maximedenes exploiting bug #9684
+  exploit: test-suite/bugs/closed/bug_9684.v
+  GH issue number: #9684
+
 Conflicts with axioms in library
 
   component: library of real numbers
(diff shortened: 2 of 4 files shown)


pretyping/nativenorm.ml
```diff
@@ -222,7 +222,12 @@ and nf_type_sort env sigma v =
   match kind_of_value v with
   | Vaccu accu -> 
       let t,s = nf_accu_type env sigma accu in
-      let s = try destSort s with DestKO -> assert false in
+      let s =
+        try
+          destSort (whd_all env s)
+        with DestKO ->
+          CErrors.anomaly (Pp.str "Value should be a sort")
+      in
```
SkySkimmer: Why bother catching the destko?

maximedenes: This was there before. I guess the idea is that `DestKO` is used as a catchable exception in the codebase. So we want to raise a fatal exception here.

Zimmi48 (approved):

Zimmi48:
Why not mark this for backporting (i.e. put this in the 8.9.1 milestone)? Otherwise, the "Impacted version" information is quickly going to become out of date when 8.9.1 is released.

maximedenes:
> Why not mark this for backporting (i.e. put this in the 8.9.1 milestone)? Otherwise, the "Impacted version" information is quickly going to become out of date when 8.9.1 is released.

Right, it is better in the 8.9.1 milestone. Note that only one of the two commits will have to be backported (the other one is fixing an aggravation of the problem after the merge of `SProp`).

pretyping/nativenorm.ml
SkySkimmer: Is there a reason to expect a destko to happen here? Considering we raise an anomaly it seems strange to me.

maximedenes: > Is there a reason to expect a destko to happen here?

No, except in a tricky scenario as was the case. It should certainly be a fatal error, otherwise debugging will be a pain (imagine this gets caught by unification).

I'm afraid I don't understand your concern. You feel more comfortable with raising an exception that can be silently caught by random clients?

ppedrot (approved):
That will teach me to look up for the place where a lexical variable is bound, a.k.a λ-calculus 101.

maximedenes:
@ppedrot can you be the assignee?

maximedenes:
@ppedrot ping

PR #9901, opened 2019-04-04
[dune] Fix include object dirs.

ejgallego:
In Dune >= 1.8 object dirs have changed; this should be stable for a
while, however we want a more robust setup for sure, which I think it
should be able to be implemented when we have a single build system.


Zimmi48 (approved):
Given that this does fix the use of

```
dune exec dev/shim/coqbyte-prelude
> Drop.
# #directory "dev";;
# #use "include_dune";;
```

with Dune >= 1.8, and given that this is of interest mostly to developers themselves, and given that the `default.nix` file is already pinned to a version of nixpkgs that includes Dune >= 1.8, I think this is fine to merge this PR even if the actual version of Dune that we are requiring for building Coq 8.10 is Dune 1.6.

ejgallego:
Thanks once again @Zimmi48, indeed this idea require 1.8 for devs but this is what we most use. We will update the base image soon anyways so I considered overkill just doing it for this.

PR #9904, opened 2019-04-04
[CI] Fix build of math-comp dependencies

maximedenes:
We were incorrectly calling the global `install` target even when
building only subcomponents of the library.

### DIFF
--- a/dev/ci/ci-common.sh
+++ b/dev/ci/ci-common.sh
@@ -108,10 +108,9 @@ install_ssreflect()
 
   git_download mathcomp
 
-  ( cd "${CI_BUILD_DIR}/mathcomp/mathcomp" && \
-    make Makefile.coq && \
-    make -f Makefile.coq ssreflect/all_ssreflect.vo && \
-    make -f Makefile.coq install )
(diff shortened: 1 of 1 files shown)


SkySkimmer (approved):

SkySkimmer:
Do we want to backport this? cc @silene

vbgl:
Thank you!

gares:
> We were incorrectly calling the global `install` target even when
> building only subcomponents of the library.

and it was working because, among the many files to install, the last one (in alphabetic order) was there masking all previous failures

gares (approved):

maximedenes:
> and it was working because, among the many files to install, the last one (in alphabetic order) was there masking all previous failures

Which can be considered a math-comp bug. Should I report it?

SkySkimmer:
>Which can be considered a math-comp bug. Should I report it?

This is coq_makefile behaviour. The coq_makefile install rule does (simplified)
~~~bash
for f in $(toinstall); do 
  install $f $dest-of-f
done
~~~
whose exit code is the code of the last `install` to run.
I guess we should set -e or `|| exit 1` or some such.

Issue #9907, opened 2019-04-04
Pin addon/plugin version for 8.9.1


MSoegtropIMC:
The addon/plugin versions for 8.9 have (mostly) not been updated since Sep 2018.

Check what is the latest versions which builds (or otherwise "best" version) for all plugins/addons delivered in the Windows installer and pin the versions in ci-basic-overlay.sh accordingly.

Zimmi48:
PR was merged. (@MSoegtropIMC We have to close most of your issue by hand because you don't take advantage of GitHub's closing feature: you should write "Close #9907" instead of "Issue #9907" in your commit messages.)

PR #9908, opened 2019-04-04
Pin addon version for 8.9.1 (Issue #9907)

MSoegtropIMC:
This PR updates the pinned version for all addons/plugins delivered with the Windows Installer.

I generally pinned only to tags or ref hashes, since I think we once agreed that release builds should be reproducible. In case there was no (suitable) tag, I used the latest (working) ref hash on a 8.9 or master branch.

### DIFF
--- a/dev/ci/ci-basic-overlay.sh
+++ b/dev/ci/ci-basic-overlay.sh
@@ -9,8 +9,8 @@
 ########################################################################
 # MathComp
 ########################################################################
-# Latest commit on master as of Sep 27, 2018
-: "${mathcomp_CI_REF:=5f8d45b54aa98732ec3de43d91814459d5a2f2e4}"
+# Latest release tag as of April 8th 2019
+: "${mathcomp_CI_REF:=mathcomp-1.8.0}"
 : "${mathcomp_CI_GITURL:=https://github.com/math-comp/math-comp}"
 : "${mathcomp_CI_ARCHIVEURL:=${mathcomp_CI_GITURL}/archive}"
 
@@ -29,13 +29,13 @@
 ########################################################################
 # Unicoq + Mtac2
 ########################################################################
-# Latest commit on master as of Sep 27, 2018
-: "${unicoq_CI_REF:=1cec038ab34e03109f5587a8aecda1f6c53495dd}"
+# This is the latest tag for 8.9 as of April 8 2019
+: "${unicoq_CI_REF:=v1.3-8.9}"
 : "${unicoq_CI_GITURL:=https://github.com/unicoq/unicoq}"
 : "${unicoq_CI_ARCHIVEURL:=${unicoq_CI_GITURL}/archive}"
 
-# Latest commit on master-sync as of Sep 27, 2018
-: "${mtac2_CI_REF:=e65c2560e5098df5e7333f19db97e9f39e46c3ee}"
+# This is the latest tag for 8.9 as of April 8 2019
+: "${mtac2_CI_REF:=v1.1-coq8.9}"
 : "${mtac2_CI_GITURL:=https://github.com/Mtac2/Mtac2}"
 : "${mtac2_CI_ARCHIVEURL:=${mtac2_CI_GITURL}/archive}"
 
@@ -81,6 +81,11 @@
 ########################################################################
 # Ltac2
 ########################################################################
+# The v8.9 branch looks unmaintained-
+# the last two commits (as of April 4 2019) are:
+# Commits on Oct 23, 2018
+# Commits on Jul 2, 2018
+# master does not build
 : "${ltac2_CI_REF:=v8.9}"
 : "${ltac2_CI_GITURL:=https://github.com/ppedrot/ltac2}"
 : "${ltac2_CI_ARCHIVEURL:=${ltac2_CI_GITURL}/archive}"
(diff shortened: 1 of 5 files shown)


dev/ci/ci-basic-overlay.sh
```diff
@@ -29,13 +29,16 @@
 ########################################################################
 # Unicoq + Mtac2
 ########################################################################
-# Latest commit on master as of Sep 27, 2018
-: "${unicoq_CI_REF:=1cec038ab34e03109f5587a8aecda1f6c53495dd}"
+# There is no (apparent) 8.9 branch
```
Zimmi48: There is a tag: v1.3-8.9 (found in https://github.com/unicoq/unicoq/releases). I'd recommend using such a tag when available.

MSoegtropIMC: Thanks for the hint! I will also include the link to the release page for next time.

Do you have an opinion on if a 8.9 branch should be used if there is one? I think we once discussed that releases should only use tags and hashes, but I am not so sure when the release will be and planned to replace the branches with current tags / hashes shortly before the release and change it back later.

Zimmi48: I wouldn't mind using the v8.9 branch when it exists instead of a commit hash / tag. But note that @silene plans to release 8.9.1 soon anyway (possibly as soon as next week).

dev/ci/ci-basic-overlay.sh
```diff
@@ -195,8 +203,11 @@
 ########################################################################
 # Bignums
 ########################################################################
-# Latest commit on master as of Sep 27, 2018
-: "${bignums_CI_REF:=f1a63cb4cce9a79e4406275fbb0ee2269947f8d2}"
+# There is no 8.9 branch
+# Below is the latest commit (Feb 4, 2019) on master as of April 4, 2019
+# 0cd435d0a3c731605536c83d0a731c3fc336cce7 => fails (requires Cyclic63)
+# This is the latest working commit (Feb 1, 2019) on master as of April 4, 2019
+: "${bignums_CI_REF:=5a7ba3d02d8c095ce31fc5f0a0dbc118c58dd433}"
```
Zimmi48: You shouldn't care about Bignums. It is not updated but for compatibility fixes. Therefore, the any working commit for 8.9 would be equivalent feature-wise.

MSoegtropIMC: There are some minor changes in CI files and gitignore in the version I picked, but I could also use the same version as 8.9.0, if you prefer.

Zimmi48: I don't actually care, I was just saying to not bother in general.

MSoegtropIMC:
@ejgallego , @Zimmi48 : could you please have a look at the failures of ssreflect build as part of Coquelicot CI in these pipelines:

<https://gitlab.com/coq/coq/-/jobs/192841629> (fails with matchcomp 1.8)
vs.
<https://gitlab.com/coq/coq/-/jobs/191160399> (passes with some previous mathcomp)

In both pipelines the make install of ssreflect produces a ton of errors, but with mathcomp 1.8 it fails, while it doesn't with the previous versions and I don't see why. The picked versions work fine in the Windows build, so I guess the issue is with the `install_ssreflect()` function in `ci_common.sh`.

MSoegtropIMC:
P.S.: please note that the CI build of mathcomp as such passed, just the ssreflect partial build in Coquelicot CI failed. I also wonder if ssreflect isn't part of main coq these days and if this partial build is required at all.

ejgallego:
@MSoegtropIMC where is the script doing the build? You need to apply #9904 AFAICT.

MSoegtropIMC:
> where is the script doing the build?

it should be ci-coquelicot.sh and ci-common.sh.

> You need to apply #9904 AFAICT.

Please note that this is a 8.9 branch PR, so if I understand you right, I should back port #9904 to 8.9 and piggy back #9904 with this PR?

ejgallego:
> Please note that this is a 8.9 branch PR, so if I understand you right, I should back port #9904 to 8.9 and piggy back #9904 with this PR?

I think so.

ejgallego:
The build method in 8.9 do seem incorrect indeed, if you are targeting math-comp master.

MSoegtropIMC:
@silene : FYI, I will backport #9904 to the 8.9 branch as suggested by @ejgallego above. Please note that this is not an issue with mathcomp 1.8 as such on 8.9.1 or with pinning 8.9.1 to use mathcomp 1.8, it is an issue with the CI of components using ssreflect.

MSoegtropIMC:
@ejgallego : I fear I need some more explanations on how the CI builds work. For quickchick my best guess on what is going wrong is that it tries to compile simple-io and extlib in in parallel, but simple-io depends on extlib and fails because extlib is not there (yet). The exact same versions of extlib and simple-io build fine in the Windows build.

Can you explain how the dependencies are managed in the CI build and where the scripts are, which trigger the build of extlib and simple-io for quickchick? I can't find this in the /dev/ci folder and the quickchick make files seem to assume that these modules are there.

ejgallego:
@MSoegtropIMC some amount of dependencies are handled in `Makefile.ci`; as far as I can see, the simple-io build should not be attempted before the `ext-lib` one suceeded.

ejgallego:
Indeed the `Makefile.ci` used in this PR https://github.com/coq/coq/blob/0d1b8863de695bd39c9fc06a4b39900d8b1c4e5f/Makefile.ci is incorrect, it is missing the dependency on `ext-lib`. See the one in master and add:
```
ci-simple-io: ci-ext-lib
```
Cheers,
E

MSoegtropIMC:
@ejgallego : Thanks for the explanation and analysis!

So it was just luck that it did work in the 8.9 branch before. I will fix it.

MSoegtropIMC:
@silene : from my point of view the pinning is finished. The failure of the windows64 build is a sporadic timeout failure. Without (substantial) changes on the Windows side it did run through here:

https://gitlab.com/coq/coq/-/jobs/193957923

Issue #6523, opened 2017-12-29
`evar_{,filtered}_env` considered harmful.


ejgallego:
Both `Evd.evar_env` and `Evd.evar_filtered_env` are implemented using a call to `Global.env ()`, their current form should eventually go away.

ejgallego:
IMHO the full fix for this is more involved, eventually these functions should not exist at all.

ppedrot:
I don't think that this issue is relevant anymore. We're removed the environment and pushing the global calls upwards, but it's hardly related to this precise API. I'm closing it because I don't see how we would fix it more than it is currently.

PR #9909, opened 2019-04-04
Remove all but one call to `Global` in the pretyper

maximedenes:
I'm making the pretyper functional w.r.t. the environment, reorganizing the code when necessary.


maximedenes:
@ejgallego may be interested

ejgallego:
> @ejgallego may be interested

Looks pretty good so far!

tactics/tactics.ml
```diff
@@ -1455,7 +1455,8 @@ exception IsNonrec
 let is_nonrec mind = (Global.lookup_mind (fst mind)).mind_finite == Declarations.BiFinite
 
 let find_ind_eliminator ind s gl =
-  let gr = lookup_eliminator ind s in
+  let env = Proofview.Goal.env gl in
```
ejgallego: Just a comment, be careful here as it is not always the case the this env will be the global env as before; in particular see #6523 and the actual code that creates `Goal.env`:
```
  let gmake_with info env sigma goal state =
    { env = Environ.reset_with_named_context (Evd.evar_filtered_hyps info) env ;
```
`tclEnv` should be better I think, but I am not sure.

ppedrot: Doesn't matter, `lookup_eliminator` only looks at the global table. Moreover, calls to `tclENV` should be an exception, most of the time the relevant data is the goal context.

gares: Indeed it would be way more civilized to have a type of global environements and a type of proof environments (that extends the latter with the proof context)

maximedenes:
So, after this PR there remains calls to the global environment only in:
- Heads, proper action is being investigated in #9915 
- Detyping, but fixing it requires improving the treatment of section variables (will try to do something)

The next steps will be to remove dependency on `Libobject` and `Lib`, then make the state explicit (removing the need for `Summary`). Then we'll have a functional pretyper, modulo `Nametab` (I'm not sure what to do about it yet).

maximedenes:
@ppedrot can you be the assignee for this PR?

interp/implicit_quantifiers.ml
```diff
@@ -231,7 +231,9 @@ let implicit_application env ?(allow_partial=true) f ty =
     | Some ({CAst.loc;v=(id, par, inst)}, gr) ->
 	let avoid = Id.Set.union env (ids_of_list (free_vars_of_constr_expr ty ~bound:env [])) in
 	let c, avoid =
-	  let c = class_info gr in
+   let env = Global.env () in
+   let sigma = Evd.from_env env in
+          let c = class_info env sigma gr in
```
gares: some indentation got wrong here

library/goptions.ml
```diff
@@ -142,7 +142,7 @@ struct
   type key = string
   let compare = String.compare
   let table = string_table
-  let encode x = x
+  let encode env x = x
```
gares: _env

pretyping/typeclasses.mli
```diff
@@ -54,19 +53,25 @@ type typeclass = {
...
 
-type instance
+type instance = {
+  is_class: GlobRef.t;
+  is_info: hint_info;
+  (* Sections where the instance should be redeclared,
+     None for discard, Some 0 for none. *)
+  is_global: int option;
+  is_impl: GlobRef.t;
+}
```
gares: Can you explain why this is needed? (the last two fields seem not to be tampered)

pretyping/typeclasses.mli
```diff
@@ -78,7 +83,8 @@ val dest_class_app : env -> evar_map -> EConstr.constr -> (typeclass * EConstr.E
 val typeclass_univ_instance : typeclass Univ.puniverses -> typeclass
 
 (** Just return None if not a class *)
-val class_of_constr : evar_map -> EConstr.constr -> (EConstr.rel_context * ((typeclass * EConstr.EInstance.t) * constr list)) option
+val class_of_constr : env -> evar_map -> EConstr.constr ->
+  (EConstr.rel_context * ((typeclass * EConstr.EInstance.t) * constr list)) option
 
 val instance_impl : instance -> GlobRef.t
```
gares: If the type is transparent, then this and the one below should go

vernac/classes.ml
```diff
@@ -49,17 +51,224 @@ let classes_transparent_state () =
...
+  let global =
+    if glob then Some (Lib.sections_depth ())
+    else None
+  in
+  if match global with Some n -> n>0 && isVarRef impl | _ -> false then
+    CErrors.user_err (Pp.str "Cannot set Global an instance referring to a section variable.");
+  { is_class = cl.cl_impl;
+    is_info = info ;
+    is_global = global ;
+    is_impl = impl }
```
gares: Ok, I guess I got my answer

gares (approved):
LG.Suggestion: in the fer calls to tclENV I'd use a name other than env to bind the result, since it is *not* the env most of the code in the monad should use, and one may just see an env in scope and use it (in a later patch, I mean). I'd pick `global_env`.

pretyping/typeclasses.mli
gares: San't some of this data be move to vernac as well? I mean leave here the fields that are needed here, and move upward the ones for libstack management (assuming I got it right)

tactics/tactics.ml
ejgallego: @maximedenes indeed you could pass something of type `Environ.globals` if it wasn't for the resolver; but something weird is going on lookup eliminator, why does the nametab need to get involved with a name that came out of the kernel?

pretyping/typeclasses.mli
maximedenes: You are right that we should improve the separation of concerns even more than this PR does. I plan to do it as part of the libobject redesign. Roughly, instead of relying on inverted control spread everywhere in the code, we should compile vernac ASTs to vernac actions, in direct style. It will be easier to clarify the interfaces at that point.

pretyping/typeclasses.mli
maximedenes: I prefer to keep it for now, for better compatibility. If they go away, it will be part of the libobject redesign.

ejgallego (commented):
A couple more comments:
- using `Proofview.Goal.env` to replace `Global.env ()` does introduce a change of semantics; I am personally OK with the change in the PR if needed, however I think: a) in most cases using this function is wrong, b) the criteria in other cases was for PRs of this kind to preserve the old semantics if broken.
- I find the title of the PR slightly misleading, as there may remain non-functional behavior. I didn't look in detail but one example is the call to `Decls.variable_opacity` in `Heads`.

ejgallego:
> * the call to `Decls.variable_opacity` in `Heads`.

Would be great to get rid of this call by the way.

maximedenes:
> * using `Proofview.Goal.env` to replace `Global.env ()` does introduce a change of semantics

If I got it right, I changed it only for functions which look for inductive types or global constants, so it does not change the semantics. Don't hesitate to point me to some occurrences if they are not clear.

> * I find the title of the PR slightly misleading, as there may remain non-functional behavior.

Yes, I should say that I mean functional w.r.t. the `env`.

> Would be great to get rid of this call by the way.

Yes, will do as soon as #9915 is merged.

ejgallego:
> Yes, will do as soon as #9915 is merged.

Cool!

Regarding the two types of envs indeed that's a huge problem, however I do think that it is "morally" wrong to use `Proofview.Goal.env` for functions that are accessing the global stuff, as morally this proof env should have a different type. So indeed I think in all these cases `tclEnv` is preferred even if not observable difference happens.

ejgallego:
> is preferred even if not observable difference happens.

Not a blocker in any way tho, but IMHO makes the code a tad confusing [due to the API's fault]

maximedenes:
> Regarding the two types of envs indeed that's a huge problem

Distinguishing the two at the type level is next on my todo-list ;)

ejgallego:
> Distinguishing the two at the type level is next on my todo-list ;)

Cool, so then you want to use `tclEnv` always here IMHO.

ppedrot:
@ejgallego I believe the exact opposite. `tclENV` vs `Goal.env` is not about globals vs. named context, it is about outside scope (i.e. section variables) vs. hypotheses. Due to the way sections work currently, I find it highly advisable never to use `tclENV` except to witness section-specific data.

ejgallego:
> I find it highly advisable never to use `tclENV` except to witness section-specific data.

Ok, I see what you mean, even tho there are some problems with the current code [sic] I take back my comment then; I was kinda envisioning a model where we would strip globals from the goal specific env and provide a `tclGlobals`, but you are right it is best to use the goal API.

Who needs to observe section-specific hyps? Can't those be told apart by some evar status?

The way you describe it made me think that we should then deprecate `tclEnv` and place it into `Unsafe`. WDYT?

ppedrot:
> Who needs to observe section-specific hyps? Can't those be told apart by some evar status?

Oh, unbearable life! This is not possible, for there exists the `clear` tactic!

TL;DR: this would require reimplementing sections from scratch to make any sense of this status.

ejgallego:
> Oh, unbearable life! This is not possible, for there exists the `clear` tactic!

I don't pronounce forbidden words until the sun sets; but now that you did, let me redo my question, who needs `tclEnv` apart from the tactic-you-shall-not-pronounce?

ejgallego:
> TL;DR: this would require reimplementing sections from scratch to make any sense of this status.

I wonder what @maximedenes is thinking, maybe using the same underlying data and some ghost types?

maximedenes:
> I wonder what @maximedenes is thinking in that front, maybe using the same underlying data and some ghost types?

Let's discuss it here instead: https://github.com/coq/coq/issues/6254

ejgallego:
There are two issues here @maximedenes one is #6254 but the other is splitting globals from the env, I am not sure we have an inside for that. And likely, we should also talk about notation data too, which lives in a bit of a limbo.

ejgallego:
There are two issues here @maximedenes one is #6254 but the other is splitting globals from the env, I am not sure we have an inside for that. And likely, we should also talk about notation data too, which lives in a bit of a limbo.

maximedenes:
> There are two issues here @maximedenes one is #6254 but the other is splitting globals from the env, I am not sure we have an inside for that. And likely, we should also talk about notation data too, which lives in a bit of a limbo.

Ok, but in both cases, this is off-topic for this PR, I'm afraid ;)

maximedenes:
> in the fer calls to tclENV I'd use a name other than env to bind the result

done

maximedenes:
@ppedrot This seems ready.

maximedenes:
@ppedrot Thank you!

PR #9910, opened 2019-04-05
[api] [proofs] Remove dependency of proofs on interp.

ejgallego:
We perform some cleanup and remove dependency of `proofs/` on
`interp/`, which seems logical.

In fact, `interp` + `parsing` are quite self-contained, so if there is
interest we could also make tactics to depend directly on proofs.


SkySkimmer (approved):
This looks simple, I'll take it.

PR #9914, opened 2019-04-05
Abort before corrupting memory.

silene:
Discovered by @proux01 in #9905


Zimmi48 (commented):
Could you please add a new item to the list in `dev/doc/critical-bugs`?

Zimmi48 (approved):

vbgl (dismissed):
The change makes perfect sense. Thanks.

silene:
Unfortunately, this abort is triggered in the contributions. For example, in CompCert, file `cparser/Parser.v`:
```coq
Theorem complete:
  Parser.complete_validator () = true.
Proof eq_refl true<:Parser.complete_validator () = true.
```

vbgl:
It also appears in the CI (Azure, macOS):

> COQC      theories/FSets/FMapFullAVL.v
make[1]: *** [theories/FSets/FMapFullAVL.vo] Abort trap: 6

vbgl:
Maybe Benjamin can shed some light on this issue.

Ping @bgregoir

silene:
For some reason, `caml_young_limit` is equal to `caml_young_alloc_end` at that point.

According to the source code, "`caml_young_limit` is the pointer that is compared to `caml_young_ptr` for allocation. It is either `caml_young_alloc_end` if a signal is pending and we are in native code, or `caml_young_trigger`."

I have no clue why a signal would be pending there.

vbgl:
May we add a non-regression test case?

silene:
What kind of non-regression do you have in mind? Since Coq's minor heap is 256MB nowadays, writing a test that exercises the failure is impossible for our CI slaves. So, it would at least depend on a plugin to modify the GC settings back to a tiny minor heap beforehand.

ejgallego:
Can't you modify the size of the minor heap using an environment var?

gares:
One of the tests that failed on pkg:nix says
```
==========> TESTING success/ROmega3.v <==========
File "./success/ROmega3.v", line 34, characters 0-14:
Error: Timeout!

0m0.000s 0m0.003s
0m1.196s 0m3.588s
    success/ROmega3.v...Error! (should be accepted)
```
and The file is:
```coq
...
Lemma lemma_5833 :
 ~ 16 * v4 + (8 * v3 + (-8192 * o4 + (-4096 * s5 + (4 * v2 +
      (-4096 * o5 + (-2048 * s6 + (2 * v1 + (-2048 * o6 +
         (-1024 * s7 + (v0 + -1024 * o7)))))))))) >= 8192
\/
  16 * v4 + (8 * v3 + (-8192 * o4 + (-4096 * s5 + (4 * v2 +
      (-4096 * o5 + (-2048 * s6 + (2 * v1 + (-2048 * o6 +
         (-1024 * s7 + (v0 + -1024 * o7)))))))))) >= 1024.
Proof.
Timeout 1 romega. (* should take a few milliseconds, not seconds *)
Timeout 1 Qed. (* ditto *)
```
(there is no line 34, the file has 31 lines, but anyway...)

Disclaimer: I don't understand the code @silene is patching, but while reading the comments and looking at the test above I recalled that the `Timeout` command is implemented using signals (`SIGALRM`). So it may be a clue to understand

> if a signal is pending and we are in native code, or `caml_young_trigger`."
> 
> I have no clue why a signal would be pending there.

Good luck.

silene:
> Can't you modify the size of the minor heap using an environment var?

Good point. I didn't know that Coq was not touching the GC settings if the user had set `OCAMLRUNPARAM`.

maximedenes:
@bgregoir would you mind having a look at this patch?

Zimmi48:
@bgregoir ping?

vbgl:
IIUC, a similar bug had been present in the OCaml runtime but has been fixed already. The main question is therefore: why does this PR propose a change that is not the same as the one for the OCaml runtime?

silene:
The fix in OCaml required to modify both the compiler and the interpreter. A quick evaluation indicated that I would have to add about 400 lines of non-trivial C code to the interpreter (about ten lines per `Alloc_small`). And I am not even talking about the compiler. This was a deal breaker.

silene:
By the way, I am using `young_trigger` in the patch, but other pointers from the OCaml GC might make more sense, e.g., `young_alloc_start`.

silene:
I was wrong in my original assessment. Presumably, only the `MAKEACCU` and `ACCUMULATE` opcodes really need to be made safe, as the other ones require the attacker to be malicious. By the way, here is a way to corrupt memory (just increase `1000` to a suitably large value):

```coq
Require Import BinNat.
Fixpoint T n := match n with O => nat | S n => nat -> T n end.
Fixpoint app n : T n -> nat :=
  match n with O => fun x => x | S n => fun f => app n (f O) end.
Definition n := N.to_nat 1000.
Axiom f : T n.
Eval vm_compute in let t := (app n f, 0) in snd t.
```

Anyway, since this is incompatible with #11102, I am closing it.

PR #9917, opened 2019-04-05
[library] Move library to `vernac`

ejgallego:
As suggested by Maxime, we move library to vernac.

This PR is just a proof-of-concept and more work would be needed to be mergeable.

cc: @maximedenes


ejgallego:
It seems that moving `Declaremods` and even `Nametab` to `interp` would be feasible; but there is at least a very big hurdle which is the use of `Namegen`.

vernac/library.mllib
```diff
@@ -7,7 +7,6 @@ Global
 Decl_kinds
```
SkySkimmer: Is moving this file intended?

SkySkimmer:
I'm a bit confused, you're only moving the `library.ml` not all of `library/` right?

vernac/library.mllib
ejgallego: Yes, this is already done in one of the proof PRs.

ejgallego: Required quite a bit of untangling.

SkySkimmer: I mean moving the mllib in this PR, not decl_kinds in the future.

ejgallego: Ah no, that was mistake sorry.

ejgallego:
> I'm a bit confused, you're only moving the `library.ml` not all of `library/` right?

Yes sorry for the mllib problem; indeed I am only moving `Library`, let me amend.

ejgallego:
This PR was just a draft to illustrate how far we are; IMHO no point on keeping it open, but can still be useful for those curious.

PR #6274, opened 2017-11-28
Attempt to fix inversion disregarding singleton types (fixes #3125)

herbelin:
We introduce a notion of "truncation family" which tells if a type has been explicitly truncated to `Prop` or impredicative `Set`. In particular, singleton inductive types which lands in `Prop` by template-polymorphism, and small predicative inductive types which lands in `Set` by template-polymorphism are not considered truncated (they have all eliminations).

I believe that this fixes #3125 in a relatively intuitive way. Let's see if it raises incompatibilities though.

PR #9918, opened 2019-04-05
Fix #9294: critical bug with template polymorphism

### LINKED ISSUE #9294: Template polymorphism breaks Prop/Set separation

maximedenes:
This is adapted from #9444, and contains only the infrastructure required to fix #9294. It does not generalize universe polymorphism to `Prop` instances.

This is still work in progress, and will require some fixes.

### DIFF
--- a/doc/changelog/01-kernel/09918-unsound-template-polymorphism.rst
+++ b/doc/changelog/01-kernel/09918-unsound-template-polymorphism.rst
@@ -1,0 +1,30 @@
+- Fix soundness issue with template polymorphism (`#9294
+  <https://github.com/coq/coq/issues/9294>`_)
+
+  Declarations of template-polymorphic inductive types ignored the
+  provenance of the universes they were abstracting on and did not
+  detect if they should be greater or equal to :math:`\Set` in
+  general. Previous universes and universes introduced by the inductive
+  definition could have constraints that prevented their instantiation
+  with e.g. :math:`\Prop`, resulting in unsound instantiations later. The
+  implemented fix only allows abstraction over universes introduced by
+  the inductive declaration, and properly records all their constraints
+  by making them by default only :math:`>= \Prop`. It is also checked
+  that a template polymorphic inductive actually is polymorphic on at
+  least one universe.
+
+  This prevents inductive declarations in sections to be universe
+  polymorphic over section parameters. For a backward compatible fix,
+  simply hoist the inductive definition out of the section.
+  An alternative is to declare the inductive as universe-polymorphic and
+  cumulative in a universe-polymorphic section: all universes and
+  constraints will be properly gathered in this case.
+  See :ref:`Template-polymorphism` for a detailed exposition of the
+  rules governing template-polymorphic types.
+
+  To help users incrementally fix this issue, a command line option
+  `-no-template-check` and a global flag :flag:`Template Check` are
+  available to selectively disable the new check. Use at your own risk.
+
+  (`#9918 <https://github.com/coq/coq/pull/9918>`_, by Matthieu Sozeau
+  and Maxime Dénès).
--- a/doc/sphinx/language/cic.rst
+++ b/doc/sphinx/language/cic.rst
@@ -70,7 +70,7 @@ and function types over these sorts.
 Formally, we call :math:`\Sort` the set of sorts which is defined by:
 
 .. math::
-   
+
    \Sort \equiv \{\SProp,\Prop,\Set,\Type(i)\;|\; i~∈ ℕ\}
 
 Their properties, such as: :math:`\Prop:\Type(1)`, :math:`\Set:\Type(1)`, and
@@ -436,7 +436,7 @@ instance the identity function over a given type :math:`T` can be written
 this a *reduction* (or a *conversion*) rule we call :math:`β`:
 
 .. math::
-   
+
         E[Γ] ⊢ ((λx:T.~t)~u)~\triangleright_β~\subst{t}{x}{u}
 
 We say that :math:`\subst{t}{x}{u}` is the *β-contraction* of
@@ -474,14 +474,14 @@ with its value, that is to expand (or unfold) it into its value. This
 reduction is called δ-reduction and shows as follows.
 
 .. inference:: Delta-Local
-   
+
    \WFE{\Gamma}
    (x:=t:T) ∈ Γ
    --------------
    E[Γ] ⊢ x~\triangleright_Δ~t
 
 .. inference:: Delta-Global
-   
+
    \WFE{\Gamma}
    (c:=t:T) ∈ E
    --------------
@@ -499,7 +499,7 @@ destroyed, this reduction differs from δ-reduction. It is called
 ζ-reduction and shows as follows.
 
 .. inference:: Zeta
-   
+
    \WFE{\Gamma}
    \WTEG{u}{U}
    \WTE{\Gamma::(x:=u:U)}{t}{T}
@@ -533,17 +533,17 @@ for :math:`x` an arbitrary variable name fresh in :math:`t`.
 
    .. math::
       f ~:~ ∀ x:\Type(2),~\Type(1)
-   
+
    then
 
    .. math::
       λ x:\Type(1).~(f~x) ~:~ ∀ x:\Type(1),~\Type(1)
-   
+
    We could not allow
 
    .. math::
       λ x:\Type(1).~(f~x) ~\triangleright_η~ f
-   
+
    because the type of the reduced term :math:`∀ x:\Type(2),~\Type(1)` would not be
    convertible to the type of the original term :math:`∀ x:\Type(1),~\Type(1)`.
 
@@ -665,7 +665,7 @@ a *subtyping* relation inductively defined by:
    .. math::
       [c_1 : ∀Γ_P' ,∀ T_{1,1}' … T_{1,n_1}' ,~t'~v_{1,1}' … v_{1,m}' ;~…;~
        c_k : ∀Γ_P' ,∀ T_{k,1}' … T_{k,n_k}' ,~t'~v_{k,1}' … v_{k,m}' ]
-   
+
    respectively then
 
    .. math::
@@ -695,7 +695,7 @@ a *subtyping* relation inductively defined by:
 The conversion rule up to subtyping is now exactly:
 
 .. inference:: Conv
-   
+
    E[Γ] ⊢ U : s
    E[Γ] ⊢ t : T
    E[Γ] ⊢ T ≤_{βδιζη} U
@@ -716,13 +716,13 @@ that :math:`t_0` is :math:`λ x:T.~u_0` then one step of β-head reduction of :m
 .. math::
    λ x_1 :T_1 .~… λ x_k :T_k .~(λ x:T.~u_0~t_1 … t_n ) ~\triangleright~
    λ (x_1 :T_1 )…(x_k :T_k ).~(\subst{u_0}{x}{t_1}~t_2 … t_n )
-   
+
 Iterating the process of head reduction until the head of the reduced
 term is no more an abstraction leads to the *β-head normal form* of :math:`t`:
 
 .. math::
    t \triangleright … \triangleright λ x_1 :T_1 .~…λ x_k :T_k .~(v~u_1 … u_m )
-   
+
 where :math:`v` is not an abstraction (nor an application). Note that the head
 normal form must not be confused with the normal form since some :math:`u_i`
 can be reducible. Similar notions of head-normal forms involving δ, ι
@@ -828,7 +828,7 @@ We have to give the type of constants in a global environment :math:`E` which
 contains an inductive definition.
 
 .. inference:: Ind
-	       
+
    \WFE{Γ}
    \ind{p}{Γ_I}{Γ_C} ∈ E
    (a:A)∈Γ_I
@@ -836,7 +836,7 @@ contains an inductive definition.
    E[Γ] ⊢ a : A
 
 .. inference:: Constr
-	       
+
    \WFE{Γ}
    \ind{p}{Γ_I}{Γ_C} ∈ E
    (c:C)∈Γ_C
@@ -917,7 +917,7 @@ condition* for a constant :math:`X` in the following cases:
 + :math:`T=(X~t_1 … t_n )` and :math:`X` does not occur free in any :math:`t_i`
 + :math:`T=∀ x:U,~V` and :math:`X` occurs only strictly positively in :math:`U` and the type :math:`V`
   satisfies the positivity condition for :math:`X`.
-  
+
 Strict positivity
 +++++++++++++++++
 
@@ -931,10 +931,10 @@ cases:
   strictly positively in type :math:`V`
 + :math:`T` converts to :math:`(I~a_1 … a_m~t_1 … t_p )` where :math:`I` is the name of an
   inductive definition of the form
-  
+
   .. math::
      \ind{m}{I:A}{c_1 :∀ p_1 :P_1 ,… ∀p_m :P_m ,~C_1 ;~…;~c_n :∀ p_1 :P_1 ,… ∀p_m :P_m ,~C_n}
-     
+
   (in particular, it is
   not mutually defined and it has :math:`m` parameters) and :math:`X` does not occur in
   any of the :math:`t_i`, and the (instantiated) types of constructor
@@ -998,7 +998,7 @@ such that :math:`Γ_I` is :math:`[I_1 :∀ Γ_P ,A_1 ;~…;~I_k :∀ Γ_P ,A_k]`
    (E[Γ_I ;Γ_P ] ⊢ C_i : s_{q_i} )_{i=1… n}
    ------------------------------------------
    \WF{E;~\ind{p}{Γ_I}{Γ_C}}{}
-   
+
 
 provided that the following side conditions hold:
 
@@ -1052,30 +1052,10 @@ between universes for inductive types in the Type hierarchy.
 Template polymorphism
 +++++++++++++++++++++
 
-Inductive types can be made polymorphic over their arguments
-in :math:`\Type`.
-
-.. flag:: Auto Template Polymorphism
-
-   This option, enabled by default, makes every inductive type declared
-   at level :math:`\Type` (without annotations or hiding it behind a
-   definition) template polymorphic.
-
-   This can be prevented using the ``notemplate`` attribute.
-
-   An inductive type can be forced to be template polymorphic using the
-   ``template`` attribute.
-
-   Template polymorphism and universe polymorphism (see Chapter
-   :ref:`polymorphicuniverses`) are incompatible, so if the later is
-   enabled it will prevail over automatic template polymorphism and
-   cause an error when using the ``template`` attribute.
-
-.. warn:: Automatically declaring @ident as template polymorphic.
-
-   Warning ``auto-template`` can be used to find which types are
-   implicitly declared template polymorphic by :flag:`Auto Template
-   Polymorphism`.
+Inductive types can be made polymorphic over the universes introduced by
+their parameters in :math:`\Type`, if the minimal inferred sort of the
+inductive declarations either mention some of those parameter universes
+or is computed to be :math:`\Prop` or :math:`\Set`.
 
 If :math:`A` is an arity of some sort and :math:`s` is a sort, we write :math:`A_{/s}`
 for the arity obtained from :math:`A` by replacing its sort with :math:`s`.
@@ -1117,10 +1097,11 @@ provided that the following side conditions hold:
     + there are sorts :math:`s_i`, for :math:`1 ≤ i ≤ k` such that, for
       :math:`Γ_{I'} = [I_1 :∀ Γ_{P'} ,(A_1)_{/s_1} ;~…;~I_k :∀ Γ_{P'} ,(A_k)_{/s_k}]`
       we have :math:`(E[Γ_{I′} ;Γ_{P′}] ⊢ C_i : s_{q_i})_{i=1… n}` ;
-    + the sorts :math:`s_i` are such that all eliminations, to
-      :math:`\Prop`, :math:`\Set` and :math:`\Type(j)`, are allowed
-      (see Section :ref:`Destructors`).
-
+    + the sorts :math:`s_i` are all introduced by the inductive
+      declaration and have no universe constraints beside being greater
+      than or equal to :math:`\Prop`, and such that all
+      eliminations, to :math:`\Prop`, :math:`\Set` and :math:`\Type(j)`,
+      are allowed (see Section :ref:`Destructors`).
 
 
 Notice that if :math:`I_j~q_1 … q_r` is typable using the rules **Ind-Const** and
@@ -1141,6 +1122,61 @@ Conversion is preserved as any (partial) instance :math:`I_j~q_1 … q_r` or
 :math:`C_i~q_1 … q_r` is mapped to the names chosen in the specific instance of
 :math:`\ind{p}{Γ_I}{Γ_C}`.
 
+.. warning::
+
+   The restriction that sorts are introduced by the inductive
+   declaration prevents inductive types declared in sections to be
+   template-polymorphic on universes introduced previously in the
+   section: they cannot parameterize over the universes introduced with
+   section variables that become parameters at section closing time, as
+   these may be shared with other definitions from the same section
+   which can impose constraints on them.
+
+.. flag:: Auto Template Polymorphism
+
+   This option, enabled by default, makes every inductive type declared
+   at level :math:`\Type` (without annotations or hiding it behind a
+   definition) template polymorphic if possible.
+
+   This can be prevented using the ``notemplate`` attribute.
+
+.. warn:: Automatically declaring @ident as template polymorphic.
+
+   Warning ``auto-template`` can be used to find which types are
+   implicitly declared template polymorphic by :flag:`Auto Template
+   Polymorphism`.
+
+   An inductive type can be forced to be template polymorphic using the
+   ``template`` attribute: it should then fullfill the criterion to
+   be template polymorphic or an error is raised.
+
+.. exn:: Inductive @ident cannot be made template polymorphic.
+
+   This error is raised when the `#[universes(template)]` attribute is
+   on but the inductive cannot be made polymorphic on any universe or be
+   inferred to live in :math:`\Prop` or :math:`\Set`.
+
+   Template polymorphism and universe polymorphism (see Chapter
+   :ref:`polymorphicuniverses`) are incompatible, so if the later is
+   enabled it will prevail over automatic template polymorphism and
+   cause an error when using the ``template`` attribute.
+
+.. flag:: Template Check
+
+   Unsetting option :flag:`Template Check` disables the check of
+   locality of the sorts when abstracting the inductive over its
+   parameters. This is a deprecated and *unsafe* flag that can introduce
+   inconsistencies, it is only meant to help users incrementally update
+   code from Coq versions < 8.10 which did not implement this check.
+   The `Coq89.v` compatibility file sets this flag globally. A global
+   ``-no-template-check`` command line option is also available. Use at
+   your own risk. Use of this flag is recorded in the typing flags
+   associated to a definition but is *not* supported by the |Coq|
+   checker (`coqchk`). It will appear in :g:`Print Assumptions` and
+   :g:`About @ident` output involving inductive declarations that were
+   (potentially unsoundly) assumed to be template polymorphic.
+
+
 In practice, the rule **Ind-Family** is used by |Coq| only when all the
 inductive types of the inductive definition are declared with an arity
 whose sort is in the Type hierarchy. Then, the polymorphism is over
@@ -1154,10 +1190,10 @@ inductive type is set in :math:`\Set` (even in case :math:`\Set` is impredicativ
 Section The-Calculus-of-Inductive-Construction-with-impredicative-Set_),
 and otherwise in the Type hierarchy.
 
-Note that the side-condition about allowed elimination sorts in the
-rule **Ind-Family** is just to avoid to recompute the allowed elimination
-sorts at each instance of a pattern matching (see Section :ref:`Destructors`). As
-an example, let us consider the following definition:
+Note that the side-condition about allowed elimination sorts in the rule
+**Ind-Family** avoids to recompute the allowed elimination sorts at each
+instance of a pattern matching (see Section :ref:`Destructors`). As an
+example, let us consider the following definition:
 
 .. example::
 
@@ -1320,7 +1356,7 @@ using the syntax:
    \Match~m~\as~x~\In~I~\_~a~\return~P~\with~
    (c_1~x_{11} ... x_{1p_1} ) ⇒ f_1 | …
    | (c_n~x_{n1} ... x_{np_n} ) ⇒ f_n~\kwend
-   
+
 The :math:`\as` part can be omitted if either the result type does not depend
 on :math:`m` (non-dependent elimination) or :math:`m` is a variable (in this case, :math:`m`
 can occur in :math:`P` where it is considered a bound variable). The :math:`\In` part
@@ -1360,7 +1396,7 @@ There is no restriction on the sort of the predicate to be eliminated.
    -----------------------
    [I:∀ x:A,~A′|∀ x:A,~B′]
 
-   
+
 .. inference:: Set & Type
 
    s_1 ∈ \{\Set,\Type(j)\}
@@ -1376,7 +1412,7 @@ is also of sort :math:`\Prop` or is of the morally smaller sort
 :math:`\SProp`.
 
 .. inference:: Prop
-	       
+
    s ∈ \{\SProp,\Prop\}
    --------------------
    [I:\Prop|I→s]
@@ -1404,7 +1440,7 @@ the proof of :g:`or A B` is not accepted:
 
       Fail Definition choice (A B: Prop) (x:or A B) :=
       match x with or_introl _ _ a => true | or_intror _ _ b => false end.
-   
+
 From the computational point of view, the structure of the proof of
 :g:`(or A B)` in this term is needed for computing the boolean value.
 
@@ -1441,7 +1477,7 @@ this type.
 :math:`\Prop` for which more eliminations are allowed.
 
 .. inference:: Prop-extended
-	       
+
    I~\kw{is an empty or singleton definition}
    s ∈ \Sort
    -------------------------------------
@@ -1589,7 +1625,7 @@ An ι-redex is a term of the following form:
 
 .. math::
    \case((c_{p_i}~q_1 … q_r~a_1 … a_m ),P,f_1 |… |f_l )
-   
+
 with :math:`c_{p_i}` the :math:`i`-th constructor of the inductive type :math:`I` with :math:`r`
 parameters.
 
@@ -1636,7 +1672,7 @@ Typing rule
 The typing rule is the expected one for a fixpoint.
 
 .. inference:: Fix
-	       
+
    (E[Γ] ⊢ A_i : s_i )_{i=1… n}
    (E[Γ;~f_1 :A_1 ;~…;~f_n :A_n ] ⊢ t_i : A_i )_{i=1… n}
    -------------------------------------------------------
@@ -1749,7 +1785,7 @@ The reduction for fixpoints is:
 
 .. math::
    (\Fix~f_i \{F\}~a_1 …a_{k_i}) ~\triangleright_ι~ \subst{t_i}{f_k}{\Fix~f_k \{F\}}_{k=1… n} ~a_1 … a_{k_i}
-   
+
 when :math:`a_{k_i}` starts with a constructor. This last restriction is needed
 in order to keep strong normalization and corresponds to the reduction
 for primitive recursive operators. The following reductions are now
@@ -1808,11 +1844,11 @@ and :math:`\subst{E}{|Γ|}{|Γ|c}` to mean the parallel substitution
         {\WF{E;~c:U;~E′;~c′:=λ x:U.~\subst{t}{c}{x}:∀x:U,~\subst{T}{c}{x};~\subst{E″}{c′}{(c′~c)}}
         {\subst{Γ}{c′}{(c′~c)}}}
 
-   
+
 .. math::
    \frac{\WF{E;~c:U;~E′;~c′:T;~E″}{Γ}}
         {\WF{E;~c:U;~E′;~c′:∀ x:U,~\subst{T}{c}{x};~\subst{E″}{c′}{(c′~c)}}{\subst{Γ}{c′}{(c′~c)}}}
-	
+
 .. math::
    \frac{\WF{E;~c:U;~E′;~\ind{p}{Γ_I}{Γ_C};~E″}{Γ}}
         {\WFTWOLINES{E;~c:U;~E′;~\ind{p+1}{∀ x:U,~\subst{Γ_I}{c}{x}}{∀ x:U,~\subst{Γ_C}{c}{x}};~
@@ -1853,7 +1889,7 @@ One can consequently derive the following property.
 .. _First-pruning-property:
 
 .. inference:: First pruning property:
-	       
+
    \WF{E;~c:U;~E′}{Γ}
    c~\kw{does not occur in}~E′~\kw{and}~Γ
    --------------------------------------
@@ -1933,5 +1969,3 @@ impredicative system for sort :math:`\Set` become:
    s ∈ \{\Type(i)\}
    ----------------
    [I:\Set|I→ s]
-
-
--- a/kernel/environ.mli
+++ b/kernel/environ.mli
@@ -51,8 +51,9 @@ type globals
 
 type stratification = {
   env_universes : UGraph.t;
-  env_engagement : engagement;
   env_sprop_allowed : bool;
+  env_universes_lbound : Univ.Level.t;
+  env_engagement : engagement
 }
 
 type named_context_val = private {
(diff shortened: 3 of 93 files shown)


pretyping/pretyping.ml
```diff
@@ -419,7 +419,10 @@ let interp_instance ?loc evd l =
          (evd, l :: univs)) (evd, [])
       l
   in
-  if List.exists (fun l -> Univ.Level.is_prop l) l' then
+  let lbound = Global.universes_lbound () in
+  if List.exists (fun l -> Univ.Level.is_prop l &&
+                           Evd.check_leq evd (Univ.Universe.super (Univ.Universe.make l))
+                             (Univ.Universe.make lbound)) l' then
```
mattam82: That looks a bit suspicious. We should still always disallow prop in instances, so I'd come back to the earlier version.

maximedenes:
With the fix you suggested, here are the tests that are failing:
```
    success/name_mangling.v...could not be checked (Error!)
    success/name_mangling.v...Error! (should be accepted)
    success/primitiveproj.v...could not be checked (Error!)
    success/primitiveproj.v...Error! (should be accepted)
    success/Record.v...could not be checked (Error!)
    success/Record.v...Error! (should be accepted)
    success/sprop_hcons.v...could not be checked (Error!)
    success/sprop_hcons.v...Error! (should be accepted)
    success/sprop.v...could not be checked (Error!)
    success/sprop.v...Error! (should be accepted)
    bugs/closed/bug_3125.v...Error! (bug seems to be opened, please check)
    bugs/closed/bug_4622.v...Error! (bug seems to be opened, please check)
    bugs/closed/bug_5281.v...Error! (bug seems to be opened, please check)
    bugs/closed/bug_sprop_14.v...Error! (bug seems to be opened, please check)
    bugs/closed/HoTT_coq_093.v...Error! (bug seems to be opened, please check)
    output/NumeralNotations.v...Error! (unexpected output)
    output/UnivBinders.v...Error! (unexpected output)
```

The output tests failures are due to more constraints being printed.

vernac/comInductive.ml
```diff
@@ -255,7 +258,7 @@ let solve_constraints_system levels level_bounds =
   done;
   v
 
-let inductive_levels env evd poly arities inds =
+let inductive_levels env evd arities inds =
   let destarities = List.map (fun x -> x, Reduction.dest_arity env x) arities in
   let levels = List.map (fun (x,(ctx,a)) ->
     if Sorts.is_prop a || Sorts.is_sprop a then None
```
mattam82: We should not do the "is_flexible_sort" stuff that handles "natural" props in the template_poly case, as that turns inductives in Prop when they were in Type@{i} for any i I think. That is why 3125 fails I think.

maximedenes: I'm a bit lost. Wasn't the `is_flexible_sort` stuff made precisely for template polymorphic inductives, that didn't really need to be?

mattam82: Nope, I think the is_flexible_sort stuff is rather for inferring the sort of "natural" propositions in Prop, but template polymorphic ones should stay in Type@{i} I believe. At least that's how I understand 3125.

Zimmi48:
Is this on purpose that you put it in 8.10+beta1 rather than in 8.9.1?

maximedenes:
> Is this on purpose that you put it in 8.10+beta1 rather than in 8.9.1?

I think it is going to be a bit painful to backport, but we can try.

pretyping/pretyping.ml
JasonGross: Why should we disallow prop in instances?

mattam82:
Alright. We were just missing the same thing for records as inductives. Now the test-suite passes almost entirely, we're down to sprop universe inconsistencies I don't understand (!).

mattam82:
I was guilty for sprop, and also for a minor change in the output tests for universes, both fixed now. This passes the test-suite. Now let's see if the more liberal elaboration constraints for template-polymorphic inductive declarations (anonymous universes can be instantiated with Prop) change anything in the CI.

mattam82:
So, unsurprisingly, people use:

```
Section Foo.
  Context (A : Type).

  blah blah doing stuff with A that shouldn't allow it to be `True`.

  Ind I : Type := ...
End Foo.
```
And they expect `I` to be template-polymorphic on `A`'s universe, which I don't allow anymore. Avoiding the problematic case is a matter of defining the inductive separately. Don't know if that's easy to do in all examples.

mattam82:
For paramcoq, the issue is that the relation for an inductive reuses the same universes as the inductive, so when checking that it is template none of the universe variables are seen as introduced by the inductive, so template-polymorphism does not apply anymore. Solution: refresh the universes.

For vst, the issue is that a failure of template-polymorphism, due to a definition of an inductive in a section in CompCert `cfrontend/Ctypes.v#fundef`, changes an automatic hyp naming (because something which was in Set turns into Type, so `H5` becomes `X`). Assigning the right name fixes it, as well as modifying the type definition in CompCert to be done outside a section.

For bedrock2, it's a definition of vectors in a section, changing hyp naming again. (Maybe other issues as well).

For ci-fiat-crypto-legacy, I'll see when it finished compiling...

SkySkimmer:
For fiat-crypto legacy it's probably this bit https://github.com/mit-plv/fiat-crypto/blob/847b18bf8f38f7262331181f25ca96aee0423f52/src/Compilers/Syntax.v#L14-L16 inductive in a section later tries to be applied to a Set.

mattam82:
I tried many things to treat template-polymorphism in sections but nothing satisfactory can be done: to be really safe we would have to elaborate everything in sections with `Prop <= i` constraints and do global checks at discharge time if some universe we're parameterizing on has been constrained elsewhere (good luck reporting a nice error there): in short that's ugly and too big a change. One halfway solution could be to have a typing_flag to govern the use of the stricter check that is implemented in this PR without delaying the release any further. @maximedenes @ppedrot @vbgl @herbelin what's your opinion on that solution?

Note that this all points to providing a clean solution to *sort* polymorphism (think PTS parameterization) and not just *universe-level* polymorphism, to support template-polymorphic style lowering to Prop but also multiple universe hierachies. This would also help improve elaboration of Prop/SProp by introducing a proper notion of sort variable that can be instantiated with impredicative or predicative universes (the later being further parameterized by a level). I'm planning to write up a CEP on that in the future (after the POPL deadline).

vbgl:
Just to be sure: disabling the stricter check — using the typing flag you propose — is unsound (but convenient in some cases).

mattam82:
Yes, disabling it means you can write unsound instances, just like with type-in-type.

vbgl:
Thanks. Then what you propose sounds like a good plan to me.

vbgl:
@mattam82 Any news regarding the implementation of your plan? It would be nice if we could at last release version 8.10.

mattam82:
I’m sorry I got really busy with something else. I’ll get on it today and tomorrow

mattam82:
Let's see where this gets up. New typing flag "-no-template-check" will have to be used on some contribs (e.g. parametricity)

maximedenes:
> Avoiding the problematic case is a matter of defining the inductive separately. Don't know if that's easy to do in all examples.

Did fixing the problematic developments  turn out to be very difficult?

I'm not a big fan of introducing a flag for compatibility with a critical bug. What would be the value of a formal proof checked with the flag?

mattam82:
I’m not a big fan either but I don’t have the time to fix fiat-parsers and parancoq and neither to implement the better solutions of using cumulative polymorphic inductives. There are cases that are not easily portable, e.g when using definitions hiding constraints on a section variable that requires moving also these definitions to polymorphic ones.

mattam82:
Without the check, one can still manually ensure if a template inductive can rightfully be lowered, but as the CI shows it’s (thankfully) rarely needed. Note that for paramcoq there is a sound solution which is to refresh universe names when creating the relations for template-polymorphic inductives. That one is doable.

maximedenes:
> I don’t have the time to fix fiat-parsers and parancoq and neither to implement the better solutions of using cumulative polymorphic inductives

Oh sure, I don't think you should be left alone dealing with this. My point was that it is more or less expected that developments relying on bugs (and even more so for critical ones) are expected to be broken by fixes. CI testing is mainly for discovering regressions early, it is not meant to prevent applying fixes.

Of course, we should collectively do our best to help downstream authors port their developments to fixed versions of Coq.

So I guess it depends what the purpose of the flag is. Is it to make (incremental) porting easier? If so, maybe we should deprecate it as we introduce it, and remove it from the next version. Also, we should not forget to open issues on the developments which require the flag and coordinate with their authors.

mattam82:
Yes, that’s the idea, to give an option to help users fix their developments incrementally. It should be deprecated only if we know that we have a proper solution that allows instantiations of what I would call sort-polymorphic constructions. My plan is to work on that in the coming months, and it has links with other efforts: treating Prop/SProp correctly during elaboration and supporting multiple sort hierarchies for Pédrot and Tabareau’s effectful type theories. That’s however a non-trivial change so it’s hard to evaluate the timeline for this without at least prototyping it.

kernel/environ.mli
```diff
@@ -51,8 +51,9 @@ type globals
 
 type stratification = {
   env_universes : UGraph.t;
-  env_engagement : engagement;
   env_sprop_allowed : bool;
+  env_universes_lbound : Univ.Level.t;
```
SkySkimmer: Should it be here or would it make more sense in the UGraph.t?

mattam82: Not sure, we set it dynamically to elaborate template inductives with `Prop <= i` by default for new universes. The graph itself is pretty agnostic to it (we just insert different constraints in it).

mattam82:
Well, that worked!

vbgl:
Thanks Matthieu for addressing this issue. It seems that only a bit of documentation is missing.

mattam82:
I’ll have that done today. I think we should make it clear in the announcement. Is there an .md / .rst file I can add a long example to?

SkySkimmer:
Around https://github.com/coq/coq/blob/master/doc/sphinx/language/cic.rst#template-polymorphism I guess

mattam82:
Thanks. I don't know where to look for the beautified version of my changelog entry, it didn't appear magically when making `make -f Makefile.dune refman-html`

vbgl:
The change log entry did not show up because the name of the file holding said entry should en with `.rst`. I’ve just renamed it.

I’ve also fixed a minor markup issue that broke the CI build.

mattam82:
I've copied from `10439-uniform-opaque-seff-handling.md`, so that one should probably be changed to a `.rst` as well.

mattam82:
I guess the OOM on fiat-crypto is unrelated.

vbgl:
> I've copied from `10439-uniform-opaque-seff-handling.md`, so that one should probably be changed to a `.rst` as well.

Or the program that builds the changelog from these files may be more lenient on the file names. Théo, what do you think? (cc @Zimmi48).

Zimmi48:
> > I've copied from `10439-uniform-opaque-seff-handling.md`, so that one should probably be changed to a `.rst` as well.
> 
> Or the program that builds the changelog from these files may be more lenient on the file names. Théo, what do you think? (cc @Zimmi48).

Well, `rst` and `md` do not bear the same format, so that would not make sense, unless we do some translation between formats which would require a dependency such as pandoc, so no.
Short-term solution would be for the build to catch that there are unknown extensions in the folder and to fail because of this (but beware of `~` and other common hidden files).
Long-term solution is probably going to be to adopt a more structured (custom) format for these files anyway, given the high-rate of errors that developers make when writing them...

Zimmi48:
I had reviewed #10439 and did not even notice the wrong extension, which proves that a CI check is actually needed for this.

vbgl:
Note that I did not suggest to accept any file format. Just any file name.

doc/changelog/01-kernel/09918-unsound-template-polymorphism.rst
```diff
@@ -0,0 +1,25 @@
+- Declarations of template-polymorphic inductive types ignored the
+  provenance of the universes they were abstracting on and did not
+  detect if they should be greater or equal to Set in general. Previous
```
Zimmi48: Could be:

```suggestion
  detect if they should be greater or equal to :g:`Set` in general. Previous
```

or

```suggestion
  detect if they should be greater or equal to :math:`\Set` in general. Previous
```

doc/changelog/01-kernel/09918-unsound-template-polymorphism.rst
```diff
@@ -0,0 +1,25 @@
...
+  polymorphic over section parameters. For a backward compatible fix,
+  simply hoist the inductive definition out of the section.
+  An alternative is to declare the inductive as universe-polymorphic and
+  cumulative in a universe-polymorphic section: all universes and
+  constraints will be properly gathered in this case.
+  See :ref:`Template-polymorphism` for a detailed exposition of the
+  rules governing template-polymorphic types.
+
+  To help users incrementally fix this issue, a command line option
+  `-no-template-check` and a global option `Unset Template Check` are
```
Zimmi48:
```suggestion
  `-no-template-check` and a global flag :flag:`Template Check` are
```

doc/sphinx/language/cic.rst
```diff
@@ -1117,10 +1117,11 @@ provided that the following side conditions hold:
     + there are sorts :math:`s_i`, for :math:`1 ≤ i ≤ k` such that, for
       :math:`Γ_{I'} = [I_1 :∀ Γ_{P'} ,(A_1)_{/s_1} ;~…;~I_k :∀ Γ_{P'} ,(A_k)_{/s_k}]`
       we have :math:`(E[Γ_{I′} ;Γ_{P′}] ⊢ C_i : s_{q_i})_{i=1… n}` ;
-    + the sorts :math:`s_i` are such that all eliminations, to
-      :math:`\Prop`, :math:`\Set` and :math:`\Type(j)`, are allowed
-      (see Section :ref:`Destructors`).
-
+    + the sorts :math:`s_i` are all introduced by the inductive
+      declaration and have no universe constraints beside being greater
+      than or equal to :math:`Prop`, and such that all
```
Zimmi48:
```suggestion
      than or equal to :math:`\Prop`, and such that all
```

doc/sphinx/language/cic.rst
```diff
@@ -1141,6 +1142,26 @@ Conversion is preserved as any (partial) instance :math:`I_j~q_1 … q_r` or
...
+.. flag:: Template Check
+
+   Unsetting option :flag:`Template Check` disables the check of
+   locality of the sorts when abstracting the inductive over its
+   parameters. This is an *unsafe* flag that can introduce
+   inconsistencies, it is only meant to help users incrementally update
+   code from Coq versions < 8.10 which did not implement this check. A
+   global ``-no-template-check`` command line option is also
+   available. Use at your own risk. Use of this flag is recorded in the
+   typing flags associated to a definition.
```
Zimmi48: And thus displayed by the `Print Assumption` command, right? If so, probably worth to say it explicitly.

Zimmi48 (commented):
Should probably be documented in `dev/doc/critical-bugs` as well.

BTW, a more general comment about the proposed fix: first, congrats on explaining the issue at stake very well, second since IIUC no inductive declaration inside a section will be template-polymorphic anymore (unless the unsafe flag is set), what about emitting a new (documented of course) warning similar to the one that @SkySkimmer introduced about template polymorphic declarations (but which was silenced by default)? This warning would say: this inductive declaration is not template polymorphic because it is inside a section, and then would go on by proposing the three safe options of setting the no-template attribute to silence the warning, making the inductive full universe polymorphic, or moving it out of the section to keep it template polymorphic.

Zimmi48:
> Note that I did not suggest to accept any file format. Just any file name.

Sorry, I don't get it.

maximedenes:
@vbgl was not suggesting to accept markdown files, just to accept any name for an rst file. No pandoc needed.

Zimmi48:
> @vbgl was not suggesting to accept markdown files, just to accept any name for an rst file. No pandoc needed.

Well, why not. Although, that would be a bit weird if we start to have many `md` files which actually contain Sphinx format.

Zimmi48:
@mattam82: the documentation says the name of the flag is `Template Check` but the one that is actually defined is `Template Polymorphic Inductive Check`.

doc/sphinx/language/cic.rst
Zimmi48: Actually it doesn't look like `Print Assumptions` print anything. Shouldn't it?

Zimmi48:
This code should fail: the explitcit attribute shouldn't be silently ignored:

```
Polymorphic Definition f@{i} : Type@{i} := nat.
Polymorphic Definition baz@{i} : Type@{i} -> Type@{i} := fun x => x.

Section Foo.
  Universe u.
  Context (A : Type@{u}).

  #[universes(template)] Inductive Bar :=
  | bar : A -> Bar.

End Foo.
```

Zimmi48:
This code should not fail, since the inductive is actually not declared template polymorphic, i.e. the warning should not be emitted:

```
Set Warnings "+auto-template".
Section Foo.
  Universe u.
  Context (A : Type@{u}).

  Inductive Bar :=
  | bar : A -> Bar.

End Foo.
```

kernel/indTyping.mli
```diff
@@ -33,3 +34,10 @@ val typecheck_inductive : env -> mutual_inductive_entry ->
      (Constr.rel_context * (Constr.rel_context * Constr.types) array) *
      Sorts.family)
     array
+
+type univ_info = { ind_squashed : bool; ind_has_relevant_arg : bool;
+                   ind_min_univ : Universe.t option; (* Some for template *)
+                   ind_univ : Universe.t }
+
+val check_univ_leq :  ?is_real_arg:bool ->
```
SkySkimmer: Why's this stuff exported?

kernel/indTyping.ml
```diff
@@ -236,22 +236,43 @@ let allowed_sorts {ind_squashed;ind_univ;ind_min_univ=_;ind_has_relevant_arg=_}
   if not ind_squashed then InType
   else Sorts.family (Sorts.sort_of_univ ind_univ)
 
+(* For a level to be template polymorphic, it must be introduced
+   by the definition (so have no constraint except lbound <= l)
+   and not to be constrained from below, so any universe l' <= l
+   can be used as an instance of l. All bounds from above, i.e.
+   l <=/< r will be valid for any l' <= l. *)
+let unbounded_from_below u cstrs =
+  Univ.Constraint.for_all (fun (l, d, r) ->
```
SkySkimmer: Can't we ask the ugraph instead of folding over constraints?

JasonGross (dismissed):
Please update the Compat/Coq810.v file (or whichever one it should be) to `Global Unset Template Check` (or whatever the option is called)

doc/sphinx/language/cic.rst
mattam82: Indeed it should, the accessors to the use of the template-check are there already.

kernel/indTyping.mli
mattam82: Good point, remnant of debugging.

vernac/vernacentries.ml
```diff
@@ -607,6 +607,24 @@ let vernac_assumption ~atts discharge kind l nl =
...
+let is_template_check () =
+  let typing_flags = Environ.typing_flags (Global.env ()) in
+  typing_flags.Declarations.check_template
+
+let () =
+  let tccheck =
+    { optdepr = true;
+      optname = "Template universe check";
+      optkey = ["Template"; "Polymorphic"; "Inductive"; "Check"];
+      optread = (fun () -> is_template_check ());
```
mattam82: What do you guys prefer? The longer name is more explicit... but maybe overkill as we don't expect many users and this is already deprecated?

kernel/indTyping.ml
```diff
@@ -286,10 +307,14 @@ let typecheck_inductive env (mie:mutual_inductive_entry) =
...
 
+  let has_template_poly = List.exists (fun oie -> oie.mind_entry_template) mie.mind_entry_inds in
+
   (* universes *)
   let env_univs =
     match mie.mind_entry_universes with
-    | Monomorphic_entry ctx -> push_context_set ctx env
+    | Monomorphic_entry ctx ->
+      let env = if has_template_poly then set_universes_lbound env Univ.Level.prop else env in
+      push_context_set ctx env
```
SkySkimmer: push_context_set catches the AlreadyDeclared exception, given that afterwards we assume the universes in ctx are new this is problematic.
We hope to someday stop catching this exception but sideeffect intricacies prevent it for now, so I guess adding an optional argument `?enforce_fresh` (defaulting with current behaviour) should be added to push_context_set

vernac/vernacentries.ml
Zimmi48: I prefer the shorter name.

mattam82:
> This code should fail: the explitcit attribute shouldn't be silently ignored:
> 
> ```
> Section Foo.
>   Universe u.
>   Context (A : Type@{u}).
> 
>   #[universes(template)] Inductive Bar :=
>   | bar : A -> Bar.
> 
> End Foo.
> ```

I don't think so, it might take other parameters for which template polymorphism would be fine.

mattam82:
> This code should not fail, since the inductive is actually not declared template polymorphic, i.e. the warning should not be emitted:
> 
> ```
> Set Warnings "+auto-template".
> Section Foo.
>   Universe u.
>   Context (A : Type@{u}).
> 
>   Inductive Bar :=
>   | bar : A -> Bar.
> 
> End Foo.
> ```

Same thing, it might be template polymorphic on other parameters. I will try to refine the code to not declare an inductive as template polymorphic if the inductive is template polymorphic on no universes after the check.

doc/changelog/01-kernel/09918-unsound-template-polymorphism.rst
```diff
@@ -0,0 +1,25 @@
+- Declarations of template-polymorphic inductive types ignored the
+  provenance of the universes they were abstracting on and did not
+  detect if they should be greater or equal to :math:`Set` in
```
Zimmi48:
```suggestion
  detect if they should be greater or equal to :math:`\Set` in
```

doc/changelog/01-kernel/09918-unsound-template-polymorphism.rst
```diff
@@ -0,0 +1,25 @@
...
+  general. Previous universes and universes introduced by the inductive
+  definition could have constraints that prevented their instantiation
+  with e.g. :math:`Prop`, resulting in unsound instantiations later. The
```
Zimmi48:
```suggestion
  with e.g. :math:`\Prop`, resulting in unsound instantiations later. The
```

doc/sphinx/language/cic.rst
```diff
@@ -1141,6 +1142,26 @@ Conversion is preserved as any (partial) instance :math:`I_j~q_1 … q_r` or
 :math:`C_i~q_1 … q_r` is mapped to the names chosen in the specific instance of
 :math:`\ind{p}{Γ_I}{Γ_C}`.
 
+.. warning::
+
+   The restriction that sorts are introduced by the inductive
+   declaration prevents inductive types declared in sections to
+   be template-polymorphic: they cannot parameterize over their
```
Zimmi48: This is not specific enough apparently, should be:

>prevents inductive types declared in sections to be template-polymorphic on section universes

or something along those lines.

doc/sphinx/language/cic.rst
Zimmi48: OK so now that `Print Assumptions` works, it should be mentioned in this description. Also, the flag is already deprecated, so it should appear in the description with the `.. deprecated:: 8.10` directive. BTW, what will coqchk say if this flag was used? Something similar to `Print Assumptions`, I suppose?

toplevel/usage.ml
```diff
@@ -82,6 +82,7 @@ let print_usage_common co command =
 \n  -sprop-cumulative      make sort SProp cumulative with the rest of the hierarchy\
 \n  -indices-matter        levels of indices (and nonuniform parameters) contribute to the level of inductives\
 \n  -type-in-type          disable universe consistency checking\
+\n  -no-template-check     disable checking of universes constraints on universes parameterizing template polymorphic inductive types\
```
Zimmi48: Should probably be documented as well in the chapter on the Coq commands.

Zimmi48 (commented):
> > This code should fail: the explitcit attribute shouldn't be silently ignored:
> > ```
> > Section Foo.
> >   Universe u.
> >   Context (A : Type@{u}).
> > 
> >   #[universes(template)] Inductive Bar :=
> >   | bar : A -> Bar.
> > 
> > End Foo.
> > ```
> 
> I don't think so, it might take other parameters for which template polymorphism would be fine.

Alright.

mattam82:
This is not the end of the story yet, I have some more work to do implementing the refined version where ˋuniverses(template)ˋ fails when there’s no template universe polymorphism to apply (which makes complete sense indeed). Bear with me...

mattam82:
> Please update the Compat/Coq810.v file (or whichever one it should be) to `Global Unset Template Check` (or whatever the option is called)

You mean `Coq89` right?

mattam82:
Let's hope for a nice CI. I had to change a lot of things to actually make sense of universes(template): now we print on which universes an inductive definition actually is polymorphic, and we can't be polymorphic for nothing (e.g. if the universes in the universe context don't actually appear in the conclusion). This prevents the attribute being ignored, making a user think his definition was meaningfully polymorphic when it's actually not doing anything. There is a tricky issue with the fact that one can have singletons that get template polymorphic even if they have no parameters, as long as at least one universe they introduce appears in its conclusion type, this is preserved: `#[universes(template)] Inductive foo := C.` gives foo the `Prop` sort which in this case means that foo can actually live in Prop or Type for any `i`. It's debatable why we want singletons like this to be template polymorphic (which will mess with extraction) while they could just be in `Prop` and get erased, but #6274 went into that direction of distinguishing the two so I followed it here.

This resulted in removing a lot of useless attributes in the stdlib. I also modified some files to use template-polymorphism where it seemed natural to keep it, e.g. micromegas' PExpr and Pol, FSets trees (although by the time we use functors template polymorphism is lost), and some generic definitions in ssr (mainly phantom types, although again they could certainly be left in Prop).

One remaining issue is that I don't know how to test the `Unset Template Check` flag which is not supported by the checker. Do we have a story for that?

mattam82:
Oh and all contribs where we blindly added the attribute whereas things were actually not needing template polymorphism are going to fail...

mattam82:
Well, that’s a relief, I will fix those tomorrow along with surely more minor things I missed. Given the non-trivial size of the fix in the end, it might be good to make an rc or last beta before the final.

SkySkimmer:
>One remaining issue is that I don't know how to test the Unset Template Check flag which is not supported by the checker. Do we have a story for that?

Make the checker support it? We're going in the direction of having the checker understand local typing_flags in #10291 already.

mattam82:
There we have it. I don’t think we should make the checker support that, it’s only meant as a transitional option for migration and not here to stay. Probably this choice should be documented though, I’ll add that along with doc on the changes of printing and the semantics of the template attribute being more strict. It’s kind of reassuring to see that not that much code relies on inductives defined in sections to be template polymorphic. In many cases I guess the inductives can not be usefully polymorphic in the end as soon as one uses constants or other things that make indirect references to the parameter universes, due to the highly syntactic nature of the criterion.

mattam82:
I rebased and cleaned up everything, and the doc and changelog's have been updated. I think it's ready now. In the end not too many overlays which is nice!

kernel/indTyping.ml
mattam82: You’re right I didn’t catch that one

mattam82:
Regarding @SkySkimmer's comment, this should now be now fixed in #10616 I'd like to get this one merged first as it's already quite big.

vbgl:
Thank you all for working on this issue.

Matthieu, would you mind opening PRs corresponding to the overlays? Also, the patch for VST seems backward compatible; is it?

Who should be the assignee?

mattam82:
Yep, the VST one is backwards compatible, PRs open now. Please don't hesitate to take over and finis this, I'm won't be available anymore this week.

mattam82:
I guess @SkySkimmer or @ppedrot ?

Zimmi48:
Sorry for thinking about this only now but I was triggered by the overlay on aac-tactic which is not backward compatible because it uses `-no-template-check`, whereas it would if it used `-set "Template Check=false"`. Wouldn't it make sense given that the option is already deprecated and the command-line flag is not there to stay either to *not* introduce a new command-line flag but recommend the use of the new generic `-set` flag instead? Of course, I don't want to be a pain in the neck with such nitpicking so feel free to ignore this comment.

ppedrot:
@mattam82 I'd prefer not to assign, expect maybe only for merging. I still have no clue about template polymorphism...

mattam82:
Maybe @herbelin then?

vbgl:
Gaëtan, do you think you’ll be able to review this PR? Thanks.

In order to correctly backport these changes, I would appreciate to get confirmation that the new `Record.check_anonymous_type` function is correctly implemented as follows. Thanks.

~~~
let check_anonymous_type ind =
  match ind with
  | { CAst.v = CSort (Glob_term.GType []) } -> true
  | _ -> false
~~~

vernac/record.ml
```diff
@@ -111,15 +122,15 @@ let typecheck_params_and_fields finite def poly pl ps records =
...
               Loc.raise ?loc (Stream.Error "pattern with quote not allowed in record parameters")) ps
-  in 
+  in
   let sigma, (impls_env, ((env1,newps), imps)) = interp_context_evars ~program_mode:false env0 sigma ps in
   let fold (sigma, template) (_, t, _, _) = match t with
-    | Some t -> 
+    | Some t ->
        let env = EConstr.push_rel_context newps env0 in
        let poly =
          match t with
```
SkySkimmer: Just below this line (github doesn't let me comment on the right line) we check that `t` is a bare `Type`, this should be reusing `check_anonymous_type` (or perhaps is redundant with the new use site of `check_anonymous_type`)

SkySkimmer:
The commit history looks a bit messy, is it going to be cleaned up or should I just review the PR as 1 big change?

>In order to correctly backport these changes, I would appreciate to get confirmation that the new Record.check_anonymous_type function is correctly implemented as follows. Thanks.

Looks fine, as I said in the other comment it should be doing the same thing that we already do on https://github.com/coq/coq/blob/09002e0c20cf4da9cbb1e27146ae1fdd205b304a/vernac/record.ml#L126 ie https://github.com/coq/coq/blob/v8.10/vernac/record.ml#L127

mattam82:
I think we should just squash everything, I tried to separate the commits by files touched but that won’t compile

mattam82:
Indeed we should factorize record/comInd more

vbgl:
Matthieu, just to clarify your last comment: do you intend to do that factorization as part of this PR?

mattam82:
No !

mattam82:
It would make rebasing even worse I guess

vbgl:
Then shall we merge this as is? Gaëtan, do you approve these changes?

vbgl:
There are now conflicts, as #10291 has been merged. This will make the port to the 8.10 branch a bit more tricky.

Shall this PR be merged?

mattam82:
I can give it a go this morning (for sure).

herbelin (commented):
Not knowing exactly if some action is (still) expected from me, but considering that noone formally approved yet, I started to look at the PR.

I find it a bit a pity to have a new option for disabling template checking and to have template polymorphism not supported anymore in sections but, at the stage of understanding I am, my feeling is that the PR otherwise reaches its goal, namely to fix #9294 in a safe way, while at the same time making some steps towards eventually reducing template polymorphism to universe polymorphism. As such, it is mergeable.

Specifically about sections (late reaction, sorry):

> I tried many things to treat template-polymorphism in sections but nothing satisfactory can be done: to be really safe we would have to elaborate everything in sections with `Prop <= i` constraints and do global checks at discharge time if some universe we're parameterizing on has been constrained elsewhere

Thinking at this, my conclusion is that the names used in the type of the arguments of a template polymorphic inductive type are actually _binding_ names, even though we use global names to represent them (at the time when template polymorphism was introduced there were no universe variables, so there were simulated using global names). So, in a configuration like:
```
Section S.
Universe u.
Variable A : Type@{u}.
Inductive box (A:Type@{u}) := Box : A -> box.
Definition B := Set : Type@{u}.
End S.
```
The constraint on `u` generated by the definition `B` has no effect on `box`. Outside the section, `box` should be:
```
Inductive box (A:Type@{i}) : Type@{i} := Box : A -> box.
 ```
or even, to adopt the notational convention of universe polymorphism:
```
Inductive box@{i} (A:Type@{i}) : Type@{i} := Box : A -> box.
```
with `i` a binding variable. In particular, when `box` is applied to `T:Type@{v}`, the name `u` (here renamed `i`) is just instantiated as a standard binding variable to give type `Type@{v}`.

The only remaining effect of the global name `u` on template polymorphism is in case `box` is unapplied and the absence of an argument does not allow to infer an instance. In this case, it is the global `u` which is used as default instance, i.e. `box` gets instantiated as `box@{u}` (otherwise said, the binding variable `u` gets instantiated by the eponym global variable `u`).

Otherwise said, I don't see what could be wrong in simply keeping the code discharging template polymorphic inductive types as it is. What is important is the local graph of an inductive type, i.e. when `u` is used inside the definition of the type, since then, this internal `u` is bound by the `u` of the type of the argument. Occurrences of `u` which are used outside the inductive definition, as the one in the body of `B`, are bound at the place of `Universe u`, not at the `box (A:Type@{u})` place. These are two distinct binders using the same binding name.

**Comment by mattam82** [comment]
I agree the universes should be binders in the case of template-polymorphism, just as in universe polymorphism. The problem is precisely when you interleave definitions depending on the same universes and the inductive (so when your definition B is above the inductive and used in its declaration). By a kind of miracle one could do that in the previous implementation (there's an example in ring of an abbreviation `Def cpol := Pol C` using that, perhaps unknowingly). To be correct one should keep track of the use of the section-introduced universe variables across definitions, to see if they're ever bounded from below, which is entirely silent in template-polymorphism mode. Also, I don't want to introduce even more confusion by interpreting global universes as binding variables rather than floating ones in general. Let's just have proper universe variables!

File: test-suite/failure/Template.v [hunk 17]
```diff
@@ -0,0 +1,32 @@
+(*
+Module TestUnsetTemplateCheck.
```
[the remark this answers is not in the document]
mattam82: Yes

test-suite/success/Template.v
```diff
@@ -46,3 +46,112 @@ Module No.
...
+
+Module ExplicitTemplate.
+  #[universes(template)]
+  Inductive identity@{i} (A : Type@{i}) (a : A) : A -> Type@{i} := id_refl : identity A a a.
+
+  (* Weird intraction of template polymorphism and inductives
+     which naturally fall in Prop: this one is template polymorphic but not on i:
+     it just lives in any universe *)
+  Check (identity Type nat nat : Prop).
+End ExplicitTemplate.
```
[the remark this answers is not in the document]
mattam82: I guess you mean we could require the parameters to be all anonymous to prevent sharing? I'm pretty sure a syntactic check like this would not be enough, one can use tactics in terms to get access to the parameters in constructor types.

mattam82:
> Maybe this is the same feeling as @SkySkimmer, I find it a bit strange that whether universes have initial constraints >= Prop or >= Set is in the global environment vs being a property at least of the current declaration being treated, if not a property of each universe under consideration.

It could be indeed, it was just simpler to implement it there. You're right it makes more sense at the declaration level or the universe one. 

> For instance, maybe the idea is odd and not easy to implement, but would it make sense that the lower bound is just enforced as an explicit constraint in the constraints part of a Univ.UContext.t? In which case, the enforce_constraint in UGraph.add_universe would just be probably moved somewhere in uState.ml to add the constraint at the same time the universe is added.

Yes, but again that's a bigger change.

> Also, does it prevent something to ensure that some u is >=Prop or would it be equivalent to declare u w/o any constraints.

It prevents having universes lower than Prop, if you don't add any constraints one can declare i < Prop.

herbelin:
> ... so when your definition B is above the inductive and used in its declaration ...

Ah, indeed, well spotted! This means that #9294 could (a priori) be exploited vernac side even before the introduction of universe polymorphism :(
```
Section Foo.
Context (A : Type).
Definition CastSet (A:Type) := Set : ltac:(let ty := type of A in exact ty).
Inductive Foo := foo : CastSet A -> Foo.
End Foo.
Check Foo True : Prop.
```
which can then be continued as in [9294](https://github.com/coq/coq/issues/9294#issuecomment-450860504). (It remains to find an equivalent to `ltac:(let ty := type of A in exact ty)` in 8.4 to get there a vernac-side critical bug.)

herbelin:
> It prevents having universes lower than `Prop`, if you don't add any constraints one can declare `i < Prop`.

Am I correct that a cheaper alternative (in terms of number of edges in the graph) could be to ensure that no constraints `i<=Prop` or `i<Prop` is added.

Otherwise, yes, other suggestions would be big changes.

herbelin:
> Also, I don't want to introduce even more confusion by interpreting global universes as binding variables rather than floating ones in general. Let's just have proper universe variables!

I agree.

herbelin:
Another version, using `ltac:` and `type of` to force the universe level to be shared, but without using sections (nor `@{i}`):
```
Inductive Foo A (S:= fun _ => Set : ltac:(let ty := type of A in exact ty))
  := foo : S A -> Foo A.
Check Foo True : Prop.
```
Ruled out by the `unbounded_from_below` test of the PR.

mattam82:
Yep, I’ll add those to the test-suite while I’m at it

vernac/comInductive.ml
```diff
@@ -345,25 +363,46 @@ let restrict_inductive_universes sigma ctx_params arities constructors =
...
+  then user_err (str "Inductives with uniform parameters may not have attached notations.");
+
+  let indnames = List.map (fun ind -> ind.ind_name) indl in
+  let sigma, env_params, infos =
+    interp_params env0 udecl uparamsl paramsl
+  in
 
   (* Interpret the arities *)
-  let sigma, arities = List.fold_left_map (fun sigma -> interp_ind_arity env_params sigma) sigma indl in
+  let arities = List.map (intern_ind_arity env_params sigma) indl in
```
herbelin: More remark: this could be moved to the `else` branch of `not poly && is_template` per symmetry and to avoid `arities` to be computed twice in the `then` branch.

mattam82:
I rebased, squashed everything and added Hugo's tests to the test-suite file. Should be ready now.

vernac/comInductive.ml
mattam82: That must be old code you looked at, now we only intern_arity twice in template-polymorphism mode (because the params and sigma must be refreshed in that case).

herbelin:
> Should an entry also be added in dev/doc/critical-bugs?

Here is a first proposal for addition:
```
  component: template polymorphism                                                                                                 
  summary: template polymorphism not collecting side constrains on the universe level of a parameter; this is a general form of the previous issue about template polymorphism exploiting other ways to generate untracked constraints                                
  introduced: morally at the introduction of template polymorphism, 23 May 2006, 9c2d70b, r8845, Herbelin                          
  impacted released versions: at least V8.4-V8.4pl6, V8.5-V8.5pl3, V8.6-V8.6pl2, V8.7.0-V8.7.1, V8.8.0-V8.8.1, V8.9.0-V8.9.1, in theory also V8.1-V8.1pl4, V8.2-V8.2pl2, V8.3-V8.3pl2 but not exploit found there yet (an exploit using a plugin to force sharing of universe level is in principle possible though)                                                                                   
  impacted development branches: all from 8.4 to 8.9 at the time of writing and suspectingly also all from 8.1 to 8.4 if a way to create untracked constraints can be found
  impacted coqchk versions: a priori all (tested with V8.4 and V8.9 which accept the exploit)                                      
  fixed in: soon in master and V8.10.0 (PR #9918, Aug 2019, Dénès and Sozeau)                                                                
  found by: Gilbert using explicit sharing of universes, exploit found for 8.5-8.9 by Pédrot, other variants generating sharing using sections, or using ltac tricks by Sozeau, exploit in 8.4 by Herbelin and Jason Gross by adding new tricks to Sozeau's variants 
  exploit: test-suite/failure/Template.v                                                                                           
  GH issue number: #9294                                                                                                           
  risk: moderate risk to be activated by chance         
```
About the risk: Gaëtan, how did you find it? I suppose this was by investigations of the semantics of template polymorphism? How would you qualify the risk?

SkySkimmer:
I don't remember how I found it.
>How would you qualify the risk?

I don't expect any developments to use the bug. OTOH if someone has to deal with malicious proofs they should be very worried, it seems tricky to notice.

ppedrot:
@herbelin's criticality entry looks good to me. Anyways, it would be good for this patch to go in as fast as possible. We can push the CEV atop of this branch ourselves, and if @SkySkimmer doesn't wish to assign I can do so instead. This PR has already been reviewed separately by several persons IIUC, so that there is little left before merging.

herbelin:
Sure, merging is ok to me.

ppedrot (approved):
The changes to the kernel seem OK to me w.r.t. the high-level description of the issue.

ppedrot:
OK, critical bug entry added. @coq/core Last round to complain, will merge tomorrow if no input by then.

PR #9924, opened 2019-04-06
Fix numeral notations test in async mode.

SkySkimmer:
Async causes output reordering in one test. Since we don't care about
the output of that test (it's just a [Fail]) we move it to success/.


ejgallego:
You can also add the `-async off` header as done in other files in `output`.

SkySkimmer:
I prefer the solution I picked, we shouldn't test output when we don't 
care about it.

Gaëtan Gilbert

ejgallego (approved):

PR #9926, opened 2019-04-07
[vernac] [ast] Make location info an attribute of vernaculars.

ejgallego:
Whether locations where attached to vernacs has been a bit of a mess for a while; we try to be more uniform in the style of the rest of the AST.

### DIFF
--- a/ide/idetop.ml
+++ b/ide/idetop.ml
@@ -64,7 +64,7 @@ let is_known_option cmd = match Vernacprop.under_control cmd with
 
 (** Check whether a command is forbidden in the IDE *)
 
-let ide_cmd_checks ~last_valid {CAst.loc;v=ast} =
+let ide_cmd_checks ~last_valid ({ CAst.loc; _ } as cmd) =
   let user_error s =
     try CErrors.user_err ?loc ~hdr:"IDE" (str s)
     with e ->
@@ -72,14 +72,14 @@ let ide_cmd_checks ~last_valid {CAst.loc;v=ast} =
       let info = Stateid.add info ~valid:last_valid Stateid.dummy in
       Exninfo.raise ~info e
   in
-  if is_debug ast then
+  if is_debug cmd then
     user_error "Debug mode not available in the IDE"
 
-let ide_cmd_warns ~id {CAst.loc;v=ast} =
+let ide_cmd_warns ~id ({ CAst.loc; _ } as cmd) =
   let warn msg = Feedback.(feedback ~id (Message (Warning, loc, strbrk msg))) in
-  if is_known_option ast then
+  if is_known_option cmd then
     warn "Set this option from the IDE menu instead";
-  if is_navigation_vernac ast || is_undo ast then
+  if is_navigation_vernac cmd || is_undo cmd then
     warn "Use IDE navigation instead"
 
 (** Interpretation (cf. [Ide_intf.interp]) *)
@@ -137,7 +137,7 @@ let annotate phrase =
   | None -> Richpp.richpp_of_pp 78 (Pp.mt ())
   | Some ast ->
     (* XXX: Width should be a parameter of annotate... *)
-    Richpp.richpp_of_pp 78 (Ppvernac.pr_vernac ast.CAst.v)
+    Richpp.richpp_of_pp 78 (Ppvernac.pr_vernac ast)
 
 (** Goal display *)
 
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -119,7 +119,6 @@ let async_proofs_workers_extra_env = ref [||]
 
 type aast = {
   verbose : bool;
-  loc : Loc.t option;
   indentation : int;
   strlen : int;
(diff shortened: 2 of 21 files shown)


vernac/ppvernac.ml
```diff
@@ -1283,5 +1283,5 @@ let pr_vernac_attributes =
       return (keyword "Fail" ++ spc() ++ pr_vernac_control v)
 
     let pr_vernac v =
-      try pr_vernac_control v
+      try pr_vernac_control v.v
```
SkySkimmer: Any reason why pr_vernac_control doesn't take the cast?

vernac/g_vernac.mlg
```diff
@@ -148,7 +148,7 @@ GRAMMAR EXTEND Gram
   ;
 
   located_vernac:
-    [ [ v = vernac_control -> { CAst.make ~loc v } ] ]
+    [ [ v = vernac_control -> { v } ] ]
```
SkySkimmer: Any reason to keep `located_vernac`?

vernac/ppvernac.ml
ejgallego: Fixed thanks.

vernac/g_vernac.mlg
ejgallego: Indeed, thanks.

ide/idetop.ml
```diff
@@ -64,22 +64,22 @@ let is_known_option cmd = match Vernacprop.under_control cmd with
 
 (** Check whether a command is forbidden in the IDE *)
 
-let ide_cmd_checks ~last_valid {CAst.loc;v=ast} =
+let ide_cmd_checks ~last_valid ({CAst.loc;v=ast} as cmd) =
```
SkySkimmer: `ast` isn't used here

ide/idetop.ml
```diff
@@ -64,22 +64,22 @@ let is_known_option cmd = match Vernacprop.under_control cmd with
...
       let (e, info) = CErrors.push e in
       let info = Stateid.add info ~valid:last_valid Stateid.dummy in
       Exninfo.raise ~info e
   in
-  if is_debug ast then
+  if is_debug cmd then
     user_error "Debug mode not available in the IDE"
 
-let ide_cmd_warns ~id {CAst.loc;v=ast} =
+let ide_cmd_warns ~id ({CAst.loc;v=ast} as cmd) =
```
SkySkimmer: `ast` isn't used here

stm/stm.ml
```diff
@@ -2307,7 +2307,7 @@ let collect_proof keep cur hd brkind id =
         (try
           let name, hint = name ids, get_hint_ctx loc  in
           let t, v = proof_no_using last in
-          v.expr <- VernacExpr([], VernacProof(t, Some hint));
+          v.expr <- CAst.make @@ VernacExpr([], VernacProof(t, Some hint));
```
SkySkimmer: Doesn't this cause the loc to be lost?

toplevel/vernac.ml
```diff
@@ -20,12 +20,12 @@ open Vernacprop
...
    (the exceptions are explained only at the toplevel). *)
 
-let checknav_simple {CAst.loc;v=cmd} =
+let checknav_simple ({CAst.loc;v=_} as cmd) =
   if is_navigation_vernac cmd && not (is_reset cmd) then
     CErrors.user_err ?loc (str "Navigation commands forbidden in files.")
 
-let checknav_deep {CAst.loc;v=ast} =
-  if is_deep_navigation_vernac ast then
+let checknav_deep ({CAst.loc;v=ast} as cmd) =
```
SkySkimmer: `ast` isn't used here

SkySkimmer (commented):
In a few places, `let foo = function Ba -> ...` is replaced by `let foo v = CAst.with_val (function Ba -> ...) v`.
It might be nicer to do `let foo v = v |> CAst.with_val (function Ba -> ...)` instead.

ejgallego:
Merci for the corrections, fixed.

> In a few places, `let foo = function Ba -> ...` is replaced by `let foo v = CAst.with_val (function Ba -> ...) v`.
> It might be nicer to do `let foo v = v |> CAst.with_val (function Ba -> ...)` instead.

Why is it nicer? I see no difference honestly [as we are forced to eta-expand anyways due to the let-rec]

I can change it if you want, I don't mind a form or the other.

SkySkimmer:
>Why is it nicer?

It's easier to see what the with_val is applied to since it's not multiple lines later IMO.

ejgallego:
> It's easier to see what the with_val is applied to since it's not multiple lines later IMO.

Ok, I'll change it if you prefer; note that there are many more occurrences in the codebase of this pattern that I won't be changing.

ejgallego:
This is IMO ready, needs assignee.

ppedrot (approved):

ppedrot:
Fairly consensual change. I'll merge this tomorrow if nobody complains.

ejgallego:
Merci.

PR #9927, opened 2019-04-08
Don't store closures in summary (reduction effects)

SkySkimmer:
We already have indirection (constant -> effect name, effect name ->
function) so we only need to stop using summary.ref for the second map.


ppedrot (approved):

SkySkimmer:
For some reason the macos build shows as in progress but it's actually finished.

PR #9928, opened 2019-04-08
Fix #9812: CoqIDE on gtk3 has wrong defaults for selection BG.

### LINKED ISSUE #9812: CoqIDE on gtk3 has wrong defaults for selection BG
In a standard gnome system (no color customization) and after wiping out the old coqide config file,
text selection has a transparent background an white text. As a result you don't see the text you are selecting. I've no idea what CoqIDE does wrong, but on gtk2 it was not the case

ppedrot:
Unfortunately, the only sane fix I was able to hack was to deactivate the possibility to change the background colour altogether. Trying to do otherwise is a real Pandora's box which is ultimately triggered by the lack of lablgtk3 bindings for the GtkStyleContext class.

I tried a lot of alternative approaches, to no avail. This includes catching the selection signal, reimplementing selection by hand in GtkTextView, and even reading the internal details of the GTK implementation turned not that helpful.

For the time being (8.10 beta) I think we do not have much choice.

Fixes #9812.

### DIFF
--- a/ide/preferences.ml
+++ b/ide/preferences.ml
@@ -410,8 +410,8 @@ let vertical_tabs =
 let opposite_tabs =
   new preference ~name:["opposite_tabs"] ~init:false ~repr:Repr.(bool)
 
-let background_color =
-  new preference ~name:["background_color"] ~init:"cornsilk" ~repr:Repr.(string)
+(* let background_color = *)
+(*   new preference ~name:["background_color"] ~init:"cornsilk" ~repr:Repr.(string) *)
 
 let attach_tag (pref : string preference) (tag : GText.tag) f =
   tag#set_property (f pref#get);
@@ -737,7 +737,7 @@ let configure ?(apply=(fun () -> ())) parent =
       ()
     in
     let () = Util.List.iteri iter [
-      ("Background color", background_color);
+(*       ("Background color", background_color); *)
       ("Background color of processed text", processed_color);
       ("Background color of text being processed", processing_color);
       ("Background color of incompletely processed Qed", incompletely_processed_color);
--- a/ide/preferences.mli
+++ b/ide/preferences.mli
@@ -88,7 +88,7 @@ val reset_on_tab_switch : bool preference
 val line_ending : line_ending preference
 val vertical_tabs : bool preference
 val opposite_tabs : bool preference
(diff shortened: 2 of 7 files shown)


ppedrot:
FTR, I don't know what is the API policy for lablgtk2-on-3. Should we ask for non Gtk2-classes that are critical for usability, as it is the case here? Or should we switch to actual Gtk3 bindings at some point? cc @garrigue.

ejgallego:
I am not sure what you mean @ppedrot but lablgtk is on beta and it does include a lot of GTK3 stuff, so it is not really gtk2-on-3; @garrige will tell more but my impression is that adding a binding for GtkStyleContext is welcome.

garrigue:
I'm not sure what you mean.
It is of course fine to add new classes to lablgtk3.
lablgtk2on3 was an intermediate step to quickly provide bindings for gtk3, but I'm open to any usability improvements.

garrigue:
Sorry for the double answer. I'm in full agreement with @ejgallego .

ppedrot:
@garrigue My question boils down to "what goes into lablgtk3"? IIUC it is still featuring legacy Gtk2 bindings that are absent from the Gtk3 API. Or am I wrong, and is lablgtk3 a plain one-to-one binding to Gtk3?

ejgallego:
> Or am I wrong, and is lablgtk3 a plain one-to-one binding to Gtk3?

At least the build system doesn't link to gtk2 at all; so whatever APIs are present there are also on GTK3 itsel.f

garrigue:
To the best of my knowledge, the is no specific code in lablgtk3 that emulates gtk2.
It may use deprecated APIs, but only as long as they are available in gtk3. I.e., do not get out of its way to switch to the "gtk3" way, but do not try to avoid it where it is needed either.

gares:
Since `GtkStyleContext` is not there yet, this PR should now be merged and eventually superseded by another one when that class will be available in a released version of lablgtk3.

With this PR we can do copy/paste in CoqIDE at the cost of dropping the possibility of customizing the background color, that is way less important than copy/paste IMO.

I'll test it locally and merge.

gares (approved):

gares:
It works as expected, thanks for the quick fix @ppedrot

PR #9933, opened 2019-04-09
Add a few missing notes to the release doc.

Zimmi48:
**Kind:** documentation

@vbgl In particular, see the point about pinning basic overlays.

### DIFF
--- a/dev/doc/release-process.md
+++ b/dev/doc/release-process.md
@@ -2,6 +2,11 @@
 
 ## As soon as the previous version branched off master ##
 
+In principle, these steps should be undertaken by the RM of the next
+release. Unfortunately, we have not yet been able to nominate RMs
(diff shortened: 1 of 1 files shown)


dev/doc/release-process.md
```diff
@@ -2,6 +2,11 @@
 
 ## As soon as the previous version branched off master ##
 
+In principle, these steps should be undertaken by the RM of the next
+release. Unfortunately, we have not yet been able to nominate RMs
+early enough in the process for this person to be known at that point
+in time.
```
ejgallego: I am not sure about this; shouldn't we add instead the nomination process as a step / pre-requisite?

Zimmi48: Yep!

vbgl (approved):

PR #9935, opened 2019-04-09
[api] [proof] Alert users that `Vernacstate.Proof_global` is not to be used.

ejgallego:
We alert users that `Vernacstate.Proof_global` is a Coq internal
module and should not be used to workaround lack of state threading.

### DIFF
--- a/stm/stm.ml
+++ b/stm/stm.ml
@@ -27,6 +27,8 @@ open Feedback
 open Vernacexpr
 open Vernacextend
 
+module PG_compat = Vernacstate.Proof_global [@@ocaml.warning "-3"]
+
 let is_vtkeep = function VtKeep _ -> true | _ -> false
 let get_vtkeep = function VtKeep x -> x | _ -> assert false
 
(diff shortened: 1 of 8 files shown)


ejgallego:
We will be able to remove this module when the STM is ported to handle proof state functionally, however this is not trivial, and in fact the STM seems to be currently workarounding a few bugs in the lower layers that should be fixed by the proof tree.

stm/stm.ml
```diff
@@ -27,6 +27,8 @@ open Feedback
 open Vernacexpr
 open Vernacextend
 
+module PG_compat = Vernacstate.Proof_global [@@ocaml.warning "-3"]
```
gares: Can't you just write
```
module Vernacstace = struct
  module Proof_global = Vernacstate.Proof_global [@@ocaml.warning "-3"]
end
```
In order to avoid all the renaming below?

ejgallego: That's another choice but you'd have to `include Vernacstate` in full as not to hide the API, which has a bad impact on code size; on the other hand a module alias is free.

vbgl:
Is this change really for 8.10? It does not look like a bug fix…

Who will assign to this PR?

stm/stm.ml
gares: > which has a bad impact on code size; on the other hand a module alias is free.

How much is the binary size increased? 10Kb? 100Kb? What does "bad" mean?
It is not for free, since you have to rename things.

ejgallego:
> Is this change really for 8.10? It does not look like a bug fix…

The change must go in 8.10 , the API should not be exposed. The implementation provided here works, feel free to deprecate using some other approach if you would like.

vbgl:
Then can someone assign, review and merge? Thanks.

maximedenes (approved):

PR #9947, opened 2019-04-11
Remove Global.env from goptions by passing from vernacentries

SkySkimmer:
Currently this env is only used to error for Printing If/Let on
non-2-constructor/non-1-constructor types so we could alternatively
remove it and not error / error later when trying to print.
Keeping the env and the error as-is should be fine though.

### DIFF
--- a/library/goptions.ml
+++ b/library/goptions.ml
@@ -42,13 +42,12 @@ let error_undeclared_key key =
 (****************************************************************************)
 (* 1- Tables                                                                *)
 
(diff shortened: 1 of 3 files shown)


ejgallego (approved):
Indeed the use of globalization seems a bit overkill but nonetheless the cleanup seems good.

ejgallego:
@SkySkimmer please rebase so get a clean CI run.

ejgallego:
Failures are spurious.

PR #9949, opened 2019-04-11
[stm] Report correct ids on some errors where it was dummy.

### LINKED ISSUE #9948: Dummy sentence ids in error reports from Stm

Error messages about proof nesting call `Stm.exn_on ~valid:Stateid.dummy Stateid.dummy`.
The new tip, at least should be known at this point.

#### Coq Version

8.10 (master)

corwin-of-amber:
**Kind:** bugfix / infrastructure.

Fixes / closes #9948

This fixes the `id` argument, I was not sure about `~valid`.


gares:
The patch looks ok. `~valid` is the last know valid id (in this case there is not such an id).

gares (approved):
Thanks!

PR #9953, opened 2019-04-12
[options] Remove string option type.

ejgallego:
As discussed in https://github.com/coq/coq/pull/9876 , "string option"
options are problematic in the wider scope of interacting with other
tools.

This is a historical shell problem, in the sense that for an
environment variable `FOO`, it is hard to distinguish whether `FOO` is
unset or `FOO` is set to `""`.

As noted by @Skyskimmer, Coq only uses this type in a couple of
places, which would be better served anyways by different approaches:

- in the case of proof using, the grammar should capture "unset"
- in the case of the dump, dump filename and dump on / off should be
  different options.

### DIFF
--- a/ide/idetop.ml
+++ b/ide/idetop.ml
@@ -345,13 +345,11 @@ let export_option_value = function
   | Goptions.BoolValue b   -> Interface.BoolValue b
   | Goptions.IntValue x    -> Interface.IntValue x
   | Goptions.StringValue s -> Interface.StringValue s
-  | Goptions.StringOptValue s -> Interface.StringOptValue s
 
 let import_option_value = function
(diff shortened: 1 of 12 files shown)


gares:
The old code and types in micromega looks better than the new one: you put a dummy value in place of an option type.

Is this what @SkySkimmer had in mind?

SkySkimmer (changes requested):
I have no desire for what this PR does, optional strings work fine. Other tools will just have to avoid reading them through shell variables.

>    in the case of proof using, the grammar should capture "unset"

I don't see how that's better.

ejgallego:
Let's close then.

Issue #7015, opened 2018-03-18
Polymorphic cumulativity and function eta break subject reduction


SkySkimmer:
Present in 8.7 and current master.
```coq
Set Universe Polymorphism.
Set Polymorphic Inductive Cumulativity.
Set Printing Universes.

(* in the real world foo@{i} might be [@paths@{i} nat] I guess *)
Inductive foo : nat -> Type :=.

(* on [refl (fun x => f x)] this computes to [refl f] *)
Definition eta_out {A B} (f g : forall x : A, B x) (e : (fun x => f x) = (fun x => g x)) : f = g.
Proof.
  change (f = g) in e. destruct e;reflexivity.
Defined.

Section univs.
  Universes i j.
  Constraint i < j. (* constraint not needed in master *)

  Definition one : (fun n => foo@{i} n) = fun n => foo@{j} n := eq_refl.

  Definition two : foo@{i} = foo@{j} := eta_out _ _ one.

  Fail Definition two' : foo@{i} = foo@{j} := Eval compute in two.

  (* Failure of SR doesn't just mean that the type changes, sometimes
     we lose being well-typed entirely. *)
  Definition three := @eq_refl (foo@{i} = foo@{j}) two.
  Fail Definition four := Eval compute in three.
End univs.
```

mattam82:
There seems to be a bug in the implementation if we don’t apply  cumulativity in this case to accept `two’` to be inhabited by `eq_refl`. Same thing for lists, conversion by cumulativity should also apply in partially applied cases to be stable by eta. @amintimany any idea how to fix/justify this?

herbelin:
Side comment: in CoqIDE, when evaluating directly to the end, `one` fails.

herbelin:
A question by curiosity:
- With `Constraint i < j`, `one` is inferred to be `eq_refl (fun n : nat => foo@{i} n)`.
- If I remove the `Constraint i < j`, then `one` enforces `i = j`.
- If I change it to `Constraint j < i`, `one` is still inferred to be `eq_refl (fun n : nat => foo@{i} n)`.

I would spontaneously suspect some arbitrariness in the order in which unification treats problems, as it is common to see. So, just wondering what is our degree of freedom, e.g. if `eq_refl (fun n : nat => foo@{j} n)` could be obtained in the latter case?

SkySkimmer:
That's outdated, in master removing the constraint should not enforce i = j.

herbelin:
[...]

PR #9959, opened 2019-04-14
[native compiler] Remove `Lconstruct` from lambda code.

maximedenes:
This is a step towards unifying the higher level ILs of the native and
bytecode compilers.

See https://github.com/coq/coq/projects/17

### DIFF
--- a/kernel/nativecode.ml
+++ b/kernel/nativecode.ml
@@ -1280,9 +1280,6 @@ let ml_of_instance instance u =
   | Lmakeblock (prefix,(cn,_u),_,args) ->
      let args = Array.map (ml_of_lam env l) args in
      MLconstruct(prefix,cn,args)
-  | Lconstruct (prefix, (cn,u)) ->
-     let uargs = ml_of_instance env.env_univ u in
-      mkMLapp (MLglobal (Gconstruct (prefix, cn))) uargs
   | Luint i -> MLapp(MLprimitive Mk_uint, [|MLuint i|])
   | Lval v ->
       let i = push_symbol (SymbValue v) in get_value_code i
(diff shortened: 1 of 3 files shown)


ppedrot:
How does this play with inductive cumulativity? Isn't systematically eta-expanding a constructor forbidding it to take place on parameters somehow?

SkySkimmer:
Rather we should be systematically eta expanding to get it to work, cf https://github.com/coq/coq/issues/7015

maximedenes:
> this

What do you mean? This patch doesn't change how/when we eta expand constructors, does it?

ppedrot:
@maximedenes Sorry, I misread your patch. Indeed the generated code is the same, because `Gconstruct` definitions are always η-expanded.

kernel/nativecode.ml
```diff
@@ -1280,9 +1280,6 @@ let ml_of_instance instance u =
   | Lmakeblock (prefix,(cn,_u),_,args) ->
      let args = Array.map (ml_of_lam env l) args in
      MLconstruct(prefix,cn,args)
-  | Lconstruct (prefix, (cn,u)) ->
-     let uargs = ml_of_instance env.env_univ u in
-      mkMLapp (MLglobal (Gconstruct (prefix, cn))) uargs
```
ppedrot: It looks like this is the only place where we generate a `Gconstruct` node, which means it is now dead code. You should probably remove it as well.

maximedenes: Oh, good point, let me do it as part of this PR, since it will simplify some further patches.

maximedenes: Done.

maximedenes:
> This patch doesn't change how/when we eta expand constructors, does it?

In fact, this was a bit inaccurate: we now eta-expand on the fly, whereas we used to precompile eta-expanded versions. The VM was already doing the latter. I don't think it matters a lot, since performance-critical examples more often involve fully applied constructors.

ppedrot (approved):

PR #9962, opened 2019-04-15
[native compiler] Encoding of constructors based on tags

maximedenes:
This serves two purposes:

1. It makes the native compiler use the same encoding and
   lambda-representation as the bytecode compiler

2. It avoid relying on fragile invariants relating tags and constructor
   indices. For example, previously, the mapping from indices to tags
   had to be increasing.

### DIFF
--- a/kernel/nativelambda.ml
+++ b/kernel/nativelambda.ml
@@ -39,9 +39,10 @@ type lambda =
   | Lif           of lambda * lambda * lambda
   | Lfix          of (int array * (string * inductive) array * int) * fix_decl
   | Lcofix        of int * fix_decl
+  | Lint          of int (* a constant constructor *)
   | Lmakeblock    of prefix * pconstructor * int * lambda array
                   (* prefix, constructor Name.t, constructor tag, arguments *)
-        (* A fully applied constructor *)
+        (* A fully applied non-constant constructor *)
   | Luint         of Uint63.t
   | Lval          of Nativevalues.t
   | Lsort         of Sorts.t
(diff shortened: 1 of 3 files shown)


kernel/nativelambda.ml
```diff
@@ -40,7 +40,7 @@ type lambda =
   | Lfix          of (int array * (string * inductive) array * int) * fix_decl
   | Lcofix        of int * fix_decl
   | Lint          of int (* a constant constructor *)
-  | Lmakeblock    of prefix * pconstructor * int * lambda array
+  | Lmakeblock    of prefix * constructor * int * lambda array
```
ppedrot: I am surprised that this data is not required to check constructor cumulativity. I am not an expert on the topic, so let's summon @SkySkimmer (who, hopefully, will dig out an unsoundness proof?).

maximedenes: Note that this is not changed by this PR, as this data was already ignored.

Last time I worked on univ poly for the VM and `native_compute`, I was told by @mattam82 that constructors were always carrying the same instance as their inductive type. If that is correct, this information is not needed.

ppedrot (approved):

maximedenes:
@ppedrot is anything holding this?

ppedrot:
Is the CI failure related?

maximedenes:
> Is the CI failure related?

I don't think so, but let's relaunch everything.

ppedrot:
Something looks reliably broken, doesn't it?

ejgallego:
> Something looks reliably broken, doesn't it?

The depending job has expired, thus the test-suite cannot find the build artifacts; I have relaunched the base flambda job, you should be able to relaunch the test-suite one when it finishes.

Note that this is the original failure: https://gitlab.com/coq/coq/-/jobs/198243378 (which happened 3 times so I dunno)

ppedrot:
The original failure looks unrelated, I think I'll merge the PR, except if @maximedenes thinks there is something wrong.

ejgallego:
> The original failure looks unrelated, I think I'll merge the PR, except if @maximedenes thinks there is something wrong.

The fact that it was reproduced 3 times reliably should maybe have you look deeper into it.

ejgallego:
A new run is here https://gitlab.com/coq/coq/-/jobs/201125501 so let's see.

ppedrot:
See, that was a spurious failure.

PR #9966, opened 2019-04-17
Add changes for -set

SkySkimmer:
I realized this was missing just as the PR got merged

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -227,6 +227,8 @@ Tools
   `coqc`/`make` as well as printing to stdout, on both python2 and
   python3.
 
+- Coq options can be set on the command line, eg `-set "Universe Polymorphism=true"`
+
 Standard Library
 
 - Added lemmas about monotonicity of `N.double` and `N.succ_double`, and about


Zimmi48:
Please don't do this (putting the milestone after merging): this breaks the automation in place.

ejgallego:
Sorry @Zimmi48 , indeed I raced against myself.

ejgallego:
Maybe we should disallow skipping a merge if the milestone is not yet?

Zimmi48:
Unfortunately, these checks were intended as something people should almost never skip but some of them have become too frequent and people have gotten used to skipping them.

ejgallego:
> Unfortunately, these checks were intended as something people should almost never skip but some of them have become too frequent and people have gotten used to skipping them.

I never skip the milestone/kind check, and indeed I see no good reason to; in this case it was just that I pushed before I clicked on the UI interface :(

ejgallego:
But indeed if they should never be skipped maybe we should remove the y/n question?

SkySkimmer:
Maybe the script could ask the user which milestone they want and set it.
That would require some work to get tokens for authenticating to github I guess.

ejgallego:
Yup, for sure @SkySkimmer at some point we want the tooling to interact with Github [for example to have a submit pr with overlays] script.

I am working on these stuff [in OCaml tho] but I don't think I'll be able to produce anything before the CUDM.

PR #9968, opened 2019-04-17
[bugfix] Fixes #9967.

### LINKED ISSUE #9967: Cast is using vm for type-checking, even when byte compiler is disabled.

The `<:` notation is translated to `VMcast`, and `vm_conv_gen` invokes the VM even if Coq was configured with `-bytecode-compiler no`. In `vm_conv_gen`, the flag `(typing_flags env).enable_VM` is tested, but this flag is set to `true` by default, unconditionally, since #8870.

#### Coq Version

8.10 (master)

corwin-of-amber:
**Kind:** bug fix.

Fixes #9967.

### DIFF
--- a/kernel/declareops.ml
+++ b/kernel/declareops.ml
@@ -22,8 +22,8 @@ let safe_flags oracle = {
   check_universes = true;
   conv_oracle = oracle;
   share_reduction = true;
-  enable_VM = true;
-  enable_native_compiler = true;
+  enable_VM = Coq_config.bytecode_compiler;
+  enable_native_compiler = Coq_config.native_compiler;
   indices_matter = true;
 }
 


SkySkimmer:
Please use a more informative commit message.

ejgallego:
Was this bug introduced in https://github.com/coq/coq/pull/8870 ?

ejgallego:
Ummm, maybe that's not a bug in Coq, but in the jsCoq toplevel [sorry @corwin-of-amber only now I could look at the code]

I think the idea of #8870 was to delegate flags to the upper layers, so maybe is jsCoq who should do:
```
 Global.set_VM false;
 Global.set_native_compiler false;
```
as done by coqtop.

ejgallego:
Anyways the handling of kernel flags need cleaning, I have a PR doing it somewhere so this kind of bugs don't happen again.

corwin-of-amber:
Ok, so should we discard this one?

corwin-of-amber:
Will be fixed by @ejgallego in a separate PR.

PR #9972, opened 2019-04-20
[build] Select uint63 using `ocamlc -config` variables.

ejgallego:
This seems more robust and avoids having another implementation of
`cp`.


kernel/dune
```diff
@@ -14,15 +14,10 @@
...
-(executable
-  (name write_uint63)
-  (modules write_uint63)
-  (libraries unix))
-
 (rule
  (targets uint63.ml)
- (deps (:gen ./write_uint63.exe) uint63_x86.ml uint63_amd64.ml)
- (action (run %{gen})))
+ (deps (:gen-file uint63_%{ocaml-config:architecture}_%{ocaml-config:int_size}.ml))
```
SkySkimmer: While we're at it, why do we depend on `architecture`? Those files don't have any assembly AFAICT

kernel/dune
```diff
@@ -14,15 +14,10 @@
...
+ (action (copy %{gen-file} %{targets})))
```
SkySkimmer: Why not `copy#`?

kernel/write_uint63.ml
```diff
@@ -31,8 +31,8 @@ let ml_file_copy input output =
 
 let write_uint63 () =
   ml_file_copy
-    (if max_int = 1073741823 (* 32-bits *) then "uint63_x86.ml"
-     else (* 64 bits *) "uint63_amd64.ml")
+    (if max_int = 1073741823 (* 32-bits *) then "uint63_i386_31.ml"
+     else (* 64 bits *) "uint63_amd64_63.ml")
     "uint63.ml"
 
 let () = write_uint63 ()
```
SkySkimmer: I guess this is kept for make
Alternatively we could dispatch in configure, ie
- configure writes `INT_SIZE` to config/Makefile
- Makefile.build copies `uint63_$(INT_SIZE)`

SkySkimmer:
@ejgallego can you respond to my earlier comments on this PR?

kernel/dune
ejgallego: No idea, I just followed previous naming.

ejgallego: cc @vbgl

kernel/write_uint63.ml
ejgallego: Indeed, this is kept for make, as well as the copy program.

ejgallego:
> @ejgallego can you respond to my earlier comments on this PR?

Done, sorry.

kernel/dune
vbgl: Indeed, this information seems redundant. What we really care about here is the int size.

gares (approved):

Zimmi48 (commented):
Can we get an issue to remember to remove the copy program and the `write_uint63.ml` file when the Makefiles are removed? And can you remove the architecture name from the files since it is apparently redundant?

maximedenes (approved):

SkySkimmer:
This needs an assignee

vbgl (approved):

maximedenes:
I had missed the file name changes. What was the motivation for adding a suffix?

Issue #9975, opened 2019-04-20
Enable a MinGW Windows bytecode debug build


MSoegtropIMC:
Since CoqIDE crashes in 8.10 on Windows, I need to debug this. Creating bytecode binaries of the main coq tools works with some patches to the makefiles, but building a bytcode version of CoqIDE is tricky cause of the dependencies to C libraries like GTK3 and Cairo2.

There has been some discussion on the topic in #8729.

Since we can hardly release 8.10 with a broken CoqIDE on Windows, I set this to high prio.

ejgallego:
You can also use the dune build which by the way will get you way further as it will properly compose in case you need to trace across libraries .

Also, gdb + native tends to work very well (in Linux tho)

MSoegtropIMC:
Some more notes on this:

The DLL `dllcairo_stubs.dll` which is not found during `OCAMLC -o bin/coqide.byte.exe` is installed in the correct path:
```
$ ocamlfind printconf
Effective configuration:
: 
The ld.conf file can be found here:
    C:\bin\cygwin_coq64_ms_coqide\usr\x86_64-w64-mingw32\sys-root\mingw\libocaml\ld.conf
```
The ld.conf file mentioned above contains:
```
C:/bin/cygwin_coq64_ms_coqide/usr/x86_64-w64-mingw32/sys-root/mingw/libocaml/stublibs
C:/bin/cygwin_coq64_ms_coqide/usr/x86_64-w64-mingw32/sys-root/mingw/libocaml
```
The DLL is in the above stublib path. Please note that all OCaml tools in the windows build are MinGW tools and can handle both / and \ in paths.

As @ejgallego suggested setting:
```
CAML_LD_LIBRARY_PATH="C:\bin\cygwin_coq64_ms_coqide\usr\x86_64-w64-mingw32\sys-root\mingw\libocaml\site-lib\stublibs"
```
fixes this although I tend to agree with @gares that it should work without this setting when the files are in the default folder. With the above setting I now get:
```
Error while linking ide/coqide_os_specific.cmo:
The external function `win32_interrupt' is not available
```
I am looking into this ...

MSoegtropIMC:
@ejgallego : so you are suggesting to base the debug build functionality on #8729? I thought about this since the old makefiles anyway will go away soon and it doesn't make that much sense to add new features to them. I might bother you quite a bit with questions though if I do this.

ejgallego:
[...]

PR #9977, opened 2019-04-21
[dune] Support for coqide as an ocamldebug target.

ejgallego:
Works fine here, cc: #9975

### DIFF
--- a/dev/dune
+++ b/dev/dune
@@ -12,6 +12,7 @@
   (deps dune-dbg.in
         ../checker/coqchk.bc
         ../topbin/coqc_bin.bc
+        ../ide/coqide_main.bc
         ; This is not enough as the call to `ocamlfind` will fail :/
         top_printers.cma)
   (action (copy dune-dbg.in dune-dbg)))
--- a/dev/dune-dbg.in
+++ b/dev/dune-dbg.in
@@ -6,6 +6,9 @@ case $1 in
         shift
         exe=_build/default/checker/coqchk.bc
         ;;
+    coqide)
+        exe=_build/default/ide/coqide_main.bc
+        ;;
     *)
         exe=_build/default/topbin/coqc_bin.bc
         ;;


Zimmi48 (commented):
Doesn't this change the required dependencies for running `dune build @dune-dbg`? Not sure there is a better solution though.

dev/dune
```diff
@@ -10,7 +10,9 @@
 (rule
   (targets dune-dbg)
   (deps dune-dbg.in
+        %{project_root}/theories/Init/Prelude.vo
```
SkySkimmer: I don't think that's a good idea, sometimes I want to debug why my change broke an init file.

ejgallego: OK

ejgallego:
> Doesn't this change the required dependencies for running `dune build @dune-dbg`? Not sure there is a better solution though.

We could have several files, I dunno, anyways this was just a PoC, so we can as well ignore it and have the devs do this by hand.

ejgallego:
I propose we merge, devs should be able to build CoqIDE anyways as to submit patches to Coq.

We will refine more later for sure.

Zimmi48 (commented):
I'd be quite happy to delegate this PR to @SkySkimmer if this is OK.

dev/dune
```diff
@@ -12,6 +12,8 @@
   (deps dune-dbg.in
         ../checker/coqchk.bc
         ../topbin/coqc_bin.bc
+        ../ide/coqide_main.bc
```
SkySkimmer: This dependency could be a pain if debugging somewhere without gtk deps. Not sure how often that happens.

dev/dune
```diff
@@ -12,6 +12,8 @@
...
+        ../topbin/coqtop_byte_bin.bc
```
SkySkimmer: Why coqtop?

SkySkimmer (approved):

dev/dune
ejgallego: The rationale here is that anyways to be a developer and submit a PR you need a system with GTK [otherwise you cannot test the full build], but I can refactor this if you want.

dev/dune
ejgallego: Oh, this likely sneaked in from a rebase as this PR dates back from the times of `coqtop` based debug.

SkySkimmer: >Oh, this likely sneaked in from a rebase as this PR dates back from the times of coqtop based debug.

Please remove then.

dev/dune
SkySkimmer: >The rationale here is that anyways to be a developer and submit a PR you need a system with GTK [otherwise you cannot test the full build]

IMO it's fine to let CI do that (the vast majority of PRs don't touch coqide) but I'm OK with the dependency if it makes things easier.

SkySkimmer (changes requested):
Waiting for removal of spurious coqtop dep, then I'll merge.

SkySkimmer (approved):

ejgallego:
Merci !

PR #9980, opened 2019-04-23
[ci] Set artifact expire date for all jobs.

### LINKED ISSUE #8735: [ci] [gitlab] `expire_in` not set in all targets.
As of today, the CI build produces some "keep forever" artifacts. IMHO, there is not point in doing so, and we should add the corresponding `expire_in` to all targets.

ejgallego:
Closes #8735 .

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -121,7 +121,7 @@ before_script:
     OPAM_VARIANT: "+flambda"
   artifacts:
     name: "$CI_JOB_NAME"
-    expire_in: 1 month
+    expire_in: 2 months
 
 # every non build job must set dependencies otherwise all build
 # artifacts are used together and we may get some random Coq. To that
@@ -140,6 +140,7 @@ before_script:
     name: "$CI_JOB_NAME"
     paths:
       - _install_ci/share/doc/coq/
+    expire_in: 2 months
 
 # set dependencies when using
 .test-suite-template:
@@ -159,6 +160,8 @@ before_script:
     when: on_failure
     paths:
       - test-suite/logs
+    # Gitlab doesn't support yet "expire_in: never" so we use the instance default
+    # expire_in: never
   variables:
     timeout: ""
 
@@ -177,7 +180,7 @@ before_script:
     name: "$CI_JOB_NAME.logs"
     paths:
       - coqchk.log
-    expire_in: 1 month
+    expire_in: 2 months
 
 .ci-template:
   stage: test
@@ -341,6 +344,8 @@ pkg:opam:
     when: on_failure
     paths:
       - nix-build-coq.drv-0/*/test-suite/logs
+    # Gitlab doesn't support yet "expire_in: never" so we use the instance default
+    # expire_in: never
 
 pkg:nix:deploy:
   extends: .nix-template
@@ -465,6 +470,8 @@ test-suite:egde:dune:dev:
     when: on_failure
     paths:
       - _build/default/test-suite/logs
+    # Gitlab doesn't support yet "expire_in: never" so we use the instance default
+    # expire_in: never
 
 test-suite:edge+trunk+make:
   stage: test


.gitlab-ci.yml
```diff
@@ -18,6 +18,9 @@ variables:
   OPAM_VARIANT: ""
   GIT_DEPTH: "10"
 
+artifacts:
+  expire_in: 3 days
```
SkySkimmer: Why not 1 week and remove the redundant settings?

SkySkimmer (changes requested):
Logs should be forever just like the main build log.

.gitlab-ci.yml
ejgallego: I prefer the information to be more "local".

ejgallego:
> Logs should be forever just like the main build log.

Ok, not a fan of this but will fix.

.gitlab-ci.yml
ejgallego: Will fix too.

ejgallego: Oh actually the global setting doesn't work:
```
jobs:artifacts config contains unknown keys: expire_in
You can also test your .gitlab-ci.yml in CI Lint
```

ejgallego:
Rebased; test-suite logs kept forever. It seems that it is not possible to set a global expire setting [except in the instance config]

ejgallego:
Turns out `expire_in: never` is not supported either; used 2 years.

SkySkimmer:
>Turns out expire_in: never is not supported either

Just don't put an `expire_in` key.

ejgallego:
> Just don't put an `expire_in` key.

The point of #8735 was indeed to have all targets documented; IMO that is better than relying on the unknown instance-wide setting.

This could be moved to a different gitlab [unlikely I know] and then logs could be lost.

Thus I prefer to keep it like this [we can put 5 years if you want?]

.gitlab-ci.yml
```diff
@@ -385,6 +388,7 @@ doc:refman:dune:
   artifacts:
     paths:
       - _build/default/doc/sphinx_build/html
+    expire_in: 2 months
```
SkySkimmer: Why not the same as dune-ci-template?

SkySkimmer:
>Thus I prefer to keep it like this [we can put 5 years if you want?]

IMO I see no gain, but 2 years should be OK so we can merge if you really want it.

ejgallego:
> IMO I see no gain, but 2 years should be OK so we can merge if you really want it.

Well I consider that ⊥⊆ date , so I prefer the more defined entry, but also I don't feel strongly about it; I dunno.

.gitlab-ci.yml
ejgallego: I used 2 months for docs and 1 month for binaries. I dunno, these values are arbitrary.

maximedenes:
@SkySkimmer can you merge if ready?

.gitlab-ci.yml
SkySkimmer: >I used 2 months for docs and 1 month for binaries. I dunno, these values are arbitrary.

Please pick one and use it for both.

ejgallego:
No rush on this.

ejgallego:
What we do with this ? Forever is not supported so we can either keep 5 years or I can just leave a comment and not `expire_in` attribute.

SkySkimmer (changes requested):
No need for a different delay for docs vs binaries IMO

SkySkimmer:
>What we do with this ? Forever is not supported so we can either keep 5 years or I can just leave a comment and not expire_in attribute.

I'll take whichever is implemented when you fix my change request.

ejgallego:
> I'll take whichever is implemented when you fix my change request.

Ok thanks, updated. I hope I got it right.

.gitlab-ci.yml
SkySkimmer: No need to repeat it when it's the same as the template.

ejgallego: well seen, let's see now.

SkySkimmer (approved):

Issue #9974, opened 2019-04-20
coq_makefile's generated Makefiles emit systematically a warning


gares:
Since d9fb8db86d5b4ddc79a207c5f0ac32eb98215e78 `coqc -opt` is deprecated but https://github.com/coq/coq/blob/master/tools/CoqMakefile.in#L737 uses that option, hence the warning (the same holds for `-byte`).

I suspect it is enough to remove `$(OPT)` from https://github.com/coq/coq/blob/master/tools/CoqMakefile.in#L179 to avoid the warning, but then one should also patch the value of `COQC` https://github.com/coq/coq/blob/master/tools/CoqMakefile.in#L81 in the conditional 
https://github.com/coq/coq/blob/master/tools/CoqMakefile.in#L163 I guess since the semantics of `-byte` is to use the bytecode version of coqc

gares:
I don't think 8.9 can be released with that warning.

ejgallego:
Actually is there a use case for calling `coqc.byte`?

We don't even build it in Dune.

gares:
Use case for building coqc.byte: Ocamldebug. Given that coqtop does not accept -compile, how do you debug coqc if you don't build it in bytecode?

About coq makefile: I don't see any use case other than the absence of native coqc. I don't recall, maybe if there is no native backend we name the bytecode binary without .byte.

ejgallego:
I was under the impression that people would call `coqc.byte` manually in that case instead of going thru coq_makefile, but if that's useful why not?

Note that both the makefiles and Dune will install `coqc.byte` as `coqc` if a native version is not available, thus `coqc` can be used by default as a synonym of "best".

gares:
I expect one calls coqc.byte by hand, but you wrote that you don't even build it, so I gave you a use case for building it.

Wrt CoqMakefile.in, since `coqc` means "best" the only fix needed is to not pass `$(OPT)` to coqc.

ejgallego:
I meant from coq makefile, sorry I was not clear. Yeah we should build it definitevely.

There are some subtlelies when plugins are involved tho, coqc.byte will need the CMA version.

maximedenes:
@gares are you going to work on fixing this? Can I mark you as the assignee?

gares:
Not this week I'm afraid.

maximedenes:
I'm afraid we are going to have to consider it is not a blocker for the beta.

Zimmi48:
[...]

PR #229, opened 2016-06-29
Bytecode compilation in a new 'make byte' rule apart from 'make world' 

letouzey:
Historically, we're accustomed to build coqtop and coqtop.byte in the same 'make world' run. That might be related with our need of the grammar.cma syntax extension, that used to build anyway the byte version of 50% of coq before the other 50% (the .ml4) could even be parsed. This could also be due to the importance of coqtop.byte and Drop for debugging.

But mixing the compilation of native code and byte code leads to delicate issues (races between ocamlc and ocamlopt about some .cmi, more details in the first commit message). 

And thanks to @ppedrot, grammar.cma is now completely self-contained.

So I think it's time to separate 'make world' (compilation of Coq with the best available OCaml compiler) from a distinct 'make byte' rule, dedicated to coqtop.byte and a few other stuff (bytecode plugins, printers.cma,...). More details in the message of the first commit. In everyday life, the main effect will be to compile with something like : 
 `make -j && make -j byte`
And then possibly: `make install install-byte`
Please do **not** use the shortcut `make -j world byte` that suffer from the races and crashes mentioned earlier.
In the second commit, I made corresponding changes to coq_makefile.
Best
Pierre

PR #9981, opened 2019-04-23
[dune] Build coqc.byte executable.

ejgallego:
This may be useful in a few cases, like testing compilation with
byte-plugins; I chose to install it globally tho it is more of a
developer target.

### DIFF
--- a/topbin/dune
+++ b/topbin/dune
@@ -28,6 +28,11 @@
  (libraries coq.toplevel)
  (link_flags -linkall))
 
+(install
+ (section bin)
+ (package coq)
+ (files (coqc_bin.bc as coqc.byte)))
+
 ; Workers
 (executables
  (names coqqueryworker_bin coqtacticworker_bin coqproofworker_bin)


topbin/dune
```diff
@@ -28,12 +28,25 @@
...
+ (action (copy# coqc_bin.ml %{targets})))
+
+(executable
+ (name coqc_byte_bin)
+ (public_name coqc.byte)
+ (package coq)
+ (modules coqc_byte_bin)
+ (libraries coq.toplevel)
+ (modes byte)
+ (link_flags -linkall))
```
rgrinberg: Isn't having the .bc target enough? If want to install it, then feel free to add an appropriate `(install ..)` stanza

ejgallego: Indeed thanks, I dunno what I was thinking....

Zimmi48 (commented):
It seems that this is reverting the change from #229 and #709. I don't have anything against it but this would be good to give a justification other than just "the maintainer of the build system is not the same person anymore".

ejgallego:
> It seems that this is reverting the change from #229 and #709. I don't have anything against it but this would be good to give a justification other than just "the maintainer of the build system is not the same person anymore".

Sorry @Zimmi48 I am not sure how are you reading this PR, but it certainly has nothing to do (at all!) with the PRs you mention. The setup discussed there is just not relevant to a Dune build as it works in a different way w.r.t. byte / opt stuff. Of course feel free to ask any question.

This PR is very trivial in that it will just build one more executable and install it, doesn't have any other impact [you can compare the install files]

Zimmi48:
I'm not blocking the PR and I can merge it without any delay, but the fact that the two build systems do not work in the same way doesn't change anything to the fact that from the user's perspective we'll have gone from a situation where both `coqc` and `coqc.byte` were installed, to a situation where only the former was installed, back to a situation where both are installed. The intermediate step was probably a mistake but this is nonetheless interesting to note how these decisions are taken...

maximedenes:
> I am not sure how are you reading this PR, but it certainly has nothing to do (at all!) with the PRs you mention

Of course it is related.

@Zimmi48 The decision of not building bytecode executables as part of the default `make` target was taken to avoid race conditions when calling `ocamlopt` and `ocamlc` concurrently. AFAIK, Dune does not suffer from these issues, due to its specific knowledge of the OCaml requirements. So making Dune build bytecode executables by default is not invalidating the previous decision, it is just taking advantage of an improvement and seems logical.

Zimmi48 (approved):

ejgallego:
> The intermediate step was probably a mistake but this is nonetheless interesting to note how these decisions are taken...

Théo, this PR was in response to a comment in #9974 .

But indeed trust me this PR is not related to the PRs you cite folks, in the sense the build rules are identical _after_ and _before_, the thing that changes is that `coqc.byte` is installed now.

You can build `coqc.byte` today, try `dune build topbin/coqc.bc`.

I am still not sure there is a use case for installing `coqc.byte` tho.

Zimmi48:
>Théo, this PR was in response to a comment in #9974.

Sometimes, it would be helpful to give that kind of information from the start. Note that it changes much in this case.

>the build rules are identical after and before, the thing that changes is that coqc.byte is installed now

I'm aware of this. I do understand what the PR does. I'm talking about the user's perspective here.

>I am still not sure there is a use case for installing coqc.byte tho.

I don't know what I should conclude from this, given that this is all the PR does.

maximedenes:
> Théo, this PR was in response to a comment in #9974 .

There you said:

> We don't even build it in Dune.

And now you seem to say the opposite...

ejgallego:
> Sometimes, it would be helpful to give that kind of information from the start. Not that it changes much in this case.

Indeed, sorry.

> I'm talking about the user's perspective here.

Well indeed my point is that PRs you mentioned were not done on any user consideration, but on problems with the build itself, as Maxime said.

> I don't know what I should conclude from this, given that this is all the PR does.

Me neither, this PR is if you want a RFC, the outcome could well be that `coqc.byte` is not useful at all. But given how little it costs we can just ship it [well, it costs about 40 MiB (!!)]

ejgallego:
> And now you seem to say the opposite...

We don't build it _by default_ I mean, the rules to build it are already there.

maximedenes:
> We don't build it _by default_ I mean, the rules to build it are already there.

Ok, but then Théo was right that it was the same question as for the `Makefile`s. But I think that's clarified now.

Zimmi48:
> > 
> 
> Well indeed my point is that PRs you mentioned were not done on any user consideration, but on problems with the build itself, as Maxime said.

I agree, and I think it was the main mistake at the time, to forget about the potential consequences for users. In the end, I think I have seen complaints from time to time about the absence of the byte outputs, so I'm still in favor of merging this PR if everybody agrees.

ejgallego:
Just a nit, the PRs we mention talk about `coqtop` and `coqtop.byte`, note that the Dune build has installed both since the beginning; so in a sense, the decision that Théo talks about was taken a few months ago in an implicit way.

Zimmi48:
Good point. And let's aim for consistency by doing the same thing for `coqc`. I will merge tomorrow if there are no barring comments.

Issue #9848, opened 2019-03-27
lia is slower than omega in large contexts


andres-erbsen:
#### Description of the problem

```coq
Require Import Lia Omega.

Goal True.
  do 1000 pose proof I.
  Time assert (42 = 42) as _ by Lia.lia. (* 10s *)
  Time assert (42 = 42) as _ by omega. (* .01s *)
  exact I.
Abort. (* Qed is also slow here *)

Goal True.
  do 1000 pose I.
  Time assert (42 = 42) as _ by Lia.lia. (* 10s *)
  Time assert (42 = 42) as _ by omega. (* .01s *)
  exact I.
Abort. (* Qed is also slow here *)
```

Rough breakdown shows that most of the time is spent in zify, but even xlia itself is slower than omega.

```coq
Require Import Lia Omega.

Goal True.
  do 1000 pose proof I.
  assert (42 = 42).
  Time progress zify_nat. (* 1.255s *)
  Time zify_positive. (* 1.121s *)
  Time zify_N. (* 1.242s *)
  Time Fail progress zify_nat. (* 1.058s *)
  Time zify_op. (* 0s *)
  Time unfold Z.succ in *. (* 1.147s *)
  Time unfold Z.pred in *. (* 1.265s *)
  Time xlia zchecker. (* .245s *)
  exact I.
Abort. (* Qed is also slow here *)
```

Cc #7878 #7872

#### Coq Version

8.9.0, a796822e5f57f74ff36e538fd2169f70a8c6c145 (the timings are from 8.9, on master the overall timings are very similar but the fine-grained breakdown cases vary within 50%).

samuelgruetter:
I ran into slow `lia` invocations as well and had to do manual workarounds such as `clear -H. lia.`

samuelgruetter:
While the above example might seem artificial, there are significant performance differences between lia and omega in practice, see https://github.com/coq/coq/issues/9921

ppedrot:
It seems `xlia` is generalizing over mere variables in a very inefficient way. I'll have a deeper look.

fajb:
I am glad @ppedrot is optimising the generalisation code!

Yet, in the present case, the main culprit seems elsewhere: the 1000x hypotheses should simply be dropped out of the proof-term. I'll look into this...

But, is-it really representative of the slow-down? True is kind of special.... 
@samuelgruetter , do you have at hand other slow instances ?
If you could pinpoint cases over Z where omega is faster that xlia that would be much helpful.
[...]

PR #9982, opened 2019-04-23
Performance improvements for `xlia`

ppedrot:
This PR mitigates part of the slowdown observed in #9848. It doesn't solve the `zify` issue, which is the major source of performance hit, but it makes `xlia` in itself faster. One of the changes is `xlia`-specific, the other is a fast path in the `keep` tactic.


ppedrot:
Something looks wrong with this PR...

ppedrot:
I removed the `keep` fast path to see whether this solves the observed issues.

PR #9964, opened 2019-04-16
Unreleased changelog folder

Zimmi48:
**Kind:** documentation infrastructure.

- [x] Corresponding documentation was updated

~~I'm submitting this as a WIP to demonstrate what the future of the unreleased changelog will be. This should be coupled with a script extracting the relevant changelog entries for a new release~~ (à la https://about.gitlab.com/2018/07/03/solving-gitlabs-changelog-conflict-crisis/), whose two advantages will be:
- no more CHANGES conflict
- no more mistake about which CHANGES section to put the entry in, especially when the PR is proposed for backporting but might be rejected, or if the PR has been waiting for a long time and the location has stalled.

The main feature demonstrated by this PR and which goes beyond what GitLab proposed is that unreleased changes are still visible in a central location (the deployed development version of the reference manual).

TODO

- [x] Remove commit with fake entries for demonstration purpose.
- [x] Wait for legacy Makefile to be removed, or update `Makefile.doc` as well.
- [x] Wait for #9939 to be merged first.

TODO for future PR

- [ ] Create the script for extracting the changelog for a new release.

PR #9987, opened 2019-04-24
Fix #9180 by reverting #9249 and #8187

### LINKED ISSUE #9180: Coq does not keep some notations

vbgl:
Fixes #9180

### DIFF
--- a/dev/doc/changes.md
+++ b/dev/doc/changes.md
@@ -203,12 +203,6 @@ Termops
 - Internal printing functions have been placed under the
   `Termops.Internal` namespace.
 
-Notations:
-
-- Notation.availability_of_notation is not anymore needed: if a
-  delimiter is needed, it is provided by Notation.uninterp_notation
-  which fails in case the notation is not available.
-
 ### Unit testing
 
 The test suite now allows writing unit tests against OCaml code in the Coq
--- a/doc/sphinx/user-extensions/syntax-extensions.rst
+++ b/doc/sphinx/user-extensions/syntax-extensions.rst
@@ -1040,8 +1040,6 @@ interpreted in the scope stack extended with the scope bound tokey.
    To remove a delimiting key of a scope, use the command
    :n:`Undelimit Scope @scope`
 
-.. _ArgumentScopes:
-
 Binding arguments of a constant to an interpretation scope
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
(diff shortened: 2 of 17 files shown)


maximedenes (approved):
I'm afraid this is the only solution for now.

Zimmi48:
I am not able to judge, but note that if it is believed that the problem in #9180 is fixable without redesigning the feature, then it would make sense to revert in v8.10 only...

Zimmi48:
BTW, it is quite puzzling to me that this large change did not have any entry in CHANGES.

vbgl:
The CHANGES entry has been moved to the manual (section Version 8.9).

Zimmi48:
OK so yet another case of a misplaced CHANGES entry that will be solved by #9964. I'll update #9939 to remove it.

vbgl:
Who shall assign to this PR? It seems ready to go.

Zimmi48:
@vbgl: From @herbelin's comment at https://github.com/coq/coq/issues/9180#issuecomment-487382878:

>In any case, a revert is the reasonable thing to do for 8.10, especially from the moment @vgbl succeeded in preparing a PR.

I didn't get what is Hugo's position wrt reverting in `v8.10` only or in `v8.10` + `master`. In any case, a solution could be to backport this PR to `v8.10` now (and then you don't need anyone else's approval to merge it) and to leave a few more days / weeks to take a decision on the revert in `master`.

ejgallego:
The revert should happen in master too, IIUC.

ejgallego:
> The revert should happen in master too, IIUC.

However indeed maybe it is a better strategy to leave #9180 open; so I suggest you follow @Zimmi48 's suggestion and apply in 8.10 -only for now until @herbelin / @ggonthier can provide more feedback on how they would like to proceed.

vbgl:
Well. I’ve opened this PR against the `master` for a reason.

I think there is no point in keeping master broken until there is a plan: let’s have a working master until there is a plan.

ejgallego:
> I think there is no point in keeping master broken until there is a plan: let’s have a working master until there is a plan.

Then, if we are going to close #9180 we need to improve the test suite so we don't reopen it in master by accident.

herbelin:
Note that, strictly speaking, regression #9180 on top of #8187 was already fixed in #9207 but the latter was not merged. What is not fixed by #9207 of the #8187 regressions are the "`%MS`" and the swap of priority wrt specialized notations and abbreviations. These are not reported in #9180 but in the body of the discussion for #9207.

So, probably, the good thing to do is to:
- close #9180, close #9207 
- add a regression test for #9180 *and* for the "`%MS`" and swap of priority wrt specialized notations and abbreviations given in the discussion of #9207 (ideally, someone should also explore #9258)
- open a new PR with #8187 + #9207, to which a fix to the `%MS`" and swap of priorities issues will eventually be integrated when the discussion converges (I'm confident that a fix could be done relatively easily, on which I'd like to get a collective support though)

Incidentally, the discussions at #9180 and #9207 gave rise to side (independent) PRs #9292, #9297, and #9311. The first two ones have been merged. The third one is still under discussion.

vbgl:
Test-case added.

maximedenes:
@ejgallego CI green, can you merge so that we can backport? Thanks!

ejgallego:
@herbelin are you OK with transforming your comment into an issue with the remaining TODO and merging this?

ejgallego:
I have opened https://github.com/coq/coq/issues/10023 , will merge later today as to leave some window for further comment, but still ensuring this PR is ready for tomorrow.

herbelin:
> @herbelin are you OK with transforming your comment into an issue with the remaining TODO and merging this?

Yes, very good. Then, I'll open a new PR after the revert is done.

ejgallego:
Thanks for the very detailed feedback and status report.

PR #6911, opened 2018-03-04
[ssr] Declare prenex implicits for `Some_inj`

anton-trunov:
**Kind:** documentation / bug fix.

Fixes / closes #6910 

Also fixes a minor typo in the documentation comment in ssrfun.v

PR #9995, opened 2019-04-24
fix `simpl_rel` and notations, `{pred T}` alias, `nonPropType` interface

ggonthier:
- Changed definition of `simpl_rel` to `T -> simpl_pred T`, so that
`inE` will now expand `a \in r b`, when `r := [rel x y | R]` to `R{b/x,
a/y}`, as the expanding coercion is now only inserted in the _last_
application.
The old definition made it possible to have a `simpl_rel >-> rel`
coercion that does not block expansion, but this can now be achieved
more economically with the `Arguments … /.` annotation.
-  Deleted the `[rel of P]` notation which is no longer needed with
the new `simpl_rel` definition, and was broken anyway.
- Added `relpre f R` definition of functional preimage of a notation.
- `comp` and `idfun` are now proper definitions, using the `Arguments
… /.` annotation to specify simplification on application.
- Added `{pred T}` syntax for the alias of `pred T` in the `pred_sort`
coercion class; deprecated the `pred_class` alias: one should either
use `pred_sort` in `Coercion` declarations, or `{pred T}` in type casts.
- Simplified the `predType` structure by removing the redundant
explicit `mem_pred` subfield, and replacing it with an interlocked
projection; deprecated `mkPredType`, now an alias for `PredType`.
- Added (and extensively documented) a `nonPropType` interface
matching types that do _not_ have sort `Prop`, and used it to remove
the non-standard maximal implicits annotation on `Some_inj` introduced
in #6911 by @anton-trunov.
- Documented the design of the four structures used to control the
matching of `inE` and related predicate rewriting lemmas.
- Used `only printing` annotations to get rid of token concatenation
hacks.
- Fixed boolean and general `if b return t then …` notation so that
`b` is bound in `t`. This is a minor source of incompatibility for
misuses of this syntax when `b` is _not_ bound in `t`, and `(if b then
…) : t` should have been used instead.
- Reserved all `ssreflect`, `ssrfun` and `ssrbool` notation at the top
of the file, adding some printing boxes, and removing some spurious
`[pred .. => ..]` reserved notation.
- Fixed parsing precedence and format of `<hidden n>` notation, and
declared and put it in an explicit `ssr_scope`. @gares check this is
fine with you.

### DIFF
--- a/plugins/ssr/ssreflect.v
+++ b/plugins/ssr/ssreflect.v
@@ -28,6 +28,11 @@ Declare ML Module "ssreflect_plugin".
         argumentType c == the T such that c : forall x : T, P x.
           returnType c == the R such that c : T -> R.
      {type of c for s} == P s where c : forall x : T, P x.
+           nonPropType == an interface for non-Prop Types: a nonPropType coerces
+                          to a Type, and only types that do _not_ have sort
+                          Prop are canonical nonPropType instances. This is
+                          useful for applied views (see mid-file comment).
+             notProp T == the nonPropType instance for type T.
            phantom T v == singleton type with inhabitant Phantom T v.
                phant T == singleton type with inhabitant Phant v.
                  =^~ r == the converse of rewriting rule r (e.g., in a
@@ -57,8 +62,6 @@ Declare ML Module "ssreflect_plugin".
  More information about these definitions and their use can be found in the
  ssreflect manual, and in specific comments below.                           **)
 
-
-
 Set Implicit Arguments.
 Unset Strict Implicit.
 Unset Printing Implicit Defensive.
@@ -77,13 +80,47 @@ Reserved Notation "(* 69 *)" (at level 69).
 (**  Non ambiguous keyword to check if the SsrSyntax module is imported  **)
 Reserved Notation "(* Use to test if 'SsrSyntax_is_Imported' *)" (at level 8).
 
-Reserved Notation "<hidden n >" (at level 200).
+Reserved Notation "<hidden n >" (at level 0, n at level 0,
+  format "<hidden  n >").
 Reserved Notation "T (* n *)" (at level 200, format "T  (* n *)").
 
 End SsrSyntax.
 
 Export SsrMatchingSyntax.
 Export SsrSyntax.
+
+(** Save primitive notation that will be overloaded. **)
+Local Notation CoqGenericIf c vT vF := (if c then vT else vF) (only parsing).
+Local Notation CoqGenericDependentIf c x R vT vF :=
+  (if c as x return R then vT else vF) (only parsing).
+Local Notation CoqCast x T := (x : T) (only parsing).
+
+(** Reserve notation that introduced in this file. **)
+Reserved Notation "'if' c 'then' vT 'else' vF" (at level 200,
+  c, vT, vF at level 200, only parsing).
+Reserved Notation "'if' c 'return' R 'then' vT 'else' vF" (at level 200,
+  c, R, vT, vF at level 200, only parsing).
+Reserved Notation "'if' c 'as' x 'return' R 'then' vT 'else' vF" (at level 200,
+  c, R, vT, vF at level 200, x ident, only parsing).
+
+Reserved Notation "x : T" (at level 100, right associativity,
+  format "'[hv' x '/ '  :  T ']'").
+Reserved Notation "T : 'Type'" (at level 100, format "T  :  'Type'").
+Reserved Notation "P : 'Prop'" (at level 100, format "P  :  'Prop'").
+
+Reserved Notation "[ 'the' sT 'of' v 'by' f ]" (at level 0,
+  format "[ 'the'  sT  'of'  v  'by'  f ]").
+Reserved Notation "[ 'the' sT 'of' v ]" (at level 0,
+  format "[ 'the'  sT  'of'  v ]").
+Reserved Notation "{ 'type' 'of' c 'for' s }" (at level 0,
+  format "{ 'type'  'of'  c  'for'  s }").
+
+Reserved Notation "=^~ r" (at level 100, format "=^~  r").
+
+Reserved Notation "[ 'unlockable' 'of' C ]" (at level 0,
+  format "[ 'unlockable'  'of'  C ]").
+Reserved Notation "[ 'unlockable' 'fun' C ]" (at level 0,
+  format "[ 'unlockable'  'fun'  C ]").
 
 (**
  To define notations for tactic in intro patterns.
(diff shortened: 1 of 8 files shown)


gares:
Mtac failure seem unrelated

plugins/ssr/ssreflect.v
```diff
@@ -500,3 +529,105 @@ Hint View for apply/ iffRLn|2 iffLRn|2 iffRL|2 iffLR|2.
...
+ of the initial canonical instance, and not on the instance itself as we do
+ in mathcomp/matrix and mathcomp/vector, because we want the unification to
+ fail when T has sort Prop. If both the test_of _and_ the result check
+ unifications were done as part of the structure telescope then the latter
+ would be a sub-problem of the former, and thus failing the check would merely
+ make the test_of unification backtrack and delta-expand and we would not get
+ failure.
+ **)
+
+#[universes(template)]
```
gares: I'm wondering why the Structures are annotate with `template`

gares (commented):
Looks good to me. I'd love to have a test in `test-suite` for an instance of `nonPropType` (or, is `Some_inj` used as a view in the few ssr*.v files? it does not seem to)

Notes:
- should `nonPropType` also non-unify with `SProp`? If so we should add a test for this (`SProp` needs a command line flags in order to be available, but in `test-suite` we can set flags on a per file basis, so it should be doable)
- the `universe(template)` attribute needs to be clarified
- the CHANGES file seems not to reflect the recent creation of the `v8.10` branch (eg this PR adds lines in the Unreleased section, but this section contains things that should be in a 8.10 section CC @vbgl )

plugins/ssr/ssreflect.v
ggonthier: "Just to be on the safe side", mostly, as I'm not sure about the semantics of the annotation.
If these only are meaningful for types with `Type` parameters then these are indeed useless, but then the `external_view` type (shouldn't it be a `Variant`?) does have a `template` annotation. I suppose the annotation on `test_of` is almost certainly useless, since that `Structure` has no parameter _or_ component in `Type`.

ggonthier:
I'm not sure about `SProp`; is it supposed to be included in `Prop`? In any case, I don't think this really matters for the intended use cases.
I can make the mechanism open-ended by adding intermediate delta layers before and/or after `maybeProp`. This would let users adjust the set of excluded types by defining new instances of `test_of`, perhaps even using an auxiliary `Structure` to pattern-match on the type itself, e.g., to specifically exclude `reflect`. I thought however that `nonPropType` was good enough for known use cases, and simpler to explain (implementation is a whole other story!).

ggonthier:
I'll see about adding something to `test-suite`. `Some_inj` is used in `seq`, but not as a view; indeed this use made me realise that putting non-standard implicits on `Some_inj` was a mistake, and got me thinking about finding another way to address the issue raised by @anton-trunov.

ggonthier:
I'm not sure what to do about `CHANGES`. I assumed this was a bit late for 8.10, so the "unreleased" category seemed fine for me, and that the owners of other items would see to their migration in the 8.10 subdivision?

gares:
> I'm not sure what to do about `CHANGES`

I'll deal with this at merge time myself

> I'm not sure about `SProp`; is it supposed to be included in `Prop`? In any case, I don't think this really matters for the intended use cases.

CC @SkySkimmer knows the detail. I think that the terms that live in SProp can live in Prop, but not the converse and that there is no cumulativity between the two sorts. 
My intuition is that, as a user, SProp is like Prop but with more proof irrelevance. So some statements that today I put in Prop, maybe tomorrow I'll put them in SProp. Hence my question about `notProp`, since I'd expect it to also rule out SProp.

plugins/ssr/ssreflect.v
gares: Cc @maximedenes @mattam82 (given #9918 they surely know better).

SkySkimmer:
There's an explicit lift from SProp to Prop but no implicit lift (cumulativity).
I don't really understand what the new thing is, but I think you'd get a typing error at some point if you tried to use it with SProp due to the non cumulativity. (perhaps needing to wait until Qed / unset https://coq.github.io/doc/master/refman/addendum/sprop.html#coq:opt.elaboration-strictprop-cumulativity)

plugins/ssr/ssreflect.v
SkySkimmer: I added the annotation mechanically when I added the auto template warning, hoping to have a look at individual cases at some point. The warning was later disabled-by-default due to low value and false positives.
AFAIK template is only useful when there's a Type parameter (not component), or when you rely on automatic lowering.

Automatic lowering: this would work without the attribute
~~~coq
#[universes(notemplate)]
Inductive foo := .
Fail Check foo : Prop. (* error foo is in Type *)
~~~

I would say don't bother annotating new types. I would almost consider reverting the existing annotations but I don't feel it's worth the time to think about it.

ggonthier:
There's no chance of a delayed type error, because a type being accepted as a `Prop` causes the unification to fail. So if Coq allows an `SProp` to be assigned to a `Prop` evar, then `nonPropType` will also exclude `SProp`, regardless of whether this assignment would be rejected on `Qed` (it will never get to this, since the unification will have failed and the assignment discarded). If not, then `nonPropType` will accept `SProp`s, so a separate (or extended) version would be needed to have that functionality.

SkySkimmer:
BTW please don't open branches on the Coq repo itself.

gares (approved):
Look OK to me.

gares:
> BTW please don't open branches on the Coq repo itself.

I'll delete the branch after merging, next time @ggonthier please use the fork button to obtain a copy of coq in you github namespace, eg github.com/ggconthier/coq, push your branch there and do the PR. I guess you already have a coq clone on your hard drive, I recommend (after you hit the fork button in coq/coq) you do the git gimmicks
```
git remote rename origin upstream
git remote add origin git@github.com:ggonthier/coq.git
git remote update
```
In this way the standard destination for a push (`origin`) points your your own space, rather than the official repo (`upstream`). 

WRT this PR, I think we are good, If I head nothing I'll merge later today.

FTR the `nonProp` thing solves a quite annoying issue @bgregoir has been reporting privately many times, I CC him here to make him aware of this.

ggonthier:
@gares, @SkySkimmer : thank you for the specific instructions. You might want to include those in `coq/CONTRIBUTING.md`, which at present quite literally instructs contributors to fork the main repository....
  Regarding this PR, I've got an additional outstanding commit fixing a multiple-coercion issue that came up while I was propagating the interface change to MathComp (math-comp/math-comp#337).
  As this commit also removes rather than deprecates two legacy `ssrbool` definitions (`mkPredType` and `pred_class`) I'm holding it until math-comp/math-comp#337 can be merged (it's still held up by minor v8.7 compatibility issues, which I hope to clear up today).
  Finally, `CONTRIBUTING.md` say I should add myself to `CREDITS`; I'm already in there, but the dates are wrong, so I guess I'll just fix that?

Zimmi48:
>which at present quite literally instructs contributors to fork the main repository

You may be confusing "fork" and "clone" but the instruction of forking the repository is indeed what we request from contributors and what @gares just explained above. We may need to clarify further what that means.

>Finally, CONTRIBUTING.md say I should add myself to CREDITS; I'm already in there, but the dates are wrong, so I guess I'll just fix that?

Yes, if you don't mind. Note though that because not many people actually follow this advice, we will probably switch to another setup wrt `CREDITS` in the future.

gares:
> I'm holding it until

OK, then just ping me back when things are set.

ggonthier:
@gares: completed, squashed and rebased; hopefully CI holds (math-comp itself was also updated).

gares:
Are we good? (I see you said go, but then pushed again)

PR #9997, opened 2019-04-25
CoqIDE: fix open-file dialog and icons on macOS

vbgl:
This fixes a few known issues of CoqIDE from the macOS “app”: crashes when using GTK+ dialogs (file open, choose background color…) and missing icons.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -52,7 +52,7 @@ jobs:
   - script: |
       set -e
       brew update
-      brew install gnu-time opam pkg-config gtksourceview3
+      brew install gnu-time opam pkg-config gtksourceview3 adwaita-icon-theme
       pip3 install macpack
     displayName: 'Install system dependencies'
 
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -52,7 +52,7 @@ jobs:
   - script: |
       set -e
       brew update
-      brew install gnu-time opam pkg-config gtksourceview3
+      brew install gnu-time opam pkg-config gtksourceview3 adwaita-icon-theme
```
SkySkimmer: Do we need to update install instructions somewhere?

gares (approved):

azure-pipelines.yml
gares: I don't think so, this package is there so that we can copy come icons in the bundle (that is supposed to work even if you don't have brew)

vbgl:
Can someone please assign themselves to this PR? I think it is ready to be merged. Thanks.

